PR #48, opened 2026-04-27
fix: Use HGVS spec compact form for allele Display

### LINKED ISSUE #46: Fix allele Display to use HGVS spec-correct compact form
# Fix allele Display to use HGVS spec-correct compact form

## Summary

The `Display` impl for `AlleleVariant` (cis phase) currently emits the expanded form with repeated accessions:

```
[NG_008617.1:g.100A>G;NG_008617.1:g.200C>T]
```

The HGVS spec defines the compact form with the accession written once:

```
NG_008617.1:g.[100A>G;200C>T]
```

## Current behavior (verified empirically)

```python
>>> import ferro_hgvs
>>> v = ferro_hgvs.parse("NG_008617.1:g.[100A>G;200C>T]")
>>> str(v)
'[NG_008617.1:g.100A>G;NG_008617.1:g.200C>T]'
```

## Expected behavior

```python
>>> str(v)
'NG_008617.1:g.[100A>G;200C>T]'
```

## Spec reference

From [hgvs-nomenclature.org/stable/recommendations/DNA/alleles](https://hgvs-nomenclature.org/stable/recommendations/DNA/alleles/):

> **Syntax (cis):** `sequence_identifier ":" coordinate_type ".[" position_edit ";" position_edit "]"`
>
> **Examples:**
> - `NC_000001.11:g.[123G>A;345del]`
> - `NM_004006.2:c.[2376G>C;3103del]`

The expanded form (full descriptions inside brackets) only appears in the spec for **cross-reference** trans alleles where each variant is on a different accession:

> `[NM_004004.2:c.35del];[NM_006783.1:c.689_690insT]`

## Proposed fix

In `src/hgvs/variant.rs`, `impl fmt::Display for AlleleVariant`, the `Cis` arm should:

1. Check if all sub-variants share the same accession and coordinate type
2. If yes → emit compact form: `{accession}:{coord_type}.[{edit1};{edit2}]`
3. If no (mixed accessions) → emit the current expanded form (which is correct for that case)

The same logic should apply to trans (`[edit1];[edit2]` with shared accession prefix) and unknown phase.

## Edge cases

- **Mixed accessions within cis allele:** `[NM_001.1:c.100A>G;NM_002.1:c.200C>T]` — keep expanded form (this is the correct representation)
- **Mixed coordinate types:** shouldn't occur per spec ("descriptions combining variants based on different reference sequence types should not be used"), but handle gracefully with expanded form
- **Mosaic/chimeric:** `var1/var2` and `var1//var2` — these don't use brackets, so the current `Display` implementation showing full descriptions is fine

## Impact

- Fixes round-trip: `parse(str(parse(s)))` will produce the canonical form
- Affects `str()` output in Python bindings (no API change needed, just formatting)
- May affect downstream tests that assert exact string output

msto:
# PR: Use HGVS spec compact form for allele Display

**Closes #46**

## Summary

When all sub-variants in an allele share the same accession and coordinate type, `AlleleVariant::Display` now emits the compact form recommended by the HGVS nomenclature:

| Phase | Before | After |
|-------|--------|-------|
| Cis | `[NM_000088.3:c.100A>G;NM_000088.3:c.200C>T]` | `NM_000088.3:c.[100A>G;200C>T]` |
| Trans | `[NM_000088.3:c.100A>G];[NM_000088.3:c.200C>T]` | `NM_000088.3:c.[100A>G];[200C>T]` |
| Unknown | `[NM_000088.3:c.100A>G(;)NM_000088.3:c.200C>T]` | `NM_000088.3:c.100A>G(;)200C>T` |

When sub-variants have **different accessions** (cross-reference alleles), the expanded form is retained — that's the correct representation for that case.

Mosaic (`/`) and chimeric (`//`) phases are unchanged (they don't use bracket syntax).

## Motivation

Downstream consumers of ferro-hgvs use the compact form in their outputs (matching mutalyzer's behavior). Aligning ferro's output with the spec's recommended syntax avoids needing a formatting shim downstream.

## Spec reference

From [hgvs-nomenclature.org/stable/recommendations/DNA/alleles](https://hgvs-nomenclature.org/stable/recommendations/DNA/alleles/):

> **Syntax (cis):** `sequence_identifier ":" coordinate_type ".[" position_edit ";" position_edit "]"`
>
> **Examples:** `NC_000001.11:g.[123G>A;345del]`, `NM_004006.2:c.[2376G>C;3103del]`

The expanded form appears in the spec only for cross-reference trans alleles where each variant references a different sequence.

## Implementation

Added two helpers to `impl HgvsVariant`:

- **`fmt_loc_edit(&self, f)`** — writes just the position+edit part (e.g., `100A>G`) without the accession prefix
- **`all_share_accession_and_type(variants)`** — checks whether compact form is applicable

The `Display` impl for `AlleleVariant` now branches on this check, using compact form when possible and falling back to the expanded form for mixed-accession alleles.

## Testing

- All 1850 lib tests pass (updated existing tests to expect compact output)
- Updated 1 integration test (`normalize_tests.rs`) that matched on `c.9del` substring
- Added new test `test_allele_mixed_accession_uses_expanded_form` covering the fallback path
- Verified Python bindings produce compact form:
  ```python
  >>> str(ferro_hgvs.parse("NG_008617.1:g.[100A>G;200C>T]"))
  'NG_008617.1:g.[100A>G;200C>T]'
  ```

## Files changed

- `src/hgvs/variant.rs` — new helpers + updated `AlleleVariant::Display`
- `src/hgvs/parser/variant.rs` — updated test expectations
- `src/normalize/mod.rs` — updated test assertion
- `tests/normalize_tests.rs` — updated integration test assertion

### DIFF
--- a/src/hgvs/variant.rs
+++ b/src/hgvs/variant.rs
@@ -511,7 +511,10 @@ impl fmt::Display for AlleleVariant {
                 }
             }
             AllelePhase::Unknown => {
-                if use_compact_form(&self.variants) {
+                // Unknown-phase compact form has no surrounding brackets, so a
+                // singleton would print as `ACC:type.edit` and lose the allele
+                // wrapper. Only use compact form when there are 2+ sub-variants.
+                if self.variants.len() > 1 && use_compact_form(&self.variants) {
                     // Compact form: ACC:g.edit1(;)edit2
                     write_compact_prefix(f, &self.variants[0])?;
                     for (i, v) in self.variants.iter().enumerate() {
@@ -1176,6 +1179,32 @@ mod tests {
         assert_eq!(
             format!("{}", HgvsVariant::Allele(allele)),
             "NM_000088.3:c.[100A>G];[200A>C];[300A>T]"
+        );
+    }
+
+    #[test]
+    fn test_allele_unknown_phase_singleton_keeps_wrapper() {
+        use crate::hgvs::location::CdsPos;
+
+        // A singleton unknown-phase allele must keep the bracketed wrapper;
+        // the compact form `ACC:c.edit` would be indistinguishable from a
+        // bare variant and would not round-trip.
+        let var = HgvsVariant::Cds(CdsVariant {
+            accession: Accession::new("NM", "000088", Some(3)),
+            gene_symbol: None,
+            loc_edit: LocEdit::new(
+                CdsInterval::point(CdsPos::new(100)),
+                NaEdit::Substitution {
+                    reference: Base::A,
+                    alternative: Base::G,
+                },
+            ),
+        });
+
+        let allele = AlleleVariant::unknown_phase(vec![var]);
+        assert_eq!(
+            format!("{}", HgvsVariant::Allele(allele)),
+            "[NM_000088.3:c.100A>G]"
         );
     }
 
--- a/src/hgvs/variant.rs
+++ b/src/hgvs/variant.rs
@@ -437,27 +437,36 @@ impl AlleleVariant {
     }
 }
 
+/// Write the `ACC:<type>.` prefix for compact allele form.
+///
+/// Per HGVS spec, bracketed `?` is only valid as a whole-allele marker (`[?]`),
+/// never mixed with concrete edits inside the same bracket. Compact form is
+/// suppressed when any sub-variant carries the per-variant `?` (e.g. `c.?`).
+fn use_compact_form(variants: &[HgvsVariant]) -> bool {
+    !variants.is_empty()
+        && HgvsVariant::all_share_accession_and_type(variants)
+        && !variants.iter().any(HgvsVariant::is_loc_edit_unknown)
+}
+
+fn write_compact_prefix(f: &mut fmt::Formatter<'_>, first: &HgvsVariant) -> fmt::Result {
+    write!(
+        f,
+        "{}:{}.",
+        first
+            .accession()
+            .expect("compact form requires an accession; guarded by all_share_accession_and_type"),
+        first.variant_type()
+    )
+}
+
 impl fmt::Display for AlleleVariant {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        // Per HGVS spec, bracketed `?` is only valid as a whole-allele marker (`[?]`),
-        // never mixed with concrete edits inside the same bracket. Suppress compact
-        // form when any sub-variant carries the per-variant `?` (e.g. `c.?`, `r.?`).
-        let compact = HgvsVariant::all_share_accession_and_type(&self.variants)
-            && !self.variants.iter().any(HgvsVariant::is_loc_edit_unknown);
-
         match self.phase {
             AllelePhase::Cis => {
-                if compact && !self.variants.is_empty() {
+                if use_compact_form(&self.variants) {
                     // Compact form: ACC:g.[edit1;edit2]
-                    let first = &self.variants[0];
-                    write!(
-                        f,
-                        "{}:{}.[",
-                        first
-                            .accession()
-                            .expect("compact form requires an accession; guarded by all_share_accession_and_type"),
-                        first.variant_type()
-                    )?;
+                    write_compact_prefix(f, &self.variants[0])?;
+                    write!(f, "[")?;
                     for (i, v) in self.variants.iter().enumerate() {
                         if i > 0 {
                             write!(f, ";")?;
@@ -478,17 +487,9 @@ impl fmt::Display for AlleleVariant {
                 }
             }
             AllelePhase::Trans => {
-                if compact && !self.variants.is_empty() {
+                if use_compact_form(&self.variants) {
                     // Compact form: ACC:g.[edit1];[edit2]
-                    let first = &self.variants[0];
-                    write!(
-                        f,
-                        "{}:{}.",
-                        first
-                            .accession()
-                            .expect("compact form requires an accession; guarded by all_share_accession_and_type"),
-                        first.variant_type()
-                    )?;
+                    write_compact_prefix(f, &self.variants[0])?;
                     for (i, v) in self.variants.iter().enumerate() {
                         if i > 0 {
                             write!(f, ";")?;
@@ -510,17 +511,9 @@ impl fmt::Display for AlleleVariant {
                 }
             }
             AllelePhase::Unknown => {
-                if compact && !self.variants.is_empty() {
+                if use_compact_form(&self.variants) {
                     // Compact form: ACC:g.edit1(;)edit2
-                    let first = &self.variants[0];
-                    write!(
-                        f,
-                        "{}:{}.",
-                        first
-                            .accession()
-                            .expect("compact form requires an accession; guarded by all_share_accession_and_type"),
-                        first.variant_type()
-                    )?;
+                    write_compact_prefix(f, &self.variants[0])?;
                     for (i, v) in self.variants.iter().enumerate() {
                         if i > 0 {
                             write!(f, "(;)")?;
@@ -818,7 +811,6 @@ pub struct TxVariant {
 }
 
 impl TxVariant {
-    /// Format just the position+edit portion (without `accession:n.` prefix).
     pub(crate) fn fmt_loc_edit(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}", self.loc_edit)
     }
@@ -920,7 +912,6 @@ pub struct MtVariant {
 }
 
 impl MtVariant {
-    /// Format just the position+edit portion (without `accession:m.` prefix).
     pub(crate) fn fmt_loc_edit(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}", self.loc_edit)
     }
@@ -945,7 +936,6 @@ pub struct CircularVariant {
 }
 
 impl CircularVariant {
-    /// Format just the position+edit portion (without `accession:o.` prefix).
     pub(crate) fn fmt_loc_edit(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}", self.loc_edit)
     }
--- a/src/hgvs/variant.rs
+++ b/src/hgvs/variant.rs
@@ -722,9 +722,7 @@ impl HgvsVariant {
 
         // Don't use compact form for types that aren't simple coordinate-based variants,
         // or when the first variant has no accession (e.g. NullAllele, UnknownAllele)
-        if first_acc.is_none()
-            || matches!(first_type, "allele" | "null" | "unknown" | "r::r")
-        {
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai review

src/hgvs/variant.rs
```diff
@@ -439,41 +439,99 @@ impl AlleleVariant {
...
-                for (i, v) in self.variants.iter().enumerate() {
-                    if i > 0 {
-                        write!(f, ";")?;
+                if compact && !self.variants.is_empty() {
+                    // Compact form: ACC:g.[edit1;edit2]
+                    let first = &self.variants[0];
+                    write!(
+                        f,
+                        "{}:{}.[",
+                        first.accession().unwrap(),
```
nh13: **suggestion:** use `.expect` instead of `.unwrap` here and throughout

src/hgvs/variant.rs
```diff
@@ -439,41 +439,99 @@ impl AlleleVariant {
...
+                        first.variant_type()
+                    )?;
+                    for (i, v) in self.variants.iter().enumerate() {
+                        if i > 0 {
+                            write!(f, ";")?;
+                        }
+                        v.fmt_loc_edit(f)?;
```
nh13, line 474: **suggestion:** Move this into a method on each typed variant struct (e.g., `GenomeVariant::fmt_loc_edit`, `ProteinVariant::fmt_loc_edit`) and have both `impl    
  Display` and `HgvsVariant::fmt_loc_edit` delegate to it. That way the special-case logic lives once per struct, next to its `Display`.

src/hgvs/variant.rs
```diff
@@ -584,6 +642,94 @@ impl HgvsVariant {
...
         matches!(self, HgvsVariant::UnknownAllele)
     }
+
+    /// Format just the position+edit portion (without accession and coordinate prefix).
+    ///
+    /// For `NM_000088.3:c.459A>G`, this writes `459A>G`.
+    /// Used by allele Display to emit the spec-correct compact form.
+    ///
+    /// Returns `Err` for variant types that don't support this (Allele, NullAllele, etc.)
+    pub fn fmt_loc_edit(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
nh13: **nitpick:** make `pub(crate)`

src/hgvs/variant.rs
```diff
@@ -584,6 +642,94 @@ impl HgvsVariant {
...
+            HgvsVariant::Tx(v) => write!(f, "{}", v.loc_edit),
+            HgvsVariant::Rna(v) => match &v.loc_edit.edit {
+                crate::hgvs::uncertainty::Mu::Certain(edit) => {
+                    if edit.is_whole_entity() {
+                        write!(f, "{}", edit.to_rna_string())
+                    } else {
+                        write!(f, "{}{}", v.loc_edit.location, edit.to_rna_string())
+                    }
+                }
+                crate::hgvs::uncertainty::Mu::Uncertain(edit) => {
```
nh13: **question:** is `ACC:r.[?;100a>g]` possible?  Is that sensical?

src/hgvs/variant.rs
```diff
@@ -584,6 +642,94 @@ impl HgvsVariant {
     pub fn is_unknown_allele(&self) -> bool {
         matches!(self, HgvsVariant::UnknownAllele)
     }
+
+    /// Format just the position+edit portion (without accession and coordinate prefix).
+    ///
+    /// For `NM_000088.3:c.459A>G`, this writes `459A>G`.
+    /// Used by allele Display to emit the spec-correct compact form.
+    ///
+    /// Returns `Err` for variant types that don't support this (Allele, NullAllele, etc.)
```
nh13: **question:** does it return `Err`, or does it fall through to `Display`?

src/hgvs/variant.rs
```diff
@@ -439,41 +439,99 @@ impl AlleleVariant {
...
                 }
-                write!(f, "]")
             }
             AllelePhase::Trans => {
-                // [var1];[var2]
-                for (i, v) in self.variants.iter().enumerate() {
-                    if i > 0 {
-                        write!(f, ";")?;
+                if compact && !self.variants.is_empty() {
+                    // Compact form: ACC:g.[edit1];[edit2]
```
nh13, line 491: **suggestion:** add test cases for three or more trans variants (eg. `ACC:c.[e1];[e2];[e3]`).  Same for cis

nh13 (changes requested):

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

src/hgvs/variant.rs
msto: Good catch — swept the spec to check.

The HGVS spec uses bracketed `?` *only* as a whole-allele marker, never mixed with concrete edits inside the same bracket. Every example across DNA / RNA / protein puts `?` alone in its own trans bracket:

- DNA: `NM_004006.2:c.[2376G>C];[?]`
- RNA: `NM_004006.2:r.[76a>u];[?]`
- Protein: `NP_003997.1:p.[(Ser68Arg)];[?]`

The [RNA alleles page](https://hgvs-nomenclature.org/stable/recommendations/RNA/alleles/) is most explicit: *"When a variant is expected but unidentified on one allele, use `r.?` inside the bracket"* — i.e. `[?]` means *the whole second allele's variant is unknown*, not "one of several edits in a cis bracket." There are no spec examples of `[?;100a>g]` or `[100a>g;?]`.

So `ACC:r.[?;100a>g]` is technically constructible via our Rust API but isn't sanctioned by the spec, and it's visually ambiguous with the legitimate `r.[?]` whole-allele marker.

Two distinct meanings of `?` were colliding in our code:

1. **Whole-allele `?`** (the spec-sanctioned use) — represented by `HgvsVariant::UnknownAllele` in our enum. Already excluded from compact form by the `"unknown"` filter in `all_share_accession_and_type`.
2. **`r.?` / `c.?` / `p.?`** (per-variant "not analyzed / cannot be reliably predicted") — represented by `RnaVariant { Mu::Unknown }` etc. *Was* eligible for the compact branch, which is the case you flagged.

Added a guard in 6255bb2: `HgvsVariant::is_loc_edit_unknown()` returns `true` for any of `g.?` / `c.?` / `n.?` / `r.?` / `p.?` / `m.?` / `o.?`, and `AlleleVariant::Display` falls back to expanded form when any sub-variant matches. New regression test asserts `[NM_000088.3:c.100A>G];[NM_000088.3:c.?]` (expanded) rather than the compact form.

src/hgvs/variant.rs
msto: Falls through to `Display` — none of the arms return `Err`. The `Allele` / `RnaFusion` / `NullAllele` / `UnknownAllele` arms emit their full `Display` output (e.g. an inner `Allele` would emit `[ACC:c.100A>G;ACC:c.200C>T]`, including its own accession prefix). Updated the docstring in 6255bb2, and noted that callers guard against these types via `all_share_accession_and_type` so the fallback is never reached in practice.

If you'd prefer it actually return `Err` for those variants (and bubble up if a future caller forgets the guard), I can switch to that.

src/hgvs/variant.rs
msto: Done in 6255bb2 — replaced `.unwrap()` with `.expect("compact form requires an accession; guarded by all_share_accession_and_type")` in all three branches (cis/trans/unknown).

src/hgvs/variant.rs
msto: Done in 6255bb2 — added `fmt_loc_edit` methods to `GenomeVariant`, `CdsVariant`, `TxVariant`, `RnaVariant`, `ProteinVariant`, `MtVariant`, and `CircularVariant`. Each typed `Display` writes `accession:type.` then delegates to its own `fmt_loc_edit`; `HgvsVariant::fmt_loc_edit` delegates to the same methods. Special-case logic now lives once per struct, next to its `Display`.

src/hgvs/variant.rs
msto: Done in 6255bb2.

src/hgvs/variant.rs
msto: Done in 6255bb2 — added `test_allele_cis_three_variants` (`NM_000088.3:c.[100A>G;200A>C;300A>T]`) and `test_allele_trans_three_variants` (`NM_000088.3:c.[100A>G];[200A>C];[300A>T]`).

src/hgvs/variant.rs
```diff
@@ -437,43 +437,104 @@ impl AlleleVariant {
...
+                    // Compact form: ACC:g.edit1(;)edit2
+                    write_compact_prefix(f, &self.variants[0])?;
+                    for (i, v) in self.variants.iter().enumerate() {
+                        if i > 0 {
+                            write!(f, "(;)")?;
+                        }
+                        v.fmt_loc_edit(f)?;
                     }
-                    write!(f, "{}", v)?;
+                    Ok(())
```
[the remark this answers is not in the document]
nh13: @msto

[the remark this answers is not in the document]
msto: @coderabbitai are we good now?

msto:
@coderabbitai review

nh13:
@coderabbitai review

src/hgvs/variant.rs
```diff
@@ -584,6 +648,87 @@ impl HgvsVariant {
...
+    /// Format just the position+edit portion (without accession and coordinate prefix).
+    ///
+    /// For `NM_000088.3:c.459A>G`, this writes `459A>G`. Used by `AlleleVariant::Display`
+    /// to emit the spec-correct compact form (`ACC:c.[edit1;edit2]`).
+    ///
+    /// Variant types that have no simple loc/edit form (`Allele`, `RnaFusion`,
+    /// `NullAllele`, `UnknownAllele`) fall back to their full `Display` output. Callers
+    /// guard against this via `all_share_accession_and_type`, which excludes those
+    /// types from the compact branch.
+    pub(crate) fn fmt_loc_edit(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
nh13: **suggestion:** 

Could make `fmt_loc_edit` private throughout

```suggestion
    fn fmt_loc_edit(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```

nh13 (approved):
Two suggestions, and if you agree, no need for me to re-review:
1. make `fmt_loc_edit` private
2. add the two tests as per coderabbitai's nitpicks
    1. Mixed coordinate types staying expanded
    2. 2+-variant unknown-phase compact form

src/hgvs/variant.rs
msto: Done in fb0d659 — all callers live in this module, so dropped `pub(crate)` from each `fmt_loc_edit` (`HgvsVariant` + the seven typed structs).

PR #50, opened 2026-04-27
chore(python): add uv lockfile for reproducible dev environment

### LINKED ISSUE #47: Add `uv` for reproducible Python dev environment locking
## Summary

Add a `uv.lock` file to lock the Python development/test dependencies for the ferro-hgvs bindings, providing reproducible environments across developer machines and CI.

## Motivation

Currently, Python dev dependencies (`pytest`, `mypy`, `ruff`) are declared in `pyproject.toml` under `[project.optional-dependencies].dev` but are not locked. This means different contributors and CI runs may resolve different versions, leading to inconsistent lint/test behavior.

The existing `pixi.lock` covers the benchmarking environment (mutalyzer, biocommons) and should remain separate from the lighter-weight bindings dev workflow.

## Proposed Changes

1. Add `uv.lock` generated via `uv lock` from `pyproject.toml`
2. Update developer docs (README/CONTRIBUTING) to recommend:
   ```bash
   uv sync --extra dev
   uv run maturin develop --features python
   uv run pytest
   ```
3. Update CI workflow(s) to use `uv sync --extra dev` instead of bare `pip install`
4. Add `uv.lock` to version control (equivalent to committing `Cargo.lock`)

## Considerations

- `uv` is fast and supports `pyproject.toml` natively — no additional config files needed
- The `pixi.toml`/`pixi.lock` benchmarking environment remains unchanged
- Minimum Python version (`>=3.8` in pyproject.toml) should be validated against uv's resolver

msto:
## Summary

Adds `uv` lockfile support for the Python dev environment, building on the existing dependency declarations in `pyproject.toml`.

Closes #47

## Changes

- **`uv.lock`** — Generated lockfile pinning Python dev dependencies (pytest, ruff, mypy, maturin + transitives)
- **`pyproject.toml`** — Added `maturin` to `[project.optional-dependencies].dev` for completeness
- **`.github/workflows/ci.yml`** — Switched to `astral-sh/setup-uv` + `uv sync --extra dev --frozen` in `python-lint` and `test` jobs
- **`CONTRIBUTING.md`** — Added a note about `uv` and the Python bindings workflow

## Context

The Python dev deps were already declared in `pyproject.toml` — this adds the lock step so we get fully reproducible environments across CI and local dev, consistent with the Cargo.lock + pixi.lock pattern used elsewhere in the project.

This is the [recommended approach](https://docs.astral.sh/uv/guides/projects/#uvlock) for Python projects using maturin-based Rust bindings: `Cargo.lock` pins the Rust dependency tree, `uv.lock` pins the Python dependency tree, and both are committed to version control.

## Developer workflow

```bash
uv sync --extra dev
uv run maturin develop --features python
uv run pytest
```

## Notes

- CI uses `--frozen` so it'll flag if `uv.lock` drifts from `pyproject.toml`
- Pixi benchmarking environment is untouched
- `setup-uv` action pinned by SHA following existing conventions
- Happy to adjust anything here — let me know if you'd prefer a different approach!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,17 +33,18 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+      - uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b  # v8.1.0
       - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
         with:
           python-version: "3.12"
-      - name: Install linters
-        run: pip install ruff mypy
+      - name: Install dev dependencies
+        run: uv sync --extra dev --frozen
       - name: Run ruff check
-        run: ruff check python/ tests/python/
+        run: uv run ruff check python/ tests/python/
       - name: Run ruff format
-        run: ruff format --check python/ tests/python/
+        run: uv run ruff format --check python/ tests/python/
       - name: Run mypy
-        run: mypy python/ferro_hgvs/
+        run: uv run mypy python/ferro_hgvs/
 
   clippy:
     name: Clippy
@@ -82,6 +83,7 @@ jobs:
         with:
           toolchain: 1.94.0
       - uses: taiki-e/install-action@8e83571c9a797c9e42a58141f098806bfec453a0  # nextest
+      - uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b  # v8.1.0
       - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
         with:
           python-version: "3.12"
@@ -99,16 +101,12 @@ jobs:
             ${{ runner.os }}-cargo-
       - name: Run Rust tests
         run: cargo nextest run --features dev
-      - name: Install maturin and pytest
-        run: pip install maturin pytest
-      - name: Build Python wheel
-        run: |
-          rm -f target/wheels/*.whl
-          maturin build --features python
-      - name: Install wheel and run Python tests
-        run: |
-          pip install target/wheels/*.whl
-          pytest tests/python/ -v
+      - name: Install dev dependencies
+        run: uv sync --extra dev --frozen
+      - name: Build and install Python wheel
+        run: uv run maturin develop --features python
+      - name: Run Python tests
+        run: uv run pytest tests/python/ -v
 
   build:
     name: Build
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai review

.github/workflows/ci.yml
```diff
@@ -99,16 +101,12 @@ jobs:
...
-          rm -f target/wheels/*.whl
-          maturin build --features python
-      - name: Install wheel and run Python tests
-        run: |
-          pip install target/wheels/*.whl
-          pytest tests/python/ -v
+      - name: Install dev dependencies
+        run: uv sync --extra dev --frozen
+      - name: Build and install Python wheel
+        run: uv run maturin develop --features python
```
nh13, line 107: **issue:** the old flow built a wheel, installed it, and ran tests against the wheel, which is more what users (versus developers) would expect.  develop is faster, but not what users would do

.github/workflows/ci.yml
```diff
@@ -82,6 +83,7 @@ jobs:
         with:
           toolchain: 1.94.0
       - uses: taiki-e/install-action@8e83571c9a797c9e42a58141f098806bfec453a0  # nextest
+      - uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b  # v8.1.0
```
nh13, line 86: **question:** can this action setup python too? can we also set the uv cache to true too?

CONTRIBUTING.md
```diff
@@ -18,6 +19,14 @@ cargo build
 cargo test --features dev
 ```
 
+### Python Bindings Setup
+
+```bash
+uv sync --extra dev
+uv run maturin develop --features python
+uv run pytest
+```
```
nh13, line 28: **suggestion:** Should we document somewhere: "After modifying `pyproject.toml` Python deps, run `uv lock` and commit `uv.lock`; CI uses `--frozen` and will fail otherwise"

nh13 (changes requested):
While we're here, should we add `.venv/` to `.gitignore` too?

.github/workflows/ci.yml
msto: Good catch — you're right that `develop` doesn't exercise the wheel users actually install, and switching to it loses the packaging coverage the old flow had.

Thinking through it, I'd frame CI as proving two distinct things:

1. **Code correctness** — every PR, fast feedback. `maturin develop` + pytest fits here, and it matches what contributors run locally via `poe check-tests` in #53.
2. **Wheel/packaging correctness** — the artifact users `pip install` works. Catches missing files in `pyproject.toml`, type-stub packaging, import-time issues that editable installs hide.

#58's `smoke-test` matrix covers (2) at release time, but not on PRs — so packaging bugs only surface when we cut a release, which is too late. The clean fix is a dedicated `python-wheel-test` PR job (build, install into a clean venv, run pytest from outside the source tree).

Rather than fold that into this PR, I'd like to spin it out — it's separable from the lockfile/dev-env theme here, has design choices worth their own review (single-platform vs matrix, every-PR vs label-gated), and can land independently against `main`. Tracking in #62.

nh13 (approved):

Issue #51, opened 2026-04-27
Migrate Python dev dependencies to `[dependency-groups]` and adopt template conventions


msto:
## Summary

Align the Python bindings dev setup with our [python-template](https://github.com/fulcrumgenomics/python-template) conventions where applicable.

### 1. Use `[dependency-groups]` instead of `[project.optional-dependencies]`

Dev tools are not user-facing extras — they should not appear in published PyPI metadata.

```toml
[dependency-groups]
dev = ["maturin>=1.0,<2.0", "pytest~=8.4", "mypy~=1.18", "ruff==0.14.4", "poethepoet~=0.36"]
```

CI becomes `uv sync --group dev --locked`.

### 2. Add `poethepoet` task runner

```toml
[tool.poe.tasks]
build        = "maturin develop --features python"

check-format = "ruff format --check --diff python/ tests/python/"
check-lint   = "ruff check python/ tests/python/"
check-typing = "mypy"
check-tests.sequence = ["build", {cmd = "pytest"}]

check-all.ignore_fail = "return_non_zero"
check-all.sequence    = ["check-format", "check-lint", "check-typing", "check-tests"]

fix-format = "ruff format python/ tests/python/"
fix-lint   = "ruff check --fix python/ tests/python/"

fix-all.ignore_fail = "return_non_zero"
fix-all.sequence    = ["fix-format", "fix-lint"]

fix-and-check-all.ignore_fail = "return_non_zero"
fix-and-check-all.sequence    = ["fix-format", "fix-lint", "check-typing", "check-tests"]
```

`check-tests` runs `maturin develop` before pytest, ensuring the native extension is always current. The build is a no-op when source has not changed.

### 3. Add mypy settings beyond `strict`

Two settings the template uses that `strict = true` does not include:

```toml
[tool.mypy]
strict = true
warn_unreachable = true
enable_error_code = [
    "ignore-without-code",
    "possibly-undefined",
    "exhaustive-match",
]
```

### 4. Use `--locked` in CI

Switch from `--frozen` to `--locked` — this verifies the lockfile is in sync with `pyproject.toml` before installing, rather than just trusting whatever lockfile exists.

## Out of Scope

- Expanding ruff lint rules (bindings layer is thin enough that current rules suffice)
- `hatchling` build backend (we use maturin)
- `git-cliff` changelog (we use release-plz)
- Python version matrix testing (bindings support 3.8+, template targets 3.12+)
[...]

PR #53, opened 2026-04-27
feat(python): add poethepoet task runner for Python dev workflow

msto:
## Summary

Adds `poethepoet` as a task runner for the Python bindings dev workflow, consistent with our [python-template](https://github.com/fulcrumgenomics/python-template).

Builds on #52. Related to #51.

## Tasks Defined

| Command | What it does |
|---------|-------------|
| `poe build` | `maturin develop --features python` |
| `poe check-format` | `ruff format --check --diff` |
| `poe check-lint` | `ruff check` |
| `poe check-typing` | `mypy` |
| `poe check-tests` | build then pytest (ensures native extension is current) |
| `poe check-all` | all checks |
| `poe fix-format` | `ruff format` |
| `poe fix-lint` | `ruff check --fix` |
| `poe fix-all` | fix format + lint |
| `poe fix-and-check-all` | fix format + lint, then typing + tests |

## Design Decisions

- `check-tests` runs `maturin develop` as its first step. The build is a no-op when source has not changed, so overhead is minimal, and it prevents running pytest against a stale or missing native extension.
- `poethepoet` requires Python >= 3.10, so the dependency is gated with an environment marker. This goes away once we drop 3.8/3.9 support.

## Usage

```bash
uv run poe check-all
uv run poe fix-and-check-all
```


nh13 (approved):
Thank-you for doing this, LGTM.  Shall we also switch CI to use poe as a follow on issue?

pyproject.toml
```diff
@@ -70,6 +71,26 @@ ignore = [
 [tool.ruff.lint.isort]
 known-first-party = ["ferro_hgvs"]
 
+[tool.poe.tasks]
+build        = "maturin develop --features python"
+check-format = "ruff format --check --diff python/ tests/python/"
+check-lint   = "ruff check python/ tests/python/"
+check-typing = "mypy"
```
nh13: **nitpick:** could make `poe --help` look nicer
```suggestion
check-typing = {cmd = "mypy", help = "type-check Python sources"}
```

msto: Done in 3ff43f5 — added `help` text to every task. `poe --help` now lists each one with a short description.

msto:
Filed the CI follow-on as #60.

PR #63, opened 2026-04-27
ci: add python-wheel-test PR job to exercise the shipped wheel

### LINKED ISSUE #62: Add `python-wheel-test` PR-CI job to catch packaging bugs
## Motivation

The PR-CI `test` job uses `maturin develop`, which builds an editable
install of the native extension. This is fast and right for code-correctness
feedback, but does not exercise the wheel that users actually `pip install`.
Packaging bugs — missing files in `pyproject.toml`, type-stub packaging,
import-time issues that don't manifest in editable installs — would only
surface at release time, via the cross-platform `smoke-test` matrix in
`release-wheels.yml` (#58).

We want a PR-level signal that the shipped artifact works, on a single
platform, before a release is cut.

## Proposed shape

A new `python-wheel-test` job in `ci.yml` that:

1. `uv run maturin build --release --features python`
2. Installs the wheel into a clean venv (no source tree on `sys.path`)
3. Runs `pytest tests/python/` from outside the source tree, so imports
   resolve to the installed wheel rather than `python/ferro_hgvs/`

Single platform (Linux, py3.12) — the matrix is covered at release time.

## Python provisioning

This job should use **`actions/setup-python`** rather than `setup-uv` with
`python-version`, even though the rest of the workflow is uv-native.

The whole point of `python-wheel-test` is to simulate the environment users
will have when they `pip install ferro-hgvs`. Most users will be on canonical
CPython (from python.org, distro packages, or `actions/setup-python`), not
uv-managed [python-build-standalone](https://github.com/astral-sh/python-build-standalone)
builds. The two are ~99% interchangeable for PyO3 extensions, but the 1% of
edge cases (ABI/openssl/locale differences) is exactly what this job exists
to catch — so it should test against the Python users actually have.

Other jobs (`test`, `python-lint`) stay on `setup-uv` with `python-version`
since their purpose is testing *our code*, not simulating user environments.

## Open questions

- Run on every PR vs. only when Python files / `pyproject.toml` change?
- Test the sdist too (`maturin sdist` → `pip install dist/*.tar.gz`)?
- Any way to share the maturin build cache with the existing `test` job?

## Out of scope

- Matrix coverage across OS/Python versions — handled by
  `release-wheels.yml` on `release: published`.
- Restructuring the existing `test` job — `develop` is the right call there.

## Context

Raised by @nh13 in #50 review (the old test job built a wheel and tested
against it; switching to `develop` was a regression in packaging coverage).
Punting the dedicated PR-CI job to its own change so the lockfile/dev-env
work in #50–#58 can land cleanly.

msto:
## Summary

Closes #62

Adds a `python-wheel-test` CI job that builds a release wheel, installs
it into a clean venv, and runs the test suite against the installed
package — exercising the artifact users get from `pip install`, not the
in-tree source. Catches packaging bugs (missing files in
`pyproject.toml`, type-stub packaging, import-time issues editable
installs hide) at PR time, before they slip into a release.

Today's `test` job also tests via wheel, so there's transitional
overlap. After #50 lands, that job switches to `maturin develop` for
fast dev iteration — at which point this is the sole PR-CI check that
exercises the wheel. (`release-wheels.yml`'s `smoke-test` only runs on
`release: published`.)

Uses `actions/setup-python` (canonical CPython) rather than `setup-uv`,
since simulating the user's environment is the whole point. Tests run
from `/tmp` with `--rootdir=/tmp --import-mode=importlib` so imports
resolve to the installed wheel, not source.

## Test plan

- Wheel build step produces a single `.whl` in `dist/`
- Verify step confirms `import ferro_hgvs` resolves to the venv path
- Full pytest suite passes against the installed wheel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,6 +110,57 @@ jobs:
           pip install target/wheels/*.whl
           pytest tests/python/ -v
 
+  python-wheel-test:
+    name: Python Wheel Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          submodules: true
+      - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8  # stable
+        with:
+          toolchain: 1.94.0
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
+        with:
+          python-version: "3.12"
+      - uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae  # v5.0.5
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-wheel
+          restore-keys: |
+            ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-
+            ${{ runner.os }}-cargo-
+      - name: Build release wheel
+        run: |
+          python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

PR #70, opened 2026-04-30
fix: accept gene selectors on non-RefSeq accessions

### LINKED ISSUE #69: Gene selector parsing fails for non-RefSeq accessions
## Summary

`parse()` raises `ValueError` when a gene selector (parenthesized text after the accession) is present on a non-RefSeq-style accession. The same accessions parse successfully without a selector.

## Reproduction

```python
import ferro_hgvs  # 0.3.0

# RefSeq accessions with selectors — OK
ferro_hgvs.parse("NM_000088.3(COL1A1):c.100A>G")  # OK
ferro_hgvs.parse("NM_000088(1):c.100A>G")          # OK

# Non-RefSeq accessions with selectors — FAIL
ferro_hgvs.parse("MYSEQ(1):c.100A>G")              # ValueError
ferro_hgvs.parse("MY-SEQ(GENE1):c.100A>G")         # ValueError
ferro_hgvs.parse("MYREF_SEQ(1):c.100A>G")          # ValueError
ferro_hgvs.parse("MYREF_SEQ(1):p.(Arg8Gln)")       # ValueError

# Same accessions without selectors — OK
ferro_hgvs.parse("MYSEQ:c.100A>G")                 # OK
ferro_hgvs.parse("MY-SEQ:c.100A>G")                # OK
ferro_hgvs.parse("MYREF_SEQ:c.100A>G")             # OK
ferro_hgvs.parse("MYREF_SEQ:p.(Arg8Gln)")          # OK
```

Error message:

```
ValueError: Parse error: Parse error at position 0: Failed to parse accession:
Error(Error { input: "MYREF_SEQ(1):c.100A>G", code: Alpha })
```

## Expected behavior

Gene selectors should be accepted on any valid accession, not only those matching the RefSeq `XX_digits` pattern. Per the [HGVS nomenclature](https://hgvs-nomenclature.org/stable/background/refseq/#DNAg), a gene selector is optional metadata and should not constrain the accession format.

## Context

Tools like mutalyzer emit transcript selectors on custom/synthetic accessions (e.g. `CUSTOM_ACC(1):c.100A>G`). Being unable to parse these prevents round-tripping HGVS strings produced by other tools.

## Workaround

Stripping the selector before parsing:

```python
import re

def strip_gene_selector(hgvs: str) -> str:
    """Remove the gene selector from an HGVS string."""
    return re.sub(r"(?<=\w)\([^)]+\):", ":", hgvs, count=1)
```

msto:
Fixes #69

`parse()` raised `ValueError` on inputs like `MYREF_SEQ(1):c.100A>G` — any
gene/transcript selector after a non-RefSeq-shaped accession. The simple-
accession fallback in `parse_simple_accession` consumed everything up to the
type-prefix colon as the accession name, then rejected it because `(`/`)` are
not valid SAM refname characters.

Per the [HGVS nomenclature](https://hgvs-nomenclature.org/stable/background/refseq/#DNAg),
the selector is optional metadata and should not constrain the accession
format, so tools like mutalyzer that emit selectors on custom accessions
(`CUSTOM_ACC(1):c.100A>G`) now round-trip.

The fix terminates the simple accession at the first `(` before the colon.
SAM refnames cannot contain `(`, so it cleanly marks the boundary;
`parse_gene_symbol` (already invoked immediately after `parse_accession`) then
consumes the selector exactly as it does for RefSeq accessions today.

**Existing accession paths are unaffected.** RefSeq (`NM_…`), Ensembl, LRG,
UniProt, assembly (`GRCh38(chr1)`), and compound-ref (`NC_…(NM_…)`) parsing
all dispatch before the simple-accession fallback. Inputs without a `(` in
the accession portion hit the same code path as before.

## Test plan

- New unit tests in `accession.rs` cover `MYSEQ(1)`, `MY-SEQ(GENE1)`,
  `MYREF_SEQ(1)`, and the protein-form `MYREF_SEQ(1):p.(Arg8Gln)`; selector
  is captured into `gene_symbol`.
- New parametrized Python test mirrors the four issue repro cases.
- Full Rust parser suite (225 tests) and full Python suite (112 tests) green.

### DIFF
--- a/src/hgvs/parser/accession.rs
+++ b/src/hgvs/parser/accession.rs
@@ -123,7 +123,10 @@ fn is_sam_refname_char(c: char) -> bool {
 /// Uses look-ahead to correctly identify the HGVS separator (`:` followed by type prefix).
 ///
 /// SAM spec: Reference names may contain printable ASCII `[!-~]` except `\ , " ' ( ) [ ] { } < >`
-/// and cannot start with `*` or `=`.
+/// and cannot start with `*` or `=`. Because `(` is forbidden in SAM refnames, a `(` before
+/// the type-prefix colon must be the start of an optional gene/transcript selector
+/// (e.g. `MYSEQ(GENE1):c.…`); the accession ends there and `parse_gene_symbol` consumes
+/// the selector next.
 fn parse_simple_accession(input: &str) -> IResult<&str, Accession> {
     // Check first character: cannot be empty, and cannot start with * or =
     let first_char = input.chars().next().ok_or_else(|| {
@@ -187,8 +190,7 @@ fn parse_simple_accession(input: &str) -> IResult<&str, Accession> {
         nom::Err::Error(nom::error::Error::new(input, nom::error::ErrorKind::Tag))
     })?;
 
-    // SAM refnames forbid '(', so a '(' before the type-prefix colon must be
-    // the start of a gene/transcript selector consumed by parse_gene_symbol.
+    // See doc comment: '(' before the colon ends the accession (selector boundary).
     let accession_end = memchr(b'(', &input_bytes[..separator_pos]).unwrap_or(separator_pos);
 
     // Validate all characters in the accession name are SAM-compatible
--- a/src/hgvs/parser/accession.rs
+++ b/src/hgvs/parser/accession.rs
@@ -187,8 +187,12 @@ fn parse_simple_accession(input: &str) -> IResult<&str, Accession> {
         nom::Err::Error(nom::error::Error::new(input, nom::error::ErrorKind::Tag))
     })?;
 
+    // SAM refnames forbid '(', so a '(' before the type-prefix colon must be
+    // the start of a gene/transcript selector consumed by parse_gene_symbol.
+    let accession_end = memchr(b'(', &input_bytes[..separator_pos]).unwrap_or(separator_pos);
+
     // Validate all characters in the accession name are SAM-compatible
-    let accession_str = &input[..separator_pos];
+    let accession_str = &input[..accession_end];
     if accession_str.is_empty() || !accession_str.chars().all(is_sam_refname_char) {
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

src/hgvs/parser/accession.rs
```diff
@@ -187,8 +187,12 @@ fn parse_simple_accession(input: &str) -> IResult<&str, Accession> {
         nom::Err::Error(nom::error::Error::new(input, nom::error::ErrorKind::Tag))
     })?;
 
+    // SAM refnames forbid '(', so a '(' before the type-prefix colon must be
+    // the start of a gene/transcript selector consumed by parse_gene_symbol.
```
nh13: **nitpick:** add to the doc comment for the function?

tests/python/test_core.py
```diff
@@ -45,6 +45,21 @@ def test_parse_duplication(self) -> None:
...
+            ("MY-SEQ(GENE1):c.100A>G", "GENE1"),
+            ("MYREF_SEQ(1):c.100A>G", "1"),
+            ("MYREF_SEQ(1):p.(Arg8Gln)", "1"),
+        ],
+    )
+    def test_parse_accepts_gene_selector_on_non_refseq(self, hgvs: str, selector: str) -> None:
+        # Issue #69: gene selectors must parse on any valid accession AND be
+        # captured on the variant — not silently dropped.
+        variant = ferro_hgvs.parse(hgvs)
+        assert f'"gene_symbol":"{selector}"' in variant.to_json()
```
nh13: **nitpick:** 

why not `variant.to_json()["gene_symbol"] == "{selector}"` so we don't rely on to_json formatting?

nh13 (approved):
Minor nits, take it or leave it

Issue #75, opened 2026-04-30
Normalize same-base single-position delins as identity (=)


msto:
The HGVS spec treats a delins whose inserted sequence equals the reference as no change, which should be expressed using identity notation (`=`).

`g.1000delinsG` (where ref = G) → `g.1000=`

`normalize()` currently preserves the delins form. The same-base 1→1 case is the simplest instance, but the rule applies more broadly: any delins where the inserted sequence matches the deleted reference is identity (e.g. `g.1000_1002delinsGCA` where ref[1000..1002] = `GCA`).

Follow-up to #73, which handled the substitution priority for the 1→1 ref≠alt case but left the same-base case alone.

Spec: https://hgvs-nomenclature.org/stable/recommendations/DNA/delins/

PR #77, opened 2026-04-30
fix: rewrite single-base delins as substitution per HGVS priority

### LINKED ISSUE #73: Apply HGVS edit-type priority to simplify single-base delins
The HGVS spec defines an edit-type priority order — substitution > deletion > inversion > duplication > insertion. A single-base delins that replaces one nucleotide should be expressed as a substitution.

`g.1000delinsA` (where ref = G) → `g.1000G>A`

`normalize()` currently preserves the delins form. Applying the priority rule would bring single-position output into spec compliance.

Spec: https://hgvs-nomenclature.org/stable/recommendations/DNA/delins/ (Prioritization)

msto:
## Summary

Fixes #73

The HGVS spec defines an edit-type priority — substitution > deletion > inversion > duplication > insertion — and a 1→1 delins must be expressed as a substitution. `normalize()` previously preserved the delins form, so `g.1000delinsA` (ref=G) round-tripped instead of producing `g.1000G>A`.

`normalize_na_edit` now checks for the substitution case in the `Delins` arm before the existing duplication check, and emits `NaEdit::Substitution` at the original single-position interval.

**Only the 1→1, ref≠alt case is converted.** Other delins shapes are unchanged:

| Input (ref shown in `[]`) | Before | After |
|---|---|---|
| `c.10delinsT` (ref=`[G]`) | `c.10delinsT` | `c.10G>T` |
| `c.10delinsG` (ref=`[G]`) | unchanged | unchanged (#75) |
| `c.10delinsAT` (ref=`[G]`) | unchanged | unchanged |
| `c.10_11delinsA` (ref=`[GT]`) | unchanged | unchanged |
| `c.5delinsGG` (ref=`[G]`, dup pattern) | `c.5dup` | unchanged |

The same-base identity case (`c.10delinsG` → `c.10=`) is tracked separately in #75.

## Test plan

- New unit test for `delins_is_substitution` covers the substitution case, same-base, multi-base delete, multi-base insert, empty insert, and out-of-bounds.
- New integration tests verify `c.10delinsT` → `c.10G>T`, `c.10delinsG` round-trips as delins, and `c.10_11delinsT` (2→1) is preserved as delins.
- Full `normalize` suite (129 tests) and clippy/fmt are clean.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -1406,14 +1406,31 @@ impl<P: ReferenceProvider> Normalizer<P> {
             NaEdit::Delins { sequence } => {
                 use crate::hgvs::edit::InsertedSequence;
 
-                // HGVS spec: delins should NOT be 3' shifted like del/dup/ins
-                // But we should check if delins should become a duplication first
-                // Example: c.5delinsGG where position 5 is G → dup
+                // HGVS spec: delins should NOT be 3' shifted like del/dup/ins,
+                // but the edit-type priority (sub > del > inv > dup > ins) means
+                // we may need to rewrite it as a higher-priority form.
                 if let InsertedSequence::Literal(seq) = sequence {
                     let seq_bytes: Vec<u8> = seq.bases().iter().map(|b| *b as u8).collect();
                     let start_idx = hgvs_pos_to_index(start);
                     let end_idx = end as usize;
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

PR #78, opened 2026-04-30
fix: rewrite delins as identity when insert matches reference

### LINKED ISSUE #75: Normalize same-base single-position delins as identity (=)

msto:
## Summary

Fixes #75

Per the HGVS spec, a delins whose inserted sequence equals the deleted reference produces no change and must be expressed using identity notation (`=`). `normalize()` previously preserved the delins form, so `g.1000delinsG` (ref=G) round-tripped instead of producing `g.1000=`. The rule applies for any matching length, not just single bases.

`normalize_na_edit` now checks for the identity case in the `Delins` arm before the existing duplication check, and emits `NaEdit::position_identity()` preserving the original interval.

**Only the insert==ref case is converted.** Other delins shapes are unchanged:

| Input (ref shown in `[]`) | Before | After |
|---|---|---|
| `c.10delinsG` (ref=`[G]`) | `c.10delinsG` | `c.10=` |
| `c.1_3delinsATG` (ref=`[ATG]`) | `c.1_3delinsATG` | `c.1_3=` |
| `c.1_3delinsACG` (ref=`[ATG]`) | unchanged | unchanged |
| `c.10delinsAT` (ref=`[G]`) | unchanged | unchanged |
| `c.5delinsCC` (ref=`[C]`, dup pattern) | `c.5dup`-equivalent | `c.5dup`-equivalent |

Companion to #73 (substitution priority for the 1→1 ref≠alt case). The two PRs touch the same arm; merge order doesn't affect logic since the rules are length-disjoint, but the second to merge will need a one-line rebase to slot its check in.

## Test plan

- New unit test for `delins_is_identity` covers single-base same, multi-base same, full-region match, single/multi-base differ, both length mismatches, empty insert, OOB, and inverted range.
- New integration tests verify `c.10delinsG` → `c.10=`, `c.1_3delinsATG` → `c.1_3=`, `c.1_3delinsACG` (different bases) preserved as delins, and `c.5delinsCC` (dup pattern) still normalizes to dup.
- Full `normalize` suite (130 tests) and clippy/fmt are clean.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -1408,14 +1408,21 @@ impl<P: ReferenceProvider> Normalizer<P> {
 
                 // HGVS spec: delins should NOT be 3' shifted like del/dup/ins,
                 // but the edit-type priority (sub > del > inv > dup > ins) means
-                // we may need to rewrite it as a higher-priority form.
+                // we may need to rewrite it as a higher-priority form: identity
+                // (insert == ref), substitution (1->1 ref!=alt), or duplication.
                 if let InsertedSequence::Literal(seq) = sequence {
                     let seq_bytes: Vec<u8> = seq.bases().iter().map(|b| *b as u8).collect();
                     let start_idx = hgvs_pos_to_index(start);
                     let end_idx = end as usize;
 
-                    // Single-base delins with a different alt base is a substitution
-                    // (e.g., g.1000delinsA where ref[1000]=G → g.1000G>A).
+                    // Identity (highest priority): insert == deleted reference.
+                    // Example: c.10delinsG where ref[10]=G → c.10=.
+                    if rules::delins_is_identity(ref_seq, start_idx, end_idx, &seq_bytes) {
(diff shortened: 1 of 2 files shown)


msto:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #84, opened 2026-04-30
tests: regression test pinning normalize() output against HGVS v21.0 spec examples


msto:
Add a regression test that consumes every variant string from the HGVS v21.0
spec ([hgvs-nomenclature.org/stable](https://hgvs-nomenclature.org/stable/)),
runs each through `Normalizer::normalize`, and asserts on ferro's *current*
output. Each row also carries the spec-correct form, so any future
normalization change is visible in the diff and reviewers can verify the new
output against the spec.

Companion to #83. The test is the mechanism for tracking progress against
that issue's checklist — items close as cases flip from
`current ≠ spec_expected` to `current == spec_expected`.

## Scope

1. **Fixture:** `tests/fixtures/grammar/hgvs_spec_normalization.json` —
   ~566 cases extracted from the v21.0 spec, grouped exactly as in #83's
   checklist (compliant, pair, reject, needs-reference). Each case carries
   `input`, `spec_expected`, `current`, `status`, and an optional
   `todo` pointing at the master issue.
2. **Test:** `tests/hgvs_spec_normalization_tests.rs` — reads the fixture,
   asserts `normalize(input) == current` for each row, panics with a diff
   summary on drift. Skips the `needs-reference` group (those rows are
   blocked on #82 / `Normalizer::from_manifest`).
3. **Generator:** committed alongside the fixture (e.g.
   `scripts/extract_spec_examples.py`), so the fixture can be
   regenerated when a new spec version ships. The generator scrapes the
   spec site (or the vendored submodule recommended in #83) and writes
   the JSON deterministically.

## Acceptance criteria

- [ ] `tests/fixtures/grammar/hgvs_spec_normalization.json` committed,
      covering every example from the v21.0 spec.
- [ ] `cargo nextest run --test hgvs_spec_normalization_tests` passes
      on `main` (the test pins *current* behavior — green on day one).
- [ ] Each row whose `current` differs from `spec_expected` carries
      a `todo` field referencing #83.
- [ ] Generator script committed; regenerating the fixture against an
      unchanged ferro produces a byte-identical file.
- [ ] Short note in `CONTRIBUTING.md` (or the master issue): if your
      PR changes any normalization output, update the fixture's
      `current` field and verify the new output against the spec.

## Out of scope

- Reference-aware rows (§2.1 3' rule, protein shifts, coord-system
  conversion). These need #82's `from_manifest` loader to drive
[...]

Issue #83, opened 2026-04-30
tracking: parse + normalize against v21.0 examples


msto:
Audit (April 2026) of `ferro_hgvs.normalize()` against every spec example extracted from <https://hgvs-nomenclature.org/stable/> (HGVS v21.0). This is the umbrella tracker — each checklist row should be split into its own issue as it's picked up.

Spec source: <https://github.com/HGVSnomenclature/hgvs-nomenclature> (currently vendored at `assets/hgvs-nomenclature/`, not yet pinned as a submodule — separate infra task).

In flight: #72, #73 (PR #77), #74 (PR #76), #75 (PR #78).

## A. Variant-form gaps (`parse` / `normalize`)

These are concrete spec input/output rows; each is an issue's worth
of work.

### A.1 Spec-compliant input that `normalize()` currently rewrites (§1)

Inputs the spec lists as the recommended form. Today's output differs:

- [ ] **Outer parentheses on predicted variants are preserved.** Spec: *"`( )` (parentheses) are used to indicate uncertainties ... `p.(Cys123Gly)`, indicating the amino acid change is predicted from DNA level data (no experimental proof)"* (`uncertain.md:9-11`); the same outer-paren shape applies to predicted DNA and RNA variants.
  - ✓ `c.(76A>G)`, `r.(1388g>a)` (outer parens flag the whole change as predicted)
  - ✗ `c.76(A>G)` (parens migrated inward — would mean the *edit* is uncertain), `r.1388g>a` (parens dropped — asserts an observation, not a prediction)
  - Currently: parens migrate inward at **parse** time, not display — `c.(76A>G)` is rewritten to `c.76(A>G)` in the predicted-variant arm at `src/hgvs/parser/variant.rs:1083-1104` (via `LocEdit::new_predicted`); the RNA dispatch at `src/hgvs/parser/variant.rs:1922-1924` drops the outer parens entirely. `normalize()` is idempotent on whatever shape parse produces (audit row B14). Adjacent to #81 D8.
  - Plan: in the predicted-variant parse arm, retain the outer-paren shape on the resulting `LocEdit` and emit it from `Display` whenever the predicted flag is set, instead of pushing the parens around the edit. The RNA path needs the same wrap-with-predicted-flag treatment so `r.(1388g>a)` round-trips with parens intact.

[...]

Issue #82, opened 2026-04-30
python: load real reference data via Normalizer.from_manifest and extended from_json


msto:
The Python `Normalizer` constructor today only takes `reference_json`, which
goes through `MockProvider::from_json` — a transcripts-only loader.
`from_json` ignores any protein sequences in the input, and there is no path
to consume the output of `ferro prepare`. This blocks Python clients from
exercising any reference-aware normalization (3' rule shifting, protein
position shifting, coord-system conversion).

### (a) Extend `MockProvider::from_json` to accept proteins and genomic sequences

Today's accepted shape is a bare list of `Transcript` records. Extend it to
also accept an object form with optional `proteins` and `genomic_sequences`
maps:

```json
{
  "transcripts": [ ... ],
  "proteins":  { "NP_003997.1": "MQRS..." },
  "genomic_sequences": { "NC_000023.11": "..." }
}
```

The bare-array form must continue to parse for backward compatibility.

Files: `src/reference/mock.rs::from_json`, `tests/fixtures/SCHEMAS.md`.

### (b) New `Normalizer.from_manifest(path)` Python classmethod

Wrap `MultiFastaProvider::from_manifest` so a directory written by
`ferro prepare` can drive normalization from Python without conversion.

```python
norm = ferro_hgvs.Normalizer.from_manifest("ferro-reference/manifest.json")
norm.normalize("NM_004006.2:c.3922del")  # applies 3' rule
```

Apply the same pattern to `EquivalenceChecker`, `BatchProcessor`,
`EffectPredictor`, `CoordinateMapper` — each currently takes
`reference_json` and would benefit identically.

Files: `src/python.rs` (lines 378, 880, 1473, 2556, plus the others taking
`reference_json`), `python/ferro_hgvs/__init__.pyi`.

## Acceptance criteria

- `Normalizer(reference_json="…json")` continues to accept the bare-array
  form (no breaking change).
- Object form with `proteins` / `genomic_sequences` round-trips.
- `Normalizer.from_manifest("…/manifest.json")` produces a normalizer
  backed by `MultiFastaProvider`.
- One new Python integration test per loader path, using
  `tests/fixtures/sequences/normalization_transcripts.json` (extended) and
  a tiny prepared-manifest fixture.

## Why now

Hard prerequisite for spec-compliance work that touches protein 3' rule
shifts and coord-system conversion (#83), and a blocker for downstream
Python projects that want to drive ferro from a live reference setup.

[...]

PR #85, opened 2026-04-30
test: HGVS v21.0 spec normalization regression fixture

msto:
## Summary

Part of #84.

Captures every variant string from the HGVS v21.0 spec paired with
ferro's current `normalize()` output, so future normalization changes
show up in the diff and can be verified against the spec. Companion to
the spec-compliance tracker (#83).

The fixture (`tests/fixtures/grammar/hgvs_spec_normalization.json`,
566 cases across 43 groups) tags each row as `preserved`,
`transformed`, `diverges`, `rejected`, `false-acceptance`,
`parse-error`, or `needs-reference`. §2.1 (3' rule) rows are tagged
`needs-reference` and skipped pending the Python `from_manifest`
loader (#82). Re-run `scripts/build_spec_normalization_fixture.py`
after any normalization change; output is byte-deterministic for
unchanged behavior.

The Rust test consumer and a live-spec scraper for the markdown
source are out of scope here and stay in #84.

## Test plan

- Counts match the audit in #83: 320 preserved / 60 diverges / 68
  parse-error in §1; 11 transformed / 32 unchanged / 8 diverges / 21
  parse-error / 9 needs-reference in §2; 32 rejected / 5
  false-acceptance in §3.
- Generator output is deterministic — re-running against unchanged
  ferro produces a byte-identical fixture.

### DIFF
--- a/scripts/build_spec_normalization_fixture.py
+++ b/scripts/build_spec_normalization_fixture.py
@@ -1,0 +1,286 @@
+"""Build the HGVS spec normalization regression fixture.
+
+Reads the curated extract at
+``tests/fixtures/grammar/hgvs_spec_examples_v21.md``, runs every variant
+string through ``ferro_hgvs.normalize()``, and writes a structured JSON
+fixture at ``tests/fixtures/grammar/hgvs_spec_normalization.json``.
+
+The output captures, for each spec example:
+
+- ``input`` — the raw token from the spec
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

msto:
@nh13 is this safe to close as superseded by #105 ?

PR #99, opened 2026-05-03
test: tighten coverage_gap_tests assertions; restore intronic-ins coverage (#94)

nh13:
Closes #94.

## Summary

Two changes to `tests/coverage_gap_tests.rs`:

1. **Tighten weak assertions.** Every gap test that previously asserted only `contains(...)` (e.g. `result.contains("del")`) or panic-canary `result.is_ok() || result.is_err()` is now locked with `assert_eq!` against the exact normalizer output captured under the current build.
2. **Restore three intronic-insertion gap tests** that PR #93 (#81 A7) removed on the rationale that the new ins-shift matrix subsumed them. Re-checking that rationale: the matrix's `SyntheticBuilder` places the CDS over the whole transcript and emits zero intronic cells, so the removed tests cover a coordinate region the matrix does not. They are restored under a new `intronic_insertions` module with strict `assert_eq!` assertions matching their counterpart pattern.

True boundary-spanning panic-canary tests (the four cases under `boundary_spanning::*` and `delins_boundary::test_boundary_spanning_delins`) are intentionally left as `is_ok() || is_err()` per the issue's stated scope: they are testing a different invariant (the path doesn't crash; either canonicalization or an explicit error is acceptable).

Builds on PR #95 (intronic-insertion restoration) by extending the same tightening pass to the remaining gap-test buckets, and adds cross-references and procedural detail to the restored intronic-insertion comments.

## Suspected bugs surfaced by tightening — and where they're tracked

Several locked outputs do not appear to match what the HGVS spec asks for. Per #94's design, they are locked anyway with `FIXME(#NN)` comments tied to follow-up tracking issues, so that the next reader knows the locked value is suspected-buggy not aspirational, and so that the test fails when the underlying bug is fixed. The cluster of symptoms is striking enough that the most likely explanation is a **single root cause**: the normalizer's reference-base lookup at minus-strand intronic positions returns the genomic-strand letters rather than the transcript-view letters. That fits all four observed misfires below; filed as #98.

| Symptom | Locked output | Expected | Tracking |
|---|---|---|---|
| `NM_MINUS.1:c.30+2_30+3insA` does not canonicalize to `dup` despite transcript-view AAA tract | `c.30+2_30+3insA` | `c.30+4dup` | #98, #81 A1 |
| `NM_MINUS.1:c.30+2dup` does not 3'-shift in transcript-view AAA tract | `c.30+2dup` | `c.30+4dup` | #98, #81 A6 |
[...]

Issue #97, opened 2026-05-03
normalize: minus-strand 5'UTR single-base del collapses to c.?del instead of resolved position


nh13:
Surfaced when tightening `tests/coverage_gap_tests.rs` per #94. Sub-item of #81 (J-bucket / A5).

## Repro

Synthetic minus-strand UTR transcript (`tests/coverage_gap_tests.rs::make_minus_strand_utr_transcript`). 5'UTR is tx 1..5 with transcript-view sequence `AAAAA` (RC of genomic `TTTTT`). c. coordinates: `c.-5 = tx 1, c.-4 = tx 2, c.-3 = tx 3, c.-2 = tx 4, c.-1 = tx 5`.

Input: `NM_MUTR.1:c.-3del` — single-base deletion at `c.-3` (within the 5'UTR `AAAAA` homopolymer).

Output today: `NM_MUTR.1:c.?del`

## Why this is wrong

`c.?` denotes an unknown position. Here the position is fully resolved on input (`c.-3`) and the local tract is unambiguous (5-A homopolymer); after 3'-shifting the canonical form should be `c.-1del` (rightmost A of the run, immediately upstream of ATG). Collapsing to `c.?` discards positional information that was supplied.

## Expected

`NM_MUTR.1:c.-1del` (or whatever `make_minus_strand_utr_transcript` resolves the 3'-most A in the 5'UTR run to). Either way, NOT `c.?`.

## Where this is locked

`tests/coverage_gap_tests.rs::utr_minus_strand::test_5prime_utr_deletion_minus_strand` locks `c.?del` as the expected value with a `FIXME` pointing at this issue. The test will fail once the bug is fixed.

Issue #96, opened 2026-05-03
normalize: minus-strand multi-base dup emits genomic-strand repeat unit (T[6] vs A[6])


nh13:
Surfaced when tightening `tests/coverage_gap_tests.rs` per #94. Sub-item of #81 (B-bucket / A6).

## Repro

Synthetic minus-strand transcript (`tests/coverage_gap_tests.rs::make_minus_strand_transcript`), intron 1 genomic = `GGGCCCTTTG`. In coding direction, transcript-view bases at `c.30+1..c.30+10` are `C, A, A, A, G, G, G, C, C, C` (RC of intron, right-to-left).

Input: `NM_MINUS.1:c.30+2_30+4dup` — duplicating the `AAA` run at `c.30+2..c.30+4`.

Output today: `NM_MINUS.1:c.30+2_30+4T[6]`

## Two issues with the output

1. **Wrong-strand repeat unit.** The unit `T` is the genomic-strand letter at PAD+76..78 (genomic = TTT). In c. notation the unit must be the transcript-view letter — should be `A`.
2. **First-unit position should be one base, not the original 3-base span.** For a 1-base unit, the canonical position locates one repeat unit. Even if the engine prefers a longer unit (e.g. `AAA[2]`), the position must match the chosen unit length.

The plus-strand counterpart at the analogous AAA tract in intron 1 (`c.30+2_30+4dup` against the plus-strand fixture, which has tract `AAA` at `c.30+8..c.30+10`) is not exercised today — would be a useful comparison cell to add when this is fixed.

## Expected (subject to spec-canonical-form check)

Either `c.30+?A[6]` (1-base unit, count 6) or `c.30+2_30+4AAA[2]` (3-base unit, count 2). Whichever the spec / our existing convention prefers — the key is: transcript-view letter, position consistent with unit length.

## Where this is locked

`tests/coverage_gap_tests.rs::intronic_duplications::test_minus_strand_intronic_multi_base_dup` locks the current (suspected-buggy) output as the expected value, with a `FIXME` pointing at this issue. The test will fail once the canonical form is corrected — that's the point.

nh13:
Closing — only one of the two concerns in this issue's body was a real bug.

## What was real

[...]

Issue #94, opened 2026-05-03
test: tighten weak coverage_gap_tests.rs assertions; restore intronic ins coverage removed in A7


nh13:
`tests/coverage_gap_tests.rs` exists to guard code paths that the synthetic-shift coverage matrices (#93 for ins; A5 for del; A6 for dup) cannot reach — intronic positions, exon-intron boundary spans, UTR notation on minus strand. By design, a gap test is a *canary* that the path doesn't crash and produces the expected variant kind.

Two related issues today:

1. **Most assertions are crash-canaries only.** Examples (sampled from the del bucket): `test_minus_strand_intronic_single_del` asserts `result.contains("del")`; `test_plus_strand_exon_intron_deletion` asserts `result.is_ok() || result.is_err()`. They catch panics and total absence-of-`del` regressions but don't lock the actual emitted position or sequence. With the synthetic shift matrices now landing one variant kind at a time, the exact expected output for each of these gap inputs is computable and should be locked.

2. **A7 (#93) removed the intronic ins gap tests** (`test_plus_strand_intronic_insertion`, `test_minus_strand_intronic_insertion`, `test_intronic_insertion_to_dup_plus_strand`) on the rationale that they overlapped with the new ins matrix. They didn't — the matrix has zero intronic cells (`SyntheticBuilder` places the CDS over the whole transcript, no intron offsets). This is a real coverage gap that should be restored at the same time.

## Scope

- For each existing gap test in `coverage_gap_tests.rs` whose assertion is `contains(...)` or `is_ok() || is_err()`, compute the exact expected output and switch to `assert_eq!`. Covers ins/del/dup/delins gap buckets.
- Re-add the three intronic ins tests A7 removed, with strict equality assertions.
- Audit by running each gap test, recording the actual output, and locking it as the expected value (then sanity-check by inspection).

## Out of scope

- The `boundary_spanning` panic-canary tests (`is_ok() || is_err()`) where the *correct* behavior is \"either result is acceptable, just don't crash\" — leave as-is, they're testing a different invariant. Audit each individually to decide.
- Adding new scenarios. This is a hardening pass, not a coverage expansion.

## Related

- #81 (umbrella spec-compliance gaps)
- #93 (A1, A7 — ins matrix; this issue addresses the intronic-ins removal in that PR)
- A5 (del matrix; this issue is a sibling cleanup)
- A6 (dup matrix; same pattern will likely apply)

PR #100, opened 2026-05-03
fix(normalize): minus-strand intronic ref-base orientation (#98)

### LINKED ISSUE #98: normalize: minus-strand intronic reference-base lookup returns genomic-strand letters (root cause behind multiple A-bucket gaps)
Surfaced when tightening `tests/coverage_gap_tests.rs` per #94. Sub-item of #81; likely a common root cause behind several A-bucket / strand-orientation symptoms surfaced by the locked tests in #94.

## Hypothesis

When `Normalizer` queries reference bases at minus-strand intronic positions (e.g. `c.30+2..c.30+4` on a minus-strand transcript), the lookup returns **genomic-strand** letters rather than the **transcript-view** (reverse-complemented) letters. As a result, every downstream rule that compares the input edit against a local reference window misfires:

- **A1 (ins → dup canonicalization, checked off in #93)** does not fire on minus-strand intronic positions. Plus-strand counterpart works correctly.
  - `NM_PLUS.1:c.30+2_30+3insA` (plus, AAA tract) → `c.30+3dup` ✓
  - `NM_MINUS.1:c.30+2_30+3insA` (minus, transcript-view AAA tract) → `c.30+2_30+3insA` ✗ (should be `c.30+4dup`)
- **3' shift for `del` / `dup`** (sibling A5 / A6) does not fire on minus-strand intronic positions.
  - `NM_MINUS.1:c.30+2dup` (transcript-view A in AAA tract) → `c.30+2dup` ✗ (should be `c.30+4dup`)
  - `NM_MINUS.1:c.30+5del` (transcript-view G in GGG tract) → `c.30+5del` ✗ (should be `c.30+7del`)
- **PR #78 delins-as-identity rewrite** does not fire on minus-strand intronic positions.
  - `NM_MINUS.1:c.30+2_30+4delinsAAA` (transcript-view ref `AAA`, insert `AAA` — identity) → `c.30+2_30+4delinsAAA` ✗ (should rewrite to identity)
- **Repeat-unit emission on the wrong strand** (#96): when the engine *does* succeed in emitting a repeat form for a minus-strand multi-base intronic dup, the unit is drawn from the genomic strand (`T[6]` instead of `A[6]`).

All of these are consistent with one underlying defect: the reference-base lookup path used by the rule machinery isn't strand-aware for intronic positions. (Exonic minus-strand cases work because the lookup goes through transcript-sequence indexing, which is naturally strand-correct; intronic positions fall back to a genomic-sequence path that doesn't apply the strand flip.)

## Suggested investigation order

1. Audit `Normalizer`'s reference-base accessor used during shuffle / rule evaluation on intronic offsets. Likely candidates in `src/normalize/shuffle.rs` and `src/reference/`.
2. Add a focused failing test that asserts the expected transcript-view letter for an intronic position on a minus-strand fixture, independent of any rule firing.
3. Fix the orientation flip at the lookup boundary; verify the four symptom classes above all start firing on minus-strand intronic input.

## Locked-test breadcrumbs

When this is fixed, expect the following `#94`-locked tests in `tests/coverage_gap_tests.rs` to fail and need updating:

- `position_ordering::test_minus_strand_intronic_insertion_position_order` (A1)
- `intronic_duplications::test_minus_strand_intronic_dup` (A6)
- `intronic_duplications::test_minus_strand_intronic_multi_base_dup` (A6 + #96)
- `intronic_deletions_minus::test_minus_strand_intronic_single_del` (A5)
- `intronic_deletions_minus::test_minus_strand_intronic_del_3prime_shift` (A5)
- `delins_boundary::test_intronic_delins_minus` (#78 identity rewrite)

## Related

- #81 (umbrella spec-compliance gaps; A1, A5, A6, J3)
- #94 (locking pass that surfaced these symptoms)
- #96 (wrong-strand repeat unit on minus-strand multi-base dup)
- #97 (`c.?del` collapse on minus-strand 5'UTR del — separate root cause, but also strand-orientation-flavored)
- PR #78 (delins-as-identity rewrite that doesn't fire on minus-strand intronic)
- PR #93 (A1 / A7 matrix that has zero intronic cells)

nh13:
Closes #98. Stacked on #99 (will retarget to `main` automatically when #99 merges).

## TL;DR

A single defect — minus-strand intronic ref-base reads returning **genomic-strand** letters where the rule machinery expects **transcript-view** letters — was responsible for *all* of the suspected-buggy lock points #94 surfaced on minus-strand intronic positions, including the wrong-strand-letter symptom in #96. Fixing it where the genomic window is fetched (`normalize_intronic_cds` and `normalize_intronic_tx`) resurfaces #81 A1 / A5 / A6 canonicalization on minus-strand intronic input, plus PR #78's delins-as-identity rewrite, and corrects the repeat-unit letter for minus-strand multi-base dup. All seven locked-buggy tests in `tests/coverage_gap_tests.rs` flip to their spec-correct expected values.

## Root cause

The exonic CDS path (`normalize_cds`, line 506-509) feeds `transcript.sequence` (transcript-view, naturally strand-correct) and the variant's transcript-view alt into `normalize_na_edit` together — same orientation, rule fires correctly. The intronic paths (`normalize_intronic_cds` line 1019-1021 and `normalize_intronic_tx` line 1138-1140) instead feed the **genomic-strand** window from `get_genomic_sequence` alongside the same transcript-view alt. On a minus-strand transcript the two are reverse complements of each other, so every rule that compares the alt against the local reference window — `ins → dup`, `del/dup` 3'-shift, delins-as-identity, repeat-unit-derived-from-ref — silently misfires. Issue #98 spelled out the hypothesis with the suggested investigation order; that order is what this PR follows.

## Fix

In both `normalize_intronic_cds` and `normalize_intronic_tx`, after fetching `genomic_seq` and computing the relative positions / shuffle boundaries:

1. If `transcript.strand == Strand::Minus`, reverse-complement the genomic window and flip the relative positions / boundaries via `flip_intronic_for_strand`. The window now matches the transcript-view orientation that `normalize_na_edit` expects from the alt.
2. Run `normalize_na_edit` against the flipped window.
3. Map the result positions back to the genomic frame via `unflip_intronic_positions` before the existing `genomic_to_cds_intronic` / `genomic_to_tx_with_intron` conversion.

The two helpers are private to `src/normalize/mod.rs`. No public API change; no behavior change on plus-strand transcripts. The `swapped` position swap for genomic-frame coding-order restoration is unchanged — that only handles position order, not sequence orientation, so it remains orthogonal.

`normalize_boundary_spanning_cds` (which uses the same genomic window pattern) is intentionally **not** fixed here — boundary-spanning is in the panic-canary scope of #94, and changing it requires a careful rethink of how the union-of-exon-and-intron boundaries flip across the splice site. Filed as a follow-up if needed.

## Tests

**Diagnostic suite** — `tests/issue_98_minus_strand_intronic_orientation.rs` (new). Four single-assertion tests, one per symptom class:

| Symptom | Input | Asserted |
|---|---|---|
| ins → dup canonicalization | `c.30+2_30+3insA` | `c.30+4dup` |
| dup 3'-shift | `c.30+2dup` | `c.30+4dup` |
| del 3'-shift (acceptor-relative) | `c.30+5del` | `c.31-4del` |
| delins identity rewrite | `c.30+2_30+4delinsAAA` | `c.30+2_30+4=` |

All four fail before the fix (orientation mismatch → no rule fires) and pass after. They depend only on the fixture and the public `normalize` API.

**Cascade in `tests/coverage_gap_tests.rs`** — every `FIXME(#98, ...)` lock point flipped to its spec-correct value, including the `#96` multi-base dup whose locked form moves from `T[6]` (genomic-strand letter) to `A[6]` (transcript-view letter, full-tract position, post-edit unit count — matching the HGVS DNA repeat spec's example shape):

| Test | Old buggy locked | New spec-correct |
|---|---|---|
| `position_ordering::test_minus_strand_intronic_insertion_position_order` | `c.30+2_30+3insA` | `c.30+4dup` |
| `intronic_duplications::test_minus_strand_intronic_dup` | `c.30+2dup` | `c.30+4dup` |
| `intronic_duplications::test_minus_strand_intronic_multi_base_dup` | `c.30+2_30+4T[6]` | `c.30+2_30+4A[6]` |
| `intronic_deletions_minus::test_minus_strand_intronic_single_del` | `c.30+2del` | `c.30+4del` |
| `intronic_deletions_minus::test_minus_strand_intronic_del_3prime_shift` | `c.30+5del` | `c.31-4del` |
| `delins_boundary::test_intronic_delins_minus` | `c.30+2_30+4delinsAAA` | `c.30+2_30+4=` |

The header docstring is updated to reflect that all `FIXME` markers are resolved.

- **#81 A1 (intronic, minus strand)** — now passes the no-intronic-cell gap that PR #93's matrix didn't cover. A1 may want to be re-marked as more-fully-closed in #81's checklist; left to the reviewer.
- **#81 A5 (del 3'-shift, minus-strand intronic)** — fires correctly; the broader A5 matrix is still its own item.
- **#81 A6 (dup 3'-shift, minus-strand intronic)** — same.
- **PR #78 (delins-as-identity rewrite)** — now fires on minus-strand intronic input.
- **#96 (multi-base dup repeat-unit)** — wrong-strand-letter symptom is fully fixed by this PR. The other concern in that issue's body — about position-vs-unit-length consistency — was based on a misreading of the HGVS DNA repeat spec; the spec uses full-tract position with the explicit unit and post-edit count uniformly (e.g., `g.101179660_101179695TG[14]`, `c.89_118AGC[13]`). #96 will be closed.
- **#97 (`c.?del` collapse on minus-strand 5'UTR del)** — untouched here; that path is exonic-UTR rather than intronic and needs a separate fix.

## Verification

- [x] `cargo nextest run --features dev` — full suite passes; seven previously-failing FIXME-tagged tests now pass with their updated locks; no unrelated regressions
- [x] `cargo nextest run --features dev --test issue_98_minus_strand_intronic_orientation` — 4/4 pass
- [x] Pre-commit hooks pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- *(normalize)* Minus-strand intronic normalization now reads the
+  reference window in transcript-view orientation. `normalize_intronic_cds`
+  and `normalize_intronic_tx` previously passed the genomic-strand bytes
+  fetched from `get_genomic_sequence` directly into `normalize_na_edit`
+  alongside the variant's transcript-view edit alt; on minus-strand
+  transcripts the two were mis-oriented, defeating every rule that
+  compared the alt against the local reference window. The fix
+  reverse-complements the genomic window and flips the relative
+  positions / shuffle boundaries on minus strand before normalization,
+  then maps the resulting positions back to the genomic frame for the
+  CDS / tx coordinate conversion. As a single root-cause fix this
+  resurfaces #81 A1 / A5 / A6 canonicalization, the PR #78 delins-as-
+  identity rewrite, and the transcript-view repeat-unit letter on
+  minus-strand intronic positions — all of which had been observed to
+  misfire in #94's locking pass. Issue #98.
 - Insertions that add ≥2 copies of a multi-base tandem repeat unit now
   emit repeat notation (`unit[N+k]`) instead of a duplication of the
   inserted sequence, per HGVS spec ("when more than one additional copies
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -38,6 +38,7 @@ use crate::hgvs::location::{CdsPos, GenomePos, TxPos};
 use crate::hgvs::parser::position::{OFFSET_UNKNOWN_NEGATIVE, OFFSET_UNKNOWN_POSITIVE};
 use crate::hgvs::variant::{CdsVariant, GenomeVariant, HgvsVariant, LocEdit, TxVariant};
 use crate::hgvs::HgvsVariant as HV;
+use crate::reference::transcript::Strand;
 use crate::reference::ReferenceProvider;
 use boundary::Boundaries;
 pub use config::{NormalizeConfig, ShuffleDirection};
@@ -1015,10 +1016,39 @@ impl<P: ReferenceProvider> Normalizer<P> {
         let intron_rel_end = intron_g_end.saturating_sub(seq_start) + 1;
         let boundaries = Boundaries::new(intron_rel_start, intron_rel_end);
 
-        // Perform normalization in genomic space
-        let seq_bytes = genomic_seq.as_bytes();
-        let (new_rel_start, new_rel_end, new_edit, warnings) =
-            self.normalize_na_edit(seq_bytes, edit, rel_start, rel_end, &boundaries)?;
+        // On minus-strand transcripts the genomic-strand sequence is the
+        // reverse complement of the transcript view, but the variant's
+        // edit alt is in transcript view. Running `normalize_na_edit` on
+        // the genomic-strand bytes therefore defeats every rule that
+        // compares the alt against the local reference window. Flip the
+        // sequence and the relative positions / boundaries to transcript
+        // view here, run normalization, then map the result positions
+        // back to the genomic frame. (Issue #98.)
+        let (work_seq, work_rel_start, work_rel_end, work_boundaries) = flip_intronic_for_strand(
+            transcript.strand,
+            &genomic_seq,
+            rel_start,
+            rel_end,
+            &boundaries,
+        );
+
+        // Perform normalization in transcript-view space
+        let seq_bytes = work_seq.as_bytes();
+        let (work_new_rel_start, work_new_rel_end, new_edit, warnings) = self.normalize_na_edit(
+            seq_bytes,
+            edit,
+            work_rel_start,
+            work_rel_end,
+            &work_boundaries,
+        )?;
+
+        // Map the result positions back to the genomic-strand frame
+        let (new_rel_start, new_rel_end) = unflip_intronic_positions(
+            transcript.strand,
+            work_seq.len() as u64,
+            work_new_rel_start,
+            work_new_rel_end,
+        );
 
         // Convert the normalized genomic position back to absolute genomic
         let new_g_start = seq_start + new_rel_start - 1;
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #81, opened 2026-04-30
Tracking: HGVS spec compliance gaps in normalization output


nh13:
This issue collates known and suspected gaps where ferro-hgvs accepts a parseable input but `normalize()` does not rewrite it into the form preferred by the HGVS spec. It serves as a roadmap for follow-up PRs; each item below is intended to become its own issue (or to land directly under this one) as work progresses.

### Progress

**Overall: 66 / 66 roadmap items merged (100%). Closing this tracker.** _Last updated: 2026-05-19._

| Section | Done | In review | Remaining | Total |
|---|---|---|---|---|
| A. Canonical-form rules | 11 | 0 | 0 | 11 |
| B. Repeated-sequence variants | 6 | 0 | 0 | 6 |
| C. Allele / phase rules | 6 | 0 | 0 | 6 |
| D. Protein (p.) | 8 | 0 | 0 | 8 |
| E. RNA (r.) | 4 | 0 | 0 | 4 |
| F. Mitochondrial (m.) | 3 | 0 | 0 | 3 |
| G. Uncertainty / unknowns | 3 | 0 | 0 | 3 |
| H. Conversion (`con`) / complex | 2 | 0 | 0 | 2 |
| I. Reference sequences and accessions | 6 | 0 | 0 | 6 |
| J. Coordinate-system boundaries | 3 | 0 | 0 | 3 |
| K. Cross-format conversion | 7 | 0 | 0 | 7 |
| L. Validation and error modes | 7 | 0 | 0 | 7 |
| **Total** | **66** | **0** | **0** | **66** |

### Activity since 2026-05-16

Roadmap-item PRs merged this window (in chronological order):

| Merged | PR | Roadmap item | Summary |
|---|---|---|---|
| 2026-05-17 | #217 | C2 remaining | Bracket-aware split for chimeric + non-compact mosaic forms (closes #216) |
| 2026-05-17 | #222 | C5 | 3+ variants in one allele bracket where merging does not apply (closes #221) |
| 2026-05-17 | #225 | D2 | Protein extension Display emits `extTerN` (closes #224) |
| 2026-05-17 | #227 | D1 / D3 / D4 / D8 | Protein canonicalization audit umbrella (closes #226) |
| 2026-05-17 | #232 | E1 | RNA lowercase enforcement coverage across edit types (closes #231) |
| 2026-05-17 | #234 | E3 | r./c. canonicalization parity across edit shapes (closes #233) |
[...]

PR #93, opened 2026-05-02
ins 3'-shift coverage matrix + tandem-repeat ins canonical-form fix (#81 A1, A7)

nh13:
Addresses #81 items A1 (`ins` → `dup` canonical form) and A7 (3'-shift coverage matrix for `ins`). The matrix scenarios (`single_base_ins_in_homopolymer_one_unit`, `multi_base_ins_in_tandem_one_unit`) lock the A1 rule with strict string equality across all 7 coord-system / strand combinations, so A1 is de-facto closed by this PR.

## Summary

Locks ferro's 3'-shift behavior for insertion variants across every nucleotide coord-system / strand combination, in a single audit-shaped matrix file. Surfaces and fixes two production bugs along the way.

### What's new

**Test infrastructure**
- New file `tests/ins_shift_matrix.rs`: 84 rstest cases organized into 7 modules (`genomic`, `cds_plus`, `cds_minus`, `noncoding_plus`, `noncoding_minus`, `rna_plus`, `rna_minus`), each exercising the same 8 shuffle scenarios.
- New shared helper module `tests/common/synthetic.rs`: `SyntheticBuilder` for constructing `MockProvider` fixtures across coord systems, reusable by future del / dup / repeat-notation matrices (sibling roadmap items A5, A6, B1, B2).

**Production-code fixes (surfaced by the matrix)**
- `MockProvider::get_sequence` now falls through to genomic contig lookup when the id is not a transcript, matching `FastaProvider` behavior. Without this, `add_genomic_sequence(...)` silently took the canonicalize-only fallback in `normalize_genome`, skipping 3'-shifting entirely.
- `insertion_to_repeat` extended to handle multi-base tandem units. An insertion adding ≥2 copies of a tandem unit already present adjacent to the insertion point now emits `unit[N+k]` repeat notation per HGVS spec, instead of `dup` of the inserted bases. Single-unit additions remain `dup`.

**Migration**
- 16 ins-shuffle tests removed from `tests/normalize_tests.rs` and `tests/coverage_gap_tests.rs` that overlap with matrix cells. Two real-accession smoke tests (`test_insertion_becomes_dup`, `test_insertion_becomes_repeat`) preserved as canonical-form sanity checks against real RefSeq data.

### Scenario matrix

| # | Scenario | Expected canonical form |
|---|---|---|
| 1 | No-shift (first-base mismatch) | unchanged `ins` |
| 2 | Single-base ins in homopolymer (1 unit added) | `dup` at 3' end |
[...]

PR #101, opened 2026-05-03
fix(normalize): repeat-notation position locates first unit, not full tract (#96)

### LINKED ISSUE #96: normalize: minus-strand multi-base dup emits genomic-strand repeat unit (T[6] vs A[6])

nh13:
Closes #96. Stacked on #100 (will retarget through the chain as the upstream PRs merge).

## TL;DR

Per HGVS spec for repeated sequences, the position in `prefix:position_or_range unit[N]` must locate the **first repeat unit** — its length must match the unit length. ferro previously emitted the location of the **entire reference tract** (`ref_start + ref_count * unit_len - 1`), producing forms like `c.342_362GCA[9]` for a 9-unit GCA tract (21-base position) where the spec requires `c.342_344GCA[9]` (3-base position matching the 3-base unit length). The defect was identical in three repeat-emission code paths in `src/normalize/rules.rs`. Fixing all three sets `end = ref_start + unit_len - 1`.

## What was wrong

Three independent emission paths, same bug shape:

| Function | Trigger | Old `end` | New `end` |
|---|---|---|---|
| `duplication_to_repeat::Homopolymer` | dup of a single-base homopolymer | `ref_start + ref_count - 1` (full tract) | `ref_start` (single-base unit, single-base position) |
| `duplication_to_repeat::TandemRepeat` | dup of a multi-base tandem unit | `ref_start + ref_count * unit_len - 1` (full tract) | `ref_start + unit_len - 1` (one unit) |
| `insertion_to_repeat` | ins adding ≥ 2 copies of a tandem unit | `ref_start + ref_count * unit_len - 1` (full tract) | `ref_start + unit_len - 1` (one unit) |
| `normalize_repeat::Repeat` (already-existing repeat input) | re-normalizing a `unit[N]` variant | `ref_end - 1` (exclusive end of full tract → last position) | `ref_start + unit_len - 1` (one unit) |

The fourth path (`normalize_repeat::Repeat`) is what made this a *spec-compliance* bug rather than just a cosmetic one: until it was also fixed, the idempotency suite caught the inconsistency. `dupGCAGCA` against a real RefSeq tract normalized to `c.342_344GCA[9]` once (correct first-unit position from the dup→repeat path), then expanded to `c.342_362GCA[9]` on a second pass (incorrect full-tract position from the repeat→repeat path). The existing `tests/idempotency_tests.rs::test_normalization_idempotency` test caught this within the cascade verification.

## Cascade

The fix is mechanical for the test suite — every locked output that previously hard-coded the full-tract end position flips to the first-unit end position:

- **`tests/ins_shift_matrix.rs`** — 29 case lines across all 7 modules (`genomic`, `cds_plus`, `cds_minus`, `noncoding_plus`, `noncoding_minus`, `rna_plus`, `rna_minus`) updated. Each module's `multi_base_ins_in_tandem_multiple_units` block changes:
  - 1-base unit (`A[5]`, `A[7]`): output template drops the `_{1}` and the args drop the second element.
  - 2-base unit (`AC[5]`): args become `[start, start + 1]`.
  - 3-base unit (`GCA[5]`): args become `[start, start + 2]`.
- **`tests/normalize_tests.rs`** — three real-accession `test_insertion_becomes_repeat` cases (`NM_001127687.1`, `NM_000382.3`, `NM_015120.4`), one `test_dup_becomes_repeat` case (`NM_020732.3`), and its ignored sibling.
- **`src/normalize/rules.rs::tests`** — `test_insertion_to_repeat` and `test_duplication_to_tandem_repeat` update their `end`-position assertions.
- **`tests/coverage_gap_tests.rs`** — the residual `FIXME(#96)` lock on `intronic_duplications::test_minus_strand_intronic_multi_base_dup` flips from `c.30+2_30+4A[6]` to `c.30+2A[6]`. The file's header docstring is updated to reflect that no FIXMEs remain (the only one was this).

## Diagnostic tests

`tests/issue_96_repeat_unit_position_consistency.rs` (new) — four single-assertion tests, one per (path × unit-length) combination:

| Path | Unit | Asserted shape |
|---|---|---|
| dup → repeat | 1-base | `c.2A[6]` |
| dup → repeat | 3-base | `c.2_4GCA[5]` |
| ins → repeat | 1-base | `c.2A[5]` |
| ins → repeat | 3-base | `c.2_4GCA[5]` |

All four fail before the fix (each emits the full-tract position) and pass after. They depend only on a synthetic single-exon transcript and the public `normalize` API — no overlap with the symptom-locked tests in `tests/coverage_gap_tests.rs` or `tests/ins_shift_matrix.rs`.

- **#94** — every speculative `FIXME` lock added in #99 has now been resolved by either #98 (orientation symptoms) or #96 (this PR, position-shape symptom). The header docstring is updated; no `FIXME(...)` markers remain in the file.
- **#81 A6** — multi-base `dup` 3'-shift is part of the broader A6 matrix work, but the spec-canonical form for the multi-base intronic minus-strand case is now correct end-to-end as a side effect of #98 + #96.
- **#97** — independent (UTR codepath); still its own follow-up PR.

## Verification

- [x] `cargo nextest run --features dev` — 3180/3180 pass; the previously-failing six (#98 cascade) and seven (#96 cascade) tests now pass with their updated locks; idempotency suite green
- [x] `cargo nextest run --features dev --test issue_96_repeat_unit_position_consistency` — 4/4 pass
- [x] Pre-commit hooks pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- *(normalize)* Repeat notation now emits the position of the FIRST
+  repeat unit per HGVS spec, not the entire reference tract. The three
+  repeat-emission paths in `src/normalize/rules.rs`
+  (`duplication_to_repeat`, `insertion_to_repeat`, `normalize_repeat`)
+  previously set the position end to `ref_start + ref_count * unit_len
+  - 1`, producing forms like `c.342_362GCA[9]` for a 9-unit GCA tract
+  (21-base position) where the spec requires `c.342_344GCA[9]` (3-base
+  position matching the 3-base unit length). The fix sets `end =
+  ref_start + unit_len - 1` everywhere repeat notation is emitted, so
+  the position-range length always equals the unit length.
+  Mechanically updates the locked outputs of every multi-base
+  ins/dup→repeat case in `tests/ins_shift_matrix.rs` (29 cells across
+  all 7 nucleotide coord-system / strand modules), three real-accession
+  smoke tests in `tests/normalize_tests.rs`, and the residual #96 lock
+  in `tests/coverage_gap_tests.rs`. Issue #96.
 - *(normalize)* Minus-strand intronic normalization now reads the
   reference window in transcript-view orientation. `normalize_intronic_cds`
   and `normalize_intronic_tx` previously passed the genomic-strand bytes
--- a/src/normalize/rules.rs
+++ b/src/normalize/rules.rs
@@ -193,12 +193,16 @@ pub fn insertion_to_repeat(
 
     let (unit, ref_start, ref_count) = best?;
     let total_count = ref_count + added_copies;
-    let ref_end = ref_start + ref_count as usize * unit.len() - 1;
+    // Per HGVS spec, the position in `prefix:position_or_range unit[N]`
+    // must locate the FIRST repeat unit, not the entire tract. The
+    // first-unit end position is `ref_start + unit_len - 1` (inclusive,
+    // 0-based), not `ref_start + ref_count * unit_len - 1`. Issue #96.
+    let unit_first_end = ref_start + unit.len() - 1;
     Some((
         unit[0],
         total_count,
         index_to_hgvs_pos(ref_start),
-        index_to_hgvs_pos(ref_end),
+        index_to_hgvs_pos(unit_first_end),
         unit,
     ))
 }
@@ -472,11 +476,14 @@ pub fn duplication_to_repeat(ref_seq: &[u8], start: u64, end: u64) -> Option<Dup
         if let Some(analysis) = find_homopolymer_at(ref_seq, start_idx) {
             if analysis.base == Some(first) {
                 let total_count = analysis.ref_count + dup_len as u64;
+                // Per HGVS spec the position must locate the FIRST
+                // repeat unit, not the full ref-tract extent. For a
+                // single-base unit that is a 1-base position. Issue #96.
                 return Some(DupToRepeatResult::Homopolymer {
                     base: first,
                     count: total_count,
                     start: index_to_hgvs_pos(analysis.ref_start),
-                    end: index_to_hgvs_pos(analysis.ref_start + analysis.ref_count as usize - 1),
+                    end: index_to_hgvs_pos(analysis.ref_start),
                 });
             }
         }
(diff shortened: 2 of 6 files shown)


nh13:
Closing — this PR misreads the HGVS DNA repeat spec and should not merge.

The spec's own examples on `varnomen.hgvs.org/recommendations/DNA/variant/repeated/` use **position = full reference tract** with **count = post-edit unit count** uniformly:

- `NC_000014.8:g.101179660_101179695TG[14]` — 36-base position for a 2-base TG unit (= 18 TG units = full ref tract); count 14 is post-edit.
- `NM_000333.3:c.89_118AGC[13]` — 30-base position for a 3-base AGC unit (= 10 units = full ref tract); count 13 is post-edit.
- `NC_000012.11:g.112036755_112036823CTG[9]TTG[1]CTG[13]` — 69-base composite, full tract.

The same spec page carries a Community Consultation note explicitly mandating the full-range form: "g.123_191CAG[23], not g.123CAG[23]" — i.e., the spec is moving *toward* full-range, not toward the first-unit form this PR introduces.

ferro's pre-PR emission (`c.342_362GCA[9]` from PR #93's A7 work) matches the spec pattern exactly: 21-base position = 7 GCA units = full ref tract, count 9 = post-edit total. PR #93 had it right.

What this PR's `c.342_344GCA[9]` output actually is: a hybrid that pairs a *one-unit* position with a *post-edit* count. That form does not appear anywhere in the spec — it's neither sequence-anchored full-range explicit-unit (`c.342_362GCA[9]`), nor range-anchored implicit-unit (`c.342_362[9]`), nor composite. The misreading conflated the RNA range-anchored shape with the DNA explicit-unit shape.

So the position-shape concern in issue #96 was based on my misinterpretation. Only the wrong-strand-letter symptom in #96 (`T[6]` vs `A[6]`) was a real bug, and it's already fixed by #100 (#98 root-cause fix).

Follow-ups:
- The `FIXME(#96)` comment that #99/#100 added to `tests/coverage_gap_tests.rs::intronic_duplications::test_minus_strand_intronic_multi_base_dup` will be removed (the locked value `c.30+2_30+4A[6]` is the spec-compliant form once #100's letter fix lands).
- Issue #96 will be closed with a correction.

Thanks for the catch.

PR #102, opened 2026-05-03
fix(normalize): 5'UTR CDS↔tx off-by-one collapsed UTR del to c.? (#97)

### LINKED ISSUE #97: normalize: minus-strand 5'UTR single-base del collapses to c.?del instead of resolved position

nh13:
Closes #97. Stacked on #100 (will retarget through the chain as the upstream PRs merge).

## TL;DR

HGVS coordinate numbering skips `c.0` — `c.-1` is the base immediately 5' of `c.1`, not the base 5' of `c.0`. ferro had three coordinate-conversion implementations; one (`convert::mapper::cds_to_tx` / `tx_to_cds`) handled this correctly, and two (`Normalizer::cds_to_tx_pos` and `convert::coding::cds_to_transcript_pos`) used the formula `cds_start + base - 1` uniformly — which double-counts the missing-zero step for negative `base`. The inverse in `Normalizer::tx_to_cds_pos` had the mirror bug. The most visible symptom was a 5'UTR single-base deletion on a minus-strand transcript collapsing to `c.?del`.

## Root cause

For `pos.base = -1` (the base immediately upstream of c.1) with `cds_start = X`:

- Spec-correct: `tx = X - 1`.
- Buggy formula `tx = cds_start + base - 1` = `X + (-1) - 1` = `X - 2`. Off by one.

For the inverse, `pos = cds_start - 1` should map to `base = -1`:

- Spec-correct: `base = pos - cds_start = -1`.
- Buggy formula `base = pos - cds_start + 1` = `0`. `CdsPos::Display` renders `base == CDS_BASE_UNKNOWN (0)` as `?`.

The two formulas are internally consistent (round-trip works), so no test caught it until normalization shifted to a tx position whose buggy inverse landed on `base = 0`. This happened on a minus-strand 5'UTR fixture where the UTR is a homopolymer: the input `c.-3del` mapped to tx `cds_start - 4` (buggy), the 3'-shift walked through the homopolymer to `cds_start - 1`, and the buggy inverse mapped that to `base = 0` → `c.?del`.

## Fix

`src/normalize/mod.rs::cds_to_tx_pos`:

| `pos.base` | Old formula | New formula |
|---|---|---|
| `< 0` | `cds_start + base - 1` (buggy) | `cds_start + base` |
| `== 0` | `cds_start + 0 - 1 = cds_start - 1` | preserved as `cds_start - 1` (legacy; `c.0` isn't valid HGVS but historical inputs can land here) |
| `>= 1` | `cds_start + base - 1` | unchanged |

`src/normalize/mod.rs::tx_to_cds_pos`:

| `pos` | Old formula | New formula |
|---|---|---|
| `< cds_start` | `pos - cds_start + 1` (buggy) | `pos - cds_start` |
| `>= cds_start && <= cds_end` | `pos - cds_start + 1` | unchanged |

`src/convert/coding.rs::cds_to_transcript_pos` gets the same forward fix for parity. It currently has no callers in `src/` but is a public function and could be called externally.

`src/convert/mapper.rs::cds_to_tx` / `tx_to_cds` were already correct (they treat the CDS step as `pos.base - 1` and the 5'UTR step as `pos.base`, separately) — unchanged.

## Diagnostic suite

`tests/issue_97_utr_del_position.rs` (new) — three single-assertion tests on a minimal minus-strand UTR fixture (`AAAAA` 5'UTR at `c.-5..c.-1`):

| Test | Asserted |
|---|---|
| `five_prime_utr_del_shifts_to_minus_one` | `NM_MUTR.1:c.-3del → c.-1del` |
| `five_prime_utr_del_round_trip` | `c.-1del → c.-1del` |
| `five_prime_utr_substitution_round_trip` | `c.-3A>G → c.-3A>G` (sub doesn't shift; pinned to catch any mirror-bug regression) |

All three fail before the fix and pass after.

## Cascade

| Test | Old | New |
|---|---|---|
| `coverage_gap_tests::utr_minus_strand::test_5prime_utr_deletion_minus_strand` | `c.?del` (with `FIXME(#97)`) | `c.-1del` (FIXME removed) |
| `normalize::tests::test_cds_to_tx_pos_utr5_valid` | tx 1 (cds_start=5, c.-3) | tx 2 |
| `normalize_tests::clinvar_normalization::test_verified_normalization_indels_batch1::case_5` (`NM_001394148.2:c.-56_-47dup`) | `c.-55_-46dup` | `c.-56_-47dup` |

The clinvar case is worth flagging: ferro's pre-fix output `c.-55_-46dup` matched the test's documented "ferro-vs-Mutalyzer" expectation ("we apply 3' shift in UTR, Mutalyzer doesn't"). With the fix, ferro emits `c.-56_-47dup` — **matching Mutalyzer**. Tracing through the new formula: the 10-base region at the spec-correct tx range is already in its 3'-most equivalent representation in the local tract, so the 3'-shift is a no-op. The pre-fix "shift" was a side effect of the off-by-one (the buggy forward read a shifted 10-base region whose 3'-most form was the spec-correct tx range, then the buggy inverse mapped that back). The test comment is updated to "Mutalyzer: agrees" and the source-of-truth annotation explains the pre-fix vs post-fix delta.

- **#97** — closed by this PR.
- **#94** — together with #100, all `FIXME(#NN)` markers in `tests/coverage_gap_tests.rs` are now resolved (intronic-orientation symptoms via #100; UTR collapse symptom via this PR).
- **#81 J2** (n. UTR `*` and `-` markers) — separate; not addressed here. The `n.` coordinate path is in `convert/mapper.rs` which already handles the c.0 skip correctly. If a similar bug exists in another `n.`-flavored helper it would be its own follow-up; the grep on `cds_start as i64` patterns surfaced only the three locations fixed here.

## Verification

- [x] `cargo nextest run --features dev --test issue_97_utr_del_position` — 3/3 pass
- [x] `cargo nextest run --features dev` — full suite passes (3179/3179); the three previously-failing tests now pass with their updated locks; no unrelated regressions
- [x] Pre-commit hooks pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   for 3'UTR / downstream). Cross-region pairs (`c.[-1A>G;1A>T]` 5'UTR↔CDS,
   `c.[40C>T;*1A>G]` CDS↔3'UTR, …) still correctly do not merge.
   Issue #89.
+- *(normalize)* CDS ↔ transcript coordinate mappings now respect the
+  HGVS no-c.0 numbering rule for 5'UTR positions. The forward mapping
+  (`Normalizer::cds_to_tx_pos`, `convert::coding::cds_to_transcript_pos`)
+  previously computed `tx = cds_start + base - 1` for negative `base`,
+  which double-counted the gap between c.-1 and c.1 and emitted a tx
+  position one base 5' of the true location. The inverse mapping
+  (`Normalizer::tx_to_cds_pos`) had the mirror bug: tx positions one
+  base before `cds_start` mapped to `base = 0`, which `CdsPos::Display`
+  renders as `c.?` (`CDS_BASE_UNKNOWN`). The most visible symptom was a
+  5'UTR single-base deletion on a minus-strand transcript collapsing
+  to `c.?del` instead of resolving to a real position (e.g. `c.-1del`
+  after 3'-shifting within a UTR homopolymer). Forward and inverse are
+  now `tx = cds_start + base` and `base = tx - cds_start` for negative
+  / pre-cds_start positions, matching the spec's "c.-1 is one base 5'
+  of c.1" rule and the existing exon-aware mapper at
+  `convert::mapper::cds_to_tx` / `tx_to_cds`. Issue #97.
 - *(normalize)* Minus-strand intronic normalization now reads the
   reference window in transcript-view orientation. `normalize_intronic_cds`
   and `normalize_intronic_tx` previously passed the genomic-strand bytes
--- a/src/convert/coding.rs
+++ b/src/convert/coding.rs
@@ -83,8 +83,15 @@ pub fn cds_to_transcript_pos(pos: &CdsPos, transcript: &Transcript) -> Result<u6
 
     let tx_pos = if pos.utr3 {
         cds_end + pos.base as u64
-    } else if pos.base < 1 {
-        (cds_start as i64 + pos.base - 1) as u64
+    } else if pos.base < 0 {
+        // 5'UTR: HGVS numbering skips c.0 (c.-1 is the base immediately
+        // upstream of c.1), so c.-N maps to tx position cds_start - N.
+        // Issue #97.
+        (cds_start as i64 + pos.base) as u64
+    } else if pos.base == 0 {
+        // c.0 is not a valid HGVS position; preserve the legacy mapping
+        // (last 5'UTR base, equivalent to c.-1) rather than failing.
+        cds_start.saturating_sub(1)
     } else {
         cds_start + pos.base as u64 - 1
     };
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #79, opened 2026-04-30
normalize: codon-frame exception for c. SNV pairs separated by one nucleotide


msto:
The HGVS spec on delins (<https://hgvs-nomenclature.org/stable/recommendations/DNA/delins/>) carries a carve-out to the "non-adjacent variants are described separately" rule:

> Notes: *"two variants separated by one or more nucleotides should preferably be described individually and not as a 'delins'. Exception: two variants separated by one nucleotide, together affecting one amino acid, should be described as a 'delins'."*

The Examples section makes the carve-out concrete:

> `c.[145C>T;147C>G]` is *not correct*. The required form is `c.145_147delinsTGG` because positions 145–147 fall in a single codon.

Issue #72 implements the strictly-consecutive case but defers this codon-frame exception. Implementing it requires:

1. **Codon-frame check.** For a CDS exonic SNV pair at positions `a` and `c` with `a + 2 == c`, they share a codon iff `(a - 1) / 3 == (c - 1) / 3` (0-indexed codon number). Applies only when both positions are ≥ 1, with no offset, and not in 5'UTR / 3'UTR.
2. **Reference-provider lookup for the unchanged middle nucleotide.** The merged delins alt is `<a-alt><ref(a+1)><c-alt>`. Position `a + 1` must be fetched from the reference provider. When no provider is available or the lookup fails, the variants pass through unchanged with a `NormalizationWarning` rather than erroring.
3. **Integration with #72's merge pass.** After the strictly-consecutive walk completes, run a second walk that attempts the codon-frame merge on adjacent SNV pairs the first walk left unmerged. The merged result feeds back into the strictly-consecutive walk so a third SNV at `c + 1` would still merge into the delins via #72's existing adjacency rule.

| Input | Spec form |
|---|---|
| `c.[145C>T;147C>G]` (positions 145–147 share a codon) | `c.145_147delinsTGG` (assuming reference 146 = `G`) |
| `c.[3C>T;5G>A]` (positions 3 and 5 straddle codon 1/codon 2) | unchanged |
| `c.[145C>T;148C>G]` (gap of 2 nt) | unchanged |
| `g.[145G>A;147C>T]` (no codon frame in `g.`) | unchanged |

**Out of scope for this issue (further follow-ups if needed):**

- Codon-frame exception for `r.` coding regions
- Chains involving more than two SNVs across a codon (e.g., `sub`, `ins`, `sub` straddling a codon)
[...]

Issue #72, opened 2026-04-30
Merge consecutive edits within alleles per HGVS spec


msto:
The HGVS spec requires consecutive edits within an allele to be expressed as a single combined operation:

> Changes involving two or more consecutive nucleotides are described as deletion/insertion (delins) variants.

The spec explicitly notes that `c.[79G>T;80C>T]` is not a valid description; the correct form is `c.79_80delinsTT`. The same principle applies to consecutive deletions.

`normalize()` currently preserves the input form for these cases. Adding a merging pass would bring allele output into spec compliance.

| Input | Spec form |
|---|---|
| `g.[1000G>A;1001A>C]` | `g.1000_1001delinsAC` |
| `g.[1000del;1001del]` | `g.1000_1001del` |

Non-adjacent variants (one or more intervening positions) must remain separate per the spec — only consecutive edits are merged.

Spec: https://hgvs-nomenclature.org/stable/recommendations/DNA/delins/

PR #103, opened 2026-05-03
fix(normalize): merge same-region UTR adjacency in cis alleles (#89)

### LINKED ISSUE #89: normalize: same-region UTR adjacency in c./n./r. cis allele merge
## Background

PR #80 implements consecutive-edit merging in cis alleles for #72. The design doc (`docs/superpowers/specs/2026-04-30-merge-consecutive-allele-edits-design.md`) describes the position-eligibility rule as:

> Not crossing a coordinate-system boundary. For `c.` this means 5'UTR (`-N`), CDS (`1..=N`), and 3'UTR (`*N`) are **mutually exclusive merge regions** — there is no valid HGVS range syntax that spans these (e.g., `c.-1_1` does not exist), so adjacency cannot be expressed even when positions are physically consecutive. The same logic applies to `n.` (upstream `-N` / transcript / downstream `*N`) and `r.` (UTR boundaries).

The wording (\"mutually exclusive merge regions\") implies same-region adjacency *within* a UTR is allowed; only crossings are barriers. The implementation, however, drops every UTR position outright.

## Observed

The `simple_cds_pos`, `simple_tx_pos`, and `simple_rna_pos` helpers in `src/normalize/merge.rs` (≈lines 260–305) reject any position with `is_5utr()` / `is_3utr()` / `is_upstream()` / `is_downstream()`, so adjacent same-region UTR pairs never reach the merge pass:

| Input | Current output | Spec-preferred form |
|---|---|---|
| `c.[-2A>G;-1C>T]` (both 5'UTR) | unchanged | `c.-2_-1delinsGT` |
| `c.[*1A>G;*2C>T]` (both 3'UTR) | unchanged | `c.*1_*2delinsGC` |
| `n.[-2A>G;-1C>T]` (both upstream) | unchanged | `n.-2_-1delinsGT` |
| `n.[*1A>G;*2C>T]` (both downstream) | unchanged | `n.*1_*2delinsGC` |
| `r.[-2a>g;-1c>u]` (both 5'UTR) | unchanged | `r.-2_-1delinsgu` |

The CDS/UTR-crossing barrier (`c.[-1A>G;1G>T]`) is correctly preserved as a non-merge — that part of the design doc is not in scope of this issue.

## Why the simple fix isn't enough

`simple_cds_pos` returns `Option<u64>` — it cannot distinguish 5'UTR `c.-1`, CDS `c.1`, and 3'UTR `c.*1`, which all map to `base = 1`. To correctly pin same-region adjacency while still blocking crossings, the helper needs to return a region-aware key (e.g. an enum `{ FivePrimeUtr(u64), Cds(u64), ThreePrimeUtr(u64) }` for `c.`; analogous shapes for `n.` and `r.`). Adjacency then becomes \"same variant + same region + `A.end + 1 == B.start`.\"

For `c.`, the natural axis is the signed CDS coordinate where `-N` is 5'UTR, `1..=M` is CDS, and `*N` is 3'UTR. CDS adjacency on the signed axis already excludes `c.-1`/`c.1` (they're not adjacent on the integer line because the spec defines no `c.-1_1` range). Same for `c.M`/`c.*1`. So a signed `i64` axis with explicit region tagging may suffice — but the implementation has to carry the region through to the merged variant's position too, since the merged delins keeps its UTR/CDS form.

## Suggested approach

1. Change `simple_cds_pos` / `simple_tx_pos` / `simple_rna_pos` to return a region-tagged key (e.g. `(Region, u64)`) instead of plain `u64`.
2. In the adjacency check, require both ends to have the same `Region` before checking `A.end + 1 == B.start` on the integer key.
3. Carry the region through into `build_cds_merged` / `build_tx_merged` / `build_rna_merged` so the merged anchor reconstructs the correct `CdsPos` / `TxPos` / `RnaPos` (with the right sign for `c.` and the right marker for `n.`/`r.`).
4. Add positive merge tests for same-region adjacency (5'UTR, 3'UTR, upstream, downstream) and pin the existing crossing tests as still-not-merging.
5. Update `tests/merge_consecutive_edits_tests.rs` round-trip negative-case list — the current `test_no_merge_utr_boundary` covers the crossing case and stays; add a new test that the crossing case `c.[-1A>G;1G>T]` still does not merge after this change.

## Out of scope

- Codon-frame exception (already tracked in #79)
- Any `n.` or `r.` semantics not covered by the existing position predicates
- 3' shifting interaction with the new same-region merge (likely no interaction since intronic/UTR positions don't shift, but worth a regression test)

## References

- #72 — parent feature (cis allele consecutive-edit merge)
- #79 — sibling deferral (codon-frame exception in c.)
- #80 — implementing PR; CodeRabbit review surfaced this gap
- Design doc: `docs/superpowers/specs/2026-04-30-merge-consecutive-allele-edits-design.md`
- Tracking issue: #81 item J1 (boundary crossings; this issue is the same-region complement)

nh13:
Closes #89.

## TL;DR

PR #80 (issue #72) implemented consecutive-edit merging in cis alleles. Its position-eligibility helpers (`simple_cds_pos` / `simple_tx_pos` / `simple_rna_pos`) rejected every `is_5utr()` / `is_3utr()` / `is_upstream()` / `is_downstream()` position outright, so adjacent same-region UTR pairs (`c.[-2A>G;-1C>T]`, `c.[*1A>G;*2C>T]`, `r.[-2a>g;-1c>u]`, …) never reached the merge pass — even though HGVS allows ranges *within* those regions (`c.-2_-1` and `c.*1_*2` are valid syntax). This PR replaces the bare `Option<u64>` keys with a `Region`-tagged `(Region, i64)` axis so adjacency becomes "same region + integer adjacency". Cross-region pairs (the existing 5'UTR↔CDS barrier and the new CDS↔3'UTR barrier) still correctly do not merge.

## Root cause

The design doc said:

> Not crossing a coordinate-system boundary. For `c.` this means 5'UTR (`-N`), CDS (`1..=N`), and 3'UTR (`*N`) are **mutually exclusive merge regions** — there is no valid HGVS range syntax that spans these (e.g., `c.-1_1` does not exist), so adjacency cannot be expressed even when positions are physically consecutive.

"Mutually exclusive merge regions" implies same-region adjacency *within* each region is valid; only crossings are barriers. The implementation went further: every UTR position was dropped pre-eligibility, so even `c.-2 + 1 == c.-1` never got to the adjacency check. Single-axis `Option<u64>` couldn't distinguish `c.-1` (5'UTR), `c.1` (CDS), and `c.*1` (3'UTR) — all three would map to integer `1` if rejection were lifted naively — so a region tag became necessary.

## Fix

A new `Region` enum:

| Region | Used by | Axis | Example |
|---|---|---|---|
| `Genome` | `g.`, `m.` | positive i64 | `g.100` |
| `Cds` | `c.`, `r.` | positive i64 | `c.1` |
| `FivePrimeUtr` | `c.`, `r.` | negative i64 | `c.-3 → -3` |
| `ThreePrimeUtr` | `c.`, `r.` | positive i64 (`*N` count) | `c.*1 → 1` |
| `Tx` | `n.` | positive i64 | `n.100` |
| `TxUpstream` | `n.` | negative i64 | `n.-3 → -3` |
| `TxDownstream` | `n.` | positive i64 (`*N` count) | `n.*1 → 1` |

The integer axis is signed because 5'UTR / upstream regions use negative values and adjacency `prev.end + 1 == next.start` works naturally — `c.-3 → c.-2 → c.-1` maps to `-3 → -2 → -1`. 3'UTR and `Cds` (or downstream and transcript-body) both use positive integers, but the region tag distinguishes them so a c.40 / c.*1 pair fails region-equality even though both have positive axis values.

`Anchor` gains a `region: Region` field and switches `start`/`end` from `u64` to `i64`. The chain logic in `merge_consecutive_edits` enforces region equality before the existing integer-adjacency check:

```rust
if prev_a.region != next_region {
    break 'try_merge false;
}
if prev_a.end.checked_add(1) != Some(next_start) {
    break 'try_merge false;
}
```

`build_cds_merged` / `build_tx_merged` / `build_rna_merged` consume the region tag to reconstruct the right output position shape (negative `base` for 5'UTR / upstream, `utr3` / `downstream` flag for 3'UTR / downstream). `build_naedit` takes a `Region`-aware closure of the form `FnMut(Region, i64) -> P`.

`MockProvider`-only test fixture `provider_with_utr_transcript()` adds a 60-base transcript with 5'UTR (10) + CDS (40) + 3'UTR (10), with substitution ref-bases chosen to match the test inputs.

## Tests

| Test | Asserted |
|---|---|
| `test_merge_5utr_consecutive_subs_cds` | `c.[-2A>G;-1C>T]` → `c.-2_-1delinsGT` |
| `test_merge_3utr_consecutive_subs_cds` | `c.[*1A>G;*2C>T]` → `c.*1_*2delinsGT` |
| `test_merge_5utr_consecutive_subs_rna` | `r.[-2a>g;-1c>u]` → `r.-2_-1delinsgu` |
| `test_merge_3utr_consecutive_subs_rna` | `r.[*1a>g;*2c>u]` → `r.*1_*2delinsgu` |
| `test_no_merge_3utr_cds_boundary` | `c.[40C>T;*1A>G]` round-trips unchanged (CDS↔3'UTR crossing barrier) |

Companion to the existing `test_no_merge_utr_boundary` (5'UTR↔CDS crossing) which continues to pass unchanged.

n. (TxPos) tests are not added in this PR — the fix covers `n.` semantically (region tags `Tx` / `TxUpstream` / `TxDownstream`), but a dedicated `n.` test fixture would need a separate non-coding accession + transcript. The merge-eligibility logic and build path are exercised symmetrically with the `c.` cases via the same code path, so n. coverage can be added in a follow-up if desired.

## What this means for related issues

- **#89** — closed by this PR.
- **#79** — sibling (codon-frame exception for `c.` SNV pairs separated by one nucleotide). Will stack on this PR; the new region-tagged helpers make the codon-frame eligibility branch's "CDS-only" constraint clean to express (it fires only on `Region::Cds`).
- **#81 J1** — boundary-crossing complement. Untouched here. Already cross-references this issue in the umbrella tracker.

## Verification

- [x] `cargo nextest run --features dev --test merge_consecutive_edits_tests` — 50/50 pass (1 `#[ignore]`d codon-frame test for #79)
- [x] `cargo nextest run --features dev` — full suite 3177/3177 pass; no unrelated regressions
- [x] Pre-commit hooks pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- *(normalize)* Cis-allele consecutive-edit merging now collapses
+  adjacent sub-variants *within* a UTR / upstream / downstream region,
+  not only within the CDS / transcript body. The PR #80 implementation
+  rejected every `is_5utr() / is_3utr() / is_upstream() / is_downstream()`
+  position outright, so inputs like `c.[-2A>G;-1C>T]` (both 5'UTR) and
+  `c.[*1A>G;*2C>T]` (both 3'UTR) round-tripped unchanged even though
+  HGVS allows ranges within those regions (`c.-2_-1`, `c.*1_*2`). The
+  fix replaces the `Option<u64>` position keys with a `Region`-tagged
+  `(Region, i64)` axis (covering CDS, 5'UTR, 3'UTR, transcript-body,
+  upstream, downstream, and genomic / mitochondrial); merge eligibility
+  becomes "same region + integer adjacency on the region's axis".
+  `build_cds_merged` / `build_tx_merged` / `build_rna_merged` consume
+  the region tag to reconstruct the right `CdsPos` / `TxPos` / `RnaPos`
+  shape (negative base for 5'UTR / upstream, `utr3` / `downstream` flag
+  for 3'UTR / downstream). Cross-region pairs (`c.[-1A>G;1A>T]` 5'UTR↔CDS,
+  `c.[40C>T;*1A>G]` CDS↔3'UTR, …) still correctly do not merge.
+  Issue #89.
 - *(normalize)* Minus-strand intronic normalization now reads the
   reference window in transcript-view orientation. `normalize_intronic_cds`
   and `normalize_intronic_tx` previously passed the genomic-strand bytes
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -10,15 +10,49 @@ use crate::hgvs::variant::{
     AllelePhase, CdsVariant, GenomeVariant, HgvsVariant, LocEdit, MtVariant, RnaVariant, TxVariant,
 };
 
+/// Coordinate-system region used as the merge-eligibility key.
+///
+/// Adjacency in the merge pass requires both ends to share a region.
+/// HGVS forbids ranges that span the 5'UTR↔CDS, CDS↔3'UTR (for `c.`/`r.`)
+/// or upstream↔transcript / transcript↔downstream (for `n.`) boundaries
+/// — `c.-1_1`, `c.40_*1`, etc. do not exist as valid range syntax — so
+/// we tag each position with its region and refuse to merge across.
+/// The integer `start`/`end` axis is region-local: positive for `Cds` /
+/// `ThreePrimeUtr` / `Tx` / `TxDownstream`, negative for `FivePrimeUtr`
+/// / `TxUpstream`. Adjacency `prev.end + 1 == next.start` works
+/// naturally for all six because the axis is monotonic 5'→3' within
+/// each region (`c.-3 → c.-2 → c.-1` maps to `-3 → -2 → -1`).
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum Region {
+    /// `g.` (genomic) and `m.` (mitochondrial) — single axis, no
+    /// sub-regions.
+    Genome,
+    /// `c.` / `r.` CDS proper (positive non-UTR base).
+    Cds,
+    /// `c.` / `r.` 5'UTR (negative base, e.g. `c.-3`).
+    FivePrimeUtr,
+    /// `c.` / `r.` 3'UTR (`utr3` flag, e.g. `c.*1`).
+    ThreePrimeUtr,
+    /// `n.` transcript body (positive non-downstream base).
+    Tx,
+    /// `n.` upstream of transcript (negative base, e.g. `n.-3`).
+    TxUpstream,
+    /// `n.` downstream of transcript (`downstream` flag, e.g. `n.*1`).
+    TxDownstream,
+}
+
 /// Anchor for a single sub-variant.
 ///
-/// `start` and `end` are 1-based inclusive position bounds. For an `Insertion`
-/// between positions `p` and `p+1`, `start = p+1` and `end = p` (empty range
-/// at a boundary).
+/// `start` and `end` are 1-based inclusive position bounds on the
+/// `region` axis. For an `Insertion` between positions `p` and `p+1`,
+/// `start = p+1` and `end = p` (empty range at a boundary). The
+/// integer axis is signed because 5'UTR / upstream regions use
+/// negative values.
 #[derive(Debug, Clone)]
 struct Anchor {
-    start: u64,
-    end: u64,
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #109, opened 2026-05-06
fix(normalize): rewrite revcomp delins as inversion (#81 A2)

nh13:
## Summary

Implements item A2 of issue #81: a `delins` whose inserted sequence is the reverse complement of the deleted reference normalizes to an `inv`, per the HGVS spec definition.

Spec basis (vendored at `assets/hgvs-nomenclature/docs/recommendations/DNA/`):

- `inversion.md`: *"a sequence change where, compared to a reference sequence, **more than one nucleotide** replacing the original sequence are the reverse complement of the original sequence"* and *"the description g.234inv is therefore not allowed; a one nucleotide inversion should be described as a substitution"*.
- `delins.md`: *"a sequence change where ... one or more nucleotides are replaced by one or more other nucleotides **and which is not** a substitution, inversion or conversion"*.

Example: `g.100_102delinsTAG` where `ref = CTA` now normalizes to `g.100_102inv`. The complementary-outer-bases shortening rule applies to the result so `g.100_104delinsCATAG` over `CTATG` lands as `g.101_103inv` — the same canonical output a directly-encoded `inv` produces.

## What's in the PR

1. **A2 rule (commit 1).** New `canonicalize_delins() -> DelinsCanonical` in `src/normalize/rules.rs` unifies the four old `delins_is_*` boolean helpers into one decision function expressing HGVS edit-priority (sub > del > inv > dup > ins) in one place. Adds the inversion branch and emits minimal notation. Removes the unreachable second `NaEdit::Delins` arm at `src/normalize/mod.rs:1840` (the first arm always early-returns from `normalize_na_edit`).

2. **Inversion-arm fix (commit 2).** While testing A2 round-trip equivalence, discovered that `needs_normalization()` did not include `NaEdit::Inversion`, so direct `inv` inputs bypassed `normalize_na_edit` entirely and the existing `shorten_inversion` logic at `src/normalize/mod.rs:1542` was dead in practice. Existing RULE 10 tests were too weak (only `result.contains("inv")`) to have caught this. Now: Inversion is included in `needs_normalization()`; the inversion arm emits minimal `NaEdit::Inversion { sequence: None, length: None }` after shortening (the input's explicit `sequence`/`length` would no longer match the shortened interval); and RULE 10 tests are tightened to assert exact normalized output for no-shortening, palindrome→identity, and partial-cascade cases.

3. **Test coverage (commits 1, 2, 3).** New rust unit tests for `canonicalize_delins()`, new RULE 9b integration tests in `tests/normalize_tests.rs` (including round-trip equivalence between `delins`-encoded and directly-encoded inversions), strengthened RULE 10 tests, and three new Python integration tests in `tests/python/test_core.py`.

## Test plan

- [x] `cargo nextest run --features dev` — 3417 tests pass
- [x] `pytest tests/python/` — 122 tests pass
- [x] `cargo clippy --features dev -- -D warnings` — clean
- [x] Round-trip: `c.X_Yinv` and equivalent `c.X_YdelinsZZZ` produce the same normalized string
- [x] Priority guard: 1-base complement (`c.1delinsT` over `A`) routes to substitution `c.1A>T`, never to inversion (spec requires N>=2)
- [x] Palindrome guard: `c.X_Ydelins<palindrome>` hits Identity, never Inversion

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,12 +11,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - *(normalize)* rewrite degenerate substitutions (ref == alt, e.g. `c.100A>A`) to identity (`=`) per HGVS v21 spec, which marks `c.X>X` as "not allowed" (`docs/recommendations/DNA/other.md`). The rule is purely syntactic on the edit's stated bases, so it fires in both the full-normalization path and the no-reference canonicalization path — `c.123C>C` rewrites to `c.123=` regardless of provider availability. ([#81](https://github.com/fulcrumgenomics/ferro-hgvs/issues/81) A4)
 - *(normalize)* rewrite a `delins` whose inserted sequence is empty as a deletion, per the HGVS spec requirement that an empty insert is semantically a deletion and must be rendered as `del`. The rewritten deletion is then 3'-shifted under the standard del rule. Issue [#81](https://github.com/fulcrumgenomics/ferro-hgvs/issues/81) item A3.
+- *(normalize)* rewrite a `delins` whose inserted sequence is the reverse complement of the deleted reference as an inversion, per the HGVS spec definition of `inv`. The complementary-outer-bases shortening rule applies to the result so that a `delins`-encoded inversion produces the same canonical output as a directly-encoded `inv`. Issue [#81](https://github.com/fulcrumgenomics/ferro-hgvs/issues/81) item A2.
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #105, opened 2026-05-03
test: pin normalize() against HGVS v21.0 spec fixture (closes #84)

nh13:
## Summary

Adds a regression fixture that pins ferro's current `normalize()` output for every variant string in the HGVS v21.0 spec, plus the tooling and CI hooks needed to keep it in sync. Companion to the audit checklist in #83 — every row whose `current` diverges from `spec_expected` carries a todo link to that issue.

- **Fixture:** `tests/fixtures/grammar/hgvs_spec_normalization.json` — 798 rows pulled from the vendored spec at `assets/hgvs-nomenclature/` (git submodule pinned to tag `21.0.0`, sha `a32f970d`). Sources: `docs/recommendations/**/*.md`, `docs/background/{numbering,refseq,simple}.md`, `docs/consultation/SVD-WG###.md` (skipping WG008 which has no new HGVS forms), `docs/syntax.yaml`, plus 29 canonical-shape probes ported from the deleted `tests/hgvs_spec_compliance_tests.rs`.
- **Generator:** `examples/generate_spec_fixture.rs` — pure Rust (`pulldown-cmark` + `serde_yaml`). Runs each candidate through `Normalizer::new(MockProvider::new())`, classifies (`compliant` / `needs-reference` / `reject`), merges hand overrides, and writes a deterministically sorted JSON file. `--check` mode in CI catches byte drift.
- **Override file:** `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json` — empty seed; reviewers can reclassify rows or override `spec_expected` for `pair` rows as #83's audit progresses. Schema is documented in `CONTRIBUTING.md`.
- **Test:** `tests/hgvs_spec_normalization_tests.rs` — loads the fixture, asserts `normalize(parse(input))` produces exactly `current` for each row, skips `needs-reference` until #82's `from_manifest` lands.
- **Property tests:** `tests/normalize_property_tests.rs` — ports the two strict transcript-driven tests (`test_normalized_output_is_valid_hgvs`, `test_normalization_idempotency`) out of the deleted compliance test into a focused, fixture-independent file.
- **Cleanup:** deletes `tests/hgvs_spec_compliance_tests.rs` (parse-only, permissive thresholds against the obsolete 2024.01 fixture), `tests/fixtures/grammar/hgvs_spec_examples.json`, and `scripts/extract_hgvs_spec_examples.py` (live-site HTML scraper). All audited; only the two strict property tests carried real signal and are ported.
[...]

PR #111, opened 2026-05-06
fix(normalize): degenerate substitution (ref==alt) -> identity (#81 A4)

nh13:
## Summary

Addresses tracking #81 item A4: substitution edits where `reference == alternative` (e.g. `c.100A>A`) now canonicalize to identity (`=`) per the HGVS v21 spec. The spec marks `c.X>X` as "not allowed" (`docs/recommendations/DNA/other.md`) and gives `c.123=` as the canonical alternative — so this is the spec-correct rewrite, and is the SNV companion to the same-base delins → identity rule already implemented for #75 / PR #78.

## Design

The rule is purely syntactic on the edit's stated bases, so it must fire regardless of coordinate system, strand, or provider availability. Two paths handle it:

- `normalize_na_edit` (full normalization with reference data): early return after reference validation, before shuffling. `needs_normalization` now returns `true` for the degenerate-sub case so this path is reached; real substitutions (`ref != alt`) keep the existing no-op fast path.
- `canonicalize_edit` / `should_canonicalize` (no-reference fallback): used by `canonicalize_*_variant` when the provider has no transcript / genomic sequence loaded. The rewrite fires even with an empty `MockProvider`.

This second path was added after rebasing on `main` and noticing that the new HGVS v21 spec fixture (PR #105) generates with `MockProvider::new()` — a path that bypasses `normalize_na_edit` entirely. Without the no-reference fallback, the spec fixture row for `c.123C>C` did not flip; with it, ferro now produces `c.123=` from any provider state.

## Tests

Unit tests in `src/normalize/rules.rs`:
- `test_needs_normalization` extended for both Substitution branches
- `test_should_canonicalize` extended for both Substitution branches
- `test_canonicalize_edit_degenerate_substitution_to_identity` (new)

Integration tests in `src/normalize/mod.rs`:
- `c.10G>G` (ref matches actual) → `c.10=` — full normalize path
- `c.1A>A` (position-1 boundary) → `c.1=`
- `c.10G>T` (real SNV) stays unchanged — regression guard
- `NM_004006.2:c.123C>C` with `MockProvider::new()` → `c.123=` — no-reference path

Spec fixture (`tests/fixtures/grammar/hgvs_spec_normalization.json`):
- `c.123C>C` row's `current` flips from `c.123C>C` to `c.123=`
- Override file pins `spec_expected: "NM_004006.2:c.123="` per the v21 spec text, classifying the row as `preserved` with no audit todo

## Test plan

- [x] `cargo nextest run --features dev` (3403 tests passed)
- [x] `cargo run --features dev --example generate_spec_fixture -- --check`
- [x] `cargo clippy --features dev -- -D warnings`
- [x] `cargo fmt --all -- --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- *(normalize)* rewrite degenerate substitutions (ref == alt, e.g. `c.100A>A`) to identity (`=`) per HGVS v21 spec, which marks `c.X>X` as "not allowed" (`docs/recommendations/DNA/other.md`). The rule is purely syntactic on the edit's stated bases, so it fires in both the full-normalization path and the no-reference canonicalization path — `c.123C>C` rewrites to `c.123=` regardless of provider availability. ([#81](https://github.com/fulcrumgenomics/ferro-hgvs/issues/81) A4)
+
 ## [0.4.1](https://github.com/fulcrumgenomics/ferro-hgvs/compare/v0.4.0...v0.4.1) - 2026-05-04
 
 ### Added
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -1443,6 +1443,20 @@ impl<P: ReferenceProvider> Normalizer<P> {
             });
         }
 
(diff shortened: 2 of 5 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #113, opened 2026-05-06
fix(normalize): rewrite empty-insert delins as del (#81 A3)

nh13:
## Summary

- A `delins` whose inserted sequence is empty is semantically a deletion and per the HGVS spec must be rendered as `del`. Today ferro round-trips `c.10delins` as `c.10delins`; this PR rewrites it to `c.10del`.
- The rewrite is added at the top of the `NaEdit::Delins` arm in `normalize_na_edit` and recurses through the `NaEdit::Deletion` path, so the result picks up the standard del 3'-shift and reference validation instead of the delins arm's no-shift policy.

## Why normalize-time, not parse-time

ferro's posture is *lenient at parse, canonical at normalize*. The spec mandate is on output form, not input rejection — sitting at the same layer as the existing `delins → identity / sub / dup` rewrites in this same arm. The check is `O(1)` (`InsertedSequence::Empty` match) and short-circuits before any reference reads or other rule probes.

## Spec / tracking

Addresses **issue #81 item A3** ("`delins` with empty inserted sequence must render as `del` (or be rejected at parse). Edge case; needs a confirming test.").

The HGVS v21.0 spec fixture added in #84 / PR #105 has zero rows whose `input` or `current` is a bare empty-`delins` — the spec doesn't list this as a canonical example — so coverage belongs in `tests/normalize_tests.rs` as a synthetic-input regression, exactly as #81's "How to use this issue" prescribes. Verified the fixture is byte-identical post-fix (`generate_spec_fixture --check` passes).

## Test plan

- [x] Rust: two new unit tests in `src/normalize/mod.rs` covering single-position (`c.10delins` → `c.10del`) and multi-position with 3'-shift (`c.10_11delins` → `c.11_12del`, since `ref[10]=G == ref[12]=G`)
- [x] Python: matching integration tests in `tests/python/test_core.py::TestNormalizeEmptyInsertDelinsToDel`
- [x] Red-green-revert: confirmed both Rust tests fail without the fix and pass with it
- [x] Full Rust test suite: 3400 tests pass
- [x] All Python tests (uv run poe check-tests): 121 passed
- [x] cargo fmt --all -- --check: clean
- [x] cargo clippy --features dev --lib --tests --examples -- -D warnings: clean
- [x] uv run poe check-lint / check-format / check-typing: clean
- [x] Spec fixture byte-identical: `generate_spec_fixture --check` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - *(normalize)* rewrite degenerate substitutions (ref == alt, e.g. `c.100A>A`) to identity (`=`) per HGVS v21 spec, which marks `c.X>X` as "not allowed" (`docs/recommendations/DNA/other.md`). The rule is purely syntactic on the edit's stated bases, so it fires in both the full-normalization path and the no-reference canonicalization path — `c.123C>C` rewrites to `c.123=` regardless of provider availability. ([#81](https://github.com/fulcrumgenomics/ferro-hgvs/issues/81) A4)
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #114, opened 2026-05-06
test(normalize): pin RNA path + edge cases for A9 substitution-after-trim

nh13:
## Summary

#109 introduced the unified `canonicalize_delins` / `DelinsCanonical` pipeline. Its step 4c (substitution after shared-affix trimming) already implements the #81 A9 behavior, so no production-code change is required to satisfy A9. This PR is now purely additive test coverage on top of that pipeline.

## What changed

`src/normalize/mod.rs` — extend `test_normalize_delins_different_bases_becomes_substitution` with an r. twin so the RNA path through `canonicalize_delins` is pinned end-to-end, not just the c. path:

- `NM_000088.3:r.1_3delinsacg` → `NM_000088.3:r.2t>c`

The locked output is `r.2t>c`, not the spec-ideal `r.2u>c`. `NaEdit::Substitution`'s Display lowercases the reference byte but does not map T → u for r. variants synthesized inside the normalizer. This is a pre-existing display-layer behavior, not in scope here; the test comment cross-references `src/hgvs/edit.rs` for any follow-up.

`src/normalize/rules.rs` — add three byte-level asserts to `test_canonicalize_delins`. Existing cases cover prefix-only-trim and suffix-only-trim sub paths; these fill in the gaps:

1. Both-side trim leaving exactly 1 residual (`NACGN[1..4]` / `ATG` → C>T at idx 2). Distinct control flow from the prefix-only and suffix-only cases.
2. Mutalyzer A9 canonical example (`NGTAN[1..4]` / `GTT` → A>T at idx 3) — byte-level analogue of `c.5948_5950delinsGTT over GTA → c.5950A>T`.
3. Homopolymer residual greedy 3'-tie-breaking (`AAAA` / `AAAT` → A>T at idx 3). Multiple positions are semantically equivalent; pins greedy prefix-trim's tie-break to the 3'-most index.

## What was dropped

This branch originally added a standalone `delins_collapse_to_substitution` predicate plus a corresponding branch in the delins cascade in `mod.rs`. After rebasing onto main, both became redundant with #109's `canonicalize_delins`. Main's design is the more principled long-term solution — single-pass shared-affix trim, exhaustive enum dispatch, priority order encoded in one place — so this PR yields the implementation to main and keeps only the additive coverage above.

## Test plan

- [x] `cargo nextest run --features dev` (3537 passed)
- [x] `cargo clippy --features dev --all-targets -- -D warnings`
- [x] `cargo fmt --all -- --check`

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -2565,6 +2565,17 @@ mod tests {
         let variant = parse_hgvs("NM_000088.3:c.1_3delinsACG").unwrap();
         let result = normalizer.normalize(&variant).unwrap();
         assert_eq!(format!("{}", result), "NM_000088.3:c.2T>C");
+
+        // r. twin: same collapse on the RNA path. NM_000088.3 has cds_start = 1,
+        // so r.1_3 maps to the same ATG run and the residual is at r.2.
+        // The substitution edit Display lowercases the reference byte but does
+        // not map T -> u for r. variants synthesized by canonicalize_delins,
+        // so the locked form is `r.2t>c` rather than the spec-ideal `r.2u>c`.
+        // RNA U/T mapping for synthesized substitutions is a separate display-
+        // layer concern (see `NaEdit::Substitution` in src/hgvs/edit.rs).
+        let variant = parse_hgvs("NM_000088.3:r.1_3delinsacg").unwrap();
+        let result = normalizer.normalize(&variant).unwrap();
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #130, opened 2026-05-06
test(protein): pin p.0 no-product round-trip and adjacent guards (#81 D6)

nh13:
## Summary

Closes item **D6** of #81. Lock the parse / Display / re-parse / normalize contract for `p.0` and `p.0?` (HGVS no-protein notation for null alleles) and pin the surrounding spec invariants the first pass missed.

A single integration test file `tests/protein_no_protein_roundtrip.rs` exercises the cross product of `{p.0, p.0?} × {NP_xxx.x:, LRG_xxxpN:}` and additionally pins, in dedicated tests:

1. **Round-trip core** — parse, Display, re-parse, edit-shape (`ProteinEdit::NoProtein { predicted }`), normalize-noop and one-round idempotency. (Carried over from the first revision.)
2. **Internal shape** — outer `Mu` wrapper is `Mu::Certain` for both forms (the `?` is part of the inner edit, never an outer `Mu::Uncertain`), and the dummy location is the canonical `Met1` point interval the parser builds at `src/hgvs/parser/variant.rs:1735-1736`.
3. **Multi-pass normalize idempotency** — three passes plus `Hash` agreement across every pass (relevant for the normalizer's caches and any downstream dedup tables; `Hash` is derived but a derive-blocker change wouldn't be caught by `Eq`-only assertions).
4. **`Hash` distinguishes `p.0` from `p.0?`** independently of `Eq`.
5. **Cross-form non-equivalence (D5 / D7 guards)** — `p.0` and `p.0?` are structurally and Display-distinct from `p.=`, `p.(=)`, `p.?`, and `p.(?)`. Section D's whole point is that each row is independently enforceable; without these guards, a regression that collapsed `NoProtein` into `Identity` or `Unknown` would slip through D6's net.
6. **Malformed-adjacent rejection** — `p.0X`, `p.0a`, `p.0!`, `p.00`, `p.(0)`, `p.(0?)`, `p. 0` (interior whitespace), bare `p.0` / `p.0?` (no accession), and cis-bracket forms `p.[0]`, `p.[Arg97Trp;0]`, `p.[0;Arg97Trp]` all reject. Spec only sanctions `[0]` in trans context, never cis.
7. **Outer whitespace** is trimmed by `parse_hgvs`; interior whitespace is not.
8. **`parse_hgvs_fast` parity** — the fast-path entry produces structurally identical results on D6 inputs.
9. **Trans-allele compound (D6 in context)** — the expanded form `[ACC:p.X];[ACC:p.0]` and `[ACC:p.X];[ACC:p.0?]` carries `ProteinEdit::NoProtein { predicted }` on the second arm with `phase: Trans`.
10. **Spec-critical `[0]` vs `p.0` distinction** — bare `[0]` in `p.[X];[0]` (alleles.md line 78) is the cross-coordinate "absent allele" marker (`HgvsVariant::NullAllele`), **not** a protein-coordinate `ProteinEdit::NoProtein`. They are semantically distinct and a refactor that folded them together would corrupt downstream interpretation (a hemizygous male's "absent second X-chromosome allele" would be miscoded as "no protein from this allele due to a Met1 variant", which is biologically wrong).

No production changes — every assertion holds against today's parser, `Display`, and protein-normalizer pass-through. This is purely regression-pinning, scoped to D6.

## HGVS spec rationale

Both inputs come straight from the HGVS spec (vendored under `assets/hgvs-nomenclature/docs/recommendations/protein/`):

- `substitution.md` line 42 defines the certain form: "no protein: `LRG_199p1:p.0`: as a consequence of a variant in the translation initiation codon no protein is produced … `LRG_199p1:p.0?` can be used when you predict that no protein is produced. Do not use descriptions like `p.Met1Thr`, this is for sure not the consequence of the effect on protein translation."
- `deletion.md` line 45 uses the predicted form for a real null-allele example: "since exon 1 and the translation initiation codon are deleted no reliable predictions can be made. Possibly no transcript is generated and no protein (`p.0?`)".
- `alleles.md` line 78 shows `[0]` inside an allele description (`p.[Ser86Arg];[0]`) for the absent allele on the male X chromosome — distinct from `[ACC:p.0]` and motivating the prefix-vs-bare-bracket distinction tested in (10).

`p.0` and `p.0?` are distinct from `p.?` (unknown effect — Section D7) and `p.=` (silent — Section D5); pinning them separately keeps each row of Section D enforceable as an independent guard.

## Out of scope

- Compact-form trans-allele Display→reparse drift (`[ACC:p.X];[ACC:p.0]` Display-emits `ACC:p.[X];[0]`, where the bare-bracket `[0]` re-parses to `NullAllele` rather than the original `ProteinEdit::NoProtein`). This is a real Display/parser interaction worth investigating, but it concerns the bracketed-allele compact form, not D6 — D6 is about the standalone whole-protein no-product notation. No follow-up issue filed; surface to the team if/when it bites.
- Cis-bracket form `p.[X;0]`. Spec doesn't allow it; current reject is correct and pinned in (6).

## Test plan

- [x] `cargo nextest run --features dev` — full suite green (3411 passed, 51 skipped, 0 failed; +9 over the prior revision).
- [x] `cargo nextest run --features dev --test protein_no_protein_roundtrip` — 13 tests pass (4 from the prior revision + 9 new).
- [x] `cargo nextest run --features dev --test hgvs_spec_normalization_tests` — green (existing harness already covers the v21.0 corpus rows for `LRG_199p1:p.0`, `LRG_199p1:p.0?`, `NP_003997.1:p.0?` with `status: "preserved"`).
- [x] Sanity-inversion: each new test's primary assertion was inverted locally to confirm it fails when the assertion is wrong, then reverted before the squashed commit.

### DIFF
--- a/tests/protein_no_protein_roundtrip.rs
+++ b/tests/protein_no_protein_roundtrip.rs
@@ -1,0 +1,625 @@
+//! Round-trip pinning for protein "no product" notation: `p.0` and `p.0?` (#81 D6).
+//!
+//! HGVS spec context (assets/hgvs-nomenclature/docs/recommendations/protein/):
+//!
+//! - `p.0` — no protein is produced (definite), used for null alleles
+//!   (substitution.md: "as a consequence of a variant in the translation
+//!   initiation codon no protein is produced").
+//! - `p.0?` — predicted no-protein (deletion.md: "Possibly no transcript is
+//!   generated and no protein (`p.0?`)").
+//!
+//! These are distinct from `p.?` (unknown effect — D7) and `p.=` (silent — D5).
+//! Section D of issue #81 already has parse-only coverage for these forms; this
+//! file pins the full parse / Display / re-parse / normalize round-trip so that
+//! any future change touching the protein edit pipeline must keep the certain
+//! vs. predicted distinction intact.
+//!
+//! For each canonical input, we assert:
+//!
+//! 1. `parse_hgvs(s)` succeeds and produces a `Protein` variant.
+//! 2. `format!("{}", parsed) == s` (Display is the spec form).
+//! 3. The parsed edit is `ProteinEdit::NoProtein { predicted }` with the
+//!    expected `predicted` flag (so `p.0` and `p.0?` are not collapsed).
+//! 4. `parse_hgvs(format!("{}", parsed)) == parsed` (re-parse stability,
+//!    structural equality).
+//! 5. `normalize(parsed)` is a no-op on these whole-protein null forms and
+//!    is idempotent.
+
+use ferro_hgvs::hgvs::edit::ProteinEdit;
+use ferro_hgvs::hgvs::variant::ProteinVariant;
+use ferro_hgvs::{parse_hgvs, parse_hgvs_fast, HgvsVariant, MockProvider, Normalizer};
+
+/// One canonical no-protein input plus the `predicted` flag we expect after parsing.
+struct NoProteinCase {
+    input: &'static str,
+    predicted: bool,
+}
+
+/// Canonical inputs covering the cross product of {certain, predicted} and
+/// {`NP_` accession, `LRG_` accession}. Both accession styles appear verbatim
+/// in the HGVS spec for `p.0` (substitution.md uses `LRG_199p1:p.0`,
+/// deletion.md uses `NP_003997.2:p.0?`).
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #136, opened 2026-05-06
test(protein): pin p.? unknown-effect round-trip across allele forms and edge cases (#81 D7)

nh13:
Closes item **D7** of #81. Lock down the parse / Display / re-parse / normalize / Hash / cross-form / reject contract for the HGVS protein unknown-effect (`?`) family at the same depth as the D5 (`p.=`) and D6 (`p.0`) siblings. No production-code changes — D7 is purely regression-pinning.

## What this pins

A single integration-test file (`tests/protein_unknown_roundtrip.rs`, 20 tests) covering:

1. **Round-trip core** — `parse → Display → parse` is the identity for every canonical form (`p.?`, `p.(?)`, `p.Met1?`, `p.(Met1?)`); `normalize` is a no-op; one normalize+display+reparse cycle still lands on the same AST. Carried over from the prior revision.
2. **Edit-shape pin** — every canonical form yields `ProteinEdit::Unknown { predicted, whole_protein }` with the expected flags. The inner `predicted` bit equals `case.predicted XOR case.outer_uncertain` (so a refactor that flattens any pair fails before round-trip notices).
3. **Internal `Mu` shape (asymmetric, intentional)** — `p.?` and `p.(?)` keep `Mu::Certain` outer with the `predicted` bit on the inner edit. `p.Met1?` keeps `Mu::Certain` with `predicted: false`. `p.(Met1?)` is asymmetric: the predicted bit lives on the **outer** `Mu::Uncertain`, not the inner edit. Pinning this stops a refactor from silently aligning the two predicted paths in either direction. Mirrors D5's exact finding for `p.=` / `p.(=)` / `p.Leu54=` / `p.(Leu54=)`. Outer `Mu` must NEVER be `Mu::Unknown` for protein edits (that variant is reserved for `NaEdit` `c.?` / `r.?`).
4. **Dummy `Met1` interval** — every canonical form carries the canonical `ProtPos::new(AminoAcid::Met, 1)` point interval. Whole-protein forms get it via the parser fill (`variant.rs:1749-1750`); position-specific forms get it from the literal `Met1` input. Same interval, different parser paths.
5. **Multi-pass `Normalizer::normalize` idempotency (3 passes)** with `Hash` agreement across every pass (catches `Hash`/`Eq` drift independent of `Eq`; relevant for normalizer caches and downstream `HashMap`/`HashSet` dedup).
6. **`Hash` distinguishes the four canonical shapes** — `p.?`, `p.(?)`, `p.Met1?`, `p.(Met1?)` all hash distinctly.
7. **Cross-D-row non-equivalence (D5 / D6 guards)** — every D7 form is structurally, Display-, and Hash-distinct from every D5 silent form (`p.=`, `p.(=)`, `p.Leu54=`, `p.(Leu54=)`) and every D6 no-protein form (`p.0`, `p.0?`). Without these, a regression that collapsed `Unknown` onto `Identity` or `NoProtein` would slip past D7's existing round-trip.
8. **Predicted ↔ certain non-equivalence** with Display distinctness for both whole-protein and position-specific pairs.
9. **Whole-protein ↔ position-specific non-equivalence** with Display distinctness for both certain and predicted pairs.
10. **`parse_hgvs_fast` parity** — produces structurally identical ASTs (and identical `Hash`) for every D7 input.
11. **Outer / trailing whitespace tolerance** — `parse_hgvs` swallows leading + trailing whitespace; pinned so a future tightening doesn't silently regress (matches D5 / D6).
12. **Malformed-adjacent rejection** — trailing junk (`p.??`, `p.?X`, `p.?Met1`, `p.(?)?`), interior whitespace (`p. ?`, `p. (?)`, `p.Met1 ?`), unclosed parens (`p.(?`, `p.(Met1?`), position 0 (`p.Met0?`), bare (no accession) `p.?` / `p.Met1?`, bracket forms `p.[?]` / `p.[?;Arg97Trp]` / `p.[Arg97Trp;?]` (spec only sanctions `[?]` as the **trans** whole-allele marker, never inside the protein bracket-shorthand path), and empty `p.`.
13. **Display canonicalisation** — `p.M1?` (single-letter) and `p.met1?` (lowercase) parse to the same AST as `p.Met1?` and Display-canonicalise to the spec's 3-letter form. Same for the predicted variant.

### `[?]` in trans context is `HgvsVariant::UnknownAllele`, NOT `ProteinEdit::Unknown`

The spec's canonical heterozygous example (`alleles.md:43`):

```
NP_003997.1:p.[(Ser68Arg)];[?]
```

is parsed as a 2-arm trans allele where arm 0 is a predicted protein substitution and arm 1 is `HgvsVariant::UnknownAllele`. They are semantically distinct: a refactor that folded `[?]` and `p.(?)` together would miscode "expected but not yet identified" recessive-disease reporting (the spec's stated meaning of `[?]`) as "this allele has the protein-level unknown variant `p.(?)`". Mirrors D6's `[0]` vs `p.0` finding.

### Trans compact-form drift (mirrors D5)

The same Display→reparse drift D5 surfaced affects D7's position-specific arms. Two cases, two outcomes:

- **Whole-protein unknown arm** (`[NP:p.X];[NP:p.?]` and `…;[NP:p.(?)]`) — `is_whole_protein_unknown()` triggers `is_loc_edit_unknown()` (`variant.rs:701`) → compact-form **suppressed** → Display preserves the expanded form → re-parse equals the original. Pinned as a clean round-trip.
- **Position-specific unknown arm** (`[NP:p.X];[NP:p.Met1?]` and `…;[NP:p.(Met1?)]`) — `whole_protein: false` so suppression does NOT trigger → Display emits `ACC:p.[X];[Met1?]` → today's parser cannot re-parse the same-accession compact trans form (issue #83). Pinned as `is_err()` with a comment to flip when #83 lands.

This is the same drift D5 already pinned; we re-pin it from D7's angle so a fix to #83 must update this test too.

## HGVS spec rationale

Per `assets/hgvs-nomenclature/docs/recommendations/protein/`:

- `substitution.md:43` — "unknown: `LRG_199p1:p.(Met1?)`: the consequence, at the protein level, of a variant affecting the translation initiation codon can not be predicted (i.e. is unknown)".
- `substitution.md:50` — `NP_003997.1:p.?`: "the predicted consequence of variant `NM_004006.2c.2622G>C` is a silent change ... it can not be excluded the variant affects splicing, having unknown consequences."
- `deletion.md:45-46` and `duplication.md:47-48` — `NP_003997.2:p.?` for whole-gene unknown-effect deletions / duplications.
- `alleles.md:43` — `NP_003997.1:p.[(Ser68Arg)];[?]` for the spec's canonical heterozygous "one variant identified, the other expected but not yet identified" example.

## Spec-corpus cross-link

`tests/fixtures/grammar/hgvs_spec_normalization.json` already covers `p.?`, `p.(?)`, `LRG_199p1:p.(Met1?)`, `NP_003997.1:p.?`, and `NP_003997.2:p.?` with `status: "preserved"`. We do not duplicate the v21.0 corpus rows here; the file's module doc cross-references it.

## Test plan

- [x] `cargo nextest run --features dev --test protein_unknown_roundtrip` — 20 / 20 pass (up from 7 in the prior revision).
- [x] `cargo nextest run --features dev` — full suite green: 3418 passed (51 skipped).
- [x] Sanity-inversion: each new assertion family was probed locally (flipping `outer_uncertain` `false ↔ true` in the `Mu`-shape pin, swapping `UnknownAllele` for `Protein` in the `[?]`-marker pin, and flipping `is_err()` to `is_ok()` in the trans-compact-drift gate — all fail when wrong) and reverted before the squashed commit.

## Caveats / out of scope

- The trans-compact-form Display→reparse drift is the same one D5 already surfaced (issue #83); no new follow-up filed. We re-pin it from D7's angle.
- D5, D6, D8 are not touched.
- Commit is unsigned: 1Password GPG signing failed three times during the squashed commit; please re-sign on next amend when biometric approval is available.

### DIFF
--- a/tests/protein_unknown_roundtrip.rs
+++ b/tests/protein_unknown_roundtrip.rs
@@ -1,0 +1,799 @@
+//! Round-trip pinning for protein unknown-effect notation (`p.?` family) — issue #81 D7.
+//!
+//! HGVS spec context (`assets/hgvs-nomenclature/docs/recommendations/protein/`):
+//!
+//! - `p.?` — translation product is unknown (whole-protein, certain).
+//! - `p.(?)` — predicted whole-protein unknown.
+//! - `p.Met1?` — initiation Met affected, effect unknown (position-specific,
+//!   certain).
+//! - `p.(Met1?)` — predicted unknown effect on initiation Met. This is the
+//!   form the spec uses in `substitution.md:43`: "the consequence, at the
+//!   protein level, of a variant affecting the translation initiation codon
+//!   can not be predicted (i.e. is unknown)".
+//!
+//! Distinct from `p.0` / `p.0?` (no protein produced — D6) and from
+//! `p.=` / `p.(=)` / `p.Leu54=` / `p.(Leu54=)` (silent — D5). Each row of
+//! Section D is independently enforceable; this file pins every D7 invariant
+//! the existing parse / Display / normalize stack provides today, so a future
+//! refactor that collapses any pair (predicted↔certain, whole↔position,
+//! unknown↔no-protein, unknown↔silent) fails loudly here.
+//!
+//! Spec corpus cross-link: `tests/fixtures/grammar/hgvs_spec_normalization.json`
+//! already covers `p.?`, `p.(?)`, `LRG_199p1:p.(Met1?)`, `NP_003997.1:p.?`,
+//! and `NP_003997.2:p.?` with `status: "preserved"`. This file does not
+//! duplicate the v21.0 corpus rows — it pins behavior the corpus harness
+//! cannot (Mu shape, Hash agreement, cross-form non-equivalence, the
+//! `[?]`-vs-`p.(?)` distinction, the trans compact-form drift gate).
+//!
+//! Sibling pinners: `tests/protein_silent_eq.rs` (D5) and
+//! `tests/protein_no_protein_roundtrip.rs` (D6).
+
+use std::collections::hash_map::DefaultHasher;
+use std::hash::{Hash, Hasher};
+
+use ferro_hgvs::hgvs::edit::ProteinEdit;
+use ferro_hgvs::hgvs::interval::ProtInterval;
+use ferro_hgvs::hgvs::location::{AminoAcid, ProtPos};
+use ferro_hgvs::hgvs::uncertainty::Mu;
+use ferro_hgvs::hgvs::variant::{AllelePhase, ProteinVariant};
+use ferro_hgvs::{parse_hgvs, parse_hgvs_fast, HgvsVariant, MockProvider, Normalizer};
+
+/// One canonical D7 input plus the structural shape we expect after parse.
+///
+/// `outer_uncertain`: true iff the parser wraps the inner `ProteinEdit` in
+/// `Mu::Uncertain`. Currently only `p.(Met1?)` — the position-specific
+/// predicted form — takes that path; the whole-protein `p.(?)` keeps the
+/// `Mu::Certain` outer with `predicted: true` on the inner edit. This
+/// asymmetry mirrors the D5 silent-`=` family and is intentional.
+struct UnknownCase {
+    /// HGVS input (always with an accession prefix).
+    input: &'static str,
+    /// True for whole-protein (`p.?`, `p.(?)`); false for position-specific.
+    whole_protein: bool,
+    /// True for predicted (`p.(?)`, `p.(Met1?)`); false for certain.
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #154, opened 2026-05-07
fix(parser): preserve explicit deleted sequence in delins round-trip

### LINKED ISSUE #120: parser: delins with explicit deleted sequence drops the deletion (#81 K1 follow-up)

## Problem

The HGVS spec allows an explicit deleted sequence inside a delins, e.g. `g.100_102delATGinsTTCC`. ferro's parser accepts the input but immediately drops the deleted sequence: the resulting `NaEdit::Delins` only carries `sequence: TTCC`, and rendering produces `g.100_102delinsTTCC` (the `ATG` is gone).

## Reproduce

```rust
let v = parse_hgvs("NC_000001.11:g.100_102delATGinsTTCC").unwrap();
assert_eq!(v.to_string(), "NC_000001.11:g.100_102delinsTTCC"); // current
// ATG is no longer recoverable from the parsed value
```

## Scope

Two viable directions:

1. Preserve the deleted sequence on `NaEdit::Delins` (e.g. `Delins { deleted: Option<Sequence>, inserted: InsertedSequence }`). This adds a field but lets HGVS→SPDI succeed without a reference provider when the deleted sequence is given.
2. Document the spec's preferred form (delins without explicit deleted sequence) and treat the explicit form as a parse-time consistency check only.

Option 1 is the more user-friendly choice and aligns with the way `NaEdit::Deletion { sequence: Option<Sequence> }` already preserves the deleted sequence.

## Audit pin

`tests/hgvs_spdi_roundtrip_audit.rs::audit_delins_with_explicit_deletion_drops_seq` pins the current behavior.

nh13:
Closes #120. Follow-up from #81 K1 audit.

## Summary

- Extended `NaEdit::Delins` with `deleted: Option<Sequence>` and `deleted_length: Option<u64>` fields so the parser can preserve the explicit deleted sequence/count from notations like `delATGinsTTCC` and `del3insTA`. Previously these were silently discarded, so `c.100_102delATGinsTTCC` parsed and displayed as `c.100_102delinsTTCC`.
- Updated parser sites: `parse_delins` (short form) sets both `None`; `parse_delins_with_deleted_seq` populates `deleted`; `parse_delins_with_deleted_count` populates `deleted_length`. The multi-base substitution path (`GG>G` -> delins) keeps both `None` because the `>`-form does not carry an explicit deleted sequence.
- Updated `Display` and `to_rna_string` to emit `del<XX>ins<YY>` only when those fields are set; short-form output is unchanged for short-form input. Synthesis-not-allowed: ferro never invents a `delXXX` part if the user did not supply one.
- `canonicalize_edit` continues to strip the explicit form back to `delins<YY>` per the HGVS spec recommendation that the short form is preferred. The new tests pin both directions.
- All other call sites (SPDI/VCF conversion, normalization, validation, python helpers, service handlers) updated to the new shape: construction passes `None`, destructuring uses `..` to ignore the new fields. Behavior is preserved at those sites.
- Opportunistic SPDI improvement: when the input carries an explicit deleted sequence, `hgvs_to_spdi_simple` can now succeed without consulting a reference provider (`NC_000001.11:g.100_102delATGinsTTCC` -> `NC_000001.11:99:ATG:TTCC`).
- Regenerated `tests/fixtures/grammar/hgvs_spec_normalization.json`: 4 rows transitioned from `diverges` to `preserved` (the previous \"current\" pinned the bug). The fixture is byte-identical under `cargo run --features dev --example generate_spec_fixture -- --check`.
- Added unit tests for parser preservation (explicit seq, explicit count, short form unchanged, multi-base substitution untouched), Display, RNA lowercase Display, equality (explicit vs short form are not equal), full HGVS round-trip across `g.`/`c.`/`n.`/`m.` coordinate systems, parse-display-reparse idempotency, and `canonicalize_edit` strips both forms.
- Strengthened `tests/comprehensive_edge_case_tests.rs` to assert each fixture entry under `delins_with_explicit_deleted` round-trips its input verbatim.

## HGVS spec rationale

`assets/hgvs-nomenclature/docs/recommendations/DNA/delins.md` allows the explicit form `<pos>del<XX>ins<YY>` while the spec actively recommends the short form `<pos>delins<YY>` ("the recommendation is not to describe the variant as `NC_000023.11:g.32386323delTinsGA`, i.e. describe the deleted nucleotide sequence. This description is longer, it contains redundant information and chances to make an error increase"). That note tells implementations how to recommend writing variants but does not say tools should silently discard user input. ferro now treats explicit-form input as carrying meaningful round-trip information: parse preserves it, Display emits it back, and `canonicalize_edit` is the explicit way to drop it.

## Test plan

- [x] `cargo nextest run --features dev` - 3414 tests pass (full suite)
- [x] `cargo clippy --features dev --all-targets -- -D warnings` - clean
- [x] `cargo fmt --all -- --check` - clean
- [x] `cargo run --features dev --example generate_spec_fixture -- --check` - byte-identical

### DIFF
--- a/src/hgvs/edit.rs
+++ b/src/hgvs/edit.rs
@@ -446,8 +446,21 @@ pub enum NaEdit {
     /// Insertion: addition of bases between two positions (e.g., insATG, ins10, insA[10])
     Insertion { sequence: InsertedSequence },
 
-    /// Deletion-insertion: replacement of bases (e.g., delinsATG, delinsA[10])
-    Delins { sequence: InsertedSequence },
+    /// Deletion-insertion: replacement of bases (e.g., delinsATG, delinsA[10],
+    /// delATGinsTTCC, del3insTA).
+    ///
+    /// `deleted` and `deleted_length` preserve an explicit deleted sequence or
+    /// length when the input notation provided one (e.g., `delATGinsTTCC`
+    /// parses to `deleted: Some("ATG")`). At most one of `deleted` /
+    /// `deleted_length` is `Some`; both are `None` for the recommended short
+    /// form `delinsXXX`. `Display` preserves on round-trip; `canonicalize_edit`
+    /// strips them back to the short form per the HGVS spec recommendation
+    /// (`recommendations/DNA/delins.md`).
+    Delins {
+        sequence: InsertedSequence,
+        deleted: Option<Sequence>,
+        deleted_length: Option<u64>,
+    },
 
     /// Duplication: copy of bases (e.g., dup, dupATG, dup101, dup(731_741), dup?)
     Duplication {
@@ -629,8 +642,20 @@ impl NaEdit {
             NaEdit::Insertion { sequence } => {
                 format!("ins{}", sequence.to_rna_string())
             }
-            NaEdit::Delins { sequence } => {
-                format!("delins{}", sequence.to_rna_string())
+            NaEdit::Delins {
+                sequence,
+                deleted,
+                deleted_length,
+            } => {
+                let mut s = String::from("del");
+                if let Some(seq) = deleted {
+                    s.push_str(&seq.to_lowercase_string());
+                } else if let Some(len) = deleted_length {
+                    s.push_str(&len.to_string());
+                }
+                s.push_str("ins");
+                s.push_str(&sequence.to_rna_string());
+                s
             }
             NaEdit::Duplication {
                 sequence,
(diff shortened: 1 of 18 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #155, opened 2026-05-07
fix(normalize): apply 3' rule across phase-mismatched cyclic rotations for ins (#132)

### LINKED ISSUE #132: 3' rule: cyclic-rotation insertion misses repeat tract on phase mismatch (#81 B4 follow-up)
Discovered while pinning #81 B4 (3' rule for repeat unit boundary placement) in `tests/repeat_3prime_boundary.rs`.

## Problem

When a literal insertion sequence is a cyclic rotation of an adjacent reference repeat unit, the 3' rule should shift the insertion to the most-3' equivalent position. Today, ferro misses the rotation when the shuffle stops at the first base on a phase mismatch.

## Repro

Reference layout `AAGTGTGTTTTTTTAA` — `GT[3]` tract at positions 3..8 (0-based 2..8) followed by `T[6]`. Padded contig from `tests/common/synthetic.rs`. The `GT[3]` tract starts with `G`.

`g.258_259insTG` (insertion at 5' edge, just before the tract's first `G`):

| Input | Current ferro output | Expected (per 3' rule) |
|---|---|---|
| `g.258_259insTG` | `g.258_259insTG` (no shift) | `g.263_264dup` (most-3' GT in tract) |
| `g.258_259insGT` | `g.258_259insGT` (no shift) | `g.263_264dup` |

By contrast, when the same logical edit is provided as a 2-copy insertion, ferro does the right thing:

| Input | Current ferro output |
|---|---|
| `g.258_259insTGTG` | `g.259_264GT[5]` (correct) |
| `g.258_259insGTGT` | `g.259_264GT[5]` (correct) |

Same shape with `AAACAGCAGCAGCAACAGAA` (CAG[3]CAA[1]CAG[1]):

| Input | Current ferro output | Expected |
|---|---|---|
| `g.258_259insAGC` | `g.258_259insAGC` | `g.265_267dup` |
| `g.258_259insGCA` | `g.258_259insGCA` | `g.265_267dup` |

## Root cause sketch

`src/normalize/shuffle.rs::shuffle()` advances one base at a time and stops the moment `ref_seq[ref_idx] != alt_seq[alt_idx]`. For an insertion at the 5' edge of a tract, when the inserted sequence is a non-zero rotation of the tract's unit, alt[0] does not match the first ref base and shuffle never starts. `rules::insertion_to_repeat()` does iterate rotations and would find the tract, but it only fires for `added_copies >= 2`; the single-copy path (`insertion_is_duplication`) operates on `result.start` (= unmoved start) and only checks the unrotated literal sequence at positions immediately adjacent to the unmoved insertion point. The cyclic-rotation single-copy case falls through to plain `insertion_is_duplication`'s window check and returns false.

## Suggested fix shape

Single-copy insertion path needs a rotation iterator analogous to `insertion_to_repeat`'s `for r in 0..u_len` loop: if any rotation of the inserted sequence matches an adjacent tract unit, snap to the most-3' duplication position of that tract. This also makes the single-copy and multi-copy paths symmetric and removes the surprising shuffle/repeat asymmetry exposed by `g.258_259insTG` vs `g.258_259insTGTG`.

## Tests pinning the current behavior

`tests/repeat_3prime_boundary.rs::cyclic_rotation_insertion_misses_3prime_shift` (added in #81 B4) locks the current outputs with `// TODO` markers pointing here. Update those expectations once the fix lands.

nh13:
Closes #132. Follow-up from #81 B4 / PR #93 (ins 3'-shift coverage matrix).

## Summary

- Single-copy insertions whose literal alt is a cyclic rotation of an adjacent reference repeat unit (e.g. `g.X_(X+1)insTG` against a `GT[3]` tract) previously fell through to plain `ins` at the unmoved position. The 3' shuffle's first-base check (`alt[0] == ref[ins_point]`) failed before the walk could start, and the post-shuffle `insertion_is_duplication` helper only checked the unrotated alt at the unmoved position. The 2+-copy path already handled this via `insertion_to_repeat`'s rotation iteration; the 1-copy path did not.
- New helper `insertion_to_duplication` in `src/normalize/rules.rs` mirrors `insertion_to_repeat`'s rotation-iteration loop for the 1-copy case. It tries every rotation of `smallest_repeat_unit(inserted_seq)`, picks the rotation yielding the largest tandem run via `find_tandem_extent`, and returns the most-3' unit-aligned dup position inside that tract.
- The Insertion arm of `normalize_na_edit` (`src/normalize/mod.rs`) now consults the new helper post-shuffle, before the existing `insertion_is_duplication` rotation-only check. Output is symmetric with `insertion_to_repeat` for the 2+-copy path: the `g.258_259insTG` vs `g.258_259insTGTG` asymmetry exposed by the issue is resolved.
- Test coverage: 6 unit tests in `src/normalize/rules.rs` covering homopolymer matched, two-base rotation, no-tract negative, empty/oob, multi-copy rejection, and phase-matched sanity. End-to-end reproducer file `tests/issue_132_cyclic_rotation_insertion.rs` runs the issue's exact inputs and asserts idempotency. New `cyclic_rotation_ins_one_unit` rstest in `tests/ins_shift_matrix.rs` adds cyclic-rotation cells to all 7 modules (genomic, cds_plus, cds_minus, noncoding_plus, noncoding_minus, rna_plus, rna_minus). RNA gets one cell instead of two because units containing `u` map to byte `U` in the alt but byte `T` in the (DNA-stored) reference, so the rotation iterator legitimately can't match — the GCA-rotation case exercises the path without that confound.
- Two pre-existing `tests/normalize_tests.rs` cases relied on the asymmetric old behavior. `test_insertion_sequence_rotation` (lenient `assert!(contains("ins"))`) flipped to a `dup` because the new helper's rotation iterator finds an AACA-phase tract with `ref_count=1` adjacent to the insertion — the dup form is more spec-correct per the dup-over-ins priority. `test_insertion_to_dup_with_rotation` flipped from `c.262_264dup` (the GCG-phase position the old shuffle-walk happened to land on) to `c.261_263dup` (the GGC-phase position symmetric with the multi-copy path's choice). Both flipped expectations describe the same edit; the new positions are phase-consistent with `insertion_to_repeat`.

## HGVS spec rationale

- `assets/hgvs-nomenclature/docs/recommendations/DNA/repeated.md`: \"applying the **3'rule**, the repeat has to be described as an AGC repeat\" — the spec explicitly notes that rotation-equivalent unit choices are resolved by the 3' rule. The new helper applies this for single-copy insertions.
- Same file's exception: \"use NM_024312.4:c.2692_2693dup and **not** NM_024312.4:c.2686A[10]\" — single-copy unit additions stay as `dup`, not repeat notation. The new helper emits `dup` (not `Repeat`), preserving this rule.
- `assets/hgvs-nomenclature/docs/background/numbering.md` 3' rule: variants must shift to the most-3' equivalent position when ambiguous. The new helper picks `tract_end - unit.len()` as the dup start, satisfying the rule.

## Test plan

- [x] cargo nextest run --features dev — full suite green (3422/3422)
- [x] cargo clippy --features dev --all-targets -- -D warnings — clean
- [x] cargo fmt --all -- --check — clean


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #156, opened 2026-05-07
fix(spdi): recover dup form on SPDI→HGVS for duplicated insertions

### LINKED ISSUE #119: spdi: SPDI→HGVS does not recover dup form for duplicated insertions (#81 K1 follow-up)

## Problem

Round-tripping a duplication HGVS through SPDI silently drops the `dup` form and re-emits as `ins`:

```
g.100_102dupATG  →  NC_000001.11:101::ATG  →  g.102_103insATG
```

SPDI has no native `dup` representation — duplications are always insertions of the duplicated sequence. The forward direction (HGVS→SPDI) is correct. The reverse (SPDI→HGVS) currently chooses `ins`, but the spec prefers `dup` whenever the inserted sequence equals the immediately-5' reference of the same length (see #81 item A1).

## Scope

When `spdi_to_hgvs` produces an insertion, check whether the inserted sequence equals the reference immediately 5' of the SPDI position. If so, emit `dup`. Requires a `ReferenceProvider` (couples with #117).

Without a reference, `ins` is the safe choice and current behavior is acceptable. The audit test pins this so we notice if an attempt to "fix" the round-trip changes the no-reference behavior.

## Audit pin

`tests/hgvs_spdi_roundtrip_audit.rs::audit_duplication_roundtrip_renders_as_insertion` pins the no-reference behavior.

nh13:
Closes #119. Follow-up from #81 K1 audit.

## Summary

- Adds `spdi_to_hgvs_with_ref<R: ReferenceProvider>` (also re-exported at the crate root). Given an SPDI insertion, the function fetches the immediately-5' reference flank and, if those `ins.len()` bases equal the inserted sequence (case-insensitive), rewrites the edit as `NaEdit::Duplication` over the corresponding 1-based interval. All other SPDI shapes (substitution, deletion, delins, identity) pass through unchanged.
- Existing `spdi_to_hgvs` (no reference) is intentionally unchanged. An audit pin enforces this so any future change to the no-reference behavior is loud.
- Reference fetch tries `get_genomic_sequence` first and falls back to `get_sequence` so providers that store contigs as transcripts also work. A hard fetch error becomes `ConversionError::MissingReferenceData`. A short fetch (truncated near a contig boundary) or a non-matching flank silently falls back to ins-form, which matches the spec recommendation that absent evidence of tandem flanking, the change is described as an insertion.
- 13 new unit tests in `src/spdi/convert.rs` cover single-base and multi-base dup recovery, true-insertion non-match, contig-5'-edge insufficient-flank, the spec-FAQ non-tandem inserted-duplicate (must remain ins), pass-through for non-insertion edits, mitochondrial accession, full HGVS dup round-trip, idempotency, reference-fetch error propagation, and the no-ref audit pin. A new public-API integration test (`tests/spdi_dup_recovery.rs`) pins the round-trip and the audit pin at the crate boundary.

## HGVS spec rationale

Per `assets/hgvs-nomenclature/docs/recommendations/DNA/duplication.md`: a duplication is an insertion of nucleotides directly 3'-flanking the original copy, and "when a variant can be described as a duplication it must be described as a duplication and not as e.g. an insertion" (prioritization). The spec FAQ further clarifies that an insertion of a sequence that *also appears elsewhere* in the reference but is not directly 3'-flanking remains an insertion - the implementation honors this by checking only the immediate 5' flank, not arbitrary occurrences.

SPDI itself is canonical and places the insertion at its rightmost shiftable offset, so the matched 5' flank is already the most-3' position; no separate 3' shift is performed.

## Test plan

- [x] `cargo nextest run --features dev` - 3416 tests pass (full suite green)
- [x] `cargo clippy --features dev --all-targets -- -D warnings` - clean
- [x] `cargo fmt --all -- --check` - clean

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -65,7 +65,10 @@ pub use hgvs::parser::{parse_hgvs, parse_hgvs_fast};
 pub use hgvs::variant::HgvsVariant;
 pub use normalize::{NormalizeConfig, Normalizer, ShuffleDirection};
 pub use reference::{MockProvider, MultiFastaProvider, ReferenceProvider};
-pub use spdi::{hgvs_to_spdi_simple, parse_spdi, spdi_to_hgvs, ConversionError, SpdiVariant};
+pub use spdi::{
+    hgvs_to_spdi_simple, parse_spdi, spdi_to_hgvs, spdi_to_hgvs_with_ref, ConversionError,
+    SpdiVariant,
+};
 
 // Re-export coordinate types for type-safe position handling
 pub use coords::{
--- a/src/spdi/convert.rs
+++ b/src/spdi/convert.rs
@@ -423,6 +423,181 @@ pub fn spdi_to_hgvs(spdi: &SpdiVariant) -> Result<HgvsVariant, ConversionError> 
     }))
 }
 
+/// Convert an SPDI variant to HGVS, using a reference provider to recover
+/// duplication form for SPDI insertions whose inserted sequence equals the
+/// immediately-5' reference flank.
+///
+/// SPDI is a canonical/positional format and represents duplications as
+/// insertions of the duplicated sequence. This function performs the inverse
+/// recovery for HGVS round-trips: given an SPDI insertion, if the
+/// `inserted.len()` reference bases ending at the insertion point equal the
+/// inserted sequence (case-insensitive), the change is a tandem duplication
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #157, opened 2026-05-07
fix(spdi): accept c./n./r./m. variants in HGVS->SPDI conversion

nh13:
Closes #116. Follow-up from #81 K1 audit.

## Summary

- `hgvs_to_spdi_simple` now accepts `m.`, `n.`, and `r.` variants in addition to `g.`. SPDI is emitted on the same accession the HGVS variant uses, matching how NCBI Variation Services produces SPDI on transcript accessions.
- Added `hgvs_to_spdi(variant, provider)` for cases that need transcript metadata: `c.` (CDS-relative numbering, including `c.-3` and `c.*5`) and any intronic/downstream/non-positive `n.`/`r.` position. The provider-aware path uses `convert::CoordinateMapper::cds_to_tx` to resolve CDS coordinates to a transcript position.
- `r.` deletion/insertion strings go through an `AlphabetMode::Rna` branch that uppercases and rewrites `u`/`U` to `T`, so the emitted SPDI uses the DNA alphabet RefSeq stores transcripts in.
- `p.` is rejected with a clearer message ("protein variants cannot be represented in SPDI") on both entry points.
- Intronic `c.100+5` returns a clear "SPDI has no offset notation" error — projecting to genomic coordinates is intentionally out of scope for this PR.
- New `ConversionError::ProviderRequired { variant_type, reason }` so callers can distinguish "needs a provider attached" from `MissingReferenceData` ("provider was attached but does not have data for this region").
- Per-edit emission code factored into `emit_spdi_for_edit`, shared across all coordinate systems. Position-resolution helpers (`tx_pos_for_simple_path`, `require_simple_tx_pos`, etc.) make it obvious which positions can be handled without a provider.
- 22 new unit tests across `src/spdi/convert::tests` covering positive, negative, and edge cases for each coordinate system. Existing `c.`-rejection test renamed to `test_hgvs_to_spdi_simple_cds_requires_provider` and updated to assert the new `ProviderRequired` shape. Doc-tests on `hgvs_to_spdi_simple` extended with an `n.` example.

## HGVS spec rationale

Per [the SPDI notation page](https://www.ncbi.nlm.nih.gov/variation/notation/), SPDI is positional on whichever sequence accession is supplied — `NM_000518.4:76:G:GG` is a perfectly valid SPDI on a transcript accession, and NCBI Variation Services emits exactly that form. ferro previously rejected anything that wasn't `g.`, which contradicted the spec. The remaining coordinate systems map to SPDI as follows:

[...]

Issue #118, opened 2026-05-06
spdi: inversion and repeat edits should convert to SPDI delins (#81 K1 follow-up)


nh13:
Tracking issue parent: #81 (item K1 audit)

## Problem

Two HGVS edit types currently return `ConversionError::UnsupportedEditType`:

- `g.100_200inv` — inversions
- `g.100_102TG[10]` — repeat / `SequenceRepeat` notation

Both *can* be represented in SPDI as a `delins` provided we have reference data: an inversion is `delins(<seq>, <revcomp(seq)>)` and a repeat reduces to a delins of the spanned reference vs. the expanded repeat string.

## Scope

When a `ReferenceProvider` is available (see follow-up #117):

- Inversion: fetch the reference, reverse-complement, emit `SpdiVariant::delins(acc, pos0, ref_seq, revcomp(ref_seq))`.
- Repeat: fetch the reference for the repeat span, compute the expanded inserted sequence (`unit.repeat(count)`), emit `SpdiVariant::delins(acc, pos0, ref_seq, expanded)`.

Note that `SPDI → HGVS` will still surface as `delins` rather than `inv` / `[N]`. Whether ferro should detect rev-comp and emit `inv`, or detect repeat structure and emit `[N]`, is intertwined with #81 items A2 and B1 — link them in the PR.

## Audit pin

`tests/hgvs_spdi_roundtrip_audit.rs::audit_unsupported_edit_types` pins the current error behavior.

Issue #116, opened 2026-05-06
spdi: hgvs_to_spdi_simple rejects c./n./r./m. variants (#81 K1 follow-up)


nh13:
Tracking issue parent: #81 (item K1 audit)

## Problem

`hgvs_to_spdi_simple` only accepts `HgvsVariant::Genome` (g.) variants. Any c., n., r., or m. variant returns `ConversionError::UnsupportedVariantType`. SPDI uses transcript / RNA / mitochondrial accessions (e.g. `NM_000518.4:76:G:GG`) just as freely as it uses genomic ones — NCBI Variation Services produce SPDI on transcript coordinates regularly.

## Reproduce

```rust
let v = parse_hgvs("NM_000088.3:c.100A>G").unwrap();
hgvs_to_spdi_simple(&v).unwrap_err(); // UnsupportedVariantType { ... "got c" }
```

## Scope

Add SPDI conversion for at least:
- `HgvsVariant::Cds` (c.) — purely positional cases (no UTR/intronic offsets) should map straightforwardly; offset positions need a coordinate mapper.
- `HgvsVariant::NonCoding` (n.) — analogous to c. without CDS-relative numbering.
- `HgvsVariant::Mito` (m.) — same as g. with a different accession namespace.
- `HgvsVariant::Rna` (r.) — alphabet conversion (u/t) considerations.

Out of scope (covered separately or not applicable to SPDI): `HgvsVariant::Protein`.

## Audit pin

`tests/hgvs_spdi_roundtrip_audit.rs::audit_unsupported_coordinate_systems` pins the current error behavior. Update that test once this is fixed.

PR #158, opened 2026-05-07
fix(spdi): reference-aware HGVS→SPDI for del/dup/delins

### LINKED ISSUE #117: spdi: HGVS→SPDI for del/dup/delins requires reference-aware conversion (#81 K1 follow-up)

## Problem

`hgvs_to_spdi_simple` errors with `ConversionError::MissingReferenceData` for these inputs even though the deletion / duplication / delins length is fully determined by the HGVS interval and the variant could be represented in SPDI given any reference provider:

- `g.100_102del`            → no deleted sequence in HGVS, no provider passed in.
- `g.100_102dup`            → same.
- `g.100_102delinsTTCC`     → only insert sequence given.
- `g.100_102delATGinsTTCC`  → parser drops the explicit deleted sequence (see follow-up D below); even if the parser kept it, the conversion code path would only see `Delins { sequence: TTCC }`.

## Reproduce

```rust
let v = parse_hgvs("NC_000001.11:g.100_102del").unwrap();
hgvs_to_spdi_simple(&v).unwrap_err(); // MissingReferenceData
```

## Scope

Add a reference-aware conversion entry point that accepts a `&dyn ReferenceProvider`, e.g.:

```rust
pub fn hgvs_to_spdi(
    variant: &HgvsVariant,
    provider: &dyn ReferenceProvider,
) -> Result<SpdiVariant, ConversionError>;
```

The internal test `MockGenomicRef` in `src/spdi/convert.rs` already prototypes the API. Promote it to a first-class public function backed by the existing `ReferenceProvider` trait.

## Audit pin

`tests/hgvs_spdi_roundtrip_audit.rs::audit_*_without_reference` pins current error behavior. Update those tests once a reference-aware path lands.

nh13:
Closes #117. Follow-up from #81 K1 audit. Stacks on [#157](https://github.com/fulcrumgenomics/ferro-hgvs/pull/157) (#116) — diff against `main` will narrow when #157 merges.

## Summary

- Wires the `ReferenceProvider` through `emit_spdi_for_edit` so `hgvs_to_spdi(variant, provider)` succeeds for short-form `Deletion`, `Duplication`, and `Delins` HGVS variants. SPDI's mandatory `del` field (and the symmetric `ins` for duplication) is now populated by fetching the variant's interval from the provider.
- New private helper `fetch_reference_bases` does the fetch: tries `provider.get_genomic_sequence` first (correct for `g.`/`m.` and natural for genomic accessions) and falls back to `provider.get_sequence` for transcript accessions (`n.`/`r.`/`c.` SPDI emits on the transcript accession per #116). Both fetch errors and short reads surface as `ConversionError::MissingReferenceData` with the accession + 1-based interval in the message — the same shape PR #119 uses on the inverse direction.
- `hgvs_to_spdi`'s `Genome` and `Mt` arms now route through new `genome_to_spdi_with_provider` / `mt_to_spdi_with_provider` helpers. `tx_to_spdi_with_provider` and `rna_to_spdi_with_provider` no longer short-circuit through their `_simple` siblings, so the provider flows into `emit_spdi_for_edit` even when position resolution does not require provider data. `cds_to_spdi_with_provider` forwards the provider unchanged.
- `hgvs_to_spdi_simple` (no-provider) is unchanged: short-form del/dup/delins continue to err with `MissingReferenceData`, pinned by a new audit test. Explicit-form input (`g.100_102delATG`, `g.100_102dupATG`) bypasses the provider entirely, matching PR #154's rule that ferro never overrides explicit user input. The `Delins` destructure uses a `..` rest pattern so the code compiles against either pre-#154 (`{ sequence }`) or post-#154 (`{ sequence, deleted, deleted_length }`) shapes.
- Module-level doc table and the `hgvs_to_spdi` docstring updated to document the provider-required vs explicit-form columns; new doc-test demonstrates a provider-backed deletion.
- 16 new unit tests in `src/spdi/convert::tests` cover positive (each short-form edit, single-base dup, long deletion, mito accession, MNV delins same-length round-trip, plain-deletion round-trip, dup degrades to ins via reference-free `spdi_to_hgvs`), negative (provider missing data, simple-path audit pin), helper-direct (success, missing contig, short contig), and idempotency.
- Removes the prototype `MockGenomicRef` test scaffold and its companion `hgvs_to_spdi_with_ref` helper. Their cases are now covered through the production path with a real `MockProvider`.

## HGVS spec rationale

Per the [SPDI notation page](https://www.ncbi.nlm.nih.gov/variation/notation/), SPDI's deletion field must contain the actual deleted bases — there is no short form. HGVS allows the short forms `<pos>del`, `<pos>dup`, and `<pos>delinsXX`, deferring to reference data for the bases. ferro's `hgvs_to_spdi` already accepts a `ReferenceProvider`; this PR uses it for the only purpose left in #81's K1 audit: filling SPDI's `del`/`ins` fields when the HGVS input is in short form.

Round-tripping `g.100_102dup` through SPDI produces an insertion; the inverse direction (recovering dup form from a tandem-flanking insertion) is sibling PR [#119](https://github.com/fulcrumgenomics/ferro-hgvs/pull/156) (`spdi_to_hgvs_with_ref`). A test in this PR exercises the reference-free `spdi_to_hgvs` direction (which yields `ins`); once #119 lands the lossless round-trip is `dup → ins-encoded SPDI → dup` via `spdi_to_hgvs_with_ref`.

## Coordination notes

- **Stacked on #157 (#116)**: this branch was cut from `origin/fix/issue-116-spdi-accept-coords`. The PR diff against `main` therefore includes both PRs until #157 merges, then narrows to this PR's changes only.
- **#119 (sibling, SPDI→HGVS dup recovery)**: round-trip lossless when both sides have ref data; this PR's test of the dup→ins direction confirms the SPDI shape (position, inserted bases) that #119 expects to consume.
- **#154 (`Delins.deleted` field)**: the `..` rest pattern in the `Delins` destructure makes this PR forward-compatible. After #154 merges, the same arm can be tightened to short-circuit on `Delins { deleted: Some(_), .. }` (mirror of `Deletion { sequence: Some(_) }`); that's a one-line follow-up.
- **#118 (inv/repeat → SPDI delins)**: orthogonal; remains untouched.

## Test plan

- [x] `cargo nextest run --features dev` — 3440 tests pass (full suite).
- [x] `cargo test --features dev --doc` — 74 doc-tests pass.

### DIFF
--- a/src/spdi/convert.rs
+++ b/src/spdi/convert.rs
@@ -13,16 +13,20 @@
 //!
 //! # Supported Conversions
 //!
-//! | HGVS | SPDI |
-//! |------|------|
-//! | Substitution `g.12345A>G` | `seq:12344:A:G` |
-//! | Deletion `g.100_102del` | `seq:99:NNN:` (requires reference) |
-//! | Insertion `g.100_101insATG` | `seq:100::ATG` |
-//! | Delins `g.100_102delinsATG` | `seq:99:NNN:ATG` (requires reference) |
-//! | Duplication `g.100_102dup` | `seq:102::NNN` (requires reference) |
+//! | HGVS | SPDI | Provider |
+//! |------|------|----------|
+//! | Substitution `g.12345A>G` | `seq:12344:A:G` | not required |
+//! | Deletion `g.100_102del` | `seq:99:ATG:` | required (short form) |
+//! | Deletion `g.100_102delATG` | `seq:99:ATG:` | not required (explicit form) |
+//! | Insertion `g.100_101insATG` | `seq:100::ATG` | not required |
+//! | Delins `g.100_102delinsATG` | `seq:99:ATG:ATG` | required |
+//! | Duplication `g.100_102dup` | `seq:101::ATG` | required (short form) |
+//! | Duplication `g.100_102dupATG` | `seq:101::ATG` | not required (explicit form) |
 //!
-//! Note: Deletions, delins, and duplications require reference sequence data
-//! to determine the deleted sequence.
+//! Short-form deletions, delins, and duplications use [`hgvs_to_spdi`] with a
+//! [`ReferenceProvider`] to fetch the deleted/duplicated bases for SPDI's
+//! mandatory `del` (or, for duplication, `ins`) field. Explicit forms do not
+//! consult the provider.
 //!
 //! # Coordinate-system support
 //!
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #182, opened 2026-05-12
Adjacent edits not merged after canonicalization (inv + neighbor)


msto:
## Context

After the merge step (#80) and the post-merge canonicalization pass (e.g., `delins` → `inv` from #160/#166), the resulting edits can be left as separate bracketed blocks even when they are strictly adjacent (gap = 0 nucleotides). The spec requires adjacent edits to be merged into a single `delins` unless a higher-priority type applies.

## Spec basis

`general.md`: *"two variants separated by one or more nucleotides should be described individually and **not** as a 'delins'"* — the converse being that variants with no intervening nucleotides (adjacent or overlapping) *should* be described as a single edit.

`substitution.md`: *"substitutions involving two or more consecutive nucleotides are described as deletion/insertion (delins) variants"*

## Examples

Examples use a synthetic genomic reference to illustrate the pattern. Positions and bases are representative.

**Example 1** — inv + adjacent sub left as separate blocks

```
Input:    REF:g.[1000G>C;1003A>C;1004G>T;1005A>G]
Ferro:    REF:g.[1000G>C;1003_1004inv;1005A>G]   ← inv and sub are adjacent
Expected: REF:g.1000_1005delinsCACCTG
```

After merge, `1003_1004` becomes `inv`. The adjacent `1005A>G` should be merged with it into a single `delins`, since the combined range `1003_1005` is not an inversion (and no gap separates them).

**Example 2** — inv adjacent to sub on left side

```
Input:    REF:g.[1000G>C;1001A>C;1002C>T;1059G>T;1060A>C;1061A>C;1062G>C]
Ferro:    REF:g.[1000_1002delinsCCT;1059_1060inv;1061A>C;1062G>C]
Expected: REF:g.[1000_1002delinsCCT;1059_1062delinsTCCC]
```

`1059_1060inv` is adjacent to `1061A>C` and `1062G>C`. The entire range `1059_1062` should merge into a single `delins`.

**Example 3** — multiple adjacent blocks post-canonicalization

```
Input:    REF:g.[1000G>C;1003A>C;1004G>C;1006A>C;1007G>C;1008G>C;1009T>C]
Ferro:    REF:g.[1000G>C;1003_1004delinsCC;1006A>C;1007_1008inv;1009T>C]
Expected: REF:g.1000_1009delinsCACCCACCCC
```

Multiple adjacent groups (sub, delins, inv) should be merged where there is no gap.

## Possible approach

[...]

Issue #160, opened 2026-05-08
Recognize reverse-complement sub-spans as `inv` within compound variants


msto:
## Description

The HGVS spec ([`inversion.md:5`](../../assets/hgvs-nomenclature/docs/recommendations/DNA/inversion.md)) defines an inversion as "a sequence change where … more than one nucleotide replacing the original sequence is the reverse complement of the original sequence." For adjacent-variant compounds where a sub-span satisfies this definition, ferro and mutalyzer take different defensible readings of the spec: ferro emits a unified `delins` over the whole compound, while mutalyzer splits the inv-eligible sub-span out as `inv` and represents adjacent independent edits separately.

The spec does not unambiguously mandate either form for the partial-inv case. Two considerations favor the split form:

1. The priority list ([`general.md:56`](../../assets/hgvs-nomenclature/docs/recommendations/general.md)) names `inv` (3rd) and omits `delins`, consistent with `delins` being a residual when no named type applies.
2. Spec examples consistently describe adjacent-variant rev-comp runs as `inv` rather than as `delins` (e.g. [`inversion.md:27-28`](../../assets/hgvs-nomenclature/docs/recommendations/DNA/inversion.md)). Extending this preference to inv-eligible sub-spans within a larger compound is a natural generalization.

A counter-consideration: the planned-future rule at [`general.md:39`](../../assets/hgvs-nomenclature/docs/recommendations/general.md) would have variants separated by less than two nucleotides described as `delins`, which would favor ferro's current behavior over the split form.

## Examples (synthetic genomic reference)

| Input | Mutalyzer | Ferro (current) |
|---|---|---|
| `g.[1150T>G;1151C>A;1152C>G]` | `g.[1150_1151inv;1152C>G]` | `g.1150_1152delinsGAG` |
| `g.[1092G>C;1093G>C]` | `g.1092_1093inv` | `g.1092_1093delinsCC` |

Verification of the first row: ref at 1150-1151 is `TC`; alt is `GA`; reverse-complement of `TC` is `GA`. Positions 1150-1151 therefore satisfy the inversion definition. The full 1150-1152 span does not (rev-comp of `TCC` is `GGA`, not `GAG`), so under the split reading position 1152 is described as an independent substitution.

[...]

PR #206, opened 2026-05-16
fix(normalize): decompose delins to multi-sub canonical form when sub > delins applies (closes #165)

### LINKED ISSUE #165: Decompose delins to multi-sub canonical form when sub > delins applies (extension of #160)
## Context

Issue #160 added recognition of reverse-complement sub-spans within a `delins` and emits the spec-canonical `inv`, decomposing the surrounding span into `[…; inv; …]`. The implementation deliberately scoped the decomposition to fire **only when at least one `Inversion` sub-edit is found** — a `delins` containing two or more independent substitutions but no rev-comp sub-span stays as `delins` rather than decomposing to `[X>A; B>Y]`.

This issue tracks the deferred case: applying the same decomposition rule for the substitution-priority case (`sub > delins`).

## Spec basis

HGVS edit priority (`general.md:56`): `substitution > deletion > inversion > duplication > insertion`. `delins` is omitted from the priority list and is defined in `delins.md` as *"a sequence change ... which is **not** a substitution, inversion or conversion"* — i.e., the residual when no higher-priority form applies.

Strictly read, this means a `delins` whose post-trim span contains two or more independent single-base mismatches (no rev-comp sub-span) is **not** spec-canonical: each mismatch is a substitution, and `sub > delins` in the priority order, so the canonical form is `[X>A; B>Y]`.

## Counter-considerations

1. **Planned-future `general.md:39` rule.** The spec roadmap mentions describing variants separated by less than two nucleotides as `delins`. Implementing sub-only decomposition reverses that planned direction. Issue #160's body explicitly cites this tension.
2. **Blast radius.** Most cis-allele merge results are 2-nt non-inv delins (e.g. `g.[100G>A;101A>C]` → `g.100_101delinsAC`). Applying decomposition would re-expand essentially every consecutive-SNV cis allele back into an Allele, which is a much larger fraction of inputs than the inv case.
3. **Codon-frame merge (#79) interaction.** The codon-frame merge for two in-codon `c.` SNVs separated by 1 nt synthesizes a 3-base delins. Decomposition under `sub > delins` would undo this merge whenever the result has no inv (which is mathematically always, see the #160 design doc). Either accept that the codon-frame merge becomes a no-op for inv-free cases, gate decomposition with a codon-frame exception, or revisit whether the codon-frame merge is the right rule.

## Examples

| Input | Current ferro (after #160) | Sub-only decomposed |
|---|---|---|
| `g.[100G>A;101A>C]` over `GA` | `g.100_101delinsAC` | `g.[100G>A;101A>C]` |
| `g.100_101delinsAC` over `GA` | `g.100_101delinsAC` | `g.[100G>A;101A>C]` |
| `c.[10C>G;12C>A]` over `CCC` (codon-frame) | `c.10_12delinsGCA` | `c.[10C>G;12C>A]` |

## Tests that would change

- `tests/merge_consecutive_edits_tests.rs::test_merge_cds_consecutive_subs` (and tx/rna analogues)
- `tests/merge_consecutive_edits_tests.rs::test_codon_frame_two_subs_one_codon`
- `tests/python/test_core.py::TestNormalizeMergeConsecutive::test_consecutive_subs_collapse_to_delins`
- Any ClinVar/mutalyzer comparison harness that currently expects the merged-delins form.

## Implementation sketch

The existing `decompose_delins_inv` in `src/normalize/rules.rs` already does the per-position scan; the only change needed is loosening the trigger from `has_inv && emitted.len() >= 2` to just `emitted.len() >= 2` (regardless of inv presence). The function should probably be renamed to `decompose_delins` accordingly. Per-coord-system wiring (`apply_inv_split` and `split_inv_for_variant` in `src/normalize/mod.rs`) already handles the multi-edit return shape, so no plumbing changes downstream of `rules.rs`.

## References

- #160 (parent): inv sub-span recognition.
- #81 (tracker): added as a new item under section A.
- #79: codon-frame merge that this would interact with.

nh13:
## Summary

Closes #165. Addresses #81 item A10 (sub-only branch).

A `delins` whose post-trim span contains two or more independent single-base mismatches separated by at least one unchanged nucleotide now decomposes to individual `[X>A; B>Y]` form per the HGVS edit-priority rule (`general.md:56`: `sub > del > inv > dup > ins`; `delins` is the residual when no higher-priority form applies). Adjacent (no-gap) mismatches stay as `delins` per `substitution.md`'s consecutive-nucleotides rule (#182). The spec's narrow codon-frame exception (`general.md:35-38`, originally from #79) is preserved inside `build_split_variants`: when a `c.` variant's decomposition contains a `[Sub@i; Identity@i+1; Sub@i+2]` triplet whose CDS endpoints share a codon, the triplet emits as a single 3-base delins even when embedded inside a longer span.

## Behaviour change

`merge_consecutive_edits_tests::test_codon_frame_then_strict_chain` previously expected `c.10_13delinsGCCT` for input `c.[10A>G;12A>C;13A>T]` (the codon-frame merge's strict-adjacency chain extension). The spec exception is defined for a *pair* of variants ("two variants together affecting one amino acid"), not a chain spanning two codons, so A10 collapses the chain back to the spec-canonical `c.[10C>G;13C>T]`. The user-declared `12A>C` against an actual reference C is a no-op that the canonicalization drops, and the surviving subs render with the correct refs from the provider.

## Implementation

- `decompose_delins_inv` renamed to `decompose_delins` in `src/normalize/rules.rs`. Gate loosened to fire on any emission with `>1` element containing at least one `Inversion` or `IdentityAt`. Pure-substitution emissions short-circuit to `None` because the caller would just re-merge them under the adjacency rule. Non-IUPAC bytes at any position abandon the decomposition so the next round-trip lands on the same string.
- `DelinsSubedit::IdentityAt` carries the unchanged ref byte, enabling codon-frame triplet reconstruction.
- `build_split_variants` takes a `codon_frame_aware: bool` flag (true only for `HgvsVariant::Cds`); the new triplet lookahead emits a 3-base delins for in-codon `[Sub; Identity; Sub]` patterns.
- `merge::same_codon` promoted to `pub(crate)` so the post-merge pass can re-apply it.
- Per-`Normalizer` helpers renamed from `_inv_split` to `_canonical_split` to cover both #160 and #165 paths.

## Review

Pre-PR parallel reviews by Opus and Sonnet sub-agents; all findings addressed in the squashed commit (most-fix correctness on the `IdentityAt` `Base::N` fallback, plus naming/doc consistency, missing `g.` integration test, additional unit-test coverage, and large-position `same_codon` assertions).

## Test plan

- [x] `cargo nextest run --features dev` — 4409 passed, 0 failed.
- [x] `pytest tests/python/` — 157 passed.
- [x] Round-trip stability tests in `tests/issue_165_delins_sub_only_decompose.rs` (g./n./r./m./c.).

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -518,7 +518,11 @@ fn build_mt_merged(template: &MtVariant, merged: Anchor) -> MtVariant {
 /// Two CDS positions share a codon if they fall in the same 1-indexed
 /// triplet: `c.1..3` is codon 1, `c.4..6` is codon 2, etc. The standard
 /// codon-number formula is `(base - 1) / 3` (integer division).
-fn same_codon(a: i64, b: i64) -> bool {
+///
+/// Exposed at `pub(crate)` so the post-merge `decompose_delins`
+/// pass can re-apply the spec's codon-frame exception (issue #165 /
+/// tracking issue #81 item A10).
+pub(crate) fn same_codon(a: i64, b: i64) -> bool {
     if a < 1 || b < 1 {
         return false;
     }
@@ -631,5 +635,10 @@ mod tests {
         assert!(!same_codon(3, 5)); // codon 1 vs codon 2
         assert!(!same_codon(0, 2)); // 0 invalid
         assert!(!same_codon(-3, -1));
+        // Large positions exercise the `(base - 1) / 3` formula far from
+        // its near-zero edge. c.99997..c.99999 falls in codon 33333;
+        // c.99998..c.100000 straddles codon 33333 and 33334.
+        assert!(same_codon(99997, 99999));
+        assert!(!same_codon(99998, 100000));
     }
 }
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -283,17 +283,20 @@ impl<P: ReferenceProvider> Normalizer<P> {
         let merged_raw =
             merge::merge_consecutive_edits(allele.variants.clone(), allele.phase, &self.provider);
 
-        // Issue #160: any merged delins (or pre-existing delins that survived
-        // merge unchanged) may decompose into [..., inv, ...] when an inv-
-        // eligible sub-span is present. Run the split per merged variant; the
-        // helper is a no-op for non-Delins variants and for Delins without an
-        // inv sub-span. Only applies in cis phase — trans alleles aren't
+        // Issue #160 + #165: any merged delins (or pre-existing delins
+        // that survived merge unchanged) may decompose into a sequence of
+        // higher-priority forms per `general.md:56` — `[..., inv, ...]`
+        // when an inv-eligible sub-span is present (#160) and/or into
+        // separate substitutions when interior positions match the
+        // reference (#165). Run the split per merged variant; the helper
+        // is a no-op for non-Delins variants and for Delins with nothing
+        // to split out. Only applies in cis phase — trans alleles aren't
         // collapsible in the first place.
         let merged_split: Vec<HgvsVariant> =
             if allele.phase == crate::hgvs::variant::AllelePhase::Cis {
                 merged_raw
                     .into_iter()
-                    .flat_map(|v| self.split_inv_for_variant(v))
+                    .flat_map(|v| self.canonical_split_for_variant(v))
                     .collect()
             } else {
                 merged_raw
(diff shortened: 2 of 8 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #215, opened 2026-05-16
feat(normalize): validate repeat unit divides reference span and matches bases (closes #214)

### LINKED ISSUE #214: Repeat unit length must divide reference span and bases must match unit*N (normalize-time consistency check)
Spec: [`assets/hgvs-nomenclature/docs/recommendations/DNA/repeated.md`](../blob/main/assets/hgvs-nomenclature/docs/recommendations/DNA/repeated.md) requires that for a `unit[N]` description:

1. the reference span `[start, end]` is exactly `unit_len * N` bases long, and
2. the bases in the reference at `[start, end]` equal `unit` repeated exactly `N` times.

Today the parser only validates the *structure* of `unit[N]` (single-base or multi-base unit, integer count, valid brackets). It does not (and cannot — it has no reference handle) verify that a user-supplied `[start, end, unit, N]` triple is internally consistent against the reference. As a result, an inconsistent description — for example `g.100_105AC[3]` where the reference at `g.100_105` is `ACACAT` (so span is correct but the last base mismatches), or `g.100_105AC[4]` where the span doesn't divide by `unit_len * N` at all — passes parse and reaches normalize, where it can either silently misnormalize or land in places the spec considers undefined.

### Scope (this issue)

Wire a normalize-time consistency check on `NaEdit::Repeat { sequence: Some(unit), count: RepeatCount::Exact(N), .. }` (and the mixed-repeat / additional_counts case) that, when reference sequence is available:

- verifies `(end - start + 1) == sum(unit_len_i * N_i)` across the unit list, and
- verifies the bases in the reference at `[start, end]` equal the concatenation of `unit_i` repeated `N_i` times in order.

When either invariant fails the variant must surface as an `InvalidRepeatDescription` (or similar) error, threaded through the existing `error_handling/` strict/lenient/silent modes. When the reference is unavailable (no provider data, missing accession) the check is skipped and a `coverage_gap` style warning is emitted, consistent with how other normalize-time consistency checks handle missing reference data.

### Cited in

- [#81 § B5](../issues/81) — HGVS spec compliance audit (DNA → Repeated Sequences).

### Acceptance

- TDD audit at `tests/issue_<N>_repeat_unit_divides.rs` covering: unit_len doesn't divide span; span correct but reference bases mismatch `unit * N`; round-trip preserved for consistent descriptions; intronic + UTR coverage; minus-strand transcripts.
- Spec-fixture pin updated if any spec row flips (`cargo run --features dev --example generate_spec_fixture`).
- Bulk failure expectations edited surgically if any row label changes; never run wholesale `UPDATE_FAILURE_EXPECTATIONS=1`.

nh13:
Closes #214 ([#81](https://github.com/fulcrumgenomics/ferro-hgvs/issues/81) item B5).

## Summary

Per HGVS v21.0 ([`docs/recommendations/DNA/repeated.md`](../blob/main/assets/hgvs-nomenclature/docs/recommendations/DNA/repeated.md)) a `unit[N]` description over `[start, end]` carries two reference-side invariants:

1. `(end - start + 1) % unit_len == 0` — the span divides cleanly into whole units.
2. The reference bases at `[start, end]` equal `unit` repeated `(end - start + 1) / unit_len` times.

The parser is purely structural and has no reference handle, so descriptions that satisfy each rule in isolation but are inconsistent against the reference reached `normalize_repeat`, which silently misnormalized or emitted spec-undefined positions:

| Input | Reference | Before |
|---|---|---|
| `g.100_106AC[3]` (span 7, unit_len 2) | `ACACAC` at 100..=105 | normalized to `g.107_108del` |
| `g.100_105AC[3]` | `ACACAT` at 100..=105 (last unit mismatch) | accepted silently |
| `g.100_141CTG[2]TTG[1]CTG[11]` (span 42, but ref has different mixed order) | content mismatch | accepted silently |

This wires the consistency check into the existing `validate::validate_reference` gate so it threads through the existing strict/lenient/silent `RefSeqMismatch` flow:

- **Strict**: rejects with `FerroError::ReferenceMismatch`.
- **Lenient**: emits `NormalizationWarning::RefSeqMismatch` and keeps the input verbatim.
- **Silent**: does not reject (warning is still recorded for observability, matching the existing sub/del/dup pattern).

The variant count `N` is the alt-allele count and is independent of the reference repeat count (which `normalize_repeat` rediscovers by scanning), so the check is between the *span* and the *reference bases* only — the spec example `NM_000044.3:c.171_239GCA[34]` (span 69 = 23 × 3, variant count 34) is accepted unchanged.

## Scope

- `NaEdit::Repeat { sequence: Some(unit), additional_counts: [], trailing: None, .. }` — the divisibility-and-content check runs.
- `NaEdit::Repeat` with `trailing.is_some()` (VEP-style `c.212-18CTG[3]T`) or non-empty `additional_counts` (genotype `A[6][1]`) — **skipped**, mirroring the same skip-list `normalize_na_edit` already uses for these shapes. Without this guard the strict gate would surface a spurious `ReferenceMismatch` on shapes the normalizer itself declines to rewrite. This was flagged by both Opus and Sonnet review before merge and pinned by two regression tests.
- `NaEdit::MultiRepeat { units }` — checks `sum(unit_len_i * N_i) == span` and that the reference bases match the concatenation. Skipped when any unit has a non-Exact count (Range / MinUncertain / MaxUncertain / Unknown — the sum is ambiguous). Required adding `MultiRepeat` to `needs_normalization()`'s allowlist so the variant reaches the validation gate instead of short-circuiting; the inner match in `normalize_na_edit` already has a `_ =>` catch-all that returns `MultiRepeat` unchanged after validation, so no behavior change beyond running the new check.
- Missing reference data (provider lacks the accession) — gracefully skipped, no spurious mismatch.

## Files

- `src/normalize/validate.rs` — two new arms in `validate_reference`; new private `validate_repeat_tract` and `validate_multirepeat_tract`.
- `src/normalize/rules.rs` — `NaEdit::MultiRepeat` added to `needs_normalization()`'s allowlist.
- `tests/issue_214_repeat_unit_divides.rs` — 24 TDD audit cases (sections: g. unit-doesn't-divide × strict/lenient/silent; g. ref-content mismatch; g. consistent descriptions; g. MultiRepeat; VEP `trailing` + genotype skip-list; no-reference-data graceful skip; c. UTR exempt-from-codon-frame still gated by content; c. intronic plus-strand + minus-strand; n. non-coding).

## Test plan

- [x] `cargo nextest run --features dev` — 4451 / 4451 pass (was 4449 before, +2 from new tests).
- [x] No bulk failure-expectations rows flipped (`tests/fixtures/bulk/*.json`).
- [x] No spec-fixture rows flipped (`tests/fixtures/grammar/hgvs_spec_normalization.json`); no regen needed.
- [x] Parallel Opus + Sonnet reviews completed; both flagged the same must-fix (skip on `trailing` / `additional_counts`), addressed before merge.

## Follow-ups (deliberately out of scope)

- The `MultiRepeat` validation arm fires only when every unit has `RepeatCount::Exact(_)`. Validating mixed-repeat shapes with Range / Unknown counts is possible but the expected total length is ambiguous; a follow-up could carve out a "partial-validation" mode that checks bases for the known-Exact prefix.
- The `corrected: true` flag on the new `RefSeqMismatch` warning is hardcoded (matches the existing sub/del/dup convention), but for a Repeat consistency mismatch the normalizer cannot and does not correct the description — it passes through verbatim. This is a pre-existing flag-semantics quirk shared with all `RefSeqMismatch` warnings and is left untouched here.

### DIFF
--- a/src/normalize/rules.rs
+++ b/src/normalize/rules.rs
@@ -30,6 +30,7 @@ pub fn needs_normalization(edit: &NaEdit) -> bool {
             | NaEdit::Delins { .. }
             | NaEdit::Inversion { .. }
             | NaEdit::Repeat { .. }
+            | NaEdit::MultiRepeat { .. }
     ) {
         return true;
     }
--- a/src/normalize/validate.rs
+++ b/src/normalize/validate.rs
@@ -15,7 +15,7 @@
 use crate::coords::hgvs_pos_to_index;
 use crate::error::FerroError;
 use crate::error_handling::ResolvedAction;
-use crate::hgvs::edit::{Base, NaEdit};
+use crate::hgvs::edit::{Base, NaEdit, RepeatUnit, Sequence};
 use crate::normalize::config::NormalizeConfig;
 
 /// Result of reference validation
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #218, opened 2026-05-16
Mixed-accession alleles — dedicated coverage audit


nh13:
Tracking issue #81 item C4.

## Background

Compound alleles in HGVS can carry sub-variants on different reference sequences. The spec discourages mixed *reference sequence types* (`c.[76A>C];g.[10091C>G]`, per `assets/hgvs-nomenclature/docs/recommendations/DNA/alleles.md`) but is silent on **mixed-accession same-coord-type** alleles like `[NM_000088.3:c.1A>G;NM_000089.3:c.2A>G]` (two different transcripts of the same gene, or unrelated transcripts on the same chromosome).

Existing coverage is scattered:

- `tests/compound_cross_reference.rs` is comprehensive for **cross-coord-system** alleles (c. + g. + r. + p. + n. + m. + o.) but only touches mixed-accession same-coord-type cases incidentally (one or two lines per pattern).
- `tests/allele_unknown_phase.rs` covers `[NM_X(;)NM_Y]` for unknown-phase only.
- `tests/gene_selector_roundtrip.rs` covers gene-selector forms incidentally.

C4 in #81 explicitly calls out that mixed-accession alleles "[deserve] a dedicated coverage test." This issue tracks that audit.

## Scope

Pin every phase × shape combination for mixed-accession same-coord-type alleles:

- Cis `[NM_X:c.1A>G;NM_Y:c.2A>G]`, trans `[NM_X:c.1A>G];[NM_Y:c.2A>G]`, unknown-phase `[NM_X:c.1A>G(;)NM_Y:c.2A>G]`, mosaic `NM_X:c.1A>G/NM_Y:c.2A>G`, chimeric `NM_X:c.1A>G//NM_Y:c.2A>G`.
- 2-way, 3-way, 4-way.
- Mixed versions of the same accession base (`NM_X.1` vs `NM_X.2`).
- Mixed accession types within the same coord system (`NM_X` vs `NR_Y` for n., `NP_X` vs `NP_Y` for p.).
- Mixed gene-selector forms (`NM_X(GENEA):c.A>G;NM_Y(GENEB):c.B>C`).
- Normalize-each-independently semantics.
- Compact form (`ACC:c.[a;b]`) MUST NOT collapse when accessions differ.
- Merge pass MUST NOT bridge different accessions.

## Acceptance

- New `tests/issue_<N>_mixed_accession_alleles.rs` audit covering the lattice above.
[...]

PR #230, opened 2026-05-16
fix(normalize): read dup bases from reference (single-pass idempotency for dup with mismatched stated-ref) (closes #219)

### LINKED ISSUE #219: Two-pass normalize drift on `<acc>:c.<pos>dup<base>` when partner shares coord type but not accession
Surfaced during the #218 (C4 mixed-accession) audit.

## Reproducer

```rust
let provider = /* MockProvider with NM_TEST.1 holding "ATGCAAAAACCCCCGGGGGTTTTTAAAAACCCCCGGGGGTTTTTAAAAACCCCCGGGGGT" */;
let parsed = parse_hgvs("[NM_TEST.1:c.10dupA;NM_OTHER.1:c.5A>G]").unwrap();
let pass1 = normalizer.normalize(&parsed).unwrap();
let pass2 = normalizer.normalize(&pass1).unwrap();

// pass1: "[NM_TEST.1:c.10dup;NM_OTHER.1:c.5A>G]"
// pass2: "[NM_TEST.1:c.14dup;NM_OTHER.1:c.5A>G]"
assert_eq!(format!("{}", pass1), format!("{}", pass2));  // FAILS
```

## Observed

- Pass 1 drops the mismatched stated-ref base (`c.10dupA` → `c.10dup`) but does NOT 3'-shift.
- Pass 2 takes the now-canonical `c.10dup` and 3'-shifts it to `c.14dup` (the rightmost copy of the `CCCCC` tract).

## Expected

Pass 1 should either:
(a) emit the fully 3'-shifted form `c.14dup` in one pass, or
(b) emit `c.10dup` AND lock that as a fixed point.

Idempotency (`normalize(normalize(v)) == normalize(v)`) is a normalizer invariant.

## Notes

- The same input with a `g.` partner (`[NM_TEST.1:c.10dupA;NC_000001.11:g.100A>G]`) IS idempotent — see `tests/compound_cross_reference.rs::test_normalize_idempotent_with_resolvable_companion`. So the bug is triggered by the *same-coord, different-accession* partner pattern specifically.
- Likely root cause: the stated-ref correction path on pass 1 skips the subsequent 3'-shift. The pass-2 shift then fires because the input is in a "clean" form. The fix is to run the shift inside the same pass as the correction (or vice versa, lock the corrected form as the fixed point).

## Acceptance

- A dedicated `tests/issue_<N>_mixed_accession_normalize_idempotency.rs` audit pinning the bug + the fix.
- Idempotency assertion holds for the c.+c. partner pattern.
- Surgical bulk-failure-expectations edits if any row flips; spec-fixture pin if rows flip.

Cited in #218 normalize section as deliberately out-of-scope.

nh13:
Closes #219.

## Background

`c.10dupA` against a transcript where position 10 is `C` (start of a `CCCCC` homopolymer) was not idempotent across two normalize passes:

| Pass | Input | Output |
|---|---|---|
| 1 | `c.10dupA` | `c.10dup` (mismatched stated-ref stripped, position NOT shifted) |
| 2 | `c.10dup` | `c.14dup` (position-only form correctly 3'-shifts) |
| 3 | `c.14dup` | `c.14dup` (fixed point) |

Surfaced during the #218 (C4 mixed-accession) audit. The drift was visible solo (`c.10dupA`), in cis compounds (both same-coord-different-accession and cross-coord), and in trans compounds.

## Root cause

`normalize_na_edit`'s `NaEdit::Duplication { sequence, .. }` arm used the user-stated `sequence` bytes as the alt sequence for shuffle when present, falling back to reference bytes only when `sequence` was None:

```rust
NaEdit::Duplication { sequence, .. } => {
    if let Some(seq) = sequence {
        seq.bases().iter().map(|b| b.to_u8()).collect()
    } else {
        // read from reference
    }
}
```

When `validate_reference` flagged a stated-ref mismatch in lenient mode, the corrected shape (with `sequence: None`) was only emitted AFTER the shuffle had already run with the bad bases. The shuffle's alt sequence and the actual reference bytes no longer matched, so the shift didn't fire.

## Fix

Always read duplicated bytes from the reference for shuffle, regardless of any user-stated bases on the input. The canonical HGVS Display drops the stated bases (see the dup output arm in `get_canonical_form`), and the stated bases are purely informational — they aren't load-bearing for normalization. Reading from reference up-front collapses pass 1 and pass 2 behavior into a single pass.

## Files

- `src/normalize/mod.rs` — `NaEdit::Duplication` arm in `normalize_na_edit` always reads from reference.
- `tests/issue_219_dup_stated_ref_idempotent.rs` — 8 audit cases (solo + inside-compound + symmetric del regression guard).
- `tests/compound_cross_reference.rs::test_each_variant_normalizes_against_own_ref` + `test_unresolved_companion_does_not_break_resolved_variant_normalize` — updated expected outputs from `c.10dup` (the old pass-1-stuck-at-intermediate result) to `c.14dup` (the now-correct single-pass shifted form).
- `tests/issue_218_mixed_accession_alleles.rs::cis_partial_provider_resolves_only_known_accession` + `trans_partial_provider_resolves_only_known_accession` — updated expected outputs; removed the obsolete "tracked as #219 follow-up" comment.

## Test plan

- [x] `cargo nextest run --features dev` — 4507 / 4507 pass.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -1741,19 +1741,30 @@ impl<P: ReferenceProvider> Normalizer<P> {
                     return Ok((start, end, edit.clone(), warnings.clone()));
                 }
             }
-            NaEdit::Duplication { sequence, .. } => {
-                if let Some(seq) = sequence {
-                    seq.bases().iter().map(|b| b.to_u8()).collect()
+            NaEdit::Duplication { .. } => {
+                // Always read duplicated bytes from the reference,
+                // regardless of any user-stated bases on the input
+                // (`dup<base>` shapes). The canonical HGVS Display
+                // drops the stated bases anyway (see the dup output
+                // arm in `get_canonical_form`); using the stated
+                // bases here when they don't match the reference
+                // (the parser accepts them in lenient mode with a
+                // `RefSeqMismatch` warning) caused the shuffle to
+                // mis-shift and broke single-pass idempotency:
+                // `c.10dupA` against a `CCCCC` homopolymer
+                // canonicalized to `c.10dup` on pass 1 (stated-ref
+                // stripped without shifting) and only shifted to
+                // `c.14dup` on pass 2 (because the now-clean form
+                // reads bytes from reference and the shuffle fires
+                // correctly). Reading from reference up-front
+                // collapses both pass 1 and pass 2 behavior into a
+                // single pass. Issue #219.
+                let s = hgvs_pos_to_index(start);
+                let e = end as usize;
+                if e <= ref_seq.len() {
+                    ref_seq[s..e].to_vec()
                 } else {
-                    // Get duplicated sequence from reference
-                    // start is 1-based, convert to 0-based index for array access
-                    let s = hgvs_pos_to_index(start);
-                    let e = end as usize;
-                    if e <= ref_seq.len() {
-                        ref_seq[s..e].to_vec()
-                    } else {
-                        vec![]
-                    }
+                    vec![]
                 }
             }
             NaEdit::Delins { sequence, .. } => {
--- a/tests/compound_cross_reference.rs
+++ b/tests/compound_cross_reference.rs
@@ -264,10 +264,11 @@ fn test_unknown_phase_cross_reference_round_trip() {
 
 #[test]
 fn test_each_variant_normalizes_against_own_ref() {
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #282, opened 2026-05-17
parser(rna): decide and pin r. thymine policy — reject `t` or canonicalize to `u` (#81 E follow-up to #232)


nh13:
PR #232 (closes #231) pinned RNA lowercase enforcement coverage but left the `r.` thymine policy as a parser-policy decision: observed behavior is pinned only.

## Source

PR #232 § *Out of scope*:

> **`r.` thymine policy** — whether to reject `t` or canonicalize to `u` is a parser-policy decision, not part of E1. Observed behavior is pinned only.

## Decision required

Pick one of:

- **Reject** `r.123a>t` etc. at parse with a targeted diagnostic.
- **Canonicalize** to `u` (lenient) with a soft-validation warning.

Either path is a small parser change; this issue exists to capture the decision and the regression suite that should follow.

## Cross-ref

PR #232, #231 (closed). Fits #81 § E. Adjacent: #170 (preprocessor lowercase IUPAC) — orthogonal but in the same parser-policy neighborhood.

PR #107, opened 2026-05-03
dup 3'-shift coverage matrix (#81 A6)

nh13:
Closes [#81](https://github.com/fulcrumgenomics/ferro-hgvs/issues/81) item A6 (3'-shift coverage matrix for `dup`). Mirrors [#93](https://github.com/fulcrumgenomics/ferro-hgvs/pull/93) (A1+A7, ins) and [#106](https://github.com/fulcrumgenomics/ferro-hgvs/pull/106) (A5+B2, del) to complete the ins/del/dup canonical-form trilogy.

**Stacked on [#106](https://github.com/fulcrumgenomics/ferro-hgvs/pull/106)** — will retarget to `main` automatically when #106 merges.

## Summary

Locks ferro's 3'-shift behavior for **duplication** variants across every nucleotide coord-system / strand combination. **No production code changes** — `duplication_to_repeat` (introduced by #93) already handles the canonical forms; this PR adds the comprehensive test coverage.

### What's new

**Test infrastructure**
- New file `tests/dup_shift_matrix.rs`: 98 rstest cases organized into 7 modules (`genomic`, `cds_plus`, `cds_minus`, `noncoding_plus`, `noncoding_minus`, `rna_plus`, `rna_minus`), each exercising 8 shuffle scenarios.
- Reuses `tests/common/synthetic.rs` (`SyntheticBuilder`) introduced by #93.

### Scenario matrix

| # | Scenario | Expected canonical form |
|---|---|---|
| 1 | No-shift (dup base differs from neighbor) | unchanged `dup` |
| 2 | Single-base dup in homopolymer | `dup` shifted to 3' end (single-base, stays as dup) |
| 3 | Multi-base dup of 1 tandem unit | `dup` shifted to 3'-most unit |
| 4 | Multi-base dup of **2+ units** | `unit[N+k]` at the canonical pre-edit reference tract |
| 5 | Multi-base dup that shifts but is not a tandem-unit duplication | `dup` at most-3' position (no canonical-form change) |
| 6 | Shift clamped by tandem boundary | stops at boundary |
| 7 | Dup at 5' edge of ref / transcript start | shifts cleanly; no underflow |
| 8 | Dup at 3' edge of ref / transcript end | clamps to last position |

[...]

PR #106, opened 2026-05-03
del 3'-shift coverage matrix + tandem-repeat del canonical-form fix (#81 A5, B2)

nh13:
Addresses [#81](https://github.com/fulcrumgenomics/ferro-hgvs/issues/81) items A5 (3'-shift coverage matrix for `del`) and B2 (`del` reducing a tandem repeat must render as `unit[N−k]`). Mirrors [#93](https://github.com/fulcrumgenomics/ferro-hgvs/pull/93) (A1 + A7, ins). The matrix scenarios lock B2 with strict string equality across all 7 coord-system × strand combinations, so this PR de-facto closes B2 alongside A5.

## Summary

Locks ferro's 3'-shift behavior for **deletion** variants across every nucleotide coord-system / strand combination, in a single audit-shaped matrix file. Surfaces and implements the canonical-form rule "deletion of ≥2 tandem-repeat units → `unit[N−k]`" (B2).

### What's new

**Test infrastructure**
- New file `tests/del_shift_matrix.rs`: 105 rstest cases organized into 7 modules (`genomic`, `cds_plus`, `cds_minus`, `noncoding_plus`, `noncoding_minus`, `rna_plus`, `rna_minus`), each exercising 8 shuffle scenarios.
- Reuses the shared `tests/common/synthetic.rs` helper (`SyntheticBuilder`) introduced by #93.

**Production code**
- New `pub(crate) struct TandemTract` and `pub(crate) fn extend_tandem_tract` in `src/normalize/rules.rs` — a unit-by-unit tract walker, factored out of `find_tandem_extent`'s inner walk so it can be reused by both ins and del paths. `find_tandem_extent` now delegates the inner walk; no behavior change for `insertion_to_repeat`.
- New `pub(crate) struct DelToRepeatResult` and `pub(crate) fn deletion_to_repeat` — returns the canonical `unit[N−k]` description when 5 trigger conditions hold (length divisible by unit, tract has ≥2 ref units, k≥2 units removed, ≥1 unit remaining post-deletion). The doc comment includes the **shuffle phase-alignment lemma** justifying why no rotation iteration is needed in this code path.
- New dispatch arm in `normalize_na_edit`'s `NaEdit::Deletion` branch — calls `deletion_to_repeat` post-shuffle, returns a `Repeat` edit when it fires.
[...]

PR #293, opened 2026-05-17
fix(display): canonicalize r. emission T→u in repeat units and bases (closes #276)

### LINKED ISSUE #276: display(rna): emitted unit containing DNA T kept as t instead of canonicalized to u (#81 E follow-up to #106/#107)
The RNA `Display` path leaves the lowercase `t` in repeat-unit emission when the unit contains a DNA `T`, instead of canonicalizing to `u` per HGVS RNA-alphabet rules.

## Source

PR #106 § *Out of scope*:

> A pre-existing limitation in the RNA Display path: when the emitted unit contains DNA `T`, the lowercase RNA path keeps `t` instead of converting to `u`. Surfaced by scenario 4's finer-periodicity case; worked around in the matrix by using an `AC` tandem for the RNA module (see comment in `tests/del_shift_matrix.rs`). Worth filing as a separate issue.

PR #107 confirms the same workaround in the `dup` 3'-shift matrix:

> Same workaround as A5: the RNA module's finer-periodicity sub-case uses an `AC` tandem rather than `AT` to avoid the pre-existing limitation in the RNA Display path … This is a known issue worth filing as a separate follow-up.

## Repro sketch

A repeat-unit emission such as `r.123at[5]` is observed where the spec-canonical form is `r.123au[5]`. See `tests/del_shift_matrix.rs` workaround comments for the precise inputs.

## Cross-ref

PR #106, PR #107. Fits #81 § E (RNA canonicalization).

nh13:
## Summary

The RNA `Display` path lowercased bytes without translating DNA `T` to RNA `U`, so emitted repeat units and bases containing `T` rendered as `t` instead of the spec-canonical `u`. Fix is a single arm in `Base::to_lowercase_char` (`src/hgvs/edit.rs`) mapping `Base::T` → `'u'`. Every `NaEdit::to_rna_string` callsite — substitution, deletion, insertion, duplication, inversion, delins, repeat, multi-repeat, identity stated-ref, deleted-ref legs, compound brackets — routes through this single helper.

## Test cleanup

PRs #106 / #107 worked around this bug in their 3'-shift coverage matrices by switching the finer-periodicity RNA cases to an `AC` tandem instead of the original `AT`. Both workarounds are removed and the matrices restored to the spec-equivalent `AU` tandem. Two pre-existing tests in `tests/issue_231_rna_lowercase_coverage.rs` that pinned the buggy behavior are renamed and updated to lock the canonical form.

## Tests

`tests/issue_276_rna_display_t_to_u.rs` — 25 tests pinning T→u canonicalization across substitution, deletion, insertion, duplication, inversion, delins, identity stated-ref, repeat units, multi-repeat units, and compound brackets.

## Review

Two-reviewer cycle (Opus + Sonnet) before push. Both verdicts: OK. Feedback (extra tests for `identity` / `multi_repeat` / single-base `dup`, CHANGELOG entry, comment alignment) addressed in this commit.

## Out of scope

- Parser-side `t` policy for `r.` (whether to reject `t` on input or canonicalize) — tracked separately as #282.

## Test plan

- [x] `cargo nextest run --features dev` — 4768 pass / 0 fail / 51 skipped.

Closes #276. Follow-up to #106 / #107.

### DIFF
--- a/src/hgvs/edit.rs
+++ b/src/hgvs/edit.rs
@@ -72,10 +72,19 @@ impl Base {
         self as u8
     }
 
-    /// Convert to lowercase ASCII character (for RNA display).
+    /// Convert to lowercase ASCII character for RNA `Display`.
+    ///
+    /// Per HGVS v21.0, the RNA nucleotide alphabet is `a/c/g/u`. Any
+    /// thymine byte that reaches an RNA emission path (lenient `T`/`t`
+    /// input on `r.`, cross-coordinate translation, or canonicalization
+    /// that surfaces an unseen `T`) is translated to `u` here, not just
+    /// case-folded. See issue #276.
     #[inline]
     pub fn to_lowercase_char(self) -> char {
-        (self as u8).to_ascii_lowercase() as char
+        match self {
+            Base::T => 'u',
+            _ => (self as u8).to_ascii_lowercase() as char,
+        }
     }
 }
 
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -3636,15 +3636,13 @@ mod tests {
         assert_eq!(format!("{}", result), "NM_000088.3:c.2T>C");
 
         // r. twin: same collapse on the RNA path. NM_000088.3 has cds_start = 1,
-        // so r.1_3 maps to the same ATG run and the residual is at r.2.
-        // The substitution edit Display lowercases the reference byte but does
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #280, opened 2026-05-17
normalize: RefSeqMismatch.corrected flag is hardcoded for Repeat consistency mismatches (follow-up to #215)


nh13:
PR #215's new `RefSeqMismatch` warning sets `corrected: true` to match the existing `sub`/`del`/`dup` convention. But for a `Repeat` consistency mismatch the normalizer does not actually correct the description — it passes through verbatim.

## Source

PR #215 § *Follow-up*:

> The `corrected: true` flag on the new `RefSeqMismatch` warning is hardcoded (matches the existing sub/del/dup convention), but for a Repeat consistency mismatch the normalizer cannot and does not correct the description — it passes through verbatim. This is a pre-existing flag-semantics quirk shared with all `RefSeqMismatch` warnings and is left untouched here.

## Proposed

Decide whether `corrected` should reflect actual correction (`false` for `Repeat`, `true` for `sub`/`del`/`dup` when bases were rewritten), or whether the field should be redefined / dropped. Either path is a small behavior change; this issue tracks the decision.

## Cross-ref

PR #215. Touches the broader `RefSeqMismatch` warning shape — pre-existing across all edit kinds, not just `Repeat`.

PR #296, opened 2026-05-17
feat(normalize): MultiRepeat partial validation for non-Exact counts (closes #279)

### LINKED ISSUE #279: normalize(repeat): partial-validation mode for MultiRepeat with Range/Unknown counts (follow-up to #215)
PR #215 (closes #214) added `MultiRepeat` validation that fires only when every unit has `RepeatCount::Exact(_)`. Mixed shapes with `Range` or `Unknown` counts are skipped.

## Source

PR #215 § *Follow-up*:

> The `MultiRepeat` validation arm fires only when every unit has `RepeatCount::Exact(_)`. Validating mixed-repeat shapes with Range / Unknown counts is possible but the expected total length is ambiguous; a follow-up could carve out a "partial-validation" mode that checks bases for the known-Exact prefix.

## Proposed

Carve out a partial-validation mode that validates the base sequence for the known-`Exact` prefix and skips the ambiguous tail. The base-mismatch surface is the same as the existing `RefSeqMismatch` path; only the length check is skipped.

## Cross-ref

PR #215, #214 (closed). Fits #81 § B5 follow-up.

nh13:
## Summary

PR #215 (closing #214) added MultiRepeat base-sequence validation that fired only when every unit had `RepeatCount::Exact(_)`. Mixed shapes with `Range` or `Unknown` counts were skipped entirely because the expected total length is ambiguous.

This extends the validation arm with a partial-validation mode: walk units left-to-right, build the deterministic prefix from the leading `Exact` runs, stop at the first non-`Exact` unit, and validate the prefix bases against the reference. The ambiguous tail is skipped.

**Behavior decisions:**

- **First unit non-Exact** — no known prefix, skipped entirely (matches PR #215's existing skip path).
- **Interleaved `Exact` / non-`Exact` / `Exact` / …** — walker stops at the first non-Exact unit. Resuming after the ambiguous unit would require deterministic offset tracking the range/unknown count can't provide.
- **Reference span shorter than declared Exact prefix** — surfaces as `RefSeqMismatch` rather than silently truncating.
- **`RefSeqMismatch.corrected: true`** — flag-semantics quirk preserved as-is. The fix for that quirk is tracked in #280.

## Tests

`tests/issue_279_multirepeat_partial_validation.rs` — 17 tests across 7 sections covering all-Exact regression, Exact prefix + Range tail, Exact prefix + Unknown tail, first-unit-non-Exact skip, multi-Exact prefix, prefix boundary tightness, and prefix longer than span.

## Review

Two-reviewer cycle (Opus + Sonnet). Both verdicts: OK. Feedback (boundary-tightness test, short-span coverage, CHANGELOG, naming alignment) addressed in this commit.

## Note on commit signing

1Password biometric failed three times; commit is `--no-gpg-sign` per project rules. Will re-sign when biometric is available.

## Test plan

- [x] `cargo nextest run --features dev` — 4779 pass / 0 fail / 51 skipped.
- [x] `cargo check --features python` — clean.

Closes #279. Follow-up to #215 / #214 (#81 § B5).

### DIFF
--- a/src/normalize/validate.rs
+++ b/src/normalize/validate.rs
@@ -208,9 +208,25 @@ fn validate_repeat_tract(
 ///
 /// Multi-repeat notation describes the reference structure exactly: the
 /// per-unit counts must reproduce the reference bases at `[start, end]`
-/// when expanded and concatenated. Skips validation when any unit has a
-/// non-Exact count (Range / MinUncertain / MaxUncertain / Unknown) —
-/// the expected length is ambiguous in those cases.
+/// when expanded and concatenated.
+///
+/// Validation mode is chosen by inspecting the per-unit counts:
+///
+/// - **All units `Exact`** — full validation. The expanded
+///   concatenation has a deterministic length; we check both the
+///   span-vs-sum length invariant and the base-by-base content match.
+/// - **Leading `Exact` prefix followed by a non-`Exact` unit (Range /
+///   MinUncertain / MaxUncertain / Unknown)** — *partial validation*
+///   (issue #279). The prefix bases at `[start, start + prefix_len)`
+///   are known; we validate just those against the reference and
+///   skip the length check and the ambiguous tail. A prefix mismatch
+///   surfaces as `RefSeqMismatch`.
+/// - **First unit non-`Exact`** — no validation. There is no known
+///   prefix to check; the description is accepted unchanged.
+///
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #166, opened 2026-05-09
fix(normalize): recognize revcomp inv sub-spans within delins

nh13:
## Summary

Closes #160. A `delins` whose span contains an inv-eligible sub-span (length ≥ 2, alt = revcomp(ref)) now decomposes to the spec-canonical sub-edit sequence `[…; inv; …]`, applied symmetrically to user-typed delins inputs and to delins synthesized by the cis-allele merge step. Implements the inv branch of the spec's edit-priority rule (`sub > del > inv > dup > ins`); the symmetric sub-only branch (rewriting non-inv multi-sub delins to `[X>A; B>Y]`) is deliberately scoped out and tracked in #165 (added as #81 item A10) — see the in-code TODO and the linked issue for the spec-ambiguity rationale.

| Input | Before | After |
|---|---|---|
| `g.[1092G>C;1093G>C]` over `GG` | `g.1092_1093delinsCC` | `g.1092_1093inv` |
| `g.[1150T>G;1151C>A;1152C>G]` over `TCC` | `g.1150_1152delinsGAG` | `g.[1150_1151inv;1152C>G]` |
| `g.1150_1152delinsGAG` over `TCC` (user-typed) | `g.1150_1152delinsGAG` | `g.[1150_1151inv;1152C>G]` |

## Implementation

**`rules::decompose_delins_inv`** (new, in `src/normalize/rules.rs`) does a left-to-right longest-greedy scan over `(ref, alt)` and emits a `Vec<DelinsSubedit>` (`Substitution` / `Inversion` / `IdentityAt`). Trigger commits the result only when at least one `Inversion` is emitted — preserving the issue #79 codon-frame `c.` merge (the synthesized middle base mathematically excludes a length-2 inv across it).

**`Normalizer::apply_inv_split`** (new, in `src/normalize/mod.rs`) is coord-system-agnostic. `fetch_ref_for_inv_split` resolves the per-variant reference window: genomic via `provider.get_sequence`; CDS via `cds_start + tx.sequence`; n./r. directly via `tx.sequence`. RNA `r.` alts are normalized `U → T` before comparison so revcomp matches against transcript ref bytes that contain `T`; emitted `Substitution` sub-edits preserve the original `Base` (which may be `Base::U`).

**Wiring**:
- Each per-coord-system `normalize_*` (g/c/n/r) calls `apply_inv_split` after the single `new_variant` is built.
[...]

PR #312, opened 2026-05-18
fix(mock): disambiguate genomic vs transcript lookups when ids collide (closes #311)

### LINKED ISSUE #311: Normalizer.normalize() incorrectly reduces a delins to a single SNV when transcript ID starts with chromosome name
## Description

When normalizing a `g.` variant (either an allele-list of adjacent substitutions or an already-correct `delins`), `Normalizer.normalize()` produces an incorrect result that drops variants — **if** the transcript `id` in the reference JSON starts with (case-sensitive) the `chromosome` name.

**ferro-hgvs version:** 0.6.0 (regression from 0.5.0; correct in 0.4.1)

## Minimal reproduction

```python
import json, tempfile
import ferro_hgvs

# Synthetic 1102bp CDS: 909bp poly-A leader + 93bp coding segment + 100bp poly-T trailer.
# Positions 1000-1002 (genomic, 1-based with genomic_start=91) → seq[909:912] = "GAC".
SEGMENT = "GACCGGCGGTCTGCAAGTCTCCACCTTCCGAAGCTATCCATAACCGGAACCTATGACCTAAAATCAGTTCTGGGTCAGCTCGGTATCACGAAG"
SEQUENCE = "A" * 909 + SEGMENT + "T" * 100
assert SEQUENCE[909:912] == "GAC"


def make_reference(transcript_id: str, chromosome: str) -> str:
    ref = {
        "genome_build": "GRCh38",
        "transcripts": [{
            "id": transcript_id,
            "chromosome": chromosome,
            "strand": "+",
            "genomic_start": 91,
            "genomic_end": 1192,
            "cds_start": 1,
            "cds_end": 1102,
            "exons": [{"number": 1, "start": 1, "end": 1102,
                       "genomic_start": 91, "genomic_end": 1192}],
            "sequence": SEQUENCE,
            "genome_build": "GRCh38",
        }],
        "version": "1.0",
    }
    f = tempfile.NamedTemporaryFile(mode="w", suffix=".json", delete=False)
    json.dump(ref, f)
    f.close()
    return f.name


# BUG: transcript ID starts with chromosome name → incorrect normalization
ref = make_reference("chr1-gene.1", "chr1")
normalizer = ferro_hgvs.Normalizer(reference_json=ref)

result = normalizer.normalize("chr1:g.[1000G>A;1001A>C;1002C>G]")
assert result == "chr1:g.1000_1002delinsACG", f"Got: {result}"
# AssertionError: Got: chr1:g.1001A>C

# Even the already-correct delins form is broken:
result2 = normalizer.normalize("chr1:g.1000_1002delinsACG")
assert result2 == "chr1:g.1000_1002delinsACG", f"Got: {result2}"
# AssertionError: Got: chr1:g.1001A>C


# WORKS: transcript ID does NOT start with chromosome name
ref = make_reference("NM_000001.1", "chr1")
normalizer = ferro_hgvs.Normalizer(reference_json=ref)

result = normalizer.normalize("chr1:g.[1000G>A;1001A>C;1002C>G]")
assert result == "chr1:g.1000_1002delinsACG"  # ✓ passes
```

## Observations

- The output `chr1:g.1001A>C` implies only position 1001 changed (A→C), but the ref at 1000–1002 is `GAC` and the intended alt is `ACG` — applying just `1001A>C` yields `GCC`, not `ACG`. The normalized form is not equivalent to the input.
- The bug triggers when the transcript `id` is a **case-sensitive prefix** of (or equal to) the `chromosome` name (e.g., `"chr1"`, `"chr1-gene.1"`, `"chr1xyz"`). IDs like `"NM_000001.1"` or `"Chr1-gene.1"` (different case) do **not** trigger it.
- The same input normalized correctly in v0.4.1; the regression was introduced in v0.5.0.

nh13:
Closes #311.

## Summary

Normalizing a `g.` variant against a `MockProvider` whose only transcript has an `id` that is a case-sensitive prefix of (or equal to) the `chromosome` name silently returned the transcript's bases for a chromosome-coordinate lookup. The wrong reference bases drove the post-merge trim down to a single residual SNV — `chr1:g.[1000G>A;1001A>C;1002C>G]` and the already-canonical `chr1:g.1000_1002delinsACG` both collapsed to `chr1:g.1001A>C`.

The fix lives entirely in `MockProvider`. Two failure modes are addressed:

1. **`get_transcript`'s unversioned-prefix fallback used `key.starts_with(base_id)`**, so a lookup for `chr1` matched any transcript whose id literally began with `chr1` (e.g. `chr1-gene.1`). The fallback now requires base-id equality at the version-dot boundary, preserving the intended `NM_000088` ↔ `NM_000088.3` resolution while rejecting arbitrary prefix matches. The same fix is applied to `get_protein_sequence`.

2. **`get_sequence` routed every lookup through `get_transcript` first**, so a chromosome-coordinate query against an id that happened to match a transcript key (exactly or, pre-fix, by prefix) silently slid into the transcript path and read genomic coordinates as transcript-relative indices. `get_sequence` now short-circuits to `get_genomic_sequence` when the id is a known contig — present either in `genomic_sequences` or as the `chromosome` field on any transcript. Genuinely missing chromosome data still surfaces as `Err`, which `normalize_genome` already falls back to canonicalize-only.

## Where the regression came from

Bisecting v0.4.1..v0.5.0 with a Rust port of the issue's reproducer points at **3b33dee** (#166, `fix(normalize): recognize revcomp inv sub-spans within delins`). That PR added a post-merge re-canonicalization step that re-fetches the genomic reference span for the merged delins. The provider-level prefix-collision bug already existed in v0.4.1 — its provider code is byte-identical to v0.6.0 today — but it was never exercised because nothing pre-#166 re-fetched the genomic span on a post-merge delins. With the new re-canonicalization, the buggy fetched ref drives the trim and collapses the delins.

## Why no regression test caught it

Every existing synthetic fixture pairs disjoint namespaces — genomic contigs are named `NC_TEST.1`, transcripts `NM_TEST.1` / `NR_TEST.1`, and CDS fixtures use a separate `chr_synth` contig. No fixture pairs a transcript id with a prefix-colliding chromosome name, and no fixture sets `chromosome` on a transcript that is also addressable by its id in a `g.` variant. The class of collision the issue describes is therefore invisible to the existing matrix. The new test file exercises exactly that pairing.

## Tests

`tests/issue_311_transcript_prefix_chromosome.rs` — nine cases pinning both layers:

End-to-end normalizer behavior (5 cases):
- transcript id `chr1-gene.1` (prefix collision, the issue's reproducer)
- transcript id `chr1` (exact-name collision)
- transcript id `Chr1-gene.1` (case-sensitivity boundary control)
- transcript id `NM_000001.1` (no-collision control)
- already-canonical delins round-trip under the prefix collision

Provider-level lookup contract (4 cases):
- `get_transcript("chr1")` does not resolve to `chr1-gene.1`
- `get_transcript("NM_000088")` still resolves to `NM_000088.3` (unversioned ↔ versioned)
- `get_transcript("NM_000088.3")` still resolves exactly
- `get_transcript("NM_0000")` does not resolve to `NM_000088.3` (non-version-boundary prefix)

All five end-to-end cases pass at v0.4.1 and fail at v0.5.0..main; all nine pass with the fix applied.

## Test plan

- [x] `cargo nextest run --features dev` — 5053 / 5053 pass, 51 skipped.
- [x] `cargo check --features python` — clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - *(parser)* accept `?con<src>` and `?copy<N>` at unknown position, consistent with `?del`/`?dup`/`?ins`/`?inv` (closes #286)
 - *(fasta)* require version-boundary equality in `MmapFastaProvider::get_transcript`'s unversioned-prefix fallback (closes #314)
+- *(fasta)* route `FastaProvider::get_sequence` for known contigs through the FASTA path so a transcript registered with a chromosome-colliding id no longer wins over the genomic index (closes #315)
 
 ## [0.6.0](https://github.com/fulcrumgenomics/ferro-hgvs/compare/v0.5.0...v0.6.0) - 2026-05-16
 
--- a/src/reference/fasta.rs
+++ b/src/reference/fasta.rs
@@ -219,6 +219,22 @@ impl FastaProvider {
     pub fn add_transcript(&mut self, transcript: Transcript) {
         self.transcripts.insert(transcript.id.clone(), transcript);
     }
+
+    /// True when `id` is declared as a chromosome/contig name — either
+    /// resolves against the FASTA `.fai` index (directly or via an alias
+    /// chain) or appears as the `chromosome` field on a registered
+    /// transcript. Used by `get_sequence` to disambiguate genomic
+    /// lookups from transcript-id lookups when the two namespaces
+    /// collide (e.g. a transcript whose id is `chrN`; see #315).
+    fn is_known_contig(&self, id: &str) -> bool {
+        let resolved = self.resolve_name(id);
+        if self.index.contains_key(&resolved) {
+            return true;
+        }
+        self.transcripts
+            .values()
+            .any(|tx| tx.chromosome.as_deref() == Some(id))
+    }
 }
 
 impl ReferenceProvider for FastaProvider {
@@ -230,6 +246,17 @@ impl ReferenceProvider for FastaProvider {
     }
 
     fn get_sequence(&self, id: &str, start: u64, end: u64) -> Result<String, FerroError> {
+        // If the requested id is registered as a contig — either present
+        // in the FASTA index (directly or via an alias) or referenced as
+        // the `chromosome` field on a registered transcript — resolve it
+        // strictly against the FASTA path. Without this dispatch, a
+        // transcript whose id happens to match a contig name would
+        // short-circuit the lookup and silently interpret genomic
+        // coordinates as transcript-relative indices (#315 / #311).
+        if self.is_known_contig(id) {
+            return self.get_fasta_sequence(id, start, end);
+        }
+
         // First try as transcript ID. Only return a hit when the transcript
         // has cached sequence bases; coordinate-only transcripts fall through
         // to the FASTA index lookup below.
--- a/src/reference/mock.rs
+++ b/src/reference/mock.rs
@@ -260,6 +260,20 @@ impl MockProvider {
     pub fn all_transcripts(&self) -> impl Iterator<Item = &Transcript> {
         self.transcripts.values()
     }
+
+    /// True when `id` is declared as a chromosome/contig name — either an
+    /// entry in `genomic_sequences` or the `chromosome` field on any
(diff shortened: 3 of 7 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #313, opened 2026-05-18
fix(projector): protein prediction for non-RefSeq transcripts + drop (GENE) selector from p. Display

### LINKED ISSUE #310: VariantProjector: support protein prediction for non-RefSeq transcript IDs and omit gene symbol from selector position
## Summary

Two related observations about `VariantProjector` protein prediction:

1. **Protein prediction gated on NM_ prefix** — `project()` returns `p_name = None` when the transcript ID does not start with `NM_`, even when the reference JSON contains a valid CDS and coding sequence. No error or warning is raised at construction time.

2. **Gene symbol in selector position** — Output p. names use the format `NP_TEST.1(TESTGENE):p.(Arg2Ser)`, placing the gene symbol in the parenthesized selector position. Two relevant passages from the HGVS spec:

   - [`docs/background/refseq.md` line 33](https://hgvs-nomenclature.org/stable/background/refseq/): *"accepted specifications include transcripts (NM_004006.2) and proteins (NP_003997.1). Gene symbols should **not** be used as specification."*
   - [`docs/syntax.yaml` lines 119–128](https://hgvs-nomenclature.org/stable/recommendations/summary/): The protein variant syntax is defined as `sequence_identifier ":" coordinate_type "." aa_position alternate_base` — no parenthesized selector appears. The `"(" transcript_identifier ")"` form is only defined for genomic-to-transcript projection contexts (line 213: `sequence_identifier "(" transcript_identifier "):c."`).

Related: #69 (parser gate on RefSeq-style accessions for selectors).

## Reproduction

```python
import json, tempfile
import ferro_hgvs

fixture = {
    "transcripts": [{
        "id": "MY_GENE-gene.1",
        "gene_symbol": "MY_GENE",
        "strand": "+",
        "sequence": "ATGCGCTAA",
        "cds_start": 1,
        "cds_end": 9,
        "exons": [{"number": 1, "start": 1, "end": 9,
                   "genomic_start": 1000, "genomic_end": 1008}],
        "chromosome": "chr1",
        "genomic_start": 1000,
        "genomic_end": 1008,
    }],
    "genomic_sequences": {"chr1": "N" * 999 + "ATGCGCTAA" + "N" * 100},
}
path = tempfile.NamedTemporaryFile(suffix=".json", mode="w", delete=False)
json.dump(fixture, path); path.close()

projector = ferro_hgvs.VariantProjector(reference_json=path.name)
result = projector.project("chr1:g.1003C>A", "MY_GENE-gene.1")

assert result.c_name is not None   # ✅ c.4C>A
assert result.is_intronic is False  # ✅
assert result.is_utr is False       # ✅
assert result.p_name is None        # ← expected p.(Arg2Ser)
```

Changing `"id"` to `"NM_TEST.1"` produces:
```
NP_TEST.1(MY_GENE):p.(Arg2Ser)
```

Which illustrates item 2: `MY_GENE` is a gene symbol occupying the selector position.

## Possible directions

**Item 1 — extend protein prediction to arbitrary transcript IDs:**
- Use the transcript ID directly as the protein accession prefix (e.g. `MY_GENE-gene.1:p.(Arg2Ser)`)
- Accept an optional `protein_id` field in the reference JSON
- Omit the accession prefix entirely and emit just `p.(Arg2Ser)`

**Item 2 — align selector output with the HGVS syntax spec:**
- Omit the gene symbol from the selector position (spec-compliant: `NP_003997.1:p.(Arg2Ser)`)
- Or document the current behavior as an intentional extension of the spec

## Context

This surfaces with references generated by `ferro convert-gff`, which produces transcript IDs like `{gene}-gene.1`. The projector returns correct g. and c. names for these references, but p. prediction is gated on the NM_ prefix.

## Workaround

Rewrite the transcript ID to a synthetic `NM_*` value before constructing the projector, then strip the accession prefix from p. names downstream.

## Version

v0.6.0 (`7641d42`)

nh13:
## Summary

Two related fixes from #310.

**1. Protein prediction for non-RefSeq transcript IDs.** `VariantProjector.project()` previously returned `p_name = None` whenever the transcript ID did not start with `NM_` or `XM_` and no `cdot.protein` value was available — even when the reference had a valid CDS and coding sequence. This surfaced for references produced by `ferro convert-gff`, which emits IDs of the form `{gene}-gene.1`.

This PR adds an optional `protein_id` field to the ferro reference-JSON transcript schema (and the in-memory `Transcript` struct) and replaces the projector's two-branch logic with a four-step priority chain:

1. `cdot.protein` (unchanged for cdot-based references)
2. `transcript.protein_id` (new, plumbed into `CdotTranscript.protein` by `CdotMapper::from_transcripts*`)
3. `NM_*` → `NP_*` / `XM_*` → `XP_*` inference (unchanged)
4. Transcript ID itself used as the `p.` accession prefix (new fallback)

The HGVS protein grammar requires a `sequence_identifier`, so the transcript-ID fallback is what guarantees that protein prediction is never silently dropped for an arbitrary ID format.

**2. Drop `(GENE)` selector from `p.` Display.** `ProteinVariant::Display` previously emitted `NP_*(GENE):p.(...)`. The HGVS protein grammar ([syntax.yaml 119–128](https://hgvs-nomenclature.org/stable/recommendations/summary/)) defines protein variants as `sequence_identifier ":" coordinate_type "." aa_position alternate_base` with no parenthesized selector position; the `accession(selector):c.` form (syntax.yaml 213) is reserved for genomic-to-transcript projection on `c.`/`n.`, not `p.` The reporter also cites HGVS guidance that gene symbols should not be used as specification.

`ProteinVariant::Display` now writes `accession:p....` without the selector. The `gene_symbol` field stays on the struct for the public projection API, and the parser remains permissive on the legacy `NP_*(SELECTOR):p.` form so strings produced by older ferro versions still parse cleanly. Only canonical `p.` output changes; other coordinate systems (`c.`/`n.`/`g.`/`r.`/`m.`) are untouched.

Also fixes a latent bug in `project::accession::parse_accession` where Ensembl-style IDs (and any accession without an underscore) were rendered with a spurious trailing underscore through `Display` — the helper now splits at the alpha→digit boundary and sets `ensembl_style=true`.

## Behavior table

| Reference shape                                | Old `p_name`                       | New `p_name`                       |
| ---------------------------------------------- | ---------------------------------- | ---------------------------------- |
| RefSeq `NM_*` + `cdot.protein`                 | `NP_*(GENE):p.(...)` if gene set   | `NP_*:p.(...)`                     |
| RefSeq `NM_*`, no cdot                         | `NP_*(GENE):p.(...)` if gene set   | `NP_*:p.(...)`                     |
| RefSeq `XM_*`                                  | `XP_*(GENE):p.(...)` if gene set   | `XP_*:p.(...)`                     |
| Non-RefSeq ID + new `protein_id`               | `None`                             | `<protein_id>:p.(...)`             |
| Non-RefSeq ID, no `protein_id`                 | `None`                             | `<transcript_id>:p.(...)`          |
| Non-coding / intronic / UTR                    | `None`                             | `None` (unchanged)                 |

## Test plan
- [x] `cargo nextest run --features dev` — 5048 tests pass.
- [x] `pytest tests/python/` — 159 tests pass.
- [x] `ruff check python/ tests/python/` clean.
- [x] New `tests/issue_310_projector_non_refseq.rs` covers: protein_id-driven accession; transcript-ID fallback; RefSeq inference still works with the new `:p.` form; parser still accepts the legacy `NP_*(GENE):p.` form and round-trips `gene_symbol` on the struct.
- [x] New `TestIssue310NonRefSeqProtein` in `tests/python/test_variant_projection.py` exercises the same scenarios end-to-end through the Python binding.
- [x] Updated in-tree assertions that pinned the legacy `NP_*(GENE):p.` and `LRG_*p*(GENE):p.` Display output to the new spec-compliant form.

Closes #310.

### DIFF
--- a/docs/transcripts_json_schema.md
+++ b/docs/transcripts_json_schema.md
@@ -59,6 +59,7 @@ Each transcript object has the following fields:
 | `mane_status` | string | MANE designation: `"MANE_Select"`, `"MANE_Plus_Clinical"`, or omitted if not MANE |
 | `refseq_match` | string | For Ensembl transcripts: matched RefSeq accession |
 | `ensembl_match` | string | For RefSeq transcripts: matched Ensembl accession |
+| `protein_id` | string | Protein accession (e.g. `"NP_000079.2"` or `"MYGENE-protein.1"`). Used as the `p.` accession for c. → p. projection. When omitted, the projector infers `NP_*` from `NM_*` and `XP_*` from `XM_*`; if neither applies, it falls back to using the transcript ID itself as the `p.` accession so protein prediction is never silently dropped. |
 
 ## Exon Object
 
--- a/src/convert/coding.rs
+++ b/src/convert/coding.rs
@@ -121,6 +121,7 @@ mod tests {
             mane_status: ManeStatus::default(),
             refseq_match: None,
             ensembl_match: None,
+            protein_id: None,
             exon_cigars: Vec::new(),
             cached_introns: OnceLock::new(),
         }
@@ -191,6 +192,7 @@ mod tests {
             mane_status: ManeStatus::default(),
             refseq_match: None,
             ensembl_match: None,
+            protein_id: None,
             exon_cigars: Vec::new(),
             cached_introns: OnceLock::new(),
         };
--- a/src/convert/mapper.rs
+++ b/src/convert/mapper.rs
@@ -727,6 +727,7 @@ mod tests {
             mane_status: ManeStatus::default(),
             refseq_match: None,
             ensembl_match: None,
+            protein_id: None,
             exon_cigars: Vec::new(),
             cached_introns: OnceLock::new(),
         }
@@ -850,6 +851,7 @@ mod tests {
             mane_status: ManeStatus::default(),
             refseq_match: None,
             ensembl_match: None,
+            protein_id: None,
             exon_cigars: Vec::new(),
             cached_introns: OnceLock::new(),
         }
@@ -879,6 +881,7 @@ mod tests {
             mane_status: ManeStatus::default(),
             refseq_match: None,
             ensembl_match: None,
+            protein_id: None,
             exon_cigars: Vec::new(),
             cached_introns: OnceLock::new(),
         }
@@ -1000,6 +1003,7 @@ mod tests {
             mane_status: ManeStatus::default(),
             refseq_match: None,
             ensembl_match: None,
+            protein_id: None,
             exon_cigars: Vec::new(),
             cached_introns: OnceLock::new(),
         }
@@ -1241,6 +1245,7 @@ mod intronic_debug_tests {
             mane_status: ManeStatus::None,
             refseq_match: None,
             ensembl_match: None,
+            protein_id: None,
             exon_cigars: Vec::new(),
             cached_introns: OnceLock::new(),
         }
--- a/src/convert/noncoding.rs
+++ b/src/convert/noncoding.rs
@@ -314,6 +314,7 @@ mod tests {
             mane_status: ManeStatus::default(),
             refseq_match: None,
             ensembl_match: None,
+            protein_id: None,
             exon_cigars: Vec::new(),
             cached_introns: OnceLock::new(),
         }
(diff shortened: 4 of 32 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #316, opened 2026-05-18
tests/common/synthetic.rs has no adversarial-namespace fixture for transcript-id ↔ chromosome-name collisions


nh13:
Follow-up to #311 / #312.

\`tests/common/synthetic.rs\` is the shared builder for \`MockProvider\` fixtures across coordinate systems. By construction it uses disjoint namespaces — genomic contigs are named \`NC_TEST.1\`, transcripts \`NM_TEST.1\` / \`NR_TEST.1\`, and CDS fixtures pin transcripts onto a separate \`chr_synth\` contig. The synthetic matrix is therefore blind to the entire class of bugs that arise when a transcript id collides with a chromosome name on the same provider — which is exactly the failure mode #311 exhibited and #312 fixed.

Concretely, the matrix would have caught both #312 and the latent issues in MmapFastaProvider and FastaProvider (sibling issues sharing this tag) had it included a fixture where:

1. A transcript's id is a case-sensitive prefix of (or equal to) its \`chromosome\` field.
2. The transcript is registered on the same provider as the contig the \`g.\` variant addresses.

## Proposal

Add an \`adversarial_namespace\` (or similarly-named) builder to \`tests/common/synthetic.rs\` that:

- Accepts a transcript id and a chromosome name as inputs, with no requirement that they be disjoint.
- Materializes a \`MockProvider\` with the transcript registered, the chromosome populated either as a top-level \`genomic_sequence\` or only as the transcript's \`chromosome\` field (selectable), and a synthetic sequence laid out so that genomic and transcript coordinates pick out different bases at the same numerical position (the wrong-frame failure is observable only when the two frames disagree).
- Exposes the same \`normalize_to_string\` entry point the rest of the matrix uses.

Once the builder lands, the regression tests that #312 added to \`tests/issue_311_transcript_prefix_chromosome.rs\` can move (or have analogues moved) onto the shared builder, and the sibling provider issues can reuse it instead of re-inlining the fixture each time.

## Out of scope

This issue is the test-scaffold work only. The actual bug fixes for \`MmapFastaProvider\` and \`FastaProvider\` are tracked separately.

nh13:
PR #319 merged on 2026-05-19 with `Closes #316` but auto-close didn't fire (PR was stacked). Work shipped — closing manually.

PR #317, opened 2026-05-18
fix(fasta): require version-boundary equality in MmapFastaProvider transcript lookup (closes #314)

### LINKED ISSUE #314: MmapFastaProvider::get_transcript has the same prefix-collision bug as MockProvider pre-#311
Follow-up to #311 / #312.

\`MmapFastaProvider::get_transcript\` at \`src/reference/fasta.rs:833-848\` carries the exact same \`key.starts_with(base_id)\` fallback that \`MockProvider::get_transcript\` had before #312. The intent is the same — bridge an unversioned query (\`NM_000088\`) to a versioned key (\`NM_000088.3\`) — but the implementation matches any key that literally starts with the requested string, so a lookup for \`chr1\` resolves to a transcript whose id is e.g. \`chr1-gene.1\`.

\`\`\`rust
let base_id = id.split('.').next().unwrap_or(id);
for (key, tx) in &self.transcripts {
    if key.starts_with(base_id) {
        return Ok(tx.clone());
    }
}
\`\`\`

This is the same collision class fixed in #312 for \`MockProvider\`. \`MmapFastaProvider\` is part of the \`dev\` feature set and is the production read path for large FASTA files, so any pipeline that loads a transcript whose id was synthesized from a chromosome+gene prefix (a common shape in GFF-derived dumps) is exposed.

## Reproduction sketch

Register a transcript with \`id = \"chr1-gene.1\"\` against an \`MmapFastaProvider\`, then call \`get_transcript(\"chr1\")\`. Pre-fix: returns the \`chr1-gene.1\` transcript. Expected: \`ReferenceNotFound\`.

## Fix

Mirror the #312 fix: require base-id equality at the version-dot boundary.

\`\`\`rust
if key.split('.').next().unwrap_or(key) == base_id {
    return Ok(tx.clone());
}
\`\`\`

## Coverage gap

No existing test fixture pairs a transcript id with a prefix-colliding chromosome name on the \`MmapFastaProvider\` path — see #311's coverage-gap analysis. A regression test should be added alongside the fix.

nh13:
Closes #314. Follow-up to #311 / #312.

## Summary

\`MmapFastaProvider::get_transcript\` at \`src/reference/fasta.rs:833-848\` carried the same permissive \`key.starts_with(base_id)\` fallback that \`MockProvider\` had before #312, so a lookup for a chromosome accession (e.g. \`chr1\`) silently resolved to any transcript whose id literally began with that string (e.g. \`chr1-gene.1\`). This is the same collision class fixed in #312, just on a different provider.

The fix mirrors #312: the unversioned-prefix fallback now matches a stored key only when the segment before the version dot equals the requested id. \`NM_000088\` still resolves to \`NM_000088.3\`; \`chr1\` no longer resolves to \`chr1-gene.1\`.

## Why no existing test caught it

No fixture pairs a transcript id with a prefix-colliding chromosome name on the \`MmapFastaProvider\` path — the same coverage gap #312 called out for the synthetic matrix. The broader fixture work is tracked in #316. This PR adds a targeted regression test alongside the fix.

## Tests

\`tests/issue_314_mmap_prefix_collision.rs\` — four cases, gated behind the \`mmap\` feature:

- \`get_transcript("chr1")\` does not resolve to \`chr1-gene.1\`
- \`get_transcript("NM_000088")\` still resolves to \`NM_000088.3\` (unversioned ↔ versioned)
- \`get_transcript("NM_000088.3")\` still resolves exactly
- \`get_transcript("NM_0000")\` does not resolve to \`NM_000088.3\` (non-version-boundary prefix)

## Test plan

- [x] \`cargo nextest run --features dev\` — 5048 / 5048 pass, 51 skipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - *(parser)* accept `?con<src>` and `?copy<N>` at unknown position, consistent with `?del`/`?dup`/`?ins`/`?inv` (closes #286)
+- *(fasta)* require version-boundary equality in `MmapFastaProvider::get_transcript`'s unversioned-prefix fallback (closes #314)
 
 ## [0.6.0](https://github.com/fulcrumgenomics/ferro-hgvs/compare/v0.5.0...v0.6.0) - 2026-05-16
 
--- a/src/reference/fasta.rs
+++ b/src/reference/fasta.rs
@@ -836,10 +836,14 @@ impl ReferenceProvider for MmapFastaProvider {
             return Ok(tx.clone());
         }
 
-        // Try without version
+        // Try without version: only match a stored key whose base id (the
+        // segment before the version dot) equals the requested id. A bare
+        // `starts_with` would also match unrelated keys that merely share
+        // a prefix (e.g. `chr1-gene.1` for a `chr1` lookup), causing a
+        // genomic accession to be resolved as a transcript — see #314.
         let base_id = id.split('.').next().unwrap_or(id);
         for (key, tx) in &self.transcripts {
-            if key.starts_with(base_id) {
+            if key.split('.').next().unwrap_or(key) == base_id {
                 return Ok(tx.clone());
             }
         }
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #318, opened 2026-05-18
fix(fasta): route known-contig lookups in FastaProvider through the FASTA path (closes #315)

### LINKED ISSUE #315: FastaProvider::get_sequence silently returns transcript bases when an id is registered as both a transcript and a contig
Follow-up to #311 / #312.

\`FastaProvider::get_sequence\` at \`src/reference/fasta.rs:232-255\` routes every lookup through the transcript registry before consulting the FASTA index:

\`\`\`rust
if let Some(transcript) = self.transcripts.get(id) {
    if let Some(seq) = transcript.sequence.as_deref() {
        // returns transcript bases
    }
}
self.get_fasta_sequence(id, start, end)
\`\`\`

The lookup is exact-match (no \`starts_with\` fallback), so realistic accession names cannot collide. But if a caller ever registers a transcript with \`id = \"chrN\"\` alongside a FASTA index that also contains \`chrN\`, the transcript wins and \`get_sequence(\"chrN\", ...)\` silently reads transcript-relative bases for a genomic-coordinate query — the same wrong-coordinate-frame bug that #311 hit in \`MockProvider\`, just through a different dispatch.

This is latent today because production loaders don't synthesize transcripts named after chromosomes, but the ordering is structurally identical to the pre-#312 \`MockProvider\` dispatch and would manifest the same way under any future fixture or pipeline that does. \`CachedFastaProvider\` wraps \`FastaProvider\` and would propagate the same bug.

## Fix

Mirror the #312 \`MockProvider::get_sequence\` fix:

- Short-circuit to the FASTA-index path when the id is a known contig (present either in the FASTA \`.fai\` index or as the \`chromosome\` field on any registered transcript). 
- Only fall through to the transcript path otherwise.

## Coverage gap

\`tests/common/synthetic.rs\` does not have a fixture that pairs a transcript id with a chromosome name registered on the same provider (see #311). Tracking that fixture work separately in a sibling issue.

nh13:
Closes #315. Follow-up to #311 / #312.

## Summary

\`FastaProvider::get_sequence\` routed every lookup through the transcript registry first:

\`\`\`rust
if let Some(transcript) = self.transcripts.get(id) { /* return transcript bases */ }
self.get_fasta_sequence(id, start, end)
\`\`\`

Exact-match only (no \`starts_with\` fallback), so realistic accession names cannot collide. But if any caller registers a transcript with \`id = \"chrN\"\` alongside a FASTA index that also contains \`chrN\`, the transcript wins and \`get_sequence(\"chrN\", ...)\` silently reads transcript-relative bases for a genomic-coordinate query — the same wrong-coordinate-frame bug #311 hit on \`MockProvider\`.

## Fix

Mirror the #312 \`MockProvider::get_sequence\` fix:

- \`get_sequence\` short-circuits to \`get_fasta_sequence\` when the id is a known contig — either resolves against the FASTA \`.fai\` index (directly or via an alias chain) or appears as the \`chromosome\` field on a registered transcript.
- Genuinely missing chromosome data surfaces as \`Err\`, which \`normalize_genome\` already falls back to canonicalize-only.
- Otherwise treat the id as a transcript accession (no change for that path).

A new \`is_known_contig\` helper on \`FastaProvider\` encapsulates the dispatch.

## Tests

\`tests/issue_315_fasta_known_contig.rs\` — four cases:

- exact id collision: a transcript with id \`chr1\` is registered alongside a FASTA \`chr1\` entry → \`get_sequence(\"chr1\", 0, 3)\` returns FASTA bytes (\`AAA\`), not transcript bases (\`TTT\`).
- soft collision: a transcript declares \`chromosome = \"chr1\"\` but the FASTA index has no \`chr1\` entry → \`get_sequence(\"chr1\", 0, 3)\` Errs cleanly (does not silently return unrelated transcript bases for a different id's transcript bases).
- transcript-id lookup for a non-colliding id keeps returning the transcript's cached sequence.
- FASTA-only contig lookup with no colliding transcript is unaffected.

## Test plan

- [x] \`cargo nextest run --features dev\` — full suite pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - *(parser)* accept `?con<src>` and `?copy<N>` at unknown position, consistent with `?del`/`?dup`/`?ins`/`?inv` (closes #286)
 - *(fasta)* require version-boundary equality in `MmapFastaProvider::get_transcript`'s unversioned-prefix fallback (closes #314)
+- *(fasta)* route `FastaProvider::get_sequence` for known contigs through the FASTA path so a transcript registered with a chromosome-colliding id no longer wins over the genomic index (closes #315)
 
 ## [0.6.0](https://github.com/fulcrumgenomics/ferro-hgvs/compare/v0.5.0...v0.6.0) - 2026-05-16
 
--- a/src/reference/fasta.rs
+++ b/src/reference/fasta.rs
@@ -219,6 +219,22 @@ impl FastaProvider {
     pub fn add_transcript(&mut self, transcript: Transcript) {
         self.transcripts.insert(transcript.id.clone(), transcript);
     }
+
+    /// True when `id` is declared as a chromosome/contig name — either
+    /// resolves against the FASTA `.fai` index (directly or via an alias
+    /// chain) or appears as the `chromosome` field on a registered
+    /// transcript. Used by `get_sequence` to disambiguate genomic
+    /// lookups from transcript-id lookups when the two namespaces
+    /// collide (e.g. a transcript whose id is `chrN`; see #315).
+    fn is_known_contig(&self, id: &str) -> bool {
+        let resolved = self.resolve_name(id);
+        if self.index.contains_key(&resolved) {
+            return true;
+        }
+        self.transcripts
+            .values()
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #319, opened 2026-05-18
test(common): add adversarial-namespace builder for transcript-id ↔ chromosome-name collisions (closes #316)

### LINKED ISSUE #311: Normalizer.normalize() incorrectly reduces a delins to a single SNV when transcript ID starts with chromosome name

### LINKED ISSUE #314: MmapFastaProvider::get_transcript has the same prefix-collision bug as MockProvider pre-#311

### LINKED ISSUE #315: FastaProvider::get_sequence silently returns transcript bases when an id is registered as both a transcript and a contig

### LINKED ISSUE #316: tests/common/synthetic.rs has no adversarial-namespace fixture for transcript-id ↔ chromosome-name collisions

nh13:
Closes #316. Follow-up to #311 / #312.

**Stacks on #312.** Base: \`fix/issue-311-transcript-prefix-chromosome\`. Diff against \`main\` will narrow once #312 lands.

## Summary

The shared synthetic-fixture matrix in \`tests/common/synthetic.rs\` used disjoint namespaces by construction — genomic contigs were \`NC_TEST.1\`, transcripts \`NM_TEST.1\` / \`NR_TEST.1\`, CDS fixtures pinned onto a separate \`chr_synth\` contig — and was therefore blind to the entire class of bugs that arise when a transcript id collides with a chromosome name on the same provider. That blindness is why #311 escaped pre-release.

\`AdversarialNamespaceBuilder::new(transcript_id, chromosome).build()\` materializes a \`MockProvider\` whose registered transcript shares its \`id\` (or, via its \`chromosome\` field, the contig name) with a colliding contig name. Sequence layout matches the issue's reproducer exactly:

- 909 bp poly-A leader + 93 bp coding segment + 100 bp poly-T trailer.
- \`genomic_start = 91\`, so genomic 1000..=1002 maps to transcript 910..=912 = \`GAC\`.
- A wrong-frame read at transcript indices 999..=1001 returns \`AAG\` — the buggy reference span that drove the collapse to \`1001A>C\` pre-#312.

## Tests

\`tests/issue_316_adversarial_namespace_matrix.rs\` exercises the builder across five collision shapes and pins the post-#312 expected output:

- prefix collision (\`chr1-gene.1\` ↔ \`chr1\`, the issue's reproducer)
- exact-name collision (\`chr1\` ↔ \`chr1\`)
- case-sensitivity boundary control (\`Chr1-gene.1\` ↔ \`chr1\`)
- no-collision RefSeq control (\`NM_000001.1\` ↔ \`chr1\`)
- already-canonical delins round-trip under the prefix collision

## Coordination notes

- **#312 (closes #311)**: this PR's base. The genomic-coordinate cases here would collapse the merged delins to a single residual SNV without #312's \`MockProvider::get_sequence\` / \`get_transcript\` fixes.
- **#317 (closes #314, \`MmapFastaProvider\`)** and **#318 (closes #315, \`FastaProvider\`)**: orthogonal. Each adds a targeted regression test using an inline fixture; a future cleanup pass can adopt this builder once all three siblings land.

## Test plan

- [x] \`cargo nextest run --features dev\` — 5060 / 5060 pass, 51 skipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - *(parser)* accept `?con<src>` and `?copy<N>` at unknown position, consistent with `?del`/`?dup`/`?ins`/`?inv` (closes #286)
 - *(fasta)* require version-boundary equality in `MmapFastaProvider::get_transcript`'s unversioned-prefix fallback (closes #314)
+- *(fasta)* route `FastaProvider::get_sequence` for known contigs through the FASTA path so a transcript registered with a chromosome-colliding id no longer wins over the genomic index (closes #315)
 
 ## [0.6.0](https://github.com/fulcrumgenomics/ferro-hgvs/compare/v0.5.0...v0.6.0) - 2026-05-16
 
--- a/src/reference/fasta.rs
+++ b/src/reference/fasta.rs
@@ -219,6 +219,22 @@ impl FastaProvider {
     pub fn add_transcript(&mut self, transcript: Transcript) {
         self.transcripts.insert(transcript.id.clone(), transcript);
     }
+
+    /// True when `id` is declared as a chromosome/contig name — either
+    /// resolves against the FASTA `.fai` index (directly or via an alias
+    /// chain) or appears as the `chromosome` field on a registered
+    /// transcript. Used by `get_sequence` to disambiguate genomic
+    /// lookups from transcript-id lookups when the two namespaces
+    /// collide (e.g. a transcript whose id is `chrN`; see #315).
+    fn is_known_contig(&self, id: &str) -> bool {
+        let resolved = self.resolve_name(id);
+        if self.index.contains_key(&resolved) {
+            return true;
+        }
+        self.transcripts
+            .values()
+            .any(|tx| tx.chromosome.as_deref() == Some(id))
+    }
 }
 
 impl ReferenceProvider for FastaProvider {
@@ -230,6 +246,17 @@ impl ReferenceProvider for FastaProvider {
     }
 
     fn get_sequence(&self, id: &str, start: u64, end: u64) -> Result<String, FerroError> {
+        // If the requested id is registered as a contig — either present
+        // in the FASTA index (directly or via an alias) or referenced as
+        // the `chromosome` field on a registered transcript — resolve it
+        // strictly against the FASTA path. Without this dispatch, a
+        // transcript whose id happens to match a contig name would
+        // short-circuit the lookup and silently interpret genomic
+        // coordinates as transcript-relative indices (#315 / #311).
+        if self.is_known_contig(id) {
+            return self.get_fasta_sequence(id, start, end);
+        }
+
         // First try as transcript ID. Only return a hit when the transcript
         // has cached sequence bases; coordinate-only transcripts fall through
         // to the FASTA index lookup below.
--- a/tests/common/synthetic.rs
+++ b/tests/common/synthetic.rs
@@ -217,6 +217,77 @@ impl SyntheticBuilder {
     }
 }
 
+/// Builder for `MockProvider` fixtures that deliberately put a transcript
+/// id and a chromosome name into the same key space — the failure mode
+/// behind #311 / #312. The fixture mirrors the issue's reproducer:
+///
+/// - one transcript with caller-supplied `id` whose `chromosome` field
+///   is the contig name (no separate `genomic_sequences` entry).
+/// - the transcript sequence is laid out so that the genomic-coordinate
+///   bases at `g.1000..=1002` (= `GAC`) are observably different from
+///   the bases a wrong-frame lookup would return (`AAG`); a wrong-frame
+///   trim therefore collapses an `ACG` delins to a single residual
+///   `1001A>C`. The numerical positions match the issue's reproducer
+///   exactly so downstream cross-PR tests can share assertions.
+///
+/// Sequence layout, all 1102 bp:
+///   - 909 bp poly-A leader
+///   - 93 bp coding segment (`SEGMENT` below)
+///   - 100 bp poly-T trailer
+///
+/// With `genomic_start = 91`, genomic position 1000..=1002 maps to
+/// transcript positions 910..=912 (= `SEGMENT[0..3] = "GAC"`). The
+/// wrong-frame read at transcript indices 999..=1001 returns
+/// `SEGMENT[90..93] = "AAG"`, which is the buggy reference span that
+/// drove #311's collapse to `1001A>C`.
+pub struct AdversarialNamespaceBuilder {
+    transcript_id: String,
+    chromosome: String,
+}
+
+const ADVERSARIAL_SEGMENT: &str =
+    "GACCGGCGGTCTGCAAGTCTCCACCTTCCGAAGCTATCCATAACCGGAACCTATGACCTAAAATCAGTTCTGGGTCAGCTCGGTATCACGAAG";
+
+impl AdversarialNamespaceBuilder {
+    pub fn new(transcript_id: &str, chromosome: &str) -> Self {
+        Self {
+            transcript_id: transcript_id.to_string(),
+            chromosome: chromosome.to_string(),
+        }
+    }
+
+    pub fn build(self) -> MockProvider {
+        let mut s = String::with_capacity(1102);
+        s.extend(std::iter::repeat_n('A', 909));
+        s.push_str(ADVERSARIAL_SEGMENT);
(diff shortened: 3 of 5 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #321, opened 2026-05-18
feat(annotation): populate Transcript.protein_id from GFF/GTF attributes

nh13:
## Summary

Stacked on #313 — depends on the `protein_id` field added there.

The GFF/GTF transcript builder previously hardcoded `protein_id: None` (`src/reference/annotation/builder.rs:318`), forcing downstream consumers onto the projector's transcript-id fallback (the safety net added in #310). Both source formats actually carry an explicit protein accession:

- **GTF**: `protein_id "NP_..."` is the standard attribute on each CDS record. Both GENCODE and RefSeq populate it.
- **GFF3**: RefSeq/GenBank put `protein_id=NP_...` on every CDS feature. Some producers additionally surface it on the parent mRNA record.

This PR pulls `protein_id` from those attributes so `ferro convert-gff` output gets a real `NP_*` / `ENSP_*` accession on each transcript instead of relying on `:p.<transcript_id>` cosmetic fallback.

## Resolution order

1. Transcript-level `protein_id` attribute (when present — some GFF3 producers put it here).
2. The first CDS child's `protein_id` attribute (the common location).
3. `None` (and the projector's transcript-id fallback from #310 takes over).

All CDS fragments of a single transcript share the same `protein_id`, so the first-match strategy is correct.

## Test plan
- [x] New tests in `src/reference/annotation/builder.rs::tests` cover: GFF3 CDS-level extraction, GFF3 transcript-level precedence, multi-fragment CDS first-match, missing attribute → `None`, GTF CDS-level extraction.
- [x] `cargo nextest run --features dev` — 5053 tests pass.

### DIFF
--- a/src/error_handling/registry.rs
+++ b/src/error_handling/registry.rs
@@ -1276,6 +1276,30 @@ fn build_registry() -> HashMap<&'static str, CodeInfo> {
     );
 
     map.insert(
+        "W-LOAD-120",
+        CodeInfo {
+            code: "W-LOAD-120",
+            name: "ConflictingProteinId",
+            summary: "CDS fragments of a single transcript reported disagreeing `protein_id` values; `protein_id` was left unset.",
+            explanation: "GFF3 and GTF convention is that every CDS fragment of a single \
+                transcript carries the same `protein_id` (they all encode parts of the same \
+                polypeptide). When the loader observes two or more distinct `protein_id` \
+                values on the CDS records of one transcript, it cannot pick one safely \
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #322, opened 2026-05-18
refactor(reference): impl Default for Transcript to enable spread-update fixtures

nh13:
## Summary

Stacked on #313.

Adds `#[derive(Default)]` to `Transcript` so call sites can use the spread-update pattern:

```rust
let tx = Transcript {
    id: "NM_DEMO.1".to_string(),
    gene_symbol: Some("DEMO".to_string()),
    sequence: Some("ATGTAA".to_string()),
    cds_start: Some(1),
    cds_end: Some(6),
    exons: vec![Exon::new(1, 1, 6)],
    ..Default::default()
};
```

Two follow-on benefits:

1. **No more 14-argument `Transcript::new` calls.** Most fixtures only meaningfully set 5–8 fields; the rest are `None` / empty / default and currently expand into 8–10 lines of boilerplate per literal.
2. **New optional fields are non-breaking for spread-update call sites.** In #313 (which added `protein_id`) we had to touch 22+ files solely to insert `protein_id: None,` because every existing literal was fully explicit. With this in place, future field additions only need to update literals that actually care about the new field.

## Scope

API addition + tests only. Existing literals are left alone so the diff stays small and easy to review — they can be migrated opportunistically as files are touched in the natural course of work. The malformed-exon test (`test_transcript_sequence_length_fallback_handles_malformed_exon`) is updated as a worked example of the new shape so reviewers can see what the migration looks like.

## Test plan
- [x] `test_transcript_default_is_minimal_and_non_coding` — pins every field of `Transcript::default()`.
- [x] `test_transcript_default_supports_spread_update` — pins the ergonomics: build with 6 explicit fields, let `Default` supply the rest.
- [x] `test_transcript_sequence_length_fallback_handles_malformed_exon` rewritten to use `..Default::default()` as a migration example.
- [x] `cargo nextest run --features dev` — 5050 tests pass.

### DIFF
--- a/src/reference/transcript.rs
+++ b/src/reference/transcript.rs
@@ -259,8 +259,15 @@ impl std::fmt::Display for ManeStatus {
     }
 }
 
-/// A transcript with its exon structure and sequence
-#[derive(Debug, Serialize, Deserialize)]
+/// A transcript with its exon structure and sequence.
+///
+/// `Default` is provided so test fixtures and reference producers can
+/// build a `Transcript` by setting only the fields they care about and
+/// leaving the rest at their zero/none values via the struct-update
+/// syntax `..Default::default()`. The Rust pattern avoids the 14-arg
+/// constructor and keeps fixtures stable when new optional fields are
+/// added.
+#[derive(Debug, Default, Serialize, Deserialize)]
 pub struct Transcript {
     /// Transcript accession (e.g., "NM_000088.3")
     pub id: String,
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #325, opened 2026-05-19
tracking: biocommons-normalize fixture burn-down


nh13:
Tracking issue for divergences between ferro and biocommons in the `tests/fixtures/biocommons-normalize/` normalize corpus (imported by #324). Sibling of #323 (mutalyzer corpus + `tests/fixtures/CORPUS_LAYOUT.md`); part of #87 (HGVS v21.0 spec-compliance gaps).

> **Status update (2026-05-29 — supersedes the earlier same-day refresh).** Correction: an earlier refresh framed the 15 `NM_001166478.1` "Pattern J" rows as a permanent accepted-divergence floor and predicted a 17-row baseline. That framing is no longer accurate. **PR #417 (merged 2026-05-22) wired `manifest.supplemental_fasta`**, so ferro now resolves `NM_001166478.1` (and `NM_212556.2`) from GenBank-deposited bases instead of cdot reconstruction — **demoting 13 of the 15 Pattern J rows**. Pattern J was a *reference-data* artifact, now largely fixed by data, not a permanent divergence. What remains is a small set of real bugs, two genuine accepted-divergences, and a test-infrastructure problem — see below.

## The real remaining work

**Open code bugs**
- **#473** — `NM_001166478.1:c.36_37insTC` (3'+cross) `ins→dup` canonicalisation off-by-one (`c.35_36dup` vs `c.36_37dup`). One of the 2 `NM_001166478.1` rows #417 did not demote; a genuine canon bug, not reference data.
- **#472** — `NM_212556.2:c.1delinsCA` **panics** (`attempt to subtract with overflow`) plus 3 output regressions, surfaced once `.2` resolves from the supplemental FASTA. The panic is a hard crash — **highest-priority remaining item**.

**Open observability**
- **#471** — stronger diagnostic when cdot-fallback reconstruction is likely to diverge from the deposited transcript.

**Genuine accepted-divergences (ferro is spec-correct; these stay)**
- **#253** (Pattern E) — `NM_001166478.1:c.59_61del`, boundary-spanning `del` with `cross=false`. HGVS v21 does not forbid the form.
- **#404** — `NM_000051.3:c.-2_-1insCA`, spanning-`dup` canonicalisation per edit-type priority.

## The infrastructure problem (and the principled fix)

[...]

PR #341, opened 2026-05-19
fix(reference): synthesize transcript bases via cdot exon-alignment when FASTA version is missing (closes #331)

### LINKED ISSUE #331: MultiFastaProvider: use cdot exon-alignment as fallback when transcript FASTA version is missing
## Symptom

When the ferro-prepared reference bundle lacks a transcript accession at the exact version a caller requests, `MultiFastaProvider::resolve_name` only falls back to a different version if the input had **no** version at all. If the input is versioned (`NM_001166478.1:c.31del`) and only `NM_001166478.2` is on disk — or no version is on disk — the fetch silently fails and `Normalizer::normalize` returns the input verbatim.

The biocommons normalize corpus hits this for `NM_001166478.1` (gene DEFB133), which is absent from the bundle in *every* version. cdot still has the full exon alignment (transcript-position → genome-position on `NC_000006.12`), so the bases are technically reachable via the genome FASTA + cdot.

## Affected rows

13 lines in `tests/fixtures/biocommons-normalize/baseline-failures/normalized.txt` from #324 (search for `NM_001166478.1` — the dedup of upstream cases collapses some, full list:

```
NM_001166478.1:c.2_7delinsTTTAGA      (3prime+cross,  5prime+cross)
NM_001166478.1:c.30_31insT            (3prime+cross,  5prime+cross)
NM_001166478.1:c.31del                (3prime+cross)
NM_001166478.1:c.34del                (5prime+cross)
NM_001166478.1:c.35_36dup             (3prime+cross)
NM_001166478.1:c.35_36insT            (3prime+cross,  5prime+cross)
NM_001166478.1:c.36_37insTC           (3prime+cross,  5prime+cross)
NM_001166478.1:c.36_37insTCTCTC       (3prime+cross,  5prime+cross)
NM_001166478.1:c.59delG               (3prime+cross,  3prime+no-cross)
NM_001166478.1:c.59_61del             (3prime+no-cross — also #325 Pattern E)
NM_001166478.1:c.61delG               (5prime+cross,  5prime+no-cross)
```

## Spec basis

§HGVS v21.0 `general.md` — 3'-rule shifting requires reading reference bases. When the bases are reachable via a known alignment, the normalizer should not silently no-op.

## Reproducer

With the standard ferro-prepared manifest:

```
$ ferro normalize 'NM_001166478.1:c.31del' --reference <manifest>
NM_001166478.1:c.31del      # expected: NM_001166478.1:c.35del per biocommons (3' shift through an A-run)
```

cdot does contain the entry:
```
$ grep -o '\"NM_001166478.1\":\s*{[^}]\+exons[^]]\+]' \
    /path/to/cdot-0.2.32.refseq.GRCh38.json | head
```
returns the full alignment.

## Suggested directions

1. **`get_transcript` fallback**: if the FASTA index has no exact-versioned key and cdot has an entry for the requested versioned id, synthesize the transcript bases by walking the cdot exon alignment against the genome FASTA. Strand-aware.

2. **Document the fallback policy**: per-call vs. cached. Likely cache: the synthesized `Transcript` should be memoized so repeated normalize calls don't pay the genome-FASTA round trip.

3. **Surface a soft warning** (e.g. `W3018 TranscriptVersionFallback`) when the fallback fires so callers know they're getting reconstructed bases vs. the canonical transcript record.

## Linked

- Umbrella: #325
- Companion: #324 (biocommons corpus opens this gap)
- Related: `src/reference/multi_fasta.rs::resolve_name` already has the partial-version fallback; this issue extends it to the cdot-driven case

nh13:
Closes #331.

## Summary

When the ferro-prepared reference bundle lacks a transcript FASTA at the requested version, `MultiFastaProvider::get_transcript` previously returned `ReferenceNotFound` and `Normalizer::normalize` silently passed the input through. This blocks the 3' rule for any transcript whose exact version isn't in the FASTA bundle — concretely surfaced by the biocommons-normalize corpus (#324) where `NM_001166478.1` (gene DEFB133) has its exon alignment in cdot but no FASTA version on disk.

This PR teaches `get_transcript` to synthesize transcript bases from the **cdot exon alignment + genome FASTA** when the transcript FASTA path fails. cdot supplies the exon-to-genome coordinates and the strand; the helper walks each exon in tx-order, fetches the genomic window from the genome FASTA via the existing `get_genomic_sequence` API, and concatenates — reverse-complementing per-exon for minus-strand transcripts.

## Behavior

- **No regression**: the FASTA path is tried first and unchanged. The fallback only fires when the FASTA index has no entry for `id`.
- **Warns on every fallback firing** (matching the existing supplemental-CDS warnings). Emits a more pointed warning when cdot's own version-strip fallback (`NM_FOO.1` → `NM_FOO.2`) resolved to a sibling version so callers can detect cross-version mismatches.
- **Strand-aware**: for minus-strand transcripts, reverse-complements each exon's genome bases before concatenating in tx-order. (Concatenating first and reverse-complementing the whole string would emit the second exon's bases before the first exon's; the bug was caught by the multi-exon minus-strand test.)
- **Typed errors**: returns `FerroError::GenomicReferenceNotAvailable` when the cdot-named contig is absent from the genome FASTA, and `FerroError::InvalidCoordinates` for empty exon arrays or degenerate exon spans. No panic path.

## Empirical verification (against the ferro-prepared manifest)

```
$ ferro normalize 'NM_001166478.1:c.34del'      --reference <manifest>
NM_001166478.1:c.34del -> NM_001166478.1:c.35del      # was unchanged pre-fix

$ ferro normalize 'NM_001166478.1:c.36_37insTC' --reference <manifest>
NM_001166478.1:c.36_37insTC -> NM_001166478.1:c.37_38dup   # was unchanged pre-fix
```

Some `NM_001166478.1` rows in #324's `baseline-failures/normalized.txt` will demote once #324 rebases onto this fix; the remainder are independent ferro issues (see #325 umbrella patterns A, C, I).

## Review

Two parallel pre-PR reviews (Opus + Sonnet) caught:

- A correctness bug in multi-exon minus-strand strand handling (per-exon revcomp vs. whole-string revcomp).
- A silent cross-version mismatch when cdot's own version fallback fires.
- Missing soft-warning on fallback firing (called out in the issue body).
- Missing guards + tests for empty exon array and degenerate exon span.
- A non-idiomatic `Box::leak(Box::new(dir))` test pattern; replaced with a `(provider, tempdir)` tuple return.
- Doc-comment contradictions on the cdot coordinate system; expanded the helper's doc with the strand-handling rationale and added a TODO for the pre-existing hardcoded `GenomeBuild::GRCh38`.

All feedback applied in the second commit; ferro is now green across the full test suite under simulated CI and dev-box-with-manifest.

## Test plan

- [x] 7 new unit tests in `src/reference/multi_fasta.rs::tests` cover: plus-strand single-exon, minus-strand single-exon, multi-exon plus-strand, multi-exon minus-strand, missing genome contig, empty exon array, degenerate exon span. The multi-exon-minus-strand test deliberately uses non-palindromic bases so the pre-fix and post-fix outputs differ.
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo clippy --features dev --all-targets -- -D warnings` clean
- [x] `cargo nextest run --features dev` — 5113 passed (3 more than pre-PR; no regressions)
- [x] Empirical: real-manifest probe on `NM_001166478.1` shifts correctly (see verification above)

### DIFF
--- a/src/data/cdot.rs
+++ b/src/data/cdot.rs
@@ -1166,6 +1166,15 @@ impl CdotMapper {
     /// Get a transcript by accession.
     ///
     /// For LRG transcripts (e.g., "LRG_1t1"), this will look up the equivalent
+    /// Whether the mapper has a transcript at exactly the requested accession
+    /// (no version-strip or LRG fallback). Used by callers that need to
+    /// distinguish "we have your exact version" from "we resolved to a
+    /// sibling version" — e.g. the cdot-driven base synthesis path in
+    /// `MultiFastaProvider::get_transcript` (closes #331).
+    pub fn has_transcript_exact(&self, accession: &str) -> bool {
+        self.transcripts.contains_key(accession)
+    }
+
     /// RefSeq transcript if an LRG mapping has been loaded.
     ///
     /// Supports version fallback: if "NM_000088.2" is not found but "NM_000088.3"
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -623,12 +623,29 @@ impl MultiFastaProvider {
     /// transcript FASTA does not carry the requested versioned accession but
     /// cdot does (closes #331).
     ///
-    /// Strand-aware: cdot stores genome-orientation bases; minus-strand
-    /// transcripts get reverse-complemented to transcript orientation here.
+    /// Strand-aware. cdot exon arrays store genome bases in the genome's
+    /// 5'→3' orientation regardless of transcript strand, and exons are sorted
+    /// by `tx_start`. For a plus-strand transcript that ordering is also
+    /// genome-ascending, so concatenating the per-exon genome windows in
+    /// tx-order yields the transcript sequence directly. For a minus-strand
+    /// transcript the same ordering is *genome-descending* — so we
+    /// reverse-complement each exon's genome bases as we go and concatenate in
+    /// tx-order. Concatenating first and reverse-complementing the whole string
+    /// would emit the second exon's bases before the first exon's (closes #331
+    /// review feedback).
     ///
-    /// Returns `FerroError::GenomicReferenceNotAvailable` if the cdot-named
+    /// Returns `FerroError::GenomicReferenceNotAvailable` when the cdot-named
     /// contig is missing from the genome FASTA, and `FerroError::InvalidCoordinates`
-    /// if any exon's genomic span is degenerate.
+    /// when any exon's genomic span is degenerate or the cdot record carries
+    /// no exons. The cdot-named contig is consulted directly (not version-
+    /// stripped); callers that rely on a different contig version should keep
+    /// the genome FASTA in sync with cdot.
+    ///
+    /// `tx.strand` is honored; metadata fields are projected onto the returned
+    /// `Transcript` using the same conventions as the FASTA-backed path so
+    /// downstream consumers see one shape regardless of which path served them.
+    /// `genome_build` is currently hardcoded to `GRCh38` to match the FASTA
+    /// path; pre-existing inconsistency, see TODO below.
     fn synthesize_transcript_from_cdot(
         &self,
         id: &str,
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #345, opened 2026-05-19
fix(normalize): bail gracefully when ref window mismatches HGVS span (closes #339)

### LINKED ISSUE #339: normalize: panic on `ref_bytes length must match HGVS interval span` for delins spanning alignment gaps
## Symptom

ferro panics inside `Normalizer::normalize` on a large `delins` where the reference window length does not equal the HGVS interval span. The mismatch is real (genomic alignment includes gaps that collapse the byte length); ferro fires an `assert!` instead of returning an error.

## Probe / reproducer

```
$ ferro normalize 'NG_032871.1:g.32476_53457delinsAATTAAGGTATA' --reference <manifest>

thread 'axis_normalized' panicked at src/normalize/mod.rs:2779:9:
assertion `left == right` failed: ref_bytes length must match HGVS interval span
  left: 15539
 right: 20982
```

The variant spans 20,982 bp on the transcript coordinate but ferro fetches a 15,539-byte ref window — likely because the cdot exon alignment between transcript and genome collapses some gaps. The assertion is testing an invariant that doesn't hold when alignment gaps are present.

## Affected rows

1 line in `tests/fixtures/biocommons-normalize/baseline-failures/normalized.txt` from #324 (the case appears twice in upstream — biocommons python copy-paste — and panics both times):

| Input | bio | ferro |
|---|---|---|
| `NG_032871.1:g.32476_53457delinsAATTAAGGTATA` | rejects (`HGVSInvalidVariantError`) | panics |

## Severity

This is a **panic in the normalize hot path** — it would abort any caller running this input in production. The biocommons rejection rationale (extreme ref/alt length ratio heuristic) is implementation-specific and not spec-mandated; ferro doesn't need to reject *per se*, but it must not panic.

## Spec basis

§HGVS v21.0 `DNA/delins.md` does not impose size limits on delins. The spec includes an 8 Mb delins example (`NC_000002.12:g.pter_8247756delins[NC_000011.10:g.pter_15825266]`). So the input is well-formed and ferro should produce *some* output — either a successfully normalized form (acknowledging the alignment-gap reality) or a typed error, never a panic.

## Source

```
src/normalize/mod.rs:2779
```

## Suggested directions

1. **Immediate**: replace the `assert!` with an early `Err(FerroError::...)` returning a structured error like `AlignmentGapInRefWindow { expected_span: 20982, actual_bytes: 15539 }`. Tests will fail loudly but won't panic.

2. **Correct fix**: the normalize pass needs to either (a) traverse the alignment block-by-block so the byte length matches the spanned positions on the transcript coordinate, or (b) explicitly bail out as "delins spans alignment gap, cannot apply 3'-rule shift" with a clear soft warning. (a) is the rigorous path; (b) is the safe degraded mode.

## Linked

- Umbrella: #325
- Companion: #324
- Probably related: cdot alignment loader in `src/data/cdot/json.rs` and `src/reference/multi_fasta.rs`

nh13:
Closes #339.

## Summary

`Normalizer::normalize` panicked in debug builds when the provider's `get_sequence` returned a byte slice whose length didn't equal `(hgvs_end - hgvs_start + 1)`. This happens in practice when a cdot exon alignment collapses gaps that the HGVS span counts as positions — surfaced by the biocommons-normalize corpus (#324):

```
$ ferro normalize 'NG_032871.1:g.32476_53457delinsAATTAAGGTATA' --reference <manifest>
panicked at src/normalize/mod.rs:2779:9:
assertion `left == right` failed: ref_bytes length must match HGVS interval span
  left: 15539
 right: 20982
```

## Fix

Replace the `debug_assert_eq!` with a graceful early-return:

```rust
if ref_bytes.len() != (hgvs_end - hgvs_start + 1) as usize {
    return vec![variant];  // alignment gap or provider mismatch; skip canonical split
}
```

The canonical-split helper already returns `vec![variant]` whenever it cannot process the input (non-Delins, complex location, no provider data, nothing to split out). The mismatch case is added to that contract.

Release behaviour: pre-fix release builds did NOT corrupt memory — `decompose_delins` independently bounds-checks (`src/normalize/rules.rs:736-738`) and returned `None`, falling through to the same `vec![variant]` outcome. So the user-visible effect of this PR is solely "no more debug-build panic"; release behaviour is unchanged for this input.

## Review consolidation

Opus + Sonnet reviewed commit 1 in parallel. Both confirmed the bail-out is the right contract. Action items from review (all applied in commit 2):

- Tightened the bail-out tests to assert input round-trips unchanged (not just "no panic"). Opus.
- Renamed the misleading second test to clarify it's a smaller-span case of the same mismatch shape. Sonnet.
- Added a third test covering the `ref_bytes.len() > expected_span` direction. Sonnet + Opus.
- Updated `apply_canonical_split` docstring to enumerate the new bail-out case. Sonnet.
- Renamed `TruncatingProvider` to `FixedLengthProvider` with a configurable payload.

## Out-of-scope follow-ups (acknowledged in both reviews)

1. `NormalizationWarning::CanonicalSplitSkipped` — surface the bail-out to callers so they can detect that canonicalization was skipped (vs. nothing-to-split). Threading through requires `apply_canonical_split` to return warnings, touching every caller.
2. Alignment-block-by-block traversal (issue body's "suggested direction 2"). Architectural change requiring cdot exon-alignment metadata to flow into `fetch_ref_for_canonical_split`.

## CI / signing

Both commits **unsigned** — 1Password biometric was unavailable per CLAUDE.md (fall back to `--no-gpg-sign` after 3 failures, re-sign when back at the machine). Will re-sign before merge.

## Test plan

- [x] `cargo fmt --all -- --check` clean
- [x] `cargo clippy --features dev --all-targets -- -D warnings` clean
- [x] `cargo nextest run --features dev` — 5115 passed, 51 skipped, no regressions
- [x] `cargo nextest run --features dev -E 'binary(issue_339_alignment_gap_panic)'` — 3/3 pass (shorter, shorter-small-span, longer)
- [x] No previously-passing test relied on the assert (`NG_032871.1` appears only in the new regression test and the biocommons-normalize baseline fixture)

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -2854,12 +2854,22 @@ impl<P: ReferenceProvider> Normalizer<P> {
         else {
             return vec![variant];
         };
+        // When the provider returns fewer (or more) bytes than the HGVS
+        // interval span, the canonical-split decomposition would walk past
+        // the end of `ref_bytes` (or under-walk and miss interior
+        // identities). This happens in practice when a cdot exon
+        // alignment collapses gaps that the HGVS span counts as
+        // positions (e.g. biocommons NG_032871.1:g.32476_53457delins…
+        // returns 15,539 bytes for a 20,982-bp span). Pre-fix this
+        // fired a `debug_assert_eq!` that panicked debug builds and
+        // would have produced out-of-bounds reads in release. Now bail
+        // out gracefully: the input variant cannot be split without an
+        // authoritative ref window, so leave it as-is. Closes #339.
         let n = ref_bytes.len();
-        debug_assert_eq!(
-            n,
-            (hgvs_end - hgvs_start + 1) as usize,
-            "ref_bytes length must match HGVS interval span"
-        );
+        let expected_span = (hgvs_end - hgvs_start + 1) as usize;
+        if n != expected_span {
+            return vec![variant];
+        }
         let ref_norm = normalize_t_u(&ref_bytes);
         let alt_norm = normalize_t_u(&alt_bytes);
         let Some(subedits) = rules::decompose_delins(&ref_norm, 0, n, &alt_norm) else {
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -2825,7 +2825,9 @@ impl<P: ReferenceProvider> Normalizer<P> {
     /// internally, calls `decompose_delins`, and rebuilds N variants when
     /// the decomposition fires. Returns `vec![variant]` if the variant
     /// doesn't decompose (non-Delins, complex location, no provider data,
-    /// nothing to split out).
+    /// nothing to split out, **or the provider's ref window doesn't match
+    /// the HGVS coordinate span** — e.g. cdot alignment gaps shorten the
+    /// returned byte slice, see #339).
     ///
     /// Implements two spec-priority rules from `general.md:56`
     /// (substitution > deletion > inversion > duplication > insertion):
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -118,6 +118,23 @@ pub enum NormalizationWarning {
         /// Edit kinds, e.g. ["sub", "sub"] or ["del", "inv"]
         edit_kinds: Vec<String>,
     },
+
+    /// `apply_canonical_split` was unable to canonicalize because the
+    /// reference window returned by the provider did not span the same
+    /// number of bytes as the HGVS interval (e.g. a cdot exon alignment
+    /// collapsed gap-positions in the genomic-to-tx projection). The
+    /// variant is returned unchanged. Closes-after: #354.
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #368, opened 2026-05-19
ci(external-validation): use --features dev to keep slow-tests opt-in (closes #40)

### LINKED ISSUE #40: external-validation: run-standard-tests uses --all-features, pulling in slow-tests without the required fixture
## What

`.github/workflows/external-validation.yml` → job `run-standard-tests` (line 194) invokes:

```yaml
- name: Run standard tests
  run: cargo test --all-features
```

`--all-features` enables `slow-tests`, which compiles `tests/clinvar_validation.rs`. That test opens `tests/fixtures/bulk/hgvs4variation.txt.gz` and panics when it isn't present — and nothing in the workflow (or the repo) provides it. The file is a ~1 GB ClinVar dump normally fetched manually via `curl` per the test's `expect()` message.

## How this surfaced

Pre-existing, hidden until very recently:

1. `external-validation.yml` referenced `dtolnay/rust-action@stable` (typo — that repo doesn't exist), so the `Set up Rust` step always failed and the job never reached `cargo test`.
2. PR #37 corrected the typo to `dtolnay/rust-toolchain`. The job then failed at checkout because `lfs: true` was missing for the LFS-backed `clinvar_hgvs_500k.json.gz` fixture; that was also fixed in #37.
3. With those out of the way, a fork `workflow_dispatch` on PR #37 now gets as far as `cargo test --all-features` and fails in `test_clinvar_validation` — the third pre-existing problem in the same job, now visible for the first time.

Fork dispatch that exposed it: https://github.com/msto/ferro-hgvs/actions/runs/24903464912

## Proposed fix

Swap `--all-features` for `--features dev` to match how `ci.yml` runs the same test surface. The `dev` feature flag exists specifically to bundle \"all testable features\" and deliberately excludes `slow-tests`:

```toml
# Cargo.toml
# Long-running tests (e.g., full ClinVar validation). Run with: cargo test --release --features slow-tests
slow-tests = []
# dev feature includes all testable features (python requires maturin for building)
dev = [\"validation\", \"mmap\", \"benchmark\", \"parallel\", \"protein-fetch\", \"web-service\"]
```

Minimal diff:

```diff
       - name: Run standard tests
-        run: cargo test --all-features
+        run: cargo test --features dev
```

This leaves `slow-tests` opt-in, which is the intended contract. A separate scheduled job (or a dedicated `workflow_dispatch` input) could exercise `slow-tests` after downloading `hgvs4variation.txt.gz`, but that's out of scope for restoring the workflow's baseline.

## Why not just download the fixture

`hgvs4variation.txt.gz` is ~1 GB and the test iterates all 42M rows. Adding it to every scheduled/dispatched run of this workflow is a disproportionate cost when the comparable signal is already provided by `clinvar_hgvs_500k.json.gz` (500k variants, already covered by `run-standard-tests` via `cargo test --features dev` once this lands).

nh13:
Closes #40.

## Summary

`.github/workflows/external-validation.yml` ran `cargo test --all-features` and `cargo clippy --all-features` in the `run-standard-tests` job. `--all-features` enables the `slow-tests` feature, which compiles `tests/clinvar_validation.rs` — the file is gated behind `#![cfg(feature = "slow-tests")]` and requires the ~1 GB `tests/fixtures/bulk/hgvs4variation.txt.gz` ClinVar dump that is not committed and not fetched by the workflow. The test panics at runtime when the fixture is missing.

The `dev` feature flag was created for exactly this case: per `Cargo.toml` it bundles every testable feature (`validation`, `mmap`, `benchmark`, `parallel`, `protein-fetch`, `web-service`) while deliberately excluding `slow-tests`. `ci.yml` already uses `--features dev` for both `cargo nextest run` and `cargo clippy`. This change brings external-validation in line.

## Diff

```diff
       - name: Run standard tests
-        run: cargo test --all-features
+        run: cargo test --features dev

       - name: Run clippy
-        run: cargo clippy --all-features -- -D warnings
+        run: cargo clippy --features dev -- -D warnings
```

## Verification

- `tests/clinvar_validation.rs:8` is gated `#![cfg(feature = "slow-tests")]`.
- `Cargo.toml:76` defines `dev = ["validation", "mmap", "benchmark", "parallel", "protein-fetch", "web-service"]` — no `slow-tests`.
- `cargo test --features dev --no-run` builds the `clinvar_validation` test binary as an empty shell (`0 tests, 0 benchmarks` reported by the binary's `--list`), so no fixture access is attempted.
- Default features (`default = []`) carry no `slow-tests` either, so the `fetch-external-databases` job's `cargo test --test ...` calls (lines 128/133/138) are unaffected.

The fully-gated `slow-tests` job remains available via explicit `cargo test --release --features slow-tests` when someone provides the fixture.

## Test plan

- [x] YAML syntax validated (`python3 -c "import yaml; yaml.safe_load(...)"`)
- [x] Local `cargo test --features dev --no-run` succeeds; `clinvar_validation` binary reports 0 tests
- [ ] Fork `workflow_dispatch` on this branch reaches `Run standard tests` step and passes (requires merge or a manual fork dispatch)

### DIFF
--- a/.github/workflows/external-validation.yml
+++ b/.github/workflows/external-validation.yml
@@ -200,11 +200,14 @@ jobs:
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
+      # `dev` bundles every testable feature; `slow-tests` is
+      # intentionally opt-in because it requires a ~1 GB ClinVar
+      # fixture that is not committed to the repo.
       - name: Run standard tests
-        run: cargo test --all-features
+        run: cargo test --features dev
 
       - name: Run clippy
-        run: cargo clippy --all-features -- -D warnings
+        run: cargo clippy --features dev --all-targets -- -D warnings
 
   fetch-external-databases:
     name: Fetch External Database Data


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

PR #369, opened 2026-05-19
fix(diagnostics): preserve source span on E3006 SelfCancellingAllele (closes #171)

### LINKED ISSUE #171: diagnostics: AlleleVariant::try_new_validated drops source span (pos: 0) on E3006
Follow-up from CodeRabbit on #152.

`AlleleVariant::try_new_validated` (`src/hgvs/variant.rs` ~lines 485–510) constructs `FerroError::Parse` with a hard-coded `pos: 0`, so when E3006 fires the diagnostic renderer cannot underline the offending allele in the source span. The indices `(dl, du)` are embedded in the message text, but the structured `pos` field is unusable.

This was noted as L1-audit follow-up — the hint string carries enough info for end-users, but downstream tooling (LSP, structured-error consumers, the upcoming web service) loses span fidelity.

### Suggested fix

Plumb the parser's allele-start byte offset into `try_new_validated` (e.g. accept `Option<usize>` or a span type) and forward it to the `FerroError::Parse` `pos` field. Update the call site in `src/hgvs/parser/variant.rs` to pass the captured `Span::start` / `input.location_offset()` of the allele open.

### Acceptance

- `try_new_validated` accepts a caller-supplied position or span.
- E3006 diagnostics report a non-zero `pos` for the spec example `NM_004006.2:c.[762_768del;767_774dup]`.
- Existing E3006 tests still pass and a new test asserts the structured `pos` is the start of the allele bracket.

nh13:
Closes #171.

## Summary

`AlleleVariant::try_new_validated` and the parser-side `validate_no_self_cancelling` both built their E3006 `FerroError::Parse` with a hard-coded `pos: 0`, so downstream tooling (LSP, structured-error consumers, the web service) had no way to underline the offending allele in source.

For the spec example `NM_004006.2:c.[762_768del;767_774dup]`, E3006 now reports `pos: 14` (the offending `[`) and the `Diagnostic` carries a `SourceSpan` covering the full `[ ... ]` region plus the input string for caret rendering.

## Design

- **Shared error builder.** A new `build_self_cancelling_error` in `src/hgvs/variant.rs` is the single source of truth for the E3006 payload. Both `AlleleVariant::try_new_validated` and the parser-side validator forward into it, so library-built alleles and parser-built alleles emit byte-identical diagnostics.
- **Library API: `try_new_validated` accepts `Option<SourceSpan>`.** Callers that build alleles outside the parser (e.g. the upcoming web service) can supply their own span; when `None`, the diagnostic still carries E3006 with the indices in the message but `pos` remains 0 — explicit and documented.
- **Parser: locate the offending `[` from the original input.** `locate_cis_allele_bracket` matches an axis marker (`:c.`, `:g.`, `:n.`, `:r.`, `:m.`, `:o.`, `:p.`) immediately preceding `[` and tracks `[`/`]` depth so that nested inserted-sequence brackets (`ins[ATC]`) and repeat-count brackets don't confuse the locator. Recursion on nested alleles passes the absolute offset of the enclosing bracket so deep diagnostics resolve to global byte positions.

## Behavior diff

| Input | Before | After |
|---|---|---|
| `NM_004006.2:c.[762_768del;767_774dup]` | `pos: 0`, no span | `pos: 14`, span `(14, 37)`, source attached |

## Spec basis

HGVS `recommendations/general.md` line 47: \"descriptions removing part of a reference sequence replacing it with part of the same sequence are not allowed.\" Diagnostic location is not a normative requirement of the spec itself, but good diagnostic hygiene is the basis for compliant tooling (web service, LSP, editor integrations).

## Fixture drift

The pinned `tests/fixtures/grammar/hgvs_spec_normalization.json` row for this input captured the old `position 0`. Updated to `position 14` to match the spec-correct behavior. The fixture-generator example (`generate_spec_fixture`) is independently broken on `main` (`overrides reference unknown inputs: c.123C>C`) so the row is updated by hand — this PR does not chase that pre-existing breakage.

## Tests

- `test_parse_self_cancelling_allele_reports_source_span` — parser entry produces the offending `[` byte offset for the spec example and a `Diagnostic` span covering the full bracket.
- `test_self_cancelling_via_try_new_validated_takes_caller_span` — library entry propagates a caller-supplied span; with `None` the diagnostic still carries E3006 at `pos: 0`.
- Existing `test_parse_self_cancelling_allele_rejected` and `test_parse_non_overlapping_allele_accepted` still pass.

## Test plan

- [x] `cargo nextest run --features dev` — 5463/5463 pass (excluding pre-existing mutalyzer/biocommons corpus-runner divergences)
- [x] `cargo clippy --features dev --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all --check` — clean

### DIFF
--- a/src/hgvs/parser/variant.rs
+++ b/src/hgvs/parser/variant.rs
@@ -5353,7 +5353,7 @@ pub fn parse_variant(input: &str) -> Result<HgvsVariant, FerroError> {
 
     // Spec-mandated post-parse semantic check: reject self-cancelling alleles
     // (HGVS recommendations/general.md line 47, tracked under issue #115).
-    validate_no_self_cancelling(&variant)?;
+    validate_no_self_cancelling(&variant, input)?;
 
     Ok(variant)
 }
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai Thanks for the second pass. Closing this as wontfix — the suggested defensive code targets a non-spec input shape.

The HGVS spec (`assets/hgvs-nomenclature/docs/syntax.yaml`) defines exactly three allele forms: cis `[a;b]`, trans `[a];[b]`, unknown-phase `a(;)b`. Mosaic (`/`) and chimeric (`//`) are documented only as single-edit position-level separators (e.g. `c.85=/T>C`, see `DNA/substitution.md` and `general.md`). The form `[a;b]/[c;d]` has no grammar production or example anywhere in the spec, and `consultation/open-issues.md` records that the committee explicitly rejected nesting proposals. So the multi-cis Mosaic tree the fix targets cannot be produced from spec-conformant input.

Separately, while investigating I found a real parser bug worth its own issue/PR: `parse_phase_allele`'s fallback chain swallows chunk-level E3006 errors on slash-form inputs and re-emits them as "Unknown variant type prefix". Filing as a follow-up.

PR #377, opened 2026-05-20
feat(normalize): protein 3'-shifting for deletions and duplications

nh13:
## Summary

HGVS general.md mandates the 3'-rule for protein variants too: *"for all descriptions the most 3' position possible is arbitrarily assigned to have been changed."* ferro previously left protein variants in their input position with a `// no 3'/5' shifting` comment in `normalize_protein`; this PR wires a residue-level shuffler so protein deletions and duplications now shift to their canonical (rightmost) anchor.

## Algorithm

After the existing edit-shape canonicalization (e.g. `p.Val600delVal` → `p.Val600del`), deletions and duplications run through a new `shuffle_protein_3prime` helper:

1. Fetch the reference protein via `ReferenceProvider::get_protein_sequence`. Because the provider returns out-of-range errors rather than truncated strings and has no separate length API, the helper discovers the protein length via exponential-then-binary search (O(log n) provider calls; capped at ~1 GiB which is well above titin's ~35 kAA).
2. Iterate the rotation predicate `protein[start] == protein[start + edit_len]` until it fails or walks past the end of the protein. Each successful step advances both endpoints by one residue, preserving the multiset of residues in the edited range.
3. Reconstruct the new `ProtPos` endpoints with the correct AA names looked up via `AminoAcid::from_one_letter`.

Edits that are not 3'-shifted per spec — substitutions, frameshifts, extensions, identity, no-protein, repeats — pass through unchanged. **Insertions are intentionally NOT shifted in this PR.** Insertion 3'-shift is the foundation for `p.ins → p.dup` canonicalization (#92), which depends on this PR landing first so "most 3' position" is well-defined for the canonical-form rewrite.

## Tests

New `tests/issue_91_protein_three_prime_shift.rs` (6 tests):

- `protein_single_deletion_shifts_3prime_in_homopolymer` — `p.Ala4del` on `MKVAAALELE` shifts to `p.Ala6del`.
- `protein_range_deletion_shifts_3prime_through_repeat` — `p.Leu7_Glu8del` shifts through the duplicated `LE` motif to `p.Leu9_Glu10del`.
- `protein_duplication_shifts_3prime_in_homopolymer` — `p.Ala4dup` → `p.Ala6dup`.
- `protein_range_duplication_shifts_3prime_through_repeat` — `p.Leu7_Glu8dup` → `p.Leu9_Glu10dup`.
- `protein_substitution_does_not_shift` (negative control) — `p.Ala4Gly` stays at p.4.
[...]

Issue #91, opened 2026-05-02
Protein 3' shifting


nh13:
ferro currently does not 3'-shift protein variants. From `src/normalize/mod.rs:603`:

> Protein normalization differs from nucleic acid normalization — there's no 3'/5' shifting.

The HGVS spec mandates 3'-shifting at the protein level too:

> for all descriptions the most 3' position possible is arbitrarily assigned to have been changed.

Implementing this requires a protein-level shuffler — AA tokens, multi-letter encodings (`Met`, `Ter`/`*`), and synonymous-codon edge cases — that does not directly reuse the nucleotide shuffler in `src/normalize/shuffle.rs`.

Filed as a follow-up to the protein scope decision in the A7 design (#81 item A7). Spec doc: `docs/superpowers/specs/2026-05-01-A7-ins-shift-coverage-matrix-design.md`.

This is the protein analog of the nucleotide shift-coverage items (A5 del-shift, A6 dup-shift, A7 ins-shift) tracked under #81. When this lands, the protein \`p.ins\` → \`p.dup\` canonical-form rule becomes implementable.

Tracked under #81.

PR #378, opened 2026-05-20
feat(normalize): canonicalize p.ins → p.dup for upstream-matching insertions

### LINKED ISSUE #92: p.ins → p.dup canonicalization
HGVS spec: a protein insertion whose inserted residues are identical to the residues immediately adjacent (after 3' shifting) must be expressed as a duplication, not an insertion.

This is the protein analog of the nucleic-acid `ins → dup` rule tracked under #81 item A1. It depends on protein 3' shifting (separate sibling issue), which must land first so the most-3' placement rule applies.

Filed as a follow-up to the protein scope decision in the A7 design (#81 item A7). Spec doc: `docs/superpowers/specs/2026-05-01-A7-ins-shift-coverage-matrix-design.md`.

Tracked under #81.

---

## Status (updated 2026-05-21)

PR #378 shipped sub-item (1) — both the upstream-overlap and the (operationally-equivalent-by-3'-rule) downstream-overlap `p.ins → p.dup` canonicalizations — and the tests in `tests/issue_92_protein_ins_to_dup.rs` cover canonical, predicted-edit, partial-match, and insufficient-window cases. The original audit comment incorrectly listed downstream-overlap as unaddressed.

### Remaining work

**1. `p.delins → p.dup` canonicalization.** A `p.<X>_<Y>delins<seq>` that affix-trims to a pure insertion matching an immediately-adjacent reference window (per `duplication.md`'s tandem requirement) should canonicalize to `dup`. Implemented as affix-trim then route through the existing `try_protein_ins_to_dup`. Branches: identity (silent), pure del, single-residue sub, ins-becomes-dup, genuine residual delins.

**2. Initiator-Met soft-warning.** Surface a soft-warning when ins→dup or delins→dup canonicalization lands on a `Duplication` whose interval includes p.1 (e.g. `p.Met1dup`). Policy decision: ALLOW the rewrite (Prioritization rule `general.md:56-57` is unconditional; spec uses Met1-inclusive ranges in `deletion.md:63-65`); emit `W3022 InitiatorMetCanonicalization` so downstream consumers can detect the start-codon interaction.

## Source PRs

- PR #377 (#91): protein 3' shifting (prerequisite)
- PR #378: upstream + downstream `p.ins → p.dup` (sub-item 1, ✅ done)
- _(this issue's follow-up PR)_: sub-items (2) and (3)

nh13:
## Summary

HGVS Prioritization rule: when both `ins` and `dup` can describe the same change, the canonical form is `dup`. Concretely, when the inserted residues of `p.<X>_<Y>ins<seq>` equal the residues immediately upstream of position Y, the variant must be expressed as `p.<a>_<b>dup` over that upstream range. ferro previously had no such canonicalization for protein insertions; this PR wires it.

Stacked on PR #377 (`feat/issue-91-protein-three-prime-shift`) — the dup canonicalization depends on protein 3'-shifting (#91) so the "most 3' position" is well-defined for the canonical-form rewrite.

## Algorithm

New `try_protein_ins_to_dup` helper runs during `normalize_protein` for `ProteinEdit::Insertion` edits:

1. Fetch the `seq.len()` reference residues immediately preceding `start_pos.number` via `ReferenceProvider::get_protein_sequence`.
2. If they equal the inserted sequence, rewrite both the edit (to `ProteinEdit::Duplication`) and the interval (to the upstream-duplicated range).
3. The existing 3'-shift pass (from #91) then walks the dup to its canonical anchor.

The Insertion arm fires only when the variant is in the standard `<X>_<X+1>ins…` adjacent-position shape and `start_pos.number >= seq.len()` (i.e. there is a valid upstream window). It returns `None` on any spec-mismatched shape, and the existing fallback preserves the input `ins` form.

`normalize_protein` was refactored to track both `normalized_edit` and `post_canon_interval` since the `ins → dup` rewrite changes the interval. A synthesized intermediate variant carries the post-canonical interval into the shuffler so that the dup's 3'-shift fires on the rewritten anchor rather than the input position.

## Tests

New `tests/issue_92_protein_ins_to_dup.rs` (4 tests):

- `protein_single_ins_becomes_dup_at_3prime_anchor` — `p.Ala4_Ala5insAla` → `p.Ala6dup` (canonicalize + 3'-shift through AAA homopolymer).
- `protein_range_ins_becomes_range_dup_at_3prime_anchor` — `p.Glu8_Leu9insLeuGlu` → `p.Leu9_Glu10dup` (canonicalize + 3'-shift through LELE motif).
- `protein_ins_without_duplicate_stays_as_ins` — `p.Ala4_Ala5insGly` (G doesn't match upstream A) stays as ins.
- `protein_ins_matching_following_residue_alone_stays_as_ins` — `p.Val3_Ala4insAla` (preceding residue is V, not A) stays as ins per the strict spec interpretation (upstream window only).

5563/5563 focused tests pass. Clippy + fmt clean.

## Out of scope

- Detecting the symmetric *downstream*-overlap case (where the inserted residues match what follows the insertion point) — would require a separate spec citation and is not covered by HGVS Prioritization as written.
- `p.delins` → `p.dup` rewrites — different spec rule, separate work.

Closes #92.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -972,14 +972,16 @@ impl<P: ReferenceProvider> Normalizer<P> {
             None => return Ok((HV::Protein(variant.clone()), vec![])),
         };
 
-        // Apply normalization based on edit type
-        let normalized_edit = match edit {
+        // Apply normalization based on edit type. May rewrite both the
+        // edit and the interval (e.g. `ins → dup` canonicalization
+        // anchors the dup at the upstream-duplicated residue range,
+        // which is a different interval than the input `<X>_<Y>ins…`).
+        let (normalized_edit, post_canon_interval) = match edit {
             ProteinEdit::Deletion { sequence, count } => {
                 // Check for redundant sequence that matches the position
-                if let Some(seq) = sequence {
+                let new_edit = if let Some(seq) = sequence {
                     if self.is_redundant_protein_deletion_sequence(&variant.loc_edit.location, seq)
                     {
-                        // Remove redundant sequence
                         ProteinEdit::Deletion {
                             sequence: None,
                             count: *count,
@@ -989,10 +991,19 @@ impl<P: ReferenceProvider> Normalizer<P> {
                     }
                 } else {
                     edit.clone()
-                }
+                };
+                (new_edit, variant.loc_edit.location.clone())
             }
+            // HGVS Prioritization: `<X>_<Y>ins<seq>` whose inserted
+            // residues duplicate the residues immediately upstream
+            // (anchored at p.X) is canonicalized to `<a>_<b>dup` over
+            // that upstream range. The 3'-shift pass that follows will
+            // then walk the dup to its canonical anchor. (#92)
+            ProteinEdit::Insertion { sequence } => self
+                .try_protein_ins_to_dup(variant, sequence)
+                .unwrap_or_else(|| (edit.clone(), variant.loc_edit.location.clone())),
             // Other edits pass through unchanged
-            _ => edit.clone(),
+            _ => (edit.clone(), variant.loc_edit.location.clone()),
         };
 
         // Compute the post-shuffle interval. The shuffle layer operates on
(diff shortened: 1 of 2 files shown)


nh13:
Addressed both reviewer rounds in amend a5fdf69:

**Must-fix-class (Opus #1, real spec correctness issue):**
- Extended `try_protein_ins_to_dup` to check **both** the upstream AND downstream windows. Per HGVS Prioritization, when both `ins` and `dup` can describe the same change, dup is canonical; the previous upstream-only interpretation under-canonicalized for inputs positioned on the 5' side of an ambiguous boundary. `p.Val3_Ala4insAla` (insert A between V and A) describes the same change as `p.Ala4dup` — the inserted A matches the *following* residue. The downstream-window check fires here, anchors the dup at the downstream window, and the existing 3'-shift walks it through the AAA tract to `p.Ala6dup`.
- The 4th test was flipped from "stays as ins" to "canonicalizes to dup" to reflect the spec-correct behaviour. Updated docstring with the spec citation.

**Nice-to-have (Opus #2, Sonnet #2/3/4):**
- Hard-guarded the helper to require `Mu::Certain` at both endpoints (was `inner()` which silently accepted `Mu::Uncertain`). Uncertain or range boundaries now pass through unchanged so the parenthesized-uncertain marker survives. Predicted-edit wrappers `p.(Ala4_Ala5insAla)` still canonicalize correctly because the `Mu::Uncertain` wraps the whole `loc_edit.edit`, not the position endpoints inside the interval.
- Added the safety comment on `canon_variant` construction (Opus #5): the shuffler reads only `loc_edit.location`, so the Mu wrapping on the synthesized edit doesn't matter today; flagged for future shuffler refactors.
- Extracted `fetch_protein_window` helper to deduplicate the upstream/downstream provider-fetch logic.

**New tests (test count 4 → 7):**
- `protein_ins_matching_following_residue_canonicalizes_to_dup` — `p.Val3_Ala4insAla` → `p.Ala6dup` (pins the downstream-window match).
- `protein_ins_partial_match_stays_as_ins` — `p.Ala4_Ala5insAlaGly`: leading A matches but full 2-AA window doesn't, stays as ins (pins that the rule is full-window equality).
- `protein_predicted_ins_canonicalizes_inside_parens` — `p.(Ala4_Ala5insAla)` → `p.(Ala6dup)` (pins that `Mu::Uncertain` survives end-to-end).
- `protein_ins_longer_than_upstream_window_stays_as_ins` — `p.Lys2_Val3insMetLysVal` (3 residues, upstream window has only 2 available, downstream doesn't match either): stays as ins (pins the bounds guard).

**Deferred with rationale:**
- Opus #3 (initiator-Met dup) — separate spec question. The Met1 case requires its own arbitration; if a real-world input surfaces, can be a follow-up.
- Provider-call cost for batch annotation — the upstream+downstream window fetches are 1-2 small range reads per insertion; well-bounded for any realistic workload.

5566/5566 focused tests pass (7 new). Clippy + fmt clean.

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

PR #358, opened 2026-05-19
feat(project): project_to_genomic for c./n./r. inputs (closes #327)

nh13:
## Summary

Adds `VariantProjector::project_to_genomic`: transcript-coordinate (c./n./r.) → genomic (g.) projection — the inverse of the existing forward `project_variant`. Closes #327, the largest single item on the mutalyzer-normalize parity tracker (#326), with 83 baseline `genomic`-axis FAILs that this unblocks for downstream demotion.

## What's in this PR

- **Public API (one new method):**
  ```rust
  impl<P: ReferenceProvider + Clone> VariantProjector<P> {
      pub fn project_to_genomic(&self, variant: &HgvsVariant) -> Result<HgvsVariant, FerroError>;
  }
  ```
  Idempotent on `Genome` input. Rejects `p./m./o./fusion/Allele/null/unknown` with `UnsupportedProjection`. Rejects transcript-coord inputs missing `Accession.genomic_context` (the parent NG/NC), per the issue body's contract. Honors `c.*N` / `c.-N` (UTR) and `c.N+M` / `c.N-M` (intronic).

- **Python bindings:** `PyVariantProjector.project_to_genomic(variant)` mirrors the Rust API. 4 new tests under `TestProjectToGenomic` in `tests/python/test_variant_projection.py`.

- **Coordinate math:** routes through `data::mapping::CoordinateMapper::cds_to_genome`, which I extended to honor `cds_pos.utr3` (was previously dropping the flag and silently mapping `c.*N` as `c.N`). The fix lives in `cds_pos_to_genome_pos`'s new `cds_to_tx_aware` closure, shared by both intronic and non-intronic arms.

- **Strand handling:** reuses `transform_edit_for_strand` for minus-strand revcomp; swaps genomic interval endpoints on minus to keep the resulting g. interval in canonical low→high order.

- **Tests:** 15 `MockProvider`-backed Rust tests under `project_to_genomic_tests`, all run in CI (no manifest needed). They pin:
  - plus-strand exonic substitution (exact `:g.1003C>A`)
  - minus-strand substitution + revcomp (exact `:g.1005G>A`)
  - minus-strand insertion + inserted-sequence revcomp (exact `:g.1004_1005ins` + `GAT`)
  - intronic `c.10+5` offset → exact genomic position
  - **3'UTR `c.*1A>G` → `:g.1012A>G`** (regression guard for the `utr3` bug above)
  - **5'UTR `c.-1T>G` → `:g.1002T>G`**
[...]

Issue #327, opened 2026-05-19
feat(project): expose c./n. → g. projection in VariantProjector


nh13:
## Summary

`VariantProjector::project_variant` only accepts genomic (`g.`) inputs today; on any `c.`/`n.`/`r.` input it returns `FerroError::UnsupportedProjection { reason: "VariantProjector currently only accepts g. variants" }` (`src/project/projector.rs:255-262`).

The mutalyzer normalizer's `equivalent_descriptions["g"]` shape — and many real-world downstream flows — require *both* directions: ferro can project `g.` forward to `c.`/`n.`/`p.`, but not back. This issue closes that asymmetry.

Surfaced by 83 FAILs on the `genomic` axis of `tests/mutalyzer_normalize_tests.rs`. Burn-down item under #326.

## Reproduction

```rust
use ferro_hgvs::{parse_hgvs, MultiFastaProvider, VariantProjector};
use ferro_hgvs::data::projection::Projector;

let provider = MultiFastaProvider::from_manifest("…/manifest.json")?;
let cdot = provider.cdot_mapper().unwrap().clone();
let projector = Projector::new(cdot);
let vp = VariantProjector::new(projector, /* Arc-wrapped provider */);

let v = parse_hgvs("NG_007485.1(NM_000077.4):c.161_162insATC")?;
let g = vp.project_to_genomic(&v)?; // → expected NG_007485.1:g.28294_28295insATC
```

## Scope

Add `VariantProjector::project_to_genomic(&self, variant: &HgvsVariant) -> Result<HgvsVariant, FerroError>` that:

- Resolves the transcript named in the input's accession.
- Maps each c./n./r. position (including intronic `pos+offset` / `pos-offset`) to the corresponding genomic coordinate via the cdot mapper. The mapper already has `cds_to_genomic`, `cds_to_genomic_with_intron`, and `tx_to_genomic` (`src/convert/mapper.rs:594-661`); the new method composes these.
- Reverse-complements the edit when the transcript is on the minus strand. The forward-projection path already does this in `transform_edit_for_strand` (`src/project/edit.rs`); reuse.
- Returns an `HgvsVariant::Genome(...)` carrying the same `Accession` as the input's parent reference (e.g. `NG_007485.1`) — *not* the chromosome accession. Mutalyzer's `genomic` field shape keeps the parent NG/NC, not the chromosome.

[...]

Issue #334, opened 2026-05-19
fix(normalize): 3' rule edge cases — exon-junction exception + repeat-region depth


nh13:
## Summary

ferro applies the HGVS 3' rule, but produces shifts that disagree with mutalyzer on ~35 inputs across two families:

1. **Exon/exon junction exception**: the HGVS spec carves out an explicit exception — *"deletions/duplications around exon/exon junctions using c., r., or n. reference sequences"* should NOT be shifted across the junction. ferro currently shifts; mutalyzer respects the exception.
2. **Repeat-region depth-by-1**: where the variant sits in a tandem repeat and ferro / mutalyzer both apply 3' shift but the chosen anchor differs by one position.

Surfaced by ~35 FAILs on the `normalized` axis of `tests/mutalyzer_normalize_tests.rs` (combining row 9 `3' shift differs by N bases` and row 3d `5' UTR boundary delins` from `failure-patterns.md`). Burn-down item under #326.

## Spec citations

- HGVS general recommendations, "3' rule":  *"for all descriptions, the most 3' position possible of the reference sequence is arbitrarily assigned to have been changed."*
- HGVS general recommendations, exception:  *"deletions/duplications around exon/exon junctions using c., r., or n. reference sequences"* are exempt.

Per arbitration in `tests/fixtures/mutalyzer-normalize/failure-patterns.md`, ferro is correct in principle (3' rule mandatory) but **wrong in two specific edge cases**: the exon-junction exception and the repeat-depth tie-break.

## Scope

Per-family triage of the affected inputs (full list in `tests/fixtures/mutalyzer-normalize/baseline-failures/normalized.txt`):

- **Exon-junction exception**: in `src/normalize/shuffle.rs` (or wherever the 3' loop terminates), check whether the shift would cross an exon boundary on a c./r./n.-axis variant; if so, halt at the boundary. Should not affect g.-axis behavior.
- **Repeat-region depth-by-1**: investigate whether ferro's repeat termination uses `<` vs `<=`; verify against `SVD-WG009` semantics (referenced by mutalyzer for the same boundary). Pick the spec-correct form and align.

Both fixes are localized to `src/normalize/`; together they likely close all ~35 FAILs.

## Tests

`MockProvider`-backed tests in `tests/normalize_tests.rs`:

- c. del spanning exon boundary where 3' shift would cross the junction → confirm no shift past boundary
[...]

PR #379, opened 2026-05-20
feat(project): VariantProjector accepts c./n./r. inputs

### LINKED ISSUE #328: feat(project): VariantProjector::project should accept c./n./r. inputs
## Summary

`VariantProjector::project_variant` rejects non-genomic inputs with `FerroError::UnsupportedProjection { reason: "VariantProjector currently only accepts g. variants" }` at `src/project/projector.rs:255-262`. Mutalyzer accepts c./n./r. inputs and projects them to (g., c., p.) via the named transcript.

Surfaced by 83 FAILs across `protein_description` and `coding_protein_descriptions` axes of `tests/mutalyzer_normalize_tests.rs`. Burn-down item under #326.

Related: #310 ([PR #313](https://github.com/fulcrumgenomics/ferro-hgvs/pull/313)) fixes non-RefSeq protein prediction + gene-symbol-selector emission inside the existing g.-only path; this issue is the orthogonal "accept c./n./r. at the entry point" gap.

## Reproduction

```rust
let v = parse_hgvs("NG_007485.1(NM_000077.4):c.161_163del")?;
vp.project_variant(&v, "NM_000077.4")?;
// → Err(UnsupportedProjection { reason: "VariantProjector currently only accepts g. variants" })
```

Expected:

```
result.coding   → NG_007485.1(NM_000077.4):c.161_163del          (input echoed; normalized within the transcript)
result.protein  → NG_007485.1(NP_000068.1):p.(Met54_Gly55delinsSer)
```

## Scope

In `project_single_inner` (currently the only path; `src/project/projector.rs:247-461`):

- Replace the early `Genome(_)`-only gate with a dispatch:
  - `Genome(_)` → existing forward path (unchanged).
  - `Cds(_)` / `Tx(_)` / `Rna(_)` → first call the new c./n./r. → g. method from sibling issue **#327** to obtain a `Genome` form, then run the existing g.-anchored projection.
- The `transcript_id` argument should default to the input's own accession when present; explicit override stays supported (callers that fan out via `project_variant_all` still pass an external id).

Dependency: ship sibling issue **#327** (c→g projection) first; this one is a thin dispatch layer on top.

## Tests

`MockProvider`-backed tests in `tests/normalize_tests.rs`:

- c. substitution input → roundtrips to same c., yields p.
- c. deletion (multi-base) → c. unchanged, p. with delins
- n. on non-coding transcript → only c.-shape result (no p.)
- minus-strand c. input → rev-comp during c→g, then forward to c./p. — same result as if input had been g.
- r. input on coding transcript → c. equivalent + p. via existing pipeline

Demotes 83 inputs from `baseline-failures/protein_description.txt` and `baseline-failures/coding_protein_descriptions.txt` (overlap counted once).

nh13:
## Summary

`VariantProjector::project_variant_inner` previously required `HgvsVariant::Genome` and rejected non-g. inputs with `FerroError::UnsupportedProjection { reason: \"VariantProjector currently only accepts g. variants\" }`. Mutalyzer accepts c./n./r. inputs and projects them to (g., c., p.) via the named transcript; this PR closes that entry-point gap by adding a thin dispatch layer on top of the existing g.-anchored projection.

## Fix

In `project_single_inner`:

- **`Genome(_)`** inputs continue to flow through the existing path unchanged — zero-cost change for the common case.
- **`Cds(_)` / `Tx(_)` / `Rna(_)`** inputs are first run through `VariantProjector::project_to_genomic` (the #327 helper that landed via PR #358) to obtain a `Genome` form, then re-enter the existing g.-anchored pipeline. The returned g. variant is in the canonical NG/NC frame, so the downstream codon-fetch and protein-prediction logic is unchanged.
- **Other variant kinds** (Protein, Mt, Circular, RnaFusion, NullAllele, UnknownAllele) error with a kind-named diagnostic instead of the generic \"only accepts g.\" message.

The `transcript_id` argument still drives the projection target. Callers can pass an external `transcript_id` to fan a c. input out across multiple transcripts, mirroring the existing g. behaviour.

## Tests

`tests/issue_328_projector_accepts_cnr.rs` (5 tests):

- `projector_accepts_cds_substitution_input` — `NC_000001.11(NM_TEST.1):c.4C>A` → `(g., c.4C>A, p.(Arg2Ser))`, same projection as parsing the equivalent g. form.
- `projector_accepts_cds_deletion_input` — `c.4_6del` round-trips through the projector with a protein output.
- `projector_accepts_rna_substitution_input` — `r.4c>a` → `c.4C>A` → `p.(Arg2Ser)`.
- `projector_rejects_cds_without_genomic_context` — diagnostic error when the input lacks the NG/NC parent (per `project_to_genomic`'s #327 contract).
- `projector_noncoding_input_currently_errors_through_existing_path` — pins that the n.-axis entry point is unblocked (no more \"only accepts g.\") but the downstream g.→c. path still requires `cds_start`. Flip-able to a success expectation once `project_single_inner` learns to dispatch on coding vs non-coding.

## Out of scope

- **End-to-end n. projection on non-coding transcripts.** The downstream `project_single_inner` g.→c. path calls `mapper.genome_to_cds`, which requires `cds_start`/`cds_end` and returns `InvalidCoordinates` (translated to `TranscriptNotOverlapping`) for non-coding transcripts. This is a separate gap from \"accept the input axis\" — adding n.-axis end-to-end requires teaching the projector to dispatch on the target's coding/non-coding status. The new test pins the current behaviour as a regression marker.
- **Allele inputs.** `project_to_genomic` currently errors on `HgvsVariant::Allele` (per its own scope note pointing at #328); compound-allele projection is a sibling follow-up.

5623/5623 focused tests pass. Clippy + fmt clean.

Closes #328.

### DIFF
--- a/src/project/projector.rs
+++ b/src/project/projector.rs
@@ -536,21 +536,95 @@ impl<P: ReferenceProvider + Clone> VariantProjector<P> {
         })
     }
 
-    /// Project a single (non-allele) g. variant, assuming it has already been
+    /// Project a single (non-allele) variant, assuming it has already been
     /// normalized.
+    ///
+    /// Accepts g./c./n./r. inputs. Non-genomic inputs are first
+    /// projected back to their parent NG/NC reference via
+    /// [`Self::project_to_genomic`], then re-enter the g.-anchored
+    /// projection path. The parent reference must be present in the
+    /// input's `Accession.genomic_context` (per #327); inputs without
+    /// it surface a clear "no parent reference" diagnostic.
+    ///
+    /// **Transcript-id constraint.** For c./n./r. inputs the caller's
+    /// `transcript_id` must equal the input's own
+    /// `accession.transcript_accession()`. Mixing the two (e.g. passing
+    /// `c.4C>A` on NM_FOO.1 but requesting projection against NM_BAR.1)
+    /// is rejected with a clear error rather than silently projecting
+    /// through one transcript's exons then re-projecting through
+    /// another's. For g. inputs there is no such input-transcript, so
+    /// the caller's `transcript_id` is the only target.
+    ///
+    /// **3'-shift asymmetry note (re #334).** This routine receives a
+    /// pre-normalized variant. The c./n./r. normalizer respects the
+    /// HGVS exon-junction exception (does not shift across exons),
+    /// while the g. normalizer does not. Consequently, the same
+    /// biological variant fed as c. vs. g. near an exon junction can
+    /// project to different `(g., c., p.)` tuples. This is intentional:
+    /// the input axis carries the spec-correct semantic for *that*
+    /// axis. Callers who want g.-axis semantics should pass the g.
+    /// variant. The c.-input projection preserves the c.-axis canonical
+    /// form by construction.
+    /// (#328)
     fn project_single_inner(
         &self,
         normalized: &HgvsVariant,
         transcript_id: &str,
     ) -> Result<VariantProjection, FerroError> {
-        // Require a g. variant.
-        let genome_variant = match normalized {
-            HgvsVariant::Genome(g) => g.clone(),
+        // Track which g. variant feeds the downstream pipeline AND will
+        // be reported in `VariantProjection.genomic`. For g. input the
+        // two are the same; for c./n./r. input we project once and use
+        // the projected form for both — the `.genomic` field is the
+        // canonical g. representation of the variant, not the input
+        // axis.
+        let projected_genome: HgvsVariant = match normalized {
+            HgvsVariant::Genome(_) => normalized.clone(),
+            HgvsVariant::Cds(_) | HgvsVariant::Tx(_) | HgvsVariant::Rna(_) => {
+                // Reject transcript_id mismatch up front: projecting a
+                // c.-on-NM_FOO through NM_BAR's exons and back through
+                // NM_FOO's would silently produce nonsensical results.
+                if let Some(acc) = normalized.accession() {
+                    let input_tx = acc.transcript_accession();
+                    if input_tx != transcript_id {
+                        return Err(FerroError::UnsupportedProjection {
+                            reason: format!(
+                                "transcript_id mismatch: input is on {} but projection \
+                                 requested against {}; transcript-coordinate inputs must \
+                                 be projected against their own transcript",
+                                input_tx, transcript_id,
+                            ),
+                        });
+                    }
+                }
+                let g = self.project_to_genomic(normalized)?;
+                match &g {
+                    HgvsVariant::Genome(_) => g,
+                    _ => {
+                        // Defensive: project_to_genomic is documented to
+                        // return Genome for these axes. If a future
+                        // refactor changes that contract, surface a
+                        // clear error rather than silently misprojecting.
+                        return Err(FerroError::UnsupportedProjection {
+                            reason: format!(
+                                "project_to_genomic returned a non-Genome variant for {} input",
+                                normalized.variant_type()
+                            ),
+                        });
+                    }
+                }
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
Addressed both reviewer rounds in amend 13b917c.

**Must-fix (Opus, real correctness bugs):**

1. **`.genomic` field returned the input variant instead of the projected g.** Fixed: the dispatch layer now plumbs the projected `Genome` variant through to `VariantProjection.genomic` for all input axes. For g. input it's the (normalized) input itself; for c./n./r. input it's the variant produced by `project_to_genomic`. The `.genomic` field is now axis-correct (always a `Genome(_)`) regardless of how the caller entered. Added explicit assertions in `projector_accepts_cds_substitution_input` so the contract is pinned.

2. **`transcript_id` mismatch silently accepted.** Fixed: for c./n./r. inputs, the dispatch checks that the caller's `transcript_id` equals the input's `accession.transcript_accession()` and errors with a clear diagnostic on mismatch. New `projector_rejects_transcript_id_mismatch_for_cds_input` test pins this.

3. **3'-shift asymmetry documented + pinned.** Per #334, the c.-axis normalizer respects the exon-junction exception while the g.-axis normalizer does not, so the same biological variant can project to different `(g., c., p.)` tuples depending on input axis. The docstring on `project_single_inner` now documents this as intentional (c.-input projections preserve the c.-axis canonical form by construction). `projector_preserves_cds_axis_normalization_semantics` test pins the contract.

**Nice-to-have (Sonnet + Opus):**

- Added a minus-strand round-trip test (`projector_accepts_minus_strand_cds_input`) — exercises `project_to_genomic`'s rev-comp path + the downstream g.→c. mapping. Pins c.→g.→c. round-trip stability on minus strand, which Opus flagged as the highest-risk path.
- Added the `.genomic` assertion to the c. test so the must-fix #1 regression is now pinned.

**Deferred with rationale:**

- **`extract_contig_and_pos` g.-only gate (Sonnet's main flag)** — this lives in the `project_normalized_all` fan-out path, which is separate scope from `VariantProjector::project`. Tracked: should land as a sibling fix. The issue body of #328 is scoped to the single-tx `project` entry point.
- **Multi-exon, `cds_start > 0`, intronic c., predicted-edit wrapper tests (Opus, Sonnet)** — the existing `tests/projection.rs::intronic_fixture` covers the multi-exon + cds_start > 0 + intronic path for g.-input; the new dispatch layer is a thin pre-filter that doesn't change those downstream paths. Adding the test coverage on c.-input is worthwhile but beyond #328's scope; the round-trip from g. to c. through those paths is already pinned by the existing `tests/projection.rs` suite.
- **Cache the cdot.get_transcript double-lookup (Opus #3 nice-to-have)** — the duplicate lookup happens once per projection call; not a hot path. Worth a refactor but separate cleanup.
- **`let projected` outer binding shape (Sonnet #3)** — restructured to a single `let projected_genome: HgvsVariant = match …` returning the cloned variant directly, removing the late-init pattern.
- **n.-axis canary test being a fragile `is_err()` assertion (Sonnet #2)** — tightened to also assert the message does NOT contain "currently only accepts g. variants", so a regression of the entry-point gate would trip the test. The deeper `TranscriptNotOverlapping` assertion isn't pinned because the upstream path may legitimately return a different diagnostic shape if/when non-coding tx support lands.

5626/5626 focused tests pass (8 in the new file, 3 added on top of the initial 5). Clippy + fmt clean.

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

Issue #129, opened 2026-05-06
Mitochondrial circular reference handling (m. coordinate system) (#81 F1)


nh13:
Tracking issue for #81 item **F1** — wraparound-at-origin and other circular-topology semantics for the `m.` coordinate system. Spun out per PR #106's note that "circular wraparound is its own design" and that the `m.` dimension was deliberately excluded from the `del` 3'-shift coverage matrix.

A behavior-pinning audit landed alongside this issue at `tests/mito_circular_audit.rs`. The asserts there pin "what ferro does today" so any F1 implementation surfaces as an intentional, reviewable diff.

## Why this is its own design

The mitochondrial genome (rCRS, NC_012920.1) is **circular**, length 16569. Position 16569 is followed by position 1 with no gap. Every place in the codebase that today assumes a linear contig has to grow a circular-aware variant — and none of the existing axes (coord system × strand × repeat shape that the A1/A5/A7/B2 matrices cover) generalize cleanly to the circular case.

Three semantic choices have to be made, none of which the spec answers in full:

1. **Range orientation.** HGVS allows `m.<high>_<low>` for ranges that cross the origin (e.g. `m.16569_1del` is a 2-nt deletion across positions 16569 and 1, in that circular order). Ferro's parser today rejects any non-structural range with `start > end` (`parse_genome_interval`'s inverted-range check). F1 must either lift the check on `m.` accessions, or introduce a circular-aware interval type that carries an explicit "wraps origin" flag.

2. **Span / length.** On a circular reference the span of `m.<a>_<b>del` with `a > b` is `(L − a + 1) + b` where `L` is the contig length, not `b − a + 1`. The `python_helpers::compute_span` path already has a defensive `None` for origin-crossing **circular (`o.`)** intervals because the contig length is not available there. The same guard does not exist for `m.`, and indel-length / frameshift heuristics will silently mis-compute on a wraparound `m.` interval if we do allow it through the parser without resolving the contig length.

3. **3'-shift / normalize semantics.** "Most-3'" loses absolute meaning on a circle: shifting is defined modulo `L`, and a homopolymer that straddles the origin is one tract, not two. `normalize_mt` is currently a stub (see `src/normalize/mod.rs::normalize_mt` — "MT variants are similar to genomic … For now, return as-is"). F1 has to define the shift target, the boundary stop conditions, and the canonical render for tracts that wrap.

## Spec citations

[...]

PR #380, opened 2026-05-20
feat(parser): spec-compliant wraparound ranges on m./o. circular refs

nh13:
## Summary

HGVS spec ([SVD-WG006](../blob/main/assets/hgvs-nomenclature/docs/consultation/SVD-WG006.md), accepted into HGVS v19.01) authorises reversed `<high>_<low>` ranges on circular references (`o.`/`m.`) "when the rearrangement includes both the last and first nucleotides of the reference sequence", with explicit examples for **both** `del` (`NC_012920.1:m.16563_13del`) and `dup` (`J01749.1:o.4344_197dup`). The `deletion.md:17` rule + `delins`'s `del + ins` composition extend the exception to `delins`. `ins` and `inv` wraparound forms are spec-silent.

ferro previously rejected the parser-side spec-authorised forms (`del`, `delins`) while silently accepting the spec-unauthorised forms (`dup`, `ins`, `inv`) — the latter with silently mis-computed indel length. Both halves of that asymmetry were pinned in `tests/mito_circular_audit.rs` as a tracking surface for this issue.

## What this PR delivers

- Parser **accepts** wraparound `del`, `delins`, and `dup` on `m.` and `o.`. Round-trip-stable.
- Parser **rejects** wraparound `ins` and `inv` with a clear `Verify` error — the spec gives no authorisation for those reversed forms.
- `normalize_mt` gates wraparound inputs (`start > end`) to the canonicalize-only fallback at the top of the function, removing the dependency on the provider error-path or window-arithmetic fail-safe that Opus flagged.
- 3'-rule wraparound shift remains **disabled**: matches mutalyzer (`mutalyzer/description.py:889` rejects all reversed ranges) and biocommons hgvs (`hgvs/location.py:437` "base start position must be <= end position"). The spec's `general.md:41` "most 3' position possible **of the reference sequence**" wording does not define circular semantics; ferro doesn't invent them.

## Comparison with other tools

| Tool | wraparound `del` | wraparound `dup` | wraparound `ins`/`inv` | wrap 3'-shift |
|---|---|---|---|---|
| **HGVS spec** | Authorised | Authorised (SVD-WG006) | Spec-silent | Undefined |
| **mutalyzer** | Rejected (`ERANGEREVERSED`) | Rejected | Rejected | No |
| **biocommons hgvs** | Rejected ("base start position must be <= end position") | Rejected | Rejected | No |
| **ferro (after this PR)** | **Accepted** ✅ | **Accepted** ✅ | Rejected | No (matches others) |

ferro is the only tool of the three that implements the spec-authorised wraparound exception. mutalyzer and biocommons are both under-compliant on this surface.

## Implementation

`parse_genome_interval` gains a sibling `parse_genome_interval_for_circular` (loosens the fast-path inverted check) plus a `check_circular_reversed_range` post-parse validator that rejects the spec-unauthorised edit kinds. Both `parse_mt_variant` and `parse_circular_variant` route through these, on the bare and `(...)` predicted-form branches. The genomic (`g.`) path is unchanged — linear references still reject reversed ranges per the general 5'→3' rule.

`normalize_mt` adds a top-of-function `if start > end { return mt_fallback(variant); }` gate so wraparound inputs route directly to the canonicalize-only fallback. Doing this here, rather than relying on the provider rejecting `start > end` and the post-arithmetic `rel_end = end - window_start` underflow, keeps the path correct under any window-size configuration.

## Tests

- New `tests/issue_129_mt_circular_wraparound.rs` (12 tests): pins the new accept/reject contract for `m.` and `o.`, including the SVD-WG006 examples verbatim (`m.16569_1del`, `J01749.1:o.4344_197dup`), non-wraparound regression coverage for `dup`/`inv` (must still parse for `start < end`), and the `g.`-axis-must-still-reject negative control. Uses `assert_eq!` round-trip checks for tighter regression coverage.
- `tests/mito_circular_audit.rs` (existing 26 tests + 2 new pins): three audit assertions flipped from "currently rejected"/"silently accepted" to the spec-compliant behaviour. Added sibling pins for wraparound `del` and `delins` indel-length miscomputes (joining the existing wraparound `dup` pin) so the contig-length-aware fix surfaces all three cases together.
- `tests/fixtures/grammar/hgvs_spec_normalization.json`: corpus entry for `NC_012920.1:m.16563_13del` flipped from `parse-error` to `preserved` per the spec-authorised parse.

5640/5640 focused tests pass. Clippy + fmt clean.

## Delivered after initial body (amend `46ecce7`)

- **Bracketed compound alleles with wraparound** now parse on both cis and trans forms. `parse_genome_bracket_member` gains a `kind` parameter and routes through `parse_genome_interval_for_circular` + `check_circular_reversed_range`. Tests added to `tests/issue_129_mt_circular_wraparound.rs`: `assert_roundtrip("{MT}:m.[16569_1del;100A>G]")`, reject test for `m.[16569_1insA;100A>G]`, and trans-form round-trip `m.[16569_1del];[100A>G]`.

## Out of scope (follow-ups)

- **Contig-length-aware indel-length / span math for wraparound del/dup/delins.** Pinned by `audit_mt_wrapping_{del,dup,delins}_indel_length_still_miscomputes_today`. Needs contig-length plumbing from the provider.
- **3'-rule shuffle across the origin.** Intentionally deferred to match mutalyzer/biocommons + strict spec reading. `general.md:41` does not define circular shift semantics.
- **Contig-length validation.** Positions past the contig end (e.g. `m.16570A>G`) still parse; separate concern from wraparound. Pinned by `audit_mt_position_past_end_silently_accepted_today`. Filed as #393.

Refs #129.

### DIFF
--- a/src/hgvs/parser/variant.rs
+++ b/src/hgvs/parser/variant.rs
@@ -235,6 +235,34 @@ fn starts_with_special_pos(bytes: &[u8]) -> bool {
 /// Optimized: fast path for simple positions and ranges (starting with digit)
 #[inline]
 fn parse_genome_interval(input: &str) -> IResult<&str, GenomeInterval> {
+    parse_genome_interval_inner(input, false)
+}
+
+/// Parse a genomic-axis interval, optionally relaxing the inverted-range
+/// check for circular references (`m.`/`o.`).
+///
+/// HGVS spec authorises reversed `<high>_<low>` ranges on `o.`/`m.`
+/// references for **deletions** (`deletion.md:17`) and **duplications**
+/// (`docs/consultation/SVD-WG006.md` line 15 + line 23 example
+/// `J01749.1:o.4344_197dup`, line 33 names "deletions/duplications" as
+/// the spec-authorised scope). `delins` inherits the exception via the
+/// `del + ins` composition. `ins` and `inv` are spec-silent — the
+/// general "5'→3'" rule applies. `check_circular_reversed_range`,
+/// called post-parse by `parse_mt_variant` / `parse_circular_variant`
+/// (and the bracket / unknown-phase / inherited-accession helpers),
+/// enforces the edit-kind authorisation so the rejection has access to
+/// the parsed edit kind.
+///
+/// When `circular = true` the parse-level inverted-range check is
+/// removed for the simple-range fast path; post-parse validation in the
+/// caller catches the spec-unauthorised edit kinds. The complex
+/// uncertain-range paths still reject reversed bounds since the spec
+/// exception only applies to plain `<high>_<low>` reversed forms.
+fn parse_genome_interval_for_circular(input: &str) -> IResult<&str, GenomeInterval> {
+    parse_genome_interval_inner(input, true)
+}
+
+fn parse_genome_interval_inner(input: &str, circular: bool) -> IResult<&str, GenomeInterval> {
     let bytes = input.as_bytes();
     if bytes.is_empty() {
         return Err(nom::Err::Error(nom::error::Error::new(
@@ -258,7 +286,10 @@ fn parse_genome_interval(input: &str) -> IResult<&str, GenomeInterval> {
                 // - either position is special (pter/qter have no comparable base values)
                 // - the edit is a structural variant type that may use inverted coordinates
                 //   (insertions, duplications, inversions, and complex delins)
-                let allows_inverted = remaining.starts_with("dupins")
+                // - this is a circular (`m.`/`o.`) reference. The caller validates
+                //   the spec-authorised edit kinds post-parse (#129).
+                let allows_inverted = circular
+                    || remaining.starts_with("dupins")
                     || remaining.starts_with("inv")
                     || remaining.starts_with("ins")
                     || remaining.starts_with("dup");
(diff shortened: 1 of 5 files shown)


nh13:
Addressed both reviewer rounds in amend eb6b799.

**Must-fix-class (Opus #2, real correctness concern):**

`normalize_mt` would underflow `rel_end = end - window_start` on wraparound inputs under non-default window sizes. Fixed: top-of-function `if start > end { return mt_fallback(variant); }` gate routes wraparound inputs to the canonicalize-only fallback directly, independent of any window-size configuration. Pinned the safety in a comment that names the failure mode.

**Nice-to-have (Opus + Sonnet):**

- Added sibling audit pins for wraparound `del` and `delins` `compute_span` miscomputes (Opus #1). The existing `dup` pin is now joined by `audit_mt_wrapping_del_indel_length_still_miscomputes_today` and `audit_mt_wrapping_delins_indel_length_still_miscomputes_today` so the contig-length-aware fix surfaces all three cases together.
- Added an explicit `_` arm comment in `check_circular_reversed_range` documenting the disposition of `Substitution`, `SubstitutionNoRef`, `Insertion`, `Inversion`, `DupIns`, `Repeat`, `MultiRepeat`, `Identity`, `Conversion`, `Unknown` (Sonnet #2, Opus #4).
- Tightened test style in `tests/issue_129_mt_circular_wraparound.rs`: replaced `out.contains(":m.16569_1del")` with `assert_eq!(out, input, ...)` via a new `assert_roundtrip` helper. Now consistent with `mito_circular_audit.rs` and catches stray leading/trailing chars that `contains` would miss (Sonnet #4, Opus #5).
- Changed PR body's `Closes #129` to `Refs #129` per Opus #6. Issue #129 has multiple semantic decisions; this PR delivers range orientation + parser asymmetry, defers span/length and 3'-shift to documented follow-ups.

**Deferred with rationale (documented in PR body's Out of scope):**

- Bracketed compound alleles with wraparound (Sonnet #3, Opus #3) — `parse_genome_bracket_member` doesn't yet use the circular interval parser. The user can write `m.16569_1del` but not `m.[16569_1del;200A>G]`. Tracked as a follow-up; the audit file pins compound-bracket rejection.
- Stale background paragraphs in `tests/mito_circular_audit.rs` were already partially updated in the original commit; remaining stale phrases in "Scenario 5" block-comment kept for context (they describe the historical state that this PR closed; rephrasing them risks erasing the audit trail).

5640/5640 focused tests pass (12 new + 2 added audit pins). Clippy + fmt clean.

nh13:
Rebased onto current main (`39907be`) and addressed two real signals in amend `46ecce7`:

**1. CI: spec fixture regeneration**

The `--check` step in CI failed because my hand-edit of `tests/fixtures/grammar/hgvs_spec_normalization.json` placed the `m.16563_13del` entry in the wrong sort bucket vs. what `generate_spec_fixture` produces canonically. Regenerated the fixture properly via the generator; content of the row is unchanged (just moved to its sorted position). Summary counter shifts `parse-error: 339 → 338`, `preserved: 651 → 652` reflecting the single status flip.

**2. CodeRabbit + Sonnet + Opus: widen circular handling to bracketed/inherited paths**

Originally documented as out-of-scope but folded in per the standing "fix all feedback" instruction. Threaded the circular interval parser through `parse_genome_bracket_member` (now takes `kind: GenomeKind`) and its callers:

- `parse_genome_kind_compound_allele` (existing kind parameter)
- `parse_genome_trans_allele_shorthand` (genome path, kind=Genome)
- `parse_mt_trans_allele_shorthand` (kind=Mt)
- `parse_circular_trans_allele_shorthand` (kind=Circular)
- `parse_genome_position_unknown_phase` (existing kind parameter)
- The m./o. arms of `parse_variant_with_inherited_accession`

Each circular-host site now uses `parse_genome_interval_for_circular` AND calls `check_circular_reversed_range` post-parse, so the same spec-authorisation rules apply uniformly: `del`/`delins`/`dup` accept reversed ranges; `ins`/`inv` reject.

New regression tests in `tests/issue_129_mt_circular_wraparound.rs`:
- `mt_compound_cis_allele_with_wraparound_del_parses` — `m.[16569_1del;100A>G]`
- `circular_compound_cis_allele_with_wraparound_dup_parses` — `J01749.1:o.[4344_197dup;500A>G]`
- `mt_compound_cis_allele_with_wraparound_ins_is_rejected` — reject propagates into brackets
- `mt_trans_allele_with_wraparound_del_parses` — `m.[16569_1del];[100A>G]`

5689/5689 focused tests pass (16 in `issue_129_mt_circular_wraparound.rs`, up from 12). Spec-fixture `--check` clean. Clippy + fmt clean.

nh13:
@coderabbitai full review

nh13:
Addressed both items from the post-amend review in `fb05758`:

1. **`audit_mt_wrapping_delins_indel_length_still_miscomputes_today`** (`tests/mito_circular_audit.rs:358-372`) — tightened from `assert!(observed.is_some())` to `assert_eq!(observed, Some(16568), ...)`. Now matches the del/dup pin style and will trip loudly if the miscompute drifts.

2. **`parse_genome_interval_for_circular` docblock** (`src/hgvs/parser/variant.rs:244-258`) — updated to reflect the validator's actual policy: `del`/`dup`/`delins` accept reversed ranges (SVD-WG006 line 15 + line 23 example + line 33 scope), `ins`/`inv` reject. The earlier draft said `dup` was unauthorised; the implementation has been authorising it since the original commit, the docblock was just stale.

All audit + suite gates green. Force-pushed (single-commit-clean PR).

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #400, opened 2026-05-21
fix(reference): cdot-fallback transcript synthesis diverges from biocommons on NM_001166478.1 (Pattern J on #325)


nh13:
## Summary

Surfaced by `tests/fixtures/biocommons-normalize/baseline-failures/normalized.txt` (umbrella #325). PR #341 / #331 added a cdot-exon-alignment + genome-FASTA fallback so ferro can resolve transcripts that are missing from the reference FASTA bundle (concretely `NM_001166478.1` / DEFB133). The fallback fires correctly — ferro no longer returns the input unchanged — but ferro's normalize output diverges from biocommons on 15 of the 16 currently-failing rows. The 16th row (`c.59_61del`, `cross=false`) is Pattern E (accepted-divergence per #253) and is out of scope here.

The umbrella issue (#325) was previously ticking Pattern A as resolved because PR #341 closed the *infrastructure* gap; in fact none of the NM_001166478.1 rows have been demoted from the baseline.

## Live divergence table

`axis_normalized` on `origin/main` (commit 93caa5d) with manifest at `/Volumes/scratch-00001/work/clients/fulcrum/ferro-hgvs/data/ferro/manifest.json`:

| Input | direction · cross | biocommons expected | ferro got |
|---|---|---|---|
| `c.2_7delinsTTTAGA` | 3'+cross | `c.3_4delinsTT` | `c.[2G>T;3_4inv;5_7delinsAGA]` |
| `c.2_7delinsTTTAGA` | 5'+cross | `c.3_4delinsTT` | `c.[2G>T;3_4inv;5_7delinsAGA]` |
| `c.30_31insT` | 3'+cross | `c.35dup` | `c.34dup` |
| `c.30_31insT` | 5'+cross | `c.31dup` | `c.30dup` |
| `c.31del` | 3'+cross | `c.35del` | `c.34del` |
| `c.34del` | 5'+cross | `c.31del` | `c.30del` |
| `c.35_36dup` | 3'+cross | `c.36_37dup` | `c.35_36dup` |
| `c.35_36dup` | 5'+cross | `c.35_36dup` | `c.34_35dup` |
| `c.35_36insT` | 3'+cross | `c.35dup` | `c.38dup` |
| `c.35_36insT` | 5'+cross | `c.31dup` | `c.36dup` |
| `c.36_37insTC` | 3'+cross | `c.36_37dup` | `c.35_36dup` |
[...]

Issue #253, opened 2026-05-16
tracking: boundary-spanning del/delins ranges (5'UTR/CDS, CDS/3'UTR) (closes #81 J1)


nh13:
Audit pinning ferro's policy on `c.` / `n.` ranges that span the 5'UTR↔CDS or CDS↔3'UTR (resp. upstream↔transcript / transcript↔downstream) boundary.

## Background

Per #81 J1, the rule "no range crosses 5'UTR/CDS or CDS/3'UTR boundary" was enforced inside the merge pass (#103, closes #89) as a merge barrier — adjacent same-coord-system edits in opposite regions cannot be folded into a single delins.

This audit covers the orthogonal case: user-supplied ranges that already span the boundary in the input.

## Spec position

HGVS Nomenclature v21 does **not** explicitly forbid these forms. `recommendations/DNA/deletion.md` and friends discuss the analogous exon/intron boundary case and accept it. The spec's "deletions extending beyond the transcribed region" prohibition applies to ranges that extend past `-244` / `*2691` (outside the transcript), not the UTR↔CDS interior boundary.

## Decision

Accept on input, preserve verbatim on Display, no warning. This matches existing user expectations and the spec's tolerance.

## Pinned

Cross-product of `del` / `delAGCC` / `delinsACGT` / `dup` / `inv` / `insAAA` × the boundaries \{ 5'UTR↔CDS (`c.-3_5`, `c.-1_1`), CDS↔3'UTR (`c.4548_*3`), entire-CDS span (`c.-3_*3`), `n.` upstream↔transcript (`n.-3_5`), `n.` transcript↔downstream (`n.40_*3`) \}. Same-region positive controls also pinned. Merge-barrier behavior at adjacent UTR/CDS positions remains unchanged (covered by #103).

PR #417, opened 2026-05-22
chore(benchmark): wire supplemental_fasta into manifest after patterns fetch

nh13:
## Summary

`ferro-benchmark prepare ferro --patterns <file>` fetched missing accessions into `supplemental/patterns_transcripts.fna` but did not (a) build a samtools-compatible `.fai` index nor (b) update the manifest's `supplemental_fasta` field. The fetched bases were on-disk but invisible to `MultiFastaProvider::from_manifest`, which only loads the supplemental directory when `supplemental_fasta` is non-null. Net effect: running the documented `ferro-benchmark prepare` recipe + then running the biocommons-normalize tests left the fetched transcripts unused, and ferro silently fell back to cdot-driven base synthesis (PR #341) or version-stripped lookups instead.

This PR adds the missing `index_fasta` + `manifest.save()` calls so the recipe "prepare with patterns → run axis test" works end-to-end with no post-prepare manual steps. It also commits `tests/fixtures/biocommons-normalize/burndown_patterns.txt` — the canonical patterns input for the biocommons-burndown reproducer, one HGVS input per line derived from `cases.json`'s `input` field for all `to_test` cases.

## Reproducible recipe

```bash
ferro-benchmark prepare ferro \
    --output-dir <ref-dir> \
    --patterns tests/fixtures/biocommons-normalize/burndown_patterns.txt
```

On a previously-prepared `<ref-dir>`, this:
- skips all already-downloaded transcripts / genome / cdot / LRG / refseqgene / clinvar (default `skip_existing=true`);
- fetches any biocommons-corpus accessions still missing from the RefSeq bundle (one transcript on the current corpus: `NM_001166478.1` / DEFB133, which has been suppressed from current RefSeq releases);
- builds the `.fai` index;
- updates `manifest.supplemental_fasta`.

## Net effect on `tests/biocommons_normalize_tests::axis_normalized`

On `origin/main`, manifest-mode:

| | Pass | FAIL |
|---|---|---|
| Before (`supplemental_fasta: null`) | 71 | 17 |
| After (this PR's recipe applied) | **77** | **11** |

**13 of the 15 NM_001166478.1 rows that #400 tracked as "manifest-data-only" are demoted** — they now resolve through the GenBank-deposited bases rather than ferro's cdot+genome reconstruction. The 2 NM_001166478.1 remainders are:

- `c.36_37insTC` — separate canonicalisation bug (will track).
- `c.59_61del` — Pattern E (#253, accepted-divergence).

## Note on NM_212556.2 regressions

This PR surfaces 4 previously-passing `NM_212556.2` rows as failing — they were silently using `NM_212556.4` bases via version-strip fallback (the RefSeq bundle ships `.4`, not `.2`). With `.2` now directly resolvable from supplemental, ferro reads `.2` bases and emits different normalize output. 3 of 4 are output-string mismatches; 1 is a panic (`c.1delinsCA`: attempt to subtract with overflow). These are pre-existing bugs surfaced, not introduced, by this change. Will file a separate issue tracking the panic + the canonicalisation differences.

## Test plan

- [x] `cargo build --release --features benchmark --bin ferro-benchmark` clean.
- [x] `cargo clippy --features benchmark --bin ferro-benchmark -- -D warnings` clean.
- [x] End-to-end: `ferro-benchmark prepare ferro --output-dir <ref-dir> --patterns tests/fixtures/biocommons-normalize/burndown_patterns.txt` updates the manifest in place and fetches the missing accession.
- [x] Manifest mirror: `grep supplemental_fasta <ref-dir>/manifest.json` now shows the populated entry (`supplemental/patterns_transcripts.fna`).
- [x] Live xfail count drops from 17 → 11 on `origin/main`.
- [x] Independent Opus-model code-review pass: APPROVE on all 10 checkpoints (code correctness, control-flow gating, fetched=0 recovery path, existence guard, patterns derivation, end-to-end reproducibility, side-effects on existing supplemental files, commit-message accuracy, NM_212556 surfacing analysis, lints).

## Companion

- #325 — biocommons-normalize fixture burn-down (umbrella; this PR drops live xfail from 17 → 11, with all but 1 NM_001166478.1 row demoted).
- #400 — Pattern J (cdot-fallback divergence); originally closed as accepted-divergence pending a manifest fix. This PR provides that fix. Will re-comment with the new outcome (13 demoted, 1 separate bug, 1 Pattern E) once the PR lands.
- #341 — original cdot-fallback infrastructure that, before this PR, was the only path serving NM_001166478.1.


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #411, opened 2026-05-22
feat: contig-length-aware math + wraparound boundary policies for m./o. variants (closes #399)

nh13:
## Summary

Closes #399 — the F1 follow-up to PR #380 — by adding full lifecycle handling for wraparound (origin-crossing) ranges on mitochondrial (`m.`) and circular (`o.`) coordinate axes.

The core addition is `ReferenceProvider::get_sequence_length`, a new default-error trait method implemented across MockProvider, FastaProvider, CachedFastaProvider, MmapFastaProvider, and MultiFastaProvider. Built on this, `compute_span_with_provider` and `get_indel_length_with_provider` compute the spec-correct circular span `(L − start + 1) + end` when a provider is available. The existing no-provider `get_indel_length` is extended to return `None` for wraparound `m.`/`o.` ranges (instead of the previous silently-wrong negative value); three audit pins in `tests/mito_circular_audit.rs` flip from `_still_miscomputes_today` to `_is_none_without_provider`, cross-referencing the new `tests/issue_399_mt_circular_followup.rs` for the provider-backed correct values (`+15`, `−2`, `−1`).

At the conversion boundary, `HgvsToVcfConverter` and all four SPDI entry points (`mt_to_spdi_simple`, `mt_to_spdi_with_provider`, `circular_to_spdi_simple`, `circular_to_spdi_with_provider`) reject wraparound variants with a clear error — VCF and SPDI have no native wraparound representation. `circular_to_spdi_simple` and `circular_to_spdi_with_provider` are new functions that also enable non-wraparound `o.` variants to convert for the first time. The validate web-service handler grows a `wraps_origin: bool` field on `ValidateResponse` (omitted from JSON when `false` via `skip_serializing_if`, so backward-compatible), surfaced via a dual-parse approach that preserves the wraparound signal that the W4001 preprocessor would otherwise destroy. Finally, `merge.rs` receives a defensive `is_wraparound_genome` guard that prevents wraparound `m.` endpoint pairs from being fed to linear-adjacency arithmetic at both the precheck and full-anchor extraction sites.

## What's in scope

- **Task 1 — Provider API.** New `ReferenceProvider::get_sequence_length(&self, id: &str) -> Result<u64, FerroError>` with a default that errors; overrides on all five concrete provider impls plus the `Box<dyn ReferenceProvider>` blanket.
[...]

PR #474, opened 2026-05-28
fix(error-handling): make detect_swapped_positions axis-aware for m./o. wraparound (closes #467)

### LINKED ISSUE #399: normalize/parse/convert: contig-length-aware math + boundary policies for m./o. wraparound (#129 follow-up)
## Context

PR #380 landed the spec-compliant parser and normalize gate for wraparound ranges on `m.`/`o.` circular references (SVD-WG006): `del`/`delins`/`dup` accepted, `ins`/`inv` rejected, `normalize_mt` gates wraparound inputs to canonicalize-only fallback. That closes the immediate trigger for #129/F1.

The remaining F1 surface is bounded and well-localized. This issue tracks it as a single rollup so the dependency chain — everything below leans on a new `Provider` length API — stays visible.

## Pinned by audit tests

`tests/mito_circular_audit.rs` already pins the three downstream consequences:

- `audit_mt_wrapping_dup_indel_length_still_miscomputes_today`
- `audit_mt_wrapping_del_indel_length_still_miscomputes_today`
- `audit_mt_wrapping_delins_indel_length_still_miscomputes_today`

These pins flip green when this issue is closed.

### 1. Provider contig-length API (load-bearing)

Add to `ReferenceProvider` (`src/reference/provider.rs`):

```rust
fn get_seq_length(&self, id: &str) -> Result<u64, FerroError>;
```

Implement across `SeqRepoProvider`, `MockProvider`, and the noop/default impls. No behavior change on its own; foundation for everything below and for #393.

### 2. Wraparound-aware span / indel-length / is-frameshift

`src/python_helpers.rs::compute_span` currently has a `Circular(_)`-only `None` guard:

```rust
if matches!(variant, HgvsVariant::Circular(_)) && end < start {
    return None;
}
```

Replace with provider-backed circular span math for both `Mt(_)` and `Circular(_)` when `end < start`: `span = (L − start + 1) + end`. Cascades through `get_indel_length` and `is_frameshift`. Flips the three audit pins above.

### 3. VCF / SPDI reject-at-boundary

Wraparound `del`/`delins`/`dup` now parses (post-#380) and reaches conversion. Today's behavior on wraparound is silent-wrong or obscurely-erroring:

- `src/vcf/from_hgvs.rs::build_vcf_record` — `Deletion`/`Delins` arms call `get_reference_sequence(chrom, start, end)` with `start > end`; provider behavior varies (mock returns empty, SeqRepo errors).
- `src/spdi/convert.rs::mt_to_spdi_simple` and `mt_to_spdi_with_provider` — pass raw `start`/`end` into `emit_spdi_for_edit`, which assumes linear coordinates.

Neither VCF nor SPDI has a natural representation for wraparound (VCF has no convention for circular-contig records; SPDI is single-edit by definition). Reject at the boundary with a clear `ConversionError` on both paths. No "split into two records" policy.

### 4. Validate handler "wraps origin" flag

`src/service/handlers/validate.rs` `HgvsVariant::Mt` branch currently emits `display: v.loc_edit.location.to_string()` only. Add a structured boolean to the response schema, set when `start > end` on `m.`/`o.`.

### 5. merge.rs defensive guard (belt-and-braces)

`src/normalize/merge.rs::simple_range_for_loc_edit` returns `(Region, start, end)` literally for `Mt(_)`/`Circular(_)` (lines 226/238). The adjacency check `prev.end + 1 == next.start` happens to fail for realistic compound cases, but a defensive `None` return when `start > end` on m./o. removes the foot-gun. Low priority.

## Out of scope

- **3'-rule shuffle across the origin.** PR #380 deliberately matched mutalyzer (`description.py:889`) and biocommons hgvs (`location.py:437`) by leaving this disabled. `general.md:41`'s "most 3' position possible *of the reference sequence*" wording doesn't define circular semantics; ferro doesn't invent them. Closed design call, not a remaining bug.
- **W4004 PositionPastEnd for m. axis** — tracked separately in #393. Uses the same Provider length API (item 1 above) so the two issues land cleanly in either order.

## Code paths confirmed correct as-is

These are listed in #129's "Inventory of code paths to update" but investigation showed no changes are needed:

- `src/liftover/` — takes raw `(contig, pos)`, no HGVS dispatch
- `src/equivalence/checker.rs` — generic over `.accession()`, inherits normalize behavior
- `src/effect/mod.rs` — amino-acid- and indel-length-level classifiers, no HGVS dispatch
- `src/vcf/annotate.rs:252` — `Consequence::Unknown` for Mt is the correct fallback
- `src/project/projector.rs:337` — already rejects Mt for `project_to_genomic`

## Suggested PR ordering

1. Provider `get_seq_length` API — foundation, no behavior change (~150–300 LOC).
2. Wraparound span math + flip the three audit pins (~200–400 LOC). Unblocks #393 too.
3. VCF/SPDI reject + validate flag + merge.rs guard (~150–300 LOC).

Total: ~500–1000 LOC across ~8 files. Could collapse into a single PR (~700–1400 LOC) if reviewers prefer.

## Refs

- #129 (F1 umbrella; PR #380 closed the immediate trigger)
- #393 (W4004 m. position-past-end, uses same Provider length API)
- PR #380 (spec-compliant wraparound parser + normalize gate)
- `tests/mito_circular_audit.rs` (three audit pins that flip with this work)

### LINKED ISSUE #467: normalize: make detect_swapped_positions axis-aware so W4001 skips m./o. wraparound ranges (PR #411 follow-up)
## Summary

PR #411 (closes #399) added wraparound (origin-crossing) handling for `m.`/`o.` axes. To surface the `wraps_origin` signal through `validate_hgvs`, it uses a dual-parse workaround because the W4001 `SwappedPositions` preprocessor in `src/error_handling/corrections.rs` rewrites reversed ranges and would otherwise destroy the wraparound signal (a wraparound range legitimately has `start > end`).

The workaround is pinned in code with `TODO(#399)` comments at both sites in `validate_hgvs`. Issue #399 is now **closed**, so those TODOs no longer point at a live tracker.

## The cleaner fix

Make `detect_swapped_positions` in `src/error_handling/corrections.rs` axis-aware so it skips `.m.` and `.o.` axes, where `start > end` is a valid origin-crossing range rather than a transposition error. With that in place, the dual-parse workaround in `validate_hgvs` can be removed and the `TODO(#399)` comments dropped.

## Acceptance criteria

- `detect_swapped_positions` does not flag `.m.`/`.o.` wraparound ranges as W4001 `SwappedPositions`.
- The dual-parse workaround in `validate_hgvs` is removed and `wraps_origin` is still reported correctly.
- The `TODO(#399)` comments are deleted.
- Existing wraparound coverage in `tests/issue_399_mt_circular_followup.rs` continues to pass; add a case asserting W4001 is **not** emitted for a wraparound `m.`/`o.` range.

## Refs

- Deferred from PR #411 ("Out of scope (intentional)" → "W4001 axis-aware SwappedPositions").

nh13:
## Summary

Closes #467. Makes `detect_swapped_positions` (the W4001 `SwappedPositions` preprocessor) axis-aware so it skips the `.m.` and `.o.` axes, where a range with `start > end` is a valid origin-crossing wraparound (per SVD-WG006) rather than a transposition error. This lets the wraparound signal survive preprocessing, which removes the dual-parse workaround introduced in PR #411.

## Background

PR #411 (closes #399) surfaced `wraps_origin` through `validate_hgvs` / `validate_single` via a secondary strict `parse_hgvs` call: the lenient parser's W4001 auto-correction would otherwise swap the reversed `<high>_<low>` endpoints before grammar parsing and destroy the wraparound signal. Both sites carried `TODO(#399)` comments noting that the cleaner fix was to make `detect_swapped_positions` axis-aware. Issue #399 is now closed, so those TODOs were orphaned.

## Changes

- `src/error_handling/corrections.rs` — `detect_swapped_positions` captures the coordinate axis while locating the marker and returns `None` early for the `m.`/`o.` axes. (Also adds `'o'` to the bare-prefix fallback set so `o.` detection is correct-by-construction rather than incidental; net behavior for `o.` is unchanged — still no W4001.)
- `src/service/handlers/validate.rs` — drops the secondary strict parse in both `validate_single` and `validate_hgvs`; `wraps_origin` is now derived directly from the lenient parse via `variant_wraps_origin`. Deletes both `TODO(#399)` blocks and refreshes the doc comments.

## Tests

New `mod w4001_axis_aware` in `tests/issue_399_mt_circular_followup.rs` (4 cases): a wraparound `m.` range does not emit W4001 and is not rewritten; a wraparound circular (`o.`) dup does not emit W4001; and a linear genomic swapped range still emits W4001 (regression guard). The pre-existing `web-service`-gated `validate_response_wraps_origin_true_for_wraparound_{mt,circular}` tests stay green after the workaround removal.

## Test plan

- [x] `cargo test --features dev` — full suite green
- [x] `cargo clippy --features dev --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all --check` — clean
- [x] `grep -rn "TODO(#399)" src tests` — empty

### DIFF
--- a/src/error_handling/corrections.rs
+++ b/src/error_handling/corrections.rs
@@ -1157,24 +1157,55 @@ fn parse_endpoint_token(
 /// offset-bearing forms (`c.100+5_99+3del`) and 3'UTR `*N` markers
 /// (`c.*5_*1del`). Returns Some with the detected swap information and
 /// the swapped endpoint pair (with offsets and markers preserved).
+///
+/// Axis-aware: the `m.` (mitochondrial) and `o.` (circular) axes are skipped
+/// entirely. On a circular contig a range with `start > end` is a valid
+/// origin-crossing wraparound (SVD-WG006), not a transposition error, so
+/// rewriting it would destroy the wraparound signal. See issue #467.
 pub fn detect_swapped_positions(input: &str) -> Option<DetectedCorrection> {
-    // Find coordinate marker.
-    let coord_markers = [".c.", ".g.", ".n.", ".m.", ".o.", ".r."];
+    // Find the coordinate marker and remember which axis it names so circular
+    // axes (m./o.) can be skipped.
+    let coord_markers = [
+        (".c.", 'c'),
+        (".g.", 'g'),
+        (".n.", 'n'),
+        (".m.", 'm'),
+        (".o.", 'o'),
+        (".r.", 'r'),
+    ];
     let mut search_start = 0;
-    for marker in &coord_markers {
-        if let Some(pos) = input.find(marker) {
-            search_start = pos + marker.len();
-            break;
-        }
+    let mut axis = None;
+    // Pick the earliest (leftmost) marker in the input, not the first marker
+    // kind in `coord_markers`. A later embedded coord prefix (e.g. a source
+    // HGVS appended after the main variant) must not override the axis of the
+    // leading expression, which would parse the wrong range. See issue #467.
+    if let Some((pos, marker_len, marker_axis)) = coord_markers
+        .iter()
+        .filter_map(|(marker, marker_axis)| {
+            input
+                .find(marker)
+                .map(|pos| (pos, marker.len(), *marker_axis))
+        })
+        .min_by_key(|(pos, _, _)| *pos)
+    {
+        search_start = pos + marker_len;
+        axis = Some(marker_axis);
     }
     if search_start == 0 {
-        if let Some(pos) = input.find(['c', 'g', 'n', 'm', 'r']) {
+        if let Some(pos) = input.find(['c', 'g', 'n', 'm', 'o', 'r']) {
             if input.get(pos + 1..pos + 2) == Some(".") {
                 search_start = pos + 2;
+                axis = input[pos..].chars().next();
             }
         }
     }
     if search_start == 0 {
+        return None;
+    }
+
+    // Skip circular axes: a reversed range here is a valid origin-crossing
+    // wraparound, not a swapped-position error.
+    if matches!(axis, Some('m') | Some('o')) {
         return None;
     }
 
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai full review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #473, opened 2026-05-28
normalize: NM_001166478.1 c.36_37insTC ins→dup canonicalisation off-by-one (c.35_36dup vs c.36_37dup) (#400 follow-up)


nh13:
## Summary

After PR #417 demoted 13 of the 15 `NM_001166478.1` divergences originally tracked in #400 (by resolving the GenBank-deposited bases instead of cdot+genome reconstruction), one row still fails as a **canonicalisation bug** distinct from the reference-data story: `c.36_37insTC` (3'+cross).

biocommons expects `c.36_37dup`; ferro emits `c.35_36dup`. This is an off-by-one in the insertion→duplication canonicalisation for this row, not the reconstructed-bases artefact that explained the other rows.

## Detail

From #400's live divergence table:

| Input | direction · cross | biocommons expected | ferro got |
|---|---|---|---|
| `c.36_37insTC` | 3'+cross | `c.36_37dup` | `c.35_36dup` |

With the GenBank-deposited `NM_001166478.1` bases now in play (post-#417), this row is no longer a reconstruction mismatch — the remaining divergence is in ferro's ins→dup canonicalisation/shift for a cross-boundary insertion.

## Acceptance criteria

- `NM_001166478.1:c.36_37insTC` (3'+cross) normalizes to `c.36_37dup` (matching biocommons), or is recorded as accepted-divergence with rationale if ferro is judged correct.
- A targeted regression test pins the resolved behavior.
- The row is demoted from `tests/fixtures/biocommons-normalize/baseline-failures/normalized.txt` if fixed.

## Refs

- Surfaced by PR #417; promised but never filed in the #400 follow-up comment: *"1 still failing (`c.36_37insTC`): a separate canonicalisation bug — will file/track independently."*
- Original Pattern J context: #400. Umbrella: #325.

nh13:
## Investigation: core ins→dup canon is correct with deposited bases

Built a repro using the **real GenBank-deposited `NM_001166478.1` sequence** (186 bp, DEFB133) in a single-exon `MockProvider`, normalizing `c.36_37insTC` with `allow_crossing_boundaries()`:

```
strand=Plus:  3'=c.36_37dup   5'=c.35_36dup
strand=Minus: 3'=c.36_37dup   5'=c.35_36dup
```

[...]

PR #482, opened 2026-05-29
fix(reference): flag cdot base-synthesis divergence risk (closes #471)

### LINKED ISSUE #471: reference: stronger diagnostic for FASTA-vs-cdot reconstruction mismatches (#400 follow-up)
## Summary

When the cdot-fallback transcript-synthesis path fires (a transcript is missing from the FASTA bundle and is reconstructed from cdot exon alignments + genome FASTA), the reconstructed bases can differ from the GenBank-deposited transcript bases — cdot carries no CIGAR/mismatch info to encode those differences (see the #400 investigation, where `NM_001166478.1` reconstructed bases diverge from GenBank with no signal to correct them).

PR #341 already emits a soft warning when the fallback fires ("synthesizing bases via cdot version fallback"). This tracks a stronger signal when the reconstruction is likely to be wrong.

## What needs doing

Surface a stronger diagnostic when FASTA-vs-cdot-reconstruction mismatches are likely — e.g. when cdot exon alignments lack CIGAR data (so per-base mismatches cannot be represented), or when a spot-check / downstream normalization disagrees with an authoritative source. The goal is to make silent reconstruction divergences observable rather than only emitting the generic fallback notice.

## Acceptance criteria

- A clearer warning/diagnostic is emitted when cdot-synthesized bases are at elevated risk of diverging from the deposited transcript (e.g. null-CIGAR exons).
- The existing soft fallback warning behavior is preserved for the low-risk case.
- Not a blocker for normalization correctness; this is observability.

## Refs

- Deferred from #400 ("A future enhancement could surface a stronger signal ... Not blocking").
- Related: #341 / #331 (cdot-fallback path), #325 (umbrella).

nh13:
## Summary

Closes #471.

When a transcript is missing from the FASTA bundle, `synthesize_transcript_from_cdot` reconstructs its bases from the genome FASTA over each exon's genomic span. cdot's GFF3 `Gap` CIGAR encodes only `M`/`I`/`D` (never substitutions), and the synthesis **does not apply the CIGAR at all** — so the reconstruction is faithful only when the alignment is gapless. Previously the only signal was a generic soft "synthesizing from cdot" notice, so silent base divergences (e.g. the `NM_001166478.1` mismatch behind #400/#473) were observable only by external comparison.

This adds a pure, unit-tested classifier and a severity-appropriate diagnostic:

- `classify_cdot_synthesis_risk(&[Option<Vec<CigarOp>>]) -> CdotSynthesisRisk` with variants:
  - `Gapless` — every exon has an `M`-only CIGAR; synthesized length is exact (only un-encodable substitutions could still differ). **Lowest risk.**
  - `NoAlignmentGapData` — empty CIGAR vector or any missing per-exon entry; the synthesis assumes an exact genome match and cannot represent any indel/substitution. **Elevated.**
  - `UnappliedIndels { insertions, deletions }` — at least one CIGAR encodes indels the synthesis drops, so length/content provably diverges. **Strongest signal.**
- `synthesize_transcript_from_cdot` emits a strong, specific `warn!` for the two elevated cases and nothing extra for `Gapless` (the call-site soft notice already fired).

Observability only — **no normalization behavior change**.

## Why

This is the root-cause diagnostic for the cdot-reconstruction-vs-deposited mismatches in the Pattern J family (#400) — including #473's residual `NM_001166478.1:c.36_37insTC` row, which is correct with deposited bases and only diverges under reconstruction. Surfacing the risk turns those silent divergences into actionable log signal.

## Test plan

- [x] 7 unit tests on `classify_cdot_synthesis_risk` (empty / all-None / partial-coverage → `NoAlignmentGapData`; all-Match → `Gapless`; insertion / deletion / indels-dominate-missing-data → `UnappliedIndels` with counts). With a `Gapless`-returning stub, 6 of 7 fail (red→green confirmed).

## Refs

- Umbrella: #325. Family: #400 (Pattern J), #473 (closed — residual folded here). Synthesis path added by #341/#331.

---
Note: commit is unsigned (`--no-gpg-sign`) — 1Password biometric signing was unavailable; please re-sign on merge/rebase.

### DIFF
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -714,6 +714,31 @@ impl MultiFastaProvider {
             }
         }
 
+        // #471: the bases above are pulled straight from the genome without
+        // applying the cdot CIGAR, so flag — at a severity matching the
+        // available alignment evidence — when the reconstruction is at elevated
+        // risk of diverging from the deposited transcript. The call site already
+        // emitted a soft "synthesizing from cdot" notice; the gapless case adds
+        // nothing further.
+        match classify_cdot_synthesis_risk(&tx.exon_cigars) {
+            CdotSynthesisRisk::UnappliedIndels {
+                insertions,
+                deletions,
+            } => warn!(
+                "{id}: cdot exon alignment encodes {insertions} inserted and {deletions} deleted \
+                 base(s) that base synthesis does not apply — the synthesized sequence will \
+                 diverge from the deposited transcript in length and content; treat normalize \
+                 output for {id} as low-confidence (issue #471)"
+            ),
+            CdotSynthesisRisk::NoAlignmentGapData => warn!(
+                "{id}: cdot record carries no per-exon alignment-gap (CIGAR) data — synthesized \
+                 bases assume an exact genome match and cannot represent transcript-specific \
+                 indels or substitutions, so normalize output for {id} may diverge from the \
+                 deposited transcript (issues #471, #400)"
+            ),
+            CdotSynthesisRisk::Gapless => {}
+        }
+
         // Project cdot exons + CDS metadata onto the transcript struct using the
         // same conventions as the FASTA-backed `get_transcript` path above so
         // downstream consumers see one shape regardless of which path served them.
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #326, opened 2026-05-19
tracking: ferro-hgvs ↔ mutalyzer normalize parity


nh13:
Tracking issue for parity between `ferro_hgvs::normalize` and `mutalyzer/mutalyzer` (v3.1.1, pinned at upstream commit `00045d6`). Surfaced by the `tests/mutalyzer_normalize_tests.rs` corpus runner introduced in #323.

## Starting line

When #323 merged (2026-05-19), the baseline was 494 axis-FAILs across 8 output axes:

| Axis | PASS | FAIL |
|---|---:|---:|
| `normalized` | 94 | 149 |
| `genomic` | 8 | 120 |
| `protein_description` | 0 | 71 |
| `coding_protein_descriptions` | 0 | 51 |
| `errors` | 0 | 57 |
| `rna_description` | 0 | 16 |
| `infos` | 0 | 22 |
| `noncoding` | 0 | 8 |
| **Total** | **102** | **494** |

The per-axis FAIL inputs live in `tests/fixtures/mutalyzer-normalize/baseline-failures/<axis>.txt`. Root-cause groupings + HGVS spec arbitration are in `tests/fixtures/mutalyzer-normalize/failure-patterns.md`.

## Current state (committed `baseline-failures/` @ `origin/main` `635b396`, 2026-06-15)

Counts below are the committed expected-FAIL membership per axis (the burn-down metric: the tracker is done when every file is empty). They are not a fresh live PASS/FAIL run — a manifest-backed run is still needed to re-arbitrate which residual rows are ferro bugs vs accepted divergences (see Phase 2 caveat below).

| Axis | FAIL | Δ vs starting line |
|---|---:|---:|
| `normalized` | 71 | −78 |
| `genomic` | 83 | −37 |
| `protein_description` | 52 | −19 |
| `coding_protein_descriptions` | 46 | −5 |
| `errors` | 19 | −38 |
| `rna_description` | 16 | 0 |
| `infos` | 7 | −15 |
| `noncoding` | 8 | 0 |
| **Total** | **302** | **−192** |

[...]

PR #483, opened 2026-05-29
fix(cdot): plumb transcript protein accession from cdot JSON

nh13:
cdot records a per-transcript CDS protein accession (e.g. `"protein": "NP_000068.1"`), but `RawCdotTranscript` never deserialized it and both conversion paths (`from_genome_build`, `from_flat_format`) hardcoded `protein: None`. The projector therefore never saw the real protein id and fell back to the `NM_* → NP_*` number-preserving inference — **wrong whenever the NP number differs from the NM number** (e.g. `NM_000077.4`’s protein is `NP_000068.1`, not `NP_000077.4`).

This affects every protein prediction backed by a cdot store that carries `protein` ids, independent of the mutalyzer corpus.

**Fix:** deserialize the `protein` field on `RawCdotTranscript` and surface it on `CdotTranscript.protein`, so the projector’s existing priority-1 `cdot_protein` path (`projector.rs`) emits the correct `p.` accession instead of the inferred one.

Adds a loader regression test (`test_protein_accession_parsed_from_cdot`). Full `cargo nextest run --features dev` suite green (6400 passed); no other test changed behavior.

**Scope note (Refs #326):** this is a prerequisite for the mutalyzer `protein_description` / `coding_protein_descriptions` burn-down but does not by itself demote corpus rows — those also need the runner `transcript_of` fix (returns the compound accession instead of the bare transcript, tripping the projector’s transcript-id-mismatch guard) and a decision on whether protein output should carry the `NG(NP)` genomic-context wrapper mutalyzer emits. Both are documented in the #326 arbitration.

⚠️ Committed with `--no-gpg-sign` (1Password biometric unavailable while away); re-sign on rebase before merge.

### DIFF
--- a/src/data/cdot.rs
+++ b/src/data/cdot.rs
@@ -92,6 +92,12 @@ struct RawCdotTranscript {
     gene_version: Option<String>,
     #[serde(default)]
     biotype: Option<Vec<String>>,
+    /// Protein accession for the transcript's CDS (e.g. `NP_000068.1`). cdot
+    /// records this per-transcript; plumbing it through lets the projector
+    /// emit the correct `p.` accession instead of falling back to the
+    /// (frequently wrong) `NM_*` → `NP_*` number-preserving inference.
+    #[serde(default)]
+    protein: Option<String>,
     #[serde(default)]
     genome_builds: Option<HashMap<String, RawGenomeBuild>>,
     /// CDS start in transcript coordinates (0-indexed)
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #478, opened 2026-05-28
docs(test): design — annotated, self-correcting, hermetic conformance harness

nh13:
## Summary

RFC-style design doc (`docs/CONFORMANCE_HARNESS_DESIGN.md`) proposing a redesign of the biocommons/mutalyzer normalize conformance suites. It targets the **root defect** behind the recurring #325 burn-down churn rather than patching symptoms.

**Root cause:** the divergence record (`tests/fixtures/biocommons-normalize/baseline-failures/normalized.txt`) is a hand-maintained, CI-unchecked, undifferentiated flat list whose correctness silently depends on out-of-band reference data. So it conflates intent (accepted-divergence vs. known-bug), drifts silently, and cannot gate merges.

**The design, five pillars** (1–2 are the high-value core, independently shippable):

1. **Annotated dispositions in `cases.json`** — per case/axis: `match` / `accepted_divergence{reason, spec_citation, ferro_output}` / `known_bug{tracking_issue, ferro_output}`.
2. **XPASS detection** — a `known_bug` that starts passing *fails the test loudly*, forcing annotation removal. Makes burn-down self-correcting; the stale-row failure mode that triggered this becomes structurally impossible.
3. **Reference-version pinning + load-time spot-check** — generalizes #471 from a soft warning into a hard guard; kills phantom divergences from wrong accession versions.
4. **Hermetic fixture-backed conformance** — commit the small reference window each case needs so the merge-gating test runs in CI without the multi-GB manifest, reproducible by any contributor.
5. **Test tiering + generated/`--check`-verified summary** — per-PR hermetic gate; nightly full-manifest sweep; any flat summary is generated, never hand-written.

## Why now

This is grounded in concrete failures from the last burn-down cycle, all documented in the doc:
- PR #410 fixed `c.1_2insCA` but its row sat stale in the file for a week (→ pillar 2).
- PR #417 (supplemental-FASTA) demoted 13 of 15 `NM_001166478.1` rows; the file still lists all 16 (→ pillars 1, 5).
- An `NG_029146.1`-vs-`.2` version mismatch produced a phantom divergence that cost ~an hour to diagnose (→ pillar 3).

## Scope of this PR

[...]

PR #490, opened 2026-05-29
feat(reference): strict exact-version transcript resolution (#478 pillar 3)

nh13:
This adds an opt-in strict-resolution capability so a caller (e.g. the conformance harness) can require that a transcript is served at the *exact* requested versioned accession, getting a structured `Err` instead of a silently-substituted transcript.

## Problem

`MultiFastaProvider` resolves a transcript by accession through a fallback chain: exact versioned FASTA -> version-strip sibling fallback (e.g. requested `NM_212556.2` served from `NM_212556.4`) -> cdot exon-alignment synthesis from the genome. Each fallback emits a `warn!` but still returns a transcript. This silent substitution causes "phantom" divergences in the conformance suites: e.g. a manifest that ships `NG_029146.2` will serve it for a request for `NG_029146.1`, producing a different normalize result with no hard signal.

## Change (additive, opt-in)

- New `pub fn MultiFastaProvider::get_transcript_strict(&self, id: &str) -> Result<Transcript, FerroError>`: returns the transcript only if its bases are directly available at the exact requested versioned accession; otherwise returns `FerroError::TranscriptVersionNotExact { requested }` **without** falling back to a sibling version or cdot-genome synthesis.
- New `pub fn MultiFastaProvider::has_transcript_exact(&self, id: &str) -> bool`: backs the exactness check (direct lookup in the FASTA/supplemental index, which is keyed by the full versioned accession). Deliberately does **not** consult the version-strip / chromosome-alias fallback in `resolve_name`.
- New `FerroError::TranscriptVersionNotExact { requested: String }` variant (mapped to the `ReferenceNotFound` error code), threaded through the mutalyzer-map exhaustiveness coverage (`debug_variant_name`, the reverse map, and `FERRO_NO_MUTALYZER_EQUIV`).

When the exact entry is present, the strict path delegates to the same construction body as the lenient path, so the returned `Transcript` is byte-for-byte identical for an exact hit (cdot metadata enrichment is preserved) -- only the *acceptance criterion* differs. It is the cdot-genome base *synthesis* fallback that strict refuses, not cdot metadata on an exact FASTA entry.

## Safety / no behavior change

The lenient `get_transcript` and the `ReferenceProvider` trait impl are unchanged. The strict method is a separate, additive entry point; existing callers see no difference.

## Test plan

Hermetic unit tests in `multi_fasta.rs` (temp FASTA + .fai, mirroring `test_load_fai_index`):

- `test_get_transcript_strict_returns_exact_version` -- exact version present in FASTA -> strict returns `Ok` with the right sequence, matching the lenient result.
- `test_get_transcript_strict_rejects_sibling_version_but_lenient_falls_back` -- only the `.4` sibling present, `.2` requested -> strict returns `TranscriptVersionNotExact`, while lenient still falls back to the `.4` sibling (proves existing behavior unchanged).
- `test_get_transcript_strict_refuses_cdot_synthesis_but_lenient_synthesizes` -- genome + cdot present, transcript absent from the FASTA index -> strict refuses the reconstruction, while lenient still synthesizes from the cdot exon alignment.

Verification:

- `cargo fmt --all -- --check` -- clean
- `cargo clippy --features dev --all-targets -- -D warnings` -- clean
- `cargo test --features dev --lib reference::multi_fasta` -- 29 passed (incl. the 3 new tests); full error/mutalyzer suite (403 tests) green.

## Refs

Implements design pillar 3 (reference-version pinning + load-time spot-check), `docs/CONFORMANCE_HARNESS_DESIGN.md`. Refs #471, #478, #325.

Commit is unsigned (--no-gpg-sign); please re-sign on merge/rebase.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -305,6 +305,22 @@ pub enum FerroError {
     #[error("Reference not found: {id}")]
     ReferenceNotFound { id: String },
 
+    /// A transcript was requested at an exact versioned accession under
+    /// strict resolution, but only a sibling version or a cdot-genome
+    /// reconstruction could satisfy it. Returned by
+    /// [`MultiFastaProvider::get_transcript_strict`] when the requested
+    /// version is not directly present, so callers (e.g. a conformance
+    /// harness) get a hard signal instead of a silently-substituted
+    /// transcript. See design pillar 3 (reference-version pinning).
+    ///
+    /// [`MultiFastaProvider::get_transcript_strict`]: crate::reference::multi_fasta::MultiFastaProvider::get_transcript_strict
+    #[error(
+        "transcript {requested} is not available at the exact requested version \
+         (a sibling version or cdot reconstruction would be required; strict \
+         resolution refuses to substitute)"
+    )]
+    TranscriptVersionNotExact { requested: String },
+
     /// Variant spans an exon-intron boundary
     #[error("Variant spans exon-intron boundary at exon {exon}: {variant}")]
     ExonIntronBoundary { exon: u32, variant: String },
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #481, opened 2026-05-29
fix(test/mutalyzer-normalize): accept infos codes ferro does not model (22→11)

nh13:
The `infos` axis hard-failed any case where an expected mutalyzer `I*` code mapped to `NO_FERRO_INFO_EQUIV`. Per the spec arbitration in #326, those codes (`ISORTEDVARIANTS`, `ICORRECTEDCOORDINATESYSTEM`, `ICORRECTEDPOINT`, `ICORRECTEDVARIANTTYPE`, `ICORRECTEDSELECTORID`, `IWHOLETRANSCRIPTEXON`, `IMRNAGENOMICTIP`, `ICORRECTEDLRGREFERENCE`) are mutalyzer **internal** diagnostics describing what mutalyzer did to the input — they are absent from the HGVS nomenclature spec (0 matches in `assets/hgvs-nomenclature`), and ferro intentionally does not model them. Requiring ferro to mirror them is not a spec obligation.

**Change:** an expected code with no ferro equivalent is now an **accepted divergence** (tallied separately, policy `mutalyzer-info-no-spec-equivalent`) rather than a FAIL. Real signal is preserved:

- ferro must still emit every *modelled* code — a mapped code ferro fails to emit (e.g. `SHUFFLE_APPLIED`) stays a hard FAIL.
- ferro **over-emission** (an info mutalyzer did not report) stays a hard FAIL.
- parse/normalize/panic errors stay hard FAILs.

**Result:** `infos` 22 FAIL → **11 `divergence_accepted` + 11 FAIL**. The 11 remaining are genuine divergences, not chatter:
- ferro over-emitting `SHUFFLE_APPLIED` where mutalyzer reported only `ISORTEDVARIANTS` (4),
- ferro not emitting `SHUFFLE_APPLIED` where mutalyzer expected the equivalent of `ICORRECTEDPOINT` (2),
- parse failures on malformed (`>CA`) / cross-reference (`insNM_...:c....`) inputs (5).

Removes the 11 accepted inputs from `baseline-failures/infos.txt`. Verified against a `FERRO_MANIFEST`-backed run; zero regressions (the 11 kept are a strict subset of the prior baseline). Test-harness adjudication change only; no `src/` change, CI mock-pin unaffected.

[...]

PR #489, opened 2026-05-29
test(mutalyzer): known_bug disposition + XPASS detection (#478)

nh13:
## Summary

Adds a `known_bug` disposition plus **XPASS detection** to the mutalyzer conformance harness (`tests/mutalyzer_normalize_tests.rs`), mirroring the biocommons machinery landed in #484.

The harness already supported `accepted_divergence` (intentional, spec-allowed, non-failing divergences). This adds the complementary disposition and closes a staleness gap:

1. **`known_bug` annotation** — ferro is *wrong* on an axis; the divergence is a tracked bug (by `tracking_issue`) rather than a policy decision. A string mismatch on the annotated axis routes into a tracked-but-non-failing `known_bug` bucket (xfail) instead of `fail`. It is reported in the per-axis summary line and is *not* written to `baseline-failures/<axis>.txt`.

2. **XPASS detection** — previously `record()` counted *any* match as a pass, even when an `accepted_divergence` or `known_bug` annotation was present for the current axis. That silently masked stale annotations (the divergence/bug had been fixed but the annotation lingered). Now a match on an annotated axis **fails loudly** so the stale annotation — and, for `known_bug`, the now-fixed fixture row — is cleaned up and the row demoted to a normal passing case.

## Safety: purely additive

- Rows with **no annotation** behave exactly as before: PASS on match, FAIL on mismatch or `Err`.
- Annotations only excuse a *string mismatch* from a successful run. An `Err` (panic / parse / normalize failure) **still fails** even with a `known_bug` annotation — same contract as `accepted_divergence` and `spec_citation`.
- `known_bug.axis` is the closed `Axis` enum, so a typo is rejected at fixture-parse time (consistent with the existing `accepted_divergence`/`spec_citation` annotations).
- Only `tests/mutalyzer_normalize_tests.rs` changed.

## Test plan

- `cargo fmt --all -- --check` — clean
- `cargo clippy --features dev --all-targets -- -D warnings` — clean
- `cargo test --features dev --test mutalyzer_normalize_tests` — 32 passed; 0 failed (manifest-gated axis tests skip without a manifest, as expected)

New unit tests in `comparator_tests`:
- `case_parses_with_known_bug`
- `tally_known_bug_on_matching_axis`
- `tally_known_bug_xpass_fails`
- `tally_accepted_divergence_xpass_fails`
- `tally_err_with_known_bug_still_fails`

Refs #478 (design), #484 (biocommons sibling), #325.

[...]

PR #491, opened 2026-05-29
refactor(test/mutalyzer-normalize): single-process baseline regeneration

nh13:
Fixes the intermittently-dropped `/tmp/ferro-xfail/<axis>.tsv` artifacts seen when regenerating baselines.

**Root cause:** the per-axis `axis_*` tests each wrote their FAIL list as a best-effort side effect (`let _ = fs::write`) and then **panic** on divergence. Under nextest’s parallel, per-test-**process** model, those writes were occasionally lost (observed: the three slowest axes — `genomic`/`errors`/`coding_protein` — dropped their `.tsv`; not reproducible deterministically, i.e. a race). Relying on best-effort writes from parallel panicking processes is inherently fragile.

**Fix — separate the gate from the generator:**
- Extract each axis body into `run_<axis>() -> Option<AxisTally>`.
- `axis_*` tests become pure gates: `run` + new `AxisTally::assert_clean` (summary + diagnostics + assert; **no filesystem I/O**).
- Add `regenerate_baselines` (`#[ignore]`d): runs all 8 axes in **one process** and calls `AxisTally::write_artifacts` for each — writes are race-free by construction.
- Split `finish()` → `assert_clean()` + `write_artifacts()`; `summary_line` takes `Option<&Path>`.

**Run the generator:**
```
FERRO_MANIFEST=/path/to/manifest.json \
  cargo nextest run --features dev -E "test(regenerate_baselines)" --run-ignored only --no-capture
```

**Verification:** gate suite passes via skip with no manifest (23 passed, 0 fail); full `cargo nextest run --features dev` green (6400 passed); the generator wrote **8/8** axis artifacts in a single run. No behavior change to the gates.

> Stacked on #481 (both touch this test file). Base will retarget to `main` once #481 merges.

Refs #326.

### DIFF
--- a/tests/mutalyzer_normalize_tests.rs
+++ b/tests/mutalyzer_normalize_tests.rs
@@ -14,10 +14,19 @@
 //! absent (CI by default), every axis test reports `skipping — no manifest`
 //! and exits 0. Same convention as `tests/real_data_normalization_tests.rs`.
 //!
-//! Each axis test also writes its current FAIL list to
-//! `/tmp/ferro-xfail/<axis>.{txt,tsv}` so contributors can regenerate the
-//! committed `baseline-failures/` snapshot or feed
-//! `tests/fixtures/mutalyzer-normalize/failure-patterns.md`.
+//! The per-axis `axis_*` tests are pure pass/fail gates and do NOT touch the
+//! filesystem. To (re)generate the FAIL lists under `/tmp/ferro-xfail/<axis>.{txt,tsv}`
+//! — used to seed the committed `baseline-failures/` snapshot and
+//! `failure-patterns.md` — run the single-process `regenerate_baselines`
+//! generator (it is `#[ignore]`d in the normal suite):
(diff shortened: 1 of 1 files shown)


nh13:
Closing — superseded by the conformance-harness redesign (#478) and its mutalyzer implementation #489, which rewrites `AxisTally::record()` in this same file and moves the source of truth from the flat `baseline-failures/*.txt` into per-case `cases.json` annotations + XPASS detection (deleting/regenerating the flat file). This PR's single-process *baseline regeneration* targets a mechanism the redesign removes, and would conflict heavily with #489.

One durable idea worth carrying into #489: writing the `/tmp` artifacts as best-effort side effects from the parallel, panicking per-axis `axis_*` processes is flaky — a `.tsv` is intermittently dropped. If any artifact generation survives the redesign, do it in a single dedicated entry point, not as a side effect of the gate tests.

PR #494, opened 2026-05-29
docs(reference): RFC — retire silent cdot-genome reconstruction (deposited-or-error)

nh13:
## Summary

RFC (`docs/RECONSTRUCTION_POLICY_DESIGN.md`) for the long-term correct fix to the cdot-reconstruction divergence class (#471/#400): **stop silently fabricating transcript sequence from the genome.**

The genome is the wrong source for transcript bases whenever the deposited transcript differs from it:
- **substitutions** are invisible to cdot's `M`/`I`/`D` CIGAR (emitted as the genome base — this is #400/#473);
- **indels** corrupt the synthesized exon length and therefore **all downstream transcript coordinates** (silent corruption that `axis_normalized` misreads as a ferro divergence).

No reconstruction logic recovers bases the genome lacks. So the target model is: **deposited bases are the single source of truth; cdot is for coordinates only; missing deposited bases → structured error, not fabrication; close the gap by completing the FASTA bundle upstream.**

## Why an RFC and not just a patch

This reverses a deliberate prior decision — the cdot-genome fallback was added on purpose in #341/#331 ("a possibly-wrong answer beats no answer"). Flipping it to *error-on-missing* changes ferro's public contract, so the default-flip is framed as an explicit, gated maintainer decision rather than folded silently into a refactor.

## Phased migration (first phases non-breaking; builds on #490)

1. **Resolution-policy seam** — add `TranscriptSourcePolicy { DepositedOnly, ReconstructBestEffort }`, default `ReconstructBestEffort` (today's behaviour, no functional change). `DepositedOnly` reuses PR #490's exact-version resolver.
2. **Refuse-on-corruption** — while reconstruction still runs, upgrade PR #482 from *warn* to *error* on any exon with CIGAR indels **or** a span-length mismatch (`e1−e0 ≠ e3−e2`, which catches no-CIGAR net indels). Turns silent coordinate corruption into a loud error. (This is "Option D" from the #471 discussion.)
3. **Flip the default to `DepositedOnly`** — the contract-changing, gated step; proceed once the bundle is complete (Phase 4) and the error-on-missing behaviour is agreed.
4. **Complete the bundle** — extend `ferro prepare` to fetch deposited bases (#417 mechanism) for every needed transcript, so deposited-only is the *complete* answer (this is what actually demotes the #400/Pattern J rows).

## Scope of this PR

Doc only — no code. Establishes the principle, the migration, and the explicit decision point. If accepted, Phases 1–2 are safe to implement immediately (additive + silent-wrong→loud-error); Phase 3 awaits sign-off.

## Refs

- #471 (cdot-mismatch), #400 (Pattern J), #325 (umbrella), #478 (conformance harness — strict sourcing is what makes `axis_normalized` a trustworthy gate). Builds on #341/#331, #417, PR #482, PR #490.

### DIFF
--- a/docs/RECONSTRUCTION_POLICY_DESIGN.md
+++ b/docs/RECONSTRUCTION_POLICY_DESIGN.md
@@ -1,0 +1,64 @@
+# Retire silent cdot-genome reconstruction — deposited-or-error transcript sourcing
+
+Parent context: [#325](https://github.com/fulcrumgenomics/ferro-hgvs/issues/325) burn-down, [#471](https://github.com/fulcrumgenomics/ferro-hgvs/issues/471) (cdot-reconstruction mismatch), [#400](https://github.com/fulcrumgenomics/ferro-hgvs/issues/400) (Pattern J).
+
+Builds on landed work: [#341](https://github.com/fulcrumgenomics/ferro-hgvs/pull/341)/#331 (the fallback this proposes to demote), #417 (deposited bases via supplemental FASTA), PR #482 (synthesis-risk diagnostic), PR #490 (strict exact-version resolution).
+
+## Principle
+
+ferro must not silently fabricate transcript **sequence** from the genome. The genome is the wrong source for transcript bases whenever the deposited transcript differs from it — by substitutions (invisible to cdot's `M`/`I`/`D` CIGAR), indels (which corrupt the synthesized length and therefore *all* downstream transcript coordinates), or non-reference haplotypes. No reconstruction logic can recover bases the genome does not contain.
+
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #500, opened 2026-05-31
feat(normalize): resolve legacy gene-symbol selectors to transcript accessions on genomic references (NG_/NC_/LRG)


nh13:
When a coding (`c.`/`n.`) variant is described against a genomic reference (`NG_`, `NC_`, `LRG_`) using a legacy gene-symbol selector — e.g. `NG_012337.1(TIMM8B_v001):c.12_13insGATC` — ferro preserves the gene-symbol selector verbatim. The HGVS specification strongly prefers a transcript (or protein) accession in that selector slot, not a gene symbol. ferro's output is valid HGVS but not the spec-preferred canonical form, and `normalize()` should converge on the preferred form.

## Spec basis

From `assets/hgvs-nomenclature/docs/background/refseq.md`:

- L38–42: parenthetical specifications on `NG_`/`NC_` references "include transcripts (`NM_004006.2`) and proteins (`NP_003997.1`). Gene symbols should **not** be used as specification." The only exception is a genomic reference with no transcript reference available (e.g. the mitochondrion) — which does not apply here (all affected genes have NM_ transcripts).
- L138–139: for a `c.` coordinate on a genomic reference, the **transcript** used should be indicated, e.g. `NG_012232.1(NM_004006.2):c.357+1G>A`.

So the spec-preferred output is `NG_(NM_):c.…` (which Mutalyzer emits); ferro's `NG_(GENE_v001):c.…` is valid but strongly non-preferred. This is the "rewrite into the form preferred by the HGVS spec" class historically tracked under #81 §I (Reference sequences and accessions).

## Why this wasn't caught before

The gene-symbol-selector policy (#121, via #81 item I3, PRs #135/#144) was reasoned and tested entirely on the `NM_(GENE)` form, where the gene symbol is redundant disambiguation on top of an already-named transcript — preserving it there is correct and remains unchanged. No test ever covered a gene symbol acting as the **sole transcript selector** on an `NG_`/`NC_`/`LRG_`. This issue addresses only the genomic-reference case.

## Current feasibility (why this is a feature, not a quick fix)

ferro cannot currently produce the spec-preferred output:

1. No code path resolves the legacy LOVD `_v001` locus-version suffix.
2. The disambiguating data (`legacy_transcripts_metadata` in the reference manifest) is unpopulated; the existing `legacy_*` machinery is for GenBank accessions (ClinVar), not LOVD gene-symbol-versions.
[...]

Issue #121, opened 2026-05-06
display: preserve gene-symbol selector on round-trip (#81 I3 follow-up)


nh13:
Per HGVS Nomenclature, the gene-symbol selector (e.g. `(COL1A1)` in `NM_000088.3(COL1A1):c.459A>G`) is informational disambiguation. It is optional in the canonical reference; the principled round-trip policy is **preserve when present in input; do not synthesize when absent**.

PR #70 (commit `b34ce35`) made parsing accept the gene selector on non-RefSeq accessions and store it on the variant struct as `gene_symbol: Option<String>`. The round-trip emission policy was left unresolved (see #81 item I3).

## Current behavior

Today, every variant `Display` impl in `src/hgvs/variant.rs` emits `{accession}:{type}.{loc_edit}` and **drops** `gene_symbol`, even when it was present in the input. Examples:

- `GenomeVariant::fmt` — `write!(f, "{}:g.", self.accession)?;`
- `CdsVariant::fmt` — `write!(f, "{}:c.", self.accession)?;`
- same for `TxVariant`, `RnaVariant`, `ProteinVariant`, `MtVariant`, `CircularVariant`
- `AlleleVariant` compact form (`write_compact_prefix`) also drops it

The one exception is `RnaFusionBreakpoint::fmt`, which already preserves the gene symbol in parens.

The current behavior is pinned by tests: `test_compound_ref_with_gene_symbol`, `test_cds_variant_with_gene_symbol`, and `test_genome_variant_with_gene_symbol`. The new tests in `tests/gene_selector_roundtrip.rs` (added in PR for #81 I3) explicitly document the strip-on-Display behavior with TODO references to this issue.

## Proposed behavior

When `self.gene_symbol` is `Some(g)` and the accession does not already include a compound-ref `(…)` wrapper, emit `{accession}({g}):{type}.{loc_edit}`. When `gene_symbol` is `None`, emit the bare form unchanged. Never synthesize a gene symbol that wasn't in the input.

Spec rationale: hgvs-nomenclature.org states the selector is an optional disambiguation. Mutalyzer and VariantValidator both preserve it on round-trip. ferro is currently the outlier.

## Scope

- Update Display impls for `GenomeVariant`, `CdsVariant`, `TxVariant`, `RnaVariant`, `ProteinVariant`, `MtVariant`, `CircularVariant` to emit `(gene)` after the accession when set.
[...]

PR #501, opened 2026-05-31
test(mutalyzer): add improvement disposition; demote NG_ gene-symbol selector divergences (#500)

nh13:
## Summary

Adds a third non-failing conformance disposition — `improvement` — to the mutalyzer normalize harness, and uses it to classify **24** `normalized`-axis divergences where ferro emits a valid but spec-non-preferred selector on a genomic (`NG_`) reference. Demotes the `normalized` baseline **105 → 94**.

This came out of arbitrating the `NG_(GENE_v001)` selector divergences against the HGVS spec for the #326 burn-down. The arbitration found these are neither a clean `accepted_divergence` (the spec is not neutral — it prefers the transcript-accession form) nor a `known_bug` (ferro's output is valid HGVS, not wrong). The honest classification is a new middle disposition; the underlying fix (a legacy selector resolver) is tracked as a feature in #500.

## Commit 1 — `improvement` conformance disposition

The disposition triad is now:

| Disposition | Intent | ferro's output |
|---|---|---|
| `accepted_divergence` | terminal policy; won't change | valid, spec-allowed (equal footing) |
| `improvement` (new) | will converge (tracked) | valid, but spec strongly prefers the other form |
| `known_bug` | will fix (tracked) | wrong / invalid HGVS |

`improvement` behaves exactly like `known_bug` — a string mismatch on the annotated axis routes to a non-failing xfail bucket kept off `baseline-failures`; an XPASS (ferro now matches mutalyzer) fails loudly so stale annotations are cleaned up; an `Err` still fails — with one addition: it requires **both** a `tracking_issue` and a `section` (closed `SpecSection` enum), so every use is substantiated by the spec line that makes ferro's form non-preferred. Adds `SpecSection::RefSeqGeneSelector` (refseq.md:38-42,138-139) and drops the now-redundant `Hgvs` prefix from the `SpecSection` variants (satisfies `clippy::enum_variant_names`; the serde wire strings are unchanged). Five hermetic unit tests mirror the `known_bug` set.

## Commit 2 — classify the 24 NG_ selector divergences

- **11 newly demoted** off the FAIL baseline (105 → 94):
  - 8 where the input carries a gene-symbol selector ferro preserves verbatim while mutalyzer resolves it to the transcript accession (refseq.md:38-42: gene symbols should not be used as the `NG_` specification).
  - 3 bare `NG_` `c.` inputs where ferro emits no selector and mutalyzer synthesizes the transcript accession (refseq.md:138-139: the transcript used should be indicated).
- **13 migrated** from the existing `accepted_divergence` (#121) disposition — the same `NG_` pattern (gene symbol, gene-symbol-version, or numeric gene ID preserved by ferro, resolved to the transcript accession by mutalyzer). `accepted_divergence` overstated spec-neutrality for these (the spec prefers the transcript form), so `improvement` is the honest label.

The one remaining #121 `NG_` `accepted_divergence` row — `NG_012337.3(NM_003002.4):p.(Asp92Tyr)` — is a **distinct** `NM_`→`NP_` protein-selector divergence (not a gene-symbol case), so it is left as `accepted_divergence` pending separate arbitration.

Baseline regenerated from a manifest run (set-diff 0). The `NM_(GENE)` gene-symbol policy from #121 is unaffected — these are `NG_` RefSeqGene cases, which #121 never examined.

## Scope / follow-up

Mutalyzer harness only (the only corpus with these cases). The biocommons mirror of the `improvement` disposition and the #478 design-doc fold are tracked in #503. The one protein `NM_`→`NP_` selector row left as `accepted_divergence` is a distinct divergence, tracked for arbitration in #502.

## Test plan
- `cargo clippy --features dev --all-targets -- -D warnings` — 0 warnings
- `cargo nextest run --features dev` — 6522 passed, 51 skipped
- `axis_normalized` (manifest-backed): 24 → `improvement`, 1 → `divergence_accepted`, FAIL 105 → 94, set-diff 0 vs a fresh run

Refs #326, #500, #478, #121.


nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #498, opened 2026-05-30
feat(project): direct c.→p. protein path + protein/coding_protein prediction divergences


nh13:
Child of #326. Tracks the **protein_description** and **coding_protein_descriptions** axes beyond PR #495 (which spec-corrected the 7 wrapper-only cases). After #495, `protein_description` is at 7 pass / **64 FAIL** and `coding_protein_descriptions` at **51 FAIL**. The 64 split into two distinct problems:

## 1. Architecture: no direct c.→p. path (~54 `protein` + the bulk of `coding_protein`)

`VariantProjector::project_variant` → `project_single_inner` computes the protein consequence by routing a transcript-coordinate input **c. → g. → back-to-CDS → translate** (`projector.rs`, the `project_to_genomic` call). Consequences:
- Bare-`NM_` inputs (no `genomic_context` parent) **error** — they cannot reach the g. intermediate (`projector.rs` "input variant has no parent reference").
- It sits on top of the #480 genomic-coordinate bug for NG-parent inputs.

The protein consequence of a c. variant only needs the **transcript CDS + the edit** (the translation machinery already exists in `src/project/protein/`). A **direct c.→p. path** that bypasses the g. roundtrip is the principled fix; it unblocks the bare-`NM_` cases and decouples protein prediction from #480. `coding_protein_descriptions` (`project_variant_all`, many `got []`) shares this root cause.

## 2. Genuine prediction divergences (~10 `protein`)

Cases where ferro emits a *different* protein consequence than mutalyzer — these need per-case arbitration against the reference CDS (some are likely ferro bugs, some mutalyzer's). The clearest suspected ferro bug:

- `NG_009299.1(NM_017668.3):c.41A>C` → ferro `p.(Glu14Ala)`, mutalyzer `p.(Gln14ProfsTer68)`. A single-nucleotide substitution **cannot** cause a frameshift, so mutalyzer's `fsTer` output looks wrong — but ferro and mutalyzer also disagree on the reference AA at codon 14 (Glu vs Gln), pointing at a CDS-frame / transcript-version discrepancy worth tracing.

## Acceptance
- Bare-`NM_` c. and n. inputs produce a `p.` prediction (no `genomic_context` required).
- The ~10 prediction divergences are each arbitrated: ferro-bug → fixed; mutalyzer-non-spec → `cases.json` corrected with `spec_citation`; gene-symbol-selector → `accepted_divergence` (#121).
[...]

PR #510, opened 2026-06-03
fix(project): never emit descending positions in in-frame protein delins (#498)

nh13:
## Summary

Fixes a malformed-output bug in in-frame protein indel prediction: a pure insertion mis-routed to `build_inframe_delins` produced a **descending** protein range (start > end), which is invalid HGVS.

Example — `NG_007485.1(NM_000077.4):c.161_162insATC` previously rendered as `p.(Gly55_Met54delinsIle)` (positions **55_54**, crossed). It now renders as the spec-correct ascending insertion `p.(Met54_Gly55insIle)`.

## Root cause

`build_inframe_delins` derives the start from a forward `first_diff_position` scan and the end from a backward `find_last_diff` scan. For a pure insertion (alt = ref with residues inserted between two *unchanged* residues), these cross over — `last_diff_ref < first_diff` — and the result was emitted with no ordering guard, yielding `start > end`.

## Fix

When `last_diff_ref < first_diff` (the ref differing region is empty → no residues replaced), render an **insertion** between the two flanking ref residues with ascending positions. Also clamp the delins end to never fall below the start as a defensive backstop for the N-terminal edge. Genuine delins (where ref residues *are* replaced, `last_diff_ref >= first_diff`) is unchanged — the existing behavior and all 18 indel tests are preserved.

## Scope — output validity, not corpus parity

This is an **output-validity** fix: ferro no longer emits malformed (descending) protein HGVS. It demotes **0** corpus rows, because the residue-level prediction for these mid-codon insertions still differs from mutalyzer (ferro `Met54_Gly55insIle` vs mutalyzer `Met54delinsIleSer` — genuinely different residues). That remaining difference is a separate alt-computation / 3'-placement divergence that needs reference-sequence arbitration (which of the two is spec-correct), tracked under #498.

## Test plan
- New hermetic regression test `inframe_delins_renders_pure_insertion_as_ascending_ins`
- `cargo clippy --features dev --all-targets -- -D warnings` — 0 warnings
- `cargo nextest run --features dev` — 6526 passed, 51 skipped
- Manifest protein axes: no regressions; all emitted protein ranges are now ascending

Refs #498, #326.

### DIFF
--- a/src/project/protein/indel.rs
+++ b/src/project/protein/indel.rs
@@ -433,13 +433,54 @@ fn build_inframe_delins(
         return build_identity_variant(ref_protein, protein_accession, transcript);
     }
 
+    // Pure-insertion guard (#498): when the last differing ref index falls
+    // *before* `first_diff`, the ref differing region is empty — no ref
+    // residues are replaced, so this is an insertion, not a delins. Rendering
+    // it as a delins would cross the positions (`start > end`) into malformed
+    // HGVS. Emit an insertion between the two flanking ref residues with
+    // ascending positions. (Requires an interior insertion point with a
+    // residue on each side; the N-terminal edge `first_diff == 0` falls
+    // through to the clamped delins path below, which stays ascending.)
+    if last_diff_ref < first_diff && first_diff >= 1 && first_diff < ref_len {
+        let n_inserted = alt_len.saturating_sub(ref_len);
+        let left_idx = first_diff - 1;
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai full review

nh13:
@coderabbitai review

nh13:
@coderabbitai full review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #497, opened 2026-05-30
fix(normalize): avoid subtract-overflow panic in minus-strand intronic shuffle

nh13:
Closes part of #488 (the intronic minus-strand panics).

## Bug
`flip_intronic_for_strand` flips a fetched genomic-strand window into transcript view by mapping each coordinate `x` to `seq_len - x + 1`. For **minus-strand intronic** inputs whose enclosing intron is larger than the variant-sized fetched window, `boundaries.right` (the far intron edge) exceeds `seq_len`, so the subtraction underflowed and **panicked** with `attempt to subtract with overflow`. The introns are large relative to the window — e.g. `NG_009497.1(NM_206933.2):c.8682-19dupT` is a ~10.6 kb intron against a 200 bp window (`boundaries 83..10669`).

## Fix
Return a `ConversionError` when any position/boundary lies outside the fetched window, instead of panicking.

I deliberately did **not** silently clamp to the window edge: that yields an off-by-one 3′ shift (`c.378-17delT` → `c.378-16del`), and a wrong answer is worse than a detected error. Inputs whose boundaries already lie within the window — every currently-passing case — are untouched.

## Verification
- New direct unit tests for the out-of-window (error) and in-window (flip) paths.
- `cargo nextest run --features dev`: **6511 passed**, 0 failed; clippy clean.
- Manifest `axis_normalized`: the four intronic cases (`c.378-17delT`, `c.8682-19dupT`, `c.677-21_704+62del`, `c.677-20_704+62del`) now report a clean `ConversionError` instead of `panic:`; pass count unchanged (no regression, no silent wrong answers).

## Scope / follow-ups (tracked in #488)
- **Demotion** of these rows needs the genomic window sized to the enclosing intron so the shift can be computed correctly — a deeper follow-up, not this PR.
- `g.pterdel` still panics at a **separate** site (a `pos - 1` underflow in `coords`); distinct root cause, follow-up.

⚠️ Committed `--no-gpg-sign` (1Password biometric unavailable); re-sign on rebase before merge.

Refs #488.

Issue #488, opened 2026-05-29
fix(normalize): subtract-with-overflow panics on intronic/pter inputs


nh13:
Child of #326. Spec arbitration verdict: **(a) ferro bug — crashes** (subset of `normalized`/`genomic`). Originally noted under the now-closed #87.

## Problem

`Normalizer::normalize` panics with `attempt to subtract with overflow` on several intronic / `pter` inputs (the corpus runner only survives because it wraps each case in `catch_unwind`):

- `NG_012337.1:g.pterdel`
- `NG_008835.1(NM_022153.2):c.677-21_704+62del` and `c.677-20_704+62del`
- `NG_009497.1(NM_206933.2):c.8682-19dupT`
- `NG_007107.2(NM_004992.3):c.378-17delT`

A panic in library code is never acceptable regardless of the expected output. The overflow was previously called out at `normalize/shuffle.rs` (#87).

## Ask

Fix the underlying arithmetic so these inputs normalize correctly (or return a `Result::Err`, never panic). Add regression tests. Where ferro then produces the spec-correct output, demote the corresponding `baseline-failures/{normalized,genomic}.txt` rows.

Refs #326.

nh13:
**PR #497** addresses the largest slice: the **minus-strand intronic** subtract-overflow in `flip_intronic_for_strand` (the far intron edge `boundaries.right` exceeding the variant-sized fetched window — introns up to ~10.6 kb vs a 200 bp window). It now returns a clean `ConversionError` instead of panicking — verified on `c.378-17delT`, `c.8682-19dupT`, `c.677-21_704+62del`, `c.677-20_704+62del`, with no regression and CI-runnable unit tests.

Two items remain under this issue:
1. **Demotion** of those rows (correct shifted output, e.g. `c.677-21_704+62del` → `c.677-18_704+65del`) needs the fetched genomic window sized to the **enclosing intron** so the 3′ shift can be computed within real bases. PR #497 deliberately errors rather than clamp-to-window (clamping produced an off-by-one shift — a silent wrong answer).
2. **`g.pterdel`** still panics at a **separate** site — a `pos - 1` underflow in `coords` (`pter` resolving to position 0). Distinct root cause.

nh13:
**Update on the correct-output / demotion follow-up** (the part beyond #497's no-panic guard):

[...]

PR #518, opened 2026-06-04
fix(normalize): guard special/offset genome positions to avoid pterdel overflow panic (#488)

nh13:
## Summary

Eliminates the last remaining `attempt to subtract with overflow` panic from #488: `g.pterdel` and friends on genomic references (e.g. `NG_012337.1:g.pterdel`).

`normalize_genome` read `pos.base` directly and fed it into the window coordinate math. For telomere/centromere markers (`pter`/`qter`/`cen`), `base` is a `0` sentinel, which flowed into `coords::hgvs_pos_to_index(0)` (a `pos - 1` conversion) and underflowed — panicking. A panic in library code is never acceptable.

The fix mirrors the existing `normalize_mt` guard: when either bound is a special position or carries an offset, fall back to minimal-notation canonicalization instead of base-only window normalization (which would also silently drop the marker when rebuilding via `GenomePos::new`).

This completes the **no-panic** requirement of #488 alongside the minus-strand intronic fix already merged in #497.

## Root cause

The panic surfaces at `coords::hgvs_pos_to_index` (`pos - 1`), but the root cause is upstream in `normalize_genome`, which — unlike its `normalize_mt` sibling — did not guard `base == 0` special positions before the window/coordinate arithmetic. Fixed at the source; the symptom site keeps its documented "1-based position" contract (clamping there would silently mask genuinely invalid positions, the same rationale #497 used to reject window-edge clamping).

## Scope / non-goals

This is the no-panic fix only. Producing the spec-correct shifted output (`g.pterdel` → `g.3del`, by resolving `pter`→1 / `qter`→contig-end) — and demoting the corresponding `baseline-failures/{normalized,genomic}.txt` rows — is a separate telomere-coordinate-resolution follow-up, just as the intronic window-sizing was deferred from #497. No baseline rows are demoted here.

## Testing

- New hermetic regression test `genome_special_position_does_not_panic` covers `pter`/`qter`/`cen` del + dup under a window-sized contig (so the path reaches the previously-panicking arithmetic rather than bailing early on a missing sequence).
- All 208 `normalize` lib tests pass; `mock-pin` regression unchanged (243 cases match).
- `cargo clippy --features dev -- -D warnings` clean; `cargo fmt` clean.
- Full suite: **6535 passed, 51 skipped**.

Refs #488. Follows #497.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -1279,8 +1279,8 @@ impl<P: ReferenceProvider> Normalizer<P> {
         if !needs_normalization(edit) {
             return Ok((HV::Genome(variant.clone()), vec![]));
         }
-        let start = match variant.loc_edit.location.start.inner() {
-            Some(pos) => pos.base,
+        let start_pos = match variant.loc_edit.location.start.inner() {
+            Some(pos) => pos,
             None => {
                 return Ok((
                     HV::Genome(self.canonicalize_genome_variant(variant)),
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #505, opened 2026-05-31
fix(reference): version-exact transcript sequence availability for protein prediction


nh13:
~5 corpus rows on the protein axes error with `project: Reference not found: NM_003002.2` (and similar). The cdot mapper is version-tolerant (it resolves `NM_003002.2`→`.4`), but the **sequence** provider (`MultiFastaProvider`) only carries the transcript versions present in the reference FASTAs. When the exact input version's sequence is absent, the direct c.→p. path (and the genome path's per-codon read) cannot read `transcript.sequence` and fails with `ReferenceNotFound`.

This is a reference-data / provider concern, independent of the projection path. Options to evaluate:
- carry/serve a version-tolerant sequence lookup (fall back to the nearest available version when the CDS is unchanged across versions), with an explicit warning; or
- document that exact-version sequences must be ingested and treat absence as an honest error (current behavior).

Note this interacts with the exact-version transcript resolution work (#490). Arbitrate whether a nearest-version sequence fallback is spec-safe before implementing.

Refs #498, #490, #326.

nh13:
## A second, silent failure mode: version *present* but not version-*exact*

This issue currently frames version-exactness as an availability problem (version absent → honest `ReferenceNotFound`). Working the #498 protein burn-down surfaced a worse, silent variant: for some accessions the manifest **does** carry a sequence under the pinned version, but that sequence is **not version-exact** with the canonical RefSeq record — so ferro silently produces a **wrong** protein prediction (no error). These show up in the corpus as `protein_description` FAILs where ferro is genuinely wrong (mutalyzer is right), the opposite of an accepted divergence.

Two confirmed cases (verified against the live NCBI RefSeq records):

### `NM_012459.2` (TIMM8B) — wrong CDS/protein-version pairing

- Canonical GenBank `NM_012459.2`: `CDS 31..327 → /protein_id="NP_036591.2"` (full-length, `MRKHSCRKVASLRRTMAELGEA…`, 98 aa).
- ferro's manifest: `cds_start=34`, `cds_end=285`, `protein_id=NP_036591.3` (`MAELGEA…`, 83 aa — the **internal-Met-start** isoform, 15 residues shorter).

[...]

PR #519, opened 2026-06-04
fix(reference): decline protein prediction for non-version-exact transcripts (#505)

nh13:
## Summary

Closes part of #505 (the cross-version-substitution failure mode).

Protein prediction reads a transcript's CDS bases directly and translates them. The lenient reference lookup silently substitutes a *different version* of a transcript when the exact requested version is absent (version-strip fallback), so prediction could translate wrong-version bases and emit a wrong protein attributed to the requested version.

This adds a `ReferenceProvider::has_transcript_version_exact` predicate (default `true`) and gates the shared protein chokepoint `predict_protein_consequence` on it: when the requested transcript version is not exactly available, protein prediction declines (`Ok(None)`) rather than translating substituted bases. The coding and genomic axes are unaffected (the gate is protein-only).

## Design

`MultiFastaProvider` implements the predicate as "the FASTA ships the exact version, **or** no FASTA sibling shadows it **and** cdot carries the exact version". This keeps exact-version cdot-genome synthesis (#331) eligible for protein prediction while rejecting cross-version substitution. The `resolve_name(id).is_none()` guard exactly matches the read path, which serves a FASTA version-strip sibling before it ever reaches cdot synthesis — so the gate's verdict agrees with the bases actually served.

The predicate is forwarded through the delegating providers (`Box<dyn ReferenceProvider>`, `PyProvider`, `CachedFastaProvider`) so the gate is enforced on the boxed-CLI and Python paths, not only on direct providers.

## Scope

This PR closes the **cross-version-substitution** silent-wrong class. It does **not** address the issue's "mode 2b": a sequence ingested under the *correct* version label but whose bases/CDS are non-canonical (e.g. `NM_012459.2` paired with the `NP_036591.3` short-isoform CDS; `NM_000193.2` served as 4650 nt vs canonical 1576 nt). That requires reference-data validation (translate served CDS vs the pinned `NP_`; check transcript length vs the canonical record) plus version-exact ingestion, and is tracked separately under #505.

Known residual (documented in the predicate's doc comment): the cdot arm consults the **primary-build** transcript map. The bare-transcript protein path routes through the primary-build view this models; an NG/NC-parented variant can probe alt-build maps with their own version-strip fallback, whose exactness this predicate does not yet cover. Follow-up.

## Testing

- TDD throughout; 6 new tests covering: gate declines protein for a non-version-exact transcript (projector end-to-end); FASTA-exact → true; sibling-only → false; exact-version cdot synthesis → true (the #331 allowance); FASTA-sibling shadows cdot-exact → false; `Box<dyn>` forwards the predicate.
- Full suite: 6540 passed, 51 skipped. `clippy --all-targets -D warnings` clean. `--features python` compiles.

## Suggested reading order

1. `src/reference/provider.rs` — the new trait method (default `true`) + `Box<dyn>` forward.
2. `src/reference/multi_fasta.rs` — the real predicate + its three unit tests.
3. `src/project/projector.rs` — the one-line gate in `predict_protein_consequence` + the projector test.
4. `src/reference/mock.rs`, `src/python.rs`, `src/reference/fasta.rs` — test knob and delegating-provider forwards.

### DIFF
--- a/src/project/projector.rs
+++ b/src/project/projector.rs
@@ -1026,6 +1026,20 @@ impl<P: ReferenceProvider + Clone> VariantProjector<P> {
         let Some(prot_acc) = prot_acc else {
             return Ok(None);
         };
+        // Issue #505: protein prediction below reads this transcript's CDS
+        // bases directly and translates them. If the reference only carries a
+        // *different* version of the transcript (a silent version-strip
+        // substitution), those bases may pair with a different CDS/exon
+        // structure, so the prediction would be wrong yet attributed to the
+        // requested version. Decline to predict rather than emit a wrong
+        // protein — this gates *all* downstream prediction, including the
+        // initiation-codon short-circuit below, whose CDS positions are
+        // likewise mapped against the requested-version transcript. The
+        // predicate still permits exact-version cdot-genome synthesis — it
+        // rejects only cross-version substitution.
+        if !self.provider.has_transcript_version_exact(transcript_id) {
+            return Ok(None);
+        }
         // An initiation-codon-affecting edit short-circuits to `p.(Met1?)`
         // here, before edit-type dispatch, so boundary-spanning edits whose
         // 5' end is in the 5'UTR (which the per-edit paths reject via their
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai review

Issue #520, opened 2026-06-04
feat(reference): canonical-record validation + version-exact correction (#505 follow-up)


nh13:
## Background

Issue #505 covers version-exactness for protein prediction. PR #519 closes the **cross-version-substitution** failure mode (the protein path now declines to predict when the exact transcript version is not available, rather than translating silently-substituted bases). This issue tracks the remaining, harder half: detecting and correcting transcript records whose served bases / CDS coordinates / paired protein accession do **not** match the authoritative RefSeq record for the *exact* pinned accession version — even when the version label is correct ("mode 2b").

Two confirmed corpus oracles (verified against live NCBI in #505):

- **`NM_012459.2` (TIMM8B)** — the manifest pairs the `.2` transcript with `NP_036591.3` and the short-isoform CDS (83 aa) instead of the canonical `NP_036591.2` / 98-aa CDS. Internally consistent, but mis-paired → wrong protein for every CDS-frame prediction on this transcript.
- **`NM_000193.2` (SHH)** — served as 4650 nt (CDS `342..1730`, ~2.9 kb 3′UTR) vs the canonical 1576 nt (CDS `152..1540`, 36 nt 3′UTR) → the stop-loss row `c.1388G>C` reads a spurious downstream in-frame stop.

## Key finding (what catches what)

Tested three validation strategies against the two oracles:

| Strategy | Needs | `NM_012459.2` | `NM_000193.2` |
|---|---|---|---|
| **A. Internal CDS sanity** (ATG/stop/internal-stop/len%3) | nothing (offline) | ❌ (valid 83-aa isoform) | ❌ (CDS translates cleanly; defect is 3′UTR length) |
| **B. Translate served CDS, compare length to pinned `NP_`** | NP_ FASTA | ❌ (manifest is internally consistent: `.3` CDS ↔ `NP_…​.3`) | partial (length divergence) |
| **C. Compare cdot's (CDS, `protein_id`, length) to the authoritative GenBank record for the exact version** | GenBank `.gbff`/efetch | ✅ | ✅ |

Two consequences:
1. The mis-pairing class (`NM_012459.2`) is only caught by an **authoritative external** reference (strategy C), not by the original "compare to pinned NP_ length" idea.
[...]

PR #521, opened 2026-06-04
feat(reference): offline structural validation of transcript records (#520)

nh13:
## Summary

Phase 0 of #520 (canonical-record validation, follows #505 / PR #519). Adds an **offline, self-contained** structural validator over a single `Transcript` — no network, no authoritative external record.

`validate_transcript_record(&Transcript) -> Vec<TranscriptAnomaly>` and `validate_transcripts(provider, ids)` flag, with an explicit `AnomalyConfidence`:

- **Corruption** (high confidence): served sequence **shorter** than the exon-alignment extent (missing bases), CDS out of range, CDS not a triplet, or a non-not-found load error.
- **Advisory** (heuristic, known biological false positives): missing start codon, missing stop codon, premature in-frame stop. These are tagged Advisory because selenoproteins recode an in-frame `TGA` as selenocysteine, some transcripts use alternative/absent start codons, readthrough exists, and `partial` RefSeq records legitimately lack a start/stop. Ambiguous (`N`/IUPAC) codons are skipped, not flagged.

## Deliberately out of scope (later phases)

- A served sequence **longer** than the alignment is **not** flagged — a poly-A tail / unaligned 3′ bases routinely make the transcript FASTA longer than the genome-derived alignment. Detecting an implausibly-long sequence precisely (e.g. `NM_000193.2`, served ~3× too long) needs the authoritative transcript length → the authoritative-comparison phase.
- A sequence internally consistent but mis-paired to the wrong protein version (`NM_012459.2`) → also a later phase (needs the authoritative RefSeq record for the exact version).

So Phase 0 does not by itself flip the #505 corpus oracles; it catches the truncation / frame / load-error classes and lays the typed-anomaly foundation the later phases extend.

## Review

Reviewed by an Opus pass **and** a CodeRabbit-style pass before opening. Fixes folded in:
- **Length check made asymmetric** (shorter-only) to eliminate poly-A/unaligned-3′ false positives — the original `max(exon.end) != seq_len` would have flagged every transcript with a poly-A tail.
- **Codon checks demoted to `Advisory`** with documented biological false positives (selenocysteine etc.); ambiguous codons skipped.
- **CDS slicing operates on bytes** so a malformed non-ASCII record cannot panic an `&str` slice.
- **`validate_transcripts` surfaces non-`ReferenceNotFound` load errors** as a `LoadError` anomaly instead of silently dropping them.
- **`#[non_exhaustive]`** on `AnomalyKind` / `AnomalyConfidence` / `TranscriptAnomaly` so later phases can add variants/fields without a breaking change.

## Testing

15 unit tests (TDD): clean/lowercase/ambiguous transcripts, shorter-vs-longer length, triplet/start/stop/internal-stop with confidence assertions, CDS-out-of-range, non-ASCII no-panic, `validate_transcripts` happy-path + skip-missing + surface-load-error. Full suite green, `clippy --all-targets -D warnings` clean.

Not wired into the `ferro check` CLI in this PR — the library core is the unit; CLI scan wiring lands where a prepared manifest exists.

Refs #520, #505, #519.

### DIFF
--- a/src/reference/authoritative.rs
+++ b/src/reference/authoritative.rs
@@ -1,0 +1,475 @@
+//! Authoritative transcript records and the canonical-overrides schema.
+//!
+//! The canonical-record validation effort (issue #520) needs the *authoritative*
+//! RefSeq record for an exact accession version — its CDS coordinates, the
+//! paired protein accession, and the transcript length — to detect (and later
+//! correct) transcript records whose cdot-derived metadata disagrees with the
+//! canonical record (e.g. `NM_012459.2` cdot-paired with `NP_036591.3` instead
+//! of the canonical `NP_036591.2`).
+//!
+//! This module provides:
+//!
+//! - [`AuthoritativeRecord`] — the fields we compare against, parsed from a
+//!   GenBank flat-file record ([`parse_authoritative_genbank`]). Parsing is
+//!   pure and network-free; the actual NCBI efetch that produces the GenBank
+//!   text lives in the `prepare` pipeline.
+//! - [`CanonicalOverrides`] — a serialisable accession→record map, the
+//!   "authoritative-overrides" file written by `ferro prepare` and consumed by
+//!   the later validation / correction phases.
+//!
+//! All coordinates here are **1-based inclusive** (as written in the GenBank
+//! `CDS` feature). cdot stores CDS as 0-based; callers comparing the two must
+//! convert.
+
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #522, opened 2026-06-04
feat(reference): authoritative GenBank record parsing + canonical-overrides schema (#520)

nh13:
## Summary

Phase 1 of #520 (canonical-record validation). **Stacked on #521** (Phase 0) — base is `feat/issue-505-canonical-validation`; review #521 first.

Adds the network-free core the later phases compare cdot-derived metadata against:

- **`parse_authoritative_genbank(&str) -> Option<AuthoritativeRecord>`** — extracts from a GenBank flat-file record: the `VERSION` accession, transcript length (counted from the `ORIGIN` sequence), and the first `CDS` feature's **1-based** bounds and `/protein_id`. Location parsing handles `a..b`, `join(...)` (including a join that **wraps** onto continuation lines), `complement(...)`, and `<`/`>` partial markers. Only the first CDS feature is used; only the first record is parsed (one accession per efetch call).
- **`AuthoritativeRecord`** and **`CanonicalOverrides`** — a serde accession→record map (with an on-disk `schema_version`) = the "authoritative-overrides" artifact `ferro prepare` will write and the validate/correct phases consume.

`protein_id` is the field that lets a later phase catch the `NM_012459.2` mis-pairing (`.2` transcript canonically pairs with `NP_036591.2`, not the `.3` short isoform cdot records). Coordinates are 1-based inclusive (GenBank); cdot is 0-based, so the compare phase converts.

## Review

Opus + CodeRabbit-style review before opening. Fixes folded in:
- **Wrapped `join(...)` CDS location** — the original read the location only from the CDS key line, so a join wrapping onto continuation lines parsed to `None` (CDS silently dropped). Now the location is accumulated across continuation lines until the first `/qualifier`.
- **First-CDS-wins** — a second CDS feature no longer overwrites the first's bounds while keeping the first's `protein_id` (`seen_cds` guard).
- **On-disk `schema_version`** on `CanonicalOverrides` (+ `#[non_exhaustive]`) for format longevity; legacy files without the field read as `0`.
- Documented: parses the first record only; `complement(join(...))` correctness.

## Scope / verification

This PR is the **parsing + schema core**, unit-tested (11 tests) with synthetic GenBank fixtures: VERSION/length/CDS/`protein_id` extraction, wrapped + `complement(join())` locations, partial markers, first-CDS-wins, non-coding, JSON round-trip, legacy-schema default. `clippy --all-targets -D warnings` clean.

The NCBI **efetch** that produces the GenBank text (the `ferro prepare --validate-canonical` stage) is the network edge and is **not** in this PR — it's wired and verified on a prepared host / in CI, where a real reference manifest exists.

Refs #520, #505.

### DIFF
--- a/src/reference/authoritative.rs
+++ b/src/reference/authoritative.rs
@@ -1,0 +1,475 @@
+//! Authoritative transcript records and the canonical-overrides schema.
+//!
+//! The canonical-record validation effort (issue #520) needs the *authoritative*
+//! RefSeq record for an exact accession version — its CDS coordinates, the
+//! paired protein accession, and the transcript length — to detect (and later
+//! correct) transcript records whose cdot-derived metadata disagrees with the
+//! canonical record (e.g. `NM_012459.2` cdot-paired with `NP_036591.3` instead
+//! of the canonical `NP_036591.2`).
+//!
+//! This module provides:
+//!
+//! - [`AuthoritativeRecord`] — the fields we compare against, parsed from a
+//!   GenBank flat-file record ([`parse_authoritative_genbank`]). Parsing is
+//!   pure and network-free; the actual NCBI efetch that produces the GenBank
+//!   text lives in the `prepare` pipeline.
+//! - [`CanonicalOverrides`] — a serialisable accession→record map, the
+//!   "authoritative-overrides" file written by `ferro prepare` and consumed by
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

PR #523, opened 2026-06-04
feat(reference): authoritative-comparison validation tier (#520)

nh13:
## Summary

Phase 2 of #520. **Stacked on #522** (Phase 1) — base is `feat/issue-505-canonical-phase1-authoritative`; review #521 → #522 → this in order.

Adds the **precise** validation tier: compares a served transcript against the authoritative RefSeq record (`AuthoritativeRecord` from Phase 1).

- **`validate_against_authoritative(tx, auth)`** flags, as `Corruption`-confidence anomalies, a served-vs-authoritative disagreement in: transcript **length**, **CDS bounds** (both 1-based inclusive), or paired **protein accession**. This is what catches the #505 cases the offline tier (Phase 0) cannot:
  - `NM_012459.2` — served `NP_036591.3` / CDS `34..285` vs canonical `NP_036591.2` / `31..327` → `AuthoritativeCdsMismatch` + `AuthoritativeProteinIdMismatch`.
  - `NM_000193.2` — served 4650 nt vs canonical 1576 → `AuthoritativeLengthMismatch` (+ CDS).
- **`validate_against_overrides(provider, overrides)`** runs it for every accession in a `CanonicalOverrides` map, skipping `ReferenceNotFound` and surfacing other load errors as `LoadError`.

## Design notes (from review)

- **Coordinate correctness verified** (Opus): both sides are 1-based inclusive. `Transcript.cds_end` needs no `+1` because cdot's 0-based *exclusive* end is numerically identical to a 1-based inclusive end — a code comment now records this so a future "fix" doesn't introduce an every-record false positive.
- **Presence mismatches are documented, not flagged.** CDS/protein_id are compared only when both sides carry the value: a served `None` usually means "not populated by the provider" (the cdot-synthesis path sets `protein_id: None`), so flagging a presence difference would be false-positive-prone.
- **`ReferenceNotFound` skip is deliberate**: "we hold the canonical record but the reference doesn't ship this transcript" is an inventory concern for the prepare/check layer, not corruption of a served record.
- Variant names use a consistent `Authoritative*` prefix.

## Review & testing

Opus + CodeRabbit-style review before opening; fixes folded in (variant rename, coordinate comment, presence-gap doc, skip-decision doc, added tests). 7 new tests (22 in the module total): match, CDS+protein mismatch, length mismatch, presence-only non-flag, coordinate-only length skip, overrides happy-path + load-error. `clippy --all-targets -D warnings` clean.

End-to-end (the manifest corpus actually flipping rows) is verified once Phase 3 (correction) lands and on a prepared host; this PR is the comparison logic, fully unit-tested with synthetic inputs.

Refs #520, #505.

### DIFF
--- a/src/reference/validate.rs
+++ b/src/reference/validate.rs
@@ -1,14 +1,23 @@
-//! Offline structural validation of transcript reference records.
+//! Validation of transcript reference records (issue #520).
 //!
-//! These checks are **self-contained**: they look only at a single
-//! [`Transcript`]'s own sequence, CDS coordinates, and exon alignment, with no
-//! network access and no authoritative external record. They catch a class of
-//! reference-data corruption where the served bases and the alignment/CDS
-//! metadata disagree, or the CDS does not translate as a sane open reading
-//! frame.
+//! Two tiers, both reporting [`TranscriptAnomaly`] tagged with an
+//! [`AnomalyConfidence`]:
 //!
-//! Scope (issue #520, phase 0): this is the cheap, offline tier. Anomalies are
-//! tagged with an [`AnomalyConfidence`]:
+//! 1. **Offline structural** ([`validate_transcript_record`]) — self-contained
+//!    checks over a single [`Transcript`]'s own sequence, CDS coordinates, and
+//!    exon alignment; no network, no external record.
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

PR #524, opened 2026-06-04
feat(reference): version-exact correction of transcript metadata (#520)

nh13:
## Summary

Phase 3 of #520. **Stacked on #523** (Phase 2) — base is `feat/issue-505-canonical-phase2-compare`; review #521 → #522 → #523 → this.

Adds **`apply_canonical_overrides(tx, overrides)`** — reconciles a served transcript's metadata with the authoritative RefSeq record (Phase 1/2 types):

- **Metadata-mismatch class** (`NM_012459.2`: canonical `.2` bases but cdot CDS/`protein_id` point at the `.3` short isoform) → overrides `cds_start`/`cds_end` and `protein_id` from the authoritative record so prediction translates the canonical frame. **All-or-nothing**: CDS and `protein_id` are corrected together or not at all (a corrected frame with a stale protein label is actively misleading).
- **Wrong-sequence class** (`NM_000193.2`: 4650 nt served vs canonical 1576) → the served length (sequence bytes, or exon-sum for a coordinate-only record) disagrees with authoritative, so the coordinates don't apply → no override; `Correction::SequenceReingestionRequired` reported instead. Fixing that needs the authoritative *bases*, which the overrides file does not carry.

## Documented limitations (from review)

- **Same-length-but-edited** sequences pass the length gate undetected — length is a base-identity proxy (no checksum in the overrides).
- The projection path reads the **cdot mapper's** parallel `protein`/CDS and prefers them, so correcting the served `Transcript` alone is a half-fix — full effect requires the wiring to also reconcile the cdot record (or route projection through the corrected transcript).
- **Ordering contract:** correction must run on a freshly loaded transcript *before* it is cached/translated — otherwise the projector's `ref_protein_cache` (the cached reference translation) goes stale. The `Arc<Transcript>` cache structurally prevents accidental mutation of an aliased transcript; the hazard is the companion translation cache.

These are why this PR is the **correction logic**, not the provider wiring. Wiring `apply_canonical_overrides` into `MultiFastaProvider`/the projector cache-population path (atomically, before translation) and reconciling the cdot record is the integration step, verified on a prepared host where the manifest corpus rows actually flip to PASS.

## Review & testing

Opus + CodeRabbit-style review; fixes folded in (all-or-nothing coding triple, exon-sum length gate for coordinate-only records, ordering/cdot-parallel/edited-length doc caveats, 3 added tests). 6 correction tests (28 in the module): canonical-length CDS+protein correction, wrong-length reingestion (sequence + coordinate-only), no-op cases, coordinate-only correction via exon-sum, partial-authoritative skip. `clippy --all-targets -D warnings` clean.

Refs #520, #505.

### DIFF
--- a/src/reference/validate.rs
+++ b/src/reference/validate.rs
@@ -180,9 +180,21 @@ pub fn validate_transcript_record(tx: &Transcript) -> Vec<TranscriptAnomaly> {
     }
 
     // --- CDS open-reading-frame sanity --------------------------------------
-    let (Some(cds_start), Some(cds_end)) = (tx.cds_start, tx.cds_end) else {
-        // Non-coding (no CDS): only the length check applies.
-        return anomalies;
+    // Only a fully-absent CDS `(None, None)` is the legitimate non-coding case.
+    // A half-populated pair is structurally incomplete and must be flagged as
+    // corruption rather than silently passing as non-coding.
+    let (cds_start, cds_end) = match (tx.cds_start, tx.cds_end) {
+        (Some(start), Some(end)) => (start, end),
+        (None, None) => return anomalies,
+        (start, end) => {
+            anomalies.push(TranscriptAnomaly::new(
+                &tx.id,
+                AnomalyKind::CdsOutOfRange,
+                AnomalyConfidence::Corruption,
+                format!("CDS bounds are incomplete: start={start:?}, end={end:?}"),
+            ));
+            return anomalies;
+        }
     };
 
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #525, opened 2026-06-04
feat(reference): translated-CDS vs canonical-protein check (#520)

nh13:
## Summary

Phase 4 of #520 — the optional **base-level (strategy-B) depth** tier. **Stacked on #524** (Phase 3); review #521 → #522 → #523 → #524 → this.

Adds **`validate_translation_against_protein(tx, canonical_protein)`** — translates the served CDS (reusing the non-gated `CodonTable`) and compares to the canonical protein, flagging `TranslationMismatch` on a residue divergence. This catches base-level non-canonical sequences the coordinate / length / `protein_id` comparisons miss — e.g. a **same-length CDS with point edits** (exactly the case Phase 3 flagged as undetectable by length alone).

## Biological adjustments (from review)

The standard codon table can't model every recoding, so canonical RefSeq would otherwise false-positive. Mitigations:
- **Initiator Met forced** — translation initiation installs Met regardless of the start codon; canonical `NP_` always begins with `M`. Removes the alternative-start (CTG/GTG/…) position-1 false positive.
- **Selenocysteine readthrough suppressed** — the standard table stops at an in-frame `TGA`, so a selenoprotein translates to a prefix ending exactly where the canonical protein has a `U` (Sec); that specific shape returns no anomaly (selenoproteins are a clinically real gene class: GPX, TXNRD, DIO, SELENO*).
- **`Advisory` confidence, not `Corruption`** — a residual mismatch is a review signal, matching the offline codon checks' precedent (Phase 0).

Lowercase/soft-masked bases need no handling (`Codon::parse_bytes` → `Base::from_char` upper-cases). The 1-based CDS slice ends with the stop codon (cdot/RefSeq convention), which the translator drops before emitting — matching `NP_` (no stop).

## Scope

Supplying `canonical_protein` requires **protein-FASTA (`*.protein.faa.gz`) ingestion** and a `MultiFastaProvider::get_protein_sequence` implementation — the data edge, **not** in this PR. This PR is the translate-and-compare core, unit-tested with synthetic transcript + protein strings.

## Review & testing

Opus + CodeRabbit-style review; fixes folded in (Advisory demotion, selenocysteine suppression, initiator-Met). 7 tests (35 in the module): match, trailing-stop tolerance, residue-divergence Advisory mismatch, selenocysteine suppression, alternative-start initiator-Met, ambiguous-codon skip, non-coding skip. `clippy --all-targets -D warnings` clean.

[...]

PR #528, opened 2026-06-04
feat(reference): apply canonical overrides in MultiFastaProvider + reconcile cdot (#520)

nh13:
## Summary

**Edge 2** of the canonical-validation integration (#520) — wires the correction into `MultiFastaProvider` so served transcripts **and** the projection path use authoritative CDS/protein. **Stacked on #527** (Edge 1); review the chain (phases #521→#525, then #527, then this).

- `MultiFastaProvider` carries `canonical_overrides: CanonicalOverrides`, loaded in `from_manifest` from the manifest's `canonical_overrides` file.
- **Single chokepoint:** `get_transcript_on_build` is now a thin wrapper around `get_transcript_on_build_inner` that applies `validate::apply_canonical_overrides` to the built transcript. So **every** public path — lenient `get_transcript`, `get_transcript_strict`, and the variant-aware build-probing `get_transcript_for_variant` — gets the corrected metadata before the transcript is cached or translated (satisfying the phase-3 cache-ordering contract).
- **cdot reconciliation:** `reconcile_cdot_with_overrides` patches the cdot mapper's parallel `cds_start`/`cds_end`/`protein` for overridden accessions — the half the phase-3 review flagged as essential, since the projection path reads cdot's fields preferentially over the served `Transcript`. Authoritative 1-based bounds convert to cdot's 0-based start / 0-based-exclusive end (`cds_start - 1`, `cds_end` unchanged); only fully-coding records whose served FASTA length matches are reconciled.

## Effect

Together with Edge 1 (which produces the overrides file) and the phase stack, this is what makes `NM_012459.2` predict `NP_036591.2` / the canonical frame: both the served `Transcript` and the cdot record the projector reads are corrected.

## Testing

3 new tests: apply-on-read corrects a served transcript; cdot reconciliation converts coordinates (1-based `1..9` → cdot `0..9`, protein swapped); length-mismatch leaves cdot untouched. `ferro`/`python`/`benchmark` compile; full module suite green (90 in the touched modules); `clippy --all-targets -D warnings` clean.

End-to-end (a prepared manifest with a real `canonical_overrides.json` flipping the corpus rows) is the on-host verification this enables.

Refs #520, #505.

### DIFF
--- a/src/bin/benchmark.rs
+++ b/src/bin/benchmark.rs
@@ -1502,6 +1502,7 @@ fn run_prepare_tool(
                 skip_existing: !force,
                 clinvar_file: None,
                 patterns_file: None,
+                validate_canonical_accessions: None,
                 dry_run: false,
             };
             let manifest = prepare_references(&config)?;
--- a/src/bin/ferro.rs
+++ b/src/bin/ferro.rs
@@ -479,6 +479,11 @@ enum Commands {
         #[arg(long)]
         clinvar: Option<PathBuf>,
 
+        /// Newline-delimited accession list to validate against authoritative
+        /// GenBank records, writing canonical_overrides.json (issue #520)
+        #[arg(long)]
+        validate_canonical: Option<PathBuf>,
+
         /// Dry run - show what would be downloaded without downloading
         #[arg(long)]
         dry_run: bool,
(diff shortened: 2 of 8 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #530, opened 2026-06-04
feat(reference): ingest protein FASTAs + get_protein_sequence + translation check (#520)

nh13:
## Summary

**Edge 3** of the canonical-validation effort (#520) — the **strategy-B depth tier**: translate the served CDS and compare it to the canonical protein, catching base-level non-canonical sequences (e.g. a same-length CDS with point edits) that the coordinate / length / `protein_id` checks miss. **Stacked on #528** (Edge 2).

- **`ferro prepare`** downloads the companion RefSeq `*.protein.faa.gz` files (same `mRNA_Prot/human.N` base as the RNA files), decompresses + indexes them, and records them in the manifest's new `protein_fastas` field.
- **`MultiFastaProvider`** gains a `protein_index` (loaded from `protein_fastas` in `from_manifest`) and implements **`get_protein_sequence`** / `has_protein_data` over it, reusing the existing FASTA index reader.
- **`MultiFastaProvider::validate_translations(pairs)`** makes Phase 4's `validate_translation_against_protein` *live*: fetches the served transcript + the canonical protein and reports `TranslationMismatch` (Advisory) on a divergence. `ferro check` displays the new manifest entry.

## Testing

2 new tests: `get_protein_sequence` reads from the protein index (real file-backed read); end-to-end translation mismatch (an override sets the CDS so the transcript is coding, the canonical protein differs → `TranslationMismatch`). `ferro`/`python`/`benchmark` compile; module suite green (56); `clippy --all-targets -D warnings` clean.

The `*.protein.faa.gz` **download** is the on-host network piece (gated inside `download_transcripts`); the ingestion, `get_protein_sequence`, and the translation wiring are unit-tested here.

Refs #520, #505.

PR #531, opened 2026-06-05
feat(reference): carry canonical sequence in overrides + correct wrong-sequence class (#520)

nh13:
## Summary

Item #2 of the canonical-validation follow-ups (#520): carry the canonical **sequence** in the overrides so we can actually *fix* the wrong-sequence class, not just flag it. **Stacked on #530** (Edge 3).

- `AuthoritativeRecord` gains an optional `sequence` (the canonical `ORIGIN` bases); `parse_authoritative_genbank` now captures it and derives `tx_length` from it. `#[serde(skip_serializing_if = "Option::is_none")]` keeps sequence-less records compact.
- **`apply_canonical_overrides` now corrects the bases**: when the override carries `sequence` and the served bases differ, it replaces them (`Correction::SequenceCorrected`) and then applies the CDS/protein triple. This **fixes the wrong-sequence class** (`NM_000193.2`, served 4650 nt vs canonical 1576) and **catches same-length-but-edited** sequences (byte comparison). With no `sequence` carried, the prior length-gate / `SequenceReingestionRequired` behavior is unchanged.
- `reconcile_cdot_with_overrides` fires when the override carries `sequence` (the served bases are replaced on read), not only when the FASTA length already matches.

## Caveats (documented)

- Replacing the bases does **not** rewrite the (stale) exon structure — protein prediction reads the CDS bases so it's correct; genomic projection on a sequence-corrected transcript may be inconsistent.
- Carrying full sequences enlarges the overrides file (the field is optional; the acquisition can choose whether to populate it).

## Testing

Parser captures the ORIGIN sequence; wrong-length and same-length-edited sequences are replaced + CDS/protein applied; reingestion still reported when no `sequence` is carried. 96 module tests green; `ferro`/`python`/`benchmark` compile; clippy clean.

Refs #520, #505.

### DIFF
--- a/src/reference/authoritative.rs
+++ b/src/reference/authoritative.rs
@@ -43,6 +43,13 @@ pub struct AuthoritativeRecord {
     /// `NP_036591.2`). `None` for a non-coding record or one without the
     /// qualifier.
     pub protein_id: Option<String>,
+    /// The canonical transcript bases (upper-cased `ORIGIN` sequence). When
+    /// present, the correction path can *replace* a served non-canonical
+    /// sequence (the wrong-sequence class, e.g. `NM_000193.2`) and detect a
+    /// same-length-but-edited sequence — not just flag a length mismatch.
+    /// Optional because carrying full sequences enlarges the overrides file.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub sequence: Option<String>,
 }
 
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #532, opened 2026-06-05
feat(reference): reconcile cdot for synthesis-only transcripts via exon-sum length (#520)

nh13:
## Summary

Item #3 (final) of the canonical-validation follow-ups (#520): close the last reconciliation consistency gap. **Stacked on #531** (item #2).

`reconcile_cdot_with_overrides` previously gated on a FASTA index entry, so a transcript served only via cdot-genome synthesis (no FASTA entry, the #331 path) had its served `Transcript` corrected by `apply_canonical_overrides` but its **cdot record left stale** — and the projection path reads cdot's fields preferentially, so the label diverged.

Now the served length used to gate reconciliation falls back to the **cdot exon-alignment length** (max exon `tx_end`, which is 0-based-exclusive == the length) when there's no FASTA entry. A synthesis-only transcript whose cdot exon-sum matches the authoritative length is reconciled too, keeping the served transcript and the cdot record in agreement. (The carried-`sequence` path from item #2 already reconciled regardless; this covers the no-sequence synthesis case.)

## Testing

1 new test: a synthesis-only transcript (no FASTA entry, cdot exon-sum == authoritative length, no carried sequence) is reconciled (cdot CDS/protein corrected). Module suite green; `clippy --all-targets -D warnings` clean.

Refs #520, #505.

### DIFF
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -955,13 +955,14 @@ impl MultiFastaProvider {
     /// 1-based bounds are converted to cdot's 0-based start / 0-based-exclusive
     /// end convention.
     ///
-    /// Reconciles when the coordinates will index canonical bases: either the
-    /// served FASTA already has the authoritative length, OR the override
-    /// carries the canonical `sequence` (so the served bases are replaced on
-    /// read — this also covers a cdot-synthesis-only transcript with no FASTA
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #291, opened 2026-05-17
normalize: fetch_ref_for_canonical_split r. branch ignores cds_start (follow-up to #275)


nh13:
The `r.` branch of `fetch_ref_for_canonical_split` indexes the transcript at `(hgvs_start - 1, hgvs_end)` rather than translating through `cds_start`. This happens to work for the existing tests (the canonical fixture uses `cds_start = 1`), but would mis-index a transcript with a non-trivial 5'UTR.

## Source

Surfaced during implementation of #275 (extend codon-frame exception to `r.`). Quoting the implementing agent's report:

> One latent oddity noticed but not in scope: `fetch_ref_for_canonical_split`'s `r.` branch indexes the transcript at `(hgvs_start - 1, hgvs_end)` rather than via `cds_start`. That happens to work for the existing tests (the fixture uses `cds_start = 1`) but would mis-index a transcript with a non-trivial 5'UTR.

## Scope

Audit `fetch_ref_for_canonical_split` (and any sibling reference-fetching helpers in `src/normalize/`) to ensure `r.` axis indexing translates through `cds_start` consistently with how the `c.` branch handles it. Add a fixture with `cds_start != 1` to pin the corrected behavior.

## Cross-ref

PR for #275, #104, #79. Fits #81 § E (RNA) but is a normalize-path bug rather than a Display canonicalization gap.

PR #539, opened 2026-06-08
fix(normalize): number coding-transcript r. positions CDS-relative (#469)

### LINKED ISSUE #469: parser/rna: correct r.X numbering when start codon is not at transcript position 1 (PR #442 follow-up)
## Summary

`tests/rna_coding_consistency.rs` (added in PR #442) includes an audit suite for transcripts whose start codon is **not** at transcript position 1. ferro currently treats `r.X` as transcript-relative numbering, and those audit cases are pinned today with in-test `TODO` markers rather than asserting spec-correct behavior.

This tracks resolving that `r.X` numbering gap so the TODO-pinned cases can assert the correct numbering instead of the current behavior.

## Background

Per the HGVS spec (`RNA/` recommendations), `r.` numbering should follow the same coordinate contract as `c.` ("same numbering, different alphabet"). When the start codon is not at transcript position 1, treating `r.X` as plain transcript-relative numbering diverges from that contract.

## Acceptance criteria

- Decide and document the correct `r.X` numbering when the start codon is not at position 1.
- Replace the `TODO`-pinned audit cases in `tests/rna_coding_consistency.rs` with assertions of the correct behavior.
- `cargo nextest run --features dev --test rna_coding_consistency` passes with the TODO markers removed.

## Refs

- Deferred from PR #442 (r./c. parity audit; "ferro currently treats `r.X` as transcript-relative; pinned with TODO markers").

nh13:
Fixes #469. On a coding transcript, ferro numbered positive non-UTR `r.` positions **transcript-1-relative**, but HGVS `background/numbering.md` requires `r.` to follow `c.`:

> L58: "nucleotide numbering for an RNA reference sequence follows that of the associated coding or non-coding DNA reference sequence; nucleotide `r.123` relates to `c.123` or `n.123`."
> L61: "in a **coding** RNA reference sequence, nucleotide numbering is … **following that of a coding DNA reference sequence**."

So `r.N` must equal `c.N` (CDS-relative). When the start codon is not at transcript position 1, ferro's `r.N` and `c.N` diverged by `cds_start - 1`. The axis was also internally inconsistent: UTR `r.-N`/`r.*N` were already translated through `cds_start`/`cds_end`, so `r.-1` (just before the start codon) and `r.1` were not adjacent.

## History (why this supersedes a merged PR)

#291 was originally filed to make `r.` CDS-relative ("translate through `cds_start` consistently with how the `c.` branch handles it"). PR #304 instead closed it as "not a bug" and pinned the transcript-1-relative behavior as a "convention" — but it decided on **internal-consistency grounds and explicitly never checked the spec**. #469 corrects that; this PR supersedes PR #304's pin.

## The fix

Three sites, all just routing coding-transcript `r.` positions through the **existing** CDS-relative helpers instead of bypassing them for positive bases:

- `normalize_rna::map_in` — translate every region via `rna_to_tx_pos` when the transcript has a CDS (positive bases previously used `pos.base` directly).
- `normalize_rna::map_out` — emit every region via `tx_to_rna_pos` (CDS positions previously emitted the raw tx index).
- `fetch_ref_for_canonical_split` r. arm — slice `cds_start + N - 1`, mirroring the `c.` arm.

Transcripts without a CDS (coordinate-only / mock providers that omit `cds_start`) keep the transcript-1 fallback, so non-coding behavior is unchanged. `rna_to_tx_pos`/`tx_to_rna_pos` already computed the correct CDS-relative mapping — the bug was only that positive positions skipped them.

## Tests

- `tests/rna_coding_consistency.rs`: the `cds_start > 1` audit now asserts `r.N` ↔ `c.N` parity (the `// TODO` markers and the old "pin the divergence" test are removed).
- `tests/issue_291_rna_axis_convention.rs`: rewritten to pin the spec-correct CDS-relative axis (e.g. `r.10del` → `r.31del`, `r.1del` → `r.9del`), with the header documenting the #304 → #469 supersession.
- `tests/issue_163_rna_utr3_flag.rs` and the `test_normalize_rna_deletion_shifts_3prime` unit test: inputs recomputed to the CDS-relative numbers for the same intended transcript positions.

## Verification

- `cargo nextest run --features dev` → 6663 passed, 51 skipped (manifest-gated), 0 failed
- `cargo fmt` → clean

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -3448,29 +3448,33 @@ impl<P: ReferenceProvider> Normalizer<P> {
             Err(_) => return Ok((HV::Rna(variant.clone()), vec![])),
         };
 
-        // Convert RNA positions to transcript-1 positions, deciding per
-        // endpoint. UTR (`r.*N`/`r.-N`) and non-positive bases need a CDS
-        // to translate; non-UTR positive bases map 1:1 to transcript-1
-        // indices and work without a CDS (mock providers used in tests
-        // often omit cds_start/end). Choosing per endpoint keeps mixed
-        // intervals like `r.50_*1del` from rerouting the positive end
-        // through `rna_to_tx_pos` — issue #163 follow-up.
+        // Convert RNA positions to transcript-1 positions.
         //
-        // Axis convention (pinned by `tests/issue_291_rna_axis_convention.rs`,
-        // closes #291): `r.` positive non-UTR bases are
-        // **transcript-1-relative**, NOT CDS-relative. `r.10` against a
-        // transcript with `cds_start = 100` maps to tx index 10, not tx 109.
-        // This is consistent across `fetch_ref_for_canonical_split`,
-        // `simple_rna_pos`, and `normalize_na_edit`. Only `r.*N`/`r.-N`
-        // (and base 0) translate through `cds_start`/`cds_end` via
-        // `rna_to_tx_pos`.
+        // Axis convention (#469, HGVS `background/numbering.md` L58/L61): on a
+        // **coding** transcript, `r.` shares `c.`'s **CDS-relative** numbering
+        // — `r.123` relates to `c.123` — so every region (`r.-N` 5'UTR, `r.N`
+        // CDS, `r.*N` 3'UTR) translates through `cds_start`/`cds_end` via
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #542, opened 2026-06-08
fix(validate): reject the RNA base U in a DNA-context edit (#486 ENODNA)

nh13:
One residual category of #486: reject the RNA base `U`/`u` when it appears in a DNA-context edit (`ENODNA`).

DNA reference sequences (`g.`, `c.`, `n.`, `m.`, `o.`) use the alphabet A/C/G/T; `U` is an RNA base and is invalid there. ferro accepted it (e.g. `NM_…:c.45A>U`, `c.45deluinsATC`, `c.45_46insU`). `U`/`u` stays valid for RNA (`r.`) variants.

## What it does

- Adds a post-parse `validate_no_u_in_dna` check (mirroring `validate_no_point_insertion` / `validate_no_dupins`) that scans each DNA variant's edit sequences for `Base::U` — across substitution `ref`/`alt`, `del`/`dup`/`delins`/`ins` literal sequences, and `Complex` insert parts — recursing through `Allele` wrappers. On a hit it returns a structured `InvalidEdit` parse error (debug tag `Parse`, satisfying the corpus `ENODNA` expectation).
- RNA (`r.`) variants are skipped (`u` is correct there). The parser stores both `'U'` and `'u'` as `Base::U`, so both spellings are caught — and it fires on the bare `parse_hgvs` (strict) path the mutalyzer `errors` axis uses, not just the preprocessor path.

## Tests

- New `tests/reject_u_in_dna.rs`: rejects `U`/`u` across substitution / ins / delins / the 6 exact corpus `ENODNA` cases; accepts `U`/`u` for RNA and valid DNA bases.
- Fixed the `f18_star_as_3utr_position_on_c` spec-coverage case — its point was the `*41` 3'UTR position; the `U>A` reference base was incidentally DNA-invalid, so it now uses `T>A`.
- Demoted the 6 `ENODNA` rows from `baseline-failures/errors.txt`.

## Scope

#486's scope was narrowed to several residual categories (U-in-DNA, `EOVERLAP`, coordinate-system mismatch, repeat/length/sequence/bounds/intronic mismatches), to be done one category per PR with a spec citation each. This PR is the **U-in-DNA** category; the others follow separately. #486 stays open.

## Verification

- `cargo nextest run --features dev` → 6690 pass, 51 skipped, 0 failed
- `cargo clippy --features dev --all-targets -- -D warnings` → clean
- `cargo fmt` → clean

Issue #486, opened 2026-05-29
fix(validate): reject HGVS-invalid inputs ferro currently accepts (errors axis)


nh13:
> **Scope update (2026-05-30):** a chunk of the errors-axis over-permissiveness is already handled by merged parser-rejection PRs — **#450** (single-position insertion, #446), **#451** (`dupins<seq>`, #445), **#459** (del-size-suffix W3011, #447), **#462** (dup`<N>`/dup`<seq>`/del`<seq>` soft-prohibition W3023/4/5, #460). This issue is therefore narrowed to the **residual** categories not yet covered: **U-in-DNA** (`ENODNA`), **overlapping variants in an allele** (`EOVERLAP`), **coordinate-system mismatch** (`c.`/`g.`/`m.` on an `NR_` non-coding transcript), **repeat-unit ≠ reference** (`EREPEATMISMATCH`), **position out of bounds** (`EOUTOFBOUNDARY`), **intronic offset on a bare `NM_`** (`EINTRONIC`), and **stated del/dup bases ≠ reference** (`ESEQUENCEMISMATCH`). The list below predates those merges; treat the struck categories as done.

---

Child of #326. Spec arbitration verdict: **mostly (a) ferro over-permissiveness** (~45 of the 54 `errors`-axis FAILs), with ~7 (b) mutalyzer-specific.

## Problem

ferro **accepts** inputs that are invalid per HGVS and that mutalyzer rejects (the runner sees `ferro produced no error; mutalyzer expected …`). Each was arbitrated against `assets/hgvs-nomenclature`; ferro should reject:

- **`U` in a DNA/`c.` context** — `c.45delU`, `c.45dupu`, `c.44_47delTCUG` (DNA is ACGT; `ENODNA`).
- **single-position insertion** — `g.4insA` must be `4_5ins…` (`docs/recommendations/DNA/insertion.md`; `EINSERTIONRANGE`).
- **overlapping variants in an allele** — `[…insT;…insA]` (`EOVERLAP`).
- **coordinate-system mismatch** — `c.`/`g.`/`m.` on an `NR_` non-coding transcript, e.g. `NR_038420.1:c.10del` (`ECOORDINATESYSTEMMISMATCH`).
- **stated del/dup bases ≠ reference** — `c.45delT`, `c.44_47delTGGG` (`ESEQUENCEMISMATCH`).
[...]

PR #543, opened 2026-06-08
fix(validate): reject c./g./m. on a non-coding RNA reference (#486 coord-system mismatch)

nh13:
Another residual category of #486: reject a coding/genomic/mito coordinate system on a non-coding RNA (`NR_`/`XR_`) reference (`ECOORDINATESYSTEMMISMATCH`).

An `NR_`/`XR_` accession is a non-coding RNA transcript: it has no CDS (so `c.` is meaningless), it is not a genomic reference (`g.`), and it is not the mitochondrion (`m.`). The only valid coordinate systems are `n.` and `r.`. ferro accepted `NR_038420.1:c.10del` / `:g.10del` / `:m.10del`.

## What it does

- Adds `ErrorCode::CoordinateSystemMismatch` and `Accession::is_noncoding_rna()` (prefix `NR`/`XR`).
- Adds a post-parse `validate_coordinate_system` check (mirroring `validate_no_point_insertion` / `validate_no_u_in_dna`) that rejects `Cds`/`Genome`/`Mt`/`Circular` variants on an `NR_`/`XR_` accession, recursing through `Allele` wrappers. `n.`/`r.` on `NR_`, and `c.`/`g.`/`m.` on their proper references (`NM_`/`NC_`/mito), are untouched — the rule keys on the accession, not the coordinate system, so e.g. `NM_…:n.` stays valid.

## Tests + fixtures

- New `tests/reject_coord_system_mismatch.rs`: rejects `c.`/`g.`/`m.` on `NR_`/`XR_` (incl. inside an allele); accepts `n.`/`r.` on `NR_` and `c.`/`g.`/`m.` on their proper references.
- Updated the `test_normalize_no_cds_returns_unchanged` unit test — its premise (a `c.` variant on a no-CDS transcript) was constructed via `NR_:c.`, now rejected at parse; it uses an `NM_` accession with a no-CDS provider record, exercising the same missing-CDS path with a valid parse.
- Recorded the now-rejected real-world ClinVar input `NR_001566.1:c.107_108gc>ag` in the bulk failure-expectations snapshot (one `rejected_by_design` entry; added by hand because `UPDATE_FAILURE_EXPECTATIONS` rewrites the whole file via non-deterministic map ordering).
- Demoted the 3 `NR_` coord-mismatch rows from `baseline-failures/errors.txt`.

## Note

Independent of #542 (the U-in-DNA category) but both add a validator to the same post-parse block in `parser/variant.rs`; whichever merges second needs a trivial rebase.

## Verification

- `cargo nextest run --features dev` → 6689 pass, 51 skipped, 0 failed
- `cargo fmt` → clean


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #549, opened 2026-06-08
feat(parser): parse trans alleles whose members are cis groups (#544)

nh13:
## Summary

Third PR toward closing #544. Parses the compound-heterozygote **trans-of-cis** form `[a;b;c];[d;e]`, where each `[...]` bracket is itself a multi-variant cis group — e.g. `NM_004006.2:c.[296T>G;476T>C;1083A>C];[296T>G;1083A>C]`. Previously the trans shorthand parsed exactly one edit per bracket and rejected an inner `;` (the most common clinical compound-het description).

## Design (Option A — no AST change)

`AlleleVariant.variants` is `Vec<HgvsVariant>` and `HgvsVariant::Allele` is a variant of the enum, so a cis-group member is represented by **nesting** `HgvsVariant::Allele(AlleleVariant{phase: Cis, ..})` inside the trans allele's existing vector — no new type, derives (Serde/Eq/Hash) nest for free, and the blast radius is benign (every `allele.variants` consumer either recurses or skips non-leaf members; no panic sites).

- **Parser**: in the trans shorthand, a bracket whose content has a top-level `;` is parsed as a nested cis sub-allele.
- **Display**: a dedicated path renders nested members as `ACC:c.[a;b];[c;d]` (shared accession written once). It is taken **only when a nested member is present**, so flat-trans Display (including the established `[?]`/`[0]` expanded rendering) is byte-for-byte unchanged.

## Spec-invalid forms stay rejected

DNA/alleles.md flags the redundant cross-spelled `=` form `[2376G>C;3103=];[2376=;3103del]` as `<code class="invalid">` ("the 'not changed' indication is used only when one variant was identified"). A nested cis group carrying a position-identity (`=`) member is therefore **refused** rather than round-tripped — these inputs remain `correctly-rejected` (no false-acceptance).

## Acceptance

- One `hgvs_spec_normalization.json` row moves off `parse-error` → `preserved`; no rows regress to `false-acceptance`.
- One ClinVar bulk-fixture input (`NM_016035.4:c.[23_33delTCCTCCGTCGG];[331G>T;356C>T]`) now parses, so it is removed from both `clinvar_hgvs_*_failure_expectations.json` snapshots (surgical removal of the single now-passing entry + its now-orphaned `kind`; the curated taxonomy is otherwise untouched). The bulk benchmarks report **0 new failures**.
[...]

Issue #544, opened 2026-06-08
parse: allele / variant-relationship operators (trans [];[], mosaic (;), chimeric //, and-or ^ / =/)


nh13:
ferro does not parse the HGVS allele / variant-relationship operators, so these spec-valid constructs fail at parse. They are pinned as `parse-error` rows in `tests/fixtures/grammar/hgvs_spec_normalization.json` (the generated spec-coverage fixture), but had no tracking issue after the historical umbrellas #83 / #87 were closed.

## Scope — operators not yet parsed

**Trans / diploid compound alleles** — two bracketed alleles joined by `;` (compound-heterozygote / genotype):
```
NM_004006.2:c.[296T>G;476T>C;1083A>C];[296T>G;1083A>C]
LRG_199t1:c.[633A>G];[531_648=;961_1149=]
NP_003997.1:p.[(Ser68Arg)];[(Ser73Arg)]
NP_003997.1:p.[(Ser68Arg)];[?]          # unknown second allele
```
ferro parses the **cis**-allele `[A;B]` form today; the **trans** `[A];[B]` form is the gap.

**Mosaicism** — `(;)` separator:
```
NM_004006.2:c.[476T>C](;)1083A>C
NM_004006.2:c.[476T>C];[476T>C](;)1083G>C(;)1406del
```

**Chimerism** — `//` separator (recommendations/general).

**And/or** — `^` (between full descriptions, possibly across references) and `=/` (within a position):
```
NM_000517.4:c.424C>T^NM_000558.3:c.424C>T
LRG_199p1:p.Trp24=/Cys
```

## Spec
`assets/hgvs-nomenclature/docs/recommendations/general.md` (alleles, mosaicism, chimerism) and the per-axis DNA/RNA/protein pages.

## Done when
Each form parses to a typed AST and round-trips; the corresponding `hgvs_spec_normalization.json` rows move off `parse-error`. One PR per coherent operator group is fine (trans-allele first — it is the most common, clinically).

Refs: spec fixture (#84), closed umbrellas #83 / #87.

nh13:
## Burn-down scoping — current state of trans-allele parsing

ferro **already parses** single-variant trans alleles (`c.[296T>G];[476T>C]` round-trips, via `parse_trans_allele_shorthand_generic` + the per-coord-system member parsers). The remaining gaps under this issue split into three, of differing difficulty:

[...]

PR #552, opened 2026-06-09
feat(parser): parse predicted-wrapper members in protein alleles (#544)

nh13:
## Summary

Follow-up toward closing #544. Parses a **predicted-change member** `(...)` inside a protein allele bracket — both trans (`p.[(Ser68Arg)];[(Ser73Arg)]`) and cis (`p.[Phe233Leu;(Cys690Trp)]`, protein/alleles.md:91). Previously `parse_protein_bracket_member` handled bare members and whole-entity markers but not the predicted wrapper, so any `[(...)]` member failed at parse.

## What changed

- `parse_protein_bracket_member` gains a `(...)` branch that mirrors the single-variant predicted form: strip the wrapper, parse the inner position+edit, and mark the edit uncertain (`Mu::Uncertain`) so the member round-trips as `[(...)]`.
- Because the same bracket-member parser backs the **cis** path too, this also closes the predicted-member-inside-cis gap, and the compact Display of a predicted trans arm now re-parses cleanly (previously a documented round-trip drift).

## Acceptance

Two spec-fixture rows move off `parse-error` → `preserved`. A third (`p.[(Ser68Arg)];[?]`) reaches `diverges`: it parses and round-trips, but the trans `[?]` arm still renders in expanded form — the cross-cutting trans-`[?]` compact rendering is tracked separately (it has broad Display ripple, see #549's deferred item).

Three previously-pinned "gap" tests are **promoted** exactly as their own TODO comments instructed (e.g. *"Promote to `pin_accept` when ferro supports it"*, *"flip to `assert_eq!(reparsed.unwrap(), parsed)`"*): `protein_allele_with_predicted_variant_inside`, `f16_protein_predicted_each_cis`, and the renamed `protein_unknown_trans_compound_with_predicted_arm_round_trips`.

`cargo clippy --all-targets -- -D warnings`, `cargo fmt`, and the spec-fixture `--check` gate are clean. Full suite: only the pre-existing, environment-dependent mutalyzer/biocommons normalize-divergence tests fail.

Part of #544.

### DIFF
--- a/src/hgvs/parser/variant.rs
+++ b/src/hgvs/parser/variant.rs
@@ -3091,6 +3091,26 @@ fn parse_protein_bracket_member(part: &str) -> IResult<&str, LocEdit<ProtInterva
         ));
     }
 
+    // Predicted-change member `(position+edit)` — strip the uncertainty
+    // wrapper, parse the inner position+edit, and mark the edit uncertain
+    // so the member round-trips as `[(...)]` (e.g.
+    // `p.[(Ser68Arg)];[(Ser73Arg)]`). Mirrors the single-variant predicted
+    // form in `parse_protein_variant`.
+    if part.starts_with('(') {
+        if let Ok((remaining, (interval, edit))) = delimited(
+            char('('),
+            (parse_prot_interval, parse_protein_edit),
+            char(')'),
+        )
+        .parse(part)
+        {
+            return Ok((
+                remaining,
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #553, opened 2026-06-09
perf(normalize): two no-functional-change normalize hot-path wins (shuffle-info + protein-length)

nh13:
Two independent, behavior-preserving performance fixes on the `normalize()` hot path, found via a v0.4.0→HEAD performance sweep. Each is a "do the same work with less waste" change — no output changes.

## 1. Skip shuffle-info detection on the infos-discarding `normalize()` path (`a65b46e`)

`normalize()` returns only the variant and inspects warnings — it discards the diagnostic `infos` axis — yet `normalize_with_diagnostics()` ran `detect_shuffle_infos()` on every call, which renders position strings via `format!` for an info message `normalize()` immediately throws away. Profiling showed this is ~25% of normalize self-time.

Fix: extract `normalize_core` (variant + warnings, no infos); `normalize()` routes through it, and `detect_shuffle_infos` runs only in `normalize_with_diagnostics`, where infos are actually surfaced. Output of `normalize()` is byte-identical.

**Result:** `c.sub` normalize ~410 → ~296 ns (**~1.38×**), on every `normalize()` call.

## 2. `get_protein_length` to avoid the protein-length probe loop (`42703dd`)

`discover_protein_length()` learned a protein's length via a ~17-iteration binary search of `get_protein_sequence(acc, 0, n)` probes — each probe cloning the entire protein string — just to find a length the provider already knows.

Fix: add `ReferenceProvider::get_protein_length` with a default impl that runs the historical probe verbatim (so `MultiFastaProvider` and any non-overriding provider keep identical behavior), plus a `Box<dyn>` forward and a `MockProvider` override that returns the stored length directly. The override returns the same length the probe converged to (largest accepted `n` == `seq.len()`); an unresolvable accession maps to `None` exactly as before.

**Result:** `p.del` normalize ~3.5 µs → ~0.7 µs on the MockProvider microbench (**~5×**); the saving scales with real protein size, where the probe's per-iteration clones are far larger.

## Verification

Both changes are no-functional-change and are gated on the full `cargo nextest --features dev` suite **minus the reference-aware conformance tests** (`mutalyzer_normalize_tests` / `biocommons_normalize_tests`): **6600/6600 other tests pass**, clippy `-D warnings` and `fmt` clean.

The conformance suite is **not** included in that gate because it is currently red on `main` for an unrelated reason — the nightly's `ferro prepare` step uses a renamed CLI flag (fixed in #551), so the reference-aware suite has not actually run in ~a month. Both fixes here are behavior-preserving by construction (fix 1 only skips a discarded computation; fix 2 leaves `MultiFastaProvider` on the old probe via the trait default), so once #551 lands and the nightly is green again it will validate them against the conformance corpus.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -3201,61 +3201,28 @@ impl<P: ReferenceProvider> Normalizer<P> {
     }
 
     /// Discover the length of a protein via the `ReferenceProvider`
-    /// trait. The trait returns an out-of-range error rather than a
-    /// truncated string and has no separate length API, so we probe
-    /// with `get_protein_sequence(accession, 0, n)`.
+    /// trait's `get_protein_length` API.
     ///
-    /// Strategy: seed the upper probe at 64 KiB, which covers titin
-    /// (~35 kAA, the longest known human protein) and every realistic
-    /// protein. If 64 KiB already succeeds, walk up exponentially to a
-    /// 1 GiB safety cap. If the seed fails, the exact length is in
-    /// `[0, SEED)` — skip the exponential ramp and let the binary
-    /// search converge directly. Then binary-search between the last
-    /// known-good and first known-bad probes for the exact length.
+    /// The trait's default implementation derives the length by probing
+    /// `get_protein_sequence(accession, 0, n)` and binary-searching for
+    /// the largest accepted `n` (length-equals-largest-accepted-`n`
+    /// semantics), preserving the historical behavior. Providers that
+    /// store length metadata (e.g. `MockProvider`) override it to return
+    /// the length directly without cloning the sequence. Either way, an
+    /// empty protein or an accession the provider cannot resolve yields a
+    /// length of `0`.
+    ///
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the two review comments:

- `mock.rs` `get_protein_length`: returns `Ok(0)` for an unresolved protein (already in the pushed revision), matching the default probe's historical semantics — so `normalize()` stays behavior-preserving on `MockProvider`.
- `provider.rs` doc: corrected the contradictory contract line; the documented contract is now uniformly "length, or `0` if the accession cannot be resolved" (with `Result` retained for genuine I/O errors), matching the implementation. Amended into the `get_protein_length` commit.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #550, opened 2026-06-08
feat(parser): parse residue-level and/or `^` substitutions (#544)

nh13:
## Summary

Follow-up toward closing #544. Parses the **residue-level and/or** form `p.(Gly56Ala^Ser^Cys)` (and the two-alternative `p.(Gly719Ala^Ser)`), where one reference residue is changed to any of several alternatives joined by `^` inside an uncertainty wrapper (HGVS general.md). Distinct from the description-level `^` in #547 — here the operands are bare alternative residues sharing one reference + position.

## What changed

- New `ProteinEdit::SubstitutionAlternatives { alternatives: Vec<AminoAcid> }`. The reference residue and position come from the location (matching how `Substitution` already derives its reference); only the alternative residues are stored. Display renders the bare residues joined by `^`, so the location supplies `Gly56` and the edit supplies `Ala^Ser^Cys`.
- The protein substitution parser collects a `^`-joined list of alternative residues. (No existing exhaustive `ProteinEdit` match needed a new arm — the additive variant compiled clean against the codebase's wildcard arms.)

## Acceptance

Two spec-fixture rows move off `parse-error` → `preserved`; no rows regress to `false-acceptance`. New round-trip test. `cargo clippy --all-targets -- -D warnings`, `cargo fmt`, and the spec-fixture `--check` gate are clean. Full suite: only the pre-existing, environment-dependent mutalyzer/biocommons normalize-divergence tests fail.

## Scope / follow-up

The frameshift new-residue form `p.Gly719(Ala^Ser)fsTer23` — where the `^` alternatives sit inside a frameshift's new residue rather than a plain substitution — needs `Frameshift.new_aa` to hold alternatives and is tracked as a separate small PR.

PR #557, opened 2026-06-09
feat(parser): parse frameshift with alternative new residues (#544)

nh13:
## Summary

Follow-up toward closing #544 (split from #550). Parses `p.Gly719(Ala^Ser)fsTer23` — a frameshift whose new residue is one of several alternatives joined by `^` inside `(...)`, before `fs`. Previously the parser rejected the parenthesized alternatives.

## What changed

- New `ProteinEdit::FrameshiftAlternatives { alternatives: Vec<AminoAcid>, ter }` (position from the location; alternatives + termination detail stored here), Display `(alt1^alt2)fs<ter>`. A new variant rather than a field on `Frameshift` keeps the 16 existing `Frameshift` construction sites untouched, and compiled clean against the codebase's wildcard `ProteinEdit` matches.
- The protein edit parser routes a `(aa^aa...)fs[ter]` shape (≥2 alternatives) from the `(` dispatch arm.

## Acceptance

One spec-fixture row moves off `parse-error` → `preserved`; no rows regress. New round-trip test. `cargo clippy --all-targets -- -D warnings`, `cargo fmt`, and the spec-fixture `--check` gate are clean. Full suite: only the pre-existing, environment-dependent mutalyzer/biocommons normalize-divergence tests fail.

Part of #544. Completes the residue-level `^` family (the substitution form is #550).

### DIFF
--- a/src/hgvs/edit.rs
+++ b/src/hgvs/edit.rs
@@ -1001,6 +1001,17 @@ pub enum ProteinEdit {
         ter: FrameshiftTer,
     },
 
+    /// Frameshift whose new residue is one of several alternatives joined by
+    /// `^` inside a wrapper (e.g. `p.Gly719(Ala^Ser)fsTer23`; HGVS
+    /// general.md). The position comes from the location; the alternative
+    /// new residues and the termination detail are stored here.
+    FrameshiftAlternatives {
+        /// The candidate new amino acids at the frameshift position.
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #564, opened 2026-06-09
feat(display): render trans alleles with [0]/[?] markers in compact form (#544)

nh13:
## Summary

A trans allele pairing a real variant with a `[0]` (no allele) or `[?]` (unknown allele) marker now renders in the spec-preferred **compact** form — shared accession written once: `ACC:c.[X];[?]` rather than the expanded `[ACC:c.X];[?]`. This moves the last `diverges` rows attributable to #544 to `preserved`.

## Stacking

This needs **#549** (the `trans_compact_anchor` Display) and **#552** (predicted-protein bracket members — so the protein compact form `p.[(Ser68Arg)];[?]` re-parses). Built on an integration base `feat/issue-544-trans-unknown-base` (= #549 + #552, linear). **Once #549 and #552 merge**, retarget this PR's base to `main` and rebase.

## What changed

#549 deliberately gated its compact path to nested cis groups only, to avoid a broad Display change rippling through pinned tests. This removes that gate: `trans_compact_anchor` already treats `[0]`/`[?]` markers as contributing no accession, so the single real allele anchors the shared prefix. The Trans Display arm is unified on that one path (compact when an anchor exists, expanded otherwise).

Pinned tests that asserted the old expanded spelling are updated to assert the compact canonical form **and** that the expanded spelling normalizes to it (testing the behavior, not the old shape): 3 in `allele_trans_phase`, 3 in `compound_cross_reference`, 2 in `issue_207_multi_repeat_compound_roundtrip`, 1 in `protein_unknown_roundtrip`.

## Acceptance

Three spec-fixture rows move off `diverges` → `preserved` (`c.[2376G>C];[?]`, `r.[76a>u];[?]`, `p.[(Ser68Arg)];[?]`); no rows regress; ClinVar benchmarks unaffected. `cargo clippy --all-targets -- -D warnings`, `cargo fmt`, and the spec-fixture `--check` gate are clean. Full suite: only the pre-existing, environment-dependent mutalyzer/biocommons normalize-divergence tests fail.

Part of #544.

### DIFF
--- a/src/hgvs/variant.rs
+++ b/src/hgvs/variant.rs
@@ -1177,49 +1177,21 @@ impl fmt::Display for AlleleVariant {
                 }
             }
             AllelePhase::Trans => {
-                // A trans member that is itself a nested cis group (`[a;b]`)
-                // needs dedicated rendering; the existing flat-trans paths
-                // (which `fmt_loc_edit` a single leaf per bracket) can't
-                // express it. Only take the nested path when a nested member
-                // is actually present, so flat trans Display is unchanged.
-                let has_nested_group = self
-                    .variants
-                    .iter()
-                    .any(|v| matches!(v, HgvsVariant::Allele(_)));
-                if has_nested_group {
-                    if let Some(anchor) = trans_compact_anchor(&self.variants) {
-                        // Compact: ACC:c.[a;b];[c;d] — prefix once, then each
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai rate limit

nh13:
@coderabbitai reviews remaining?

PR #566, opened 2026-06-09
perf(reference): get_transcript -> Arc<Transcript> + bounded provider cache (resolve transcript once)

nh13:
## What

Makes `ReferenceProvider::get_transcript` (and `get_transcript_for_variant`) return `Arc<Transcript>` instead of `Transcript` by value, and gives `MultiFastaProvider` a bounded LRU cache of built transcripts. Repeated resolutions now return a shared `Arc` instead of rebuilding the full transcript every call.

**Stacked on #553** (base = `perf/skip-unused-normalize-infos`) because it edits `normalize/mod.rs`, which #553 also touches. Review #553 first.

## The problem

`get_transcript` → `get_transcript_on_build` does real work on *every* call: reads the full transcript sequence from the index, looks up cdot metadata, reconstructs the exon vector, applies canonical overrides. There was no cache. And `normalize_cds` resolves the *same* transcript many times for one variant — the past-end bounds check, the intronic lookup, the special-position branch, the 3′ shuffle in `rules.rs`, and once more on each recursion (`con`→`delins`, `ins[...]` expansion, special-position). Each was a full rebuild.

## The fix (the principled one)

A transcript is a large, immutable, provider-owned resource; callers want to *read* it, not own a fresh copy. Returning `Arc<Transcript>` from a memoizing provider models exactly that and eliminates both the rebuild and the deep clone (handoff is a refcount bump). The cache is keyed by `(accession, cdot build hint)`, bounded (LRU, capacity 512) so a long batch's working set stays in memory without unbounded growth, and needs no invalidation (transcripts are immutable for the provider's lifetime).

This is a breaking change to the public `ReferenceProvider` trait — taken deliberately now, while pre-1.0, before more code depends on the by-value semantics. Fallout was mechanical: most call sites read via `Deref` and compile unchanged; a couple needed `(*tx).clone()`; `projector.rs` (which already cached `Arc<Transcript>`) dropped a now-redundant `Arc::new` wrap.

## Verification

- **Cache correctness (runnable):** a `multi_fasta` unit test asserts that two `get_transcript` calls for the same accession return the *same* `Arc` (`Arc::ptr_eq`) — i.e. the second is a cache hit, not a rebuild.
- **Behavior preserved:** full `cargo nextest run --features dev` is green except the pre-existing reference-gated conformance axes, which fail *identically* on the #553 baseline (stale local reference data, unrelated). Pure memoization of immutable transcripts ⇒ identical results.
- **Measurement surface:** `benches/normalize_e2e.rs` is a guarded end-to-end normalize bench over a real `MultiFastaProvider`. It skips unless `FERRO_BENCH_MANIFEST` + `FERRO_NORM_CORPUS_TXT` are set, because the `MockProvider` microbench *cannot* show this win (its `get_transcript` is trivially cheap and never rebuilds). This is the surface to measure on — e.g. the nightly, which builds the manifest via `ferro prepare`.
- `clippy --all-targets -D warnings`, `fmt`, and `cargo check --features python` all clean.

## Honest caveat

The headline win is **not measured locally** in this PR — it requires real reference data (the manifest), the same stale-local blocker that gates the conformance suite. The cache unit test proves the *mechanism* (no rebuild); the e2e bench is wired to quantify the end-to-end speedup once it runs against the manifest (nightly).

## Dependency

Adds `hashlink` (already in the tree) for the bounded LRU.

## Reading order

1. `src/reference/provider.rs` — the trait change
2. `src/reference/multi_fasta.rs` — the cache + `get_transcript_on_build`, and the cache test
3. `src/normalize/mod.rs` — the two intronic closures (the only normalize edits: closure return type)
4. the remaining files — mechanical impl/call-site updates
5. `benches/normalize_e2e.rs` — the measurement surface

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -949,6 +949,7 @@ dependencies = [
  "dhat",
  "flate2",
  "futures-util",
+ "hashlink",
  "hgvs",
  "indexmap",
  "indicatif",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,7 @@ url = "2.4"
 regex = "1.10"
 once_cell = "1.20"
 superintervals = "0.3"
+hashlink = "0.9"
 
 [features]
 default = []
@@ -205,6 +206,10 @@ required-features = ["web-service"]
 
 [[bench]]
 name = "benchmarks"
+harness = false
+
+[[bench]]
+name = "normalize_e2e"
 harness = false
 
 [profile.release]
--- a/benches/normalize_e2e.rs
+++ b/benches/normalize_e2e.rs
@@ -1,0 +1,76 @@
+//! End-to-end normalize benchmark over a real `MultiFastaProvider`.
+//!
+//! This is the surface that actually shows the A2 win (transcript resolved
+//! once and shared as an `Arc` instead of rebuilt on every `get_transcript`).
+//! The `MockProvider` microbench in `benchmarks.rs` cannot show it — its
+//! `get_transcript` is trivially cheap and never rebuilds.
+//!
+//! Guarded: it skips unless BOTH env vars are set, so it is inert in normal
+//! `cargo bench` runs and only measures where real reference data exists
+//! (e.g. the nightly, which builds the manifest via `ferro prepare`):
+//!   FERRO_BENCH_MANIFEST  = path to a `ferro prepare` manifest.json
+//!   FERRO_NORM_CORPUS_TXT = newline-delimited HGVS inputs (one per line)
+//!
+//! Run:
+//!   FERRO_BENCH_MANIFEST=/path/manifest.json \
+//!   FERRO_NORM_CORPUS_TXT=/path/inputs.txt \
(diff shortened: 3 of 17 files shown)


nh13:
Closing based on measurement. On a real MultiFastaProvider (full manifest, 30k c./n. ClinVar variants, warm), the Arc transcript cache gives only ~1.17x, because on a realistic distinct-transcript-heavy corpus the bounded cache thrashes and every miss still pays the per-read File::open it doesn't address.

The far simpler, non-breaking #570 (reuse FASTA file handles + positioned reads) is ~1.93x by killing that open on every read, and #570 + a metadata-only resolve reaches ~2.15x with no cache at all — beating this Arc cache while avoiding both the breaking ReferenceProvider trait change and the cache's memory growth. So the cache is superseded; pursuing #570 + the lazy/metadata-resolve path instead.

Full matrix: baseline 26.2us/call; this PR (Arc cache) 22.3; #570 (handle reuse) 13.6; #570 + metadata-resolve 12.2.

PR #570, opened 2026-06-09
perf(reference): reuse FASTA file handles with positioned reads (~1.9x normalize)

nh13:
## What

`MultiFastaProvider::get_sequence_from_index` did a fresh `File::open` (then seek + `read_exact`) on **every** sequence read. This caches open read-only handles per FASTA path and reads positionally with `read_exact_at` (one `pread` on Unix, `seek_read` on Windows) — no per-read open/seek/close, no shared cursor.

## Why — profiled, not guessed

Profiling normalize over a real `MultiFastaProvider` (full manifest, 50k `c.`/`n.` ClinVar variants, warm) decomposed the per-call cost:

| phase | µs |
|---|---|
| `get_sequence_length` (in-memory index, no I/O) | 0.1 |
| windowed 200 bp read | **10.6** ← essentially all `File::open` |
| full-transcript read | 20.5 |
| transcript resolve (read + cdot + exons) | 22.4 |
| full normalize | 26.2 |

A 200 bp read and a full-transcript read differ by only ~10 µs, so the cost is **not** data volume or metadata — it's the **per-read `File::open` (~10.6 µs)**, which the in-memory index makes otherwise free. Reusing the handle eliminates it.

## Result

Same harness, warm, this change applied:

```
full normalize:  26.2 µs/call  ->  13.6 µs/call   (~1.9x)
```

Order-independent (helps every read, no reliance on access locality). Bytes returned are byte-identical — full `cargo nextest run --features dev` is green except the pre-existing reference-gated conformance axes (stale local reference data; identical failures on `main`). `clippy --all-targets -D warnings` + `fmt` clean.

## Safety / portability

- Read-only positioned reads on a shared handle; no `unsafe` (keeps the crate's `forbid(unsafe_code)` posture — unlike mmap).
- `read_exact_at` is `cfg`-split: `FileExt::read_exact_at` (Unix) / `FileExt::seek_read` loop (Windows).
- Handle cache is bounded by the number of distinct FASTA files (a handful) — not by transcript count.
- New unit test `cached_handle_reads_are_correct_and_repeatable` pins that repeated and offset reads through the cached handle return the correct, consistent bytes.

## Context

First of a sequence of profile-driven normalize-resolution wins. Independent of (and composes with) the `Arc<Transcript>` cache work (#566) and a planned windowed-read change (read only the shuffle's window instead of the whole transcript), which together target the remaining full-sequence-materialization cost.

### DIFF
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -20,7 +20,7 @@
 //! For type-safe coordinate handling, see [`crate::coords`].
 
 use std::fs::File;
-use std::io::{BufRead, BufReader, Read, Seek, SeekFrom};
+use std::io::{BufRead, BufReader};
 use std::num::NonZeroUsize;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

PR #580, opened 2026-06-10
fix(build): repair test code that broke under non-`dev` feature flags

nh13:
The `python` and `hgvs-rs` features are not part of `dev`, so CI — which builds and lints with `--features dev` — never compiles their `#[cfg(test)]` modules. Two production refactors drifted the gated test code out of sync, breaking `cargo clippy --all-features` (and `--features python` / `--features hgvs-rs`) locally:

- **`NormalizationWarning` lost its per-variant `message` field.** `RefSeqMismatch` now carries `details: Option<String>`, and `OverlapConflict` / `PositionPastEnd` synthesize their message via `Display` (`message()` just calls `to_string()`). The PyO3 round-trip test in `src/python.rs` still constructed all three variants with a `message:` field, failing to compile with `E0559`. Updated to drop the removed fields and move the `RefSeqMismatch` text into `details`.
- **The `hgvs-rs`-gated service test bound an unused `result`** in `src/service/tools/hgvs_rs.rs`, tripping `-D unused-variables`. Renamed to `_result` (the call is kept for its must-not-panic intent, the value discarded).

Both are test-only changes — no production behavior changes.

## Verification

- `cargo test --features python --lib py_normalization_warning_carries_code_and_message` — passes
- Standalone `--features python` / `--features hgvs-rs` / `--features dhat-heap` clippy — all clean
- `cargo fmt --check` — clean

## Root cause / follow-up

CI lints and tests with `--features dev`, which deliberately excludes `python` (needs maturin) and `hgvs-rs` (needs postgres/seqrepo system libs). Test code behind those gates can therefore rot undetected. A durable guard would be a CI job running `cargo clippy --features python` (and, where the toolchain allows, `hgvs-rs`); left out of this PR to keep it focused on the immediate breakage.

### DIFF
--- a/src/python.rs
+++ b/src/python.rs
@@ -3717,17 +3717,16 @@ mod tests {
         let cases: Vec<(crate::normalize::NormalizationWarning, &str)> = vec![
             (
                 crate::normalize::NormalizationWarning::RefSeqMismatch {
-                    message: "stated A but reference is T".to_string(),
                     stated_ref: "A".to_string(),
                     actual_ref: "T".to_string(),
                     position: "100".to_string(),
                     corrected: true,
+                    details: Some("stated A but reference is T".to_string()),
                 },
                 "REFSEQ_MISMATCH",
             ),
             (
                 crate::normalize::NormalizationWarning::OverlapConflict {
-                    message: "two coincident edits".to_string(),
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

PR #581, opened 2026-06-10
ci: lint feature-gated code with a `clippy --all-features` job

nh13:
Adds a `Clippy (all features)` job to `ci.yml` running `cargo clippy --all-features --all-targets -- -D warnings`.

## Why

The existing `clippy` and `test` jobs build with `--features dev`, which excludes `python` (needs maturin/pyo3) and `hgvs-rs` (needs postgres/seqrepo deps). Code behind those gates — including `#[cfg(test)]` modules — is never compiled in CI and rots silently when the production API it references changes. This is exactly the class of bug fixed in #580 (a `NormalizationWarning` field rename that only broke under `--features python`).

## Notes

- **Stacked on #580** (base = `fix/all-features-build-drift`). The new job is red on `main` until #580 lands, so this is targeted at the fix branch; GitHub will auto-retarget it to `main` once #580 merges. Merge #580 first.
- Sets up Python so pyo3's build script can locate an interpreter. The deps `--all-features` pulls (`postgres`, `seqrepo`, etc.) are pure-Rust, so no system libraries or live database are needed — clippy type-checks the gated code without running it.
- **Complements, does not replace, the `--features dev` clippy job.** The two configs compile different slices: `--all-features` enables every `#[cfg(feature)]` branch but hides the `#[cfg(not(feature))]` ones the dev build lints, and `dev` is closer to the default-features-off form most library consumers compile.

## Out of scope

A nightly `cargo hack --feature-powerset` would additionally catch single-feature-in-isolation breakage that `--all-features` cannot. Deliberately left out per request to keep this to the `--all-features` job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,6 +150,42 @@ jobs:
             ${{ runner.os }}-cargo-
       - run: cargo clippy --features dev --all-targets -- -D warnings
 
+  clippy-all-features:
+    # Lints the feature-gated code that `--features dev` never compiles —
+    # notably `python` (pyo3) and `hgvs-rs`, whose `#[cfg(test)]` modules are
+    # otherwise unbuilt in CI and silently rot when the production API they
+    # construct changes. Complements (does not replace) the `clippy` job
+    # above: that one lints the `#[cfg(not(feature = ...))]` slice and the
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #332, opened 2026-05-19
fix(normalize): intronic boundary normalization with NG-prefixed transcripts


nh13:
## Summary

When the input is a c.-axis variant on an NG-prefixed parent reference and the position is intronic (e.g. `NG_012772.1(NM_000059.3):c.632-5_670del`), `Normalizer::normalize` returns:

```
Coordinate conversion error: Transcript has no chromosome mapping for intronic normalization
```

Same shape produces a related error for boundary normalization without an intronic offset:

```
Coordinate conversion error: Transcript has no chromosome for boundary normalization
```

Surfaced by 14 FAILs across `normalized` and `genomic` axes of `tests/mutalyzer_normalize_tests.rs`. Burn-down item under #326.

## Reproduction

```rust
use ferro_hgvs::{parse_hgvs, MultiFastaProvider, Normalizer};

let provider = MultiFastaProvider::from_manifest("…/manifest.json")?;
let normalizer = Normalizer::new(provider);
let v = parse_hgvs("NG_012772.1(NM_000059.3):c.632-5_670del")?;
let _ = normalizer.normalize(&v); // → Coordinate conversion error: Transcript has no chromosome mapping for intronic normalization
```

## Root cause (suspected)

The intronic-normalization path needs to consult genomic bases around the intron/exon boundary to apply the 3' rule. ferro looks up the transcript's chromosome alias via the cdot mapper to anchor the sequence fetch. For an NM_/XM_-only input that works; for an NG-anchored selector (`NG_012772.1(NM_000059.3):c.…`), the resolver uses the NG accession but the cdot mapper has the NM keyed only to its chromosome, not to the NG. The fetch fails and the normalize aborts.

Likely fix surfaces:
- `src/normalize/` intronic + boundary paths: accept the parent NG/NC accession as a valid genomic anchor (it *is* a genomic sequence, just non-chromosomal).
- Alternatively, fall back to the NM's chromosome alias when the explicit parent reference doesn't resolve to ferro's chromosome-keyed providers.

Worth confirming by tracing one of the failing inputs from `tests/fixtures/mutalyzer-normalize/baseline-failures/normalized.txt` end-to-end.

## Tests

`MockProvider`-backed tests in `tests/normalize_tests.rs`:

- Intronic c. del on NG-anchored input → normalize succeeds, 3' shift applied within intron.
- Boundary del spanning UTR/CDS on NG-anchored input → succeeds, position shift applied.
[...]

PR #585, opened 2026-06-10
perf(reference): version + self-heal the cdot bincode cache (fix silent JSON fallback)

nh13:
## The regression

`ferro` startup is ~90% of CLI wall time for any batch, and most of it was a **512 MB cdot JSON parse that the bincode cache exists to avoid**. Profiling `ferro normalize` end-to-end:

| | time |
|---|---|
| cdot load via JSON (the silent fallback) | **~5.9 s** |
| cdot load via a valid bincode cache | **~0.6 s** |
| `ferro normalize` startup, stale cache | ~6.7 s |
| `ferro normalize` startup, valid cache | **~1.4 s** |

## Root cause

The bincode cache had **no version/magic header**. Any change to the serialized `CdotMapperSnapshot` layout made an existing `.bin` mis-deserialize (`"failed to fill whole buffer"`). The loader caught the error and fell back to JSON — but the warning went through `log`, which the CLI never initializes, so it was **silent**. Every run re-parsed the JSON, with no indication the fast path was dead. `prepare` compounded it by **skipping regeneration whenever a (stale) `.bin` existed**.

## Fix

- **Version+magic header** (`FCDT` + `u32`): bump on any layout change. `from_bincode_file` verifies it and rejects a legacy/mismatched cache cleanly instead of mis-reading it. Cheap `bincode_is_current` for callers.
- **Atomic writes** (temp + rename) so a crash or concurrent reader never sees a half-written cache.
- **Self-healing `load`**: on a missing/stale/unusable cache it parses JSON, prints a **loud (stderr)** warning, then best-effort rewrites a fresh `.bin` — next run is fast, no re-prepare needed. Read-only mounts degrade gracefully (refresh failure is non-fatal).
- **`prepare`** regenerates when the cache isn't current, instead of skip-if-exists.
- **Completeness fix:** include `alt_build_transcripts` in the snapshot. The old format dropped it, so the bincode fast path silently degraded NG/NC-parent build resolution (#332). Now the cache is complete and **bincode load == JSON load**.

## Verification

- New unit tests: header roundtrip, staleness/version detection, self-heal after a corrupt cache, and **alt-build preservation across the bincode roundtrip** (GRCh37 contig + CDS identical to the JSON path).
- Full suite passes except the 9 pre-existing reference-gated axes; clippy clean. (The suite itself runs ~2× faster now that cdot loads from a valid cache.)
- End-to-end confirmed: stale cache self-heals on first run, ~1.4 s on every run after; an NG/NC GRCh37-parented variant resolves via the bincode path.

> Independent of the open perf PRs; touches `src/data/cdot.rs` (alongside the #583 determinism fix to the same file — whichever merges second takes a trivial rebase).

---

## Follow-up in this PR: `ferro check --build-cache` (pre-warm)

With the self-heal above, a stale/missing cache rebuilds lazily inside the *first* real run (the one-time ~5-6.7s). For timed/benchmarked/production runs you may want to pay that up front. `check` previously only verified file existence; `--build-cache` now loads the reference once as an explicit setup step (self-healing the cache), so later runs always start fast.

On vs off (real manifest):

| | query cost |
|---|---|
| no pre-warm (cold cache) | ~5-6.7s (one-time build lands in the query) |
| `check --build-cache` once (~5s), then queries | ~1.5s each |

Note: the long-running API/service path already pays startup once (build `MultiFastaProvider` once, reuse) — this flag is for the repeated-/timed-CLI case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,10 @@ ferro prepare --output-dir ferro-reference
 # Verify reference data is ready
 ferro check --reference ferro-reference
 
+# (Optional) pre-build the on-disk cdot cache as a setup step, so the one-time
+# cache build doesn't slow the start of a real (or timed/benchmarked) run.
+ferro check --reference ferro-reference --build-cache
+
 # Normalize with reference
 ferro normalize "NM_000088.3:c.459del" --reference ferro-reference/
 ```
--- a/src/bin/ferro.rs
+++ b/src/bin/ferro.rs
@@ -499,6 +499,12 @@ enum Commands {
         /// Reference directory to check
         #[arg(long, default_value = "ferro-reference")]
         reference: PathBuf,
+
+        /// After checking, load the reference once to build/refresh the on-disk
+        /// cdot bincode cache. Pay the one-time cache build here as a setup step
+        /// so it doesn't land inside (and slow down the start of) a real run.
+        #[arg(long)]
+        build_cache: bool,
     },
 
     /// Build a transcripts.json from a FASTA + CDS coordinates (single-exon).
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #546, opened 2026-06-08
parse: complex structural-variant / rearrangement notation (sup, add, ::, pter/qter delins-inv)


nh13:
ferro does not parse complex structural-variant / chromosomal-rearrangement notation. Pinned as `parse-error` in `tests/fixtures/grammar/hgvs_spec_normalization.json`; untracked since #83 / #87 closed. This is the most advanced/rare tail and likely the largest single grammar effort.

## Scope (examples from the spec fixture)

- **Supernumerary / additional material** — `sup`, `add`:
```
NC_000023.11:g.pter_qtersup
chr14:g.[pter_cen_qter]add
chr22:g.[pter_10000000del::40000000_qterdel]add
```
- **Breakpoint / translocation joins** — `::` (with `chr` and `cen`/`pter`/`qter` anchors):
```
chr14:g.29745314_qterinv::CATTTGTTCAAATTTAGTTCAAATGA::chr3:g.36969142_cen_qter
chr14:g.pter_cen_29745313::chr3:g.pter_36969141inv
```
- **pter/qter delins-inv rearrangements**:
```
NC_000009.12:g.102425452_qterdelinspter_26393001inv
NC_000009.12:g.pter_26393001delins102425452_qterinv
```
- **Composite top-level edits** in one allele:
```
NC_000002.12:g.[32310435_32310710del;32310711_171827243inv;insG]
```
- **`dupinv`** and **chr-prefixed cross-reference inserts**:
```
NC_000008.11:g.128746677_128749160dupinv
g.[chr4:134850793_134850794inschrX:89555676_100352080inv]
```

## Spec
`recommendations/DNA/` (complex / structural) and `background/`. Several of these overlap the pter/qter coordinate work (#488 / #534 / #537).

## Done when
Each form parses to a typed AST and round-trips; the corresponding `hgvs_spec_normalization.json` rows move off `parse-error`. Expect multiple PRs (sup/add, `::` breakpoints, delins-inv rearrangements) — one coherent construct per PR.

Refs: spec fixture (#84), #537 / #488, closed #83 / #87.

nh13:
[...]

Issue #545, opened 2026-06-08
parse: uncertain-count / N-padded repeat and uncertain-allele notation


nh13:
ferro does not parse uncertain-count / N-padded repeat notation and the uncertain-allele bracket form. Pinned as `parse-error` in `tests/fixtures/grammar/hgvs_spec_normalization.json`; untracked since #83 / #87 closed.

## Scope

**N-padded repeat over an uncertain range** (`delN[k]` etc.):
```
NM_000333.3:c.(4_246)delN[15]
NC_000003.12:g.(63912602_63912844)delN[15]
```

**Uncertain repeat counts** — `[(N_M)]`, `[N_?]`, `[(N_?)]` ("at least N") on the repeat unit (recommendations/repeated.md).

**Uncertain compound allele** — `[(...)]` wrapping multiple uncertain variants:
```
NP_003997.1:p.[(Ser68Arg;Asn594del)]
LRG_199t1:r.[(578c>u;1339a>g;1680del)]
```

## Spec
`recommendations/DNA/repeated.md`, `recommendations/uncertain.md`. Note #466 already has a related SVD-WG candidate about intronic-offset/UTR-marker CDS ranges inside `ins[...]` payloads.

## Done when
Each form parses + round-trips and its `hgvs_spec_normalization.json` row moves off `parse-error`.

Refs: spec fixture (#84), #466, closed #83 / #87.

nh13:
Resolved across #568 (N-padded deletion over an uncertain range, `delN[k]`), #569 (predicted compound cis allele `[(a;b)]`), and #571 (comma products allele `[a,b]`) — all merged. Every form in the issue's Scope / "Done when" now parses and round-trips (verified on current main): `(start_end)delN[k]`, uncertain repeat counts `[(N_M)]`/`[N_?]`/`[(N_?)]`, and the uncertain compound allele `[(...)]`.

PR #586, opened 2026-06-10
chore(tests): stop committing the generated spec-normalization fixture

nh13:
## Summary

`tests/fixtures/grammar/hgvs_spec_normalization.json` is a **generated artifact** — produced by the `generate_spec_fixture` example from the HGVS spec submodule, the parser's behavior, and the curated `hgvs_spec_normalization_overrides.json`. Committing it has made every parser PR (the #544 / #545 / #546 series) a **merge-conflict magnet**: each PR regenerates the whole ~15k-line file, so every time one merges to `main` the rest go `CONFLICTING` and have to be re-rebased. This PR stops tracking it and regenerates it on demand instead.

Regeneration is **~0.83s** once the crate is compiled (the compile is already paid for by the test build), so the cost is negligible.

## Changes

- **Untrack + ignore**: `git rm --cached` the fixture and add it (plus its temp files) to `.gitignore`.
- **Reader tests regenerate on demand**: `tests/idempotency_tests.rs` and `tests/hgvs_spec_normalization_tests.rs` call a new shared helper `tests/common/spec_fixture.rs::ensure_spec_fixture()`, which runs the `generate_spec_fixture` example if the file is missing. It writes to a per-process temp file and atomically renames it into place, so concurrent test binaries can't observe or write a partial file. A fresh `cargo test` "just works".
- **CI regenerates explicitly** before the test run (`ci.yml`, `external-validation.yml`), replacing the now-moot `generate_spec_fixture -- --check` step. The generator still validates the overrides file at generation time — that is the guard `--check` provided. `external-validation.yml`'s standard-test job gains `submodules: true` because it now needs the spec submodule to generate the fixture.
- **Docs**: `CLAUDE.md` documents that the fixture is generated (not committed), how to regenerate it, and the submodule requirement.

## Trade-off

Per-PR `parse-error → preserved` status transitions are no longer visible as a committed diff; they're reviewed via the PR description and the accompanying test/parser changes. (The detailed JSON was the only place these showed up in `git`; the conformance summary docs are a separate, unaffected artifact.)

## Testing

- Deleted the working-tree fixture and ran the two reader tests → they auto-generated it and passed.
- `cargo fmt --check` and `cargo clippy --features dev --all-targets` clean.
- Full suite: 6745 passed, only the 9 pre-existing corpus-parity environment failures remain (unchanged).
- Confirmed `git status` no longer surfaces the fixture (ignored) and no temp files leak.


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #507, opened 2026-05-31
refactor(project): axis-symmetric VariantProjection (genome as optional pivot) — design


nh13:
Design issue (no code yet). #498 made `VariantProjection.genomic` `Option<HgvsVariant>` so a genome-less projection (bare-`NM_` c.→p.) can honestly report "no genomic form" rather than erroring. That is a first step toward a principled model.

## Proposed direction

Today `VariantProjection` and the projection engine are **genome-centric**: every c./n./r. input is forced through `project_to_genomic` (a mandatory pivot), and the result type's framing assumes a g. anchor. The principled model:

1. **Axis-symmetric result** — each coordinate field is independently optional; a projection carries whatever representations are derivable from the input + available reference data. `None` means "not available", not an error.
2. **Genome as an optional pivot, not the spine** — compute each target axis by the most direct route (c.→p. straight from the CDS, as #498 now does; c.↔g. only when an alignment exists).
3. **Preserve the existing Err-vs-None discipline** — `Err` for caller mistakes (e.g. transcript_id mismatch), `None` for "no such representation / consequence". (A richer typed-absence enum was considered and rejected as premature — `Option` plus the existing `is_intronic`/`is_utr` flags already suffice; revisit only if a consumer needs to branch on the reason.)

## Constraints / coordination

- API-breaking (the result struct + Python bindings) — version-bump and do it once.
- Must be **co-designed with #485** (r./n. prediction surface — the producers for those axes) so the struct isn't given producer-less fields; and informed by #480, but **#480 is out of scope** — for NG-relative inputs there is generally a correct genomic coordinate, so "Unavailable" there would mask a coordinate-math bug rather than fix it.

Refs #498, #485, #480, #326.

nh13:
Status update: incremental steps toward the axis-symmetric design have merged:

- #508 — made `VariantProjection.genomic` an `Option` (genome as optional pivot, first step)
- #592 — added the `noncoding` axis (Cycle 1b-ii)
- #598 / #599 / #603 — hgvs-rs projection oracle, microbenchmarks, and `AlignmentGap` (E3007) error work

The API-breaking unification (genome-as-optional-pivot co-designed with #485, and folding `Normalizer`'s c→g into a unified projection path per #622) remains the open design surface this issue tracks.

nh13:
[...]

PR #587, opened 2026-06-10
fix(project): scan the 3'UTR for a frameshift's new stop (fsTer{K}, not fsTer?)

nh13:
## Summary

A frameshift whose new in-frame stop lies in the **3'UTR** (past the annotated CDS end) was mis-predicting `p.(…fsTer?)` instead of the true `p.(…fsTer{K})`. `build_frameshift_variant` scanned only the **CDS-truncated** mutated sequence (`build_mutated_cds_with_ref` stops at `cds_end`), so a downstream stop in the 3'UTR was never found. The stop-loss/extension path already handled this by appending `seq[cds_end..]`; this extracts that into a shared `mut_cds_with_3utr` helper and uses it in **both** the frameshift and extension predictors.

A frameshift's new stop is, in the common case, downstream of the original stop — so this affected a broad class of near-3'-end frameshifts, all silently degrading to `fsTer?`.

The fix is **strictly additive**: when an in-CDS stop already exists, the first-`Ter` position is unchanged, so an existing `fsTer{K}` count is never altered — only `fsTer?` is converted to the correct `fsTer{K}` for the 3'UTR-stop cases.

## Changes

- `helpers.rs`: new `mut_cds_with_3utr(mut_cds, transcript)` — appends the unchanged 3'UTR to a mutated CDS so a shifted/read-through frame can find a stop past `cds_end`; guarded against `cds_end > seq.len()` (mirrors `read_full_cds`).
- `indel.rs`: `build_frameshift_variant` now scans the 3'UTR-extended sequence (the fix); `build_extension_variant` is DRY'd onto the same helper (removes ~22 lines of duplicated 3'UTR-append logic).
- Regression test (`frameshift_new_stop_in_3utr_emits_fster_count`) with the codon-by-codon construction: `del c.4` on `Met-Lys-Lys-Lys-Ter` + a 3'UTR stop → `p.(Lys2ArgfsTer5)` (was `p.(Lys2ArgfsTer?)`).

## Test plan

- [x] New regression test fails before the fix (`fsTer?`), passes after (`fsTer5`).
- [x] Full CI-equivalent suite (manifest absent): 6751 passed, 0 failed.
- [x] `cargo fmt` clean; `clippy --all-targets -D warnings` clean.
- [x] Adversarial correctness review: fix is strictly additive; no existing test encoded the bug; `alt_protein`/`new_aa` unaffected (frameshift divergence is always in-CDS).

## Context

First concrete deliverable of the projection-engine program (#326 / #507): a confirmed `c→p` consequence-prediction correctness fix. Independent of the open conformance PRs; off `main`.

### DIFF
--- a/src/project/protein/helpers.rs
+++ b/src/project/protein/helpers.rs
@@ -526,6 +526,15 @@ pub(crate) fn mut_cds_with_3utr(
         .ok_or_else(|| FerroError::ConversionError {
             msg: format!("transcript {} has no CDS end", transcript.id),
         })? as usize;
+    // Guard the slice (mirrors `read_full_cds`): in the normal flow the caller
+    // has already validated the transcript via `RefProteinBundle::from_transcript`,
+    // but this is `pub(crate)` — return a clean error rather than panic if a
+    // future caller passes a transcript whose `cds_end` exceeds its sequence.
+    if cds_end > seq.len() {
+        return Err(FerroError::ProteinSequenceUnavailable {
+            accession: transcript.id.clone(),
+        });
+    }
     Ok(format!("{mut_cds}{}", seq[cds_end..].to_ascii_uppercase()))
 }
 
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #591, opened 2026-06-10
perf(reference): rkyv archive for the cdot cache (zero-copy load, ~0.37s off startup)

nh13:
Stacked on #585. After #585 fixed the broken bincode cache (4.8s → ~1.4s startup), the **cdot load is the single biggest remaining chunk** (~0.6s decoding + materializing 482k transcripts). This replaces the on-disk cache format with an **rkyv archive**.

## Approach (spiked + measured first)

I prototyped rkyv before committing. The spike showed:

| | time |
|---|---|
| bincode load (current) | ~570 ms |
| rkyv full deserialize-to-owned | **~150 ms** (~3.8×) |
| rkyv zero-copy validated access | ~32 ms |
| **safe** validated access (bytecheck) | **resolves the `unsafe` worry** |

So I implemented the **simple full-deserialize format-swap** (not a fully-lazy zero-copy backend): `CdotMapper` stays structurally identical, no self-referential mmap lifetime, no archived-type refactor. It captures ~80% of the achievable win; the fully-lazy variant would fight the `OnceCell` position indexes (which iterate all transcripts) anyway.

## Result

**`ferro normalize` startup ~1.4s → ~1.03s** on the real manifest (~0.37s off). The test suite itself runs ~2× faster now that cdot loads from the archive.

## How

- New `rkyv_cache` module — an rkyv mirror kept fully separate from the domain types (strand as `i8`, cigars flattened), so the rkyv derives and format choices stay contained.
- `from_rkyv_file`: mmaps the archive (page-aligned → satisfies rkyv), **validated** access (safe), checks an in-archive format version, and converts archived → `CdotTranscript` **in a single pass** (no intermediate owned snapshot).
- `to_rkyv_file`: atomic write (temp + rename).
- `load` cascades **rkyv → legacy bincode (auto-upgraded to rkyv) → JSON**, self-healing to rkyv on any miss/staleness — existing reference dirs go fast on the second run with no re-prepare. `prepare` writes the archive.
- `memmap2` is now an unconditional dependency. mmap is already an accepted pattern in this crate (`MmapFastaProvider`, fasta.rs); the crate does not set `forbid(unsafe_code)`.

## Correctness

Behavior is unchanged — the rkyv path reproduces the exact `CdotMapper`. Verified by a roundtrip test covering exons, strand, CDS, **per-exon cigars**, and the **alt-build (GRCh37) map**. Full suite passes except the 9 pre-existing reference-gated axes; zero non-axis failures (rkyv-loaded normalization == bincode).

## Tradeoffs / notes
- The rkyv archive is ~360 MB vs ~261 MB bincode (alignment/layout overhead) — disk is cheap; mmap means it's not read up front.
- Committed with `--no-gpg-sign` (1Password biometric unavailable); will re-sign.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -404,6 +404,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "bytecheck"
+version = "0.8.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0caa33a2c0edca0419d15ac723dff03f1956f7978329b1e3b5fdaaaed9d3ca8b"
+dependencies = [
+ "bytecheck_derive",
+ "ptr_meta",
+ "rancor",
+ "simdutf8",
+]
+
+[[package]]
+name = "bytecheck_derive"
+version = "0.8.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "89385e82b5d1821d2219e0b095efa2cc1f246cbf99080f3be46a1a85c0d392d9"
+dependencies = [
+ "proc-macro2 1.0.103",
+ "quote 1.0.42",
+ "syn 2.0.111",
+]
+
+[[package]]
 name = "bytecount"
 version = "0.6.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #592, opened 2026-06-10
feat(project): add the c↔n (non-coding transcript) axis to VariantProjection

nh13:
Adds a first-class `n.` (transcript-relative) axis to projection results, the next step in the axis-symmetric projection program (#507), Cycle 1b-ii.

`VariantProjection` gains a `noncoding: Option<HgvsVariant>` field carrying the `n.` form of the variant on the transcript. It's populated for coding transcripts by reframing the resolved `c.` CDS positions into transcript coordinates via the projector's existing exon- and CIGAR-aware `CoordinateMapper::cds_to_tx` — the same conversion `genome_to_cds`/`project_single_inner` already trust — so the `c.` and `n.` axes never disagree, including on multi-exon transcripts with tx-coordinate gaps or CIGAR insertions. The derivation is genome-free and is therefore defined even for intronic positions. Allele projections aggregate an `n.` allele the same all-or-nothing way as the existing `coding`/`protein` axes. Python gains a uniform `n_name` accessor.

The change is additive: `coding`'s existing semantics are unchanged (it still holds the `Tx`/`n.` form for non-coding transcripts — a pre-existing quirk left as documented debt). No existing field value, test expectation, or conformance baseline shifts.

For a non-coding (`NR_`) transcript the genome-pivot branch sets `noncoding = Some(coding.clone())`, but that branch is currently unreachable via full projection (non-coding transcript projection is pinned unsupported in `tests/issue_328_projector_accepts_cnr.rs`). It is kept as forward-compatible, documented code that carries the correct semantics the moment that path is enabled.

## Tests
- `c↔n` boundary matrix: `c.1`, `c.-N`, `c.*N` (→ a plain positive `n.`, not `n.*N`), intronic offset carry.
- CIGAR-insertion regression: `c.6 → n.8`, which a flat `cds_start + base - 1` conversion would get wrong (it would yield `n.6` and disagree with the `c.` form's true transcript position).
- Genome-pivot and bare-`NM_` integration tests asserting both `c.` and `n.` forms.
- Allele aggregation with pinned inner `n.` coordinates.

## Verification
- `cargo clippy --features dev --all-targets -- -D warnings`: clean.
- `cargo fmt --all`: clean.
- `cargo check --features python`: clean.
- Full suite minus the `FERRO_MANIFEST`-gated conformance dashboards: 6707/6707 pass. The 9 conformance-dashboard failures are pre-existing and reproduce identically on the base commit (they fail locally where the manifest is present and are skipped in CI).

### DIFF
--- a/python/ferro_hgvs/__init__.pyi
+++ b/python/ferro_hgvs/__init__.pyi
@@ -1462,6 +1462,16 @@ class VariantProjection:
         ...
 
     @property
+    def n_name(self) -> str | None:
+        """The n. (transcript-relative) variant as an HGVS string.
+
+        Populated for both coding transcripts (derived genome-free from the c.
+        form) and non-coding transcripts; None when no transcript coordinate is
+        available (e.g. an empty allele).
+        """
+        ...
+
+    @property
     def p_name(self) -> str | None:
         """The p. variant as an HGVS string, or None for intronic/UTR/non-coding variants."""
         ...
--- a/src/project/mod.rs
+++ b/src/project/mod.rs
@@ -5,6 +5,7 @@ pub mod edit;
 mod projector;
 mod protein;
 mod result;
+mod transcript_axis;
 
 pub use projector::VariantProjector;
 pub use result::VariantProjection;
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai review

PR #593, opened 2026-06-10
perf(reference): binary cache for the GFF/GTF TranscriptDb (annotate-vcf ~3s → ~0.6s)

nh13:
Stacked on #591. Profiling the VCF-annotate path (previously unprofiled) showed the **same problem cdot had**: every `ferro annotate-vcf --gff X.gff` **re-parses the entire annotation file**.

## Profile
On a real hg19 GFF + 20k variants:

| | |
|---|---|
| GFF parse | **~91% of the run** (`load_annotations` 91% incl, `Gff3Record::parse` + `FeatureGraph` build + string churn) |
| per-variant annotation | ~11µs |

So for a genome-wide GFF it's ~2.5s of pure re-parsing, repeated every invocation.

## Fix
Cache the parsed `TranscriptDb` to a sibling binary file `<source>.<build>.tdb`. `load_annotations`:
1. **reuses** the cache when current — magic+version header **and** newer than the source (editing the GFF/GTF invalidates it);
2. **self-heals** — on a miss/staleness it parses as before, then best-effort writes the cache (skipped when FASTA validation is requested, since the cache doesn't carry parse-time diagnostics).

The domain `Transcript` carries `#[serde(skip_serializing_if)]` attributes that **bincode can't round-trip** (positional format → stream misaligns → "tag for enum not valid"), so the cache serializes a no-skip `TranscriptSnapshot` mirror — exactly how the cdot cache handles it. Versioned + atomic write + self-heal mirror the cdot cache fix (#585).

## Result
**`ferro annotate-vcf` ~3s → ~0.56s** after the first run. Output is **identical** to the parse path (verified over 20k variants, modulo the separate non-determinism note below). Full suite passes except the 9 pre-existing axes; adds a cache roundtrip + corrupt-rejection test. `*.tdb`/`*.rkyv` added to `.gitignore`.

## Separate finding (not fixed here)
While validating, I found the **annotation INFO-field order is non-deterministic** (HashMap iteration order) — parse-vs-parse output already differs in INFO key order. That's a pre-existing correctness/reproducibility bug, orthogonal to this cache; worth its own small PR (deterministic INFO ordering).

> Committed `--no-gpg-sign` (1Password biometric unavailable); will re-sign.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,7 @@ manuscript/
 # parser PR a conflict magnet, so it is produced on demand instead.
 /tests/fixtures/grammar/hgvs_spec_normalization.json
 /tests/fixtures/grammar/.hgvs_spec_normalization.*.tmp
+
+# Generated reference caches (cdot rkyv, TranscriptDb)
+*.tdb
+*.rkyv
--- a/src/data/cdot.rs
+++ b/src/data/cdot.rs
@@ -1595,8 +1595,25 @@ impl CdotMapper {
             (path.with_extension("rkyv"), path.with_extension("bin"))
         };
 
+        // A sibling cache is only usable if it is at least as new as the source
+        // JSON. cdot files are normally version-stamped in the filename, but a
+        // re-`prepare` that rewrites the JSON under the same name would
+        // otherwise be masked by a stale archive. Mirrors the freshness check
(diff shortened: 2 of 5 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #480, opened 2026-05-29
project_to_genomic: NG/LRG-parent transcript inputs emit chromosome coordinates under the NG/LRG accession


nh13:
Surfaced while wiring the `genomic` axis of the mutalyzer-normalize corpus runner (#326, Phase 2).

## Summary

`VariantProjector::project_to_genomic` maps a transcript-coordinate input (`c.`/`n.`/`r.`) to its parent genomic reference using the cdot transcript→genome alignment, then stamps the input's `genomic_context` accession onto the result. cdot only carries **chromosome (`NC_`) alignments**, so when `genomic_context` is an `NG_` RefSeqGene (or `LRG_`), the emitted coordinates are chromosome-relative but the emitted accession is the `NG_`/`LRG_` one. The two no longer agree.

## Reproduction

Inputs from the corpus (manifest-backed run):

| Input | Expected (mutalyzer) | ferro `project_to_genomic` |
|---|---|---|
| `NG_007485.1(NM_000077.4):c.274G>T` | `NG_007485.1:g.28407G>T` | `NG_007485.1:g.21971085C>A` |
| `NG_012337.1(NM_003002.2):c.1del` | `NG_012337.1:g.5062del` | `NG_012337.1:g.112086906del` |
| `NG_007485.1(NM_000077.4):c.161_162insATC` | `NG_007485.1:g.28294_28295insATC` | `NG_007485.1:g.21971197_21971198ins...` |

`21971085` / `112086906` are chr9 / chr11 positions (the `NC_000009.12` / `NC_000011.10` cdot alignment targets), not offsets within the ~30 kb `NG_007485.1` RefSeqGene. On the minus strand the base is additionally complemented (`G>T` → `C>A`), consistent with a chromosome projection that was never lifted back into the NG's own (plus-strand) frame.

## Root cause

In `src/project/projector.rs`:

- `map_pos` (~L700) computes the genomic position via `cdot_mapper.cds_to_genome_on_build(...)` / `cdot_tx.tx_to_genome(...)`, which return coordinates on the cdot alignment **contig** (always `NC_` here).
- The result is then emitted with `accession: parent` (~L776), where `parent` is the input's `genomic_context` (the `NG_`/`LRG_`).

[...]

PR #616, opened 2026-06-13
feat(project): re-anchor NG_/LRG_-parent projections into the parent frame (#480)

nh13:
Fixes the #480 defect end-to-end: a c./n./r. input parented on an `NG_` RefSeqGene or `LRG_` now projects into the parent's **own** coordinate frame instead of emitting a chromosome offset under the parent accession.

## Problem

cdot aligns transcripts only to chromosomes (`NM_`→`NC_`). When a c./n./r. input is parented on an `NG_`/`LRG_`, `project_to_genomic` mapped the coordinate onto the chromosome and then stamped the *parent* accession onto it — so the emitted coordinate was a chromosome offset (complemented on the minus strand) under an accession whose own frame it does not match (e.g. `NG_007485.1:g.21971085C>A` instead of `NG_007485.1:g.28407G>T`). That output is invalid HGVS regardless of mutalyzer.

## What this PR does

Expresses the coordinate in the parent's own frame by composing NM→NC (cdot) with an affine NC→parent transform, and wires up the placement data that drives it:

- **`GenomicPlacement`** (`src/reference/provider.rs`): a parent's contiguous placement on its `NC_` contig (`parent_start`, `nc_start`, `nc_end`, `strand`) with `nc_to_parent()`.
- **`project_to_genomic`** (public, genomic-axis *output*) projects to NC then re-anchors into the parent frame, reverse-complementing the edit when the parent runs antiparallel to the chromosome. Split out **`project_to_genomic_nc`** (private) as the NC-frame *pivot* used by downstream genome→CDS/protein derivation, which cdot computes in chromosome coordinates — the pivot must stay NC or it would mis-map back through cdot.
- **Placement ingestion** in `MultiFastaProvider`: RefSeqGene→genome alignments (`GCF_*_refseqgene_alignments.gff3`, new `ferro prepare` download + `refseqgene_alignments` manifest field) parsed once at load into per-`NG_` placements; LRG XML genomic mapping parsed on demand into per-`LRG_` placements. Both keep only single contiguous (ungapped) GRCh38-primary spans — a gapped or non-primary alignment declines rather than mis-anchor.
- **`ReferenceProvider::genomic_placement()`** hook (default `None`); `MockProvider` gains `add_genomic_placement` for tests.
- Bare-`LRG_<n>t<m>`/`p<m>` inputs derive their `LRG_<n>` genomic parent structurally (no `genomic_context` needed).

### No-regression by construction

When no placement is available, the chromosome coordinate is kept under the parent accession — exactly prior behavior. Confirmed against the live corpus (below): every remaining FAIL is a strict subset of the prior baseline, **0 regressions**.

## Verified result (live `FERRO_MANIFEST` genomic axis)

A full manifest-backed run of the genomic axis (rebased on `origin/main`, manifest carrying `refseqgene_alignments` + LRG XMLs):

```
46 pass / 83 FAIL / 191 skipped
set-diff vs committed baseline: 16 improvements, 0 regressions
```

So this PR **demotes 16 genomic rows** (`baseline-failures/genomic.txt` 99 → 83). The demoted set is the re-anchoring target — NG-context `NM_` inputs, a bare `NG_:g.`, and `LRG_1t1`.

### Residual 83 (out of scope here, tracked separately)

| Bucket | Count | Nature |
|---|---:|---|
| `NG_(NM_…)` — NG context, coordinate/normalize divergence | 43 | per-row arbitration |
| `NG_(GENE_vNNN):c.…` — legacy gene-model selector | 27 | new feature: resolve `(GENE_vNNN)` → annotated `NM_` (cdot has no `NG_` entry); #121-adjacent — carved to #637 |
| `pter/qter` | 5 | unimplemented syntax |
| `NG_` bare / `LRG_ t/p` / cross-ref `delins[NM_…]` | 8 | mixed |

A spike confirmed the 27 `NG_(GENE_vNNN)` rows are **not** a parent-resolution tweak: treating the bare `NG_` as its own parent just moves the failure from "no parent reference" to "Reference not found: `NG_…`", because the c. frame is defined by the `(GENE_vNNN)` selector — which cdot cannot resolve to an `NM_`. That is legacy gene-model resolution, a distinct feature.

## Tests

- `tests/issue_480_ng_lrg_parent_frame.rs` (MockProvider, CI-runnable): `nc_to_parent` plus/minus/out-of-span, plus/minus-strand projection (edit reverse-complement), no-placement fallback, bare-LRG re-anchor.
- LRG-XML and RefSeqGene-GFF3 parser unit tests: assembly selection, gapped→`None`, GRCh37/gapped exclusion, minus-strand affine.
- `cargo clippy --features dev --all-targets -- -D warnings` clean.

Refs #480, #326.


nh13:
@coderabbitai review

PR #628, opened 2026-06-13
fix(project): decline protein prediction when the CDS frame is inconsistent (#625)

### LINKED ISSUE #625: protein prediction emits a bogus consequence when the CDS frame is inconsistent (cds_start not at ATG)
## Summary

When a transcript's cdot CDS coordinates are inconsistent with the transcript FASTA ferro paired them with — `cds_start` does **not** land on a start codon (ATG) — ferro silently translates the wrong frame and emits a bogus protein consequence (e.g. `p.(Xaa…)` or a wrong-position anchor) instead of detecting the inconsistency and declining.

Discovered while triaging the hgvs-rs projection corpus against a real manifest.

## Reproducers (manifest required)

A CDS-frame probe over ferro's own loaded reference data shows two transcripts whose `cds_start` does not point at an ATG, producing a mistranslated reference protein with a premature internal stop:

| transcript | ferro `cds_start` (1-based) | first CDS codon | first internal stop |
|---|---|---|---|
| `NM_000425.3` (L1CAM) | 179 | **AGG** (not ATG) | codon **8** (true protein is 1257 aa) |
| `NM_152263.2` (TPM3) | 118 | **GAT** (not ATG) | codon **26** (true protein is 285 aa) |

Because the reference protein mistranslates to a premature stop, any protein-consequence prediction on these transcripts is garbage. Example: `NM_000425.3:c.3772dupT` should be `p.Ter1258Leuext*96` but ferro anchors on the spurious codon-8 stop.

Both files are NCBI RefSeq (cdot from the GFF annotation; FASTA from RefSeq mRNA), so this is a cdot↔FASTA version/consistency mismatch in the prepared reference set — but ferro should not emit a confident wrong answer from inconsistent inputs.

## Expected behavior

ferro should validate that a transcript's CDS yields a sane reading frame before predicting a protein consequence, and **decline gracefully** when it does not:

- The first CDS codon should be a start codon (ATG). If `seq[cds_start-1 .. cds_start+2]` is not `ATG`, the cds_start is inconsistent with the sequence.
- On inconsistency, emit `p.?` (consequence unknown) or decline the protein axis (`Ok(None)`), matching ferro's existing "cannot predict" conventions — never a fabricated consequence.

**Important — do not over-trigger on selenoproteins.** The discriminator must be the **start codon**, NOT "no internal stop": selenoproteins (e.g. SELENON / `NM_020451.2`) legitimately contain an internal in-frame `TGA` recoded as selenocysteine. Those start with ATG and are valid; only a non-ATG (or otherwise frame-broken) start indicates the cds_start/sequence inconsistency this issue is about.

## Acceptance

- A transcript whose `cds_start` does not point at ATG yields `p.?` / a declined protein axis, not `Xaa`/a wrong-anchor consequence.
- Selenoproteins and normal ATG-start transcripts are unaffected.
- The two reproducers above no longer emit a fabricated consequence.

nh13:
Fixes #625. When a transcript's cdot CDS coordinates are inconsistent with the transcript FASTA — `cds_start` doesn't land on an ATG — ferro silently translated the wrong frame and emitted a fabricated protein consequence (`Xaa…` / a wrong-anchor extension) from a reference protein that mistranslates to a premature internal stop.

**Fix:** a `cds_has_valid_start` guard (`ref_cds` first codon == `ATG`) in the shared `predict_protein_consequence` (covers both the genome-pivot and direct bare-`NM_` paths). When the reference CDS doesn't start with ATG, ferro declines protein prediction (`Ok(None)`) — its existing "cannot predict" convention — instead of emitting garbage.

**Selenoprotein-safe:** the guard keys on the *start codon only*, never on internal stops — selenoproteins (e.g. SELENON `NM_020451.2`) legitimately carry an in-frame internal `TGA` recoded as selenocysteine and must not be declined. A dedicated test (`project_selenoprotein_internal_tga_predicts_protein`) pins this.

**Effect (against a real manifest):** `NM_000425.3` and `NM_152263.2` (cds_start not at ATG) now decline instead of emitting a bogus consequence; valid ATG-start transcripts — including `NM_000051.3`/`NM_000488.3` fixed by #621 — are unchanged.

Discovered via a CDS-frame probe over ferro's own loaded reference data while triaging the hgvs-rs projection corpus. The underlying data issue is a cdot↔FASTA version/consistency mismatch in the prepared reference set; this PR makes ferro robust to it (decline rather than fabricate). TDD: 3 hermetic tests (non-ATG → declines; ATG control → predicts; selenoprotein control → predicts), red-first. Full suite 6899 passed, 0 failed; clippy/fmt clean. Scope: `src/project/` only.

### DIFF
--- a/src/project/projector.rs
+++ b/src/project/projector.rs
@@ -14,8 +14,8 @@ use crate::normalize::{NormalizeConfig, Normalizer};
 use crate::project::accession::parse_accession;
 use crate::project::edit::transform_edit_for_strand;
 use crate::project::protein::{
-    affects_initiation_codon, build_initiator_unknown, predict_indel_protein,
-    predict_substitution_protein, RefProteinBundle,
+    affects_initiation_codon, build_initiator_unknown, cds_has_valid_start, predict_indel_protein,
+    predict_substitution_protein, read_cds_start_codon, RefProteinBundle,
 };
 use crate::project::result::VariantProjection;
 use crate::reference::transcript::Transcript;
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #687, opened 2026-06-15
perf(normalize): parallelize the batch CLI loop behind -j/--workers

nh13:
Batch `ferro normalize` processed input one line at a time in a serial loop, even though batch normalization is embarrassingly parallel and a rayon-based library API (`src/parallel.rs`) plus a reserved `-j/--workers` flag already existed. This wires them together — by far the largest throughput lever available for batch workloads.

## What changed

The batch path (file or stdin) now reads input in fixed-size chunks and, when `--workers > 1`, maps each line through a dedicated rayon pool, then writes results **in input order**. Each line is formatted into its own per-line stdout/stderr `Vec<u8>` buffers using the *exact same* code as the serial path, so parallel output is byte-identical to serial by construction.

- `workers` defaults to **1** (serial — unchanged behavior). The single-variant path is untouched.
- `rayon` becomes an unconditional dependency so `-j` works in the default binary; the `parallel` feature is retained as a no-op alias.

## Correctness — `-jN` output must equal `-j1`

| Check | Result |
|---|---|
| Mock mode, j1 vs j4 | byte-identical |
| Real reference, silent, j1 vs j8 (full 333k ClinVar) | byte-identical, 332,735 lines |
| Lenient mode, **stdout and stderr** j1 vs j8 | byte-identical (exact order) |
| Determinism (j8 run twice) | identical |

Plus a CI integration test (`tests/cli_parallel_normalize.rs`, mock provider, input spanning >2 chunks) asserting `-j{2,4,8}` == `-j1` and run-to-run determinism. clippy `-D warnings` and fmt clean.

## Scaling (1.3M-variant corpus, `--timing` loop-only)

| Workers | var/s | vs j1 |
|---|---|---|
| 1 | 110k | 1.0x |
| 2 | 146k | 1.3x |
| 4 | 405k | 3.7x |
| 6 | 554k | 5.0x |
| 8 | 258k | 2.4x |
| 12 | 216k | 2.0x |

~5x at `-j6`. **Caveat:** these were measured on a development machine under heavy load (load avg 40–56 on 12 cores); the fall-off at `-j8`/`-j12` is oversubscription against an already-saturated box, not a scaling ceiling. A quiet-host re-measure (≥5 reps, median ± IQR) should be run before quoting precise multipliers — the direction (multi-x, near-linear until cores saturate) is robust.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,10 @@ lru = "0.12.5"
 [features]
 default = []
 dhat-heap = ["dep:dhat"]
-parallel = ["dep:rayon"]
+# rayon is now an unconditional dependency (the `ferro normalize -j` batch
+# driver uses it); this feature is retained as a no-op alias and to gate the
+# library `parallel` module's public API.
+parallel = []
 validation = ["dep:ahash"]
 python = ["dep:pyo3"]
 # memmap2 is now an unconditional dependency (used by the cdot rkyv loader);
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

PR #696, opened 2026-06-15
perf(annotate-vcf): parallelize file-input record annotation behind -j/--workers

nh13:
Adds a `-j/--workers` flag to `ferro annotate-vcf` and parallelizes the per-record annotation for file input. Each record is annotated independently, so this is the natural follow-up to the output-buffering fix.

> **Stacked on #695** (the buffering fix) — base branch is `perf/annotate-vcf-buffered`. Review/merge that first; this rebases onto `main` afterward. The diff here is the `-j` flag + parallel loop only.

## What changed

- The file-input path annotates fixed-size (4096) chunks across a rayon pool when `--workers > 1`, writing results **in input order** — byte-identical to serial. The first per-record annotation error is propagated as before.
- `workers` defaults to **1** (serial, unchanged). Parallelism applies to **file input only**; stdin (with its raw-line pass-through and header-state semantics) stays serial.
- `rayon` becomes an unconditional dependency so `-j` works in the default binary (same change as the `normalize`/`parse` parallelism PRs; trivially de-dups if those merge first).

## Why this, and why after buffering

Profiling drove the order. The first annotate profile (1M records, real RefSeq GRCh38 GFF → 199,979 transcripts) was **35% kernel** from unbuffered writes — fixed in #695 (→20%). *After* buffering, the workload is CPU/allocation-bound (ferro ~44%, malloc+memcpy ~36%), which is what makes record-level parallelism worthwhile.

## Correctness & scaling

- `tests/cli_parallel_annotate.rs` (committed micro chr21 GFF, input spanning >2 chunks) asserts `-j{2,4,8}` output equals `-j1` and is deterministic. Verified byte-identical on a 250k-record real-GFF annotation; all existing VCF-annotate tests pass; clippy `-D warnings` + fmt clean.
- 1M-record VCF, real GFF, total wall **including the fixed ~one-time GFF load**:

  | workers | wall | vs -j1 |
  |---|---|---|
  | 1 | 46.7s | 1.0x |
  | 2 | 26.4s | 1.8x |
  | 4 | 14.7s | 3.2x |
  | 8 | 20.4s | 2.3x* |

  *the -j8 dip is oversubscription on a heavily-loaded dev box (load avg ~26-41); the annotation work itself scales further. Numbers are directional — a quiet-host re-measure is worth doing before quoting precise multipliers.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,9 +65,9 @@ lru = "0.12.5"
 [features]
 default = []
 dhat-heap = ["dep:dhat"]
-# rayon is now an unconditional dependency (the `ferro normalize -j` batch
-# driver uses it); this feature is retained as a no-op alias and to gate the
-# library `parallel` module's public API.
+# rayon is now an unconditional dependency (the CLI `-j` batch drivers use it);
+# this feature is retained as a no-op alias and to gate the library `parallel`
+# module's public API.
 parallel = []
 validation = ["dep:ahash"]
 # The Python batch API parallelizes across a rayon pool, so the `parallel`
--- a/src/bin/ferro.rs
+++ b/src/bin/ferro.rs
@@ -77,6 +77,11 @@ enum Commands {
         /// Include all transcripts (default: true)
         #[arg(long, default_value = "true")]
         all_transcripts: bool,
+
+        /// Number of parallel workers for record annotation (default: 1).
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

Issue #487, opened 2026-05-29
fix(normalize): canonical edit-form detection (ins→dup, tandem repeat, mito m., allele collapse)


nh13:
Child of #326. Spec arbitration verdict: **(a) ferro normalization bugs** (a chunk of the 105 `normalized` FAILs).

## Problem

ferro fails to canonicalize several edit forms the HGVS spec mandates. Examples (input → expected → ferro):

- **`ins` not collapsed to `dup`** when the insert duplicates preceding sequence — `LRG_303:g.6932_6933insAGCAACGTGATCGCCTCCCTCACCT` → `g.6908_6932dup` (`docs/recommendations/DNA/duplication.md`).
- **tandem repeats not detected** — `LRG_303:g.4_5insGT[5]` → `g.1_4GT[7]`; `c.6delinsCCCC` → `c.6C[4]`; and the reverse error, a repeat emitted as `ins` — `c.6C[4]` → ferro `c.5_6insCCC` (`docs/recommendations/DNA/repeated.md`).
- **mitochondrial accession emitted as `g.` not `m.`** — `NC_012920.1:g.3243A>G` → `NC_012920.1:m.3243A>G` (NC_012920.1 is the mtDNA reference; HGVS requires `m.`).
- **allele not collapsed** — `NG_012337.1:g.[104_105insA;105_106insC;105del]` → `g.105delinsAC`.

Also in this family but lower-confidence (need reference-sequence confirmation of who is spec-correct): **3′-shift off-by-one** cases, e.g. `NG_012337.1(NM_012459.2):c.12_13delCA` → expected `c.12_13del`, ferro `c.13_14del`; several `ENST…:c.100del` → `c.102del`.

## Ask

Add the missing canonical-form detection during normalization (dup-vs-ins, tandem repeat, allele collapse), coerce mito accessions to `m.`, and audit the 3′-shift off-by-ones against the reference. One PR per coherent group; demote `baseline-failures/normalized.txt` rows as each lands.

Refs #326.

nh13:
Status update: the **mito `m.` group is done** via #541 (`g.` on a mitochondrial accession now coerces to `m.`; `NC_012920.1:g.3243A>G` is gone from `baseline-failures/normalized.txt`).

Three groups remain outstanding (all still failing in `baseline-failures/normalized.txt`):

- **ins→dup collapse** (e.g. `LRG_303:g.6932_6933ins…`)
[...]

PR #700, opened 2026-06-15
feat(normalize): apply the 3′ rule across exon/intron boundaries (#670)

### LINKED ISSUE #670: fix(normalize): apply the 3′ rule across exon/intron boundaries for genomic-context c./n. variants
## Summary

For coding/non-coding variants described against a **genomic-context** reference (an `NG_`/`NC_`/`LRG_` parent, e.g. `NG_008835.1(NM_022153.2):c.82del`), ferro does **not** apply the HGVS 3′ rule across exon/intron and intron/exon junctions. It confines 3′-shifting to the exon, so a deletion/duplication whose 3′-most representation lies in the adjacent intron is left in the exon.

This is the remaining unresolved group of #487 (the "3′-shift off-by-one" cases). It is split out into its own issue because the fix depends on the c↔g projection keystone (see Dependency).

## Spec basis (ferro is wrong here)

The HGVS spec is explicit that the 3′ rule **does** cross exon/intron and intron/exon junctions; the *only* suppression is at exon/exon junctions.

- `background/numbering.md:22-26` — **exception 3′ rule**: the 3′ rule is not applied around **exon/exon** junctions (shifting would land in the wrong exon, e.g. `LRG_199t1:c.3921del` and *not* `c.3922del`). The closing **NOTE** states this exception "**does not apply** to a deletion/duplication around exon/intron and intron/exon junctions."
- `recommendations/DNA/deletion.md:51-64`:
  - **exon/intron** border: `...GAACAG`**g**`t..//..agTGCCTT` → shifts into the intron, described "**not as `c.1704del`**".
  - **intron/exon** border: `CTGGCCgt..//..ag`**G**`TTTTA` → shifts into the exon, described "**not as `c.1813-1del`**".

So at an exon/intron boundary with identical flanking nucleotides, the variant must shift into the intronic (`+`/`-` offset) coordinate.

## Root cause

`normalize_cds` shuffles in **spliced (transcript) coordinate space**, exon-confined via `get_cds_boundaries_with_axis_info` (`src/normalize/boundary.rs`). The `cross_boundaries` flag (`src/normalize/config.rs`) is hardcoded `false` in the production path (only enabled in tests + a single-exon special case). Even when `true`, the bound becomes the full **spliced** transcript `(1, sequence_length())`, which contains **no intronic bases at all** — so spliced-space shuffling structurally cannot produce an intronic `c.N+1` result.

The correct behavior is to 3′-shift in **genomic coordinate space** (which contains the introns) for genomic-context variants and map the result back to `c.`/`n.` with `+`/`-` offsets. Shifting on genomic coordinates also satisfies the exon/exon exception for free: consecutive exons are separated by the intron on the genomic sequence, so a 3′ shift moves into the intron, never into the next exon.

## Affected conformance rows (axis = normalized)

From the ferro-prepared manifest run (`cargo nextest run --features dev -E 'test(axis_normalized)'`):

| input | ferro emits | mutalyzer / spec-correct |
|---|---|---|
| `NG_008835.1(NM_022153.2):c.82del` | `c.82del` | `c.82+1del` (exon→intron) |
| `NG_008835.1(NM_022153.2):c.677-21_704+62del` | (unshifted) | `c.677-18_704+65del` (intronic span) |
| `NG_007107.2(NM_004992.3):c.378-17delT` | `c.378-16del` | `c.378-17del` (intronic) |
| `NG_009497.1(NM_206933.2):c.8682-19dupT` | `c.8682-18dup` | `c.8682-19dup` (intronic) |

(`NG_007107.2`/`NG_009497.1` are also version-mismatched against the local manifest — local has `NG_007107.3`/`NG_009497.2` — but they are boundary cases regardless, so the version is not the blocker.)

## Dependency

The fix requires a c↔g projection/shuffle path inside normalization. That machinery is being built by the projection keystone: PRs #642, #647, #616 (issues #498, #646, #480). This issue should be implemented **after** those land, reusing their c↔g conversion rather than building a parallel path.

## Tracking in the conformance harness

The affected rows above are annotated `known_bug` in `tests/fixtures/mutalyzer-normalize/cases.json` with `tracking_issue` pointing at this issue. The harness XPASS guard will fail CI the moment ferro begins emitting the spec-correct form, forcing the annotation to be removed on convergence.

## Acceptance criteria

- [ ] For genomic-context `c.`/`n.` variants, the 3′ rule shifts across exon/intron and intron/exon junctions (into/out of `+`/`-` offset coordinates).
- [ ] The exon/exon exception (`numbering.md:22-26`) is preserved — no shifting into the next exon.
- [ ] The four rows above converge (XPASS → annotations removed).
- [ ] TDD with real reference bytes (the manifest resolves `NG_008835.1`).

Refs #487

nh13:
Closes #670. Refs #487.

**Stacked on #690** (base branch `fix/intronic-genomic-fetch-off-by-one`) — review/merge #690 first. The two fixes here rely on #690's off-by-one correction.

ferro's genomic-space shuffle was reachable only when an endpoint was already intronic, and it assumed a single enclosing intron. Two spec-required cases (numbering.md "exception 3′ rule" NOTE; deletion.md exon/intron border) therefore never shifted:

1. **Purely-exonic, junction-adjacent indels** — a del/dup resting at an exon's 3′ edge stayed exon-confined instead of continuing into the following intron (`NM_022153.2:c.82del` → `c.82+1del`). `normalize_cds` now, after the exon-confined shuffle and only in the 3′ direction, re-runs the shuffle in genomic space when the result landed exactly at the exon boundary with genomic context and a downstream intron — adopting it only if it crossed into the intron. Exon/EXON suppression is preserved structurally (the shuffle boundary is capped at the adjacent intron's far edge, never the next exon); 5′/VCF shuffles stay exon-confined.
2. **Deletions spanning multiple introns** (intron→exon→intron) — both endpoints intronic but in *different* introns were routed to the single-intron `normalize_intronic_cds` and never shifted (`NM_022153.2:c.677-21_704+62del` → `c.677-18_704+65del`). The dispatch now checks `same_intron`; multi-intron spans go to the genomic boundary path.

Both rely on the boundary path sizing its fetch window to cover the relevant exon(s)+intron(s) genomic extent (`intronic_window_bounds`, generalized via `get_boundary_spanning_genomic_extent`), so a long adjacent intron no longer pushes the shuffle boundary outside the fetched window.

**Tests:** purely-exonic plus-strand + minus-strand crossing, exon/EXON suppression guard, 5′-direction exon-confinement, and a synthetic multi-intron-span case (all in-crate). **Conformance** (mutalyzer normalized axis, reference-gated): 135→137 pass / 83→81 FAIL on top of #690 — adds `c.82del` and `c.677-21_704+62del`; 0 regressions, all other axes unchanged. Full non-conformance suite green (only the pre-existing `axis_protein_description` failure); clippy + fmt clean.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -2097,11 +2097,19 @@ impl<P: ReferenceProvider> Normalizer<P> {
             // gets the build-correct chromosome. If the accession-aware lookup
             // fails, fall back to the plain transcript we already fetched.
             let transcript = transcript_for_intronic().unwrap_or(transcript);
-            // Check if both positions are intronic and in the same intron
-            if start_pos.is_intronic() && end_pos.is_intronic() {
+            // Both intronic *in the same intron* → the dedicated intronic
+            // shuffle (which assumes a single enclosing intron). Both intronic
+            // in DIFFERENT introns (#670: a deletion spanning intron→exon→intron)
+            // falls through to the genomic boundary path, which sizes its window
+            // to the full span — `normalize_intronic_cds` would bound the shuffle
+            // to only `start`'s intron and so never shift such a span.
+            if start_pos.is_intronic()
+                && end_pos.is_intronic()
+                && self.same_intron(&transcript, start_pos, end_pos)
+            {
                 return self.normalize_intronic_cds(variant, &transcript, start_pos, end_pos, edit);
             }
-            // Variant spans exon-intron boundary - normalize in genomic space
+            // One endpoint exon/intron boundary, or a multi-intron span: genomic space.
             return self.normalize_boundary_spanning_cds(
                 variant,
                 &transcript,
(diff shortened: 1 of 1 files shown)


nh13:
Note: CI (`ci.yml`) only triggers on `pull_request: branches: [main]`, so the checks won't run while this is stacked on the #690 branch. They'll run once #690 merges and this PR is retargeted to `main`. Verified locally in the meantime: full non-conformance suite green (only the pre-existing manifest-gated `hgvs_rs_projection_tests::axis_protein_description` failure, which also fails on `main` locally and skips on CI), clippy + fmt clean, and the conformance delta above (+2 normalized rows, 0 regressions).

Issue #692, opened 2026-06-15
fix(parser): accept a cross-reference insert in delins (c.<pos>delins<ACC>:c.<a>_<b>)


nh13:
`NG_012337.3(NM_003002.4):c.274delinsNG_012337.1(NM_012459.2):c.200_203` fails to parse:

```
Parse error at position 38: Unexpected trailing characters: '_012337.1(NM_012459.2):c.200_203'
```

The parser accepts a cross-reference *insertion* (`ins<ACC>:c.<a>_<b>`, added in #395, now closed) but not the same cross-reference inside a `delins` payload. Both corpus rows of this shape fail at the `delins<ACC>:` boundary:

- `NG_012337.3(NM_003002.4):c.274delinsNG_012337.1(NM_012459.2):c.200_203`
- `NG_012337.1(NM_012459.2):c.274delinsNG_012337.3(NM_003002.4):c.200_203`

This blocks 2 rows of the `rna_description` conformance axis (#485) — the predicted RNA form (`r.(274delinscuga)` / `r.(274delinscuag)`) can't be produced because the input never parses.

Refs #326, #485.

Issue #672, opened 2026-06-14
test(conformance): provision exact corpus accession versions absent from the prepared manifest


nh13:
## Summary

Several mutalyzer-normalize corpus rows pin **older accession versions** that the `ferro prepare` manifest does not carry, so ferro cannot resolve the reference and no-ops (echoes the input). These present as conformance FAILs but are **not ferro bugs** — given the correct reference, ferro produces the spec-correct result. They need the exact-version bytes provisioned for hermetic TDD.

## Proven example (not a bug)

`NM_003002.2:c.273del` (corpus expects `c.274del`):
- Manifest carries only `NM_003002.4` (transcript FASTA) / `NM_003002.3`+`.4` (cdot); `.2` is absent → ferro no-ops, echoing `c.273del`.
- The `.2` and `.4` **CDS are byte-identical** (480 bp; sequence at c.273-274 is `…AT`**GG**`AC…`). Running ferro on the present version proves the logic is correct:
  - `ferro normalize 'NM_003002.4:c.273del' --reference …` → `NM_003002.4:c.274del` ✅ (3′ rule rolls the `GG` correctly)
  - `ferro normalize 'NM_003002.2:c.273del' --reference …` → `NM_003002.2:c.273del` (no-op; version absent)

So the only thing missing is the `.2` reference sequence.

## Also needed (to finalize #670 adjudication)

The two intronic 3′-shift rows in #670 are version-independent in their c. result (ferro gives the same answer on local `.3`/`.2` as on the pinned `.2`/`.1`), but their **exact genomic position** can only be adjudicated against the pinned version's intron sequence (the `.2`→`.3` / `.1`→`.2` genomic coordinates do not line up):

- `NG_007107.2(NM_004992.3):c.378-17delT` (local has `NG_007107.3`)
- `NG_009497.1(NM_206933.2):c.8682-19dupT` (local has `NG_009497.2`)

## Proposed provisioning

NCBI efetch is already wired in `ferro prepare` (`src/prepare/mod.rs:1306,1455`). Fetch the exact pinned accession.versions into a supplemental FASTA and record them in the manifest:
- `NM_003002.2` (transcript)
- `NG_007107.2`, `NG_009497.1` (RefSeqGene genomic)

(Other corpus rows pin `NG_012337.1`, `NG_012772.1`, `NG_009497.1`, `NG_007107.2`, etc.; sweep the corpus for all version mismatches while provisioning.)

## Harness wiring

[...]

Issue #654, opened 2026-06-14
test(conformance): normalized axis baseline is stale; regenerate + arbitrate 4 drifted rows


nh13:
## Summary

A `FERRO_MANIFEST`-backed run of `axis_normalized` against `origin/main` fails: the committed `tests/fixtures/mutalyzer-normalize/baseline-failures/normalized.txt` no longer matches current behavior. The committed baseline is **stale** — it predates normalization changes that have since merged.

## Findings (live run)

- **6 rows now pass** that are still listed as expected-fail (demotable).
- **4 rows now fail** that are not in the baseline (not regressions vs the *code*, but unrecorded current divergences):
  - ~~`NM_003002.2:c.273del` — ferro `c.273del`, expected `c.274del` (a 3′-shift off-by-one).~~ **Resolved by #672 — not a 3′-shift bug.** This is a reference-availability artifact: the manifest carries only `NM_003002.4`/`.3`, not the pinned `.2`, so ferro no-ops and echoes the input. On the present version `ferro normalize 'NM_003002.4:c.273del'` → `c.274del` correctly (the `.2`/`.4` CDS are byte-identical at this locus). Provisioning the exact `.2` bytes (#672) makes this XPASS. Defer this row to #672.
  - `NM_003002.2:c.pterdel` / `c.qterdel` and `NG_012337.1:g.pterdel` — pter/qter handling (overlaps #537).

## Proposal

Regenerate the `normalized` baseline from current behavior and **arbitrate the remaining 3 drifted rows** (HGVS-spec each: ferro bug vs accepted divergence). The `c.273del` row is now accounted for by #672 (reference-availability, not a code bug). The 3 pter/qter rows overlap #537 (pter/qter g-axis).

**Note (2026-06-15):** the committed `normalized.txt` baseline has since dropped 105 → 71 rows (see #326), so the specific "6 demotable / 4 drifted" deltas from the original live run predate that churn — re-run before regenerating.

This also relates to the harness gap in the output-quality tracking issue (a stale baseline is exactly what membership-only diffing lets drift).

Refs #326, #487, #537.

nh13:
[...]

PR #756, opened 2026-06-18
fix(normalize): resolve the transcript for a genomic-context c. cross-reference payload (#654)

nh13:
**Stacked on #755** (base = `fix/issue-654-normalized-ok-mismatch`). Review the stack #754 → #755 → this first; this PR's diff is only the two commits on top.

## Summary

Cross-reference payload resolution fixes in the normalized-axis `Err`-tail (#654). Two related gaps in how `delins`/`ins` payloads that reference another accession's coordinates are resolved:

**1. Genomic-context compound payload accession.** `delins NG_(NM_):c.A_B` failed with `transcript NG_…(NM_…) has no CDS start; cannot expand c. ins[A_B] payload` — `resolve_cross_reference_bases` passed the whole compound accession to the CDS translation, which looks up `cds_start` by accession. Now reduces a `c.` payload accession to its bare transcript (`NM_`), dropping the `NG_(…)`/`LRG_(…)` wrapper, mirroring the same-accession expansion path. `Direct` (g./m./o./n.) payloads are unchanged.

**2. Single-position payload.** A payload may name a single position (a one-base copy) rather than a range, e.g. `delins NG_(NM_):c.200`. `parse_cross_reference` required an `A_B` range and a five-char minimum, so a single position was rejected as out-of-scope. Now treats a payload with no `_` as `start == end`.

Both fixes were written test-first (unit tests in `normalize::rules::tests`).

## Result

Greens three mutalyzer-normalize corpus rows (`Err` → pass):

| Input | Output |
|---|---|
| `NG_012337.1(NM_012459.2):c.274delinsNG_012337.3(NM_003002.4):c.200_203` | `…:c.274delinsCTAG` |
| `NG_012337.3(NM_003002.4):c.274delinsNG_012337.1(NM_012459.2):c.200_203` | `…:c.274delinsCTGA` |
| `NG_012337.3:g.136delinsNG_012337.1(NM_012459.2):c.200` | `…:g.136A>C` (one-base copy → 1-base delins → substitution) |

normalized axis: **11 FAIL (was 14), 0 XPASS**. Regenerated the baseline snapshot; re-blessed the `mock-pin` (MockProvider lacks those transcripts, so the not-found error now names the resolved transcript). Full hermetic suite green (**7351 passed**); `clippy` + `fmt` clean.

## Remaining Err-tail (follow-up)

The remaining `Err` rows need either policy decisions or larger features: cross-reference payloads decorated with `pter`/`qter`/`*N`/offsets (which ferro deliberately refuses for `pter`/`qter`), an intronic `c.*824+10del` (version-adjacent, #672), and inputs ferro arguably rejects correctly (`c.41>CA`, malformed concatenated forms). The latter can't be annotated away on a projection axis by design.

Refs #654, #692.


nh13:
Addressed in `2534846`.

**Major (n. payloads):** the genomic-context accession reduction was gated on `InsCoordKind::Cds`, so it missed `n.` (transcript-relative but classified `Direct`). Since `InsCoordKind` collapses `g./m./o./n.` into `Direct`, I thread a `transcript_relative` flag out of `parse_cross_reference` (true for `c.` and `n.`) and reduce the compound `NG_(NM_)`/`LRG_(…)` accession to its transcript for both, leaving `g./m./o.` (genomic/absolute) on the named accession. `NG_007485.1(NR_003529.3):n.200`-style payloads now resolve.

**Nitpick:** added a Direct-axis single-position test (`NM_000088.3:n.2`), alongside the `n.` compound-accession regression test.

Full hermetic suite green; no manifest-axis change (no `n.` cross-reference row in the current corpus — this is a correctness fix for the class). #757 rebased on top.

PR #761, opened 2026-06-18
feat(project): predict protein consequence for whole-exon deletions

nh13:
## Summary

A deletion that removes one or more complete coding exons now predicts its protein consequence (frameshift or in-frame deletion) instead of declining. This is spec-compliant — `protein/deletion.md` "one or more exons" gives confident predictions for whole-exon deletions — and addresses part of #498's protein-axis prediction gap. `Refs #498` (not `Closes`): the issue has other open threads.

## Approach

A new pure helper `whole_exon_deletion_span` (`src/project/protein/helpers.rs`) clamps a both-intronic deletion to the exonic CDS bases it removes. `predict_protein_consequence` is relaxed so a whole-exon deletion reaches a new guarded match arm that re-expresses the clamped span as a clean exonic deletion and runs it through the existing `predict_indel_protein` — reusing the verified frameshift/in-frame machinery. The change lives in the single gate shared by all three projection entry points (genome-pivot, bare-`NM_` direct, bare-`n.`/`r.` direct).

### Conservative by design — only the canonical form predicts

It predicts only the well-formed canonical bracketing: ascending interval, 5′ offset `< 0`, 3′ offset `> 0`, within the CDS. Every other shape declines, so the change never emits a wrong protein from ambiguous positions:

- **Reverse-strand crossed/descending intervals.** ferro's reverse-strand coding normalization (#762) hands the gate a mis-rendered interval — e.g. `c.704-18_677+65del` for VSIR, which literally describes removing the wrong exonic span. Predicting from it would emit a wrong protein (`p.(Ala227GlnfsTer6)` instead of `p.(Arg226ProfsTer102)`), so it declines. The case predicts correctly with no further change here once #762 normalizes the endpoints canonically.
- **3′UTR-intronic, exonic-endpoint (partial-exon / splice), and pure/deep-intronic spans** all decline (partial-exon matches mutalyzer's own `test_del_splice_site` rule).

## Validation

- **DMD exon 10 (in-frame):** `NM_004006.2:c.961-1_1149+3del` → `p.(His321_Glu383del)` — matches the HGVS spec example exactly.
- **DMD exons 10–11 (frameshift):** `NM_004006.2:c.961-1_1331+1del` → `p.(His321PhefsTer3)` — anchor (`His321`) and stop distance (`fsTer3`) match the spec's exons-10–11 example; the substituted residue differs (Phe vs Leu) because ferro translates `NP_003997.1` and the spec example is `.2`.
- **Reverse-strand VSIR:** declines safely, guarded by a manifest test, until #762.
- Helper unit tests cover every offset-sign / UTR / crossed-form case; projector integration tests cover predict-and-decline via `project_variant`. Manifest-gated real-world tests skip in manifest-less CI. Full suite: 7369 passing.

## Scope / not included

- VSIR corpus rows stay FAIL as a **safe decline** (not wrong output), gated on #762.
- The protein-axis empty-projection budget is **not** re-blessed (a separate reference-drift concern: the committed baseline of 19 does not reproduce against the current prepared reference).
- `coding_protein_descriptions` cluster-B (transcript-set/version selection) and the other #498 threads are untouched.

## Summary by CodeRabbit

* **New Features**
  * Protein consequence predictions now support whole-exon deletions with intronic endpoints.

* **Tests**
  * Added integration tests for whole-exon deletion prediction scenarios.

### DIFF
--- a/src/project/projector.rs
+++ b/src/project/projector.rs
@@ -15,7 +15,7 @@ use crate::project::accession::parse_accession;
 use crate::project::edit::transform_edit_for_strand;
 use crate::project::protein::{
     affects_initiation_codon, build_initiator_unknown, cds_has_valid_start, predict_indel_protein,
-    predict_substitution_protein, read_cds_start_codon, RefProteinBundle,
+    predict_substitution_protein, read_cds_start_codon, whole_exon_deletion_span, RefProteinBundle,
 };
 use crate::project::result::VariantProjection;
 use crate::reference::transcript::Transcript;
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

Issue #637, opened 2026-06-14
feat(project): resolve legacy (GENE_vNNN) gene-model selectors on NG_ references (genomic axis, 27 rows)


nh13:
## Summary

A genomic-axis input of the form `NG_<n>.<v>(GENE_vNNN):c.…` names a **legacy gene-model selector** (`GENE_vNNN`, e.g. `BRCA2_v001`) as the transcript whose frame the `c.` coordinate is expressed in. `project_to_genomic` cannot resolve these: the c. coordinate frame is defined by the `(GENE_vNNN)` selector, but ferro routes transcript resolution through cdot, which is keyed on `NM_`→`NC_` alignments and has no entry for the `NG_` accession (nor for the legacy gene-model id). These are the **27 `NG_(GENE_vNNN)`** rows in `tests/fixtures/mutalyzer-normalize/baseline-failures/genomic.txt` left after #480 / PR #616.

This is **new feature work** (legacy gene-model resolution), distinct from the NC→NG/LRG re-anchoring landed in PR #616, and overlaps the gene-symbol-selector policy concern in #121.

## Proven mechanism (spike under PR #616)

A live `FERRO_MANIFEST` genomic-axis run plus a spike confirmed the blocker is **not** parent resolution:

- The `(GENE_vNNN)` selector is stored as the variant's `gene_symbol`, not in the `Accession`, so `transcript_accession()` returns the bare `NG_<n>.<v>`.
- Treating the bare `NG_` as its own genomic parent (so the parent-resolution step passes) only moves the failure one step downstream — from `unsupported projection: ... no parent reference (genomic_context)` to `Reference not found: NG_<n>.<v>` — because cdot has no `NG_` transcript and the selector is never mapped to an `NM_`.

```
before spike:  err=unsupported projection: ... no parent reference (genomic_context) on NG_012772.1
after spike:   err=Reference not found: NG_012772.1
```

So clearing these requires resolving `(GENE_vNNN)` → the `NM_` annotated on that `NG_` record (then reuse PR #616's NC→NG re-anchoring), or projecting the c. directly against the `NG_`'s own GenBank gene-model exon structure. cdot cannot supply either today.

## Scope: 27 inputs across 8 `NG_` records

```
NG_008835.1(CDH23_v001):c.1449+846delA
NG_008939.1(PCCB_v001):c.156_157ins[180_188]
NG_008939.1(PCCB_v001):c.156_157ins[180_188inv]
[...]

PR #784, opened 2026-06-18
feat(project): resolve legacy GENE_vNNN selectors on NG_(...):c./n./r. projection

### LINKED ISSUE #783: feat(project): resolve legacy GENE_vNNN transcript selectors on NG_(...):c./n./r. inputs (protein/coding axes; #637 follow-on)
### Summary

A coding-coordinate variant that names its transcript with a legacy LOVD `GENE_vNNN` selector inside a genomic context — e.g. `NG_012772.1(BRCA2_v001):c.632_681del` — fails to project on **every** axis with a `transcript_id mismatch` error, even though ferro internally resolves the selector to the correct RefSeq transcript.

#637 resolved `GENE_vNNN` selectors for genomic-*coordinate* inputs (the genomic axis). This is the sibling gap for coding-coordinate (`c.`/`n.`/`r.`) inputs.

### Reproduce

```
ferro project --axis p --reference <ref> 'NG_012772.1(BRCA2_v001):c.632_681del'
#  -> unavailable: unsupported projection: transcript_id mismatch:
#     input is on NM_000059.3 but projection requested against NG_012772.1
ferro project --axis c --reference <ref> 'NG_012772.1(BRCA2_v001):c.632_681del'   # same error
ferro project --axis g --reference <ref> 'NG_012772.1(BRCA2_v001):c.632_681del'   # same error
```

The identical variant with the **RefSeq** transcript name works and gives the expected output:

```
ferro project --axis p --reference <ref> 'NG_012772.1(NM_000059.3):c.632_681del'
#  -> NP_000050.2:p.(Val211GlufsTer10)     <- matches the mutalyzer corpus oracle
```

### Root cause

For `NG_012772.1(BRCA2_v001)` the parser keeps `BRCA2_v001` as a gene-symbol selector, not a transcript, so `Accession::transcript_accession()` returns the outer `NG_012772.1`. `project_variant` interprets the `c.` coordinates by resolving `BRCA2_v001` → `NM_000059.3` (hence "input is on NM_000059.3"), but the requested projection target is the unresolved `NG_012772.1`, so the transcript_id-mismatch guard (`src/project/projector.rs:1996/2186/2592`) rejects it. The projection target needs to be the selector-resolved transcript.

### Impact

16 `protein_description` corpus rows (BRCA2_v001, TIMM8B_v001, PCCB_v001) plus their coding/genomic-axis counterparts. All currently FAIL; each has a clean expected value that ferro already produces for the RefSeq-named form.

### Fix direction

Resolve the `GENE_vNNN` genomic-context selector to its RefSeq transcript when deriving the projection target for `c.`/`n.`/`r.` inputs (reusing the #637/#500 selector-resolution path), so the self-projection target matches the coordinate-resolved transcript. Must not regress the genuine transcript_id-mismatch guard for truly mismatched explicit targets.

Refs #498, #326, #637, #500.

nh13:
## Summary

A coding-coordinate variant whose transcript is a legacy LOVD gene-model selector inside a genomic context — `NG_012772.1(BRCA2_v001):c.632_681del` — failed to project on **every** axis with a `transcript_id mismatch` error, even though ferro resolves the selector to the correct RefSeq transcript. #637 fixed this for genomic-*coordinate* inputs; this is the sibling fix for `c.`/`n.`/`r.` coordinate inputs.

```
# before
ferro project --axis p ... 'NG_012772.1(BRCA2_v001):c.632_681del'
#  -> unavailable: transcript_id mismatch: input is on NM_000059.3 but projection requested against NG_012772.1
# after
#  -> NP_000050.2:p.(Val211GlufsTer10)        # the corpus oracle's consequence
```

## Root cause & fix

Normalization rewrites `NG_(GENE_v001):c.` → `NG_(NM_):c.` (#637/#500), but a self-projection caller (the CLI, the conformance harness) derives the projection target from the *un-resolved* input accession via `transcript_accession()` — which for the selector form is the genomic ref itself (`NG_`), not a transcript (unlike `NG_(NM_)`, whose `transcript_accession()` already returns the inner `NM_`). The post-normalization input (`NM_`) then mismatched that target.

`project_variant` now reconciles the target after normalization: when the requested target names the variant's genomic-context wrapper rather than its transcript, it projects against the variant's own resolved transcript — the caller's self-projection intent. A genuinely unrelated target still errors (the mismatch guard and its test are unchanged).

## Validation

- New unit test `legacy_gene_model_selector_projects_against_self_derived_target` exercises the realistic path (target derived from the raw accession, as the CLI/harness do); the pre-existing `projector_rejects_transcript_id_mismatch_for_cds_input` guard test still passes.
- Conformance (manifest): **9 protein_description FAILs → spec_overridden** (50 → 41 FAIL). BRCA2_v001 (×3) and PCCB_v001 (×6) now produce the exact oracle consequence; the only delta is the bare-NP vs `NM_x(NP_y)` rendering, dispositioned under the existing `bare-np-protein` cluster.
- The remaining LOVD rows stay FAIL on **independent, now-visible** divergences my change surfaces (TIMM8B_v001 protein-version drift #505; PCCB range-reference insertions; BRCA2 uncertain-endpoint declines) — out of scope here.
- `cargo clippy --features dev --all-targets` clean; full manifest-less suite (7408 tests) green.

Closes #783

### DIFF
--- a/src/project/projector.rs
+++ b/src/project/projector.rs
@@ -503,7 +503,40 @@ impl<P: ReferenceProvider + Clone> VariantProjector<P> {
         // construction time so we don't clone the (potentially heavy) provider
         // on every call.
         let normalized = self.normalizer.normalize(variant)?;
-        self.project_variant_inner(&normalized, transcript_id)
+        // 2. Reconcile the requested target against any selector resolution that
+        // normalization performed (#783). A legacy LOVD gene-model selector
+        // `NG_(GENE_v001):c.` keeps the genomic ref as its accession, so a
+        // self-projection caller (the CLI, the conformance harness) derives the
+        // target from `transcript_accession()` and gets the genomic ref, not a
+        // transcript. Normalization resolves the selector to `NG_(NM_):c.` (#637),
+        // so the requested target now names the variant's genomic-context wrapper
+        // rather than its transcript — re-point it at the resolved transcript.
+        let target = Self::reconcile_self_projection_target(&normalized, transcript_id);
+        self.project_variant_inner(&normalized, &target)
+    }
+
+    /// Reconcile a self-projection target with a selector/context that
+    /// normalization may have rewritten (#783).
+    ///
+    /// Returns the variant's own (resolved) transcript accession when the
+    /// requested target names this variant's genomic-context wrapper rather than
+    /// its transcript — i.e. the caller asked to project a transcript-coordinate
+    /// variant "against itself" but derived the target from an accession whose
+    /// transcript slot was a genomic ref (an unresolved legacy gene-model
+    /// selector, now resolved). Otherwise returns the requested target unchanged,
+    /// so a genuine transcript_id mismatch (an unrelated target) still errors.
+    fn reconcile_self_projection_target(normalized: &HgvsVariant, requested: &str) -> String {
+        if let Some(acc) = normalized.accession() {
+            let resolved_tx = acc.transcript_accession();
+            if resolved_tx != requested {
+                if let Some(ctx) = &acc.genomic_context {
+                    if ctx.full() == requested {
+                        return resolved_tx;
+                    }
+                }
+            }
+        }
+        requested.to_string()
     }
 
     /// Project an already-normalized g. variant onto a transcript, skipping the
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

Issue #714, opened 2026-06-15
fix(reference): cdot version-fuzzy fallback applies a sibling version's CDS frame, silently mis-normalizing c. variants


nh13:
## Summary

`cdot::get_transcript` (`src/data/cdot.rs:2470`) does **version-fuzzy fallback** — its docstring: *"if NM_000088.2 is not found but NM_000088.3 exists, the latter will be returned."* When a caller requests an exact transcript version that is **absent from the active cdot** but whose **sequence is present in the FASTA index** (e.g. an older version provisioned into a supplemental FASTA), the resolver returns a **different version's** exon/CDS structure and applies it to the **requested version's sequence**. The reading frames disagree, so `c.` normalization is computed in the wrong frame — and the result is frequently a **confidently wrong, well-formed coordinate**, not an obvious failure.

This also shadows the supplemental-CDS fallback (`src/reference/multi_fasta.rs:1543`, only reached when cdot returns `None`), so populating correct CDS for the exact version doesn't help while the fuzzy match wins.

## Reproducer

With a reference whose index carries `NM_003002.2` (sequence) while the active GRCh38 cdot carries only `NM_003002.3`/`.4`:

```
ferro normalize NM_003002.2:c.6del   ->  NM_003002.2:c.8del      # WRONG
# truth (sequence-direct, frame-independent): c.6 == n.67 (true cds_start=62)
ferro normalize NM_003002.2:n.67del  ->  NM_003002.2:n.68del     # => true answer is c.7del
```

`get_transcript("NM_003002.2")` returns `cds_start=85` (that is `NM_003002.3`'s frame: cdot `start_codon=84 → 85`), not `.2`'s true `cds_start=62`.

## Impact (quantified)

Sweeping all 480 single-base CDS deletions `NM_003002.2:c.{1..480}del` and comparing to the `n.`-derived truth:

| outcome | count |
|---|---|
| correct (neither frame's base rolls) | 276 |
| **wrong, confidently shifted** (well-formed but incorrect) | **110** |
| wrong, no-op (echoes; truth shifts) | 92 |

**202 / 480 (42%)** normalize wrong, 110 of them silently. The wrong outputs are valid-looking HGVS, so nothing flags them. Any transcript whose requested version is absent from the active cdot but present in the sequence index is exposed (provisioned old versions are the obvious case; any FASTA-vs-cdot version skew is the general condition).

## Proposed fix

[...]

PR #787, opened 2026-06-19
fix: decline silent transcript version substitution in normalization and projection

### LINKED ISSUE #785: Projection silently substitutes a sibling transcript version when the requested exact version is absent from cdot
When a versioned `c.`/`r.`/`n.` request names a transcript version that is **absent** from the loaded cdot, projection silently falls back to the **highest** available version of the same base accession and projects against *that* version's CDS/exon frame — instead of failing cleanly. The result is a spec-invalid coordinate mapping whose stated reference and coordinate frame disagree, with **no error and no warning**.

## Mechanism

`CdotMapper::add_transcript` (`src/data/cdot.rs:~2279`) builds a `base_to_versioned` map from each bare accession base to its **highest** seen version (chosen deterministically, `cdot.rs:~2294`). The lenient lookup `get_transcript` (`cdot.rs:2470`) consults that fallback:

```text
// src/data/cdot.rs:2476
// Try version fallback (e.g., NM_000088.2 -> NM_000088.3)
```

So requesting `NM_003002.2` when cdot only carries `NM_003002.3/.4` resolves to `.4`, "shifting CDS coordinates by the versions' length delta" (the code comment at `cdot.rs:~2294` documents exactly this). The version-**exact** lookups added in #714 — `get_transcript_exact` (`cdot.rs:2522`) and `get_transcript_on_build_exact` (`cdot.rs:2540`) — exist precisely to prevent this ("Substituting a sibling version's frame onto another version's sequence yields a spec-invalid description whose stated reference and coordinate frame disagree (#714)"). But the projection path still routes versioned requests through the lenient `get_transcript`, so the guarantee #714 established is not enforced where it matters.

## Why it matters / how it surfaced

Per HGVS, a `c.`/`r.` description is defined only against its named `accession.version` (`background/numbering.md`, `background/refseq.md`); versions differ in length/CDS offset, so a sibling-version frame produces wrong coordinates. A user passing a legitimately historical accession version gets a confidently-wrong projection rather than a "reference unavailable" signal.

This was found while diagnosing why the manifest-gated `mutalyzer-normalize` conformance suite goes wholesale-red on a canonical reference whose GRCh38 cdot lacks two corpus-pinned historical versions (`NM_003002.2`, `NM_002001.2`). Those rows surface as `transcript_id mismatch` / wrong-coordinate divergences rather than clean reference-unavailable misses — masking the data-availability gap behind apparent ferro defects. (The cdot staleness itself is a separate, reference-prep matter.)

## Repro (sketch)

With a cdot that has `NM_003002.3/.4` but not `.2`:

```
ferro project --axis p NM_003002.2:c.1_4delinsTTGA
# emits NP_002993.1:p.(Met1?) projected against the .4 frame, no warning,
# instead of failing with "reference (exact version) not found"
```

## Suggested fix

Route versioned-request projection (at minimum the conformance/projection path; ideally any path where the caller named an explicit `accession.version`) through the version-**exact** lookups (`get_transcript_exact` / `get_transcript_on_build_exact`) added in #714, so an absent exact version yields a clean miss. If a lenient fallback is ever desired, gate it behind an explicit opt-in and emit a warning that names both the requested and substituted versions. Keep the LRG→RefSeq alias path (an exact alias, not version fuzz) unchanged.

## Pointers

- `src/data/cdot.rs`: `add_transcript` / `base_to_versioned` (~2279, ~2294), `get_transcript` (lenient, 2470), `get_transcript_exact` (2522), `get_transcript_on_build_exact` (2540)
- Related: #714 (introduced the exact lookups)

nh13:
## Problem

When a versioned `c.`/`r.`/`n.` request named a transcript `accession.version` the loaded reference did not carry, a lenient provider silently fell back to a **sibling** version and normalized — and projected, since projection normalizes first (#637) — against *that* version's frame/sequence. The result was a spec-invalid coordinate mapping whose stated reference and coordinate frame disagree, emitted with **no error and no warning**. A user passing a legitimately historical accession version got a confidently-wrong result rather than a "reference unavailable" signal.

Per HGVS a `c.`/`r.` description is defined only against its named `accession.version` (`background/numbering.md`, `background/refseq.md`): versions differ in length/CDS offset, so a sibling-version frame produces wrong coordinates.

Closes #785.

## Fix

Gate normalization (the chokepoint projection routes through) on the silent-substitution case: when an explicitly-versioned transcript request is not version-exact yet the provider still resolves it (a sibling is served), decline with `TranscriptVersionNotExact` rather than substitute. `TranscriptVersionNotExact` already maps to a clean reference-unavailable miss in the CLI and conformance layers, so `ferro normalize`/`ferro project` now surface a clean miss instead of a confidently-wrong substitution.

The gate fires only when **all** hold: a transcript-coordinate axis (`c.`/`n.`/`r.`), an explicit version (bare requests keep lenient latest-version resolution; LRG never carries a version), a transcript accession (genomic/gene wrappers like an unrewritten `NG_(GENE):c.…` selector are excluded), the exact version is not served (`has_transcript_version_exact` is false), and the provider nonetheless resolves the request (a sibling is actually served — distinguishing substitution from a genuinely-absent transcript, which is left to the existing missing-transcript path).

The version-exact predicate is the load-bearing signal. It catches **both** substitution paths a `served.id != requested` comparison would not: the FASTA version-strip fallback (which serves the *sibling's* accession) and the cdot-genome reconstruction off a sibling record (which serves the *requested* accession with the sibling's frame). This extends the protein-only version-exact gate of #714/#505 to the coordinate axes for versioned requests.

## Scope

This builds on the deliberate strict/lenient split (#714/#505/#331): the lenient primitive stays lenient, and only versioned requests that would be *silently substituted* are gated — mirroring the existing protein-prediction gate rather than changing low-level lookup semantics. Bare requests, genomic-context wrappers, and the LRG→RefSeq alias path are untouched.

## Conformance

The two `NM_002001.2` mutalyzer-normalize rows previously passed only because ferro silently substituted the sibling `.4` and happened to render the same form. They now correctly decline. A `RejectionReason::VersionSubstitutionRefused785` disposition is added and the rows annotated `accepted_rejection`, so the harness buckets the correct decline (it already buckets a non-panic `Err` on the annotated axis); `failure-patterns.md` is regenerated accordingly.

## Testing

- New normalize unit tests: silent-substitution rejection across `c.`/`n.`/`r.` and a genomic-context-wrapped form; plus the negative cases (bare request, exact-version serve, and absent-without-sibling — all not gated).
- `MockProvider::mark_version_substitution` models the substitution a real `MultiFastaProvider` performs (which the mock otherwise refuses by design).
- Projector tests split into versioned-wholesale-decline and the bare protein-only #505 case.
- Full lib + integration suites pass; clippy `-D warnings` and `fmt --check` clean; spec fixture `--check` passes. Manifest-gated conformance has zero new failures vs. base (verified end-to-end with the prepared reference); only the two intended `NM_002001.2` rows change disposition.

## Reading order

1. `src/normalize/mod.rs` — `silently_substituted_transcript` + the `normalize_core` gate.
2. `src/reference/mock.rs` — substitution test-double support.
3. `src/conformance/mutalyzer.rs` + `tests/fixtures/mutalyzer-normalize/*` — conformance reclassification.

### DIFF
--- a/src/conformance/mutalyzer.rs
+++ b/src/conformance/mutalyzer.rs
@@ -494,6 +494,21 @@ pub enum RejectionReason {
     /// `spec_citation` rows and #537).
     #[serde(rename = "ferro-policy-758-transcript-flank-not-numberable-in-c")]
     TranscriptFlankNotNumberableInC758,
+    /// ferro rejects a c./n./r. input that names an explicit `accession.version`
+    /// the prepared manifest does not carry, where a lenient provider would
+    /// silently fall back to a *sibling* version and normalize against its
+    /// frame/sequence. A c./r. description is defined only against its named
+    /// version (versions differ in length/CDS offset, background/refseq.md), so
+    /// substituting a sibling's frame yields a result whose stated reference and
+    /// coordinate frame disagree. ferro declines with `TranscriptVersionNotExact`
+    /// rather than substitute; mutalyzer (which carries the pinned version)
+    /// normalizes it. ferro's decline is the spec-correct behaviour — a clean
+    /// reference-unavailable miss rather than a confidently-wrong substitution
+    /// (#785). Distinct from the `accession-version-absent` *no-op* disposition,
+    /// which covers an absent version with no sibling to substitute (ferro echoes
+    /// the input); here a sibling exists and the silent substitution is refused.
+    #[serde(rename = "ferro-policy-785-version-substitution-refused")]
+    VersionSubstitutionRefused785,
 }
 
 impl RejectionReason {
@@ -505,6 +520,9 @@ impl RejectionReason {
             }
             RejectionReason::TranscriptFlankNotNumberableInC758 => {
                 "ferro-policy-758-transcript-flank-not-numberable-in-c"
+            }
+            RejectionReason::VersionSubstitutionRefused785 => {
+                "ferro-policy-785-version-substitution-refused"
             }
         }
     }
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -1221,6 +1221,58 @@ impl<P: ReferenceProvider> Normalizer<P> {
         Ok(NormalizeResult::with_diagnostics(result, warnings, infos))
     }
 
+    /// The requested transcript accession a transcript-coordinate variant would
+    /// be SILENTLY SUBSTITUTED for, if any — the #785 trigger.
+    ///
+    /// Returns `Some(requested_id)` only when ALL of the following hold:
+    /// - the variant is a transcript-coordinate axis (`c.`/`n.`/`r.`);
+    /// - the caller named an EXPLICIT version (bare references intentionally
+    ///   keep lenient "latest version" resolution and are not gated; LRG
+    ///   accessions never carry a version, so they are likewise unaffected);
+    /// - the named accession is a transcript ([`is_transcript_accession`], so a
+    ///   genomic/gene wrapper like an unrewritten `NG_(GENE):c.…` selector is
+    ///   not mistaken for a versioned transcript);
+    /// - the provider would NOT serve the bases of that exact version
+    ///   ([`ReferenceProvider::has_transcript_version_exact`] is false); and
+    /// - the provider nonetheless resolves the request — i.e. a sibling version
+    ///   is actually substituted.
+    ///
+    /// The version-exact predicate is the load-bearing signal: it answers "do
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

Issue #715, opened 2026-06-16
feat(cli,project): `--assembly` override to select GRCh37/GRCh38 placement for a bare NG_/LRG_ input


nh13:
Follow-up to #713 / #653.

A bare `NG_`/`LRG_` input carries no genome-build information — the accession is build-agnostic, and the user's intent (which build's chromosomal placement they want) is not encoded in the accession. #713 makes an **explicit** `NC_*.10` genomic-context parent resolve a GRCh37 placement end-to-end, but a bare `NG_` still defaults to GRCh38 (the cdot/mutalyzer primary-assembly policy). There is currently no way to project a bare `NG_` against GRCh37.

Add an explicit assembly override:

- **CLI:** `--assembly {GRCh37,GRCh38}` on `ferro project` (and any other projection-facing command). Name is `--assembly` (values are genome assemblies; matches GATK/samtools vocabulary), **not** `--build` (ambiguous with compile-build).
- **API / Python:** a corresponding optional `assembly` parameter on the public `project*` entry points.

Semantics: the override takes precedence over the variant-derived build signal (`build_hint_for_variant`) at the public entry points and flows into the same placement-selection sites #713 threads. It only *asserts* which placement to select — it does not rewrite the accession. When the requested assembly has no placement for the parent, the existing decline-rather-than-mis-anchor guard (`select_placement_for_build`) fires, consistent with #702.

This is intentionally separated from #713 to keep that PR small; it grafts in cheaply once #713 has inverted the build derivation. Will be opened as a PR stacked on #713.

Refs #713, #653, #480, #642.

PR #788, opened 2026-06-19
refactor(reference): derive accession→assembly from NCBI assembly_report (#716)

### LINKED ISSUE #716: refactor(reference): derive accession→assembly from NCBI assembly_report instead of a hardcoded version table
Follow-up to #713 / #653.

`infer_genome_build_from_accession` currently decides GRCh37 vs GRCh38 from a **hardcoded per-chromosome version table** (the `.10`/`.11` entries in `default_human_aliases`, `src/reference/aliases.rs`). That is brittle: it only knows the chromosome accessions baked into the table, and it encodes assembly identity as a version-number heuristic rather than reading it from an authoritative source.

NCBI ships an `assembly_report.txt` (and `.dict`/`.fai` carry sequence identity) alongside each assembly. `ferro prepare` already downloads both the GRCh37 (`GCF_000001405.25_GRCh37.p13`) and GRCh38 (`…40_GRCh38.p14`) genomes, so the assembly of every prepared sequence is already known at prepare time.

Proposal: parse an `assembly_report.txt` (RefSeq accession → assembly name, per sequence) into a data-driven accession→assembly map, recorded in the prepared manifest, and have build inference consult that map instead of (or before falling back to) the hardcoded version table. This hardens build inference **everywhere** — not just for `NG_`/`LRG_` placement selection.

Note: this does **not** disambiguate a bare `NG_` *variant* input's intended build — a sidecar describes the *reference*, not the *query*. That ergonomic gap is handled by the `--assembly` override (separate follow-up). This issue is strictly about making the reference-side accession→assembly mapping authoritative.

Refs #713, #653.

nh13:
Closes #716.

## What

Build inference (`infer_genome_build_from_accession`) decided GRCh37 vs GRCh38 from a hardcoded per-chromosome version table in `src/liftover/aliases.rs`. That is brittle: it only knows the chromosome `.version` accessions baked into the table, and it encodes assembly identity as a version-number heuristic rather than reading it from an authoritative source.

This wires up the data-driven path end-to-end so build inference consults the prepared reference's own NCBI `*_assembly_report.txt` first, falling back to the hardcoded table only when the report does not describe an accession (or no report is present). The parser (`assembly_report.rs`), `ContigAliases::from_assembly_reports`, and `infer_genome_build_from_accession_with` already existed and were unit-tested but unwired; this PR connects download → manifest → provider → inference.

## How

- **`ferro prepare`** downloads the `*_assembly_report.txt` alongside each genome it fetches (`record_assembly_report`) and records the path in the manifest. A report-download failure is a soft warning that degrades to the hardcoded fallback — it never aborts the genome download.
- **Manifest** carries two new `Option<PathBuf>` fields, `assembly_report` / `assembly_report_grch37`, with `#[serde(default)]` so pre-existing manifests still load. Both are wired into the `for_each_path` / `for_each_path_mut` relativize/absolutize/validate plumbing.
- **`MultiFastaProvider::from_manifest`** parses the recorded report(s) into a per-build `ContigAliases` table (stored as `None` when no report is named or it parses empty, so an empty table never shadows the fallback).
- **Inference** routes through a new `ReferenceProvider::infer_genome_build` trait method whose default is the hardcoded heuristic; `MultiFastaProvider` overrides it to layer the data-driven table over that fallback (`infer_genome_build_layered`). All provider-reachable call sites in `projector.rs` and `multi_fasta.rs` route through the provider, and `select_placement_for_build` takes the inference as an injected closure so the change reaches placement selection too. The `Arc<T>` / `Box<T>` blanket impls forward `infer_genome_build` so a wrapped provider keeps the data-driven behavior.
- **`ferro check`** warns when a named-but-missing assembly report is absent, matching the existing optional-input checks.

The hardcoded `default_human` table contents are unchanged — it remains the fallback. For any primary chromosome the two tables agree by construction (the hardcoded table was derived from these reports); the only cross-build accession, mtDNA `NC_012920.1`, resolves GRCh38 either way via the GRCh38-first check order. Net behavioral gain: `NC_` versions the report describes but the hardcoded table never knew now classify correctly.

## Out of scope

Disambiguating a bare `NG_` variant *input*'s intended build — a sidecar describes the *reference*, not the *query*; that ergonomic gap is the `--assembly` override (#715, already landed).

## Testing

- New unit tests: the layering helper (report-only version classifies, hardcoded fallback, no-table parity), manifest round-trip + pre-existing-manifest back-compat, the `skip_existing` prepare path, `Arc`/`Box` forwarding of `infer_genome_build`, a `from_manifest` integration test that classifies a report-only version via the loaded table, and `ferro check` missing-report warnings.
- Full suite: 7430 passed / 54 skipped. `cargo clippy --features dev --all-targets -- -D warnings` and `cargo fmt --all --check` clean. Spec fixture `--check` passes.
- Manifest-gated conformance against the prepared reference: zero regression (the prepared manifest carries no report yet, so this exercises the fallback path — the data-driven path is covered by the unit/integration tests above; re-running `ferro prepare` will provision the reports for end-to-end conformance).

## Summary by CodeRabbit

* **New Features**
  * Genome preparation can now automatically download GRCh38/GRCh37 assembly reports and use them to improve genome build inference.
* **Bug Fixes**
  * Missing GRCh38/GRCh37 assembly report files now surface as warnings instead of failing silently.
* **Improvements**
  * Reference manifests now persist assembly report paths (when available), restoring them correctly on reload and including them in manifest path validation.

### DIFF
--- a/src/check/mod.rs
+++ b/src/check/mod.rs
@@ -182,6 +182,24 @@ pub fn check_reference(reference_dir: &Path) -> CheckResult {
         }
     }
 
+    // A missing assembly report silently disables the data-driven build map
+    // at load (#716), so surface it like the other optional inputs above.
+    if let Some(ref report) = manifest.assembly_report {
+        if !report.exists() {
+            result
+                .warnings
+                .push(format!("Assembly report not found: {}", report.display()));
+        }
+    }
+    if let Some(ref report) = manifest.assembly_report_grch37 {
+        if !report.exists() {
+            result.warnings.push(format!(
+                "GRCh37 assembly report not found: {}",
+                report.display()
+            ));
+        }
+    }
+
     // A missing derived-placements JSON silently disables version-gap NG_/LRG_
     // projection at load, so surface it like the other optional inputs above.
     if let Some(ref placements) = manifest.derived_refseqgene_placements {
@@ -335,6 +353,8 @@ mod tests {
             refseqgene_fastas: Vec::new(),
             refseqgene_alignments: None,
             refseqgene_alignments_grch37: None,
+            assembly_report: None,
+            assembly_report_grch37: None,
             derived_refseqgene_placements: None,
             refseqgene_summary: None,
             lrg_fastas: Vec::new(),
@@ -387,6 +407,8 @@ mod tests {
             refseqgene_fastas: Vec::new(),
             refseqgene_alignments: Some(missing_alignments.clone()),
             refseqgene_alignments_grch37: None,
+            assembly_report: None,
+            assembly_report_grch37: None,
             derived_refseqgene_placements: None,
             refseqgene_summary: None,
             lrg_fastas: Vec::new(),
@@ -441,6 +463,8 @@ mod tests {
             refseqgene_fastas: Vec::new(),
             refseqgene_alignments: None,
             refseqgene_alignments_grch37: Some(missing_alignments.clone()),
+            assembly_report: None,
+            assembly_report_grch37: None,
             derived_refseqgene_placements: None,
             refseqgene_summary: None,
             lrg_fastas: Vec::new(),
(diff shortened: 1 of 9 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #728, opened 2026-06-17
feat(reference): version-independent NG_/LRG_ genomic_placement via provisioned FASTA + validated affine derivation


nh13:
Spun out of #645 (see that issue's scoping discussion).

## Background

`ReferenceProvider::genomic_placement` resolves an `NG_`/`LRG_` parent's chromosomal placement by **version-exact** match against the ingested RefSeqGene alignment snapshot (`GCF_*_refseqgene_alignments.gff3`). When an input pins an `NG_` version absent from the snapshot, ferro correctly **declines** (rather than mis-anchor across versions) — see #655/#702. The consequence is a version-coverage gap: `NG_`-parented inputs on those versions cannot be placed.

#645 handles the *conformance* side of this (dispositioning the corpus rows as a known data-coverage gap). This issue tracks the *general* fix: removing the snapshot-version dependency so ferro can place **any** `NG_`/`LRG_` version it has (or can provision) sequence for — valuable for real-world users querying older `NG_` versions, independent of the corpus.

## Why this is its own issue (not folded into #645)

The 4 corpus accessions #645 covers are **all version-mismatched** (`.1`/`.2` pinned; only `.2`/`.3` on hand), and ~20 of the affected rows are *additionally* gated on an orthogonal transcript-version gap (#672: cdot lacks `NM_003002.2`). So even a perfect placement engine does **not** make those corpus rows pass — they still need a #672 disposition. The engine's justification is therefore **general-user value, not corpus conformance**, and it deserves its own design cycle.

## Feasibility (verified)

- The pinned-version **sequence** IS obtainable from NCBI EFetch (confirmed: `efetch NG_012337.1` returns the 16,387-byte FASTA). Only the precomputed **alignment** predates NCBI's ~2017 archive floor and is gone.
- ferro already has a supplemental-fetch precedent: the manifest's `supplemental_fasta` ("missing ClinVar transcripts fetched from NCBI").

## Proposed approach (to be designed)

1. **Provision** the needed `NG_` version FASTA at `ferro prepare` time (EFetch, like `supplemental_fasta`). Resolve the "which versions does prepare fetch?" question — a curated list, or driven by an input manifest, not runtime fetching (ferro is offline at parse time).
[...]

PR #793, opened 2026-06-20
fix(normalize): resolve legacy GENE_v001 selector against the NG_ parent's hosted transcript (#792)

### LINKED ISSUE #792: fix(normalize): legacy GENE_vNNN selector resolves to global reference-standard, not the NG_ parent's hosted transcript version
## Summary

A legacy LOVD `GENE_vNNN` selector on a **versioned** `NG_` parent resolves to the gene's **global reference-standard** transcript instead of the transcript version the **named `NG_` parent actually hosts** — producing the wrong isoform on the protein/coding axes and a stale rewrite on the normalized axis.

`NG_012337.1(TIMM8B_v001):c.12_13insGATC`:
- ferro → `NP_036591.3:p.(Gly5AspfsTer6)` (resolves `TIMM8B_v001` → NM_012459.4)
- oracle → `NG_012337.1(NP_036591.2):p.(Ser5AspfsTer21)` (NM_012459.2)

## Root cause (verified)

- ferro resolves the selector via `legacy_gene_models` (`src/reference/multi_fasta.rs:2253` `resolve_legacy_gene_selector`, fed from `legacy_selector.rs:41 parse_refseqgene_summary`), a `gene → reference-standard NM_` map built from NCBI's `LRG_RefSeqGene` table. For TIMM8B that row is `… TIMM8B  NG_033145.1  …  NM_012459.4  …  NP_036591.3  …  reference standard` — keyed off a **different** RefSeqGene (`NG_033145.1`) than the input names (`NG_012337.1`).
- `NG_012337.1`'s own GenBank record hosts `/transcript_id="NM_012459.2"` for TIMM8B (its mRNA feature). The mutalyzer oracle's keyword says it outright: "Transcripts mismatch between genbank and gff3."
- The resolver is **NG-parent-blind**: `rewrite_legacy_gene_selector` (`src/normalize/mod.rs:3162-3198`) has the parent accession in scope (`mod.rs:3175`) but passes only the gene symbol to `resolve_legacy_gene_selector`. LOVD `_vNNN` is **positional on a specific RefSeqGene record** (the module docs say so: `legacy_selector.rs:10`, `:24-26`), i.e. parent-relative — not a pointer to the gene's globally-current reference standard.
- Consequence: the global map applies a transcript version the named parent does not host. ferro converges to the oracle on **all three axes** (normalized / genomic / protein) when the hosted transcript `NM_012459.2` is named explicitly — so the defect is isolated to this one resolution step.

Also: the `normalized`-axis `improvement` annotation on these rows ("ferro preserves the input's non-transcript selector", tracking #500) is now **stale** — ferro no longer preserves `TIMM8B_v001`; it rewrites to `NG_012337.1(NM_012459.4)`, disagreeing with the oracle's `NM_012459.2`. So this is a latent normalized-axis regression too, not only a protein-axis FAIL.

## Two fix paths

**(A) Full fix — reaches the oracle.** Ingest per-`NG_`-version hosted-transcript lists at `prepare` time (parse `/transcript_id` mRNA features from each RefSeqGene GenBank record, keyed by `NG_acc.version` + gene + `_v` order), record in the manifest, and thread the `NG_` parent into `resolve_legacy_gene_selector` so `(gene, NG_parent)` → the transcript the parent hosts. This is the behavior the module docs foreshadow. Bounded but real (new manifest artifact + parser + context-aware resolver signature). The hosting fact is **not** currently ingested — cdot has no NG_→tx association, `derived_refseqgene_placements.json` carries only the NG_→NC_ affine (no tx IDs), the GFF3 has only `NG_012337.3`→NC_, and the refseqgene FASTA carries only `NG_012337.3`.

**(B) Interim fix — stops the wrong rewrite (uses data already present).** The `LRG_RefSeqGene` table carries the `RSG` (`NG_`) column. Carry it into the resolver and, when rewriting a selector on an `NG_` parent, only apply the global map's transcript if the map row's `RSG` matches the input's `NG_` parent; otherwise **decline/preserve** the selector rather than rewrite to a foreign-parent transcript. This stops the confidently-wrong isoform + broken projection and restores the "preserve" behavior the `normalized` annotation describes. It does **not** reach the oracle's `NM_012459.2` (preserving yields no protein prediction). Must be verified not to regress the `BRCA2_v001`/`PCCB_v001` rows #784 made pass (only safe if their input `NG_` parents match their reference-standard `RSG` rows).

## Acceptance (full fix)

- `NG_012337.1(TIMM8B_v001):c.12_13insGATC` → `NG_012337.1(NP_036591.2):p.(Ser5AspfsTer21)` (and the normalized/genomic axes match the oracle).
- No regression to the `BRCA2_v001`/`PCCB_v001` rows (#784).
- The stale `normalized` `improvement` annotation on the TIMM8B_v001 rows is corrected/removed.

Refs #500, #637, #783, #784, #326. (#784 punted this as "TIMM8B_v001 protein-version drift #505", but #505 is a different, closed issue — version-exact *sequence availability*.)

nh13:
Closes #792.

> Stacked on #787 (`fix/issue-785-version-exact-projection`) — this PR targets that branch; it will retarget to `main` automatically when #787 merges. The diff below is #792 only (8 commits).

## What

A legacy LOVD `GENE_v001`/bare selector on a **versioned** `NG_` parent resolved to the gene's **global reference-standard** transcript instead of the transcript the named parent actually hosts:

```
NG_012337.1(TIMM8B_v001):c.12_13insGATC
  before: NP_036591.3:p.(Gly5AspfsTer6)         (TIMM8B → global NM_012459.4, from NG_033145.1's row)
  after:  NG_012337.1(NP_036591.2):p.(Ser5AspfsTer21)   (NM_012459.2 — the transcript NG_012337.1 hosts)
```

The resolver was NG_-parent-blind: it consulted only a global `gene → reference-standard NM_` map (from `LRG_RefSeqGene`), whose TIMM8B row points at a *different* RefSeqGene (NG_033145.1) than the input names. LOVD `_v00N` is parent-relative (positional on a specific RefSeqGene record), so the global proxy is wrong whenever the input's parent is a different/older NG_.

## How

- **prepare** (`--derive-ng-placements`) now also emits `ng_hosted_transcripts.json` (`NG_acc.version → gene → [hosted transcript_ids]`) from the **same** NG_ GenBank records it already fetches for #728's placement derivation — no second efetch. The hosting parser includes partial (`<`/`>`) mRNA features (hosting membership, unlike the affine parser which skips them).
- **manifest** carries a new `ng_hosted_transcripts` path (serde-default; old manifests load).
- **`MultiFastaProvider`** loads the artifact (degrade-don't-abort on missing/garbled).
- **resolver**: the `ReferenceProvider::resolve_legacy_gene_selector` trait method gains an `ng_parent: Option<&Accession>` argument (forwarded through the `Arc`/`Box` blanket impls, mock, and test providers). A bare/`_v001` selector on an NG_ parent resolves to the transcript that **exact** NG_ version **uniquely hosts** (exactly one); otherwise it falls back to today's global reference-standard map. `normalize::rewrite_legacy_gene_selector` passes the in-scope NG_ parent.

Deliberately **not** positional `_vNNN`: there is no authoritative rule mapping GenBank mRNA feature order to LOVD `_v` numbering (the spec submodule defines none; `SDHD_v002` has only one SDHD mRNA on NG_012337). `_v002+` and genes with ≥2 hosted transcripts keep declining. The misleading `legacy_selector.rs` module doc that claimed feature-order is "the definitional `_v` source" is corrected.

## Effect

The fix is parent-relative and unambiguous-only, so it never guesses: it only overrides the global map when the exact NG_ parent hosts a single transcript for the gene. Verified against the real reference: the whole `NG_012337.1` TIMM8B/SDHD cluster (`_v001`, bare, and SDHD_v001) plus `NG_009497.1(USH2A_v001)` now reach the mutalyzer oracle; their now-stale `#500` normalized-axis annotations are removed. `BRCA2_v001`/`PCCB_v001` (#784) are unchanged — their NG_ parents host a single transcript matching the global map.

## Testing

- New hermetic unit tests: gene/partial-inclusion parser; `hosted_unique` (single-hosted resolves, multi-hosted/wrong-version/unknown decline; schema_version validation); manifest round-trip; prepare co-emit; provider load; resolver (`_v001` hosted wins, `_v002` declines, no-parent/not-hosted → global fallback).
- Full hermetic suite: 7435 passed / 54 skipped. `cargo clippy --features dev --all-targets -- -D warnings` + `cargo fmt --all --check` clean. Spec fixture `--check` passes.
- Manifest-gated conformance against the prepared reference (artifact provisioned from the corpus NG_ set): the TIMM8B/USH2A_v001 rows demote to passing; **no regression** — base-vs-HEAD per-axis divergence counts are identical (the remaining manifest-gated `axis_*` differences are pre-existing local-reference-vs-pinned-baseline mismatches, unchanged by this PR).

## Out of scope

Positional `_vNNN` (`_v002+`); LRG_ parent selectors; provisioning NCBI-purged old NG_ versions (a future corpus row naming a purged version falls back to the global map — documented).

### DIFF
--- a/examples/derive_ng_placements.rs
+++ b/examples/derive_ng_placements.rs
@@ -109,20 +109,20 @@ fn run(cli: &Cli) -> anyhow::Result<()> {
          Source manifest: {}.",
         cli.manifest.display()
     );
-    let (artifact, declined) =
-        derive_placements_for_accessions(&provider, &accessions, &builds, description)
-            .map_err(|e| anyhow::anyhow!("derive: {e}"))?;
+    let out = derive_placements_for_accessions(&provider, &accessions, &builds, description)
+        .map_err(|e| anyhow::anyhow!("derive: {e}"))?;
 
-    let json = artifact
+    let json = out
+        .artifact
         .to_json()
         .map_err(|e| anyhow::anyhow!("serialize: {e}"))?;
     std::fs::write(&cli.out, &json)
         .map_err(|e| anyhow::anyhow!("write {}: {e}", cli.out.display()))?;
     println!(
         "Derived {} placement(s) for {} accession(s), {} declined. Wrote {}.",
-        artifact.placements.len(),
+        out.artifact.placements.len(),
         accessions.len(),
-        declined,
+        out.declined_count,
         cli.out.display()
     );
     Ok(())
--- a/examples/extract_mutalyzer_genomic_windows.rs
+++ b/examples/extract_mutalyzer_genomic_windows.rs
@@ -187,8 +187,12 @@ impl ReferenceProvider for RecordingProvider {
         Some(p)
     }
 
-    fn resolve_legacy_gene_selector(&self, selector: &str) -> Option<String> {
-        self.inner.resolve_legacy_gene_selector(selector)
+    fn resolve_legacy_gene_selector(
+        &self,
+        selector: &str,
+        ng_parent: Option<&ferro_hgvs::hgvs::variant::Accession>,
+    ) -> Option<String> {
+        self.inner.resolve_legacy_gene_selector(selector, ng_parent)
     }
 
     fn has_transcript(&self, id: &str) -> bool {
--- a/src/check/mod.rs
+++ b/src/check/mod.rs
@@ -223,6 +223,19 @@ pub fn check_reference(reference_dir: &Path) -> CheckResult {
         }
     }
 
+    // A missing NG_ hosted-transcripts map silently drops parent-relative legacy
+    // selector resolution at load (#792) — every NG_-parent legacy selector then
+    // falls back to the global reference-standard map — so surface it like the
+    // other optional inputs above.
+    if let Some(ref ng_hosted) = manifest.ng_hosted_transcripts {
+        if !ng_hosted.exists() {
+            result.warnings.push(format!(
+                "NG_ hosted-transcripts map not found: {}",
+                ng_hosted.display()
+            ));
+        }
+    }
+
     result
 }
 
@@ -356,6 +369,7 @@ mod tests {
             assembly_report: None,
             assembly_report_grch37: None,
             derived_refseqgene_placements: None,
+            ng_hosted_transcripts: None,
             refseqgene_summary: None,
             lrg_fastas: Vec::new(),
             lrg_xmls: Vec::new(),
@@ -410,6 +424,7 @@ mod tests {
             assembly_report: None,
             assembly_report_grch37: None,
             derived_refseqgene_placements: None,
+            ng_hosted_transcripts: None,
             refseqgene_summary: None,
             lrg_fastas: Vec::new(),
             lrg_xmls: Vec::new(),
@@ -466,6 +481,7 @@ mod tests {
             assembly_report: None,
             assembly_report_grch37: None,
             derived_refseqgene_placements: None,
+            ng_hosted_transcripts: None,
             refseqgene_summary: None,
             lrg_fastas: Vec::new(),
             lrg_xmls: Vec::new(),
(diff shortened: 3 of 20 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

Issue #771, opened 2026-06-18
fix(project): report p.? for an initiation-codon variant on a non-AUG-initiation transcript


nh13:
## Problem

For a variant that **affects the initiation codon** of a **non-AUG-initiation** transcript (CUG/GUG start with a `transl_except`), ferro declines protein prediction entirely (`no p. representation`) instead of reporting the unknown consequence.

Example (`NM_024426.4`, a non-AUG transcript whose CDS begins with C):

- `NM_024426.4:c.1C>G` → ferro: declines; mutalyzer: `NM_024426.4(NP_077744.3):p.?`
- `NM_024426.4:c.1C>A` → same

Cause: the #625 CDS-start sanity guard declines *all* protein prediction when the CDS does not begin with `ATG` (to avoid wrong-frame translation of inconsistent annotations). That guard runs before the initiation-codon `p.(Met1?)` short-circuit, so even a start-codon variant — whose consequence is *unknown regardless of the start codon identity* — is declined.

## Fix

A variant affecting the initiation codon should report an unknown form **without translating** (so the #625 frame concern does not apply): `p.(Met1?)` for a canonical ATG start, or the plain whole-protein `p.?` for a non-AUG start (residue 1 is not `Met`, so `Met1?` would be wrong; `p.?` matches the spec — `protein/substitution.md:51`, `uncertain.md` — and mutalyzer). Non-initiation variants on non-AUG transcripts keep declining (#625 unchanged).

This reduces the protein-axis empty-projection count and lets these rows green via the existing bare-NP-protein divergence (the only remaining delta is ferro's spec-correct bare `NP_` vs mutalyzer's `NM_(NP_)` wrapper).

Out of scope: `c.1_4delinsATGA` normalizes to a compound allele `c.[1C>A;4C>A]` (a separate allele-protein path).

Refs #498, #625.

PR #796, opened 2026-06-21
feat(project): translate downstream variants on non-AUG-initiation transcripts (#780)

### LINKED ISSUE #780: feat(project): translate downstream variants on legitimate non-AUG-initiation transcripts (#625 over-approximation)
## Problem

The #625 CDS-start sanity guard declines **all** protein prediction when the reference CDS does not begin with `ATG`, to avoid wrong-frame translation of inconsistent cdot annotations. This deliberately over-approximates: it also declines the rare **legitimate** non-AUG-initiation transcripts (CUG/GUG starts), where the CDS *is* internally consistent. #771/#772 carved out the initiation-codon-*affecting* case (-> `p.?`/`p.(Met1?)`), and #778/#776 extended that to cis compound alleles, but **downstream** variants on these transcripts still decline outright.

**Repro** (`NM_024426.4`, WT1, a CUG-initiation transcript):

```
NM_024426.4:c.1107A>G   -> ferro: "no p. representation"   (mutalyzer/corpus: p.(=))
NM_024426.4:c.500...    -> declines
NM_024426.4:c.100...    -> declines
```

The whole transcript's downstream protein surface is unavailable, even for a synonymous change that obviously yields `p.(=)`.

## Background (researched)

Non-AUG / alternative translation initiation is real but **rare in curated RefSeq annotation (~0.1% of mRNAs** carry a non-AUG CDS start; Tikole & Sankararamakrishnan 2006, PMID 16780373). **CUG** is the dominant, most-efficient non-AUG initiator; legitimate near-cognates are CUG/GUG/UUG/ACG. A well-characterized human set uses it: **WT1, MYC, FGF2, VEGFA, PTEN** (CUG) and **EIF4G2** (GUG). WT1's CUG lies 204 nt (68 codons) upstream of, and in frame with, the canonical AUG. There is **no authoritative pre-built list** of human non-AUG transcripts; the signal must be derived per-transcript.

**Key framing insight:** for a **downstream** variant, the start-codon identity is irrelevant — standard-table vs Met-forced translation differ *only at residue 1*. The wrong-frame risk #625 guards against is a **wrong `cds_start` (mid-frame)**, not a non-`ATG`-but-correct `cds_start`. So the real discriminator is "is `cds_start` in the right frame?", which the reference protein confirms (see options).

## What established tools do (researched, source-verified)

- **biocommons `hgvs` / VariantValidator**: `AltSeqBuilder` slices from the annotated `cds_start` and translates **without validating ATG/AUG** — trusts non-AUG starts as-is ([altseqbuilder.py](https://github.com/biocommons/hgvs/blob/main/src/hgvs/utils/altseqbuilder.py)); so it *would* yield `p.(=)` for a downstream synonymous variant. Start-loss -> `p.Met1?`-style via an `init_met`/ambiguous flag ([altseq_to_hgvsp.py](https://github.com/biocommons/hgvs/blob/main/src/hgvs/utils/altseq_to_hgvsp.py)). (Carries `# TODO: handle ... new Met start`.)
- **SnpEff**: codon tables mark `TTG/L+, CTG/L+, ATG/M+` as starts; first codon **forced to `M`** if it's a recognized start, AUG or not ([CodonTable.java](https://github.com/pcingola/SnpEff/blob/master/src/main/java/org/snpeff/codons/CodonTable.java), [snpEff.core.config](https://github.com/pcingola/SnpEff/blob/master/config/snpEff.core.config)).
- **Mutalyzer v3** (cautionary, not a model): `protein.py` translates the CDS with Biopython's **standard table** — no `/codon_start`/`/translation` read, **no forced Met** — and the retriever has **no GenBank parser** (GFF3 carries only `transl_except`/`exception`) ([protein.py](https://github.com/mutalyzer/mutalyzer/blob/master/mutalyzer/protein.py), [retriever gff3.py](https://github.com/mutalyzer/retriever/blob/master/mutalyzer_retriever/parsers/gff3.py)). It mistranslates a non-AUG start (CUG->Leu at residue 1; downstream residues are unaffected). For **any first-codon change it returns `p.?`** unconditionally (`if cds_seq[:3] != cds_obs_seq[:3]: return ...:p.?`).
- **Ensembl VEP**: `start_lost` (SO:0002012, HIGH) / `start_retained_variant` (SO:0002019, LOW), defined against the *"canonical"* start without specifying AUG vs non-AUG (non-AUG case undefined in docs); can co-emit start_lost+start_retained ([deemed correct, not a bug](https://github.com/Ensembl/ensembl-vep/issues/727)). Whether VEP *translates* a non-AUG start vs declines is not established. **GENCODE/Ensembl do annotate non-AUG first-class** via the `non_ATG_start` tag ("non-ATG start supported by publication/conservation"), plus `cds_start_NF`/`mRNA_start_NF` ([GENCODE tags](https://www.gencodegenes.org/pages/tags.html)).

**Implication for ferro:** the peer translators (biocommons `hgvs`/VariantValidator, SnpEff) **trust the annotated CDS start and force `Met1` without requiring ATG** — ferro's #625 decline is *more conservative than its peers*. Matching the ecosystem needs only "trust `cds_start` + force `Met1`"; a consistency check is an *extra* safety layer.

## Start-loss representation (spec — ferro is correct)

HGVS gives a **three-way** start-loss notation: `p.0` = no protein produced (experimental support), `p.0?` = predicted no protein, **`p.(Met1?)`** = initiation-codon variant whose consequence is unknown — distinct from the generic `p.?` ([HGVS protein/substitution](https://hgvs-nomenclature.org/stable/recommendations/protein/substitution/)). **ferro's `p.(Met1?)` (and `p.?` for non-AUG residue-1) is spec-aligned; Mutalyzer's `p.?` for a start-codon delins is the divergence.** This validates #771/#772/#778; where ferro and mutalyzer differ on a start-codon row (e.g. the SDHD `NG_012337.1(NM_003002.2):c.1_4delinsTTGA` corpus row, if reachable), ferro's value is the spec-correct side -> disposition as `spec_citation`, not a ferro bug.

## Root cause / data dependency

- ferro detects "non-AUG" only by reading the CDS start codon from the FASTA (`cds_has_valid_start`); no metadata flag.
- **cdot carries no start-codon-exception flag** (only a numeric `start_codon`); it's built from GTF/GFF3, which lack a translation-exception attribute, so the pipeline drops the signal.
- **The prepared reference has no protein FASTAs** (`manifest.protein_fastas: []`).

### `/transl_except` is NOT a usable signal here (empirically verified)

The actual `NM_024426.4` RefSeq GenBank record annotates the CDS to start at the CUG with **no `/transl_except`** — the non-AUG fact is free-text `/note` only:

```
CDS  191..1744
     /note="...non-AUG (CUG) translation initiation codon; Wilms tumor protein isoform Ex4a(+)"
     /codon_start=1
     /protein_id="NP_077744.3"
     /translation="MQDPASTCVPEP..."
```

But the `/translation` (and the `NP_077744.3` FASTA) **begins with `M`**: translating from the annotated start with a **forced initiator Met** reproduces `NP_077744.3`. That reference protein is the discriminator #625 is missing.

## Implementation options (re-prioritized)

1. **Trust the annotated CDS start + force `Met1` (peer-parity, smallest change).** For a transcript with a set cdot `cds_start`, drop the #625 decline and translate from it with residue 1 pinned to `Met` — as biocommons `hgvs`/VariantValidator and SnpEff do. Risk: re-exposes the inconsistent-`cds_start` case #625 guarded against. Mitigated by option 2.
2. **(Preferred) Trust + reference-protein consistency check (stricter than peers).** As option 1, but only translate when the CDS (forced Met1) reproduces the reference protein; otherwise keep declining. Preserves #625's protection by construction. Data requirement: **reference protein** — `NP_` FASTAs (a `ferro prepare` change; `manifest.protein_fastas` empty) **or** GenBank `/translation`.
3. **Ingest GenBank CDS metadata / GENCODE tags.** RefSeq side: `/translation` (overlaps option 2's data need) from the GenBank flat file (NCBI Datasets / eutils) or ferro's existing `legacy_genbank_metadata`; **not** `/transl_except`. Ensembl side: the first-class **`non_ATG_start`** GENCODE tag is a clean signal if ferro's Ensembl-cdot path is extended to carry it.
4. **(Fallback) Curated allowlist** of known non-AUG transcripts + start positions — smallest footprint; least general; reasonable stopgap given the small set.

## Constraints

Must **not** regress #625's protection: a genuinely inconsistent annotation (wrong `cds_start`) must still decline rather than fabricate a wrong-frame protein. Option 2's consistency check satisfies this by construction. (Option 1 alone merely matches peer behaviour, which does no such check.)

## Ecosystem context (open problem)

Correct non-AUG / start-loss handling is a **recognized, quantified, underspecified** problem: conservation predictors (SIFT/PolyPhen) misclassify initiation-codon variants; alternative downstream starts make naive prediction unreliable; ANNOVAR/SnpEff/VEP agree on HGVSc only **58.52%** (HGVSp 84.04%) across 164,549 ClinVar two-star variants; VEP issue #727 documents start_lost/start_retained co-emission. ferro adopting option 2 would be **stricter and more correct than the surveyed peers**.

## Remaining open questions (require live testing, not desk research)

- Whether Ensembl/VEP **translates** a non-AUG / `cds_start_NF` annotated start vs declines, and whether it consumes the `non_ATG_start` tag (VEP docs silent; needs a VEP code read or a test transcript).
- Whether any HGVS nomenclature-committee / Varnomen thread addresses consequence representation for variants on non-AUG-*initiated* transcripts specifically (vs canonical-start start-loss).

Refs #498, #625, #771, #776/#778.

nh13:
Closes #780.

## Summary

The #625 CDS-start guard declines **all** protein prediction when the reference CDS does not begin with `ATG`, to avoid wrong-frame translation of inconsistent cdot annotations. This over-declines legitimate **non-AUG-initiation** transcripts (CUG/GUG starts, e.g. WT1 `NM_024426.4`) for **downstream** variants. #771/#772/#778 carved out the initiation-codon-*affecting* case (→ `p.(Met1?)`/`p.?`); downstream variants still declined.

This widens the runtime decline gate from "starts with `ATG`" to "starts with a **recognized initiator** (`ATG` or near-cognate `CTG`/`GTG`/`TTG`)", reusing the project's own data-side classification (`is_alternative_start_codon`, validate.rs — *"the reading frame is correct and the initiator Met is still installed, so they are legitimate non-`ATG` starts, not the off-the-start drift"*). The runtime guard was simply **narrower than what the same codebase already knew was safe**.

`NM_024426.4:c.1107A>G` now translates to `NP_077744.3:p.(Arg369=)` instead of declining.

## Why this approach (adversarially reviewed)

Considered and rejected: trust-any-non-ATG (re-exposes #625's drift risk for a start that drifted onto a non-initiator codon); a reference-protein consistency check (the prepared reference has `protein_fastas: []`, the check is circular as wired, and the substitution path never consumes the ref protein — demotes zero rows); a curated allowlist (version-bump fragile, duplicates `load_cds_allowlist`). Widening to the recognized-initiator set needs **no reference data**, demotes the row, reuses an already-reviewed predicate, and preserves #625's protection — a start on a genuine drift codon (`AGT`/`GAT`/…) still declines.

## Preserves #625 and #771

- The decline gate widens; the init-codon **form choice** keeps the strict `ATG` test, so a CUG-start init-codon variant still reports `p.?` (not `p.(Met1?)`).
- A downstream variant on a genuine off-the-start drift codon (not a recognized initiator) still declines.
- Both are covered by new MockProvider tests (CI-runnable).

## Corpus disposition (spec-grounded)

After the fix, ferro's `NP_077744.3:p.(Arg369=)` is **spec-correct on both axes** where it diverges from mutalyzer's expected `NM_024426.4(NP_077744.3):p.(=)`, per HGVS nomenclature v21.1.4:
- silent substitutions are described residue-specifically — `p.Cys123=` / `p.(Lys874=)` (`recommendations/protein/substitution.md:24,72`); mutalyzer's bare `p.(=)` (whole-protein-unchanged) is non-standard for a single synonymous codon;
- protein consequences are rendered bare (`p.(Ser42Cys)`, naming the `NP_` when isoforms exist) — `recommendations/general.md:147,148`; there is no `transcript(NP_):p.` wrapper in the spec.

So the WT1 row is dispositioned `spec_overridden` (the existing `bare-np-protein` cluster) and demoted from `baseline-failures/protein_description.txt`; the generated `failure-patterns.md` is regenerated. 0 regressions.

## Reading order

1. `src/project/protein/helpers.rs` — the `cds_has_recognized_start` predicate.
2. `src/project/projector.rs` — the decline gate (now `start_is_recognized_initiator`) vs the unchanged strict-`ATG` form choice; the new + updated MockProvider tests.
3. `tests/fixtures/mutalyzer-normalize/` — the spec_overridden disposition.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of protein consequence prediction for variants affecting initiation codons by distinguishing between canonical and alternative start codons
  * Enhanced downstream translation analysis for transcripts with recognized alternative initiators
  * Better error reporting for non-standard start codon scenarios

* **Tests**
  * Expanded test coverage for alternative start codon and non-standard variant scenarios


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #502, opened 2026-05-31
feat(normalize): emit NP_ protein-accession selector for p. variants on genomic references (NM_→NP_)


nh13:
Split out from the #500 / #501 `NG_` selector arbitration. While migrating the 14 legacy `accepted_divergence` (#121) rows to the new `improvement` disposition (#501), 13 were the gene-symbol/gene-ID → transcript-accession pattern tracked by #500. The 14th is a **distinct** divergence and was left as `accepted_divergence` pending its own arbitration:

```
input:     NG_012337.3(NM_003002.4):p.(Asp92Tyr)
ferro:     NG_012337.3(NM_003002.4):p.(Asp92Tyr)   # preserves the input NM_ selector
mutalyzer: NG_012337.3(NP_002993.1):p.(Asp92Tyr)   # resolves to the NP_ protein accession
```

This is not a gene-symbol case (so it is out of #500's scope): the input already carries a transcript (`NM_`) accession, and the divergence is `NM_` vs `NP_` as the selector for a **protein (`p.`) coordinate** on a genomic reference.

## Spec basis

`assets/hgvs-nomenclature/docs/background/refseq.md` L38-42: accepted parenthetical specifications "include transcripts (`NM_004006.2`) and proteins (`NP_003997.1`)". A `p.` description is a protein-level statement, and the spec's protein examples use the `NP_` accession (e.g. `NG_007485.1(NP_478102.2):p.(...)`, as already present in the corpus). Pairing a coding-transcript (`NM_`) selector with a `p.` coordinate is at best non-preferred; the protein-accession (`NP_`) selector is the spec-appropriate form mutalyzer emits.

## Arbitration question

Decide whether ferro's `NG_(NM_):p.` output is:
- (b) spec-allowed / both-fine → keep as `accepted_divergence`, or
- improvement → ferro should emit the `NP_` selector for `p.` variants on genomic references (likely, on the spec evidence above), or
- (a) invalid → `known_bug`.

If the verdict is "converge to `NP_`", this becomes a normalization feature analogous to #500 (resolve the spec-appropriate accession from the input selector + reference data), and the corpus row migrates from `accepted_divergence` to `improvement` tracking this issue.

## Scope

- 1 corpus row today (`NG_012337.3(NM_003002.4):p.(Asp92Tyr)`); likely a small class once other `NG_(NM_):p.` inputs are considered.
[...]

PR #815, opened 2026-06-21
fix(data/cdot): drop frequently-wrong NM_→NP_ protein-accession inference

### LINKED ISSUE #808: fix(data/cdot): "frequently wrong" NM_→NP_ protein-accession inference used as fallback
[`cdot.rs#L429`](https://github.com/fulcrumgenomics/ferro-hgvs/blob/7f69ef21ea823c68d86b0f77035a708a66e8643a/src/data/cdot.rs#L429) falls back to number-preserving `NM_*` → `NP_*` protein-accession inference that the comment itself flags as `(frequently wrong)`.

**Expected:** drop the inference and decline when the protein accession is genuinely absent, rather than emit a likely-wrong `NP_`.

_Found via a code-comment audit; permalink pinned to audited commit `7f69ef2`._

nh13:
## Summary

`ferro` predicts a protein consequence (`p.`) by translating CDS bases, and that prediction needs a protein sequence identifier. When the transcript carried no authoritative protein accession, the fallback fabricated one by number-preserving substring substitution: `NM_000077` → `NP_000077`. RefSeq does **not** guarantee the NM and NP numbers match (e.g. `NM_000077.4` ↔ `NP_000068.1`), so — as the code comment itself flagged — this inference is "(frequently wrong)". It attached a real, translated prediction to a **non-existent** protein accession.

This PR removes the number-preserving `NM_*→NP_*` / `XM_*→XP_*` inference everywhere it was used as a user-facing fallback, declining to fabricate an accession while still satisfying the HGVS `p.` grammar (which requires a sequence identifier).

## What changed

- **`src/project/projector.rs`** (production `ferro project` / `ferro normalize` / Python bindings): collapse the protein-accession resolution chain to **authoritative cdot/reference value, else the transcript id itself**. The #310 grammar-required sequence-identifier fallback is preserved — only the wrong middle step is removed. A protein-less `NM_` transcript now emits `NM_…:p.(…)` rather than a fabricated `NP_…` — honest, because the identifier names the transcript whose CDS was actually translated.
- **`src/service/handlers/{effect,convert,vcf_convert}.rs`** (web-service handlers, position-based `p.`): prefer the authoritative `cdot_tx.protein`, else fall back to the transcript accession. In `vcf_convert`, the coding gate previously keyed off the *inferred* `NP_`/`XP_` prefix; it is reworked to key off the transcript's coding status (`NM_`/`XM_` or an authoritative protein present) so a coding transcript lacking an authoritative protein still emits a `p.` and a genuinely non-coding transcript still drops it.

## Out of scope (intentionally unchanged)

- **Benchmark comparison shims** (`benchmark.rs`, `benchmark/cache.rs`, `benchmark/biocommons.rs`) keep the substitution — they build like-for-like comparison keys against external tools that themselves infer `NP_`, and are not user-facing nomenclature.
- **`normalize/mod.rs::resolve_genomic_protein_selector`** (#502) already resolves only from the authoritative `Transcript.protein_id` and declines when absent — no number-preserving inference, so no change.

## Tests

- New projector regression tests: a protein-less `NM_`/`XM_` transcript emits the transcript id, never a fabricated `NP_`/`XP_`.
- New `vcf_convert` unit tests: authoritative protein preferred; coding `NM_` without protein emits transcript-id `p.`; non-coding transcript drops the `p.`.
- Updated the `#310` integration tests and Python projection tests that pinned the old inferred output to the transcript-id fallback (these assertions *were* the bug encoded as tests).
- `cargo nextest run --features dev`: 7475 passed. `cargo clippy --all-features --all-targets -- -D warnings`: clean. `pytest tests/python/`: 185 passed.

Closes #808

## Release Notes

* **Bug Fixes**
  * Removed protein accession inference: The system no longer fabricates protein identifiers by transforming transcript IDs. When no authoritative protein accession is available, the transcript ID itself is used as the protein accession in HGVS notation.

* **Documentation**
  * Updated schema and inline documentation to reflect the revised protein accession fallback behavior, clarifying that no automatic inference occurs.

* **Tests**
  * Updated test expectations to match the new protein accession handling; added coverage for transcript ID-based fallback scenarios.

### DIFF
--- a/docs/transcripts_json_schema.md
+++ b/docs/transcripts_json_schema.md
@@ -59,7 +59,7 @@ Each transcript object has the following fields:
 | `mane_status` | string | MANE designation: `"MANE_Select"`, `"MANE_Plus_Clinical"`, or omitted if not MANE |
 | `refseq_match` | string | For Ensembl transcripts: matched RefSeq accession |
 | `ensembl_match` | string | For RefSeq transcripts: matched Ensembl accession |
-| `protein_id` | string | Protein accession (e.g. `"NP_000079.2"` or `"MYGENE-protein.1"`). Used as the `p.` accession for c. → p. projection. When omitted, the projector infers `NP_*` from `NM_*` and `XP_*` from `XM_*`; if neither applies, it falls back to using the transcript ID itself as the `p.` accession so protein prediction is never silently dropped. |
+| `protein_id` | string | Protein accession (e.g. `"NP_000079.2"` or `"MYGENE-protein.1"`). Used as the `p.` accession for c. → p. projection. When omitted, the projector falls back to using the transcript ID itself as the `p.` accession so protein prediction is never silently dropped. It does **not** fabricate an `NP_*`/`XP_*` accession from an `NM_*`/`XM_*` transcript by preserving the number — RefSeq does not guarantee the `NM_` and `NP_` numbers match (e.g. `NM_000077.4` ↔ `NP_000068.1`), so that inference was frequently wrong and was removed (#808). |
 
 ## Exon Object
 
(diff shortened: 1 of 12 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #816, opened 2026-06-21
fix(equivalence/service): make unsupported variant classes explicit

### LINKED ISSUE #810: fix(equivalence/service): unsupported variant classes silently return None
Two spots silently return `None` for variant classes they don't handle, so unsupported inputs look like clean "no result" rather than an explicit unsupported signal:

- [`equivalence/checker.rs#L217`](https://github.com/fulcrumgenomics/ferro-hgvs/blob/7f69ef21ea823c68d86b0f77035a708a66e8643a/src/equivalence/checker.rs#L217) — `HgvsVariant::Allele(_) => return None, // Complex case, skip for now`, so allele equivalence checks silently resolve as unknown.
- [`service/tools/ferro.rs#L101`](https://github.com/fulcrumgenomics/ferro-hgvs/blob/7f69ef21ea823c68d86b0f77035a708a66e8643a/src/service/tools/ferro.rs#L101) — `_ => None, // For alleles, RNA fusions, etc. - skip for now`, so those variant classes get no detail output.

**Expected:** return an explicit `Unsupported` variant/error rather than a silent `None`, so callers can tell "not equivalent / no detail" apart from "not supported."

_Found via a code-comment audit; permalinks pinned to audited commit `7f69ef2`._

nh13:
Closes #810

Two spots silently returned `None` for variant classes they did not handle, so unsupported inputs looked like a clean "no result" rather than carrying an explicit signal. This addresses #810 at each named site according to what its `None` actually means, rather than mechanically converting every `None` to an error (which would regress correct behavior).

## `equivalence/checker.rs` — the `Allele` arm was already correct; the comment lied

`check_accession_version_difference` is a private predicate that answers exactly one question: are these two variants the same change on two different *accession versions*? An allele has no single accession version to compare (`accession()` reports only its first member, which would be misleading), so the helper declines with `None`. That `None` is "not-applicable", not "unsupported": `check()` falls through to the normalizer, which handles alleles directly (`normalize_allele`), so allele equivalence is fully determined — just not via the version-comparison path.

Converting this helper to error on alleles would make `check()` *error* on allele pairs that currently compare correctly — a real regression and exactly the "indeterminate ≠ unsupported" trap. So the only defect here was the misleading `// Complex case, skip for now` comment. Replaced it with an accurate rationale and added regression tests proving allele pairs flow through `check()` without erroring:

- two identical compound alleles → `Ok(Identical)`
- two differing compound alleles → `Ok(NotEquivalent)` (no error)

## `service/tools/ferro.rs` and `service/handlers/validate.rs` — eliminate accidental silence

`extract_variant_details` (two diverged copies of the same function) was dropping supported axes to `None`:

- `ferro.rs` dropped `Rna`, `Mt`, and `Circular` — all carry a `loc_edit` and can produce real, correct details.
- `validate.rs` dropped `Rna` (`Mt`/`Circular` arms already existed there).

Added the missing arms so these axes are populated (`r`/`m`/`o`), mirroring the existing `Cds`/`Genome`/`Tx`/`Protein` arms. Then replaced the catch-all `_ => None` with **explicit named arms** for the classes that genuinely have no single coordinate/position to flatten — `RnaFusion | GenomeRing | Supernumerary | Allele | NullAllele | UnknownAllele => None`. With no wildcard, a future `HgvsVariant` variant fails to compile here instead of being silently swallowed (registration-completeness guarantee).

The `validate.rs` sibling is the same function by copy and shares the latent silent-drop pattern, so the matching fix is applied there too — a sibling-pattern fix, not scope creep.

## Why not change the return types

Neither site changes its `Option`/`Result` contract. At the checker, returning an error would regress allele checks (above). At the service sites, switching to `Result` would ripple into `ParseResult`/`ValidateResponse`, the JSON/HTTP contract, and the web UI — disproportionate, and `None` = "no structured breakdown" is a legitimate value. `FerroError::UnsupportedVariant` exists but is not the right tool at either site.

## Tests

Added unit tests for the new arms (`r`/`m`/`o` populated; `Allele` returns `None`) and the checker regression tests above. Full suite (`cargo nextest run --features dev`, 7479 tests) green; `cargo clippy --all-features --all-targets -- -D warnings` clean.

## Release Notes

* **Tests**
  * Added regression tests for compound allele equivalence checking to ensure consistent variant comparison behavior.

* **Refactor**
  * Consolidated shared HGVS component extraction logic across validation and parsing modules for improved code maintainability and consistency. No changes to validation behavior or output.

### DIFF
--- a/src/equivalence/checker.rs
+++ b/src/equivalence/checker.rs
@@ -211,10 +211,17 @@ impl<P: ReferenceProvider> EquivalenceChecker<P> {
         str1: &str,
         str2: &str,
     ) -> Option<EquivalenceResult> {
-        // Get accessions (handle allele and null/unknown cases gracefully)
+        // This helper only detects whether two variants are the *same change on
+        // different accession versions*. `None` here means "not an accession-version
+        // difference" — it is not an "unsupported" signal. For compound/null/unknown
+        // alleles there is no single accession version to compare (an allele's
+        // `accession()` reports only its first member, which would be misleading), so
+        // we decline. The caller (`check`) then falls through to the normalizer, which
+        // handles alleles directly, so allele equivalence is fully determined — just
+        // not via this version-comparison path.
         let acc1 = match v1 {
             HgvsVariant::NullAllele | HgvsVariant::UnknownAllele => return None,
-            HgvsVariant::Allele(_) => return None, // Complex case, skip for now
+            HgvsVariant::Allele(_) => return None,
             _ => v1.accession()?,
         };
         let acc2 = match v2 {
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #228, opened 2026-05-16
vcf/annotate: extension misclassified as Nonsense due to `hgvs.contains("Ter")` substring heuristic


nh13:
Surfaced during the [PR #225](https://github.com/fulcrumgenomics/ferro-hgvs/pull/225) Opus review.

## Background

`src/vcf/annotate.rs::determine_consequence` (around line 220-235) classifies protein-level variant consequences by matching substrings against the `Debug` format of the `ProteinEdit` and against the HGVS string text:

```rust
if edit_str.contains("Frameshift") { return Consequence::Frameshift; }
if edit_str.contains("Nonsense") || hgvs.contains("Ter") {
    return Consequence::Nonsense;
}
if edit_str.contains("Extension") { return Consequence::StopLoss; }
// ... etc
```

The `hgvs.contains("Ter")` branch fires for **any** HGVS string containing the substring `Ter` — including `p.Ter315TyrextTer10` (a C-terminal extension / stop-loss variant), which contains `Ter` in both the start position `Ter315` and (post-#225) the extension count `extTer10`. The check fires before the Extension check is reached, so extensions are misclassified as `Nonsense` (`stop_gained`) rather than `StopLoss` (`stop_lost`).

## Reproducer

```rust
use ferro_hgvs::{parse_hgvs, HgvsVariant};
use ferro_hgvs::vcf::annotate::{determine_consequence, HgvsAnnotation, Consequence};
use ferro_hgvs::reference::ManeStatus;

let variant = parse_hgvs("NP_003997.1:p.Ter315TyrextTer10").unwrap();
let ann = HgvsAnnotation {
    variant,
    gene_symbol: None,
    transcript_accession: Some("NP_003997.1".to_string()),
    is_coding: true,
    is_intronic: false,
    mane_status: ManeStatus::None,
    intron_number: None,
    intron_position: None,
    intronic_consequence: None,
};
// Expected: Consequence::StopLoss
// Actual:   Consequence::Nonsense
assert_eq!(determine_consequence(&ann), Consequence::StopLoss);
```

The bug pre-dates #225: even with the pre-#225 Display form `p.Ter315Tyrext*10`, the start position `Ter315` already triggers the substring match. #225 (D2 Ter canonicalization) only makes the substring more pervasive by adding `extTer10` — it does not introduce the bug.

## Acceptance

[...]

PR #817, opened 2026-06-21
fix(vcf): classify coding indels by typed NaEdit and emit Frameshift

### LINKED ISSUE #803: fix(vcf): annotate never emits Frameshift; classifies indels via Debug-string matching
[`src/vcf/annotate.rs#L201-L218`](https://github.com/fulcrumgenomics/ferro-hgvs/blob/7f69ef21ea823c68d86b0f77035a708a66e8643a/src/vcf/annotate.rs#L201-L218) classifies coding indels by `format!("{:?}", edit).contains("Insertion"/"Deletion")` and unconditionally returns `InframeInsertion` / `InframeDeletion`. It never checks length mod 3, so **no variant is ever classified `Frameshift`**, and the Debug-string matching is the same fragile pattern fixed for extensions in #228.

> ```rust
> // For a simple approximation, single base indels cause frameshifts
> // A more complete implementation would analyze the exact length
> ```

**Expected:** compute the indel span length and emit `Frameshift` when not divisible by 3; match on the typed `Edit` enum rather than its `Debug` rendering.

_Found via a code-comment audit; permalink pinned to audited commit `7f69ef2`._

nh13:
## Problem

`determine_consequence` in `src/vcf/annotate.rs` classified coding (`HgvsVariant::Cds`) indels by rendering the edit to its `Debug` string and substring-matching. This had two defects, both verified against current code:

1. **`Consequence::Frameshift` was never emitted for coding indels.** The net indel length mod 3 was never checked, so every coding insertion became `InframeInsertion` and every coding deletion `InframeDeletion` — even a 1 bp or 2 bp frameshift was mislabeled in-frame.
2. **Classification matched on the `Edit` `Debug` rendering** — the same fragile, order-dependent anti-pattern that caused the prior protein-arm bug fixed in #228 (`p.Ter…ext…` misclassified via a `.contains("Ter")` substring heuristic).

## Fix

Rewrite only the `HgvsVariant::Cds(v)` arm to match the parsed `NaEdit` enum directly (via `v.loc_edit.edit.inner()`, the same pattern the protein arm uses post-#228), and split each indel category into frameshift vs in-frame using the crate-wide `crate::hgvs::variant::is_frameshift` rule (net indel length known, nonzero, and not divisible by 3).

- **Insertion / BreakpointInsertion** → `Frameshift` if `is_frameshift`, else `InframeInsertion`.
- **Deletion / NPaddedDeletion** → `Frameshift` if `is_frameshift`, else `InframeDeletion`.
- **Delins** → net length is `inserted_len - deleted_span`; `Frameshift` if `is_frameshift`, else `InframeDeletion` (a delins is a replacement, bucketed under deletion per the pre-existing convention).
- **Duplication / DupIns** → treated as an insertion of the copied span; `Frameshift` if `is_frameshift`, else `InframeInsertion`. (Previously `dup` fell through to `Unknown` — a strict improvement.)
- **Everything else** (substitution, inversion, identity, repeat, conversion, etc.) → `Unknown`, unchanged (no protein prediction available in this path).

This reuses the canonical, already-tested net-length logic (`get_indel_length`: deletion `-span`, insertion `+ins_len`, delins `ins_len - span`, duplication `+span`) so the VCF path and the rest of the crate share one definition of "frameshift". When the net length is non-deterministic — uncertain/ranged inserted length (`ins(10_20)`), named/reference insertion, N-padded deletion — `is_frameshift` is `false`, so the variant stays in its in-frame category rather than being asserted a frameshift. This is the conservative choice: we never claim a frameshift we cannot prove.

No `format!("{:?}", …).contains(…)` classification remains in the VCF consequence path.

## Tests

Added coding-indel tests (TDD; the 1 bp `c.100del` case fails against the old code, returning `InframeDeletion`):

- Insertions: 1/2/4 bp → `Frameshift`; 3/6 bp → `InframeInsertion`.
- Deletions: 1/2/4 bp → `Frameshift`; 3/6 bp → `InframeDeletion`.
- delins net-frame: `delinsA` over a 3 bp del (net −2), `delinsACG` over a 1 bp del (net +2), `delinsAC` (net +1), `c.100_103delinsACG` (net −1), `c.100_103delinsACGTAC` (net +2) → `Frameshift`; net 0 and net −3 → `InframeDeletion`.
- Duplication: `c.100dup` (1 bp) → `Frameshift`; `c.100_102dup` (3 bp) → `InframeInsertion`.
- Unknown-length insertion `ins(10_20)` → `InframeInsertion` (never `Frameshift`).
- Non-indel coding (`c.100A>G`, `c.100_105inv`) → `Unknown`.

All existing protein/UTR/intronic tests stay green.

## Verification

- `cargo nextest run --features dev` — 7478 passed.

## Follow-up (out of scope here)

`python_helpers::edit_type_from_debug` is a sibling `format!("{:?}", …).contains(…)` classifier with the same #228/#803 anti-pattern (and a latent ordering issue). It is `pub` but only exercised by its own tests — no production call site — and is not on the VCF consequence path, so it is left for a separate cleanup that replaces it with the typed `na_edit_type_str`.

Closes #803

## Release Notes

* **Bug Fixes**
  * Improved frameshift variant detection for duplicate insertions.
  * Enhanced consequence classification accuracy for coding indels.

* **Tests**
  * Extended test coverage for variant frameshift and in-frame behavior classification.

### DIFF
--- a/src/python_helpers.rs
+++ b/src/python_helpers.rs
@@ -386,6 +386,7 @@ fn compute_span(variant: &HgvsVariant) -> Option<i64> {
 /// - Insertion: +(inserted length), or None if length is unknowable
 /// - Delins: inserted_length - span, or None if inserted length is unknowable
 /// - Duplication: +(span)
+/// - DupIns: +(span + inserted_length), or None if inserted length is unknowable
 /// - Inversion: 0
 /// - Identity: 0
 ///
@@ -406,6 +407,13 @@ pub fn get_indel_length(variant: &HgvsVariant) -> Option<i64> {
         NaEdit::BreakpointInsertion { sequence } => inserted_sequence_len(sequence),
         NaEdit::Delins { sequence, .. } => {
             Some(inserted_sequence_len(sequence)? - compute_span(variant)?)
+        }
+        // DupIns is a duplication of the interval span followed by an
+        // insertion: net = duplicated span + inserted length, when both are
+        // deterministic (the inserted length is unknown for non-literal
+        // payloads, so `inserted_sequence_len` returns None there).
+        NaEdit::DupIns { sequence } => {
+            Some(compute_span(variant)? + inserted_sequence_len(sequence)?)
         }
         _ => None,
     }
@@ -461,6 +469,12 @@ pub fn get_indel_length_with_provider<P: ReferenceProvider + ?Sized>(
         NaEdit::BreakpointInsertion { sequence } => inserted_sequence_len(sequence),
         NaEdit::Delins { sequence, .. } => {
             Some(inserted_sequence_len(sequence)? - compute_span_with_provider(variant, provider)?)
+        }
+        // DupIns is a duplication of the interval span followed by an
+        // insertion: net = duplicated span + inserted length, when both are
+        // deterministic.
+        NaEdit::DupIns { sequence } => {
+            Some(compute_span_with_provider(variant, provider)? + inserted_sequence_len(sequence)?)
         }
         _ => None,
     }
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #820, opened 2026-06-21
fix(vcf): emit RefSeq NC_ accessions on the genomic HGVS axis

nh13:
Closes #804

## Problem

The VCF→HGVS converter built genomic accessions directly from the VCF chromosome name (`Accession::new("NC", chrom, …)`), producing spec-invalid genomic HGVS like `NC_1:g.…` (from `vcf_to_genomic_hgvs`) and `NC_chr1:g.…` (from `VcfToHgvsConverter::convert_genomic_only`) instead of build-appropriate RefSeq assembly accessions (`NC_000001.11` on GRCh38, `NC_000001.10` on GRCh37). The code carried a comment admitting the gap.

## Fix

Resolve the chromosome name plus the record's genome build to its RefSeq accession through the existing build-aware `ContigAliases::default_human` table (in `src/liftover/aliases.rs`) and the shared `parse_accession` helper (in `src/project/accession.rs`). A single new private helper, `genomic_accession_for(chrom, build)`, is used by both genomic call sites.

Key points:

- **Reuses existing infrastructure** — no new chromosome→accession table. `ContigAliases::default_human` already maps autosomes, X, Y, and MT under both builds with the correct per-chromosome NC versions, and `resolve_to_refseq` tolerates UCSC (`chr1`), Ensembl (`1`), and RefSeq (`NC_000001.11`) spellings. The cached singleton `default_human_aliases()` is promoted to `pub(crate)` so there is exactly one process-wide table.
- **Build context was already available** on `VcfRecord::genome_build` (default GRCh38), so no new plumbing was required; the build reaches both call sites via the existing `vcf` argument.
- **Mitochondrial**: `chrM`/`MT` resolve to the rCRS accession `NC_012920.1` (shared across builds) and are emitted on the `g.` axis. The `g.`→`m.` coordinate-system coercion is left to `normalize()`, consistent with how the rest of the crate treats mitochondrial genomic variants.
- **Unmapped contigs**: contigs the human alias table does not cover (unplaced/alt scaffolds, unknown names) now decline with an error rather than emit a malformed `NC_<name>` accession.

## Tests

[...]

Issue #805, opened 2026-06-21
tracking(vcf): HGVS→VCF converter completeness gaps


nh13:
Umbrella for several deferred shortcuts in the HGVS↔VCF conversion path found during a code-comment audit. Each can be split out into a scoped fix.

- [`from_hgvs.rs#L585`](https://github.com/fulcrumgenomics/ferro-hgvs/blob/7f69ef21ea823c68d86b0f77035a708a66e8643a/src/vcf/from_hgvs.rs#L585) — simplified 1→0-based coordinate mapping without proper validation: `// This is a simplification - in practice we'd need proper coordinate mapping`.
- [`from_hgvs.rs#L670-L685`](https://github.com/fulcrumgenomics/ferro-hgvs/blob/7f69ef21ea823c68d86b0f77035a708a66e8643a/src/vcf/from_hgvs.rs#L670-L685) — deletion/insertion conversion can't recover the anchor base when the reference is unavailable, with no fallback.
- [`from_hgvs.rs#L74`](https://github.com/fulcrumgenomics/ferro-hgvs/blob/7f69ef21ea823c68d86b0f77035a708a66e8643a/src/vcf/from_hgvs.rs#L74) — RNA variant → VCF unsupported (`"RNA variant conversion to VCF not yet supported"`).
- [`from_hgvs.rs#L115`](https://github.com/fulcrumgenomics/ferro-hgvs/blob/7f69ef21ea823c68d86b0f77035a708a66e8643a/src/vcf/from_hgvs.rs#L115) — complex / multi-variant alleles rejected rather than decomposed: `// For now, return an error suggesting decomposition`.
- [`parser.rs#L239-L255`](https://github.com/fulcrumgenomics/ferro-hgvs/blob/7f69ef21ea823c68d86b0f77035a708a66e8643a/src/vcf/parser.rs#L239-L255) — VCF parser only captures FORMAT keys; per-sample genotype values are skipped (`// For now, skip sample value extraction`).

_Found via a code-comment audit; permalinks pinned to audited commit `7f69ef2`._

nh13:
Related VCF-module issues from the same code-comment audit:
- #803 — `annotate` never emits `Frameshift` (correctness)
- #804 — bare chromosome names instead of RefSeq accessions (correctness)
- #813 — assertion-free converter unit tests (test-side gap)

nh13:
Closing — all five sub-tasks from the code-comment audit have landed and merged:

[...]

PR #821, opened 2026-06-21
fix(vcf): correct genomic anchor-base recovery and consolidate NC_→chrom mapping

nh13:
Part of #805 (genomic-conversion sub-items: coordinate mapping + anchor-base recovery; plus the reverse NC_→chrom consolidation flagged by #820). Remaining #805 sub-items (RNA→VCF, complex-allele decomposition, parser per-sample values) are out of scope here.

### Coordinate mapping in the converter reference fetch
`HgvsToVcfConverter::get_reference_base` / `get_reference_sequence` previously indexed the *transcript-relative* cached sequence with a *genomic* position, and on miss fell back to `provider.get_sequence(transcript.id, …)` — i.e. the transcript accession. Every caller (`build_vcf_record`, reached from the c./n./g./m. paths) passes genomic coordinates, so this read the wrong base for genomic positions and silently masked invalid inputs (`pos.saturating_sub(1)` mapped `pos == 0` to index 0).

The fetch now goes to the provider's genomic contig (keyed by the same `chrom` the VCF record is emitted under), mirroring the canonical `spdi/convert.rs` pattern: validate the 1-based interval (`start >= 1`, `end >= start`), convert to 0-based half-open `[start-1, end)`, try `get_genomic_sequence` then fall back to `get_sequence`, and length-check the result so a short provider read errors instead of truncating REF.

### Anchor-base recovery
With the fetch corrected, the deletion / insertion / delins / duplication arms recover the VCF anchor base from the provider when genomic data is available, and decline with a clear error (no silently wrong base) when it is not. The provider-less free function `genomic_hgvs_to_vcf` keeps its existing "anchor base required" decline (it has no provider to recover from); its doc now points callers to the converter for recovery, and the dead/shadowed bindings in its del/ins arms were tidied without changing which error message wins.

### Reverse NC_→chromosome consolidation
`from_hgvs.rs::accession_to_chromosome` and the web-service `vcf_convert.rs::extract_chromosome_from_accession` each hand-rolled their own `NC_`→chromosome ladder. Both now resolve the primary-assembly mapping through the shared, build-aware `ContigAliases::refseq_to_ucsc` table (`default_human_aliases()` is now `pub(crate)`), keeping the existing verbatim passthrough for non-primary contigs (`NC_06*`, `NZ_`, `NT_`, `NW_`).

## Tests
New unit tests cover anchor recovery for del & ins (correct REF/ALT/POS from a provider-seeded genomic contig), the clean decline when no reference is available, coordinate validation including an off-by-one boundary (`pos == 0`, `end < start`), and reverse `NC_`→chrom via the shared table (MT + an unmapped contig + non-primary passthrough). The `issue_390` coverage matrix cells that previously pinned the documented converter bug (`GenomicReferenceNotAvailable`) now assert successful provider-recovered conversions, with fixtures made genome-consistent. `cargo nextest run --features dev` (7480 passing) and `cargo clippy --features dev --all-targets -D warnings` are green.

## Note
There are sibling hand-rolled `NC_`↔chrom helpers in other service handlers (`service/handlers/convert.rs`, `service/handlers/liftover.rs`, plus `liftover::get_refseq_accession`) that were left untouched to keep this PR bounded; consolidating those is a candidate follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chromosome name resolution from RefSeq accessions using shared mapping tables.
  * Enhanced HGVS to VCF conversion with strict reference sequence validation and genomic range checking.
  * Stricter handling of deletion/insertion operations now requires reference data availability.
  * Better error messages for missing reference data in conversion operations.

* **Tests**
  * Added comprehensive tests for genomic anchor recovery and reference sequence validation.

### DIFF
--- a/src/service/handlers/vcf_convert.rs
+++ b/src/service/handlers/vcf_convert.rs
@@ -758,24 +758,18 @@ fn get_refseq_accession(chrom: &str, build: &GenomeBuild) -> Option<String> {
     }
 }
 
-/// Extract chromosome from RefSeq accession
+/// Extract the UCSC chromosome name from a RefSeq accession.
+///
+/// Resolves the primary-assembly `NC_` → UCSC mapping through the shared,
+/// build-aware [`ContigAliases`](crate::liftover::aliases::ContigAliases)
+/// reverse table rather than a hand-rolled accession ladder, so the forward and
+/// reverse directions share one source of truth. Accessions the table does not
+/// describe are returned unchanged.
 fn extract_chromosome_from_accession(accession: &str) -> String {
-    if accession.starts_with("NC_") {
-        if let Some(num_part) = accession.strip_prefix("NC_") {
-            if let Some(chrom_num) = num_part.split('.').next() {
-                if let Ok(num) = chrom_num.parse::<u32>() {
-                    return match num {
-                        1..=22 => format!("chr{}", num),
-                        23 => "chrX".to_string(),
-                        24 => "chrY".to_string(),
-                        12920 => "chrM".to_string(),
-                        _ => format!("chr{}", num),
-                    };
-                }
-            }
-        }
-    }
-    accession.to_string()
+    crate::liftover::aliases::default_human_aliases()
+        .refseq_to_ucsc(accession)
+        .map(|ucsc| ucsc.to_string())
+        .unwrap_or_else(|| accession.to_string())
 }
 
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

PR #822, opened 2026-06-21
feat(vcf): RNA→VCF conversion and complex-allele decomposition

nh13:
Part of #805 (sub-items: RNA→VCF + complex-allele decomposition). Stacked on #821.

Implements the two remaining `HgvsToVcfConverter::convert` arms in `src/vcf/from_hgvs.rs` that previously declined with "not yet supported".

## RNA (`r.`) → VCF

RNA variants now lower to VCF by reusing the existing coordinate mappers and `build_vcf_record` — no new mapping math. HGVS `r.` numbering mirrors the matching DNA reference:

- **Coding transcript** (`Transcript::is_coding()`): `r.` is CDS-relative (identical to `c.`), so each `RnaPos` lowers to the corresponding `CdsPos` and reuses `cds_to_genomic` (the same path as `convert_cds`).
- **Non-coding transcript**: `r.` is transcript-relative (identical to `n.`), so each `RnaPos` base lowers to a `TxPos` and reuses `tx_to_genomic` (the same path as `convert_tx`).

Genomic coordinates then feed the shared `build_vcf_record`, so REF/ALT/anchor-base handling (including the genomic anchor recovery #821 fixed) is identical to the c./n. paths.

### Shapes that decline (clear error, never wrong coordinates)

- Whole-entity RNA edits (`r.=`, `r.?`, `r.spl`, `r.0`) — caught by the `NaEdit::is_whole_entity()` guard *before* coordinate mapping. `r.=` in particular would otherwise hit `build_vcf_record`'s Identity arm and emit a spurious no-op REF=ALT record at the parser's placeholder position.
- Unknown edits (`Mu::Unknown`) — declined rather than panicking on `inner()`.
- `r.*N`/`r.-N` on a **non-coding** transcript — a non-coding transcript has no CDS frame, so a UTR position is meaningless; declined rather than silently dropping the `utr3` flag and mapping `*N` as transcript position `N` (mirrors the non-coding guard in `normalize::rna_pos_to_txpos`).

## Complex / multi-variant allele decomposition

A multi-member allele maps to multiple, independently-positioned VCF records, which a single `VcfConversionResult` cannot represent. New `convert_all` entry point returns one `VcfConversionResult` per member:

- scalar (non-allele) variant → 1-element vec;
- multi-member allele → one record per member, in declaration order;
- nested alleles are flattened recursively (`[[a;b];c]` → `[a, b, c]`);
- a member that cannot be lowered fails the whole call with a clear per-member error naming its 0-based index (no silent drop);
- empty allele rejected.

`convert` keeps its single-record contract: single-member alleles still convert (unchanged, `test_converter_single_variant_allele`), and multi-member alleles now error pointing at `convert_all` rather than silently dropping all but one member. No production caller relied on `convert` handling multi-member alleles (`HgvsToVcfConverter` is exercised only via the converter API/tests; the annotator uses the reverse direction).

## Out of scope

- VCF parser per-sample genotype values (`src/vcf/parser.rs`) — separate PR.
- Anything #821 already changed (genomic anchor recovery, `accession_to_chromosome`/`ContigAliases`).

## Tests

Added inline + integration coverage: RNA sub (coding & non-coding), SubstitutionNoRef provider read, del/ins anchor recovery, the `r.=`/`r.?`/`r.spl`/`r.0` declines, the non-coding-UTR decline, and `convert_all` 2-/3-member, single-member, scalar, nested-flatten, unsupported-member, and empty-allele cases. Repointed the integration test that asserted `r.` was unsupported. Full suite green (`cargo nextest run --features dev`); clippy clean (`--features dev --all-targets`).

## Release Notes

* **New Features**
  * Added support for converting HGVS RNA variants (`r.`) into VCF, including coding vs non-coding coordinate mapping.

* **Improvements**
  * Added `convert_all` to convert multi-member variants into separate VCF results with clear per-member error reporting.
  * Improved RNA-to-DNA normalization so `U` never appears in VCF alleles, including repeat-based edits.

* **Bug Fixes**
  * Tightened reference sequence fallback to only handle “not available” cases.

* **Tests**
  * Expanded integration and unit coverage for RNA conversion cases, allele parsing/round-trips, and plus-strand repeat `U`→`T` handling.

### DIFF
--- a/src/mave/parser.rs
+++ b/src/mave/parser.rs
@@ -350,16 +350,22 @@ mod tests {
     #[test]
     fn test_allele_variant() {
         let ctx = test_context();
-        // Alleles like c.[32T>C;39G>A] should work with context
-        let result = parse_mave_hgvs("c.[32T>C;39G>A]", &ctx);
-        // Note: this may fail depending on allele support, but we test the context injection
-        if let Err(e) = &result {
-            // Should not be a MissingAccession error
-            assert!(
-                !matches!(e, MaveParseError::MissingAccession { .. }),
-                "Should have injected accession"
-            );
-        }
+        // A short-form coding allele must parse with the context's coding
+        // accession (NM_000518.5) injected, yielding a full HGVS allele.
+        let variant = parse_mave_hgvs("c.[32T>C;39G>A]", &ctx)
+            .expect("short-form coding allele should parse");
+        // Assert the parsed structure directly (independent of Display): a
+        // two-member allele carrying the injected coding accession.
+        let HgvsVariant::Allele(allele) = &variant else {
+            panic!("expected an allele variant, got: {variant:?}");
+        };
+        assert_eq!(allele.variants.len(), 2);
+        assert_eq!(
+            variant.accession().map(|a| a.to_string()).as_deref(),
+            Some("NM_000518.5")
+        );
+        // And pin the exact deterministic compact-form round-trip.
+        assert_eq!(variant.to_string(), "NM_000518.5:c.[32T>C;39G>A]");
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #798, opened 2026-06-21
test(conformance): disposition bare-NP protein-framing rows as spec_overridden (ferro spec-correct)


nh13:
## Problem

A cluster of `protein_description` (and `coding_protein_descriptions`) rows fail **only** on the protein-reference framing: ferro emits the bare `NP_…:p.(…)`, while the corpus expects mutalyzer's parent-wrapped `NM_x(NP_y):p.(…)` / `NG_x(NP_y):p.(…)`. The `p.(…)` body is byte-identical.

ferro's bare-`NP_` form is **spec-correct** (verified during #780):
- HGVS nomenclature `recommendations/general.md` §147 renders protein consequences bare (`p.(Ser42Cys)`), §148 names the `NP_` when multiple isoforms exist. There is **no** `transcript(NP_):p.` wrapper anywhere in the spec.
- This is the basis of the existing `bare-np-protein` `spec_citation` cluster (#483/#495).

## Evidence (current corpus, ~15 rows in `protein_description` alone)

```
NM_000193.2:c.1388G>C                  exp_ref NM_000193.2(NP_000184.1)  got_ref NP_000184.1
NG_012337.1(NM_012459.2):c.129G>T      exp_ref NG_012337.1(NP_036591.2)  got_ref NP_036591.2
NG_008835.1(NM_022153.2):c.677-21_704+62del  exp_ref NG_008835.1(NP_071436.1)  got_ref NP_071436.1
NM_003002.4:n.206_210del               exp_ref NM_003002.4(NP_002993.1)  got_ref NP_002993.1
…(≈15 total; plus the CDKN2A NG_007485.1(NP_478102.2) rows on the coding_protein axis)
```

These were the residual blockers behind several rows attributed to #790/#791/#646.

## Proposal

Sweep the only-framing-different protein rows and disposition them `spec_overridden` under the `bare-np-protein` cluster (extending the established policy), demoting them from `baseline-failures/{protein_description,coding_protein_descriptions}.txt`. 0 behavior change.

**Exclude** the LRG case `LRG_1t1:c.266G>T` (expects `LRG_1p1`, the LRG protein reference — a separate LRG-protein-mapping question, not bare-NP).

Refs #780, #483/#495, #646.

PR #832, opened 2026-06-22
fix(reference): cover canonical-sequence reingestion through the provider (#791)

### LINKED ISSUE #791: fix(reference): canonical-sequence reingestion for wrong-sequence transcripts (NM_000193.2; #529 residual)
## Summary

`apply_canonical_overrides` corrects transcript **metadata** (CDS bounds, protein id) from an authoritative GenBank record, but it cannot fix a transcript whose **served sequence is the wrong sequence**. For `NM_000193.2` the prepared reference serves ~4650 nt where the canonical record is 1576 nt; the validator correctly reports `Correction::SequenceReingestionRequired` (src/reference/validate.rs:477, raised at :601), but the application path (`MultiFastaProvider::apply_canonical_overrides_to` → `apply_canonical_overrides`, src/reference/multi_fasta.rs:1457) only applies metadata corrections. `reconcile_cdot_with_overrides` can replace bases on read **only when the override carries the canonical `sequence`**, which the metadata-only `canonical_overrides.json` does not.

So `NM_000193.2` stays wrong even after `--validate-canonical` runs, and its `baseline-failures/protein_description.txt` rows (2) cannot demote.

## Why this is its own issue

This is the residual carved out of #529 ("run --validate-canonical on-host"). #529's metadata-correction path (#520/#523/#524/#527/#528) is complete and handles the `NM_012459.2`-class (right sequence, wrong metadata). `NM_000193.2` is the distinct **wrong-sequence** class that needs canonical-base re-ingestion, which no code path implements today.

## What's needed (decide in design)

Extend the canonical-overrides mechanism to carry/repoint the **canonical bases** for a `SequenceReingestionRequired` record, so the served sequence is replaced on read (the `reconcile_cdot_with_overrides` "override carries `sequence`" path already exists — the gap is that `build_canonical_overrides` / `fetch_canonical_overrides` do not populate `sequence` for these records). Alternatively, re-ingest the canonical FASTA for the affected accession into a supplemental/legacy source that takes precedence.

## Acceptance (sketch)

- `NM_000193.2` is served at its canonical 1576 nt (not ~4650 nt).
- The 2 `NM_000193.2` rows in `baseline-failures/protein_description.txt` demote; 0 regressions.

Refs #520, #529, #523, #524.

nh13:
## Summary

Investigation found that the canonical-sequence **replace-on-read mechanism #791 asks for already exists** (landed in #530, after #791 was filed): `from_genbank` populates `AuthoritativeRecord.sequence`, `apply_canonical_overrides` has the `Some(canonical)` branch that replaces wrong served bases and collapses exons, and `reconcile_cdot_with_overrides` reconciles cdot when the override carries a sequence even on length mismatch. The data side for `NM_000193.2` is likewise already solved — it is served at the canonical 1576 nt from the **supplemental FASTA** (the issue's "Alternative"), not the ~4650 nt main-FASTA record. So the wrong-sequence condition no longer reproduces on the current reference.

The genuine residual was **missing provider-level test coverage**: the mechanism was exercised only by isolated `validate.rs` unit tests. This PR adds provider-level tests in `src/reference/multi_fasta.rs` proving:
- a metadata-only override leaves the served FASTA bases byte-for-byte unchanged;
- a no-override transcript is unchanged;
- (the replace-on-read path is covered alongside the existing `Some(canonical)` mechanism).

## Scope note

The 2 `NM_000193.2` `protein_description` rows mentioned in the issue are a **bare-`NP_` framing** divergence (ferro emits spec-correct bare-NP), not a wrong-sequence problem — they are owned by #798 (PR #826), which dispositions them in its 13-row bare-NP sweep. This PR therefore does **not** touch `cases.json` / `baseline-failures` (that would duplicate and conflict with #826); it is test-coverage only.

Closes #791

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive regression test suite for canonical override functionality to verify sequence reingestion and metadata corrections. Tests ensure correct handling of canonical sequence replacements, metadata-only updates, unchanged default behavior, transcript normalization, and reconciliation edge cases including transcript length mismatches.

### DIFF
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -5033,6 +5033,157 @@ NC_000001.11\tRefSeq\tmatch\t9000\t9099\t100\t+\t.\tID=a1;Target=NG_008000.1 1 1
         assert_eq!(tx.protein_id.as_deref(), Some("NP_OV.2"));
     }
 
+    // --- canonical-sequence reingestion through the provider (issue #791) -----
+    //
+    // The metadata-only override class (`sequence: None`) is covered above; these
+    // exercise the **wrong-sequence** class end-to-end through the provider: an
+    // override that carries the canonical bases must cause `get_transcript` to
+    // serve those bases (not the wrong served FASTA sequence), and the cdot side
+    // must reconcile even when the served length disagrees with the
+    // authoritative length. The base-level behavior of `apply_canonical_overrides`
+    // is unit-tested in `validate.rs`; this proves the wiring at the provider
+    // chokepoint and the no-regression contract for normal / metadata-only paths.
+
+    #[test]
+    fn get_transcript_replaces_served_sequence_when_override_carries_canonical_bases() {
+        use crate::reference::authoritative::{AuthoritativeRecord, CanonicalOverrides};
+        // Served FASTA carries a WRONG, longer sequence (24 nt); the canonical
+        // record is 9 nt. The override carries the canonical bases, so the served
+        // sequence must be REPLACED on read with the 9 nt canonical sequence
+        // (the NM_000193.2 wrong-sequence class, scaled down).
+        let served_wrong = "ATGAAATAACCCGGGTTTAAACCCG"; // 25 nt, not canonical
+        let canonical = "ATGAAATAA"; // 9 nt canonical
+        let (mut provider, _kept) =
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #802, opened 2026-06-21
Provision NM_002001.2 — old transcript version whose mRNA is absent, blocking sibling-anchored derivation (#790 residual)


nh13:
The `mutalyzer-normalize` corpus pins **`NM_002001.2`** (2 rows), an old transcript version that the GRCh38 cdot lacks (cdot carries `.3`/`.4` only). After PR #790/#795 provisioned `NM_003002.2` via sibling-anchored exon→genome derivation, `NM_002001.2` is the remaining drifted corpus accession.

## Current behavior (NOT a silent bug)
`NM_002001.2` is correctly **declined** today via #785 — `TranscriptVersionNotExact { requested: "NM_002001.2" }` (infos axis) / `ferro-policy-785-version-substitution-refused` (normalized axis). It does not silently substitute a sibling version. So this is a *coverage* gap, not a correctness defect.

## Why #790's mechanism can't provision it as-is
The #790 sibling-anchoring derivation (`examples/derive_tx_placements.rs` → `derive_transcript_structure`) **readback-validates** by reconstructing the target version's spliced mRNA against the genome. That requires the target's mRNA bytes to be present in the prepared reference. They are **not** for `NM_002001.2`.

Verified against the canonical reference FASTAs (`transcripts/human.*.rna.fna.gz` + `supplemental/patterns_transcripts.fna`):

| accession | versions present in reference FASTAs |
|---|---|
| `NM_003002` (provisioned by #790) | `.1 .2 .3 .4` — `.2` present → derivation worked |
| `NM_002001` (this issue) | `.3 .4` only — **`.2` ABSENT** → derivation declines (nothing to readback) |

The RefSeq RNA release (`human.N.rna.fna.gz`) happens to retain `NM_003002`'s historical `.2` but not `NM_002001`'s. So `derive_tx_placements --accessions NM_002001.2` declines today (no `.2` sequence to validate against), and the default accession set (`DEFAULT_ACCESSIONS = "NM_003002.2"`) does not include it.

## What it takes to provision NM_002001.2
1. **Source `NM_002001.2`'s mRNA** (absent from the standard RefSeq RNA bundle) — e.g. `efetch -db nuccore -id NM_002001.2` — and ingest it into the prepared reference (supplemental FASTA), wired through `ferro prepare`.
2. Add `NM_002001.2` to the `derive_tx_placements` production accession set (`DEFAULT_ACCESSIONS` / however the producer is driven).
[...]

Issue #790, opened 2026-06-19
feat(reference): provision genomic exon structure for corpus-pinned old transcript versions (g.↔c. projection)


nh13:
## Summary

Corpus rows that pin an **old transcript version** (e.g. `NM_003002.2`, `NM_004992.3`, `NM_206933.2`) cannot be projected `g.↔c.` onto that pinned version, because ferro has the transcript's *spliced mRNA sequence* but not its **genomic exon structure** (an exon→genome map). The version is present only as a degenerate single-exon supplemental (whole sequence = one exon, no genomic placement) and is absent from cdot, so any projection that needs real exon boundaries — `g.→c.` mapping onto the pinned version, or resolving an intronic-offset position like `c.*824+10` / `c.378-17` — fails or diverges.

This is the residual that blocks a cluster of mutalyzer-corpus rows after #728 (which solved the *genomic-parent* side) and after the sequence-level supplemental provisioning (which already satisfies #672's exonic headline case). It is its own problem: provisioning the **structure**, not the sequence.

## Evidence (against the prepared reference)

`NM_003002.2` is present as a supplemental FASTA (length 1382) but only as a synthetic single exon (`get_supplemental_cds_info`, `src/reference/multi_fasta.rs`), and is not in cdot (only `.3`/`.4` are).

- `ferro project 'NG_012337.1:g.7125G>T' --axis c --transcript NM_003002.2` → `ERROR: Reference not found: NM_003002.2` (no projectable structure for the pinned version).
- `ferro project 'NG_012337.1:g.7125G>T' --axis c --transcript NM_003002.4` → `ERROR: variant does not overlap transcript NM_003002.4` (g.7125 falls in `.2`'s exon layout, not `.4`'s).
- `ferro normalize 'NM_003002.2:c.*824+10del'` → `c.*841del`; mutalyzer expects `c.*834del`. With no intron in the single-exon model, ferro folds `*824+10`→`*834` then 3′-shifts along the synthetic 3′UTR to the terminus. (This is the `tracking_issue: 672` row at `cases.json:2474`.)
- The purely-exonic 3′-shift case already works (`NM_003002.2:c.273del` → `c.274del`), confirming the gap is structural, not sequence.

## Why existing work doesn't cover it

[...]

PR #835, opened 2026-06-22
fix(reference): suppress derived-tx injection for the #790 producer

### LINKED ISSUE #800: fix(prepare): derive_tx_placements re-declines when the manifest already references its own output
## Problem

`examples/derive_tx_placements.rs` (the #790 build-time producer) builds its provider via `MultiFastaProvider::from_manifest`, which — once the manifest's `derived_transcript_placements` key is wired — **loads and injects the producer's own prior output** into the cdot mapper. On a re-run, the target accession (e.g. `NM_003002.2`) is then present in cdot (via the injected structure), which pollutes sibling selection / the candidate logic and makes the derivation **decline**:

```
$ cargo run --features dev --example derive_tx_placements -- --manifest <ref>/manifest.json --out <ref>/derived_transcript_placements.json
Loaded 1 derived transcript structures
Derived 0 structure(s) for 1 accession(s), 1 declined.
```

So regenerating the artifact in place (or a CI `--check` against a wired manifest) silently produces an empty result. The first generation (before the key is wired, or after removing the file) works correctly.

## Direction

- Have the producer ignore injected/derived structures when selecting siblings and enumerating candidates (only trust *real* cdot records), or load the provider with the `derived_transcript_placements` key suppressed; or document that regeneration requires the key be unset.
- Low severity (dev-tool only; runtime unaffected), but it makes `--check`/regeneration unreliable.

Refs #790, #795.

nh13:
## Summary

`examples/derive_tx_placements.rs` (the #790 build-time producer) builds its provider via `MultiFastaProvider::from_manifest`, which loads the manifest's `derived_transcript_placements` artifact and injects each structure into the cdot mapper (`inject_derived_tx_structures` → `cdot.add_transcript`). Once that key is wired to the producer's own prior output, the target accession (e.g. `NM_003002.2`) becomes present in cdot, so the producer's `has_transcript_exact(old_id)` guard fires and the derivation declines — silently emitting an **empty** artifact on an in-place re-run or a CI `--check` against a wired manifest. First generation (key not yet wired, or file removed) worked.

## Fix

Add `MultiFastaProvider::from_manifest_without_derived_tx`, which shares `from_manifest`'s body (refactored into a private `from_manifest_inner(path, inject_derived_tx)`) but skips **only** the `derived_transcript_placements` injection block; every other manifest key (cdot, genome, supplemental, NG_ placements, canonical overrides) loads identically. The producer uses it so it reasons over **real cdot records only** — its view is identical to the pre-wired state.

The runtime `from_manifest` path is **unchanged** (`inject_derived_tx = true`): projection still injects and uses derived placements. The change is scoped to the dev-only producer.

### Why the load-boundary seam

The injected old-version record flips the `has_transcript_exact(old_id)` guard true (the active bug today) and is a latent sibling-selection hazard if the artifact ever carries multiple sibling versions of one base. Suppressing at the provider-load boundary removes the pollution at the source and forecloses the latent hazard, without threading producer-specific knowledge (the `--accessions` target set) into the provider API.

## Reproduction (against a wired manifest)

Before:
```
Loaded 1 derived transcript structures
  skipping NM_003002.2: present in cdot (not an old/absent version)
Derived 0 structure(s) for 1 accession(s), 1 declined.
```

After:
```
Derived 1 structure(s) for 1 accession(s), 0 declined.
```
The derived `structures` are byte-identical to the committed artifact (only the path-suffixed `description` differs, as the current code already produces a path-free description).

## Tests

Adds a manifest-level regression test (`from_manifest_without_derived_tx_suppresses_injection`) asserting `from_manifest` injects the derived old version while `from_manifest_without_derived_tx` does not, with the real sibling record present in both cases. `cargo nextest run --features dev` green (7521 passed); `cargo clippy --features dev --all-targets -- -D warnings` clean.

This unblocks #802's re-derive step (in-place regeneration / `--check` on a wired manifest is now reliable).

Closes #800

## Summary by CodeRabbit

* **New Features**
  * Introduced option to load reference manifests while excluding previously derived transcript information, ensuring more accurate derivation results.

* **Tests**
  * Added regression test to validate correct behavior when loading manifests with and without derived transcript data.

### DIFF
--- a/examples/derive_tx_placements.rs
+++ b/examples/derive_tx_placements.rs
@@ -123,7 +123,12 @@ fn candidate_siblings<'a>(
 }
 
 fn run(cli: &Cli) -> anyhow::Result<ExitCode> {
-    let provider = MultiFastaProvider::from_manifest(&cli.manifest)
+    // Load WITHOUT injecting the manifest's `derived_transcript_placements`
+    // artifact (#800): once that key is wired to this producer's own prior output,
+    // injecting it would make the target accession present in cdot and the
+    // derivation would decline (skip) it, silently producing an empty artifact on
+    // a re-run or CI `--check`. The producer derives over real cdot records only.
+    let provider = MultiFastaProvider::from_manifest_without_derived_tx(&cli.manifest)
         .map_err(|e| anyhow::anyhow!("load manifest {}: {e}", cli.manifest.display()))?;
     let cdot = provider
         .cdot_mapper()
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -482,8 +482,49 @@ impl MultiFastaProvider {
         merged
     }
 
-    /// Create a provider from a manifest file
+    /// Create a provider from a manifest file.
+    ///
+    /// This is the standard (runtime) entry point: it loads every manifest key,
+    /// including injecting build-time–derived transcript structures
+    /// (`derived_transcript_placements`, #790) into the cdot mapper so projection
+    /// resolves cdot-absent old versions with real exons.
     pub fn from_manifest<P: AsRef<Path>>(manifest_path: P) -> Result<Self, FerroError> {
+        Self::from_manifest_inner(manifest_path, true)
+    }
+
+    /// Like [`Self::from_manifest`], but skips injecting the manifest's
+    /// `derived_transcript_placements` artifact into the cdot mapper.
+    ///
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #823, opened 2026-06-21
feat(vcf): parse per-sample FORMAT genotype values

nh13:
Part of #805 (sub-item: VCF parser per-sample genotype values).

## Problem

`src/vcf/parser.rs::convert_record` captured only FORMAT keys and skipped per-sample value extraction (`// FORMAT and samples - simplified: just capture format keys` / `// For now, skip sample value extraction`). Every parsed `VcfRecord` from a multi-sample VCF reported the FORMAT string but left each sample's map empty, so per-sample genotype data was silently lost and round-tripped through `Display` as all-`.` columns.

## Change

Populate `VcfRecord.samples` (already typed `Vec<HashMap<String, String>>`, and already consumed by `Display`) with one map per sample column, in sample-column order matching `sample_names()`, keyed by FORMAT field name. Values are rendered as canonical VCF text via noodles-vcf 0.73's `record_buf` samples API (`Samples::values()` -> `Sample`, zipping `keys()` with `values()`).

- Scalars use their natural string form; arrays are comma-joined with missing elements rendered `.`.
- Genotypes are rendered VCF 4.0/4.3-style (first allele position only; subsequent alleles prefixed by their phasing separator `|`/`/`; missing allele position `.`), mirroring noodles' own `write_genotype`. The project emits VCF 4.3.
- Missing field values (`.`) and fields absent from a short sample column are omitted from the map; `Display` already renders any absent FORMAT key as `.`, so omission round-trips correctly and keeps maps minimal.

No struct, field, or signature changes; no consumer changes (the only other `samples` constructors build empty vecs and nothing reads map contents).

## Tests

Added parser tests: single-sample field extraction, multi-sample column ordering with phased GT preserved, missing-value omission with `./.` genotype, ragged (short) sample columns, sites-only VCF still parses, and a parse->Display round-trip. Full suite green; clippy clean with `--features dev`.

## Summary by CodeRabbit

* **Bug Fixes**
  * VCF parser now correctly extracts and preserves per-sample FORMAT values
  * Improved handling of missing values in sample columns
  * Enhanced data integrity for VCF file display formatting

PR #841, opened 2026-06-22
fix(vcf): drop per-sample genotype when splitting multi-allelic records

nh13:
`split_multiallelic` previously cloned the source record into each bi-allelic output and copied the per-sample FORMAT/genotype verbatim. A multi-allelic genotype encodes its ALT alleles by index into the original `alternate` list (e.g. `1/2` references the first and second ALTs), so once the record is split so each output retains a single ALT, those indices no longer reference the right allele — each split record is left with a miscalled genotype (the wrong verbatim `1/2`).

Re-indexing genotypes correctly is non-trivial (and ambiguous for fields like AD/PL whose layout depends on the allele count), and no caller relies on split-record sample data, so the per-sample FORMAT/genotype is now dropped on split (`format = None`, `samples` empty) rather than propagated into each bi-allelic record. Also corrects the `samples` round-trip-fidelity doc comment to state that omitting a missing/absent FORMAT value is VCF-semantically equivalent but not byte-identical — a trailing-absent field from a ragged/short sample column is re-emitted by `Display` as an explicit `.` (e.g. input `1/1` under `GT:DP:GQ` renders back as `1/1:.:.`). Adds regression tests for both behaviors.

Follow-up to #823: this fix was prepared during the review of #823 but #823 merged before it was applied.

Refs #823

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed multi-allelic VCF splitting to exclude invalid per-sample genotype/FORMAT data in the generated bi-allelic records
  * Improved VCF rendering for ragged sample columns so trailing-absent fields display as explicit missing values (`.`)

* **Documentation**
  * Clarified how missing and ragged `FORMAT`/sample fields are handled during round-tripping and display

* **Tests**
  * Added coverage for dropping sample genotype data on split records and for correct rendering of ragged trailing fields

### DIFF
--- a/src/vcf/parser.rs
+++ b/src/vcf/parser.rs
@@ -254,8 +254,13 @@ fn convert_record(
     // One map per sample column, in sample-column order (matching the header's
     // `sample_names()`), keyed by FORMAT field name. A field whose value is
     // missing (`.`) or absent from a short sample column is omitted from the
-    // map; `VcfRecord`'s `Display` renders any absent FORMAT key as `.`, so the
-    // omission round-trips correctly and keeps the maps minimal.
+    // map, keeping the maps minimal. On output, `VcfRecord`'s `Display` walks
+    // every FORMAT key and re-materializes any omitted key as an explicit `.`
+    // placeholder. This is VCF-semantically equivalent but not byte-identical:
+    // an interior missing scalar round-trips verbatim (`./.:.:.`), whereas a
+    // trailing-absent field from a ragged/short sample column is re-emitted as
+    // a `.` rather than dropped (e.g. input `1/1` under `GT:DP:GQ` renders back
+    // as `1/1:.:.`).
     let samples: Vec<std::collections::HashMap<String, String>> = samples_buf
         .values()
         .map(|sample| {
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
Addressed the outside-diff CodeRabbit finding on `src/vcf/parser.rs` (`split_multiallelic`, per-allele INFO copied verbatim).

Chose the documentation option rather than subsetting: per-allele INFO (`Number=A`/`R`/`G`) can only be subset to the retained ALT with the header's `Number=` definitions, which the record-only `split_multiallelic(record: &VcfRecord)` signature does not have — and blindly dropping all INFO would discard valid scalar (`Number=1`) fields like `DP`. Extended the doc comment to state that per-allele INFO is carried verbatim and is unreliable in split output for per-allele keys (scalar fields stay correct), and added `test_split_multiallelic_carries_per_allele_info_verbatim` pinning that behavior (`AF=0.1,0.2` verbatim on both bi-allelic records, `DP=50` correct). A future change that subsets per-allele INFO should update that test.

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

Issue #811, opened 2026-06-21
test(normalize): file the 18 ignored cigar_aware_normalization cases as a tracked backlog


nh13:
[`tests/it/normalize_tests.rs` → `mod cigar_aware_normalization` (L4216+)](https://github.com/fulcrumgenomics/ferro-hgvs/blob/7f69ef21ea823c68d86b0f77035a708a66e8643a/tests/it/normalize_tests.rs#L4216) contains **18 `#[ignore]`d cases** that each assert the *desired* (correct) normalization behavior and will pass only once the underlying fixes land. They document known intronic-normalization failures across:

- CIGAR insertions in the transcript-to-genome alignment
- complex / multi-operation CIGARs
- transcript version mismatches
- LRG transcript mapping
- non-contiguous exons

This is the largest in-code-but-not-in-tracker backlog surfaced by the audit. Filing so the dormant cases are tracked and can be burned down (and so a sweep can confirm none silently start passing without being un-ignored).

_Found via a code-comment audit; permalink pinned to audited commit `7f69ef2`._

nh13:
Overlaps the cdot CIGAR-indel synthesis problem in #807 — several of these ignored cases (CIGAR insertions, non-contiguous exons) depend on that fix.

PR #844, opened 2026-06-24
fix(normalize): treat LRG transcript references as bare for EINTRONIC (#834)

### LINKED ISSUE #834: fix(normalize): LRG-addressed intronic normalization fails where NM-addressed equivalent succeeds
Surfaced while resolving #811 (PR #833).

7 LRG intronic-normalization cases decline with `ConversionError` (E5001) where the **NM-addressed equivalent succeeds** — an LRG→genomic mapping gap, not a coordinate/offset problem (the NM equivalents passing proves the normalization logic is correct; the LRG path can't resolve the genomic alignment).

Cases (asserted as the current decline in `tests/it/normalize_tests.rs::lrg_intronic_still_unsupported`, gated `#[ignore]` and run via `--ignored` against the prepared reference):
- `LRG_741t1:c.7007+30del`
- `LRG_763t1:c.52+1del`, `LRG_763t1:c.53-6del`
- `LRG_293t1:c.1-59_1-57del`
- `LRG_486t1:c.1000+5del`
- `LRG_720t1:c.1183+4dup`, `LRG_720t1:c.1327+4del` (these two lack any genomic alignment in the reference)

**Expected:** LRG-addressed intronic variants normalize equivalently to their NM-addressed counterparts (resolve the LRG→genomic mapping for the intronic path). When fixed, flip the `lrg_intronic_still_unsupported` assertions from decline to the normalized value and repoint the `#[ignore]` reasons.

Refs #811/#833. (Note: the last two may need genomic-alignment provisioning for those LRGs — distinct from the mapping-path gap.)

nh13:
Closes #834. Context: surfaced while resolving #811 (PR #833, merged); parent #811/#833 introduced the `lrg_intronic_still_unsupported` decline-bucket this PR resolves.

## Root cause

An LRG transcript reference (`LRG_<N>t<k>:c.…` / `:n.…`) is a **bare transcript** with no `NG_`/`NC_` genomic context — semantically identical to a bare `NM_`/`XM_` (`c.`) or `NR_`/`XR_` (`n.`) reference. The HGVS spec rule "a (non-)coding DNA reference sequence does not contain introns" applies to it equally. ferro models that rule as the EINTRONIC / W4007 (`IntronicOnBareTranscript`) path.

The classifier `intronic_on_bare_transcript_warning` (`src/normalize/mod.rs`) recognized only RefSeq prefixes (`NM`/`XM` on `c.`, `NR`/`XR` on `n.`) — its doc comment explicitly listed LRG as "out of scope." So an LRG-addressed intronic variant was denied the bare-transcript handling.

Why that produced the divergence: in the default **lenient** mode, `normalize_core` recovers a bare-transcript intronic variant whose genomic intronic conversion declines by **echoing the input** with W4007 attached. Because LRG was excluded, the underlying `ConversionError` (E5001) propagated for an LRG input while the **identical NM-addressed variant echoed its input** via the recovery. That asymmetry — not any genuine LRG→genomic mapping gap — is the entire bug. For the 7 reported cases, the genomic intronic conversion (`cds_to_genomic_with_intron` → `intronic_to_genomic`) declines for **both** the LRG and the NM accession; only the NM accession got the recovery.

## Fix

Extend the classifier to treat LRG transcript accessions as bare coding (`c.`) and non-coding (`n.`) references via `Accession::is_lrg()`, guarded by the same `genomic_context.is_none()` condition. LRG now behaves identically to its NM equivalent in every error mode:

- genuinely resolvable LRG intronic variants still 3'-shift / re-anchor (e.g. `LRG_311t1:c.80+1del` → `LRG_311t1:c.79+1del`, mirroring `NM_000314.4:c.80+1del` → `c.79+1del`) — the W4007 recovery only fires on a dispatch error, so these are untouched;
- LRG intronic variants whose genomic conversion declines echo the input under the lenient W4007 recovery, mirroring the NM equivalent;
- strict mode rejects an LRG bare intronic as EINTRONIC, symmetric with NM (this was already NM's strict-mode behavior).

## Scope: deviation from the issue's 5-fixed / 2-deferred split

The issue proposed fixing 5 cases (a "pure LRG→genomic mapping-path gap") and deferring 2 (`LRG_720t1`, said to "lack any genomic alignment", blocked on reference provisioning). **Careful verification against the prepared reference shows that split does not hold**, and the issue itself asked me to report rather than force it if so:

- The fix mechanism is the bare-transcript EINTRONIC recovery, **not** genomic-alignment resolution. Whether a case has a genomic alignment is irrelevant to whether the LRG path matches the NM path — the NM path of all 7 cases also fails genomic resolution and only "passes" via the same recovery.
- `LRG_720t1 → NM_000280.3`: the NM equivalent **also lacks a genomic alignment** (`chromosome=None`, synthetic single exon) and likewise "passes" only via the W4007 recovery. So the 2 `LRG_720t1` cases are indistinguishable from the other 5 under this fix.
- Forcing the 2 to remain declining would require **suppressing** the recovery for `LRG_720t1` while `NM_000280.3` keeps it — re-introducing the exact LRG-vs-NM asymmetry #834 exists to eliminate, with no spec basis.

Therefore the correct root-cause fix resolves **all 7 cases uniformly**. The scratch-SSD-offline provisioning blocker cited for the 2 is moot because the fix does not depend on genomic alignment. All 7 LRG cases move from `lrg_intronic_still_unsupported` into `lrg_intronic_normalization`, asserting identity (the NM-addressed output is the independent oracle) plus idempotency. The now-empty decline module and its unused `normalize_expect_err` helper are removed.

## Tests

- 3 new non-reference unit tests in `eintronic_helper_tests` pinning the LRG classifier on both axes (intronic → warn, minus-offset → warn, exonic → no-warn).
- 7 reference-gated cases flipped from decline to identity in `lrg_intronic_normalization::test_lrg_intronic_normalizes_ok` (16 cases total, all `#[ignore]`d; run with `--run-ignored all` + `FERRO_MANIFEST`).
- `cargo clippy --features dev --all-targets -- -D warnings`: clean.
- Targeted suites (eintronic / issue_486 / lrg / intronic / cigar / cdot_gap / version_mismatch) with the prepared reference: 532 passed.
- Full `cargo nextest run --features dev --run-ignored all`: the only failures are 37 pre-existing, environment-dependent external-oracle comparison tests (biocommons/UTA/docker, gnomad, mutalyzer/hgvs-rs) — identical on the unmodified base (verified). Zero new failures; no LRG cases appear in any baseline-failures ledger or corpus, so no ledger regeneration was needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of intronic LRG transcript variants so valid `c.` and `n.` forms are now recognized consistently.
  * Expanded support for LRG coding and non-coding transcript references, including cases with `+`/`-` offsets and unknown positions.
  * Reduced false warnings for exon-only LRG variants, which now normalize without the intronic warning.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -600,14 +600,17 @@ fn boundary_has_intronic<T>(
 /// reference** (no `NG_(…)`/`NC_(…)` genomic context). Returns the warning
 /// to emit, or `None`.
 ///
-/// In scope: a bare coding transcript (`NM_`/`XM_`) used with `c.` and a bare
-/// non-coding transcript (`NR_`/`XR_`, via `is_noncoding_rna()`) used with
-/// `n.`, with `Accession.genomic_context == None`. The spec's "a (non-)coding
-/// DNA reference sequence does not contain introns" rule applies equally to
-/// curated (`NM_`/`NR_`) and predicted (`XM_`/`XR_`) transcripts, so both are
-/// covered on each axis. Out of scope: genomic-context forms
-/// (`genomic_context: Some`), `NG_`/`NC_` references (which never reach the
-/// c./n. transcript path), LRG, Ensembl `ENST`, and the r. axis. Both `Single`
+/// In scope: a bare coding transcript (`NM_`/`XM_` or LRG `LRG_<N>t<k>`) used
+/// with `c.` and a bare non-coding transcript (`NR_`/`XR_` via
+/// `is_noncoding_rna()`, or LRG) used with `n.`, with
+/// `Accession.genomic_context == None`. The spec's "a (non-)coding DNA
+/// reference sequence does not contain introns" rule applies equally to curated
+/// (`NM_`/`NR_`), predicted (`XM_`/`XR_`), and LRG transcript references — an
+/// LRG transcript is itself a bare reference with no `NG_`/`NC_` genomic
+/// context — so all are covered on each axis (#834). Out of scope:
+/// genomic-context forms (`genomic_context: Some`), `NG_`/`NC_` references
+/// (which never reach the c./n. transcript path), Ensembl `ENST`, and the r.
+/// axis. Both `Single`
 /// and `Range` (uncertain-breakpoint) position boundaries are inspected; an
 /// unknown (`?`) offset still counts as intronic (`CdsPos::is_intronic` treats
 /// the unknown-offset sentinel as intronic), which is correct — it is an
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai full review

Issue #807, opened 2026-06-21
fix(reference): served transcript silently diverges from deposited sequence on unapplied cdot indels


nh13:
[`multi_fasta.rs#L1440-L1450`](https://github.com/fulcrumgenomics/ferro-hgvs/blob/7f69ef21ea823c68d86b0f77035a708a66e8643a/src/reference/multi_fasta.rs#L1440-L1450) (`CdotSynthesisRisk::UnappliedIndels`): when cdot carries per-exon CIGARs with indels, the genome-derived synthesized transcript provably diverges in length and content from the GenBank-deposited transcript — but it is still served (only a `warn!` is emitted). Same family of synthesis caveat already tracked in #471 / #400.

The comment is explicit that the bases "assume an exact genome match and cannot represent transcript-specific indels or substitutions."

**Expected:** apply cdot CIGAR indels during synthesis, or decline (serve nothing on that axis) rather than serve a knowingly-divergent sequence.

_Found via a code-comment audit; permalink pinned to audited commit `7f69ef2`._

nh13:
The ignored `cigar_aware_normalization` cases in #811 exercise this from the normalization side (CIGAR insertions / non-contiguous exons), and will only pass once cdot CIGAR indels are applied.

PR #848, opened 2026-06-25
feat(prepare): version-aware transcript backfill (#842)

### LINKED ISSUE #842: prepare: backfill superseded transcript accession.versions so synthesis isn't needed for version-skew
## Problem

`ferro prepare` populates the transcript FASTA from NCBI's **bulk** RefSeq RNA release files (`human.N.rna.fna.gz`, `src/prepare/mod.rs:262-264`). That feed is a *current-snapshot* product: it ships essentially one version per accession. cdot metadata, fetched separately (`src/prepare/mod.rs:656`), carries exon alignments for **many historical versions**.

When a query names a superseded version (e.g. `NM_000088.3`) whose bases are not in the prepared FASTA but whose alignment *is* in cdot, `MultiFastaProvider::get_transcript` falls back to `synthesize_transcript_from_cdot` (`src/reference/multi_fasta.rs:2089-2116`): it reconstructs the bases from the **genome** FASTA via the cdot exon CIGAR rather than from the deposited transcript sequence.

That reconstruction is fundamentally lossy. The cdot GFF3 Gap CIGAR encodes only `M`/`I`/`D`, never substitutions, so:

- **Insertions** are declined (`TranscriptSequenceUnreconstructable`, E2005 — #807).
- **Deletions** are applied (#807).
- **Substitutions** are *undetectable* from genome + CIGAR alone — an aligned `M` base can differ from the deposited transcript at a substitution site with no structural signal. We can only warn, never correct.

The root cause is purely a reference-coverage gap: the authoritative sequence for the requested `accession.version` exists at NCBI and is retrievable, it just isn't in the prepared reference.

## Proposed fix

Have `ferro prepare` fetch the deposited sequence for any `accession.version` that is referenced by cdot but absent from the bulk RNA feed, and add it to the prepared transcript FASTA. The per-accession GenBank fetch path added for NG-hosted transcripts (`src/prepare/mod.rs:858`, behind the `benchmark` feature) is the same mechanism — generalize it to a version-aware backfill:

1. Diff the set of `accession.version` keys present in cdot against those present in the transcript FASTA.
2. For the missing ones (typically superseded versions), fetch each by `accession.version` (NCBI efetch / equivalent) and append to the prepared FASTA + reindex.
3. Cache so re-prepare is incremental.

With the authoritative sequence present, the primary FASTA path serves it directly and synthesis is never invoked for those accessions — eliminating the substitution caveat entirely for version-skew cases.

## Scope / non-goals

- Genome + CIGAR synthesis stays as the last-resort fallback for accessions genuinely unavailable from any source.
- This does not change query-time behavior (ferro stays offline at query time); it's a prepare-time data-acquisition improvement.

Related: #807 (apply CIGAR deletions / decline on insertions), #471, #400 (synthesis-divergence caveats).

nh13:
## Summary

`ferro prepare` populates the transcript FASTA from NCBI's **bulk** RefSeq RNA release (`human.N.rna.fna.gz`), a current-snapshot product carrying ~one version per accession, while cdot (downloaded separately) carries exon alignments for **many historical versions**. When a query names a superseded version (e.g. `NM_000088.3`) whose bases are absent from the prepared FASTA but whose alignment is in cdot, `MultiFastaProvider` falls back to `synthesize_transcript_from_cdot`, reconstructing bases from the genome via the cdot exon CIGAR. That path is lossy: insertions decline (#807), deletions apply, and **substitutions are undetectable** (genome + CIGAR cannot see a deposited-transcript substitution). The root cause is a reference-coverage gap — the authoritative `accession.version` sequence exists at NCBI, it just isn't in the prepared reference.

This PR adds an opt-in, prepare-time **version-aware backfill** that closes the gap:

```
ferro prepare --backfill-transcripts targets.txt
```

reads a newline-delimited `accession.version` list, diffs it against the cdot keys (`CdotMapper::transcript_ids`) and the transcript-FASTA `.fai` keys (`load_available_versions`), fetches the genuine gap (present in cdot, absent from the FASTA) from NCBI EFetch, appends it to `backfill/backfill_transcripts.fna`, reindexes, and wires a new manifest field (`backfill_transcripts_fasta`) that the provider ingests. The backfilled `accession.version` is then served by the **primary index path** with version-exact cdot CDS — the lossy genome+CIGAR synthesis is never invoked for it. Synthesis remains the last-resort fallback for accessions unavailable anywhere, and query time stays fully offline.

## Design

- **Injectable fetcher boundary** (mirrors `authoritative::build_canonical_overrides`): `backfill::execute_backfill` takes `F: FnMut(&str) -> Option<String>`, so the diff/append/manifest logic is unit-tested with a mocked fetcher and **CI stays network-free**. The production fetcher (`fetch_transcript_fasta`) curls EFetch `rettype=fasta` with `--fail` + connect/total timeouts and a 100 ms NCBI rate-limit.
- **Two-phase**: pure `plan_backfill` (diff → `to_fetch` / `already_present` / `not_in_cdot`, sorted + deduped) then `execute_backfill` (lazy file creation so a fully-failed/empty run leaves no orphan 0-byte FASTA; appends `>accession\n`-headed records wrapped at 70 cols).
- **Dedicated `backfill/` directory**, not `supplemental/`, so the provider's whole-directory scan does not silently fold `legacy_genbank.fna` / benchmark-gated supplemental FASTAs into the primary index.
- **Incremental cache**: the present-set folds in the prior backfill `.fai`, so a re-run skips already-fetched accessions; `--force` discards the prior output and re-fetches without duplicate records.
- **Pre-flight guard** (mirrors `--derive-ng-placements`): requires cdot — downloaded this run *or* already present in the reference (enables cheap incremental backfill against a prepared reference).
- **Continue-on-failure**: a per-accession fetch/parse/version-mismatch failure warns and continues; only output-file I/O errors abort.

## Test coverage (all network-free, mocked fetcher)

- `plan_backfill`: exact-missing diff; `not_in_cdot` routing; incremental present-wins; sort/dedup determinism.
- `parse_fasta_record`: sequence extraction; header-accession **version-mismatch** rejection (efetch resolving bare→current); empty-sequence rejection; stop-at-next-record (no multi-record fold).
- `execute_backfill`: append with exact headers; continue-on-failure; version-mismatch → not written; incremental append without clobber; **no orphan file** when all fetches fail; empty plan no-ops.
- manifest: `backfill_transcripts_fasta` relative/absolute round-trip; pre-#842 manifest deserializes (`None`).
- provider: a manifest naming the backfill FASTA serves the accession from the index.
- prepare/CLI: config default `None`; pre-flight guard rejects a cdot-less selection.

`cargo build --features dev`/`--features benchmark`, `cargo nextest run --features dev` (7555 passed), `cargo clippy --features dev --all-targets -- -D warnings`, and `cargo fmt --all --check` are all clean. The generated spec fixture is untouched.

## Downstream

The mechanism is drivable for a specific target accession set, so **#802** (stacked on this PR) can provision `NM_002001.2` — a superseded version present in cdot but absent from the bulk feed — by passing a one-line targets file.

Closes #842

## Summary by CodeRabbit

* **New Features**
  * Added `ferro prepare --backfill-transcripts <PATH>` to fetch missing exact `accession.version` transcript sequences and write a backfill FASTA.
  * Persisted the backfill FASTA path in the reference manifest and enabled downstream transcript lookup to index/serve it when present (including from Python).
* **Bug Fixes**
  * Transcript lookups now prefer backfilled exact-version FASTA records instead of falling back to less precise behavior.
  * Check summaries now report backfilled transcripts when available.

### DIFF
--- a/src/bin/benchmark.rs
+++ b/src/bin/benchmark.rs
@@ -1515,6 +1515,7 @@ fn run_prepare_tool(
                 patterns_file: None,
                 validate_canonical_accessions: None,
                 derive_ng_placements: None,
+                backfill_transcripts: None,
                 genome: match genome {
                     GenomeOption::None => "none".to_string(),
                     GenomeOption::Grch37 => "grch37".to_string(),
--- a/src/bin/ferro.rs
+++ b/src/bin/ferro.rs
@@ -542,6 +542,19 @@ enum Commands {
         #[arg(long)]
         derive_ng_placements: Option<PathBuf>,
 
+        /// Newline-delimited file of exact `accession.version`s (e.g.
+        /// `NM_002001.2`) to backfill at prepare time (#842): for each that is
+        /// present in cdot but absent from the bulk RefSeq RNA FASTA, fetch its
+        /// deposited sequence and append it to `backfill/backfill_transcripts.fna`,
+        /// wiring the manifest field so the primary path serves it instead of
+        /// synthesizing lossy bases from the genome. Requires cdot (downloaded
+        /// in this run or already in the reference). Networked (NCBI EFetch per
+        /// accession); per-accession failures warn and continue.
+        ///
+        /// Example: `ferro prepare --backfill-transcripts targets.txt`
+        #[arg(long)]
+        backfill_transcripts: Option<PathBuf>,
+
         /// Dry run - show what would be downloaded without downloading
         #[arg(long)]
         dry_run: bool,
@@ -871,6 +884,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             validate_canonical,
             ensembl,
             derive_ng_placements,
+            backfill_transcripts,
             dry_run,
         } => run_prepare(
             &output_dir,
@@ -886,6 +900,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             validate_canonical.as_deref(),
             ensembl,
             derive_ng_placements,
+            backfill_transcripts,
             dry_run,
         ),
         Commands::Check {
@@ -3238,6 +3253,7 @@ fn run_prepare(
     validate_canonical: Option<&Path>,
     ensembl: bool,
     derive_ng_placements: Option<PathBuf>,
+    backfill_transcripts: Option<PathBuf>,
     dry_run: bool,
 ) -> Result<(), Box<dyn std::error::Error>> {
     use ferro_hgvs::prepare::{prepare_references, PrepareConfig};
@@ -3258,6 +3274,7 @@ fn run_prepare(
         patterns_file: patterns.map(|p| p.to_path_buf()),
         validate_canonical_accessions: validate_canonical.map(|p| p.to_path_buf()),
         derive_ng_placements,
+        backfill_transcripts,
         genome: genome.to_string(),
         dry_run,
     };
--- a/src/check/mod.rs
+++ b/src/check/mod.rs
@@ -317,6 +317,10 @@ pub fn print_check_summary(result: &CheckResult, reference_dir: &Path) {
         eprintln!("  Supplemental transcripts: {}", supp.display());
     }
 
+    if let Some(ref backfill) = manifest.backfill_transcripts_fasta {
+        eprintln!("  Backfilled transcripts: {}", backfill.display());
+    }
+
     if let Some(ref legacy) = manifest.legacy_transcripts_fasta {
         eprintln!("  Legacy transcripts: {}", legacy.display());
     }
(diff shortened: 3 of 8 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

Issue #891, opened 2026-07-01
test(conformance): re-derive protein_description for NM_000143.3 insCAT rows after #882 dup→ins correction


nh13:
Follow-up from #882.

In fixing #882, the `mutalyzer-normalize` corpus rows `NM_000143.3:c.-1_1insCAT` and `NM_000143.3:c.1_2insCAT` had their `normalized` value corrected from the stale Mutalyzer2 value `c.-1_2dup` to the (spec- and mutalyzer-3.1.1-correct) insertion form.

Their `protein_description` was `NP_000134.2:p.(Met1?)`, which was derived from the **old** `c.-1_2dup` — whose duplicated copy was inserted between c.2 and c.3, splitting the ATG initiation codon (hence `Met1?`). The corrected insertion forms differ and likely have **different** protein consequences:

- `c.-1_1insCAT` inserts CAT entirely 5' of c.1, leaving the ATG intact — plausibly `p.(=)` (the old note recorded mutalyzer computing `p.(=)` for this input).
- `c.1_2insCAT` inserts CAT between c.1 (A) and c.2 (T), disrupting the ATG — plausibly `p.?`/`p.(Met1?)`.

So retaining `p.(Met1?)` on both rows is no longer verified. The `protein_description` axis for these two rows is deferred via a `known_bug` annotation pointing here until the values are re-derived from the corrected insertion consequence (run the protein axis with `FERRO_MANIFEST` set, cross-check against mutalyzer 3.1.1, and re-bless).

Scope: two corpus rows; `protein_description` axis only. The `normalized` axis is already correct.

Issue #890, opened 2026-07-01
chore(benchmark): bump comparator pins (hgvs-rs, mutalyzer-hgvs-parser, biocommons hgvs) and record versions in corpus provenance


nh13:
## Motivation

Fixing #882 surfaced that the `mutalyzer-normalize` conformance corpus carried **Mutalyzer2-era** expected values that current mutalyzer 3.1.1 no longer produces — e.g. `NM_000143.3:c.-1_1insCAT` was recorded as `→ c.-1_2dup` (an `M2:`-tagged value), but mutalyzer 3.1.1 returns `c.-1_1insCAT` (confirmed against mutalyzer.nl and the local `mutalyzer_normalizer` 3.1.1).

A full sweep of all 246 corpus rows carrying a mutalyzer expectation against live mutalyzer 3.1.1 found the corpus is **96% faithful** (236 match); the only genuine drift is the 2 `insCAT` rows (fixed in #882). The rest of the apparent "drift" is either documented ferro-vs-mutalyzer divergences (correct) or dormant `to_test:false` rows. So the corpus itself is healthy — but the root cause is a **provenance gap**: expected values were imported from upstream mutalyzer's test-suite repo (`source_commit 00045d63`, partly M2-era) rather than **derived from the pinned comparator**, and the comparator runtime versions are not recorded in the corpus header.

Meanwhile several benchmark comparators are behind latest.

## Version status (2026-06-30)

| comparator | repo pin / recorded | latest | action |
|---|---|---|---|
| mutalyzer | `==3.1.1` (pixi.toml) | 3.1.1 | current |
| mutalyzer-hgvs-parser | `==0.3.9` (pixi.toml) | 0.4.0 | bump |
| biocommons hgvs | `>=1.5.6,<2`; 1.5.6 recorded | 1.5.7 | refresh + record |
| hgvs-rs | 0.19.1 (BENCHMARK_GUIDE) | 0.22.0 | bump |

## Tasks

1. Bump comparator pins to latest: `mutalyzer-hgvs-parser 0.4.0`, `hgvs-rs 0.22.0`, refresh `biocommons hgvs` to 1.5.7; rebuild the benchmark env. (mutalyzer already 3.1.1.)
2. Re-run the parse/normalize comparisons and note any expected-value shifts introduced by the upgrades.
[...]

PR #892, opened 2026-07-01
fix(normalize): keep out-of-phase insertions as ins, not spurious dup/repeat (#882)

### LINKED ISSUE #882: fix(normalize): out-of-phase insertion mis-converted to duplication (changes the variant)
## Summary

`normalize` converts an **out-of-phase** insertion into a tandem duplication, producing a **different variant** — the dup form decodes to a different sequence than the input insertion.

Surfaced during #864 (normalization idempotency) while cross-checking against mutalyzer. This is distinct from #864's 3'-slide fix: that fixes the *position* of a legitimate dup; this is a false dup-*detection*. It is pre-existing (reproduces on `main`, unchanged by #864).

## Reproducer (real locus, GRCh38)

Reference `NC_000001.11` g.20004502–20004514 = `GACACACATAGGT` — an `ACACACA` run at g.20004503–20004509, flanked by `G` (5') and `T` (3').

| input | ferro | mutalyzer | correct |
|---|---|---|---|
| `NC_000001.11:g.20004503_20004504insAC` | `g.20004508_20004509dup` ❌ | `g.20004503_20004504insAC` | mutalyzer |
| `NC_000001.11:g.20004503_20004504insCA` | `g.20004508_20004509dup` ✅ | `g.20004508_20004509dup` | both |

Both insert at the same point, but:
- `insCA` → `GACACAC**CA**ATAGGT`… stays alternating → a genuine tandem dup of the `AC`/`CA` unit. ✅
- `insAC` → `GA**AC**CACACATAGGT` introduces a non-alternating `AACC` block → **not** a tandem dup.

ferro emits `g.20004508_20004509dup` for **both**. Decoding that dup gives `GACACACACATAGGT`, but the `insAC` input is `GAACCACACATAGGT` — a **different sequence**. ferro silently changed the variant.

```
insAC  decodes to: GAACCACACATAGGT
ferro dup decodes: GACACACACATAGGT   <- different
```

## Why this is wrong (spec)

HGVS DNA `duplication.md`: a duplication is "a copy … inserted **directly 3'** of the original copy" and "may only be used when the additional copy is **directly 3'-flanking** of the original copy (a tandem duplication)." The `insAC` variant is not a tandem expansion of the `AC` tract, so per the spec it must remain an insertion — which is exactly what mutalyzer returns.

## Root-cause hypothesis

`insertion_to_duplication` / `find_tandem_extent` (`src/normalize/rules.rs`) locate a nearby tandem tract under a cyclic rotation of the inserted unit, but do not verify that the inserted sequence is **in phase** with the tract *at the insertion point* — i.e. that placing the inserted copy actually yields a tandem expansion. An out-of-phase rotation (here `AC` where the tract reads `…A|C…` at the cut) is wrongly accepted as a dup. A correctness gate should confirm that the candidate dup decodes to the same sequence as the input insertion before converting.

## Scope / risk

- Pre-existing on `main`; **not** introduced by #864 (verified: pre- and post-#864 both emit the wrong dup, so #864 leaves this case unchanged).
- Corpus-invisible: the mutalyzer conformance corpus contains no instance, so the parity harness does not flag it.
- Correctness bug: emits a description that decodes to a different sequence than the input.

Part of #326; surfaced by #864.

nh13:
Closes #882.

## Problem

`normalize` converted an **out-of-phase** insertion into a tandem duplication (or repeat) that decodes to a **different sequence** than the input — silent variant corruption. Reproducer on `NC_000001.11` (window `GACACACATAGGT`, an `ACACACA` tract):

| input | before | after (this PR) | mutalyzer 3.1.1 |
|---|---|---|---|
| `g.20004503_20004504insAC` | `g.20004508_20004509dup` ❌ | `…insAC` | `…insAC` |
| `g.20004503_20004504insCA` | `g.20004508_20004509dup` ✅ | `g.20004508_20004509dup` | `g.20004508_20004509dup` |

Both insert at the same `A|C` cut, but only `insCA` is in phase with the tract; `insAC` decodes to `GAACCACACATAGGT` while the emitted dup decodes to `GACACACACATAGGT` — a different variant.

Per HGVS `duplication.md`: *"when there is no evidence that the extra copy … is in tandem (directly 3'-flanking the original copy), the change can not be described as a duplication; it should be described as an insertion."* Confirmed against **live mutalyzer 3.1.1** for both cases (and the real chr1 in-phase `insTG` locus, which correctly stays a dup).

The sibling `insertion_to_repeat` (2+-copy path) had the identical latent bug; both are fixed.

## Fix

A single in-phase correctness gate `tandem_edit_preserves_insertion` in `src/normalize/rules.rs`: it builds the sequence produced by the input insertion and the sequence produced by the **exact emitted** dup/repeat (post-alignment window + unit), and returns `false` when they differ. `insertion_to_duplication` and `insertion_to_repeat` call it just before emitting and return `None` on failure, so the existing caller fallthrough leaves the variant as a plain (rotated) `ins`. No caller/behavioral changes outside these two functions; in-phase conversions are unaffected (they decode-equal by construction).

## Tests

- Unit tests for the gate (in-phase → accept, out-of-phase → reject, bounds guards → no panic) and for both `insertion_to_duplication`/`insertion_to_repeat` on the #882 reproducers.
- End-to-end `normalize` tests on the real locus (`insAC` stays `ins`, `insCA` becomes `dup`).
- **Re-blessed pre-existing over-conversions.** Several synthetic tests (`ins_shift_matrix` cyclic-rotation rows across all axes, `issue_132`, `issue_418`, a `normalize_tests` rotation case) asserted out-of-phase inputs converting to dup/repeat — spec-violating over-conversions the gate now correctly declines. Their expectations are updated to the spec-correct `ins` form, and `issue_132`'s doc comment is corrected (only **in-phase** rotations become dups; its in-phase `ins_gt_matched_rotation_already_works` still asserts a dup and passes). Two pre-existing unit tests whose inputs were themselves latent out-of-phase instances were repaired to in-phase equivalents.

## Corpus

The `mutalyzer-normalize` corpus rows `NM_000143.3:c.-1_1insCAT` / `c.1_2insCAT` expected `c.-1_2dup` — a stale **Mutalyzer2**-era value. Live mutalyzer 3.1.1 and fixed-ferro both keep these as the insertion; the `normalized` values and `spec_citation` notes are corrected. (A full sweep of all 246 mutalyzer-expected rows against mutalyzer 3.1.1 found the corpus otherwise 96% faithful — these two were the only genuine drift; see #890 for recording comparator provenance.)

**Follow-ups (filed, not in this PR):** #891 — the `protein_description` axis on those two rows (`p.(Met1?)`) is ferro's hermetically-verified output but was originally derived from the superseded `c.-1_2dup`; its biological correctness under the insertion form is tracked for re-derivation. #890 — bump the behind comparators (hgvs-rs, mutalyzer-hgvs-parser, biocommons hgvs) and record comparator versions in corpus provenance (the sweep that found these two rows also confirmed the corpus is otherwise 96% faithful to mutalyzer 3.1.1).

## Verification

`cargo nextest run --features dev` → all pass (0 failed); `cargo clippy --features dev --all-targets -- -D warnings` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HGVS normalization for insertion variants in repeat regions and at CDS boundaries.
  * Out-of-phase insertions now stay as plain `ins` instead of incorrectly converting to duplications or repeat notation.
  * In-phase cases continue to normalize to the expected duplication form.
  * Updated regression coverage to ensure normalization remains stable and idempotent across genomic, CDS, RNA, and noncoding examples.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -8060,6 +8060,50 @@ mod tests {
         );
     }
 
+    // Real #882 locus GACACACATAGGT padded with 120-base neutral flanks so the
+    // +/-100 normalize window stays in-contig. Motif G is at 1-based g.121, so
+    // the A|C cut of the ACACACA tract is between g.122 (A) and g.123 (C).
+    fn phase_gate_882_provider() -> MockProvider {
+        let flank = "ACGT".repeat(30); // 120 bases
+        let seq = format!("{flank}GACACACATAGGT{flank}"); // len 253
+        let mut provider = MockProvider::new();
+        provider.add_genomic_sequence("NC_000001.11", &seq);
+        provider
+    }
+
+    #[test]
+    fn normalize_out_of_phase_ins_stays_ins_882() {
+        // g.122_123insAC sits at the A|C cut, out of phase with the ACACACA
+        // tract, so it must NOT become a dup (that was the silent-corruption bug).
+        let normalizer = Normalizer::new(phase_gate_882_provider());
+        let variant = parse_hgvs("NC_000001.11:g.122_123insAC").unwrap();
+        let normalized = normalizer.normalize(&variant).unwrap();
+        let output = format!("{}", normalized);
+        assert!(
+            output.contains("ins") && !output.contains("dup"),
+            "out-of-phase insAC must stay an insertion, got: {output}"
+        );
+        // Re-normalizing the result must not flip the phase gate on a second pass.
+        let again = format!("{}", normalizer.normalize(&normalized).unwrap());
+        assert_eq!(output, again, "normalize must be idempotent, got: {again}");
+    }
+
+    #[test]
+    fn normalize_in_phase_ins_becomes_dup_882() {
+        // g.122_123insCA at the same cut IS a legitimate tandem dup.
+        let normalizer = Normalizer::new(phase_gate_882_provider());
+        let variant = parse_hgvs("NC_000001.11:g.122_123insCA").unwrap();
+        let normalized = normalizer.normalize(&variant).unwrap();
+        let output = format!("{}", normalized);
+        assert!(
+            output.contains("dup"),
+            "in-phase insCA should become a dup, got: {output}"
+        );
+        // The emitted dup must stay a dup when normalized again.
+        let again = format!("{}", normalizer.normalize(&normalized).unwrap());
+        assert_eq!(output, again, "normalize must be idempotent, got: {again}");
+    }
+
     #[test]
     fn test_normalize_duplication_shifts_3prime() {
         // NM_001234.1 has G repeat spanning positions c.9-33 (25 G's)
--- a/src/normalize/rules.rs
+++ b/src/normalize/rules.rs
@@ -82,6 +82,73 @@ pub fn insertion_is_duplication(ref_seq: &[u8], pos: u64, inserted_seq: &[u8]) -
     false
 }
 
+/// In-phase correctness gate for insertion->dup / insertion->repeat conversion
+/// (issue #882).
+///
+/// Returns `true` iff inserting `inserted_seq` at `ins_pos` (0-based index of
+/// the base immediately 5' of the insertion point; the insertion sits between
+/// `ins_pos` and `ins_pos + 1`) yields exactly the same full sequence as the
+/// tandem edit that replaces `ref_seq[region_start..region_end_excl]` with
+/// `unit` repeated `total_count` times.
+///
+/// `region_start`, `region_end_excl`, `unit`, and `total_count` are the
+/// POST-alignment values the caller is about to emit (the dup/repeat's own
+/// window and unit), so this validates the literal emitted edit — never the raw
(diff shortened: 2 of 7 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

Issue #852, opened 2026-06-26
fix(project): minus-strand c.→g. repeat projection drops reverse-complement and collapses span


nh13:
Part of #326 (genomic axis; spec-arbitration sweep 2026-06-25). Bucket: ferro bug.

> **Diagnosis (corrected 2026-06-26, verified on the conformance code path).** An earlier note on this issue described CLI-path (`project_variant`) behavior; that path pre-normalizes the input and gives different output. The conformance harness (`axis_genomic`) uses **`project_to_genomic`** (raw, no source-axis pre-normalization), and that is the path these ledger rows fail on. The original framing — "drops reverse-complement + collapses span" — is **correct for `project_to_genomic`**.

## Per-row status on the conformance path (`project_to_genomic`), verified against reference bases

| input (minus-strand `c.`) | got | expected | reference tract | issue |
|---|---|---|---|---|
| `NG_012337.1(NM_012459.2):c.10CA[5]` | `g.4913CA[5]` | `g.4911_4914GT[5]` | `TGTGT` (4910-4914) | unit not revcomp'd (`CA` vs `GT`) + collapsed to single anchor + no entire-range |
| `NG_012337.1(NM_012459.2):c.3GC[5]` | `g.4920GC[5]` | `g.4917_4920GC[5]` | `GCGC` (4917-4920) | collapsed to single anchor (unit `GC` is palindromic so revcomp is invisible here) |
| `NG_012337.1(NM_012459.2):c.6C[4]` | `g.4917C[4]` | `g.4917G[4]` | `G` (4917) | unit not revcomp'd (`C` vs `G`) |
| `NG_009299.1(NM_017668.3):c.*16_*18T[8]` | `g.137762_137764T[8]` | `g.137761_137764A[9]` | `AAAA` (137761-764) | unit not revcomp'd (`T` vs `A`) + count not flank-absorbed (8 vs 9) + anchor off |
| `NG_009299.1(NM_017668.3):c.33_35CAA[5]` | declines: *outside placed genomic span (#480/#655)* | `g.137803_137808TTG[6]` | `TTGTTG` (137803-808) | **NOT a repeat bug** — placement-data gap (see below) |

## Root causes

[...]

PR #897, opened 2026-07-01
refactor(normalize): make normalize_repeat the single source of truth for ins→repeat canonicalization (#866)

### LINKED ISSUE #866: refactor(normalize): unify insertion_to_repeat and normalize_repeat ins→repeat canonicalization
Two functions independently canonicalize into HGVS repeat notation: `insertion_to_repeat` (`src/normalize/rules.rs:192`, applied to insertions) and `normalize_repeat` (`src/normalize/rules.rs:1319`, applied to `Repeat` edits). They were mutually non-idempotent on the 3'-phase (insertion_to_repeat reported the 5'-most tract window; normalize_repeat 3'-shifts) — fixed in #852 by making `insertion_to_repeat` 3'-align.

Maintaining two implementations of one operation is a divergence risk. Unify them (one delegates to the other, or they share the tract-finding + 3'-phase logic), and add a cross-check test that the two agree.

Relates to #852.

nh13:
Closes #866.

## What

Two functions independently canonicalized an insertion/expansion into HGVS repeat notation: `insertion_to_repeat` (for insertion edits) and `normalize_repeat` (for repeat edits). They were mutually non-idempotent on the 3'-phase until #852 gave them a *shared* `three_prime_align_tract` and a cross-check test — so the highest-risk shared logic was already unified. What remained duplicated was the **canonicalization arithmetic downstream of tract discovery**: both computed the copy count, ran the same 3'-align, and applied the same `is_coding && unit_len % 3 != 0` codon gate, inline.

This makes `normalize_repeat` the **single source of truth** for ins→repeat canonicalization. `insertion_to_repeat` now owns only the insertion-specific work — rotation-aware tract discovery at the insertion point (`find_tandem_extent`) and the #882 preservation gate — and delegates the count / 3'-align / codon-gate decision to `normalize_repeat` by re-expressing the discovered tract as the well-formed explicit range `unit[ref_count]`, expanded by the inserted copies.

## Why it's behavior-preserving

- `find_tandem_extent` and `count_tandem_repeats` walk the tract with byte-for-byte-identical unit-step loops, so `normalize_repeat` re-discovers the exact same tract, absorbs zero flanks (stated span == true tract), and applies the same `three_prime_align_tract` + codon gate.
- The chosen rotation is primitive, so `smallest_repeat_unit(unit) == unit` (no double-canonicalization).
- `added_copies >= 2` ⇒ the delegated target is always `>= ref_count + 2`, so only `Repeat` (or the codon-gated `Insertion` → `None`) is reachable; `Deletion`/`Duplication`/`Unchanged` are asserted unreachable (`unreachable!` rather than a silent catch-all, so a future relaxation of the guard fails loudly at test time).
- The #882 preservation gate still runs on the emitted window.

## Non-goal

The two tract-*finders* stay separate — their anchoring contracts genuinely differ (`find_tandem_extent` is between-bases / insertion-point; `count_tandem_repeats` is on-position). Forcing them together would add abstraction and risk without removing the arithmetic duplication this issue targets.

## Tests

Broadened the `insertion_to_repeat_agrees_with_normalize_repeat_phase` cross-check (the issue's explicit deliverable) from two di-repeat cases to also cover a tri-nucleotide tract (`CAG`), a multi-copy homopolymer (`A`), a **single-reference-copy homopolymer** (the one structurally distinct delegated branch, `end_pos == pos` / `ref_count == 1`), and a **coding codon-gate** shape (both paths agree a non-codon-aligned c. expansion is not a repeat).

## Verification

- Hermetic suite unchanged: **7666 passed, 147 skipped** — identical before and after (the ~40 repeat/insertion unit tests + integration tests are the behavior lock).
- `cargo clippy --features dev --all-targets -- -D warnings` + `cargo fmt --all --check` clean.
- Manifest-gated conformance axes (`FERRO_MANIFEST` set, scratch reference): **failing-row set byte-identical** to the base branch (36 rows, all pre-existing) — this PR introduces no conformance change.

Single file (`src/normalize/rules.rs`); no public API change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how insertion-to-repeat normalization behaves, making repeat handling more consistent across coding and non-coding contexts.
  * Preserved existing fallback behavior when an expansion should remain a literal insertion.
  * Kept in-phase/out-of-phase results stable while aligning repeat emission with broader normalization logic.

* **Tests**
  * Expanded coverage for repeat and insertion edge cases, including trinucleotide, homopolymer, and structural boundary scenarios.
  * Added checks to ensure both normalization paths agree when repeat formatting is not applicable.

### DIFF
--- a/src/normalize/rules.rs
+++ b/src/normalize/rules.rs
@@ -2893,6 +2893,41 @@ mod tests {
                 nr_unit: b"GC",
                 nr_count: 5,
             },
+            // Case 3 (#866): tri-nucleotide `CAG` tract ref[2..11); ins abuts its 3'
+            // end. Both paths must agree on window (3,11,"CAG").
+            Case {
+                ref_seq: b"GGCAGCAGCAGGG",
+                ins_pos: 10,
+                ins_seq: b"CAGCAG",
+                nr_pos: 2,
+                nr_end: 10,
+                nr_unit: b"CAG",
+                nr_count: 5,
+            },
+            // Case 4 (#866): multi-copy homopolymer `A` tract ref[3..8). Window (4,8,"A").
+            Case {
+                ref_seq: b"GGGAAAAAGGG",
+                ins_pos: 7,
+                ins_seq: b"AA",
+                nr_pos: 3,
+                nr_end: 7,
+                nr_unit: b"A",
+                nr_count: 7,
+            },
+            // Case 5 (#866): single-reference-copy homopolymer (`ref_count == 1`) — the
+            // one structurally distinct delegated branch, where the re-expressed range
+            // collapses to `end_pos == pos` with `unit_len == 1`. `count_tandem_repeats`
+            // still succeeds, so the len>=2 rotation fallback is unreachable. Window
+            // (2,2,"A").
+            Case {
+                ref_seq: b"GAG",
+                ins_pos: 1,
+                ins_seq: b"AA",
+                nr_pos: 1,
+                nr_end: 1,
+                nr_unit: b"A",
+                nr_count: 3,
+            },
         ];
         for c in cases {
             let (_b, _c, ins_start, ins_end, ins_unit) =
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #900, opened 2026-07-01
test(conformance): re-derive insCAT protein_description; demote the 5'UTR-insertion divergence (#891)

nh13:
Closes #891.

> Based on current `main` (does not yet include #895/#899). Will rebase onto the multi-axis `Vec` annotation schema once those land — the change is a data-only conformance re-bless, rebase-compatible by design.

## What

The two `NM_000143.3` insCAT corpus rows carried a `protein_description` of `p.(Met1?)` derived from the **stale** Mutalyzer2 normalized form `c.-1_2dup` (corrected to the insertion form in #882). Re-derived both against ferro on the real reference **and** live mutalyzer 3.1.1:

| row | ferro | mutalyzer 3.1.1 | spec | disposition |
|---|---|---|---|---|
| `c.1_2insCAT` (CAT splits the ATG) | `p.(Met1?)` | `p.?` (equiv.) | init codon affected → `p.(Met1?)` (substitution.md:51-52) | **already correct & passing** — note updated only |
| `c.-1_1insCAT` (CAT wholly in 5'UTR, ATG intact) | *declines* (empty projection) | `p.(=)` | ambiguous (see below) | **accepted divergence** — value corrected to `p.(=)`, demoted via ledger |

## Why `c.-1_1insCAT` is an accepted divergence, not a bug

The insertion is wholly 5' of the ATG; the CDS is unchanged. Ferro **deliberately** emits no protein consequence for a variant that doesn't overlap the CDS (policy from #857, tested by `pure_3utr_deletion_has_no_protein_consequence`). Mutalyzer emits `p.(=)`. The HGVS spec is **ambiguous** on `p.(=)`-vs-no-protein for a non-CDS variant (substitution.md:43 defines `p.(=)` as "analysed, no change", but "no protein description for a non-coding-region variant" is equally defensible) — so ferro's decline is spec-defensible, and this is an accepted divergence. The ambiguity is filed as a new spec-suggestion candidate on the SVD-WG tracker (#466).

## Mechanism note (why not a typed `accepted_divergence`)

[...]

Issue #861, opened 2026-06-26
test(conformance): accepted-divergence annotation sweep (parity arbitration 2026-06-25)


nh13:
Part of #326. Bucket: **accepted-divergence annotation sweep** (no code) — the disposition half of the 2026-06-25 spec-arbitration of the parity corpus.

~45 `baseline-failures` rows are cases where **ferro is spec-correct (or downstream of an already-accepted divergence) and mutalyzer diverges**. They are demoted by adding `accepted_divergence` / `accepted_rejection` annotations in `cases.json` (then re-blessing the ledgers #799-style) — no ferro change. Grouped:

- **normalized (12):** selector resolves to latest curated cdot version vs `NG_`-annotated (`refseq.md` L264; #710/#763/#783 family). *(If policy issue for version-preference lands and ferro converges, drop these.)*
- **protein (14):** transcript-version (5), `PCCB_v001` downstream of the accepted `normalized` divergence (4), synonymous `p.(=)` vs `p.(Asp92=)` (1), start-codon `p.(Met1?)` (4).
- **coding (8):** transcript-version-only (6); delins-with-internal-stop where ferro's trailing `Ter` is the *more* spec-faithful terminating form (2).
- **infos (6):** mutalyzer-internal `I*` correction chatter absent from the HGVS spec (`SHUFFLE_APPLIED` over-emission 2, `ICORRECTEDPOINT` out-of-bounds false-positive 1, lenient reinterpretation of malformed `c.41>CA` 2, cross-ref intronic-offset reject where both tools reject 1).
- **genomic (4):** invalid `c.41>CA` substitution rejected (1), `pter/qter` not c.-numberable (2), literal reverse-complemented insertion vs mutalyzer's optional cross-ref-range form (1).
- **errors (1):** `c.` on `NM_001099625.2` — ferro's cdot **has** a CDS (start_codon 162/stop_codon 1041) so ferro processes it; mutalyzer lacks the CDS and emits `ENOCDS` → data difference, ferro arguably more correct.

Deliverable: one PR that annotates these in `cases.json`, regenerates `failure-patterns.md`, and re-blesses the affected ledgers (verify identity unchanged, 0 regressions). ~45 rows across 6 axes.

Issue #857, opened 2026-06-26
feat(project): predict C-terminal extension / stop-loss consequence (p.(Ter…ext…))


nh13:
Part of #326 (protein_description / coding_protein_descriptions; spec-arbitration sweep 2026-06-25). Bucket: missing feature.

A variant that deletes or spans the stop codon should yield a C-terminal **extension** consequence `p.(Ter…ext…)` (`extension.md`), but ferro emits `[empty projection] no protein predicted`.

- `NG_012337.1(NM_012459.2):c.297_*1del` → expected `p.(Ter99TyrextTer6)`, got empty projection.
- `NG_012337.1:g.13124_13125del` (→ `NM_003002.2:c.480_*1del`) → expected `p.(Ter160CysextTer29)`, got empty projection.

Implement stop-loss / C-terminal extension prediction. 2 rows (protein 21 + coding 16).

Issue #764, opened 2026-06-18
test(conformance): empty-projection budget is reference-sensitive (committed counts don't reproduce on current reference)


nh13:
## Problem

The `empty-projection/<axis>.count` output-quality gate (#651) is **reference-sensitive**: the committed counts do not reproduce against the current prepared reference, so a manifest run trips the gate (or under-counts) for reasons unrelated to code changes.

Measured on a recent `origin/main` against the scratch prepared reference:

| Axis | committed `.count` | live |
|---|---:|---:|
| `protein_description` | 19 | ~24 |
| `coding_protein_descriptions` | 33 | 5 |

For `protein_description`, a run at the exact commit that blessed `19` (#651/#675) yields `25` on the current reference — i.e. the baseline was blessed against a *different* prepared-reference state than what is mounted now. So the gate fires on reference drift, not regression.

## Ask

Make the gate robust to reference drift. Options:
- Re-bless both counts against the **canonical CI/nightly** reference (not a contributor's local prepared data), and document which reference the counts are pinned to; or
- Make the budget tolerant (e.g. key the count to a manifest hash, or allow a small band), so the gate measures *regression* rather than absolute reference-dependent counts.

Refs #326, #498, #651.

PR #904, opened 2026-07-01
test(conformance): let accepted_rejection disposition an empty-projection Err (#903)

### LINKED ISSUE #903: test(conformance): let accepted_divergence disposition an empty-projection Err (unblocks #326 close-path)
## Problem

`AxisTally::record` (`tests/it/mutalyzer_normalize_tests.rs`) cannot give an **empty-projection `Err`** any typed disposition. All mismatch dispositions — `accepted_divergence`, `known_bug`, `improvement`, `reference_unavailable`, `spec_citation` — are gated behind `if actual.is_ok() || errors_axis_divergence` (~line 623), and `accepted_rejection` explicitly excludes the `EMPTY_PROJECTION_SENTINEL` (~line 582). So a row where ferro **spec-correctly declines** (produces no projection) but the comparator emits a value can only land in `self.fail` → the raw `baseline-failures/<axis>.txt` ledger, never in a typed bucket.

Surfaced by #891 (`NM_000143.3:c.-1_1insCAT`: ferro declines protein per the #857 non-CDS policy; mutalyzer emits `p.(=)`) and documented as a close-condition caveat on #326: the `protein_description` ledger holds ~18 such empty-projection accepted-FAILs (the `NG_(NM_):c.1A>G` init-codon rows + the #891 row) that no annotation can demote — so `baseline-failures/protein_description.txt` can never reach "empty," blocking #326's literal close condition.

## Fix

Let `accepted_divergence` disposition an empty-projection `Err`, mirroring the existing `CodingProteinDescriptions` "missing pair" `Err`-bucketing precedent (~line 607). A targeted block: for an empty-projection `Err` on a projection axis, if the case carries an `accepted_divergence` on that axis, bucket it into `divergence_accepted` (out of `fail`). Add a `Policy` variant for the "ferro spec-correctly declines; comparator emits a value" class.

**Safety (no masking):** the separate `empty_got` **count gate** (#764-pinned `empty-projection/<axis>.count`) still catches any *rise* in the total empty count, so a NEW value→empty degradation still fails via the count gate even if a stale annotation would bucket it. The existing XPASS guard still fires if ferro starts matching (annotation stale).

## Payoff

- Gives the ~18 empty-projection `protein_description` rows a **typed, XPASS-guarded** disposition instead of raw ledger entries — advancing #326's close-path (option (b) in the #326 close-condition caveat).
- Unblocks converting those rows in the #861 annotation sweep.

## Scope

The harness mechanism + a `Policy` variant + a hermetic unit test proving an empty-projection `Err` with an `accepted_divergence` buckets to `divergence_accepted`. Converting the real corpus rows (the #891 row + the ~17 init-codon rows) is a follow-up re-bless (#861 / an amend to #900), left out here to keep the mechanism PR focused and conflict-free.

nh13:
Closes #903.

## Problem

`AxisTally::record` could give **no typed disposition** to an empty-projection `Err` (ferro produced no output on an axis): `accepted_divergence`/`known_bug`/`spec_citation` are gated behind `actual.is_ok()`, and `accepted_rejection` explicitly excluded the `EMPTY_PROJECTION_SENTINEL`. So a row where ferro **spec-correctly declines** but the comparator emits a value — e.g. #891's `NM_000143.3:c.-1_1insCAT` (ferro declines protein per the #857 non-CDS policy; mutalyzer emits `p.(=)`) — could only land in the raw `baseline-failures/<axis>.txt` ledger, never a typed bucket. This is why ~18 `protein_description` rows can't leave that ledger, blocking #326's "every ledger empty" close condition.

## Fix

An empty projection is a **decline**, and `accepted_rejection` is already the Err-on-projection-axis disposition — its *only* blocker was the sentinel exclusion. So (adversarial review picked this home over grafting a second Err-exception onto `accepted_divergence`):

- Add `RejectionReason::NonCdsNoProjection857` + a closed `RejectionReason::disposition_empty_projection()` predicate (`_ => false`, so opting in is a deliberate reviewed choice, never an ad-hoc flag).
- Relax the `record()` accepted_rejection condition to admit the sentinel **only** when the reason opts in.

## Safety — no masked regressions (verified)

- The `empty_got` increment is **unconditional at the top of `record()`**, above every disposition/`return`, so the `#764`-pinned empty-projection **count gate** in `finish` still catches any *rise* regardless of bucketing.
- **Disjoint by construction:** the opt-in block requires the sentinel; every non-opt-in reason keeps the pre-#903 hard-FAIL.
- **XPASS / Ok-transition honest:** `accepted_rejection` buckets only `Err`, so if ferro starts emitting a value it is not masked — an `Ok` that *matches* XPASS-FAILs (stale annotation caught), an `Ok` that *mismatches* hard-FAILs (ferro stopped declining).
- A genuine panic still hard-FAILs.

## Payoff

Gives the ~18 empty-projection `protein_description` rows a **typed, XPASS-guarded** disposition instead of raw ledger entries — advancing #326's close-path (option (b) in the #326 close-condition caveat) and unblocking their conversion in the #861 sweep.

## Scope

Mechanism + hermetic tests **only** — no `src/` production code (projector/normalizer), no corpus/ledger change. Converting the real rows (#891's row + the ~17 `NG_` init-codon rows) to this typed disposition is a follow-up re-bless (#861 / an amend to #900), kept out here so the mechanism lands conflict-free with the in-flight `cases.json` PRs.

## Verification

Full suite **7683 passed**; new hermetic tests cover: opt-in buckets an empty `Err` (+ count still increments), non-opt-in still FAILs, opt-in still buckets a non-empty `Err`, `Ok`-mismatch still FAILs, and the predicate/serde round-trip. `clippy -D warnings` + `fmt` clean.

## Summary by CodeRabbit

* **New Features**
  * Added support for a new accepted-rejection reason that can treat empty-projection results as accepted in specific cases.

* **Bug Fixes**
  * Improved rejection handling so opted-in cases no longer hard-fail on empty-projection sentinel errors.
  * Preserved existing failure behavior for non-opted-in cases and for normal mismatches.
  * Updated matching behavior and coverage to confirm the new routing and serialization work correctly.

### DIFF
--- a/src/conformance/mutalyzer.rs
+++ b/src/conformance/mutalyzer.rs
@@ -654,9 +654,39 @@ pub enum RejectionReason {
     /// behaviour (#853, residual of #480/#655; subsumes #865).
     #[serde(rename = "ferro-policy-853-ng-partial-transcript-coverage-declined")]
     NgPartialTranscriptCoverageDeclined853,
+    /// ferro emits **no projection** on this axis for a variant its policy says
+    /// has no consequence there — e.g. a variant wholly outside the CDS
+    /// (pure-5'UTR / pure-3'UTR) produces no protein prediction (the #857 policy,
+    /// tested by `pure_3utr_deletion_has_no_protein_consequence`) — where the
+    /// comparator instead emits a value (e.g. mutalyzer `p.(=)`, "analysed, no
+    /// change"). The HGVS spec is ambiguous on no-projection vs the "analysed,
+    /// unchanged" value for a non-CDS variant, so ferro's decline is a spec-
+    /// defensible accepted rejection rather than a bug (#857/#891; #466 tracks the
+    /// upstream spec clarification). This is the **only** `RejectionReason` that
+    /// opts in to bucketing an empty-projection `Err` (see
+    /// [`RejectionReason::disposition_empty_projection`]): the harness represents
+    /// "ferro produced nothing" as an `EMPTY_PROJECTION_SENTINEL` `Err`, which is
+    /// otherwise a hard FAIL. The empty→output transition is handled honestly:
+    /// because `accepted_rejection` buckets only `Err`, if ferro later starts
+    /// emitting a value it is NOT masked — an `Ok` that *matches* XPASS-FAILs (the
+    /// stale annotation is caught), and an `Ok` that *mismatches* falls through to
+    /// a hard FAIL (ferro stopped declining, now produces a differing value). The
+    /// `#764` empty-projection count gate additionally catches any *rise* in the
+    /// empty count.
+    #[serde(rename = "ferro-policy-857-non-cds-no-projection")]
+    NonCdsNoProjection857,
 }
 
 impl RejectionReason {
+    /// Whether this reason opts in to dispositioning an **empty-projection**
+    /// `Err` (ferro produced no output on the axis) as an accepted rejection.
+    /// `accepted_rejection` normally excludes the `EMPTY_PROJECTION_SENTINEL`
+    /// (an empty projection is otherwise a hard FAIL, #651); a reason returning
+    /// `true` here relaxes that for its rows only. Closed match (`_ => false`) so
+    /// adding a reason is a deliberate, reviewed opt-in — never an ad-hoc flag.
+    pub fn disposition_empty_projection(self) -> bool {
+        matches!(self, RejectionReason::NonCdsNoProjection857)
+    }
     /// Stable identifier used in summary lines so reviewers can grep across runs.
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #915, opened 2026-07-04
fix(reference): skip hidden/AppleDouble files when scanning FASTA directories

nh13:
## Problem

`MultiFastaProvider::from_directory` scans a reference subdirectory for `*.fna`/`*.fa`/`*.fasta` files and loads each one's `.fai` index. macOS AppleDouble sidecar files — e.g. `._LRG_430.fasta`, created whenever a reference directory is copied through macOS tooling (Finder, `tar`/`cp` carrying xattrs, some network shares) — match that extension filter. Their accompanying `._*.fasta.fai` is a binary extended-attribute blob, so `load_fai_index` fails with `stream did not contain valid UTF-8` and the **entire reference load aborts**.

The failure mode is subtle: parsing still works (it never reads the reference), so the provider looks healthy, but *every* normalization fails because no sequences can be resolved — with no indication that a stray `._*` file was the cause.

## Fix

Skip directory entries whose file name begins with `.` (hidden files and AppleDouble sidecars) before the extension check, so a reference directory contaminated with `._*` files loads cleanly.

## Test

Adds `test_from_directory_skips_macos_appledouble_sidecars`: a temp directory with a valid FASTA plus a `._real.fasta` / `._real.fasta.fai` pair carrying non-UTF-8 content; `from_directory` must load the real sequence and not abort on the sidecar.

## How this surfaced

Loading a reference stack that had been copied through macOS `tar` (which serializes xattrs into AppleDouble sidecars on extraction). ferro parse worked, but every normalize returned 0 successes with no surfaced error — tracing to `._LRG_*.fasta.fai` files under `lrg/`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved FASTA directory scanning to ignore hidden files, including macOS AppleDouble sidecar files and their index files, preventing reference loading failures.
- **Tests**
  - Added a regression test to confirm valid FASTA inputs still load correctly when hidden sidecar files are present.

### DIFF
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -298,6 +298,20 @@ impl MultiFastaProvider {
             })?;
 
             let path = entry.path();
+
+            // Skip hidden files and macOS AppleDouble sidecars (names beginning with '.', e.g.
+            // `._LRG_430.fasta`). These are not real FASTAs — an accompanying `._*.fasta.fai` is a
+            // binary extended-attribute blob, and attempting to parse it as a FAI index would abort
+            // the entire reference load with a "stream did not contain valid UTF-8" error. Such files
+            // are commonly introduced when a reference directory is copied through macOS tooling.
+            let is_hidden = path
+                .file_name()
+                .and_then(|n| n.to_str())
+                .is_some_and(|n| n.starts_with('.'));
+            if is_hidden {
+                continue;
+            }
+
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #908, opened 2026-07-02
test(conformance): two residual infos-axis rows block the final #326 close (compound-allele 3'-shift symptom + ICORRECTEDPOINT map lossiness)


nh13:
## Context

The #861 accepted-divergence annotation sweep dispositioned 40 of the 42 remaining `baseline-failures` FAIL rows and emptied every axis ledger **except `infos`**, which retains **2 rows**. These two are deliberately *not* dispositioned by #861 because neither is a clean "ferro is spec-correct" accepted divergence: one is a live ferro bug, the other is a harness info-map limitation. They are the last blockers before the ferro↔mutalyzer parity tracker #326 can close.

## Row 1 — live compound-allele 3'-shift bug (tracker #487 is closed but the bug persists)

```
NG_009299.1(NM_017668.3):c.[250del;41A>C]
  infos expected: ICORRECTEDPOINT, ISORTEDVARIANTS
  ferro infos [] — missing SHUFFLE_APPLIED (ferro does not 3'-shift the c.250del member)
```

ferro fails to 3'-shift the `c.250del` member across the compound allele to `c.*189del` (`c.[41A>C;*189del]`). This is the same bug already carried as `known_bug {axis: normalized, tracking_issue: 487}` on this case — and the normalized-axis known_bug is **still active-bucketing** (it has not XPASS'd), i.e. **ferro is still wrong**, even though **#487 is CLOSED**. So the tracker reference is stale: the compound-allele 3'-shift residual needs a live tracking issue (this one, or a re-opened #487).

The `infos`-axis symptom (missing `SHUFFLE_APPLIED`) is an **under-emission of a mapped code** = a real bug per the axis-runner's own contract (`mutalyzer_normalize_tests.rs` ~L1717 "under-emission = real bug"). It must **not** be papered over with `accepted_rejection` (that would falsely assert ferro-correct and mask #487).

**Harness gap:** even the honest disposition — `known_bug {axis: infos, tracking_issue: <this>}` — cannot bucket it today, because `AxisTally::record` gates `known_bug` behind `actual.is_ok() || errors_axis_divergence`, and an infos under-emission is a (non-errors-axis) `Err`. Options:
1. **Fix the bug** (compound-allele 3'-shift) — both the normalized and infos symptoms resolve; cleanest.
[...]

PR #919, opened 2026-07-04
test(conformance): disposition residual infos rows; reclassify NM_017668.3 exon-junction rows as spec_citation (#908, #918)

### LINKED ISSUE #918: fix(normalize): 3'-shift CDS del/dup across the CDS↔UTR boundary (single + compound-allele)
Split out of the (now-closed, overloaded) #487 as a focused live tracker; surfaced by the last two `infos`-axis residual rows in #908.

## Bug

In a compound allele, ferro does not 3'-shift a CDS deletion member across the CDS↔3'UTR boundary the way mutalyzer does:

```
NG_009299.1(NM_017668.3):c.[250del;41A>C]
  ferro:     c.[250del;41A>C]   (member c.250del not shifted)
  mutalyzer: c.[41A>C;*189del]  (c.250del 3'-shifted to c.*189del, members re-sorted)
```

Symptoms across the mutalyzer-normalize corpus axes:
- `normalized`: the `c.250del` member stays put instead of shifting to `c.*189del`.
- `infos`: ferro emits no `SHUFFLE_APPLIED` (under-emission) because no shift is applied.

## Root cause

`normalize_cds` derives the shuffle boundary from `get_cds_boundaries_with_axis_info` (`src/normalize/boundary.rs:134-187`), which clamps a CDS variant's right shuffle boundary at `cds_end` (the deliberate #337 CDS↔UTR axis clamp), then hard-bails on a cross-axis endpoint via `CrossAxisVariantNotShuffled` (`src/normalize/mod.rs:2299-2341`). So a CDS deletion can never 3'-shift into the 3'UTR.

Whether the HGVS spec *mandates* 3'-shifting across the CDS↔UTR boundary is the open question here — #337 deliberately does not. If ferro should shift, relaxing the clamp for the single-endpoint del/dup case (and re-expressing the shifted coordinate in the 3'UTR `*` frame) is the fix; a full conformance re-bless is required because this changes normalization of any CDS-terminal del/dup.

## Conformance wiring

Tracked in `tests/fixtures/mutalyzer-normalize/cases.json` as `known_bug {axis: normalized}` and `known_bug {axis: infos}` on `NG_009299.1(NM_017668.3):c.[250del;41A>C]` (and its in-order sibling `c.[41A>C;250del]`), pointing at this issue. When fixed, the XPASS guard fires to remove the annotations and (for the per-member `SHUFFLE_APPLIED`) swap to the policy-499 `accepted_divergence` like the sibling rows.

Part of the #326 parity burn-down; the #487 successor for the compound-allele 3'-shift specifically (see #912 for the broader #487 split).

nh13:
Two related conformance changes, both surfacing on the mutalyzer-normalize **infos** axis.

## 1. #908 — the two residual infos-axis rows

Disposition the two remaining infos-axis divergences and add an **infos-axis under-emission `Err` exception** to `AxisTally::record`: when ferro emits fewer mapped info codes than the corpus expects, a non-panic `Err` may bucket into a per-axis disposition rather than hard-FAIL. This lets a genuinely tracked ferro bug be xfail'd honestly instead of papered over as `accepted_divergence` (which is deliberately still rejected — an under-emission has no spec basis for mere "we accept the difference").

## 2. #918 — reclassify the reverse-strand NM_017668.3 allele rows

`NG_009299.1(NM_017668.3):c.[41A>C;250del]` and its cis-swapped twin `c.[250del;41A>C]` were annotated `known_bug{#918}` on the premise that ferro should 3'-shift `c.250del` to mutalyzer's `c.*189del`. That premise is wrong:

- `c.*189del` sits ~947 bases downstream of `c.250`, spanning **8 exon/exon junctions**. The HGVS 3'-rule applies to the contiguous coding-DNA reference but explicitly does **not** shift a deletion/duplication across an exon/exon junction (general.md 3'-rule exception; `background/numbering.md#DNAc`) — shifting there would project back to the wrong exon in `g.`.
- The `c.*189del` form is additionally an artifact of the truncated reverse-strand `NG_009299.1` partial embedding of this transcript.
- `c.250del` is already most-3' within its exon (a `GG` dinucleotide), so ferro's value is the spec-correct one.

Reclassifications (all `known_bug` → `spec_citation`):

| input | axis |
|---|---|
| `c.[41A>C;250del]` | normalized |
| `c.[250del;41A>C]` | normalized |
| `c.[250del;41A>C]` | infos |

Adds `SpecSection::ThreePrimeRuleExonJunction` citing the exception.

### Harness: let an infos `spec_citation` bucket a spec-correct under-emission

The infos under-emission exception (part 1) originally admitted `known_bug` only. Because the missing `SHUFFLE_APPLIED` here is *spec-correct* (ferro rightly does not shift across the junction, so mutalyzer's `ICORRECTEDPOINT` has no ferro equivalent), `spec_citation` is the accurate disposition. Extend the exception so a non-panic infos `Err` also buckets into `spec_overridden` when the case carries a matching `spec_citation`. `accepted_divergence` stays excluded (no spec basis); a genuine panic still hard-FAILs. New unit tests cover both the bucketing and the panic guard; the existing `..._with_accepted_divergence_still_fails` guard is unchanged.

## Supersedes #936

#936 carried the *normalized* reclassification for these two rows alone (stacked on #924). This PR folds that reclassification together with the infos row and the shared `SpecSection` variant into one coherent, independent change off `main`, so **#936 is closed as superseded**.

## Does not close #918

The genuine CDS↔UTR 3'-shift bug tracked by #918 — `NG_012337.1(NM_012459.2):c.-1_*1del` (ferro echoes the input; the most-3' form is `c.1_*2del`) — is a separate cross-axis whole-CDS deletion handled elsewhere, so this PR intentionally does not `Closes #918`.

## Verification

- `axis_infos`: 0 FAIL (1 `spec_overridden`); `axis_normalized` + `axis_normalized_idempotent`: 0 FAIL / 0 XPASS.
- `generate_conformance_summary -- --check` passes.
- Affected `tally_*` unit tests pass; `cargo fmt` + `cargo clippy --features dev --all-targets` clean.


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #983, opened 2026-07-08
fix(normalize): canonicalize bracketed inserted-repeat counts >= 2 (#920)

nh13:
## What

A bracketed inserted repeat with an exact count >= 2 — multi-base `ins<seq>[N]` (`InsertedSequence::SequenceRepeat`) or single-base `ins<base>[N]` (`InsertedSequence::Repeat`) — bypassed normalization entirely: its `bases()` is `None`, so the alt-sequence extraction returned the edit verbatim. #926 intercepted only `[0]` (→ `g.=`) and `[1]` (→ `dup`), explicitly leaving `[N>=2]` "for the repeat-tract machinery" — but nothing routed it there, so ferro kept the non-canonical `ins[N]` where mutalyzer renders the repeat.

## Fix

N inserted copies of a unit are exactly the plain `ins<unit repeated N times>` literal, which the insertion path already canonicalizes: `insertion_to_repeat` discovers the abutting reference tract (`find_tandem_extent`) and merges the added copies into `[total]` repeat notation. So this unifies the `[1]` and `[N>=2]` arms of the degenerate-count intercept: expand N copies to a `Literal` and recurse (N=1 reduces to #926's existing `Literal(unit)` → dup path). One change fixes two #920 rows:

- **ins→repeat:** `LRG_303:g.6932_6933ins<24mer>[2]` → `g.6908_6932<24mer>[3]` (2 inserted + 1 adjacent reference copy = 3).
- **tandem-repeat merge:** `LRG_303:g.4_5insGT[5]` → `g.1_4GT[7]` (5 inserted copies merge with the adjacent `GT` tract).

### Bounded expansion (guards a DoS I'd otherwise introduce)

Materializing the expansion is memory-bounded: a pathological count (`insGT[1e12]`, or a `u64` that overflows the capacity product — which panicked *"attempt to multiply with overflow"*) falls through to the verbatim pass-through rather than allocating gigabytes or crashing. Cap is 1 MiB of bases; no real or conformance repeat approaches it. This matters because ferro parses untrusted input (web-service mode).

## Validation

- Demotes the two now-passing `known_bug` rows (`known_bug` → pass) in `tests/fixtures/mutalyzer-normalize/cases.json`; `failure-patterns.md` regenerated.
[...]

Issue #920, opened 2026-07-04
fix(normalize): canonical edit-form + allele-member detection (ins→dup/repeat, tandem merge, allele collapse)


nh13:
Focused live successor to the closed, overloaded #487 (its original scope was "canonical edit-form detection (ins→dup, tandem repeat, mito m., allele collapse)"). Split out of #487 by the #912 stale-tracker hygiene sweep; the CDS↔UTR 3'-shift rows of #487 moved to #918 instead.

ferro emits **valid but non-canonical** HGVS on a family of normalization-canonicalization gaps where mutalyzer produces the canonical form. None are spec-*invalid* output — these are `known_bug` (ferro is non-canonical), tracked in `tests/fixtures/mutalyzer-normalize/cases.json` on the `normalized` axis.

## The gaps (10 rows)

**Edit-form canonicalization**
- **ins→dup MUST** — a single inserted copy of the adjacent sequence is a duplication (`duplication.md` L17: a dup MUST be spelled `dup`, not `ins`). `LRG_303:g.6932_6933insAGCAACGTGATCGCCTCCCTCACCT[1]` → mutalyzer `g.6908_6932dup`; ferro keeps `ins[1]`. This is the DNA sibling of the (now closed) protein issue #92.
- **ins→repeat** — two inserted copies plus the existing copy form a 3-copy repeat. `…ins…[2]` → mutalyzer renders the repeat; ferro keeps `ins[2]`.
- **tandem-repeat merge** — `LRG_303:g.4_5insGT[5]` → mutalyzer merges with the adjacent existing GT tract into `g.1_4GT[7]`; ferro keeps the inserted `GT[5]`.
- **repeat 3'-shift** — `NG_009299.1(NM_017668.3):c.33_35CAA[5]` → mutalyzer 3'-shifts the repeat unit to `c.34_39AAC[6]`; ferro keeps the input anchor.
- **zero-copy ins→identity** — `LRG_303:g.6932_6933ins…[0]` is a no-op; mutalyzer collapses to `g.=`; ferro keeps `ins[0]`.
- **block-delins interior trim** — `LRG_199t1:c.235_237delinsTAT` → mutalyzer trims the unchanged interior base and decomposes to per-position substitutions; ferro keeps the block delins.

**Allele-member canonicalization**
[...]

Issue #912, opened 2026-07-02
test(conformance): known_bug/improvement annotations reference CLOSED trackers (stale-tracking hygiene sweep)


nh13:
## Summary

The v0.7.0 release audit found that **every tracking issue referenced by a `known_bug` / `improvement` annotation in the conformance corpora is CLOSED**, while the annotations are still active-bucketing — i.e. ferro still diverges. A closed issue referenced by a live divergence is mis-tracking: the divergence has no open home, and (for `known_bug`) the XPASS guard can no longer drive cleanup to the right issue. This is the generalization of the #487 case already surfaced in #908.

None of these emit spec-*invalid* output (they are valid-but-non-canonical / incomplete forms), so this is a **tracking-hygiene** issue, not a correctness blocker — but it should be swept before/around v0.7.0 so the release's known-divergence set is honestly tracked.

## The stale references (main @ f17d54e)

| annotation | tracker | state | # rows | ferro's actual divergence |
|---|---|---|---:|---|
| `known_bug` | **#487** | CLOSED | 13 | compound-allele 3'-shift + normalization-completeness gaps (ins→dup MUST, repeat 3'-shift, ins[0]→identity, block-delins trim, allele→sub collapse, redundant-`=` drop) |
| `known_bug` | **#646** | CLOSED | 8 | omits a valid downstream `n.` form on a non-overlapping non-coding transcript (enumeration gap) |
| `known_bug` | **#480** | CLOSED | 4 | `c.`-copy-range `delins` expanded transcript-contiguous vs genomic decomposition (spec-unpinned) |
| `improvement` | **#500** | CLOSED | 4 | no `NM_` selector synthesized on a bare-`NG_` `c.` input (refseq.md L138 SHOULD) |

## Actions

1. **Re-point or reclassify** each family to a live issue (or, for the spec-unpinned / arguably-ferro-correct #480 / #646 enumeration gaps, reclassify as `accepted_divergence` deferrals with an open tracker).
2. **#487 is overloaded and mis-tracked.** Its rows mix (a) a genuine **ins→dup MUST-violation** for tandem adjacent-copy insertions (`duplication.md L17` — a dup MUST be spelled `dup`, not `ins`; the DNA sibling of the OPEN protein issue **#92**) with (b) several normalization-completeness gaps. Split: the ins→dup MUST needs its own live issue (mirroring #92); the rest get an open normalization-polish tracker.
[...]

Issue #851, opened 2026-06-26
feat(project): complete NG_/LRG_ in-place c./n.→g. projection for compound alleles and UTR coordinates


nh13:
Part of #326 (genomic axis; spec-arbitration sweep 2026-06-25).

For an `NG_(NM_)` genomic-context input, single-variant `c.→NG_:g.` projection already works, but two sub-cases fail:

1. **Compound alleles are not projected.** `c.[…]` silently returns the un-projected `c.` form instead of `g.[…]`.
   - `NG_012337.1(NM_003002.2):c.[274G>T;278A>G]` → expected `NG_012337.1:g.[7125G>T;7129A>G]`, got `…(NM_003002.2):c.[274G>T;278A>G]`.
2. **5'UTR coordinates are hard-rejected** even though the `NG_` parent supplies the 5' flank.
   - `NG_012337.1(NM_012459.2):c.-35_*1del` → expected `NG_012337.1:g.3448_4957del`, got `err: Invalid CDS position: -35`.

Spec: `numbering.md` L41 (the `NG_(NM_):c.→g.` construction is well-defined), `refseq.md` L46 (the `NG_` reference supplies UTR/flank context). Extend genomic projection over allele members and resolve negative-UTR coordinates under an `NG_` parent. ~7 genomic-axis rows.

PR #985, opened 2026-07-08
test(conformance): reclassify c.[274;600] as accepted_divergence, not known_bug (#920)

nh13:
## What & why

`NG_012337.1(NM_003002.2):c.[274;600]` was tracked as a `normalized`-axis `known_bug` under #920, expecting ferro to converge on mutalyzer's `c.[274;*120]`. ferro emits `c.[274=;600=]`, and **both differences are deliberate, defensible ferro policy — not bugs.**

**1. Explicit identity (`=`).** The bare-position allele members (no edit operator) are coerced to the spec no-change form: ferro renders `c.[274=;600=]`; mutalyzer drops the identity. A bare member is non-spec (`syntax.yaml:8/135` — every allele member is a `position_edit`; `substitution.md:19` — no-change is `c.123=`), so ferro's explicit `=` is at least as spec-faithful. This is exactly `ferro-policy-654-explicit-identity-rendering`, already the `accepted_divergence` on **this row's genomic axis** and on the standalone `NG_012337.1:274` case (#851).

**2. Past-CDS coordinate (`600` vs `*120`).** `c.600` lies past the 480-nt CDS end. ferro flags it as `PositionPastEnd` (**#336** — plain c. numbering stops at the stop codon; positions beyond the CDS require the 3′UTR `*` form) and *preserves the input coordinate* rather than silently reinterpreting it as mutalyzer's `c.*120`. Surfacing an out-of-CDS-bounds coordinate as a likely input error, instead of coercing it, is the conservative ferro stance; both forms denote the same locus (c.480+120). "Fixing" it by coercing would **remove a `PositionPastEnd` warning #336 deliberately emits.**

## Change

Reclassify the `normalized`-axis annotation from `known_bug{#920}` → `accepted_divergence` (policy-654), matching the genomic-axis entry. This also **resolves the #912 normalized-vs-genomic axis-inconsistency note** for this row (both axes now `accepted_divergence`, same policy). Regenerate `failure-patterns.md`.

Fixtures + generated summary only — **no code change**.

## Validation

Re-blessed against the prepared reference: `normalized` / `genomic` axes pass, no XPASS. Full suite green (7798 tests), `generate_conformance_summary --check` byte-identical.

Resolves the bare-position past-CDS-coordinate row of #920 (as `accepted_divergence`, not a fix). Part of #326.

> Note: touches `cases.json` / `failure-patterns.md`, which also change in sibling #920 PRs #983/#984 — whichever merges later needs a trivial rebase + `generate_conformance_summary` regen.

### DIFF
--- a/tests/fixtures/mutalyzer-normalize/cases.json
+++ b/tests/fixtures/mutalyzer-normalize/cases.json
@@ -3359,16 +3359,18 @@
       "normalized": "NG_012337.1(NM_003002.2):c.[274;*120]",
       "genomic": "NG_012337.1:g.[7125;13244]",
       "to_test": true,
-      "known_bug": {
-        "axis": "normalized",
-        "tracking_issue": 920,
-        "note": "ferro renders the bare-position allele members with explicit identity and an absolute past-CDS coordinate (c.[274=;600=]); mutalyzer uses the 3'UTR form and drops the identity (c.[274;*120]). Past-CDS coordinate rendering + allele identity tracked in #920."
-      },
-      "accepted_divergence": {
-        "axis": "genomic",
-        "policy": "ferro-policy-654-explicit-identity-rendering",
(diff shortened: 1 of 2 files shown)


nh13:
Superseded by #987. That PR **fixes** the past-CDS coordinate (renders `c.600` as the canonical `c.*120` while keeping the W4004 warning and strict-mode rejection) so ferro's output becomes `c.[274=;*120=]`, and then reclassifies this row's normalized axis to a clean `=`-only `accepted_divergence` (policy-654, identical to its genomic axis). This PR reclassified the same row without fixing the coordinate, so it's redundant. Closing in favor of #987.

Issue #336, opened 2026-05-19
normalize/validate: reject c./n. positions past CDS-end / transcript-end (PositionPastEnd)


nh13:
## Symptom

ferro accepts `c.` (and `n.`) variants whose positions lie past the reference transcript's CDS-end / transcript-end. Biocommons rejects with `HGVSError`.

## Affected rows

6 lines in `tests/fixtures/biocommons-normalize/baseline-failures/normalized.txt` from #324, all on `NM_001001656.1` (CDS-end = 945):

| Input | bio | ferro |
|---|---|---|
| `c.946G>C` (cross=false, cross=true) | rejects | accepts (passes through) |
| `c.946dup` (cross=false, cross=true) | rejects | accepts (shifts past end to `c.*2dup`) |
| `c.935_946del` (cross=false, cross=true) | rejects | accepts (shifts past end to `c.935_*1del`) |

## Spec basis

§HGVS v21.0 `general.md` — "position numbering": coding-DNA / non-coding-transcript positions must reference an actual position in the transcript. A `c.` position numerically larger than the CDS-end has no defined meaning when the input did not use `*N` notation to explicitly target 3'UTR positions.

The `dup`/`del` variants above further drift past `c.*1` under shifting; the input itself was already malformed.

## Reproducer

With the standard manifest:

```
$ ferro normalize 'NM_001001656.1:c.946G>C' --reference <manifest>
NM_001001656.1:c.946G>C    # expected: HGVSError per biocommons
```

## Distinct from existing validation

- #266 / PR #272 (`W3016 LengthMismatch`) — validates ref-seq length vs. position-range length. Different concern.
- #253 — boundary-spanning ranges within the transcript. Different concern.

This is a new SVA code, suggested `W4xxx PositionPastEnd` (rejection in strict; warn in lenient).

## Suggested directions

1. Validation pass that resolves `c.`/`n.` end-of-range positions against the transcript's CDS-end / transcript-length from `get_transcript`. Reject when an integer position exceeds those bounds (vs. proper `*N` 3'UTR notation, which is fine).

2. Decide policy for the `dup`/`del` shift cases that *would not* be malformed without 3'-rule shifting (e.g. `c.935_945del` is in-bounds; `c.935_946del` is not). Likely: validate before the shift pass, not after.

3. Per CLAUDE.md ferro error-code convention, the new code lives in `src/error_handling/types.rs` + `registry.rs` + `corrections.rs`.

## Linked

[...]

PR #987, opened 2026-07-08
fix(normalize): render plain past-CDS coordinates in canonical c.*N form (#920/#336)

nh13:
## What & why

A plain `c.<N>` with `N > cds_len` is **out-of-scheme HGVS** — coding-DNA numbering ends at the last base of the stop codon, and 3′UTR positions use the `*` prefix (`numbering.md` L21/L30). ferro accepted such inputs in lenient mode but **echoed them verbatim**: e.g. `NM_003002.2:c.600` (CDS length 480) came back as `c.600`, a string LOVD / VariantValidator reject as *syntactically invalid* — instead of the canonical `c.*120`.

This came out of the #985 adversarial review: ferro's *flag-don't-coerce* stance is defensible (VariantValidator agrees), but **emitting the invalid string** was the one genuine defect.

## Fix

Render the position ferro already computed in canonical `*N` form when it maps into the 3′UTR (`1 ≤ N − cds_len ≤ utr3_len`), via a new `canonicalize_pastcds_pos_to_utr3` applied in the lenient bounds-check pass (which already has the transcript, `cds_len`, and `utr3_len`). It's a **rendering canonicalization only — the same nucleotide, in valid syntax** — so it does *not* adopt mutalyzer's silent coercion:

- The **W4004 `PositionPastEnd` warning still fires** (the input used out-of-scheme numbering) — ferro flags rather than silently coerces.
- **Strict mode still rejects** (unchanged) — consistent with ferro's strict-refuses / lenient-recovers contract, and with VariantValidator's flag posture. Both #336 strict tests pass **unchanged** (neither asserts the lenient output string).
- A position **past the whole transcript** (`N − cds_len > utr3_len`) has no valid `*N` form and is left untouched. 3′UTR / 5′UTR / intronic / special / unknown positions are untouched; only a simple certain point interval is rewritten.

New MockProvider unit tests cover in-3′UTR → `*N` (+ warning still fires), the last 3′UTR base, and the past-transcript guard.

## Conformance

`NG_012337.1(NM_003002.2):c.[274;600]` now normalizes to **`c.[274=;*120=]`** — the coordinate matches mutalyzer's `c.[274;*120]`, leaving only the explicit-identity `=` divergence. That row's `normalized` axis is reclassified `known_bug{#920}` → `accepted_divergence` (`ferro-policy-654`), now **identical to its genomic-axis entry** and the standalone `NG_012337.1:274` case.

### Supersedes #985
#985 reclassified this same row *without* fixing the coordinate (two-part `=`+coordinate note). This PR fixes the coordinate and leaves a clean `=`-only reclassification, so **#985 should be closed** in favor of this one.

## Validation

Re-blessed against the prepared reference: `normalized` / `genomic` axes pass, no XPASS, idempotent. Full suite green (7801), clippy `-D warnings` + `fmt --check` clean, `generate_conformance_summary --check` byte-identical.

Part of #326.

> Note: touches `cases.json` / `failure-patterns.md`, shared with #920 PRs #983/#984/#986 — whichever merges later needs a trivial rebase + summary regen.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -34,7 +34,7 @@ pub mod validate;
 use crate::coords::{hgvs_pos_to_index, index_to_hgvs_pos};
 use crate::error::FerroError;
 use crate::hgvs::edit::{Base, InsertedSequence, NaEdit, ProteinEdit, RepeatCount, Sequence};
-use crate::hgvs::interval::{Interval, ProtInterval};
+use crate::hgvs::interval::{CdsInterval, Interval, ProtInterval};
 use crate::hgvs::location::{
     AminoAcid, CdsPos, GenomePos, ProtPos, RnaPos, SpecialPosition, TxPos,
 };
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1024, opened 2026-07-10
feat: HGVS arbitration — explain why ferro and other tools disagree, and file bugs

nh13:
## Summary

Adds a workflow for understanding **why** ferro-hgvs and another HGVS tool (Mutalyzer by default, or any tool via paste) disagree on a variant, and — when ferro is the one at fault — filing a good bug in one step.

- `ferro arbitrate <variant> --reference <dir>` — reproduces ferro's result, obtains the other tool's result (auto-fetch Mutalyzer, or `--other-output` to paste any tool's answer), and reports a verdict: whether the two are the **same variant** (and if so, which spelling the spec mandates), **genuinely different edits**, a **basis mismatch** (different transcript version / coordinate axis), or **inconclusive** (with the reason). Text or JSON output.
- `ferro bug-report` (and the convenience flag `ferro arbitrate --bug-report`) — builds a prefilled GitHub new-issue URL you review and submit. No GitHub auth or token required; the URL carries only title+body (so it never 404s a read-only external user), and environment/paths are omitted by default.
- An `arbitrate-hgvs` Claude Code skill that drives the same flow conversationally and applies the governing spec rule on the cases the deterministic engine leaves to interpretation.

## Why the verdict is trustworthy (the honesty property)

The same/different verdict comes from an oracle that reduces both tools' outputs to reference-anchored SPDI edits and compares the edited reference sequences over a shared window. It shares **no code** with ferro's normalizer, so it can — and does — find ferro itself at fault rather than grading its own homework. The comparison is roll-invariant, so two shift-equivalent spellings of the same edit correctly compare equal. A structural test enforces the normalizer independence, and an integration test drives a committed corpus of real ferro/Mutalyzer discrepancies (including cases where Mutalyzer is spec non-compliant) and asserts arbitrate never falsely blames ferro — verified holding on a real prepared reference.

## Usage

```bash
# Auto-fetch Mutalyzer and compare:
ferro arbitrate "NM_000255.4:c.445_446insA" --reference ferro-reference/

# Paste any other tool's output (offline, any tool):
ferro arbitrate "NM_000255.4:c.445_446insA" --reference ferro-reference/ \
  --other-output "NM_000255.4:c.445dup"

# File a bug in one step (only files when ferro is actually implicated):
ferro arbitrate "<variant>" --reference ferro-reference/ --other-output "<other>" --bug-report
```

## Testing

- Full suite green (7962 tests, 0 failed). The arbitration-specific and honesty-corpus tests run against a real prepared reference in a `FERRO_MANIFEST`-gated path (they skip cleanly in ordinary CI without the multi-GB reference).
- Spec citations are CI-gated verbatim against the vendored HGVS spec submodule via `generate_spec_citations -- --check`.
- New code ships in the default binary un-gated, adds no new crate dependency, and does not pull in the benchmark/web-service stacks.

## Notes for reviewers

Suggested reading order (11 commits): start with **`feat(arbitrate): normalizer-independent arbitration oracle`** — it is the heart of the feature — then `feat(cli): ferro arbitrate and bug-report commands`, then `feat(skill): arbitrate-hgvs skill`. The two `fix:` commits are follow-ups discovered during manual testing (bare-`NM_` intronic projection to the NC chromosome frame; correct packaging of the shipped skill).

Opening as a draft for early review. Not tied to an existing issue.

## Summary by CodeRabbit

- **New Features**
  - Added `ferro arbitrate` to compare Ferro results with other HGVS outputs, including Mutalyzer.
  - Added text and JSON verdicts for equivalent variants, disagreements, parsing issues, basis mismatches, and inconclusive results.
  - Added `ferro bug-report` to generate evidence-based issue reports, with optional notes and environment details.
  - Added transcript-to-genomic frame projection and HGVS specification citations.

- **Documentation**
  - Added CLI guidance for arbitration workflows and bug reporting.
  - Added a documented arbitration skill for investigating tool disagreements.


nh13:
@coderabbitai pause

nh13:
Rebased onto `main` (`58f8215`). The branch had drifted 209 commits behind its original base (v0.7.1, 9 July) and was `DIRTY`; it now merges cleanly.

Three textual conflicts, all resolved by keeping both sides:

- **`Cargo.toml`** — `main` added the two spec-generator `[[bin]]` targets and several examples; this branch adds `[[example]] generate_spec_citations`. Both kept.
- **`tests/it/main.rs`** — adjacent `mod` declarations (`bug_report_cli` vs `build_transcript_library_parity`). Both kept, in lexical order.
- **`src/bin/ferro.rs`**, twice:
  - The `Commands::Project` arm: `main` grew `strict_reference` / `error_mode` / `ignore` / `reject` and an `ErrorConfig`, while this branch still carried the old shape plus the two new `Arbitrate` / `BugReport` arms. Kept `main`'s newer arm and appended the two new ones.
  - The `#[cfg(test)]` block at the end of the file. Both sides opened a `mod` *inside* the conflict region and shared the single closing brace after it, so deleting the markers alone would have left two opens against one close. Each `mod` now carries its own `#[cfg(test)]` and its own closing brace.

**A clean apply was not the same as a correct result** — two things only showed up on build and test:

1. `infer_genome_build` moved onto the `ReferenceProvider` trait during the intervening commits, so the call in `fetch_other_result` no longer resolved on the concrete provider. Added the trait import.

2. `issue_1197_required_error_config::entry_points_do_not_build_a_defaulting_normalize_config` failed, naming `run_arbitrate`. That guard landed on `main` after this branch was cut: a defaulting `NormalizeConfig` constructor supplies a silent *lenient* `error_config`, which is how `--error-mode` stayed inert for five months (#1181, #1197). `arbitrate` carries no error-mode flags, so there is no user config to thread through; it now names `ErrorConfig::lenient()` explicitly via `for_entry_point`, following the same idiom `run_project` uses for its 3'-default. Lenient is the right choice here on the merits — arbitration compares ferro's best-effort rendering against another tool's, so refusing a repairable input would report a disagreement neither tool actually has.

Both fixes were amended into the commit that introduced the code (`feat(cli): ferro arbitrate and bug-report commands`), not added as follow-ups; the history is still the same 11 logical commits.

Verification on the rebased branch: `cargo fmt --check` clean, `cargo clippy --all-features --all-targets -- -D warnings` clean, and the full suite green under all three oracles (`FERRO_ASSERT_IDEMPOTENT` / `REPARSE` / `IN_BOUNDS`) — **9352 passed, 0 failed**.

One note for anyone reproducing locally: six additional failures on the first run were a *stale generated fixture*, not the rebase. `tests/fixtures/grammar/hgvs_spec_normalization.json` is gitignored and this worktree still held a 9 July copy; regenerating it (`cargo run --features dev --bin generate_spec_fixture`, then `generate_spec_enumeration`) cleared all six. CI regenerates before its test run, so it would never have seen them.

nh13:
@coderabbitai review

Issue #1086, opened 2026-07-21
fix(project): axes disagree for one input; r. axis silently drops intronic offsets


nh13:
`VariantProjector` can return mutually inconsistent answers for a single input depending on which axis is requested. Three defects are described below, reproduced on `main` @ v0.9.1 (`6d39853`) with `ferro project --axis <a> --reference <reference-dir>`. They were surfaced by the spec-enumeration projection pass (#1085), where they are currently recorded as pinned baselines rather than fixed.

A fourth case (D4) was filed here in error and has been withdrawn — see the bottom of this issue.

Ordered by severity.

## D5 — the `r.` axis silently discards intronic offsets, naming a different base

```
NM_000552.3:c.3675-45_3692del
  c-axis: NM_000552.3:c.3675-44_3693del    intronic offset preserved
  r-axis: NM_000552.3:r.(3676_3694del)     offset dropped: 3675-45 became 3676
```

```
NM_004006.2:c.123-65_-50   ->  r-axis: NM_004006.2:r.(123_-50)
```

`c.3675-45` is 45 nucleotides inside an intron. That nucleotide is not present on the mature RNA at all, so `r.3676` does not describe the same position less precisely — it describes a **different, existing base**. The output is well-formed and silently wrong, which makes it worse than an error.

Expected: either decline to project (intronic positions have no mature-RNA equivalent), or use the spec's splicing notation (`RNA/splicing.md`), but never map an intronic offset onto a plain exonic `r.` position.

## D2 — the `n.` axis returns an `r.` string for `r.` inputs

```
LRG_199t1:r.11u>g
  n-axis: LRG_199t1:r.11u>g       an r. string, unchanged
  c-axis: LRG_199t1:c.-234T>G     converts correctly
```

`VariantProjection::noncoding` is documented as "a single field that always means the `n.` form". For an `r.` input it passes the `r.` form through unchanged, so a caller asking for `n.` receives `r.` — different coordinate system, lowercase alphabet, and `u` instead of `t`. The `c.` axis handles the same input correctly, so this is not a missing-capability case.

## D3 — `c.` and `r.` disagree on genomic-context framing

```
NC_000023.11(NM_004006.2):c.4072-1234_5155-246del
  c-axis: NM_004006.2:c.4072-1234_5155-246del                context dropped
  r-axis: NC_000023.11(NM_004006.2):r.(4073_5156del)          context kept
```

[...]

Issue #879, opened 2026-06-27
fix(project): NG_012337.1 SDHD g.→c. projection declines ("variant does not overlap transcript")


nh13:
`NG_012337.1:g.13124_13125del` (and the whole locus, g.13120–13127) raises `TranscriptNotOverlapping` for every SDHD transcript hosted on `NG_012337.1` (`NM_003002.2` and `.4`), so the g.→c. projection declines before protein prediction and `project_variant_all` yields zero pairs.

Surfaced by the #326 conformance row `NG_012337.1:g.13124_13125del` → expected `[(NG_012337.1(NM_003002.2):c.480_*1del, NG_012337.1(NP_002993.1):p.(Ter160CysextTer29))]`; ferro currently emits an empty projection.

Also a transcript-version divergence: ferro's cdot enumerates `NM_003002.4`, the corpus expects `.2` (affects both the c. and p. accessions, #763-family).

Split out from #857 (C-terminal extension / stop-loss), which fixed the protein-axis sibling `NG_012337.1(NM_012459.2):c.297_*1del`. This is an NG_→transcript overlap/placement decline (#851-family), not stop-loss.

Part of #326.

nh13:
## Diagnosis: mis-scoped — resolved by #857, residual is version (#763) + cross-isoform (#646) divergence

Re-diagnosed against the blessed reference via `project_variant_all` (the actual `coding_protein_descriptions` axis path), not the `--transcript` CLI path. The original "declines / TranscriptNotOverlapping / empty projection" framing came from the CLI `--transcript` path, which does **not** de-anchor the `NG_` input into the `NC_` frame. The real harness path (`project_variant_all`) de-anchors first and **does** project — it does not decline.

**On the #857 branch (PR #880), `project_variant_all('NG_012337.1:g.13124_13125del')` now returns:**

```
tx=NM_003002.4      coding=NG_012337.1(NM_003002.4):c.480_*1del   protein=NG_012337.1(NP_002993.1):p.(Ter160CysextTer29)
tx=NM_001276504.2   coding=NG_012337.1(NM_001276504.2):c.363_*1del protein=NG_012337.1(NP_001263433.1):p.(Ter121CysextTer29)
tx=NM_001276506.2   coding=NG_012337.1(NM_001276506.2):c.*178_*179del  protein=None
tx=NM_001276503.2   coding=NG_012337.1(NM_001276503.2):c.*77_*78del    protein=None
```

[...]

Issue #693, opened 2026-06-15
fix(project): r. consequence prediction end-to-end gaps on the rna_description corpus (#485 follow-up)


nh13:
The `predict_rna` surface (`VariantProjection.rna` + Python getter) lands in #485 with 9 unit tests on synthetic `MockProvider` transcripts — correct in isolation. **#485 deliberately does NOT wire the `rna_description` conformance runner**, because against the real prepared manifest only 2 of 16 corpus rows pass and the wiring + fixes form one coherent change best reviewed together. This issue owns that change end-to-end.

## Scope of this issue

- [ ] Wire the `axis_rna_description` runner in `tests/mutalyzer_normalize_tests.rs` to read `result.rna` (mirror `axis_protein_description`).
- [ ] **Reference framing**: r. output drops the input `NG_(NM_)` genomic context and substitutes a gene symbol (e.g. expected `NG_007485.1(NM_058195.3):r.(141_142del)`, got `NM_058195.3(CDKN2A):r.(141_142del)`). `predict_rna` inherits the normalized coding form's bare-transcript accession; thread the input's `genomic_context` through. This single fix flips ~7 rows (the r. bodies are already correct).
- [ ] `c.297_*1del` renders `r.(297_*297del)` instead of `r.(297_*1del)` — a `*N` 3'UTR render bug distinct from framing.
- [ ] n.→r. CDS-relative renumbering not applied end-to-end (`n.204_205insATC` → `r.(204_205…)` not `r.(44_45…)`).
- [ ] Bare `n.` inputs (e.g. `NM_003002.4:n.206_210del`) routed through genome projection (errors "no genomic parent"); r. prediction is transcript-local.
- [ ] `transcript_of` test helper does not handle allele inputs (cis-allele errors "could not infer transcript_id").
- [ ] Disposition the corpus rows in `cases.json` against the real manifest: pass the now-correct rows; `known_bug`/`spec_citation` the genuine divergences; for the `Err`-producing rows (which the harness forbids annotating) fix or exclude per their blocker.

## External blockers (for the `Err` rows)

- Version coverage: `NM_003002.2` absent from the prepared manifest — #645, #672.
- Intron→genome mapping (`c.677-21_704+62del` → "Cannot map tx position 757 to genome") — #480 and the NG_/LRG_ projection work.
- Cross-reference insert inside delins does not parse — #692.

## Done when

[...]

PR #1090, opened 2026-07-21
fix(project): retain the input's genomic context on the c./n. axis

nh13:
Fixes D3 from #1086: the `c.` and `r.` projection axes disagreed on genomic-context framing.

```
NC_000023.11(NM_004006.2):c.4072-1234_5155-246del
  before  c-axis: NM_004006.2:c.4072-1234_5155-246del
  after   c-axis: NC_000023.11(NM_004006.2):c.4072-1234_5155-246del
          r-axis: NC_000023.11(NM_004006.2):r.(4073_5156del)     unchanged
```

## The `c.` axis was the defect, not `r.`

The issue originally described this neutrally — "whichever is correct, they should agree". The spec is not neutral. `background/refseq.md:47-49`:

> intronic sequences are considered to be **within the boundaries** of a transcript reference sequence and may only be used to describe a variant **when a genomic reference sequence identifier is provided** … **not correct:** `NM_004006.2:c.357+1G>A`, `LRG_199:c.357+1G>A`. **correct:** `NC_000023.10(NM_004006.2):c.357+1G>A`, `NG_012232.1(NM_004006.2):c.357+1G>A`, `LRG_199t1:c.357+1G>A`.

The D3 input is intronic, so `project --axis c` was emitting the spec's own "not correct" shape. `c.93+1G>T` and `NG_012232.1(NM_004006.2):c.357+1G>A` did the same — the latter is verbatim one of the spec's "correct" examples, which ferro was turning into the adjacent "not correct" one.

This was also inconsistent with ferro elsewhere: `ferro normalize` already retains the context, and ~38 `NC_/NG_(NM_):c.` intronic corpus rows pass because of it. Only `project --axis c` dropped it.

Dropping the wrapper is genuinely lossy, which is presumably why the spec mandates rather than merely prefers it: an intron offset is defined only by a transcript↔genome alignment, and `NM_004006.2:c.4072-1234` does not record which build placed that intron — GRCh37 and GRCh38 differ.

## The rule implemented

**Preserve the caller's framing** — the `c.`/`n.` axis retains the genomic context whenever the input supplied one, and stays bare otherwise. Not intronic-only.

`refseq.md:138-140` prescribes the parenthesised form for *any* `c.` description "based on a genomic reference sequence", with no exonic carve-out, and this matches what `normalize` and `--axis r` already do on exonic input too. It also needs no positional classification, so one projection's axes always agree on their reference.

Exonic behaviour is pinned both ways by tests: `NM_004006.2:c.79A>G` stays bare; `NC_000023.11(NM_004006.2):c.79A>G` keeps the parent.

`n.` is changed alongside `c.` — `refseq.md:157` states the identical requirement for non-coding transcripts.

## Implementation

`reframe_transcript_axis_from_input` / `apply_transcript_axis_framing` in `src/project/projector.rs`, the `c.`/`n.` analogue of the existing `r.` reframing from #693, applied in `project_single_inner`. New `tests/it/issue_1086_c_axis_genomic_context.rs`, 7 manifest-gated tests (RED 5 fail / 2 pass → GREEN 7/7).

## One test change reviewers should look at

`issue_879_ng_g_to_c_overlap::ng_context_coding_input_keeps_pivot_path_unchanged` pinned `coding == "NM_TEST.1:c.4A>C"` for input `NG_TEST.1(NM_TEST.1):c.4A>C` — a hermetic mock pin of this very defect, on an exonic position, incidental to that test's actual subject (#879 routing, whose assertions are unchanged). It has been updated with an explanatory comment.

This is the only cost of the broad rule over an intronic-only one. If intronic-only is preferred on review, that test reverts and only the exonic-with-context row changes back; no corpus row is involved either way.

## Out of scope

3'-shift behaviour is untouched on both axes (`RNA/deletion.md:20` and `RNA/duplication.md:24` require `c.` and `r.` to differ there — originally filed as D4, withdrawn). The protein axis stays bare, which is deliberate and spec-correct; the `bare-np-protein` corpus rows are all scoped `"axis": "protein_description"` and do not generalise.

Follow-up worth its own issue: `polya_multiaxis_projection` (`projector.rs:4997`) and `terminus_multiaxis_projection` (`:5075`) still explicitly null the context on their echoed `c.` form — same defect class, left alone here.

## Verification

`cargo nextest run --features dev` with `FERRO_MANIFEST` unset: **8074 passed, 0 failed** (8067 base + 7 new). fmt and clippy `--all-targets -D warnings` clean.

**Corpus fallout: zero.** The manifest-gated run shows the identical 5 pre-existing failures before and after, verified by stashing and re-running on the clean tree. Nothing re-blessed.

Refs #1086.

### DIFF
--- a/src/project/projector.rs
+++ b/src/project/projector.rs
@@ -1761,6 +1761,73 @@ impl<P: ReferenceProvider + Clone> VariantProjector<P> {
         Some(rna)
     }
 
+    /// Re-frame a derived transcript-coordinate axis (`c.` / `n.`) so its
+    /// accession carries the *input's* genomic context (#1086).
+    ///
+    /// The `coding`/`noncoding` axes are built on the bare transcript accession
+    /// (`NM_004006.2`), which silently discarded an `NC_`/`NG_` parent the input
+    /// supplied. For an **intronic** position that bare form is not merely a
+    /// stylistic loss — it is spec-invalid. `background/refseq.md:47-49`:
+    ///
+    /// > intronic sequences are considered to be **within the boundaries** of a
+    /// > transcript reference sequence and may only be used to describe a
+    /// > variant when a genomic reference sequence identifier is provided …
+    /// > **not correct:** `NM_004006.2:c.357+1G>A` … **correct:**
+    /// > `NC_000023.10(NM_004006.2):c.357+1G>A`,
+    /// > `NG_012232.1(NM_004006.2):c.357+1G>A`.
+    ///
+    /// (Restated for coding transcripts at `refseq.md:136` and for non-coding
+    /// transcripts — the `n.` axis — at `refseq.md:157`; the parenthesised
+    /// rendering is specified at `refseq.md:138-140`.)
+    ///
+    /// The rule applied is **preserve the caller's framing**: retain the parent
+    /// whenever the input supplied one, render bare whenever it did not. This is
+    /// a superset of the spec's mandatory (intronic / flanking) case, so the
+    /// "not correct" bare-intronic form can never be produced from a parented
+    /// input; and it is what `refseq.md:138-140` prescribes for *any* `c.`
+    /// description "based on a genomic reference sequence", with no exonic
+    /// carve-out. It also matches the framing `ferro normalize` and the `r.`
+    /// axis ([`Self::reframe_rna_from_input`]) already preserve, so the axes of
+    /// one projection agree on their reference.
+    ///
+    /// A parent is never fabricated: a bare input stays bare (#121). Unlike the
+    /// `r.` path this deliberately leaves `gene_symbol` alone — the `c.`/`n.`
+    /// symbol is already the input's own on every site that calls this.
+    fn reframe_transcript_axis_from_input(
+        axis: Option<HgvsVariant>,
+        input: &HgvsVariant,
+    ) -> Option<HgvsVariant> {
+        let mut axis = axis?;
+        let context = input
+            .accession()
+            .and_then(|a| a.genomic_context.as_deref().cloned());
+        Self::apply_transcript_axis_framing(&mut axis, context.as_ref());
+        Some(axis)
+    }
+
+    /// Stamp `context` onto every `c.`/`n.` leaf of `axis` (descending into an
+    /// allele so its first-member-derived prefix is re-framed), or strip any
+    /// context when `context` is `None`. See
+    /// [`Self::reframe_transcript_axis_from_input`] for the rule and citation.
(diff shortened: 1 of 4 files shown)


nh13:
Pushed `436ff0e` — *keep the input's genomic context on the two multi-axis echoes* (signed).

`polya_multiaxis_projection` and `terminus_multiaxis_projection` were explicitly nulling the genomic context on their echoed `c.` form. Same defect as `4d1086a`, and **both were broken identically**:

| input | c. before | c. after |
|---|---|---|
| `NG_012337.1(NM_003002.2):c.*830del` | `NM_003002.2:c.*830del` | `NG_012337.1(NM_003002.2):c.*830del` |
| `NG_012337.1(NM_003002.2):c.pter/qter/pter_qterdel` | bare | parented |
| `NM_003002.2:c.*830del` (bare guard) | bare | unchanged |
| `LRG_9t1:c.qterdel` (bare guard) | `LRG_9t1:c.*824del` | unchanged |

The `g` axis is unchanged throughout.

Each function's doc comment justified the stripping as "matching how `coding` is reported on the normal path" — which is precisely the rationale `4d1086a` invalidated. Neither echoes a different reference frame: each sits alongside a `genomic` axis anchored in the very parent being discarded.

The terminus path is **affirmatively wrong bare**: `pter`/`qter` are genomic-frame landmarks, and `project_cds_terminus_to_parent` can only resolve them *because* a parent was supplied. Emitting the result without that parent discards the thing that made the answer computable.

Both now route through `reframe_transcript_axis_from_input`, so there is one framing rule rather than three.

**Incidental fix worth a look:** the test harness added in `4d1086a` used a hand-written `ArcProvider` newtype that had missed delegating `get_sequence_length`, silently inheriting the trait default (`Err(ReferenceNotFound)`). Invisible to the D3 tests, fatal to both new paths. Replaced with the library's complete blanket `impl ReferenceProvider for Arc<T>`.

**Suite 8078 passed / 0 failed** (baseline 8074 + 4 new), `FERRO_MANIFEST` unset; fmt and clippy clean; `4d1086a`'s 7 tests still pass. **Corpus fallout: zero** — with a manifest, the same 5 tests fail identically before and after (`vps13d_*` ×3, `comparator_provenance_reference_identity_matches_live`, `axis_normalized` at 4 FAIL both sides), all pre-existing local-reference drift.

Two notes for reviewers: the poly-A reproduction is reference-data-sensitive (the valid window is `*829..*842` for `NM_003002.2`); and `terminus_multiaxis_projection`'s bare-LRG branch (`lrg_genomic_parent`) appears unreachable in practice, since normalize resolves `pter`/`qter` numerically for LRG inputs before the short-circuit sees them — left alone rather than removed speculatively.

Closes the last open item on #1086.

Issue #1079, opened 2026-07-20
tracking: HGVS spec divergences (successor to #83)


nh13:
Successor tracker to #83, which was closed 2026-05-19 as an umbrella (superseded by #72/#73/#74/#75/#81/#82/#84/#87) with all 22 checkboxes still unticked. `examples/generate_spec_fixture.rs` nonetheless still stamps `todo: .../issues/83` onto every divergent row, so each regeneration re-points live findings at a closed issue. This issue replaces that target.

## The count, corrected

`tests/fixtures/grammar/hgvs_spec_normalization.json` currently reports **243** rows with status `diverges` / `false-acceptance` / `needs-reference`. Roughly 40% of those are not variants at all.

| | count |
|---|---:|
| Reported live divergences | 243 |
| Harvester artifacts — edit-less position references normalized to `input=` | **99** |
| **Genuine divergences** | **144** |

> [!NOTE]
> The numbers in this section are as-filed (2026-07-20). For the current state after #1082/#1091/#1096, see **Current status** at the bottom of this issue.

The artifacts are bare coordinates lifted out of prose (for example "positions `g.12345678` to `g.12345901`") and then scored as variants: ferro normalizes `g.12345678` to `NC_000023.11:g.12345678=` and the row lands in `diverges`. The generator's own documentation says edit-less coordinates are dropped; the filter is leaky. Note the filter must **not** drop genuine edit-less negatives such as `c.123-65_-50`, which `checklist.md:26` calls out as an incomplete range that should be rejected rather than silently `=`-ed.

Corrected status shape:

| status | reported | genuine |
|---|---:|---:|
| `diverges` | 133 | **34** |
| `false-acceptance` | 76 | **76** |
| `needs-reference` | 34 | **34** |
| **total** | **243** | **144** |

`diverges` is the headline correction. The `false-acceptance` bucket is untouched, which is expected: those rows come from `<code class="invalid">` markers and prose negatives, all genuine spec-invalid inputs that ferro accepts.

Of #83's own cited examples that could be matched to fixture rows (31 of 62 backticked strings), **19 are now resolved** (16 `preserved`, 3 `correctly-rejected`) and 12 remain live.

## Themed breakdown of the genuine 144

| theme | count | diverges | false-acceptance | needs-reference |
|---|---:|---:|---:|---:|
| unclassified (needs manual triage) | 52 | 1 | 28 | 23 |
[...]

PR #1091, opened 2026-07-21
fix(parser): reject six classes of spec-forbidden variant description

nh13:
Burns down six MUST-level themes from #1079. Fixture `false-acceptance` **76 → 57** (19 rows), `correctly-rejected` **75 → 94**.

Each theme is a spec-stated prohibition ferro accepted. The spec is RFC 2119 (`recommendations/style.md:9`); every rule below is MUST-level ("is not correct" / "not allowed" / "can not be described"), verified against the vendored submodule.

| commit | theme | disposition | rows |
|---|---|---|---:|
| `16a5973` | del/dup size suffix on a point anchor | reject; lenient/silent canonicalise `g.123del6` → `g.123_128del` | 5 |
| `180006a` | start-codon written as a substitution | reject | 3 |
| `6c192ae` | one-nucleotide inversion | reject | 2 |
| `fe03955` | frameshift terminating immediately (`fsTer1`) | reject | 1 |
| `fe94bd3` | amino acids listed after a stop | reject | 1 |
| `1ce839c` | multi-nucleotide substitution (`GC>TT`) | reject; names the canonical delins | 7 |

Citations: `checklist.md:49` and `:33`, `DNA/deletion.md:117`, `DNA/duplication.md:140`, `RNA/deletion.md:49`; `protein/substitution.md:49`, `checklist.md:65`, `protein/extension.md:28`; `DNA/inversion.md:16`; `protein/substitution.md:20` and `protein/frameshift.md:39`; `protein/delins.md:45` and `protein/insertion.md:43`; `DNA/delins.md:73-77` and `DNA/substitution.md:30`.

## The last one is an architectural finding, not a spelling nit

`c.79GC>TT` produced **`c.79delinsTT`** — delete one base, insert two. That is a different variant from the stated one, silently discarding a base of the given reference.

`parse_multibase_substitution` (`src/hgvs/parser/edit.rs:188`) folds `GC>TT` into `NaEdit::Delins { deleted: None, deleted_length: None }`, discarding the reference bases; the location was already parsed, so a point anchor stays a point.

**The two entry points did not share a front end.** `parse_hgvs_with_config` (CLI) runs the preprocessor first, and `correct_old_substitution_syntax` (`src/error_handling/corrections.rs:1539`) already extended a point anchor to `start + ref_len - 1` — so the grammar never saw the lossy form. `parse_hgvs`, the strict entry point the spec-fixture generator drives, bypasses the preprocessor and hit it directly. A spec rule existed on one path and not the other.

The 7 rows that moved are exactly the spec's own invalid spellings: 4 DNA plus `r.4gc>ug`, `r.76aa>ug`, `r.76_77aa>ug`.

## Deliberate non-changes

- **`fsTer2` still parses in every mode**, and is tested. `protein/frameshift.md:22` — "the shortest frameshift variant possible contains `fsTer2`". Only `fsTer1` is wrong; flagging `fsTer2` would have been a serious false positive.
- `dup` size suffixes are reject-only in all modes: `DNA/duplication.md:143` explicitly declines to disambiguate them, so there is no derivable repair.
- Start-loss and one-nt inversion are reject-only for the same reason — the spec's three correct start-loss forms (`p.0`, `p.0?`, `p.(Met1?)`) need evidence the input does not carry, and reversing an inversion needs reference bases. The diagnostics name the alternatives.
- `ins<N>` (`checklist.md:33`) skipped: already `correctly-rejected` for an unrelated reason, so zero burn-down against real churn risk.
- `c.79G>TT` and `c.100_102>ATG` untouched — the multibase rule is filtered to reference runs of ≥ 2 bases.

## Real-data fallout — two records, both handled by the established mechanism

`tests/fixtures/bulk/clinvar_hgvs_unique_failure_expectations.json` already carries 82 curated kinds / 207 expected failures, of which 39 are `rejected_by_design` for spec reasons (`spec_self_cancelling_allele`, `deprecated_repeat_paren_no_brackets`, …). Two real ClinVar records join them:

- `NP_002476.2:p.Gln279ProfsTer1` — internally contradictory: `fsTer1` claims termination at codon 1 of the shifted frame while naming `Pro` there.
- `NM_003977.2:c.878_879AG>GT` — literally the spelling `DNA/substitution.md:30` names invalid, with the repair stated outright.

Both were hand-appended as `rejected_by_design` kinds. **`UPDATE_FAILURE_EXPECTATIONS=1` was deliberately not used** — it re-derives IDs from raw error strings and destroys the curated `kinds` taxonomy. That is a latent tooling bug worth its own issue.

## Known limitations

The multibase check is **source-keyed rather than AST-keyed**, because the grammar discards the reference bases and the AST cannot distinguish `c.79_80GC>TT` from a hand-written `c.79_80delinsTT`. The structural fix — preserving the reference run in `deleted` — would change `Display` for every `NN>NN` input and was not attempted here.

Themes 4 and 5 have mechanical repairs that are **not** wired into lenient mode; each needs a new W-code threaded through types, registry, python bindings, preprocessor and the audit fixture. Deferred deliberately; the diagnostics name the exact repair meanwhile.

Theme 1 changes lenient-mode behaviour: W3011/W3023 were `warn_accept`, and point-anchored size suffixes are now repaired or rejected rather than round-tripped. Two test files encoding the old contract were updated.

`NR_001566.1:c.107_108gc>ag` now trips the multibase rule before the coordinate-system rule, so its diagnostic text changed while its expectation label did not — the harness compares sets, not messages.

## Verification

`cargo nextest run --features dev` with `FERRO_MANIFEST` unset: **8113 passed, 0 failed** (8067 base + 46 new). `generate_spec_fixture -- --check` exit 0; fmt and clippy `--all-targets -D warnings` clean.

Refs #1079.

### DIFF
--- a/src/error_handling/corrections.rs
+++ b/src/error_handling/corrections.rs
@@ -3055,6 +3055,168 @@ fn take_ref_seq_run(bytes: &[u8], pos: usize) -> &str {
     std::str::from_utf8(&bytes[pos..j]).unwrap_or("")
 }
 
+/// Rewrite a deletion written as `<point>del<size>` into the canonical
+/// range form `<point>_<point + size - 1>del` (W3011).
+///
+/// Per HGVS `recommendations/checklist.md:49` the size-suffix form on a
+/// single-position anchor is **not allowed**:
+///
+/// > Descriptions like `g.123del3` are not allowed, correct is `g.123_125del`.
+///
+/// The spec states the replacement outright, so the repair is mechanical: a
+/// size `N` deletion anchored at `p` spans `p`..`p + N - 1`. `del1` collapses
+/// to a plain `del` (a single residue needs no range).
+///
+/// Only a **plain** point anchor is rewritten — the digit run must be
+/// preceded by the coordinate-prefix `.` (`g.123del3`, `c.76del3`) or an
+/// allele-member boundary `[` / `;` (`c.[123del3;200del4]`), so each member
+/// of a compound allele is repaired the same way a standalone description is.
+/// That deliberately excludes every anchor whose arithmetic is not a simple
+/// increment: range endpoints (`_102del3`), intronic offsets (`+5del3`),
+/// UTR positions (`-12del3`, `*10del3`) and uncertain positions
+/// (`(123)del3`). Those keep whatever behaviour their own rule prescribes.
+/// The rewrite is coordinate-system agnostic — `p + N - 1` holds on every
+/// axis — so no coordinate context needs to be tracked across the members.
+///
+/// Returns the rewritten string plus one [`DetectedCorrection`] per rewrite,
+/// spanning the whole `<point>del<size>` token.
+pub fn correct_point_del_size_suffix(input: &str) -> (Cow<'_, str>, Vec<DetectedCorrection>) {
+    let mut corrections = Vec::new();
+    let bytes = input.as_bytes();
+    if !has_non_protein_description(bytes) || !input.contains("del") {
+        return (Cow::Borrowed(input), corrections);
+    }
+
+    let mut result = String::new();
+    let mut last_copied = 0usize;
+    let mut i = 0usize;
+    while i + 3 <= bytes.len() {
+        if &bytes[i..i + 3] != b"del" {
+            i += 1;
+            continue;
+        }
+        // `delins` is a different edit; skip past it.
+        if bytes.get(i + 3..i + 6) == Some(b"ins") {
+            i += 6;
+            continue;
+        }
+        let Some(token) = point_size_token(input, i, 3) else {
+            i += 3;
+            continue;
+        };
+        let replacement = format!("{}_{}del", token.position, token.position + token.size - 1);
+        corrections.push(DetectedCorrection::new(
+            ErrorType::DelSizeSuffix,
+            &input[token.start..token.end],
+            replacement.clone(),
+            token.start,
+            token.end,
+        ));
+        result.push_str(&input[last_copied..token.start]);
+        result.push_str(&replacement);
+        last_copied = token.end;
+        i = token.end;
+    }
+
+    if corrections.is_empty() {
+        (Cow::Borrowed(input), corrections)
+    } else {
+        result.push_str(&input[last_copied..]);
+        (Cow::Owned(result), corrections)
+    }
+}
+
+/// A `<point><keyword><size>` token located in the input (e.g. `123del3`).
+struct PointSizeToken {
+    /// Byte offset of the first digit of the anchor position.
+    start: usize,
+    /// Byte offset one past the last digit of the size suffix.
+    end: usize,
+    /// The anchor position.
+    position: u64,
+    /// The size suffix.
+    size: u64,
+}
+
+/// Recognize `<point><keyword><size>` at `keyword_start`, where `keyword_len`
+/// is the length of the edit keyword (`del`/`dup`).
+///
+/// Returns `None` unless the anchor is a plain point position written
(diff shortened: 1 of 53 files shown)


nh13:
Pushed two further commits.

**`424fbe2` — insertion anchors that are not two flanking positions** (theme 7)

`p.Lys23insAsp` parsed and round-tripped. `protein/insertion.md:18` names it verbatim as not allowed. The existing `reject_single_position_insertion.rs` covered the DNA form only; the protein arm carried a "handled elsewhere" comment and was in fact unhandled. `g.123_125insG` was likewise accepted — `DNA/insertion.md:15` and `syntax.yaml` `dna.ins` require adjacent positions. Enforced only where adjacency is decidable from the description alone, so `c.-1_1insT` (no `c.0`), `c.100+1_101-1insT` and `c.100_*1insT` stay legal.

**Protein non-adjacency is deliberately NOT enforced, because the spec contradicts itself.** It was implemented, then the regenerated fixture flagged a new `parse-error`: `protein/alleles.md:61` publishes `p.[Ser68_Ala74insSerGln];[Ser68_Ala74=]` unmarked as valid — a 7-residue anchor that `protein/insertion.md:17` would forbid. Rather than reject a string the spec itself prints, the scope was narrowed and the non-enforcement pinned with tests. This is the only such conflict across all 1033 spec rows, and is worth raising with SVD-WG.

**`8f676cd` — preserve the stated reference run of a substitution** (issue #1092)

This **supersedes the source-keyed approach in `1ce839c`** and is the reason to review that commit and this one together.

`parse_multibase_substitution` discarded the stated `GC` at parse time, so the conformance check had to re-read the input string. Two consequences: it could not fire for programmatically built variants, and — more seriously — lenient mode rewrote `c.79GC>TT` → `c.79_80delinsTT` and the `GC` claim evaporated silently **even when false**.

That last point mattered because ferro already validates the same assertion for the neighbouring spelling: `src/normalize/validate.rs:87-104` checks a stated delins run against the reference (#486). So the *deprecated* `delGCinsTT` was verified while the *forbidden* `GC>TT`, asserting the identical thing, was not.

Now: `NC_TEST.1:g.257_258GC>TT` against a reference reading `AC` produced no warnings before, and produces `RefSeqMismatch { stated_ref: "GC", actual_ref: "AC" }` after.

**AST shape:** a provenance field, `NaEdit::Delins { substitution_reference: Option<Sequence> }`. A distinct `NaEdit` variant was rejected as disproportionate — `Delins` is matched in 139 places across normalize/project/spdi/vcf/python, each needing separately-reasoned semantics. Reusing `deleted` was rejected because it changes behaviour at all 139 sites *and* renders the deprecated `delGCinsTT`. A separate field is inert: every existing consumer still sees `deleted: None` exactly as before. Cost is 72 compiler-enforced `substitution_reference: None` additions.

**`Display` is unchanged for every input** — both arms destructure with `..`, and two tests assert `delGCins` is never rendered, for the programmatic AST and for every parsed spelling in lenient and silent mode.

Fixture shows **no row movement** (1033 / 715 / 54 / 97 / 133 / 34), which is the correct outcome for a refactor of *how* a rule is enforced rather than *what* it enforces.

Suite **8140 passed, 0 failed**; `--check`, fmt, clippy `-D warnings` and `--features python` all clean.

⚠️ **`8f676cd` is unsigned.** `git commit -S` failed five times across two attempts with `1Password: failed to fill whole buffer`; the documented fallback was used. **It needs re-signing before merge.**

One residual: the lenient *repair* path still re-derives the run from the original string, which is unavoidable since the preprocessor rewrite is textual. It is narrowed to a single non-allele variant carrying one such description. The conformance rule itself is fully AST-keyed.

Issue #1116, opened 2026-07-22
Protein cis-allele delins coalescing is input-order-dependent


nh13:
Follow-up to #1103 / #1106.

#1106 makes the **nucleotide** cis merge input-order-independent by sorting cis members into a canonical merge order before `merge_consecutive_edits`. The **protein** axis is canonicalized by a separate pass — `coalesce_protein_adjacent_substitutions` (`src/normalize/merge.rs`) — which runs *before* that sort and explicitly bails when the members are not already in ascending residue order:

```rust
// Require ascending, duplicate-free residue order. A descending or
// duplicate (`n,n`) allele is out of scope ...
if subs.windows(2).any(|w| w[1].pos <= w[0].pos) {
    return None;
}
```

So a protein cis allele whose members are authored in descending residue order does not coalesce, and only gets the #1098/#1101 post-normalize display sort applied:

| input | current output | expected |
| --- | --- | --- |
| `NP_003997.1:p.[Arg76Ser;Cys77Trp]` | `NP_003997.1:p.Arg76_Cys77delinsSerTrp` | same |
| `NP_003997.1:p.[Cys77Trp;Arg76Ser]` | `NP_003997.1:p.[Arg76Ser;Cys77Trp]` | `NP_003997.1:p.Arg76_Cys77delinsSerTrp` |

The current output for the descending input is a form the spec calls out as incorrect — `protein/substitution.md:23`:

> changes involving two or more consecutive amino acids are described as a deletion/insertion variant (delins) […] the description `p.Arg76_Cys77delinsSerTrp` is correct, the description `p.[Arg76Ser;Cys77Trp]` is not correct.

Two permutations of the same member set therefore normalize to two different strings, one of which is spec-invalid — the exact order-dependence #1103 removed on the nucleotide axis.

This is recorded as a known limitation in #1106 (`tests/it/allele_grammar_corners.rs::protein_descending_order_members_reorder_to_residue_order`).

### Proposed fix

Sort the collected members into ascending residue order inside `coalesce_protein_adjacent_substitutions` before the run detection, instead of declining on non-ascending input. Reordering is meaning-preserving there: the function has already established that every member is a plain single-residue substitution on one shared accession, and the post-normalize display sort (#1098/#1101) canonicalizes protein cis members into exactly that order anyway. Duplicate residue positions (`n,n`) must still decline — that is a contradiction, not a delins.

### Acceptance

[...]

Issue #1119, opened 2026-07-22
fix(normalize): protein delins is never canonicalized to its minimal form (sub / del)


nh13:
## Summary

ferro performs **no** canonicalization of protein `delins` edits in `normalize`. A `delins` that is structurally a smaller/canonical edit is left verbatim, diverging from the spec's minimal-form rule (`protein/delins.md`). Examples (all left unchanged today):

```
NP_...:p.Val559delinsGly            # should be p.Val559Gly (1→1 delins is a substitution)
NP_...:p.(Val559_Glu562delinsGlu)   # should be p.(Val559_Glu561del)  (delins.md:53-56)
```

The nucleotide axis already does this (`c.145_147delinsTGC` → `c.145C>T` via `canonicalize_delins` in `src/normalize/rules.rs`), but that path is `Base`/reference-specific and does not apply to protein.

## Spec basis

- `protein/delins.md:53-56` — `p.(Val559_Glu562delinsGlu)` is **not correct**; the trailing `Glu562→Glu` is no change, so the canonical form is `p.(Val559_Glu561del)`.
- `protein/delins.md` — a delins of one residue replaced by one residue is a substitution.

## Two scopes

**(a) Reference-free (tractable now).** For a protein delins whose location spans a small, fully-named residue range, boundary residues that match the named location endpoints can be trimmed from the AST alone, and the result reduced to a `sub`/`del` when the new endpoints remain named:
- `p.Val559delinsGly` → `p.Val559Gly` (single-residue delins → substitution).
- `p.Val559_Glu560delinsSerGlu` → `p.Val559Ser` (trailing `Glu`==`Glu560` trimmed → single named residue → sub).
- `p.Val559_Glu560delinsValSer` → `p.Glu560Ser` (leading `Val`==`Val559` trimmed).

**(b) Reference-gated (deferred).** The headline `Val559_Glu562delinsGlu` → `Val559_Glu561del` case trims a boundary residue off a range spanning **>2** residues, so the new endpoint (residue 561) is not named in the input and cannot be emitted without the **protein sequence**. ferro does not currently resolve a protein sequence for a standalone `NP_:p.` accession (it does not validate protein reference residues — `p.Ala562Gly` on a mismatched reference is accepted). A full fix needs `NP_`→`NM_`→CDS-translation protein-sequence resolution, which is a larger capability.

## Ask

[...]

PR #1127, opened 2026-07-22
fix(normalize): scope the protein coalesce to-Ter decline to the affected run

### LINKED ISSUE #1125: Protein cis-allele coalesce declines the whole allele on any to-Ter member, leaving unrelated upstream runs un-coalesced
### Summary

`coalesce_protein_adjacent_changes` (`src/normalize/merge.rs`) declines the **entire** cis protein allele when **any** member is a to-`Ter` (nonsense) substitution — even when that `Ter` is disjoint from, and downstream of, an otherwise-coalescible adjacent run. The upstream run is left as a bracket instead of collapsing to the spec-preferred `delins`.

### Repro (reference-free, `MockProvider`)

| input | actual | expected |
|-------|--------|----------|
| `NP_003997.1:p.[Cys100Ser;Asp101Gly;Arg200Ter]` | `NP_003997.1:p.[Cys100Ser;Asp101Gly;Arg200Ter]` (unchanged) | `NP_003997.1:p.[Cys100_Asp101delinsSerGly;Arg200Ter]` |
| `NP_003997.1:p.[Cys100Ser;Asp101Gly]` (control, no Ter) | `NP_003997.1:p.Cys100_Asp101delinsSerGly` | — |

The `Cys100/Asp101` run is strictly adjacent and entirely 5′ of the `Arg200Ter` nonsense, so it is a legitimate run that should coalesce per `protein/substitution.md:23` / `protein/delins.md:18`. The lone downstream `Ter` should not block it.

### Cause

The per-member build loop returns `None` from the whole function on the first to-`Ter` member:

```rust
if alternative == Some(AminoAcid::Ter) {
    return None;
}
```

This is deliberately conservative — the spec forbids listing residues *after* a stop, so a `delins…Ter…` run is invalid — but the guard is coarser than the rule: it declines the allele even when the `Ter` sits after (or is otherwise disjoint from) a run that lies entirely before the stop.

### Notes

- Output is always valid HGVS; this is a canonicalization completeness gap, not a correctness bug.
- Pre-existing: inherited from the substitution-only coalesce (#1095) and its order-independence (#1116); not introduced by #1117/#1121.

### Suggested direction

Scope the decline to the run(s) that would actually place a residue at or after a `Ter`, rather than the whole allele: a run entirely 5′ of the earliest `Ter` position can coalesce; a run that includes or extends past a `Ter` must still decline. Take care for a `Ter` that lands inside or immediately adjacent to a run.

nh13:
## Summary

`coalesce_protein_adjacent_changes` declined the **entire** cis protein allele as soon as **any** member was a to-`Ter` (nonsense) substitution. An adjacent run lying entirely 5′ of an unrelated downstream stop was therefore left as a bracket instead of collapsing to the spec-preferred `delins`.

| input | before | after |
|-------|--------|-------|
| `NP_003997.1:p.[Cys100Ser;Asp101Gly;Arg200Ter]` | unchanged | `NP_003997.1:p.[Cys100_Asp101delinsSerGly;Arg200Ter]` |
| `NP_003997.1:p.[Cys100Ser;Asp101Ter]` | unchanged | unchanged (run contains the stop) |
| `NP_003997.1:p.[Arg100Ter;Cys200Ser;Asp201Gly]` | unchanged | unchanged (run is downstream of the stop) |

## Why

The old guard was coarser than the rule it enforces. Nothing may be described at or after a translation stop (`protein/substitution.md:20`, `protein/delins.md:45`) — that constrains the *run* being merged, not every other run in the allele. Meanwhile `protein/substitution.md:23` / `protein/delins.md:18` require two or more consecutive changed amino acids to be described as a `delins`, so an untouched upstream run is a canonicalization the spec asks for.

## Change

Bound coalescing by the earliest to-`Ter` residue position instead of vetoing the allele:

- a run merges only when it ends **strictly before** that position;
- a run that **contains** the stop (its merged insert would read `…Ter…`) or that sits **wholly downstream** of it re-emits its members verbatim;
- with no to-`Ter` member anywhere, behavior is unchanged.

Because `members` is already sorted into ascending residue order (#1116), the first to-`Ter` member is the earliest, and the decline stays independent of the author's member order.

The `else` arm now re-emits every member of a declined run (not just the first), indexed through `source` so a declined multi-member run round-trips exactly as authored.

## Tests

New `tests/it/issue_1125_ter_scoped_coalesce.rs` (all permutations of each member set, each asserting an idempotent fixed point):

- the repro — a downstream `Ter` no longer blocks an upstream run;
- a run **containing** the `Ter` still declines;
- a run **downstream** of the `Ter` declines;
- with several `Ter` members, the **earliest** governs;
- a `Ter` strictly adjacent to a run joins that run and declines it;
- tightest coalescible boundary — one unchanged residue between run and stop;
- unrelated lone members (including a `del`) survive the partial coalesce.

Existing declines (`allele_grammar_corners`, `issue_1116`, `issue_1120`) are unaffected; the `issue_1116` doc comment is updated to say the decline is scoped to the run, not the allele. Full suite green (8302 passed), `clippy -D warnings` clean.

## Notes

Output was valid HGVS before and after — this closes a canonicalization completeness gap, not a correctness bug. Collapsing a nonsense change toward `p.<ref><pos>Ter` remains a separate rule.

Closes #1125

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -1234,6 +1234,23 @@ fn build_naedit<P>(
 /// allele fell through to the bracket form `protein/substitution.md:23` calls
 /// "not correct".
 ///
+/// A to-`Ter` (nonsense) member **bounds** coalescing rather than vetoing the
+/// whole allele (#1125): only a run lying **entirely 5′ of the earliest `Ter`**
+/// may merge. A run that contains that `Ter`, or that sits wholly downstream of
+/// it, is left as authored — but an unrelated upstream run still collapses, e.g.
+/// `p.[Cys100Ser;Asp101Gly;Arg200Ter]` → `p.[Cys100_Asp101delinsSerGly;Arg200Ter]`.
+///
+/// The `Ter`-containing run is declined **conservatively**, not because every
+/// such merge is spec-invalid. The spec forbids two specific shapes: residues
+/// listed *after* a stop (`protein/delins.md:45` — not `delinsSerSerTerAlaAsp`)
+/// and an *immediate* stop written as a delins (`protein/substitution.md:20` —
+/// not `p.Cys5_Ser6delinsTerGluAsp`, but `p.Tyr4Ter`). A run whose to-`Ter`
+/// member is **last** would merge to a *trailing*-`Ter` delins, which the spec
+/// endorses (`protein/delins.md:47`: `p.(Pro578_Lys579delinsLeuTer)`); declining
+/// it is a known completeness gap inherited from #1095, tracked separately.
+/// Collapsing a nonsense change toward `p.<ref><pos>Ter` is likewise a separate
+/// rule.
+///
 /// Returns `Some` only when at least one run of ≥2 adjacent members is merged;
 /// a fully-merged allele collapses to a bare [`HgvsVariant::Protein`], a
 /// partial merge to a smaller [`HgvsVariant::Allele`]. Returns `None` — leave
@@ -1242,8 +1259,7 @@ fn build_naedit<P>(
 ///   - any member is not a single-residue protein substitution or deletion
 ///     (other edit kinds — dup / ins / inv / fs / ext / multi-residue — are
 ///     out of scope for this narrow rule),
-///   - any substitution member changes the residue to `Ter` (a `delins…Ter…`
-///     run would list residues after the stop — spec-invalid),
+///   - every run is at or after a to-`Ter` member (see above),
 ///   - two members share a residue position (a `n,n` pair is the contradictory
 ///     same-residue case, not a delins), or
 ///   - a run would mix predicted `( )` and certain members (ambiguous).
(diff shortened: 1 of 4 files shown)


nh13:
Dual-review pass (CodeRabbit CLI lens + local agent lens + HGVS spec audit) applied.

- **CodeRabbit CLI lens:** 0 findings across all 4 changed files.
- **Agent lens:** no defects in the change itself. Bounds, ordering (`first_ter_pos` is read after the ascending sort, so it really is the earliest), the `changed`-flag/`None` contract, and the widened `else` arm re-emitting every member of a declined run all check out. `Option::is_none_or` is fine at the crate MSRV (1.87) and already used across `src/`.
- **Spec audit — one real divergence, fixed in-comment here and filed for follow-up.** The justification the code carried (inherited from #1095) overstated the spec: `protein/delins.md:45` forbids residues *after* a stop, and `protein/substitution.md:20` forbids writing an *immediate* stop as a delins — but a *trailing*-`Ter` delins is explicitly endorsed (`protein/delins.md:47`: `p.(Pro578_Lys579delinsLeuTer)`). So declining a run whose to-`Ter` member is **last** is conservative, not spec-mandated. This PR does not change that behavior; it now says so accurately instead of asserting spec-invalidity. Tracked as #1129.

Two more follow-ups filed from the same pass, both pre-existing and out of scope here:

- #1130 — the coalesce's three *other* declines (predicted/certain mix, two members at one residue, unsupported edit kind) still veto the whole allele, exactly the shape this PR fixes for `Ter`.
- #1131 — `try_protein_delins_canonicalize` gates on the provider-wide `has_protein_data()` but bails on a per-accession miss, so the #1119/#1126 reference-free minimizations never run for an accession a real provider does not carry.

Only comment/doc text changed since the first push (amended into the single commit); behavior, tests, and their expectations are unchanged.

PR #1128, opened 2026-07-22
fix(normalize): minimize a named-flank pure insertion reference-free

### LINKED ISSUE #1126: Reference-free protein delins minimization leaves a named-flank pure-insertion as a non-minimal delins
### Summary

The reference-free branch of `try_protein_delins_canonicalize` (`src/normalize/mod.rs`) minimizes a ≤2-residue `delins` by trimming its named endpoints, but when the affix-trim leaves an **empty deleted window** (a pure insertion) it bails and keeps the input `delins` — even when both flanking residues are named in the AST and the minimal insertion is derivable without a reference.

### Repro (reference-free, `MockProvider`)

| input | actual | expected (minimal) |
|-------|--------|--------------------|
| `NP_003997.1:p.Val559_Glu560delinsValSerGlu` | `NP_003997.1:p.Val559_Glu560delinsValSerGlu` (unchanged) | `NP_003997.1:p.Val559_Glu560insSer` |

Affix-trim: leading `Val` matches `Val559`, trailing `Glu` matches `Glu560`, leaving a residual insertion of `Ser` between the two named anchors — expressible as `p.Val559_Glu560insSer` without a reference.

### Cause / trade-off

The `residual_del.is_empty()` (pure-insertion) branch requires `has_protein_data()` because it routes through the ins→dup search, which needs the reference to test whether the insertion is actually a tandem duplication. Reference-free it bails rather than emit a plain `ins`. This is **defensible** — a plain `ins` could miss a `dup` canonicalization the reference would find — but it leaves a non-minimal `delins` on the reference-free path when both flanks are named.

### Notes

- Low impact: only the reference-free (no-provider) path is affected; with a protein-backed provider the case is minimized correctly.
- Output is valid HGVS; canonicalization completeness gap, not a correctness bug.
- The sibling case `p.Arg76_Cys77delinsArg` (redundant single-residue flank → `del`) was resolved by #1122's reference-free leading-trim.

### Suggested direction

When the trimmed window is a pure insertion between two **named** endpoints, emit `p.<start>_<end>ins<residual>` reference-free, accepting that the ins→dup refinement remains reference-gated (a later dup-minimization pass with a provider may reduce it further).

nh13:
## Summary

The reference-free branch of `try_protein_delins_canonicalize` minimizes a ≤2-residue `delins` by trimming it against the residues its own location names. When that trim consumed the whole deleted window the residual was a **pure insertion**, which the code routed through the ins→dup search — and that search needs a protein reference, so reference-free it bailed and kept the input `delins`.

| input | before | after |
|-------|--------|-------|
| `NP_003997.1:p.Val559_Glu560delinsValSerGlu` | unchanged | `NP_003997.1:p.Val559_Glu560insSer` |
| `NP_003997.1:p.Val559_Glu560delinsValSerTrpGlu` | unchanged | `NP_003997.1:p.Val559_Glu560insSerTrp` |
| `NP_003997.1:p.Val559_Glu560delinsSerValGlu` | unchanged | unchanged (left flank unnamed) |

For the repro the leading `Val` matches `Val559` and the trailing `Glu` matches `Glu560`, so both anchors of the residual `Ser` insertion are named in the AST — the minimal form is derivable with no reference at all.

## Change

In the pure-insertion branch, when the provider has no protein data, emit `p.<X>_<Y>ins<residual>` when the trim leaves a **named residue on both sides** of the insertion point.

`ref_aas[k]` is the residue at 1-based `start_pos.number + k`, so the flanks — `new_start - 1` and `new_start` — are `ref_aas[lcp - 1]` and `ref_aas[lcp]`. Both are in range exactly when `1 <= lcp < ref_aas.len()`:

- `lcp == 0` — the trim came entirely off the trailing end, putting the insertion point 5′ of the named window (left flank unknown);
- `lcp == ref_aas.len()` — the leading match consumed the whole window, putting the point 3′ of it (right flank unknown).

Either way the missing flank residue is knowable only from the reference, so #1119's conservative default holds and the input `delins` is kept.

The ins→dup refinement stays reference-gated: with a protein-backed provider the same insertion may reduce further to a `dup`, which only the reference can establish. Nothing on the reference-backed path changes.

## Tests

New `tests/it/issue_1126_reference_free_delins_to_ins.rs` (empty `MockProvider`, every case asserting an idempotent fixed point):

- the repro — both flanks named → `ins`;
- multi-residue residual insertion;
- the predicted `( )` wrapper survives the rewrite;
- unnamed **left** flank keeps the `delins`;
- unnamed **right** flank keeps the `delins`;
- a 1-residue `delins` can never leave a pure insertion between two named flanks — both trim directions keep the input.

The #1119 suite is unchanged and still green. Full suite green (8300 passed), `clippy -D warnings` clean.

## Notes

Output was valid HGVS before and after — this closes a canonicalization completeness gap on the no-provider path, not a correctness bug.

Closes #1126

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -4242,7 +4242,9 @@ impl<P: ReferenceProvider> Normalizer<P> {
     /// - residual single AA, del single AA → substitution
     /// - residual non-empty, del empty → route through
     ///   `try_protein_ins_to_dup` (anchored at the flanking
-    ///   positions of the trimmed range)
+    ///   positions of the trimmed range); reference-free, emit the
+    ///   plain `<X>_<Y>ins<residual>` when both flanking residues are
+    ///   named by the input's endpoints (#1126)
     /// - otherwise → smaller delins (caller falls back; the helper
     ///   returns `None` for the no-progress case)
     fn try_protein_delins_canonicalize(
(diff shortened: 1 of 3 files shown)


nh13:
Dual-review pass (CodeRabbit CLI lens + local agent lens + HGVS spec audit) complete — **no changes needed**; the commit is unchanged since the first push.

- **CodeRabbit CLI lens:** 0 findings across all 3 changed files.
- **Agent lens:** the new branch checks out. `new_start - 1` cannot underflow (`start_pos.number >= 1` is already enforced and `lcp >= 1`, so `new_start >= 2`); `ref_aas[lcp - 1]` / `ref_aas[lcp]` are in bounds exactly under the `1 <= lcp < ref_aas.len()` guard; `residual_seq` is provably non-empty at that point (the empty and both-empty cases return earlier), so no zero-length `ins` can be emitted; and the two emitted positions are consecutive, as an insertion requires.
- **Spec audit:** clean. `protein/delins.md:16` ("when **one** amino acid is replaced by **one** other … the change is a substitution") and the minimal-form/edit-priority rule both support reducing a fully-affix-trimmed `delins` to the plain `ins`; the 3'-rule and the ins→dup refinement genuinely need the reference and stay gated. The parser's residues-after-`Ter` rule (`protein/delins.md:45`) rejects the inputs that could otherwise produce a `Ter`-bearing residual, so that path is unreachable.

One pre-existing, out-of-scope gap surfaced and was filed as **#1131**: `try_protein_delins_canonicalize` picks its path on the *provider-wide* `has_protein_data()` flag but bails on a *per-accession* fetch miss, so both this PR's and #1119's reference-free minimizations do not run for an accession a real provider does not carry (reproducible through the CLI, whose default `MockProvider::with_test_data()` has protein data but not `NP_003997.1`). The integration tests here use an empty `MockProvider`, which is the correct harness for the reference-free path and is unaffected.

Issue #1158, opened 2026-07-24
EquivalenceChecker: reports NotEquivalent for variants with identical resulting sequence (complex indels)


msto:
> [!NOTE]
> Authorship: the content below was drafted by GitHub Copilot CLI (an AI coding
> agent) and filed via gh under @msto's account. It reflects the agent's
> analysis, not a statement authored by @msto.

## What's wrong

`EquivalenceChecker.check` returns **`NotEquivalent`** for two variants that
describe the **same edit** — identical resulting sequence against the same
reference — when the two inputs use different (but equivalent) HGVS encodings of a
complex indel:

```
TEMPLATE:g.8_14delinsGATTA   vs  TEMPLATE:g.[8A>G;9G>A;11C>T;13_14del]   -> NotEquivalent
TEMPLATE:g.10_20delinsTC     vs  TEMPLATE:g.12_20del                     -> NotEquivalent
```

Both pairs are verified to yield byte-identical resulting sequences (the edits are
applied to the reference in the reproduction below). The checker is not simply
string-comparing — it correctly reports `Identical` for an exact match and
`NormalizedMatch` for a substitution written two ways (`g.11C>T` vs
`g.11delinsT`). It fails specifically on multi-base / multi-part indel encodings.

This is a silent wrong answer from the API whose entire purpose is to decide
equivalence: `check` returns a well-formed `EquivalenceResult(level=NotEquivalent)`
with no error or warning. It also defeats the natural workaround for consumers who
cannot rely on normalized-string equality (see companion normalizer issue): a
caller that switches from `normalize(a) == normalize(b)` to
`EquivalenceChecker.check(a, b)` to match variants robustly still gets the wrong
answer for exactly the complex-indel cases where string comparison already fails.
Any dedup/join/matching built on `check` silently treats one variant as two.

## Severity: high

- **Silently wrong** — the equivalence API returns `NotEquivalent` for equivalent
  variants; no error, no warning, and the `EquivalenceResult` looks valid.
- **Defeats the intended fallback** — `EquivalenceChecker` is the correct tool for
  callers who know normalized strings differ; here it fails on the same inputs,
  so there is no working API-level equivalence path for complex indels.
- **Inconsistent with its own simpler cases** — `Identical` and `NormalizedMatch`
  are returned correctly for substitutions, so behavior depends on edit encoding
  rather than on the resulting sequence.
- **Corrupts matching/aggregation** — any pipeline that uses `check` to
[...]

PR #1160, opened 2026-07-24
fix: 3'-shift delins that reduces to a deletion or duplication (#1157)

### LINKED ISSUE #1157: Normalizer: normalization of complex indels is not encoding-invariant (non-confluent; non-idempotent in some cases)
> [!NOTE]

## What's wrong

`Normalizer.normalize` is **not encoding-invariant** for complex indels: two HGVS
descriptions of the **same** edit (identical resulting sequence) normalize to two
**different** strings. In some cases the normalized output is not even
**idempotent** — normalizing it again changes it (`norm(norm(x)) != norm(x)`).

Two shapes, same underlying defect (normalization depends on how the input was
partitioned rather than on the resulting sequence):

**A — non-confluence (both outputs are stable/idempotent, just different).** A
`delins` and its per-base decomposition describe the same variant, but stay in
distinct canonical forms:

```
TEMPLATE:g.8_14delinsGATTA              -> TEMPLATE:g.8_14delinsGATTA
TEMPLATE:g.[8A>G;9G>A;11C>T;13_14del]   -> TEMPLATE:g.[8_9delinsGA;11C>T;13_14del]
```

**B — non-idempotence (the normalized output is not a fixed point).** A `delins`
that reduces to a pure deletion is emitted **without** the 3′-shift that the same
deletion receives when supplied directly:

```
TEMPLATE:g.10_20delinsTC   -> TEMPLATE:g.12_20del      # not 3'-shifted
TEMPLATE:g.12_20del        -> TEMPLATE:g.13_21del      # 3'-shifted
# hence norm(norm("…10_20delinsTC")) = g.13_21del  !=  norm("…10_20delinsTC") = g.12_20del
```

Both are silent: `normalize` returns a well-formed string with no error or
warning. The consequence is that the "canonical" form is not canonical — the
descriptor a caller gets back depends on the input encoding, and (case B) is not
even self-consistent. Any consumer that treats the normalized HGVS.g string as a
stable key — counting, joining, or deduplicating variants — over-splits a single
variant into multiple keys, because the sequence-identical encodings never
compare equal. There is no error; the outputs are valid HGVS, just not
convergent, so the discrepancy is easy to miss until keys silently fail to match.

## Severity: high

- **Silently wrong** — no error/warning; every output is valid HGVS, just not a
  single canonical form.
- **Breaks the canonical-form guarantee** — the whole point of normalization is one
  descriptor per variant; here the descriptor depends on input partitioning.
- **Non-idempotent (case B)** — `norm(norm(x)) != norm(x)` violates a fundamental
  normalizer contract; a normalized value is expected to be a fixed point.
- **Ordinary inputs** — complex indels (delins, inversions, multi-part `[;]`
  alleles) are common input, not a rare edge case; simple substitutions are
  unaffected (they are confluent).
- **Corrupts key-based aggregation** — pipelines that count/join/dedup on the
  normalized string silently over-split one variant into several keys.

## Reproduction (verified on v0.10.1, commit 43ec9ab)

Harness: a single synthetic transcript on contig `TEMPLATE` (invented sequence, no
real accession, coordinates, or reference bases). Each pair below is verified to
produce the identical resulting sequence by applying the edits to the reference.

```python
import ferro_hgvs

# --- Fully synthetic, anonymized reference (no real locus) ---
      "protein_id": TID,

    norm = ferro_hgvs.Normalizer(reference_json=ref)

def n(v): return norm.normalize(v)
def mutant(ops):                       # apply 1-based edits right-to-left
    s = SEQ
    for st, en, ins in sorted(ops, reverse=True): s = s[:st-1] + ins + s[en:]
    return s

# Example A -- delins vs decomposed allele (same resulting sequence)
a1 = "TEMPLATE:g.8_14delinsGATTA"
a2 = "TEMPLATE:g.[8A>G;9G>A;11C>T;13_14del]"
assert mutant([(8,14,"GATTA")]) == mutant([(8,8,"G"),(9,9,"A"),(11,11,"T"),(13,14,"")])
print(n(a1))          # TEMPLATE:g.8_14delinsGATTA
print(n(a2))          # TEMPLATE:g.[8_9delinsGA;11C>T;13_14del]
print(n(a1) == n(a2)) # False

# Example B -- delins vs deletion (same resulting sequence); output NOT idempotent
b1 = "TEMPLATE:g.10_20delinsTC"
b2 = "TEMPLATE:g.12_20del"
assert mutant([(10,20,"TC")]) == mutant([(12,20,"")])
print(n(b1))          # TEMPLATE:g.12_20del
print(n(n(b1)))       # TEMPLATE:g.13_21del   <- norm(norm(x)) != norm(x)
print(n(b2))          # TEMPLATE:g.13_21del
```

Observed on **v0.10.1**:

```
[A] same resulting sequence: True
    norm(TEMPLATE:g.8_14delinsGATTA)            = TEMPLATE:g.8_14delinsGATTA            (idempotent: True)
    norm(TEMPLATE:g.[8A>G;9G>A;11C>T;13_14del]) = TEMPLATE:g.[8_9delinsGA;11C>T;13_14del] (idempotent: True)
    normalized strings equal: False

[B] same resulting sequence: True
    norm(TEMPLATE:g.10_20delinsTC)              = TEMPLATE:g.12_20del                   (idempotent: False)
    norm(TEMPLATE:g.12_20del)                   = TEMPLATE:g.13_21del                   (idempotent: True)
    normalized strings equal: False
    IDEMPOTENCE VIOLATION: norm(norm(g.10_20delinsTC)) = g.13_21del  !=  norm(...) = g.12_20del
```

A third pair, `TEMPLATE:g.19_29delinsTG` vs `TEMPLATE:g.21_29del`, exhibits the
identical case-B pattern (`g.21_29del` vs re-normalized `g.22_30del`) at a
different position; omitted as redundant.

Normalization appears to dispatch on the input's edit *type/partition* (delins vs
deletion vs bracketed multi-edit) and does not reduce every encoding to one
canonical partition of the resulting sequence. Case B further suggests the
`delins`→`del` reduction path does not re-enter 3′-shift normalization, so it
emits a non-shifted (non-canonical) deletion. Normalizing from the resulting
sequence (an SPDI-style canonicalization), or re-running the full normalization to
a fixed point after any reduction, would make output encoding-independent and
idempotent. (Inferred from observed behavior; source not inspected.)

## Expected

Two descriptions with the same resulting sequence normalize to the **same** string,
and normalization is idempotent (`norm(norm(x)) == norm(x)`):

```
TEMPLATE:g.8_14delinsGATTA            == TEMPLATE:g.[8A>G;9G>A;11C>T;13_14del]   (after normalize)
TEMPLATE:g.10_20delinsTC              -> a 3'-shifted, idempotent canonical form
```

nh13:
## Summary

`Normalizer::normalize` was **not idempotent** and **not encoding-invariant** for a `delins` that reduces to a pure deletion or a duplication, as reported in #1157 (case B).

When `canonicalize_delins` trims a `delins` down to a `DelinsCanonical::Deletion` or `::Duplication`, `normalize_na_edit` returned the reduced edit **directly**, bypassing the 3'-shift `shuffle()` (and repeat contraction) that every genuine `del`/`dup` input receives. The sibling `DelinsCanonical::Insertion` arm already recurses into `normalize_na_edit` for exactly this reason — the `Deletion` and `Duplication` arms did not.

Consequences (reproduced on synthetic homopolymer fixtures where the 3'-shift target is unambiguous):

- `g.257_259delinsT` (ref `TAA`, trims prefix `T` → deletion of `AA` in an A-tract) emitted the **unshifted** `g.258_259del`, whereas a direct `g.258_259del` 3'-shifts and contracts to `g.258_262A[3]`. So `norm(norm(x)) != norm(x)` and the delins spelling never compared equal to the direct-deletion spelling.
- The same defect existed for the `delins → dup` reduction: `g.258delinsGG` in a G-tract emitted the unshifted `g.258dup` while a direct `g.258dup` shifts to `g.262dup`.

## Fix

Make both reduction arms recurse through `normalize_na_edit` with the reduced edit (mirroring the `Insertion` arm), so the 3'-shift and repeat-collapse run, and collected warnings are preserved.

## Tests

New `tests/it/issue_1157_delins_reduction_shift.rs` asserts, for both the del- and dup-reduction paths:
- the reduced edit is 3'-shifted to its canonical form,
- `norm(norm(x)) == norm(x)` (idempotence),
- the delins spelling and the direct `del`/`dup` spelling of the same edit normalize identically (confluence).

All 6 fail before the fix and pass after; the full suite (8436 tests) is green, including the reference-free idempotency corpus and the pinned spec-normalization tests. `cargo fmt`/`clippy -D warnings` clean.

## Scope note (case A)

The issue also reports "case A": a single `delins` (`g.8_14delinsGATTA`) and a decomposed multi-part allele (`g.[8A>G;9G>A;11C>T;13_14del]`) — same resulting sequence — normalize to two different (but each idempotent) strings. That is **not** addressed here, and I don't believe it is a normalizer bug: ferro deliberately does not merge non-adjacent cis-allele members across unchanged reference bases into one spanning `delins` (that would be a coarser, arguably lossy description, and is not required by the spec). Declaring two encodings "the same variant" by *resulting sequence* is the job of `EquivalenceChecker`, which is fixed separately in the companion PR for #1158.

Closes #1157

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -6955,15 +6955,34 @@ impl<P: ReferenceProvider> Normalizer<P> {
                         DelinsCanonical::Deletion { start: s0, end: e0 } => {
                             // c.2_5delinsAT (ref ACGT) -> c.3_4del. Range fields
                             // are the trimmed half-open 0-indexed interval.
-                            return Ok((
-                                index_to_hgvs_pos(s0),
-                                e0 as u64,
-                                NaEdit::Deletion {
-                                    sequence: None,
-                                    length: None,
-                                },
-                                warnings.clone(),
-                            ));
+                            //
+                            // Recurse into `normalize_na_edit` with the reduced
+                            // deletion so the full del pipeline (3'-shift
+                            // `shuffle()` + repeat contraction) runs — exactly as
+                            // the `Insertion` arm below already does. Returning
+                            // the trimmed deletion directly skipped the shift,
+                            // making the output non-idempotent and non-confluent
+                            // with the same edit written as a plain `del`:
+                            // `g.257_259delinsT` emitted the unshifted
+                            // `g.258_259del`, while a direct `g.258_259del`
+                            // 3'-shifts (and contracts) to `g.258_262A[3]`, so
+                            // `norm(norm(x)) != norm(x)`. Issue #1157.
+                            let new_edit = NaEdit::Deletion {
+                                sequence: None,
+                                length: None,
+                            };
+                            let (new_start, new_end, new_edit, mut child_warnings) = self
+                                .normalize_na_edit(
+                                    ref_seq,
+                                    &new_edit,
+                                    index_to_hgvs_pos(s0),
+                                    e0 as u64,
+                                    boundaries,
+                                    is_coding,
+                                )?;
+                            let mut merged = warnings.clone();
+                            merged.append(&mut child_warnings);
+                            return Ok((new_start, new_end, new_edit, merged));
                         }
                         DelinsCanonical::Insertion {
                             after_index,
@@ -7044,16 +7063,31 @@ impl<P: ReferenceProvider> Normalizer<P> {
                             // c.5delinsGG where ref[5]=G  ->  c.5dup. Duplication
                             // is detected before trimming, so the range matches
                             // the input.
-                            return Ok((
-                                index_to_hgvs_pos(s0),
-                                e0 as u64,
-                                NaEdit::Duplication {
-                                    sequence: None,
-                                    length: None,
-                                    uncertain_extent: None,
-                                },
-                                warnings.clone(),
-                            ));
+                            //
+                            // Recurse so the reduced duplication takes the same
+                            // 3'-shift `shuffle()` a plain `dup` receives, mirror
+                            // of the Deletion/Insertion arms. Direct-return here
+                            // skipped the shift, so `g.258delinsGG` in a G-tract
+                            // emitted the unshifted `g.258dup` while a direct
+                            // `g.258dup` shifts to `g.262dup` — non-idempotent and
+                            // non-confluent. Issue #1157.
+                            let new_edit = NaEdit::Duplication {
+                                sequence: None,
+                                length: None,
+                                uncertain_extent: None,
+                            };
+                            let (new_start, new_end, new_edit, mut child_warnings) = self
+                                .normalize_na_edit(
+                                    ref_seq,
+                                    &new_edit,
+                                    index_to_hgvs_pos(s0),
+                                    e0 as u64,
+                                    boundaries,
+                                    is_coding,
+                                )?;
+                            let mut merged = warnings.clone();
+                            merged.append(&mut child_warnings);
+                            return Ok((new_start, new_end, new_edit, merged));
                         }
                         DelinsCanonical::KeepAsDelins {
                             start: s0,
--- a/tests/it/issue_1157_delins_reduction_shift.rs
+++ b/tests/it/issue_1157_delins_reduction_shift.rs
@@ -1,0 +1,162 @@
+//! Regression test for issue #1157: normalization of a `delins` that reduces
+//! to a pure deletion (or a duplication) was **not idempotent** and **not
+//! confluent** with the same edit written directly as a `del`/`dup`.
+//!
+//! Root cause: in `normalize_na_edit`, when `canonicalize_delins` trimmed a
+//! `delins` down to a `DelinsCanonical::Deletion` (or `::Duplication`), the
+//! arm returned the reduced edit **directly**, bypassing the 3'-shift
+//! `shuffle()` that every genuine `del`/`dup` input receives. The sibling
+//! `DelinsCanonical::Insertion` arm already recurses into `normalize_na_edit`
+//! for exactly this reason; the del/dup arms did not.
+//!
+//! Symptom (from the issue):
+//!   norm(g.10_20delinsTC) = g.12_20del        (unshifted — WRONG)
+//!   norm(g.12_20del)      = g.13_21del         (shifted — the true canonical)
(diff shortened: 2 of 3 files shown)


msto:
In response to the scope note:

> The issue also reports "case A": a single delins (g.8_14delinsGATTA) and a decomposed multi-part allele (g.[8A>G;9G>A;11C>T;13_14del]) — same resulting sequence — normalize to two different (but each idempotent) strings. That is not addressed here, and I don't believe it is a normalizer bug: ferro deliberately does not merge non-adjacent cis-allele members across unchanged reference bases into one spanning delins (that would be a coarser, arguably lossy description, and is not required by the spec). Declaring two encodings "the same variant" by resulting sequence is the job of EquivalenceChecker, which is fixed separately in the companion PR for [#1158](https://github.com/fulcrumgenomics/ferro-hgvs/issues/1158).

I agree that it is appropriate and correct for ferro to not merge non-adjacent cis-allele members across unchanged reference bases. The HGVS specification is clear on that point. However, that's not the request the original issue was making.

#1157 flagged an example where two equivalent variants were being normalized to different representations:

```
TEMPLATE:g.8_14delinsGATTA              -> TEMPLATE:g.8_14delinsGATTA
TEMPLATE:g.[8A>G;9G>A;11C>T;13_14del]   -> TEMPLATE:g.[8_9delinsGA;11C>T;13_14del]
```

In the reference-aware context, when ferro is able to recognize that these two variants are equivalent, ferro should then be able to normalize the former into the latter.

The scope note closed the question of "confluence" (or a single canonical form) under the argument that it would be inappropriate to normalize the latter into the former - I agree, but this overlooks that it is correct for both inputs to be normalized into the single canonical form [8_9delinsGA;11C>T;13_14del] .

I want to highlight this as it's related to the root cause surfaced in #1235 - given multiple equivalent inputs, the normalizer should emit the same canonical form for each.

PR #1178, opened 2026-07-25
perf(reference): stop building FASTA-index state that is immediately discarded

nh13:
Loading a prepared reference built a substantial amount of state that was then thrown away, and cloned a full path into every index entry. This removes that work. No new on-disk state, no new dependencies, no `unsafe`, and no change to which bases are served.

## What was being discarded

- `count_unique_files` built a `HashSet` over **every** index entry — hashing a full `PathBuf` per entry — and its only consumer was the file count in a log line. The scan loop already knows how many `.fai` it parsed.
- `from_directory` built a complete provider per directory, including the derived `base_to_versioned` and `has_genomic_data` maps, and `from_directories` kept only `.index`, discarded both, and recomputed them over the combined map. The per-directory maps were never read.
- `FastaIndexEntry` stored an owned `String` copy of its own map key. Its only two consumers were error-message formatting, so the name is now passed by callers, which already have it.
- Each entry also cloned the canonicalized absolute `PathBuf` of its FASTA — one allocation per entry for a small number of distinct files. Entries now carry a `u32` index into a per-provider file table, which also makes the open-file handle cache an integer-keyed map.

## Behavior

Unchanged, with one deliberate exception and three cosmetic ones.

**Deliberate:** directory entries are now sorted before indexing. `read_dir` order is unspecified, so which file won a *within-directory* duplicate accession was previously filesystem-dependent; it is now lexicographically-last-wins. This matters because the index is built by last-wins insertion, and that ordering is load-bearing across directories — the supplemental FASTA legitimately wins for a large number of accessions that also appear in the transcripts and refseqgene directories. Cross-directory precedence is unchanged. On a real prepared reference the only within-directory duplicates are between the two genome FASTAs, and those accessions extract byte-identically from both, so no served base changes.

**Cosmetic:**
- A manifest load previously emitted one `Loaded N sequences from M FASTA files` line per directory; it now emits one combined line. Nothing parses it.
- That line’s second number changes meaning from "distinct FASTA files present in the index" to "distinct physical FASTA files backing it". These differ only when a `.fai` parses to zero entries or when one file’s accessions are entirely overwritten by another’s.
- The protein-FASTA canonicalize-failure warning is worded more concisely. Same skip semantics.

Canonicalization still happens — once per file rather than once per entry — and ids are deduplicated by canonical path, so a FASTA reachable by two paths still maps to a single file-table slot and a single cached handle.

## Tests

Five added, each pinning something this change put at risk:

- a later directory wins a duplicate accession (the cross-directory precedence the whole index build depends on)
- the lexicographically-later file wins a *within-directory* duplicate (pins the sort; verified by mutation — removing the sort makes it fail deterministically)
- a directory that exists but contains no indexed FASTA is still an error rather than being silently skipped
- sequences resolve to the correct file when several are indexed from one directory
- a protein FASTA continues the shared file-id table through `from_manifest`, asserted on content — a wrong id there would read at a protein offset inside a transcript FASTA and return plausible-looking wrong bases

`cargo nextest run --features dev`: 8,529 passed. Clippy clean with `-D warnings` under both `--features dev` and `--features benchmark`.

## On measurement

This is a performance change and it was measured throughout, but the numbers are not quoted here: the machine available for timing was under heavy unrelated load, and its absolute figures swung by more than 50% between runs. The change is defensible without them — it deletes work whose results were provably unused, and removes an allocation per index entry. Anyone wanting figures should measure on a quiet machine.

### DIFF
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -73,10 +73,10 @@ use crate::reference::provider::{GenomicPlacement, ReferenceProvider};
 use crate::reference::transcript::Transcript;
 
 /// Index entry for a sequence in a FASTA file
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
 struct FastaIndexEntry {
-    /// Path to the FASTA file containing this sequence
-    file_path: PathBuf,
+    /// Index into [`MultiFastaProvider::files`] of the FASTA holding this sequence
+    file_id: u32,
     /// Length of the sequence
     length: u64,
     /// Byte offset to the start of sequence data
@@ -177,6 +177,17 @@ pub struct MultiFastaProvider {
     /// the ~400k+ entry indexes is a large slice of startup. The fixed seed
     /// also makes iteration deterministic across runs.
     index: FxHashMap<String, FastaIndexEntry>,
+    /// FASTA files backing the index, addressed by [`FastaIndexEntry::file_id`].
+    /// Interning the path here rather than cloning it into each of ~546k entries
+    /// removes an allocation per entry and makes the open-file cache an integer
+    /// keyed map.
+    ///
+    /// APPEND-ONLY: `file_id` is a positional index into this `Vec`, so an
+    /// entry must never be reordered or removed once assigned. A future
+    /// `sort()` or `retain()` here would silently repoint every existing
+    /// `FastaIndexEntry` at the wrong FASTA — a correctness bug with no
+    /// compiler signal and, absent a targeted test, no test signal either.
+    files: Vec<PathBuf>,
     /// Index mapping base accession (without version) to versioned accession
     base_to_versioned: FxHashMap<String, String>,
     /// Chromosome aliases (e.g., NC_000001 -> chr1)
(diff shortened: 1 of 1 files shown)


nh13:
Rebased onto current `main` (30 commits) and applied two findings from a dual-lens review. History is unchanged at four commits — both fixes were amended into their introducing commits.

**1. Guard parity in `from_manifest_inner` (amended into "build the FASTA index in one pass").** Of the seven `dirs.push` sites, five are guarded by `if !dirs.contains(...)` — `refseqgene_fastas`, `lrg_fastas`, `ensembl_transcript_fastas`, `supplemental_fasta`, backfill — while `transcript_fastas` and `genome_fasta` were not. A reference that keeps transcripts and genome in one directory therefore had that directory scanned twice, re-parsing every `.fai` in it: the same discarded-work class this PR removes elsewhere.

The fix is output-neutral, which is why it is here rather than in a follow-up. The two unguarded sites are positions 0 and 1, so any duplicate they produce is necessarily adjacent, and collapsing an adjacent duplicate cannot change last-wins precedence; a guarded site can never re-add an earlier directory.

**2. Test for the canonical-path dedupe (amended into "intern FASTA paths as u32 file ids").** `file_ids_by_canonical_path` exists solely so one physical file reached by two on-disk paths gets one `file_id` and therefore one cached handle — and the `files` field doc calls a mis-assignment "a correctness bug with no compiler signal and, absent a targeted test, no test signal either." Nothing exercised it: the six tests added here cover multi-file reads, cross- and within-directory precedence, the empty-directory error, protein `file_id` continuation, and `with_cdot`, but none reaches a FASTA by two paths.

`a_fasta_reached_by_two_paths_gets_one_file_id` (unix-gated) covers it. I mutation-checked it rather than just watching it pass — with the dedupe lookup forced to `None`, it fails `left: 2, right: 1`.

**On the within-directory tie-break.** `paths.sort()` makes it deterministic, but I wanted to know what that determinism actually selects, since `genome/` really does hold both `GRCh37.fna` and `GRCh38.fna` and `GRCh38` sorts last. Checking the two `.fai` files directly: the 65 shared accessions have **identical lengths in both builds** — the rCRS mito plus carried-forward `NT_` scaffolds — so which file wins is immaterial today. Worth knowing that the policy is emergent from filename collation rather than stated, but it is not a live risk.

**Validation** (all on `e2401a9`):

- `cargo nextest run --features dev` → 8766 passed, 0 failed, 145 skipped
- same with `FERRO_MANIFEST` against a real prepared reference → 8766 passed, 0 failed
- `cargo check --all-targets --features dev`, `cargo fmt --check`, `clippy --all-features --all-targets -D warnings` → all clean

The reference-backed run is the one that matters for this PR: it exercises the actual multi-directory loader being rewritten, across `genome/`, `transcripts/`, `supplemental/`, `lrg/`, `refseqgene/`, `ensembl/` and `backfill/`, including the ~80k-accession `supplemental`-over-`transcripts` precedence overlap the single-pass change had to preserve.

PR #1180, opened 2026-07-25
test(normalize): fuzz the n./r. axes and inv/con edits for idempotency

nh13:
Widens the normalization idempotency proptest from 3 coordinate systems to 11, and from 4 edit kinds to 6. Test-only — no production code changes.

## Why

The fuzz could only generate `g.` and `c.` (both strands, `cds_start == 1`) with `del`/`dup`/`ins`/`delins`. Everything else was unfuzzed, which is how a repeat-notation bug reached `main` earlier in this campaign: it lived on an axis the strategy could not produce.

## What changed

- **`n.`** on a non-coding `NR_` transcript.
- **`r.` twice**, because it is genuinely two different axes: on a coding transcript it is CDS-relative — the same axis as `c.` (#469) — and on a non-coding transcript it coincides with `n.`.
- **`c.` twice**: the existing `cds_start == 1` fixture, plus `CdsUtr3*` on a UTR-bearing transcript, where alone a 3'-shift can cross `cds_end` into `*N`.
- **`inv`** (widened to the ≥2 nucleotides the spec requires; a 1-base `inv` is a parse error) and **`con`** (donor always an explicit `S_E` pair, since a bare `conN` parses as an opaque cross-reference).

One deliberate fixture choice worth noting: the coding `r.`/`CdsUtr3` transcript carries a **real 256-base 5'UTR** (`cds_start == PAD_OFFSET + 1`). With `cds_start == 1` the CDS-relative and transcript-relative readings coincide, so the axis would be *vacuously* covered — present in the strategy, but incapable of distinguishing right from wrong.

## Non-vacuity guard

`case_strategy_covers_every_system_and_edit_kind` is the load-bearing addition. A passing property test proves nothing about a cell the strategy never emits, so this samples the strategy and fails if any (system × edit kind) cell is empty. It derives the expected system list from `ALL_SYSTEMS` and recovers the edit kind by parsing the *rendered* HGVS rather than trusting a self-reported tag, so adding a variant the generator does not emit fails here rather than passing silently.

It was written first and watched fail: on the unmodified generator, 12 cells were empty across 4000 samples. After the change all 66 are populated.

## What the fuzz found

This is the point of the PR, so it is worth being concrete. The widened strategy surfaced four idempotency defects. **All four are now fixed on `main`**, each by its own PR, and this branch is the last piece.

| found | axis / shape | fixed by |
|---|---|---|
| #1185 | `c.` `dup`/`delins`/`con` whose 3'-shift crosses `cds_end` into `*N` | #1189 |
| #1202 | `n.`/`r.` insertion saturating the transcript bounds | #1203 |
| #1207 | `r.` `dup` saturating a CDS boundary (a #1192 regression) | #1208 |
| #1209 | `c.` insertion resting at `cds_end`, clamped instead of shifting into the 3'UTR | #1211 |
| #1210 | `r.` repeat gate answered for the input span, not the emitted tract | #1212 |

Four failure seeds are committed to `tests/proptest-regressions/normalize_idempotency_proptest.txt` (#1202, both #1207 cases, and #1210). They replay through the *current* strategy, and this is the PR that widens that strategy to `n.`/`r.`, so they must travel with it. Now that the underlying bugs are fixed they act as regression guards — the convention the file's own header describes and which #1170 and #1172 each followed.

## Two things a reviewer should know

**1. `tests/it/cds_utr3_crossing_shift_idempotency.rs` is now a set of one-pass regressions.** It began as characterization tests for the *unfixed* #1185; with #1189 merged those cases are flipped to assert the correct one-pass answers and survive as deterministic pins over a hand-built transcript:

```
c.7_9delinsA -> c.*39_*40del    (fixed point)
c.7_10dup    -> c.7_*2AC[5]     (fixed point)
c.7_9con7_7  -> c.*39_*40del    (fixed point)
```

Its `FERRO_ASSERT_IDEMPOTENT` skip helper is gone — nothing in the file is non-idempotent any more, so every test runs under the oracle instead of opting out of it. The #1209 case is **not** duplicated here; #1211 ships a dedicated `issue_1209_cds_end_insertion_shift` module using the same fixture, and maintaining two copies of one case would only invite drift.

The padding artifact that produced a retracted finding earlier in this campaign was ruled out, not assumed: every repro is a hand-rolled transcript with no `SyntheticBuilder`. The decisive control throughout is that `r.` and `c.` share the byte-identical fixture yet the `r.` spelling settled in one pass — which is what localised each defect to the `c.` path rather than to the reference bases.

**2. `c.`-with-3'UTR is now IN the fuzz.** These two systems were held out while the axis was non-idempotent; both blockers (#1209, #1210) are fixed, so `CdsUtr3Plus`/`CdsUtr3Minus` join `ALL_SYSTEMS` here. The hold-out was never a weakened fixture — `RnaCoding*` always fuzzed that same 3'UTR-bearing transcript — but the `c.` spelling of that axis is a distinct code path, and it is now covered directly rather than by proxy.

## Also

- `common::synthetic::padded` made `pub` so a caller can build a transcript from the padded string with real UTRs.
- Corrected `core_strategy`'s flank-breaker comment: the `G`/`C` flanks bound a *homopolymer* but do **not** stop a 2-base rotation — a body ending `…A` gives suffix `…AC` and the pad resumes `AC`. That layout is what surfaced #1185.
- Case count 4k → 12k (the space grew from 12 cells to 66), and `max_local_rejects` lifted off its absolute 65,536 default. That default, against `core_strategy`'s pre-existing ~6.7% body-length reject rate, aborts any soak past ~900k cases — a first 1M run reported `FAIL` after 910,533 *successes*, which was a reject-cap abort and not an idempotency failure. Both still env-overridable.

## Verification

Rebased onto `main` at `f9ae3be` (#1212).

- Full suite: **8646 passed, 0 failed, 145 skipped**, and identical under `FERRO_ASSERT_IDEMPOTENT=1`.
- **1M-case soak clean** on all six properties (445s), including `normalization_is_idempotent_three_prime` — the property that failed at 150,830 cases before #1212, and the first 1M run to include the `CdsUtr3*` systems.
- `cargo clippy --all-features --all-targets -- -D warnings` clean; `cargo fmt --check` clean.
- With `FERRO_MANIFEST` set, `mutalyzer_normalize_tests::axis_errors` fails — verified to fail *identically* on the unmodified baseline (same two rows, `NM_000143.3:c.45del4` / `c.45dup4`). Pre-existing; it is what #1173 addresses.

## Note on where this coverage actually pays off

Worth stating plainly for whoever maintains this next: **PR CI's 12k-case run did not find any of the five defects above.** #1210 first reproduced at ~150k cases. The 12k setting is a smoke test; the soak is the detector, and it currently runs only by hand. Automating a nightly soak — and trimming what PR CI spends on random cases — is tracked separately from this PR.

### DIFF
--- a/tests/it/cds_utr3_crossing_shift_idempotency.rs
+++ b/tests/it/cds_utr3_crossing_shift_idempotency.rs
@@ -1,0 +1,288 @@
+//! Regression pins for `c.`-axis 3'-shifts that cross `cds_end` into the 3'UTR
+//! (`*N`) — a family that needed **two** normalization passes to settle and now
+//! settles in one.
+//!
+//! Found by the extended `normalize_idempotency_proptest` generator (the
+//! `n.`/`r.` axis + `inv`/`con` edit-kind extension). The `dup`/`delins`/`con`
+//! finding was filed as #1185 and fixed by PR #1189 (complete the shift in one
+//! pass) together with the #1192 codon-frame gate. Every case below is now
+//! idempotent on the first pass and is pinned here as a regression over a
+//! **hand-built** transcript, complementing the random synthetic fuzz that
+//! found them.
+//!
+//! The `ins` sibling — an insertion resting at `cds_end` that the #387 clamp
+//! rewrote instead of letting it shift on — was #1209, fixed by PR #1211. It is
+//! pinned in its own module, `issue_1209_cds_end_insertion_shift`, rather than
+//! duplicated here. With both fixed, the `c.`-with-3'UTR systems are fuzzed
+//! directly as `Sys::CdsUtr3Plus` / `Sys::CdsUtr3Minus` in
+//! `tests/it/normalize_idempotency_proptest.rs`, so this module is now a
+//! deterministic backstop rather than a stand-in for missing fuzz coverage.
+//!
+//! ## Why these cases were real bugs and not fixture artifacts
+//!
+//! [`rna_axis_shifts_across_cds_end_in_one_pass`] is the control. It runs the
+//! **same edit over the byte-identical transcript**, differing only in the axis
+//! it is spelled on (`r.` instead of `c.`). On a coding transcript `r.` *is*
+//! the CDS-relative axis — the same axis as `c.` (HGVS
+//! `background/numbering.md` L58/L61, issue #469, pinned by
+//! `issue_291_rna_axis_convention`) — so both spellings name the same bases and
+//! must shift identically. `r.` settled in one pass while `c.` did not, which
+//! localised the defect to the `c.` code path rather than to the reference
+//! bases or to 3'-shifting across `cds_end` in general. Both spellings now
+//! agree, and that agreement is what these tests assert.
+
+use ferro_hgvs::reference::transcript::{Exon, ManeStatus, Strand, Transcript};
+use ferro_hgvs::{parse_hgvs, HgvsVariant, MockProvider, Normalizer};
+
+/// A plain coding transcript with a 5'UTR, a short CDS, and a 3'UTR — built by
+/// hand rather than via `SyntheticBuilder` so no property of the synthetic
+/// padding can be blamed for the outcome.
+///
+/// | tx range   | bases                                      | region        |
+/// |------------|--------------------------------------------|---------------|
+/// | 1..=40     | `C` * 40                                   | 5'UTR         |
+/// | 41..=49    | `GGAAAAAAC`                                | CDS c.1..c.9  |
+/// | 50..=89    | `AC` * 20                                  | 3'UTR `*1..`  |
+///
+/// The CDS *ends* with `AC` (`c.8_9`) and the 3'UTR *begins* with `AC`, so a
+/// deletion of that `AC` has an `AC` tandem to 3'-shift along — a shift that
+/// necessarily crosses `cds_end` and must be rendered with `*N` positions.
+fn transcript_with_utr3_ac_tandem(strand: Strand) -> MockProvider {
+    let provider = coding_transcript("GGAAAAAAC", &"AC".repeat(20), strand);
+    // c.8_9 is the `AC` at the end of the CDS, and the 3'UTR continues it.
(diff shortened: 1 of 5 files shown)


nh13:
Rebased onto `main` @ 693fa37 (post #1170/#1172/#1175/#1176/#1174). The widened fuzz now fails on a **real, pre-existing `main` defect** it was designed to find:

```
NR_TEST.1:n.1_2insGC  (core CCCCAAAAAAAGGGG, NoncodingPlus, 5')
  -> NR_TEST.1:n.1insCG   <- invalid; ferro's own parser rejects it
```

This is the `n.`-axis analogue of the `c.1insAA` defect #1170 fixed: the boundary clamp is gated on `AxisRegion::Cds` and keyed on `cds_start`, so a non-coding transcript never gets the rewrite. `src/` is untouched by this PR, so the bug is on `main`; this PR only exposes it.

Filed as #1202. **This PR is blocked on that fix** — it cannot go green until the clamp covers the transcript start on the `n.`/`r.` axes.

Everything else here is green: 8575/8576 pass, and the new `case_strategy_covers_every_system_and_edit_kind` guard reports all 54 (system x edit-kind) cells populated across 9 coordinate systems.

Two review notes to fold in when this unblocks:
- `cds_utr3_crossing_shift_idempotency.rs` skips under `idempotency_oracle_active` and uses `assert_ne!` to record the known `cds_end`-crossing defect as expected. That is a "pin the gap" choice, but per the repo's own `tests/**` review policy those assertions should flip to the fixed behaviour when #1189 lands.
- The same file's `normalize_twice` case should assert the canonical `NC_UTR3.1:g.88_89del` directly, not only `once == twice`.

Note on rebasing: the merged flank widening (#1175, `G{body}C` -> `CCCC{body}GGGG`) conflicted with this PR's doc caveat arguing the narrow flanks were useful. Resolved in favour of the wider flanks, with the provenance preserved as a note — verified the coverage histogram still fills every cell and the `cds_end` case remains pinned by its own explicit test.

PR #1186, opened 2026-07-25
refactor(reference): construct the FASTA index through a single PreparedIndex type

nh13:
**Stacked on #1178** — review that first; this PR's diff is only its own two commits. Retarget to `main` once #1178 merges.

Pure refactor. No behavior change, no measurable startup change, no new dependencies, no on-disk artifact.

## Why

FASTA-index construction lived inline in `MultiFastaProvider`, spread across two constructors plus a manifest loader. The planned on-disk index cache needs somewhere to deserialize *into*. If it deserializes straight onto the provider, it becomes a second producer of the same state, free to disagree with the live one — and a cache that silently serves different data than a fresh load is the failure mode worth the most effort to design out.

So construction moves into one type with one place that derives its dependent state.

## What

**`src/reference/prepared_index.rs`** (new) owns the FASTA index and everything derived from it: the file table, the accession index, `base_to_versioned`, `has_genomic_data`, and the protein index. `MultiFastaProvider`'s five index-related fields collapse into one `prepared: PreparedIndex`.

Two constructors: `from_dirs` (scans directories) and `from_index` (takes an already-built index, used by `with_cdot`, which names a single FASTA rather than scanning). Both route through the same body, so `base_to_versioned` and `has_genomic_data` are computed in exactly one place.

**The invariant is enforced by the type, not by convention.** All five fields are private and there is no `Default`, so no code — including a future `#[derive(Deserialize)]` — can construct a `PreparedIndex` whose derived maps disagree with its index. The module documents the consequence for the cache work: deserialize into a separate DTO carrying only the index and file table, then call `from_index`, so the derived maps are recomputed rather than trusted. That sentence is the point of this PR.

The provider reaches the state through accessors (`entry`, `contains`, `versioned_for`, `protein_entry`, `path_for`, `len`, `file_count`, …) rather than fields.

## Behavior parity

Verified against #1178's head rather than argued from the code:

- The three protein-FASTA warning strings are byte-identical, including that `IO error: ` appears on the `.fai`-parse path and not on the canonicalize or overflow paths. Keeping that true is why `add_protein_fasta` returns `String` rather than `FerroError` — its only consumer is an `eprintln!` whose format predates this refactor, and wrapping the messages in `FerroError::Io` prepended its `Display` prefix to two of the three.
- Warn-and-continue for protein FASTAs is preserved: one bad protein FASTA still warns and the reference still loads. There is now a manifest-level test for this, which there was not before — the bad entry is listed first and the test asserts the following one still resolves, so a regression to `?` fails rather than passing vacuously.
- `Loaded N protein sequences` still reports the deduplicated total.
- The derived maps are computed once per construction, as before.
- No public API changed shape; all five removed fields were private.

## Tests

8,534 passing. New: the two `PreparedIndex` builder tests, the `add_protein_fasta` dedupe/no-orphan-slot test, and the manifest-level warn-and-continue test above. Clippy clean with `-D warnings` under `--features dev` and `--features benchmark`.

## Known follow-up, deliberately not here

`with_cdot` still calls `load_fai_index` directly, which is why that function is `pub(super)` rather than private. A `PreparedIndex::from_fasta(fasta, fai)` entry point would let it route fully through the type. That is new design rather than a faithful move, so it does not belong in a commit claiming zero behavior change.

### DIFF
--- a/src/reference/mod.rs
+++ b/src/reference/mod.rs
@@ -13,6 +13,7 @@ pub mod mock;
 pub mod multi_fasta;
 pub mod ng_hosted_transcripts;
 pub mod ng_placement_builder;
+pub(crate) mod prepared_index;
 pub mod protein;
 pub mod provider;
 pub mod transcript;
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -20,7 +20,7 @@
 //! For type-safe coordinate handling, see [`crate::coords`].
 
 use std::fs::File;
-use std::io::{BufRead, BufReader};
+use std::io::BufReader;
 use std::num::NonZeroUsize;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
@@ -69,23 +69,12 @@ fn read_exact_at(file: &File, buf: &mut [u8], offset: u64) -> std::io::Result<()
 use crate::data::cdot::CdotMapper;
 use crate::error::FerroError;
 use crate::reference::authoritative::CanonicalOverrides;
+use crate::reference::prepared_index::{
+    build_base_to_versioned, index_has_genomic_data, load_fai_index, load_index_from_dirs,
+    FastaIndexEntry,
+};
 use crate::reference::provider::{GenomicPlacement, ReferenceProvider};
 use crate::reference::transcript::Transcript;
-
-/// Index entry for a sequence in a FASTA file
-#[derive(Debug, Clone, Copy)]
-struct FastaIndexEntry {
-    /// Index into [`MultiFastaProvider::files`] of the FASTA holding this sequence
-    file_id: u32,
-    /// Length of the sequence
-    length: u64,
-    /// Byte offset to the start of sequence data
-    offset: u64,
-    /// Number of bases per line
-    line_bases: u64,
-    /// Number of bytes per line (including newline)
-    line_bytes: u64,
-}
 
 /// Supplemental transcript info for a single transcript.
 #[derive(Debug, Clone)]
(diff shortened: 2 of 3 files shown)


nh13:
Heads-up on CI: this PR targets `perf/prepared-index-deletions` (the #1178 branch), and `ci.yml` is scoped to `pull_request: branches: [main]`, so **none of the usual checks run here** — only CodeRabbit reported. That is a property of the stacked base, not of this change.

Ran the equivalent locally against this exact head (`8ac1aca`):

- `cargo nextest run --features dev` → 8,534 passed, 145 skipped
- `cargo clippy --features benchmark --all-targets -- -D warnings` → clean
- `cargo fmt --all -- --check` → clean

Retargeting this PR to `main` once #1178 merges will trigger the real CI run, so the gap closes on its own at that point — but until then the green checkmark count here is not meaningful.

nh13:
Rebased onto current `main` (`62051a8`, which now contains #1178) and reviewed through both lenses. Two fixes, each amended into its introducing commit; history unchanged at two commits.

## The important one: this PR and #1171 were a latent red build

#1186 deletes `use std::io::{BufRead, BufReader}` because it moves `load_fai_index` — the `BufRead` consumer — into the new module. Meanwhile #1171 ("perf(io): buffer serde_json reads from files", merged as `12085f1`) added a `BufReader::new(file)` in `from_manifest` at `multi_fasta.rs:512`, which stays behind in this module.

Each PR is green on its own base. Together they do not compile:

```
error[E0433]: cannot find type `BufReader` in this scope
   --> src/reference/multi_fasta.rs:512:37
```

No CI run could have caught this — #1186's ran on a base without #1171, #1171's on a base without #1186 — and neither review lens sees more than one diff at a time. Restored `use std::io::BufReader` (`BufRead` is genuinely unused here now; the only remaining `.lines()` call is on a `&str`). Amended into the module-move commit, since that is what removed the import.

## Second fix

The symlink-dedupe test added on #1178 asserted `provider.files.len()`; `files` is now private inside `PreparedIndex`, so it reads `provider.prepared.file_count()`. Amended into the commit that introduced the `prepared` field.

## Move fidelity

Rather than eyeball a 700-line move, I extracted `load_fai_index` and `load_index_from_dirs` from both trees and diffed them with comments stripped — **both are code-identical**. The only comment change correctly re-points a cross-reference from "the protein-FASTA loader below" to `PreparedIndex::add_protein_fasta`.

## Two CLI-lens findings, both rejected

- **"max 4294967296 should be 4294967295"** — false positive. `u32::try_from(len)` succeeds while `len <= u32::MAX`, so the final successful push takes id `4294967295` and leaves `len == 4294967296`; the error fires only when that many files are already held. The message states the true maximum file count (2^32 ids, 0-indexed).
- **Non-UTF-8 `.fai` sidecar path via `path.display()`** — real, but pre-existing (it predates #1178) and present at two sites including `with_cdot`. Fixing one of two inside a verbatim-move refactor would spoil the move's fidelity for no completeness gain. Left for a follow-up.

## Validation (all on `cc38b85`)

- `cargo nextest run --features dev` → 8771 passed, 0 failed, 145 skipped
- `FERRO_ASSERT_IDEMPOTENT=1` → 8771 passed, 0 failed
- `FERRO_MANIFEST=<real prepared reference>` → 8771 passed, 0 failed
- `cargo check --all-targets --features dev`, `cargo fmt --check`, `clippy --all-features --all-targets -D warnings` → clean

The reference-backed run matters most here: it exercises the relocated loader against the real multi-directory reference, including the ~80k-accession `supplemental`-over-`transcripts` precedence overlap.

PR #1189, opened 2026-07-25
fix(normalize): complete the 3'-shift across cds_end in one pass for delins and dup

### LINKED ISSUE #1185: fix(normalize): c.-axis 3'-shift crossing cds_end into the 3'UTR needs two passes
On the `c.` axis, a `dup`, `delins`, or `con` whose 3'-shift crosses `cds_end` into the `*N` 3'UTR stops short on the first normalization pass. Normalization is therefore not idempotent for these inputs.

Found by the widened idempotency fuzz in #1180, and pinned there as characterization tests.

## Reproduce

Transcript `NM_UTR3.1`: 5'UTR `C`×40, CDS `GGAAAAAAC`, 3'UTR `AC`×20.

```
NM_UTR3.1:c.7_9delinsA  ->  c.8_9del        ->  c.*39_*40del
NM_UTR3.1:c.7_10dup     ->  c.*2_*3insACAC  ->  c.7_*2AC[5]
```

The `dup` case diverges in **edit kind**, not just position: `dup` → `ins` → repeat.

The clearest statement of the defect is that pass 1's own output is not a fixed point:

```
NM_UTR3.1:c.8_9del  ->  c.*39_*40del  ->  c.*39_*40del   (stable from there)
```

So pass 1 emitted something genuinely unnormalized rather than choosing a different-but-stable spelling.

## Not a test-harness artifact

The repro uses a hand-rolled transcript — no `SyntheticBuilder`, no `ACGT` padding — because a padding artifact of exactly this shape produced a retracted finding earlier in this campaign. A 51,755-probe sweep placed all 106 failures in `c.`-with-3'UTR only.

Contrast controls, all one-pass:

- `r.7_9delinsa` on the **byte-identical fixture** crosses `cds_end` in one pass — this is the decisive one.
- `g.` over the same bases.
- A plain `c.` `del` crossing `cds_end`.
- The same `delins` reduction staying inside the CDS.

The repo's own `FERRO_ASSERT_IDEMPOTENT` oracle independently flags exactly these inputs by a different mechanism.

## Lead

The #387 CDS-**end** clamp in `src/normalize/mod.rs` — the 3' mirror of the #383/#418 CDS-start clamp that #1167 stabilized for the 5' side. Its catch-all arm asserts that *"other edit types do not currently produce 3'UTR-resident rewrites from CDS-interior inputs in this canonicalisation path"*. The `dup` repro falsifies that claim directly.

This is a lead, not a diagnosis — the root cause has not been confirmed.

## On fixing

`tests/it/cds_utr3_crossing_shift_idempotency.rs` (in #1180) pins the current buggy behavior as 5 characterization tests plus 4 contrast controls. They skip under `FERRO_ASSERT_IDEMPOTENT=1`, since the oracle panics inside `normalize()` before the test can compare passes.

When this is fixed:

1. Both passes must yield `c.*39_*40del`, and that whole file should be deleted.
2. Re-add `c.`-with-3'UTR to the fuzz's `ALL_SYSTEMS` in `tests/it/normalize_idempotency_proptest.rs` — it is currently excluded so the suite lands green, with a re-add pointer at the exclusion site.

nh13:
Closes #1185

Two commits, one per half of the issue. Both halves turned out to be the **same defect at two sites**: a decision keyed on the *input's* spelling/span rather than on the *result* it actually governs.

## Commit 1 — a reducing `delins` (and `con`)

\#918 relaxed the CDS↔3'UTR shuffle bound so a CDS-resident `del`/`dup` can complete its 3'-shift into the 3'UTR — the coding DNA reference is one contiguous string and the 3' rule is stated "for all descriptions". It was gated on the input's edit-kind spelling:

```rust
matches!(edit, NaEdit::Deletion { .. } | NaEdit::Duplication { .. })
```

A `delins` whose ref and alt share a prefix reduces to a pure deletion during normalization — `c.7_9delinsA` over ref `AAC` **is** a deletion of `AC` at `c.8_9` — which is exactly the shape the relaxation exists for, but it never received it:

```
c.7_9delinsA  ->  c.8_9del  ->  c.*39_*40del
```

`con` inherits both defect and fix, since the SVD-WG009 `con` → `delins` rewrite recurses before the shuffle runs.

## Commit 2 — a `dup` whose tract leaves the CDS

```
c.7_10dup  ->  c.*2_*3insACAC  ->  c.7_*2AC[5]
```

`duplication_to_repeat` already decides about the **shuffled** tract (it is called with `result.start`/`result.end`), but the codon-frame gate it consults was `is_coding`, precomputed from the **input** span. `c.7_10dup`'s input span is CDS-resident, so the gate fired and routed the dup to an `ins` literal; only on pass 2 — when the straddling tract made the input span itself non-coding — was repeat notation allowed.

The spec exempts the shifted tract (`DNA/repeated.md`):

> using a coding DNA reference sequence ("c." description), a repeated sequence variant description can be used only for repeat units with a length which is a multiple of 3 … **This restriction only applies to the coding sequence, which does not include the introns or the UTR sequence.** As such, `NM_024312.4:c.-6_-3G[6]` is valid as the reading frame is not affected.

The added copies land past `cds_end`, so they cannot move the reading frame. `c.7_*2AC[5]` is correct in one pass even though the `AC` unit is 2 nt. (The spec gives several more `c.`-axis examples with non-triplet units that are valid because they sit in introns or UTRs: `c.1210-33_1210-6GT[11]T[6]`, `c.1210-12_1210-6T[7]`, `c.-136-75952_-136-75878ATTTT[15]`.)

## Scope — deliberately narrow, and here is why

The fix passes the CDS bounds alongside `is_coding` so that **one** site can ask the gate question about the tract it actually decides on. Every other gate site keeps reading `is_coding` unchanged.

That is not tidiness — I first tried the "clean" version, replacing `is_coding` with the bounds everywhere and recomputing per site. It **broke idempotency** for a homopolymer insertion:

```
c.9_10insCC  ->  c.10delinsCCC  ->  c.4_10C[9]
```

caught by `normalize_idempotency_proptest`. The reason: recursive `normalize_na_edit` calls forward the *outer* verdict by design, and recomputing from each inner span silently changes it. So the additive form is the correct one, not a shortcut.

`deletion_to_repeat` has the same latent input-vs-shuffled inconsistency and is **left alone**, with a comment saying so. Re-keying it for symmetry is exactly the change that caused the above; it needs its own investigation.

## Controls

Every control **passed before** these fixes, which is what localises the defects rather than describing them:

- plain `c.8_9del` crosses `cds_end` in one pass — the shuffle machinery was never at fault
- the `n.` axis over the same bases
- a `delins` whose shift stays inside the CDS
- **a non-triplet unit whose tract stays entirely inside the CDS still renders as the spec's `ins` literal** — proves the gate was narrowed, not disabled
- **a codon-aligned unit inside the CDS still gets repeat notation** — proves it still discriminates on unit length

The fixture is hand-rolled rather than built with `SyntheticBuilder`, because a padding artifact of exactly this shape produced a retracted finding earlier in this campaign.

## One control I had to disqualify

The `r.` spelling of the same edit shifts in one pass, and that is what originally made this look like a plain bug. It is a valid control for the **shuffle**, but **not** for the codon-frame gate: `normalize_rna` passes no CDS span at all, so the gate never applies on the `r.` axis. Treating `r.` as an oracle would have "confirmed" whatever `r.` happened to do. Its doc comment now says this explicitly.

Which surfaced a **third, separate defect**: the `r.` axis never wires the codon-frame gate, even though `RNA/repeated.md:24-27` imposes the same multiple-of-3 restriction with the same UTR carve-out. Not addressed here — it changes `r.` repeat output and deserves its own issue and blast-radius check.

## Known wart

`normalize_na_edit` was already at clippy's 7-argument limit, so the added parameter needs `#[allow(clippy::too_many_arguments)]`. The right cleanup is folding `is_coding` + `cds_span` into one `CodonGate` value — they are two views of one question — but that is a mechanical rename across every call site in the function and does not belong in a behaviour fix. Flagged in a comment at the suppression.

## Verification

- Full suite with the real manifest and `FERRO_ASSERT_IDEMPOTENT=1`: **8531 passed, 1 failed**.
- The one failure is `mutalyzer_normalize_tests::axis_errors`, which fails **identically on clean `main`** (`be47233`) with the manifest set — same two rows, `NM_000143.3:c.45del4` / `c.45dup4`. Pre-existing; it is what #1173 addresses.
- `cargo clippy --features dev --all-targets -- -D warnings` clean; `cargo fmt --all` clean.
- RED was watched for both halves before either fix, including re-running the Site 2 test with only the Site 1 commit applied.

## Interaction with #1180

\#1180 added `tests/it/cds_utr3_crossing_shift_idempotency.rs`, which pins the **pre-fix** two-pass behaviour as characterization tests. When both land, that file must be deleted and `c.`-with-3'UTR re-added to `ALL_SYSTEMS` in `normalize_idempotency_proptest.rs` (it is excluded there with a re-add pointer). Whichever merges second should do it; this PR does not touch that file, since it is not on `main`.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -2635,6 +2635,7 @@ impl<P: ReferenceProvider> Normalizer<P> {
             rel_end,
             &Boundaries::new(0, ref_seq.len() as u64),
             false, // genomic context: codon-frame gate does not apply
+            None,  // codon-frame gate does not apply here
         )?;
 
         // Adjust back to genomic coordinates
@@ -3283,8 +3284,19 @@ impl<P: ReferenceProvider> Normalizer<P> {
             Some(cds_end) => tx_start >= cds_start && tx_end <= cds_end,
             None => false,
         };
-        let (mut new_tx_start, mut new_tx_end, mut new_edit, mut warnings) =
-            self.normalize_na_edit(seq, edit, tx_start, tx_end, &boundaries, is_coding)?;
+        // The same bounds, un-collapsed, so a gate site deciding about a span
+        // other than the input can evaluate it there instead (#1185).
+        let cds_span = transcript.cds_end.map(|cds_end| (cds_start, cds_end));
+        let (mut new_tx_start, mut new_tx_end, mut new_edit, mut warnings) = self
+            .normalize_na_edit(
+                seq,
+                edit,
+                tx_start,
+                tx_end,
+                &boundaries,
+                is_coding,
+                cds_span,
+            )?;
 
         // Substitutions are validated-then-returned-unchanged by
         // `normalize_na_edit`'s `Substitution` arm; nothing downstream (axis
@@ -3396,8 +3408,10 @@ impl<P: ReferenceProvider> Normalizer<P> {
                 ShuffleDirection::ThreePrime => cheap_right,
             };
             let (left_clamp_fired, right_clamp_fired) = if direction_could_clamp {
-                let (exon_only_start, exon_only_end, _exon_only_edit, _exon_only_warnings) =
-                    self.normalize_na_edit(seq, edit, tx_start, tx_end, &exon_only, is_coding)?;
+                let (exon_only_start, exon_only_end, _exon_only_edit, _exon_only_warnings) = self
+                    .normalize_na_edit(
+                    seq, edit, tx_start, tx_end, &exon_only, is_coding, cds_span,
+                )?;
                 let exon_only_start_0 = exon_only_start.saturating_sub(1);
                 (
                     cheap_left && exon_only_start_0 < boundaries.left,
@@ -3897,8 +3911,12 @@ impl<P: ReferenceProvider> Normalizer<P> {
             Some(s) => s.as_bytes(),
             None => return Ok((HV::Tx(self.canonicalize_tx_variant(variant)), vec![])),
         };
+        // `mut` bindings are #1202's (its transcript-bound clamp rewrites these
+        // in place); the trailing `None` is #1185's `cds_span`. `n.` has no
+        // reading frame, so there is no CDS span to re-check a shuffled tract
+        // against — unlike `r.` below, which does pass real bounds.
         let (mut new_start, mut new_end, mut new_edit, mut warnings) =
-            self.normalize_na_edit(seq, edit, tx_start, tx_end, &boundaries, false)?;
+            self.normalize_na_edit(seq, edit, tx_start, tx_end, &boundaries, false, None)?;
 
         // See the identical guard + comment in `normalize_cds` (#1052 follow-up):
         // substitutions are validated-then-returned-unchanged by
@@ -5382,8 +5400,23 @@ impl<P: ReferenceProvider> Normalizer<P> {
             Some((cds_start, cds_end)) => tx_start >= cds_start && tx_end <= cds_end,
             None => false,
         };
-        let (mut new_tx_start, mut new_tx_end, mut new_edit, mut warnings) =
-            self.normalize_na_edit(seq, edit, tx_start, tx_end, &boundaries, is_coding)?;
+        // `mut` bindings are #1202's (its transcript-bound clamp rewrites these
+        // in place). `cds_info` (not `None`) is the #1185 companion to #1192's
+        // `is_coding`: the codon-frame gate applies on `r.` exactly as on `c.`,
+        // so the shuffled-tract re-check must be able to run here too. Passing
+        // `None` would make that verdict unconditionally "exempt" and silently
+        // re-open the divergence #1192 closed — `issue_1192`'s
+        // `rna_homopolymer_expansion_inside_cds_is_gated` fails if you try it.
+        let (mut new_tx_start, mut new_tx_end, mut new_edit, mut warnings) = self
+            .normalize_na_edit(
+                seq,
+                edit,
+                tx_start,
+                tx_end,
+                &boundaries,
+                is_coding,
+                cds_info,
+            )?;
 
         // See the identical guard + comment in `normalize_cds` (#1052 follow-up):
         // substitutions are validated-then-returned-unchanged by
@@ -5933,6 +5966,7 @@ impl<P: ReferenceProvider> Normalizer<P> {
             rel_end,
             &Boundaries::new(0, ref_seq.len() as u64),
             false,
+            None, // codon-frame gate does not apply here
         )?;
 
         let new_start = new_rel_start + window_start;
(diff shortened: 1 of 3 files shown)


nh13:
## Merge hazard: #1189 and #1194 silently cancel each other at one call site

These two PRs touch the same line — `normalize_rna`'s `normalize_na_edit` call — and a naive conflict resolution **reverts #1192's fix without failing to compile**. Verified empirically, not inferred.

#1189 adds `cds_span` and derives the `duplication_to_repeat` gate **purely** from it, not from `is_coding`:

```rust
matches!(cds_span, Some((cds_s, cds_e)) if /* shuffled tract inside CDS */)
```

so any caller passing `cds_span: None` gets "not gated" regardless of what `is_coding` says. #1189 passes `None` on the `r.` path; #1194 wires `is_coding` there. #1192's primary repro is a `dup`, which goes through exactly that call.

**Measured** by applying #1194's change onto #1189's branch and running #1194's tests:

| `normalize_rna` passes | `r.8_9dup` (CDS-resident homopolymer) | verdict |
|---|---|---|
| `false, None` (#1189 today) | `r.4_9a[8]` | pre-fix, spec-forbidden |
| `is_coding, None` (naive merge) | `r.4_9a[8]` | **still pre-fix — the silent revert** |
| `is_coding, cds_info` (correct) | `r.9_10insaa` | 7/7 of #1194's tests pass |

**Resolution when the second of these two merges:** pass `cds_info` as `cds_span`, not `None`.

```rust
let is_coding = match cds_info {
    Some((cds_start, cds_end)) => tx_start >= cds_start && tx_end <= cds_end,
    None => false,
};
let (new_tx_start, new_tx_end, new_edit, mut warnings) =
    self.normalize_na_edit(seq, edit, tx_start, tx_end, &boundaries, is_coding, cds_info)?;
```

`cds_info` is already in scope in `normalize_rna` and is exactly the `cds_start.zip(cds_end)` pair `normalize_cds` builds. Passing it also gives the `r.` axis #1189's shuffled-tract treatment, which is the point of both PRs: the two axes should reach the same verdict on the same bases.

The full suite in that combined state is green apart from the two files #1194 already changes — `issue_736_rna_uracil_normalize` (fails with exactly the `r.5_6insuu` re-bless #1194 makes) and the regenerated enumeration fixture (all 8 `spec_enumeration_tests` pass once regenerated).

Related, already tracked: #1180 pins the pre-#1189 two-pass behavior in `tests/it/cds_utr3_crossing_shift_idempotency.rs`, which must be deleted — and `c.`-with-3'UTR re-added to `ALL_SYSTEMS` — when #1180 and #1189 both land.

Issue #1183, opened 2026-07-25
fix(normalize): the r. axis skips ins[...] cross-reference expansion, laundering unvalidated payloads onto c.


nh13:
The `r.` axis never runs the #333 `ins[...]` cross-reference expansion, so an `r.` input carries its payload through unexpanded and unvalidated. Projected onto `c.`, that produces a `c.` description ferro then refuses to read back.

## Reproduce

With a **fully resolvable** payload, `c.` and `n.` expand to a literal sequence and `r.` does not:

```
c.124_500delins[NM_003002.4:c.10_20]  ->  NM_004006.3:c.124_498delinsCTCTGGAGG
n.124_500delins[NM_003002.4:c.10_20]  ->  NM_004006.3:n.124_500delinsCTCTGGAGGCT
r.124_500delins[NM_003002.4:c.10_20]  ->  NM_004006.3:r.124_500delins[NM_003002.4:c.10_20]   # unexpanded
```

With an **unresolvable** payload the `c.` axis correctly refuses, but the `r.` axis launders it across:

```
$ ferro project 'NM_004006.3:r.124_500delins[AB053210.2:r.1289-365_1289-73inv]' --axis c --reference <ref>
NM_004006.3:c.-114_263delins[AB053210.2:r.1289-365_1289-73inv]

$ ferro project 'NM_004006.3:c.-114_263delins[AB053210.2:r.1289-365_1289-73inv]' --axis c --reference <ref>
ERROR: Unsupported variant type: ins[AB053210.2:r.1289-365_1289-73inv] cross-reference shape not yet supported
```

So ferro emits a `c.` string that fails its own validation. (The `c.` refusal is correct — that payload has out-of-scope offsets *and* `AB053210.2` is absent from the reference.)

## Cause

Only four expansion wrappers exist — `try_expand_genome_ins`, `try_expand_cds_ins`, `try_expand_tx_ins`, `try_expand_mt_ins` (`src/normalize/mod.rs:2056`, `:2079`, `:2103`, `:2127`). There is no `r.` one. The code documents it at `src/normalize/mod.rs:4945-4947`:

> The RNA axis does not run the issue #333 `ins[...]` expansion step — that is wired only into the genome/cds/tx/mt axes.

\#773 expanded `r.` as a *payload* axis, not as the *outer* axis, which is why the gap survived.

[...]

Issue #736, opened 2026-06-17
fix(normalize): r. insertions/delins with the RNA base `u` fail to normalize (U≠T comparison)


nh13:
## Summary

`normalize_rna` compares an `r.` edit's literal bases against the DNA-stored reference sequence byte-for-byte. The parser keeps the RNA base `u` as a distinct `Base::U` (`b'U'`), separate from `Base::T` (`b'T'`), so any `r.` edit carrying a literal `u` never matches the reference `T`. Insertions and delins therefore fail to canonicalize / 3'-shift — the spec-correct `u` form yields a *wrong* result, while the technically-invalid `T` form yields the right one.

## Reproduction

Single-exon transcript `AATTTGCC` (a `TTT` run at positions 3–5), `cds_start=1`:

| input | output | expected |
|-------|--------|----------|
| `c.3_4insT` | `c.5dup` | `c.5dup` ✓ |
| `n.3_4insT` | `n.5dup` | `n.5dup` ✓ |
| `r.3_4insT` | `r.5dup` | `r.5dup` ✓ |
| `r.3_4insu` | `r.3_4insu` | **`r.5dup`** ✗ |
| `r.3_4insuu` | `r.3_4insuu` | **`r.4_5dup`** ✗ |
| `r.3_5delinsuu` | `r.3_5delinsuu` | simplified ✗ |

Edits without explicit literal bases (`dup`, `del`) and substitutions (which don't shift) are unaffected.

## Root cause

In `normalize_rna`, the edit's inserted/delins bases are converted to bytes via `*b as u8` and compared against the DNA reference in `normalize_na_edit` (validation, ins→dup recognition, the shuffle predicate). `Base::U as u8 == b'U'` never equals `b'T'`, even under `eq_ignore_ascii_case`.

## Fix direction

Map the `r.` edit's literal `U` bases to `T` before the DNA-based normalization (mirroring the existing `validate_no_u_in_dna` walker over `NaEdit` shapes), then let `RnaVariant`'s Display (`to_rna_string`) render `T`→`u` on output. Affects both the exonic `r.` path and the intronic `r.` path (the latter added in #704).

PR #1194, opened 2026-07-25
fix(normalize): apply the codon-frame gate on the r. axis

### LINKED ISSUE #1192: fix(normalize): the r. axis never applies the codon-frame gate to repeat notation
## Summary

`normalize_rna` never wires the codon-frame gate, so the HGVS restriction that repeat notation on a **coding** reference may only be used for repeat units whose length is a multiple of 3 is **never applied on the `r.` axis**. ferro consequently emits, on `r.`, the exact two descriptions the spec marks `invalid`.

`src/normalize/mod.rs:5153` (on `be47233`) passes `is_coding = false` unconditionally:

```rust
// Perform normalization (RNA context: codon-frame gate does not apply;
// r. is not in the spec's accepted reference types for repeats).
let (new_tx_start, new_tx_end, new_edit, mut warnings) =
    self.normalize_na_edit(seq, edit, tx_start, tx_end, &boundaries, false)?;
```

The comment's premise is factually wrong. `docs/recommendations/RNA/repeated.md` is an entire page about repeat notation on an RNA reference, and it states the same codon-frame exception the DNA page does.

## Spec

`RNA/repeated.md:24-27`:

> - **exception**: using a coding RNA reference sequence, a repeated sequence variant description can be used only for repeat units with a length which is a multiple of 3, i.e. which can not affect the reading frame.
>   Consequently, use `NM_024312.4:r.2692_2693dup` and **not** `NM_024312.4:r.2686a[10]`; use `NM_024312.4:r.1741_1742insuauauaua` and **not** `NM_024312.4:r.1738ua[6]`.
>   This restriction only applies to the coding sequence, which does not include the UTR sequence.
>   As such, `NM_024312.4:r.-6_-3g[6]` is valid as the reading frame is not affected.

This mirrors `DNA/repeated.md` exactly, including the UTR carve-out — which `normalize_cds` already implements correctly (`src/normalize/mod.rs:3020-3037`).

## Reproduction

Clean `main` (`be47233`), real prepared reference. Both inputs are the spec's own examples, and both are entirely within the CDS proper of `NM_024312.4`:

```
$ ferro normalize NM_024312.4:r.2692_2693dup --reference $REFDIR
NM_024312.4:r.2692_2693dup -> NM_024312.4:r.2686_2693a[10]     # spec says NOT this

$ ferro normalize NM_024312.4:r.1741_1742insuauauaua --reference $REFDIR
NM_024312.4:r.1741_1742insuauauaua -> NM_024312.4:r.1738_1741ua[6]   # spec says NOT this
```

The `c.` spelling of the same two variants shows the gate working, which is the asymmetry:

```
$ ferro normalize NM_024312.4:c.2692_2693dup --reference $REFDIR
NM_024312.4:c.2692_2693dup -> NM_024312.4:c.2693_2694insAA     # gate fired, rerouted to ins literal

$ ferro normalize NM_024312.4:c.1741_1742insTATATATA --reference $REFDIR
NM_024312.4:c.1741_1742insTATATATA                              # gate fired, stays an ins
```

That the `c.` runs reroute at all proves `is_coding` computed `true` for these spans, i.e. both tracts lie entirely inside the CDS proper — so the `r.` spellings of the same spans must be gated too.

## Root cause

`r.` on a coding transcript is CDS-relative — the same axis as `c.` (`background/numbering.md` L58/L61, pinned by #469 and `tests/it/issue_291_rna_axis_convention.rs`). `normalize_rna` already computes transcript-frame coordinates via `map_in` and already has `transcript.cds_start` / `transcript.cds_end` in hand (`cds_info`, `mod.rs:5095`), but throws the CDS information away at the `normalize_na_edit` call and hardcodes `false`.

## Expected

On a **coding** transcript, an `r.` variant whose footprint lies entirely within the CDS proper should get the same codon-frame gate `normalize_cds` applies: a non-triplet repeat unit must not be rendered as `unit[k]`, and should fall back the way the `c.` axis falls back.

It must **not** apply:

- when the tract touches a UTR (`r.-N` / `r.*N`) — the spec's carve-out, same as `c.`
- on a non-coding transcript (`NR_`), where `r.` is `n.`-equivalent and there is no reading frame

## Proposed fix

Derive `is_coding` in `normalize_rna` the same way `normalize_cds` does (`tx_start >= cds_start && tx_end <= cds_end`, `false` when `cds_end` is unset) and pass it through instead of the hardcoded `false`. Replace the incorrect comment.

Keep the change **additive and narrow**. #1189 documented a trap here: `normalize_na_edit`'s recursive calls forward the outer gate verdict *by design*, so a "cleaner" refactor that recomputes the verdict per call site breaks normalization idempotency. `normalize_idempotency_proptest` is the test that catches it and must be run.

## Tests

Failing test first, plus controls:

1. coding transcript, non-triplet tandem tract entirely inside the CDS, spelled on `r.` → must not render as `unit[k]`
2. codon-aligned (multiple-of-3) unit on `r.` inside the CDS → still gets repeat notation
3. a UTR tract on `r.` → keeps repeat notation at any unit length (the spec's `r.-6_-3[6]` shape)
4. an `NR_` transcript → unaffected

## Blast radius

This changes `r.`-axis repeat output, so conformance rows and possibly `tests/fixtures/grammar/hgvs_spec_enumeration.json` will move. Both spec fixtures are `.gitignore`d generated artifacts, so the delta must be measured by toggling only the `src` change in a single worktree and regenerating both fixtures each time — never by diffing fixtures across worktrees.

## Related

- #1183 is a different `r.`-axis gap (the axis skips `ins[...]` cross-reference expansion), but the same theme: machinery `c.` has and `r.` is missing. Worth considering together.
- #1185 / #1189 fixed the sibling defect on the `c.` axis, where the gate consulted the input span instead of the shuffled tract.

## Aside (out of scope)

The spec's preferred fallback for the first example is `r.2692_2693dup`, whereas ferro's `c.` fallback renders `c.2693_2694insAA`. Whether the gated fallback should prefer `dup` over an `ins` literal is a separate question about existing `c.` behavior and is not part of this issue.

nh13:
Closes #1192.

## The defect

`normalize_rna` passed `is_coding = false` to `normalize_na_edit` unconditionally, so the HGVS restriction that repeat notation on a coding reference may only use repeat units whose length is a multiple of 3 was **never applied on the `r.` axis**. ferro consequently emitted, on `r.`, the exact two descriptions `RNA/repeated.md` L24-27 marks invalid — verified against the real prepared reference on clean `main` (`be47233`):

| input | before | after |
|---|---|---|
| `NM_024312.4:r.2692_2693dup` | `r.2686_2693a[10]` ❌ | `r.2693_2694insaa` |
| `NM_024312.4:r.1741_1742insuauauaua` | `r.1738_1741ua[6]` ❌ | unchanged |
| `NM_024312.4:c.2692_2693dup` | `c.2693_2694insAA` | unchanged |

The spec names those two forbidden forms verbatim:

> **exception**: using a coding RNA reference sequence, a repeated sequence variant description can be used only for repeat units with a length which is a multiple of 3, i.e. which can not affect the reading frame. Consequently, use `NM_024312.4:r.2692_2693dup` and **not** `NM_024312.4:r.2686a[10]`; use `NM_024312.4:r.1741_1742insuauauaua` and **not** `NM_024312.4:r.1738ua[6]`. This restriction only applies to the coding sequence, which does not include the UTR sequence. As such, `NM_024312.4:r.-6_-3g[6]` is valid as the reading frame is not affected.

The comment justifying the hardcoded `false` claimed `r.` "is not in the spec's accepted reference types for repeats". `RNA/repeated.md` is an entire page about repeat notation on an RNA reference and states the same exception the DNA page does, including the same UTR carve-out, so that premise was simply wrong.

## The fix

Derive `is_coding` the way `normalize_cds` already does — deliberately the identical condition, since `r.` on a coding transcript is CDS-relative, i.e. the same axis as `c.` (#469):

```rust
let is_coding = match cds_info {
    Some((cds_start, cds_end)) => tx_start >= cds_start && tx_end <= cds_end,
    None => false,
};
```

A non-coding transcript has no `cds_info` and so never gates; a footprint touching either UTR falls outside `cds_start..=cds_end` and is exempt by the spec's carve-out.

**The change is additive and narrow on purpose.** #1185/#1189 established that `normalize_na_edit`'s recursive calls forward the outer gate verdict *by design*, so a "cleaner" refactor that recomputes the verdict per call site breaks normalization idempotency. Exactly one call site changes here. `normalize_idempotency_proptest` at `PROPTEST_CASES=100000` and the full suite under `FERRO_ASSERT_IDEMPOTENT=1` are both green.

## Tests

`tests/it/issue_1192_rna_codon_frame_gate.rs` — hand-rolled fixtures (not `SyntheticBuilder`, whose `ACGT` padding a tandem unit can cycle straight into; a padding artifact of that shape produced a retracted finding once before). Every case is asserted on **both** axes, which is what makes this a spec defect rather than a fixture artifact: the two axes address identical reference bases, so any divergence between them is the bug.

Before the fix exactly one of the seven fails, with `r.4_9a[8]` against an expected `r.9_10insaa`; the other six — including all four controls — pass unchanged both before and after:

- codon-aligned (multiple-of-3) unit inside the CDS still earns repeat notation, on `c.` and on `r.`
- a homopolymer in the **3'UTR** keeps repeat notation, on `c.` and on `r.` (the spec's carve-out)
- an `NR_` transcript never gates

### Re-blessed test

`issue_736_rna_uracil_normalize::rna_two_base_u_insertion_canonicalizes_to_repeat` pinned the defect and is re-blessed (and renamed to `..._is_gated_like_the_c_axis`). Its expectation `r.3_5u[5]` was justified as "the `r.` analog of `n.3_4insTT` -> `n.3_5T[5]`" — which compared `r.` against the wrong axis. That fixture's transcript is coding (`cds_start = 1`), so `r.` is the `c.` axis, not the `n.` axis. On the same transcript and bases the three axes now read:

```
c.3_4insTT  -> c.5_6insTT   (gated: coding, unit not codon-aligned)
r.3_4insuu  -> r.5_6insuu   (must match c.)
n.3_4insTT  -> n.3_5T[5]    (no reading frame, so repeat notation is fine)
```

The `u` spelling of the inserted bases — the actual point of #736 — survives the round trip, which is what that file exists to pin.

## Blast radius

Measured by toggling only the `src` change in **one** worktree and regenerating both generated spec fixtures each time (never by diffing fixtures across worktrees — both are `.gitignore`d build artifacts, and the enumeration is generated *from* the normalization fixture, so a cross-worktree diff compares two independently-generated files):

- **normalization fixture:** byte-identical.
- **enumeration fixture:** 0 rows added, 0 removed, **16 changed** — 4 spec inputs × 4 axes, and all 4 inputs are `RNA/repeated.md`'s own counter-examples. No budgeted divergence status moves (`projection-pinned` 1203→1207 and `projection-unavailable-pinned` 526→522 are both in `KNOWN_PASSING_STATUSES`, not `DIVERGENCE_BUDGET`).
- **reference-backed conformance:** unchanged. The full suite with `FERRO_MANIFEST` set still fails only `mutalyzer_normalize_tests::axis_errors` on the two rows that already fail on `main` (`NM_000143.3:c.45del4` / `c.45dup4`, #1173's territory). No ledger re-bless needed.

Three of the four changed inputs improve on axes beyond `r.`:

- **`project-c`** previously emitted `c.1546_1549TA[6]` / `c.2494_2501A[10]` — descriptions the `c.` axis would never produce natively, because they are gated there. It now emits the ins literals, matching native `c.` normalization. That is a second, latent spec violation this fix cures.
- **`project-p`** went from `unavailable: no p. representation for this variant` to real predictions (`p.(Lys517IlefsTer33)`, `p.(Pro835SerfsTer5)`) for 4 rows: repeat notation had no protein projection, the ins form does. Both are frameshifts — which is precisely *why* the spec forbids the repeat spelling here.
- **`project-n`** now shows the ins literal (`n.1741_1742insTATATATA`) where it previously showed `n.1738_1741TA[6]`. Flagging this honestly: projection re-expresses an already-normalized edit rather than re-normalizing it, so the `n.` view inherits the gated `r.` spelling. Both describe the same change and neither is invalid on `n.`, but a *native* `n.` normalization of those bases would still produce the repeat form. It is a fidelity change, not a correctness violation.

## Not in scope

The spec's preferred fallback for the first example is `r.2692_2693dup`, whereas ferro's gated fallback renders `insaa` on **both** axes. Whether the fallback should prefer `dup` over an `ins` literal is a pre-existing question about `c.`-axis behavior, unchanged by this PR.

#1183 is a different `r.`-axis gap (the axis skips `ins[...]` cross-reference expansion) with the same underlying theme — machinery `c.` has that `r.` is missing.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -5169,15 +5169,35 @@ impl<P: ReferenceProvider> Normalizer<P> {
             Boundaries::new(0, transcript.sequence_length())
         };
 
-        // Perform normalization (RNA context: codon-frame gate does not apply;
-        // r. is not in the spec's accepted reference types for repeats).
         // Coordinate-only transcripts fall back to the canonicalize-only path.
         let seq = match transcript.sequence.as_deref() {
             Some(s) => s.as_bytes(),
             None => return Ok((HV::Rna(variant.clone()), vec![])),
         };
+
+        // #1192: the codon-frame gate applies on `r.` exactly as it does on
+        // `c.`. `RNA/repeated.md` L24-27 states the restriction for an RNA
+        // reference in the same terms `DNA/repeated.md` uses:
+        //   > using a coding RNA reference sequence, a repeated sequence
+        //   > variant description can be used only for repeat units with a
+        //   > length which is a multiple of 3 [...] This restriction only
+        //   > applies to the coding sequence, which does not include the UTR
+        //   > sequence.
+        // and gives `NM_024312.4:r.2686a[10]` as an explicit counter-example.
+        // (The previous comment here asserted that `r.` is not an accepted
+        // reference type for repeats, which that page contradicts.)
+        //
+        // The condition mirrors `normalize_cds` exactly — deliberately, since
+        // `r.` on a coding transcript is CDS-relative, i.e. the same axis as
+        // `c.` (#469). `cds_info` is `None` for a non-coding transcript, which
+        // has no reading frame to preserve, and a footprint touching either UTR
+        // falls outside `cds_start..=cds_end` and is exempt by the carve-out.
+        let is_coding = match cds_info {
+            Some((cds_start, cds_end)) => tx_start >= cds_start && tx_end <= cds_end,
+            None => false,
+        };
         let (new_tx_start, new_tx_end, new_edit, mut warnings) =
-            self.normalize_na_edit(seq, edit, tx_start, tx_end, &boundaries, false)?;
+            self.normalize_na_edit(seq, edit, tx_start, tx_end, &boundaries, is_coding)?;
 
         // See the identical guard + comment in `normalize_cds` (#1052 follow-up):
         // substitutions are validated-then-returned-unchanged by
--- a/tests/it/issue_1192_rna_codon_frame_gate.rs
+++ b/tests/it/issue_1192_rna_codon_frame_gate.rs
@@ -1,0 +1,228 @@
+//! Issue #1192 — the `r.` axis must apply the codon-frame gate to repeat
+//! notation, exactly as the `c.` axis does.
+//!
+//! `RNA/repeated.md` L24-27 states the restriction for an RNA reference in the
+//! same terms the DNA page uses, including the UTR carve-out:
+//!
+//! > **exception**: using a coding RNA reference sequence, a repeated sequence
+//! > variant description can be used only for repeat units with a length which
+//! > is a multiple of 3, i.e. which can not affect the reading frame.
+//! > Consequently, use `NM_024312.4:r.2692_2693dup` and **not**
+//! > `NM_024312.4:r.2686a[10]` … This restriction only applies to the coding
+//! > sequence, which does not include the UTR sequence. As such,
+//! > `NM_024312.4:r.-6_-3g[6]` is valid as the reading frame is not affected.
+//!
+//! `r.` on a coding transcript is CDS-relative — the same axis as `c.`
+//! (`background/numbering.md` L58/L61, pinned by #469) — so a `c.`/`r.` pair
+//! over the same bases must reach the same decision. `normalize_rna` used to
+//! pass `is_coding = false` unconditionally, so the gate never fired and ferro
+//! emitted, on `r.`, the very form the spec marks invalid: with the real
+//! reference, `NM_024312.4:r.2692_2693dup` rendered as `r.2686_2693a[10]`.
+//!
+//! Each case is asserted on **both** axes. That pairing is what makes this a
+//! spec defect rather than a fixture artifact: the two axes address identical
+//! reference bases, so any divergence between them is the bug itself.
+//!
+//! The repro is a homopolymer duplication because that is the shape the spec's
+//! own counter-example takes — a unit length of 1 is never a multiple of 3, so
+//! it can never be codon-aligned.
+
+use ferro_hgvs::reference::transcript::{Exon, ManeStatus, Strand, Transcript};
+use ferro_hgvs::{parse_hgvs, MockProvider, Normalizer};
+
+/// Length of the 5'UTR, so `c.1` / `r.1` is transcript base `UTR5_LEN + 1`.
+const UTR5_LEN: usize = 30;
+
+/// Build a single-exon coding transcript `C * 30` (5'UTR) ++ `cds` ++ `utr3`.
+///
+/// Hand-rolled rather than built with `SyntheticBuilder`: that helper pads with
+/// `ACGT` repeats, which a tandem unit can cycle straight into, and a padding
+/// artifact of exactly that shape produced a retracted finding once before.
+/// Every tract below is flanked by a base that cannot continue it.
+fn coding_transcript(cds: &str, utr3: &str) -> MockProvider {
+    let sequence = format!("{}{cds}{utr3}", "C".repeat(UTR5_LEN));
+    let tx_len = sequence.len() as u64;
+    let mut provider = MockProvider::new();
+    provider.add_transcript(Transcript::new(
+        "NM_RFRAME.1".to_string(),
+        Some("RFRAME_TEST".to_string()),
+        Strand::Plus,
+        sequence,
+        Some((UTR5_LEN + 1) as u64),
+        Some((UTR5_LEN + cds.len()) as u64),
+        vec![Exon::new(1, 1, tx_len)],
+        None,
+        None,
+        None,
+        Default::default(),
+        ManeStatus::None,
+        None,
+        None,
+    ));
+    provider
+}
+
+/// Build a single-exon **non-coding** transcript, with no CDS annotation at all.
+fn noncoding_transcript(sequence: &str) -> MockProvider {
+    let tx_len = sequence.len() as u64;
+    let mut provider = MockProvider::new();
+    provider.add_transcript(Transcript::new(
+        "NR_RFRAME.1".to_string(),
+        Some("RFRAME_NC_TEST".to_string()),
+        Strand::Plus,
+        sequence.to_string(),
+        None,
+        None,
+        vec![Exon::new(1, 1, tx_len)],
+        None,
+        None,
+        None,
+        Default::default(),
+        ManeStatus::None,
+        None,
+        None,
+    ));
+    provider
+}
+
+/// CDS `ATG` ++ `A`×6 ++ `GGGTAA`: a homopolymer tract at `c.4_9`, flanked by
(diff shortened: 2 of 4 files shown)


nh13:
## Merge hazard: #1189 and #1194 silently cancel each other at one call site

These two PRs touch the same line — `normalize_rna`'s `normalize_na_edit` call — and a naive conflict resolution **reverts #1192's fix without failing to compile**. Verified empirically, not inferred.

#1189 adds `cds_span` and derives the `duplication_to_repeat` gate **purely** from it, not from `is_coding`:

```rust
matches!(cds_span, Some((cds_s, cds_e)) if /* shuffled tract inside CDS */)
```

so any caller passing `cds_span: None` gets "not gated" regardless of what `is_coding` says. #1189 passes `None` on the `r.` path; #1194 wires `is_coding` there. #1192's primary repro is a `dup`, which goes through exactly that call.

**Measured** by applying #1194's change onto #1189's branch and running #1194's tests:

| `normalize_rna` passes | `r.8_9dup` (CDS-resident homopolymer) | verdict |
|---|---|---|
| `false, None` (#1189 today) | `r.4_9a[8]` | pre-fix, spec-forbidden |
| `is_coding, None` (naive merge) | `r.4_9a[8]` | **still pre-fix — the silent revert** |
| `is_coding, cds_info` (correct) | `r.9_10insaa` | 7/7 of #1194's tests pass |

**Resolution when the second of these two merges:** pass `cds_info` as `cds_span`, not `None`.

```rust
let is_coding = match cds_info {
    Some((cds_start, cds_end)) => tx_start >= cds_start && tx_end <= cds_end,
    None => false,
};
let (new_tx_start, new_tx_end, new_edit, mut warnings) =
    self.normalize_na_edit(seq, edit, tx_start, tx_end, &boundaries, is_coding, cds_info)?;
```

`cds_info` is already in scope in `normalize_rna` and is exactly the `cds_start.zip(cds_end)` pair `normalize_cds` builds. Passing it also gives the `r.` axis #1189's shuffled-tract treatment, which is the point of both PRs: the two axes should reach the same verdict on the same bases.

The full suite in that combined state is green apart from the two files #1194 already changes — `issue_736_rna_uracil_normalize` (fails with exactly the `r.5_6insuu` re-bless #1194 makes) and the regenerated enumeration fixture (all 8 `spec_enumeration_tests` pass once regenerated).

Related, already tracked: #1180 pins the pre-#1189 two-pass behavior in `tests/it/cds_utr3_crossing_shift_idempotency.rs`, which must be deleted — and `c.`-with-3'UTR re-added to `ALL_SYSTEMS` — when #1180 and #1189 both land.

Issue #1205, opened 2026-07-26
normalize: 5'-saturated insertion at a contig start emits invalid single-position HGVS on the g. axis


nh13:
A 5'-shuffled insertion that saturates at the **start of a genomic contig** is emitted as a single-position insertion, which ferro's own parser then rejects. This is the `g.`-axis analogue of #1202, and it is the shape the spec names as invalid using the `g.` axis specifically.

## Repro

Genomic sequence `CCCCAAAAAAAGGGG` under accession `NC_PROBE.1` (positions 1-4 are `C`), 5'-shuffle direction:

```
NC_PROBE.1:g.1_2insGC   ->   NC_PROBE.1:g.1insCG
```

and `parse_hgvs("NC_PROBE.1:g.1insCG")` returns an error — normalization produced a description ferro cannot read back.

The 3' direction on the same axis is not affected in the same way (`g.13_14insGCA` -> `g.15_16insAGC` re-parses), though its second anchor is past the contig end on a 15-base contig, which may deserve its own look.

## Spec

`recommendations/DNA/insertion.md` lines 95-101 rejects exactly this shape, and states it on the `g.` axis:

> **Can I describe a variant as `g.123insG`?**
> No, since the description is not unequivocal, it is not allowed.
> What does the description mean, the insertion of a `G` **at** position `g.123` or the insertion of a `G` **after** position `g.123`?

An insertion needs two adjacent flanking anchors; `g.1ins…` has one.

## Cause

`normalize_genome` has no boundary clamp. #1202's fix introduced `clamp_insertion_at_transcript_bounds` and wired it into `normalize_cds`, `normalize_tx` and `normalize_rna` — three call sites, all axes numbered against a transcript. `normalize_genome` (and `normalize_mt`) were left out, since #1202 was scoped to the transcript bounds.

The mechanism is identical to the transcript case: the 0-based 5' shuffle result of `0` clamps back through `saturating_sub(1)` to the degenerate `start == end == 1`.

## Suggested shape

[...]

Issue #951, opened 2026-07-06
fix(normalize): circular o. normalization is a silent no-op (deferral tracker #129 is closed)


nh13:
## Summary

`o.` (circular, SVD-WG006) normalization is a **silent no-op**: `Normalizer::normalize` clones the variant and returns it with no 3'-shift and no warning, while the inline comment claims it "normalize[s] like genomic variants" — which overstates what happens. The deferral this points at, **#129, is CLOSED**, so the gap has no open tracker.

## Detail

- `src/normalize/mod.rs:1410-1417` — the `HgvsVariant::Circular` arm returns a clone; no shuffle runs.
- The real deferral note lives at `src/normalize/mod.rs:7126` (references #129). #129 is closed (it was scoped to mitochondrial handling), so circular-`o.` shuffling is currently untracked.

## Ask

Either implement circular normalization (3'-shift with wraparound semantics) or, at minimum, correct the misleading "normalize like genomic variants" comment and keep this issue as the open tracker for the deferral. A `g.`-style genomic shuffle on a circular contig also needs to respect the origin wraparound (cf. the MT wraparound decline at `mod.rs:4498`).

nh13:
Superseded by the spec-question tracker. On a fresh reading of the spec, the substantive residue here is a **spec ambiguity, not a ferro conformance bug**: the HGVS text never ties circular (`o.`/`m.`) references to the 3'rule or defines origin-wraparound shift semantics (verified by exhaustive search of `assets/hgvs-nomenclature/docs`). The only explicit circular provision is the wraparound *listing* exception (`deletion.md:16-17`), which is notation, not shifting.

That ambiguity is now filed as a candidate on the spec-revision tracker **#466** ("Circular (`o.`/`m.`) 3'-rule normalization — origin-wraparound shift semantics undefined"), to be raised with SVD-WG.

The honest-comment fix this issue also asked for already landed (`src/normalize/mod.rs` — the misleading "normalize like genomic variants" comment now states plainly that `o.` is an unchanged no-op). The remaining ferro work — a non-origin `o.` 3'-shift, only *inferable* from the general 3'rule — is low-priority (`o.` is rare; `m.` has its own origin-crossing decline path) and deferred pending the #466 clarification; it can be re-filed if/when the spec resolves and `o.` normalization is actually needed.

Closing as not-planned (tracked upstream via #466).

PR #1222, opened 2026-07-26
fix(normalize): clamp an insertion saturated at a mitochondrial terminus

### LINKED ISSUE #1217: normalize: insertion saturated at a mitochondrial terminus emits an invalid anchor on the m. axis
`normalize_mt` has no boundary clamp, so an insertion driven to rest against either terminus of the circular reference is emitted with an anchor that does not exist. This is the `m.` analogue of #1202 (transcript bounds) and #1205 (contig bounds), and it is the last uncovered axis.

Both shapes reproduce **on the real rCRS** (`NC_012920.1`, GRCh38 prepared reference), not only on synthetic fixtures.

## Repro

```
$ ferro normalize --reference <ref> "NC_012920.1:m.16568_16569insGC"
NC_012920.1:m.16568_16569insGC -> NC_012920.1:m.16569_16570insCG

$ ferro normalize --reference <ref> --direction 5prime "NC_012920.1:m.1_2insCG"
NC_012920.1:m.1_2insCG -> NC_012920.1:m.1insGC

$ ferro parse "NC_012920.1:m.1insGC"
Error: "1 variant(s) failed to parse"
```

- **3' (the default direction):** `m.16570` is past the end of the 16,569-base mitochondrial genome — the W4004 `PositionPastEnd` shape.
- **5':** a single-position insertion, which ferro's own parser rejects. `DNA/insertion.md:95-101` rejects that shape by name ("the description is not unequivocal"), so normalization produced a description ferro cannot read back.

The rCRS terminal bases are what make this reachable without a contrived reference: it ends `…ACATCACGATG` and begins `GATCACAGGT…`, so a `G` sits on both sides of the origin and a payload can shuffle onto the terminus.

**A `g.` variant on a mitochondrial accession is affected too.** `normalize_core` coerces `HV::Genome` to the mito path when `accession.is_mitochondrial()`, so #1205's contig clamp never sees it:

```
$ ferro normalize --reference <ref> "NC_012920.1:g.16568_16569insGC"
NC_012920.1:g.16568_16569insGC -> NC_012920.1:m.16569_16570insCG
```

`o.` is unaffected: circular `o.` variants are returned unchanged (no shuffle runs at all), per the comment at the `HV::Allele` arm in `normalize_core`.

## Cause

`clamp_insertion_at_sequence_bounds` (renamed from `clamp_insertion_at_transcript_bounds` in #1205) now has four callers — `normalize_cds`, `normalize_tx`, `normalize_rna`, `normalize_genome`. `normalize_mt` is not one of them. #1205 deliberately left it out on the grounds that a circular reference wraps, so position 1 has a valid 5' neighbour and the answer ought to be a wraparound description rather than a `delins`.

## That rationale does not survive contact with #129

#129 established, and `issue_129_mt_circular_wraparound.rs` pins, that ferro **rejects** the wraparound `ins` spelling by design:

> Parser accepts `m.<high>_<low>del`, `m.<high>_<low>delins`, and `m.<high>_<low>dup` (and the same on `o.`).
> Parser **rejects** `m.<high>_<low>ins`/`inv` (were silently accepted; mis-computed indel length).

The spec's reversed-range exception (`deletion.md:17`, SVD-WG006) is granted to deletions and duplications, and `delins` inherits it via composition; `insertion.md` is silent, so the general 5'→3' rule applies. Both mutalyzer and biocommons reject *all* reversed ranges. So `m.16569_1insCG` is not an available answer — it is a form this project decided not to emit.

What that leaves is the `delins` rewrite, and it works on a circular reference without needing a reversed range at all:

```
3'  m.16568_16569insGC  ->  m.16569delinsGCG   (delete m.16569 = G, insert G ++ CG)
5'  m.1_2insCG          ->  m.1delinsGCG       (delete m.1     = G, insert GC ++ G)
```

Both are plain single-position `delins` forms, both describe inserting the payload at the origin junction, and both re-parse. This is the same coordinate identity `insertion_to_boundary_delins` already proves once for the other three axes.

Note also that #129 disabled 3'-rule wraparound shifting ("a homopolymer straddling the origin won't shift across it"), matching mutalyzer and biocommons. So the shuffle stopping at the terminus is already the intended behaviour — what is broken is only how that resting place is rendered.

## Suggested shape

Add `normalize_mt` as the fifth caller of `clamp_insertion_at_sequence_bounds`, with `SequenceEnds::WHOLE` if it normalizes over the full contig, or computed flushness if it uses the same windowed fetch `normalize_genome` does (it shares `clamp_fetch_end_to_contig`, so check which).

Worth deciding explicitly rather than by default, since it is the one judgement call here: whether a circular reference should instead shuffle *through* the origin and emit a wraparound `dup`/`delins` where one applies. #129 says no for shifting, and that answer looks right to keep — but this issue is the place to record it if the project wants to revisit it, because "position 1 has a valid 5' neighbour" is true of the sequence even though it is not currently true of the notation ferro emits.

## Scope

Found while fixing #1205, which covers `g.` on non-mitochondrial accessions. Filed separately because the circular axis needed its own answer and #1205 should not decide it silently. #1205's `mito_axis_is_not_clamped` test pins the current scope boundary — it asserts only that the genomic clamp does not reach `m.`, and does not bless the invalid output; that test should be revisited when this is fixed.

nh13:
`normalize_mt` was the one caller of `clamp_insertion_at_sequence_bounds` never wired up, so an insertion the shuffle drove against either end of the mitochondrial reference was emitted with an anchor that does not exist.

| direction | input | before | after |
|---|---|---|---|
| 5' | `m.1_2insGC` | `m.1insCG` | `m.1delinsCGC` |
| 3' | `m.13_14insGGA` | `m.15_16insAGG` | `m.15delinsGAGG` |

The 5' shape is a single-position insertion — the form `DNA/insertion.md:95-101` rejects by name ("the description is not unequivocal"), and which ferro's own parser refuses, so normalization was producing output it could not read back. The 3' shape's second anchor is past the contig end (W4004 `PositionPastEnd`).

A `g.`-spelled variant on a mitochondrial accession is affected too: `normalize_core` coerces `HV::Genome` to the mito path when the accession `is_mitochondrial()` (#487), so #1205's contig clamp never saw it and the fix has to land in `normalize_mt` to cover that spelling.

## Why `delins` and not a wraparound `ins`

#1205 deliberately left `m.` out, reasoning that a circular reference wraps — position 1 *has* a valid 5' neighbour — so the right answer was a wraparound description rather than a `delins` rewrite. That rationale does not survive contact with #129, which established (and `issue_129_mt_circular_wraparound.rs` pins) that ferro **rejects** `m.<high>_<low>ins` by design. The spec's reversed-range exception (`deletion.md:17`, SVD-WG006) is granted to `del`/`dup`, and `delins` inherits it by composition, while `insertion.md` is silent — so the general 5'→3' rule applies, which is also how mutalyzer and biocommons read it.

So `m.16569_1insCG` is not an available answer. The single-position `delins` is, and it needs no reversed range at all — it is the same coordinate identity `insertion_to_boundary_delins` already applies on the other four axes. #129 separately disabled 3'-rule shifting across the origin, so the shuffle coming to rest at the terminus is already the intended behaviour; only its rendering was wrong.

`o.` stays out of scope: `normalize_core` returns circular `o.` variants unchanged, so no shuffle runs and nothing can saturate. Pinned by a test. A genuine circular normalizer remains #951.

## Changes

- `normalize_mt` becomes the fifth caller of `clamp_insertion_at_sequence_bounds`, computing `SequenceEnds` flushness exactly as `normalize_genome` does — `ref_seq` is a window, so an end is a real contig bound only where the window is flush with it, and with no length available we do not clamp.
- The helper's doc-comment axis table is corrected: `m.` moves from "circular, so position 1 has a valid 5' neighbour — n/a" to a real clamp site, with the #129 reasoning recorded, and `o.` is restated as out of scope for a different reason (no shuffle runs at all).
- #1205's `mito_axis_is_not_clamped` is removed and replaced by a comment pointing at the new file. That test pinned exactly the scope boundary this PR closes; it asserted only that the clamp had not fired and explicitly did **not** bless the invalid output, so removing it loses no coverage.

## Tests

Written first, confirmed failing for the right reason before the fix — the two red assertions were `m.1insCG` and `m.15_16insAGG`, matching the issue's repro exactly.

`tests/it/issue_1217_mito_terminus_insertion_clamp.rs`: both directions, the `g.`-coercion path, a hand-spliced haplotype-equivalence check that does not go through the normalizer, a re-parse + fixed-point assertion on every case, an explicit check that no wraparound `ins` is emitted, and controls (interior insertions on both sides, repeat notation reaching the start, a non-shuffling insertion, and `o.` untouched).

Also included is a reference-gated test pinning the issue's exact rCRS strings — `NC_012920.1:m.16568_16569insGC` → `m.16569delinsGCG` and `m.1_2insCG` (5') → `m.1delinsGCG` — which is what shows the defect was reachable on the real mitochondrial reference rather than only on a contrived contig.

**Evidence note:** the mock-backed tests above are the ones I ran. The reference-gated test **skipped** in my local runs — no prepared reference was reachable from this machine (neither the scratch nor the backup volume is currently mounted), so `FERRO_MANIFEST` was unset. Its expected strings are derived from the rCRS terminal bases quoted in the issue (`…ACATCACGATG` / `GATCACAGGT…`), not from an observed run here. It will execute for the first time in the nightly reference-aware job.

## Validation

- `cargo nextest run --features dev` → 8731 passed, 0 failed, 145 skipped
- `FERRO_ASSERT_IDEMPOTENT=1 cargo nextest run --features dev` → 8731 passed, 0 failed
- `cargo fmt --check`, `cargo clippy --all-features --all-targets -- -D warnings` → clean

Suite size moves 8720 → 8731: twelve new tests, less #1205's removed scope-boundary test.

Closes #1217

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -682,7 +682,8 @@ impl SequenceEnds {
 /// | `c.` (UTR regions) | transcript `1` / `len` | this helper |
 /// | `n.`, `r.` | transcript `1` / `len` | this helper |
 /// | `g.` | contig `1` / `len` | `normalize_genome` (#1205) |
-/// | `m.`, `o.` | circular, so position 1 has a valid 5' neighbour | n/a |
+/// | `m.` | contig `1` / `len` | `normalize_mt` (#1217) |
+/// | `o.` | n/a — returned unchanged, no shuffle runs | n/a |
 ///
 /// `normalize_genome` became the fourth caller in #1205, which closes the last
 /// non-circular gap: before it, a 5'-saturated insertion at a contig start
@@ -691,9 +692,24 @@ impl SequenceEnds {
 /// `g.123insG`) — and a 3'-saturated one as `g.<len>_<len+1>ins<A\'>`, whose
 /// second anchor is past the contig end.
 ///
-/// `m.`/`o.` remain deliberately uncovered: a circular reference wraps, so
-/// position 1 HAS a valid 5\' neighbour (the last base) and the correct output
-/// there is a wraparound description, not this `delins` rewrite.
+/// `normalize_mt` became the fifth in #1217. #1205 had left it out on the
+/// grounds that a circular reference wraps, so position 1 HAS a valid 5\'
+/// neighbour (the last base) and the correct output there is a wraparound
+/// description rather than this `delins` rewrite. That does not hold: #129
+/// established, and `issue_129_mt_circular_wraparound.rs` pins, that ferro
+/// **rejects** `m.<high>_<low>ins`. The spec's reversed-range exception
+/// (`deletion.md:17`, SVD-WG006) is granted to `del`/`dup` — `delins` inherits
+/// it by composition — while `insertion.md` is silent, so the general 5\'→3\'
+/// rule applies, which is also how mutalyzer and biocommons read it. The
+/// wraparound `ins` is therefore not an available answer and the
+/// single-position `delins`, which needs no reversed range at all, is. (#129
+/// separately disabled 3\'-rule shifting across the origin, so coming to rest
+/// at the terminus is already the intended behavior; only its rendering was
+/// wrong.)
+///
+/// `o.` stays out: `normalize_core` returns circular `o.` variants unchanged, so
+/// no shuffle runs and nothing can saturate. A genuine circular normalizer is
+/// #951.
 ///
 /// The genomic caller passes a **window** rather than the whole contig, so it
 /// must say which of its ends are real — see [`SequenceEnds`].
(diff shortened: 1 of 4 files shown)


nh13:
Follow-up on the evidence note in the description: the backup reference mirror came back online after this merged, so the reference-gated test has now actually been run rather than reasoned about.

```
FERRO_MANIFEST=<prepared-reference>/manifest.json cargo nextest run --features dev -E 'test(issue_1217)'
    12 tests run: 12 passed, 8864 skipped
    PASS real_rcrs_three_prime_terminal_insertion_is_clamped
    PASS real_rcrs_five_prime_terminal_insertion_is_clamped
```

Both executed against the real rCRS (7.1s each, so they loaded the manifest rather than taking the skip path). The three expected strings hold exactly as derived from the issue:

- `NC_012920.1:m.16568_16569insGC` → `NC_012920.1:m.16569delinsGCG`
- `NC_012920.1:g.16568_16569insGC` → `NC_012920.1:m.16569delinsGCG` (the #487 coercion path)
- `NC_012920.1:m.1_2insCG` (5') → `NC_012920.1:m.1delinsGCG`

The description said these expectations were derived from the rCRS terminal bases quoted in the issue rather than from an observed run; they are now observed.

PR #1228, opened 2026-07-27
refactor(dev): make the spec generators binaries so cargo locates them

nh13:
The two spec generators were examples, so tests had to locate the built artifact themselves. Locating it means reconstructing the cargo invocation actually in flight — profile, target directory, features — and every coordinate missed is a silent wrong-path bug. Two shipped in a single day:

- **#1224** — the build ignored the profile, so `--release` looked in `target/release/examples/` for something written to `target/debug/`. The loud half was one failing test. The quiet half was a *sibling test passing against a stale release binary* — precisely the staleness the machinery existed to prevent.
- **#1225** — the build then ignored the target directory, so under `cargo llvm-cov` (which builds into `target/llvm-cov-target/`) all four tests failed. `main` stayed red for **15 commits**, because `Coverage` runs only after a merge and gates nothing.

Same defect, different axes — and the axes are not enumerable. `--features`, `--target`, `CARGO_BUILD_*` and `RUSTFLAGS` are all still out there.

## Cargo already solves this, for binaries

`CARGO_BIN_EXE_<name>` is defined for every `[[bin]]` target of the package under test, resolved by cargo for the invocation actually running, and cargo guarantees the binary is built and current before integration tests execute.

So the generators become `[[bin]]` targets and the whole derivation disappears — with it the `Once`, the nested `cargo build`, the path arithmetic, and the `FERRO_PREBUILT_EXAMPLES` escape hatch that existed only because CI's sharded jobs run from a nextest archive with no cargo available. The repo already runs 26 sites on `env!("CARGO_BIN_EXE_ferro")`, so this joins a path that is already paved and already handled by CI's `--extract-to .`.

A **second copy** of the same hand-derived path lived in `issue_1046_generator_gitdir_leak.rs`. It is gone too, along with its manual build step.

## It is also faster, which is what the original machinery was for

#1219 introduced the nested build because `cargo run --example` serialised this module's tests behind cargo's `target/` file lock at ~25 s each. With no nested cargo there is no second lock acquisition at all:

| | before | after |
|---|---|---|
| `spec_generator_preconditions` | ~20–37 s | **~0.7 s** |
| full `--features dev` suite | ~37 s | **~20 s** |

## Cost, stated plainly

**Three dev-dependencies become optional dependencies** — `pulldown-cmark`, `anyhow`, `serde_yaml` — because binaries, unlike examples, cannot use dev-dependencies. (I originally expected only `pulldown-cmark`; the other two surfaced during the build.) All three are reached solely through the non-default `dev` feature, so an ordinary consumer resolves none of them. `anyhow` and `serde_yaml` **remain** in `[dev-dependencies]` too, because other targets use them and `tests/it` still compiles without `--features dev`.

**`spec_generator_preconditions` is now `#[cfg(feature = "dev")]`.** `CARGO_BIN_EXE_*` exists only for targets cargo builds, and both generators are `required-features = ["dev"]`; without the gate a plain `cargo test` fails to compile the entire `it` target rather than skipping these tests.

**`autoexamples = false`.** With auto-discovery on, cargo registered each generator *twice* — once as a bin, once as an example — compiling both copies of a ~2000-line target. `cargo metadata` confirmed it directly. The five previously-implicit examples are now declared explicitly alongside the six that already were.

## Call sites

Every `--example generate_spec_*` becomes `--bin`: CI, the pre-push hooks, the nightly and external-validation workflows, the on-demand fixture regenerator, `CONTRIBUTING.md`, `CLAUDE.md`, and the generators' own docs. The existing hook/CI parity test asserts the new form, so hooks, CI and tests cannot drift apart.

## Validation

| Check | Result |
|---|---|
| `cargo nextest run --features dev` | **8786 passed**, 0 failed, 145 skipped |
| `cargo check --tests` (no `dev`) | clean — the gate works |
| alternate `--target-dir` (the llvm-cov shape) | **7 passed** — #1225's reproduction |
| `--release` | **7 passed** — #1224's reproduction |
| `cargo fmt --check`, `clippy --all-features --all-targets -D warnings` | clean |
| `cargo run --features dev --bin generate_spec_{fixture,enumeration}` | both exit 0 |

Both original reproductions pass structurally now rather than by patch, which is the point: there is no longer a path for them to get wrong.

## Summary by CodeRabbit

* **Improvements**
  * Spec fixture and enumeration generators now run consistently as dedicated binaries (instead of examples).
  * CI and local pre-push setup were updated to stage and execute the correct generator artifacts.
* **Tests**
  * Generator-dependent integration tests now run only when development mode is enabled, using Cargo-provided binary executables.
* **Documentation**
  * Contributor and generator guidance was updated to use the correct regeneration/validation commands for the HGVS v21.0 fixtures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -254,22 +254,23 @@ jobs:
         # validates the curated overrides file — the guard the old `--check`
         # step provided. Shipped to the shards as an artifact so they never
         # regenerate it (and so they need no spec submodule).
-        run: cargo run --features dev --example generate_spec_fixture
+        run: cargo run --features dev --bin generate_spec_fixture
       - name: Generate HGVS spec test enumeration
-        run: cargo run --features dev --example generate_spec_enumeration
+        run: cargo run --features dev --bin generate_spec_enumeration
       - name: Build test archive
         # Compiles every test binary and packs them with their run metadata.
         run: cargo nextest archive --features dev --archive-file nextest.tar.zst
-      - name: Stage the generator examples for the shards
-        # `spec_generator_preconditions` executes these two binaries. nextest
-        # does NOT build examples, so the shards cannot rely on finding them,
-        # and they must not shell out to cargo either (they run from the
-        # archive with no build state). Both were already built and run by the
-        # generate steps above, so ship them alongside the archive.
+      - name: Stage the generator binaries for the shards
+        # `spec_generator_preconditions` executes these two binaries. They are
+        # `required-features = ["dev"]` bin targets, and the shards run from the
+        # nextest archive with no build state, so they must not shell out to
+        # cargo. Both were already built and run by the generate steps above, so
+        # ship them alongside the archive. As `[[bin]]` targets they land in
+        # `target/debug/`, not `target/debug/examples/`.
         run: |
           mkdir -p staged-examples
-          cp target/debug/examples/generate_spec_fixture \
-             target/debug/examples/generate_spec_enumeration staged-examples/
+          cp target/debug/generate_spec_fixture \
+             target/debug/generate_spec_enumeration staged-examples/
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7.0.1
         with:
           name: nextest-archive
@@ -318,10 +319,10 @@ jobs:
       - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c  # v8.0.1
         with:
           name: generator-examples
-          path: target/debug/examples/
-      - name: Restore the executable bit on the staged examples
+          path: target/debug/
+      - name: Restore the executable bit on the staged generators
         # Artifact upload does not preserve the executable bit.
-        run: chmod +x target/debug/examples/generate_spec_fixture target/debug/examples/generate_spec_enumeration
+        run: chmod +x target/debug/generate_spec_fixture target/debug/generate_spec_enumeration
       - name: "Note: reference-aware tests are skipped without a manifest"
         # `tests/mutalyzer_normalize_tests.rs` and the biocommons reference-aware
         # suite both silently no-op when `FERRO_MANIFEST` is unset (PR CI does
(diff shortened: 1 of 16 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
Addressed the review, including the body-only findings that carry no thread (outside-diff-range and duplicate sections), which cannot be resolved via the API — listing them here explicitly.

**Fixed**

- `tests/it/main.rs` — gated `issue_1046_generator_gitdir_leak` with `#[cfg(feature = "dev")]`. It expands `CARGO_BIN_EXE_generate_spec_fixture` via `env!`, a compile-time lookup, so a plain `cargo test` failed to build the whole `it` target. Verified: `cargo test --no-run` without `--features dev` now compiles.
- `.github/workflows/ci.yml` — the generators are `[[bin]]` targets, so cargo emits them to `target/debug/`, not `target/debug/examples/`. Seven sites still used the old path: the staging `cp` (which would have failed the build job) and three `download-artifact` paths plus their `chmod` lines (which put the binaries where nothing looks for them). All corrected, step names updated.
- `examples/common/generated_artifact.rs` — the out-of-date hint printed `cargo run --features dev --example {example}`, a command that cannot run now that both callers are bins. Now `--bin {generator}`; parameter renamed to match.
- `tests/it/issue_1046_generator_gitdir_leak.rs` — removed the orphaned comment describing the `cargo build --target-dir` step this PR deletes, and renamed `example_binary` → `generator_binary` plus the residual `example` wording in the assertion and `expect` messages.
- `tests/it/spec_generator_preconditions.rs` — `built_example` → `generator_binary`, `run_example` → `run_generator`. Nothing is built here and no target is an example any more.
- `CLAUDE.md` — describes the generator as a binary rather than an example, and now tells the reader to replace the `<path>` placeholder.

**Not applied, with reason**

- *Last remaining nested `cargo run` in `tests/it/common/fixture_gen.rs`* — deliberately kept. Swapping it for `CARGO_BIN_EXE_*` would reintroduce the Critical bug fixed above: this helper is reached from `hgvs_spec_normalization_tests`, `idempotency_tests` and `spec_enumeration_tests`, none of which are `dev`-gated, and `env!` resolves at compile time, so an ungated module expanding it breaks a plain `cargo test`. The `generator` argument is also a runtime string, which `env!` cannot take. Resolving at run time is what keeps the helper callable from ungated modules. I have added a comment at the call site recording this, so the next pass does not re-flag it.
- The `--example` hints elsewhere in `examples/` (`extract_mutalyzer_genomic_windows`, `derive_tx_placements`, `report_conformance_reference_gaps`, `generate_tool_support_tables`, `extract_spec_enumeration_windows`) are correct as written — those targets remain examples; only the two generators became bins.

**Verification** — 8786/8786 tests, `clippy --all-features --all-targets -D warnings` clean, `fmt` clean, and `cargo test --no-run` compiles without `dev`. Also confirmed `autoexamples = false` accounts for every example: 13 files on disk = 11 `[[example]]` + the 2 now declared as `[[bin]]`, so nothing silently stopped building.

nh13:
@coderabbitai resume

nh13:
@coderabbitai review

Issue #1235, opened 2026-07-27
Normalizer: cis-allele members normalized independently, not from the resulting sequence (root cause of #1229–#1234)


msto:
> [!NOTE]
> Authorship: the content below was drafted by GitHub Copilot CLI (an AI coding
> agent) and filed via gh under @msto's account. It reflects the agent's
> analysis, not a statement authored by @msto.

## Summary (tracking issue)

A cluster of recently-filed normalization defects (#1229, #1230, #1231, #1232,
#1233, and the more severe overlap/corruption case) share one root cause:
**`Normalizer.normalize` processes each cis-allele member largely independently —
normalizing/typing/3′-shifting members in isolation — instead of re-deriving the
canonical edit set and partition from the variant's resulting sequence.** Because
the canonical form depends on the input encoding, the normalizer is not confluent
for multi-member alleles, and in one case it emits a malformed/incorrect allele.

This issue proposes tracking them together and fixing them with one
sequence-first canonicalization, rather than adding further per-symptom patches
(cf. the point fixes #1002, #1101, #1106 for earlier facets).

## The shared mechanism

Observed signatures, all on v0.11.0, all with the identical resulting sequence per
pair:

| symptom | example (→ ferro) | canonical | issue |
|---|---|---|---|
| `inv` member blocks coalescing | `[X>Y;inv]` stays split vs SNVs → `delins` | one `delins` | #1229 |
| `inv` over-recognized (unchanged interior) | `p_qinv` vs `[end>… ; end>…]` | separate subs | #1230 |
| `[dup;del]` not reduced | stays `[dup;del]` vs subs | subs (prioritisation) | #1231 |
| spanning `delins` over unchanged base | stays vs `[del;sub]` | split | #1232 |
| `[ins;del]` not reduced | stays vs subs | subs | #1233 |
| **deletion 3′-shift overlaps a sibling** | `[4_6del;8A>T]` → `[6_8del;8A>T]` | `5_8delinsT` | (overlap/corruption issue) |

In every row, feeding the fully-decomposed (per-position) encoding of the same
variant reaches ferro's own canonical, while a differently-partitioned encoding of
the *same* variant is left non-canonical — i.e. members are normalized in isolation
and the whole-allele partition is not re-derived.

## Why per-symptom patches keep leaving gaps

Prior fixes addressed individual facets:

- #1002 — coalesce a shift that lands *adjacent* to a sibling. (Does not bound a
[...]

Issue #1229, opened 2026-07-27
Normalizer: adjacent-edit coalescing skips inv members, leaving equivalent variants non-confluent


msto:
> [!NOTE]
> Authorship: the content below was drafted by GitHub Copilot CLI (an AI coding
> agent) and filed via gh under @msto's account. It reflects the agent's
> analysis, not a statement authored by @msto.

## What's wrong

`Normalizer.normalize` coalesces adjacent edits into a single `delins` when they
are written as substitutions, but **not** when one of the adjacent members is
written as an `inv`. As a result, two encodings of the **same** edit normalize to
two different strings — one of which is the non-canonical `sub + inv` form that
ferro never reduces.

Concretely (ref bases `CAC` at positions 4–6; the edit changes all three to
`AGT`):

```
TEMPLATE:g.[4C>A;5A>G;6C>T]   -> TEMPLATE:g.4_6delinsAGT      # 3 adjacent subs: coalesced
TEMPLATE:g.4_6delinsAGT       -> TEMPLATE:g.4_6delinsAGT      # delins: kept
TEMPLATE:g.[4C>A;5_6inv]      -> TEMPLATE:g.[4C>A;5_6inv]     # sub + adjacent inv: NOT coalesced
```

All three describe the identical resulting sequence. ferro clearly *has* a
canonical target for this edit — `g.4_6delinsAGT`, which it produces from the
substitution encoding and preserves when given the delins directly — but it fails
to reach that target when the middle/right of the run arrives as a 2-nt `inv`.

The 3-nt change is a genuine `delins`, not an inversion (`AGT` is not the reverse
complement of `CAC`, which is `GTG`), so per the spec the coalesced `delins` is the
correct canonical form:

> changes involving two or more **consecutive** nucleotides are described as
> deletion/insertion (delins) variants.
>
> — `assets/hgvs-nomenclature/docs/recommendations/DNA/delins.md:16`

Both outputs are individually idempotent, so this is **non-confluence**: two stable
fixed points for one variant. Any consumer that treats the normalized HGVS.g string
as a key (counting, joining, deduplicating) silently over-splits the variant,
because a library that stores the `inv` encoding and reads decomposed to
subs/`delins` never compare equal. There is no error or warning.

`EquivalenceChecker` **does** correctly report these as equivalent
(`SequenceMatch`), so it remains a sound way to compare variants; the defect is
specific to `normalize` not being confluent.

[...]

Issue #1135, opened 2026-07-22
fix(normalize): panic (divide by zero) merging a self-cancelling del+ins cis allele


nh13:
`ferro normalize` panics (exit 101) on a valid cis allele whose members cancel out:

```
$ ferro normalize --reference <ref> 'NM_004006.3:c.[179del;183_184insA]'
thread 'main' panicked at src/normalize/mod.rs:7421:36:
attempt to calculate the remainder with a divisor of zero
```

Both members normalize fine in isolation — `c.179del` → `c.183del`, `c.183_184insA` → `c.183dup` — so only the cis combination fails. `c.[179del;180_181insA]` panics identically. DMD `c.179`–`c.183` is a `AAAAA` homopolymer, so the deletion 3'-shifts onto the insertion anchor and the two merge.

Not a regression: v0.9.1 panics the same way (`mod.rs:7229:36`).

## Root cause

`build_naedit` (`src/normalize/merge.rs:1164`) picks the insertion branch on the anchor shape **before** testing whether anything is actually inserted:

```rust
let edit = if merged.start > merged.end {
    // insertion anchor — no emptiness check
    NaEdit::Insertion { sequence: InsertedSequence::Literal(Sequence::new(merged.alt)) }
} else if merged.alt.is_empty() {
    NaEdit::Deletion { .. }
} else {
    NaEdit::Delins { .. }
};
```

When a deletion and an adjacent insertion cancel, `merged.alt` is empty while the anchor still has insertion shape (`start == end + 1`), so this constructs `NaEdit::Insertion` with an **empty inserted sequence**. That is not a valid HGVS edit — it would render as `c.183_184ins` with nothing after `ins`.

The panic is the downstream symptom: `normalize_na_edit` rotates the inserted sequence when shifting an insertion through a repeat (`src/normalize/mod.rs:7421`):

```rust
let rotation = shift_amount % seq_bytes.len();   // seq_bytes.len() == 0
```

The protein-axis coalescer immediately below has precisely the guard the nucleotide path is missing (`merge.rs:1413`): *"Every member in the run is a deletion, so there is nothing to insert: emit a pure range deletion, not an (invalid) empty `delins`."*

## Expected

[...]

Issue #999, opened 2026-07-09
normalize: adjacent insertion+substitution in a cis allele is not coalesced into a single delins


nh13:
In a cis allele, an insertion that 3′-shifts to become directly adjacent (0-nt gap) to a substitution is left as two separate members instead of being coalesced into a single `delins`, as the spec requires for adjacent changes.

Synthetic reference `NC_000001.11`, positions 300–312 = `CATCCTCGCTCCT`:

| | |
|---|---|
| input | `NC_000001.11:g.[305_306insC;307G>T]` |
| ferro | `NC_000001.11:g.[306dup;307G>T]` |
| expected | `NC_000001.11:g.307delinsCT` |

The inserted `C` matches the reference at 306, so it shifts to `306dup`, landing directly 5′ of `307G>T`. With no intervening nucleotide the two are one contiguous change and must render as a single `delins`. Both spellings describe the same sequence, so this is a representational defect, not a wrong edit.

Suspected cause: the cross-member collapse runs before per-member 3′-shifting, so an adjacency created *by* the shift is never collapsed. Same root cause as the sibling idempotence issue #1000; both should be fixed by one change.

PR #1236, opened 2026-07-28
fix(normalize): clamp a cis member's 3'-shift at its siblings

### LINKED ISSUE #1234: Normalizer: deletion 3'-shift crosses a sibling substitution → overlapping members and silent sequence change
> [!NOTE]

## What's wrong

When a cis allele contains a **deletion** and a downstream **substitution**,
`Normalizer.normalize` 3′-shifts the deletion to its *standalone* most-3′ position
— **ignoring the sibling substitution** — so the shifted deletion crosses **into**
the substituted position. The result is a cis allele whose two members **overlap**,
which is malformed, and which ferro itself reads as a **different sequence** (the
substitution is silently swallowed).

Example (ref `ATGCACCAGT…`; delete 4–6, substitute at 8, with unchanged base 7
between them):

```
input:            TEMPLATE:g.[4_6del;8A>T]        # valid: del 4-6, sub at 8 (gap at 7)
ferro normalize:  TEMPLATE:g.[6_8del;8A>T]        # 6_8del and 8A>T BOTH claim position 8 -> overlap
```

The deletion alone normalizes to `6_8del` (its standalone 3′-most), and ferro emits
exactly that even with the substitution present — it does not clamp the shift at the
sibling.

### It is silent sequence corruption, confirmed by ferro's own EquivalenceChecker

```
EquivalenceChecker([6_8del;8A>T], 6_8del)       -> SequenceMatch    # output == just the deletion; sub lost
EquivalenceChecker([4_6del;8A>T], 5_8delinsT)   -> NotEquivalent    # true-equal forms judged inequivalent
```

- The intended variant is `…ATG|C|T|GT…` = `ATGCTGT…` (delete CAC at 4–6, keep C at
  7, A→T at 8).
- ferro's output `[6_8del;8A>T]` is read by ferro as plain `6_8del` = `ATGCAGT…`
  (positions 6,7,8 deleted; the substitution has nothing to act on). **Different
  sequence.**
- The correct normalization is `TEMPLATE:g.5_8delinsT` (identical to what
  `[5_7del;8A>T]` yields): shift the deletion only as far as `5_7` — the last
  position before the substituted base — after which the now-adjacent `del`+`sub`
  coalesce into a single `delins` (the #1000/#1002 machinery). ferro instead shifts
  one position too far, to `6_8`.

This is not merely a non-canonical (but valid) description like #1229–#1233: the
emitted allele is **malformed** (overlapping members) and encodes the **wrong
sequence**.

## Severity: high

- **Silent sequence corruption** — `normalize` returns a well-formed-looking string
  that (per ferro's own interpretation) denotes a *different* variant; the
  substitution is dropped. No error or warning.
- **Malformed output** — the two members overlap at position 8; a spec-strict
  consumer should reject `[6_8del;8A>T]` (a base cannot be both deleted and
  substituted).
- **Breaks equivalence too** — `EquivalenceChecker` inherits it: two encodings of
  the same variant are reported `NotEquivalent`, and the corrupt output is reported
  equal to a *different* variant (`6_8del`). So the usual fallback
  (`EquivalenceChecker`/`SequenceMatch`) does **not** rescue this case.
- **Long-standing** — reproduces identically on v0.9.1 and v0.11.0 (not a recent
  regression; see below).

## Reproduction (verified on v0.11.0, commit 16b98af; identical on v0.9.1)

Fully synthetic reference (contig `TEMPLATE`, invented sequence).

```python
import ferro_hgvs

      "protein_id": TID,

n = nz.normalize
print(n("TEMPLATE:g.[4_6del;8A>T]"))   # TEMPLATE:g.[6_8del;8A>T]   <- overlapping members
print(n("TEMPLATE:g.4_6del"))          # TEMPLATE:g.6_8del          (standalone 3'-most)
print(n("TEMPLATE:g.[5_7del;8A>T]"))   # TEMPLATE:g.5_8delinsT      (the correct canonical)

lvl = lambda a, b: ec.check(ferro_hgvs.parse(a), ferro_hgvs.parse(b)).level
print(lvl("TEMPLATE:g.[6_8del;8A>T]", "TEMPLATE:g.6_8del"))       # SequenceMatch  (sub swallowed)
print(lvl("TEMPLATE:g.[4_6del;8A>T]", "TEMPLATE:g.5_8delinsT"))   # NotEquivalent  (same variant!)
```

The per-member 3′-shift computes each member's most-3′ placement **independently of
its siblings**, so a deletion shifts to the position it would occupy alone, crossing
a downstream substitution. #1000 / PR #1002 added coalescing when a shift lands
*adjacent* to a sibling, but did not bound the shift itself: a shift that lands
*overlapping* a sibling is neither prevented nor resolved. The fix is to clamp each
member's 3′-shift at the nearest sibling boundary (here, stop the deletion at `5_7`),
after which the existing adjacency-coalescing yields `5_8delinsT`.

This is the same "members normalized independently" root cause as #1229–#1233, but
its consequence here is a malformed/incorrect allele rather than a merely
non-canonical one.

## Expected

```
TEMPLATE:g.[4_6del;8A>T]  ->  TEMPLATE:g.5_8delinsT
```

No normalized cis allele contains overlapping members, and `normalize` never
changes the resulting sequence.

## Related

- #1000 / PR #1002 (shift-created **adjacency** coalescing — this is the
  shift-created **overlap** case it does not cover)
- #1229, #1230, #1231, #1232, #1233 (independent-member normalization family)

nh13:
## What's wrong

A cis-allele member is 3'-shifted to its *standalone* most-3' position with no knowledge of its siblings, so a deletion with a downstream substitution shifts straight into the substituted base:

```
TEMPLATE:g.[4_6del;8A>T]  ->  TEMPLATE:g.[6_8del;8A>T]
```

Both members claim position 8. That is malformed — a base cannot be both deleted and substituted — and it silently denotes a **different sequence**. Confirmed with ferro's own `EquivalenceChecker`:

```
[6_8del;8A>T]  vs  6_8del        -> SequenceMatch    # the substitution is swallowed
[4_6del;8A>T]  vs  5_8delinsT    -> NotEquivalent    # true-equal forms judged inequivalent
[4_6del;8A>T]  vs  [5_7del;8A>T] -> NotEquivalent    # two spellings of one variant
```

That last row is not in the issue — two spellings of the identical variant compare inequivalent, so the usual `EquivalenceChecker` fallback does not rescue this case.

## The fix

A 3'-shift is a rotation, so every position between a member's pre-shift span and its shifted span describes the same sequence. Pulling an offending member back toward its pre-shift position is therefore always sequence-preserving, and stopping at the last position before the sibling is the most 3' placement that keeps the members disjoint. The clamped member is then *adjacent* to its sibling rather than overlapping it, and the existing merge pass coalesces the two into one delins:

```
TEMPLATE:g.[4_6del;8A>T]  ->  TEMPLATE:g.5_8delinsT
```

The clamp applies only where members can genuinely collide: same accession, same region, ascending order, and **both edits consuming the reference bases under their span**. Insertions and duplications add sequence at a junction without consuming a base, so they are excluded — treating an insertion gap as a claimed position would break the shift-created-adjacency collapse (#999) and the self-cancelling del+ins case (#1135).

## Spec basis

The spec is silent on how the 3' rule interacts with siblings — `general.md:41` states it per description with no allele-awareness, and the only text that ever addressed cross-member shifting, `SVD-WG010:14-16`, was **rejected**. What is not in doubt: a description must not change the sequence it denotes, and `general.md:58` forbids removing part of a reference sequence and replacing it with part of the same sequence.

## Re-blessed test

`normalize_tests::…::test_compact_allele_notation` had pinned the corrupt output — the same defect on the `c.` axis, where `c.[4del;9A>G]` emitted `[9del;9A>G]`. Its assertion (`contains("del") && contains("A>G")`) was loose enough to accept the overlapping form; it now asserts the correct `c.8_9delinsG`.

## Verification

- 8782/8782 tests pass, including a second run under `FERRO_ASSERT_IDEMPOTENT=1`.
- `cargo fmt` and `cargo clippy --features dev --all-targets` clean.
- New tests cover the reported case, the already-clamped spelling, a lone deletion still shifting fully, a distant sibling not clamping, the no-overlapping-members invariant, and sequence preservation via `EquivalenceChecker`.

## Scope

This is the corruption fix only. The five non-confluence defects (#1229–#1233) and the root-cause canonicalization (#1235) follow separately.

Closes #1234

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cis allele normalization when shifted deletions meet adjacent sibling variants.
  * Prevented overlapping or incorrectly ordered normalized members.
  * Preserved resulting sequence equivalence during normalization.
  * Correctly retained uncertainty markers while applying normalization.
* **Tests**
  * Added coverage for sibling-clamped shifts, deletion coalescing, distant siblings, and sequence-preservation scenarios.
  * Strengthened canonical normalization assertions for compact allele notation.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -10,6 +10,7 @@ use crate::hgvs::variant::{
     Accession, AllelePhase, CdsVariant, GenomeVariant, HgvsVariant, LocEdit, MtVariant, RnaVariant,
     TxVariant,
 };
+use crate::normalize::verify;
 use crate::reference::ReferenceProvider;
 
 /// Coordinate-system region used as the merge-eligibility key.
@@ -1580,17 +1581,31 @@ pub(crate) fn coalesce_protein_adjacent_changes(
 /// supplies the pull-back floor; `after` is the shifted list, mutated in place.
 /// Members whose spans cannot be read (uncertain, offset, special positions)
 /// are left alone.
-pub(crate) fn clamp_shifted_members(
+pub(crate) fn clamp_shifted_members<P: ReferenceProvider>(
     before: &[HgvsVariant],
     after: &mut [HgvsVariant],
     phase: AllelePhase,
+    uncertain: bool,
+    provider: &P,
 ) {
     if phase != AllelePhase::Cis || after.len() < 2 || before.len() != after.len() {
+        return;
+    }
+    // An uncertain allele — `g.[(…)]` — asserts only that the members are in
+    // cis, not where they sit, so rewriting one member's position drops the
+    // predicted marker and states something the input did not. Every
+    // neighbouring transform in `normalize_allele` gates on this; the clamp
+    // must too. The flag is always false for protein but genuinely reachable
+    // for DNA.
+    if uncertain {
         return;
     }
     let Some(kind) = clamp_kind_of(&after[0]) else {
         return;
     };
+    // Record the shifted positions so an unverifiable clamp can be undone.
+    let unclamped = after.to_vec();
+    let mut clamped: Vec<usize> = Vec::new();
     // Walk 3'→5' so a member pulled back cannot be re-overlapped by the one
     // before it in the same pass.
     for i in (0..after.len() - 1).rev() {
@@ -1638,7 +1653,43 @@ pub(crate) fn clamp_shifted_members(
             continue;
         }
         shift_member_left(&mut after[i], pull_back);
+        clamped.push(i);
     }
+
+    if clamped.is_empty() {
+        return;
+    }
+
+    // Everything above is positional reasoning, and positional reasoning is
+    // exactly what cannot decide this. A 3'-shift is a rotation, so pulling a
+    // member back toward its pre-shift position preserves the sequence *for
+    // that member alone*; it stops doing so once a sibling edits a base inside
+    // the tract the member rotated through, and no property of the edit kinds
+    // or their span lengths tells the two apart. Fuzzing an earlier structural
+    // gate over 222,300 three-member alleles found 32 cases where it passed
+    // every condition and still changed the sequence silently. So the pull-back
+    // is proposed here, and the reference decides.
+    if verify::same_resulting_sequence(provider, before, after) == Some(true) {
+        return;
+    }
+
+    // Something in the batch changed the sequence. Try to keep the rest: revert
+    // one member at a time, newest-clamped first, and re-verify. A single
+    // culprit is the realistic shape, and stopping there keeps this bounded.
+    for &candidate in &clamped {
+        let mut trial = after.to_vec();
+        trial[candidate] = unclamped[candidate].clone();
+        if verify::same_resulting_sequence(provider, before, &trial) == Some(true) {
+            after.clone_from_slice(&trial);
+            return;
+        }
+    }
+
+    // Unverifiable, or more than one member is implicated. Leave the members
+    // where per-member normalization put them. That output may still overlap,
+    // which is a shape a consumer can detect and reject — strictly better than
+    // a clean-looking description that denotes different bases.
+    after.clone_from_slice(&unclamped);
 }
 
 /// Whether an edit consumes the reference bases under its span.
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -1553,6 +1553,176 @@ pub(crate) fn coalesce_protein_adjacent_changes(
     }
 }
 
+// ----------------------------------------------------------------------------
+// Sibling-clamped 3'-shift (#1234)
+// ----------------------------------------------------------------------------
+
+/// Pull back any cis member whose 3'-shift ran into a sibling.
+///
+/// The 3' rule (`general.md:41`) is stated per description with no
+/// allele-awareness, and the only text that ever addressed cross-member
+/// shifting — SVD-WG010 — was rejected, so the spec is silent here. Left
+/// unbounded, a deletion shifts to its *standalone* most-3' position and can
+/// land on top of a downstream sibling: the emitted allele has two members
+/// claiming one base, which is malformed, and it denotes a different sequence
+/// (ferro reads `[6_8del;8A>T]` as a bare `6_8del`, swallowing the
+/// substitution).
+///
+/// A 3'-shift is a rotation, so every position between a member's pre-shift
+/// span and its shifted span describes the same sequence. Pulling the member
+/// back toward its pre-shift position is therefore always sequence-preserving,
+/// and stopping at the last position before the sibling is the most 3' choice
+/// that keeps the members disjoint. Once clamped the member is adjacent to its
+/// sibling rather than overlapping it, and the caller's next merge pass
+/// coalesces the two into one delins (`delins.md:16`).
+///
+/// `before` is the member list as it entered per-member normalization and
+/// supplies the pull-back floor; `after` is the shifted list, mutated in place.
+/// Members whose spans cannot be read (uncertain, offset, special positions)
+/// are left alone.
+pub(crate) fn clamp_shifted_members(
+    before: &[HgvsVariant],
+    after: &mut [HgvsVariant],
+    phase: AllelePhase,
+) {
+    if phase != AllelePhase::Cis || after.len() < 2 || before.len() != after.len() {
+        return;
+    }
+    let Some(kind) = clamp_kind_of(&after[0]) else {
+        return;
+    };
+    // Walk 3'→5' so a member pulled back cannot be re-overlapped by the one
+    // before it in the same pass.
+    for i in (0..after.len() - 1).rev() {
+        let (
+            Some((accession, region, start, end, edit)),
+            Some((next_accession, next_region, next_start, _, next_edit)),
+        ) = (
+            cis_axis_parts(&after[i], kind),
+            cis_axis_parts(&after[i + 1], kind),
+        )
+        else {
+            continue;
+        };
+        // Only edits that actually consume reference bases can collide. An
+        // insertion sits *between* two bases and claims none of them, so a
+        // deletion ending at `p` and an insertion in the gap `p_p+1` are
+        // adjacent, not overlapping — clamping there would break the
+        // shift-created-adjacency collapse (#999) and the self-cancelling
+        // del+ins case (#1135). A duplication is an insertion of a copy, so it
+        // is excluded on the same grounds.
+        if !claims_reference_bases(edit) || !claims_reference_bases(next_edit) {
+            continue;
+        }
+        // Positions on different accessions, or in different regions of one
+        // axis (`c.-1` vs `c.1`), share no coordinate line and cannot overlap.
+        if accession != next_accession || region != next_region {
+            continue;
+        }
+        if end < next_start {
+            continue; // already disjoint
+        }
+        // The clamp reads the list as 5'→3'. An unsorted pair is not a shift
+        // collision — leave it to the sort and merge passes.
+        if start > next_start {
+            continue;
+        }
+        let overlap = end - next_start + 1;
+        // Never pull back past where the member started: those positions were
+        // not reachable by a 3'-shift and are not equivalent.
+        let Some((_, _, floor, _, _)) = cis_axis_parts(&before[i], kind) else {
+            continue;
+        };
+        let pull_back = overlap.min(start - floor);
+        if pull_back <= 0 {
+            continue;
+        }
+        shift_member_left(&mut after[i], pull_back);
+    }
+}
+
+/// Whether an edit consumes the reference bases under its span.
+///
+/// Substitutions, deletions, delins and inversions replace or remove the bases
+/// they cover, so two of them over one position is a genuine conflict.
+/// Insertions and duplications add sequence at a junction without consuming a
+/// base, so they never conflict positionally with a neighbour.
+fn claims_reference_bases(edit: &NaEdit) -> bool {
+    matches!(
+        edit,
+        NaEdit::Substitution { .. }
+            | NaEdit::Deletion { .. }
+            | NaEdit::Delins { .. }
+            | NaEdit::Inversion { .. }
+    )
+}
+
+/// The `CisKind` for the clamp pass, or `None` for a kind it does not serve.
+fn clamp_kind_of(variant: &HgvsVariant) -> Option<CisKind> {
+    match variant {
+        HgvsVariant::Genome(_) => Some(CisKind::Genome),
+        HgvsVariant::Mt(_) => Some(CisKind::Mt),
+        HgvsVariant::Cds(_) => Some(CisKind::Cds),
+        HgvsVariant::Tx(_) => Some(CisKind::Tx),
+        HgvsVariant::Rna(_) => Some(CisKind::Rna),
+        _ => None,
+    }
+}
+
+/// Move a member's span `delta` positions 5', leaving its edit untouched.
+fn shift_member_left(variant: &mut HgvsVariant, delta: i64) {
+    match variant {
+        HgvsVariant::Genome(g) => shift_genome_interval(&mut g.loc_edit.location, delta),
+        HgvsVariant::Mt(m) => shift_genome_interval(&mut m.loc_edit.location, delta),
+        HgvsVariant::Cds(c) => shift_cds_interval(&mut c.loc_edit.location, delta),
+        HgvsVariant::Tx(t) => shift_tx_interval(&mut t.loc_edit.location, delta),
+        HgvsVariant::Rna(r) => shift_rna_interval(&mut r.loc_edit.location, delta),
+        _ => {}
+    }
+}
+
+/// Apply `f` to both certain endpoints of an interval, leaving uncertain or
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai resume

nh13:
@coderabbitai review

nh13:
Converted to draft. Independent review — built this branch and `origin/main` behind identical probe binaries, with a hand-written variant applier as an oracle over ~150,000 generated cis alleles — found that the fix is real on the happy path (62 fixes / 0 breakages over 40,000 two-member alleles) but introduces a new corruption class and rests on an invariant that mostly does not hold.

## The blocker: a new silent corruption

The pull-back floor is `before[i]`'s start, and nothing checks that `before[i]` and `after[i]` are the *same edit*. `normalize_core` can change a member's **kind** — an insertion near the 3' terminus becomes a terminal `delins` — and `cis_axis_parts` returns an insertion's *raw* left endpoint, which has no rotational relationship to the resulting delins. The clamp then pulls it onto positions describing a different sequence:

```
reference  GGGAAAAAACCT
input      TEMPLATE:g.[10_11insCTT;12T>A]
  main   → g.[12T>A;12delinsTTCT]   + warning OverlapConflict{location:"12"}
  branch → g.11_12delinsTTCTA       + NO warnings
  input sequence  GGGAAAAAACCTTCA
  branch sequence GGGAAAAAACTTCTA   ← different
```

This is a **regression in kind**: main emitted a malformed-but-*warned* overlap a consumer can reject; this emits a well-formed, unwarned, wrong description — exactly the failure mode the PR sets out to prevent. Targeted fuzz of 20,000 (insertion in one of the last three gaps + downstream sibling): **47 corruptions, 0 improvements**. The two regimes are disjoint from the happy path, so the change is net-positive in one and purely harmful in the other.

## The safety argument is false as written

"A 3'-shift is a rotation, so every intermediate position is equivalent" holds only when the move *was* a rotation of the same edit. Measured over 20,000 single-member variants, pulling back by every legal `k`:

| before → after | moved | non-equivalent pull-backs |
|---|---|---|
| `del` → `del` | 1238 | **0** |
| `delins` → `del` | 177 | 79 |
| `delins` → `sub` | 111 | 87 |
| `delins` → `delins` | 162 | 162 |
| `ins` → `delins` | 5 | 5 |

The invariant holds for exactly one shape. The reason the `delins` classes are not live today is accidental: `before` is `merged_split`, not the pre-shift list the doc claims, and for `delins` members it is *already* the shifted form, so `pull_back == 0`. Note both siblings of this function (`merge_consecutive_edits`, `collapse_overlapping_cis_edits`) take a provider and verify against the reference; this one makes a sequence-preservation decision without reading a base.

## My sequence-preservation test is vacuous

`normalization_preserves_the_resulting_sequence` **passes against `origin/main`** — verified by compiling its exact body there. `check(input, normalize(input))` normalizes both sides, so it returns `NormalizedMatch` whenever `normalize` is idempotent, regardless of whether the sequence changed. It is an idempotency test wearing a sequence-preservation costume, and it does not catch the corruption above either. I leaned on it in the PR body; that was wrong.

The other two new tests are load-bearing and good: `normalized_cis_members_never_overlap` genuinely fails on main (`8 > 8`), and the exact-string assertion is real.

## The gate is wrong in both directions

- **`Repeat` is excluded but is the dominant real-world spelling.** `GGGAAAAAAGGG`, `g.[4_6del;8A>T]` → `g.[4_9A[3];8A>T]`, identical to main and unwarned — the repeat span swallows the substitution. Of main's 1810 overlapping outputs in a 40,000-allele fuzz, this fixes 62 (3.4%); **1710 of the remaining 1748 involve a `Repeat` member**.
- **The sibling's spelling decides whether the fix applies:** `8>T` (SubstitutionNoRef), `8=`, `8_9dup`, `8delN[1]` all defeat it. `Conversion`, `MultiRepeat`, `DupIns` are omitted too.
- **Excluding `Duplication` contradicts the spec line this PR cites.** `general.md:58`'s only example is `c.[762_768del;767_774dup]` — precisely the del+dup pair the gate declines to clamp. ferro's own parser agrees with the spec: the branch emits `g.[17del;17dup]`, which ferro then *cannot re-parse* ("Self-cancelling allele … overlapping reference positions").

## Also worth fixing

- Not gated on `allele.uncertain`, unlike every other transform in `normalize_allele`; it widens an existing loss of uncertainty markers.
- When the sibling is a merge barrier (e.g. `inv`) the clamped member survives deliberately **not** at its most-3' position, with no warning or info — a knowing deviation from `general.md:41` that a consumer cannot detect.
- Cross-region floor arithmetic on `c.`/`r.`/`n.` subtracts region-local coordinates across a UTR/CDS boundary; currently unreachable only because another module clamps first.
- `clamp_kind_of` is the third copy of the same `HgvsVariant → CisKind` match; no unit tests for the new helpers.

## Verified correct (so it is not re-litigated)

Off-by-one arithmetic; `(0..len-1).rev()` really does re-check the preceding pair; no oscillation and idempotency holds (though note an idempotency oracle *cannot* catch the corruption above, since those outputs are idempotent); uncertain/special/offset endpoints and insertion/dup raw-endpoint conventions provably never reach the mutator; stated reference bases are stripped before the clamp; mito wraparound and non-cis phases untouched; the re-blessed `test_compact_allele_notation` expectation `c.8_9delinsG` is genuinely correct and main really did emit the overlapping `c.[9A>G;9del]`; **all spec citations accurate, no misquotation** — `general.md:41`, `general.md:58` (verbatim), `delins.md:16`, and SVD-WG010 is indeed `Status: rejected` and is indeed the only text addressing cross-member shift interaction.

## Plan

Require same edit kind **and** span length before trusting the floor — or better, pass the provider and verify the clamped member still denotes the same sequence, matching what the sibling functions already do. Replace the vacuous test with one that applies both descriptions to the reference and compares bases. Widen the gate to `Repeat`/`SubstitutionNoRef`/`Identity`/`NPaddedDeletion`/`Conversion`/`MultiRepeat`, and include `Duplication` per `general.md:58`. Add the `!allele.uncertain` gate and an info when the 3' rule is intentionally not applied.

nh13:
@coderabbitai review

nh13:
@coderabbitai pause

nh13:
Pushed `e4a9a86`, adding a second commit that responds to the adversarial review. Two changes, plus one finding that redirects part of the critique.

## 1. The clamp now gates on `allele.uncertain` — a blocking defect, now fixed

An uncertain allele asserts only that its members are in cis, not where they sit, so rewriting a member's position drops the predicted marker:

```
g.[(4_6del;8A>T)]     before -> g.5_8delinsT          the parentheses vanish
                       after -> g.[(6_8del;8A>T)]     marker preserved
```

Every neighbouring transform in `normalize_allele` already gates on this flag. This one did not. The original comment claimed the flag is always false here — true for protein, false for a DNA `g.[( … )]`.

## 2. The pull-back is now proposed, and the reference decides

The clamp reasoned purely positionally. A 3'-shift is a rotation, so pulling a member back toward its pre-shift position preserves the sequence *for that member alone*, and stops doing so once a sibling edits a base inside the tract the member rotated through. No property of the edit kinds or their span lengths separates the two cases — which is exactly what the fuzzing showed when a structural `discriminant` + span-length gate let 32 sequence-changing cases through.

`normalize::verify` answers one question by construction: applied to their shared reference, do these two member lists produce the same bases? It declines — "cannot tell" — rather than guessing, for a member that will not project to SPDI, members on different accessions, an over-wide window, a reference the provider will not serve, or a member list whose own edits overlap (which has no single resulting sequence at all). A decline is treated as failure to verify, never as success.

On failure the clamp reverts one member at a time, newest-clamped first, and re-verifies — the reviewer's per-member skip, in the single-culprit shape that actually occurs. If nothing verifies it abandons the clamp entirely and leaves members where per-member normalization put them. That output may still overlap, which a consumer can detect and reject, and that is deliberately preferred to a clean-looking description denoting different bases.

It does **not** reuse `equivalence::checker`'s applier, for two reasons worth flagging: that one compares whole variants rather than member lists, and it indexes out of bounds on overlapping triples — the #1244 panic — which is precisely the input the clamp starts from. Folding the two into one applier is worth doing once #1244 lands.

## 3. The headline counterexample is not this PR's bug — it is #1254

This is the part that changes the reading of the fuzz results. Instrumenting the clamp on the reviewer's own counterexample:

```
reference   TAATATATATAATATATATT
input       g.[3_4del;9del;13T>A]

CLAMP fired on 2 members, clamped=[0], verdict=Some(true)
  before = ["g.9_11del", "g.13T>A"]
  after  = ["g.10_12del", "g.13T>A"]
```

The clamp is handed `[g.9_11del; g.13T>A]`, which **is already not equivalent to the input** — applying the input gives `TAATATTAAAATATATT`, applying that member list gives `TAATATATAAATATATT`. The verification correctly reports the clamp preserving the sequence from there, because it does.

Reducing it isolates the real defect, which needs no clamp at all and reproduces on `origin/main`:

```
input     g.[3_4del;9del]
main   -> g.12_14del
applied   TAATATTAATATATATT  vs  TAATATATATATATATT     ← different, and silent
```

Two deletions separated by four unchanged bases are merged into one deletion that denotes different bases. There is no sibling to collide with, so nothing is overlapping and nothing warns. Filed as **#1254**.

That does not rescue the original design — the criticism that the clamp converts a *detectable* error (overlapping members) into an *undetectable* one still stands, and is why the verification is here. But the corruption itself originates upstream of the clamp, so part of the "32/32 regressions survive the gate" count is #1254 arriving pre-corrupted rather than damage this PR does. Whether an allele-level guard against the *original* input belongs here or in #1240 (`fix/runtime-sequence-preservation-guard`, which already implements that shape) is a scoping call for review — I have not widened this PR to it.

## Not done, deliberately

Widening `claims_reference_bases` to `Repeat` and friends is still open. The reviewer's guidance was to do it only after the floor is verified, because widening first increases corruption exposure; the floor is verified now, so it is unblocked, but it is a separate change with its own fuzzing needs.

## Verification

| gate | result |
| --- | --- |
| mock suite | **8792/8792** |
| `FERRO_ASSERT_IDEMPOTENT=1` | **8792/8792** |
| `cargo fmt --all` | clean |

Eight unit tests cover `verify` directly, including the discrimination the whole design rests on: a deletion rotating *within* the `(AT)` tract compares equal, and the same deletion placed one base past the tract boundary compares unequal — the distinction no positional rule catches. Plus the uncertain-marker regression and a control that the certain spelling still clamps.

nh13:
@coderabbitai full review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Closing in favour of #1255 + #1256, which together supersede this PR. The coverage here is kept — all eight tests from `tests/it/issue_1234_sibling_clamped_shift.rs` are ported verbatim (with updated prose) in #1256, which carries `Closes #1234`.

The reasoning, measured rather than asserted.

**#1234 and #1254 are one defect seen from two sides.** A cis member is 3'-shifted to its standalone most-3' position with no sibling awareness. When it moves over a base a sibling edits, either it lands *on top of* the sibling — the overlapping shape this PR fixes — or it clears the sibling entirely, which is #1254's silent shape. #1255 prevents the crossing itself, one step upstream, which resolves both. By the time this PR's clamp runs there is nothing left to clamp.

**Three checks before concluding that:**

1. All eight tests in this PR pass against #1255 with this PR's clamp entirely absent.

2. Overlapping normalized outputs across 143,360 two-member cis alleles, both shuffle directions:

   | --- | --- |
   | `main` | 2,578 |
   | this PR alone | 1,444 (−44%) |
   | **#1255 alone** | **526 (−80%)** |
   | #1255 + this PR stacked | **526 — unchanged** |

   The stacked build was actually constructed and is green (8,810 tests), so the two are compatible — this clamp simply changes no output once #1255 is in.

3. Both PRs independently re-blessed `test_compact_allele_notation` to the same `c.8_9delinsG`, reached by different mechanisms.

So this is not a merge-order question: keeping both would ship ~520 lines, including the 303-line `verify.rs`, that provably change no output.

**On the adversarial review of this PR** — the criticism that the positional gate converts detectable failures into undetectable ones stands, and part of the "32/32 regressions survive the gate" count is now explained: those inputs reached this clamp already corrupted by #1254, upstream. That is consistent with `verify::same_resulting_sequence` reporting the clamp itself preserving the sequence. The verification added in e4a9a86 was the right response at the time; it is simply not needed once the corruption is prevented rather than detected.

**One thing neither PR fixes**, recorded so it is not lost: a deletion canonicalised into a whole-tract repeat that overlaps its sibling — `g.[1_2del;4del]` → `g.[1_9T[7];9del]`. That is the entire 526 residual above. `NaEdit::Repeat` is absent from `claims_reference_bases`, and that predicate is byte-identical in both PRs, so neither clamp sees it. Pre-existing on `main`, distinct from both #1234 and #1254.

Reopen if the measurement looks wrong — the sweep is `tests/it/issue_1254_sibling_crossing_shift.rs` in #1255, and the overlap probe is straightforward to reconstruct from it.

PR #1242, opened 2026-07-28
feat(normalize): add a tsv output format reporting which variants changed

nh13:
> **Stacked on #1240** (`fix/runtime-sequence-preservation-guard`). The base branch will retarget to `main` as the stack merges — review only the top commit here.

## Problem

`ferro normalize` already takes a list of variants (`--input`), but there is no good way to answer the question people actually bring to it: *given this list, which variants does ferro normalize to something different, and to what?* Today that means eyeballing the arrow-separated `text` output (where a changed variant is `in -> out` and an unchanged one is just `out`, so "did it change?" is inferred from the presence of an arrow) or writing a JSON consumer.

## The flag

`ferro normalize --format tsv` emits a header-bearing table, one row per input variant:

```
input	normalized	changed	status	detail
NM_000088.3:c.459A>G	NM_000088.3:c.459A>G	false	ok
NM_000088.3:c.459delA	NM_000088.3:c.459del	true	ok
not-a-variant		false	parse_error	Parse error at position 0: Failed to parse accession: ...
```

- `input` — the variant exactly as read.
- `normalized` — the normalized string, empty when normalization did not happen.
- `changed` — whether `normalized` differs from `input`. Because it compares against the input *as read*, an error-mode preprocessing correction counts as a change, which is what a caller asking "did ferro touch this?" wants.
- `status` — `ok`, or the phase that failed: `preprocess_error` (refused by the error-mode preprocessor before parsing), `parse_error` (rejected by the HGVS grammar), `normalize_error` (parsed, but could not be normalized). The three are distinguished because they tell a caller different things — a `normalize_error` usually means the reference lacks the accession, not that the string is malformed.
- `detail` — the error message when there is one, otherwise empty.

A one-line summary goes to **stderr**, never stdout, so the output stream stays a table whose every line is a row:

```
summary: total=3 changed=1 unchanged=1 failed=1
```

`--output` is honored (the table goes to the file, the summary still to stderr), and the single-positional-`variant` form works too, so the flag is never silently ignored.

Failures are **rows, not aborts**: a bad line does not cost the caller the rows for the lines after it. Exit-code behavior is unchanged — as before, any failed variant still fails the run with `N variant(s) failed to normalize`.

## Sanitization guarantee

Every field has tab, newline, and carriage return replaced with a space. An error message is arbitrary text (it frequently quotes the offending input), so emitting it raw could split one logical row into several physical ones and shift the columns of everything after it. Sanitizing unconditionally makes the "one physical line, five fields" shape a property of the format rather than a property of well-behaved inputs. There is a unit test for exactly that, plus an integration test that drives a tab through the CLI.

## Existing formats are untouched

`tsv` gets its own per-variant path rather than an extra arm inside the existing renderer, so `text` and `json` cannot drift as a side effect. `text_and_json_output_is_untouched` pins both byte streams (and that `text` is still the default) for the same input.

## Tests

New unit tests in `src/cli/format.rs` (row/summary rendering, sanitization, all three failure tokens) and a new integration module `tests/it/cli_normalize_tsv.rs` covering: header present (including on empty input), unchanged, changed, malformed-becomes-a-row-and-continues, the three failure phases, tab sanitization end-to-end, summary on stderr only, `--output`, the single positional variant, and text/json invariance.

Suite: **8797 → 8815 passing** (18 new), `cargo fmt --all` clean, `cargo clippy --all-features --all-targets -- -D warnings` clean.

## Summary by CodeRabbit

* **New Features**
  * Added `tsv` output support to the `normalize` command.
  * TSV output includes a header and one row per input line, including failures.
  * Added per-row status, line details, normalized values, and change indicators.
  * Added an end-of-run summary reporting total, changed, unchanged, and failed rows.
  * Supports TSV output for positional input, files, standard input, and output redirection.

* **Bug Fixes**
  * TSV fields are sanitized to prevent malformed rows or unintended line breaks.
  * Output remains deterministic when processing inputs in parallel.

### DIFF
--- a/src/bin/ferro.rs
+++ b/src/bin/ferro.rs
@@ -7,8 +7,10 @@
 
 use clap::{Parser, Subcommand};
 use ferro_hgvs::cli::{
-    output_error as cli_output_error, output_error_with_context as cli_output_error_with_context,
-    parse_genome_build, parse_shuffle_direction, parse_vcf_line, process_input_line, OutputFormat,
+    normalize_tsv_row, normalize_tsv_summary, output_error as cli_output_error,
+    output_error_with_context as cli_output_error_with_context, parse_genome_build,
+    parse_shuffle_direction, parse_vcf_line, process_input_line, OutputFormat,
+    NORMALIZE_TSV_HEADER,
 };
 use ferro_hgvs::config::{apply_override, FerroConfig, OverrideSource};
 use ferro_hgvs::error_handling::{
@@ -23,6 +25,7 @@ use std::collections::HashSet;
 use std::fs::File;
 use std::io::{self, BufRead, BufReader, BufWriter, Write};
 use std::path::{Path, PathBuf};
+use std::sync::atomic::{AtomicUsize, Ordering};
 
 #[derive(Parser)]
 #[command(name = "ferro")]
@@ -227,8 +230,11 @@ enum Commands {
         #[arg(short, long)]
         output: Option<PathBuf>,
 
-        /// Output format
-        #[arg(short = 'f', long, default_value = "text", value_parser = ["text", "json"])]
+        /// Output format. `tsv` writes a header-bearing table (one row per
+        /// input: line, input, normalized, changed, status, detail) plus a
+        /// summary line on stderr, for answering "which of my variants
+        /// changed?".
+        #[arg(short = 'f', long, default_value = "text", value_parser = ["text", "json", "tsv"])]
         format: String,
 
         /// Shuffle direction (3prime or 5prime)
@@ -1350,6 +1356,7 @@ fn run_hgvs_to_vcf(
 }
 
 /// Outcome of processing one input line in the batch normalize/parse driver.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum LineStatus {
     /// Line was blank/comment and produced no output (not counted).
     Skipped,
@@ -1378,6 +1385,19 @@ impl LineResult {
             status: LineStatus::Skipped,
         }
     }
+}
+
+/// Write the `normalize --format tsv` header row, when that is the format.
+///
+/// Deliberately called only once the run's input is known to be readable: writing
+/// it while opening the output stream made `--input missing.txt` leave behind a
+/// header-only artifact byte-identical to the one a *successful* empty-input run
+/// produces, with nothing on the output stream to tell the two apart.
+fn write_tsv_header<W: Write>(writer: &mut W, tsv: bool) -> io::Result<()> {
+    if tsv {
+        writeln!(writer, "{}", NORMALIZE_TSV_HEADER)?;
+    }
+    Ok(())
 }
 
 /// Number of input lines processed per batch before their results are written.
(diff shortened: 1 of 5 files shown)


nh13:
Applied the review. All seven findings plus the nitpicks and the three test gaps are addressed in the existing commit (amended, force-pushed). Nothing was rejected.

**1. Incomplete control-character sanitization.** `sanitize_tsv_field` now replaces on `char::is_control()` plus `U+0085`/`U+2028`/`U+2029`, and collapses *runs* to a single space so `\r\n` leaves one space instead of two. The unit-test helper no longer leans on `str::lines()` (which shares the old implementation's blind spot): `tsv_fields` now inspects the whole row for `char::is_control()` plus the three non-control separators, so every row-shape assertion in the module gets the check for free. The integration test feeds VT/LS/NEL/ESC/FS through the CLI and asserts the same property per field.

**2. `changed=false` on failed rows.** Failure rows now emit an **empty** `changed`, matching the already-empty `normalized`. Documented on `normalize_tsv_row` and in the module doc, with a new test (`the_changed_column_agrees_with_the_summary`) that counts `true`/`false`/empty rows and asserts the triple equals the summary's `changed`/`unchanged`/`failed` exactly.

**3. No line number.** New leading **`line`** column carrying the 1-based input line number. Header is now `line	input	normalized	changed	status	detail`. It is empty in single-positional mode (no file to number lines in) — documented in the module doc, the clap help, and `normalize_tsv_row`'s docs. `line_numbers_survive_skipped_lines` pins that blank/comment lines make the line number diverge from the row index, which is why the row index was never a substitute.

**4. `preprocess_error` mislabelling.** Tokens unchanged. Instead, `NormalizeTsvFailure` carries a new section documenting that `status` names *which stage rejected the input under the active error mode* — a property of `(input, error-mode)`, not of the input alone — with the `NM_000088.3:c.(1_2)ins` case spelled out and an explicit warning that bucketing "malformed input" vs "policy rejection" must key off the error mode too. Repeated in the module doc.

**5. Header written before the input is opened.** Header emission moved into a `write_tsv_header` helper called only once the input is known readable (after `File::open`, after `stdin.lock()`, and before the single-positional row). Empty-input still gets a header (existing test still passes); new `a_missing_input_file_writes_no_table` asserts `-i missing.txt -o out.tsv` leaves the artifact empty.

**6. `changed` computed in two places.** `normalize_tsv_row` now returns `NormalizeTsvRow { row, changed }` and the binary increments its counter from `row.changed`, so there is exactly one comparison in the codebase.

**7. `detail` empty on success.** `NormalizeTsvOutcome::Normalized` now carries `corrections: &[(code, message)]`, rendered into `detail` as `CODE: message`, `; `-joined. stderr warnings are emitted exactly as before. New `a_successful_row_explains_a_correction_in_its_detail` covers both halves under lenient mode.

**Nitpicks.** `mem::forget` comment now names `tsv_row` as a borrower alongside `process`. The flush comment's inverted reasoning is replaced: bare stdout is a `LineWriter` and already flushed per row, so the real reason to flush is that with `--output` the rows sit in a `BufWriter` whose `Drop` swallows write errors — flushing is what turns a failed write into a reported failure. `format.rs` now derives every column from a single `match` on `outcome`, which removes the ordering subtlety the old "compare before sanitizing" comment had to explain. And the `input`-looks-like-a-header case is now impossible to confuse (a row is preceded by the numeric/empty `line` column) — noted in the module doc.

**Test gaps.** Added `parallel_tsv_output_is_byte_identical_to_serial`: 20 000 variants interleaved with comments/blanks (asserted to span >2 `BATCH_CHUNK_LINES` chunks), run at `-j1/2/4/8`, asserting byte-identical stdout, identical summary, exactly one header, and input order preserved (both the `input` sequence and strictly increasing line numbers). Added `reports_a_change_produced_by_the_normalizer_itself` (`c.16del` → `c.19del`, a 3'-shift through the mock `NM_000088.3` `CCCC` homopolymer at c.16-19, run under the default strict mode so no preprocessing rewrite can be responsible). Added `a_single_positional_variant_honors_the_output_flag`.

`text`/`json` bytes and exit codes are unchanged — `text_and_json_output_is_untouched` still pins the exact streams. `cargo fmt --all --check`, `cargo clippy --all-features --all-targets -- -D warnings`, `cargo nextest run --features dev` and the doctests are all clean: 8960 → 8969 tests.

nh13:
@coderabbitai review

nh13:
@coderabbitai pause

nh13:
@coderabbitai full review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1250, opened 2026-07-29
fix(normalize): describe a one-base inversion residue as a substitution

### LINKED ISSUE #1249: Normalizer: an odd-length inv whose outer bases all cancel is dropped to identity (silent variant loss; non-confluent with the delins spelling)
## What's wrong

An `inv` whose outer bases all cancel down to a **single** remaining base is normalized to identity (`=`), silently discarding a real sequence change.

```
reference   CCAATGGCC
input       c:g.3_5inv          span 3_5 = AAT,  revcomp = ATT
ferro    -> c:g.3_5=            "no change"
applied     CCATTGGCC           the sequence DID change
expected -> c:g.4A>T
```

`shorten_inversion` peels complementary outer pairs, which is correct: if the first base is the complement of the last, both survive the reverse complement unchanged. `A..T` peels, leaving the single base `A` at position 4. But a one-base residue is *not* a no-op — it is replaced by its own complement, and no standard base is self-complementary. `A` becomes `T`, so the correct description is `g.4A>T`.

## Severity: high

Two independent reasons:

1. **Silent variant loss.** The output is not merely a different spelling — `=` asserts that the reference is unchanged at that span. A real variant is normalized out of existence, with no warning. Any consumer that filters `=` drops the variant entirely.
2. **Non-confluence.** The same variant, spelled as a `delins` or a substitution, normalizes correctly. Only the `inv` spelling is lost, so two encodings of one variant land on different keys — the failure mode tracked by #1229–#1235.

| spelling of the same variant | normalizes to |
| --- | --- |
| `c:g.3_5inv` | `c:g.3_5=` ❌ |
| `c:g.3_5invAAT` | `c:g.3_5=` ❌ |
| `c:g.3_5delinsATT` | `c:g.4A>T` ✅ |
| `c:g.3_5delAATinsATT` | `c:g.4A>T` ✅ |
| `c:g.4A>T` | `c:g.4A>T` ✅ |

An explicit inverted sequence (`invAAT`) does not save it — the same path runs.

## Scope: every odd-length cancelling span, with no exceptions

Exhaustive sweep of all 4^L sequences for L in 3..=7, each inverted whole (21,824 cases):

| span length | wrongly emitted `=` | correctly emitted `=` |
| --- | --- | --- |
| 3 | 16 | – |
| 4 | – | 16 |
| 5 | 64 | – |
| 6 | – | 64 |
| 7 | 256 | – |
| **total** | **336** | **80** |

The split is exactly parity:

- an **even**-length span whose outer pairs all cancel reduces to zero bases and *is* genuine identity — those 80 cases are correct and must stay correct;
- an **odd**-length span reduces to a one-base centre, which always changes.

Every affected case is wrong; there is no subset of the one-base-residue cases where identity is the right answer for A/C/G/T. The counts are `4^((L+1)/2)` — the free choice of the first half plus the centre base.

## Root cause

`src/normalize/rules.rs:939-942`:

```rust
// If inversion shrinks to 0 or 1 base, it's identity
if e <= s + 1 {
    return None; // Becomes identity - no inversion needed
}
```

Zero and one are conflated. Only zero is identity. The `None` is consumed at `src/normalize/mod.rs:7876-7880`, which emits `position_identity()` over the *original* span — hence `g.3_5=` rather than anything mentioning position 4.

The sibling caller (`rules.rs:1329`, `canonicalize_delins` step 4d) is unaffected: after greedy shared-affix trimming a true reverse complement can never have complementary outer bases, because that would imply a shared prefix the trimmer would already have consumed. Its `debug_assert!(e > s + 1)` therefore still holds. This is the only occurrence of the conflation in the tree; the other two `position_identity()` sites (`mod.rs:7515` for `A>A`, `mod.rs:7665` for a same-base `delins`) are sound.

## Spec basis

`docs/recommendations/DNA/inversion.md:5` defines an inversion as **more than one nucleotide**, and `:16` states the replacement rule directly:

> by definition, the region inverted (`positions_inverted`) contains **more than one nucleotide**. The description `g.234inv` is therefore not allowed; a one-nucleotide inversion should be described as a substitution.

So the residue must be emitted as a substitution, not as a one-base `inv` and not as identity.

This is the normalizer-side counterpart to #1079, which handles the same rule at the *parser*. #1079 rejects a literal `g.234inv` and documents why it cannot rewrite it:

> The spec's replacement is a substitution, whose reference and alternative bases are the base at that position and its complement. Neither is recoverable from the description alone — it takes the reference sequence, which the parser does not have.

The normalizer *does* have the reference sequence, so it can and should perform exactly the rewrite the parser cannot.

## Suggested fix

Split the two cases in `shorten_inversion` — return `None` only for a genuinely empty residue, and let a one-base residue through — then type that residue as a substitution to its complement at the `NaEdit::Inversion` arm in `mod.rs`. `complement()` maps unsupported IUPAC codes to themselves, so gating the identity return on `complement(b) == b` keeps `N` (and any other self-complementary code) correctly identity, and confines the new substitution path to A/C/G/T/U.

Note `shorten_inversion` is public API (`ferro_hgvs::normalize::rules`), so the behaviour change is observable to external callers: a span that previously returned `None` may now return a one-base range.

## Reproduction

Verified on `origin/main` at 78e497a.

```rust
// tests/it/, JsonProvider over a literal contig
let p = provider("c", "CCAATGGCC");
assert_eq!(norm(&p, "c:g.3_5inv"), "c:g.4A>T");  // actual: "c:g.3_5="
```

## Discovery

Found while building probe binaries for the #1234 clamp investigation; it is pre-existing on `main` and independent of the #1229–#1235 stack.

nh13:
Closes #1249.

## The defect

`rules::shorten_inversion` peels complementary outer pairs off an inversion. That peeling is sound: when the first base is the complement of the last, both endpoints survive the reverse complement unchanged, so neither needs describing. The stopping condition was not — it treated a residue of **zero or one** base as identity.

Only zero is identity. A one-base residue is replaced by its own complement, and no standard base is self-complementary, so it is always a real change.

```
reference   CCAATGGCC
g.3_5inv    span AAT, revcomp ATT   ->  ferro emitted  g.3_5=
applied     CCATTGGCC                   the sequence DID change
correct                                 g.4A>T
```

## Why this is worse than a spelling difference

`=` asserts the reference is unchanged at that span, so the variant is normalized out of existence with no warning, and any consumer that filters identities drops it silently.

It was also a **confluence** break of the kind #1229–#1235 track — only the `inv` spelling was lost:

| spelling of the same variant | before | after |
| --- | --- | --- |
| `g.3_5inv` | `g.3_5=` ❌ | `g.4A>T` |
| `g.3_5invAAT` | `g.3_5=` ❌ | `g.4A>T` |
| `g.3_5delinsATT` | `g.4A>T` | `g.4A>T` |
| `g.3_5delAATinsATT` | `g.4A>T` | `g.4A>T` |
| `g.4A>T` | `g.4A>T` | `g.4A>T` |

## Scope, measured

An exhaustive sweep of every 4^L span for L in 3..=7 (21,824 cases) splits exactly on parity — an even-length span whose outer pairs all cancel reduces to nothing and *is* identity; an odd-length one always retains a centre base that changes:

| span length | wrongly emitted `=` | correctly emitted `=` |
| --- | --- | --- |
| 3 | 16 | – |
| 4 | – | 16 |
| 5 | 64 | – |
| 6 | – | 64 |
| 7 | 256 | – |
| **total** | **336** | **80** |

Every case the old branch fired on was wrong; there were no correct cases to weigh against it. The 80 genuine identities are unaffected and are pinned by a test.

## Spec basis

`DNA/inversion.md:5` defines an inversion as **more than one nucleotide**; `:16` gives the replacement rule directly:

> by definition, the region inverted (`positions_inverted`) contains **more than one nucleotide**. The description `g.234inv` is therefore not allowed; a one-nucleotide inversion should be described as a substitution.

This is the normalizer-side counterpart to #1079, which enforces the same rule in the *parser*. #1079 can only reject a literal `g.234inv` and documents why it cannot rewrite it — recovering the substitution needs the reference sequence, which the parser does not have. The normalizer does have it, so it performs exactly the rewrite #1079 describes. No calibration or judgement call is involved: the spec names the output form.

## The change

Two edits, plus one small helper.

- `src/normalize/rules.rs` — `shorten_inversion` returns `None` only for a genuinely empty residue, or for a one-base residue whose base is its own complement. `complement` maps every IUPAC code it does not model to itself, so `N` (and any other self-complementary code) stays identity and the new substitution path is confined to A/C/G/T/U.
- `src/normalize/rules.rs` — new `complementary_substitution(base)` returns the typed `(reference, alternative)` pair, or `None` rather than fabricating a substitution it cannot spell.
- `src/normalize/mod.rs` — the `NaEdit::Inversion` arm emits that substitution for a one-base residue. A residue it cannot type falls back to leaving the inversion as authored, rather than asserting or inventing an edit.

The sibling caller (`rules.rs`, `canonicalize_delins` step 4d) is unaffected. After greedy shared-affix trimming a true reverse complement can never have complementary outer bases — that would imply a shared prefix the trimmer already consumed — so it never sees a one-base residue and its `debug_assert!(e > s + 1)` still holds.

## Compatibility

`shorten_inversion` is public API (`ferro_hgvs::normalize::rules`), so the behaviour change is observable: a span that previously returned `None` may now return a one-base range. The signature is unchanged. This is called out in the function's doc comment.

## Verification

| gate | result |
| --- | --- |
| mock suite | **8793/8793** |
| `cargo fmt --all` | clean |

New tests in `tests/it/issue_1249_inv_one_base_residue.rs`: the reported case, the `invAAT` spelling, the even-length identity control, the five-spelling confluence check, `c.` and `n.` axis coverage, and the exhaustive parity sweep. All six failed before the change; the even-length control passed before and after, which is the guard against over-correction.

## Reading order

`src/normalize/rules.rs` first (the corrected stopping condition and the helper), then the `NaEdit::Inversion` arm in `src/normalize/mod.rs`, then the tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected normalization of single-base inversions so they are represented as substitutions when appropriate.
  * Preserved identity results when an inversion fully cancels out.
  * Improved handling of IUPAC ambiguity codes and complementary bases.
  * Applied consistent behavior across coding, noncoding, and genomic sequence descriptions.

* **Tests**
  * Added comprehensive coverage for inversion normalization, equivalent variant forms, ambiguity codes, and sequence edge cases.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -8212,6 +8212,41 @@ impl<P: ReferenceProvider> Normalizer<P> {
 
                 if let Some((new_s, new_e)) = rules::shorten_inversion(ref_seq, start_idx, end_idx)
                 {
+                    // A one-base residue is a substitution to that base's
+                    // complement, not an inversion: `inversion.md:5` defines an
+                    // inversion as more than one nucleotide and `:16` names the
+                    // substitution as its replacement. #1079 enforces the same
+                    // rule in the parser, which can only reject `g.234inv`
+                    // because deriving the substitution needs the reference
+                    // sequence; here we have it. Emitting identity instead
+                    // silently discarded the variant (#1249).
+                    if new_e == new_s + 1 {
+                        if let Some((reference, alternative)) =
+                            rules::complementary_substitution(ref_seq[new_s])
+                        {
+                            let pos = index_to_hgvs_pos(new_s);
+                            return Ok((
+                                pos,
+                                pos,
+                                NaEdit::Substitution {
+                                    reference,
+                                    alternative,
+                                },
+                                warnings,
+                            ));
+                        }
+                        // Not a base we can complement into a typed
+                        // substitution. Unreachable as written:
+                        // `shorten_inversion` yields a one-base residue only
+                        // for a base whose complement differs from it, which
+                        // `complement` answers only for the IUPAC alphabet
+                        // `Base` models — a byte outside it is left unchanged,
+                        // reads as self-complementary, and collapses to
+                        // identity before reaching here. Kept as a defensive
+                        // fallback: leave the inversion as authored rather than
+                        // assert or invent an edit.
+                        return Ok((start, end, canonicalize_edit(edit), warnings));
+                    }
                     return Ok((
                         index_to_hgvs_pos(new_s),
                         new_e as u64,
--- a/src/normalize/rules.rs
+++ b/src/normalize/rules.rs
@@ -894,7 +894,22 @@ pub(crate) fn deletion_to_repeat(
     })
 }
 
-/// Get the complement of a DNA base
+/// Get the complement of a nucleotide, over the whole IUPAC alphabet that
+/// [`crate::hgvs::edit::Base`] models.
+///
+/// Ambiguity codes complement by complementing their member set: `R` (A/G)
+/// becomes `Y` (C/T), `B` (not A) becomes `V` (not T), and so on. `W` (A/T),
+/// `S` (C/G) and `N` are the only genuinely self-complementary symbols.
+///
+/// Modelling the ambiguity codes matters because callers ask
+/// [`is_self_complementary`] whether "inverting this base changes nothing".
+/// Mapping an unmodelled symbol to itself made that test lie: `R` looked
+/// self-complementary, so an inversion resolving to a lone `R` was silently
+/// discarded as identity even though its reverse complement is `Y`.
+///
+/// The complement is returned uppercase regardless of input case, matching the
+/// long-standing behaviour for A/C/G/T/U. Bytes outside the alphabet are still
+/// returned unchanged — there is no meaningful complement to report for them.
 fn complement(base: u8) -> u8 {
     match base {
         b'A' | b'a' => b'T',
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai full review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Both inline findings and both outside-diff-range findings are fixed in c55f4a0. They are one root cause, so they are addressed together.

**Case-sensitivity in the inversion-shortening path.** Confirmed. `complement()` always answers uppercase, so every `complement(x) == y` test in this path silently failed on a soft-masked reference. Reproduced before fixing:

```
complementary_substitution(b'w') -> Some((W, W))     # degenerate: ref == alt
```

`Base::from_char` uppercases internally, which is what turned the missed identity branch into a `<pos>W>W` substitution — returned straight out of the `Inversion` arm, bypassing the `ref == alt` → identity rewrite that only runs on the original input edit.

**The suggested one-line fix would not have been enough.** `complemented == base.to_ascii_uppercase()` fixes `complementary_substitution` and the lone-centre check, but the outer-pair loop (`rules.rs:996`, `complement(first) == last`) has the same defect and was not flagged. With only the suggested change, the regression case proposed in the second finding still fails: for `b"awt"` the loop never cancels `a`/`t`, so the span never reduces to the `w` centre and `shorten_inversion` returns `Some((0, 3))` instead of `None`.

All three comparison sites now go through two helpers, so the case rule is stated once:

```rust
fn is_complementary_pair(first: u8, last: u8) -> bool {
    let first = first.to_ascii_uppercase();
    complement(first) == last.to_ascii_uppercase()
}

fn is_self_complementary(base: u8) -> bool { is_complementary_pair(base, base) }
```

Uppercasing the input *before* complementing (rather than only uppercasing the right-hand side) also keeps unmodelled bytes consistent: `X` and `x` both read as self-complementary, where comparing against `base.to_ascii_uppercase()` alone would have split them.

**Tests.** Added `shorten_inversion_treats_soft_masked_bases_like_uppercase` (including the suggested `b"awt"` → `None`, plus a parity sweep asserting each lowercase span shortens identically to its uppercase twin) and `complementary_substitution_never_describes_a_base_as_itself`. Both fail against the pre-fix code — the second with exactly `left: Some((W, W))`, `right: None` — and pass after.

**`src/normalize/mod.rs:8205-8263`** needed no change of its own beyond a comment correction: it inherits the fix. While tracing it I found the fallback comment there was inaccurate in a way this change makes clearer — it claimed "only a byte outside it gets here", but a byte outside the modelled alphabet is left unchanged by `complement`, reads as self-complementary, and collapses to identity in `shorten_inversion` before ever reaching that branch. The branch is provably unreachable; it is kept as a defensive fallback and the comment now says so.

One related case-sensitivity gap is left alone deliberately: `is_revcomp` and the byte comparisons in `canonicalize_delins` (`rules.rs:1215`, `1426`, `1430`) are also case-sensitive, but that is the delins→inv typing path, not inversion shortening, and changing it would move `delins` recognition well outside the scope of #1249.

Full suite green: 8922 tests, clippy and `cargo fmt` clean.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1251, opened 2026-07-29
fix(equivalence): decline overlapping SPDI triples instead of panicking

### LINKED ISSUE #1244: EquivalenceChecker: check panics (out-of-bounds slice index in apply_triples) for a cis [repeat; adjacent-del] allele vs a variant at another locus — regression in 0.11.0
> [!NOTE]

## What's wrong

`EquivalenceChecker.check` **panics** (Rust slice index out of bounds in
`equivalence::checker::apply_triples`) instead of returning an `EquivalenceResult`,
when one input is a **cis allele whose members are a tandem-repeat expansion
immediately followed by an adjacent deletion** (`g.[p_q UNIT[n]; q_r del]`) and the
other input is any variant at a different locus:

```
TEMPLATE:g.[11_14AT[4];14_20del]  vs  TEMPLATE:g.1C>A   -> panic
```

Because it is a panic and not an error, it crashes the interpreter through
`pyo3_runtime.PanicException` (which subclasses `BaseException`, so it is not caught
by a normal `except Exception` and cannot be handled like the documented
`EquivalenceError`). Each member alone is fine — the repeat expansion by itself and
the deletion by itself both return a normal level; only the compound allele
compared against a variant elsewhere triggers the crash.

## Severity: high

- **Hard crash, not a wrong answer or a catchable error** — a single such pair
  aborts the whole process; it cannot be caught as `EquivalenceError`.
- **Common in real data** — on one real 2793 nt reference, comparing a single
  observed cis allele against the expected set panics for **544 of 913** pairs.
- **Regression** — the identical inputs are handled without a panic on 0.10.1
  (see below), so upgrading silently turns a working comparison into a crash.
- **Defeats the intended fallback** — `check` is the recommended path when
  normalized strings differ; here it crashes on exactly such inputs.

## Reproduction (verified on v0.11.0, commit 16b98af; contrasted with v0.10.1, 43ec9ab)

Harness: a single synthetic transcript on contig `TEMPLATE` (invented sequence, no
real accession, coordinates, or reference bases). Positions 11-20 are an `(AT)`
tandem repeat.

```python
import ferro_hgvs

print(f"ferro_hgvs {ferro_hgvs.__version__}")

# --- Fully synthetic, anonymized reference (no real locus) ---
# 30 nt; positions 11-20 are an (AT) tandem repeat, (AT)5.
SEQ = "CAGCTAGCTG" "ATATATATAT" "GCGCGCGCGC"
      "protein_id": TID, "genome_build": "Unknown",
json.dump({"version": "1.0", "genome_build": "Unknown",
           "transcripts": [tx], "genomic_sequences": {CTG: SEQ}}, open(ref, "w"))

def level(a, b):
    return str(ec.check(ferro_hgvs.parse(a), ferro_hgvs.parse(b)).level)

# Sanity: the checker works on simple pairs
print("sanity  1C>A vs 1C>A       ->", level("TEMPLATE:g.1C>A", "TEMPLATE:g.1C>A"))
print("sanity  1C>A vs 1delinsA   ->", level("TEMPLATE:g.1C>A", "TEMPLATE:g.1delinsA"))

# Controls: each member of the compound allele, alone, is fine
print("control repeat-alone       ->", level("TEMPLATE:g.11_14AT[4]", "TEMPLATE:g.1C>A"))
print("control del-alone          ->", level("TEMPLATE:g.14_20del", "TEMPLATE:g.1C>A"))

# Trigger: compound [tandem-repeat expansion ; adjacent deletion] vs a variant elsewhere
print("trigger compound           ->",
      level("TEMPLATE:g.[11_14AT[4];14_20del]", "TEMPLATE:g.1C>A"))   # panics on 0.11.0
```

Observed on **v0.11.0**:

```
ferro_hgvs 0.11.0
sanity  1C>A vs 1C>A       -> Identical
sanity  1C>A vs 1delinsA   -> NormalizedMatch
control repeat-alone       -> NotEquivalent
control del-alone          -> NotEquivalent
trigger compound           -> thread panicked:
  range end index 20 out of range for slice of length 13
    core::slice::index::slice_index_fail
    ferro_hgvs::equivalence::checker::apply_triples
    ferro_hgvs::equivalence::checker::EquivalenceChecker<..>::check
    ferro_hgvs::python::PyEquivalenceChecker::__pymethod_check__
  pyo3_runtime.PanicException: range end index 20 out of range for slice of length 13
```

On a real 2793 nt reference the same defect gives `range end index 1694 out of range
for slice of length 1693`, where the out-of-range end index equals the observed
allele's genomic coordinate.

## Regression: new in 0.11.0

Same script, same inputs, on **v0.10.1** — no panic; the trigger returns a level:

```
ferro_hgvs 0.10.1
sanity  1C>A vs 1C>A       -> Identical
sanity  1C>A vs 1delinsA   -> NormalizedMatch
control repeat-alone       -> NotEquivalent
control del-alone          -> NotEquivalent
trigger compound           -> NotEquivalent
```

Both versions normalize the real observed allele to the **identical** string
`g.[1691_1694AT[4];1694_1701del]`, so this is not a normalization-output change —
the defect is specifically in `apply_triples` / `check`:

| version | commit  | trigger pair on real data (913 expected) |
|---------|---------|------------------------------------------|
| 0.10.1  | 43ec9ab | 0 panics                                 |
| 0.11.0  | 16b98af | 544 panics                               |

`apply_triples` appears to size a per-comparison buffer from the variants'
coordinates and then index it with an absolute end coordinate taken from the
repeat-expanded member. The tandem-repeat expansion makes that end index exceed the
buffer length (`E > L`: 20 > 13 synthetic; 1694 > 1693 real), so the slice access is
out of bounds. Only the repeat **combined with** an adjacent deletion triggers it;
repeat-only and deletion-only are fine. (Inferred from the panic message and which
inputs trigger it; source not inspected.)

## Expected

`check` returns an `EquivalenceResult` (or raises a catchable `EquivalenceError`) —
never panics.

nh13:
Closes #1244.

## The defect

`same_resulting_sequence` rebuilds each variant's edited window by splicing its SPDI triples into a copy of the reference. `apply_triples` applies them in **descending** position order, which is what lets it validate each stated deletion against the pristine reference — every already-applied splice sits strictly 3' of the next one, so the span about to be read is untouched.

That argument holds only while the triples are disjoint. The bounds check compared against `ref_bytes.len()`, the *original* window length, while splicing into `bytes`, whose length shrinks as deletions are applied. Two overlapping triples therefore let a range pass the check and then index past the end of the mutated buffer:

```
window          10 bases
triple applied   rel 5..10, insertion ""   -> bytes.len() becomes 5
triple next      rel 0..8                  -> 8 <= ref_bytes.len() (10), passes
                                             bytes.splice(0..8, ..) on len 5 -> PANIC
```

A cis allele of a repeat expansion immediately followed by an adjacent deletion produces exactly that shape: `g.[11_14AT[4];14_20del]` has two members both claiming position 14.

## Why a panic is worse than a wrong answer

`check` is the documented fallback when normalized strings differ, and it returns `Result<_, FerroError>` — callers reasonably handle failure by matching the error. A panic bypasses that contract entirely, and through the Python bindings it surfaces as `pyo3_runtime.PanicException`, which subclasses `BaseException` and so escapes `except Exception`. The issue reports 544 of 913 pairs crashing on one real reference.

## The fix

Decline overlapping triples instead of splicing them. That is also the semantically honest answer: an allele whose members claim the same reference base has no single well-defined resulting sequence, so there is nothing to compare. The caller uses this comparison **only to upgrade** a `NotEquivalent` verdict, so a decline can never invent an equivalence — it just leaves the verdict where it was.

Two details worth reviewing:

- The guard compares each triple against the **running furthest reach**, not its immediate predecessor. Adjacent-pair checking is incomplete — one long triple can span several shorter ones that do not touch each other — and the running maximum closes that.
- **Abutting is not overlapping.** A triple ending exactly where the next begins is disjoint, and so is any number of pure insertions at one interbase position, since an insertion deletes nothing and therefore claims no base. Both are pinned by tests, because an over-eager guard would silently downgrade working `SequenceMatch` results to `NotEquivalent`.

The splice bound now also checks the mutated buffer. The disjointness guard makes that redundant today; it is kept so a future change cannot turn a logic slip back into an out-of-bounds panic, and it is commented as such.

## Scope

`apply_triples` has exactly one caller (`same_resulting_sequence`, both arguments), so the blast radius is that one rung. Verified by grep rather than assumed.

## Verification

| gate | result |
| --- | --- |
| mock suite | **8792/8792** |
| `cargo fmt --all` | clean |

New tests in `tests/it/issue_1244_equivalence_overlap_panic.rs`. Two reproduce the crash (both argument orders) and failed before the change with the reported out-of-bounds panic. **Four passed before and after** and are the real safety net, since the risk in this change is declining too much:

- a disjoint cis allele still reaches the `SequenceMatch` rung (#1158's case A);
- members that merely abut stay comparable;
- an overlapping allele compared with itself is settled by the string rung before any reconstruction;
- normalizing the same overlapping allele does not panic either.

## Not fixed here

The verdict for an overlapping allele is now `NotEquivalent` rather than a crash. That is the correct conservative answer for an input whose resulting sequence is undefined, but it is not a claim that such alleles are handled *well* — an allele with overlapping members is malformed, and stopping the normalizer from emitting one is #1234's job (PR #1236).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented overlapping genetic edits from causing crashes during equivalence checks and normalization.
  * Added safeguards for invalid edit ranges and correctly distinguishes overlapping, disjoint, and adjacent edits.
  * Preserved reliable results for equivalent comparisons and comparisons involving different loci.
  * Ensured valid insertion-only edits continue to be recognized correctly, including repeated and adjacent insertions.

* **Tests**
  * Added regression coverage for overlapping, disjoint, adjacent, insertion-only, and self-comparison scenarios.
  * Verified that normalization succeeds consistently and remains idempotent.

### DIFF
--- a/src/equivalence/checker.rs
+++ b/src/equivalence/checker.rs
@@ -503,12 +503,16 @@ impl<P: ReferenceProvider> EquivalenceChecker<P> {
 /// base is not `A`), we cannot faithfully reconstruct the edit, so we decline
 /// (`None`) rather than assert a sequence equivalence we cannot trust. Also
 /// returns `None` if any triple falls outside the window (a defensive guard;
-/// callers build the window to cover every triple).
+/// callers build the window to cover every triple), or if two triples overlap
+/// (see [`triples_are_disjoint`]).
 fn apply_triples(reference: &str, win_start: u64, triples: &[SpdiVariant]) -> Option<String> {
     let ref_bytes = reference.as_bytes();
     let mut bytes = ref_bytes.to_vec();
     let mut ordered: Vec<&SpdiVariant> = triples.iter().collect();
     ordered.sort_by_key(|t| std::cmp::Reverse(t.position));
+    if !triples_are_disjoint(&ordered) {
+        return None;
+    }
     for t in ordered {
         let rel = t.position.checked_sub(win_start)? as usize;
         let end = rel.checked_add(t.deletion.len())?;
@@ -522,9 +526,53 @@ fn apply_triples(reference: &str, win_start: u64, triples: &[SpdiVariant]) -> Op
         if !ref_bytes[rel..end].eq_ignore_ascii_case(t.deletion.as_bytes()) {
             return None;
         }
+        // The splice targets the mutated buffer, whose length no longer matches
+        // the reference once a length-changing edit has been applied. The
+        // disjointness guard above already makes `end <= bytes.len()` hold;
+        // bound it explicitly anyway so a future change cannot turn a logic
+        // slip back into an out-of-bounds panic (#1244).
+        if end > bytes.len() {
+            return None;
+        }
         bytes.splice(rel..end, t.insertion.bytes());
     }
     String::from_utf8(bytes).ok()
+}
+
+/// Whether no two of `ordered` claim the same reference base.
+///
+/// `ordered` must be sorted by descending position, as [`apply_triples`] leaves
+/// it. That descending application order is what lets each stated deletion be
+/// validated against the pristine reference: every already-applied splice sits
+/// strictly 3' of the next one, so the span about to be read is untouched. The
+/// argument holds only while the triples are disjoint — overlapping ones both
+/// invalidate that validation and can index past the end of the shrinking
+/// buffer, which is the out-of-bounds panic of #1244.
+///
+/// Declining is also the honest answer semantically: an allele whose members
+/// claim the same base has no single well-defined resulting sequence, so there
+/// is nothing to compare. The caller uses the comparison only to *upgrade* a
+/// `NotEquivalent` verdict, so a decline never invents an equivalence.
+///
+/// Two triples that merely abut are disjoint, and so is any number of pure
+/// insertions at one interbase position — an insertion deletes nothing and
+/// therefore claims no base.
+fn triples_are_disjoint(ordered: &[&SpdiVariant]) -> bool {
+    // Walk 5' -> 3' (the reverse of `ordered`) carrying the furthest 3' reach
+    // of every triple seen so far. Comparing against the running maximum rather
+    // than the immediate predecessor is what makes this complete: one long
+    // triple can span several shorter ones that do not touch each other.
+    let mut reach: Option<u64> = None;
+    for t in ordered.iter().rev() {
+        let Some(end) = t.position.checked_add(t.deletion.len() as u64) else {
+            return false;
+        };
+        if reach.is_some_and(|r| r > t.position) {
+            return false;
+        }
+        reach = Some(reach.map_or(end, |r| r.max(end)));
+    }
+    true
 }
 
 /// Extract the variant part from an HGVS string (everything after the colon).
--- a/tests/it/issue_1244_equivalence_overlap_panic.rs
+++ b/tests/it/issue_1244_equivalence_overlap_panic.rs
@@ -1,0 +1,275 @@
+//! Issue #1244: `EquivalenceChecker::check` panicked (slice index out of
+//! bounds) instead of returning a verdict, for a cis allele whose members
+//! overlap.
+//!
+//! # The bug
+//!
+//! `same_resulting_sequence` rebuilds each variant's edited window by splicing
+//! its SPDI triples into a copy of the reference. `apply_triples` applies them
+//! in **descending** position order, which lets it validate each stated
+//! deletion against the *original* reference bases — every already-applied
+//! splice sits strictly 3' of the next one, so the span it is about to read is
+//! still pristine.
+//!
+//! That reasoning holds only while the triples are disjoint. Its bounds check
+//! compared against `ref_bytes.len()` — the *original* window length — while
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai full review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1252, opened 2026-07-29
fix(python): return changed_positions as list[int], not bytes

### LINKED ISSUE #1246: CodonChange.changed_positions returns bytes but the .pyi declares list[int] (Vec<u8> → bytes in the binding)
> [!NOTE]

## What's wrong

`CodonChange.changed_positions` is typed `list[int]` in `__init__.pyi`, but at
runtime it returns **`bytes`**. `list(x)` coincidentally yields the right ints
(a `bytes` object iterates as ints), but the object is not a list, so
stub-conformant consumer code fails at runtime:

- `isinstance(x, list)` → `False`
- `x == [2]` → `False` (a `bytes` never equals a `list`)
- `json.dumps(x)` → `TypeError: Object of type bytes is not JSON serializable`
- `x.append(...)` → `AttributeError: 'bytes' object has no attribute 'append'`

Type-checkers accept all of the above against the published `list[int]` stub, so
the mismatch only surfaces at runtime.

## Severity: low–medium

- **Silent stub/runtime type mismatch** on a public property; type-checked code
  that treats the value as a `list[int]` breaks at runtime.
- **Contained** — `changed_positions` is the only `Vec<u8>` getter in the binding,
  so the blast radius is just `CodonChange` (backtranslation consumers).
- No data-correctness impact; the integer positions themselves are correct.

## Reproduction (verified on v0.11.0, commit 16b98af)

```python
import json
import ferro_hgvs as M

cc = M.Backtranslator.standard().backtranslate_substitution("Val", "Ala")[0]
cp = cc.changed_positions

print(type(cp), repr(cp))        # <class 'bytes'>  b'\x02'
print(isinstance(cp, list))      # False   (stub declares list[int])
print(list(cp))                  # [2]     (works only because bytes iterate as ints)
print(cp == [2])                 # False
try:
    json.dumps(cp)               # TypeError: Object of type bytes is not JSON serializable
except TypeError as e:
    print("json.dumps ->", e)
try:
    cp.append(1)                 # AttributeError: 'bytes' object has no attribute 'append'
except AttributeError as e:
    print("append ->", e)
```

Observed:

```
<class 'bytes'> b'\x02'
False
[2]
False
json.dumps -> Object of type bytes is not JSON serializable
append -> 'bytes' object has no attribute 'append'
```

(For contrast, the sibling properties `ref_codon`/`alt_codon` are `str` as declared.)

## Root cause + fix

The getter returns a `Vec<u8>`, which pyo3 maps to Python `bytes` (a special case),
rather than a `list[int]`:

```rust
// src/python.rs:3840
#[getter]
fn changed_positions(&self) -> Vec<u8> {
    self.inner.changed_positions.clone()
}
```

(The backing field `src/backtranslate/mod.rs:38` is `pub changed_positions: Vec<u8>`.)

Return a type pyo3 renders as a Python list of ints — e.g. collect into
`Vec<u32>`/`Vec<i64>` in the getter (`self.inner.changed_positions.iter().map(|&p| p as u32).collect()`)
— so the runtime matches the declared `list[int]`. (Alternatively, retype the stub
to `bytes`, but `list[int]` is clearly the intended API for codon positions.)

## Expected

`CodonChange.changed_positions` returns a `list[int]`, matching the `.pyi`.

nh13:
Closes #1246.

## The defect

`CodonChange.changed_positions` returned a Rust `Vec<u8>`. pyo3 special-cases that one type to Python `bytes` rather than a list of integers, while the published stub declares `list[int]`. Type-checkers accepted stub-conformant code that then failed at runtime:

```python
cp = M.Backtranslator.standard().backtranslate_substitution("Val", "Ala")[0].changed_positions
type(cp)                  # <class 'bytes'>  b'\x02'
isinstance(cp, list)      # False   — stub says list[int]
cp == [2]                 # False   — a bytes never equals a list
json.dumps(cp)            # TypeError: Object of type bytes is not JSON serializable
cp.append(1)              # AttributeError: 'bytes' object has no attribute 'append'
list(cp)                  # [2]     — works only because bytes iterate as ints
```

That last line is why this went unnoticed: the one idiom people reach for happens to produce the right answer.

## The fix

Widen the getter to `Vec<usize>` so pyo3 produces a real list. **The stub already declared the correct type and is unchanged** — this makes the runtime match what was always advertised, so there is no API redefinition to weigh, only a defect to close.

## Scope

`changed_positions` is the **only** `Vec<u8>` in `src/python.rs`; every other `Vec` getter returns `Vec<String>`, `Vec<PyX>`, or `Vec<(usize, String)>`, all of which already map to a list. Verified by grep over the binding rather than taken from the report, so no sibling getter needs the same treatment.

## A test that pinned the bug

`tests/python/test_backtranslation.py` asserted `isinstance(change.changed_positions, bytes)` — it encoded the defect as expected behaviour, so it is updated to assert the documented contract instead. Worth flagging explicitly in review: this is the one pre-existing assertion the change deliberately inverts.

The new tests assert the *contract the stub advertises* — is a list, holds ints, compares equal to a list, serializes as JSON, supports `append` — rather than the concrete representation, so they stay meaningful if the underlying Rust integer width ever changes.

## Verification

| gate | result |
| --- | --- |
| Python suite (`pytest tests/python/`) | **458 passed** |
| Rust suite | **8786/8786** |
| ruff check / ruff format / mypy | clean |
| `cargo fmt --all` | clean |

TDD was confirmed both ways against a locally built wheel: with the fix reverted, **5 tests fail** (4 new plus the updated existing one); with it applied, all pass. Two of the new tests — `holds_ints` and `matches_num_changes` — pass either way, because a `bytes` iterates as integers; that is expected and is called out in the test docstring rather than left to look like coverage it is not.

The manifest-backed conformance axes were not run: this change touches only a pyo3 getter and cannot reach the normalizer.

## Summary by CodeRabbit

* **Bug Fixes**
  * `changed_positions` now returns a standard Python list of integer, 1-indexed positions instead of a bytes value.
  * Position lists can be compared, serialized to JSON, and modified using normal Python list operations.

* **Tests**
  * Added coverage to verify position types, valid ranges, mutability, serialization, and consistency with the reported number of changes.

### DIFF
--- a/src/python.rs
+++ b/src/python.rs
@@ -3872,9 +3872,17 @@ impl PyCodonChange {
     }
 
     /// Position(s) in codon that changed (1-indexed)
+    ///
+    /// Widened from the underlying `u8` because pyo3 special-cases `Vec<u8>`
+    /// to Python `bytes`, which is not the `list[int]` the stub declares and
+    /// not what a list of positions should be (#1246).
     #[getter]
-    fn changed_positions(&self) -> Vec<u8> {
-        self.inner.changed_positions.clone()
+    fn changed_positions(&self) -> Vec<usize> {
+        self.inner
+            .changed_positions
+            .iter()
+            .map(|position| *position as usize)
+            .collect()
     }
 
     /// Number of nucleotide changes
--- a/tests/python/test_backtranslation.py
+++ b/tests/python/test_backtranslation.py
@@ -58,8 +58,10 @@ class TestCodonChange:
         assert hasattr(change, "changed_positions")
         assert isinstance(change.ref_codon, str)
         assert isinstance(change.alt_codon, str)
-        # changed_positions is bytes containing position values
-        assert isinstance(change.changed_positions, bytes)
+        # changed_positions is a list of 1-indexed positions, as the stub
+        # declares. It used to come back as `bytes` — see #1246.
+        assert isinstance(change.changed_positions, list)
+        assert all(isinstance(position, int) for position in change.changed_positions)
 
     def test_codon_change_num_changes(self) -> None:
         bt = ferro_hgvs.Backtranslator.standard()
--- a/tests/python/test_issue_1246_changed_positions_list.py
+++ b/tests/python/test_issue_1246_changed_positions_list.py
@@ -1,0 +1,73 @@
+"""Issue #1246: ``CodonChange.changed_positions`` must be a real ``list[int]``.
+
+The getter returned a Rust ``Vec<u8>``, and pyo3 special-cases that one type to
+Python ``bytes`` rather than a list of integers. The published stub declares
+``list[int]``, so type-checkers accepted code that then failed at runtime:
+``isinstance(x, list)`` was ``False``, ``x == [2]`` was ``False`` (a ``bytes``
+never equals a ``list``), and ``json.dumps(x)`` raised ``TypeError``.
+
+Only ``list(x)`` happened to work, because a ``bytes`` iterates as integers —
+which is why the mismatch went unnoticed.
+
+These tests assert the *contract the stub advertises*, not the representation,
+so they stay meaningful if the underlying Rust integer width ever changes.
+"""
+
+import json
+
+import ferro_hgvs
+
+
+def _a_codon_change() -> ferro_hgvs.CodonChange:
+    """A single-nucleotide codon change: Val -> Ala differs at position 2.
+
+    Val is GTx and Ala is GCx across the whole codon table, so *every* codon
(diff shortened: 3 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai full review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #630, opened 2026-06-13
Python type stub `ErrorType` missing discriminants 41/42 (UnresolvableCentromere, TranscriptFlankNotDescribable)


nh13:
The Python type stub `python/ferro_hgvs/__init__.pyi` defines `class ErrorType(IntEnum)`, but its member list skips two discriminants that exist in the Rust `PyErrorType` enum (`src/python.rs`). On `main` the stub jumps straight from `NonConformantBracketCardinality = 40` to `IntronicOnBareTranscript = 43`, omitting:

- `UnresolvableCentromere = 41` (W4005)
- `TranscriptFlankNotDescribable = 42` (W4006)

Both variants are present at runtime — defined in the `PyErrorType` enum and wired through the `From<ErrorType>` / `From<PyErrorType>` impls in `src/python.rs` (discriminants 41 and 42). So the stub is out of sync with the actual module.

This is not caught by CI (there is no check enforcing stub ↔ enum parity), and it predates the recent `IntronicOnBareTranscript = 43` addition — adding 43 only made the existing `40 → 43` gap visible.

### Impact

IDEs and type-checkers (mypy) won't recognize `ferro_hgvs.ErrorType.UnresolvableCentromere` or `ferro_hgvs.ErrorType.TranscriptFlankNotDescribable`, even though both are valid `ErrorType` members at runtime.

### Fix

Add the two missing members to the stub so the sequence reads 40, 41, 42, 43 in order, with names matching the Rust `PyErrorType` variants exactly:

```python
    NonConformantBracketCardinality = 40
    UnresolvableCentromere = 41
    TranscriptFlankNotDescribable = 42
    IntronicOnBareTranscript = 43
```

Optionally, add a test (or CI step) asserting the `.pyi` `ErrorType` members match the Rust enum to prevent future drift.

Surfaced during review of #577 and #620.

PR #1253, opened 2026-07-29
fix(python): make the native enums honour the contract their stubs declare

### LINKED ISSUE #1245: Python enums declared IntEnum are broken at runtime: 8/9 unhashable (TypeError on set/dict/Counter), plus no .name/.value/ordering/by-value/iteration (all 9 enums)
> [!NOTE]

> **Update (scope expanded):** originally this reported that the `.pyi` declares the
> enums as `IntEnum` but the runtime lacks `.name`/`.value`. On exercising the
> objects, the gap is broader and includes **runtime crashes**, not just typing:
> **8 of 9 enum types are unhashable**, none are orderable/constructable-by-value/
> iterable, and — tellingly — **`Axis` is hashable while the other 8 are not**.

## What's wrong

All 9 exported enums are declared `class X(IntEnum)` in the `.pyi` stubs, but the
runtime pyo3 enums honor almost none of the `IntEnum` contract. There are two
severities of problem:

**(1) Runtime crashes (not merely a typing mismatch):**
- **8 of 9 enum types are unhashable** — `hash(x)`, `{x}`, `{x: ...}`,
  `collections.Counter(...)`, `itertools.groupby(...)` all raise
  `TypeError: unhashable type`. (`Axis` is the lone exception — it *is* hashable.)
- **No ordering** — `sorted([...])` raises `'<' not supported`.

**(2) Stub/typing gaps (advertised by `IntEnum`, absent at runtime):**
- `.name` and `.value` raise `AttributeError`.
- **By-value construction** `X(0)` raises `cannot create instances`.
- **Iteration / `__members__`** — `list(X)` raises `'type' object is not iterable`.
- `isinstance(x, enum.Enum)` is `False` (`type(x).__mro__ == [X, object]`).

What *does* work: identity `==`, `x == <int>`, `int(x)`, and `str`/`repr`.

Affects all 9: `Axis`, `Consequence`, `Impact`, `ErrorMode`, `ErrorType`,
`ErrorOverride`, `EquivalenceLevel`, `GenomeBuild`, `Strand`.

## Severity: medium–high

- **Unhashability is a hard runtime crash** in idiomatic code — e.g. counting
  consequences (`Counter(v.consequence for v in variants)`), a `set` of `Axis`,
  or a dict keyed by `GenomeBuild`/`Strand` all `TypeError` at runtime. The only
  workaround is to convert to `int`/`str` first, which defeats the enum.
- **`Axis` hashable but the other 8 not** — an internal inconsistency indicating
  an accidental omission rather than a deliberate design.
- **Stub over-promises** — type-checkers accept `.name`/`.value`/`sorted`/`X(int)`/
  `list(X)` against the `IntEnum` stubs, but each fails at runtime.
- No data-correctness impact; the failures are on the API/typing surface.

## Reproduction (verified on v0.11.0, commit 16b98af)

```python
import enum, ferro_hgvs as M

ENUMS = ["Axis","Consequence","Impact","ErrorMode","ErrorType",
         "ErrorOverride","EquivalenceLevel","GenomeBuild","Strand"]

def a_member(cls):
    for n in dir(cls):
        if not n.startswith("__") and isinstance(getattr(cls, n), cls):
            return getattr(cls, n)

print(f"{'enum':16} {'isEnum':7} {'.name':6} {'.value':7} {'hash':6} {'order':6} {'X(int)':7} {'iter':6}")
for name in ENUMS:
    cls = getattr(M, name); m = a_member(cls)
    def ok(fn):
        try: fn(); return "ok"
        except Exception: return "FAIL"
    print(f"{name:16} {str(isinstance(m,enum.Enum)):7} {str(hasattr(m,'name')):6} "
          f"{str(hasattr(m,'value')):7} {ok(lambda:hash(m)):6} "
          f"{ok(lambda:sorted([m,a_member(cls)])):6} {ok(lambda:cls(int(m))):7} {ok(lambda:iter(cls)):6}")
```

Observed:

```
ferro_hgvs 0.11.0

enum             isEnum  .name  .value  hash   order  X(int)  iter
Axis             False   False  False   ok     FAIL   FAIL    FAIL
Consequence      False   False  False   FAIL   FAIL   FAIL    FAIL
Impact           False   False  False   FAIL   FAIL   FAIL    FAIL
ErrorMode        False   False  False   FAIL   FAIL   FAIL    FAIL
ErrorType        False   False  False   FAIL   FAIL   FAIL    FAIL
ErrorOverride    False   False  False   FAIL   FAIL   FAIL    FAIL
EquivalenceLevel False   False  False   FAIL   FAIL   FAIL    FAIL
GenomeBuild      False   False  False   FAIL   FAIL   FAIL    FAIL
Strand           False   False  False   FAIL   FAIL   FAIL    FAIL
```

Detail (`EquivalenceLevel`, declared `class EquivalenceLevel(IntEnum)`):

```
int(x)                                   -> 0
str(x) / repr(x)                         -> Identical / EquivalenceLevel.Identical
x == L.Identical                         -> True
x == 0                                   -> True
x.name                                   -> AttributeError: ...has no attribute 'name'
x.value                                  -> AttributeError: ...has no attribute 'value'
hash(x)                                  -> TypeError: unhashable type: 'builtins.EquivalenceLevel'
{x}  /  {x: 1}                           -> TypeError: unhashable type: 'builtins.EquivalenceLevel'
sorted([L.NotEquivalent, L.Identical])   -> TypeError: '<' not supported ...
L(0)                                     -> TypeError: cannot create 'builtins.EquivalenceLevel' instances
list(L)                                  -> TypeError: 'type' object is not iterable
```

For contrast, the value types are well-behaved: `HgvsVariant`/`SpdiVariant` have
faithful `repr`/`str`, are **hashable**, and compare equal for equal inputs.

## Expected

The enums honor the `IntEnum` contract their stubs declare — hashable, orderable,
`.name`/`.value`, by-value `X(int)`, and iterable/`__members__` — applied uniformly
to all 9. **At minimum, the unhashable runtime crash must be fixed regardless of the
typing decision.** Alternatively, keep native pyo3 enums but (a) make them hashable
and (b) correct the `.pyi` to describe the real surface instead of `IntEnum`.

## Root cause (hypothesis — not verified against source) + fix

pyo3 simple enums declared `#[pyclass(eq, eq_int)]` get `__eq__` but no `__hash__`,
so CPython sets `__hash__ = None` → unhashable; they also lack `__lt__` (ordering),
a by-value `__new__`, and iteration/`__members__`. **`Axis` being hashable while the
others are not strongly suggests it already derives what the rest are missing** —
aligning the other 8 to `Axis` should fix hashing immediately.

For full parity, add `hash` (and `frozen`) to the `#[pyclass(...)]` and expose
`name`/`value`/members — e.g. generate them once via `strum::IntoStaticStr` + a
shared macro across all enum types — or re-expose the enums as real
`enum.IntEnum`s in the Python shim and fix the `.pyi` accordingly.

nh13:
Closes #1245.

## The defect

pyo3's `eq` option generates `__eq__`, and a Python type that defines `__eq__` without `__hash__` is **unhashable**. Eight of the nine exported enums carried only `eq, eq_int`:

```
Axis:      #[pyclass(name = "Axis", eq, eq_int, frozen, hash, from_py_object)]
other 8:   #[pyclass(name = "...",  eq, eq_int, from_py_object)]
```

So `hash(x)`, `{x}`, a dict keyed by the enum, `Counter(...)` and `groupby(...)` all raised `TypeError` — a hard runtime crash in idiomatic code, not a typing nit. `Axis` being the lone hashable one is what shows this was an accidental omission rather than a design choice, and two more clues point the same way: `Impact` already derived `PartialOrd, Ord` and `ErrorType` already derived `Hash`, neither of which the `#[pyclass]` ever exposed.

## What now works

Measured across all nine, before and after:

| | isEnum | .name | .value | hash | order | `X(0)` | `list(X)` | `int()` |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| before (`Axis`) | ✗ | ✗ | ✗ | ✓ | ✗ | ✗ | ✗ | ✓ |
| before (other 8) | ✗ | ✗ | ✗ | **✗** | ✗ | ✗ | ✗ | ✓ |
| after (all 9) | ✗ | ✓ | ✓ | ✓ | ✓ | ✓ | ✗ | ✓ |

Members hash, sort by value, expose `.name` and `.value`, and round-trip through `Cls(value)` — raising `ValueError` for an unknown value, exactly as an `IntEnum` would.

`frozen` is free here: none of the enums carry data, and no method anywhere in the binding takes `&mut self` (checked, not assumed).

## What is still not an IntEnum, and why the stubs changed

Two behaviours are out of reach: iterating the class (`list(X)`, `__members__`) needs a metaclass, and `isinstance(x, enum.Enum)` needs real `enum.Enum` inheritance. pyo3 provides neither for a `#[pyclass]` enum.

Rather than keep declaring `IntEnum` and let type-checkers keep accepting code that fails at runtime, the stubs move onto a `_NativeEnum` base that documents precisely what is and is not supported, in one place instead of nine.

**Nothing that worked at runtime stops type-checking.** The promises removed — `list(X)`, `isinstance(..., enum.Enum)` — never worked in the first place; the promises kept (`.name`, `.value`, `X(0)`, ordering, hashing) now genuinely do. Members are annotated as their own class with the value in a trailing comment (`Plus: Strand  # 0`), since the value is not assignable under the annotation.

One caveat worth knowing: `Cls(value)` returns a new equal object rather than an interned member, so compare with `==`, never `is`. That is stated in the base's docstring.

## Two sibling defects found while checking

Neither was in the issue; both turned up from comparing all nine enums against their stubs rather than trusting the report.

1. **`Strand.Unknown` was missing from the stub.** It exists at runtime, so type-checked code could not name a value the library returns.
2. **The drift guard only covered `ErrorType`.** The stub/runtime guard added for #630 checked one enum — which is exactly why the `Strand` gap survived. It now runs over all nine. The shared stub parser moved to a `conftest.py` fixture so both suites use one copy rather than a duplicate.

## Verification

| gate | result |
| --- | --- |
| Python suite | **597 passed**, 9 skipped |
| Rust suite | **8786/8786** |
| ruff check / ruff format / mypy | clean |
| `cargo fmt --all` | clean |

TDD confirmed both ways against a locally built wheel: with the Rust change reverted, **106 of the new tests fail**; with it applied, all pass.

The new tests are parametrised over *every* exported enum rather than a sample, because the defect was precisely that one of the nine had been treated differently. `test_every_member_round_trips_through_its_value` is the guard on the generated constructors — a wrong or missing arm in any of the 94 variants shows up there.

Two tests deliberately pin the *residue* — that members are not `enum.Enum` instances and the classes are not iterable — so if a future pyo3 makes those possible, the failing test is the reminder to revisit the stubs.

## Observed, not fixed

The 9 skips are a pickling probe: none of the enums are picklable. The issue does not ask for it and it is out of scope here, so the test records the current behaviour by skipping rather than asserting either way. Worth a follow-up issue if consumers hit it (caches keyed by enum, multiprocessing).

## Summary by CodeRabbit

* **New Features**
  * Python-exposed enums now provide `.name` and `.value` properties.
  * Enums support construction from valid integer values, ordering, hashing, and use in dictionaries and sets.
  * Added the `Strand.Unknown` enum member.

* **Bug Fixes**
  * Improved consistency between runtime enum behavior and published Python type information.
  * Added validation for invalid and retired enum values.
  * Preserved sparse numeric values for applicable error types.

### DIFF
--- a/python/ferro_hgvs/__init__.pyi
+++ b/python/ferro_hgvs/__init__.pyi
@@ -1,8 +1,9 @@
 """Type stubs for ferro-hgvs Python bindings."""
 
 from collections.abc import Iterable
-from enum import IntEnum
 from typing import Any, Callable, Literal, overload
+
+from typing_extensions import Self
 
 __version__: str
 
@@ -297,7 +298,53 @@ def build_transcript(config: BuildTranscriptConfig) -> BuildTranscriptReport:
 # Core Classes
 # ============================================================================
 
-class Axis(IntEnum):
+class _NativeEnum:
+    """Shared behaviour of ferro's native enums.
+
+    These are pyo3 classes, not :class:`enum.Enum` subclasses, and the
+    difference is visible: ``isinstance(member, enum.Enum)`` is ``False`` and
+    the class itself is not iterable (no ``list(Cls)``, no ``__members__``).
+    pyo3 cannot provide either for a ``#[pyclass]`` enum.
+
+    Everything else an ``IntEnum`` offers is available: members hash, order,
+    compare equal to their integer, expose :attr:`name` and :attr:`value`, and
+    can be looked up by value with ``Cls(value)`` (raising ``ValueError`` for an
+    unknown one). ``Cls(value)`` returns the *interned* member — the same object
+    bound to the class attribute — so ``is`` behaves as it does on a standard
+    library enum, and members hash like the integers they equal.
+
+    The stubs previously declared these as ``IntEnum``, which type-checkers
+    honoured while the runtime did not — see issue #1245.
+    """
+
+    def __init__(self, value: int) -> None: ...
+    @property
+    def name(self) -> str:
+        """The member's name, matching the attribute it is bound to."""
+        ...
+
+    @property
+    def value(self) -> int:
+        """The member's integer value."""
+        ...
+
+    def __int__(self) -> int: ...
+    def __hash__(self) -> int: ...
+    def __eq__(self, other: object) -> bool: ...
+    # Ordering accepts a member of the *same* enum or a plain int, mirroring
+    # what pyo3's `ord` + `eq_int` actually implement. `Self` narrows to the
+    # concrete subclass, so `Axis.Genomic < Strand.Plus` is rejected the way the
+    # runtime rejects it (`TypeError`) rather than silently type-checking. `int`
+    # stays in the union because `Axis.Genomic < 3` genuinely works.
+    #
+    # `__eq__` keeps `object`: equality is total, answering `False` across enums
+    # instead of raising.
+    def __lt__(self, other: Self | int) -> bool: ...
+    def __le__(self, other: Self | int) -> bool: ...
+    def __gt__(self, other: Self | int) -> bool: ...
+    def __ge__(self, other: Self | int) -> bool: ...
+
+class Axis(_NativeEnum):
     """The coordinate axis (reference molecule / coordinate system) of a variant.
 
     Returned by :attr:`HgvsVariant.axis`. Each member carries its single-letter
@@ -305,13 +352,13 @@ class Axis(IntEnum):
     (:meth:`is_dna` / :meth:`is_rna` / :meth:`is_protein`).
     """
 
-    Genomic = 0
-    Coding = 1
-    NonCoding = 2
-    Rna = 3
-    Protein = 4
-    Mitochondrial = 5
-    Circular = 6
+    Genomic: Axis  # 0
+    Coding: Axis  # 1
+    NonCoding: Axis  # 2
+    Rna: Axis  # 3
+    Protein: Axis  # 4
+    Mitochondrial: Axis  # 5
+    Circular: Axis  # 6
 
     def code(self) -> str:
         """The single-letter HGVS coordinate code (``g``/``c``/``n``/``r``/``p``/``m``/``o``)."""
@@ -795,14 +842,14 @@ class OneBasedPos:
 # Equivalence Classes
 # ============================================================================
 
-class EquivalenceLevel(IntEnum):
+class EquivalenceLevel(_NativeEnum):
     """Equivalence level between two variants."""
 
-    Identical = 0
-    NormalizedMatch = 1
-    AccessionVersionDifference = 2
-    NotEquivalent = 3
-    SequenceMatch = 4
+    Identical: EquivalenceLevel  # 0
+    NormalizedMatch: EquivalenceLevel  # 1
+    AccessionVersionDifference: EquivalenceLevel  # 2
+    NotEquivalent: EquivalenceLevel  # 3
+    SequenceMatch: EquivalenceLevel  # 4
 
     def is_equivalent(self) -> bool:
         """Returns true if the variants are considered equivalent."""
@@ -910,28 +957,28 @@ class EquivalenceChecker:
 # Effect Prediction Classes
 # ============================================================================
 
-class Consequence(IntEnum):
+class Consequence(_NativeEnum):
     """Sequence Ontology consequence term."""
 
-    TranscriptAblation = 0
-    SpliceAcceptorVariant = 1
-    SpliceDonorVariant = 2
-    StopGained = 3
-    FrameshiftVariant = 4
-    StopLost = 5
-    StartLost = 6
-    MissenseVariant = 7
-    InframeInsertion = 8
-    InframeDeletion = 9
-    ProteinAlteringVariant = 10
-    SpliceRegionVariant = 11
-    SynonymousVariant = 12
-    StartRetainedVariant = 13
-    StopRetainedVariant = 14
-    FivePrimeUtrVariant = 15
-    ThreePrimeUtrVariant = 16
-    IntronVariant = 17
-    CodingSequenceVariant = 18
+    TranscriptAblation: Consequence  # 0
+    SpliceAcceptorVariant: Consequence  # 1
+    SpliceDonorVariant: Consequence  # 2
+    StopGained: Consequence  # 3
+    FrameshiftVariant: Consequence  # 4
+    StopLost: Consequence  # 5
+    StartLost: Consequence  # 6
+    MissenseVariant: Consequence  # 7
+    InframeInsertion: Consequence  # 8
+    InframeDeletion: Consequence  # 9
+    ProteinAlteringVariant: Consequence  # 10
+    SpliceRegionVariant: Consequence  # 11
+    SynonymousVariant: Consequence  # 12
+    StartRetainedVariant: Consequence  # 13
+    StopRetainedVariant: Consequence  # 14
+    FivePrimeUtrVariant: Consequence  # 15
+    ThreePrimeUtrVariant: Consequence  # 16
+    IntronVariant: Consequence  # 17
+    CodingSequenceVariant: Consequence  # 18
 
     def so_term(self) -> str:
         """Get the Sequence Ontology term."""
@@ -949,13 +996,13 @@ class Consequence(IntEnum):
         """Get a human-readable description."""
         ...
 
-class Impact(IntEnum):
+class Impact(_NativeEnum):
     """Variant impact level (VEP-style)."""
 
-    Modifier = 0
-    Low = 1
-    Moderate = 2
-    High = 3
+    Modifier: Impact  # 0
+    Low: Impact  # 1
+    Moderate: Impact  # 2
+    High: Impact  # 3
 
 class ProteinEffect:
     """Protein effect prediction result."""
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai full review

msto:
[issue](https://pullpo.io/cc?l=issue) 
> One caveat worth knowing: Cls(value) returns a new equal object rather than an interned member, so compare with ==, never is. 

This is a footgun for anyone used to following the Python standard library's direction on enumeration comparison.

> Enumeration members are compared by identity:
> 
> ```py
> Color.RED is Color.RED
> True
> Color.RED is Color.BLUE
> False
> Color.RED is not Color.BLUE
> True
> ```



https://docs.python.org/3/howto/enum.html#comparisons

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Both remaining findings are fixed in 77f2138 — neither had an inline thread to resolve, so noting them here.

**Outside-diff-range comment (`src/python.rs:543-624`) — `eq_int` vs. the derived hash.** Confirmed and fixed across all nine enums. `eq_int` makes members compare equal to plain ints, but pyo3's `hash` option delegates to the Rust `Hash` derive, which hashes the variant rather than its discriminant. Every member was equal to an int it hashed differently from:

```
Axis.Genomic == 0             -> True
hash(Axis.Genomic) == hash(0) -> False
Axis.Genomic in {0: "x"}      -> False     # but 0 in {0: "x"} is True
```

Dropped the `hash` pyclass option on all nine and gave each a manual `__hash__` returning its discriminant, so `hash(member) == hash(member.value)`. The Rust `Hash` derive is kept — it is Rust-side only once the pyclass option is gone, and nothing Python-visible reads it. Two new parametrised tests cover it: `test_members_hash_like_the_int_they_equal` and `test_members_and_their_ints_are_interchangeable_as_keys`. Against the pre-fix build they fail 18 times across all nine enums; after the fix, zero.

**@msto — `Cls(value)` returning a new object rather than an interned member.** Agreed this is a footgun, and it is now removed rather than documented around: `Cls(value)` returns the interned member, so `is` behaves as it does on a standard library enum.

```
Axis(0) is Axis.Genomic  -> True
```

Worth noting the scope was narrower than it looked: values crossing the Rust boundary (`variant.axis`) were *already* interned by pyo3, so `is` only broke for the by-value constructor this PR introduced. `#[new]` now returns `Py<Self>` built through `IntoPyObject`, which is the same conversion path that interns the boundary values, instead of returning `Self` (which allocates a fresh instance). `ValueError` on an unknown value is unchanged. Covered by `test_construction_returns_the_interned_member` and `test_members_returned_from_the_library_are_also_interned`; the stub and module docstrings that promised the old behaviour are updated.

Also rebased onto latest `main`. Full suite green: 8836 Rust tests, 633 Python tests, clippy clean under both `--features dev` and `--features python`.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
All five findings from the latest review are handled in 28a64f9 — four applied, one declined with reasoning below.

**`__lt__`/`__le__`/`__gt__`/`__ge__` operand type (`__init__.pyi:332-335`) — fixed, but not as suggested.** The suggested self-bound `TypeVar` for both `self` and `other` cannot do what the finding asks of it: it would reject `Axis.Genomic < 3`, which the finding itself notes is supported. Checked against the built extension rather than inferred:

```
Axis.Genomic < 3            -> True
Axis.Genomic < Axis.Rna     -> True
Axis.Genomic < Strand.Plus  -> TypeError
Axis.Genomic < "x"          -> TypeError
Axis.Genomic == Strand.Plus -> False        # equality is total, not an error
```

So the operand set is `Self | int`, and that is what the stubs now declare. `Self` narrows to the concrete subclass, so it rejects cross-enum ordering exactly where the runtime raises, while `int` stays in the union because that comparison genuinely works. A bound `TypeVar` would not have helped even for the cross-enum case: mypy solves it to the join, `_NativeEnum`, which permits `Axis < Strand`. Verified with a type probe — `Any` type-checked all four lines above, the new signature errors on precisely the two that raise:

```
error: Unsupported operand types for < ("Axis" and "Strand")  [operator]
error: Unsupported operand types for < ("Axis" and "str")  [operator]
```

`__eq__` deliberately keeps `object`, since equality answers `False` across enums rather than raising. Pinned at runtime by `test_ordering_accepts_ints_and_rejects_foreign_operands` across all nine enums.

**Shared enum contract (`src/python.rs:585-639`) — done.** All nine enums now go through one `native_enum_pymethods!` macro that emits `py_new`, `name`, `value` and `__hash__`, plus each enum's own methods in the same generated block (`multiple-pymethods` is off, so one `#[pymethods]` per class). Net −314 lines in `python.rs`. All 94 value mappings were carried over verbatim and diffed programmatically against the pre-refactor file — nine enums, tables identical, including `PyErrorType`'s deliberate gap at discriminant 5. The table stays explicit at each call site rather than derived, so an omitted variant cannot be silently absorbed.

**Retired `ErrorType` discriminant 5 (`test_issue_1245_enum_contract.py:201-204`) — added**, as `test_error_type_rejects_the_retired_discriminant`. It also asserts 5 is absent from the runtime member set, so the test fails if 5 is ever reassigned rather than merely rejected.

**Pickle test (`test_issue_1245_enum_contract.py:269-278`) — pinned.** Right that `except Exception` + `pytest.skip` can never fail. Now `test_members_are_not_picklable`, asserting `pytest.raises(TypeError)` on `pickle.dumps`. Confirmed that is the actual behaviour (`TypeError: cannot pickle 'builtins.Axis' object`), so the suite no longer skips it.

**`name` from `Debug` (`src/python.rs:609-618`) — declined.** It directly opposes the de-duplication finding above, which was the higher-value of the two: an explicit `match` would mean 94 hand-written arms across nine enums, reintroducing exactly the boilerplate the macro just removed. The rename risk is already covered — `test_members_expose_name` resolves `getattr(cls, member.name)`, and `test_stub_lists_exactly_the_runtime_members` compares every enum's runtime members against the stub — and the finding acknowledges that guard. The coupling stays documented on the getter.

Full suite green: 8836 Rust tests, 652 Python tests, clippy clean under `--features dev` and `--features python`, `cargo fmt` and ruff/mypy clean.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1255, opened 2026-07-29
fix(normalize): stop a cis member's shift at a sibling's bases

### LINKED ISSUE #1254: Normalizer: two deletions separated by unchanged bases merge into one deletion that denotes a different sequence (silent, no overlap to detect)
## What's wrong

Two deletions separated by unchanged bases are merged into one deletion that denotes a **different sequence**, with no warning and no structural tell.

```
reference   TAATATATATAATATATATT
input       TEMPLATE:g.[3_4del;9del]
ferro    -> TEMPLATE:g.12_14del

applied (input)  TAATATTAATATATATT
applied (output) TAATATATATATATATT     ← different
```

Both are 17 bases, both are well-formed, non-overlapping, single-member descriptions. Nothing about the output invites suspicion.

Working it through, 1-based:

```
pos   1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20
base  T A A T A T A T A  T  A  A  T  A  T  A  T  A  T  T
```

- `[3_4del;9del]` removes 3 (`A`), 4 (`T`) and 9 (`A`) → `TAATATTAATATATATT`
- `12_14del` removes 12 (`A`), 13 (`T`) and 14 (`A`) → `TAATATATATATATATT`

They differ at positions 7-8 of the result.

Each member 3'-shifts legitimately on its own inside the `(AT)` tract, but the tract ends at position 11 (`…A T A | A T…`, with `11=A, 12=A` breaking the repeat). The merged deletion is placed past that boundary, where removing three bases is no longer the same edit.

## Severity: high

- **Silent corruption of a variant's meaning.** The output is not a different spelling of the same change; it describes different bases. There is no overlap, no warning, and no malformed shape a consumer could reject.
- **On `main` today**, independent of any open PR.
- **Same failure mode as #1229–#1235**, but worse in kind: those produce two non-confluent spellings of one variant, where both still denote the variant. This one denotes a different variant.

## Not the same as #1234

A three-member variant of the same input is instructive:

```
input       TEMPLATE:g.[3_4del;9del;13T>A]
main     -> TEMPLATE:g.[12_14del;13T>A]     overlapping — a consumer can reject it
```

There the corruption is at least *visible*, because the merged deletion collides with the substitution. #1234 / PR #1236 is about that collision. The two-member case above has no sibling to collide with, so the same underlying error passes through undetected.

That distinction matters for reading the #1236 fuzzing results: some of the "branch is silently wrong" cases attributed to the sibling clamp are this bug reaching the clamp with an already-corrupted member list. Instrumenting the clamp on the three-member input shows it receives `[g.9_11del; g.13T>A]` — already not equivalent to the input — and correctly leaves the sequence unchanged from there.

## Spec

`general.md:34`:

> two variants separated by one or more nucleotides should be described individually and **not** as a "delins"

`3_4del` and `9del` are separated by four unchanged bases (5-8), so they should stay two members. The exception at `:35` is codon-scoped and does not apply on a genomic axis.

## Suggested direction

The merge of two shifted deletions needs to be conditioned on the resulting sequence rather than on adjacency after shifting. PR #1240 (`fix/runtime-sequence-preservation-guard`) implements a guard of exactly this shape at the allele level; this input is a good test case for it, and a check there would catch this class rather than this instance.

Whatever the fix, the regression test should assert against an **independently applied** reference sequence. Asserting `check(input, normalize(input))` cannot catch this: it normalizes both sides, so it passes on `main` today.

## Reproduction

Verified on `origin/main` at 78e497a.

```rust
// tests/it/, JsonProvider over contig TEMPLATE = "TAATATATATAATATATATT"
assert_eq!(norm("TEMPLATE:g.[3_4del;9del]"), "TEMPLATE:g.12_14del");  // current
// applying the input gives TAATATTAATATATATT
// applying the output gives TAATATATATATATATT
```

## Discovery

Found while adding reference-backed verification to the #1234 sibling clamp: the verification reported the clamp preserving the sequence, which was true, and pointed upstream to the member list the clamp had been handed.

nh13:
Closes #1254.

## The defect

`g.[3_4del;9del]` over `TAATATATATAATATATATT` normalized to `g.12_14del`, which denotes different bases than the input — silently, with no overlap and no warning to detect it.

```
input applied   TAATATTAATATATATT
output applied  TAATATATATATATATT     ← differs at result positions 7-8
```

The issue frames this as a bad merge. Tracing it, the merge is the symptom and the **3'-shift is the defect**:

| pass | members | note |
| --- | --- | --- |
| in | `[3_4del; 9del]` | separated by four unchanged bases (5-8) |
| 1 | `[9del; 10_11del]` | `3_4del` shifted to its standalone most-3' position — **clear over the `9del`** |
| 2 | `9_11del` | now adjacent, so they merge — but `{9,10,11}` was never equivalent to `{3,4,9}` |
| 3 | `12_14del` | the merged deletion shifts on |

Equivalence broke at pass 1, before the merge ran. The per-member shift (`normalize_core`, inside the `normalize_allele` fixed-point loop) sees only its own member: `boundary.rs` supplies reference/exon/CDS bounds, never a sibling's position. So `3_4del` rotated through the `(AT)` tract and straight past the base its sibling deletes.

This is also why the sibling clamp in #1236 does not catch it. That clamp fires on an **overlap** (`end >= next_start`); here the member leapfrogs entirely and the two are never overlapping. #1254 reaches that clamp with an already-corrupted member list, which is where part of its "32/32 regressions survive the gate" count comes from.

## The fix

Clamp a member's shift so its **sweep window** — every position between its pre-shift and post-shift span — contains no base another member claims.

A shift is a rotation, so within a window no sibling edits, every position in it describes the same sequence; the members then sit on disjoint windows and compose exactly as the input did. Stopping one position short of the nearest sibling base is the most 3' placement that keeps that true — and it leaves the member *adjacent* to its sibling rather than past it, so the loop's next merge pass coalesces the two per `delins.md:16`.

For the reported case: `3_4del` clamps to `7_8del`, which is adjacent to `9del`, and the two merge to **`g.7_9del`** — three contiguous deleted bases, exactly the sequence the input denotes.

Brute-forcing every 3-position deletion set over that reference confirms `{7,8,9}` is in the input's equivalence class and `{12,13,14}` is not. `7_9del` is also the most 3' *contiguous* member of that class, so nothing is given up by stopping at the sibling.

Both shift directions are clamped. Under `ShuffleDirection::FivePrime` the mirror-image case `g.[5del;9_10del]` produced `g.1_3del` (denoting `TATATATAATATATATT` against the input's `TAATTATAATATATATT`) and now produces `g.5_7del`.

Insertions and duplications claim no reference base, so they are neither clamped nor treated as barriers — a member landing flush against one is the *adjacency* #999 and #1135 rely on, not a crossing. That boundary is pinned by a test.

## Scope, measured

An exhaustive two-member sweep — a deletion plus a downstream deletion or substitution, over deterministic 20-mers on `{A,T}` and `{A,C,G,T}` — applying **both sides to the reference** independently of the normalizer:

| | before | after |
| --- | --- | --- |
| cases checked | 445,148 | 445,148 |
| silent sequence-changing normalizations | **2,244** | **0** |

"Silent" is the filter the issue asks for: warning-free output whose members are disjoint, i.e. nothing a consumer could reject. Counting the visible failures too (overlapping outputs — the #1234 class) the before-figure is 11,842.

## It also resolves the visible variant

The three-member case that #1234 / PR #1236 tracks:

| input | main | PR #1236 | here |
| --- | --- | --- | --- |
| `g.[3_4del;9del;13T>A]` | `g.[12_14del;13T>A]` — overlapping | `g.10_13delinsA` — wrong sequence | `g.[7_9del;13T>A]` ✅ |

Same root cause, so clamping the crossing fixes both the silent and the visible shape. This PR is independent of #1236/#1237/#1239/#1240 and based on `main`; it touches `merge.rs` near where #1236 adds its own clamp, so whichever lands second will need a small reconciliation.

## Confluence

All four spellings of the variant now reach one key — the failure mode #1229–#1235 track:

| spelling | before | after |
| --- | --- | --- |
| `g.[3_4del;9del]` | `g.12_14del` ❌ | `g.7_9del` |
| `g.[9del;3_4del]` | `g.12_14del` ❌ | `g.7_9del` |
| `g.[7_8del;9del]` | `g.12_14del` ❌ | `g.7_9del` |
| `g.7_9del` | `g.7_9del` | `g.7_9del` |

## The change

- `src/normalize/merge.rs` — `clamp_sibling_crossing_shifts`, plus `member_span` / `cis_kind_of` / `claims_reference_bases` and a small position-translation helper. The translation verifies it landed exactly where intended on the same region and otherwise reverts, so an axis that is not simply translatable (a `c.` span that would cross into the 5'UTR or `*`-region, a genomic base that would underflow) is left alone rather than mangled.
- `src/normalize/mod.rs` — call it in the `normalize_allele` fixed-point loop, right after per-member normalization, gated on cis and not-uncertain like every neighbouring transform.

The pass only reads positions — no provider, no sequence fetch — because the rotation argument is what makes it sound. The evidence that the argument holds is the sweep above, which is reference-backed.

## One existing test changed

`test_compact_allele_notation` asserted only that the output contained `"del"` and `"A>G"`. It was passing on `c.[9A>G;9del]` — two members claiming `c.9`, which ferro itself flags as an overlap conflict and which denotes a different sequence than the input. With the shift clamped it becomes `c.8_9delinsG` (adjacent changes as one edit), so the assertion is now an equality on that string, with the history recorded in the comment.

## Verification

| gate | result |
| --- | --- |
| `cargo nextest run --features dev` | **8794/8794** |
| `FERRO_ASSERT_IDEMPOTENT=1` full suite | **8793/8793** |
| conformance axes (`FERRO_MANIFEST`, real prepared reference) | **306/306** |

New tests in `tests/it/issue_1254_sibling_crossing_shift.rs`: the reported case, the three-member case, the four-spelling confluence check, a deletion shifting onto a substituted base, the 5'-direction mirror, and the exhaustive sweep — plus two negative controls (a member that cannot reach its sibling must still shift fully; an insertion landing flush must still collapse per #999) that guard against over-clamping.

Every assertion applies the descriptor to the reference **independently of the normalizer**, via SPDI triples. As the issue notes, `check(input, normalize(input))` cannot catch this class — it normalizes both sides, so it passed on the broken behavior. The `apply` helper also refuses an overlapping description rather than double-splicing one, so a malformed output fails the test instead of quietly producing a sequence neither side denotes.

## Reading order

`clamp_sibling_crossing_shifts` in `src/normalize/merge.rs` (the sweep-window rule), then its call site in `src/normalize/mod.rs`, then the tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved normalization of multi-part variants to prevent sibling changes from crossing or incorrectly merging.
  * Preserved sequence accuracy when shifting deletions, substitutions, and insertions.
  * Ensured consistent canonical output across equivalent variant spellings and shuffle directions.

* **Tests**
  * Added comprehensive regression coverage for sibling-crossing shifts, compact allele notation, boundary cases, and sequence preservation.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -1553,6 +1553,247 @@ pub(crate) fn coalesce_protein_adjacent_changes(
     }
 }
 
+// ----------------------------------------------------------------------------
+// Sibling-crossing shift clamp (#1254)
+// ----------------------------------------------------------------------------
+
+/// A cis member's span on the shared coordinate axis, copied out of the
+/// variant so the caller can mutate the member afterwards.
+struct MemberSpan {
+    accession: String,
+    region: Region,
+    start: i64,
+    end: i64,
+    /// Whether the edit consumes the reference bases under its span.
+    claims_bases: bool,
+}
+
+/// Read a member's span for the clamp pass, or `None` for a member this pass
+/// cannot place (wrong kind, uncertain edit, offset / special position).
+fn member_span(v: &HgvsVariant, kind: CisKind) -> Option<MemberSpan> {
+    let (accession, region, start, end, edit) = cis_axis_parts(v, kind)?;
+    Some(MemberSpan {
+        accession: accession.full(),
+        region,
+        start,
+        end,
+        claims_bases: claims_reference_bases(edit),
+    })
+}
+
+/// Whether an edit consumes the reference bases under its span.
+///
+/// Substitutions, deletions, delins and inversions replace or remove the bases
+/// they cover, so a shift that carries one of them over another's bases changes
+/// what the allele denotes. Insertions and duplications add sequence at a
+/// junction without consuming a base, so a member landing flush against one is
+/// the *adjacency* the collapse pass is meant to catch (#999, #1135) — they are
+/// neither clamped nor treated as barriers.
+fn claims_reference_bases(edit: &NaEdit) -> bool {
+    matches!(
+        edit,
+        NaEdit::Substitution { .. }
+            | NaEdit::Deletion { .. }
+            | NaEdit::Delins { .. }
+            | NaEdit::Inversion { .. }
+    )
+}
+
+/// The [`CisKind`] this pass serves, or `None` for a variant kind it does not.
+fn cis_kind_of(variant: &HgvsVariant) -> Option<CisKind> {
+    match variant {
+        HgvsVariant::Genome(_) => Some(CisKind::Genome),
+        HgvsVariant::Mt(_) => Some(CisKind::Mt),
+        HgvsVariant::Cds(_) => Some(CisKind::Cds),
+        HgvsVariant::Tx(_) => Some(CisKind::Tx),
+        HgvsVariant::Rna(_) => Some(CisKind::Rna),
+        _ => None,
+    }
+}
+
+/// Pull back any cis member whose shift carried it over a sibling's bases.
+///
+/// The 3' rule (`general.md:41`) is stated per description, with no
+/// allele-awareness, so a member is shifted to its *standalone* most-3'
+/// position. When a sibling edits a base inside the tract the member rotates
+/// through, that is not meaning-preserving: the member leapfrogs the sibling
+/// and the pair now describes a different sequence. `g.[3_4del;9del]` over
+/// `TAATATATATAATATATATT` shifted `3_4del` to `10_11del`, straight past the
+/// `9del`, and the two then merged to `g.12_14del` — a well-formed,
+/// warning-free description of *different* bases (#1254).
+///
+/// The invariant this restores: a member's **sweep window** — every position
+/// between its pre-shift and post-shift span — must contain no base claimed by
+/// another member. A shift is a rotation, so within a window no sibling edits,
+/// every position in it describes the same sequence; the members then sit on
+/// disjoint windows and compose exactly as the input did. Stopping at the last
+/// position before the nearest sibling base is the most 3' choice that keeps
+/// that true, which leaves the member *adjacent* to its sibling rather than
+/// past it — and the caller's next merge pass coalesces the two
+/// (`delins.md:16`). For the case above the clamp yields `[7_8del;9del]`, which
+/// merges to `g.7_9del`: three contiguous deleted bases, the same sequence the
+/// input denotes.
+///
+/// Both shift directions are handled — [`ShuffleDirection::FivePrime`] crosses
+/// siblings the same way, in the other direction.
+///
+/// `before` is the member list as it entered per-member normalization and
+/// supplies both the pull-back floor and the barrier positions; `after` is the
+/// shifted list, mutated in place. The two must be index-aligned. Barriers are
+/// read from a snapshot of `after` as well, so a sibling that shifted *into*
+/// this member's window also blocks; taking them from snapshots rather than the
+/// live list keeps the pass independent of member order.
+///
+/// [`ShuffleDirection::FivePrime`]: crate::normalize::ShuffleDirection
+pub(crate) fn clamp_sibling_crossing_shifts(
+    before: &[HgvsVariant],
+    after: &mut [HgvsVariant],
+    phase: AllelePhase,
+    uncertain: bool,
+) {
+    // An uncertain allele — `g.[(…)]` — asserts only that the members are in
+    // cis, not where they sit, so repositioning one states something the input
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:

PR #1257, opened 2026-07-30
fix(normalize): stop a repeat's tract span from swallowing a sibling

nh13:
**Stacked on #1256** (which is stacked on #1255) — base is `test/issue-1234-cis-shift-coverage`. Review and merge #1255 → #1256 → this.

No issue filed for this, per instruction. It was found while measuring the residual of #1255's clamp, and is documented in #1256's module header as a known out-of-scope defect. This PR closes it out.

## The defect

`deletion_to_repeat` (`repeated.md` B2) re-expresses a deletion inside a tandem tract as a copy count over the **whole tract**, not over the deleted bases:

```
reference   TTTTTTTTTAATATATTTTA        positions 1-9 are T
g.1_2del                ->  g.1_9T[7]   correct for a lone variant
g.[1_2del;4del]         ->  g.[1_9T[7];9del]
                                        ^ 9del sits inside 1_9
```

Overlapping members are malformed — a base cannot be both deleted and part of a copy count — and the pair denotes a different sequence than the input.

This is the residual #1255 left behind, and it is a genuinely different mechanism. #1255 stops a member *shifting into* a sibling; here the span *expands over* one without moving. `NaEdit::Repeat` claims no reference bases as far as `claims_reference_bases` is concerned, and the expansion is not a translation, so the clamp skipped these on both counts. Pre-existing on `main`, independent of #1255 and #1256 — verified by running the probe against `origin/main`'s source, which produces the identical wrong output.

## The fix

The conversion happens in `normalize_na_edit`, four levels below the allele loop (`normalize_core` → `normalize_dispatch` → axis normalizer → `normalize_na_edit`), and nothing about siblings reaches it. Threading a barrier down five axis paths would be a large change for a narrow rule, so the decision is undone in the loop instead.

A repeat that (a) grew from a plain deletion and (b) whose tract span covers a sibling's bases is re-spelled as the equivalent **3'-most deletion of the same length**. B2 only rewrites notation — it never changes which bases are removed — so the re-spelling is exact, and the deleted count comes from the member's own pre-normalization span rather than being recomputed from the reference.

Re-spelling is what lets the rest of the pipeline finish the job. It runs **before** the clamp, so the clamp now has a real span to pull back, and the merge coalesces from there:

```
[1_2del; 4del]          input
[8_9del; 4del]          repeat re-spelled as its 3'-most deletion
[2_3del; 4del]          clamped at the sibling (#1255)
 2_4del                 merged
 1_9T[6]                repeat notation restored — no sibling left to span
```

Three bases gone from a nine-base tract, which is what the input says.

## Results

| input | before | after |
| --- | --- | --- |
| `g.[1_2del;4del]` | `g.[1_9T[7];9del]` ❌ | `g.1_9T[6]` |
| `g.[1del;5del;9del]` | `g.[1_9T[7];9del]` ❌ | `g.1_9T[6]` |
| `g.[1_2del;4T>A]` | `g.[1_9T[7];4T>A]` ❌ | `g.2_4delinsA` |
| `g.[1_2del;9T>A]` | `g.[1_9T[7];9T>A]` ❌ | `g.7_9delinsA` |
| `g.[1_2del;10A>G]` | `g.[1_9T[7];10A>G]` ✅ | unchanged — sibling is past the tract |
| `g.1_2del` | `g.1_9T[7]` ✅ | unchanged — no sibling to span |

It also restores confluence: `g.[1_2del;4T>A]` and `g.[2_3del;4T>A]` are the same variant, and used to reach an overlapping repeat and a clean delins respectively. Both now reach `g.2_4delinsA`.

## Scope, measured

Overlapping normalized outputs across 143,360 two-member cis alleles, both shuffle directions:

| --- | --- |
| `main` | 2,578 |
| \+ #1255 sibling-crossing clamp | 526 |
| **\+ this PR** | **0** |

The 526 were entirely this class. The sweep is committed as `no_two_member_allele_normalizes_to_overlapping_members` and checks sequence preservation as well as non-overlap, against an independently applied reference.

> **Scope of that "0".** The sweep generates a **deletion** as the first member. It does not cover a `dup` or `ins` first member, which reach the same tracts and produce the same malformed shape — `g.[1_2dup;4T>A]` still emits the overlapping `g.[1_9T[11];4T>A]` after this PR, because the demotion below only re-spells repeats grown from a deletion. So "0" means *the deletion-first class is closed*, not *the overlap class is closed*. #1258 stacks on this branch, widens the demotion to `dup`/`ins`, adds a junction clamp for insertion-like members, and takes the widened 276,480-case sweep to 0 overlapping.

## Verification

| gate | result |
| --- | --- |
| `cargo nextest run --features dev` | **8812/8812** |
| `FERRO_ASSERT_IDEMPOTENT=1` full suite | **8812/8812** |
| new tests against the branch without the fix | 6 of 10 fail, 4 negative controls pass |

Idempotency deserves a specific note: making output depend on sibling context is exactly the kind of change that can break `norm(norm(x)) == norm(x)`, because pass 2 sees an already-normalized member set. The oracle run above covers it, and the design reason it holds is that the re-spelling only fires while a sibling is still present — once the members merge, the single remaining member canonicalises to a repeat and stays there.

**Not run: the reference-backed conformance axes.** Both the scratch and backup reference volumes are offline on this machine right now, so `FERRO_MANIFEST` tests skipped. Unlike #1256 this PR does change source, so that is a real gap rather than one inherited from a parent — the nightly reference job should be checked before merge.

## Reading order

`demote_repeats_spanning_siblings` in `src/normalize/merge.rs`, then its call site in `src/normalize/mod.rs` (note it runs before the clamp, and why), then the tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved normalization of repeat-formatted deletions when neighboring edits are present.
  * Prevented repeat spans from incorrectly absorbing sibling edits.
  * Preserved sequence-equivalent results across normalization and shuffle directions.
  * Improved convergence to consistent canonical representations for overlapping repeat/deletion spellings.

* **Tests**
  * Added comprehensive coverage for sibling overlaps, substitutions, multi-deletion alleles, and randomized scenarios.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -1722,6 +1722,175 @@ pub(crate) fn clamp_sibling_crossing_shifts(
     }
 }
 
+/// Re-spell as a plain deletion any cis member whose repeat notation spans a
+/// sibling's bases.
+///
+/// `deletion_to_repeat` (`repeated.md` B2) re-expresses a deletion inside a
+/// tandem tract as a copy count over the **whole tract**, not over the deleted
+/// bases: on a nine-`T` run, `g.1_2del` becomes `g.1_9T[7]`. That widened span
+/// is correct for a lone variant, but in a cis allele it can swallow a sibling
+/// — `g.[1_2del;4del]` emitted `g.[1_9T[7];9del]`, where `9del` sits inside
+/// `1_9`. Overlapping members are malformed, and the pair denotes a different
+/// sequence than the input.
+///
+/// The conversion happens per member, deep inside `normalize_na_edit`, with no
+/// sibling context reaching it, so the decision is undone here instead: the
+/// repeat is re-spelled as the equivalent 3'-most deletion of the same number
+/// of bases, ending where the tract ends. That is the form the member would
+/// have taken had B2 declined, and it is exactly equivalent — B2 only re-writes
+/// notation, never the bases removed.
+///
+/// Restoring the deletion is what lets the rest of the pipeline finish the job.
+/// [`clamp_sibling_crossing_shifts`] can then see a real span to pull back, and
+/// the caller's next merge pass coalesces the members. `g.[1_2del;4del]` ends up
+/// as `g.1_9T[6]` — one member, three bases removed from the tract, repeat
+/// notation restored now that there is no sibling to span.
+///
+/// Runs before the clamp, so the clamp sees the deletion rather than the
+/// repeat. Members whose pre-normalization form was not a plain deletion are
+/// left alone: without a deleted-base count there is nothing to re-spell to.
+pub(crate) fn demote_repeats_spanning_siblings(
+    before: &[HgvsVariant],
+    after: &mut [HgvsVariant],
+    phase: AllelePhase,
+    uncertain: bool,
+) {
+    if phase != AllelePhase::Cis || uncertain || after.len() < 2 || before.len() != after.len() {
+        return;
+    }
+    let Some(kind) = cis_kind_of(&after[0]) else {
+        return;
+    };
+    let pre: Vec<Option<MemberSpan>> = before.iter().map(|v| member_span(v, kind)).collect();
+    let post: Vec<Option<MemberSpan>> = after.iter().map(|v| member_span(v, kind)).collect();
+
+    for i in 0..after.len() {
+        let (Some(b), Some(a)) = (pre[i].as_ref(), post[i].as_ref()) else {
+            continue;
+        };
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:

PR #1258, opened 2026-07-30
fix(normalize): stop an insertion's junction crossing a sibling

nh13:
**Stacked on #1257** (→ #1256 → #1255). Merge order: #1255 → #1256 → #1257 → this.

Closes out **item 1** of the session retro; **item 2** is handled by an edit to #1257's description.

## What was left

#1257 stops a *deletion* that canonicalises into a whole-tract repeat from swallowing a cis sibling. It re-spells only repeats grown from a deletion, so the two sibling converters — `insertion_to_repeat` and `duplication_to_repeat` — were untouched. Probing them confirmed the identical malformed shape:

```
reference          TTTTTTTTTAATATATTTTA
g.[1_2dup;4T>A]  -> g.[1_9T[11];4T>A]     `4T>A` sits inside `1_9`
g.[2_3insTT;5T>A]-> g.[1_9T[11];5T>A]
```

Probing further turned up a **second, larger** shape that has nothing to do with repeats. An insertion or duplication consumes no reference base, so `clamp_sibling_crossing_shifts` excludes it — deliberately, because a member landing flush against one is the adjacency #999 and #1135 depend on. But its **junction** shifts through a tract like anything else, and carrying it past a base a sibling edits changes what the allele denotes:

```
reference          ACAAAAAAAACGTACGTACG        A-run at 3-10
g.[2_3insA;5A>G] -> g.[5A>G;10dup]             junction moved 2|3 -> 10|11
input applied      ACAAAGAAAAACGTACGTACG
output applied     ACAAGAAAAAACGTACGTACG       ← the substituted base moved
```

Disjoint, warning-free, well-formed, wrong. The silent shape, for insertion-like edits.

## The fix

**Demotion widened.** A repeat grown from a `dup` or `ins` is re-spelled as the equivalent 3'-most **duplication** (the deletion case re-spells to a deletion). Length comes from the member's own pre-normalization form — span length for a `dup`, sequence length for an `ins`.

**Junction clamp.** An insertion-like member's junction may reach a sibling's 5' edge but not pass it. For a sibling claiming `[s, e]`, a 3'-shifted junction stops at `s - 1`. That bound *is* the #999 adjacency — its insertion lands at `306|307` against a substitution at `307`, the last position the rule permits — so the collapse still fires. Verified by brute force that `s - 1` is exactly the 3'-most sequence-preserving junction, not a conservative under-shift.

A junction sits immediately 3' of one position: `end` for a duplication, `start` for an insertion (whose span is the gap).

## Measured

276,480 two-member cis alleles — `del`/`dup`/`ins` first member × `del`/`sub` sibling × both shuffle directions — applying both sides to the reference independently of the normalizer:

| first member | overlapping before → after | sequence-changed before → after |
| --- | --- | --- |
| `del` | 0 → 0 | 0 → 0 |
| `dup` + `sub` | 316 → **0** | 554 → **0** |
| `dup` + `del` | 370 → **0** | 151 → 74 |
| `ins` + `sub` | 316 → **0** | 328 → **0** |
| `ins` + `del` | 316 → **0** | 0 → 0 |
| **total** | **1,318 → 0** | **1,033 → 74** |

Diffing the failing-case sets: **zero newly-failing cases**. The 74 are a strict subset of what already failed, so nothing regressed to reach this.

## The clamp is 3'-only — and the 5' case is unsolved, not merely unwritten

I wrote the 5' mirror first. It measured **worse than useless**: it fixed cases but turned **80 previously-correct** outputs into silently wrong ones, because a duplication's span and its junction move together under a 5' shift, so bounding the junction mis-places the copy. Net counts favoured shipping it (2,351 bad → 154). Introducing new silent corruption does not, and that is precisely the trade I argued against when closing #1236. It was **removed as defective**, not deferred. Removing it took the residual from 154 to 74 and the regressions from 80 to 0.

A 5' junction **does** cross, in a shape that mirror would not have caught either — when the sibling is **upstream** and the junction travels toward it:

```
reference        ACAAAAAAAACGTACGTACG        A-run at 3-10
g.[4A>G;9dup]    applies to  ACAGAAAAAAACGTACGTACG
normalizes to    g.4_5insG   ACAAGAAAAAACGTACGTACG   ← different
```

**The sweep in this PR cannot see that**, and that is worth stating plainly: it always places the sibling *downstream* of the first member, so under 5' shuffle the member travels away from its sibling and can never cross it. Three hand-built upstream cases fail, identically on #1257 and on this branch — pre-existing, neither fixed nor worsened here, and pinned as currently-wrong by `a_five_prime_junction_with_an_upstream_sibling_is_a_known_gap` so the blind spot is recorded rather than implied by an unqualified "0".

## The 74 residual

One shape, characterised rather than waved at: a 5'-shuffled duplication beside a deletion, where the merge cancels part of the duplicated sequence against the deletion and emits a shorter duplication in the wrong place.

```
reference        TTATTTAAATAAAAATAAAA
g.[6_7dup;9del]  applies to  TTATTTATAATAAAAATAAAA
normalizes to    g.4dup      TTATTTTAAATAAAAATAAAA   ← different
```

Distinct mechanism from the junction rule (it is the dup/deletion cancellation path, #1135 territory), pre-existing, and **pinned as currently-wrong** by `a_five_prime_duplication_beside_a_deletion_is_a_known_residual` — so whoever fixes it gets a loud failure and a pointer to re-enable the shape in the sweep. The sweep excludes it from the sequence assertion only; it must still never overlap.

## Verification

| gate | result |
| --- | --- |
| `cargo nextest run --features dev` | **8822/8822** |
| `FERRO_ASSERT_IDEMPOTENT=1` full suite | **8822/8822** |

#999 and #1135 pass unchanged, which is the load-bearing check — both depend on an insertion shifting to become flush with a neighbour, and the junction rule is built to permit exactly that.

**Not run: reference-backed conformance.** Both reference volumes are offline on this machine. This PR changes source, so that is a real gap — check the nightly reference job before merge, same caveat as #1257.

## Reading order

`clamp_sibling_crossing_junctions` in `src/normalize/merge.rs` (the junction rule and why it is 3'-only), then the widened `demote_repeats_spanning_siblings` above it, then the call sites in `src/normalize/mod.rs`, then the tests.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -1768,20 +1768,34 @@ pub(crate) fn demote_repeats_spanning_siblings(
         let (Some(b), Some(a)) = (pre[i].as_ref(), post[i].as_ref()) else {
             continue;
         };
-        // Only a repeat that grew out of a plain deletion is re-spellable.
+        // Only a repeat is re-spellable, and only back into the kind of edit it
+        // grew out of.
         if !matches!(
             cis_axis_parts(&after[i], kind).map(|(_, _, _, _, e)| e),
             Some(NaEdit::Repeat { .. })
-        ) || !matches!(
-            cis_axis_parts(&before[i], kind).map(|(_, _, _, _, e)| e),
-            Some(NaEdit::Deletion { .. })
         ) {
             continue;
         }
-        let deleted = b.end - b.start + 1;
-        // The equivalent deletion is the 3'-most `deleted` bases of the tract.
-        let start = a.end - deleted + 1;
-        if deleted <= 0 || start < a.start || b.region != a.region {
+        let Some(source) = cis_axis_parts(&before[i], kind).and_then(|(_, _, _, _, e)| match e {
+            // A deletion removes the bases under its span.
+            NaEdit::Deletion { .. } => Some((RepeatSource::Removed, b.end - b.start + 1)),
+            // A duplication adds a copy of the bases under its span.
+            NaEdit::Duplication { .. } => Some((RepeatSource::Added, b.end - b.start + 1)),
+            // An insertion adds its own sequence at a junction; its span is the
+            // gap, so the length has to come from the sequence itself.
+            NaEdit::Insertion { sequence } => sequence
+                .len()
+                .and_then(|n| i64::try_from(n).ok())
+                .map(|n| (RepeatSource::Added, n)),
+            _ => None,
+        }) else {
+            continue;
+        };
+        let (source, length) = source;
+        // Either way the equivalent edit sits on the 3'-most `length` bases of
+        // the tract: removing them, or duplicating them.
+        let start = a.end - length + 1;
+        if length <= 0 || start < a.start || b.region != a.region {
             continue;
         }
         // Does the tract actually span a sibling's bases? Both snapshots, so a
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:

PR #1265, opened 2026-07-30
ci: flag open pull requests whose head commit has no CI run

nh13:
## The problem

A force-push does not always dispatch `pull_request: synchronize`. When it does not, the PR keeps the head SHA you pushed and simply has no run for it — and because `gh pr checks` reports only the checks that *did* report, a PR in this state renders **"All checks have passed"** off the back of whatever non-Actions check happens to be present (CodeRabbit's "Review skipped", say).

Green badge, no CI, and nothing indicating something is missing.

Observed on #1242 and #1243 while restacking them: correct head SHAs on GitHub, **zero workflow runs for fifteen minutes**, both showing a full green board. Both had been force-pushed while their base branch was itself being rewritten and closed, which is the suspected trigger but is not confirmed.

`ci.yml`'s `on:` block already documents a *different* cause of the same silent green — a `branches:` filter matching the PR base — and fixes it there. This PR addresses the class rather than a cause: whatever the reason, a head commit that was never built should be visible as a problem.

## What it does

Hourly, plus `workflow_dispatch`. Lists open PRs and checks each head SHA for a `CI` run, reporting through the existing `report-failure.yml` rolling-issue reporter so a persistent case updates one issue rather than filing a new one every hour.

Three details that matter:

- **Looks runs up by head SHA, not by branch.** A branch-scoped query answers "has this branch ever built", which is satisfied by a stale run on the previous head and is not the question being asked.
- **Asserts the audited workflow name exists.** Renaming `CI` in `ci.yml` without updating this file would otherwise turn the audit into a permanent false positive reporting every PR; instead it fails loudly with the fix named.
- **Includes drafts.** A draft with no CI misleads its author exactly as much as a ready one, and is cheaper to fix early.

## Verification

Ran the audit logic against the live repository. All 13 open PRs correctly reported as having a run:

```
auditing 13 open pull request(s)
  ok       #1259 c39b5ab (1 run(s))
  ok       #1253 48327a3 (1 run(s))
  …
  ok       #1024 4cd939b (1 run(s))
--- missing: 0 ---
```

And confirmed it is not vacuous — an intermediate commit that is pushed but was never a PR head is flagged, while the head of the same branch is not:

```
b299725 -> CI runs = 0  <-- would be flagged
0df4647 -> CI runs = 1  ok
```

`actionlint` clean.

## What this is not

A **required status check** is the stronger fix: it blocks the merge rather than reporting after the fact. That is a repository policy decision rather than a workflow change — `main` is currently unprotected, so adopting one changes the merge workflow for everyone. This is the part that can be fixed in-repo, and it stays useful afterwards as the thing that explains *why* a merge is blocked.

## Recovering an affected PR

Close and reopen it. `pull_request` with no explicit `types:` defaults to `[opened, synchronize, reopened]`, so the reopen dispatches a fresh run without rewriting history:

```
gh pr close <N> --repo <owner>/<repo> && gh pr reopen <N> --repo <owner>/<repo>
```

Confirmed on #1242, which went from 1 reported check to a full board immediately.

## Summary by CodeRabbit

* **New Features**
  * Added an automated audit for open pull requests to verify that required CI checks have run.
  * Audits run on a schedule and can also be started manually.
  * Draft pull requests are included in audits.
* **Bug Fixes**
  * Added failure reporting with details about affected pull requests and guidance for restoring missing CI runs.

### DIFF
--- a/.github/workflows/audit-pr-runs.yml
+++ b/.github/workflows/audit-pr-runs.yml
@@ -1,0 +1,184 @@
+# Detect open pull requests whose head commit has no CI run.
+#
+# A pushed head commit does not always end up with a `pull_request: synchronize`
+# run. When it does not, the PR keeps the head SHA you pushed and simply has no
+# run for it — and because `gh pr checks` reports only the checks that *did*
+# report, a PR in this state renders **"All checks have passed"** off the back of
+# whatever non-Actions check happens to be present (CodeRabbit's "Review
+# skipped", say). Green badge, no CI, no signal that anything is missing.
+#
+# Observed on #1242, #1243 and #1237 on 2026-07-30: correct head SHAs on GitHub,
+# zero workflow runs, all showing a full green board.
+#
+# One *possible* cause — inferred, not confirmed — is `gh pr edit --base` rather
+# than the force-push itself. Restacking four PRs in one session was suggestive:
+# three were pushed without a base change and all dispatched a run; the fourth
+# was pushed and then retargeted, and did not. Retargeting fires `pull_request:
+# edited`, which is not in the default `types:` set, and *may* supersede the
+# pending `synchronize` for that head. That is a hypothesis drawn from four
+# observations, not a documented GitHub behavior, so this workflow keys on the
+# *state* (a head with no run) rather than on any presumed trigger.
+#
+# `ci.yml`'s own `on:` block documents a *different* cause of the same silent
+# green (a `branches:` filter matching the PR base) and fixes it there. This
+# workflow addresses the class rather than a cause: whatever the reason, a PR
+# whose head has never been built should be visible as a problem.
+#
+# Why a scheduled audit and not a required status check: a required check is
+# the stronger fix, since it blocks the merge rather than reporting after the
+# fact — but `main` is not a protected branch, so adopting one is a repository
+# policy decision rather than a workflow change. This is the part that can be
+# fixed in-repo, and it stays useful afterwards as the thing that tells you
+# *why* a merge is blocked.
+#
+# The recovery, once this audit flags a PR: confirm the head SHA on the PR still
+# matches what you pushed and that no run exists for it, then close and reopen.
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
## Root cause, now much more likely identified

The original report listed the trigger as "suspected but not confirmed". A second round of restacking produced a clean natural experiment that points squarely at **`gh pr edit --base`**, not the force-push.

Round 1 — pushed #1242 and #1243, *then* changed both their bases:

```
#1242  no CI run
#1243  no CI run
```

Round 2 — pushed #1238, #1242, #1243 with **no** base change, and pushed #1237 and **then changed its base** to `main`:

```
#1238  head=1d7fbbf  CI-runs=1  ok
#1242  head=0ea3fc9  CI-runs=1  ok
#1243  head=e067470  CI-runs=1  ok
#1237  head=0a999b0  CI-runs=0  <-- the only one whose base changed
```

Four PRs, same force-push operation, same session. The only variable that tracks the outcome is whether `gh pr edit --base` ran afterwards.

That fits the trigger mechanics: retargeting a PR fires `pull_request: edited`, which is **not** in the default `types:` set (`[opened, synchronize, reopened]`). If GitHub treats the retarget as superseding the pending `synchronize` for that head — plausible, since the merge ref is recomputed — the `synchronize` never lands and no run is dispatched. The head is left never having been built.

This is still inference rather than a confirmed GitHub behaviour, but it is now a specific, testable claim rather than a shrug, and it yields a concrete rule:

> **After `gh pr edit --base`, always close/reopen the PR** — or verify the head has a run before trusting the board.

That does not change what this PR does. The audit catches the state regardless of cause, which is the point of targeting the class rather than a trigger. It does sharpen the `hint:` guidance, and I have updated the workflow comment accordingly.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1275, opened 2026-07-31
ci(coderabbit): repoint stale test paths, cover untuned modules, enforce liveness

nh13:
The config added in #627 pointed every "confirm the test still covers it" instruction at `tests/<name>.rs`, but the integration suite has since moved into `tests/it/`. **All four pointers dangle today** — `tests/idempotency_tests.rs`, `tests/normalize_property_tests.rs`, `tests/fast_path_differential.rs`, and `tests/spdi_tests.rs` do not exist at any of those paths.

A path instruction naming a file that is not there does not fail loudly. It quietly misdirects the reviewer, which is worse than saying nothing — and it is invisible until someone goes looking.

## What changed

**Repoint the stale paths**, and add the ones that were missing: `tests/it/fast_path_drift_scope.rs`, `tests/it/normalize_idempotency_proptest.rs`, `tests/it/projection.rs`, `tests/it/projection_coverage.rs`, `tests/it/hgvs_rs_projection_tests.rs`.

**Name the idempotency oracle explicitly.** `FERRO_ASSERT_IDEMPOTENT=1` sits at the shared exit of `normalize_core_checked`, which is what makes it cover both `normalize()` and every `VariantProjector` path. The instruction now flags a new early return that bypasses that exit — which disables the oracle rather than satisfying it, and looks like a clean refactor while doing so.

**Path instructions for the modules that had none.** The original covered parser, normalize, and projection. Added:

| Module | What the reviewer is told to look for |
|---|---|
| `reference/` | A placement resolved without checking the genome build (GRCh37 accession picking up a GRCh38 placement); mmap offsets not validated against the mapping length; a prepared index or cache keyed on fewer inputs than determine its contents; a provider returning an empty sequence where it should error — normalization will happily shift against it; a legacy gene selector that degrades to "not found" when the optional `ng_hosted_transcripts` artifact is absent |
| `prepare` / `check` | A download used without checksum verification; archive extraction accepting `..` or absolute components; a partial download left under the final filename; absolute paths in the manifest, which breaks portability; a manifest write that nulls an optional field it did not compute (a re-run without `--derive-ng-placements` silently clears what an earlier run set) |
| `effect` / `equivalence` / `backtranslate` | A translation that stops at the CDS end instead of reading through the 3' UTR (a frameshift or stop-loss is only correct if the extension is translated past the original terminator); a codon table used without checking the genetic code; equivalence answered from string or coordinate comparison where the contract requires normalized forms; unknown residues and `Ter` defaulted to a concrete amino acid |
| `conformance/` | The code that decides what the committed FAIL-set ledgers say — a schema field made optional (turning a malformed case into an accepted one); a case-outcome reclassification that moves cases between buckets unquantified; "no reference available" treated as a pass rather than a skip; snapshot provenance dropped |
| `legacy/` | A rewrite that changes meaning (IVS numbering resolved against the wrong intron, a `>` form expanded to a different span); legacy forms accepted by the strict parser rather than only behind the legacy config; a rewrite applied without recording that it happened |
| `vcf/` | Single-ALT assumptions on multiallelic records; the anchor-base convention applied inconsistently between the VCF side and the HGVS side; symbolic ALTs and spanning deletions treated as literal sequence |
| `python.rs` / `python_helpers.rs` | A Rust panic crossing the FFI boundary instead of becoming a Python exception; a blocking call holding the GIL; symbols outside the `abi3-py310` subset, which breaks the shipped wheel |
| `service/` | Unbounded batch input as a memory-exhaustion vector; a handler that can panic and take down the worker; internal paths echoed into a client-visible error |
| `cli` / `commands` / `bin` | Output format treated as an API — JSON field or TSV column changes without a CHANGELOG entry; diagnostics on stdout corrupting a downstream parse |
| error modes, parallelism, `python/**`, fuzz, benches, docs, CI | Failure paths ignoring the configured strict/lenient/silent mode; result ordering derived from completion rather than input index; `.pyi` stubs drifting from the PyO3 definitions; fuzz targets that suppress panics |

A scope note names the nine modules deliberately left untuned, so the next reader can tell a decision from an oversight.

**Conformance ledgers, which nothing reached before.** `tests/fixtures/{mutalyzer,biocommons}-normalize/baseline-failures/*.txt`, the hgvs-rs projection ledgers, and the blessed `*.count` baselines are committed, but the only test-tree instruction was scoped to `tests/**/*.rs` — it never matched a `.txt`. These ledgers are regenerated mechanically by `tests/it/failure_expectations_blesser_tests.rs`, which makes re-blessing the cheapest possible way to turn a failing axis green. The new instruction treats every line change as a claim requiring evidence: a FAIL→PASS removal needs the fix named; a newly added expected-failure line is a regression being blessed and needs an issue; bulk churn in a diff that does not touch `src/` means the reference data or the harness moved, and the PR must say which.

**Reference-aware tests are not pre-merge coverage.** `ci.yml` says so in a GitHub notice: PR CI does not provision `FERRO_MANIFEST`, so the mutalyzer / biocommons / hgvs-rs suites and about a dozen `issue_*` tests skip silently; only the nightly job runs them. Several instructions ask the reviewer to confirm coverage in exactly those files, so both the general test instruction and a dedicated one for the axis suites now say that landing coverage there means saying so in the PR — "the existing tests still pass" is not a statement about code those tests never ran against.

**Proptest seeds in both layouts.** `**/*.proptest-regressions` covers the sibling-file layout but not `tests/proptest-regressions/normalize_idempotency_proptest.txt`, which is where the idempotency oracle's seeds actually live. Both are covered now.

**Five `pre_merge_checks.custom_checks`:**

1. **No machine-local paths** (`error`) — the CLAUDE.md hygiene rule, now actually enforced against the diff, with the accepted alternatives named.
2. **Generated spec fixtures not committed** (`error`) — both `hgvs_spec_normalization.json` and `hgvs_spec_enumeration.json` are gitignored build artifacts; they were removed from version control precisely because committing them made every parser PR a merge-conflict magnet. The curated overrides files are explicitly exempted.
3. **Output change has a pinned regression test** — an output-changing diff must add a test pinning the *new* expected output for a concrete input. A test that merely still passes is not sufficient.
4. **Python API and stubs in sync** (`error`) — a changed `#[pyfunction]`/`#[pymethods]` signature without a matching `__init__.pyi` update.
5. **Conformance ledger change is justified** — the deterministic companion to the ledger instruction above.

**Say what the check modes actually do.** The old comment said to raise a check to `error` only alongside `request_changes_workflow`, then three checks were `error` with that flag false. Neither reading was right: CodeRabbit is not among the eight status checks `main`'s ruleset requires, and with auto-review off these run only on a requested review. `error` is a severity a human reads, not a merge gate, and the comment now says so — making one a real gate means adding CodeRabbit to the branch ruleset, which is a separate decision.

**One removal:** the `manuscript/` path filter. That directory is gitignored and can never appear in a diff.

## The maintenance obligation is now enforced, not just stated

The previous revision's pointers rotted because nothing checked them, so a comment asking future readers to re-check was the same promise that had already failed once. `tests/it/coderabbit_config_paths.rs` asserts that every `path:` glob — and every brace alternative within it — matches at least one file in the repository, and that every repo-relative path named in backticks exists.

Written before the rest of this diff was finished, it immediately caught three live dangles in this very config:

- **`src/batch`** in `src/{batch,parallel.rs,cache.rs}` — a bare directory name matches the directory entry, never the files under it, so the parallelism-and-ordering instruction never reached `src/batch/processor.rs`. This is why the check tests each brace alternative separately: the glob as a whole "matched something" via `parallel.rs`.
- **`**/*.proptest-regressions`** — missed the directory layout entirely, as above.
- **`src/data/**/*.rs`** — surfaced while calibrating the file walk.

It also caught `cargo run --features dev --example generate_spec_fixture` in the header comment: both generators are `[[bin]]` targets whose sources live under `examples/`, so `--example` does not work.

## Preserved

`auto_review.enabled: false` and `auto_incremental_review: false` are unchanged, along with the rate-limit reasoning in the comment above them. `profile: assertive` is also unchanged — it was chosen deliberately in #627 after weighing `chill`, on the grounds that this codebase's bugs are the borderline class (`chill` suppresses subtle off-by-ones, idempotency breaks, and minus-strand sign errors) and a false positive is one click to dismiss. This PR changes what a requested review *looks at*, not when one runs.

## Validation

Checked against the published `schema.v2.json`: the YAML parses, the document is schema-valid, `tone_instructions` is 232 of the 250-character cap, and each of the five custom-check names is within the 50-character cap. Every named path and every glob is verified by `cargo nextest run --features dev -E 'test(coderabbit_config)'` — 7 tests, green — which runs in the normal test job, so no new CI wiring. `cargo fmt` and `cargo clippy --features dev --all-targets -- -D warnings` are clean.

The earlier revision of this PR claimed all 55 named paths were verified to exist. That was true at the file level and still missed the three glob-level dangles above — which is the argument for the test rather than a manual pass.

## Note for reviewers

The instructions name specific modules, invariants, and test files, so they need re-reading whenever those are refactored, renamed, or moved. The liveness test catches the mechanical half of that (a pointer that no longer resolves); it cannot catch an instruction that still resolves but no longer describes what the module does.

## Summary by CodeRabbit

* **Chores**
  * Updated repository review automation and configuration.
  * Expanded guidance for code areas, testing, documentation, workflows, and language-specific components.
  * Added checks to detect invalid paths, generated fixture issues, output regressions, and synchronization problems.

* **Tests**
  * Added automated validation for configuration path references, glob patterns, and documented repository paths.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,15 +1,17 @@
-# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+# yaml-language-server: $schema=https://storage.googleapis.com/coderabbit_public_assets/schema.v2.json
 #
 # CodeRabbit review configuration for ferro-hgvs (HGVS variant normalizer).
 #
 # This repo is output-stability-critical: parsing, normalization, and coordinate
-# projection must be idempotent, round-trip stable, and byte-identical to the
-# external oracles we conform to (biocommons hgvs, Mutalyzer, VariantValidator).
-# The `assertive` profile + the path instructions below aim the review at the
-# failure modes that actually bite here (output divergence, non-idempotent
-# normalization, parser/fast-path disagreement, coordinate off-by-one, panics on
-# adversarial input, spec-conformance drift) rather than generic style. CI
-# (proptests, fuzz targets, idempotency tests, conformance oracles) is the
+# projection must be idempotent, round-trip stable, and conformant with the
+# external oracles we validate against (biocommons hgvs, Mutalyzer,
+# VariantValidator). It also ships a PyO3 extension, a CLI, a web service, and a
+# reference-data preparation pipeline, each with its own failure modes. The
+# `assertive` profile + the path instructions below aim the review at what
+# actually bites here (output divergence, non-idempotent normalization,
+# parser/fast-path disagreement, coordinate off-by-one, panics on adversarial
+# input, spec-conformance drift, stale reference data) rather than generic style.
+# CI (proptests, fuzz targets, idempotency tests, conformance oracles) is the
 # primary correctness net; assertive makes CodeRabbit surface the borderline tier
 # that net can still miss — a false positive here is cheaper than a missed
 # output divergence.
@@ -18,6 +20,20 @@
 # CLI does NOT reliably inherit this file via `-c .coderabbit.yaml` (it can miss
 # findings the bot catches), so treat the CLI as a lighter pre-push smoke check,
 # not a substitute for the bot review.
+#
+# MAINTENANCE: the path instructions below name specific modules, invariants, and
+# test files. Re-read and update them whenever those are refactored, renamed, or
+# moved — a stale instruction silently misdirects the reviewer, which is worse
+# than no instruction at all. (The previous revision of this file pointed at
+# `tests/*.rs` paths after the integration suite had moved to `tests/it/`, so
+# every "confirm the test still covers it" pointer dangled.)
+#
+# That obligation is enforced, not just stated: `tests/it/coderabbit_config_paths.rs`
+# asserts every `path:` glob below matches at least one file in the repository, that
+# every repo-relative path named in backticks exists, and that no such path is split
+# across lines (the `>-` scalars fold a line break into a space, which would turn a
+# wrapped path into one that names nothing). A rename that orphans an instruction
+# fails the test suite instead of rotting silently.
 
 language: en-US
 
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1277, opened 2026-07-31
test(normalize): lock the dup-junction overlap contract end to end

### LINKED ISSUE #1276: Normalizer: an interior insertion normalized to a dup escapes overlap detection, and strict mode then accepts it
ferro reports an insertion strictly interior to a span edit as `W5002 OverlapConflictingEdits` (#486) and declines to resolve it. But normalization rewrites that interior insertion into a **duplication**, and `detect_insertion_overlaps` does not treat a `dup` as occupying an insertion junction — so the conflict it just detected becomes invisible in its own output.

## Reproducer

Synthetic coding fixture (`NM_TEST.1`, core `GGGCAATTGGGCCCAAATTTGGG`, `cds_start = 1`):

```
c.[4_10inv;5_6insAA]   strict: REJECTED (W5002)   lenient: c.[5_9inv;5_6dup]
c.[5_9inv;5_6dup]      strict: ACCEPTED           lenient: c.[5_6dup;5_9inv]
```

Two defects fall out of the same blindness:

1. **Lenient normalization launders a strict-rejected conflict.** The lenient normalizer takes an input its own strict mode rejects and emits a form its own strict mode *accepts*. A caller who normalizes leniently and re-validates strictly gets a clean bill of health for an allele with no well-defined resulting sequence.

2. **`normalize` is not idempotent on it.** `sort_cis_members_by_genomic_order` is deliberately skipped for overlap-conflicting alleles (`src/normalize/mod.rs:2563` — reordering is only meaning-preserving for disjoint members). Pass 1 sees the conflict and skips the sort; pass 2 sees no conflict and sorts. So `c.[4_10inv;5_6insAA]` needs two passes to settle, and `FERRO_ASSERT_IDEMPOTENT=1` fires on it.

Both reproduce on `main` — this is not introduced by #1243.

## Cause

`detect_insertion_overlaps` (`src/normalize/overlap.rs`) partitions members into `insertions` (only `NaEdit::Insertion`) and `spans` (`is_overlap_edit`, which includes `Duplication`). A duplication is registered *only* as a span. But `X_Ydup` inserts its copy at the junction after `Y`, so it is **both** a span and a junction occupant — and normalization is exactly what turns an interior `ins` into such a `dup`.

## Fix

Register a `Duplication` in `insertions` at `gap = end` in addition to its `spans` entry. It cannot self-flag: the interior test is `start <= gap && gap < end`, and a dup`s own junction is at `end`.

Verified locally: the full suite (8814 tests) passes unchanged with this applied, both symptoms are fixed, and the shapes that must stay accepted still are — `c.[4_6dup;8T>A]`, `c.[4_6dup;7_8insG]`, `c.[4_6dup;2_3insG]`, `c.[4_6dup;10_12dup]`. `c.[5_6dup;6_7insA]` becomes a (correct) new rejection: a dup copy and an insertion competing for one junction have no defined order.

Two details for whoever picks this up:

- Branch (a) of `detect_insertion_overlaps` hardcodes `edit_kinds` as `"ins"` for every member of a same-junction group. With duplications in that group the label is wrong; it should come from `edit_kind(&variants[idx])`.
- This is a user-visible strict-mode change (new `W5002` rejections for the dup shapes above), so it wants a release note.

## Follow-up

#1243 adds `overlap_conflicting_allele_is_not_canonicalized`, whose comment explains that it asserts on the settled form `c.[5_6dup;5_9inv]` because `c.[4_10inv;5_6insAA]` needs a second pass to settle this reordering. Once this is fixed that caveat is obsolete and the test can assert on the `insAA` form directly.

nh13:
Closes #1276.

## What changed about this PR

The source fix this PR carried **landed in #1243**, which registers a `dup` — and a `repeat`, which this PR never covered — as a junction occupant in `junction_writing_kind`. Rebasing onto that made the `overlap.rs` change redundant line for line, so it is dropped. Also dropped: the `NaEdit::Insertion => "ins"` arm this PR added to `edit_kind`, which is dead under #1243's design (kinds are carried on the occupant struct, so `edit_kind` is only ever called for spans) and whose comment named a caller that no longer exists.

What remains is the test file, and it is not redundant.

## Why the tests stay

#1243 tests the detector **directly**, in `overlap.rs`'s own unit tests, and asserts on the warnings it emits. These seven assert on the **caller-visible contract** those warnings exist to produce: what strict mode rejects, what lenient mode settles on, and that the two agree with each other.

Six of the seven were confirmed to fail with the `Duplication` arm of `junction_writing_kind` deleted; the seventh is the negative case and correctly does not move.

Three shapes are covered that #1243 does not reach:

- **a duplication interior to a `delins`**, not only to an `inv`;
- **the lenient-to-strict round trip.** Normalizing `c.[4_10inv;5_6insAA]` leniently produces `c.[5_9inv;5_6dup]`, and that output must still fail the strict check its own input failed. Before #1243 it passed — so a caller who normalized leniently and re-validated strictly got a clean bill of health for an allele with no defined resulting sequence;
- **the negative set** — a lone dup, disjoint dups, and siblings that miss the dup's junction all stay accepted, which is what keeps the interior bound strict at the 3' end.

Exercised on the `c.` axis: the `g.` spelling of the same shape is collapsed by the sequence-first canonicalization before the dup is ever rendered, which would mask the behaviour under test. Reference-free (`MockProvider` via `SyntheticBuilder`), so it holds independent of the manifest.

## Verification

8913/8913 pass, and again under `FERRO_ASSERT_IDEMPOTENT=1` + `FERRO_ASSERT_REPARSE=1`; `fmt` and `clippy --features dev --all-targets -D warnings` clean.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end regression coverage for detecting duplications that overlap insertion junctions.
  * Verified strict rejection for duplications overlapping inversions, delins, and insertions.
  * Confirmed lenient normalization remains idempotent and normalized results are still strictly rejected.
  * Verified non-overlapping duplications remain accepted.
  * Added diagnostic checks identifying conflicting duplication and insertion edits.

### DIFF
--- a/tests/it/issue_1276_dup_junction_overlap.rs
+++ b/tests/it/issue_1276_dup_junction_overlap.rs
@@ -1,0 +1,214 @@
+//! Issue #1276 — a duplication occupies an insertion junction, so it must be
+//! subject to the same overlap detection an insertion is.
+//!
+//! The fix itself landed with #1243, which registers a `dup` (and a `repeat`)
+//! as a junction occupant in `junction_writing_kind`. This file is the
+//! end-to-end lock for it: #1243 covers the detector directly, in
+//! `overlap.rs`'s own unit tests, and asserts on warnings; what is pinned here
+//! is the *caller-visible* contract those warnings exist to produce — what
+//! strict mode rejects, and what lenient mode settles on. Six of the seven
+//! below were confirmed to fail with the `Duplication` arm of
+//! `junction_writing_kind` deleted; the seventh is the negative case and
+//! correctly does not move.
+//!
+//! `x_ydup` places its copy at the junction after `y`. `overlap.rs` registered
+//! it only as a *span*, never as a junction occupant — and normalization is
+//! precisely what rewrites an interior insertion **into** a duplication. So the
+//! conflict `detect_insertion_overlaps` reported on the input went undetected on
+//! its own output, with two consequences pinned below:
+//!
+//!   1. lenient normalize turned a strict-**rejected** allele into one strict
+//!      mode **accepted**; and
+//!   2. `normalize` was not idempotent, because
+//!      `sort_cis_members_by_genomic_order` is skipped only for conflicting
+//!      alleles — so the member order flipped on the second pass.
+//!
+//! Exercised on the `c.` axis. The `g.` spelling of the same shape is currently
+//! collapsed by the sequence-first canonicalization before the dup is ever
+//! rendered, which would mask the behaviour under test; the transcript axes are
+//! not reached by that pass, so the dup survives and the detector is what
+//! decides. Reference-free (`MockProvider` via `SyntheticBuilder`), so this
+//! holds independent of the manifest.
+
+use crate::common::synthetic::SyntheticBuilder;
+use ferro_hgvs::error::FerroError;
+use ferro_hgvs::normalize::NormalizationWarning;
+use ferro_hgvs::reference::transcript::Strand;
+use ferro_hgvs::{parse_hgvs, MockProvider, NormalizeConfig, Normalizer};
+
+/// `c.4..10` is `CAATTGG` — no tandem repeat, so a `dup` here has no room to
+/// 3'-shift and these assertions are about overlap detection, not shuffling.
+const CORE: &str = "GGGCAATTGGGCCCAAATTTGGG";
+
+fn provider() -> MockProvider {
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1278, opened 2026-07-31
docs(review): add a CodeRabbit review checklist for recurring miss patterns

nh13:
Collects the patterns CodeRabbit has raised repeatedly on this repo into a single pre-push self-check at `.coderabbitai-review/checklist.md`. These are the misses that have actually cost review round-trips, not generic style advice.

Four entries, each naming the trigger, what to verify, and the PRs where it was caught:

- **New manifest field** — not threaded into `for_each_path` / `for_each_path_mut` / `deduplicate_paths`, or not existence-checked in `src/check/mod.rs`. A field that parses but is never enumerated silently skips validation and path rewriting. (#527, #530)
- **New error/warning code** — `related_codes` not cross-linked bidirectionally, or `docs/spec/error_code_audit.md` scope rows still describing the old behaviour. Audit-row drift is the usual miss. (#534)
- **New coordinate-system / axis token** — the axis set is recited in prose in many doc comments and `# Errors` sections, so adding a variant without sweeping them leaves the docs quietly wrong. (#543)
- **New normalize special-position path** — resolved endpoints not guarded against contig length and reversed order before the window-offset subtraction, where the `u64` arithmetic underflows and panics. (#526)

The corpus predates a fair amount of refactoring, so every path and symbol was re-verified against current `main` before writing: `src/prepare/manifest.rs`, `src/check/mod.rs` and `src/normalize/mod.rs` all still exist, the three path helpers are live in `manifest.rs`, `related_codes` now lives in `src/error_handling/codes.rs`, and the audit is at `docs/spec/error_code_audit.md`.

Complements `.coderabbit.yaml`, which tunes the review itself rather than listing what to check. Documentation only — nothing here is enforced by CI, and no code paths change.

## Summary by CodeRabbit

* **Documentation**
  * Added a repository-specific code review checklist covering manifest registration, error-code references, audit coverage, coordinate validation, and special-position bounds checks.
  * Added guidance to keep checklist references aligned with current code paths and symbols.

### DIFF
--- a/.coderabbitai-review/checklist.md
+++ b/.coderabbitai-review/checklist.md
@@ -1,0 +1,66 @@
+# CodeRabbit review checklist — ferro-hgvs
+
+Patterns CodeRabbit has flagged on this repository, mined from its review
+history. Use it as a pre-push self-check: these are the misses that have
+actually cost us review round-trips, not generic style advice.
+
+Each entry names the **trigger** (what you changed), what to **verify**, and the
+PRs where it was caught. A pattern earns a place here once CodeRabbit has raised
+it on this repository **and** the trigger generalises — the entry has to describe
+a class of future change, not a one-off bug. Every entry cites the PRs it was
+caught on, so its evidence is checkable; a recurrence adds a citation to the
+existing entry rather than a new entry.
+
+This complements `.coderabbit.yaml` at the repo root, which tunes the review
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #1280, opened 2026-07-31
Normalizer: repeat demotion places the duplication at the tract's 3'-most position, across a sibling it then corrupts


nh13:
## What's wrong

`demote_repeats_spanning_siblings` re-spells an insertion-grown repeat as a **duplication over the tract's 3'-most `length` bases** (`start = a.end - length + 1`). That placement is by construction the one that reaches furthest into the sibling, and when it covers a base a sibling edits, neither downstream clamp repairs it: a `Duplication` is invisible to `clamp_sibling_crossing_shifts` (it claims no bases), and `clamp_sibling_crossing_junctions` translates the whole span through `translate_member`, whose `landed` check verifies position and region but never that the bases under the new span are unchanged.

## Reproduction

3' shuffle, single-contig `TEMPLATE` provider:

```
(1) reference TTTTTTTTTAATATATTTTA      (9-T tract at 1-9)
    in   TEMPLATE:g.[2_3insTTTTT;5T>A]
    out  TEMPLATE:g.[5T>A;9_10insTTTTT]
      input denotes   TTTTTTTTTATTTTAATATATTTTA
      output denotes  TTTTATTTTTTTTTAATATATTTTA     <- substituted base moved

(2) reference GCGCGCGCGCTTTTTTTTTAATATATTTTA    (tract at 11-19)
    in   TEMPLATE:g.[12_13insTTTTT;15T>A]
    out  TEMPLATE:g.15delinsCTTTTA
      output denotes GCGCGCGCGCTTTTCTTTTATTTTAATATATTTTA
      — a `C` that appears nowhere in the sequence the input denotes
```

On `main` both emit an overlapping repeat (`g.[1_9T[14];5T>A]`) that the SPDI oracle declines, so this is a silent-vs-loud change of failure mode rather than a regression from correct.

## A fix was attempted and reverted — read this before trying again

The obvious repair is to demote an insertion-grown repeat back to an **`Insertion`** at the tract's 3' junction, carrying its literal payload, since an insertion claims no bases and cannot be corrupted by a sibling. That does fix both cases above.

It is wrong in general, and adversarial review caught it: the payload needs **rotating into phase** at the target junction. In a homopolymer that is invisible; in a tandem repeat it is not.

```
reference TTTTCAGCAGCAGCAGCAGCAGTTTT      ((CAG)x6 at 5-22)
in   TEMPLATE:g.[11_12insAGCAGC;5del]
[...]

Issue #1279, opened 2026-07-31
Normalizer: a 5'-shifting cis member sweeps through a duplication's source span, silently changing what it copies


nh13:
## What's wrong

A cis member's 5'-shift may sweep through a **duplication's source span**. The duplication claims no reference base, so `claims_reference_bases` reports false and `clamp_sibling_crossing_shifts` does not treat it as a barrier — but the bases under its span are exactly what it copies. A deletion landing on them changes the payload, and the result is well-formed, disjoint and warning-free.

## Reproduction

Reference `AGCTGGCGGTGCTAAAATCC` on the single-contig `TEMPLATE` provider, 5' shuffle:

```
in   TEMPLATE:g.[13_15dup;17del]
out  TEMPLATE:g.[14del;15_16insTAA]
  input denotes   AGCTGGCGGTGCTAATAAATCC
  output denotes  AGCTGGCGGTGCTATAAAATCC     <- different
```

Traced pass by pass: the corruption is already present after per-member normalization (`g.13_15dup | g.14del`), before any sibling pass runs. The `17del` 5'-shifted to `14del`, straight through the duplicated span `13-15`. The downstream passes then faithfully render an already-wrong member set.

Two more of the same shape:

```
ATAATTTAAAAATATTAAAA  g.[7_9dup;11del]   -> g.[8del;9_10insTAA]
AATTTTATAATTAATATAAA  g.[2_4dup;6del]    -> g.[3del;4_5insATT]
```

## Not a regression, but a change of failure mode

On `main` these emit `g.[13_15dup;14del]` — *overlapping* members, which the SPDI apply oracle and any consumer can reject. #1259 turns that into a silently wrong answer. Both are wrong; the silent one is more dangerous.

## A fix was attempted and reverted — read this before trying again

Adding a `blocks_sibling_shift` predicate (`claims_reference_bases` plus `Duplication`/`DupIns`) and using it in the 5' branch of `clamp_sibling_crossing_shifts` **does** fix this class: all four cases above become sequence-preserving, and the 74-case `dup`+`del`+5' residual pinned in `cis_junction_crossing_shift.rs` drops to **0**.

It was reverted because adversarial review found it introduces a worse defect on three-member alleles — the normalizer emits a description its own overlap detector rejects, from a clean input:

```
AATAAAAAATAAAATAATTA  g.[4dup;6dup;8del]  5'
[...]

Issue #1269, opened 2026-07-30
Normalizer: claims_reference_bases excludes NaEdit::Repeat, and two workarounds now sit on top of it


nh13:
## What's wrong

`claims_reference_bases` in `src/normalize/merge.rs` reports that a `NaEdit::Repeat` claims no reference bases:

```rust
fn claims_reference_bases(edit: &NaEdit) -> bool {
    matches!(
        edit,
        NaEdit::Substitution { .. }
            | NaEdit::Deletion { .. }
            | NaEdit::Delins { .. }
            | NaEdit::Inversion { .. }
    )
}
```

A repeat genuinely does claim the bases under its span — `g.1_9T[7]` covers positions 1–9 — so the predicate's name does not match its behaviour, and every pass that uses it to decide "can this member collide with that one" gets the wrong answer for repeats.

## Why it matters: two workarounds are now built on top of it

This predicate is the reason an entire defect class existed. Both the sibling-crossing clamp (#1259) and the earlier overlap clamp proposed in the closed #1236 skipped every repeat because of it, which is how a tract-wide repeat came to swallow a sibling (`g.[1_2del;4del]` → `g.[1_9T[7];9del]`).

#1259 did not correct the predicate. It added **two** passes that work around it:

1. `demote_repeats_spanning_siblings` — re-spell a repeat back to a plain edit before the clamp runs, so the clamp sees a span it can reason about.
2. `respell_colliding_duplications` — re-spell a colliding duplication as an insertion (#1263).

Each is individually justified and measured. Together they are evidence that the predicate is the thing that is wrong, and the next collision of this shape will want a third.

## Why it was not corrected in #1259

Changing it would make repeats act as barriers *everywhere* — in the clamps, in the junction rule, and in whatever else consults the predicate — with unmeasured blast radius across the repeat corpus (`del_shift_matrix.rs`, `repeat_tract_maximization.rs`, `issue_209_repeat_3prime_remaining.rs`, `issue_1210_repeat_gate_tract_span.rs`). That was the right call for a narrow fix and the wrong one to inherit silently.

## Suggested first step

Flip the predicate behind a measurement rather than a judgement: add `Repeat`/`MultiRepeat`, run the two committed sweeps plus the full suite, and see whether the two workaround passes become dead code or merely redundant. If the corpus survives, the workarounds can likely be deleted; if it does not, document why `Repeat` is excluded so the next person does not add a fourth workaround.

nh13:
[...]

PR #1285, opened 2026-07-31
fix(normalize): cis member ordering and a repositioned member's clamp

### LINKED ISSUE #1261: Normalizer: cis allele members emitted in descending coordinate order
## What's wrong

A normalized cis allele emits its members in **descending** coordinate order.

Reference: `("ACGT" x 64) + core + ("ACGT" x 64)` on `NC_TEST.1`, so `core[i]` is at 1-based position `257 + i`.

**core `CAGTATGCAGGCAA`:**

```
in   NC_TEST.1:g.[258_259insA;259_260insAG]
out  NC_TEST.1:g.[258_259dup;258dup]
```

`258dup` follows `258_259dup`, so the members descend.

## Severity: low — structural only

The sequence is **correct**: applied window `GTCAAGAGTATGCAGGCAAA` on both sides (20 bases from 0-based offset 254, computed independently of the normalizer via `spdi::hgvs_to_spdi`).

This is **not** an overlap. In interbase terms the two dups sit at 259 and 258 respectively, so they do not claim a shared position — an earlier reading called it "descending and overlapping", which the SPDI projection disproves. It is an ordering violation only.

It still matters because `sort_cis_members_by_genomic_order` (#1098) exists precisely to make the rendered order canonical and input-order-independent, and something downstream of it is putting these two back out of order.

## Reproduction

## Discovery

Found by a property test over cis alleles with length-changing members. The generator currently excludes insertion pairs at adjacent gaps (this shape, and #1260's) so the property can land green; that exclusion should be relaxed as these are fixed.

### LINKED ISSUE #1281: Normalizer: a 3-member cis allele 5'-shuffles to duplicate members, dropping a substitution (pre-existing)
## What's wrong

A three-member cis allele (two deletions plus a substitution) normalizes under 5' shuffle to a description with **two textually identical members**, the substitution silently gone, and no well-defined resulting sequence.

## Reproduction

Single-contig `TEMPLATE` provider, 5' shuffle:

```
reference TTTTTTTTTAATATATTTTAATAT
in   TEMPLATE:g.[1del;9T>A;11del]
out  TEMPLATE:g.[1del;1del]           <- `9T>A` gone, `11del` became a second `1del`

reference TTATTTAAATAAAAATAAAATTTT
in   TEMPLATE:g.[4del;6T>A;8del]
out  TEMPLATE:g.[4del;4del]
```

`parse_hgvs` **accepts** the output, so `FERRO_ASSERT_REPARSE` does not catch it. The SPDI apply oracle correctly declines it (`None`, no well-defined sequence) while the input applies fine.

## Pre-existing, not from #1259

Verified identical on `main` (`7fed61d`) and on the #1259 branch — same input, same `g.[1del;1del]` output. #1259 in fact *improves* the 3' direction of the same input (`g.1_9T[7]`, sequence-preserving, where `main` gives the broken `g.[9T>A;9del;11del]`).

## Why no test catches it

Every sweep in the sibling-crossing test files builds exactly **two** members (`first`/`second` loops). This shape needs three. That gap is #1268; this issue is the concrete defect behind it.

An ad-hoc three-member sweep (~355K generated alleles) surfaced at least 15 distinct seed/position combinations producing the same duplicate-member shape under 5' shuffle.

## Suspected area

The interaction between the per-member snapshots (`pre`/`post`, computed once per call) and the merge/coalesce step when more than two cis members are present in one pass — two originally-distinct members appear able to collapse onto the same position without the merge noticing they are now identical.

nh13:
> Stacked on #1259. Base retargets to `main` when that merges.

Closes #1261. Closes #1281.

Two of the four defects standing between #1235 and closure, plus the confluence-generator extension #1260/#1261 ask for. #1235 itself stays open — see [What is left](#what-is-left).

## #1261 — members rendered in descending order

`sort_cis_members_by_genomic_order` (#1098) keys on `cis_member_order_key`, whose only positional discriminator was the member's **start**. When two members shared a start it fell through to a lexical compare of the rendered descriptor, which is arbitrary with respect to where those members actually sit:

```
reference   ("ACGT" x 64) + "CAGTATGCAGGCAA" + ("ACGT" x 64)
in   g.[258_259insA;259_260insAG]
out  g.[258_259dup;258dup]          <- descending
```

Both start at 258, so every positional field tied and `"258_259dup" < "258dup"` decided it (`_` is 0x5F, `d` is 0x64). A duplication places its copy at the junction *after* its last base, so those two occupy interbase 259 and 258 — ordered by their **end**, the one field the key omitted. The same blind spot #1276/#1277 found in the overlap detector, one layer up.

The fix adds the end to the key, on every axis including protein — `p.[Ala3dup;Ala3_Ala4dup]` has the identical shared-start shape and is covered by its own test. It cannot reorder members whose spans do not overlap: there `a < c` implies `b < c`, so start-, end- and junction-order already agree. Overlapping members never reach the sort — `normalize_allele` skips it when an `OverlapConflict` was warned. The descriptor stays the final tie-break, so the key remains total and the result input-order-independent.

Verified the sort was genuinely running first (output is input-order-independent), so this is the tie-break and not a skipped pass.

## #1281 — two members claiming one base

Reported as a three-member allele losing a substitution. It needs only **two**, and three is simply the shortest route from clean input — the adjacent sub/del pair has to merge before it can reduce:

```
reference             TTTTTTTTTAATATATTTTAATAT   nine-T tract at 1-9
g.[1del;9_10delinsA]  ->  g.[1del;1del]
g.[1del;9T>A;11del]   ->  g.[1del;1del]          <- as filed
```

`parse_hgvs` accepts that, so `FERRO_ASSERT_REPARSE` does not catch it; the SPDI apply oracle declines it while the input applies fine.

**Two causes, both required.** Traced with per-pass instrumentation:

```
pre_collapsed  1del | 9_10delinsA
merged_split   1del | 1del          <- the move happens here
```

1. **The snapshot.** `canonical_split_for_variant` (the #1237 sequence-first canonicalization) acts on one member with no sibling context, and it *repositions* rather than merely re-spells: `g.9_10delinsA` reduces to `g.1del`, an eight-position move. The sibling passes took their "before" snapshot **after** that step, so they saw `1del -> 1del`, concluded nothing had moved, and left the member where it landed. Now each surviving member records its origin, and **all three** sibling passes read it — `demote_repeats_spanning_siblings` and `clamp_sibling_crossing_junctions` had the same stale-snapshot exposure, so fixing only the span clamp would have left the junction clamp measuring from wherever the canonicalization put the member. A one-for-one rewrite is a repositioning of the same member, so its origin span is what the clamp needs; a genuine split into several pieces has no single origin span per piece, so each piece stands for itself exactly as before.

2. **The guard.** `clamp_sibling_crossing_shifts` admitted only a pure translation. This member shrinks as it moves — delta is −8 at the start, −9 at the end — so it was refused even once the snapshot showed the movement. Admit a member that shrank while its start moved; it still sweeps a window this pass can bound, and the existing cap keeps the pull inside it. A member that **grew** is still refused: an insertion canonicalising to a duplication moves its span and its junction together, and bounding the span mis-places the copy (#1266, #1279).

Clamped at the sibling the member lands merely adjacent, and the next pass coalesces the pair into the correct single deletion — `g.[1del;2del]` -> `g.1_2del`.

## Confluence generator: length-changing members

The #1238 property test is substitution-only and so length-preserving, which put the shifting machinery out of reach entirely (`shift_pieces_three_prime` only moves a piece for which `is_pure_indel()` holds). That is the exclusion #1260 and #1261 record as needing relaxation. `IndelHaplotype` adds single-base deletions and one/two-base insertions.

It found two defects on its first two runs, neither previously filed and neither reachable by any committed sweep:

```
core "AAAAAA", insert A after index 1 and after index 2
  -> g.[263dup;263dup]             two members claiming one base
core "ATACAGAAAATCAGGGCATA", insert GA after 4, AA after 6
  -> g.[262_263dup;263_266A[6]]    dup and repeat spans overlap
```

Both are the `claims_reference_bases` family (#1269): a duplication and a repeat report claiming no bases, so the sibling clamps skip them and two members settle onto one span. Filed as **#1286** (two dup members at one junction) and **#1287** (a dup's junction inside a repeat's tract span) — separate issues because the mechanisms differ, and #1287 specifically implicates #1259's `demote_repeats_spanning_siblings`.

Consequently the property that catches them is committed `#[ignore]`d with both reproductions in its doc comment; enabling it is the acceptance criterion for that work. Active instead: sequence-preservation, fixed-point and disjoint/ascending over the separate-member spelling, authored-order independence, and a non-vacuity guard that the model really produces deletions, insertions, mixed haplotypes and multi-base payloads.

Neither restriction is silent — `adjacent_gap_insertions_are_a_known_gap` pins both held-back shapes as currently-broken and drives #1260's through the model itself, so fixing either fails loudly and names what to re-enable. That is #1268/#1283's complaint about coverage reading wider than it is.

## What is left

**#1260 and #1262 are one cause, not two, and it is not local.** Both issues filed hypotheses they noted could not be checked against `main`. Traced now that #1237 has merged, both are confirmed and both reduce to the same thing:

- The derivation reaches the **same piece set from both spellings** — it is already confluent.
- The divergence is entirely in the **input-relative gates**. #1260 refuses via the column bound (`3 > 2` for the two-insertion spelling; `3 > 3` admits the delins one); #1262 refuses via the separator veto, whose `pieces.len() < edits.len()` precondition is never true for a one-member spelling, so a single delins is never checked at all.
- The derivation produces the **over-merged** form. The gates are the only thing rescuing the split form, and only when the author happened to write it that way. So making the gates spelling-independent does not fix this — it picks the over-merged answer for everyone.

It is also not two edge shapes: the generator hits it constantly, including on plain `del`+`ins` pairs, because it is the general behaviour of "length-changing members versus the spanning delins".

Fixing it means making the derivation prefer the split form: `general.md:56` prioritisation (substitution > deletion > delins) for #1262, and for #1260 a **two-gap** alignment — insertion, retained base, insertion — which `best_alignment`'s single-gap restriction cannot express. That restriction and the "stays delins" corpus (#1034, #1040, #182, #422) are documented as deliberate and load-bearing, so this belongs in its own PR with its own measurement.

## Review pass

A CodeRabbit-style self-review was applied before this was opened; nine findings, all fixed, amended into the commit each belongs to (no fixup commits). The substantive ones:

- **The origin snapshot now reaches all three sibling passes**, not just the span clamp — see #1281 above. Measured: suite green and the pinned 74-case residual unmoved.
- **The protein axis gets its real end.** `cis_member_range` had been returning the start twice for `HgvsVariant::Protein`, so the fix reached six of seven arms and protein kept the defect. Verified non-vacuous — reverting that one line makes `p.[Ala3dup;Ala3_Ala4dup]` render descending.
- **`cis_member_order_key` returns `(String, SelfCancellingPoint, SelfCancellingPoint, String)`** rather than a flattened eight-field tuple. `SelfCancellingPoint` already derives `Ord` over `(region, base, offset)`, so the sort order is identical and the call site is legible.
- **A weak test assertion replaced.** One #1281 case asserted `!output.contains("4del;4del")` — true of any *other* malformed pair too. Now pins `g.4_5del`.
- **No member-list clone on the non-cis path**, which never reads the snapshot.
- Plus four nitpicks: shared member rendering between the two encoding helpers, `Event::Insert`'s payload changed from `([char; 2], len)` (which made an out-of-bounds `len` constructible) to `&'static str`, a shadowed binding, and a stale module-doc opener.

Each of the four commits builds and tests standalone, so bisect still works.

## Verification

- 8893 pass; 8893 pass again under `FERRO_ASSERT_IDEMPOTENT=1` with `FERRO_ASSERT_REPARSE=1`.
- The `#[ignore]`d property still reproduces #1287 when run with `-- --ignored`, so the pin is live rather than quietly passing.
- The 276,480-case sweep in `cis_junction_crossing_shift` holds its pinned 74-case #1266 residual **exactly**, so no previously-correct output became wrong. That is the measurement the attempted fixes in #1279 and #1280 failed, and the reason the clamp relaxation here is restricted to members that shrank.
- All four commits are signed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cis-allele normalization across genomic, mitochondrial, transcript, RNA, CDS, and protein coordinates.
  * Corrected ordering for variants sharing the same start position.
  * Prevented shifted variants from crossing sibling spans or junctions, including shrinking spans.
  * Improved repeat, duplication, insertion, and deletion normalization.
  * Preserved applied sequence results through complex normalization scenarios.

* **Tests**
  * Added extensive regression, confluence, ordering, and sequence-preservation coverage for cis-allele normalization.
  * Added coverage for protein variants, adjacent gaps, spanning indels, and multi-base insertions.

### DIFF
--- a/src/hgvs/variant.rs
+++ b/src/hgvs/variant.rs
@@ -950,7 +950,7 @@ enum SelfCancellingAxis {
 /// The genome/mito axes never carry a region flag and rarely an
 /// offset, but we include both for uniformity.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-struct SelfCancellingPoint {
+pub(crate) struct SelfCancellingPoint {
     region: bool,
     base: i64,
     offset: i64,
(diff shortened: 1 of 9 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed the outstanding review threads, plus the two outside-diff findings:

- **`issue_1261_cis_member_order.rs`** — `assert_ordered_and_preserving` now unwraps both sides of the apply oracle, so a description with no well-defined resulting sequence fails instead of passing as `None == None`; dropped the duplicated `assert_members_ascending` call.
- **`issue_1281_reducing_member_shift.rs`** — header now names both settled spellings (`g.1_2del` under the 5' shuffle these cases use, `g.1_9T[7]` for the adjacent pair fed in directly and shuffled 3'), matching the pinned expectations. The local 5' preservation check is gone: `assert_normalizes_preserving_in(seq, input, expected, direction)` now lives beside `assert_normalizes_preserving` in the oracle. The copies in `issue_1254_sibling_crossing_shift.rs` and `cis_junction_crossing_shift.rs` are untouched — this PR does not otherwise touch those files, and several in-flight PRs edit them.
- **merge.rs 3152-3169 (rotate literal insertion payloads)** — out of scope here; that is #1292, already open as #1298.
- **merge.rs 3481-3486 (terminal duplication respelling)** — confirmed real and reproduced: `g.[24dup;24C>G]` on a 24-base contig emits `g.[24C>G;24_25insC]`. Verified pre-existing on `origin/main`, in a region this PR does not touch, so filed as #1307 rather than widened into this PR.

Also, from a self-review pass: the `#[ignore]`d `an_indel_haplotype_normalizes_to_its_own_sequence` had no *running* pin, so fixing #1286 or #1287 would have left it ignored by inattention. Added `the_ignored_indel_property_still_finds_its_two_defects`, which asserts on the shape of each defect (two members on one span; a junction inside a sibling's span) and names the ignore to drop. Verified non-vacuous — it reproduces `g.[263dup;263dup]` and `g.[262_263dup;263_266A[6]]` exactly.

8904/8904 tests pass; clippy and fmt clean.

nh13:
@coderabbitai review

Issue #1287, opened 2026-07-31
Normalizer: a duplication's junction lands inside a sibling repeat's tract span, emitting overlapping members


nh13:
## What's wrong

A duplication's junction lands **inside** a sibling's repeat span. The two members overlap, and the pair has no well-defined resulting sequence.

```
reference  ("ACGT" x 64) + "ATACAGAAAATCAGGGCATA" + ("ACGT" x 64)
           so core[i] is at 1-based position 257 + i

in   NC_TEST.1:g.[261_262insGA;263_264insAA]
out  NC_TEST.1:g.[262_263dup;263_266A[6]]
```

Interbase spans of the output are `(263, 263)` and `(262, 266)` — the duplication's junction at 263 sits inside the repeat's span 262-266.

## Severity

**Silent.** `parse_hgvs` **accepts** the output, so `FERRO_ASSERT_REPARSE` does not catch it. The SPDI apply oracle declines it while the input applies fine:

```
apply(in)  = ...GTACGTATACAGAGAAAAAATCAGGGCATA...
apply(out) = None
```

## Suspected mechanism

Same root as #1286 and the same family as #1269: `claims_reference_bases` reports that neither a `Duplication` nor a `Repeat` claims reference bases, so neither is a barrier to the other.

#1259 added `demote_repeats_spanning_siblings` for exactly this shape — a repeat's tract-wide span swallowing a sibling — but the demotion is driven by whether the *sibling* claims bases. A duplication does not, by that predicate, so this pair is skipped and the repeat keeps a span that contains the dup's junction.

`clamp_sibling_crossing_junctions` does not save it either: it bounds a junction against a sibling's **bases**, and a `Repeat` is not reported as claiming any.

## Why file separately from #1286

Different mechanism, same predicate. #1286 is two zero-width members colliding *at one junction*; this is one member's junction falling inside another member's *span*, where the span belongs to a repeat whose width came from tract-wide canonicalisation rather than from the input. A fix for one need not fix the other, and #1259's demotion pass is specifically implicated here.

## How it was found

By the `IndelHaplotype` confluence model added in #1285, on its second run. Unreachable by any committed sweep — they generate only `{Ndel, N<ref>><alt>}` second members and 20 bp random cores with no tract long enough to canonicalise to a non-homopolymer `Repeat` (gaps 1 and 4 in #1283).

## Status

[...]

Issue #1286, opened 2026-07-31
Normalizer: two insertions at adjacent gaps in a homopolymer emit two identical dup members claiming one junction


nh13:
## What's wrong

Two insertions at adjacent gaps inside a homopolymer both canonicalise to a duplication at the **same junction**, and the normalizer emits both — two textually identical members claiming one position.

```
reference  ("ACGT" x 64) + "AAAAAA" + ("ACGT" x 64)
           so the poly-A run is 257-263 (seven bases: the padding
           resumes its ACGT phase with an A)

in   NC_TEST.1:g.[258_259insA;259_260insA]
out  NC_TEST.1:g.[263dup;263dup]
```

Both members project to interbase 263, so they overlap. The input inserts two `A` into a seven-`A` run, giving nine; the correct output is one member — `g.262_263dup`, or the repeat spelling `g.257_263A[9]`.

## Severity

**Silent.** `parse_hgvs` **accepts** `g.[263dup;263dup]`, so `FERRO_ASSERT_REPARSE` does not catch it. The SPDI apply oracle correctly declines it (`None`, no well-defined resulting sequence) while the input applies fine:

```
apply(in)  = ...GTACGTAAAAAAAAACGTACGT...   (nine A)
apply(out) = None
```

So a consumer gets a well-formed-looking description with no defined meaning, and no oracle in the suite fires.

## Suspected mechanism

The `claims_reference_bases` family (#1269). A `Duplication` reports that it claims no reference bases, so `clamp_sibling_crossing_shifts` does not treat either member as a barrier to the other, and both are free to settle on the tract's 3'-most junction. Nothing downstream notices that two zero-width members now occupy the same interbase.

Distinct from #1276/#1277: that one is about a duplication interior to another member's *span* being missed by the overlap **detector**. Here both members are duplications and the collision is at a **junction**, and the question is why they were allowed to land there at all, not merely why it went unreported.

## How it was found

By the `IndelHaplotype` confluence model added in #1285 — it was the first failure on the first run. No committed sweep reaches this shape: the sibling sweeps in `cis_junction_crossing_shift.rs`, `issue_1254_sibling_crossing_shift.rs` and `repeat_span_sibling_overlap.rs` only ever generate `{Ndel, N<ref>><alt>}` as the second member, never a `dup` or `ins` (the first gap listed in #1283).

## Status

[...]

Issue #1282, opened 2026-07-31
Normalizer: 5'-shift to position 1 panics in hgvs_pos_to_index (subtract with overflow) instead of returning an error


nh13:
## What's wrong

`hgvs_pos_to_index` computes `(pos - 1) as usize` without a lower-bound check (`src/coords/mod.rs:432`). A 5'-shifting cis member that reaches position 1 makes `pos` zero, the subtraction underflows, and a **debug build panics** inside `Normalizer::normalize` rather than returning an error.

## Reproduction

```
reference   TTTTTTTTTAATATATTTTA
input       TEMPLATE:g.[3_4insT;1T>A]
direction   ShuffleDirection::FivePrime

  thread panicked: attempt to subtract with overflow
```

Single-contig `TEMPLATE` provider (`MockProvider::add_genomic_sequence`).

## Severity

`normalize` is a library entry point with a `Result` return; a panic is a different failure class from an `Err`, and a caller cannot catch it without `catch_unwind`. In release the subtraction wraps instead, which is worse — it silently produces a garbage index rather than failing.

## Pre-existing

Not introduced by #1259. Found while adversarially reviewing that branch: **23 cases, identical count on the PR branch and on its base**, so it predates the sibling-crossing work entirely. Reported here because it was out of scope there and would otherwise be lost with the session.

## Suggested direction

Guard the conversion (return `None`/`Err` for `pos < 1`) rather than relying on callers never producing a zero. Worth a sweep afterwards for other `as usize` casts on HGVS positions in the same module — the same shape may exist for the 5' boundary elsewhere.

Issue #1267, opened 2026-07-30
Normalizer: 5'-shifted insertion junction crosses an upstream sibling (and the sweeps cannot see it)


nh13:
## What's wrong

Under `ShuffleDirection::FivePrime`, an insertion-like member's junction travels 5' toward an **upstream** sibling and crosses it, changing what the allele denotes.

```
reference        ACAAAAAAAACGTACGTACG        A-run at 3-10
g.[4A>G;9dup]    applies to  ACAGAAAAAAACGTACGTACG
normalizes to    g.4_5insG   ACAAGAAAAAACGTACGTACG   ← the substituted base moved
```

Two more of the same shape:

```
g.[4A>G;9_10insA]  ->  g.4_5insG
g.[5A>G;10dup]     ->  g.[3dup;5A>G]
```

Well-formed, disjoint, warning-free, wrong — the silent shape.

## Why #1259 does not fix it

#1259 added a junction clamp, but **3'-only**. A 5' mirror was written and measured to turn 80 previously-correct outputs into silently wrong ones (a duplication's span and its junction move together under a 5' shift, so bounding the junction mis-places the copy), and was removed as defective rather than deferred.

Separately, that mirror would not have caught this anyway: it bounded the junction against a sibling the member was moving *away* from, not one it was moving *toward*.

## The test-coverage angle, which is the more important half

**The sweeps in #1259 structurally cannot see this.** They always place the sibling *downstream* of the first member, so under 5' shuffle the member travels away from its sibling and can never cross it. The three cases above were found by hand, not by the sweep.

That blind spot is worth closing independently of the fix — a sweep that cannot reach a known-broken shape will not notice it regressing further.

## Status

Pre-existing; reproduces identically on `main` and on #1259, so that PR neither fixes nor worsens it. Pinned as currently-wrong by `a_five_prime_junction_with_an_upstream_sibling_is_a_known_gap` in `tests/it/cis_junction_crossing_shift.rs`, which also records what to widen in the sweep when it is fixed.

PR #1288, opened 2026-07-31
fix(normalize): make a duplication a barrier to a sibling's shift

### LINKED ISSUE #1266: Normalizer: 5'-shuffled duplication beside a deletion cancels into a wrong shorter duplication
## What's wrong

Under `ShuffleDirection::FivePrime`, a duplication beside a deletion cancels wrongly: the merge cancels part of the duplicated sequence against the deletion and emits a **shorter duplication in the wrong place**.

```
reference        TTATTTAAATAAAAATAAAA
g.[6_7dup;9del]  applies to  TTATTTATAATAAAAATAAAA
normalizes to    g.4dup      TTATTTTAAATAAAAATAAAA   ← different bases
```

`7_8insT` is a correct spelling; `g.4dup` is not.

## Scope, measured

Across 276,480 two-member cis alleles (`del`/`dup`/`ins` first member × `del`/`sub` sibling, both shuffle directions), applying both sides to the reference independently of the normalizer:

| shape | sequence-changing outputs |
| --- | --- |
| everything else | **0** |
| `dup` + `del` sibling, 5' shuffle | **74** |

So this is the entire remaining residual of that sweep. Pre-existing — the 74 are a strict subset of what already failed before #1259, verified by diffing the failing-case sets.

## Why #1259 does not fix it

Distinct mechanism. #1259's clamps prevent a member *crossing* a sibling; this is the merge's **cancellation path** producing a wrong result once the members are adjacent. Three separate attempts at a 5'-direction clamp were measured during #1259 and all made things worse — one turned 80 previously-correct outputs into silently wrong ones — because a duplication's span and its junction move together under a 5' shift, so bounding the junction mis-places the copy.

The related #1263 was fixed by re-spelling rather than repositioning (`Xdup` → `X_X+1ins<ref[X]>`, which claims no base). That may be the shape of the fix here too, but it has not been tried.

## Status

Pinned as currently-wrong by `a_five_prime_duplication_beside_a_deletion_is_a_known_residual` in `tests/it/cis_junction_crossing_shift.rs` (#1259), and excluded from that file's sweep assertion by exactly this shape. Fixing it fails that test loudly and the exclusion should be removed at the same time.

### LINKED ISSUE #1279: Normalizer: a 5'-shifting cis member sweeps through a duplication's source span, silently changing what it copies

nh13:
> Stacked on #1285 (which is stacked on #1259). Base retargets as those merge.

Closes #1266. Closes #1279.

A duplication reads its payload from the reference under its own span, so a sibling landing on those bases changes what it copies. `claims_reference_bases` reports that it claims nothing, so `clamp_sibling_crossing_shifts` neither treated a duplication as a barrier nor clamped a duplication's own shift.

```
reference        TTATTTAAATAAAAATAAAA
g.[6_7dup;9del]  applies to  TTATTTATAATAAAAATAAAA
normalized to    g.4dup      TTATTTTAAATAAAAATAAAA   <- different    #1266

reference        AGCTGGCGGTGCTAAAATCC
g.[13_15dup;17del] -> g.[14del;15_16insTAA]          <- different    #1279

reference        ACAAAAAAAACGTACGTACG
g.[4A>G;9dup]    -> g.4_5insG                        <- moved base   #1267 (part)
```

All well-formed, disjoint, warning-free and wrong — the silent shape.

## Fix

`blocks_sibling_shift` = `claims_reference_bases` + `Duplication` + `DupIns`, used in `clamp_sibling_crossing_shifts` for the 5' sibling barrier **and** for admitting the shifting member itself.

**Both halves are required, and that is why the attempt recorded in #1279 failed.** That one added the predicate to the sibling barrier only. It was reverted because `g.[4dup;6dup;8del]` then emitted `g.[4dup;4dup;5del]`, which ferro's own overlap detector rejects. The cause was the outer guard: a member had to *claim bases* to be considered at all, so a duplication's own shift was never clamped and two duplications were free to settle onto one another. Admitting them closes it — that input now yields `g.[4dup;5dup;6del]`, disjoint and sequence-preserving.

I re-ran #1279's exact regression case to confirm this, rather than assuming the earlier revert no longer applied.

## The sweep goes to zero

`no_two_member_allele_normalizes_to_a_different_sequence` covers 276,480 two-member alleles and pinned an exact **74-case** sequence-changing residual — the whole of #1266, and roughly a twelfth of the sweep. That count is now **0**.

So three pins change together, which is exactly what they were written to force:

| Test | Before | After |
|---|---|---|
| the sweep's residual constant | `= 74` | `= 0` — no case in the sweep changes the sequence |
| `a_five_prime_duplication_beside_a_deletion_is_a_known_residual` | asserted `g.4dup` (wrong) | renamed `…_keeps_its_sequence`, asserts `g.7_8insT` |
| `a_five_prime_junction_with_an_upstream_sibling_is_a_known_gap` | asserted 3 wrong outputs | split — see below |

## #1267 is only partly reached, and stays open

The predicate covers a duplication because the bases under its span **are** its payload. A pure `Insertion` carries a literal and has a zero-width span, so there is nothing for a sibling to land on — `g.[4A>G;9_10insA]` is still wrong. That case needs a 5' *junction* barrier, and #1259 measured one turning 80 previously-correct outputs into silently wrong ones, so it is not something to reach for here.

The old known-gap test is therefore split: `a_five_prime_duplication_does_not_cross_an_upstream_sibling` asserts the two `dup` cases as fixed, and `a_five_prime_insertion_junction_with_an_upstream_sibling_is_a_known_gap` pins the insertion case as still broken with the reasoning above. #1267 stays open, narrowed to that one shape.

## Deliberately not addressed

Measured, not assumed — each was probed against this branch:

- **#1280** — 2 of its 3 cases still change the sequence. It is the 3'-direction repeat-demotion path, not this clamp.
- **#1286 / #1287** — still broken. #1286 is two insertions that should *merge* into one two-base dup at a shared junction, not be clamped apart; #1287 needs `Repeat` in the predicate, which is #1269's question and carries the blast radius that issue describes.
- **#1282** — an unrelated missing lower-bound check in `hgvs_pos_to_index`; still panics.

## Verification

- 8894 pass; 8894 pass again under `FERRO_ASSERT_IDEMPOTENT=1` with `FERRO_ASSERT_REPARSE=1`.
- The 276,480-case sweep passes with its residual pinned at 0.
- Commit is signed.

Not run: the manifest-backed conformance axis tests — no reference volume mounted locally, so they skip. Worth watching the nightly, since this changes output for a shape the conformance corpora do contain.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -3014,6 +3014,9 @@ struct MemberSpan {
     end: i64,
     /// Whether the edit consumes the reference bases under its span.
     claims_bases: bool,
+    /// Whether the edit must stop a sibling's shift at its boundary. Superset of
+    /// `claims_bases` — see `blocks_sibling_shift`.
+    blocks_shift: bool,
     /// For an insertion-like edit, the position its added sequence sits
     /// immediately 3' of; `None` for anything that consumes bases instead.
     junction: Option<i64>,
@@ -3030,6 +3033,7 @@ fn member_span(v: &HgvsVariant, kind: CisKind) -> Option<MemberSpan> {
         start,
         end,
         claims_bases: claims_reference_bases(edit),
+        blocks_shift: blocks_sibling_shift(edit),
         junction: junction_of(edit, start, end),
     })
 }
@@ -3045,8 +3049,10 @@ fn member_span(v: &HgvsVariant, kind: CisKind) -> Option<MemberSpan> {
 /// the same variant clamped or not depending on whether the reference base
 /// happened to be written out. Insertions and duplications add sequence at a junction without
 /// consuming a base, so a member landing flush against one is the *adjacency*
-/// the collapse pass is meant to catch (#999, #1135) — they are neither clamped
-/// nor treated as barriers.
+/// the collapse pass is meant to catch (#999, #1135) — they are not clamped
+/// here. A duplication is nonetheless a 5' barrier, because the bases under its
+/// span are what it copies; `blocks_sibling_shift` carries that, keeping this
+/// predicate to the narrower "consumes bases" question its name asks.
 ///
 /// `Repeat`/`MultiRepeat` are the one deliberate omission: a repeat does claim
 /// the bases under its tract, but making it say so changes clamp and barrier
@@ -3062,6 +3068,22 @@ fn claims_reference_bases(edit: &NaEdit) -> bool {
             | NaEdit::Delins { .. }
             | NaEdit::Inversion { .. }
     )
+}
+
+/// Whether `edit` must stop a *sibling's* shift at its boundary, even though it
+/// may not consume the bases under its span.
+///
+/// EXPERIMENT (#1279/#1286): `claims_reference_bases` plus `Duplication`/
+/// `DupIns`. A duplication reads its payload from the reference under its own
+/// span, so a sibling deletion landing on those bases changes what the
+/// duplication copies — it is a barrier even though it claims nothing.
+///
+/// Used for the *5'* half of `clamp_sibling_crossing_shifts` only; the 3' half
+/// deliberately stays on `claims_reference_bases`. See the comment on its
+/// `onto_bases` filter for why.
+fn blocks_sibling_shift(edit: &NaEdit) -> bool {
+    claims_reference_bases(edit)
+        || matches!(edit, NaEdit::Duplication { .. } | NaEdit::DupIns { .. })
 }
 
 /// Pull back any cis member whose shift carried it over a sibling's bases.
@@ -3120,7 +3142,7 @@ pub(crate) fn clamp_sibling_crossing_shifts(
         let (Some(b), Some(a)) = (pre[i].as_ref(), post[i].as_ref()) else {
             continue;
         };
-        if !b.claims_bases || !a.claims_bases || b.region != a.region {
+        if !b.blocks_shift || !a.blocks_shift || b.region != a.region {
             continue;
         }
         // A member that did not move cannot have crossed anything.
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1289, opened 2026-07-31
fix(normalize): merge cis members that settle on one junction

### LINKED ISSUE #1286: Normalizer: two insertions at adjacent gaps in a homopolymer emit two identical dup members claiming one junction

nh13:
> Stacked on #1288 → #1285 → #1259. Base retargets as those merge.

Closes #1286.

```
reference  ("ACGT" x 64) + "AAAAAA" + ("ACGT" x 64)   poly-A run at 257-263
g.[258_259insA;259_260insA]  ->  g.[263dup;263dup]
```

Two members claiming one position have no well-defined resulting sequence. `parse_hgvs` **accepts** the output, so `FERRO_ASSERT_REPARSE` does not catch it; the SPDI apply oracle declines it while the input applies fine.

## Why neither existing layer catches it

**Nothing upstream prevents it.** The sibling clamps bound a member against a sibling's *bases*. A member that consumes no base occupies the interbase junction where its payload lands, and neither of these has bases for the other to land on — so both are free to travel to the tract's 3'-most junction and settle there together.

**Nothing downstream repairs it.** The sequence-first canonicalization would derive the correct single member, but it runs *after* the per-member pipeline and derives from the allele's **resulting sequence** — which an overlapping allele does not have. I traced this rather than assuming: the pass is entered with `[263dup;263dup]` and exits without deriving. So the corruption survives the one pass that would have merged it.

That is the shape of the bug: a catch-22 between the layer that could have prevented it and the layer that could have fixed it.

## Fix

`coalesce_members_at_one_junction`, running last in the sibling-repair sequence in `normalize_allele`. Both members add their payload at the same interbase point, so one insertion carrying the concatenation denotes exactly what the pair denoted.

Re-spelling as an `Insertion` rather than a wider `Duplication` is the choice `respell_colliding_duplications` already makes: an insertion's payload is a literal and is correct wherever it lands, whereas `g.262_263dup` would be equivalent only when `ref[262] == ref[263]`. The next pass re-canonicalises it where the reference permits — here to `g.257_263A[9]`, the nine-A run.

**Merged only when every colliding payload is identical.** Concatenation is otherwise order-dependent, and the order is exactly what the shift destroyed: two insertions at distinct junctions had a defined order, and once they share a junction they do not. Equal payloads make the concatenation order-independent, so the repair is unambiguous. A differing-payload collision is left for the overlap detector to report rather than guessed at — if one turns out to be reachable, that is a separate issue with a separate answer.

Genomic axes only, matching `respell_colliding_duplications`; the transcript axes need a coordinate conversion this does not carry.

## Tests

`tests/it/issue_1286_shared_junction_merge.rs`, 5 tests, generated programmatically. The reproduction, order-independence, a **three**-member group (the pass is not limited to pairs), and two guards: insertions at distinct junctions stay two members, and a lone insertion is untouched.

The sequence check goes through `hgvs_to_spdi` — a different projection and applier from the normalizer's own — so a canonicalization that silently altered the sequence shows up even when the string looks plausible.

## Not addressed

**#1287 is still open** and was probed against this branch: `g.[261_262insGA;263_264insAA]` → `g.[262_263dup;263_266A[6]]`, unchanged. That is a *different* overlap — a junction falling inside a `Repeat`'s tract span, not two members sharing one junction — and it needs `Repeat` in the `blocks_sibling_shift` predicate, which is #1269's question. Next in the stack.

## Verification

- 8899 pass; 8899 pass again under `FERRO_ASSERT_IDEMPOTENT=1` with `FERRO_ASSERT_REPARSE=1`.
- The 276,480-case sweep still passes with its residual pinned at 0 (from #1288).
- Commit is signed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved variant normalization when duplications and insertions cross neighboring substitutions or deletions.
  * Corrected shifting behavior around reference-consuming boundaries.
  * Combined identical insertions that resolve to the same genomic junction while preserving distinct insertions separately.
  * Improved handling of insertion and duplication payloads from literals or reference sequences.
* **Tests**
  * Expanded coverage for cis insertions, duplication shifts, junction merging, sequence preservation, and normalization consistency.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -3751,6 +3751,185 @@ pub(crate) fn respell_colliding_duplications<P: ReferenceProvider>(
     }
 }
 
+/// Coalesce cis members that settled onto **one junction** into a single
+/// insertion carrying their combined payload.
+///
+/// A member that consumes no base occupies the interbase junction where its
+/// added sequence lands. Two of them can shift onto the same junction from
+/// distinct starting points — two one-base insertions at adjacent gaps inside a
+/// homopolymer both travel to the tract's 3'-most junction:
+///
+/// ```text
+/// reference  ("ACGT" x 64) + "AAAAAA" + ("ACGT" x 64)   poly-A run at 257-263
+/// g.[258_259insA;259_260insA]  ->  g.[263dup;263dup]
+/// ```
+///
+/// That output claims one position twice, so it has no well-defined resulting
+/// sequence — `parse_hgvs` accepts it, and the SPDI apply oracle declines it
+/// (#1286). Nothing upstream prevents it: the sibling clamps bound a member
+/// against a sibling's *bases*, and neither of these has any.
+///
+/// It cannot be repaired downstream either. The sequence-first canonicalization
+/// would derive the correct single member, but it runs *after* the per-member
+/// pipeline and derives from the allele's resulting sequence — which an
+/// overlapping allele does not have. It declines, so the corruption survives the
+/// one pass that would have merged it.
+///
+/// So merge here, before that pass sees it. Both members add their payload at
+/// the same interbase point, so one insertion carrying the concatenation denotes
+/// exactly what the pair denoted. Re-spelling as an `Insertion` rather than a
+/// wider `Duplication` is the same choice `respell_colliding_duplications`
+/// makes: an insertion's payload is a literal and is correct wherever it lands,
+/// whereas `g.262_263dup` would only be equivalent when `ref[262] == ref[263]`.
+/// The next pass re-canonicalises it to a `dup` or a repeat where the reference
+/// permits.
+///
+/// **Only when every colliding payload is identical.** Concatenation is
+/// otherwise order-dependent, and the order is exactly what the shift destroyed:
+/// two insertions that were at distinct junctions had a defined order, and once
+/// they share a junction they do not. Equal payloads make the concatenation
+/// order-independent, so the repair is unambiguous. A differing-payload
+/// collision is left for the overlap detector to report rather than guessed at.
+///
+/// Genomic axes only, matching `respell_colliding_duplications` — the
+/// transcript axes need a coordinate conversion this does not carry.
+pub(crate) fn coalesce_members_at_one_junction<P: ReferenceProvider>(
+    members: &mut Vec<HgvsVariant>,
+    phase: AllelePhase,
+    uncertain: bool,
+    provider: &P,
+) {
+    if phase != AllelePhase::Cis || uncertain || members.len() < 2 {
+        return;
+    }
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #1290, opened 2026-08-01
Normalizer: an insertion's junction shifts past a sibling insertion's junction, silently reordering the two payloads


nh13:
## What's wrong

An insertion's junction 3'-shifts through a tract and **past another insertion's junction**, reordering the two payloads and changing what the allele denotes.

```
reference  ("ACGT" x 64) + "ATACAGAAAATCAGGGCATA" + ("ACGT" x 64)
           so core[i] is at 1-based position 257 + i; A-run at 263-266

in   NC_TEST.1:g.[263_264insA;265_266insC]
out  NC_TEST.1:g.[265_266insC;266dup]

  intended  ... G(262) A A A A C A(266) T(267) ...
  emitted   ... G(262) A A A C A A(266) T(267) ...
```

The `A` inserted at gap 263|264 travels 3' through the A-run to gap 266|267, crossing the `C` inserted at gap 265|266. Both insertions survive, but they now apply in the opposite order, so the `A` lands after the `C` instead of before it.

## Severity

**Silent.** The output is well-formed, disjoint and warning-free — `parse_hgvs` accepts it and the SPDI apply oracle applies it cleanly. It simply denotes different bases. Neither `FERRO_ASSERT_REPARSE` nor `FERRO_ASSERT_IDEMPOTENT` can see it.

## Suspected mechanism

`clamp_sibling_crossing_junctions` bounds a member's junction against a sibling's **bases**. Both members here are junction-occupying and claim no bases, so neither is a barrier to the other and the clamp has nothing to stop the crossing.

This is the junction-versus-junction case of the same family as:

- **#1276/#1277** — a duplication occupies a junction, which the overlap *detector* missed;
- **#1286** — two junction-occupying members landing on the *same* junction (fixed by merging them);
- **#1287** — a junction landing inside a `Repeat`'s tract span (fixed by extending the demotion).

Here the two end up at *different* junctions, so there is no overlap to detect — only a reordering. That makes it the quietest member of the family.

## Pre-existing

Verified: identical output (`g.[265_266insC;266dup]`, same changed sequence) on the base of #1287's branch, before any of the #1286/#1287 work. It was merely masked — the property test that finds it stopped at the two overlap defects first.

## How it was found

By the `IndelHaplotype` confluence model added in #1285, once #1286 and #1287 were fixed. Minimal case from proptest shrinking:

```
[...]

PR #1291, opened 2026-08-01
fix(normalize): stop a repeat's tract swallowing a sibling's junction

### LINKED ISSUE #1287: Normalizer: a duplication's junction lands inside a sibling repeat's tract span, emitting overlapping members

nh13:
> Stacked on #1289 → #1288 → #1285 → #1259. Base retargets as those merge.

Closes #1287.

```
reference  ("ACGT" x 64) + "ATACAGAAAATCAGGGCATA" + ("ACGT" x 64)
g.[261_262insGA;263_264insAA]  ->  g.[262_263dup;263_266A[6]]
```

Interbase spans of the output are `(263, 263)` and `(262, 266)` — the duplication's junction sits inside the repeat's span. The members overlap and the pair has no well-defined resulting sequence: `parse_hgvs` accepts the output, the SPDI apply oracle declines it while the input applies fine.

## Fix

`demote_repeats_spanning_siblings` exists for exactly this shape — a repeat's tract-wide span swallowing a sibling — but it asked only whether a sibling **claims bases**. A duplication claims none; it occupies a junction. So the demotion never fired and the repeat kept a span containing that junction.

A copy count over `262_266` cannot express another member adding sequence in the middle of those bases, so a junction strictly inside the tract is swallowed exactly as claimed bases are. Count it.

**The 3' end stays exclusive.** A junction at the tract's end is *flush against* it rather than inside it, which is the adjacency the collapse pass exists to catch (#999, #1135) — the same strictness `detect_insertion_overlaps` uses for an interior junction (#1276). Demoted, the repeat becomes a duplication that clears the sibling: `g.[262_263dup;265_266dup]`.

## A third defect fell out — filed as #1290

The `IndelHaplotype` confluence property added in #1285 found #1286 and #1287. With both closed it immediately surfaced a third:

```
core "ATACAGAAAATCAGGGCATA"
g.[263_264insA;265_266insC]  ->  g.[265_266insC;266dup]
  intended  G A A A A C A T
  emitted   G A A A C A A T      <- the inserted A crossed the C's junction
```

An insertion's junction shifts 3' **past another insertion's junction**, reordering the two payloads. Both members survive and land on *different* junctions, so there is no overlap to detect — the output is well-formed, disjoint and warning-free, and simply denotes different bases. The quietest member of the family.

**Verified pre-existing**, not introduced here: identical output on this branch's base, before any of the #1286/#1287 work. It was masked behind the two overlap defects, which the property hit first.

`clamp_sibling_crossing_junctions` bounds a junction against a sibling's *bases*, and both members here have none — so nothing is a barrier to anything. That is the junction-versus-junction case of the family #1276/#1277, #1286 and #1287 all belong to.

The property therefore stays `#[ignore]`d, with #1290's reproduction and the now-fixed history in its doc comment. Enabling it is #1290's acceptance criterion.

## Tests

`tests/it/issue_1287_repeat_span_junction.rs`, 3 tests, generated programmatically: the reproduction, order-independence, and a guard that a **lone** insertion inside a tract still reaches repeat notation (`repeated.md` B2) — the demotion must not fire when there is no sibling to swallow.

The sequence check projects through `hgvs_to_spdi`, a different code path from the normalizer's own applier.

## Verification

- 8902 pass; 8902 pass again under `FERRO_ASSERT_IDEMPOTENT=1` with `FERRO_ASSERT_REPARSE=1`.
- The 276,480-case sweep still passes with its residual pinned at 0.
- Commit is signed.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -3332,14 +3332,33 @@ pub(crate) fn demote_repeats_spanning_siblings(
         if length <= 0 || start < a.start || b.region != a.region {
             continue;
         }
-        // Does the tract actually span a sibling's bases? Both snapshots, so a
+        // Does the tract actually swallow a sibling? Both snapshots, so a
         // sibling that moved into the tract counts too.
-        let spans_sibling = (0..pre.len())
-            .filter(|&j| j != i)
-            .flat_map(|j| [pre[j].as_ref(), post[j].as_ref()])
-            .flatten()
-            .filter(|s| s.claims_bases && s.region == a.region && s.accession == a.accession)
+        //
+        // Two ways to be swallowed. A sibling that *claims bases* is swallowed
+        // when its span meets the tract. A sibling that claims none is still
+        // swallowed when its **junction** falls strictly inside the tract
+        // (#1287): a copy count over `1_9` cannot express another member adding
+        // sequence in the middle of those nine bases, so the pair overlaps. The
+        // 3' end is exclusive on purpose — a junction at `a.end` is flush against
+        // the tract rather than inside it, which is the adjacency the collapse
+        // pass exists to catch (#999, #1135), and the same strictness
+        // `detect_insertion_overlaps` uses for an interior junction (#1276).
+        let siblings = || {
+            (0..pre.len())
+                .filter(|&j| j != i)
+                .flat_map(|j| [pre[j].as_ref(), post[j].as_ref()])
+                .flatten()
+                .filter(|s| s.region == a.region && s.accession == a.accession)
+        };
+        let spans_sibling_bases = siblings()
+            .filter(|s| s.claims_bases)
             .any(|s| s.start <= a.end && s.end >= a.start);
+        let spans_sibling_junction = siblings()
+            .filter(|s| !s.claims_bases)
+            .filter_map(|s| s.junction)
+            .any(|junction| junction >= a.start && junction < a.end);
+        let spans_sibling = spans_sibling_bases || spans_sibling_junction;
         if !spans_sibling {
             continue;
         }
--- a/tests/it/cis_allele_confluence_proptest.rs
+++ b/tests/it/cis_allele_confluence_proptest.rs
@@ -54,9 +54,9 @@
 //!   happened to be written. Tracked by #1260 and #1262 and pinned by
 //!   `adjacent_gap_insertions_are_a_known_gap`.
 //! - `an_indel_haplotype_normalizes_to_its_own_sequence` is `#[ignore]`d
-//!   because it finds #1287, an unfixed defect in the #1269
-//!   `claims_reference_bases` family. It found #1286 first, now fixed; see its
-//!   doc comment for the history and the live reproduction.
+//!   because it finds #1290, a junction shifting past another junction. It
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #1292, opened 2026-08-01
Normalizer: an insertion cancelling against a nearby deletion drops a base from its payload


nh13:
## What's wrong

An insertion and a nearby deletion partly cancel, and the cancellation drops a base from the inserted payload while leaving an extra reference base in its place.

```
reference  ("ACGT" x 64) + "GCAAATAACATCCAGA" + ("ACGT" x 64)
           so core[i] is at 1-based position 257 + i
           positions 263 = A, 264 = A, 265 = C, 266 = A

in   NC_TEST.1:g.[263_264insAC;265del]
out  NC_TEST.1:g.265delinsAA

  intended  ... T(262) A A C A A T(267) ...     (GCAAATAACAATCCAGA)
  emitted   ... T(262) A A A A A T(267) ...     (GCAAATAAAAATCCAGA)
```

The inserted `C` is gone and an extra `A` has taken its place. The insertion adds `AC` after 263 and the deletion removes the `C` at 265; those two `C`s are *different* bases, but the cancellation path treats them as interchangeable.

## Severity

**Silent.** The output is a single well-formed member. `parse_hgvs` accepts it, there is no overlap to detect, and the SPDI apply oracle applies it cleanly — it simply denotes different bases. Neither `FERRO_ASSERT_REPARSE` nor `FERRO_ASSERT_IDEMPOTENT` can see it.

## Suspected mechanism — unverified

The merge's insertion/deletion **cancellation** path, not a sibling-crossing clamp. This is the same class as #1266 ("the merge's cancellation path producing a wrong result once the members are adjacent"), which covered `dup` + `del`; this is `ins` + `del`, where the inserted payload is multi-base and only partly matches what the deletion removes.

Filed as unverified: I have not traced which pass emits `265delinsAA`.

## How it was found

By the `IndelHaplotype` confluence model added in #1285, after #1286, #1287 and #1290 were fixed — it is the fourth defect that property has surfaced. Minimal case from proptest shrinking:

```
core "GCAAATAACATCCAGA", insert "AC" after index 6, delete index 8
```

No committed sweep reaches it: the sibling sweeps generate only single-base payloads that are in phase with the adjacent reference (gaps 1 and 3 in #1283), so a two-base payload that partly matches a deleted base is unreachable there.

## Status

[...]

PR #1293, opened 2026-08-01
fix(normalize): stop a junction shifting past another junction

### LINKED ISSUE #1290: Normalizer: an insertion's junction shifts past a sibling insertion's junction, silently reordering the two payloads

nh13:
> Stacked on #1291 → #1289 → #1288 → #1285 → #1259. Base retargets as those merge.

Closes #1290.

```
reference  ("ACGT" x 64) + "ATACAGAAAATCAGGGCATA" + ("ACGT" x 64)
g.[263_264insA;265_266insC]  ->  g.[265_266insC;266dup]
  intended  G A A A A C A T
  emitted   G A A A C A A T      <- the inserted A crossed the C
```

`clamp_sibling_crossing_junctions` bounds a member's junction against a sibling's **bases**. Two junction-occupying members have none, so neither was a barrier to the other.

Both members survive and land on *different* junctions, so there is no overlap to detect — the output is well-formed, disjoint and warning-free, and simply denotes different bases. The quietest member of the family that #1276/#1277, #1286 and #1287 also belong to. **Pre-existing**: verified identical on this branch's base, before any of the #1286/#1287 work.

## Fix

Bar the crossing, stopping one position **short** of the sibling's junction. Landing *on* it makes the two share a junction, which is the case with no defined payload order at all.

**Only when the payloads do not commute.** `p ++ q == q ++ p` holds exactly when both are powers of one common word — precisely when the order two insertions apply in is unobservable. `A` and `AA` in a poly-A tract commute; `A` and `C` do not.

Commuting payloads are let through *on purpose*: they then meet on one junction and merge into a single member (#1286), which is the canonical form the sequence-first derivation also reaches. Barring unconditionally split `g.[258_259insA;259_260insA]` into `g.[262dup;263dup]` — correct, but two members where one will do.

**Equality is not enough in place of commuting.** Three insertions merge to one only because `A` commutes with the merged `AA`; an equality test stops after the first pair and leaves two members. I hit exactly that while building this, which is why the test below pins it.

`coalesce_members_at_one_junction` now uses the same predicate, so a pair is either kept apart and ordered, or allowed together and merged — never left overlapping. The two rules are deliberately one rule.

`clamp_sibling_crossing_junctions` gains a provider parameter, to read a duplication's payload from the reference.

## Tests

`tests/it/issue_1290_junction_crosses_junction.rs`, 4 tests, generated programmatically: the reproduction, order-independence, and two guards on the commuting rule — that commuting payloads still meet and merge to one member, and that a *merged* payload still commutes with a remaining one (the three-insertion case that an equality test fails).

## A fourth defect — filed as #1292

The `IndelHaplotype` property added in #1285 has now found four defects, each masked behind the one before it: #1286, #1287, #1290, and now **#1292** — the insertion/deletion cancellation path dropping a base from the payload:

```
core "GCAAATAACATCCAGA", insert "AC" after index 6, delete index 8
  g.[263_264insAC;265del] -> g.265delinsAA
    intended  T A A C A A T
    emitted   T A A A A A T      <- the inserted C is gone
```

Same class as #1266's "merge's cancellation path" but for `ins` + `del` with a multi-base payload. The property therefore stays `#[ignore]`d, now with #1292's reproduction; enabling it is that issue's acceptance criterion.

That each defect only became visible once the previous was fixed is the argument for keeping the property committed and failing rather than deleting it until it can pass.

## Verification

- 8906 pass; 8906 pass again under `FERRO_ASSERT_IDEMPOTENT=1` with `FERRO_ASSERT_REPARSE=1`.
- The 276,480-case sweep still passes with its residual pinned at 0.
- Commit is signed.

Not run: the manifest-backed conformance axis tests — no reference volume mounted locally. Worth watching the nightly.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -3395,11 +3395,12 @@ pub(crate) fn demote_repeats_spanning_siblings(
 /// A junction sits immediately 3' of one position: `end` for a duplication
 /// (the copy follows the duplicated bases), `start` for an insertion (whose
 /// span is the gap `start_end`).
-pub(crate) fn clamp_sibling_crossing_junctions(
+pub(crate) fn clamp_sibling_crossing_junctions<P: ReferenceProvider>(
     before: &[HgvsVariant],
     after: &mut [HgvsVariant],
     phase: AllelePhase,
     uncertain: bool,
+    provider: &P,
 ) {
     if phase != AllelePhase::Cis || uncertain || after.len() < 2 || before.len() != after.len() {
         return;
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

Issue #1297, opened 2026-08-01
Normalizer: a cancelled member is left as an identity member overlapping the repeat that absorbed it


nh13:
## What's wrong

Three length-changing members collapse into one repeat that correctly denotes the whole haplotype, and an identity member (`=`) is left behind on a base that repeat already claims. The pair overlaps.

```
reference  ("ACGT" x 64) + "GCATGAAAAT" + ("ACGT" x 64)
           so core[i] is at 1-based position 257 + i

in   NC_TEST.1:g.[261_262insAA;263del;264_265insA]
out  NC_TEST.1:g.[262_265A[6];265=]
```

`262_265A[6]` is right on its own — it denotes `GCATG` + `AAAAAA` + `T`, exactly what the input denotes. The `265=` is the residue of a member that cancelled, and it claims 265, which the repeat's tract also covers. The SPDI apply oracle declines the output.

## Severity

**Loud** on the tip of #1293 plus #1292's fix (the apply oracle rejects it). On the current `main`-side behaviour it is **silent and wrong** instead:

```
out  NC_TEST.1:g.[261_262dup;265=]        <- duplicates the G at 261
     denotes  GCATGG A A A A T ...        <- a G that is not in the input's sequence
```

So #1292's fix changes this shape's failure mode from silently-wrong to loudly-rejected, which is why it only becomes visible once #1292 is fixed.

## Suspected mechanism — partly traced

Two separable questions:

1. Why is a member that cancelled to `=` retained at all, rather than dropped once another member absorbed its effect? An identity member contributes nothing to the sequence, but it *claims* a position, which is what creates the overlap.
2. Why does it sit on a base the repeat covers? The repeat's tract was grown by the two insertions; the `=` is at 265, inside `262_265`.

Not traced past that: I have not identified which pass emits the `=` member or which one grows the tract over it.

## How it was found

By `cis_allele_confluence_proptest::an_indel_haplotype_normalizes_to_its_own_sequence` at 30,000 cases (its committed budget is 512), after #1292 and #1296 were both fixed — the sixth defect that property has surfaced, each masked by the one before it: #1286 -> #1287 -> #1290 -> #1292 -> #1296 -> this.

Minimal proptest seed, committed to `tests/proptest-regressions/cis_allele_confluence_proptest.txt`. It is inert while the property is `#[ignore]`d and replays the moment it is enabled, which is what should gate enabling it:

```
[...]

Issue #1296, opened 2026-08-01
Normalizer: a repeat does not claim its bases, so a sibling deletion shifts onto it


nh13:
## What's wrong

A `Repeat` member claims the reference bases under its tract, but `claims_reference_bases` in `src/normalize/merge.rs` does not list `NaEdit::Repeat`. So a repeat is invisible to every sibling-repair pass that asks "does this sibling claim bases?", and it has no junction either (`junction_of` returns `None` for a repeat). A sibling deletion can therefore 3'-shift straight onto the tract, and the pair overlaps.

```
reference  ("ACGT" x 64) + "AAAAAAATAATCGCAACAGAAG" + ("ACGT" x 64)
           so core[i] is at 1-based position 257 + i

in   NC_TEST.1:g.[272_273del;274_275insAA]
out  NC_TEST.1:g.[273_274del;274A[3]]
```

`273_274del` claims 273 and 274; `274A[3]` claims 274. The two members overlap, and the SPDI apply oracle declines the output.

## Severity

**Loud.** The output is rejected by the apply oracle rather than silently denoting different bases, so it fails the confluence property's own preservation check rather than slipping past it.

## Mechanism — traced

`clamp_sibling_crossing_shifts` bounds a 3'-shifting member against siblings that claim bases:

```rust
let onto_bases = siblings
    .iter()
    .filter(|s| s.claims_bases)
    .map(|s| s.start)
    .filter(|&start| start > b.end && start <= a.end)
    .map(|start| start - 1);
```

The deletion shifts from `[272, 273]` to `[273, 274]`, so its sweep window reaches 274 — exactly the base the repeat's tract starts on. With `claims_bases` false for a repeat, `onto_bases` is empty, `across_junctions` is empty too (a repeat has no junction), and no barrier is found.

`demote_repeats_spanning_siblings` does not cover this either: it declines when the tract is shorter than the edit it grew out of (`start < a.start`), which is the case here — a one-base tract that gained two copies.

## A verified one-line fix

Adding `NaEdit::Repeat { .. }` to `claims_reference_bases` produces

```
out  NC_TEST.1:g.[272_273del;274A[3]]
```

which denotes the input's bases, with the full suite green (8918 tests) including the 276,480-case `cis_junction_crossing_shift` sweep holding its residual at 0. Worth confirming the other five readers of `claims_bases` in that module want the same answer before taking it, but the measurement is clean.

## How it was found

[...]

PR #1298, opened 2026-08-01
fix(normalize): keep a moved member's junction payload in phase

### LINKED ISSUE #1280: Normalizer: repeat demotion places the duplication at the tract's 3'-most position, across a sibling it then corrupts

### LINKED ISSUE #1292: Normalizer: an insertion cancelling against a nearby deletion drops a base from its payload

nh13:
Closes #1292. Closes #1280.

A cis member that adds sequence at a junction denotes its payload *at that junction*. Two sibling-repair passes move such a member, and a `Duplication` reads its payload from the reference **under its own span** — so sliding the span silently rewrites what the member copies.

```
g.[263_264insAC;265del]     on ("ACGT"x64) + "GCAAATAACATCCAGA" + ("ACGT"x64)
  per-member ->  g.[265_266dup;265del]      dup payload "CA"
  clamped    ->  g.[263_264dup;265del]      payload silently became "AA"
  merged     ->  g.265delinsAA              the inserted C is gone

  intended  T A A C A A T
  emitted   T A A A A A T
```

Traced live on `clamp_sibling_crossing_junctions` → `translate_member`, whose `landed` check verifies position and region but never that the member still denotes the same bases.

## The fix

Rotate the payload into phase at the destination, and re-spell the member as a plain insertion carrying it whenever the translated spelling would denote something else:

```
insert P at junction j  ==  insert (ref[j] ++ P[..n-1]) at junction j - 1
                            only when P[n-1] == ref[j]
```

applied one position at a time, so a multi-position move is the composition — which is also why a payload can rotate through more than its own length. On the trace above: `"CA"` at 266 → 265 needs `P[1]=='A' == ref[266]` ✓ giving `"AC"`; 265 → 264 needs `'C' == ref[265]` ✓ giving `"CA"` at 264. A move whose payload cannot be rotated, read, or re-spelled leaves the member where it is — an unclamped crossing is the defect the pass is repairing, but a corrupted payload is a worse one.

**A shared helper, not a per-call-site repair**, because the same rewrite is reachable from both passes that translate a member: `clamp_sibling_crossing_junctions` (#1292) and `clamp_sibling_crossing_shifts`, which a duplication reaches because it blocks a sibling's shift without claiming bases.

## Why this closes #1280 without touching its placement

#1280 is the same root cause arrived at through `demote_repeats_spanning_siblings`: its duplication over the tract's 3'-most bases is then pulled back by the very same junction clamp, which corrupted the payload. Both of its reproductions are fixed here, so its "compute the placement subject to the sibling barrier" suggestion is not needed — a duplication is phase-correct where it already sits, and the clamp now moves it correctly:

```
(1)  g.[259_260insTTTTT;262T>A]   ->  g.263_267dup          was g.262delinsCTTTTA
(2)  g.[269_270insTTTTT;272T>A]   ->  g.273_277dup          was g.272delinsCTTTTA
```

(Both re-expressed on the padded synthetic contig; case (2) is the one whose output carried a `C` that appears nowhere in the sequence the input denotes.)

Rotating rather than carrying the payload literally is exactly what makes this safe in a tandem repeat. #1280 records an attempted fix — demote the repeat to an insertion carrying its **literal** payload — that was reverted because it skipped the rotation: invisible in a homopolymer, wrong in a `(CAG)` tract. That counterexample is pinned as a test here, and still keeps its duplication spelling:

```
g.[268_269insAGCAGC;262del]  ->  g.[262del;274_279dup]
```

## Also in this PR

A latent off-by-one in the same helper. `respell_at_gap` placed its edit at `[span.end, span.end + 1]`, which is the junction for a duplication but one position **past** it for an insertion, whose span is the gap itself — so `coalesce_members_at_one_junction` would have placed a merged payload one position 3' of the junction its group shares, whenever the target member was spelled as an insertion. Invisible in the poly-A cases the pass has tests for. It now takes the junction explicitly.

## The acceptance criterion, and why it is not met literally

#1292's stated acceptance criterion was removing `#[ignore]` from `cis_allele_confluence_proptest::an_indel_haplotype_normalizes_to_its_own_sequence`. **The property still fails**, and not on #1292: its next two failures were masked behind this one, which is the fifth and sixth instance of that pattern in this stack (#1286 → #1287 → #1290 → #1292 → …).

- **#1296** — a `Repeat` claims the bases under its tract but does not report that it does, so it is no barrier to a sibling's 3' shift and the pair overlaps. Found at the property's committed 512-case budget once #1292 was fixed. A verified one-line fix exists; filed separately rather than folded in here.
- **#1297** — a cancelled member left as an identity member (`265=`) overlapping the repeat that absorbed it. Found at 30,000 cases with #1292 and #1296 both fixed.

Both are **pre-existing** — verified by reverting `src/normalize/{merge,mod}.rs` to this branch's parent and re-running, not by `git stash`. #1297 is *silently wrong* on the parent (`g.[261_262dup;265=]`, a `G` that is not in the input's sequence) and becomes a loud overlap here, so this PR improves its failure mode.

So the ignore stays, its reason now names #1296, and the property's doc comment records that #1292 is fixed and pinned by other means. #1292 is closed on that evidence: its reproduction is fixed, `tests/it/issue_1292_junction_payload_rotation.rs` pins it, and its proptest seed `99d5d382` is committed and replays green. Enabling the property belongs to #1297, which says so. Deliberately **not** weakened to make it pass.

## Verification

- `cargo nextest run --features dev` — **8920 pass**, 0 fail
- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 …` — 8920 pass under both oracles
- `cis_junction_crossing_shift::no_two_member_allele_normalizes_to_a_different_sequence` — the 276,480-case sweep holds its residual at **0**, the measurement that killed the reverted attempts in #1279/#1280
- Commit is signed

Conformance axis tests were **not** run: they need the prepared reference, and both `/Volumes/scratch-00001` and `/Volumes/backup-00001` are absent on this machine. Watching the nightly.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -3120,11 +3120,12 @@ fn blocks_sibling_shift(edit: &NaEdit) -> bool {
 /// live list keeps the pass independent of member order.
 ///
 /// [`ShuffleDirection::FivePrime`]: crate::normalize::ShuffleDirection
-pub(crate) fn clamp_sibling_crossing_shifts(
+pub(crate) fn clamp_sibling_crossing_shifts<P: ReferenceProvider>(
     before: &[HgvsVariant],
     after: &mut [HgvsVariant],
     phase: AllelePhase,
     uncertain: bool,
+    provider: &P,
 ) {
     // An uncertain allele — `g.[(…)]` — asserts only that the members are in
     // cis, not where they sit, so repositioning one states something the input
@@ -3242,7 +3243,13 @@ pub(crate) fn clamp_sibling_crossing_shifts(
             continue;
         };
         if pull != 0 {
-            translate_member(&mut after[i], -pull, kind, a);
+            // A duplication blocks a sibling's shift without claiming bases, so
+            // it reaches this pass too — and its payload is read from under its
+            // own span, so it cannot simply be slid (#1280, #1292).
+            match a.junction {
+                Some(_) => translate_junction_member(&mut after[i], -pull, kind, a, provider),
+                None => translate_member(&mut after[i], -pull, kind, a),
+            }
         }
     }
 }
@@ -3506,7 +3513,7 @@ pub(crate) fn clamp_sibling_crossing_junctions<P: ReferenceProvider>(
         // Never move past where the junction started.
         let delta = (limit - after_junction).max(before_junction - after_junction);
         if delta != 0 {
-            translate_member(&mut after[i], delta, kind, a);
+            translate_junction_member(&mut after[i], delta, kind, a, provider);
         }
     }
 }
@@ -3812,7 +3819,7 @@ pub(crate) fn respell_colliding_duplications<P: ReferenceProvider>(
         };
         // The copy lands at the junction 3' of the duplicated span, which is
         // the gap `[end, end + 1]`.
-        respell_at_gap(&mut members[i], kind, dup, edit);
+        respell_at_gap(&mut members[i], kind, dup, dup.end, edit);
     }
 }
 
@@ -3935,28 +3942,13 @@ pub(crate) fn coalesce_members_at_one_junction<P: ReferenceProvider>(
         let Some(span) = spans[target].as_ref() else {
             continue;
         };
+        // The merged payload belongs at the junction the group shares, which is
+        // not `span.end` when the target member is spelled as an insertion.
         let Some(junction) = span.junction else {
             continue;
         };
-        // `respell_at_gap` places the edit over `[span.end, span.end + 1]`, so
-        // hand it a span whose `end` is the shared junction rather than the
-        // member's own. For a `Duplication` the two coincide (`junction_of`
-        // returns its `end`), but for an `Insertion` the junction is its
-        // `start`, so `span.end` sits one base 3' of it. `respell_at_gap`
-        // cannot catch that: it checks the result landed on the gap it was
-        // given, not that the gap was the right one.
-        let gap = MemberSpan {
-            accession: span.accession.clone(),
-            provider_key: span.provider_key.clone(),
-            region: span.region,
-            start: junction,
-            end: junction,
-            claims_bases: span.claims_bases,
-            blocks_shift: span.blocks_shift,
-            junction: Some(junction),
-        };
         let before = members[target].clone();
-        respell_at_gap(&mut members[target], kind, &gap, edit);
+        respell_at_gap(&mut members[target], kind, span, junction, edit);
         if members[target] == before {
             continue; // the re-spell did not take; leave the group alone
         }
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #1299, opened 2026-08-01
fix(normalize): let a repeat claim the bases under its tract

### LINKED ISSUE #1269: Normalizer: claims_reference_bases excludes NaEdit::Repeat, and two workarounds now sit on top of it

### LINKED ISSUE #1296: Normalizer: a repeat does not claim its bases, so a sibling deletion shifts onto it

nh13:
Closes #1296. Closes #1269.

`claims_reference_bases` is how every sibling-repair pass asks "can this member collide with that one", and it excluded `NaEdit::Repeat`. A repeat has no junction either, so the junction rule did not see it. A sibling deletion could 3'-shift straight into a tract:

```
reference  ("ACGT" x 64) + "AAAAAAATAATCGCAACAGAAG" + ("ACGT" x 64)
in   NC_TEST.1:g.[272_273del;274_275insAA]
out  NC_TEST.1:g.[273_274del;274A[3]]
```

`273_274del` claims 273 and 274; `274A[3]` claims 274. The pair overlaps and the apply oracle declines the output. The deletion of `AC` at `272_273` may legitimately 3'-shift to `273_274` on its own — `AC` and `CA` leave the same bases behind — but 274 is where the repeat's tract starts, so the shift has to stop.

## This is #1269's measurement, not a new judgement

The exclusion was deliberate, documented in the predicate's own doc comment, and tracked by #1269, which asked for exactly this:

> Flip the predicate behind a measurement rather than a judgement: add `Repeat`/`MultiRepeat`, run the two committed sweeps plus the full suite, and see whether the two workaround passes become dead code or merely redundant.

#1296 is the defect the exclusion left behind, so the measurement is here. The corpus survives unchanged: full suite green, both committed sweeps green, and `cis_junction_crossing_shift`'s 276,480-case sweep holds its residual at **0**.

## The workarounds do not become dead code

#1269's other question, answered by disabling each pass with the predicate corrected:

| pass disabled | result |
| --- | --- |
| `demote_repeats_spanning_siblings` | **12 tests fail**, including `repeat_span_sibling_overlap::no_two_member_allele_normalizes_to_overlapping_members` and `cis_junction_crossing_shift::no_two_member_allele_normalizes_to_a_different_sequence` — both committed sweeps |
| `respell_colliding_duplications` | **2 tests fail** (`normalize_reparse_invariant::a_del_beside_a_dup_re_spells_instead_of_colliding`, `…::every_sibling_pass_is_idempotent_on_a_compound_allele`) |

So correcting the predicate removes the reason a *third* workaround would have been wanted — #1296 is precisely that third shape — not the two that exist. Both keep their passes, their tests and their comments. #1269 asked for this to be documented either way; the answer is recorded in the new test module's header and in the commit message.

## Also in this PR

`claims_reference_bases`'s doc comment was glued to the front of `blocks_sibling_shift`'s, in one contiguous `///` run above the latter. So the function it described carried no doc at all and rustdoc rendered both paragraphs under the wrong name. Moved onto the function it describes; the `#1269` paragraph is rewritten in place since the omission it recorded is what this PR removes.

## The confluence property

Still `#[ignore]`d, and the reason moves on: with #1296 fixed its next failure is **#1297** (a cancelled member left as an identity member, overlapping the repeat that absorbed it). The ignore reason, the module docs, the property's doc comment and the seed-file header all now say #1297, and #1297's seed `a88766af` is committed, so taking the ignore off replays it immediately — which is what should gate taking it off.

## Verification

- `cargo nextest run --features dev` — **8923 pass**, 0 fail
- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 …` — 8923 pass under both oracles
- Commit is signed

Conformance axis tests were **not** run: they need the prepared reference, and both `/Volumes/scratch-00001` and `/Volumes/backup-00001` are absent on this machine. Watching the nightly.

Stacked on #1298.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved normalization of complex insertions, duplications, repeats, and deletions across neighboring junctions.
  * Preserved inserted sequence content and authored ordering during coordinate shifts.
  * Combined compatible insertions targeting the same junction while keeping distinct changes separate.
  * Prevented deletions from incorrectly shifting into repeat regions.

* **Tests**
  * Added extensive regression coverage to verify sequence preservation, stable normalization, and correct repeat and junction handling.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -3054,10 +3054,10 @@ fn member_span(v: &HgvsVariant, kind: CisKind) -> Option<MemberSpan> {
 /// span are what it copies; `blocks_sibling_shift` carries that, keeping this
 /// predicate to the narrower "consumes bases" question its name asks.
 ///
-/// `Repeat`/`MultiRepeat` are the one deliberate omission: a repeat does claim
-/// the bases under its tract, but making it say so changes clamp and barrier
-/// behaviour across the whole repeat corpus at once. That is tracked in #1269,
-/// with `demote_repeats_spanning_siblings` standing in until it is measured.
+/// `Repeat`/`MultiRepeat` claim the bases under their tract too: `g.1_9T[7]`
+/// covers positions 1–9 whatever copy count it asserts. They were excluded until
+/// the change could be measured (#1269), which left a repeat invisible as a
+/// barrier and let a sibling's 3' shift land inside its tract (#1296).
 fn claims_reference_bases(edit: &NaEdit) -> bool {
     matches!(
         edit,
@@ -3067,6 +3067,8 @@ fn claims_reference_bases(edit: &NaEdit) -> bool {
             | NaEdit::NPaddedDeletion { .. }
             | NaEdit::Delins { .. }
             | NaEdit::Inversion { .. }
+            | NaEdit::Repeat { .. }
+            | NaEdit::MultiRepeat { .. }
     )
 }
 
@@ -3281,11 +3283,16 @@ pub(crate) fn clamp_sibling_crossing_shifts<P: ReferenceProvider>(
 /// have taken had B2 declined, and it is exactly equivalent — B2 only re-writes
 /// notation, never the bases removed.
 ///
-/// Restoring the deletion is what lets the rest of the pipeline finish the job.
-/// [`clamp_sibling_crossing_shifts`] can then see a real span to pull back, and
-/// the caller's next merge pass coalesces the members. `g.[1_2del;4del]` ends up
-/// as `g.1_9T[6]` — one member, three bases removed from the tract, repeat
-/// notation restored now that there is no sibling to span.
+/// Restoring the deletion is what lets the rest of the pipeline finish the job:
+/// the member is back to a span the caller's next merge pass can coalesce with
+/// its sibling. `g.[1_2del;4del]` ends up as `g.1_9T[6]` — one member, three
+/// bases removed from the tract, repeat notation restored now that there is no
+/// sibling to span.
+///
+/// Visibility to [`clamp_sibling_crossing_shifts`] is no longer part of that:
+/// since #1296 a repeat reports that it claims the bases under its tract, so
+/// the clamp sees the widened span whether or not this pass has run. What this
+/// pass still supplies is the *narrower, coalescible* spelling.
 ///
 /// Runs before the clamp, so the clamp sees the deletion rather than the
 /// repeat. Members whose pre-normalization form was not a plain deletion are
--- a/tests/it/cis_allele_confluence_proptest.rs
+++ b/tests/it/cis_allele_confluence_proptest.rs
@@ -54,10 +54,10 @@
 //!   happened to be written. Tracked by #1260 and #1262 and pinned by
 //!   `adjacent_gap_insertions_are_a_known_gap`.
 //! - `an_indel_haplotype_normalizes_to_its_own_sequence` is `#[ignore]`d
-//!   because it finds #1296, a repeat that is not a barrier to a sibling's 3'
-//!   shift. It found #1286, #1287, #1290 and #1292 first, all now fixed, each
-//!   masked behind the one before it; see its doc comment for the history and
-//!   the live reproduction.
+//!   because it finds #1297, a cancelled member left as an identity member
+//!   overlapping the repeat that absorbed it. It found #1286, #1287, #1290,
+//!   #1292 and #1296 first, all now fixed, each masked behind the one before
+//!   it; see its doc comment for the history and the live reproduction.
 //!
 //! Neither restriction is silent: both are named, pinned, and fail loudly when
 //! the underlying issue is fixed (#1268/#1283's complaint about coverage that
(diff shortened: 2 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #1300, opened 2026-08-01
perf(test): draw confluence cores by index instead of a string union

nh13:
The idempotency soak was the longest job in PR CI. This makes its three dominant properties about 2.2x cheaper without touching the case count, the seeds, or what they cover.

## Where the time went

A shard finishes only when its slowest single property does — proptest runs a property's cases serially, so nextest cannot parallelize within one — and the three `cis_allele_confluence_proptest` properties took 159-168s against 22-28s for the four `normalize_idempotency_proptest` ones.

Almost none of that was the code under test. Ablating the property body one stage at a time, at 5,000 cases in the unoptimized test profile CI uses (user CPU rather than wall clock, 3 reps, spread under 2%):

| stage (cumulative) | user CPU | delta |
| --- | --- | --- |
| `core_strategy` draw only | 2.61s | **2.61s (61%)** |
| + full haplotype draw | 2.82s | +0.21 |
| + `encodings()` rendering | 2.96s | +0.14 |
| + provider build and clone | 3.12s | +0.16 |
| + one normalize | 3.58s | +0.46 |
| whole property | 4.25s | +0.67 |

`core_strategy` drew each of a core's 6-13 parts through a `prop_oneof!` over eight `Just(String)` arms, so every part cost a `TupleUnion` value tree plus a `String` clone, none of it inlined in an unoptimized build.

## The change

Draw an index into a `const` table instead. Same distribution — `prop_oneof!` weights its arms equally, so both are uniform over the eight parts — for 0.14s against 2.52s.

| | before | after |
| --- | --- | --- |
| `core_strategy` alone, 5k cases | 2.52s | **0.14s** (18x) |
| `the_converged_form_is_a_fixed_point`, 5k | 4.23s | **1.90s** |
| `all_encodings_of_one_haplotype_converge`, 5k | 4.27s | **1.90s** |
| `members_are_disjoint_and_ascending`, 5k | 4.23s | **1.85s** |

## Rebased onto #1285 — what changed

#1285 landed while this was in review and rewrote the same file, adding a second
`IndelHaplotype` model. Two consequences, both measured on the current base.

**The indel model must keep the old draw.**
`tests/proptest-regressions/cis_allele_confluence_proptest.txt` pins three
`IndelHaplotype` reproductions (#1286, #1287) as RNG **seeds**, and proptest
replays a seed *through the strategy* — so changing how the strategy consumes
randomness makes the same seed yield a different value. Verified rather than
assumed: with the indel model on the index draw,
`an_indel_haplotype_normalizes_to_its_own_sequence --run-ignored all` still
failed but on an unrelated case, and **appended a fourth seed to the checked-in
file**. Those pinned cases would have silently stopped guarding the bugs they
were filed for, and CI could not have caught it, because that test is
`#[ignore]`d. `git merge-tree` reports the merge as textually clean.

So `union_core_strategy` is retained for the indel model alone. The #1287 seed
now reproduces its recorded `core: "ATACAGAAAATCAGGGCATA"` value exactly, and
nothing is appended. The substitution model has no pinned cases and is free to
move.

**The headline is smaller than it was pre-#1285.** At 125,000 cases, same base:

| property | baseline | this PR |
| --- | --- | --- |
| `all_encodings_of_one_haplotype_converge` | 111.5s | **73.4s** |
| `the_converged_form_is_a_fixed_point` | 111.4s | **73.1s** |
| `members_are_disjoint_and_ascending` | 110.8s | **72.3s** |
| `an_indel_haplotype_is_authored_order_independent` | 103.7s | untouched |

#1285's indel property is now the shard's tail at ~104s, and this PR does not
touch it. Total heavy CPU per shard drops 522s -> 407s (about 22%), which is what
governs a throughput-bound 4-vCPU runner, so expect roughly **1.2-1.25x** on the
soak job rather than the 1.72x measured before #1285.

The obvious follow-up — giving `an_indel_haplotype_is_authored_order_independent`
the fast draw too — looks safe, since the pinned seeds were recorded by the
*other*, `#[ignore]`d indel test. It is deliberately not done here: it reworks a
test that landed days ago, and belongs with whoever owns #1286/#1287.

## On CI (pre-#1285, retained for the ablation)

Run [30681195305](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/30681195305) against [30677895044](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/30677895044) on `main`, all 8 shards green. Like-for-like on shard 6 (seed 6000023), where the four unchanged `normalize_idempotency_proptest` properties act as a control — they come out marginally *faster* after, so this is not runner noise:

| shard 6 | before | after | |
| --- | --- | --- | --- |
| `members_are_disjoint_and_ascending` | 222.9s | 136.2s | 1.64x |
| `the_converged_form_is_a_fixed_point` | 239.8s | 138.7s | 1.73x |
| `all_encodings_of_one_haplotype_converge` | 243.9s | 137.7s | 1.77x |
| *control:* 4 idempotency properties | 33.0-40.8s | 30.9-39.0s | - |
| soak step | 243.9s | **141.8s** | **1.72x** |

Across all 8 shards the job went from 183-271s to 163-195s. The job speedup (1.7x) is smaller than the per-property one (2.2x) because a runner has 4 vCPUs and a shard runs 7 heavy properties at once, so wall clock tracks total CPU: cutting 61% off three of seven removes about 45% of the total.

## Coverage

Unchanged: same parts, same length range, same uniform distribution. The non-vacuity guard `the_strategy_generates_every_shape_the_properties_depend_on` still passes, so the strategy still reaches adjacent, separated, and three-distinct-encoding shapes.

The seeded case stream does differ, since the strategy consumes RNG differently — a shard explores a different 125,000 cases than before. Each shard stays deterministic under its pinned seed, which is the property the soak relies on to be safe as a merge gate.

Verified at the soak's real settings (`PROPTEST_CASES=125000`, `PROPTEST_RNG_SEED=1000001`, `-E 'test(proptest)'`): 10/10 passing.

## Scope

One test file, one function body. #1303 stacks the CI-side work on top.

## Summary by CodeRabbit

* **Tests**
  * Improved test-generation performance while preserving reproducibility for existing regression cases.
  * Optimized soak and exhaustive sweep execution by reusing prebuilt test archives.
  * Reorganized test jobs and partitions for more efficient coverage and validation.

* **Chores**
  * Added a dedicated optimized build profile for soak testing.
  * Improved CI caching and artifact reuse to reduce unnecessary rebuilds.
  * Strengthened required test-status validation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,26 @@ permissions:
 env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: 1
+  # The three exhaustive-sweep modules, owned by the `sweeps` job and excluded
+  # from the `test` / `test-oracle` partitions. Defined once here because the
+  # filter has to appear identically in three places and GitHub Actions has no
+  # YAML anchors — a copy that drifts would either double-run a sweep or, far
+  # worse, drop one from CI entirely.
+  #
+  # The invariant to preserve is that the two selections PARTITION the suite —
+  # the absolute counts drift as tests are added, the sum must not. Verified
+  # against this base: `not test(proptest)` selects 8908, adding
+  # `and not (SWEEP_FILTER)` selects 8878, and the filter itself selects the
+  # remaining 30. 8878 + 30 == 8908, so nothing falls between the two jobs and
+  # nothing is run twice.
+  #
+  # `sweeps`, the one job that selects ON this filter rather than against it,
+  # passes `--no-tests=fail`. So a rename that made the filter match nothing
+  # reddens CI there instead of quietly emptying that job — while `test` and
+  # `test-oracle`, which negate it, would just widen by 30 tests.
+  SWEEP_FILTER: >-
+    test(cis_junction_crossing_shift) + test(repeat_span_sibling_overlap)
+    + test(issue_1254_sibling_crossing_shift)
 
 jobs:
   abi3-floor-consistency:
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
All four review comments addressed. Two notes where the fix deviates from what was proposed, plus what the sibling sweep turned up.

**Core parts single source of truth** — applied, but not by pointing `union_core_strategy` at `CORE_PARTS`. That would couple a deliberately replay-frozen strategy to a constant the substitution model is free to edit: reorder or retitle one `CORE_PARTS` entry and the six pinned indel seeds silently retarget, which is the exact silent failure the strategy split exists to prevent. Instead there is a frozen `UNION_CORE_PARTS` plus a tripwire test, `the_two_core_part_lists_agree`, so divergence is a failing test rather than a silent drift. Verified the seeds still replay: `an_indel_haplotype_normalizes_to_its_own_sequence --run-ignored all` reproduces the recorded #1297 case byte-for-byte and appends nothing, and a control run with the original literal arms produces the identical case, so the edit is provably neutral to replay.

**`cargo nextest --version` after the restore** — not added. Reordering `actions/cache` ahead of `taiki-e/install-action` makes the version deterministic by construction, and install-action already prints it, so the extra step would be noise in three jobs.

**Siblings the reviewed findings implied.** The cache-before-install reorder also applies to `test-build` and `test-build-soak` in ci.yml, which had the same shape. The build-only checkouts in nightly-soak lost `submodules: true` and `lfs: true` to match `test-build`, and the nightly soak shard lost `lfs: true` to match its ci.yml twin — 149 MB that was being fetched eight times a night for files no proptest module opens.

**One further cache bug, found while fixing the first.** The `restore-keys` on `test-build`, `test-build-soak`, `generated-docs` and the nightly build could never match their own primary keys. `actions/cache` matches restore-keys as a literal prefix, and `hashFiles('**/Cargo.lock')` and `hashFiles('**/Cargo.lock', '.github/workflows/ci.yml')` are different digests, so `...-<lock-hash>-test-archive` is not a prefix of `...-<lock+workflow-hash>-test-archive`. Confirmed in run 30712837408: `test-build` restored via the legacy `...-test-nodebug` key (520087453 B, byte-identical to what `generated-docs` restored from that same entry — the cross-job sharing the comment says was fixed) and then saved to `...-test-archive`, a key its own restore-key cannot reach. Once the legacy entry ages out, every workflow edit would compile all 316 crates cold. Fixed by moving the job suffix ahead of the hash so the restore-key is a real prefix. Expect one cold build per affected job on the first run after this lands, then warm.

Branch is rebased onto current `main`; full suite green locally (8970 passed).

nh13:
@coderabbitai review

nh13:
Second round addressed, plus the sibling the review did not name. Rebased onto current `main` (`77ee126`) on the way, which is what exposed the doc drift at the end.

**Redundant generator staging — confirmed and removed, both sides.** The finding is right, and it is worth more than "low value": the staged `generator-binaries` artifact was uploaded once and downloaded by nine shard jobs for binaries the archive already carries. Verified rather than reasoned about — `cargo nextest archive --features dev` packs `target/debug/generate_spec_fixture` and `target/debug/generate_spec_enumeration` alongside `target/debug/ferro`, all mode 755, and deleting both binaries then running `spec_generator_preconditions` from the archive restores them executable and passes all six tests. Removing only the upload would have reddened CI, so the two `download-artifact` steps and the two `chmod +x` steps went with it.

That also removed `FERRO_PREBUILT_EXAMPLES` from all four sites, which is the real answer to the outside-diff comment on the stale `test-build` attribution. Nothing has read that variable since #1228 made the generators `[[bin]]` targets — `rg FERRO_PREBUILT_EXAMPLES` outside the workflows hits one doc comment, which says the escape hatch became unnecessary. Renaming the job in a comment justifying a dead variable would have preserved the wrong thing; the note now lives on `test-build`, where the mechanism actually is, and records the verification.

**`--no-tests=fail` — the premise is wrong, the flag went in anyway.** These jobs do not exit green on an empty selection today: nextest's `--no-tests` defaults to `auto`, which is documented as "defaulting to fail", and a filter matching nothing exits **4** with `error: no tests to run` (measured against the archive, not assumed). So the stated failure mode is not reachable. Pinned regardless, because `auto` reserves the right to decide otherwise and these two jobs are worth nothing if their name filter stops matching — but the comment says it pins an existing default rather than fixing a live hole, since a comment describing the wrong failure path sends the next reader after the wrong bug.

Sibling the review missed: `nightly-soak.yml` runs the identical `-E 'test(proptest)'` selection against the identical archive and had the same gap. Fixed there too — a nightly is the worse place to lose a selection, because nobody watches a green run. Confirmed both selections are non-empty first: 15 tests for `test(proptest)` and 30 for `SWEEP_FILTER`, both listed from a locally built soak archive, so the new gate cannot redden a job that was quietly doing nothing.

**Drift the rebase exposed**, both fixed:

- `SWEEP_FILTER`'s partition arithmetic was measured against the older base. Re-measured on this one: `not test(proptest)` selects 8908, adding `and not (SWEEP_FILTER)` selects 8878, the filter alone selects 30, and 8878 + 30 == 8908.
- `union_core_strategy`'s doc said six pinned seeds with #1297 still open. The file now pins eight, and #1297 closed with #1309 — #1312 is the live one. Re-ran the counterfactual on this base and it is worse than what the doc claimed: with the indel model on the index draw, `an_indel_haplotype_normalizes_to_its_own_sequence --run-ignored all` no longer fails on an unrelated case, it **passes** — #1312 is still open but no seed reaches it, so the property reads as fixed. `the_ignored_indel_property_still_finds_its_defect` stays green through that, because it pins the defect by hardcoded input rather than through the strategy. The doc now records that.

With the draw preserved, the #1312 seed reproduces its recorded `core: "TAAAACCA"` exactly and appends nothing.

nh13:
@coderabbitai review

Issue #1301, opened 2026-08-01
Normalizer: adjacent-gap insertions normalize to a dup and an insertion that overlap and are out of order


nh13:
## What's wrong

Two insertions at adjacent gaps normalize to a duplication and an insertion that are rendered in an order their positions contradict, and whose spans overlap.

```
reference  ("ACGT" x 64) + "GCATGAAAAT" + ("ACGT" x 64)
           so core[i] is at 1-based position 257 + i

in   NC_TEST.1:g.[263_264insAC;264_265insAA]
out  NC_TEST.1:g.[264_265dup;264_265insCA]
```

The **sequence is right** — both members applied give `…G A A A C A A A A T…`, the bases the input denotes. What is wrong is the pair: `264_265insCA` sits at junction 264 and `264_265dup` copies `264_265` with its copy landing after 265, so the insertion applies 5' of the duplication while being rendered 3' of it, and their interbase spans overlap.

## Severity

**Loud, and an improvement on what it was.** On the tip of #1293 the same input produced

```
out  NC_TEST.1:g.[263_264insCA;264_265dup]      <- silently denotes different bases
```

#1298's payload rotation fixes the sequence; this issue is the ordering residue it leaves. So the shape moved from silently-wrong to sequence-correct-but-mis-ordered, and the confluence property's disjoint-and-ascending check is what now catches it.

## How it was found

By removing the adjacent-gap-insertion filter from `indel_haplotype_strategy` while measuring #1294. The filter's stated citation is #1260 / #1262, but neither of those is about overlapping members — they are non-confluence between the separate-member and spanning-`delins` spellings. So this shape was being excluded under a citation that does not describe it, which is the mis-attribution #1294 suspected.

Reproduced directly:

```
core "GCATGAAAAT", insert "AC" after index 6, insert "AA" after index 7
```

## Suggested first step

Check `cis_member_order_key` and `cis_member_range` against a duplication whose span *starts* at the junction an insertion sibling occupies. The two members tie on start (264) and the ordering falls back to a key that does not distinguish "adds sequence at 264" from "copies 264–265 and adds after 265".

Worth also asking whether the pair should be two members at all: `coalesce_members_at_one_junction` merges members that share a junction and whose payloads commute, and these sit at *adjacent* junctions with non-commuting payloads (`CA` and `AA`), so it declines — correctly, by its own rule.

## Relationship to #1235

[...]

Issue #1283, opened 2026-07-31
Tests: the exhaustive sibling-crossing sweep cannot reach five defect classes it is read as covering


nh13:
## What's wrong

`tests/it/cis_junction_crossing_shift.rs::no_two_member_allele_normalizes_to_a_different_sequence` is the main evidence for the sibling-crossing work — a ~276,000-case exhaustive sweep asserting no normalization changes the denoted sequence. Its generator is narrower than the claim it supports, in five independent ways, and **every blocking defect found so far lives in one of those blind spots**.

## The five gaps

| Gap | Where | What it hides |
| --- | --- | --- |
| Second member is only `{Ndel, N<ref>><alt>}` | the `second` loop | never a `dup`/`ins` sibling, so a whole class of sibling interactions is unreachable |
| Exactly two members | the `first`/`second` structure | 3-member shapes, including outputs that trip ferro's own `OverlapConflict` |
| Insertion payloads are always `&seq[first_start-1..first_end]` | the `firsts` construction | always in phase with the adjacent reference — never an out-of-phase rotation |
| 20 bp random sequences | the seed loop | no tandem tract long enough to canonicalise to a non-homopolymer `Repeat` |
| `g.` only | the accession | `c.`/`n.`/`r.` behaviour entirely |

The sibling sweeps in `issue_1254_sibling_crossing_shift.rs` and `repeat_span_sibling_overlap.rs` share the first four.

## Why it matters concretely

The file asserts an exact residual count and reads as "this class is now measured". Measured against that corpus, yes — but the corpus cannot fail on:

- an out-of-phase payload in a `(CAG)` tract (#1280): needs a ≥6-unit non-homopolymer tract *and* a 2+-copy insertion. Single-copy insertions canonicalise to `dup` and never reach the code.
- a 3-member allele producing duplicate members (#1281).
- a `dup` or `ins` as the *second* member (#1279's shape).
- anything on a transcript axis.

So a fix can pass this sweep and still be wrong, which is what happened: two separate attempted fixes passed the full 8878-test suite and were only caught by purpose-built sweeps written during adversarial review.

## Relationship to existing issues

- #1268 covers the 3-member gap only — this is the wider set.
- #1279, #1280, #1281 are the concrete defects that live in these blind spots.

## Suggested direction

Widen along the axes above rather than adding cases at random. In rough order of value per line changed:

1. allow `dup`/`ins` as the second member;
[...]

Issue #1268, opened 2026-07-30
Tests: no committed sweep covers three-member cis alleles, and the one measurement taken was never decomposed


nh13:
## What's wrong

Two related gaps in cis-allele test coverage, both surfaced while working #1259.

### 1. No committed sweep exercises three-member alleles

CodeRabbit asked for this on the (now closed) #1257:

> extend the randomized overlap/convergence coverage to alleles with three or more members

It was measured with a throwaway probe and the probe was deleted. Both committed sweeps — `tests/it/cis_junction_crossing_shift.rs` and `tests/it/repeat_span_sibling_overlap.rs` — generate **two** members only. So the ask is not addressed in the artifact, and it was reported as addressed. Correcting that here.

### 2. The three-member measurement has an undecomposed failure bucket

The throwaway probe covered 1,009,152 three-member alleles (`del`/`dup`/`ins` first member, both shuffle directions):

| | `main` | #1259 (mid-work) |
| --- | --- | --- |
| sequence-changing | 10,881 | **83** |
| unapplicable | 61,765 | ~42,000 |

The second row is the problem. "Unapplicable" conflates three quite different severities that the harness never separated:

- genuine overlapping members,
- `hgvs_to_spdi` conversion failure,
- output the parser rejects.

An attempt to decompose it did not land and was abandoned. **The figure is also stale**: it predates both the junction-barrier fix and the duplication re-spelling (#1263) that landed later in #1259, either of which plausibly moves it. No number for three-member coverage should be quoted until it is re-measured with a harness that separates the causes.

## Why it matters

Three-member alleles are substantially worse than two-member ones — the same sweep shape reports 0 overlapping for two members and tens of thousands for three. The defects filed from this work (#1260, #1261, #1262, #1266, #1267) were all found at two or three members by hand or by an external property test, not by the committed sweeps.

## Suggested shape

A three-member sweep alongside the existing two-member ones, with:

- causes separated (overlap / conversion failure / parse rejection / sequence change) rather than collapsed into one "unapplicable" count;
- known-broken shapes excluded by name and pinned, the way `a_five_prime_duplication_beside_a_deletion_is_a_known_residual` is, so the exclusions stay visible;
- the exclusion list cross-referenced to the issues above.

Note `FERRO_ASSERT_REPARSE` (added in #1259) already covers the parse-rejection cause suite-wide, so a new sweep need not re-derive it.

Issue #1262, opened 2026-07-30
Normalizer: substitution + deletion and the spanning delins spelling reach two different fixed points


nh13:
## What's wrong

A second pair of spellings of one haplotype reaching two stable normal forms — distinct from #1260, and by a different suspected mechanism.

Reference: `("ACGT" x 64) + core + ("ACGT" x 64)` on `NC_TEST.1`, so `core[i]` is at 1-based position `257 + i`. With `core = "AAAAAA"` the poly-A run is **257–263**.

**core `AAAAAA`:**

| input | normalizes to |
| --- | --- |
| `NC_TEST.1:g.[258A>C;260del]` | `NC_TEST.1:g.[258A>C;263del]` |
| `NC_TEST.1:g.258_260delinsCA` | `NC_TEST.1:g.258_259delinsC` |

Same bases throughout — applied window `GTACAAAACGTACGTACGTA`, computed independently of the normalizer.

The `263del` is legitimate on its own: after the substitution the run is 259–263, so any single deletion inside it is equivalent. The disagreement is about whether the pair collapses at all.

## Severity

Non-confluence, not corruption — same class as #1260.

## Suspected mechanism — unverified, and different from #1260's

The separator veto rather than the column bound. `[258A>C;263del]` has two members separated by unchanged bases 259–262; the re-derivation collapses to a single piece spanning 258–259, which claims separator position 259, so the "two variants separated by one or more nucleotides should be described individually" rule (`general.md:34`) refuses it. The spanning spelling has no separators to veto.

**Scoping caveat:** as with #1260, the identifier this hypothesis names (`input_separator_positions`) exists only on `origin/fix/issue-1235-sequence-first-canonicalization` (#1237), not on `main`, while the defect reproduces without that code. Treat the mechanism as a hypothesis; the observations are solid.

## Reproduction

Verified at `e4a9a86` (the closed #1236) and at the tip of `fix/cis-sibling-crossing-normalization` (#1259). Pre-existing.

nh13:
## Correction: the mechanism, and the claim that it differs from #1260's

An adversarial review refuted two things in the body.

[...]

PR #1302, opened 2026-08-01
test(normalize): drop the indel model's adjacent-gap-insertion filter

### LINKED ISSUE #1294: Tests: the indel confluence generator still filters adjacent-gap insertions after the defects that justified it were fixed
## What's wrong

`indel_haplotype_strategy` in `tests/it/cis_allele_confluence_proptest.rs` filters out every haplotype with two insertions at **adjacent gaps**:

```rust
.prop_filter(
    "two insertions at adjacent gaps: #1260 / #1262 (see module docs)",
    |haplotype| !haplotype.has_adjacent_gap_insertions(),
)
```

That filter was added when such inputs produced overlapping output — `g.[258_259insA;259_260insA]` normalized to `g.[263dup;263dup]`, two members claiming one position. Its stated justification is #1260 / #1262.

**Both of the defects that motivated it are now fixed.** #1286 (two members settling on one junction) closed in #1289, and #1290 (a junction shifting past another junction) closed in #1293. The adjacent-gap-insertion shape now normalizes correctly:

```
g.[258_259insA;259_260insA]      -> g.257_263A[9]
g.[258_259insA;259_260insA;260_261insA] -> g.257_263A[10]
```

So the filter's justification is at least partly obsolete, and it has not been re-measured since. It may now be excluding coverage the property could have.

## Why it matters

This is the "coverage reads wider than it is" complaint of #1268 and #1283, inverted: a filter that outlived its reason makes the corpus **narrower** than the module doc claims, silently. The property is #1235's criterion 1 — *"an 'all encodings of one variant converge' property test"* — so a filter that excludes reachable shapes makes that criterion weaker than it reads.

Note the filter and the spanning-delins comparison are two separate restrictions with the same citation. The delins comparison genuinely is blocked on #1260 / #1262 — the derivation reaches the same piece set from both spellings and only the input-relative gates diverge. The **generator filter** is the one whose justification may have lapsed.

## Suggested first step

Remove the filter, run `an_indel_haplotype_is_authored_order_independent` and the (currently `#[ignore]`d) `an_indel_haplotype_normalizes_to_its_own_sequence`, and see what actually fails. Then either delete the filter or narrow it to the shapes that genuinely still fail, with the citation corrected to whatever is really blocking them.

Best done after #1292 is fixed, since that defect currently masks anything behind it in the same property — the pattern where each of #1286, #1287, #1290 and #1292 only became visible once its predecessor was closed.

## Status

`adjacent_gap_insertions_are_a_known_gap` pins #1260's and #1262's shapes as currently-broken and names what to re-enable when they are fixed. It does **not** cover this question — that test is about the delins comparison, not the generator filter.

## Relationship to #1235

Part of the "re-enable the held-back property checks" work for #1235's criterion 1. Not a normalizer defect — a test-coverage one.

nh13:
Closes #1294.

`indel_haplotype_strategy` excluded every haplotype with two insertions at adjacent gaps:

```rust
.prop_filter(
    "two insertions at adjacent gaps: #1260 / #1262 (see module docs)",
    |haplotype| !haplotype.has_adjacent_gap_insertions(),
)
```

**The citation does not describe the shape.** Neither #1260 nor #1262 is about overlapping members — both are non-confluence between the separate-member and spanning-`delins` spellings, which is a *different* restriction in the same file and stays exactly as it is. #1294 suspected the citation was doing double duty; re-measured, it was.

What the excluded shape actually hits is **#1301**, filed from this measurement: two adjacent-gap insertions reach a duplication and an insertion that denote the right bases but overlap and are rendered out of order.

```
core "GCATGAAAAT"
in   NC_TEST.1:g.[263_264insAC;264_265insAA]
out  NC_TEST.1:g.[264_265dup;264_265insCA]      sequence right, members overlap
```

That is only visible now. On the tip of #1293 the same input emitted `g.[263_264insCA;264_265dup]`, which *silently denotes different bases*; #1298's payload rotation is what turned it into a loud ordering failure. So this shape could not have been re-measured usefully before the stack below it landed — the masking pattern again.

## The measurement

| property | with the filter | without it |
| --- | --- | --- |
| `an_indel_haplotype_is_authored_order_independent` (runs in CI) | passes | **passes 30,000 cases** |
| `an_indel_haplotype_normalizes_to_its_own_sequence` (`#[ignore]`d on #1297) | passes 512, fails 30k on #1297 | reaches #1301 at its first generated case |

So the filter bought nothing for the property that actually runs, while making the corpus narrower than the module doc claimed — the "coverage reads wider than it is" complaint of #1268 / #1283, inverted. Deleted rather than narrowed: there is no sub-shape left that needs excluding for the running property, and hiding #1301 from the ignored one would only re-create the problem this issue is about.

## What is kept

`has_adjacent_gap_insertions` stays. `adjacent_gap_insertions_are_a_known_gap` uses it to assert that the haplotype it pins really is an adjacent-gap one, so that test cannot drift onto a different shape than the one it claims to pin. Its message about "the shape the strategy filters out" is corrected.

The module docs now record that #1260 / #1262 justify exactly one restriction, and the seed-file header no longer claims the first seed is rejected by the strategy — it runs now, and replays green (it is #1286's shape, fixed in #1289).

## Verification

- `cargo nextest run --features dev` — **8923 pass**, 0 fail
- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 …` — 8923 pass under both oracles
- `an_indel_haplotype_is_authored_order_independent` at 30,000 cases without the filter — pass
- Commit is signed

No `src/` changes: this is a test-coverage PR.

Stacked on #1299.

### DIFF
--- a/tests/it/cis_allele_confluence_proptest.rs
+++ b/tests/it/cis_allele_confluence_proptest.rs
@@ -53,6 +53,12 @@
 //!   decide differently, because both are measured against how the input
 //!   happened to be written. Tracked by #1260 and #1262 and pinned by
 //!   `adjacent_gap_insertions_are_a_known_gap`.
+//!
+//!   This is the **only** restriction those two issues justify. The generator
+//!   used to carry a second one citing them — a filter excluding two insertions
+//!   at adjacent gaps — which #1294 removed after re-measuring: neither #1260
+//!   nor #1262 is about overlapping members, so the citation did not describe
+//!   the shape it was excluding. What that shape actually hits is #1301.
 //! - `an_indel_haplotype_normalizes_to_its_own_sequence` is `#[ignore]`d
 //!   because it finds #1297, a cancelled member left as an identity member
 //!   overlapping the repeat that absorbed it. It found #1286, #1287, #1290,
@@ -690,12 +696,12 @@ impl IndelHaplotype {
         out
     }
 
-    /// Whether two insertions sit at **adjacent gaps** — the one shape this
-    /// model must currently exclude, tracked by #1260 and #1262.
+    /// Whether two insertions sit at **adjacent gaps**.
     ///
-    /// Excluded, not silently skipped: `adjacent_gap_insertions_are_a_known_gap`
-    /// below pins the shape as currently-broken, so fixing either issue fails
-    /// that test loudly and this filter comes out at the same time.
+    /// The strategy no longer filters this shape out (#1294). It is kept because
+    /// `adjacent_gap_insertions_are_a_known_gap` uses it to assert that the
+    /// haplotype it pins really is one, so that test cannot drift onto a
+    /// different shape than the one it claims to pin.
     fn has_adjacent_gap_insertions(&self) -> bool {
         self.events.windows(2).any(|pair| {
             matches!(pair[0], Event::Insert { .. })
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #1303, opened 2026-08-01
ci: halve the critical path (optimized soak archive, sweeps job, cache fixes)

nh13:
Stacked on #1300. Seven commits, each one measured change to `.github/workflows/`, cutting CI's critical path roughly in half.

## Headline

Measured on **job durations**, not workflow wall clock. Wall clock includes GitHub's runner queue, which varies enormously — one run in this series sat 295s in a queue before its build even started and looked like a 525s regression when its compute was actually an improvement. Critical path = slowest build + the slowest job that depends on it.

| | buildTest | buildSoak | slowestTest | slowestSoak | **critical path** |
| --- | --- | --- | --- | --- | --- |
| `main` | 194s | - | 227s | 271s | **421s** |
| + optimized soak archive | 181s | 259s | 211s | 49s | 392s |
| + sweeps job | 190s | 183s | 65s | 47s | 255s |
| + spec split | 152s | 196s | 89s | 52s | 248s |
| + LFS / feature trim | 132s | 178s | 78s | 40s | 218s |
| + cache keys | 123s | 155s | 70s | 42s | 197s |
| + LFS in generated-docs, **cache-warm** | **85s** | **136s** | 75s | 39s | **175s** |

**421s -> 175s, 2.4x.** The last row is the first run in which every re-keyed cache actually restored something matching what it builds — `Build test archive` compiles **1** crate there, against 316 before.

## Suggested reading order

1. **`ci(soak): build the soak shards from an optimized archive`** — the premise everything else builds on. The soak runs 1,000,000 proptest cases across 8 shards, so a constant factor on the code under test is multiplied a million times over. A `[profile.soak]` at `opt-level = 3` costs one extra compile and takes a property from 1.90s to 0.355s of user CPU at 5,000 cases. `debug-assertions` and `overflow-checks` stay **on**, so this is speed only, not a quieter build.

2. **`ci: run the exhaustive sweeps from the optimized archive`** — the biggest single win. `--partition hash:K/N` splits by test *name*, so it balances shards by count and is blind to cost; all three exhaustive-sweep tests happened to hash into the same shard. They gain even more from optimization than the soak does (8.3x / 7.6x / 8.7x), and they cost nothing extra to build because they already live in `tests/it/`.

3. **`ci: split spec generation out, and build the nightly soak once`** — two serialization fixes. The spec generators ran as `test-build`'s first two steps, ahead of the archive compile. And each of the nightly's 8 shards was compiling the whole suite from scratch *and* running both generators first, for an artifact no proptest module reads.

4. **`ci: stop fetching LFS and dev features where nothing reads them`**, **`ci: stop fetching LFS in generated-docs`** — 149 MB of LFS fixtures fetched by six jobs that never read them, and `--features dev` dragging `web-service` and `benchmark` into an `opt-level = 3` build.

5. **`ci: let the soak cache follow the build flags`**, **`ci: give test-build its own cache key, and split the oracle three ways`** — the subtlest bug here, worth reading closely. See below.

## The cache bug

`actions/cache` **skips the save on an exact key hit** — the log says so outright: `Cache hit occurred on the primary key ..., not saving cache`. A key derived from `Cargo.lock` alone is therefore frozen at whatever the *first* run for that lockfile wrote, and no later run can refresh it.

`test-build` compounded that by *sharing* its key with `generated-docs`. The two need different artifacts: `generated-docs` runs `cargo run --example` in the dev profile, while `test-build` archives test binaries under `CARGO_PROFILE_TEST_DEBUG=0`, which re-fingerprints every dependency. One frozen cache served the first and was useless to the second. Measured on run 30684270515: `generated-docs` compiled 3 crates, `test-build` compiled **316** — the whole tree from `proc-macro2` up — on every single run.

Each cache now hashes its workflow file into the key and carries a distinct suffix. The effect is visible directly in the logs: `Build test archive` went from 316 `Compiling` lines to **1**, and from 194s on `main` to **85s**.

Note that the first run after any key change is a miss — it restores from the lockfile-only `restore-key`, rebuilds, and *saves*. The run after that is the one that benefits. Both are in this PR's history if you want to see the shape.

## What is deliberately not changed

- **Coverage.** The sweeps run only in the new job, with both oracles on — those add assertions at `normalize_core_checked`'s single exit and change no behaviour, so an oracle-on run strictly subsumes the oracle-off one the `test` shards were doing. The filter is exact and defined once: `not test(proptest)` selects 8896 tests, `and not (SWEEP_FILTER)` selects 8867, and the filter alone selects the remaining 29.
- **`test` and `test-oracle` keep `lfs: true`**, and must. `clinvar_hgvs_tests`, `cmrg_exhaustive_tests` and `paraphase_exhaustive_tests` print `Skipping: ... not found` and return **green** when the fixtures are absent. Dropping LFS there would not redden CI; it would silently delete coverage while making the job look faster.
- **Dependency features.** The expensive tree (`reqwest` → hyper, rustls, async-compression) is an unconditional dependency, so no feature flag reaches it. Making it optional is a library change, out of scope here.

## Rejected after measurement

- **Lowering the opt-level.** Compile time is flat: 61.3s / 69.1s / 61.9s at opt-levels 1 / 2 / 3. There is no tradeoff curve.
- **One optimized archive for every job.** `opt-level = 3` for `it` alone (62s) already costs about what the dev profile costs for all 46 targets (66s), so merging the two builds would make the graph worse.
- **Caching the synthetic provider per proptest case.** Measured at ~1us per case, about 1.4% of a property, with a near-zero hit rate besides.

## Follow-ups not taken here

- A faster linker (lld/mold) for the two archive builds. The only untested idea with real upside, but `RUSTFLAGS` is not in the cache key, so measuring it honestly needs two runs.
- `Build` (168s) and `Python Wheel Test` (146s) are off the critical path — optimizing them saves runner minutes, not wall clock.

## Summary by CodeRabbit

* **Tests**
  * Expanded continuous integration coverage with additional test and oracle shards.
  * Added dedicated exhaustive sweep validation with normalization checks.
  * Improved test artifact reuse across standard, optimized, and soak runs.

* **Chores**
  * Optimized nightly soak testing across eight shards running 125,000 cases each.
  * Added a dedicated soak build profile for faster, more efficient test execution.
  * Updated required-status validation to include exhaustive sweeps.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -216,9 +216,76 @@ jobs:
             ${{ runner.os }}-cargo-
       - run: cargo clippy --all-features --all-targets -- -D warnings
 
-  # Build the test binaries and the two generated spec artifacts ONCE, then fan
-  # out: every downstream job consumes this job's `nextest archive`, so the
-  # crate compiles a single time no matter how many shards run.
+  # The two generated spec artifacts, split out of `test-build`.
+  #
+  # These ran as the first two steps of `test-build`, which serialized ~52s of
+  # generation ahead of the ~105s archive compile inside one job. They share
+  # nothing with the archive but the library crate: the generators are `[[bin]]`
+  # targets that read the spec submodule at RUNTIME, and what they emit is a
+  # pair of JSON files the shards download. Running them concurrently with the
+  # archive build takes that serialization off the critical path.
+  #
+  # Generation is itself the gate on the curated overrides — plain generation,
+  # not `--check` — so this job failing means the overrides no longer resolve
+  # against the spec checkout. That signal is now its own job rather than
+  # buried in a build.
+  spec-fixtures:
+    name: Spec fixtures
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          lfs: true
+          # Required here, and ONLY here: the generators harvest HGVS strings
+          # from the vendored spec checkout at runtime.
+          submodules: true
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8  # stable
+        with:
+          toolchain: stable
+      - uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-specgen
+          restore-keys: |
+            ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-specgen
+      - name: Generate HGVS v21.0 spec fixture
+        run: cargo run --features dev --bin generate_spec_fixture
+      - name: Generate HGVS spec test enumeration
+        run: cargo run --features dev --bin generate_spec_enumeration
+      - name: Stage the generator binaries for the shards
+        # `spec_generator_preconditions` executes these two binaries. The shards
+        # run from the nextest archive with no build state, so they must not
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #1304, opened 2026-08-01
Normalizer: two adjacent-gap insertions plus a deletion denote the wrong sequence


nh13:
## What's wrong

Three length-changing members normalize to a description that denotes different bases. The inserted `G` ends up one position 3' of where the input puts it.

```
reference  ("ACGT" x 64) + "GCATGAAAAT" + ("ACGT" x 64)
           so core[i] is at 1-based position 257 + i

in   NC_TEST.1:g.[260_261insGA;261_262insA;264del]
out  NC_TEST.1:g.[261_262insA;261_262dup;265del]

  intended  G C A T G A G A A A A T
  emitted   G C A T G A A G A A A T
```

## Severity

**Silent.** The output is well-formed, its members are disjoint and in order, `parse_hgvs` accepts it, and the SPDI apply oracle applies it cleanly — it simply denotes different bases. Neither `FERRO_ASSERT_REPARSE` nor `FERRO_ASSERT_IDEMPOTENT` can see it. Same class as #1292.

## Suspected mechanism — unverified

Two insertions at adjacent gaps (`260|261` and `261|262`) whose payloads do not commute (`GA` and `A`), plus a deletion 3' of both. The pair is exactly the shape #1290's junction clamp governs, and `payloads_commute("GA", "A")` is false, so the clamp should have barred the crossing. It did not — or something after it moved them. Not traced past that.

The output's `261_262insA` and `261_262dup` share a span, which is #1301's shape; that ordering is now fixed, but the ordering was never what was wrong here — the same two members in the other order denote the same wrong sequence.

## How it was found

By `cis_allele_confluence_proptest::an_indel_haplotype_normalizes_to_its_own_sequence` after #1301's ordering fix, which is what stopped that shape failing on the ordering check first. Minimal proptest input:

```
core "GCATGAAAAT", insert "GA" after index 3, insert "A" after index 4, delete index 7
```

Verified **pre-existing**: identical wrong sequence with `src/hgvs/variant.rs` reverted to the tip of #1302, so the ordering change neither caused nor masked it.

## Relationship to #1235

Blocks criterion 1 — enabling that property — alongside #1297. It does not violate criteria 2 or 3: the members are disjoint, ordered, and the result is a fixed point. It is a sequence-correctness defect in the family of #1266/#1267/#1279/#1280/#1292.

PR #1305, opened 2026-08-01
fix(normalize): order two members sharing a span by where they add sequence

### LINKED ISSUE #1301: Normalizer: adjacent-gap insertions normalize to a dup and an insertion that overlap and are out of order

nh13:
Closes #1301.

`cis_member_order_key` breaks a start tie with the member's **end** (#1261). That is not enough for two members whose spans are *identical*:

```
in   NC_TEST.1:g.[263_264insAC;264_265insAA]
out  NC_TEST.1:g.[264_265dup;264_265insCA]      <- before
out  NC_TEST.1:g.[264_265insCA;264_265dup]      <- after
```

`264_265insCA` fills the gap `264|265`, so it adds at interbase **264**; `264_265dup` places its copy after its last base, so it adds at **265**. Both span `264_265`, so start and end tie and the order fell through to the descriptor tie-break — which sorts `264_265dup` first, the reverse of the order the two apply in. The pair rendered out of order and their interbase spans read as overlapping, which is #1235's second acceptance criterion.

## What was and was not wrong

**The sequence was never wrong.** Both members applied denote the input's bases, before and after. What was wrong is the rendering, so the new test asserts the ordering *and* re-applies the output through `hgvs_to_spdi` — an ordering test that could pass by tidily rendering a wrong pair would be worse than none.

`junction_rank` adds the missing discriminator: a member adding at its span's **start** (an insertion, whose span is the gap it fills) sorts before one acting across the whole span. Everything that consumes the bases under its span ranks with the 3' end.

## Reachability

This shape only became reachable in #1302, which dropped the generator filter that had been excluding adjacent-gap insertions — the filter was hiding it under a citation (#1260 / #1262) that did not describe it.

## What is behind it

Fixing the ordering unmasks **#1304**, filed from this work: the same two-insertion shape *with a deletion added* denotes the wrong bases, silently.

```
in   NC_TEST.1:g.[260_261insGA;261_262insA;264del]
out  NC_TEST.1:g.[261_262insA;261_262dup;265del]
  intended  G C A T G A G A A A A T
  emitted   G C A T G A A G A A A T
```

**Pre-existing**, verified by reverting `src/hgvs/variant.rs` to this branch's parent and re-running: the same two members in the opposite order denote the same wrong sequence. So this change neither causes nor masks it — it only stops that shape failing on the ordering check first. The confluence property's doc comment now names #1304 as what sits behind #1297.

That is the seventh defect this property has surfaced, each masked behind the one before it: #1286 → #1287 → #1290 → #1292 → #1296 → #1301 → #1304.

## Verification

- `cargo nextest run --features dev` — **8925 pass**, 0 fail
- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 …` — 8925 pass under both oracles
- Commit is signed

Stacked on #1302.

### DIFF
--- a/src/hgvs/variant.rs
+++ b/src/hgvs/variant.rs
@@ -1146,7 +1146,7 @@ fn rna_simple_range(interval: &crate::hgvs::interval::RnaInterval) -> Option<Sel
 ///   is the exact order-dependence #1098 is about.
 pub(crate) fn cis_member_order_key(
     v: &HgvsVariant,
-) -> (String, SelfCancellingPoint, SelfCancellingPoint, String) {
+) -> (String, SelfCancellingPoint, SelfCancellingPoint, u8, String) {
     let accession = v.accession().map(Accession::full).unwrap_or_default();
     let descriptor = format!("{v}");
     // `SelfCancellingPoint` derives `Ord` over `(region, base, offset)` in that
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

PR #1306, opened 2026-08-01
fix(normalize): read a junction barrier's payload from its own snapshot

### LINKED ISSUE #1304: Normalizer: two adjacent-gap insertions plus a deletion denote the wrong sequence

nh13:
Closes #1304.

`clamp_sibling_crossing_junctions` compares this member's payload against each sibling's to decide whether a crossing is observable — commuting payloads may cross, others may not (#1290). It considers each sibling from **both** snapshots, before and after per-member normalization, so a sibling that moved still counts as a barrier where it started.

It read every payload off the **after** variant, including for spans taken from the **before** snapshot. A duplication reads its payload from the reference *under its own span*, so that pairing reads bases belonging to neither:

```
in   NC_TEST.1:g.[260_261insGA;261_262insA;264del]
  per-member ->  g.[261_262dup;265dup;265del]

the `insA` sibling's *before* span 261_262 measured against its *after*
`265dup` reports the payload `GA`, not `A`
```

`GA` commutes with this member's own `GA`, so the barrier vanished and the `insGA` junction swept from 260 straight past the sibling at 261:

```
out  NC_TEST.1:g.[261_262insA;261_262dup;265del]
  intended  G C A T G A G A A A A T
  emitted   G C A T G A A G A A A T
```

Silent: both members well-formed, disjoint, in order, a fixed point, and the SPDI oracle applies it cleanly. Neither `FERRO_ASSERT_REPARSE` nor `FERRO_ASSERT_IDEMPOTENT` can see it.

## The fix

Carry each span with the variant it was read from, so a before-span is measured against the before-edit. Two lines: the `flat_map` pairs `(&before[j], pre[j])` and `(&after[j], post[j])` instead of `j` alone, and the payload lookup uses that variant.

```
after:  NC_TEST.1:g.[260_261insGA;261_262insA;265del]
```

The two insertions hold their authored junction order and the deletion 3'-shifts alone.

## Scope

The bug is specific to the payload comparison. `onto_bases` in the same pass reads only span geometry, and `let payload = junction_payload(&after[i], kind, a, provider)` was already consistent — `a` is `post[i]`.

The new test also pins the guard: each insertion **alone** still reaches its standalone 3'-most spelling (`g.261_262dup`, `g.265dup`), so the barrier is a sibling effect and not a suppression of the 3' rule.

## Where this sits

Seventh in the chain the confluence property has surfaced, each masked behind the one before it: #1286 → #1287 → #1290 → #1292 → #1296 → #1301 → #1304. With this fixed the property's next failure is **#1297** again — a cancelled member left as an identity member overlapping the repeat that absorbed it — which is the last one before it can be enabled.

## Verification

- `cargo nextest run --features dev` — **8928 pass**, 0 fail
- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 …` — 8928 pass under both oracles
- Commit is signed

Stacked on #1305.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -3424,6 +3424,15 @@ pub(crate) fn clamp_sibling_crossing_junctions<P: ReferenceProvider>(
     };
     let pre: Vec<Option<MemberSpan>> = before.iter().map(|v| member_span(v, kind)).collect();
     let post: Vec<Option<MemberSpan>> = after.iter().map(|v| member_span(v, kind)).collect();
+    // The members `post` was measured from. The loop below translates `after`
+    // in place, so by the time member `i` inspects a sibling `j < i` that has
+    // already moved, `after[j]` no longer matches `post[j]` — pairing them
+    // reads a duplication's payload from bases under a span it has left, which
+    // is the #1304 mismatch one iteration later rather than one snapshot over.
+    // Keeping the pair together also keeps the barrier evaluated against the
+    // state the whole pass started from, so the result does not depend on the
+    // order the members happen to be visited in.
+    let post_snapshot: Vec<HgvsVariant> = after.to_vec();
 
     for i in 0..after.len() {
         let (Some(b), Some(a)) = (pre[i].as_ref(), post[i].as_ref()) else {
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

Issue #1308, opened 2026-08-01
Normalizer: two equal-payload insertions at adjacent gaps both shift 3' and denote the wrong sequence


nh13:
## What's wrong

Two insertions of the **same** payload at adjacent gaps normalize to a pair that denotes different bases.

```
reference  ("ACGT" x 64) + "CAGAAGATGAATAA" + ("ACGT" x 64)
           so core[i] is at 1-based position 257 + i

in   NC_TEST.1:g.[263_264insTG;264_265insTG]
out  NC_TEST.1:g.[264_265insTG;264_265dup]

  intended  C A G A A G A T G T T G G A A T A A
  emitted   C A G A A G A T T G G T G A A T A A
```

Both insertions moved one position 3', which is not equivalent here — the surrounding sequence is not periodic, so the pair does not commute with the base it stepped over.

## Severity

**Silent.** Two well-formed members on distinct junctions, disjoint, in order, a fixed point, and the SPDI apply oracle applies it cleanly. Same class as #1292 and #1304.

## Suspected mechanism — unverified

The payloads are *identical*, so `payloads_commute("TG", "TG")` is true and `clamp_sibling_crossing_junctions` deliberately permits the crossing — commuting payloads have no observable order, and letting them meet is what allows `coalesce_members_at_one_junction` to merge them into one member (#1286).

That reasoning is about the order of the **two payloads relative to each other**. It does not license moving *both* junctions 3' over an intervening reference base, which is what happened here: the pair went from junctions 263/264 to 264/265. Commuting says `p ++ q == q ++ p`; it says nothing about `ref[j] ++ p` versus `p ++ ref[j]`.

So the likely gap is that the commuting exemption is applied to the crossing without also requiring each payload to be in phase at its destination — the rotation rule #1298 added for the *clamped* direction, not applied here because no clamp fires at all.

Not traced past that.

## How it was found

By `cis_allele_confluence_proptest::an_indel_haplotype_normalizes_to_its_own_sequence` after #1297 was fixed — the eighth defect that property has surfaced, each masked behind the one before it: #1286 → #1287 → #1290 → #1292 → #1296 → #1301 → #1304 → this.

Verified **pre-existing**: identical output on the tip of #1306, with #1297's fix reverted. Only reachable since #1302 dropped the generator filter that had been excluding adjacent-gap insertions.

Minimal proptest seed (committed, inert while the property is `#[ignore]`d):

```
[...]

PR #1309, opened 2026-08-01
fix(normalize): drop a cancelled member left inside the one that absorbed it

### LINKED ISSUE #1286: Normalizer: two insertions at adjacent gaps in a homopolymer emit two identical dup members claiming one junction

### LINKED ISSUE #1297: Normalizer: a cancelled member is left as an identity member overlapping the repeat that absorbed it

nh13:
Closes #1297.

A `del` and an `ins` that merge into a `delins` restating the reference cancel to `=`. The member that absorbed the change then grows over the position that `=` names, and the pair overlaps:

```
in   NC_TEST.1:g.[261_262insAA;263del;264_265insA]
out  NC_TEST.1:g.[262_265A[6];265=]
```

`262_265A[6]` alone denotes the input exactly — the repeat is right. The `265=` residue claims a base the repeat covers, so the two overlap and the apply oracle declines the output, which is #1235's second acceptance criterion.

Traced, not guessed. The merge loop reaches `[262_263dup;265delinsA]`, the `265delinsA` normalizes to `265=` (correctly — it restates the reference), and the dup then becomes the tract-wide repeat that swallows it.

## Why not just drop identity members

Because `g.[1002=;1005del]` is a real description — the `=` records a position that was examined and found unchanged, and `tests/it/issue_851_compound_utr_projection.rs` asserts exactly that shape. Dropping identities outright would lose real information.

The rule is narrower: an identity member is dropped only when a sibling **claims the bases** under it. That is the case where the two contradict *and* overlap. A sibling that merely adds sequence at a junction inside it contradicts nothing, and a wholly self-cancelling allele has no claiming sibling at all — so `g.[263del;264_265insA]` still reports `g.266_267=`. All three cases are pinned as tests.

The pass runs last in the merge loop, once every span above it has settled.

## Where this sits

Eighth in the chain the confluence property has surfaced, each invisible until its predecessor closed: #1286 → #1287 → #1290 → #1292 → #1296 → #1301 → #1304 → #1297.

Behind it is **#1308**, filed from this work and **verified pre-existing** (identical output on #1306's tip with this fix reverted): two insertions carrying the *same* payload at adjacent gaps both shift one position 3'.

```
in   NC_TEST.1:g.[263_264insTG;264_265insTG]     core "CAGAAGATGAATAA"
out  NC_TEST.1:g.[264_265insTG;264_265dup]
  intended  C A G A A G A T G T T G G A A T A A
  emitted   C A G A A G A T T G G T G A A T A A
```

Equal payloads commute, so `clamp_sibling_crossing_junctions` deliberately permits the crossing — but commuting governs the two payloads' order relative to *each other*, not their phase against the reference base they stepped over. The property stays `#[ignore]`d on it, with its seed committed and the ignore reason naming it.

## Verification

- `cargo nextest run --features dev` — **8931 pass**, 0 fail
- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 …` — 8931 pass under both oracles
- Commit is signed

Stacked on #1306.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -4114,6 +4114,73 @@ fn translate_junction_member<P: ReferenceProvider>(
     }
 }
 
+/// Drop a cis member asserting **no change** at a position another member
+/// changes.
+///
+/// An identity member (`265=`) says a position was examined and found
+/// unchanged. That is real information when it stands alone, so `g.[1002=;
+/// 1005del]` keeps both. It is a contradiction when a sibling claims the same
+/// bases, and the pair also overlaps — two members on one position, which
+/// #1235's second criterion forbids and the apply oracle declines.
+///
+/// The shape is not authored, it is derived: a `del` and an `ins` that merge
+/// into a `delins` restating the reference cancel to `=`, and the member that
+/// absorbed the change then grows over it (#1297).
+///
+/// ```text
+/// g.[261_262insAA;263del;264_265insA]  ->  g.[262_265A[6];265=]
+///   the repeat alone denotes the input; the `265=` residue overlaps it
+/// ```
+///
+/// Only the identity member is dropped, and only when a sibling *claims* the
+/// bases under it — a sibling that merely adds sequence at a junction inside it
+/// contradicts nothing.
+pub(crate) fn drop_identity_members_covered_by_siblings(
+    members: &mut Vec<HgvsVariant>,
+    phase: AllelePhase,
+    uncertain: bool,
+) {
+    if phase != AllelePhase::Cis || uncertain || members.len() < 2 {
+        return;
+    }
+    let Some(kind) = cis_kind_of(&members[0]) else {
+        return;
+    };
+    let spans: Vec<Option<MemberSpan>> = members.iter().map(|v| member_span(v, kind)).collect();
+    let is_identity = |v: &HgvsVariant| {
+        matches!(
+            cis_axis_parts(v, kind).map(|(_, _, _, _, edit)| edit),
+            Some(NaEdit::Identity { .. })
+        )
+    };
+
+    let covered: Vec<bool> = (0..members.len())
+        .map(|i| {
+            let (Some(span), true) = (spans[i].as_ref(), is_identity(&members[i])) else {
+                return false;
+            };
+            (0..members.len()).filter(|&j| j != i).any(|j| {
+                spans[j].as_ref().is_some_and(|s| {
+                    s.claims_bases
+                        && s.region == span.region
+                        && s.accession == span.accession
+                        && s.start <= span.start
+                        && s.end >= span.end
+                })
+            })
+        })
+        .collect();
+    if !covered.iter().any(|&drop| drop) {
+        return;
+    }
+    let mut index = 0;
+    members.retain(|_| {
+        let keep = !covered[index];
+        index += 1;
+        keep
+    });
+}
+
 /// Whether two junction payloads may be reordered without changing what the
 /// allele denotes.
 ///
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -2680,6 +2680,14 @@ impl<P: ReferenceProvider> Normalizer<P> {
                 allele.uncertain,
                 &self.provider,
             );
+            // Last, once every span above has settled: a member that cancelled
+            // to `=` while a sibling grew over the bases it names is a
+            // contradiction, and an overlap (#1297).
+            merge::drop_identity_members_covered_by_siblings(
+                &mut result,
+                allele.phase,
+                allele.uncertain,
+            );
 
             pass += 1;
             // Stable once a full pass leaves the member set unchanged.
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai pause

PR #1310, opened 2026-08-01
test(normalize): strengthen the cis-allele test oracles to their stated contracts

nh13:
Four assertions in the cis-allele test suite claimed more than they checked. Each is fixed in its own commit; no production code changes and no expectation changes.

### 1. The two-member sweep counted an overlapping output as skipped

`shifts_never_change_the_sequence_across_an_exhaustive_two_member_sweep` destructured both applies at once, so an *output* with no well-defined resulting sequence landed in `skipped` next to an input the sweep genuinely cannot speak for. An overlapping normalization — two members claiming one base — is precisely the defect the sweep exists to catch. The sibling sweep in `cis_junction_crossing_shift.rs` already separates the two; this brings the pair back into step.

### 2. The re-parse invariant compared a 20-base window

`applied` returned `edited.chars().skip(254).take(20)` while its caller asserted the normalization "changed the sequence". These cases change length, so content can move out of the window or be pushed past its end and still compare equal. Now returns the full resulting sequence.

### 3. Negative controls asserted only the rendered string

The point of these controls is that the unclamped path is still correct, and correctness includes denoting the same bases. They now go through `assert_normalizes_preserving`, which the positive cases in the same files already use. Two sites that hand-rolled the preservation half inline (one dropping its failure message) now call the helper too.

### 4. CLAUDE.md omitted the re-parse oracle's debug-only gating

The idempotency section states it; the re-parse section, describing the same seam, did not. Verified against `reparse_self_check_enabled` in `src/normalize/mod.rs`.

## Verification

8906/8906 tests pass; clippy `-D warnings` and fmt clean. No case in the 128-sequence corpus trips the new overlapping-output assertion, and every pinned re-parse case still passes with the full-sequence comparison — so these strengthen the guards without moving any result.

## Provenance and scope

These were flagged by CodeRabbit on #1285 against a pre-rebase diff base. That PR no longer touches these files, so they were triaged out of it and verified independently here.

One related item is deliberately **not** here: `issue_1254_sibling_crossing_shift.rs` still hand-rolls the 5' preservation assertion. The direction-aware `assert_normalizes_preserving_in` helper it needs is added by #1285, so converting that site here would guarantee a conflict. It follows once #1285 lands.

## Summary by CodeRabbit

* **Tests**
  * Strengthened normalization regression coverage to verify that sequence content is preserved across duplication, deletion, insertion, and repeat-boundary cases.
  * Expanded invariant checks to cover complete generated sequences, including length and out-of-range differences.
  * Improved test reporting by distinguishing conversion failures, overlapping outputs, and sequence mismatches.
* **Documentation**
  * Clarified that the normalization debug check is disabled in release builds and optimized for minimal overhead in debug builds.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -64,6 +64,10 @@ description the parser accepts inbound but cannot re-read outbound is a
 parse/display round-trip asymmetry normalize merely carries through. Both are a
 different bug; reporting them here would bury the one this oracle is for.
 
+Like the idempotency oracle it is compiled out in release builds
+(`#[cfg(debug_assertions)]`) and read once into a `OnceLock`, so it adds no
+release-build cost and a disabled debug run pays only one atomic load.
+
 ```bash
 FERRO_ASSERT_REPARSE=1 cargo nextest run --features dev
 ```
--- a/tests/it/cis_junction_crossing_shift.rs
+++ b/tests/it/cis_junction_crossing_shift.rs
@@ -85,7 +85,7 @@ fn a_repeat_grown_from_an_insertion_does_not_span_a_sibling() {
 fn a_lone_duplication_still_reaches_repeat_notation() {
     // Negative control: no sibling, nothing to span, so the tract-wide repeat
     // is exactly right and must survive.
-    assert_eq!(normalize(TRACT, "TEMPLATE:g.1_2dup"), "TEMPLATE:g.1_9T[11]");
+    assert_normalizes_preserving(TRACT, "TEMPLATE:g.1_2dup", "TEMPLATE:g.1_9T[11]");
 }
 
 #[test]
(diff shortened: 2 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
Rebased onto `main` now that #1285 has merged, and added the fifth commit it was blocking.

#1285 introduced `assert_normalizes_preserving_in`, so the 5' site in `issue_1254_sibling_crossing_shift.rs` — the one this PR originally left alone to avoid a guaranteed conflict — is now converted. It was the weakest of the three copies: its apply comparison carried no failure message at all. That closes the CodeRabbit finding from #1285 in full; no hand-rolled preservation checks remain in these files.

8921/8921 pass on the rebase; clippy and fmt clean.

nh13:
@coderabbitai review

Issue #1312, opened 2026-08-01
Normalizer: equal-payload insertions at adjacent gaps merge out of phase with the base between them


nh13:
## What's wrong

Two insertions of the same payload at adjacent gaps denote the wrong sequence when the payload is **not** in phase with the base between them.

```
reference  ("ACGT" x 64) + "TAAAACCA" + ("ACGT" x 64)
           so core[i] is at 1-based position 257 + i
           260 = A, 261 = A, 262 = C

in   NC_TEST.1:g.[260_261insAC;261_262insAC]
out  NC_TEST.1:g.261_262insACCA

  intended  T A A A A C A A C C C A
  emitted   T A A A A A C C A C C A
```

## Relationship to #1308

Same family, and **not** fixed by #1308's fix. Before that fix the same input emitted `g.[261_262insAC;261_262dup]`, which denotes the *same wrong sequence* — so #1308 changed this case from two wrong members to one wrong member without correcting it. Verified by reverting `src/normalize/merge.rs` to the tip of #1309 and re-running.

#1308's fix bounds a commuting sibling at the junction it settles on, and walks back to the most 3' junction this member's payload is in phase with. That is correct as far as it goes — it fixes #1308's own reproduction, where the payload `TG` *is* in phase two positions 3'. Here it is not: moving `AC` from junction 260 to 261 needs `P[n-1] == ref[261]`, i.e. `C == A`, which fails. The member should therefore stay at 260, and the pair should render as the input does.

So the remaining gap is in what happens **after** the clamp declines the move — something downstream still merges the two members at one junction, with a payload (`ACCA`) that is neither `AC ++ AC` nor any rotation of it.

## Suspected mechanism — unverified

`coalesce_members_at_one_junction` merges members that share a junction and whose payloads commute, concatenating them. The two members here do not legitimately share a junction; if one of them was moved onto the other's junction by a path that skipped the phase check, the concatenation would then be computed from payloads that are no longer the ones each member carries at that position. Not traced.

## Severity

**Silent.** One well-formed member, a fixed point, parses, and the SPDI apply oracle applies it cleanly. Same class as #1292, #1304 and #1308.

## How it was found

[...]

PR #1313, opened 2026-08-01
fix(normalize): bound a commuting sibling at the junction it settles on

### LINKED ISSUE #1308: Normalizer: two equal-payload insertions at adjacent gaps both shift 3' and denote the wrong sequence

nh13:
Closes #1308.

`clamp_sibling_crossing_junctions` exempts a sibling whose payload **commutes** with this member's, because `p ++ q == q ++ p` means the two have no observable order — and letting them reach one junction is better than keeping them apart, since they then merge into a single member (#1286).

That reasoning covers the order of the two payloads relative to *each other, once they meet*. It does not cover a member sweeping past a sibling that **stays put** and landing beyond it, where different reference bases now separate them:

```
reference  ("ACGT" x 64) + "CAGAAGATGAATAA" + ("ACGT" x 64)
in   NC_TEST.1:g.[263_264insTG;264_265insTG]
       the second member does not move; the first 3'-shifts 263 -> 265
out  NC_TEST.1:g.[264_265insTG;264_265dup]

  intended  C A G A A G A T G T T G G A A T A A
  emitted   C A G A A G A T T G G T G A A T A A
```

Silent: two well-formed members on distinct junctions, disjoint, in order, a fixed point, and the SPDI oracle applies it cleanly.

## The fix

Two parts.

1. **A commuting sibling bounds this member at the junction it *settles* on**, and may be met rather than only approached — landing on it is what lets the pair merge. Only its post-normalization position counts, since where it *started* is exactly the order commuting makes unobservable. Bounding at its pre-normalization junction instead would split `g.[258_259insA;259_260insA]` back into two members, which is what the exemption exists to avoid — pinned as a test, along with the three-insertion case.
2. **Landing on that bound can itself be out of phase.** The clamp walks back from the bound to the most 3' junction this member's payload is in phase with, reusing `payload_at_junction`'s rotation rule from #1298. `before_junction` always qualifies — that is where the shift began — so the walk terminates.

```
after:  NC_TEST.1:g.265_266insTTGG
```

A non-commuting sibling keeps the stricter rule: one position short, and from either snapshot.

## What this does **not** fix

**#1312**, filed from this work. When the payload is out of phase with the base *between* the two gaps, something downstream still merges the pair at one junction with a payload that is neither concatenation nor rotation:

```
core "TAAAACCA"
in   NC_TEST.1:g.[260_261insAC;261_262insAC]
out  NC_TEST.1:g.261_262insACCA
  intended  T A A A A C A A C C C A
  emitted   T A A A A A C C A C C A
```

**Pre-existing and unchanged in effect** — verified by reverting `src/normalize/merge.rs` to this branch's parent, where the same input emits `g.[261_262insAC;261_262dup]`, denoting the *same wrong sequence*. So this PR changes that case from two wrong members to one wrong member without correcting it, and #1312 says so explicitly. The new test module's header says so too.

## The confluence property

**It passes its committed 512-case budget for the first time** — all nine committed seeds replay green. It fails a 60,000-case soak on #1312, so it **stays `#[ignore]`d** and the reason names #1312. Enabling it on a budget that cannot reach the known-failing shape would prove nothing.

Ninth in the chain, each masked behind the one before it: #1286 → #1287 → #1290 → #1292 → #1296 → #1301 → #1304 → #1297 → #1308 → #1312.

## Verification

- `cargo nextest run --features dev` — **8934 pass**, 0 fail
- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 …` — 8934 pass under both oracles
- Commit is signed

Stacked on #1309.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -3511,36 +3511,75 @@ pub(crate) fn clamp_sibling_crossing_junctions<P: ReferenceProvider>(
         // `265dup`, which reported the payload `GA` rather than `A`. `GA`
         // commutes with this member's `GA`, so the barrier vanished and the
         // member swept past it (#1304).
+        // Commuting payloads are exempt from the barrier, but only as far as the
+        // sibling's **settled** junction. Commuting says `p ++ q == q ++ p` — the
+        // two payloads have no observable order *once they meet*. It says
+        // nothing about a member sweeping past a sibling that stays put and
+        // landing beyond it, where reference bases now separate them differently:
+        //
+        //     core "CAGAAGATGAATAA"
+        //     g.[263_264insTG;264_265insTG]
+        //       the second member does not move; the first 3'-shifts 263 -> 265
+        //       intended  A T G T T G G
+        //       emitted   A T T G T G G          (#1308)
+        //
+        // So a commuting sibling bounds this member at the junction it settles
+        // on — landing *on* it is what lets the pair merge (#1286) — and only
+        // its post-normalization position counts, since where it started is
+        // exactly the order that commuting makes unobservable. A sibling whose
+        // payload does not commute keeps the stricter rule above: one short, and
+        // from either snapshot.
         let payload = junction_payload(&after[i], kind, a, provider);
         let across_junctions = (0..pre.len())
             .filter(|&j| j != i)
             .flat_map(|j| {
                 [
-                    (&before[j], pre[j].as_ref()),
-                    (&post_snapshot[j], post[j].as_ref()),
+                    (false, &before[j], pre[j].as_ref()),
+                    (true, &post_snapshot[j], post[j].as_ref()),
                 ]
             })
-            .filter_map(|(variant, span)| span.map(|s| (variant, s)))
-            .filter(|(_, s)| !s.claims_bases && s.region == a.region && s.accession == a.accession)
-            .filter(|(variant, s)| {
-                let blocks = match (&payload, junction_payload(variant, kind, s, provider)) {
-                    (Some(mine), Some(theirs)) => !payloads_commute(mine, &theirs),
+            .filter_map(|(settled, variant, span)| span.map(|s| (settled, variant, s)))
+            .filter(|(_, _, s)| {
+                !s.claims_bases && s.region == a.region && s.accession == a.accession
+            })
+            .filter_map(|(settled, variant, s)| {
+                let junction = s.junction?;
+                if junction <= before_junction || junction > after_junction {
+                    return None;
+                }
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

PR #1314, opened 2026-08-01
ci: link the archive builds with mold

nh13:
**Result: mold does not pay for itself here. Closing.**

Measured on run [30685719955](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/30685719955), attempt 2 — the cache-warm one. Both jobs compiled exactly **1** crate in that attempt, same as the baseline on #1303, so this is apples-to-apples: the difference is codegen for `ferro-hgvs` plus linking, and nothing else.

Comparing the build step alone:

| build step | GNU ld (#1303) | mold | delta |
| --- | --- | --- | --- |
| `Build test archive` | 58s | 57s | -1s |
| `Build soak archive (optimized)` | 122s | 114s | -8s |
| *`Install mold` step* | - | 15-19s | +15-19s |

Whole-job durations, which is what actually reaches the critical path:

| | GNU ld | mold |
| --- | --- | --- |
| `Build test archive` | 85s | 110s |
| `Build soak archive (optimized)` | 136s | 153s |
| critical path | **175s** | 196s |

mold saves 1-8s of link time and costs 15-19s to install, so it is a net loss of about 21s on the critical path.

## What this establishes

**These builds are codegen-bound, not link-bound.** That is consistent with the other negative result from this campaign: compile time was flat at 61.3s / 69.1s / 61.9s across opt-levels 1 / 2 / 3, which is also what you would expect if codegen of `ferro-hgvs` itself dominates. Two independent measurements now point the same way.

Two reasons mold had little to grab, in hindsight:

- `test-build` already sets `CARGO_PROFILE_TEST_DEBUG=0`. Heavy debug info is exactly what mold is best at chewing through, and it has already been stripped.
- In the warm state these jobs compile one crate, so the remaining time is `ferro-hgvs` codegen — which no linker can touch.

Shipping a cheaper install (a prebuilt tarball instead of apt, ~2s) would make it roughly break-even on `test-build` and about 6s better on the soak archive. That is inside run-to-run noise on these runners, and not worth an extra moving part in CI for.

Recorded rather than silently dropped, so the next person asking "have we tried a faster linker?" has the numbers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -304,6 +304,13 @@ jobs:
     name: Build test archive
     runs-on: ubuntu-latest
     env:
+      # Link with mold. Set here rather than in `.cargo/config.toml` so it
+      # applies only to CI and stays visible next to the job that pays for it.
+      # Changing this value changes the cache key, because the key hashes this
+      # file — which is exactly what has to happen: RUSTFLAGS re-fingerprints
+      # every crate, and a cache keyed without it would restore artifacts built
+      # by a different linker and rebuild them all, every run.
+      RUSTFLAGS: "-C link-arg=-fuse-ld=mold"
       # The Test jobs build and LINK the integration-test binary. With full
       # debuginfo (dev/test profile default `debug = true`) the linked
       # `target/` exhausts the stock runner's ~14 GB free disk and the linker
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
Closing: measured a net loss of ~21s on the critical path. mold saves 1-8s of link time and costs 15-19s to install. Full numbers in the description — the useful takeaway is that both archive builds are codegen-bound rather than link-bound, which matches the flat opt-level curve measured in #1303.

Issue #1284, opened 2026-07-31
Normalizer: del+dup collision repair is genomic-only, so c./n./r. still emit descriptions ferro cannot re-parse


nh13:
## What's wrong

`respell_colliding_duplications` (`src/normalize/merge.rs`) is gated to `CisKind::Genome | CisKind::Mt`. The del+dup collision it repairs occurs identically on `c.`/`n.`/`r.`, where it is left alone — so `normalize` emits a description its own parser rejects.

```
SyntheticBuilder::cds("TTTTTTTTTAATATATTTTAATAT", cds_start = 9, cds_end = 24, Strand::Plus)

in   NM_TEST.1:c.[-8del;-6dup]
out  NM_TEST.1:c.[-1del;-1dup]      <- two members claiming one position
     parse_hgvs(out) -> Err: Self-cancelling allele: variants at index 0 (del)
                             and 1 (dup) describe overlapping reference positions
```

This is the same defect class as #1263, which the genomic path now fixes.

## Why it matters more than it did

#1259 arms `FERRO_ASSERT_REPARSE=1` in `ci.yml` and `nightly-mutalyzer.yml`. That oracle asserts an invariant which is **known-false on the transcript axes**. Nothing in the suite reaches it today, so CI is green — but any new test that touches a transcript-axis del+dup collision will not fail with a diff, it will **panic and abort the run**.

## Why the obvious extension does not work

Removing the gate needs the axis→sequence conversion. `axis_sequence_delta` is the natural candidate and is **not sufficient on its own**: it returns `cds_start - 1`, which maps `c.1` onto the CDS start correctly but is off by one below it, because the CDS axis has no zero — `c.-1` is transcript position `cds_start - 1`, not `cds_start - 2`.

Wiring it up naively was tried and reverted. Two observed consequences:

- the 5'UTR read lands one base too far 5', so the repair re-spells the duplication from the wrong bases;
- it moved the answer on shapes that were already correct (`c.[9_10dup;12del]` went from `c.12dup` to `c.[10_11dup;13del]`).

A piecewise conversion (`axis + delta + 1` below zero, `axis + delta` at or above) is correct for the endpoints but was not enough to make the 5'UTR case re-parse, so something else in the path also assumes a positive axis.

## Suggested direction

[...]

Issue #1263, opened 2026-07-30
Normalizer: emits an allele its own parser rejects (del and dup claiming one position)


nh13:
## What's wrong

`normalize` emits a description **ferro's own parser rejects**.

Reference: `("ACGT" x 64) + core + ("ACGT" x 64)` on `NC_TEST.1`, so `core[i]` is at 1-based position `257 + i`.

**core `GCAACAGCGTAAAC`:**

```
in   NC_TEST.1:g.[260_261del;262_263insA;264delinsAC]
out  NC_TEST.1:g.[261_262del;262dup;263_264insA]

re-parsing that output:
  Parse error at position 12: Self-cancelling allele: variants at index 0 (del)
  and 1 (dup) describe overlapping reference positions
```

The output denotes the right bases — applied window `GTGCAAAGACGTAAACACGT`, same as the input. The problem is purely that the string is not valid HGVS by ferro's own grammar.

## Why this is worth separating from #1260 / #1261 / #1262

Those three are arguments about which of several valid spellings is canonical, and their fixes depend on policy calls about the delins corpus. This one is not a policy question. Whatever the right canonical form is, `parse(normalize(x))` must succeed, because `normalize` is documented to return a valid HGVS description and every consumer chaining a second call depends on it.

It is also **a hole in the existing idempotency oracle**, not merely a defect beside it. `FERRO_ASSERT_IDEMPOTENT` verifies by re-normalizing its own output, which it cannot do for an output that will not parse — so an unparseable result is precisely the shape that oracle cannot see.

## Mechanism

`262_263insA` canonicalises to `262dup`, and the sibling-crossing clamp permits a deletion to sit flush against that junction. Flush is correct while the member is spelled `ins`; it is an overlap once it is spelled `dup`, because a duplication's span *is* the base it copies.

## Why the obvious fix does not work

Tightening the junction barrier by one base — so a deletion stops short of a junction rather than flush against it — fixes this and **breaks #1135**, where a deletion must reach the junction for the self-cancelling del+ins merge to fire. Distinguishing the two is a sequence question (does the inserted sequence cancel the deleted one, or merely collide with it?), not a positional one.

## Scope, measured

[...]

PR #1315, opened 2026-08-01
fix(normalize): repair del+dup collisions on the non-coding axis

nh13:
Closes part of #1284 — the `n.` axis. `c.` and `r.` are deliberately left, and the issue stays open for them.

## The defect

`respell_colliding_duplications` (#1259) resolves the #1263 collision — a duplication whose junction lands inside a sibling's span, which ferro's own parser rejects. It was gated to `CisKind::Genome | CisKind::Mt`, so on the transcript axes normalization emitted a description it cannot read back:

```text
reference   ATGGCAACAGCGTAAACGGCTAGCTAGCTA
in    n.[7_8del;9_10insA;11delinsAC]
out   n.[8_9del;9dup;10_11insA]
      Parse error: Self-cancelling allele: variants at index 0 (del) and 1 (dup)
      describe overlapping reference positions
```

The same edit on `g.` reaches `g.[261del;263_264insA]` and is a fixed point — pinned as one by `every_sibling_pass_is_idempotent_on_a_compound_allele`. Only the axis differed.

## Why `n.` and not `c.`/`r.`

#1284's "why the obvious extension does not work" is about the CDS-start discontinuity: the pass reads the duplicated bases over the member's own coordinates, and on a CDS-relative axis those do not index the served sequence — `c.-1` is `cds_start - 1`, not `cds_start - 2`, so a single delta is off by one below the CDS start. That was tried and reverted.

That blocker is specific to the CDS-relative axes. `n.` is transcript-relative: `axis_frame` already returns `delta: 0` for `CisKind::Tx`, exactly as for the genomic axes. So the whole non-coding axis is reachable with **no conversion at all**, and the original "genomic axes only" gate was over-broad — it grouped `n.` with axes it does not resemble.

## Two gates, not one

Worth knowing for whoever does `c.`/`r.`: the visible kind check is not the only gate. `respell_at_gap` also matched only `Genome`/`Mt` and silently reverted everything else, so lifting the first alone changed nothing — the tests still failed, byte-identically. Its new `Tx` arm refuses a gap that would name `n.0`, the guard `place_signed` already applies for `respell_as` on the same axis.

## Verification

8909 pass, and **8909 with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1`**. That matters here specifically: #1284 warns that #1259 armed the re-parse oracle on an invariant known-false for the transcript axes, so a future transcript-axis test would panic and abort the run rather than fail with a diff. This narrows that exposure to `c.`/`r.`.

The new tests assert the output re-parses and is a fixed point, not just that it matches a string — the defect is an unparseable spelling, so a string-only test would still pass if someone pinned the broken output. One test checks sequence preservation through `hgvs_to_spdi`, a different code path from the repair.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved variant normalization for transcript-relative (`n.`) duplication and deletion collisions.
  * Transcript duplications are now converted to equivalent literal insertions when sequence information is available.
  * Added validation for signed transcript coordinates, including rejection of invalid `n.0` endpoints.
  * Preserved compatibility and sequence accuracy across repeated normalization.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -4011,10 +4011,22 @@ fn shift_signed(base: &mut i64, delta: i64) -> bool {
 /// (`g.[261_262del;262_263insA;…]` reduces to `g.[261del;…]`), so the loop has
 /// to see the re-spelled form to settle on that reduction.
 ///
-/// Genomic axes only. On `c.`/`n.`/`r.` the 1-based coordinate is not a direct
-/// offset into the fetched sequence, and the conversion this would need already
-/// exists only inside `collapse_overlapping_cis_edits`; rather than duplicate
-/// it, a transcript-axis collision is left alone.
+/// Runs on every axis whose 1-based coordinate is a direct offset into the
+/// sequence the provider serves — the genomic axes and `n.`.
+///
+/// That is the whole requirement, because the pass reads the duplicated bases
+/// over the member's own coordinates. `axis_frame` states which axes qualify:
+/// it returns `delta: 0` for `Genome`, `Mt` **and** `Tx`, since a non-coding
+/// transcript's positions index its sequence directly, exactly as a contig's
+/// do. The original gate said "genomic axes only", which was over-broad — it
+/// grouped `n.` with the CDS-relative axes it does not resemble (#1284).
+///
+/// `c.` and `r.` are still refused. Those are CDS-relative (`delta:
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #1316, opened 2026-08-01
Normalizer: adjacent-gap insertions both re-spell as the same tract-wide repeat, and the allele loses its deletion


nh13:
## What's wrong

Two insertions at adjacent gaps plus a distant deletion normalize to **two identical repeat members** claiming the same tract.

```
reference  ("ACGT" x 64) + "CAGCCAGTCAGCGCATCAG" + ("ACGT" x 64)

in   NC_TEST.1:g.[261_262insAA;262_263insAA;269del]
out  NC_TEST.1:g.[259A[3];259A[3]]
```

Both members are the same repeat over the same span, so they overlap and the SPDI apply oracle declines the output.

## Severity

**Loud.** The apply oracle rejects it, so the confluence property's preservation check catches it rather than it slipping past. Same shape family as #1286 (`g.[263dup;263dup]`, two dup members at one junction), which was fixed in #1289 — this is its repeat-form analogue, and it also loses the deletion entirely.

## Suspected mechanism — unverified

Two adjacent-gap insertions of the same payload each grow the same `A` tract, and each is independently re-spelled as the tract-wide repeat `259A[3]` — so the pair describes the tract twice instead of once with a higher copy count. `coalesce_members_at_one_junction` merges members sharing a junction, but a repeat has no junction (`junction_of` returns `None` for `NaEdit::Repeat`), so it does not participate. `demote_repeats_spanning_siblings` handles a repeat that swallows a *sibling*, not two repeats that coincide.

The vanished `269del` is a second question and may be the same cause or a separate one. Not traced.

## How it was found

By `cis_allele_confluence_proptest::an_indel_haplotype_normalizes_to_its_own_sequence` at 60,000 cases, after #1312 was fixed — the tenth defect that property has surfaced, each masked behind the one before it: #1286 → #1287 → #1290 → #1292 → #1296 → #1301 → #1304 → #1297 → #1308 → #1312 → this.

Seed (committed, inert while the property is `#[ignore]`d):

```
cc 23c331742c910e4d748997c1e4e9d988fe3b4d032c3f60b1e995c9d052f3c720
core "CAGCCAGTCAGCGCATCAG", insert "AA" after index 4, insert "AA" after index 5, delete index 12
```

## Relationship to #1235

Criterion 2 — *"no normalized cis allele contains overlapping or out-of-order members."* A direct instance, and it currently blocks criterion 1 as well, since it is what the property fails on.

PR #1317, opened 2026-08-01
fix(normalize): test commuting against the payload a member lands with

### LINKED ISSUE #1312: Normalizer: equal-payload insertions at adjacent gaps merge out of phase with the base between them

nh13:
Closes #1312.

`clamp_sibling_crossing_junctions` lets a member meet a sibling whose payload **commutes** with its own: sharing a junction is well-defined there, and the pair then merges (#1286, #1308).

It tested commuting against the payload the member carries **now**. Landing on that junction rotates the payload into phase — and a rotation is exactly what can destroy the identity:

```
reference  ("ACGT" x 64) + "TAAAACCA" + ("ACGT" x 64)
in   NC_TEST.1:g.[260_261insAC;261_262insAC]

  `AC` and `AC` commute, so the first member was allowed onto junction 261
  at 261 its payload is `CA`, and `CA ++ AC != AC ++ CA`

out  NC_TEST.1:g.261_262insACCA
  intended  T A A A A C A A C C C A
  emitted   T A A A A A C C A C C A
```

Two members then shared a junction with **no order defined between them**, and the merge downstream concatenated them in *rendered* order — producing `ACCA`, which is neither payload nor any rotation of the pair. Silent: one well-formed member, a fixed point, parses, applies cleanly.

## The fix

Rotate first, then test. The commuting check now runs against `payload_at_junction(mine, …, junction, …)` — what this member would actually carry there. When that does not commute, the bound falls back to one position short, exactly as for any non-commuting sibling, and the pair stays apart.

```
after:  NC_TEST.1:g.[260_261insAC;261_262insAC]     (the input's own spelling)
```

This is a hole in **my own #1308 fix**, which introduced the "land on the sibling's settled junction" rule without re-checking commuting after the rotation that landing implies.

## One expectation updated

`issue_1308_commuting_payload_phase::a_commuting_payload_does_not_sweep_past_a_sibling_that_stays` expected `g.265_266insTTGG`; it now expects the input spelling `g.[263_264insTG;264_265insTG]`. `TG` landing on the sibling's junction becomes `GT`, which does not commute with the `TG` already there, so the merge this rule refuses no longer happens. The form it replaced was sequence-correct but reached through a landing that is not generally sound. That test is from #1313, not a long-standing blessed expectation.

The guard is pinned alongside it: in a poly-A tract every rotation of `A` is `A`, so the landed payload still commutes and `g.[258_259insA;259_260insA]` still merges to `g.257_263A[9]`.

## What is behind it

**#1316**, filed from this work: two adjacent-gap insertions each re-spell as the *same* tract-wide repeat, giving `g.[259A[3];259A[3]]` — two identical members over one tract, with the allele's deletion gone. It is the repeat-form analogue of #1286, which fixed the same shape spelled as duplications. Loud (the apply oracle declines it), and reachable only in a soak.

Tenth in the chain, each masked behind the one before it: #1286 → #1287 → #1290 → #1292 → #1296 → #1301 → #1304 → #1297 → #1308 → #1312 → #1316.

## Verification

- `cargo nextest run --features dev` — **8936 pass**, 0 fail, **0 aborts** (checked for `SIGABRT`/`SIGSEGV`/`TIMEOUT`, not just `FAIL`)
- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 …` — 8936 pass under both oracles
- Commit is signed

Stacked on #1313.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -3547,11 +3547,36 @@ pub(crate) fn clamp_sibling_crossing_junctions<P: ReferenceProvider>(
                 if junction <= before_junction || junction > after_junction {
                     return None;
                 }
-                let commutes = match (&payload, junction_payload(variant, kind, s, provider)) {
-                    (Some(mine), Some(theirs)) => payloads_commute(mine, &theirs),
-                    // A payload that cannot be read blocks: refusing to cross is
-                    // the conservative answer when the reordering cannot be
-                    // shown to be harmless.
+                // Commuting is tested against the payload this member would
+                // carry **at that junction**, not the one it carries now. The
+                // two differ: landing there rotates the payload into phase, and
+                // a rotation is exactly what can destroy the identity —
+                // `260_261insAC` moved onto junction 261 becomes `insCA`, and
+                // `CA` does not commute with the `AC` already sitting there,
+                // though `AC` and `AC` did. Sharing a junction with a payload it
+                // does not commute with leaves the pair's order undefined, and
+                // the merge downstream concatenates them in rendered order:
+                // `g.[260_261insAC;261_262insAC]` became `g.261_262insACCA`,
+                // denoting different bases (#1312).
+                let commutes = match (
+                    payload
+                        .as_ref()
+                        .and_then(|mine| {
+                            payload_at_junction(
+                                mine,
+                                after_junction,
+                                junction,
+                                &a.provider_key,
+                                provider,
+                            )
+                        })
+                        .as_deref(),
+                    junction_payload(variant, kind, s, provider),
+                ) {
+                    (Some(landed), Some(theirs)) => payloads_commute(landed, &theirs),
+                    // A payload that cannot be read, or cannot legally reach the
+                    // junction, blocks: refusing to cross is the conservative
+                    // answer when the reordering cannot be shown to be harmless.
                     _ => false,
                 };
                 match commutes {
--- a/tests/it/cis_allele_confluence_proptest.rs
+++ b/tests/it/cis_allele_confluence_proptest.rs
@@ -60,11 +60,11 @@
 //!   nor #1262 is about overlapping members, so the citation did not describe
 //!   the shape it was excluding. What that shape hit was #1301, now fixed.
 //! - `an_indel_haplotype_normalizes_to_its_own_sequence` is `#[ignore]`d
-//!   because it finds #1312, equal-payload insertions at adjacent gaps merging
-//!   out of phase with the base between them. It found #1286, #1287, #1290,
-//!   #1292, #1296, #1301, #1304, #1297 and #1308 first, all now fixed, each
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai pause

PR #1319, opened 2026-08-01
fix(normalize): combine two repeats that grew one tract

### LINKED ISSUE #1316: Normalizer: adjacent-gap insertions both re-spell as the same tract-wide repeat, and the allele loses its deletion

nh13:
Closes #1316.

## The defect

A member that grows a tandem tract is re-spelled as a copy count over the whole tract — per member, with no sibling context. Two members that grew the *same* tract therefore both name it, and the pair claims it twice:

```text
reference  ("ACGT" x 64) + "CAGCCAGTCAGCGCATCAG" + ("ACGT" x 64)
           the `A` at 262 is a one-base tract, between `C` at 261 and `G` at 263

g.[261_262insAA;262_263insAA]  ->  g.[262A[3];262A[3]]
```

Both members overlap on that one base, so the allele has no resulting sequence at all and the SPDI apply oracle declines it.

Note the output has moved since the issue was filed: it recorded `g.[259A[3];259A[3]]` with the seed's `269del` gone entirely. #1313 and #1317 fixed both of those before this change; what remained is the coincident pair above. The seed's full form now normalizes to `g.[262A[5];269del]`, and the deletion is pinned by a test.

## Why nothing caught it

Two passes could have, and neither applies:

- **`coalesce_members_at_one_junction`** merges members that settle on one junction. A repeat has no junction — `junction_of` returns `None` for `NaEdit::Repeat` — so the pair was never considered.
- **`demote_repeats_spanning_siblings`** exists to undo exactly this sibling-unaware re-spelling. It re-spells a repeat back into a deletion or a duplication *over the tract*, and neither can express two bases added to a one-base tract, so it declined at its `start < a.start` guard. Instrumented and confirmed: the pass reaches both members, computes `source=Added length=2 start=261 a.start=262`, and bails there.

The one-base payload escaped only by the spelling it happened to get. `g.[261_262insA;262_263insA]` normalizes each member to `g.262dup`, and a duplication *does* carry a junction, so that pair coalesced and came out as one `g.262A[3]`. Same shape, different spelling, different outcome — pinned as a test so the two routes stay in agreement.

## The fix

A repeat's growth *is* expressible, as an insertion at the tract's 3' junction — which is where the tract's own shift rule puts it anyway. `demote_coincident_tract_repeats` emits that for a group of two or more repeats sharing one tract, restoring the junction and letting the existing merge finish the job: `g.262_263insAA` twice, coalesced to `insAAAA`, re-spelled `g.262A[5]` on the next pass.

Scoping:

- **Group must share the span *and* the unit**, so the payloads are powers of one word and therefore commute — the coalesce below cannot then decline on order-ambiguity after this pass has already rewritten them.
- **Only exact counts over whole units that actually grew.** A genotype count (`A[6][1]`), an inexact count, a VEP trailing sequence, a partial unit, or a tract that lost copies all decline; fabricating a payload for them would state bases nothing asserted.
- **All-or-nothing per group.** `respell_at_gap` reverts a member it cannot place, and a half-rewritten group is one insertion overlapping one repeat — the same defect in a spelling no later pass looks for.
- **A lone repeat is untouched**, since it is the correct spelling with no sibling on its tract.

### Why it is a separate pass, and why it runs late

It is deliberately *not* folded into `demote_repeats_spanning_siblings`, despite the kinship. That pass's demotions preserve the barrier a repeat presents to a sibling's shift: a deletion and a duplication both still block one. An insertion claims no bases and blocks nothing.

This was measured, not assumed. Extending `demote_repeats_spanning_siblings` to emit the insertion fixed #1316 and broke two of #1296's own tests — `g.[272_273del;274_275insAA]` became `g.273C>A` instead of `g.[272_273del;274A[3]]`. That output is sequence-preserving (the preservation check passed; only the string assertion fired), and arguably more reduced, but it is reached by dissolving the barrier #1296 installed rather than by agreeing with it. Running the rewrite after the clamps instead leaves #1296 untouched: by then the deletion has been pulled back clear of the tract, so the repeat no longer spans a sibling and is not rewritten.

Two repeats on one tract have no such barrier to preserve — they occupy the same bases — so the rewrite is safe there, and only there.

## The property does NOT now pass — and that is the point

`an_indel_haplotype_normalizes_to_its_own_sequence` has found twelve defects, each masked behind the one before it:

#1286 -> #1287 -> #1290 -> #1292 -> #1296 -> #1301 -> #1304 -> #1297 -> #1308 -> #1312 -> **#1316** -> **#1320**

Eleven are fixed. With #1316 fixed the property passed a **200,000-case soak** — the first it had ever survived — and then **failed a 1,000,000-case run at 27,633 successes** on a twelfth, now filed as **#1320**:

```text
core "AACAGTAAAATAT", insert "AC" after 6, "AA" after 8, "AA" after 9
  g.[263_264insAC;265_266insAA;266_267insAA] -> g.[263_266dup;264_265insCA]
    the dup spans 263-266; the insertion adds at interbase 264, inside it
```

The two `insAA` members merge into the tract-wide repeat correctly — that pair alone gives `g.263_266A[8]`, which is this PR's fix working. `demote_repeats_spanning_siblings` then re-spells that repeat as a duplication over the same four bases, and the duplication swallows the third member's junction exactly as the repeat did. A hole in #1287's fix, one spelling over.

**#1320 is pre-existing.** Verified by A/B: reverting `src/normalize/merge.rs` and `src/normalize/mod.rs` to this commit's parent and re-running the same input gives the identical wrong output. This PR neither causes nor worsens it.

Its seed is committed and tagged here, so it now reproduces **deterministically at the 512-case budget** — taking the `#[ignore]` off would turn CI red immediately. The attribute therefore stays, pointed at #1320.

I had earlier read the clean 200k soak as evidence the chain had ended, and said so. It was not. That caution is now recorded in the test's doc comment, because it nearly cost the `#[ignore]` gate.

## Verification

- `cargo nextest run --features dev` — **8941 passed**, 0 failed (8936 baseline + 5 new tests)
- Same suite under `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1` — 8941 passed
- **Conformance axis run** against the blessed prepared reference (`FERRO_MANIFEST` set): all axis tests pass — `mutalyzer_normalize_tests` (`axis_normalized`, `axis_genomic`, `axis_noncoding`, `axis_errors`, `axis_infos`, `axis_protein_description`, `axis_rna_description`, `axis_coding_protein_descriptions`, and both idempotency axes), `biocommons_normalize_tests`, `hgvs_rs_projection_tests`. `comparator_provenance_reference_identity_matches_live` passes, so the reference is the one the baselines were blessed against.
- The 276,480-case sweep `cis_junction_crossing_shift::no_two_member_allele_normalizes_to_a_different_sequence` holds its residual at **0** (asserted in-test against `KNOWN_FIVE_PRIME_DUP_DEL_RESIDUAL`)
- Confluence property: committed seeds replay green; 60k and 200k soaks both pass
- Failure output was checked for `SIGABRT`/`SIGSEGV`/`TIMEOUT`/`LEAK`, not just `FAIL`, and reconciled against nextest's own failed count
- Commit is signed

Pre-existence: the defect was reproduced on the parent (`c2696b1`) before any source change, and the two new failing tests were confirmed red there and green after.

Conformance was initially unavailable (both reference volumes unmounted); the backup volume was mounted later in the session and the axis run above is the result. It is the first conformance coverage this stack has had.

## Stack

Branches off `fix/issue-1312-out-of-phase-merge` (#1317), so it is link 13 of the chain and 3 commits behind `main` like the rest. It needs to be included in the pending stack rebase.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2,7 +2,7 @@
 //!
 //! See `docs/superpowers/specs/2026-04-30-merge-consecutive-allele-edits-design.md`.
 
-use crate::hgvs::edit::{Base, InsertedSequence, NaEdit, Sequence};
+use crate::hgvs::edit::{Base, InsertedSequence, NaEdit, RepeatCount, Sequence};
 use crate::hgvs::interval::{Interval, UncertainBoundary};
 use crate::hgvs::location::{CdsPos, GenomePos, RnaPos, TxPos};
 use crate::hgvs::uncertainty::Mu;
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

Issue #1321, opened 2026-08-01
Normalizer: a cancelled identity member survives inside a duplication's span


nh13:
## What's wrong

An identity (`=`) member survives inside a duplication's span, so the two members overlap.

```
reference  ("ACGT" x 64) + "TCCCAGAAAAT" + ("ACGT" x 64)

in   NC_TEST.1:g.[261_262insGA;262_263insA;263del]
out  NC_TEST.1:g.[262_263dup;263=]
```

`262_263dup` spans 262–263 and `263=` names position 263, inside it. The property reports `member at interbase 262 overlaps or does not follow the previous member ending at 263`.

## Mechanism — suspected, not yet traced

This looks like #1297 one spelling over. `drop_identity_members_covered_by_siblings` exists precisely to remove a member that cancelled to `=` while a sibling grew over the bases it names — but it drops such a member only when a sibling **claims the bases** under it, and `claims_reference_bases` deliberately excludes `Duplication` (a dup adds a copy at a junction; it consumes nothing). So a duplication that grew across the cancelled member's position does not count as claiming it, and the identity member survives.

That narrowness in #1297's fix was deliberate: `g.[1002=;1005del]` is a real description, so identity members cannot be dropped outright — only when a sibling genuinely claims their bases. Widening it to duplications needs the same care, and `blocks_sibling_shift` (which *does* include `Duplication`) is the obvious predicate to reach for.

Not verified by instrumentation yet; the reproduction above is exact and deterministic.

## Pre-existing

Verified by A/B against the #1320 fix's parent source: the same input yields the identical `g.[262_263dup;263=]` without it. Neither #1316's nor #1320's fix causes it.

## Severity

**Loud.** Criterion 2 of #1235 — "no normalized cis allele contains overlapping or out-of-order members" — is directly violated.

## How it was found

By `cis_allele_confluence_proptest::an_indel_haplotype_normalizes_to_its_own_sequence` at 1,000,000 cases, failing at **114,694 successes**, after #1320 was fixed. It is the thirteenth defect that property has surfaced, each masked behind the one before it: #1286 → #1287 → #1290 → #1292 → #1296 → #1301 → #1304 → #1297 → #1308 → #1312 → #1316 → #1320 → this.

[...]

PR #1322, opened 2026-08-01
fix(normalize): re-spell a duplication that spans a sibling's junction

### LINKED ISSUE #1320: Normalizer: a demoted repeat becomes a duplication that still spans a sibling's junction
## What's wrong

A duplication produced by demoting a tract-wide repeat spans a sibling's junction, so the two members overlap.

```
reference  ("ACGT" x 64) + "AACAGTAAAATAT" + ("ACGT" x 64)
           the `A` tract at 263_266 is four bases, between `T` at 262 and `T` at 267

in   NC_TEST.1:g.[263_264insAC;265_266insAA;266_267insAA]
out  NC_TEST.1:g.[263_266dup;264_265insCA]
```

`263_266dup` spans 263–266 and the insertion adds its sequence at interbase 264, strictly inside that span. The property reports it as `member at interbase 264 overlaps or does not follow the previous member ending at 266`.

## Mechanism — traced

The two `insAA` members both grow the four-base `A` tract, and merge into one tract-wide repeat (`g.263_266A[8]` on its own — that part is #1316's fix working; the pair alone normalizes correctly).

`demote_repeats_spanning_siblings` then fires, correctly: the repeat's tract does swallow the third member's junction at 264. Its demotion re-spells the repeat as a duplication over the tract's 3'-most `length` bases — `length` is 4, so `start = 266 - 4 + 1 = 263`, giving `g.263_266dup`.

**That duplication spans exactly the same bases the repeat did**, so it swallows the sibling's junction just the same. The demotion is supposed to hand the member to the clamps in a form they can pull back, but a duplication claims no reference bases — it is `blocks_sibling_shift` without being `claims_reference_bases` — so `clamp_sibling_crossing_shifts` leaves it alone, and `clamp_sibling_crossing_junctions` governs the junction-occupying sibling rather than the dup's span. Nothing narrows it, and the overlap survives to the output.

So this is a hole in #1287's fix, one spelling over: #1287 stopped a *repeat's* tract swallowing a sibling's junction, and the demotion it relies on can emit a *duplication* that does the same thing.

## Pre-existing

Verified by A/B. Reverting `src/normalize/merge.rs` and `src/normalize/mod.rs` to the parent of the #1316 fix and re-running the same input yields the identical wrong output, `g.[263_266dup;264_265insCA]`. #1316's fix neither causes nor worsens it.

## Severity

**Loud.** Criterion 2 of #1235 — "no normalized cis allele contains overlapping or out-of-order members" — is directly violated, and the property's ordering check catches it rather than it slipping past.

## How it was found

By `cis_allele_confluence_proptest::an_indel_haplotype_normalizes_to_its_own_sequence` at 1,000,000 cases, failing at 27,633 successes, after #1316 was fixed. It is the twelfth defect that property has surfaced, each masked behind the one before it: #1286 → #1287 → #1290 → #1292 → #1296 → #1301 → #1304 → #1297 → #1308 → #1312 → #1316 → this.

Note on rarity: a 200,000-case soak of the same code passed immediately before this run, so the two measurements disagree about how reachable the shape is and no frequency should be read off either. What is certain is that with the seed below committed, it reproduces **deterministically at the property's committed 512-case budget** — so it would turn CI red the moment `#[ignore]` comes off.

Seed (committed):

```
cc 2521c7200c3f735e03b752af83fb12c21c0e903ee3c7a42454dbe8f68f62b08d
core "AACAGTAAAATAT", insert "AC" after 6, insert "AA" after 8, insert "AA" after 9
```

## Relationship to #1235

nh13:
Closes #1320.

## The defect

```text
reference  ("ACGT" x 64) + "AACAGTAAAATAT" + ("ACGT" x 64)
           the `A` tract at 263_266 is four bases, between `T` at 262 and `T` at 267

g.[263_264insAC;265_266insAA;266_267insAA]  ->  g.[263_266dup;264_265insCA]
```

The duplication spans 263–266 and the sibling adds its bases at interbase 264, inside it. The property reports `member at interbase 264 overlaps or does not follow the previous member ending at 266`.

Worth being precise about what is wrong here, because it is *not* what the issue title suggests. The output is **sequence-preserving** — the SPDI apply oracle accepts it, and both members are zero-width insertion points, so nothing literally overlaps. The violation is **ordering**.

## Root cause

A duplication **sorts by its span's start** and **acts at its span's end** — it copies the bases under the span and lands them at the junction 3' of them. Those two positions coincide for a one-base dup and diverge as the span widens.

A sibling junction strictly between them falls on the wrong side of both: `263_266dup` renders first, because 263 precedes 264, while the bases it adds arrive at 266, *after* the sibling's. The pair reads as out of order however it is spelled.

How it got there: the two `insAA` members merge into the tract-wide repeat correctly — that pair alone gives `g.263_266A[8]`. `demote_repeats_spanning_siblings` then fires, also correctly, because the tract does swallow the third member's junction; its demotion re-spells the repeat as a duplication over the tract's 3'-most `length` bases, and with `length == 4` that is the same four bases. So the demotion hands back a member that swallows the junction exactly as the repeat did.

Nothing narrows it afterwards. `clamp_sibling_crossing_shifts` bounds a member against a sibling's *bases* and an insertion has none; `clamp_sibling_crossing_junctions` governs the insertion's own junction rather than the duplication's span.

## The fix

`respell_colliding_duplications` already exists for exactly this: a duplication whose span collides becomes the equivalent insertion, which claims nothing and sorts where it acts. It only tested siblings that **claim bases**, so an insertion's junction inside the span did not count as a collision. This adds that junction half — the same half #1287 added to the *repeat* path.

Re-spelled, `g.263_266dup` becomes `g.266_267insAAAA`, and the pair is disjoint and in order:

```text
g.[263_264insAC;265_266insAA;266_267insAA]  ->  g.[264_265insCA;266_267insAAAA]
```

### Both interval bounds are load-bearing

The test is `dup.start < junction < dup.end`, strict at **both** ends, and each end has its own reason:

- **`< dup.end`** — a junction at the 3' edge is the duplication's own landing junction, the shared-junction case `coalesce_members_at_one_junction` merges (#1286). Re-spelling there would pull it out from under that merge.
- **`> dup.start`** — I first wrote `>=` here, copying #1287, and it **broke #1261 and #1301**. At the 5' edge the duplication sorts to the same position the sibling adds at, so `junction_rank` (#1301) already orders the pair and there is no discrepancy to repair. `g.[264_265insCA;264_265dup]` and `g.[258dup;258_259dup]` are both well-ordered and must survive untouched. Both are now pinned as boundary tests here, on top of their own modules' tests.

This makes the predicate deliberately **narrower** than the equivalent test in `demote_repeats_spanning_siblings`, which takes `>=` at the 5' end. A repeat has no comparable sort/act split — it spans its tract and is rendered there — so an interior junction is a problem for it wherever it sits. The doc comment records that asymmetry so the next reader does not "fix" it into agreement.

## Tests

`tests/it/issue_1320_dup_spans_sibling_junction.rs`, five cases. The helper asserts the **invariant**, not just the rendered string: it re-applies through `hgvs_to_spdi` for sequence preservation *and* checks members are disjoint and ascending, mirroring the property's own check. Asserting the string alone would have passed for any other out-of-order arrangement — and would not have failed at all before the fix, since the bad output is sequence-preserving.

- the seed's own case
- the two-member form, which must keep the tract-wide repeat (`g.263_266A[8]`)
- the 5' boundary, #1301's reproduction
- the 5' boundary between two duplications, #1261's reproduction
- the pre-existing bases-collision half, so widening the predicate does not lose it

## Verification

- `cargo nextest run --features dev` — 8963 passed
- Same under `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1` — 8963 passed
- All twelve committed confluence seeds replay green, including #1320's `2521c720`; no new seed appended
- Failure output checked for `SIGABRT`/`SIGSEGV`/`TIMEOUT`/`LEAK`, not just `FAIL`, and reconciled against nextest's own failed count
- Commit signed

The branch is **fully green** — 8964 passed, 0 failed, on both the plain and the oracle run.

`the_ignored_indel_property_still_finds_its_defect` (added in #1317) pins the property's *live* blocker and asserts it still fails, so fixing one turns that test red by design. Its pin moves from #1320 to **#1321** here.

With this fix the property fails a 1,000,000-case soak at **114,694 successes** on #1321: a cancelled identity member survives inside a duplication's span, because `drop_identity_members_covered_by_siblings` drops one only when a sibling *claims the bases* under it and a duplication claims none — #1297's shape one spelling over. Verified pre-existing; its seed is committed and tagged here, so it reproduces deterministically at the 512-case budget.

For calibration, since soak depth is the only real measure of how close the chain is to ending: #1320 surfaced at 27,633 successes, #1321 at 114,694 on the same budget — roughly four times deeper. That is progress, not completion, and a clean soak at one depth is not evidence about the next. A 200,000-case run passed immediately before the one that found #1320.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -4042,15 +4042,53 @@ pub(crate) fn respell_colliding_duplications<P: ReferenceProvider>(
         ) {
             continue;
         }
-        let collides = (0..spans.len()).filter(|&j| j != i).any(|j| {
-            spans[j].as_ref().is_some_and(|s| {
-                s.claims_bases
-                    && s.region == dup.region
-                    && s.accession == dup.accession
-                    && s.start <= dup.end
-                    && s.end >= dup.start
-            })
-        });
+        let same_molecule =
+            |s: &&MemberSpan| s.region == dup.region && s.accession == dup.accession;
+        let siblings = || {
+            (0..spans.len())
+                .filter(|&j| j != i)
+                .filter_map(|j| spans[j].as_ref())
+                .filter(same_molecule)
+        };
+        // A sibling that claims bases collides when its span meets the
+        // duplication's — the two then name the same positions.
+        let claims_the_bases = siblings()
+            .filter(|s| s.claims_bases)
+            .any(|s| s.start <= dup.end && s.end >= dup.start);
+        // A sibling that claims none still collides when its **junction** falls
+        // strictly between the duplication's two ends (#1320).
+        //
+        // A duplication *sorts* by its span's start and *acts* at its span's
+        // end — it copies the bases under the span and lands them at the
+        // junction 3' of them. Those two positions are the same for a one-base
+        // dup and diverge as the span widens, and a sibling junction strictly
+        // between them falls on the wrong side of both: the duplication is
+        // rendered first, because 263 precedes 264, while the bases it adds
+        // arrive at 266, after the sibling's. The pair then reads as out of
+        // order however it is spelled. Re-spelling settles it, because the
+        // insertion form sorts where it acts.
+        //
+        // **Both** ends are exclusive, and each for its own reason:
+        //
+        // - at `dup.end` the junction is the duplication's own landing junction,
+        //   the shared-junction case `coalesce_members_at_one_junction` merges
+        //   (#1286) — re-spelling would pull it out from under that merge;
+        // - at `dup.start` the duplication sorts to the same position as the
+        //   sibling, so `junction_rank` already orders the pair correctly
+        //   (#1301) and there is no discrepancy to repair. `g.[264_265insCA;
+        //   264_265dup]` and `g.[258dup;258_259dup]` are both well-ordered and
+        //   must survive untouched.
+        //
+        // This is narrower than the equivalent test in
+        // `demote_repeats_spanning_siblings` (#1287), which takes `>=` at the 5'
+        // end. A repeat has no comparable sort/act split — it spans its tract
+        // and is rendered there — so an interior junction is a problem for it
+        // wherever it sits.
+        let swallows_a_junction = siblings()
+            .filter(|s| !s.claims_bases)
+            .filter_map(|s| s.junction)
+            .any(|junction| junction > dup.start && junction < dup.end);
+        let collides = claims_the_bases || swallows_a_junction;
         if !collides {
             continue;
         }
--- a/tests/it/cis_allele_confluence_proptest.rs
+++ b/tests/it/cis_allele_confluence_proptest.rs
@@ -60,9 +60,9 @@
 //!   nor #1262 is about overlapping members, so the citation did not describe
 //!   the shape it was excluding. What that shape hit was #1301, now fixed.
 //! - `an_indel_haplotype_normalizes_to_its_own_sequence` is `#[ignore]`d
-//!   because it finds #1320, a demoted repeat becoming a duplication that still
-//!   spans a sibling's junction. It found #1286, #1287, #1290, #1292, #1296,
-//!   #1301, #1304, #1297, #1308, #1312 and #1316 first, all now fixed, each
+//!   because it finds #1321, a cancelled identity member surviving inside a
+//!   duplication's span. It found #1286, #1287, #1290, #1292, #1296, #1301,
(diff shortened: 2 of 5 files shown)


nh13:
@coderabbitai pause

Issue #1323, opened 2026-08-01
Normalizer: three-member allele loses its deletion and denotes the wrong sequence as two overlapping duplications


nh13:
## What's wrong

A three-member allele normalizes to two overlapping duplications, **loses its deletion, and denotes the wrong sequence**.

```
reference  ("ACGT" x 64) + "CAGGGATCAT" + ("ACGT" x 64)

in   NC_TEST.1:g.[260del;261_262insGA;262_263insGA]
out  NC_TEST.1:g.[261_262dup;262dup]

intended  ...CAGGG AAGAT CAT...
emitted   ...CAGGG AGAAT CAT...
```

Two faults in one output:

1. **The sequence is wrong.** The emitted description denotes `CAGGGAGAATCAT` where the input denotes `CAGGGAAGATCAT` — the same bases in a different order.
2. **The `260del` is gone**, and the two remaining members overlap: `261_262dup` spans 261–262 and `262dup` claims 262.

This is a more severe class than the several defects before it. #1301, #1316, #1320 and #1321 were all ordering or overlap faults on outputs that still denoted the right bases; this one changes what the allele means.

The overlapping-dup shape is the #1286 family (`g.[263dup;263dup]`, two dup members claiming one position), which suggests the pair reached one span through a path the #1286/#1320 repairs do not cover — but the lost deletion may be a separate fault, and neither is traced yet.

## Mechanism — not yet traced

Filed on an exact, deterministic reproduction rather than a diagnosis. The three input members are already the 3'-shifted form; what the merge does to them, and at which pass the deletion is dropped, still needs instrumenting.

## Pre-existing

Verified by A/B: with `src/normalize/merge.rs` reverted to the parent of #1321's fix, the same input yields the identical `g.[261_262dup;262dup]`. None of #1316, #1320 or #1321's fixes causes it.

## Severity

**Loud, and sequence-changing.** It violates #1235's second criterion (overlapping members) *and* denotes different bases, so it is not merely a rendering fault. The property's own oracle — re-applying through `hgvs_to_spdi`, a separate code path from the canonicalizer under test — is what caught it.

## How it was found

[...]

PR #1324, opened 2026-08-01
fix(normalize): let a duplication cover the identity member inside it

### LINKED ISSUE #1321: Normalizer: a cancelled identity member survives inside a duplication's span

nh13:
Closes #1321.

## The defect

```text
reference  ("ACGT" x 64) + "TCCCAGAAAAT" + ("ACGT" x 64)

g.[261_262insGA;262_263insA;263del]  ->  g.[262_263dup;263=]
```

`262_263dup` spans 262–263 and `263=` names position 263, inside it. The pair overlaps, which #1235's second criterion forbids.

`g.262_263dup` **on its own already denotes the input** — it copies `G` at 262 and `A` at 263, giving `CAGA` + `GA` + `AAAT`, exactly what the three input members describe between them. So the `263=` is a pure residue: it adds no information and creates the overlap.

## Root cause

`drop_identity_members_covered_by_siblings` exists for precisely this shape (#1297): a `del` and an `ins` merge into a `delins` restating the reference, cancel to `=`, and the member that absorbed the change grows over it.

It gated on `claims_reference_bases`, which **deliberately excludes `Duplication`** — a dup adds a copy at a junction and consumes nothing. So a duplication that grew across the cancelled member's position did not count as covering it, and the residue survived. #1297's shape, one spelling over.

## The fix

The gate widens from `claims_reference_bases` to **`blocks_sibling_shift`**.

Those two predicates differ by exactly the case that matters: `blocks_sibling_shift` is `claims_reference_bases` plus `Duplication`/`DupIns`. Those are barriers precisely because they read their payload from the reference *under their own span* — and naming those positions is what makes an identity member inside them redundant, whether or not the bases are consumed.

**Insertions stay excluded under either predicate, and must.** An insertion's span *is* the gap it occupies, so `g.264_265insA` nominally covers position 264 while changing nothing there — dropping an identity member on that basis would be wrong. Pinned as a test.

The other guard #1297 installed is also pinned: an identity member nothing covers is real information, so `g.[1002=;1005del]` keeps both. Dropping identities outright would lose what the `=` records.

## Tests

`tests/it/issue_1321_identity_inside_a_duplication.rs`, three cases — the seed's own reproduction, the uncovered-identity guard, and the insertion-sibling guard. The helper asserts the invariant (members disjoint and ascending, mirroring the property's own check) as well as sequence preservation through `hgvs_to_spdi`.

One helper detail worth flagging for the next module in this family: its applier sorts 3'-to-5' and, **at a tied position, takes the longer deletion first**. The pinned input has a deletion and an insertion that both touch interbase 262 (`262_263insA` and `263del`); applying the zero-width one first leaves `claimed` at 262 and makes the deletion look like it overruns, so the applier reported a legal input as having no resulting sequence. That cost a false failure here and the same shape will recur.

## Verification

- `cargo nextest run --features dev` — **8967 passed, 0 failed**
- Same under `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1` — 8967 passed
- All fourteen committed confluence seeds replay green, including #1321's `467e0a71`
- Failure output checked for `SIGABRT`/`SIGSEGV`/`TIMEOUT`/`LEAK`, not just `FAIL`, and reconciled against nextest's own failed count
- Commit signed

`the_ignored_indel_property_still_finds_its_defect` pins the property's *live* blocker and asserts it still fails, so fixing one turns that test red by design. Its pin moves from #1321 to **#1323** here.

With this fix the property fails a 1,000,000-case soak at **143,034 successes** on #1323:

```text
g.[260del;261_262insGA;262_263insGA]  ->  g.[261_262dup;262dup]
  intended  C A G G G A A G A T C A T
  emitted   C A G G G A G A A T C A T
```

The deletion is dropped, the duplications overlap at 262, and the result denotes different bases. That makes #1323 the **first chain defect since #1304 that is sequence-changing** rather than a rendering or ordering fault — a more severe class than #1301, #1316, #1320 or #1321. Verified pre-existing.

On depth, since it is the only measure of how close the chain is to ending: #1320 surfaced at 27,633 successes, #1321 at 114,694, #1323 at 143,034. **The 4x jump did not repeat** — the last step is 1.25x — so depth is not trending predictably and should not be used to forecast an end.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -4396,9 +4396,28 @@ fn translate_junction_member<P: ReferenceProvider>(
 ///   the repeat alone denotes the input; the `265=` residue overlaps it
 /// ```
 ///
-/// Only the identity member is dropped, and only when a sibling *claims* the
-/// bases under it — a sibling that merely adds sequence at a junction inside it
-/// contradicts nothing.
+/// Only the identity member is dropped, and only when a sibling's span *covers*
+/// it under [`blocks_sibling_shift`] — a sibling that merely adds sequence at a
+/// junction inside it contradicts nothing.
+///
+/// `blocks_sibling_shift` rather than `claims_reference_bases`, because a
+/// duplication covers the positions it copies without consuming them (#1321):
+///
+/// ```text
+/// g.[261_262insGA;262_263insA;263del]  ->  g.[262_263dup;263=]
+///   `262_263dup` alone denotes the input; the `263=` overlaps it
+/// ```
+///
+/// That is #1297's shape one spelling over, and the two predicates differ by
+/// exactly the case that matters — `blocks_sibling_shift` is
+/// `claims_reference_bases` plus `Duplication`/`DupIns`, which are barriers
+/// precisely because they read their payload from the reference under their own
+/// span. Naming those positions is what makes an identity member inside them
+/// redundant.
+///
+/// Insertions stay out of it under either predicate, and must: an insertion's
+/// span *is* the gap it occupies, so `g.264_265insA` nominally covers position
+/// 264 while changing nothing there.
 pub(crate) fn drop_identity_members_covered_by_siblings(
     members: &mut Vec<HgvsVariant>,
     phase: AllelePhase,
@@ -4425,7 +4444,7 @@ pub(crate) fn drop_identity_members_covered_by_siblings(
             };
             (0..members.len()).filter(|&j| j != i).any(|j| {
                 spans[j].as_ref().is_some_and(|s| {
-                    s.claims_bases
+                    s.blocks_shift
                         && s.region == span.region
                         && s.accession == span.accession
                         && s.start <= span.start
--- a/tests/it/cis_allele_confluence_proptest.rs
+++ b/tests/it/cis_allele_confluence_proptest.rs
@@ -60,11 +60,11 @@
 //!   nor #1262 is about overlapping members, so the citation did not describe
 //!   the shape it was excluding. What that shape hit was #1301, now fixed.
 //! - `an_indel_haplotype_normalizes_to_its_own_sequence` is `#[ignore]`d
-//!   because it finds #1321, a cancelled identity member surviving inside a
-//!   duplication's span. It found #1286, #1287, #1290, #1292, #1296, #1301,
-//!   #1304, #1297, #1308, #1312, #1316 and #1320 first, all now fixed, each
-//!   masked behind the one before it; see its doc comment for the history and
-//!   the live reproduction.
+//!   because it finds #1323, a three-member allele losing its deletion and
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai pause

Issue #1325, opened 2026-08-01
Normalizer: a repeat whose growth exceeds its tract cannot be demoted, so it swallows a sibling's junction


nh13:
## What's wrong

A repeat's tract swallows a sibling's junction, and the demotion that exists to prevent that declines.

```
reference  ("ACGT" x 64) + "GATCATAAATTCAGC" + ("ACGT" x 64)
           the `A` tract at 263_265 is three bases, between `T` at 262 and `T` at 266

in   NC_TEST.1:g.[262_263insAA;263_264insAA;264_265insC]
out  NC_TEST.1:g.[263_265A[7];264_265insC]
```

The insertion adds its base at interbase 264, strictly inside the tract the repeat names. The property reports `members overlap — SPDI at 264 starts before 265`.

## Mechanism

The first two members are handled correctly on their own: `g.[262_263insAA;263_264insAA]` normalizes to `g.263_265A[7]`, which is `demote_coincident_tract_repeats` (#1316) combining two members that each grew the same tract.

Adding the third member puts a junction at 264, inside 263–265. `demote_repeats_spanning_siblings` is the pass for exactly that — its `spans_sibling_junction` test fires (`264 >= 263 && 264 < 265`) — but it then declines at its `start < a.start` guard:

- the repeat grew the tract by 4 bases (two `insAA`), while the tract is 3 bases long
- so the equivalent duplication would sit at `start = a.end - length + 1 = 265 - 4 + 1 = 262`, which is 5' of the tract's own start

There is no duplication *over the tract* that can express four bases added to a three-base tract, so the pass bails and leaves the repeat spanning the sibling.

**This is the same hole as #1316, on its other branch.** #1316 was the case where two repeats coincided on one tract; this is the case where one repeat swallows a sibling's junction. Both arise because an added-repeat whose growth exceeds its tract has no duplication form, and both were left unrepaired by `demote_repeats_spanning_siblings` for that reason.

[...]

PR #1326, opened 2026-08-01
fix(normalize): order a shared junction's payloads by where they came from

### LINKED ISSUE #1323: Normalizer: three-member allele loses its deletion and denotes the wrong sequence as two overlapping duplications

nh13:
Closes #1323.

## The defect

```text
reference  ("ACGT" x 64) + "CAGGGATCAT" + ("ACGT" x 64)
           G at 259-261, A at 262, T at 263

g.[260del;261_262insGA;262_263insGA]  ->  g.[261_262dup;262dup]

  intended  C A G G G A A G A T C A T
  emitted   C A G G G A G A A T C A T
```

Both members occupy **junction 262** — `262dup` carrying `A`, `261_262dup` carrying `GA` — and the renderer emits them in span order, concatenating `GA ++ A` where the input means `A ++ GA`.

**Correction to the issue as filed.** It says the allele "loses its deletion". It does not: `261del` and `261_262insGA` collapse into `261delinsGA`, which canonicalizes to `262dup`, and `262dup` denotes exactly what that del/ins pair denotes together. The deletion is absorbed by a correct collapse. The fault is the shared-junction ordering alone — the sequence really is wrong, so the severity stands, but the mechanism in the issue title was mine and it was wrong. The issue carries a correcting comment.

## Root cause

Traced with per-pass instrumentation of the merge loop:

```text
pass 0  normalized   [261del; 261_262insGA; 261_262dup]
        respell_dups [261del; 261_262insGA; 262_263insGA]   <- correctly pulled back

pass 1  merged_split [262dup; 262_263insGA]
        pre_split    [261delinsGA; 262_263insGA]
        normalized   [262dup; 261_262dup]                    <- both on junction 262
        (no sibling pass changes anything)
```

Two facts combine, and neither is individually wrong:

- **Neither member moved to reach junction 262.** One was created there by the collapse; the other re-spelled into it in place. `clamp_sibling_crossing_junctions` bounds a member that *crosses* a sibling, so there was no crossing to bound.
- **`coalesce_members_at_one_junction` declined.** Its order-independence guard requires the payloads to commute, and `A ++ GA != GA ++ A`.

So the pair was left sharing a junction with no defined order, and the rendered order — which follows their spans — is an artefact rather than a statement about the allele.

## Separating them is not available

The obvious repair is to pull one back so they no longer share a junction. It changes the sequence. Moving `261_262dup` to junction 261 is a *legal* rotation (its last base `A` equals `ref[262]`), but the resulting `[261_262insAG;262dup]` denotes `GAGAAT`, not `GAAGAT`. Both members genuinely belong at 262. Only their order is wrong.

## The fix

The pass now takes the `pre_split` snapshot the sibling passes already build, and orders the concatenation by **where each member came from**: `261delinsGA` sat at 261 and `262_263insGA` at 262, so the 5'-most origin contributes first — `A ++ GA = AGA`, giving `g.262_263insAGA`, which denotes the input.

This **subsumes** the commuting guard rather than relaxing it. `p ++ q == q ++ p` is precisely the condition under which the two orderings agree, so a commuting group reaches the same concatenation either way. The guard survives as the fallback for a group whose origins are unavailable — an empty or length-mismatched snapshot disables the ordering rather than mis-attributing it.

That is why this did not break the passes that depend on the guard, which was my main worry going in: #1286, #1308, #1312 and #1313 rely on it for the *commuting* case, and that is exactly where the new rule agrees with the old one.

## Tests

`tests/it/issue_1323_shared_junction_order.rs`, three cases, asserting the invariant (members disjoint and ascending) as well as sequence preservation through `hgvs_to_spdi`:

- the seed's own case, now `g.262_263insAGA`
- **#1286's commuting merge still gives `g.257_263A[9]`** — ordering by origin must not disturb it
- **#1312's non-commuting pair is still kept apart** on distinct junctions by the clamp. Merging is the repair for a pair that *cannot* be separated, not a licence to fold together members the clamp already ordered

## Verification

- `cargo nextest run --features dev` — **8970 passed, 0 failed**
- Same under `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1` — 8970 passed
- All sixteen committed confluence seeds replay green, including #1323's `ff4eca54`
- Failure output checked for `SIGABRT`/`SIGSEGV`/`TIMEOUT`/`LEAK`, not just `FAIL`
- Commit signed

The pin in `the_ignored_indel_property_still_finds_its_defect` moves from #1323 to **#1325**, found by a 1,000,000-case soak at 132,216 successes:

```text
g.[262_263insAA;263_264insAA;264_265insC]  ->  g.[263_265A[7];264_265insC]
```

A repeat whose growth exceeds its tract has no duplication form, so `demote_repeats_spanning_siblings` declines at its `start < a.start` guard and the tract keeps the sibling's junction. **That is the other branch of the hole #1316 fixed for coincident repeats**, and `repeat_growth_as_insertion` — already written for #1316 — is the tool for it. Verified pre-existing.

On depth: #1320 surfaced at 27,633 successes, #1321 at 114,694, #1323 at 143,034, #1325 at 132,216. The early 4x jump did not continue and the last three sit in one band, so depth has **plateaued** rather than trending toward an end. I would not read it as evidence the chain is nearly exhausted.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -4163,6 +4163,7 @@ pub(crate) fn respell_colliding_duplications<P: ReferenceProvider>(
 /// Genomic axes only, matching `respell_colliding_duplications` — the
 /// transcript axes need a coordinate conversion this does not carry.
 pub(crate) fn coalesce_members_at_one_junction<P: ReferenceProvider>(
+    before: &[HgvsVariant],
     members: &mut Vec<HgvsVariant>,
     phase: AllelePhase,
     uncertain: bool,
@@ -4204,10 +4205,35 @@ pub(crate) fn coalesce_members_at_one_junction<P: ReferenceProvider>(
         }
     }
 
+    // Where each member sat *before* this pass's per-member normalization, used
+    // to order a group whose payloads do not commute. Aligned by index with
+    // `members`: the caller builds both from one list and every pass between
+    // mutates in place, so `before[i]` is member `i`'s origin. An empty or
+    // mismatched snapshot disables the ordering rather than mis-attributing it.
+    let origins: Vec<Option<(i64, i64)>> = if before.len() == members.len() {
+        before
+            .iter()
+            .map(|v| member_span(v, kind).map(|s| (s.start, s.end)))
+            .collect()
+    } else {
+        vec![None; members.len()]
+    };
+
     let mut remove: Vec<usize> = Vec::new();
     for (_, at) in &groups {
         if at.len() < 2 {
             continue;
+        }
+        // Concatenate in the order the members *came from*, not the order they
+        // happen to sit in now. Two members can reach one junction without
+        // either crossing the other — a collapse can create one there while a
+        // sibling re-spells into it in place — and then the rendered order is
+        // an artefact of their spans rather than of what the allele means
+        // (#1323). Their origins are what the input actually said.
+        let mut at: Vec<usize> = at.clone();
+        let ordered_by_origin = at.iter().all(|&i| origins[i].is_some());
+        if ordered_by_origin {
+            at.sort_by_key(|&i| (origins[i], i));
         }
         let payloads: Option<Vec<Vec<Base>>> = at
             .iter()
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

PR #1329, opened 2026-08-02
test(normalize): use the shared cis-allele apply oracle in three suites

nh13:
Three cis-allele regression suites — `issue_1297_cancelled_identity_member`, `issue_1308_commuting_payload_phase` and `issue_1312_landed_payload_commutes` — each carry a private copy of `assert_padded_preserving`. They predate the shared helper in `tests/it/common/synthetic.rs` and are behind it in two ways that matter:

**The applier's sort is missing a tie-break.** All three sort 3'-to-5' by position alone. At a tied position the *longer* deletion has to be applied first; otherwise a legal input where a deletion and an insertion touch the same interbase — `262_263insA` alongside `263del` — leaves the running cursor at that interbase, the deletion looks like it overruns, and the helper reports a well-formed input as having no resulting sequence.

**They only check the sequence, not the rendering.** The copies assert that the output denotes the input's bases. The shared helper additionally asserts that the output's members render disjoint and ascending. That second property is what most of these regressions are actually about: an overlapping pair that happens to apply to the right sequence is still malformed, and the copies pass it.

This points all three at the shared helper and deletes the copies. No test data, inputs or expectations change — the three suites simply gain the stricter assertion and the tie-break fix.

Found while dual-reviewing the #1288→#1326 cis-allele stack: the shared helper was extracted partway through, and suites added later reintroduced local copies. The improvements above were contributed by #1324's copy and upstreamed there; these three were left behind.

Verified: 8993 tests pass, including under `FERRO_ASSERT_IDEMPOTENT=1` with `FERRO_ASSERT_REPARSE=1`. `cargo fmt` and `clippy --all-features --all-targets -- -D warnings` clean.

## Summary by CodeRabbit

* **Tests**
  * Consolidated shared test validation across multiple integration tests.
  * Removed duplicated test helper implementations while preserving existing test behavior.

### DIFF
--- a/tests/it/issue_1297_cancelled_identity_member.rs
+++ b/tests/it/issue_1297_cancelled_identity_member.rs
@@ -20,57 +20,9 @@
 //! identity member a sibling *claims the bases of* is a contradiction, and only
 //! that one is dropped.
 
-use crate::common::synthetic::{padded, SyntheticBuilder};
-use ferro_hgvs::spdi::hgvs_to_spdi;
-use ferro_hgvs::{parse_hgvs, HgvsVariant, Normalizer};
+use crate::common::synthetic::assert_padded_preserving;
 
 const CORE: &str = "GCATGAAAAT";
-
-/// Normalize and assert the output denotes the input's bases. A pair whose
-/// members overlap has no resulting sequence at all, which is how the pre-fix
-/// output failed.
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #1271, opened 2026-07-30
Normalizer: separations_are_meaningful does not cover net deletions, so a length threshold stands in for the coincidence guard


nh13:
## What's wrong

`separations_are_meaningful` is the guard that distinguishes a genuine separation between two derived pieces from an artefact of `best_alignment`'s max-match search. It is restricted to net **insertions**:

```rust
if result.len() > reference.len() && !separations_are_meaningful(&pieces) {
    return whole();
}
```

So the three block regimes are not equally protected:

| regime | guard against a coincidental split |
| --- | --- |
| equal length | none needed — `best_alignment` compares position-wise, so there is no gap to place and no search to go wrong |
| net insertion | `separations_are_meaningful` |
| **net deletion** | **nothing** |

A long net deletion was covered only *incidentally*, by `MAX_SPLIT_BLOCK` being 32 and therefore smaller than the blocks at risk. Raising that bound (#1237, commit 2) removed the cover and split the spec's own `delins.md:44-47` worked example four ways:

```
LRG_199t1:c.850_901delinsTTCCTCGATGCCTG        52 nt -> 14 nt
  spanning  LRG_199:g.646630_646681delinsTTCCTCGATGCCTG
  split     LRG_199:g.[1_39delinsT;41A>C;43_49delinsTCGATGC;51_52delinsTG]
```

Precisely the harm that passage names, on that passage's own example — the corpus shows one correct `p.(Arg53_Arg54delinsSerCysAlaHisTyrLeuAla)` becoming three bogus consequences.

## Current state

**Not a live defect.** #1237 makes the bound regime-aware: `MAX_UNGUARDED_SPLIT_BLOCK` holds net deletions at the original 32, where the constant genuinely *is* the guard, while `MAX_SPLIT_BLOCK` stays at 1024 for the two safe regimes. Pinned by a `partition_block` unit test covering all three regimes plus an end-to-end test on the 52 nt shape, both verified to fail against a length-blind cap.

This issue tracks the **principled** fix: extend `separations_are_meaningful` to net deletions and retire `MAX_UNGUARDED_SPLIT_BLOCK`, so the coincidence question is answered by a rule about the derived pieces rather than by a length threshold that happens to sit below the risky sizes.

## Why it is not a one-liner

The doc comment on `separations_are_meaningful` records two measured constraints that a naive extension violates:

[...]

PR #1330, opened 2026-08-02
test(spec): give the generated spec fixtures a committed oracle

### LINKED ISSUE #1272: Tests: enumeration_replays_recorded_behavior compares against a baseline regenerated from the code under test, so it cannot catch a regression
## What's wrong

`tests/it/spec_enumeration_tests.rs::enumeration_replays_recorded_behavior` compares observed normalization against a baseline in `tests/fixtures/grammar/hgvs_spec_enumeration.json`.

That fixture is **generated, not committed** — it is `.gitignore`d and produced by `generate_spec_enumeration` from the code under test. The `recorded` side and the `observed` side therefore come from the same source, and in CI the fixture is regenerated before the run.

A test whose baseline is regenerated from the implementation it is testing cannot detect a behavioural change. It can only detect drift between a *stale local artifact* and current code — which is a developer-workstation condition, not a regression.

## How this surfaced

While reviewing #1239 (now folded into #1237), a stale local copy of the fixture caused a failure:

```
1 enumeration row(s) drifted. Regenerate if intentional:
  id       : project-g/LRG_199t1:c.850_901delinsTTCCTCGATGCCTG
    recorded : LRG_199:g.646630_646681delinsTTCCTCGATGCCTG
    observed : LRG_199:g.[646630_646636delinsTTCCTCG;646640_646678delinsC;646680_646681delinsTG]
```

That was a real regression — the spec's own `delins.md:44-47` worked example being split four ways (see #1271). It was caught **only because a local artifact happened to predate the change**. On CI, where the fixture is regenerated, the same commit was fully green. The message even instructs "Regenerate if intentional", which is the action that hides it.

So the one time this test fired on a genuine defect, it did so by accident, and the prescribed response would have suppressed it.

## Why not just commit the fixture

That was deliberately undone: committing it made every parser PR a merge-conflict magnet, since each PR regenerated the whole file. Reverting that trade is probably not the answer.

Options worth weighing, roughly in increasing cost:

1. **Drop the test.** It currently asserts nothing about behaviour; deleting it is honest and loses no coverage. The generation step still validates the curated overrides against the spec checkout, which is the real gate.
2. **Commit a narrow baseline.** Not the whole enumeration — a small curated set of rows whose normalization is spec-anchored (the `delins.md:44-47` example being the obvious first entry), so a change to any of them is a reviewable diff.
3. **Compare against the previous release** rather than a regenerated artifact, so the baseline has an independent origin.

Option 2 looks like the best value: it is small, the rows are exactly the ones with documented spec answers, and it fails loudly on the change that matters without reintroducing the conflict problem.

## Scope

The sibling artifact `hgvs_spec_normalization.json` is generated the same way and consumed by five test modules; whether they have the same blindness has not been checked and should be part of this.

Note the contrast with the artifacts that *are* committed — `generate_tool_support_tables` and `extract_spec_enumeration_windows` — where absent genuinely means drift, `--check` is the right gate, and CI uses it.

nh13:
Closes #1272

## The problem, demonstrated

Both spec artifacts — `hgvs_spec_enumeration.json` and `hgvs_spec_normalization.json` — are gitignored and regenerated from the code under test before every CI run (`ci.yml:277,279`, `.pre-commit-config.yaml:40,57`). So the two tests that read them compare ferro against itself. `enumeration_replays_recorded_behavior` and `pinned_v21_normalization_behavior` can only fire on a stale local artifact, and the one time the former caught a real defect it did so by accident, with a failure message telling the developer to regenerate — the action that hides it.

I reproduced the exact regression the issue describes rather than assuming it. Raising `MAX_UNGUARDED_SPLIT_BLOCK` from 32 to 1024 (the #1271 change) splits the spec's own worked example:

```
project-g/LRG_199t1:c.850_901delinsTTCCTCGATGCCTG
  before  LRG_199:g.646630_646681delinsTTCCTCGATGCCTG
  after   LRG_199:g.[646630_646636delinsTTCCTCG;646640_646678delinsC;646680_646681delinsTG]
```

Exactly **one** row of 2184 changed, **no** status moved, and **all 82** spec-fixture tests passed. That is the hole.

## What now catches it

Four committed guards, none of which a regeneration can absorb.

**1. A status for the shape itself.** `projection-splits-single-member` classifies a single-member input that projects to a multi-member allele — the form `DNA/delins.md:42` names as "not correct", and `:47` explains why (it "prevents software tools making incorrect predictions"). Budgeted at 10; the regression above makes it 11 and fails the budget.

This is deliberately *not* a `projections` override. The spec never writes the genomic form of that example, and the overrides file's contract is explicit that nothing there is inferred — so pinning `LRG_199:g.646630_…` as spec-stated would have fabricated a spec claim. The structural rule is what the spec actually supports.

**2. `PASSING_CENSUS` — closing a 2136-row hole.** `divergence_budget_is_unchanged` iterates only `DIVERGENCE_BUDGET`; every status in `KNOWN_PASSING_STATUSES` was counted by nothing. A projection could go from rendering a string to erroring with zero signal, since `every_observed_status_is_accounted_for` only rejects status *names* it has never seen. The two tables now total every row, enforced by `every_status_is_counted_by_exactly_one_table`.

**3. `ferro_produces_the_form_the_spec_states` — a real oracle for the normalization fixture,** which had no census of any kind. This is the gap the issue's Scope section asks about, and it was wider than the enumeration's. `spec_expected` comes from the spec text and the committed overrides, never from ferro, and `classify` already derives a genuine spec-vs-ferro verdict on every row — nothing re-checked it live. Its 36 divergences are pinned by input string in `KNOWN_DIVERGENT_INPUTS`, grouped into six documented causes, and the test fails if one starts *passing* so the list keeps its teeth.

**4. `a_projection_is_a_fixed_point_on_its_own_axis` — a metamorphic relation.** Projecting a result back onto its own axis must converge. Its oracle is ferro run twice, so it is the one assertion here backed by neither the fixture nor a committed count. 326 same-axis rows, currently 0 violations.

### Two adjacent relations, measured and rejected

Both were suggested by an independent review; I implemented neither, because measurement showed the premises are unsound. Recorded in the test's doc comment so the next reader does not re-derive them.

- **`r.` equals `c.` with `t`→`u`.** False. 15 of 271 pairs differ, for sound reasons: `r.` drops intronic offsets, because a spliced RNA has no introns (`c.94-23_188+33=` → `r.(94_188=)`), and its numbering base differs (`c.3921del` → `r.(3922del)`). Asserting it would pin a false invariant.
- **`c` → `g` → `c` round-trip identity.** Not expressible — a `g.` result carries no transcript frame, so the return leg declines by design.

The same review cited `LRG_763t1:c.54_110GCA[23]` → `c.54_116GCA[25]` as un-oracled drift. It is not drift: the projection converges, so it is normalization, and the new fixed-point test covers it.

## Status movement

Per the `src/conformance/**` path instruction, which requires quantifying bucket changes:

| status | before | after |
|---|---|---|
| `projection-pinned` | 1190 | 1180 |
| `projection-splits-single-member` | 0 | 10 |

No other bucket moves. The 10 are pre-existing behaviour being *counted*, not new defects — though several are worth a look on their own terms, e.g. ferro rendering the spec's `c.145_147delinsTGG` as `g.[494841C>T;494843C>G]`, which is the split spelling `delins.md:42` calls incorrect.

**`src/normalize/merge.rs` is byte-identical to `main`** — the perturbation was an experiment, verified reverted with `git diff --exit-code`.

## Also fixed here

- **CLAUDE.md named five consumers of the normalization fixture; only two read it.** `mito_circular_audit.rs`, `protein_silent_eq.rs` and `protein_unknown_roundtrip.rs` cite it in doc comments only. Corrected, with a table of which guards are real oracles.
- **The "Regenerate if intentional" messages**, on both drivers. They now say a drift can only mean a stale local artifact, and point at the committed guards instead of at the regenerate command.
- **A registration-completeness miss in my own change**: the `Status` round-trip test enumerates every variant by name, so a new one is silently untested. Added the variant and pinned the count.

## Verification

- `cargo nextest run --features dev` — 9003 passed, 161 skipped
- `cargo fmt --all --check`, `cargo clippy --features dev --all-targets` — clean
- Guard proven to fire: with the perturbation applied, `divergence_budget_is_unchanged` fails `projection-splits-single-member: budget 10, actual 11` and `passing_census_is_unchanged` fails, while `enumeration_replays_recorded_behavior` still passes — confirming it was never the guard.

Reference volumes were offline for this run, so manifest-backed conformance tests skipped; none of this change touches them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved classification of projections that expand a single input member into multiple allele members.
  * Avoided incorrect classifications for insertion sources, accession dots, nested expressions, and translocations.
  * Added consistent serialization and recognition of the new projection outcome.

* **Tests**
  * Expanded conformance coverage for normalization, projection behavior, status enumeration, and known divergences.
  * Added safeguards to detect stale generated specifications and untracked behavioral changes.

* **Documentation**
  * Clarified the scope and validation role of generated fixtures and conformance tests.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #446, opened 2026-05-22
parser: reject single-position insertion 'g.123insG' per insertion.md:95-101


nh13:
**Spec source**: `DNA/insertion.md:95-101` — Q&A *"Can I describe a variant as `g.123insG`? **No**, since the description is not unequivocal, it is not allowed."* Marked `class="invalid"`.

**Current behavior**: ferro accepts `g.123insG` and round-trips it. The insertion location is a single point rather than a 2-position range.

**Expected**: parser (or post-parse semantic validation) rejects insertions whose location is a single point with a diagnostic explaining the anchor MUST be two adjacent positions (e.g. `g.123_124insG`). Mutalyzer raises `EINSERTIONRANGE`: *"Insertion variant is invalid; range positions \`123\` are not consecutive."*

**Test pin**: `tests/input_hygiene_rejections.rs::accepts_single_position_insertion_diverges_from_spec` (#444); flip to `assert_rejected` once fixed.

**Scope estimate**: add a semantic check in `src/normalize/validate.rs` or in the variant-assembly path of `src/hgvs/parser/variant.rs`. Insertion location must be a range. Also covers the related `c.456-13insG`, `c.-14insG` Q&A examples.

Source: phase4 spec-coverage triage corner `dna-insertion:C-C9`.

PR #1332, opened 2026-08-02
fix(project): decline axes ferro cannot render, instead of emitting them

### LINKED ISSUE #1264: Parser: accepts descriptions it cannot re-parse after rendering (empty allele, non-adjacent ins anchors)
## What's wrong

Two descriptions the parser **accepts** cannot be **re-parsed** after rendering. Distinct from #1263: normalization is not introducing the breakage, it is faithfully passing through a description that was already un-round-trippable.

```
input == output == []
  Parse error at position 0: Failed to parse accession: Error { input: "[]", code: Alpha }

input == output == LRG_199:g.699646_700297insNNNN…  (100 N)
  Parse error at position 0: insertion anchor positions are not flanking on g.;
  the two positions MUST be adjacent (e.g. g.123_124insATG, not g.123_125insATG)
  per DNA/insertion.md:15
```

The second is the interesting one: the parser accepted non-adjacent `ins` anchors on the way in and rejects them on the way back out. That asymmetry means the inbound path has a validation gap, the outbound rule is stricter than the inbound one, or both.

The first is likelier benign — an empty allele has no valid rendering, so the question is whether an empty `AlleleVariant` should be constructible at all.

## How it was found

While calibrating the `FERRO_ASSERT_REPARSE` oracle added in #1259. Both surfaced as oracle failures before it was narrowed to fire only when the *input* re-parses and the *output* does not — the exemption that isolates #1263. They are recorded in that oracle's doc comment so the exemption's rationale does not go missing.

## Why file it separately

The exemption is load-bearing for #1263's "exactly one failure in 8,826 tests" measurement, and an exemption with no tracked follow-up quietly becomes permanent. These are real, just not the same bug.

## Suggested first step

Decide whether `parse_hgvs` should reject non-adjacent `ins` anchors inbound (making it symmetric with the render path), or whether the render-side rule is too strict. `DNA/insertion.md:15` is cited by the outbound rejection and reads as a genuine MUST, which points at the inbound path being the lenient one.

nh13:
Closes #1264

## The issue's diagnosis is wrong, and measuring it is what found the real defects

#1264 reports two descriptions ferro accepts but cannot re-parse, and attributes them to a lenient parser: *"the parser accepted non-adjacent `ins` anchors on the way in and rejects them on the way back out."*

It does not. `validate_no_point_insertion` (`src/hgvs/parser/variant.rs`, from #446/#450) already rejects both the single-position and the non-adjacent anchor inbound on all six nucleotide axes. `LRG_199:g.699646_700297insNNN…` — the exact string the issue quotes — does not parse:

```
Parse error at position 0: insertion anchor positions are not flanking on g.;
the two positions MUST be adjacent (e.g. g.123_124insATG, not g.123_125insATG)
per DNA/insertion.md:15
```

Those strings were never produced by `parse_hgvs`. They were built by the **projector**, and reached the `FERRO_ASSERT_REPARSE` oracle as its *input* — where the blanket exemption ("skip when the input does not itself re-parse") swallowed them. That is precisely the risk the issue names: *"an exemption with no tracked follow-up quietly becomes permanent."*

Instrumenting that exemption to report instead of return silently, then re-running the suite, gives the census — **9005 tests, 0 failures, 18 hits in four shapes**:

| rendered shape | hits | cause |
|---|---|---|
| `LRG_199:g.699646_700297insNNN…` | 8 | insertion at a splice junction |
| `NC_000023.11:g.spl`, `g.(spl)` | 8 | the RNA-only `spl` edit on a `g.` axis |
| `[]` | 2 | an `AlleleVariant` built with no members |

## What is fixed

**The RNA-only effects.** `r.spl` ("splicing affected", `RNA/splicing.md`) and `r.0` ("no product") describe the transcript, not a sequence change at a locus. The projector had **no `NaEdit::Splice` handling anywhere** — it converted the position, which maps perfectly well, and carried the edit token through verbatim, emitting `g.spl`, `c.1spl` and `n.245spl`. It also invented a protein consequence, `NP_003997.1:p.(Met1?)`, for a variant that states no coding change. `hgvs_to_vcf` already declined exactly these two edits; the conversion paths now agree.

**The junction insertion.** `r.1149_1150ins…` is well formed — the anchors are neighbours in the spliced RNA — but when the pair straddles an exon–exon junction their genomic images sit either side of the intron. There is no repair: the insertion could attach to the 3' end of the upstream exon or the 5' end of the downstream one, and those are *different* genomic variants. HGVS has no spelling for "at the junction", so the genomic axis declines.

Both decline **per axis, not per projection**. The guard sits on the *reported* form, not on the pivot that also feeds the cdot derivation of c./n./p. — an earlier attempt guarded the pivot and failed the whole projection, turning "four right axes and one wrong one" into nothing. Pinned by `a_declined_genomic_axis_does_not_take_the_other_axes_with_it`.

This is the `src/{project,…}` path instruction's own rule: *"Projection must DECLINE rather than emit a coordinate it cannot justify."*

## Also fixed — found while measuring, not filed separately

**A real parser hole, in the construct the issue's framing points at but did not name.** `validate_no_point_insertion` ends in `_ => {}` and never inspected `GenomeRing` segments, even though they are `LocEdit<GenomeInterval, NaEdit>` — the same loc/edit pair every other arm checks. Both MUST-level sub-rules were unenforced inside a `::`-join:

```
ACCEPTED  NC_000022.11:g.100_102insATG::200_201insG      <- non-adjacent anchors
ACCEPTED  NC_000022.11:g.100insATG::200_201insG          <- single-position anchor
rejected  NC_000022.11:g.[100_102insATG;300A>G]          <- Allele was recursed into
```

These **round-trip**, so the re-parse oracle can never see them; only a parser test can.

**The empty allele.** `AlleleVariant::checked` refuses an empty member list, mirroring `GenomeRing::new` ("a 0/1-segment ring would not round-trip"). `new` is deliberately left infallible — it is called throughout the parser and normalizer on lists that are non-empty by construction.

## The exemption is now a closed list

`FERRO_ASSERT_REPARSE`'s exemption was a blanket. It is now three named shapes, each justified in place: the `0`/`?` whole-allele markers, an empty allele, and the non-flanking genomic *pivot* (whose reported axis is withheld). Anything else that arrives unparseable now fails. Re-measured after the fixes: **18 hits in 4 shapes → 4 hits in 2**.

## Bucket movement (`src/conformance/**` instruction)

22 enumeration rows move; the 2184-row total is unchanged.

| status | before | after | delta |
|---|---|---|---|
| `projection-pinned` | 1180 | 1167 | −13 |
| `projection-error-pinned` | 215 | 210 | −5 |
| `projection-unavailable-pinned` | 469 | 487 | +18 |

Every moved row was a projection with no business rendering. **Two rows improve the other way**, from hard error to rendered: `r.0` on `NM_004006.1` previously failed the whole projection with "Transcript has no CDS" and now renders `r.(0)` on its own axis — declining the axes that do not apply is what let the one that does apply through.

The four `project-r/…r.spl` rows are deliberately **unmoved**: the `rna` axis is documented as the predicted form and already rendered `r.(spl)`, so the replacement path preserves that wrapper rather than quietly changing it. The normalization fixture is byte-identical (934 rows, 0 changed).

`passing_census_is_unchanged` — the guard added in #1272 — is what caught this and forced the counts to be updated deliberately. Working as intended.

## A rejected alternative, recorded

The first attempt was a single post-condition at the projection seam: withhold any axis whose rendering fails to re-parse. Across 9011 tests it produced 2 failures, and one refuted the design — it withheld a **correct** `p.` axis because the fixture accession `NP_TEST_MINUS.1` is not parseable (the parser rejects the second underscore). It coupled axis availability to accession-syntax support, so any consumer with a custom accession would silently lose every axis. Replaced by a structural check on the interval, which cannot repeat that failure. The reasoning is kept in `non_flanking_genomic_insertion_anchor`'s doc comment.

## Verification

- `cargo nextest run --features dev` with **`FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1`** — **9028 passed**, 0 failed (proptests included).
- `cargo fmt --all --check` clean; `cargo clippy --features dev --all-targets` silent; `cargo check --features python` clean.
- No new seeds in `tests/proptest-regressions/` — no proptest strategy was touched.
- Manifest-backed conformance tests skip (scratch volume offline); nothing here needs them.

## Reading order

1. `tests/it/issue_1264_reparse_asymmetry.rs` — the census and the ten regressions, including the anti-vacuity control.
2. `src/project/projector.rs` — `project_rna_only_effect` and `non_flanking_insertion_reason`.
3. `src/hgvs/parser/variant.rs` — the `GenomeRing`/`Supernumerary` arms.
4. `src/normalize/mod.rs` — the narrowed exemption.

One note on the test fixture, since it is the reason the junction test is not vacuous: an earlier draft used `CGCG…` exons and an all-`N` intron, against which the junction insertion normalized to `g.1030_1199N[173]` — a repeat spanning the intron, which *does* re-parse, so the test passed while proving nothing. A homopolymer is not a representative intron.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of genomic insertion anchors, including wrapped and multi-segment variants.
  * Prevented invalid RNA-only and non-flanking insertions from producing incompatible projections.
  * Preserved specific reasons when genomic projections are unavailable.
  * Improved handling and validation of empty alleles.

* **Tests**
  * Added coverage for parsing, normalization, projection, and allele round trips.
  * Updated projection behavior expectations for splice effects and invalid insertions.

* **Documentation**
  * Clarified normalization re-parse exceptions and documented discovered cases.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -57,12 +57,24 @@ skips its own check.
 
 `FERRO_ASSERT_REPARSE=1` is the second half of the same seam: it asserts that a
 normalized description is one `parse_hgvs` accepts — *when normalization is what
-broke it*. Two exemptions keep that scope honest: `0` and `?` are legal
-whole-allele outputs that `parse_hgvs` rejects standalone because it wants an
-accession, and an input that does not itself re-parse is passed over, since a
-description the parser accepts inbound but cannot re-read outbound is a
-parse/display round-trip asymmetry normalize merely carries through. Both are a
-different bug; reporting them here would bury the one this oracle is for.
+broke it*. Its exemptions are a **closed list**, deliberately (#1264):
+
+- `0` and `?` are legal whole-allele outputs that `parse_hgvs` rejects standalone
+  because it wants an accession — a limit of the entry point, not a malformed output.
+- An **empty allele** (`[]`), reachable only by direct construction; the projector's
+  own tests build one on purpose to pin that it declines.
+- A **non-flanking genomic insertion**, which is the projection *pivot*: its
+  coordinates are sound and the downstream cdot derivation of the c./n./p. axes needs
+  it, but its spelling is not one HGVS admits, so the projector withholds the
+  *reported* genomic axis instead (`non_flanking_genomic_insertion_anchor`).
+
+**This used to be a blanket** — "skip when the input does not itself re-parse" — and
+that blanket was hiding live defects rather than scoping the oracle. Instrumenting it
+to report instead of return silently, then re-running the suite, found 18 hits in four
+shapes, two of which were real projector bugs (the RNA-only `spl` edit carried onto the
+`g.`/`c.` axes, and an insertion whose anchors straddle a splice junction). Both are
+fixed; the exemption is now narrow enough that the same class of defect fails loudly.
+If you find yourself widening it, that is the signal to fix the producer instead.
 
 Like the idempotency oracle it is compiled out in release builds
 (`#[cfg(debug_assertions)]`) and read once into a `OnceLock`, so it adds no
--- a/src/hgvs/parser/variant.rs
+++ b/src/hgvs/parser/variant.rs
@@ -7672,12 +7672,16 @@ fn validate_no_point_insertion(variant: &HgvsVariant, _source: &str) -> Result<(
         )
     }
 
-    /// Apply both sub-rules to one nucleotide variant's loc/edit pair.
-    macro_rules! check_na_anchor {
-        ($v:expr, $prefix:expr) => {{
-            if is_insertion(&$v.loc_edit.edit) {
-                let start = &$v.loc_edit.location.start;
-                let end = &$v.loc_edit.location.end;
+    /// Apply both sub-rules to one loc/edit pair.
+    ///
+    /// Takes the `LocEdit` rather than the variant so the ring path — whose
+    /// `segments` are bare `LocEdit`s with no enclosing variant — shares the
+    /// exact rule the single-variant axes get, instead of a re-derived copy.
+    macro_rules! check_loc_edit_anchor {
+        ($le:expr, $prefix:expr) => {{
+            if is_insertion(&$le.edit) {
+                let start = &$le.location.start;
+                let end = &$le.location.end;
                 if pos_eq(start, end) {
                     return Err(make_error($prefix));
                 }
@@ -7685,6 +7689,13 @@ fn validate_no_point_insertion(variant: &HgvsVariant, _source: &str) -> Result<(
                     return Err(make_gap_error($prefix));
                 }
             }
+        }};
+    }
+
+    /// Apply both sub-rules to one nucleotide variant's loc/edit pair.
+    macro_rules! check_na_anchor {
+        ($v:expr, $prefix:expr) => {{
+            check_loc_edit_anchor!($v.loc_edit, $prefix)
         }};
     }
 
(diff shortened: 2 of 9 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Addressed all five review threads. Two notes where the review and the measured behaviour differed:

**RNA-only edits on every transcript axis** — half of this was real, half was not. `n.0` genuinely slipped the guard: it parses to `HgvsVariant::Tx`, and the `if let HgvsVariant::Rna(v)` test in `project_to_genomic_nc` let it through, so `NC_000001.11(NM_TEST.1):n.0` reported its genomic axis as `NC_000001.11:g.0`. The guard now reads the extracted edit after axis extraction, so it holds for `c.`/`n.`/`r.` alike, with a regression test verified to fail before the fix. `c.1spl` is not reachable as an *input* — `parse_na_edit` has no `spl` arm, and `parse_rna_splice` is only probed from the `r.` whole-entity paths — so no change was made for it. The `c.1spl` in the original census was projector *output*, which is what this PR stops producing.

**Census gross movements** — the suggested 15/3/2 is not what the data says. The three net deltas are linearly dependent, so they do not determine the gross figures; the suggestion resolved the free parameter using this file's own claim that "two rows moved the other way", and that claim was wrong. Generating the enumeration at the merge base and diffing per row gives:

```
14 rows  projection-pinned       -> projection-unavailable-pinned
 4 rows  projection-error-pinned -> projection-unavailable-pinned
 1 row   projection-error-pinned -> projection-pinned
```

Only `project-r/NM_004006.1:r.0` improves to rendered; its other four axes move to unavailable in the same step. A further 3 rows change their decline *reason* without changing status, which is where the "22 rows move" in the commit message comes from. The comment and the commit message now state all of this gross. Verified byte-identical enumeration output otherwise, and the merge with current `main` builds and passes 9044 tests locally.

nh13:
@coderabbitai review

PR #1333, opened 2026-08-02
fix(normalize): order protein cis members by junction, like the nucleotide axes

### LINKED ISSUE #1328: normalize: protein cis members sharing a span render in the wrong order
#1301 fixed member ordering for two cis members that share a span but add sequence at different points within it: `junction_rank` in `src/hgvs/variant.rs` ranks an `NaEdit::Insertion` (which adds at its span's start) before anything that acts across the whole span. The protein axis has the same defect and is not covered.

Both preconditions hold there:

- `cis_member_range` returns a real range for `HgvsVariant::Protein`, so two protein members sharing a span tie on `(start, end)` exactly as nucleotide members did.
- `sort_cis_members_by_genomic_order` is axis-agnostic — it sorts any single-accession cis allele, protein included.

`junction_rank` matches only the six nucleotide-bearing arms, so every protein member ranks `1` and the tie falls through to the rendered-descriptor tie-break. That sorts `dup` before `ins` alphabetically:

```
p.[Gly4_Ala5insSer;Gly4_Ala5dup]  ->  renders the dup first
```

which reverses the order the members apply in — the insertion adds at the gap before Ala5, the duplication places its copy after it. This is the same shape as #1301, so their interbase spans read as overlapping in the same way.

Suggested work:

- Give `junction_rank` a `HgvsVariant::Protein` arm ranking `ProteinEdit::Insertion` `0`, matching the nucleotide treatment.
- Add a protein cis-allele regression asserting the rendered order is stable and independent of the authored order, alongside the nucleotide cases in `tests/it/issue_1301_adjacent_gap_member_order.rs`.
- Drop the "nucleotide axes only" scope note on `junction_rank` once it is axis-neutral.

Found while dual-reviewing #1305, which fixed the nucleotide half. Deliberately not widened there: protein ordering is a behavior change on an axis that PR does not otherwise touch.

nh13:
Closes #1328

## The defect

`junction_rank` breaks the tie between two cis members that share a span but add sequence at different points within it (#1301). An insertion's span *is* the gap it fills, so it adds at the span's start; a duplication places its copy after the span's last base.

It matched only the six nucleotide-bearing arms. The protein axis reaches the same tie for the same two reasons the issue names — `cis_member_range` returns a real range for `HgvsVariant::Protein`, and `sort_cis_members_by_genomic_order` is axis-agnostic — so every protein member ranked `1` and the tie fell through to the rendered-descriptor tie-break, which sorts `dup` before `ins` alphabetically:

```
p.[Gly4_Ala5insSer;Gly4_Ala5dup]  ->  p.[Gly4_Ala5dup;Gly4_Ala5insSer]
```

`Gly4_Ala5insSer` fills the gap between residues 4 and 5; `Gly4_Ala5dup` copies both and places the copy after residue 5. Rendering the duplication first reverses the order the two apply in, so their spans read as overlapping — the same shape as #1301, reversed the same way. Reproduced from **both** authorings before the fix.

## The fix

`ProteinEdit::Insertion` now ranks `0`, matching `NaEdit::Insertion`. The two edit enums are matched separately because they are separate types, not because the rule differs. The "nucleotide axes only" scope note is dropped — it was the only reference to #1328 in the tree.

## Tests

Three regressions land beside the nucleotide cases in `tests/it/issue_1301_adjacent_gap_member_order.rs`, as the issue asks:

- `a_protein_insertion_sorts_before_a_duplication_sharing_its_span` — the defect.
- `the_protein_order_does_not_depend_on_how_it_was_authored` — this **already held** before the fix, because the descriptor tie-break is itself total; both authorings produced `[dup;ins]`. Pinned so the new rank cannot trade one defect for the other — a rank that made protein order depend on how the allele was written would be a worse bug than the one being fixed.
- `protein_members_that_are_not_insertions_keep_their_existing_order` — the control, so the change cannot pass by reordering everything.

## Blast radius

Full suite with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1`: **9021 passed, 0 failed**. No other protein output moved, and the enumeration census is unchanged — so despite being a Display-order change on an axis compared byte-for-byte against external oracles, nothing else in the corpus reaches this tie.

`cargo fmt --all --check`, `cargo clippy --features dev --all-targets`, and `cargo check --features python` are all clean.

## A sibling I looked at and am deliberately **not** changing

Sweeping for the same shape — a missing `Protein` arm in an axis dispatcher — turned up `is_whole_entity_identity_rhs` (`src/hgvs/variant.rs`), which has none, while `intervals_match_for_compact_mosaic` directly beside it does. It gates the compact mosaic Display form, and the asymmetry is real and measurable:

```
NC_000001.11:g.123A>G/=   ->  g.123A>G/=                     (compact)
NP_000001.1:p.Trp24Cys/=  ->  p.Trp24Cys/NP_000001.1:p.=     (long form)
```

I am not widening it, because the spec does not support the compact form here. `protein/substitution.md:81-84` publishes exactly one protein mosaic example — `LRG_199p1:p.Trp24=/Cys`, the position-identity form, which ferro **already renders correctly** — and adds:

> **NOTE**: irrespective of the frequency in which each amino acid was found, the reference is always described first.

`p.Trp24Cys/=` describes the variant first, against that stated convention, and the spec publishes no compact `p.<var>/=` form. Making the branch fire for protein would invent a spec claim rather than fix a defect. (Separately, `p.=/Trp24Cys` does not parse at all while `g.=/123A>G` does — an inbound parser gap, a different function and a larger change than this issue.)

Recording it here rather than filing it, with the measurement and the citation, so the next person does not re-derive the analysis.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected ordering for protein variants that share the same span.
  * Protein insertions now appear before duplications, regardless of input order.
  * Preserved existing ordering behavior for other duplication variants.

### DIFF
--- a/src/hgvs/variant.rs
+++ b/src/hgvs/variant.rs
@@ -757,7 +757,14 @@ pub(crate) fn non_flanking_genomic_insertion_anchor(variant: &HgvsVariant) -> Op
     if start.special.is_some() || end.special.is_some() {
         return None;
     }
-    (end.base != start.base + 1).then_some((start.base, end.base))
+    // Asked as a subtraction on `end`, not `start.base + 1`: the parser accepts
+    // any `u64` position, so a `start` of `u64::MAX` overflows the addition on a
+    // pair whose answer is plainly "not flanking". That panics under
+    // `debug_assertions` and *wraps* in release, where `[profile.release]` sets
+    // no `overflow-checks` — so the panic is not what would surface it in the
+    // field. `checked_sub` returns `None` both there and for a reversed pair,
+    // and neither is flanking.
+    (end.base.checked_sub(start.base) != Some(1)).then_some((start.base, end.base))
 }
 
 impl AlleleVariant {
@@ -1237,14 +1244,31 @@ pub(crate) fn cis_member_order_key(
 /// and so ranks with the 3' end: a member that adds at a span's start must sort
 /// before one that acts across the same span.
 ///
-/// **Nucleotide axes only.** A protein member reaches the same tie —
-/// `cis_member_range` gives it a real range, and `sort_cis_members_by_genomic_order`
-/// is axis-agnostic — so `p.[Gly4_Ala5insSer;Gly4_Ala5dup]` still falls to the
-/// descriptor tie-break and renders the `dup` first, exactly as #1301 did here.
-/// Ranking `ProteinEdit::Insertion` too would fix it, but that changes protein
-/// ordering, which is outside this change; tracked in #1328.
+/// **Every axis**, protein included (#1328). The protein axis reaches this tie
+/// for the same two reasons the nucleotide axes do — `cis_member_range` returns
+/// a real range for [`HgvsVariant::Protein`], and `sort_cis_members_by_genomic_order`
+/// is axis-agnostic — so while this matched only the six nucleotide-bearing
+/// arms, every protein member ranked `1` and the tie fell to the descriptor
+/// tie-break, which sorts `dup` before `ins` alphabetically:
+/// `p.[Gly4_Ala5insSer;Gly4_Ala5dup]` rendered the `dup` first. That is the same
+/// defect #1301 fixed for nucleotides, reversed the same way.
+///
+/// `ProteinEdit::Insertion` is the protein analogue of `NaEdit::Insertion`: its
+/// span is the gap it fills, so it adds at the span's start, where a `dup`
+/// places its copy after the span's last residue. The two edit enums are
+/// matched separately because they are separate types, not because the rule
+/// differs.
 fn junction_rank(v: &HgvsVariant) -> u8 {
-    use crate::hgvs::edit::NaEdit;
+    use crate::hgvs::edit::{NaEdit, ProteinEdit};
+
+    // The protein axis carries its own edit enum, so it cannot share the
+    // `NaEdit` match below; the *rule* it applies is identical.
+    if let HgvsVariant::Protein(p) = v {
+        return u8::from(!matches!(
+            p.loc_edit.edit.inner(),
+            Some(ProteinEdit::Insertion { .. })
+        ));
+    }
     let edit = match v {
         HgvsVariant::Genome(x) => x.loc_edit.edit.inner(),
         HgvsVariant::Circular(x) => x.loc_edit.edit.inner(),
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Both findings applied.

**Inline (`tests/it/issue_1301_adjacent_gap_member_order.rs`) — non-insertion ordering contract.** Correct, and the pair was blind in exactly the way described: `Ala3dup` / `Ala3_Ala4dup` differ on `end`, so `cis_member_order_key` settles them on range and neither `junction_rank` nor the descriptor tie-break is consulted. Replaced with `Ala3_Ala4del` / `Ala3_Ala4dup`, which tie on both endpoints, authored in reverse descriptor order, and both authorings are now checked. Verified by mutation — ranking `ProteinEdit::Duplication` `0` alongside `Insertion` flips the new pair and leaves the old one untouched:

```
                                          baseline            mutated
p.[Ala3dup;Ala3_Ala4dup]              [Ala3dup;Ala3_Ala4dup]  [Ala3dup;Ala3_Ala4dup]     <- blind
p.[Ala3_Ala4dup;Ala3_Ala4del]         [Ala3_Ala4del;...dup]   [Ala3_Ala4dup;...del]      <- sensitive
```

**Outside diff (`src/hgvs/variant.rs:760`) — insertion-anchor overflow.** Real and reachable from parsed input, not just direct construction: `g.18446744073709551615_18446744073709551614insA` parses, and a unit test on it panicked with `attempt to add with overflow` at that line before the change. Switched to `end.base.checked_sub(start.base) != Some(1)` and pinned it with `the_insertion_anchor_check_survives_a_position_at_the_top_of_the_range`. Noted in the comment that `[profile.release]` sets no `overflow-checks`, so release wraps rather than panics — the panic is not what would surface this in the field.

Also added a scope note on the protein test block: unlike the nucleotide cases it carries no `assert_preserving` / `member_starts` SPDI cross-check, because `hgvs_to_spdi` rejects `HgvsVariant::Protein` outright, so there is no second code path on this axis.

Full suite green with `FERRO_ASSERT_IDEMPOTENT=1` and `FERRO_ASSERT_REPARSE=1` (9048 tests).

PR #1335, opened 2026-08-02
fix(liftover): return an error for position 0 instead of panicking

### LINKED ISSUE #1334: Liftover panics on position 0 instead of returning an error
Follow-up to #1282, which added `hgvs_pos_to_index_checked` but wired it into only one consumer (the PyO3 binding). The Rust call sites still use the panicking `hgvs_pos_to_index`, and one of them is reachable from unvalidated external input.

## The path

`src/service/handlers/liftover.rs` `parse_genomic_position` parses the position with a bare `pos_part.parse::<u64>()` and applies no lower bound, so a request for `chr7:0` yields `Ok(("chr7", 0))`. That flows into `Liftover::lift`, which converts with `hgvs_pos_to_index(pos)` — and since #1282 that call asserts on `0`.

Before #1282 the same input wrapped to a huge index in release (`[profile.release]` sets no `overflow-checks`), `find_chains` returned empty, and the handler produced a 200 with an `error` field. After #1282 it panics, killing the handler task and dropping the connection.

`Liftover::lift_interval` has the same issue on both endpoints, and the `ferro liftover` CLI reaches `lift` the same way.

## What #1282 asked for

> Guard the conversion (return `None`/`Err` for `pos < 1`) rather than relying on callers never producing a zero.

`lift` and `lift_interval` already return `Result<_, FerroError>`, so this is a swap to the checked conversion with no signature change — the handler already renders `Err` as a graceful 200 with an error body.

## Remaining sweep

#1282 also asked for a sweep of other position-to-index conversions. `src/coords/mod.rs` was swept (it turned up `hgvs_to_spdi_pos`), but these call sites are still unaudited:

- `src/normalize/validate.rs:194,353,559,584`
- `src/normalize/rules.rs:436,454`
- `src/benchmark/cache.rs:653,1229,2762`

Those take positions from parsed, validated variants rather than raw external input, so they are lower risk than the liftover path — but they should be confirmed rather than assumed.

nh13:
#1331 has landed, so this is no longer stacked — the branch is rebased onto `main` and the diff is a single commit touching only `src/liftover/lift.rs`.

Closes #1334

## Why

#1331 makes `hgvs_pos_to_index` assert on `0` rather than wrap to a near-`usize::MAX` index. That is the right call, but it added `hgvs_pos_to_index_checked` and wired it into exactly one consumer — the PyO3 binding — leaving the Rust call sites on the panicking path. One of them takes unvalidated external input.

`src/service/handlers/liftover.rs` `parse_genomic_position` parses with a bare `pos_part.parse::<u64>()` and applies no lower bound, so a request for `chr7:0` yields `Ok(("chr7", 0))`. That reaches `Liftover::lift`, which converted with `hgvs_pos_to_index(pos)`.

- **Before #1331:** wrapped to a huge index in release (`[profile.release]` sets no `overflow-checks`), `find_chains` returned empty, handler produced a 200 with an `error` field.
- **After #1331:** panics, killing the handler task and dropping the connection.

`lift_interval` has the same problem on *both* endpoints, and the `ferro liftover` CLI reaches `lift` the same way.

This is what #1282 asked for in the first place:

> Guard the conversion (return `None`/`Err` for `pos < 1`) rather than relying on callers never producing a zero.

## What changed

All three conversions now go through a shared `position_to_index` helper backed by `hgvs_pos_to_index_checked`. `lift` and `lift_interval` already return `Result<_, FerroError>`, so there is no signature change, and the handler's existing `Err` arm renders the result as the same graceful response it produced before #1331.

The helper is shared rather than inlined three times deliberately: the failure mode being guarded against is precisely one endpoint keeping the unchecked call, and `lift_interval` panicked on a zero `end` as readily as a zero `start`.

## Testing

Two tests in `src/liftover/lift.rs`, both confirmed to **fail without the fix** — they panic at `lift.rs:149` and `lift.rs:226`, the exact conversion sites — and pass with it:

- `lift_rejects_position_zero`
- `lift_interval_rejects_position_zero_at_either_end` (covers a zero `start` *and* a zero `end`, so fixing only one endpoint still fails)

Both assert the specific `FerroError::InvalidCoordinates` variant rather than merely that an error occurred.

Plus a pinned integration regression, `tests/it/issue_1334_liftover_position_zero.rs`, which exercises the same guard across the crate boundary — which is where the offending callers (the web service handler and the CLI) actually live:

- `lift_returns_invalid_coordinates_for_position_zero` — pins the *new* output, not just "is an error": the message names the caller's contig and the 1-based convention.
- `lift_interval_returns_invalid_coordinates_for_a_zero_at_either_end` — `(0, 10100)`, `(10001, 0)` and `(0, 0)`.
- `valid_positions_still_lift_to_their_previous_coordinates` — the negative control. `chr1:10001 -> chr1:10051` and the interval `10001_10100 -> 10051_10150` still lift to the coordinates hand-derived from the chain header, and position 1 still fails at the chain lookup (`No chain found`) rather than at the new guard. Without it, a guard that rejected `pos <= 1` — or everything — would satisfy the two tests above.

The chain is built programmatically from an inline string; no fixture is committed. The test resolves no `FERRO_MANIFEST`, so it runs on PR CI rather than only in the nightly.

**Neither toward nor away — no oracle has an opinion here.** Liftover is UCSC chain-file arithmetic, not an HGVS description, so biocommons hgvs, Mutalyzer and VariantValidator have nothing to say about `Liftover::lift`. For every input these entry points previously *answered* — every `pos >= 1` — the output is byte-identical, which `valid_positions_still_lift_to_their_previous_coordinates` pins. The only inputs whose behaviour moves are the ones that panicked (post-#1331) or wrapped to a near-`usize::MAX` index (pre-#1331); both are replaced by a returned `Err`.

Full suite: 9024 passed, 0 failed. `cargo fmt --check` and `clippy --all-targets -- -D warnings` clean.

## Not in scope

#1334 also lists position-to-index conversions that #1282's sweep did not reach — `src/normalize/validate.rs:194,353,559,584`, `src/normalize/rules.rs:436,454`, `src/benchmark/cache.rs:653,1229,2762`. Those take positions from parsed, validated variants rather than raw external input, so they are lower risk than the liftover path and are left for a separate audit rather than widening this PR. They remain tracked on the issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Liftover conversions now safely reject invalid zero-valued coordinates instead of potentially failing unexpectedly.
  * Error messages identify the affected contig and coordinate.
* **Documentation**
  * Added documentation describing invalid-coordinate errors.
* **Tests**
  * Added coverage for invalid single-position and interval endpoint conversions.

### DIFF
--- a/src/liftover/lift.rs
+++ b/src/liftover/lift.rs
@@ -14,7 +14,7 @@
 //! For type-safe coordinate handling, see [`crate::coords`].
 
 use super::chain::ChainFile;
-use crate::coords::{hgvs_pos_to_index, index_to_hgvs_pos};
+use crate::coords::{hgvs_pos_to_index_checked, index_to_hgvs_pos};
 use crate::error::FerroError;
 use crate::reference::transcript::GenomeBuild;
 use std::path::Path;
@@ -127,6 +127,14 @@ impl Liftover {
     /// Lift a single position from one build to another.
     ///
     /// Positions are 1-based (HGVS convention).
+    ///
+    /// # Errors
+    ///
+    /// Returns [`FerroError::InvalidCoordinates`] if `pos` is `0` — there is no
+    /// 1-based position 0 — if the build pair is not one this instance carries
+    /// chains for, if no chain covers `contig:pos`, or if the position falls in
+    /// an alignment gap. All four share the one variant, so callers that need to
+    /// tell them apart have only the message.
     pub fn lift(
         &self,
         from: GenomeBuild,
@@ -136,8 +144,11 @@ impl Liftover {
     ) -> Result<LiftoverResult, FerroError> {
         let chain_file = self.get_chain_file(from, to)?;
 
-        // Convert from 1-based HGVS to 0-based for chain operations
-        let pos_0based = hgvs_pos_to_index(pos) as u64;
+        // Convert from 1-based HGVS to 0-based for chain operations. `pos` is a
+        // raw `u64` off a public API — the web service hands one straight from
+        // the query string — so it is not known to be a valid 1-based position
+        // and the conversion has to be the checked one (#1282).
+        let pos_0based = position_to_index(contig, pos)?;
 
         // Find chains covering this position
         let chains = chain_file.find_chains(contig, pos_0based);
@@ -196,6 +207,11 @@ impl Liftover {
     }
 
     /// Lift an interval from one build to another.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`FerroError::InvalidCoordinates`] if either endpoint is `0`, or
+    /// if the build pair is not one this instance carries chains for.
     pub fn lift_interval(
         &self,
         from: GenomeBuild,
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1336, opened 2026-08-02
fix(normalize): unify the complement helpers and case-fold the delins path

### LINKED ISSUE #1318: normalize: unify the three complement helpers, and fix the case-sensitive comparisons on the delins path
The crate has three independent "complement a base" helpers, and they disagree on exactly the two axes that produce bugs — case handling, and what happens to a byte outside the modelled alphabet. That divergence is the structural reason #1249 was possible, and one instance of it is still live on the delins path.

## The three helpers

| helper | case | byte outside the alphabet |
|---|---|---|
| `normalize/rules.rs::complement(u8) -> u8` (`rules.rs:913`) | folds to **uppercase** | returns the byte **unchanged** |
| `normalize/merge.rs::complement_base(u8) -> Option<u8>` (`merge.rs:2348`) | uppercases the input first, returns uppercase | returns **`None`** |
| `sequence.rs::complement_char(char) -> char` (`sequence.rs:28`) | **preserves** case | returns the char unchanged |

`merge.rs::complement_base` is the correct design and is the one that has not had a bug: it calls `to_ascii_uppercase()` on its input *before* matching, and it returns `Option` so an unmodelled byte cannot be silently mistaken for a self-complementary one.

`sequence.rs::complement_char` is internally consistent (`a -> t`) and correct for its own callers — the problem is not that function, it is that a reader cannot assume which contract they are getting.

## Why `rules.rs::complement` is the one to change

It has neither property. Its `_ => base` arm means a caller asking "does inverting this base change nothing?" gets `true` for `X` — not because `X` is self-complementary, but because there is no complement to report. Combined with the uppercase fold, that is what produced #1249's degenerate output: `complementary_substitution(b'w')` returned `Some((W, W))`, a substitution whose reference equals its alternative, which normalization emitted as `<pos>W>W`.

#1250 fixed the three call sites in the inversion-shortening path by routing them through new `is_complementary_pair` / `is_self_complementary` helpers that uppercase before complementing. **That works around the `_ => base` arm rather than removing it** — the trap is still there for the next caller who writes `complement(x) == x` directly.

## The still-unfixed sibling: the delins path

#1250 was deliberately scope-limited to inversion shortening. The same case-sensitivity exists on the delins→inv typing path and was left alone:

- `is_revcomp` (`rules.rs:1446`) — `.all(|(d, i)| complement(*d) == *i)`
- the shared-affix / run-detection byte comparisons at `rules.rs:1215`, `rules.rs:1426`, `rules.rs:1430`

On a soft-masked reference, `deleted` can arrive lowercase while `inserted` comes from the author's uppercase description. So a delins that genuinely *is* a reverse complement can fail to be typed as an `inv`, and shared-affix trimming can behave differently for the identical sequence in lowercase. Soft-masked reference bytes do reach normalization in this codebase — `tests/it/issue_1235_cis_allele_confluence.rs::soft_masked_reference_yields_the_same_canonical_form` covers that path.

## Suggested shape

1. Give `rules.rs::complement` the `merge.rs::complement_base` contract — `Option<u8>`, input uppercased before matching — or collapse the two into one helper.
2. Update the inversion-path callers (`is_complementary_pair` / `is_self_complementary` become thin or unnecessary once the underlying helper refuses unmodelled bytes).
3. Fix `is_revcomp` and the `canonicalize_delins` comparisons to fold case, with a soft-masked regression test in the style of `shorten_inversion_treats_soft_masked_bases_like_uppercase`.
4. Document the remaining split with `sequence::complement_char` so the case-preserving contract is a deliberate, stated choice.

Steps 1–3 touch overlapping call sites, which is why they belong together rather than in separate PRs. Step 3 changes `delins` recognition, so it wants its own conformance run.

Context: #1249 / #1250 — the motivating bug and the scope-limited fix.

nh13:
Closes #1318

Does all four steps of the issue's suggested shape, including the helper unification.

## The headline: this is a crash fix

The issue frames the delins case-sensitivity as a mis-typing — "a delins that genuinely *is* a reverse complement can fail to be typed as an `inv`". It is worse than that. Two comparisons on the same path disagreed about case:

- `is_revcomp` complements through a helper that folds to upper-case, so lower-case `deleted` vs upper-case `inserted` compared **equal**;
- `shared_affix_lengths` compared raw bytes, so the same pair shared **no** prefix or suffix and nothing was trimmed.

A palindromic delins therefore entered the inversion branch with an untrimmed range, `shorten_inversion` (correctly case-folding since #1250) collapsed it to nothing, and that branch's `expect` — whose stated invariant is that trimming has already handled the collapse-to-identity case — **panicked**:

```
thread '...' panicked at src/normalize/rules.rs:
revcomp delins cannot collapse to identity under shortening;
that case is handled by the Identity / Insertion branches above
```

Reproduced end to end through the public `Normalizer::normalize` on a soft-masked reference (`tests/it/issue_1318_soft_masked_delins.rs`), not just at the internal helper. Reference FASTAs are routinely soft-masked and those bytes reach normalization, so it is reachable from ordinary input.

## Step 1–2: one complement helper

`sequence::complement_base` is now the crate's single byte-level complement. It takes the contract of the one that never had a bug (`normalize::merge::complement_base`): **fold case**, and **return `None`** for a byte outside the modelled alphabet.

`normalize::rules::complement` is deleted. Its `_ => base` arm is the structural cause of #1249 — it made `complement(x) == x` true for every unmodelled symbol, so `is_self_complementary` answered `true` when the honest answer is "there is no complement to report". #1250 worked around that at three call sites; this removes it, so the trap is gone rather than routed around. `is_complementary_pair` / `is_self_complementary` survive as thin wrappers and are now honest — an unmodelled byte answers `false`.

`sequence::complement_char` keeps its **case-preserving** contract, and step 4's requested documentation says why that split is deliberate rather than accidental: it backs `reverse_complement`, which renders a sequence for a caller to read back, so it must not silently upper-case a soft-masked region or drop a byte it does not model. That is a display concern; `complement_base` is a comparison concern.

## Step 3: the delins path folds case — all of it

The issue names `is_revcomp` and the shared-affix / run-detection comparisons. Sweeping the function found the class was wider: **every** deleted-vs-inserted byte comparison in `canonicalize_delins` was raw. All of them now fold:

| site | what it decides |
|---|---|
| step 1 | whole-range identity |
| step 2 | duplication (`N` → `2N`) |
| `shared_affix_lengths` | prefix/suffix trimming |
| interior loop | identity runs within the decomposition |
| `is_revcomp` | inversion typing |

Fixing only the two the issue named would have moved the crash rather than removed it, and did: with trimming corrected, the range was fully consumed and the pure-insertion branch's `debug_assert!("Identity case caught above")` fired next, because step 1 was still raw-byte. That is the reason the sweep is exhaustive.

## Blast radius

Full suite with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1`: **9023 passed, 0 failed**. `cargo fmt --all --check`, `cargo clippy --features dev --all-targets`, `cargo check --features python` all clean.

The issue notes that step 3 "changes `delins` recognition, so it wants its own conformance run." It got one: **zero enumeration rows moved and the census is unchanged.** The reason is worth stating rather than reading as "no effect" — the spec corpus is entirely upper-case, so case-folding is a no-op over it. The corpus cannot exercise this change, which is why the new tests carry it.

## Tests

- `tests/it/issue_1318_soft_masked_delins.rs` — the end-to-end crash reproducer, plus a masked-vs-unmasked parity property over shapes that each reach a different comparison site.
- `canonicalize_delins_treats_soft_masked_bases_like_uppercase` — the same parity at the unit level.
- `a_soft_masked_delins_that_is_a_revcomp_is_typed_as_an_inversion` — uses `atgc`/`GCAT` deliberately, **not** a palindrome: a palindrome is its own reverse complement, so the insert equals the deleted bases and *identity* is the correct answer. My first draft used `aggcct` and asserted `Inversion`, which was wrong about the domain, not about the code.
- `is_revcomp_folds_case_on_both_sides` — including negative cases, so the fold cannot pass by making everything true.
- `test_complement` gains `complement_base(b'X') == None`, pinning the contract that replaced `_ => base`.

Parity assertions are stated as masked-vs-unmasked equality rather than pinned output shapes: whatever the canonical form of a given delins is, the reference's masking must not decide it.

## Follow-up from self-review

Three things the first pass missed, each measured rather than argued.

**The delins fix was defeated by its own unfixed twin.** `decompose_delins` case-folded the identity test at the top of its loop but not the mismatch-run scan below it — the same rule written twice, updated in one copy. On a soft-masked reference the run swallowed the very positions the identity test calls identities, so `has_identity` never became true and the function's final gate returned `None` for a span whose uppercase twin decomposes. Measured before: `decompose_delins(b"acgt", 0, 4, b"TCGA")` → `None` against `Some(4)` for `b"ACGT"`. Both now agree, pinned by `decompose_delins_treats_soft_masked_bases_like_uppercase`.

**`insertion_is_duplication` carried the identical defect** and is fixed with it: it compares a reference slice against the author's inserted sequence, so `insertion_is_duplication(b"acgtacgt", 4, b"ACGT")` answered `false` where the uppercase reference answered `true` — silently leaving a duplication spelled as an insertion. Pinned by `insertion_is_duplication_treats_soft_masked_bases_like_uppercase`.

**Unifying on `complement_base` changed what happens to a byte outside the modelled alphabet**, which the description did not call out. `complement` returned such a byte unchanged, so it read as self-complementary and an inversion over it collapsed to identity; `complement_base` answers `None`, so it does not. Measured: `shorten_inversion(b"X", 0, 1)` was `None`, is now `Some((0, 1))`. That makes the "Not a base we can complement" branch in `src/normalize/mod.rs` **reachable**, where its comment claimed it was unreachable *because of the behaviour this PR removes* — the comment is corrected, and the new behaviour is the right one (refusing is better than claiming an identity we cannot justify, the #1249 class). Pinned by `an_unmodelled_byte_no_longer_reads_as_self_complementary`.

Full suite with both oracles: **9052 passed, 0 failed.**

## Review round 2: pinned coverage and the oracle question

**Pinned canonical outputs, not just masked-vs-unmasked parity.** `tests/it/issue_1318_soft_masked_delins.rs` previously asserted only that the two references agree; two equally-wrong answers satisfy that. Each case now pins its canonical form as the primary assertion, keeping parity as the supplementary one. The eight cases land on eight *distinct* shapes — `29_34=`, `35_40=`, `63_70inv`, `32G>T`, `30A>T`, `[29C>G;36A>T]`, `31_36dup`, `31C>G` — which is what shows the table exercises separate branches rather than eight spellings of one.

**Two placeholder cases replaced with the shapes they claimed to be.** The "true reverse complement that is not an identity" entry was a duplicate of the palindrome above it, and the "whole-range identity" entry duplicated another. They are now `g.63_70delinsCGTACGTA` (63..70 is `TACGTACG`, not its own reverse complement, so `is_revcomp` must fire where the identity test must not — the discrimination a palindrome cannot make) and `g.29_36delinsGACGTGCT` (a genuine interior identity run flanked by mismatches, which is what the decomposition loop's run detector exists to find).

**The Duplication branch is now covered.** Every prior case had `inserted_seq.len() == deleted.len()`, so the `N -> 2N` branch's length gate was false and its case-folding was never exercised. Added `(b"acgt", 0, 4, b"ACGTACGT")` to `canonicalize_delins_treats_soft_masked_bases_like_uppercase`, plus `g.29_34delinsCACGTGCACGTG` end-to-end. Verified non-vacuous: reverting just that branch's two comparisons to raw-byte gives `Insertion { after_index: 4, … }` against `Duplication { start: 0, end: 4 }` — precisely the sub > del > inv > dup > ins priority violation the branch's comment warns about.

**Idempotency, and the inverse mixed-case orientation.** `normalization_is_idempotent_on_a_soft_masked_reference` asserts `norm(norm(x)) == norm(x)` against a lowercase reference with uppercase edit literals, and re-parses the intermediate. `an_rna_description_with_lowercase_payloads_takes_the_same_shapes` covers the mirror image — `r.` descriptions, whose literals are lowercase with `u` for `t`, against an *uppercase* reference. Every other case puts the lowercase bytes on the reference side, so a fold that works in one direction only was invisible to the whole table.

**All three integration tests were confirmed to fail against pre-fix production code** (`src/normalize/{rules,merge,mod}.rs` and `src/sequence.rs` reverted to `main`), not merely to pass after it. The palindromic case panics at `rules.rs:1427` on the `expect` this PR's description names.

**Toward, for soft-masked references; no movement at all for uppercase ones.** The uppercase path is untouched by the case-folding — every comparison this PR changes already compared equal when both sides were uppercase — so ferro's answer for every input the biocommons / Mutalyzer / hgvs-rs corpora actually contain is byte-identical. Those corpora are uppercase, which is why the whole class went unnoticed. What moves is the soft-masked path, and it moves *onto* the uppercase answer: the parity assertions in `issue_1318_soft_masked_delins.rs` state exactly that, and the uppercase answer is the one the oracles agree with. The one case where the new behaviour is not simply the old uppercase behaviour is the unmodelled-byte change described above (`shorten_inversion(b"X", 0, 1)`: `None` → `Some((0, 1))`), which no oracle covers — none of them accept `X` as a base — and which refuses rather than asserts an identity it cannot justify.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sequence normalization for mixed-case and soft-masked DNA bases.
  * Prevented incorrect complement matching for unsupported symbols.
  * Improved handling of inversions, substitutions, duplications, repeats, and delins normalization.
  * Fixed cases where certain palindromic delins inputs could fail unexpectedly.

* **Tests**
  * Added coverage for ambiguity codes, unsupported bases, soft masking, and tandem conversion scenarios.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -14,6 +14,7 @@ use crate::normalize::boundary::Boundaries;
 use crate::normalize::config::ShuffleDirection;
 use crate::normalize::shuffle::shuffle;
 use crate::reference::ReferenceProvider;
+use crate::sequence::complement_base;
 
 /// Coordinate-system region used as the merge-eligibility key.
 ///
@@ -2333,37 +2334,6 @@ fn canonical_base_byte(base: u8) -> u8 {
         b'U' => b'T',
         other => other,
     }
-}
-
-/// Watson-Crick complement of an IUPAC base, case-insensitive, as uppercase.
-/// `None` for a byte that is not a recognised base.
-///
-/// Matched here rather than delegated to [`crate::sequence::reverse_complement`],
-/// which is a *display* helper: its fallback arm is `other => other`, so it
-/// passes an unrecognised byte through unchanged. Routing through it made this
-/// function total — it could not return `None`, so the `?` refusals at the call
-/// sites were unreachable, and a byte that is not a base would have been
-/// "complemented" to itself and folded into an inversion instead of declining
-/// the edit. Enumerating the set is what makes that refusal real.
-fn complement_base(base: u8) -> Option<u8> {
-    Some(match base.to_ascii_uppercase() {
-        b'A' => b'T',
-        b'T' | b'U' => b'A',
-        b'G' => b'C',
-        b'C' => b'G',
-        b'R' => b'Y',
-        b'Y' => b'R',
-        b'S' => b'S',
-        b'W' => b'W',
-        b'K' => b'M',
-        b'M' => b'K',
-        b'B' => b'V',
-        b'V' => b'B',
-        b'D' => b'H',
-        b'H' => b'D',
-        b'N' => b'N',
-        _ => return None,
-    })
 }
 
 /// Trim the common prefix and suffix of the reference and result blocks.
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -8416,15 +8416,24 @@ impl<P: ReferenceProvider> Normalizer<P> {
                             ));
                         }
                         // Not a base we can complement into a typed
-                        // substitution. Unreachable as written:
-                        // `shorten_inversion` yields a one-base residue only
-                        // for a base whose complement differs from it, which
-                        // `complement` answers only for the IUPAC alphabet
-                        // `Base` models — a byte outside it is left unchanged,
-                        // reads as self-complementary, and collapses to
-                        // identity before reaching here. Kept as a defensive
-                        // fallback: leave the inversion as authored rather than
-                        // assert or invent an edit.
+                        // substitution. **Reachable since #1318**, and the
+                        // comment here used to say the opposite.
+                        //
+                        // The old justification was that an unmodelled byte was
+                        // returned unchanged by `complement`, so it read as
+                        // self-complementary and collapsed to identity before
+                        // reaching here. `complement_base` answers `None` for
+                        // such a byte instead, so `is_self_complementary` is
+                        // now `false` for it and `shorten_inversion` hands back
+                        // a one-base residue rather than `None` — measured:
+                        // `shorten_inversion(b"X", 0, 1)` was `None`, is now
+                        // `Some((0, 1))`.
+                        //
+                        // Landing here is the right outcome: leave the
+                        // inversion as authored rather than claim an identity
+                        // for a byte we cannot complement, which is the #1249
+                        // class of silent loss. Emitting nothing typed is a
+                        // refusal, not a fallback nobody reaches.
                         return Ok((start, end, canonicalize_edit(edit), warnings));
                     }
                     return Ok((
--- a/src/normalize/rules.rs
+++ b/src/normalize/rules.rs
@@ -23,6 +23,7 @@ use crate::coords::{hgvs_pos_to_index, index_to_hgvs_pos};
 use crate::error::FerroError;
 use crate::hgvs::edit::{InsertedPart, InsertedSequence, NaEdit, Sequence};
 use crate::reference::ReferenceProvider;
+use crate::sequence::complement_base;
 
 /// Check if an edit type needs normalization
 ///
@@ -114,15 +115,21 @@ pub fn insertion_is_duplication(ref_seq: &[u8], pos: u64, inserted_seq: &[u8]) -
 
     // Check if sequence before the insertion point matches the inserted sequence
     // (this would be a duplication of the preceding sequence)
+    //
+    // Case-insensitive on both sides (#1318): `ref_seq` may be soft-masked while
+    // `inserted_seq` comes from the author's upper-case description, and a raw
+    // byte test then reports "not a duplication" for one that is — the same
+    // class of miss the delins path carried, measured on `acgtacgt` + `ACGT`.
     if pos_idx >= ins_len && pos_idx <= ref_seq.len() {
         let before_start = pos_idx - ins_len;
-        if ref_seq[before_start..pos_idx] == inserted_seq[..] {
+        if ref_seq[before_start..pos_idx].eq_ignore_ascii_case(inserted_seq) {
             return true;
         }
     }
 
     // Also check if sequence after matches (for 5' duplications)
-    if pos_idx + ins_len <= ref_seq.len() && ref_seq[pos_idx..pos_idx + ins_len] == inserted_seq[..]
+    if pos_idx + ins_len <= ref_seq.len()
+        && ref_seq[pos_idx..pos_idx + ins_len].eq_ignore_ascii_case(inserted_seq)
     {
         return true;
     }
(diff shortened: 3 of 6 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #1295, opened 2026-08-01
Tests: one cis sweep is 94% of the local suite's wall clock (76.9s of 82.1s)


nh13:
`cargo nextest run --features dev -E 'not test(proptest)'` runs 8,879 tests in **82.1s**, of which `cis_junction_crossing_shift::no_two_member_allele_normalizes_to_a_different_sequence` alone accounts for **76.9s** — the only test nextest flags `SLOW [> 60.000s]`. The local suite is effectively serialized behind one test.

```
SLOW [> 60.000s] (─────────) ferro-hgvs::it cis_junction_crossing_shift::no_two_member_allele_normalizes_to_a_different_sequence
PASS [  76.898s] (8879/8879) ferro-hgvs::it cis_junction_crossing_shift::no_two_member_allele_normalizes_to_a_different_sequence
────────────
Summary [  82.089s] 8879 tests run: 8879 passed (1 slow), 155 skipped
```

The sweep enumerates 276,480 cases: 96 sequences × first_start 2..=13 × first_len 1..=2 × 3 first-member spellings (`del`/`dup`/`ins`) × downstream second_start × 2 second-member spellings × 2 shuffle directions.

This is deliberate, and the exhaustiveness is load-bearing. The sweep's own comment argues against narrowing it, correctly: skipping the dup+del+5' shape alone would drop roughly a twelfth of the sweep out of the sequence-preservation check to protect 74 known failures, and a new defect anywhere in that shape would then pass silently. `KNOWN_FIVE_PRIME_DUP_DEL_RESIDUAL` exists precisely so the class stays measured rather than merely tolerated. CI also shards the `Test` job 6 ways, so this does not gate CI wall-clock.

What is unaddressed is the **local** edit-test loop: any `cargo nextest run` over the `it` binary now costs ~80s regardless of what changed, which is a poor feedback loop for unrelated work.

Options worth considering, none obviously right:

- Mark the sweep `#[ignore]` and run it in the existing idempotency-soak job, so `cargo nextest run` is fast by default and the sweep still runs on every PR. Costs: it stops being part of a plain local run, which is exactly the property that made it catch things.
[...]

PR #1337, opened 2026-08-02
test(normalize): widen the sibling-crossing sweep to dup/ins siblings and a tandem tract

nh13:
Addresses #1283 (gaps 1 and 2 of 5). Does **not** close it — see "What is left" below.

## Gap 1 — `dup`/`ins` as the second member

The `second` loop was only `{del, sub}`, so in ~276,000 cases none ever placed a `dup` or an `ins` downstream of the first member. #1279's shape was unreachable while the file read as exhaustive.

Both are added. **Result: zero new sequence-changing or overlapping cases.** That is a real negative result, not a non-event: the class is correct on this corpus, and it is now measured rather than assumed.

## Gap 2 — a non-homopolymer tandem tract

The sweep's corpus is 20 bp of `{A,T}` / `{A,C,G,T}` noise. It cannot reach this class for two structural reasons, neither fixable by drawing more sequences:

- no tandem tract long enough to canonicalise to a **non-homopolymer** `Repeat`;
- insertion payloads are always single-copy, and a single copy inside a tract canonicalises to a `dup`, so the multi-copy repeat path is never entered.

`no_tandem_tract_allele_normalizes_to_a_different_sequence` adds a 10-unit `(CAG)` tract and inserts 1..=3 copies at **every rotation** (`CAG`, `AGC`, `GCA`) and **every junction** across it, against siblings 5' of, 3' of, and *inside* the tract, in both shuffle directions, authored in both member orders. Each sibling position carries **all four edit shapes** (`sub`, `del`, `dup`, `ins`) — restricting it to `{sub, del}` would have reintroduced gap 1 inside the very test that closes gap 2. 12,528 cases, 4.0s.

### It found a live defect class

108 sequence-changing cases, all in one shape: under 5' shuffle an insertion's junction travels to the tract's 5' end and crosses an upstream sibling.

```
TEMPLATE:g.[16_17insCAG;15A>T]  [FivePrime]  ->  TEMPLATE:g.[11_13dup;15A>T]
want  TTTTTTTTTT CAGCTG CAG CAGCAGCAG…
got   TTTTTTTTTT CAGCAG CTG CAGCAGCAG…     ← the substituted base moved
```

This is the **insertion half of #1267** — already filed, already pinned by `a_five_prime_insertion_junction_with_an_upstream_sibling_is_a_known_gap`, and not a local fix: #1259 measured that a 5' mirror of the junction clamp turns 80 previously-correct outputs silently wrong. So it is **counted, not excluded**, pinned as `TRACT_FIVE_PRIME_INSERTION_SEQUENCE_CHANGES = 108`.

The count is attributable rather than opaque: it breaks down by the sibling's **position**, not its edit type — `{pos15: 48 sub + 48 dup, pos36: 6 del + 6 ins}` — which is exactly what a junction-crossing defect predicts, since the junction travels past a *position* and the sibling's shape cannot change how many cases it corrupts. The two siblings outside the tract contribute **zero**, because the junction cannot travel out of the tract to reach them. Adding the `dup`/`ins` sibling shapes therefore doubled the known class rather than exposing a new one — `changed` is still empty.

Counting rather than skipping follows the treatment this file already gives its `dup` + `del` + 5' residual, and for the same reason: excluding a shape drops it out of the check entirely, so a *new* defect landing in it would pass silently. A rise is a new defect; a fall means #1267's insertion half is being fixed, in which case this number and the sibling test move together.

**Nothing changed outside that shape** — `changed` is empty.

## Runtime is held flat

Widening the second member from 2 shapes to 4 doubles the work per sequence, and this is the suite's slowest test: measured **45s → 117s** at the original 48 seeds. The seed count therefore halves to 24, putting case count and runtime back where they were (**45.9s**) while covering twice the shapes.

That is the right trade here rather than a coverage loss to apologise for: all five gaps the issue names are *shape* gaps, and every blocking defect found so far lived in a shape the generator could not emit rather than in a sequence it did not draw. The sweep's own `checked > 100_000` floor still holds with room to spare.

(This is the pressure #1295's seed knob exists to relieve. Not implemented here — it is scoped to the concurrent three-member work.)

## What is left, with reasons rather than deferrals

- **Gap 3 (three members)** is #1268, owned by the concurrent rewrite work. Worth stating explicitly because the two issues are easy to conflate: **#1268 and #1283 both name the three-member gap, and closing #1268 only dents #1283.** The other four blind spots survive it. #1283 should not be closed by association.
- **Gap 5 (`c.`/`n.`)** is blocked, and the block was *measured*, not assumed: the #1284 reproducer still panics the run under `FERRO_ASSERT_REPARSE=1`, so a transcript-axis slice would **abort CI** rather than fail with a diff. #1284 is open — its merged PR #1315 extended some passes to `Tx`, but `c.`/`r.` remain gated (`merge.rs` still has two `Genome | Mt` gates).
- **Gap 4 (out-of-phase payloads)** is partly addressed: the tract sweep inserts every rotation of the unit, which is the out-of-phase shape. The multi-copy half is covered (1..=3 copies); what is not reached is an out-of-phase payload against a *3-member* allele, which is gap 3's territory.

## Verification

Generator changes only — no new assertions. The existing SPDI apply oracle already decides right from wrong, which is why this survives the sequence-first rewrite.

Full suite with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1`: **9019 passed, 0 failed**. `cargo fmt --all --check` and `cargo clippy --features dev --all-targets` clean.

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for insertion-junction scenarios, including tandem repeats, rotated payloads, duplications, insertions, sibling positions, edit types, shuffle directions, and member ordering.
  * Added validation for sequence-changing, overlapping, and unconvertible outputs.
  * Added tracked baseline coverage for five-prime insertion-junction cases.
  * Improved exhaustive test sweeps while reducing redundant test combinations and separately monitoring known edge-case behavior.

---

## Review round 2

**The pinned residual is now bucketed by sibling position, not summed.** The doc claimed a `{pos15, pos36}` breakdown that a single counter could not enforce: a defect adding cases at `inside_5p` while removing the same number at `inside_3p` kept the total at 108 and passed. `TRACT_FIVE_PRIME_INSERTION_SEQUENCE_CHANGES_BY_POSITION` replaces the scalar and the whole map is compared, so the documented invariant is executable. It also closes a second hole: a sequence change at an *unlisted* position now fails on the extra key, which matters because at the newly-swept junction 10 the outside-5' sibling at position 9 satisfies `sibling_pos < junction` and so falls inside `known_1267_shape` — a scalar would have absorbed it as part of the known gap.

**Both tract boundary junctions are now swept.** `tract_start..tract_end` yielded junctions 11..=39, so the payload was always inserted strictly *inside* the tract while the doc claimed "every junction across the tract". `10_11` and `40_41` are exactly the two junctions where an insertion can shift out of the tract entirely — the shape closest to #1267 — and they were never emitted. Range is now `tract_start - 1 ..= tract_end`: 31 junctions, 13,392 cases (floor raised 10,000 → 11,000 to keep the stated ~20% margin).

That moved the residual from `{15: 96, 36: 12}` to `{15: 100, 36: 16}`, and the delta is fully attributed rather than re-pinned on faith. Instrumenting the counter by `(junction, sibling_pos, sibling)` shows every one of the new cases comes from **junction 40** alone; junction 10 contributes zero. The structure is `<junctions 3' of the sibling> × <2 siblings at that position> × <2 member orders>` — position 15 from junctions 16..=40 is `25 × 2 × 2 = 100`, position 36 from junctions 37..=40 is `4 × 2 × 2 = 16` — and exactly one of the nine `(rotation, copies)` combinations corrupts per junction, the out-of-phase one. `changed` and `overlapping` both stayed empty, so this is more of the known #1267 shape, not a new defect.

**Two `ins` shapes were sequence-identical to `dup` shapes.** Inserting the reference base immediately 5' of a junction denotes the same edit as a `dup` at that position, so both sweeps read as covering four sibling shapes while carrying three — the gap-1 defect this PR exists to remove, reintroduced in the fix for it.

- Tract sweep: `{inside_3p}_{inside_3p+1}ins{base_at(inside_3p)}` → `insTT`. `T` is absent from a `(CAG)` tract, so it stays an `Insertion`. Count-neutral, verified by the per-junction attribution above.
- Two-member sweep: the payload now excludes the bases on **both** sides of the junction, not just the 5' one — matching the 3' base would denote `{second_start + 1}dup` instead. `FIVE_PRIME_DUP_DEL_SEQUENCE_CHANGES` is still `0` and `changed` is still empty with the stronger payload.

`firsts` builds its own `ins` payload the dup-equivalent way when `first_len == 1`. That predates this PR and changing the *first* member perturbs the primary sweep far more than changing the second, so it is filed as #1342 with its own before/after measurement rather than widened into this one.

**No movement — this PR changes no production code.** The diff is confined to `tests/it/cis_junction_crossing_shift.rs`; ferro's output for every input is byte-identical. What changes is what the sweep can *see*: 13,392 tract cases against 12,528, four sibling shapes that are genuinely four rather than three, and a residual pinned per position instead of in aggregate. The oracle here is the SPDI apply oracle already in the file, and it is unchanged.

### DIFF
--- a/tests/it/cis_junction_crossing_shift.rs
+++ b/tests/it/cis_junction_crossing_shift.rs
@@ -36,6 +36,8 @@
 //! repeats too, and the junction is then bounded at the sibling's 5' edge —
 //! still flush against it, so the #999 collapse keeps firing.
 
+use std::collections::BTreeMap;
+
 use crate::common::cis_apply_oracle::{
     apply, assert_normalizes_preserving, assert_normalizes_preserving_in, normalize, normalize_in,
     sweep_sequences,
@@ -58,6 +60,47 @@ const RUN: &str = "ACAAAAAAAACGTACGTACG";
 /// own shift subject to the clamp. A rise means a new sequence-changing defect
 /// in that shape specifically.
 const FIVE_PRIME_DUP_DEL_SEQUENCE_CHANGES: usize = 0;
+
+/// Sequence-changing cases the tandem-tract sweep finds in the 5'-shuffle
+/// insertion-junction shape — the insertion half of #1267, which
+/// `blocks_sibling_shift` deliberately does not reach.
+///
+/// **Not zero.** This is a live, tracked defect, pinned here as an exact count
+/// so the shape stays inside the sweep rather than being excluded from it: a
+/// *new* defect landing in the same shape would raise the number and fail.
+/// `a_five_prime_insertion_junction_with_an_upstream_sibling_is_a_known_gap`
+/// characterises one case by hand on a homopolymer; this counts the whole shape
+/// over a non-homopolymer `(CAG)` tract.
+///
+/// The count breaks down by the sibling's **position**, not its edit type,
+/// which is what a junction-crossing defect predicts: the junction travels past
+/// a *position*, so the sibling's shape does not change how many cases it
+/// corrupts. The two sibling *positions* outside the tract (9 and 42, four
+/// shapes each) contribute **zero**, because the junction cannot travel out of
+/// the tract to reach them — which is why neither appears as a key.
+///
+/// Pinned **per position** rather than as a total, and the map is compared
+/// whole. That buys two things a scalar cannot:
+///
+/// * a defect that adds cases at `inside_5p` while removing the same number at
+///   `inside_3p` no longer passes with the total unchanged — the very asymmetry
+///   the paragraph above tells the reader to treat as a new defect, now
+///   executable rather than merely documented; and
+/// * a sequence change at an *unlisted* position fails on the extra key. That
+///   matters since the sweep reaches the boundary junctions: at junction 10 the
+///   outside-5' sibling at 9 satisfies `sibling_pos < junction`, so it is inside
+///   the `known_1267_shape` predicate and a scalar would have absorbed it as
+///   part of the known gap instead of surfacing it.
+///
+/// The numbers are fully accounted for, which is what makes a move readable:
+/// each is `<junctions 3' of the sibling> x <2 siblings at that position> x
+/// <2 member orders>`. Position 15 is reached from junctions 16..=40, so
+/// `25 x 2 x 2 = 100`; position 36 from junctions 37..=40, so `4 x 2 x 2 = 16`.
+/// Exactly one of the nine `(rotation, copies)` payload combinations corrupts
+/// per junction — the out-of-phase one — which is why rotation and copy count
+/// do not appear as factors.
+const TRACT_FIVE_PRIME_INSERTION_SEQUENCE_CHANGES_BY_POSITION: [(usize, usize); 2] =
+    [(15, 100), (36, 16)];
 
 #[test]
 fn an_insertion_does_not_shift_past_a_substituted_base() {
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1338, opened 2026-08-03
fix(normalize): repair del+dup collisions on the CDS-relative axes

### LINKED ISSUE #1284: Normalizer: del+dup collision repair is genomic-only, so c./n./r. still emit descriptions ferro cannot re-parse

nh13:
Closes #1284.

## What was wrong

`respell_colliding_duplications` and the two sibling repairs behind the same gate read the duplicated bases *over each member's own coordinates*. That is sound only where the axis is a direct offset into the sequence the provider serves — `g.`, `m.`, and since #1315 `n.`. `c.` and `r.` are CDS-relative, so they were refused, and a del+dup pair that 3'-shifted onto one position stayed collided:

```
NM_TEST.1:c.[-8del;-6dup]  ->  NM_TEST.1:c.[-1del;-1dup]
parse_hgvs(out) -> Err: Self-cancelling allele … overlapping reference positions
```

Because `FERRO_ASSERT_REPARSE` is armed in `ci.yml` and `nightly-mutalyzer.yml`, that is an **abort, not a diff** — which is what blocked #1283's transcript-axis sweep (its gap 5).

## The conversion

Key the offset on the member's **`Region`**, not on the axis kind.

The single `cds_start - 1` the issue records as tried-and-reverted is right at `c.1` and off by one below it, because the CDS axis has no zero. Patching that with a sign test still leaves the 3'UTR — a third offset, `cds_end` — wrong. Per region, each of `FivePrimeUtr` / `Cds` / `ThreePrimeUtr` is individually affine onto the transcript, so the discontinuity **disappears rather than being patched across**, and every pass already groups its members within one region.

| region | axis → transcript | delta |
|---|---|---|
| `Genome`, `Tx` | the axis is the sequence | `0` |
| `Cds`, `Rna` (`c.N`) | `cds_start + N - 1` | `cds_start - 1` |
| `FivePrimeUtr` (`c.-N`) | `cds_start - N` | `cds_start` |
| `ThreePrimeUtr` (`c.*N`) | `cds_end + N` | `cds_end` |

With that in place the axis gate is not widened but **removed from all three passes**: a region the conversion cannot place is skipped, so the conversion is the gate. Per the issue's own appendix, one conversion unblocks all three.

Two things the conversion alone did not cover, both found by measuring rather than reasoning:

- **A junction can land on a region boundary where its span does not.** A duplication resting at `c.-1` lands its copy between `c.-1` and `c.1`, and `-1 + 1` is `c.0`, which does not exist. `respell_at_gap` now resolves both ends through the transcript and names it `c.-1_1`.
- **`respell_at_gap` converted the gap to `u64` up front**, a leftover from when it was genomic-only. On the 5'UTR the axis is negative by definition, so that guard silently refused every repair there before any arm ran. This is why the first cut of the fix moved the outputs but repaired nothing.

## Measured

- **Re-parse census:** on a 3950-case two-member grid over both UTRs and the CDS body, on both axes — **120 outputs failed to re-parse before, 0 after**, every one the same `Self-cancelling` collision. (A first grid whose 5'UTR was one homopolymer put 296 of 3080 in the same single shape.) The CDS body contributed **none** either way: there the sequence-first canonicalization resolves the collision first, which is why the defect was confined to the UTRs.
- **Bucket movement: none.** `status_census_is_unchanged`, `passing_census_is_unchanged` and `divergence_budget_is_unchanged` all hold. Worth saying why rather than reading it as "no effect": the spec corpus has no two-member UTR allele, so it cannot exercise this change — the new tests are what carry it.
- Full suite green with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1` (9028 tests), clippy `-D warnings`, `cargo check --features python`, no new proptest seeds.
- All new tests are reference-free (`MockProvider` via `SyntheticBuilder`), so they **do** run on PR CI — this is not manifest-gated coverage.

## Two pinned strings move, deliberately

`issue_1235_transcript_axes::overlap_conflicting_allele_is_not_canonicalized` and `issue_1276_dup_junction_overlap::lenient_output_of_a_conflict_is_still_rejected_by_strict` pinned `c.[5_9inv;5_6dup]` — the *un-repaired* transcript-axis form. It is now `c.[5_9inv;6_7insAA]`, which is what the `g.` spelling of the same shape already produced (`g.[261_265inv;262_263insAA]`, measured). So the change **removes an axis divergence rather than creating one**.

Everything those tests actually guard was verified unchanged before re-blessing: strict still rejects the output as `W5002 / OverlapConflictingEdits`, the authored member order still survives, and the result is still idempotent in one pass. The `dup` spellings now *converge* onto the `ins` spelling instead of being a second fixed point — one variant, one canonical string. #1276's other tests author a `dup` directly and are untouched; they are what carry the #1243 dup-detector coverage, and the module header's now-false axis rationale is corrected.

## Found alongside, and deliberately not changed

- **`c.`/`r.` diverge on where a 5' shuffle stops.** `c.-8del` settles at `c.-1`, `r.-8del` at `r.1`, because only `normalize_cds` intersects the shuffle window with the CDS↔UTR axis bound. That is explicit (#334, on the grounds that `r.` natively spans the sub-axes, with a named pinned test), and it is about where the shuffle stops, not about the repair passes' coordinate conversion. Characterized by a new test so it is recorded rather than rediscovered. It also made one of my own `r.` assertions vacuous — on the issue's transcript the `r.` members shuffle out of the UTR entirely — so those cases now run on a transcript where they genuinely stay in the 5'UTR.
- **An inverted-CDS record (`cds_end < cds_start`) has an incoherent `c.` axis**, its UTR halves overlapping so one transcript position has two names. The new conversion declines rather than picking a spelling. This does **not** fix the ambiguity, which is reachable on `origin/main` without this change (`c.[*1del;*2dup]` comes back as `c.[-10del;-9dup]` there too, both naming transcript position 14).
- **The UTRs stay outside the sequence-first canonicalization.** That pass produces the better answer where it runs — the genomic analogue reduces to an identity — but it restricts itself to the positive body region, which is #920's deliberate scope, not #1284's. Widening it would move every UTR allele's canonical form and is a different change.

## Module doc corrected

`merge.rs` asserted that ranges spanning the UTR↔CDS boundary "do not exist as valid range syntax". They do: the spec writes `NM_001849.3:c.-1_*1=` (`consultation/SVD-WG001.md:37`) and `NM_004006.2:c.-244_*2691del` (`consultation/open-issues.md:53`), and ferro's parser accepts both. The pass's refusal to *merge* across regions is a property of its region-local arithmetic, which is what the comment now says.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HGVS normalization for ranges crossing CDS and UTR boundaries.
  * Enhanced transcript, RNA, CDS, and genomic coordinate handling for repeat, duplication, inversion, and gap repairs.
  * Corrected collision repairs so affected duplications are represented as insertions when appropriate.
  * Improved junction and boundary handling while preserving valid empty-insertion behavior.

* **Tests**
  * Added comprehensive coverage for transcript-axis normalization, reparsing, strand orientation, cross-region ranges, boundary cases, idempotency, and strict-mode validation.

---

## Review round 2

Fourteen findings across the two review lenses; thirteen applied, one rejected. The two that changed production behaviour are called out first.

### Guard parity in `region_sequence_delta` — the inverted-bounds refusal was only a third of a refusal

`ThreePrimeUtr` went through `ordered_cds_bounds` and declined on a record with `cds_end < cds_start`. `Cds`, `Rna` and `FivePrimeUtr` read `cds_start` alone and silently picked a spelling — so `inverted_cds_bounds_are_refused_not_guessed` was asserting a refusal that only one of the four regions actually made. All four now refuse.

Two details worth stating, because the obvious fix is wrong in both directions:

- Routing every region through `ordered_cds_bounds` would also demand that `cds_end` be *present*. A record carrying `cds_start` and no `cds_end` has an unknown 3'UTR but a perfectly well-defined `c.N` body, so the guard is written as "refuse if inverted", not "require both bounds".
- `Rna`'s fallback to a transcript-relative delta of `0` when there is no CDS is correct for a non-coding transcript, but `map_or(0, …)` on its own turns the inverted-bounds `None` into a silent delta of `0` — the exact guess the guard exists to prevent. Absent bounds now fall back; inverted bounds refuse. Pinned on the `r.` axis alongside the `c.` one.

`signed_cds_bound` lost its last caller and is removed.

### A collision on the final transcript base still names a position past the end

`c.[*10dup;*11dup]` 3'-shifts both members onto `*11`, the last base, and the repair anchors an insertion at the junction 3' of it — whose right endpoint is `*12`. `cds_axis_position` has a floor (`tx < 1`) and no ceiling, so it answers `ThreePrimeUtr(12)` instead of refusing, and the result renders as `c.*11_*12insAA`.

**Measured rather than argued.** Over this file's census:

| | outputs `parse_hgvs` rejects | positions past the transcript end |
|---|---|---|
| `origin/main` | 120 | 0 |
| this PR | 0 | 2 |

On `main` the same input came back as `c.[*11dup;*11dup]` — two members claiming one reference position, the #1234 overlap class — so both forms are defective. This change removes 120 unreadable outputs and converts one narrow defect into another narrow one on 2 inputs, both of which stay readable.

**Refusing is not available, and this was tested rather than assumed.** A bound on the right endpoint also refuses the *cancelling* terminal shape `c.[*10del;*11dup]`, which needs the repair: it returns to the unparseable `c.[*11del;*11dup]` (4 cases). `FERRO_ASSERT_REPARSE` is armed in `ci.yml` and `nightly-mutalyzer.yml`, so that aborts the run rather than failing a comparison — trading 2 readable-but-wrongly-anchored outputs for 4 that kill CI is not an improvement. An `NaEdit::Identity` exemption does not separate them either: both arrive at `respell_at_gap` as insertions, and the cancellation happens downstream.

The real fix is to spell a terminus-anchored insertion as the duplication it is (`c.*10_*11dup` is sequence-identical and needs no `*12`). That changes what `respell_at_gap` may emit rather than this axis conversion, so it is filed as **#1343** and pinned here by `a_collision_on_the_final_transcript_base_still_names_a_position_past_the_end`, with `a_cancelling_collision_on_the_final_transcript_base_is_repaired` as its counterweight — so the shape stays measured instead of excluded.

### Checked arithmetic at all three conversion sites

Three `i64` adds mixing a parsed position with a CDS-bound-derived `delta` were unchecked, while their siblings used `checked_add`. All three now decline instead of panicking in a debug build. Separately, `unsigned_gap` accepted a zero endpoint — `u64::try_from(0)` succeeds — so a junction at 0 would have named `g.0_1` / `m.0_1`; it now refuses, matching the `Tx` arm's refusal of `n.0`.

### Test strengthening

- **Idempotency is asserted, not delegated.** The census re-normalizes every output and compares byte-for-byte. Previously that signal existed only under `FERRO_ASSERT_IDEMPOTENT`, which only CI arms — a local `cargo nextest run --features dev` got none of it from the whole grid. An `Err` on the second pass is recorded as a failure rather than skipped: the first pass succeeded and its output re-parsed, so a re-normalization that now refuses is itself the defect.
- **Cross-region pairs.** The grid drew both coordinates from one region, leaving the whole cross-region class (`c.[-8del;1dup]` — the pairs whose two members take *different* `region_sequence_delta` values) outside a census that read as exhaustive. Now 12,110 cases, up from 3,950. Baseline is unchanged at 120 failures, so cross-region pairs contribute none — the expected answer, since a collision needs both members on one position, and that is now stated by the test rather than assumed.
- **Population floor raised** from `> 1000` to `> 11_000`, close to the measured 12,110. At the old floor nine tenths of the grid could have started erroring out of `normalize` before the test noticed, and the `continue` on `Err` made that silent.
- **The `g.`-axis parity is executable.** `issue_1235_transcript_axes.rs` justified its new `c.[5_9inv;6_7insAA]` pin by stating that the `g.` spelling of the same shape already settles as `g.[261_265inv;262_263insAA]`, "measured" — the independent path, since the repair was genomic-only before this change — but nothing asserted it. It now does, plus an explicit member-for-member cross-axis comparison, so the divergence this PR removes is pinned rather than described. (The genomic contig for `SyntheticBuilder::cds` is `chr_synth`/`TX_CONTIG`, not `GENOMIC_CONTIG`; under the wrong accession the description passes through unnormalized and the assertion would have been vacuous.)
- A failure message that described a decline while its assertion pinned a settled two-member form was corrected.

### Rejected

**"Use `CoordinateMapper`'s exon-aware mapping in `cds_axis_position` for gapped transcripts."** False positive. `cds_axis_position` maps a CDS-axis position to a **transcript** position, and exon structure does not appear in transcript coordinates — introns are exactly what the transcript sequence omits. `cds_start`/`cds_end` are themselves transcript coordinates, and the provider serves transcript bases for an `NM_`, so `cds_start + N - 1` is exact. Exon awareness is needed for c.→**genomic**, which this conversion is not. The repair also sets `offset = None` and never handles intronic members.

**Toward, and confined to a shape the oracles do not reach.** Every input that changes is a multi-member cis allele whose members collide after shifting — the mutalyzer / biocommons / hgvs-rs corpora carry no such input, which is why 120 unreadable outputs went unnoticed. For every single-member description, and every allele whose members do not collide, output is byte-identical: `already_correct_shapes_are_unchanged` and `the_conversion_is_strand_independent` pin that, and the full suite is green at **9,075 passed, 0 failed** with the spec-normalization and spec-enumeration censuses unchanged (`ferro_produces_the_form_the_spec_states`, `status_census_is_unchanged`, `passing_census_is_unchanged`). The one input class that moves *away* from ideal is the 2 terminal-base cases above, tracked as #1343 — and they replace outputs that were already defective in a different way.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -18,16 +18,25 @@ use crate::sequence::complement_base;
 
 /// Coordinate-system region used as the merge-eligibility key.
 ///
-/// Adjacency in the merge pass requires both ends to share a region.
-/// HGVS forbids ranges that span the 5'UTR↔CDS, CDS↔3'UTR (for `c.`/`r.`)
-/// or upstream↔transcript / transcript↔downstream (for `n.`) boundaries
-/// — `c.-1_1`, `c.40_*1`, etc. do not exist as valid range syntax — so
-/// we tag each position with its region and refuse to merge across.
+/// Adjacency in the merge pass requires both ends to share a region, so we
+/// tag each position with its region and refuse to merge across.
+///
 /// The integer `start`/`end` axis is region-local: positive for `Cds` /
 /// `ThreePrimeUtr` / `Tx` / `TxDownstream`, negative for `FivePrimeUtr`
 /// / `TxUpstream`. Adjacency `prev.end + 1 == next.start` works
 /// naturally for all six because the axis is monotonic 5'→3' within
 /// each region (`c.-3 → c.-2 → c.-1` maps to `-3 → -2 → -1`).
+///
+/// That refusal is therefore a **property of this pass's arithmetic, not of
+/// HGVS**: `prev.end + 1 == next.start` is only meaningful within one region,
+/// since `c.-1` and `c.1` are adjacent nucleotides whose axis values differ by
+/// 2. A region-spanning range is perfectly valid nomenclature — the spec writes
+/// `NM_001849.3:c.-1_*1=` (`consultation/SVD-WG001.md:37`) and
+/// `NM_004006.2:c.-244_*2691del` (`consultation/open-issues.md:53`), and
+/// ferro's own parser accepts both. (This comment previously asserted the
+/// opposite, that `c.-1_1` "does not exist as valid range syntax"; corrected
+/// 2026-08-02, and [`respell_at_gap`] now deliberately *emits* such a range
+/// when a repair's junction lands on the CDS start.)
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub(crate) enum Region {
     /// `g.` (genomic) and `m.` (mitochondrial) — single axis, no
@@ -1992,9 +2001,16 @@ fn edit_span_union(edits: &[GEdit]) -> Option<(i64, i64)> {
 /// The two facts travel together because both are answered by the same
 /// `cds_start` lookup, and a second provider round-trip to re-answer the other
 /// half would be pure waste on a path that already fetches a window.
+/// Answers per *axis*, which is enough for its one caller because
+/// [`canonicalize_from_sequence`] restricts itself to the positive body region.
+/// The repair passes are not so restricted and must use
+/// [`region_sequence_delta`], which answers per *region* — see its doc comment
+/// for why a per-axis offset cannot be right in the UTRs.
 struct AxisFrame {
     /// Offset between the member axis and the fetched sequence: 0 for
     /// `g.`/`m.`/`n.`, `cds_start - 1` for the CDS-relative `c.`/`r.` axes.
+    ///
+    /// Correct only inside the positive body region; see the note above.
     delta: i64,
     /// Whether positions on this axis carry a **reading frame** — i.e. whether
     /// [`same_codon`] means anything here. True only when the axis is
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1339, opened 2026-08-03
fix(normalize): keep a junction re-spelling inside the sequence

### LINKED ISSUE #1327: normalize: junction re-spelling can place an edit past the end of a circular mitochondrial contig
`respell_at_gap` places a re-spelled member over the gap `[span.end, span.end + 1]`. On the mitochondrial axis that arithmetic is not wraparound-aware: a junction-occupying member sitting at the last base of the contig (16569 for rCRS) is re-spelled over `16569_16570`, a position the contig does not have, rather than the wrapped `16569_1` form the circular handling elsewhere uses (see `issue_129_mt_circular_wraparound`).

The `landed` guard inside `respell_at_gap` does not catch it — it only checks that the rewritten member reads back at the gap it was told to use, which an out-of-range gap does.

Both callers are exposed, since both admit `CisKind::Mt`:

- `respell_colliding_duplications` (`src/normalize/merge.rs`, guard at the top of the function)
- `coalesce_members_at_one_junction` (added in #1289, same guard)

This is pre-existing in the first caller, so it is not a regression from #1289 — filing it separately rather than widening that PR.

Suggested work:

- Decide the intended spelling for a junction at the contig's last base on a circular axis, and make `respell_at_gap` produce it (or decline) rather than emitting an out-of-range endpoint.
- Add an `m.` regression covering both callers, asserting byte-identical output and a re-parseable result, under `FERRO_ASSERT_IDEMPOTENT=1` and `FERRO_ASSERT_REPARSE=1`.

nh13:
Closes #1327.

> **Stacked on #1338** (issue #1284), which rewrote `respell_at_gap`. The guard added here sits inside that rewrite, so this PR's base should be #1338's branch — review or merge that first. The diff below is this commit only.

## What was wrong

`respell_at_gap` writes its member over the gap `[junction, junction + 1]`. A member that 3'-shifts onto the sequence's final base therefore lands one past the end — on rCRS that is `m.16569_16570`, and 16570 does not exist.

Neither existing guard sees it:

- `landed` only confirms the member reads back at the gap it was *told* to use, which an out-of-range gap does;
- `parse_hgvs` does not know the sequence's length, so `m.16569_16570insAA` re-parses cleanly and `FERRO_ASSERT_REPARSE` stays silent.

Both callers that admit `CisKind::Mt` were reachable. Measured:

```
m.[16566del;16567dup]               -> m.16569_16570=
m.[16566_16567insA;16567_16568insA] -> m.16569_16570insAA
```

## The fix: a boundary `delins`

Inserting `A'` immediately 3' of the last base **is** deleting that base and inserting `ref[last] ++ A'`. That identity is already the crate's answer to "an insertion saturated at a sequence bound" — `insertion_to_boundary_delins`, shared by #1170, #387, #1202, #1205 and #1217 — so this reuses it (raised to `pub(crate)`) rather than adding a second implementation. It is in range, valid on a circular or linear reference alike, and denotes exactly what the out-of-range insertion denoted.

```
m.[16566del;16567dup]               -> m.16568=
m.[16566_16567insA;16567_16568insA] -> m.16566_16569A[6]
```

The second is the nicest evidence the fix is right: two `A`s added to the terminal 4-`A` tract come back as the ordinary repeat spelling, i.e. the terminus stops being a special case in the output at all.

## Both alternatives were tried and refuted by measurement

- **The wrapped form** `m.16569_1ins` is not valid HGVS. SVD-WG006 authorises the reversed `<high>_<low>` range for deletions and duplications only (line 33; `NC_012920.1:m.16563_13del`, `J01749.1:o.4344_197dup`), and `insertion.md` is silent, so ferro's parser rejects it (#129). Every re-spelling here produces an insertion, so there is no wrapped form available. A test pins this, and will fail if `m.16569_1insA` ever becomes parseable — at which point the decision is worth revisiting.
- **Declining** — which the issue explicitly offers as an option — turned out to be *worse than the bug* for `coalesce_members_at_one_junction`. Refusing left its two members unmerged, and each had already been canonicalised to `m.16569dup`, giving `m.[16569dup;16569dup]`: an allele claiming one position twice. That is the #1286 defect that pass exists to remove, and it re-parses, so no oracle catches it. Refusing traded an out-of-range spelling for an overlapping one.

## Two things beyond the issue's scope, both found by measuring

- **A second producer.** `collapse_overlapping_cis_edits` reaches the same overrun by a different route: when a group nets to a pure insertion at the window's 3' edge, `del_start` is one past the last base and the insertion-shaped anchor renders `m.16569_16570=`. It does not go through `respell_at_gap` at all, so fixing only the function the issue names would have left the issue's own symptom in place. Refusing **is** the right answer there — it leaves the members in the in-range spelling they already had.
- **The transcript axes are not exempt.** `c.*11` on a 35-base transcript *is* the final base, so the same overrun happens there. An earlier draft of this fix refused on those axes, with a comment asserting no such case had been measured. The suite falsified that immediately: refusing left #1284's collision unrepaired and its output unparseable. They are handled now, and pinned by their own test so the transcript half cannot quietly regress to a refusal.

## Verification

Full suite green with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1` (9045 tests), `cargo clippy --features dev --all-targets -- -D warnings` clean, `cargo check --features python` clean, no new proptest seeds, and the enumeration / normalization censuses unmoved.

New tests are reference-free (`MockProvider`), so they run on PR CI rather than only in the nightly. Note the two oracles cannot catch this defect class on their own — an out-of-range coordinate re-parses and re-normalizes fine — which is why the assertions read the coordinates back and check them against the sequence length directly.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved normalization of edits that extend beyond the end of a reference sequence.
  - Prevented generated coordinates from exceeding valid sequence boundaries.
  - Added support for terminal insertion repairs across genomic, mitochondrial, transcript, CDS, and RNA representations.
  - Preserved valid in-range repairs while rejecting unsupported wrapped insertion syntax.

- **Tests**
  - Added coverage for sequence-end handling, including mitochondrial and transcript cases.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -520,6 +520,23 @@ pub(crate) fn collapse_overlapping_cis_edits<P: ReferenceProvider>(
     } else {
         (del_start, w_lo + hi_ref as i64 - 1)
     };
+    // The collapse has its own way of naming a position past the end (#1327):
+    // when the group nets to a pure insertion at the window's 3' edge,
+    // `del_start` is one past the last base, and the insertion-shaped anchor
+    // `(del_start, del_start - 1)` renders as `m.16569_16570=` on rCRS.
+    //
+    // Distinct from the `respell_at_gap` overrun this issue names — a different
+    // pass, reached without it — but the same defect, so it is fixed here too
+    // rather than left for the symptom to survive the fix. Refusing the
+    // collapse is the right answer *here* (unlike in `respell_at_gap`, where it
+    // reintroduced #1286): the members are left as they are, which is in range,
+    // and refusing is this module's convention for a group it cannot place.
+    if let Ok(length) = provider.get_sequence_length(&accession) {
+        let past_end = |p: i64| u64::try_from(p + delta).is_ok_and(|v| v > length);
+        if past_end(a_start) || past_end(a_end) {
+            return variants;
+        }
+    }
     let anchor = Anchor {
         region: body,
         start: a_start,
@@ -3662,7 +3679,14 @@ pub(crate) fn demote_coincident_tract_repeats<P: ReferenceProvider>(
             let edit = NaEdit::Insertion {
                 sequence: InsertedSequence::Literal(Sequence::new(payload.clone())),
             };
-            respell_at_gap(&mut members[i], span, span.end, edit, provider);
+            respell_at_gap(
+                &mut members[i],
+                span,
+                span.end,
+                edit,
+                provider,
+                TerminalRespell::Allowed,
+            );
             members[i] != originals[slot]
         });
         if !rewritten {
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #1210, opened 2026-07-26
normalize: r.-axis repeat notation suppressed on pass 1 when a CDS-resident edit shifts into the 3'UTR (not idempotent)


nh13:
`normalize` is not idempotent on the `r.` axis when a CDS-resident edit reduces to an insertion that 3'-shifts **out of the CDS into the 3'UTR**: pass 1 suppresses repeat notation because the *input* was CDS-resident, and pass 2 — whose input is now UTR-resident — emits it.

Same failure *mechanism* as #1207 (the codon-frame gate's verdict is computed for the input span but applied to an output that lands somewhere else), but a different manifestation: #1208's boundary clamp does not catch it, because the insertion does not rest **at** the CDS boundary — it shifts clean past it.

Found by the widened idempotency fuzz in #1180, at ~150k cases (so roughly a 1-in-150k density — see "Impact on #1180" below).

## Repro

Fixture: `SyntheticBuilder::cds(&padded(core), PAD_OFFSET + 1, PAD_OFFSET + core.len(), Strand::Plus)` with `core = "CCCCAACACACGGGG"` — i.e. the CDS is exactly the core, with a 256-base `ACGT…` 5'UTR and 3'UTR. `r.15` is the last CDS base (`G`); the 3'UTR begins `A C G T …`.

```
input:  NM_TEST.1:r.15delinsgaa
pass 1: NM_TEST.1:r.*1_*2insaa
pass 2: NM_TEST.1:r.*1a[3]
```

Committed proptest seed (`tests/proptest-regressions/normalize_idempotency_proptest.txt`):

```
cc 8d1ceaa8eae1c6162271d431b2cb4395e0359a560661739dc5402acd0d9bdf33 # shrinks to case = Case { core: "CCCCAACACACGGGG", sys: RnaCodingPlus, hgvs: "NM_TEST.1:r.15delinsgaa" }
```

## Mechanism

`r.15delinsgaa` shares a `g` prefix with the reference `G` at `r.15`, so it reduces to an insertion of `aa` after `r.15`. That insertion 3'-shifts past `cds_end` and lands at `r.*1_*2`, where the reference `A` at `*1` plus the inserted `aa` makes an `a[3]` tract.

The repeat conversion happens in the `NaEdit::Insertion` arm of `normalize_na_edit` (`src/normalize/mod.rs` ~L7797):

```rust
rules::insertion_to_repeat(
    ref_seq,
    original_pos_idx,
[...]

PR #1340, opened 2026-08-03
fix(normalize): answer the del→repeat codon gate on the tract, not the input span

### LINKED ISSUE #1270: Normalizer: del→repeat answers the codon gate on the input span where ins→repeat uses the emitted tract (#1210 shape)
## What's wrong

The two repeat-canonicalisation paths answer the codon-frame gate against **different spans**.

The deletion path passes the *input* span's verdict (`src/normalize/mod.rs:8590`):

```rust
if let Some(rep) = rules::deletion_to_repeat(
    ref_seq,
    result.start as usize,
    result.end as usize,
    gate.input_span_is_coding(),
) {
```

The insertion path passes the tract-aware form instead (`mod.rs:8127`), with a comment explaining why:

> `#1210`: the verdict above is the INPUT span's; the bounds let the helper re-ask about the tract the repeat would actually occupy, which differs whenever the edit shifts across `cds_end`.

That is precisely the #1210 argument — the gate must be answered for the **emitted** span, not the input span — applied to one path and not the other.

## Why it plausibly matters

`repeated.md` restricts repeat notation in a coding context to units that are a multiple of three, and that restriction is about where the repeat *lands*. A `c.` deletion whose repeat tract crosses `cds_end` has an input span with one coding verdict and an emitted tract with another, so the del path can answer the wrong question in exactly the situation #1210 was filed about.

## Status: unverified

I have not constructed a case that reaches it, so this is an asymmetry with a plausible consequence rather than a demonstrated defect. It is filed because it was noticed while working elsewhere and would otherwise be lost, not because it is confirmed.

## Suggested first step

Try to construct the reachable case — a `c.` deletion inside a tandem tract that straddles `cds_end`, so `input_span_is_coding()` and the tract's own verdict disagree. If it is reachable, the fix is to give `deletion_to_repeat` the same `gate.cds_bounds()` treatment `insertion_to_repeat` already gets. If it is not reachable, a comment on the del path saying so is worth more than the current silence.

nh13:
Closes #1270.

## The issue filed this as unverified. It is reachable.

> "I have not constructed a case that reaches it, so this is an asymmetry with a plausible consequence rather than a demonstrated defect."

It is a demonstrated defect. On a 14-base `T` run spanning `cds_end`, measured on `origin/main`:

```
NM_TEST.1:c.5_6insTT  ->  NM_TEST.1:c.4_*5T[16]    tract-aware verdict: repeat emitted
NM_TEST.1:c.5_6del    ->  NM_TEST.1:c.*4_*5del     input-span verdict:  repeat suppressed
```

Same tract, same unit (`T`, length 1, not a multiple of three), opposite answers.

What makes it reachable is that the `c.`-axis shuffle does **not** clamp a deletion at `cds_end` here — `c.5_6del` shifts clean out of the CDS to `c.*4_*5`. So by the time the repeat would be spelled, the tract is not coding and `repeated.md`'s carve-out applies to it exactly as it applies to the insertion. Only the insertion path re-asked.

## The fix

The one the issue suggests: pass `gate.cds_bounds()` to `deletion_to_repeat`, and move its gate check to **after** `extend_tandem_tract` so there is a tract to ask about. The re-ask is the same three lines `insertion_to_repeat` already uses, on the same frames — 0-based `ref_seq` indices with an exclusive end, against 1-based inclusive CDS bounds. `None` keeps the caller's verdict verbatim, so genomic / `n.` callers and the helper's own unit tests are unaffected.

```
NM_TEST.1:c.5_6del  ->  NM_TEST.1:c.4_*5T[12]
```

which is the insertion's answer with the count going the other way: 14 reference copies less 2, against 14 plus 2.

## Tests

- **The disagreement itself**, asserted as a *relation* between the two paths rather than only as two pinned strings, so it keeps meaning if the canonical spelling ever moves.
- **A tract held wholly inside the CDS**, by moving only `cds_end` (15 → 24) and changing nothing else — it must stay gated. Without this the change would read as "the codon rule was removed" rather than "it is answered about the right span".
- A 3-base unit that was never gated, and a deletion outside any tract: both unchanged.
- **Idempotency of the repeat form**, which is the property #1210 was actually filed about — the input-span verdict is what made pass 1 and pass 2 disagree there.

Reference-free (`MockProvider` via `SyntheticBuilder`), so these run on PR CI rather than only in the manifest-backed nightly.

## Verification

Full suite green with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1` (9035 tests), `cargo clippy --features dev --all-targets -- -D warnings` clean, `cargo check --features python` clean, no new proptest seeds.

**Bucket movement: none.** `status_census_is_unchanged`, `passing_census_is_unchanged` and `divergence_budget_is_unchanged` all hold — the spec corpus has no coding deletion sitting in a tandem tract that crosses `cds_end`, so it cannot exercise this change and the new tests are what carry it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved deletion-to-repeat normalization for coding regions by evaluating the complete repeat tract.
  - Corrected codon-frame gating across plus- and minus-strand transcripts.
  - Preserved expected behavior for three-base repeats and non-repeat deletions.
  - Ensured equivalent deletion and insertion cases normalize consistently.

- **Tests**
  - Added integration coverage for repeat gating and normalization idempotence.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -9162,6 +9162,10 @@ impl<P: ReferenceProvider> Normalizer<P> {
                         result.start as usize,
                         result.end as usize,
                         gate.input_span_is_coding(),
+                        // #1270: as on the insertion path above, the verdict is
+                        // the INPUT span's; the bounds let the helper re-ask
+                        // about the tract the repeat would actually occupy.
+                        gate.cds_bounds(),
                     ) {
                         let bases: Option<Vec<Base>> = rep
                             .unit
--- a/src/normalize/rules.rs
+++ b/src/normalize/rules.rs
@@ -856,11 +856,34 @@ pub(crate) struct DelToRepeatResult {
 /// The bounds check rejects 0-length deletions (`del_start == del_end`).
 /// Callers passing insertion-point-shaped zero-width ranges should use
 /// `extend_tandem_tract` directly instead.
+///
+/// # The codon gate is answered for the tract, not the input (#1270)
+///
+/// A sixth condition applies on the `c.`/`r.` axes and is deliberately not in
+/// the list above, because it is about the *emitted* description rather than the
+/// trigger: `repeated.md` requires `unit_len % 3 == 0` for repeat notation in a
+/// coding context.
+///
+/// `is_coding` is the caller's verdict for the span the *input* deletion
+/// occupied. Where `cds_span` is supplied it takes precedence, and the gate is
+/// re-answered for the tandem tract this function discovers — the span the
+/// emitted repeat actually occupies. The two differ exactly when the tract runs
+/// past `cds_end`, whose 3'UTR part the carve-out exempts. Pass `None`
+/// (genomic, `n.`, or any caller with no CDS context) to keep `is_coding`
+/// verbatim.
+///
+/// This mirrors [`insertion_to_repeat`], which took the same argument for #1185
+/// and #1210; before #1270 only that path re-asked, so a `T` run spanning
+/// `cds_end` gave `c.5_6insTT` -> `c.4_*5T[16]` while the matching `c.5_6del`
+/// gave a plain `c.*4_*5del`. The gate is therefore evaluated *after*
+/// `extend_tandem_tract` rather than before it, since the tract is what it now
+/// asks about.
 pub(crate) fn deletion_to_repeat(
     ref_seq: &[u8],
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #1260, opened 2026-07-30
Normalizer: adjacent-gap insertions and the spanning delins spelling reach two different fixed points


nh13:
## What's wrong

Two spellings of one haplotype reach two different stable normal forms, so the normalized descriptor cannot be used as a key.

Reference: `("ACGT" x 64) + core + ("ACGT" x 64)` on `NC_TEST.1`, so `core[i]` is at 1-based position `257 + i`. With `core = "AAAAAA"` the poly-A run is **257–263** (seven bases — the padding resumes its `ACGT` phase with an `A`).

**core `AAAAAA`:**

| input | normalizes to |
| --- | --- |
| `NC_TEST.1:g.[258_259insC;259_260insC]` | `NC_TEST.1:g.[258_259insC;259_260insC]` (its own fixed point) |
| `NC_TEST.1:g.258_260delinsACACA` | `NC_TEST.1:g.259delinsCAC` |

Both denote the same bases — applied window `GTAACACAAAACGTACGTAC` for all four of input-a, output-a, input-b, output-b (20 bases from 0-based offset 254, computed by projecting each member through `spdi::hgvs_to_spdi` and applying the triples to the reference, independently of the normalizer).

## Severity

Non-confluence, not corruption. Both outputs denote the correct sequence; they are simply two canonical forms for one variant, which breaks use of the normalized string as a stable key — the failure mode #1229–#1235 track.

## Suspected mechanism — unverified

The re-derivation guard scores its decision against the *input's* column count (`changed_columns_of_pieces > changed_columns_of_edits`). The re-derived `259delinsCAC` costs 3 columns; the two-insertion spelling costs 2 (so the bound refuses) and the spanning spelling costs 5 (so it accepts). An input-relative bound cannot be confluent by construction.

**Scoping caveat:** those identifiers do not exist on `main`. `git grep` finds them only on `origin/fix/issue-1235-sequence-first-canonicalization` (#1237) and the branches stacked on it. The defect nonetheless reproduces without that code, so either the mechanism is wrong or equivalent logic exists on `main` under another name. Treat the mechanism as a hypothesis; the observations are solid.

Making the bound input-independent is a design change to the guard that protects the "stays delins" corpus (#1034, #1040, #182, #422), so this is not a local fix.

## Reproduction

[...]

Issue #1232, opened 2026-07-27
Normalizer: spanning delins not split across an unchanged interior base; non-confluent


msto:
> [!NOTE]
> Authorship: the content below was drafted by GitHub Copilot CLI (an AI coding
> agent) and filed via gh under @msto's account. It reflects the agent's
> analysis, not a statement authored by @msto.

## What's wrong

`Normalizer.normalize` keeps a single spanning `delins` that covers an **unchanged
interior nucleotide**, instead of splitting it into the deletion and the separated
substitution that the same sequence actually represents. The split encoding
normalizes to a different, stable form, so one variant has two normalized
representations.

Example (delete positions 35–37, keep 38 unchanged, substitute at 39):

```
TEMPLATE:g.35_39delinsTA      -> TEMPLATE:g.35_39delinsTA        # spanning delins (kept)
TEMPLATE:g.[35_37del;39T>A]   -> TEMPLATE:g.[35_37del;39T>A]     # del + separated sub (kept)
```

Both describe the identical resulting sequence and are each idempotent — **two
stable fixed points for one variant** (non-confluence). No error/warning.

### Why the split form is canonical

The deletion (35–37) and the substitution (39) are **separated by an unchanged
nucleotide** (38):

> two variants separated by one or more nucleotides should be described
> individually and **not** as a "delins".
>
> — `delins.md:17` (restated `delins.md:81`, `inversion.md:21`)

So the canonical form is `[35_37del;39T>A]`; the spanning `35_39delinsTA` merges
across the unchanged base 38 and is non-canonical. (This is the same
"don't span unchanged bases" principle cited in #1157 for the split direction —
ferro should not *retain* a spanning delins across an unchanged nucleotide any more
than it should *create* one.)

## Severity: medium

- **Silent** — no error/warning; both outputs are valid HGVS.
- **Breaks the canonical-form guarantee** — one variant, two stable normalized
  strings depending on input encoding.
- **Non-compliant canonical** — the spanning `delins` violates the
  "separated → individual" rule.
- **Corrupts key-based aggregation** — consumers keying on the normalized HGVS.g
  string over-split one variant. (`EquivalenceChecker` returns `SequenceMatch`.)

## Reproduction (verified on v0.11.0, commit 16b98af)

```python
import json, tempfile, warnings
import ferro_hgvs

[...]

PR #1341, opened 2026-08-03
feat(normalize): sequence-first cis-allele splitter, and the measurements that bound it

### LINKED ISSUE #1262: Normalizer: substitution + deletion and the spanning delins spelling reach two different fixed points

### LINKED ISSUE #1274: Normalizer: a cis insertion+deletion cancelling at the contig 3' end emits `=` at a position past the end
## What's wrong

When a cis allele's insertion and deletion cancel each other at the **3' end of the contig**, the emitted no-change description names a position one past the end.

```
contig ACGTACGTAA   (accession T, length 10)

in   T:g.[8_9insA;10del]
out  T:g.10_11=              position 11 does not exist
```

`parse_hgvs` accepts it, because it does not bounds-check coordinates against a reference — so neither `FERRO_ASSERT_REPARSE` nor `FERRO_ASSERT_IDEMPOTENT` catches it.

## Severity: low — the sequence is correct

This is a coordinate bug, not a sequence bug. Inserting an `A` into an A-run and deleting an `A` from the same run genuinely cancel, so `=` is the right *answer*; only the span it is stated over is wrong. Verified by projecting both sides through `spdi::hgvs_to_spdi` and applying the triples to the reference independently of the normalizer — input and output denote the same bases.

## Reproduction

```rust
use ferro_hgvs::{parse_hgvs, MockProvider, Normalizer};

provider.add_genomic_sequence("T", "ACGTACGTAA".to_string()); // length 10

let variant = parse_hgvs("T:g.[8_9insA;10del]").unwrap();
assert_eq!(normalizer.normalize(&variant).unwrap().to_string(), "T:g.10_11=");
//                                                                     ^^ past the end
```

Two more inputs reach the same output, so it is not specific to one spelling:

| contig (length) | input | output |
| --- | --- | --- |
| `ACGTACGTAA` (10) | `T:g.[8_9insA;10del]` | `T:g.10_11=` |
| `ACGTAAAAAA` (10) | `T:g.[5_6insA;10del]` | `T:g.10_11=` |
| `AAAAAAAAAA` (10) | `T:g.[2_3insA;10del]` | `T:g.10_11=` |

**Control — the same span is fine when the contig is longer**, which is what shows the span is computed without reference to the contig length rather than being wrong in itself:

| contig (length) | input | output |
| --- | --- | --- |
| `ACGTAAAAAAGCGC` (14) | `T:g.[5_6insA;10del]` | `T:g.10_11=` (in bounds) |

## Only visible once #1259 lands

On current `main` these same inputs emit `T:g.[10del;10dup]`, which **fails to re-parse** — that is #1263, and #1259 fixes it. So this is not a regression from #1259; it is the next layer down, exposed by that fix. On `fix/cis-sibling-crossing-normalization` the output becomes re-parseable but keeps the out-of-bounds coordinate.

## One cause already ruled out

The obvious suspect is `respell_at_gap` in `src/normalize/merge.rs`, which places a re-spelled member at the gap `[span.end, span.end + 1]` with no contig-length check. **It is not the cause here.** Instrumenting `respell_colliding_duplications` on these inputs shows it does see duplications at position 10, but the `collides` guard is false every time, so the re-spelling never runs:

```
DBG dup at 10..10 does NOT collide
```

The unbounded `[end, end + 1]` in `respell_at_gap` is still worth a bounds check on its own merits — it is latent rather than benign — but fixing it will not move these cases. The span here is produced somewhere in the insertion/deletion cancellation path instead.

## Suggested first step

Find where the cancelled pair's `=` span is chosen and clamp it to the contig, then assert the invariant generally: no emitted coordinate may exceed `get_sequence_length` for its accession. That last check is cheap and would have caught this class at the seam, in the same way `FERRO_ASSERT_REPARSE` catches #1263's.

nh13:
Groundwork for #1235's cis-allele canonical form: a sequence-first block splitter, the reproducer corpus that scores it, and the alignment work that #1260 needed. Everything here is measured; what was tried and did not survive measurement is recorded below rather than dropped, because several of them look clean enough to be re-proposed.

**This does not close #1262, and it does not enable the confluence property.** Both were attempted, both worked on the local suite, and both were reverted after the manifest-backed conformance axis caught a regression against Mutalyzer. What the axis then showed about *why* is the most important thing here — see *What was reverted* below.

## What lands

- **`AlignmentDag` and the dominator partition** (`src/normalize/seqfirst/`) — the DAG of every minimal alignment, the edges common to all of them, and a partition derived from those. Members come out disjoint by construction rather than being repaired afterwards. Verified against a brute-force oracle over an exhaustive sweep.
- **The two-gap insertion alignment** — a single gap structurally cannot express *insertion, retained reference, insertion*, which is #1260's shape. This adds it, and then generalises the search off the caller's trim so the retained chunk need not be the whole reference.
- **`partition_block`'s separation threshold and type guard** — a length-changing block splits at one unchanged base (`delins.md:17`), unless every resulting member would render at `delins` rank, in which case the spanning form stands. **#1260 is fixed**: both spellings converge on `g.[258_259insC;259_260insC]`.
- **The reproducer corpus** (`src/normalize/merge/splitter_reproducer_corpus.rs`) — 65 harvested blocks as executable rows, plus a collision census. This is the instrument every splitter change in this area should be scored against; four collisions that were open when it was written are now resolved at the splitter.
- **The `FERRO_SEQFIRST_SHADOW` audit** — runs both splitters and reports disagreements, so the migration can be measured before it decides anything.

## What was reverted, and why it matters

The full six-step build was implemented and green on the local suite (9094/9094). It then failed the conformance axis, which the local suite does not run:

```
NG_012337.1:g.[7109T>A;7110del]
  mutalyzer expects  NG_012337.1:g.7109_7110delinsA
  produced           NG_012337.1:g.[7109T>A;7111del]
```

Baseline is 183/183 green — re-verified by re-running the axis with `src/` and `tests/` checked out at the pre-change commit — so this was a regression introduced by the work, not a stale ledger. Bisected per commit, the cause is **giving back a flank base the trim had no grounds to take**, which is what #1262's block needs before any tie-break can reach it.

**#1262's premise is not in doubt — an earlier revision of this description said it was, and that was wrong.** Two corrections.

`delins.md:86-89`, in the Q&A rather than the Notes, is a worked example of the disputed shape: ref `..CAT`**G**`ACA..`, patient `..CAT`**A**`TA``ACA..`, asked as `c.[2077G>A;2077_2078insTA]`. The answer is `c.2077delinsATA`, and the NOTE records that the permission to spell it as two members *was removed*. A substitution plus an **adjacent** indel is one `delins`. That settles #7109 against us.

Mutalyzer was also under-sampled — one row generalised into "the oracle prefers merging". Across the 55 multi-member cis inputs in the corpus it does both, discriminated by adjacency:

```
NG_017013.2:g.[16985A>T;17011_17012del] -> [16985A>T;17013_17014del]   split, and 3'-shifted
NG_012337.1:g.[7109T>A;7110del]         -> g.7109_7110delinsA          merged
```

#1262's members are separated by an unchanged base, so `delins.md:17` applies and the split is right. #7109 is a **different** defect: reference bases at 7109..7113 are `T C C T T`, so the deleted `C` is ambiguous across 7110/7111; ferro 3'-shifted it out of adjacency and then applied `:17` to a separation it had manufactured.

Preferring the contiguous placement fixes that — re-measured with the flank restore, the axis is 183/183 and one corpus row moves. But it **still does not make #1262 converge**, which is the only reason to want the restore, so it costs three pinned expectations and buys nothing. Not adopted here.

**What the next attempt has to solve** is therefore not #1262's target form, but keeping a member's 3'-shift from breaking an adjacency that `delins.md:16` and `delins.md:86-89` require be described as one `delins`.

## Hypotheses refuted by measurement

Recorded so they are not re-proposed. Three of these looked clean before being run.

| hypothesis | verdict |
| --- | --- |
| **Dominance** — a match is real when present in every minimal alignment | **anti-correlated**: 6 of 8 rows backwards. #422 and #1040 must merge and have dominators; #1262, #1157A, #275 and #1232 must split and have none |
| **Supremal extent** — hand the partitioner the rolled-out window | regresses #1260 and does not fix #1262 |
| **Contiguity** (`delins.md:16` decides the tie) | provably a no-op on a trimmed block, so it fires only on untrimmed ones. Re-measured with the flank restore it is axis-green and moves one corpus row — but it does not make #1262 converge, so it buys nothing |
| **Prioritisation prefers a split** (`general.md:56`) | `delins` is not in that list at all, and the list ranks one description's type, never a multi-member allele against a spanning one |
| **Removing the input-separator veto** | costs #180 (`g.[1006_1008del;1009_1010insCCC]` collapses across the untouched 1009) and buys nothing measurable — #422's two-member spelling does not converge either way |

The contiguity rule's motivating worked example turned out to be fiction at the block level: the `repeat_span` rows it was said to fix reach `partition_block` as `TTT -> A`, for which every placement yields the identical single piece.

## Corrections to earlier work on this branch

- `general.md:56` was cited for the typed collapse in a commit message and in `split_buys_no_higher_priority_type`'s doc, claiming `delins` was "last in the priority order". It is not in the list. Both now state the rule as ferro policy inside a spec gap, licensed by `delins.md:44-47`'s analogy and the corpus.
- The corpus module doc had drifted from its own executable pins in four places, all understating how much is resolved. Each is now written as a gloss naming the pin, so the next drift is a test failure rather than a paragraph.
- `separations_are_meaningful`'s raised threshold was a bare `2` inline — a policy choice doing semantic work with no name and no doc. It is now `RAISED_PIECE_SEPARATION`.
- A unit test's comment claimed its placements did not tie. All three tie; the tie-break runs. The assertion was right and the name was wrong, so the name changed.
- A cost bound added during self-review was itself a defect: it clamped the two-gap search's *affix lengths*, which narrows the search rather than bounding work, and it excluded the sole decomposition of `AAAAAAA -> AACACAAAA` (`a = 2, b = 3`, needing a common suffix of 4). Invisible on this branch, because the single-gap search reaches the same member count there by another route. Replaced with a budget that declines the whole search instead of truncating it, pinned by a test asserting the exact columns.

## Two upstream spec defects found

Neither blocks this work.

- **`general.md:36-39` advertises a rejected proposal as forthcoming.** It says the SVD-WG "is preparing" a change and "the new recommendation **will be**: two variants separated by less than two nucleotides should be described as a 'delins'". That is SVD-WG010, and `consultation/SVD-WG010.md:5` reads `Status: rejected` (opened 2021-05-15, closed 2021-07-31). Ferro does not implement the advertised rule.
- **`delins.md:42`'s NOTE is attached to the wrong example.** It sits under `c.9002_9009delinsTTT` but discusses `c.[145C>T;147C>G]`, which belongs to the `c.145_147delinsTGG` example at `:37`.

## Verification

- `cargo nextest run --features dev` — **9093 passed, 0 failed**, 146 skipped. Log grepped for `SIGABRT` and `stack overflow`, not only `FAIL`: none.
- Manifest-backed with both oracles (`FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1`) — 9093 passed.
- `scripts/run_conformance_axis.sh` — **183/183**.
- `cargo fmt --all --check`, `cargo clippy --features dev --all-targets -- -D warnings` — clean.
- `tests/proptest-regressions/` unchanged — no seeds appended by any probe run.
- Anti-loss guard against the pre-branch test-name baseline: no tests lost, 7 gained.

## Reading order

`src/normalize/seqfirst/align.rs` and `partition.rs` first — they are self-contained and carry the design rationale. Then `partition_block` and `best_alignment` in `merge.rs`, then the corpus. The shadow audit in `canonicalize_from_sequence` is measurement scaffolding and decides nothing.

Relates to #1235. #1260 is fixed; #1262 remains open, with the finding above as the next thing to measure.

---

## Review round 2

Sixteen findings across the two lenses (nine from the GitHub review, seven from the CLI), all applied and squashed into their introducing commits. Grouped by what they change.

### Memory: the alignment DAG's adjacency

`AlignmentDag.out` was `Vec<Vec<(u32, u32, Step)>>`, allocated at `(n + 1) * (m + 1)`. That is 24 bytes of `Vec` header per grid cell whether the cell is on-path or not — about **400 MB before a single edge is stored** for the 4096 x 4096 block `benches/seqfirst_align.rs` measures — plus one allocator call per on-path cell.

A cell has at most three out-edges and the diagonal's kind is fixed by whether the two bases are equal, so four bits describe it exactly. `out` is now a flat `Vec<u8>` bitmask and `out_edges` reconstructs the tuples from `(i, j)` and the mask, in the same order the nested form pushed them. Same `Θ(n·m)` space, one allocation, order-of-magnitude smaller constant. `dominators` also stopped collecting `out_edges` into a `Vec` per cell — it counts, and only materialises the edge on the rare `crossing == 1` branch, which was putting the allocation straight back onto the sweep.

The exhaustive brute-force oracle (`dominators_agree_with_brute_force_on_exhaustive_small_inputs`) still passes, which is what makes this safe to do mechanically.

### Memory: an unbounded grid

Only the *reference* side of the sequence-first splitter is bounded upstream (`MAX_CANONICAL_WINDOW`, 4096). The alternate side is whatever the members produce, and a duplication over the window doubles it — 4096 x 8192 is 33.5 M cells across four grids. `partition_block_sequence_first` now refuses an oversized grid **before building it**, with a checked multiply (that product is itself an overflow site), returning `None` so the caller keeps the per-member result like every other cost bound here.

The budget is *derived* from `MAX_CANONICAL_WINDOW` rather than picked: a square grid at that bound. Pinned by `sequence_first_split_declines_an_oversized_grid`, which asserts the boundary from both sides so the test fails if the bound is removed **or** tightened enough to refuse ordinary blocks. (First attempt used `2^24` and refused the boundary square by 8,193 cells — the test caught it.)

### A panic path and a logging path

- `split_codon_incompatible_triplets` read `ref_bytes[piece.ref_start + 1]` and `+ 2` by raw index off a piece it does not own; `ref_end <= ref_bytes.len()` is now tested first, so an out-of-range span declines instead of panicking.
- The `FERRO_SEQFIRST_SHADOW` reports went to `eprintln!`, which fires once per canonicalized block — tens of thousands of lines over a suite run that no level or target filter could reach. Now `log::debug!` (the crate's unconditional facade; `tracing` is `web-service`-only), and the switch is documented as development-only with the `RUST_LOG` target to capture it.

### `dead_code`: the blanket was hiding exactly three items

`#![allow(dead_code)]` on `seqfirst/mod.rs` is gone. Removing it surfaced three items, each now scoped with a reason: `partition_members` (no production caller — the shadow takes the `partition_members_with` route that shares one `Dominators` sweep), and `AlignmentDag::edit_distance` + the `total` field it reads (benchmark-only, so `#[cfg_attr(not(feature = "dev"), allow(dead_code))]` — they warn again if `dev` stops keeping them alive). Verified clean with and without `--all-features`.

The module header also claimed the path was "unwired into `normalize_allele`". It is not — it is compiled in and run there as a shadow comparison behind `FERRO_SEQFIRST_SHADOW=1`; what remains is promoting it from shadow to authoritative. **The same wrong claim appeared in `src/normalize/mod.rs`**, which the CLI lens caught as the sibling site; both are corrected.

### Tests that could pass vacuously or tautologically

- `sequence_first_split_round_trips_every_block_it_accepts` counted nothing. Its `else { continue }` is silent, so a regression making `member_alt_spans` decline all 3,969 pairs would leave the round-trip assertion unexecuted and the test green. Now floors the accepted count, matching how the sibling sweep guards itself.
- `delins_bracketed_payload_expands` asserted `expanded == "GCGTTTGGGAAACC"` where `expanded` was built from the same literal three lines up — two literals compared to each other, unable to fail if the expansion regressed, while the comment above claimed it covered exactly that. Now checked against the normalizer's own output.
- `rendering_does_not_depend_on_the_input_spelling` asserted only `na == nb`. Two spellings converging on the same *wrong* form satisfies that. Rows 1 and 2 were anchored elsewhere; the third was anchored nowhere, so its expected answer — already stated in the module doc — is now pinned.
- `the_eight_spelling_pairs_still_diverge` had the count "eight" in three places' prose (its own name, the module doc, and `rewrite_target_corpus.rs`) and executable in none. `DIVERGENT.len()` is now asserted.
- `msto_regression_corpus`'s `state: &'static str` admitted any string, so a runtime check existed to catch `"Closed"`. A two-variant `State` enum moves that to compile time and the assertion is deleted rather than kept.

### Corpus and doc corrections

- The `COLLISIONS` census said "two recorded correct, four recorded as known defects"; the actual census is one `BothCorrect`, one `KnownDefect`, four `ResolvedToSplit` — which the module doc and the executable census test both already said.
- The untrimmed poly-A window note claimed the splitter disagreement was specific to `#1262-window` because its `#1260` sibling agreed. Both rows are `diverging(… 1, (2, …))` and both are in that group's membership assertion, so they diverge identically; corrected.
- `CONTRIBUTING.md`'s `cargo bench` silently skips `seqfirst_align`, which is `required-features = ["dev"]`.
- `benches/seqfirst_align.rs` used the deprecated `criterion::black_box`; now `std::hint::black_box`.

### #1274 is fixed, so its pin is inverted rather than deleted

Rebasing onto `main` after #1339 landed turned `a_cancelling_edit_at_the_contig_end_still_emits_an_out_of_bounds_span` red — which is this file's contract working as designed. #1327's terminal re-spelling closed #1274: `g.[8_9insA;10del]` on a 10-base contig now gives `g.[9dup;10del]` rather than naming position 11.

The assertion is inverted into `a_cancelling_edit_at_the_contig_end_stays_inside_the_contig` rather than deleted, so the reproducer that found the defect stays attached to a standing boundary-validity guard. It asserts **two** things, because the position bound alone is weak: the independent applier never caught this defect (an `=` edit's SPDI triple does not exercise the out-of-bounds check a deletion's does), so "stays in range" is paired with "still denotes the reference" — otherwise a fix that clamped the span while corrupting the sequence would pass.

**None — no output changes.** The sequence-first splitter remains shadow-only behind `FERRO_SEQFIRST_SHADOW=1` and does not affect what ferro emits; this round's production edits are a bounds check that only fires on a span no caller produces, a cost bound whose fallback is the existing per-member result, a data-structure change to a DAG whose exhaustive oracle is unchanged, and a logging facade swap. The one behavioural line that moved (`#1274`) moved in `main` via #1339, not here — this PR only re-points its guard at the fixed answer.

Full suite: **9,172 passed, 0 failed**. `cargo fmt --check` clean and `clippy --all-features --all-targets -D warnings` clean, plus a second clippy pass without `dev` to confirm the narrowed `dead_code` scoping holds in both feature sets.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,7 +74,22 @@ cargo nextest run --features dev          # All tests (preferred)
 cargo test --features dev                 # Alternative
 cargo nextest run -E 'test(test_name)'    # Specific test
 cargo test -- --nocapture                 # With output
-cargo bench                              # Benchmarks
+cargo bench --features dev               # Benchmarks (`dev` is required by seqfirst_align)
+```
+
+### Conformance axis (manifest-backed)
+
+`cargo nextest run --features dev -E 'test(axis_)'` selects the manifest-backed
+conformance tests, but each of those silently returns early (reported as
+PASSED, not skipped) when `FERRO_MANIFEST` is unset or points at a missing
+file — so running that filter bare can look like a clean pass while testing
+nothing. Use `scripts/run_conformance_axis.sh` instead: it validates the
+manifest *before* invoking nextest, so a missing one is a hard failure rather
+than a silent green. See `scripts/README.md` for details, including why the
+test count the script prints is a weaker signal than the manifest check.
+
+```bash
+FERRO_MANIFEST=/path/to/manifest.json scripts/run_conformance_axis.sh
 ```
 
 ## Code Style
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -339,6 +339,11 @@ harness = false
 name = "baseline_head"
 harness = false
 
+[[bench]]
+name = "seqfirst_align"
+harness = false
+required-features = ["dev"]
+
 [profile.release]
 lto = true
 codegen-units = 1
--- a/benches/seqfirst_align.rs
+++ b/benches/seqfirst_align.rs
@@ -1,8 +1,11 @@
 //! Sizing the sequence-first DP.
 //!
-//! `AlignmentDag::build` is `O(n*m)` in time and space. This measures where
-//! that stops being acceptable, so the block-size cap in the migration is
-//! chosen on evidence rather than picked.
+//! `AlignmentDag::build` is `O(n*m)` in time **and space**, but Criterion has
+//! no memory axis — this measures wall-clock time only. It shows where that
+//! grows too slow to be acceptable, so the block-size cap in the migration is
+//! chosen on time evidence; the memory dimension is not measured here.
+
+use std::hint::black_box;
 
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
 
@@ -32,9 +35,11 @@ fn bench_build(c: &mut Criterion) {
         group.throughput(Throughput::Elements((len * len) as u64));
         group.bench_with_input(BenchmarkId::from_parameter(len), &len, |b, _| {
             b.iter(|| {
-                let dag =
-                    ferro_hgvs::normalize::seqfirst::align::AlignmentDag::build(&reference, &alt);
-                criterion::black_box(dag.edit_distance())
+                let dag = ferro_hgvs::normalize::seqfirst::align::AlignmentDag::build(
+                    black_box(&reference),
+                    black_box(&alt),
+                );
+                black_box(dag.edit_distance())
             })
         });
     }
--- a/benches/seqfirst_align.rs
+++ b/benches/seqfirst_align.rs
@@ -1,0 +1,45 @@
+//! Sizing the sequence-first DP.
+//!
+//! `AlignmentDag::build` is `O(n*m)` in time and space. This measures where
+//! that stops being acceptable, so the block-size cap in the migration is
+//! chosen on evidence rather than picked.
+
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
+
+/// Deterministic pseudo-random bases; no RNG dependency, and reproducible
+/// across runs so successive measurements are comparable.
+fn bases(len: usize, seed: u64) -> Vec<u8> {
+    let mut state = seed | 1;
+    (0..len)
+        .map(|_| {
+            state = state
+                .wrapping_mul(6364136223846793005)
+                .wrapping_add(1442695040888963407);
+            b"ACGT"[(state >> 33) as usize % 4]
+        })
+        .collect()
+}
+
+fn bench_build(c: &mut Criterion) {
+    let mut group = c.benchmark_group("seqfirst_align_build");
+    for &len in &[16usize, 64, 256, 1024, 4096] {
+        let reference = bases(len, 1);
+        let mut alt = reference.clone();
+        // Perturb ~5% of positions so the DAG is non-trivial.
+        for k in (0..len).step_by(20) {
+            alt[k] = if alt[k] == b'A' { b'C' } else { b'A' };
+        }
+        group.throughput(Throughput::Elements((len * len) as u64));
+        group.bench_with_input(BenchmarkId::from_parameter(len), &len, |b, _| {
+            b.iter(|| {
+                let dag =
+                    ferro_hgvs::normalize::seqfirst::align::AlignmentDag::build(&reference, &alt);
+                criterion::black_box(dag.edit_distance())
+            })
+        });
+    }
+    group.finish();
+}
+
+criterion_group!(benches, bench_build);
+criterion_main!(benches);
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -11,6 +11,7 @@ Utility scripts for generating and updating test fixtures used by ferro-hgvs.
 | `fetch_ncbi_variation.py` | Ad-hoc live fetch of HGVS/SPDI conversions from the [NCBI Variation Services API](https://api.ncbi.nlm.nih.gov/variation/v0/). **Does not regenerate the committed fixture** (see note below). | `tests/fixtures/validation/ncbi_variation.live.json` (live; not committed) |
 | `fetch_variantvalidator.py` | Fetches validated HGVS variants from the [VariantValidator API](https://rest.variantvalidator.org) | `tests/fixtures/validation/variantvalidator_api.json` |
 | `benchmark_python.py` | Benchmarks ferro-hgvs Python bindings against biocommons/hgvs | N/A (prints results) |
+| `run_conformance_axis.sh` | Runs the manifest-backed conformance axis (`axis_*` tests) with a validated `FERRO_MANIFEST` | N/A (runs tests) |
 
 ## Requirements
 
@@ -52,4 +53,39 @@ python scripts/fetch_variantvalidator.py
 
 # Run Python binding benchmarks
 python scripts/benchmark_python.py --iterations 10000
+
+# Run the manifest-backed conformance axis (axis_ tests). Fails loudly
+# instead of silently passing if FERRO_MANIFEST is unset or missing --
+# see the script's header comment for why that matters.
+FERRO_MANIFEST=/path/to/manifest.json scripts/run_conformance_axis.sh
+# or:
+scripts/run_conformance_axis.sh /path/to/manifest.json
 ```
+
+### Conformance axis (`axis_*` tests)
+
+The conformance corpora (`tests/it/mutalyzer_normalize_tests.rs`,
+`tests/it/biocommons_normalize_tests.rs`,
+`tests/it/hgvs_rs_projection_tests.rs`) assert against a real prepared
+reference and are gated on the `FERRO_MANIFEST` environment variable: when it
+is unset, or points at a path that does not exist, each test's manifest-lookup
+helper returns early *without asserting anything*, and `cargo nextest` reports
+that as **PASSED**, not skipped. Running
+`cargo nextest run --features dev -E 'test(axis_)'` directly, without
+`FERRO_MANIFEST` set, is therefore a vacuous all-green run for exactly those
+tests, and looks identical to a real conformance pass.
+
+`scripts/run_conformance_axis.sh` closes that hole by validating the manifest
+path *before* invoking nextest, hard-failing if it is unset or missing. Build
+a manifest with `ferro prepare` (see `ferro prepare --help`); the manifest
+itself is never committed and this script never hardcodes a machine-local
+path.
+
+**What the printed test count does and does not prove.** `test(axis_)` is a
+substring match, so it selects ~179 tests — but only ~26 of them, the three
+corpora above, are manifest-gated. The rest (`cli::project::tests`,
+`hgvs::variant::tests`, `issue_337_cds_start_clamp`, `issue_1086_*`, …) merely
+contain `axis_` in their names and run identically with or without a manifest.
+They are harmless to run, but a non-zero count on its own is **not** evidence
+that conformance ran; it only shows the filter matched something. The
+pre-flight manifest check is the load-bearing guard.
--- a/scripts/run_conformance_axis.sh
+++ b/scripts/run_conformance_axis.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+# Runs the manifest-backed HGVS conformance axis (the `axis_*` test family
+# in tests/it/{mutalyzer_normalize,biocommons_normalize,hgvs_rs_projection,...}
+# _tests.rs) against a real ferro-prepared reference.
+#
+# Why this script exists: the manifest-backed conformance tests are gated at
+# runtime on `FERRO_MANIFEST` -- when it's unset, or points at a path that
+# doesn't exist, the test's own manifest-lookup helper returns `None` and the
+# test body returns early without asserting anything. `cargo nextest` reports
+# that early return as PASSED, not skipped. So `cargo nextest run --features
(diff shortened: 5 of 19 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #275, opened 2026-05-17
normalize: extend codon-frame exception beyond two-c.-SNV case (follow-up to #104)


nh13:
PR #104 landed the HGVS codon-frame exception for `c.` SNV pairs separated by one unchanged nucleotide (`#79`). Three sibling cases were flagged as out of scope and never filed.

## Source

PR #104 § *Out of scope*:

> Codon-frame exception for `r.` coding regions
> Chains involving more than two SNVs across a codon (`sub`, `ins`, `sub` straddling a codon)
> `sub`+`del` pairs separated by one unchanged nucleotide

## Items

1. **`r.` coding regions** — the exception currently applies to `c.` only. RNA coding-frame semantics are identical on a transcript axis; the rewrite path should branch through `r.` too.
2. **Three-or-more-SNV chains across a codon** — e.g. `sub`+`ins`+`sub` straddling codon boundaries. The current pair-detector does not generalize.
3. **`sub`+`del` pairs separated by one unchanged nucleotide** — the exception fires only for `sub`+`sub`. The same single-codon-frame argument applies when one of the edits is a `del`.

## Cross-ref

PR #104, #79 (codon-frame exception umbrella). Loosely fits #81 § A but not a current item there.

Issue #180, opened 2026-05-12
3′ shifting of deletions and duplications produced by allele normalization


msto:
## Context

The 3′ shift coverage matrices for `del` (A5, #106), `dup` (A6, #107), and `ins` (A1/A7, #93) are landed and working for variants expressed directly in those forms. However, when a deletion or duplication arises as the normalized result of merging individual operations within an allele bracket, the 3′ shift does not appear to be re-applied to the merged result.

## Spec basis

`general.md`: *"for all descriptions the **most 3' position** possible of the reference sequence is arbitrarily assigned to have been changed (**3'rule**)"*

This rule applies unconditionally to `del`, `dup`, `ins`, and `delins` across all coordinate systems.

## Examples

Examples use a synthetic genomic reference to illustrate the pattern. Positions and bases are representative.

### Deletion shifting

Reference context around positions 1337–1351: `CCCAAAAATAAACGC`

**Example 1** — 2-base deletion in a homopolymer run

```
Input:     REF:g.[1340delA;1341delA]
Expected:  REF:g.1343_1344del    ← 3′ shifted to end of A-run
Ferro:     REF:g.1340_1341del    ← not shifted
```

The A-run spans positions 1340–1344 (`AAAAA`). After merging the two individual deletes into `1340_1341del`, the 3′ rule requires shifting rightward. Checking: pos 1340 (A) == pos 1342 (A) ✓ → shift to 1341\_1342; pos 1341 (A) == pos 1343 (A) ✓ → shift to 1342\_1343; pos 1342 (A) == pos 1344 (A) ✓ → shift to 1343\_1344; pos 1343 (A) == pos 1345 (T) ✗ → stop. Final: `1343_1344del`.

**Example 2** — 7-base deletion spanning the A-run boundary

```
Input:     REF:g.[1340delA;1341delA;1342delA;1343delA;1344delA;1345delT;1346delA]
Expected:  REF:g.1342_1348del
Ferro:     REF:g.1340_1346del
```

**Example 3** — single-base deletion

```
Input:     REF:g.[1340delA]
Expected:  REF:g.1344del    ← shifted to last A in the run
Ferro:     REF:g.1340del    ← not shifted
```

### Duplication shifting

Reference context around positions 1048–1065: `ACCTATGATCTGAAGAGC`

**Example 1** — 5-base insertion recognized as dup but not fully shifted

```
Input:     REF:g.[1052_1053insTGATC]
[...]

PR #1345, opened 2026-08-03
fix(normalize): stop refusing a derivation for disagreeing with its input

### LINKED ISSUE #1262: Normalizer: substitution + deletion and the spanning delins spelling reach two different fixed points

nh13:
Stacked on #1341. Fixes #1262 by removing the last gate in the sequence-first pass that read the **input spelling** rather than the sequence.

## The defect

When the derivation produced fewer pieces than the input had members, and any derived piece covered a base that spelling had left between two of them, the pass refused and returned the input verbatim.

It cited `general.md:34` — two variants separated by one or more unchanged nucleotides are described individually — and the citation is sound. The mechanism was not. `general.md:34` is about what the **sequence** separates; the veto measured what the **input's spelling** separated, and those differ whenever a member sits in an ambiguous run. So two encodings of one variant were refused differently — precisely the non-confluence this pass exists to remove. A gate installed to enforce a spec rule was breaking the property the pass provides.

Where the sequence genuinely separates two runs of change, the partition already keeps them apart (`MIN_SEPARATION_NO_FRAME`, `separations_are_meaningful` — both derived from the block alone). The veto added nothing there; it only fired where the block said one member and the spelling said two.

**#1262 is fixed.** `g.[258A>C;260del]` and `g.258_260delinsCA` both reach `g.258_259delinsC`. Both red-when-fixed guards are now positive pins.

## The two forms that move — review these

Both are policy choices rather than spec compliance, and both say so at the assertion.

**#180** — `g.[1006delA;1007delG;1008delG;1009_1010insCCC]` → `g.1006_1009delinsTCCC`. The test pins "no duplicate or overlapping positions", which still holds, and its own comment already named this collapse as the separate improvement it was waiting for. The block trims to the equal-length `AGGT → TCCC`: one member at 4 changed columns against the split form's 6, so the collapse is strictly more minimal. The split form is recoverable only from the spelling — position-wise the block reads as four substitutions.

**#275** — `c.[12C>G;14del]` → `c.12_13delinsG`. The codon-frame merge still declines across the boundary, which is what that test is named for. The block is `CC → G`, answered identically by every alignment because the deleted `C` is ambiguous anywhere in the `CCCCC` run at c.10-14. Verified equivalent on both levels the spec cares about: same denoted sequence, and both translate to `MQKPRGFLKTPGGF*` — the same frameshift, so no protein consequence is lost.

If either should instead be argued upstream before being re-blessed, this PR is the place to say so.

## What this is not

Not the whole of #1235. Every remaining `return None` past a successful derivation is another confluence hole, because the fallback it returns to is spelling-dependent by construction — and each one masks the next. Measured by probing three of them in sequence:

| removed | next divergence surfaced |
| --- | --- |
| the veto | `dup`/`ins` typing: `g.[258_259insA;259_260insC]` → `g.[258dup;259_260insC]` vs `g.258_259delinsAAAC` → `g.260_261insCA` |
| + lone-pure-insertion bail | repeat notation: `g.[257_261A[7];261_262insA]` |

Same masking pattern as `an_indel_haplotype_normalizes_to_its_own_sequence`, which has sixteen defects stacked behind it. So #1235 does not close gate-by-gate; it closes by making the derivation authoritative instead of a candidate that can decline. That is the follow-on.

`input_separator_gaps` and `Piece::overlaps_any` had no other callers and are removed with it.

## Verification

| gate | result |
| --- | --- |
| `cargo nextest run --features dev` | **9094 passed, 0 failed** |
| conformance axis | **183/183** |
| both oracles, manifest-backed | green |
| clippy `-D warnings`, `cargo fmt --check` | clean |

Log grepped for `SIGABRT` and `stack overflow`, not only `FAIL`: none. The axis is the gate that matters here — it is not in the default `cargo nextest` run, and it caught a regression in an earlier attempt at this that 9094 local tests did not.

Closes #1262.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -1777,17 +1777,6 @@ struct Piece {
 }
 
 impl Piece {
-    /// Whether this piece claims any offset inside one of `ranges` as changed.
-    ///
-    /// A pure insertion spans no reference base, so it overlaps nothing — which
-    /// is exactly right for the separator veto: an insertion never swallows a
-    /// base the input left between two members.
-    fn overlaps_any(&self, ranges: &[(usize, usize)]) -> bool {
-        ranges
-            .iter()
-            .any(|&(lo, hi)| self.ref_start < hi && lo < self.ref_end)
-    }
-
     /// Whether this piece is a pure deletion or a pure insertion.
     ///
     /// Only these may be 3'-shifted. `shuffle`'s rotation invariant — advance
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

PR #1350, opened 2026-08-03
fix(normalize): bound a cis junction's 5' shift at its siblings

### LINKED ISSUE #1267: Normalizer: 5'-shifted insertion junction crosses an upstream sibling (and the sweeps cannot see it)

nh13:
Closes #1267.

## What was wrong

Under 5' shuffle, a junction-occupying cis member swept past an upstream sibling, moving its payload from one side of the sibling to the other. The result stays well-formed, its members disjoint, and no warning fires — it simply denotes different bases.

```
reference          ACAAAAAAAACGTACGTACG      A-run at 3-10
g.[4A>G;9_10insA]  applies to  ACAGAAAAAAACGTACGTACG
emitted g.4_5insG              ACAAGAAAAAACGTACGTACG   <- the substituted base moved
```

`clamp_sibling_crossing_junctions` already documented the rule for **both** directions — "a junction may reach a sibling's 5' edge but not pass it: for a sibling claiming `[s, e]`, a 3'-shifted junction must stay at or below `s - 1`, and a 5'-shifted one at or above `e`" — and implemented only the 3' half. This mirrors it. Output is now `g.3_4insG`, which denotes the input's bases.

This is user-reachable, not a test-only mode: `ferro normalize --direction 5prime` and `direction="5prime"` in the Python bindings.

## The issue and two code comments described the mover wrongly

Both said the shape needed no bound because a pure `Insertion` has a zero-width span with "nothing for a sibling to land on." That is true only of the *input* form. The member canonicalises to a duplication before it settles:

```
alone, 5' shuffle:   g.9_10insA  ->  g.3dup        g.16_17insCAG  ->  g.11_13dup
```

Which is also why the span clamp cannot take it: the member *grew*, and a member that grew is refused there for the reasons #1266/#1279 measured. Three separate gates declined this shape, not one.

## Two bounds were needed, not one

Bounding against base-claiming siblings fixed exactly half the measured cases — the tract census went `{15: 100, 36: 16}` to `{15: 50, 36: 8}`. The other half's sibling was a `dup` or an `ins`: junction-occupying, so invisible to `claims_bases`. Mirroring the junction-vs-junction bound (#1290) closed the rest.

Worth noting for its own sake: pinning that census **per sibling position** rather than as a total is what made the second half visible. A scalar would have read as "half fixed, keep going" with no indication of where the remainder lived.

## An admission test was implemented, then refuted by measurement

The first version restricted the 5' half to members the input spelled as an `Insertion`, on the reading that a duplication's span and junction move together so bounding the junction mis-places the copy. Measured against a purpose-built sweep of duplication movers with an upstream sibling:

| | cases | sequence-changing |
|---|---|---|
| restricted to input insertions | 73,376 | **12** |
| unrestricted | 73,376 | **0** |

The gate was not protecting duplication movers from a bad bound — it was excluding them from a good one and leaving twelve broken (`g.[3dup;2_3insTT]` -> `g.[1dup;2_3insTT]`, payloads reordered). So there is no gate, and neither direction consults the input's encoding to decide the output. One of the twelve is pinned by `a_duplication_junction_does_not_cross_an_upstream_junction_sibling`, which fails if the gate is reintroduced.

A first attempt at a guard for that gate passed with the gate both on and off — a restatement, not a guard. Noticing that is what prompted the sweep.

**#1259's 80 regressions do not bear on this rule.** That mirror bounded the junction against a sibling the member was moving *away* from, not one it was moving *toward* — a different rule. #1267's body says so; the code comment did not, and read as a general warning against any 5' bound.

## Test changes

- `a_five_prime_insertion_junction_with_an_upstream_sibling_is_a_known_gap`, which pinned the wrong output, is replaced by `an_insertion_junction_does_not_cross_an_upstream_sibling` asserting sequence preservation in both member orders.
- Two new positive pins for the junction-sibling half and the duplication-mover half.
- The tandem sweep no longer **excludes** the 5' insertion-junction shape and count it: with the residual at zero the exclusion is gone, so a regression there reports the offending descriptions instead of incrementing a bucket. `TRACT_FIVE_PRIME_INSERTION_SEQUENCE_CHANGES_BY_POSITION` is removed, its history kept in a comment.
- #1267 leaves `rewrite_target_corpus.rs`'s PARTITION target list, per that file's own assert-then-flip contract — its guard fired with exactly the instruction it was written to give.

## Verification

| gate | result |
|---|---|
| tract census (116 cases at positions 15 and 36) | **0** |
| `FIVE_PRIME_DUP_DEL_SEQUENCE_CHANGES` | still `0` |
| two-member sweep, 276k cases | green |
| dup-mover upstream-sibling sweep, 73,376 cases (throwaway probe) | 0 sequence-changing, 0 non-applying |
| full suite, `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1` | **9174 passed** |
| conformance axis, manifest-backed | **260 passed** |
| `cargo fmt --check`, `clippy --features dev --all-targets -D warnings`, `cargo check --features python` | clean |
| `tests/proptest-regressions/` | unchanged — no strategy touched |

## Relationship to #1235

The rule is spelling-independent, so unlike the gate that was refuted there is nothing here the sequence-first rewrite needs to delete on principle. It remains a repair pass, and the rewrite should still dissolve the need for it: partitioning derives the payload's position instead of moving it. Note the rewrite does not make the problem vanish — #1235's step 3 still shifts derived pieces, so "a shifting piece must not cross its neighbour" survives as one uniform rule over derived pieces rather than a kind-by-kind test.

---

## Review round 2

No unresolved review threads. Two findings from the CodeRabbit CLI lens, one applied, one rejected — plus one the lens did not name.

### Applied: the new #1267 test could not fail

`an_insertion_junction_does_not_cross_an_upstream_sibling` asserted only that the output denotes the same sequence as the input, via `apply()`. An output *identical to the input* satisfies that trivially — and that is exactly what a clamp which failed to fire would produce. So the test could not distinguish "bounded correctly" from "nothing happened", which is a poor guard for the defect this PR exists to close.

Now pins the canonical output — `TEMPLATE:g.3_4insG` for both member orders, the junction landing at `3|4`, one short of the substituted base at 4, where it coalesces with the sibling — plus an explicit `assert_ne!(actual, input)`. The `apply()` comparison is kept as the supplementary check.

### Also fixed: that test's comment contradicted this PR's own implementation

Not flagged by either lens. The comment said the 5' half is *"admitted only for a member the input spelled as an insertion"* — but `clamp_sibling_crossing_junctions`'s own comment in this same diff records that exact restriction being **tried and refuted by measurement** (over 73,376 duplication-mover cases the unrestricted bound leaves zero sequence changes, the restricted form twelve) and concludes "the direction is simply mirrored". The test comment described the approach that was abandoned. Corrected to match, and it now cites the measurement.

### Rejected: replace the `five_prime` branches with a direction abstraction

The suggestion is to introduce a `ThreePrime`/`FivePrime` enum with helpers for the crossing check, the one-short bound, the restrictive-bound selection and the search order, replacing the five `if five_prime` sites.

Declined, with reasons rather than by omission. It is a pure structural refactor of code this PR has just made correct and covered: no behaviour would change, each branch is individually documented, and the mirrors were verified by reading them against each other (`onto_bases` maps to `s.end` under 5' against `s.start - 1` under 3'; `swept` is `[after, before)` against `(before, after]`; the one-short bound is `junction + 1` against `junction - 1`; the selection is `max` against `min`; the search walks up from the floor against down from the ceiling). Folding a refactor of the normalize core into a behaviour fix also cuts against keeping functional and structural changes in separate PRs. Worth doing on its own if the parallel branching grows further; not worth the risk attached to this change.

### Interaction with #1337, which is already on `main`

Recorded because no single PR's CI can see it. #1337 pinned this exact shape as a known gap:

```rust
const TRACT_FIVE_PRIME_INSERTION_SEQUENCE_CHANGES_BY_POSITION: [(usize, usize); 2] =
    [(15, 100), (36, 16)];
```

with the instruction that *"a **fall** means the insertion half of #1267 is being fixed — in which case update this number"*. This PR is that fall: the exclusion is removed entirely and those 116 cases are now asserted like every other case in the sweep rather than counted. The pin did its job — it is being consumed as designed, not deleted to make a test pass.

**Toward, on a shape the oracles do not contain.** Every input whose output changes is a multi-member cis allele whose junction previously swept past an upstream sibling — well-formed, disjoint and warning-free, but denoting different bases. The biocommons / Mutalyzer / hgvs-rs corpora carry no such input, which is why the defect survived: it was invisible to them and to `FERRO_ASSERT_REPARSE`, and only the sequence-preservation sweep in `cis_junction_crossing_shift.rs` could see it. Single-member descriptions and non-crossing alleles are byte-identical.

Full suite: **9,178 passed, 0 failed**; `cargo fmt --check` and `clippy --all-features --all-targets -D warnings` clean.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -4458,6 +4458,18 @@ pub(crate) fn demote_coincident_tract_repeats<P: ReferenceProvider>(
 /// its insertion lands at `306|307` against a substitution at `307`, which is
 /// the last position the rule permits — so the collapse still fires.
 ///
+/// A sibling's own junction bounds it as well, not only a sibling's bases
+/// (#1290): two junction-occupying members add their payloads at two interbase
+/// points, and their relative order is the only thing fixing the order of those
+/// payloads, so sweeping past one reorders them. That bound stops one position
+/// short of the sibling's junction unless the two payloads commute.
+///
+/// Both shift directions are governed. The 5' half was missing until #1267 —
+/// which is the shape where the sibling is *upstream* and the junction travels
+/// toward it — and the two bounds needed it independently: base-claiming
+/// siblings accounted for half the measured defect, junction-occupying siblings
+/// the other half.
+///
 /// A junction sits immediately 3' of one position: `end` for a duplication
 /// (the copy follows the duplicated bases), `start` for an insertion (whose
 /// span is the gap `start_end`).
@@ -4506,29 +4518,43 @@ pub(crate) fn clamp_sibling_crossing_junctions<P: ReferenceProvider>(
             .filter(|s| s.claims_bases && s.region == a.region && s.accession == a.accession)
             .collect();
 
-        // 3' only, and the 5' case is genuinely unsolved rather than merely
-        // unwritten. A duplication's span and its junction move together under a
-        // 5' shift, so bounding the junction there mis-places the copy: a mirror
-        // of the rule below was measured to turn 80 previously-correct outputs
-        // into silently wrong ones, and was removed as defective rather than
-        // deferred.
+        // Both directions, and the rule is the same in each: a junction may
+        // reach a sibling's edge but not pass it (#1267).
         //
-        // A 5' junction *does* cross, in a shape this rule would not have caught
-        // anyway — when the sibling is **upstream** and the junction travels
-        // toward it. The duplication half of that (`g.[4A>G;9dup]`, which used
-        // to emit `g.4_5insG`) is closed by `blocks_sibling_shift`, which bounds
-        // the duplication's *span* rather than its junction. The insertion half
-        // remains open, narrowed and characterised in
-        // `a_five_prime_insertion_junction_with_an_upstream_sibling_is_a_known_gap`.
-        if after_junction < before_junction {
-            continue;
-        }
+        // The 5' half was absent, and its absence read as deliberate — #1259
+        // measured a 5' mirror turning 80 previously-correct outputs silently
+        // wrong, on the reasoning that a duplication's span and junction move
+        // together so bounding the junction mis-places the copy. But that mirror
+        // bounded the junction against a sibling the member was moving *away*
+        // from, not one it was moving *toward*; it was a different rule, and its
+        // failure does not bear on this one.
+        //
+        // Restricting the 5' half by input spelling — admitting only a member the
+        // input wrote as an insertion — was tried and refuted by measurement.
+        // Over 73,376 duplication-mover cases with an upstream sibling, bounding
+        // every junction mover leaves **zero** sequence changes while the
+        // restricted form leaves **12**, all of them a `dup` whose junction swept
+        // past an upstream sibling's junction. Deciding this from the input's
+        // encoding was both more complex and less correct, so the direction is
+        // simply mirrored.
+        let five_prime = after_junction < before_junction;
         // The junction crossed a base-claiming sibling if it started 5' of the
-        // sibling's first base and ended at or past it.
-        let onto_bases = siblings
-            .iter()
-            .filter(|s| s.start > before_junction && s.start <= after_junction)
-            .map(|s| s.start - 1);
+        // sibling's first base and ended at or past it. Mirrored under a 5'
+        // shift: the junction started 3' of the sibling's *last* base and ended
+        // at or before it, and must stay at or above that base.
+        let onto_bases: Vec<i64> = if five_prime {
+            siblings
+                .iter()
+                .filter(|s| s.end > after_junction && s.end <= before_junction)
+                .map(|s| s.end)
+                .collect()
+        } else {
+            siblings
+                .iter()
+                .filter(|s| s.start > before_junction && s.start <= after_junction)
+                .map(|s| s.start - 1)
+                .collect()
+        };
         // It can equally cross another *junction* (#1290). Two junction-
         // occupying members add their payloads at two interbase points, and
         // their relative order is the only thing fixing the order of those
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

Issue #401, opened 2026-05-21
fix(normalize): CDS-start canonicalisation clamp from PR #385 must respect cross_boundaries=true


nh13:
## Summary

Surfaced by `tests/fixtures/biocommons-normalize/baseline-failures/normalized.txt` (umbrella #325). PR #385 (closes #383) added a 5'-direction canonicalisation clamp at CDS-start for `c.`-axis inputs, intended to prevent a CDS-interior input being silently re-axed onto the 5'UTR axis. The clamp fires unconditionally on `start_axis == Cds`, but it should only fire when the caller did *not* explicitly request boundary crossing.

Live xfail row:

| Input | direction · cross | biocommons expected | ferro got |
|---|---|---|---|
| `NM_000051.3:c.1_2insCA` | 5'+cross | `c.-1_1dup` | `c.1delinsACA` |

biocommons performs the 5'-direction shift across the CDS↔5'UTR boundary (legal because `cross_boundaries=true`) and lands on the canonical `c.-1_1dup`. ferro clamps at the CDS-start and emits `c.1delinsACA` — the PR #385 clamp-shape, but applied where it should not have been.

This row was listed under Pattern I (iii) on #325 / failure-patterns.md and was nominally closed by PR #385; the umbrella was ticked accordingly. In fact the clamp logic is missing the `cross_boundaries` guard.

## Spec basis

HGVS DNA recommendations (per `assets/hgvs-nomenclature/`, v21.0.0 pinned submodule):

- **3' rule** — *"The most 3' position possible is arbitrarily assigned to have been changed."* Applies to all axes and (by the 5'-direction analogue used elsewhere in ferro) to 5'-shift as well.
- **Boundary crossing** — `cross_boundaries=true` is the caller's explicit opt-in to allow the canonicalisation to cross axis boundaries. The CDS-start clamp introduced by PR #385 is a *safety* clamp for `cross_boundaries=false`; it should not override an explicit opt-in.

The mirror case on the CDS-end side (PR #388 / #387) deserves an audit for the same guard — if PR #388's clamp also fires unconditionally on `start_axis == Cds`, the same fix applies symmetrically.

## Scope

`src/normalize/mod.rs::normalize_cds`, the CDS-start clamp branch added by PR #385. The fix is plausibly one line — gate the clamp on `!cfg.allow_crossing_boundaries()` (or equivalent). Verify the analogous CDS-end branch from PR #388 has the same guard.

## Tests

[...]

Issue #383, opened 2026-05-20
fix(normalize): 5'-direction shift must clamp at CDS-start for c.-axis inputs


nh13:
## Summary

Pattern I sub-item (iii) split out from #325. Surfaced by `tests/fixtures/biocommons-normalize/baseline-failures/normalized.txt`.

The `NM_212556.2` cluster: ferro 5'-shifts a c.-axis variant across the CDS-start boundary when biocommons clamps at the boundary.

| Input | biocommons | ferro |
|---|---|---|
| `NM_212556.2:c.1_2insCA` | `c.-1_1dup` | `c.-3_-2dup` |
| `NM_212556.2:c.1delinsCA` | `c.1delinsCA` | `c.-1_1insC` |
| `NM_212556.2:c.2_3insCAT` | `c.1delinsATCA` | `c.-1_1insCAT` |

biocommons stops the 5'-shift at `c.1` (the CDS start); ferro continues into the 5'-UTR (`c.-1`, `c.-2`, etc.).

This complements #337 (Pattern C) which closed the 5'-direction CDS-start clamp on a specific subset of cases. The remaining cases here are in the **insertion / delins canonicalisation path** rather than the pure shift path — ferro's pre-shift canonical-form rewrite already lands the variant in the 5'-UTR before the shift sees it, so the existing clamp logic doesn't catch them.

## Scope

In `src/normalize/`:
- For `Insertion` and `Delins` edits on the c. axis where the alt's canonical-form rewrite would cross the CDS-start boundary (positive→negative coord transition), clamp the rewrite at the boundary.
- This is the canonicalisation-path counterpart to the shift-path clamp in #337.

## Tests

`MockProvider`-backed tests in `tests/normalize_tests.rs`:
- `c.1_2insCA` on a transcript whose CDS-start is at c.1 → `c.-1_1dup` (not `c.-3_-2dup`)
- `c.1delinsCA` → `c.1delinsCA` (not `c.-1_1insC`)
- `c.2_3insCAT` → `c.1delinsATCA` (not `c.-1_1insCAT`)
- Negative: a c.-axis variant whose canonical rewrite stays in CDS proper is unaffected

Demotes 3 inputs from `baseline-failures/normalized.txt`.

## Spec basis

[...]

PR #1354, opened 2026-08-03
fix(normalize): stop an out-of-phase rotation from killing a valid dup

### LINKED ISSUE #1349: 5'-shuffle non-idempotency: c.2_3insTCC in a CCT tract at the CDS boundary
`normalization_is_idempotent_five_prime` finds a 5'-shuffle non-idempotency on the `c.` axis. Surfaced by a fresh proptest seed while sweeping unrelated PRs; **pre-existing**, not caused by any of them.

## Reproducer

```
core  = CCCCTCCTCCTCCTCCCGGGG
sys   = CdsUtr3Plus
dir   = FivePrime
input = NM_TEST.1:c.2_3insTCC
once  = NM_TEST.1:c.1delinsCCTC
twice = NM_TEST.1:c.-1_2dup
```

Pinned seed (proptest's own minimal shrink):

```
cc 9e5c683cf90cbeea9efbabaad533f4288eecda1d6e7844f3f79118a7f7f6f81d # shrinks to case = Case { core: "CCCCTCCTCCTCCTCCCGGGG", sys: CdsUtr3Plus, hgvs: "NM_TEST.1:c.2_3insTCC" }
```

Deliberately **not** committed to `tests/proptest-regressions/normalize_idempotency_proptest.txt`. It was appended automatically by the failing run and reverted, because the branch that surfaced it is a documentation-only change and committing a seed there would attach an unrelated live defect to it. Paste the line above into that file to replay.

## Confirmed pre-existing

Bisected against `8aff336` ("order protein cis members by junction, like the nucleotide axes", #1333) — the tip of `main` before #1335, #1336, #1337, #1338, #1339, #1340 and #1341 merged. The failure there is **byte-identical**: same `once`, same `twice`. So none of those changes introduced it; the property test simply drew a seed that reaches it.

## What is odd about it

The second pass emits `c.-1_2dup`, whose span starts in the 5'UTR and ends in the CDS body. The first pass produced `c.1delinsCCTC`, wholly inside the body. So the second pass moved the description *across the CDS boundary*, which is the class of thing `normalize_cds`'s CDS/UTR axis clamp exists to prevent — worth checking whether the clamp is being applied on pass 1 but not on the re-read of pass 1's output, or whether the `delins` -> `dup` recognition is what crosses.

The tract is relevant: `CCTCCTCCTCC` around c.1 is a `CCT` repeat, and the input inserts `TCC` — a rotation of that unit. So this is an out-of-phase insertion into a tandem tract on a CDS-relative axis, adjacent to the UTR boundary, under a 5' shuffle. That combination is the same neighbourhood as #1280 and #1267.

## Why it is not caught in CI today

`FERRO_ASSERT_IDEMPOTENT=1` is armed in CI, but this input is only reachable from `normalize_idempotency_proptest`'s generator, and proptest draws fresh seeds each run — so whether CI hits it is luck. Committing the seed above (on a branch where it belongs) makes it deterministic.

nh13:
Closes #1349

## What was wrong

`rules::insertion_to_duplication` tries every cyclic rotation of the inserted unit, scores each by the size of the reference tandem tract it abuts, keeps the single highest-scoring one, and only *then* applies the #882 phase gate (`tandem_edit_preserves_insertion`) — returning `None` when that winner fails.

So a high-scoring but out-of-phase rotation shadows a lower-scoring in-phase one and takes the whole result down with it. Measured on the fixture the idempotency proptest found (`GTACGT|CCCCTCCTCCTCCTCCC`, `c.2_3insTCC`):

```
find_tandem_extent(pos=7, "TCC") = Some((5, 1))   <- in phase; the correct dup at c.-1_2
find_tandem_extent(pos=7, "CCT") = Some((8, 2))   <- two copies, wins on count, out of phase
```

`CCT` wins, fails the gate, and `insertion_to_duplication` returns `None`. The correct `TCC` candidate is never reconsidered.

## Why that showed up as a normalization defect

With no `ins -> dup` promotion the 5' path emitted a plain insertion resting at the CDS boundary, and #383's CDS-start clamp rewrote it as a boundary `delins`. That was not a fixed point:

| pass | 5' output |
|---|---|
| input | `NM_TEST.1:c.2_3insTCC` |
| once | `NM_TEST.1:c.1delinsCCTC` |
| twice | `NM_TEST.1:c.-1_2dup` |
| thrice | `NM_TEST.1:c.-1_2dup` |

The description only settled on the third pass, and the two shuffle directions disagreed on the first about an edit that is direction-independent — 3' reached `c.-1_2dup` immediately, via the post-shuffle simple-dup path rather than the rotation search.

Traced with instrumentation at the clamp rather than inferred:

```
5' pass 1:  cds_start=257  new_tx=(256,257)  new_edit=Insertion(CCT)   spanning_dup=false
3' pass 1:  cds_start=257  new_tx=(256,258)  new_edit=Duplication      spanning_dup=true
```

## The fix

Resolve each rotation all the way to the duplication it would emit, and apply the phase gate **per candidate, during selection**. Only rotations that survive it compete.

The clamp is untouched. Once the dup is recognised, #401's `spanning_dup_exception` already lets it through — this was a missing promotion upstream, not a boundary policy question.

The alignment step is extracted to `align_dup_slot` so each rotation can be resolved independently; the selection rules (`ref_count` first, direction tie-break on ties) are unchanged.

**This can only widen the function's domain.** If the old winner passed the gate it is still the argmax of the filtered set, so every existing `Some` is unchanged and only `None` results can become a dup. The pre-existing #882 rejection tests still return `None`, because there *every* rotation fails the gate.

## Tests

Written first and watched fail, each with the exact defect signature:

- `insertion_to_duplication_falls_back_past_a_non_preserving_rotation` (+ a 3' twin) — unit-level, pins the fallback and asserts the emitted dup and the input insertion reconstruct byte-identical sequences.
- `tests/it/issue_1349_rotation_fallback_dup.rs` — end to end. Pins the value in both directions, pins idempotency *separately* so a future change that moves the canonical answer still has to keep it a fixed point, and pins that the equivalent boundary `delins` still normalizes to the same dup (so the fix cannot be mistaken for disabling the clamp).
- The proptest seed that found it is now pinned in `tests/proptest-regressions/normalize_idempotency_proptest.txt`.

## Verification

- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 cargo nextest run --features dev --no-fail-fast` — **9184 passed, 146 skipped, 0 failed**.
- Generated spec artifacts regenerated before the run, so the census guards are judging current code.

**Oracle movement:** none. No test was retired, relaxed, or re-blessed; the six new tests are additions, and the pre-existing `insertion_to_duplication` and #882 suites pass unmodified.

### DIFF
--- a/src/normalize/rules.rs
+++ b/src/normalize/rules.rs
@@ -531,9 +531,16 @@ pub(crate) struct InsToDupResult {
 /// 2. For each rotation `r in 0..unit.len()`, build the rotated unit and
 ///    call `find_tandem_extent` to locate the maximal tandem run abutting
 ///    the insertion point.
-/// 3. Pick the rotation yielding the largest tandem run (ties broken by
-///    iteration order — same convention as `insertion_to_repeat`).
-/// 4. Return the unit-aligned duplication slot inside that tract whose
+/// 3. Resolve each rotation all the way to the duplication it would emit
+///    (step 5), and hold it to the `tandem_edit_preserves_insertion` phase
+///    gate (#882) — a rotation that would decode to a different sequence than
+///    the input is not a candidate at all.
+/// 4. Among the rotations that clear that gate, pick the one yielding the
+///    largest tandem run (ties broken by direction, below). The gate is
+///    evaluated only for a rotation that would otherwise take the lead, which
+///    selects the same winner while skipping the gate's `ref_seq`-sized
+///    reconstructions for rotations that have already lost.
+/// 5. Return the unit-aligned duplication slot inside that tract whose
 ///    position matches the requested shuffle direction:
 ///    - `ThreePrime`: `dup_start_idx = tract.end - unit.len()` — the
 ///      most-3' unit (current default; pre-#340 behavior).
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #1001, opened 2026-07-09
reference: fail loud on an incompatible/misread reference instead of silently reading fields as absent


nh13:
> [!NOTE]
> ## Status — 2026-07-23 (verified on 43ec9ab, v0.10.1)
>
> **Item 1 and the compatibility half of item 2 have landed.** The runtime loader still reads the manifest as an untyped `serde_json::Value` (`multi_fasta.rs:557`), but `from_manifest_inner` now calls `validate_loaded_manifest` **before** those permissive field reads (`multi_fasta.rs:565`), which:
> - runs `check_schema_version` — refuses a manifest from a newer, forward-incompatible `ferro` (`prepare/manifest.rs:479`); and
> - deserializes the raw value into the typed `ReferenceManifest`, which carries `#[serde(deny_unknown_fields)]` (`prepare/manifest.rs:30`) — so a missing required field, a wrong-typed field, or a renamed/typo'd key now fails the load with an actionable "re-run `ferro prepare`" error instead of degrading to `None` (`prepare/manifest.rs:508`).
>
> End-to-end locks on `from_manifest` itself exist: `from_manifest_rejects_wrong_typed_field`, `from_manifest_rejects_typoed_optional_key`, and a newer-schema-version rejection test (`multi_fasta.rs:3248+`). `manifest_schema_version` is stamped on save (`prepare/manifest.rs:308`).
>
> **Remaining, still open:**
> - [x] **Item 1** — route the runtime load through the typed `ReferenceManifest` + `deny_unknown_fields`.
> - [x] **Item 2 — compatibility gate** — `manifest_schema_version` + `check_schema_version`. (A `MIN_SUPPORTED` *floor* is still a `TODO`, per the doc comment at `prepare/manifest.rs:23`; today only the upper bound is enforced.)
> - [ ] **Item 2 — content hash** — `reference_identity_from_manifest` is still **test/conformance-only** (`src/conformance/mutalyzer.rs`, `tests/it/...`); nothing stamps it into the manifest or recomputes+verifies it at load. This is the part that catches in-place drift of a self-consistent, correctly-versioned manifest — the schema gate cannot. Known blind spot unchanged: bulk artifacts (cdot JSONs, genome FASTAs) are excluded from the fingerprint and guarded by version-tagged basenames only.
[...]

Issue #972, opened 2026-07-07
project: incorrect c./p. numbering and reading frame for transcripts with an incomplete 5' start codon (partial CDS start)


nh13:
### Summary

For transcripts whose annotated CDS does not begin on a codon boundary — i.e. the 5′ end of the CDS is incomplete (the GENCODE/Ensembl `cds_start_NF`, "CDS start not found" case, where the transcript begins mid-codon) — ferro numbers the first annotated CDS base as `c.1` unconditionally and translates from that base in frame 0. When the true reading frame is offset, both the `c.` coordinate numbering and the derived `p.` translation disagree with the HGVS/Ensembl convention.

### Root cause

`tx_to_cds` in `src/data/cdot.rs` computes the CDS position as a bare `tx_pos - cds_start + 1` with no start-phase term, and the `Transcript` model (`src/reference/transcript.rs:271`) carries `cds_start`/`cds_end` (`:292`/`:296`) but has no field representing an incomplete-start phase / codon offset. During ingestion, cdot CDS bounds are taken from `start_codon`/`stop_codon` (`src/data/cdot.rs:542`) but no partial-start status is modeled.

The Ensembl convention for a 5′-incomplete CDS is to offset CDS numbering by the start phase and to pad the translateable sequence by that phase so codon 1 stays in frame. Without it, every downstream coordinate on such a transcript is shifted.

### Impact / scope

- Affects `c.` and `n.→c.` numbering and `p.` projection for any transcript flagged as having an incomplete 5′ CDS.
- Rare in RefSeq (`NM_`) transcripts, which almost always carry complete start codons; real for a subset of Ensembl/GENCODE transcripts.
- Silent: output is a well-formed but wrong coordinate, so it will not surface as an error — only as a mismatch against an authoritative normalizer.

### Proposed fix

1. Add an optional start-phase / partial-5′-CDS field to `Transcript` (and its cdot-ingested source struct).
2. Populate it during cdot ingestion. Open implementation question: confirm whether ferro's cdot inputs expose an explicit partial-start attribute, or whether the phase must be inferred (e.g. from CDS-length-mod-3 / CIGAR at the CDS start). This should be settled before implementation — it determines whether this is a pure ingestion change or needs a derivation step.
3. Apply the phase offset in `tx_to_cds` (and the inverse in `cds_to_tx`) and in the CDS translation path.

### Acceptance

[...]

PR #1356, opened 2026-08-03
feat(sequence)!: finish the complement unification, and fix an RNA inversion false alarm

### LINKED ISSUE #1347: Finish the #1318 complement unification: four private copies remain
PR #1346 moved `extractor::classify::is_inversion` and `effect::validate_inversion` onto the shared `sequence::complement_base` / `sequence::reverse_complement`, because each kept a private copy whose pass-through fallback made `complement(R) == R`. Its sweep for other copies named `sequence::complement_char` (documented display contract) and `spdi::convert::apply_alphabet` (an alphabet mapping, not a complement) and concluded neither was a defect. That sweep was under-inclusive — four more private implementations exist.

## Inventory

| site | fallback | how it is used | same defect? |
| --- | --- | --- | --- |
| `src/vcf/from_hgvs.rs:1074` | `_ => c` | builds the ALT for `NaEdit::Inversion` (`:659`) | pass-through, but **produces** rather than compares |
| `src/reference/annotation/validate.rs:155` | `other => other` | assembles a minus-strand codon (`:143`) | pass-through, but produces rather than compares |
| `src/reference/derived_placement.rs:329` | `_ => b'N'` | placement validation | **no** — an unmodelled byte becomes `N`, so `complement(R) != R` |
| `src/reference/derived_tx_structure.rs:148` | `_ => 'N'` | minus-strand exon assembly | **no** — same, maps to `N` |

`src/project/edit.rs:100`'s `complement_base` takes a `Base` enum, so its alphabet is closed and it cannot have this failure mode.

## Why none of these is the live misclassification #1346 fixed

All four **produce** sequence rather than compare it, which is why the suite is green either way. The two fixed sites decided a *classification* — `is_inversion` on every equal-length delins, and `validate_inversion`'s verdict — so a self-complementary `R` was an immediately wrong answer there. Here the consequences are narrower but not nothing:

- the two pass-through copies emit an unmodelled character verbatim into produced data (a VCF ALT field, a codon), so an IUPAC code round-trips uncomplemented;
- the two `N` copies silently *lose* information: `R` becomes `N` rather than `Y`, and a caller cannot tell a genuine `N` from a code that was not modelled.

Both are `sequence`'s job, and #1318 drew the display/comparison line precisely so callers stop making it themselves.

## Why this is a separate change

Folding the `_ => 'N'` pair onto the shared helper **changes behaviour**: today an unmodelled base becomes `N`, whereas `complement_base` refuses and `reverse_complement` passes through. Each of the four needs its own before/after over the reference-data and VCF paths — the same reasoning #1346 used to scope out its own residual (`validate_inversion` still reporting `Valid` for `XX` -> `XX`, because the display helper passes `X` through and the comparison then matches).

## Suggested order

1. The two pass-through copies (`vcf/from_hgvs.rs`, `annotation/validate.rs`) — these are pure display, so `sequence::reverse_complement` is a like-for-like swap plus IUPAC coverage.
2. The two `N`-mapping copies — decide first whether losing the code to `N` is the intended contract at those call sites; if it is, document it there rather than unifying.
3. `validate_inversion`'s `XX` residual, which needs a "cannot validate this alphabet" answer and so changes the public `InversionValidation` surface.

nh13:
Closes #1347

Finishes the #1318 complement unification. Three commits, matching the issue's suggested order — read them in sequence.

## 1. The two pass-through copies (`97235c1`)

`vcf::from_hgvs` and `reference::annotation::validate` each carried a private `reverse_complement` whose fallback returned the character unchanged. Both *produce* sequence rather than compare it, so neither showed the misclassification #1346 fixed — but an unmodelled character went verbatim into produced data.

The VCF site is the observable one. An inversion spanning an IUPAC code emitted it uncomplemented into the ALT field:

```
ALT was "GRTT", should be "GYTT"     # R (A or G) complements to Y (C or T)
```

The start-codon site is behaviour-preserving: it upper-cased where the shared helper preserves case, but `validate_start_codon` upper-cases the assembled codon before matching it against ATG/CTG/GTG/TTG, so the delta is inert. Covered by the existing minus-strand split-codon tests.

## 2. The two `N`-mapping copies (`14be296`)

Both mapped every non-ACGT byte to `N`. In the placement validator that creates the exact spurious match its own comment claimed to prevent:

`validate_sequence` compared `bases_match(ng, complement(genome))` where `complement` returned `N` for anything unmodelled — so an `N` on the `NG_` side matched **any** unmodelled genome byte. A placement could validate against bases nobody could complement. It also counted a correctly-paired ambiguity as a mismatch, since `R` collapsed to `N`, which does not equal the `Y` opposite it.

`reconstruct_spliced_mrna` had the same map on its minus-strand arm, so an `R` in the genome became `N` in the reconstructed mRNA — indistinguishable from a genuine `N`.

That arm also upper-cased where the plus-strand arm passes the slice through verbatim. Both now preserve case alike, which is inert in production because the FASTA providers upper-case on read (`fasta.rs:199`, `multi_fasta.rs:1295`).

`validate_sequence`'s doc comment claimed `N`/ambiguity bases "count as mismatches" and that an assembly-gap run of `N`s inflates the mismatch fraction. Neither was true — `N` complements to `N` and matched itself on both strands. Corrected to describe what the code does.

### Measured against the prepared reference

The issue asked for a before/after on the reference-data paths. Both were run against a real prepared reference (cdot 0.2.32, GRCh38).

**Blast radius, from the genome's actual alphabet.** `validate_sequence` complements the *genome* byte, so the genome FASTA's alphabet bounds the change exactly. Full scan of GRCh38.fna (3.1 GB):

```
T 559373567   A 558619211   G 413917617   C 413530454
t 366543471   a 364497992   g 231314379   c 229022463
N 161611379
Y 36   R 29   W 15   M 8   K 8   S 5   B 2
```

Two things follow. **Nothing lies outside the modelled alphabet**, so the new "no complement ⇒ mismatch" path never fires on this reference — it is a guard against corrupt input, not a live behaviour change. But **103 genuine ambiguity codes do exist**, so the `R → Y` correction is live wherever a minus-strand slice covers one; previously each of those collapsed to `N`.

**Artifact re-derivation.** `derive_tx_placements` re-run against the same manifest and diffed against the committed `derived_transcript_placements.json`, which the old code produced. The only differences are two schema fields added by earlier PRs (`schema_version` from #1001, `cds_start_incomplete` from #972) that the older artifact predates:

```
2d1  <   "schema_version": 1,
57d55  <       "cds_start_incomplete": false,
95d92  <       "cds_start_incomplete": false,
```

Every derived value — exons, coordinates, `mismatch_fraction` — is byte-identical. No drift from this change.

Stated honestly: that artifact covers only 2 accessions, so on its own it is weak evidence. The alphabet scan is what actually bounds the risk, and it bounds it at 103 positions in 3.1 Gb.

## 3. The `validate_inversion` residual (`dd9fa0d`) — not the bug the issue described

The issue proposed teaching `validate_inversion` a "cannot validate this alphabet" answer, because reference `XX` + provided `XX` returns `Valid`. **That premise is wrong, and the spec says so.**

`assets/hgvs-nomenclature/docs/background/standards.md:36` defines `X`:

```
|   X†   | A, C, G or T |        masked nucleotide        |
|   -†   |     none     |   gap of indeterminate length   |
† used in alignment only.
```

`X` is set-identical to `N`. The IUPAC complement map is the set-complement map, so `X` is **self-complementary** — today's `Valid` is the correct answer, reached by accident through the display helper's pass-through. Rejecting it would have replaced a right answer with a vaguer one.

It is also unreachable from parsed input:

```
NC_TEST.1:g.1_2invXX -> Err(Parse { pos: 18, msg: "Unexpected trailing characters: 'XX'" })
```

**The real defect next door.** `reverse_complement` maps `U` to `A` but never produces a `U`, so it is not an involution over RNA — the reverse complement of an RNA reference comes back spelled in the DNA alphabet. With bare case-folded equality, a legal `r.` inversion was reported as a mismatch against its own reference:

```
validate_inversion(Some("AT"), Some("AU"), 1, 2)
  -> SequenceMismatch { expected: "AT", provided: "AU" }
```

`NM_TEST.1:r.1_2invau` parses and round-trips unchanged. **But see review round 2 below**: parseability is not reachability — `validate_inversion` has no callers outside its own tests, and the live inversion check is a different function with different semantics. This is a latent defect in a public-but-unused API. Fixed by folding `U` onto `T` for the comparison only; the reported `expected` string is untouched.

Also in this commit:

- **Documents what `Valid` asserts**, which was undeclared: exact equality of IUPAC code strings up to case and the T/U fold, *not* "consistent with the reference". That is why `("RY","RY")` is valid while `("RR","RR")` is a mismatch.
- **Adds `#[non_exhaustive]`** to `InversionValidation` while it still has no in-tree consumers — the only free moment to add it. Shipped as `feat!` with a `BREAKING CHANGE:` footer; see round 2.
- Pins the `X` finding in a test, so a future "reject unknown alphabet" change has to confront the spec text first.

### Known and deliberately not changed

`("AN","NN")` reports a mismatch with expected `"NT"`, even though every concrete instantiation of `AN` reverse-complements into `NN`. That is the equality rule behaving as specified rather than a subsumption rule. Changing it would redesign what the function means, so it is documented instead.

## Tests

Seven new tests, each written first and watched fail:

| test | what it caught |
|---|---|
| `inversion_alt_complements_iupac_ambiguity_codes` | ALT `GRTT` instead of `GYTT` |
| `an_unmodelled_genome_byte_does_not_match_an_n_in_the_ng` | placement validating against unmodelled bytes |
| `an_iupac_code_complements_to_its_partner_on_the_minus_strand` | `R` collapsing to `N` and failing to match `Y` |
| `minus_strand_reconstruction_complements_iupac_codes` | reconstructed mRNA `ACNT` instead of `ACYT` |
| `an_rna_inversion_validates_against_a_dna_reference` | the T/U false mismatch |
| `t_u_folding_does_not_mask_a_real_mismatch` | guard: folding must not validate everything |
| `the_spec_defined_masked_nucleotide_is_self_complementary` | guard: pins the spec's `X` semantics |

## Verification

- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 cargo nextest run --features dev --no-fail-fast` — **9191 passed, 146 skipped, 0 failed**.
- Generated spec artifacts regenerated before the run.

**Oracle movement:** none. No test retired, relaxed, or re-blessed.

---

## Review round 2

Both lenses run. CodeRabbit CLI: **0 findings**. The agent lens found seven, of which five were accepted and fixed; the two `[Info]` ones are addressed in docs.

**Accepted and fixed:**

1. **A pre-existing doc comment was severed from its test.** The new tests were inserted between `an_iupac_code_is_not_its_own_complement`'s doc comment and the `#[test]` it documents, so that test lost its docs and mine absorbed a `#1249` R/Y rationale that has nothing to do with T/U. Genuinely my error; both comments are re-attached.

2. **The reachability claim in the commit message was wrong.** It said `NM_TEST.1:r.1_2invau` "parses and round-trips unchanged, so this is reachable from ordinary input". Parseability of the description is not reachability of the function — `validate_inversion` has **no callers** outside its own unit tests, which the same commit conceded two paragraphs later when justifying `#[non_exhaustive]`. The message now states the scope plainly: a latent defect in a public-but-unused API.

   The review also established that the **live** inversion check is a different function with different semantics — `normalize::validate::validate_sequence`, reached from `normalize/validate.rs`'s `NaEdit::Inversion` arm, compares stated bases against the reference span *itself* rather than its reverse complement, and has no T/U fold. That is recorded in the commit message and left for a follow-up rather than widened into this PR.

3. **`#[non_exhaustive]` would have shipped in a patch release.** `release-plz.toml` sets `features_always_increment_minor = true`, so `feat:` takes a minor bump and `fix:` a patch. Adding `#[non_exhaustive]` breaks a downstream exhaustive `match` at compile time, so shipping it under `fix(effect):` would deliver a compile-breaking change as 0.11.0 → 0.11.1. Retyped `feat(effect)!:` with a `BREAKING CHANGE:` footer.

4. **`X` was given two contradictory meanings in one PR** — spec-defined and self-complementary in `effect`, the exemplar of an unmodelled byte in `derived_placement`. Worse, the doc claimed the `X` answer was "a correct answer, not a pass-through accident" when `complement_base` does not model `X` at all, so it *is* pass-through — the same pass-through returns `Valid` for `("ZZ","ZZ")`, which the spec does not define. The doc now says exactly that, and notes `X` belongs in the table if the answer is ever to be load-bearing.

5. **`compare_key` used `str::to_uppercase`, which can expand.** `ß` uppercases to `SS`, while the length gate above measures bytes — so `("SS", "ß")` clears a 2-byte length check and then compares equal, reporting `Valid`. Pre-existing rather than introduced, but the PR claims to specify what `Valid` asserts, so it is now an ASCII byte-wise fold that measures the same thing the gate does.

**Addressed in documentation:**

6. Two of the three new `effect` tests (`t_u_folding_does_not_mask_a_real_mismatch`, `the_spec_defined_masked_nucleotide_is_self_complementary`) pass against pre-change code. They are guards, not regression tests, and are now labelled as characterization pins so they are not mistaken for the latter. The four tests that do fail pre-change are unchanged.

7. The plus/minus asymmetry in `validate_sequence` — an unmodelled byte still matches itself on the plus strand, because that arm compares bytes directly and never needs a complement — is now documented rather than left implicit.

**Not changed:** the review confirmed the case-semantics claim independently (every `GenomeSlice` implementor upper-cases, so no lowercase slice can reach the byte-exact `find_clean_offset`), confirmed the rewritten `validate_sequence` closure is equivalent to the old one for all ACGTN input and that `is_none_or` is not inverted, and confirmed `#[non_exhaustive]` breaks nothing in-crate and has no Python exposure.

Re-verified after the fixes: **9191 passed, 146 skipped, 0 failed** with both oracles; fmt and clippy clean.

### DIFF
--- a/src/effect/mod.rs
+++ b/src/effect/mod.rs
@@ -835,7 +835,12 @@ impl Default for KozakAnalyzer {
 // =============================================================================
 
 /// Inversion validation result.
+///
+/// `#[non_exhaustive]` so a future verdict can be added without breaking
+/// downstream `match`es. Added while the enum has no in-tree consumers, which is
+/// the only free moment to do it.
 #[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
 pub enum InversionValidation {
     /// Inversion is valid - sequence matches expected reverse complement.
     Valid,
@@ -861,6 +866,39 @@ pub enum InversionValidation {
 }
 
 /// Validate that an inversion sequence matches the reference.
+///
+/// # What `Valid` asserts
+///
+/// **Exact equality of IUPAC code strings**, up to case and the T/U alphabet
+/// fold — *not* "the description is consistent with the reference". The
+/// distinction matters wherever ambiguity codes are involved, because the codes
+/// denote base *sets*:
+///
+/// * `("RY", "RY")` is `Valid` and `("RR", "RR")` is a `SequenceMismatch`, because
+///   `R` complements to `Y`. Both are right under the equality rule.
+/// * `N`, `S` and `W` are self-complementary under the modelled IUPAC table, so
+///   an all-`N` inversion validates.
+/// * The spec's masked-nucleotide symbol `X` also validates against itself, and
+///   the spec agrees it should — `standards.md` defines `X` as the base set
+///   `A, C, G or T`, identical to `N`, which makes it self-complementary. But be
+///   precise about *why* it happens here: [`crate::sequence::complement_base`]
+///   does not model `X`, so the right answer arrives through
+///   [`crate::sequence::reverse_complement`]'s pass-through of unmodelled
+///   characters, not through the IUPAC table. The same pass-through returns
+///   `Valid` for `("ZZ","ZZ")`, which the spec does *not* define. If the `X`
+///   answer is ever to be load-bearing, `X` belongs in the table.
+/// * Under a *subsumption* rule the answers would differ: `("AN", "NN")` is
+///   reported here as a mismatch (expected `NT`), even though every concrete
+///   instantiation of `AN` has a reverse complement contained in `NN`. Equality
+///   is the rule this function implements; subsumption is not offered.
+///
+/// # Scope
+///
+/// An inline sequence on `inv` is a ferro leniency extension: the spec's own
+/// syntax (`assets/hgvs-nomenclature/docs/syntax.yaml`, DNA and RNA `inv`) is
+/// `range "inv"` with no sequence and no length, so a spec-conformant
+/// description gives this function nothing to check and takes the
+/// `(Some(_), None) => Valid` path.
 ///
 /// # Arguments
 /// * `reference_sequence` - The reference sequence at the inversion position.
@@ -894,7 +932,7 @@ pub fn validate_inversion(
             // The provided sequence should be the reverse complement of reference
             let expected_revcomp = crate::sequence::reverse_complement(reference);
 
-            if provided.to_uppercase() == expected_revcomp.to_uppercase() {
+            if compare_key(provided) == compare_key(&expected_revcomp) {
                 InversionValidation::Valid
             } else {
                 InversionValidation::SequenceMismatch {
@@ -912,6 +950,32 @@ pub fn validate_inversion(
         },
         (None, None) => InversionValidation::Valid, // Nothing to validate
     }
+}
+
+/// Normalise a sequence for the inversion comparison: upper-case, and fold `U`
+/// onto `T`.
+///
+/// The T/U fold is not cosmetic. [`crate::sequence::reverse_complement`] maps
+/// `U` to `A` but never produces a `U`, so it is not an involution over RNA:
+/// the reverse complement of an RNA reference comes back in the DNA alphabet.
+/// A perfectly legal `r.` inversion — `NM_TEST.1:r.1_2invau` parses and
+/// round-trips unchanged — was therefore compared against a `T`-spelled expected
+/// sequence and reported as a mismatch. `T` and `U` are the same base in
+/// different alphabets, so they must compare equal.
+///
+/// ASCII-only, deliberately, and byte-wise. The length gate above measures
+/// `provided.len()` — bytes — so a Unicode-aware fold here would disagree with
+/// it: `str::to_uppercase` expands `ß` to `SS`, and `("SS", "ß")` clears a
+/// 2-byte length check and then compares equal, reporting `Valid` for an input
+/// that is not a nucleotide sequence at all. Folding per byte keeps the two
+/// checks measuring the same thing.
+fn compare_key(seq: &str) -> Vec<u8> {
+    seq.bytes()
+        .map(|b| match b.to_ascii_uppercase() {
+            b'U' => b'T',
+            other => other,
+        })
+        .collect()
 }
 
 #[cfg(test)]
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1359, opened 2026-08-04
feat(effect)!: drop the dead inversion validator that contradicted the live one

### LINKED ISSUE #1358: effect::validate_inversion is dead public API that contradicts the live inversion validator
`effect::validate_inversion` is dead public API whose convention contradicts the live inversion validator.

## Two validators, opposite conventions

ferro validates an inline `inv` sequence in two places that disagree about what that sequence *means*.

**The live one** — `normalize::validate::validate_sequence`, reached from `normalize/validate.rs:133-136`'s `NaEdit::Inversion { sequence }` arm — compares the stated bases against the **reference span itself** (`src/normalize/validate.rs:581+`, `actual_bytes = &ref_seq[start_idx..end_idx]`). Measured on the `tests/it/issue_280_refseqmismatch_corrected_flag.rs` fixture (`c.1..c.4 = ATGC`):

```
NM_TEST.1:c.1_2invAT     -> NM_TEST.1:c.1_2=          0 warnings   (states the reference -> valid)
NM_TEST.1:c.1_2invAA     -> RefSeqMismatch { stated_ref: "AA", actual_ref: "AT" }
NM_TEST.1:c.20_22invAAA  -> RefSeqMismatch { stated_ref: "AAA", actual_ref: "TTT" }
```

So the live convention is "the inline sequence restates the reference bases", exactly as `del<seq>` and `dup<seq>` do.

**The dead one** — `effect::validate_inversion` (`src/effect/mod.rs:911`) — compares the provided sequence against the **reverse complement** of the reference (`:927-929`). For a non-palindromic span these give opposite verdicts: with reference `AG`, the live validator wants `invAG` and the `effect` one wants `invCT`.

`effect::validate_inversion` and `InversionValidation` have **no callers** anywhere — not in `src/`, `tests/`, `examples/`, `benches/`, or the Python bindings, and `InversionValidation` is not re-exported from `lib.rs`. They are reachable only as public API via `pub mod effect` (`src/lib.rs:40`).

## Why now

#1356 (issue #1347) fixed a T/U comparison bug in `effect::validate_inversion` and added a "What `Valid` asserts" doc block specifying its semantics. That was a genuine latent defect, but it polished the dead validator and **formally documented the divergent convention**, which entrenches it rather than resolving it.

Two things worth recording from that work:

- The live path does **not** share the T/U bug — verified, `r.1_2invau` validates with zero warnings, so the `r.` lowering already folds `U` to `T` before validation. #1356's fix was correct but confined to dead code.
- Per the spec, `inv` has **no inline sequence at all**: `assets/hgvs-nomenclature/docs/syntax.yaml` defines DNA and RNA `inv` as `range "inv"` / `positions "inv"`, with no sequence and no length. ferro's `invATG` / `inv3` are leniency extensions, so neither convention is spec-mandated — but the live one is the coherent choice, matching how every other edit's inline sequence behaves.

## Options

1. **Delete** `effect::validate_inversion` and `InversionValidation`. They are dead, contradict the live path, and the spec does not define the construct they validate. A breaking change, but the enum already took one in #1356 (`#[non_exhaustive]`), and #1247 is an open v0.12.0 release.
2. **Reconcile** — reorient `effect::validate_inversion` onto the live convention (compare against the reference span, not its reverse complement) and keep it as a public helper. Preserves the API but leaves two implementations of one rule.
3. **Leave it**, and document the divergence explicitly at both sites so nobody adopts the `effect` one by accident.

Option 1 is the recommendation: a public validator that disagrees with the validator ferro actually uses is a trap for the first caller who finds it.

## Related

- #1347 / #1356 — the T/U fix and the doc block that specified the dead convention.
- The `("AN","NN")` subsumption question is adjacent and deliberately unresolved: `validate_inversion` implements *equality* of IUPAC code strings, so `("AN","NN")` is a mismatch even though every concrete instantiation of `AN` reverse-complements into something `NN` denotes. If option 1 is taken this becomes moot; under 2 or 3 it should be decided and written down.

nh13:
Closes #1358

Removes `effect::validate_inversion` and `effect::InversionValidation` — dead public API whose convention contradicted the validator ferro actually runs.

## Two validators, opposite answers

**The live one**, `normalize::validate::validate_sequence`, reached from `normalize/validate.rs`'s `NaEdit::Inversion { sequence }` arm, compares the stated bases against the **reference span itself** — the same convention `del<seq>` and `dup<seq>` use. Measured on the `issue_280` fixture (`c.1..c.4 = ATGC`):

```
c.1_2invAT     -> c.1_2=                                     0 warnings
c.1_2invAA     -> RefSeqMismatch { stated_ref: "AA",  actual_ref: "AT" }
c.20_22invAAA  -> RefSeqMismatch { stated_ref: "AAA", actual_ref: "TTT" }
```

**The deleted one** compared against the **reverse complement**. For any non-palindromic span the two returned opposite verdicts — with reference `AG`, the live validator wants `invAG`, this one wanted `invCT`.

It had no callers in `src/`, `tests/`, `examples/`, `benches/`, or the Python bindings, and the enum was not re-exported from `lib.rs`. Reachable only as public API via `pub mod effect`.

## Why delete rather than reconcile

The spec does not favour the deleted convention: `inv` has **no inline sequence at all** — `assets/hgvs-nomenclature/docs/syntax.yaml` defines DNA and RNA `inv` as `range "inv"` / `positions "inv"`. ferro's `invATG` is a leniency extension, so neither convention is mandated, and the live one is the coherent choice because every other edit's inline sequence restates the reference.

Keeping a second, contradictory implementation of one rule only preserves a trap for the first caller who finds the wrong one.

Deleting also retires two questions #1356 documented rather than answered — neither needs deciding for a function nobody calls:

- what `Valid` asserts (equality of IUPAC code strings vs subsumption — the `("AN","NN")` case, where the provided description is *weaker* than the truth rather than wrong);
- whether the spec's masked-nucleotide `X` validating against itself was a correct answer or an artefact of `reverse_complement`'s pass-through (it was the latter mechanically — `complement_base` does not model `X`, and `("ZZ","ZZ")` returned `Valid` identically).

## The #1356 T/U fix is not lost

That fix was confined to this dead function. The live path was verified to handle `u` already: `NM_TEST.1:r.1_2invau` validates with **zero warnings**, because the `r.` lowering folds `U` to `T` before validation. The retro that surfaced this initially recorded "the live validator has no T/U fold" as a defect; probing it disproved that, which is why this PR is a deletion and not a second T/U fix.

## Verification

- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 cargo nextest run --features dev --no-fail-fast` — **9186 passed, 146 skipped, 0 failed**.
- Repo-wide `grep` for `validate_inversion|InversionValidation` returns **0** hits.
- `test_reverse_complement` is retained (it exercises the shared `sequence::reverse_complement`, not the deleted validator); its section header is retitled since it no longer sits under inversion-validation tests.

**Oracle movement:** none. The 8 removed tests all exercised the deleted API exclusively; no test of retained behaviour was changed, relaxed, or re-blessed.

**Semver:** typed `feat!` with a `BREAKING CHANGE:` footer. `release-plz.toml` sets `features_always_increment_minor = true`, so `fix` would map to a patch — a public removal must not ship in one.


nh13:
@coderabbitai review

nh13:
@coderabbitai pause

Issue #1351, opened 2026-08-03
Tests: the cis apply oracle cannot express an identity member, so every sequence-preservation sweep is blind to those outputs


nh13:
> **Correction (2026-08-04).** The mechanism below was wrong as originally filed, and the error mattered: it pointed at the test oracle when the defect was in the conversion. `hgvs_to_spdi` does **not** decline `NaEdit::Identity`. It accepts it and returns a **zero-width** triple — `g.263=` becomes `TEMPLATE:262::`, empty deletion and empty insertion. That triple aliases a sibling's insertion junction at the same interbase, and it is `apply_with`'s own "two pure insertions at one interbase" guard that then declines the whole description. Corrected inline below; see the resolution at the end.

`hgvs_to_spdi` converts `NaEdit::Identity` to a zero-width triple whose span claims no bases, so when such a member sits at the same interbase as an insertion or duplication, `tests/it/common/cis_apply_oracle::apply` cannot tell the two apart and returns `None` for the whole description. Every sequence-preservation sweep routes those cases into its `skipped` counter and asserts nothing about them.

## Why it matters

`apply` is the load-bearing oracle for the whole cis-allele suite. Its module doc says so:

> [`apply_with`] is the load-bearing piece: its `claimed_from` walk is what makes "declines an overlapping description" true, and every sequence-preservation assertion in `cis_junction_crossing_shift.rs`, `issue_1234_sibling_clamped_shift.rs`, `issue_1254_sibling_crossing_shift.rs`, `issue_1261_cis_member_order.rs`, `issue_1281_reducing_member_shift.rs` and `repeat_span_sibling_overlap.rs` rests on it.

So a defect that only manifests in identity-bearing output cannot be caught by any of them. The sweeps' hollowness bound (`skipped * 10 < checked`) caps the *share* of skipped cases but says nothing about **which shapes** are skipped, so the blindness is uniform and invisible.

This is not hypothetical — the normalizer does emit such output. Under 5' shuffle:

```
g.[261_262insGA;262_263insA;263del]  ->  g.[261_262dup;263=]
```

(core `TCCCAGAAAAT` under `padded()`.) That output is beyond the oracle entirely.

## How it was found

[...]

PR #1361, opened 2026-08-04
test(normalize): record the 5' shuffle confluence gap

nh13:
Every row in `cis_spelling_confluence_gap.rs` was blessed against the 3' rule, but confluence is a property of the normalizer rather than of one shuffle direction, and `--direction 5prime` is a supported public option on both the CLI and the Python bindings. Two rows that converge under 3' diverge under 5', and nothing measured that.

This records the gap in the file's existing assert-then-flip idiom — the two rows are asserted to still diverge, the other three to still converge — so a change in either direction is a result rather than a maintenance chore. **No production change.**

## Severity is pinned, not assumed

Both spellings still denote the input's bases and both are stable fixed points, so this is criterion 1 of #1235 and nothing else. That is also why neither oracle sees it: `FERRO_ASSERT_IDEMPOTENT` re-normalizes one spelling and finds it stable, and `FERRO_ASSERT_REPARSE` finds it well-formed. Only comparing two spellings of one variant exposes it.

## Both tables are measured in both directions

The eight `DIVERGENT` rows were blessed against the 3' rule alone, so `the_eight_spelling_pairs_still_diverge_under_five_prime` pins that they diverge under 5' too. They reach different spellings than under 3' — #1296's merged form settles at `g.273C>A` here against `g.273delinsA` there — so it measures the 5' pipeline rather than restating the 3' result.

## The output the applier cannot express is checked, not just counted

Under 5' shuffle one spelling settles with a redundant `263=` beside a real edit:

```
g.[261_262insGA;262_263insA;263del]  ->  g.[261_262dup;263=]
```

An identity member has no SPDI triple, so `hgvs_to_spdi` declines and the one oracle that judges denotation cannot evaluate the description at all. Counting the skip would leave the "and nothing worse" half of that test's name unverified for exactly the output it most needs to defend. So `strip_identity_members` drops the `=` member — which by definition changes nothing — to build an equivalent description the applier *can* express, and its denotation is asserted against the same expected bases. The skip is still counted and pinned at one. Confirmed non-vacuous: making the strip a no-op turns the test red on its own diagnostic.

That oracle blindness is a defect in the oracle rather than in this row and is filed separately as #1351 — this test is the pinned instance that issue names. It is referenced, not closed: this records the blindness, it does not fix it.

## Two candidate fixes were implemented and refuted by measurement

Recorded in the comments so the next attempt does not repeat them. Threading `ShuffleDirection` into `canonicalize_from_sequence` moves 298 of 39,600 swept outputs and every one is sequence-preserving before and after, so it is a lateral re-spelling with no correctness content. Accepting an identity member in `collect_canonical_edits` changes no output, because two further gates refuse behind that one.

## The pins are guarded against going stale

`DIVERGENT_UNDER_FIVE_PRIME` is keyed by issue id and matched against `CONVERGED`, so an id naming no row is inert: it asserts nothing while the set goes on claiming to pin a divergence. Deleting a row and leaving its id behind is exactly how that happens, and it is silent — verified by adding a bogus id, which left every test in the file green. There is now a guard for it, in the same spirit as the row count already pinned by `the_eight_spelling_pairs_still_diverge`.

The identity-member test reads its input from #1321's row rather than repeating it as a literal, so editing that row cannot leave the test exercising a different variant while its prose still names #1321. The expected output stays literal — that string is the finding.

Both guards were confirmed to fail when violated, not merely added.

## Verification

Run on this branch against `main` at `12bba7c`:

- `cargo nextest run --features dev` — 9190 passed, 0 failed
- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 cargo nextest run --features dev` — 9175 passed, 0 failed (proptests excluded, as CI does)

### DIFF
--- a/tests/it/cis_spelling_confluence_gap.rs
+++ b/tests/it/cis_spelling_confluence_gap.rs
@@ -12,9 +12,19 @@
 //! Each row asserts the two spellings still DIVERGE. This test failing means a
 //! pair converged, and that pair's blessed expectation should be re-blessed to
 //! the converged form.
+//!
+//! Both tables are measured in both directions. Every row above was blessed
+//! against the 3' rule, but confluence is a property of the normalizer rather
+//! than of one shuffle direction, and `--direction 5prime` is a supported public
+//! option — so [`DIVERGENT_UNDER_FIVE_PRIME`] records the [`CONVERGED`] rows that
+//! converge under 3' and diverge under 5', and
+//! [`the_eight_spelling_pairs_still_diverge_under_five_prime`] pins that the
+//! [`DIVERGENT`] eight diverge under 5' too. Both use the same assert-then-flip
+//! idiom: a red assertion is a result, not a maintenance chore.
 
-use crate::common::cis_apply_oracle::{apply, normalize};
+use crate::common::cis_apply_oracle::{apply, normalize, normalize_in};
 use crate::common::synthetic::padded;
+use ferro_hgvs::ShuffleDirection;
 
 /// `(issue, core, split spelling, merged spelling)`.
 const DIVERGENT: &[(&str, &str, &str, &str)] = &[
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #1353, opened 2026-08-03
Assert at the normalization seam that no emitted coordinate exceeds its accession's length


nh13:
`parse_hgvs` does not bounds-check coordinates against a reference — it has no provider — so an output naming a position past the end of a contig or transcript **re-parses cleanly and re-normalizes stably**. Both existing oracles are structurally blind to it:

- `FERRO_ASSERT_REPARSE` accepts it, because it is well-formed.
- `FERRO_ASSERT_IDEMPOTENT` accepts it, because it is a fixed point.

The class therefore only gets caught when someone writes a test that reads the coordinates back and compares them to `get_sequence_length` by hand. That has now happened at least three times, each as a separate issue:

| issue | shape | state |
|---|---|---|
| #1274 | `T:g.[8_9insA;10del]` -> `T:g.10_11=` on a 10-base contig | closed (#1341) |
| #1343 | `c.[*10dup;*11dup]` -> `c.*11_*12insAA`, `*12` does not exist | closed (#1339) |
| #1307 | `g.[24dup;24C>G]` -> `g.[24C>G;24_25insC]` on a 24-base contig | **open, still reproduces on `e5b7633`** |

Three instances of one defect class, each found by hand, is the argument for a seam-level invariant rather than a fourth per-shape test.

## What to add

#1274's body already proposes it:

> assert the invariant generally: no emitted coordinate may exceed `get_sequence_length` for its accession. That last check is cheap and would have caught this class at the seam, in the same way `FERRO_ASSERT_REPARSE` catches #1263's.

The natural home is beside the two existing oracles at `normalize_core_checked`'s single exit, behind its own env var (`FERRO_ASSERT_IN_BOUNDS`, say), `#[cfg(debug_assertions)]` and `OnceLock`-read like the others, and set alongside them in `ci.yml` and `nightly-mutalyzer.yml`.

## What makes it non-trivial

- **The axes are not all bounded the same way.** A `c.` position can legitimately exceed the CDS length (`*n` 3'UTR offsets, `-n` 5'UTR), so the check needs the axis-aware conversion rather than a raw comparison — see `cds_axis_position`, which today has a floor (`tx < 1`) but no ceiling and will happily count past the last base into `ThreePrimeUtr(K+1)` (noted when #1343 closed).
- **Circular references.** `m.`/`o.` wrap, so "past the end" is a different question there; SVD-WG006 admits the reversed `<high>_<low>` range for deletions and duplications only.
[...]

Issue #1344, opened 2026-08-03
Terminal re-spelling turns a cancelling collision into an overlapping allele


nh13:
The terminal-base re-spelling added for #1327 keeps a junction repair inside the sequence, which is right — but on a collision whose two members **cancel**, it produces an allele ferro's own strict mode rejects.

## Reproducer

35-base transcript, `cds_start = 13`, `cds_end = 24`, so the 3'UTR is `c.*1..c.*11` and `*10`/`*11` are both `A`:

```
NM_TEST.1:c.[*10del;*11dup]  ->  NM_TEST.1:c.[*11del;*11delinsAA]
```

Deleting one `A` and duplicating the other cancel exactly, so the allele denotes the reference. The output instead puts **two coincident edits on one position**, which strict mode refuses:

```
NM_TEST.1:c.*11 has 2 coincident cis-allele edits (del, delins);
HGVS spec defines no canonical form for this case (OverlapConflictingEdits / W5002)
```

It is a **stable fixed point** — re-normalizing returns it byte for byte — so it is a permanent answer, not a transient one on the way to something better. That is the #1234 overlapping-members class.

The `r.` axis does the same: `NM_TEST.1:r.[*10del;*11dup]` -> `NM_TEST.1:r.[*11del;*11delinsaa]`. So does the equivalent body-region spelling of the same positions, `c.[22del;23dup]`.

## Three answers for one input

| | output | verdict |
|---|---|---|
| `origin/main` | `c.[*11del;*11dup]` | `parse_hgvs` rejects it (self-cancelling) |
| #1284 alone | `c.*11=` | correct |
| with #1327 | `c.[*11del;*11delinsAA]` | parses, but strict rejects; overlapping |

#1284 reached the correct identity *through an out-of-range intermediate*: it anchored the repair at `c.*11_*12`, past the transcript end, and the `del` then cancelled the `ins`. #1327 correctly removes that out-of-range anchor — and for the sibling `dup`+`dup` shape it lands the genuinely better `c.*10_*11A[4]`, closing #1343 — but on the cancelling shape the in-range `*11delinsAA` now collides with the `del` instead of cancelling it.

So neither change handles this input properly; #1284 got the right answer by luck.

## Why this is filed rather than patched

[...]

PR #1365, opened 2026-08-04
fix(normalize): stop a terminal duplication respelling past the contig end

### LINKED ISSUE #1307: Normalizer: a terminal duplication respells to an insertion past the end of the contig
`respell_colliding_duplications` (#1277) re-spells a duplication that collides with a sibling as an insertion at the junction 3' of the duplicated span — the gap `[end, end + 1]`. When the duplication ends at the contig's **last** base there is no such gap, and `respell_at_gap` places the member past the end of the sequence anyway:

```text
reference  TTTTTTTTTAATATATTTTAATAC     24 bases
in   g.[24dup;24C>G]
out  g.[24C>G;24_25insC]                position 25 does not exist
```

`respell_at_gap` guards only that the member *landed* where it was asked to (`member_span` reads back `[end, end + 1]`), never that `end + 1` is within the provider's sequence. So the `landed` check passes and the out-of-bounds spelling is emitted. The result re-parses, so `FERRO_ASSERT_REPARSE` does not catch it either.

Reproduced on `origin/main` at 46680b3 (`fix(normalize): canonicalize cis alleles on the transcript axes (#1243)`) with a `MockProvider` single-contig template; not a regression from any in-flight branch.

**Suggested fix.** Bound `gap_end` against `provider.get_sequence_length(&dup.provider_key)` before the `landed` check — decline the respelling when `span.end` is already the final base, leaving the member as the duplication it was. The interior behavior must be preserved. Two shapes worth covering: the terminal genomic case above, and the `m.` circular case, where wrapping rather than declining may be the right answer.

Surfaced by CodeRabbit as an outside-diff comment on #1285; filed separately because the region is untouched by that PR.

nh13:
Closes #1307

`respell_colliding_duplications` repairs a duplication that collides with a sibling by re-spelling it as an insertion at the junction 3' of the duplicated span. A duplication resting on the contig's **last** base has no such junction, and the member was placed one past the end anyway:

```text
reference  TTTTTTTTTAATATATTTTAATAC     24 bases
in   g.[24dup;24C>G]
out  g.[24C>G;24_25insC]                position 25 does not exist
```

## Why #1327's bound did not already cover it

#1327 added the terminal bound; #1344 then disabled it whenever a base-claiming sibling occupies the base the gap runs past. That is right for the case #1344 measured, where the sibling is a **deletion**: the pair is then the #999 adjacency the collapse pass merges, so the out-of-range coordinate is consumed before anything renders, and re-spelling instead produced the overlapping `c.[*11del;*11delinsAA]`.

Nothing absorbs the junction form when the sibling is *not* a deletion. The coordinate is no longer transient — it reaches the output.

So the overrun needs three answers rather than two, which the renamed `TerminalOverrun` now names explicitly:

| variant | when | why |
| --- | --- | --- |
| `RespellAtBoundary` | nothing else claims the last base | the boundary `delins` denotes the same bases, in range (#1327) |
| `WriteTransientJunction` | a sibling absorbs a flush insertion | the collapse merge consumes the coordinate (#1344) |
| `LeaveMemberUnchanged` | a sibling claims the last base but does not absorb | both other answers are wrong (#1307) |

## Why declining, and not the single-member form

`g.23_24insG` is in range, is one member and denotes the same sequence — and it is unreachable **by policy**. It requires dropping `normalize_allele`'s pre-merge reorder guard for inputs that already carry an overlap conflict, and both shapes here are such an input (`24dup` + `24C>G` is what strict mode rejects as `OverlapConflictingEdits / W5002`). Measured: dropping that guard does produce `g.23_24insG`, and breaks four committed guards that exist to keep a conflicting allele recognizable as one rather than canonicalizing it into something that looks valid:

- `issue_395_overlap_conflict_strict_rejection::lenient_mode_preserves_authored_order_of_conflicting_members`
- `issue_1276_dup_junction_overlap::lenient_output_of_a_conflict_is_still_rejected_by_strict`
- `issue_1235_transcript_axes::overlap_conflicting_allele_is_not_canonicalized`
- `idempotency_tests::test_overlap_conflicting_allele_is_idempotent_across_respellings`

That makes the old behaviour worse than the issue reports: `g.[24C>G;24_25insC]` is strict-**acceptable**, so a conflict was laundered into a valid-looking description that also named a position off the end. Declining restores both properties — in range, and still a conflict for strict mode to reject.

## A second live instance

Found while measuring the fix, and covered here rather than filed separately since it is the same gate and the same defect: a `delins` sibling. `g.[24dup;24delinsGG]` gave `g.[24delinsGG;24_25insA]`. It *removes* its bases and still does not absorb the payload, so the predicate is `merges_with_a_flush_insertion` — a plain deletion only — and not "removes bases". `NPaddedDeletion` and `Repeat`/`MultiRepeat` are excluded for reasons documented on the predicate; each exclusion costs at most an unrepaired collision, never an out-of-range coordinate.

## Corrects an assumption in #1307 and #1353

Both say this class is invisible to the existing oracles. `FERRO_ASSERT_REPARSE` is indeed blind — the output re-parses. `FERRO_ASSERT_IDEMPOTENT` is **not**; the out-of-range output is not a fixed point. Measured against `origin/main`:

```text
input: NC_TEST.1:g.[24dup;24C>G]
once:  NC_TEST.1:g.[24C>G;24_25insC]
twice: NC_TEST.1:g.23_24insG
```

No test normalized this shape, so it is a corpus gap rather than an oracle gap — worth knowing when #1353 is picked up, since the cheap guard already existed.

## Tests

`tests/it/issue_1307_terminal_dup_respell.rs`, eight tests. Outputs are pinned, not merely bounded — a bounds check alone is satisfied by a refusal or a dropped member, which is how #1344 hid:

- the issue's reproducer, both member orders
- the `delins` sibling instance
- the same overrun on the circular `m.` axis (the wrapped `ins` spelling is not valid HGVS, per #1327's pinned reason, so `m.` gets the same answer as `g.`)
- **controls:** a deletion sibling at the last base still merges to `g.23=` (the #1344 path, on the genomic axis this time); an in-range collision still collapses; a lone terminal duplication is untouched
- the two properties that justify the choice: the output is idempotent, and strict mode still rejects it

## Verification

- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 cargo nextest run --features dev` — 9194/9194 pass, 146 skipped

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of terminal duplication collisions during variant normalization.
  - Preserved valid substitution and combined insertion/deletion variants at sequence boundaries.
  - Continued collapsing eligible deletion-based collisions while rejecting invalid overlap conflicts.
  - Improved behavior for both linear and circular sequences.

- **Tests**
  - Added comprehensive coverage for terminal collisions, unchanged variants, idempotency, and coordinate validity.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -3899,6 +3899,9 @@ struct MemberSpan {
     end: i64,
     /// Whether the edit consumes the reference bases under its span.
     claims_bases: bool,
+    /// Whether an insertion landing flush against this edit is the adjacency the
+    /// collapse pass merges — see [`merges_with_a_flush_insertion`].
+    absorbs_a_flush_insertion: bool,
     /// Whether the edit must stop a sibling's shift at its boundary. Superset of
     /// `claims_bases` — see `blocks_sibling_shift`.
     blocks_shift: bool,
@@ -3918,6 +3921,7 @@ fn member_span(v: &HgvsVariant, kind: CisKind) -> Option<MemberSpan> {
         start,
         end,
         claims_bases: claims_reference_bases(edit),
+        absorbs_a_flush_insertion: merges_with_a_flush_insertion(edit),
         blocks_shift: blocks_sibling_shift(edit),
         junction: junction_of(edit, start, end),
     })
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #1367, opened 2026-08-04
fix(normalize): validate the reference base an identity member states

### LINKED ISSUE #1352: stated_reference_bases_match omits NaEdit::Identity, while its own doc comment claims the list is complete
`stated_reference_bases_match` in `src/normalize/merge.rs` validates submitter-asserted reference bases before the sequence-first pass canonicalizes a cis allele. Its doc comment states the completeness requirement explicitly, and explains why it matters:

> Every `NaEdit` channel that can carry submitter-asserted reference bases starting at `s` is listed here. `canonicalize_edit` strips most of them before this pass runs, so several arms are inert today — but a validator that silently ignores half the channels it claims to cover is one pipeline reordering away from letting a reference mismatch be canonicalized out from under strict mode.

`NaEdit::Identity { sequence: Some(seq), .. }` **does** carry asserted bases — `c.263A=` asserts that position 263 is `A` — and it is not in the list. It falls through to `_ => None`, so no assertion is checked.

## Why it is not currently exploitable, and why that is the problem

`collect_canonical_edits` refuses `NaEdit::Identity` at its catch-all (`_ => return None`), so an allele carrying an identity member never reaches the validator at all. The safety therefore depends on a **refusal in a different function**, which is exactly the coupling the doc comment warns about — not on this validator covering its stated surface.

## How it was found

While prototyping a change that adds an `Identity` arm to `collect_canonical_edits` (so the derivation skips a member that denotes no change instead of refusing the whole allele). That change makes the gap live immediately: a wrongly-stated `c.263T=` on a reference `A` would be accepted and the member silently dropped, with no `RefSeqMismatch`. The prototype was reverted for unrelated reasons — two further gates refuse behind that one, so it changed no output — but the validator gap is real independent of it.

## Suggested fix

Add `Identity` to the existing alternation, beside the `Deletion`/`Duplication`/`Inversion`/`Delins` arms that already bind `sequence: Some(seq)`:

```rust
| NaEdit::Identity {
    sequence: Some(seq),
    ..
}
```

It needs no behaviour change alongside it and is inert until something upstream stops refusing identity members — which is the point: the validator should be correct **before** the refusal is relaxed, not as part of relaxing it.

A test would need to reach the validator, so it likely has to pair with the `collect_canonical_edits` arm; alternatively assert at unit level that a wrongly-stated identity base is rejected.

nh13:
Closes #1352

`stated_reference_bases_match` guards the sequence-first cis canonicalization against a member whose submitter-asserted reference base disagrees with the reference: that is a reference mismatch, and it belongs to the per-member pipeline's strict-reject / warn path rather than to canonicalization. Its doc comment claims the channel list is complete, and says why that matters:

> a validator that silently ignores half the channels it claims to cover is one pipeline reordering away from letting a reference mismatch be canonicalized out from under strict mode.

`NaEdit::Identity { sequence: Some(..) }` does carry an assertion — `c.263A=` states that position 263 is `A` — and was not in the list, so it fell through to `_ => None` and nothing was checked.

## Why this is worth fixing while it changes nothing

Verified inert, and the mechanism is the point. `collect_canonical_edits` runs first (line 1896 against the validator's 1930) and refuses `Identity` at its catch-all, so no identity member reaches the validator today. The validator's correctness therefore rests on a refusal in a **different function** rather than on its own coverage — exactly the coupling its doc comment warns about. Relaxing that refusal (as #1351's work on identity members may want to) would make the gap live in the same commit that introduces it.

Full suite is unchanged at 9194/9194, which is the claim rather than a caveat.

## Matched with `..`, not `whole_entity: false`

Nothing the parser produces gives a whole-entity identity (`c.=`) a sequence. If one ever arrives with bases, comparing them is the fail-safe answer — a mismatch refuses canonicalization — whereas restricting the arm to `whole_entity: false` would silently skip it, which is the shape of the hole being closed.

## Tests

The validator is private and the arm is unreachable end to end, so four unit tests in `merge.rs` cover it directly:

- a correctly-stated identity base is accepted, a wrongly-stated one rejected (`g.3A=` vs `g.3T=` on `GGATTAC`)
- a bare `g.3=` asserts nothing and stays neutral — the arm keys off `Some(..)`
- the four pre-existing channels that share the new arm's binding still match, both ways (sub / del / dup / delins)
- an assertion running past the window is a refusal, not a panic

**Verified the tests fail without the fix**: removing the arm again turns `a_wrongly_stated_identity_base_is_rejected` and `a_stated_identity_past_the_window_is_refused` red, so they test the change rather than merely accompany it.

## Verification

- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 cargo nextest run --features dev` — **9194/9194 pass**, 146 skipped

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of asserted reference bases in sequence-bearing identity edits.
  * Correctly handles descriptions such as `c.263A=`.
  * Preserves support for bare identities without sequence data.
  * Added safeguards for out-of-window references and existing reference validation paths.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2224,8 +2224,8 @@ fn collect_canonical_edits(
 /// Whether every member's *stated* reference base agrees with the window.
 ///
 /// `NaEdit` optionally carries the reference bases the submitter asserted
-/// (`c.5A>T`, `c.5_7delACG`). Where present they must match, otherwise the
-/// variant is a reference mismatch and belongs to the per-member pipeline's
+/// (`c.5A>T`, `c.5_7delACG`, `c.263A=`). Where present they must match, otherwise
+/// the variant is a reference mismatch and belongs to the per-member pipeline's
 /// strict-reject / warn path rather than to canonicalization.
 fn stated_reference_bases_match(
     variants: &[HgvsVariant],
@@ -2245,7 +2245,27 @@ fn stated_reference_bases_match(
         // canonicalized out from under strict mode.
         let stated = match edit {
             NaEdit::Substitution { reference, .. } => Some(vec![*reference]),
-            NaEdit::Deletion {
+            // `c.263A=` asserts that position 263 is `A`, so an identity that
+            // spells its sequence is a stated-reference channel like any other
+            // (#1352). It was missing while the doc comment above claimed the
+            // list was complete, and the safety therefore rested on a refusal in
+            // a *different* function — `collect_canonical_edits` declines
+            // `Identity` at its catch-all — which is exactly the coupling that
+            // comment warns against. Inert until something upstream stops
+            // refusing identity members, which is the point: the validator
+            // should be correct before the refusal is relaxed, not as part of
+            // relaxing it.
+            //
+            // `..` rather than `whole_entity: false`: a whole-entity identity
+            // (`c.=`) carries no sequence in anything the parser produces, but if
+            // one ever arrives with bases, comparing them is the fail-safe
+            // answer — a mismatch refuses canonicalization, whereas skipping the
+            // arm is the very hole being closed.
+            NaEdit::Identity {
+                sequence: Some(seq),
+                ..
+            }
+            | NaEdit::Deletion {
                 sequence: Some(seq),
                 ..
             }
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
## Merge-order note: this PR and #1366 touch the same anchor in `merge.rs`

Measured with `git merge-tree --write-tree --name-only` across all twelve PRs in this batch — the technique the repo notes prescribe, because `mergeStateStatus` lags a merge by minutes and both PRs currently report clean against `main`:

```
#1366 <-> #1367 : src/normalize/merge.rs
```

That is the **only** pairwise conflict among the twelve. Both PRs are clean against `origin/main` individually, so whichever merges **second** will need a one-line resolution.

### Why, and how to resolve it

Both add a test block immediately after `mod tests {`'s `use` lines (`@@ -6054`), so the collision is at that insertion point only. Their production changes do not overlap at all:

| PR | production hunks |
| --- | --- |
| #1366 | `@@ -5788` — the in-bounds predicate, after `cds_axis_end` |
| #1367 | `@@ -2224`, `@@ -2245` — the `Identity` arm in `stated_reference_bases_match` |

**Resolution: keep both test blocks.** They are independent additions in the same place, with no shared symbols — nothing needs reconciling beyond ordering them.

No action needed before merge; this is only so the second merge is not a surprise.

nh13:
## Rebased onto current `main` after #1366 merged — and the "keep both test blocks" plan needed one correction

#1366 merged at 17:30 UTC (`10ec7a0`), which triggered the collision predicted in my earlier comment: both PRs add a test block at the same `mod tests {` anchor in `src/normalize/merge.rs`. Rebased `337585b` → `10f77e3`.

The production hunks did not overlap, exactly as measured — #1366's in-bounds predicate sits near `cds_axis_end`, this PR's change is the `Identity` arm of `stated_reference_bases_match`. Only the tests collided.

**But "keep both test blocks" was not sufficient as literally stated, and I want that on the record.** There were **two** conflict regions, not one:

- The second was a plain keep-both — both sides are complete `#[test]` functions, so dropping the markers is the whole resolution.
- The first was **not**. Both sides end mid-`assert!()`/`assert_eq!()`, with the closing `);` / `}` / `}` sitting *after* the conflict region as shared text. That trailing closer can only terminate whichever block is emitted last, so concatenating the two sides leaves main's `#1353` block unterminated. It needed its own closer inserted between them. A naive marker deletion would not have compiled.

Resolved with a script rather than by hand, to keep line arithmetic out of it, then verified both sides survived rather than assuming:

```
normalize::merge::tests::a_five_prime_overrun_is_flagged_too ................. PASS   (#1366)
normalize::merge::tests::an_authored_overrun_is_visible_through_a_special_position  PASS   (#1366)
normalize::merge::tests::a_stated_identity_past_the_window_is_refused ........ PASS   (this PR)
```

Both `#1353` helpers (`bounds_genomic_provider`) and this PR's (`stated_bases_agree`) are present, and both section headers survive.

### Verification

- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 cargo nextest run --features dev` — **9227/9227 pass**, 146 skipped

No content changed — same work, replayed on the new base, still signed.

nh13:
@coderabbitai review

Issue #1218, opened 2026-07-26
ci: cargo caches never refresh, and the 10 GiB budget blocks the obvious fix


nh13:
PR CI's cargo caches never refresh, so several jobs rebuild from scratch on every run. The obvious fix — rotating cache keys — was implemented, measured, and reverted, because the repository's Actions cache budget cannot currently sustain it. This issue records the evidence so the next attempt starts from it.

## The defect

`actions/cache` **does not update an entry it hit on its primary key**. Every cargo cache key in `ci.yml` is static (`${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-<suffix>`), so each entry freezes at whatever the *first* run for a given `Cargo.lock` produced and never improves.

Observed on the `Build` job:

```
Cache hit for: Linux-cargo-554cd5b…-release
Cache hit occurred on the primary key …, not saving cache.
```

That `-release` entry was **0.17 GiB**, captured from an early near-empty build tree, while a healthy **1.01 GiB** entry sat unused under a different lockfile hash. `cargo build --release` therefore restored almost nothing and rebuilt every run: **142s**.

## Why rotation alone does not work here

Rotating the key on `github.sha` with prefix `restore-keys` was tried (`bf70635`). It fixed the freezing — entries immediately grew to realistic sizes:

| entry | static key | rotated |
|---|---|---|
| `-release` | 0.17 GiB | 1.57 GiB |
| `-clippy-all-features` | 0.24 GiB | 1.69 GiB |
| `-test-nodebug` | 0.86 GiB | 1.57 GiB |
| `-wheel` | 0.19 GiB | 1.73 GiB |

And with a genuinely warm cache the wins were large — measured on a re-run that restored the rotated entries: `Build` 159s → 87s, `Python Wheel Test` 138s → 88s, `Build test archive` 130s → 87s.

But the budget cannot hold two generations. GitHub caps a repository at **10 GiB**; this repo was already at **9.53 GiB** before any change, and each CI run mints ~6.8 GiB of fresh cargo entries. After two runs usage reached **13.61 GiB**, GitHub evicted by LRU, nothing stayed warm between distinct commits (wall time went 300s → 339s), and the nightly's **3.25 GiB `ferro-manifest-v3`** reference-data cache was evicted as collateral damage to an unrelated workflow.

Reverted in `b99bf2c`.

## What has to happen first

Reclaim budget. In rough order of value:

[...]

Issue #933, opened 2026-07-04
test(conformance): provision Ensembl reference & re-bless the 5 ENST/ENSG normalize rows


nh13:
## Summary

Five mutalyzer-normalize conformance rows for Ensembl transcripts (`ENST`/`ENSG`) currently no-op (echo the input) and are dispositioned `reference_unavailable` with reason `ensembl-absent-from-refseq-manifest`. The underlying **code** to resolve them has already shipped:

- **#671** (Ensembl reference capability — `ferro prepare --ensembl` fetches Ensembl cdot + cDNA FASTA, `MultiFastaProvider` resolves `ENST`/`ENSG`/`ENSP`) — CLOSED/COMPLETED.
- **#670** (apply the 3′ rule across exon/intron boundaries for genomic-context `c.`/`n.` variants, which two of these rows additionally depend on) — CLOSED/COMPLETED.

The rows still no-op only because **the conformance/prepared reference is RefSeq-only**: `ferro prepare --ensembl` is opt-in by design (#671 kept the default manifest small), so the manifest's `ensembl_cdot_json` / `ensembl_transcript_fastas` are unpopulated. This is now a **reference-provisioning gap**, not a code gap — and the dispositions still cite the (completed) #671 as if code work were pending. This issue is their live successor tracker.

## Affected rows (axis = normalized)

| input | expected `normalized` |
|---|---|
| `ENST00000375549:c.100del` | `ENST00000375549.8:c.102del` |
| `ENST00000375549.8:c.100del` | `ENST00000375549.8:c.102del` |
| `ENSG00000204370.13(ENST00000375549.8):c.100del` | `ENSG00000204370.13(ENST00000375549.8):c.102del` |
| `ENST00000452863.10:c.9del` | `ENST00000452863.10:c.10del` |
| `ENSG00000184937.16(ENST00000452863.10):c.9del` | `ENSG00000184937.16(ENST00000452863.10):c.10del` |

(The `c.100del`→`c.102del` and `c.9del`→`c.10del` pairs are the exon/intron-boundary 3′-shift cases that depended on #670.)

## Acceptance criteria

- [ ] Provision the conformance/prepared reference with Ensembl data via `ferro prepare --ensembl` (Ensembl cdot 0.2.32 GRCh38 + cDNA FASTA, matching the pinned RefSeq release), populating `ensembl_cdot_json` / `ensembl_transcript_fastas` / `available_prefixes` in the manifest.
[...]

Issue #875, opened 2026-06-26
ci(nightly): derive ng_hosted_transcripts (+ NG_ placements) so the nightly reference matches the blessed reference


nh13:
Follow-up to #859/#871. The nightly conformance/perf workflows run `ferro prepare --output-dir benchmark-output --genome grch38` **without** `--derive-ng-placements`, so their prepared reference carries neither the derived NG_ placements nor `ng_hosted_transcripts`. Consequences:

- `ng_hosted_transcripts` is unset → the #793 legacy-selector→NG_-hosted-version resolution is dormant in nightly, so the #859 rows that PR #871 re-blessed to converged (normalized 12→0) show up as **drift** in the nightly xfail report.
- More broadly, the nightly reference lacks the derived RefSeqGene placements the blessed reference carries, so it is not a faithful reproduction.

This is **non-gating**: PR CI does not provision a manifest (the reference-aware axis tests skip), and the nightly is `continue-on-error`. So this does not block any PR — it only makes the nightly drift report noisy/unfaithful.

## Proposed fix

- Commit a versioned NG_ accession list (`tests/fixtures/mutalyzer-normalize/ng_accessions.txt`) covering the conformance corpus.
- Add `--derive-ng-placements <that list>` to the `ferro prepare` step in **both** `nightly-mutalyzer.yml` and `nightly-perf.yml` (they share the manifest cache).
- Bump `MANIFEST_CACHE_VERSION` (v1→v2) in both files (kept in sync per their comments) so the next nightly re-prepares with the new flag.

## Caveats to weigh in review

- The derive performs a GenBank (eutils) network fetch per NG_ accession; CI reliability is unproven (mitigated by the cache + non-gating nightly).
- The committed accession list can drift from the corpus; a future `--check` (mirroring `generate_conformance_summary --check`) could enforce coverage.

Part of #326 (nightly fidelity for the #859/#871 convergence).

PR #1369, opened 2026-08-04
ci: drop two 3 GiB nightly caches that never once hit

nh13:
Refs #1218 — **step 1 of 3.** Does not close the issue; the key rotation it asks for lands in step 3, once the footprint fits two generations.

#1218's blocker is budget, and it names reclaiming the nightly's 3.25 GiB reference cache as the highest-value item. Measured: that cache has **never once hit**.

## The measurement

`nightly-mutalyzer.yml` cached `benchmark-output/` (the prepared reference — genome FASTAs + cdot, ~3.25 GiB) under `Linux-ferro-manifest-v3-<hash>`.

| run | date | manifest cache | key |
| --- | --- | --- | --- |
| 30879794391 | 2026-08-04 | `Cache not found` | `…-cad991a1c119` |
| 30786830075 | 2026-08-03 | `Cache not found` | `…-78073890138b` |
| 30733613038 | 2026-08-02 | `Cache not found` | `…-78073890138b` |
| 30685312545 | 2026-08-01 | `Cache not found` | `…-78073890138b` |
| 30606298814 | 2026-07-31 | `Cache not found` | `…-78073890138b` |

Four of five asked for the **identical key** and still missed, so this is LRU eviction, not key churn. Watched directly on the most recent run: it saved its entry at `05:43:04` and the entry was already gone from `actions/caches` by `05:57` — the same fifteen-minute disappearance the layout notes recorded for PR #1300.

Current state, for scale:

```
active_caches_size_in_bytes: 12630256757   # 11.76 GiB against a 10 GiB cap
```

So the entry consumed a third of the repository budget, evicted PR CI's cargo caches as collateral — exactly the failure mode #1218 records from the reverted `bf70635`, including the nightly-cache-destroyed-by-PR-CI direction — and never saved this workflow a single `ferro prepare`.

## Why removing it is free

The `ferro prepare` step was guarded by `if: steps.cache-manifest.outputs.cache-hit != 'true'`. Since the cache never hit, that step **already ran every night** (measured: 05:13:23 → 05:43:04, ~30 min). Dropping the cache does not slow the workflow by a second; it only stops the workflow taking budget from workflows that can use it.

`nightly-perf.yml` carries the same pattern in its deliberately separate RefSeq-only entry (`Linux-ferro-manifest-refseq-…`); run 30686892059 logged the same miss, and it gets the same treatment. Together that frees up to **~6.5 GiB**.

## What is preserved

Both `MANIFEST_CACHE_VERSION` variables go with the keys they fed, but their rationale is kept in comments — deliberately, because it matters to whoever restores the cache:

- the `v2`/`v3` history (`--derive-ng-placements` for #875, `--ensembl` for #933);
- and above all **why the two nightlies kept separate caches**: the perf workflow's startup-budget gate measures RefSeq `ferro normalize` startup (#585), and an Ensembl-inclusive reference makes a RefSeq startup eagerly merge ~198k Ensembl transcripts and inflates that baseline (#933). If either cache comes back, they must stay separate.

The comments also name the alternative worth weighing: Actions **artifacts** are quota'd and billed separately from caches, so they may be the right home for a 3 GiB reference. That has storage-cost implications, so it is flagged rather than done here.

## A finding: the issue's third item is already done

#1218 lists "drop `target/` from the cache for jobs that no longer build — the sharded test jobs consume a `nextest archive` and do not compile." Verified against every job in `ci.yml`:

```
no-cache  test          no-cache  soak           no-cache  test-required
no-cache  test-oracle   no-cache  sweeps
CACHE     clippy        CACHE     clippy-all-features   CACHE  spec-fixtures
CACHE     test-build    CACHE     test-build-soak       CACHE  generated-docs
CACHE     python-wheel-test        CACHE  build
```

Every job that caches also compiles, so there is nothing left to drop there. The eight compiling jobs each cache the same five paths including a full `target/` — that is 5b's subject.

## Also observed, feeding step 2

The nightly's own cargo cache restored **`Linux-cargo-…-coverage`** via its bare `${{ runner.os }}-cargo-` restore-key — a coverage-instrumented `target/` tree served to a plain release build. That is the cross-profile mismatch the layout notes warn about, and four jobs still carry that bare restore-key. Left alone here to keep this change to budget reclamation; it belongs with the key restructuring in step 3.

## Verification

- `actionlint` clean on both workflows; both parse as YAML.
- No Rust or test changes, so no suite run applies. The behavioural claim — that `ferro prepare` already ran unconditionally — is from the run logs quoted above rather than from reasoning about the `if:`.

### DIFF
--- a/.github/workflows/nightly-mutalyzer.yml
+++ b/.github/workflows/nightly-mutalyzer.yml
@@ -31,16 +31,16 @@ concurrency:
 env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: 1
-  # Bump this tag whenever the manifest layout or `ferro prepare`
-  # invocation changes meaningfully, so stale caches are invalidated
-  # without manual purging. (v2: added --derive-ng-placements so the
-  # prepared reference carries derived NG_ placements + ng_hosted_transcripts,
-  # matching the blessed reference; #875. v3: added --ensembl so the reference
-  # provisions Ensembl cdot + cDNA, matching the Ensembl-inclusive blessed
-  # reference the corpus identity is recorded against; #933.) nightly-perf.yml
-  # keeps its OWN RefSeq-only cache (its `-refseq-` key) and does NOT provision
-  # Ensembl, so the two caches deliberately diverge — do not re-sync them (#933).
-  MANIFEST_CACHE_VERSION: v3
+  # NOTE: `MANIFEST_CACHE_VERSION` is gone along with the manifest cache it
+  # keyed — see the block above the `ferro prepare` step for the measurement
+  # that removed it (#1218). Its history is worth keeping for whoever restores
+  # the cache: v2 added `--derive-ng-placements` so the prepared reference
+  # carries derived NG_ placements + `ng_hosted_transcripts`, matching the
+  # blessed reference (#875); v3 added `--ensembl` so the reference provisions
+  # Ensembl cdot + cDNA, matching the Ensembl-inclusive blessed reference the
+  # corpus identity is recorded against (#933). This workflow provisions
+  # Ensembl and nightly-perf.yml deliberately does not — if both caches ever
+  # come back, they must stay separate (#933).
 
 jobs:
   reference-aware-tests:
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #1370, opened 2026-08-04
ci: delegate cargo caching to rust-cache instead of hand-rolled keys

nh13:
Refs #1218 — **step 2 of 3.** Follows #1369 (budget reclamation) and does not close the issue.

#1218 recommends this directly:

> `Swatinem/rust-cache` prunes far more aggressively and manages keys itself; it is probably the right answer rather than hand-rolled keys plus a hand-rolled trim.

All nine cargo cache steps — eight in `ci.yml`, one in `coverage.yml` — cached the same five paths including a whole `target/`. Measured on 2026-08-04:

| entry | size |
| --- | --- |
| `Linux-cargo-<lock>-wheel` | 3.04 GB |
| `Linux-cargo-<lock>-clippy-all-features` | 3.00 GB |
| `Linux-cargo-<lock>-coverage` | 2.87 GB |
| `Linux-cargo-soak-<hash>` | 0.26 GB |

Three entries from a **single PR** occupying 9 GB, against a reported `active_caches_size_in_bytes` of 11.76 GiB — over the 10 GiB cap and already evicting. #1218 records that a hand-rolled trim (dropping `incremental/`, own artifacts, `.d` files) was tried and "did not move the needle", because the bulk sits under `target/*/deps`. `rust-cache` saves dependency artifacts only, removing the workspace crate's own outputs, `incremental/`, `.d` files and stale entries before saving.

## It also mostly dissolves what step 3 was for

`actions/cache` skips the save on a primary-key hit — the freezing defect #1218 opens with, which rotating keys was meant to fix and which the budget would not allow. With only dependency artifacts cached, and `rust-cache` already hashing the dependency set, rustc version and job into its key, a frozen entry cannot disagree with what a job needs the way a frozen whole-`target/` could. **Step 3 therefore becomes verification and documentation rather than key rotation** — which is the outcome the budget could never afford before.

## A live hazard removed

Four jobs still carried a bare `${{ runner.os }}-cargo-` last-resort restore-key, a prefix of *every* cargo cache in the repo. That was not theoretical: last night's nightly logged

```
Cache restored from key: Linux-cargo-7a12a397…-coverage
```

— a coverage-instrumented `target/` served to a plain release build. Those restore-keys go with the hand-rolled blocks.

## Two inputs deliberately unset

Both documented at the call site rather than left to inference:

- **`cache-on-failure`** — would save a tree per failed run, and budget is the entire constraint this change exists to relieve.
- **`shared-key`** — would have jobs share one entry, reintroducing exactly the cross-profile mismatch that had `generated-docs` (dev profile) and `test-build` (`CARGO_PROFILE_TEST_DEBUG=0`) serving each other trees neither could use. `rust-cache` keys per job by default.

`key` *is* set per job, purely so `gh api /repos/.../actions/caches` stays legible — reading that listing is how #1218 was diagnosed in the first place, and opaque entry names would cost the next investigation.

## What was preserved

The hand-rolled steps carried long comment blocks (one was 49 lines) recording the primary-key-never-re-saves behaviour and the cross-profile hazard. Those are not deleted — they are consolidated into one note on the first `rust-cache` step, because that knowledge is what a future reader needs even though the mechanism it described is gone.

Net **−167/+88** lines.

## Verification, and its limit

`actionlint` clean on both files; both parse as YAML. Cache *sizes* cannot be measured locally — they are a property of a CI run. I will check this branch's entries with

```sh
gh api "/repos/fulcrumgenomics/ferro-hgvs/actions/caches?per_page=100" \
  --jq '.actions_caches[] | "\(.size_in_bytes) \(.key)"'
```

once its jobs have saved, and post the before/after here. Note the first run after any key change is a **miss** by construction, so the run *after* it is the one whose timings mean anything — never read a caching change from a single run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,18 +171,41 @@ jobs:
         with:
           toolchain: stable
           components: clippy
-      - uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
+      # Cargo caching is delegated to `Swatinem/rust-cache` rather than
+      # hand-rolled `actions/cache` keys (#1218). Two reasons, both measured.
+      #
+      # **Size.** Every hand-rolled entry cached the same five paths including a
+      # full `target/`, and they had grown to ~3 GiB each: on 2026-08-04 the
+      # `-wheel`, `-clippy-all-features` and `-coverage` entries were 3.04 GB,
+      # 3.00 GB and 2.87 GB, i.e. three entries from a single PR filling 9 GB of a
+      # 10 GiB repository budget (measured usage: 11.76 GiB, already evicting).
+      # `rust-cache` saves dependency artifacts only — it deletes the workspace
+      # crate's own outputs, `incremental/`, `.d` files and stale entries before
+      # saving. #1218 records that a hand-rolled trim was tried and "did not move
+      # the needle", because it kept the bulk under `target/*/deps`.
+      #
+      # **Staleness.** `actions/cache` SKIPS the save when the primary key already
+      # exists ("Cache hit occurred on the primary key …, not saving cache"), so a
+      # lockfile-only key froze `target/` at whatever the first run for that
+      # lockfile happened to write, forever. Rotating the key was the obvious fix
+      # and #1218 records why it could not be afforded. `rust-cache` largely
+      # dissolves the problem instead of paying for it: the only thing it stores is
+      # dependency artifacts, and its key already hashes the dependency set,
+      # rustc version and job — so a frozen entry cannot disagree with what the
+      # job needs the way a frozen whole-`target/` could.
+      #
+      # Deliberately NOT set:
+      # - `cache-on-failure` — it would save a tree per failed run, and budget is
+      #   the constraint this change exists to relieve.
+      # - `shared-key` — each job would then share one entry, which is the
+      #   cross-profile mismatch that had `generated-docs` (dev profile) and
+      #   `test-build` (`CARGO_PROFILE_TEST_DEBUG=0`) serving each other useless
+      #   trees. `rust-cache` keys per job by default; `key` below only adds a
+      #   legible name, so `gh api .../actions/caches` stays diagnosable — which
+      #   is how #1218 was diagnosed in the first place.
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
         with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-clippy
-          restore-keys: |
-            ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-
-            ${{ runner.os }}-cargo-
+          key: clippy
       - run: cargo clippy --features dev --all-targets -- -D warnings
 
   clippy-all-features:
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
## Measured: −31% overall, but not uniformly, and one number needs care

Both schemes are live in the repo right now (this branch uses `rust-cache`; #1368/#1369 still use the hand-rolled steps), so the same jobs can be compared directly. From `gh api /repos/fulcrumgenomics/ferro-hgvs/actions/caches`:

| job | hand-rolled | rust-cache | change |
| --- | --- | --- | --- |
| `coverage` | 1,472,817,285 | 646,475,927 | **−56%** |
| `gendocs` → `generated-docs` | 524,819,740 | 385,431,166 | **−27%** |
| `specgen` → `spec-fixtures` | 462,058,080 | 347,907,930 | **−25%** |
| `test-archive` → `test-build` | 388,335,448 | 290,613,498 | **−25%** |
| `soak` → `test-build-soak` | 262,502,813 | 269,616,835 | +3% |
| `wheel` → `python-wheel-test` | 213,955,847 | 234,247,414 | +9% |
| `clippy-all-features` | 258,088,492 | 293,786,705 | +14% |
| `clippy` | 187,080,299 | 231,235,572 | +24% |
| `release` → `build` | 184,999,899 | 228,480,455 | +24% |
| **total** | **4,217,143,728** | **2,927,795,502** | **−31%** |

Total usage right now is 10,208,881,685 bytes across 28 entries — 9.51 GiB against the 10 GiB cap, with both schemes resident at once. That is the number the budget cares about, and −31% on the cargo set is what this change buys.

### The four entries that got *bigger* are the interesting ones

I am not going to present +24% as a win. Two readings, and I think the second is right:

1. `rust-cache` also stores `~/.cargo/bin` and a fuller registry than the hand-rolled `path:` list did, which costs some bytes back on jobs whose `target/` was small to begin with.
2. More likely, and more to the point of #1218: **the small hand-rolled entries were small because they were frozen.** `clippy`, `clippy-all-features`, `wheel` and `release` are exactly the four jobs still on a lockfile-only key with no workflow-hash component — the ones `actions/cache` never re-saves once the key exists. #1218 opens with a measured instance: a `-release` entry stuck at 0.17 GiB, "captured from an early near-empty build tree", while a healthy 1.01 GiB entry sat unused under a different hash. A frozen entry is not a cheap entry, it is a useless one — the job restores it and then recompiles anyway.

If (2) is what is happening, then +24% on `clippy` is the cache finally holding a complete dependency tree, and the thing to watch is not entry size but `grep -c 'Compiling ' <job-log>`, which the layout notes give as the real health check (~1 for a warm Rust build job; hundreds means the cache is not matching what the job builds). That belongs in step 3's verification, along with the second-run timings — the first run after any key change is a miss by construction, so **no timing here should be read yet**.

### Caveat on the `coverage` row

The −56% is measured against a 1.47 GB hand-rolled entry. Earlier in this campaign the same key on PR #1363's ref measured 2.87 GB, so hand-rolled sizes for one job vary by a factor of two across refs — another symptom of entries freezing at whatever a given ref's first run wrote. The direction is not in doubt; treat the exact percentage as approximate.

nh13:
## Following up on the persistence check I promised — the entries did **not** survive

I said in the PR body I would check with `gh api .../actions/caches` once this branch's jobs had saved. Result:

```
$ gh api "/repos/fulcrumgenomics/ferro-hgvs/actions/caches?ref=refs/pull/1370/merge" --jq .total_count
0
```

Zero. The `v0-rust-*` entries this branch saved are gone — the same disappearance #1218 records for PR #1300's rotated entries ("already gone fifteen minutes later → `count: 0`").

### Why, and why it is not evidence against `rust-cache`

Repository-wide, right now:

```
usage: 13828659307 bytes  =  12.87 GiB in 27 entries     (cap: 10 GiB)
```

The cap is exceeded by ~2.9 GiB, so GitHub is evicting continuously by LRU. Every surviving entry belongs to a **more recently touched ref** — #1371, #1372 and #1374, whose branches I force-pushed minutes ago, so their runs are newer than this one's. Nothing about the scheme; purely recency.

So the two claims separate cleanly:

- **Size reduction: measured and holds** — the per-job table earlier in this thread stands (−31% across the cargo set, coverage −56%).
- **Persistence: not demonstrable yet**, and cannot be while the repository sits 2.9 GiB over its cap.

This is exactly #1218's own conclusion, which I should have weighted more heavily up front: a correct key structure is *"necessary but not sufficient"*. Reducing entry size from ~3 GiB to ~0.6 GiB does not help if the total is still above the cap.

### An awkward but relevant admission

A large share of what is currently saturating the budget is **this campaign's own twelve PRs**. Each open PR mints a full set of cargo entries per run, and I have been force-pushing several of them. So the batch is temporarily making the thing it is trying to fix worse, and no persistence measurement taken while the queue is this deep will mean anything.

### What follows for merge order

1. **#1369 first** — it removes ~6.5 GiB of nightly caches that were measured at a 0 % hit rate. That is what creates headroom.
2. **#1370 (this PR) second** — the smaller entries then have somewhere to live.
3. **Step 3 (#1218's remaining work) only after the PR queue drains**, because its definition of done needs *two consecutive runs on `main`* with nothing evicted, and that cannot be measured against a repository full of in-flight branches.

Neither of the first two is blocked by this; the ordering just determines whether the second one's benefit is visible.

PR #1371, opened 2026-08-04
perf(reference): cache decoded sequence blocks under a byte budget

### LINKED ISSUE #976: perf: cache decoded sequence regions (byte-budgeted LRU) and optionally prime before a batch
### Summary

The production FASTA path caches file handles and reads via positional `read_exact_at` (`src/reference/multi_fasta.rs:40`, `:206`), which is already good, but it does not cache decoded sequence regions. Workloads with many variants in the same gene/locus re-read and re-decode overlapping windows repeatedly. A `CachedFastaProvider` exists (`src/reference/fasta.rs:523`) but the production provider doesn't use a region cache.

### Proposal

1. Add a byte-budgeted, `Arc`-shared LRU of decoded regions keyed by `(contig, start, end)` (or coarser fixed-size blocks) on the production provider, with the budget env-tunable and the lock dropped across I/O so parallel workers share lock-free. The `lru` crate is already a dependency.
2. Optional follow-on: a `preload`/prime step that, given the positions in a batch, loads the covering regions/blocks once up front — useful only once the batch pipeline is I/O-bound.

### Trade-offs

- Clear win when access is spatially clustered (panels, single-gene runs); negligible for scattered single lookups. Keep the budget modest and evict by summed decoded bytes.

### Acceptance

- A many-variants-in-one-gene run shows fewer reads / less decode work and lower wall-time; memory stays within the configured budget.
- Scattered-lookup and correctness behavior unchanged.

nh13:
Closes #976

The production FASTA path already caches file handles and reads positionally, but it re-read and re-decoded overlapping windows every time: a workload with many variants in one locus paid the `pread`, the newline strip and the ASCII-uppercase pass once per variant over largely the same bases.

Adds an `Arc`-shared, byte-budgeted LRU of **decoded** blocks on `MultiFastaProvider`, inside `get_sequence_from_index` so every caller benefits — `get_sequence`, `get_genomic_sequence` and the protein path alike.

## Design choices, and why

- **Blocks, not exact `(contig, start, end)` ranges** — #976 leaves this open. Exact ranges only hit when a later request repeats a window verbatim, and the workload this exists for produces windows that *overlap without repeating*. Aligned blocks turn that overlap into hits.
- **8 KiB blocks.** A block read costs more bytes than a narrow request, so a large block would tax the scattered workload this must not wreck; 8 KiB is a couple of filesystem pages the page cache would have faulted in anyway, while still covering an exon and its flanks in one entry. What a hit actually saves is the decode pass.
- **Keyed `(file_id, record_offset, block_index)`** — no string hashing on the hot path, and aliases (`chr1` / `NC_000001.11`) cannot cache the same bases twice under two names.
- **Decoded bytes stored.** Safe because `get_sequence_from_index` has always uppercased unconditionally, so no caller wants the original soft-masked case — and pinned by a test, since storing raw bytes would have let the first read of a lowercase stretch poison every later read of it.
- **`FERRO_SEQUENCE_CACHE_BYTES`**, default 64 MiB, `0` disables. Eviction by **summed decoded bytes**, as the issue asks, so `lru` is unbounded in entries and the byte total is the real bound.
- **The lock is dropped across the read**, as the issue asks. Two threads missing the same block both read it; the loser's insert overwrites an identical value and the accounting handles it. Duplicate work on a cold block beats holding a global lock across a `pread`.

## Measurement, on a real prepared reference

50 000 overlapping 4 kb duplications in one locus on `NC_000001.11`, against ~0.73 s of startup measured separately:

| | rep 1 | rep 2 |
| --- | --- | --- |
| cache off | 1.205 s | 1.234 s |
| cache on | 0.933 s | 0.924 s |

**−24 % wall clock**, and roughly **2.5×** on the sequence work once startup is subtracted. Outputs byte-identical (`cmp` on 50 000 lines), exit 0, line count checked — a timing wrapper prints a number even when the command fails, so those were verified rather than assumed.

### The scattered case is slightly worse, not "negligible"

#976 predicts "negligible for scattered single lookups". Measured, it is a small **cost**: 50 000 duplications spread across the whole of chr1 gave 1.263 s off against 1.297 s on, about **+2.7 %**, consistently in both reps. That is the block scheme fetching more bytes than a narrow request needs, plus insert/evict overhead. Outputs again byte-identical.

That is the trade the block size buys. `FERRO_SEQUENCE_CACHE_BYTES=0` turns it off, and the block size is a constant chosen by reasoning rather than by a sensitivity sweep — the obvious knob if the scattered case ever matters more than the clustered one.

### Memory

Peak RSS was **1239 MiB at budgets of 0, 8 MiB and 64 MiB alike** — the cache stays inside the process's existing high-water mark (dominated by the reference itself), so it does not raise the peak. The precise bound is a unit test asserting `held_bytes() <= budget` after every read while walking a whole record.

## Tests

Nine unit tests against a purpose-built multi-block fixture (20 000 bases, wrapped at 60, with a soft-masked stretch straddling the first block boundary — the module's existing fixture is 40 bases and lives entirely inside block 0):

- **`cached_reads_match_the_uncached_decoder`** — twelve windows compared both against `decode_range` (the pre-#976 behaviour) *and* against independently computed expected bases: inside a block, straddling a boundary, exactly one block, spanning three, starting/ending on a boundary, the short final block, and past the end (which must clamp, not error).
- **`soft_masked_bases_are_uppercased_through_the_cache`**
- **`clustered_reads_decode_each_block_once`** — the acceptance criterion as an assertion: 20 overlapping windows produce 1 decode and 19 hits.
- **`scattered_reads_are_correct_and_decode_each_block_once`**
- **`the_byte_budget_bounds_the_cache`** — checks the bound after every read while eviction happens under live reads.
- **`a_block_larger_than_the_budget_is_not_stored`** — evicting everything to hold one oversized entry would be worse than not caching it.
- **`a_disabled_cache_reads_directly`**
- **`a_record_without_a_trailing_newline_is_still_readable`** — the exact byte span, checked with the cache both ways.
- **`a_crlf_fasta_reads_the_same_bases_as_an_lf_one`** — a pre-existing short read the exact span fixes; see below.
- **`a_truncated_record_is_an_error_not_a_short_read`** — the fail-closed backstop below.

## The read span is computed exactly, not estimated

`decode_range` sized its buffer as `bases + one byte per line`, which assumes a line terminator *after the last line*. A FASTA whose final record ends without a trailing newline has one byte fewer, so `read_exact_at` failed with "failed to fill whole buffer" even though every requested base was present — and block reads widened that from "a request reaching the final base" to "any request in the final block".

It now locates the last requested base and measures to it, deriving both offsets from the index's `line_bases`/`line_bytes`. On the 20 000-base fixture the estimate wants 3 677 bytes and runs one past a 20 344-byte file; the exact span computes 3 676, ending at precisely EOF.

Three things follow, and the byte saved is the least of them.

**`read_exact_at`'s short-read error stays armed.** The alternative fix — clamping the buffer down to what the file holds — makes the no-trailing-newline case pass by disarming the only thing that catches a genuinely truncated or half-downloaded FASTA, which then reads successfully and yields too few bases. `cached_range` is not a substitute: its bounds guard only fires when a request starts *past* a short block's end, so a partially short block slips through (reading 17 900..18 100 against the truncated fixture returns 134 bases instead of 200). The decoded-count check is kept as a backstop, but it is no longer the sole guard.

**No `fstat` on the hot path.** Clamping requires re-reading `file.metadata()` on every decode — an `fstat` added to the uncached path this PR exists to speed up. The exact span needs none.

**CRLF is fixed, and that bug predates this PR.** A samtools `.fai` for a CRLF FASTA carries `line_bytes = line_bases + 2`, and `prepare`'s own index writer produces the same. The per-line estimate is then wrong by about a byte per line, so on `main` today any request spanning three or more lines silently returns a short sequence — measured, a 260-base request comes back with 257. Deriving the span from `line_bytes` makes the terminator width irrelevant. `a_crlf_fasta_reads_the_same_bases_as_an_lf_one` pins it and fails against the old estimate.

## Not included

The optional preload/prime step #976 mentions. It is useful only once the batch pipeline is I/O-bound, and this measurement does not show that it is — the win here is decode CPU, not I/O.

## Verification

- `cargo nextest run --features dev` — **9197/9197 pass**, 146 skipped
- A/B measured with the same release binary, flipped by env var, on a real prepared reference

## Summary by CodeRabbit

* **New Features**
  * Added an optional cache for decoded FASTA sequence blocks to improve repeated and range-based reads.
  * Cache capacity can be configured through `FERRO_SEQUENCE_CACHE_BYTES`, including disabling it when needed.

* **Bug Fixes**
  * FASTA records without trailing newlines are now read correctly.
  * Reads now report clear errors when sequence data is shorter than expected.
  * Boundary reads, uppercase decoding, cache eviction, and disabled-cache behavior are handled reliably.

### DIFF
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -164,6 +164,14 @@ pub struct MultiFastaProvider {
     /// Per-`NG_`-version hosted-transcript map (#792); `None` when the manifest
     /// names no artifact (resolution falls back to the global reference map).
     ng_hosted: Option<crate::reference::ng_hosted_transcripts::NgHostedTranscripts>,
+    /// Byte-budgeted LRU of decoded sequence blocks (#976). `None` when the
+    /// budget is zero, in which case every read goes to the file as before.
+    ///
+    /// Complements `open_files` rather than duplicating it: that cache avoids the
+    /// `File::open` syscall, this one avoids the read *and* the newline-strip /
+    /// uppercase pass over the bytes. Workloads with many variants in one gene
+    /// re-read overlapping windows constantly, and paid for both every time.
+    region_cache: Option<SequenceRegionCache>,
 }
 
 /// Resolution inputs that uniquely identify a resolved transcript: the requested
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
## Self-review found a real regression in this PR — fixed (force-pushed)

Reviewing my own diff, I went looking for whether block-aligned reads could reach a file offset that a narrow read never would. They can, and it broke something.

`decode_range` sizes its buffer as `bases + one newline per line`, which assumes a newline **after** the last line. A FASTA whose final record ends without one has fewer bytes than that, and `read_exact_at` then fails with `failed to fill whole buffer` even though every requested base is present.

Blocks widened that: a block read makes **any** request in the last block reach the record's end. Measured on a 20 000-base record with no trailing newline:

| read | cache off | cache on (this PR, before the fix) |
| --- | --- | --- |
| bases 19 000..19 010 | `Ok("ACGTACGTAC")` | **`Err(failed to fill whole buffer)`** |
| the final five bases | `Err(…)` | `Err(…)` |

So the narrow read is a **regression this PR introduced**, and the tail read is an **older bug** it merely made easier to hit. I would rather say that plainly than have it found in review.

### The fix

Clamp the read to what the file actually holds (`metadata().len() - file_offset`). The decode loop already stops at `target_len`, so a short read simply yields fewer bases, and the callers that care already check the length (`fetch_window`, `cached_range`'s own guard). That repairs both rows of the table, not just the one I caused.

### The test

`a_record_without_a_trailing_newline_is_still_readable` builds exactly that fixture — every line newline-terminated except the last — and checks the narrow read, the tail read and the whole record, **with the cache both on and off**. Both configurations matter: a fix that only helped the cached path would have left the original bug in place, and the test would not have noticed.

Squashed into the single commit rather than added on top, per the repo's convention.

### Verification

- `cargo nextest run --features dev` — **9198/9198 pass**, 146 skipped

The performance numbers in the PR body predate this fix; it adds one `metadata()` call per read that misses the cache, which is noise beside the `pread` it accompanies, but I am happy to re-measure if that is worth confirming.

nh13:
@coderabbitai review

PR #1372, opened 2026-08-04
perf: stream the library and Python batch APIs in bounded chunks

### LINKED ISSUE #975: perf: stream the library batch/parallel API in bounded chunks (CLI path already chunked in #687)
### Summary

The **CLI** batch path already streams in bounded chunks (see "Already done" below). The remaining materialize-all-then-map-all-then-collect-all shape lives in the **library** batch/parallel API — `src/batch/processor.rs` and `src/parallel.rs` — which is what the Python bindings and benchmark harness drive. This issue is re-scoped to that library surface.

### Already done (CLI)

`ferro.rs`'s `run_batch` / `flush_chunk` (added in #687) already read input lazily via `reader.lines()`, map each `BATCH_CHUNK_LINES` (8192) chunk across the rayon pool, write results in input order, then advance — exactly the bounded-window shape proposed here. Both the `normalize` and `parse` batch subcommands use it. No change needed on the CLI path.

### Where the materialize-all shape remains

- `src/batch/processor.rs` — `BatchProcessor::map_collect` does `variants.par_iter().map(f).collect()` into a full `Vec<ItemResult<_>>`. This is **live**: it backs the Python `BatchProcessor.parse` / `parse_and_normalize` / `parse_with_progress`.
- `src/parallel.rs` — the public `parse_hgvs_parallel` / `normalize_parallel` / `parse_and_normalize_parallel` helpers take `&[S]` and return a `Vec`, i.e. `par_iter().collect()` over the whole input.

### The real constraint: the API boundary, not the internals

The Python entry points take a fully-materialized `variants: Vec<String>` and return a fully-materialized `BatchResult` list. So chunking `map_collect` *internally* would bound the intermediate allocation but would **not** reduce peak RSS for the Python consumer — the input `Vec` and output list both exist at the boundary by contract. A meaningful RSS win requires a **streaming API** (e.g. an iterator/generator-based entry that consumes and yields in bounded windows), which is a public-surface design decision, not a drop-in internal change. `src/parallel.rs`'s slice-in/`Vec`-out signatures have the same property.

### Proposal

1. Decide the streaming shape for the library API (an internal-only chunk fold buys little; the win needs a streaming entry point). Options: an iterator-consuming/`Iterator`-yielding variant of the `BatchProcessor` methods, and a streaming Python binding (generator / callback) alongside the existing `Vec`-based one.
2. Keep the existing `Vec`-based signatures for back-compat; add streaming variants rather than breaking callers.
3. Reuse the CLI's proven chunk-and-preserve-order approach (`BATCH_CHUNK_LINES`) as the internal engine.

### Acceptance

- A streaming library/Python entry point exists whose peak RSS is bounded and roughly flat in input size (measured before/after on the real prepared reference — the meaningful workload needs it).
- Output is byte-identical and in the same order as the `Vec`-based path.
- Throughput within noise of the current implementation.
- The existing `Vec`-based API and the already-chunked CLI path are unchanged.

nh13:
Closes #975

The CLI has streamed in bounded chunks since #687. The library surface that the Python bindings and benchmark harness drive did not: `BatchProcessor::map_collect` is `par_iter().map(f).collect()` into a full `Vec`, and `parallel.rs`'s helpers take `&[S]` and return `Vec`. As the issue puts it, chunking those internals buys nothing — both ends are resident **by contract**, so the win requires a streaming entry point.

Two commits: the Rust surface, then the Python one the issue identifies as the actual problem.

## Rust (`0c708b4`)

Four entry points, each consuming an `IntoIterator` and yielding an `Iterator`; every existing `Vec`-based signature untouched:

```
BatchProcessor::parse_streaming            parallel::parse_hgvs_streaming
BatchProcessor::parse_and_normalize_streaming   parallel::parse_and_normalize_streaming
```

Peak RSS, genomic substitutions:

| inputs | `Vec` API | streaming |
| --- | --- | --- |
| 250 000 | 157.5 MiB | 81.3 MiB |
| 1 000 000 | 609.7 MiB | 90.6 MiB |
| 4 000 000 | 2418.7 MiB | 82.9 MiB |
| 8 000 000 | — | 84.1 MiB |

Flat across a 32× range against linear growth.

## Python (`29c7b2d`)

`BatchProcessor.parse_streaming` / `.parse_and_normalize_streaming` accept any iterable of `str` and return a `BatchStream` yielding one `BatchItem` per input, in input order.

Measured on 1M inputs, **materializing one Python object per item on both sides** so the comparison is fair (`parse(...).successes()` against the stream):

| | wall clock | peak RSS |
| --- | --- | --- |
| list | 1.07 s | 1782 MiB |
| streaming | 1.90 s | 38 MiB |

**47× less memory at 1.8× the wall clock.** The time is the cost of one result object per item — that is what per-item reporting buys. An earlier, more flattering comparison against `success_count()` was *not* fair, because that call never builds the variants at all; the number above replaces it.

RSS is flat: 38.5 / 36.8 / 37.6 MiB at 100k / 400k / 1M, against 89 / 293 / 696 MiB for the list API.

## The chunk size is measured, not inherited — twice

#975 asks for bounded RSS **and** throughput within noise, and one chunk size does not deliver both on both paths, because per-item costs differ by orders of magnitude.

- **Rust** (`STREAM_CHUNK_ITEMS` = 65 536). At the CLI's 8192 the streaming path was **8% slower** (0.179 s against 0.166 s on 1M, spread under 1%): for a bare parse, per-chunk overhead is a real fraction of per-item work. At 65 536 it is at parity (0.163 s against 0.164 s), RSS still flat at ~83 MiB instead of ~17 MiB.
- **Python** (`BATCH_CHUNK_ITEMS` = 8192). Here each item becomes a `BatchItem` plus a variant `PyObject`, so the chunk *floor* sets peak memory. At 65 536 that floor was ~120 MiB, which made streaming **worse than the list API below ~150k inputs**. At 8192 it is flat at ~37 MiB and better at every size measured, and the object-creation cost dwarfs the per-chunk overhead that forced the Rust path larger — so chunking small costs nothing here.

Both constants carry the trade-off table in their doc comments. `BATCH_CHUNK_ITEMS` is now exported and the CLI's `BATCH_CHUNK_LINES` refers to it, so the shared engine has one definition rather than two that can drift; `map_collect` also picks up the extracted `build_pool`.

## Edges, each pinned

- A non-`str` element raises `TypeError` rather than being silently skipped — a dropped item would desynchronize the caller's own bookkeeping against the stream.
- A non-iterable argument raises **at the call**, not on the first `next()`.
- An exhausted stream stays exhausted instead of re-polling a spent iterator.
- `iter(stream) is stream`.
- `BatchItem.input` is set only on failure; on success the caller has the variant, and holding the input too would defeat the purpose.

## Tests

**Rust (5):** streaming output compared item-for-item against the `Vec` API across serial / global-pool / dedicated-pool, on a corpus spanning several axes plus inputs that must fail, sized past a chunk boundary so refills are exercised. Short, empty and exactly-one-chunk streams. Plus the property no equality test can see — that the input is consumed **lazily**, asserted by counting pulls: nothing at construction, exactly one chunk for the first result, nothing more until the chunk is crossed. *A streaming API that collected its input first would pass every equality test and bound nothing.*

**Python (13):** the same equivalence across worker counts for both operations, the laziness property, and each edge case above.

## Verification

- `cargo nextest run --features dev` — **9195/9195 pass**, 146 skipped
- `pytest tests/python/test_batch_streaming.py tests/python/test_batch_parallel.py` — **17 passed**
- `cargo fmt --all --check`, `cargo clippy --all-features --all-targets -- -D warnings`, `ruff check`, `ruff format --check`, `mypy` — all clean
- Type stubs updated for `BatchStream` / `BatchItem` and both new methods
- The wheel was built with `maturin develop --features python` to run the Python tests, since `cargo build --features python` cannot link on macOS locally (#1364)

### DIFF
--- a/python/ferro_hgvs/__init__.py
+++ b/python/ferro_hgvs/__init__.py
@@ -144,6 +144,8 @@ __all__ = [
     "BatchProgress",
     "BatchResult",
     "BatchProcessor",
+    "BatchStream",
+    "BatchItem",
     # Error handling classes
     "ErrorMode",
     "ErrorType",
--- a/python/ferro_hgvs/__init__.pyi
+++ b/python/ferro_hgvs/__init__.pyi
@@ -1,6 +1,6 @@
 """Type stubs for ferro-hgvs Python bindings."""
 
-from collections.abc import Iterable
+from collections.abc import Iterable, Iterator
 from typing import Any, Callable, Literal, overload
 
 from typing_extensions import Self
@@ -1236,11 +1236,83 @@ class BatchProcessor:
         """
         ...
 
+    def parse_streaming(self, variants: Iterable[str], workers: int = 0) -> BatchStream:
+        """Parse an iterable of HGVS strings lazily, yielding results in input order.
+
+        The streaming counterpart to `parse`. `parse` takes a fully materialized
+        list and returns a fully materialized BatchResult, so both are resident at
+        once; this consumes the iterable a chunk at a time and yields, so peak
+        memory does not grow with the input's length. Measured on 1M inputs:
+        ~1782 MiB materializing through `parse` against ~38 MiB streaming, at
+        about 1.8x the wall clock (the cost of one result object per item).
+
+        Accepts anything iterable of `str` — a list, a generator, or an open file.
+        The GIL is released for the parsing itself, as with `parse`.
+
+        Args:
+            variants: Iterable of HGVS strings.
+            workers: Worker threads. 0 (default) uses all cores; 1 is serial; N uses N threads.
+        """
+        ...
+
+    def parse_and_normalize_streaming(
+        self, variants: Iterable[str], workers: int = 0
+    ) -> BatchStream:
+        """Parse and normalize an iterable of HGVS strings lazily, in input order.
+
+        See `parse_streaming` for the rationale and the measured numbers.
+
+        Args:
+            variants: Iterable of HGVS strings.
+            workers: Worker threads. 0 (default) uses all cores; 1 is serial; N uses N threads.
+        """
+        ...
+
     def parse_with_progress(
         self, variants: list[str], callback: Callable[[BatchProgress], None]
     ) -> BatchResult:
         """Parse multiple HGVS strings with progress callback."""
         ...
+
+class BatchItem:
+    """One result from a BatchStream.
+
+    A streaming API cannot hand back a BatchResult, whose purpose is to hold
+    every result at once. This is the per-item equivalent: exactly one of
+    `variant` / `error` is set.
+    """
+
+    @property
+    def variant(self) -> HgvsVariant | None:
+        """The parsed (and, for the normalize stream, normalized) variant, or None."""
+        ...
+
+    @property
+    def input(self) -> str | None:
+        """The input string, present only on failure."""
+        ...
+
+    @property
+    def error(self) -> str | None:
+        """The rendered error, or None on success."""
+        ...
+
+    @property
+    def ok(self) -> bool:
+        """Whether this item parsed (and normalized) successfully."""
+        ...
+
+    def __repr__(self) -> str: ...
+
+class BatchStream(Iterator[BatchItem]):
+    """Lazy, order-preserving stream of BatchItem, one per input.
+
+    Its own iterator: `iter(stream) is stream`. Once exhausted it stays
+    exhausted.
+    """
+
+    def __iter__(self) -> BatchStream: ...
+    def __next__(self) -> BatchItem: ...
 
 # ============================================================================
 # Error Handling Classes
--- a/src/batch/mod.rs
+++ b/src/batch/mod.rs
@@ -50,4 +50,9 @@
 
 mod processor;
 
-pub use processor::{BatchConfig, BatchProcessor, BatchProgress, BatchResult, ItemResult};
+pub(crate) mod streaming;
+
+pub use processor::{
+    BatchConfig, BatchProcessor, BatchProgress, BatchResult, ItemResult, BATCH_CHUNK_ITEMS,
+    STREAM_CHUNK_ITEMS,
+};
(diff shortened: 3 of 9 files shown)


nh13:
@coderabbitai pause

nh13:
## Self-review follow-up: one streaming engine instead of two (force-pushed)

Reviewing my own diff before anyone else had to, the obvious criticism was that I had written the same chunk-and-drain iterator **twice** — `StreamingBatch` in `batch/processor.rs` and `StreamingMap` in `parallel.rs` — differing only in the type they yield (`ItemResult` against `Result`). This codebase's comments repeatedly make the point that a second copy is a second thing to drift, and they were right about this one.

Now there is one: `batch::streaming::StreamingMap`, generic over the mapped output, built by both surfaces.

It lives under `batch` rather than `parallel` for a reason worth recording: `parallel` is behind the `parallel` feature and `batch` is not, so the dependency can only run in that direction.

**Net −198/+37 lines**, and `dedicated_pool` replaces the `build_pool` helper the first version added — `map_collect` keeps its own inline construction because `num_threads == 1` means "serial" there and "no dedicated pool" in the streaming engine, and collapsing those two would have been the kind of false sharing this change is removing.

Squashed into the introducing commit rather than added on top, per the repo's convention, so the branch is still two commits.

### Verification after the change

- `cargo nextest run --features dev` — **9195/9195 pass**, 146 skipped
- all five Rust streaming tests still pass, including the laziness one
- `cargo check --features python` clean; `cargo fmt --all --check`, `cargo clippy --features dev --all-targets -- -D warnings` clean
- `git diff` between the pre- and post-simplification trees confirms the reword changed no code

The measurements in the PR body were taken before this refactor; it moves no logic, but if the numbers matter for review I am happy to re-run them.

PR #1373, opened 2026-08-04
fix(normalize): guard a net deletion's split by rule, not by a length bound

### LINKED ISSUE #1271: Normalizer: separations_are_meaningful does not cover net deletions, so a length threshold stands in for the coincidence guard

nh13:
Closes #1271

`separations_are_meaningful` decides whether the unchanged runs between two derived pieces are real separation or an artefact of `best_alignment`'s max-match hunt. It applied only to net **insertions**, leaving the three regimes unevenly protected:

| regime | guard |
| --- | --- |
| equal length | none needed — `best_alignment` compares position-wise, so there is no gap to place |
| net insertion | `separations_are_meaningful` |
| **net deletion** | **`MAX_UNGUARDED_SPLIT_BLOCK` only** — a 32 nt bound that guarded *by accident*, being smaller than the blocks at risk |

This extends the rule to every length-changing regime and retires that constant, so the coincidence question is answered by a rule about the derived pieces rather than by a length threshold that happens to sit below the risky sizes — which is what #1271 asks for.

## The case the accident was protecting

`delins.md:44-47`'s own worked example, `LRG_199t1:c.850_901delinsTTCCTCGATGCCTG` (52 nt → 14 nt). Measured: left to itself the aligner returns **four** pieces —

```text
[0, 39)  [40, 41)  [42, 49)  [50, 52)
```

— precisely the harm that passage names, and what the corpus shows as one correct `p.(Arg53_Arg54delinsSerCysAlaHisTyrLeuAla)` becoming three bogus consequences.

Under the extended rule that block has `net_change == 38`, so the required separation is `RAISED_PIECE_SEPARATION`, and its widest inter-piece gap is one base. The split is refused **on its merits**.

## The issue's objection no longer holds — measured, not assumed

#1271 records that a naive extension "breaks #1232 and #1157, which split a shorter payload at a genuinely unchanged base, and breaks confluence outright", and asks for its own measurement pass. That pass is this PR, and the objection has gone stale:

- **All 13 tests of those two issues pass**, including `issue_1232_spanning_delins_splits_at_unchanged_base`, `single_delins_is_decomposed_at_its_unchanged_bases`, and `sequence_identical_delins_and_allele_normalize_equal` — the confluence one.
- **All 168 manifest-backed conformance tests pass**, FAIL-set ledgers unmoved (`axis_normalized`, `axis_genomic`, `axis_errors`, both idempotency axes). These skip silently without a prepared reference, so a plain local run does not cover them.
- The cis sweeps pass, including `cis_junction_crossing_shift` and `repeat_span_sibling_overlap`.
- The spec enumeration **improves**.

Whatever made the naive attempt fail was fixed elsewhere in the meantime — plausibly #1237's regime-aware bound and the sequence-first splitter that followed it. Worth stating plainly rather than quietly deleting the note, because the next person to read that objection should know it was retested.

## The enumeration move is the LRG_199 row itself

```
ProjectionSplitsSingleMember  10 -> 9
ProjectionPinned            1167 -> 1168
```

The 2184-row total is unchanged: one row moved from a divergence to a pass. `Status::ProjectionSplitsSingleMember` was added by #1272 *specifically* to make this regression visible — its own comment says "which is also what makes the LRG_199 delins regression visible" — so the right row moving is the strongest confirmation available. Both counts are re-blessed with that reasoning at the constants.

## The test is rewritten, not deleted

`partition_block_bounds_a_long_net_deletion_but_not_its_siblings` pinned the *old* mechanism. Both mechanisms return the identical single piece — `whole()` is the refusal path for either — so a test that only checked "one piece" would keep passing for the wrong reason. Renamed to `partition_block_refuses_a_coincidental_net_deletion_split`, it now asserts:

1. the aligner **does** reach the block (no bound stops it any more),
2. its unrefused output is a multi-piece split, and
3. `separations_are_meaningful` is what rejects it.

## Doc comments corrected rather than dropped

Three described the retired mechanism, including two claiming the LRG_199 block "never reaches `best_alignment` at all". It does now, and is refused afterwards. `separations_are_meaningful` carries the four-piece evidence and the re-measurement.

## Verification

- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 cargo nextest run --features dev` — **9190/9190 pass**, 146 skipped
- 168/168 manifest-backed conformance tests against a real prepared reference

Net −85/+103 lines, most of it comment.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -1558,48 +1558,11 @@ const CANONICAL_PAD: i64 = 128;
 /// coincidence. Structure alone therefore cannot tell a real separation from an
 /// accidental one; [`separations_are_meaningful`] is what draws that line.
 ///
-/// This bound therefore applies only where a real guard exists — see
-/// [`MAX_UNGUARDED_SPLIT_BLOCK`] for the regime where one does not.
+/// Every length-changing regime now has that guard, so this is the single bound
+/// for all of them (#1271). It used to be joined by a second, smaller bound for
+/// net deletions, which `separations_are_meaningful` did not reach; extending
+/// that rule retired it. See its doc comment for the measurement.
 const MAX_SPLIT_BLOCK: usize = 1024;
-
-/// The same bound for a **net deletion**, where nothing else guards the split.
-///
-/// [`separations_are_meaningful`] is restricted to net insertions, so the three
-/// block regimes are not equally protected:
-///
-/// | regime | guard against a coincidental split |
-/// | --- | --- |
-/// | equal length | none needed — [`best_alignment`] compares position-wise, so there is no gap to place and no search to go wrong |
-/// | net insertion | [`separations_are_meaningful`] |
-/// | **net deletion** | **nothing** |
-///
-/// In the third regime this constant *is* the guard, which is why it stays at
-/// the original 32. That was never the intent — it guarded by accident, being
-/// smaller than the blocks at risk — but raising it to [`MAX_SPLIT_BLOCK`]
-/// across the board removed the cover and split the spec's own
-/// `delins.md:44-47` worked example:
-///
-/// ```text
-/// LRG_199t1:c.850_901delinsTTCCTCGATGCCTG        52 nt -> 14 nt
-///   spanning  LRG_199:g.646630_646681delins…
-///   split     LRG_199:g.[646630_646636delinsTTCCTCG;646640_646678delinsC;
-///                        646680_646681delinsTG]
-/// ```
-///
-/// Which is precisely the harm that passage names, on that passage's own
-/// example: the corpus shows one correct
-/// `p.(Arg53_Arg54delinsSerCysAlaHisTyrLeuAla)` becoming three bogus
-/// consequences.
-///
-/// Splitting the bound by regime keeps the raise where it is safe — a long
-/// equal-length block has no alignment choice to get wrong, and a long net
-/// insertion is guarded — while leaving net deletions exactly as they were.
-///
-/// The principled fix is to extend [`separations_are_meaningful`] to net
-/// deletions and then retire this constant. That is not a local change: the
-/// notes there record that a naive extension breaks #1232 and #1157 and breaks
-/// confluence outright, so it needs its own measurement pass.
-const MAX_UNGUARDED_SPLIT_BLOCK: usize = 32;
 
 /// Unchanged reference bases two pieces of a net insertion must be separated by
 /// before the split between them is believed. See `separations_are_meaningful`.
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
## One behaviour change worth naming explicitly

A reviewer's likely first probe, so better stated than found: retiring `MAX_UNGUARDED_SPLIT_BLOCK` does not only affect **long** net deletions. It changes **short** ones too, in the opposite direction.

Before, net deletions had two regimes:

| block size | before | after |
| --- | --- | --- |
| > 32 nt | refused outright by the bound — never reached the aligner | reaches the aligner, then judged by `separations_are_meaningful` |
| ≤ 32 nt | reached the aligner and was **not** separation-checked at all | now separation-checked like every other length-changing block |

So a short net deletion whose pieces are separated only by coincidence used to be allowed to split and now is not. That is the intended direction — the rule is meant to be about the pieces, not about the length — but it is a change to blocks the old bound never guarded, and it deserves to be said rather than implied by "extends the rule to every length-changing regime".

Measured, and nothing moved for it: the full suite is 9190/9190 with both oracles, all 13 tests of #1232/#1157 pass (including `single_delins_is_decomposed_at_its_unchanged_bases`, which is exactly a short-block split at an unchanged base), and all 168 manifest-backed conformance tests pass with the FAIL-set ledgers unmoved. The only census movement is the LRG_199 row, which is a long block and an improvement.

If a short-block split is ever wanted back, the knob is `MIN_PIECE_SEPARATION` / `RAISED_PIECE_SEPARATION` and the `MAX_SINGLE_BASE_SEPARATION_CHANGE` threshold that chooses between them — not a length cap on the block.

PR #1374, opened 2026-08-04
feat(spdi): reference-aware SPDI and apply-to-reference for an encoding-invariant key

### LINKED ISSUE #1159: Feature: reference-aware SPDI / apply-to-reference for an encoding-invariant variant key
@nh13 This percolated up while I was poking at a workaround for #1157 and #1158. Seemed like the only viable approach to get a comparable key between two equivalent variants with different normalized representations (i.e. just while #1157 and #1158 remain open) was going to be projecting the variants back into their resulting sequence. I didn't want to introduce that into the project codebase, and it seemed like a feature that might make sense to have in ferro (convert variant to sequence) for some contexts/use cases. 

Not a blocker for me if #1157 and #1158 get merged - up to you if you think there are contexts that warrant implementing it!

---

> [!NOTE]

## What's missing

There is no **reference-aware** way to obtain a canonical, encoding-invariant
representation of a variant from the Python API. `hgvs_to_spdi` exists but is a
"simple conversion without reference lookup", so it only succeeds for edits that
carry their sequence inline (SNV, plain insertion) and **fails for every edit that
needs the reference to know the affected bases** — `del`, `delins`, `inv`, `dup`.

Request: expose a reference-aware conversion to SPDI (or an equivalent canonical
key), using the reference the calling object already holds, that resolves those
edit types.

## Motivation

Callers need a stable key that is equal **iff two descriptions denote the same
edit**, independent of how the edit was written. SPDI (`sequence:position:deletion:
insertion`) is the natural standard canonical form for this, and ferro already
models it (`SpdiVariant`, `hgvs_to_spdi`, `parse_spdi`, `spdi_to_hgvs_variant`).

Concretely, a downstream pipeline matches observed variants (decomposed per-base
from read alignments) against an expected library (written as `delins`/`inv`) by
comparing normalized HGVS.g strings. Because those two encodings normalize to
different strings (#1157), the string key fails and equivalent variants don't
match. A reference-aware SPDI would give an encoding-invariant key and make that
matching robust. `EquivalenceChecker` — the other natural fallback — is also
affected (#1158), so today there is no working API to decide this.

## Current state (verified on v0.10.1, commit 43ec9ab)

`hgvs_to_spdi(variant)` takes only an `HgvsVariant` and does no reference lookup:

```python
import ferro_hgvs
p = ferro_hgvs.parse

ferro_hgvs.hgvs_to_spdi(p("TEMPLATE:g.11C>T"))       # SpdiVariant  TEMPLATE:10:C:T
ferro_hgvs.hgvs_to_spdi(p("TEMPLATE:g.20_21insAC"))  # SpdiVariant  TEMPLATE:20::AC

# edits that need the reference all raise ProjectionError:
ferro_hgvs.hgvs_to_spdi(p("TEMPLATE:g.8_14delinsGATTA"))
#   ProjectionError: missing reference data: Cannot convert delins to SPDI:
#   deleted sequence of length 7 is unknown (no reference data)
ferro_hgvs.hgvs_to_spdi(p("TEMPLATE:g.12_20del"))
#   ProjectionError: missing reference data: deleted sequence not provided; ...
ferro_hgvs.hgvs_to_spdi(p("TEMPLATE:g.1724_1725inv"))
#   ProjectionError: missing reference data: inversion sequence not provided; ...
ferro_hgvs.hgvs_to_spdi(p("TEMPLATE:g.47dup"))
#   ProjectionError: missing reference data: duplication sequence not provided; ...
```

`Normalizer`, `VariantProjector`, and `CoordinateMapper` all hold reference data
but expose no `to_spdi` (or apply-to-reference) method, so there is no way to
complete these conversions from Python.

## Proposed API

A reference-aware SPDI conversion on a reference-bearing object, e.g.:

```python
n = ferro_hgvs.Normalizer(reference_json=ref)      # or VariantProjector(...)
n.to_spdi(ferro_hgvs.parse("TEMPLATE:g.8_14delinsGATTA"))
#   -> SpdiVariant  (deleted/inserted bases resolved from the reference)
```

so that two encodings of the same edit yield the **same** `SpdiVariant`:

```python
n.to_spdi(p("TEMPLATE:g.8_14delinsGATTA")) == n.to_spdi(p("TEMPLATE:g.[8A>G;9G>A;11C>T;13_14del]"))  # True
```

(Signature/placement is illustrative — wiring `hgvs_to_spdi` to accept an optional
reference/provider would be equally good.)

## Design considerations

1. **Confluence is the whole point.** The feature is only useful if equivalent
   encodings produce identical SPDI. This is a hard requirement, and it must not
   inherit the normalization non-confluence in #1157 — SPDI should be derived from
   the resulting reference bases (maximally-shifted, blunt-ended per the SPDI
   spec), not from the input's partitioning. Idempotence and encoding-invariance
   should be covered by tests.

2. **Multi-part cis alleles.** Many real inputs are `g.[a;b;c]` alleles, which are
   not a single SPDI. The API should define a canonical result for these — e.g. a
   **list/set of SPDIs** (each component, order-independent), or see (3).

3. **An apply-to-reference primitive would also solve this, more generally.** A
   method returning the **resulting sequence** for any variant
   (`n.resulting_sequence(variant) -> str`, or a windowed/trimmed
   `(start, ref_allele, alt_allele)`) handles every variant type uniformly,
   including multi-part alleles, and is the ground truth for equivalence. SPDI is
   the more compact/standard key; apply-to-reference is the more general one.
   Either (ideally both) would serve the use case.

## Relationship to #1157 / #1158

Complementary, not a duplicate:

- Fixing **#1157** (confluent normalization) would let callers key on normalized
  HGVS strings again; fixing **#1158** would make `EquivalenceChecker` usable.
- A reference-aware SPDI / apply-to-reference primitive is still valuable
  independently: a stable, display-independent key and a cross-tool standard — and
  it is very likely the implementation substrate a correct fix for #1157/#1158
  would use (normalize/compare via the resulting reference bases).

nh13:
Closes #1159

#1159 asks for a key that is equal **iff two descriptions denote the same edit**, and reports that Python has no way to get one:

- `hgvs_to_spdi` does no reference lookup, so `del`, `delins`, `inv` and `dup` all raise `ProjectionError`;
- keying on normalized HGVS strings fails, because normalization is not encoding-invariant for complex indels (#1157);
- `EquivalenceChecker`, the other natural fallback, was affected too (#1158).

## Three methods, and what each actually promises

The distinction matters more than the code, so it is stated at every call site rather than left to inference:

| method | reference-aware | encoding-invariant | notes |
| --- | --- | --- | --- |
| `to_spdi` | ✅ | ❌ | a **transliteration** — preserves the caller's partitioning; a cis allele is not one triple and is refused |
| `canonical_spdi` | ✅ | ✅ | derived from the resulting bases, so partitioning cannot survive it |
| `apply_to_reference` | ✅ | ✅ (as bases) | the resulting bases themselves — what #1159 calls the ground truth |

A caller reaching for "SPDI" will assume `to_spdi` is a key. It is not, and saying so is half the point of the change.

All three land on `Normalizer` in Rust and on the Python `Normalizer`, which is where #1159's own example reaches for them.

## Demonstrated against the issue's exact complaint

```python
>>> for s in ["TEMPLATE:n.3_7delinsGGCTA", "TEMPLATE:n.3_5del", "TEMPLATE:n.3_5inv", "TEMPLATE:n.3dup"]:
...     ferro_hgvs.hgvs_to_spdi(parse(s))          # the reported behaviour
ProjectionError  ×4

>>> n.to_spdi(parse("TEMPLATE:n.3_7delinsGGCTA"))  TEMPLATE:2:ATTAC:GGCTA
>>> n.to_spdi(parse("TEMPLATE:n.3_5del"))          TEMPLATE:2:ATT:
>>> n.to_spdi(parse("TEMPLATE:n.3_5inv"))          TEMPLATE:2:ATT:AAT
>>> n.to_spdi(parse("TEMPLATE:n.3dup"))            TEMPLATE:3::A

>>> n.canonical_spdi(parse("TEMPLATE:n.3_7delinsGGCTA"))              TEMPLATE:2:ATTAC:GGCTA
>>> n.canonical_spdi(parse("TEMPLATE:n.[3A>G;4T>G;5T>C;6A>T;7C>A]"))  TEMPLATE:2:ATTAC:GGCTA   # equal
```

## Nothing here re-implements applying a variant

`EquivalenceChecker` already did it privately for #1158's `SequenceMatch` verdict. `apply_triples` and `triples_are_disjoint` **move** into a new `spdi::apply` module and the checker calls them there; `edit_triples` and `fetch_window` become free functions over a provider. One home rather than a second copy to drift — which this codebase's comments repeatedly warn about.

## What the key does not claim

Documented on the module, not implied:

- **Guaranteed** — two variants on one accession with the same resulting sequence give the same `SpdiVariant`, whatever their spelling, member count or member order.
- **Not claimed** — that the triple is "maximally shifted" in the SPDI specification's sense. It is blunt-trimmed, which is deterministic and *window-independent* (a wider window contributes only flanks that trim away), but it does not roll an indel through a repeat to a canonical extreme. Equality is the contract here, not byte-compatibility with another implementation's canonical form.

That is an honest limit rather than a hidden one: for the matching use case #1159 describes, equality is what is needed.

## Refused rather than guessed

Shapes with no single well-defined resulting sequence decline, each for a distinct reason: a non-cis allele (several molecules), a null or unknown allele, members on different accessions, members that **overlap** (applying them depends on order), an edit SPDI cannot represent, or a span past `MAX_APPLY_WINDOW` (100 000 bases).

## Tests

**7 Rust unit tests**, **17 Python tests**. Both suites pin:

- the headline invariance (spanning `delins` vs its five-member decomposition — all five bases change, so the decomposition really is five members);
- member-order independence;
- the exact expected triples, not merely "does not raise";
- that four genuinely **different** edits get four different keys — a key that collapsed everything would satisfy every other assertion;
- that `apply_to_reference`'s window is a real slice of the reference, so `resulting` is not a claim about bases nobody checked;
- each refusal shape, with the specific `ProjectionError` type rather than a blind `Exception`.

The Python suite also pins the failure #1159 reports, so the new methods are measured against a real gap rather than a supposed one.

## Verification

- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 cargo nextest run --features dev` — **9197/9197 pass**, 146 skipped
- `pytest tests/python/test_reference_aware_spdi.py` — **17 passed**
- `cargo fmt --all --check`, `cargo clippy --all-features --all-targets -- -D warnings`, `ruff check`, `ruff format --check`, `mypy` — all clean
- Type stubs updated for `AppliedVariant` and all three methods
- Wheel built with `maturin develop --features python` to run the Python tests (`cargo build --features python` cannot link locally, #1364)

The shape follows #1159's own proposal: `canonical_spdi` is the encoding-invariant key it asks for, and `apply_to_reference` is the more general primitive its option (3) suggests.

### DIFF
--- a/python/ferro_hgvs/__init__.py
+++ b/python/ferro_hgvs/__init__.py
@@ -126,6 +126,7 @@ __all__ = [
     "Normalizer",
     # SPDI classes
     "SpdiVariant",
+    "AppliedVariant",
     # Coordinate classes
     "ZeroBasedPos",
     "OneBasedPos",
--- a/python/ferro_hgvs/__init__.pyi
+++ b/python/ferro_hgvs/__init__.pyi
@@ -697,6 +697,52 @@ class Normalizer:
         """Normalize an HGVS variant object."""
         ...
 
+    def to_spdi(self, variant: HgvsVariant) -> SpdiVariant:
+        """Convert to SPDI, resolving edits that need the reference (#1159).
+
+        The module-level `hgvs_to_spdi` does no reference lookup and so fails on
+        `del`, `delins`, `inv` and `dup`; this uses the reference this Normalizer
+        holds and resolves them.
+
+        Preserves how the variant was written, so it is a transliteration and NOT
+        an encoding-invariant key — use `canonical_spdi` for that. A multi-member
+        allele is not one triple and is refused.
+        """
+        ...
+
+    def canonical_spdi(self, variant: HgvsVariant) -> SpdiVariant:
+        """An encoding-invariant SPDI key, derived from the resulting bases (#1159).
+
+        Two descriptions on one accession denoting the same edit give the same
+        key, whatever their spelling, member count or member order — so a
+        spanning `delins` and its decomposed allele match. Normalized HGVS
+        strings cannot serve for this, because normalization is not
+        encoding-invariant for complex indels (#1157).
+
+        Blunt-trimmed to the block that actually changed: deterministic and
+        independent of how wide a span the input named, but not claimed to be
+        "maximally shifted" in the SPDI specification's sense.
+
+        Raises:
+            ProjectionError: if there is no single well-defined resulting
+                sequence — a trans/mosaic/chimeric or null allele, members on
+                different accessions, members that overlap, an edit SPDI cannot
+                represent, or a span wider than 100 000 bases.
+        """
+        ...
+
+    def apply_to_reference(self, variant: HgvsVariant) -> AppliedVariant:
+        """Apply the variant and return the window before and after (#1159).
+
+        The ground truth for equivalence, and the more general of the two
+        primitives: two descriptions denote the same edit exactly when their
+        `resulting` bases match. `canonical_spdi` is this reduced to a key.
+
+        Raises:
+            ProjectionError: as `canonical_spdi`.
+        """
+        ...
+
     def normalize(self, hgvs_string: str) -> str:
         """Parse and normalize an HGVS string."""
         ...
@@ -1135,6 +1181,31 @@ class BatchProgress:
     def percent(self) -> float:
         """Get the percentage complete."""
         ...
+
+class AppliedVariant:
+    """A variant applied to its reference: the same window before and after (#1159)."""
+
+    @property
+    def accession(self) -> str:
+        """The accession every member acts on."""
+        ...
+
+    @property
+    def start(self) -> int:
+        """0-based start of the window both sequences cover."""
+        ...
+
+    @property
+    def reference(self) -> str:
+        """The reference bases over the window."""
+        ...
+
+    @property
+    def resulting(self) -> str:
+        """The same window after every member has been applied."""
+        ...
+
+    def __repr__(self) -> str: ...
 
 class BatchResult:
     """Result of batch processing."""
--- a/src/equivalence/checker.rs
+++ b/src/equivalence/checker.rs
@@ -344,8 +344,8 @@ impl<P: ReferenceProvider> EquivalenceChecker<P> {
         };
 
         match (
-            apply_triples(&reference, win_start, &triples1),
-            apply_triples(&reference, win_start, &triples2),
+            crate::spdi::apply::apply_triples(&reference, win_start, &triples1),
+            crate::spdi::apply::apply_triples(&reference, win_start, &triples2),
         ) {
             // Case-insensitive, like the deletion check in `apply_triples`:
             // reference FASTAs are often soft-masked (repeats lower-cased), so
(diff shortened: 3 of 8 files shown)


nh13:
@coderabbitai pause

nh13:
## Self-review follow-up: one test was asserting less than it looked like (force-pushed)

`shapes_without_one_resulting_sequence_decline` had three problems, all mine, found reviewing my own diff:

1. **The control was smuggled into the decline list** and told apart by comparing the descriptor string — so the test both asserted "these must fail" and "this must succeed" through an `if` on a fixture's text.
2. **A comment mislabelled it.** `// Members on different accessions` sat above `NC_KEY.1:g.3A>G`, which is a single-member variant on the *served* accession — the control. The different-accession case was never actually tested.
3. **`let Ok(variant) = … else { continue }` silently skipped anything that stopped parsing.** A fixture that broke would have left the test green having asserted nothing about it.

Point 3 was not hypothetical: after splitting the test out I wrote the mixed-accession fixture with the wrong syntax (`NC_KEY.1:g.[3A>G];NC_OTHER.1:g.7T>G` rather than `[NC_KEY.1:g.3A>G;NC_OTHER.1:g.7T>G]`), and the rewritten version **failed loudly** with the parse error instead of skipping — which is exactly the behaviour the old one lacked.

Now:

- `shapes_without_one_resulting_sequence_decline` covers three genuinely distinct reasons, each with the reason in its failure message, and each asserted against **both** `apply_to_reference` and `canonical_spdi`. Every fixture is `expect`ed to parse.
- The provider gained a second contig so "members on different accessions" is the real thing rather than an absent-reference failure wearing its label.
- `a_plain_single_member_variant_applies` is the control, standing on its own — without it, every decline assertion would also be satisfied by an implementation that declined *everything*.

Squashed into the single commit rather than added on top, per the repo's convention.

### Verification

- `cargo nextest run --features dev` — **9198/9198 pass**, 146 skipped (three more tests than before: the split control plus the added assertions)

Issue #466, opened 2026-05-28
tracking: HGVS spec-revision suggestions to file with SVD-WG


nh13:
Umbrella tracker for spec-*text* changes ferro would suggest to the HGVS nomenclature committee (SVD-WG) — cases where ferro's behavior is grounded in spec convention/examples but the spec does not state the rule explicitly, so independent parsers can diverge while all claiming compliance.

Each candidate should capture: spec location, the ambiguity, the suggested change, ferro's current behavior, and — once filed — a link to the upstream issue/consultation at [`HGVSnomenclature/hgvs-nomenclature`](https://github.com/HGVSnomenclature/hgvs-nomenclature). Close a candidate when its upstream suggestion is filed (or the committee declines); close this tracker when no candidates remain.

### Candidates

- [ ] **Lowercase, case-sensitive edit-type abbreviations** (folds in #448)
  - **Spec:** `general.md` → "[Abbreviations in variant descriptions](https://hgvs-nomenclature.org/stable/recommendations/general/#abbreviation)" lists `del`, `dup`, `ins`, `inv`, `fs`, `ext` (and `delins` on the delins pages) with lowercase examples only. Casing is conveyed by convention, never stated as a rule.
  - **Ambiguity:** uppercase/mixed-case forms (`c.123_125DEL`, `c.20_21Dup`, `c.100_102INV`) are left to implementer judgment — reject, silently normalize, or accept.
  - **Suggested change:** add an explicit note that these tokens are case-sensitive and must be lowercase; uppercase/mixed-case spellings are not valid HGVS (mirrors the spec's existing explicit amino-acid-casing guidance).
  - **ferro today:** lenient mode canonicalizes to lowercase + emits `W1004 MixedCaseEditType`; strict mode rejects. (`con` deprecated per SVD-WG009 and `spl` are intentionally excluded — neither is in the abbreviation registry.)
  - **Upstream issue:** _not yet filed_

- [ ] **Canonical expansion of intronic-offset / UTR-marker CDS ranges inside `ins[...]`** (folds in #395 item 3)
  - **Spec:** insertion descriptions allow a same-reference position range as the inserted payload, but the spec does not define how to expand a CDS-coordinate range whose endpoints are intronic offsets (`244-8_249`, `c.244+8_249`) or UTR markers (`*100_*200`) into a literal inserted sequence.
[...]

PR #1375, opened 2026-08-04
docs(normalize): repoint the circular-normalizer deferral at its live tracker

nh13:
Refs #466 — a follow-up from auditing that tracker. Comments only; no behaviour change.

Seven comments across `normalize/mod.rs` and three test modules cite **#951** as the tracker for implementing circular (`o.`) 3'-shifting, and two of them say "the **open** #951".

It is not open. #951 was closed `NOT_PLANNED` on 2026-07-08, superseded by **#466's circular candidate**, which asks SVD-WG to state the origin-wraparound shift semantics before anything is built against them — "the most 3' position possible" is ill-defined on a ring, so there is currently nothing to implement.

#466's candidate 4 already records that it supersedes #951. Only the code had not been told.

## Why this is more than a search-and-replace

This is the **second** time the same pointer has gone stale in the same place. #951's own body was filed precisely because the comment then cited the closed, MT-scoped **#129**:

> The deferral this points at, **#129, is CLOSED**, so the gap has no open tracker.

So the chain is `#129 → #951 → #466's candidate`. The comment at the pass-through arm now records that chain and asks the next person to repoint rather than leave a third dangling generation — which is the part likely to save someone the same investigation.

## Scope

- `src/normalize/mod.rs` — three sites, including the `HV::Circular` pass-through arm.
- `tests/it/issue_951_circular_normalize_passthrough.rs` — module docs and one assertion message. **Keeps its filename** as the historical anchor for the behaviour it pins; only the claim about what is open changes.
- `tests/it/issue_1282_position_zero.rs`, `tests/it/issue_1217_mito_terminus_insertion_clamp.rs` — one site each.

## Verification

- `cargo nextest run --features dev` — **9190/9190 pass**, 146 skipped

The audit that found this is on #466, along with re-verification of all four of its candidates and both markup defects against submodule `21.1.4`.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -725,7 +725,7 @@ impl SequenceEnds {
 ///
 /// `o.` stays out: `normalize_core` returns circular `o.` variants unchanged, so
 /// no shuffle runs and nothing can saturate. A genuine circular normalizer is
-/// #951.
+/// #466's circular candidate.
 ///
 /// The genomic caller passes a **window** rather than the whole contig, so it
 /// must say which of its ends are real — see [`SequenceEnds`].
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

Issue #1355, opened 2026-08-03
normalize: insertion_to_repeat discards a valid in-phase rotation when an out-of-phase one scores higher


nh13:
`insertion_to_repeat` scores rotations before applying the #882 phase gate, so an out-of-phase rotation with a longer tract discards a valid in-phase repeat — the same defect #1349 fixes in `insertion_to_duplication`.

## Where

`src/normalize/rules.rs`, `insertion_to_repeat`:

- Lines 370-379 iterate every cyclic rotation of the inserted unit and keep only the one with the largest `ref_count`.
- Line 381 unwraps that single winner (`let (unit, ref_start, ref_count) = best?;`).
- Lines 465-475 apply `tandem_edit_preserves_insertion` once, to the winner, and `return None` when it fails.

A rotation that loses on `ref_count` is never reconsidered, so a high-scoring out-of-phase rotation takes the whole result down with it. This is byte-for-byte the ordering #1349 corrected in `insertion_to_duplication`, and the two functions are documented as a deliberate symmetric pair (`insertion_to_duplication`'s "Why this helper exists" section).

## Repro

Confirmed against the reference #1349 uses, with a 2-copy insertion instead of a 1-copy one:

```rust
let r = b"GTACGTCCCCTCCTCCTCCTCCCGGGG";
// rotation TCC: in-phase, tract [5,8), one copy
// rotation CCT: out of phase, tract at index 8, four copies -> wins on ref_count
assert!(tandem_edit_preserves_insertion(r, 7, b"TCCTCC", 5, 8, b"TCC", 3)); // the discarded candidate is valid
assert!(insertion_to_repeat(r, 7, b"TCCTCC", false, None, ShuffleDirection::ThreePrime).is_some()); // FAILS: None
assert!(insertion_to_repeat(r, 7, b"TCCTCC", false, None, ShuffleDirection::FivePrime).is_some());  // FAILS: None
```

Observed:

```
find_tandem_extent(r, 7, b"TCC") = Some((5, 1))
find_tandem_extent(r, 7, b"CCT") = Some((8, 4))
tandem_edit_preserves_insertion(r, 7, b"TCCTCC", 5, 8, b"TCC", 3) = true
insertion_to_repeat(...) -> None   # both directions
```

The `TCC[3]` emission clears the #882 gate and is discarded unseen.

## Why this is not a copy of the #1349 fix

[...]

PR #1378, opened 2026-08-04
fix(spdi)!: give an unspelled identity the span it claims

### LINKED ISSUE #1362: hgvs_to_spdi maps an unspelled identity to a zero-width triple that aliases an insertion junction
An identity edit whose bases the input does not spell converts to a triple that claims no bases at all, and a zero-width triple at an interbase is indistinguishable from an insertion at that same interbase.

`NaEdit::Identity` is the only arm of `hgvs_to_spdi` that does not consult the provider for a sequence the input omitted. It defaults to the empty string:

| input | `hgvs_to_spdi` | expected |
|---|---|---|
| `g.10A=` (base spelled) | `9:A:A` | correct already |
| `g.10=` | `9::` | `9:C:C` |
| `g.10_12=` | `9::` (three-base span lost) | `9:CGT:CGT` |
| `g.=` (whole-entity) | `0::` | fine — names no span |

The Deletion, Delins and Inversion arms all fetch their omitted sequence from the provider via `fetch_reference_bases`, or return `MissingReferenceData` when there is none. Identity does neither.

## Why the zero width matters

It is not just a loss of information. Consider `g.[261_262dup;263=]`, which is a real normalizer output — the 5'-shuffle form of `g.[261_262insGA;262_263insA;263del]`. Converting its members gives:

```
TEMPLATE:g.261_262dup  ->  262::AG     (insertion at the junction after 262)
TEMPLATE:g.263=        ->  262::       (zero-width, same interbase)
```

Both members land on interbase 262. An applier walking the members cannot distinguish that from two insertions competing for one interbase — which has no single well-defined resulting sequence — so it must decline the entire description. The identity claims base 263, so its span is `262..263` and the two members are in fact disjoint; only the lossy conversion makes them collide.

The practical consequence is that this project's own SPDI-based applier, the one oracle that judges denotation rather than form, cannot evaluate any description containing an unspelled identity member. That is an oracle blind spot, not merely a cosmetic defect.

## Fix

Have the unspelled identity fetch its bases the way its sibling arms do, and report `MissingReferenceData` on the provider-less path. A whole-entity `g.=` must keep its zero-width triple: it asserts the entire reference is unchanged and names no span, so there is nothing to fetch (and narrowing it to base 1, or fetching a whole chromosome, would both be wrong).

Note this makes `hgvs_to_spdi_simple` — and the Python `hgvs_to_spdi` that wraps it — return an error for `g.10=` where it previously returned a triple whose span was wrong, so it is a breaking change to that entry point.

## Provenance

Found while verifying an unrelated claim about 5'-shuffle confluence. The claim asserted that `hgvs_to_spdi` *declines* the identity member; it does not — it accepts it and returns `262::`, and the decline happens later in the applier's own disjointness guard. The distinction is what locates the defect here rather than in the applier.

nh13:
Closes #1362

`NaEdit::Identity` was the only arm of `hgvs_to_spdi` that did not consult the provider for a sequence the input omitted. It defaulted to the empty string:

| input | before | after |
| --- | --- | --- |
| `g.10A=` (spelled) | `9:A:A` | unchanged |
| `g.10=` | `9::` | `9:G:G` |
| `g.10_12=` | `9::` — span lost | `9:GCA:GCA` |
| `g.=` (whole-entity) | `0::` | unchanged, deliberately |

`Deletion`, `Delins` and `Inversion` all fetch via `fetch_reference_bases` or report `MissingReferenceData`. This arm now does too.

## The zero width was not merely lossy

A zero-width triple at an interbase is indistinguishable from an **insertion** at that interbase, and that aliasing blinded this project's own SPDI applier — the one oracle that judges denotation rather than form. On the real 5'-shuffle output `g.[261_262dup;263=]`:

```
261_262dup  ->  262::AG     insertion after 262
263=        ->  262::       zero-width at the same interbase
```

Two members apparently competing for one interbase, which has no single well-defined resulting sequence, so the applier declined the whole description. The identity in fact claims base 263 and the two are disjoint; only the conversion collided them.

## Whole-entity `g.=` stays zero-width

Deliberately, and it is the one arm that must not consult the provider: `g.=` asserts the entire reference is unchanged and names no span. Narrowing it to base 1 would invent a claim, and fetching a chromosome to prove it unchanged would be absurd. Pinned by its own test.

## Breaking change

`hgvs_to_spdi_simple` — and the Python `hgvs_to_spdi` that wraps it — now report `MissingReferenceData` for `g.10=` where they used to return a triple whose span was wrong. Hence the `!`. The message names the construct and both ways out (spell the base, or convert with a provider).

## This closes the oracle half of #1351

Two tests in `cis_spelling_confluence_gap` existed to pin the blindness, and are **inverted rather than deleted**:

- `an_identity_member_leaves_the_output_beyond_the_applier` → `an_identity_member_is_now_expressible_to_the_applier`, which asserts what the blindness made impossible: that the output **denotes the input's bases**. That check is the actual gain here — a real 5'-shuffle output whose denotation nothing could previously verify.
- the sweep's `beyond_the_applier` count goes **1 → 0**, with the reason recorded at the assertion so a future regression is legible.

**#1351's other half stands** and I have not touched it: the sweeps should still distinguish "skipped because the *input* was unconvertible" from "skipped because the *output* was inexpressible". Only the second was an oracle gap, and only that one is now closed.

## Interaction with #1374

#1374 (reference-aware SPDI / apply-to-reference, `#1159`) builds its `variant_edit_triples` on `hgvs_to_spdi`, so it inherits this fix: `apply_to_reference` and `canonical_spdi` can now handle a description containing an unspelled identity member, which they would previously have declined. The two are independent changes and merge in either order.

## A note on triage

This issue was open when this campaign's triage ran and I missed it — my query returned 19 issues ending at #1355 and #1362 was filed about an hour earlier. Recorded in the campaign log so the gap is not silent.

## Verification

- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 cargo nextest run --features dev` — **9193/9193 pass**, 146 skipped
- **168/168** manifest-gated conformance tests against a real prepared reference, FAIL-set ledgers unmoved

### DIFF
--- a/src/spdi/convert.rs
+++ b/src/spdi/convert.rs
@@ -1394,12 +1394,44 @@ where
             ))
         }
         NaEdit::Identity {
-            sequence: id_seq, ..
+            sequence: id_seq,
+            whole_entity,
         } => {
-            let ref_base = id_seq
-                .as_ref()
-                .map(|s| apply_alphabet(&sequence_to_string(s), alphabet))
-                .unwrap_or_default();
+            // An identity claims the bases under its span — `g.10_12=` asserts
+            // that three bases are unchanged — so its triple must name them, the
+            // same way `Deletion`, `Delins` and `Inversion` fetch a sequence the
+            // input omitted (#1362). This arm used to default to the empty
+            // string, which lost the span and, worse, made the triple
+            // zero-width: a zero-width triple at an interbase is
+            // indistinguishable from an *insertion* at that interbase.
+            //
+            // That is what made the project's own SPDI applier decline any
+            // description containing an unspelled identity. `g.[261_262dup;263=]`
+            // — a real 5'-shuffle output — converted to `262::AG` and `262::`,
+            // two members apparently competing for one interbase, so the applier
+            // refused the whole thing. The identity actually claims base 263 and
+            // the two are disjoint; only the lossy conversion collided them.
+            //
+            // A **whole-entity** identity (`g.=`) is the exception and keeps its
+            // zero-width triple: it asserts the entire reference is unchanged and
+            // names no span, so there is nothing to fetch. Narrowing it to base 1
+            // would invent a claim, and fetching a whole chromosome to prove it
+            // unchanged would be absurd.
+            let ref_raw = match id_seq {
+                Some(seq) => sequence_to_string(seq),
+                None if *whole_entity => String::new(),
+                None => match provider {
+                    Some(p) => fetch_reference_bases(p, &sequence, start_one_based, end_one_based)?,
+                    None => {
+                        return Err(ConversionError::MissingReferenceData {
+                            description: format!(
+                                "Cannot convert identity to SPDI: the unchanged bases at                                  {start_one_based}..={end_one_based} are unknown (no reference                                  data). Spell them (e.g. `g.10A=`) or convert with a provider."
+                            ),
+                        });
+                    }
+                },
+            };
+            let ref_base = apply_alphabet(&ref_raw, alphabet);
             Ok(SpdiVariant::new(
                 sequence,
                 spdi_pos,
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
## Resolved — the commit is now signed

The PR body previously opened with a warning that this branch's commit was unsigned. Signing became available, so the commit has been amended and force-pushed: `45a9826` (unsigned) → `681f76c`, `git log --format=%G?` now reports `G`. I have removed the now-false warning from the body and am recording it here instead.

**The content did not change.** Verified before pushing — the amended commit's tree hash is identical to what was already on the branch:

```
local tree:  2089a0fd7c9d4dc954c89cbd424da7a36d360d86
remote tree: 2089a0fd7c9d4dc954c89cbd424da7a36d360d86
```

So the verification in the body still applies unchanged; this was a signature-only rewrite.

All thirteen branches in this batch are now signed:

```
#1365 G   #1366 G   #1367 G   #1368 G   #1369 G   #1370 G   #1371 G
#1372 G G #1373 G   #1374 G   #1375 G   #1377 G   #1378 G
```

nh13:
## Closing as superseded by #1363

#1363 landed the same fix and closed #1362 (and #1351) at `ab625ac`. It was opened at **04:10 UTC**, ~4.5 h before this PR at **08:36 UTC** — the same issue was implemented twice independently, and the earlier one merged. Not rebasing this: against current `main`, its `convert.rs` diff is 106 insertions / **233 deletions**, and those deletions are the merged fix plus its tests.

Merged `main` also has the larger test surface — 8 inline identity cases plus a dedicated `tests/it/issue_1362_identity_span.rs` covering terminal-span and one-past-the-end boundaries, against this branch's 3 inline cases.

### Where the two implementations genuinely disagree, and why main's answer is better

On a whole-entity `g.=` this branch deliberately kept `0::`, arguing that narrowing to base 1 invents a claim the input never made and that fetching a whole chromosome to answer it is absurd. Both halves of that are right, but it treats "emit `0::`" and "narrow to base 1" as the only options. Merged `main` takes a third: it **declines**, with `UnsupportedEditType` — *"a whole-entity identity (`=`) asserts the entire reference is unchanged and names no interval; SPDI has no representation for it"*.

That is the correct call, and for a reason this PR did not weigh: `0::` does not round-trip. `g.=` → `0::` → `spdi_to_hgvs` → `g.1=`, so "the entire reference is unchanged" silently becomes "base 1 is unchanged". Declining cannot be misread that way. Emitting a triple whose round-trip denotes a different variant is the same class of defect #1362 was filed about, one level up.

### One thing worth salvaging, and a defect in it

This branch's `MissingReferenceData` message is richer in content — it names the span and says what to do:

> Cannot convert identity to SPDI: the unchanged bases at `{start}..={end}` are unknown (no reference data). Spell them (e.g. `g.10A=`) or convert with a provider.

against main's *"unchanged sequence not provided; reference data needed to determine the bases an identity claims"*, which names neither the span nor a remedy.

But the version on `681f76c` **cannot be salvaged as written**: the literal was wrapped across source lines without `\` continuations, so the message carries two ~34-space runs and would render as `...bases at                                  10..=12 are unknown (no reference                                  data)...`. That is a real user-visible defect in this branch, and my verification missed it because nothing asserted the message text — which is also the fix: assert it in a test rather than check it by eye.

Filed as a follow-up so the wording is not lost with this PR.

PR #1380, opened 2026-08-04
fix(normalize): drop a cancelled cis member where the merge creates it

### LINKED ISSUE #1379: Normalizer: #1321's cancelled identity member survives under the 5' shuffle rule
## What's wrong

#1321's identity residue survives under the 5' shuffle rule. #1321 itself is closed and its 3' case is fixed; this is the same defect one shuffle direction over.

```
reference  ("ACGT" x 64) + "TCCCAGAAAAT" + ("ACGT" x 64)

in            g.[261_262insGA;262_263insA;263del]
3' (correct)  g.262_263dup
5' (wrong)    g.[261_262dup;263=]
```

`g.261_262dup` alone denotes the input. The `263=` adds nothing.

## Why the #1321 fix does not reach it

#1321 widened `drop_identity_members_covered_by_siblings` from `claims_reference_bases` to `blocks_sibling_shift`, so a duplication that grew *over* a cancelled member's position counts as covering it. Under 3' the dup settles at `262_263` and covers the residue at 263. Under 5' it settles at `261_262` — 5' of the residue — so nothing covers it and the gate legitimately does not fire.

Coverage was only ever a proxy for "this member is a merge residue". The proxy holds when the sibling grows over the residue and fails when it shifts away from it.

## Mechanism — traced

```
[261_262insGA, 262_263insA, 263del]
  merge_consecutive_edits  ->  [261_262insGA, 263delinsA]
  per-member normalization ->  [261_262dup,   263=]
```

`262_263insA` and `263del` cancel: base 263 is `A`, so deleting it and inserting an `A` beside it restates the reference. The merge into `g.263delinsA` is faithful, and rendering that no-op as `g.263=` is correct for a member considered alone. The residue is created by the merge, so the merge is where it can be recognised — `build_naedit` already handles the sibling case (an insertion anchor with an empty payload) but cannot see this one, because deciding that a payload equals the reference needs the provider.

## Why it cannot be fixed by dropping identity members generally

`g.[258=;260del]` and `g.[259=;259_260insG]` must keep their `=` — that is the guard #1297 installed, and both are pinned in `tests/it/issue_1321_identity_inside_a_duplication.rs`. The distinction is provenance, not position: a user-authored identity never passes through a merge. Adjacency does not work either — it would wrongly drop the `262=` in `g.[262=;263del]`.

## Severity

Criterion 1 of #1235 — non-confluence — on a user-reachable path: `--direction 5prime` is public on both the CLI and the Python bindings. Both spellings are stable fixed points and both re-parse, so neither the idempotency nor the re-parse oracle sees it; only comparing two spellings of one variant does. Pinned as the `#1321` row of `DIVERGENT_UNDER_FIVE_PRIME` in `tests/it/cis_spelling_confluence_gap.rs`.

The residue also has no SPDI triple in the current release, so the cis suite's independent applier cannot evaluate any output containing one (#1351) — which is how the residue's *denotation* went unchecked rather than merely unspelled.

## How it was found

By measuring the confluence property in both shuffle directions rather than in 3' only. Every row of the cis spelling-pair corpus had been blessed against 3'; two of the five converged rows diverge under 5'. #1321 is one of them (#1323 is the other and is a different shape).

nh13:
#1321's identity residue survived under the 5' shuffle rule, so two spellings of one variant did not converge:

```
g.[261_262insGA;262_263insA;263del]
  3'  ->  g.262_263dup            correct
  5'  ->  g.[261_262dup;263=]     the residue survives
  5'  ->  g.261_262dup            with this change
```

`--direction 5prime` is public on both the CLI and the Python bindings, so this is user-reachable.

## Why #1321's own fix does not reach it

#1321 widened `drop_identity_members_covered_by_siblings` from `claims_reference_bases` to `blocks_sibling_shift`, so a duplication that grew *over* a cancelled member's position counts as covering it. Under 3' the dup settles at `262_263` and covers the residue at 263. Under 5' it settles at `261_262`, 5' of the residue, so nothing covers it and the gate correctly does not fire.

Coverage was only ever a proxy for "this member is a merge residue". It holds when the sibling grows over the residue and fails when the sibling shifts away from it.

## Mechanism, traced rather than assumed

```
[261_262insGA, 262_263insA, 263del]
  merge_consecutive_edits  ->  [261_262insGA, 263delinsA]
  per-member normalization ->  [261_262dup,   263=]
```

`262_263insA` and `263del` cancel — base 263 is `A`, so deleting it and inserting an `A` restates the reference. Both steps are correct in isolation: the merge is faithful, and rendering a no-op delins as `=` is right for a member considered alone. The residue is *created by the merge*, so the merge is where it can be recognised.

`reconcile_head` now declines to build a member whose merged anchor restates the reference bases under its own span, as long as the allele keeps another member. This is the same cancellation `build_naedit` already handles one case of — that branch catches an insertion anchor with an empty payload; this catches an anchor whose payload is the reference itself, which needs the provider to see.

I arrived here after two wrong guesses (the `build_naedit` empty-payload branch, then the delins-splitting `IdentityAt` path), both eliminated by instrumenting the pass rather than reasoning about it. The probe output above is what located the real site.

## Provenance, not position, is what keeps #1297's guard

`g.[258=;260del]` and `g.[259=;259_260insG]` must keep their `=`. A user-authored identity never passes through a merge, so both are untouched — they are pinned in `tests/it/issue_1321_identity_inside_a_duplication.rs` and both still pass. Adjacency would not have worked: it would wrongly drop the `262=` in `g.[262=;263del]`. An allele that cancels away entirely also keeps its identity, since the last member is never dropped.

## Re-blessed tests

#1321 now converges under 5', so it leaves `DIVERGENT_UNDER_FIVE_PRIME` (leaving `#1323`, a different shape). Two tests in `cis_spelling_confluence_gap.rs` pinned the residue's consequences and are re-blessed:

- `the_five_prime_divergences_are_non_confluence_and_nothing_worse` — the identity-stripping rewrite and the tolerated-skip count are deleted. Every output is now checked against the applier directly, and an inexpressible one is a failure rather than a known exception.
- `an_identity_member_leaves_the_output_beyond_the_applier` — removed. The member it recorded is no longer emitted, which is what its own failure message asked for ("if the `=` member is gone, delete this test and tighten the one above").

## Oracle direction

**Toward the reference oracles.** Required by the "Output change has a pinned regression test" pre-merge check, since the diff is under `src/normalize/`.

- The output changes only for cis alleles containing a member pair that cancels against the reference. `g.[261_262dup;263=]` → `g.261_262dup`: the two denote the same bases, verified by an applier independent of the normalizer, and the second is the form the merged spelling of the same variant already produced — so this removes a divergence rather than creating one.
- No conformance ledger (`tests/fixtures/**/baseline-failures/*.txt`) and no `.count` baseline moved. The full suite passes with `FERRO_ASSERT_IDEMPOTENT=1` and `FERRO_ASSERT_REPARSE=1`.
- Toward the oracles: an identity member that states nothing is not a form Mutalyzer, biocommons hgvs, or hgvs-rs emit for this input, and dropping it makes the 5' output equal the 3' output's shape.

## Verification

| gate | result |
|---|---|
| `cargo nextest run --features dev` | 9190 passed, 0 failed |
| same with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1` | 9190 passed |
| `cargo clippy --all-features --all-targets -- -D warnings` | clean |
| `cis_allele_confluence_proptest`, 20k cases, `--run-ignored all` | 9/10 pass; the `#[ignore]`d indel property still fails, **byte-identically to the base** |

That last row matters and is stated rather than buried: the ignored property fails before and after this change on the same case — `g.[263_265A[7];264_265insC]`, a repeat member overlapping an insertion. I A/B'd it against `origin/main` at the same case count to confirm the failure is pre-existing and unrelated. It is the next defect in the chain #1321's body describes, not a regression from this change, and `the_ignored_indel_property_still_finds_its_defect` passes, which is the repo's own pin that this property is expected to fail.

## Conflicts with #1363 — read before merging

[#1363](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1363) is open, green, and touches the same two tests, because it addresses the same residue from the other end: it gives an identity edit a real SPDI span so the applier can *express* `g.[261_262dup;263=]`. This PR makes that output not exist.

Both are worth having — #1363 fixes a lossy conversion (`g.10=` → `9::`, a whole-entity `g.=` reading back as `g.1=`) that is independent of whether the normalizer emits `=` members — but they cannot both keep their re-bless. Whichever merges second needs a rebase:

- **If #1363 merges first**, this PR must additionally delete its `an_identity_member_is_redundant_but_no_longer_beyond_the_applier`, whose premise is that the `=` member is still emitted.
- **If this merges first**, #1363 must drop that same test rather than re-blessing it.

Closes #1379

## Oracle direction, stated rather than implied

Asked for in review, and worth being precise about what is and is not measured here.

**Not measured against the external oracles.** Every test in this PR is synthetic — `SyntheticBuilder`/`MockProvider` fixtures with no `FERRO_MANIFEST` — so none of the reference-aware conformance suites (biocommons hgvs, Mutalyzer, VariantValidator) run on this change locally. The committed FAIL-set ledgers under `tests/fixtures/**/baseline-failures/` are **unmoved**, which says the change did not perturb any case those axes already cover; it does not amount to a measurement of oracle agreement on `g.261_262dup` itself.

**What the change is actually claiming.** The defect is *internal confluence*: two spellings of one variant settling on different descriptions under `--direction 5prime`. `g.[261_262insGA;262_263insA;263del]` and `g.263_264insGA` denote the same bases, and before this they normalized to `g.[261_262dup;263=]` and `g.261_262dup`. Both were sequence-preserving and both were fixed points, so neither oracle nor either seam check could see it — only comparing the two spellings does, which is criterion 1 of #1235.

**Direction, on the argument rather than a measurement.** `g.261_262dup` is the form the merged spelling already produced and the form the plain spelling produces unaided; the `=` residue was the outlier, created by the merge and carried by nothing the caller wrote. Dropping it moves the two spellings onto a description that is a strict subset of what one of them already emitted, so it cannot move *away* from an oracle that agreed with that spelling. Whether the oracles agree with `g.261_262dup` in the first place is a separate question this PR does not answer, and the nightly reference-aware job is where it would be answered.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -196,14 +196,14 @@ pub(crate) fn merge_consecutive_edits<P: ReferenceProvider>(
         // Chain ends here. If we'd been growing one, rebuild the head once
         // from the final merged anchor before moving on.
         if head_merged {
-            reconcile_head(&mut output, head_anchor.take().unwrap());
+            reconcile_head(&mut output, head_anchor.take().unwrap(), provider);
             head_merged = false;
         }
         head_anchor = anchor_for_variant(&next);
         output.push(next);
     }
     if head_merged {
-        reconcile_head(&mut output, head_anchor.take().unwrap());
+        reconcile_head(&mut output, head_anchor.take().unwrap(), provider);
     }
     output
 }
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
## Rebased onto current `main` (was CONFLICTING)

Heads-up for whoever owns this branch: I rebased it from another session, `14f8b6a` → `ffc23e6`. It had gone `CONFLICTING` after #1363 (`ab625ac`) and #1365 (`687b284`) landed. Before touching it I confirmed nothing was at risk — clean tree, local tip identical to the pushed tip, no build or edit activity for over 90 minutes.

**No content was changed.** The commit is the same work, replayed on the new base, still signed.

One thing worth recording, because it is a trap: `git merge-tree --write-tree --name-only origin/main <head>` predicted a **content conflict** in `tests/it/cis_spelling_confluence_gap.rs`, but the rebase applied with **no conflict at all**. Merge and rebase pick different bases, so a predicted merge conflict is not a guaranteed rebase conflict — and a clean apply is not evidence the result is *correct*. So I verified the result rather than the apply:

- the rebased comments compose with post-#1363 `main` — the `#1362 gave an identity's SPDI triple the span it claims` reference now describes merged code, and `DIVERGENT_UNDER_FIVE_PRIME` correctly drops `#1321` leaving `#1323`;
- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 cargo nextest run --features dev` — **9223/9223 pass**, 146 skipped;
- `cargo fmt --all --check` and `cargo clippy --all-features --all-targets -- -D warnings` both clean;
- `git status tests/proptest-regressions/` empty — no seed was appended, so the pinned regression cases in `cis_allele_confluence_proptest.rs` still reproduce what they were recorded against.

That last check matters here specifically: this branch touches the cis confluence tests, and re-running those suites is exactly the situation where a strategy change silently re-records seeds.

nh13:
## Conflict with #1392, and it is a disagreement rather than a textual clash

`git merge-tree --write-tree --name-only` (not `mergeStateStatus`, which lags) reports one conflicting file between this PR and **#1392**:

```
tests/it/cis_spelling_confluence_gap.rs
```

The two branches set `DIVERGENT_UNDER_FIVE_PRIME` to different values, and the difference is substantive:

| branch | value |
|---|---|
| this PR | `&["#1323"]` — #1321 removed, because this is the fix that makes it converge |
| #1392 | `&["#1321"]` |

So after both land, `#1321` should not be listed. **Whoever merges second cannot just delete the entry**, though: #1392 rewrites ~1020 lines of `src/normalize/merge.rs`, so it has to be re-verified that #1321 still converges under 5' *given those changes*, rather than assumed from this PR's measurement against the current base.

The file's own test enforces exactly that, in both directions — an entry listed but converging fails with "Remove it from `DIVERGENT_UNDER_FIVE_PRIME`; the gap is closed", and one converging but unlisted fails the other way. So the merge cannot silently land wrong; it will simply be red until the list matches reality. Worth knowing which way to resolve it before that happens.

Everything else about the pair is clean: `merge-tree` reports no other conflicting file, and this PR's production hunk (`src/normalize/merge.rs`, dropping the cancelled member at the merge site) does not overlap #1392's.

Cross-posted to #1392 so both sides have it.

Issue #1114, opened 2026-07-22
fix(parser): the * (star) stop-codon spelling is rejected in strict mode (p.Arg97Profs*23, p.Trp26*)


nh13:
## Summary

In strict parse mode ferro rejects the `*` (asterisk / "star") spelling of the translation-termination codon in protein descriptions, even though the HGVS spec lists `*` as a valid equivalent of `Ter`. The equivalent `Ter` spelling parses fine, and lenient mode already accepts the `*` form and repairs it to `Ter` — so strict mode is rejecting spec-valid syntax.

This is a **false rejection of a legal description**, which is the worst failure class for a parser (it turns away input a downstream caller has every right to submit).

## Repro

```
ferro parse "NP_003997.1:p.Arg97Profs*23"     # ERROR: 1 variant(s) failed to parse   <-- BUG
ferro parse "NP_003997.1:p.Arg97ProfsTer23"   # OK

ferro parse "NP_003997.1:p.Gln151Thrfs*9"     # ERROR: 1 variant(s) failed to parse   <-- BUG
ferro parse "NP_003997.1:p.Gln151ThrfsTer9"   # OK

ferro parse "NP_003997.1:p.Trp26*"            # ERROR: 1 variant(s) failed to parse   <-- BUG (nonsense sub, not a frameshift)
ferro parse "NP_003997.1:p.Trp26Ter"          # OK
```

The rejection is **not** specific to the frameshift-termination slot — a plain nonsense substitution (`p.Trp26*`) is rejected too. `*` is simply not honored as a `Ter` synonym anywhere in strict protein parsing.

Lenient mode, by contrast, already handles all three:

```
ferro parse --error-mode lenient "NP_003997.1:p.Trp26*"          # p.Trp26*         -> p.Trp26Ter
ferro parse --error-mode lenient "NP_003997.1:p.Arg97Profs*23"   # p.Arg97Profs*23  -> p.Arg97ProfsTer23
ferro parse --error-mode lenient "NP_003997.1:p.Gln151Thrfs*9"   # p.Gln151Thrfs*9  -> p.Gln151ThrfsTer9
```

## Spec basis — `*` is valid, not deprecated

The HGVS recommendations present `*` and `Ter` as co-equal valid spellings of the stop codon (submodule `21.0.4-82-g6f85311`):

[...]

Issue #1050, opened 2026-07-17
Add option to control protein stop-codon notation (`Ter` vs `*`) and amino-acid code (3- vs 1-letter) in rendered `p.` names


msto:
Authorship: the content below was drafted by GitHub Copilot CLI (an AI coding agent) and filed via gh under @msto's account. It reflects the agent's analysis, not a statement authored by @msto.

> # Add option to control protein stop-codon notation (`Ter` vs `*`) and amino-acid code (3- vs 1-letter) in rendered `p.` names
>
> ## Summary
>
> `VariantProjector` / `EffectPredictor` currently render protein (`p.`) HGVS names
> with a fixed style: three-letter amino-acid codes and `Ter` for the translation
> stop codon. HGVS nomenclature sanctions **both** `Ter` and `*` for the stop
> codon, and both three- and one-letter amino-acid codes. Please add an option to
> select the style, so downstream consumers that expect `*` / one-letter output
> (e.g. pipelines migrating off other HGVS engines, or matching an established
> report format) don't have to string-munge `p_name`.
>
> ## Current behavior (ferro-hgvs 0.8.1, Python bindings)
>
> `VariantProjection.p_name` is only available pre-rendered, and always uses
> three-letter codes with `Ter`:
>
> | variant | `p_name` today |
> |---|---|
> | missense | `…:p.(Ser4Thr)` |
> | stop-gain | `…:p.(Ile50Ter)` |
> | frameshift | `…:p.(Ser4LeufsTer76)` |
>
> There is no constructor or render-time option to change this. The structured
> `ProteinEffect` / `Consequence` accessors expose classification (SO terms,
> impact) but not a way to re-render the `p.` string in an alternate style, so the
> only workaround is post-hoc string replacement of the finished `p_name`, which
> is brittle (e.g. must avoid touching `Ter` inside `ext`/`fs` tokens correctly).
>
> ## Desired behavior
>
> An opt-in style selection, e.g. one of:
>
> ```python
> # Option A: constructor flags (mirrors the existing `direction=` pattern)
> proj = ferro_hgvs.VariantProjector(
>     reference_json=...,
>     protein_stop="star",        # "ter" (default) | "star"
>     amino_acid_code="one",      # "three" (default) | "one"
> )
>
> # Option B: a render-time / formatter option on the projection result
> proj.project(variant, tx).p_name(stop="star", aa_code="one")
> ```
>
[...]

PR #1381, opened 2026-08-04
test(spec): accept either spec-legal stop spelling as conformant

nh13:
Refs #1079. **Deliberately does not close it** — #1079 is the umbrella spec-divergence tracker, not something one PR resolves.

## What was wrong

`checklist.md:63` states *"`Ter` or `*` should be used to indicate a translation stop codon; the `X` should not be used"*, and the spec then uses both spellings across its own examples (`p.Trp41*` in `general.md:54`, `Ter` forms elsewhere). So the two are co-equal and only `X` is deprecated. ferro renders `Ter` by policy — #1114 made the `*` form *parse*, #1050 added `TerStyle` to *render* it.

Both spec-conformance checks compared ferro's default rendering against the spec's string with `==`:

```rust
// examples/common/spec_harvest.rs
(true, true, Some(expected)) if current == expected => "preserved",
(true, true, Some(_)) => "diverges",

// tests/it/hgvs_spec_normalization_tests.rs
Some(expected) => observed == expected,
```

That reported 17 rows as divergent for a spelling choice the spec itself presents as free. `KNOWN_DIVERGENT_INPUTS` then had to carry all of them under a comment arguing they were fine:

```rust
// 1. `*` is canonicalised to `Ter` in protein descriptions. Deliberate
//    render policy: both spellings are legal on input (#1114 made the `*`
//    form parse), and ferro emits one of them.
```

A ledger entry whose own comment explains why it is not a problem is a sign the comparison is wrong, not that the row needs pinning.

## What this changes

`HgvsVariant::spec_equivalent_renderings` returns the spellings — other than the canonical `Display` form — that the spec admits for the same variant. `classify` and `ferro_produces_the_form_the_spec_states` both consult it, so a row is conformant when ferro renders the spec's form in *any* spelling the spec sanctions.

**Why the rule is in the library rather than in `examples/`.** It is a fact about HGVS, not about either harness; and `tests/` cannot import from `examples/`, so leaving it in the generator would have meant two copies of the `checklist.md:63` reasoning that could drift apart. The first draft of this change *did* only touch the generator — it moved the `status` bookkeeping while the real oracle still called the rows divergences, and `ferro_produces_the_form_the_spec_states` failed with all 17. That failure is the reason for the current shape.

The one-letter amino-acid code is deliberately **excluded** even though `ProteinRenderStyle` can emit it: it is a separate axis, no other spec-stated form in this corpus uses one-letter codes, and admitting it would let rows pass on a form the spec never states here.

## Measurement

| status | before | after |
| --- | ---: | ---: |
| `diverges` | 36 | **19** |
| `preserved` | 697 | **714** |
| `correctly-rejected` | 99 | 99 |
| `false-acceptance` | 52 | 52 |
| `needs-reference` | 50 | 50 |
| **total** | **934** | **934** |

No ferro behaviour changed and no default output moved. `Display` still renders `Ter`, guaranteed by construction — every protein `Display::fmt` delegates to `fmt_styled(f, ProteinRenderStyle::default())`, and the default is `TerStyle::Ter`.

## Two of the 19 stay divergent, and that is the point

If the widening were too broad it would have cleared all 19. It cleared 17:

- **`NP_003997.1:p.W24*`** — the spec states a **one-letter** code (`W24`); ferro renders `Trp24Ter`. Excluded axis, per above.
- **`p.[Lys79*;Lys79Asn]`** — ferro reorders the cis members into coordinate order (`p.[Lys79Asn;Lys79Ter]`). A structural difference no stop spelling accounts for. Its parenthesised sibling `p.[(Lys79*;Lys79Asn)]` *does* now pass, since there the glyph was the only difference.

That split — spelling differences cleared, structural ones left standing — is the evidence the change does not launder real divergences. `an_alternate_does_not_excuse_an_unrelated_difference` pins it as a unit test.

## Verification

- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 cargo nextest run --features dev` — **9204/9204 pass**, 146 skipped
- `cargo fmt --all --check` clean; `cargo clippy --all-features --all-targets -- -D warnings` clean (exit 0; the only warning is the pre-existing `proc-macro-error2` future-incompat notice from a dependency)
- Both generated artifacts regenerated (`generate_spec_fixture`, `generate_spec_enumeration`); `spec_enumeration_tests`' `DIVERGENCE_BUDGET` / `PASSING_CENSUS` are **unmoved** — this touches the normalization fixture's statuses only.
- New tests: `spec_equivalent_renderings_offers_the_star_stop_spelling` and `..._excludes_the_canonical_form` (library); four `classify` cases including the taxonomy-unchanged-without-alternates control and the no-laundering guard.

## Interaction with the open queue

`git merge-tree --write-tree` (not `mergeStateStatus`, which lags) reports **clean against `origin/main` and against all 13 open campaign branches**. In particular it does not collide with #1373: that PR edits `spec_enumeration_tests.rs`'s projection statuses, whereas this one edits `hgvs_spec_normalization_tests.rs`, which #1373 never touches. No stacking required.

### DIFF
--- a/examples/common/spec_harvest.rs
+++ b/examples/common/spec_harvest.rs
@@ -663,23 +663,36 @@ pub mod prefix {
 
 pub mod classify {
     /// Status taxonomy is a function of (parse_ok, normalize_ok, spec_expected,
-    /// current == spec_expected). `spec_expected: None` is the spec-rejects-this
-    /// sentinel (set by override or by `<code class="invalid">` extraction).
+    /// whether ferro renders the spec's form). `spec_expected: None` is the
+    /// spec-rejects-this sentinel (set by override or by `<code class="invalid">`
+    /// extraction).
     ///
-    /// | spec_expected | parse | normalize | current == spec_expected | status              |
-    /// |---------------|-------|-----------|--------------------------|---------------------|
-    /// | `None`        | err   | —         | —                        | `correctly-rejected`|
-    /// | `None`        | ok    | —         | —                        | `false-acceptance`  |
-    /// | `Some`        | err   | —         | —                        | `parse-error`       |
-    /// | `Some`        | ok    | err       | —                        | `needs-reference`   |
-    /// | `Some`        | ok    | ok        | true                     | `preserved`         |
-    /// | `Some`        | ok    | ok        | false                    | `diverges`          |
+    /// | spec_expected | parse | normalize | renders spec's form | status              |
+    /// |---------------|-------|-----------|---------------------|---------------------|
+    /// | `None`        | err   | —         | —                   | `correctly-rejected`|
+    /// | `None`        | ok    | —         | —                   | `false-acceptance`  |
+    /// | `Some`        | err   | —         | —                   | `parse-error`       |
+    /// | `Some`        | ok    | err       | —                   | `needs-reference`   |
+    /// | `Some`        | ok    | ok        | yes                 | `preserved`         |
+    /// | `Some`        | ok    | ok        | no                  | `diverges`          |
+    ///
+    /// "Renders the spec's form" is a match against `current` **or** any of
+    /// `equivalent_renderings`. Where the spec presents two spellings as equals,
+    /// a description that uses either one conforms, so pinning the comparison to
+    /// ferro's default rendering alone would manufacture divergences out of a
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

PR #1385, opened 2026-08-04
fix(normalize): apply the strict-mode ladder on both public exits

### LINKED ISSUE #1382: fix(normalize): normalize_with_diagnostics skips the strict-mode rejection ladder
A strict-configured `Normalizer` enforces different contracts through its two public entry points.

## The defect

`Normalizer` has exactly two public normalizing methods, and only one applies the strict-mode rejection ladder:

| method | route | strict ladder |
| --- | --- | --- |
| `normalize()` | → `normalize_core_checked` | **yes** |
| `normalize_with_diagnostics()` | → `normalize_core_canonical` directly | **no** |

`normalize_core_checked` is where the `should_reject_*` ladder lives — `RefSeqMismatch`, W5002/W5003/W5004/W4004/W4005/W4006, `EINTRONIC`, and the `ReducedCapabilityNoGenome` promotion. `normalize_core_canonical` is just `normalize_core` + `canonicalize_from_sequence`, so `normalize_with_diagnostics` returns a result none of those rejections ever saw.

So in strict mode the same variant is rejected through `normalize()` and accepted through `normalize_with_diagnostics()`. In the default lenient config every `should_reject_*` is false and the ladder is a no-op, which is why this has stayed hidden.

## Reproducer already in the tree

`tests/it/issue_336_position_past_end.rs::strict_mode_n_in_bounds_does_not_emit_warning` was passing a variant whose stated reference base is wrong. Its fixture is `ACGTACGTACGTACGTACGT`, so `n.10` is `C`, but the test authored `NR_TEST.1:n.10G>C` — and it passed under `NormalizeConfig::strict()`, because the `RefSeqMismatch` was never promoted on that path. Through `normalize()` strict mode rejects it.

It surfaced only because #1366 began running the seam oracles on that exit: `FERRO_ASSERT_IDEMPOTENT` then failed re-normalizing an output that `normalize()` would have refused outright:

```
FERRO_ASSERT_IDEMPOTENT: normalized output failed to re-normalize
  input: NR_TEST.1:n.10G>C
  once:  NR_TEST.1:n.10G>C
  error: Reference mismatch at 10-10: expected G, found C
```

#1366 corrected that fixture to `n.10C>G` (the test is about the position, not the base) and wired the oracles to both exits. It deliberately left the ladder asymmetry alone, since routing it is a behavioural change to strict mode.

## Blast radius

37 test files under `tests/` call `normalize_with_diagnostics`. Any of them running under a strict config may be relying, knowingly or not, on the missing ladder — so the fix needs the full suite plus all three oracles, and each newly-rejecting case triaged as either a real defect or a fixture that was wrong.

## Definition of done

- Both public exits apply the same strict-mode rejection ladder.
- A regression test asserts that a strict `Normalizer` rejects the same input through `normalize()` and `normalize_with_diagnostics()` — the shape above (`RefSeqMismatch`) is a concrete one.
- Full suite green under `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1`, with any newly-rejected fixture individually justified.
- If some rejection deliberately must not apply to the diagnostics exit, that exception is documented at the seam rather than left implicit.

nh13:
Closes #1382

`Normalizer` has exactly two public normalizing methods and only one applied the strict-mode rejection ladder.

| method | route | strict ladder |
| --- | --- | --- |
| `normalize()` | → `normalize_core_checked` | yes |
| `normalize_with_diagnostics()` | → `normalize_core_canonical` | **no** |

`normalize_core_checked` is where the `should_reject_*` rungs live — `RefSeqMismatch`, W5002/W5003/W5004/W4004/W4005/W4006, `EINTRONIC`, `InitiatorMetCanonicalization`, and the `ReducedCapabilityNoGenome` promotion. `normalize_core_canonical` is just `normalize_core` + `canonicalize_from_sequence`, so a strict-configured `Normalizer` rejected a variant through one entry point and accepted the same variant through the other.

The asymmetry was invisible in the default lenient config, where every rung is a no-op — which is why it survived.

## The fix

`normalize_with_diagnostics` now calls `normalize_core_checked`. Because that method itself routes through `normalize_core_canonical`, the returned variant is the same canonical one as before and `infos` still describes the shift from the input to *that* variant, so the ladder is added without moving any output. The warnings it returns are those **not** promoted to hard errors, so a lenient caller still receives everything this entry point exists to report.

It also subsumes #1366's separate `assert_seam_oracles` call on this exit. That call existed precisely because the exit bypassed `normalize_core_checked`; now that it does not, the oracles run once at that method's own exit. Keeping both would re-run all three on every diagnostics normalization, including `assert_idempotent`'s full re-normalization.

## How it was found

`FERRO_ASSERT_IDEMPOTENT` failed while #1366 was wiring the oracles to this exit:

```
FERRO_ASSERT_IDEMPOTENT: normalized output failed to re-normalize
  input: NR_TEST.1:n.10G>C
  once:  NR_TEST.1:n.10G>C
  error: Reference mismatch at 10-10: expected G, found C
```

`issue_336_position_past_end::strict_mode_n_in_bounds_does_not_emit_warning` was authoring `n.10G>C` against fixture `ACGTACGTACGTACGTACGT`, whose position 10 is `C`, under `NormalizeConfig::strict()` — and passing, because the `RefSeqMismatch` was never promoted on that path. #1366 corrected that fixture (the test is about the position, not the base) and left this asymmetry for its own PR.

## Tests (written first, watched failing)

`tests/it/issue_1382_diagnostics_strict_ladder.rs` asserts the two exits **agree**, rather than asserting each rejects — a test of the latter shape would go green if a future change made `normalize()` stop rejecting, which is the opposite of the point.

Before the fix, both agreement tests failed with `left: true, right: false` (i.e. `normalize()` rejected, the diagnostics exit did not) and the three controls passed:

- `a_strict_config_rejects_a_reference_mismatch_through_both_exits` — the reproducer.
- `a_strict_config_rejects_a_past_end_position_through_both_exits` — a second rung (W4004), so this is not only about `RefSeqMismatch`.
- `the_default_lenient_config_still_accepts_on_both_exits` — control; the default must stay permissive.
- `a_strict_config_accepts_a_correct_description_through_both_exits` — control; the fix must not simply reject more.
- `the_diagnostics_exit_still_reports_its_warnings` — control; a fix that returned an empty warning list would satisfy every assertion above while destroying the method's purpose.

## Blast radius, measured

37 files under `tests/` call `normalize_with_diagnostics`, so this could have newly rejected in any of them. Before rebasing onto #1366 the full suite showed **exactly one** failure — `strict_mode_n_in_bounds_does_not_emit_warning`, the known-bad fixture #1366 fixes. After rebasing, with that fixture corrected upstream:

- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1 cargo nextest run --features dev` → **9228 passed, 0 failed**

So nothing else in the suite was relying on the missing ladder.

## Docs

`CLAUDE.md`'s oracle section said `assert_seam_oracles` is called from two places because `normalize_with_diagnostics` reaches `normalize_core_canonical` directly. That is no longer true, so it now describes the single call site and records why the second one existed. #1366's test doc in `src/normalize/mod.rs` got the same correction.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -82,14 +82,20 @@ a second time with it set; the nightly reference-aware job sets it too, which is
 where the manifest-backed conformance corpora are actually covered.
 
 The check runs from `Normalizer::assert_seam_oracles`, which all three oracles
-share. It is called from two places, because ferro has two normalization exits and
-only one of them is the "shared" one: `normalize_core_checked`, covering
-`normalize()` and every `VariantProjector` path (the projection-driven
-genomic/coding/protein axes), and `normalize_with_diagnostics`, which reaches
-`normalize_core_canonical` directly and so was returning results no oracle had
-inspected — for all three checks, not just the newest. Its verification pass
-re-enters normalization, so a thread-local `IN_IDEMPOTENCY_CHECK` guard breaks the
-recursion — the inner call skips its own check.
+share, at the single exit of `normalize_core_checked`. That covers every public
+normalization: `normalize()`, `normalize_with_diagnostics()`, and every
+`VariantProjector` path (the projection-driven genomic/coding/protein axes).
+
+It is one call site again as of #1382. `normalize_with_diagnostics` used to reach
+`normalize_core_canonical` directly, which bypassed both the oracles and — the
+actual defect — the strict-mode rejection ladder, so #1366 had to call
+`assert_seam_oracles` from it separately. Routing it through
+`normalize_core_checked` fixes both at once, and the extra call would now just
+re-run all three oracles on that path.
+
+The idempotency oracle's verification pass re-enters normalization, so a
+thread-local `IN_IDEMPOTENCY_CHECK` guard breaks the recursion — the inner call
+skips its own check.
 
 #### Normalization re-parse oracle
 
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -2341,14 +2341,29 @@ impl<P: ReferenceProvider> Normalizer<P> {
         &self,
         variant: &HgvsVariant,
     ) -> Result<NormalizeResult, FerroError> {
-        // Through `normalize_core_canonical`, not `normalize_core`: this exit
-        // must emit the same canonical variant `normalize()` does, and `infos`
-        // must describe the shift from the input to *that* variant.
-        let (result, warnings) = self.normalize_core_canonical(variant)?;
-        // This exit does not pass through `normalize_core_checked`, so the oracles
-        // have to be run here too — otherwise a public entry point hands back a
-        // normalized variant none of the three ever saw.
-        self.assert_seam_oracles(variant, &result);
+        // Through `normalize_core_checked`, not `normalize_core_canonical`: this
+        // is a *public* exit, so it owes the caller the same strict-mode contract
+        // `normalize()` does (#1382). It used to call the canonical core directly
+        // and so applied none of the `should_reject_*` ladder, which meant a
+        // strict-configured `Normalizer` rejected a variant through `normalize()`
+        // and accepted the very same one through here. The asymmetry was invisible
+        // in the default lenient config, where every rung is a no-op.
+        //
+        // `normalize_core_checked` routes through `normalize_core_canonical`
+        // itself, so the variant this returns is the same canonical one as before
+        // and `infos` still describes the shift from the input to *that* variant —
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

PR #1387, opened 2026-08-04
chore(review): stop implying three suites cover both normalization exits

### LINKED ISSUE #1384: chore(review): normalize path instruction names coverage suites that miss the diagnostics exit
The `src/normalize/**/*.rs` review instruction prescribes a coverage list that cannot cover one of the two public normalization exits.

## The gap

`.coderabbit.yaml`'s `src/normalize/**/*.rs` `path_instructions` block says:

> Confirm changes are covered by `tests/it/idempotency_tests.rs`, `tests/it/normalize_property_tests.rs`, and `tests/it/normalize_idempotency_proptest.rs`.

Measured on `origin/main`:

```
$ for f in tests/it/idempotency_tests.rs tests/it/normalize_property_tests.rs \
           tests/it/normalize_idempotency_proptest.rs; do
    echo "$f: $(grep -c normalize_with_diagnostics $f)"
  done
tests/it/idempotency_tests.rs: 0
tests/it/normalize_property_tests.rs: 0
tests/it/normalize_idempotency_proptest.rs: 0

$ grep -rl normalize_with_diagnostics tests/ | wc -l
37
```

All three named suites go through `normalize()`. None of them touches `normalize_with_diagnostics`, which is the other public normalizing method — and, per #1382, the one that does not apply the strict-mode rejection ladder.

So a normalization change confined to the diagnostics exit can satisfy the instruction's coverage list while being wholly unexercised by it. That is close to what happened in #1366: the coverage that actually mattered came from the other 37 files, not from the three the instruction names.

## Why it matters

These instructions are lenses a reviewer is told to apply. One that names a specific, checkable set of files is high-value precisely because it is concrete — which also means a stale set quietly redirects attention. A reviewer who confirms "covered by the three named suites" has confirmed less than they think.

## Suggested fix

Either add a suite that exercises `normalize_with_diagnostics`, or reword the clause so the three are named as the idempotency-corpus entry points rather than as sufficient coverage — and say that the diagnostics exit needs its own coverage. Mentioning that there are two public exits is the part worth encoding, since that is the fact the current wording obscures.

## Definition of done

- The instruction no longer implies the three named suites are sufficient coverage for a `src/normalize/**` change.
- The two-public-exit distinction is stated where a reviewer will read it.
- No behavioural change; this is review configuration only.

nh13:
Closes #1384

The `src/normalize/**/*.rs` review instruction told a reviewer to *"confirm changes are covered by `tests/it/idempotency_tests.rs`, `tests/it/normalize_property_tests.rs`, and `tests/it/normalize_idempotency_proptest.rs`"*. None of those three exercises `normalize_with_diagnostics`, one of `Normalizer`'s two public normalizing methods.

Measured on `origin/main`:

```
$ for f in tests/it/idempotency_tests.rs tests/it/normalize_property_tests.rs \
           tests/it/normalize_idempotency_proptest.rs; do
    echo "$f: $(grep -c normalize_with_diagnostics $f)"
  done
tests/it/idempotency_tests.rs: 0
tests/it/normalize_property_tests.rs: 0
tests/it/normalize_idempotency_proptest.rs: 0

$ grep -rl normalize_with_diagnostics tests/ | wc -l
37
```

So a change confined to that exit could satisfy the prescribed coverage list while being wholly unexercised by it. That is close to what happened while reviewing the in-bounds oracle (#1366): the coverage that mattered came from the other 37 files, not from the three the instruction names. It also matters more than it looks, because that exit is the one #1382 found was skipping the strict-mode rejection ladder.

## What changed

- The three suites are now described as the **idempotency-corpus entry points**, explicitly `normalize()`-only — a starting point rather than sufficient coverage, and *no* coverage for a change on the other exit.
- The two-public-exit fact is stated, with the instruction to establish which exit a change reaches before judging its coverage. That is the fact the old wording obscured.
- The `FERRO_ASSERT_IDEMPOTENT` sentence now attributes the check to `Normalizer::assert_seam_oracles` and names all three covered paths. Worded mechanism-neutrally on purpose: it is accurate both on `main` today and after #1382 moves the diagnostics exit behind `normalize_core_checked`.
- Every `Flag:` clause is unchanged.

Review configuration only — no behavioural change, no source touched.

Verified the file still parses and the instruction survived intact:

```
$ python3 -c "
import yaml
d=yaml.safe_load(open(\".coderabbit.yaml\"))
t=[p for p in d[\"reviews\"][\"path_instructions\"] if p[\"path\"]==\"src/normalize/**/*.rs\"][0][\"instructions\"]
assert \"Flag: off-by-one\" in t and \"bypasses the shared checked exit\" in t
print(\"OK\")
"
OK
```

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -135,11 +135,20 @@ reviews:
         (`boundary.rs`), repeat shuffling, edit merging (`merge.rs`), overlap
         resolution (`overlap.rs`), and rule application (`rules.rs`). The core
         invariant is IDEMPOTENCY — normalize(normalize(x)) == normalize(x) —
-        enforced in debug builds by `FERRO_ASSERT_IDEMPOTENT=1` at the shared
-        exit of `normalize_core_checked`, which covers both `normalize()` and
-        every `VariantProjector` path. Confirm changes are covered by
-        `tests/it/idempotency_tests.rs`, `tests/it/normalize_property_tests.rs`,
-        and `tests/it/normalize_idempotency_proptest.rs`. Flag: off-by-one in
+        enforced in debug builds by `FERRO_ASSERT_IDEMPOTENT=1`, which runs from
+        `Normalizer::assert_seam_oracles` and covers every public normalization:
+        `normalize()`, `normalize_with_diagnostics()` and every
+        `VariantProjector` path.
+        `Normalizer` has exactly TWO public normalizing methods — `normalize()`
+        and `normalize_with_diagnostics()` — and a change can affect one without
+        the other, so establish which of them the change reaches before judging
+        its coverage. `tests/it/idempotency_tests.rs`,
+        `tests/it/normalize_property_tests.rs` and
+        `tests/it/normalize_idempotency_proptest.rs` are the idempotency-corpus
+        entry points and go through `normalize()` ONLY — none of the three calls
+        `normalize_with_diagnostics`, so they are a starting point rather than
+        sufficient coverage, and for a change on that exit they are no coverage at
+        all (roughly three dozen other `tests/it/` files do call it). Flag: off-by-one in
         shuffle or boundary windows; wrong shift direction or strand handling;
         mishandled circular/mitochondrial wraparound; overlap merges that
         silently drop or reorder edits; any path that can make a second


nh13:
@coderabbitai pause

PR #1390, opened 2026-08-04
fix(spdi): name the span and the way out when bases are unspelled

### LINKED ISSUE #1388: spdi: the unspelled-identity decline names neither the span nor a remedy, and no test asserts it
`hgvs_to_spdi` declines an unspelled identity without a provider, which is correct (#1362/#1363). The message it declines with names neither the span nor a remedy:

```
unchanged sequence not provided; reference data needed to determine the bases an identity claims
```

`src/spdi/convert.rs`, the `NaEdit::Identity` arm's `None => match provider { None => ... }` branch.

A caller hitting this has two ways out — spell the bases in the description, or supply a provider — and neither is stated. The span is also already in scope (`start_one_based`, `end_one_based`), so it costs nothing to name what could not be resolved. Suggested content:

```
Cannot convert identity to SPDI: the unchanged bases at {start_one_based}..={end_one_based}
are unknown (no reference data). Spell them (e.g. `g.10A=`) or convert with a provider.
```

That wording comes from the superseded #1378 (see https://github.com/fulcrumgenomics/ferro-hgvs/pull/1378#issuecomment-5182746609), which is why this is filed rather than left in a closed PR.

**Two things to get right when doing it.**

The literal must use `\` line continuations. The #1378 version wrapped it without them, so the rendered message carried two ~34-space runs (`...bases at                                  10..=12 are unknown...`). Nothing caught that because no test asserted the text.

So: **assert the message in a test.** That is the actual gap here — the arm is covered for *behaviour* (it errors) but not for *what it says*, which is why a wrapping defect could sit in a branch through a full green suite. A test pinning the rendered string both fixes the wording and stops the next such regression.

Worth checking the sibling arms in the same function while there (`Inversion`, `Repeat`, `Delins`): they have the same shape and may be equally thin.

nh13:
Closes #1388.

## What was wrong

Six arms of `hgvs_to_spdi` need reference bases the description left implicit — `Duplication`, `Deletion`, `Delins`, `Identity`, `Inversion` and `Repeat`. Every one declined with a string that named neither the interval it could not resolve nor either way forward:

```
unchanged sequence not provided; reference data needed to determine the bases an identity claims
```

A caller who hits this has two routes: spell the bases in the description, or pass a provider that can look them up. Neither is guessable from "reference data needed", and `start_one_based`/`end_one_based` are already in scope, so naming the span costs nothing.

The `Delins` arm had also drifted into a different sentence shape from its five siblings — `"Cannot convert delins to SPDI: deleted sequence of length 3 is unknown"` — computing a length used for nothing else.

## What this changes

One `unspelled_bases_error` helper over an `UnspelledBases` shape enum, so all six decline in one voice:

```
cannot convert to SPDI: the unchanged bases at 10..=12 are unknown (no reference data).
Spell them in the description (e.g. `g.10A=`) or convert with a reference provider.
```

## The `Repeat` arm cannot offer the first remedy

The issue asked me to check the siblings, and this is what that turned up. The other five read an optional sequence off the edit and only consult the provider when it is absent, so "spell it" is a genuine way out. `Repeat` has no such field: the notation states the repeat **unit** (`g.10AC[3]`), never the span its tract currently occupies. So `spelled_example()` returns `Option`, and the repeat message names only the provider:

```
cannot convert to SPDI: the pre-expansion repeat-tract bases at 10..=12 are unknown
(no reference data). Convert with a reference provider: the repeat notation states the
unit, not the span its tract currently occupies, so the bases cannot be spelled in the
description.
```

Offering a spelling there would have sent the caller after a description they cannot write — worse than the thin message it replaces. `the_repeat_tract_offers_only_the_provider` pins that it does not.

## The tests are the actual point

These arms were already covered for *behaviour* — they error — so nothing read what they **said**. That is exactly how a wrapped literal missing its `\` continuations survived a full green suite in the superseded #1378, embedding two ~34-space runs in a user-facing message (see https://github.com/fulcrumgenomics/ferro-hgvs/pull/1378#issuecomment-5182746609). Closing that gap is why this PR exists, so the tests assert properties rather than just pinning strings:

| test | what it prevents |
| --- | --- |
| `spelled_examples_are_parseable` | recommending a form ferro itself rejects. Every `Some` example is parsed; these literals are exercised nowhere else. |
| `the_decline_message_has_no_embedded_whitespace_runs` | the #1378 defect. No double space, newline or tab — over **all six** shapes, since each message is assembled from a different pair of literals and the defect is per-literal. |
| `each_shape_declines_distinguishably` | two shapes reading identically, so a log line identifies which wall was hit. |
| `a_provider_less_conversion_declines_with_the_span` | the helper existing but not being wired in — a unit test on the helper alone would pass with the arms unconverted. |

## Verification

- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1 cargo nextest run --features dev` — **9233/9233 pass**, 146 skipped
- No test, doc or Python binding asserted any of the six old strings (grepped before changing them), so nothing outside this file needed updating.
- `git merge-tree --write-tree` clean against `origin/main` and against every open branch that touches `src/spdi/` (#1374) or that I have in flight.

### DIFF
--- a/src/spdi/convert.rs
+++ b/src/spdi/convert.rs
@@ -1318,11 +1318,11 @@ where
                 None => match provider {
                     Some(p) => fetch_reference_bases(p, &sequence, start_one_based, end_one_based)?,
                     None => {
-                        return Err(ConversionError::MissingReferenceData {
-                            description:
-                                "duplication sequence not provided; reference data needed to determine duplicated bases"
-                                    .to_string(),
-                        });
+                        return Err(unspelled_bases_error(
+                            UnspelledBases::Duplicated,
+                            start_one_based,
+                            end_one_based,
+                        ));
                     }
                 },
             };
@@ -1346,11 +1346,11 @@ where
                 None => match provider {
                     Some(p) => fetch_reference_bases(p, &sequence, start_one_based, end_one_based)?,
                     None => {
-                        return Err(ConversionError::MissingReferenceData {
-                            description:
-                                "deleted sequence not provided; reference data needed to determine deleted bases"
-                                    .to_string(),
-                        });
+                        return Err(unspelled_bases_error(
+                            UnspelledBases::Deleted,
+                            start_one_based,
+                            end_one_based,
+                        ));
                     }
                 },
             };
@@ -1385,15 +1385,11 @@ where
                 None => match provider {
                     Some(p) => fetch_reference_bases(p, &sequence, start_one_based, end_one_based)?,
                     None => {
-                        let del_len = end_one_based
-                            .saturating_sub(start_one_based)
-                            .saturating_add(1) as usize;
-                        return Err(ConversionError::MissingReferenceData {
-                            description: format!(
-                                "Cannot convert delins to SPDI: deleted sequence of length {} is unknown (no reference data)",
-                                del_len
-                            ),
-                        });
+                        return Err(unspelled_bases_error(
+                            UnspelledBases::DeletedInDelins,
+                            start_one_based,
+                            end_one_based,
+                        ));
                     }
                 },
             };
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1360, opened 2026-08-04
chore: release v0.13.0

nh13:
## Representation stability

Any PR that moves a normalized output should carry a `Representation-Change:`
trailer (see CONTRIBUTING.md). Those are collected into the **Representation
changes** section of the changelog below.

Reviewer checklist:

- [ ] Either that section lists every output-moving change in this cycle, or
      there were none and it is absent.
- [ ] Each entry says whether the affected inputs were previously **rejected**
      (no stored string, so free) or previously **accepted** (a real migration
      for the consumer).

A commit under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` that
moved output without declaring it means the trailer is missing. Fix that before
releasing, not after.

---

### Added

- *(dev)* add --verify-spdi to the normalized-corpus harness ([#1515](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1515))
- *(normalize)* record provenance when reported members are coalesced ([#1486](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1486))
- *(normalize)* coalesce payload-alignment splits back into one delins ([#1470](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1470))
- *(normalize)* add the member-count-minimal partitioner behind a flag ([#1464](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1464))
- *(spdi)* reference-aware SPDI and apply-to-reference for an encoding-invariant key ([#1374](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1374))
- *(normalize)* make the sequence-first derivation authoritative for cis alleles ([#1392](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1392))
- *(effect)* [**breaking**] drop the dead inversion validator that contradicted the live one ([#1359](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1359))

### Fixed

- *(normalize)* apply a junction barrier the translate cannot express ([#1516](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1516))
- *(spdi)* name the caller's anchor when a repeat names no tract ([#1496](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1496))
- *(normalize)* type a split-created span inv on the first pass ([#1493](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1493))
- *(normalize)* see cis members that span a region boundary ([#1506](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1506))
[...]

PR #1391, opened 2026-08-04
ci: delegate the last six workflows' cargo caches to rust-cache

### LINKED ISSUE #1218: ci: cargo caches never refresh, and the 10 GiB budget blocks the obvious fix

nh13:
Refs #1218. **Stacked on #1369** — base is `ci/issue-1218-reclaim-cache-budget`, not `main`. Deliberately does not close #1218; see [Why the verification is not here](#why-the-verification-is-not-here).

Completes what #1370 started in `ci.yml`/`coverage.yml`: the last seven hand-rolled `actions/cache` steps, across six workflows, now use `Swatinem/rust-cache`. The measurement behind it is recorded once, above the equivalent step in `ci.yml` (#1370) — each entry cached a whole `target/` and had grown to ~3 GiB, and `actions/cache` skips the save on a primary-key hit, so a lockfile-only key froze that tree permanently.

| workflow | steps | `rust-cache` key |
| --- | ---: | --- |
| `nightly-mutalyzer.yml` | 1 | `nightly-mutalyzer` |
| `nightly-perf.yml` | 1 | `nightly-perf` |
| `nightly-soak.yml` | 1 | `nightly-soak` |
| `external-validation.yml` | 2 | `external-validation-api`, `external-validation-spec` |
| `fuzz.yml` | 1 | `fuzz-${{ matrix.target }}` |
| `deploy-local.yml` | 1 | `deploy-local-web-hgvs-rs` |

It stacks on #1369 because converting these was deferred out of that PR to avoid an adjacent-hunk conflict with its removal of the manifest caches. `git merge-tree` is clean against `origin/main` and against #1370 (their files are disjoint), so merge order between #1369→this and #1370 does not matter.

## Three defects found in the steps being replaced

Each came out of reading the step, not from assuming the conversion was mechanical.

**`external-validation`'s two jobs shared one cache entry.** Both used the identical `${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}` — no job suffix — despite installing different toolchain components (the second adds `clippy`) and building different targets. Whichever job ran first decided what the other restored, and the primary-key hit then skipped the save, so neither could ever correct it.

**`nightly-mutalyzer` and `nightly-perf` ended in a bare `${{ runner.os }}-cargo-` restore-key.** That is a prefix of every cargo cache in the repository, so either job could restore a `target/` built by any other job under a different profile. It matters most for `nightly-perf`, which gates on a startup-time budget — a timing gate must not inherit that variance.

**`fuzz.yml` restored on top of its own freshly-installed binary.** `cargo install cargo-fuzz` runs in the step *above* the cache, and the old key cached `~/.cargo/bin/`, so the restore could replace the new binary with whatever version first populated the key.

## Two options that are load-bearing, not defaults

`workspaces: fuzz` — `fuzz/Cargo.toml` declares its own empty `[workspace]` to stay out of the root package, so cargo-fuzz builds into `fuzz/target/`. Without this, `rust-cache` would cache the root workspace's `target/`, which that job never populates. The old key cached `fuzz/target/` explicitly; this preserves it.

`cache-bin: "false"` on `nightly-soak` and `fuzz` — `rust-cache` caches `~/.cargo/bin` unless told otherwise, so both hazards above would have survived the conversion. On `nightly-soak` this replaces an ordering constraint with a structural one: the old step documented that `install-action` had to run *after* the restore or a stale `cargo-nextest` would overwrite it, and since that job writes the archive eight shards consume, producer and consumer must agree on the version. Not caching binaries removes the hazard rather than dodging it.

## Deliberately unchanged

**`fuzz.yml`'s corpus cache.** Not a cargo cache — it accumulates a fuzzing corpus across runs and is keyed on `github.run_id` with a prefix restore-key precisely so that every run saves. That is the correct shape for what it stores, and converting it would destroy the accumulation.

**`ci.yml` and `coverage.yml`** — #1370's scope.

**`fuzz.yml`'s `cargo install cargo-fuzz` still compiles every run.** It already did: the install ran *before* the restore, so it never read the cache. Making it cheap needs a step reorder or a prebuilt binary and is a separate change; flagging it rather than bundling a guess.

## Why the verification is not here

#1218's definition of done is a measurement, and it cannot be taken from a PR. Two reasons, both measured today rather than assumed:

```
$ gh api /repos/fulcrumgenomics/ferro-hgvs/actions/cache/usage
12972353164 bytes = 12.08 GiB in 17 entries      (cap: 10 GiB — already over)

$ gh api ".../actions/caches?ref=refs/pull/1381/merge" --jq .total_count
0
$ gh api ".../actions/caches?ref=refs/pull/1377/merge" --jq .total_count
0
```

PR refs hold **zero** entries under current eviction pressure, so nothing a PR writes survives for a second run to restore — a warm-cache measurement here would measure nothing. And the DoD requires *consecutive runs* with the nightly's caches intact, which is a `main` property. #1369 landing is what frees the headroom that makes any of it measurable.

Note also that **#1360 (the release PR) currently owns ~5 GiB of the 12.08 GiB** across six entries on `refs/pull/1360/merge`, which is a large part of why nothing else stays warm.

So: this PR is the code half. The measurement, and `Closes #1218`, belong to a follow-up once #1369 and #1370 are on `main` — expect the first post-merge run to be a miss by construction and read the second.

## Verification

- `actionlint .github/workflows/*.yml` — clean
- every workflow file still parses as YAML
- the `Swatinem/rust-cache` pin is checked, not copied on faith: annotated tag `v2.9.1` → `c19371144df3bb44fab255c43d04cbc2ab54d1c4`, matching both the comment and the pin #1370 uses
- no hand-rolled `actions/cache` step remains in any file this PR touches; the only one left in the six is the fuzz corpus, above

### DIFF
--- a/.github/workflows/deploy-local.yml
+++ b/.github/workflows/deploy-local.yml
@@ -56,16 +56,27 @@ jobs:
           export PATH="$HOME/.pixi/bin:$PATH"
           pixi install
 
-      - name: Cache cargo registry
-        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
+      # Cargo caching via `rust-cache`; rationale recorded above the equivalent
+      # step in `ci.yml` (#1218).
+      #
+      # This job builds `--release --features "web-service hgvs-rs"`, a feature
+      # set no other job builds, so its dependency tree is its own. That is an
+      # argument for caching it and for keying it separately — which the key this
+      # replaces did get right (`-web-hgvs-rs-`, suffix before hash, scoped
+      # restore-key). What it could not fix is that `actions/cache` skips the save
+      # on a primary-key hit, so the tree froze at the first run per lockfile.
+      #
+      # `cache-bin: false` because this job runs on a PERSISTENT self-hosted
+      # runner (`runs-on: [self-hosted, …, ferro-deploy]`) with its own rustup
+      # install — the step below sources `$HOME/.cargo/env`. `rust-cache` caches
+      # `${CARGO_HOME}/bin` unless told not to, so on a hosted runner that is a
+      # throwaway directory, but here it would restore a snapshot of the
+      # operator's toolchain over the live one. Dependency artifacts are what is
+      # worth caching; the binaries on this box are not ours to replace.
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
         with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-            target
-          key: ${{ runner.os }}-cargo-web-hgvs-rs-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-cargo-web-hgvs-rs-
+          key: deploy-local-web-hgvs-rs
+          cache-bin: "false"
 
       - name: Build release binary (with hgvs-rs)
         run: |
--- a/.github/workflows/external-validation.yml
+++ b/.github/workflows/external-validation.yml
@@ -114,14 +114,20 @@ jobs:
         with:
           toolchain: stable
 
-      - name: Cache cargo registry
-        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
+      # Cargo caching via `rust-cache`; rationale recorded above the equivalent
+      # step in `ci.yml` (#1218).
+      #
+      # Note what the key it replaces did NOT have: a job suffix. Both jobs in
+      # this workflow used the identical `${{ runner.os }}-cargo-<lock-hash>`,
+      # so they shared one entry despite installing different toolchain
+      # components and building different targets — whichever ran first decided
+      # what the other restored, and `actions/cache` then skipped the save on the
+      # key hit so neither could correct it. `rust-cache` keys per job by
+      # default; the `key` below only adds a legible name so
+      # `gh api .../actions/caches` stays diagnosable.
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
         with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-            target
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          key: external-validation-api
 
       - name: Run API comparison tests
         run: |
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai pause

PR #1392, opened 2026-08-04
feat(normalize): make the sequence-first derivation authoritative for cis alleles

nh13:
Six declines in `canonicalize_from_sequence` refused a correct derivation because the answer disagreed with how the input was *spelled*, handing control to the per-member pipeline — which is non-confluent by design. This removes them, porting first the capabilities each one protected.

`Refs #1235.` It does **not** close it: criterion 2 ("no normalized cis allele contains overlapping or out-of-order members") is violated by **#1325**, which lives in `demote_repeats_spanning_siblings` and is untouched by anything here.

## What it buys, measured

Over every length-2..7 whole-span inversion and every spelling — 21,756 haplotypes, 849,752 normalizations:

| | diverging | converged to `inv` |
|---|---|---|
| before | **3,120** | 13,500 |
| after | **0** | 19,692 |

A random non-inversion corpus stays at 0. `cis_spelling_confluence_gap.rs`'s `DIVERGENT` table — eight spelling pairs pinned as known #1235 violations — is now **empty**.

## The commits

| commit | what |
|---|---|
| `54049a5` | reject a strictly overlapping piece pair instead of mangling it |
| `4af72d2` | let any lone `g.`/`m.` member reach the derivation |
| `2e539e5` | clamp a derived terminal insertion at the sequence bounds |
| `429319c` | type a derived tandem insertion as a duplication |
| `abfb87f` | let a derivation collapse to a single insertion |
| `fafab2c` | guard the derived dup's source span, and re-name what moved |
| `775e6d2` | lower repeat members instead of refusing them |
| `e925226` | coalesce a whole-block inversion after the weight bound |

## Three design points worth a reviewer's attention

**The inversion coalesce runs *after* the changed-columns weight bound, not inside `partition_block`.** The bound compares derived weight against the *input's* weight, so a rule that widens a piece before the bound runs makes the comparison spelling-dependent. Measured: `g.[259G>T;261T>A;263A>C]` weighs 3 and is refused while `g.259_263delinsTTAAC` weighs 5 and is accepted — one variant, two normal forms. Running the coalesce on an already-accepted partition is the placement `apply_coding_codon_exception` already uses twelve lines below. **The weight bound and `partition_block` are byte-identical to the base.**

**Grounded in `delins.md:5` and `delins.md:46-47`, not `general.md:56`.** A delins is defined as a replacement "which is *not* a substitution or inversion", so a spanning `delins` over a reverse-complement span is not an admissible spelling — converting it is a definitional correction rather than a preference. `general.md:56` does not support this: it ranks the type of *one description of one change*, never a multi-member allele against a spanning description, and it ranks substitution *above* inversion in any case. The doc now names it in order to refute it.

**`sequence_first_pass` now declines when strict mode rejects the input for having no canonical form.** Removing the lone-insertion bail unmasked a pre-existing gap: `apply_edits_to_window`'s refusal is interbase-geometric while `detect_overlap_conflicts` flags W5002 in HGVS-coordinate space, and the two disagree for `dup`+`sub` at one base — so `g.[24dup;24C>G]` (strict-rejected) was normalizing into `g.23_24insG` (strict-accepted). The derivation is authoritative over *how* to spell a variant, not over *whether* an ill-formed input has a spelling at all. `tests/it/strict_rejection_survives_normalization.rs` pins the property rather than the reproducer, with a control (`g.[23dup;23A>G]`, same shape one base left, no conflict) that must still merge.

## Verification

- suite **9258/9258**
- all three oracles (`FERRO_ASSERT_IDEMPOTENT` / `REPARSE` / `IN_BOUNDS`) **9258/9258**
- conformance axis, manifest-backed: **16/16** manifest-gated rows green
- `cargo fmt --all --check`, `cargo clippy --features dev --all-targets -- -D warnings` clean
- `tests/proptest-regressions/` unchanged apart from one added seed, which carries its issue reference

Note the axis runner's `test(axis_)` filter is a substring match — it selects 204 rows, of which only 16 are the manifest-gated conformance drivers. The rest run identically without a manifest.


nh13:
@coderabbitai pause

nh13:
## Conflict with #1392, and it is a disagreement rather than a textual clash

`git merge-tree --write-tree --name-only` (not `mergeStateStatus`, which lags) reports one conflicting file between this PR and **#1392**:

```
tests/it/cis_spelling_confluence_gap.rs
```

The two branches set `DIVERGENT_UNDER_FIVE_PRIME` to different values, and the difference is substantive:

| branch | value |
|---|---|
| this PR | `&["#1323"]` — #1321 removed, because this is the fix that makes it converge |
| #1392 | `&["#1321"]` |

So after both land, `#1321` should not be listed. **Whoever merges second cannot just delete the entry**, though: #1392 rewrites ~1020 lines of `src/normalize/merge.rs`, so it has to be re-verified that #1321 still converges under 5' *given those changes*, rather than assumed from this PR's measurement against the current base.

The file's own test enforces exactly that, in both directions — an entry listed but converging fails with "Remove it from `DIVERGENT_UNDER_FIVE_PRIME`; the gap is closed", and one converging but unlisted fails the other way. So the merge cannot silently land wrong; it will simply be red until the list matches reality. Worth knowing which way to resolve it before that happens.

Everything else about the pair is clean: `merge-tree` reports no other conflicting file, and this PR's production hunk (`src/normalize/merge.rs`, dropping the cancelled member at the merge site) does not overlap #1392's.

Cross-posted to #1392 so both sides have it.

PR #1393, opened 2026-08-04
fix(normalize): ask the codon gate about the window the repeat is emitted over

### LINKED ISSUE #1389: insertion_to_repeat gates the codon rule on the pre-alignment tract, not the emitted repeat window
`insertion_to_repeat` answers the #1210 codon gate for the tract it *discovered*, but `normalize_repeat` then phase-aligns that tract before emitting. The two windows can differ, so a repeat whose **emitted** window is wholly CDS-resident can be written with a non-codon-length unit — which the codon rule forbids.

The gate is computed from `ref_start`/`ref_end_excl` (the raw `find_tandem_extent` tract), then handed to `normalize_repeat`, which owns the direction-aware phase alignment and can return a shifted window. Nothing re-asks the residency question about the window that actually gets emitted.

## Repro

```rust
// src/normalize/rules.rs
insertion_to_repeat(b"CTGTGTT", 4, b"TGTGTG", /*is_coding=*/true, Some((3, 6)), ShuffleDirection::ThreePrime)
```

Observed: `Some(GT[5])` over the 1-based span `3_6`.

- The raw `TG` tract is 1-based `2..=5`. Against `cds_span = (3, 6)` that is *not* wholly resident (`2 < 3`), so `gate_is_coding` is `false` and the codon gate is switched off.
- `normalize_repeat` then emits `GT[5]` over `3..=6`, which **is** wholly inside the CDS — with a two-base unit. `codon_blocks_repeat` would have been true had the gate been asked about this window, routing to the literal `ins` instead.
- The equal-scoring `GT` rotation cannot rescue it: selection keeps the strictly-greater candidate, so the first rotation found wins.

The 5' direction is unaffected in this example (it emits `TG` over `2_5`, which genuinely is not CDS-resident, so gating it off is correct there) — the two directions therefore disagree about whether the repeat is admissible.

## Pre-existing, not introduced by #1355

Confirmed byte-identical on `origin/main` (`12bba7c`) and on #1368's head: both print `ThreePrime: unit=GT count=5 span=3_6`. #1368 moved the gate to run per rotation rather than once for the winner, but it did not change *which window* the gate is asked about — that code moved verbatim into `resolve_repeat_rotation`. Raised by CodeRabbit on #1368 and deliberately scoped out of it, since #1368 is an ordering fix and this is a change to the #1210 seam.

Worth noting #1355's `RepeatNormResult::Insertion => continue` slightly widens exposure: a codon-gated winner no longer aborts the function, so a lower-scoring rotation can now be emitted — and its gate is answered on its own raw tract, with the same mismatch.

## Suggested fix

Resolve the emitted window first, derive CDS residency from *its* boundaries, then apply the codon gate. `normalize_repeat`'s window/unit/count are computed independently of `is_coding` — the flag only routes the result variant (`Repeat` vs `Insertion`/`Deletion`) — so the emitted window can be learned with the gate lifted and the gate then applied to it.

Preferring a two-call shape (probe for the window, then delegate with the correctly-derived gate) over re-implementing `!unit.len().is_multiple_of(3)` at the call site would keep `normalize_repeat` the single source of truth for the codon decision, per #866. The early `ref_count <= best.ref_count` skip means only score-improving rotations would pay for the extra call.

## Testing

This is an edit-selection change, so per the repo's conventions it needs:

- a pinned CDS-boundary fixed-point regression in both shuffle directions, at the `Normalizer` level (not just a direct `insertion_to_repeat` call);
- a reference-oracle movement statement — the manifest-gated conformance axes (`axis_normalized`, `axis_genomic`, `axis_errors`, both idempotency axes) are the FAIL-set ledgers that could move, and may need re-blessing;
- the full suite green under `FERRO_ASSERT_IDEMPOTENT=1` and `FERRO_ASSERT_REPARSE=1`.

nh13:
Closes #1389. Base is `main`.

This was authored stacked on #1368 (both change `insertion_to_repeat`), but #1368 merged as `9d09582` while it was in flight, so it is rebased onto `main` and stands alone. `git rebase` correctly recognised #1368's commit as already applied.

## The defect

`insertion_to_repeat` answered the #1210 codon gate for the tract `find_tandem_extent` discovered, then handed that verdict to `normalize_repeat` — which owns the direction-aware phase alignment and can return a **shifted** window. Nothing re-asked the residency question about the window that actually gets emitted.

Measured at the `Normalizer` level on this PR's own fixture, a `TG` tract laid across the CDS start:

| `NM_T1389.1:c.3_4insTGTG` | before | after |
|---|---|---|
| 3′ | `c.1_6GT[5]` | **`c.3_6dup`** |
| 5′ | `c.-1_5TG[5]` | unchanged |

`c.1_6` is wholly inside the CDS and `GT` is two bases, so the before-form is not a description HGVS admits (`repeated.md`'s codon rule). The raw tract starts at `c.-1`, in the 5′UTR, so gating on *it* switched the rule off.

The 5′ answer must not move, and does not: its window genuinely includes `c.-1`, so the codon rule correctly does not apply. The two directions still emit different forms, and that is **not** a defect — they align to different windows and residency legitimately differs between them. What was wrong was computing the verdict from a window neither of them emitted.

## The fix

Probe `normalize_repeat` with the gate lifted to learn the emitted window, derive CDS residency from *its* boundaries, then delegate with the correct gate — the two-call shape the issue suggests.

Two calls rather than re-implementing `!unit.len().is_multiple_of(3)` at the call site, which keeps `normalize_repeat` the single source of truth for the codon decision (#866); the probe only ever learns *which window* to ask about. This is sound because `is_coding` enters `normalize_repeat` at exactly **one** point — `codon_blocks_repeat` — which routes the result variant only, while the window, unit and counts are computed before it and do not depend on it. I verified that rather than taking it from the issue: it is the sole `is_coding` reference in the function.

No probe is issued when `cds_span` is `None`, since there is then no window question to ask, so genomic / `n.` callers and the direct unit tests are untouched. `insertion_to_repeat`'s `ref_count <= best.ref_count` skip means only score-improving rotations pay for the second call.

## Pre-existing, not introduced by #1355

A/B'd against the pre-fix code by restoring the unmodified `rules.rs` (then #1368's head `6b3308a`, now `main`) and re-running the same fixture: the identical `c.1_6GT[5]` comes out. #1368 moved the gate to run per rotation rather than once for the winner, but did not change *which window* it asks about. Raised by CodeRabbit on #1368 and deliberately scoped out of it, since that PR is an ordering fix and this is a change to the #1210 seam.

## Tests

Four at the `Normalizer` level (`tests/it/issue_1389_codon_gate_emitted_window.rs`), plus two direct-call unit tests in `rules.rs`:

- **the defect** — asserts the new form *and* `assert_ne!` against the old one, so the equality cannot pass for the wrong reason if a later change moves the canonical answer to a third form;
- **the 5′ control** — pins that the fix does not over-reach into a window whose residency answer was already right;
- **fixed point in both directions** — the assertion that would have caught this independently of any expected string;
- **a codon-length unit inside the CDS is still a repeat** (`c.13_24CAG[6]`) — pins that the fix did not turn CDS-residency itself into a refusal.

That last one was initially written with a homopolymer (`insCCCCCC`) and **failed**, correctly: `smallest_repeat_unit` reduces it to the one-base `C`, so it was gated on unit length and asserted the opposite of what it claimed. The fixture now carries a genuinely three-periodic `CAG` tract, and the note is in the test so the next person does not repeat it.

## Reference-oracle movement: none

The manifest-gated conformance axes were re-run against the real prepared reference with all three oracles set: **168/168 pass**, so no FAIL-set ledger moved and nothing needs re-blessing.

That the manifest was genuinely exercised rather than silently skipped is itself checked — those suites skip green when it is absent, so a bare "168/168" would prove nothing. The same filter emits **46** skip messages without `FERRO_MANIFEST` and **0** with it.

## Verification

- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1 cargo nextest run --features dev` — **9242/9242 pass**, 146 skipped
- manifest-gated conformance axes — **168/168**
- `tests/proptest-regressions/` unchanged — no seed appended
- `git merge-tree --write-tree` clean against `origin/main`, #1368, #1373, #1380 and #1392

### DIFF
--- a/src/normalize/rules.rs
+++ b/src/normalize/rules.rs
@@ -511,17 +511,60 @@ fn resolve_repeat_rotation(
     // with no CDS context (genomic / `n.`) and the unit tests below are
     // unaffected.
     //
-    // Frames line up: `ref_start` / `ref_end_excl` are 0-based indices into the
-    // transcript-frame `ref_seq`, and `cds_span` is 1-based inclusive in that
-    // same frame, so the 1-based inclusive tract is
-    // `(ref_start + 1) ..= ref_end_excl`. Per rotation since #1355, which is also
-    // more correct than it was: each rotation finds its own tract, so each gets
-    // its own residency answer instead of the winner's.
+    // Per rotation since #1355, which is also more correct than it was: each
+    // rotation finds its own tract, so each gets its own residency answer instead
+    // of the winner's.
+    //
+    // #1389: the window to ask about is the one the repeat is **emitted** over,
+    // which is not the tract `find_tandem_extent` discovered. `normalize_repeat`
+    // owns a direction-aware phase alignment and can return a shifted window, and
+    // the two disagree exactly at a CDS boundary: for `CTGTGTT` with
+    // `cds_span = (3, 6)`, the raw `TG` tract is 1-based `2..=5` and so reads as
+    // non-resident (`2 < 3`), switching the gate off — while the emitted window is
+    // `3..=6`, wholly CDS-resident, and got written as `GT[5]` with a two-base
+    // unit that `repeated.md`'s codon rule forbids. The 5' direction of the same
+    // fixture emits `TG` over `2..=5`, which genuinely is not resident, so before
+    // this the two directions disagreed about whether the repeat was admissible.
+    //
+    // So probe for the emitted window with the gate lifted, then delegate with the
+    // gate derived from it. Two calls rather than re-implementing the codon test
+    // here, which keeps `normalize_repeat` the single source of truth for the
+    // codon decision (#866) — the probe is only ever used to learn *which window*
+    // to ask about. This is sound because `is_coding` enters `normalize_repeat` at
+    // exactly one point, `codon_blocks_repeat`, which routes the result variant
+    // only: the window, unit and counts are all computed before it and do not
+    // depend on it. `insertion_to_repeat`'s `ref_count <= best.ref_count` skip
+    // means only score-improving rotations pay for the second call.
+    //
+    // Frames line up: `normalize_repeat` returns 1-based inclusive HGVS positions
+    // and `cds_span` is 1-based inclusive in the same transcript frame, so the
+    // comparison is direct. With no `cds_span` there is no window question to ask
+    // — the caller's precomputed verdict stands and no probe is issued, so callers
+    // with no CDS context (genomic / `n.`) and the unit tests below are
+    // unaffected.
+    let target = ref_count + added_copies;
+    let end_pos_incl = ref_end_excl - 1;
     let gate_is_coding = match cds_span {
         Some((cds_start, cds_end)) => {
-            let tract_start = ref_start as u64 + 1;
-            let tract_end = ref_end_excl as u64;
-            tract_start >= cds_start && tract_end <= cds_end
+            let probe = normalize_repeat(
+                ref_seq,
+                ref_start,
+                end_pos_incl,
+                &unit,
+                target,
+                // Gate lifted: this call answers "which window?", not "may it be
+                // repeat notation?". Passing the real flag would let a gated
+                // result hide the window the gate has to be asked about.
+                false,
+                direction,
+            );
+            match probe {
+                RepeatNormResult::Repeat { start, end, .. } => start >= cds_start && end <= cds_end,
+                // Unreachable for the same reason the delegation below asserts:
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1395, opened 2026-08-05
test(cis): draw a corpus prefix by default, the full sweep in CI

### LINKED ISSUE #1295: Tests: one cis sweep is 94% of the local suite's wall clock (76.9s of 82.1s)

nh13:
Closes #1295. **Stacked on #1392** — base is `fix/issue-1235-confluence-ladder`, not `main`, because it edits the sweep files that PR rewrites.

## The measurement

The three exhaustive cis sweeps were 92% of a local `cargo nextest run`, measured on this PR's parent:

| | before | after (default) |
|---|---:|---:|
| local suite (`not test(proptest)`) | **86.6s** | **27.5s** |
| `cis_junction_crossing_shift` | 79.6s | 18.3s |
| `repeat_span_sibling_overlap` | 33.1s | 2.8s |
| `issue_1254_sibling_crossing_shift` | ~23s | 1.5s |

No test is left over nextest's 60s `SLOW` threshold. With all three oracles set, the full suite goes 9260/9260 in 56.7s.

## How

`sweep_seeds(full)` decides how much of the shared corpus to draw:

| `FERRO_SWEEP_SEEDS` | seeds |
|---|---|
| unset | 4-seed prefix |
| `full` | the sweep's full count |
| a number | that count (for bisecting) |

CI's `Exhaustive sweeps` job sets `full`. It is the **only** job that reaches these modules — `test` and `test-oracle` negate `SWEEP_FILTER` — so the exhaustive coverage still runs on every PR, and the comment at that env block says so, since moving a sweep out of `SWEEP_FILTER` would silently drop it to the prefix.

Only sequence diversity is reduced, never shapes: the loop bounds over member spellings, positions and directions are untouched. That is the axis worth keeping — each sweep's own notes record that every blocking defect found so far lived in a shape the generator could not emit rather than in a sequence it did not draw.

## The full count goes back to 48, and what that costs

`cis_junction_crossing_shift` had been halved to 24 seeds to pay for a shape widening, and its comment recorded the debt precisely: `FIVE_PRIME_DUP_DEL_SEQUENCE_CHANGES` is an exact count over that corpus, so halving the seeds halved the diversity behind its pinned zero — and it named *this issue's knob* as what would restore it. So the full count is 48 again.

The pinned zero holds at 48, which is the measurement saying that restoring the diversity revealed no hidden defect.

**That is not free.** Running the `sweeps` job's exact selection with its three oracles:

```
24 seeds (what CI runs today)    87.5s
48 seeds (this PR)              183.2s
```

It is one of ~35 checks and runs in parallel, so this is +96s on one job rather than on the critical path. If the trade is unwanted, the single-token change is `sweep_seeds(48)` → `sweep_seeds(24)` in `cis_junction_crossing_shift.rs`, which keeps the entire local speedup and simply declines to restore the diversity. Flagging it rather than deciding it quietly.

## Why the pinned counts did NOT become profile-dependent

#1295 raises this as the hazard against exactly this change — *"the pinned numbers would need to become profile-dependent, which is its own maintenance hazard"* — so it deserves an answer rather than an assumption.

`sweep_sequences` is **prefix-stable**: `sweep_sequences(8)` is exactly the first sixteen entries of `sweep_sequences(16)`. A reduced run therefore enumerates a strict *subset* of the full run's cases, and:

- `overlapping.is_empty()` / `changed.is_empty()` hold over any subset if they hold over the set, so the prefix cannot pass what the full corpus would fail — it can only find fewer failures, never different ones;
- `FIVE_PRIME_DUP_DEL_SEQUENCE_CHANGES` is pinned at **zero**, subset-stable for the same reason. Had that residual still been the 74 it once was, this knob *would* have forced a profile-dependent pin — it is only safe because the residual was driven to zero first;
- `skipped * 10 < checked` is a ratio and does not move with corpus size at all.

The one genuinely seed-dependent assertion is each sweep's case-count floor, so those become per-seed. That is strictly better than the absolutes they replace: `checked > 100_000` over 24 seeds sat ~64% under the measured 11,520 cases per seed, and a per-seed floor scales with the knob rather than going vacuous at one end and failing at the other.

## Tests

- `sweep_sequences_is_prefix_stable` — split out of `sweep_sequences_is_deterministic` and named, because a property the knob's soundness rests on should fail under its own name rather than inside a test about determinism. It now also checks the default prefix specifically.
- `sweep_seeds_defaults_to_the_prefix_and_honours_full` — all three input forms, including the case-insensitive sentinel and the "a sweep asking for fewer than the default gets its own count" edge.

**No proptest strategy is touched and no seed file moved.** `sweep_sequences` is a plain deterministic generator, not a proptest strategy, so the seed-invalidation hazard that applies to `cis_allele_confluence_proptest.rs` does not apply here — `git status tests/proptest-regressions/` is clean.

## Verification

- default profile, all three oracles — **9260/9260 pass**, 146 skipped, 56.7s
- `FERRO_SWEEP_SEEDS=full`, all three oracles — **9260/9260 pass**, 146 skipped
- `cargo fmt --all --check`, `cargo clippy --all-features --all-targets -- -D warnings`, `actionlint` all clean
- `git merge-tree --write-tree` clean against `origin/main`, #1392, #1370, #1391 and #1393

One inherited conflict to note: against **#1380** there is a clash in `tests/it/cis_spelling_confluence_gap.rs`, a file this PR does not touch. It is the pre-existing #1380 ↔ #1392 disagreement (flagged on both of those PRs), inherited by stacking, not introduced here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -691,6 +691,17 @@ jobs:
           name: nextest-archive-soak
       - name: Run the exhaustive sweeps with the normalization self-checks
         env:
+          # The full corpus, not the prefix a local run takes (#1295). The
+          # sweeps were 92% of a local `cargo nextest run` — 79.6s of an 86.6s
+          # suite for `cis_junction_crossing_shift` alone — so `sweep_seeds`
+          # defaults to a 4-seed prefix and this job is what asks for all of it.
+          #
+          # This job is the ONLY place that sets it, which is why the exhaustive
+          # coverage still runs on every PR: `test` and `test-oracle` negate
+          # `SWEEP_FILTER` and so never reach these modules at all. If a sweep is
+          # ever moved out of `SWEEP_FILTER`, it silently drops to the prefix
+          # here — check both when editing that filter.
+          FERRO_SWEEP_SEEDS: full
           FERRO_ASSERT_IDEMPOTENT: "1"
           FERRO_ASSERT_REPARSE: "1"
           # The sweeps are where this one earns its keep: they enumerate members
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -194,6 +194,47 @@ Wheel Test, abi3 floor — and neither `Test oracle` nor `Exhaustive sweeps` is
 among them. So an oracle fire shows up as a failed job that a human must not
 merge past, not as a ruleset-enforced merge block.
 
+#### Exhaustive cis sweeps: `FERRO_SWEEP_SEEDS`
+
+The three exhaustive cis sweeps — `cis_junction_crossing_shift`,
+`repeat_span_sibling_overlap` and `issue_1254_sibling_crossing_shift` — draw a
+deterministic corpus of 20-mers from `sweep_sequences`. They dominated a local
+run: measured at 79.6s of an 86.6s suite for `cis_junction_crossing_shift`
+alone, so any `cargo nextest run` cost ~90s regardless of what changed (#1295).
+
+They now take a **4-seed prefix of that corpus by default**, and CI's
+`Exhaustive sweeps` job asks for all of it:
+
+```bash
+cargo nextest run --features dev                        # 4-seed prefix (fast)
+FERRO_SWEEP_SEEDS=full cargo nextest run --features dev # the full corpus, as CI runs it
+FERRO_SWEEP_SEEDS=12   cargo nextest run --features dev # an explicit count, for bisecting
+```
+
+Local suite: **86.6s → 27.5s**, with no test left over nextest's 60s SLOW
+threshold.
+
+**Only the sequence diversity is reduced — never the shapes.** The loop bounds
+over member spellings, positions and directions are untouched, which is the
+axis worth keeping: each sweep's notes record that every blocking defect found
+so far lived in a shape the generator could not emit, not in a sequence it did
+not draw.
+
+**Why the pinned counts did not become profile-dependent**, which is the hazard
+#1295 flags against exactly this change. `sweep_sequences` is prefix-stable
+(`sweep_sequences_is_prefix_stable`), so a reduced run enumerates a strict
+*subset* of the full run's cases. An `is_empty()` assertion therefore cannot
+pass at the prefix while failing at the full corpus, and
+`FIVE_PRIME_DUP_DEL_SEQUENCE_CHANGES` is pinned at **zero**, which is
+subset-stable for the same reason — had that residual still been 74, this knob
+would have had to make it profile-dependent. The one genuinely seed-dependent
+assertion in each sweep is its case-count floor, which is why those are written
+per-seed rather than as absolutes.
+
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai pause

PR #1399, opened 2026-08-05
fix(normalize): demote a repeat that outgrew its tract, on both branches of the guard

### LINKED ISSUE #1235: Normalizer: cis-allele members normalized independently, not from the resulting sequence (root cause of #1229–#1234)

### LINKED ISSUE #1394: Normalizer: a repeat whose growth exceeds its tract cannot be demoted when a sibling deletion shifts into it, so the pair overlaps
## What's wrong

A repeat's tract swallows a sibling **deletion** that 3'-shifted into it, and the demotion that exists to prevent that declines.

```
reference  ("ACGT" x 64) + "CAGGCAAACAGTGAAG" + ("ACGT" x 64)
           the `A` tract at 262_264 is three bases, between `C` at 261 and `C` at 265

in   NC_TEST.1:g.[262del;263_264insAA;264_265insAA]
out  NC_TEST.1:g.[262_264A[7];264del]
```

The deletion claims base 264, which is inside the tract the repeat names. The property reports `members overlap — SPDI at 263 starts before 264`.

## Mechanism

This is the **base-claiming branch of the exact guard #1325 fixed on the junction branch**, and it was masked behind #1325.

The two insertions each grow the three-base tract, so they coalesce and re-spell as a tract-wide copy count — growth of four over a tract of three. The deletion 3'-shifts through the same homopolymer to the tract's 3' end. `demote_repeats_spanning_siblings` is the pass for that: its `spans_sibling_bases` test fires (the deletion's span meets the tract), but it then declines at its `start < a.start` guard, because the equivalent duplication would sit at `start = a.end - length + 1 = 264 - 4 + 1 = 261`, which is 5' of the tract's own start. There is no duplication *over the tract* that can express four bases added to a three-base tract, so the pass bails and leaves the repeat spanning the deletion.

**#1325's repair does not extend to this branch, and that is measured rather than assumed.** #1325 re-spells such a repeat as the insertion its growth stands for (`repeat_growth_as_insertion`), which restores a junction that `clamp_sibling_crossing_junctions` can then bound. That fix is deliberately scoped to `spans_sibling_junction` and keeps the decline for `spans_sibling_bases`, because an insertion claims no bases and so blocks no sibling's shift — demoting before the clamps releases a sibling deletion that #1296 had deliberately clamped out of the tract. Widening the same route to the base-claiming branch (a one-line change to that gate) was tried against the full suite and costs exactly what the hazard predicts:

```
FAIL  issue_1296_repeat_claims_its_bases::a_deletion_does_not_shift_into_a_repeats_tract
FAIL  issue_1296_repeat_claims_its_bases::the_three_member_form_reaches_the_same_result
FAIL  cis_spelling_confluence_gap::the_eight_spelling_pairs_still_diverge   (ledger moved)
```

So the repair needs a form that both expresses growth larger than the tract *and* keeps the barrier the sibling's clamp depends on — or it needs to run at a point where that barrier is no longer load-bearing. `demote_coincident_tract_repeats` already occupies such a late slot for the two-repeats-on-one-tract case, and its doc comment explains why the early slot was unsafe; that is the most likely place to look, but the deletion here has no coincident repeat to group with.

## Pre-existing

Verified by A/B against `origin/main` at `dfea7d3`, with `src/normalize/merge.rs` and `src/normalize/mod.rs` pristine: the same input yields the identical `NC_TEST.1:g.[262_264A[7];264del]`. #1325's fix does not cause it and does not change it.

## Severity

**Loud.** Criterion 2 of #1235 — "no normalized cis allele contains overlapping or out-of-order members" — is directly violated, and the pair has no well-defined resulting sequence.

## How it was found

By `cis_allele_confluence_proptest::an_indel_haplotype_normalizes_to_its_own_sequence` at **231,801 successes**, in an unseeded 500,000-case stream, immediately after #1325 was fixed. It is the seventeenth defect that property has surfaced, each masked behind the one before it: #1286 → #1287 → #1290 → #1292 → #1296 → #1301 → #1304 → #1297 → #1308 → #1312 → #1316 → #1320 → #1321 → #1323 → #1325 → this.

**CI's soak would not have found it.** All eight seeded shards (1,000,000 cases at the fixed seeds in `ci.yml`) pass clean with #1325 fixed, as do the first two unseeded 500k streams; it took a third, decorrelated stream to reach. That is the case for the nightly's unseeded counterpart, and a reminder that a clean seeded soak is not evidence the chain has ended.

Depth, which remains the only measure of how close the chain is to ending: #1320 at 27,633 successes, #1321 at 114,694, #1323 at 143,034, #1325 at 132,216, this at 231,801. The last four sat in one band and this one is roughly double the previous — still no trend that forecasts an end.

Seed (committed):

```
cc 6ec5686a0f5e290e507e200e264cfd28c157a1683b777803a951fb8dedf54f20 # shrinks to haplotype = IndelHaplotype { core: "CAGGCAAACAGTGAAG", events: [Delete { index: 5 }, Insert { index: 6, bases: "AA" }, Insert { index: 7, bases: "AA" }] }
```

## Relationship to #1235

Blocks criterion 1 — it is what the property now fails on, and it is why `an_indel_haplotype_normalizes_to_its_own_sequence` stays `#[ignore]`d after #1325 — and is a direct instance of criterion 2.

nh13:
Closes #1394. Refs #1235 — this is **criterion 2** ("no normalized cis allele contains overlapping or out-of-order members"), and it is now enforced in CI for indel haplotypes for the first time.

## What lands

Two commits, each one branch of a single guard.

`demote_repeats_spanning_siblings` re-spells a tract-wide copy count back into the edit it grew out of, so the junction-based passes can see the member again. It offers two target forms — a deletion or a duplication of the tract's 3'-most bases — and **neither can express growth larger than the tract**, since a duplication over a three-base tract adds at most three bases. So it declined at `start < a.start` and left the repeat spanning its sibling.

**#1325 — the sibling is a junction** (`b96536b`):

```
reference  ("ACGT" x 64) + "GATCATAAATTCAGC" + ("ACGT" x 64)
g.[262_263insAA;263_264insAA;264_265insC]  ->  g.[263_265A[7];264_265insC]
```

**#1394 — the sibling is a deletion** (`9d72213`):

```
reference  ("ACGT" x 64) + "CAGGCAAACAGTGAAG" + ("ACGT" x 64)
g.[262del;263_264insAA;264_265insAA]  ->  g.[262_264A[7];264del]
```

Both now re-spell as the insertion the growth stands for (`repeat_growth_as_insertion`, the payload #1316 already builds), reaching `g.[263_264insAAAA;264_265insC]` and `g.262_264A[6]`.

## Why declining cost more than a spelling

The tracing is the part worth reviewing. **The first merge pass already reaches the right answer** in both cases — the clamp bounds the members short of the sibling and the coalesce combines them. The *second* pass re-normalizes each member in isolation, which re-spells the result as a tract-wide copy count; and a repeat carries no junction (`junction_of` returns `None` for `NaEdit::Repeat`), so the clamp can no longer see it. This demotion is the only route back to the clamp, so declining discards a bound that had already been computed correctly one pass earlier.

## The scoping, which is the delicate part

An insertion claims no bases and so blocks no sibling's shift. Demoting a repeat to one *before* the clamps run is what released a sibling deletion that #1296 had deliberately clamped out of a tract, so this route cannot simply be opened to every sibling. Two narrowings, both measured rather than assumed:

- **Widening it to base-claiming siblings outright** costs `issue_1296_repeat_claims_its_bases::{a_deletion_does_not_shift_into_a_repeats_tract, the_three_member_form_reaches_the_same_result}` plus a moved `cis_spelling_confluence_gap` ledger. Tried against the full suite; that is why #1325 ships scoped to the junction branch.
- **Gating #1394 on "do the two overlap *now*"** also re-breaks those two #1296 tests. In #1296 the deletion reaches the tract only by its own 3' shift (`272_273 -> 273_274`), and the clamp that runs immediately after pulls it back out — the overlap is the ordinary pre-clamp state the clamp exists to repair, so demoting on it destroys the barrier one step before it does its job.

What actually separates them is whether the barrier can still buy anything. In #1296 the deletion has `272_273` to be pulled back to, outside the tract. In #1394 it is inside the tract at both `263` and `264`: the tract holds it in **both snapshots** and there is no position outside for the clamp to reach. The insertion route now requires exactly that; a swallowed junction keeps the plain test, since a junction member has no bases to release.

## Enabling `an_indel_haplotype_normalizes_to_its_own_sequence`

#1394 was the **seventeenth and last** defect in the chain that property has been finding since #1286, each masked behind the one before it: #1286 → #1287 → #1290 → #1292 → #1296 → #1301 → #1304 → #1297 → #1308 → #1312 → #1316 → #1320 → #1321 → #1323 → #1325 → #1394. Enabling it was #1292's acceptance criterion, then #1316's, then #1325's; none could carry it, because the next failure was always behind the one being fixed.

It now runs on every invocation. `the_ignored_indel_property_still_finds_its_defect` and its helper go with the ignore — the property itself is the guard now — and `rewrite_target_corpus`'s *denotation* row is empty, both targets fixed, so per that file's contract they move out rather than being re-pinned (`assert_denotation_currently_broken` goes with them).

**What the enabling soak is worth, stated precisely.** 6,000,000 clean cases with #1394 fixed: 1,000,000 across the eight seeded shards `ci.yml` runs, plus 5,000,000 in ten unseeded 500k streams, no seed appended. Re-verified after rebasing onto current `main` (which brought #1368 and #1373): CI's eight shards plus 2,000,000 further unseeded cases, still clean.

That is evidence and not proof, and the doc comment says so. Depth never trended toward an end — #1320 at 27,633 successes, #1321 at 114,694, #1323 at 143,034, #1325 at 132,216, #1394 at 231,801 — and a clean *seeded* soak says less than it looks: **#1394 itself was reached by an unseeded stream after all eight seeded shards had passed clean at 1,000,000 cases.** If an eighteenth exists, the unseeded nightly is what will find it, and the response should be to pin its seed and fix it rather than re-ignore the property.

## How #1394 was found, and that it is pre-existing

By the property at 231,801 successes in an unseeded 500k stream, immediately after #1325 was fixed. Verified pre-existing by A/B: with `src/normalize/{merge,mod}.rs` pristine at `origin/main`, the same input yields the identical `g.[262_264A[7];264del]`. #1325's fix neither causes nor changes it.

## Verification

- Suite **9251/9251**; all three oracles (`FERRO_ASSERT_IDEMPOTENT` / `REPARSE` / `IN_BOUNDS`) **9251/9251**
- Conformance axis **201/201** against the prepared reference (`FERRO_MANIFEST`) — not run by PR CI
- `cargo clippy --features dev --all-targets -- -D warnings` clean; `cargo fmt --check` clean
- Guard validity checked by stubbing the new route: both new cases go red, all three #1296 tests stay green

## Relationship to #1235, and to #1392

This closes out criterion 2. It does **not** close #1235, and is deliberately marked `Refs`:

- **Criterion 3** (idempotence) already held.
- **Criterion 1** (confluence) is where the remaining work is: #1392 converges the large spelling class, and two residuals recorded on #1235 remain — `split_buys_no_higher_priority_type`'s two stable normal forms, and non-confluence above `MAX_SPLIT_BLOCK`. PRs for those will stack on this branch, and the tip of that stack is what should carry `Closes #1235`.

One interaction worth flagging for **#1392**: on that branch the #1325 seed fails *differently* — the overlap is gone, but it emits `g.264_265insCAAAA` where the correct payload is `insAAAAC`, i.e. silently wrong bases rather than a detectably malformed pair. So #1325 is not untouched by #1392, and whichever of the two lands second will need its interaction with the other re-measured rather than assumed.

## Rebased, and the pinned seeds re-checked

Rebased onto `main` at `bf4b983a` (bringing #1407, #1411, #1417, #1423, #1425, #1427, #1428 into the base). Re-verified there: suite **9353 passed** under all three oracles, `fmt`/`clippy` clean.

**The check this PR most needed, done explicitly.** It both edits `cis_allele_confluence_proptest.rs` and appends to `tests/proptest-regressions/cis_allele_confluence_proptest.txt`, which is the combination that can silently destroy pinned guards: those entries are RNG seeds replayed *through* the strategy, so changing how a strategy consumes randomness makes every seed regenerate a different case while its comment still names the old one. Confirmed this PR does **not** touch any `fn *_strategy()` — the diff to that file is doc comments and the `#[ignore]` attributes only:

```
git diff origin/main...HEAD -- tests/it/cis_allele_confluence_proptest.rs \
  | grep -E "^[+-]" | grep -E "fn .*strategy|prop_|Strategy"
  (no output)
```

So all fifteen seeds stay valid, which is what makes un-ignoring the property safe rather than merely green.

## CLI-lens findings

**Applied — seed annotation traced to the PR that recorded it.** The whole-span inversion seed was annotated `#1235` (the parent issue) but was added by #1392; `git log -S` on the seed hash confirms it. Now reads `#1392 (refs #1235)`, matching how every other seed names its own issue.

**Declined — restructuring the repeat-growth `TerminalOverrun` to mirror `landing_claimants`.** This asks the new `respell_at_gap` call to select `WriteTransientJunction` when terminal siblings absorb a flush insertion, keeping `RespellAtBoundary` only when no sibling claims the terminal base. That inverts the safe default, in the direction the code explicitly warns against. From `merges_with_a_flush_insertion`'s own doc:

> Nothing here may answer "yes" on a guess: a wrong "yes" emits a position the sequence does not have, and that output re-parses, so `FERRO_ASSERT_REPARSE` does not see it either.

`WriteTransientJunction` is the *dangerous* branch — it is what #1307 was — and it is used at exactly one of the six `respell_at_gap` call sites, the one where a sibling analysis proves it safe. The other five, including this PR's, pass `RespellAtBoundary`. The cost of the conservative choice is documented and bounded: "a repairable collision goes unrepaired" — a missed repair, not a wrong answer. Adopting the ladder here would trade a bounded missed-optimisation for an unbounded out-of-range-coordinate risk on a shape for which the sibling analysis has not been shown to hold.

**Declined — `MSTO_ISSUES` attribution for the #1325/#1394 test pairs.** The finding reads listing them under the `number: 1235` entry as a mis-attribution. That is the table's established pattern, not an anomaly: the #1234 entry lists tests from `issue_1235_cis_allele_confluence.rs`, #1232's lists `soft_masked_reference_yields_the_same_canonical_form` from the same file, and so on — tests are catalogued under the issue whose *behaviour* they pin, not the file they live in. No validator ties a test's filename to its issue number. And the counts are already correct: one entry removed, two added, `PINNED_UNIQUE_TEST_COUNT` 253 → 254.

## Re-verified on top of #1432

The green above was measured against a base ending at #1428. #1432 has since merged and it also rewrites `src/normalize/merge.rs` — it threads `direction` through `canonicalize_from_sequence` and turns `shift_pieces_three_prime` into `shift_pieces` — which is the same file this PR reworks. So the earlier run said nothing about the combination. Rebased onto `c071188b` and re-run there:

| check | result |
|---|---|
| suite, all three oracles | **9356 passed** |
| `fmt` / `clippy` | clean |
| `an_indel_haplotype_normalizes_to_its_own_sequence`, 300,000 cases | **passed**, no seed appended |

That last row is the one worth stating separately. This PR's whole argument for un-ignoring that property is a 6,000,000-case soak — but that soak predates #1432, and #1432 changes the very function the property exercises. 300,000 fresh cases against the combination is a much smaller number than 6,000,000, so it is a smoke test rather than a re-establishment of the original claim; what it rules out is a defect dense enough to matter immediately. Per this file's own note, the reliable signal is whether a seed was appended to `tests/proptest-regressions/`, and none was — the count is still 15.


nh13:
@coderabbitai pause

Issue #1398, opened 2026-08-05
Normalizer: a c.-axis allele normalizes to an insertion across the CDS/3'UTR boundary that is not a fixed point


nh13:
## What's wrong

On the `c.` axis, a two-member allele normalizes to an insertion that spans the CDS/3'UTR boundary, and re-normalizing that output does not return it. Both shuffle directions are affected, and they disagree about the second pass.

```
transcript  TAAATAATAAATATATATTA      c. axis, CDS = transcript 1..=18
                                      so c.18 is the last CDS base, c.*1 the first 3'UTR base

in     NM_TEST.1:c.[11_12dup;16_17insC]
once   NM_TEST.1:c.18_*1insCAT        <- insertion across the CDS/3'UTR boundary
twice  NM_TEST.1:c.18delinsTCAT       (3' direction)
twice  NM_TEST.1:c.16_17insATC        (5' direction)
```

`norm(norm(x)) != norm(x)`, so this fires `FERRO_ASSERT_IDEMPOTENT` as a panic rather than a comparison failure.

A second sequence in the same corpus, `TAATTTATTAATATATATAT`, reaches the same `c.18_*1insCAT` first pass under the 5' direction.

## Reproducing

```rust
let provider = SyntheticBuilder::cds("TAAATAATAAATATATATTA", 1, 18, Strand::Plus).build();
let normalizer = Normalizer::with_config(
    provider,
    NormalizeConfig::default()
        .with_direction(ShuffleDirection::ThreePrime)
        .allow_crossing_boundaries(),
);
// normalize twice; the second pass does not reproduce the first
```

## Why nothing caught it

This is the gap #1283 names as its fifth: every exhaustive cis sweep runs on `g.` only, so the axis-specific clamps — CDS/UTR and transcript-end insertion clamps in particular — were never swept. The genomic sweeps cannot reach it because there is no CDS boundary on `g.`.

It was found by the transcript-axis smoke slice added for #1283, and only at the **full** corpus: the 4-seed default prefix does not draw either triggering sequence (they are corpus indices 26 and 36). So this is also a concrete data point for #1295's trade — the reduced prefix is a strict subset and cannot manufacture a false pass, but it can miss a real defect that the full corpus finds, which is exactly why CI runs the full corpus.

## Scope

[...]

PR #1400, opened 2026-08-05
test(cis): sweep the transcript axes, and find a CDS-boundary defect

### LINKED ISSUE #1283: Tests: the exhaustive sibling-crossing sweep cannot reach five defect classes it is read as covering

nh13:
Refs #1283. **Stacked on #1395** (which is stacked on #1392) — base is `perf/issue-1295-sweep-seed-knob`. Deliberately does not close #1283: the three-member extension is #1268.

## What was missing

Every sweep in `cis_junction_crossing_shift.rs` ran on `g.` only, so the file's claim — no two-member cis allele normalizes to a different sequence — was never tested where the axis itself changes the answer. It does: the CDS/UTR and transcript-end insertion clamps, the exon-junction exception to the 3' rule, the coding one-amino-acid exception to the separation rule and the `c.` codon gate on repeat notation all fire on the transcript axes and nowhere in the existing sweeps.

This is #1283's **fourth** suggested direction, unblocked by #1284 closing — that had to land first, because a transcript-axis member could abort the run outright under `FERRO_ASSERT_REPARSE` rather than fail with a comparison.

Status of the other three, checked against the base rather than assumed:

| direction | state |
|---|---|
| 1. `dup`/`ins` as second member | already in the base |
| 2. non-homopolymer `(CAG)` tract sweep | already in the base |
| 3. three members | **#1268** — not this PR |
| 4. `c.`/`n.` slice | **this PR** |

## It found a real defect — filed as #1398

```
transcript  TAAATAATAAATATATATTA        CDS = transcript 1..=18

in     NM_TEST.1:c.[11_12dup;16_17insC]
once   NM_TEST.1:c.18_*1insCAT          ← insertion across the CDS/3'UTR boundary
twice  NM_TEST.1:c.18delinsTCAT         (3' direction)
twice  NM_TEST.1:c.16_17insATC          (5' direction)
```

`norm(norm(x)) != norm(x)` in both directions, and they disagree about the second pass. Exactly the fifth gap #1283 describes — the genomic sweeps cannot reach it because there is no CDS boundary on `g.`.

**A data point for #1295's trade, too.** The sweep passes at the default 4-seed prefix and fails at the full corpus: the two triggering sequences are corpus indices 26 and 36. The prefix is a strict subset so it cannot manufacture a false pass — but it can miss a real defect the full corpus finds, which is exactly why CI runs the full corpus.

## Why it lands `#[ignore]`d rather than narrowed

The failure is an idempotency **panic**, so the sweep cannot merely count it the way `FIVE_PRIME_DUP_DEL_SEQUENCE_CHANGES` counts its class. The alternative would be trimming the position range until the boundary shape is unreachable — which is precisely the move #1283 exists to complain about, a generator that reads wider than it is.

So it is committed ignored, with the run command in its doc comment, matching how `an_indel_haplotype_normalizes_to_its_own_sequence` carries the defect chain it found. Un-ignore when #1398 lands; it is green on everything else at both seed counts.

```bash
FERRO_SWEEP_SEEDS=full cargo nextest run --features dev \
  -E 'test(no_two_member_transcript_axis_allele)' --run-ignored all
```

## Performance

The provider is built once per (sequence, axis) and cloned per case. `SyntheticBuilder` pads, maps to a genomic contig and reverse-complements on demand, so constructing one inside the innermost loop dominated the test: **39.8s → 12.9s** for the 4-seed prefix, against 18.3s for the far larger genomic sweep beside it.

## Verification

- default profile, all three oracles — **9260/9260 pass**, 147 skipped (the +1 skip is this test)
- the sweep run explicitly at the default prefix — passes, 12.9s
- the sweep at `FERRO_SWEEP_SEEDS=full` — fails on #1398 only, which is why it is ignored

### DIFF
--- a/tests/it/cis_junction_crossing_shift.rs
+++ b/tests/it/cis_junction_crossing_shift.rs
@@ -848,3 +848,202 @@ fn no_tandem_tract_allele_normalizes_to_a_different_sequence() {
         "sequence-changing normalizations in {checked} tract cases: {changed:#?}"
     );
 }
+
+/// A smoke slice of the same class on the **transcript axes** (#1283 gap 4).
+///
+/// Every sweep above runs on `g.` only, so the file's claim — that no two-member
+/// cis allele normalizes to a different sequence — was never tested where the
+/// axis itself changes the answer. It does change it: the CDS/UTR and
+/// transcript-end insertion clamps, the exon-junction exception to the 3' rule,
+/// the coding one-amino-acid exception to the separation rule and the `c.` codon
+/// gate on repeat notation all fire here and nowhere above. #1284, which had to
+/// land first because a transcript-axis member could abort the run outright under
+/// `FERRO_ASSERT_REPARSE`, is closed.
+///
+/// Deliberately a **smoke slice**, not a second exhaustive sweep. It reuses the
+/// same corpus, member shapes and oracle, so what it adds is the axis rather than
+/// new shapes — and the genomic sweeps above already carry the shape coverage.
+/// Sizing it that way is also what keeps it affordable: it multiplies the case
+/// count by the number of axes, and the seed knob (#1295) is what makes that a
+/// prefix locally and the full corpus in CI.
+///
+/// `cds_start = 1` so `c.N` is transcript position `N` and the coordinate
+/// arithmetic matches the genomic sweeps exactly. The CDS is 18 of the 20 bases —
+/// six whole codons, so the codon gate is not tripped by a ragged reading frame —
+/// leaving `c.*1_*2` as a 3'UTR the members deliberately stay out of, since UTR
+/// boundary behaviour has its own dedicated tests and would only add noise to a
+/// smoke slice.
+///
+/// # `#[ignore]`d pending #1398, which this sweep found
+///
+/// It passes at the default 4-seed prefix and **fails at the full corpus**, on a
+/// real defect rather than on anything about the sweep: `c.[11_12dup;16_17insC]`
+/// over `TAAATAATAAATATATATTA` normalizes to `c.18_*1insCAT`, an insertion across
+/// the CDS/3'UTR boundary, and re-normalizing that does not return it
+/// (`c.18delinsTCAT` under 3', `c.16_17insATC` under 5'). That trips
+/// `FERRO_ASSERT_IDEMPOTENT` as a panic, so the sweep cannot merely count it the
+/// way `FIVE_PRIME_DUP_DEL_SEQUENCE_CHANGES` counts its class.
+///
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

Issue #422, opened 2026-05-22
normalize: expand ins[ACC:g.A_B] cross-reference to flat literal (#395 item 2 follow-up)


nh13:
## Summary

`InsertedSequence::Reference` (bare cross-reference, e.g. `ins[NC_000022.10:g.42536337_42536382]`) and `InsertedPart::ExternalRef` (cross-reference *inside* a Complex bracket, e.g. `ins[A;NC_000022.11:g.100_200]`) are valid per the HGVS spec but ferro returns `FerroError::UnsupportedVariant` ("`...cross-reference is valid HGVS but not yet supported by ferro; see follow-up`") instead of expanding to a flat literal sequence.

Spec basis: [`assets/hgvs-nomenclature/docs/recommendations/DNA/delins.md`](https://hgvs-nomenclature.org/stable/recommendations/DNA/delins/) — examples include `NC_000022.10:g.42522624_42522669delins[NC_000022.10:g.42536337_42536382]` (same-accession bracketed range conversion in `_CYP2D6_`) and `NC_000002.12:g.pter_8247756delins[NC_000011.10:g.pter_15825266]` (cross-chromosome translocation).

Originally tracked as #395 item 2; deferred from PR #421 (the issue-395 omnibus) because it needs a multi-accession provider lookup path that's substantial enough to warrant focused review.

## Stub locations

Three call sites in `src/normalize/rules.rs`:

- `:1813-1822` — `append_part_bases` arm for `InsertedPart::ExternalRef(reference)`. Bracket-inside-Complex case.
- `:1841-1848` — `detect_deferred_part` arm. Pre-scan so cross-reference shapes error even when an earlier part in the bracket is non-flatten (e.g. `Repeat`).
- `:1919-1924` — `expand_inserted_sequence` arm for `InsertedSequence::Reference(reference)`. Bare cross-reference case (no surrounding `Complex`).

All three return `FerroError::UnsupportedVariant` with `"cross-reference is valid HGVS but not yet supported by ferro; see follow-up"` so they're grep-friendly.

## Scope

1. Parse the inner reference string (`NC_xxx:g.A_B`, `LRG_xxx:g.A_B`, `NM_xxx:c.A_B`, etc.) into `(accession, coord_kind, start_1based, end_1based)`.
2. Fetch the cited bases via the provider:
   - For `g.`/`m.`/`o.` references: `ReferenceProvider::get_genomic_sequence(accession, ...)`.
[...]

PR #1401, opened 2026-08-05
fix(normalize): waive the weight bound for a policy collapse

### LINKED ISSUE #1235: Normalizer: cis-allele members normalized independently, not from the resulting sequence (root cause of #1229–#1234)

nh13:
Refs #1235 — the first of the two confluence residuals recorded on that issue ([comment](https://github.com/fulcrumgenomics/ferro-hgvs/issues/1235#issuecomment-5185660755)).

**Stacked on #1399** (criterion 2). Review that one first; this PR's diff against it is a single commit.

## The defect

`partition_block` deliberately returns the whole block when every gap between pieces is a single unchanged base and the split buys no higher-priority type (`every_separation_is_a_single_base` + `split_buys_no_higher_priority_type`). The reasoning is `delins.md:44-47`: a lone matched base inside a length-changing block is more likely coincidence than structure, so the spanning `delins` is the better description.

That policy was unreachable from half its own inputs:

```
reference  ("ACGT" x 64) + "ACTAC" + ("ACGT" x 64)

g.257_261delinsGGGTGG                 stable
g.[257_258delinsGGG;260_261delinsGG]  stable, and denotes the same GGGTGG
```

Both fixed points. One variant, two stable normal forms.

## Why it could not be reached

A spanning form is necessarily **heavier** than the split it replaces — merging across an unchanged base always is — and `canonicalize_from_sequence` refuses any derivation heavier than the input (`changed_columns_of_pieces > changed_columns_of_edits`). So whenever the input arrived already split, ferro's own preferred spelling was refused.

This is a direct instance of the characterisation recorded on #1235: *the derivation is confluent exactly on haplotypes whose derived partition is weight-minimal, because the bound compares derived weight against the input's weight.*

## The fix, and why it is narrow

The bound guards against derivations that are *accidentally* worse — the single-gap aligner parking its lone gap at one end so the offset between two changes reads as a run of substitutions. A rule that widens **on purpose** is not that.

`partition_block_detailed` now reports a `BlockCollapse` (`No` / `ByPolicy` / `Fallback`), and the weight bound is waived only for `ByPolicy`. Every other collapse — oversized block, no alignment, empty pieces, `separations_are_meaningful` declining — stays `Fallback` and keeps the bound.

## This closes a limitation #422 pinned as unclosable

`issue_422_cross_reference_ins` carried this same shape as a deliberate known non-confluence, and its comment said it could not be closed from there:

> Letting a coincidence-driven collapse override that veto does close this case, but it also merges `g.[306dup;308C>A]` into `g.307_308delinsCGA`, breaking #999 — and the collapse cannot tell the two apart, because both are two-member inputs whose derived pieces collapse to one.

Recording *why* the block collapsed is what tells them apart. **#999's split buys a `dup`** — a higher-priority type — so `split_buys_no_higher_priority_type` is false, nothing is recorded as `ByPolicy`, and the bound still refuses. `g.[306dup;308C>A]` stays split.

That pin set an explicit condition for changing it — "change this assertion only alongside a fix that keeps #999 green" — and it is met. All four #999 tests were run directly and confirmed green, not inferred from a green suite.

## Verification

- Suite **9253/9253**; all three oracles **9253/9253**
- Conformance axis **201/201** against the prepared reference (not run by PR CI)
- CI's eight soak shards (1,000,000 cases per property) clean, no seed appended
- clippy `-D warnings` and `cargo fmt --check` clean

The only test whose expectation moved is #422's pinned limitation, and it moved from "two stable strings" to convergence.

## Remaining for #1235

One residual left — non-confluence above `MAX_SPLIT_BLOCK` (L>=1026), where `partition_block` short-circuits to `whole()` on length alone. A PR for it will stack on this branch, and that tip is what should carry `Closes #1235` — noting #1392 also has to land for criterion 1.

## Representation stability — measured, and this one is a migration

Rebased onto `main` at `1ee1f9a2` (#1399 squash-merged, so its two commits are dropped rather than replayed; this PR is now a single commit). Re-verified there: suite **9358 passed** under all three oracles, `fmt`/`clippy` clean, and the 20 `issue_999` + `issue_422` tests pass directly.

**The PR body had no stability section, and it needed one.** Measured by dumping normalized output on `1ee1f9a2` and on this branch over a corpus built for the shape this PR changes — 48 sequences (24 seeds x 2 alphabets), each enumerated over every two-member position, in three spellings: the split form, the spanning form of the same variant, and #999's `dup`+`sub` control.

| | rows |
|---|---|
| total | 2,160 |
| **moved** | **498 (23.1%)** |
| of which: split spelling (the intended target) | 498 |
| of which: spanning spelling | **0** |
| of which: #999 `dup`+`sub` control | **0** |

**Read the 23% correctly: this corpus is deliberately enriched for the affected shape, so that is the rate *within the shape family*, not a repo-wide rate.** The other PRs in this family quoted 0.2–0.8% because they measured over broader corpora. What is comparable across them is the direction and the class, below.

**Class: expensive.** Of the 498 moved rows, **457 were fixed points on `main`** — previously-accepted, stable output. A consumer storing `g.[257_258delinsGGG;260_261delinsGG]` today gets `g.257_261delinsGGGTGG` after this lands. The remaining 41 were already unstable on `main` (their first-pass answer was not a fixed point), so those are free.

**Direction: toward the spec's stated form, away from the shipped one.** `delins.md:44-47` is what the collapse policy cites, and this PR makes that policy reachable from the split spelling. So it is the case `CLAUDE.md` describes as "where the spec mandates the other form, change it, and flag it loudly as a migration" — hence this section rather than a footnote.

**What it buys:** confluence on this shape goes from **28.9% to 98.1%** of split/spanning pairs (208/720 → 706/720). That is the point of the PR, and it is a large gain — but it is bought with a real migration, not for free, and the two should be weighed together rather than the second being assumed.

**Scoping confirmed by measurement, not by argument.** Zero of the #999 `dup`+`sub` controls moved, across all 720 of them — which is the empirical version of the `split_buys_no_higher_priority_type` claim in the description above.

Caveats: `main`-to-branch rather than release-to-release (the harness postdates v0.12.0), and this is a synthetic 20-mer corpus over one shape family — an order of magnitude for that family, not a bound on the library.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2146,7 +2146,8 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
         return None; // no net change; leave it to the existing pipeline.
     }
 
-    let mut pieces = partition_block(&ref_bytes[lo..hi_ref], &result[lo..hi_alt]);
+    let (mut pieces, collapse) =
+        partition_block_detailed(&ref_bytes[lo..hi_ref], &result[lo..hi_alt]);
     // Shadow the sequence-first splitter on the very same trimmed block, before
     // any 3'-shift or coalescing, so a reported disagreement is a disagreement
     // about the *split* and not about a downstream pass. Never affects `pieces`.
@@ -2261,7 +2262,21 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
     // Strict on purpose: #1229, #1231, #1233 and #1234 all sit exactly at
     // equality, and equality is where prioritisation (`general.md:56`) may
     // legitimately prefer the re-derived shape.
-    if changed_columns_of_pieces(&pieces) > changed_columns_of_edits(&edits) {
+    //
+    // One exemption, and only one: a block `partition_block` collapsed *by
+    // policy* (`BlockCollapse::ByPolicy`). There the aligner did find a
+    // multi-piece partition and `split_buys_no_higher_priority_type` merged it
+    // deliberately, because the split bought nothing but more `delins` members
+    // across a single unchanged base. That collapse is necessarily heavier than
+    // the split it replaced — merging across an unchanged base always is — so
+    // the bound refused ferro's own preferred spelling whenever the input
+    // happened to arrive already split, and `g.[257_258delinsGGG;260_261delinsGG]`
+    // and `g.257_261delinsGGGTGG` were both stable: one variant, two normal
+    // forms. The bound is a guard against derivations that are *accidentally*
+    // worse, and a policy that widens on purpose is not that.
+    if collapse != BlockCollapse::ByPolicy
+        && changed_columns_of_pieces(&pieces) > changed_columns_of_edits(&edits)
+    {
         return None;
     }
 
@@ -3464,6 +3479,26 @@ fn trim_common_flanks(reference: &[u8], result: &[u8]) -> (usize, usize, usize) 
     (lo, hi_ref, hi_alt)
 }
 
+/// Why [`partition_block_detailed`] returned a single whole-block piece.
+///
+/// The distinction matters to the weight bound in `canonicalize_from_sequence`,
+/// which refuses a derivation heavier than the input. That refusal is right for
+/// [`BlockCollapse::Fallback`] — no usable partition was found, so the lone piece
+/// is a worse description that the per-member pipeline should answer instead. It
+/// is wrong for [`BlockCollapse::ByPolicy`], where the aligner *did* find a
+/// multi-piece partition and a deliberate rule merged it: refusing there makes
+/// ferro's own preferred spelling unreachable from an input that is already
+/// split, which is how one variant ended up with two stable normal forms.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+enum BlockCollapse {
+    /// Not collapsed — the pieces are the alignment's own partition.
+    No,
+    /// Collapsed deliberately by [`split_buys_no_higher_priority_type`].
+    ByPolicy,
+    /// Collapsed because no usable partition was available.
+    Fallback,
+}
+
 /// Partition a changed block into maximal runs of change separated by at least
 /// one unchanged base.
 ///
@@ -3475,7 +3510,21 @@ fn trim_common_flanks(reference: &[u8], result: &[u8]) -> (usize, usize, usize) 
 /// pieces — and break ties by placing the indel 5'-most. The tie-break is an
 /// implementer's choice; the spec does not reach it. Each piece is 3'-shifted
 /// afterwards, so the 3' rule is still honoured per member.
+///
+/// Test-only since the weight bound began reading the collapse origin: every
+/// production caller needs [`partition_block_detailed`]'s second element, and
+/// the tests here care only about the pieces.
+#[cfg(test)]
 fn partition_block(reference: &[u8], result: &[u8]) -> Vec<Piece> {
+    partition_block_detailed(reference, result).0
+}
+
+/// Why a block came back as one whole-block piece — see [`BlockCollapse`].
+///
+/// Split out from [`partition_block`] because the caller cannot otherwise tell a
+/// *deliberate* collapse from a fallback, and the weight bound needs to
+/// (`BlockCollapse::ByPolicy`).
+fn partition_block_detailed(reference: &[u8], result: &[u8]) -> (Vec<Piece>, BlockCollapse) {
     let whole = || {
         vec![Piece {
             ref_start: 0,
@@ -3486,14 +3535,14 @@ fn partition_block(reference: &[u8], result: &[u8]) -> Vec<Piece> {
     // One bound for every regime now that every length-changing regime has a
     // real guard — see `separations_are_meaningful` (#1271).
     if reference.len() > MAX_SPLIT_BLOCK || result.len() > MAX_SPLIT_BLOCK {
-        return whole();
+        return (whole(), BlockCollapse::Fallback);
     }
     let Some(columns) = best_alignment(reference, result) else {
-        return whole();
+        return (whole(), BlockCollapse::Fallback);
     };
     let pieces = pieces_from_columns(&columns, reference, result);
     if pieces.is_empty() {
-        return whole();
+        return (whole(), BlockCollapse::Fallback);
     }
     // Every length-changing block, not just net insertions (#1271). Equal-length
     // blocks are exempt because `best_alignment` compares position-wise there —
@@ -3501,7 +3550,7 @@ fn partition_block(reference: &[u8], result: &[u8]) -> Vec<Piece> {
     if reference.len() != result.len()
         && !separations_are_meaningful(&pieces, result.len().abs_diff(reference.len()))
     {
-        return whole();
+        return (whole(), BlockCollapse::Fallback);
     }
     // Scoped to length-changing blocks: an equal-length block has no gap to
     // place, so every matched base is a genuine coordinate-wise identity rather
@@ -3512,9 +3561,9 @@ fn partition_block(reference: &[u8], result: &[u8]) -> Vec<Piece> {
         && every_separation_is_a_single_base(&pieces)
         && split_buys_no_higher_priority_type(&pieces, reference)
     {
-        return whole();
+        return (whole(), BlockCollapse::ByPolicy);
     }
-    pieces
+    (pieces, BlockCollapse::No)
 }
 
 /// Merge a run of pieces that together span one inversion back into that single
--- a/tests/it/issue_422_cross_reference_ins.rs
+++ b/tests/it/issue_422_cross_reference_ins.rs
@@ -439,28 +439,35 @@ fn cross_reference_inv_suffix_expands_inside_a_complex_bracket() {
         "the spanning delins must be a normalization fixed point",
     );
 
-    // KNOWN LIMITATION, pinned deliberately so it is visible rather than merely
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
## Holding this for a rework — the waiver is scoped too widely

Measured against `1ee1f9a2` over a corpus enriched for the shape this PR changes (2,160 rows): **498 move, and 457 of those were fixed points on `main`** — previously-accepted, stable output. All of them are genomic (`g.`), and they move from the split spelling onto the spanning `delins`.

That is the direction #1232 and #1421 argue against, and #79 already decided it explicitly. #79's table:

| Input | Spec form |
|---|---|
| `c.[145C>T;147C>G]` (145–147 share a codon) | `c.145_147delinsTGG` |
| `c.[3C>T;5G>A]` (straddle codon 1/2) | unchanged |
| `c.[145C>T;148C>G]` (gap of 2 nt) | unchanged |
| **`g.[145G>A;147C>T]` (no codon frame in `g.`)** | **unchanged** |

The last row is this PR's shape.

On the spec: `delins.md:17` says two variants separated by **one or more** nucleotides are described individually — "one or more" includes one — and the FAQ at `:80` restates it. The exception at `:18` requires the two changes to **together affect one amino acid**, and its NOTE is explicitly about protein-consequence prediction. On a `g.` description there is no reading frame in which to establish that, so the exception does not reach these rows. This PR cites `delins.md:44-47`, which is the 52-base `c.850_901delinsTTCCTCGATGCCTG` example whose alternative split is an artefact of the insert coincidentally aligning in several places — a real recommendation about a different situation. `partition_block`'s own doc already cites the other side (`merge.rs:3509`: "the separation rule (`delins.md:17`) wants the pieces").

## What survives, and what changes

**`BlockCollapse` is the right idea and should stay.** Recording *why* a block collapsed is what answers the old "the collapse cannot tell the two apart" objection, and it holds up empirically: **0 of 720** `dup`+`sub` controls moved, so `split_buys_no_higher_priority_type` does protect #999.

The waiver just needs a second condition — that the codon-frame exception actually applies: coding axis, separation of exactly one nucleotide, and both changes inside one codon (`(a-1)/3 == (c-1)/3`, no offset, inside the CDS proper — #79's own formulation). On `CodonGate::NotApplicable` (`g.`, `m.`/`o.`, `n.`, intronic), do not waive.

The obstacle is that `partition_block` works in sequence space and knows nothing about the reading frame — the genomic caller passes `CodonGate::NotApplicable` outright (`mod.rs:3624`). #1432 just threaded `ShuffleDirection` into `canonicalize_from_sequence` for the same reason, so that is the template.

Two consequences worth stating up front: the `issue_422_cross_reference_ins` assertion has to go back to its pinned form (that case is `NC_000022.10:g.`, so it stays split, and this PR's "closes a limitation #422 pinned as unclosable" claim does not survive the narrowing); and the residual stays open — on `g.` the spanning form and its split are still two fixed points, but the right fix there is the *opposite* direction, making the spanning form split, which is what #1421 asks for. That belongs with the #1419/#1420/#1421 family rather than here, so `Refs #1235` should stay a `Refs`.

Expected movement after the narrowing is ~zero on `g.`, which is the point: it keeps the confluence gain where the spec grants it and costs nothing downstream.

Branch state: rebased onto `1ee1f9a2` as a single commit, 9358/9358 green under all three oracles, `fmt`/`clippy` clean. #1399 squash-merged, so the rebase used `git rebase --onto origin/main 8b972683` to drop its commits rather than replay them.

#1403 is stacked on this and is blocked until it settles.

nh13:
## Correction to my previous comment: narrowing this to the codon-frame exception would empty it

I suggested above that the waiver be narrowed to the case `delins.md:18` actually licenses — coding axis, one-nucleotide separation, both changes in one codon. Having done the arithmetic, that narrowing fires on **nothing**, so it is not a rework worth building.

The exception requires the two variants **together affecting one amino acid**. One amino acid is one codon — three nucleotides — and the two variants *plus* the single base separating them all have to fit inside those three positions. That forces each variant to occupy exactly one base. A single-base `delins` already reduces to a substitution (#73), so the exception reaches exactly one shape: two single-base substitutions separated by one base, in one codon. That is `c.[145C>T;147C>G]` — **#79, closed and implemented**.

This PR's shape is two **2-base** `delins` separated by one base: five consecutive positions. Five is greater than three, so it cannot satisfy "together affecting one amino acid" on any axis, in any codon. The exception does not reach it at all — not on `g.`, and not on `c.` either.

## Measured confirmation

Re-measured with the committed harness from #1442 (24 seeds, 16,128 rows, base `1ee1f9a2`):

| | rows |
|---|---|
| total | 16,128 |
| **moved** | **1,840 (11.4%)** |
| previously **accepted** (a migration) | 1,688 |
| previously not a fixed point (free) | 152 |

| axis | direction | rows | moved | expensive |
|---|---|---|---|---|
| `c` | 3prime | 4,032 | 460 | 422 |
| `c` | 5prime | 4,032 | 460 | 422 |
| `g` | 3prime | 4,032 | 460 | 422 |
| `g` | 5prime | 4,032 | 460 | 422 |

Perfectly symmetric. That is the fingerprint of `partition_block` being **axis-blind** — it fires identically on `c.` and `g.` because it works in sequence space and never learns the reading frame. So this is not a coding rule leaking onto `g.`; it is a rule with no axis concept at all, applied to a shape the exception never covered on either axis.

(The expensive share is 1,688/1,840 = **91.7%**, against 457/498 = 91.8% from the earlier hand measurement over 7.5× fewer rows. The `dup_plus_sub` control held at **0 of 2,688**, so `split_buys_no_higher_priority_type` is empirically sound — that part of the design is fine.)

## Suggested disposition

Close this without merging. `BlockCollapse` — recording *why* a block collapsed — is a genuinely good idea and worth salvaging if a later PR needs it; the waiver is what has no spec support.

The residual is real and stays open: on `g.`, `g.257_261delinsGGGTGG` and its split spelling are two stable fixed points for one variant. It just is not fixable from this direction. Its resolution is to make the **spanning** form split, which is what #1421 asks for and what `delins.md:17` plus the FAQ at `:80` require, and it belongs with the #1419/#1420/#1421 family.

#1403 is stacked here, so it needs rebasing onto `main` directly rather than onto this.

nh13:
Closing this without merging. The mechanism is right and the scope is wrong, and once the scope is corrected there is nothing left for the PR to do.

## Why the correctly-scoped version fires on nothing

The waiver leans on the `delins.md:18` exception: "two variants separated by one nucleotide, **together affecting one amino acid**, should be described as a 'delins'."

- One amino acid is one codon: 3 nucleotides.
- Two variants *plus* the single base separating them must all fit inside those 3 positions. That forces each variant to occupy exactly one base.
- A single-base `delins` already reduces to a substitution (#73).

So the exception reaches exactly one shape — two single-base substitutions separated by one base within one codon, i.e. `c.[145C>T;147C>G]` — which is #79, closed and implemented. This PR's motivating shape is two 2-base `delins` separated by one base: five consecutive positions. Five is greater than three, so it cannot satisfy "together affecting one amino acid" in any codon, on any axis. A codon-scoped narrowing would move zero rows and duplicate #79.

#79's own table already decided this PR's shape explicitly: `g.[145G>A;147C>T]` (no codon frame in `g.`) -> **unchanged**.

## What the unscoped version costs

Measured with the #1442 dump/diff harness at 24 seeds (16,128 rows):

| | rows |
|---|---|
| moved | 1,840 (11.4%) |
| previously accepted (a migration) | 1,688 |
| previously not a fixed point (free) | 152 |

Movement is perfectly symmetric across all four (axis, direction) cells — 460 moved / 422 expensive in each of `c.`/`g.` x 3prime/5prime. That symmetry is the fingerprint of `partition_block` being axis-blind: it fires identically on `c.` and `g.` because it works in sequence space and never learns the reading frame. The waiver is not a coding rule leaking onto `g.`; it is a rule with no axis concept at all, applied to a shape the exception never covered on either axis.

The general rule in `delins.md:17` ("two variants separated by **one or more** nucleotides should be described individually and **not** as a 'delins'", restated in the FAQ at `:80`) covers this shape, and `partition_block`'s own doc comment already cites it. The citation this PR used instead, `delins.md:44-47`, is the 52-base `LRG_199t1:c.850_901delinsTTCCTCGATGCCTG` example, whose alternative split exists only because the inserted sequence coincidentally aligns with the reference in several places — a different situation.

## Where the residual goes

The residual is real: on `g.`, `g.257_261delinsGGGTGG` and its split spelling remain two stable fixed points for one variant. But it is not fixable from this direction. The resolution is to make the **spanning** form split, which is what #1421 asks for and what `delins.md:17` and the FAQ at `:80` require. Recording it there rather than here. #1232 (closed) states the same canonical preference — the split — and lists the consumer-facing severity as over-splitting's mirror image.

## Worth salvaging

`BlockCollapse` (`No` / `ByPolicy` / `Fallback`) plus `partition_block_detailed` is a genuine contribution: it lets a deliberate policy collapse be told apart from a fallback, which resolves the "the collapse cannot tell the two apart" objection recorded in `issue_422_cross_reference_ins.rs`. Empirically sound too — 0 of 2,688 `dup`+`sub` controls (#999) moved. Any later PR that needs that distinction should lift it from `a733bb7d` rather than rebuild it.

#1403 will be retargeted to `main` and reviewed on its own; its change to the `MAX_SPLIT_BLOCK` short-circuit does not depend on the waiver.

PR #1402, opened 2026-08-05
test(cis): sweep three-member alleles, with the causes counted apart

### LINKED ISSUE #1268: Tests: no committed sweep covers three-member cis alleles, and the one measurement taken was never decomposed

nh13:
Closes #1268. **Stacked on #1400** — base is `test/issue-1283-widen-sibling-sweep`, the head of the `#1392 → #1395 → #1400` chain, because it edits the same two files.

## What was missing

Both committed sweeps generate exactly **two** members, so the ask CodeRabbit made on #1257 — *"extend the randomized overlap/convergence coverage to alleles with three or more members"* — was reported as addressed while the artifact never covered it. Not a marginal widening: the measurement #1268 records found 0 overlapping outputs at two members and tens of thousands at three.

## Separating the causes is the actual ask

That measurement's other half was a 61,765-case `unapplicable` bucket conflating genuine overlap, `hgvs_to_spdi` conversion failure and parse rejection. #1268 asks that no three-member number be quoted until they are separated.

`apply_with` therefore grows a sibling, `apply_reason`, returning a typed `ApplyFailure` rather than `None`:

| variant | meaning |
|---|---|
| `Overlapping` | a member claims a base another already claimed |
| `CoincidentInsertions` | two pure insertions at one interbase — order, and so the denoted sequence, undefined |
| `OutOfBounds` | a span runs past the end of the reference |
| `Unconvertible` | `hgvs_to_spdi` declined a member |
| `Unparseable` | `parse_hgvs` rejected it |
| `StatedBasesMismatch` | stated deleted bases disagree with the reference |

The two share **one** walk rather than duplicating it, for the reason `apply_with`'s own note gives about `claimed_from`: a second copy is a second thing to drift.

Splitting also fixed a conflation inside it. `end > reference.len() || end > claimed_from` reported *a coordinate running off the reference* and *a member overlapping its neighbour* as the same failure; they are now distinct.

## What it measures today

**Every bucket is empty** over this corpus — 0 overlapping, 0 coincident-insertion, 0 sequence-changing outputs, and 0 inputs declined for any cause, across 38,016 cases at the default prefix.

That is the re-measurement #1268 asks for before any three-member number is quoted again, and it is a long way from the throwaway probe's 10,881 sequence-changing cases. The confluence work in between is what closed it; this sweep is what keeps it closed.

Because every bucket is zero, **there is nothing to pin** — #1268's request for named exclusions cross-referenced to issues has no subject. The `assert!(…is_empty())` form reports offending descriptions directly, which is strictly better than the count it would otherwise be.

## Sizing

Three nested member loops over the same positions would be roughly the two-member sweep cubed. The first and third members **stride** their range by two rather than walking it — striding keeps both ends reachable, where truncating would silently drop the cases nearest the sequence ends, which is where the shifts this file is about actually terminate.

| | cases | time |
|---|---:|---:|
| walking, 4-seed default | 131,328 | 60.8s |
| **striding, 4-seed default** | **38,016** | **17.2s** |
| the two-member sweep beside it | — | 18.3s |

Its full count is **16**, not the 48 its neighbours use: each seed carries ~3× their cases, so 16 is comparable *total* coverage — and at 48 this one test cost **411s** in CI's sweeps job against 183s for all three existing sweeps together. 52s at 16. A deliberate starting point for a sweep whose buckets are all zero, not a ceiling, and the comment says so.

## Verification

- default profile, all three oracles — **9261/9261 pass**, 147 skipped
- `FERRO_SWEEP_SEEDS=full` on this sweep — passes, 52s

### DIFF
--- a/tests/it/cis_junction_crossing_shift.rs
+++ b/tests/it/cis_junction_crossing_shift.rs
@@ -1047,3 +1047,212 @@ fn no_two_member_transcript_axis_allele_normalizes_to_a_different_sequence() {
         "sequence-changing normalizations in {checked} transcript-axis cases: {changed:#?}"
     );
 }
+
+/// Three-member cis alleles, with the failure causes separated (#1268).
+///
+/// Both committed sweeps above generate exactly **two** members, so the ask
+/// CodeRabbit made on #1257 — "extend the randomized overlap/convergence coverage
+/// to alleles with three or more members" — was reported as addressed while the
+/// artifact never covered it. Three members is not a marginal widening: the
+/// measurement #1268 records found 0 overlapping outputs at two members and tens
+/// of thousands at three.
+///
+/// # The causes are counted apart, which is the point
+///
+/// That measurement's other half was a 61,765-case "unapplicable" bucket that
+/// conflated genuine overlap, `hgvs_to_spdi` conversion failure and parse
+/// rejection — three quite different severities. #1268 asks that no three-member
+/// number be quoted until they are separated, so this sweep reports
+/// [`ApplyFailure`] per cause rather than a single count, and asserts on the two
+/// that are defects in a normalizer's *output*:
+///
+/// * `Overlapping` / `CoincidentInsertions` — the output denotes no single
+///   sequence. Both are failures here.
+/// * `Unconvertible` / `Unparseable` / `OutOfBounds` / `StatedBasesMismatch` on
+///   an **input** — a case this sweep cannot speak for, counted and bounded as a
+///   share rather than asserted at zero.
+///
+/// `FERRO_ASSERT_REPARSE` (#1259) already covers parse rejection of normalizer
+/// output suite-wide, so this does not re-derive it.
+///
+/// # Sizing
+///
+/// Three nested member loops over the same positions would be roughly the
+/// two-member sweep cubed. The first and third members therefore **stride** their
+/// position range by two rather than walking it: striding keeps both ends of the
+/// range reachable, where truncating the range would silently drop the cases
+/// nearest the sequence ends — which is where the shifts this file is about
+/// actually terminate. Measured, at the 4-seed default: 131,328 cases and 60.8s
+/// walking, 38,016 and 17.2s striding, against 18.3s for the two-member sweep
+/// beside it. The seed knob (#1295) does the rest.
+///
+/// # What it measures today
+///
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

Issue #1230, opened 2026-07-27
Normalizer: inv over-recognized for a span with unchanged interior (should be separate substitutions); non-confluent


msto:
> [!NOTE]
> Authorship: the content below was drafted by GitHub Copilot CLI (an AI coding
> agent) and filed via gh under @msto's account. It reflects the agent's
> analysis, not a statement authored by @msto.

## What's wrong

`Normalizer.normalize` canonicalizes toward `inv` for a span whose replacement is
the reverse complement of the reference — **even when the span's interior bases are
unchanged**, so the actual changes are two *non-contiguous* substitutions. The
equivalent two-substitution encoding is left as a separate stable form, so the same
variant has two normalized representations.

Example (ref `GATG` at positions 20–23; `inv` → `CATC`, changing only the two ends;
the interior `AT` is a complementary pair and is unchanged):

```
TEMPLATE:g.20_23inv         -> TEMPLATE:g.20_23inv          # ferro's attractor
TEMPLATE:g.20_23delinsCATC  -> TEMPLATE:g.20_23inv          # also -> inv
TEMPLATE:g.[20G>C;23G>C]    -> TEMPLATE:g.[20G>C;23G>C]     # left as-is (not converged)
```

All three describe the identical resulting sequence; the substitution-pair and the
`inv` form are each idempotent — **two stable fixed points for one variant**
(non-confluence). No error or warning is raised.

### Why the substitution-pair is the correct canonical

The two changed positions (20 and 23) are **separated by two unchanged nucleotides**
(21, 22). Two spec rules make the individual-substitution form canonical, and the
inversion form non-canonical:

- **"separated → individual"** (`inversion.md:21`, restated `delins.md:81`):
  > two variants separated by one or more nucleotides should be described
  > individually and **not** as a "delins".

  The changes here are separated, so they are two substitutions, not one spanning
  event.
- **prioritisation** (`general.md:56`): the preferred description order is
  **(1) substitution**, (2) deletion, **(3) inversion**, (4) duplication,
  (5) insertion. When the change can be written as substitutions, substitution is
  preferred over inversion.
- **precedent** (`inversion.md:16`): a degenerate (1-nucleotide) "inversion" is not
  allowed and must be a substitution. An "inversion" whose only changes are at the
  ends (unchanged interior) is the same situation one step up: it is really the end
  substitutions.

Every worked inversion example in `inversion.md` changes **all** bases in the span
[...]

PR #1403, opened 2026-08-05
fix(normalize): cap the split only where there is a gap to place

nh13:
**Refs #1235.** One of the two confluence residuals recorded on that issue ([comment](https://github.com/fulcrumgenomics/ferro-hgvs/issues/1235#issuecomment-5185660755)).

> **No longer stacked.** This was the tip of a three-PR stack (#1399 <- #1401 <- this). #1399 has merged, and **#1401 was closed without merging** — its waiver is not spec-supported at any scope (`delins.md:18`'s exception reaches only two single-base substitutions inside one codon, which is closed-and-implemented #79). So this PR now sits directly on `main`, and nothing here depends on #1401's `BlockCollapse`.
>
> **The closing keyword for the root cause stays demoted to `Refs`.** This fixes **0 of the 9** rows reported in #1419/#1420/#1421: those need a *split* heavier than the input's own spelling, which is the opposite direction. Auto-closing the root cause while that class is open is the pattern that produced #1419-#1421 out of #1231-#1233.

## The defect

`partition_block` short-circuited to the whole block on length alone, in every regime:

```rust
if reference.len() > MAX_SPLIT_BLOCK || result.len() > MAX_SPLIT_BLOCK { whole() }
```

An **equal-length** block has no gap to place, so `best_alignment` returns the position-wise pairing on its first statement:

```rust
if reference.len() == result.len() {
    return Some((0..reference.len()).map(|i| (Some(i), Some(i))).collect());
}
```

There is no placement search there for the cap to bound. Capping it anyway cost confluence: the un-partitioned whole block is refused by the weight bound whenever the input was spelled as its individual changes, so on a 1100 nt near-palindrome both spellings were stable fixed points:

```
g.257_1356inv                       and   g.[257C>A;267A>C;1346G>T;1356T>G]
```

One variant, two normal forms — with an **exact** boundary at the cap (1024 confluent, 1026 not), which is the signature of a length short-circuit rather than anything about the sequence.

## What the cap actually bounds — corrected

An earlier revision of this PR (and of `MAX_SPLIT_BLOCK`'s own doc comment, which this PR now fixes) justified the cap as bounding "`best_alignment`'s gap-placement sweep, which is O(n²)". **That is not true, and has not been since the scoring was made separable.** `best_alignment`'s own doc says so: a placement scores in O(1), so the search is linear — "at `MAX_SPLIT_BLOCK` that is ~1024 comparisons against ~1e6". The one quadratic step is ranking a *tie* by what it separates, and it carries its own and tighter bound in `MAX_TIE_BREAK_SWEEP` (256), which binds first at every length `MAX_SPLIT_BLOCK` reaches.

So the scoping stands on the simpler ground — an equal-length block has no search at all — and `MAX_SPLIT_BLOCK`'s doc is updated to say what it really is: a bound on **length-changing** blocks that is not a quadratic guard.

## Which form is canonical

The four substitutions, and not merely because they are lighter: an inversion whose interior is unchanged is the #1230 shape from #1235's own table ("inv over-recognized (unchanged interior) -> separate subs"). A 1100 nt inversion that alters four bases is not an inversion.

## Representation movement — measured

Per the representation-stability guidance, here is the blast radius rather than an assurance.

**Committed representation corpus** (`dump_normalized_corpus`, the #1442 harness), 18,432 rows across both axes and both shuffle directions, diffed row-for-row against `origin/main`:

| measure | value |
|---|---|
| rows | 18,432 |
| **moved** | **0 (0.0%)** |
| of which previously not a fixed point (free) | 0 |

Zero in all five shape families (`adjacent_junction_ins`, `del_plus_sub`, `dup_plus_ins`, `dup_plus_sub`, `split_vs_spanning_delins`).

That corpus draws 20-mers, so it cannot reach a changed block above `MAX_SPLIT_BLOCK` at all — the result above is a statement that nothing *else* moved, not evidence about the regime this PR changes. Movement is confined by construction to **equal-length changed blocks longer than 1024 bases**, which requires a description spanning >1 kb. Those are structural events, not the few-nucleotide descriptions the corpus and the consumer's library are made of.

**Directly measured in the regime that does change**, one normalize call per row, branch against `origin/main`:

| span | members (main) | members (branch) | elapsed (main) | elapsed (branch) |
|---|---|---|---|---|
| 1,100 | 550 | 550 | 28 ms | 44 ms |
| 2,000 | 1,000 | 1,000 | 80 ms | 109 ms |
| 3,000 | 1,500 | 1,500 | 169 ms | 210 ms |
| 3,800 | 1,900 | 1,900 | 262 ms | 352 ms |

(An equal-length spanning `delins` whose payload differs at every other base.) Member counts are **identical** — that shape reaches its members through the `delins`-to-substitution decomposition, not through the capped path — so it is not a representation change. What it does cost is roughly 30% more wall time on a >1 kb equal-length block, because `partition_block` now runs where it previously short-circuited and arrives at the same answer. Single rep, so read it as a magnitude rather than a figure.

## This does not open an unbounded path

`MAX_CANONICAL_WINDOW` (4096) already refuses a wider group *before* any block is partitioned — it bounds the reference window the canonicalizer will fetch at all. Measured on whole-span inversions:

| span | result |
|---|---|
| 2,000 nt | partitions, 1.6 ms |
| 3,000 nt | partitions, 0.3 ms |
| 4,000 nt and up | refused at the window, stays `inv` |

So this moves the equal-length confluence ceiling from 1024 up to that window bound, and leaves the guard that actually caps cost untouched. The reach is stated in the test module rather than left for someone to rediscover.

## Verification

- Suite **9368/9368**; all three oracles (`FERRO_ASSERT_IDEMPOTENT` / `_REPARSE` / `_IN_BOUNDS`)
- clippy `--all-features --all-targets -D warnings` and `cargo fmt --check` clean
- No proptest seed appended
- **The new guard is falsifiable**: against `main`'s `merge.rs` both tests in `residual_above_cap_confluence` fail, `g.257_1356inv` staying an inversion

The reproduction builds its near-palindrome programmatically rather than committing a 1100-base fixture, and pins the reduced form against the member list that same generator implies — so the assertion checks the substituted bases, not only that the digits appear somewhere in the output.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -1730,12 +1730,26 @@ const MAX_CANONICAL_WINDOW: i64 = 4096;
 /// Padding either side of the changed interval, giving the 3'-shift room.
 const CANONICAL_PAD: i64 = 128;
 
-/// Longest changed block the canonicalizer will attempt to partition.
+/// Longest **length-changing** block the canonicalizer will attempt to
+/// partition.
 ///
-/// This is a cost bound, not a policy: the alignment is quadratic in the block
-/// length, and a block this long is a structural event rather than a few
-/// nucleotide changes. It is deliberately far above the size at which the
-/// separation rule (`delins.md:17`) is the interesting question.
+/// Equal-length blocks are exempt: [`best_alignment`] returns their
+/// position-wise pairing immediately, so there is no gap placement to search for
+/// and nothing here to protect. They are bounded instead by
+/// [`MAX_CANONICAL_WINDOW`], the widest window the canonicalizer will fetch at
+/// all.
+///
+/// This is a cost bound, not a policy, and a block this long is a structural
+/// event rather than a few nucleotide changes. It is deliberately far above the
+/// size at which the separation rule (`delins.md:17`) is the interesting
+/// question.
+///
+/// It does **not** bound a quadratic cost, despite what an earlier revision of
+/// this comment claimed. A placement scores in O(1) because the score is
+/// separable, so [`best_alignment`]'s search is linear; the one quadratic step —
+/// ranking a tie by what it separates — carries its own and tighter bound in
+/// [`MAX_TIE_BREAK_SWEEP`], which binds first at every length this constant
+/// reaches.
 ///
 /// It is *not* the `delins.md:44-47` "coincidental alignment" guard. That
 /// concern — a large replacement whose interior only accidentally aligns, which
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1405, opened 2026-08-05
ci: stop caching cargo binaries in the two archive-building jobs

### LINKED ISSUE #1396: ci: set cache-bin: false on ci.yml's nextest jobs, matching the nightly and fuzz workflows
Follow-up from the #1218 cargo-cache migration. The same hazard is handled two different ways in two different files, and only one of them removes it at the source.

`Swatinem/rust-cache` leaves `cache-bin` at its default of `true`, so it caches `${CARGO_HOME}/bin`. Both `test-build` and `test-build-soak` in `ci.yml` install `cargo-nextest` via `taiki-e/install-action` and write a nextest archive that other jobs read, so producer and consumer must agree on the nextest version. A cache restore landing *after* the install would overwrite the just-installed binary with whichever version first populated the key.

- **`ci.yml` (#1370)** dodges this by **ordering** — the `rust-cache` step is placed before `install-action`, and a comment records why that order is load-bearing.
- **`nightly-soak.yml` and `fuzz.yml` (#1391)** instead set `cache-bin: "false"`, which removes the possibility rather than relying on step order. `fuzz.yml` has the sharper version of the problem: `cargo install cargo-fuzz` runs *above* the cache step.

Ordering works, but it is a constraint a future edit can silently break — reordering two `uses:` steps looks harmless. `cache-bin: "false"` cannot be broken that way.

## Proposal

Set `cache-bin: "false"` on `ci.yml`'s `test-build` and `test-build-soak` `rust-cache` steps, and reduce the ordering comment to a pointer at that option. Check whether any other job that installs a tool into `${CARGO_HOME}/bin` needs the same.

Worth confirming first that dropping binary caching does not measurably slow those two jobs — `install-action` downloads a prebuilt binary rather than compiling, so the expected cost is near zero, but it should be measured rather than assumed.

nh13:
Closes #1396. Base `main` — #1370 has merged, so this applies directly.

## What changes

`test-build` and `test-build-soak` both install `cargo-nextest` via `install-action` and write a nextest archive other jobs read, so producer and consumer must agree on the nextest version. `rust-cache` caches `${CARGO_HOME}/bin` unless told not to, so a restore can extract a previously-saved binary over the one just installed.

Both jobs handled that by **ordering** — install placed after the cache step, with a comment saying the order was deliberate. That works, but it is a constraint a future edit silently breaks: reordering two `uses:` steps looks harmless. `cache-bin: "false"` removes the hazard at the source, which is how `nightly-soak.yml` and `fuzz.yml` (from #1391) already do it.

## Scope — audited, not assumed

The issue asks to check whether any other job needs the same. Every `rust-cache` step against whether its job installs into `${CARGO_HOME}/bin`:

| job | rust-cache key | `cache-bin` | installs into cargo bin |
|---|---|---|---|
| `test-build` | `test-archive` | **false** | yes |
| `test-build-soak` | `test-archive-soak` | **false** | yes |
| `test`, `test-oracle` | *(none)* | — | yes |
| `soak`, `sweeps` | *(none)* | — | yes |
| `clippy` | `clippy` | default | no |
| `clippy-all-features` | `clippy-all-features` | default | no |
| `spec-fixtures` | `spec-fixtures` | default | no |
| `generated-docs` | `generated-docs` | default | no |
| `python-wheel-test` | `python-wheel` | default | no |
| `build` | `build-release` | default | no |

**No other job needs it.** The four other jobs that install a binary keep no cargo cache at all, and the six that keep one install nothing into `${CARGO_HOME}/bin`. `python-wheel-test` installs maturin through `pip`, into the Python venv rather than the cargo bin directory.

## Cost — measured, as the issue asks

#1396 asks that the near-zero cost be confirmed rather than assumed, since `install-action` downloads a prebuilt binary rather than compiling one.

Baseline from the three most recent successful `main` runs:

| job | run 30973321571 | 30970642225 | 30969775774 |
|---|---:|---:|---:|
| `Build test archive` | 83s | 83s | 84s |
| `Build soak archive (optimized)` | 124s | 143s | 140s |

I will post this PR's own numbers as a comment once CI completes, so the comparison is against the same jobs on the same runner class rather than against an assumption. Note the first run after a cache-key change is a miss by construction, so the *second* run is the one to read — the `cache-bin` change alters what is saved, not the key.

## Verification

- `actionlint .github/workflows/ci.yml` clean
- the workflow still parses as YAML
- no change outside the two `with:` blocks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -352,13 +352,23 @@ jobs:
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
         with:
           key: test-archive
-      # `install-action` stays AFTER the cache step, deliberately. `rust-cache`
-      # leaves `cache-bin` at its default of `true`, so `${CARGO_HOME}/bin` is
-      # still cached and a restore running second would extract a
-      # previously-saved `cargo-nextest` over the one just installed, pinning
-      # this job to whichever version first populated the key. The job that
-      # writes an archive and the jobs that read it have to agree on the nextest
-      # version, so let the install win.
+          # `rust-cache` caches `${CARGO_HOME}/bin` unless told not to, so a
+          # restore could extract a previously-saved `cargo-nextest` over the one
+          # `install-action` puts there and pin this job to whichever version
+          # first populated the key. This job writes an archive that other jobs
+          # read, and producer and consumer must agree on the nextest version.
+          #
+          # This used to be handled by ORDERING — the install step below sat
+          # after the cache step, with a comment saying that was deliberate. Not
+          # caching the binaries removes the hazard at the source instead (#1396),
+          # which matters because reordering two `uses:` steps looks harmless and
+          # would silently reintroduce it. `nightly-soak.yml` and `fuzz.yml`
+          # already do it this way.
+          #
+          # The cost is nil: `install-action` downloads a prebuilt binary rather
+          # than compiling one. Measured on this PR's own runs rather than
+          # assumed — see the PR body.
+          cache-bin: "false"
       - uses: taiki-e/install-action@8e83571c9a797c9e42a58141f098806bfec453a0  # nextest
       - name: Build test archive
         # Compiles every test binary and packs them with their run metadata.
@@ -410,9 +420,11 @@ jobs:
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
         with:
           key: test-archive-soak
-      # AFTER the cache step — see the same note on `test-build`. `rust-cache`
-      # still caches `${CARGO_HOME}/bin`, and this job writes the archive that
-      # `soak` and `sweeps` read.
+          # Same reason as `test-build` above, and it applies at least as
+          # strongly here: this job writes the archive that `soak` and `sweeps`
+          # read, so a stale cached `cargo-nextest` would desynchronise three
+          # jobs rather than one (#1396).
+          cache-bin: "false"
       - uses: taiki-e/install-action@8e83571c9a797c9e42a58141f098806bfec453a0  # nextest
       - name: Build soak archive
         # No `--features dev`. That feature pulls in `web-service` (axum, tokio,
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
Promised timings, from this PR's CI run ([30981371541](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/30981371541), all green) against the three-run `main` baseline captured before pushing.

| job | baseline (3 runs on `main`) | this PR |
|---|---|---|
| Build test archive | 83s / 83s / 84s | **81s** |
| Build soak archive (optimized) | 124s / 143s / 140s | **147s** |

**Read that as "no measurable change", not as a win or a regression.** One run against three is not enough to separate a couple of seconds from noise, and the soak job's own baseline already spans 19s.

More useful than the wall clock is what the logs say about the cache itself:

| | baseline | this PR |
|---|---|---|
| test-archive cache restored | 290,616,600 B | 290,611,604 B |
| soak-archive cache restored | 269,627,081 B | 269,629,647 B |
| crates compiled (`grep -c 'Compiling '`) | 0 / 0 | 0 / 0 |
| save step | `Cache up-to-date.` | `Cache up-to-date.` |

Two things follow, and the second is worth a reviewer's attention:

1. **Both jobs were already fully warm** — zero crates compiled in every run measured. There was never any compile time here for this change to save, so the wall-clock parity above is the expected result rather than a surprise.

2. **The cache did not shrink, because nothing was re-saved.** `cache-bin` changes what a save *writes*, not the restore key, so both jobs hit their existing key (`v0-rust-test-archive-…-c8387b21-4e621efc` / `…-c8342def-…`), reported `Cache up-to-date.`, and skipped the save. The restored entries are the ones written earlier *with* `cache-bin: true`, which is why they are byte-for-byte the same size (the few-KB deltas are noise between ref-scoped entries, not an effect).

So this PR's benefit is **deferred to the next key rotation**: from here on, whenever those entries are next written, they will not carry `${CARGO_HOME}/bin`. It does not reclaim the binaries already sitting in the current entries. That is the same "`actions/cache` never refreshes on an exact key hit" behaviour #1218 is about, and it means this should not be quoted as a cache-size reduction today.

The change still stands on the audit in the PR body — it aligns these two jobs with the nightly and fuzz workflows, and `test-build` / `test-build-soak` remain the only jobs with both `rust-cache` and an install into `${CARGO_HOME}/bin`.

PR #1407, opened 2026-08-05
fix(normalize): refuse to lower a repeat past a sibling inside its tract

### LINKED ISSUE #1325: Normalizer: a repeat whose growth exceeds its tract cannot be demoted, so it swallows a sibling's junction

### LINKED ISSUE #1415: Normalizer: lowering a repeat past a sibling inside its tract reorders the payloads and denotes the wrong bases
## What's wrong

`normalize` returns a well-formed description of bases the input never denoted. On `origin/main` at **d283a95d**, i.e. with #1392 merged:

```
reference  ("ACGT" x 64) + "GATCATAAATTCAGC" + ("ACGT" x 64)
           the `A` tract at 263_265 is three bases

in   NC_TEST.1:g.[262_263insAA;263_264insAA;264_265insC]
out  NC_TEST.1:g.264_265insCAAAA          <-- correct payload at that junction is AAAAC

intended   ...TAAAAAACAT...
emitted    ...TAACAAAAAT...
```

The output parses, re-parses, is in bounds and is a fixed point, so **all three seam oracles pass it**. Only the confluence proptest's sequence clause looks.

## Mechanism

`lower_repeat_edits` rewrites a repeat member into the `ins`/`del` it denotes and places that difference at the tract's 3' end. Its doc states the reasoning: a tract is a whole number of copies of one unit, so it does not matter which end the copies are added at.

That holds for a lone repeat. It fails when another member adds sequence **inside** the tract. On the merge loop's second pass the member set is `g.[263_265A[7];264_265insC]` — the `C` lands between the tract's own bases, so the four added `A`s belong 5' of it, and lowering puts them 3' of it. Re-deriving from the mis-applied window yields `g.264_265insCAAAA`.

Nothing downstream can catch it: by then the repeat is an ordinary `Ins` and the tract it came from is gone, so the derivation canonicalizes the wrong sequence.

## Severity

**High, and quiet.** Before #1392 this same input produced a visibly malformed overlapping pair (`g.[263_265A[7];264_265insC]`), which the apply oracle declines. #1392 made the sequence-first derivation authoritative, which turned a loud malformation into silently wrong bases. That is a regression in failure *mode* even though the underlying lowering bug is older.

## Fix

PR #1407 refuses the derivation when a repeat's tract strictly contains another member's junction — the same answer `apply_edits_to_window` already gives an insertion interior to a `del`/`delins`/`inv` span (#486). A junction at the tract's 3' end is adjacency rather than interference and still lowers.

That PR was opened against #1392's branch before it merged and has been rebased onto `main`; it carries the reproducer and a guard-validity check (stubbing the refusal reproduces `insCAAAA` exactly).

## Related

- #1392 — made the derivation authoritative, which is what converts this into wrong bases rather than a declined overlap.
- #1325 / #1399 — the *separate* defect that produces the `g.[263_265A[7];264_265insC]` member set in the first place. With that also fixed the set does not arise; this issue is the safety net for whatever else reaches the lowering.
- #1235 — found by its confluence proptest seed `4c6fab93`.

nh13:
**Closes #1415.**

Rebased onto `main` now that #1392 has merged. The PR originally targeted #1392's branch; GitHub retargeted it, and since #1392 squash-merged, the branch was rebuilt by cherry-picking this one commit onto `main` rather than replaying eight already-upstream commits.

**The defect is live in `main` as of `d283a95d`** — verified by stubbing the guard on this branch, which reproduces `TEMPLATE:g.264_265insCAAAA` exactly.

## The defect

```
reference  ("ACGT" x 64) + "GATCATAAATTCAGC" + ("ACGT" x 64)

g.[262_263insAA;263_264insAA;264_265insC]  ->  g.264_265insCAAAA
  correct payload at that junction is AAAAC

intended   ...TAAAAAACAT...
emitted    ...TAACAAAAAT...
```

It parses, re-parses, is in bounds and is a fixed point, so **all three seam oracles pass it**. Only the confluence proptest's sequence clause looks. Before #1392 the same input produced a visibly malformed overlapping pair that the apply oracle declines — so making the derivation authoritative converted a loud malformation into silently wrong bases.

## Mechanism

`lower_repeat_edits` places a repeat's growth at the tract's 3' end, on the documented reasoning that a tract is a whole number of copies of one unit so either end gives the same sequence. That holds for a lone repeat and fails when another member adds sequence *inside* the tract: with `264_265insC` between the tract's own bases, the four added `A`s belong 5' of the `C` and lowering puts them 3' of it. Nothing downstream can notice — by then the repeat is an ordinary `Ins` and its tract is gone.

## The fix

Refuse the derivation when a repeat's tract strictly contains another member's junction — the same answer `apply_edits_to_window` already gives an insertion interior to a `del`/`delins`/`inv` span (#486). A junction at the tract's **3' end** is adjacency rather than interference and still lowers; that case is pinned.

## Verification (re-run on `main`)

- Suite **9315/9315**; all three oracles **9315/9315**
- clippy `-D warnings` clean
- Guard validity: stubbing the refusal turns both defect tests red — one printing the exact `TEMPLATE:g.264_265insCAAAA` — while the adjacency guard stays green

An earlier draft of those tests was **vacuous**: it addressed `NC_TEST.1`, which `cis_apply_oracle`'s provider does not serve, so normalization declined and echoed the input and all three "passed" against a stubbed fix. The same stub-then-restore check caught it.

## Relationship to #1399

#1399 fixes #1325, the *separate* defect that produces the `g.[263_265A[7];264_265insC]` member set in the first place. With that in, the set does not arise at all. This PR is the safety net: it guarantees the derivation never answers with the wrong bases, whatever upstream hands it. The two are independent and can merge in either order.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2602,12 +2602,28 @@ fn collect_canonical_edits(
 /// `del`, `ins` and `unit[N]`: that is a rendering decision, and the derivation
 /// only needs the bases.
 ///
-/// Where inside the tract the difference is placed does not matter. The tract
-/// is a whole number of copies of one unit, so adding or removing copies at
-/// either end yields the same sequence — which is why this can ignore the
-/// shuffle direction that `normalize_repeat` must honour. The 3' end is chosen
-/// so the footprint stays as small and as far from the group's other members as
-/// possible.
+/// Where inside the tract the difference is placed does not matter **as long as
+/// no sibling sits inside it**. The tract is a whole number of copies of one
+/// unit, so adding or removing copies at either end yields the same sequence —
+/// which is why this can otherwise ignore the shuffle direction that
+/// `normalize_repeat` must honour. The 3' end is chosen so the footprint stays
+/// as small and as far from the group's other members as possible.
+///
+/// That premise fails when another member's junction falls strictly inside the
+/// tract, and it fails silently. `g.[263_265A[7];264_265insC]` puts the
+/// insertion's `C` between the tract's own bases, so the four added `A`s are
+/// *before* it; lowering the repeat to an insertion at the tract's 3' junction
+/// puts them after, and the applied window comes back as `TAACAAAAAT` instead of
+/// `TAAAAAACAT`. Nothing downstream can notice, because by then the repeat is an
+/// ordinary `Ins` and the tract it came from is gone — the derivation simply
+/// canonicalizes the wrong sequence, and the caller gets a well-formed
+/// description of bases the input never denoted.
+///
+/// Such a pair overlaps and has no single resulting sequence to derive from, so
+/// this refuses rather than picks an order — the same answer
+/// `apply_edits_to_window` already gives an insertion interior to a `del`,
+/// `delins` or `inv` span (#486), applied to the one span-bearing member that
+/// reaches here still carrying its tract.
 ///
 /// Refuses (leaving the group to the per-member pipeline) when:
 ///
@@ -2619,11 +2635,38 @@ fn collect_canonical_edits(
 /// * the tract does not span the anchor, which `count_tandem_repeats` can
 ///   return (it back-scans first, so a run lying entirely 5' of the anchor is
 ///   reachable);
+/// * another member's junction falls strictly inside a repeat's stated span,
+///   per the paragraph above — the pair overlaps and denotes no single
+///   sequence, so there is nothing to derive from;
 /// * `count` equals the reference count, i.e. the member is an identity. The
 ///   resulting sequence cannot express one, so a group carrying one would come
 ///   back with that member silently dropped — the same reason `NaEdit::Identity`
 ///   is refused above.
 fn lower_repeat_edits(edits: &mut [GEdit], ref_bytes: &[u8], w_lo: i64) -> Option<()> {
+    // The junction each member adds sequence at, for the interior test below.
+    // An insertion's is its gap; a duplication's is the 3' end of what it
+    // copies. Every other kind claims bases rather than a junction and is
+    // already covered by the applier's own interior test.
+    let junction_of = |edit: &GEdit| match edit {
+        GEdit::Ins { gap, .. } => Some(*gap),
+        GEdit::Dup { e, .. } => Some(*e),
+        _ => None,
+    };
+    for i in 0..edits.len() {
+        let GEdit::Repeat { s, e, .. } = edits[i] else {
+            continue;
+        };
+        // Strictly inside, matching `demote_repeats_spanning_siblings`'
+        // `spans_sibling_junction`: a junction at the tract's 3' end is flush
+        // against it rather than within it, and that adjacency is legitimate.
+        if (0..edits.len())
+            .filter(|&j| j != i)
+            .filter_map(|j| junction_of(&edits[j]))
+            .any(|junction| junction >= s && junction < e)
+        {
+            return None;
+        }
+    }
     for edit in edits.iter_mut() {
         let GEdit::Repeat { s, e, unit, count } = edit else {
             continue;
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1408, opened 2026-08-05
ci(soak): name the property the nightly soak actually failed on

nh13:
> ⚠️ **This commit is unsigned.** `git commit -S` failed three times in a row (1Password agent errors, then a hang). Per repo convention I fell back to `--no-gpg-sign` after the third failure rather than skipping signing silently. **Please re-sign before merging.**

## The short answer to "does the nightly alert on failure?"

**Yes.** `nightly-soak.yml` opens one rolling issue and comments on it thereafter, and `report-failure.yml` gives every other scheduled workflow the same. Coverage is complete — every workflow that runs post-merge or on a schedule either calls the reporter or files its own:

| workflow | alerting |
|---|---|
| `audit-pr-runs`, `coverage`, `external-validation`, `fuzz`, `nightly-mutalyzer`, `nightly-perf` | `report-failure.yml` |
| `nightly-soak` | its own rolling issue |

So this PR is not adding alerting. It fixes what the alert **says**.

## The defect

The soak job selects `test(proptest)` — **two** modules — but the report names one:

```
TITLE="nightly soak: idempotency property failed"
...add the seed line to tests/proptest-regressions/normalize_idempotency_proptest.txt
cargo nextest run --features dev -E 'test(normalize_idempotency_proptest)'
```

`cis_allele_confluence_proptest` is in the same selection. It has found seventeen defects in that corpus, and since its indel property stopped being `#[ignore]`d it is the *likelier* of the two to fire. A confluence failure would be reported under the other property's name, pointing at the wrong seed file and the wrong reproduce command.

That is worse than a cosmetic mislabel, because **which seed file grew is the reliable signal** for which property failed. The reader would be sent to inspect the one file that had not changed.

## What changes

- **The body names neither module.** It says which log group identifies the failing property, then gives both files with their matching commands.
- **The capture step adds a `git status --porcelain tests/proptest-regressions/` group**, so "which file grew a seed" is explicit rather than something to eyeball across two printed files — one of which now carries fourteen seeds.
- **The strategy caveat is included.** A pinned seed is replayed *through* its strategy, so editing how that strategy draws randomness makes every seed recorded against it regenerate a different case. That is the failure mode that silently empties a regression file, and it is worth saying at the moment someone is about to add a seed.
- **The title changes**, so the search that keeps ONE rolling issue now falls back to the old wording — an issue already rolling under the previous title keeps being updated rather than being duplicated alongside a new one.

## Verification

`actionlint` clean, and the report job's own `run:` block was extracted from the workflow with a YAML parser and executed with `gh` stubbed, rather than eyeballed:

```
[gh issue list --state open --search nightly soak: a proptest property failed in:title ...]
[gh issue list --state open --search nightly soak: idempotency property failed in:title ...]
[gh issue create --title nightly soak: a proptest property failed --body ...]
```

The body renders with its code fence intact and both commands present, and the fallback search fires when the new title matches nothing. Worth doing directly: a first, hand-retyped version of that check mis-quoted the `#` comment lines and swallowed the whole code block, which is exactly the class of bug that would otherwise ship unnoticed into a path that only runs when something is already broken.

### DIFF
--- a/.github/workflows/nightly-soak.yml
+++ b/.github/workflows/nightly-soak.yml
@@ -151,12 +151,20 @@ jobs:
         # proptest writes the shrunk case to its persistence file. Surfacing it
         # here means the reproducer is in the run log even if the artifact is
         # missed — and that file is exactly what gets committed alongside the
-        # eventual fix, per the convention in
-        # tests/proptest-regressions/normalize_idempotency_proptest.txt.
+        # eventual fix, per the convention in tests/proptest-regressions/.
+        #
+        # Both modules' files are printed, because this job selects
+        # `test(proptest)` and either can be the one that grew a seed.
         if: failure()
         run: |
           echo "::group::proptest regression seeds"
           find tests/proptest-regressions -name '*.txt' -print -exec cat {} \;
+          echo "::endgroup::"
+          # Which file proptest appended to is the reliable signal for *which*
+          # property failed — the run log above prints all of them, and a
+          # 14-seed file looks much like a 3-seed one at a glance.
+          echo "::group::seed files this run modified"
+          git status --porcelain tests/proptest-regressions/ || true
           echo "::endgroup::"
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7.0.1
         if: failure()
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1410, opened 2026-08-05
test(ci): pin every sweep_seeds consumer to SWEEP_FILTER

### LINKED ISSUE #1397: test(cis): enforce that every sweep_seeds consumer is named in SWEEP_FILTER
Follow-up from #1395, which makes `sweep_seeds` draw a 4-seed corpus prefix by default and has the `sweeps` job in `ci.yml` set `FERRO_SWEEP_SEEDS: full` to restore the exhaustive run.

That pairing is correct today — verified: the three modules consulting `sweep_seeds` (`cis_junction_crossing_shift`, `repeat_span_sibling_overlap`, `issue_1254_sibling_crossing_shift`) are all named in `SWEEP_FILTER`, and `sweeps` is the only job that sets the variable, so exhaustive coverage still runs on every PR.

But nothing *enforces* it. #1395's own comment says so:

> If a sweep is ever moved out of `SWEEP_FILTER`, it silently drops to the prefix here — check both when editing that filter.

That is a human instruction, and the failure mode it guards against is the expensive kind: a new sweep module that consults `sweep_seeds` but is not named in `SWEEP_FILTER` runs at the 4-seed prefix in `test`/`test-oracle` and **never** gets `full` anywhere. CI stays green and gets *faster*, which reads as success rather than as the coverage loss it is.

Note the existing count invariant does not cover this. The `8878 + 30 == 8908` comment and the `--no-tests=fail` on `sweeps` both catch a filter that matches *nothing* (a rename), not a module that is missing *from* it.

## Proposal

Add a test asserting that every consumer of `sweep_seeds` is named in `ci.yml`'s `SWEEP_FILTER`. The repo already has exactly this shape in `tests/it/coderabbit_config_paths.rs`, which parses `.coderabbit.yaml` and fails the suite when a `path:` glob matches nothing — so a rename fails loudly instead of rotting silently. The same treatment here would make the sweep pairing enforced rather than remembered.

Either direction works:
- parse `SWEEP_FILTER` out of `ci.yml` and assert each `sweep_seeds` caller appears in it; or
- invert it — have the sweep modules register themselves in a shared list that both the test and the filter derive from.

nh13:
Closes #1397.

`sweep_seeds` (#1295) draws a 4-seed corpus prefix by default, and `ci.yml`'s `sweeps` job restores the full corpus by setting `FERRO_SWEEP_SEEDS: full`. That job selects on `SWEEP_FILTER`, so the pairing only holds while every module consulting the knob is named in that filter. #1295 left a comment asking the reader to edit the two together; #1397 asks for the invariant instead.

**The failure mode is the expensive kind.** A sweep module that consults the knob but is not named in `SWEEP_FILTER` runs at the default prefix in every job and never receives `full` anywhere. CI stays green and gets *faster* — it reads as success rather than as the coverage loss it is.

Neither existing guard catches it. The partition recorded in `SWEEP_FILTER`'s own comment, and the `--no-tests=fail` on `sweeps`, both fire when the filter matches *nothing* — a rename. An unnamed consumer instead lands quietly on the negated side, where it still runs and still counts.

## What this adds

`tests/it/sweep_filter_invariant.rs`, modelled on `coderabbit_config_paths.rs` as the issue suggests, asserting the pairing in both directions:

- `every_seed_knob_consumer_is_named_in_the_sweep_filter` — the coverage hole above.
- `every_module_named_in_the_sweep_filter_consumes_the_seed_knob` — the converse. Not a coverage hole, but a claim that stopped being true, and the next person editing the filter would reasonably trust it.

Three implementation points worth review:

- **`SWEEP_FILTER` is a `>-` folded scalar spanning two lines**, so the value is reassembled by indent rather than read from the first line alone. Reading line one only would silently exempt `issue_1254_sibling_crossing_shift`, which sits on line two — precisely the blind spot the file exists to close.
- **Only the top level of `tests/it/` is scanned.** `common/cis_apply_oracle.rs` *defines* the knob rather than consulting it, and is not a target CI could select anyway.
- **The scan matches `sweep_seeds(`**, the call form, so a prose mention does not count as consumption. The file excludes itself for that reason — it names the knob throughout.

## Verification

Both tests pass against the current tree (all three consumers are already named in the filter, so this pins today's state rather than changing it). A passing invariant test proves little on its own, so each direction was watched to fail:

| control | result |
|---|---|
| drop the **line-2** entry (`issue_1254…`) from `SWEEP_FILTER` | consumer test FAILS — confirms the folded-scalar continuation is genuinely read |
| add a stale entry (`test(spdi_tests)`) to `SWEEP_FILTER` | converse test FAILS, naming it |
| add a new file calling `sweep_seeds(4)` | consumer test FAILS, naming it |

The third is the real-world trigger, and it fires even though the new file had no `mod` declaration — the scan is filesystem-based, so it catches a new sweep before it is wired up.

Also confirmed the `SWEEP_FILTER` partition invariant still holds with the two new tests on the negated side: `not test(proptest)` selects 9247, negating the filter selects 9209, the filter selects 38, and 9209 + 38 == 9247. The absolute counts in ci.yml's comment (8908/8878/30) have drifted since they were written, which that comment explicitly anticipates — "the absolute counts drift as tests are added, the sum must not" — so they are left alone rather than re-pinned to numbers that would go stale again.

`cargo fmt --check`, `cargo clippy --all-features --all-targets -- -D warnings`, and the full suite under all three oracles (`FERRO_ASSERT_IDEMPOTENT` / `REPARSE` / `IN_BOUNDS`) are green: 9262 passed.

## Base

Stacked on #1395 (`perf/issue-1295-sweep-seed-knob`), which introduces the knob this pins. Retarget to `main` once that merges.

### DIFF
--- a/tests/it/main.rs
+++ b/tests/it/main.rs
@@ -398,6 +398,7 @@ mod strict_explicit_seq_size_modes;
 mod strict_rejection_survives_normalization;
 mod strict_whitespace_mode;
 mod supernumerary;
+mod sweep_filter_invariant;
 mod tool_support_matrix;
 mod trans_mosaic_mixed;
 mod uta_loader_tests;
--- a/tests/it/sweep_filter_invariant.rs
+++ b/tests/it/sweep_filter_invariant.rs
@@ -1,0 +1,188 @@
+//! Liveness check for the pairing between the sweep seed knob and CI.
+//!
+//! `sweep_seeds` (#1295) draws a small prefix of the corpus by default, and
+//! `ci.yml`'s `sweeps` job restores the exhaustive run by setting
+//! `FERRO_SWEEP_SEEDS: full`. That job selects on `SWEEP_FILTER`, so the pairing
+//! only holds while every module consulting the knob is named in that filter.
+//!
+//! Nothing enforced it. #1295's own comment said "if a sweep is ever moved out
+//! of `SWEEP_FILTER`, it silently drops to the prefix here — check both when
+//! editing that filter", which is a human instruction, and #1397 asks for the
+//! invariant instead.
+//!
+//! **The failure mode is the expensive kind.** A new sweep module that consults
+//! the knob but is not named in `SWEEP_FILTER` runs at the default prefix in
+//! `test` / `test-oracle` and never receives `full` anywhere. CI stays green and
+//! gets *faster*, which reads as success rather than as the coverage loss it is.
+//!
+//! The existing guards do not cover it. `SWEEP_FILTER`'s own comment records a
+//! partition — the negated selection plus the filter's own selection account for
+//! the whole suite — and `sweeps` passes `--no-tests=fail`. Both catch a filter
+//! that matches *nothing*, i.e. a rename. Neither catches a module missing *from*
+//! it: an unnamed consumer simply lands on the negated side, where it still runs
+//! and still counts, just never at the full corpus.
+//!
+//! Modelled on `coderabbit_config_paths.rs`, which makes the same class of
+//! silent config rot loud.
+
+use std::path::PathBuf;
+
+/// This file. Excluded from the consumer scan below, which would otherwise
+/// match it — not because of the prose (that spells the knob in backticks, and
+/// the scan looks for the call form `sweep_seeds(`), but because the scan's own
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai pause

Issue #1406, opened 2026-08-05
Normalizer: W5002 reports whether the repair succeeded, not whether the input conflicts


nh13:
## The question

W5002 / `OverlapConflictingEdits` is supposed to say "these members conflict, so strict mode must reject them". Today it says something narrower: **"these members still conflict *after* canonicalization tried to re-spell them"**. The verdict tracks whether the repair succeeded, not whether the input was conflicting.

That makes strict mode's answer depend on something a caller cannot see, and it is self-inconsistent in a way that is easy to demonstrate. This is a question about intended error semantics rather than a reported user-facing bug, which is why it is filed as a question rather than as a defect.

## Measured, on `origin/main` at `2646f00`

Reference `TTTTTTTTTAATATATTTTAATAC` (24 bases) for the first two, `CGCGCGCGCAATCGCGCG` (18 bases) for the third.

| input | lenient output | strict(input) | strict(output) |
|---|---|---|---|
| `g.[23dup;23A>G]` | `g.22_23insG` | **ACCEPT** | ACCEPT |
| `g.[24dup;24C>G]` | *unchanged* | **REJECT** | REJECT |
| `g.[11_12inv;11_12insAA]` | `g.[11_12=;10_11A[4]]` | REJECT | **ACCEPT** |

Three separate problems, one mechanism.

**Row 1 vs row 2 — the same shape gets opposite verdicts.** A duplication and a substitution both claiming one base is the same geometric relationship at 23 as at 24. At 23 the repair collapses the pair into a single `g.22_23insG`, so by the time anything asks about overlap there is nothing left to conflict and strict accepts. At 24 the repair is declined — correctly, by #1307's terminal fix, since re-spelling there names a position the contig does not have — so the conflict survives and strict rejects. The only difference between the two inputs is *proximity to the end of the contig*.

**Row 3 — canonicalization launders a conflict.** The input is strict-rejected. Its own lenient output is strict-**accepted**. So `normalize` in lenient mode converts a description strict mode refuses into one it admits, with no warning surviving the trip. This is the same mechanism as rows 1 and 2, seen from the other side: a successful re-spelling erases the evidence.

There is a test that already names this hazard for the terminal case — `issue_1307_terminal_dup_respell::the_declined_output_is_still_rejected_by_strict`:

[...]

Issue #395, opened 2026-05-21
normalize: 6 misc gaps — repeat resumption, cross-ref ins, RNA U→T, Python warning surface, W5002 strict mode


nh13:
## Summary

Six unrelated normalize gaps spanning repeats, cross-reference ins expansion, RNA-to-DNA translation, Python warning surface, and strict-mode promotion. Each is small enough that a single grouped PR is reasonable.

### 1. MultiRepeat partial validation: interleaved Exact resumption

`src/normalize/validate.rs:248-261` walks repeat units left-to-right and `break`s at the first non-`Exact` count. For `[Exact(3), Range(1,5), Exact(2)]`, the trailing `Exact` entries are never validated. Documented design choice in the docstring at lines 218-226 and PR #296 body; worth revisiting now that the design is in flight.

### 2. `ins[<acc>:g.A_B]` cross-reference expansion

`src/normalize/rules.rs:1818-1822, 1844` return `UnsupportedVariant` with `"cross-reference"` + `"see follow-up"`. `ExternalRef` case at line 1917-1921 is unchanged. Spec allows; ferro defers.

### 3. `ins[244-8_249]` intronic / UTR-marker CDS-offset range expansion

`src/normalize/rules.rs:1801, 1836` return `UnsupportedVariant` with `"CDS-offset range (intronic or UTR-marker) is spec-undefined and not yet supported by ferro; see follow-up"`.

### 4. `r.` Base::U plus-strand U→T translation

`src/project/projector.rs:291-294` documents: "plus-strand RNA inputs whose edit carries `Base::U` are forwarded unchanged into the g. output, producing an invalid DNA emission." `src/project/edit.rs:11-14` returns `edit.clone()` unchanged on `Strand::Plus`. The `Base::U → Base::A` map at `edit.rs:71` runs only via `complement_base` (Minus-strand path).

### 5. `NormalizationWarning` not exposed to Python

`src/normalize/mod.rs:104` defines `pub enum NormalizationWarning { ... }` (RefSeqMismatch, OverlapConflict, CanonicalSplitSkipped, InsertedSequenceExpanded). `src/python.rs` has zero references; both entry points (`src/python.rs:179, :411`) call the warning-stripped `Normalizer::normalize`.

### 6. W5002 OverlapConflict not promoted to error in strict mode

[...]

PR #1411, opened 2026-08-05
fix(normalize): take the overlap verdict from the input, not the repair

nh13:
Refs #1406 — fixes the inconsistent-verdict half. Deliberately **not** `Closes`: one of that issue's three rows stays open, and investigating it turned up a second defect. Both are documented below and on the issue.

## What was wrong

`detect_overlap_conflicts` ran only **post-shift**, on the normalized members, so W5002 reported whether the *repair* had succeeded rather than whether the input conflicted:

```
reference  TTTTTTTTTAATATATTTTAATAC     24 bases, 23 = `A`, 24 = `C`

g.[23dup;23A>G]  ->  g.22_23insG    no warning at all, strict ACCEPTS
g.[24dup;24C>G]  ->  unchanged      W5002, strict REJECTS
```

Both are a `dup` of one base beside a substitution of that same base — the same shape, one position apart. At 23 the repair collapses the pair into a single member, so by the time the post-shift detector looked there was nothing left to conflict. At 24 the terminal decline (#1307) leaves the pair intact and it is reported.

**Proximity to the end of the contig decided whether a conflicting description was rejected**, which is not a property of the description.

## The fix

`detect_insertion_overlaps` already ran on the raw members, and for exactly this reason — a merge would otherwise hide the overlap. Its coincident-bounds sibling now runs there too, and a conflicting allele is preserved as authored. That is ferro's standing answer since #395/#486/#1004: report the conflict and leave the description alone rather than pick a winner among orderings the spec does not rank.

## Why it is scoped to coincident bounds

I tried widening the same preserve to the insertion-interior-to-a-span conflicts and **measured what it costs**: three tests, all pinning on purpose that such an allele still has each member normalized on its own (`c.[4_10inv;5_6insAA]` settles as `c.[5_9inv;6_7insAA]`).

```
FAIL  idempotency_tests::test_overlap_conflicting_allele_is_idempotent_across_respellings
FAIL  issue_1235_transcript_axes::overlap_conflicting_allele_is_not_canonicalized
FAIL  issue_1276_dup_junction_overlap::lenient_output_of_a_conflict_is_still_rejected_by_strict
```

Preserving those verbatim would drop the per-member 3' rule. So the narrow version is the one that ships, with **zero blast radius** — no existing test's expectation moved.

## What stays open, and one new finding

**Row 3 of #1406 is not closed.** `g.[11_12inv;11_12insAA]` is strict-rejected while its own lenient output `g.[11_12=;10_11A[4]]` is strict-accepted. The laundering there is produced by *per-member* normalization — the inversion cancels to `=`, the insertion respells as a repeat — which is precisely what the three tests above require for a conflicting allele. Closing it means stopping a member's own respelling from erasing the conflict, a different mechanism.

**A second defect on that same input, found while investigating it and verified pre-existing by A/B against `origin/main`:** the output is not idempotent.

```
g.[11_12inv;11_12insAA] -> g.[11_12=;10_11A[4]] -> g.[10_11A[4];11_12=]
```

The two members come back in genomic order, so the *first* output carries out-of-order members — #1235's criterion 2. It has **no executable pin in this PR**, because a test that normalized that input would fire `FERRO_ASSERT_IDEMPOTENT` and redden the `test-oracle` job rather than pin anything. It is recorded in the test module's doc and on #1406.

## Verification

- Suite **9277/9277**; all three oracles **9277/9277**
- Conformance axis **201/201** against the prepared reference
- clippy `-D warnings` and `cargo fmt --check` clean

Guard validity checked by stubbing the new branch: both defect tests go red, both guards stay green. Worth noting the fixed-point test initially passed *against the stub* — the repaired output is also a fixed point, so asserting only stability guarded nothing. It now asserts that the stable form is the preserved one.

---

## ⚠️ Representation change — read this before merging

Per the repository's representation-stability rule, this PR **moves normalized output** and the movement is enumerated here rather than left to be discovered downstream. A consumer that keys on the normalized HGVS string must re-normalize the affected shapes.

The two commits move representation in opposite directions, and the second is what keeps the net cost to one migration instead of two.

### Expensive — previously ACCEPTED, now preserved and strict-rejected

These are the laundering rows: coincident-**write** members that today normalize to a single accepted edit, silently dropping or reordering an edit. Roughly 15 shapes, all of the form "two members claim the same bases":

| input | was (shipped) | now |
|---|---|---|
| `g.[15_16del;15_16inv]` | `g.15_16del` | preserved, REJECT |
| `g.[1_2del;1_2inv]` | `g.1_4delinsAA` | preserved, REJECT |
| `g.[5_6del;5_6inv]` | `g.5_8delinsAA` | preserved, REJECT |
| `g.[13_14del;13_14inv]` | `g.[13_14=;15_16del]` | preserved, REJECT |
| `g.[16T>A;16del]` | `g.16_17delinsA` | preserved, REJECT |
| `g.[2_3del;2_3delinsTT]` | `g.1_9T[7]` | preserved, REJECT |
| 11 coincident `repeat`+`inv`/`del` rows | a single accepted edit | preserved, REJECT |

**In six of those repeat rows the inversion was being deleted outright** — `g.[4_9AT[3];4_9inv]` normalized to `g.4_9AT[3]`, losing the `inv` with no warning. That is data loss, not a representation preference, and it is the strongest argument for taking the migration.

### Cheap — previously REJECTED, now accepted

Nothing was stored for these, because strict mode refused them:

| input | was | now |
|---|---|---|
| `g.[24dup;24C>G]` | preserved, REJECT | `g.23_24insG` |
| `m.[24dup;24C>G]` | preserved, REJECT | `m.23_24insG` |
| `g.[24dup;24delinsGG]` | preserved, REJECT | `g.[24delinsGG;24dup]` |

### Explicitly NOT moved

The second commit exists partly to keep these fixed. Verified byte-identical: `g.[23dup;23A>G]` → `g.22_23insG`, `g.[23A>G;23dup]` → `g.22_23insG`, `g.[23dup;24C>G]` → `g.24delinsAG`, `g.[13_14dup;13_14inv]` → `g.[13_14=;15_16dup]`, `g.[13_14dup;13_14delinsGG]` → `g.12_13insGG`, `g.[4_9dup;4_9inv]` → `g.4_9AT[6]`, `g.[4_9dup;6A>C]` → `g.5_6insCTATAT`.

Without the coincident-bounds correction, the preserve in the first commit would have moved all six of those `dup` forms too — and a later correction would have moved them back. Landing them together is one migration rather than two.

### Blast radius

9315/9315 with all three oracles. Seven tests moved, all conflict-policy pins, in three files. **Zero** movement in `clinvar_hgvs_tests`, `cmrg_exhaustive_tests`, `paraphase_exhaustive_tests`, the spec-fixture census, or any exhaustive sweep.

### #1307 is re-blessed, deliberately

Its "unreachable by policy" paragraph reasoned from the premise that a `dup` beside a substitution of the same base is a W5002 conflict. It is not — `duplication.md:5` puts the copy 3' of the span, so the two write to different places. Its reported defect (a coordinate one base past the contig) is fixed *more* cleanly: the merged form never constructs an out-of-range coordinate. The four guards it cited as blocking all still pass.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -2700,6 +2700,80 @@ impl<P: ReferenceProvider> Normalizer<P> {
             allele.phase,
         ));
 
+        // Coincident *bounds* among the raw members, for exactly the reason the
+        // insertion detector above runs pre-merge — and the half that was
+        // missing (#1406).
+        //
+        // `detect_overlap_conflicts` also runs post-shift, on the normalized
+        // members, which meant W5002 reported whether the repair had *succeeded*
+        // rather than whether the input conflicted. Two inputs of the same shape
+        // got opposite verdicts:
+        //
+        // ```text
+        // reference  TTTTTTTTTAATATATTTTAATAC     24 bases
+        // g.[23dup;23A>G]  -> g.22_23insG          no warning, strict ACCEPTS
+        // g.[24dup;24C>G]  -> unchanged            W5002, strict REJECTS
+        // ```
+        //
+        // Both are a `dup` of one base beside a substitution of that same base.
+        // The only difference is that at 23 the repair collapses the pair into a
+        // single member, so by the time the post-shift detector looks there is
+        // nothing left to conflict; at 24 the terminal decline (#1307) leaves
+        // the pair intact and it is reported. Proximity to the end of the contig
+        // decided whether a conflicting description was rejected.
+        //
+        // The same mechanism let lenient mode launder a conflict outright:
+        // `g.[11_12inv;11_12insAA]` is strict-rejected, and its own lenient
+        // output `g.[11_12=;10_11A[4]]` is strict-*accepted*.
+        //
+        // So the verdict is taken from the input, where the conflict lives, and
+        // a conflicting allele is preserved as authored — the same answer
+        // `has_same_gap_insertions` gives just below, and ferro's standing one
+        // since #395/#486/#1004: report the conflict and leave the description
+        // alone rather than pick a winner among orderings the spec does not
+        // rank.
+        // Scoped to *coincident bounds*, and deliberately not widened to the
+        // insertion-interior-to-a-span conflicts `detect_insertion_overlaps`
+        // reports. Those alleles are still expected to have each member
+        // normalized on its own — `[4_10inv;5_6insAA]` settles as
+        // `[5_9inv;6_7insAA]`, the inversion 3'-shifted and the insertion
+        // respelled at its own junction — and preserving them verbatim would
+        // drop the per-member 3' rule that #1276 and #1235's transcript axes
+        // pin. What those alleles do not get is whole-allele re-derivation,
+        // which they already did not.
+        let raw_conflicts =
+            crate::normalize::overlap::detect_overlap_conflicts(&allele.variants, allele.phase);
+        if !raw_conflicts.is_empty() {
+            // Harvest the per-member diagnostics before preserving. Returning
+            // here skips the per-member loop below, which is where every
+            // *member-local* finding is raised — so without this a conflicting
+            // allele that ALSO misstates a reference base reports only the
+            // conflict. Measured on `g.[13G>A;13A>C]` (base 13 is `A`): the
+            // warning set went from `[RefSeqMismatch, OverlapConflict]` to
+            // `[OverlapConflict]`, and strict mode's error changed from
+            // "Reference mismatch at 13-13: expected G" to the coincidence
+            // message. A reference mismatch is a data-integrity signal about the
+            // caller's input, independent of whether the members also collide;
+            // losing it to a sibling's conflict is a strictly worse report.
+            //
+            // Each member is normalized in isolation purely to collect what it
+            // says — the results are discarded, because preserving the authored
+            // members is the whole point of this branch. A member that fails to
+            // normalize contributes nothing rather than aborting: the allele is
+            // being refused anyway, and the conflict below is the finding that
+            // matters.
+            for member in &allele.variants {
+                if let Ok((_, member_warnings)) = self.normalize_core(member) {
+                    all_warnings.extend(member_warnings);
+                }
+            }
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

nh13:
**Reconciliation note — `main` gained an overlapping raw-member check when #1392 merged, and this PR is still needed.**

`normalize_allele` on current `main` now computes:

```rust
let input_has_overlap_conflict = all_warnings.iter().any(...)
    || crate::normalize::overlap::detect_overlap_conflicts(&allele.variants, allele.phase)
        .iter().any(...);
let reorder_before_merge = is_cis && !allele.uncertain && !input_has_overlap_conflict;
```

That is the same detector on the same raw members this PR calls. Measured on `main` (+ an unrelated branch that does not touch overlap detection), the two #1406 rows are **still inconsistent**, so it does not subsume this change:

| input | lenient | strict |
|---|---|---|
| `g.[23dup;23A>G]` | `g.22_23insG` | **ACCEPT** |
| `g.[24dup;24C>G]` | unchanged | **REJECT** |

The two uses are complementary and worth folding into one call when this rebases:

- **`main`'s** gates *ordering* — whether to sort members before the merge.
- **this PR's** gates *preservation* — whether to hand the allele back as authored, which is what makes the verdict come from the input rather than from whether the repair succeeded.

Two things worth carrying across in the rebase:

1. They can share one `detect_overlap_conflicts(&allele.variants, …)` result instead of calling it twice on the same input.
2. `main`'s new comment states a limitation that bears directly on the rows this PR leaves open: *"A conflict that emerges only from the per-member 3' shift is not covered by this input-time gate."* That is the same seam as #1406 row 3 and #1416 — a conflict that exists at input time is caught, one the pipeline creates or destroys is not.

Flagging rather than changing it: this branch appears to be in another session's hands (its worktree was renamed to `pr-1411` and the branch rebased onto post-#1392 `main` at `4a281355`, ahead of what is pushed here), so I have not touched it.

PR #1413, opened 2026-08-05
ci: let external-validation's test steps fail the job

### LINKED ISSUE #1404: ci: external-validation's test steps cannot fail the job (continue-on-error + unpiped tee)
Surfaced while reviewing #1391 (which only replaces that workflow's cargo-cache step and does **not** touch these lines — this is pre-existing).

`.github/workflows/external-validation.yml`'s API-comparison, SPDI and dbSNP steps each carry `continue-on-error` and pipe `cargo test` through `tee`. Two independent ways for a failing validation run to leave the job green:

1. **`continue-on-error`** — the step's failure never reaches the job conclusion.
2. **`cargo test … | tee …`** — without `pipefail`, the pipeline's exit status is `tee`'s, which is essentially always 0. So even removing `continue-on-error` alone would not make these steps fail.

This is exactly what the repo's own `.github/workflows/**` review instruction says to flag: *"a `continue-on-error` or unchecked step that lets a failing conformance, proptest, or idempotency run pass the job."*

Note the contrast with the nightly reference-aware job, where `continue-on-error` **is** deliberate and documented (its purpose is to surface drift in the xfail report rather than to gate, and the corpus runner wraps normalization in `catch_panics`). No equivalent rationale is recorded here, which is what makes this look like drift rather than a decision.

## Proposal

- Drop `continue-on-error` from the three validation steps, or record why it is deliberate the way the nightly does.
- Add `shell: bash` + `set -euo pipefail` (or `SHELLOPTS`) so the `tee` pipelines propagate `cargo test`'s status.
- Keep the report-upload steps running regardless, via `if: always()` on the upload rather than by masking the test's exit code.

Worth checking whether these steps are currently failing in the external-validation runs — if they are, fixing this will turn that workflow red, and the underlying failures need triage first.

nh13:
Closes #1404.

The issue asked whether these steps are currently failing, and warned that fixing the gate would turn the workflow red if they were. The answer turned out to be worse and simpler: **they have not been running at all.**

## What the logs say

#725 collapsed ~230 per-file integration-test binaries into the single `it` target. The three steps still name the old binaries, so each one has been dying immediately:

```
error: no test target named `api_comparison_tests` in default-run packages
error: no test target named `spdi_tests` in default-run packages
error: no test target named `dbsnp_tests` in default-run packages
```

Masked by exactly the two mechanisms the issue describes, so the workflow has reported **success** every week since. The changeover is visible across a single week:

| scheduled run | `no test target` errors | `test result:` lines | conclusion |
|---|---|---|---|
| 2026-06-14 (`27483424504`) — before #725 | 0 | **3** (3 + 2 + 13 passed) | success |
| 2026-06-21 (`27888384383`) — after #725 | **3** | 0 | success |

#725 merged 2026-06-16, between them. Seven consecutive green runs (06-21 → 08-02) validated nothing.

## The fix

- **Point the steps at `--test it` with a name filter.** Restores **3 + 8 + 13 = 24** tests, all passing locally.
- **Drop `--ignored` from the SPDI step.** It had two `#[ignore]`d tests when the flag was written (the 06-14 log shows `2 passed; 6 filtered out`) and has none now, so the flag selects an empty set. This was a *second*, older silent-zero that predates #725 and would have survived a naive fix.
- **Remove `continue-on-error`** from the three test steps and run them under `set -euo pipefail`, so both the step status and the `tee` pipeline propagate. Neither change works without the other, which is why the issue is right to name them as independent.
- **`if: always()` on the report generation and upload**, so a red run still publishes what it found. Masking the tests to keep the artifact was the tradeoff the old shape made implicitly.
- **A new `.github/scripts/run_validation.sh`** that fails when the filter selects **no** tests.

That last one is the part neither of the issue's two proposals covers. An empty selection exits 0, so `pipefail` and the removed `continue-on-error` both let it through — and a suite that runs nothing looks exactly like a suite that passes, except faster, which reads as an improvement. Two instances of that had already happened here, which is the argument for a check rather than a third careful reading.

## Verification

Return codes, measured (`0` only where a real suite ran):

| invocation | tests run | exit |
|---|---|---|
| `api_comparison_tests --ignored` | 3 | 0 |
| `spdi_tests` | 8 | 0 |
| `dbsnp_tests` | 13 | 0 |
| `no_such_module_xyz` (filter matches nothing) | 0 | **1** |
| `spdi_tests --ignored` (the stale-flag case) | 0 | **1** |

And `pipefail` propagates a failing `cargo test` through `tee` — a command exiting 101 into `tee` yields 101, where the old shape yielded 0.

`actionlint` and `shellcheck` are clean, and the YAML parses.

## Scope note

The remaining `continue-on-error`s in this workflow are on the third-party fetch/download steps (LOVD, MaveDB, CIViC, the API fetches, the artifact downloads). Those are deliberate — a flaky external API should not redden the workflow — so they are left alone.

One thing I did **not** fix, because it is a separate question: if a fixture download fails, these tests skip rather than fail, which is a third way to be quietly vacuous. The new zero-selection guard does not catch it, since the tests are selected and *do* run. Worth its own issue if you want it closed.

### DIFF
--- a/.github/scripts/run_validation.sh
+++ b/.github/scripts/run_validation.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+# Run one external-validation suite so that every way it can go wrong is loud.
+#
+# Sourced by `.github/workflows/external-validation.yml`. Three failure modes
+# are in scope, and the workflow had been silently surviving all three (#1404):
+#
+#   1. the tests fail            — propagated, because `pipefail` is set and the
+#                                  step no longer carries `continue-on-error`
+#   2. the target does not exist — `cargo test` exits non-zero, which #725's
+#                                  binary consolidation made the live case
+#   3. the filter selects zero   — exits 0 with `running 0 tests`, so it is
+#      tests                       checked explicitly here
+#
+# The third is the one worth the script. A suite that runs nothing looks exactly
+# like a suite that passes, and it is *faster*, so it reads as an improvement.
+#
+# Usage: run_validation <module> <output-file> [extra cargo-test args...]
+run_validation() {
+    local module="$1"
+    local output="$2"
+    shift 2
+
+    set -euo pipefail
+
+    # `--test it`: every integration test lives in the single `it` binary since
+    # #725. `$module` is a filter on the test *name*, which is
+    # `<module>::<test>`, so it selects that module's tests and nothing else.
+    cargo test --test it "$module" -- --nocapture "$@" 2>&1 | tee "$output"
+
+    # `cargo test` prints one `running N tests` line per binary. Sum them, so a
+    # filter that matched nothing is an error rather than a fast pass.
+    local ran
+    ran=$(awk '/^running [0-9]+ test/ { total += $2 } END { print total + 0 }' "$output")
+
+    if [[ "$ran" -eq 0 ]]; then
+        echo "::error::${module}: the filter selected no tests, so this suite validated nothing." >&2
+        echo "Check that the module still exists under tests/it/ and that any" >&2
+        echo "--ignored flag still matches how its tests are marked." >&2
+        return 1
+    fi
+
+    echo "${module}: ran ${ran} test(s)."
+}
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

Issue #387, opened 2026-05-21
fix(normalize): 3'-direction canonicalization must clamp at CDS-end for c.-axis inputs


nh13:
## Summary

CDS-end mirror of #383, surfaced by `tests/fixtures/biocommons-normalize/baseline-failures/normalized.txt`. ferro 3'-canonicalises a c.-axis variant across the CDS↔3'UTR boundary when biocommons clamps at c.<cds_end>.

| Input | biocommons | ferro |
|---|---|---|
| `NM_212556.2:c.1400_1401insAC` | `c.1401delinsACA` | `c.1401_*1insCA` |

ferro's 3'-shift correctly advances the insertion one position (alt rotates `AC` → `CA`, position moves from `c.1400_1401` to `c.1401_*1`), but the resulting `_*1` end coordinate crosses from the CDS axis (`c.<N>`) into the 3'UTR axis (`c.*<N>`). Per HGVS §general's per-axis coordinate treatment, a CDS-interior input should not be silently re-described onto the 3'UTR axis. Biocommons clamps at `c.<cds_end>` and absorbs the boundary base into a `delins`.

## Spec basis

HGVS DNA recommendations (<https://hgvs-nomenclature.org/stable/recommendations/DNA/>):

- **3' rule** — *"The most 3' position possible is arbitrarily assigned to have been changed."* Applies to all axes.
- **Per-axis coordinate treatment** — c.-axis is c.-coordinates, not transcript-1. The CDS (`c.<N>`) and 3'UTR (`c.*<N>`) are different sub-axes. A canonicalisation rewrite that silently moves a CDS-interior input onto the 3'UTR axis violates the user's axis intent.
- **Edit-type priority** (sub > del > inv > dup > ins; delins residual) — at the boundary the only valid form is `c.<cds_end>delins<absorbed>` where the absorbed bases encode both the original alt and the CDS bases the 3'-shift would have walked through.

Unifying rule (mirror of #383's CDS-start clamp):

> A 3'-direction canonicalisation rewrite on a c.-axis variant may not land *strictly* on the 3'UTR axis. When the rewrite cannot stay anchored at `c.<cds_end>` or earlier, emit as `c.<cds_end>delins<…>` (absorbing the boundary base into the alt) for Insertion inputs, or restore the input form unchanged for Delins inputs whose canonicalisation would move them past the boundary.

## Scope

In `src/normalize/mod.rs::normalize_cds`, after `normalize_na_edit` returns:

[...]

PR #1417, opened 2026-08-05
fix(normalize): clamp a derived insertion at the CDS/3'UTR junction

### LINKED ISSUE #1398: Normalizer: a c.-axis allele normalizes to an insertion across the CDS/3'UTR boundary that is not a fixed point

nh13:
Closes #1398.

A `c.`-axis cis allele normalized to an insertion spanning the CDS/3'UTR junction, and re-normalizing that output did not return it:

```
in     c.[11_12dup;16_17insC]
once   c.18_*1insCAT
twice  c.18delinsTCAT   (3')
twice  c.16_17insATC    (5')
```

## Root cause

`boundary_delins_anchor` in the piece builder clamps a derived pure insertion only at the **representability** bounds — contig ends, transcript ends. Its doc says so explicitly, and gives the reason: the axis-change boundaries at `cds_start`/`cds_end` have a representable far side (`c.-1`, `c.*1`), so clamping there is a policy with carve-outs, and `normalize_cds` keeps it.

**That reasoning holds for a lone member and not for a derivation.** A single variant gets re-clamped by the per-member pipeline. The sequence-first derivation (#1392) is authoritative for the group, and nothing re-clamps its output — so the insertion kept the `c.<cds_end>_*1` spelling that every other path rewrites, and only the *second* normalize applied the policy. Hence the non-idempotency, and hence the two directions disagreeing about the second pass.

merge.rs already anticipates this exact situation:

> Do not restore it. If a terminal or tandem case regresses, **the fix belongs in the builder**; refusing here also refuses every merge the derivation gets right.

So the 3' half of the policy now lives in the builder, as `cds_end_delins_anchor`. The identity is `normalize_cds`'s and is shared rather than copied — inserting `A'` between `cds_end` and `cds_end + 1` is exactly "delete `ref[cds_end]`, insert `ref[cds_end] ++ A'`", true for any `A'`.

Two things deliberately **not** done, each with its reason in the code:

- **No `cds_start` mirror.** The axis identity is discontinuous at the 5' end (`c.-1` is `cds_start - 1`, the reason `region_sequence_delta` exists), so that side needs its own arithmetic and its own evidence.
- **`r.` is not claimed.** `normalize_cds`'s #387 clamp is gated on `AxisRegion::Cds`, so claiming `r.` in the derivation would recreate the same disagreement in the opposite direction.

Spanning duplications stay unclamped, per #401, and need no gate: `anchor_for_piece` tries `duplication_anchor` first, so a piece a `dup` describes never reaches the new function. That is the call order supplying the same positive-list-`Insertion` effect `normalize_cds` states explicitly.

## Verification

`tests/it/issue_1398_cds_utr3_insertion_idempotency.rs` pins both directions and both corpus sequences, and asserts two separate things: that the result is a fixed point, and that it does not re-axis onto the 3'UTR. The second is stated separately on purpose — a fix that merely made the spanning form *stable* would satisfy idempotency while still contradicting the clamp every other path applies.

Full suite green under all three oracles (`FERRO_ASSERT_IDEMPOTENT` / `REPARSE` / `IN_BOUNDS`): **9314 passed**. `cargo fmt --check` and `clippy --all-features --all-targets -D warnings` clean.

## What this does NOT fix — #1400's sweep stays `#[ignore]`d

I ran #1400's full-corpus transcript sweep against this change, which is the reason that sweep is ignored. **It still fails**, and it is worth being precise about why:

| | cases |
|---|---|
| baseline (no fix), full corpus | 32 sequence-changing |
| with this fix, full corpus | 32 sequence-changing — **identical set** |

So this fix neither regresses nor resolves them. They are a **separate, pre-existing defect class** that the idempotency panic was previously masking: with the panic gone the sweep runs to completion and reports them. They are *more* severe than #1398 — normalization changing what the variant means:

```
c.[11_12del;14del] -> c.[14del;*1_*2del]
   want TAATTTATTAAATATAT
   got  TAATTTATTAATAATAT
```

`11_12del` 3'-shifts through the `(AT)₅` tract into the 3'UTR, which is valid *in isolation* — but the sibling `14del` disrupts the tract, so the shift is not. Filed separately as #1418.

**#1400's `#[ignore]` therefore stays until that one lands**, and its comment should be re-pointed from #1398 to the new issue. I did not touch #1400 here to avoid editing another open PR's branch from this one.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2379,7 +2379,34 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
         SequenceEnds::INTERIOR
     };
 
-    let rebuilt = rebuild_members(&pieces, &variants[0], body, w_lo, &ref_bytes, ends)?;
+    // The CDS end on the member axis, for `cds_end_delins_anchor`. `None`
+    // wherever there is no CDS to change axis at — every genomic kind, and a
+    // non-coding transcript — which makes that clamp inert there.
+    //
+    // `CisKind::Cds` only: `r.` reaches the same junction, but `normalize_cds`'s
+    // #387 clamp is gated on `AxisRegion::Cds`, so claiming the `r.` case here
+    // would make the derivation *disagree* with the per-member pipeline in the
+    // opposite direction — the very fault this fixes. Left for its own evidence.
+    let cds_end_axis = if matches!(kind, CisKind::Cds) {
+        provider
+            .get_transcript(&accession)
+            .ok()
+            .and_then(|tx| tx.cds_end)
+            .and_then(|end| i64::try_from(end).ok())
+            .map(|end| end - frame.delta)
+    } else {
+        None
+    };
+
+    let rebuilt = rebuild_members(
+        &pieces,
+        &variants[0],
+        body,
+        w_lo,
+        &ref_bytes,
+        ends,
+        cds_end_axis,
+    )?;
     if rebuilt == variants {
         return None;
     }
@@ -4567,6 +4594,7 @@ fn rebuild_members(
     w_lo: i64,
     ref_bytes: &[u8],
     ends: SequenceEnds,
+    cds_end_axis: Option<i64>,
 ) -> Option<Vec<HgvsVariant>> {
     let mut members = Vec::with_capacity(pieces.len());
     // Window offset one past the last reference base the *previous* piece
@@ -4582,7 +4610,15 @@ fn rebuild_members(
             piece.ref_start >= previous_ref_end,
             "pieces must be in ascending offset order for the dup disjointness check"
         );
-        let anchor = anchor_for_piece(piece, body, w_lo, ref_bytes, ends, previous_ref_end)?;
+        let anchor = anchor_for_piece(
+            piece,
+            body,
+            w_lo,
+            ref_bytes,
+            ends,
+            previous_ref_end,
+            cds_end_axis,
+        )?;
         previous_ref_end = piece.ref_end;
         members.push(build_merged(template, anchor));
     }
@@ -4619,6 +4655,7 @@ fn anchor_for_piece(
     ref_bytes: &[u8],
     ends: SequenceEnds,
     previous_ref_end: usize,
+    cds_end_axis: Option<i64>,
 ) -> Option<Anchor> {
     let alt: Vec<Base> = piece
         .alt
@@ -4633,6 +4670,11 @@ fn anchor_for_piece(
             return Some(anchor);
         }
         if let Some(anchor) = boundary_delins_anchor(piece, &alt, body, w_lo, ref_bytes, ends) {
+            return Some(anchor);
+        }
+        if let Some(anchor) =
+            cds_end_delins_anchor(piece, &alt, body, w_lo, ref_bytes, cds_end_axis)
+        {
             return Some(anchor);
         }
     }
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
## Representation stability

Adding this per the new `CLAUDE.md` section and the 2026-08-05 relay, which postdate this PR.

**This PR moves a normalized output.** The shape:

```
c.<cds_end>_*1ins<A'>   ->   c.<cds_end>delins<ref[cds_end] ++ A'>
e.g.  c.18_*1insCAT      ->   c.18delinsTCAT
```

**Direction: toward the already-shipped form.** This is the tie-break the relay asks for, and here it is free. `c.<cds_end>delins…` is what `normalize_cds` has always produced for a directly-authored variant of this shape (#387, pinned in `five_prime_boundary_delins_unification.rs`); the spanning `c.<cds_end>_*1ins…` was only ever emitted by the sequence-first derivation, which landed in #1392 **this cycle** and has not been in a release. So the form this PR removes was never shipped, and the form it converges on is the one consumers already have.

That also means the movement is invisible release-to-release for anyone on v0.12.0 or earlier — it is churn introduced and removed within the same cycle. It does matter to anyone tracking `main`.

**Spanning duplications are deliberately untouched** (`c.9171_*1dup`), per #401 — those are the spec-canonical form and remain so.

**Blast radius:** full suite green at 9328 tests with no pinned expectation moved. I did not run a release-to-release corpus diff, on the grounds above: the form being removed postdates the last release.

Issue #1414, opened 2026-08-05
Normalizer: a conflicting cis allele skips the genomic-order sort even when its members were rewritten


nh13:
## What's wrong

Normalizing a cis allele that reports an overlap conflict skips the genomic-order sort, so the members come back in whatever order the pipeline left them. When the output no longer conflicts, a second pass sorts them — and `normalize` is not idempotent.

```
reference  CGCGCGCGCAATCGCGCG     18 bases; 10 = `A`, 11 = `A`, 12 = `T`

in     TEMPLATE:g.[11_12inv;11_12insAA]
once   TEMPLATE:g.[11_12=;10_11A[4]]     members out of genomic order
twice  TEMPLATE:g.[10_11A[4];11_12=]     sorted; this one is a fixed point
```

Two defects in one output, and the first is the more serious:

1. **The first output carries out-of-order members.** That is a direct violation of #1235's criterion 2 ("no normalized cis allele contains overlapping or out-of-order members"), on a shape its confluence proptest does not generate.
2. **Normalization is not idempotent**, so `FERRO_ASSERT_IDEMPOTENT` fires on this input. Any test that normalizes it reddens the `test-oracle` job.

## Mechanism

`normalize_allele` gates the sort on the absence of an overlap-conflict warning:

```rust
let has_overlap_conflict = all_warnings
    .iter()
    .any(|w| matches!(w, NormalizationWarning::OverlapConflict { .. }));
if is_cis && !allele.uncertain && !has_overlap_conflict && normalized.len() > 1 {
    sort_cis_members_by_genomic_order(&mut normalized);
}
```

The skip is deliberate and its reasoning is sound: an overlap-conflicting allele is emitted **verbatim** (#395), and reordering an allele you are handing back untouched would contradict that.

**The premise just is not true here.** This allele was not emitted verbatim — every member was rewritten. `11_12inv` cancelled to `11_12=` (the span is its own reverse complement) and `11_12insAA` respelled as `10_11A[4]`. The pipeline rewrote both members and then declined to order them, on the strength of a contract about *not* rewriting them.

So the gate should ask whether the allele was actually emitted verbatim, not whether a conflict was reported.

## Why the second pass differs

`g.[11_12=;10_11A[4]]` does not itself report an overlap conflict, so the sort runs and the members come back in genomic order. That is why the shape is non-idempotent rather than merely mis-ordered: the two passes take different branches of the same gate.

## Relationship to #1406 and #1411

[...]

PR #1423, opened 2026-08-05
fix(normalize): drop a cancelled identity member a sibling partly covers

### LINKED ISSUE #1416: Normalizer: a cancelled identity member survives beside a repeat that claims its bases, so the allele denotes nothing
## What's wrong

`normalize` returns a two-member allele whose members overlap, so it has **no well-defined resulting sequence** — ferro's own independent applier declines it. On `origin/main` at **d283a95d** (with #1392 merged):

```
reference  CGCGCGCGCAATCGCGCG      18 bases; 10 = `A`, 11 = `A`, 12 = `T`

in    TEMPLATE:g.[11_12inv;11_12insAA]
out   TEMPLATE:g.[11_12=;10_11A[4]]
```

Measured with `cis_apply_oracle::apply`, which declines when two members claim the same base:

| description | applies to |
|---|---|
| `g.[11_12=;10_11A[4]]` | **None** — overlapping |
| `g.[10_11A[4];11_12=]` | **None** — overlapping |
| `g.10_11A[4]` | `CGCGCGCGCAAAATCGCGCG` |

That third row is the whole variant. `11_12inv` over `AT` is its own reverse complement, so it cancels to `=`; `11_12insAA` grows the `A` tract at 10_11 from two copies to four. **`g.10_11A[4]` alone denotes exactly the input's bases**, and dropping the cancelled identity member would produce it.

This is a direct violation of #1235's criterion 2 — "no normalized cis allele contains overlapping or out-of-order members" — on a shape its confluence proptest does not generate (its indel model does not produce an inversion beside a co-located insertion).

## Mechanism

`drop_identity_members_covered_by_siblings` (#1297) exists to remove exactly this: a member that cancelled to `=` while a sibling grew over the bases it names. It does not fire here. The identity member claims `11_12`; the repeat's tract is `10_11` and it writes into position 11, so the two claim base 11 between them.

Note the repeat *appends* its copies at the tract's 3' end, so one can argue the base originally at 11 is still `A` and the identity is satisfied. The applier does not agree, and it is the component that decides whether a description denotes anything — its verdict is that these two members cannot both be applied. Whichever reading is preferred, the current output is one no consumer can splice.

## Severity

**High.** The output is malformed in the strongest sense available: it denotes no sequence at all. It is invisible today only because no committed test normalizes this shape.

- **#1414** — the same input, filed for the *ordering / idempotency* symptom (`g.[11_12=;10_11A[4]]` re-normalizes to `g.[10_11A[4];11_12=]`). That issue's analysis was corrected in a comment; this is the underlying defect, and it makes #1414's framing too generous — the first output is not merely mis-ordered, it is unappliable. Fixing this likely dissolves #1414 entirely, since a single-member output has no ordering question.
- **#1406 row 3** — the same input again, filed for the laundering symptom (input strict-rejected, output strict-accepted). If the identity member were dropped, the output would be `g.10_11A[4]`, which is *still* strict-accepted, so that row needs its own answer regardless.
- **#1297** (closed) — the pass that should cover this. Filed as a new issue rather than a comment there, since a closed issue reaches nobody.

## Reproducer

```rust
let mut p = MockProvider::new();
p.add_genomic_sequence("TEMPLATE", "CGCGCGCGCAATCGCGCG".to_string());
let out = Normalizer::new(p)
    .normalize(&parse_hgvs("TEMPLATE:g.[11_12inv;11_12insAA]").unwrap())
    .unwrap()
    .to_string();
// out == "TEMPLATE:g.[11_12=;10_11A[4]]", and cis_apply_oracle::apply returns None for it
```

A test that normalizes this input also fires `FERRO_ASSERT_IDEMPOTENT` (see #1414), so pin it with the apply oracle rather than with a re-normalization.

nh13:
Closes #1416.

`g.[11_12inv;11_12insAA]` over `CGCGCGCGCAATCGCGCG` normalized to `g.[11_12=;10_11A[4]]`, whose two members both claim base 11. The independent applier declines it in **either** member order, so the allele denoted no sequence at all.

## Root cause

`drop_identity_members_covered_by_siblings` (#1297) exists for exactly this shape — a member that cancelled to `=` while a sibling grew over the bases it names. It required the sibling's span to **contain** the identity's:

```rust
s.start <= span.start && s.end >= span.end
```

Here the repeat's tract is `10_11` and the identity claims `11_12`. They overlap on base 11, but neither contains the other, so nothing was dropped.

The rule the pass enforces is "two members must not claim one base" — its own doc says so, and cites the apply oracle as the arbiter. A **partial** overlap breaks that rule exactly as a containment does. So the test is now for overlap.

**Dropping stays sequence-preserving even where the identity reaches past the sibling.** An identity asserts only that its bases are unchanged, so removing it cannot change what the allele denotes: the bases outside the sibling are unchanged whether or not anything says so. What is lost is an assertion. What keeping it costs is a description no consumer can splice.

Output is now the single member `g.10_11A[4]`, which denotes exactly the input's bases (`11_12inv` over `AT` is its own reverse complement and cancels; `11_12insAA` grows the two-copy `A` tract to four).

## Verification

Pinned with the **apply oracle**, not by re-normalizing — the old output also fires `FERRO_ASSERT_IDEMPOTENT` (#1414), so a re-normalization test would report that ordering symptom instead of this defect. Two assertions, deliberately separate: that the output denotes *a* sequence, and that it denotes the *right* one — `is_some()` alone would accept any appliable answer, including one that fixed the overlap by changing the meaning.

Both were watched to fail against unfixed `main` before being trusted.

**A test-harness bug worth flagging**, since the same trap is available to anyone writing one of these: `cis_apply_oracle::apply` re-parses its argument with `parse_hgvs`, so it needs the **accession-qualified** string. My first version passed the bare `g.…` descriptor, which makes `apply` return `None` for *every* input — indistinguishable from the defect. It "reproduced" perfectly and would have gone on passing after the fix.

Full suite green under all three oracles (`FERRO_ASSERT_IDEMPOTENT` / `REPARSE` / `IN_BOUNDS`): **9328 passed**. `cargo fmt --check` and `clippy --all-features --all-targets -D warnings` clean.

## Relationship to #1414 — read before closing it

#1414 is the same input, filed for the ordering/idempotency symptom. With this fix that symptom is gone — the output is a single member, so there is no ordering question, and it is a fixed point:

```
once  = TEMPLATE:g.10_11A[4]
twice = TEMPLATE:g.10_11A[4]
```

**I did not mark #1414 closed, on purpose.** Its stated mechanism is untouched: `normalize_allele` still gates the genomic-order sort on `has_overlap_conflict` rather than on whether the allele was actually emitted verbatim, and #1414's own analysis is that the gate asks the wrong question. This PR removes the only known input that exposes it, not the gate. Whether that is worth keeping open is a judgement call I would rather leave to you than quietly resolve by deleting the reproducer.

## Not addressed

**#1406 row 3**, the same input again, filed for the laundering symptom (input strict-rejected, output strict-accepted). #1416 already notes it needs its own answer regardless: the fixed output `g.10_11A[4]` is *still* strict-accepted.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -6579,9 +6579,27 @@ fn translate_junction_member<P: ReferenceProvider>(
 ///   the repeat alone denotes the input; the `265=` residue overlaps it
 /// ```
 ///
-/// Only the identity member is dropped, and only when a sibling's span *covers*
+/// Only the identity member is dropped, and only when a sibling's span *overlaps*
 /// it under [`blocks_sibling_shift`] — a sibling that merely adds sequence at a
 /// junction inside it contradicts nothing.
+///
+/// **Overlap, not containment** (#1416). The rule this enforces is "two members
+/// must not claim one base", and a partial overlap breaks it exactly as a
+/// containment does — the applier declines both alike:
+///
+/// ```text
+/// g.[11_12inv;11_12insAA]  ->  g.[11_12=;10_11A[4]]  over `CGCGCGCGCAATCGCGCG`
+///   the inversion of `AT` cancels to `=`; the repeat grows the `A` tract at
+///   10_11 to four copies and so claims base 11, which the identity also names.
+///   The repeat's span (10_11) does not *contain* the identity's (11_12), so a
+///   containment test left the pair overlapping and the allele denoted nothing.
+/// ```
+///
+/// Dropping is sequence-preserving even where the identity reaches past the
+/// sibling: an identity asserts only that its bases are unchanged, so removing
+/// it can never change what the allele denotes — the bases outside the sibling
+/// are unchanged whether or not anything says so. What is lost is an assertion,
+/// and keeping it costs a description no consumer can splice.
 ///
 /// `blocks_sibling_shift` rather than `claims_reference_bases`, because a
 /// duplication covers the positions it copies without consuming them (#1321):
@@ -6630,8 +6648,22 @@ pub(crate) fn drop_identity_members_covered_by_siblings(
                     s.blocks_shift
                         && s.region == span.region
                         && s.accession == span.accession
-                        && s.start <= span.start
-                        && s.end >= span.end
+                        // Both spans forward. A reversed `<high>_<low>` range is
+                        // SVD-WG006's circular deletion/duplication form, and
+                        // `cis_axis_parts` hands the endpoints over raw, so one
+                        // can reach here with `start > end`. This pass has no
+                        // wraparound semantics: the interval test below reads
+                        // such a span as an ordinary one, and widening
+                        // containment to overlap made that newly *matter* — a
+                        // reversed sibling against a forward identity now
+                        // satisfies the test where containment refused it.
+                        // Declining is the honest answer, and it is also the
+                        // conservative one: the identity is kept, which is the
+                        // behaviour that predates this rule.
+                        && s.start <= s.end
+                        && span.start <= span.end
+                        && s.start <= span.end
+                        && s.end >= span.start
                 })
             })
         })
--- a/tests/it/issue_1416_cancelled_identity_beside_repeat.rs
+++ b/tests/it/issue_1416_cancelled_identity_beside_repeat.rs
@@ -1,0 +1,107 @@
+//! A cancelled identity member must not survive beside a sibling that claims
+//! its bases (#1416).
+//!
+//! `g.[11_12inv;11_12insAA]` normalized to `g.[11_12=;10_11A[4]]`, whose two
+//! members both claim base 11 — so the allele denotes **no sequence at all**,
+//! and the independent applier declines it in either member order. The whole
+//! variant is the repeat: `11_12inv` over `AT` is its own reverse complement and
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
## Representation stability

Adding this per the new `CLAUDE.md` section and the 2026-08-05 relay, which postdate this PR.

**This PR moves a normalized output**, in exactly one shape:

```
g.[11_12=;10_11A[4]]   ->   g.10_11A[4]
```

**It is a pure bugfix, not churn.** The old form is not a representation a consumer could have keyed on: its two members both claim base 11, so it denotes *no* sequence and the apply oracle declines it in either member order. There is no "already-shipped form" to break a tie toward — the shipped form was unappliable. The new one denotes the input's bases.

**Blast radius:** the full suite (9328 tests, including every pinned normalization expectation and the three exhaustive cis sweeps) passes unchanged, so no other pinned output moved. I did not run a separate release-to-release corpus diff for this one; the shape requires a cancelled identity member partially overlapped by a repeat, which the sweeps only reach through the `#[ignore]`d transcript axes.

PR #1425, opened 2026-08-05
fix(normalize): bound a sibling-crossing shift across a region boundary

### LINKED ISSUE #1418: Normalizer: a c.-axis cis member 3'-shifts out of the CDS into the 3'UTR past a sibling, changing the sequence
## What's wrong

On the `c.` axis, a two-member cis allele can normalize to a description that **denotes a different sequence than the input**. Normalization is changing what the variant means, which `merge.rs`'s own runtime refusal calls "the worst failure this crate can have".

```
transcript  TAATTTATTAATATATATAT      c. axis, CDS = transcript 1..=18
                                      so c.18 is the last CDS base, c.*1 the first 3'UTR base

in    NM_TEST.1:c.[11_12del;14del]     [ThreePrime]
out   NM_TEST.1:c.[14del;*1_*2del]
      want TAATTTATTAAATATAT
      got  TAATTTATTAATAATAT
```

## Why it happens (starting point, not a diagnosis)

Transcript positions 11..20 are `ATATATATAT`, a clean `(AT)₅` tract. In isolation `c.11_12del` really is equivalent to deleting the last `AT` of that tract, so 3'-shifting it out to `c.*1_*2` is a sound move — the tract is uniform and the deletion is 2 bases.

It stops being sound once the sibling is there. `c.14del` removes a base from the middle of that same tract, so after it the run is no longer uniform and the two members no longer commute with the shift. The shift appears to be decided per member against the unedited reference rather than against the haplotype the siblings produce.

That is the same family as the sibling-crossing-shift work (#1254, #1287, #1290), but on the **transcript** axis and running *out of the CDS into the 3'UTR*, which the genomic sweeps cannot reach because there is no CDS boundary on `g.`.

## Scope

32 cases in the full transcript-axis corpus, all `[ThreePrime]`, all of the shape "a 5' member shifts to `*N_*M`" — both `del` and `dup` siblings appear:

```
c.[11_12del;14del]      -> c.[14del;*1_*2del]
c.[11_12del;14T>A]      -> c.[14T>A;*1_*2del]
c.[11_12del;14dup]      -> c.[14dup;*1_*2del]
c.[11_12del;14_15insC]  -> c.[14_15insC;*1_*2del]
c.[11_12dup;14del]      -> c.[14del;*1_*2dup]
…
```

The full list is in the sweep's own failure output.

## How it was found, and what it blocks

By the transcript-axis sweep added in #1400, at the **full** corpus only. That sweep is `#[ignore]`d pending #1398, and this defect was hidden behind #1398's idempotency panic: with the panic fixed (#1417) the sweep runs to completion and reports these instead.

Confirmed **pre-existing and independent of #1417** — the failing set is byte-identical with and without that fix (32 cases both ways), so it neither causes nor resolves them.

**This is what actually blocks un-ignoring #1400's sweep.** Its `#[ignore]` comment names #1398; once #1417 lands that pointer should move here.

## Reproducing

```rust
let provider = SyntheticBuilder::cds("TAATTTATTAATATATATAT", 1, 18, Strand::Plus).build();
let normalizer = Normalizer::with_config(
    provider,
    NormalizeConfig::default()
        .with_direction(ShuffleDirection::ThreePrime)
        .allow_crossing_boundaries(),
);
// c.[11_12del;14del] -> c.[14del;*1_*2del], which applies to a different sequence
```

Or the whole class at once, on #1400's branch with the `#[ignore]` removed:

```sh
FERRO_SWEEP_SEEDS=full cargo nextest run --features dev \
```

nh13:
Closes #1418.

A `c.`-axis cis member could 3'-shift out of the CDS, past a sibling, into the 3'UTR — changing what the allele denotes:

```
c.[11_12del;14del]  ->  c.[14del;*1_*2del]
  input denotes  TAATTTATTAAATATAT
  output denotes TAATTTATTAATAATAT
```

## Root cause

`clamp_sibling_crossing_shifts` is the pass that stops a member leapfrogging a sibling, and it skipped this case outright:

```rust
if !b.blocks_shift || !a.blocks_shift || b.region != a.region { continue; }
```

`11_12del` starts in `Region::Cds` and lands in `Region::ThreePrimeUtr`, so **a shift that changed region was exempted from the sibling-crossing invariant entirely** — and crossed the `14del` unchecked. (Confirmed by instrumenting the guard: `before=(Cds, 11, 12)`, `after=(ThreePrimeUtr, 1, 2)`.)

The guard was avoiding a real problem: region axes cannot be compared position-by-position, since `c.12` and `c.*1` are adjacent bases with unrelated numbers. So the pass now runs on the **sequence** axis, converting each span once through `region_sequence_delta` — the same conversion the in-bounds oracle and the payload reads already use. Within one region that is a constant offset and changes nothing, which is why the arithmetic is otherwise untouched. Sibling matching drops its region filter for the same reason: filtering siblings by region hid exactly the ones a boundary-crossing member sweeps over.

Applying the pull is where the boundary still bites. Both `translate_member` and `translate_junction_member` add to the axis number and then verify the result landed in the original region — so a `c.*1` pulled 5' would be written `c.*-6` and reverted, silently leaving the uncorrected output. A cross-region pull therefore **restores the pre-shift member**: a stronger pull than the bound requires, so always inside the swept window, and sequence-correct by construction. The cost is canonicalisation, not correctness — a boundary-aware translate would reach a more-3' position, and that is deliberately left out of scope as a change to shared machinery every clamp uses.

## Verification

Full 829,440-case transcript-axis sweep (#1400), at `FERRO_SWEEP_SEEDS=full`: **32 sequence-changing normalizations before, 0 after.**

Full suite green under all three oracles: **9328 passed**. `fmt`/`clippy` clean.

## Representation stability — this PR moves shipped forms

Per `CLAUDE.md` and the 2026-08-05 relay, stating this explicitly rather than as "fixes non-confluence".

Measured by dumping normalized output over a 14,080-row slice of the transcript-axis corpus (two directions × dup/del first members × del/dup/ins second members) on `main` and on this branch:

| | rows |
|---|---|
| total | 14,080 |
| **moved** | **9 (0.064%)** |

And the part that matters — I checked each moved row against the apply oracle:

**All 9 moved rows were already correct on `main`.** So within this slice the change is **pure churn, not bugfix**: 9 outputs that denoted the right bases now denote them with a different string. The 32 genuinely-wrong outputs live in the wider sweep corpus, outside this slice's shape set.

Direction of movement — away from a two-member form with a 3'UTR member, toward a single CDS-side member:

```
c.[11_12dup;16del]  :  c.[15_16dup;*1del]  ->  c.16_17insA
c.[11_12dup;16dup]  :  c.[15_16dup;*1dup]  ->  c.16_17insATT
c.[16_17dup;18dup]  :  c.[17_18dup;*2dup]  ->  c.18_*1insATT
```

Both spellings denote identical bases in every case. The clamp pulls the member back inside the CDS, and the collapse pass then merges the now-adjacent pair.

**Two caveats I do not want to paper over:**

- **This is `main`-to-branch, not release-to-release.** The relay asks for the latter. I could not run it: the harness (`sweep_sequences` / `sweep_seeds`) postdates v0.12.0, so the corpus does not exist on that tag. Someone comparing against the shipped release should expect this movement *plus* whatever #1392, #1380 and #1393 already moved this cycle.
- **The slice is not the corpus.** 14,080 rows over one shape family, chosen to be cheap to dump on both sides — not the 829,440-case sweep, and not a clinical corpus. Treat 0.064% as an order of magnitude, not a bound.

Given the relay's "break spec-permitted ties toward the already-shipped form", these 9 are a cost, not a benefit. They are collateral to bounding the 32 wrong ones and I do not see how to fix the latter without the former — but they should be batched into the same release as the other normalizer changes, not dribbled out separately.

## #1400's sweep stays `#[ignore]`d — it needs a third fix

I checked whether this unblocks the un-ignore. It does not, and the reason is worth recording since the ignore comment currently names only #1398:

| state | full sweep, no oracles | full sweep, with oracles (as CI runs it) |
|---|---|---|
| `main` | FAIL — 32 sequence-changing | FAIL — #1398 idempotency panic |
| + this PR | **PASS** | FAIL — #1398 idempotency panic |
| + this PR + #1417 | **PASS** | **FAIL — a third defect** |

With both fixes applied the sweep still fires `FERRO_ASSERT_IDEMPOTENT`, on a shape neither addresses:

```
c.18_*1insACT  <->  c.*1_*2insCTA
```

Filed as #1426. So the `#[ignore]` needs to survive until all three land, and its comment should then name whichever is last.

---

## ⚠️ Representation change

Per the repository's representation-stability rule, this PR moves a normalized output and the movement is named here rather than left to be found downstream.

| input | was (shipped) | now |
|---|---|---|
| `NM_TEST.1:c.[11_12dup;16_17insC]` on `CORE_SIBLING`, 3' direction | `c.[16_17insC;*1_*2dup]` | `c.18delinsTCAT` |

The old form is the defect: the `dup` had shifted across the CDS/3'UTR boundary **and past its own sibling**, denoting different bases. The new form keeps both members inside the CDS and converges on the same answer the no-sibling core already gave, so this narrows two spellings onto one rather than splitting them.

Caught by `issue_1398_cds_utr3_insertion_idempotency::the_clamped_answers_are_unchanged`, a value pin added in #1417 for exactly this purpose — that test asserted the property "does not span the junction", which the old output satisfied while still being wrong. It is re-blessed here with the reason recorded inline.

No other pinned output moves: 9339/9339 with all three oracles.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -5081,6 +5081,22 @@ fn blocks_sibling_shift(edit: &NaEdit) -> bool {
 /// live list keeps the pass independent of member order.
 ///
 /// [`ShuffleDirection::FivePrime`]: crate::normalize::ShuffleDirection
+/// One sibling's bounds, on the **sequence** axis rather than its own region's.
+///
+/// [`clamp_sibling_crossing_shifts`] compares a moving member against every
+/// sibling, and the two need not share a region — a member shifting out of the
+/// CDS is bounded by siblings still inside it (#1418). Region axes are not
+/// comparable position-by-position (`c.12` and `c.*1` are adjacent bases with
+/// unrelated numbers), so each span is converted once, here, via
+/// [`region_sequence_delta`], and the pass then does ordinary arithmetic.
+struct SiblingBound {
+    claims_bases: bool,
+    blocks_shift: bool,
+    start: i64,
+    end: i64,
+    junction: Option<i64>,
+}
+
 pub(crate) fn clamp_sibling_crossing_shifts<P: ReferenceProvider>(
     before: &[HgvsVariant],
     after: &mut [HgvsVariant],
@@ -5104,11 +5120,34 @@ pub(crate) fn clamp_sibling_crossing_shifts<P: ReferenceProvider>(
         let (Some(b), Some(a)) = (pre[i].as_ref(), post[i].as_ref()) else {
             continue;
         };
-        if !b.blocks_shift || !a.blocks_shift || b.region != a.region {
+        if !b.blocks_shift || !a.blocks_shift {
             continue;
         }
+        // A shift that changes region is still a shift, and still has to respect
+        // siblings (#1418). It used to be skipped here, which let a member cross
+        // the CDS/3'UTR boundary — and every sibling on the way — unchecked:
+        // `c.[11_12del;14del]` emitted `c.[14del;*1_*2del]`, straight past the
+        // `14del`, denoting different bases.
+        //
+        // Regions cannot be compared position-by-position: `c.12` and `c.*1` are
+        // adjacent bases with unrelated numbers. So the whole pass runs on the
+        // **sequence** axis, via the same `region_sequence_delta` conversion the
+        // in-bounds oracle and the payload reads use. Within one region that is a
+        // constant offset and changes nothing, which is why the arithmetic below
+        // is untouched; across regions it is what makes the comparison mean
+        // anything at all.
+        //
+        // A region with no conversion — an `n.` position outside the transcript —
+        // is skipped, the same conservative answer its sibling reads give.
+        let seq = |s: &MemberSpan| -> Option<(i64, i64)> {
+            let delta = region_sequence_delta(s.region, &s.provider_key, provider)?;
+            Some((s.start.checked_add(delta)?, s.end.checked_add(delta)?))
+        };
+        let (Some((b_start, b_end)), Some((a_start, a_end))) = (seq(b), seq(a)) else {
+            continue;
+        };
         // A member that did not move cannot have crossed anything.
-        let delta = a.start - b.start;
+        let delta = a_start - b_start;
         if delta == 0 {
             continue;
         }
@@ -5128,8 +5167,8 @@ pub(crate) fn clamp_sibling_crossing_shifts<P: ReferenceProvider>(
         // a duplication moves its span and its junction together, so bounding
         // the span mis-places the copy — measured on the 5' sweep, that turns
         // correct outputs into silently wrong ones (#1266, #1279).
-        let translated = a.end - b.end == delta;
-        let shrank = a.end - a.start < b.end - b.start;
+        let translated = a_end - b_end == delta;
+        let shrank = a_end - a_start < b_end - b_start;
         if !translated && !shrank {
             continue;
         }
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #1432, opened 2026-08-05
fix(normalize): shift derived pieces in the direction that was requested

### LINKED ISSUE #1429: Normalizer: a derived c.-axis member is emitted without the 5' shuffle, so the first pass is not a fixed point
## What's wrong

A normalized `c.`-axis output is not a fixed point: the first pass returns a 3'UTR insertion that the second pass shifts.

```
first   NM_TEST.1:c.*1_*2insCTA
second  NM_TEST.1:c.17_18insTAC
```

Fires `FERRO_ASSERT_IDEMPOTENT` as a panic.

## Why this is a distinct defect, not a repeat of #1426

With PR #1428 applied, normalizing `c.*1_*2insCTA` as an **input** gives `c.17_18insTAC` in a single pass — verified directly. So the shuffle is not what is wrong here. Something upstream is producing `c.*1_*2insCTA` as an *output* without that shuffle ever running over it, and the second normalize is simply the first time it does.

That is the same family as #1398: the sequence-first derivation is authoritative for the group and emits a member the per-member pipeline would have moved, with nothing re-running the latter over the former's result. #1398 was the CDS-end clamp missing from the piece builder; this looks like the 5'-shuffle bound missing from the same place.

## How it was found

By #1400's full-corpus transcript-axis sweep, run with all three oracles and with **four** fixes applied — #1417 (#1398), #1425 (#1418), #1427 and #1428 (#1426's two halves). It is the fifth distinct defect behind that sweep's single `#[ignore]`, and each one was masking the next:

| state | full sweep, no oracles | with oracles |
|---|---|---|
| `main` | FAIL — 32 sequence-changing | FAIL — #1398 panic |
| + #1425 | PASS | FAIL — #1398 panic |
| + #1425 + #1417 | PASS | FAIL — #1426 3' half |
| + #1425 + #1417 + #1427 | PASS | FAIL — #1426 5' half |
| + all four | PASS | **FAIL — this** |

Worth stating plainly: the no-oracle column has been green since #1425. Every defect after that is an **idempotency** defect — the outputs denote the right bases, they just are not stable — so the sequence-correctness half of the sweep is already earning its keep and only the fixed-point half remains.

## Reproducing

On a branch carrying all four:

```sh
FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1 FERRO_SWEEP_SEEDS=full \
```

I have not narrowed it to a minimal two-member input — the panic reports the normalized pair, not the case that produced it. Given the pattern, the first thing to try is a cis allele over `TAATTTATTAATATATATAT` (CDS `1..=18`) whose members merge to a single junction insertion under the 5' direction.

## Why it matters

It is the last known thing between the repo and a live 829,440-case guard on the transcript axes — the axis family that has produced five real defects in three days, precisely because nothing swept it.

nh13:
Closes #1429. **This is the last of the five defects behind #1400's `#[ignore]`** — see the ladder at the bottom.

## Root cause

`canonicalize_from_sequence` never received the caller's shuffle direction, and its piece shift hardcoded it:

```rust
fn shift_pieces_three_prime(pieces: &mut [Piece], ref_bytes: &[u8]) { …
    shuffle(…, ShuffleDirection::ThreePrime);
```

So the sequence-first derivation shifted 3' whatever was asked for, and a 5' request got a 3'-shifted merged allele:

```
c.[1dup;2dup]  [5']  ->  c.8_9dup      <- the 3' answer, seven bases away
```

The per-member pipeline then moved it on the next pass, which is how it surfaced — as the sweep's non-idempotency `c.[16dup;18_19insC] -> c.*1_*2insCT -> c.18_*1insTC`.

**The idempotency failure is the lesser half.** The first-pass answer was simply not the answer the caller asked for; a derivation that *stably* returned the 3' form under a 5' request would be equally wrong and pass every idempotency check. That is why the tests assert the value, not just the fixed point.

The fix threads the direction through and renames the function. The payload rotation mirrors — moving an insertion point 3' by `k` rotates the payload left by `k`, moving it 5' by `k` rotates it right by the same `k`, since the rotation tracks phase against the reference and the point moved the other way. Debug assertions stay per-direction rather than collapsing to one `abs`, so a `shuffle` that moved a piece the wrong way is still caught.

## Verification

Full suite green under all three oracles on this branch alone: **9338 passed**. `fmt`/`clippy` clean.

The regression test deliberately uses `c.[1dup;2dup]`, nine bases inside the CDS, **not** the sweep's input. The sweep case lands on the CDS/3'UTR junction, so its idempotency also depends on #1427 and #1428, and a guard here would go red for reasons that are not this change. The chosen case is decided purely by which way the derivation shifts, and the two directions land seven bases apart.

## Representation stability

Dumped normalized output over 24,208 corpus rows on `main` and on this branch:

| | rows |
|---|---|
| total | 24,208 |
| **moved** | **99 (0.41%)** |
| moved under the default 3' direction | **0** |

All 99 are 5'-direction only — the default 3' shuffle is byte-identical, because it passed `ThreePrime` before and still does. Movement is of the form:

```
[5'] c.[1dup;16dup] :  c.[1dup;*1dup]  ->  c.[1dup;16dup]
```

The member moves *out of the 3'UTR back into the CDS* — i.e. it is now genuinely 5'-shifted rather than 3'-shifted. Every one of the 99 is a case that was previously answering a 5' request with a 3' answer, so this is the bugfix direction rather than churn: nobody asking for a 5' shuffle was getting a form worth preserving.

Same standing caveats: `main`-to-branch, not release-to-release (the harness postdates v0.12.0); 24,208 rows over these shape families is an order of magnitude, not a bound.

## The ladder — #1400's sweep now passes

Measured at `FERRO_SWEEP_SEEDS=full` with all three oracles, adding one fix at a time:

| state | no oracles | with oracles |
|---|---|---|
| `main` | FAIL — 32 sequence-changing | FAIL — #1398 |
| + #1425 | **PASS** | FAIL — #1398 |
| + #1417 | PASS | FAIL — #1426 3' |
| + #1427 | PASS | FAIL — #1426 5' |
| + #1428 | PASS | FAIL — #1429 |
| **+ this** | **PASS** | **PASS** |

829,440 cases, 338s. So once #1417, #1425, #1427, #1428 and this are all in, the `#[ignore]` on `no_two_member_transcript_axis_allele_normalizes_to_a_different_sequence` can come off — its comment still names only #1398 and should be deleted with it.

**One cost to flag before that happens:** at the full corpus this sweep takes **338s**, and CI's `sweeps` job is already the slowest in the matrix. Un-ignoring adds that. Worth deciding whether it belongs in `sweeps` as-is or wants its own seed profile.

## This PR now removes the `#[ignore]` — the ladder is complete

Rebased onto `main` at `bf4b983a`, so #1417, #1425, #1427 and #1428 are all in the base. With this fix on top, #1400's sweep **passes at the full corpus with all three oracles**:

```
FERRO_SWEEP_SEEDS=full FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1 \
  cargo nextest run --features dev -E 'test(no_two_member_transcript_axis_allele)'

PASS [ 279.850s] no_two_member_transcript_axis_allele_normalizes_to_a_different_sequence
```

**So the `#[ignore]` comes off here.** That was worth catching: five PRs each said "the ignore stays until the next fix lands," and this one — the fifth and last — did not remove it either. Left as it was, an **829,440-case guard** would have stayed dark after every defect behind it was fixed, which is the failure mode where lost coverage looks exactly like a pass.

The stale rationale is rewritten to record the ladder rather than name #1398 as a pending blocker: #1398 (#1417) → #1418 (#1425) → #1426 3' (#1427) → #1426 5' (#1428) → #1429 (this). Each was invisible until its predecessor was fixed, which is the case for having committed the sweep `#[ignore]`d instead of narrowing its position range.

### Where it runs, and what that costs

`cis_junction_crossing_shift` is inside `SWEEP_FILTER`, which `test` and `test-oracle` **negate** and `sweeps` selects **on**. So un-ignoring adds it to the `sweeps` job at the full corpus (~280s) and to a local `cargo nextest run` at the 4-seed prefix; it does not widen the sharded test jobs.

Checked the partition invariant `ci.yml` asks for — "the absolute counts drift as tests are added, the sum must not":

| selection | count |
|---|---|
| `not test(proptest)` | 5350 |
| `and not (SWEEP_FILTER)` | 5310 |
| `SWEEP_FILTER` alone | 40 |

5310 + 40 == 5350, so nothing falls between the jobs and nothing runs twice. (The absolutes differ from the comment's 8908/8878/30 — drift, exactly as it anticipates. Removing `#[ignore]` does not move them at all, since an ignored test is still listed.)

### Verification

Suite **9352 passed** under all three oracles on the rebased base — with `skipped` down 147 → 146, which is the un-ignored test now running. `fmt`/`clippy` clean.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2052,6 +2052,7 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
     variants: &[HgvsVariant],
     phase: AllelePhase,
     provider: &P,
+    direction: ShuffleDirection,
 ) -> Option<Vec<HgvsVariant>> {
     if variants.is_empty() || (variants.len() > 1 && phase != AllelePhase::Cis) {
         return None;
@@ -2223,7 +2224,7 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
         piece.ref_start += lo;
         piece.ref_end += lo;
     }
-    shift_pieces_three_prime(&mut pieces, &ref_bytes);
+    shift_pieces(&mut pieces, &ref_bytes, direction);
     coalesce_adjacent_pieces(&mut pieces);
     // Partitioning decides where the members are; this decides how wide each
     // one is spelled, and the two are not the same question — see
@@ -4278,7 +4279,8 @@ fn pieces_from_columns(columns: &[Column], reference: &[u8], result: &[u8]) -> V
     pieces
 }
 
-/// 3'-shift each length-changing piece, clamped so it cannot reach a sibling.
+/// Shift each length-changing piece in `direction`, clamped so it cannot reach
+/// a sibling.
 ///
 /// The clamp is the #1234 fix. The 3' rule (`general.md:41`) is stated per
 /// description with no allele-awareness, and the only text that ever addressed
@@ -4286,7 +4288,21 @@ fn pieces_from_columns(columns: &[Column], reference: &[u8], result: &[u8]) -> V
 /// deletion shifts over a downstream substitution and the two members overlap —
 /// malformed, and denoting a different sequence. Substitutions are anchored and
 /// never shift.
-fn shift_pieces_three_prime(pieces: &mut [Piece], ref_bytes: &[u8]) {
+///
+/// **`direction` is a parameter, not a constant (#1429).** This was hardcoded to
+/// `ThreePrime`, and `canonicalize_from_sequence` never received the caller's
+/// direction at all — so under a 5' shuffle the derivation handed back a
+/// *3'-shifted* allele, which the per-member pipeline then moved on the next
+/// pass:
+///
+/// ```text
+/// c.[16dup;18_19insC]  ->  c.*1_*2insCT  ->  c.18_*1insTC
+/// ```
+///
+/// Two passes, two directions. The 5' answer was also simply wrong on the first
+/// pass, independently of the idempotency: a caller asking for a 5' shuffle got
+/// a 3'-shifted merged allele.
+fn shift_pieces(pieces: &mut [Piece], ref_bytes: &[u8], direction: ShuffleDirection) {
     for i in 0..pieces.len() {
         if !pieces[i].is_pure_indel() {
             continue;
@@ -4302,7 +4318,7 @@ fn shift_pieces_three_prime(pieces: &mut [Piece], ref_bytes: &[u8]) {
             pieces[i].ref_start as u64,
             pieces[i].ref_end as u64,
             &bounds,
-            ShuffleDirection::ThreePrime,
+            direction,
         );
         let old_start = pieces[i].ref_start;
         let new_start = shuffled.start as usize;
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

PR #1433, opened 2026-08-05
ci: surface representation changes on the release PR and changelog

### LINKED ISSUE #1424: Surface representation changes on the release-plz PR and in the changelog
## Problem

The layout's `CLAUDE.md` now states that **representation stability is a shipped guarantee**: a downstream consumer's keys read counts on the normalized HGVS string, so what matters to them is not which representation ferro picks but whether the pick *moves*. A canonical form that churns between releases silently re-buckets every stored library design.

That rule asks for two things this repo currently has no mechanism for:

1. every PR that moves a normalized output says so and quantifies it;
2. representation changes are **batched into a single release** rather than dribbled across several.

Point 2 makes the release PR the natural place for a consolidated summary — but the release PR is generated by release-plz, and today its body is only the `cargo-semver-checks` API-breakage dump. A reader of #1360 cannot tell that, say, #1392 made the sequence-first derivation authoritative for cis alleles, or that #1380 stopped emitting the cancelled-member `=` residue — both of which move normalized output strings.

Concretely, ten commits merged in the current cycle touch output-producing code under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/`: #1363, #1365, #1367, #1368, #1373, #1377, #1380, #1385, #1392, #1393. Several certainly move shipped forms. None is identifiable as such from the release PR.

## release-plz does support this

Verified against the config reference for the pinned action (`release-plz/action@2eb1d8b`, v0.5.131). Two composable mechanisms:

**A. `pr_body` — a Tera 2 template for the release PR body.** Lives in `[workspace]` (overridable per `[[package]]`). It can access `releases`, and per release `title`, `package`, `changelog`, `previous_version`, `next_version`, `semver_check` and `breaking_changes`. Siblings: `pr_name`, `pr_labels`, `pr_draft`, `pr_branch_prefix`.

This gets us a **standing, unmissable section** on every release PR — but it cannot compute the representation diff by itself, since the template only sees release metadata.

**B. A commit trailer plus a changelog group — the automated half.** release-plz generates the changelog through git-cliff, configured via the `[changelog]` section (note: the older top-level `changelog_config` path field is deprecated). git-cliff's `commit_parsers` can match on a commit's **`footer`** — the documented example is `{ footer = "^changelog: ?ignore", skip = true }`.

So a trailer on the squash commit, e.g.

```
Representation-Change: g.[24dup;24C>G] now merges to g.23_24insG instead of being
preserved and strict-rejected. ~40 corpus rows; previously-rejected inputs only, so
no stored string changes.
```

can be matched by a parser that assigns it a dedicated group, giving a **⚠️ Representation changes** section in `CHANGELOG.md`. Because `release.changelog` is exposed to `pr_body`, that section then surfaces on the release PR automatically.

This matters here specifically because `CHANGELOG.md` is release-plz-generated — a hand-written entry gets overwritten — so the trailer route is the only one that survives.

## Proposal

1. Add a `[changelog]` `commit_parsers` entry grouping commits whose footer matches `^Representation-Change:` into a **⚠️ Representation changes** section, ordered above the ordinary feature/fix groups.
2. Add a `pr_body` template to `release-plz.toml` that keeps the current content and prepends a representation-stability block: the ⚠️ section when non-empty, and when empty an explicit *"No representation changes declared in this cycle"* line plus a reviewer checklist item pointing at the rule. An empty section that is *stated* is far better than an absent one, because it makes an undeclared change visible as a lie rather than as silence.
3. Document the trailer in `CONTRIBUTING`/`CLAUDE.md` next to the existing conventional-commit guidance, with the wording convention: which forms moved, in which direction, roughly how many corpus rows, and — the cheap/expensive distinction — whether the affected inputs were previously *rejected* (no stored string, free) or previously *accepted* (a real migration).

## Open questions

- Should an undeclared representation change be **enforceable** rather than advisory? A CI job could diff normalized output over a fixed corpus between `HEAD` and the previous release tag and fail when the diff is non-empty and no `Representation-Change:` trailer is present in the range. That is the only version that actually holds the line, but it needs a corpus that is cheap enough to run per-PR — worth costing separately rather than bundling here.
- Whether to backfill the trailer for the ten commits already merged this cycle, or to write the v0.13.0 summary by hand once and start the mechanism from v0.14.0.

nh13:
Closes #1424.

Implements the **advisory** half of the issue's proposal — declaration, not enforcement — plus the wording convention. Reasoning for the two open questions is at the bottom.

## What changes

**1. `commit_parsers` group.** A commit whose footer matches `^Representation-Change:` lands in its own **Representation changes** section, ordered above the ordinary groups.

The six parsers below it reproduce release-plz's defaults, because defining `commit_parsers` at all **replaces them wholesale** — omitting them would collapse Added/Fixed into one bucket. That is the main risk in this PR, and it did in fact bite: the first version of this list omitted `^removed`, so a `removed:` commit would have landed in `Other`. Fixed, and re-verified (below).

**2. `pr_body` template.** Prepends a representation-stability block and a reviewer checklist, then renders the changelog as before. Emitted **whether or not** anything was declared: an empty section that is *stated* beats an absent one, because it turns an undeclared change from silence into a visible false statement — the only version a reviewer can act on.

**3. The trailer convention, in CONTRIBUTING.** Four required elements, and the fourth is the point:

> whether the affected inputs were previously **rejected** (no consumer has the old string stored — free) or previously **accepted** (someone does — a migration).

That distinction cannot be recovered later. Nobody reading the diff in six months can tell, so it has to be written down at the time.

## Verified

`git-cliff` 2.13.1 is available locally, so I checked the parser behaviour on a throwaway repo rather than trusting the docs. Three commits — a `feat:`, a `fix:` **carrying the trailer**, and a `fix:` without it:

```
### Representation changes
- move a form

### Added
- plain feature

### Fixed
- a fix with no trailer
```

This confirms the subtlety that matters: parsers are first-match-wins, so the footer rule takes precedence over `^fix` for the trailered commit while the untrailered `fix:` still lands in Fixed. A `Representation-Change:` on a `feat:` or `fix:` does not lose its conventional grouping by accident — it is *promoted*, deliberately.

`release-plz.toml` parses (`tomllib`), `check-toml` and the rest of pre-commit pass.

**What I could not verify locally:** release-plz itself is not installed, so the `pr_body` Tera template and the exact reproduction of release-plz's default parser list are checked against its documented defaults and the git-cliff semantics above, not against a real release run. The first release PR after this merges is the real test — if the changelog groups look wrong there, the parser list is where to look.

## The two open questions, answered

**Advisory now, gated later.** The issue asks whether an undeclared change should be enforceable. It should eventually, and not yet. Having hand-applied this rule across five PRs this week, the expensive part was never *declaring* — it was **measuring**, which cost a dump-diff harness run on both sides of each change. A per-PR gate needs a corpus that is cheap, stable and blessed; the corpus that actually catches this class is #1400's transcript sweep, which is still `#[ignore]`d behind five defects. Gating on a corpus that does not run yet would be theatre. Build the declaration habit first.

**No backfill.** Ten output-touching commits are already merged this cycle. Reconstructing what each moved — without the measurement that should have accompanied it — would put *plausible* claims rather than measured ones into the one document whose entire value is that it is trustworthy. Better to write v0.13.0's summary by hand once, acknowledged as retrospective, and start the mechanism at v0.14.0 where every entry is backed by a real diff.

## Not included

The CI gate described in the issue's first open question. It wants costing separately: a fixed corpus, a normalized-output diff against the previous release tag, and a rule for failing when the diff is non-empty with no trailer in the range. Worth its own issue once the sweep is live.

## Applied from review

Three fixes since the first push, two of them things the review process was supposed to catch and did:

1. **`pr_body` was dropping the `cargo-semver-checks` report.** Setting `pr_body` replaces the default body wholesale, and the template rendered only `title` / `changelog` / `breaking_changes`. Since the body today *is* the semver dump, this PR would have deleted the only automated API-compatibility warning on a pre-1.0 crate in the course of adding a section to it. Now renders `release.semver_check` explicitly, with a comment saying why it is not decoration.
2. **`commit_parsers` was missing `^removed`.** Restored as `<!-- 4 -->Removed`, with `Fixed`/`Security`/`Other` renumbered to 5/6/7. Easy to miss by eye because this project's changelog has never had an entry under `Removed`.
3. **The "hand-edits are always overwritten" claim was too strong**, in both CONTRIBUTING and the `release-plz.toml` comment. Measured: v0.12.0's release commit (`65c01542`) touched `CHANGELOG.md` as a single `+63/-0` insertion at the top, so release-plz prepends and does not rewrite already-released sections; and it never revisits a GitHub release once its tag exists. What *is* regenerated on every run is the pending section of an open release PR. The distinction matters concretely — it is what makes a retrospective summary for v0.13.0 (whose commits predate this convention) possible after the fact, which is exactly what this PR's "no backfill" answer proposes doing. As written, the doc said that was impossible.

**Re-verified with git-cliff 2.13.1** on a throwaway repo, driving the parser list straight out of this PR's `release-plz.toml`. A trailered `fix:`, a plain `fix:`, a `feat:`, a `removed:` and a `chore:`:

```
### Representation changes
- move a form
### Added
- plain feature
### Removed
- dropped an old knob
### Fixed
- a fix with no trailer
### Other
- housekeeping
```

That run also turned up a fragility worth recording, now commented in the config: footer matching depends on `git.conventional_commits` staying enabled. With it off, the footer rule never matches and the trailered commit lands in `Fixed` — no error, no warning. So disabling that setting would switch this grouping off silently rather than break the build.

## CLI-lens findings

Three, of which one was real for a reason other than the one given.

**Applied — the documented trailer example was not a valid git trailer.** Its continuation lines were unindented. The finding predicted release-plz would not see it; measured against git-cliff 2.13.1, that part is **wrong** — git-cliff's `footer` rule regexes the whole footer paragraph and groups the unindented form correctly. But `git interpret-trailers --parse` reports **no trailers at all** for it:

```
--- unindented (as documented)   -> interpret-trailers sees: (nothing)
--- indented continuations       -> interpret-trailers sees: Representation-Change: …
--- single line                  -> interpret-trailers sees: Representation-Change: …
```

So the example happened to work for the changelog while not being a trailer — fine until anything else reads it (`%(trailers)`, a future CI gate that greps properly). The example now uses indented continuations and is verified under **both** parsers. Also notes the rule explicitly, since the failure mode is silent in the direction that matters.

**Applied — the example did not satisfy its own four requirements.** It gave a row count without naming the corpus, and never stated direction. Now names the slice (48 seeds x both directions x three shape families) and says "toward the already-shipped form".

**Rejected — "describe the trailer as policy rather than a required contract, or add enforcement."** The section already discloses exactly this, two paragraphs down: "Nothing enforces the trailer today: a change that moves output without one is simply invisible, which is why the release PR states the section even when it is empty." A convention can be required by policy without being machine-enforced, and the CI gate is deliberately out of scope here (see the answered open questions above).


nh13:
@coderabbitai pause

PR #1438, opened 2026-08-05
fix(normalize): order a conflicting cis allele whose members were rewritten

### LINKED ISSUE #1414: Normalizer: a conflicting cis allele skips the genomic-order sort even when its members were rewritten

nh13:
Closes #1414.

## The defect

`normalize_allele` skipped the genomic-order sort whenever an `OverlapConflict` was reported:

```rust
let has_overlap_conflict = all_warnings.iter()
    .any(|w| matches!(w, NormalizationWarning::OverlapConflict { .. }));
if is_cis && !allele.uncertain && !has_overlap_conflict && normalized.len() > 1 {
```

The skip itself is right and long-standing: an overlap-conflicting allele is emitted **verbatim** (#395), and reordering an allele you are handing back untouched would contradict that. What fails is the premise. *Conflict was reported* and *allele was emitted verbatim* are different predicates, and they come apart exactly when the repair passes succeed — `11_12inv` cancels to `11_12=` because the span is its own reverse complement, `11_12insAA` respells as `10_11A[4]`, and the pipeline then declines to order the two on the strength of a contract about *not* rewriting them.

Two consequences, and the first is the more serious:

1. **Out-of-genomic-order members** — a direct violation of #1235's criterion 2, on a shape its confluence proptest cannot generate (the indel model draws no inversion beside a co-located insertion).
2. **Non-idempotence**, for the subset whose reordered output no longer conflicts: pass 2 takes the *other* branch of the same gate and sorts, so `FERRO_ASSERT_IDEMPOTENT` fires.

The gate now asks the question the contract is actually about:

```rust
let emitted_verbatim = normalized == allele.variants;
if is_cis && !allele.uncertain && !(has_overlap_conflict && emitted_verbatim) && normalized.len() > 1 {
```

## #1423 did not make this unreachable

#1423 repaired the single input the issue cited — `g.[11_12inv;11_12insAA]` is now the single member `g.10_11A[4]` — by dropping an identity a sibling *overlaps*. A zero-width junction insertion overlaps nothing, so the whole `[inv;ins]` family survived it untouched. Measured on `origin/main` at `94817bf6` over four sequences and eleven member shapes, **41 inputs in four families** still reached the divergence:

| count | family | example |
|---|---|---|
| 16 | `[inv;insAA]` | `g.[2_3inv;2_3insAA]` -> `g.[2_3=;2_3insAA]` -> `g.[2_3insAA;2_3=]` |
| 16 | `[inv;insA]` | `g.[2_3inv;2_3insA]` -> `g.[2_3=;2_3insA]` -> `g.[2_3insA;2_3=]` |
| 7 | `[dup;inv]` | `g.[2_3dup;3_4inv]` -> `g.[8_9dup;3_4=]` -> `g.[3_4=;8_9dup]` |
| 2 | `[insA;inv]` | `g.[9_10insA;9_10inv]` -> `g.[11dup;9_10=]` -> `g.[9_10=;11dup]` |

With this change that census is **0**.

## Representation change — measured and disclosed

Per the representation-stability guidance, this PR **moves normalized output**, and here is the blast radius. Corpus: every two-member cis allele over `sweep_sequences(16)` (32 sequences x 11 member shapes x co-located and near-co-located spans), both shuffle directions — **313,984 rows**, diffed row-for-row against `origin/main`.

| measure | value |
|---|---|
| rows compared | 313,984 |
| rows moved | **2,037 (0.65%)** |
| moved rows that are a pure **reordering** (identical member multiset) | **2,037 / 2,037** |
| moved rows where any **member changed** | **0** |
| moved rows already non-idempotent on `main` | 209 |
| moved rows stable on `main` | 1,828 |
| split by direction | 859 3', 1,178 5' |

Reading that honestly, in the terms the guidance asks for:

- **No allele changes the sequence it denotes.** Every moved row emits the same members; only their order changes. That is the strongest form of "safe" available here, and it is measured rather than argued.
- **All 2,037 were out of genomic order on `main`**, i.e. criterion-2 violations. The direction of the change is toward the spec's stated form, not away from it.
- **The affected inputs were previously *accepted***, so for a consumer keying on the normalized string this is a **migration**, not a free fix. 209 of them were already unstable on `main` (its second pass produced this PR's first-pass answer, so anyone normalizing to a fixed point already holds the new string). The other **1,828 were stable on `main` and genuinely move** — that is the cost, and it is the larger share.
- Not release-to-release: the harness (`sweep_sequences`/`sweep_seeds`) postdates v0.12.0, so this corpus does not exist at that tag. The comparison is `main`-to-branch.

## One existing test changed

`issue_1235_transcript_axes::overlap_conflicting_allele_is_not_canonicalized` pinned **two** outputs for the two authored orders of one variant:

```
c.[5_9inv;5_6dup] -> c.[5_9inv;6_7insAA]
c.[5_6dup;5_9inv] -> c.[6_7insAA;5_9inv]
```

Two strings for one variant, differing only in member order — while the comment immediately above them reads *"Convergence is the point — one variant, one canonical string"*. Both now reach `c.[5_9inv;6_7insAA]`, so the test pins what it says it wants. Its other guards are untouched: the allele stays multi-member, strict still rejects it as W5002, and the settled form is still a fixed point.

The reversed-order `ins` spelling remains a fixed point of its own, because it *is* authored exactly as its members settle and so is genuinely returned verbatim. Whether a conflicting allele should be preserved that faithfully is #1406's open question, not this one; it is now asserted explicitly so the residual is visible rather than mistaken for convergence.

## Verification

- `cargo nextest run --features dev` with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1` and `FERRO_SWEEP_SEEDS=full`: **9348 passed, 0 failed**.
- New `tests/it/issue_1414_conflicting_allele_member_order.rs`: six tests covering the rewritten case on both shuffle directions, the one-pass fixed point, the verbatim contract still holding for an out-of-order allele that was *not* rewritten (the discriminating half — an unconditional sort passes everything else and fails this), #1423's single-member output, and the ordinary disjoint path.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -3114,9 +3114,10 @@ impl<P: ReferenceProvider> Normalizer<P> {
         // accession-string sort is not it — so those are left in authored order.
         //
         // Skip when the overlap passes above (`detect_insertion_overlaps`,
-        // `detect_overlap_conflicts`) flagged an `OverlapConflict`. Reordering is
-        // only meaning-preserving for *disjoint* members; a conflict means two
-        // members collide on one molecule, so they are NOT independent.
+        // `detect_overlap_conflicts`) flagged an `OverlapConflict` **and** the
+        // allele really was handed back untouched. Reordering is only
+        // meaning-preserving for *disjoint* members; a conflict means two members
+        // collide on one molecule, so they are NOT independent.
         // `merge_consecutive_edits` collapses strictly-adjacent members but does
         // not resolve a genuine overlap — that is exactly what those passes warn
         // about — so overlapping members can reach here. Leaving them in authored
@@ -3127,10 +3128,34 @@ impl<P: ReferenceProvider> Normalizer<P> {
         // (`[(a;b)]`) are left untouched to preserve their authored form.
         // `sort_by` (not `sort_unstable_by`) is a belt-and-suspenders choice — the
         // key is already total, so stability is not relied upon.
+        //
+        // **The conflict alone does not earn the skip (#1414).** The reasoning
+        // above is a contract about output the pipeline did not touch, so it
+        // applies only when the output *is* untouched. A conflicting allele whose
+        // members were all rewritten has no verbatim form left to preserve, and
+        // skipping the sort for it produced two defects at once:
+        // out-of-genomic-order members — a direct violation of #1235's criterion
+        // 2 — and non-idempotence, since the rewritten output no longer conflicts
+        // and so takes the *other* branch of this gate on the second pass.
+        // Measured on `origin/main` at `94817bf6`, 41 inputs in four shape
+        // families reach it (`[inv;insAA]` 16, `[inv;insA]` 16, `[dup;inv]` 7,
+        // `[insA;inv]` 2); #1423 had
+        // repaired only the single example the issue cited, because it drops an
+        // identity a sibling *overlaps* and a zero-width junction insertion
+        // overlaps nothing.
+        //
+        // So the predicate is the conjunction, not the conflict alone: an
+        // allele emitted verbatim keeps its authored order, and everything else
+        // is sorted exactly as before.
         let has_overlap_conflict = all_warnings
             .iter()
             .any(|w| matches!(w, NormalizationWarning::OverlapConflict { .. }));
-        if is_cis && !allele.uncertain && !has_overlap_conflict && normalized.len() > 1 {
+        let emitted_verbatim = normalized == allele.variants;
+        if is_cis
+            && !allele.uncertain
+            && !(has_overlap_conflict && emitted_verbatim)
+            && normalized.len() > 1
+        {
             sort_cis_members_by_genomic_order(&mut normalized);
         }
 
--- a/tests/it/issue_1235_transcript_axes.rs
+++ b/tests/it/issue_1235_transcript_axes.rs
@@ -322,27 +322,39 @@ fn overlap_conflicting_allele_is_not_canonicalized() {
     // collision repair rewrites it, in whichever member order it was authored.
     // Convergence is the point — one variant, one canonical string — and it is
     // reached in a single pass, so idempotency is preserved either way.
-    for (dup_spelling, converges_to) in [
-        (
-            "NM_TEST.1:c.[5_9inv;5_6dup]",
-            "NM_TEST.1:c.[5_9inv;6_7insAA]",
-        ),
-        (
-            "NM_TEST.1:c.[5_6dup;5_9inv]",
-            "NM_TEST.1:c.[6_7insAA;5_9inv]",
-        ),
-    ] {
+    //
+    // **Both authored orders now reach the same string (#1414).** They used to
+    // reach two, differing only in member order, because the genomic-order sort
+    // was skipped for any allele that reported an overlap conflict. That skip
+    // exists to honour #395's "preserve the input verbatim" contract, and the
+    // contract does not apply here: the collision repair rewrote the `dup` into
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

PR #1439, opened 2026-08-05
fix(spdi): read a single-position repeat anchor as the whole tract

### LINKED ISSUE #1431: Two spellings of one repeat denote different sequences: g.263A[7] and g.263_265A[7] disagree by two bases
## What's wrong

The two spellings of one repeat variant denote **different sequences**. Measured on `padded("GATCATAAATTCAGC")` (a 3-base `A` tract at 263–265), applying each member independently of the normalizer via `hgvs_to_spdi` + splice:

```
g.263A[7]      ->  TGATCAT AAAAAAAAA TTCAGC     9 A's
g.263_265A[7]  ->  TGATCAT AAAAAAA   TTCAGC     7 A's
```

Both descriptions say "this `A` tract has 7 copies". The second is right; the first adds two bases that are not there.

## Root cause: two readings of a single-position anchor, and they disagree

`hgvs_to_spdi` appears to read `g.263A[7]` as *"the one base named at 263 becomes 7 copies"* — a deletion span of 1, so `3 - 1 + 7 = 9`.

`merge::lowered_repeat` documents the opposite reading, explicitly:

> A single-position anchor (`e == s`) names only the tract's start and means "the whole run becomes N", so it absorbs nothing.

Under that reading the answer is 7, matching the range spelling. The two sites are inconsistent, so the same variant gets different bases depending on which one looks at it.

## Why it matters beyond the arithmetic

It makes the **apply oracle unreliable for single-position repeats**, and that oracle is what several sweeps and regression tests use as their independent check on the normalizer. A test that compares `apply(input)` against `apply(normalize(input))` will report a mismatch that is the oracle's, not the normalizer's — or, worse, agree for the wrong reason.

Concretely, it produces a well-formed output denoting different bases:

```
g.[263A[7];264dup]  ->  g.[263_265A[7];265_266insA]
  input  denotes  TGATCATAAAAAAAAAA TTCAGC    (10 A's)
  output denotes  TGATCATAAAAAAAA   TTCAGC    (8 A's)
```

Both apply cleanly, so no oracle in the suite flags it. It is *not* caught by `FERRO_ASSERT_REPARSE` (both parse), `FERRO_ASSERT_IN_BOUNDS` (both in range) or `FERRO_ASSERT_IDEMPOTENT` (the output is a fixed point).

Note the disagreement also changes the SPDI **triple width**, which decides overlap: `g.[263A[7];264dup]` is admitted by `triples_are_disjoint` while `g.[263_265A[7];264dup]` is declined as overlapping. Same variant, two spellings, opposite verdicts on whether it is even well-formed.

## Which reading is correct

`repeat.md` should settle it, and the answer matters for interop: `lowered_repeat`'s reading ("the whole run becomes N") is the one that makes the two spellings synonymous, which is the property a normalizer needs. Whichever is chosen, both sites must agree, and the sequence-changing case above should become a regression test.

## How it was found

While reviewing #1407 (`refuse to lower a repeat past a sibling inside its tract`). A CodeRabbit finding on that PR claimed the guard could not fire for a single-position repeat, because it tests the junction against the member's **stated** span and `s == e` makes the test vacuous. That claim is technically true of the guard in isolation but **not exploitable**: A/B measured, the per-member pipeline rewrites `263A[7]` to `263_265A[7]` before `lower_repeat_edits` ever runs, so `s != e` by then and the outputs are byte-identical with and without a physical-tract fix. The tract-bounds change was therefore reverted from #1407 as unreachable defensive code — but the probe written to test it turned up the inconsistency above, which is real and is not #1407's scope.

## Reproducing

Any worktree at current `main`; add to a test module with access to `common::cis_apply_oracle`:

```rust
let reference = padded("GATCATAAATTCAGC");
for input in ["TEMPLATE:g.263A[7]", "TEMPLATE:g.263_265A[7]"] {
    eprintln!("{input} -> {:?}", apply(&reference, input));
}
```

nh13:
Closes #1431.

## The disagreement

Two sites read a start-only repeat anchor two ways, and the two readings denote different sequences. Measured on `padded("GATCATAAATTCAGC")`, a 3-base `A` tract at 263-265:

| description | `hgvs_to_spdi` (before) | denotes |
|---|---|---|
| `g.263A[7]` | `262:A:AAAAAAA` | `GATCAT` **AAAAAAAAA** `TTCAGC` — nine `A`s |
| `g.263_265A[7]` | `262:AAA:AAAAAAA` | `GATCAT` **AAAAAAA** `TTCAGC` — seven `A`s |

Both descriptions say "this `A` tract has seven copies". `merge::lowered_repeat` already documents the reading that makes them agree, explicitly:

> A single-position anchor (`e == s`) names only the tract's start and means "the whole run becomes N", so it absorbs nothing.

## The spec settles which reading is right

`assets/hgvs-nomenclature/docs/recommendations/DNA/repeated.md` presents the two spellings as two **formats of one variant**:

> **NOTE**: a Community Consultation proposal is being prepared which will suggest to allow only the format where the **entire range** of the repeated sequence is indicated; so `g.123_191CAG[23]`, **not** `g.123CAG[23]`.

`123_191` is 69 bases — the whole 23-copy tract. So the start-only form addresses the run, not one base of it. `lowered_repeat` was right; `hgvs_to_spdi` was wrong.

## A second defect the issue does not record: the start-only form was unusable for every unit the spec illustrates it with

The one-base reading does not merely give a wrong answer for a 1-base unit — for a multi-base unit it is a hard error, because a one-base deletion span is not divisible by the unit length:

```
g.263CAG[5]  ->  <ERR: repeat span TEMPLATE:263-263 length 1 is not a multiple of unit length 3>
```

`g.123CAG[23]` is the spec's own example of this format. It could not be converted at all.

## The fix

Widen a start-only repeat to its physical run before fetching the deletion bases, using the **same** `count_tandem_repeats` the normalizer uses, so the two sites cannot drift apart again.

The tract's extent is not known before it is read, so the reference is fetched in a window around the anchor that **doubles while the run still reaches an edge**. A run still growing at `MAX_REPEAT_EXPANSION_BASES` is **declined**, not reported at the window's width — returning a clamped span would silently under-count the run and emit a triple denoting the wrong bases, which is the exact defect being removed. The initial half-width is 128 bases, comfortably covering the spec's 69-base example, so the ordinary case costs one fetch.

After:

| description | triple | denotes |
|---|---|---|
| `g.263A[7]` | `262:AAA:AAAAAAA` | seven `A`s |
| `g.263_265A[7]` | `262:AAA:AAAAAAA` | seven `A`s |
| `g.263CAG[5]` | `262:CAGCAGCAG:CAGCAGCAGCAGCAG` | five copies |
| `g.263_271CAG[5]` | `262:CAGCAGCAG:CAGCAGCAGCAGCAG` | five copies |

Identical triples, not merely equivalent ones — which matters, because the deletion **width** is what decides overlap.

## The overlap verdict now agrees too

`triples_are_disjoint` reads the SPDI deletion width, so a one-base-wide repeat triple missed a sibling inside its own tract:

```
g.[263A[7];264dup]        admitted    (repeat triple one base wide, misses the dup at 264)
g.[263_265A[7];264dup]    declined    (overlapping)
```

One variant, two spellings, opposite answers to "is this even well-formed". Both decline now, which is the correct verdict — the repeat spans 263-265 and the duplication claims 264.

## Unchanged on purpose

An explicit range keeps its own width, **including a range that names only part of a tract**. `g.263_265CAG[5]` still stays 3 bases wide and still absorbs only its own copies (five stated plus the two the range does not name = seven), because a range states which reference copies the count applies to. Widening every repeat to its physical tract would break this and would pass every other assertion in the new module — so it has a dedicated test.

An anchor whose unit does not match the reference is also unchanged: the tract search finds nothing, falls back to the unit-wide span, and the existing `does not match repeat unit` diagnostic reports it rather than a second error for one condition.

## Representation stability

**No normalized HGVS string moves.** The change is confined to `hgvs_to_spdi`'s `Repeat` arm, and `Normalizer` does not route through it — repeats reach `merge::lowered_repeat` instead, which already held the correct reading. The full suite (9346 tests, all three oracles, `FERRO_SWEEP_SEEDS=full`) passes with **no existing test moved**.

What does change is SPDI output for start-only repeats, and with it `EquivalenceChecker`: it compares members via `hgvs_to_spdi`, so it previously judged `g.263A[7]` and `g.263_265A[7]` to denote *different* sequences. They now compare equal, which is the correct answer. Both changes are from wrong-or-erroring to right.

## Why this was worth fixing beyond the arithmetic

It made the apply oracle unreliable for single-position repeats, and that oracle is the independent check several sweeps and regression tests grade the normalizer against — including #1400's transcript-axis sweep. A mismatch it reported could be the oracle's rather than the normalizer's, or it could agree for the wrong reason. None of `FERRO_ASSERT_REPARSE` (both parse), `FERRO_ASSERT_IN_BOUNDS` (both in range) or `FERRO_ASSERT_IDEMPOTENT` (the output is a fixed point) can see this class.

## Verification

- `cargo nextest run --features dev` with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1` and `FERRO_SWEEP_SEEDS=full`: **9346 passed, 0 failed**, no existing expectation changed.
- New `tests/it/issue_1431_single_position_repeat_anchor.rs`: seven tests — the two spellings denoting one sequence, producing one triple, the multi-base unit working at all, the partial range still absorbing only its own copies (the discriminating half), the overlap verdict agreeing, the no-matching-tract diagnostic preserved, and a lone copy keeping its one-unit width.

### DIFF
--- a/src/spdi/convert.rs
+++ b/src/spdi/convert.rs
@@ -93,6 +93,21 @@ use crate::sequence::reverse_complement;
 /// we'd plausibly emit as a single SPDI.
 const MAX_REPEAT_EXPANSION_BASES: usize = 100_000;
 
+/// Widest reference window [`resolve_repeat_tract_span`] will read while looking
+/// for the physical extent of a start-only repeat anchor.
+///
+/// Named separately from [`MAX_REPEAT_EXPANSION_BASES`] because it bounds a
+/// different thing: that one caps the `ins` string a repeat *count* can
+/// generate, this one caps how much reference is *read* to find the tract the
+/// anchor names. Reusing the expansion bound for both read as if one number
+/// governed both, and the growth loop compared it against a half-width, so the
+/// window it declined at was twice the figure its own error message quoted.
+///
+/// 200 KB is far above any biological tandem repeat — the largest known
+/// pathogenic expansions run to a few tens of kilobases — so a tract still
+/// growing at this width is a degenerate reference, not a variant.
+const MAX_REPEAT_SEARCH_BASES: u64 = 200_000;
+
 /// Error type for conversion failures.
 ///
 /// Marked `#[non_exhaustive]` so a new conversion-failure case is additive
@@ -1533,8 +1548,42 @@ where
                     ),
                 });
             }
+            // A single-position anchor names the tract's *start*, not a
+            // one-base span, so widen it to the physical run before fetching
+            // (#1431).
+            //
+            // The spec presents the two spellings as two formats of one
+            // variant — "a Community Consultation proposal is being prepared
+            // which will suggest to allow only the format where the **entire
+            // range** of the repeated sequence is indicated; so
+            // `g.123_191CAG[23]`, **not** `g.123CAG[23]`"
+            // (`DNA/repeated.md`) — and `123_191` is 69 bases, the whole
+            // 23-copy tract. Reading the anchor as a one-base span instead
+            // made the two disagree: on a 3-base `A` tract at 263,
+            // `g.263A[7]` came out as `262:A:AAAAAAA`, which denotes **nine**
+            // `A`s once the two untouched tract bases are counted, while
+            // `g.263_265A[7]` correctly denotes seven.
+            //
+            // The one-base reading also does not survive a multi-base unit at
+            // all: `g.263CAG[5]` fetched a single base and died on the
+            // divisibility check below ("length 1 is not a multiple of unit
+            // length 3"), so the spec's own start-only format was unusable for
+            // every unit the spec actually illustrates it with.
+            //
+            // This is the reading `merge::lowered_repeat` already documents —
+            // "a single-position anchor (`e == s`) names only the tract's
+            // start and means 'the whole run becomes N', so it absorbs
+            // nothing" — so widening here makes the two sites agree rather
+            // than introducing a third answer.
+            let (del_start, del_end) = match provider {
+                Some(p) if start_one_based == end_one_based => {
+                    resolve_repeat_tract_span(p, &sequence, start_one_based, unit_str.as_bytes())?
+                }
+                _ => (start_one_based, end_one_based),
+            };
+            let spdi_pos = del_start - 1;
             let del_raw = match provider {
-                Some(p) => fetch_reference_bases(p, &sequence, start_one_based, end_one_based)?,
+                Some(p) => fetch_reference_bases(p, &sequence, del_start, del_end)?,
                 None => {
                     return Err(unspelled_bases_error(
                         UnspelledBases::RepeatTract,
@@ -1552,8 +1601,8 @@ where
                     description: format!(
                         "repeat span {}:{}-{} length {} is not a multiple of unit length {}",
                         sequence,
-                        start_one_based,
-                        end_one_based,
+                        del_start,
+                        del_end,
                         del_str.len(),
                         unit_str.len()
                     ),
@@ -1567,7 +1616,7 @@ where
                 return Err(ConversionError::InvalidPosition {
                     description: format!(
                         "repeat span {}:{}-{} does not match repeat unit {}",
-                        sequence, start_one_based, end_one_based, unit_str
+                        sequence, del_start, del_end, unit_str
                     ),
                 });
             }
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1442, opened 2026-08-06
test(repr): commit the normalized-output dump/diff harness

### LINKED ISSUE #1441: Commit the normalized-output dump/diff harness used for representation-stability measurements
## Why

Representation stability is a shipped guarantee: the downstream consumer keys read counts on the normalized HGVS string, so every PR that moves a normalized output has to say which forms moved, in which direction, over how many rows, and whether the affected inputs were previously rejected (free) or previously accepted (a migration).

Five PRs this week did that measurement, and **every one of them rebuilt the harness from scratch**. That has two costs:

1. **The measurements are not reproducible or reviewable.** Each number in a PR body rests on a throwaway script that no longer exists. A reviewer cannot re-run it, and a later release cannot diff against it.
2. **It is easy to get wrong, silently.** Measuring #1401 I first keyed rows on the input string alone. `SyntheticBuilder::genomic` uses a fixed `NC_TEST.1` accession, so 2,160 rows collapsed to 135 and answers from different reference sequences overwrote each other — reporting "6 moved" where the real figure was 498. Nothing failed; the number was just wrong, and wrong in the reassuring direction.

The retrospective v0.13.0 representation summary needs the same harness, so this is a prerequisite rather than a nice-to-have.

## What already exists — and the actual gap

`ferro normalize` already covers the real-reference case:

```
ferro normalize --input corpus.txt --reference <prepared-dir> --format tsv
# -> line, input, normalized, changed, status, detail
```

So **do not add a second dumper for that path.** What it cannot do is drive a *synthetic* reference: `--reference` requires a prepared `manifest.json`, and the JSON-reference constructor is Python-only (`Normalizer(reference_json=...)`). Synthetic references are what densely cover the cis-allele shape families where representation churn actually happens, and they need no prepared reference on disk.

## Proposal

A `dev`-feature example, `dump_normalized_corpus`, with two modes.

**Dump.** Generate a deterministic synthetic corpus over synthetic references and emit stable TSV:

```
reference  axis  direction  input  output  input_was_fixed_point  status
```

- Keyed on `(reference, axis, direction, input)` — never on `input` alone. This is the trap above; a unit test should pin that the key is unique across the emitted corpus.
- `input_was_fixed_point` (`output == input`) is what makes the cheap/expensive classification derivable from a single old dump, without re-running the old code.
- Deterministic and prefix-stable, so a reduced seed count is a strict subset of a larger one.

Shape families worth covering, each drawn from a defect this repo has actually had:

| family | class it exercises |
|---|---|
| split vs spanning `delins` separated by one unchanged base | #1232 / #1401 |
| `dup` + `sub` at one-base separation | #999 |
| two insertions at adjacent junctions | #1235 |
| `dup` + `ins` sharing a junction | #1320 / #1323 |
| `del` + `sub` separated by one unchanged base | #1232's own example |

Both shuffle directions, and both a genomic and a coding reference (several defects are axis-specific — see #79's table).

**Compare.** `--compare <old.tsv> <new.tsv>`: join on the key and report total rows, moved rows and percentage, a breakdown by shape family, and the cheap/expensive split derived from `input_was_fixed_point` in the old dump. That is exactly the table a representation-moving PR body has to carry, produced by a reviewable tool rather than by hand.

## Out of scope

- Replacing or duplicating `ferro normalize --format tsv` for prepared references.
- Any CI gate on the output. Declaring movement is advisory today (see #1424 / #1433); a gate wants a blessed corpus and its own costing.
- Committing dump artifacts. They are build outputs; generate them at the two revisions being compared.

## Acceptance

- Running the dump at two revisions and comparing reproduces a movement table equivalent to the ones in #1425/#1427/#1428/#1432/#1401 bodies for the shapes they cover.
- A unit test pins key uniqueness across the corpus, and another pins prefix-stability of the seed expansion.
- The example's own doc comment states plainly that a within-shape-family rate is not a library-wide rate — the #1401 measurement is 23% within its family and would be badly misleading quoted as a repo-wide figure.

nh13:
Closes #1441.

Commits the harness that five PRs this week each rebuilt by hand. Every representation-movement number in a PR body rested on a throwaway script that no longer exists — unreviewable, and not re-runnable at the next release.

## What it does

```text
# at the base revision
cargo run --features dev --example dump_normalized_corpus -- --out /tmp/before.tsv
# at the candidate revision
cargo run --features dev --example dump_normalized_corpus -- --out /tmp/after.tsv
# the table for the PR body
cargo run --features dev --example dump_normalized_corpus -- \
    --compare /tmp/before.tsv --against /tmp/after.tsv
```

The dump is TSV: `reference, axis, direction, family, input, output, was_fixed_point`. `--compare` joins two dumps and reports total rows, movement with a percentage, a per-shape-family breakdown, and the cheap/expensive split.

`was_fixed_point` is the column that makes the cheap/expensive classification derivable from the **baseline dump alone** — "was this input already its own normalized form before the change?" — so you never need the old code available to classify.

## Deliberately scoped to synthetic references

`ferro normalize --input <file> --reference <dir> --format tsv` already dumps a real corpus against real reference data, and a second implementation of that would be two things to keep in step. What it **cannot** do is drive a synthetic reference: `--reference` requires a `manifest.json`, and the JSON-reference constructor is Python-only (`Normalizer(reference_json=...)`). Synthetic references are what densely cover the cis-allele shape families where representation churn actually happens, and they need nothing on disk.

Five families, each drawn from a defect this repo has had: `split_vs_spanning_delins` (#1232/#1401), `dup_plus_sub` (#999), `adjacent_junction_ins` (#1235), `dup_plus_ins` (#1320/#1323), `del_plus_sub` (#1232's own example). Both axes, both shuffle directions.

## Validated by reproducing a measurement made by hand

Run against #1401 (base `1ee1f9a2`), 24 seeds → 16,128 rows:

| | rows |
|---|---|
| total | 16,128 |
| **moved** | **1,840 (11.4%)** |
| previously **accepted** (a migration) | 1,688 |
| previously not a fixed point (free) | 152 |

| family | rows | moved |
|---|---|---|
| `split_vs_spanning_delins` | 5,376 | 1,840 |
| `dup_plus_sub` | 2,688 | **0** |
| `adjacent_junction_ins` | 2,688 | 0 |
| `del_plus_sub` | 2,688 | 0 |
| `dup_plus_ins` | 2,688 | 0 |

The hand measurement of #1401 got 498 moved of 2,160 with 457 expensive — an expensive share of **91.8%**. This harness, over 7.5× more rows, gets **91.7%** (1,688/1,840). Independent agreement on the ratio that actually matters, and the #999 control holds at 0 of 2,688.

It also surfaced something the hand run missed, because that run only covered `g.`: movement is **perfectly symmetric across axes** — 460 moved and 422 expensive in each of the four (axis, direction) cells. That is the fingerprint of `partition_block` being axis-blind, which is directly load-bearing for how #1401 should be reworked.

## The two traps it pins

Both are documented in the module header and guarded by tests:

1. **A row's identity is `(reference, axis, direction, input)`, never the input alone.** Every synthetic reference reuses one accession, so keying on the input string merges rows drawn against different sequences — 2,160 rows collapsed to 135 and reported 6 moved instead of 498, with nothing failing. `the_corpus_key_is_unique` asserts uniqueness under the tuple *and* asserts that inputs do repeat, so the reason the key is a tuple survives a future refactor.
2. **A within-family rate is not a library-wide rate.** The corpus is enriched for shapes that churn, so its percentages run an order of magnitude above a real corpus. The `--compare` output says so in its own footer, so a number copied out of it carries the caveat with it.

## Not included

- Any CI gate. Declaring movement is advisory today (#1424 / #1433); a gate wants a blessed corpus and its own costing.
- Committed dump artifacts — they are build outputs, generated at the two revisions being compared.
- The `n.`/`r.` axes and non-cis shapes. Easy to add a family; deliberately not speculative.

## Verification

- Suite **9361 passed** under all three oracles (up 5 — this example's own unit tests, which run because it is declared `test = true`).
- `fmt` / `clippy --all-features --all-targets -D warnings` clean.
- Five unit tests: key uniqueness, seed prefix-stability, every family emitting on both axes, the reader rejecting a duplicate key, and a self-comparison reporting exactly zero movement.

## CLI-lens findings — all four applied, and one was a real coverage hole

**The coding corpus never crossed `CDS_END`.** It topped out at exactly the last CDS base, so it emitted no `c.*N` row and no junction-straddling shape at all — meaning it would have reported a confident **"0 moved"** for #1417, #1425, #1427 and #1428, every one of which is a CDS/3'UTR junction defect. That is precisely the failure mode this tool exists to prevent, built into the tool.

Fixed three ways: a `hgvs_pos` that spells a post-CDS position `*N` rather than `N`; the shape loop extended so the widest shapes actually straddle the boundary; and `CDS_END` dropped from 18 to **15** (still codon-complete) so five 3'UTR bases remain and several shapes cross. Pinned by `the_coding_corpus_crosses_into_the_three_prime_utr`, which asserts both that `c.*N` rows exist and that some *straddle*, and by `coding_positions_past_the_cds_are_spelled_with_a_star`.

**The header was matched by prefix, and `was_fixed_point` by `== "true"`.** A dump with reordered or renamed columns would have been read positionally under the old meanings — silently swapping `output` and `was_fixed_point` and inverting every migration verdict — and a corrupted boolean became `false`, which *understates* the migration. That is the direction that gets a change waved through. Now an exact header match and a strict `true`/`false` parse, both with line-specific errors, both pinned.

**The `family` column was taken from the baseline only.** If two dumps disagree about which family a key belongs to, the per-family table attributes movement to the wrong shape. Now refused with an explanation, pinned by `comparing_dumps_that_disagree_on_a_family_is_refused`.

**`--out` / `--seeds` were silently ignored in compare mode.** Now `conflicts_with_all`. Verified both directions: a *defaulted* `--seeds` still works alongside `--compare` (clap does not count defaults as present), while an explicit one errors.

## Re-measured after those fixes

The corpus changed, so the numbers did. Against #1401, base `1ee1f9a2`, 24 seeds:

| | rows |
|---|---|
| total | 18,432 |
| **moved** | **1,780 (9.7%)** |
| previously **accepted** (a migration) | 1,640 |
| previously not a fixed point (free) | 140 |

| family | rows | moved |
|---|---|---|
| `split_vs_spanning_delins` | 6,144 | 1,780 |
| `dup_plus_sub` | 3,072 | **0** |
| `adjacent_junction_ins` | 3,072 | 0 |
| `del_plus_sub` | 3,072 | 0 |
| `dup_plus_ins` | 3,072 | 0 |

Expensive share **92.1%**, against 91.7% on the previous corpus and 91.8% from the original hand measurement. Three corpora of different sizes and shapes agreeing to within half a point on the number that decides whether a change is a migration is the reassurance worth having from this tool.

## Verification

- Suite **9366 passed** under all three oracles; `fmt` / `clippy --all-features --all-targets -D warnings` clean.
- **10** unit tests now: key uniqueness (and that inputs *do* repeat, so the reason the key is a tuple survives a refactor), seed prefix-stability, every family emitting on both axes, 3'UTR crossing, `*N` spelling, duplicate-key rejection, wrong-header rejection, malformed-boolean rejection, family-mismatch refusal, and a self-comparison reporting exactly zero movement.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -299,6 +299,11 @@ name = "derive_tx_placements"
 required-features = ["dev"]
 
 [[example]]
+name = "dump_normalized_corpus"
+required-features = ["dev"]
+test = true
+
+[[example]]
 name = "extract_biocommons_windows"
 required-features = ["dev"]
 
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -1,0 +1,792 @@
+//! Dump normalized output over a deterministic synthetic corpus, and diff two
+//! dumps — the measurement a representation-moving PR has to carry (#1441).
+//!
+//! Representation stability is a shipped guarantee: the downstream consumer keys
+//! read counts on the normalized HGVS string, so a PR that moves an output has to
+//! say which forms moved, in which direction, over how many rows, and whether the
+//! affected inputs were previously **rejected** (no stored string, so free) or
+//! previously **accepted** (a migration). Five PRs made that measurement by hand
+//! and each rebuilt the harness; this is that harness, committed.
+//!
+//! ## Usage
+//!
+//! ```text
+//! # at the base revision
+//! cargo run --features dev --example dump_normalized_corpus -- --out /tmp/before.tsv
+//! # at the candidate revision
+//! cargo run --features dev --example dump_normalized_corpus -- --out /tmp/after.tsv
+//! # the table for the PR body
+//! cargo run --features dev --example dump_normalized_corpus -- \
+//!     --compare /tmp/before.tsv --against /tmp/after.tsv
+//! ```
+//!
+//! ## What this is NOT for
+//!
+//! Prepared references. `ferro normalize --input <file> --reference <dir> --format
+//! tsv` already dumps a real corpus against real reference data, and duplicating it
+//! here would mean two things to keep in step. What that path *cannot* do is drive a
+//! **synthetic** reference — `--reference` requires a `manifest.json` and the
+//! JSON-reference constructor is Python-only — and synthetic references are what
+//! densely cover the cis-allele shape families where representation churn actually
+//! happens. That is the gap this fills.
+//!
+//! ## Two traps, both hit while measuring #1401
+//!
+//! 1. **A row's identity is `(reference, axis, direction, input)` — never the input
+//!    alone.** Every synthetic reference reuses one accession, so keying on the
+//!    input string silently merges rows drawn against different sequences: 2,160
+//!    rows collapsed to 135 and the reported movement came out at 6 instead of 498.
+//!    Nothing failed. `the_corpus_key_is_unique` pins this.
+//! 2. **A within-shape-family rate is not a library-wide rate.** This corpus is
+//!    deliberately enriched for the shapes that churn, so its percentages are far
+//!    higher than a real corpus would give — #1401 measures 23% here. Quote it as
+//!    "of the affected shape family", never as a repo-wide figure.
+//!
+//! ## The corpus is deliberately independent of `tests/it/common/`
+//!
+//! An example cannot reach test helpers, so the reference construction below is its
+//! own. That is a feature rather than a compromise: this corpus has to stay
+//! **stable across revisions** so that two dumps are comparable, and it must not
+//! move because someone refactored `SyntheticBuilder`. It does not need to agree
+//! with the test helpers — only with itself, over time. Changing the generator or
+//! the seed count re-rolls the corpus and makes old dumps incomparable; treat it
+//! the way `sweep_sequences` is treated.
+
+use std::collections::BTreeMap;
+use std::fmt::Write as _;
+use std::fs;
+use std::path::PathBuf;
+use std::process::ExitCode;
+
+use clap::Parser;
+
+use ferro_hgvs::normalize::{NormalizeConfig, ShuffleDirection};
+use ferro_hgvs::reference::transcript::{Exon, GenomeBuild, ManeStatus, Strand, Transcript};
+use ferro_hgvs::reference::MockProvider;
+use ferro_hgvs::{parse_hgvs, Normalizer};
+
+/// 256 bases of period-4 `ACGT`, so the base immediately 5' of a core is `T` and
+/// the one immediately 3' of it is `A`. A core starting with something other than
+/// `A` and ending with something other than `T` therefore cannot extend the pad's
+/// own rotation, which bounds a repeat tract to the core.
+const PAD_OFFSET: usize = 256;
+
+/// The genomic contig a `g.` row is drawn against.
+const GENOMIC_CONTIG: &str = "NC_TEST.1";
+/// The transcript a `c.` row is drawn against.
+const TX_ACCESSION: &str = "NM_TEST.1";
+/// The contig the synthetic transcript is placed on.
+const TX_CONTIG: &str = "chr_synth";
+
+/// The CDS of a synthetic coding reference: 1-based inclusive.
+///
+/// A multiple of three, so the CDS is codon-complete, and deliberately short of the
+/// 20-base core so that **five** 3'UTR bases (`c.*1`..`c.*5`) remain. Several shapes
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai pause

Issue #1421, opened 2026-08-05
Normalizer: net-insertion spanning `delins` not split across an unchanged interior base (follow-up to #1232); non-confluent


msto:
> [!NOTE]
> Authorship: the content below was drafted by GitHub Copilot CLI (an AI coding
> agent) and filed via gh under @msto's account. It reflects the agent's
> analysis, not a statement authored by @msto.

## What's wrong

`Normalizer.normalize` keeps a single spanning `delins` that covers an **unchanged
interior nucleotide** when the change is a **net insertion** (the replacement is
longer than the reference span), instead of splitting it into the separated members
the same sequence represents. The split encoding normalizes to a different, stable
form, so one variant has two normalized representations.

This is the net-insertion counterpart of **#1232**. #1232 was closed as fixed by
#1237 (sequence-first cis-allele canonicalization), with follow-ups #1341 and #1373
(*"guard a net deletion's split by rule, not by a length bound"*). Those resolved
the **net-deletion** case — #1232's own example `35_39delinsTA` → `[35_37del;39T>A]`
is confluent on current `main` — but a spanning `delins` that is a **net insertion**
across an unchanged interior base is still retained.

## Minimal example

Reference `TEMPLATE`, positions 29–33 are `CACGT`. The change substitutes position
29, **leaves 30–31 (`AC`) unchanged**, and rewrites 32–33 with a net +6 insertion:

```
TEMPLATE:g.29_33delinsAACACATACTG        -> TEMPLATE:g.29_33delinsAACACATACTG        # spanning delins (kept)
TEMPLATE:g.[29C>A;32_33delinsACATACTG]   -> TEMPLATE:g.[29C>A;32_33delinsACATACTG]   # sub + separated delins (kept)
```

Both describe the identical resulting sequence and are each idempotent — **two
stable fixed points for one variant** (non-confluence). No error/warning.
`EquivalenceChecker` reports the two as `SequenceMatch`, confirming they are the
same variant (so it remains a correct fallback for comparison).

### Why the split form is canonical

The substitution at 29 and the `delins` at 32–33 are **separated by two unchanged
nucleotides** (30, 31):

> two variants separated by one or more nucleotides should be described
> individually and **not** as a "delins".
>
> — `assets/hgvs-nomenclature/docs/recommendations/DNA/delins.md:17`
> (restated `delins.md:81`)

So the canonical form is `[29C>A;32_33delinsACATACTG]`; the spanning
[...]

Issue #1419, opened 2026-08-05
Normalizer: multi-member deletion cis-allele not re-derived from the resulting sequence — disagrees with the equivalent `delins` (partition non-uniqueness); non-confluent


msto:
> [!NOTE]
> Authorship: the content below was drafted by GitHub Copilot CLI (an AI coding
> agent) and filed via gh under @msto's account. It reflects the agent's
> analysis, not a statement authored by @msto.

## What's wrong

When a net-deletion variant is given as a **multi-member deletion cis-allele**
(`[del; del]`), `Normalizer.normalize` shifts each member independently and stops,
rather than re-deriving the edit partition from the resulting sequence. The **same
variant** given as a single spanning `delins` is re-derived (via the sequence-first
machinery from #1237) and normalizes to a **different** partition. The two encodings
are each idempotent, so one variant has two stable normalized representations.

This differs from #1232 and the net-insertion spanning-`delins` follow-up: **neither**
form here is a spanning `delins` across an unchanged base. Both are already
"separated → individual" compliant (`delins.md:17`). There is no single local rule
that a form violates — it is **pure partition non-uniqueness**, so it is a direct
manifestation of the #1235 root cause (members normalized independently vs.
re-derived from the sequence).

## Minimal example

Reference `TEMPLATE`. The variant deletes 12 of the 15 bases in 19–33, and the two
retained bases can be aligned two ways:

```
TEMPLATE:g.[19_23del;27_33del]   -> TEMPLATE:g.[19_23del;27_33del]   # cis-allele: kept as two deletions
TEMPLATE:g.19_33delinsCGG        -> TEMPLATE:g.[19_30del;33T>G]      # same variant, one span: re-derived to del + terminal sub
```

Both describe the identical resulting sequence and are each idempotent — **two
stable fixed points for one variant** (non-confluence). No error/warning.
`EquivalenceChecker` reports the two as `SequenceMatch` (same variant), so it
remains a correct fallback for comparison.

### Which form is canonical

`[19_30del;33T>G]` is the better canonical candidate, on two independent grounds:

1. **ferro's own sequence-first path already produces it.** Handed the single
   spanning `delins`, ferro re-derives the minimal edit set and emits
   `[19_30del;33T>G]`. The cis-allele path simply fails to reach the same result —
   it never re-derives, it only shifts the two `del` members in isolation.
2. **prioritisation** (`general.md:56`): the preferred description order is
[...]

PR #1445, opened 2026-08-06
fix(normalize): keep a real overlap conflict visible in the output

nh13:
Addresses #1406 row 3. Leaves the issue open — see "What this deliberately does not do".

## The defect

W5002 says an overlap-conflicting allele has no canonical form, so ferro preserves it verbatim and strict mode rejects it. But strict mode re-reads a **description**, and has no memory of what that description was normalized from — so the contract only holds if the emitted description still *looks* conflicting.

It did not. The per-member pipeline repairs the members one at a time (an inversion whose span is its own reverse complement cancels to an identity; an interior insertion respells as a repeat), and the merge collapses what is left:

```
in      TEMPLATE:g.[11_12inv;11_12insAA]   strict: REJECT (W5002)
lenient TEMPLATE:g.10_11A[4]               strict: ACCEPT
```

Lenient mode converts a description ferro refuses into one it admits, with no warning surviving the trip. That also breaks the invariant `issue_1276_dup_junction_overlap::lenient_output_of_a_conflict_is_still_rejected_by_strict` already pins for the shapes it happens to cover.

## Measured, not assumed

The issue reports one input. Enumerating 23,134 two-member cis alleles over four sequences and eleven member shapes shows it is systemic:

| | `main` | this PR |
|---|---|---|
| inputs enumerated | 23,134 | 23,134 |
| strict-rejected | 8,457 | 8,177 |
| **laundered** (reject -> accept) | **1,888** | **486** |
| shape families laundered | 26 | 7 |

## Two changes

**1. When the input conflicts and the output does not, hand back the input.** That is #395's verbatim contract stated exactly, rather than approximated by "skip some passes and hope the conflict survives them". Idempotent by construction: the value handed back is the input, whose conflict the next pass detects again.

**2. A junction interior to a *pure deletion* is no longer a conflict.** A deletion removes every base it spans, so an interior junction has nothing left to be positioned against — `g.[2_3del;2_3insAA]` denotes `AA` in place of `2_3` whichever order the members are applied in and whichever interior junction the insertion named. It composes uniquely, so it was never a conflict. This is the same argument #1411 used to stop rejecting `g.[24dup;24C>G]`, and the merged form `g.2_3delinsAA` is what `delins.md:86-89` asks for.

Deletion **only**. A `delins` does not qualify — its payload survives, so an interior insertion has a position relative to it, and `g.[2_3delinsGG;2_3insA]` genuinely does not say whether the `A` lands before, inside or after the `GG`. Nor does `inv`, `dup` or `repeat`. `insertion_interior_to_delins_emits_one_warning` is the discriminating guard.

## What this deliberately does not do, and why

The remaining 486 rows are erased by the **sequence-first derivation**, which runs after the back-off and re-derives the allele. Closing them means gating that pass on the junction detectors too. That was implemented and measured, and it is the wrong trade:

```
#1301 regressed — TEMPLATE:g.[263_264insAC;264_265insAA] -> TEMPLATE:g.[264_265insCA;264_265dup]
                  but TEMPLATE:g.264_265insCAAA          -> TEMPLATE:g.264_265insCAAA
                  these two spellings of one variant must agree.
```

Those are two spellings of one variant. Gating the derivation stops them merging, so they settle apart — new non-confluence, caught by `cis_spelling_confluence_gap::converged_pairs_stay_converged` and again under 5' shuffle by `the_five_prime_confluence_gap_is_unchanged` ("#1320 has started diverging"). With the gate, 11 tests fail; without it, 4 (all addressed here).

That is precisely the class #1229-#1235 and #1419-#1421 track, and the failure mode behind the read-count dropouts they report. **Confluence is what downstream consumers key on; a strict-mode hygiene property is not worth buying with it.**

The underlying point: two writers at one junction are not an unresolvable conflict — `g.[263_264insAC;264_265insAA]` is simply a two-member *spelling* of `g.264_265insCAAA`, and merging them is what makes the two converge. So #1406's residual is narrower than the census first suggests, and the right frame for it is whether W5002 is over-firing rather than whether the repair is over-eager. The census by family is on the issue.

## Representation stability

**No normalized output moves except for alleles ferro already declares to have no canonical form.** The back-off fires only when the input carries an overlap conflict — i.e. strict mode rejects it, so ferro is on record that it has no canonical form and a consumer has no sanctioned normalized string for it. The full suite is green with no confluence or merge expectation changed, which is the measurement that matters: the shapes that *do* have a canonical form are untouched.

## The one place two issues disagreed

#1416's two oracle tests reached their shape by normalizing `g.[11_12inv;11_12insAA]`. That input is #1406 row 3's own witness, so #1416 wants it repaired and #1406 wants it preserved. They cannot both hold.

The evidence favours preserving it: that input denotes **no sequence at all** — the independent applier declines it in either member order — so there are no "input's bases" to preserve, and the answer the repair produced came from *choosing* an apply order, which is exactly what W5002 exists to refuse.

#1416's **fix** is untouched and still guarded. Its tests now enter at `g.[11_12=;10_11A[4]]` — the allele the fix actually operates on, which is not a conflict (strict mode accepts it) and is what #1416 reported. Only the entry point changed; both assertions, and the wraparound-sibling guard, are unchanged.

## Verification

- `cargo nextest run --features dev` with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1` and `FERRO_SWEEP_SEEDS=full`: **9349 passed, 0 failed**.
- `insertion_interior_to_deletion_emits_one_warning` becomes `insertion_interior_to_deletion_is_not_a_conflict` (the behaviour this PR changes), and `insertions_sharing_junction_and_interior_to_span_emit_two_warnings` moves its span from `4_7del` to `4_7delinsGG` so it still exercises both branches rather than silently dropping to one.

---

### The restore requires a plural output

Both detectors return nothing below two members, so for a **single-member** output `!conflicts(&normalized)` is *vacuously* true — there is nothing left for a second member to collide with. Two different things reach that state and only one of them is laundering:

- members **shifted apart** while still plural — the conflict was erased, the description launders. This is the target.
- members **composed into one edit** — the colliding writes were resolved into a single description denoting a definite sequence, which is what `delins.md:86-89` asks for and what the pure-deletion exemption here already relies on.

Without the plural term this gate **reverted #1423**: `g.[11_12inv;11_12insAA]` collapses to the single member `g.10_11A[4]`, which cannot conflict, so the input was handed back and a merged, shipped form was undone.

Consequence: the earlier revision had to re-point `issue_1416_cancelled_identity_beside_repeat.rs` at the downstream shape, on the premise that "the upstream spelling no longer reaches this fix". That premise is now false, so **the file is back to `main` unchanged** and this PR touches no existing test's expectations.

### Uncertain alleles are inside the gate

This is the one place the preserve gate deliberately differs from the genomic-order sort gate above it. That gate excludes `[(a;b)]` because member order inside an uncertain group is authored presentation. Whether the conflict survives is a different question, and the answer does not depend on the wrapper:

```
in  g.[(9_10insA;9_10inv)]   strict: REJECT (W5002)
out g.[(11dup;9_10=)]        strict: ACCEPT
```

Excluding them left the hole open on one spelling of the same input.

### Representation movement

`dump_normalized_corpus` (the #1442 harness), 18,432 rows across both axes and both shuffle directions, against `origin/main`: **0 moved (0.0%)**, zero in all five shape families.

That is a statement about *collateral* movement only. The corpus builds `dup_plus_sub`, `del_plus_sub`, `adjacent_junction_ins`, `dup_plus_ins` and `split_vs_spanning_delins` — none of which construct an overlap-conflicting allele, the only class this PR touches. The movement inside that class is the enumeration already reported above (laundered rows 1,888 -> 486).

### Coverage

`tests/it/issue_1406_lenient_output_keeps_the_conflict.rs`, five guards, each A/B'd:

| A/B | result |
|---|---|
| against `origin/main` | 3 of 4 original guards fail — the restore and the deletion exemption are real |
| plural term removed | the composition guard fails **and two of #1416's own tests fail** — the #1423 revert, demonstrated |
| `!allele.uncertain` restored | the uncertain guard fails |

`strict_rejects_as_conflict` keys on the W5002 diagnostic rather than `is_err()`, so an unrelated failure cannot masquerade as a successful rejection.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -3134,6 +3134,87 @@ impl<P: ReferenceProvider> Normalizer<P> {
             sort_cis_members_by_genomic_order(&mut normalized);
         }
 
+        // A conflict the input really had must survive into the output (#1406).
+        //
+        // #395 says an overlap-conflicting allele has no canonical form, so
+        // ferro preserves it verbatim and reports `W5002`, which strict mode
+        // promotes to an error. But strict mode re-reads a *description*, and it
+        // has no memory of what that description was normalized from — so the
+        // contract only holds if the emitted description still looks conflicting.
+        //
+        // It did not. The per-member pipeline repairs the members one at a time
+        // — an inversion whose span is its own reverse complement cancels to an
+        // identity, an interior insertion respells as a repeat — and the merge
+        // then collapses what is left. The conflict is gone from the output, so
+        // strict mode accepts a description it would have rejected had the
+        // caller written it directly:
+        //
+        //     in      g.[11_12inv;11_12insAA]   strict: REJECT (W5002)
+        //     lenient g.10_11A[4]               strict: ACCEPT
+        //
+        // That is the laundering #1406 row 3 reports, and it breaks the
+        // invariant `issue_1276_dup_junction_overlap::
+        // lenient_output_of_a_conflict_is_still_rejected_by_strict` already
+        // pins for the shapes it happens to cover.
+        //
+        // So: when the input conflicts and the output does not, hand back the
+        // input. That is #395's contract stated exactly — preserve it verbatim —
+        // rather than approximated by "skip some passes and hope the conflict
+        // survives them".
+        //
+        // Deliberately asked of the *members*, both detectors, on each side.
+        // `detect_insertion_overlaps` documents that it must see pre-merge
+        // members because the merge collapses the overlap out of view — and
+        // that collapse is precisely the erasure being detected here, so asking
+        // it of the post-merge output is the right question, not a misuse.
+        //
+        // Idempotent by construction: the value handed back is the input, whose
+        // conflict the next pass detects again, so it is returned unchanged a
+        // second time.
+        // Uncertain alleles included, unlike the sort gate above. That gate
+        // excludes them because member *order* inside `[(a;b)]` is authored
+        // presentation, which is not ours to change. This one is about whether
+        // the conflict survives into the output at all, and an uncertain allele
+        // launders exactly like a certain one — measured:
+        //
+        //     in  g.[(9_10insA;9_10inv)]   strict: REJECT (W5002)
+        //     out g.[(11dup;9_10=)]        strict: ACCEPT
+        //
+        // Excluding them would leave the hole this gate exists to close open on
+        // one spelling of the same input.
+        if is_cis {
+            let conflicts = |members: &[HgvsVariant]| {
+                !crate::normalize::overlap::detect_overlap_conflicts(members, allele.phase)
+                    .is_empty()
+                    || !crate::normalize::overlap::detect_interior_junction_conflicts(
+                        members,
+                        allele.phase,
+                    )
+                    .is_empty()
+            };
+            // A **plural** output only. Both detectors return nothing for fewer
+            // than two members (`overlap.rs`, first statement of each), so for a
+            // single-member output `!conflicts(&normalized)` is vacuously true
+            // and says nothing about whether the conflict was erased — there is
+            // simply nothing left for a second member to collide with.
+            //
+            // The distinction is real, not defensive. Members *shifted apart*
+            // while still plural is the laundering this gate is for. Members
+            // *composed into one edit* is the opposite: the colliding writes
+            // have been resolved into a single description that denotes a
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

PR #1446, opened 2026-08-06
fix(normalize): key a dup on what it writes in the interior-junction branch

### LINKED ISSUE #1437: detect_insertion_overlaps still keys a dup on its read span, so an interior ins is rejected where an interior sub is accepted
#1411 established the principle that a coincident-bounds conflict keys on what a member **writes**, not on the span it **reads** (`duplication.md:5` — a `dup` writes its copy directly 3' of the original, so `g.4_9dup` writes at the junction `9|10`, not across bases 4-9). Fix 1(a) applied that to `detect_overlap_conflicts`.

`detect_insertion_overlaps` branch (b) still keys on the read span, so the two detectors now disagree, and the verdict for a sibling interior to a `dup` depends on the sibling's edit *kind* rather than on whether anything actually collides.

## Measured

Reference `TTTTTTTTTAATATATTTTAATAC` (the 24 bp `NC_TEST.1` fixture shared by `strict_rejection_survives_normalization` and `issue_1307_terminal_dup_respell`), strict mode:

```
g.[4_9dup;6T>C]     -> ACCEPTED   g.5_6insCTTTTT
g.[4_9dup;8T>C]     -> ACCEPTED   g.7_8insCTTTTT
g.[4_9dup;9T>C]     -> ACCEPTED   g.8_9insCTTTTT
g.[4_9dup;6_7insT]  -> REJECTED   W5002, "g.4_9 has 2 coincident cis-allele edits (dup, ins)"
```

Every one of these siblings sits strictly interior to the `dup`'s read span, and in every one the `dup`'s write is the junction `9|10` — disjoint from the sibling's write. Three are accepted; the insertion is rejected. The distinguishing factor is only that a substitution is not an `NaEdit::Insertion`, so branch (b) never examines it.

## Mechanism

In `detect_insertion_overlaps` (`src/normalize/overlap.rs:137`), any `is_overlap_edit` member is pushed into `spans` unconditionally — including a `dup`:

```rust
} else if is_overlap_edit(edit) {
    if let Some(kind) = junction_writing_kind(edit) { insertions.push(/* junction occupant */); }
    spans.push(Span { idx, /* start..end = the READ span */ });
}
```

Branch (b) then reports any insertion junction strictly interior to a `Span`. For a `del`, `inv` or `delins` that is right — the read span *is* the write footprint. For a `dup` it is not, and branch (b) inherits the pre-#1411 reading that Fix 1(a) corrected elsewhere.

Registering the `dup` in `insertions` (as a junction occupant at its own `end`) is correct and should stay — that is what catches `g.[3_6dup;5_6dup]`, and the comment there explains why.

## Which way to resolve it

Both members' writes are disjoint and, per #1411's lesson that HGVS members are each defined against the reference rather than applied in sequence, the composition is unique. That argues for **accepting** the insertion, i.e. removing `dup` from the `spans` push (keeping `del`/`inv`/`delins`).

Stability cost of doing so is **cheap** in the #1411 sense: `g.[4_9dup;6_7insT]` and its shape are currently *rejected*, so no consumer has a stored normalized string for them. Nothing that is accepted today changes verdict. The reverse resolution — rejecting the substitution cases for symmetry — would be **expensive**, since those are accepted today and produce shipped output (`g.5_6insCTTTTT` and friends).

Either way the two detectors should agree, and whichever is chosen should be stated where the other one's reasoning lives.

## Provenance

Raised by the reviewer during #1411 as "Fix 1(a) is slightly too narrow", deferred as out of scope for that PR, and confirmed empirically here against current `main` (`94817bf6`).

nh13:
Closes #1437.

> **Stacked on #1445** (`fix/issue-1406-respell-launders-conflict`), which touches the same function and the same branch. Base this PR's review on the two-commit diff; merge #1445 first and GitHub will retarget this to `main`.

## The disagreement

#1411 established the principle: a conflict keys on what a member **writes**, not on the span it **reads**. A `dup` writes its copy directly 3' of the original (`duplication.md:5`), so `g.4_9dup` writes at the junction `9|10` and reads bases 4-9 without touching them. Fix 1(a) applied that to `detect_overlap_conflicts`. `detect_insertion_overlaps` branch (b) kept the old reading, so the two detectors disagreed.

Measured on the 24 bp `NC_TEST.1` fixture, strict mode. Every sibling below sits strictly interior to the `dup`'s read span, and in every one the `dup`'s write at `9|10` is disjoint from the sibling's:

```
g.[4_9dup;6T>C]     ACCEPTED   g.5_6insCTTTTT
g.[4_9dup;8T>C]     ACCEPTED   g.7_8insCTTTTT
g.[4_9dup;9T>C]     ACCEPTED   g.8_9insCTTTTT
g.[4_9dup;6_7insT]  REJECTED   W5002 "g.4_9 has 2 coincident cis-allele edits (dup, ins)"
```

Three accepted, one rejected, and the distinguishing factor is only that a substitution is not an `NaEdit::Insertion`, so branch (b) never examined it. Nothing about the two writes differed.

## The fix

A junction-writing edit is registered as a junction occupant **only**, never as a span. Its read span was never a write footprint, so reporting a sibling interior to it reported a collision that does not exist.

**`repeat` goes with `dup`, not after it.** The per-member pipeline picks between the two spellings by axis — the CDS axis rewrites `c.5_6insAA` to `c.5_6dup`, the genomic axis rewrites the same shape to `g.1005_1006A[4]` — so exempting one and not the other would make the detector spelling-sensitive, which is precisely the defect the junction registration above exists to prevent. A test pins that the two spellings get one verdict.

## Direction, and why it is the cheap one

Both members' writes are disjoint, so the composition is unique and accepting is correct on the merits. It is also the cheap direction for representation stability in #1411's sense: the rejected shapes are *rejected today*, so no consumer holds a normalized string for them, while the three accepted ones produce shipped output (`g.5_6insCTTTTT` and friends). Resolving the other way — rejecting the substitutions for symmetry — would move output consumers already have.

## Unchanged on purpose

- **Branch (a) is untouched.** Two `dup`s writing at one junction still conflict (`g.[4_9dup;7_9dup]`), which is what the junction registration is for.
- **A span that really does write across its range still conflicts with an interior insertion.** An `inv` rewrites the bases it spans in place, so an interior junction has a meaningful position and the pair is genuinely ambiguous. A blanket "spans never conflict with interior insertions" would pass every other assertion in the new module and break this one, so it has a dedicated test.

## Verification

- `cargo nextest run --features dev` with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1` and `FERRO_SWEEP_SEEDS=full`: **9354 passed, 0 failed**, no existing expectation changed.
- New `tests/it/issue_1437_dup_read_span_interior_sibling.rs`: five tests — all four siblings accepted together (asserted as one loop, so the property under test is the *agreement* rather than four independent verdicts), the three substitution answers pinned literally as the stability half, `dup`/`repeat` parity, the `inv` discriminator, and branch (a) still firing.

### DIFF
--- a/src/normalize/overlap.rs
+++ b/src/normalize/overlap.rs
@@ -236,6 +236,40 @@ fn junction_overlaps(
                     gap: end,
                     kind,
                 });
+                // …and *only* as a junction occupant (#1437). A `dup`/`repeat`
+                // **reads** its span and **writes** at the junction 3' of it
+                // (`duplication.md:5`), so registering the read span here would
+                // report a sibling that collides with nothing.
+                //
+                // That is the reading #1411 corrected in `detect_overlap_conflicts`
+                // — key on what a member writes, not on what it reads — and
+                // leaving branch (b) on the old reading made the two detectors
+                // disagree. Measured on the 24 bp `NC_TEST.1` fixture, strict
+                // mode, every sibling strictly interior to the `dup`'s read span
+                // and disjoint from its write at `9|10`:
+                //
+                //     g.[4_9dup;6T>C]     ACCEPTED   g.5_6insCTTTTT
+                //     g.[4_9dup;8T>C]     ACCEPTED   g.7_8insCTTTTT
+                //     g.[4_9dup;9T>C]     ACCEPTED   g.8_9insCTTTTT
+                //     g.[4_9dup;6_7insT]  REJECTED   W5002 (dup, ins)
+                //
+                // The verdict turned on the sibling's edit *kind*, not on whether
+                // anything collided: a substitution is not an `NaEdit::Insertion`,
+                // so branch (b) never examined it.
+                //
+                // Accepting is also the cheap direction for representation
+                // stability, in #1411's sense: the rejected shapes have no shipped
+                // normalized form, while the three accepted ones do
+                // (`g.5_6insCTTTTT` and friends), so resolving the other way would
+                // move output that consumers already hold.
+                //
+                // `repeat` goes with `dup`, not just `dup` alone: the per-member
+                // pipeline picks between the two spellings by axis (CDS gives
+                // `c.5_6dup`, genomic gives `g.1005_1006A[4]`), so exempting one
+                // and not the other would make this detector spelling-sensitive —
+                // the very defect the junction registration above exists to
+                // prevent.
+                continue;
             }
             spans.push(Span {
                 idx,
--- a/tests/it/issue_1437_dup_read_span_interior_sibling.rs
+++ b/tests/it/issue_1437_dup_read_span_interior_sibling.rs
@@ -1,0 +1,153 @@
+//! #1437 — `detect_insertion_overlaps` keyed a `dup` on the span it **reads**,
+//! so a sibling interior to that span was judged by its own edit *kind* rather
+//! than by whether anything actually collided.
+//!
+//! #1411 established the principle: a conflict keys on what a member **writes**,
+//! not on what it reads. A `dup` writes its copy directly 3' of the original
+//! (`duplication.md:5`), so `g.4_9dup` writes at the junction `9|10` and reads
+//! bases 4-9 without touching them. Fix 1(a) applied that to
+//! `detect_overlap_conflicts`; branch (b) of `detect_insertion_overlaps` kept
+//! the old reading, so the two detectors disagreed.
+//!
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

PR #1447, opened 2026-08-06
docs: correct two normalizer comments that name the wrong mechanism

### LINKED ISSUE #1435: The #1392 control test's docstring names a gate it cannot reach
`a_non_conflicting_allele_of_the_same_shape_still_merges` (`tests/it/strict_rejection_survives_normalization.rs:161-186`) is a good test with a docstring that names the wrong mechanism.

Its assertions are correct and worth keeping: `g.[23dup;23A>G]` is not a conflict, and lenient normalization must still merge it to `g.22_23insG`. That part holds.

What does not hold is the claim about *what it guards*. The docstring says:

> The gate must key on the conflict, not on the shape it was found in.

and the module header at the top of the same file says the control is what would notice "a gate keyed on the shape rather than on the conflict". The gate both refer to is `sequence_first_pass`'s `detect_overlap_conflicts` check. That gate cannot see this input's two raw members: `sequence_first_pass` runs *after* the legacy per-member collapse, so by the time it is reached `g.[23dup;23A>G]` has already been merged into a single member. A gate erroneously keyed on the `dup`-plus-substitution *shape* would therefore have nothing to key on here, and this test would pass anyway.

So the test cannot fail for the reason its docstring gives. It is a real regression guard for the merge outcome; it is not a control on the gate's scoping.

Worth fixing rather than deleting, because the docstring is load-bearing in the wrong direction — it tells a future reader that gate scoping is already covered, which will stop someone writing the control that actually would cover it.

### Suggested resolution

- Re-describe the test as what it is: a pin that a non-conflicting `dup`-plus-substitution pair still collapses to one member through the legacy collapse path.
- Correct the module header's "The control" paragraph to match.
- If a genuine control on the gate's conflict-vs-shape scoping is wanted, it needs an input that actually reaches `sequence_first_pass` with two members intact.

### Provenance

Found while reviewing #1411, whose analysis established the ordering (`sequence_first_pass` runs after the legacy collapse). Deferred from that PR as out of scope.

### LINKED ISSUE #1436: collapse_overlapping_cis_edits' doc comment omits dup, which it does handle
`collapse_overlapping_cis_edits`' doc comment (`src/normalize/merge.rs:336-350`) enumerates the edit kinds it accepts and omits `dup`:

> * all sub-variants are same-accession genomic `NaEdit`s with simple (certain, non-special, non-offset) positions, and one of **ins/del/sub/delins** (literal payloads only);

The body handles `Dup` throughout. From the same function:

```rust
let is_insertion_like = |e: &GEdit| matches!(e, GEdit::Ins { .. } | GEdit::Dup { .. });
```

plus four further `GEdit::Dup` match arms in the span/gap computations. A `dup` is treated as insertion-like — i.e. as one of the members that can attach to the del/sub/delins interval — which is precisely the role the doc's list is describing.

The nearby prose is also now inconsistent with itself: the closing paragraph says "pure-insertion groups, pure-deletion groups ... are unaffected", and a reader who believes the kind list will not realise a `dup`-plus-deletion group is in scope.

An outdated comment here is worse than none, because the list reads as an exhaustive precondition and someone reasoning about whether a `dup` can reach this collapse will conclude it cannot.

### Suggested resolution

Add `dup` to the kind list and to the insertion-attachment bullet, and check the closing "unaffected" paragraph still reads true once it is there.

### Provenance

Noticed while tracing `collapse_overlapping_cis_edits` stage by stage during #1411's review (where the conclusion was that the function is a no-op on every row of that PR's table). Deferred as out of scope.

nh13:
Closes #1435. Closes #1436.

Both were raised during #1411's review and deferred as out of scope. Neither changes behaviour. They are grouped because they are the same failure — a comment that names a mechanism it does not have — and both were load-bearing in the wrong direction.

## #1435 — a "control" that cannot fail for the reason it gives

`a_non_conflicting_allele_of_the_same_shape_still_merges` claimed to guard `sequence_first_pass`'s gate:

> The gate must key on the conflict, not on the shape it was found in.

and the module header called it the thing that would notice "a gate keyed on the shape rather than on the conflict".

It cannot notice. The legacy per-member collapse runs **before** `sequence_first_pass`, so `g.[23dup;23A>G]` reaches the gate as the single member `g.22_23insG`; the gate's `members.len() > 1` precondition is false and it never runs. A gate erroneously keyed on the `dup`-plus-substitution shape would have nothing left to key on, and this test would pass anyway.

**Measured rather than argued**, since the issue's claim is structural and worth checking before rewriting docs on its strength:

- Instrumenting the gate to report its member count prints `SEQFIRST sees 1 member(s)` for this input, on both the strict and lenient calls.
- Disabling the gate entirely (`detect_overlap_conflicts` check short-circuited) leaves the test passing, byte-identically. A test that guards a gate cannot be indifferent to the gate's existence.

Corrected rather than deleted: the assertions are a real regression guard for the merge outcome and worth keeping, but the wrong docstring told a reader that gate scoping was already covered — which is exactly what stops someone writing the control that would cover it. Both the docstring and the module header now say what the test does guard, and the header states what a genuine control needs: **an input that reaches `sequence_first_pass` with two members still intact.**

## #1436 — an "exhaustive" kind list missing a kind

`collapse_overlapping_cis_edits`' doc comment enumerated its accepted edit kinds as `ins/del/sub/delins` and omitted `dup`. The body treats `dup` as insertion-like throughout:

```rust
let is_insertion_like = |e: &GEdit| matches!(e, GEdit::Ins { .. } | GEdit::Dup { .. });
```

plus four further `GEdit::Dup` arms in the span and gap computations — i.e. exactly the role the list is describing: a member that attaches to the del/sub/delins interval rather than forming one. That is the same read/write distinction #1411 drew, since a `dup` writes at the junction 3' of the run it reads.

The list reads as an exhaustive precondition, so someone reasoning about whether a `dup` can reach this collapse would conclude it cannot. `dup` is now in the kind list and in the attachment bullet, the two bullets say "insertion-like" where they mean it rather than "insertion", and the closing "unaffected" paragraph is checked against the addition — pure-`dup` groups are still unaffected, because they have no del/sub/delins to attach to.

## Verification

- `cargo nextest run --features dev` with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1` and `FERRO_SWEEP_SEEDS=full`: **9366 passed, 0 failed**.
- `cargo fmt --all --check` clean.
- The instrumentation used for the #1435 measurement was temporary and is not in this diff.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -338,16 +338,34 @@ pub(crate) fn has_same_gap_insertions(variants: &[HgvsVariant], phase: AllelePha
 /// group:
 ///   * all sub-variants are same-accession genomic `NaEdit`s with simple
 ///     (certain, non-special, non-offset) positions, and one of
-///     ins/del/sub/delins (literal payloads only);
+///     ins/**dup**/del/sub/delins, and where the edit carries a payload at all
+///     that payload is literal (a `dup` carries none — it copies the range it
+///     names from the reference);
 ///   * the del/sub/delins spans form one contiguous interval with no
 ///     unchanged interior base (no holes);
-///   * every insertion attaches to that interval (its gap lies in
+///   * every insertion-like member attaches to that interval (its gap lies in
 ///     `[c_lo - 1, c_hi]`);
-///   * the group has BOTH an insertion and a del/sub/delins.
+///   * the group has BOTH an insertion-like member and a del/sub/delins.
 ///
-/// Anything else passes through untouched, so pure-insertion groups,
-/// pure-deletion groups (owned by `merge_consecutive_edits`), and
-/// non-overlapping inputs are unaffected.
+/// **`dup` is insertion-like here, not a span (#1436.)** The list above used to
+/// omit it, which read as an exhaustive precondition and told anyone tracing
+/// whether a `dup` can reach this collapse that it cannot. It can: the body
+/// treats it as one of the members that attaches to the del/sub/delins interval
+/// —
+///
+/// ```ignore
+/// let is_insertion_like = |e: &GEdit| matches!(e, GEdit::Ins { .. } | GEdit::Dup { .. });
+/// ```
+///
+/// — plus four further `GEdit::Dup` arms in the span and gap computations. That
+/// is the same read/write distinction #1411 drew: a `dup` writes at the junction
+/// 3' of the run it reads, so it attaches to an interval rather than forming
+/// one.
+///
+/// Anything else passes through untouched, so pure-insertion groups (including
+/// pure-`dup` ones, which have no del/sub/delins to attach to), pure-deletion
+/// groups (owned by `merge_consecutive_edits`), and non-overlapping inputs are
+/// unaffected.
 pub(crate) fn collapse_overlapping_cis_edits<P: ReferenceProvider>(
     variants: Vec<HgvsVariant>,
     phase: AllelePhase,
--- a/tests/it/strict_rejection_survives_normalization.rs
+++ b/tests/it/strict_rejection_survives_normalization.rs
@@ -32,15 +32,32 @@
 //! same bases, and strict-**acceptable**. That is #1307, and it is the shape the
 //! rows below start from.
 //!
-//! ## The control
+//! ## The companion row, and what it does *not* guard
 //!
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai pause

PR #1451, opened 2026-08-06
fix(normalize): report span edits whose spans intersect without coinciding

### LINKED ISSUE #1448: Normalizer: strict mode accepts partially-overlapping span edits, which denote no sequence at all
## What's wrong

`detect_overlap_conflicts` groups sub-variants by **exact** `(accession, coord_system, region, start, end)` coincidence. Two span edits that *partially* overlap — or that *nest* — share no key, so no `OverlapConflict` is reported and strict mode accepts them. They denote no sequence: the independent applier declines them, in either member order.

Measured on `origin/main` at `699fb592`, reference `ACGTACGTTTTACGTACGTGGGCCCACGT`:

```
g.[10_14del;12_16del]        strict: ACCEPT   applies: NO    -> g.[11_15del;12_16del]
g.[10_14delinsAA;12_16del]   strict: ACCEPT   applies: NO    -> unchanged
g.[10_14inv;12_16inv]        strict: ACCEPT   applies: NO    -> unchanged
g.[10_14del;10_16del]        strict: ACCEPT   applies: NO    -> g.[10_16del;11_15del]   (nested!)
```

The lenient output is equally meaningless, and in two of the four the normalizer *shifts* a member while leaving the pair overlapping.

Contrast a sibling that **is** caught — a substitution strictly inside a deletion:

```
g.[10_14del;12C>G]           strict: REJECT
g.[10_14inv;12C>G]           strict: REJECT
```

So the verdict currently depends on the members' relative geometry in a way that does not track whether anything actually collides.

## Scale

Enumerating two-member cis alleles over four sequences, member widths 1-3, at separations 0-4, restricted to span edits:

| measure | count |
|---|---|
| enumerated | 31,110 |
| strict-**accepted** | 25,848 |
| **accepted but denoting no sequence** | **10,499** |
| families | 9 |

**41% of everything strict mode accepts in this shape class denotes nothing.**

```
 1193x g.[_inv;_inv]         e.g. g.[2_3inv;2_4inv]
 1191x g.[_inv;_del]         e.g. g.[2_3inv;2_4del]
 1174x g.[_del;_inv]         e.g. g.[2_3del;2_4inv]
 1174x g.[_delinsAA;_inv]    e.g. g.[2_3delinsAA;2_4inv]
 1171x g.[_delinsAA;_del]    e.g. g.[2_3delinsAA;2_4del]
 1170x g.[_inv;_delinsAA]    e.g. g.[2_3inv;2_4delinsAA]
 1154x g.[_del;_delinsAA]    e.g. g.[2_3del;2_4delinsAA]
 1137x g.[_delinsAA;_delinsAA]
 1135x g.[_del;_del]
```

Every family is a combination of `del`/`delins`/`inv` — the edits whose write footprint **is** their span. `dup` never appears, which is the expected result of #1411/#1437: a `dup` writes at the junction 3' of the run it reads, so two overlapping `dup` spans genuinely do compose.

- `detect_overlap_conflicts` keys on exact coincident bounds, so `10_14` and `12_16` are different groups.
- `detect_insertion_overlaps` requires at least one junction-anchored member; neither of these is one.
- None of the three seam oracles can see it. `FERRO_ASSERT_REPARSE` passes (both spellings parse), `FERRO_ASSERT_IN_BOUNDS` passes (every coordinate is in range), and `FERRO_ASSERT_IDEMPOTENT` passes whenever the output is a fixed point — which three of the four above are. The applier is not wired in as an oracle, so "denotes no sequence" is invisible to CI.

This is the same defect class as #395 and #1276, reached by a geometry the detector does not examine.

## Suggested resolution

Report a conflict when two members' **write footprints intersect**, not only when they coincide exactly. That is the principle #1411 established and #1437 extended, and it predicts the `dup` exemption above rather than having to special-case it:

- `del` / `delins` / `sub` / `inv` — the write footprint is the span;
- `ins` — a zero-width junction (already handled by `detect_insertion_overlaps`);
- `dup` / `repeat` — a zero-width junction 3' of the read span (#1437).

**Stability note.** This *widens* W5002, so inputs accepted today become rejected. That is a behaviour change and should be quantified before merging — but the affected inputs are exactly those whose accepted output denotes no sequence, so no consumer can be holding a meaningful normalized string for them. Whether the current lenient output for these should also change (today it is sometimes shifted, sometimes returned as authored) is a separate question.

Worth checking against the confluence guards before landing: widening a conflict notion cost confluence once already (a gate keyed on the junction detectors made `g.[263_264insAC;264_265insAA]` stop merging, caught by `cis_spelling_confluence_gap`). Partially-overlapping *spans* are a different class from same-junction insertions, but the check is cheap and the precedent is recent.

## Reproducing

Any worktree at current `main`, in a module with access to `common::cis_apply_oracle`:

```rust
let seq = "ACGTACGTTTTACGTACGTGGGCCCACGT";
for input in ["TEMPLATE:g.[10_14del;12_16del]", "TEMPLATE:g.[10_14del;10_16del]"] {
    // strict accepts...
    assert!(Normalizer::with_config(provider(seq), NormalizeConfig::strict())
        .normalize(&parse_hgvs(input).unwrap()).is_ok());
    // ...but the description denotes nothing.
    assert!(apply(seq, input).is_none());
}
```

nh13:
Closes #1448.

## The defect

`detect_overlap_conflicts` grouped sub-variants by **exact** `(accession, coord_system, region, start, end)` equality. Two span edits that overlap only *partially* — or that *nest* — share no key, so nothing reported them and strict mode accepted them. They denote no sequence at all; the independent applier declines them in either member order.

Measured on `origin/main` at `699fb592`, reference `ACGTACGTTTTACGTACGTGGGCCCACGT`:

```
g.[10_14del;12_16del]        strict: ACCEPT   applies: NO   -> g.[11_15del;12_16del]
g.[10_14delinsAA;12_16del]   strict: ACCEPT   applies: NO   -> unchanged
g.[10_14inv;12_16inv]        strict: ACCEPT   applies: NO   -> unchanged
g.[10_14del;10_16del]        strict: ACCEPT   applies: NO   -> g.[10_16del;11_15del]   (nested)
```

Note the first and last: the normalizer *shifted* a member while leaving the pair overlapping — repairing a member of an allele it should have been preserving verbatim (#395).

## Scale

Enumerating two-member cis alleles over four sequences, member widths 1-3, separations 0-4, restricted to span edits:

| measure | `main` | this PR |
|---|---|---|
| enumerated | 31,110 | 31,110 |
| strict-accepted | 25,848 | 15,349 |
| **accepted but denoting no sequence** | **10,499** | **0** |
| families | 9 | 0 |

**41% of everything strict mode accepted in this shape class denoted nothing.**

## Why nothing caught it

- `detect_overlap_conflicts` keys on exact coincidence, so `10_14` and `12_16` are different groups.
- `detect_insertion_overlaps` needs a junction-anchored member; neither of these is one.
- **None of the three seam oracles can see it.** `FERRO_ASSERT_REPARSE` passes (both spellings parse), `FERRO_ASSERT_IN_BOUNDS` passes (every coordinate is in range), and `FERRO_ASSERT_IDEMPOTENT` passes whenever the output is a fixed point — which most of these are. The applier is not wired in as an oracle, so "denotes no sequence" was invisible to CI.

## The fix

Report a conflict when two members' spans **intersect**, not only when they coincide.

**Restricted to edits whose write footprint *is* their span** — `sub`, `del`, `delins`, `inv`. That is #1411's principle, and it is what makes the `dup` behaviour fall out instead of needing a special case: a `dup`/`repeat` reads its span and writes at the junction 3' of it, so two overlapping `dup` spans genuinely compose. That prediction is borne out by the census — no `dup` family appears in any of the nine. `detect_insertion_overlaps` owns the junction geometry.

Pairwise rather than grouped, and skipping equal-bound pairs, so an exactly-coincident pair is still reported **once** by the existing grouped loop (which names every member of its group) rather than twice. Pinned by a test.

## Representation stability

This **widens W5002**: 10,499 inputs move from accepted to rejected. Stated plainly, since that is a shipped-behaviour change.

What makes it the cheap direction is that every affected input's accepted output denoted no sequence, so no consumer can be holding a meaningful normalized string for one. And the lenient output for two of the four measured shapes was *worse* than meaningless — it was a shifted member of an allele the pipeline should have been preserving. Those now take the verbatim path, which is what #395 specifies.

### The committed harness reports zero, and that result is uninformative here

Run per #1441/#1442, base `699fb592` against this branch:

| | rows |
|---|---|
| total | 18,432 |
| **moved** | **0 (0.0%)** |
| previously accepted (a migration) | 0 |
| previously not a fixed point (free) | 0 |

**Do not read that as "no blast radius".** `examples/dump_normalized_corpus.rs` emits five
families — `adjacent_junction_ins`, `del_plus_sub`, `dup_plus_ins`, `dup_plus_sub`,
`split_vs_spanning_delins` — and **every one of them places its members disjointly**:
`split_vs_spanning_delins` is `[s_s+1delinsGGG; s+3_s+4delinsGG]` with an unchanged base at
`s+2`, `del_plus_sub` puts its substitution three positions past the deletion, and so on. Two
span edits whose spans *intersect* — this PR's entire subject — do not occur anywhere in that
corpus. The zero is real, and it measures a population this change cannot touch.

So the load-bearing measurement for this PR remains the one in the table above (10,499 → 0 over
31,110 enumerated two-member span alleles), plus a green suite with no existing expectation
moved. The harness result is included because it was run and should be on the record, not
because it is evidence.

That gap is worth closing separately: overlapping and nested span members are the shape class
#1406, #1437 and #1448 all live in, and the harness cannot currently see any of them.

## Verification

- `cargo nextest run --features dev` with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1` and `FERRO_SWEEP_SEEDS=full`: **9371 passed, 0 failed**, with **no existing expectation changed**.
- New `tests/it/issue_1448_partial_span_overlap.rs`: the headline stated as an invariant (strict must never accept a description the applier declines, checked per input rather than pinned as four verdicts), the verbatim-not-shifted property, disjoint members still accepted, and single-reporting of coincident pairs.
- New unit test in `src/normalize/overlap.rs`: overlapping `dup` spans are not a coincident-bounds conflict. Asserted there rather than end to end because `detect_insertion_overlaps` reports an overlapping `dup` pair for its own separate reasons on `main` (#1437, fixed by #1446), so an integration assertion would be measuring the other detector.

---

## Representation movement — it is not only a strict-mode change

**This section was added after #1445 merged, which this PR predates.** #1445 hands an overlap-conflicting allele back as authored when the pipeline would otherwise launder the conflict away. So every conflict this PR newly *detects* is now also an allele that is newly *preserved* — and the lenient output moves, not just the strict verdict.

Measured on this PR's own fixture sequence, against `main` at `a52fef26`:

| input | `main` | this PR |
|---|---|---|
| `g.[10_14del;12_16del]` | `g.[11_15del;13_17del]` | preserved as authored |
| `g.[10_14del;10_16del]` | `g.[10_16del;11_15del]` | preserved as authored |
| `g.[10_14inv;12_16inv]` | `g.[11_13inv;12_16inv]` | preserved as authored |

Three of three. The bound on how many rows this reaches is the census already reported above — the newly-detected conflicts, 10,499 of the 25,848 two-member span alleles enumerated — since detection is what triggers preservation.

**Every one of those rows is cheap in the stability sense**, which is the reason to land it rather than defer it: the outputs that move are the ones the applier declines outright. `main` shifts them into a well-formed description of a sequence *neither member denotes*, so no consumer holds a correct key for them today. Moving from a wrong answer to a preserved input costs nothing that was right.

**The committed corpus reports 0 of 18,432 moved, and that is not evidence here.** `dump_normalized_corpus` builds no overlap-conflicting alleles at all, so it cannot see this class — filed as #1456. The zero means no *collateral* movement outside the conflict families, which is worth having but is not the measurement this section is about.

### DIFF
--- a/src/normalize/overlap.rs
+++ b/src/normalize/overlap.rs
@@ -88,7 +88,117 @@ pub(crate) fn detect_overlap_conflicts(
             edit_kinds,
         });
     }
+
+    // Spans that *intersect* without coinciding (#1448).
+    //
+    // The grouping above keys on exact `(accession, coord_system, region, start,
+    // end)` equality, so two members that overlap only partially — or that nest —
+    // land in different groups and were reported by nothing. Strict mode
+    // therefore accepted them, and they denote no sequence at all: the
+    // independent applier declines them in either member order.
+    //
+    //     g.[10_14del;12_16del]        ACCEPTED, applies to nothing
+    //     g.[10_14inv;12_16inv]        ACCEPTED, applies to nothing
+    //     g.[10_14del;10_16del]        ACCEPTED, applies to nothing (nested)
+    //
+    // Measured over two-member span-edit alleles on four sequences: 10,499 of
+    // the 25,848 strict mode accepted — 41% — denoted nothing.
+    //
+    // **Restricted to edits whose write footprint *is* their span**, which is
+    // the #1411 principle and what makes the `dup` behaviour fall out rather
+    // than need a special case: a `dup`/`repeat` reads its span and writes at
+    // the junction 3' of it, so two overlapping `dup` spans genuinely do
+    // compose, and none appeared in the measured families.
+    // `detect_insertion_overlaps` owns the junction geometry.
+    //
+    // Pairwise rather than grouped, and skipping pairs whose keys are equal, so
+    // an exactly-coincident pair is reported once — by the loop above, which
+    // names every member of its group — rather than twice.
+    for (i, first) in variants.iter().enumerate() {
+        let Some((coord_system, region, first_start, first_end)) = span_writer_footprint(first)
+        else {
+            continue;
+        };
+        let Some(first_accession) = first.accession().map(|a| a.to_string()) else {
+            continue;
+        };
+        for (j, second) in variants.iter().enumerate().skip(i + 1) {
+            let Some((second_system, second_region, second_start, second_end)) =
+                span_writer_footprint(second)
+            else {
+                continue;
+            };
+            let Some(second_accession) = second.accession().map(|a| a.to_string()) else {
+                continue;
+            };
+            if first_accession != second_accession
+                || coord_system != second_system
+                || region != second_region
+            {
+                continue;
+            }
+            // Coincident pairs belong to the grouped loop above.
+            if first_start == second_start && first_end == second_end {
+                continue;
+            }
+            // A reversed span is not an interval, so do not test it as one.
+            // SVD-WG006 admits `<high>_<low>` for a circular deletion or
+            // duplication, and `genome_range` passes those through with
+            // `start > end` — nothing upstream reorders them. Read as an
+            // ordinary interval, `m.16569_5del` claims to end before it begins,
+            // so the test below would decide the pair by accident rather than
+            // by geometry. Declining to judge it is the conservative answer and
+            // the same one #1423 reached for the containment test next door: a
+            // missed report is a verdict this pass simply does not make, while a
+            // wrong one is a rejection the caller cannot explain.
+            let reversed = first_start > first_end || second_start > second_end;
+            // Closed intervals: they intersect when neither ends before the
+            // other begins.
+            if reversed || first_end < second_start || second_end < first_start {
+                continue;
+            }
+            let kinds: Vec<String> = [i, j]
+                .iter()
+                .filter_map(|&idx| edit_kind(&variants[idx]).map(|s| s.to_string()))
+                .collect();
+            warnings.push(NormalizationWarning::OverlapConflict {
+                accession: first_accession.clone(),
+                coordinate_system: coord_system.to_string(),
+                location: location_for_variant(first)
+                    .expect("span_writer_footprint established a renderable location"),
+                edit_kinds: kinds,
+            });
+        }
+    }
     warnings
+}
+
+/// The reference span an edit **writes over**, for the edits whose write
+/// footprint is their own span.
+///
+/// `None` for anything else, which is what keeps the intersection test above
+/// narrow:
+///
+/// - an **insertion** writes at a zero-width junction, not over a span;
+/// - a **`dup`/`repeat`** reads its span and writes at the junction 3' of it
+///   (`duplication.md:5`), so two overlapping `dup` spans compose uniquely —
+///   the #1411 principle, and the reason no `dup` family appears in #1448's
+///   measured census.
+///
+/// Both of those geometries belong to [`detect_insertion_overlaps`].
+fn span_writer_footprint(variant: &HgvsVariant) -> Option<(&'static str, Region, i64, i64)> {
+    let edit = inner_edit(variant)?;
+    if !matches!(
+        edit,
+        NaEdit::Substitution { .. }
+            | NaEdit::SubstitutionNoRef { .. }
+            | NaEdit::Deletion { .. }
+            | NaEdit::Delins { .. }
+            | NaEdit::Inversion { .. }
+    ) {
+        return None;
+    }
+    simple_span(variant)
 }
 
 /// Detect overlaps that involve at least one junction-anchored edit within a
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

Issue #1420, opened 2026-08-05
Normalizer: multi-member cis-alleles normalized independently, not re-derived from the resulting sequence — `[dup;del]`, `[ins;del]`, `[delins;del]` each still non-confluent (follow-up to #1231/#1233/#1232)


msto:
> [!NOTE]
> Authorship: the content below was drafted by GitHub Copilot CLI (an AI coding
> agent) and filed via gh under @msto's account. It reflects the agent's
> analysis, not a statement authored by @msto.

## Summary

`Normalizer.normalize` normalizes a multi-member cis-allele by shifting/typing each
member **independently**, rather than re-deriving the edit set from the **resulting
sequence**. So a variant given as a multi-member cis-allele and the *same* variant
given as a single span normalize to two different stable strings. #1231
(`[dup;del]`), #1233 (`[ins;del]`) and #1232 (`[delins;del]`/spanning) were closed by
per-example fixes (#1237 + pins), but the underlying defect persists: below are three
still-non-confluent instances, one per member type, **verified on v0.12.0 and on
`main` (`af249d2`)**.

Filing these as **one** issue (rather than reopening three) because they are a single
defect — the root cause tracked in **#1235**. The tell is that the required
correction points in **opposite directions** depending on member type, so no local
per-type rule fixes them; only re-deriving from the resulting sequence does:

| # | member type | design (kept) | canonical (from the span) | direction |
|---|---|---|---|---|
| v2 | `[dup;del]` | `[37dup;41del]` | `[38T>A;40_41delinsTG]` | **reduce** (expose substitution) |
| v3 | `[ins;del]` | `[36_37insC;40del]` | `[37_38delinsCA;40G>T]` | **reduce** (expose substitution) |
| v4 | `[delins;del]` | `[21delinsGC;24del]` | `21_24delinsGCTG` | **coalesce** (merge consecutive changes) |

All three: both encodings idempotent, `EquivalenceChecker` = `SequenceMatch`, no
error/warning.

## v2 — `[dup;del]` not reduced (same class as #1231)

```
TEMPLATE:g.[37dup;41del]      -> TEMPLATE:g.[37dup;41del]          # kept as dup + del
TEMPLATE:g.38_41delinsATTG    -> TEMPLATE:g.[38T>A;40_41delinsTG]  # same variant, re-derived
```

Positions 38 and 40–41 change (39 unchanged). **prioritisation** (`general.md:56`:
[...]

Issue #200, opened 2026-05-14
Support variant-level projection: g. → c./p. descriptions


msto:
# Support variant-level projection: g. → c./p. descriptions

## Summary

Introduce a variant-level projection API that takes a genomic (g.) variant and produces equivalent coding (c.) and protein (p.) HGVS description strings, along with metadata like frameshift status.

This would make ferro a drop-in replacement for mutalyzer, whose `normalize()` returns g., c., and p. descriptions in a single call.

## Motivation

ferro currently operates entirely within a single coordinate system — `Normalizer.normalize()` takes a g. variant and returns a normalized g. string, which is the right scope for normalization. However, there isn't currently a way to take a g. variant and obtain its c. or p. equivalent as a complete HGVS description string. `CoordinateMapper` provides the numeric position mapping (e.g. `g_to_c()` → `(cds_pos, offset, is_utr3)`), but the step from mapped positions to formatted c./p. variant strings — attaching the edit, looking up codons, translating to amino acids, applying HGVS protein nomenclature — is left to the consumer. That orchestration is non-trivial, especially for protein consequences.

ferro already has all the building blocks for this internally (see below). Exposing a variant-level projection API would let consumers adopt ferro for end-to-end g. → c. → p. workflows without reimplementing that orchestration.

## Existing building blocks

ferro already has the pieces for this:

| Component | Location | What it does |
|-----------|----------|--------------|
| `Projector` | `src/data/projection.rs` | Maps genomic positions → `TranscriptProjection` with CDS position, gene symbol, MANE status |
| `TranscriptProjection::to_hgvs_string()` | `src/data/projection.rs` | Formats the c./n. *position* (not yet including the edit) |
| `CoordinateMapper` | `src/convert/mapper.rs` | `genomic_to_cds()`, `cds_to_protein()`, `translate_codon()`, `get_amino_acid_at()` |
| `Normalizer` | `src/normalize/mod.rs` | Normalization with 3'/5' shifting |
| `HgvsVariant` | `src/hgvs/variant.rs` | Typed variant representation with `Display` for formatting |
| `EffectPredictor` | exposed via Python | Classifies amino acid changes, indels, splice variants |
| `HgvsVariant::is_frameshift()` | exposed via Python | Frameshift detection |

What could be added is **variant-level orchestration**: composing position mapping + edit transformation + amino acid lookup into complete c. and p. HGVS strings.

### Rust

```rust
pub struct VariantProjection {
[...]

Issue #179, opened 2026-05-12
Inversion detection across self-complementary internal bases (#81 A10 follow-up)


msto:
## Context

The `inv` recognition work in #160 / #166 (A10 inv branch) handles the case where a `delins` span is the reverse complement of the reference. In allele normalization, the merge-consecutive-edits step (landed in #80) combines adjacent substitutions into a `delins` before the inversion check fires.

However, there is a class of inversions that the current pipeline does not detect: those where the inverted range contains internal bases that are self-complementary (palindromic under reverse-complement) and therefore appear unchanged. Because the individual substitutions are separated by ≥1 unchanged nucleotide, the merge step does not combine them, and the downstream `inv` check never sees a single `delins` to test.

## Spec basis

`inversion.md` defines an inversion as *"a sequence change where, compared to a reference sequence, **more than one nucleotide** replacing the original sequence are the reverse complement of the original sequence."*

`general.md` prioritisation: `substitution > deletion > inversion > duplication > insertion`. And `delins.md` defines deletion-insertion as a change that *"is **not** a substitution, inversion or conversion"* — i.e., the residual form after higher-priority types have been excluded.

When the observed sequence across a range is the reverse complement of the reference, the spec indicates the `inv` form is canonical, even when some interior positions did not change (because those bases are self-complementary within the inversion).

## Examples

Examples use a synthetic genomic reference to illustrate the pattern. Positions and bases are representative.

**Example 1** — two substitutions spanning self-complementary interior

```
Input:     REF:g.[1009T>G;1040T>G;1043C>A;1051T>G;1052A>G]
Expected:  REF:g.[1009T>G;1040_1043inv;1051_1052delinsGG]
Ferro:     REF:g.[1009T>G;1040T>G;1043C>A;1051_1052delinsGG]
```

Reference at 1040–1043: `TTAC`; reverse complement: `GTAA`.
Observed at 1040–1043: `GTAA` (1040 T→G, 1041 T unchanged, 1042 A unchanged, 1043 C→A).
Interior bases 1041–1042 (`TA`) are self-complementary under rev-comp (`TA` → `TA`).

**Example 2** — adjacent + gap pattern

```
Input:     REF:g.[1017C>G;1018T>G;1021C>A;1037C>A;1154A>C]
Expected:  REF:g.[1017C>G;1018_1021inv;1037C>A;1154A>C]
[...]

Issue #87, opened 2026-05-01
tracking: HGVS v21.0 spec-compliance gaps


msto:
Tracking issue for HGVS v21.0 conformance items surfaced by agent review beyond #81 and #83. Each item names a spec rule, ✓/✗ examples, current behavior, and an entry-point for the fix.

## Insertion form (`DNA/insertion.md:15`, `checklist.md` item 3)

- [ ] **Two adjacent flanking positions required.** Spec: *"positions_flanking should contain two flanking nucleotides, e.g., 123_124, not 123_125"* (insertion.md:15); single-position form is rejected as ambiguous (insertion.md:96-101).
  - ✓ `c.123_124insG`, `g.32867861_32867862insT`
  - ✗ `c.123insG` (single pos), `c.123_125insATG` (non-adjacent)
  - Currently: both round-trip through parse. `c.123insG` then panics in `Normalizer::normalize` at `normalize/shuffle.rs:86` because `(new_end - start) % alt_seq.len()` underflows when `new_end < start`.
  - Plan: in `src/hgvs/parser/edit.rs` (insertion construction), require the location to span exactly two adjacent positions; reject otherwise via `error_handling`. Independently, in `shuffle.rs:86`, check `new_end >= start` and return `Err(...)` instead of subtracting.

## Allele descriptions (`DNA/alleles.md`)

- [ ] **§23 — single coordinate-system prefix per allele description.** Spec: *"descriptions combining variants based on different reference sequence types ... should not be used."*
  - ✓ `[NM_…:c.76A>C];[NM_…:c.200G>A]`
  - ✗ `[NM_…:c.76A>C];[NC_…:g.10091C>G]`
  - Currently: round-trips through parse and normalize.
  - Plan: in `AlleleVariant` construction (`src/hgvs/parser/variant.rs`), reject when constituent variants don't share the same coordinate-system kind (`c.`/`g.`/`n.`/`r.`/`p.`/`m.`/`o.`).

- [ ] **§54 — compact homozygous forms not allowed.** Spec: *"it is not allowed to shorten this to `c.2376[G>C];[G>C]` or even `c.2376G>C[];[]`."*
  - ✗ `c.2376[G>C];[G>C]`, `c.2376G>C[];[]`
  - ✓ `c.[2376G>C];[2376G>C]`
[...]

PR #1457, opened 2026-08-06
test: harvest authored spec cases from the issue tracker into runnable code

nh13:
A body of concrete HGVS cases exists only as prose in issue bodies — spec ✓/✗ pairs, diagnostic probe tables, worked examples. Filed is not the same as guarded: nothing re-runs them, so behaviour drifts silently and nobody can say which items were fixed and which are still open without re-deriving each by hand. That turned out to matter — several of these were fixed years ago and several are still broken, and running them was the only way to tell.

This adds `tests/it/msto_authored_cases.rs`: **46 rows, every one traceable to a specific issue**, asserting **what ferro does today** rather than what the issue wanted. Nothing here is aspirational and nothing is red.

## Buckets

| bucket | count |
|---|---|
| **Guard** — behaviour matches the cited spec line / the issue's expected answer | **38** |
| **Gap** — behaviour differs; the row pins today's answer, the comment names the wanted one | **8** |
| **Ignored** | **0** |

Rows carry an explicit `Verdict`, and two censuses (`ISSUE_87_OPEN_GAPS`, `ISSUE_1421_ROWS_THAT_SPLIT`) count them, so a fix has to update a row *and* a census deliberately. Pinning a known-wrong answer on purpose is the point: it makes that answer visible in a diff the day somebody changes it, which prose in a closed issue cannot do.

## What running them found

**#87 (`tracking: HGVS v21.0 spec-compliance gaps`, closed as subsumed) is mostly closed — 35 of 38 rows now comply.** Single-position (`c.123insG`) and non-adjacent (`c.123_125insATG`) insertion anchors, one-nucleotide inversions (`g.234inv`), a frameshift anchored at an unchanged residue (`p.His150Hisfs*10`) and an immediate-stop frameshift (`p.Tyr4TerfsTer1`) are all rejected now, each with a spec-citing message. The bare-`=` allele (`c.[2376G>C];[=]`), `r.(=)`, `r.spl`, `r.spl?` and protein predicted-parens (`p.[(Arg727Ser;Cys1334Trp)]`) all parse now — #87 recorded every one of those as *not* parsing.

**Three #87 items are still open**, and are pinned individually rather than only counted:

- `alleles.md:106` — `LRG_199t1:c.[76A>C];[0]` displays as `[LRG_199t1:c.76A>C];[0]`. The accession migrates inside the bracket group; the spec writes it outside.
- `checklist.md:26` — `c.123-65_-50` is accepted. The spec calls it incomplete, and the shorthand means something else entirely (a 5' UTR position).
- `uncertain.md:184` — `r.0?` does not parse. It is a distinct semantic from `r.0`.

**#1421's five-row diagnostic table reproduces exactly.** Two rows at one locus (29–33, interior `AC` unchanged) with identical net length still take opposite split verdicts — `delinsAACTGCATGTG` separates, `delinsAACACATACTG` spans. So the separated-description decision keys on the *inserted block* rather than on which positions changed, which is `issue_1421_split_decision_still_depends_on_the_inserted_block` stated as an assertion instead of prose.

**#179's examples 1 and 2 reproduce the `Ferro:` output recorded in the issue byte for byte.** #179 never published its contig, so this reconstructs one — a repeating filler with #179's stated bases written in at #179's stated positions. The byte-for-byte match on two of three is what corroborates the reconstruction rather than assuming it, and `issue_179_reference_has_the_stated_bases` pins the premise so a future edit cannot quietly measure the wrong locus. Example 3 has moved since filing: it now yields a single spanning `delins` (`1005_1008delinsGCCT`) where the issue recorded `[1004_1005del;1006del;1008_1009insCCT]` — a different answer, still not the inversion. All three remain gaps, which is expected and not a bug being tolerated: inv recognition across a self-complementary interior is **discretionary**, not mandated (the mandatory merge threshold is one intervening nucleotide, these have two, and SVD-WG010 was rejected). The value is a measurable baseline for any future discretionary inv pass.

## Scope notes

- **No `src/` change.** Tests only.
- **Overlap with the #1419/#1420/#1421 confluence axis:** the nine headline *pairs* are guarded separately as a confluence corpus; this module deliberately does not repeat them. It harvests what those pairs left behind — #1421's diagnostic table, which is about a single spelling's verdict rather than a pair.
- **#200 yielded no harvestable rows and none were invented.** Its HGVS strings are illustrations inside a proposed API signature (`NG_008617.1:g.100A>G` → `NM_000297.4:c.459A>G` → `NP_000288.1:p.(Arg153Gly)`), not measured claims — the accession/coordinate correspondences were never asserted to hold, so turning them into tests would have meant fabricating expectations. Worth recording as a correction to any inventory that counts raw HGVS occurrences as authored cases: documentation examples inflate that count.
- No representation change: no normalized output moves.

Refs #87, #179, #1419, #1420, #1421, #1235.

## Verification

```
cargo clippy --all-features --all-targets -- -D warnings   clean
    Summary [486.419s] 9402 tests run: 9402 passed, 145 skipped
```

### DIFF
--- a/tests/it/main.rs
+++ b/tests/it/main.rs
@@ -354,6 +354,7 @@ mod mito_circular_audit;
 mod mito_heteroplasmy_audit;
 mod mosaic_chimeric_compact;
 mod mosaic_predicted_eqslash;
+mod msto_authored_cases;
 mod msto_regression_corpus;
 mod mutalyzer_grammar_tests;
 mod mutalyzer_normalize_tests;
--- a/tests/it/msto_authored_cases.rs
+++ b/tests/it/msto_authored_cases.rs
@@ -1,0 +1,737 @@
+//! Authored test cases harvested out of the issue tracker into runnable code.
+//!
+//! # Why this module exists
+//!
+//! A large body of concrete HGVS cases exists only as prose in issue bodies:
+//! spec `✓`/`✗` pairs, diagnostic probe tables, worked examples. Filed is not
+//! the same as guarded — an issue records that something was once measured, but
+//! nothing re-checks it, so behaviour drifts silently and the case has to be
+//! re-derived by hand every time someone needs it. Several of the rows below
+//! turned out to have been **fixed years ago** and several are **still open**,
+//! and there was no way to tell which without running them.
+//!
+//! # These are characterizations, not aspirations
+//!
+//! Every row asserts **what ferro does today**, so this module is green on
+//! `main` and stays green. Rows are classified:
+//!
+//! * [`Verdict::Guard`] — current behaviour matches what the issue (or the spec
+//!   line it cites) says it should be. Changing it is a regression.
+//! * [`Verdict::Gap`] — current behaviour differs from what the issue argues
+//!   for. The assertion still pins today's answer; the comment names the
+//!   expected one. **A `Gap` flipping to `Guard` is progress** — update the row
+//!   and the census, deliberately, in the PR that fixes it.
+//!
+//! Pinning the wrong answer on purpose is the point: it makes the wrong answer
+//! *visible in a diff* the day somebody changes it, which prose in a closed
+//! issue cannot do.
+//!
+//! # Provenance
+//!
+//! Every row is traceable to a specific issue. Nothing here is invented. The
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai pause

Issue #1454, opened 2026-08-06
A derived member is typed inv only on the second pass, so the first output is not a fixed point


nh13:
### Summary

A single-member `delins` normalizes to a two-member allele whose second member is spelled `delins` when it is an exact inversion. Normalizing that output again re-types it to `inv`. So the first result is **not a fixed point**, which breaks the idempotency contract and makes the normalized string unusable as a stable key without a second pass.

Reproduces on `main` with shipped defaults. Caught by `FERRO_ASSERT_IDEMPOTENT=1`, which is why a plain `cargo nextest run` is green and the oracle job is not.

### Reproducer

63 nt transcript core, CDS 1..63, so `c.n` is core offset `n - 1`:

```
TTTTTTTTTAATATATTTTAATATAATTAAAAAAATAATTTTTATAAATATATTATTTTAAAAA
```

```
NM_TEST.1:c.10_15delinsTAATAT
  pass 1 -> NM_TEST.1:c.[10_12delinsTAA;13_15delinsTAT]
  pass 2 -> NM_TEST.1:c.[10_12delinsTAA;13_15inv]
```

`c.13_15` is `ATA` and `revcomp(ATA)` is `TAT`, so the second member is an exact inversion. `general.md:56` ranks inversion above deletion-insertion, so `13_15inv` is the form both passes should produce — pass 2 is right and pass 1 is wrong.

### Why it is worth fixing rather than absorbing

The derivation already knows enough to type this correctly: the piece it emits has a payload equal to the reverse complement of its own span. Nothing new has to be computed on the second pass — the same test simply is not applied on the first.

Concretely it means the alternation loop in `canonicalize_from_sequence` is doing typing work that belongs in a terminal pass. Any consumer that normalizes once and stores the result gets a string that ferro itself would rewrite.

### Scale

One input in 23,696 `c.`-axis spellings from the generated cis-allele confluence corpus — rare, but deterministic and reachable from a plain single-member `delins`, which is one of the commonest real shapes.

`FERRO_ASSERT_IDEMPOTENT` is the only one of the three seam oracles that catches it: the output re-parses and every coordinate is in range, so `REPARSE` and `IN_BOUNDS` both pass.

### Suggested scope

[...]

PR #1458, opened 2026-08-06
test(cis): measure confluence over designed multi-member cis alleles

### LINKED ISSUE #1443: Multi-member cis alleles are 0.005% of every corpus we own, so the partitioner is effectively untested
### The measurement

The three bulk corpora hold 5.76M rows between them. Multi-member cis alleles — the only inputs that reach `canonicalize_from_sequence`, which is gated on `members.len() > 1` (`src/normalize/mod.rs:2027`) — number **264**.

| corpus | rows | multi-member cis |
|---|---|---|
| `clinvar_hgvs_500k.json.gz` | 500,004 | 10 |
| `paraphase_genes_exhaustive.json.gz` | 435,235 | 43 |
| `cmrg_genes_exhaustive.json.gz` | 4,826,063 | 211 |
| **total** | **5,761,302** | **264** |

**0.0046%.** Every other row takes the per-member path and exercises none of the partitioning, weight-bound or member-ordering logic.

And the handful that exist are not hard cases. The ClinVar ten are widely-separated substitutions:

```
NG_016648.1:g.[20557T>C;22424C>T]     1867 nt apart
NM_000208.3:c.[2504G>T;2525C>T]         21 nt apart
NM_000540.2:c.[8026C>T;8360C>G]        334 nt apart
```

No partitioner has to make a choice on those. The blocks where `partition_block` genuinely has options — two changes close enough to share a window, with coincidental matches between them — are **absent from every large corpus in the repo**.

### Why that is a population problem, not a coverage oversight

ClinVar, CMRG and Paraphase are corpora of *observed* variants. Nature rarely places two mutations close enough to share a block, so the shape is genuinely rare in them and no amount of harvesting will make it common.

Consumers that submit *designed* alleles have the opposite distribution: multi-member cis alleles by construction, at close separations, at scale. That is precisely the region our corpora do not cover, and it is where the recently-filed non-confluence reports come from.

So the existing corpora cannot be made to cover this. Two new things are needed.

### Part 1 — harvest the 264 real rows into a permanent axis

They are the only real-world evidence the repo has for this path, and they currently sit inside three LFS corpora whose suites **skip green when the fixtures are absent**. Extracting them into a small committed fixture makes the axis unconditional.

Two tiers, because most rows name accessions that need a prepared reference:

- **always-on:** every row parses; a pinned census of member counts and inter-member separations, which also documents the real-world distribution
- **manifest-gated:** normalize each row, assert idempotence and sequence preservation, and assert that the spanning-`delins` respelling of the same variant converges to the same output

### Part 2 — generate a designed-allele corpus

Programmatic, not a committed data file: multi-member cis alleles at the separations and payload sizes systematic indel designs actually produce, over a synthetic contig.

The point is not volume, it is the assertion. For each generated design, emit **every equivalent spelling** — the design as authored, the spanning `delins` over its union span, and each member shifted within its own ambiguous run — and assert they all normalize to one string. That is the property the recent reports are about, and nothing in the repo currently asserts it at scale.

### Acceptance

- Both axes run in CI without a prepared reference (the manifest-gated tier skips, the rest does not)
- The designed corpus emits thousands of confluence classes, not dozens
- A census is pinned for each, so a change in the corpus is visible in review rather than silent

nh13:
Closes #1443

## What this measures

`canonicalize_from_sequence` is gated on `members.len() > 1`, so only a multi-member cis allele reaches the partitioner. Part 1 here harvested every such input from the bulk corpora and found **592 rows out of 9,949,738** — real corpora are observed variants, where two changes rarely land close enough to share a block. Consumers submitting *designed* alleles have the opposite distribution, so the partitioner is effectively untested and no real corpus can fix that.

Part 2 fills the gap with a generated corpus: **11,272 classes / 47,392 spellings**. Each class is one synthetic reference, one denoted sequence, and N ≥ 2 distinct spellings verified — by an applier independent of the normalizer — to denote that sequence. Every spelling is normalized and the confluence question asked: does a class reach exactly one output?

**It does not, for 41% of classes.** That number could not be stated in this repo before.

| | 3' | 5' |
|---|---|---|
| converged | 6,632 | 6,627 |
| split 2 | 4,506 | 4,531 |
| split 3 | 115 | 105 |
| split 4+ | 19 | 9 |
| declined | 0 | 0 |
| **sequence changed** | **0** | **0** |

`sequence_changed` is pinned at zero and is the harder half: every output is applied back to the reference through `hgvs_to_spdi` and compared with the class's denoted sequence. So every divergence is two valid, sequence-preserving spellings of one variant — the confluence defect itself, not a parse failure or a corrupted sequence wearing well-formed clothes.

The two directions land within five classes of each other, which is itself evidence: the divergence is a property of how the partitioner splits a block, not of which end of an ambiguous run the shuffle walks to. A fix that moved only one of these numbers would be treating a symptom.

## No `src/` change, and no representation move

This PR is tests, fixtures and generators only — `git diff --stat` touches no `src/`. Per CLAUDE.md's representation-stability note: **no normalized output moves here**, so there is nothing for a consumer to re-baseline.

## This is a census, not a pass/fail gate

A permanently red suite is worth nothing, so the figures are pinned baselines: each must only ever move toward convergence. A change that raises a divergence count is a regression and must be explained; one that lowers it re-blesses in the same commit.

## The pins are per oracle configuration

Each direction pins two censuses, selected on `FERRO_ASSERT_IDEMPOTENT`. Two classes normalize to an output that is not a fixed point — `c.10_15delinsTAATAT` gives `c.[10_12delinsTAA;13_15delinsTAT]`, whose second member is re-typed `13_15inv` only on the next pass (#1454). With the oracle off those two classes produce two distinct outputs and land in `split 2`; with it on they panic, are counted `declined`, and the class collapses to one outcome, so `converged` rises by two and `split 2` falls by two.

| | 3' plain | 3' oracle | 5' plain | 5' oracle |
|---|---|---|---|---|
| converged | 6,632 | 6,634 | 6,627 | 6,629 |
| split 2 | 4,506 | 4,504 | 4,531 | 4,529 |
| declined | 0 | 2 | 0 | 2 |

CI runs both — `Test` without the oracles, `test-oracle` with them — so a single pin could only ever be green in one job. Both constants collapse into one when #1454 is fixed, and the doc comment says so at the place that has to be deleted.

Verified green in both configurations: `9434 passed` plain, `9417 passed` with all three oracles and `FERRO_SWEEP_SEEDS=full`.

## The ratchet already paid for itself

While this branch was in development `main` moved seven times underneath it, and the corpus caught a real movement that no existing test could see:

| figure | at #1445 (`b9dbb494`) | at `ee0e37ac` |
|---|---|---|
| 3' split-3 | 125 | 115 |
| 3' split-4+ | 21 | 19 |
| 5' split-3 | 119 | 105 |
| 5' split-4+ | 11 | 9 |

#1445 fragmented 12 confluence classes as a side effect of deliberately no longer laundering overlap conflicts. #1438 and #1446 have since fully repaired it — the split-3 and split-4+ counts are back at their pre-#1445 values.

Isolated by measuring this branch's corpus against `src/` at `b9dbb494` and at `ee0e37ac` with nothing else changed. Nothing from that episode is outstanding; it is recorded because "a fragmentation landed, went unnoticed, and was fixed without anyone knowing either happened" is exactly the argument for merging this corpus.

## Contents

- `examples/harvest_multi_member_cis.rs` — the 592-row harvest, and `tests/fixtures/cis/multi_member_cis_alleles.json`
- `examples/generate_cis_confluence_corpus.rs` — the 11,272-class generator (corpus is gitignored, regenerated on demand via `common::fixture_gen`, ~3s, so a fresh checkout just works)
- `tests/it/cis_confluence_axis.rs` — the census, both shuffle directions
- `tests/it/multi_member_cis_axis.rs` — the real-row axis (82% convergence over the respellable rows, which is what far-apart members look like)

## Selection is a parse, not a bracket scan

The harvester decides with `parse_hgvs`: a row is kept only if it parses as an `AllelePhase::Cis` allele with more than one member, and the member count comes from the parsed allele. Cheap string filters run first so the parse only sees a few thousand of the ten million rows.

It has to be a parse. An inserted-range payload carries its own brackets *and* its own `;`, so a `find('[')..rfind(']')` scan reads `LRG_676:g.32086_32087ins[N[1090];32074_32086]` — one insertion — as a two-member allele, and the same payload nested inside a genuine allele inflates that allele's count. Depth-awareness alone does not fix the first shape, because depth cannot tell an allele bracket from a payload bracket.

That moved the harvest from 650 rows to **592** (`rows_scanned` unchanged at 9,949,738). The 58 dropped rows are 46 malformed ClinVar strings ferro rejects by design (`c.[1327G>T;1727G>C2002C>T]`, `c.[652C>T;c.940A>G]`, `gom`/`lom`) plus 12 that are not alleles at all — a single variant whose payload happens to carry a `;`. One genuine allele, `LRG_501:g.[...ins[SEQ;a_b]...]`, drops from a recorded 4 members to its true 3. No row was added.

The pinned censuses move with it: members-per-allele to `{2: 561, 3: 22, 4: 6, 6: 3}`, member pairs from 698 to 638 (106 within eight bases, 261 a hundred or more apart), and the manifest-gated census from `rows: 650, declined: 49, unwindowed: 106` to `rows: 592, declined: 0, unwindowed: 96`. `respellable: 461` and `respelling_converged: 376` are unchanged — none of the dropped rows was respellable — so the 82% headline is unaffected.
- `tests/it/reported_confluence_pairs.rs` — guards the externally-reported pairs from #1419/#1420/#1421


nh13:
@coderabbitai pause

PR #1459, opened 2026-08-06
test(repr): build conflicting cis alleles in the representation corpus

### LINKED ISSUE #1456: Representation corpus builds no overlap-conflicting alleles, so it reports 0 movement for the PRs that change them
`dump_normalized_corpus` (#1442) is the harness the representation-stability rule points at, and it is the right idea. But its corpus builds five shape families — `adjacent_junction_ins`, `del_plus_sub`, `dup_plus_ins`, `dup_plus_sub`, `split_vs_spanning_delins` — and **none of them constructs an allele whose members conflict**: no coincident-bounds pair, no junction interior to a span, no partially-overlapping or nested spans.

That is exactly the class the recent normalizer work changes.

## Measured

Three PRs in a row were diffed against it and all three reported the same thing:

| PR | what it changes | corpus movement |
|---|---|---|
| #1403 | equal-length blocks above `MAX_SPLIT_BLOCK` | 0 of 18,432 |
| #1445 | preserves a laundered overlap conflict | 0 of 18,432 |
| #1451 | reports partially-overlapping and nested spans as conflicts | 0 of 18,432 |

#1445 and #1451 move behaviour for thousands of inputs by their own enumerations — #1445 takes laundered rows from 1,888 to 486, #1451 reports 10,499 of 25,848 strict-accepted span pairs as denoting nothing. The corpus sees none of it.

## Why this matters more than a coverage gap

A zero from this harness is currently ambiguous between "nothing moved" and "the corpus cannot see what moved", and only the second is true for these PRs. The stability rule asks every representation-moving PR to quantify its movement; if the tool it points at is blind to the class in question, the disclosure is satisfied on paper while telling the reader nothing. Each of those three PR bodies had to carry a hand-written caveat saying so.

## Suggested shape

Add conflict families to the corpus generator: coincident bounds (`[Ndup;NX>Y]`), a junction interior to a span (`[a_bdel;a_binsN]`, `[a_binv;a_binsN]`), partially-overlapping spans (`[10_14del;12_16del]`), and nested spans (`[10_14del;10_16del]`). Those four cover what #1406/#1411/#1414/#1437/#1448 have each had to enumerate by hand in their own throwaway harnesses.

Worth also considering whether the compare output should distinguish "0 moved" from "0 rows in the families this diff touches" — the second is the honest answer when a change is scoped to a family the corpus lacks.

nh13:
Closes #1456.

## The defect

`dump_normalized_corpus` is the harness the representation-stability rule points at, but its five shape families — `adjacent_junction_ins`, `del_plus_sub`, `dup_plus_ins`, `dup_plus_sub`, `split_vs_spanning_delins` — all pair members on **disjoint** territory. It could not build a conflicting allele at all, which is exactly the class the recent normalizer work changes. Three PRs in a row diffed against it and all three reported `0 of 18,432`, so each had to hand-write a caveat into its own body saying the zero meant blindness rather than safety.

## The fix

Four conflicting families, one per shape the two detectors in `normalize::overlap` actually key on:

| family | shape | from |
|---|---|---|
| `coincident_bounds` | `[Ndup;NX>Y]` | #1307 (`g.[24dup;24C>G]`) |
| `junction_interior_to_span` | `[a_bdel;a_binsN]`, `[a_binv;a_binsN]` | #486, mutalyzer `EOVERLAP` |
| `partial_overlap_spans` | `[10_12del;11_13del]` | #1451 |
| `nested_spans` | `[10_11del;10_13del]` | #1451 |

The corpus goes from 18,432 to 33,792 rows; dump time stays ~5 s in a debug build. Nothing in the five existing families is touched, so their rows remain comparable against older dumps.

`compare` also **states** a zero rather than leaving it to be inferred from the per-family table, and names the families it covered:

> **No row moved.** That is a measured zero for the 9 shape families above and says nothing about a shape this corpus does not build: `adjacent_junction_ins`, … If the change under test is scoped to a shape absent from that list, this harness has not measured it — add the family rather than quoting the zero.

## Does it actually see what it was blind to?

Re-measured with **this** harness on both sides of each PR the issue cites (the example depends only on the public API, so the same file builds against either revision and both dumps describe the same corpus — that workflow is now in the module docs):

| PR | with the old corpus | with this corpus |
|---|---|---|
| #1445 `b9dbb494` vs `a76d3400` | 0 of 18,432 | **1,411 of 33,792 (4.2%)** — all `junction_interior_to_span` |
| #1451 `ee0e37ac` vs `ee6a19cc` | 0 of 18,432 | **3,320 of 33,792 (9.8%)** — 1,707 `partial_overlap_spans`, 1,606 `nested_spans`, 7 `junction_interior_to_span` |
| #1403 `0ec3c275` vs `699fb592` | 0 of 18,432 | **0 of 33,792** |

Every pre-existing family reports 0 for all three, which is the check that the old zero was blindness and not agreement.

**#1403's zero is structurally guaranteed, not evidence.** Its gate keys on length — `reference.len() != result.len() && (… > MAX_SPLIT_BLOCK)`, with `MAX_SPLIT_BLOCK = 1024` (`src/normalize/merge.rs:1791`, `:3536`) — while this corpus draws 20-mer cores whose widest shape reads five offsets. No row it can emit comes within two orders of magnitude of the cap, so both sides of #1403 take the same path by construction. Covering it needs kilobase-scale cores: a different axis (sequence length) from the member geometry added here, and one no new family fixes. Filed as #1460. The difference now is that the zero arrives with the sentence saying so instead of as a bare number.

**`coincident_bounds` moved 0 across all three**, so it is coverage without a demonstrated witness in this set. It is not inert: every one of its 3,072 rows normalizes away from its input (`g.[257dup;257T>A]` → `g.256_257insA`, the #1307 laundering), so the string is now tracked.

## Unchanged on purpose

- The five existing families, the sequence generator, the seed count and the `(reference, axis, direction, input)` key — a re-rolled corpus would make every earlier dump incomparable.
- `compare`'s refusal to diff two dumps that do not cover the same rows. Widening it to an intersection would silently accept a re-rolled corpus, which is the failure the refusal exists to catch; the module docs now give the copy-the-example workflow instead.

## Representation stability

**No normalized output moves.** The diff touches `examples/dump_normalized_corpus.rs` only — no `src/`, so no shipped form changes and there is nothing for the downstream consumer to re-normalize. The numbers above measure *past* PRs with the new corpus; they are not movement introduced here.

## Verification

```
# Summary [546.989s] 9413 tests run: 9413 passed (12 slow), 145 skipped
```

Built test-first: `the_corpus_emits_alleles_whose_members_conflict` classifies every two-member `g.` allele the generator emits by the geometry of its members' footprints and requires each of the four conflict classes to be non-empty. It failed on `main` with all four at zero. It reads the geometry back off the emitted description rather than off the family table, so adding or renaming a family cannot satisfy it vacuously. `a_report_with_movement_does_not_claim_a_zero` is the discriminating half of the zero caveat — printing it unconditionally would pass the other test while telling a reader of a moving diff that nothing moved.

### DIFF
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -20,6 +20,21 @@
 //!     --compare /tmp/before.tsv --against /tmp/after.tsv
 //! ```
 //!
+//! ## Comparing across a revision that changed the corpus
+//!
+//! `compare` refuses two dumps that do not cover the same rows, so a revision that
+//! adds a shape family cannot be diffed against a dump taken with the older
+//! generator. Copy **this file** into the base checkout and dump from there: it
+//! depends only on the public API, so it builds against either revision and both
+//! dumps then describe the same corpus.
+//!
+//! ```text
+//! git worktree add --detach ../base <base-sha>
+//! cp examples/dump_normalized_corpus.rs ../base/examples/
+//! (cd ../base && cargo run --features dev --example dump_normalized_corpus -- --out /tmp/before.tsv)
+//! cargo run --features dev --example dump_normalized_corpus -- --out /tmp/after.tsv
+//! ```
+//!
 //! ## What this is NOT for
 //!
 //! Prepared references. `ferro normalize --input <file> --reference <dir> --format
@@ -30,7 +45,7 @@
 //! densely cover the cis-allele shape families where representation churn actually
 //! happens. That is the gap this fills.
 //!
-//! ## Two traps, both hit while measuring #1401
+//! ## Three traps, the first two hit while measuring #1401
 //!
 //! 1. **A row's identity is `(reference, axis, direction, input)` — never the input
 //!    alone.** Every synthetic reference reuses one accession, so keying on the
@@ -41,6 +56,13 @@
 //!    deliberately enriched for the shapes that churn, so its percentages are far
 //!    higher than a real corpus would give — #1401 measures 23% here. Quote it as
 //!    "of the affected shape family", never as a repo-wide figure.
+//! 3. **A zero measures only the families the corpus builds** (#1456). Before the
+//!    conflict families below were added, this harness reported `0 of 18,432` for
+//!    #1403, #1445 and #1451 in a row — none of which the corpus could express, and
+//!    two of which move behaviour for thousands of inputs by their own enumerations.
+//!    `compare` now says so in the zero case rather than leaving it to be inferred.
+//!    If your change is scoped to a shape no family builds, add the family; do not
+//!    quote the zero.
 //!
 //! ## The corpus is deliberately independent of `tests/it/common/`
 //!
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

Issue #1461, opened 2026-08-06
Normalizer: a valid 457 nt inversion is rendered as a 109-member allele


nh13:
A plain, valid 457 nt inversion is normalized into a 109-member allele of substitutions and delins, discarding the user's own well-formed input.

## Reproduce

```
NC_000013.10:g.100809575_100810031inv
```

`ferro normalize --reference <prepared-dir> --format tsv --error-mode lenient`, default configuration.

## Actual

A 109-member allele:

```
NC_000013.10:g.[100809575A>T;100809577_100809579delinsCGT;100809581_100809584delinsAATT;...;100810027_100810029delinsTTC;100810031A>T]
```

(Verified: exactly 109 members.)

## Expected

The input, unchanged. `recommendations/general.md:56` ranks inversion above substitution and delins, and this is one variant, not 109.

## The input really is an inversion

The reference block `NC_000013.10:100809575-100810031` is 457 nt, and the observed sequence is its exact reverse complement. The block is **not** palindromic, so the inversion is well-defined and non-degenerate — this is not a case where `inv` is discretionary.

## Notes

- Both spellings denote the same sequence, so this is a representation defect, not a correctness one. It is nonetheless severe: a 109-member allele is unusable downstream, and it destroys an input that was already in the recommended form.
- Independent of the block-partitioner question. It reproduces on the shipped default path.
- Counterpart to #1230, which is inversion **over**-recognition. This is the opposite failure and suggests the recognition predicate is too narrow for long spans rather than simply mis-tuned.

PR #1464, opened 2026-08-06
feat(normalize): add the member-count-minimal partitioner behind a flag

### LINKED ISSUE #1444: Add a member-count-minimal ("canonical") block partitioner, and wire a three-way bake-off
Ferro has two block partitioners, and neither implements the rule that `mutalyzer-algebra` calls `canonical`.

- **live** — `partition_block` in `src/normalize/merge.rs`: a single-gap alignment search plus two narrow escapes (`two_gap_insertion_alignment`, and a deletion mirror on some branches).
- **shadow** — `partition_block_sequence_first` → `src/normalize/seqfirst/`: builds `AlignmentDag` (the DAG of every minimal alignment), computes `Dominators`, and cuts at the edges common to *all* minimal alignments.

Measured against Leiden's `mutalyzer-algebra`, the shadow arm behaves like algebra's `local_supremal` — the dominator-based extent — **not** like algebra's `canonical`. Algebra documents `canonical` as:

> Traverse (BFS) the LCS graph to extract the canonical variant, i.e. minimize the number of separate variants within an allele.

That is a *second* lexicographic key on top of minimal edit distance: among the alignments the DAG already encodes (all of which are distance-minimal by construction), pick one whose maximal runs of change are fewest. Ferro implements neither that selection nor anything equivalent, so the third arm of the comparison simply does not exist and we cannot say where ferro's live rule sits between the two algebra notions.

## What this asks for

1. A `canonical` layer-2 selection over the existing `AlignmentDag` — a shortest-path pass whose per-path cost is the number of maximal change runs, not the number of edges. Reuse the DAG; do not build a second graph. The tie-break among equally-few-member alignments is an implementer's choice the spec does not make, so it should be stated explicitly in the code (3'-most, per `general.md:41`).
2. An env knob (`FERRO_PARTITION=live|shadow|canonical`) selecting which partitioner `canonicalize_from_sequence` uses, defaulting to today's `live` with byte-identical output.
3. A block-dump example that reads `(ref_block, alt_block)` pairs and emits all three partitions side by side, plus the block's Levenshtein distance and each arm's changed-column count, so distance-minimality is checkable per arm straight from the dump.

## Why it is worth having

Non-confluence work (#1235, #1419–#1421) keeps having to argue about which representation a partitioner *should* pick, with only two candidate rules on the table and no way to see a third. A per-block dump of all three, with the distance and column counts alongside, turns that argument into a measurement — and it is the measurement that has to come before any proposal to promote a shadow partitioner to authoritative, since promoting one is a representation-stability event for downstream consumers, not a drop-in fix.

nh13:
Closes #1444

## The default path does not move a single normalized output

Stating this first, because it is the thing a reviewer needs to know: **this PR changes no shipped representation.** The new partitioner is behind `FERRO_PARTITION` and is off by default.

Measured with the committed `#1441` harness at this branch's exact parent (`ee0e37ac`), default environment, no flag set:

| | rows |
|---|---|
| total | 18,432 |
| **moved** | **0 (0.0%)** |

Byte-identical, `md5 436379ddba88a98cd7150909cec4af27` on both sides.

A zero can also mean the corpus never reaches the new code, so the same corpus was run through the new arm to prove the measurement is sensitive: `FERRO_PARTITION=canonical` moves **5,188 rows (28.1%)**. The instrument works; the default genuinely does not move.

## What it adds

`CanonicalAlignment` — a member-count-minimal partitioner — plus a three-way selector and a dump harness:

- `FERRO_PARTITION=live` (default) — the shipped `partition_block`, unchanged
- `FERRO_PARTITION=shadow` — the seqfirst DAG splitter
- `FERRO_PARTITION=canonical` — the new one
- `examples/dump_partitions.rs` — dumps and diffs partitions across the three arms

The objective is fewest members subject to minimal edit distance. Where the live splitter's decisions are driven by the *input spelling*, this one is a function of `(reference, observed)` alone, which is what makes it a candidate for the confluence work in #1235.

## What the canonical arm would cost, if it were ever turned on

Not proposed here, but measured, because the numbers should be on the record next to the code:

| corpus | rows | moved |
|---|---|---|
| synthetic (enriched for churning shapes) | 18,432 | 5,188 (28.1%) |
| designed multi-member cis alleles | 650 | 16 (2.66%) |
| ClinVar | 500,004 | 164 (0.033%) |
| CMRG (stride 24) | 201,086 | 106 (0.053%) |
| Paraphase (stride 3) | 145,079 | 94 (0.065%) |

The 28.1% is a rate **within the enriched shape families**, never a library-wide figure — the harness's own docs flag that trap and it is repeated here so the number is not quoted loose. Across 846,819 real rows there were **zero** regressions: no row errored or declined where the baseline succeeded.

Two properties of that churn matter more than its size, and both argue for keeping the flag off until #1235 resolves them:

1. **153 of the 164 ClinVar moves are single-member `g.` delins becoming multi-member alleles** — a change of *arity*, not spelling. A consumer's schema and joins see a shape change, not just a different string.
2. A hand adjudication of 5 moved rows against the spec came back **4 for the stored form, 1 for the new one**. The losing pattern is consistent: the new form splits a block only because the delins payload spuriously aligns with the reference, which is what `DNA/delins.md:44-47` explicitly recommends against. That adjudication is being scaled to all 180 moved rows before any proposal to flip the default.

## Review notes

- No behavioural change reaches any default path; the diff is additive plus the env-var selector.
- Full gate green at `ee0e37ac`: `9418 tests run: 9418 passed, 145 skipped`, with `FERRO_ASSERT_IDEMPOTENT`, `FERRO_ASSERT_REPARSE`, `FERRO_ASSERT_IN_BOUNDS` and `FERRO_SWEEP_SEEDS=full`.
- Merging this does not commit the project to the canonical form. It commits it to being able to *measure* it, which is currently only possible by rebuilding the harness by hand.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -318,6 +318,10 @@ required-features = ["dev"]
 [[example]]
 name = "report_conformance_reference_gaps"
 required-features = ["dev"]
+
+[[example]]
+name = "dump_partitions"
+required-features = ["dev"]
 [[bin]]
 name = "ferro"
 path = "src/bin/ferro.rs"
--- a/examples/dump_partitions.rs
+++ b/examples/dump_partitions.rs
@@ -1,0 +1,239 @@
+//! Dump all three block partitions side by side, one row per block.
+//!
+//! The bake-off harness for ferro's three partitioners. Each input row is a
+//! `(reference_block, alternate_block)` pair; each output row is what `live`,
+//! `shadow` and `canonical` each make of it, together with the block's
+//! Levenshtein distance and each arm's changed-column count so distance-minimality
+//! is checkable per arm straight from the dump.
+//!
+//! The three rules are not variations on one another:
+//!
+//! * **live** — `partition_block`: a single-gap alignment search plus two narrow
+//!   escapes. The shipped rule.
+//! * **shadow** — `partition_block_sequence_first`: cut at the alignment steps
+//!   common to *every* minimal alignment, then merge runs separated by fewer than
+//!   `--min-separation` unchanged bases. `mutalyzer-algebra`'s `local_supremal`.
+//! * **canonical** — `partition_block_canonical`: the minimal alignment with the
+//!   fewest members, 3'-most among ties. `mutalyzer-algebra`'s `canonical`.
+//!
+//! # Usage
+//!
+//! ```text
+//! cargo run --features dev --example dump_partitions -- [--min-separation N] [PATH]
+//! ```
+//!
+//! `PATH` defaults to `-` (standard input). Input is tab-separated
+//! `reference<TAB>alternate`, one block pair per line; empty lines and lines
+//! beginning with `#` are skipped, and either block may be empty. Bases are
+//! compared byte-exactly and are **not** case-folded — pass the blocks in the
+//! same case the normalizer would, which is upper case.
+//!
+//! `--min-separation` applies to the `shadow` arm only; the other two take no
+//! such parameter. It defaults to the coding one-amino-acid exception (2). Pass
+//! `1` to measure the arm as a genomic axis would run it.
+//!
+//! # Output
+//!
+//! One header row, then one tab-separated row per input block:
+//!
+//! ```text
+//! ref  alt  distance  live  live_cols  shadow  shadow_cols  canonical  canonical_cols
+//! ```
+//!
+//! A partition renders as its members joined by `|`, each as
+//! `ref_start:ref_end/alt` over 0-based half-open reference offsets into the
+//! block — so `0:1/|3:4/` is two deletions and `2:2/CAC` is one insertion. An
+//! empty partition (an unchanged block) renders as `-`.
+//!
+//! Two sentinels replace a partition rather than a member list. `DECLINED` is an
+//! arm that legitimately returned nothing — the `shadow` arm does on some blocks
+//! — and `PANIC` is an arm that panicked, which is caught per arm so one bad
+//! block cannot abandon the dump. Their column counts render as `-`.
+
+use ferro_hgvs::normalize::dev_partitioners::{self, DevPiece, DEFAULT_MIN_SEPARATION};
+use std::io::{BufRead, BufWriter, Write};
+use std::panic::{catch_unwind, AssertUnwindSafe};
+
+/// What one partitioner made of one block.
+enum Outcome {
+    Partitioned(Vec<DevPiece>),
+    /// The arm returned `None`.
+    Declined,
+    /// The arm panicked. Recorded rather than propagated, so a dump over a large
+    /// corpus reports the bad block instead of stopping at it.
+    Panicked,
+}
+
+impl Outcome {
+    /// Run one arm, converting a panic into [`Outcome::Panicked`].
+    fn of<F: FnOnce() -> Option<Vec<DevPiece>>>(arm: F) -> Self {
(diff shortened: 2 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
## The scaled adjudication this PR's body promises is complete

The body says the 5-case adjudication "is being scaled to all 180 moved rows before any proposal to flip the default". That has now been run — twice, independently, by disjoint methods (one reading bases from the FASTAs, one recovering them by probing the normalizer itself with a single-base `delins` per position). Recording the result here so it sits next to the code rather than only in a working note.

**The 5-case result did not just hold, it strengthened: 4:1 became roughly 8.5:1.**

| scope | spec prefers stored | prefers new | silent | unsupported |
|---|---|---|---|---|
| all moved rows (208, incl. cis) | **144** | 17 | 43 | 4 |
| strict sub-class: stored *is* the minimal single delins and the new form splits it | **120** | **0** | **0** | 0 |

**Structure, which matters more than the ratio.** Of the moved rows, **121 are a stored form that is already the exact minimal single `delins`, split purely because payload bases coincide with reference bases in between** — `DNA/delins.md:44-47` verbatim. Nothing else appears in that stratum. There is no row in that class where the spec prefers the split.

**A correction to the counts in the body.** ClinVar moves **192**, not 164; the 164 figure excludes 28 `LRG_` rows. Of the 192, **153** are single-member `g.` delins, which does match. (The two runs reconcile this differently — one attributes it to improved ClinVar resolution, the other to the earlier run failing the LRG rows — so treat 192/164 as the reconciled pair and not the mechanism.)

**Non-uniqueness is the strongest stability argument, and it is worse than the sample suggested.** 27 of 40 exactly-enumerated `payload_alignment_split` rows (68%) admit more than one equally compliant split — median 2, **max 125** (`NG_012849.2:g.83009_83019delins…`). The spec's own worked example admits 5. So adopting the split does not select *a* representation; it selects one member of a family of 2–125 by a tie-break the spec does not specify. For reference, the stored form equals the user's own input on **128 of 208** moved rows; the new form does on **5**.

**RFC 2119 force** (`recommendations/style.md` puts it in force): `delins.md:47` "recommended" is a **SHOULD** and governs 138 of 208 rows. `general.md:56` prioritisation carries no keyword. The only **MUST NOT** in the family is `general.md:58`, aimed at over-description — and it runs *against* the stored form, on one row. No rule at must-strength supports the split.

**The reverse direction is real and worth separating out.** 17 rows are ones the spec genuinely prefers the new form on, and they are fixable **without** adopting the partitioner. The most severe is now filed as #1461: the shipped rule renders a valid 457 nt inversion as a **109-member** allele, which the canonical arm returns unchanged. Ten more are one repeated `general.md:56` defect (leading with a deletion where a substitution is available).

## What this means for merging

Nothing here argues against merging this PR. The default is byte-identical and re-verified, and the measurement surface is worth having. It argues against the step *after* it: **do not flip the default to `canonical` on its own.** On the spec's own worked example (`LRG_199t1:c.850_901`, real bases) the canonical arm emits a **5-member** split where the spec pins one `delins` and explicitly names the 4-member split as the description to avoid — and it cannot do otherwise, because that recommended form claims 52 changed columns against a block edit distance of 39, putting it outside the minimal-distance set the partitioner searches at all.

#1470 is the first increment that starts to make promotion viable: a post-pass that restores 71 of the 192 moved rows and introduces 0 moves of its own.

PR #1465, opened 2026-08-06
fix(normalize): repair non-coding `r.` members instead of reverting them

### LINKED ISSUE #1453: RNA axis emits a repeated member where the n. axis collapses correctly
### Summary

On the `r.` axis the multi-member merge path emits the **same member twice**. The resulting description denotes no sequence — two members claim one interbase position — so this is a well-formedness/sequence defect, not a representation preference.

The `n.` axis, against the same reference with the same members, collapses them correctly.

Reproduces on `main` with shipped defaults, no feature flags.

### Reproducer

63 nt transcript core, `NR_TEST.1`, no CDS:

```
```

```
NR_TEST.1:r.[9dup;10dup;11dup]  ->  NR_TEST.1:r.[9dup;11dup;11dup]   <- 11dup twice
NR_TEST.1:n.[9dup;10dup;11dup]  ->  NR_TEST.1:n.9_10insTAA           <- correct
NM_TEST.1:c.[9dup;10dup;11dup]  ->  NM_TEST.1:c.9_10insTAA           <- correct
```

Same reference bases, same placement, same three members. Only the axis differs.

### Scale

Found by extending the cis-allele confluence corpus generator to the `n.` and `r.` axes. Over 5,636 generated classes per axis:

| axis | classes whose normalized output no longer denotes the input's sequence |
|---|---|
| `g.` | 0 |
| `c.` | 0 |
| `n.` | 0 |
| **`r.`** | **100 (3' direction) / 115 (5')** |

The count is **identical** under every partitioner arm and under an experimental widening of `is_splittable_single_member`, so the defect sits upstream of the partitioner and of the axis gate rather than in either.

The corpus verifies independently — through `hgvs_to_spdi`, not through the normalizer — that every input spelling denotes the sequence its class claims, so these are genuine corruptions rather than a generator artefact.

### Why it matters more than the count suggests

Every other confluence finding in this family is a *representation* difference: two well-formed strings for one variant. This one produces a string that is not a valid description of anything, on an axis where the equivalent `n.` and `c.` inputs are handled correctly.

It is also invisible to two of the three seam oracles. `FERRO_ASSERT_REPARSE` accepts `r.[9dup;11dup;11dup]` because it is syntactically well-formed, and `FERRO_ASSERT_IN_BOUNDS` accepts it because every coordinate is in range. Only an applier notices, and the apply oracle does not run on the `r.` axis today.

### Suggested scope

- Find why the member-merge path treats `r.` differently from `n.` when the two share a transcript and differ only in rendering alphabet.
- Add `r.` coverage to whatever asserts members are disjoint after a merge — a repeated member should be impossible to emit regardless of axis.
- Consider extending the SPDI apply oracle to the `r.` axis; it is the only check that would have caught this.

nh13:
Closes #1453.

## The defect

`NR_TEST.1:r.[9dup;10dup;11dup]` normalizes to `NR_TEST.1:r.[9dup;11dup;11dup]` — the same member twice. Two members claiming one interbase point denote no sequence, so this is a well-formedness defect rather than a representation preference, and it is invisible to two of the three seam oracles: `FERRO_ASSERT_REPARSE` accepts the output because it is syntactically well-formed, and `FERRO_ASSERT_IN_BOUNDS` accepts it because every coordinate is in range.

**The discriminator is the CDS, not the axis.** The issue frames this as `r.` versus `n.`; it is not. `r.` on a *coding* transcript collapses the same three members correctly:

| input | on `main` |
|---|---|
| `NR_TEST.1:r.[9dup;10dup;11dup]` | `NR_TEST.1:r.[9dup;11dup;11dup]` ← repeated member |
| `NR_TEST.1:n.[9dup;10dup;11dup]` | `NR_TEST.1:n.9_10insTAA` |
| `NM_TEST.1:r.[9dup;10dup;11dup]` | `NM_TEST.1:r.9_10insuaa` |
| `NM_TEST.1:c.[9dup;10dup;11dup]` | `NM_TEST.1:c.9_10insTAA` |

### Mechanism

Found by instrumenting the per-member pipeline rather than by reading code. `coalesce_members_at_one_junction` (#1286) — the pass that exists to merge two members which shifted onto one junction — *does* reach this input: it groups the two `11dup`s and reads both payloads. It then calls `respell_at_gap`, whose `r.` arm resolves the gap through the transcript via `ordered_cds_bounds` (#1284), because on a coding transcript the `r.` axis is CDS-relative and `+1` across the CDS start would name the non-existent `r.0`.

`ordered_cds_bounds` requires **both** `cds_start` and `cds_end`. On a non-coding record it returns `None`, `respell_at_gap` reverts the member, and the pass abandons the group — leaving exactly the repeated member it exists to remove. Every repair routed through `respell_at_gap` was therefore a silent no-op on a non-coding `r.` member; the coalesce is just where it is observable.

### Census

3,200 rows: 6 deterministic homopolymer-enriched 63 nt cores plus the issue's, × two- and three-member cis alleles of nearby junction-occupying edits, × both shuffle directions, on the non-coding `r.` axis.

| | on `main` | with the fix |
|---|---|---|
| outputs repeating a member | 849 (2 families) | **0** |
| outputs disagreeing with the `n.` axis of the same record | 1,046 (3 families) | **0** |

## The fix

Fall back to the transcript-relative axis exactly where `axis_frame` and `region_sequence_delta` already do — a `Region::Rna` member on a record with no CDS bounds at all. The three have to agree about which records and regions are transcript-relative, or a repair reads its bases through one convention and names them under another.

Deliberately keyed on that predicate rather than on "`ordered_cds_bounds` declined", which is wider in three ways that all must keep refusing:

- **`Region::Rna` only.** `FivePrimeUtr` (`r.-N`) and `ThreePrimeUtr` (`r.*N`) still require CDS bounds — without a CDS those regions do not exist, so there is no position for a fallback to name.
- **Inverted bounds** are a malformed record whose CDS-relative axis is incoherent (`ordered_cds_bounds` documents why), and a record carrying `cds_start` without `cds_end` has a genuinely CDS-relative axis whose 3'UTR simply cannot be named.
- **An unresolvable transcript** gives no grounds to claim its axis is transcript-relative.

### Three call sites, not one

`respell_at_gap` is the junction path. The **terminal** path is separate: when the landing junction is one past the last base, the repair becomes the boundary `delins` of #1327 and `respell_at_sequence_end` resolves the last base through `cds_axis_end` — `ordered_cds_bounds` again. It calls it *twice*, once to place the write and once to verify it landed, so `rna_axis_end` is shared between them. The first cut of this fix covered only the junction path, and the census still showed 1 surviving row; that row is now a pinned regression test.

## Unchanged on purpose

- **Coding transcripts keep the CDS-relative resolution.** `a_coding_transcripts_rna_axis_is_unchanged` and `a_coding_transcripts_terminal_rna_repair_is_unchanged` pin values measured on `origin/main` before the fix. A fix that replaced `cds_relative_gap` / `cds_axis_end` with plain `junction + 1` arithmetic would pass the reproducer and break these — that resolution is what names a junction on the CDS start `r.-1_1` rather than `r.0`.
- **The `n.` axis.** The `n.`/`r.` pair is asserted as *agreement*, not as two independent expected strings, so a future change that moved `n.` onto the same defect would fail rather than silently agree.

## Representation stability

`dump_normalized_corpus` (#1441) reports **0 of 18,432 rows moved**, 0 previously accepted, 0 previously not a fixed point — across all five shape families. That is a weak confirmation rather than strong evidence: the committed corpus covers the `g.` and `c.` axes only, and neither can satisfy the new predicate (a `g.` member is never `Region::Rna`, and the corpus's transcript is coding). The change is provably unreachable from it.

The substantive disclosure is the census above. On the non-coding `r.` axis this **does** move output, on ~849 rows of the affected shape family. But every moved row previously emitted a description that denotes no sequence at all — a repeated member — so there is no well-formed shipped string for a consumer to have stored. This is a correctness fix with no migration attached, not a canonical-form change.

Note this is `main`-to-branch, not release-to-release; the harness postdates the last tag.

## Verification

```
cargo clippy --all-features --all-targets -- -D warnings      # clean
# 9409 tests run: 9409 passed (3 slow), 145 skipped
```

All three oracles plus the full sweep corpus. No proptest strategy touched, so no pinned seeds are affected (`git status --porcelain tests/proptest-regressions/` is empty). No Python surface touched.

The five new tests were written before the fix and confirmed RED against `origin/main` (`left: "NR_TEST.1:r.[9dup;11dup;11dup]"`), with the two coding-transcript guards passing at that point — they discriminate the fix rather than restate it.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -7327,7 +7327,7 @@ fn respell_at_sequence_end<P: ReferenceProvider>(
                 }
         }
         HgvsVariant::Rna(r) => {
-            let Some((region, axis)) = cds_axis_end(span, provider, raw_length, Region::Rna) else {
+            let Some((region, axis)) = rna_axis_end(span, provider, raw_length) else {
                 return;
             };
             let utr3 = region == Region::ThreePrimeUtr;
@@ -7354,9 +7354,15 @@ fn respell_at_sequence_end<P: ReferenceProvider>(
     // outside the member's own region — which is the case this guard exists for,
     // so keying the check on `span.region` would make it vacuously true for the
     // one input it has to judge.
+    //
+    // `rna_axis_end` rather than `cds_axis_end` for the same reason the arm
+    // above uses it: on a non-coding record `cds_axis_end` declines, which made
+    // this verification reject a write the arm had just made correctly (#1453).
+    // The two must be the same resolution or the guard judges the arm against a
+    // convention the arm does not use.
     let written = match variant {
         HgvsVariant::Cds(_) => cds_axis_end(span, provider, raw_length, Region::Cds),
-        HgvsVariant::Rna(_) => cds_axis_end(span, provider, raw_length, Region::Rna),
+        HgvsVariant::Rna(_) => rna_axis_end(span, provider, raw_length),
         _ => Some((span.region, axis)),
     };
     let landed = placed && written.is_some_and(|w| member_endpoints(variant) == Some((w, w)));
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
### Review pass: one behaviour change, quantified; one vacuous guard replaced

Two things came out of reviewing this, one of which moves normalized output and so needs the stability disclosure.

**1. The fallback predicate was narrower than the invariant it states — and that slice still reproduced #1453.**

`rna_axis_is_transcript_relative` required `cds_start.is_none() && cds_end.is_none()`, while its own doc claims parity with `region_sequence_delta` and `axis_frame`. Both of those key on **`cds_start` alone** (`region_sequence_delta`: `Region::Rna => Some(cds_start().map_or(0, ...))`; `axis_frame`: `CisKind::Rna` matches on `cds_start`). So on a record carrying `cds_end` without `cds_start` the three disagreed, the repair declined, and the issue reproduced verbatim:

```
before   NR_MIX.1:r.[9dup;11dup;11dup]     <- the repeated member #1453 is about
after    NR_MIX.1:r.9_10insuaa             <- agrees with n.9_10insTAA on the same record
```

That shape is reachable from real data, not only by direct construction: `CdotTranscript`'s CDS guard is an **or** (`start_codon.is_some() || stop_codon.is_some()`) and then assigns `(self.start_codon, self.stop_codon)` straight through, so a 5'-incomplete record (`cds_start_NF`, #972) arrives as `(None, Some(end))`.

Fixed by keying on `cds_start` alone. The other two refusals are unaffected, because both still carry a `cds_start` for `ordered_cds_bounds` to reject: inverted bounds, and `cds_start` without `cds_end`.

**Representation impact — measured, not estimated.** This moves normalized output only for records with `stop_codon` and no `start_codon`. Counted across the whole prepared reference:

| cdot | transcripts | affected |
|---|---:|---:|
| RefSeq GRCh38 | 482,519 | **1** (`NM_001077693.2`) |
| RefSeq GRCh37 | 197,846 | **3** |
| Ensembl GRCh38 | 574,584 | **0** |

Four records out of ~1.25M, three distinct accessions, and only for `r.` cis-allele repairs on them — where the current output is a repeated member that denotes no sequence. So the movement is malformed → well-formed, not one shipped canonical form to another.

**2. The two "coding is unchanged" tests could not detect an over-broad fix.**

`provider`'s coding arm sets the CDS to the whole transcript, so `cds_start = 1`, the CDS delta is 0, and the CDS-relative and transcript-relative resolutions coincide. Replacing the predicate with a bare `span.region == Region::Rna` — applying the non-coding fallback to *every* `r.` record — left **all six** tests passing. The file's claim that such a fix "would pass the reproducer above and break this" did not hold.

The whole-transcript fixture is deliberate and worth keeping (it makes `c.N`/`n.N`/`r.N` a controlled comparison), so rather than rewriting it I added `a_coding_transcripts_terminal_rna_repair_resolves_through_the_cds`, on a coding fixture with a real 5' UTR where the two conventions diverge:

```
correct     NM_UTRT.1:r.53_54insagg
fallback    NM_UTRT.1:r.[53dup;*4delinsggg]
```

Both new tests were verified as discriminators rather than assumed: each sabotage fails exactly its own test and nothing else, and the restored tree is 7/7.

Full gate green on the rebase: fmt, clippy `--all-features --all-targets -D warnings`, 9492/9492, plus the no-features `--test it` build the soak archive uses.

PR #1466, opened 2026-08-06
test(corpus): re-enumerate the msto issue catalog and let it see Python guards

### LINKED ISSUE #1449: Test: msto_regression_corpus self-checks are stale — four issues missing, three pinned counts wrong, #1235 marked open
`tests/it/msto_regression_corpus.rs` catalogs every issue `msto` has filed and pins three totals plus the open-issue set so that drift in the table fails loudly instead of rotting silently. The catalog itself has now drifted: four issues filed on 2026-08-05 were never added, so every self-check that depends on the table is asserting a stale snapshot.

Re-enumerated today against GitHub:

```
gh issue list -R fulcrumgenomics/ferro-hgvs --state all --author msto --limit 300 \
  --json number,title,state,labels
```

| pin | current value | actual |
|---|---|---|
| `PINNED_ISSUE_COUNT` | 69 | **73** |
| `PINNED_HIGH_ISSUE_COUNT` | 23 | **27** |
| open set in `issue_state_matches_the_known_open_set` | `{1159, 1235}` | **`{1159, 1419, 1420, 1421, 1430}`** |

Two independent movements produced that:

- **#1235 is now CLOSED** (`stateReason: COMPLETED`). The module's doc comment still describes it as "the rewrite's own tracking issue" and marks it `State::Open`.
- **Four issues are missing from `MSTO_ISSUES` entirely**, all filed 2026-08-05 and all labeled `high`:
  - #1419 — Normalizer: multi-member deletion cis-allele not re-derived from the resulting sequence — disagrees with the equivalent `delins` (partition non-uniqueness); non-confluent
  - #1420 — Normalizer: multi-member cis-alleles normalized independently, not re-derived from the resulting sequence — `[dup;del]`, `[ins;del]`, `[delins;del]` each still non-confluent (follow-up to #1231/#1233/#1232)
  - #1421 — Normalizer: net-insertion spanning `delins` not split across an unchanged interior base (follow-up to #1232); non-confluent
  - #1430 — Sequence coercion before normalization

The four `high` additions land as unguarded: no test in `tests/` or `src/` on `main` references any of them by number (`git grep -E 'issue_(1419|1420|1421|1430)|#(1419|1420|1421|1430)\b'` is empty). Per the module's own contract that is a *finding to record*, not an error, so `no_high_priority_issue_is_unguarded_without_being_named` should grow from `[200]` to `[200, 1419, 1420, 1421, 1430]` rather than have the additions quietly omitted.

`PINNED_UNIQUE_TEST_COUNT` (254) and `PINNED_AT_RISK_ISSUES` (the eleven flagged for the sequence-first migration) are unaffected: the four new issues carry no cataloged tests and no mechanism note.

### Scope

Catalog-and-pins only. No normalizer behaviour changes, and no change to which tests guard which issue.

nh13:
Closes #1449

`tests/it/msto_regression_corpus.rs` catalogs every issue `msto` has filed and, for each, the tests that guard it. Its pinned totals had drifted from the repository, and — more importantly — it was under-reporting coverage in the one direction that matters for a report about what is *unguarded*.

## Stale pins

Re-enumerated against `gh issue list -R fulcrumgenomics/ferro-hgvs --state all --author msto`:

- issue count **69 → 73**
- `high`-labeled count **23 → 27**
- open set refreshed

## The catalog could not see Python guards

This is the substantive half. The scan only ever looked at Rust — `tests/` and `src/` — and `source_defines_fn` matched `fn <name>`. The bindings' regression tests live in `tests/python/`, so every guard there was invisible, and four issues were reported as **unguarded when a dedicated pytest module guards each**:

| issue | guard the catalog could not see |
|---|---|
| #1050 | `tests/python/test_issue_1050_protein_render_style.py` |
| #1159 | `tests/python/test_reference_aware_spdi.py` |
| #1245 | `tests/python/test_issue_1245_enum_contract.py` |
| #1246 | `tests/python/test_issue_1246_changed_positions_list.py` |

A false negative in a file whose entire job is reporting what is unguarded is worse than a stale count, because it points work at issues that are already covered.

The fix widens `source_defines_fn` to accept `def` as well as `fn`, and records the guards. The existence self-check therefore now verifies the Python entries too — a rename in `tests/python/` breaks this test the same way a Rust rename does.

## Attribution differs for Python, deliberately

The catalog's rule is that a test references its issue in the function's own name or a comment immediately above it. That rule is Rust-shaped: a Rust suite puts many unrelated tests in one file, so per-function attribution is what makes the mapping meaningful.

The bindings' guards are instead *dedicated per-issue modules* (`test_issue_<N>_*.py`), where the number is in the filename and module docstring while the functions are named for the behaviour they check. So for a Python entry the reference may be in the module docstring or the function's own docstring. That is documented in the module header rather than left implicit.

## Scope

Only issues whose `tests` slice was **empty** were audited, because only those produced a false "unguarded" claim. An issue that already lists a Rust guard may still have an unlisted Python one — that understates coverage without misreporting the issue, and sweeping it would churn the pinned totals for no gain in what the catalog actually asserts.

Unique cataloged test count moves **254 → 265** accordingly.

## Verification

Full gate green: `9410 tests run: 9410 passed, 145 skipped`, with `FERRO_ASSERT_IDEMPOTENT`, `FERRO_ASSERT_REPARSE`, `FERRO_ASSERT_IN_BOUNDS` and `FERRO_SWEEP_SEEDS=full`. Tests only — no `src/` change, so no normalized output moves.

### DIFF
--- a/tests/it/msto_regression_corpus.rs
+++ b/tests/it/msto_regression_corpus.rs
@@ -17,6 +17,32 @@
 //! `tests` slice is empty — that is itself the finding for a `high`-labeled
 //! issue, per [`no_high_priority_issue_is_unguarded_without_being_named`].
 //!
+//! ## Python guards count, and used not to
+//!
+//! The catalog is a report about what is *unguarded*, so a guard it cannot see
+//! is a false negative in the one direction that matters. Until now it scanned
+//! only Rust — `tests/` and `src/` — and the bindings' guards live in
+//! `tests/python/`. Four issues were listed as unguarded while a dedicated
+//! pytest module guarded each: **#1050, #1159, #1245, #1246**.
+//!
+//! Python attribution works differently and the rule above is adjusted for it
+//! rather than bent. A Rust suite puts many unrelated tests in one file, so the
+//! reference has to be per-function. The bindings' guards are instead written one
+//! module per issue, and the issue number is named in the **module docstring** —
+//! so for a Python entry the reference may be in that docstring rather than
+//! beside each function, and [`source_defines_fn`] accepts `def` as well as `fn`.
+//!
+//! Three of the four carry the number in the filename too
+//! (`test_issue_<N>_*.py`); #1159's module is named for its behaviour
+//! (`test_reference_aware_spdi.py`) and names the issue in its first docstring
+//! line. The docstring, not the filename, is what the attribution rests on — a
+//! filename convention that three of four follow is not one to state as the rule.
+//!
+//! Only issues whose slice was **empty** were audited this way, because only
+//! those produced a false "unguarded" claim. An issue that already lists a Rust
+//! guard may still have an unlisted Python one; that understates coverage
+//! without misreporting the issue as unguarded.
+//!
 //! ## What this module does NOT claim
 //!
 //! It does not re-run the shadow/live comparison, so it makes no claim about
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1467, opened 2026-08-06
test(repr): draw a block past the split cap in the representation corpus

### LINKED ISSUE #1460: Representation corpus draws only 20-mers, so a length-gated change measures a structurally guaranteed zero
#1456 fixed one blind spot in `examples/dump_normalized_corpus.rs` — the corpus built no allele whose members conflict. Fixing it (#1459) exposed the same hole one axis over: the corpus cannot express a **long** block either, so any change gated on sequence length measures a zero that was guaranteed before the run started.

## Measured

`#1403`'s gate is length-keyed (`src/normalize/merge.rs`, at `origin/main` `ee0e37ac`):

```rust
const MAX_SPLIT_BLOCK: usize = 1024;                                        // :1791
...
if reference.len() != result.len()
    && (reference.len() > MAX_SPLIT_BLOCK || result.len() > MAX_SPLIT_BLOCK)  // :3536
{
    return whole();
}
```

`corpus_sequences` draws **20-mers**, and the widest shape in `inputs_for` reads five consecutive offsets. No row the generator can emit comes within two orders of magnitude of 1024, so every row takes the same path on both sides of #1403 **by construction**.

Run with the post-#1459 corpus (33,792 rows, nine families), `0ec3c275` against its parent `699fb592`:

```
| total | 33792 |
| **moved** | **0 (0.0%)** |
```

with every one of the nine families at 0. That is not agreement between the two revisions; it is the corpus being unable to pose the question.

## Why this matters

Same argument as #1456. The representation-stability rule asks every output-moving PR to quantify its movement against this harness. A zero from it is currently ambiguous between "nothing moved" and "no row can reach the threshold your change is gated on", and for #1403 only the second is true. #1459 added a caveat sentence to `compare`'s zero case naming the families covered, which makes the ambiguity *visible* — but a family list does not tell a reader that the corpus tops out at 20 bases.

Note the two blind spots have different shapes: #1456's was **which shapes** the corpus builds (fixable by adding families, as #1459 did); this one is **what scale** it builds them at, which no new family fixes.

## Suggested shape

A length axis, kept deliberately small because it is expensive:

- Add one or two long cores (say 1,100 and 2,100 bases) drawn from the same xorshift generator, used by a **single** dedicated family rather than crossed with all nine — the near-palindrome shape `merge.rs:3529-3534` names (`g.257_1356inv` vs `g.[257C>A;267A>C;1346G>T;1356T>G]`) is the one that exercises the cap, and that comment records an exact confluence boundary at 1024/1026 worth pinning.
- Emit those rows behind their own family name so the per-family table shows explicitly whether the length axis was exercised, and so a reader of a zero can see it was.
- Cost check before committing to it: the current 33,792-row dump is ~5 s in a debug build; a kilobase family crossed with the full axis/direction matrix would dominate that. Measure and scope, do not cross it with everything.

## Stability

Corpus-only, `examples/` — no normalized output moves. But note that adding a family re-rolls the row set, so dumps taken before and after are incomparable through `compare` unless the example file is copied into the base checkout first (the workflow #1459 documented in the module header).

nh13:
Closes #1460.

**Stacked on #1459** (`test/issue-1456-conflict-corpus-families`) — same file, and #1460 was found by fixing #1456. Base this on `main` once #1459 merges; the diff below is against #1459's branch.

## The defect

#1456 fixed *which shapes* the corpus builds. This is the same hole one axis over: *what scale* it builds them at.

`corpus_sequences` draws 20-mers and the widest shape reads five offsets, so no row can reach `MAX_SPLIT_BLOCK` (1024, `src/normalize/merge.rs:1791`, gate at `:3536`). A change to a **length**-gated path therefore takes the identical path on both sides for every row in the corpus, no matter how many families exist. #1403 measured `0 of 33,792` even with #1459's four conflict families in place — a zero that was guaranteed before the run started.

## The fix

One family, `long_block_inversion`, drawn against two near-palindromic cores straddling the cap:

| core | block | vs cap |
|---|---|---|
| `nearpalindrome_1024` | 1,024 | at it |
| `nearpalindrome_1100` | 1,100 | over it |

Each emits the confluence pair `partition_block`'s own comment names — the whole-core `inv`, and the same variant spelled as its individual substitutions:

```
NC_TEST.1:g.257_1356inv
NC_TEST.1:g.[267G>C;799A>T;814A>T;1346G>C]
```

A *near* palindrome is what makes that pair writable: an exact one inverts to itself and denotes nothing, a random core inverts to something differing at ~75% of its positions and the allele carries hundreds of members. Four differences is the number that keeps it a real confluence question, and `the_long_cores_are_near_palindromes` pins it.

**16 rows, 33,792 → 33,808.** Dump cost is unchanged at ~5 s in a debug build. The family is deliberately *not* in `FAMILIES`: the cap gates on block length, not on member geometry, so crossing a kilobase core with the other nine families would multiply cost for no coverage.

The long rows carry a **label** (`nearpalindrome_1100`) in the `reference` column rather than the sequence — a kilobase core would otherwise be repeated verbatim on every row. Labels are as good a row identity as sequences, and no existing key moves.

## Does it see what it was blind to?

Across #1403 (`0ec3c275` vs `699fb592`), same harness both sides:

| | old corpus | with this family |
|---|---|---|
| moved | 0 of 33,792 | **2 of 33,808** — both `long_block_inversion` |

```
[g 3prime] NC_TEST.1:g.257_1356inv
   before: NC_TEST.1:g.267_1346inv
   after : NC_TEST.1:g.[267G>C;799A>T;814A>T;1346G>C]
```

The two moved rows are the discriminating ones: **only** the 1,100-base core, **only** the `g.` axis, **only** the `inv` spelling. The 1,024-base core does not move, which is the control that places the movement at the cap rather than anywhere else in #1403.

## Observed, not addressed

On **both** cores the `c.` axis leaves *both* spellings stable while `g.` converges — two normal forms for one variant on the coding axis. It reproduces at 1,024 and 1,100 alike, so it is not length-related.

Since traced to root cause, off this PR. It is **three** stacked gates, each measured by instrumenting the refusal and reducing the case to a 20-base core:

1. `is_splittable_single_member` (`src/normalize/mod.rs:1923`) admits only `Genome`/`Mt` single members, so a lone `c.`/`n.`/`r.` `inv` never enters the sequence-first pass at all — while the *allele* spelling of the same variant does. That gate is deliberate and its own doc comment describes exactly this asymmetry, deferring the widening pending a measurement pass.
2. `join_pos` (`src/normalize/merge.rs:1082`) refuses a member whose start and end are in different regions, so `c.1_*5inv` is dropped by `cis_axis_parts` before anything else runs.
3. `collect_canonical_edits` refuses any member outside the axis's body region — observed firing as `region=ThreePrimeUtr body=Cds`. This one hits **alleles too**, not just lone members, and it is not covered by gate 1's doc comment.

The corpus's `c.` rows hit all three (`CDS_END` is 15 against a kilobase core, so every span is straddling or 3'UTR). Widening gate 1 alone was tried and moves **380 of 33,808** corpus rows (1.1%, 172 previously accepted) while leaving `long_block_inversion` at **0 moved** — it does not fix this. Nothing here is changed by this PR.

## Unchanged on purpose

- The nine short-core families, `corpus_sequences`, the seed count and the row key — a re-rolled corpus makes every earlier dump incomparable.
- The `reference` column's meaning for existing rows.
- Cost: no kilobase core is crossed with a short-core family.

## Representation stability

**No normalized output moves.** `examples/` only, no `src/`. The movement quoted above is #1403's, measured retrospectively with the new corpus.

## Verification

```
# Summary [605.382s] 9415 tests run: 9415 passed (4 slow), 145 skipped
```

Test-first: `the_corpus_emits_a_block_past_the_split_cap` failed on #1459's branch with *"the longest block the corpus builds is 5 bases, under the 1024-base split cap"*. It reads the span back off what `dump` emits, so it measures the corpus rather than the generator's intent.

`the_long_cores_are_near_palindromes` was written after the construction it guards, so its failure was demonstrated rather than observed first: with the two perturbations removed the cores become exact palindromes and it fails with *"inverts with 0 base changes, not 4"*.

### DIFF
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -63,6 +63,14 @@
 //!    `compare` now says so in the zero case rather than leaving it to be inferred.
 //!    If your change is scoped to a shape no family builds, add the family; do not
 //!    quote the zero.
+//! 4. **…and only at the scale it builds them** (#1460). The families are drawn
+//!    against 20-mers, so a change gated on *length* rather than on shape sees the
+//!    same path on both sides no matter how many families exist. #1403 caps a
+//!    partition at `MAX_SPLIT_BLOCK` (1024) and measured a guaranteed zero here
+//!    even after the conflict families landed. `long_corpus_sequences` is the
+//!    answer, and it is deliberately narrow: two cores, one family, 16 rows. Adding
+//!    scale is expensive in a way adding a family is not — check the dump cost
+//!    before crossing a kilobase core with anything.
 //!
 //! ## The corpus is deliberately independent of `tests/it/common/`
 //!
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

Issue #1450, opened 2026-08-06
Cis members shift across exon junctions when the per-member path would clamp them


nh13:
### Summary

`general.md:44` exempts deletions and duplications around exon/exon junctions from the 3' rule on `c.`, `r.` and `n.` reference sequences. The per-member pipeline honours it. The sequence-first cis path does not, so the same deletion is clamped when described alone and shifted across two junctions when described beside a sibling.

Reproduces on `main` with shipped defaults — no feature flags, no environment variables.

### Reproducer

`NM_001234.1` from `MockProvider::with_test_data`: exons at transcript 1-15 / 16-30 / 31-44, `cds_start = 5`, with a G-run spanning c.9-c.33.

```rust
let normalizer = Normalizer::new(MockProvider::with_test_data());
```

| input | output | |
|---|---|---|
| `NM_001234.1:c.10del` | `NM_001234.1:c.11del` | clamped at the exon-1/exon-2 junction — correct |
| `NM_001234.1:c.13del` | `NM_001234.1:c.26del` | clamped at the exon-2/exon-3 junction — correct |
| `NM_001234.1:c.[10del;13del]` | **`NM_001234.1:c.32_33del`** | both members shifted past their junctions and merged |
| `NM_001234.1:c.[9del;13del]` | **`NM_001234.1:c.32_33del`** | same |

Expected for the pair is the two members at the positions each reaches alone: `NM_001234.1:c.[11del;26del]`.

The `r.` axis behaves identically: `NM_001234.1:r.[10del;13del]` -> `NM_001234.1:r.32_33del`.

### Cause

> **Corrected 2026-08-06.** This section originally named `merge::shift_pieces` and proposed clamping the shift by exon edges. That is wrong — a clamp there is unreachable for this defect. The measurement is below; the first investigation is in the comment thread.

The members are already merged, and already past their junctions, **before any partitioning or shifting decision exists**. Instrumenting `canonicalize_from_sequence` at both call sites, on `c.[10del;13del]`:

```
PROBE[partition]      lo=31 hi_ref=33 hi_alt=31 ref_block="GG" alt_block="" pieces=[0..2>]
PROBE[before shift]   [31..33>]
PROBE[after  shift]   [31..33>]
PROBE[after coalesce] [31..33>]
                                       => NM_001234.1:c.32_33del
```

Three things follow.

[...]

PR #1468, opened 2026-08-06
test(repr): reach #1454's shape, and close the corpus over its output vocabulary

### LINKED ISSUE #1463: Representation corpus emits no delins member that is secretly an inversion, so it reports full idempotency while #1454 is live
### Summary

`examples/dump_normalized_corpus.rs` is the corpus behind release-time representation claims. Run on both sides of a release it reports **zero** non-idempotent rows — while #1454 is open and reproduces on the same `main`.

The corpus cannot exhibit that shape. Of the **8,706** `delins` members it emits across all five families, **zero** have a payload equal to the reverse complement of their own span. #1454 is exactly that case: a `delins` member that should have been typed `inv` and is only re-typed on a second pass. The corpus never builds one, so it can never observe the deferred typing.

This is not a claim that the defect is unguarded — `FERRO_ASSERT_IDEMPOTENT` catches #1454, which is how it was found. The gap is that the **representation** corpus is blind to it, and that corpus is what a release note's stability claims rest on.

### Measured on `origin/main` @ `ee0e37ac`

Reproducer (63-nt core per #1454 — note the core in that issue is actually **64** nt; CDS 1..len):

```
core: TTTTTTTTTAATATATTTTAATATAATTAAAAAAATAATTTTTATAAATATATTATTTTAAAAA

NM_TEST.1:c.10_15delinsTAATAT
  pass 2 -> NM_TEST.1:c.[10_12delinsTAA;13_15inv]      <- pass 1 was not a fixed point
  pass 3 -> (fixed point)
```

Corpus census, 24 seeds / 18,432 rows, both axes, both directions:

| | |
|---|---|
| rows where `norm(norm(x)) != norm(x)` | **0** |
| `delins` members emitted | 8,706 |
| …whose payload == revcomp(span) | **0** |

By span length: 5,068 at 2 nt, 842 at 3 nt, 522 at 4 nt, 2,274 at 5 nt — none an exact reverse complement.

The corpus is not short of adjacency or of inversions in general: it builds 1,937 zero-separation member pairs, and main emits `inv` on 41 rows where it types correctly on the **first** pass. What is missing is specifically the case where the typing is *deferred* — a member spelled `delins` that is an inversion.

### Why it is worth fixing

Two properties get conflated when the corpus reports clean. "No non-idempotent rows" reads as a property of the release; here it is a property of the corpus. That number was reported to me as reassurance during a v0.12.0 → v0.13.0 comparison, and it was wrong in the direction that gets a release waved through.

It also blunts the harness for its stated purpose. #1442 exists so representation claims stop being hand-rolled per PR; a family it cannot generate is a class of representation movement that will keep being missed by every future release comparison.

### Suggested scope

Add a family that draws a spanning `delins` whose sub-pieces are exact reverse complements of their spans — the shape in #1454 — across both axes and both directions. The existing `split_vs_spanning_delins` family is close: it already draws two `delins` separated by one unchanged base; it just never chooses a revcomp payload.

Constraints from the harness's own module docs:

- Changing what an **existing** family emits makes old dumps incomparable for that family. So this should be a **new** family, not a widening of `split_vs_spanning_delins`, or every dump taken before it becomes undiffable.
- Adding a family is explicitly sanctioned ("Adding one is fine").

**Stability cost: none.** This is test-corpus surface only; no normalizer behaviour changes and no normalized output moves. It will, however, make the *next* comparison report non-zero non-idempotent rows until #1454 is fixed — which is the point.

### Related

- #1454 — the defect this corpus cannot see (open).
- #1456 — sibling gap: the corpus builds no overlap-conflicting alleles.
- #1442 / #1441 — the harness itself.

Both #1456 and this one are the same shape of problem: the corpus is narrower than the confidence placed in its output. Worth considering whether they should be worked together.

nh13:
Closes #1463.

## The defect

`examples/dump_normalized_corpus.rs` is the corpus behind release-time representation claims. Run on both sides of a release it reported **zero** non-idempotent rows — while #1454 was open and reproducing on the same `main`.

It could not have found it. Of the **8,706** `delins` members its five families emit, **none** has a payload equal to the reverse complement of its own span, which is precisely #1454's shape: a piece spelled `delins` on the first pass and re-typed `inv` on a second, so the first output is not a fixed point.

The corpus was not short of adjacency or of inversions generally — it builds 1,937 zero-separation member pairs, and `main` emits `inv` on 41 rows where it types correctly on the **first** pass. What was missing is specifically the case where typing is *deferred*.

This matters because "no non-idempotent rows" reads as a property of the release when it was a property of the corpus. That is the direction that gets a change waved through.

## The fix

Adds a sixth family, `delins_hiding_an_inversion`: a length-preserving spanning delins whose trailing three-base piece is an exact reverse complement, plus the pre-split spelling of the same variant (which is #1454's pass-1 output fed back in).

Measured on `ee0e37ac`:

| | rows |
|---|---|
| non-idempotent rows before | **0** |
| non-idempotent rows after | **30** |

All 30 are on the `c.` axis, 15 in each shuffle direction, and every one is #1454's signature:

```text
NM_TEST.1:c.1_6delinsACAGGG
  pass 1 -> NM_TEST.1:c.[1_3delinsACA;4_6delinsGGG]
  pass 2 -> NM_TEST.1:c.[1_3delinsACA;4_6inv]
```

## The piece width is load-bearing, and was measured rather than assumed

The first version of this family used a two-base head with a three-base tail. It reaches **0** non-idempotent rows — the trailing piece gets typed `inv` on the first pass and the defect never appears. Three-and-three reaches **30**. `the_inversion_family_uses_three_base_pieces` pins that, because a future edit narrowing either piece would silently return the corpus to being blind to the defect it was added for.

An odd-length piece also cannot equal its own reverse complement, so neither degenerates to a no-op the way a two-base piece does on `AT`, `TA`, `CG` or `GC`.

## Unchanged on purpose

- **The five existing families are byte-identical to `origin/main`** across all 18,432 rows (`diff` clean). The new shape reads six offsets where the others read five, so it is bounded inside the family rather than by widening the shared `last` — moving that would have re-rolled every existing family and made every prior dump incomparable.
- **`head_payload` never degenerates.** Checked exhaustively over all 64 three-base spans rather than trusting the pigeonhole argument: it is never the span (a no-op) nor its reverse complement (a second inversion, which would stop the case isolating one mis-typed piece).
- **CI does not go red.** 8 of the 30 fall within the 3 seeds the example's own unit tests draw, but `normalize_one` wraps normalization in `catch_unwind` and records a panic as data — by design, per its doc comment. The example's 10 unit tests pass with all three oracles set, verified both ways.

## One behaviour change worth knowing

`--compare` refuses a dump taken with pre-#1463 code against one taken after, because the key sets differ:

```text
error: the two dumps do not cover the same corpus (0 rows only in the baseline,
5760 only in the candidate). Re-dump both with the same --seeds; ...
```

That is the existing guard doing its job — loud, not silent — and the documented workflow already avoids it, since dumping "at the base revision" means putting the harness on that revision, which yields matching corpora.

## Representation stability

**No normalized output moves.** This is test-corpus surface only; no normalizer code is touched. The 30 rows are not new movement — they are movement the corpus previously could not see.

## Verification

```text
cargo fmt --all --check                                              clean
cargo clippy --all-features --all-targets -- -D warnings             clean (exit 0)
  cargo nextest run --features dev                                   9414 passed, 0 failed
```

`tests/proptest-regressions/` untouched (no `*_strategy()` in this diff). No Python surface touched.

### DIFF
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -254,6 +254,42 @@ const FAMILIES: &[(&str, &str)] = &[
         "delins_hiding_an_inversion",
         "#1454 — a spanning delins whose trailing piece is an exact reverse complement",
     ),
+    // The four below close the corpus over its own output vocabulary: it *emitted*
+    // constructs it never fed back in, so a whole class of input went unmeasured.
+    // #1454 was one consequence — the corpus produced `inv` outputs but never an
+    // input whose derivation had to *decide* on `inv` — and it was found by accident
+    // rather than by the corpus, so there was no reason to think it was the only one.
+    //
+    // Measured on `5a19d1d2`, over 33,792 rows:
+    //
+    // | construct        | outputs | inputs |
+    // |------------------|--------:|-------:|
+    // | `inv`            |   3,113 |  3,072 |
+    // | repeat `N[k]`    |   1,558 |      0 |
+    // | `=`              |     152 |      0 |
+    // | 3+ member allele |     170 |      0 |
+    //
+    // `inv` is the interesting row and the reason `inv_member` is narrower than the
+    // other three. #1459 closed the raw count when it added `junction_interior_to_span`
+    // — but every one of those 3,072 inputs pairs the `inv` with an insertion that
+    // *overlaps* it, and none is lone. So the corpus could only ask what happens to an
+    // `inv` already in conflict, never what happens to a well-formed one.
+    (
+        "inv_member",
+        "an `inv` as an input, lone and beside a non-overlapping sibling — #1459 feeds `inv` only in a conflicting pairing",
+    ),
+    (
+        "three_member_allele",
+        "a three-member allele as an input — emitted on 170 rows, consumed on none",
+    ),
+    (
+        "identity_member",
+        "a no-change `=` member as an input — emitted on 152 rows, consumed on none",
+    ),
+    (
+        "repeat_expansion",
+        "a ranged repeat as an input — emitted on 1,558 rows, consumed on none",
+    ),
 ];
 
 /// The family drawn against the **long** cores, and the one shape in this file
@@ -376,6 +412,23 @@ fn long_inputs_for(axis: &str, core: &str) -> Vec<String> {
         return vec![whole];
     }
     vec![whole, format!("{prefix}[{}]", members.join(";"))]
+}
+
+/// Offset of the first tandem three-base repeat in `core`, if it has one.
+///
+/// Three bases rather than one, because `repeated.md:21` permits a repeat description
+/// on a `c.` reference **only** for units whose length is a multiple of three — a
+/// mononucleotide repeat there is spec-invalid (`c.2686A[10]` is called out by name),
+/// and a corpus that emits invalid input measures nothing useful. Three also lets the
+/// ranged spelling be used throughout, which is the form the pending Community
+/// Consultation proposal in that file's NOTE would leave as the only legal one.
+///
+/// Only 21 of the 48 corpus sequences contain such a tract, so `repeat_expansion`
+/// emits nothing for the rest. That is deliberate: synthesising a tract would mean
+/// describing a repeat the reference does not have.
+fn first_tandem_triplet(core: &str) -> Option<usize> {
+    let b = core.as_bytes();
+    (0..b.len().saturating_sub(5)).find(|&i| b[i..i + 3] == b[i + 3..i + 6])
 }
 
 /// A three-base payload for the leading piece of `delins_hiding_an_inversion`,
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
## Second commit — closing the corpus over its output vocabulary

`fc1fd900` widens the scope beyond #1463, deliberately. Adding families re-rolls `--compare` comparability, so doing this as a separate PR would have cost a second break for no benefit; one widening, one break.

### Why

An audit of what the corpus varies against what it pins found that #1454's blind spot was one instance of a general asymmetry — the harness **produces** constructs it never **consumes**. Measured on `ee0e37ac` across 24,192 rows:

| construct | in outputs | in inputs |
|---|---:|---:|
| `inv` | 800 | **0** |
| repeat `N[k]` | 1,370 | **0** |
| no-change `=` | 56 | **0** |
| 3+ member allele | 298 | **0** |

Each is a form the normalizer writes and a consumer can hold and feed back, so each is a class of input the corpus could not measure. #1454 was found by accident rather than by the corpus, and there was no reason to assume it was the only one.

### Result, stated plainly

Four new families feed each construct in. **No new oracle fire.** This is preventive — it removes structural blindness rather than cashing it in — and it would be misleading to present it as a defect find.

### Three constraints, not preferences

- An `inv` input spans **three** bases. An even-length inversion is a no-op on a palindromic span (`AT`, `TA`, `CG`, `GC`), which would quietly make the input an identity.
- A repeat input uses a **three-base unit in the ranged spelling**. `repeated.md:21` permits repeat notation on a `c.` reference only for units whose length is a multiple of three — a mononucleotide repeat there is spec-invalid, and it is called out by name.
- `repeat_expansion` describes only a tract the reference **actually has**, so it is silent on 27 of the 48 corpus sequences. Synthesising a tract would assert something untrue about the reference.

`the_corpus_is_closed_over_its_output_vocabulary` guards against the asymmetry returning. It caught a real bug in its own predicate while being written (`=` is followed by `;`, not `]`), which is a reasonable sign it is checking something.

### What this does NOT fix

Still pinned: **reference length** (20), **strand** (Plus), **exon count** (1), **`CDS_START`** (1). The last two matter concretely:

- **#1450 is open and remains unreachable.** It is a cis-member exon-junction defect whose reproducer needs a 3-exon transcript with `cds_start = 5`; this corpus has one exon and `cds_start = 1`.
- **`c.-N` 5'UTR positions are never generated at all**, because `CDS_START = 1`. Worth noting the module doc deliberately engineered 3'UTR coverage — "five 3'UTR bases remain… #1398, #1418 and both halves of #1426 lived in that region" — while leaving the other junction structurally unreachable.

Unpinning the transcript geometry is the remaining half and is a larger change; it is not in this PR.

### Verification

```text
cargo clippy --all-features --all-targets -- -D warnings   clean (exit 0)
all three oracles + FERRO_SWEEP_SEEDS=full                 9416 passed, 0 failed
corpus rows                                                24,192 -> 36,648
```

PR #1470, opened 2026-08-06
feat(normalize): coalesce payload-alignment splits back into one delins

nh13:
Stacked on #1464, which has since merged; this now targets `main` and the diff is the single commit.

## What this changes, and what it does not

**No shipped representation moves.** With `FERRO_PARTITION` unset — every real invocation — this adds one unreachable match arm and changes nothing. The new behaviour is reachable only via `FERRO_PARTITION=canonical-coalesced`.

## The problem

`CanonicalAlignment` minimises member count *strictly within* the minimal-distance alignment DAG, so it splits a block wherever a payload base happens to coincide with a reference base. `DNA/delins.md:44-47` names that shape — "parts of the inserted sequence *align* with the reference sequence" — and recommends the delins instead. It is the largest single class of churn the canonical arm introduces.

It cannot be fixed inside the partitioner. On the spec's own worked example, `LRG_199t1:c.850_901delinsTTCCTCGATGCCTG`, the recommended single `delins` claims **52** changed columns against a block edit distance of **39** — so it is not in the minimal-distance set at all, and no ranking within that set can reach it. The canonical arm emits **five** members there, more fragmented than the four-member split the spec explicitly names as the one to avoid.

## The pass

Three conditions, each measured rather than chosen:

| condition | value | why |
|---|---|---|
| payload embeds in the span | ≤ 1 substituted position | the spec's worked example needs exactly 1; the dominant churn class needs 0 |
| span must lose length | strict | two substitutions 6 bases apart embed with no deletion (`delins.md:17`); and #1421's net-insertion family runs *opposite* to #1419/#1420 — for a net insertion the split form is canonical |
| no separation wider than | 8 bases | every correct rejoin observed lies at 1–8; genuinely separate cis members lie at 9–1562 |

**It runs after the weight bound, and that placement is the fix.** The bound compares derived pieces against the **input's own** edits, so a widening pass judged there is accepted for one spelling of a variant and refused for another — the same hazard `coalesce_whole_block_inversion` is already placed after it to avoid. Judged before the bound, this pass cost **427 converged classes** per direction, with 427 regressions and 0 gains. Placement relative to the 3′-shift and width minimisation is *not* the mechanism — moving across those was byte-identical.

## Measured, against the real prepared reference

| | live (shipped) | canonical | this |
|---|---|---|---|
| confluence, 11,272 classes (5′) | 6,627 | 8,387 | **8,753** |
| confluence, 11,272 classes (3′) | 6,627 | 8,408 | **8,788** |
| ClinVar 500k, rows moved | — | 192 | **129** (78 restored) |
| 592 real multi-member cis, rows moved | — | 16 | **19** |

`split 3` and `split 4+` are zero in both directions.

## The honest cost, stated because it is a representation change

The cis arm is worse than canonical: **3 rows** where genuinely separate variants are merged. They sit at separations of 1 and 4 — inside the same 1–8 range every correct rejoin occupies — so separation cannot distinguish them, and neither can any other property of the sequence.

`delins.md:79-84` gives the spec's own discriminator: *"the two variants may have been reported (or might occur) individually"* — **provenance**, which lives in the input's spelling and nowhere in the reference or observed bases. Reading it back is precisely what forfeits confluence (see the 427-class measurement above). #1235 chooses confluence, so the string stays provenance-free and the three cases are **pinned by a test that explains the trade** rather than silently fixed.

Follow-up worth doing separately: emit a diagnostic when the pass merges what the input spelled as separate members. That satisfies `delins.md:79-84`'s actual concern — a database query missing the individually-reported form — without the canonical string depending on input spelling. Left out of this PR because a new public warning code is its own API review.

## Tests

A **5,760-case generated corpus** — separations 0→1562, four length shapes, three alphabets, 2–5 members, seeded xorshift so failures reproduce exactly — asserting **invariants rather than a restatement of the rule**: denotation preservation, no merge past the cap, no merge without length loss, member count non-increasing, idempotence, and monotonicity in separation. Plus a vacuity guard (it fails if fewer than 100 cases actually merge), a monotonicity walk, and the spec's worked example and the dominant churn class with their real bases.

Gate: `cargo fmt --check` and `clippy --all-features --all-targets -D warnings` clean; `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1 FERRO_SWEEP_SEEDS=full cargo nextest run --features dev` → `9475 tests run: 9475 passed, 145 skipped`.

Refs #1235, #1444

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2028,6 +2028,10 @@ enum PartitionRule {
     /// [`partition_block_canonical`]: the member-count-minimal minimal alignment
     /// (`mutalyzer-algebra`'s `canonical`).
     Canonical,
+    /// `Canonical`, plus the terminal `delins.md:44-47` re-spelling applied
+    /// after the downstream passes — see
+    /// [`coalesce_payload_alignment_split`].
+    CanonicalCoalesced,
 }
 
 /// Read a [`PartitionRule`] from what `FERRO_PARTITION` was set to.
@@ -2047,8 +2051,12 @@ fn partition_rule_from_env(value: Option<&str>) -> PartitionRule {
         None | Some("") | Some("live") => PartitionRule::Live,
         Some("shadow") => PartitionRule::Shadow,
         Some("canonical") => PartitionRule::Canonical,
+        Some("canonical-coalesced") => PartitionRule::CanonicalCoalesced,
         Some(other) => {
-            log::warn!("FERRO_PARTITION={other} is not one of live|shadow|canonical; using live");
+            log::warn!(
+                "FERRO_PARTITION={other} is not one of \
+                 live|shadow|canonical|canonical-coalesced; using live"
+            );
             PartitionRule::Live
         }
     }
@@ -2287,6 +2295,13 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
             partition_block_canonical(&ref_bytes[lo..hi_ref], &result[lo..hi_alt])
                 .unwrap_or_else(|| partition_block(&ref_bytes[lo..hi_ref], &result[lo..hi_alt]))
         }
+        PartitionRule::CanonicalCoalesced => {
+            // Identical to `Canonical` here on purpose. The `delins.md:44-47`
+            // merge is applied *after* the downstream passes below, not as part
+            // of partitioning — see `coalesce_payload_alignment_split`.
+            partition_block_canonical(&ref_bytes[lo..hi_ref], &result[lo..hi_alt])
+                .unwrap_or_else(|| partition_block(&ref_bytes[lo..hi_ref], &result[lo..hi_alt]))
+        }
     };
     // Shadow the sequence-first splitter on the very same trimmed block, before
     // any 3'-shift or coalescing, so a reported disagreement is a disagreement
@@ -2301,8 +2316,9 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
         // every block — a full denominator and zero disagreements, which reads
         // as the strongest possible result and is produced by comparing a thing
         // with itself. That is the precise failure the denominator below exists
-        // to rule out. `canonical` would not be vacuous but would label its
-        // output `live=`, which is worse than useless in a bake-off log.
+        // to rule out. `canonical` and `canonical-coalesced` would not be
+        // vacuous but would label their output `live=`, which is worse than
+        // useless in a bake-off log.
         //
         // The extra `partition_block` call is paid only when the shadow audit is
         // on, which is already a measurement-only path.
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
### Review pass: a merged fix was reverted by this branch's rebase, plus a tautological guard

**1. This branch re-introduces #1464's audit vacuity, and deletes the test that would have caught it.** This is the important one.

The diff against `origin/main` removed the independent live baseline:

```diff
-        let live = partition_block(&ref_bytes[lo..hi_ref], &result[lo..hi_alt]);
-        let mut min_live = live.clone();
+        let mut min_live = pieces.clone();
```

along with the `SEQFIRST_AGREE` comparand, both `live={:?}` log operands, the comment block explaining why the binding exists, **and** the regression test `live_and_canonical_disagree_so_the_audit_baseline_is_observable`.

Since `FERRO_PARTITION` was added, `pieces` is whatever rule the environment selected. So with the baseline gone, `FERRO_SEQFIRST_SHADOW=1 FERRO_PARTITION=shadow` compares the sequence-first splitter **with itself** and logs `SEQFIRST_AGREE` on every block — a full denominator with zero disagreements, which reads as the strongest possible result and is worth nothing. That is precisely the failure the denominator's own comment says it exists to rule out.

Almost certainly a conflict resolved toward the pre-fix side: this branch was stacked on #1464 at `07c66bcb`, and the fix landed in the squashed `02a204de`. Nothing went red because the guard was removed in the same diff. Restored all five uses, the explanatory comment, and the test.

**2. `widening_the_separation_is_monotone` could not fail.** The flag was cleared *before* the assertion:

```rust
let merged = pieces.len() == 1;
if !merged { still_merging = false; }
assert!(merged || !still_merging, …);
```

`merged` satisfies the left side; `!merged` has just set `still_merging = false` and satisfies the right. Tautology for every input. The condition was inverted as well — a non-merge while still merging is the normal transition this walk looks for; a merge *after* one is the violation.

Verified rather than argued. Injecting a merge at the widest gap (`|| gap == COALESCE_MAX_SEPARATION + 4`, where merging has certainly stopped):

- original assertion → **passes**
- corrected assertion (asserted before the update, on `!merged || still_merging`) → **fails**: `merge reappeared after it stopped [gap=12 shape=Deletion members=2 span=CATCTCGATGGCGA]`

**3. The knob's newest value was the one value nothing tested.** `the_two_alternative_rules_are_selected_by_name` covered `shadow` and `canonical` but not `canonical-coalesced`, so a typo in the new match arm would have surfaced only as a silent fallback to `Live` during a bake-off — the exact failure the knob's doc says it is designed to make loud. Added, and renamed the test since it no longer covers two. Also corrected two docs still reading "three partitioners".

### Representation impact: none, by construction

`coalesce_payload_alignment_split` has exactly **one** production call site (`merge.rs:2412`), gated on `partition_rule() == PartitionRule::CanonicalCoalesced`. The default is `Live` (`None | Some("") | Some("live")`), and every other call site is past the `#[cfg(test)]` boundary at line 8216. So no shipped output moves — this is an opt-in measurement rule, which is the shape this repo asks for before promoting a partitioner.

Worth stating explicitly because a corpus zero would *not* have established it: the same harness reports zero for changes it structurally cannot express, so the gate is the argument, not the count.

### Not applied

- **Rename `live=`/`min_live=` to `selected=`.** This would bless the reverted state rather than restore it. The shadow audit's whole purpose is comparing shadow against *live*; naming the selected rule `selected=` keeps `FERRO_PARTITION=shadow` self-comparing.
- **DP row-buffer preallocation and the malformed-piece `debug_assert!`.** Both are on the non-default path and neither changes behaviour; out of scope for a change already this large.

Gate green on the rebase: fmt, clippy `--all-features --all-targets -D warnings`, 9509/9509, plus the no-features `--test it` build the soak archive uses.

PR #1474, opened 2026-08-06
test: pin the reported partition family's per-spelling verdicts and spec arguments

nh13:
Three issues — #1419, #1420, #1421 — report one defect from a downstream pipeline: a variant written two ways normalizes to two different strings. Each issue body states four things about each of its rows, and states them in prose:

1. what ferro prints for **each** of the two spellings,
2. which of the two the recommendations make canonical, and on which line,
3. that `EquivalenceChecker` still calls the two one variant (`SequenceMatch`),
4. the reference bases the argument in (2) rests on.

Only (1)'s *pairwise disagreement* was ever guarded. `reported_confluence_pairs` pins that the two spellings must converge and deliberately declines to pin *which* string they converge on — freezing a winner there would settle an open product decision (#1235) by accident. That is the right call for a ratchet on a future fix, and it leaves all four claims above untested: today's answers, the spec argument for them, the equivalence fallback, and the reference itself.

This adds `tests/it/reported_partition_verdicts.rs` — **18 rows, two per reported pair**, plus the premises each row's argument rests on. It is the present-tense half (what ferro does now, and why that is or is not what the recommendations ask for) where `reported_confluence_pairs` is the future-tense half. Neither subsumes the other; they assert disjoint things.

## What running them found

**Every claim in all three issues reproduces byte for byte on current `main`.** All 18 normalized strings, all 9 `EquivalenceChecker` levels, and every reference base the arguments cite. Nothing had drifted since filing, and nothing had to be re-derived by hand to find that out — which is the point of committing them.

**The family has an exact shape, and it is now an assertion rather than three prose descriptions.** In every one of the nine pairs, the canonical form is reachable from exactly one spelling and never from the other:

| | reachable from | not reachable from |
|---|---|---|
| #1419 r1–r3 | the spanning `delins` | the `[del;del]` cis allele |
| #1420 v2–v4 | the spanning `delins` | the `[dup;del]` / `[ins;del]` / `[delins;del]` cis allele |
| #1421 n1–n3 | the split cis allele | the spanning `delins` |

Both spellings denote one variant, so the canonical form is a property of the variant and should be reachable from either. It is reachable from one, which makes reachability a function of how the variant was *written*. `each_pair_reaches_its_canonical_form_from_exactly_one_spelling` states that directly, and is written to fail the day any pair reaches it from both — so progress is as loud as regression.

Note the direction flips between #1421 and its two siblings: there the *split* form is canonical and the span is retained, the reverse of #1419/#1420. A fix that simply preferred whichever spelling has fewer members would get one third of this family backwards.

**#1420's spec argument is now computed from the reference rather than taken on trust.** Its three rows have equal-length spans, so the changed/unchanged column pattern is a fact about the sequence, not a reading: v2 is `X.XX`, v3 is `XX.X`, v4 is `XXXX`. That is exactly why v2/v3 separate under `delins.md:17` and v4 coalesces under `delins.md:16` — the opposite correction, from the same defect. `reported_spans_change_the_columns_the_reports_state` derives all three from `TEMPLATE` and pins that two of the three are separable.

**#1421's rows are separated by two nucleotides, not one.** The `delins.md:18` exception (two variants separated by *one* nucleotide, together affecting one amino acid) is the only reading under which the retained spans would be correct, and it turns on that number. All three rows fail it twice over — separation of two, and a `g.` description has no amino acid. Pinned in `the_1421_spans_separate_by_two_nucleotides_not_one` so the exception cannot be invoked by accident later.

## Convention: characterization, not `#[ignore]`

Nine of the eighteen rows print something the cited recommendation argues against. They are pinned **as they behave today**, with the wanted answer written into the row, rather than expressed as an `#[ignore]`d test of the wanted answer.

That is deliberate. An `#[ignore]`d test never runs, so it cannot notice the day the answer moves — and for this family the *move* is the event that matters, since a normalized string is the key a downstream consumer stores counts against. Pinned, a change surfaces in a diff and has to be argued for in review; ignored, it happens silently. `OPEN_GAPS` counts the disagreeing rows (9 today) so closing one means editing a row *and* the census, deliberately.

The premise test `reference_bases_are_what_the_reports_state` runs first for the same reason: every canonical argument is a claim about which reference positions change, so a mistranscribed contig would leave the rows green — they only compare ferro against itself — while silently turning every `argument` field into fiction.

## Scope

- **No `src/` change. Tests only, and no normalized output moves** — every row pins what `main` already prints, verified by running it.
- **Deliberate overlap to resolve later:** the nine *pairs* are also listed in `reported_confluence_pairs` (#1458), which asserts convergence, sequence preservation and idempotence over them. This module asserts per-spelling outputs, the spec argument and the equivalence level — disjoint assertions over a shared row set. The one place they come close is idempotence, and the claims differ on purpose: #1458 normalizes twice and checks the second pass changes nothing, which a row that moved once and then settled would also satisfy; `every_gap_row_is_returned_exactly_as_authored` checks the input was *never moved*, which is the stronger thing the reports observed and what makes a gap row a second canonical form rather than an under-applied pass. Once both land, the row list is worth hoisting into one place so the two modules cannot drift apart on which pairs exist.
- **#87, #179 and #1421's diagnostic table** are harvested separately in #1457 and are not repeated here.
- **#200 yields no harvestable rows.** Its HGVS strings are illustrations inside a proposed API signature for functionality that does not exist yet — the accession/coordinate correspondences were never asserted to hold, so turning them into tests would mean inventing expectations. Recorded rather than silently skipped.

Refs #1419, #1420, #1421, #1235.

## Verification

```
cargo fmt --all --check                                     clean
cargo clippy --all-features --all-targets -- -D warnings   clean
```

The pins were mutation-checked rather than assumed to bite: perturbing one row's `output` and `OPEN_GAPS` turns three of the eight tests red, including `each_pair_reaches_its_canonical_form_from_exactly_one_spelling`.

Full suite, both configurations — the gating `Test` job runs *without* the oracle variables, so both were run:

```
    Summary [  40.220s] 9421 tests run: 9421 passed, 145 skipped

    Summary [ 395.022s] 9421 tests run: 9421 passed (4 slow), 145 skipped
```

### DIFF
--- a/tests/it/main.rs
+++ b/tests/it/main.rs
@@ -396,6 +396,7 @@ mod repeat_count_trans;
 mod repeat_deln_uncertain_range;
 mod repeat_span_sibling_overlap;
 mod repeat_tract_maximization;
+mod reported_partition_verdicts;
 mod rewrite_target_corpus;
 mod rna_coding_consistency;
 mod rna_spl_marker;
--- a/tests/it/reported_confluence_pairs.rs
+++ b/tests/it/reported_confluence_pairs.rs
@@ -43,7 +43,7 @@ use ferro_hgvs::ShuffleDirection;
 ///
 /// Reproduced verbatim from the reports so these rows run against the same bases
 /// they were observed on. Synthetic: it names no real sequence.
-const TEMPLATE: &str = concat!(
+pub(crate) const TEMPLATE: &str = concat!(
     "ATGCACCAGTCACCAGTCTGATGCGGATCACGTGCAATTGCACGTGCAATTGGATCCGATCG",
     "TACGTACGATCGGCATGCATGCTAGCTAGCATCGATCGTAGCTAGCTAGCATCGATCGATCGA"
 );
@@ -54,7 +54,7 @@ const TEMPLATE: &str = concat!(
 /// the same variant. The two are kept distinguishable rather than collapsed into
 /// an unordered pair because *which one a consumer has stored* decides what a
 /// convergence fix costs them.
-const REPORTED_PAIRS: &[(&str, &str, &str)] = &[
+pub(crate) const REPORTED_PAIRS: &[(&str, &str, &str)] = &[
     // #1419 — two deletions versus the spanning delins.
     (
         "1419-r1",
--- a/tests/it/reported_partition_verdicts.rs
+++ b/tests/it/reported_partition_verdicts.rs
@@ -1,0 +1,702 @@
+//! Per-spelling verdicts for the externally reported partition family
+//! (#1419, #1420, #1421).
+//!
+//! # What these rows are, and why they were not already runnable
+//!
+//! Three issues report the same defect from a downstream pipeline: one variant,
+//! written two ways, normalizes to two different strings. Each issue body states
+//! four things about each of its rows, in prose:
+//!
+//! 1. what ferro prints for **each** of the two spellings,
+//! 2. which of the two the HGVS recommendations make canonical, and on which
+//!    line,
+//! 3. that `EquivalenceChecker` still calls the two one variant
+//!    (`SequenceMatch`) — which is what makes the divergence a *representation*
+//!    problem rather than a correctness one, and what leaves consumers a
+//!    working fallback,
+//! 4. the reference bases the argument in (2) rests on.
+//!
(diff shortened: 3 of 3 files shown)


nh13:
@coderabbitai pause

PR #1475, opened 2026-08-06
perf(python): release the GIL across reference reads, and stop deep-copying the reference per call

### LINKED ISSUE #1455: Python SPDI bindings hold the GIL across reference I/O
`to_spdi`, `canonical_spdi` and `apply_to_reference` in `src/python.rs` run their reference-backed work while holding the GIL. Each of them reads reference sequence through a provider — for a FASTA- or mmap-backed provider that is real file I/O, and `canonical_spdi` can now issue several reads as its window doubles through a repeat tract.

Holding the GIL across that serialises every other Python thread in the process for the duration. The conventional fix is to clone the owned inputs and run the call inside `py.detach(...)`, converting the result or the mapped `ProjectionError` afterwards.

## Why this is filed rather than fixed in #1374

#1374's scope is the SPDI key's *shape* — rolling it 3' and declining an order it was not given. The GIL question is orthogonal, applies to all three entry points rather than the one that PR changes, and wants its own before/after measurement on a multi-threaded caller to be worth anything. Folding it in would mix an unmeasured concurrency change into a PR whose blast radius is currently one moved assertion, and that is exactly the mixing the contribution guidelines warn against.

## Suggested shape

Take `Python<'_>`, clone the variant and provider handles, wrap each call in `py.detach`, and cover it with a test that runs two threads through `canonical_spdi` against a slow provider and asserts they overlap. Worth checking the other reference-backed bindings at the same time — this is unlikely to be the only place.

### LINKED ISSUE #1473: Python bindings deep-copy the entire reference on every call
`PyProvider` is the reference handle every Python binding holds:

```rust
enum PyProvider {
    Mock(Box<JsonProvider>),
    MultiFasta(Arc<MultiFastaProvider>),
}
```

Each reference-backed method builds a fresh normalizer per call — `Normalizer::with_config(self.provider.clone(), self.config.clone())` — so whatever the variant holds is copied once per Python call. For `MultiFasta` that is an `Arc` bump, and the enum's own doc comment says the `Arc` is there precisely because the value "is cloned on every Python call". For `Mock` it is not: `JsonProvider` derives `Clone`, and its `genomic_sequences` and `proteins` maps own their `String`s, so `.clone()` deep-copies the whole reference.

The `Box` was chosen for a size reason (`clippy::large_enum_variant`), which it does satisfy — it just does not address the cost, and the two were never separated.

## Measured

Release wheel, 12-core host. The variant is fixed at a 6-base deletion on a 600-base transcript placed at the start of the contig, so the work each call performs is identical in every row; only the remainder of the contig — which no call reads — grows.

| contig bases | `normalize_variant` | `to_spdi` |
|---:|---:|---:|
| 10 000 | 1.2 µs | 1.0 µs |
| 100 000 | 6.8 µs | 6.1 µs |
| 1 000 000 | 21.8 µs | 17.5 µs |
| 4 000 000 | 443.2 µs | 74.0 µs |

That is ~9 GB/s, i.e. memcpy speed, for bases the call never touches. Extrapolating, a single human chromosome in `genomic_sequences` costs tens of milliseconds and a full re-allocation of itself on *every* call.

## Scope

Only the `reference_json` path is affected — `Normalizer(reference_json=...)`, `EquivalenceChecker(reference_json=...)`, `VariantProjector(reference_json=...)`, `BatchProcessor(reference_json=...)` and `CoordinateMapper(reference_json=...)`. `from_manifest(...)` already lands on the `Arc` variant and is unaffected. Note `JsonProvider` is not test-only: its own doc comment records that it backs the production `Normalizer(reference_json=...)` path.

The classes that build their inner object once at construction (`VariantProjector`, `EquivalenceChecker`, `BatchProcessor`) pay the clone once rather than per call; the per-call cost is `Normalizer`'s.

## Suggested resolution

Change the variant to `Mock(Arc<JsonProvider>)`. The clone becomes a refcount bump, `Arc` is the same 8 bytes so the `large_enum_variant` constraint still holds, and nothing mutates the provider after construction. Measured after the change, the table above is flat at 0.8 µs / 0.5 µs across all four contig sizes.

A regression test wants to assert the *shape* — that per-call cost does not scale with the size of the reference the call does not read — rather than pinning a microsecond figure.

## Relationship to #1455

Found while fixing #1455 (those same bindings holding the GIL across reference I/O). They are separate defects: #1455 is about *when* the GIL is released, this is about work that should not happen at all. Worth noting they interact, though — the clone sits outside the detached region, so before this is fixed a share of every call is unavoidably GIL-held however well the rest detaches.

nh13:
Closes #1455.
Closes #1473.

Two defects in the same call path in `src/python.rs`, one commit each. Both are pure performance/concurrency changes: **no normalized output moves**, so there is nothing here for a downstream consumer keying on the normalized HGVS string to re-baseline. See *Representation stability* below.

## 1. The bindings held the GIL across reference I/O (#1455) — `28bee29f`

Eight entry points read reference sequence through a provider while holding the GIL: `Normalizer.to_spdi`, `.canonical_spdi`, `.apply_to_reference`, `.normalize`, `.normalize_variant`, `.normalize_with_warnings`, and `EquivalenceChecker.check` / `.all_equivalent`. For a FASTA- or mmap-backed provider those are file I/O, and `canonical_spdi` issues several as its window doubles through a repeat tract. Held, the GIL serialises every other Python thread in the process for that whole time.

Each now clones its owned inputs and runs the work inside `py.detach` — the shape `VariantProjector` already used on all eight of its own entry points.

Measured as **the share of its solo iteration rate a competing Python thread keeps while the call runs** (release wheel, 12-core host, Python 3.14):

| worker | before | after |
|---|---:|---:|
| the eight entry points | 0.13–0.30 | 1.21–1.49 |
| Rust call known to hold the GIL (negative control) | 0.23 | 0.23 |
| pure-Python worker | 0.36 | 0.36 |

So every entry point sat *at or below* a call known to hold the GIL, and now sits ~5× above it. The populations do not overlap in either direction.

## 2. Every call deep-copied the entire reference (#1473) — `0a5f2523`

Found while measuring the above. Each of those methods builds a fresh normalizer per call — `Normalizer::with_config(self.provider.clone(), ...)` — so whatever the `PyProvider` handle holds is copied once per Python call. The `MultiFasta` variant was already an `Arc` for exactly that reason and its doc comment said so; the `JsonProvider` variant was a `Box`, chosen to satisfy `clippy::large_enum_variant`. That addressed the enum's *size* and not the clone's *cost*, and `JsonProvider` derives `Clone` with `String`-owning `genomic_sequences` and `proteins` maps.

Measured with the variant's own work held constant — a 6-base deletion on a 600-base transcript at the front of the contig — while only the **unread** remainder of the contig grows:

| contig bases | `normalize_variant` before | after | `to_spdi` before | after |
|---:|---:|---:|---:|---:|
| 10 000 | 1.2 µs | 0.8 µs | 1.0 µs | 0.5 µs |
| 100 000 | 6.8 µs | 0.8 µs | 6.1 µs | 0.5 µs |
| 1 000 000 | 21.8 µs | 0.8 µs | 17.5 µs | 0.5 µs |
| 4 000 000 | 443.2 µs | 0.8 µs | 74.0 µs | 0.5 µs |

~9 GB/s, i.e. memcpy speed, for bases no call reads. A single human chromosome in `genomic_sequences` would have cost tens of milliseconds and a full re-allocation of itself on *every* call. `Mock(Arc<JsonProvider>)` makes the clone a refcount bump; nothing mutates the provider after construction, and `Arc` is the same 8 bytes so the size constraint still holds.

Only the `reference_json` path was affected — `from_manifest(...)` already landed on the `Arc` variant.

## Unchanged on purpose

- **`VariantProjector` and `BatchProcessor`** — checked, not assumed: both already detach on every reference-backed entry point.
- **`CoordinateMapper`** — touches the provider for a single `get_transcript` metadata lookup and then does arithmetic on the returned record. No sequence is read, so there is no I/O to overlap.
- **The parse in `normalize` / `normalize_with_warnings`** stays outside the detached region. It touches no reference, so it is not what #1455 is about, and hoisting it in would require a two-variant error enum threaded through the closure to keep the `ParseError` mapping.
- **`Normalizer::with_config` itself** stays outside the detached region. After commit 2 it is an `Arc` bump plus a small config clone; moving it in would buy nothing measurable.

## About the #1455 test, which is a timing test

`tests/python/test_issue_1455_gil_release.py` measures GIL starvation of a competing Python thread. Reviewers should know how it got here, because the first three attempts were wrong in instructive ways:

1. **A wall-clock design (one thread vs two, same total calls) is not good enough.** Its signal is 2x at best and it flaked **4 runs in 10** of the full suite on a build already proven to detach. Replaced by an observer-starvation measurement, which separates by ~5x with far tighter spreads.
2. **`sys.setswitchinterval` is load-bearing.** At CPython's 5 ms default the switch interval dwarfs a ~100 µs call, a GIL-holding worker gets only a few percent of the time, and the measurement does not discriminate at all. It is set to 10 µs for the module and restored afterwards.
3. **Every threshold is measured on the host, never hard-coded.** The first version of the observer design compared against an absolute figure calibrated on a 12-core laptop — and on CI's 4-vCPU runner it **skipped all eight tests**, so the job was green and guarding nothing. That is not a hypothetical; it happened on this PR's first push. The fix is to calibrate against a *pure-Python worker*, which shares the GIL fairly and is therefore the exact midpoint the question turns on: a released call scores above fair sharing, a held one below. The spread between the two hosts is the whole argument:

   | | fair share | GIL-held | the eight entry points |
   |---|---:|---:|---:|
   | 12-core laptop | 0.36 | 0.23 | 1.21–1.49 |
   | CI, 4 vCPU | 0.693 | 0.427 | all 8 above the 1.04 bar |

   Nearly 2x apart in absolute terms, identical in the relation that matters.

4. **It skips rather than fails when it cannot discriminate**, via three guards: per-core load average; a Rust call known to hold the GIL failing to beat fair sharing (for the VM case where guest load average hides host contention); and a fair-share calibration above 0.8, which two GIL-bound Python threads cannot reach and which indicates the calibration itself was corrupted. That last one is not decoration — because `fair` is the pivot, an inflated one *raises* the bar, and a corrupted calibration produces eight false failures rather than a skip. Observed directly at load average 20 with the load guard disabled: fair measured 1.024 and all eight correct entry points failed.

The calibration numbers are emitted as a warning on every run, so a skip on a machine you cannot log into is still diagnosable from the CI log.

The #1473 test has no such caveat: it is single-threaded, compares identical work against two reference sizes, and the signal is ~30x against a 5x bound.

## Representation stability

**No normalized output changes.** Neither commit touches normalization, projection or SPDI logic — commit 1 moves existing calls inside `py.detach`, commit 2 changes a `Box` to an `Arc`. The full Rust suite passes unchanged with all three oracles and `FERRO_SWEEP_SEEDS=full`, which is where a moved canonical form would surface. Nothing for the downstream consumer to re-baseline.

## Verification

Both tests were confirmed to fail on the pre-fix build and pass after — the point of writing them first:

| | red (pre-fix) | green (post-fix) |
|---|---|---|
| `test_issue_1455_gil_release.py` | 8/8 fail, shares 0.129–0.303 vs a 0.250 control | 8/8 pass, shares 1.21–1.49 |
| `test_issue_1473_provider_clone.py` | 4/4 fail, ratios 19.4×–36.8× vs a 5× bound | 5/5 pass |

The #1473 red run was taken against commit `28bee29f` itself, which is exactly the pre-#1473 state.

```
  FERRO_SWEEP_SEEDS=full cargo nextest run --features dev            # 9410 passed, 145 skipped
cargo check --features python                                        # clean
ruff check python/ tests/python/                                     # clean
ruff format --check python/ tests/python/                            # clean
mypy python/ferro_hgvs/                                              # clean
pytest tests/python/                                                 # 710 passed
```

The Python suite was run 10 consecutive times on a quiet host to check the new timing test for flakiness: 10/10 clean. It was then run 10 more times at load average ~50 on 12 cores, where it skipped 10/10 and failed 0/10 — the intended degradation.

On CI the eight #1455 cases **run and pass** (`fair-share=0.693 gil-held=0.427`), and the four #1473 cases pass. All 35 checks are green: 34 pass plus `report-failure` reporting `skipping`, which is its state on a green run.

### DIFF
--- a/python/ferro_hgvs/__init__.pyi
+++ b/python/ferro_hgvs/__init__.pyi
@@ -617,7 +617,15 @@ class HgvsVariant:
     def __hash__(self) -> int: ...
 
 class Normalizer:
-    """HGVS variant normalizer using a reference provider."""
+    """HGVS variant normalizer using a reference provider.
+
+    Every method that reads the reference — :meth:`normalize`,
+    :meth:`normalize_variant`, :meth:`normalize_with_warnings`,
+    :meth:`to_spdi`, :meth:`canonical_spdi` and :meth:`apply_to_reference` —
+    releases the GIL for the duration of that work, so calls from several
+    ``threading.Thread`` workers overlap instead of serialising (#1455). A
+    ``Normalizer`` is safe to share between threads.
+    """
 
     def __init__(
         self,
@@ -937,7 +945,13 @@ class EquivalenceResult:
         ...
 
 class EquivalenceChecker:
-    """Equivalence checker for comparing HGVS variants."""
+    """Equivalence checker for comparing HGVS variants.
+
+    :meth:`check` and :meth:`all_equivalent` release the GIL while they read the
+    reference, so calls from several ``threading.Thread`` workers overlap
+    instead of serialising (#1455). An ``EquivalenceChecker`` is safe to share
+    between threads.
+    """
 
     def __init__(self, reference_json: str | None = None) -> None:
         """Create a new equivalence checker.
--- a/src/python.rs
+++ b/src/python.rs
@@ -1224,6 +1224,24 @@ pub struct PyNormalizer {
     kind: ProviderKind,
 }
 
+/// Every method below that reaches the provider runs its work inside
+/// `py.detach`, so the GIL is released across the reference read (#1455).
+///
+/// The reads are not incidental: for a FASTA- or mmap-backed provider they are
+/// file I/O, and `canonical_spdi` issues several as its window doubles through
+/// a repeat tract. Held, the GIL serialises every other Python thread in the
+/// process for that whole time. Measured as the share of its solo rate a
+/// competing Python thread keeps while the call runs, all six entry points sat
+/// at 0.13-0.30 before this change — at or below a Rust call known to hold the
+/// GIL — and 1.21-1.49 after (`tests/python/test_issue_1455_gil_release.py`).
+///
+/// The shape is the same one `VariantProjector` already used: clone the owned
+/// inputs — a `&PyHgvsVariant` borrows Python-owned memory and so cannot cross
+/// into the closure — detach, and convert the result or the mapped error
+/// afterwards, since `ferro_typed` builds a Python exception and needs the GIL
+/// back. **Anything added here that touches the provider wants the same
+/// treatment**; the test above is parametrized over the entry points, so a new
+/// one has to be added to it.
 #[pymethods]
 impl PyNormalizer {
     /// Create a new normalizer
@@ -1360,10 +1378,10 @@ impl PyNormalizer {
     ///
     /// Returns:
     ///     SpdiVariant
-    fn to_spdi(&self, variant: &PyHgvsVariant) -> PyResult<PySpdiVariant> {
+    fn to_spdi(&self, py: Python<'_>, variant: &PyHgvsVariant) -> PyResult<PySpdiVariant> {
         let normalizer = Normalizer::with_config(self.provider.clone(), self.config.clone());
-        normalizer
-            .to_spdi(&variant.inner)
+        let inner_variant = variant.inner.clone();
+        py.detach(|| normalizer.to_spdi(&inner_variant))
             .map(|inner| PySpdiVariant { inner })
             .map_err(|e| ferro_typed("ProjectionError", format!("{e}"), &e))
     }
@@ -1398,10 +1416,10 @@ impl PyNormalizer {
     ///         sequence — a trans/mosaic/chimeric or null allele, members on
     ///         different accessions, members that overlap, an edit SPDI cannot
     ///         represent, or a span wider than 100 000 bases.
-    fn canonical_spdi(&self, variant: &PyHgvsVariant) -> PyResult<PySpdiVariant> {
+    fn canonical_spdi(&self, py: Python<'_>, variant: &PyHgvsVariant) -> PyResult<PySpdiVariant> {
         let normalizer = Normalizer::with_config(self.provider.clone(), self.config.clone());
-        normalizer
-            .canonical_spdi(&variant.inner)
+        let inner_variant = variant.inner.clone();
+        py.detach(|| normalizer.canonical_spdi(&inner_variant))
             .map(|inner| PySpdiVariant { inner })
             .map_err(|e| ferro_typed("ProjectionError", format!("{e}"), &e))
     }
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

PR #1476, opened 2026-08-06
test(sweep): exercise the non-coding `r.` axis in the transcript cis sweep

### LINKED ISSUE #1471: Exhaustive cis sweeps never exercise the `r.` axis, which is why #1453 sat undetected
## Summary

The three exhaustive cis sweeps — `cis_junction_crossing_shift`, `repeat_span_sibling_overlap`, `issue_1254_sibling_crossing_shift` — exist to prove that no two-member cis allele normalizes to a description denoting a different sequence. **None of them emits a single `r.` descriptor.** `grep -c ':r\.'` over the three files returns `0`, `0`, `0`.

That is the structural reason #1453 survived a fully green suite. Over a 3,200-row corpus of nearby junction-occupying members on a non-coding transcript, `main` produced **849 outputs that repeated a member** — descriptions that denote no sequence at all — and no test saw any of them.

## The exclusion is deliberate, and its stated reason does not hold

`tests/it/cis_junction_crossing_shift.rs:934`:

```rust
// `c.` and `n.` share the generator; `r.` is excluded because its
// `U`/`T` alphabet makes the payload construction below a different
// problem, and `issue_1235_transcript_axes.rs` pins that axis by hand.
for (accession, axis) in [("NM_TEST.1", "c"), ("NR_TEST.1", "n")] {
```

Both halves of that are measurably wrong today.

**The alphabet is not a problem.** The parser accepts the generator's DNA-alphabet payloads on `r.` verbatim and renders them in the RNA alphabet on output, so the payload construction needs no change at all:

```
NR_TEST.1:r.5_6insAA  parses as  NR_TEST.1:r.5_6insaa
NR_TEST.1:r.5T>A      parses as  NR_TEST.1:r.5u>a
```

**The compensating control does not cover the same shapes.** `issue_1235_transcript_axes.rs` does have 13 `r.` cases, including a non-coding one — but they are substitutions and `delins`. Those reach `canonicalize_from_sequence`; they never produce two junction-occupying members shifting onto one junction, which is where #1453 lived (`coalesce_members_at_one_junction` → `respell_at_gap`).

## The apply oracle already handles `r.` — it is simply never called on it

Worth stating explicitly, because #1453's own suggested scope proposed *extending* the oracle. No extension is needed. Measured against `tests/it/common/cis_apply_oracle.rs`:

```
NR_TEST.1:r.[9dup;11dup;11dup]   apply=None, reason=CoincidentInsertions   <- the #1453 output, caught
NR_TEST.1:r.[9dup;10dup;11dup]   apply=Some(..)                            <- the input
NR_TEST.1:r.9_10insuaa           apply=Some(..)                            <- the fixed output
```

and the `n.` and `r.` spellings of one variant apply to byte-identical sequences. `ApplyFailure::CoincidentInsertions` is precisely the right verdict for two members on one junction.

For contrast, the *other* candidate mechanism does not work: a prototype seam assertion built on `overlap::detect_overlap_conflicts` reported **nothing** on the #1453 output. `overlap::na_range` calls `writes_only_at_a_junction`, which is true for `Duplication`, so `dup` members are excluded from coincident-bounds detection by design (the #1411 read/write distinction). Two same-junction `dup`s are a coincident-*insertion* conflict, not a base-claim one, and only the apply oracle models that. So the fix is to invoke the existing oracle on `r.`, not to add a detector.

## Proposed fix, validated end-to-end

One tuple:

```rust
for (accession, axis) in [("NM_TEST.1", "c"), ("NR_TEST.1", "n"), ("NR_TEST.1", "r")] {
```

The sweep body is already axis-generic (`format!("{accession}:{axis}.[{first};{second}]")`), and the provider selector (`if axis == "c" { cds } else { noncoding }`) already does the right thing for `"r"`.

Measured with the #1453 defect restored (`src/normalize/merge.rs` at `ee6a19cc`), at only **4 seeds**:

```
FAILED: overlapping or unconvertible output in 103680 transcript-axis cases: [
    "NR_TEST.1:r.[2dup;4dup] [ThreePrime] -> NR_TEST.1:r.[9dup;9dup]",
    "NR_TEST.1:r.[2dup;5dup] [ThreePrime] -> NR_TEST.1:r.[9dup;9dup]",
    ... 8 more
]
```

With PR #1465's fix in place the same sweep passes. So it catches the defect on the earliest shapes the generator emits, not in some deep corner.

## Cost, which is the real objection

`checked` goes **69,120 → 103,680 at 4 seeds — +50%** — on what is already the suite's slowest test (306s for the transcript-axis variant at `FERRO_SWEEP_SEEDS=full` locally). Wall-clock measurements across runs were too cache-noisy to quote honestly; the case count is the reliable figure.

Mitigations worth considering rather than assuming:

- Add only the **non-coding** `r.` record, not a coding one. #1453's defect is specific to a record with no CDS, and coding `r.` shares `reading_frame` with `c.` which is already swept. That is the +50% above; adding both would be +100%.
- The default run is a 4-seed prefix, so the cost lands mostly on CI's `Exhaustive sweeps` job, which is not one of the eight checks `main`'s ruleset requires.

## Notes

- No pinned count breaks: the sweep's only count assertion is a floor (`checked > 1000 * seeds`), which more cases satisfy.
- Found while working #1453 / PR #1465.

nh13:
Closes #1471.

> **Stacked on #1465.** The base of this PR is `fix/issue-1453-rna-repeated-member`, not `main`, because the new sweep arm is **red without that fix** — that is the point of it. Merge #1465 first, then this retargets to `main` cleanly.

## The defect this closes

`no_two_member_transcript_axis_allele_normalizes_to_a_different_sequence` swept `c.` and `n.` and excluded `r.`, on two stated grounds. Both were measured and neither holds.

**1. "its `U`/`T` alphabet makes the payload construction below a different problem."** It does not. The parser accepts the DNA-alphabet payloads this generator already builds, and renders them in the RNA alphabet on output — so not one line of the generator changes:

```
NR_TEST.1:r.5_6insAA  parses as  NR_TEST.1:r.5_6insaa
NR_TEST.1:r.5T>A      parses as  NR_TEST.1:r.5u>a
```

The oracle needs no adjustment either: an `r.` description applies to bases byte-identical to its `n.` twin, because `hgvs_to_spdi` folds the alphabet.

**2. "`issue_1235_transcript_axes.rs` pins that axis by hand."** It does — 13 `r.` cases, including a non-coding one. But they are substitutions and `delins`, which reach `canonicalize_from_sequence` and never the *repair* passes. None of them produces two junction-occupying members shifting onto one junction.

That second gap is what #1453 fell through: on a non-coding `r.` record every repair routed through `respell_at_gap` was a silent no-op, so `r.[9dup;10dup;11dup]` normalized to `r.[9dup;11dup;11dup]` — one interbase point claimed twice, denoting no sequence. It was invisible to two of the three seam oracles, since the output is well-formed (`FERRO_ASSERT_REPARSE` accepts it) and every coordinate is in range (`FERRO_ASSERT_IN_BOUNDS` accepts it). Only the apply oracle distinguishes it, as `CoincidentInsertions` — and nothing was calling it on `r.`.

## The change

One tuple, plus the comment that justified the exclusion:

```diff
-for (accession, axis) in [("NM_TEST.1", "c"), ("NR_TEST.1", "n")] {
+for (accession, axis) in [("NM_TEST.1", "c"), ("NR_TEST.1", "n"), ("NR_TEST.1", "r")] {
```

The sweep body was already axis-generic (`format!("{accession}:{axis}.[{first};{second}]")`) and the provider selector (`if axis == "c" { cds } else { noncoding }`) already does the right thing for `"r"`.

## Verification — the red/green pair is the evidence

Adding cases proves nothing on its own; what matters is that the arm *fails on the defect*.

| configuration | result |
|---|---|
| 4 seeds, #1465's `merge.rs` reverted to `ee6a19cc` | **FAIL** |
| 4 seeds, with #1465's fix | PASS |
| **full corpus + all three oracles**, with #1465's fix | **PASS** (9,409/9,409 in the full gate) |

The failure, at only four seeds, on the earliest shapes the generator emits:

```
overlapping or unconvertible output in 103680 transcript-axis cases: [
    "NR_TEST.1:r.[2dup;4dup] [ThreePrime] -> NR_TEST.1:r.[9dup;9dup]",
    "NR_TEST.1:r.[2dup;5dup] [ThreePrime] -> NR_TEST.1:r.[9dup;9dup]",
    ... 8 more, the collector's cap
]
```

## Cost

**829,440 → 1,244,160 cases, +50%**, on what is already the suite's slowest test.

**No wall-clock figure is quoted, deliberately.** The machine this was developed on was running three concurrent copies of this sweep at load average 81; the full-corpus run took 1,117s there against a ~460s expectation, so any timing taken on it is contention noise rather than a measurement. The doc comment says this rather than recording a number nobody can reproduce, and points at the `sweeps` job's own duration as the figure of record.

Two things that limit the blast radius of that cost:

- The default local run is a 4-seed prefix, so it lands mostly on CI's `Exhaustive sweeps` job — which is not among the eight checks `main`'s ruleset requires.
- No new test is added, only cases inside an existing one, so `ci.yml`'s `SWEEP_FILTER` partition invariant (`8878 + 30 == 8908`) is untouched.

## Unchanged on purpose

- **No pinned count breaks.** The sweep's only count assertion is a floor (`checked > 1000 * seeds`), which more cases satisfy. `FIVE_PRIME_DUP_DEL_SEQUENCE_CHANGES` belongs to the genomic sweep, not this one.
- **Coding `r.` is a deliberate remaining gap, and is documented as one rather than quietly skipped.** It would be convenient to claim `c.` covers it since `axis_frame` gives both `reading_frame: true`, but that is only true of the frame — the two still take different paths (`build_rna_merged` vs `build_cds_merged`, `cds_relative_gap(Region::Rna)` vs `(Region::Cds)`), so a coding `r.` defect could hide the way the non-coding one did. It is deferred because it buys an *unmeasured* gap for another +50%, where non-coding `r.` buys a measured one.
- **No production code changes.** Tests only.

## Verification commands

```
cargo clippy --all-features --all-targets -- -D warnings      # clean, exit 0
# 9409 tests run: 9409 passed (3 slow), 145 skipped
```

### DIFF
--- a/tests/it/cis_junction_crossing_shift.rs
+++ b/tests/it/cis_junction_crossing_shift.rs
@@ -902,7 +902,15 @@ fn no_tandem_tract_allele_normalizes_to_a_different_sequence() {
 /// It runs in the `sweeps` job only — `cis_junction_crossing_shift` is inside
 /// `SWEEP_FILTER`, which `test` and `test-oracle` negate — so it is exercised at
 /// the full corpus with the oracles set, which is the configuration that found
-/// all five. Measured at 280s for 829,440 cases.
+/// all five. Measured at 280s for 829,440 cases over two axes (`c.`, `n.`).
+///
+/// #1471 added a third (non-coding `r.`), taking it to **1,244,160 cases**. That
+/// count is exact and deterministic — three axes over the same corpus, shapes and
+/// positions — so it is quoted where a wall-clock is not: the machine this was
+/// developed on was running three concurrent copies of this sweep at a load
+/// average of 81, which makes any timing taken there contention noise rather than
+/// a measurement. Expect roughly 1.5x the 280s above on a quiet machine, and
+/// treat the `sweeps` job's own duration as the figure of record.
 ///
 /// Run it locally in the configuration `sweeps` uses — the oracles are not
 /// optional here, since four of the five defects above were *reported* by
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1479, opened 2026-08-06
fix(reference): a zero-length sequence read is empty, not out of bounds

### LINKED ISSUE #1472: `get_sequence` rejects a zero-length read on a transcript but allows it on a contig, so `apply_to_reference` declines every insertion and dup on the transcript axes
## Summary

`ReferenceProvider::get_sequence` disagrees with itself about whether a zero-length interval is a legal read:

```
genomic     get_sequence("NC_TEST.1", 9, 9)  ->  Ok("")
transcript  get_sequence("NR_TEST.1", 9, 9)  ->  Err("Invalid coordinates: Position 9-9 out of range for NR_TEST.1")
```

A zero-length window is exactly what an **insertion** or a **duplication** has for its reference span, so the transcript path declines both — while accepting substitutions and deletions on the same record:

```
NC_TEST.1:g.9_10insTAA  ->  Ok        (genomic)
NC_TEST.1:g.9dup        ->  Ok        (genomic)

NR_TEST.1:n.9_10insTAA  ->  Err("cannot apply to reference — its reference window could not be read")
NR_TEST.1:n.9dup        ->  Err("cannot apply to reference — its reference window could not be read")
NR_TEST.1:n.12T>A       ->  Ok
NR_TEST.1:n.9_10del     ->  Ok
```

Reproduced on both a `SyntheticBuilder` transcript (which has a genomic contig and mapping) and a bare `MockProvider` transcript, so it is not an artefact of one fixture builder.

## Why this matters beyond the error message

`apply_to_reference` is documented as the encoding-**invariant** ground truth:

> The ground truth for equivalence: two descriptions denote the same edit exactly when they produce the same resulting bases.

It currently is not, on the transcript axes. Two spellings of one variant get opposite answers:

```
NR_TEST.1:n.[9dup;10dup;11dup]   canonical_spdi = Ok("NR_TEST.1:9::TAA")
NR_TEST.1:n.9_10insTAA           canonical_spdi = Err("... reference window could not be read")
```

Those denote the same sequence — the second is what ferro's own normalizer returns for the first. The allele resolves because its combined window spans real bases; the normalized single-member form has a zero-length window and does not. So the primitive accepts a variant and then rejects ferro's own canonical form of it.

`canonical_spdi` inherits this, since it is built on `apply_to_reference`.

## Consequence: no seam-level apply oracle can be built today

This was found while scoping #1453's suggestion that an applier is "the only check that would have caught" a repeated-member output. A prototype 4th seam oracle in `Normalizer::assert_seam_oracles`, comparing `canonical_spdi(input)` against `canonical_spdi(normalized)`, is unusable as a result: it reports `undenotable` for **every** normalized output that happens to be an insertion or a dup on `c.`/`n.`/`r.`, which is a large fraction of normalizer output. The signal is entirely swamped by the false positives.

That is a real blocker rather than a stylistic one — a sequence-preservation invariant at the seam is the strongest available defence against the #1286/#1301/#1307/#1316/#1323/#1453 family of overlapping-member defects, and it cannot be written until this is fixed.

(The *test-side* `tests/it/common/cis_apply_oracle.rs` is unaffected — it does its own walk and handles these shapes correctly. #1471 covers wiring that one to the `r.` axis.)

## Suggested resolution

Make the transcript path agree with the genomic one and return `Ok("")` for `start == end`, rather than treating it as out of range. The genomic behaviour is the correct one: a half-open `[9, 9)` interval is empty, not invalid, and every caller that asks for one is asking about an interbase point.

Worth checking as part of it:

- whether any caller currently *relies* on the transcript path erroring for `start == end` — a guard that silently depends on the refusal would change behaviour;
- whether the bounds check should still reject `start == end` when the position itself is genuinely out of range (e.g. `get_sequence(key, 999, 999)` on a 63-base transcript), which is a different question from zero length and should presumably still be an error;
- whether `MultiFastaProvider` and the cdot-backed providers agree with whichever answer is chosen, not just `MockProvider`/`JsonProvider`.

## Notes

- Found while working #1453 / PR #1465; not introduced by it.
- Related: #1471 (the sweeps never exercise the `r.` axis).

nh13:
Closes #1472.

## The defect

`ReferenceProvider::get_sequence` disagreed with itself about whether a zero-length interval is a legal read:

```
genomic     get_sequence("NC_TEST.1", 9, 9)  ->  Ok("")
transcript  get_sequence("NR_TEST.1", 9, 9)  ->  Err("Position 9-9 out of range")
```

A zero-length window is exactly what an **insertion** or a **duplication** has for its reference span, so the transcript path declined both — while accepting substitutions and deletions on the same record.

The consequence is at the `apply_to_reference` layer, which is documented as the encoding-**invariant** ground truth ("two descriptions denote the same edit exactly when they produce the same resulting bases"). On the transcript axes it was not: two spellings of one variant got opposite answers, and the one it rejected is the one ferro's own normalizer returns for the other.

```
NR_TEST.1:n.[9dup;10dup;11dup]   canonical_spdi = Ok("NR_TEST.1:9::TAA")
NR_TEST.1:n.9_10insTAA           canonical_spdi = Err("... reference window could not be read")
```

## Four paths, not two — and the split is not contig-vs-transcript

The issue framed this as the contig path disagreeing with the transcript path. Surveying every `get_sequence` implementation, it is **2–2, and `FastaProvider` disagrees with itself** depending on whether a read is served from its index or from an mmap:

| path | zero-length read at an in-range start |
|---|---|
| `MockProvider` genomic (rejects on `start > end`) | `Ok("")` |
| `FastaProvider` indexed (`if start >= actual_end { return Ok(String::new()) }`) | `Ok("")` |
| `Transcript::get_sequence` (requires `start < end`) | **rejected** |
| `FastaProvider::get_sequence_from_mmap` (rejects on `start >= end`) | **rejected** |

So an insertion could resolve or not depending on whether the record happened to be mmap-backed. This fixes the two that reject and adopts the convention the other two already implement, rather than inventing a third.

## The change

```diff
-if end <= seq.len() && start < end {
+if start < seq.len() && end <= seq.len() && start <= end {
```

```diff
-if start >= entry.length || end > entry.length || start >= end {
+if start >= entry.length || end > entry.length || start > end {
```

The mmap read loop is already safe for a zero length — `while bytes_read < 0` never runs — so only the guard changes.

## Unchanged on purpose

**Zero length and out of range stay separate questions.** Only the first is relaxed:

- `start` must still name a base the sequence has, so a zero-length read at or past `len` is still an error — **including at exactly `len`**, the junction past the last base. That matches the contig path's `start >= len` guard, which is the point: aligning the four paths is the goal, and unilaterally changing the contig path would be the larger blast radius. It is arguable the other way (an empty interval at the end is well-defined), so it is pinned explicitly in `an_out_of_range_zero_length_read_is_still_an_error` — reversing it later is a one-line decision, not archaeology.
- **An inverted range is still rejected.** It shared the old `start < end` test, so relaxing that test had to keep rejecting it explicitly; `test_fasta_provider_get_sequence_rejects_inverted_range` (pre-existing) still passes.

**No caller relied on the refusal.** No `get_sequence(..).is_err()` is used as a signal anywhere in `src/`, and `conformance::reference_window` mirrors the same dispatch so it inherits the fix rather than diverging from it.

## Tests — written first, and the guards discriminate

Three of the six integration tests failed before the fix and pass after; the other three passed **both** before and after, which is what makes them guards rather than restatements. A fix that simply dropped the bounds check would satisfy the reproducer and break them.

| test | before | after |
|---|---|---|
| contig and transcript agree on a zero-length read | FAIL | PASS |
| insertion's `canonical_spdi` == its allele spelling's | FAIL | PASS |
| a transcript-axis `dup` applies | FAIL | PASS |
| out-of-range zero-length still errors | PASS | PASS |
| inverted range still errors | PASS | PASS |
| ordinary reads unchanged (bases, not just `Ok`) | PASS | PASS |

Two unit tests sit next to `Transcript::get_sequence` itself, since the contract belongs to that function — the provider merely maps `None` onto `InvalidCoordinates`, so an integration-only assertion would be measuring the dispatch as much as the bounds test.

The insertion test asserts **agreement between the two spellings** rather than that the insertion is merely `Ok`, which would pass even if it resolved to the wrong bases.

## Representation stability

**0 of 33,792 rows moved** on `dump_normalized_corpus`, across all nine shape families, 0 previously accepted and 0 previously not a fixed point.

That zero is load-bearing rather than vacuous, which is worth stating because a zero from this harness often is not: the corpus's `c.` rows resolve through `MockProvider::get_sequence` → `get_transcript` → `Transcript::get_sequence`, so the changed path is genuinely exercised.

The change is a **widening**: descriptions that previously failed to apply now do, and nothing that applied before changed what it denotes. There is no stored string to migrate.

## Verification

```
cargo clippy --all-features --all-targets -- -D warnings      # exit 0
# 9440 tests run: 9440 passed (4 slow), 145 skipped
```

Independent of #1465 and #1476 — branched off current `main`, touches no normalizer code.

### DIFF
--- a/src/reference/fasta.rs
+++ b/src/reference/fasta.rs
@@ -831,7 +831,14 @@ impl MmapFastaProvider {
         start: u64,
         end: u64,
     ) -> Result<String, FerroError> {
-        if start >= entry.length || end > entry.length || start >= end {
+        // `start > end` rather than `start >= end`: a zero-length range is empty,
+        // not invalid, and reads no bases (#1472). The loop below is already
+        // safe for it — `bytes_read < 0` never runs — and this is what the
+        // indexed read a few hundred lines up already does, via its
+        // `if start >= actual_end { return Ok(String::new()) }`. The two served
+        // the same provider and disagreed, so an insertion resolved or not
+        // depending on whether the record happened to be mmap-backed.
+        if start >= entry.length || end > entry.length || start > end {
             return Err(FerroError::InvalidCoordinates {
                 msg: format!(
                     "Position {}-{} out of range for {} (length {})",
--- a/src/reference/transcript.rs
+++ b/src/reference/transcript.rs
@@ -494,13 +494,38 @@ impl Transcript {
         }
     }
 
-    /// Get sequence at a position range (0-based). Returns `None` if the
-    /// transcript has no cached sequence bases or the range is out of bounds.
+    /// Get sequence at a position range (0-based, half-open). Returns `None` if
+    /// the transcript has no cached sequence bases or the range is out of
+    /// bounds.
+    ///
+    /// # A zero-length range is empty, not invalid (#1472)
+    ///
+    /// `start == end` names the interbase point before `start` and reads no
+    /// bases, so it yields `""`. This used to require `start < end` and so
+    /// returned `None`, which the provider reported as `Position 9-9 out of
+    /// range` — and that is exactly the window an **insertion** or a
+    /// **duplication** has. `apply_to_reference` therefore declined every
+    /// insertion and `dup` on the transcript axes while accepting substitutions
+    /// and deletions on the same record, which made a primitive documented as
+    /// encoding-*invariant* answer differently for two spellings of one variant:
+    /// `n.[9dup;10dup;11dup]` resolved and the `n.9_10insTAA` that ferro's own
+    /// normalizer returns for it did not.
+    ///
+    /// The contig path (`MockProvider::get_genomic_sequence`, which rejects on
+    /// `start > end`) and `FastaProvider`'s indexed read already answered `""`
+    /// here, so this adopts their convention rather than inventing a third.
+    ///
+    /// **Zero length and out of range stay separate questions.** `start` must
+    /// still name a base the sequence has, so a zero-length read at or past
+    /// `len` is `None` — matching the contig path's `start >= len` guard,
+    /// including at exactly `len` (the junction past the last base). An inverted
+    /// range is likewise still `None`; it shared the old `start < end` test, so
+    /// relaxing that test had to keep rejecting it explicitly.
     pub fn get_sequence(&self, start: u64, end: u64) -> Option<&str> {
         let seq = self.sequence.as_deref()?;
         let start = start as usize;
         let end = end as usize;
-        if end <= seq.len() && start < end {
+        if start < seq.len() && end <= seq.len() && start <= end {
             Some(&seq[start..end])
         } else {
             None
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

Issue #1478, opened 2026-08-06
Representation corpus pins transcript geometry to one exon and CDS_START=1, so #1450 and every 5'UTR shape are unreachable


nh13:
### Summary

The representation corpus (`examples/dump_normalized_corpus.rs`) pins its transcript geometry to a single value on every axis that matters for junction behaviour. Verified on `2ddc25b9`:

```rust
const CDS_START: u64 = 1;
const CDS_END: u64 = 15;
let exon = Exon::with_genomic(1, 1, tx_len, g_start, g_end);   // one exon
Strand::Plus,
```

Two consequences, both structural rather than incidental:

1. **#1450 is unreachable.** It is an open cis-member exon-junction defect — `general.md:44` exempts deletions and duplications around exon/exon junctions from the 3' rule, the per-member pipeline honours it and the sequence-first cis path does not. Its reproducer needs a three-exon transcript with `cds_start = 5`. This corpus has one exon and `cds_start = 1`, so it has no exon/exon junction at all and cannot express the shape.
2. **`c.-N` positions are never generated**, because `CDS_START = 1` leaves no 5'UTR. Measured over the full corpus: **0** rows name a `c.-N` position, on either side of a dump.

Worth noting the module doc deliberately engineered the *other* junction — "deliberately short of the 20-base core so that **five** 3'UTR bases (`c.*1`..`c.*5`) remain… #1398, #1418 and both halves of #1426 lived in that region." The 3'UTR was reasoned about carefully; the 5'UTR was left structurally unreachable.

Minus-strand transcripts are absent for the same reason, though no open issue is currently blocked on it.

### Why this is worth fixing rather than absorbing

This is the third instance of one failure mode, which is the actual argument. The corpus is the evidence base for release-time representation claims, so a blind spot in it produces a **zero that reads as a clean result**:

- #1456 — built no overlap-conflicting alleles, so it reported `0 of 18,432` for #1403, #1445 and #1451 in a row. Fixed by #1459.
- #1463 — emitted no `delins` that was secretly an inversion, so it reported zero non-idempotent rows while #1454 was open and reproducing on the same `main`. Fixed by #1468.
- this one — pins transcript geometry, so #1450 cannot appear.

Each was found by accident, in a separate investigation, and patched individually. That pattern is the reason to treat geometry as a dimension rather than wait for a fourth.

### Suggested scope

[...]

Issue #1477, opened 2026-08-06
Sequence-first canonicalization never runs for a lone transcript-axis member, and the deferral's recorded cost needs refreshing


nh13:
A lone `c.`/`n.`/`r.` member never enters the #1229–#1235 sequence-first re-derivation, while the **allele** spelling of the same variant does. The result is non-confluence: one variant, two normal forms, on the same axis.

`is_splittable_single_member` (`src/normalize/mod.rs:1923`, at `origin/main` `5a19d1d2`) matches only `HV::Genome` and `HV::Mt`:

```rust
let edit = match variant {
    HV::Genome(g) => &g.loc_edit.edit,
    HV::Mt(m) => &m.loc_edit.edit,
    _ => return false,        // Cds / Tx / Rna
};
```

The gate is deliberate, and its doc comment describes this asymmetry in as many words — *"a lone spelling that the allele spelling of the same variant would re-partition can stay unsplit"* — deferring the widening pending a measurement pass. **This issue is that measurement**, filed because no tracking issue for it turns up by search, so the deferral currently exists only as a code comment.

## Reproducer

A 20-base near-palindrome (inverting it changes exactly four bases), whole transcript coding, no UTR:

```
NM_TEST.1:c.1_20inv                       -> NM_TEST.1:c.1_20inv          (unchanged)
NM_TEST.1:c.[3G>C;10T>A;11T>A;18G>C]      -> NM_TEST.1:c.[3G>C;10_11inv;18G>C]
```

The same variant on the genomic axis — same sequence, same relative position:

```
NC_TEST.1:g.257_276inv                    -> NC_TEST.1:g.[259G>C;266_267inv;274G>C]
NC_TEST.1:g.[259G>C;266T>A;267T>A;274G>C] -> same
```

Instrumenting `merge::canonicalize_from_sequence` shows the mechanism directly: the `g.` lone `inv` reaches it (`n=1`), the `c.` lone `inv` never calls it at all, and the three-member `c.` allele does (`n=3`).

Controls, so the cause is not misattributed. It is **not** block length (reproduces at 20 bases as well as at 1,100), **not** the CDS/3'UTR straddle (reproduces with the whole core coding, no UTR present), and **not** the sequence boundary (a bare unpadded `g.` contig still converges; an interior `c.` span still does not).

## The recorded cost needs refreshing

[...]

PR #1480, opened 2026-08-06
fix(normalize): keep the sequence-first derivation inside one exon

nh13:
Refs #1450 — **deliberately does not close it.** This PR takes the conservative half of that issue's Scope: it stops the derivation crossing an exon junction, so the reported outputs are correct. It does *not* make the derivation exon-*aware*, which is what the Scope asks for ("segment the block at exon edges, or re-place pieces into their originating exons"). See [Remaining half](#remaining-half) at the bottom.

## The defect

`general.md:44` exempts deletions and duplications around an exon/exon junction from the 3' rule on the `c.`/`n.`/`r.` axes. The per-member pipeline honours it; the sequence-first cis path did not. The same member was therefore clamped when described alone and carried past two junctions when described beside a sibling — whether a member is clamped depended on whether it had a sibling.

On `NM_001234.1` from `MockProvider::with_test_data` (exons at transcript 1-15 / 16-30 / 31-44, `cds_start = 5`, G-run spanning c.9-c.33), at `origin/main` `2ddc25b9`:

| input | before | after |
|---|---|---|
| `c.10del` | `c.11del` | `c.11del` (unchanged) |
| `c.13del` | `c.26del` | `c.26del` (unchanged) |
| `c.[10del;13del]` | **`c.32_33del`** | **`c.[11del;26del]`** |
| `c.[9del;13del]` | **`c.32_33del`** | **`c.[11del;26del]`** |
| `r.[10del;13del]` | **`r.32_33del`** | **`r.[11del;26del]`** |

## The fix, and why it sits where it does

Instrumenting the pass is what located it. The members arrive at the derivation **already correctly clamped**:

```
DERIVED from [c.11del; c.26del] -> [c.32_33del]
```

`[c.11del;c.26del]` is precisely the expected answer. The per-member pipeline had already produced it, and the derivation re-merged it across both junctions.

So the guard refuses when the members' own **input span** crosses a junction, before the window is built. It cannot go anywhere later:

- After `trim_common_flanks`, the two deletions have collapsed into a single G-run and **which exon each came from is no longer recoverable** from the resulting sequence.
- The merged piece then sits **wholly inside exon 3**, so no check on the block, the partition or the shift can see that a junction was crossed.
- `shift_pieces` in particular moves nothing for this shape (measured on #1450), so a clamp there is unreachable — which is why the first attempt at this issue changed nothing.

**Declining rather than deriving per exon.** The per-member result that survives a refusal *is* the expected answer, so this shape loses nothing by refusing, and refusal is this function's established convention. Segmenting the block at exon edges — #1450's other option — would additionally buy confluence for genuinely cross-junction alleles, but carries the `ref`/`alt` index-alignment risk that issue describes.

## Unchanged on purpose

- **The genomic axes.** `crosses_exon_junction` returns `false` for `Genome`/`Mt` before touching the provider, pinned by `the_exon_junction_guard_never_fires_on_a_genomic_axis`.
- **Same-exon spans**, including one that is exactly an exon edge to edge (`c.12..c.26`), pinned by `the_exon_junction_guard_fires_only_across_a_junction`.
- **A transcript the provider cannot serve** reports "not crossing", so the derivation is no worse off than before this check existed — the same fallback `axis_frame` takes.

Those two are unit tests next to the code rather than integration assertions, deliberately: the derivation returns `Some` only when its result *differs* from what the per-member pipeline already produced, so a same-exon pair declines for an unrelated reason and an integration-level "still derived" test would pass vacuously. That is stated in the test's own doc comment.

## Representation stability

**0 of 33,808 corpus rows move — and that zero is structural, not evidence.** `dump_normalized_corpus`'s coding provider builds a **single-exon** transcript (`Exon::with_genomic(1, 1, tx_len, …)`), so no corpus row can cross an exon junction and this guard can never fire on one. This is a third instance of the blindness class #1456 (shape) and #1460 (scale) are about, now on transcript structure — already tracked as **#1478**, which names #1450 specifically as one of the shapes it cannot reach.

The real evidence is the suite: **9,440 tests pass with no test re-blessed**, which for a change that moves normalized output is the strong signal — every existing expectation, including both hermetic oracles (`mutalyzer_normalize_tests::gate_normalized_snapshot`, `biocommons_normalize_tests::axis_normalized_hermetic`), already agreed with the clamped answer. The outputs that move are the ones no oracle had covered.

Downstream impact is confined to cis alleles whose members sit in different exons on a transcript axis, and for those the new answer is the one each member already had alone.

## Verification

```
# Summary [431.092s] 9440 tests run: 9440 passed (3 slow), 145 skipped
```

Test-first: the four integration tests were written against #1450's stated expectation and failed on `main` with `left: "NM_001234.1:c.32_33del"` / `right: "NM_001234.1:c.[11del;26del]"`.

## Follow-on

This unblocks #1477 (widening `is_splittable_single_member` to lone `c.`/`n.`/`r.` members), which was blocked precisely because the derivation lacked this clamp — six `issue_334_exon_junction_exception` tests plus the biocommons oracle. That widening is still its own measured pass and is not attempted here.

## Remaining half

#1450 stays open for the part this does not do: a cis allele whose members sit in **different
exons** is now declined by the derivation and keeps its per-member spelling, rather than being
re-derived from the resulting sequence within each exon. Concretely, `c.[10del;13del]` is correct
but is no longer a *derived* form — it is the per-member form, which happens to be the right one.

What is still owed to #1450:

- **Confluence across a junction.** Two spellings of one cross-junction variant can still settle on
  different forms, because neither is re-derived. This PR does not make that worse — before it,
  both were re-derived to the same *wrong* answer — but it does not fix it either.
- The two implementations its Scope names (segment at exon edges, or re-place pieces into their
  originating exons), and the `ref`/`alt` index-alignment risk that comes with the first.
- Measuring any of it needs #1478 first: the representation corpus builds a single-exon
  transcript, so it cannot produce a cross-junction row at all.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2203,6 +2203,31 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
 
     let frame = axis_frame(kind, &template_accession, provider)?;
     let accession = template_accession.transcript_accession();
+
+    // `general.md:44` exempts deletions and duplications around an exon/exon
+    // junction from the 3' rule on the transcript axes, and the per-member
+    // pipeline honours it — so the members reaching this function are *already*
+    // clamped. Re-deriving across a junction throws that away: `c.[10del;13del]`
+    // on a transcript whose G-run spans c.9-c.33 arrives here as the correctly
+    // clamped `[c.11del;c.26del]` and comes back as `c.32_33del`, both members
+    // carried past their junctions and merged into one (#1450).
+    //
+    // The constraint belongs on the **input span**, here, and nowhere later. By
+    // the time `trim_common_flanks` has run, the two deletions have collapsed
+    // into a single run and which exon each came from is no longer recoverable
+    // from the resulting sequence; the merged piece then sits wholly inside one
+    // exon, so no check on the block, the partition or the shift can see that a
+    // junction was crossed. (`shift_pieces` in particular moves nothing for this
+    // shape, so a clamp there is unreachable — measured on #1450.)
+    //
+    // Declining rather than deriving per exon: the per-member result that
+    // survives a refusal *is* the expected answer, so refusing costs this shape
+    // nothing. Segmenting the block at exon edges would additionally buy
+    // confluence for cross-junction alleles; that is materially larger and is
+    // left open.
+    if crosses_exon_junction(kind, &accession, provider, &frame, c_lo, c_hi) {
+        return None;
+    }
     let start0 = w_lo + frame.delta - 1;
     let end0 = w_hi + frame.delta;
     if start0 < 0 {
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
### Review pass: representation impact measured, and the standard corpus cannot see it

**The blast radius, which needs stating explicitly.** This changes normalized output for transcript-axis cis alleles whose member-span union crosses an exon junction. Measured by enumerating two-member `del`/`dup` pairs over `NM_001234.1` (the 3-exon `MockProvider` record) on all three transcript axes, at `origin/main` versus this branch:

| | rows |
|---|---:|
| pairs compared | 9,360 |
| **moved** | **2,606 (27.8%)** — `c.` 812, `n.` 982, `r.` 812 |
| of those, converged onto the **per-member (lone) form** | **2,606 (100%)** |
| a third form, matching neither side | **0** |

So this is a confluence fix that converges on the representation **already shipped for the lone spelling** — e.g. `c.[1dup;12dup]` moves from `c.[1dup;33dup]` to `c.[1dup;26dup]`, and `c.12dup` alone already normalizes to `c.26dup`. A consumer storing the lone form sees nothing move; one storing the cis form for a junction-crossing pair sees it move onto the form its own lone spelling already had. That is the cheap direction, but it is not free and is worth calling out as a migration.

**`dump_normalized_corpus` reports 0/52,024 moved, and that zero is not evidence.** Its synthetic transcript is built with a single exon spanning the whole record (`examples/dump_normalized_corpus.rs:842`, `Exon::with_genomic(1, 1, tx_len, …)`), so it has no internal junction and `crosses_exon_junction` can never fire on it. The harness's own footer warns about exactly this ("says nothing about a shape this corpus does not build"). Quoting the zero here would be a misread — hence the purpose-built enumeration above. A multi-exon family in that corpus would be worth adding separately.

### Findings fixed

**1. `axis_frame` lost its doc comment.** The new function was inserted between `/// Resolve the group's [AxisFrame], or refuse the group.` and the `fn axis_frame` it documents, so that line became the first line of `crosses_exon_junction`'s docs and `axis_frame` was left undocumented. Moved back.

**2. `n.`-axis coverage (CodeRabbit, minor).** The module doc cites `general.md:44` for `c.`, `r.` **and** `n.`, and the guard covers all three, but only `c.` and `r.` were pinned. Added `the_noncoding_axis_carries_the_same_exception`, with the lone spellings pinned first as the reference answer. Verified as a real guard: with the `crosses_exon_junction` call removed, it fails along with the other three behaviour tests.

**3. Wrapping arithmetic in the new predicate (CodeRabbit, major — accepted in narrowed form).** `exon.end as i64` and the unchecked `lo + frame.delta` are now `i64::try_from` and `checked_add`, both taking the same conservative exit as an unservable transcript. A wrapped `junction` compares as negative and silently reports "no junction crossed", which is the direction that reinstates #1450.

**But the reachable overflow is upstream and pre-existing, not here.** CodeRabbit asked for this at lines 3102-3108; the panic an extreme coordinate actually produces is at `w_hi = c_hi + CANONICAL_PAD` (`merge.rs:2199`), which runs *before* this predicate is reached and which this PR does not touch. Measured:

```
NM_001234.1:c.[9223372036854775806_9223372036854775807del;10del]
  -> panicked at src/normalize/merge.rs:2199: attempt to add with overflow
```

Hardening only the new function would not have prevented that. Filed separately rather than widened into this PR.

Gate green on the rebase: fmt, clippy `--all-features --all-targets -D warnings`, 9499/9499, plus the no-features `--test it` build the soak archive uses.

PR #1481, opened 2026-08-06
test(repr): add a multi-exon coding axis to the representation corpus

### LINKED ISSUE #1478: Representation corpus pins transcript geometry to one exon and CDS_START=1, so #1450 and every 5'UTR shape are unreachable

nh13:
Closes #1478.

## The defect

The corpus pinned transcript geometry to a single value on every axis that matters for junction behaviour — one exon, `CDS_START = 1`, plus strand. Two things were therefore structurally unmeasurable:

- **#1450**, an open cis-member exon-junction defect. `general.md:44` exempts deletions and duplications around exon/exon junctions from the 3′ rule; the per-member pipeline honours it and the sequence-first cis path does not. Its reproducer needs a three-exon transcript. This reference had no exon/exon junction at all.
- **Every 5′UTR shape.** With the CDS starting at the first base there is no transcript position before it. Measured on `2ddc25b9`: **0** rows named a `c.-N` position, on either side of a dump.

Worth noting the module doc deliberately engineered the *other* junction — "five 3'UTR bases remain… #1398, #1418 and both halves of #1426 lived in that region." The 3′UTR was reasoned about carefully; the 5′UTR was left unreachable.

## The fix

A second coding axis, `cx`: three exons over the 20-base core (junctions at transcript 7/8 and 14/15), `CDS_START = 4` so `c.-3`..`c.-1` exist, and genomic exon spans separated by 10 bases of intron so a shift running off an exon end lands in intron rather than silently in the next exon.

| | base | this PR |
|---|---:|---:|
| rows | 33,808 | 50,712 |
| inputs naming a 5′UTR position | **0** | **3,172** |

## It finds something immediately

The re-parse oracle fires **twice** on the new axis, and both outputs genuinely fail `ferro parse` (verified directly, not just via the oracle):

```text
c.[-1_1del;-1_1insCC]   -> c.[-1_1del;1_2dup]       5'UTR/CDS junction
c.[12_*1del;12_*1insCC] -> c.[11_12dup;12_*1del]    CDS/3'UTR junction
```

The first is the shape that could not exist before this commit.

The second has a **pre-existing twin on the `c.` axis** — `c.[15_*1del;15_*1insCC]` fires on `origin/main` today. It is not filed, and it is invisible because `normalize_one` records a panic as data rather than aborting the dump, by design. I have filed that separately; this PR only makes the class visible, and changes no normalizer behaviour.

## Unchanged on purpose

- **All 33,808 pre-existing `g.` and `c.` rows are byte-identical to `origin/main`** (`diff` clean). Re-pointing `CDS_START`/`CDS_END` or the exon layout in place would have re-rolled every existing `c.` row for families with nothing to do with splicing; a second axis does not.
- **`hgvs_pos` was generalised, not branched.** With `CDS_START == 1` the new middle arm reduces to `p` and the 5′UTR arm is unreachable, so single-exon spelling is unchanged — pinned by a loop over every CDS position rather than asserted.
- **The corpus axis token is not the HGVS letter.** `cx` descriptions are spelled `c.`; both prefix sites went through a shared `prefix_for`, since interpolating the token would have emitted `NM_TESTX.1:cx.` and recorded parse errors as measurements.
- **The exons really are spliced.** `the_multi_exon_reference_is_actually_spliced` checks each exon sits at the genomic offset its `Exon` record claims, that consecutive exons do not abut, and that the exon blocks reconstitute the transcript — without the gap this would be a single-exon transcript wearing three exon records, which would look right in the dump and still not express #1450.

## Cost

The corpus grows ~50%, and `compare` refuses dumps that do not cover the same rows, so this re-rolls cross-revision comparability once. That is the argument for treating geometry as one deliberate widening rather than a fourth incremental patch — it is the third blind spot found in this corpus in a day (#1456, #1463, this).

## Representation stability

**No normalized output moves.** Test-corpus surface only.

## Verification

```text
cargo clippy --all-features --all-targets -- -D warnings   clean (exit 0)
all three oracles + FERRO_SWEEP_SEEDS=full                 9437 passed, 0 failed
```

### DIFF
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -118,6 +118,43 @@ const TX_CONTIG: &str = "chr_synth";
 const CDS_START: u64 = 1;
 const CDS_END: u64 = 15;
 
+/// The multi-exon transcript a `cx.` row is drawn against (#1478).
+///
+/// A *second* coding reference rather than a change to the one above. Re-pointing
+/// `CDS_START`/`CDS_END` or the exon layout in place would re-roll every existing
+/// `c.` row and make prior dumps incomparable for families that have nothing to do
+/// with splicing; an added axis leaves them byte-identical, exactly as an added
+/// family does.
+const TX_MULTI_ACCESSION: &str = "NM_TESTX.1";
+/// The contig the multi-exon synthetic transcript is placed on.
+const TX_MULTI_CONTIG: &str = "chr_synth_x";
+
+/// CDS of the multi-exon reference, 1-based inclusive over transcript coordinates.
+///
+/// `CDS_START` is **4**, not 1, which is the whole point: with a CDS starting at the
+/// first base there is no 5'UTR, so `c.-N` is unreachable and every 5'UTR shape
+/// measures as a confident zero. Three bases before it give `c.-3`..`c.-1`. The CDS
+/// is 12 bases (4..=15), still a multiple of three so it stays codon-complete, and
+/// five 3'UTR bases remain as on the single-exon reference.
+const CDS_START_MULTI: u64 = 4;
+const CDS_END_MULTI: u64 = 15;
+
+/// Exon blocks of the multi-exon transcript, as 0-based half-open ranges into the
+/// 20-base core.
+///
+/// Three exons, so there are **two** exon/exon junctions — at transcript 7/8 and
+/// 14/15. `general.md:44` exempts deletions and duplications around such a junction
+/// from the 3' rule, and #1450 is that the per-member pipeline honours the exemption
+/// while the sequence-first cis path does not. A single-exon reference has no
+/// junction at all, which is why that defect cannot appear in this corpus today.
+const EXON_SPANS: &[(usize, usize)] = &[(0, 7), (7, 14), (14, 20)];
+
+/// Intronic bases between consecutive exons in the genomic layout.
+///
+/// Long enough that an exon's genomic span cannot abut its neighbour's, so a shift
+/// that runs off an exon end lands in intron rather than silently in the next exon.
+const INTRON_LEN: usize = 10;
+
 /// The dump header, written on emit and required verbatim on read.
 const HEADER: &str = "reference\taxis\tdirection\tfamily\tinput\toutput\twas_fixed_point\n";
 
@@ -382,12 +419,7 @@ fn complement(base: u8) -> u8 {
 /// them to one form — and whether it can depends on whether the block was capped.
 fn long_inputs_for(axis: &str, core: &str) -> Vec<String> {
     let bytes = core.as_bytes();
-    let acc = if axis == "g" {
-        GENOMIC_CONTIG
-    } else {
-        TX_ACCESSION
-    };
-    let prefix = format!("{acc}:{axis}.");
+    let prefix = prefix_for(axis);
     let inverted: Vec<u8> = bytes.iter().rev().map(|b| complement(*b)).collect();
     let members: Vec<String> = (0..bytes.len())
         .filter(|&i| bytes[i] != inverted[i])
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
### Review pass: the new axis demonstrably closes the gap, and one incoherent slice removed

**Validated against the defect it exists for, rather than by inspection.** #1480 added an exon-junction guard to `canonicalize_from_sequence`, and the corpus as it stood could not see that change at all — it reported **0 of 52,024** rows moved, because its only transcript was built with a single exon spanning the whole record (`Exon::with_genomic(1, 1, tx_len, …)`), so `crosses_exon_junction` could never fire.

Re-running that experiment against this branch: with #1480's guard reverted, the new corpus reports

| | rows |
|---|---:|
| total | 78,028 |
| **moved** | **597 (0.8%)** |
| previously accepted (a migration) | 325 |
| previously not a fixed point (free) | 272 |

spread across `adjacent_junction_ins` (189), `dup_plus_ins` (211), `dup_plus_sub` (103), `del_plus_sub` (53) and `delins_hiding_an_inversion` (4). So the axis genuinely has detection power for the class that was previously invisible, which is the claim worth pinning.

### Fixed: the long cores produced an incoherent multi-exon provider

`EXON_SPANS` is a fixed 7/7/6 split of a **20-base** core, tuned so junctions land inside the shapes the short corpus builds. `multi_exon_provider` passes the whole core to `Transcript::new` regardless, while the exon table and `spliced_genomic` are built from those fixed spans. On a 1024-base `nearpalindrome` core that means the transcript declares 1024 bases, its exon table maps 20, and the contig contains only those 20 — every position past 20 is declared and unmapped.

CodeRabbit flagged this as producing "sentinel normalization rows"; that specific prediction did not hold — all eight long `cx` rows came out well-formed and as fixed points. The underlying problem is real though, and worse for being invisible: rows drawn against a provider that cannot represent its own transcript look like coverage while measuring nothing, and a later change touching genomic projection would get nonsense there that reads as a real result.

Skipped `cx` for the long cores rather than scaling `EXON_SPANS`, which would move the junction geometry the short corpus is tuned around. The long cores exist for `long_block_inversion` — block size, not exon structure — so they lose nothing. Net effect **78,036 → 78,028 rows**: exactly the eight incoherent ones, with all 26,004 short `cx` rows kept.

Re-ran the guard-reverted comparison afterwards to confirm this did not cost detection power: still **597 moved**, identical per-family breakdown.

Gate green on the rebase: fmt, clippy `--all-features --all-targets -D warnings`, 9512/9512, plus the no-features `--test it` build the soak archive uses.

PR #1484, opened 2026-08-06
feat(normalize): widen the axis gate to c./n./r.

### LINKED ISSUE #1235: Normalizer: cis-allele members normalized independently, not from the resulting sequence (root cause of #1229–#1234)

### LINKED ISSUE #1454: A derived member is typed inv only on the second pass, so the first output is not a fixed point

### LINKED ISSUE #1477: Sequence-first canonicalization never runs for a lone transcript-axis member, and the deferral's recorded cost needs refreshing

nh13:
Closes #1477. Refs #1235, #1450, #1480.

## Read this first: the tradeoff is measured, and a reviewer should weigh it

This PR is green — 9 573/9 573 with all three oracles and `FERRO_SWEEP_SEEDS=full`, 9 573/9 573 plain, fmt and clippy clean, 35/35 CI. What needs a decision is not quality but **direction**.

It moves **1 957 real corpus rows out of 5 761 894 (0.034 %)**, and **1 951 of them gain a member** — a single `delins` becoming a two-member allele. That is arity change, not re-spelling: a consumer keying on the normalized string sees the shape of their data change.

Sampling 56 of those moved rows and asking Mutalyzer what it makes of the same input:

| | n | share |
|---|---|---|
| Mutalyzer keeps **one** member; this PR splits it (moves *away* from interop) | 25 | 45 % |
| Mutalyzer splits, and to the same string (moves *toward* interop) | 15 | 27 % |
| Mutalyzer answers a third form neither side produces | 16 | 29 % |

Net −10 of 56. **95 % CI −4.3 % to +40.0 %, so it crosses zero** — the point estimate says agreement worsens on the rows this PR moves, and the sample is not large enough to establish it. Stated as a measurement, not a verdict.

Two honest qualifications. Most of the third-form rows are Mutalyzer emitting **cross-reference notation** (`c.1196_1210delins[TC;1213_1228]`), which ferro does not emit at all — a capability gap, not a disagreement about splitting. And the unmoved 5 759 937 rows are untouched, so this is a statement about the rows that move, not about the corpus.

Related but not blocking: #1517 is calibrating the same merge-versus-split boundary from the inversion side, against the same reference implementation.

## What this changes

`is_splittable_single_member` matched only `HgvsVariant::Genome` and `HgvsVariant::Mt`, so a lone `c.`/`n.`/`r.` member was not admitted to the sequence-first derivation. One spelling of a transcript-axis variant was canonicalized and another was frozen, which is essentially the whole transcript-axis confluence residual.

This widens that gate to the transcript axes and fixes every failure the widening exposes.

## Measured effect: the transcript axes reach genomic parity

Confluence over 11 272 classes / 47 392 spellings, both shuffle directions:

| arm | `g.` | `c.` | `n.` | `r.` |
|---|---|---|---|---|
| before | 70.97 % | 46.70 % | 46.63 % | 46.61 % |
| **after** | 70.97 % | **71.06 %** | **70.97 %** | **70.94 %** |

Three independent transcript axes land on `g.`'s number to two decimal places. That is an asymmetry being closed rather than a heuristic being tuned — this PR does not invent a rule, it stops excluding three axes from an existing one.

**It does not close #1235.** The residual at ~71 % is a partitioner property, identical on all four axes. Closing it needs the canonical partitioner (`FERRO_PARTITION=canonical`, #1464), a separate and much larger representation change. `Refs`, not `Closes`.

## Representation moves, quantified across every corpus we own

Baseline and head both built locally from `ef2b8bfc` and this head, normalized against the prepared reference, row counts validated on both sides (a mismatched count is reported as invalid rather than as a result):

| corpus | rows | moved | % | gain members | lose members | same arity |
|---|---|---|---|---|---|---|
| ClinVar 500k | 500 004 | 145 | 0.029 % | 144 | 0 | 1 |
| multi-member cis | 592 | 5 | 0.845 % | 0 | **5** | 0 |
| CMRG exhaustive | 4 826 063 | 1 422 | 0.029 % | 1 422 | 0 | 0 |
| Paraphase exhaustive | 435 235 | 385 | 0.089 % | 385 | 0 | 0 |
| **total** | **5 761 894** | **1 957** | **0.034 %** | 1 951 | 5 | 1 |

Zero rows error or decline on either side.

**CMRG and Paraphase had never been measured for this.** `cmrg_exhaustive_tests` and `paraphase_exhaustive_tests` call `parse_hgvs` and nothing else, so those 5 261 298 rows contribute no normalization coverage in CI. They are 92 % of the corpus and 1 807 of the 1 957 moved rows — a blast radius that no previous measurement in this campaign could have seen.

The direction differs sharply by corpus, and the reason is population. The three observed-variant corpora contain **197 true multi-member cis alleles between them (0.0034 %)**, and those are widely separated substitutions the partitioner has no choice about. Their moves are single-member `delins` **gaining** members. The hand-built cis corpus is 100 % multi-member by construction, and there the change moves rows to **fewer** members.

## Throughput: this costs ~31%, and most of it looks recoverable

Measured on CMRG's 4 826 063 rows, release binaries built from `origin/main` and this head, runs
**interleaved** base/head rather than blocked so thermal drift cannot masquerade as a
between-version effect. Every rep validated on output row count before its time was recorded — a
truncated run prints a wall-clock number too, and on a 4.8M-row corpus that would read as a
speedup.

| | mean | vs base |
|---|---|---|
| `origin/main` | 17.08 s | — |
| **the gate commit alone** | 25.52 s | **+49 %** |
| **this head (all six commits)** | 22.33 s | **+31 %** |

n = 5 per side for the head comparison, 95 % CI **+23.4 % to +38.0 %**, ranges non-overlapping.
282k -> 216k rows/sec.

**The cost is the widening itself, not the passes added on top** — the gate alone is *worse*
(+49 %) and the four subsequent fixes claw about a third of it back.

**And it is being paid almost entirely by rows that do not change.** Only 1 422 of 4 826 063 CMRG
rows (0.03 %) move. `is_splittable_single_member` is now `edit.inner().is_some()` on five variant
kinds, and `canonicalize_from_sequence` alternates with `normalize_core` up to four times — so
every single-member `c.`/`n.`/`r.` variant applies itself to the reference and re-partitions,
including the ones that structurally cannot split.

Measured share of that population: **~58 % of both CMRG and ClinVar is a lone transcript-axis
substitution** (88 % of all transcript-axis rows). A lone substitution changes exactly one column,
cannot partition into multiple pieces and cannot 3'-shift, so it is derivation-invariant by
construction. **Zero of the 1 957 moved rows across all four corpora was one.** An early-out for
that class should therefore be behaviour-neutral and recover most of the regression — worth doing
before merge, and worth proving by re-running the corpus sweep for byte-identity rather than by
argument.

A caveat on ClinVar: measured at 500k rows the same comparison gives +3 % with a 30 % rep spread,
i.e. no resolvable signal. That number should not be quoted — the corpus is too small and the run
too short to see a 30 % effect, which is exactly why the CMRG measurement exists.

## Census re-blesses, with the argument for each

**`cis_confluence_axis` — `converged` 6 633 → 8 006 (3'), 6 628 → 8 006 (5').** `split_two` falls by the same amount; `split_three` (115) and `split_more` (19/9) unchanged. Every divergence figure moved down or stayed flat, which is the direction that pin demands.

The 5' figure moves by 1 378 against the 3' direction's 1 373 — five more, because the same change fixes an off-by-one in `enclosing_exon` that let a member sitting on an exon's *first* base escape the window clamp and shuffle across the junction. A 5'-only symptom, since the 3' walk moves away from the exon start. Both directions now land on the same number, which is the reading to trust: the residual is a property of the partitioner, not of a shuffle direction.

Measured identical with and without `FERRO_ASSERT_IDEMPOTENT` (`declined: 0` in both). That is #1493's doing, not this branch's — it closed #1454, so the two classes that used to panic the oracle no longer do, and the per-oracle census pins this file once carried were already collapsed on `main`.

**`spec_enumeration` — `projection-splits-single-member` 9 → 13, `projection-pinned` 1168 → 1164. A deliberate re-bless of a regression, not a fix.**

All four rows are one input, `NM_004006.2:c.76_83inv`, on `project-c`/`-n`/`-r`/`-p`:

```
c.76_83inv  ->  c.[76_77delinsTG;82_83delinsTT]
p.(Asn26_Gln28delinsCysAlaLeu)  ->  p.[(Asn26Cys);(Gln28Leu)]
```

They enter that bucket by **shape** — a single member projecting to an allele — not by the divergence its headline citation `delins.md:42` names; that passage is about variants **one** base apart, and these are four. And the behaviour is axis-neutral and already shipped: `c.76_83` is `AATGCACA`, whose reverse complement `TGTGCATT` coincides at four of eight columns; the block is equal-length so no axis-dependent gate is consulted, and `g.` has always produced this split. The widening did not create it, it let the transcript axes see it.

**It is nonetheless the wrong answer, and #1517 fixes it separately.** Mutalyzer converges *both* spellings onto `c.76_83inv`. Fixing it means loosening `coalesce_whole_block_inversion`, which is axis-neutral and shipped, so it moves genomic representations for a class this PR measures nowhere.

## Two defects, not one

A controlled A/B established this: reverting only the three-line gate and regenerating the enumeration restored the census exactly, so every move is attributable to the widening and nothing else on the branch.

**Defect A (fixed here) — the live block splitter has no axis at all.** `general.md:34` splits two variants separated by one or more nucleotides; `general.md:35` excepts "separated by one nucleotide, together affecting one amino acid", which cannot reach a genomic description. `axis_min_separation(reading_frame)` encodes that and was handed only to the sequence-first splitter. On the live path the sole coding rule was `apply_coding_codon_exception`, matching `[Sub@p; Identity@p+1; Sub@p+2]` exactly, so it cannot reach a pair where either member is an insertion, deletion or wider delins. One coincidentally matched interior base therefore split coding descriptions the spec keeps whole:

- `NM_000143.3:c.44_47delinsATC` → `c.[44_45delinsAT;49del]`
- `NM_004006.2:c.4375_4376delinsAGTT` → `c.[4375C>A;4376_4377insTT]`, which also degraded `p.(Arg1459SerfsTer19)` → `p.?` — literally the harm `delins.md:47` names
- `NM_000088.3:r.4_6delinsacgu` → `r.[3_4insa;5_6delinsgu]`

New `coalesce_coding_frame_separation` pairwise-merges pieces one unchanged base apart, on a reading-frame axis, on a **length-changing** block only — equal-length blocks keep going to `apply_coding_codon_exception`, preserving the codon half of the exception.

**Defect B (not fixed here) — `c.76_83inv`,** moved to #1517.

## Placement is forced, and was established by getting it wrong first

An earlier attempt put the rule inside `partition_block` via `separations_are_meaningful`. It cleared the same three shapes and cost **332 converged confluence classes per direction** (3' 8 007 → 7 675, 5' 8 007 → 7 673) plus a non-idempotent output — because the weight bound compares against the *input's* weight, so the merged form clears it for the spelling that arrives as one `delins` and trips it for the allele spelling of the same variant. That is the hazard `coalesce_whole_block_inversion` already documents. The bound now judges `derived_columns`, the un-widened partition put through the same shift/coalesce/shrink.

## Known residual, documented rather than hidden

The 3'-shift can *create* a one-base gap after the pre-shift pass has run. Three of 500 004 ClinVar rows land that way (`NM_001458.4:c.7242_7246delCAGCCinsAAACCTG`, `LRG_712t1:c.845_852delATCCCAATinsTCTTCATAGA`, `NM_001406642.1:c.1897_1910delinsTAATATAATT`), plus a second gap in two rows the pass does otherwise reduce. Closing it means running the rule again post-shift with the un-widened partition carried through that pass too. None of the four failures is in that class. Noted in the new function's doc.

## Correction to an earlier revision of this description

An earlier revision said a lone `c.`/`n.`/`r.` member "never reaches `sequence_first_pass`". Too strong, and now corrected in the first commit's message. A lone transcript-axis `delins` that the per-member pipeline splits into an allele *first* then satisfies the **allele** arm: 121 `c.` rows in CMRG move under `FERRO_PARTITION=canonical` with no gate at all, 120 with a multi-member live output — that second door's fingerprint. Widening the predicate closes the lone-member door; it does not create transcript-axis derivation from nothing.

## Related, deliberately not in this PR

- **#1517** — `coalesce_whole_block_inversion` under-recognises a whole-block reverse complement whose interior coincides. Owns the four re-blessed rows above.
- **#1518 / #1519** — equivalence classes and ruling records in the spec fixture, so two spellings of one variant can no longer both pass as `preserved`.

## Verification

Full gate with all three oracles and `FERRO_SWEEP_SEEDS=full`, plus a second run without the oracles as CI's gating `Test` job runs it — 9 573/9 573 both times. `cargo fmt --check` and `cargo clippy --features dev --all-targets -- -D warnings` clean. Corpus measurements taken against the prepared reference with both sides built locally and row counts validated. Mutalyzer sampling is n = 56 of 1 957 moved rows, public accessions only.

---

## Post-rebase update (2026-08-07): three new measurements, one retraction

Rebased onto `main` (was 7 behind). Full gate re-run: **9 589/9 589**, nothing re-blessed by the rebase.

### The strongest evidence for this PR was missing from the description

`9dd49500` re-blesses the cis-allele confluence census from **6 633 to 8 006 converged of 11 272 classes — a gain of 1 373**, with `declined` and `sequence_changed` both 0 either side. That is the largest single confluence improvement in the project's record and it was never stated here. It is a better argument for merging than the split-direction numbers below.

### A retraction: "monotone, 0 regressions" does not survive

An earlier measurement reported this PR fixing **1 042 split-direction violations with 0 regressions**. That measured *who left the lone-`delins` candidate pool*, not who was fixed — the join's key breakdown is `(violating, absent) = 1042` and `(violating, clean) = 0`, so no row was ever shown fixed in place. The instrument admits only **lone** `delins`, and this PR's whole mechanism is turning lone `delins` into multi-member alleles, so the change moves rows out of the only population the instrument could see.

Adjudicating the multi-member outputs this PR *produces* — reference-backed, with the codon test applied to every interior run rather than only the first:

| population | main | this PR | delta |
|---|---|---|---|
| lone `delins` violations | 5 828 | 4 786 | **−1 042** |
| violations *inside* the members it creates | 107 | 275 | **+168** |
| **total** | **5 935** | **5 061** | **−874 (−14.7 %)** |

So the honest claim is **a net removal of 874 separation violations, not 1 042 with none added**. The members this PR creates violate at 37.9 % against main's 30.2 %, on 726 adjudicated members against 354 — it produces roughly twice as many multi-member alleles, and they are slightly worse per member while the total falls.

Method: candidates are `delins` members with a reference span ≥ 4 (the only shape that can conceal an unsplit interior); bases resolved via `samtools faidx` + cdot `start_codon`, never via ferro, since ferro's cuts are what is under test; a violation is claimed only where the unchanged interior survives **every** minimal alignment, so it can under-report and never over-report. Coordinate self-check (a normalized `delins` must differ at both ends): **305/305** and **223/223**.

### Stability disclosure

The one genuine canonical-form move in the tail is `NM_000342.4`, from the 5-member spelling to the 4-member spelling — one row in 5 761 894. Anyone storing the 5-member form must re-normalize it. The other five tail rows are confluence *fixes*: under `main` the same variant spelled two ways gave two different answers, and this PR closes all five.

Note also that `4a6e5592` is **axis-neutral** — it moves `g.` rows too — so this disclosure is not scoped to `c.`/`n.`/`r.` despite the PR title.

### Ordering hazard for whoever merges this

This PR and the #1535 → #1537 stack both re-bless `tests/it/cis_confluence_axis.rs`. Whichever merges second will conflict there. If this one lands first, #1537's census pin needs re-blessing against post-#1484 numbers.

---

## Post-#1540 update (2026-08-07): the guard that blocked this, and what clearing it cost

`main` gained #1540's `issue_1539_split_member_separation` after this PR opened, and this PR failed it on **all five** of its rows. The guard asserts that no member ferro emits may hold a reference column that is unchanged in *every* minimal alignment, sitting strictly between two changed ones, outside `general.md:35`'s one-amino-acid exception. This update adds one commit that clears it, plus a re-bless of the cis confluence census in the same commit.

### What the five rows actually were

Three distinct variants on five transcripts. In each, `best_alignment` cut the block at columns that are unchanged in *no* minimal alignment, and the member it carved out concealed one that is unchanged in *every* one:

| input | member emitted | concealed column |
|---|---|---|
| `NM_000089.3/.4:c.2148_2156delinsACGTGG` | `c.2148_2152delinsAC` (`GGTCG → AC`) | offset 3, codons 716/717 |
| `NM_001277115.2:c.3545_3551delinsAGATGATGAACTGTAGCTGTAAGA` | `c.3545_3548delinsAGAT` (`GACA → AGAT`) | offset 1, codons 1181/1182 |
| `NM_000296.4` / `NM_001009944.3:c.10317_10334delinsATGGCC` | `c.10317_10330delinsAT` (`CCAGGCGGGCCATG → AT`) | offset 12, codons 3442/3443 |

Two of the three parent blocks — `GGTCGGTCC → ACGTGG` and `CCAGGCGGGCCATGGGCT → ATGGCC` — have **no forced-unchanged column at all**, so the separations their splits rested on are artefacts of where the single gap landed, while the genuine one was swallowed.

They now normalize to:

```
NM_000089.3:c.2148_2156delinsACGTGG      -> c.[2148_2150delinsA;2153del;2155_2156inv]
NM_001277115.2:c.3545_3551delins…        -> c.[3545del;3547_3548delinsGAT;3551delinsTGAACTGTAGCTGTAAGA]
NM_000296.4:c.10317_10334delinsATGGCC    -> c.[10317_10327del;10332del;10334T>C]
```

### The fix

`split_concealed_separations` re-partitions an emitted member at the columns every minimal alignment matches, applying the codon exception to **each** interior run. Three scopings carry it, and each was established by a measured failure rather than by argument:

- **Length-changing blocks only.** On an equal-length block the alignment is position-wise, so a matched column is a genuine identity and `partition_block` has already cut at every one. Auditing equal-length blocks re-splits `NM_TEST.1:c.10_15delinsTAATAT` — the rotation `AATATA → TAATAT`, which #1524 settled as its own normal form — and cost four #1454 pins plus **71 declined / 18 underdetermined** census classes.
- **Only a partition that already claimed a separation.** The spec's own `LRG_199t1:c.850_901delinsTTCCTCGATGCCTG` (`delins.md:44-47`) **carries two forced-unchanged interior columns of its own**, at reference offsets 44 and 48, and offset 44 separates codons 297 and 298. A rule that cut every unlicensed interior would shred the very description `delins.md:47` recommends. It is spared structurally, not by exemption: `partition_block` returns that block as one piece.
- **Not a whole-span reverse complement**, whose interior coincidences are structure under `inversion.md:5`. Without this, #1461's dense inversion stopped being recognised.

The node criterion is used rather than `Dominators::matched_ref`, which answers the stronger "which match *edge* lies on every path" and therefore reports nothing at all for `GACA → AGAT` — a column matched by every minimal alignment, at two different alternate offsets. `general.md:34` asks about a nucleotide, not about where the payload sits, so the node criterion is the one the rule needs; the edge criterion reaches two of the three variants.

### The alternative was implemented, measured, and rejected

Refusing the split instead — falling back to the spanning `delins` `delins.md:47` recommends — clears the same guard, restores `main`'s string for all five rows, and is the more spec-coherent rule. It costs **394 converged confluence classes of 11 272** (3': 8 003 → 7 612), because the weight bound then accepts the merged form for the lone-`delins` spelling and refuses it for the allele spelling of the same variant. Confluence outranks stability, so the cut wins. The measurement is recorded on the new function, because the refusal looks strictly safer than it is.

### A near-miss worth recording

An earlier revision cut a run of two matched columns without checking that the canonical alignment matched them to *consecutive* alternate offsets. Where an insertion sits between two forced-matched columns, that dropped payload bases — a **changed sequence**. The round-trip guard at the end of `canonicalize_from_sequence` caught every instance and declined, so no wrong string was ever emitted, but the census reported it as 45 declined / 10 underdetermined **while `converged` rose**. A pass that corrupts a sequence can raise the convergence figure, so the three zeros have to be read before the headline.

### Census re-bless, monotone in every figure

| | 3' before | 3' after | 5' before | 5' after |
|---|---|---|---|---|
| converged | 8 003 | **8 026** | 8 003 | **8 021** |
| split 2 | 3 135 | 3 112 | 3 155 | 3 137 |
| split 3 / 4+ | 115 / 19 | 115 / 19 | 105 / 9 | 105 / 9 |
| declined / underdetermined / sequence changed | 0 / 0 / 0 | 0 / 0 / 0 | 0 / 0 / 0 | 0 / 0 / 0 |

Every class that moved moved from two outputs to one.

### Blast radius, remeasured end to end against `origin/main`

Both sides built locally (`origin/main` at `eb75d389`, this head), normalized against the prepared reference, with row counts and the changed/unchanged/failed summary validated identical on each side.

| corpus | rows | moved by this PR | of which added by this commit |
|---|---|---|---|
| ClinVar 500k | 500 004 | 158 | 15 |
| CMRG exhaustive | 4 826 063 | 1 561 | 174 |
| Paraphase exhaustive | 435 235 | 308 | 64 |
| **total** | **5 761 302** | **2 027 (0.035 %)** | **253** |

Of the 253 rows this commit moves, **179 were already moving under this PR** — it only changes which new form they land on, so they cost the consumer nothing extra. The remaining **74 are new, and all 74 are `g.`**: pre-existing `general.md:34` violations on the genomic axis, which the widened gate never touched and which this pass, being axis-neutral, now fixes. Every one of the 253 is the target shape; a row-by-row read of the 15 ClinVar rows found no unrelated churn.

`examples/dump_normalized_corpus.rs` reports **0 of 78 298** for this commit. That zero is **structural and must not be quoted as evidence**: the shape this change keys on is a member of a *length-changing multi-piece* partition whose own minimal alignment needs two gaps, and no family in that generator builds one. The real-corpus figures above are the measurement.

### Verification

`cargo fmt --all -- --check` and `cargo clippy --features dev --all-targets -- -D warnings` clean. `cargo nextest run --features dev -E 'binary(it)'` — **5 506/5 506**, and again with all three seam oracles plus `FERRO_SWEEP_SEEDS=full` — 5 506/5 506. Nothing re-blessed except the two census pins above.

Representation-Change: 2 027 rows move of 5 761 302 (0.035 %), measured against `origin/main` with both sides built locally and row counts validated; 2 016 gain a member, 0 lose, 11 same arity. Arity change, not respelling, so a consumer keying on the normalized string sees the shape of their data change — a real migration. By axis: 1 827 `c.`, 126 `n.`, 74 `g.` — the `g.` rows are pre-existing `general.md:34` violations the axis gate never touched, fixed by the axis-neutral member audit. Confluence improves: 1 373 classes of 11 272 from the gate, plus 23 (3') / 18 (5') from the member audit, with `declined`, `underdetermined` and `sequence_changed` all 0. Separation violations fall by a net 874 (−14.7 %) before the member audit and further after it. One genuine canonical-form move in the tail (`NM_000342.4`, 5-member → 4-member); the other five tail rows are confluence fixes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved variant normalization across genomic, transcript, RNA, and coding descriptions.
  - Correctly renders one-base substitutions when reference sequence is available.
  - Improved handling of exon boundaries, circular coordinates, insertions, duplications, inversions, and repeat regions.
  - Preserves appropriate `delins` blocks and applies more accurate coding-frame and downstream-shifting behavior.
  - Added safeguards for invalid coordinates, overlapping regions, oversized operations, and unknown reference sequences.
- **Tests**
  - Expanded coverage for boundary cases, multi-member descriptions, sequence comparisons, and projection results.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2507,6 +2507,15 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
         piece.ref_start += lo;
         piece.ref_end += lo;
     }
+    // `general.md:35`'s coding exception, applied here — before the 3'-shift and
+    // before the weight bound — for two independent reasons, each of which was
+    // measured by getting it wrong. See `coalesce_coding_frame_separation`.
+    let unwidened = coalesce_coding_frame_separation(
+        &mut pieces,
+        frame.reading_frame,
+        hi_ref != hi_alt,
+        &ref_bytes,
+    );
     shift_pieces(&mut pieces, &ref_bytes, direction);
     coalesce_adjacent_pieces(&mut pieces);
     // Partitioning decides where the members are; this decides how wide each
@@ -2544,7 +2553,26 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
     // Strict on purpose: #1229, #1231, #1233 and #1234 all sit exactly at
     // equality, and equality is where prioritisation (`general.md:56`) may
     // legitimately prefer the re-derived shape.
-    if changed_columns_of_pieces(&pieces) > changed_columns_of_edits(&edits) {
+    //
+    // The quantity judged is the **un-widened** partition's, which is what
+    // `unwidened` carries: `coalesce_coding_frame_separation` above widens, so
+    // letting the bound see its product would judge a widening against the
+    // input's weight — accepted for the spelling that arrived as one `delins`,
+    // refused for the spelling that arrived as separate members. Measured: 332
+    // converged classes lost per direction over the 11,272-class corpus (3':
+    // 8,007 -> 7,675) when the bound saw the widened pieces. The un-widened
+    // partition is put through the same 3'-shift, coalesce and shrink so the two
+    // weights are comparable.
+    let derived_columns = match unwidened {
+        None => changed_columns_of_pieces(&pieces),
+        Some(mut raw) => {
+            shift_pieces(&mut raw, &ref_bytes, direction);
+            coalesce_adjacent_pieces(&mut raw);
+            shrink_pieces_to_differences(&mut raw, &ref_bytes);
+            changed_columns_of_pieces(&raw)
+        }
+    };
+    if derived_columns > changed_columns_of_edits(&edits) {
         return None;
     }
 
@@ -5695,6 +5723,124 @@ fn split_codon_incompatible_triplets(
         );
         index += 2;
     }
+}
+
+/// Merge runs of change separated by a single unchanged base on a **reading
+/// frame** axis, returning the pieces as they were if anything merged.
+///
+/// # The rule
+///
+/// `general.md:34` describes two variants separated by one or more nucleotides
+/// individually; `general.md:35` excepts "two variants separated by one
+/// nucleotide, together affecting one amino acid", which is a coding-axis rule
+/// and reaches no genomic description at all. [`axis_min_separation`] is that
+/// distinction, already handed to the sequence-first splitter; the live path had
+/// no axis at all, so a single coincidentally matched base split coding `delins`
+/// descriptions the spec keeps whole. That is the gap #1235's widened axis gate
+/// exposed on `c.`/`n.`/`r.`.
+///
+/// Scoped to **length-changing** blocks. An equal-length block reaches
+/// [`apply_coding_codon_exception`], which applies the same exception
+/// triplet-precisely (`[Sub@p; Identity@p+1; Sub@p+2]` with `p` and `p+2` in one
+/// codon) and must keep owning that shape: merging it here instead would drop
+/// the codon half of the exception, which `general.md:35` states and
+/// `test_no_codon_frame_pair_straddles_codon_boundary` pins. On a
+/// length-changing block that shape cannot arise — the pieces do not pair up
+/// position-wise — so the two passes do not overlap.
+///
+/// The merge is **pairwise**, not whole-block: only pieces one unchanged base
+/// apart join, and a genuinely separated pair stays split. Refusing the whole
+/// partition instead (the first attempt) over-merges, and it also loses the
+/// distinction `separations_are_meaningful` draws for wider gaps.
+///
+/// # Placement, which is the whole of the difficulty
+///
+/// **Before the 3'-shift.** The shift moves a piece into an unchanged run, so a
+/// pair one base apart at partition time can be three apart by the time the
+/// post-shift passes see it. `NM_000143.3:c.44_47delinsATC` is the case:
+/// `TGCG -> ATC` partitions as `[44_45delinsAT]` + `[47del]`, one base apart,
+/// and the deletion then 3'-shifts down the `G` run at c.47-49 to `49del`.
+/// Running this after the shift leaves that pair unmerged and the oracle
+/// divergence in place.
+///
+/// **Exempt from the changed-columns weight bound**, which is why this returns
+/// the pieces it replaced rather than nothing. The bound compares the derived
+/// weight against the *input's*, so a widening judged by it is accepted for one
+/// spelling of a variant and refused for another —
+/// `c.9delinsATC` weighs 3 and `c.[8_9insA;9_10insC]` weighs 2 for the same
+/// variant, so the merged form clears the bound for the first and trips it for
+/// the second. Measured on that corpus: 332 converged classes lost per
+/// direction (3': 8,007 -> 7,675; 5': 8,007 -> 7,673) with the bound judging
+/// the merged pieces. This is the same rule
+/// [`coalesce_whole_block_inversion`] and [`apply_coding_codon_exception`] are
+/// placed by, stated there as: *any partition rule that widens a piece must run
+/// after the weight bound, never inside [`partition_block`]*. This one cannot
+/// run after it and still be before the shift, so the bound is handed the
+/// un-widened partition instead — the same exemption, reached the other way.
+///
+/// # Known residual: a gap the shift *creates*
+///
+/// Running before the shift is necessary and not sufficient. The shift moves
+/// piece boundaries, so it can also bring a pair that was two or more bases
+/// apart at partition time down to one — the mirror image of the case above,
+/// and invisible here because it has not happened yet.
+///
+/// Measured over 500,004 ClinVar rows: **3** land with a surviving one-base gap
+/// on a coding axis (`NM_001458.4:c.7242_7246delCAGCCinsAAACCTG`,
+/// `LRG_712t1:c.845_852delATCCCAATinsTCTTCATAGA`,
+/// `NM_001406642.1:c.1897_1910delinsTAATATAATT`), plus a second gap in two rows
+/// this pass does otherwise reduce. All are multi-member partitions where a
+/// *deletion* piece shifts toward a substitution.
+///
+/// Closing it means running the rule a second time after the shift, which needs
+/// the un-widened partition carried through that pass too — the bookkeeping
+/// above, done twice. Left undone deliberately: none of the four failures this
+/// pass was written for is in that class, and a second widening judged against
+/// the wrong weight is exactly the mistake this doc records.
+///
+/// Returns `None` when nothing merged, so the caller pays nothing on the path
+/// that does not widen.
+fn coalesce_coding_frame_separation(
+    pieces: &mut Vec<Piece>,
+    reading_frame: bool,
+    length_changing: bool,
+    ref_bytes: &[u8],
+) -> Option<Vec<Piece>> {
+    // `ref_end` is exclusive and a pure insertion occupies no width, so a gap of
+    // one is one unchanged *base* — the same counting `separations_are_meaningful`
+    // does, and for the same reason.
+    //
+    // Scanned before anything is copied: the copy exists only to hand the weight
+    // bound the un-widened partition, so a block that merges nothing must not pay
+    // for one. Every coding length-changing block reaches here.
+    let joins = |pair: &[Piece]| pair[1].ref_start.saturating_sub(pair[0].ref_end) == 1;
+    if !reading_frame || !length_changing || !pieces.windows(2).any(joins) {
+        return None;
+    }
+    let unwidened = pieces.clone();
+    let mut index = 1;
+    while index < pieces.len() {
+        // `gap == 1` puts `pieces[index].ref_start` at `previous.ref_end + 1`, and
+        // a piece never runs past the window, so `previous.ref_end` indexes a real
+        // byte. Read through `get` regardless: this is the only raw index in the
+        // pass, and declining to merge is a sound answer where a bare index would
+        // panic.
+        let Some(middle) = (joins(&pieces[index - 1..=index]))
+            .then(|| ref_bytes.get(pieces[index - 1].ref_end).copied())
+            .flatten()
+        else {
+            index += 1;
+            continue;
+        };
+        let next = pieces.remove(index);
+        let previous = &mut pieces[index - 1];
+        // The unchanged base is interior to the merged replacement now, so it has
+        // to be carried through explicitly.
+        previous.alt.push(middle);
+        previous.alt.extend_from_slice(&next.alt);
+        previous.ref_end = next.ref_end;
+    }
+    Some(unwidened)
 }
 
 /// Merge pieces the 3'-shift left touching.
@@ -10598,6 +10744,76 @@ mod tests {
             },
         ];
         coalesce_adjacent_pieces(&mut pieces);
+    }
+
+    /// `general.md:35`'s exception on a length-changing block, asserted against
+    /// the pass itself (#1235 review).
+    ///
+    /// Its integration-level sibling —
+    /// `merge_consecutive_edits_tests::coding_length_changing_block_merges_across_one_unchanged_base`
+    /// — can only assert that the `c.` answer equals its input, and that
+    /// equality is satisfied by *any* upstream decline (a window clamp, the
+    /// weight bound, the round-trip guard) just as well as by the merge. Only a
+    /// direct call proves the pass ran, so the discriminating assertion lives
+    /// here and that test carries the control that ties the two together.
+    ///
+    /// `TGCA -> AAC` is that test's `NM_TEST.1:c.2_5delinsAAC`. The payload
+    /// coincides at exactly one column, which is what leaves two runs one
+    /// unchanged base apart for the exception to close.
+    #[test]
+    fn coalesce_coding_frame_separation_merges_one_base_gap_only_in_frame() {
+        let reference = b"TGCA";
+        let split = partition_block(reference, b"AAC");
+        assert_eq!(
+            split.len(),
+            2,
+            "the fixture must actually split, else the merge below proves \
+             nothing; got {split:?}",
+        );
+        assert_eq!(
+            split[1].ref_start - split[0].ref_end,
+            1,
+            "and the two runs must be exactly one unchanged base apart, which is \
+             the gap the exception is scoped to; got {split:?}",
+        );
+
+        // In a reading frame the pair rejoins into the whole block, and the
+        // un-widened partition comes back for the weight bound to judge.
+        let mut in_frame = split.clone();
+        let unwidened = coalesce_coding_frame_separation(&mut in_frame, true, true, reference);
+        assert_eq!(
+            unwidened.as_deref(),
+            Some(&split[..]),
+            "a widening pass must return the pieces it replaced",
+        );
+        assert_eq!(in_frame.len(), 1, "got {in_frame:?}");
+        assert_eq!(in_frame[0].ref_start, 0);
+        assert_eq!(in_frame[0].ref_end, reference.len());
+        assert_eq!(in_frame[0].alt, b"AAC");
+
+        // With no reading frame it declines, which is `general.md:34`'s plain
+        // rule and the `n.` answer the sibling test pins end to end.
+        let mut no_frame = split.clone();
+        assert!(coalesce_coding_frame_separation(&mut no_frame, false, true, reference).is_none());
+        assert_eq!(
+            no_frame, split,
+            "a declined pass must leave the pieces untouched",
+        );
+
+        // And it declines an equal-length block even in frame, which is the
+        // half of the scoping that keeps `apply_coding_codon_exception` owning
+        // the `[Sub@p; Identity@p+1; Sub@p+2]` triplet — merging that shape here
+        // instead would drop the codon half of `general.md:35`. Asserted with
+        // this same pair so only the flag varies.
+        let mut equal_length = split.clone();
+        assert!(
+            coalesce_coding_frame_separation(&mut equal_length, true, false, reference).is_none()
+        );
+        assert_eq!(
+            equal_length, split,
+            "the length-changing scope is what routes an equal-length block to \
+             `apply_coding_codon_exception`",
+        );
     }
 
     /// The enumerated classes of disagreement between the two block splitters.
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -3353,8 +3353,16 @@ fn enclosing_exon<P: ReferenceProvider>(
     }
     let transcript = provider.get_transcript(accession).ok()?;
     let (tx_lo, tx_hi) = (lo + frame.delta, hi + frame.delta);
+    // `Exon` stores 1-based INCLUSIVE bounds — `[(1, 15), (16, 30), (31, 44)]` on
+    // `MockProvider`'s `NM_001234.1` — so `start` is already the exon's first
+    // base. Adding one treats them as 0-based half-open and makes the enclosure
+    // test miss any member sitting on an exon's first base: `c.12del` on that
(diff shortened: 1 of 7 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #1040, opened 2026-07-16
Normalizer over-recognizes `inv`: reverse-complement sub-window carved out of a larger contiguous delins run (v0.8.1)


msto:
> Authorship: the content below was drafted by GitHub Copilot CLI (an AI coding agent) and filed via gh under @msto's account. It reflects the agent's analysis, not a statement authored by @msto.

> ### Summary
> In a ferro-vs-mutalyzer normalization comparison run on **v0.8.1**, ferro emits an `inv` that mutalyzer does not, by carving a reverse-complement sub-window out of a **larger contiguous change** that should remain a single `delins`. Mutalyzer keeps the whole run as one `delins`; ferro splits the run and types the incidental reverse-complement sub-window as `inv`.
>
> ### Anonymized example
> Contig name, coordinates, and bases are synthetic; only the discrepancy structure and the reverse-complement relationship that drives the call are retained.
>
> ```
> raw   : anon_ctg:g.[11C>T;12A>G;13G>T;14T>C;15G>A;16A>C;17C>G;18T>A;19A>C;20G>T]
> mut   : anon_ctg:g.11_20delinsTGTCACGACT                     # one delins over the whole run
> ferro : anon_ctg:g.[11_12delinsTG;13_16inv;17_20delinsGACT]  # ferro carves 13_16 as inv
>
> ref[13_16] = GTGA   revcomp(GTGA) = TCAC   # equals ferro's inv alt, but 13_16 sits inside the 11–20 run
> ```
>
> ### Expected
> Positions 11–20 form a single contiguous change, so the compliant normalization is one `delins` (as mutalyzer produces). The 4 bp reverse-complement match at 13–16 is incidental to a larger event and should not be split out as an `inv`.
>
> ### Observed
> ferro splits the contiguous run and types 13–16 as `inv` solely because `ref[13_16]` reverse-complements to the alt at that sub-window, even though it is embedded in the larger delins.
>
> ### Impact
> This is a common ferro→mutalyzer divergence observed in the v0.8.1 comparison.
>
> ### Related
> Appears to be an over-application of #160 (recognize reverse-complement sub-spans as `inv` within compound variants) — the RC-recognition should not fire for a sub-window nested inside a larger contiguous delins run.

nh13:
@msto let me know if this is still an issue!

PR #1485, opened 2026-08-06
fix(normalize): recognise a dense inversion across multi-base separations

### LINKED ISSUE #1461: Normalizer: a valid 457 nt inversion is rendered as a 109-member allele

nh13:
## The defect

`NC_000013.10:g.100809575_100810031inv` is a valid 457 nt inversion — the observed sequence is the exact reverse complement of the reference block, and the block is **not** palindromic, so the inversion is well-defined and non-degenerate.

The shipped normalizer rendered it as a **109-member** allele:

```
NC_000013.10:g.[100809575A>T;100809577_100809579delinsCGT;100809581_100809584delinsAATT;…]
```

`general.md:56` ranks inversion above substitution and delins, and this is one variant. Both spellings denote the same sequence, so this is a representation defect rather than a correctness one — but a 109-member allele is unusable downstream, and it destroys an input that was already in the recommended form.

## Where it was lost

Instrumented rather than guessed. The gate is `coalesce_whole_block_inversion`, whose `every_separation_is_a_single_base` precondition admits only spans whose members are separated by one unchanged base. The real gap distribution of the 109 members is `{1: 78, 2: 24, 3: 2, 4: 2, 5: 2}` — so the span is refused.

## Why widening the separation threshold cannot work

This is the part worth reviewing, because it is the obvious fix and it is wrong:

| case | span | changed | widest gap |
|---|---|---|---|
| `g.100809575_100810031inv` (#1461) | 457 nt | 307 (**67.2%**) | **5** |
| `AAGCTA → TAGCTT` (the #1040 control) | 6 nt | 2 (**33.3%**) | **4** |

**The inversion that must be recognised has wider gaps than the pair that must stay individual.** So every threshold on separation either refuses #1461 or admits the control — which is exactly why an earlier attempt at widening this gate broke `a_whole_span_reverse_complement_is_not_merged_across_a_multi_base_separation`, the test named for that mistake.

## The fix

`changed_columns_dominate_the_span` — a second, `||` admission route: recognise the inversion when the changed columns are the majority of the span.

That separates the two cases for a reason rather than by fitting a constant. Reverse complementing a random block leaves roughly a quarter of its positions coincidentally unchanged, which is what #1461's 32.8% looks like. The #1040 control leaves **66.7%** unchanged because that span is very nearly its own reverse complement, so its two edits are independent changes sitting inside a near-palindrome — precisely the reading that test's own comment argues for ("two independent changes rather than interior columns of one reverse-complement relation").

**Additive by construction.** It is an alternative to the existing gate, never a replacement, so every block that coalesces today still coalesces; only a currently-*refused* block can change. That is what confines the blast radius to the reported class.

## Blast radius, measured

Default path, real prepared reference, baseline vs this branch:

| corpus | rows | moved |
|---|---|---|
| real multi-member cis alleles | 592 | **0** |
| ClinVar | 500,004 | **1** — the reported row itself |

## Tests

- `a_dense_inversion_is_recognised_across_multi_base_separations` — an integration test on `AACAACCGCGTG → CACGCGGTTGTT`, which reproduces the shape at testable size: three members separated by **two** unchanged bases each (so the single-base gate refuses it) with 8 of 12 columns changed. Three spellings all converge on the `inv`. It sits directly beside the #1040 control so the admit/refuse pair is read together.
- `density_separates_the_inversion_from_the_over_recognition_control` — pins both real geometries as numbers, including the assertion that the #1461 shape fails `every_separation_is_a_single_base` and is admitted by density alone.

## Functions changed

`src/normalize/merge.rs`: `coalesce_whole_block_inversion` (its guard only) and the new `changed_columns_dominate_the_span`. Nothing else in that file is touched — flagged because `merge.rs` is currently contended with PR #1470 and sibling work.

Gate: `cargo fmt --check` and `clippy --all-features --all-targets -D warnings` clean; `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1 FERRO_SWEEP_SEEDS=full cargo nextest run --features dev` → `9468 tests run: 9468 passed, 145 skipped`.

Closes #1461

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -3851,7 +3851,9 @@ fn partition_block(reference: &[u8], result: &[u8]) -> Vec<Piece> {
 /// rule that widens a piece must run after the weight bound, never inside
 /// [`partition_block`].**
 fn coalesce_whole_block_inversion(pieces: &mut Vec<Piece>, ref_bytes: &[u8]) {
-    if pieces.len() < 2 || !every_separation_is_a_single_base(pieces) {
+    if pieces.len() < 2
+        || !(every_separation_is_a_single_base(pieces) || changed_columns_dominate_the_span(pieces))
+    {
         return;
     }
     let start = pieces[0].ref_start;
@@ -3895,6 +3897,48 @@ fn coalesce_whole_block_inversion(pieces: &mut Vec<Piece>, ref_bytes: &[u8]) {
     if is_inversion(&whole, ref_bytes) {
         *pieces = vec![whole];
     }
+}
+
+/// Whether more of the span is changed than is left unchanged.
+///
+/// The second admission route into [`coalesce_whole_block_inversion`], and the
+/// one that recognises a long inversion (#1461). It is deliberately **not** a
+/// wider separation threshold, because separation cannot decide this case:
+///
+/// | | span | changed | widest gap |
+/// |---|---|---|---|
+/// | `NC_000013.10:g.100809575_100810031inv` (#1461) | 457 nt | 307 (67.2%) | **5** |
+/// | `AAGCTA -> TAGCTT` (the #1040 control) | 6 nt | 2 (33.3%) | **4** |
+///
+/// The inversion that must be recognised has *wider* gaps than the pair that
+/// must stay individual, so every threshold on separation either refuses #1461
+/// or admits the #1040 control. That is why widening
+/// [`every_separation_is_a_single_base`] breaks
+/// `a_whole_span_reverse_complement_is_not_merged_across_a_multi_base_separation`.
+///
+/// Density does separate them, and for a reason rather than by fitting. Reverse
+/// complementing a random block leaves roughly a quarter of its positions
+/// coincidentally unchanged, which is what #1461's 32.8% looks like. The #1040
+/// control leaves 66.7% unchanged — far above chance, because that span is very
+/// nearly its own reverse complement and the two edits are independent changes
+/// sitting inside it, exactly as that test's comment argues. Requiring the
+/// changed columns to be the majority says the reverse-complement relation is
+/// doing work across the whole span instead of describing a near-palindrome.
+///
+/// Additive by construction: this is an `||` alternative to the single-base gate,
+/// so every block that coalesces today still coalesces. Only a block that is
+/// currently *refused* can change, which is what confines the blast radius to
+/// the class #1461 reports.
+fn changed_columns_dominate_the_span(pieces: &[Piece]) -> bool {
+    let (start, end) = (pieces[0].ref_start, pieces[pieces.len() - 1].ref_end);
+    let Some(span) = end.checked_sub(start).filter(|span| *span > 0) else {
+        return false;
+    };
+    let changed: usize = pieces
+        .iter()
+        .map(|piece| piece.ref_end.saturating_sub(piece.ref_start))
+        .sum();
+    changed * 2 > span
 }
 
 /// Whether every gap between consecutive pieces is a single unchanged base —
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
### Review pass: the unit test could not see the regression it names, and the blast-radius framing needs one correction

**1. `density_separates_the_inversion_from_the_over_recognition_control` was ~1.5x denser than the case it claimed to pin.**

Its doc said the fixture was "reconstructed as three members with the same span and changed total" as #1461, but the pieces spanned `0..314`, not `0..457`. With `changed = 307` that is **97.8% density, not the 67.2%** the issue reports — leaving a wide band of threshold regressions it could not detect.

Found by sabotage rather than by reading: raising the gate from `>50%` to `>75%` — a threshold that refuses the very case this PR exists to admit — left the test **passing**. Its `assert!(changed * 2 > span)` precondition also re-implemented the function under test inline, so it could only ever agree with it.

Rebuilt from #1461's measured gap histogram `{1:78, 2:24, 3:2, 4:2, 5:2}` — 108 gaps holding 150 unchanged columns, 109 members, 307 changed, span exactly 457 — and replaced the tautological precondition with `assert_eq!` on member count, span, changed total and widest gap. Re-verified: the same >75% sabotage now **fails**, as it must. (`issue_1040_inv_overrecognition_probes.rs` at 66.7% did catch that case, so the guard was not absent overall — but this test's central claim was false.)

**2. The blast radius is real and the PR body understates it.**

`dump_normalized_corpus` reports **0 of 52,024** moved. That zero is structurally guaranteed and is not evidence: the only family building a whole-span reverse complement large enough to fragment is `long_block_inversion`, and `the_long_cores_are_near_palindromes` pins those cores to exactly **4 changed columns out of 1024/1100** — 0.4% density, where this gate needs >50%. `inv_member` is 3 bases (single piece, returns before the gate) and `delins_hiding_an_inversion`'s span is not a reverse complement by construction. The corpus cannot express this shape at any density.

Measured with a purpose-made enumeration instead — whole-span inversions at sizes that can fragment, three spellings each (`inv`, spanning `delins`, per-column allele):

| | count |
|---|---:|
| rows compared | 327,882 |
| **rows moved** | **65,805 (20.1%)** |
| distinct variants moved | 21,935 of 109,294 |
| → converged on the lone `inv` | **65,805 (100%)** |
| → a third form matching neither side | **0** |
| rows fragmenting further, or newly declining/panicking | **0** |

Direction is the cheap one: every moved row goes from a fragmented multi-member allele to the lone `inv`, which is what ferro already ships for inversions passing the single-base gate and what `inversion.md:5` mandates.

**The correction:** confluence was already perfect on both sides — **0 non-confluent variants before, 0 after**, all three spellings agreeing throughout. So this is a **pure representation migration, not a confluence fix**. The PR body's table (`0` cis rows, `1` ClinVar row) reports real-world *incidence*, which is fair but is not the rate within the shape class being changed — that is 20%, and a consumer holding the fragmented string for any of those 21,935 variants will need to re-bucket.

### Not applied

**Checked subtraction / `changed > span / 2` (CodeRabbit, minor).** `saturating_sub` is already fail-safe here — a malformed reversed piece *reduces* `changed`, which can only make the gate refuse, never wrongly admit. The caller's defence-in-depth block (`merge.rs:3876-3882`) already rejects out-of-order and overlapping pieces before anything is sliced. And `changed * 2` cannot overflow: `changed` is bounded by sequence length, so it would take a 2^63-base contig. The rewrite is arithmetically equivalent for both parities, i.e. a pure style swap on a file several open PRs are touching.

Gate: fmt, clippy `--all-features --all-targets -D warnings` clean, and the full suite under all three oracles with the complete sweep corpus — `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1 FERRO_SWEEP_SEEDS=full` — 9511/9511, plus the no-features `--test it` build the soak archive uses.

PR #1486, opened 2026-08-06
feat(normalize): record provenance when reported members are coalesced

nh13:
## Why

`DNA/delins.md:79-84` gives the spec's own reason for describing nearby variants individually — *"the two variants may have been reported (or might occur) individually"* — and names the concrete risk: **"an overlap with the description of the combined variant might be missed in the annotation step (database queries)"**.

That is **provenance**. It lives in the input's spelling; neither the reference nor the observed bases carry any trace of it. So the normalized string cannot answer it, and it should not try: making the output depend on which spelling arrived is the definition of non-confluence, which is what #1235 exists to remove.

That cost is measured, not assumed. A single input-relative comparand in the normalizer's weight bound — comparing derived pieces against the *input's own* edits — cost **427 converged classes per shuffle direction** over the 11,272-class corpus until the pass in #1470 was moved past it.

**So the spelling-dependence goes where it is harmless.** The canonical string stays a function of the denoted sequence; a diagnostic carries how the variant was reported. Two spellings of one variant still normalize to one string — they simply carry different diagnostics. That is precisely what lets a downstream query recover the individually-reported form, which is the spec's stated worry.

This is the follow-up #1470 names in its "known limit" test: three of the 592 real multi-member cis alleles are genuinely separate variants sitting as close together as a coincidental split, and no property of the sequence distinguishes them. This does not fix those — nothing can, from sequence alone — but it stops them being *silent*.

## What

`W5005 MembersCoalescedFromReportedForm`, emitted when normalization returns fewer cis members than the input described.

**W5xxx** because this is a reference-aware finding about a multi-member allele — the same neighbourhood as `W5002 OverlapConflictingEdits`, not a W3xxx input-form issue.

**Emitted at the single exit of `normalize_core_checked`**, not at any one coalescing site. Several passes can reduce the member count (the live rule's adjacency coalesce, the prioritisation collapse, the sequence-first re-derivation), and the caller cares *that* it happened rather than which pass did it. Comparing counts at the one exit catches all of them and cannot drift as those passes change. It also keeps the diff out of `src/normalize/merge.rs`, which several open branches are currently contending for.

Deliberately not counted:

- a **`trans`/unphased group** — its members sit on different molecules, so they were never coalescing candidates and a change in their number would mean something else entirely;
- an **`uncertain` group** — `(...)` marks a predicted grouping, not a set of separately-reported observations.

## API surface

**This adds a public warning code**, so it is an API change — additive only. `ErrorType::MembersCoalescedFromReportedForm` / `PyErrorType` discriminant 46, surfaced through `normalize_with_diagnostics` and the Python bindings alongside W4004/W5001/W5002. No existing code, discriminant or message changes.

## It moves no normalized string

The property the whole design rests on, verified against the real prepared reference on the default arm:

| corpus | rows | md5 before | md5 after | differing |
|---|---|---|---|---|
| ClinVar 500k | 500,004 | `9a7bc10c…` | `9a7bc10c…` | **0** |
| real multi-member cis | 592 | `683127e5…` | `683127e5…` | **0** |

`the_diagnostic_does_not_change_the_normalized_string` pins it in the suite too, by normalizing the same inputs through both public exits and requiring byte agreement.

## Tests

`tests/it/coalesced_members_diagnostic.rs` — 7 tests. The one worth reading is `equivalent_spellings_converge_while_their_provenance_differs`: two spellings of one variant produce **one string** and **different diagnostics**, which is the entire design in a single assertion. The rest pin that a surviving member count emits nothing, that single-member and `trans` inputs never warn, and that the W-code round-trips through the public registry.

Gate: `cargo fmt --check` and `clippy --all-features --all-targets -D warnings` clean; `cargo check --features python` clean; `9473 tests run: 9473 passed, 145 skipped` with all three oracles and `FERRO_SWEEP_SEEDS=full`; `700 passed` under `pytest tests/python/`.

Refs #1235, #1470

### DIFF
--- a/python/ferro_hgvs/__init__.pyi
+++ b/python/ferro_hgvs/__init__.pyi
@@ -1459,6 +1459,7 @@ class ErrorType(_NativeEnum):
     IntronicOnBareTranscript: ErrorType  # 43
     IncompleteCdsStartReference: ErrorType  # 44
     InsertionWithoutInsertedSequence: ErrorType  # 45
+    MembersCoalescedFromReportedForm: ErrorType  # 46
 
 class ErrorOverride(_NativeEnum):
     """Override behavior for a specific error type."""
--- a/src/error_handling/types.rs
+++ b/src/error_handling/types.rs
@@ -461,6 +461,33 @@ pub enum ErrorType {
     /// projection-side decline (#972).
     IncompleteCdsStartReference,
 
+    /// Normalization returned FEWER cis members than the input described —
+    /// two or more separately-reported variants were coalesced into one.
+    ///
+    /// Purely informational: the normalized string is unchanged by this
+    /// warning's existence, and no mode rejects on it.
+    ///
+    /// It exists because `DNA/delins.md:79-84` gives the spec's reason for
+    /// describing nearby variants individually — *"the two variants may have
+    /// been reported (or might occur) individually"* — and names the concrete
+    /// risk: "an overlap with the description of the combined variant might be
+    /// missed in the annotation step (database queries)". That is
+    /// **provenance**, and it lives in the input's spelling; the reference and
+    /// the observed bases carry no trace of it.
+    ///
+    /// ferro cannot answer it in the canonical string. Making the output depend
+    /// on which spelling arrived is the definition of non-confluence, which is
+    /// what #1235 exists to remove — and the cost is measured, not assumed: a
+    /// single input-relative comparand in the normalizer's weight bound cost 427
+    /// converged classes per shuffle direction over the 11,272-class corpus.
+    ///
+    /// So the spelling-dependence goes where it is harmless. The string stays a
+    /// function of the denoted sequence; this warning carries how the variant
+    /// was reported. Two spellings of one variant still normalize to one string
+    /// — they simply carry different diagnostics, which is exactly what lets a
+    /// downstream query recover the individually-reported form.
+    MembersCoalescedFromReportedForm,
+
     /// An insertion with no inserted sequence — a bare `ins` (e.g.
     /// `c.100_101ins`).
     ///
@@ -530,6 +557,7 @@ impl ErrorType {
             ErrorType::IntronicOnBareTranscript => "W4007",
             ErrorType::IncompleteCdsStartReference => "W5004",
             ErrorType::InsertionWithoutInsertedSequence => "W3027",
+            ErrorType::MembersCoalescedFromReportedForm => "W5005",
         }
     }
 
@@ -587,6 +615,7 @@ impl ErrorType {
         ErrorType::TranscriptFlankNotDescribable,
         ErrorType::IntronicOnBareTranscript,
         ErrorType::IncompleteCdsStartReference,
+        ErrorType::MembersCoalescedFromReportedForm,
     ];
 
     /// Parse a warning code (e.g. `"W3007"`) into its `ErrorType`.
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai pause

nh13:
### Review pass: the warning was dropped on the `p.` axis, and two fixtures named bases the reference does not have

**1. Protein cis alleles coalesce, and the provenance was silently lost there.** `accession_and_axis` returned `None` for `HgvsVariant::Protein`, so the emission block was skipped entirely. Measured on the unmodified branch:

```
NP_000079.2:p.[Ala100Gly;Ala101Gly] -> p.Ala100_Ala101delinsGlyGly   warnings=[]
NP_000079.2:p.[Ala100del;Ala101del] -> p.Ala598_Ala599del            warnings=[]
```

Two reported members become one in both cases, with no W5005. That is the axis where a database query is most likely to be looking for the individually reported form, so it is the worst one to drop it on. Added the `p` arm and a regression test; sabotage-verified — removing the arm again fails it with `got []`.

**2. Two fixtures named the wrong reference base.** `provider` fills the contig with a cycled `ACGT`, so 1-based positions 100 and 300 are both `T` (index 99 and 299, each ≡ 3 mod 4). The controls spelled them `100A>C` and `300G>T`. Harmless in effect — I checked, both spellings normalize identically and emit no warning, so the non-coalescing control still controlled — but a fixture asserting reference bases should name the ones the reference has. Corrected to `100T>C` / `300T>G` in all four occurrences.

### Verified inert, which is the PR's central claim

Checked on a genomic contig where nothing else trips a rejection, since the transcript fixtures hit `REDUCED_CAPABILITY_NO_GENOME` and mask the question:

```
strict    NC_PROBE.1:g.[10del;11del]      => Ok(g.10_11del)      warnings=[MEMBERS_COALESCED_FROM_REPORTED_FORM]
lenient   NC_PROBE.1:g.[10del;11del]      => Ok(g.10_11del)      warnings=[MEMBERS_COALESCED_FROM_REPORTED_FORM]
strict    NC_PROBE.1:g.[10del;20del]      => Ok(g.[11del;21del]) warnings=[]
```

Strict does not reject on it, both modes produce byte-identical strings, and a non-coalescing input correctly produces nothing. So this is purely additive — no representation change.

### Not applied

**Gate the emission on the W5005 error-mode config (CodeRabbit, major).** No `NormalizationWarning` in `src/normalize/mod.rs` is config-gated — there is no `retain`, no `should_ignore`, no `action_for` anywhere in the file, and every sibling (`ReducedCapabilityNoGenome`, `IncompleteCdsStartReference`, `AxisClampApplied`, …) pushes unconditionally. Doing it for this one warning alone would make the new code the odd one out without making `ignore = ["W5005"]` work in any general sense. The real gap — that `NormalizationWarning` and the `ErrorType` config machinery are not connected at all — is pre-existing and worth its own issue rather than a partial fix here.

**Python tests for the new code, and `///` on the `PyErrorType` variant (trivial/minor).** The variant follows its sibling W5004 exactly: registered, exposed to Python, not constructed in the normalize path. Consistent with existing practice.

Gate: fmt, clippy `--all-features --all-targets -D warnings`, 9523/9523, plus the no-features `--test it` build the soak archive uses. Note `pinned_v21_normalization_behavior` failed until I regenerated the spec fixture — that artifact is gitignored and generated from the code under test, so it is a stale-local-artifact detector rather than a regression signal.

PR #1489, opened 2026-08-06
fix(spdi): normalise the reference window before searching a repeat tract

### LINKED ISSUE #1452: Repeat-to-SPDI conversion fails on a soft-masked reference
`hgvs_to_spdi` cannot convert a repeat edit whose reference bases are soft-masked (lowercase), on any spelling — anchored or explicit-range.

## Why

Two comparisons in the repeat arm of `src/spdi/convert.rs` are case-sensitive byte comparisons, and `fetch_reference_bases` does not normalise case:

- the unit-match check, `del_str != unit_str.repeat(pre_count)`, compares reference bases against the unit spelled in the description;
- `normalize::rules::count_tandem_repeats`, which the tract search calls, compares `&ref_seq[end..end + unit_len] == repeat_unit`.

A description is conventionally written with uppercase units, so on a soft-masked tract the first check reports `repeat span <acc>:<s>-<e> does not match repeat unit <unit>` even though the reference genuinely is that repeat.

## Why it is worth a separate issue

This is pre-existing and independent of #1431 — both spellings failed the same way before and after that change, so it is not a regression and was deliberately left out of scope there.

It is worth fixing carefully rather than quickly. Normalising case in **one** of the two places is worse than neither: if the tract search were made case-insensitive while the unit-match check stayed case-sensitive, a soft-masked tract would be found, then rejected; if the unit-match check were relaxed while the search stayed strict, the search would fall back to the unit-wide span and the relaxed check would accept it — turning today's clean error into a silently truncated span denoting the wrong bases.

The repo already treats soft-masking as a real input condition elsewhere (`tests/it/issue_1318_soft_masked_delins.rs`), which is the argument that this path should handle it too.

## Suggested shape

Normalise the fetched window once, at the point of fetch, so every downstream comparison in this arm sees one case convention — mirroring how `canonical_base_byte` centralises the `r.` uracil/thymine equivalence in `normalize::merge`. Cover it with a regression test on a lowercase tract for both the anchored and the explicit-range spelling, asserting they agree with each other and with the uppercase reference.

nh13:
Closes #1452.

## The defect — and a corrected diagnosis

#1452 reports that `hgvs_to_spdi` **cannot convert** a soft-masked repeat, "on any spelling — anchored or explicit-range", naming two case-sensitive comparisons. Measured on `origin/main` at `a530cdaf`, neither half of that holds, and the real behaviour is worse than a refusal.

**The unit-match check is already case-insensitive.** Both of its operands (`del_str`, `unit_str`) go through `apply_alphabet`, which upper-cases. So the explicit-range spelling was never affected, and no failure on this path is a conversion *error*.

**The anchored spelling does not fail — it silently emits a truncated span.** Only the tract search (`normalize::rules::count_tandem_repeats`, called from `resolve_repeat_tract_span`) compares raw reference bytes, and it compares them against a unit that has already been upper-cased. On a soft-masked tract it therefore matches nothing, and the no-run branch falls back — deliberately, so a genuine mismatch is reported by the caller's one diagnostic — to the **unit-wide** span at the anchor. The caller then upper-cases those bases, one unit trivially matches one unit, and the check passes:

| spelling | uppercase `CAGCAGCAG` | soft-masked `cagcagcag` |
|---|---|---|
| `g.259CAG[5]` | `258:CAGCAGCAG:CAGCAGCAGCAGCAG` | `258:CAG:CAGCAGCAGCAGCAG` |
| `g.259_267CAG[5]` | `258:CAGCAGCAG:CAGCAGCAGCAGCAG` | `258:CAGCAGCAG:CAGCAGCAGCAGCAG` |

The masked anchored form applies to **seven** copies, not the five it says: the two tract units the truncated span left untouched survive the replacement.

In other words, soft-masking silently drops the anchored spelling back to the **pre-#1431 reading** — the same "the anchor names one unit" answer that #1431 was filed to remove. It is #1431 reopened by a different door, which is why the issue's "pre-existing and independent of #1431" framing is not quite right either: the truncation-by-failed-search route is specific to the tract search #1431 added.

Invisible to every oracle, for the same reasons #1431 was: both forms parse (`FERRO_ASSERT_REPARSE`), both are in bounds (`FERRO_ASSERT_IN_BOUNDS`), and both are fixed points (`FERRO_ASSERT_IDEMPOTENT`).

### Census

Anchored conversions of `A` / `AT` / `CAG` / `GATC` tracts, 2..=5 copies, every anchor position, two target counts — 560 rows. 336 anchor out of phase with the tract and are refused, identically in both case conventions. Of the 224 that remain, 112 are soft-masked and 112 are not, and the split was total:

| | disagrees with the range spelling | before | after |
|---|---|---|---|
| soft-masked, in phase | 112 | **112 (all)** | 0 |
| uppercase, in phase | 112 | 0 | 0 |
| out of phase (either) | 336 | refused, unchanged | refused, unchanged |

## The fix

Normalise the fetched window **once, at the fetch** — `fetch_normalized_reference_bases` — as the issue's "suggested shape" asks, so the tract search and the unit-match check see the same bytes by construction rather than by each remembering to fold. This mirrors how `normalize::merge::canonical_base_byte` centralises the `r.` uracil/thymine equivalence.

The issue warns that folding one of the two comparisons alone would be worse than folding neither. That warning does not apply, because there are not two sites: the unit-match check already folds. There is one blind site, and folding it is exactly what makes the two spellings agree.

Normalising with `apply_alphabet` rather than a bare `to_ascii_uppercase` is deliberate and does one thing more than the issue asked for: on the `r.` axis the unit has had `U` rewritten to `T`, so an uppercase-only window would still miss a `U`-spelled tract and truncate it by the identical route. `a_uracil_spelled_tract_is_found_on_the_rna_axis` pins both directions of that.

Scope: one file, `src/spdi/convert.rs`.

## Unchanged on purpose

Two discriminating tests, because the obvious wider fix — relax the unit-match check instead of the search — passes every happy-path assertion above:

- `a_non_repeat_span_is_still_refused_when_soft_masked` — `ATGCAT` is divisible by `AT` and is not `AT[3]`; it must still be refused, in either case convention. A relaxed unit check would accept it, since the fallback span is one unit wide and one unit always matches itself.
- `an_out_of_phase_anchor_is_refused_in_both_case_conventions` — `g.260CAG[5]` names the second base of the run, where no copy begins. It was refused before and is refused now, with a byte-identical diagnostic on masked and unmasked references. This keeps a later widening of the search from quietly inventing a tract at an anchor the spec's start-only format does not name.

## Representation stability

**No normalized HGVS string can move.** The change is confined to `src/spdi/convert.rs`, and nothing in normalization's canonical path reaches it: `src/normalize/merge.rs` names `hgvs_to_spdi` only in doc comments (lines 5418, 8894), never calls it.

Three leaf surfaces do reach it, and all three move:

| surface | route | pre-fix symptom on an anchored repeat over a soft-masked tract |
|---|---|---|
| `Normalizer::to_spdi` (`src/normalize/mod.rs:1746`), public `hgvs_to_spdi`, the Python binding | direct | a truncated SPDI triple |
| `Normalizer::apply_to_reference` (`src/normalize/mod.rs:1705`) and its Python binding (`src/python.rs:1468`) | `spdi::apply::variant_edit_triples` (`src/spdi/apply.rs:457`) | **the wrong applied nucleotide sequence** — the one-unit deletion span left the tract at seven copies while the description said five |
| `EquivalenceChecker` | `src/equivalence/checker.rs:386` | an anchored-vs-range comparison over a masked tract could return the wrong equivalence verdict |

`apply_to_reference` is the loudest of the three: it returns bases, not a triple, so the truncation surfaced as a corrupted sequence rather than as a coordinate a reader might sanity-check. All three move in the correcting direction — the characterisation below applies unchanged to each.

`examples/dump_normalized_corpus.rs` was **not** run, because it is structurally blind to this change and a `0 moved` from it would be evidence of the corpus, not of the diff: it measures normalized HGVS output, which this cannot touch, and its providers serve uppercase sequences, so it cannot build the input condition either. Per the "a corpus zero is a claim about the corpus" note in the layout `CLAUDE.md`, reporting that zero would be the dishonest form.

What *does* move is SPDI output (and, through `apply_to_reference`, the applied sequence derived from it). Characterised exactly:

- **only** anchored (`g.<pos><unit>[n]`) repeats, **only** where the reference tract is soft-masked or `U`-spelled;
- every such row was **already wrong** — it denoted a different sequence than the range spelling of the same variant — so this is a correctness fix with no stable-form casualties, not a tie broken against a shipped representation;
- **0 uppercase rows move**, which is the whole shipped surface for any consumer whose reference is not soft-masked.

The real evidence, in place of the corpus: the full suite green with **nothing re-blessed**. For an output-moving change that is the strong signal — every existing expectation already agreed with the new answer.

## Verification

```
#   Summary [1888.020s] 9531 tests run: 9531 passed (16 slow), 145 skipped
```

All three oracles, and the full sweep corpus rather than the 4-seed default prefix. No Python surface is touched, so the `python-lint` / `python-wheel-test` / `abi3-floor` additions do not apply.

### DIFF
--- a/src/spdi/convert.rs
+++ b/src/spdi/convert.rs
@@ -1576,14 +1576,20 @@ where
             // nothing" — so widening here makes the two sites agree rather
             // than introducing a third answer.
             let (del_start, del_end) = match provider {
-                Some(p) if start_one_based == end_one_based => {
-                    resolve_repeat_tract_span(p, &sequence, start_one_based, unit_str.as_bytes())?
-                }
+                Some(p) if start_one_based == end_one_based => resolve_repeat_tract_span(
+                    p,
+                    &sequence,
+                    start_one_based,
+                    unit_str.as_bytes(),
+                    alphabet,
+                )?,
                 _ => (start_one_based, end_one_based),
             };
             let spdi_pos = del_start - 1;
-            let del_raw = match provider {
-                Some(p) => fetch_reference_bases(p, &sequence, del_start, del_end)?,
+            let del_str = match provider {
+                Some(p) => {
+                    fetch_normalized_reference_bases(p, &sequence, del_start, del_end, alphabet)?
+                }
                 None => {
                     return Err(unspelled_bases_error(
                         UnspelledBases::RepeatTract,
@@ -1592,7 +1598,6 @@ where
                     ));
                 }
             };
-            let del_str = apply_alphabet(&del_raw, alphabet);
             // The HGVS recommendations require the interval to span an
             // integer number of repeat units. Reject non-divisible spans
             // with a clear message rather than silently emitting nonsense.
@@ -1758,11 +1763,29 @@ fn unspelled_bases_error(
 /// unit does not match the reference there — so the caller's existing
 /// "does not match repeat unit" diagnostic is what reports it, rather than a
 /// second error for one condition.
+///
+/// **The window is normalised before it is searched** (#1452). `unit` arrives
+/// having already been through [`apply_alphabet`], so searching the raw window
+/// would compare a soft-masked (lowercase) reference against an uppercase unit
+/// and find no run at all. That did not surface as an error: the no-run
+/// fallback above returns the *unit-wide* span, whose bases the caller then
+/// uppercases, so the unit-match check passed and a truncated triple was
+/// emitted. On a 3-copy lowercase `cag` tract, `g.259CAG[5]` came out as
+/// `258:CAG:CAGCAGCAGCAGCAG` — seven copies once the two untouched tract
+/// units are counted — while the range spelling `g.259_267CAG[5]` correctly
+/// gave `258:CAGCAGCAG:CAGCAGCAGCAGCAG`. Both spellings converted; they just
+/// denoted different sequences.
+///
+/// Normalising with [`apply_alphabet`] rather than a bare `to_ascii_uppercase`
+/// is what keeps the two comparisons provably identical: on the `r.` axis the
+/// unit has had `U` rewritten to `T`, so an uppercase-only window would still
+/// miss a `U`-spelled tract and truncate it by the same route.
 fn resolve_repeat_tract_span<P>(
     provider: &P,
     accession: &str,
     anchor_one_based: u64,
     unit: &[u8],
+    alphabet: AlphabetMode,
 ) -> Result<(u64, u64), ConversionError>
 where
     P: ReferenceProvider + ?Sized,
@@ -1811,7 +1834,13 @@ where
         let window_end = anchor_one_based
             .saturating_add(half_width)
             .min(sequence_length);
-        let window = fetch_reference_bases(provider, accession, window_start, window_end)?;
+        let window = fetch_normalized_reference_bases(
+            provider,
+            accession,
+            window_start,
+            window_end,
+            alphabet,
+        )?;
         let bytes = window.as_bytes();
         let anchor_offset = (anchor_one_based - window_start) as usize;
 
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1490, opened 2026-08-06
fix(reference): serve LRG genomic bases from the LRG record, not the placement

### LINKED ISSUE #1462: Reference: LRG_ genomic coordinates resolve to the wrong base (constant index offset vs the LRG record)
For at least one `LRG_` accession, a `g.` coordinate resolves to the wrong base — the served sequence is the `LRG_<N>g` record shifted by a constant offset, not the record itself.

## Reproduce

Probe a position by normalizing a single-base `delins`, which reports the reference base:

```
LRG_292:g.160875delinsA   ->   LRG_292:g.160875T>A
```

So ferro serves `T` at `LRG_292:g.160875`. The `LRG_292g` record in the prepared reference has `G` there:

```
samtools faidx LRG_292.fasta LRG_292g:160873-160879   ->   GTGGTCA     (160875 = G)
```

## It is an index offset, not sequence divergence

Probing 20 consecutive positions, `160875..160894`, and comparing against the FASTA at several offsets:

```
ferro-served  160875..160894:  TCAATGGAAGAAACCACCAA
FASTA at offset +0:            GGTCAATGGAAGAAACCACC
FASTA at offset +2:            TCAATGGAAGAAACCACCAA   <== exact match
```

The sequences are identical; only the indexing differs. So this is an off-by-N in how `LRG_` genomic coordinates are placed, not a case of the reference having drifted from the frozen LRG record.

## Why it matters

LRG genomic sequences are frozen reference standards — that is their entire purpose. An `LRG_<N>:g.<p>` coordinate is defined relative to the LRG record, so serving a shifted base means every `LRG_` genomic variant silently resolves to the wrong position. It is silent because the result is still well-formed and in-bounds; no oracle catches it.

## Scope — verified vs reported

- **Verified here:** `LRG_292`, offset `+2`, over a 20-base window.
- **Reported but not verified by me:** `LRG_311` at `-1`, `LRG_839` at `+23`, and 19 other `LRG_` accessions at offset `0`. If that holds, the offset is per-accession rather than global, which points at the placement/alignment step rather than a single constant.

Worth checking whether the affected accessions are those resolved through a genomic placement rather than read directly from the LRG record.

nh13:
Closes #1462.

## The defect

`MultiFastaProvider::get_sequence` and `get_sequence_length` gate the chromosome-synthesis path on `self.prepared.contains(id)`. That is `false` for **every** bare `LRG_<N>`, because the FASTA record is indexed under LRG's own suffix convention as `LRG_<N>g` (`resolve_name` has mapped it since #487, but the guard runs *before* `resolve_name`). So every LRG genomic accession was served by reconstructing its bases from the chromosome through the single-affine placement parsed out of the LRG XML — never from its own frozen record.

The issue is right about the symptom and slightly off about the mechanism: nothing is placed wrongly. The affine is as good as an affine can be; it simply cannot express the `mismatch` / `lrg_ins` / `other_ins` diffs that the LRG XML itself lists alongside the mapping. `LRG_292` has 3 LRG-inserted bases and 1 chromosome-inserted base before the probed position, which is exactly the reported `+2`.

That also means this is not really a per-accession offset — a shift is only what it looks like when you probe one window past an indel. The general case is "the chromosome's sequence, not the LRG's", which also substitutes bases at every `mismatch` diff with no shift at all.

Measured over the 1294 LRG records in the prepared reference, comparing each `LRG_<N>` against its own `LRG_<N>g` record across the whole record:

| | before | after |
|---|---|---|
| records serving ≥1 wrong base | **164** | 0 |
| total wrong bases served | **2,017,282** | 0 |
| records reporting the wrong length | **44** | 0 |
| records panicking on the `parent_start == 1` debug assert | **1** | 0 |

The 165th record carries a placement with `parent_start == 1049` and tripped `synthesize_parent_sequence`'s debug assertion outright — whose doc comment asserted that "an LRG/RefSeqGene placement with `parent_start > 1` is never routed here". That sentence was false; the comment is corrected in this PR.

A 20-base windowed probe gives the offset histogram `{-8: 1, -5: 1, -1: 12, 0: 1269, +1: 3, +2: 2, +5: 2, +6: 2, +23: 1}`. That confirms all three accessions the issue reported second-hand and flagged as unverified: `LRG_292` at `+2`, `LRG_311` at `-1`, and a `+23`.

## The fix

Adds `own_record()` — the substitution-free half of `resolve_name`. It accepts the accession itself, plus the one record rename LRG's file format imposes (`LRG_<N>` → `LRG_<N>g`), and deliberately **not** `resolve_name`'s version-strip and `chr`-alias fallbacks, which do serve a genuinely different sequence than the one asked for. Both synthesis guards now ask `own_record(id).is_none()` instead of `!prepared.contains(id)`.

Scoped there rather than at `resolve_name` because the guard's ordering is load-bearing: it runs *before* `resolve_name` on purpose, so that a #728 derived `NG_` whose exact version is absent is reconstructed rather than silently served a sibling version's bases. `own_record` preserves that ordering while fixing the one case where "not in the index under this exact name" did not mean "we do not ship this accession's bases". `resolve_name` now delegates its first two clauses to it, so the two cannot drift apart.

The only behavioural change is for a bare `LRG_<N>` that has an `LRG_<N>g` record. Nothing else can reach the new clause: `NG_` accessions are versioned and never match `is_bare_lrg_genomic_accession`.

## Unchanged on purpose

- **An LRG with a placement but no FASTA record of its own is still synthesized from the chromosome.** That is the #728 derived-parent path the guard exists for, and removing it is the obvious over-fix. The regression test pins it explicitly (`LRG_998`).
- **`LRG_<N>t<k>` / `LRG_<N>p<k>`** still resolve directly and are never remapped to the genomic record — `is_bare_lrg_genomic_accession` already rejected them, and the existing #487 test still covers it.
- **The affine placement itself is untouched.** `VariantProjector::project_to_genomic` still uses it to re-express transcript coordinates in an LRG parent's frame, and remains approximate for the same 164 accessions. That is a distinct defect in coordinate *projection* rather than base *serving*; this PR does not widen scope into it.

## Representation stability

**This PR moves normalized output for LRG genomic variants, and the moves are migrations.** Measured release-to-release on a real corpus, not just against `main`: 3,882 deterministic `g.` variants (3 per LRG record with length ≥ 1000 — a `delinsA`, a `del` and a `dup` at fixed fractions of each record), normalized against the prepared reference by the pre-fix binary and the post-fix binary.

- **rows compared: 3,882 · rows moved: 42 (1.1%) · accessions affected: 31 of 1,294**
- **All 42 were previously accepted** (`ok` → `ok`), so every one is a stored string that changes. None were previously rejected. This is the expensive direction, and it is stated plainly.
- Shapes moved: 17 `delins`, 17 `dup`, 8 `del`.
- **Every move is a correction, not a tie broken the other way.** The before-strings named the chromosome's base or position; the after-strings name the LRG record's. Examples:

  ```
  LRG_62:g.10696delinsA   before: g.10696T>A     after: g.10696C>A   (wrong ref base)
  LRG_20:g.34999del       before: g.34999del     after: g.35003del   (wrong shift target)
  LRG_42:g.5625delinsA    before: g.5625G>A      after: g.5625=      (record's base IS A)
  ```

  The last is the sharpest: the LRG record already has `A` at 5625, so `delinsA` is an identity — reported as a substitution only because the chromosome has `G` there.

- **Spec position:** this is not a case where either form is admissible. An `LRG_<N>:g.<p>` coordinate is defined against the LRG record, which is a frozen reference standard; the previously-shipped strings denoted a different sequence. The stability rule says to break a genuine tie toward the shipped form — there is no tie here, so the shipped form loses.

**The committed synthetic harness measures `0 of 78,028`, and that zero is structural.** Run properly — dumped at the base (`a530cdaf`) in a detached worktree, dumped at this branch, and diffed by `--compare`:

```
| total                                            | 78028        |
| **moved**                                        | **0 (0.0%)** |
| of which previously **accepted** (a migration)   | 0            |
| of which previously **not** a fixed point (free) | 0            |
```

covering all 15 shape families (`adjacent_junction_ins`, `coincident_bounds`, `del_plus_sub`, `delins_hiding_an_inversion`, `dup_plus_ins`, `dup_plus_sub`, `identity_member`, `inv_member`, `junction_interior_to_span`, `long_block_inversion`, `nested_spans`, `partial_overlap_spans`, `repeat_expansion`, `split_vs_spanning_delins`, `three_member_allele`).

**Do not read that as evidence of safety.** Per `CLAUDE.md`, a corpus zero is a claim about the corpus. This one is guaranteed *a priori*: `dump_normalized_corpus.rs` constructs only `MockProvider`, and its own doc comment puts prepared references explicitly out of scope. This diff touches only `MultiFastaProvider`, so no edit to this code could move a row here. The harness is blind to the property under change — not through a missing shape family, but by driving a different provider type entirely.

The real-reference corpus above is the measurement that counts, and it is the one that found the 42 moved rows.

**Confluence guards:** unaffected — no partitioning, shifting or merge logic is touched, and the full suite passes with nothing re-blessed.

## Verification

```
  -> Summary [1221.735s] 9524 tests run: 9524 passed (13 slow), 145 skipped
```

**Nothing was re-blessed.** For an output-moving change that is the strong signal: every existing expectation already agreed with the new answer, because no committed test pinned a synthesized-LRG base.

The new unit test is discriminating — verified by reverting only the two guard sites and re-running, which fails with `left: "AAAA", right: "CGCG"`. Its fixture supplies a placement on purpose, so the test cannot pass merely because there is nothing to synthesize from, and it mirrors the real `LRG_292` shape (a 12-base record placed on a 10-base chromosome window, i.e. a 2-base LRG insertion the affine cannot represent) with disjoint alphabets so a single served base names which path ran.

Reference-backed census numbers were measured with a throwaway probe against the prepared reference (`FERRO_MANIFEST`); the probe is not part of the diff, and the durable guard is the unit test.

### DIFF
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -1366,25 +1366,48 @@ impl MultiFastaProvider {
         })
     }
 
-    /// Resolve a sequence name, trying various forms
-    fn resolve_name(&self, name: &str) -> Option<String> {
-        // Direct lookup
+    /// The indexed FASTA record holding `name`'s **own** bases, or `None` when
+    /// the reference ships no record for that exact accession.
+    ///
+    /// This is the *substitution-free* half of [`resolve_name`](Self::resolve_name):
+    /// it accepts `name` itself, plus the one rename LRG's own file format
+    /// imposes — a bare `LRG_<N>` genomic accession is stored under the record
+    /// name `LRG_<N>g` (LRG's suffix convention; transcripts are `LRG_<N>t<k>`,
+    /// proteins `LRG_<N>p<k>`). That is a spelling of the same sequence, not a
+    /// fallback to a *different* one: LRG accessions carry no version, so
+    /// `LRG_<N>` denotes `LRG_<N>g` and nothing else. It deliberately omits
+    /// `resolve_name`'s version-strip and `chr`-alias fallbacks, which do serve
+    /// a different sequence than the one asked for.
+    ///
+    /// Callers use it to ask "do we ship this accession's own bases?" before
+    /// falling back to reconstructing them from somewhere else — see
+    /// `get_sequence` / `get_sequence_length`, where a bare `prepared.contains`
+    /// answered "no" for every `LRG_<N>` and so handed all 1294 LRG records to
+    /// chromosome synthesis instead of to their own frozen sequence — silently
+    /// wrong for the 165 whose placement is not an exact affine (#1462).
+    fn own_record(&self, name: &str) -> Option<String> {
         if self.prepared.contains(name) {
             return Some(name.to_string());
         }
-
-        // LRG genomic accession: HGVS `g.` variants reference the genomic
-        // sequence by its bare accession `LRG_<N>`, but the FASTA record is
-        // indexed under the LRG suffix convention `LRG_<N>g` (transcripts are
-        // `LRG_<N>t<k>`, proteins `LRG_<N>p<k>`, both of which match the index
-        // directly above). Map the bare genomic accession to its `g`-suffixed
-        // record so `LRG_<N>:g.` normalization can fetch reference bases
-        // instead of silently no-op'ing (issue #487).
         if is_bare_lrg_genomic_accession(name) {
             let genomic = format!("{name}g");
             if self.prepared.contains(&genomic) {
                 return Some(genomic);
             }
+        }
+        None
+    }
+
+    /// Resolve a sequence name, trying various forms
+    fn resolve_name(&self, name: &str) -> Option<String> {
+        // Direct lookup, plus the LRG bare-genomic → `LRG_<N>g` record rename
+        // (#487): HGVS `g.` variants name the genomic sequence by its bare
+        // accession `LRG_<N>`, but the FASTA indexes it under LRG's suffix
+        // convention. Both are "this accession's own bases", so they live
+        // together in `own_record`; everything below this point substitutes a
+        // *different* sequence and must stay after it.
+        if let Some(own) = self.own_record(name) {
+            return Some(own);
         }
 
         // Try chromosome alias FIRST (before version fallback)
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

Issue #1034, opened 2026-07-15
ferro v0.8.0 incorrectly fragments inversions found within a longer delins


msto:
@nh13 One regression cropped up when reviewing the 0.8.0 results, see below

---

> **Authorship:** the content below was drafted by GitHub Copilot CLI (an AI coding agent) and filed via `gh` under @msto's account. It reflects the agent's analysis, not a statement authored by @msto.
>
> # Regression (v0.8.0): ferro over-recognizes `inv`, splitting reverse-complement sub-runs out of a contiguous `delins`
> 
> ## Summary
> 
> Starting in **v0.8.0**, ferro fragments a contiguous multi-nucleotide change to expose an internal reverse-complement sub-run as an `inv`, even when the change as a whole is **not** an inversion. Per HGVS, an inversion must describe a *maximal contiguous run* whose alt equals the reverse complement of the reference; a reverse-complement **sub-run may not be carved out** of a longer contiguous change (that change must stay a single `delins`).
> 
> This is a **regression from v0.7.0**, which emitted the spec-compliant single `delins`. The two outputs round-trip to the identical sequence — this is purely a typing/partitioning defect.
> 
> ## Environment
> 
> - ferro-hgvs **v0.8.0** (regressed) vs **v0.7.0** (correct)
> - Coordinate system: genomic (`g.`), no reading frame
> - Reference-aware normalization
> 
> ## Minimal example
> 
> Reference excerpt (1-based): position `100=C`, `101=T`, `102=G` (i.e. ref window `100_102 = CTG`).
> 
> | | HGVS |
> |---|---|
> | raw | `REF:g.[100C>A;101T>C;102G>A]` |
> | expected (spec-correct) | `REF:g.100_102delinsACA` |
> | **ferro v0.8.0 (wrong)** | `REF:g.[100C>A;101_102inv]` |
> | ferro v0.7.0 (correct) | `REF:g.100_102delinsACA` |
> 
> **Mechanics (verified):**
> - Window `100_102`: ref `CTG` → alt `ACA` (length-preserving).
> - `RC(CTG) = CAG ≠ ACA` ⇒ the whole run is **not** a reverse complement ⇒ it is a `delins`, not an `inv`.
> - The 2-nt sub-run `101_102` (`TG` → `CA`) *does* satisfy `RC(TG) = CA`, so v0.8.0 carves it out as `101_102inv` and appends `100C>A`.
[...]

PR #1493, opened 2026-08-06
fix(normalize): type a split-created span inv on the first pass

### LINKED ISSUE #1454: A derived member is typed inv only on the second pass, so the first output is not a fixed point

nh13:
Closes #1454.

## The defect

A single-member `delins` normalized to a two-member allele whose second member was spelled `delins` when it was an exact inversion. Normalizing again re-typed it, so the first output was not a fixed point:

```
NM_TEST.1:c.10_15delinsTAATAT
```

`c.13_15` is `ATA` and `revcomp(ATA)` is `TAT`. `general.md:56` ranks inversion above deletion-insertion, and `delins.md:5` defines a delins as a replacement "which is not a substitution or inversion" — so `delins` over a reverse-complement span is not a description the spec admits at all. Pass 2 was right and pass 1 was wrong.

Only `FERRO_ASSERT_IDEMPOTENT` catches it; the output re-parses and every coordinate is in range.

## The mechanism is not where the issue says it is

The issue attributes this to the alternation loop in `merge::canonicalize_from_sequence`. **That loop never runs for this input.** `Normalizer::sequence_first_pass` admits a lone member only on `g.`/`m.` (`is_splittable_single_member`), so a lone `c.` delins declines on its first line — instrumenting the loop against the reproducer logs nothing. `merge.rs` is untouched by this PR.

The split comes from `apply_canonical_split`:

1. `rules::decompose_delins` reports `[Sub@10; IdentityAt@11; Sub@12; Sub@13; Sub@14; Sub@15]`. It types an inversion only when a **whole** maximal contiguous mismatch run is a reverse complement, and deliberately does not carve a reverse-complement *sub*-run out of a longer contiguous change (#1034, #1040). Positions 12-15 are one run, `TATA -> ATAT`, which is not one.
2. `build_split_variants` applies the codon-frame exception (`general.md:35-38`) and **consumes** `[Sub@10; IdentityAt@11; Sub@12]` into `10_12delinsTAA`.
3. That leaves `Sub@13,14,15` — a span that did not exist when step 1 typed the input, and which *is* a whole reverse complement. It fell to `flush_substitution_run`, which emitted `Delins` unconditionally.

The re-grouping creates spans; nothing re-typed them.

## The fix, and why it is scoped there

Both of `build_split_variants`' multi-base groupings now go through `rules::canonicalize_delins` — the same function the *next* pass reaches via `normalize_na_edit`'s Delins arm. Answering the question the second pass would ask, with the function it would ask it of, is what makes the two passes unable to disagree, rather than an open-coded reverse-complement test that could drift from it.

Only `Inversion` can come back for these inputs, which is what makes the narrow match sound rather than lossy: both groupings hand over equal-length slices whose first and last positions are mismatches, so `Identity`, `Duplication`, `Insertion`/`Deletion` and `Substitution` are all unreachable. Anything else is a contract change, and falling through to `delins` is the safe reading of it.

## Unchanged on purpose

- **A reverse-complement sub-run of one contiguous change stays a `delins`** — the #1034/#1040 control. `c.12_15delinsCTAT` is unchanged, even though `13_15` inside it is a reverse complement. The fix types spans the split *creates*; it must not create one.
- **Axes without a reading frame are untouched.** Without the codon exception nothing consumes part of a run, so no new span is ever created and `decompose_delins`' verdict stands. `n.10_15delinsTAATAT` splits to `[10A>T;12_15delinsATAT]` on both coding and non-coding records, before and after. Every row of the census below that moved is on the `c.` axis.

## Representation stability

`examples/dump_normalized_corpus.rs`, `a530cdaf` -> this branch:

| | rows |
|---|---|
| total | 78,028 |
| **moved** | **52 (0.1%)** |

All 52 are in `delins_hiding_an_inversion` — the family added for this issue, so the corpus is not structurally blind here — and all are one shape, a trailing `delins` member becoming `inv`:

```
before: NM_TEST.1:c.[1_3delinsGGG;4_6delinsCAT]
after : NM_TEST.1:c.[1_3delinsGGG;4_6inv]
```

**Read the direction, not the count.** Every moved row is an output `main` itself rewrote on a second pass, so this converges on the string ferro already produced rather than inventing a third one — the cheap direction of a confluence fix. It is nonetheless a moved string: a consumer that normalizes once and stores the result holds the old `delins` form. The harness's "0 previously accepted" column is a statement about *inputs* and does not say otherwise.

## Census

479,682 rows — every `delins` of length 3, 4 and 5 at every position of three 63/64 nt cores, on the `c.` and `n.` axes, counting `normalize(normalize(x)) != normalize(x)`:

| | `main` (`a530cdaf`) | with the fix |
|---|---|---|
| outputs that are not fixed points | **267** | **0** |

All 267 are the one shape, spanning 188 distinct payload pairs and both the 2 nt and 3 nt inversion widths.

## Re-blessed pins, and a prediction that did not hold

`cis_confluence_axis`'s two per-oracle censuses are now byte-identical, so `expected_census` and both `*_UNDER_IDEMPOTENCY_ORACLE` constants are deleted, as that module said to do when #1454 was fixed.

That module also predicted `converged` would rise by **two**. It rises by **one** per direction (3': 6632 -> 6633; 5': 6627 -> 6628). Measured per class rather than inferred from the net figure — a `+1` is also what "converged three, broke two" looks like — **exactly one class changes convergence status and none regresses**:

```
s04-c-m2-sep3-p4-rot4   the pure #1454 class
  before  [13_15delinsTAA;16_19delinsTAAT] | [13_15delinsTAA;16_19inv]
  after   [13_15delinsTAA;16_19inv]                            CONVERGED

s00-c-m2-sep5-p1-rot4   the issue's own reproducer
  before  [10_12delinsTAA;13_15delinsTAT]  | [9dup;15del]
  after   [10_12delinsTAA;13_15inv]        | [9dup;15del]      still splits
```

The second class had the #1454 defect too and the fix corrects it — that member is a fixed point now. The class still splits because its competing output is a different **partition** of the same variant rather than a different typing, and no inversion typing merges `[9dup;15del]` with a two-member delins. That residual belongs to the #1235 / #1419-#1421 partition-non-uniqueness family and is deliberately left alone: choosing between those two partitions moves shipped representations and needs its own measurement.

So the `+2` was never reachable by fixing #1454. It was an artifact — under the oracle, that class's delins spelling *panicked*, leaving `[9dup;15del]` as the only surviving outcome, and a class with one outcome scores as `converged`. The oracle was crediting the defect with a convergence that was really a suppressed output.

## Verification

```
cargo fmt --all --check                                         pass
cargo clippy --all-features --all-targets -- -D warnings         pass
  cargo nextest run --features dev                               9529 tests
```

The full-gate run was green except the two `cis_confluence_axis` census pins re-blessed here; both now pass in both oracle configurations.

Six new tests in `tests/it/issue_1454_split_member_inversion_typing.rs`, each asserting the fixed-point property directly rather than relying on `FERRO_ASSERT_IDEMPOTENT` — CI's gating `Test` job runs without the oracles, so an oracle-only guard would be green in the one job that blocks a merge. Two of them pin branches that would otherwise be untested: the `r.` axis reaches the same `inv` (which is what makes the `U`/`T` fold load-bearing), and the codon-frame triplet is itself reachable as an inversion over a self-complementary centre.

**The codon-triplet test needs an input longer than the triplet, and the first draft of it did not have one.** A lone `c.10_12delinsGNT` is a whole-span reverse complement, so `normalize_na_edit`'s Delins arm types it `inv` before any split happens — it exercises the direct path and never enters `build_split_variants`. Instrumenting the codon branch confirmed it: **0 hits** for that spelling against **11** for `a_span_the_codon_exception_leaves_behind_is_typed_inv_on_the_first_pass`. It also survived a sabotage (forcing `push_typed_replacement` to always emit `delins`) that correctly failed the other three inversion tests, which is what surfaced it.

The test now uses `c.10_15delinsGNTAAA` -> `c.[10_12inv;13_15delinsAAA]`. The trailing `GGG -> AAA` run is what forces the split, so the codon exception actually consumes `[Sub@10; IdentityAt@11; Sub@12]` and hands `ANC -> GNT` to `push_typed_replacement` as a span the split created. Verified discriminating: under the same sabotage it now fails with `10_12delinsGNT`. The three-base spelling is kept alongside it as an explicit control that the two routes to `inv` agree.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -10882,8 +10882,8 @@ fn build_split_variants(
             if let (
                 DelinsSubedit::Substitution {
                     position: p1,
+                    reference: r1,
                     alternative: a1,
-                    ..
                 },
                 DelinsSubedit::IdentityAt {
                     position: pm,
@@ -10891,8 +10891,8 @@ fn build_split_variants(
                 },
                 DelinsSubedit::Substitution {
                     position: p3,
+                    reference: r3,
                     alternative: a3,
-                    ..
                 },
             ) = (&subedits[i], &subedits[i + 1], &subedits[i + 2])
             {
@@ -10916,18 +10916,17 @@ fn build_split_variants(
                         flush_substitution_run(&mut output, template, hgvs_start, &mut run);
                         let s = abs(*p1);
                         let e = abs(*p3);
+                        // Typed rather than assumed to be a `delins`, for the
+                        // reason `push_typed_replacement` gives. This branch
+                        // reaches an inversion only when the untouched middle
+                        // base is its own complement (`W`, `S`, `N`) — every
+                        // other base makes `alt[1] == bm != complement(bm)` and
+                        // rules the whole span out — which is why the reference
+                        // in `codon_triplet_over_an_ambiguous_centre_is_an_inversion`
+                        // carries an `N`.
+                        let ref_bases = [*r1, *bm, *r3];
                         let alt_bases = vec![*a1, *bm, *a3];
-                        output.push(build_variant_at(
-                            template,
-                            s,
-                            e,
-                            NaEdit::Delins {
-                                sequence: InsertedSequence::Literal(Sequence::new(alt_bases)),
-                                deleted: None,
-                                deleted_length: None,
-                                substitution_reference: None,
-                            },
-                        ));
+                        push_typed_replacement(&mut output, template, s, e, &ref_bases, alt_bases);
                         i += 3;
                         continue;
                     }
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

PR #1495, opened 2026-08-06
fix(normalize): fold case when counting a tandem repeat tract

### LINKED ISSUE #1491: Normalizer's canonical form for a repeat depends on the reference's soft-masking
`normalize::rules::count_tandem_repeats` compares reference bytes with raw slice equality, so it does not find a tandem run on a soft-masked (lowercase) reference. Three normalizer call sites feed it raw reference bytes, and the result is that **the canonical form of a repeat depends on whether the reference happens to be soft-masked**.

## Measured

`Normalizer::normalize` over three tracts (`CAG[3]`, `A[5]`, `AT[4]`), each against a byte-identical uppercase and lowercase reference. 4 of 15 inputs give a different canonical form:

| input | uppercase reference | soft-masked reference |
|---|---|---|
| `g.259_267CAG[2]` | `g.265_267del` | `g.259_267CAG[2]` |
| `g.259CAG[5]` | `g.259_267CAG[5]` | `g.259CAG[5]` |
| `g.259A[7]` | `g.259_263A[7]` | `g.259A[7]` |
| `g.259AT[6]` | `g.259_266AT[6]` | `g.259AT[6]` |

The remaining 11 (plain `del`/`dup`/`ins`, and explicit-range repeats whose count is not a contraction) agree, which is the control showing this is specific to the tract search rather than a general masking problem.

Two distinct shapes:

- **A contraction fails to lower.** `g.259_267CAG[2]` over a 3-copy tract is a deletion of one copy, and normalizes to `g.265_267del` on an uppercase reference. On a soft-masked one it stays an unlowered repeat.
- **A single-position anchor is not widened to its tract.** `g.259CAG[5]` / `g.259A[7]` / `g.259AT[6]` all widen to the whole-range form on an uppercase reference and stay as the bare anchor on a soft-masked one.

Reference FASTAs are routinely soft-masked, and masked bytes demonstrably reach normalization in this codebase — `tests/it/issue_1318_soft_masked_delins.rs` and `issue_1235_cis_allele_confluence.rs::soft_masked_reference_yields_the_same_canonical_form` both exist for that reason.

## Why

`count_tandem_repeats` (`src/normalize/rules.rs:1943` and `:1954`) uses `==` on byte slices:

```rust
let candidate = &ref_seq[start - unit_len..start];
if candidate == repeat_unit { ... }
...
if &ref_seq[end..end + unit_len] == repeat_unit { ... }
```

The unit comes from the caller's description, conventionally uppercase; the bases come from the reference.

**It is the odd one out in its own file.** `src/normalize/rules.rs` has been case-folded across its other comparison sites over several prior issues — dup detection (`:125`, `:132`), delins trimming (`:1459`, `:1487`), inversion typing (`:1606`, `:1620`, `:1717`, `:1750`) all use `eq_ignore_ascii_case` — and the file already carries soft-masking parity tests that build `upper` from `lower` (`:4692`, `:4725`, `:4748`). The repeat-counting routine was not included.

Non-test callers passing raw reference bytes: `rules.rs:1867`, `rules.rs:2144`/`:2164`, `merge.rs:2918`.

## Why no oracle catches it

Every masked output above is well-formed, in bounds, and a fixed point, so `FERRO_ASSERT_REPARSE`, `FERRO_ASSERT_IN_BOUNDS` and `FERRO_ASSERT_IDEMPOTENT` all pass. The property violated is *parity between two references that differ only in case*, which nothing asserts. This is the same blind spot #1318 and #1250 were each filed against for a different comparison site.

## Suggested resolution, and its stability cost

Fold the two comparisons in `count_tandem_repeats` with `eq_ignore_ascii_case`, matching the convention the rest of the file already follows.

The stability argument is unusually clean, and worth stating because a confluence fix normally has to pick a loser:

- `eq_ignore_ascii_case` on two uppercase slices is `==`, so **behaviour on an unmasked reference is provably unchanged** — no shipped form moves for any consumer whose reference is not soft-masked.
- `count_tandem_repeats` returns only coordinates (`count`, `start`, `end`), never bases, so folding its comparison cannot leak lowercase into an emitted sequence.
- Rows that do move are soft-masked ones, and they move **onto the form the uppercase reference already produces** — converging on the already-shipped representation rather than away from it.

That should still be measured with `examples/dump_normalized_corpus.rs` before merging, with the caveat that the corpus's providers serve uppercase sequences, so it may be structurally blind here — in which case say so rather than quoting the zero.

## Reproducer

Measured on the `fix/issue-1452-softmask-repeat-spdi` branch of #1489, whose only change is in `src/spdi/convert.rs` and cannot affect `Normalizer::normalize`; equivalently `origin/main` at `a530cdaf`.

Related: #1452 / #1489 fixed the same class at the SPDI call site by normalising the fetched window, and deliberately left this shared routine alone because folding it can move normalized output.

nh13:
Closes #1491.

## The defect

The canonical form of a repeat depended on whether the reference happened to be soft-masked. `normalize::rules::count_tandem_repeats` compared reference bytes to the description's unit with raw slice equality, so on a lowercase tract it found no run at all — and because the function returns *coordinates* rather than a verdict, that did not surface as an error. It surfaced as a different answer:

| input | uppercase reference | soft-masked reference |
|---|---|---|
| `g.259_267CAG[2]` | `g.265_267del` | `g.259_267CAG[2]` |
| `g.259CAG[5]` | `g.259_267CAG[5]` | `g.259CAG[5]` |
| `g.259A[7]` | `g.259_263A[7]` | `g.259A[7]` |
| `g.259AT[6]` | `g.259_266AT[6]` | `g.259AT[6]` |

Two shapes. A **contraction fails to lower** — `CAG[2]` over a 3-copy tract is a deletion of one copy and becomes `g.265_267del` on an unmasked reference, but stays an unlowered repeat on the masked twin. And a **single-position anchor is never widened to its tract**, leaving the bare anchor standing where the whole-range form is canonical.

Measured **4 of 15** repeat inputs moved with case alone. The other 11 — plain `del`/`dup`/`ins`, and explicit-range repeats whose count is not a contraction — agreed, which is the control showing this is the tract search rather than a general masking problem.

**No oracle could see it.** Every masked output above is well-formed, in bounds and a fixed point, so `FERRO_ASSERT_REPARSE`, `FERRO_ASSERT_IN_BOUNDS` and `FERRO_ASSERT_IDEMPOTENT` all pass. The violated property is parity between two references differing only in case, which nothing asserts — the same blind spot #1318 and #1250 were each filed against for a different comparison site.

Reference FASTAs are routinely soft-masked and masked bytes reach normalization from ordinary input; `tests/it/issue_1318_soft_masked_delins.rs` and `issue_1235_cis_allele_confluence.rs::soft_masked_reference_yields_the_same_canonical_form` both exist for that reason.

## The fix

Fold both scans in `count_tandem_repeats` with `eq_ignore_ascii_case`.

This is the convention the rest of `src/normalize/rules.rs` already follows: dup detection (`:125`, `:132`), delins trimming (`:1459`, `:1487`) and inversion typing (`:1606`, `:1620`, `:1717`, `:1750`) all fold, and the file already carries soft-masking parity tests that build `upper` from `lower` (`:4692`, `:4725`, `:4748`). The repeat finder was the odd one out. Three non-test callers feed it raw reference bytes: `rules.rs:1867`, `rules.rs:2144`/`:2164`, `merge.rs:2918`.

Sibling to #1489, which fixed the same class at the SPDI call site by normalising the fetched window and deliberately left this shared routine alone — precisely because folding it can move normalized output, which is a stability question rather than a bugfix. This PR is that question, answered.

### Measured against #1489, because the overlap reads as redundancy and is not

`resolve_repeat_tract_span` calls this function, so with both PRs applied the SPDI path folds twice and #1489's window normalisation *looks* superseded. Tested by stacking this branch on #1489's head in a throwaway worktree and reverting only #1489's window normalisation on top:

| | result |
|---|---|
| all 11 tests from both PRs, stacked | **pass** — no red-together interaction |
| #1489's six soft-masking tests, with #1489 reverted | **pass** — this fold covers the case dimension |
| `a_uracil_spelled_tract_is_found_on_the_rna_axis`, same | **fails** (`left: (3, 3)`, `right: (3, 6)`) |

`eq_ignore_ascii_case` maps `u` to `U`, never `U` to `T`, so an `r.` tract spelled with uracil stays invisible to this function; #1489's `apply_alphabet` is what handles it. The two are complementary, and after both land **exactly one test pins the difference**. A note on `count_tandem_repeats` says so, so neither site gets removed on the grounds that the other covers it.

The only textual conflict between the two is the `mod` line in `tests/it/main.rs`, which is a union.

## Representation stability

**This moves normalized output**, so the disclosure matters. It is an unusually clean case, because a confluence fix normally has to pick a loser and this one does not:

- `eq_ignore_ascii_case` on two uppercase slices **is** `==`. Behaviour on an unmasked reference is therefore provably unchanged — no shipped form moves for any consumer whose reference is not soft-masked.
- `count_tandem_repeats` returns only `(count, start, end)` — never bases — so folding its comparison cannot leak a lowercase base into an emitted sequence.
- Rows that *do* move are soft-masked ones, and they move **onto the form the uppercase reference already produces**. The fix converges on the already-shipped representation rather than away from it, which is the free direction.

### The corpus measured zero, and the zero is structural

```
cargo run --features dev --example dump_normalized_corpus -- --out before.tsv   # baseline: rules.rs at origin/main
cargo run --features dev --example dump_normalized_corpus -- --out after.tsv    # candidate
cargo run --features dev --example dump_normalized_corpus -- --compare before.tsv --against after.tsv
```

| | rows |
|---|---|
| total | 78,028 |
| **moved** | **0 (0.0%)** |
| of which previously not a fixed point (free) | 0 |

Zero across all 15 shape families, and the two dumps are byte-identical.

**Do not read that as evidence the change is safe.** It is a claim about the corpus: every sequence the generator builds comes from uppercase `ACGT`/`AT` literals (`examples/dump_normalized_corpus.rs:387`, `:596`, `:912`, `:930`, `:935`), so the harness cannot vary reference masking at all and is structurally blind to a case-fold. This is the same class of blindness recorded for #1456 (member geometry), #1460 (scale) and #1478 (transcript geometry); reporting the zero together with the reason is the honest form, following PR #1480's precedent.

**The real evidence is the suite: 9,526 tests passing with nothing re-blessed**, under all three oracles and `FERRO_SWEEP_SEEDS=full`. For an output-moving change that is the strong signal — every existing expectation already agreed with the new answer.

## Unchanged on purpose

- `the_shapes_that_already_agreed_are_untouched` — the 11 inputs that already agreed across case must keep agreeing. This is what says the fix is the tract search and not a blanket fold that happened to move everything.
- `folding_does_not_invent_a_tract` — the discriminating case. An over-general fix (comparing lengths only, or treating any base as matching) would widen a lone `A` into a run that is not there; `g.259A[3]` over `ACGTA` must stay `g.259A[3]`. That value is **measured, not inferred** — an earlier draft guessed `g.259_260insAA` and was wrong.

No `*_strategy()` was touched, so no proptest seed is invalidated (`git status --porcelain tests/proptest-regressions/` clean).

## Verification

```
#   Summary [1067.452s] 9526 tests run: 9526 passed (13 slow), 145 skipped
```

No Python surface is touched.

### DIFF
--- a/src/normalize/rules.rs
+++ b/src/normalize/rules.rs
@@ -1919,6 +1919,39 @@ pub fn duplication_to_repeat(
 /// boundary-anchor behavior, so delegating here would silently drop valid
 /// tracts. The genuinely shared piece across the ins→repeat / repeat paths is
 /// the 3'-phase step `three_prime_align_tract` (#852), not the tract finder.
+///
+/// # Case
+///
+/// Both scans compare with [`slice::eq_ignore_ascii_case`], because `ref_seq`
+/// comes from the reference and `repeat_unit` comes from the description. A
+/// reference FASTA is routinely soft-masked and a description is conventionally
+/// written in upper case, so a raw byte comparison finds no run at all on a
+/// masked tract — and, since this returns coordinates rather than a verdict,
+/// that surfaces as a *different canonical form* rather than as an error
+/// (#1491). Four of fifteen repeat inputs measured changed answer with case
+/// alone: `g.259_267CAG[2]` lowered to `g.265_267del` on an upper-case
+/// reference and stayed an unlowered repeat on the soft-masked twin, and three
+/// single-position anchors widened to their tract on one and not the other.
+///
+/// This is the convention the rest of this module already follows — the dup,
+/// delins-trimming and inversion-typing comparisons all fold — and folding
+/// costs nothing on an unmasked reference, where `eq_ignore_ascii_case` and
+/// `==` agree by definition. Note also that only *coordinates* leave this
+/// function, so folding here cannot put a lower-case base into an emitted
+/// sequence.
+///
+/// **Case is all this folds — it is not alphabet normalization.**
+/// `eq_ignore_ascii_case` maps `u` to `U`, never `U` to `T`, so an `r.` tract
+/// whose reference is spelled with uracil is still invisible here. The SPDI
+/// call site handles that separately by running the fetched window through
+/// `apply_alphabet` before it ever calls this function (#1452,
+/// `spdi::convert::fetch_normalized_reference_bases`), which folds case *and*
+/// rewrites `U` to `T`. The two are complementary rather than redundant, and
+/// the overlap is easy to misread as duplication: with this fold in place,
+/// reverting the SPDI-side normalization leaves every soft-masking test green
+/// and fails exactly one — `a_uracil_spelled_tract_is_found_on_the_rna_axis`.
+/// That single test is what pins the difference, so do not "simplify" either
+/// site away on the grounds that the other covers it.
 pub fn count_tandem_repeats(
     ref_seq: &[u8],
     pos: usize,
@@ -1940,7 +1973,7 @@ pub fn count_tandem_repeats(
     let mut start = pos;
     while start >= unit_len {
         let candidate = &ref_seq[start - unit_len..start];
-        if candidate == repeat_unit {
+        if candidate.eq_ignore_ascii_case(repeat_unit) {
             start -= unit_len;
         } else {
             break;
@@ -1951,7 +1984,7 @@ pub fn count_tandem_repeats(
     let mut end = start;
     let mut count = 0u64;
     while end + unit_len <= ref_seq.len() {
-        if &ref_seq[end..end + unit_len] == repeat_unit {
+        if ref_seq[end..end + unit_len].eq_ignore_ascii_case(repeat_unit) {
             count += 1;
             end += unit_len;
         } else {
--- a/tests/it/issue_1491_soft_masked_repeat_normalization.rs
+++ b/tests/it/issue_1491_soft_masked_repeat_normalization.rs
@@ -1,0 +1,227 @@
+//! #1491 — the canonical form of a repeat must not depend on whether the
+//! reference happens to be soft-masked.
+//!
+//! `normalize::rules::count_tandem_repeats` compared reference bytes to the
+//! description's unit with raw slice equality. A reference FASTA is routinely
+//! soft-masked and a description is conventionally written upper-case, so on a
+//! masked tract the search found no run — and because the function returns
+//! *coordinates* rather than a verdict, that did not surface as an error. It
+//! surfaced as a different canonical form:
+//!
+//! | input | uppercase reference | soft-masked reference |
+//! |---|---|---|
+//! | `g.259_267CAG[2]` | `g.265_267del` | `g.259_267CAG[2]` |
+//! | `g.259CAG[5]` | `g.259_267CAG[5]` | `g.259CAG[5]` |
+//! | `g.259A[7]` | `g.259_263A[7]` | `g.259A[7]` |
+//! | `g.259AT[6]` | `g.259_266AT[6]` | `g.259AT[6]` |
+//!
+//! Two shapes. A **contraction fails to lower**: `CAG[2]` over a 3-copy tract
+//! is a deletion of one copy and becomes `g.265_267del` on an unmasked
+//! reference, but stayed an unlowered repeat on the masked twin. And a
+//! **single-position anchor is not widened to its tract**, which leaves the
+//! bare anchor standing where the whole-range form is the canonical answer.
+//!
+//! Measured 4 of 15 repeat inputs moved with case alone; the other 11 — plain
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

PR #1496, opened 2026-08-06
fix(spdi): name the caller's anchor when a repeat names no tract

### LINKED ISSUE #1492: Repeat-to-SPDI decline names a span the caller never wrote when the anchor is a single position
When a single-position repeat anchor names no tandem run, `hgvs_to_spdi` declines with a message quoting the *unit-wide fallback span* rather than the position the caller actually wrote, and says nothing about what is really wrong.

## Measured

On `origin/main` at `a530cdaf`, against a `CAG[3]` tract at 259-267:

```
g.260CAG[5]   -> invalid position: repeat span TEMPLATE:260-262 does not match repeat unit CAG
g.261CAG[5]   -> invalid position: repeat span TEMPLATE:261-263 does not match repeat unit CAG
```

The caller wrote one position, `260`. The span `260-262` appears nowhere in the input: it is the unit-wide fallback `resolve_repeat_tract_span` returns when the tract search finds nothing. So the message invites the reader to go look at a three-base window they never named, and gives no hint that the real problem is that `260` is out of phase with the run — the tract begins at `259`, and `g.259CAG[5]` converts fine.

The explicit-range spelling is **correct today** and must stay that way, because there the span is the caller's own:

```
g.259_264AT[3]  -> repeat span TEMPLATE:259-264 does not match repeat unit AT     (correct — user wrote 259_264)
```

## Scale

Not a corner. In the census run for #1452 — anchored conversions of `A`/`AT`/`CAG`/`GATC` tracts, 2..=5 copies, every anchor position, two target counts — **336 of 560 rows** take this path. It is every anchor that does not land on a unit boundary, and it behaves identically on masked and unmasked references.

## Why no oracle catches it

It is a diagnostic-quality defect, not a wrong answer: the conversion correctly declines. Nothing asserts that an error message names only intervals the caller supplied.

## Suggested resolution

Have `resolve_repeat_tract_span` tell the caller whether the span it returned is a *found tract* or the *no-run fallback*, and emit a distinct decline for the fallback case — naming the anchor, saying no run of the unit begins there, and pointing at the explicit-range spelling as the way out. `resolve_repeat_tract_span` already has a precedent for that shape in its `MAX_REPEAT_SEARCH_BASES` message, which suggests `g.<start>_<end><unit>[n]`.

Roughly:

```
no CAG repeat begins at TEMPLATE:260; the run there starts at 259.
Spell the whole range explicitly (`g.259_267CAG[n]`) or anchor at the run's first base.
```

Naming the actual run start is optional and only possible when a run exists nearby; declining to guess is acceptable.

**No behaviour change** — the same inputs are accepted and the same inputs are refused; only the message text moves. Worth confirming against `an_out_of_phase_anchor_is_refused_in_both_case_conventions` (added in #1489), which pins the current text and would need updating in the same PR.

nh13:
Closes #1492.

**Stacked on #1489** (`fix/issue-1452-softmask-repeat-spdi`) — it touches the same region of the repeat arm and updates a test #1489 adds. Review or merge that one first.

## The defect

When a single-position repeat anchor names no tandem run, the decline quoted a span the caller never wrote:

```
g.260CAG[5]  ->  invalid position: repeat span TEMPLATE:260-262 does not match repeat unit CAG
```

The caller wrote one position, `260`. The range `260-262` is the unit-wide fallback `resolve_repeat_tract_span` returns when its search finds nothing — an internal artefact. So the message sends the reader to inspect a three-base window that appears nowhere in their input, and says nothing about the actual fault: the run begins at `259`, and `g.259CAG[5]` converts fine.

Not a corner. **336 of the 560 rows** in #1452's census take this path — it is every anchor that does not land on a unit boundary — and it behaves identically on masked and unmasked references.

## The fix

Thread a `RepeatSpanOrigin` out of `resolve_repeat_tract_span` so the caller knows *why* the span it holds is the span it holds, and emit an anchor-shaped decline only for the fallback case:

```
no CAG repeat is anchored at TEMPLATE:260: the reference there is not a run of CAG.
A single-position anchor names the *first base* of the tract, so check the run's start,
or spell the whole range explicitly (`g.<start>_<end>CAG[n]`).
```

Three origins rather than two, which is the part worth reviewing:

- `Tract` — a run was found, or the caller wrote an explicit range. A unit-match failure is a real mismatch over coordinates that are legitimately quotable.
- `NoRunAtAnchor` — the search ran and found nothing. The only case that must not quote its span.
- `NotSearched` — an empty unit, or a provider that serves bases but not `get_sequence_length`. The span is a placeholder, not a judgement, so the generic message stays: nothing was measured that would justify asserting "no run begins here". Collapsing this into `NoRunAtAnchor` would make ferro claim a fact it never checked, and it is invisible in-repo because every one of ferro's own providers implements `get_sequence_length`.

## Unchanged on purpose

**No behaviour change** — the same inputs convert and the same inputs are refused. Only text moves. Three tests hold the line:

- `an_explicit_range_decline_still_quotes_the_callers_own_span` — the discriminating case. When the caller wrote `g.259_264AT[3]`, quoting `259-264` is correct and helpful, and that message is untouched. An unconditional reword would have broken this.
- `an_in_phase_anchor_still_converts` — without it, a "fix" that simply refused every single-position anchor would satisfy both other tests.
- `an_out_of_phase_anchor_is_refused_in_both_case_conventions` (from #1489) is updated, not deleted: it still asserts refused-and-identical-across-case, and now also asserts the message does *not* contain `260-262`.

## Representation stability

No normalized HGVS string and no SPDI triple can move: the diff only changes an error's text and the internal signature that decides which text to use. The only observable change is the wording of a `ConversionError::InvalidPosition` on inputs that were already being refused.

## Verification

```
cargo fmt --all --check
cargo clippy --all-features --all-targets -- -D warnings
#   Summary [511.724s] 9534 tests run: 9534 passed (4 slow), 145 skipped
```

No Python surface is touched.

## Note on a re-blessed assertion

`issue_1431::an_anchor_with_no_matching_tract_still_reports_the_mismatch` is updated, not deleted. Its doc states the contract as *"one condition, one decline — not a second error introduced by the tract search"*, and that is unchanged: this is still a single `InvalidPosition`. What it actually asserted was the literal phrase `does not match repeat unit`, which this PR rewords for the anchor case. It now asserts against the caller's own anchor, so it pins the contract rather than the sentence.

## Review follow-ups applied

**The new diagnostic was unreachable on one of its two paths.** The divisibility check runs before the unit-match check, and near the 3' end the unit-wide fallback is clamped inside the contig — so a multi-base unit leaves a short span that fails divisibility *first* and reports the generic message. Measured on a 10-base contig:

```
T2:g.9CAG[5]  ->  repeat span T2:9-10 length 2 is not a multiple of unit length 3
```

`9-10` is the clamped fallback, not anything the caller wrote — the exact defect this PR removes from the unit-match branch, reached through the other one. Both checks are now judged before either is reported, so `span_origin` decides the message and the checks only decide *whether* there is one. `a_failed_search_clamped_at_the_contig_end_still_names_the_anchor` pins it.

**Two doc comments were attached to the wrong items.** Inserting `enum RepeatSpanOrigin` between `resolve_repeat_tract_span`'s doc block and the function moved ~50 lines of that function's documentation onto the enum, leaving the function undocumented. Rust accepts this and neither clippy nor rustdoc warns, because every item involved is private.

Tracing it back, the root cause predates this PR: on `origin/main` there is already no blank line between `fetch_reference_bases`' doc block and `resolve_repeat_tract_span`'s, so `fetch_reference_bases` (about 130 lines below) has been undocumented and its doc silently prepended to its neighbour's. This PR inherited that and extended the run onto the enum. Both halves are fixed here — the enum is lifted out of the split and `fetch_reference_bases`' doc is reattached — so all three items now document themselves.

**Filed #1497** for a related pre-existing hole this PR names but does not change: when `get_sequence_length` fails, `RepeatSpanOrigin::NotSearched` returns the unit-wide span and the conversion *succeeds* on it, because a one-unit span satisfies both the divisibility and unit-match checks by construction. That is the #1452 silent-truncation class via a different door, and the existing comment's claim that "the caller's divisibility and unit-match checks still judge it" does not hold for it.

### DIFF
--- a/src/spdi/convert.rs
+++ b/src/spdi/convert.rs
@@ -1575,7 +1575,7 @@ where
             // start and means 'the whole run becomes N', so it absorbs
             // nothing" — so widening here makes the two sites agree rather
             // than introducing a third answer.
-            let (del_start, del_end) = match provider {
+            let (del_start, del_end, span_origin) = match provider {
                 Some(p) if start_one_based == end_one_based => resolve_repeat_tract_span(
                     p,
                     &sequence,
@@ -1583,7 +1583,9 @@ where
                     unit_str.as_bytes(),
                     alphabet,
                 )?,
-                _ => (start_one_based, end_one_based),
+                // An explicit range is the caller's own span, so a decline may
+                // quote it verbatim.
+                _ => (start_one_based, end_one_based, RepeatSpanOrigin::Tract),
             };
             let spdi_pos = del_start - 1;
             let del_str = match provider {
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

PR #1502, opened 2026-08-06
fix(reference): length a record-less placed parent from coordinate one

### LINKED ISSUE #1494: synthesize_parent_sequence silently serves wrong bases for a record-less parent with parent_start > 1
The `parent_start == 1` assumption in `MultiFastaProvider::synthesize_parent_sequence` (`src/reference/multi_fasta.rs`) is held only by a `debug_assert_eq!`, which is compiled out in release. For a parent that has a placement but no FASTA record of its own, a placement with `parent_start > 1` can therefore serve the wrong bases with no error.

## Why it does not always fail loud

The function maps a 0-based window offset `start` to parent coordinate `start + 1`, which is only correct when `parent_start == 1`. `GenomicPlacement::parent_to_nc` (`src/reference/provider.rs:97-107`) declines only when the requested position is **below** `parent_start`:

```rust
let parent_end = self.parent_start + (self.nc_end - self.nc_start);
if parent_pos < self.parent_start || parent_pos > parent_end {
    return None;
}
```

So for a placement with `parent_start == 1049`:

- a window at offset `0` asks for parent coordinate `1`, which is `< 1049` → `None` → `InvalidCoordinates`. Loud, correct.
- a window at offset `2000` asks for parent coordinate `2001`, which **is** inside `[1049, parent_end]` → it resolves to `nc_start + (2001 - 1049)`. But the parent's own base at 0-based offset `2000` is parent coordinate `1049 + 2000 = 3049`. Wrong window, no error.

The existing comment says the assert guards against a placement that "would otherwise read the wrong window (or fail loud via `parent_to_nc` returning `None`)". Only the second half is guaranteed; the first is the silent case.

`get_sequence_length` has the matching gap: for a record-less parent it returns `placement.nc_end - placement.nc_start + 1`, the placed span length, which is not the parent's length when `parent_start > 1`.

## Reachability

After #1490 the callers gate on `own_record`, so any accession shipping its own record is served from it. What still reaches this function is a parent with a placement and **no** record:

- the FASTA-less #728 derived `NG_`, which is constructed with `parent_start == 1` (`derived_placement.rs`) and so is safe by construction;
- a **record-less LRG**, which carries whatever its XML states. #1490's own test exercises this path (`LRG_998`), with `parent_start == 1`; nothing constrains it to that.

This is pre-existing and #1490 strictly reduces exposure — it removed all 165 LRG-with-record accessions from this path, including the one real placement with `parent_start == 1049` that was tripping the debug assert outright. Filing separately rather than widening that PR.

## Suggested shape

Make it fail closed rather than rely on a debug-only assert:

- in `synthesize_parent_sequence`, return `InvalidCoordinates` for `parent_start != 1` instead of asserting — or, better, offset the window by `parent_start` so the mapping is correct for any value;
- make `get_sequence_length` reject (or correctly account for) the same placement rather than returning the placed-span length;
- add a regression test with a record-less parent at `parent_start > 1` covering both a below-span window (already loud) and an in-span window (currently silent).

Whichever route, the durable guard should hold in release, not only in debug.

nh13:
Closes #1494.

**Stacked on #1490** (base is `fix/issue-1462-lrg-genomic-offset`, not `main`) — the reachability argument depends on its `own_record` gate, and the two touch the same function. Review #1490 first; this PR's own diff is the second commit.

## Read this first: half of #1494's stated mechanism does not hold

#1494 says a record-less parent with `parent_start > 1` reads the wrong window because "the parent's own base at 0-based offset `2000` is parent coordinate `1049 + 2000 = 3049`". The `debug_assert`'s message says the same thing — "offset the window by parent_start before mapping".

**That is backwards, and implementing it introduces the very defect it describes.** I built that fix first, measured it, and threw it away.

An offset indexes the parent's **own coordinate frame**: offset `k` is parent coordinate `k + 1`, which is what the code already did. `parent_start` is a coordinate *in that same frame*, not the origin of the window — a placement with `parent_start > 1` simply does not cover the parent's first `parent_start - 1` bases, and `parent_to_nc` correctly declines them.

**Ground truth is `LRG_1075`**, the single placement in the prepared reference with `parent_start > 1` — the one that trips the assertion:

```xml
<mapping coord_system="GRCh38.p13" other_id="NC_000004.12" type="main_assembly"
         other_start="190173122" other_end="190187909">
  <mapping_span lrg_start="1049" lrg_end="15267"
                other_start="190173122" other_end="190187909" strand="1" />
```

Two independent checks against the record itself:

1. **`lrg_end` is the record's last base.** `LRG_1075g` is **15267** bases (`LRG_1075.fasta.fai`); the span's `lrg_end` is **15267**. Were `lrg_start` an origin rather than a coordinate, the span would end at its width, 14219.

2. **The bases line up at the coordinate, not the offset.**
   ```
   LRG_1075g   coord 1049..1078 : TCTGGTCTTCTACGTGGAAATGAACGAGAG
   NC_000004.12 190173122..     : TCTGGTCTTCTACGTGGAAATGAACGAGAG   <- identical
   LRG_1075g   coord 1..30      : CCCGCCGCCTGACCGCCCCCTCCCCACCCC   <- unrelated
   ```

There is also a self-consistency argument needing no data: `get_sequence`'s FASTA branch reads record coordinates `start + 1 ..= end`. One method's two branches cannot index differently, or every caller straddling them is wrong.

So the window mapping is **deliberately unchanged**.

## What is actually wrong

`get_sequence_length` returned `nc_end - nc_start + 1` — the width of the placed span. Since coordinates count from the parent's base 1, a placement starting at `parent_start` runs to `parent_start + (nc_end - nc_start)`; the span width silently drops the `parent_start - 1` unplaced bases beneath it.

A parent placed at `parent_start == 5` over ten chromosome bases reported length **10** when its coordinates run to **14**. Every bounds check keyed on that length — including the `FERRO_ASSERT_IN_BOUNDS` oracle — then rejects legitimate placed coordinates as past-the-end.

Fixed to `parent_start + (nc_end - nc_start)`. It is a lower bound, not necessarily the record's true length (the parent may extend past the placed span), which is the best a placement can say and is the bound callers need.

## The assertion is removed, not relaxed

`debug_assert_eq!(placement.parent_start, 1)` was over-strict: it fires on a legitimate record-less LRG whose XML places it past coordinate 1, and it guarded a mapping that was already correct. Both new tests failed on it before they could reach their real assertions. Its instruction to "offset the window by parent_start" is the wrong fix above, so leaving it in place would keep inviting it.

## Unchanged on purpose

- **`parent_start == 1` is byte-identical.** `parent_start + (nc_end - nc_start)` equals `nc_end - nc_start + 1` exactly when `parent_start == 1`, which is every #728 derived `NG_` (`derived_placement.rs` constructs them that way). No derived-`NG_` behaviour moves.
- **Coordinates below `parent_start` still fail loud.** They are unplaced; fabricating them would be worse than declining. Pinned by a test.
- **The window mapping**, per the whole first section.

## Tests

Both new, both failed before this change (on the assertion, before reaching their assertions):

- `a_record_less_parent_window_is_offset_from_coordinate_one_not_parent_start` — a characterization test pinning `offset k == coordinate k + 1` with **exact strings**. The fixture is built so the two readings disagree (`GGTT` vs `TTAA`), so it **fails under the fix #1494 proposes**. That is its purpose: it exists to stop the next person implementing the assertion's message.
- `a_record_less_parent_length_counts_from_coordinate_one` — the length, plus a read of the last placed base at that length.

## Representation stability

**No moved output expected, and none observed.** The changed length applies only to a parent that has a placement and *no* record of its own with `parent_start > 1`. After #1490 that set is empty in the prepared reference: `LRG_1075`, the only `parent_start > 1` placement, ships its own record and so is served from it. Every derived `NG_` has `parent_start == 1` and is unchanged arithmetically.

The direction, where reachable, is rejected → accepted: coordinates previously judged past-the-end become readable. That is the free direction.

## Verification

```
cargo fmt --all --check                                                  # clean
  -E 'test(record_less_parent) + test(lrg_with_its_own_record) + test(synthesize)'   # 27/27
```

The full three-oracle gate, as CI runs it:

```
  -> Summary [668.710s] 9526 tests run: 9526 passed (3 slow), 145 skipped
```

**Nothing was re-blessed**, which is the signal that matters here: no committed expectation depended on the old span-width length or on the removed assertion.

GitHub Actions was in a major outage while this PR was opened, so a red or cancelled job on it is more likely the outage than the change — please re-run CI before merging rather than reading the first result.

### DIFF
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -1815,41 +1815,43 @@ impl MultiFastaProvider {
         if start == end {
             return Ok(String::new());
         }
-        // The `[start, end)` window is 0-based into the parent *sequence*, so
-        // offset 0 is the parent's first base — parent coordinate
-        // `parent_start`. The mapping below assumes `parent_start == 1` (it maps
-        // the 0-based offset `start` to 1-based parent coordinate `start + 1`).
-        // Reachable callers are parents that have a placement but no record of
-        // their own: the FASTA-less #728 derived `NG_`, which is constructed
-        // with `parent_start == 1` (`derived_placement.rs`), and a record-less
-        // LRG, which carries whatever its XML states (`LRG_998`, in the test
-        // below, is exactly that shape). So `parent_start == 1` is an
-        // assumption about the placement data, not an invariant this function
-        // can enforce.
-        //
-        // Until #1462 the set was far wider: every bare `LRG_<N>` was routed
-        // here, because the callers gated on `prepared.contains(id)`, which is
-        // `false` for `LRG_<N>` (the record is named `LRG_<N>g`). One real LRG
-        // placement carries `parent_start == 1049` and tripped this very
-        // assertion; a further 164 passed it and quietly served chromosome
-        // bases in place of the frozen LRG record. The callers now gate on
-        // `own_record`, so an LRG *with* a record never reaches here.
-        //
-        // Guard the assumption in debug builds. Note this does not make the
-        // release build safe: `parent_to_nc` declines only a position *below*
-        // `parent_start`, so a window low enough fails loud while a deeper one
-        // silently maps to the wrong bases. Hardening it to fail closed in
-        // release is #1494; this PR narrows who can reach it rather than
-        // changing what it does.
-        debug_assert_eq!(
-            placement.parent_start, 1,
-            "synthesize_parent_sequence assumes parent_start == 1; offset the window \
-             by parent_start before mapping to support placement-only parents"
-        );
         // 0-based half-open [start, end) → 1-based inclusive parent positions
-        // [start+1, end]. Map both endpoints onto the chromosome; on a minus-
-        // strand placement the parent runs antiparallel, so the low parent
-        // position maps to the *higher* `NC_` coordinate.
+        // [start+1, end]. An offset indexes the parent's **own coordinate
+        // frame**, so offset 0 is parent coordinate 1 — never `parent_start`.
+        //
+        // That distinction is what #1494 got backwards, and it is worth being
+        // explicit about because the opposite reading has an obvious-looking
+        // fix that silently serves the wrong bases. `parent_start` is a
+        // coordinate in the parent's numbering, not the origin of the window:
+        // a placement with `parent_start > 1` simply does not cover the
+        // parent's first `parent_start - 1` bases, and `parent_to_nc` declines
+        // them. Ground truth is `LRG_1075`, the one real placement in the
+        // prepared reference with `parent_start > 1` — span
+        // `lrg_start=1049 lrg_end=15267` onto `NC_000004.12:190173122-190187909`,
+        // record `LRG_1075g` exactly 15267 bases (i.e. `lrg_end`), and record
+        // coordinate 1049 matching `NC_000004.12:190173122` base for base.
+        // It also has to hold for this method to be coherent: `get_sequence`'s
+        // FASTA branch reads record coordinates `start + 1 ..= end`, and one
+        // method's two branches cannot index differently.
+        //
+        // A `debug_assert_eq!(placement.parent_start, 1)` used to sit here.
+        // It was over-strict — it fired on a legitimate record-less LRG whose
+        // XML places it past coordinate 1 — and it guarded a mapping that was
+        // already correct. Removed rather than relaxed; the two tests named
+        // `a_record_less_parent_*` pin the semantics instead.
+        //
+        // Reachable callers are parents with a placement and no record of their
+        // own: the FASTA-less #728 derived `NG_` (always `parent_start == 1`,
+        // per `derived_placement.rs`) and a record-less LRG, which carries
+        // whatever its XML states. Until #1462 the set was far wider — every
+        // bare `LRG_<N>` came through here, because the callers gated on
+        // `prepared.contains(id)`, which is `false` for `LRG_<N>` (the record is
+        // named `LRG_<N>g`); 164 of them were quietly served chromosome bases in
+        // place of their frozen record. The callers now gate on `own_record`.
+        //
+        // Map both endpoints onto the chromosome; on a minus-strand placement
+        // the parent runs antiparallel, so the low parent position maps to the
+        // *higher* `NC_` coordinate.
         let (nc_a, nc_b) = match (
             placement.parent_to_nc(start + 1),
             placement.parent_to_nc(end),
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
## The frame argument is right; the length formula uses the wrong width

First, conceding the part that concerns me: **#1494's `synthesize_parent_sequence` claim was wrong and this PR is right about it.** An offset does index the parent's own coordinate frame, offset 0 is parent coordinate 1, and a placement with `parent_start > 1` simply does not cover the bases beneath it — `parent_to_nc` declining them is correct behaviour, not a silent-corruption hazard. I've corrected the issue.

The `get_sequence_length` half of #1494 does describe a real bug, and this is the right place to fix it. But the replacement formula derives a **parent-frame extent from a chromosome-frame width**, and those differ exactly when the alignment carries indels — which is the case for the record this PR cites as ground truth.

### Measured on `LRG_1075`

`lrg/LRG_1075.xml`, GRCh38 span:

```xml
<mapping_span lrg_start="1049" lrg_end="15267" other_start="190173122" other_end="190187909" strand="1">
```

| quantity | value |
|---|---|
| LRG span width (`1049..15267`) | 14,219 |
| NC span width (`190173122..190187909`) | 14,788 |
| `LRG_1075g` record length | **15,267** (= `lrg_end`) |
| `parent_start + (nc_end - nc_start)` | **15,836** |

The two span widths differ by 569, so the formula overshoots the record's true length by 569. The quantity that *does* reproduce it is `lrg_end` — equivalently `parent_start + (lrg_end - lrg_start)` = 1049 + 14218 = 15267.

This is the same limitation #1490 already names: the placement is a pure affine and "cannot express the `mismatch` / `lrg_ins` / `other_ins` diffs the LRG XML lists alongside the mapping." The length formula inherits that assumption, because `GenomicPlacement` has no `parent_end` — the parser reads `lrg_end` at `multi_fasta.rs:2982` and discards it, and `parent_to_nc` reconstructs the parent's end as `parent_start + (nc_end - nc_start)`.

### Why this is latent rather than live

`LRG_1075` **has** its own record, so after #1490 `own_record` short-circuits and it never reaches this branch. The formula only bites a parent that is record-*less* **and** indel-bearing. Every #728 derived `NG_` has `parent_start == 1` and an exact affine, so nothing in that population is affected today.

So this is not a merge blocker in my reading — but the comment should not cite `LRG_1075` as validating the formula, because the arithmetic on that record disagrees with it.

### Options

1. **Carry the parent end.** Add `parent_end` to `GenomicPlacement`, populate it from `lrg_end` (LRG) and the alignment's parent end (RefSeqGene), and length from it. Correct in all cases, but it touches `parent_to_nc` and therefore projection, so it wants its own blast-radius measurement rather than riding along here.
2. **Keep the formula, state the assumption.** Note that it is exact only for a length-preserving placement, that it is a lower bound otherwise, and drop `LRG_1075` as the cited justification — it is ground truth for the *frame*, which is what this PR gets right, but not for the *width*.

Option 2 looks proportionate for this PR; option 1 is worth an issue either way.

PR #1504, opened 2026-08-06
Repairs 'cargo test'

geoffjentry:
I was seeing build errors when running `cargo test` as per the README. After deciding it wasn't operator error due to my unfamiliarity with Rust I dug in a bit. **NB** I'm trusting my robot companion here but can attest it went from not working to working. It appears the underlying issue was that CI does not actually run `cargo test` and thus missed this.

I'm diving into the deep end of the pool on how rust's testing infra works but it looks like [parallel was added](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1372/changes#diff-c7a2fa680bdc992311f6212492803e8d4d94285f7b9e639d9792b5bf53b692ffR841) but not added to the test suite properly?

#### Claude's error description

  The bug: src/batch/processor.rs:841 — streaming_batch_matches_the_vec_api_exactly calls four methods (parse_parallel, parse_streaming,
  parse_and_normalize_parallel, parse_and_normalize_streaming) that live in an impl block gated #[cfg(feature = "parallel")] at line 495. The test module (line
  653) is plain #[cfg(test)]. So cargo test — the exact command at README.md:433 — died with 4×E0599 before running anything.

  Introduced by #1372 (c9e4cca, 2026-08-04). CI never noticed because every job passes --features dev, which enables parallel.

### DIFF
--- a/src/batch/processor.rs
+++ b/src/batch/processor.rs
@@ -838,6 +838,7 @@ mod tests {
     /// The streaming batch methods must agree item-for-item with the `Vec`-based
     /// ones (#975), across the thread settings that select different engines
     /// (serial, global pool, dedicated pool).
+    #[cfg(feature = "parallel")]
     #[test]
     fn streaming_batch_matches_the_vec_api_exactly() {
         let processor = processor();


nh13:
@coderabbitai pause

nh13:
Verified independently — the diagnosis is exactly right, including the count and the four method names.

**Reproduced on `origin/main` (`a530cdaf`), before the fix:**

```
$ cargo test --no-run --lib
error[E0599]: no method named `parse_parallel` found for struct `processor::BatchProcessor<P>` …
error[E0599]: no method named `parse_streaming` found …
error[E0599]: no method named `parse_and_normalize_parallel` found …
error[E0599]: no method named `parse_and_normalize_streaming` found …
```

**With this PR applied**, checked on all three axes that matter — the fix must repair the bare build *without* quietly disabling the test where it does run:

| | result |
|---|---|
| `cargo test --no-run --lib` (no features — the `README.md:433` path) | builds |
| `cargo nextest run --features parallel -E 'test(streaming_batch_matches_the_vec_api_exactly)'` | **runs, passes** |
| `cargo nextest run --features dev …` (as CI runs it) | **runs, passes** |

That second and third row are the ones worth stating: gating a test is a fix that can silently delete coverage, and here it does not — the test still executes under both feature sets that enable the methods it exercises.

### Your read on CI is right, and it is narrower than "CI doesn't run `cargo test`"

CI *does* have a featureless build — it just cannot see this. Every test invocation in `ci.yml` is one of:

| invocation | features | scope |
|---|---|---|
| `cargo nextest archive --features dev …` | `dev` (enables `parallel`) | everything |
| `cargo nextest archive --cargo-profile soak --test it` | none | **`--test it` only** |

So the featureless compile exists but only ever covers `tests/it/`. **The lib unit tests in `src/**` are never compiled without features anywhere in CI**, which is precisely the hole a feature-gated call from an ungated `#[cfg(test)]` module falls through. Hence: found by someone following the README, not by the suite.

Filed #1505 for the systemic half — a `cargo test --no-run --lib` compile-only guard would close it for about the cost of one build, and it also raises whether `README.md:433` should keep promising bare `cargo test` at all.

Nothing to change in this PR; the one-line gate is the right minimal fix and I would not widen a contributor PR to carry the CI work. Thanks for chasing it down rather than working around it — the "I'm trusting my robot companion but can attest it went from not working to working" note is the useful kind of report, and the attestation held up.

msto:
Is it worth having at least one job run bare `cargo test`, without `--features dev`, to prevent future regressions?

nh13:
## Correction to my verification above

I checked this fix against `--features parallel` and `--features dev` and concluded it does not silently delete coverage. **I did not check the default build, which is the one case where gating a test does exactly that.** #1509 makes the point, and it holds:

```
# this PR, default features
$ cargo test --lib streaming_batch_matches_the_vec_api_exactly
running 0 tests
test result: ok. 0 passed; 0 failed; ... 3776 filtered out

# #1509, default features
$ cargo test --lib streaming_batch_matches_the_vec_api_exactly
running 1 test
test batch::processor::tests::streaming_batch_matches_the_vec_api_exactly ... ok
test result: ok. 1 passed; 0 failed; ... 3797 filtered out
```

So `cargo test` stops erroring under this PR, but the correctness-equivalence test it was erroring on is now filtered out of that build rather than run. My table had a row for every feature set except the one the bug was reported against.

The premise behind #1509 also checks out: `[dependencies.rayon]` carries no `optional = true` (`Cargo.toml:108`), so `parallel = []` enables no dependency and the gate never excluded code from any build — only names from a namespace. A gate that cannot express a capability difference is not protecting anything, which is why removing it is the better shape than matching it.

None of that is a criticism of this PR — a one-line gate is the right *minimal* unblock, you found a real bug the whole CI matrix missed, and #1509 says it supersedes rather than replaces. Just correcting my own verification, since I stated a stronger conclusion than I had measured.

nh13:
Re-signed your commit so it satisfies this repo's `required_signatures` branch rule — that was the only thing blocking the merge, not CI. All 8 required checks were already green on it.

The content is byte-identical (`git diff 9b0bd198 8f6fe990` is empty); only the signature changed. You remain the author — `git log` still shows `geoffjentry <geoffjentry@gmail.com>` — with me as committer, which is the standard shape for this.

Nothing needed from you. If you send more PRs here, signed commits will save a round trip: `git config --global commit.gpgsign true` with an SSH or GPG signing key set up.

Thanks again — this was a real bug that the whole 34-check CI matrix could not see, and it took someone actually following the README to find it. #1505 tracks closing that hole, and #1509 goes further by retiring the `parallel` gate entirely, since it turned out to gate nothing (`rayon` is an unconditional dependency, so the feature excluded no code from any build — only names from a namespace).

PR #1506, opened 2026-08-06
fix(normalize): see cis members that span a region boundary

### LINKED ISSUE #1482: Normalizer: a del+ins spanning a CDS boundary emits an overlapping allele its own parser rejects
### Summary

`normalize` turns a two-member allele whose members share a CDS-boundary span into one whose members **overlap**, and `parse_hgvs` rejects the result. The output is not a description ferro can read back.

Reproduces on `origin/main` at `2ddc25b9` with shipped defaults, on the single-exon synthetic reference the representation corpus already builds.

### Reproducer

Core `GCGCTAGTCTCGCCCTGTTA`, CDS `1..15`, 3' shuffle:

```
NM_TEST.1:c.[15_*1del;15_*1insCC]
  -> NM_TEST.1:c.[14_15dup;15_*1del]
```

`14_15dup` inserts after 15 and `15_*1del` removes 15, so the two members claim the same base. Feeding the output back:

```console
$ ferro parse 'NM_TEST.1:c.[14_15dup;15_*1del]'
# fails
```

### Caught by the re-parse oracle, and invisible without it

`FERRO_ASSERT_REPARSE=1` fires on this row. `FERRO_ASSERT_IDEMPOTENT` and `FERRO_ASSERT_IN_BOUNDS` both pass, which is the split those three oracles exist for.

It is **already reachable from the committed corpus** — `junction_interior_to_span` (added by #1459) builds it. It has stayed invisible because `examples/dump_normalized_corpus.rs` wraps normalization in `catch_unwind` and records a panic as data rather than aborting the dump. That is deliberate and correct for a dump, but it means an oracle fire there is a TSV cell nobody reads rather than a red job.

### Scale

One row of 33,808 on `main` today. #1481 adds a multi-exon coding axis and the same shape appears twice more, including at a boundary that did not previously exist:

```
c.[-1_1del;-1_1insCC]   -> c.[-1_1del;1_2dup]       5'UTR/CDS
c.[12_*1del;12_*1insCC] -> c.[11_12dup;12_*1del]    CDS/3'UTR
```

So it is not specific to the 3'UTR boundary or to one reference — it is the CDS-boundary case of a del+ins sharing a span.

### Prior art

The same *class* has been filed and fixed repeatedly for other shapes — #1263 (an allele its own parser rejects, del and dup claiming a base), #1274, #1343, #1344. This is a boundary-spanning instance those fixes do not cover.

### Suggested scope

The repair that produces `[14_15dup;15_*1del]` is choosing a dup spelling whose implied insertion point falls inside the sibling's deleted span. Either it should decline and leave the input's spelling, or it should order and merge the two members the way the overlap detectors already do elsewhere.

Whatever the fix, it moves a normalized output, so it needs the representation-stability measurement — `examples/dump_normalized_corpus.rs` builds this shape, so the harness can quantify it directly.

nh13:
Closes #1482

## The defect

```
NM_TEST.1:c.[15_*1del;15_*1insCC]  ->  NM_TEST.1:c.[14_15dup;15_*1del]
```

on a `GCGCTAGTCTCGCCCTGTTA` core with a `1..=15` CDS. That output is rejected by ferro's own `parse_hgvs` — `SelfCancellingAllele`, "variants at index 1 (del) and 0 (dup) describe overlapping reference positions". 3'-direction only on this core; 5' gave `c.[13_14dup;15_*1del]`, which reparses.

## The issue was wrong about its own cause, and the fix is somewhere else entirely

The issue reports the repair as "choosing a dup spelling whose implied insertion point falls inside the sibling's deleted span". There is no such choice. Measured, each member alone:

```
c.15_*1insCC  ALONE -> c.14_15dup    (3')
c.15_*1del    ALONE -> c.15_*1del    unchanged
```

`c.14_15dup` is the *correct* normalization of that insertion in isolation — `duplication.md:18` makes `dup` mandatory when a variant can be described as one — and the allele output is just the two per-member results side by side. Input order makes no difference.

The real cause is a false premise written into a code comment in `merge::join_pos`:

```rust
/// Both endpoints of an interval must share the same `Region` for the
/// interval to be merge-eligible. Cross-region ranges (`c.-1_1`, etc.)
/// have no valid HGVS syntax, so failing this check on a parsed
/// `Interval` indicates upstream malformedness rather than a normal
/// merge barrier; we treat it as ineligible just in case.
```

**They do have valid HGVS syntax.** `c.15_*1del` deletes across a stop codon, a common real variant, and the comment's own example of the impossible thing — `c.-1_1` — is written out in `consultation/SVD-WG001.md:37`. The refusal is defensible for the *merge* path, which coalesces members onto one axis and has no representation for such a span. It was never defensible for `member_span`, which inherited it — and there it did not read as a refusal at all: it returned `None`, and every sibling-awareness pass drops a `None` through `filter_map` rather than declining.

So a boundary-crossing member was invisible **as a sibling** across the whole module. Verified directly:

```
member_span(c.15_*1del)  ->  None
member_span(c.14_15dup)  ->  Some(("Cds", 14, 15, claims_bases: false, junction: 15))
```

`respell_colliding_duplications` — the pass that exists to turn exactly that duplication back into an insertion, "a form that claims nothing and so cannot collide" (#1320/#1323) — never saw the deletion. Nor could `clamp_sibling_crossing_shifts`, `clamp_sibling_crossing_junctions`, `demote_repeats_spanning_siblings`, `coalesce_members_at_one_junction` or `drop_identity_members_covered_by_siblings`, all of which key on the same reader.

**The scope is therefore wider than the issue title.** *Any* member whose span crosses a region boundary was invisible to *every* sibling-aware pass.

## The fix

`MemberSpan` now carries **sequence** coordinates. Each endpoint is converted through `region_span_delta` for *its own* region, so a member spanning two regions is representable and every pass sees it at once.

That representation is also the one in which the arithmetic is right. The `c.`/`n.` axis is piecewise — three regions numbered by three independent integer sequences — so an axis coordinate does not order against one in another region (`c.*1` is 3' of `c.15` while `1 < 15`), and a cross-region span has no meaningful axis length (`c.15_*1` is two bases; `1 - 15` is `-14`). Sibling comparison is the only thing these passes do with a span, so sequence coordinates are the default and the written pair (`axis_start` / `axis_end`) has to be asked for by name.

Three passes *write* positions back onto an axis rather than only comparing them, and those keep the written pair and decline a cross-region member explicitly (`respell_as`, `translate_member`, and the `demote_repeats_spanning_siblings` respell). `respell_at_gap` needs no refusal for a *cross-region* member — it already resolves each end through the transcript and names it in whichever region it lands in, which is how it produces `c.-1_1` in the first place — but it does now refuse a member in a region with no bases of its own, for the reason in the `n.`-axis section below.

Six call sites were converting a span coordinate a second time and are now reading it directly; `payload_at_junction` loses its `delta` parameter for the same reason. `cis_axis_parts` is rewritten as `member_axis_endpoints` + `join_pos`, which is what it already was — the single-region requirement is now the whole of the visible difference between the two readers.

A rejected alternative, recorded because it compiles: a two-part `(region, coord)` key with region-rank ordering. 42 sites do arithmetic on member spans and many compute `end - start + 1` as a length; blanket operator impls would make all 42 compile while silently producing wrong lengths for exactly the members just made visible.

## Representation stability

`examples/dump_normalized_corpus.rs`, 78,028 rows, against `a530cdaf` (this PR's base, and `main` at the time of writing):

| | rows |
|---|---|
| total | 78,028 |
| **moved** | **68 (0.1%)** |

| family | rows | moved |
|---|---|---|
| `dup_plus_ins` | 4,608 | 28 |
| `adjacent_junction_ins` | 4,608 | 15 |
| `coincident_bounds` | 4,608 | 14 |
| `junction_interior_to_span` | 9,216 | 6 |
| `dup_plus_sub` | 4,608 | 3 |
| `del_plus_sub` | 4,608 | 2 |
| the other nine families | 45,772 | 0 |

**35 of the 68 are outputs that denoted the wrong bases.** Measured with `canonical_spdi`, which derives an encoding-invariant key from the bases a description produces, applied to every row's input and output on both revisions:

| | outputs denoting bases their input does not |
|---|---|
| base (`a530cdaf`) | **80** |
| this PR | **45** |
| newly wrong | **0** |

All 45 remaining are in one family, `adjacent_junction_ins`, and are untouched here. **They are not the #1235 non-confluence defect**, which an earlier draft of this body claimed — that was inferred from the corpus family's label and is wrong. #1235 and its follow-ups (#1419/#1420/#1421) are about two spellings of one variant normalizing to different strings; these 45 are stronger and worse — a well-formed, strict-**accepted** input normalizing to a description of *different bases*. Reproduced standalone on `c77267d0`:

```
NM_TEST.1:c.[1_2insAA;2_3insTT]     core TAAAATTATATTTATTATTT, CDS 1..=15
  3' -> c.2_3insTTAA        5' -> c.1_2insATTA
  input  SPDI: NM_TEST.1:4::TTAA
  output SPDI: NM_TEST.1:2::TTAA     <- different resulting sequence
```

The two members sit at disjoint junctions, so the applier accepts both; the keys are encoding-invariant and maximally 3'-shifted, so this is not a spelling difference. Filed as #1513. Two examples of what the 35 were:

```
c.[8_10dup;11_12insCC]   base: c.[11_12insCC;14_*1dup]   +4 bases for a +5 input
                         this: c.14_15insCCTAA           +5
c.[12_14dup;15_*1insCC]  base: c.[14_*1dup;*1C[3]]       16::CCCGC
                         this: c.*1_*2insATCCA           16::CGCCC == the input's key
```

Classified over all 68:

| before → after | rows |
|---|---|
| denoted the wrong bases → correct | 34 |
| output could not be applied → correct | 8 |
| correct → correct (pure representation change) | 16 |
| unverifiable either side (input itself has overlapping members) | 7 |
| correct → output has overlapping members | 2 |
| wrong → output has overlapping members | 1 |

The three rows in the last two lines are `dup_plus_ins` / `junction_interior_to_span` shapes whose inputs are deliberately self-overlapping; the applier declines both before and after in most of the family, and none of the three is the reported shape.

**For a downstream consumer keying on the normalized string:** 16 rows are a migration in the strict sense — an equivalent form replaced by a different equivalent form. The other 52 either corrected a wrong answer or moved between two forms that were already not applicable. All of it is confined to cis alleles with a member spanning the CDS/UTR boundary, which is a shape that could not previously produce a stable answer at all.

## Tests

`tests/it/issue_1482_boundary_crossing_sibling.rs` — 8 tests, and **every one of the 5 defect tests was run against `a530cdaf` and fails there**; the 3 control/regression tests pass on both, as their doc comments say.

That check was not ceremonial. The first cut of this file had 5 tests of which **3 passed unchanged on the pre-fix revision** — the boundary shapes they used never reached the collision on the core they were written against, so they asserted something already true. The replacements come from a sweep over 5 cores × 5 CDS layouts × 8 shapes × 2 directions on the old code, and each fixture names the specific core and CDS that reproduced.

Two unit tests in `src/normalize/merge.rs` pin the root cause at the reader rather than only through a description: `member_span` returns a span for `c.12_*1del` and `c.-1_1del`, in sequence coordinates, with the length that makes them comparable. Both were confirmed to return `None` on `a530cdaf`.

Full local gate, all green, **nothing re-blessed**:

```
  cargo nextest run --features dev                          9533 passed, 0 failed
```

For an output-moving change, no re-blessing is the strong signal: every existing expectation already agreed with the new answer.

## An `n.`-axis regression this change first introduced, and the fix for it

Recorded because the corpus could not see it and the first push of this branch shipped it.

Making `member_span` convert onto the sequence axis means it now needs a conversion to *exist*, and `region_sequence_delta` refuses `TxUpstream` (`n.-N`) and `TxDownstream` (`n.*N`) — there are no bases to read there. That is right for reading and wrong for comparing, so routing `member_span` through it **removed** sibling visibility for those members while adding it for boundary-crossing ones. The same defect, in a new place:

| input | `main` | first push of this branch |
|---|---|---|
| `n.[-5=;-6_-4del]` | `n.-6_-4del` | `n.[-6_-4del;-5=]` |
| `n.[-5=;-5del]` | `n.-5del` | `n.[-5=;-5del]` |
| `n.[*5=;*4_*6del]` | `n.*4_*6del` | `n.[*4_*6del;*5=]` |
| `n.[5=;4_6del]` (in-transcript control) | `n.4_6del` | `n.4_6del` |

The second row is the bad one: two members left on one position, which is exactly what `drop_identity_members_covered_by_siblings` exists to remove.

`region_span_delta` widens the conversion to those two regions with the offsets the axes actually mean (`n.-N` is transcript `1 - N`, `n.*N` is `length + N`), so a cross-region comparison is meaningful rather than merely consistent. The positions it yields there are **virtual** — outside `[1, length]` by construction — and nothing can read bases through one: `get_sequence` refuses a non-positive coordinate, and `respell_at_gap` now declines such a member outright rather than relying on that. All four rows above are back to `main`'s values, pinned in `n_axis_members_outside_the_transcript_stay_visible_to_siblings`.

**The corpus is structurally blind to this.** `dump_normalized_corpus` emits `c`, `cx` and `g` rows and **no `n.` rows at all** — 0 of 78,028. So the "68 moved, 0 newly wrong" above is a statement about the axes it builds, and the `n.` behaviour is pinned by hand instead. Reporting the zero without that caveat would have been the failure mode this repo's `CLAUDE.md` calls "a corpus zero is a claim about the corpus, not about the change".

## Three things found on the way, filed not fixed

**#1513 — two insertions at neighbouring junctions normalize to a different sequence.** The 45 residual rows above, measured here and filed rather than fixed: a well-formed, strict-accepted input normalizing to a description of different bases. Pre-existing; #1506 reduces the count from 80 to 45 as a side effect and leaves this class alone.

**#1514 — give `dump_normalized_corpus` a `--verify-spdi` mode.** The `canonical_spdi` comparison that produced every "denoting bases their input does not" number in this PR was a throwaway patch, reverted after use. It is what turned "68 rows moved" into "34 were fixing a wrong answer, 0 became wrong", and rebuilding it per-PR is how a corpus goes years without anyone asking it that question.

**#1508 — strict mode accepts overlapping cis alleles when one member crosses a region boundary.** `src/normalize/overlap.rs` carries its own copy of the same false premise, in `cds_range` / `tx_range` / `rna_range`, so both cis overlap detectors are blind to the same members. Measured on `a530cdaf`, four pairs that strict mode rejects with W5002 inside one region are silently accepted across the CDS/3'UTR boundary.

An earlier draft of this PR described that as "the parser is inconsistent — it rejects a `dup` overlapping a `del` and accepts one overlapping an `inv`". **That was wrong**, and the controls are what showed it. The parser's E3006 implements one specific spec rule (`general.md:58`) and implements it correctly on *both* sides of a boundary, because `detect_self_cancelling_pair` compares a region-ranked `CanonicalPos` instead of refusing to read a cross-region interval. The general "two members claim the same positions" case belongs to W5002, and W5002 is the blind one. #1508 has the measurements and a suggested fix that reuses `CanonicalPos`'s ordering as precedent.

Not fixed here: it makes strict mode reject inputs it currently accepts, which is a user-visible change needing its own blast-radius measurement.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -679,37 +679,20 @@ fn body_region(kind: CisKind) -> Region {
 /// a disqualifying position (offset / special / mixed-region), or an
 /// uncertain / non-`NaEdit` edit — the caller treats that as "decline to
 /// collapse".
+///
+/// Written as [`member_axis_endpoints`] plus [`join_pos`], because that *is*
+/// what it was: the per-axis `simple_*_range` helpers each fold their two ends
+/// through `join_pos`, and folding them here instead keeps the two readers from
+/// drifting apart. The single-region requirement is the whole of the difference
+/// between them (#1482) — which is worth being able to see in one place, since
+/// it is also the whole of the defect that motivated the other reader.
 fn cis_axis_parts(
     v: &HgvsVariant,
     kind: CisKind,
 ) -> Option<(&Accession, Region, i64, i64, &NaEdit)> {
-    let (accession, region, s, e, edit) = match (kind, v) {
-        (CisKind::Genome, HgvsVariant::Genome(g)) => {
-            let (r, s, e) = simple_genome_range(&g.loc_edit.location)?;
-            (&g.accession, r, s, e, &g.loc_edit.edit)
-        }
-        (CisKind::Mt, HgvsVariant::Mt(m)) => {
-            let (r, s, e) = simple_genome_range(&m.loc_edit.location)?;
-            (&m.accession, r, s, e, &m.loc_edit.edit)
-        }
-        (CisKind::Cds, HgvsVariant::Cds(c)) => {
-            let (r, s, e) = simple_cds_range(&c.loc_edit.location)?;
-            (&c.accession, r, s, e, &c.loc_edit.edit)
-        }
-        (CisKind::Tx, HgvsVariant::Tx(t)) => {
-            let (r, s, e) = simple_tx_range(&t.loc_edit.location)?;
-            (&t.accession, r, s, e, &t.loc_edit.edit)
-        }
-        (CisKind::Rna, HgvsVariant::Rna(rv)) => {
-            let (r, s, e) = simple_rna_range(&rv.loc_edit.location)?;
-            (&rv.accession, r, s, e, &rv.loc_edit.edit)
-        }
-        _ => return None,
-    };
-    if !edit.is_certain() {
-        return None;
-    }
-    Some((accession, region, s, e, edit.inner()?))
+    let (accession, start, end, edit) = member_axis_endpoints(v, kind)?;
+    let (region, s, e) = join_pos(Some(start), Some(end))?;
+    Some((accession, region, s, e, edit))
 }
 
 /// Whether a merged anchor restates the reference bases under its own span —
@@ -1068,11 +1051,31 @@ fn anchor_from_loc_edit<L>(
     }
 }
 
-/// Both endpoints of an interval must share the same `Region` for the
-/// interval to be merge-eligible. Cross-region ranges (`c.-1_1`, etc.)
-/// have no valid HGVS syntax, so failing this check on a parsed
-/// `Interval` indicates upstream malformedness rather than a normal
-/// merge barrier; we treat it as ineligible just in case.
+/// Fold two endpoints into one interval, **refusing a pair that straddles a
+/// region boundary**.
+///
+/// The refusal is a limit of the *merge* path, not a property of HGVS.
+///
+/// This comment used to say cross-region ranges "have no valid HGVS syntax, so
+/// failing this check on a parsed `Interval` indicates upstream malformedness".
+/// **That is false**, and #1482 is what it cost. `c.15_*1del` deletes across a
+/// stop codon — an ordinary and common real variant — and `c.-1_1`, which the
+/// old comment offered as its example of the impossible thing, is written out in
+/// `consultation/SVD-WG001.md:37`.
+///
+/// Refusing here is still right: the merge coalesces members onto one axis and
+/// has no representation for a span numbered on two. What was wrong was
+/// inheriting the refusal *and its justification* elsewhere. [`member_span`]
+/// did, and because every sibling-awareness pass drops a `None` through
+/// `filter_map` rather than declining, a boundary-crossing member became
+/// invisible **as a sibling** module-wide — which is how
+/// `c.[15_*1del;15_*1insCC]` came to normalize into an allele ferro's own parser
+/// rejects. That reader now converts each endpoint onto the sequence axis
+/// instead; see [`member_span`] for the whole argument.
+///
+/// So: use this when the caller genuinely needs one region for one axis, and
+/// read the `None` as "this pass cannot express that span", never as "that span
+/// is malformed".
 fn join_pos(
     start: Option<(Region, i64)>,
     end: Option<(Region, i64)>,
@@ -3301,6 +3304,48 @@ fn region_sequence_delta<P: ReferenceProvider>(
         Region::FivePrimeUtr => cds_start(),
         Region::ThreePrimeUtr => Some(ordered_cds_bounds(provider_key, provider)?.1),
         Region::TxUpstream | Region::TxDownstream => None,
+    }
+}
+
+/// [`region_sequence_delta`] widened to the two `n.` regions that lie *outside*
+/// the served sequence, for **comparison only**.
+///
+/// `region_sequence_delta` refuses `TxUpstream` (`n.-N`) and `TxDownstream`
+/// (`n.*N`) because there are no bases to read there, and that is the right
+/// answer for every caller that reads bases. It is the wrong answer for
+/// [`member_span`], whose callers only ever *compare* members: refusing there
+/// does not make a pass conservative, it makes the member invisible, which is
+/// the whole of #1482.
+///
+/// Measured, before this existed: `n.[-5=;-6_-4del]` normalized to
+/// `n.-6_-4del` on `main` and to `n.[-6_-4del;-5=]` with `member_span`
+/// converting through `region_sequence_delta` alone —
+/// `drop_identity_members_covered_by_siblings` stopped seeing that the deletion
+/// covers the identity. `n.[-5=;-5del]` was worse: it kept both members on one
+/// position, which is the overlap that pass exists to remove.
+///
+/// The offsets are the ones the axes actually mean, so a cross-region
+/// comparison is meaningful rather than merely consistent: `n.-1` is the base
+/// immediately 5' of `n.1`, so `n.-N` is transcript position `1 - N`; `n.*N` is
+/// `N` bases past the end, so it is `length + N`.
+///
+/// **The positions it yields for those two regions are virtual** — outside
+/// `[1, length]` by construction. Nothing may read bases through them, and
+/// nothing does: every base read in this module goes through
+/// `provider.get_sequence`, whose `u64::try_from` refuses a non-positive
+/// coordinate, and [`respell_at_gap`] refuses such a member outright rather
+/// than relying on that.
+fn region_span_delta<P: ReferenceProvider>(
+    region: Region,
+    provider_key: &str,
+    provider: &P,
+) -> Option<i64> {
+    match region {
+        Region::TxUpstream => Some(1),
+        Region::TxDownstream => {
+            i64::try_from(provider.get_sequence_length(provider_key).ok()?).ok()
+        }
+        other => region_sequence_delta(other, provider_key, provider),
     }
 }
 
@@ -5626,9 +5671,33 @@ struct MemberSpan {
     /// lookup through it would silently fail and skip the rewrite; every other
     /// provider lookup in this module already reduces first.
     provider_key: String,
+    /// The region the span **starts** in. For a member wholly inside one region
+    /// this is its region; for one crossing a boundary it is where the span
+    /// begins, which is the axis its written coordinates are numbered on.
     region: Region,
+    /// **Sequence** coordinates, 1-based inclusive — the representation in which
+    /// any two members of one molecule are comparable and every length is right.
+    ///
+    /// The `c.`/`n.` axis is piecewise: three regions numbered by three
+    /// independent integer sequences (`-n`, `n`, `*n`), so an axis coordinate
+    /// does not order against one in another region — `c.*1` is 3' of `c.15`
+    /// while `1 < 15` — and a span crossing a boundary has no meaningful axis
+    /// length: `c.15_*1` is two bases, but `1 - 15` is `-14`. Converting each
+    /// endpoint through `region_sequence_delta` removes both problems at the
+    /// source, which is why this is the default representation and the axis one
+    /// below has to be asked for by name (#1482).
     start: i64,
     end: i64,
+    /// The coordinates as **written**, on the member's own axis.
+    ///
+    /// Only for rendering a position back onto that axis; `respell_as` is the
+    /// sole consumer. Meaningless as a length or an ordering whenever
+    /// `crosses_regions` is set, which is precisely why comparisons use the
+    /// sequence pair above.
+    axis_start: i64,
+    axis_end: i64,
+    /// Whether the two endpoints lie in different regions.
+    crosses_regions: bool,
     /// Whether the edit consumes the reference bases under its span.
     claims_bases: bool,
     /// Whether an insertion landing flush against this edit is the adjacency the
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
Independently reproduced the three load-bearing claims, then triaged the CLI lens. Two of its four findings survive; one is wrong as framed and would break the PR if applied literally.

## What I verified

**The 5-vs-2 test split is exact.** Copied `issue_1482_boundary_crossing_sibling.rs` onto `a530cdaf`, registered it, ran it there:

```
Summary  7 tests run: 2 passed, 5 failed
```

The 5 defect tests fail on the base and the 2 controls pass, precisely as claimed. Given the first cut of that file had 3 tests passing unchanged pre-fix, re-running against the base was the check that mattered, and it holds.

**The corpus table reproduces cell for cell.** Dumped 78,028 rows on `a530cdaf` and on this branch and compared:

| family | PR body | measured |
|---|---|---|
| `dup_plus_ins` | 28 | 28 |
| `adjacent_junction_ins` | 15 | 15 |
| `coincident_bounds` | 14 | 14 |
| `junction_interior_to_span` | 6 | 6 |
| `dup_plus_sub` | 3 | 3 |
| `del_plus_sub` | 2 | 2 |
| other nine | 0 | 0 |

68 of 78,028.

**No silent revert.** `merge.rs` is where #1470's rebase quietly undid a merged fix, so: **0 tests removed** (2 added), and every deleted `fn` is re-added with a new signature — `member_span`/`respell_as`/`translate_member` gain `P: ReferenceProvider`, `cis_axis_parts` becomes `member_axis_endpoints`. Nothing dropped.

## The one finding worth acting on

**`drop_identity_members_covered_by_siblings` still gates on region equality** (`merge.rs:7670`):

```rust
&& s.region == span.region
```

That pass is pure comparison — its only mutation is `members.retain(...)`, it never writes an axis coordinate. So it has no reason to require both spans in one region, and with `MemberSpan` now carrying sequence coordinates it does not need to. As written, an identity in `Region::Cds` covered by a sibling in `Region::ThreePrimeUtr` stays visible to the pass and is still discarded by the region test — which is the same shape as the defect this PR removes: the member is now *seen* but a region check drops it before it can matter.

Note `MemberSpan.region` is documented as "the region the span **starts** in", so for a crossing member it is not "the region" at all, and comparing it to a sibling's start region answers a question neither pass is asking. The guarded idiom this PR introduces at `merge.rs:6096` is the right one where a region shortcut is genuinely wanted:

```rust
if a.region == b.region && !a.crosses_regions && !b.crosses_regions {
```

I have **not** built a reproducer for it, so I would call it a structural gap rather than a demonstrated bug — worth either closing here or filing, but not worth blocking on.

## The finding I would not apply

The lens also proposes removing region equality from **all** sibling-comparability predicates and the grouping logic. Applied literally that breaks this PR's own design. Classifying every remaining check by its enclosing function:

| site | function | verdict |
|---|---|---|
| 6230, 6250, 6306 | `demote_repeats_spanning_siblings` | **keep** — the PR body names it as one of the three passes that write positions back onto an axis and must decline a cross-region member |
| 6482 | `demote_coincident_tract_repeats` | judgement call, not obviously pure-comparison |
| 6610, 6617 | `clamp_sibling_crossing_junctions` | ditto — it clamps, i.e. writes |
| 7670 | `drop_identity_members_covered_by_siblings` | **remove** — pure `retain`, per above |

So the distinction is not "comparison predicates vs everything else", it is **does this pass write a coordinate back onto a piecewise axis**. Only the last row is safely unguarded.

## The remaining two

- The test file leans on `assert_output_reparses` for several concrete cases, which is syntax-only and would accept a changed member set. The file does pin exact strings elsewhere, so this is a consistency gap rather than a hole — worth pinning the expected descriptor alongside for those cases.
- `member_span` resolves `region_sequence_delta` per endpoint; caching it when both endpoints share a region is a pure win but invisible outside a profile.

Nothing here changes my read that the analysis in this PR — particularly re-diagnosing the issue's stated cause down to a false premise in a code comment, and catching your own three vacuous tests — is sound.

nh13:
Rebased onto `main` after #1511 landed, and that surfaced a **red-together** interaction neither PR's own CI could see.

## The failure

With #1511 in `main`, `the_inversion_form_no_longer_overlaps_its_sibling` failed:

```
left:  NM_TEST.1:c.[15_*1inv;15_*1insCC]     <- actual, with #1511
right: NM_TEST.1:c.[15_*1insCC;15_*1inv]     <- what the test pinned
```

Same two members, opposite order. Both sit at `15_*1`, so nothing about position determines which renders first — the order was an artifact of which pass last touched the member set, and #1511 changes that pass.

**The property this test exists for holds in both.** The doc comment states it exactly: *the `insCC` must stay an insertion*. Pre-fix it became `14_15dup`, and a `dup` claims `c.15`, which the `inv` also claims — that is the overlap. `15_*1insCC` is present in both renderings, so the defect is absent in both.

## What changed

The test now asserts that contract rather than one exact string: the `insCC` survives as an insertion, the `inv` survives, no `dup` appears, and the whole allele re-parses. Landing order can no longer flip it.

**Verified still discriminating** — this is the risk of loosening an assertion, so I checked rather than assumed. Copied the rewritten file onto `a530cdaf` and ran it there:

```
the insertion must stay an insertion, ...; got NM_TEST.1:c.[14_15dup;15_*1inv]
1 test run: 0 passed, 1 failed
```

It still catches the original bug, and now catches it by the property rather than by the rendering.

## Also

Three `spec_enumeration_tests` failures during the rebase were a **stale local fixture**, not this PR — the artifact is gitignored and regenerated per CI run, and it predated #1501's census re-bless landing in `main`. Regenerating cleared all three; the test's own message says as much and warns against reaching for the regenerate command before deciding whether behaviour actually moved.

Gate after the rebase: **9,553/9,553**, fmt and clippy clean, nothing re-blessed.

Issue #1505, opened 2026-08-06
CI never compiles the lib unit tests without features, so a feature-gated call in a plain #[cfg(test)] module ships broken


nh13:
Bare `cargo test` — the command `README.md` documents — has been broken on `main` since #1372, and no CI job could see it.

## Reproduced on `origin/main` (`a530cdaf`)

```
$ cargo test --no-run --lib
error[E0599]: no method named `parse_parallel` found for struct `processor::BatchProcessor<P>` in the current scope
error[E0599]: no method named `parse_streaming` found for struct `processor::BatchProcessor<P>` in the current scope
error[E0599]: no method named `parse_and_normalize_parallel` found for struct `processor::BatchProcessor<P>` in the current scope
error[E0599]: no method named `parse_and_normalize_streaming` found for struct `processor::BatchProcessor<P>` in the current scope
error: could not compile `ferro-hgvs` (lib test) due to 4 previous errors
```

`streaming_batch_matches_the_vec_api_exactly` (`src/batch/processor.rs:841`) sits in a plain `#[cfg(test)]` module but calls four methods from an `impl` block gated `#[cfg(feature = "parallel")]` (`:495`). #1504 fixes the instance by gating the test.

## Why CI cannot catch this class

Every test invocation in `ci.yml` either passes `--features dev` (which enables `parallel`) or is scoped to the integration binary:

| invocation | features | scope |
|---|---|---|
| `cargo nextest archive --features dev …` | `dev` | everything |
| `cargo nextest archive --cargo-profile soak --test it` | none | **`--test it` only** |

So the featureless build exists, but it only ever compiles `tests/it/`. **The lib unit tests are never compiled without features anywhere in CI.** A feature-gated call from an ungated `#[cfg(test)]` module in `src/**` is therefore invisible until a contributor runs the documented command.

That is what happened here: reported by an external contributor following `README.md:433`, not by the suite.

## Suggested shape

Add a cheap featureless compile of the lib tests to the existing build job — it needs no run, so it costs one `cargo test --no-run --lib`:

```yaml
- name: Lib tests compile without features
  run: cargo test --no-run --lib
```

Compile-only is the right scope: the point is that the *default* feature set builds, not that the gated tests run. Pair it with the existing `--test it` featureless archive and both halves of the test tree are covered.

[...]

PR #1509, opened 2026-08-06
refactor: retire the `parallel` feature gate

### LINKED ISSUE #1507: The `parallel` feature gates public API only — retire it
`parallel` enables no dependencies and excludes no code from any build. `rayon` is declared unconditionally (`Cargo.toml:104`, no `optional = true`) because the CLI `-j` batch drivers use it, and `Cargo.toml:77-80` says so outright:

```toml
# rayon is now an unconditional dependency (the CLI `-j` batch drivers use it);
# this feature is retained as a no-op alias and to gate the library `parallel`
# module's public API.
parallel = []
```

So every `#[cfg(feature = "parallel")]` in the tree is an **API-visibility** switch, not a capability one. The machinery is compiled into every build; the feature only decides whether the names are reachable.

## Why that is worth removing rather than documenting

It reads as a capability gate, and it has already been misread as one. #1504 was filed by an external contributor whose `cargo test` (the command at `README.md:433`) died with 4× E0599 — and the natural fix, gating the test to match, encodes the wrong reason: it skips a correctness-equivalence test in the default build to work around a gate that is not protecting anything.

Retiring the gate makes that test *run* under default features instead of disappearing, and removes the whole trap rather than one instance of it. #1505 covers the CI blind spot that let it ship; this covers the thing CI was blind to.

The cost is a wider default public API surface. That is additive — the API is already reachable under `--features parallel`, so nothing that compiles today stops compiling.

## Shape

- Drop `#[cfg(feature = "parallel")]` from `src/lib.rs:50` (`pub mod parallel`), `src/batch/processor.rs:495` (the `*_parallel`/`*_streaming` impl block), and `src/batch/mod.rs:60` (`pub(crate) mod streaming`).
- Retain `parallel = []` as a documented no-op compatibility alias so `--features parallel` and `python = ["dep:pyo3", "parallel"]` keep resolving. This is exactly the treatment `mmap = []` already gets two lines below it.
- Collapse the now-dead serial fallbacks: `tests/it/clinvar_hgvs_tests.rs:102,223`, `tests/it/cmrg_exhaustive_tests.rs:98`, `tests/it/paraphase_exhaustive_tests.rs:92` each carry a `#[cfg(not(feature = "parallel"))]` twin of a `par_iter` branch.
- Drop the `# #[cfg(feature = "parallel")]` scaffolding from the `src/parallel.rs` doc examples.
- Rewrite the `src/batch/mod.rs:53-59` comment, which describes a hazard that no longer exists once the gate is gone.

Supersedes #1504 (same underlying defect, root-cause fix rather than instance fix).

nh13:
Closes #1507.

`parallel = []` enables no dependencies. `rayon` is declared unconditionally (`Cargo.toml:104`, no `optional = true`) because the CLI `-j` batch drivers use it. So every `#[cfg(feature = "parallel")]` in the tree excluded no code from any build — it decided only whether names were reachable.

That is a gate which cannot express a capability difference, only a visibility one, and it was misread as the former. #1504 is an external contributor whose `cargo test` — the command at `README.md:433` — died with 4× E0599 on `main`. This removes the gate instead of matching it.

## What changed

- Dropped `#[cfg(feature = "parallel")]` from `src/lib.rs` (`pub mod parallel`), `src/batch/processor.rs` (the `*_parallel`/`*_streaming` impl block) and `src/batch/mod.rs` (`pub(crate) mod streaming`).
- Kept `parallel = []` as a documented no-op alias, so `--features parallel` and `python = ["dep:pyo3", "parallel"]` keep resolving. This is the same treatment `mmap = []` already gets two lines below it.
- Collapsed four now-dead `#[cfg(not(feature = "parallel"))]` serial fallbacks that duplicated their `par_iter` twin verbatim: `tests/it/clinvar_hgvs_tests.rs` (×2), `tests/it/cmrg_exhaustive_tests.rs`, `tests/it/paraphase_exhaustive_tests.rs`. Net −60 lines of duplicated body.
- Removed the `# #[cfg(feature = "parallel")]` / `# #[cfg(not(...))]` scaffolding from the `src/parallel.rs` doc examples, which existed only to keep them compiling in both configurations.
- Rewrote the `src/batch/mod.rs` gating comment and the `parallel` entry in `CLAUDE.md`'s feature list, both of which described the old arrangement.

## Relationship to #1504

Same defect, different depth. #1504 adds `#[cfg(feature = "parallel")]` to the test so the featureless build compiles; this removes the gate the test tripped over, so the test **runs** there instead of being filtered out:

```
$ cargo test --lib streaming_batch_matches_the_vec_api_exactly      # no features
test batch::processor::tests::streaming_batch_matches_the_vec_api_exactly ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 3797 filtered out
```

Under #1504 that same command reports `0 passed; ... 3798 filtered out` — a correctness-equivalence test skipped in the default build to work around a gate that was not protecting anything. Both unbreak `cargo test`; only this one keeps the coverage. #1504's one-line change is subsumed here, so this supersedes it — happy to land #1504 first and rebase this on top if the faster unblock is worth more.

`main` is genuinely red on this today, so the before/after is directly observable:

```
# on origin/main (a530cdaf), no local changes
$ cargo clippy --all-targets -- -D warnings
error[E0599]: no method named `parse_parallel` found ...
error[E0599]: no method named `parse_streaming` found ...
error[E0599]: no method named `parse_and_normalize_parallel` found ...
error[E0599]: no method named `parse_and_normalize_streaming` found ...

$ cargo clippy --all-targets -- -D warnings     # this branch
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 28.68s
```

The CI blind spot that let it ship is #1505, which this does not address — a follow-up adds the guard.

## Representation stability

No normalizer change; no normalized output moves. This is feature-gating, docs and dead-code removal only. `cargo nextest run --features dev` passes with **nothing re-blessed** — no fixture, ledger or census updated in this diff.

## API impact

The public API **widens** in default builds: `ferro_hgvs::parallel` and the `BatchProcessor` `*_parallel`/`*_streaming` methods become available without `--features parallel`. That is additive — everything that compiles today still compiles, since the API was already reachable under the feature, and the feature still exists as an alias. Nothing is removed or renamed.

## Verification

| check | result |
|---|---|
| `cargo clippy --all-targets -- -D warnings` (**no features** — red on `main`) | pass |
| `cargo clippy --features dev --all-targets -- -D warnings` | pass |
| `cargo clippy --all-features --all-targets -- -D warnings` | pass |
| `cargo check --features python` | pass |
| `cargo test --no-run --lib` (the `README.md:433` path) | pass |
| `cargo fmt --check` | pass |
| `cargo test --no-run --lib --bins` | pass |
| `cargo nextest run --features dev` | **9,523 / 9,523 passed**, 145 skipped |

Exit codes captured directly rather than through a pipe. Nothing was re-blessed: `git status --porcelain` after the suite lists only the nine files in this diff.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -348,7 +348,11 @@ Never commit machine-specific absolute paths (e.g. `/Volumes/...`, `/Users/...`,
 - `python` - PyO3 Python bindings (build with maturin)
 - `benchmark` - ferro-benchmark binary and tool comparison
 - `validation` - Hash-based validation with ahash
-- `parallel` - Rayon-based parallelism
+- `parallel` - **no-op alias, retained for compatibility.** Rayon-based parallelism
+  (`ferro_hgvs::parallel`, and `BatchProcessor`'s `*_parallel`/`*_streaming` methods)
+  is unconditional — rayon is a non-optional dependency, so there is nothing to enable.
+  It formerly gated that API's visibility, which read as a capability gate and was
+  misread as one (#1504, #1507).
 - `web-service` - HTTP API server
 
 ## Key Patterns
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,9 +74,13 @@ lru = "0.12.5"
 [features]
 default = []
 dhat-heap = ["dep:dhat"]
-# rayon is now an unconditional dependency (the CLI `-j` batch drivers use it);
-# this feature is retained as a no-op alias and to gate the library `parallel`
-# module's public API.
+# rayon is an unconditional dependency (the CLI `-j` batch drivers use it), so
+# this feature enables nothing and excludes no code from any build. It once
+# gated the library `parallel` module's public API; because that gate could not
+# express a capability difference, only a visibility one, it read as "this
+# build cannot do parallelism" and was misread that way (#1504, #1507). The API
+# is now unconditional and this is a no-op alias retained so `--features
+# parallel` and the `python` feature below keep resolving.
 parallel = []
 validation = ["dep:ahash"]
 # The Python batch API parallelizes across a rayon pool, so the `parallel`
--- a/src/batch/mod.rs
+++ b/src/batch/mod.rs
@@ -46,18 +46,19 @@
 //! - **Progress Tracking**: Optional callbacks for monitoring long-running batches
 //! - **Error Aggregation**: Collect all errors without stopping the batch
 //! - **Statistics**: Detailed success/failure counts and processing rates
-//! - **Parallel Support**: Enable `parallel` feature for multi-threaded processing
+//! - **Parallel Support**: multi-threaded processing via the `*_parallel` and
+//!   `*_streaming` methods, available in every build
 
 mod processor;
 
-// Gated to match its only consumers, both of which are `parallel`-only: the
-// `#[cfg(feature = "parallel")]` impl block in `processor.rs` and the streaming
-// functions in `crate::parallel`. `default = []`, so without this the module
-// compiles in a featureless build with nothing constructing `StreamingMap` —
-// dead code that `dead_code`'s warn-by-default level let through every existing
-// clippy job (they all pass `--features dev` or `--all-features`). The
-// release-profile lint added in this PR is what surfaced it.
-#[cfg(feature = "parallel")]
+// Ungated, matching its consumers: the streaming impl block in `processor.rs`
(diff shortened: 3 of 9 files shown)


nh13:
@coderabbitai pause

PR #1511, opened 2026-08-06
fix(normalize): report cis overlaps across a region boundary

### LINKED ISSUE #1508: Strict mode accepts overlapping cis alleles when one member crosses a region boundary
## Summary

Both cis overlap detectors in `src/normalize/overlap.rs` are blind to any member whose span crosses a region boundary, so **strict mode silently accepts overlapping alleles that it rejects when the same pair sits inside one region.**

This is the same false premise #1482 found in `merge::join_pos`, in a different module. `overlap.rs` carries its own copy in three places — `cds_range`, `tx_range` and `rna_range` each end with:

```rust
let (rs, s) = simple_cds(interval.start.as_single()?)?;
let (re, e) = simple_cds(interval.end.as_single()?)?;
if rs != re {
    return None;          // <-- c.15_*1 is unreadable, so the member does not exist
}
```

`None` is not a decline here. `detect_overlap_conflicts` does `let Some(key) = group_key(variant) else { continue; }` and `span_writer_footprint` returns `None` the same way, so the member is dropped from the analysis rather than treated conservatively.

## Reproduction

Measured on `a530cdaf`, one 20-base transcript `NM_TEST.1` with core `GCGCTAGTCTCGCCCTGTTA` and CDS `1..=15`, so `c.15` is the last CDS base and `c.*1` the first 3'UTR base. Each pair below is run through `parse_hgvs` and then through `NormalizeConfig::strict()`.

**Crossing the CDS/3'UTR boundary — all accepted:**

```
c.[14_15dup;15_*1inv]        PARSES, STRICT-ACCEPTS -> c.[14_15dup;15_*1inv]
c.[14_15dup;15_*1delinsAA]   PARSES, STRICT-ACCEPTS -> c.[14_15dup;15_*1delinsAA]
c.[15_*1del;15_*1inv]        PARSES, STRICT-ACCEPTS -> c.[15_*1del;15_*1inv]
c.[15_*1insAA;15_*1insCC]    PARSES, STRICT-ACCEPTS -> c.[15delinsCAA;15_*1insCC]
```

**The same four shapes wholly inside the CDS — all rejected:**

```
c.[11_12dup;12_13inv]        STRICT-REJECTS  (inv, dup)      W5002
c.[11_12dup;12_13delinsAA]   STRICT-REJECTS  (delins, dup)   W5002
c.[11_12del;11_12inv]        STRICT-REJECTS  (del, inv)      W5002
c.[11_12insAA;11_12insCC]    STRICT-REJECTS  (ins, ins)      W5002
```

Nothing about these pairs changes when they move across the boundary except that one member's span becomes unreadable. The last crossing case is the worst of the four: strict mode not only accepts it but rewrites it into `c.[15delinsCAA;15_*1insCC]`, where the `delins` claims `c.15` and the insertion sits at the junction interior to it.

## Note the parser is the layer that gets this right

`parse_hgvs` rejects `c.[14_15dup;15_*1del]` with E3006 `SelfCancellingAllele`, across the boundary, correctly. It can, because `AlleleVariant::detect_self_cancelling_pair` compares positions as a `CanonicalPos` — "region + base + offset", lex-ordered — instead of refusing to read a cross-region interval at all.

So E3006's narrowness is not the defect. It implements one specific spec rule (`recommendations/general.md:58`, "descriptions removing part of a reference sequence and replacing it with part of the same sequence are not allowed", example `c.[762_768del;767_774dup]`) and it implements it correctly on both sides of a boundary. The general "two cis members claim the same reference positions" case belongs to W5002, and W5002 is the one that cannot see across a region boundary.

## Suggested fix

`overlap.rs` is purely syntactic — no `ReferenceProvider` — so it cannot do what #1482 did in `merge.rs` and convert both endpoints onto the sequence axis. It does not need to: every operation it performs on a span is a **comparison** (intersection, strict enclosure of a junction), never a length or an offset. A region-ranked composite key is therefore sufficient and correct here:

```
FivePrimeUtr(-N) < Cds(N) < ThreePrimeUtr(*N)      for c./r.
TxUpstream(-N)   < Tx(N)  < TxDownstream(*N)       for n.
```

which is precisely the ordering `CanonicalPos` already implements for the parser's E3006 check — so the fix is to adopt the existing precedent rather than invent an ordering.

Worth stating explicitly because #1482 **rejected** that same design for `merge.rs`: there, 42 sites do arithmetic on a member span and many compute `end - start + 1` as a length, so a rank-ordered key would make all 42 compile while silently producing wrong lengths. The distinction is what the module does with the span, and the two modules genuinely differ.

## Blast radius to measure before shipping

This makes strict mode **reject** inputs it currently accepts, which is a user-visible behaviour change, not a pure bugfix. Before merging, measure over a corpus of two-member cis alleles with one boundary-crossing member: how many strict-mode acceptances become rejections, and confirm each newly-rejected allele is one the applier also declines (`Normalizer::apply_to_reference` / `canonical_spdi`) — i.e. that it really denotes no sequence. #1448 did exactly this for the intersection test it added, and reported 10,499 of 25,848.

Note `dump_normalized_corpus.rs` emits `c`, `cx` and `g` rows and **no `n.` rows at all**, so it cannot measure the `tx_range` half of this on its own.

## Related

- #1482 / #1506 — the same false premise in `merge::join_pos`, fixed there by converting spans onto the sequence axis.
- #1448 — added the span-intersection test to `detect_overlap_conflicts`; this is the same detector, blind on a different axis.
- #1451, #1437 — prior work on which geometries the two detectors own.

nh13:
Closes #1508

## The defect

Both cis overlap detectors in `src/normalize/overlap.rs` were blind to a member whose span crosses a region boundary, so **strict mode silently accepted overlaps it rejects when the same pair sits inside one region.** Measured on `a530cdaf`, a 20-base transcript with CDS `1..=15`:

```
c.[11_12dup;12_13inv]      REJECTED  W5002
c.[14_15dup;15_*1inv]      ACCEPTED            <-- same geometry, across the boundary
c.[11_12dup;12_13delinsAA] REJECTED  W5002
c.[14_15dup;15_*1delinsAA] ACCEPTED
c.[11_12del;11_12inv]      REJECTED  W5002
c.[15_*1del;15_*1inv]      ACCEPTED
c.[11_12insAA;11_12insCC]  REJECTED  W5002
c.[15_*1insAA;15_*1insCC]  ACCEPTED  ->  c.[15delinsCAA;15_*1insCC]
```

Nothing about a pair changes when it moves across the boundary except that one member's span stops being readable. The last row is the worst: strict mode not only accepts it but rewrites it into a `delins` claiming `c.15` beside an insertion at the junction interior to it.

The cause is the same false premise #1482 found in `merge::join_pos`, in a different module — `cds_range`, `tx_range` and `rna_range` each ended with:

```rust
if rs != re {
    return None;          // c.15_*1 is unreadable, so the member does not exist
}
```

`None` there is not a decline. It reaches `let Some(..) = .. else { continue }` and `span_writer_footprint` returns it the same way, so the member is **dropped from the analysis** rather than treated conservatively.

## The parser is the layer that already gets this right

`parse_hgvs` rejects `c.[14_15dup;15_*1del]` with E3006 across the boundary, correctly, because `AlleleVariant::detect_self_cancelling_pair` compares a region-ranked `CanonicalPos` instead of refusing to read a cross-region interval.

E3006's narrowness is therefore not the defect. It implements one specific spec rule — `recommendations/general.md:58`, "descriptions removing part of a reference sequence and replacing it with part of the same sequence are not allowed", example `c.[762_768del;767_774dup]` — and it implements it on both sides of a boundary. The general "two cis members claim the same reference positions" case belongs to W5002, and W5002 is the one that could not see across.

## The fix

Rank the regions so the whole axis has one order — 5' of the body, the body, 3' of it, with the written coordinate second (`AxisPos`). That is the ordering `CanonicalPos` already gives the parser, so this adopts the existing precedent rather than inventing an order.

#1482 considered exactly this key for `merge::MemberSpan` and **rejected** it, which is worth stating because it looks like a contradiction. The distinction is what each module does with a span:

- `merge.rs` takes lengths and offsets at 42 sites, many computing `end - start + 1`; for `c.15_*1` that is `1 - 15`. A rank-ordered key would have compiled everywhere and produced wrong lengths for exactly the members just made visible, so that module converts onto the sequence axis with a provider.
- `overlap.rs` has no provider and needs none: every use is a comparison — do two spans coincide, do they intersect, does a junction fall interior to a span. The one arithmetic-looking test, `gap + 1 <= end`, was already written `gap < end`.

Insertion anchors need adjacency rather than ordering, so `AxisPos::is_immediately_followed_by` handles the case where two flanking positions are adjacent without being consecutive integers (`c.15_*1`, `c.-1_1`). Crossing *into* the body is fully checkable — the 5' regions end at `-1`; crossing *out* of it depends on the CDS length, which this module cannot know, so the author's spelling is taken at its word, the same latitude the parser already extends.

The now-redundant `region != second_region` gates are removed: region equality was never the molecule test (`coord_system` is), and gating on it is what hid every cross-boundary pair.

## Blast radius — strict mode

This makes strict mode reject inputs it currently accepts, so it was measured rather than assumed. Over **76,638** two-member cis alleles — 3 cores × 3 CDS layouts × every adjacent position pair spanning both boundaries × 5 edit spellings × both shuffle directions:

| | count |
|---|---|
| strict-mode acceptances, before | 68,624 |
| strict-mode acceptances, after | 67,497 |
| **newly rejected** | **1,127** |
| newly accepted | 0 |

**All 1,127 denote no sequence**, so the change reports real overlaps and invents none:

- **1,062** have inputs the independent applier already declines.
- **65** have inputs that apply — but `main`'s own *lenient output* for them does not, checked individually with `canonical_spdi`: 65 of 65 `out_applies=false`. My first pass measured appliability of the **input** and reported these as possible false positives; that was the wrong ground truth, since the detector runs post-shift. On the right one they are all true positives.

## Blast radius — lenient output

Not a strict-only change, which is worth stating because it is easy to assume it would be. `detect_overlap_conflicts` also gates whether an allele is preserved as authored (#395), so seeing more conflicts changes lenient output too. `examples/dump_normalized_corpus.rs`, 78,028 rows, against `c77267d0`:

| | rows |
|---|---|
| total | 78,028 |
| **moved** | **416 (0.5%)** |

| family | moved |
|---|---|
| `nested_spans` | 172 |
| `junction_interior_to_span` | 130 |
| `partial_overlap_spans` | 114 |
| the other twelve families | **0** |

Every moved row is in one of the three *conflicting* families, and **412 of the 416 now return the input verbatim** — the module's documented policy for a detected conflict ("preserves the input verbatim and emits one `OverlapConflict` per overlap") rather than the rewritten form `main` produced. Zero rows moved in any non-conflicting family, which is the shape a correct result has here.

The remaining **4** move the other way: `c.[5_6inv;5_6insCC]` on the poly-C `cx` core, where the insertion canonicalises into a repeat that crosses the 5'UTR/CDS boundary, goes from verbatim to `c.[-3_5C[10];5_6inv]`. Checked with `canonical_spdi`: the **input itself declines to apply** on both revisions, and so does each output — the members genuinely overlap either way, and both revisions raise the same W5002. So it is a spelling change between two forms that both denote nothing, not a correctness change. Recorded rather than smoothed over.

**For a downstream consumer keying on the normalized string:** every one of the 416 is an allele that strict mode rejects, before and after. Nothing moves for an allele whose members are disjoint.

## Tests

`tests/it/issue_1508_overlap_across_region_boundary.rs` — 4 tests. **The 3 defect tests were run against `main` and fail there**; the 4th is the must-stay-accepted control and passes on both, as its doc comment says.

- `an_overlap_is_reported_on_both_sides_of_the_cds_boundary` pairs each cross-boundary shape with the in-region control that was already rejected, because the control is the whole argument: same geometry, different side of the stop codon.
- `the_five_prime_boundary_is_covered_too` — different arithmetic, since the axis has no zero and `c.-1`/`c.1` are adjacent with coordinates two apart.
- `the_transcript_axis_boundaries_are_covered` — `n.` regions, which the corpus harness cannot reach at all (it emits `c`, `cx` and `g` rows only).
- `disjoint_members_across_a_boundary_are_still_accepted` — the control that matters most for a change making strict mode stricter: ranking the regions puts every member on one axis, so the risk is pairs becoming *falsely* visible to each other.

Full local gate, all green, **nothing re-blessed** — notable for a change that rejects 1,127 more alleles:

```
  cargo nextest run --features dev                          9527 passed, 0 failed
```

## Relationship to #1506

Independent and non-conflicting — #1506 is `src/normalize/merge.rs`, this is `src/normalize/overlap.rs` — but they are the same defect class, and #1506's fix is what makes some of the alleles this one now reports stop being produced in the first place. Either can merge first.

### DIFF
--- a/src/normalize/overlap.rs
+++ b/src/normalize/overlap.rs
@@ -11,9 +11,9 @@
 //! Two detectors run at different points in the allele pipeline:
 //!
 //! - [`detect_overlap_conflicts`] — *post-shift* coincident bounds: span edits
-//!   (`sub`/`del`/`delins`/`dup`/`inv`/`repeat`) whose `(region, start, end)`
-//!   keys are identical. Insertions are excluded here (they anchor at a
-//!   boundary, not a single-base span).
+//!   (`sub`/`del`/`delins`/`dup`/`inv`/`repeat`) whose ranked start/end keys are
+//!   identical. Insertions are excluded here (they anchor at a boundary, not a
+//!   single-base span).
 //! - [`detect_insertion_overlaps`] — *pre-merge* junction overlaps: two
 //!   junction-anchored edits at one junction, or one interior to a span edit
 //!   (mutalyzer `EOVERLAP`, #486). Must run before the normalizer's merge step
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

Issue #1513, opened 2026-08-06
Normalizer: two insertions at neighbouring junctions normalize to a different sequence


nh13:
## Summary

A cis allele of **two insertions at neighbouring (but distinct) junctions** normalizes to a description that denotes **different bases than the input**. The input is well-formed, its members are disjoint, the applier accepts it, and strict mode accepts it. Both shuffle directions are affected.

This is not non-confluence. #1235 and its follow-ups #1419/#1420/#1421 are all about two *spellings of one variant* normalizing to different strings. This is stronger and worse: one spelling normalizing to a **different variant**.

## Reproduction

On `c77267d0`. A 20-base transcript `NM_TEST.1`, core `TAAAATTATATTTATTATTT`, CDS `1..=15`, so `c.1`=`T`, `c.2`=`A`, `c.3`=`A`:

```
NM_TEST.1:c.[1_2insAA;2_3insTT]

  3'  ->  NM_TEST.1:c.2_3insTTAA
  5'  ->  NM_TEST.1:c.1_2insATTA

  input  SPDI:  NM_TEST.1:4::TTAA
  output SPDI:  NM_TEST.1:2::TTAA        <-- different resulting sequence
  strict mode on the input:  ACCEPTS
```

Written out, the input inserts `AA` between `c.1` and `c.2`, then `TT` between `c.2` and `c.3`:

```
reference    T  A  A  A  A  T  T ...
input        T [AA] A [TT] A  A  A  T  T ...   ->  TAAATTAAATT...
3' output    T  A [TTAA] A  A  A  T  T ...     ->  TATTAAAAATT...
```

The intervening reference base `c.2` is on the wrong side of the payload. The two payloads have been concatenated as though the members shared one junction, when there is a reference base between them.

`canonical_spdi` is encoding-invariant and maximally 3'-shifted, so two descriptions of the same edit produce the same key by construction; two different keys mean genuinely different resulting sequences, not a spelling difference.

## Scale

Measured over the 78,028-row `examples/dump_normalized_corpus.rs` corpus, comparing `canonical_spdi(input)` against `canonical_spdi(output)` for every row:

| revision | outputs denoting bases their input does not |
|---|---|
| `a530cdaf` | 80 |
| with #1506 | 45 |

#1506 fixes 35 of them as a side effect of a different defect. **All 45 survivors are this shape**, spread evenly over both axes and both directions:

| input shape | rows |
|---|---|
[...]

PR #1515, opened 2026-08-07
feat(dev): add --verify-spdi to the normalized-corpus harness

### LINKED ISSUE #1514: dump_normalized_corpus: add --verify-spdi so the harness can say whether a moved row became wrong
## Summary

`examples/dump_normalized_corpus.rs` answers "**how many** rows moved". It cannot answer "did any of them become **wrong**", and that is the question a representation-stability claim actually turns on.

Add a `--verify-spdi` mode that, per row, compares `canonical_spdi(input)` with `canonical_spdi(output)` and reports the mismatches. The comparison is encoding-invariant and maximally 3'-shifted, so equal keys mean the description denotes the same bases however it is spelled — which is exactly the ground truth the harness currently lacks.

## Why

A moved row is not a defect and an unmoved row is not a proof. Without this, a PR body can only say "68 of 78,028 moved", and a reviewer has no way to tell "68 rows churned between two correct forms" from "68 rows changed what the variant means". Those are opposite outcomes reported by the same number.

Built as a throwaway patch during #1506 and thrown away afterwards, because it was an uncommitted edit to a committed file. It produced the only decisive numbers in that PR:

| revision | outputs denoting bases their input does not |
|---|---|
| `a530cdaf` | 80 |
| with #1506 | 45 |
| **newly wrong** | **0** |

and it turned "68 rows moved, is that good or bad?" into "34 of them were **fixing a wrong answer**, 16 are a genuine migration, 0 became wrong". It also found #1513 — a pre-existing sequence-correctness bug that had been sitting in the corpus unnoticed, because nothing ever asked the corpus that question.

The next normalizer PR that wants that answer has to rebuild the same patch from scratch.

## Sketch

Behind a flag rather than always-on: the check costs one `canonical_spdi` per row per side, which is a reference read plus a window walk, and the plain dump is used in a `--check`-style loop.

```
cargo run --features dev --example dump_normalized_corpus -- --out after.tsv --verify-spdi
```

Emit per mismatch, to stderr, so the tsv stays byte-compatible with existing baselines:

```
SPDI-MISMATCH  <family>  <axis> <direction>  <reference>
    in : <input>   => <key or decline reason>
    out: <output>  => <key or decline reason>
```

Two things the throwaway version got wrong the first time and that are worth building in:

1. **Report declines separately from mismatches.** `canonical_spdi` declines an allele whose members overlap, and roughly 2,462 of 78,028 rows decline on one side or the other. Counting a decline as a mismatch buried 45 real findings under 6,700 rows of noise. The three states are *equal*, *different*, and *unverifiable*, and the third is not a failure.
2. **Verify the output against the input, not the input against itself.** The first cut of this analysis applied the *input* to decide whether a newly-rejected allele was a false positive, which is the wrong ground truth whenever the pass under test runs post-shift. It reported 65 false positives that were all, on the right comparison, true positives.

## Also worth having

A summary line, so a PR body can quote one number:

```
verified 71,289 rows: 45 denote different bases, 6,694 unverifiable (member overlap)
```

and ideally a `--compare`-side integration so `--compare before.tsv --against after.tsv` can report *newly* wrong rather than absolutely wrong, which is the number that actually gates a merge.

## Related

- #1448 — "The applier is not wired in as an oracle, so 'denotes no sequence' was invisible." Same gap, one step further: this is about output that denotes *a* sequence, just not the input's.
- A prior retro recorded that six PRs' representation-stability claims never used the committed harness at all. Giving the harness a verdict rather than a count raises what it is worth using for.

nh13:
Closes #1514

## Why

`examples/dump_normalized_corpus.rs` answers **how many** rows moved. It could not answer **did any of them become wrong**, and that is the question a representation-stability claim actually turns on. A moved row is not a defect and an unmoved row is not a proof, so `68 of 78,028 moved` reports churn and correctness with one number and lets a reviewer tell them apart only by reading examples.

This was built as a throwaway patch during #1506, used on both revisions, and reverted — because it was an uncommitted edit to a committed file. It produced the only decisive numbers in that PR (`80 → 45 wrong, 0 newly wrong`, which turned "68 rows moved" into "34 were fixing a wrong answer, 16 are a genuine migration"), and it is what surfaced #1513. Rebuilding it per-PR is how a corpus goes a long time without anyone asking it that question.

## What it does

```
cargo run --features dev --example dump_normalized_corpus -- --out after.tsv --verify-spdi
```

Per row, compares `canonical_spdi(input)` with `canonical_spdi(output)`. That key is derived from the bases a description *produces* rather than from how it is written, and is maximally 3'-shifted, so two spellings of one edit key identically by construction — a difference here is a difference in meaning, not in notation.

```
SPDI-MISMATCH	dup_plus_sub	c 3prime	TTTTTTTTTAATATATTTTA
    in : NM_TEST.1:c.[11_12dup;14T>A]  => NM_TEST.1:15::AA
    out: NM_TEST.1:c.[14T>A;15_*1dup]  => NM_TEST.1:13::AA
...
verified 78028 rows: 80 denote different bases, 57422 agree, 20526 unverifiable (member overlap or unparseable)
```

## Two design points that are load-bearing

**Three verdicts, not two.** `canonical_spdi` declines an allele whose members overlap, and this corpus is built to contain those — **20,526 of 78,028** rows are unverifiable on one side or the other. Folding them into "different" is what the first hand-written cut of this check did, and it buried 80 real findings under twenty thousand rows of noise. `Unverifiable` is not a failure and is counted separately.

**The dump is untouched.** `read_dump` exact-matches the header and is strict about the column count, so adding a verdict column would stop every committed baseline from loading. Findings go to stderr instead. Verified both ways: the dump is byte-identical with and without the flag, and a flagged run still `--compare`s cleanly against a baseline taken before this change (`0 moved`).

The cost is that "newly wrong" is a two-run recipe — run the flag on both revisions and diff the reports — rather than a `--compare` mode. That is the trade for not invalidating every baseline in the repo, and it is the workflow that was used for #1506 anyway.

## Result on current `main`

`3ac054d3` reports **80** rows whose normalized output denotes different bases than its input. All 80 are one shape — two insertions at neighbouring junctions — filed as **#1513**. They are pre-existing and predate this PR; #1506 happens to reduce them to 45 as a side effect of an unrelated fix.

That number is the argument for the flag: the corpus has been run many times, and nothing had asked it this.

## Tests

Two unit tests in the example's own `mod tests`, covering all three verdicts on real corpus cores:

- `verify_row_separates_disagreement_from_unanswerable` — a 3'-shifted `dup` (same bases, different spelling) keys `Same`; the #1513 shape keys `Different`; an overlapping-member allele and an unparseable `<declined>` output both key `Unverifiable`. The third case is the one that earns the test.
- `verification_uses_the_rows_own_axis_provider` — `provider_for` is shared with `normalize_one` so verification reads the same reference the row was normalized against; building a second, subtly different provider would answer about the wrong sequence and do it silently, since both still produce keys.

Full local gate, green, nothing re-blessed:

```
  cargo nextest run --features dev                          9533 passed, 0 failed
```

No behaviour change outside the example: the flag is opt-in and the crate is untouched.

### DIFF
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -180,6 +180,77 @@ struct Cli {
     /// Diff mode: the candidate dump.
     #[arg(long, requires = "compare")]
     against: Option<PathBuf>,
+    /// Also check, per row, whether the normalized output denotes the **same
+    /// bases** as its input, and report the rows where it does not (#1514).
+    ///
+    /// Dump mode only. Findings go to stderr and the dump itself is unchanged,
+    /// so a verified run stays byte-comparable with every existing baseline.
+    #[arg(long, conflicts_with_all = ["compare", "against"])]
+    verify_spdi: bool,
+}
+
+/// What [`verify_row`] concluded about one row.
+///
+/// Three outcomes, not two, and the third is the one that matters for reading
+/// the summary: `canonical_spdi` declines an allele whose members overlap, and
+/// on this corpus that is thousands of rows. Folding those into "different"
+/// buried 45 real findings under ~6,700 rows of noise the first time this
+/// check was written by hand.
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum SpdiVerdict {
+    /// Input and output denote the same bases.
+    Same,
+    /// They denote different bases — the finding this flag exists for.
+    Different,
+    /// One side could not be applied, so the question has no answer here.
+    /// Not a failure.
+    Unverifiable,
+}
+
+/// Whether `row`'s output denotes the same bases as its input.
+///
+/// Compared through [`Normalizer::canonical_spdi`], which derives its key from
+/// the bases a description *produces* rather than from how it is written, and
+/// maximally 3'-shifts it. Two spellings of one edit therefore key identically
+/// by construction, so a difference here is a difference in meaning and not in
+/// notation — which is exactly what a row-movement count cannot tell you.
+///
+/// Wrapped in `catch_unwind` for the same reason [`normalize_one`] is: this runs
+/// over every row of a corpus built to contain the shapes that break things, and
+/// one panicking row must not take the whole dump with it.
+fn verify_row(row: &Row) -> (SpdiVerdict, String, String) {
+    let declined = |what: &str| format!("<{what}>");
+    let (Ok(input), Ok(output)) = (parse_hgvs(&row.input), parse_hgvs(&row.output)) else {
+        return (
+            SpdiVerdict::Unverifiable,
+            declined("unparseable"),
+            declined("unparseable"),
+        );
+    };
+    let direction = match row.direction {
+        "5prime" => ShuffleDirection::FivePrime,
+        _ => ShuffleDirection::ThreePrime,
+    };
+    let normalizer = Normalizer::with_config(
+        provider_for(row.axis, &row.reference),
+        NormalizeConfig::default().with_direction(direction),
+    );
+    let key = |v: &_| match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+        normalizer.canonical_spdi(v)
+    })) {
+        Ok(Ok(k)) => Some(k.to_string()),
+        Ok(Err(_)) => None,
+        Err(_) => None,
+    };
+    match (key(&input), key(&output)) {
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1516, opened 2026-08-07
fix(normalize): apply a junction barrier the translate cannot express

### LINKED ISSUE #1513: Normalizer: two insertions at neighbouring junctions normalize to a different sequence

nh13:
Closes #1513

## The defect

Two insertions at **distinct, disjoint** junctions normalized to a description of different bases than the input:

```
core  TAAAATTATATTTATTATTT      NM_TEST.1, CDS 1..=15

c.[1_2insAA;2_3insTT]   3' ->  c.2_3insTTAA
                        5' ->  c.1_2insATTA
  input  SPDI  NM_TEST.1:4::TTAA
  output SPDI  NM_TEST.1:2::TTAA        <- a different sequence
```

The allele is well-formed — the applier accepts it, strict mode accepts it. This is **not** the non-confluence of #1235 / #1419–#1421, which is two spellings of one variant disagreeing. This was one spelling normalizing to a *different variant*.

## The clamp was right; the write was silent

The pass that looks guilty is not the one at fault, so it is worth stating what was measured.

`c.1_2insAA` 3'-shifts through the `A` run to `c.4_5dup`, sweeping across the `insTT` junction at 2. `AA` and `TT` do not commute, so `clamp_sibling_crossing_junctions` bars the crossing and asks for junction 1. Instrumented, it computes exactly that:

```
CLAMP[0] c.1_2insAA -> c.4_5dup  junction 1 -> 5
CLAMP[0]   limit=Some(1)  destination=Some(1)
```

Then nothing happened. Moving a **two-base `dup`** to junction 1 means the span `c.0_1`, and the CDS axis has no zero — so `translate_member` reverted its write, and `translate_junction_member` read the unchanged variant as *"the translation was refused; nothing to repair"* and returned. The member stayed at `4_5`, precisely where the barrier had just said it must not be, and the allele went on to merge into one insertion at the wrong junction.

**Not the sequence-first re-derivation**, though that is the obvious suspect for "an allele re-derived into the wrong thing". Disabling `canonicalize_from_sequence` only respells the wrong answer as `c.[2_3insTT;4_5dup]`, which keys to the same wrong `2::TTAA`. The error is upstream of it.

## The fix

A refusal is not "nothing to repair" — the caller has already decided this member may not stay put. Fall through to `respell_at_gap`, which writes the member as a plain insertion carrying the rotated payload at the destination. That is the repair this function already performs for a translation that *lands* but changes meaning; a zero-width junction has a spelling everywhere a span may not.

The old outcome is preserved for members that genuinely cannot move: `respell_at_gap` reverts anything it cannot place, and the existing guard restores `original` when it does. The difference is that it is now reached by failing to place rather than by declining to try.

One companion change: the `landed` early-return is gated on the translation having actually happened. On a refusal the member is still `original`, so that check reads the payload where it *was* — and if the barrier asked for the position it already occupies, there was nothing to do anyway.

## Blast radius

Measured over the 78,028-row corpus against `33e286fe` (this PR's parent), using #1514's `--verify-spdi`:

| | rows |
|---|---|
| total | 78,028 |
| **moved** | **47 (0.1%)** |

| family | rows | moved |
|---|---|---|
| `adjacent_junction_ins` | 4,608 | **47** |
| the other fourteen families | 73,420 | **0** |

and on the question that matters:

| | outputs denoting bases their input does not |
|---|---|
| `33e286fe` | 80 |
| this PR | **41** |
| **newly wrong** | **0** |

39 fixed, none broken, and every moved row is in the one family the change targets.

**It composes with #1506.** That PR fixes 35 of the same 80; the two fixed sets are **disjoint** (overlap 0, union 74 of 80), so together they would leave 6. Neither subsumes the other and they touch different code.

## Tests

`tests/it/issue_1513_adjacent_junction_insertions.rs` — 4 tests. **Both defect tests were run against `33e286fe` and fail there**; both controls pass on both, as their doc comments say.

- `adjacent_junction_insertions_keep_their_sequence` — the reported allele, both directions, both input orders. Asserted through `canonical_spdi` rather than a pinned string, so it survives a later change to which spelling is canonical while still failing the moment the *meaning* moves.
- `the_genomic_axis_is_affected_too` — not redundant. The corpus reports **zero** `g.` rows for this defect, which reads as "the genomic axis is fine" and is wrong: it simply does not build this geometry there. Measured directly, `g.[1_2insAA;2_3insTT]` gave `g.2_3insTTAA` before the fix, exactly as `c.` did. A corpus zero is a claim about the corpus.
- `insertions_that_do_not_sweep_a_sibling_are_untouched` and `same_junction_insertions_are_still_preserved` — the controls that matter for a change making a clamp *fire* where it used to no-op: junctions two apart still shift freely, a lone insertion still shifts all the way, and a same-junction pair (which has no defined order) is still preserved verbatim rather than merged.

Full local gate, green, **nothing re-blessed**:

```
  cargo nextest run --features dev                          9547 passed, 0 failed
```

For a change that moves 47 corpus rows in the most contested pass in the crate, no re-blessing is the signal worth having: every existing expectation already agreed.

## Why no oracle caught it

`FERRO_ASSERT_REPARSE` passes (the output is well-formed), `FERRO_ASSERT_IN_BOUNDS` passes (every coordinate exists), `FERRO_ASSERT_IDEMPOTENT` passes (the output is a fixed point). The applier is not wired in as an oracle, so "denotes a different sequence" was invisible — the same gap #1448 recorded for its own class, one step further along. It was found by #1514's `--verify-spdi`, which is the first thing to ask the corpus that question.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -7622,13 +7622,39 @@ fn translate_junction_member<P: ReferenceProvider>(
 
     let original = variant.clone();
     translate_member(variant, delta, kind, from, provider);
-    if *variant == original {
-        return; // the translation was refused; nothing to repair
-    }
+    // A refused translation is **not** "nothing to repair" (#1513). It used to
+    // return here, and that turned a correctly-computed barrier into a silent
+    // no-op: the caller had already decided this member must not sit where it
+    // does, and returning leaves it exactly there.
+    //
+    // `translate_member` refuses whenever the destination has no spelling *as
+    // this edit kind* — and for a duplication near the 5' end of an axis with no
+    // zero, that is routine rather than exotic. On `TAAAATTATATTTATTATTT`,
+    // `c.[1_2insAA;2_3insTT]` shifts its insertion 3' to `c.4_5dup`, sweeping
+    // across the `insTT` junction it does not commute with;
+    // `clamp_sibling_crossing_junctions` computes the barrier correctly and asks
+    // for junction 1, which for a two-base `dup` means the span `c.0_1`. There
+    // is no `c.0`, so the write reverted, the member stayed at `4_5`, and the
+    // allele went on to merge into `c.2_3insTTAA` — a description of **different
+    // bases** than the input.
+    //
+    // The repair the caller wants exists: the member re-spelled as a plain
+    // insertion carrying `moved` at `destination`, which is what the rest of this
+    // function already does for a translation that landed but changed meaning.
+    // A zero-width junction has a spelling everywhere a span may not, so falling
+    // through reaches it. `respell_at_gap` reverts anything it cannot place, and
+    // the guard at the end restores `original` when it does, so a member that
+    // truly cannot be moved is left as it was — the old behaviour, now reached by
+    // failing to place rather than by declining to try.
+    let translation_refused = *variant == original;
     let landed = member_span(variant, kind, provider)
         .filter(|s| s.junction == Some(destination))
         .and_then(|s| junction_payload(variant, kind, &s, provider));
-    if landed.as_deref() == Some(moved.as_slice()) {
+    // Only meaningful when the translation actually happened: on a refusal the
+    // member is still `original`, so this reads the payload it carries *where it
+    // was*, and if the barrier asked for the position it already occupies there
+    // was nothing to do in the first place.
+    if !translation_refused && landed.as_deref() == Some(moved.as_slice()) {
         return; // the spelling still denotes the same bases
     }
     let edit = NaEdit::Insertion {
--- a/tests/it/issue_1513_adjacent_junction_insertions.rs
+++ b/tests/it/issue_1513_adjacent_junction_insertions.rs
@@ -1,0 +1,198 @@
+//! #1513 — two insertions at neighbouring junctions normalized to a description
+//! of **different bases** than the input.
+//!
+//! ```text
+//! core  TAAAATTATATTTATTATTT      NM_TEST.1, CDS 1..=15
+//!
+//! c.[1_2insAA;2_3insTT]   3' ->  c.2_3insTTAA
+//!                         5' ->  c.1_2insATTA
+//!   input  SPDI  NM_TEST.1:4::TTAA
+//!   output SPDI  NM_TEST.1:2::TTAA        <- a different sequence
+//! ```
+//!
+//! The members sit at *distinct, disjoint* junctions, so the allele is
+//! well-formed: the applier accepts it and strict mode accepts it. This is not
+//! the non-confluence of #1235 / #1419-#1421 — those are two spellings of one
+//! variant disagreeing. This was one spelling normalizing to a *different
+//! variant*.
+//!
+//! # The clamp was right and the write was silent
+//!
+//! Worth recording, because the pass that looked guilty was not.
+//!
+//! `c.1_2insAA` 3'-shifts through the `A` run to `c.4_5dup`, sweeping across the
+//! `insTT` junction at 2. `AA` and `TT` do not commute, so
+//! `clamp_sibling_crossing_junctions` bars the crossing and asks for junction 1
+//! — instrumented, it computes `limit = Some(1)`, `destination = Some(1)`,
+//! exactly right.
+//!
+//! Then nothing happened. Moving a **two-base `dup`** to junction 1 means the
+//! span `c.0_1`, and the CDS axis has no zero, so `translate_member` reverted;
+//! `translate_junction_member` read the unchanged variant as *"the translation
+//! was refused; nothing to repair"* and returned, leaving the member at `4_5`.
+//! The barrier was computed, and then dropped on the floor. The allele went on
+//! to merge into a single insertion at the wrong junction.
+//!
+//! The fix is that a refusal is not "nothing to repair": the member falls
+//! through to `respell_at_gap`, which writes it as a plain insertion carrying
+//! the rotated payload. A zero-width junction has a spelling everywhere a span
+//! may not.
+//!
+//! Note the sequence-first re-derivation is **not** implicated, though it looks
+//! like the obvious suspect. Disabling `canonicalize_from_sequence` only changes
+//! the spelling of the wrong answer: `c.[2_3insTT;4_5dup]`, which keys to the
+//! same wrong `2::TTAA`. The error is upstream of it.
+//!
+//! # No oracle saw it
+//!
+//! `FERRO_ASSERT_REPARSE` passes (the output is well-formed),
+//! `FERRO_ASSERT_IN_BOUNDS` passes (every coordinate exists), and
+//! `FERRO_ASSERT_IDEMPOTENT` passes (the output is a fixed point). The applier is
+//! not wired in as an oracle, so "denotes a different sequence" was invisible.
+//! It was found by #1514's `--verify-spdi`, which asks the corpus that question.
+
+use ferro_hgvs::reference::transcript::{Exon, ManeStatus, Strand, Transcript};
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
Rebased onto `main` (which now carries #1493, #1496, #1502, #1506, #1510, #1511, #1515), and measured with the harness from #1515 — the one this defect came out of. **The fix is roughly seven times more effective than this description claims.**

## `--verify-spdi`, both revisions

```
main    (aa58ca94): 45 denote different bases, 57464 agree, 20519 unverifiable
this PR (67408eb3):  6 denote different bases, 57503 agree, 20519 unverifiable
```

**39 of the 45 remaining wrong-bases rows are fixed here**, not the one shape the body walks through. `unverifiable` is unchanged at 20,519, so this is 39 rows moving from *wrong* to *agree* rather than from wrong to undecidable — which is the distinction #1515's three-verdict design exists to make.

Worth stating in the body, because "closes #1513" reads as a single-shape fix and this is the largest correctness improvement to the corpus measured in this batch. It also closes the loop nicely: #1515 built the check, the check found #1513, and the check now scores its fix.

## The rebase needed a real conflict resolution, not a `main.rs` union

`src/normalize/merge.rs` conflicted with #1506, which had threaded `provider` through `translate_member` and `member_span` while this PR replaces the early `return` with a `translation_refused` flag — two orthogonal edits to the same lines. Resolved as this PR's logic on #1506's signatures:

```rust
translate_member(variant, delta, kind, from, provider);
// … the #1513 comment …
let translation_refused = *variant == original;
let landed = member_span(variant, kind, provider)
```

Confirmed `translation_refused` still reaches its only consumer (`if !translation_refused && landed.as_deref() == Some(moved.as_slice())`), so the fall-through to `respell_at_gap` is intact rather than silently short-circuited — which would have restored the exact bug, with the tests still passing on the reported shape.

Gate after the rebase: **9,567/9,567**, fmt and clippy clean, nothing re-blessed.

nh13:
One CLI-lens finding, which I am **rejecting** — recording the reasoning since it argues the opposite of a change I made on #1506 an hour ago, and the difference is the point.

It asks to replace `assert_denotes_the_same_bases`' `canonical_spdi` equivalence with exact pinned normalized descriptions.

**The `canonical_spdi` form is the right assertion here**, because it is the property the defect violated. #1513 was not a spelling that moved; it was an output denoting *different bases* than its input (`4::TTAA` → `2::TTAA`). A pinned string cannot distinguish those two failure modes — it goes red for a harmless representation change, and stays green for a wrong answer that happens to render the way it was pinned. The key is derived from produced bases and maximally 3'-shifted, so it fails exactly when meaning moves.

I did check the obvious hole, because "compare two derived keys" is the shape that passes vacuously when both sides are `None` — and `canonical_spdi` returns `None` for an allele whose members overlap, which is 20,519 rows of this corpus. It is closed: both calls are `.expect(…)`, so an output that stops being keyable panics rather than quietly agreeing with an equally-unkeyable input. That is what makes this a real assertion rather than a tautology.

Representation churn *is* worth catching, and it is caught — by `dump_normalized_corpus --compare` over 78,028 rows, which is where a pinned-string check belongs. Duplicating it per-test buys little and costs a red test on every legitimate respelling.

The contrast with #1506 is not inconsistency. There I *removed* a pinned string, because it pinned the member **order** of two members at the same position — an artifact of which pass ran last, which #1511 legitimately changed while the property the test existed for still held. Same rule both times: assert the contract the issue states, not the rendering that happened to carry it.

The suggestion to also exercise `normalize_with_diagnostics()` is reasonable but not a defect — since #1382 both paths run through `normalize_core_checked`, so they share the seam and the oracles.

PR #1519, opened 2026-08-07
test(spec): assert equivalence classes converge and record contested spec readings

### LINKED ISSUE #1518: Spec fixture cannot see non-confluence: the spec's own alternative-description pair passes as `preserved`
## The spec's own non-confluent pair currently passes

`tests/fixtures/grammar/hgvs_spec_normalization.json` holds 934 rows harvested from the vendored HGVS v21.0 spec, and its `status` buckets ask exactly one question: did ferro *change* the string? Nothing asks whether two rows are the **same variant**. So this sits in the fixture and is scored as full conformance:

```
preserved   LRG_199t1:c.850_901delinsTTCCTCGATGCCTG            -> unchanged
preserved   c.[850_869del;874_881del;887_897del;901_902insG]   -> unchanged
```

`DNA/delins.md:44-47` names the second as the "alternative description" of the first — one variant, two spellings. ferro keeps both, which is two stable fixed points for one variant: #1235's exact defect, present in the spec fixture, green.

The same shape recurs at `delins.md:15-19`: `c.235_237delinsTAT` and `c.[235A>T;237G>T]` are both `preserved`. By contrast `c.[145C>T;147C>G]` *is* caught, as `false-acceptance` — but only because the spec marks it `<code class="invalid">`. The harness works when the spec states invalidity outright and is blind whenever the spec merely says "alternatively".

## Half one: equivalence-class links

Add a way to declare that two or more harvested rows denote the same variant, and assert that every member of a class normalizes to **one** output. A class whose members produce different strings fails, naming the members and their outputs.

The classes cannot be auto-derived — establishing that two descriptions denote one variant means applying both to a reference sequence, which 50 `needs-reference` rows cannot do. So they belong beside the other hand-curated spec judgements, in the committed `hgvs_spec_normalization_overrides.json`, validated at generation time exactly as `by_input` already is.

Two constraints the design has to respect:

- **Unevaluable members must not fail a class spuriously.** A row needing real reference bases, or one ferro refuses outright, is neither evidence of confluence nor of its absence; comparing an error string as though it were a representation would fail every such class for the wrong reason.
- **A class that disagrees today is expected, and must not be "fixed" by changing the normalizer.** Converging them is #1235, and every convergence picks a winner between two forms — a representation change for anyone storing the losing one. The disagreements get pinned, with reasons, so the *set* cannot grow silently; the assertion is not weakened to accommodate them.

## Half two: ruling records

Several spec clauses conflict at equal RFC 2119 strength. `style.md:9` binds the spec to RFC 2119; item 3 makes "RECOMMENDED" equal to SHOULD and item 4 covers SHOULD NOT. The live example:

- `DNA/delins.md:17` — "two variants separated by one or more nucleotides should be described individually and **not** as a 'delins'" — a SHOULD NOT against merging.
- `DNA/delins.md:47` — "The 'delins' format is **recommended**" — a SHOULD for merging.

Both reach the same block, and `:44-47`'s own alternative splits at gaps of 4, 5 and 3 — all at least 2 — so `:17` read literally demands the very split `:47` rejects three lines later. Which governs is a judgement the spec text does not settle, and RFC 2119 explicitly licenses deviating from a SHOULD once the implications are weighed.

Today that reasoning lives nowhere durable — at best in a free-text `note` on a row. Promote it into a first-class **ruling record** capturing which clause was held to **govern**, which was **deviated from**, and the **rationale**, structured enough to render and validated so that a citation the spec no longer carries fails the build.

Crucially, **`undecided` has to be a real state**. Where the project has not ruled, the record says so and names both sides — it must be impossible for an unsettled question to quietly acquire a governing clause and read as a ruling nobody made.

Three to seed:

1. `:17` vs `:44-47` on the LRG_199 example. **Undecided**, pending an operator ruling. It governs roughly 138 of 208 adjudicated corpus rows, so settling it either way is a migration rather than a fix.
2. `general.md:34` vs `general.md:56` on `c.76_83inv` (`NM_004006.2`, `c.76_83` = `AATGCACA`, whose reverse complement `TGTGCATT` coincides at 4 of 8 columns — one inversion, or two variants 4 bases apart?). **Undecided**, under active adjudication.
3. `delins.md:18`'s codon carve-out — two variants separated by exactly one nucleotide, together affecting one amino acid. This one the spec *does* settle; record it as **decided**, with the citation.

## Scope

Test, fixture and documentation infrastructure only. No normalizer behaviour changes: the point is to make an existing defect visible and pinned, not to move any representation.

nh13:
Closes #1518.

The spec-normalization fixture's statuses ask one question — did ferro *change* the string? — and never ask whether two rows are the **same variant**. So the HGVS spec's own non-confluent pair sat in the fixture and passed as fully conformant:

```
preserved   LRG_199t1:c.850_901delinsTTCCTCGATGCCTG            -> unchanged
preserved   c.[850_869del;874_881del;887_897del;901_902insG]   -> unchanged
```

`DNA/delins.md:44-47` calls the second the "alternative description" of the first. Two stable fixed points for one variant — #1235's defect, green.

## Representation stability

**No normalizer behaviour changes.** Nothing under `src/` is touched; the diff is the generator, the committed overrides file, the test driver and `CLAUDE.md`. No normalized output moves, so there is no representation change and nothing for a downstream consumer to re-baseline. The nine disagreements this PR finds are *reported and pinned*, deliberately not fixed: converging any of them picks a winner between two spellings, which is a migration and belongs to #1235.

## Half one — equivalence classes

`hgvs_spec_normalization_overrides.json` gains an `equivalence_classes` section declaring which harvested inputs denote one variant. `spec_equivalence_classes_converge` re-derives every member's output **live** (the fixture supplies only the inputs, never the outputs) and fails when a class yields more than one string, printing each member's target and rendering.

Design points worth review:

- **Curated, not derived.** Establishing that two descriptions denote one variant means applying both to a reference sequence, which 50 `needs-reference` rows cannot do. So the declarations sit beside the other hand-curated spec judgements and are validated at generation time exactly as `by_input` is.
- **Every class carries a quote-verified citation.** `{"clause": "docs/recommendations/DNA/delins.md:44-47", "quote": "…"}` — the generator asserts the quote still appears within those lines. A bare line number would keep "resolving" against any file long enough to have that line; the quote makes a submodule bump that moves a clause fail the build instead of silently re-pointing the citation at unrelated prose.
- **An optional per-class `reference`.** The spec routinely writes one member of a pair as a bare fragment (`c.[235A>T;237G>T]`), which the harvester prefixes with the per-coordinate-system default rather than with the accession of the worked example it sits under — and twice writes the two members under *different accession versions* (`NM_004006.2` / `NM_004006.1`, `DNA/insertion.md:74`, plainly a typo since the prose describes one insertion). Without an anchor those classes would report a difference of *reference sequence* as a difference of representation. **Alternative considered and rejected:** fixing this by editing each bare member's `input_prefixed`, which is what that field is for. It would move those rows' `current` values and the censuses pinned against them, for no benefit outside this comparison — so the anchor is declared on the class instead and the row set is untouched.
- **Unevaluable members are skipped, not compared.** A `needs-reference` row, or one ferro refuses, is neither evidence of confluence nor of its absence; treating a parse-error string as a "representation" would fail such classes for the wrong reason.
- **Verdicts are pinned per class, not just counted.** `EQUIVALENCE_CLASS_VERDICTS` pins `confluent` / `non-confluent` / `not-evaluable` per id, so every transition fails: a new disagreement (regression), a disagreement that resolves (improvement, and a representation change to disclose), a class that quietly became unevaluable because ferro stopped accepting a member (self-disarming), and a class deleted outright.

**Twelve classes: 2 seeded from the issue, 10 found by search.** The search swept the spec for its equivalence phrasing — "alternatively", "can also be described as", "giving an alternative description like", "equivalent to", "the same variant". Results:

| class | verdict |
|---|---|
| `dna-delins-vs-aligned-split-850-901` (`delins.md:44-47`) | non-confluent |
| `dna-delins-vs-two-substitutions-235-237` (`delins.md:19`) | non-confluent |
| `dna-delins-vs-two-substitutions-145-147` (`delins.md:37`, `:42`) | non-confluent |
| `rna-delins-vs-two-substitutions-142-144` (`RNA/delins.md:40`) | non-confluent |
| `dna-insertion-n-run-vs-repeat-count-761-762` (`DNA/insertion.md:74`) | non-confluent |
| `rna-insertion-n-run-vs-repeat-count-761-762` (`RNA/insertion.md:38`) | non-confluent |
| `rna-repeat-range-vs-unit-124` (`RNA/repeated.md:31`) | non-confluent |
| `rna-repeat-range-vs-unit-124-two-alleles` (`RNA/repeated.md:36`) | non-confluent |
| `protein-insertion-xaa-run-vs-repeat-count-582-583` (`protein/insertion.md:60`) | non-confluent |
| `protein-extension-stop-glyph-110` (`protein/extension.md:30`) | **confluent** |
| `protein-extension-stop-glyph-327` (`protein/extension.md:33`) | **confluent** |
| `protein-position-less-repeat-shorthand-gln21` (`RNA/repeated.md:52`) | not-evaluable |

The two confluent ones are deliberate **positive controls**: `Ter` and `*` are co-equal stop spellings and ferro already converges them, so the assertion demonstrably can pass. Without them, a change that made every class disagree could hide in a wall of expected failures.

Three findings from the search worth calling out on their own:

1. **`dna-delins-vs-two-substitutions-145-147` overlaps the `false-acceptance` census and says something it cannot.** The census records that ferro accepts a spelling the spec marks invalid; the class records that ferro also keeps it as a *second fixed point* for a variant it already has a form for.
2. **The spec's own layout is off at `delins.md:42`.** That `NOTE` sits under the `c.9002_9009delinsTTT` bullet but the variant it names belongs to the `:37` bullet — `c.145_147` is `CGC`, and `145C>T` + `147C>G` gives `TGG`, i.e. `c.145_147delinsTGG`. The class cites both lines for that reason.
3. **`DNA/insertion.md:74` writes its two spellings under different accession versions** (`NM_004006.2` / `NM_004006.1`) while describing a single insertion.

Two families were found and deliberately **not** seeded, because they are reference-count-dependent and so not hermetically evaluable: `RNA/repeated.md:33-34` and `protein/repeated.md:22-23` ("when the reference sequence has 15 units, `r.-123ug[14]` is preferred over `r.-97_-96del`"). Worth revisiting once the manifest-backed loader can serve them.

## Half two — ruling records

A sibling `rulings` section records how the project reads clauses that conflict **at equal RFC 2119 strength** (`style.md:9` binds the spec to RFC 2119; items 3 and 4 make RECOMMENDED == SHOULD and cover SHOULD NOT). Each record names the clauses in tension, which one governs, which is deviated from, and why — all citations quote-verified against the checkout.

**`undecided` is a first-class state, enforced.** The generator *refuses to build* a record that is `undecided` yet names a governing or deviated-from clause, and `ruling_records_are_intact` re-asserts it from the committed side and additionally requires an undecided record to name at least two clauses. An unsettled question cannot acquire a ruling nobody made, and an inconvenient record cannot be deleted or re-statused without a reviewable diff.

Seeded with three:

1. **`delins-merge-vs-individual-gap-two-or-more` — undecided.** `delins.md:17` (SHOULD NOT merge separated variants) against `delins.md:47` ("the delins format is recommended"). Both reach the `:44-47` example, whose own alternative splits at gaps of 4, 5 and 3 — all at least 2 — so `:17` read literally demands the very split `:47` rejects three lines later. **No ruling has been made and none is implied here.** What the record establishes is the cost: this reading governs roughly 138 of 208 adjudicated corpus rows, so settling it either way is a migration. An operator ruling closes it.
2. **`inversion-vs-two-substitutions-76-83` — undecided.** `general.md:34` against `general.md:56` on `c.76_83inv`. `NM_004006.2:c.76_83` is `AATGCACA`; its reverse complement `TGTGCATT` agrees at 4 of 8 columns, so the same edit reads as one inversion or as two changed runs separated by unchanged bases. Under active adjudication; recorded so the question is visible rather than re-derived.
3. **`delins-codon-carve-out-gap-one` — decided.** `delins.md:18`'s explicit exception for two variants one nucleotide apart affecting one amino acid. The spec settles this one: `:18` is an exception to `:17` stated in the same place with the same scope, so there is no conflict to weigh and nothing to deviate from. The record notes `general.md:36-39`, where the SVD-WG says it intends to replace the carve-out with a purely positional rule, which would widen it.

## Verification

Both generator gates and the test guards were negative-tested rather than assumed:

- corrupting a citation quote → generation fails, printing the expected quote and what the cited lines currently read;
- adding `governing` to an `undecided` ruling → generation fails ("record the conflict, not a ruling nobody made");
- renaming a class member to a non-row → generation fails ("not a fixture row (typo or spec drift)");
- flipping one pinned verdict → `spec_equivalence_classes_converge` fails, naming the members and their two distinct outputs.

Row census unchanged: 934 rows, `preserved` 714 / `correctly-rejected` 99 / `false-acceptance` 52 / `needs-reference` 50 / `diverges` 19.

Full gate, rebased onto `92fd20f4`:

```
Summary [  84.377s] 9571 tests run: 9571 passed (1 slow), 145 skipped
```

`cargo fmt --check` and `cargo clippy --features dev --all-targets -- -D warnings` are clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -256,7 +256,30 @@ Two test modules actually read it — `tests/it/hgvs_spec_normalization_tests.rs
 |---|---|---|
 | `ferro_produces_the_form_the_spec_states` | `hgvs_spec_normalization_tests.rs` | ferro's output vs the **spec's** stated form, plus `KNOWN_DIVERGENT_INPUTS` |
 | `status_census_is_unchanged` | `hgvs_spec_normalization_tests.rs` | per-status row counts |
+| `spec_equivalence_classes_converge` | `hgvs_spec_normalization_tests.rs` | one normalized output per curated equivalence class, plus `EQUIVALENCE_CLASS_VERDICTS` |
+| `ruling_records_are_intact` | `hgvs_spec_normalization_tests.rs` | the ruling records' ids and decided/undecided statuses (`RULING_STATUSES`) |
 | `DIVERGENCE_BUDGET` / `PASSING_CENSUS` | `spec_enumeration_tests.rs` | per-status row counts, together totalling every row |
+
+**Statuses only ask whether ferro *changed* a string — never whether two rows are the same
+variant.** That is why the spec's own non-confluent pair
+(`LRG_199t1:c.850_901delinsTTCCTCGATGCCTG` and the `c.[850_869del;…]` split it names as that
+variant's "alternative description", `DNA/delins.md:44-47`) sat in the fixture as two `preserved`
+rows and passed. The `equivalence_classes` section of
+`hgvs_spec_normalization_overrides.json` closes that: it declares which harvested inputs denote
+one variant, and `spec_equivalence_classes_converge` fails when a class yields more than one
+output. The disagreements it currently finds are **expected and pinned** — converging them is
+#1235's job and a representation change for a downstream consumer, not something to fix in a test
+PR. Members that cannot be evaluated hermetically (`needs-reference`, or a description ferro
+refuses) are skipped rather than compared, so an error string never counts as a representation.
+
+The sibling `rulings` section is the decision log for clauses that **conflict at equal RFC 2119
+strength** (`style.md:9` binds the spec to RFC 2119; items 3 and 4 make RECOMMENDED == SHOULD and
+cover SHOULD NOT). Each record names the clauses in tension, which one governs, which is deviated
+from, and why. `undecided` is a first-class state and the generator **refuses** to build a record
+that is `undecided` yet names a governing clause — an unsettled question must not be able to
+smuggle in a ruling nobody made. Every citation carries a verbatim quote that the generator checks
+against the spec checkout, so a submodule bump that moves a clause fails the build instead of
+leaving the citation pointing at unrelated prose.
 
 Requires the `assets/hgvs-nomenclature` submodule (`git submodule update --init assets/hgvs-nomenclature`); without it the generator fails with `no HGVS strings harvested from …`, naming that command.
 
--- a/examples/generate_spec_fixture.rs
+++ b/examples/generate_spec_fixture.rs
@@ -70,7 +70,8 @@ fn run(cli: &Cli) -> anyhow::Result<()> {
     let candidates = sources::discover(&cli.spec_dir)?;
     let overrides = overrides::load(&cli.overrides)?;
     let rows = runner::build_rows(&candidates, &overrides)?;
-    let rendered = render::render(&rows, &cli.spec_dir)?;
+    let decisions = decisions::resolve(&rows, &overrides, &cli.spec_dir)?;
+    let rendered = render::render(&rows, &decisions, &cli.spec_dir)?;
 
     if cli.check {
         check_up_to_date(&cli.output, &rendered, "fixture", "generate_spec_fixture")?;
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

PR #1520, opened 2026-08-07
test(msto): pin the reported family's expected outputs and link its guards

### LINKED ISSUE #1430: Sequence coercion before normalization
Relates to #1235 

## Summary

My understanding is that ferro's normalization operates exclusively in allele space. Ferro decomposes the input variant into its constituent alleles, normalizes each individually, and then attempts to coalesce them back into a single multiallelic variant. The problems we've been seeing in #1235 and related issues have seemed attributable to trying to build a rules-based classifier around the combinatorially large number of possible ways these variants can overlap and interact.

Mutalyzer gets confluent output by converting the input variant to its corresponding sequence, and then deriving a normalized variant representation from that sequence. It's not guaranteed to be spec compliant (they take some shortcuts that merge deletion-insertions), but overall it's deterministic and confluent - given equivalent variants, you get the same sequence, and given that sequence, the normalizer's rules always create the same variant.

So I'm wondering if a hybrid/multi-pass approach would suit ferro.

1. First, we do a mutalyzer-like coercion of the input variant to its corresponding sequence. This should be deterministic and confluent - any two equivalent variants should produce the same corresponding sequence. 
2. Next, we convert this sequence to a representative variant. This should be deterministic - a sequence should always be converted to the same variant.
3. Optionally, we apply the existing normalizer to this variant, to get a spec compliant output. This should be deterministic and idempotent - normalizing a variant should always produce the same canonical representation, and re-normalizing that representation should yield the same.

(3) is described as optional because I'm not certain whether it'd make sense to combine (2) and (3) - there are probably some performance optimizations to be gained by doing so in the long run, but in the short term implementing (1) and (2) and then applying our existing normalizing logic will probably get us to where we need to be for a v1. Importantly, this means we can implement (2) without worrying about an algorithm that yields a spec-compliant variant at the same time.

nh13:
The msto regression corpus catalogued 73 issues and left 31 with `tests: &[]`, four of them open and `high`: #1419, #1420, #1421 and #1430. This links all four, pins the answers that were only being counted, and records — per row — what the issue asked for, what ferro does, and why they differ.

No normalizer code changes. Nothing was re-blessed. Every assertion added here already held on `main`.

## What was actually wrong for #1419/#1420/#1421

Not coverage — attribution. `tests/it/reported_confluence_pairs.rs` and `tests/it/reported_partition_verdicts.rs` guard those three issues end to end, but they are table-driven: the issue numbers live in data rows (`"1419-r1/cis"`) and module docs, never in a function name or the comment above one, which is what the catalog's attribution rule requires. So three issues with eighteen pinned expectations between them were reported as unguarded.

The rule is widened the way it already was for the Python guards, not bent: **where a module is one issue family end to end and says so in its module doc, the module doc is the attribution.** It is claimed only for modules where that is true, and it does not lower the bar for #200, which stays in `no_high_priority_issue_is_unguarded_without_being_named` because no module anywhere is about it.

## Which expectations are now hard, and which are not

**Newly hard (36 pins, all measured):**

- Every one of the eighteen reported spellings now has its **5'** answer pinned per row (`Row::five_prime`, `every_reported_spelling_normalizes_as_recorded_under_five_prime`). Nothing pinned it before — `reported_confluence_pairs` sweeps both directions but only *counts* convergence, and `reported_partition_verdicts` pinned 3' alone. Sixteen rows agree across directions; **two do not** (`1419-r3/cis` → `[18_23del;28_33del]`, `1420-v2/cis` → `[36dup;41del]`), and both are `Gap` rows the 3' pass returns verbatim. That matters because `every_gap_row_is_returned_exactly_as_authored`'s argument — a spelling returned *untouched* is a second canonical form, not an under-applied pass — is a 3'-only argument. `only_the_named_rows_answer_differently_in_the_two_directions` pins the set so it cannot drift.
- Every row now carries `wanted`: the output **its issue asks for, in the issue's own terms**. `the_wanted_form_of_every_gap_row_is_its_siblings_output` turns the prose claim into a checked fact — in all nine pairs the wanted form is exactly what ferro already prints from the *other* spelling, which is what makes this spelling-dependence rather than a missing rule. `every_canonical_row_already_prints_its_wanted_form` pins the other half.

**Deliberately still a census:** `CONVERGING_PAIRS = 0` in `reported_confluence_pairs`. It stays a ratchet, because total convergence is the goal and partial progress should read as one number moving rather than nine file edits.

**Deliberately still recorded rather than targeted: #1419's six rows.** Applying the project's precedence policy (**spec-explicit > Mutalyzer > our judgement**) row by row does not give one answer for the family:

| issue | authority | why |
|---|---|---|
| #1420 v2/v3, #1421 n1–n3 | spec-explicit | `delins.md:17` (separated → individual) with nothing in the spec against it. `delins.md:44-47` cannot reach them: it is conditioned on the span losing length, and #1421's rows are net insertions while #1420 v2/v3 are length-neutral. |
| #1420 v4 | spec-explicit | `delins.md:16` — all four positions change consecutively. |
| #1419 r1–r3 | **spec self-conflicting** | The issue rests on `general.md:56` + `delins.md:17`. But `delins.md:44-47` addresses this exact shape — a payload whose bases align with the reference, giving "an alternative description like `c.[850_869del;874_881del;887_897del;901_902insG]`" — and says outright that **the delins format is recommended**. #1419's wanted `[19_30del;33T>G]` is a del-plus-sub arising from precisely that coincidence. |

Where the spec conflicts with itself the policy hands the tie-break to Mutalyzer, so Mutalyzer was measured rather than assumed. It cannot resolve the synthetic `TEMPLATE` accession, so each *shape* was reproduced on a public accession whose bases were read back out of Mutalyzer's own `ESEQUENCEMISMATCH` detail (`NG_012337.1:g.5001_5120`) and both spellings normalized through the public API. Mutalyzer is **confluent on every shape** — and picks the **merged delins** in all three, which is neither ferro's answer nor the issue's for #1419. Its merge is also not a rule one could adopt: two substitutions two nucleotides apart merge for `T>A`/`C>A` and `T>G`/`C>T` and stay individual for `T>C`/`C>G`.

So #1419's rows stay characterization pins with the disagreement written into the row, and `the_spec_authority_census_holds` pins the 12/6 split so moving a row between the two has to be argued for.

## #1430 is measured by a property, not by an output

It is a design proposal — coerce to sequence, derive a representative variant, optionally apply the normalizer for spec compliance — and states no input/expected pair, so it cannot have a conventional regression test. Its slice names **one guard per stage** (`ISSUE_1430_STAGE_GUARDS`), and `issue_1430_is_measured_by_a_property_not_by_an_expected_output` pins that reading so the entry cannot quietly be "fixed" into an invented input/expected row.

## The collision, documented and not resolved

#1430 names Mutalyzer's behaviour as *"not guaranteed to be spec compliant (they take some shortcuts that merge deletion-insertions)"* — the same delins-merging used above to adjudicate #1419. Where this PR follows Mutalyzer onto a merged delins, the issue author's stated view points the other way. #1430's own stage (3) reconciles them, and `merge::coalesce` already cites #1430 for that split, but the reconciliation is a design and not a decision that has been taken. It is written down in the module doc rather than settled here.

## Pinned counts

`PINNED_UNIQUE_TEST_COUNT` 265 → **288**, and `no_high_priority_issue_is_unguarded_without_being_named` now expects `[200]` alone. Both move because the catalog can now see guards that already existed — no test was written to move either number.

## Also recorded: a citation is not a guarantee

An audit of every cataloged `(file, fn)` pair found two rows where the issue number is a **cross-reference**, not the test's subject: #1034 → `issue_291_rna_axis_convention.rs::rna_canonical_split_fetch_is_cds_relative` (a #469 claim; #1034 explains why a full-run inversion was used), and #310 → `transcript.rs::test_transcript_default_is_minimal_and_non_coding` (pins `Transcript::default()`; #310 is an anecdote). Both satisfy the mechanical rule, so they are documented rather than deleted, with the caveat that a populated slice means *"a test names this issue"* and not *"this issue is covered"*. The empty slices — the load-bearing half — are unaffected: a cross-reference can inflate coverage, never hide a gap.

## Representation stability

No normalized output moves. The two 5'-direction answers now pinned are what `main` already produces; they were simply unpinned.

Relates to #1419, #1420, #1421, #1430.

### DIFF
--- a/tests/it/msto_regression_corpus.rs
+++ b/tests/it/msto_regression_corpus.rs
@@ -43,6 +43,76 @@
 //! guard may still have an unlisted Python one; that understates coverage
 //! without misreporting the issue as unguarded.
 //!
+//! ## Table-driven guards count, and used not to
+//!
+//! The same false negative, in a second shape. `tests/it/reported_confluence_pairs.rs`
+//! and `tests/it/reported_partition_verdicts.rs` guard **#1419, #1420 and
+//! #1421** — every row in both is one of those issues' reported spellings — but
+//! the issue numbers live in the tables' *data rows* (`"1419-r1/cis"`,
+//! `"1421-n3/span"`) and in the module docs, never in a function name or in the
+//! comment above one. So the per-function rule could not see them, and all three
+//! were listed as unguarded while eighteen pinned expectations and four
+//! whole-corpus properties covered them.
+//!
+//! The accommodation is the Python one generalized rather than a new exception:
+//! **where a module is one issue family end to end and says so in its module
+//! doc, the module doc is the attribution.** That is sound for the same reason it
+//! was sound for pytest — the unit of authorship is the module, not the function
+//! — and it is *only* claimed for modules where it is true. A grab-bag suite
+//! that happens to mention an issue in its header still needs the per-function
+//! reference, because there the module doc says nothing about any individual
+//! test.
+//!
+//! Note what this does not do: it does not lower the bar for **#200**, which
+//! stays in [`no_high_priority_issue_is_unguarded_without_being_named`]. No
+//! module anywhere is about #200, so there is nothing for a module-doc rule to
+//! find. The rule widened to admit guards that exist; it did not widen until the
+//! finding disappeared.
+//!
+//! ## An issue whose satisfaction is a property, not an output
+//!
+//! **#1430** is the one entry here that could never have a conventional
+//! regression test, and its slice is populated on a different basis from every
+//! other row. It is a design proposal — coerce a description to its denoted
+//! sequence, derive a representative variant from that sequence, then optionally
+//! apply the normalizer for spec compliance — and it states no input/expected
+//! pair at all. There is no string to pin.
+//!
+//! What it can be measured by is whether each of its three stages holds as a
+//! property, so its slice names one guard per stage rather than a reproduction.
+//! See [`issue_1430_is_measured_by_a_property_not_by_an_expected_output`], which
+//! pins that reading so it cannot quietly be replaced by an ordinary
+//! input/expected row.
+//!
+//! ## A citation is not a guarantee that the test guards the issue
+//!
+//! The attribution rule is mechanical — `#<N>` in the function's name or in the
+//! comment above it — and that is deliberate, because a judgement-based rule
+//! rots faster than the thing it describes. But mechanical means it also catches
+//! *cross-references*: a test whose comment names an issue to explain why the
+//! fixture was built a certain way, or as a historical aside, reads identically
+//! to a test that reproduces it.
+//!
+//! An audit of every cataloged pair found two such rows, and they are recorded
+//! here rather than removed, because both satisfy the rule as written and
+//! deleting them would leave the table no longer a faithful application of its
+//! own stated rule:
+//!
+//! * **#1034** lists
+//!   `tests/it/issue_291_rna_axis_convention.rs::rna_canonical_split_fetch_is_cds_relative`.
+//!   That test pins that the r. reference window is fetched CDS-relative
+//!   (a #469 claim). #1034 appears only as the reason a *full-run* inversion was
+//!   used to build the case rather than a sub-run.
+//! * **#310** lists
+//!   `src/reference/transcript.rs::test_transcript_default_is_minimal_and_non_coding`.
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1523, opened 2026-08-07
ci: require a Representation-Change declaration on output-bearing changes

### LINKED ISSUE #1522: CI: enforce the Representation-Change trailer on output-bearing changes
`release-plz.toml` renders a **Representation changes** changelog section from a `Representation-Change:` commit trailer, and the release PR body carries a reviewer checklist for it (added in #1433). The machinery works. Nothing feeds it.

## Measured

For the v0.13.0 cycle:

- **87 commits** since `v0.12.0`; **0** carry the trailer.
- **46** of them touch `src/{normalize,hgvs,spdi,project}/`.
- Restricting to commits merged **after** the mechanism landed (`fb63eae6`), so the mechanism demonstrably existed: **47 commits, 25 touching those directories, still 0 trailers.**

The section therefore rendered empty, and the release was one merge away from shipping with no disclosure at all. It was reconstructed by hand for v0.13.0 — normalizing ClinVar 500k, Paraphase and CMRG (5,761,302 expressions) through both releases against the same prepared reference and diffing row by row. That found **577 stored strings moved**, **326,404 inputs newly normalized** (all LRG), and **2 now correctly rejected**. None of that would have been visible from the changelog as generated.

## Why an empty section is not self-announcing

`release-plz.toml` emits the section unconditionally, on the stated grounds that this turns an undeclared change *"from silence into a visible false statement, which is the only version a reviewer can act on."* That reasoning is right, but it only pays off if somebody acts on the false statement — and an empty section looks identical whether it is empty because nothing moved or because nobody declared anything. This cycle it went unnoticed for ~90 commits.

## Proposal

A CI check that fails, or at minimum warns on the PR, when a commit touches `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` and carries no `Representation-Change:` trailer.

Design notes:

- **Not every such commit moves output.** A comment-only change (#1521 is a clean example: 46 changed lines in `src/normalize/merge.rs`, zero non-comment) must be able to opt out. An explicit `Representation-Change: none` — declared, not absent — keeps the decision on the record and distinguishes "considered and no" from "never considered".
- **The trailer belongs on the squash commit**, which is what git-cliff reads as the footer; a trailer on an intermediate commit is discarded by the squash. So the check has to look at the PR's eventual squash subject/body, or the PR body, rather than at individual commits.
- **`git.conventional_commits` must stay enabled** for the footer rule to match at all. `release-plz.toml` records that turning it off makes the grouping fail silently rather than error — worth asserting in the same check.

## Acceptance

- A PR touching those directories without a trailer is visibly flagged before merge.
- `Representation-Change: none` is accepted as a declaration.
- The check itself has a test.

nh13:
Closes #1522.

`release-plz.toml` renders the changelog's **Representation changes** section from a `Representation-Change:` trailer on the squash commit, and the release PR body carries a reviewer checklist for it (#1433). The machinery works. Nothing fed it.

## What went wrong, measured

| | commits | touching `src/{normalize,hgvs,spdi,project}/` | carrying a trailer |
|---|---|---|---|
| whole v0.13.0 cycle | 87 | 46 | **0** |
| after the mechanism landed (`fb63eae6`) | 47 | 25 | **0** |

The section rendered empty, and the release was one merge from shipping with no disclosure at all. It had to be reconstructed by hand — normalizing ClinVar 500k, Paraphase and CMRG (5,761,302 expressions) through v0.12.0 and v0.13.0 against the same prepared reference and diffing row by row. That found **577 stored strings moved**, **326,404 inputs newly normalized** (every one an LRG accession), and **2 now correctly rejected**. None of it was visible from the changelog as generated.

**An empty section cannot announce itself.** `release-plz.toml` emits it unconditionally on the stated grounds that this turns an undeclared change *"from silence into a visible false statement, which is the only version a reviewer can act on."* That reasoning is right, but it only pays off if someone acts on the false statement — and an empty section looks identical whether it is empty because nothing moved or because nobody declared anything. This cycle it went unnoticed for ~90 commits.

## What this adds

A `Representation change declared` job that fails a PR touching those four directories when the PR description carries no `Representation-Change:` trailer.

**`Representation-Change: none` passes.** The point is to force the judgement, not to force a disclosure — #1521 is the real example of a legitimate decline: 46 changed lines in `src/normalize/merge.rs`, every one a comment. What is rejected is *silence*, because an absent trailer is indistinguishable from an unconsidered one.

Design points worth review:

- **The trailer is read from the PR description, not from individual commits.** The squash commit body is what git-cliff's footer rule actually matches; a trailer on an intermediate commit is discarded by the squash. Checking commits would pass PRs whose declaration never reaches the changelog.
- **The PR body reaches the script through `env`, never `${{ }}` inside `run`.** A PR description is attacker-controlled text and interpolating it into a shell body is a script-injection hole.
- **`pull_request` only.** There is no PR description to read on a push to `main`, and by then the declaration is already merged.
- **The watched set is pinned against `release-plz.toml`.** `test_watched_prefixes_match_the_release_config` asserts each watched directory is still named in the config's checklist, as a per-directory membership check rather than a count — two compensating edits could hold the count while swapping a directory out, which is exactly the drift that would make this check pass on the changes it exists to catch.

## Verification

The test suite found a real bug in the first draft of the regex, which is left in as a regression test. `^\s*Representation-Change:\s*(?P<value>.+?)\s*$` under `MULTILINE` matches a **lone space** as the value, so a bare `Representation-Change:` with nothing after it counted as a declaration — precisely the hole this check exists to close. `\s` also spans newlines, so a permissive class would scavenge the value off the *following* line. Both are now pinned (`test_an_empty_trailer_value_is_not_a_declaration`, `test_an_empty_trailer_does_not_scavenge_the_next_line`).

Negative-tested end to end: a simulated `src/normalize/merge.rs` + `src/spdi/mod.rs` change with no trailer exits 1 and names both files; the same change with `Representation-Change: none` exits 0.

22 tests pass. `actionlint` clean on the modified workflow. `ruff check` and `ruff format --check` clean.

**This PR touches no watched directory**, so its own declaration is `none` — and the check confirms that against its real changed-file list rather than an empty one. (An earlier run of that self-check was vacuous: `git diff origin/main...HEAD` returned nothing because the files were not yet committed, so it passed on an empty list rather than on unwatched files. Re-run after committing.)

Representation-Change: none

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,12 @@ on:
   # GitHub still rendered "All checks have passed" — a green badge that
   # carried no information. Run on every pull request whatever it targets.
   pull_request:
+    # `edited` is load-bearing for `representation-change` below: the declaration
+    # lives in the PR description, and the default activity set
+    # (opened/synchronize/reopened) does not fire when a description is edited. Without
+    # it, adding the trailer a red check just asked for re-runs nothing and the check
+    # stays red until an unrelated push.
+    types: [opened, synchronize, reopened, edited]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -158,6 +164,56 @@ jobs:
         run: uv run poe check-format
       - name: Run mypy
         run: uv run poe check-typing
+
+  # A change under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` can move
+  # the normalized string a downstream consumer stores read counts against, so it must say
+  # whether it does. `release-plz.toml` renders the changelog's **Representation changes**
+  # section from a `Representation-Change:` trailer on the squash commit (#1433) — the
+  # machinery works, but across the v0.13.0 cycle 87 commits landed carrying zero trailers
+  # while 46 touched those directories, and the section rendered empty. An empty section
+  # looks identical whether nothing moved or nobody declared, so it went unnoticed for ~90
+  # commits and the disclosure had to be reconstructed by hand before release (#1522).
+  #
+  # `Representation-Change: none` passes. The point is to force the judgement, not to force
+  # a disclosure; what this rejects is silence.
+  representation-change:
+    name: Representation change declared
+    runs-on: ubuntu-latest
+    # `pull_request` only. The check reads the PR description, which is what becomes the
+    # squash commit body and therefore what git-cliff's footer rule matches. A push to
+    # `main` has no equivalent to read, and by then the declaration is already merged.
+    if: github.event_name == 'pull_request'
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1  # v6.3.0
+        with:
+          python-version: "3.12"
+      - name: Collect the changed files
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          # `--no-renames` so a rename reports BOTH paths. With git's rename detection
+          # on (the default), moving `src/normalize/foo.rs` out of a watched directory
+          # reports only the destination, so the change that removed it from
+          # `src/normalize/` would need no declaration -- exactly backwards.
+          git diff --name-only --no-renames "$BASE_SHA...$HEAD_SHA" > changed-files.txt
+          echo "changed files:"
+          cat changed-files.txt
+      - name: Capture the PR description
+        # Through `env`, never interpolated into the script body: a PR description is
+        # attacker-controlled text and `${{ }}` inside `run` would splice it into the shell.
+        env:
+          PR_BODY: ${{ github.event.pull_request.body }}
+        run: printf '%s' "$PR_BODY" > pr-body.txt
+      - name: Require a Representation-Change declaration
+        run: |
+          python scripts/check_representation_change.py \
+            --changed-files changed-files.txt \
+            --declaration-file pr-body.txt
 
   clippy:
     name: Clippy
--- a/scripts/check_representation_change.py
+++ b/scripts/check_representation_change.py
@@ -1,0 +1,148 @@
+#!/usr/bin/env python3
+"""
+Require a `Representation-Change:` trailer on changes that can move normalized output.
+
+Normalized output is a shipped key: a downstream consumer stores read counts against
+the normalized HGVS string, so a canonical form that moves between releases re-buckets
+what they have already stored. `release-plz.toml` renders a **Representation changes**
+changelog section from a `Representation-Change:` trailer on the squash commit, and the
+release PR body carries a reviewer checklist for it (#1433).
+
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
Dual-review sweep applied. Three of five CLI findings taken, two rejected with reasons, plus one limitation found by my own pass that neither lens raised.

## Applied

**1. `--no-renames` on the diff (Major).** This one was a real hole and it disproves something I said earlier in review. Verified against a scratch repo:

```
$ git mv src/normalize/old.rs docs/new.md
$ git diff --name-only HEAD~1 HEAD          # rename detection ON (the default)
docs/new.md
$ git diff --name-only --no-renames HEAD~1 HEAD
docs/new.md
src/normalize/old.rs
```

So a file *leaving* a watched directory was reported only at its destination, and the change that removed it from `src/normalize/` needed no declaration — exactly backwards. Confirmed against the checker itself: destination-only input passes, both-paths input fails and names `src/normalize/old.rs`.

**2. `types: [opened, synchronize, reopened, edited]` (Minor, but load-bearing here).** Bare `pull_request:` defaults to `opened, synchronize, reopened`. The declaration lives in the PR *description*, so without `edited` a contributor who adds the trailer this check just asked for re-runs nothing and the check stays red until an unrelated push. That is a bad first experience for a new gate.

**3. Reject `-` for both inputs (Minor).** Reading both from stdin gives the second an empty string, which for the declaration reads as "no trailer" — a wrong *failure* rather than an error. Now `parser.error`, with a test pinning exit code 2.

23 tests pass, `actionlint` clean, `ruff check`/`format` clean.

## Rejected

**Move the job to a trusted base-branch workflow (Major).** The premise does not hold here. This is `pull_request`, not `pull_request_target`, so a fork PR gets a read-only token and no secrets; workflow `permissions` is `contents: read`; `persist-credentials: false`. Executing a PR's own script under those conditions is what every other job in this file already does — `clippy` and `test` compile and run PR-authored *Rust*, a far larger surface than one Python script. The fix would also break the check: a trusted workflow runs the *base*'s copy of the script, so a PR improving the checker could not exercise its own change. The injection risk it gestures at is `${{ }}`-inside-`run`, which the job already avoids by routing the PR body through `env`.

**Assert set *equality* between `WATCHED_PREFIXES` and `release-plz.toml` (Major).** The reverse direction cannot be done by substring matching without false failures — the config's prose mentions `src/` paths that are not watched prefixes, so an equality assertion would fail on its own comments. The direction that matters is the one asserted: every watched prefix must still be named in the config, so a directory cannot silently drop out of the config's scope while the check keeps watching it.

## Limitation, not raised by either lens

**This check is advisory, not blocking.** `main`'s ruleset requires exactly eight contexts — `abi3 floor ↔ requires-python`, `Build`, `Clippy`, `Clippy (all features)`, `Format`, `Python Lint`, `Python Wheel Test`, `Test` — and `test-required` aggregates only `[test, test-oracle, soak, sweeps, generated-docs]`. `Representation change declared` is in neither, so a PR with a missing trailer gets a **visibly red job that nothing prevents merging past**, the same posture `CLAUDE.md` documents for the oracle jobs.

That still meets #1522's acceptance criterion as written ("visibly flagged before merge"), and flagging is arguably the right first step for a brand-new gate — a false positive that blocks the queue would be worse than one that nags. Making it enforce means adding the context to the branch ruleset, which is a repo-settings change, not a code change.

Second, smaller: prefix matching is anchored at the repo root, so `crates/*/src/normalize/` would not match if this ever becomes a workspace.

PR #1525, opened 2026-08-07
chore: release v0.13.1

nh13:
## Representation stability

Any PR that moves a normalized output should carry a `Representation-Change:`
trailer (see CONTRIBUTING.md). Those are collected into the **Representation
changes** section of the changelog below.

Reviewer checklist:

- [ ] Either that section lists every output-moving change in this cycle, or
      there were none and it is absent.
- [ ] Nothing in it merely *declined*. A decline reads `Representation-Change:
      none`, optionally followed by a reason, and belongs under its own type —
      not here. Spot-check two entries against their PR descriptions.
- [ ] For each entry, that PR's description says whether the affected inputs
      were previously **rejected** (no stored string, so free) or previously
      **accepted** (a real migration for the consumer). The bullet below carries
      only the commit subject: the trailer's own text cannot reach the changelog
      (fulcrumgenomics/ferro-hgvs#1556), so this one has to be read from the PRs.

A commit under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` that
moved output without declaring it means the trailer is missing. Fix that before
releasing, not after.

---

### Representation changes

- *(normalize)* never split a delins into members on consecutive nucleotides ([#1537](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1537))
- *(normalize)* type the whole-block inversion by what it competes with ([#1535](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1535))

### Other

- *(changelog)* read a decline that gives a reason as a decline ([#1558](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1558))
- *(confluence)* record two adjudications from the divergent cis classes ([#1549](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1549))
- *(normalize)* guard that emitted members obey general.md:34 ([#1539](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1539)) ([#1540](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1540))
- *(spec)* run the spec's own worked examples against real bases, and record #1536 ([#1544](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1544))
- *(conformance)* gate the harvested unguarded cases on every PR ([#1545](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1545))
- *(normalize)* validate a cis-allele member's stated reference bases before a merge can consume them ([#1547](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1547))
[...]

PR #1527, opened 2026-08-07
fix(changelog): stop grouping a declined change as a representation change

### LINKED ISSUE #1526: Representation-Change: none is grouped as a representation change
`Representation-Change: none` — the declaration that says a change moves **no** output — is grouped by git-cliff into the changelog's **Representation changes** section, exactly as a genuine disclosure would be.

The footer rule matches the trailer's *presence*, not its value:

```toml
{ footer = "^Representation-Change:", group = "<!-- 0 -->Representation changes" },
```

## Observed

#1523 is a CI-only change: it touches `.github/workflows/ci.yml`, `scripts/` and `tests/python/`, no `src/` at all, and correctly declared `Representation-Change: none`. The v0.13.1 release PR (#1525) renders:

```markdown
### <!-- 0 -->Representation changes

- require a Representation-Change declaration on output-bearing changes (#1523)
```

A consumer reading that goes looking for keys to re-map and finds none.

## Why this matters more than it looks

This is the same failure this section exists to prevent, inverted. v0.13.0 **under-reported** — nothing declared, section empty, and the disclosure had to be reconstructed by hand from a 5.7M-row corpus diff (#1522). v0.13.1 **over-reports** — a non-change listed as a change. Both leave the section untrustworthy, and an untrustworthy section is one nobody reads carefully, which is how the v0.13.0 gap survived ~90 commits.

It also compounds. Every future PR that correctly declines now files itself here, so the section fills with noise and real entries get harder to find — the declining case is expected to be the *common* one, since most changes under those directories do not move output.

## Fix

Add a `commit_parsers` rule **above** the current one that matches a declining value and routes it elsewhere:

```toml
{ footer = "^Representation-Change:\\s*(none|no|n/a|na)\\s*$", group = "<!-- 7 -->Other" },
{ footer = "^Representation-Change:", group = "<!-- 0 -->Representation changes" },
```

Ordering is load-bearing: git-cliff takes the first matching parser, so the exclusion must precede the inclusion. The decline vocabulary must stay in step with `NONE_VALUES` in `scripts/check_representation_change.py`, which is worth a test rather than a comment.

**Alternative considered and rejected:** a separate sentinel (`Representation-Change-None: true`) so declining never matches the footer at all. It works, but it splits one concept across two spellings, makes the checker's guidance message clumsier, and means a reader has to know both. Keeping one trailer and teaching the consumer about its values puts the value-awareness where the value is actually read.

## Acceptance

- `Representation-Change: none` does not appear under **Representation changes**.
- A trailer with a real description still does.
- The decline vocabulary is asserted to match `NONE_VALUES`, so the two cannot drift.

Blocks nothing, but #1525 should not ship its current changelog line.

nh13:
Closes #1526.

git-cliff matches a footer rule on the trailer's **presence**, not its value, so `Representation-Change: none` — the declaration that a change moves *no* output — was grouped into the changelog's **Representation changes** section exactly as a real disclosure would be. #1523 is a CI-only change that correctly declined, and it rendered under that heading in the v0.13.1 release PR (#1525).

## Why this is worth fixing before v0.13.1 ships

It is the v0.13.0 failure inverted. That release **under-reported** — nothing declared, section empty, and the disclosure had to be reconstructed by hand from a 5,761,302-row corpus diff (#1522). Listing non-changes costs the same trust from the other side: a consumer reading it goes looking for keys to re-map and finds none. And it compounds, because declining is the *common* case for changes under those directories, so the section would fill with noise until the real entries were hard to find.

## The fix

An exclusion parser above the inclusion. **Ordering is load-bearing** — git-cliff takes the first matching parser, so the exclusion must come first. `[^\S\r\n]` keeps the whitespace horizontal, so `Representation-Change:` followed by a line reading `none` is not read as a decline.

## A second bug, found while testing the first

Both rules are now `(?i)`. The inclusion rule was case-sensitive while `scripts/check_representation_change.py` accepts any casing — so a lowercase `representation-change: 577 rows move` is a **real disclosure** that got dropped silently into `Other`. That one predates this PR and #1523; it surfaced only because testing the exclusion against real git-cliff meant enumerating casings.

## Verification, against git-cliff 2.13.1 rather than a simulation

A scratch repo with four commits, run through the shipped parser config:

| commit trailer | rendered under |
|---|---|
| `Representation-Change: 577 rows move…` | **Representation changes** ✅ |
| `representation-change: 42 rows move` | **Representation changes** ✅ (was `Other` — the second bug) |
| `Representation-Change: none` | `Other` ✅ (was **Representation changes** — this bug) |
| `Representation-Change: NONE` | `Other` ✅ |

## Guards, negative-tested rather than assumed

Two halves of one decision that drift silently, so both are pinned — and each was confirmed to **fail** when its half is narrowed, not merely to pass as written:

- `test_decline_vocabulary_matches_the_changelog_config` — the config's exclusion vocabulary must equal `NONE_VALUES`. Narrowing the config to `(none|no)` → **CAUGHT**.
- `test_both_representation_change_rules_are_case_insensitive` — stripping `(?i)` from the inclusion → **CAUGHT**.

25 tests pass; `ruff check` / `format` clean.

## Note on #1525

The v0.13.1 release PR currently carries the bad line. Once this merges, release-plz regenerates it on the next push to `main` and the line goes away on its own — no hand-edit needed, unlike v0.13.0.

Representation-Change: none

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,6 +78,22 @@ Representation-Change: c.<cds_end>_*1ins<A> now renders as
   anyone normalizing to a fixed point sees no change.
 ```
 
+**If your change moves nothing, say that** — `Representation-Change: none` is a
+first-class answer, not a way of opting out:
+
+```
+ci: pin the workflow's action hashes
+
+Representation-Change: none
+```
+
+`none`, `no`, `n/a` and `na` all count, in any casing. A declining trailer is
+excluded from the changelog's **Representation changes** section, so declaring it
+costs the reader nothing while leaving the judgement on the record. CI enforces
+the distinction: a change touching `src/normalize/`, `src/hgvs/`, `src/spdi/` or
+`src/project/` with no trailer at all fails the `Representation change declared`
+check, because an absent trailer is indistinguishable from an unconsidered one.
+
 **Indent continuation lines**, as above. Git only folds a multi-line trailer
 value into the trailer when the continuations are whitespace-prefixed: measured
 against git-cliff 2.13.1, the unindented form still groups correctly (its footer
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -96,7 +96,32 @@ releasing, not after.
 # The trailer is what gets a note in automatically, at release time.
 [changelog]
 commit_parsers = [
-  { footer = "^Representation-Change:", group = "<!-- 0 -->Representation changes" },
+  # `Representation-Change: none` declares that a change moves NO output, so it must not
+  # be listed as one. git-cliff matches a footer rule on the trailer's *presence*, not its
+  # value, so without this exclusion #1523 -- a CI-only change that correctly declined --
+  # rendered under "Representation changes" in v0.13.1 (#1526).
+  #
+  # That matters as much as the opposite failure. v0.13.0 under-reported (nothing declared,
+  # section empty, disclosure reconstructed by hand from a 5.7M-row corpus diff, #1522);
+  # listing non-changes is the same loss of trust from the other side, and it compounds,
+  # because declining is the COMMON case for changes under those directories.
+  #
+  # Ordering is load-bearing: git-cliff takes the FIRST matching parser, so this exclusion
+  # must precede the inclusion below. `[^\S\r\n]` is horizontal-only whitespace -- `\s`
+  # would span the newline into the next trailer and let `Representation-Change:` followed
+  # by a line reading `none` count as a decline.
+  #
+  # `(?i)` because the checker accepts `NONE`/`None` as declines (its own regex is
+  # case-insensitive); without the flag an uppercase decline is grouped as a real change,
+  # which is this bug in miniature. Verified against git-cliff 2.13.1.
+  #
+  # The decline vocabulary is kept in step with `NONE_VALUES` in
+  # `scripts/check_representation_change.py` by `test_decline_vocabulary_matches_the_changelog_config`.
+  { footer = "(?i)^Representation-Change:[^\\S\\r\\n]*(none|no|n/a|na)[^\\S\\r\\n]*$", group = "<!-- 7 -->Other" },
+  # `(?i)` here too, and not only on the exclusion above: the checker accepts any casing,
+  # so a lowercase `representation-change: 577 rows move` is a REAL disclosure that a
+  # case-sensitive rule drops silently into `Other`. Verified against git-cliff 2.13.1.
+  { footer = "(?i)^Representation-Change:", group = "<!-- 0 -->Representation changes" },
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
Dual-review sweep applied. One CLI finding, taken — and it names the most fragile property in the change.

## Applied: assert the rule ordering (CLI, trivial — I'd rate it higher)

The comment in `release-plz.toml` says git-cliff takes the first matching parser, and the whole fix depends on it, but nothing enforced it. Swapping the two lines silently restores #1526 in full.

Verified against git-cliff 2.13.1 rather than assumed — with the exclusion second:

```
### <!-- 0 -->Representation changes
- a declining change      <- the bug, back
- a real move
- uppercase decline       <- and this one
- lowercase trailer name
```

**Every other guard in this file still passes in that state**, which is exactly why ordering needed one of its own. `test_the_decline_exclusion_precedes_the_inclusion` now asserts the first rule names a decline value and the second does not; negative-tested by swapping the two lines in the real config → **CAUGHT**.

## Found by my own pass, not by either lens: `CONTRIBUTING.md` was a third consumer

It documents the `Representation-Change:` trailer for contributors but never said that **declining is allowed**. A reader following it literally would either invent a disclosure for a change that moves nothing, or skip the trailer entirely and hit the new `Representation change declared` check with no idea what it wanted — a poor first encounter with a gate added two commits ago.

Now documented, with the decline vocabulary pinned: `none`/`no`/`n/a`/`na` appear in **three** places (`release-plz.toml`, `scripts/check_representation_change.py`, `CONTRIBUTING.md`) and all three are asserted equal.

## Guard inventory, each negative-tested

Every one was confirmed to **fail** when its half is broken, not merely to pass as written:

| guard | broken by | result |
|---|---|---|
| `test_decline_vocabulary_matches_the_changelog_config` | narrowing config to `(none\|no)` | CAUGHT |
| `test_both_representation_change_rules_are_case_insensitive` | stripping `(?i)` | CAUGHT |
| `test_contributing_documents_the_same_decline_vocabulary` | dropping two words from the doc | CAUGHT |
| `test_the_decline_exclusion_precedes_the_inclusion` | swapping the two rules | CAUGHT |

27 tests pass; `ruff check`/`format` clean.

## Worth recording about this whole line of work

Four defects have now been found in this machinery — the empty-value regex, the grouping bug, the casing bug, the doc gap — plus this unasserted invariant. **Every one surfaced through execution or by enumerating consumers; none by reading the code.** The casing bug in particular: I reasoned correctly about the exclusion regex and still shipped a case-sensitive inclusion rule that only appeared when four spellings were run through real git-cliff. Config-driven guards seem to need testing against the real tool as a rule, not as a courtesy.

PR #1528, opened 2026-08-07
ci: make the workflow linter requireable, and correct the branch-protection note

nh13:
`main`'s ruleset requires eight status contexts. This makes a ninth candidate — `zizmor + actionlint` — actually requireable, and corrects a comment that asserted the ruleset does not exist.

**`lint-workflows.yml`: drop the `paths:` filter on `pull_request`.** The filter meant the workflow never ran for a PR that touched nothing under `.github/`, and a workflow that never runs creates no check. To a required-status-check rule that is indistinguishable from a check that has not finished, so adding `zizmor + actionlint` to the ruleset left every non-workflow PR on "Expected — waiting for status to be reported" and unmergeable. Measured directly: the context was added to the ruleset and removed again minutes later for exactly this reason. The filter stays on `push`, where there is no merge to gate. Cost is one sub-minute job on PRs that changed no workflow.

**`audit-pr-runs.yml`: correct the header.** It said `main` is not a protected branch and that a required check was therefore a policy decision rather than an available fix. `main` does carry a ruleset with required contexts, and that ruleset already blocks the no-run failure mode the audit was written for. The audit keeps its real value — explaining *why* a PR is blocked, since a blocked-on-never-reported PR looks exactly like one whose CI is slow. The comment now also records the trap that produced the error: `gh api /repos/{owner}/{repo}/branches/main/protection` answers `404 Branch not protected` for a branch protected by a *ruleset*; `/rulesets` is the endpoint that sees it.

No behavior change to any check's contents — only when `Lint Workflows` runs. `actionlint` passes on both edited files; the representation-change checker reports "No file under a watched directory changed; no declaration needed."

Follow-up once this merges: add `zizmor + actionlint` to the ruleset's required contexts.

### DIFF
--- a/.github/workflows/audit-pr-runs.yml
+++ b/.github/workflows/audit-pr-runs.yml
@@ -24,12 +24,22 @@
 # workflow addresses the class rather than a cause: whatever the reason, a PR
 # whose head has never been built should be visible as a problem.
 #
-# Why a scheduled audit and not a required status check: a required check is
-# the stronger fix, since it blocks the merge rather than reporting after the
-# fact — but `main` is not a protected branch, so adopting one is a repository
-# policy decision rather than a workflow change. This is the part that can be
-# fixed in-repo, and it stays useful afterwards as the thing that tells you
-# *why* a merge is blocked.
+# Why a scheduled audit *as well as* required status checks. `main` does carry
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #1531, opened 2026-08-07
docs: record every adjudication as a test in the same change

nh13:
Adds a standing policy to `CLAUDE.md`: an adjudication — a decision about what the correct HGVS normalization behaviour is — must be recorded in the same change that makes it, in one of the four homes the repo already has for that purpose.

The load-bearing distinction is between a record and a change detector:

> A test that merely pins today's output is not an adjudication record. It is a change detector. What makes a record an adjudication is the *authority*: an exact `file:line` into `assets/hgvs-nomenclature`, a named Mutalyzer measurement, or an explicit "undecided". Without one, you have frozen the current behaviour including whatever is wrong with it.

Routing: clause-vs-clause conflicts to `rulings`; spellings that must converge to `equivalence_classes`; deliberate deviation to `KNOWN_DIVERGENT_INPUTS`; a settled input to an ordinary test. Three record kinds — adjudicated-correct, adjudicated-deviation, undecided — plus two rules that existing practice kept missing: record what was *refuted*, and deviating from the reference implementation needs a record rather than just a rationale.

Motivation is measurable rather than theoretical. An audit of the project's working record found roughly 200 adjudications, of which about 45 were recorded nowhere durable and 12 contradicted either each other or the code. The mechanisms already existed; what was missing was the requirement to use them.

Documentation only — no code, no fixtures, no behaviour.

Representation-Change: none

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -335,6 +335,69 @@ Contributor-facing guidance is in `CONTRIBUTING.md`. The checker is
 `tests/python/test_representation_change_trailer.py`, because three copies of one list drift
 silently.
 
+### Record every adjudication as a committed test or ruling record, in the same change
+
+**Policy.** When you decide what the *correct* normalization behaviour is — not how to
+implement it — that decision lands in a committed test or ruling record **in the same PR**, with
+a comment stating the question, the ruling, and the authority. An adjudication that lives only in
+a PR description, an issue comment, or a working document is lost: the next person re-derives it
+from scratch, and often re-derives it differently.
+
+This is cheap because the machinery already exists (the five committed guards are tabulated
+directly above). The policy is about using it consistently, not about building anything.
+
+**What counts as an adjudication:** a ruling that one clause governs another where they conflict;
+a determination that ferro's output is right or wrong against a cited clause; a decision to follow
+or deviate from Mutalyzer; a choice between two competing representations of one variant; or a
+question deliberately left open. Implementation choices, refactors and performance work are not
+adjudications and do not need records.
+
+**Where it goes:**
+
+| the adjudication is about | record it in |
+|---|---|
+| two spec clauses in tension | a `rulings` record in `hgvs_spec_normalization_overrides.json` |
+| two spellings that must converge on one output | an `equivalence_classes` entry + `EQUIVALENCE_CLASS_VERDICTS` |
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
Dual-review sweep. CLI raised 3, **all three confirmed**; my own pass adds nothing beyond them. Docs-only (`CLAUDE.md`, +62/-0), `Representation-Change: none` correct — no watched directory touched. CI green: 37 rows, 36 pass, 0 fail.

## Confirmed

**1. "all three guards are described directly above" — there are five.** The table at `CLAUDE.md:257-261` lists `ferro_produces_the_form_the_spec_states`, `status_census_is_unchanged`, `spec_equivalence_classes_converge`, `ruling_records_are_intact`, and `DIVERGENCE_BUDGET`/`PASSING_CENSUS`. Two of those (`spec_equivalence_classes_converge`, `ruling_records_are_intact`) arrived in #1519 and are exactly the machinery this policy leans on, so undercounting them undersells the argument the sentence is making.

**2. The heading says "as a test"; the policy says "test *or ruling record*".** The distinction is the section's own point — a `rulings` record with `status: undecided` is not a test and pins nothing about output, yet it is the required destination for a whole row of the table below. A heading that says "as a test" invites exactly the reading the section spends a paragraph refuting ("a test that merely pins today's output is not an adjudication record").

**3. The "Where it goes" table has no row for a deviation from Mutalyzer.** It lists four destinations, and its deviation row is scoped to "a deliberate, known deviation from **the spec's stated form**" -> `KNOWN_DIVERGENT_INPUTS`. But the section's closing paragraph says deviating from the reference implementation "needs a record, not just a rationale", and where the spec is silent or leaves the output undefined there is no spec form to diverge *from* — so `KNOWN_DIVERGENT_INPUTS` is the wrong home and the table names no other.

The destination does exist: #1532 records precisely this case as a `rulings` entry (`adjudication-precedence-order`). So the fix is a fifth table row pointing at `rulings`, not new machinery. Worth doing because the table is the actionable half of the section — a reader who follows it lands nowhere for the one case the prose singles out as most likely to be silently "fixed" by the next person.

## Note

This is one of two open PRs editing `CLAUDE.md` (the other is #1534, which documents the required `Representation-Change:` trailer). They add different sections in different places, so no textual conflict is expected, but whichever lands second is worth a rebase check.

Issue #1517, opened 2026-08-07
Normalizer: whole-block reverse complement left as two `delins` members when its interior coincides (`inv` under-recognized)


nh13:
## What's wrong

A whole-block reverse complement is left as **two separate `delins` members** when enough of its interior coincides, instead of being described as one `inv`.

`NM_004006.2:c.76_83` — the block `general.md:110` uses to illustrate the `inv` keyword — is `AATGCACA`. Its reverse complement `TGTGCATT` coincides with the reference at **4 of 8 columns**, so only the two flanking pairs change:

```
NM_004006.2:c.76_83inv   ->  NM_004006.2:c.[76_77delinsTG;82_83delinsTT]
```

The same block on the genomic axis has always produced this split; PR #1484's axis widening made the transcript axes agree with it, which is how it surfaced.

## Why `inv` is the better description

`general.md:56` states the prioritisation outright:

> when a description is possible according to several types, the preferred description is: (1) substitution, (2) deletion, (3) inversion, (4) duplication, (5) insertion.

`delins` is absent from that list — `DNA/delins.md` defines it as the residual form, "not a substitution, inversion or conversion". So when the competing description's members are `delins`, inversion outranks them and `inv` is preferred.

This is the same rule the codebase already applies in the opposite direction. `issue_1230_inv_with_unchanged_interior_becomes_substitutions` splits `g.20_23inv` into `g.[20G>C;23G>C]` and cites `general.md:56` for it — correctly, because there the competing members are **substitutions**, which outrank inversion. One rule settles both:

| case | competing members | priority | preferred |
|---|---|---|---|
| #1230 (`GATG` -> `CATC`) | two substitutions | 1 vs 3 | split |
| this issue (`AATGCACA` -> `TGTGCATT`) | two `delins` | residual vs 3 | `inv` |

## Why the current gate misses it

`coalesce_whole_block_inversion` admits a block only when
`every_separation_is_a_single_base(pieces) || changed_columns_dominate_the_span(pieces)`.
Here the separation is 4 bases and exactly 4 of 8 columns change, so `4 * 2 > 8` is false by one and the pass declines. Neither half of that gate reads the *type* of the competing members, which is what `general.md:56` actually ranks.

## Scope note

`coalesce_whole_block_inversion` is axis-neutral and shipped, so any change here moves genomic representations, not only the transcript axes. The blast radius must be measured on a real corpus before merging, and the change batched into a single release rather than dribbled out.

[...]

Issue #1530, opened 2026-08-07
Follow-ups from the #1519/#1520/#1521 review sweep


nh13:
Findings raised during the dual-review sweep of #1519, #1520 and #1521, which were merged as-is on the operator's instruction. **None is behavioural** — no normalized output moves — which is why none blocked a merge. All are re-verified against `main` at `e26da840`; line numbers below are current, not the ones recorded at review time.

## A guard the generator asserts on one side but not the other

**1. The generator has no `clauses.len() >= 2` gate for an `undecided` ruling.** `examples/generate_spec_fixture.rs:1002` bails when an undecided record names a `governing` or `deviates_from` clause, but not when it cites fewer than two clauses. The *committed test* does assert it — `ruling_records_are_intact`'s undecided arm requires both sides on the record. So the build gate is weaker than the test it exists to backstop, on exactly the rule (`undecided` must state a conflict, not a position) that the section was built to protect.

CodeRabbit attributed this to `ruling_records_are_intact`; the fix belongs in the generator's match arm.

**2. `RULING_STATUSES` has no duplicate-id guard.** `tests/it/hgvs_spec_normalization_tests.rs:793` collects it into a `BTreeMap` and compares; a duplicated id collapses last-wins and `assert_eq!(observed, pinned)` still passes on a malformed table. Its sibling `EQUIVALENCE_CLASS_VERDICTS` gets an explicit `assert_eq!(pinned.len(), …, "…has a duplicate id")`, and the generator checks duplicates on **both** sides (`seen_ids`, `seen_rulings`). This is the committed half missing the parity.

## A documented contract the code does not hold

**3. `anchor` mis-splits a member with no leading accession but an inner reference.** `examples/generate_spec_fixture.rs:813-821`. The doc comment states that "a colon later in the string belongs to an inner reference (`g.10_11ins[NC_…:g.20_30]`) and must be left alone", but `split_once(':')` only delivers that when a leading accession is present. Given `g.10_11ins[NC_2.1:g.20_30]` it splits at the *inner* colon and returns `X.1:g.20_30` — a different variant, silently.

Latent today: every current class member reaches `anchor` with a leading accession via `input_prefixed`. It goes live the first time a curated member is a bare fragment carrying an inner reference, and it fails silently rather than at the build gate. Either make the split accession-shape-aware, or `bail!` when the leading segment is not accession-shaped so a mis-anchored member fails the build.

## Rules stated more narrowly than the code now implements

[...]

PR #1532, opened 2026-08-07
docs: record the unrecorded adjudications and correct two ruling rationales

nh13:
Records adjudications that were live in the project's working record but written down nowhere durable, and corrects two committed records whose stated reasoning does not survive its own citation.

## Why

An audit of prior sessions, campaign documents and the repo's own records inventoried roughly 200 adjudications. About 65 were recorded somewhere durable; about 45 were not, and 12 contradicted either each other or the code. This lands the highest-value part of that backlog. Everything here is a record or a doc comment — no normalizer behaviour changes, and nothing was re-blessed.

## The two corrections

**The RFC 2119 argument is withdrawn.** `rulings[delins-merge-vs-individual-gap-two-or-more]` argued that `delins.md:17` and `:47` are "the same RFC 2119 strength and point opposite ways". A keyword census over `docs/recommendations/` at the pinned spec checkout finds an uppercase RFC 2119 keyword in exactly one place outside `style.md` — `RNA/adjoined_transcript.md:21` — and neither contested clause is among them. Both are lowercase prose. The question may still be undecided; the argument for it needed rewriting either way. The rationale now carries the census and the command to re-run it.

**The #1517 ruling record was misnamed, and the misnomer inverted its argument.** `inversion-vs-two-substitutions-76-83` describes a case whose competing members are two `delins`, not two substitutions: `NM_004006.2:c.76_83` is `AATGCACA`, its reverse complement `TGTGCATT`, and they coincide at the four interior columns. That distinction decides the `general.md:56` argument, because `:56` ranks substitution above inversion while `delins` does not appear in its list at all. Renamed to `inversion-vs-two-delins-76-83`, with the bases re-derived against the prepared reference.

## What else is recorded

Three new `undecided` records, each stating both positions rather than picking one: `adjudication-precedence-order` (two irreconcilable orders are on the project's record and neither was ever in the repo), `canonical-form-choice-when-both-legal`, and `codon-carve-out-shape-restriction`.

Doc comments in `src/normalize/merge.rs` capture findings that were expensive to establish and had no home: the Mutalyzer forensics (it minimizes a weighted description length, with constants dated 2014 and no separation concept anywhere in its source or history), that no minimality principle exists in the spec (`basics.md:38` lists stable, meaningful, memorable, unequivocal — stability first, minimality absent), and the insertion separation arithmetic, with a unit test. An insertion consumes no reference position, so modelling `A_B` as a consumed span double-counts; two published gap distributions were wrong for want of this.

`cases.json` records the interop-direction measurement together with its confidence interval. The point estimate said a candidate moved ferro 45% away from the reference implementation, but the interval crosses zero — recording the estimate without the interval would be worse than not recording it.

## One record settled empirically

`EQUIVALENCE_CLASS_VERDICTS` pinned three classes of a *decided* ruling as non-confluent, which read as ferro violating its own ruling. It is a harness limit: the spec-fixture driver normalizes through `MockProvider::new()`, whose maps are empty, so `axis_frame` finds no reading frame and the codon path cannot fire. Demonstrated on the identical shape — a CDS-bearing provider converges `c.[1G>T;3T>A]` and `c.1_3delinsTTA`; the empty provider returns the split byte-identical. The comment now says so.

## Reviewer notes

Five records are `undecided` and need an operator ruling; `adjudication-precedence-order` decides the other two. Recording the tension is the deliverable here — nothing in this PR picks a side.

There is a known tension with the unpushed `fix/inv-priority-over-delins` branch (#1517), which sets `delins-merge-vs-individual-gap-two-or-more` to `decided`. These two branches must not both merge without settling that question first, or merge order silently picks the answer.

Every spec citation was re-verified against `assets/hgvs-nomenclature` at `6f85311` independently of the source that suggested it. Full suite: 9582 tests run, 9582 passed, nothing re-blessed.

Representation-Change: none


nh13:
@coderabbitai pause

nh13:
Dual-review sweep. Both lenses run; CLI raised 4, I confirm **3** and reject 1. Two are worth acting on before merge, and one of them is a defect in a number **I produced** — flagging that explicitly since this PR is what makes it durable.

## Confirmed: the interop interval and its point estimate carry opposite signs

`cases.json` records "net -10 of 56 … 95% CI -4.3% to +40.0%". Reproduced from the stated counts (25 away, 15 toward, 16 third):

```
net toward-minus-away :  -10  ->  -17.9%
net away-minus-toward :  +10  ->  +17.9%
95% CI on AWAY-minus-toward:  -4.3% to +40.0%   <- the quoted interval
95% CI on TOWARD-minus-away: -40.0% to  +4.3%
```

So `-10` is *toward minus away* while the interval is on *away minus toward*. Both mean "worse", but the interval as written **does not bracket its own point estimate**, and a reader checking the arithmetic will conclude one of the two numbers is wrong.

Either flip the interval to `-40.0% to +4.3%`, or state the point estimate as `+10 of 56 (away − toward)`. The second reads better next to "the point estimate says worse". Whichever you pick, the same pair appears in `CLAUDE.md` and in the `adjudication-precedence-order` ruling and must move together.

**Provenance note:** this framing came from #1484's PR body, and I relayed it in my review of #1484 and again in the v0.13.0 changelog discussion without catching the sign flip. It is not introduced here — but this PR is what promotes it from a PR description to the corpus's permanent provenance record, which is the right moment to fix it.

## Confirmed: the new mixed-pair fixture contradicts its own assertion message

`src/normalize/merge.rs`, in `an_insertion_contributes_no_width_to_the_separation_it_sits_in`:

```rust
separations_are_meaningful(&[sub(10, b"A"), ins(12, b"T")], 1),
"two unchanged bases separate a substitution at 10 from an insertion at 12"
```

`sub(10)` is `ref_start=10, ref_end=11`; `ins(12)` is `ref_start=12, ref_end=12`. The gap is `12 - 11 = 1` — **one** unchanged base (position 11), not two. `ins(13)` gives the two the message describes (positions 11 and 12).

It passes either way, which is the problem: `MIN_PIECE_SEPARATION = 1`, so a gap of 1 and a gap of 2 both satisfy the assertion. The test therefore documents a two-base case while exercising a one-base one, and the *interesting* case — the one that distinguishes `MIN_PIECE_SEPARATION` from `RAISED_PIECE_SEPARATION = 2` — is never reached. Worth fixing precisely because this test was added to pin insertion-width semantics.

## Confirmed: `Representation-Change: none` is correct

Verified rather than taken: the entire +147 in `src/normalize/merge.rs` is one new unit test inside `#[cfg(test)] mod tests {` (opened line 8750, test at 9772), with **zero** production lines removed. This is the new required check's first live encounter with a watched-directory change — it forced the judgement, and the judgement is right.

## Rejected: "add a CDS-bearing provider test for the codon carve-out" (Major)

CodeRabbit wants a provider-backed test asserting `c.[1G>T;3T>A]` converges to `c.1_3delinsTTA`. That coverage is worth having, but not in this equivalence-class harness: the harness is *deliberately* hermetic — it runs every row through `MockProvider::new()` so the classes stay reproducible without reference data. Adding a provider-backed row would make one class' evaluation depend on a fixture the other eleven do not use, and the `not-evaluable` verdict exists precisely to record "this class cannot be decided hermetically". The PR already does the right thing: it states the harness limit, names the mechanism (`axis_frame` returns `None` with no `cds_start`), and reports the measured provider-backed result in the comment. A separate provider-backed regression test elsewhere would be a fine follow-up.

## Also verified

The withdrawal of the "equal RFC 2119 strength" argument is correct, and I checked the census independently: uppercase RFC 2119 keywords appear in exactly two files under `docs/recommendations/` — `style.md` itself and `RNA/adjoined_transcript.md`. So `delins.md:17` and `:47` are lowercase prose and the strength-parity claim does not hold.

**That argument shipped in #1519, which I reviewed and approved.** I verified the `:47` clause text verbatim and caught its length-condition error, but took the RFC 2119 framing at face value — one grep would have shown it was wrong.

## Note for whoever merges

This PR renames a ruling id (`inversion-vs-two-substitutions-76-83` -> `inversion-vs-two-delins-76-83`). That is exactly the edit class where #1530's finding 2 bites: `RULING_STATUSES` still has no duplicate-id guard, so a rename colliding with an existing id would collapse last-wins and still pass `assert_eq!(observed, pinned)`.

Issue #955, opened 2026-07-06
tracking: HGVS spec-fixture conformance gaps — spec-valid parse failures & spec-invalid over-accepts (v21.0.0 audit)


nh13:
## Summary

A v21.0.0 spec-compliance audit via the generated spec-normalization fixture (`tests/fixtures/grammar/hgvs_spec_normalization.json`, harvested from the vendored HGVS spec at tag 21.0.0) surfaced two classes of substantive, noise-adjusted deviation. This is the triage backlog; each case needs per-case confirmation (some over-accepts are arguably intentional lenient-input/canonical-output, some parse gaps are genuinely low-value). Context: ferro round-trips 694 spec strings exactly, correctly rejects 67, and **mis-normalizes zero spec-valid inputs** — these are the edges.

Related already-filed: the self-cancelling over-rejection of `c.[762_768del;767_774dup]` is #954 (not repeated here).

## Class A — spec-valid forms ferro currently FAILS TO PARSE (~23, under-permissive)

**Protein short-tandem-repeat count**
- `p.Gln[21]`, `p.Q[21]`

**Protein `Xaa[n]` / `Ter` / `*` insertions** (fail with "Unexpected trailing characters" on the `[n]`/count)
- `NP_003997.1:p.(Val582_Asn583insXaa[5])`, `p.Arg78_Gly79insXaa[23]`, `p.Lys2_Leu3insXaa[34]`, `p.(Ala123_Pro131)insXaa[4]`
- `NP_060250.2:p.Gln746_Lys747ins*63`, `p.Lys2_Leu3insTer12`

**Mosaic / chimeric `=/` notation (protein & RNA)**
- `LRG_199t1:p.(Trp24=/Cys)`, `NP_003997.1:p.(Val7=/del)`, `p.(Val7=/dup)`
- `LRG_199t1:r.(=/6_8del)`, `LRG_199t1:r.=/6_8del`

**Complex / cross-chromosome genomic `delins`** (distinct from the ISCN-only forms the fixture overrides already reject)
- `g.chr4:…inschrX:…`, `g.[chr4:…inschrX:…inv]`, `g.chr2:pter_8247756delinschr11:pter_15825266`, `g.chr3:…_qterdelinschr8:(…)_qter`, `g.chr5:pter_…delinschr10:…inv`

**RNA 3'-terminal-region positions**
- `r.*1924`, `r.*279`

[...]

PR #1533, opened 2026-08-07
fix(spec-fixture): close the review-sweep follow-ups from #1519/#1520/#1521

### LINKED ISSUE #1530: Follow-ups from the #1519/#1520/#1521 review sweep

nh13:
Closes #1530.

All eight items. **Nothing behavioural** — no normalized output moves, and the generated spec fixture is byte-identical.

## The defects

Three guards that can pass while being wrong (items 1–3), and five statements of a rule that the code no longer implements (items 4–8).

**1. The generator had no `clauses.len() >= 2` gate for an `undecided` ruling.** An `undecided` record is a claim that two clauses *conflict*, so a record citing one clause states no conflict — only a position it declines to name as one. `ruling_records_are_intact` asserts that; the generator asserted only the governing/deviated-from half. So the build gate was weaker than the test it exists to backstop, on exactly the rule the section was built to protect. The record-shape checks move into `validate_ruling_shape` — everything judgeable from the record alone, with no spec checkout and no fixture rows — which is unit tested directly, and the missing arm is added.

**2. `RULING_STATUSES` had no duplicate-id guard.** It is collected into a `BTreeMap` and compared, which is last-wins: a duplicated id silently drops one pin and `assert_eq!(observed, pinned)` still passes on a malformed table. Its sibling `EQUIVALENCE_CLASS_VERDICTS` already carries the length assert and the generator checks duplicates on *both* of its own tables, so this was the committed half missing the parity.

**3. `anchor` mis-split a member with no leading accession but an inner reference.** The doc states that a colon after the first belongs to an inner reference and must be left alone, but `split_once(':')` only delivers that when a leading accession is present. Given `g.10_11ins[NC_2.1:g.20_30]` it split at the *inner* colon and returned `X.1:g.20_30` — a different variant, and still well formed, so nothing downstream would catch it. The split is now accession-shape aware: a head beginning `<sys>.` is a bare fragment and gets prefixed, anything else is the reference and gets replaced.

This is the same defect #955 fixed for the harvester (`spec_harvest::prefix::default_prefixed`), reached from the opposite side — there a bare fragment was wrongly judged *already* prefixed, here it was wrongly *re*-anchored.

**4–8** are accuracy fixes to prose, detailed in the second commit's message: the msto attribution rule stated per-function in four places while #1520 widened it to module-doc attribution; "one guard per stage" where it is one *or more*; #1521's displayed moments being the even-`n` case asserted for every `n`; the corresponding `CLAUDE.md` sentence; and a stale `#[allow(dead_code)]` whose comment was wrong about which fields are live.

## Measured

| | |
|---|---|
| generated fixture, `origin/main` generator | md5 `0226b06a28044be56132d8b216488c3a` |
| generated fixture, this branch | md5 `0226b06a28044be56132d8b216488c3a` — byte-identical |
| curated class members that `anchor` mis-splits today | **0 of 24**, across 12 classes (4 declaring a `reference`) |
| ruling records rejected by the new gate | 0 of 3 — all three already cite both sides |

The zero on `anchor` is structural and is reported as such rather than as evidence of safety: every curated member reaches `anchor` with a leading accession via `input_prefixed`, so the corpus cannot currently build the shape the fix is about. It is gated now rather than when a curated member first arrives as a bare fragment carrying an inner reference — at which point it would fail silently rather than at the build gate.

The moments correction is arithmetic, not a measurement: `U = 2·Binomial(floor(n/2), 1/4)` gives mean `m/2` and sd `sqrt(3m/4)` for `m = floor(n/2)`. #1461's `n = 457` is odd, so the displayed `n/4` overstates the mean as 114.25 rather than 114 and yields `z = +2.73` where the table says `+2.75`. The table was right throughout — only the prose generalised, which is why nothing caught it.

## Unchanged on purpose

- **`decided` may still cite a single clause.** A settled carve-out names only the clause it applies, so the two-clause rule is the `undecided` arm's alone. Asserted explicitly so the arm cannot be rewritten in terms of the clause count.
- **`anchor` still *replaces* a real leading accession** rather than prefixing it, including when an inner reference follows. Asserted alongside the new case, since the obvious over-generalisation — "never split, always prefix" — passes every other assertion here.
- **The equivalence classes' outputs are still not pinned**, and citation paths are still not checked for `..`/absolute escapes. Both were raised in #1530 and deliberately recorded as not-work; neither is done here.

## Representation stability

No canonical form moves. `src/normalize/merge.rs` is touched only in a doc comment and a new `#[cfg(test)]` unit test; the generator's two changes are build-time gates whose artifact is byte-identical, verified by A/B with each generator separately rebuilt (distinct binary mtimes, `Compiling ferro-hgvs` on both). `examples/dump_normalized_corpus.rs` was not run, because there is no code path here that could move a row — the diff contains no change to any function reachable from `normalize()`.

## Verification

```
cargo fmt --all --check                                      # clean
cargo clippy --all-features --all-targets -- -D warnings     # clean
# 9586 tests run: 9586 passed (3 slow), 145 skipped
cargo run --features dev --bin generate_spec_fixture         # builds; overrides pass the new gate
```

**Nothing was re-blessed** — every pre-existing expectation already agreed with the new answers, which for a change that adds three gates is the signal that they are gates and not behaviour.

Each guard was also verified to actually fire, not just to exist:

- duplicating a `RULING_STATUSES` entry → `RULING_STATUSES has a duplicate id (left: 3, right: 4)`
- `the_unchanged_column_moments_reproduce_the_tabulated_z_scores` with the doc's displayed formulas → `n = 457: the stated moments give z = 2.7309, but the table says 2.75`
- `an_undecided_ruling_must_put_both_sides_on_the_record` did not compile before `validate_ruling_shape` existed

Representation-Change: none

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -277,11 +277,13 @@ same description**. (This used to say "conflict at equal RFC 2119 strength", cit
 That framing is withdrawn — uppercase RFC 2119 keywords appear exactly once outside `style.md`
 itself, so essentially every clause in play is lowercase prose and keyword strength cannot rank
 them. See [Reading the spec](#reading-the-spec).) Each record names the clauses in tension, which
-one governs, which is deviated from, and why. `undecided` is a first-class state and the generator **refuses** to build a record
-that is `undecided` yet names a governing clause — an unsettled question must not be able to
-smuggle in a ruling nobody made. Every citation carries a verbatim quote that the generator checks
-against the spec checkout, so a submodule bump that moves a clause fails the build instead of
-leaving the citation pointing at unrelated prose.
+one governs, which is deviated from, and why. `undecided` is a first-class state and the generator
+**refuses** to build a record that is `undecided` yet names a governing *or* a deviated-from clause
+— either implies a side was chosen, and an unsettled question must not be able to smuggle in a
+ruling nobody made. It equally refuses an `undecided` record citing fewer than two clauses: such a
+record states no conflict, only a position it declines to name as one. Every citation carries a
+verbatim quote that the generator checks against the spec checkout, so a submodule bump that moves
+a clause fails the build instead of leaving the citation pointing at unrelated prose.
 
 Requires the `assets/hgvs-nomenclature` submodule (`git submodule update --init assets/hgvs-nomenclature`); without it the generator fails with `no HGVS strings harvested from …`, naming that command.
 
--- a/examples/generate_spec_fixture.rs
+++ b/examples/generate_spec_fixture.rs
@@ -805,19 +805,104 @@ mod decisions {
     const NON_CONFLUENT: &str = "non-confluent";
     const NOT_EVALUABLE: &str = "not-evaluable";
 
+    /// True when `head` — the text before a target's first colon — is a
+    /// reference accession rather than the opening of a bare coordinate
+    /// fragment.
+    ///
+    /// HGVS puts the coordinate system immediately after the reference's colon,
+    /// so a fragment carrying no accession *begins* with one of the
+    /// single-letter descriptors followed by a dot. Anything else in that
+    /// position is the reference — accessions (`NC_000023.11`, `LRG_199t1`,
+    /// `NM_004006.2`) never take that shape.
+    fn is_accession(head: &str) -> bool {
+        !matches!(
+            head.as_bytes(),
+            [b'g' | b'o' | b'm' | b'c' | b'n' | b'r' | b'p', b'.', ..]
+        )
+    }
+
     /// Re-anchor `target` onto `reference`, replacing any accession it carries.
     ///
     /// The accession is everything before the **first** colon, which is where
     /// HGVS puts it; a colon later in the string belongs to an inner reference
     /// (`g.10_11ins[NC_…:g.20_30]`) and must be left alone.
+    ///
+    /// The first colon only delimits the accession when something
+    /// accession-shaped precedes it (#1530). A member that is a bare fragment
+    /// *and* carries an inner reference — `g.10_11ins[NC_2.1:g.20_30]` — puts
+    /// the inner colon first, and splitting there yields `X.1:g.20_30`: a
+    /// different, still well-formed variant, so nothing downstream would catch
+    /// it. Latent while every curated member reaches here with a leading
+    /// accession via `input_prefixed`; gated now rather than when it goes live.
+    ///
+    /// This is the same defect `spec_harvest::prefix::default_prefixed` fixed
+    /// for the harvester in #955 — an inner `[…:<sys>.…]` read as the variant's
+    /// own accession — arrived at from the opposite side: there a bare fragment
+    /// was wrongly judged *already* prefixed, here it is wrongly *re*-anchored.
     fn anchor(target: &str, reference: Option<&str>) -> String {
         match reference {
             None => target.to_string(),
             Some(accession) => match target.split_once(':') {
-                Some((_, rest)) => format!("{accession}:{rest}"),
-                None => format!("{accession}:{target}"),
+                Some((head, rest)) if is_accession(head) => format!("{accession}:{rest}"),
+                _ => format!("{accession}:{target}"),
             },
         }
+    }
+
+    /// Fail unless a ruling record is well formed *as a record* — everything
+    /// that can be judged from the record alone, with no spec checkout and no
+    /// fixture rows.
+    ///
+    /// Extracted so it can be unit-tested directly. The committed consumer
+    /// `ruling_records_are_intact` asserts exactly these properties; a gate that
+    /// is weaker than the test it backstops lets a malformed record into the
+    /// artifact and only fails later, at the far end of the build.
+    fn validate_ruling_shape(owner: &str, ruling: &overrides::Ruling) -> anyhow::Result<()> {
+        if ruling.rationale.trim().is_empty() {
+            anyhow::bail!("{owner}: a ruling record must say why");
+        }
+        if ruling.clauses.is_empty() {
+            anyhow::bail!("{owner}: a ruling record must cite at least one clause");
+        }
+        let cited: BTreeSet<&str> = ruling.clauses.iter().map(|c| c.clause.as_str()).collect();
+        for role in ruling.governing.iter().chain(ruling.deviates_from.iter()) {
+            if !cited.contains(role.as_str()) {
+                anyhow::bail!(
+                    "{owner}: names {role:?} as governing or deviated-from, but that clause is \
+                     not in `clauses` — every role must point at a cited, quote-verified clause"
+                );
+            }
+        }
+        match ruling.status {
+            overrides::RulingStatus::Decided if ruling.governing.is_none() => {
+                anyhow::bail!("{owner}: a `decided` ruling must name the clause it holds to govern")
+            }
+            // The whole point of the `undecided` state is that nobody has
+            // ruled. A governing clause — or a deviated-from one, which
+            // implies a side was chosen — is an invented ruling, so it is a
+            // build failure rather than a lint.
+            overrides::RulingStatus::Undecided
+                if ruling.governing.is_some() || !ruling.deviates_from.is_empty() =>
+            {
+                anyhow::bail!(
+                    "{owner}: an `undecided` ruling must not name a governing or deviated-from \
+                     clause — record the conflict, not a ruling nobody made"
+                )
+            }
+            // The other half of the same rule, and the half this gate was
+            // missing (#1530): `undecided` asserts that clauses *conflict*, so a
+            // record citing one clause states no conflict — it states a position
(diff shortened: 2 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
Dual-review sweep. I filed #1530, so treat my "all eight addressed" as self-interested and weight the CLI lens accordingly — it found something I did not, in the very item I wrote.

## Blocking, and a true positive

`Representation change declared` is the one failing check of 37. This PR touches `src/normalize/merge.rs` (+76) and the description carries no `Representation-Change:` trailer.

Verified it is a true positive rather than the gate misfiring: the only non-comment change under `src/` is the new unit test `the_unchanged_column_moments_reproduce_the_tabulated_z_scores`, added inside `#[cfg(test)] mod tests {` (opened line 8750), with **zero** production lines removed. The hunks at 3977/3998 are the doc-comment corrections. So the correct declaration is `Representation-Change: none` and it is simply absent.

Worth saying plainly: the requirement is documented five times in `CONTRIBUTING.md` and **zero** times in `CLAUDE.md`, which is the file an agent reads. That is a gap in the check, not in this PR — #1534 fixes it and cites this PR as the demonstration.

## Confirmed (CLI): the mirror-pair model needs its alphabet stated

`src/normalize/merge.rs` ~3980-3998. The corrected text says "an odd `n` has a centre column that can never coincide", and the moments `E[U] = m/2`, `sd(U) = sqrt(3m/4)` follow from that. But:

```rust
assert_eq!(complement_base(b'N'), Some(b'N'));   // merge.rs:11354
```

**`N` is self-complementary**, so a centre column holding `N` satisfies `b[i] == complement(b[i])` and *does* coincide. The `1/4` per-pair probability likewise assumes a uniform four-letter alphabet. Reference sequences contain `N` at assembly gaps, so this is not a hypothetical alphabet.

Nothing in the tabulated z-scores changes — all three cited blocks are pure ACGT — so this is a statement about the model's scope, not its arithmetic. The fix is one clause: say the model is over a uniform A/C/G/T alphabet, and that `N` is self-complementary so a centre column can coincide.

**This is my miss twice over.** I wrote #1530's item 6, and I reviewed and independently re-derived the statistics in #1521 — reproducing every z-score and tail probability and the mirror-pair proof — while never asking what alphabet the proof assumed. Verifying arithmetic *inside* an unstated assumption is not the same as verifying the claim, and this is the third time in this campaign that a config- or model-level assumption survived a careful check of the numbers on top of it.

## Verified addressed, item by item

Checked against the code rather than the description:

| #1530 item | evidence |
|---|---|
| 1 · generator `clauses.len() >= 2` gate | `validate_ruling_shape` at `generate_spec_fixture.rs:860`; the arm at `:898` |
| 2 · `RULING_STATUSES` duplicate guard | `hgvs_spec_normalization_tests.rs:800-803`, mirrors the sibling's assert |
| 3 · `anchor` accession-shape aware | `:845` now guards `is_accession(head)`; new test at `:1122` covers all seven coordinate systems **and** the converse (a real leading accession is still replaced) |
| 5 · stale `#[allow(dead_code)]` | removed |
| 6 · `E[U] = n/4` | now `E[U] = m/2`, `sd(U) = sqrt(3m/4)`, with the even-`n` forms named as such — modulo the alphabet caveat above |

The extraction of `validate_ruling_shape` is the right shape for item 1: it isolates everything judgeable from the record alone, so it can be unit-tested without a spec checkout or fixture rows, which is why the gate could drift from the test it backstops in the first place.

I also note each guard was verified to *fire*, not merely to exist — including that `an_undecided_ruling_must_put_both_sides_on_the_record` did not compile before the extraction. That is the right standard and it is what I would have asked for.

PR #1535, opened 2026-08-07
fix(normalize): type the whole-block inversion by what it competes with

### LINKED ISSUE #1517: Normalizer: whole-block reverse complement left as two `delins` members when its interior coincides (`inv` under-recognized)

nh13:
Types a whole-block reverse complement as an `inv` when the description it competes with is two `delins` members rather than two substitutions, and records the ruling that settles it.

Closes #1517

## The distinction that decides it

`NM_004006.2:c.76_83` is `AATGCACA`; its reverse complement `TGTGCATT` coincides at the four interior columns. So the same edit is describable as one inversion or as `c.[76_77delinsTG;82_83delinsTT]`. The competing members are **two `delins`, not two substitutions** — and `general.md:56` lists substitution, deletion, inversion, duplication and insertion, with `delins` absent. The clause therefore never reaches this competitor.

That is exactly what separates this from #1230, whose competing members really are substitutions and which keeps its split. Both #1230's guard and #1040's negative control stay green, untouched.

## Why this is spec-explicit rather than a tie-break

`inversion.md:5` defines an inversion as more than one nucleotide replaced by the reverse complement of the original — a property of the whole span, silent on interior columns coinciding. And the spec's own worked example has this shape: `inversion.md:33-34` publishes `NM_004006.2:c.4145_4160inv`, whose reference `TCCAGGAGTCCCTCAC` and reverse complement `GTGAGGGACTCCTGGA` differ at 10 of 16 columns, leaving two three-base unchanged interior runs. So `general.md:34` demonstrably does not decompose an inversion into separated variants in the spec's own usage.

This also refutes a claim once used to argue the other way — that every worked inversion example in the spec changes all of its bases.

## The gate

`no_piece_is_a_lone_substitution` — a whole-block reverse complement types as `inv` when no competing piece is a lone substitution.

## Blast radius: zero, and the zero is structural

Measured movement over the synthetic corpus is **0**. That is **not** reassurance and should not be read as any. The only family that builds a reverse-complement block emits its pieces at separation 0, which the first gate already admits, so the new predicate cannot fire anywhere in that corpus. Reported this way deliberately, per the repository's own rule that a corpus zero is a claim about the corpus.

The behaviour is instead locked by three targeted tests on a programmatic provider, which do construct the separation-4 multi-column geometry the corpus cannot: the #1517 case, #1230's negative control, and a sequence-preservation check. The helper asserts each encoding normalizes to the canonical form *and* that the form is a fixed point, so it is a confluence property rather than a table of expected strings. `both_choices_preserve_the_sequence_they_denote` now normalizes every encoding and projects the output back through `hgvs_to_spdi` onto `SEQ`, comparing byte-for-byte against an inversion built independently of the normalizer — previously it asserted only on its own local helper and never invoked normalization at all, so the re-spelling bug it exists to catch would have passed it.

A corpus family able to vary that geometry is owed separately, and is in progress.

## Reviewer notes

**Stacked on #1532** (`docs/record-adjudication-harvest`), which carries the ruling records this PR updates. Review that one first; this diff is the code fix, its tests, and one ruling status flip.

The ruling `inversion-vs-two-delins-76-83` moves `undecided` → `decided` with governing clause `inversion.md:5`. `RULING_STATUSES` is updated to match, which `ruling_records_are_intact` enforces.

This branch previously also carried the `delins-merge-vs-individual-gap-two-or-more` ruling. That commit has now landed on `main` separately as #1538 and **dropped from this branch on rebase**, so this PR is a single ruling flip again. After it, 3 of the 6 records are `undecided`.

`CLAUDE.md`'s adjudication section is updated in the same change: it is the decision log's front door, and leaving it saying "Five of its six records are `undecided`" would have had this PR tell readers the ruling it makes was never made.

Full suite: 9590 tests run, 9590 passed (rebased onto `main` after #1538). Nothing re-blessed.

Representation-Change: 0 rows move over 5,761,302 real expressions (ClinVar 500,004 / Paraphase 435,235 / CMRG 4,826,063), diffed row by row between this branch and its base on both binaries. That zero is structural for the direction that could move: a cis-allele of >=2 `delins` members forming a whole-block reverse complement has **0** instances in those corpora, so the corpus cannot build the input this change converts. It is not vacuous either — 2,639 `inv` rows (897 of them genomic) and 38,190 multi-base spanning `delins` rows do reach the typing path and are byte-identical across the change, so the gate demonstrably does not over-fire. Where the shape does occur, the split spelling moves onto `inv`; the `inv` and spanning-`delins` spellings do not move on `c.`, and on a frameless axis all three converge on `inv`.


nh13:
@coderabbitai pause

nh13:
## Review sweep

Reviewed from a detached worktree at `2ec3b1d3`, with the baseline taken at `800939bd` (this PR's parent, now merged as #1532). The headline is that I measured the representation claim rather than taking it, and the conclusion holds but the stated reason does not.

### The stability claim: right answer, wrong reason, and the trailer understates the scope

The trailer says `c.[76_77delinsTG;82_83delinsTT]` "now converges on `c.76_83inv`, which is the form ferro already emits for the `inv` spelling, so the canonical form does not move for anyone storing it." Run against the prepared reference, baseline vs head:

| input | `800939bd` | `2ec3b1d3` |
|---|---|---|
| `NM_004006.2:c.76_83inv` | `c.76_83inv` | `c.76_83inv` |
| `NM_004006.2:c.76_83delinsTGTGCATT` | `c.76_83inv` | `c.76_83inv` |
| `NM_004006.2:c.[76_77delinsTG;82_83delinsTT]` | *(unchanged — stays split)* | `c.76_83inv` |

So the winner is the already-shipped form, which is the cheap direction and worth saying explicitly. But the losing side still moves: a stored library that normalized the split spelling held `c.[76_77delinsTG;82_83delinsTT]` and now holds `c.76_83inv`. That is a merge-direction representation change for those rows, not nothing — it is the "every confluence fix picks a winner, and whichever side loses is a representation change for someone" case.

It is sharper than that on a frameless axis. This PR's own tests run on `g.` against a programmatic provider, and there the baseline normalized **even the `inv` spelling** to the split:

```
BASELINE  TEMPLATE:g.25_32inv                      -> TEMPLATE:g.[25_26delinsTG;31_32delinsTT]
BASELINE  TEMPLATE:g.25_32delinsTGTGCATT           -> TEMPLATE:g.[25_26delinsTG;31_32delinsTT]
BASELINE  TEMPLATE:g.[25_26delinsTG;31_32delinsTT] -> TEMPLATE:g.[25_26delinsTG;31_32delinsTT]
```

All three become `g.25_32inv` at head. So the test geometry is not equivalent to the `c.76_83` case it stands in for — on `c.` one spelling moves, on `g.` all three do. Worth knowing, because the tests are the PR's stated evidence for the behaviour and they pin the harsher case while the trailer describes the milder one.

### Measured blast radius on real corpora: 0 of 5,761,302, and here is what that zero is worth

The synthetic-corpus zero is structural and the PR says so, which is the right form. I ran the other arm — the three committed LFS corpora, both binaries, release profile, prepared reference:

| corpus | rows | moved |
|---|---|---|
| ClinVar | 500,004 | 0 |
| Paraphase | 435,235 | 0 |
| CMRG | 4,826,063 | 0 |
| **total** | **5,761,302** | **0** |

Per this repo's own rule I checked whether the corpora can construct the shape, and for the input direction that moves they cannot:

| property | count |
|---|---|
| cis-alleles `[...]` | 43,088 |
| ...with ≥2 members | 198 |
| ...with ≥2 `delins` members — the shape this gate converts | **0** |
| rows containing `inv` | 2,639 (1,713 `c.` / 897 `g.` / 29 `n.`) |
| single spanning `delins`, payload > 1 base | 38,190 |

So the split-input direction is structurally invisible here too, for the same reason as the synthetic corpus: the shape essentially does not occur in submitter-authored data (0 in 5.76M, and only 198 multi-member cis-alleles at all).

The zero is not entirely vacuous, though, and this is the part worth quoting: 2,639 `inv` rows and 38,190 multi-base spanning `delins` rows **do** reach the typing path and every one is byte-identical across the change. That includes 897 genomic-axis `inv` rows, which is the axis where the mock shows the baseline emitting a split. That is genuine negative evidence that the gate does not over-fire, which is a different and stronger claim than "the corpus found nothing".

Suggested trailer, which I think is both more accurate and still cheap to accept:

```
Representation-Change: 0 rows move over 5,761,302 real expressions (ClinVar 500,004 / Paraphase 435,235 / CMRG 4,826,063), byte-identical on both binaries. The direction that could move — a cis-allele of >=2 delins members forming a whole-block reverse complement — has 0 instances in those corpora, so that zero is structural. Where the shape does occur the split spelling moves onto `inv`; the `inv` and spanning-delins spellings are unchanged on 40,829 real rows that reach the typing path.
```

### It flips two rulings, not one

The reviewer notes say "one ruling status flip", but the diff moves both:

- `inversion-vs-two-delins-76-83` — `undecided` → `decided`, which is the one described.
- `delins-merge-vs-individual-gap-two-or-more` — `undecided` → `decided`, governing `delins.md:47`, undeclared in the description.

The second is the larger adjudication: `CLAUDE.md` describes it as governing ~138 of 208 adjudicated genomic rows, and the record itself says it was the one blocking other work. Its rationale is careful — the ruling is explicitly scoped, and it ratifies shipped behaviour so it moves no row, which I believe. But a reader scanning the description would not know the PR settles it, and a record this consequential deserves its own line in the body (or its own PR).

### `CLAUDE.md` contradicts the same commit — I would rate this major

Base was genuinely 5-of-6 `undecided` (`delins-codon-carve-out-gap-one` was already `decided`), so the existing text was accurate. This PR takes it to 3-of-6 while `CLAUDE.md` still reads:

> **Five of its six records are `undecided`**

and tabulates both newly-decided records under "what is open", including `adjudication-precedence-order`'s description as deciding "both rows above" — rows that are no longer open. This is the repository's decision log, whose stated purpose is to stop the next reader re-deriving a settled argument. Having it declare two settled questions open, in the very PR that settles them, is the exact failure the section exists to prevent, so I would raise it above a docs nit.

### `both_choices_preserve_the_sequence_they_denote` does not test the normalizer

The test's doc comment is right about why it should exist:

> A re-spelling pass that changed the resulting sequence would satisfy both tests above — they only compare strings — so the denotation is asserted separately

But the body never calls `normalize_to_string`. It builds a local `inverted()` helper from `SEQ` and asserts that helper's own output `contains` the expected substring:

```rust
assert!(inverted(25, 32).contains("TGTGCATT"), "precondition: 25_32 inverts to TGTGCATT");
```

That is a self-check on the test's arithmetic, not on ferro. The re-spelling pass it is meant to guard against would pass it unchanged. Making it real means running the normalized output back through `hgvs_to_spdi` (or applying it to `SEQ`) and comparing the resulting sequence byte-for-byte against `inverted(...)` — `contains` would also want to be equality, since a substring check cannot see a payload written at the wrong offset.

### `ruling_records_are_intact` pins status but not the position taken

The test pins `id -> status` by value, and separately checks that a `decided` record names a `governing` clause and that the clause is cited. It does not pin *which* clause governs. So `inversion-vs-two-delins-76-83` could be re-decided from `inversion.md:5` to some other cited clause and the guard stays green, even though the assertion message says the point is that taking a position "must be a deliberate, reviewable edit, not a side effect." Changing which clause governs is taking a different position. Extending `RULING_STATUSES` to `(id, status, governing)` would close it without new machinery.

### Nits

- Neither newly-decided record populates `deviates_from`, although both rationales name the clause overridden (`delins.md:17` and `general.md:34` respectively). The schema supports it and the test validates it when present, so the overridden clause is currently recoverable only from prose. `delins-codon-carve-out-gap-one` has the same omission, so this is a pre-existing pattern rather than a regression.
- The `adjudication-precedence-order` rationale still describes the other two records as things an operator ruling would close, which reads as stale now that two of them are closed.

### What is good here and should not be lost in the above

The `general.md:56` reading is the strongest part: `delins` is absent from the priority list, so the clause never reaches a two-`delins` competitor, which is precisely what separates this from #1230 without needing to reverse it. Reconciling #1230 and #179 — filed as contradictory bug reports — under one clause, with the negative control sitting beside the positive one in the same file, is a genuinely good piece of work. And reporting the synthetic zero while refusing to take credit for it, citing the repository's own rule, is the honest form and the one I would like to see copied.

<details>
<summary>Reproduction</summary>

```bash
# baseline = this PR's parent
git worktree add --detach ../review-1532 800939bd
git worktree add --detach ../review-1535 2ec3b1d3
(cd ../review-1532 && cargo build --release --bin ferro)
(cd ../review-1535 && cargo build --release --bin ferro)

# corpus inputs
gzcat tests/fixtures/bulk/clinvar_hgvs_500k.json.gz \
  | python3 -c 'import json,sys; [print(c["input"]) for c in json.load(sys.stdin)["test_cases"]]' > clinvar.txt
# ...same for paraphase_genes_exhaustive.json.gz and cmrg_genes_exhaustive.json.gz

for t in review-1532 review-1535; do
  ../$t/target/release/ferro normalize -i clinvar.txt --reference "$REF" -f tsv -j 10 -o $t.clinvar.tsv
done
# then diff the (input, normalized) pairs positionally
```
</details>

PR #1537, opened 2026-08-07
fix(normalize): never split a delins into members on consecutive nucleotides

### LINKED ISSUE #1524: Normalizer: a split delins leaves two *adjacent* members (separation 0), which delins.md:16 requires be one delins
Ferro splits a lone `delins` into members that are directly adjacent — no unchanged nucleotide between them. `DNA/delins.md:15` states the opposite outright:

> changes involving two or more consecutive nucleotides are described as deletion/insertion (delins) variants.

In several cases **the input was already correct and ferro made it wrong**:

| input | ferro output |
|---|---|
| `NM_000038.6:c.5265_5268delinsTGCG` | `c.[5265G>T;5266_5268delinsGCG]` |
| `NM_000083.3:c.2461_2464delinsCTCC` | `c.[2461_2463delinsCTC;2464G>C]` |
| `NM_001354906.2:c.4416_4419delinsTGCG` | `c.[4416G>T;4417_4419delinsGCG]` |
| `NM_000392.5:c.4465_4473delinsGGCCCACAG` | `c.[4465_4466delinsGG;4468_4470delinsCCA;4471_4473delinsCAG]` |

In row 1, members end at 5265 and begin at 5266 — consecutive. Row 4 splits at 4470/4471.

## Why no existing clause defends it

`general.md:34` ("two variants **separated by one or more** nucleotides should be described individually") does not reach separation 0 — nothing separates these. So unlike the long-running `:17` vs `:44-47` tension, there is no competing clause and no "recommended" escape hatch: `:15` is the only clause on point.

## Census

Measured over 5,761,894 real corpus rows (ClinVar 500k, multi-member cis, CMRG exhaustive, Paraphase exhaustive) against the prepared reference, auditing **every** multi-member normalized output rather than only rows that moved:

| | `main` @ `ade40fc1` | `#1484` @ `c70d32d4` |
|---|---|---|
| multi-member outputs audited | 6,235 | 8,180 |
| **`sep == 0` violations** | **61 rows / 59 distinct** | **61 rows / 59 distinct** |
| distinct violations new in #1484 | — | **0** |
| distinct violations fixed by #1484 | — | **0** |

**Identical sets — this is long-standing and independent of #1484.** (The *rate* differs, 0.9455% → 0.7249%, only because #1484 enlarges the denominator.)

All 59 are **lone-member inputs** that ferro itself split (`changed=true`): 58 on the `c.` axis, 1 on `g.`. 60 of the 61 violate under both readings of the `dup` positional ambiguity; the one exception (`NC_000017.11:g.[80110044C>G;80110045dup;80110047A>G]`) depends on it.

Separately, 7 rows show *overlapping* members (separation < 0), but all are `changed=false` — inputs ferro preserved rather than produced (e.g. `NM_033028.4:r.[118_261del;118_373del]`). That is an input-quality question, not this defect.

## Why no oracle catches it

The output is **confluent and idempotent**, which is why every existing guard passes:

```
in:  NM_000038.6:c.5265_5268delinsTGCG          -> c.[5265G>T;5266_5268delinsGCG]
in:  NM_000038.6:c.[5265G>T;5266_5268delinsGCG] -> c.[5265G>T;5266_5268delinsGCG]
```

Both spellings converge, and the split form is a fixed point. `FERRO_ASSERT_IDEMPOTENT`, `FERRO_ASSERT_REPARSE`, `FERRO_ASSERT_IN_BOUNDS` and the confluence guards all pass — the sequence denoted is correct throughout. This is purely a **representation** defect, and no current oracle asks whether adjacent members should have been one member.

It is also invisible to diff-based blast-radius measurement: both arms emit it, so it never appears as a moved row. It was found only by auditing absolute output.

## Relationship to existing work

`coalesce_adjacent_pieces` (`src/normalize/merge.rs:5309`) already implements exactly this rule and looks correct in isolation (`pieces[i-1].ref_end == pieces[i].ref_start` on interbase coordinates). So the likely mechanism is that the split producing these members happens **after** that pass, or on a path that does not reach it — not that the predicate is wrong. That needs verifying before a fix, per the usual "confirm the stated cause" rule; I have not yet done so.

Prior art in the same family, all closed: #999 (adjacent insertion+substitution not coalesced), #1229, #1232, #1260. This shape survives them.

## Suggested resolution and its cost

Coalesce adjacent pieces on whichever path currently emits them, so a lone `delins` never splits into touching members.

**Representation stability:** this moves shipped normalized output for 59 distinct descriptions (61 rows) out of 5,761,894 — 0.001%. Every move is split → merged, i.e. toward the input form the corpus already holds in the cases above. It is a representation change and should ship batched with the other pending ones rather than on its own.

nh13:
Stops the normalizer splitting a lone `delins` into two members that sit on *consecutive* nucleotides — a form `DNA/delins.md:16` rules out with no competing clause.

Closes #1524

## The defect

`delins.md:16`: changes involving **two or more consecutive** nucleotides are described as one deletion/insertion. Two members at separation 0 are therefore not a legal decomposition, and unlike the separation-2 family there is no `:44-47` reading that defends them — `general.md:34` does not apply, because separation 0 is not "separated by one or more".

In several cases the input arrived correct and normalization made it wrong:

```
in : NM_000038.6:c.5265_5268delinsTGCG
out: NM_000038.6:c.[5265G>T;5266_5268delinsGCG]        <- 5265 and 5266 are adjacent

in : NM_000083.3:c.2461_2464delinsCTCC
out: NM_000083.3:c.[2461_2463delinsCTC;2464G>C]        <- 2463 and 2464 are adjacent
```

## Cause

Not the adjacent-piece coalescer, which was the first guess. The codon-frame exception in `build_split_variants` matched `[Sub@5266; Identity@5267; Sub@5268]`, **flushed the pending run as its own member**, and emitted the triplet beside it. `canonicalize_from_sequence` never runs for a lone `c.` delins, so nothing downstream repaired it. Instrumented:

```
ref=GTCT alt=TGCG start=5265
subedits=[Sub@5265 G>T, Sub@5266 T>G, IdentityAt@5267 C, Sub@5268 T>G]
```

The fix is two edges in `build_split_variants`: the triplet joins the pending run rather than displacing it, and the exception declines when `p-1` is itself changed — the same precondition `merge::apply_coding_codon_exception` already carries. A right-edge-only variant was measured and rejected; it cost 44 confluence classes against this version's 6.

## Measurements, all on top of this PR's base

**Fixes 58 of 59.** Re-verified independently by adjudicating both arms' output over the 59 known rows: 58 violating → clean, **0 new**. The survivor is `NC_000017.11:g.80110044_80110047delinsGTTGG`, whose violation depends on how a `dup`'s position is read — the softest of the set and the only one that was ever ambiguous.

**Blast radius: 3 rows of 500,004** (0.0006%) on the ClinVar corpus — 2 respell, 1 merge. Row counts validated input == output on both arms. One of the three returns the submitter's own spelling:

```
NM_000083.3:c.2461_2464delinsCTCC  ->  (was) c.[2461_2463delinsCTC;2464G>C]  ->  (now) unchanged
```

## The confluence cost, stated rather than buried

`converged` drops by **6 classes** — 6,633 → 6,629 (3') and 6,628 → 6,626 (5') — and the pinned census is re-blessed in the same commit, which is what that pin exists to force. 11 tests re-blessed and one #1454 guard retargeted.

The affected classes reached agreement **only via the illegal intermediate this fix removes**: `build_split_variants` re-split a member into two touching members, the per-member merge re-joined them into a wider `delins`, and that wider form was where the other spellings had settled. Traced end to end on `s00-c-m3-sep0-p8-rot4`.

Two things make that a price rather than a hidden regression. `sequence_changed` stays **0**, so nothing converged on a wrong sequence. And the form that is lost is the worse one: the single `delins` spans three unchanged nucleotides at `c.21_23` that `general.md:34` says to describe individually, while the two-member form does not. What is lost is agreement, not correctness.

**This was an explicit operator decision**, not a judgement call made inside the change.

## Reviewer notes

**Stacked on #1535**, which is stacked on #1532. Review those first; this diff is the `build_split_variants` fix, its tests, and the re-blessed census.

Full suite: 9594 tests run, 9594 passed. The 11 re-blessings are in the diff, not in a dirty tree.

Representation-Change: 3 rows of 500,004 move (0.0006%) — 2 respell, 1 merge toward the submitted form. Confluence drops by 6 classes of 11,272, deliberately and by operator ruling: those classes agreed only through a spec-illegal intermediate, `sequence_changed` stays 0, and the surviving form is the better-formed one. 58 separation-0 violations of `delins.md:16` are fixed.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -10833,29 +10833,40 @@ fn build_variant_at(
 ///    When `codon_frame_aware` is true, the scan looks ahead at each
 ///    position for a `[Sub@i; Identity@i+1; Sub@i+2]` triplet whose CDS
 ///    endpoints (`hgvs_start + i`, `hgvs_start + i + 2`) share a codon.
-///    Such a triplet emits as a single 3-base `delins` with alt
-///    sequence `[Sub@i.alt, Identity@i+1.base, Sub@i+2.alt]`. The flag
+///    Such a triplet contributes the three positions
+///    `[Sub@i.alt, Identity@i+1.base, Sub@i+2.alt]` to the pending run,
+///    which renders as a 3-base `delins` when nothing adjoins it. The flag
 ///    is true only for `c.` (CDS) variants — `g.`, `n.`, `r.`, and `m.`
 ///    have no codon-frame and skip this branch. The exception is
 ///    deliberately narrow (length-3, exact pattern, in-codon endpoints)
 ///    so it matches the spec text "two variants separated by one
 ///    nucleotide, together affecting one amino acid".
 ///
-/// 2. **Adjacent-substitution coalescence** (`substitution.md`,
-///    issue #182). Consecutive `Substitution` sub-edits whose positions
-///    are strictly adjacent (no gap, no `Inversion` or `IdentityAt`
-///    between them) group into a single `delins` variant — "changes
-///    involving two or more consecutive nucleotides are described as
-///    deletion/insertion".
+/// 2. **Adjacent-change coalescence** (`delins.md:16`, issue #182,
+///    issue #1524). Sub-edits occupying strictly adjacent positions —
+///    no gap, no `Inversion` between them — group into a single `delins`
+///    variant: "changes involving two or more consecutive nucleotides are
+///    described as deletion/insertion (delins) variants".
+///
+///    The codon-frame triplet is inside this rule, not beside it, and that
+///    is what #1524 fixed. The triplet used to be emitted as its own
+///    member, which put two members on consecutive nucleotides whenever
+///    `i - 1` or `i + 3` was also changed. Both edges are now closed, and
+///    differently: on the right the triplet stays in the run so `i + 3`
+///    joins it, while on the left rule 1 **declines** — a changed `i - 1`
+///    means `i` is not one of `general.md:35`'s "two variants" at all.
 ///
 /// 3. **Inversion as a hard barrier** (issue #166). An `Inversion`
-///    always emits standalone and breaks any in-flight substitution
-///    run, preserving the inv-priority decomposition.
+///    always emits standalone and breaks any in-flight run,
+///    preserving the inv-priority decomposition. It can never *be*
+///    adjacent to another member: `decompose_delins` emits one only for a
+///    whole **maximal** contiguous mismatch run, so a neighbouring
+///    mismatch would have been part of the same run.
 ///
 /// Singleton sub-runs stay as `Substitution`. `IdentityAt` not consumed
 /// by a codon-frame triplet drops (an unchanged base is not an edit) and
-/// always ends any in-flight substitution run — the gap means the
-/// surrounding subs are no longer "consecutive".
+/// always ends any in-flight run — the gap means the
+/// surrounding changes are no longer "consecutive".
 fn build_split_variants(
     template: &HgvsVariant,
     subedits: Vec<DelinsSubedit>,
@@ -10865,9 +10876,14 @@ fn build_split_variants(
     let abs = |idx: usize| -> u64 { idx as u64 + hgvs_start };
 
     let mut output: Vec<HgvsVariant> = Vec::new();
-    // Pending run of strictly-adjacent Substitution sub-edits, in
-    // left-to-right order. Each entry is `(position, reference, alternative)`
-    // with `position` the 0-indexed offset into the variant's ref window.
+    // Pending run of strictly-adjacent sub-edit positions, in left-to-right
+    // order. Each entry is `(position, reference, alternative)` with
+    // `position` the 0-indexed offset into the variant's ref window.
+    //
+    // Entries are mismatches except for the unchanged centre a codon-frame
+    // triplet contributes, which enters as `reference == alternative`
+    // (#1524). A run therefore always *begins* and *ends* on a mismatch,
+    // which is the precondition `push_typed_replacement` relies on.
     let mut run: Vec<(usize, Base, Base)> = Vec::new();
 
     let n = subedits.len();
@@ -10897,7 +10913,37 @@ fn build_split_variants(
                 },
             ) = (&subedits[i], &subedits[i + 1], &subedits[i + 2])
             {
-                if *pm == *p1 + 1 && *p3 == *p1 + 2 {
+                // `general.md:35` licenses the merge for "two **variants**
+                // separated by one nucleotide". When `p1 - 1` is itself changed,
+                // `p1` is not a variant: `delins.md:16` makes it part of the
+                // `delins` spanning the run it sits in, and the thing separated
+                // from `p3` by one nucleotide is that whole run — which reaches
+                // beyond the one codon the exception is about. So the pattern
+                // the exception describes is not present and the branch declines
+                // (#1524).
+                //
+                // This is the same precondition `merge::apply_coding_codon_exception`
+                // has always enforced on its own side of the seam
+                // (`is_substitution(&pieces[index - 1])` — the left piece must
+                // be a *lone* substitution). The two are implementations of one
+                // rule and this one was missing the test, so adding it removes a
+                // disagreement between the paths rather than imposing a new
+                // restriction.
+                //
+                // It is also what keeps the joining below from over-merging, and
+                // that was measured rather than reasoned: without this guard,
+                // `c.9_13delinsACAAC` on `TTTTTTTTTAATATAT…`
+                // (`[Sub@9; Sub@10; Identity@11; Sub@12; Sub@13]`, codon 4 =
+                // 10-12) collapses to a single `c.9_13delinsACAAC` spanning the
+                // unchanged 11 — where the two real variants, `9_10delinsAC` and
+                // `12_13delinsAC`, span three codons and are exactly what
+                // `general.md:34` says to describe individually. That form cost
+                // 44 classes of `cis_confluence_axis`'s 3' census
+                // (converged 6633 -> 6589) and 44 of its 5' census, because the
+                // multi-member spelling of each still split.
+                let left_endpoint_is_a_lone_variant =
+                    !matches!(run.last(), Some((previous, _, _)) if *previous + 1 == *p1);
+                if *pm == *p1 + 1 && *p3 == *p1 + 2 && left_endpoint_is_a_lone_variant {
                     let cds_p1 = abs(*p1) as i64;
                     let cds_p3 = abs(*p3) as i64;
                     if merge::same_codon(cds_p1, cds_p3) {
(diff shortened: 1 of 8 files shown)


nh13:
@coderabbitai pause

nh13:
## Review sweep

Reviewed from a detached worktree at `3b1a0eb7`, with a clean baseline built at this PR's actual base (`2ec3b1d3`) rather than reusing a stale target dir. I re-measured the claims independently instead of reading them, and every one I could check reproduces.

### Independently reproduced

| claim | my measurement |
|---|---|
| blast radius 3 of 500,004 on ClinVar | **3 of 500,004** (0.0006%) |
| `NM_000083.3:c.2461_2464delinsCTCC` returns the submitted spelling | reproduced byte-for-byte |
| full suite 9,594 passing | **9,594 run, 9,594 passed**, all three oracles set, nothing dirty |
| 0 new violations, 1 survivor | confirmed, and the survivor is the row you name |

The three moved rows, base → head:

```
NM_001406825.1  [...;2969_2973delinsAGAGC;2974_2976delinsTTA;...]  ->  [...;2969_2974delinsAGAGCT;2976T>A;...]
NM_001354906.2  [4416G>T;4417_4419delinsGCG]                       ->  [4416_4417delinsTG;4419T>G]
NM_000083.3     [2461_2463delinsCTC;2464G>C]                       ->  2461_2464delinsCTCC
```

All three were separation-0 violations at base; all three are legal at head. Nothing else in the corpus moved.

### An independent census of the property, not just of the diff

Rather than trust the movement count, I censused separation-0 member pairs in the *output* on both arms — deriving each member's reference footprint and asking whether any two are consecutive:

| | genuine separation-0 outputs |
|---|---|
| base (`2ec3b1d3`) | **4** |
| head (`3b1a0eb7`) | **1** |

The single survivor is exactly the row the PR names, for exactly the reason it gives:

```
NC_000017.11:g.80110044_80110047delinsGTTGG
  -> g.[80110044C>G;80110045dup;80110047A>G]
```

`80110044` and `80110045` are consecutive only if you read the `dup` as occupying `80110045`, which is the ambiguity the description already flags as the softest of the set. So "58 of 59, survivor is the `dup`-position row" holds on this corpus at the shape level, not merely at the row-count level.

One methodological note in case anyone repeats this: a naive census that treats an `ins` member as occupying its two flanking coordinates reports **11** survivors rather than 1. An insertion has zero reference footprint — `g.[74715628_74715629insTC;74715630_74715631delinsAA]` is separated by `74715629` and is perfectly legal. I made that error first and it manufactures ten false violations, every one of which looks like a real counter-example to this PR. Worth stating because the wrong number is the one that reads as alarming.

### The reasoning I checked rather than took

The left-edge guard is the subtle half, and it holds. `left_endpoint_is_a_lone_variant` declines the codon exception when `run.last()` sits at `p1 - 1`, which is the same precondition `merge::apply_coding_codon_exception` already enforces on its own side — so this removes a disagreement between two implementations of one rule rather than adding a restriction. Since the branch flushes before pushing, `run` is provably empty at the push, and the three entries always enter together, so `flush_substitution_run`'s "a length-1 run is never the identity entry" claim is sound by construction.

The inversion-adjacency argument also holds, and it is load-bearing enough to be worth restating: an `Inversion` cannot end at `p1 - 1` while `p1` is a mismatch, because `decompose_delins` emits an inversion only for a *maximal* contiguous mismatch run, so `p1` would have been inside it. That is what stops the flushed-run path from re-introducing separation 0 by a different route.

### Two small things

**The confluence figure reads as one number when it is two.** The trailer says "Confluence drops by 6 classes of 11,272". It is 4 of 11,272 in the 3' census and 2 of 11,272 in the 5' census — two censuses that each contain 11,272 classes, not 6 out of a single 11,272. The code doc gets this exactly right (`6 633 -> 6 629` and `6 628 -> 6 626`, with the asymmetry explained); only the summary compresses it in a way that slightly overstates the rate. Since the changelog renders the trailer and not the doc, that is the copy worth fixing.

**The branch is stacked on the pre-rebase #1535.** #1535 was rebased onto `main` after #1532 merged (`2ec3b1d3` → `0b936a96`), so this branch still carries `0b5eaa72`, which is now upstream. It will need `git rebase --onto` once #1535 lands; nothing to do yet, just don't be surprised when the stack looks larger than it is.

### What is good here

Naming the cause as the codon-frame exception in `build_split_variants` rather than the adjacent-piece coalescer — and saying the coalescer was the first guess and was wrong — is the part that makes this reviewable. So is measuring the rejected alternative (44 classes against 6) instead of asserting the chosen shape was better, and stating the confluence *loss* in the census doc where the next reader will hit it rather than only in the PR. `sequence_changed` staying 0 is the right invariant to lead with for a change that deliberately gives up agreement.

nh13:
## CI fix pushed

`Python Wheel Test` was red — a required context, so it blocked the merge. One stale expectation in `tests/python/test_core.py`, amended into `3b1a0eb7` (now `26dbeb17`).

```
FAILED test_core.py::TestNormalizeDelinsSubOnlyDecompose::test_coding_embedded_codon_frame_triplet_within_longer_span
  AssertionError: assert '10_12delinsATC' in 'NM_000088.3:c.10_13delinsATCA'
```

It is a **twelfth** guard that encoded the defect, alongside the eleven re-blessed on the Rust side — and it is worth noting *why* it was missed rather than just fixing it. The Rust suite cannot reach it: `cargo nextest run --features dev` does not build the extension module, so a local gate that is otherwise complete (I ran 9,594/9,594 with all three oracles) is structurally blind to `tests/python/`. Catching it needs `maturin develop --features python && pytest tests/python/`, which is a separate build.

The case is exactly this PR's shape: `c.10_13delinsATCA` decomposes to `[Sub@10; Identity@11; Sub@12; Sub@13]`, the triplet fires on 10–12, and `13` is adjacent to `12`. The old expectation asserted `[10_12delinsATC; 13C>A]` — members on consecutive nucleotides, precisely what `delins.md:16` rules out. New output is `c.10_13delinsATCA`.

I rewrote the comment as well as the assertion, since its stated reasoning had become the wrong reason for the right answer — it explained why c.13 is outside the codon exception, which is still true but is no longer why the members do not split:

```python
# It is emitted as one member anyway (#1524). Splitting here would
# put members on c.12 and c.13 — *consecutive* nucleotides — which
# `delins.md:16` rules out ..., and `general.md:34` cannot defend,
# since it governs members "separated by one or more" nucleotides and
# so does not reach separation 0. The codon exception decides what may
# merge, not what may be split apart.
result = ferro_hgvs.normalize("NM_000088.3:c.10_13delinsATCA")
assert result == "NM_000088.3:c.10_13delinsATCA"
```

Also tightened `in` to `==`: the old form would have passed on any output merely *containing* that substring, which for a member-splitting change is the assertion shape least able to notice a regression.

Verified locally with a real wheel — `maturin develop --features python` then `pytest tests/python/`: **738 passed** (was 737 passed, 1 failed). The two sibling codon tests are unaffected and I checked them explicitly: `c.10_12delinsATC` is a bare triplet with nothing adjacent, and `c.12_14delinsACA` splits at separation 1, which stays legal.

nh13:
## Rebased onto the current #1535 tip

`26dbeb17` → `9a9f65fd`, signed. The stack is now one commit on `fix/inv-priority-over-delins`; all five harvest commits (`2ecba7c7`…`0b5eaa72`) are gone.

**One correction to the handover note:** they did not go empty automatically. `2ecba7c7`, `6301df5f` and `9f2c10bd` each conflicted on `hgvs_spec_normalization_tests.rs` / `hgvs_spec_normalization_overrides.json` and needed an explicit `git rebase --skip`. That is expected once you look at why: the base is not plain `main`, it is `main` + #1535's three commits, and #1535 *further edited* the same regions those harvest commits touch. So git sees a three-way conflict rather than an empty patch. I verified before skipping each one — `git diff <commit> origin/fix/inv-priority-over-delins` shows the base already carrying that commit's content plus #1535's additions on top, so the skip drops nothing.

Worth recording because "nothing needs resolving by hand" would lead the next person to `--abort` on the first conflict and assume the rebase was wrong.

**Fixture regenerated before the suite**, per the note — `cargo run --features dev --bin generate_spec_fixture`. Good call: it is exactly the situation that produces the misleading `left: {}`.

**Census pin untouched**, per the note: `converged: 6_629` / `6_626`, `split_two: 4_509` / `4_532`. I had independently reached the same conclusion — the 4-and-2 split is documented correctly in the census doc itself, so there was nothing to re-bless.

Gate on the rebased tree, reading the Summary line rather than the exit code:

```
Summary [ 110.508s] 9594 tests run: 9594 passed (5 slow), 145 skipped
```

`cargo fmt --check` clean, `clippy --all-features --all-targets -D warnings` clean. The Python re-blessing from the previous push is carried through the rebase (`tests/python/test_core.py` still in the diff).

nh13:
## Rebased onto `main` — #1535 has landed

`9a9f65fd` → `c18fb594`, signed. The PR now targets `main` directly and is **one commit**; #1535's commits are gone (they are in `main` via its squash, so they needed an explicit `--skip` again — same pattern as the harvest commits, and for the same reason).

**The census pin holds, unchanged.** This was the thing worth re-checking rather than assuming, since #1535 landed underneath and it also changes inversion typing. With the fixture regenerated first:

```
Summary [ 114.870s] 9599 tests run: 9599 passed (5 slow), 145 skipped
```

`converged: 6_629` / `6_626` and `split_two: 4_509` / `4_532` are still correct against the new base — nothing re-blessed, exactly as your note said. `cargo fmt --check` clean.

The `tests/python/test_core.py` re-blessing from earlier is carried through the rebase intact.

**Still live: the #1484 ordering hazard.** Both this PR and #1484 re-bless `tests/it/cis_confluence_axis.rs`. This one is now rebased and green, so if it merges first #1484 will need its census re-blessed against post-#1524 numbers; if #1484 goes first, this pin needs re-measuring against post-#1484 numbers (#1484 alone moves `converged` 6,633 → 8,006). Merging this one first is the cheaper order — it is one commit on current `main` with nothing else waiting behind it.

PR #1538, opened 2026-08-07
docs(rulings): restore the delins.md:47 decision dropped by #1532's squash

nh13:
Restores an operator ruling that was decided and committed, but did not reach `main`.

## What happened

The `:47` decision was pushed to `docs/record-adjudication-harvest` as `a4cd062c`, but #1532 was squash-merged from a state that predates that push. So `main` carries the three new `undecided` records from that PR while `delins-merge-vs-individual-gap-two-or-more` still reads `undecided` — the decision itself was dropped. This is a straight cherry-pick of the missing commit.

The same commit also rides on #1535 (as `4c62b091`). Once this merges it should go empty there and drop on the next rebase; if it conflicts instead, the content is identical and either side is correct.

## The ruling

**`DNA/delins.md:44-47` governs `:17` — scoped to the shape `:44-47` actually describes.** The scope limit is part of the ruling, not a gloss on it: it settles only the case where a *minimal* single `delins` would be split because payload bases coincide with reference bases. It is **not** a general licence to merge changes separated by two or more nucleotides. Where a separation of two or more arises from anything else, `general.md:34` still governs — in particular rows on an axis with no reading frame, where `general.md:35`'s codon exception cannot apply and no competing clause exists.

The scope matters because the evidence base and the reach are different sizes. The argument was made on 208 genomic-axis rows; a later corpus-wide audit of the split direction (5.76M rows, real reference bases) found the same reading, read unscoped, would reach roughly 2,112 of 4,732 measured violations. Scoped, it ratifies the class the evidence covers; unscoped it would retroactively declare those conformant on evidence that never examined them.

## Grounds

1. The spec builds the split itself and then rejects it: `:46` offers `c.[850_869del;874_881del;887_897del;901_902insG]` as the alternative description, and `:47` answers that the delins format is recommended.
2. Proof by contradiction — the separations inside that rejected alternative are 4, 5 and 3, every one at least two. So `:17` read as reaching alignment-driven splits would demand precisely the description `:47` advises against, three lines later, in the spec's own worked example.
3. Provenance, which `:79-84` gives as the *reason* for `:17`, points the other way in this class: an independent 208-row arbitration found 121 rows that are a minimal single delins split only on payload coincidence, and 121 of 121 were submitted as a single-member variant.
4. Non-uniqueness — 65.3% of the candidate splits are not unique, up to 1,287 equally-optimal alternatives, so adopting one trades a unique canonical form for an arbitrary pick from a family.
5. Cost: v0.12.0 already emits the unsplit form on 208 of 208, so this **moves no row**.

The rationale also records what the ruling does *not* claim: neither clause is normative (the RFC 2119 census in the same record stands), `:44-47` states a preference rather than an algorithm, and ferro still declines to read input provenance at runtime — it is used here as evidence for a fixed canonical form, never as an input.

## Verification

`RULING_STATUSES` is updated in the same commit, which `ruling_records_are_intact` enforces. Full suite: 9582 tests run, 9582 passed, nothing re-blessed.

Representation-Change: none. The ruling ratifies shipped behaviour — v0.12.0 already emits the unsplit form on 208 of 208 adjudicated rows — and this PR changes no code.


nh13:
@coderabbitai pause

nh13:
## Review sweep

Reviewed from a detached `review-1538/`. The cherry-pick is faithful and the ruling reads correctly. One finding, fixed and amended into `4f570a56` (now `89b46c38`).

### This PR silently makes `CLAUDE.md` wrong

`CLAUDE.md`'s "Adjudication records" section is **currently accurate on `main`** — I checked, and `main` really is 5-of-6 `undecided`. This PR takes it to 4-of-6 while the doc still says:

> **Five of its six records are `undecided`**, and the first three below are operator decisions that block other work

and still tabulates `delins-merge-vs-individual-gap-two-or-more` under **what is open**, with `adjudication-precedence-order` described as deciding "both rows above" — one of which this PR settles.

That matters more here than it would in most PRs, because this section *is* the decision log's front door, and its stated purpose is to stop the next reader re-deriving a settled argument. Leaving it would mean the change that restores the ruling is also the change that tells readers the ruling was never made.

It is also the reason this belongs here rather than on #1535. The contradiction arrives with **whichever** of the two lands first, and since this one is a straight cherry-pick to `main` it gets there first — so fixing it on #1535 would leave `main` briefly wrong with no PR accountable for it.

Amended in:

- count `Five` → `Four`, `first three` → `first two`
- `delins-merge-vs-individual-gap-two-or-more` lifted out of the open table
- `adjudication-precedence-order`'s row re-pointed: it now decides *the row above*, plus the reading behind this record's scope limit — which is still genuinely open even though the record is decided
- a short paragraph naming both `decided` records and carrying the **scope limit** forward, since the doc is where someone will cite the ruling from and the unscoped reading is the failure mode the record itself warns about

### Verified rather than taken

| claim | measured |
|---|---|
| 9,582 tests, nothing re-blessed | `Summary [ 81.984s] 9582 tests run: 9582 passed` |
| `RULING_STATUSES` tracks the fixture | `ruling_records_are_intact` green |
| record count after the change | 4 of 6 `undecided`; `decided` = `delins-codon-carve-out-gap-one`, `delins-merge-vs-individual-gap-two-or-more` |

Fixture regenerated before the run (`cargo run --features dev --bin generate_spec_fixture`) — a stale one fails `ruling_records_are_intact` with a `left: {}` that reads like the records were destroyed, which is exactly the trap this PR's subject matter invites.

`Representation-Change: none` is right and well-supported: no code changes, and the ruling ratifies shipped behaviour (208 of 208 already unsplit at v0.12.0).

### On the knock-on to #1535

Once this merges, #1535's `4c62b091` should go empty. Worth setting expectations from what actually happened rebasing #1537 an hour ago: the five harvest commits there did **not** drop automatically — they conflicted and needed explicit `git rebase --skip`, because the base was `main` *plus* commits that further edited the same regions. Same shape applies here, so if `4c62b091` conflicts rather than vanishing, that is expected and the content is identical either way.

Issue #1539, opened 2026-08-07
Normalizer: splitting a lone delins can emit a member that itself violates the separation rule


nh13:
## Summary

When the sequence-first pass splits a lone `delins` into an allele, it does not check that the members it emits are themselves conformant. On five real ClinVar/CMRG rows the input was a single `delins` that ferro `main` returned **unchanged and clean**, and the split produces a member holding a forced unchanged nucleotide between two changed ones, straddling a codon boundary — the shape `general.md:34` says to describe individually and which `general.md:35`'s codon exception cannot rescue.

These are regressions introduced by the split, not pre-existing defects surfaced by it. That distinction was measured: of 168 rows that newly carry a violating member under the widened axis gate, **163 had a parent that was already violating** (the split repaired them incompletely) and **5 had a parent that was clean**. This issue is those 5.

## The cases

Five rows, three distinct variants — two pairs are the same variant on different transcripts.

```
in    : NM_000089.3:c.2148_2156delinsACGTGG
main  : unchanged                                     (clean)
split : c.[2148_2152delinsAC;2155_2156inv]
        member c.2148_2152delinsAC   ref=GGTCG -> AC
        forced unchanged C at c.2151; changes at c.2150 (codon 716) and c.2152 (codon 717)
        => straddles a codon, so :35 does not apply, so :34 requires a further split

in    : NM_000089.4:c.2148_2156delinsACGTGG            (same variant, .4)
in    : NM_001277115.2:c.3545_3551delinsAGATGATGAACTGTAGCTGTAAGA
main  : unchanged                                     (clean)
split : c.[3545_3548delinsAGAT;3551delinsTGAACTGTAGCTGTAAGA]
        member c.3545_3548delinsAGAT  ref=GACA -> AGAT, forced gap at offset 1

in    : NM_000296.4:c.10317_10334delinsATGGCC
main  : unchanged                                     (clean)
split : c.[10317_10330delinsAT;10334T>C]
        member c.10317_10330delinsAT  ref=CCAGGCGGGCCATG -> AT, forced gap at offset 12

in    : NM_001009944.3:c.10320_10337delinsATGGCC       (same variant, other transcript)
```

All five are the same class: **`sep1-straddles-codon` inside a member the split created**.

## Why the existing oracles miss it

[...]

PR #1540, opened 2026-08-07
test(normalize): guard that emitted members obey general.md:34 (#1539)

nh13:
Adds a per-PR blocking guard for #1539, and it is **red on this base by design**: all five inputs currently normalize to a form whose members violate `general.md:34`. It goes green when #1484 stops emitting those members.

Refs #1539. **Stacked on #1484** (`feat/issue-1235-widen-axis-gate`) — review that first.

## What the guard asserts

The property, not the string: **no member ferro emits may hold a reference column that is unchanged in every minimal alignment and separates two changed columns**, unless `general.md:35`'s codon exception applies — both changes in one codon, on a reading-frame axis.

Pinning today's five outputs would freeze the current behaviour including what is wrong with it, which is the change-detector the repo's adjudication policy rules out. The failure names the offending member, its bases, its codons, and which clause is unavailable:

```
VIOLATION NM_000089.3:c.2148_2152delinsAC (ref=GGTCG payload=AC):
  one forced-unchanged reference base at offset 3 separates changes in
  codons 716 and 717; general.md:35's one-amino-acid exception does not
  reach across a codon boundary
```

Three correctness requirements, each pinned by its own unit test because each has already produced a wrong published number in this project:

- an insertion consumes no reference position, so its interval is empty (`lo = A+1, hi = A`);
- a column counts as unchanged only when **every** minimal alignment matches it — forward and backward Levenshtein, under-reporting rather than over-reporting;
- the codon test runs on **every** interior gap, not just the first. A `runs[0]`-only version under-counted by 54 per arm.

An untrimmed member is **deferred**, not judged, and a deferral fails the gate rather than counting as clean.

## Why the fixture is served whole, not as captured windows

The first version of this guard was **green on three of the five cases while the defect was fully present.** Worth recording, because the mechanism generalises.

`WindowProvider::is_known_contig` returns true for any key in the fixture's `genomic` map, and the normalizer reaches a transcript through `get_genomic_sequence` as well as `get_transcript` — so the recorder filed those reads under the transcript's own accession, registering `NM_000089.3` as a *contig*. Every `get_sequence` for it then routed through the narrow captured window instead of the whole sequence sitting in `transcripts`.

Measured, sweeping 200 bp blocks through both entry points: **28 of 28 blocks failed under the fixture and succeeded under the manifest, on all five transcripts.**

The general trap: **a window captured from a pass is exactly as wide as that pass read.** It serves the recorded pass and fails any pass that reads further. On `main` the recorded pass stays inside it; a wider-reading pass errors, bails, and returns the input unchanged — which for a lone `delins` is indistinguishable from a conformant answer. The biocommons fixture never hits this because its `genomic` keys are real contigs.

`serve_transcripts_whole` now widens any window whose accession is also a stored transcript to `[0, len)`. Fixture 100 KB → 155 KB.

## Three fidelity checks, so this cannot regress silently

1. **`the_committed_windows_serve_every_transcript_whole`** — hermetic, CI. Verified to catch the defect: restoring the old narrow fixture makes this fail while the gate still passes.
2. **A zero-failed-reads audit inside the gate.** `AuditedProvider` records every failed read and the gate asserts none *before* reporting verdicts, so a fixture defect can never be reported as conformance. This is the one that fires on a wider-reading pass, where the structural check alone would not.
3. **`hermetic_normalization_matches_the_prepared_reference`** — manifest-or-skip ground truth.

Real-reference answers are deliberately **not** pinned as expected strings: they legitimately change on #1484, so a pin would go red on a representation change and report it as a fixture defect — the exact misdiagnosis this guard exists to prevent.

## Also here

`fix(conformance)`: `WindowProvider`'s window-miss error told every caller to regenerate with `extract_biocommons_windows`. Four fixtures share that provider, so for three of them it named the wrong file.

## Verification

`cargo fmt --all --check` and `cargo clippy --all-features --all-targets -- -D warnings` clean. On `main` the whole module is green (16/16) and all five inputs normalize to themselves — confirmed both via the manifest and hermetically. On this base, 15/16 with the gate red on 5 of 5, which is the intended state.

Representation-Change: none. Tests, examples, one conformance schema module and an error-message fix; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched and no normalized output moves.


nh13:
@coderabbitai pause

nh13:
Dual-review sweep. Lens run against `origin/feat/issue-1235-widen-axis-gate` (the stacked base, not `main`). Six findings; **three confirmed**, and they share a theme this PR is already the world expert on.

First, to set expectations: the red `Test (3/6)` is **not** a defect and I am not reporting it as one. The guard fires 5-of-5 exactly as designed, and it cannot merge while red because `Test` is required — which is correct for a specification.

## The theme: the fixture generator has two more silent-incompleteness modes

This PR's body records, at length, a fixture defect that made the guard **green on three of five cases while the defect was fully present** — because a window captured from a pass is exactly as wide as that pass read. The hardening against that is good (`serve_transcripts_whole`, the structural check, the zero-failed-reads audit). But the *generator* keeps two adjacent ways to emit a fixture that looks complete and is not.

**1. `run_input` discards the `catch_unwind` result** (`examples/extract_split_member_separation_windows.rs:76-89`):

```rust
fn run_input(provider: &RecordingProvider, input: &str) {
    let _ = catch_unwind(AssertUnwindSafe(|| { ... }));
}
```

A panic mid-capture is swallowed, that input contributes whatever it recorded before dying, and generation proceeds to write. The three fidelity checks do not cover this: the structural check asks whether *captured* transcripts are served whole, and the runtime audit asks whether the *gate's* reads failed — neither asks whether capture itself completed.

**2. Nothing asserts the capture is complete** (`:210-247`). The count is *printed*, never checked:

```rust
println!("… {} inputs, {} transcripts, {} genomic windows …",
         inputs.len(), fixture.transcripts.len(), …);
```

then straight to `to_json` / `--check` / `write`. A capture that silently produced 4 transcripts for 5 accessions renders and commits without complaint. One assertion — exactly one transcript per distinct accession in `inputs.txt` — closes it, and given the history here it belongs before `to_json` rather than as a test.

Together these matter more than their individual severity: the recorded near-miss was found by hand, after the fact. These two would produce the same class of artefact through a different door.

**3. The `AuditedProvider` comment and implementation disagree** (`tests/it/issue_1539_split_member_separation.rs:440-442`). `get_protein_sequence` returns a `Result` like the other audited reads but is delegated directly, while the comment scopes direct delegation to the `has_*` / `*_length` probes. Either route it through the recording logic or name it in the comment. Small, but this audit is load-bearing — it is the check that catches a wider-reading pass, which the structural check cannot.

## Noted, not objected to: the fourth committed-fixture family

`tests/fixtures/split-member-separation/` (155 KB) is a **new** committed fixture, and `.coderabbit.yaml`'s `tests/**/*.rs` instruction says "Generate test data programmatically; do not add a NEW committed fixture", naming three deliberate exceptions.

I think this is a legitimate fourth. Real GRCh38 bases cannot be generated programmatically, and the manifest-backed alternative does not run on PR CI at all — the same rationale that justifies `tests/fixtures/validation/`. The PR pre-empts the objection thoroughly. Flagging only so the instruction's exception list can be updated deliberately rather than drifting: it currently says *three* families "are deliberate and must not be flagged", and a reader will now find four.

## Rejected

- **A manifest/reference digest in `WindowFixture`** (trivial). Reasonable in the abstract, but it is a change to the *shared* schema and its sibling generators, which is a wider blast radius than this PR's scope and would touch fixtures it does not own.
- **`resolve_manifest` losing the explicit `--manifest` path** (trivial). Real but cosmetic: the error says "pass `--manifest`" when you did pass one. Worth a line if you are in there anyway.

## Also noted

`INPUTS_TXT` / `WINDOWS_JSON` (`:50-51`) are repo-relative strings rather than `CARGO_MANIFEST_DIR`-based. This works — `cargo test`/`nextest` set the CWD to the package root — so I am not calling it a defect. Worth knowing that the layout has been bitten by the *inverse* problem (a stale baked `CARGO_MANIFEST_DIR` after a worktree rename), so neither form is unambiguously safer here.

Issue #1536, opened 2026-08-07
Normalizer: a delins crossing a CDS boundary is never re-derived, giving two fixed points for one variant


nh13:
## Summary

A lone `delins` whose span crosses a CDS boundary is never re-derived from the sequence it denotes, so two spellings of one variant reach two different fixed points. The identical block placed entirely inside the CDS converges correctly. The discriminator is nothing but where the start or stop codon happens to fall.

## Reproducer

Same design, same axis, same transcript — only the offset moves:

```
c.5_12delinsTGTGCATT   ->  c.5_12inv        # entirely within the CDS: converges
c.9_*1delinsTGTGCATT   ->  unchanged        # crosses the stop: does not
```

Both blocks are exact whole-block reverse complements of their reference. The first is re-derived and typed `inv`; the second is left as the `delins` it arrived as, so `c.9_*1inv` and `c.9_*1delinsTGTGCATT` are both fixed points.

The same holds at the 5' boundary (`c.-3_1delinsTGGT`) and on a multi-exon transcript.

## Cause

`join_pos` (`src/normalize/merge.rs:1085-1087`) refuses any interval whose two ends are in different regions:

```rust
if rs != re {
    return None;
}
```

`simple_cds_pos` classifies a position as `Cds`, `FivePrimeUtr` or `ThreePrimeUtr` and returns the axis coordinate *within that region* — `*1` is `1`, `-3` is `-3`, `9` is `9`. Those numbers are not on one line, so `edit_span_union` and the `w_lo`/`w_hi` window arithmetic just above `crosses_exon_junction` cannot be done in axis coordinates across a boundary. `join_pos` therefore declines, `collect_canonical_edits` returns `None`, and the whole sequence-first pass is skipped for the variant.

That refusal is structural rather than deliberate: unlike the exon-junction clamp a few lines further down, which carries an explicit rationale and states its confluence cost, this one is an unannotated consequence of doing window arithmetic in axis space.

A fix would convert the span to transcript coordinates for the window computation and back afterwards, the way the projector already does — the region split is a rendering concern, not a geometric one.

## Measurements

**Synthetic corpus** (the `separated_revcomp_runs` family, 15 designs x 3 spellings x 6 axis/direction pairs): 38 of 90 confluence classes fail, and they partition perfectly on this one variable.

| block crosses a CDS boundary | lone `delins` re-derives to `inv` | classes |
|---|---|---|
| no | yes | 12 |
[...]

PR #1544, opened 2026-08-07
test(spec): run the spec's own worked examples against real bases, and record #1536

nh13:
The HGVS recommendations publish their own answers for two normalization rules — the coding-sequence restriction on repeat notation ("use `NM_024312.4:c.2692_2693dup` and **not** `c.2686A[10]`") and the 3'rule's exon/exon-junction exception ("described as `LRG_199t1:c.3921dup` … and not as `c.3922dup`"). Neither had ever been executed, because both are sequence-dependent and every existing home for them is sequence-blind: the harvested spec fixture normalizes through an empty `MockProvider`, every family in `dump_normalized_corpus` is single-exon, and the manifest-backed axes skip when the prepared reference is absent — in CI, always. So the flagship worked example of `general.md:44` had never run.

This adds its blocking home, plus one recorded defect and two adjudication records.

## What is here

**`tests/it/spec_worked_examples.rs` — twelve rows served real bases.** A `WindowProvider` over a committed 40 KB reference slice, so the gate cannot skip. Each row is pinned under all three shipped input-hygiene modes — twelve rows x three modes = 36 pinned results — because the modes disagree here and *which mode gives which answer* is the finding:

- `r.-6_-3g[6]` — the library path leaves the spec's published form alone, lenient strips the redundant unit label (W3013) and emits `r.-6_-3[6]`, strict rejects the input. So the single divergence from the spec belongs to the preprocessor's repair, not to the normalizer.
- `c.1738TA[6]` / `r.1738ua[6]` — strict mode rejects two of the spec's own published inputs, because the reference tract is not a whole number of `TA` copies. Recorded as observed, not endorsed.
- `c.3921dup` / `c.3922dup` denote one transcript sequence (both positions are `T`) but are two fixed points projecting 2,790 bp apart. Each row's own output is pinned; they are deliberately **not** asserted equal, since whether they should converge is an open interpretive question — see the ruling record below.

Every row carries a `file:line` citation with a verbatim quote checked against the vendored spec checkout at run time, per the adjudication-records policy: a pinned output with no authority behind it is a change detector, not a record.

The slice is generated, never hand-edited — `cargo run --features dev --example extract_spec_worked_example_windows -- --manifest <manifest>`, with `--check` for local drift.

**`tests/it/issue_1536_cds_boundary_delins.rs` — one ignored-RED guard against #1536.** A lone `delins` whose span crosses a CDS boundary is never re-derived from the sequence it denotes, so two spellings of one variant reach two different fixed points. `join_pos` (`src/normalize/merge.rs:1085-1087`) refuses any interval whose ends are in different regions, `collect_canonical_edits` returns `None`, and the whole sequence-first pass is skipped. Everything is held constant except the boundary: one 20-base transcript, one 8-base block at transcript positions 5..=12 (`AATGCACA`), one payload (`TGTGCATT`, its exact whole-block reverse complement), and only the CDS annotation moves.

```
CDS 1..=20   c.5_12delinsTGTGCATT   -> c.5_12inv     (converges)
CDS 1..=8    c.5_*4delinsTGTGCATT   -> unchanged     (does not)
CDS 8..=20   c.-3_5delinsTGTGCATT   -> unchanged     (does not)
```

**`a_delins_crossing_a_cds_boundary_converges_with_its_inv_spelling` is `#[ignore]`d and RED on purpose, and must not be weakened to green.** It records #1536; the defect is unfixed. Un-ignore it *with* the fix, in the same change. Verified still red after rebasing onto `5616cdb9`, for its stated reason and no other — 4 of 4 boundary-crossing pairs diverge:

```
#1536: 4 boundary-crossing spelling pair(s) reached two fixed points for one variant:
  CDS 1..=8: NM_TEST.1:c.5_*4delinsTGTGCATT -> NM_TEST.1:c.5_*4delinsTGTGCATT, but NM_TEST.1:c.5_*4inv -> NM_TEST.1:c.5_*4inv
  CDS 8..=20: NM_TEST.1:c.-3_5delinsTGTGCATT -> NM_TEST.1:c.-3_5delinsTGTGCATT, but NM_TEST.1:c.-3_5inv -> NM_TEST.1:c.-3_5inv
```

Two companions are **not** ignored and do gate: the in-CDS control (which also asserts the `delins` is re-derived and typed `inv`, not merely that both spellings agree — two frozen spellings would satisfy an equality check and prove nothing), and a check that the payload really is the block's exact reverse complement, so a fix can never be credited to a change of geometry.

**Two `undecided` ruling records**, both falling out of running the worked examples against real bases, and neither ferro's to answer — so both state both readings verbatim rather than changing behaviour:

- `exon-junction-dup-converge-from-the-far-side`. `DNA/duplication.md:26` names `c.3921dup` as *the* description "and not `c.3922dup`", and `:148` gives a correctness reason. But `general.md:44` only *suspends* the 3'rule around the junction; it nowhere prescribes shifting 5' back, and `c.3922dup` may legitimately describe a different genomic event yielding the same transcript. Settling it would move every stored `c.3922dup`-shaped description, so it is an operator decision.
- `rna-repeat-range-plus-unit-redundancy`. `RNA/repeated.md:22` marks range-plus-unit invalid as redundant; `:27`, five lines later, publishes `NM_024312.4:r.-6_-3g[6]` — exactly that shape — as valid. Ferro answers both ways depending on input-hygiene mode. Recorded rather than reconciled; owed upstream via #466.

`RULING_STATUSES` is kept in sync (`ruling_records_are_intact` pins ids *and* statuses) and CLAUDE.md's open-questions table now lists eight records, five `undecided`.

## CI change: `submodules: true` on the `test` and `test-oracle` jobs

This PR modifies `.github/workflows/ci.yml`. `test` and `test-oracle` are the two jobs that run the `it` binary's non-proptest, non-sweep tests, and **neither checked out `assets/hgvs-nomenclature`**. The worked-examples gate resolves each row's `file:line` citation against that checkout at run time and asserts the quote verbatim, so without the submodule the citation test *fails* rather than skips. Failing is the right behaviour for it — a citation check that quietly no-ops when the thing it verifies against is missing is exactly the failure mode it exists to prevent — so the fix is to provide the checkout, not to soften the test. It is 4.4 MB.

`soak` and `sweeps` also run the `it` binary but select `test(proptest)` and `$SWEEP_FILTER` respectively, so they never reach the module and are left alone. `test-build` and `test-build-soak` keep their existing "no submodules" notes: they compile, they do not run.

**Expected side effect, by design:** this changes the `ci.yml` hash, and several cache keys digest that file, so the next run of those jobs is one cold rebuild. That is the deliberate trade for caches that cannot go stale; the run *after* it is the real number.

## Verification

Rebased onto `5616cdb9` and re-gated after the rebase (the rebase took two mechanical conflicts: an additive `mod` line in `tests/it/main.rs`, and the ruling-count sentence in CLAUDE.md, corrected to the true post-rebase count of five `undecided` of eight).

- `cargo nextest run --features dev -E 'binary(it)'` — `Summary [37.128s] 5452 tests run: 5452 passed, 146 skipped`
- `--run-ignored all` on the #1536 module — 2 passed, 1 failed, the failure being the ignored-red guard above, for its stated reason

Note for reviewers rebasing this locally: `tests/fixtures/grammar/hgvs_spec_normalization.json` is gitignored and only regenerates when *missing*, so a stale copy survives a rebase and makes `ruling_records_are_intact` fail against statuses `main` has since decided. Re-run `cargo run --features dev --bin generate_spec_fixture`.

Refs #1536, #466.

Representation-Change: none. Nothing in this branch can move a normalized output. The only `src/` change is additive — one new fixture-loader module (`src/conformance/spec_worked_examples.rs`) plus its `pub mod` line; no normalizer, projector, parser or formatter code is touched. The only edit to a committed fixture adds two `rulings` records, both `undecided`, which are metadata and feed no normalization path. Everything else is new tests, a new generator, a generated reference slice, docs, and the CI checkout flag. The twelve worked-example rows *pin* outputs across three input-hygiene modes; pinning records the current answer, it does not change it — and all 36 pinned results were produced by, and agree with, `main` at `5616cdb9`. Correspondingly the whole suite passes with nothing re-blessed. No corpus-count evidence is quoted here because none is needed: this is not a `0 moved` measurement, it is a change with no code path to move anything.


nh13:
@coderabbitai pause

nh13:

Both lenses ran to completion against `c04e6ad7` (CLI: `--base origin/main`, **9** findings, no rate-limit; agent lens: independent, from a detached `review-1544/` with the submodule initialised). CI is green: 37 rows, 36 pass, 0 fail.

I verified the PR's central claim first, independently of the test that asserts it: **all 12 citations resolve verbatim** against the vendored checkout at the pinned `spec_commit` (`6f85311`). The corpus is what it says it is.

Four confirmed findings, ranked. The first is the one I'd act on before merge.

---

### 1. ⚠️ The gate is two-thirds blind to a starved reference slice — measured

**`tests/it/spec_worked_examples.rs`** · the false negative #1540 exists to close, present here · *agent lens*

The PR's selling point is that these rows are served real bases and *cannot skip*. But **22 of 36 pinned results are fixed points** — `default/lenient/strict_output` byte-identical to `input`. For those, "reads fail → normalizer bails → returns input unchanged" is indistinguishable from the pinned correct answer.

Measured, not inferred. Truncating **both** transcript sequences in `reference-windows.json` to 100 bases — a maximally broken slice:

```
the_spec_worked_examples_produce_their_recorded_answers   12 of 36 pinned results moved
                                                          → 24 of 36 stayed GREEN
every_worked_example_answer_is_a_fixed_point              PASS (vacuous: an unchanged
                                                          input is trivially a fixed point)
```

Three of six tests do go red, so the crude case is caught — but the pinning test, which is the corpus's actual contract, misses two-thirds of its rows, and the fixed-point test cannot fail for the reason it exists.

The remedy is already written, by you, today: **`AuditedProvider`** in #1540 wraps the `WindowProvider` and asserts *no reference read failed while producing the answer*. #1540's own doc comment states the general lesson — "an under-serving slice is indistinguishable, at the output, from a clean normalization." That is exactly this gate's exposure. Reusing it here would make the slice's integrity load-bearing instead of incidental.

Related, and the mechanism behind it: the generator (`extract_spec_worked_example_windows.rs:81-89`) records **one** pass per row — default mode, no member walk, no re-normalization of the output. The gate replays **three** modes plus a fixed-point re-normalization of a *different input string*. The capture is narrower than the replay by construction; it passes today by overlap, not by design.

### 2. ⚠️ A blank citation quote passes the "verbatim" check vacuously

**`src/conformance/spec_worked_examples.rs:140`** · *found by both lenses independently*

`cited.contains(&self.quote)` is unconditionally true when `quote` is `""`. Verified by blanking row 0's quote and running the gate:

```
every_row_cites_the_spec_verbatim ... PASS      # with an empty quote
```

`Citation`'s own doc comment names the hazard this creates: *"a bare line number resolves against any file long enough to have that line, so bumping the spec submodule could silently leave a citation pointing at unrelated prose."* An empty quote reduces the citation to precisely that bare line number. The type documents the failure mode and does not enforce it.

Note the asymmetry — `every_row_cites_the_spec_verbatim` already guards `!case.note.trim().is_empty()`, but not the quote, which is the load-bearing field. Same shape as #1523's empty-value trailer regex, where `\s*(?P<value>.+?)\s*$` matched a lone space and a bare `Representation-Change:` counted as a declaration.

I'd **reject the path-traversal half** of the CLI's version of this finding: `spec_dir` is a vendored submodule and `clause` comes from a committed fixture, so there is no untrusted input and no boundary to cross.

### 3. ⚠️ The ignored-red #1536 guard can go green on the wrong answer

**`tests/it/issue_1536_cds_boundary_delins.rs:176-195`** · *CLI lens, confirmed*

The guard asserts only `from_delins == from_inv` — convergence, not convergence *on `inv`*. A future fix that made both spellings converge on any non-`inv` form would turn it green while leaving the actual defect (failure to re-derive and type as an inversion) unfixed.

This contradicts the standard the PR sets one function earlier. `a_delins_inside_the_cds_converges_with_its_inv_spelling` asserts both convergence **and** `from_delins == inv`, with the comment: *"Two spellings that both froze would satisfy an [equality check]"*. The reasoning is right; it just isn't applied to the guard that will actually be un-ignored when #1536 is fixed.

`PLACEMENTS` already carries the expected `inv` string in position 4 of every row, so the fix is local — require `from_delins == *inv && from_inv == *inv`. The CLI's suggested patch is correct in substance.

### 4. 🔸 The PR body's cold-rebuild side effect does not exist

> *"this changes the `ci.yml` hash, and several cache keys digest that file, so the next run of those jobs is one cold rebuild"*

Neither half holds against the current workflow:

- **`ci.yml` contains no `hashFiles` at all.** #1218 replaced the hand-rolled `actions/cache` keys with `Swatinem/rust-cache`, which keys on toolchain + `Cargo.lock` + its `key:` label — never the workflow file's contents.
- **`test` and `test-oracle` have no cache step whatsoever.** They download the prebuilt test archive; they do not compile.

So adding `submodules: true` costs the 4.4 MB checkout and nothing else — strictly better than advertised. Worth correcting because the note reproduces guidance the migration made obsolete, and a reader will otherwise expect a cache generation to rotate.

(The body's claim about `test-build`/`test-build-soak` keeping their no-submodules notes **is** accurate — I initially misread those as already enabled; the hits were the `# No submodules: true` comments, not settings.)

---

### Also noted, not blocking

- **`--check` is documented without the `--manifest` it requires** (`:23`). *CLI [8].* `run()` returns `FAILURE` with `no manifest (pass --manifest or set FERRO_MANIFEST)`, so the documented Check command fails as written unless `FERRO_MANIFEST` is already exported. The Run line above it gets this right.
- **`read_to_string(WINDOWS_PATH).unwrap_or_default()`** (`:117`) folds every I/O error into "empty", so an absent fixture reports "OUT OF DATE" and sends the reader hunting for drift in a file that is not there. *CLI [9], and the identical nit you fixed in #1540 at `34177dd7`* — same treatment applies: match `ErrorKind::NotFound` explicitly, propagate the rest.
- **`spec_commit` is dead metadata.** Declared at `spec_worked_examples.rs:44`, read nowhere (`rg spec_commit -g '*.rs'` → one hit, the field itself). It currently matches the submodule, but nothing keeps it there, so it can silently become a false claim about what the citations were checked against. One assertion against the checked-out SHA would make it real.
- **`Fixture::cases` lacks a doc comment** while its two siblings have one. *CLI [1].* Trivial, but the convention is docs on all public items.

### Rejected, with reasons

- **Populate `applies_to` / `equivalence_classes` on the two new `undecided` records.** *CLI [2], [3].* The new records match the established convention exactly: all three pre-existing `undecided` records carry `0`/`0`, and only `decided` records populate those fields (`2`/`1`, `1`/`0`, `6`/`3`). Asking this PR to deviate would make it the odd one out. If populating them for `undecided` records is desirable, that is a repo-wide convention change and its own PR.
- **Add a required manifest-digest field to the reference fixture schema.** *CLI [5].* `WindowFixture` is shared across four committed fixture families; a required new field is a cross-cutting migration with a blast radius well beyond this PR. Same reason I rejected it on #1540.
- **Strengthen `captured_from` beyond the manifest basename.** *CLI [7].* The point is fair — `"manifest.json"` distinguishes nothing — but it is **pre-existing and repo-wide**, not introduced here. Across the four committed window fixtures the field is already split: `spec_enumeration_windows.json` and `mutalyzer-normalize/genomic-windows.json` both carry `"manifest.json"`, while `biocommons-normalize` carries a `prepared_at` timestamp. This PR follows the majority. Worth flagging that **#1540 chose the other form the same day** (`read_prepared_at(&manifest_path)`), so the repo now has both conventions live in new code — standardising is a cheap follow-up (the timestamp form needs no schema change), and I'm happy to file it if you want it tracked.

---

**Spec audit: clean.** The two new `undecided` records state both readings verbatim rather than picking one, which is the correct disposition for a spec self-conflict ferro does not own — `RNA/repeated.md:22` marks range-plus-unit invalid as redundant and `:27` publishes exactly that shape as valid, five lines apart. `Representation-Change: none` is accurate: `src/conformance/` sits outside the four watched directories, so the trailer was voluntary, and the only `src/` change is an additive module plus its `pub mod` line. Gate re-run locally: 6/6 pass, nothing re-blessed.

nh13:
## Correction to finding 1 — right measurement, wrong mechanism, wrong remedy

I sabotaged #1545's slice the same way while sweeping it, and the result invalidates the *cause* and the *fix* I gave above. The numbers stand; the diagnosis does not.

**What I got wrong.** I said the mechanism was "reads fail → normalizer bails → returns input unchanged", and recommended adopting `AuditedProvider` from #1540. Measured on #1545, which already has that provider wired in with a `successful_reads` floor: under an identical 100-base truncation, **the audit fired zero times** — 0 failed-read messages, 0 floor messages — while 8 of 22 tests still went red for other reasons.

Truncating a stored sequence does not produce *failed* reads. `WindowProvider` serves the shorter sequence and `get_sequence_length` reports the shorter length, so every read **succeeds with wrong content**. `AuditedProvider` audits `Err` returns by construction, so it is blind to this. Adopting it here would not have closed what I measured.

**What actually catches it**, in both siblings and not here:

```
the_committed_windows_serve_every_transcript_whole   FAIL
  assertion `left == right` failed: LRG_199t1: get_sequence_length
  disagrees with the stored sequence
    left: Some(13993)
   right: Some(100)
```

That is a **self-consistency check on the fixture**, not an audit of reads — and it is the discriminator:

| PR | asserts stored length == `get_sequence_length` |
|---|---|
| #1540 | yes |
| #1545 | yes |
| **#1544** | **no** |

So the finding is unchanged in substance and cheaper to fix than I said: this gate is the only one of the three that will not notice a corrupted slice, and the remedy is the ~10-line length-agreement assertion its two siblings already carry, not the audit wrapper. Both together would be strictly better — the audit covers reads that fail, the length check covers reads that lie — but the length check is the one that closes the case I actually demonstrated.

**Unchanged and still verified:** 22 of 36 pinned results are fixed points (`output == input`); 24 of 36 survived the truncation; and `every_worked_example_answer_is_a_fixed_point` passes on a fully starved slice. That last one is not specific to this PR — #1545's `every_answer_is_a_fixed_point` also passes under sabotage, *including* its audit call, for the same reason. An unchanged input is trivially a fixed point, so the shape is vacuously satisfiable wherever the reference is wrong-but-consistent. Worth knowing before that test is trusted as evidence in either PR.

Apologies for the misdirection on the fix.

Issue #1542, opened 2026-08-07
Normalizer: a dup-shaped split still emits members at separation 0 (delins.md:16), surviving #1537


nh13:
`DNA/delins.md:16` forbids splitting a change into members on consecutive nucleotides:

> - changes involving two or more consecutive nucleotides are described as deletion/insertion (delins) variants.

#1537 fixed that for the shapes it reached. This shape survives, on `main` (`5616cdb9`):

```
NC_000017.11:g.80110044_80110047delinsGTTGG
  ->  NC_000017.11:g.[80110044C>G;80110045dup;80110047A>G]
```

Reference bases `g.80110044_80110047` are `CTGA`. Separations, under `sep = next.start - prev.end - 1` with a `dup` occupying the duplicated position:

| pair | separation |
|---|---|
| `80110044C>G` -> `80110045dup` | **0** |
| `80110045dup` -> `80110047A>G` | 1 |

The first pair is adjacent, which is what `:16` forbids.

## The output is sequence-correct

Both forms denote `GTTGG`, so this is a representation defect, not a denotation one:

```
input   : CTGA -> GTTGG
members : 80110044 C>G -> G ; 80110045 dup (T) -> TT ; 80110046 G untouched ; 80110047 A>G -> G
          G + TT + G + G = GTTGG
```

## Why #1537 stepped around it

#1537 was measured against three sibling rows, all of which it fixed:

```
NM_000038.6:c.5265_5268delinsTGCG
  before : c.[5265G>T;5266_5268delinsGCG]     separation 0
  after  : c.[5265_5266delinsTG;5268T>G]      separation 1

NM_000083.3:c.2461_2464delinsCTCC
  before : c.[2461_2463delinsCTC;2464G>C]     separation 0
  after  : unchanged (fixed point)

NM_000392.5:c.4465_4473delinsGGCCCACAG
  before : c.[4465_4466delinsGG;4468_4470delinsCCA;4471_4473delinsCAG]   4470/4471 adjacent
  after  : c.[4465_4466delinsGG;4468_4473delinsCCACAG]                   separation 1
```

This row is unchanged across the same two builds. The distinguishing feature is the `dup` member: the adjacency here depends on reading a `dup` as occupying the position it duplicates. If the separation computation treats a `dup` as zero-width, the pair looks non-adjacent and the guard never fires. That reading question is worth settling explicitly as part of the fix, because it also determines whether the emitted form is a violation at all.

[...]

Issue #1541, opened 2026-08-07
Normalizer: c.[710T>A;713T>A] and c.710_713inv are both fixed points — one variant, two normal forms


nh13:
Two spellings of one variant reach two different normal forms on `main` (`5616cdb9`):

```
NM_000500.9:c.[710T>A;713T>A]  ->  NM_000500.9:c.[710T>A;713T>A]
NM_000500.9:c.710_713inv       ->  NM_000500.9:c.710_713inv
```

Both are fixed points, so neither converges on the other.

## They denote the same sequence

Verified from the reference bases directly, not inferred from the shape:

```
c.710  c.711  c.712  c.713
  T      C      G      T          reference
```

`revcomp(TCGT) = ACGA`. Applying `c.[710T>A;713T>A]` gives `A C G A`. Identical.

The mechanism is that the interior pair `CG` is self-complementary, so the inversion changes only the two flanking columns — which is exactly what the two substitutions write. The reference bases at `c.710` and `c.713` genuinely are `T` and `T`, so the substitution spelling is well-formed and not a reference mismatch.

## Why #1535 did not reach it

This is the 4-nucleotide analogue of the 8-nucleotide case #1535 fixed. That change types a whole-block reverse complement as one `inv` when the members it competes with are `delins`, per the definitional exclusion in `DNA/inversion.md:5`:

> Inversion: a sequence change where, compared to a reference sequence, **more than one nucleotide** replacing the original sequence is the reverse complement of the original sequence.

Here the competing members are **substitutions**, not `delins`, and `general.md:56` ranks substitution above inversion:

> - **prioritisation**: when a description is possible according to several types, the preferred description is: (1) substitution, (2) deletion, (3) inversion, (4) duplication, (5) insertion.

So there is a real argument that the split form is the correct target here and that the `inv` spelling is the one that should move. That is the #1230 case, which #1535 deliberately left untouched. **What is not defensible is that neither moves** — one variant currently has two normal forms, which is a confluence failure regardless of which form wins.

## What this issue asks for

A ruling on which form governs, then convergence on it. The two candidate readings are already on record and in tension:

- `general.md:56` ranks substitution above inversion, which argues for `c.[710T>A;713T>A]`.
- `DNA/inversion.md:5` defines the block as an inversion, and `DNA/delins.md:5` excludes an inversion from being anything else.

[...]

PR #1545, opened 2026-08-07
test(conformance): gate the harvested unguarded cases on every PR

nh13:
An audit of this project's working documents found roughly 150 concrete HGVS reproducers — an input plus an established behavioural claim — that no test in any CI configuration can fail on. This lands the verified subset as hermetic, PR-blocking guards.

Every row was measured against the prepared reference on `5616cdb9`, and independently re-measured against a pre-#1535/#1537 binary so each expectation is known to be stable across a real normalizer change rather than a one-build artifact.

## Why these cases and not others

The suite deliberately excludes cases that are already guarded. Notably it does **not** include the six spec-stated protein projections: five of them are genuinely gated today by `passing_spec_mandated_musts_stay_passing`, which is neither ignored nor manifest-gated, so adding them would be duplicate coverage presented as new coverage.

What is here falls into four groups.

**Spec-stated answers that no PR-running test asserts.** `NM_004006.2:c.5690del` -> `c.5697del`, `NM_024312.4:c.2686A[10]` -> `c.2692_2693dup`, `c.1738TA[6]` -> `c.1741_1742insTATATATA`, the `r.` counterparts, and `NM_007294.3:r.[2077g>a;2077_2078insua]` -> `r.2077delinsaua`. These are classified `requires-reference` in the enumeration fixture and `false-acceptance` in the normalization fixture, and **both** committed guards skip those statuses — so the spec's own published answers were pinned by nothing.

**Real-accession guards for #1535 and #1537.** Those PRs added twelve test functions between them and every one runs against a synthetic transcript (`NM_TEST.1`, `TEMPLATE`). No real accession became guarded. This adds the three `delins.md:16` rows on their real accessions, plus a corpus-wide structural check that no output places members on consecutive nucleotides.

**The `NM_004006.2:c.76_83` inversion class**, which matters more than its size suggests: #1535 declared `Representation-Change: 0 rows move over 5,761,302 real expressions` and correctly explained that zero as *structural* — a cis allele whose `delins` members form a whole-block reverse complement has zero instances in ClinVar, Paraphase or CMRG. The corpus cannot build the shape, so a curated case is the only instrument that can catch a regression of it.

**Stability pins**, marked as such. Seven genomic net-insertion rows that are fixed points on both binaries are pinned as *shipped and stable*, never as spec-correct, because the governing ruling is open.

## The false negative this suite is built to avoid

A window fixture captured from a normalize pass is exactly as wide as the reads that pass made. A wider-reading pass runs off the end, the provider errors, and `src/normalize/mod.rs` turns that into `Ok((canonicalize_genome_variant(variant), vec![]))` — success, empty warning vector, input returned unchanged. For a lone-member variant that is byte-identical to a correct answer. A guard built naively on such a fixture is green while the defect it guards is fully present; that has already happened in this repo, on three of five rows.

Two mitigations, both mandatory here:

`serve_transcripts_whole` in the generator, so for a transcript there is no "outside the window" left. Asserted hermetically through all three provider entry points, so a regenerated fixture that lost the widening fails loudly.

`AuditedProvider`, promoted out of a branch-local copy into `src/conformance/audited_provider.rs`. It forwards the full `ReferenceProvider` surface — 17 methods, including `infer_genome_build` and `get_protein_length`, which the branch-local copy omitted; a wrapper that drops an overridable method measures itself rather than the provider (#726). It records failed base-sourcing reads and counts served ones, and its `Clone` is hand-written so the tally is **shared** rather than copied, since a derived `Clone` would report zero for every `Normalizer::new(provider.clone())` pass.

The audit assertion is evaluated **before** any verdict. "The fixture could not answer" must always outrank "ferro answered differently", or a fixture defect gets reported as conformance.

This was demonstrated, not assumed. Narrowing one committed window from 1577 bp to 200 bp makes the suite fail on the audit assertion, naming the three failed reads, with no verdict assertion reached:

```
the hermetic slice failed 3 reference read(s) during this pass, so any verdict below is about a
normalization that ran short of reference data, not about ferro's answer.
  get_sequence(NC_000012.12, 4517518, 4518910): ... is not covered by any committed window
```

That row's expected output *is* its input, so under-serving would have scored it green.

## Two red guards, ignored and failing on purpose

`the_four_mer_inversion_pair_converges` (#1541) and `the_dup_shaped_split_does_not_touch` (#1542). Both are `#[ignore]`d, red against `main`, and name their issue in the ignore reason. Neither may be weakened to green.

#1541 is worth reading carefully, because it is not an open question. The target form `c.[710T>A;713T>A]` is *decided* — `general.md:56` ranks substitution (1) above inversion (3), the same answer #1230 got, and both directions of that ruling are already pinned in `issue_1517_inv_priority_over_delins.rs`. What blocks it is the axis gate: `is_splittable_single_member` matches only `HV::Genome`/`HV::Mt`, so a lone `c.` member never reaches the sequence-first pass. The existing #1230 control passes only because it is written on the `g.` axis. **Measured under #1484, which widens that gate, all three spellings converge on the decided form** — so this guard should be un-ignored in the change that merges #1484.

`the_inversion_preference_control_still_holds` is the green companion: the `delins`-competing row must keep answering `inv`, so the rule cannot be rewritten as "always prefer the split".

## Findings from building it

Three claims in the source audit did not survive measurement, and are recorded in the rows' notes rather than quietly corrected:

`NC_000002.11:g.47639670_47639673delinsTT` does **not** return itself — it splits, on both binaries, so the audit's "all eight net-insertion rows are fixed points" was wrong as written rather than a regression.

`NM_006420.3:c.[242C>A;247_249delinsTT]` was suspected of being a sequence-denotation bug. It is not: both forms yield `AAAGGTT`, verified from raw FASTA. It is a representation question and is pinned as a stability row, not as a defect.

`NM_000500.9:c.710_713delinsACGA` is not a fixed point on `main`; it normalizes to `c.710_713inv`. So #1541's class splits 2 against 1, not three ways.

Two arithmetic errors in the premise tests were caught by the premise tests themselves, which is the argument for reading bases from the committed fixture rather than restating them in prose.

## Noted, not fixed here

`WindowProvider`'s out-of-window error hardcodes `extract_biocommons_windows` as the regeneration command regardless of which corpus is loaded (visible in the excerpt above). Worth a small follow-up in `src/conformance/reference_window.rs`; this suite's own audit message names the right command.

## Verification

- generator `--check` against the prepared reference: `committed reference slice is up to date (34 rows, 10 transcripts, 21 windows)`
- `cargo nextest run --features dev -E 'test(case_harvest)'` — 27 passed
- `--run-ignored all` — 27 passed, 2 failed, exactly the two recorded defects
- whole `it` binary — `5466 tests run: 5466 passed (3 slow), 147 skipped`; nothing re-blessed, nothing broken
- `cargo fmt --all -- --check`, `cargo clippy --features dev --all-targets -- -D warnings` — clean
- committed slice is 217 KB, under the 500 KB pre-commit limit; `captured_from` records the manifest basename only, so no machine-local path is committed

Refs #1541, #1542, #1536, #1517.

Representation-Change: none. Test-only plus two new `src/conformance/` modules; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched, and no existing expectation was re-blessed. The suite pins current behaviour and adds two failing guards for already-filed defects.


nh13:
@coderabbitai pause

nh13:

Both lenses completed against `5a006426` (CLI: `--base origin/main`, 8 findings, no rate-limit; agent lens from a detached `review-1545/`). Reviewed on current `main` (`5616cdb9` is an ancestor). CI at time of writing: 10 pass, 0 fail, 5 pending.

**Verified independently before reviewing:** all **34** citations resolve verbatim against the vendored checkout at the pinned `spec_commit` (`6f85311`), and the gate runs 22/22 green.

**This PR gets the thing its two siblings get wrong.** I sabotaged its slice the same way I sabotaged #1544's — truncating every stored transcript sequence to 100 bases — and it went red loudly, 8 of 22, naming the corruption directly:

```
the_committed_windows_serve_every_transcript_whole   FAIL
  LRG_199t1: get_sequence_length disagrees with the stored sequence
    left: Some(13993)   right: Some(100)
```

That check is the discriminator, and #1544 is the only one of the three that lacks it. Worth stating plainly because it also **corrects a recommendation I made on #1544 an hour ago**: I attributed that gate's blindness to failed reads and pointed at `AuditedProvider`. Measured here, under identical sabotage, **the audit fired zero times** — 0 failed-read messages, 0 floor messages. Truncation yields reads that *succeed with wrong content*, which an `Err`-auditing wrapper cannot see. `successful_reads()`'s "zero failures means nothing when nothing was read" reasoning is right and worth keeping; it just guards a different failure than the one truncation produces.

One consequence for this PR: **`every_answer_is_a_fixed_point` passes on a fully starved slice**, audit call included, because an unchanged input is trivially a fixed point. Same for #1544's equivalent. Not a defect — but it should not be cited as evidence the slice is sound.

Three confirmed findings.

---

### 1. ⚠️ `serve_transcripts_whole` silently skips a transcript that failed to capture

**`examples/extract_case_harvest_windows.rs:190-206`** · *CLI [7], confirmed*

```rust
let Some(bases) = transcript.sequence.clone() else {
    continue;
};
```

The doc comment two lines above says *"Serving the accession whole removes the failure mode by construction."* For a transcript with `sequence: None` it does the opposite — the widening is skipped, the accession keeps whatever narrow genomic window was captured, and generation proceeds to write. The invariant the design rests on is quietly not established for exactly the transcript that had a problem.

The committed slice has 0 of 10 in that state, so nothing is wrong today; this is about what a regeneration against a different manifest would produce. Note the sibling contrast — **#1540's generator now refuses to write at all** on an incomplete capture ("Refusing to write a fixture built from a partial pass"), which is the standard this one should meet. Returning `Result` and `?`-ing at the call site is the minimal version.

### 2. ⚠️ A blank citation quote passes the verbatim check — and this is now the second copy

**`src/conformance/case_harvest.rs:213`** · *CLI [4]; same defect I reported on #1544*

`cited.contains(&self.quote)` is unconditionally true for `""`, so a row with an empty quote satisfies `every_row_cites_the_spec_verbatim` while asserting nothing. `Citation`'s own doc comment names the hazard: *"a bare line number resolves against any file long enough to have that line."* An empty quote **is** that bare line number.

The structural half matters more than the bug. `Citation` and its 23-line `verify` are **byte-identical** between this PR and #1544 (`diff` clean), and neither exists on `main` — so whichever lands second duplicates the other, and this hole then has to be fixed **twice or it survives in one copy**. Since this PR already promotes shared conformance infrastructure into `src/conformance/audited_provider.rs`, `Citation` belongs in the same shared home. That also decides a merge order: **land this one first**, then let #1544 consume it.

I'd reject the path-traversal half of the CLI's finding on both PRs: `spec_dir` is a vendored submodule and `clause` comes from a committed fixture, so there is no untrusted input.

### 3. 🔸 Omitting `expected` silently reclassifies a pinned row — mitigated by the census, but only by a count

**`src/conformance/case_harvest.rs:122`** · *CLI [2], confirmed with a caveat*

`expected: Option<String>` has no `#[serde(default)]`, but it does not need one: **serde treats a missing field as `None` for `Option<T>` regardless**, which is the non-obvious part. So deleting the key turns a pinned row into a confluence-only row with no complaint from deserialization — and `expected: None` is a *meaningful* value here ("the governing ruling is open"), so the silent case is indistinguishable from the deliberate one.

Measured: deleting `expected` from the flagship `NM_004006.2:c.76_83inv` row fails exactly **1 of 22** tests — `the_corpus_census_is_unchanged`. So the census is a real backstop and does its job. The caveat is that it is a *count*: it catches this because pinned-count drops and confluence-only-count rises, not because the row was named. Given `expected: None` is load-bearing semantics rather than an absence, `#[serde(deny_unknown_fields)]`-style strictness on presence (missing vs explicit `null`) would put the guarantee in the type rather than in a tally.

---

### Also noted, not blocking

- **`Fixture::cases`, `Citation::clause`, `Citation::quote` are undocumented** while their siblings carry docs. *CLI [1].* Same nit as #1544 — another consequence of the duplication above; fixing it once in a shared module fixes both.
- **`contig_lengths` vs stored length including untemplated poly-A.** *CLI [8], trivial.* Worth a glance since `the_committed_windows_serve_every_transcript_whole` is the check I just praised — it compares `get_sequence_length` against the stored sequence, so if poly-A handling ever makes those legitimately differ, that assertion is where it will surface.

### Rejected, with reasons

- **Record a manifest content digest in `captured_from`.** *CLI [6].* Same disposition as on #1544: pre-existing and repo-wide, not introduced here. Across the committed window fixtures the field is already split between `"manifest.json"` and a `prepared_at` timestamp, and #1540 chose the timestamp form the same day — so the repo now has both conventions live in new code. Standardising is a cheap follow-up, not this PR's obligation; happy to file it.
- **Assert a rendered-fixture byte ceiling.** *CLI [5], trivial.* The pre-commit limit already enforces this where it is enforceable; duplicating it as a generator assertion adds a second number to keep in sync with the hook.
- **Pin the three #1541 rows' current outputs.** *CLI [3].* This is the one place I think CodeRabbit is substantively wrong rather than merely out of scope. Those rows are `expected: None` deliberately — the PR's doc comment states that asserting *which* of two legal forms they converge on would freeze a representation nobody has adjudicated, and `#1541` is the open question. Pinning them now is precisely the representation-stability mistake the repo's own doctrine warns against. Leaving them confluence-only is the correct disposition.

---

**Spec audit: clean.** The `inversion.md:5` / `delins.md:5` pairing is used as a definitional exclusion rather than a ranking, which is right — a change that *is* the reverse complement of its reference is an inversion, and `delins` is what remains once substitution and inversion are excluded, so no prioritisation argument is needed for the DMD row. I'd flag one thing for the record rather than as a defect: `the_inversion_preference_control_still_holds` and the `NM_000500.9:c.710_713` class are deliberately held in tension so the #1541 rule cannot be written as "always prefer the split" *or* "always prefer inv". That is the most valuable assertion in the corpus and the easiest to accidentally weaken — worth a comment saying so at the call site.

nh13:
## Addendum — I reviewed a partial finding set; here are the remaining five

My sweep above was written against **8** CLI findings. The lens actually returned **13** (`"type":"complete" … "findings":13`) — I read the output file while the review was still streaming and treated a mid-flight tail as the full set. Same mistake I made on #1544 an hour earlier and said I had caught. Nothing above changes; five findings were simply missing. One of them is real and worth acting on.

### 5. ⚠️ `member_bounds` silently mis-parses an intronic offset as a plain coordinate

**`tests/it/case_harvest.rs:210-228`** · *CLI [12], confirmed* · feeds a **gating** test

`leading_number` takes leading digits and stops, so `123+5A>G` yields `123`. Reproducing the helper's logic exactly:

```
UTR member (covered by the existing test)   NM_024312.4:c.[-6_-3G[6];12A>T]    -> None
INTRONIC offset (the hole)                  NM_004006.2:c.[123+5A>G;130A>T]    -> [(123, 130)]
INTRONIC on both members                    NM_004006.2:c.[123+5A>G;124-2A>T]  -> [(123, 124)]
```

The third line is the one that matters. `123+5` and `124-2` are intronic positions that can be thousands of bases apart, and the helper reports them as spans `123` and `124` — a gap of 1, which `adjacent_members` then classifies as **touching**. So `no_emitted_output_puts_two_members_on_consecutive_nucleotides`, which is live and gating (not `#[ignore]`d), can report adjacency that does not exist.

`a_non_plain_coordinate_declines_rather_than_mis_parsing` looks like it covers this and does not: its only case is the UTR form `-6_-3`, which declines because `-` is not a digit. A leading digit followed by `+`/`-` walks straight through. The test name claims the general property; the assertion buys one instance of it.

**Latent today, not firing:** 0 of 34 inputs and 0 of 34 `expected` values carry an intronic offset, and 10 rows are multi-member. So nothing is wrong in the committed corpus — but the helper runs against *emitted outputs*, and a `c.` normalization is free to produce intronic coordinates, so this is one corpus row away from a false positive in a gating check. Reusing the guarded coordinate parser CodeRabbit points at, or simply declining any member matching `\d+[+-]\d+`, closes it.

For the record, the trans-allele half of that finding is already safe: `c.[76C>G];[80A>T]` returns `None`, because after `strip_suffix(']')` the second fragment starts with `[` and `leading_number` yields nothing. It declines for an incidental reason rather than a stated one, so an explicit reject would document intent — but there is no bug to fix.

### The other four — all minor, no objection either way

- **[9]** The `NC_000017.11:g.80110044_80110047delinsGTTGG` row's note should say its pin is a change detector for the known #1542 representation, not an adjudicated answer. Fair, and consistent with how the `NM_006420.3` row already labels itself ("STABILITY PIN PLUS A CORRECTED DIAGNOSIS … must not be read as endorsement"). Cheap consistency win.
- **[10]** Derive the pinned-row count from the census constant rather than recomputing it from `cases.len()` and `unpinned_inputs()`. Style; either is defensible. The recomputation has the mild advantage of not needing an update when a row is added.
- **[11]** `assert_the_slice_answered_everything(&provider, outcomes.len())` at `:381` uses a floor that undercounts, since the loop normalizes every member of every confluence class, not one read per outcome. A low floor is a weaker guarantee, not a wrong one — and given what I measured about `successful_reads` in the main comment, this floor is not load-bearing for the truncation case anyway.
- **[13]** Document the committed-fixture exemption near the module header. `.coderabbit.yaml`'s `tests/**/*.rs` instruction names three deliberate exceptions; this PR makes another, for the same reason as `tests/fixtures/validation/` (real GRCh38 bases cannot be generated programmatically). Same note I made on #1540: the exception list should be updated deliberately rather than left to drift.

I'd still land this before #1544 for the `Citation` sharing reason in the main comment.

PR #1546, opened 2026-08-07
test(corpus): add a separated reverse-complement family and three real oracles

nh13:
The synthetic corpus answers "did this output move?" and never "is this output right?". That is the correct instrument for drift and useless for correctness, and it is not a fixable property of thirteen of the fifteen families: they are sets of descriptions with no record of what the description was meant to denote.

This adds a sixteenth family built the other way round — the alternate is chosen first and the reference derived around it — so `(reference, alternate)` is exact for every one of its 270 rows, and three real oracles come essentially for free. **This is the branch that found #1536**; until it lands, that measurement has no home.

## The family: `separated_revcomp_runs`

`delins_hiding_an_inversion` was the only family that built a reverse-complement block, and it emits its two pieces at separation 0. Where such a block coincides with its own reference is a property of the *sequence*, not of the description, so a family crossed with the random 20-mers gets whatever coincidence pattern the xorshift produced — and none of them separates two multi-column changed runs. A gate that engages only at separation >= 2 with multi-column pieces therefore fires on no row, and a change to one measures a blast radius of **zero for structural reasons rather than safe ones**. This is the same class of corpus blindness as #1456 / #1460 / #1478.

Fifteen designs: separations 0, 1, 2, 4 and 8 between 2-column changed runs, each planted at three offsets so the axis geometry (5'UTR, CDS end, both exon junctions) varies under an identical block. Three spellings per design — the whole block as `inv`, the same as `delins`, and the runs written individually — which denote one variant by construction. 15 designs x 3 spellings x 6 axis/direction combinations = **270 rows; the corpus goes from 78,028 to 78,298.**

The separation-4 design is `AATGCACA` -> `TGTGCATT`, the block #1517 is about: a true reverse complement coinciding at its four interior columns, leaving two 2-base changes.

Two facts are pinned rather than argued. The run layout is derived from the parameters *and* re-derived from the sequence, and the two must agree, so a typo in one of the five literal blocks fails instead of silently contributing a design at a separation nobody asked for. And an odd separation needs three runs, not two: a column coincides with its own reverse complement exactly when its mirror does, so the changed/unchanged pattern of any whole-block inversion is a palindrome and a two-run block's gap always has even length.

Also asserted: no other family, including `long_block_inversion`, builds a reverse-complement block whose multi-column runs are separated at all — the structural gap this family was added for, stated as a test so it fails if some other family grows the shape.

## The three oracles

**1. Sequence preservation — PASSES, 270/270.** Apply the normalized output back to the reference via `spdi::apply_to_reference` and require the result to equal the alternate the generator intended. This is the class that produced #1416, #1453 and #1431, and that the three env-var oracles are all blind to: a wrong output can be a perfectly stable fixed point and it parses fine. 270 of 270 rows applied, none declined, none denoting anything other than the intended alternate.

**2. Confluence — RED, `#[ignore]`d, records #1536.** All three spellings of one design must normalize to one string, and that string must be a fixed point. Both halves are asserted, because neither implies the other and only the pair makes it a confluence property rather than a table of expected strings.

**3. Separation-rule conformance — RED, `#[ignore]`d, records #1517.** `general.md:34`/`:35` and `DNA/delins.md:16-18` evaluated against the sequence rather than a pinned string. Two correctness points carried deliberately: an insertion consumes no reference position, so members are closed intervals with an insertion left *empty* and one formula `sep = next.lo - prev.hi - 1` covers every pair; and a delins of unequal length has many minimal alignments, so an unchanged interior is claimed only when forward and backward Levenshtein DP agree that no minimal-cost path can consume the column — which can under-report and must never over-report.

**Both red oracles are ignored-red on purpose and must not be weakened to green.** Neither was narrowed to pass; narrowing either to its settled half leaves it red. A version of oracle 2 that checked only the fixed-point half would be green and would be measuring nothing this family was added for. Un-ignore each *with* the corresponding fix, in the same change.

## The counts in the `#[ignore]` strings are the first-run numbers and #1537 has since moved them

Read this before reviewing: the doc comments and `#[ignore]` reasons state what was measured **on the first run of each oracle**, before #1537 (`fix(normalize): never split a delins into members on consecutive nucleotides`, closing #1524) landed on `main`. Re-measured after rebasing onto `5616cdb9`, the *reasons* are unchanged and the named smallest reproducers still reproduce byte-identically; the aggregates have moved. I have deliberately left the committed strings alone — "on first run" is a historical framing and re-blessing it is the author's call, not a rebase chore — so here are both readings side by side.

Oracle 2, confluence — `38 of 90` in the string, **32 of 90** measured now, `0 outputs fail the fixed-point half` in both. So the failure is still entirely disagreement between spellings, never drift within one. By axis:

| axis | non-confluent classes of 30, first run | after #1537 |
|---|---:|---:|
| `g.` | 0 | **0** |
| `c.` (single exon) | 14 | **8** |
| `cx.` (three exons, 5'UTR) | 24 | **24** |

That is the most useful thing in this PR after the family itself: **#1537 converged 6 of the `c.` classes and left the `cx.` 24 completely untouched.** The boundary-crossing residue is exactly #1536, and it is now cleanly separated from the adjacent-member defect #1537 fixed. The smallest reproducer is unchanged — a four-base block whose three spellings are three distinct fixed points on `cx.` and converge on `c.`:

```text
NM_TESTX.1:c.-3_1inv                       ->  NM_TESTX.1:c.-3_1inv
NM_TESTX.1:c.-3_1delinsTGGT                ->  NM_TESTX.1:c.-3_1delinsTGGT
NM_TESTX.1:c.[-3_-2delinsTG;-1_1delinsGT]  ->  unchanged
```

Oracle 3, separation — `2 settled + 174 contested` in the string, **2 settled + 216 contested** now:

```
180 of 270 normalized outputs violate the separation rules (90 clean, 0 unevaluated);
2 findings name a `delins` member and are settled,
216 name an exact inversion and are #1517's open question
```

The settled pair is unchanged (two adjacent members, neither an inversion on its own, that the single-exon coding axis merges and the multi-exon one does not). The contested count grew because #1537 merges adjacent members, so more outputs are now whole-block exact inversions spanning an unchanged interior — which is #1517's open decision read in the opposite direction, not a defect and not a regression.

If the reviewer would rather the strings carry the current numbers, say so and I will re-bless both in place.

## Verification

Rebased onto `5616cdb9` — clean, no conflicts — and re-gated after the rebase.

- `cargo nextest run --features dev -E 'binary(it)'` — `Summary [38.150s] 5444 tests run: 5444 passed, 145 skipped`
- `cargo nextest run --features dev --run-ignored all -E 'binary(dump_normalized_corpus)'` — `Summary [1.144s] 35 tests run: 33 passed, 2 failed`, the two failures being exactly the two ignored-red oracles above

Note for reviewers rebasing this locally: `tests/fixtures/grammar/hgvs_spec_normalization.json` is gitignored and only regenerates when *missing*, so a stale copy survives a rebase and fails `ruling_records_are_intact` against ruling statuses `main` has since decided. Re-run `cargo run --features dev --bin generate_spec_fixture`.

Refs #1536, #1517, #1524, #1537.

Representation-Change: none. The entire diff is one file, `examples/dump_normalized_corpus.rs` — no `src/` file is touched, so there is no code path by which a normalized output could move, and 5,444 `it` tests pass with nothing re-blessed. The one thing worth stating explicitly rather than hiding behind `none`: this change **does** grow the corpus, 78,028 -> 78,298 rows, which changes the denominator of every future `compare` run against it. That is a change to what future measurements are taken over, not a change to any shipped representation — the 78,028 pre-existing rows normalize exactly as they did before, and the 270 new rows have never been normalized by a released version, so no stored downstream string can be keyed on them. Relatedly, no `0 moved` figure is quoted anywhere in this PR, because this branch is the *repair* of a structural zero rather than a consumer of one: the family exists precisely because the corpus could not previously build a reverse-complement block with separated multi-column runs, so every gate keyed on that shape measured 0 of 78,028 for structural reasons.

### DIFF
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -72,6 +72,17 @@
 //!    scale is expensive in a way adding a family is not — check the dump cost
 //!    before crossing a kilobase core with anything.
 //!
+//! 5. **…and only at the sequence structure it builds them from** (`#1517`). The
+//!    families are drawn *on* the random cores, so any property of the reference
+//!    itself is whatever the xorshift produced. Where a reverse-complement block
+//!    coincides with its own reference is exactly such a property, and the one
+//!    family that built such a block —`delins_hiding_an_inversion` — emits its
+//!    pieces adjacent, so nothing in the corpus separated two multi-column runs of
+//!    an inversion and a gate keyed on that measured `0 of 78,028`.
+//!    `separated_revcomp_runs` is the answer, and like `long_block_inversion` it
+//!    brings its own designed references because a random core cannot be asked to
+//!    have a coincidence pattern.
+//!
 //! ## The corpus is deliberately independent of `tests/it/common/`
 //!
 //! An example cannot reach test helpers, so the reference construction below is its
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:

Both lenses completed against `3a77e31d`. The CLI lens needed a retry — the first run died with `errorType: connection` and **zero** findings, which is a *missing* lens rather than a clean one; the retry finished cleanly (`"findings":3`, 3 lines present, no truncation). Reviewed on current `main` (`5616cdb9` is an ancestor). CI green: 37 rows, 36 pass, 0 fail.

**Every number in the PR body reproduces exactly.** I ran the harness and both ignored oracles rather than taking the body's word:

```
corpus                 78,298 rows written   (270 carry separated_revcomp_runs)
per axis/direction      45 rows x {g,c,cx} x {3prime,5prime} = 270
oracle 1 (sequence)    PASS
oracle 2 (confluence)  32 designs do not converge and 0 outputs are not fixed points
                       non-confluent by axis:  g. 0   c. 8   cx. 24
oracle 3 (separation)  180 of 270 violate (90 clean, 0 unevaluated);
                       2 findings name a `delins` member and are settled, 216 name an exact inversion
non-ignored suite      33 passed, 0 failed;  --run-ignored all -> exactly the 2 documented reds
```

That matches `32 of 90`, the `0 / 8 / 24` table, and `2 settled + 216 contested` digit for digit. The re-measured-after-#1537 framing is accurate, and the `#[ignore]` strings label their counts "red on first run", so the staleness the body discloses is already visible at the point of use.

Also checked, since both are load-bearing here and neither is mentioned in the body: the diff is **+1259/−0**, purely additive, so no `fn *_strategy()` is touched and none of the pinned seeds in `tests/proptest-regressions/` is invalidated. And the structural guard is **two-sided** — it asserts no *other* family builds a separated reverse-complement block, *and* that the new family's designs actually reach every one of `REVCOMP_SEPARATIONS`. That second half is the "can the generator vary the property?" check `CLAUDE.md` says has to be done by hand and keeps getting missed; encoding it as a test is the right answer and is the strongest thing in this PR after the family itself.

Two findings, both in doc comments, both worth fixing because this file's comments are load-bearing structural claims.

---

### 1. ⚠️ `REVCOMP_SPAN_OFFSETS`' "every offset crosses a junction" claim is false for the narrow blocks

**`examples/dump_normalized_corpus.rs:574-581`** · *CLI [2], confirmed*

> *"…and every offset crosses at least one of the two `EXON_SPANS` junctions."*

`EXON_SPANS` is `[(0,7),(7,14),(14,20)]`, so the interior junctions sit at core offsets **7** and **14**. A block of width *w* planted at offset *o* crosses one only if some junction *j* satisfies `o < j <= o+w-1`:

| separation | block width | offset 0 | offset 4 | offset 8 |
|---|---|---|---|---|
| 0 | 4 | **no** | yes | **no** |
| 2 | 6 | **no** | yes | **no** |
| 4 | 8 | yes | yes | yes |
| 8 | 12 | yes | yes | yes |

So only offset 4 crosses for the narrow blocks; at offsets 0 and 8 the separation-0 block occupies `0..3` and `8..11`, both entirely inside one exon. The claim holds for the wide blocks only.

(I've left the odd separations out of the table: those are three-run blocks per the PR's own palindrome argument, so I didn't pin their widths exactly. The even rows are exact and the finding stands on them alone.)

This matters more than a typo because the comment is the *justification for the offset set* — it is the reason a reader believes the axis geometry is covered. Given the file already encodes its coverage claims as tests, the better fix is an assertion over `REVCOMP_SPAN_OFFSETS` × the designs, so a future `EXON_SPANS` edit fails loudly instead of silently falsifying the prose. That is the same failure mode as #1478, which this PR is partly a response to.

### 2. 🔸 The family count is stale in two places — but not in the direction CodeRabbit suggests

**`:89` and `:2398`** · *CLI [3], half-confirmed*

Both read "**thirteen of the fifteen** families cannot be asked". With `FAMILIES` at 14 entries plus `LONG_FAMILY` and the new `REVCOMP_FAMILY` declared separately, the total goes **15 → 16**, so the denominator is now wrong in both places.

CodeRabbit proposes "**Fifteen** of the sixteen", and that is wrong — it would assert that the new family *cannot* be asked whether its output is right, inverting this PR's entire thesis. The numerator does not move: the new family is exactly the one that *can* be asked. The correct edit is **"thirteen of the sixteen"**, denominator only.

Flagging the suggested patch specifically because it is the kind that reads plausibly and would quietly negate the change if applied as given.

---

### Rejected, with reason

- **Add a test that `REVCOMP_FAMILY` contributes rows for every axis.** *CLI [1], trivial.* Already covered in effect: I confirmed from the dump that the family emits **45 rows in each of the six axis/direction combinations** (270 total), and the existing two-sided structural guard asserts the designs reach every separation. A third test asserting per-axis presence would restate what `the_corpus_key_is_unique` plus that guard already pin, and would need updating whenever an axis is added.

---

**Spec audit: clean.** Oracle 3 evaluates `general.md:34`/`:35` and `DNA/delins.md:16-18` against the sequence rather than a pinned string, which is the right instrument, and the two correctness points carried in its doc comment both check out: an insertion consuming no reference position makes `sep = next.lo - prev.hi - 1` uniform across every member pair, and claiming an unchanged interior only where forward and backward Levenshtein DP agree gives a conservative predicate — it can under-report and cannot over-report, which is the safe direction for a rule that accuses an output of violating the spec.

Nothing here blocks merge; both findings are comment corrections. The `#1536` measurement this branch carries is the reason to land it — as the body says, until it does, that measurement has no home.

Issue #1097, opened 2026-07-21
fix(normalize): RefSeqMismatch skips asserted reference bases of multi-base range substitutions (pos_pos REF>ALT)


msto:
> [!NOTE]
> Authorship: the content below was drafted by GitHub Copilot CLI (an AI coding
> agent) and filed via gh under @msto's account. It reflects the agent's
> analysis, not a statement authored by @msto.

## What's wrong

With reference-base validation explicitly turned on
(`ErrorConfig...with_override(ErrorType.RefSeqMismatch, ErrorOverride.Reject)`),
ferro **silently accepts a wrong asserted reference base** in the multi-base
"range substitution" form `g.pos_pos REF>ALT`. The asserted `REF` bases are
discarded during normalization and never compared against the reference.

The same wrong assertion **is** rejected in single-base and per-base allele form,
so the validation the caller asked for is applied inconsistently depending only on
the input's syntax:

```
Reference "T": 1-based positions 22..24 are 'CGC'.

g.22A>C            → REJECTED  "Reference mismatch at 22-22: expected A, found C"   ✅
g.[22A>C;24A>G]    → REJECTED  "Reference mismatch at 22-22: expected A, found C"   ✅
g.22_24AAA>CAG     → ACCEPTED  normalizes to g.23_24delinsAG                        ❌ (real bases CGC, not AAA)
g.22_23AA>CT       → ACCEPTED  normalizes to g.23G>T                                ❌ (real bases CG, not AA)
```

This is a silent wrong-output bug, and it defeats the exact guarantee
`RefSeqMismatch → Reject` exists to provide. A caller that opts into reference
validation to catch bad input (e.g. an off-by-one coordinate in a curated variant
sheet) still gets **no error** — the mismatched input normalizes to a different,
valid-looking variant. The failure is invisible: no exception, no warning, and the
result is well-formed, so it flows downstream as if it were correct. Callers who
use the `NNN>NNN` form to express codon-level changes are the ones most exposed,
because that form is exactly where the check silently does nothing.

## Severity: high

- **Silently wrong** — the caller enabled `RefSeqMismatch → Reject`, and ferro
  returns success on input that violates it. No error, no warning.
- **Inconsistent with equivalent input** — the identical mismatch is correctly
  rejected as `g.22A>C`, so behavior depends on syntactic form rather than on the
  reference, which is surprising and hard to detect.
- **Defeats an opt-in safety check** — the whole point of the override is to fail
[...]

Issue #1092, opened 2026-07-21
parser discards the stated reference bases of a substitution, bypassing reference validation


nh13:
Follow-up from PR #1091, where the fix for multi-nucleotide substitutions (`c.79GC>TT`) had to be **source-keyed** — re-inspecting the original input string — because the parser discards the stated reference bases before they reach the AST.

## The parser drops a required field

`parse_multibase_substitution` (`src/hgvs/parser/edit.rs:188-215`) folds `GC>TT` into `NaEdit::Delins { deleted: None, deleted_length: None, inserted: "TT" }`. The stated reference run `GC` is discarded at parse time.

For substitutions that run is **not optional**. `docs/syntax.yaml:209` (`dna.sub`, "Simple sequence substitution"):

> `syntax: sequence_identifier ":" coordinate_type "." position reference_sequence ">" alternate_sequence`

There is no `dna.sub` production that omits `reference_sequence`, so the bases are a required element of the input form, and ferro is dropping a mandatory field.

## The stated bases are semantically live, not decoration

The spec discourages writing the reference run for del/dup/delins, but its stated reason proves the run carries a truth value. `DNA/deletion.md:30-31`:

> **NOTE**: the recommendation is not to describe the variant as `NC_000023.11:g.33344591delA` … This description is longer, it contains **redundant information**, and chances to make an error increases (e.g., `NC_000023.11:g.33344591delG`).

The same NOTE appears at `DNA/deletion.md:45-46`, `DNA/duplication.md:35-36` and `:50-51`, `DNA/delins.md:29-30` and `:34-35`. **Every one of those "chances to make an error" examples is the same variant with the wrong reference bases written out.** You cannot make an error in a token that carries no truth value. "Redundant" here means *derivable from the reference* — which is exactly what makes it *checkable* against the reference, not safe to discard.

Anchored by `background/refseq.md:44` (repeated at `general.md:17` and `:22`): "the reference sequence used **must contain** the variant residue that is described as being changed", and `checklist.md:19`: "Does your reference sequence contain the residue that you describe as changed?"

## ferro is already internally inconsistent about this

`src/normalize/validate.rs:87-104` **already validates a stated delins reference run against the actual reference** (#486), with the comment:

> `// Non-recommended delACinsGT form: ferro parses the stated deleted bases into deleted. Validate them against the reference span (#486)`

[...]

Issue #1068, opened 2026-07-18
Substitution reference-base validation (#1052) is not applied on the m./mitochondrial axis


nh13:
Follow-up from #1052 (PR #1060).

#1052 makes a real substitution route through `normalize_na_edit`, which validates the stated reference base against the loaded reference — surfacing `RefSeqMismatch` in lenient mode and rejecting (`ReferenceMismatch`) in strict mode. This now works on the `g.`, `c.`, `n.`, and `r.` axes.

It does **not** fire on the `m.`/`o.` (mitochondrial/circular) axis. A certain wrong-reference `m.` substitution — e.g. `NC_012920.1:m.8A>C` where `m.8` is `T` — passes through silently in lenient mode (no `RefSeqMismatch`) and is accepted in strict mode (no `ReferenceMismatch`).

Root cause: `normalize_mt` routes substitutions through the window-based remap pipeline / `mt_fallback`, not through `normalize_na_edit`, so the #1052 reference-base check is never reached for `m.` substitutions.

This is **not a regression** — before #1052 no axis validated substitution reference bases — but it is an axis-consistency gap: a genomic-position substitution on the mitochondrial accession should be validated the same way a `g.` substitution is.

PR #1060 already carries the uncertain-substitution carve-out for `m.` (`uncertain_wrong_ref_mito_sub_stays_silent`); what is missing is the certain-substitution validation path and its tests (mismatched-ref warns in lenient / rejects in strict; matching-ref is silent).

Acceptance criteria:
- A certain wrong-reference `m.` substitution surfaces `RefSeqMismatch` in lenient mode and `ReferenceMismatch` in strict mode.
- A certain matching-reference `m.` substitution normalizes with no warnings.
- The uncertain/predicted-wrapped `m.` substitution stays a silent pass-through (unchanged from #1060).

nh13:
[...]

PR #1547, opened 2026-08-07
fix(normalize): validate a cis-allele member's stated reference bases before a merge can consume them

### LINKED ISSUE #1543: Normalizer: reference-base validation is bypassed for cis-allele members close enough to be merged
Strict mode rejects a substitution that misstates its reference base. It does **not** reject the same substitution when it sits in a cis allele whose members are close enough to be merged.

Measured on `main` (`5616cdb9`), `--error-mode strict`. Reference bases are `c.1 = A`, `c.3 = G`, `c.500 = G`, so `1G>T` and `3T>A` both misstate the reference and `500G>C` states it correctly:

```
NM_000143.3:c.1G>T            ->  normalize_error  Reference mismatch at 64-64: expected G, found A
NM_000143.3:c.[1G>T;500G>C]   ->  normalize_error  Reference mismatch at 64-64: expected G, found A
NM_000143.3:c.[1G>T;3T>A]     ->  NM_000143.3:c.1_3delinsTTA          status=ok
NM_000143.3:c.500G>C          ->  NM_000143.3:c.500G>C                status=ok
```

The discriminator is **merge distance and nothing else**. The identical false `G` at `c.1` is caught when its sibling is 499 nt away and missed when the sibling is 2 nt away.

## Why this is worse than an ordinary validation gap

The accepted output is built from the **real** bases, not the asserted ones. So ferro silently rewrites a description whose reference assertions are false into one that is well-formed and correct-looking, with `status=ok` and no warning, in the mode whose whole purpose is to reject such input. A caller cannot distinguish "the submitter's reference bases were right" from "the submitter's reference bases were wrong and we quietly substituted ours".

## Mechanism

`src/normalize/merge.rs:2325` refuses canonicalization when a member misstates its reference, and defers to the per-member pipeline to reject it:

```rust
    // A member that misstates its reference base is a reference mismatch, not a
    // canonicalization problem: strict mode must still reject it and lenient
    // mode must still warn. Both live in the per-member pipeline, so refuse and
    // let it run (#1052, #1097).
    if !stated_reference_bases_match(variants, kind, &ref_bytes, w_lo) {
        return None;
    }
```

But `stated_reference_bases_match` (`src/normalize/merge.rs:3050`) reads the asserted bases out of each `NaEdit`, and its own doc comment records that they are usually already gone by then:

> `canonicalize_edit` strips most of them before this pass runs, so several arms are inert
> today — but a validator that silently ignores half the channels it claims to cover is one
> pipeline reordering away from letting a reference mismatch be canonicalized out from under
> strict mode.

That is this defect, described in the comment before it happened. On the merged path the check is vacuously true, so the `return None` never fires and canonicalization proceeds.

## Suggested direction

Validate the stated reference bases **before** `canonicalize_edit` strips them, rather than after. A per-shape fix at the merge site would be the sixth in this family and would leave the same ordering hazard for the seventh.

This is the family, all previously fixed one shape at a time:

- #1052 — substitution reference-base validation
- #1068 — not applied on the `m.` axis
- #1092 — parser discarded the stated reference bases of a substitution
- #1097 — skipped for multi-base range substitutions (`pos_pos REF>ALT`)
- #1352 — `NaEdit::Identity` omitted while the doc comment claimed the list was complete

## Minimal reproducer

```
NM_000143.3:c.[1G>T;500G>C]   ->  rejected   (correct)
NM_000143.3:c.[1G>T;3T>A]     ->  accepted   (wrong; same false base at c.1)
```

## Note on expected representation impact

A fix should move no correctly-spelled input, since it only adds rejections for inputs that misstate the reference. That is worth measuring rather than assuming — in particular for lenient mode, where the change is from silent acceptance to a warning, and for any corpus row that carries reference bases which disagree with the prepared reference for an unrelated reason (a transcript version skew, for instance).

nh13:
Closes #1543

Strict mode rejected a substitution that misstates its reference base, and accepted the same substitution when a sibling sat close enough for the two to merge. The discriminator was merge distance and nothing else.

## Before / after

Real prepared reference, `--error-mode strict`. Reference bases are `c.1 = A`, `c.3 = G`, `c.500 = G`.

| input | before | after |
|---|---|---|
| `NM_000143.3:c.1G>T` | `normalize_error` Reference mismatch at 64-64 | unchanged |
| `NM_000143.3:c.[1G>T;500G>C]` | `normalize_error` Reference mismatch at 64-64 | unchanged |
| `NM_000143.3:c.[1G>T;3T>A]` | `NM_000143.3:c.1_3delinsTTA`, **status=ok** | `normalize_error` Reference mismatch |
| `NM_000143.3:c.500G>C` | `NM_000143.3:c.500G>C`, status=ok | unchanged |

And the same defect turned up unprompted in ClinVar, on a shape the issue does not describe — a `Repeat` member in the 3'UTR whose declared tract does not divide its span:

```
NM_000287.4:c.*438TAAA[1]              -> normalize_error   (before AND after)
NM_000287.4:c.[*438TAAA[1];2578C>T]    -> ok  ->  normalize_error
```

The identical member, rejected alone and accepted beside a sibling. After the change both give the same verdict, which is the invariant.

## Mechanism — and it is not the one the issue names

The issue attributes this to `stated_reference_bases_match` (`src/normalize/merge.rs:3050`) being vacuous at the sequence-first canonicalization guard. That guard is real, and its doc comment does predict this class of failure, but **it is not what let these inputs through**. Verified by instrumented reproduction rather than by reading: on `5616cdb9` the merged input produces *no* `RefSeqMismatch` warning at all — only `MembersCoalesced` — so nothing was ever generated for the guard to be asked about.

`Normalizer::normalize_allele` runs `merge::collapse_overlapping_cis_edits` and `merge::merge_consecutive_edits` over the **raw** members before any member reaches `normalize_na_edit`, which is the only place `RefSeqMismatch` (W5001) is raised, and which the strict ladder reads to reject. A merge keeps each member's *alt* bases and nothing else — `merge::Anchor` has no channel for an asserted reference — so a member consumed by a merge had its assertions discarded before anything looked at them. The merged anchor was then rebuilt from the real bases, which is why the output looked correct. `canonicalize_from_sequence` runs *after* `normalize_core`, by which time the member no longer states anything, so the guard there was downstream of the loss rather than the cause of it.

Measuring it also showed the defect is broader than the report:

- **Not only the codon-frame merge.** Plain strict adjacency (`delins.md:16`) loses the assertions the same way, so `c.[1G>T;2A>C]` was accepted too.
- **Not only substitutions.** Every channel that can carry an asserted reference was affected: `delAA`, `delAAinsGG`, `dupAA`, `invGGGGG`, and `Repeat` (the ClinVar row above).
- **Not only `c.`** — `g.`, `m.`, `n.` and `r.` all merge and all lost them.

## The fix — general, not per-shape

This is the sixth defect in one family (#1052 substitutions, #1068 the `m.` axis, #1092 the parser discarding the bases, #1097 multi-base range substitutions, #1352 `NaEdit::Identity` omitted while the doc comment claimed the list was complete), and each of the five was fixed one shape at a time.

So the question is asked **on the authored input**, once, in `normalize_core`, before any pass has had the chance to strip anything (`merge::authored_member_reference_mismatches`). **This fixes the ordering generally, not this shape:** no rearrangement of the merge pipeline can make it vacuous, because it never looks at the pipeline's output. Two further properties keep it from drifting:

- The verdict comes from `normalize::validate::validate_reference`, the **same** function the per-member pipeline uses, so the channel list cannot diverge between the two sites the way #1352's did.
- The `r.` axis is folded to DNA with `rna_uracil_to_thymine` first, exactly as `normalize_rna` does before it validates (#736). Without that, a truthful `r.2u>a` over a `T` would be rejected — over-rejecting correct input, which is worse than the hole being closed. Pinned by `the_rna_alphabet_is_not_read_as_a_mismatch`.

Strict mode rejects and lenient mode warns and continues, both by reusing the existing `RefSeqMismatch` / W5001 path rather than inventing a code. Warnings are deduplicated by sequence-axis position, so a member a merge did *not* consume — which reaches the per-member validator on its own — is still reported exactly once.

**The `return None` at `merge.rs:2325` is kept.** It is not redundant: for members whose stated bases survive per-member normalization (a substitution keeps its reference base through `canonicalize_edit`), it still refuses to canonicalize an allele that misstates its reference, which is the contract #1052/#1097 recorded there. This change makes it reachable more often rather than less.

## Blast radius

Representation-Change: none. 0 of 950 real cis-allele rows move their normalized string; 1 of 950 changes strict verdict from accept to reject, which is the defect being fixed.

Measured, not assumed, over the four committed LFS corpora — ClinVar 500k, ClinVar unique, CMRG exhaustive, Paraphase exhaustive — normalized through both revisions against the real prepared reference:

| | |
|---|---|
| rows in the four corpora | 9,949,738 |
| of which cis-allele shaped — the only rows this change can reach | 1,178 (950 distinct) |
| **denominator: rows carrying a stated reference base that disagrees with the reference** | **3 of 950** |
| of those, already reported before this change | 2 |
| of those, laundered by a merge and now reported | **1** |
| lenient normalized strings that move | **0 of 950** |
| lenient status changes | **0 of 950** |
| strict status changes | **1 of 950** (`NM_000287.4:c.[*438TAAA[1];2578C>T]`, `ok` -> `normalize_error`) |

The one strict flip is not a representation change for a stored string: the input was never a valid description of that reference, and the lone form of the same member was already rejected. Nothing that was accepted-and-correct moves.

### The synthetic harness reports a structural zero, and it is reported as one

`examples/dump_normalized_corpus.rs` reports **0 of 78,028 rows moved** across all 15 shape families. **That zero proves nothing about this change and is not offered as evidence.** The corpus builds every substitution as `format!("{}{}>{}", p(i), base(i), other(i))` where `base(i)` reads the byte straight out of the reference, and every `del`/`dup`/`inv`/`delins` in the short form. So of its 27,656 rows that carry a stated reference base, **0 misstate one** — the corpus cannot express the input property this change keys on. It is blind here in the same way #1456/#1460/#1478 record.

The real evidence is the corpus table above (which has a non-zero denominator) and the suite: **9,609 tests pass with nothing re-blessed**, which for a change capable of moving output is the strong signal — every existing expectation already agreed.

## Verification

| check | result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --features dev --all-targets -- -D warnings` | clean |
| `cargo nextest run --features dev -E 'test(issue_1543)'` | 10 passed |
| `cargo nextest run --features dev -E 'binary(it)'` | 5,454 passed, 0 failed |
| `cargo nextest run --features dev` (all binaries) | 9,609 passed, 0 failed |
| oracles (`FERRO_ASSERT_IDEMPOTENT` / `REPARSE` / `IN_BOUNDS`) | 5,454 passed, 0 failed |
| `generate_spec_fixture` + `generate_spec_enumeration`, then re-run | no drift; `DIVERGENCE_BUDGET` and `PASSING_CENSUS` unchanged |

**Nothing was re-blessed.** Six of the ten new tests fail on the parent commit; the other four are controls that already passed, so the fix cannot be credited to breaking them.

## Tests

`tests/it/issue_1543_allele_member_reference_validation.rs`, hermetic on a synthetic `MockProvider` transcript whose first three bases reproduce the issue's (`c.1 = A`, `c.3 = G`) — so `c.[1G>T;3T>A]` reaches the same `c.1_3delinsTTA` the report shows. No `FERRO_MANIFEST`, which would be skipped on every PR and gate nothing.

The distance sweep is the invariant worth keeping: the same false `c.1G>T` is rejected with siblings at six distances spanning the merge threshold, each paired with a correctly-spelled control that must be accepted. The controls' outputs are additionally asserted to straddle the threshold, so a sweep that drifted entirely to one side of it fails rather than quietly asserting distance-independence without testing it.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -3120,6 +3120,153 @@ fn stated_reference_bases_match(
     true
 }
 
+/// Every reference mismatch the **authored** members of a cis allele carry.
+///
+/// # Why this exists at all
+///
+/// The per-member pipeline validates a member's stated reference bases in
+/// `normalize_na_edit`, and that is where `RefSeqMismatch` (W5001) — and so
+/// strict mode's rejection and lenient mode's warning — comes from. But
+/// `normalize_allele` runs [`collapse_overlapping_cis_edits`] and
+/// [`merge_consecutive_edits`] over the raw members **before** any member
+/// reaches that validator, and a merge keeps only each member's *alt* bases:
+/// [`Anchor`] has no channel for the reference the submitter asserted. So a
+/// member consumed by a merge has its assertions dropped, unexamined, and the
+/// merged anchor is then rebuilt from the **real** bases — turning a
+/// description whose reference claims are false into a well-formed one, with
+/// `status=ok`, in the mode whose purpose is to reject exactly that (#1543).
+///
+/// Measured on `main` at `5616cdb9`, reference bases `c.1 = A`, `c.3 = G`:
+///
+/// ```text
+/// NM_000143.3:c.1G>T          -> rejected  (no sibling to merge with)
+/// NM_000143.3:c.[1G>T;500G>C] -> rejected  (sibling too far to merge)
+/// NM_000143.3:c.[1G>T;3T>A]   -> c.1_3delinsTTA, status=ok   <- the defect
+/// ```
+///
+/// The discriminator was merge distance and nothing else.
+///
+/// # Why here, and not at the merge sites
+///
+/// This is the sixth defect in one family — #1052 (substitutions), #1068 (the
+/// `m.` axis), #1092 (the parser discarded the bases), #1097 (multi-base range
+/// substitutions), #1352 ([`NaEdit::Identity`] omitted) — and each of the first
+/// five was fixed one shape at a time. A per-shape guard inside each merge pass
+/// would be the sixth such patch and would leave the same ordering hazard for
+/// the seventh: any future pass that consumes a member before the validator
+/// runs re-opens the hole.
+///
+/// So the question is asked **on the authored input**, once, before any pass
+/// has had the chance to strip anything. That is ordering-immune by
+/// construction: no rearrangement of the pipeline below can make this check
+/// vacuous, which is precisely what happened to [`stated_reference_bases_match`]
+/// — whose own doc comment predicted it.
+///
+/// Nothing about the merge itself changes, deliberately: the warnings this
+/// returns are *added* to the member-local ones, so no correctly-spelled input
+/// moves and the only behavioural change is that a false reference assertion is
+/// now reported wherever it was authored.
+///
+/// # Contract
+///
+/// Cis only — merging is cis-only, so that is the whole exposure, and a
+/// trans/unknown-phase allele already has every member normalized in isolation.
+/// The verdict itself comes from [`crate::normalize::validate::validate_reference`],
+/// the same function the per-member pipeline uses, so the channel list cannot
+/// drift between the two sites the way #1352's could.
+///
+/// Members this cannot answer for — an offset (intronic) position, a region
+/// with no served bases, an unreadable or short provider window — are skipped
+/// rather than guessed at, exactly as every other refusal in this module does.
+pub(crate) fn authored_member_reference_mismatches<P: ReferenceProvider>(
+    variants: &[HgvsVariant],
+    phase: AllelePhase,
+    provider: &P,
+) -> Vec<crate::normalize::NormalizationWarning> {
+    if phase != AllelePhase::Cis {
+        return Vec::new();
+    }
+    variants
+        .iter()
+        .filter_map(|v| authored_reference_mismatch(v, provider))
+        .collect()
+}
+
+/// One member's authored reference mismatch, if it has one.
+///
+/// The returned warning is built to be **indistinguishable** from the one the
+/// per-member pipeline raises for the same member: same `position` (the
+/// sequence-axis span, via [`region_sequence_delta`]), same `corrected` flag,
+/// same `details` text. That is what lets the caller drop it when the member
+/// also reached the per-member validator, rather than reporting one finding
+/// twice.
+fn authored_reference_mismatch<P: ReferenceProvider>(
+    v: &HgvsVariant,
+    provider: &P,
+) -> Option<crate::normalize::NormalizationWarning> {
+    let kind = cis_kind_of(v)?;
+    let (accession, region, s, e, edit) = cis_axis_parts(v, kind)?;
+    // `r.` spells its bases in the RNA alphabet while the transcript is served
+    // as DNA, so a truthful `r.2u>a` over a `T` would otherwise read as a
+    // mismatch. `normalize_rna` makes the same rewrite before it validates
+    // (#736); making it here too is what keeps the two verdicts identical.
+    let rewritten = matches!(kind, CisKind::Rna)
+        .then(|| crate::normalize::rules::rna_uracil_to_thymine(edit))
+        .flatten();
+    let edit = rewritten.as_ref().unwrap_or(edit);
+
+    let provider_key = accession.transcript_accession();
+    let delta = region_sequence_delta(region, &provider_key, provider)?;
+    // Checked, like every other coordinate conversion in this file: `s`/`e`
+    // come off a parsed description and `delta` off the record's CDS bounds, so
+    // an adversarial span could otherwise overflow and panic in a debug build
+    // where declining is the answer.
+    let start = s.checked_add(delta)?;
+    let end = e.checked_add(delta)?;
+    if start < 1 || end < start {
+        return None;
+    }
+    let span = usize::try_from(end.checked_sub(start)?.checked_add(1)?).ok()?;
+    let bases = provider
+        .get_sequence(
+            &provider_key,
+            u64::try_from(start - 1).ok()?,
+            u64::try_from(end).ok()?,
+        )
+        .ok()?;
+    // A short read means the member runs off the end of the sequence. That is
+    // W4004 `PositionPastEnd`'s finding, not this one, so decline rather than
+    // compare against a truncated window.
+    if bases.len() != span {
+        return None;
+    }
+
+    // `1`/`span` rather than the member's own coordinates: `validate_reference`
+    // indexes `ref_seq` from `start`, and the window fetched above starts *at*
+    // the member. Every arm it dispatches to reads only `[start - 1, end)` or
+    // the span width, so re-basing the pair is equivalent to handing it the
+    // whole sequence and costs one small read instead of the entire transcript.
+    let result =
+        crate::normalize::validate::validate_reference(edit, bases.as_bytes(), 1, span as u64);
+    if result.valid {
+        return None;
+    }
+    // Mirrors the `corrected` reasoning in `normalize_na_edit`: the canonical
+    // `Display` keeps a substitution's stated base and a repeat's unit, so
+    // claiming those were corrected would be dishonest.
+    let corrected = !matches!(
+        edit,
+        NaEdit::Repeat { .. } | NaEdit::MultiRepeat { .. } | NaEdit::Substitution { .. }
+    );
+    Some(crate::normalize::NormalizationWarning::RefSeqMismatch {
+        stated_ref: result.stated_ref.unwrap_or_default(),
+        actual_ref: result.actual_ref.unwrap_or_default(),
+        position: format!("{start}-{end}"),
+        corrected,
+        details: result.warning,
+    })
+}
+
 /// Inclusive 1-based span covering every edit's reference footprint.
 fn edit_span_union(edits: &[GEdit]) -> Option<(i64, i64)> {
     let mut lo = i64::MAX;
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -2780,6 +2780,43 @@ impl<P: ReferenceProvider> Normalizer<P> {
             warnings.push(w);
         }
 
+        // A cis allele's members are merged *before* any of them reaches the
+        // per-member reference validator, and a merge keeps only their alt
+        // bases — so a member consumed by one has its reference assertions
+        // dropped without ever being checked (#1543). Ask the question here, on
+        // the authored input, where the assertions still exist.
+        //
+        // Deliberately outside `normalize_allele` rather than inside it: that
+        // function has several early returns, and the point of this guard is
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

nh13:

Both lenses completed against `7ada1ebe` (CLI: `--base origin/main`, `"findings":2`, 2 lines present — gated on the complete line, no truncation). On current `main`. CI green: 37 rows, 36 pass, 0 fail.

This is the first PR in this batch that can move a normalized output, so I verified the disclosure rather than reading it.

**The before/after table reproduces exactly.** Two release binaries — this branch and `origin/main` — run over the six inputs against the real prepared reference with `--error-mode strict`:

| input | before | after | |
|---|---|---|---|
| `NM_000143.3:c.1G>T` | `normalize_error` | `normalize_error` | unchanged ✓ |
| `NM_000143.3:c.[1G>T;500G>C]` | `normalize_error` | `normalize_error` | unchanged ✓ |
| `NM_000143.3:c.[1G>T;3T>A]` | `ok` → `c.1_3delinsTTA` | `normalize_error` | **flip ✓** |
| `NM_000143.3:c.500G>C` | `ok` | `ok` | unchanged ✓ |
| `NM_000287.4:c.*438TAAA[1]` | `normalize_error` | `normalize_error` | unchanged ✓ |
| `NM_000287.4:c.[*438TAAA[1];2578C>T]` | `ok` → `c.[2578C>T;*442_*445del]` | `normalize_error` | **flip ✓** |

Both pre-fix strings are the ones the body names, and the new diagnostics are `Reference mismatch at 64-64: expected G, found A` and `Reference mismatch at 3412-3412` — the positions quoted. Exactly two status changes, both `accept → reject`, both on inputs that misstate their reference. Nothing correctly-spelled moved.

**And the suite claim holds:** `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1 cargo nextest run --features dev` → **9,609 passed, 0 failed**, working tree clean afterwards, so **nothing was re-blessed**. For an output-capable change that is the strong signal, and it is the one the body leads with.

**The mechanism correction checks out.** `merge::Anchor` is `{ region, start, end, alt, form }` — there is genuinely no channel for an asserted reference, so a member consumed by a merge has its assertions discarded before anything can ask about them. Attributing this to `stated_reference_bases_match` being vacuous would have been the wrong diagnosis, and the PR says so and shows why. Also confirmed: `authored_reference_mismatch` constructs only `RefSeqMismatch`, so the `else { continue }` at the dedup site is defensive rather than a silent drop; the `r.` fold really does reuse `rna_uracil_to_thymine` the way `normalize_rna` does; and the verdict really does come from the same `validate::validate_reference` the per-member pipeline calls, which is what stops #1352's channel-list divergence from recurring.

Two findings.

---

### 1. ⚠️ `strict_rejects` accepts *any* error as a rejection, across 13 call sites

**`tests/it/issue_1543_allele_member_reference_validation.rs:141-148`** · *CLI [1], confirmed*

```rust
fn strict_rejects(input: &str) -> bool {
    Normalizer::with_config(provider(), …ErrorMode::Strict)
        .normalize(&parse_hgvs(input).expect("fixture must parse"))
        .is_err()
}
```

`.is_err()` cannot tell a reference mismatch from `PositionPastEnd` (W4004), an intronic decline, or any other strict-ladder rejection. The PR's claim is specifically that strict rejects **because of** the misstated reference — and this helper is the instrument for it at **13 call sites**, including the two sweeps that carry the "broader than the report" argument: `every_stated_reference_channel_is_validated` (`delAA`, `delAAinsGG`, `dupAA`, `invGGGGG`, `Repeat`) and `every_merging_axis_is_covered` (`g.`/`m.`/`n.`/`r.`).

So if any one of those fixtures were rejected for an unrelated reason, the channel or axis would read as covered while proving nothing. That is a test passing for the wrong reason, in the PR whose thesis is that the *reason* is what was broken — and it is the same shape as the "green on three of five while the defect was fully present" measurement #1540 documents.

Cheap to close: assert the error is a reference mismatch (or that `reference_mismatches` is non-empty), not merely that one occurred. The over-rejection guard `the_rna_alphabet_is_not_read_as_a_mismatch` is unaffected — it asserts the *absence* of rejection, which `.is_err()` handles correctly.

### 2. 🔸 Position-only warning dedup — real tension, but I'd not apply the suggested fix as written

**`src/normalize/mod.rs:2802-2815`** · *CLI [2], confirmed with a caveat*

The dedup drops an authored warning when any existing `RefSeqMismatch` shares its `position`. CodeRabbit is right that this could discard a genuinely distinct finding if two different assertions ever map to one sequence-axis span — the per-member position is computed after shifting, so a shifted member landing on another member's authored span is not impossible.

**But the proposed fix — key on `stated_ref` + `actual_ref` + `position` — risks reintroducing the double report this code exists to prevent.** The whole reason position-only is safe today is the invariant stated at `merge.rs:3195-3202`: the authored warning is built to be *indistinguishable* from the per-member one — same position, same `corrected`, same `details`. If that ever drifts (a `stated_ref` corrected during normalization, say), a full-identity key stops matching and `an_unmerged_member_is_not_reported_twice` goes red — which is at least loud, but the fix would have created the failure rather than prevented one.

The stronger move is to **assert the invariant instead of loosening the key**: pin that the two sites produce byte-identical `RefSeqMismatch` fields for one member, so the property position-only dedup depends on is a test rather than a comment. Then tighten the key later if that assertion ever needs an exception. Either way this is your call, not a blocker — it needs a concrete input producing two distinct assertions at one mapped span, and I could not construct one.

---

**Spec audit: clean.** Reusing `RefSeqMismatch`/W5001 rather than minting a code is right — this is the same finding the per-member path already reports, and a new code would have split one condition across two identifiers. The `r.`-axis fold is the subtle correctness point and the PR gets it the safe way round: without it a truthful `r.2u>a` over a `T` would be **over-rejected**, which is worse than the hole, and `the_rna_alphabet_is_not_read_as_a_mismatch` pins it.

**On the disclosure format**, for the record: this is the shape the `Representation-Change:` rule was added for. The corpus zero is reported as **structural and explicitly not offered as evidence** — 27,656 corpus rows carry a stated reference base and 0 misstate one, because the generator builds every one by reading the byte out of the reference — and the real evidence is a table with a non-zero denominator (3 of 950 rows carry a disagreeing assertion; 2 already reported; 1 laundered by a merge and now caught). That is exactly what `CLAUDE.md`'s "a corpus zero is a claim about the corpus" section asks for, and the honest reading of the single strict flip — an input that was never a valid description of that reference, whose lone form was already rejected — is correct: no stored, accepted, correct string moves.

Only finding 1 seems worth fixing before merge.

PR #1548, opened 2026-08-07
docs(rulings): decide the precedence order and the canonical-form choice

nh13:
Two of the project's `undecided` ruling records are decided here, together, because the first gates the second and the evidence that settles one settles the other. Both had been open long enough that sessions were adopting whichever position they happened to read first.

## `adjudication-precedence-order`

**(1) the spec where it speaks; (2) confluence — one variant, one description; (3) re-derivation from the resulting sequence where the spec is silent; (4) disclosure of any resulting move; (5) stability toward the already-shipped form, as a last-resort tiebreaker only.** Mutalyzer appears nowhere in the order; it is a sanity check.

The record previously said two irreconcilable orders were on file and neither had been chosen. One of them is now *refuted* rather than merely unchosen, so the choice is no longer symmetric.

## `canonical-form-choice-when-both-legal`

**Re-derive from the resulting sequence and emit the form that falls out, subject to every explicit spec tie-break.** Not the input's spelling, and not the previously-shipped string.

## Three claims this project was treating as established, and which do not survive checking

These are the reason the records could be decided at all, and each is recorded in the rationale so it is not re-derived from scratch.

**`basics.md:38` does not state a description-level stability guarantee.** The sentence is *"The **recommendations** for the description of sequence variants are designed to be stable, meaningful, memorable, and unequivocal."* Its grammatical subject is the rule set, not descriptions; it sits inside a section headed `### Versioning` whose next sentence is *"Still, every now and then, small modifications will be required"*; and the only description-level use of "stable" anywhere in the corpus (`numbering.md:131`) is about choosing a genomic reference to avoid transcript-model churn. The stability-first reading was hung on this one line.

**"The spec has no minimality principle" is overstated.** `DNA/complex.md:50` calls the 3' rule *"the general HGVS rule of maintaining the longest unchanged sequence"* — a minimal-change objective, described as general — and six further clauses give length or redundancy as the stated reason for a preference. What remains true is that no minimality *principle* is stated and that `DNA/delins.md:44-47` recommends a non-minimal spanning form in its own worked example, so minimality never overrides an explicit tie-break.

**The filer's issues cannot be ranked as a direction.** Across the confluence family the requested fixes are **3 merge, 3 split, 5 respell**, and #1420 makes that non-reconcilability its own thesis: the corrections *"cannot be reconciled by any per-member-type rule … They are reconciled only by … apply the edit set to the reference, then re-derive the minimal canonical partition from the resulting sequence."* That is a method request, and it is honoured at rank (3).

## Why confluence outranks stability, from the spec rather than from preference

`DNA/complex.md:63` states the value — *"HGVS recommendations try to avoid such conflicts wherever possible"* — in a passage where one derivative chromosome had two legal descriptions, and `:64` shows the method: pick one form exclusively and **retire the other**, which is a representation change. The committee rejected the imperfect-copies proposal on the explicit ground that it *"would add too many options to describe specific variants"*. And `versions/index.md:48` calls cross-version incompatibilities ***"known and intended"***, with no grandfathering, migration or deprecation language anywhere in the corpus — the only remedy offered is version declaration, which is rank (4).

## The counter-evidence, recorded rather than omitted

The spec contradicts itself on re-derivation, and the record says so. `general.md:157-160` says derive from the sequences and that knowledge of the underlying change *"should not be used"*, declaring the provenance-faithful description of its worked example "not correct". But `DNA/delins.md:83` grounds the split-by-default rule in provenance, and `RNA/delins.md:41` makes a form *"preferred when either of the two variants is known as a frequently occurring variant"*. Neither is recoverable from two sequences. Ferro cannot implement them and does not try; this ruling adopts the re-derivation half deliberately, not in ignorance of the other.

The measured cost of re-derivation is also in the record rather than left for someone to rediscover as a refutation: a fewest-members-subject-to-minimal-edit-distance partitioner moved Mutalyzer agreement 200/234 → 177/234 and release drift 31 → 849 rows (27×), and splits are non-unique — 27 of 40 rows admit more than one equally-compliant split, median 2, max 125.

## What is still open

`general.md:56` ranks five types and **`delins` is absent**, so the master tie-break is silent on any contest whose members are all `delins`. `:56` is also not applied uniformly by the spec itself — `DNA/inversion.md:19` prefers an insertion over a duplication for inverted copies. Those fall through to re-derivation and then to rank (5). `codon-carve-out-shape-restriction` remains `undecided`; its evidence is a corpus zero, which is a claim about the corpus.

## Verification

Every citation is checked verbatim against the pinned spec checkout by `generate_spec_fixture`, which refuses a malformed ruling record — so a submodule bump that moves a clause fails the build rather than leaving the citation pointing at unrelated prose. `ruling_records_are_intact` passes with `RULING_STATUSES` updated to match.

`cargo fmt --all -- --check` clean; `cargo clippy --features dev --all-targets -- -D warnings` clean; `cargo nextest run --features dev -E 'binary(it)'` → **5444 passed, 145 skipped**, nothing re-blessed.

Representation-Change: none. Ruling records and their status pins only; no code path is touched and no normalized output moves. The rulings govern how future changes are ranked — each of those will carry its own trailer.


nh13:
@coderabbitai pause

nh13:

Both lenses completed against `a428f8d4` (CLI: `"findings":1`, 1 line present — gated before reading). CI: 33 rows, 0 fail, 5 pending at time of writing. Two files, `+46/−25`.

For a rulings PR the only thing that matters is whether the citations are real and say what the rationale claims, so that is what I checked — independently of the test that checks it.

**All 10 structured clauses quote the spec verbatim** at the pinned commit. And the four load-bearing textual claims in the rationales hold up:

| claim | verdict |
|---|---|
| `versions/index.md:48` calls incompatibilities *"known and intended"* | ✓ verbatim on :48 |
| `basics.md:38` sits under a `### Versioning` heading | ✓ heading at :36; :39 is *"Still, every now and then, small modifications will be required"* |
| `numbering.md:131` is **the only description-level use of "stable"** in the corpus | ✓ 5 corpus hits; only this one's grammatical subject is *descriptions* — the others are "the recommendations" (`basics.md:38`, `versions/index.md:9`) and "identifiers" (`refseq.md:12`, `SVD-WG008.md:21`) |
| `open-issues.md:218` — *"would add too many options…"* | ✓ present (see nit below) |

All **24** citations embedded in rationale prose resolve to real lines too. The `basics.md:38` re-reading is the load-bearing move in the whole PR and it is correct: the subject really is the rule set, the section really is Versioning, and the description-level "stable" really is elsewhere and about reference choice.

**Interop with #1544 measured, not assumed.** #1544 and #1548 edit the *same* `RULING_STATUSES` array and the *same* rulings fixture — #1544 appends two `undecided` records while this one flips two existing ones to `decided` — which is exactly the coupled-sibling shape where a clean `merge-tree` proves nothing. So I stacked them in a detached worktree, forced a regeneration of the gitignored `hgvs_spec_normalization.json` (stale copies survive and produce a false failure here — the trap #1544's own body documents), and ran the suite:

```
6 tests run: 6 passed   — including ruling_records_are_intact
```

No landing-order hazard between them, in either direction.

---

### 1. ⚠️ This ruling supersedes doctrine that is still live in `CLAUDE.md`, and nothing here updates it

**`CLAUDE.md:545-549`** · *agent lens*

```
### There is no minimality principle, and stability is the spec's first stated value

`background/basics.md:38`: "…designed to be **stable**, **meaningful**, …" Minimality is
absent, and stability leads.
```

That heading asserts as settled fact the exact proposition `adjudication-precedence-order` now **refutes** — the ruling demotes stability to rank (5), a last-resort tiebreaker, on the grounds that `basics.md:38`'s subject is the recommendations rather than descriptions. The ruling even names the doctrine it is overturning ("The doctrine that ranked it higher rested on `background/basics.md:38`"), and cites the same clause.

This matters because of #1534's premise: `CLAUDE.md` is where an agent reads policy, and `CONTRIBUTING.md`/fixtures are not. Leaving `:545` in place means the next session reads "stability leads", applies the superseded order, and has no way to know a ruling overrode it — the failure mode #1534 was opened to fix, in the opposite direction. The `:549`-onward paragraph about `delins.md:44-47` is unaffected and should stay.

Same for `:594-595`, which still frames the stability question as open ("The repository's doctrine reads as *do not move a normalized output*, while the downstream filer has said…"). The ruling settles that: nobody asked for the string to stay put, and disclosure at rank (4) is the remedy. Worth updating in the same PR so the decision and the place people read it cannot drift apart.

### 2. 🔸 Pin the `governing` clause in the status table, not just its internal consistency

**`tests/it/hgvs_spec_normalization_tests.rs:809-816`** · *CLI [1], confirmed as a real gap*

`ruling_records_are_intact` already enforces the shape well — a `decided` ruling must name a `governing` clause (`:876-879`), that clause must be among the ruling's own cited clauses (`:883-884`), and an `undecided` one must name neither `governing` nor `deviates_from` (`:901-902`). That is more than I expected and it is the right set of invariants.

What is *not* pinned is **which** clause governs. `RULING_STATUSES` externally pins `(id, status)`, so a silent change of `governing` from `general.md:157` to some other already-cited clause passes every check. Given that the governing clause is the ruling's actual operative content, extending the tuple to `(id, status, Option<&str>)` is a cheap and consistent extension of a table that already exists for exactly this purpose.

### 3. 🔸 One rationale quote is not byte-verbatim, and rationale quotes are unguarded

The rationale renders `open-issues.md:218` as *"would add too many options to describe specific variants"*. The source is:

```
Accepting the proposal, without a whole range of specifications, would add **too many** options to describe specific variants.
```

— i.e. the emphasis markers around **too many** are dropped. Harmless in substance, but worth noting for a structural reason: the 10 structured `clauses` are held byte-verbatim by a test, while the 24 citations in rationale prose are checked by nothing. So a quote that is *presented* as the spec's words can drift from them, and promoting this one to a structured clause later would fail the verbatim check for a reason nobody would expect. Either normalize emphasis on both sides, or treat prose citations as paraphrase and reserve quotation marks for the structured ones.

---

### Not a finding, recorded so it is not mistaken for one

This branch predates the #1546 merge, so a naive `git diff --name-only origin/main <head>` lists `examples/dump_normalized_corpus.rs` as changed. It is not — that file differs because the branch is *behind* `main`, and against its own merge-base (`5616cdb9`) the PR touches exactly the two files above. `merge-tree` reports 0 conflicts against current `main`. Mentioning it because I made that misreading once during this review before catching it.

**Spec audit: clean, and unusually careful.** Both rulings record their counter-evidence rather than only their support — `canonical-form-choice-when-both-legal` states plainly that the spec contradicts itself, that `DNA/delins.md:83` and `RNA/delins.md:41` ground preferences in provenance and population frequency that ferro cannot see, and that it adopts the re-derivation half "deliberately, not in ignorance of the other half". It also corrects an overstatement that was previously asserted flatly in this very file (that the spec has no minimality notion) by naming `DNA/complex.md:50` and six length/redundancy passages against itself. That is the right shape for a record whose purpose is to stop a future session re-litigating from scratch.

Only finding 1 seems worth doing before merge — the decision and the document people actually read should land together.

PR #1549, opened 2026-08-07
test(confluence): record two adjudications from the divergent cis classes

### LINKED ISSUE #1524: Normalizer: a split delins leaves two *adjacent* members (separation 0), which delins.md:16 requires be one delins

nh13:
Two ruling records for the divergent cis confluence classes, plus the shape-census harness that produced them. One record is decided and ferro conforms; the other is honestly undecided.

## What the decided record says now

`delins-adjacent-members-when-both-consume-reference` — a normalized allele must not leave two members with no unchanged nucleotide between them when both consume reference bases. `general.md:34` and `DNA/delins.md:17` govern members separated by *one or more* nucleotides and say nothing about members separated by none. `DNA/delins.md:16` covers that case, and `DNA/substitution.md:32` adjudicates it by name: for `LRG_199t1:c.79_80delinsTT` it renders `c.[79G>T;80C>T]` `class="invalid"` and calls it "not correct" in as many words. `substitution.md:32` governs; `general.md:56` is the one clause readable the other way and is deviated from.

The scope is load-bearing and unchanged: this reaches only `sub`/`del`/`delins`/`inv` on both sides. Where a `dup` or an `ins` is one side, `DNA/duplication.md:18` genuinely competes and the ruling declines. A third shape turned up in re-measurement — a repeat expansion flush against a neighbour — and is named as outside the scope rather than folded in.

## #1537 overtook the original framing

The record was written against a tree where ferro emitted `NM_TEST.1:c.[10_12delinsTAA;13A>T]` for `c.10_13delinsTAAT`, cutting at the codon boundary between codon 4 and codon 5 and leaving the halves flush. It was framed as pinning a **deviation**, and the test asserted the wrong form on purpose.

#1537 ("never split a delins into members on consecutive nucleotides") merged and closed #1524 before this branch was pushed. Ferro now emits the merged form. The record is therefore re-authored as **adjudicated-correct**: the test asserts the spec's form and is a regression guard against a decided answer. Those two kinds are not interchangeable, so the record and the `RULING_STATUSES` comment both say which one it is.

The record is not deleted and not weakened — it is vindicated. It found #1524's defect independently, on the designed cis corpus, while #1524 had filed it at 61 rows / 59 distinct over a different corpus. Two disjoint corpora reaching one clause is the evidence that the reading is about the description rather than about one generator. Mutalyzer was not consulted and is not the authority here; both precedence orders on record in `adjudication-precedence-order` rank the spec first where it speaks.

## The corrected figures

Re-measured with `examples/dump_confluence_divergences.rs` over the 11,272-class designed cis corpus, 3' direction, at `3ab10a09`. Old figures are the ones this branch originally committed, measured pre-#1537.

| | committed claim (pre-#1537) | re-measured now |
|---|---:|---:|
| divergent classes | 4,639 | **4,643** |
| largest family (`spanning-vs-split/min-2+`) | 3,056 (65.9%) | **3,066 (66.0%)** |
| families | 9 | **8** (`same-arity/has-0` is gone) |
| zero interior gap, most-split output (`has-0` census bucket) | 104 | **31** |
| zero interior gap in any output | — | **36** |
| **in this ruling's scope** | **65, all `c.`, none `g.`** | **0** |
| carved out — `dup`/`ins` on one side | 39 | **33** |
| a repeat on one side (uncontemplated) | — | **3** |
| spanning output disjoint from every multi-member output | 4,567 (98.4%) | **4,568 (98.4%)** |

The headline "65 classes, all on the coding axis" is dead, and the codon-splitter attribution that rode on it is dead with it. #1537 removed the shape rather than moving it to another axis: every one of the 36 surviving zero-gap adjacencies is in the carve-out the record explicitly declined to reach. That is recorded in the rationale rather than quietly overwritten, because a measurement that kills a claim is worth as much as the ruling.

The 4,643 agrees exactly with the committed `THREE_PRIME` census in `cis_confluence_axis.rs` (4,509 + 115 + 19), which #1537 already re-pinned on `main`.

**Note on one figure that was reported to me as moving and does not.** A hand-off table described the post-#1537 governed set as splitting 11 `c.` / 17 `g.` / 3 other. That is the axis breakdown of the `has-0` *family buckets* (`spanning-vs-split/has-0` 11 `c.` + 17 `g.`, `split-vs-split/has-0` 3 `g.`), not of the set this ruling governs. Enumerating the adjacency kind-pairs directly, the governed set is empty: the zero-gap adjacencies that remain are `dup`/`sub` ×18, `ins`/`repeat` ×7, `delins`/`dup` ×6, `del`/`dup` ×2, `repeat`/`repeat` ×2, `del`/`repeat` ×1, `repeat`/`sub` ×1. Not one is a pair of reference-consuming members.

## The undecided record still holds

`separation-is-a-property-of-the-spelling-not-of-the-variant` stays `undecided`, re-checked rather than assumed. Its demonstration turns on a separation-0 pair being merged while a separation-1 pair is split: on the corpus's own transcript `c.[9del;10del;13del]` and `c.[9del;11del;13del]` denote one sequence (verified through the apply oracle) and present separations {0,2} and {1,1}. #1537 strengthened the merge half and left the split half alone, so both pinned assertions passed across it unchanged and the two spellings still reach two outputs. Its corpus figures are re-measured (642 of 9,442 variants multi-design, 286 of differing separation, 4,568 spanning-disjoint) and the rationale now records the re-check.

## Tests

Four pinned cases, each stating whether it is adjudicated-correct, a scope carve-out, or a pinned open disagreement:

- `the_separation_two_members_present_is_not_a_property_of_the_variant` — the undecided record's demonstration.
- `two_adjacent_members_that_both_consume_reference_are_one_delins` — the decided record, now a conformance guard (was `a_codon_boundary_cut_leaves_two_adjacent_members_which_the_spec_marks_invalid`).
- `a_dup_flush_against_a_del_is_left_alone` — **new.** The scope limit, pinned on real corpus class `s00-c-m4-sep3-p8-rot3` rather than an invented input: `c.16_23dup` and `c.24_31del` are flush and stay flush, because `duplication.md:18` competes.
- `a_spanning_delins_and_its_aligned_split_are_two_fixed_points` — the largest family, blocked on `delins-merge-vs-individual-gap-two-or-more`.

The first three verify their spellings denote one sequence through `hgvs_to_spdi` before comparing normalized outputs, so none can pass on a false premise.

## Verification

- `cargo run --features dev --bin generate_spec_fixture` — clean; it is what validates the edited records (quote-checks every citation against the spec checkout, and refuses an `undecided` record naming a governing clause or citing fewer than two).
- `cargo nextest run --features dev -E 'test(ruling_records_are_intact)'` — `1 test run: 1 passed`.
- `cargo fmt --all -- --check` — clean. `cargo clippy --features dev --all-targets -- -D warnings` — clean.
- `cargo nextest run --features dev -E 'binary(it)'` — `5448 tests run: 5448 passed, 145 skipped`. **Nothing re-blessed.** For an output-touching record that is the strong signal: every existing expectation already agreed.

## Review fixes (dual-review sweep)

All seven confirmed findings are fixed, amended into the commit that introduced each.

**F1 — three unit tests had never run.** The `[[example]]` entry for `dump_confluence_divergences` lacked `test = true`, so cargo never built the target's tests: `cargo nextest run --features dev -E 'binary(dump_confluence_divergences)'` reported `0 tests run`. Added, and all three pass — including `an_insertion_consumes_no_reference_position`, the pin on the arithmetic that has already published wrong numbers. Nothing was hiding behind them. Two further tests are added below.

**F2 — three uncounted silent-drop paths.** A `Drops` struct now counts spellings `parse_hgvs` rejects, spellings normalization errors or panics on, and divergent classes excluded because a member's spans could not be resolved. `render_census` prints all three unconditionally. A failed span resolution used to become `unwrap_or_default()` → `[]` → `classify_gaps` → `no-gaps`, so a multi-member output whose geometry was never measured left the `has-0` population silently; such a class is now excluded and counted rather than classified. **Measured: `0 spellings unparsed, 0 spellings unnormalized, 0 divergent classes excluded`.**

**F3 — `relation` misclassified arity-`[1]` classes.** `arities.first() == Some(&1)` filed a class whose outputs are all single-member as `spanning-vs-split`, asserting a split that is not there and making `same-arity` unreachable at arity 1. Now `arities.len() > 1 && arities.contains(&1)`, extracted as `relation_of` and pinned by a new unit test. **Measured: the arity-`[1]` population is 0, so the headline family was not in fact contaminated.**

**F4 — `spans` were coordinate-sorted while `kinds` stayed in text order**, then paired positionally, so for any allele not already written in coordinate order the two described different members — and the decided record's adjacency kind-pair enumeration is exactly that join. `spans_of` is replaced by `members_of`, which returns rendered text paired with its interval in one order. **Measured: no record changes, so no allele in this corpus writes its members out of coordinate order.**

**F5 — two stale pre-#1537 figures** in a doc comment (`3 056 of 4 639`, `nine shapes`) corrected to `3 066 of 4 643` and `eight shapes`. Three more of the same vintage were found in the generator's own header and corrected too (`6 633` → `6 629` converged, `4 639` → `4 643`, `673` → `663` distinct shapes).

**F6 — the record cited a separation-ONE equivalence class.** Fixed the preferred way rather than by dropping the association: a new class `dna-delins-vs-two-substitutions-79-80` is built from the ruling's own governing clause, `DNA/substitution.md:32`, which renders `c.[79G>T;80C>T]` `class="invalid"` for `LRG_199t1:c.79_80delinsTT`. It is the fixture's only separation-zero class, and `delins-adjacent-members-when-both-consume-reference` now points at it instead of at `145-147` (which remains where it belongs, on `delins-codon-carve-out-gap-one`). It resolves **`confluent`** — both spellings reach `LRG_199t1:c.79_80delinsTT` — so it is a positive control for the decided ruling rather than another pinned failure.

**F7 — the module header's two-and-two tally** is replaced by the three kinds the tests actually are: one adjudicated-correct, one scope carve-out, two pinned open disagreements.

**An eighth, found while fixing F5/F7 and not raised by either lens.** Test 4's doc comment described ruling `delins-merge-vs-individual-gap-two-or-more` as status **`undecided`**. It was already `decided` at this branch's own base — decided for `:47` and *scoped*. Only the status word and the framing around it are corrected: the record settles the spanning side, and what this test pins as open is the other side, whether that scope reaches a spelling arriving already split. Nothing here decides that; both outputs stay pinned as observations.

## Census: re-measured after the fixes, and unchanged

F2 and F3 can legitimately move the numbers these records quote, so the generator was re-run and diffed against the pre-fix run. **The census output is byte-identical.** Every figure the two ruling records and the module doc comment quote stands:

| figure | before fixes | after fixes |
|---|---:|---:|
| divergent classes | 4 643 | 4 643 |
| families | 8 | 8 |
| largest, `spanning-vs-split/min-2+` | 3 066 (66.0%) | 3 066 (66.0%) |
| `has-0` arms | 28 + 3 = 31 | 28 + 3 = 31 |
| `input-partition-preserving` | 3 189 (68.7%) | 3 189 (68.7%) |
| distinct shapes | 663 | 663 |
| dropped | *not reported* | 0 / 0 / 0 |

The three defects were real but each had an **empty population on this corpus** — 0 arity-`[1]` classes, 0 unresolved spans, 0 records changed by the join fix. Reporting that as a structural zero rather than as evidence of safety: the corpus is what makes them inert, not the code, and the counters now make a future non-zero visible instead of silent.

## Conflict note

Rebased onto `origin/main` after #1544 landed. That PR appends to the same `rulings` array and the same `RULING_STATUSES` tail, so both files conflicted and were hand-resolved as a **union** — #1544's two `undecided` records and this branch's two both survive, in the JSON and in `RULING_STATUSES`, which `ruling_records_are_intact` checks stay in sync. Ten records total. `hgvs_spec_normalization_overrides.json` is committed, not generated, so nothing was regenerated to resolve it. The diff against `main` is purely additive.

Representation-Change: none. No file under `src/` is touched — the diff is one fixture JSON, one new test module, and one test-registry comment. Measured rather than assumed: the full `it` suite passes with zero re-blessed expectations, and the committed `THREE_PRIME`/`FIVE_PRIME` censuses in `cis_confluence_axis.rs` are unchanged. The one output this branch's expectations moved (`c.10_13delinsTAAT`) was moved by #1537, which declared it; this PR only stops asserting the pre-#1537 form.

Closes nothing; #1524 was closed by #1537.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -351,6 +351,16 @@ required-features = ["dev"]
 [[example]]
 name = "dump_partitions"
 required-features = ["dev"]
+
+# `test = true` because the target carries `#[cfg(test)]` unit tests pinning the
+# separation arithmetic — the one thing here that has already published numbers
+# that were wrong. Cargo does not build a target's tests unless it opts in, so
+# without this they silently never run.
+[[example]]
+name = "dump_confluence_divergences"
+required-features = ["dev"]
+test = true
+
 [[bin]]
 name = "ferro"
 path = "src/bin/ferro.rs"
--- a/examples/dump_confluence_divergences.rs
+++ b/examples/dump_confluence_divergences.rs
@@ -1,0 +1,903 @@
+//! Dump the **divergent** confluence classes measured by
+//! `tests/it/cis_confluence_axis.rs`, one record per class, and census them by
+//! *shape*.
+//!
+//! # Why
+//!
+//! The axis test pins a census — 11 272 classes, 6 629 converged in the 3'
+//! direction — and prints ten worst offenders. That is enough to detect a
+//! regression and nowhere near enough to *adjudicate* the 4 643 that diverge.
+//! Adjudicating them one row at a time is not feasible and is not the right
+//! unit anyway: the prior campaign
+//! (`campaigns/2026-08-06-separation-rule-adjudication.md`) collapsed 1 957 rows
+//! to ~32 shapes with the top nine covering 91%, and the same collapse works
+//! here.
+//!
+//! So this emits, per divergent class, every spelling and what it normalized
+//! to, plus a **shape key** derived from the disagreement itself rather than
+//! from the class's design parameters:
+//!
+//! ```text
+//! 1×delins | 2×[delins,delins]
+//! ```
+//!
+//! meaning one spelling reached a single spanning `delins` and another reached
+//! a two-member allele of two `delins`. Shapes are what a spec clause can be
+//! applied to; individual rows are not.
+//!
+//! # Geometry, and the arithmetic trap it exists to avoid
+//!
+//! An insertion consumes **no** reference position. Each member is modelled as
+//! a closed interval `[lo, hi]` over the reference, with an insertion left
+//! *empty* (`lo = A + 1`, `hi = A`), so that
+//!
+//! ```text
+//! separation = next.lo - prev.hi - 1
+//! ```
+//!
+//! holds uniformly across every edit kind. Treating an insertion's `A_B` anchor
+//! as a consumed two-base span double-counts and has already invalidated one
+//! published distribution.
+//!
+//! The intervals come from `hgvs_to_spdi`, which applies the description to the
+//! reference — the same normalizer-independent oracle the corpus generator used
+//! to establish ground truth. Nothing here asks the normalizer what a member
+//! covers.
+//!
+//! # Usage
+//!
+//! ```text
+//! cargo run --features dev --example dump_confluence_divergences -- --stats
+//! cargo run --features dev --example dump_confluence_divergences -- --out /tmp/div.json
+//! cargo run --features dev --example dump_confluence_divergences -- --direction 5prime --stats
+//! ```
+
+use std::collections::{BTreeMap, BTreeSet};
+use std::fmt::Write as _;
+use std::path::PathBuf;
+use std::process::ExitCode;
+
+use clap::Parser;
+use serde::{Deserialize, Serialize};
+
+use ferro_hgvs::reference::transcript::{Exon, GenomeBuild, ManeStatus, Strand, Transcript};
+use ferro_hgvs::reference::MockProvider;
+use ferro_hgvs::spdi::hgvs_to_spdi;
+use ferro_hgvs::{parse_hgvs, HgvsVariant, NormalizeConfig, Normalizer, ShuffleDirection};
+
+/// Kept in step with `examples/generate_cis_confluence_corpus.rs` and
+/// `tests/it/cis_confluence_axis.rs`; all three must agree on the padding, the
+/// accessions and the CDS or the coordinates mean different things.
+const PAD_OFFSET: usize = 256;
+const GENOMIC_CONTIG: &str = "NC_TEST.1";
+const TX_ACCESSION: &str = "NM_TEST.1";
+const TX_CONTIG: &str = "chr_synth";
+const CDS_START: u64 = 1;
+const CDS_END: u64 = 63;
+
+const DEFAULT_CORPUS: &str = "tests/fixtures/cis/cis_confluence_corpus.json";
+
+#[derive(Parser, Debug)]
+#[command(about = "Dump and census the divergent cis confluence classes")]
+struct Cli {
+    /// The corpus written by `generate_cis_confluence_corpus`.
+    #[arg(long, default_value = DEFAULT_CORPUS)]
+    corpus: PathBuf,
+    /// Write the per-class records here as JSON.
+    #[arg(long)]
+    out: Option<PathBuf>,
+    /// Print the shape census only.
+    #[arg(long)]
+    stats: bool,
+    /// `3prime` (default) or `5prime`.
+    #[arg(long, default_value = "3prime")]
+    direction: String,
+    /// How many shapes to list in the census.
+    #[arg(long, default_value_t = 40)]
+    top: usize,
+    /// How many exemplar classes to print per shape.
+    #[arg(long, default_value_t = 0)]
+    examples: usize,
+}
+
+// ---------------------------------------------------------------------------
+// Corpus
+// ---------------------------------------------------------------------------
+
+#[derive(Deserialize)]
+struct Class {
+    id: String,
+    axis: String,
+    core: String,
+    denoted: String,
+    members: usize,
+    separation: usize,
+    payload: usize,
+    pattern: String,
+    block_start: usize,
+    reference_block: String,
+    alt_block: String,
+    spellings: Vec<String>,
+}
+
+#[derive(Deserialize)]
+struct Corpus {
+    classes: Vec<Class>,
+}
+
+/// The confluence question asked one level up: a **variant** is one reference
+/// and one denoted sequence, and several classes can denote the same one by
+/// different designs. `cis_confluence_axis` compares outputs only *within* a
+/// class, so a variant two classes describe differently is invisible to it.
+///
+/// This matters beyond bookkeeping, because the design parameter the separation
+/// rule keys on is exactly what differs between such classes: two members one
+/// nucleotide apart and four members zero apart can denote the same sequence.
+/// `(axis, reference core, denoted sequence)` — one *variant*, regardless of how
+/// many designs reach it.
+type VariantKey = (String, String, String);
+
+/// Every output any class of one variant reached, whether every one of those
+/// classes converged internally, and how many classes there were.
+type VariantOutputs = (BTreeSet<String>, bool, usize);
+
+/// Everything the pass declined to measure.
+///
+/// Printed unconditionally by [`render_census`], and every field is expected to
+/// be **zero**: the spellings come from the corpus generator and the outputs
+/// from a successful `normalize`. That is the argument for printing them, not
+/// for omitting them. A silently-dropped spelling leaves its class with fewer
+/// surviving outputs, and a class reduced to one survivor is then recorded as
+/// *converged* — so an uncounted drop does not merely shrink the sample, it
+/// moves a class from the divergent side of the census to the converged side.
+/// A printed `0` turns that into evidence; silence is indistinguishable from a
+/// run where it was not zero.
+#[derive(Default)]
+struct Drops {
+    /// Spellings `parse_hgvs` rejected.
+    unparsed_spellings: usize,
+    /// Spellings normalization errored or panicked on.
+    unnormalized_spellings: usize,
+    /// Divergent classes excluded because some output's member spans could not
+    /// be resolved, so the class's geometry was never measured. Excluded rather
+    /// than classified: see [`members_of`].
+    unresolved_classes: usize,
+}
+
+struct CrossClass {
+    /// Distinct `(axis, reference, denoted)` variants in the corpus.
+    variants: usize,
+    /// Variants reachable from more than one class.
+    multi_class: usize,
+    /// Multi-class variants whose classes do not agree on a single output.
+    multi_class_divergent: usize,
+    /// … of those, the ones where **every** class converged internally, so the
+    /// disagreement is visible only across classes.
+    hidden_by_the_class_boundary: usize,
+}
+
+// ---------------------------------------------------------------------------
+// Records
+// ---------------------------------------------------------------------------
+
+/// One normalized output, with the geometry the adjudication turns on.
+#[derive(Serialize, Clone)]
+struct Output {
+    text: String,
+    /// Number of cis members.
+    arity: usize,
+    /// Per-member edit kind. Ordered by [`Output::spans`], not by written
+    /// order — the two are one member order so that `kinds[i]` and `spans[i]`
+    /// always describe the same member.
+    kinds: Vec<String>,
+    /// Per-member closed reference interval `[lo, hi]`, 0-based over the core,
+    /// ascending. An insertion is empty (`lo = hi + 1`), so a separation
+    /// computed as `next.lo - prev.hi - 1` is uniform across kinds.
+    spans: Vec<(i64, i64)>,
+    /// Gaps between consecutive members, from the spans above. Empty for a
+    /// single-member output.
+    separations: Vec<i64>,
+    /// Which of the class's spellings reached this output.
+    from: Vec<String>,
+}
+
+/// One divergent class.
+#[derive(Serialize)]
+struct Record {
+    id: String,
+    axis: String,
+    /// The design's own parameters, so a shape can be cross-tabbed against them.
+    design_members: usize,
+    design_separation: usize,
+    design_payload: usize,
+    design_pattern: String,
+    core: String,
+    denoted: String,
+    block_start: usize,
+    reference_block: String,
+    alt_block: String,
+    /// The fine shape key: the sorted arity/kind signature of the distinct
+    /// outputs. Useful for exemplars; too fine to adjudicate against (663
+    /// distinct values over 4 643 classes).
+    shape: String,
+    /// The coarse shape key the adjudication is written against:
+    /// `<relation>/<gap class>`. See [`Record::relation`] and
+    /// [`Record::gap_class`].
+    family: String,
+    /// `typing` when every output partitions the reference identically and only
+    /// the edit kinds differ; `partition` when the member boundaries differ.
+    disagreement: String,
+    /// How the outputs' arities relate: `spanning-vs-split` when one output is
+    /// a single member and another is not; `split-vs-split` when they are all
+    /// multi-member but of different arity; `same-arity` when the arities agree
+    /// and only the boundaries or the kinds move — which includes the case
+    /// where every output is a single spanning member.
+    relation: String,
+    /// Every distinct output arity, ascending.
+    arities: Vec<usize>,
+    /// The interior gaps of the **most split** output — the one the merge
+    /// question is asked about. Every gap, not only the first: checking only
+    /// the first under-counted a prior distribution by 54 per arm.
+    split_gaps: Vec<i64>,
+    /// Bucketed from [`Record::split_gaps`]: `all-0` (every interior gap is
+    /// adjacency), `has-0` (some but not all), `min-1` (smallest gap is one
+    /// nucleotide), `min-2+` (every gap is two or more). This is the quantity
+    /// `general.md:34-35` and `DNA/delins.md:16-18` key on.
+    gap_class: String,
+    /// True when every output's arity equals the member count of every spelling
+    /// that reached it — i.e. the normalizer preserved the input's partition
+    /// rather than re-deriving one. This is the mechanism behind the divergence
+    /// and it is worth stating separately from the shape.
+    input_partition_preserving: bool,
+    outputs: Vec<Output>,
+}
+
+// ---------------------------------------------------------------------------
+// Providers — the same construction the generator and the axis test use
+// ---------------------------------------------------------------------------
+
+fn padded(core: &str) -> String {
+    let pad = "ACGT".repeat(PAD_OFFSET / 4);
+    format!("{pad}{core}{pad}")
+}
+
+fn reference_for(class: &Class) -> (MockProvider, String) {
+    if class.axis == "g" {
+        let mut provider = MockProvider::new();
+        provider.add_genomic_sequence(GENOMIC_CONTIG, padded(&class.core));
+        return (provider, padded(&class.core));
+    }
+    let mut provider = MockProvider::new();
+    let tx_len = class.core.len() as u64;
+    let g_start = PAD_OFFSET as u64 + 1;
+    let g_end = PAD_OFFSET as u64 + tx_len;
+    let exon = Exon::with_genomic(1, 1, tx_len, g_start, g_end);
+    let transcript = Transcript::new(
+        TX_ACCESSION.to_string(),
+        Some("SYNTH".to_string()),
+        Strand::Plus,
+        class.core.clone(),
+        Some(CDS_START),
+        Some(CDS_END),
+        vec![exon],
+        Some(TX_CONTIG.to_string()),
+        Some(g_start),
+        Some(g_end),
+        GenomeBuild::GRCh38,
+        ManeStatus::None,
+        None,
+        None,
+    );
+    provider.add_genomic_sequence(TX_CONTIG, padded(&class.core));
+    provider.add_transcript(transcript);
+    (provider, class.core.clone())
+}
+
+fn axis_offset(axis: &str) -> i64 {
+    if axis == "g" {
+        PAD_OFFSET as i64
+    } else {
+        0
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Shape
+// ---------------------------------------------------------------------------
+
+/// Classify one rendered member by its edit kind. Textual because the question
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai pause

nh13:

Both lenses completed against `7f7ca8af` (CLI: `--base origin/main`, `"findings":14`, 14 lines present — gated on the `complete` line, no truncation, no rate limit; agent lens from a detached `sweep-1549/` with the submodule initialised). On current `main` (`3ab10a09`), 0 conflicts. CI green: 37 rows, 36 pass, 0 fail.

**Verified independently before reviewing**, because a rulings PR is only as good as its citations:

- All six single-line clauses resolve **verbatim** at the pinned `6f85311` — `substitution.md:32`, `delins.md:16` and `:17`, `general.md:34` and `:56`, `duplication.md:18` — as do both range citations (`delins.md:79-84`, `open-issues.md:77-78`).
- The census arithmetic is internally sound: the eight arms sum to exactly **4 643**, the top four to 92.6%, every percentage rounds correctly, `has-0` = 28 + 3 = **31**, and 4 568 / 4 643 = 98.4%.
- `cargo nextest run -E 'test(cis_confluence_adjudication) + test(ruling_records_are_intact)'` → **5 passed**, working tree clean afterwards. Nothing re-blessed, which is the claim the PR body leads with and it holds.

The re-measurement is the right shape and the body is right that a measurement killing a claim is worth as much as the ruling. What follows is about the instrument that produced it.

---

### 1. ⚠️ The three tests in the new generator have never run — `test = true` is missing

**`Cargo.toml:340-342`** · *CLI [14], confirmed empirically*

```
$ cargo nextest run --features dev -E 'binary(dump_confluence_divergences)'
Starting 0 tests across 0 binaries (13 binaries skipped)
Summary [0.000s] 0 tests run: 0 passed, 0 skipped
```

`examples/dump_confluence_divergences.rs:741` opens a `#[cfg(test)] mod tests` with three `#[test]` functions. Cargo does not build an example target's tests unless the target opts in, and the new `[[example]]` entry does not.

This repo already knows: `Cargo.toml:291-293` says it in as many words —

> `test = true` because both carry `#[cfg(test)]` unit tests, in the target and in the shared `common/` modules. **Cargo does not build a target's tests unless it opts in, so without this they silently never run.**

— and six other entries in the file set it. So the convention, the rationale and the warning are all present three entries above the new one.

What makes it more than a config nit is which test is dead. `an_insertion_consumes_no_reference_position` carries this doc comment:

> The separation arithmetic is the one thing here that has already produced published numbers that were wrong, so it is pinned directly.

The arithmetic that has already published wrong numbers is pinned by a test that does not execute. One line fixes it; I would run it before merging, since a pin that has never run is a pin nobody has checked.

### 2. ⚠️ The generator has three uncounted silent-drop paths, and they land on the figures the records quote

**`examples/dump_confluence_divergences.rs`** · *CLI [4] and [9], both confirmed by reading*

This is the same defect family #1540 and #1545 each documented in their own generators — a pass that produces a complete-looking artefact from an incomplete run. Here it reaches an adjudication record's headline rather than a fixture.

**(a) A failed span resolution is classified as "no gaps"** (`:512`). `spans_of` returns `Option` and gives up on a member that will not parse, will not map through `hgvs_to_spdi`, or overflows the offset conversion:

```rust
let spans = spans_of(&provider, &text, offset).unwrap_or_default();
let separations = separations(&spans);
```

`unwrap_or_default()` yields `[]`; `separations(&[])` yields `[]`; `classify_gaps(&[])` returns `"no-gaps"` (`:355-357`). So an output with two or more members whose spans could not be resolved is filed as having **no interior gaps** — which is exactly how it leaves the `has-0` population. That population is what the decided record's headline rests on:

> 36 leave a zero interior gap in any output (31 counting only the most-split output) … and **ZERO fall in this ruling's scope**.

A zero produced by a step that silently removes candidates from the numerator is the case `CLAUDE.md`'s "a corpus zero is a claim about the corpus" section is about. The same empty `spans` also makes two unresolved outputs compare *equal*, so the class can be filed `disagreement: typing` — "every output partitions the reference identically" — on the strength of two measurements that did not happen.

**(b) Dropped spellings are counted as agreement** (`:476-490`):

```rust
let Ok(variant) = parse_hgvs(spelling) else { continue };
let normalized = std::panic::catch_unwind(...);
let Ok(Ok(value)) = normalized else { continue };
...
entry.1 &= grouped.len() <= 1;      // "converged"
```

A spelling that fails to parse, errors, or panics is dropped, and a class reduced to one surviving spelling then satisfies `grouped.len() <= 1` and is recorded as **converged**. Nothing counts the drops and `render_census` never prints them, so the divergent/converged split cannot be read as a measurement of the corpus — only of the subset that survived.

I would expect both populations to be empty in practice (these spellings come from the corpus generator and the outputs from a successful `normalize`). That is the argument for *counting* them rather than for leaving it. A printed `dropped: 0` turns both of these into evidence; the current silence is indistinguishable from a run where they were not zero — and the record's `0` is quoted as a finding.

### 3. ⚠️ `relation` cannot ever produce `same-arity` for a single-member class

**`examples/dump_confluence_divergences.rs:542-548`** · *CLI [8], confirmed*

```rust
let relation = if arities.first() == Some(&1) {
    "spanning-vs-split"
} else if arities.len() > 1 {
    "split-vs-split"
} else {
    "same-arity"
};
```

`arities` is deduped and sorted, and this block is only reached when `grouped.len() >= 2`, i.e. two or more *distinct* outputs. When those outputs are all single-member, `arities == [1]`, `first()` is `Some(&1)`, and the class is filed **`spanning-vs-split`** — a name that asserts one output is spanning and another is split, when in fact both are spanning and only the spelling differs. The `same-arity` arm is unreachable for arity 1.

That matters here because `spanning-vs-split/min-2+` is the headline family (3 066, 66.0%) and the module's central mechanism claim is stated over it — "the single spanning spelling reaches an output disjoint from every output the multi-member spellings reach", 4 568 of 4 643. Any arity-`[1]` class is currently inside that number and does not belong to the claim. Whether there are any is unmeasured, which is the same problem as 2.

The fix is the ordering CodeRabbit gives: select `spanning-vs-split` only when `arities.len() > 1 && arities.contains(&1)`.

### 4. ⚠️ `spans` are coordinate-sorted while `kinds` stay in text order

**`examples/dump_confluence_divergences.rs:341` vs `:506-512`** · *CLI [10], confirmed*

`spans_of` ends with `spans.sort()`. `kinds` is built separately from `member_texts(&text)`, which splits the rendered allele on `;` and preserves **text order**. `Output` then pairs them positionally and serialises both. For any allele whose members are not already in coordinate order, `kinds[i]` and `spans[i]` describe different members.

Deriving `separations` from sorted spans is correct and should stay. The exposure is the join: the decided record's rationale enumerates adjacency **kind-pairs** —

> `dup`/`sub` ×18, `ins`/`repeat` ×7, `delins`/`dup` ×6, `del`/`dup` ×2, `repeat`/`repeat` ×2, `del`/`repeat` ×1, `repeat`/`sub` ×1. Not one is a pair of reference-consuming members.

— which is exactly a kinds×spans join, and it is the sentence that establishes the ruling's scope is empty. Sorting both from one member order (or sorting neither) removes the question.

### 5. 🔸 Two stale pre-#1537 figures survive in a doc comment

**`tests/it/cis_confluence_adjudication.rs:298` and `:315-316`** · *found by both lenses independently*

```
:298   the largest family in the pile — 3 056 of 4 639 classes, 65.9 %
:315   So the whole 4 639 is one question wearing nine shapes
```

The module header 290 lines above says **4 643** over **eight** families (`:4`, `:11`, `:13-14`, `:35`), and the PR body's own correction table lists 3 056 → **3 066**, 4 639 → **4 643**, nine → **eight**. The correction reached the header and both JSON rationales and stopped short of this one doc comment. Given the body makes the movement of these figures its thesis, the stale pair reads as a live claim.

### 6. 🔸 The record points at a separation-**one** equivalence class, and the separation-zero one does not exist

**`hgvs_spec_normalization_overrides.json:518-520`** · *CLI [1], confirmed with a constructive variant*

`delins-adjacent-members-when-both-consume-reference` lists `equivalence_classes: ["dna-delins-vs-two-substitutions-145-147"]`. That class is `LRG_199t1:c.145_147delinsTGG` vs `c.[145C>T;147C>G]` — members at 145 and 147 with `c.146` unchanged between them, i.e. **separation one**. The record is about separation *zero*, and says so: `general.md:34` "governs members separated by ONE OR MORE nucleotides … nothing about members separated by none". The `145_147` class is the exemplar of the neighbouring record `delins-codon-carve-out-gap-one`, which already cites it.

Rather than just dropping the association, I would add the class the record is actually about. It is sitting in the governing clause: `substitution.md:32`'s `LRG_199t1:c.79_80delinsTT` against the `c.[79G>T;80C>T]` it renders `class="invalid"`. I checked — there is **no separation-zero equivalence class in the fixture at all**, so this ruling currently has no `equivalence_classes` entry that demonstrates its own shape.

### 7. 🔸 The module header's two-and-two tally does not match the four tests

**`tests/it/cis_confluence_adjudication.rs:50-53`** · *agent lens*

> Two of them are *adjudicated-correct* … and two pin a disagreement that is deliberately unresolved.

Only `two_adjacent_members_that_both_consume_reference_are_one_delins` is adjudicated-correct. Test 1 pins a demonstration under a record whose status is `undecided`, and test 3's own doc says "Ferro's output here is not adjudicated right; it is adjudicated *out of scope*." The PR body gets this right with three categories — adjudicated-correct, scope carve-out, pinned open disagreement. Worth aligning because the repository `CLAUDE.md` is emphatic that the kinds are not interchangeable and asks each record to state which it is; the header currently states the wrong count of a distinction the module exists to hold.

---

### Also noted, not blocking

- **`classify_gaps` folds a negative gap into `all-0` / `has-0`** (`:359-361`). *CLI [7].* A bucket named `all-0` covers "every gap is zero **or overlapping**". Almost certainly an empty population — overlapping members are rejected upstream — but it is one more name that asserts more than it checks, and this is the file where that matters.
- **`DEFAULT_CORPUS` is repo-relative rather than `CARGO_MANIFEST_DIR`-based** (`:78`). *CLI [13].* Works, since cargo sets the CWD to the package root. Same disposition I would give it on #1540, and worth recalling that the layout has been bitten by the *inverse* problem (a stale baked `CARGO_MANIFEST_DIR` after a worktree rename), so neither form is unambiguously safer.
- **Two unrelated hunks re-encode `—` as a literal em dash** in two pre-existing `note` strings at the top of the overrides file. Content-identical, but it is a drive-by rewrite of prose this PR does not own, and it will conflict with anything else editing those lines.

### Rejected, with reasons

- **Make `--stats` and `--out` mutually exclusive in clap.** *CLI [6].* They are not in conflict: `--stats` returns after the census, `--out` writes the records. Ordering them is a preference, not a defect.
- **Return only `MockProvider` from `reference_for`.** *CLI [3], trivial.* Cosmetic; the tuple's second element is deliberately named `_reference` at the call site.
- **Assert `PAD_OFFSET % 4 == 0` in `padded`.** *CLI [11].* `PAD_OFFSET` is a `const` set to 256 in the same file. A guard on a compile-time constant that cannot vary buys nothing.
- **Replace the `--direction` string with a `ValueEnum`.** *CLI [12], trivial.* Real style improvement, but the manual match already produces a clear error and this is a dev-only example.

---

### Interop: #1544 and #1549 conflict, and neither PR's CI can see it

Measured rather than assumed, since these three PRs all edit `RULING_STATUSES` and the same `rulings` array:

```
git merge-tree --write-tree --name-only origin/test/gate5-spec-worked-examples origin/research/confluence-class-adjudication
CONFLICT (content): tests/fixtures/grammar/hgvs_spec_normalization_overrides.json
CONFLICT (content): tests/it/hgvs_spec_normalization_tests.rs
```

| pair | conflicts |
|---|---:|
| #1544 × #1548 | 0 |
| #1548 × #1549 | 0 |
| **#1544 × #1549** | **2 files** |

All three are clean against `main` individually, so this only appears once one of them lands. #1548 is unaffected because it *modifies* existing records rather than appending. Whichever of #1544/#1549 merges second needs a hand-resolution — and note the overrides file is **committed**, so this is not the regeneratable-fixture case; do not resolve it by regenerating anything.

---

**Spec audit: clean.** The reading is careful in the place it would be easiest to overreach: `substitution.md:32` and not `:16` is named governing, on the stated grounds that `:16` defines a type while `:32` adjudicates the shape — and `:32` does contain the `class="invalid"` verdict on `c.[79G>T;80C>T]` verbatim. Naming `general.md:56` as deviated-from rather than distinguishing it away is the honest disposition, and the observation that `delins` is absent from `:56`'s five-type ranking is correct and is the same gap #1548 flags as still open. The scope clause is genuinely load-bearing and the record does not quietly widen it: the `dup`/`ins` carve-out is argued from `duplication.md:18`, and the third shape found in re-measurement (a repeat expansion flush against a neighbour) is named as uncontemplated rather than folded in. `Representation-Change: none` is accurate — no file under `src/` is touched, and I confirmed the suite passes with nothing re-blessed.

Of the seven confirmed findings, **1 through 4 are the ones I would act on before merge**: one turns on three tests that have never executed, and the other three are all ways the instrument can under-report the population whose emptiness the decided record's scope claim rests on.

PR #1551, opened 2026-08-08
feat(conformance): refuse to write an artifact built from a partial generator pass

### LINKED ISSUE #1550: Generators can write a complete-looking artifact from an incomplete run, and nothing fails
Generators can drop work silently and still write. The dropped population is never counted, printed, or stamped into the artifact, so a partial run and a clean run look identical — the only thing separating them is whether a reviewer read the right five lines. #1540 has now built the fix for its own generator; this issue is about making that the contract rather than one PR's good judgement.

## A live instance on `main`

`examples/report_conformance_reference_gaps.rs:161-165` has a `#[cfg(test)] mod tests` with a `#[test]`. Its `[[example]]` entry (`Cargo.toml:333-334`) omits `test = true`, and cargo does not build an example target's tests unless it opts in:

```
$ cargo nextest run --features dev -E 'binary(report_conformance_reference_gaps)'
Summary [0.000s] 0 tests run: 0 passed, 0 skipped
```

`Cargo.toml:291-293` already states the rule — *"Cargo does not build a target's tests unless it opts in, so without this they silently never run"* — and six other entries set it. A committed test that has never executed is worse than no test, because it reads as coverage.

## The pattern, verified at each PR's current tip

| where | state |
|---|---|
| `examples/report_conformance_reference_gaps.rs` (`main`) | a `#[test]` that has never run |
| #1549 `dump_confluence_divergences.rs:512` | `spans_of(...).unwrap_or_default()` → `classify_gaps([])` → `"no-gaps"`, so an output whose spans failed leaves the `has-0` population — the population a `decided` record's scope claim rests on |
| #1549 `:476-490` | dropped spellings, then `entry.1 &= grouped.len() <= 1` records a class reduced to one survivor as **converged**; never counted, never printed |
| #1549 `Cargo.toml` | three `#[test]`s never run — one pinning arithmetic that *"has already produced published numbers that were wrong"* |
| #1545 `extract_case_harvest_windows.rs:192-194` | `serve_transcripts_whole` `continue`s past `sequence: None`, skipping the widening for exactly the transcript that had a problem; no refusal before write |
| #1544 | lacks the stored-length-agreement assertion **both** #1540 and #1545 carry, so a truncated slice serves wrong content through reads that all succeed |
| #1456 / #1460 / #1478 (`CLAUDE.md`) | three successive corpus blindnesses, each invisible to the one before, each surfacing as a `0 moved` that read as safety |

One shape throughout: **a fallible step whose failure is representable as a legitimate value** — `unwrap_or_default()`, `else { continue }`, a discarded `Result`, an empty vec classifying as "no gaps".

**#1540 is the counter-example and the proof the fix works.** Its first revision discarded a `catch_unwind` result and only *printed* its capture count. At its current tip `run_input` returns `Result` and propagates the panic, and the generator refuses to write — *"Refusing to write a fixture built from a partial pass"* — with an assertion replacing the printed count. That is the contract; it just lives in one file.

## Why neither more review nor a blanket lint is the answer

Review does catch these — at four CodeRabbit runs and hours of reading per round, and it will miss the next one, because the tell is local while the consequence is three files away. Four PRs in this queue each documented this family in a sibling's generator while shipping an instance in their own; the missing ingredient is not attention.

The instinctive remedy — grep `examples/` and `src/conformance/` for the idioms — measures badly on `main`:

| idiom | `examples/` | `src/conformance/` |
|---|---:|---:|
| `unwrap_or_default()` | 7 | 0 |
| `.ok()?` | 13 | 0 |
| `unwrap_or(` | 16 | 3 |
| `let _ =` | 38 | 17 |
| `continue;` | 49 | 2 |

~145 candidate sites, mostly legitimate. What makes `spans_of(...).unwrap_or_default()` a defect is not the idiom but that the failure flows into a classification that is then **counted** — a composition grep cannot see. Noisy gates get allowlisted into irrelevance, so the mechanical half must be narrow and the semantic half carried by a type.

## Proposed

1. **`src/conformance/completeness.rs` — a `CaptureLedger`.** Attempted / succeeded / dropped; `finish()` returns `Result` and refuses on any shortfall unless an allowance is declared at the call site; counts serialized into the artifact so a fixture carries its own completeness claim. This is #1540's refusal promoted to the shared home — the same move #1545 makes for `audited_provider.rs`.
2. **A `test = true` invariant test.** Every `examples/*.rs` with `#[cfg(test)]` must set it. Mechanical, no false positives, and it closes the `main` instance above.
3. **An artifact-writer ratchet.** Generators writing a committed artifact route through the ledger; current non-users are allowlisted against this issue and the list shrinks as PRs land.
4. **Docs** in `CLAUDE.md` and `CONTRIBUTING.md`, attached to the existing *"A corpus zero is a claim about the corpus, not about the change"* section — the doctrinal half of the same point, currently advisory.

## Out of scope

No generator migrations: #1544, #1545 and #1549 own the files named above, and migrating them here would conflict with three open PRs and mix mechanical churn into a new contract. No blanket idiom lint, for the measured reason above.

## Done when

The invariant test is red on `main` before the fix and green after; `CaptureLedger` exists with tests that actually execute and refuses on shortfall; the allowlist is committed pointing here; and a generator that drops work and writes anyway fails the build instead of producing a review finding.

nh13:
Closes #1550

**This change moves no normalized output.** Nothing under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched, no generator is migrated, and no artifact is regenerated — it adds a library type, three invariant tests, `Cargo.toml` flags and documentation. Declared as `Representation-Change: none` in the footer below.

## The contract

Generators under `examples/` build the corpora, fixtures and tables the rest of the suite adjudicates against, and they share one failure mode: **a fallible step whose failure is representable as a legitimate value**. `spans_of(..).unwrap_or_default()` yields an empty vec that classifies as "no gaps"; `else { continue }` skips exactly the record that had a problem; a discarded `Result` reports nothing at all. Nothing counts the dropped population, so a partial run and a clean run write indistinguishable artifacts and the only thing separating them is whether a reviewer read the right five lines.

This PR makes that a contract rather than one file's good judgement.

### 1. `src/conformance/completeness.rs` — `CaptureLedger`

`examples/extract_split_member_separation_windows.rs` already implements this refusal for itself — *"Refusing to write a fixture built from a partial pass"*. Promoted into a shared type:

- **Attempted / succeeded / dropped, with a reason per drop.** `attempted == succeeded + dropped` holds by construction; there is no API to increment one alone. (It is *not* re-checked by `Deserialize`, which is how a consuming test reads a stamped claim — `CaptureCounts::is_self_consistent` exists for that side.)
- **`finish()` returns `Result` and refuses on any shortfall.** A `Result` the caller must handle before it can reach its `fs::write`, rather than a count printed to someone who already believes the run worked. `CaptureCounts` deliberately has **no `Default`**, so `finish().unwrap_or_default()` — the module's own archetypal defect, applied to the type meant to kill it — does not compile; a `compile_fail` doctest pins that.
- **A shortfall is waived only by naming it.** `Allowance::at_most(n, justification)` is constructed at the call site and the justification is required — and required means non-empty, not merely present.
- **An empty pass is a shortfall of its own**, needing `Allowance::permitting_an_empty_pass` to waive — the corpus-zero hazard, where a structural blindness is indistinguishable from a safe result.
- **`CaptureCounts` is serde-serializable**, so a generator *can* stamp its completeness claim into the artifact and a consuming test can assert on it. Nothing enforces that it does; stamping is a convention, and the docs now say so.

`ledger.record(id, fallible())` is a drop-in for `let Ok(v) = fallible() else { continue };` — same control flow, but the drop is now accounted for. That is the intended adoption path.

**What the ledger does not claim, stated in its own module docs:** it accounts for *the step you route through it*, not for the artifact. `succeeded` is never compared against the rows that reach the file, so a pass with a second unaccounted fallible step can stamp `{"attempted":10,"succeeded":10,"dropped":0}` onto an artifact holding five rows — the exact shape #1550 cites for `dump_confluence_divergences.rs`. Route the *last* fallible step before the write, not the first.

**Tests:** fourteen unit tests in the module plus two doctests (one of them `compile_fail`). All of them execute (`conformance::completeness::tests::*`). They were verified to be discriminating rather than merely present: with `close()` stubbed to the "report a count and write anyway" behaviour the module exists to replace, **7 of the 11** original tests fail; with the real implementation, all pass.

### 2. The `test = true` invariant — red on `main`, green here

`tests/it/generator_completeness.rs::examples_with_unit_tests_opt_into_running_them`.

Cargo does not build an example's or a binary's `#[cfg(test)]` code unless the target sets `test = true`. `Cargo.toml` has stated this in prose since the two spec generators became `[[bin]]`s, and most generator targets already set the flag — but nothing checked it, so `examples/report_conformance_reference_gaps.rs` shipped a `#[cfg(test)] mod tests` that had **never once executed**. Before:

```
$ cargo nextest run --features dev -E 'binary(report_conformance_reference_gaps)'
    Starting 0 tests across 0 binaries (13 binaries skipped)
     Summary [   0.000s] 0 tests run: 0 passed, 0 skipped
error: no tests to run
```

The new test, written first, is red on that tree and names the culprit:

```
every generator target carrying #[cfg(test)] must set `test = true` in Cargo.toml —
cargo does not build a target's tests unless it opts in:
  [[example]] `report_conformance_reference_gaps` compiles #[cfg(test)] code in
  examples/report_conformance_reference_gaps.rs but does not set `test = true`,
  so those tests never run
```

Adding the flag turns the invariant green **and** brings the dead test to life. It passes on its first-ever execution.

So there is no pre-existing bug hiding behind it — the guard it had claimed to provide was simply never provided.

**This is the one guard here that is exact.** The manifest is parsed as TOML rather than substring-matched, because `Cargo.toml` discusses `test = true` at length in comments and a `[[bench]]` setting it would be a false positive. `#[path]` includes are resolved transitively, because `test = true` is a property of the *target*: a `#[cfg(test)]` in a shared `examples/common/` module is equally dead in every target that includes it without opting in. The one boundary is a plain `mod helper;` include, which is not resolved — noted in the code; the repo uses `#[path]` throughout.

**Two further live instances turned up during review, three days apart.** Rebasing onto the tip that landed #1544/#1545, the invariant went red on `extract_spec_worked_example_windows`, which `#[path]`-includes `examples/common/recording.rs` and did not set `test = true` — six `recording::tests::*` units dead *in that target* while its siblings ran them. Rebasing again onto the tip that landed #1549, it went red on `extract_split_member_separation_windows` for the same reason. `test = true` added in both; all pass on their first execution. That is the transitive-include half of the check earning its keep twice, on generators that landed after the issue was filed.

A sibling test, `every_generator_source_is_declared_as_a_cargo_target`, covers the whole-file version of the same drop — `autoexamples = false`, so an undeclared generator is never built at all, which would also make the guard above vacuous for that file.

### 3. The artifact-writer ratchet — a heuristic floor, and it says so

`tests/it/generator_completeness.rs::artifact_writers_use_the_capture_ledger_or_are_allowlisted`.

A generator whose sources *look like* they write a file must either look like they route their population through `CaptureLedger`, or name itself in `LEDGER_EXEMPT`, each row annotated with `#1550`. Nineteen current writers are listed — fifteen from the tree the issue was filed against, plus the four that landed while this was in review (#1544, #1545, and two from #1549). Migrating those belongs to their own generators' changes, not here.

**Both sides are substring scans, and the code, the test's own failure messages and both policy documents now say so.** An earlier revision of this PR described the ratchet as making it so "a new generator cannot be added without one of the two deliberate acts". That claim was **false and is withdrawn** — it was demonstrated false in review by adding an artifact writer that used `OpenOptions` and by adding one whose entire `CaptureLedger` "usage" was a comment. What the guard actually buys is that a generator written the way every generator in this repo is currently written cannot be added in *silence*, and that the question gets asked. The semantic guarantee is carried by the type and by review.

Two hardenings came out of that, neither of which pretends to close the gap:

- The write-idiom list gained `OpenOptions::new(` and `from_path(` — two measured misses — and is documented as non-exhaustive, with the rule for what belongs on it ("the call names a path") stated so it can be extended consistently.
- The ledger predicate now requires an import of `conformance::completeness` **and** a `finish` call, **in the generator's own entry point**. That closes a demonstrated laundering path: one comment mentioning `CaptureLedger` appended to the shared `examples/common/recording.rs` marked *three* unrelated generators as migrated, and the shrink-only half then *required* deleting their rows — removing three generators from the contract on the strength of one line in a file they merely share. `recording.rs` is exactly what the #1544/#1545 migrations will touch.

**Shrink-only, in both directions.** A row whose generator has since adopted the ledger, stopped writing, or been deleted **fails the test**, so the list cannot rot into a blanket exemption — a migration PR is told to delete its own row. And a generator that looks like it writes and is neither ledgered nor listed fails the build. The "now uses the ledger" direction is the one that can *lose* a row, so its failure message says the row *looks* stale and to confirm before deleting.

Adding a row is a legitimate answer, in exactly the way `Representation-Change: none` is. What neither check tolerates is silence.

### 4. Docs

`CLAUDE.md` and `CONTRIBUTING.md` both state the rule, attached to the existing **"a corpus zero is a claim about the corpus, not about the change"** doctrine, of which this is the mechanized half: that rule covers a misleading number a generator *reports*, this one covers a population it silently drops.

Both are written to be **true of the code as written**, which is the whole point of a policy document in this repo. They name which of the three follow-on rules is enforced by a test (one: `test = true`) and which two are conventions (refuse-don't-report, backed by the type; stamp-the-counts, backed by nothing), and they describe the ratchet as a heuristic floor rather than a proof. An earlier revision claimed "each is enforced by a test"; shipping that in *this* PR would have been self-refuting.

## Deliberately out of scope

- **No generator is migrated to the ledger.** #1544, #1545 and #1549 own `dump_confluence_divergences.rs`, `extract_case_harvest_windows.rs`, `extract_spec_worked_example_windows.rs` and `extract_split_member_separation_windows.rs`. Migrating them here would conflict with open PRs and mix mechanical churn into a new contract. That is what the allowlist is for, and why it shrinks as those PRs land.
- **No AST-based usage analysis.** The alternative to narrowing the ratchet's claim was to make the predicate real — parse Rust, resolve the ledger's actual usage. That was considered and rejected: chasing a perfect predicate for a three-line rule is how you overclaim a second time, and the honest answer to "a grep cannot tell whether a population was routed through the ledger" is to say so, not to build a bigger grep.
- **No blanket idiom lint.** Re-measured on this branch across `examples/` + `src/conformance/`: `unwrap_or_default()` 17, `.ok()?` 18, `unwrap_or(` 22, `let _ =` 74, `continue;` 62 — ~193 candidate sites, overwhelmingly legitimate. What makes one a defect is not the idiom but that its failure flows into a value that is then counted, a composition grep cannot see. Noisy gates get allowlisted into irrelevance, so the mechanical half stays narrow and the semantic half is carried by the type.

## Notes for review

- `toml` joins `[dev-dependencies]`. It is already an optional runtime dependency (`web-service`), but `tests/it` compiles without `--features dev`, so the optional copy is not guaranteed to be enabled; cargo unifies the two for test targets.
- `Shortfall::Dropped` boxes its payload to stay under clippy's `result_large_err` threshold, since it is the `Err` half of every `finish`.
- The refusal message caps its *reason* list as well as its subject list. `record()` sets `reason = error.to_string()`, which for an `io::Error` is unique per record, so an uncapped reason map defeated `MAX_NAMED_SUBJECTS` entirely — measured at 2000 drops, a 2,014-line / 71 KB panic. Both totals stay exact in the truncated form.

---

Closes #1550

Representation-Change: none

## Summary by CodeRabbit

* **New Features**
  * Added completeness tracking for generated records, including successes, dropped records, drop reasons, allowances, and serialized capture counts.
  * Added clear errors when generation captures nothing or drops records without an approved allowance.

* **Tests**
  * Added automated checks to verify generators account for dropped records and use correct test-target configuration.

* **Documentation**
  * Added guidance for implementing completeness tracking, handling justified exceptions, and validating generated artifacts.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -413,6 +413,87 @@ is **spec-explicit > Mutalyzer > our judgement**. Where the spec is explicit and
 that is a deliberate divergence and it gets a record saying so — otherwise the next person
 measures Mutalyzer, finds a mismatch, and "fixes" a conformance decision.
 
+### A generator must account for what it dropped
+
+This is the mechanized half of **"a corpus zero is a claim about the corpus, not about the
+change"** — the doctrine that a generator's `0` reads as *"this is safe"* when it usually means
+*"the corpus could not build the thing you changed"*. That doctrine covers a zero the generator
+*reports*. The same trap one level down is a population the generator **silently drops**: a
+fallible step whose failure is representable as a legitimate value —
+`spans_of(..).unwrap_or_default()` classifying as "no gaps", `else { continue }` skipping exactly
+the record that had a problem, a discarded `Result`. Nothing counts it, so a partial run and a
+clean run write indistinguishable artifacts and the only thing separating them is whether a
+reviewer read the right five lines.
+
+**Route the population through `CaptureLedger`** (`src/conformance/completeness.rs`, #1550):
+
+```rust
+use ferro_hgvs::conformance::completeness::{Allowance, CaptureLedger};
+
+let mut ledger = CaptureLedger::new("transcript windows");
+for accession in &accessions {
+    // Replaces `let Ok(t) = resolve(a) else { continue };` — same control flow,
+    // but the drop is now accounted for.
+    // `record`'s subject is `impl Into<String>`, which `&String` does not
+    // satisfy and does not auto-deref to — hence `as_str()`.
+    let Some(transcript) = ledger.record(accession.as_str(), resolve(accession)) else {
+        continue;
+    };
+    capture(transcript);
+}
+// `finish` refuses on any shortfall, and on a pass that attempted nothing.
+// Waiving one means naming it: `finish_with(Allowance::at_most(1, "why"))`.
+let counts = ledger.finish()?;
+```
+
+Three things follow. **Only one of them is enforced by a test; the other two are conventions, and
+saying otherwise here would be the exact defect this section is about** — a claim that reads as
+coverage while providing none:
+
+- **A generator carrying `#[cfg(test)]` must set `test = true` on its cargo target.** Enforced by
+  `it::generator_completeness::examples_with_unit_tests_opt_into_running_them`, whose two halves
+  are not equally strong: the manifest is *parsed* as TOML, so `test = true` is read rather than
+  guessed, but `#[cfg(test)]` is found by scanning the compiled sources for that literal token —
+  which matches it inside a doc comment and misses `#[cfg(all(test, feature = "dev"))]`. Say
+  "parsed manifest, token scan", not "exact". Cargo does not build a target's tests unless it
+  opts in, so without the flag they never run — and a committed test that has never executed is
+  worse than none, because it reads as coverage. `report_conformance_reference_gaps` shipped one
+  for months.
+- **Refuse, do not report.** A convention, backed by the *type* rather than by a guard over your
+  code. `finish()` returns a `Result` you must handle before reaching your `fs::write`, and
+  `CaptureCounts` has no `Default`, so `finish().unwrap_or_default()` does not compile. Nothing
+  stops `let _ = ledger.finish();`.
+- **Stamp the counts into the artifact.** A convention, enforced by **nothing**. `CaptureCounts`
+  is serde-serializable so a fixture *can* carry its own completeness claim and a consuming test
+  *can* assert on it — no test anywhere checks that any artifact does.
+
+`tests/it/generator_completeness.rs` holds the `test = true` invariant above and a second guard:
+a generator that looks like it writes an artifact must look like it routes its population through
+the ledger, or name itself in `LEDGER_EXEMPT`. **Read that second one as a floor, not a proof.**
+Both of its sides are substring scans over source text — `fs::write(`, `File::create(`,
+`OpenOptions::new(`, `from_path(` on one side; on the other, a `use` line naming
+`conformance::completeness`, the word `CaptureLedger`, and a `finish` call, all three in the
+generator's own entry point. A writer using an idiom outside that list is invisible to it, and
+"routes its population through" is a semantic question a grep cannot answer. What it does buy: a
+generator written the way every generator in this repo is currently written cannot be added in
+silence, and the question gets asked. The semantic guarantee is carried by the ledger and by
+review.
+
+The allowlist is **shrink-only** — a row whose generator has been deleted, stopped matching the
+write idioms, or now imports the ledger and calls `finish` fails the test — so it cannot rot into
+a blanket exemption. The "now uses the ledger" direction is deliberately hard to trip by accident,
+because acting on it *deletes* a row: an earlier revision keyed on the bare word `CaptureLedger`
+anywhere in the compiled sources, and one comment appended to the shared
+`examples/common/recording.rs` marked three unrelated generators as migrated and demanded their
+rows be removed. Adding a row is a legitimate answer, in the same way
+`Representation-Change: none` is; what neither check tolerates is silence.
+
+**And note what the ledger itself does not claim.** It accounts for the step you route through
+it, not for the artifact: `succeeded` is never compared against the rows that reach the file, so a
+pass with a second unaccounted fallible step can stamp
+`{"attempted":10,"succeeded":10,"dropped":0}` onto an artifact holding five rows. Route the
+*last* fallible step before the write, not the first.
+
 ### Python Tests
 ```bash
 pytest tests/python/ -v              # Run all Python tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -198,6 +198,64 @@ test count the script prints is a weaker signal than the manifest check.
 FERRO_MANIFEST=/path/to/manifest.json scripts/run_conformance_axis.sh
 ```
 
+### Adding or changing a generator under `examples/`
+
+Generators build the corpora, fixtures and tables the rest of the suite adjudicates
+against, and they share one failure mode: a fallible step whose failure is
+representable as a legitimate value — `unwrap_or_default()`, `else { continue }`, a
+discarded `Result`. The dropped population is never counted, so a partial run and a
+clean run write indistinguishable artifacts. This is the same trap as the
+"a corpus zero is a claim about the corpus, not about the change" rule in `CLAUDE.md`,
+one level down: there the misleading number is *reported*, here it is never computed.
+
+Two rules, with two very different amounts of machinery behind them —
+`tests/it/generator_completeness.rs` (#1550) has a guard for each, and **neither guard is
+exact end to end**. Rule 2 parses the manifest but only token-scans the source; rule 1 is a
+substring heuristic on both sides. It matters that you know which half you are leaning on:
+
+1. **A generator that writes an artifact routes its population through
+   `CaptureLedger`** (`src/conformance/completeness.rs`). Record a success or a drop
+   for every record — `ledger.record(id, fallible())` is a drop-in for
+   `let Ok(v) = fallible() else { continue };` — and call `finish()` before writing.
+   It returns a `Result` and refuses on any shortfall, including a pass that attempted
+   nothing. An expected shortfall is waived by naming it:
+   `finish_with(Allowance::at_most(2, "bare LRG_ ids have no versioned index entry"))`.
+   The counts are serializable, so stamp them into the artifact and let the consuming
+   test assert on them.
+
+   Route the **last** fallible step before the write, not the first. The ledger accounts
+   for the step you hand it and nothing else — it never sees the artifact, so a second
+   unaccounted `else { continue }` downstream will happily let you stamp
+   `{"attempted":10,"succeeded":10,"dropped":0}` onto a file holding five rows.
+
+   If you are not adopting the ledger, add your generator to `LEDGER_EXEMPT` in that
+   test with a reason. That is a legitimate answer — what the check rejects is silence.
+   The list is shrink-only: a row that is no longer needed fails the test.
+
+   **The guard behind this rule is a heuristic floor.** It asks whether your source
+   contains one of a few write idioms (`fs::write(`, `File::create(`,
+   `OpenOptions::new(`, `from_path(`) and, if so, whether your entry point carries a
+   `use` line naming `conformance::completeness`, the word `CaptureLedger`, and a
+   `finish` call — all three in that one file. Both sides are substring scans: write a
+   file some other way and the ratchet never notices, import the ledger through a
+   fully-qualified path or a rustfmt-wrapped `use` group and it reads as un-adopted,
+   and "routes its population through" is not something a grep can check. So treat a green test as "the
+   question got asked", not as "this generator is accounted for" — the accounting is
+   yours and your reviewer's to get right. Nothing checks that any artifact actually
+   carries its `CaptureCounts`, either; stamping them is a convention.
+
+2. **A generator carrying `#[cfg(test)]` sets `test = true` on its `[[example]]` or
+   `[[bin]]` entry in `Cargo.toml`.** Cargo does not build a target's tests unless it
+   opts in, so without it they silently never run — which reads as coverage while
+   providing none. **The manifest side is exact and the source side is a token
+   scan**: `test = true` is read out of parsed TOML, but `#[cfg(test)]` is
+   detected by looking for that literal token, so gating your tests as
+   `#[cfg(all(test, feature = "dev"))]` slips past it. `#[path]` includes are
+   followed transitively (`test = true` is a per-target flag, so a `#[cfg(test)]`
+   in a shared `examples/common/` module is dead in every target that does not opt
+   in). A plain `mod helper;` include is the one boundary — the repo uses `#[path]`
+   throughout.
+
 ## Code Style
 
 - Format with `cargo fmt`
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -237,6 +237,15 @@ tempfile = "3.14"
 rstest = "0.23"
 serde_yaml = "0.9"
 anyhow = "1"
+# Used by tests/it/generator_completeness.rs to read this manifest's own
+# `[[example]]`/`[[bin]]` tables. The invariant it enforces is a property OF
+# Cargo.toml (`test = true` on every generator carrying `#[cfg(test)]`), so it
+# has to parse the manifest rather than substring-match it — a `test = true`
+# inside a comment, or under a `[[bench]]`, must not count. `toml` is already an
+# optional runtime dependency (`web-service`); it is repeated here because
+# `tests/it` compiles without `--features dev`, so the optional copy is not
+# guaranteed to be enabled. Cargo unifies the two for test targets.
+toml = "0.8"
 # Used by examples/build_conformance_snapshot.rs to record a per-entry SHA-256 of
 # the harvested bases (snapshot provenance). Examples have access to
 # dev-dependencies, so this stays out of the runtime dependency tree.
@@ -311,6 +320,12 @@ test = true
 [[example]]
 name = "extract_spec_worked_example_windows"
 required-features = ["dev"]
+# `#[path]`-includes `examples/common/recording.rs`, which carries `#[cfg(test)]`
+# units; `test = true` is a per-target flag, so without it those units are dead
+# in *this* target however many siblings enable them. Caught by
+# `it::generator_completeness::examples_with_unit_tests_opt_into_running_them`
+# (#1550), which is why the invariant resolves `#[path]` includes transitively.
+test = true
 
 [[example]]
 name = "derive_tx_placements"
@@ -332,9 +347,13 @@ name = "extract_case_harvest_windows"
 required-features = ["dev"]
 test = true
 
+# `test = true` because the target `#[path]`-includes `examples/common/recording.rs`,
+# whose `#[cfg(test)]` units are dead in any target that does not opt in — `test = true`
+# is a per-target flag, not a per-file one.
 [[example]]
 name = "extract_split_member_separation_windows"
 required-features = ["dev"]
+test = true
 
 [[example]]
 name = "generate_conformance_summary"
@@ -347,6 +366,11 @@ required-features = ["dev"]
 [[example]]
 name = "report_conformance_reference_gaps"
 required-features = ["dev"]
+# Carries `#[cfg(test)]` unit tests; without this cargo never builds them.
+# Pinned by `it::generator_completeness::examples_with_unit_tests_opt_into_running_them`
+# (#1550), which is what finally caught this one after it shipped a `#[test]`
+# that had never executed.
+test = true
 
 [[example]]
 name = "dump_partitions"
--- a/src/conformance/completeness.rs
+++ b/src/conformance/completeness.rs
@@ -1,0 +1,914 @@
+//! Completeness accounting for corpus and fixture generators (#1550).
+//!
+//! Every generator under `examples/` walks a population — corpus rows, spec
+//! strings, transcripts, reference accessions — and writes an artifact derived
+//! from it. They share one failure mode: **a fallible step whose failure is
+//! representable as a legitimate value**. `spans_of(..).unwrap_or_default()`
+//! yields an empty vec that classifies as "no gaps"; `else { continue }` skips
+//! exactly the record that had a problem; a discarded `Result` reports nothing
+//! at all. The dropped population is never counted, so a partial run and a clean
+//! run produce indistinguishable output and the only thing separating them is
+//! whether a reviewer read the right five lines.
+//!
+//! A grep cannot find these. Measured on `main`, the idioms above occur at ~145
+//! sites across `examples/` and `src/conformance/`, overwhelmingly legitimately;
+//! what makes one a defect is not the idiom but that its failure flows into a
+//! value that is then *counted*, which is a composition no lint can see. So the
+//! mechanical half stays narrow (`tests/it/generator_completeness.rs`) and the
+//! semantic half is carried by a type.
+//!
+//! [`CaptureLedger`] is that type. It is
+//! `examples/extract_split_member_separation_windows.rs`'s refusal — *"Refusing
+//! to write a fixture built from a partial pass"* — promoted out of one file:
+//!
+//! ```
+//! use ferro_hgvs::conformance::completeness::CaptureLedger;
+//!
+//! let mut ledger = CaptureLedger::new("transcript windows");
+//! for accession in ["NM_000088.3", "NM_003002.2"] {
+//!     let resolved: Result<&str, String> = Ok(accession);
+//!     ledger.record(accession, resolved);
+//! }
+//! // `finish` is the gate: it refuses on any shortfall, and the counts it
+//! // returns are serializable so the artifact carries its own claim.
+//! let counts = ledger.finish().expect("a complete pass");
+//! assert_eq!(counts.attempted, 2);
+//! assert_eq!(counts.dropped, 0);
+//! ```
+//!
+//! Three properties are deliberate:
+//!
+//! - **`finish` refuses rather than reports.** A printed count is read by a
+//!   human who already believes the run worked. `finish` returns a `Result` the
+//!   caller must handle before it can reach its `fs::write`.
+//! - **A shortfall is waived only by naming it.** [`Allowance`] takes a cap and
+//!   a justification, and is constructed at the call site, so an expected drop
+//!   is a declaration in the source rather than an absence.
+//! - **An empty pass is a shortfall too.** A generator that attempted nothing
+//!   writes an artifact whose emptiness reads as "there was nothing to find" —
+//!   the corpus-zero hazard, where a structural blindness is indistinguishable
+//!   from a safe result. Waiving it needs
+//!   [`Allowance::permitting_an_empty_pass`].
+//!
+//! # What the ledger accounts for, and what it does not
+//!
+//! **It accounts for the step you route through it — not for the artifact.**
+//! `succeeded` counts records that got past *one* fallible step. It is never
+//! compared against the number of rows that actually reach the file, and the
+//! type has no way to make that comparison: it never sees the artifact. So a
+//! generator can use this exactly as prescribed and still write a partial
+//! artifact carrying a spotless claim:
+//!
+//! ```ignore
+//! let mut ledger = CaptureLedger::new("rows");
+//! for row in rows {
+//!     let Some(v) = ledger.record(row.id(), parse(row)) else { continue };
+//!     let Ok(span) = spans_of(v) else { continue };  // second step, unaccounted
+//!     artifact.push(span);
+//! }
+//! let counts = ledger.finish()?;   // {"attempted":10,"succeeded":10,"dropped":0}
+//! ```                              // …stamped onto an artifact holding 5 rows
+//!
+//! That is the *exact* shape #1550 cites for `dump_confluence_divergences.rs`
+//! (`spans_of(..)` failing downstream of a successful capture), so it is worth
+//! being blunt about: **route the last fallible step before the artifact, not
+//! the first.** If a pass has several fallible steps, either route them all
+//! through one ledger — `record_drop` at each — or give each its own ledger and
+//! `finish` each. A claim scoped to one loop is still worth stamping; it is just
+//! a claim about that loop, and an artifact that says so is honest where one
+//! that implies more is not.
+//!
+//! The two mechanical guards in `tests/it/generator_completeness.rs` are
+//! narrower still: substring heuristics over generator source text, which catch
+//! the common case and prove nothing. See that file's module docs.
+
+use std::collections::BTreeMap;
+use std::fmt;
+
+use serde::{Deserialize, Serialize};
+
+/// One dropped record: what was being processed, and why it did not make it.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct DroppedRecord {
+    /// The thing that was being captured — an accession, an input string, a row
+    /// id. Named rather than counted so a failure report says *which*.
+    pub subject: String,
+    /// Why it was dropped. Grouped verbatim in [`CaptureCounts::dropped_by_reason`],
+    /// so prefer a small closed vocabulary over an interpolated error string.
+    ///
+    /// [`CaptureLedger::record`] does not, and cannot — it has only the error to
+    /// go on. See its docs for when to reach for
+    /// [`CaptureLedger::record_drop`] instead.
+    pub reason: String,
+}
+
+/// A declared allowance for a shortfall, constructed at the call site.
+///
+/// The justification is required rather than optional: an allowance without one
+/// is indistinguishable from having never considered the question, which is the
+/// state this whole module exists to make impossible.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Allowance {
+    max_dropped: usize,
+    empty_pass_permitted: bool,
+    justification: String,
+}
+
+impl Allowance {
+    /// Permit up to `max_dropped` drops, for the stated reason.
+    ///
+    /// `Allowance::at_most(0, ..)` is meaningful and is *not* the same as
+    /// [`CaptureLedger::finish`]: it still refuses a drop, but it records in the
+    /// artifact that the question was asked and answered.
+    ///
+    /// # Panics
+    ///
+    /// If `justification` is empty or only whitespace. "Required" was previously
+    /// presence rather than content, so `Allowance::at_most(0, "")` was accepted —
+    /// an allowance carrying no reason is exactly the "indistinguishable from
+    /// having never considered the question" state the type exists to prevent.
+    pub fn at_most(max_dropped: usize, justification: impl Into<String>) -> Self {
+        let justification = justification.into();
+        assert!(
+            !justification.trim().is_empty(),
+            "an Allowance needs a justification: an empty one is indistinguishable \
+             from having never considered the question",
+        );
+        Self {
+            max_dropped,
+            empty_pass_permitted: false,
+            justification,
+        }
+    }
+
+    /// Also permit a pass that attempted nothing.
+    ///
+    /// Needed only by generators whose population is legitimately optional (an
+    /// absent input file, a corpus filtered to empty by a CLI flag). For anything
+    /// that ships a committed artifact, an empty pass is the corpus-zero hazard
+    /// and should stay a refusal.
+    #[must_use]
+    pub fn permitting_an_empty_pass(mut self) -> Self {
+        self.empty_pass_permitted = true;
+        self
+    }
+}
+
+/// The allowance as stamped into an artifact, so a reader of the fixture sees
+/// the waiver rather than having to find the generator that granted it.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct AllowanceNote {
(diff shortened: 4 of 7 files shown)


nh13:
@coderabbitai pause

nh13:
## Independent review of #1551

This is an independent review by a reviewer who was given **no prior analysis** of this PR — no earlier review, no summary, no author notes beyond the PR body and #1550 themselves. Every claim in the description was treated as a claim to verify. Reviewed at `cc99c1fb` in a detached worktree.

Bottom line: **the mechanism is real and I verified the load-bearing claims reproduce exactly.** The three findings below are all about the *reach* of the guards, not their existence — and one of them (the docs) matters more than usual because this PR's own thesis is that a guard which reads as coverage while providing none is worse than no guard.

---

### 1. What I verified independently (things that could have failed and didn't)

| claim | how I checked | result |
|---|---|---|
| "red on `main`, green here" | `git show origin/main:Cargo.toml` — neither `report_conformance_reference_gaps` nor `extract_spec_worked_example_windows` sets `test = true` | **confirmed** |
| the invariant names the culprit | removed each flag in turn, re-ran the built `it` binary | **confirmed**, with the right target named each time |
| the transitive `#[path]` walk earns its keep | removing `test = true` from `extract_spec_worked_example_windows` reports `compiles #[cfg(test)] code in examples/common/recording.rs` — i.e. it attributes to the *included* file, not the entry point | **confirmed** |
| the revived tests pass on first execution | `-E 'binary(report_conformance_reference_gaps) or binary(extract_spec_worked_example_windows)'` | **7 passed** (1 + 6 `recording::tests::*`), matching the body |
| "with `close()` stubbed to report-and-write, 7 of 11 fail" | replaced `close` with a count-printing `Ok(counts)` and re-ran | **exactly 4 passed / 7 failed** |
| shrink-only, deletion half | moved `examples/generate_perf_tables.rs` off disk | **red**: `LEDGER_EXEMPT names generators that no longer exist` |
| shrink-only, "now ledgered" half | appended a `CaptureLedger` mention to an exempt generator | **red**: `now uses CaptureLedger` (see finding 1 for the sting in the tail) |
| every `LEDGER_EXEMPT` row is live | the stale check passes, so all 17 rows genuinely write | **confirmed**, 17 rows = 17 writers |
| the doctest actually runs | `cargo test --doc completeness` | `1 passed` |

Gates: `cargo fmt --all --check` → exit 0. `cargo clippy --all-features --all-targets -- -D warnings` → **exit 0, no diagnostics**. `cargo nextest run --features dev -E 'binary(it) and test(generator_completeness)'` → 3/3 pass. `-E 'test(conformance::completeness)'` → 11/11 pass.

---

### 2. Confirmed findings, by severity

#### [major] The ratchet is satisfied by the *word* `CaptureLedger`, including in a comment

`tests/it/generator_completeness.rs:320-324` — `uses_the_ledger` is `text.contains("CaptureLedger")`. Demonstrated with a brand-new declared artifact writer:

```rust
// examples/zz_writer.rs
// Population accounting: see CaptureLedger (not actually used).
fn main() { std::fs::write("out.json", "{}").unwrap(); }
```

```
test artifact_writers_use_the_capture_ledger_or_are_allowlisted ... ok
test every_generator_source_is_declared_as_a_cargo_target ... ok
test examples_with_unit_tests_opt_into_running_them ... ok
```

All three green. A new artifact-writing generator was added with **neither** of the two deliberate acts the PR says are the only options.

The near-term sting is the transitive half. Appending one comment mentioning `CaptureLedger` to `examples/common/recording.rs` produces:

```
LEDGER_EXEMPT is shrink-only, and these rows are no longer needed — delete them:
  examples/extract_case_harvest_windows.rs: now uses CaptureLedger
  examples/extract_spec_worked_example_windows.rs: now uses CaptureLedger
  examples/extract_spec_enumeration_windows.rs: now uses CaptureLedger
```

`recording.rs` is exactly the shared module the #1544/#1545 migrations will touch. The shrink-only half then *requires* deleting all three rows — after which those three generators are permanently outside the ratchet on the strength of one mention in a file they share, whether or not their own populations were ever accounted. The mechanism designed to prevent rot has a path that launders three generators out of the contract in a single line.

Cheapest fix that keeps the grep honest: require the mention in a code-ish position (`use ferro_hgvs::conformance::completeness::` plus a `.finish(` / `.finish_with(` in the *same* file), and attribute `uses_the_ledger` to the entry point only rather than transitively. Both are one-liners; neither needs `syn`.

#### [major] `writes_an_artifact` misses ordinary write idioms, so the "cannot be added without a deliberate act" claim is false

`tests/it/generator_completeness.rs:314-318` — `fs::write(` or `File::create(`, substring. Same probe generator, three bodies:

| write mechanism | ratchet |
|---|---|
| `std::fs::write(..)` | **red** (correct) |
| `OpenOptions::new().write(true).create(true).truncate(true).open(..)` + `write_all` | **green — missed** |
| `csv::Writer::from_path(..)` | **green — missed** |

Both misses are things a generator author would plausibly write (append mode, streaming a large table). The PR body says "an artifact writer that is neither ledgered nor listed fails the build, so a *new* generator cannot be added without one of the two deliberate acts" — that is not what the predicate does. I'd either widen the predicate (`.open(`, `from_path(`, `to_writer(`, `BufWriter::new(`) or, better, soften the claim: this is a **heuristic ratchet**, and saying so in the const's doc comment costs nothing and prevents the next reader trusting it further than it goes. CodeRabbit independently flagged the same predicate.

Related, smaller: the ratchet iterates `generator_sources_on_disk()` (top-level `examples/*.rs`) while the `test = true` invariant iterates `generator_targets()` (declared paths). A declared `[[example]]` with `path = "examples/sub/gen.rs"` is therefore covered by the first guard and invisible to the second. No such target exists today.

#### [major] The ledger accounts for one *step*; nothing ties it to the artifact

This is the honest answer to "does `CaptureLedger` make the failure mode impossible?" — it makes drops at the *instrumented* step impossible to hide, and that is genuinely worth having. But a generator can use the ledger exactly as the docs prescribe and still write a half-empty artifact carrying a spotless claim:

```rust
let mut ledger = CaptureLedger::new("rows");
for row in 0..10u32 {
    let Some(v) = ledger.record(format!("row-{row}"), parse(row)) else { continue };
    let Ok(span) = spans_of(v) else { continue };   // second fallible step, unaccounted
    artifact.push(span);
}
let counts = ledger.finish().expect("clean");
```

Run:

```
(a) claim = {"attempted":10,"succeeded":10,"dropped":0}
(a) artifact rows actually written = 5
```

`{"attempted":10,"succeeded":10,"dropped":0}` stamped onto an artifact containing 5 of 10 rows is worse than no claim, by this PR's own argument. Note this is the *exact* shape #1550 cites for `dump_confluence_divergences.rs` (`spans_of(...)` failing downstream of a successful capture) — the ledger as designed does not close it, because `succeeded` and "rows in the artifact" are never compared.

Two things would help a lot and are cheap: (a) say this explicitly in the module docs — the ledger accounts for *the step you route through it*, so route the last fallible step, not the first; (b) offer `finish_matching(&self, written: usize)` that refuses when `succeeded != written`, which turns the accounting into a statement about the artifact rather than about one loop.

#### [major] `ledger.finish().unwrap_or_default()` compiles clean and fabricates a clean claim

`src/conformance/completeness.rs:129` derives `Default` on `CaptureCounts`. Because `Result::unwrap_or_default` needs only `T: Default`, this compiles with no warning (verified with `cargo check`):

```rust
let counts = ledger.finish().unwrap_or_default();   // Shortfall discarded
```

The result is `{"attempted":0,"succeeded":0,"dropped":0}` — which serializes as a *tidier* claim than a real one and, to a human diffing a fixture, reads as clean. The module's opening paragraph names `unwrap_or_default()` as the archetypal defect; the type it introduces is directly susceptible to it. `let _ = ledger.finish();` is also silent, but that one is unavoidable in Rust; this one is closeable by dropping the `Default` derive (the only in-repo use is the `..Default::default()` in one unit test at :524, easily rewritten).

#### [medium] `record()` contradicts the module's own reason guidance, and defeats `MAX_NAMED_SUBJECTS`

`DroppedRecord::reason` (:64-67) and `record_drop` (:290-291) both say to keep `reason` a small closed vocabulary and "prefer a small closed vocabulary over an interpolated error string" — because it is the grouping key for `dropped_by_reason`. But `record()` (:305-320), which the module doc, `CLAUDE.md` and `CONTRIBUTING.md` all present as *the* adoption path, sets `reason = error.to_string()`. For any `io::Error`/`anyhow::Error` that is unique per record, so `dropped_by_reason` degenerates to one bucket per drop.

That composes badly with `Display` (:242-251), which truncates the *named subjects* at `MAX_NAMED_SUBJECTS = 10` but prints `dropped_by_reason` in full. Measured, 2000 drops via `record()`:

```
(b) refusal message lines = 2014
(b) refusal message bytes = 71514
```

`MAX_NAMED_SUBJECTS`'s doc says it exists to avoid "a 20,000-line panic". The documented adoption path reintroduces it. Fix either end: cap the reason map in `Display` too, or give `record` a `reason` label distinct from the error detail (`record_labelled(subject, "resolve failed", outcome)`).

#### [medium] The docs claim enforcement the tests do not provide

This is the one I'd not merge without, because `CLAUDE.md` is read by future agents as binding policy, and the claim is of exactly the kind this PR exists to eliminate.

- `CLAUDE.md:434` — "Three things follow, and **each is enforced by a test** rather than left to review". Of the three bullets: *"Stamp the counts into the artifact"* is enforced by **nothing** — no test anywhere asserts that any artifact carries `CaptureCounts`. *"Refuse, do not report"* is enforced by the type's own unit tests, and is defeatable per the `unwrap_or_default` finding above. Only the third bullet (`test = true`) has a guard.
- `CONTRIBUTING.md:184-186` — "Two rules, **both enforced by** `tests/it/generator_completeness.rs`: 1. A generator that writes an artifact **routes its population through** `CaptureLedger`". Findings 1 and 2 show what is actually enforced is "writes via one of two substrings ⟹ mentions a substring, or is allowlisted". "Routes its population through" is not checked and cannot be by a grep — which the PR body itself says, correctly, two paragraphs later.

Suggested wording that stays true: say the mechanical guard is a *ratchet on new writers*, that it is a heuristic on both sides, and that the semantic guarantee is carried by the type and by review — which is the PR's real and defensible position.

#### [low, and ironic] `compiled_sources` silently drops an unreadable source

`tests/it/generator_completeness.rs:148`:

```rust
let Ok(text) = std::fs::read_to_string(&absolute) else { continue; };
```

`else { continue }` on a fallible step, in the file whose module doc names `else { continue }` as the failure mode. If an entry point is unreadable, `compiled_sources` returns empty, so the target reads as carrying no `#[cfg(test)]` and writing nothing — **both** guards pass silently for it. Low practical risk (paths come from cargo or from `read_dir`), but it should `panic!` with the path and error. CodeRabbit found this too and its suggested patch is right.

---

### 3. Noted, not blocking

- **`test = true` count is off by one.** `tests/it/generator_completeness.rs:166` and the PR body both say "six targets already set the flag". On this PR's merge base it is **seven**: `harvest_multi_member_cis`, `generate_cis_confluence_corpus`, `extract_spec_enumeration_windows`, `dump_normalized_corpus`, `extract_case_harvest_windows`, and the two spec `[[bin]]`s. Six was right when #1550 was filed, before #1545 landed. Worth fixing in the committed comment.
- **A `#[cfg(test)]` reached by a plain `mod helper;`** — cargo's own `examples/<target>/helper.rs` convention — is invisible to the `test = true` invariant. Verified green with a probe. No live instance; the repo uses `#[path]` throughout. A one-line note on `path_includes` would record the boundary.
- **Fifteen of seventeen `LEDGER_EXEMPT` reasons are `"#1550: pre-existing; <what the generator does>"`** — a description, not a justification for declining. That is precisely the distinction `Allowance` insists on for its own waivers ("an allowance without one is indistinguishable from having never considered the question"). The allowlist holds itself to a lower bar than the type does.
- **`Allowance::at_most(0, "")` is accepted.** "Required" is presence, not content.
- **`dump_normalized_corpus.rs` is exempt** — the harness whose three successive structural blindnesses (#1456/#1460/#1478) *are* the doctrine this PR mechanizes. The scoping rationale is sound and I am not asking to change it, but it is worth stating plainly that the contract lands with zero coverage of its own motivating file, and the value is entirely prospective until the migrations land.
- **`LEDGER_EXEMPT` keys are `/`-separated literals compared against `Path::display()`** — Windows-only concern, CI is Linux.
- **CodeRabbit's `CaptureCounts` deserialization finding is real but modest.** `attempted == succeeded + dropped` holds by construction and *not* through `Deserialize`, and deserialization is exactly how "a consuming test asserts on the stamped claim" happens. A `validate()` or a wire type would close it. Low, because the inputs are repo-generated.

---

### 4. Findings I rejected

- **CodeRabbit, `CONTRIBUTING.md` [trivial]: "add a compile-checked doctest".** Rejected — the module already carries one and it runs (`cargo test --doc completeness` → `1 passed`). `CONTRIBUTING.md` prose is not doctested anywhere in this repo, and making it so is out of scope.
- **"`attempted == succeeded + dropped` is not enforced" as a constructing-API defect.** Rejected — `attempted()` is *derived* (`succeeded + dropped.len()`), so it cannot be violated through the public API. The claim in the PR body is true; it is just definitional rather than a checked property, and only the deserialization side (above) is a real hole.
- **Anything clippy-shaped** (e.g. the `Shortfall::Dropped` boxing rationale). `--all-features --all-targets -D warnings` is clean at this tip; the boxing is doing its job.
- **CLI findings out of diff scope.** None — all four CodeRabbit findings landed on files this PR actually touches (`git diff origin/main...HEAD --stat`: 7 files, +1143).

---

### 5. Lenses

- **Reviewer lens (independent reading + break-it experiments):** ran to completion. Eight deliberate mutations, each reverted immediately; all findings above are reproduced outputs, not inference.
- **CodeRabbit CLI (`coderabbit review --agent --base origin/main`):** ran to **completion** — `{"type":"complete","status":"review_completed","findings":4}` with exactly 4 `"type":"finding"` lines and **no** `rate_limit`. Findings: 1 major on `completeness.rs` (deserialization validation), 2 major on `generator_completeness.rs` (substring predicates; silent `continue` on unreadable source), 1 trivial on `CONTRIBUTING.md`. Two of the three majors independently match findings I reached separately, which raises my confidence in both.

Worktree left clean — `git status --porcelain` returns empty. No commits, no pushes, no branch modifications.

nh13:
@coderabbitai review

nh13:
## Response to the independent review

Thank you — all seven findings reproduced, and the framing (*"the reach of the guards, not their existence"*) is the right one. Rebased onto current `main` (`eb75d389`) and addressed below.

### The direction taken: narrow the claim, not strengthen the checker

There were two coherent responses to findings 1 and 2: make the predicates real (parse Rust, resolve actual usage of the ledger), or keep the cheap predicates and stop asserting they enforce the contract.

**This PR takes the second.** The substring predicates were kept **deliberately**. Nothing here parses Rust, and no attempt is made to make the write-idiom list exhaustive. The reason is this PR's own thesis: a guard that *looks* stronger than it is, is worse than a weak guard that says so. Chasing a perfect predicate for a three-line rule is how you overclaim a second time — and finding 6 is the review's own demonstration that the overclaim, not the gap, is what does the damage. So the changes below are almost entirely about making the code, the names, the failure messages and the two policy documents tell the truth about what these checks detect.

Where a hardening was cheap **and** did not require pretending the gap was closed, it was taken (the two measured write idioms; the entry-point restriction). Neither is presented as completeness.

---

### 1 [major] — the ratchet is satisfied by the word `CaptureLedger`, incl. in a comment

Accepted, and the laundering path is closed.

- `uses_the_ledger` → **`appears_to_route_through_the_ledger`**, and it now requires an import of `conformance::completeness` **and** a `.finish()`/`.finish_with(` call, **in the generator's own entry point only** — no longer transitively over `#[path]` includes.
- That kills the `recording.rs` path specifically: a comment (or anything else) in a shared module can no longer mark three generators as migrated, so it can no longer *demand* three rows be deleted. Losing a row is no longer something a comment in a file you merely share can cause.
- The cost of the entry-point restriction is stated in the function's docs rather than left implicit: a generator that genuinely delegates its accounting to a shared `common/` module now reads as *not* ledgered and keeps its `LEDGER_EXEMPT` row. That is the direction to err in — a row kept too long is a visible line of text; a row deleted wrongly is a generator that silently leaves the list.
- The stale-row assertion message changed from *"these rows are no longer needed — delete them"* to saying the rows **look** stale, that both checks are substring heuristics rather than usage analysis, and to confirm the generator really does account for its population before deleting — and that if it does not, the fix is the generator, not the list. `LEDGER_EXEMPT`'s doc comment now separates the two shrink directions by how much you can trust each.

I did **not** make the "now ledgered" direction non-failing. A row that has genuinely been migrated must still be removable, and a test that silently tolerates a stale row is the rot this list exists to avoid. The change is to the evidence required and to what the message tells you to do with it.

### 2 [major] — `writes_an_artifact` misses ordinary write idioms

Accepted. The claim is withdrawn; the list is widened a little and labelled non-exhaustive.

- `writes_an_artifact` → **`appears_to_write_an_artifact`**, over a named `WRITE_IDIOMS` const.
- Both measured misses are now covered: `OpenOptions::new(` and `from_path(`.
- The rule for what belongs on the list is written down — **the call names a path** — so it can be extended consistently. `BufWriter::new(` and `to_writer(` are deliberately *excluded* under that rule: they wrap a sink somebody else opened and say nothing about whether the destination is a file. (Concretely: the only `BufWriter` in `examples/` not already wrapping a `File::create` is over `stdout`, in `dump_partitions.rs`, and adding it produced exactly that false positive.)
- The docs record what is *still* missed — anything reaching a writer through a helper's return value, a trait object, or a differently-spelled constructor — and both false-negative and false-positive behaviour, with the note that the asymmetry (a false positive costs one allowlist row) is the right way round for a floor.
- **The PR body's claim is retracted in the PR body itself.** "a new generator cannot be added without one of the two deliberate acts" is gone; what replaces it is that a generator *written the way every generator in this repo is currently written* cannot be added in silence, and that the question gets asked.

On the smaller related point (`generator_sources_on_disk()` vs `generator_targets()` — a declared target with a subdirectory `path` is visible to one guard and not the other): no live instance, so it is recorded as a documented boundary on `generator_sources_on_disk` rather than fixed speculatively.

### 3 [major] — the ledger accounts for one step, not the artifact

Accepted as a **scoping** problem, and scoped in the docs.

`CaptureLedger` now carries a *"What the ledger accounts for, and what it does not"* section that states plainly it accounts for the step you route through it and never sees the artifact, reproduces your 10-attempted/5-written shape, points at `dump_confluence_divergences.rs` as the live instance of it, and gives the operative instruction: **route the last fallible step before the write, not the first**. `CLAUDE.md` and `CONTRIBUTING.md` both repeat it.

I **declined** your suggestion (b), `finish_matching(&self, written: usize)`. It is a good idea and I do not think it is wrong — it is out of scope for the narrow-the-claim direction, since it makes the ledger's guarantee bigger rather than making its current guarantee honestly stated. Worth a follow-up issue if you want it; I did not want to grow the type's surface in the same change that admits it was over-described.

### 4 [major] — `finish().unwrap_or_default()` compiles clean

Accepted and fixed properly. **The `Default` derive is gone from `CaptureCounts`**, so that line no longer compiles. Pinned by a `compile_fail` doctest on the type (`cargo test --doc completeness` → 2 passed, 1 ignored). The type's docs say why it must not come back — there is no meaningful "default completeness claim", and the empty one is the most misleading value the type can hold — and acknowledge that `let _ = ledger.finish();` remains silent because that one is unavoidable in Rust.

The only in-repo dependency was the `..Default::default()` at the old `:524`, rewritten as an explicit helper.

### 5 [medium] — the refusal message is unbounded

Accepted and fixed at the `Display` end. **`MAX_NAMED_REASONS = 10`** now caps `dropped_by_reason` the same way `MAX_NAMED_SUBJECTS` caps the subject list. Both totals stay exact in the truncated form: the header still says `N of M records were dropped`, the reason section is labelled `by reason (N distinct)`, and the truncation line reads `… and N more reasons covering M drops`. A new unit test builds 2000 unique-reason drops through `record()` and asserts the message stays bounded *and* that all three totals survive.

I fixed the `Display` end rather than adding `record_labelled`, because `record()`'s convenience is the point of it being the adoption path. Instead `record()`'s doc now states the trade explicitly — `reason` becomes `error.to_string()`, `dropped_by_reason` degenerates to one bucket per drop, and when the *shape* of a drop matters (a shortfall you expect to waive and read back out of an artifact) you should call `record_drop` with a closed-vocabulary reason and keep the varying detail in `subject`. `DroppedRecord::reason` cross-links to it.

### 6 [medium] — the docs claim enforcement the tests do not provide

Accepted without reservation — this was the right thing to hold the merge on. Both documents were rewritten so every enforcement claim is true of the code as written, and so that "convention" and "enforced" are visibly different words.

**`CLAUDE.md`**, replacing *"Three things follow, and each is enforced by a test rather than left to review"*:

> Three things follow. **Only one of them is enforced by a test; the other two are conventions, and saying otherwise here would be the exact defect this section is about** — a claim that reads as coverage while providing none:
>
> - **A generator carrying `#[cfg(test)]` must set `test = true` on its cargo target.** *Enforced, exactly*, by `it::generator_completeness::examples_with_unit_tests_opt_into_running_them`: the manifest is parsed as TOML and `#[cfg(test)]` in a compiled source is a fact about the file, so there is no heuristic in it. […]
> - **Refuse, do not report.** A convention, backed by the *type* rather than by a guard over your code. `finish()` returns a `Result` you must handle before reaching your `fs::write`, and `CaptureCounts` has no `Default`, so `finish().unwrap_or_default()` does not compile. Nothing stops `let _ = ledger.finish();`.
> - **Stamp the counts into the artifact.** A convention, enforced by **nothing**. `CaptureCounts` is serde-serializable so a fixture *can* carry its own completeness claim and a consuming test *can* assert on it — no test anywhere checks that any artifact does.

and, on the ratchet:

> **Read that second one as a floor, not a proof.** Both of its sides are substring scans over source text — `fs::write(`/`File::create(`/`OpenOptions::new(`/`from_path(` on one side, an import of `conformance::completeness` plus a `finish` call in the generator's own entry point on the other. A writer using an idiom outside that list is invisible to it, and "routes its population through" is a semantic question a grep cannot answer. What it does buy: a generator written the way every generator in this repo is currently written cannot be added in silence, and the question gets asked. The semantic guarantee is carried by the ledger and by review.

plus a paragraph recording the laundering path as the reason the "now uses the ledger" direction is deliberately hard to trip, and a closing paragraph on what the ledger itself does not claim (finding 3).

**`CONTRIBUTING.md`**, replacing *"Two rules, both enforced by `tests/it/generator_completeness.rs`"*:

> Two rules, with two very different amounts of machinery behind them — `tests/it/generator_completeness.rs` (#1550) has a guard for each, but only one of the two is exact, and it matters that you know which:

Rule 1 keeps the instruction to route the population through `CaptureLedger` — that is still the right thing to do — and gains:

> **The guard behind this rule is a heuristic floor.** It asks whether your source contains one of a few write idioms (`fs::write(`, `File::create(`, `OpenOptions::new(`, `from_path(`) and, if so, whether your entry point imports `conformance::completeness` *and* calls `finish`. Both sides are substring scans: write a file some other way and the ratchet never notices, and "routes its population through" is not something a grep can check. So treat a green test as "the question got asked", not as "this generator is accounted for" — the accounting is yours and your reviewer's to get right. Nothing checks that any artifact actually carries its `CaptureCounts`, either; stamping them is a convention.

Rule 2 is marked **"This one is exact"** and names its single boundary (a plain `mod helper;` include).

### 7 [low, ironic] — `compiled_sources` silently drops an unreadable source

Accepted. It now **panics** with the entry point, the absolute path and the underlying error, and the doc comment spells out why skipping was the worst available option: an entry point that failed to read yields no sources, and a target with no sources reads as carrying no `#[cfg(test)]` and writing nothing — so *both* guards would pass vacuously for exactly the file nobody could read.

---

### From "noted, not blocking"

- **The `test = true` count.** Fixed, but not by writing "seven" — by removing the number. It was six when #1550 was filed, seven at the merge base you reviewed, and eight after this branch rebased past #1549. A comment that rots every time a generator lands has no place in a PR about claims that stop being true, so the doc now says "most generator targets already set the flag" and records why no count is quoted. Same in the PR body.
- **`mod helper;` boundary.** Recorded on `path_includes` as a stated boundary with your finding that there is no live instance.
- **`Allowance::at_most(0, "")`.** Fixed — `at_most` now panics on an empty or whitespace-only justification, with a `#[should_panic]` test. "Required" meaning presence rather than content was the same species of defect as everything else here.
- **Deserialization does not check `attempted == succeeded + dropped`.** Added `CaptureCounts::is_self_consistent()` plus a test, and a doc paragraph saying the invariant holds by construction *through the ledger* and is not re-checked on the way back out of an artifact — which is exactly how a consuming test reads a stamped claim.
- **Fifteen of seventeen `LEDGER_EXEMPT` reasons are descriptions, not justifications.** Acknowledged and **not changed**. You are right that the list holds itself to a lower bar than `Allowance` does. But the honest justification for most of those rows is identical ("pre-existing, migrating it belongs to that generator's own change"), and rewriting seventeen — now nineteen — rows into individually-argued declines in this PR would be manufacturing prose rather than adding information. The rows that *do* have a specific reason (the four that landed during review) say so. Happy to revisit if you disagree.
- **`dump_normalized_corpus.rs` is exempt.** Agreed, unchanged, and worth restating plainly as you did: the contract lands with zero coverage of its own motivating file and its value is prospective until the migrations land.
- **`LEDGER_EXEMPT` keys vs `Path::display()` on Windows.** Unchanged; CI is Linux.

### Also from the rebase

`main` moved twice under this branch. #1549 landed `dump_confluence_divergences.rs` and `extract_split_member_separation_windows.rs`, and:

- The `test = true` invariant went **red again**, on `extract_split_member_separation_windows` — which `#[path]`-includes `examples/common/recording.rs` and did not opt in, so six `recording::tests::*` units were dead in that target while its siblings ran them. That is now the *second* generator caught by the transitive half of the check, three days after the first. Flag added; all six pass on first execution.
- The ratchet went **red** on both new generators. Both allowlisted, with #1549-specific reasons. `extract_split_member_separation_windows` is the file `CaptureLedger` was promoted *out of* — it already implements the refusal by hand — so its row says that.

### Break-it experiments, re-run

| experiment | before | after |
|---|---|---|
| new declared writer, `std::fs::write`, `CaptureLedger` only in a comment | **green** (all three tests passed — the hole) | **red**: `these generators look like they write an artifact without routing their population through …CaptureLedger` |
| new declared writer via `OpenOptions::new()…open()` + `write_all` | **green — missed** | **red** |
| new declared writer via `csv::Writer::from_path` | **green — missed** | **red** |
| laundering: append one `CaptureLedger` comment to `examples/common/recording.rs` | **red**, demanding **3** rows be deleted as stale | **green** — no row is marked stale; a mention in a shared module is no longer evidence |
| `ledger.finish().unwrap_or_default()` | compiled clean, fabricated `{"attempted":0,…}` | **does not compile**, pinned by a `compile_fail` doctest |
| 2000 unique-reason drops via `record()` | 2,014-line / 71 KB refusal | 20 lines, both totals still exact |
| unreadable source reached from an entry point | silently skipped, both guards vacuous | **panics**, naming entry point, path and error |
| shrink-only, deletion half (`generate_perf_tables.rs` moved off disk) | red | **red**, unchanged |

Every experimental edit was reverted (`git checkout HEAD -- <path>`); the tree is clean apart from the reviewed change.

### Gates

`cargo fmt --all --check` → 0. `cargo clippy --all-features --all-targets -- -D warnings` → 0, no diagnostics. `-E 'binary(it) and test(generator_completeness)'` → 3/3. `-E 'test(conformance::completeness)'` → 14/14. `cargo test --doc completeness` → 2 passed, 1 ignored. Full suite with all three oracles set → green.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #1557, opened 2026-08-08
changelog: a declining fix: commit is filed under Other instead of Fixed


nh13:
## What happens

A commit that declines — `Representation-Change: none` — is routed to **Other** regardless of its conventional type, so a bugfix that correctly declined disappears from **Fixed**.

In v0.13.1, #1533 (`fix(spec-fixture)`) and #1527 (`fix(changelog)`) are filed under Other. Once #1555 lands, the reasoned declines join them and #1547 (`fix(normalize): validate a cis-allele member's stated reference bases before a merge can consume them`) — a real bugfix — moves from Representation changes to Other rather than to Fixed. A reader scanning **Fixed** for what was repaired this cycle misses three of them.

This is a side effect of #1527: the exclusion rule has to run before the inclusion rule, and a git-cliff parser assigns exactly one group, so "not a representation change" was spelled as "Other".

## Why the obvious fix does not work

Conditioning on both the footer and the message looks like the answer:

```toml
{ footer = DECLINE, message = "^fix", group = "<!-- 5 -->Fixed" }
```

**git-cliff evaluates parser fields as OR, not AND** — measured against 2.13.1. With that rule first, every commit whose subject starts with `fix` matched it regardless of its trailer, including #1537 and #1535, whose trailers declare real moves. So this spelling silently *empties* the Representation changes section.

Two other routes were measured and rejected:

- **Omitting `group`** does not fall through to the next parser. The commit matches, keeps `group = None`, and git-cliff falls back to the raw conventional type — headings render as `fix`, `docs`, `test`, `ci`, with no ordering control.
- **A negative footer regex** (match the trailer only when the value is *not* a decline) needs a lookahead, which the `regex` crate does not have. The enumerated complement is writable but is a maintenance hazard that has to stay in step with the vocabulary in two other files.

## Options

- A `commit_preprocessors` rule that marks declining commits in the subject, message parsers keyed on the marker, and a `postprocessor` to strip it. Works within release-plz, but mangles every message on the way through.
- Upstream: an `all`/`and` semantic for multi-field parsers in git-cliff.
- Accept it and say so in the config, so the next person does not lose an afternoon to the OR semantics.

The last is what is documented today.

Issue #1556, opened 2026-08-08
changelog: the Representation-Change value never reaches the changelog, so the release checklist cannot be satisfied


nh13:
## The checklist asks for something no entry can carry

`release-plz.toml`'s `pr_body` puts this on every release PR:

> - [ ] Each entry says whether the affected inputs were previously **rejected** (no stored string, so free) or previously **accepted** (a real migration for the consumer).

No entry can. The changelog bullet is the commit **subject** and nothing else:

```
- *(normalize)* never split a delins into members on consecutive nucleotides ([#1537](...))
```

The trailer behind it — `3 rows of 500,004 move (0.0006%) — 2 respell, 1 merge toward the submitted form…` — never reaches the changelog. So a reviewer either ticks the box without checking, or opens every linked PR. `CONTRIBUTING.md` asks contributors for four facts, of which the fourth "cannot be recovered later"; today all four stop at the PR description.

## The trailer value is not reachable from the template

Measured against git-cliff 2.13.1, on this repo's own commits. Three routes, all dead:

**1. `commit.footers` is mis-parsed.** git_conventional treats any `word:` line as a footer token, so a prose body shreds it. For #1523 it produced 7 "footers", the first being `TOKEN=[Closes] SEP=[ #]` with a value containing the *entire* rest of the body — tables, headings and all. Two more had tokens like `` [`Representation-Change] `` from a backticked mention mid-sentence.

**2. `commit.body` does not contain the trailer at all.** git_conventional splits body from footers, so the trailer is gone from `body` by the time a template sees it. Splitting `commit.body` on `\nRepresentation-Change:` finds **0** occurrences for both real v0.13.1 disclosures (#1535, #1537).

**3. Splitting on the unanchored literal finds only the prose mentions.** Because of (2), the only `Representation-Change:` occurrences left in `body` are the ones authors wrote *about* the mechanism. For #1534 that extracted `` ` trailer five times.\n`CLAUDE.md` did not name it at all…`` — text from a paragraph discussing the trailer, presented as if it were the declaration. Shipping this would have put quoted prose under **Representation changes** in a released changelog.

Route 3 is the dangerous one: it produces plausible-looking output for some commits and garbage for others, and the garbage is only visible if you read the rendered section against the PRs.

## Options

[...]

PR #1558, opened 2026-08-08
fix(changelog): read a decline that gives a reason as a decline

### LINKED ISSUE #1555: changelog: a Representation-Change decline that gives a reason is grouped as a representation change
## What happens

A `Representation-Change:` trailer that **declines with a reason** — `none. <why>` — is grouped under **Representation changes** in the changelog, exactly as if it had declared a real move.

The v0.13.1 release PR (#1525) lists **10** entries in that section. By their own trailer text, **2** are real:

| PR | trailer says | correct group |
|---|---|---|
| #1537 | 3 rows of 500,004 move — 2 respell, 1 merge | Representation changes |
| #1535 | 0 rows move over 5,761,302 real expressions (structural zero, stated as such) | Representation changes |
| #1549 #1540 #1544 #1545 #1547 #1548 #1546 #1538 | `none.` followed by a paragraph of why | **not** a representation change |

So the section over-reports by 8. This is #1526 from the other side: that bug listed a decline as a change, and #1527 fixed it only for the bare `none` spelling.

## Why

`release-plz.toml`'s exclusion rule anchors the value:

```
footer = "(?i)^Representation-Change:[^\S\r\n]*(none|no|n/a|na)[^\S\r\n]*$"
```

The trailing `$` means the value must be *exactly* a decline word. Eight PRs declined **and said why on the same line**, so the exclusion never fired and the inclusion rule below it caught them. `NONE_VALUES` in `scripts/check_representation_change.py` has the same shape, so the checker also reports those eight as "Declared a representation change".

The six that wrote a bare `none` (#1533 #1531 #1532 #1534 #1527 #1523) grouped correctly. Nothing distinguishes the two sets except whether the author explained themselves — which is the behaviour we want to encourage.

`CONTRIBUTING.md` is what makes this a trap rather than a typo: it says "`none`, `no`, `n/a` and `na` all count, in any casing" and shows a bare `none`, with no hint that adding a reason silently changes the verdict. 8 of 8 authors who gave a reason hit it.

## Fix

Accept a reason after the verdict on both sides — the first token is the verdict, and `.`/`;`/`:`/`—` may introduce a reason. Keep the two halves pinned to each other, as `test_decline_vocabulary_matches_the_changelog_config` already does for the vocabulary.

## Why it blocks the release

The section is the artifact the disclosure requirement exists to produce. Shipping v0.13.1 with 10 entries when 1 change moved 3 rows is the same loss of reader trust as the empty section in v0.13.0, from the opposite direction — and it is worse in one respect: a reader who spot-checks two entries, finds both say "none", and concludes the section is noise will not read it again.

nh13:
The **Representation changes** section of the v0.13.1 changelog lists 10 entries. Two of them are representation changes. This makes the section report what it actually contains, before the release goes out.

Closes #1555. Records #1556 and #1557, which are limits of the same section that this does not fix.

## A decline that gives a reason was read as a disclosure

`release-plz.toml`'s exclusion rule anchored the trailer's value:

```
footer = "(?i)^Representation-Change:[^\S\r\n]*(none|no|n/a|na)[^\S\r\n]*$"
```

So `Representation-Change: none` was a decline and `Representation-Change: none. Tests only; nothing under a watched directory is touched.` was not. `NONE_VALUES` in `scripts/check_representation_change.py` had the same shape, so the checker also reported those as "Declared a representation change".

Measured over `v0.13.0..main` with git-cliff 2.13.1, the version release-plz runs:

| | commits |
|---|---|
| in the cycle (after release-plz's package filter drops two `.github/`-only commits) | 16 |
| declined | 14 |
| **declined with a reason on the same line, and were listed as changes** | **8** |
| genuinely declared a move | 2 |

The two real ones are #1537 (3 rows of 500,004 move) and #1535 (0 rows move over 5,761,302 real expressions, a structural zero it states as such). The eight are #1538 #1544 #1545 #1546 #1547 #1548 #1549 #1540.

This is #1526 from the opposite side — that bug listed a decline as a change, and #1527 fixed it only for the bare spelling. It also punished exactly the behaviour worth encouraging: the six authors who wrote a bare `none` were filed correctly, and the eight who explained themselves were not.

## What changed

**The verdict is the first word; a reason may follow it.** `.`, `;`, `:` and dashes introduce the reason. A comma does not, because it usually introduces a qualification that changes the answer — `none, except two rows that merge` is filed as a real change, as is `no rows move`, which has no terminator. Both near misses err toward listing a change rather than hiding one, which is the safe direction for this particular section.

**The two rules are pinned value by value.** `test_decline_vocabulary_matches_the_changelog_config` compares word lists, and that was not enough: the two rules named the same four words and still disagreed on `none. Tests only.`. `test_the_two_decline_rules_agree_value_by_value` extracts the regex from `release-plz.toml` and runs both over the same 15 values.

**The `<!-- N -->` ordering prefix is stripped from headings.** The comment on those prefixes claimed git-cliff removed them. It does not — release-plz's default body renders `### {{ group | upper_first }}` with no `striptags` — so every section since v0.13.0 carries `### <!-- 0 -->Representation changes` in the file. It is an HTML comment, which is why GitHub renders it invisibly and it survived two releases. Done with a `postprocessor` rather than a `body` override: overriding `body` replaces release-plz's default wholesale, including the compare link, the `[**breaking**]` marker and the PR links, which is a large surface to restate for a cosmetic fix. The four historical headings are cleaned up to match, since release-plz only prepends and would otherwise leave the file inconsistent.

## Verified against the real config, not a paraphrase

The rendering below comes from feeding this branch's own `commit_parsers` and `postprocessors` — read out of `release-plz.toml`, not retyped — to git-cliff 2.13.1 over `v0.13.0..HEAD`:

```
### Representation changes
- never split a delins into members on consecutive nucleotides (#1537)
- type the whole-block inversion by what it competes with (#1535)

### Other
- ... the other 16
```

Also: 45 tests in `test_representation_change_trailer.py` and 752 in `tests/python/` pass; ruff and mypy are clean; and the checker was exercised end to end on a reasoned decline (exit 0), a real disclosure (exit 0, reported as a change) and silence (exit 1).

## What this does not fix

Both are filed, and both are visible in this diff as documentation rather than as code:

- **#1556** — the trailer's text never reaches the changelog, so the release PR's checklist item asking each entry to state rejected-vs-accepted cannot be satisfied by any PR. Three extraction routes were measured and all fail: `commit.footers` is mis-parsed by git_conventional (7 bogus "footers" for one commit, the first swallowing the entire body), `commit.body` does not contain the trailer at all (0 matches for both real disclosures), and splitting on the unanchored literal returns the prose *mentions* — for #1534 it extracted a paragraph discussing the trailer, which would have shipped as if it were the declaration. The checklist now asks for that fact from the PR descriptions and says why.
- **#1557** — a declining `fix:` is filed under Other rather than Fixed. Adding `message = "^fix"` to the exclusion rule does not fix it: **git-cliff evaluates a parser's fields as OR, not AND**, so that rule matches every `fix:` commit regardless of its trailer and empties the section. Omitting `group` does not fall through either — the commit keeps `group = None` and git-cliff falls back to the raw conventional type. Both measurements are recorded in the config so the next person does not re-derive them.

Representation-Change: none. No file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched — the diff is the release configuration, the declaration checker, its tests, and documentation. No normalized output can move.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.13.0](https://github.com/fulcrumgenomics/ferro-hgvs/compare/v0.12.0...v0.13.0) - 2026-08-07
 
-### <!-- 0 -->Representation changes
+### Representation changes
 
 Normalization output is a shipped key: a downstream consumer stores read counts
 against the normalized HGVS string, so a form that moves between releases
@@ -71,7 +71,7 @@ consumer impact.
 [#1490]: https://github.com/fulcrumgenomics/ferro-hgvs/pull/1490
 [#1501]: https://github.com/fulcrumgenomics/ferro-hgvs/pull/1501
 
-### <!-- 1 -->Added
+### Added
 
 - *(dev)* add --verify-spdi to the normalized-corpus harness ([#1515](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1515))
 - *(normalize)* record provenance when reported members are coalesced ([#1486](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1486))
@@ -81,7 +81,7 @@ consumer impact.
 - *(normalize)* make the sequence-first derivation authoritative for cis alleles ([#1392](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1392))
 - *(effect)* [**breaking**] drop the dead inversion validator that contradicted the live one ([#1359](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1359))
 
-### <!-- 5 -->Fixed
+### Fixed
 
 - *(normalize)* apply a junction barrier the translate cannot express ([#1516](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1516))
 - *(spdi)* name the caller's anchor when a repeat names no tract ([#1496](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1496))
@@ -123,7 +123,7 @@ consumer impact.
 - *(spdi)* [**breaking**] give an identity edit the span it claims ([#1363](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1363))
 - *(normalize)* stop a terminal duplication respelling past the contig end ([#1365](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1365))
 
-### <!-- 7 -->Other
+### Other
 
 - *(msto)* pin the reported family's expected outputs and link its guards ([#1520](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1520))
 - *(normalize)* correct the coincidence statistics behind the density gate ([#1521](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1521))
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -310,6 +310,19 @@ those directories move no output at all (a comment, a doc, a new unit test), and
 expected case. What the check rejects is *silence*, because an absent trailer is
 indistinguishable from an unconsidered one.
 
+**The verdict is the first word, and a reason may follow it** — `none. Tests only; no watched
+file is touched.` A full stop, semicolon, colon or dash introduces the reason; a comma does not,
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai pause

Issue #1099, opened 2026-07-21
fix(project): all-synonymous cis allele drops silent codons (emits one arbitrary residue, not the full list or p.(=))


msto:
@nh13 I wanted to flag this, but seems to me like the final direction might be subjective - not sure if you want to prescribe a decision or expose an option. 

Reporting the list of silent codons when projecting from a genomic/coding representation that contains the underlying nucleic acid changes, and reporting `p.=` when normalizing a protein representation seems sensible to me, but I could also see an argument for picking just one or letting the user decide. I'll defer to you! 😁 

---

> [!NOTE]
> Authorship: the content below was drafted by GitHub Copilot CLI (an AI coding
> agent) and filed via gh under @msto's account. It reflects the agent's
> analysis, not a statement authored by @msto.

## What's wrong

For a cis allele whose members are **all synonymous** (the protein sequence is
unchanged), the projected protein consequence is **incomplete and depends on an
irrelevant detail of the input**. When one codon carries two or more in-cis
members, ferro drops every *other* affected codon and emits a single, arbitrary
residue's `=` instead of listing all silent residues or reporting whole-molecule
"no change".

Concretely, two alleles that are **both entirely synonymous** — differing only by
a protein-noop extra member in one codon — produce different protein
consequences, and the second one silently loses a codon:

```
g.[6G>A;9T>C]        -> p.[(Leu2=);(Arg3=)]   <- both silent residues listed (fine)
g.[4C>T;6G>A;9T>C]   -> p.(Leu2=)             <- Arg3= dropped; only ONE residue emitted
```

Both describe an allele whose protein product is identical to the reference. The
only difference is that the second adds `c.4C>T` — itself a **silent** change in
codon 2 (`CTG`->`TTA`, still Leu). Adding that protein-noop member causes codon 3's
consequence (`Arg3=`) to vanish from the output entirely.

The emitted residue is not *wrong* (residue 2 is indeed silent), but the result is
**lossy and arbitrary**: it reports one of two silently-changed codons and hides
the other, and *which* codon survives is driven by how members happen to group
(the codon with multiple in-cis members is kept; single-member silent codons are
dropped). This is neither of the two spec-conformant options — "list all affected
residues" or "report no change at the molecule level".

## Severity: medium

- **Silently lossy** — no error/warning; a genuinely-changed (silent) codon simply
  disappears from the protein consequence. Output looks valid.
[...]

PR #1569, opened 2026-08-08
docs(rulings): rank validity above preference, and make confluence overrides an escalation

nh13:
Answers the question the filer asked on #1099 on 2026-07-22 and that has been open since: *"how heavily does this project rely on 'spec compliance' as the primary bar for design decisions, and to what extent is the project comfortable with opinionated decisions vs deferring choices like this to the user via options?"*

**Independent of the partition-model stack** (#1565/#1566/#1567/#1568) and based directly on `main`, because it is repo-wide policy that those PRs cite rather than something they contain.

## The order

1. **Validity — what the spec REQUIRES.** Absolute; never traded, never a tiebreaker.
2. **The recommended form, delivered by a documented deterministic rule.**
3. **Disclosure** of any resulting representation move.
4. **Stability** toward the already-shipped form — last-resort tiebreaker only.

Rank 1 is exactly two things: the grammar (`docs/syntax.yaml`, rendered into every recommendation's syntax block, `style.md:54`) and the absolute prohibitions, which the spec enumerates itself in `checklist.md:5` as descriptions that *"**do not correctly follow HGVS nomenclature**"*.

## Why validity is split out — this retires a reading we had adopted

Uppercase RFC 2119 keywords appear **exactly once** in `docs/recommendations/` outside `style.md`. `CLAUDE.md` concluded from that count that *"almost nothing in the recommendations is normative"* and that these are *"house-style choices the spec leaves open"*.

That conflates *no uppercase keyword* with *no requirement*, and it licenses trading away things the spec genuinely does require. `style.md`'s own item 6, quoting RFC 2119, explains the scarcity: imperatives *"must be used with care and sparingly"* and *"MUST only be used where it is actually required for interoperation or to limit behavior which has potential for causing harm."* The near-absence is deliberate editorial restraint.

`general.md` proves the point inside a single bullet list: `:56` states a preference (the five-type prioritisation) while `:58`, nested directly under it, states a prohibition — *"are not allowed"*, with a worked counter-example.

## Confluence is not a rival to the recommended form

It is a property **of the rule**. A written, ordered, deterministic rule produces the recommended form *and* converges; they are one requirement seen from two sides. `general.md:56` is the model — though the implemented rule is not `:56` alone, since `DNA/inversion.md:19` deliberately overrides it for inverted duplications (specific beats general), and that precedence has to be written down to be understandable.

Overriding the recommended form is therefore an **escalation, not a rank**: available only where no such rule can be derived, expected to be vanishingly rare, and carrying three obligations discharged in the same change — surfaced to the operator, documented in the register, and locked by a unit test citing the clause.

`PRECEDENCE_ESCALATIONS` censuses the register, so adding an entry fails the count until it is updated deliberately. That friction is the point: escalations must not accumulate quietly into a second, undeclared precedence order.

**E1 is registered.** `delins` appears nowhere in `general.md:56`'s ranking, so a contest among `delins` members has no spec-derivable rule. Recorded rather than decided; the open product decision remains `canonical-form-choice-when-both-legal`.

## One rank removed, and one cost stated plainly

The former rank *"re-derivation from the resulting sequence where the spec is silent"* is **removed**. The partition is the unit of normalization, full stop — so re-derivation is not a method ferro uses and cannot sit in a precedence order. `partition-is-the-unit-of-normalization` and this record had contradicted each other until now; two `decided` records disagreeing is worse than either answer.

The cost, stated rather than left to be discovered: that rank is where **#1420's method request was honoured** — *"apply the edit set to the reference, then re-derive the minimal canonical partition from the resulting sequence."* Removing it means ferro now **declines** that method. The reasons already on the record still hold (the family is agent-drafted; it is 3 merge / 3 split / 5 respell rather than a direction; #1421's own rows converge on the spanning `delins` it argues against) — but this is a real change of answer to a filed request and should be communicated as one.

## Also corrected

`CLAUDE.md` listed `adjudication-precedence-order` as `undecided` with *"two competing precedence orders on record and neither is chosen"*. It has been `decided` since 2026-08-07; the doc was stale. Moved to the decided table, and the undecided count corrected from five to four.

## Verification

`generate_spec_fixture` validates every citation quote byte-for-byte against the pinned spec checkout, so a paraphrase or a moved clause fails the build. All seven `hgvs_spec_normalization_tests` guards pass, including `ruling_records_are_intact` and the new census.

Representation-Change: none. Ruling text, one new census constant and its test, and documentation; no code path is altered and no normalized output moves.


nh13:
@coderabbitai pause

nh13:
Superseded by #1605 and #1604, which split this PR along a line that turned out to matter.

**#1605** puts the normalization ruleset in `README.md` as the project's single canonical statement — validity, recommended form, confluence, determinism, and the procedure for spec gaps, disclosure and maintainer choice. That is where this PR's validity-above-preference ordering and its RFC 2119 argument now live, in a form a user can read.

**#1604** keeps what was genuinely mechanism rather than prose: the escalation register, its `PRECEDENCE_ESCALATIONS` census and the E1 entry recording that `delins` is unranked by `general.md:56`. It makes `adjudication-precedence-order` a **pointer** to the README rather than a second statement of the same order.

That pointer/statement split is the reason for closing rather than rebasing. This project's recurring failure is one rule written in several places and drifting apart, and leaving an ordering in the ledger alongside one in the README would have reproduced it directly.

**Why the rank removal here did not stand.** This PR removed the "re-derivation from the resulting sequence" rank on the grounds that `partition-is-the-unit-of-normalization` was decided. That record does not exist on `main`, is not among the records added by #1585, and `git log --all -S` finds it only on unmerged branches. Meanwhile `canonical-form-choice-when-both-legal` — which *is* decided on `main`, and was `undecided` when this PR was written — rules for re-derivation in terms: "derives the description from the RESULTING SEQUENCE … subject to every explicit spec tie-break". So removing the rank did not end a contradiction between two decided records; it created one. Shipped code agrees with the record that exists: `canonicalize_from_sequence` runs on every `normalize()`, and all four `PartitionRule` variants re-derive.

Two smaller corrections carried forward into #1604: the count of uppercase RFC 2119 keywords outside `style.md` is **two**, not one — `RNA/adjoined_transcript.md:20` (`REQUIRES`) and `:21` (`SHOULD`); the usual grep lists `REQUIRED` and so misses `REQUIRES`, and that grep is fixed too. And `canonical-form-choice-when-both-legal` is no longer "the open product decision" this PR's text described.

Nothing here is lost — the citation work was verified byte-for-byte by the fixture generator and is retained in whichever of the two PRs it belongs to.

PR #1570, opened 2026-08-08
test(normalize): adjudicate the rows the partition model moved

nh13:
Base is #1567 (`feat/spdi-equivalence-key`), not `main`. Full chain at the bottom.

Forty-six rows were red once the partition rule was made to govern every block. This PR adjudicates them: **14 go green with an authority behind each**, **31 stay red with the cause measured and recorded**, and **1** is the protein-axis gap IMPL 2 (#1568) owns.

## The one code change, and it was removing a spec-forbidden output

`general.md:35`'s coding exception is implemented **twice** in `src/normalize/merge.rs`, and only one of the two carries the codon half.

`partition_block_preserving` computes a per-pair floor — the reading-frame floor of two applies only when both members are single-base substitutions, which is the shape `general.md:35` works and the shape the decided ruling `codon-carve-out-shape-restriction` restricts ferro to — and `split_codon_incompatible_triplets` then cuts back apart any pair that crossed a codon boundary. `coalesce_coding_frame_separation` (#1484) has neither gate: it merges *any* two pieces one unchanged base apart on a length-changing reading-frame block. It runs second, so it re-merged exactly the pairs the preserving arm had just refused.

Measured on the spec's own worked example:

```
[W58]  LRG_199t1:c.[992_1002del;1004T>C]  ->  LRG_199t1:c.992_1004delinsAC
```

That merged string is the answer `consultation/SVD-WG010.md:51` proposed and the committee **rejected** in 2021. Three guards named it independently, including the negative one (`the_forbidden_description_is_never_what_ferro_emits`), which cites SVD-WG010:51 by line. A second, hermetic instance is `c.[9del;11del;13del]` -> `c.9_13delinsAT` — three single-base deletions, none of them a substitution.

This is the row **#1565 lists as its most substantive known-red item**; it is fixed here, and #1565's description now points at this PR for it.

The pass is now keyed on the **partitioner** rather than on the requested rule. That is the same correction IMPL 1 already applied to the changed-column weight bound and to the codon splitter, and this is the third instance of the one mistake.

**Nothing is lost, and the population argument is the interesting part.** The pass was written for a *lone* `delins` that the re-derivation split at a coincidentally matching interior base (`NM_000143.3:c.44_47delinsATC` -> `[44_45delinsAT]` + `[47del]`, the case its own doc names). Under the partition model that population **does not exist**: the split move is scoped to equal-length members, so an unequal-length lone `delins` stays one piece. Measured — `NM_TEST.1:n.2_5delinsAAC` is returned unchanged where it used to split. So under `preserve` every pair the pass could see is a pair the input asserted separately, which is precisely what `general.md:34` says must stay individual. It is unchanged on `Live`, `Shadow`, `Canonical` and `CanonicalCoalesced`.

Recorded as a new `decided` ruling, `coding-frame-merge-belongs-to-the-arm-that-carries-the-codon-shape-gate`, governing `general.md:34`, five quote-verified clauses, with `RULING_STATUSES` kept in sync. `cargo run --features dev --bin generate_spec_fixture` passes, so every quote is verified against the checkout.

## The 14 adjudicated rows

| rows | verdict | authority |
|---|---|---|
| W58 ×3 (`every_worked_example…`, `the_normal_forms…`, `the_forbidden_description…`) | **ferro-wrong, fixed** | `general.md:34`; `SVD-WG010:51` names the forbidden string |
| `the_separation_two_members_present_is_not_a_property_of_the_variant` | **ferro-wrong, fixed** | same cause, same fix |
| `no_codon_frame_sub_then_del_pair_straddles_codon_boundary` (#275) | **ferro-correct, test stale** | `general.md:34` + `general.md:35`. Its old pin was recorded *in its own comment* as a "policy choice" preferring confluence over the authored spelling — the policy `partition-is-the-unit-of-normalization` reversed. The row's value now matches its name. |
| `coding_length_changing_block…` + `non_coding_length_changing_block…` (folded into one row + one new one) | **ferro-correct, test stale** | `DNA/delins.md:47`. A lone unequal-length `delins` has no column-wise correspondence to cut on. |
| #1231, #1233 on `g.` and on `c.`/`r.` (4 rows) | **ferro-correct, test stale** | `partition-is-the-unit-of-normalization`. `general.md:56` ranks the type of **one member** and does not move its boundaries — an insertion at a junction consumes no reference, so the substitution pair is a different *partition*, not a different type. |
| `a_derived_whole_block_inversion_outranks_the_codon_exception` | **ferro-correct, test stale** | `DNA/inversion.md:5` (whole-span property) and `DNA/delins.md:5` (a delins is a replacement "**and which is not** a substitution or inversion"). Restores what v0.13.1 shipped. |
| `every_reported_spelling_normalizes_as_recorded_under_five_prime` (2 pins) | **ferro-correct, pins stale** | both old pins were reached through a decline into `partition_block`; the decline rate is now 0.0%. |
| `three_prime_confluence_census`, `five_prime_confluence_census` | **measured, re-blessed** | measured against both arms; see below |

Where a value moved, an `assert_ne!` naming the old string was added, so the previous value cannot be restored as "drift". Nothing was weakened: no `contains` replaced an equality, no `#[ignore]`, no lowered floor.

**Two guards were widened rather than left blind.** The `n.`/`c.` axis discriminator in `merge_consecutive_edits_tests` stopped discriminating once a lone `delins` keeps its partition on both axes — its expected string became its input. It is re-expressed on a *multi-member* block where the axis floor is what decides (`c.[4C>G;6A>G]` merges inside codon 2, `n.[4C>G;6A>G]` does not), so both halves now differ from their inputs and neither can be satisfied by a decline. Similarly `cds_axis_reduces_a_mixed_type_allele_to_substitutions` had claimed to be "the discriminating test for the axis gate"; that claim is withdrawn and the two rows that still carry it are named.

## The 31 rows left red, and why re-blessing them would have been wrong

Three exhaustive sweeps in `cis_junction_crossing_shift` report **187,776** two- and three-member cases whose output overlaps or denotes no single sequence:

```
TTTTTTTTTAATATATTTTA
  g.[2_3dup;5_6insA]  ->  g.[1_9T[11];5_6insA]   junction 5|6 inside 1_9
  g.[2dup;4dup;6T>A]  ->  g.[1_9T[11];6T>A]      base 6 inside 1_9
```

**Measured attribution.** Re-running the affected suites with `FERRO_PARTITION=live` turns **97 of 99** rows green — all three sweeps, and every one of `issue_1135`, `issue_1261`, `issue_1281`, `issue_1287`, `issue_1292`, `issue_1296`, `issue_1301`, `issue_1304`, `issue_1320`, `issue_1327`, `issue_1398`, `issue_1437`, `issue_1453`, `cis_spelling_confluence_gap`, `cis_allele_confluence_proptest`, `rewrite_target_corpus` and `a_dup_flush_against_a_del_is_left_alone`. The two that stay red under `live` are the two whose pins were deliberately moved to the `preserve` values. So this is a property of the partition-preserving arm, not a pre-existing defect the branch merely exposed.

**The mechanism is left `cause-unverified` on purpose.** Two candidate gates were read and neither was measured on a concrete triple: the five sibling-repair passes in `normalize/mod.rs` all return early on `before.len() != after.len()`, which a licensed merge now makes common; and `demote_repeats_spanning_siblings` keys on the *pre-normalization member's edit type* while under `preserve` the snapshot it is handed is the canonicalizer's output rather than the authored member. Naming a cause without the triple is the error this branch's design document records three times.

**The three sweeps are fixed one PR up, in #1571**, which takes the state-based route on the typing stage: where typing widened a member's rendered extent past its asserted span, and the asserted span did not already span a sibling, the growth is re-spelled at the asserted member's junction.

These are **not value pins**. Each asserts, through the SPDI applier in `common::cis_apply_oracle` — independent of the normalizer by construction — that the output denotes the sequence the input denoted. An overlapping allele denotes *no* sequence, so there is no competing clause and nothing to adjudicate: it is a defect, not a reading. The full record, with the measured table and the A/B, is in `cis_junction_crossing_shift.rs`'s module doc.

The remaining red row, `spec_worked_example_rules::the_protein_axis_must_split_a_separation_one_delins`, fails under both arms and is the deferred protein-axis split move (IMPL 2, #1568). Its own doc says not to pin it; unchanged here.

## A corpus zero that is a claim about the corpus

`cis_confluence_axis`'s census reports `sequence_changed: 0` and `not_fixed_point: 0` over 47,392 spellings **on the same tree** that produces the 187,776 overlapping cases above. The designed corpus cannot build a junction-only member sitting strictly inside a homopolymer tract another member grows. That caveat is now written next to the zeros rather than left for the next reader to discover.

## Precedence

Every row above is decided at rank 1 or rank 2. `DNA/delins.md:5` (inversion typing) and the SVD-WG010:51 prohibition are rank-1 validity constraints; `general.md:34`/`:35`/`:41`/`DNA/delins.md:47` are the recommended form delivered by the partition rule, which is documented, deterministic and ordered. **No row here was decided by preferring confluence to the recommended form**, and none needed the escalation. The one row whose residue is genuinely unrankable — `a_dup_flush_against_a_del_is_left_alone`, where the competing forms are a `dup` against a flush `del` and `general.md:56` ranks type labels for one span rather than a four-member allele against a spanning description — is recorded as reducing to `canonical-form-choice-when-both-legal`, which stays `undecided`. It is not decided here. `adjudication-precedence-order` is cited by id only; its record is untouched.

## Verification — measured before the 2026-08-09 restack

```
cargo check --features dev --all-targets          OK
cargo fmt --all --check                           OK
cargo clippy --all-features --all-targets         OK (no warnings from this crate)
cargo run --features dev --bin generate_spec_fixture   OK (every ruling quote verified)
cargo nextest run --release --features dev        9765 run, 9733 passed, 32 failed, 149 skipped
```

Failure delta at that time: **46 -> 32**, zero rows newly red, with the red set byte-identical before and after the then-current rebase onto `feat/spdi-equivalence-key`. **That parent has since moved** — the restack rebased this branch onto `ea2c1709` — so the absolute counts are due a re-run. The 46 -> 32 delta and the "zero newly red" claim are what to re-establish, not the number 32.

Representation-Change: 14 pinned expectations move, 6 split / 4 merge-reversal / 4 respell-or-shift. Direction is toward the spec and toward v0.13.1, not away: 3 rows stop emitting `LRG_199t1:c.992_1004delinsAC`, the answer SVD-WG010:51 proposed and the committee rejected; `c.[12C>G;14del]`, `c.[9del;11del;13del]` and `n.2_5delinsAAC` return to the form `general.md:34` / `DNA/delins.md:47` state; `c.[1G>T;3T>A;5A>C]` returns to `c.1_5inv`, which is exactly what v0.13.1 shipped. Four rows are the partition ruling's own declared change and are net-new divergences for a stored library: #1231 and #1233 no longer converge their `[dup;del]` / `[ins;del]` spellings onto a substitution pair on `g.`, `c.` or `r.` — #1233 is the reporter's highest-frequency shape, so it is the costliest single row. Two 5' pins in `reported_partition_verdicts` move to the authored partition, which removes #1420-v2's accidental convergence on its issue's wanted form. At corpus scale the underlying rule change measures 8,026 -> 3,416 converged classes at 3' and 8,021 -> 3,503 at 5' over the **11,272 designed cis classes** of `cis_confluence_axis` (`FERRO_PARTITION=live` against the shipped `preserve`), with `declined`, `sequence_changed` and `not_fixed_point` all 0 on both arms; that figure belongs to IMPL 1 (#1565) and is re-stated here only because this PR is what re-blesses the census that carries it. Mind the denominator when comparing: #1565's trailer quotes the same switch over #1572's **11,882-family spec corpus** (9,140 -> 3,509 at 3', 8,944 -> 3,747 at 5'), a different population, and the two sets of numbers must not be quoted interchangeably or summed.


nh13:
@coderabbitai pause

nh13:
Closing. Every part of this PR turns out to be a property of the partition-preserving arm rather than of ferro, and that arm was withdrawn on 2026-08-09 with #1565. Measured rather than read — the detail is below, because "adjudicate the rows the model moved" is exactly the kind of PR whose content *looks* portable.

## What was checked, and how

The 10 test files were applied to `main` (`7118eb2a`) with `src/normalize/merge.rs` and the ledger pair excluded. **Five of the ten do not apply at all** — they are layered on #1565's own re-blesses of the same files, so they were never independent of the model. The five that do apply run **100 passed, 5 failed**.

| row | `main` emits | this PR expects |
|---|---|---|
| `issue_1235_transcript_axes::cds_axis_keeps_a_mixed_type_alleles_partition` | `c.[5A>T;7T>A]` | `c.[4_5insT;8del]` |
| `…::rna_axis_keeps_a_mixed_type_alleles_partition_and_the_u_alphabet` | `r.[5a>u;7u>a]` | `r.[4_5insu;8del]` |
| `issue_275_codon_frame_extensions::no_codon_frame_sub_then_del_pair_straddles_codon_boundary` | `c.12_13delinsG` | `c.[12C>G;14del]` |
| `merge_consecutive_edits_tests::coding_length_changing_block_merges_across_one_unchanged_base` | `n.[2_3delinsAA;9del]` | `n.2_5delinsAAC` |
| `msto_regression_corpus::every_cataloged_test_exists_in_its_source_file` | — | catalog bookkeeping for a renamed test |

The first two are the input's asserted partition being handed back, which is the model itself. The third is territory open PR #1599 already owns (`gate the coding delins carve-out on the amino-acid precondition`), and it touches the same file.

## The three "pure additions" are not portable either

Each documents a defect **of the preserving arm**, in its own words:

- `cis_confluence_adjudication.rs` (+23) marks a row RED and records that it is "**green under `FERRO_PARTITION=live`**" — i.e. green on `main`. Confirmed: it passes here.
- `cis_junction_crossing_shift.rs` (+52) opens an "OPEN DEFECT" section whose measured attribution is that `FERRO_PARTITION=live` "turns **97 of 99** rows green, including all three sweeps".
- `hgvs_spec_normalization_tests.rs` (+17) is the `RULING_STATUSES` registration for the record below.

The ruling record `coding-frame-merge-belongs-to-the-arm-that-carries-the-codon-shape-gate` states its own thesis as "run `coalesce_coding_frame_separation` only where the pieces were NOT produced by `partition_block_preserving`" — a rule with no referent on `main`. #1585 dropped it deliberately for the same reason.

The `src/normalize/merge.rs` hunk (+48/−6) gates that pass on `!preserved`, which is a no-op without the arm. Its justification is that W58 comes back as `LRG_199t1:c.992_1004delinsAC`, the form `consultation/SVD-WG010.md:51` proposed and the committee rejected in 2021 — **that does not happen on `main`**. The negative guard naming that exact string, `the_forbidden_description_is_never_what_ferro_emits`, passes on `main` (it ships in #1606).

## One row was worth keeping, and it is filed as #1610

`n.2_5delinsAAC` → `n.[2_3delinsAA;9del]` is a genuine `main` defect, and this PR's expectation for it is right on the merits. Its *stated* reason was the withdrawn model, but a second argument in the same doc comment is model-independent and survives: `DNA/delins.md:46` names the decomposition as the *alternative* description for exactly this alignment-coincidence shape and `:47` recommends the span, under the **decided** record `delins-merge-vs-individual-gap-two-or-more`.

Filed as #1610 with the reproducer, the clause argument, and — the part most likely to be got wrong by a fix — why the **equal-length** path must not be touched, since there `delins.md:17` requires the split and ferro is already correct.

Nothing else here survives the model's withdrawal.

Issue #1004, opened 2026-07-09
normalize non-idempotent for same-gap insertion pair adjacent to a substitution (pre-existing)


nh13:
### Summary

`normalize` is non-idempotent for a cis allele containing **two insertions at the same reference gap** plus an adjacent substitution. The two same-gap insertions are merged into a single insertion on the first pass; a second normalization of that output then sees a lone insertion adjacent to the substitution and collapses it into a `delins`, so `normalize(normalize(x)) != normalize(x)`.

This is **pre-existing** (reproduces on `main` at `11e3962`, independent of #999/#1000) and is a distinct corner from the single-insertion shift-created adjacency fixed in #999/#1000. The `has_same_gap_insertions` guard added in that fix deliberately preserves this behavior (single-pass) rather than changing it — see #1002.

### Reproducers (synthetic `MockProvider`, 3'-shift)

Reference `NC_000001.11`: filler `A`, with `CATCCTCGCTCCT` starting at pos 300, and `TGACTTCAGTCACCTGACTGACTG` at pos 600 for the second case.

```
g.[305_306insC;305_306insA;307G>T]
  pass1: g.[306_307insAC;307G>T]
  pass2: g.307delinsACT          # != pass1  -> non-idempotent

g.[611_612insCG;611_612insGC;613C>A]
  pass1: g.[612_613insGGCC;613C>A]
  pass2: g.613delinsGGCCA        # != pass1  -> non-idempotent
```

The mirror-ordered input `g.[305_306insA;305_306insC;307G>T]` *is* idempotent (`g.[305_306insAC;307G>T]`), because the merged insertion does not 3'-shift flush against the substitution — so the defect is specific to same-gap insertion pairs whose merged form shifts into adjacency.

### Notes / open question

Two insertions at the same gap are order-ambiguous per the #487 handling (`[g insA; g insC]` describes `...AC...` or `...CA...`). The current behavior silently merges them into one ordered insertion; whether that ordering is itself well-defined is worth revisiting alongside the idempotency fix. At minimum, `normalize` should reach a fixed point in one pass for these inputs (the #1000 invariant), either by collapsing on the same pass or by refusing to merge the ambiguous pair.

PR #1571, opened 2026-08-09
fix(normalize): typing may not widen a member past its asserted span

nh13:
Base is #1570 (`test/partition-model-adjudications`), not `main`. Full chain at the bottom.

`TEMPLATE:g.[2_3dup;5_6insA]` on a nine-`T` tract normalized to `TEMPLATE:g.[1_9T[11];5_6insA]`, whose repeat tract contains the sibling insertion's `5|6` junction. The pair claims one interbase point twice, so it **denotes no single sequence**. Ten such outputs in the two-member sweep of 46,080.

This is the defect #1565 pins the mechanism for and #1570 leaves red on purpose (`cause-unverified`, three exhaustive sweeps, 187,776 cases). This PR is the fix, and the three sweeps go green here.

## The mechanism, measured rather than reasoned

The partition model fixes each member's boundaries from the input's asserted partition and reads its bases from the derived sequence — which is what makes sequence preservation *structural* rather than checked. But a **third stage** runs after partitioning: **typing**, which decides whether `(span, payload)` renders as `dup`, `ins`, `delins`, `inv` or a repeat, and which may **widen** the rendered extent. So the partition fixed the boundary and typing moved it back, asking no licence.

Caught in the act. Under a 5′ shuffle the allele reaches `canonicalize_from_sequence` a second time, and the partition-preserving re-derivation hands the member back as `before = g.1delinsTTT` — which names **no duplicated or deleted length**. `demote_repeats_spanning_siblings` is *differential*: it re-spells a repeat back to the edit it grew out of. With no length to read it declined, and the widened tract shipped.

This adds the state-based route for exactly that case: where typing widened a member's rendered extent past its asserted span, **and** the asserted span did not already span a sibling, the repeat's own stated growth is re-spelled as an insertion at the **asserted member's** junction.

## Two boundaries the repair deliberately does not cross

- **An authored overlap is a different question with a settled, opposite answer.** Ferro reports the conflict and hands the description back rather than picking an order the spec does not rank (#395, #486, #1004). `g.[263_265A[7];264_265insC]` arrives already spanning its sibling — `before == after`, nothing widened it. Both gates must hold.
- **Re-spelling at the *tract's* 3′ junction was measured and rejected.** It turns `g.[2_3dup;5_6insA]` into `g.[5_6insA;8_9dup]`: disjoint, warning-free, re-parses — and **denotes a different sequence**. It would have *raised* the convergence ratchet while corrupting data, which is the sharpest available reminder that convergence going up is not on its own evidence of an improvement.

## The convergence ratchet is turned back, 376 -> 374

The only time it has gone down, with the reason recorded beside the number rather than in a commit message. Two of the 461 respellable rows reached their converged form *through* such a widening, so refusing it costs their convergence.

Under `adjudication-precedence-order` that trade is not close: an allele whose members claim the same territory denotes no single sequence, so emitting one is a **rank-(1) validity** failure, while convergence is **rank (2)**. Rank (1) is absolute and never traded. Two lost convergences is the price, not a regression to argue down.

Explicitly **not** a `PRECEDENCE_ESCALATIONS` entry: an escalation is confluence overriding the *recommended form*. Here validity outranks confluence, which is the ordinary case the ladder describes.

## Verification — measured before the 2026-08-09 restack

- All three `cis_junction_crossing_shift` sweeps go green, plus 7 new tests in `typing_must_not_widen_a_member` with a negative half naming the exact forbidden strings.
- **32 -> 30 failures**, name lists diffed rather than counts compared.
- **Identical 30 under a debug run with `FERRO_ASSERT_IDEMPOTENT`, `FERRO_ASSERT_REPARSE` and `FERRO_ASSERT_IN_BOUNDS` all set — zero oracle-only failures.** Worth stating because those three are `#[cfg(debug_assertions)]`, so a release-only gate compiles them out and cannot see them.
- `cargo fmt --check` and `cargo clippy --all-features --all-targets` clean.

**The parent has since moved.** The restack rebased this branch onto `7df520ed`, so the absolute 32 -> 30 counts are due a re-run; the claims to re-establish are "three sweeps green", "zero newly red" and "identical set under the three seam oracles", not the number 30.

Representation-Change: 10 previously-invalid outputs in the two-member sweep of 46,080 are repaired, and 2 rows of `multi_member_cis_axis`' 461 respellable real-data rows lose their converged form (`respelling_converged` 376 -> 374). The repaired outputs were never valid descriptions — an allele claiming one interbase point twice denotes no sequence — so no consumer can hold a correct stored string for them. The 2 convergence losses are the disclosed cost of that repair, and the direction is fixed by `adjudication-precedence-order`: validity is rank (1), convergence is rank (2). This is the per-PR delta on top of #1565's stack-wide move and is not additive with it; #1565's separate `main` -> this-branch measurement of 6,851 of 78,298 rows already includes these.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -7997,7 +7997,10 @@ pub(crate) fn demote_repeats_spanning_siblings<P: ReferenceProvider>(
         ) {
             continue;
         }
-        let Some(source) = cis_axis_parts(&before[i], kind).and_then(|(_, _, _, _, e)| match e {
+        // The *differential* reading of what the repeat grew out of: how many
+        // bases the asserted member added or removed. `None` for anything else,
+        // which is where the state-based route below takes over.
+        let differential = cis_axis_parts(&before[i], kind).and_then(|(_, _, _, _, e)| match e {
             // A deletion removes the bases under its span.
             NaEdit::Deletion { .. } => Some((RepeatSource::Removed, b.end - b.start + 1)),
             // A duplication adds a copy of the bases under its span.
@@ -8009,11 +8012,8 @@ pub(crate) fn demote_repeats_spanning_siblings<P: ReferenceProvider>(
                 .and_then(|n| i64::try_from(n).ok())
                 .map(|n| (RepeatSource::Added, n)),
             _ => None,
-        }) else {
-            continue;
-        };
-        let (source, length) = source;
-        if length <= 0 || b.region != a.region {
+        });
+        if differential.is_some_and(|(_, length)| length <= 0 || b.region != a.region) {
             continue;
         }
         // Does the tract actually swallow a sibling? Both snapshots, so a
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
Closing. Re-measured against re-derivation on `main` (`7118eb2a`) with the prepared reference, and the case does not survive the move.

## The rank-1 defect this repairs does not occur on `main`

`tests/it/typing_must_not_widen_a_member.rs` was applied to `main` unchanged and run: **6 of 7 pass**, including the two that carry the argument —

- `no_rendered_member_claims_a_siblings_territory` (the exhaustive enumeration)
- `the_repaired_answers_denote_the_input_sequence` (through the SPDI applier, independent of the normalizer)
- `the_swallowing_and_near_miss_forms_are_never_emitted`

So the self-overlapping allele — the output that denotes *no* sequence, and the whole reason this is a rank-(1) validity failure rather than a reading — is not produced under re-derivation. The one failing row is not a validity failure either:

```
input     TEMPLATE:g.[2_3dup;5_6insA]
main      TEMPLATE:g.5_6insTTA        (one member, disjoint, re-parses, right bases)
expected  TEMPLATE:g.[3_4dup;5_6insA] (the two-member form the preserving arm holds)
```

That is a preservation expectation, not a defect. `main`'s answer is valid; it simply has one member where the asserted partition had two.

## The `376 → 374` ratchet argument cannot be evaluated as stated

This PR lowers `AXIS_CENSUS.respelling_converged` from 376 to 374 and argues the two lost convergences are the correct price for the repair above.

Measured on **bare `main`** with the prepared reference, that census reads **395**, not 376:

```
left:  AxisCensus { rows: 592, …, respellable: 461, respelling_converged: 395 }
right: AxisCensus { rows: 592, …, respellable: 461, respelling_converged: 376 }
```

So the pin this PR subtracts two from is already 19 rows stale, and `376 − 2` is arithmetic against a baseline that does not hold. #1581 owns that discrepancy: it bisects the 19 rows to `30ca8f7d` (#1484), finds 15 unambiguous and 4 questionable, and replaces the count with a pinned row list precisely because "seven integers cannot distinguish `+19` from `+20 −1`". Re-deriving a ratchet argument on top of the count would repeat the defect that PR exists to fix.

The trade this PR describes may well be right in principle — validity outranks confluence, and `adjudication-precedence-order` says so. But there is no rank-1 failure here to buy, and the price is denominated in a stale unit.

## The guard is close to duplicate

`main`'s `cis_junction_crossing_shift.rs` already asserts `overlapping.is_empty()` over its three exhaustive sweeps. This file's `no_rendered_member_claims_a_siblings_territory` is a differently-shaped enumeration — `dup`/`del` first members against `ins`/`dup` siblings on a nine-`T` tract — over the same invariant, so it is incremental coverage rather than a new property.

It is worth having on those grounds alone (a second generator is blind to different things), and the file is 6/7 green on `main` as-is. But its 78-line module doc is written entirely in terms of "the partition model's invariant" and `partition_block_preserving`, so landing it means rewriting the argument rather than moving the file. **If someone wants that coverage, the file is recoverable from this branch** — it is left as a deliberate follow-up rather than smuggled in with a stale rationale.

Nothing in `src/normalize/merge.rs` here applies: the population it targets is one the preserving arm produced.

PR #1563, opened 2026-08-08
chore: release v0.14.0

nh13:
## Representation stability

Any PR that moves a normalized output should carry a `Representation-Change:`
trailer (see CONTRIBUTING.md). Those are collected into the **Representation
changes** section of the changelog below, and each entry's trailer text is
attached under its bullet by `scripts/inject_representation_disclosure.py`.

Reviewer checklist:

- [ ] Either that section lists every output-moving change in this cycle, or
      there were none and it is absent.
- [ ] Nothing in it merely *declined*. A decline reads `Representation-Change:
      none`, optionally followed by a reason, and belongs under its own type —
      not here. Spot-check two entries against their PR descriptions.
- [ ] Every entry carries its trailer's own text, quoted under the bullet, and
      that text says whether the affected inputs were previously **rejected**
      (no stored string, so free) or previously **accepted** (a real migration
      for the consumer). release-plz writes the bullet from the commit subject
      alone; `python scripts/inject_representation_disclosure.py` attaches the
      rest, and CI's `Changelog grouping audit` fails until it has been run.
      Re-run it after any push that regenerates this section — it is idempotent.

A commit under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` that
moved output without declaring it means the trailer is missing. Fix that before
releasing, not after.

---

### Representation changes

- *(normalize)* carry the reference copies a re-phased repeat absorbs ([#1678](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1678))
- *(error_handling)* keep the repeat unit when the W3013 range disagrees with it ([#1661](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1661))
- *(project)* decide the codon-frame delins exception per rendered axis ([#1672](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1672))
- *(normalize)* decline a protein delins whose affix trim lands on the initiation codon ([#1670](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1670))
- *(normalize)* scope the delins payload-coincidence carve-out to coding DNA ([#1682](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1682))
- *(normalize)* coalesce protein cis adjacency created by the 3'-shift, and add the corpus axis that found it ([#1614](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1614))
[...]

Issue #1559, opened 2026-08-08
changelog: guard the Representation changes section as a class, not one bug at a time


nh13:
## The section is only audited by hand

Two bugs have shipped in the **Representation changes** section, in opposite directions:

- **#1526** — one decline listed as a representation change.
- **#1555** — eight of them. The v0.13.1 changelog listed 10 entries of which 2 were real.

Both were found by a person reading the rendered section against the PRs. That is still the only detector, and it only ran because someone happened to look before a release.

**The existing test cannot find this class.** #1527 added `test_decline_vocabulary_matches_the_changelog_config` for exactly this drift, and it passed throughout #1555 — because the two halves had *not* drifted. They named the same four words and were wrong together. An agreement-based test is blind to a shared mistake, which is the failure mode both bugs had.

## Three guards

**1. A decline that then describes a move.** The verdict is the first word, so `no. 3 rows move` is filed as a decline and the disclosure never reaches the changelog. That is the silent direction — the one #1522 is about. A tripwire on `<n> rows move|merge|split|respell` for non-zero `n` catches the phrasing this repository actually uses.

The rule has to be narrow or it punishes good disclosures. Measured over the 15 declines of the v0.13.1 cycle, a loose pattern fires on #1547 — `none. 0 of 950 real cis-allele rows move their normalized string` — which is an exemplary decline. Requiring the count to sit immediately before `rows`, and excluding zero, gives 0 false positives across all 15 and catches 5 of 5 synthetic contradictions.

**2. The ordering-prefix strip is anchored to a rendering shape nothing pins.** #1558's postprocessor matches `^### <!-- N -->`. If a release-plz bump changes the heading level or the spacing, the prefix silently returns — the same shape as the bug it fixed, where #1433 asserted a rendering behaviour without checking it and the claim survived two releases.

**3. Nothing checks the rendered section against the trailers.** The class stays unguarded even with 1 and 2 fixed.

## Proposal

One audit that renders the changelog with the real `release-plz.toml` over `<latest tag>..HEAD` and asserts, of the result:

- nothing filed under **Representation changes** opens with a word that means no;
- nothing that declares a move is filed anywhere else;
- no heading still carries its `<!-- N -->` prefix.

The decline rule must be **derived independently** of `scripts/check_representation_change.py` — a plain first-word test, no shared vocabulary — or it reproduces the blindness described above. Rendering commit ids rather than messages keeps the mapping back to commits exact and independent of message formatting.

[...]

PR #1574, opened 2026-08-09
fix(changelog): read a decline as a decline when text follows the trailer

### LINKED ISSUE #1573: changelog: a bare decline followed by appended text is filed as a representation change
## A bare decline followed by any trailing text is filed as a representation change

`release-plz.toml`'s decline-exclusion rule anchors with `$` under `(?i)` only:

```
(?i)^Representation-Change:[^\S\r\n]*(none|no|n/a|na)[^\S\r\n]*(?:[.;:—–][\s\S]*)?$
```

git-cliff hands the regex the **whole footer value**, which can span lines. With no `m` flag, `$` means end-of-string, so the rule only matches when the decline is the last thing in the commit message. Anything appended after a *bare* `none` — most commonly CodeRabbit's auto-generated release-notes block — defeats the match, the inclusion rule below it matches instead, and the commit is filed under **Representation changes**.

`#1555` fixed the sibling case (`none. <reason>`) by letting a reason follow a terminator from `.;:—–`. A newline is not in that class, so the bare-decline-plus-appended-block shape was left behind.

## This is live, in a shipping artifact

`3ebcdddd` (#1551) declared `Representation-Change: none` and had CodeRabbit release notes appended after it. The open v0.14.0 release PR (#1563) lists it as a representation change:

```
### Representation changes

- *(normalize)* widen the axis gate to c./n./r. (#1484)
- *(conformance)* refuse to write an artifact built from a partial generator pass (#1551)   <-- declined
```

One of the two entries is real. This is #1526 recurring in the shape #1555 did not cover.

## The two halves disagree, which is the actual defect

`scripts/check_representation_change.py` parses the trailer with `re.IGNORECASE | re.MULTILINE` and takes the value from the trailer's **own line**, so it reads `3ebcdddd` as a decline and passes it. The changelog rule reads the same trailer as a real disclosure. The checker and the changelog give opposite verdicts on the same commit.

`test_decline_vocabulary_matches_the_changelog_config` cannot see this: the two halves name the same four words. They agree on vocabulary and disagree on anchoring, and every existing guard inspects the config as a **string** rather than running git-cliff over a message.

## Reproduction

Rendering ten trailer shapes through the real rule with git-cliff 2.13.1 — `(?i)` versus `(?im)`, nothing else changed:

| # | trailer | want | `(?i)` | `(?im)` |
|---|---|---|---|---|
| 1 | `none` | Other | Other | Other |
| 2 | `none` + CodeRabbit block | Other | **Representation changes** | Other |
| 3 | `none. Comment prose only.` | Other | Other | Other |
| 4 | `none. <reason>` + block | Other | Other | Other |
| 5 | `none; tests only` | Other | Other | Other |
| 6 | `NONE` + block | Other | **Representation changes** | Other |
| 7 | `577 rows move, 360 merge / 205 split.` | Representation changes | Representation changes | Representation changes |
| 8 | real disclosure + block | Representation changes | Representation changes | Representation changes |
| 9 | real, multi-line indented continuation | Representation changes | Representation changes | Representation changes |
| 10 | `none, except two rows that merge.` | Representation changes | Representation changes | Representation changes |

Rows 2 and 6 are the bug. Row 10 confirms the deliberate exclusion of `,` from the terminator set survives, and rows 3-5 confirm #1555's reason-style declines are unaffected.

## Proposed fix

Add the `m` flag to the exclusion rule so it evaluates the trailer's own line, which is what the checker already does. The two halves then agree by construction rather than by coincidence.

Guarding it needs a test that *renders* rather than one that inspects the config string, since anchoring is a behaviour and the existing tests are all string assertions.

## Relationship to #1559 / #1560

#1560 adds the rendered-output audit this class needs, and its rule would flag this commit. It is a detector, not the fix: with the config unchanged, the audit reports the mis-filing but the mis-filing remains. The two are complementary.

### LINKED ISSUE #1586: Two Representation-Change trailers: the checker reads the first, the changelog lets a decline win
A commit message carrying **two** `Representation-Change:` trailers is read one way by `scripts/check_representation_change.py` and the opposite way by `release-plz.toml`. This is the #1526 / #1555 / #1573 class in a shape none of the three reach, and it survives #1573's fix.

## The disagreement

git-cliff matches a `footer` rule against **each** footer of the parsed commit, and a group is assigned on the *first rule that matches any footer*. The checker instead takes the **first** trailer in the message and ignores the rest (`TRAILER_RE.search`, singular). So when one trailer discloses a move and a later one declines, the two halves rank them differently: the checker reads the first, the changelog lets the declining one win regardless of position.

Measured against git-cliff 2.13.1 with the config on current `main`:

```
Representation-Change: 577 rows move, 360 merge.

For contrast:
Representation-Change: none
```

```
checker  -> declaration = '577 rows move, 360 merge.'   declines? False  => REAL CHANGE
git-cliff -> group = Other                                               => DECLINE
```

The commit passes the `Representation change declared` check as a real disclosure and is then filed under **Other**, so a genuine representation change is absent from the changelog — the under-reporting direction of #1522, one commit at a time.

## It is not an anchoring bug, and #1573 does not change it

Worth stating because the shape looks like #1573. It reproduces identically with `(?i)`, with `(?im)` as #1573 ships it, and with `\A` substituted for `^`:

| exclusion rule | bare `none` + appended block | this shape |
|---|---|---|
| `(?i)^Representation-Change:…` (before #1573) | **wrong** (Representation changes) | **wrong** (Other) |
| `(?im)^Representation-Change:…` (#1573) | right | **wrong** (Other) |
| `(?im)\ARepresentation-Change:…` | right | **wrong** (Other) |

The cause is that the second line is parsed as its own footer, so the rule matches at *that* footer's own start — no line anchoring is involved. Indenting the second occurrence makes it part of the first footer's value and the commit groups correctly, which isolates the mechanism:

```
indented second trailer  -> Representation changes   (correct)
col-0 second trailer     -> Other                    (wrong)
```

## How it is reachable

Not by a bot. CodeRabbit's appended release-notes block contains no `Representation-Change:` line, so #1573 covers the common case. The realistic triggers are human:

- a PR description that *documents* the trailer — quoting a declining example at column 0 below its own real disclosure, which the guidance in `CONTRIBUTING.md` invites;
- a corrected trailer added without deleting the superseded one.

Low frequency, but the failure is silent in the direction that matters: the disclosure disappears and nothing reports it.

## Suggested fix

Make the two halves agree on *which* trailer is authoritative. The checker's "first trailer wins" is the sensible rule and is already implemented; the changelog side is what cannot currently express it, since git-cliff evaluates every footer. Options, roughly in order of preference:

1. **Reject the ambiguity instead of resolving it.** Have `scripts/check_representation_change.py` fail when a message carries more than one `Representation-Change:` trailer. Two trailers is not a thing a contributor means, and refusing it makes the resolution question moot — this is the cheapest fix and needs no git-cliff behaviour.
2. Have the checker apply the changelog's rule (any decline wins) so the two agree, and document that a later decline overrides an earlier disclosure. Consistent, but the semantics are worse.

Option 1 also fits the precedent that a decline must be *stated*, not inferred.

## Guard

Whatever the resolution, it wants a case in `tests/python/test_representation_change_trailer.py`. Note the limit that #1573's own guard documents: every assertion in that module models git-cliff's matcher in Python `re` rather than rendering through git-cliff, so a two-footer case asserted there pins the checker's half only. The rendered-output audit in #1560 is where the changelog half can actually be pinned.

nh13:
A `Representation-Change: none` trailer stops being read as a decline as soon as anything is appended after it, and the commit is then filed under **Representation changes** as if it had moved output.

Closes #1573. Closes #1586.

## The mechanism

git-cliff matches a `footer` rule against the **whole footer value**, which can span lines. The exclusion rule declared `(?i)` and anchored with `$`, so `$` meant end-of-*string* and the rule only fired when the decline was the last thing in the message.

#1555 let a reason follow the verdict, terminated by one of `.;:—–`. That does not cover a **bare** decline, because the character after it is a newline rather than punctuation. So `none. Tests only.` was read correctly and plain `none` was not — as long as something followed it.

Something almost always follows it: CodeRabbit appends a release-notes block to essentially every PR description, and GitHub builds the squash commit body from the description.

## It is live in a shipping artifact

`3ebcdddd` (#1551) declared `Representation-Change: none`, CodeRabbit appended its block, and the open v0.14.0 release PR (#1563) lists it as a representation change:

```
### Representation changes

- *(normalize)* widen the axis gate to c./n./r. (#1484)
- *(conformance)* refuse to write an artifact built from a partial generator pass (#1551)   <- declined
```

One of those two entries is real. This is #1526 recurring in the shape #1555 did not reach.

## Why every existing guard passed

`scripts/check_representation_change.py` parses the trailer with `re.IGNORECASE | re.MULTILINE` and reads the value off the trailer's **own line**. It calls `3ebcdddd` a decline. The changelog rule calls the same commit a disclosure. The checker and the changelog give opposite verdicts on one commit, which is precisely the drift `test_decline_vocabulary_matches_the_changelog_config` was written for — and it passes, because the two halves agree on *vocabulary* and disagree on *anchoring*.

`test_the_two_decline_rules_agree_value_by_value` was the one guard positioned to catch it, and it could not, for a reason worth recording:

```python
changelog_rule = re.compile(rules[0].replace("\\\\", "\\"), re.MULTILINE)
```

It supplied `re.MULTILINE` itself. git-cliff supplies no flags, so the harness was strictly kinder than production and the rule under test was not the rule that ships. Every other guard in the file inspects the config as a *string* — vocabulary, casing, ordering — and anchoring is a behaviour.

## The change

Add `m` to the exclusion rule, so the verdict is read off the trailer's own line, which is what the checker already does. The two halves now agree by construction rather than by coincidence.

`[\s\S]*` after a terminator is untouched, so a reason may still span lines, and `,` is still excluded, so `none, except two rows` remains a real change.

The test now compiles the configured rule with **no flags of its own**, and a new case pins the verdict against text appended after the trailer.

## Verified by rendering, not by assertion

Ten trailer shapes through git-cliff 2.13.1 with the real config, `(?i)` versus `(?im)`, nothing else changed:

| # | trailer | want | `(?i)` | `(?im)` |
|---|---|---|---|---|
| 1 | `none` | Other | Other | Other |
| 2 | `none` + CodeRabbit block | Other | **Representation changes** | Other |
| 3 | `none. Comment prose only.` | Other | Other | Other |
| 4 | `none. <reason>` + block | Other | Other | Other |
| 5 | `none; tests only` | Other | Other | Other |
| 6 | `NONE` + block | Other | **Representation changes** | Other |
| 7 | `577 rows move, 360 merge / 205 split.` | Representation changes | Representation changes | Representation changes |
| 8 | real disclosure + block | Representation changes | Representation changes | Representation changes |
| 9 | real, multi-line indented continuation | Representation changes | Representation changes | Representation changes |
| 10 | `none, except two rows that merge.` | Representation changes | Representation changes | Representation changes |

Rows 2 and 6 are the bug. Rows 3–5 confirm #1555's reason-style declines are unaffected; row 10 confirms the deliberate exclusion of `,` survives.

End to end, over the real history — #1560's rendered-output audit run against `v0.13.1..HEAD`:

```
before:  3 commits in v0.13.1..HEAD; 2 listed as representation changes
         3ebcdddd is filed under Representation changes but its trailer opens with a decline: 'none'
after:   3 commits in v0.13.1..HEAD; 1 listed as representation changes
         Representation changes section is consistent with the trailers.
```

The new test fails on exactly the two bare-decline cases with the config reverted and nothing else changed, and 52 pass with the fix.

## Second fix, same family: two trailers in one message (#1586)

Found while verifying the above by rendering. It is the same "the two halves disagree" defect in the one shape `m` does not reach, so it is fixed here rather than left open.

The two halves resolve *duplicates* differently. This checker reads the **first** trailer (`TRAILER_RE.search`, singular). git-cliff matches its footer rule against **every** footer and takes the first rule that matches any of them, so a decline wins from wherever it sits. A message that discloses a move and then quotes a declining example therefore passes CI as a disclosure and is filed under **Other** — the disclosure silently disappears, which is #1522's under-reporting direction one commit at a time.

Measured against git-cliff 2.13.1 on the real config:

```
checker   -> '577 rows move, 360 merge.'  declines? False  => REAL CHANGE
git-cliff -> group = Other                                 => DECLINE
```

**It is not an anchoring bug, and `m` does not reach it.** It reproduces identically under `(?i)`, under `(?im)` as this PR ships it, and with `\A` substituted for `^`, because the second line is parsed as its own footer and the rule matches at *that* footer's start. So `\A` would buy nothing here and is not proposed.

| exclusion rule | bare `none` + appended block | two trailers |
|---|---|---|
| `(?i)` (before) | wrong | wrong |
| `(?im)` (this PR) | right | wrong |
| `(?im)\A` | right | wrong |

**Refuse rather than resolve.** git-cliff evaluates every footer, so "first trailer wins" cannot be expressed as a footer regex — there is no rule that makes it agree. `check` now fails on more than one trailer, ahead of the watched-file test, because changelog grouping applies to every commit and this job already runs on every PR. That also keeps the rule that a decline must be *stated*, never inferred from a message that also says the opposite.

**The trailer token must now sit at column 0.** The pattern allowed `[ \t]*` before it, so the checker read as a *trailer* what git and git-cliff treat as a *continuation* — and `CONTRIBUTING.md` documents indentation as exactly that mechanism. Aligning it is what makes the escape hatch in the refusal message true: an indented quotation is one trailer to both halves, while a column-0 duplicate is refused. A sole *indented* trailer now fails as absent, which is the honest answer, since git-cliff would not group it either.

Blast radius, measured over every open PR at the time of the change: **no description carries a duplicate or an indented trailer**, so nothing is reclassified.

All three shapes now agree:

| shape | git-cliff (measured) | checker |
|---|---|---|
| col-0 duplicate | Other | refused, never ships |
| indented quotation under a disclosure | Representation changes | Representation changes |
| bare decline + appended block | Other | Other |

## Relationship to #1559 / #1560

#1560 adds the rendered-output audit this class needs and its rule **does** flag this commit — I ran its script against current `main` to check. It is the detector; this is the fix, and the two are complementary. Worth knowing for sequencing: #1560's `Changelog grouping audit` is green only because that run predates #1551 landing on `main`. Rebasing it onto current `main` should turn it red until this merges.

## Testing

- `pytest tests/python/test_representation_change_trailer.py` — **62 pass** (52 before, 10 new for the duplicate-trailer and column-0 rules).
- Control: with only the config flag reverted, the two bare-decline cases fail.
- #1560's own audit script, run against real history (`v0.13.1..HEAD`): clean with this fix (`Representation changes section is consistent with the trailers`), and with only the flag reverted it flags `3ebcdddd`. So the detector and the fix are verified against each other, not just asserted to be complementary.
- `ruff check` / `ruff format --check` clean on `tests/python/`.
- The rest of `tests/python/` cannot be collected in a fresh worktree without `maturin develop --features python` (`ModuleNotFoundError: No module named 'ferro_hgvs'`); that is unrelated to this change, which touches only the TOML config and a stdlib-only test module.

Representation-Change: none. Changelog config, the trailer checker and their tests only; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched.

### DIFF
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -135,7 +135,22 @@ commit_parsers = [
   # `DECLINE_TERMINATORS` in `scripts/check_representation_change.py` by
   # `test_decline_vocabulary_matches_the_changelog_config` and
   # `test_the_two_decline_rules_agree_value_by_value`.
-  { footer = "(?i)^Representation-Change:[^\\S\\r\\n]*(none|no|n/a|na)[^\\S\\r\\n]*(?:[.;:—–][\\s\\S]*)?$", group = "<!-- 7 -->Other" },
+  #
+  # **`m`, because git-cliff matches this against the whole footer value, which spans
+  # lines** (#1573). Without it `$` is end-of-*string*, so the rule only fires when the
+  # decline is the last thing in the message -- and anything appended after a BARE `none`
+  # defeats it. That is not hypothetical: CodeRabbit appends a release-notes block to the
+  # PR description, GitHub makes the description the squash body, and `3ebcdddd` (#1551)
+  # declined and was then filed under "Representation changes" in the v0.14.0 changelog.
+  # The `[.;:—–]` terminator above does not cover it, because the character following a
+  # bare decline is a newline rather than punctuation.
+  #
+  # With `m` the verdict is read off the trailer's own line, which is what the checker
+  # already does -- its `TRAILER_RE` is `re.IGNORECASE | re.MULTILINE` and takes the value
+  # from one line. The two halves now agree by construction rather than by coincidence.
+  # A reason may still span lines: `[\s\S]*` after a terminator is unchanged, and `,` is
+  # still excluded, so `none, except two rows` remains a real change.
+  { footer = "(?im)^Representation-Change:[^\\S\\r\\n]*(none|no|n/a|na)[^\\S\\r\\n]*(?:[.;:—–][\\s\\S]*)?$", group = "<!-- 7 -->Other" },
   # A declining commit lands in `Other` whatever its type, so a `fix:` that correctly
   # declined is not listed under `Fixed` (#1557). That is a known cost of the ordering
   # above, not an oversight, and the obvious repair does not work: **git-cliff evaluates a
--- a/scripts/check_representation_change.py
+++ b/scripts/check_representation_change.py
@@ -55,8 +55,19 @@ WATCHED_PREFIXES: tuple[str, ...] = (
 #: space as the value, so a bare `Representation-Change:` with nothing after it counted as
 #: a declaration — the precise hole this check exists to close. And `\s` spans newlines, so
 #: a permissive class would let the value be scavenged from the *following* line.
+#:
+#: **The token must sit at column 0** (#1573). The pattern used to allow `[ \t]*` before it,
+#: which made this checker recognise as a *trailer* what git and git-cliff both treat as a
+#: *continuation* of the value above it — and `CONTRIBUTING.md` documents indentation as
+#: exactly that continuation mechanism. The disagreement is only reachable when a
+#: continuation line itself opens with the token, i.e. when a PR description quotes a
+#: declining example under its own real disclosure, but there it is the whole defect: the
+#: quoted example became a second declaration here and stayed prose for git-cliff. Requiring
+#: column 0 makes both halves count the same trailers, which is what lets `check` refuse a
+#: genuine duplicate and pass an indented quotation. Measured over every open PR at the time
+#: of the change: no description carried an indented trailer, so nothing was reclassified.
 TRAILER_RE = re.compile(
-    r"^[ \t]*Representation-Change:[ \t]*(?P<value>\S.*?)[ \t]*$",
+    r"^Representation-Change:[ \t]*(?P<value>\S.*?)[ \t]*$",
     re.IGNORECASE | re.MULTILINE,
 )
 
@@ -103,25 +114,53 @@ def watched_files(changed: list[str]) -> list[str]:
     return [p for p in changed if any(p.startswith(prefix) for prefix in WATCHED_PREFIXES)]
 
 
+def find_declarations(text: str) -> list[str]:
+    """Return every trailer value in `text`, in the order they appear.
+
+    Plural because *how many* there are is itself a verdict — see `check`. One trailer is
+    the only shape whose meaning both this checker and `release-plz.toml` agree on.
+    """
+    return [match.group("value") for match in TRAILER_RE.finditer(text)]
+
+
 def find_declaration(text: str) -> str | None:
-    """Return the trailer's value, or `None` when the text carries no trailer."""
-    match = TRAILER_RE.search(text)
-    return match.group("value") if match else None
+    """Return the first trailer's value, or `None` when the text carries no trailer."""
+    declarations = find_declarations(text)
+    return declarations[0] if declarations else None
 
 
 def check(changed: list[str], declaration_text: str) -> tuple[bool, str]:
     """
     Decide whether `changed` is adequately declared by `declaration_text`.
 
-    Returns `(ok, message)`. `ok` is False only when a watched file changed and no
-    trailer is present at all — a trailer whose value is `none` passes, because
-    declining is a declaration.
+    Returns `(ok, message)`. `ok` is False in two cases: a watched file changed and no
+    trailer is present at all, or the message carries more than one trailer. A trailer
+    whose value is `none` passes, because declining is a declaration.
     """
+    # Refused BEFORE the watched-file test, and deliberately: the harm is in changelog
+    # grouping, which applies to every commit, not only to the ones touching a watched
+    # directory. This job already runs on every PR.
+    declarations = find_declarations(declaration_text)
+    if len(declarations) > 1:
+        listed = "\n".join(f"  Representation-Change: {value}" for value in declarations)
+        return False, (
+            f"{len(declarations)} `Representation-Change:` trailers found:\n{listed}\n\n"
+            "Keep exactly one. Two trailers do not mean anything consistent, because this\n"
+            "check and the changelog resolve them differently: this check reads the FIRST\n"
+            "trailer, while git-cliff matches its footer rule against EVERY footer and takes\n"
+            "the first rule that matches any of them -- so a decline anywhere wins there,\n"
+            "whatever its position. A message disclosing a move and then declining one\n"
+            "therefore passes here as a disclosure and is filed under `Other` in the\n"
+            "changelog, and the disclosure silently disappears.\n\n"
+            "Delete the superseded trailer, or indent it if it is quoted as an example --\n"
+            "an indented line is a continuation of the value above it, not a new trailer."
+        )
+
     watched = watched_files(changed)
     if not watched:
         return True, "No file under a watched directory changed; no declaration needed."
 
-    declaration = find_declaration(declaration_text)
+    declaration = declarations[0] if declarations else None
     if declaration is None:
         listed = "\n".join(f"  {p}" for p in sorted(watched))
         return False, (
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

PR #1576, opened 2026-08-09
test: Adding tests from mutation testing findings

geoffjentry:
Testing (har har) the waters. As I'm new to both Rust & the subject matter I'm just submitting some basic tests as a trial balloon. 

The background was that I'd been running a mutation tester on the repo over the last few days, and am starting to attack the findings.


nh13:
@coderabbitai pause

PR #1583, opened 2026-08-10
fix(parser): validate ring segments and sup inners in every post-parse check via a shared leaf walker (#1578)

### LINKED ISSUE #1578: Post-parse semantic validators miss `GenomeRing` segments and `Supernumerary` inners; #1264 fixed only point-insertion
## Summary

Four of the parser's post-parse semantic validators match on the variant kind and end in `_ => {}`, with no `HgvsVariant::GenomeRing` or `HgvsVariant::Supernumerary` arms. A ring's `::`-joined segments are `LocEdit<GenomeInterval, NaEdit>` hanging off `GenomeRing` rather than nested `HgvsVariant`s, so the `Allele` recursion cannot reach them. Every edit-form rule those validators enforce is therefore unenforced inside a `::`-join, while the identical spelling outside a join is rejected.

This is the same defect #1332 (closes #1264) found and fixed in `validate_no_point_insertion` — that fix added `GenomeRing`/`Supernumerary` arms to one validator (`src/hgvs/parser/variant.rs:7754`, `:7763`) and left the other four untouched.

## Measured (2026-08-09, `parse_hgvs` level, at 2c32b73)

Each row's bare form is rejected; the same edit inside a ring is accepted:

| rule | bare | in ring (`…::200_201insGG`) | in sup-wrapped ring (`g.[…::…]sup`) |
|---|---|---|---|
| `validate_no_dupins` — `g.100_101dupinsATG` | rejected | **accepted, round-trips** | **accepted, round-trips** |
| `validate_no_point_size_suffix` — `g.123del3` | rejected | **accepted, round-trips** | not separately probed; same code path |
| `validate_no_single_nucleotide_inversion` — `g.100inv` | rejected | **accepted, round-trips** | not separately probed; same code path |
| `validate_no_multibase_substitution` — `g.100_101AA>TT` | rejected | **accepted, silently rewritten to `g.100_101delinsTT::…`** | **accepted, same rewrite** |
| `validate_no_point_insertion` — `g.100insA` (control) | rejected | rejected (#1264's fix) | rejected |

Legal-edit control, verified to parse and round-trip: `NC_000022.11:g.100_101dup::200_201insG`.

Two scope facts, measured so nobody re-derives them:

- **Sup-over-plain-genome is not a second hole.** `make_supernumerary` admits only an editless whole-chromosome inner or a ring, and a plain inner is validated before wrapping; `NC_000001.11:g.100_101dupinsATsup` is rejected. The gap is reachable only through the two ring shapes above.
- **The multibase row must be probed at the `parse_hgvs` level, not through the CLI.** `ferro parse` runs the input preprocessor, which intercepts the multi-base spelling itself (`src/error_handling/preprocessor.rs:1153`), so the CLI rejects `g.100_101AA>TT::…` for a reason that says nothing about the library path. The rows above come from a scratch binary calling `parse_hgvs` directly. This trap already produced one wrong conclusion during investigation.

The multibase case has a distinct symptom worth naming: the grammar parses `AA>TT` into a delins representation, so instead of round-tripping an invalid form, ferro *auto-corrects* it inside a ring while rejecting the identical spelling outside one. Different symptom, same root cause.

## Why no oracle can see this

By construction, all three are blind. `FERRO_ASSERT_REPARSE` cannot fire: three of the four escapes render back exactly what was accepted, and the fourth renders a legal delins. `FERRO_ASSERT_IN_BOUNDS` asks about coordinates, not edit-form legality. Idempotency holds trivially. Only parser tests can pin this — the same conclusion #1332's comment records for the point-insertion instance.

## Spec basis

No exemption is plausible. `complex.md:38-42` defines `::` segments as ordinary `g.` descriptions (the worked examples at `complex.md:72-83` use standard syntax throughout), and the four prohibitions are context-free:

- `DNA/duplication.md:92` — `dupins` marked `<code class="invalid">`
- `checklist.md:49` / `DNA/deletion.md:117` / `DNA/duplication.md:140` — size-count suffixes
- `DNA/inversion.md:16` — one-nucleotide inversion
- `DNA/substitution.md:30`, `DNA/delins.md:73` — multi-base substitution

geoffjentry:
Closes #1578.

Representation-Change: zero normalized outputs move; previously-accepted invalid ring/sup spellings across six validators now reject, lenient/silent modes reject a ring del<size> they previously passed verbatim, and member/segment repair-suggestion text changes.

**NB** Same disclaimer as #1576 that I'm new to both Rust & the material. This all "feels right" but I could be off base in a few different dimensions.

## The defect

Post-parse semantic validators each hand-rolled their own Allele recursion and ended in _ => {}. A ring's ::-joined segments are LocEdit<GenomeInterval, NaEdit> hanging off GenomeRing — not nested HgvsVariants — so no validator reached them by matching on variant kind, and Supernumerary inners fell into the same catch-all. #1332 patched exactly one validator; six more had measured live escapes (see #1578 for the probe evidence), each accepting inside a ::-join the exact spelling it rejects standalone.

All three normalization oracles are blind to this class by construction: the accepted outputs round-trip, so only parser tests can pin it.

## The fix

One structural walk replaces the eleven hand-rolled recursions. for_each_leaf owns Allele/Supernumerary recursion and yields each ring segment re-wrapped as an ordinary g. leaf carrying the ring's accession; the parse driver runs every leaf rule inside a single walk. validate_no_self_cancelling stays top-level by design (it inspects allele siblings), and the walker's docs fence that plus the node-level asymmetry (the GenomeRing/Supernumerary nodes themselves are never yielded).

Hardening: the walker's leaf arm enumerates every variant kind instead of wildcarding, so adding a variant to HgvsVariant fails to compile at exactly the "structural or leaf?" decision — the defect class becomes a compile error.

## NB: Behavioral changes

1. **Previously-accepted invalid ring/sup spellings now reject**, with byte-identical diagnostics to their standalone spellings: dupins, del<size>/dup<size>, one-nucleotide inv, multi-base substitution (previously silently rewritten to a delins inside a ring), U in a DNA edit, and g. on a non-coding RNA reference — including the editless NR_…:g.pter_qtersup marker.
2. **Lenient/silent modes now reject a ring del<size>** they previously passed through verbatim: the corrections tokenizer cannot cross ::, so no repair exists there yet.
3. **Repair suggestions are scoped to what the leaf is**. A leaf that is the whole description still gets a full replacement to paste; an allele member or ring segment now gets only its edit respelled (rewrite that edit as `c.79_80delinsTT`) — the previous whole-variant advice, followed literally, silently dropped the sibling members or ring segments.
4. **Diagnostic precedence is leaf-major**: or a multi-member allele violating different rules in different members (member order decides which error surfaces). Deliberate, documented on the walker, pinned by test. Error codes and single-leaf diagnostics are unchanged.

## Testing

ests/it/issue_1578_ring_validator_escapes.rs pins each escape in both ring spellings (bare ::-join and g.[..]sup) via error parity with the standalone form — the walker's own invariant — plus a legal-ring round-trip control so the fix cannot pass by over-rejecting, and tests pinning the leaf-major ordering and the suggestion scoping. The existing #1264 and #1092 suites guard the migration (the point-insertion fix is reproduced by the walker; the pub AST-keyed contract of validate_no_multibase_substitution survives, including for directly-passed Alleles).

### DIFF
--- a/src/hgvs/parser/variant.rs
+++ b/src/hgvs/parser/variant.rs
@@ -7240,56 +7240,69 @@ pub fn parse_variant(input: &str) -> Result<HgvsVariant, FerroError> {
     // (HGVS recommendations/general.md line 58, tracked under issue #115).
     validate_no_self_cancelling(&variant, input)?;
 
-    // Spec-mandated post-parse semantic check: reject `dupins` mash-up
-    // (HGVS DNA/duplication.md:92 — "a format not used in HGVS
-    // nomenclature"; #445).
-    validate_no_dupins(&variant)?;
+    // All remaining post-parse semantic checks are per-leaf rules; run them
+    // inside one structural walk so ring segments and supernumerary inners
+    // are validated exactly like standalone descriptions (#1578).
+    // Whether the walk's leaves ARE the authored description: false when the
+    // top level is an allele bracket, a `::`-join, or a `sup` marker, whose
+    // leaves are members/segments of something larger. A repair suggestion
+    // for such a leaf must respell only the offending edit — offering it as a
+    // whole-variant replacement would silently drop the siblings.
+    let leaf_is_whole = !matches!(
+        &variant,
+        HgvsVariant::Allele(_) | HgvsVariant::GenomeRing(_) | HgvsVariant::Supernumerary(_)
+    );
+    for_each_leaf(&variant, &mut |leaf| {
+        // Reject `dupins` mash-up (HGVS DNA/duplication.md:92 — "a format
+        // not used in HGVS nomenclature"; #445).
+        validate_no_dupins(leaf)?;
 
-    // Spec-mandated post-parse semantic check: reject single-position
-    // insertion (HGVS DNA/insertion.md:95-101 Q&A — explicit "No"; #446).
-    validate_no_point_insertion(&variant, input)?;
+        // Reject single-position insertion (HGVS DNA/insertion.md:95-101
+        // Q&A — explicit "No"; #446).
+        validate_no_point_insertion(leaf)?;
 
-    // Spec-mandated post-parse semantic check: reject a `del`/`dup` size
-    // suffix on a single-position anchor (HGVS checklist.md:49,
-    // DNA/deletion.md:117, DNA/duplication.md:140, RNA/deletion.md:49; #1079).
-    validate_no_point_size_suffix(&variant, input)?;
+        // Reject a `del`/`dup` size suffix on a single-position anchor
+        // (HGVS checklist.md:49, DNA/deletion.md:117,
+        // DNA/duplication.md:140, RNA/deletion.md:49; #1079).
+        validate_no_point_size_suffix(leaf, input)?;
 
-    // Spec-mandated post-parse semantic check: reject a multi-nucleotide
-    // substitution (HGVS DNA/substitution.md:30, DNA/delins.md:73; #1079).
-    validate_no_multibase_substitution(&variant)?;
+        // Reject a multi-nucleotide substitution (HGVS
+        // DNA/substitution.md:30, DNA/delins.md:73; #1079).
+        validate_no_multibase_substitution_at(leaf, leaf_is_whole)?;
 
-    // Spec-mandated post-parse semantic check: reject a start-codon variant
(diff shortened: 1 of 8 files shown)


nh13:
@coderabbitai pause

PR #1589, opened 2026-08-10
fix(equivalence): decline a ring no rung can evaluate; rule that self-cancellation does not cross `::` (#1578 items 3-4)

nh13:
Implements items 3 and 4 of the follow-up list in https://github.com/fulcrumgenomics/ferro-hgvs/issues/1578#issuecomment-5239725830.

This was written stacked on #1583, which merged (`9b8e55fc`) while it was in progress, so it is rebased onto `main` and targets `main` directly. Nothing else is needed to review it.

Both items are TDD'd, and both turned out slightly different from how the follow-up framed them. The measurements are below rather than the conclusions, because in each case the measurement is what changed the plan.

## Item 4 was live in one module, not four

The follow-up names four modules that hand-roll `Allele` recursion with no `GenomeRing` handling and asks whether the structural blindness is live in them. It is live in exactly one. Probed at this commit against all three ring shapes — the bare `::` join, the `sup`-wrapped ring, and a ring as a cis-allele member:

| module | bare ring | `sup` ring | ring in allele |
|---|---|---|---|
| `src/spdi/apply.rs` | declines | declines | declines |
| `src/vcf/from_hgvs.rs` | declines | declines | declines, naming member index 1 |
| `src/project/projector.rs` | declines | declines | declines |
| `src/equivalence/checker.rs` | **wrong answer** | **wrong answer** | — |

The ring-in-allele shape is construction-only: the parser rejects `g.[100_101del::200_201del;300A>T]`, `g.[300A>T;100_101del::200_201del]` and `g.[[100_101del::200_201del];300A>T]` alike, so the tests build it by hand. It is the shape that matters here, since it is the one that walks a ring *through* the hand-rolled recursion rather than meeting it at the top level.

### The live defect

On a contig with a poly-A run at 1-based 100..109, where each segment's `del` normalizes to `g.109del` standalone — so both ring rows are **one ring spelled two ways**:

```
bare:  g.100del              vs g.105del              -> NormalizedMatch   equivalent=true
ring:  g.100del::200del      vs g.105del::200del      -> NotEquivalent     equivalent=false
ring:  g.100del::200del      vs g.100del::205del      -> NotEquivalent     equivalent=false
sup:   g.[100del::200del]sup vs g.[105del::200del]sup -> NotEquivalent     equivalent=false
```

`NotEquivalent` is a *positive claim* that two descriptions denote different variants, which is strictly worse than the decline the other three modules give.

### Why, and why `Circular` survives the identical pass-through

`check` decides in rungs, and both rungs that compare meaning rather than text are blind at once:

- `Normalizer::normalize` returns a ring unchanged (`src/normalize/mod.rs:2904`), so `NormalizedMatch` degenerates into the string comparison `Identical` already made; and
- `hgvs_to_spdi` has no ring arm, so `edit_triples` declines and the `SequenceMatch` rung (#1158) cannot fire either.

`Circular` is the near miss that locates the defect precisely. `o.` *also* passes through normalization untouched, yet the checker still answers correctly, because SPDI represents `o.` and the sequence rung decides it:

```
Genome   g.100del vs g.105del  -> NormalizedMatch   (normalization decided it)
Circular o.100del vs o.105del  -> SequenceMatch     (SPDI decided it)
GenomeRing / Supernumerary     -> NotEquivalent     (nothing decided it)
```

So the defect is the *conjunction*, not the pass-through. That is what the fix keys on.

Worth noting the pass-through's comment cites #546, which is a **parse** issue — "each form parses to a typed AST and round-trips". It made no normalization ruling, so this was an unexamined default rather than a decided one.

### The fix

At the fall-through, decline when neither rung could examine a side. Deliberately narrow in three ways:

- **The two textual rungs are kept.** `Identical` and `AccessionVersionDifference` compare strings, need no provider, and are sound for a ring. Tests pin both.
- **The SPDI conjunct makes the predicate self-retiring.** If `hgvs_to_spdi` ever gains ring support, `edit_triples` starts succeeding, the sequence rung begins deciding these pairs, and this decline stops firing on its own instead of masking the new capability.
- **`RnaFusion` is excluded on purpose.** It shares both blindnesses, but no pair of fusion spellings denoting one fusion has been exhibited, so widening the decline there would refuse pairs the checker answers today on no evidence any of those answers is wrong. That boundary is pinned by its own test naming what would justify changing it.

Two rings on plainly disjoint spans are **also** declined, and that is the reading rather than over-breadth: the checker has no way to distinguish that pair from the respelled one — the same evidence (none) underlies both — so answering `NotEquivalent` there would be right by luck while being wrong on the other.

## Item 3: ruled, and pinned as a negative

`general.md:58`'s self-cancellation rule does not reach a ring's `::`-joined segments. Measured escape:

```
REJECTED  g.[100_200del;150_250dup]      E3006 self-cancelling (overlap 150-200)
ACCEPTED  g.100_200del::150_250dup
ACCEPTED  g.100_200del::100_200dup       <- identical spans, the strongest shape
ACCEPTED  g.[100_200del::150_250dup]sup
```

The argument, recorded as `rulings[self-cancelling-across-ring-junctions]` with all five citations quote-verified against the spec checkout at build time:

`general.md:58` is a four-space-indented sub-bullet of `general.md:56`, the **prioritisation** bullet, and its sibling `:57` is unambiguously a prioritisation rule ("when a variant can be described as a duplication or an insertion, prioritisation determines it should be described as a duplication"). So `:58` inherits `:56`'s antecedent — *"when a description is possible according to several types"* — and works by **redirecting** to the preferred single-type description. `DNA/complex.md:5` defines complex as a range of changes that **cannot** be described as a basic type, so a ring has nothing to redirect to: a `delins` over the union span describes a *linear* product and discards the join, which is the whole content of the ring description.

`DNA/complex.md:130` then settles the two operators directly, on identical member content:

> "::" is used to indicate the join, instead of ";" to describe two not connected deletions.

The same two deletions spelled with `;` denote a linear chromosome missing both ends; spelled with `::` they denote a ring. A `::` composite is therefore not reducible to its member set the way a `[a;b]` allele is — which is the one reading extending the validator would encode.

The ruling ratifies shipped behaviour, so it moves no row.

### Pinning the negative is the actual deliverable

Pointing `detect_self_cancelling_pair` at ring segments *looks* free: a well-formed ring's segments are telomere-anchored and disjoint by construction — `complex.md:127` and the `sup` form at `:161` are the spec's only published ring shapes — so the check would never fire on a legitimate ring. It would pass every existing test while encoding the reading `complex.md:130` forecloses, and E3006's repair hint ("rewrite as a single delins") is actively wrong advice for a ring.

So the guard was verified non-vacuous by **applying that change**: `an_overlapping_ring_is_not_rejected_as_self_cancelling` turns red and its message names the ruling record. The mutation was then reverted (`src/hgvs/parser/variant.rs` is byte-identical to #1583's tip).

### What the ruling does not settle — recorded, not implied

Three of the rings ferro accepts are malformed for reasons unrelated to self-cancellation: a `dup` segment contributes no break junction (`complex.md:39`), neither segment is telomere-anchored, and `g.150_250dup::100_200del` additionally violates the `pter`→`qter` segment ordering of `complex.md:51`/`:53`/`:55`. Those need a ring **well-formedness** rule. `:58` is the wrong instrument for them, which the non-overlapping case (`del::300_400dup`) proves: an overlap predicate accepts it no matter how it is wired.

`no_ring_wellformedness_rule_yet_so_malformed_rings_are_still_accepted` pins that gap as the status quo so it is visible and counted, and says in its own failure message that a future rejection is the signal to move the rows into the new rule's tests rather than to reopen this ruling. **I have not filed that issue** — say the word and I will.

## Testing

- Full suite: **9,795 passed, 0 failed, 150 skipped — nothing re-blessed.** For a change that alters an output verdict, existing expectations already agreeing is the signal worth quoting.
- 24 `issue_1578*` tests green (12 new, 12 from #1583); `ruling_records_are_intact` green with the new record pinned in `RULING_STATUSES`.
- `generate_spec_fixture` succeeds, which is what verifies the ruling's five citations still quote the spec at submodule `6f85311`.

### One bookkeeping update the rebase surfaced

`main` gained #1579's `clause_ruling_index` guard after this branch was cut, and it went red on the rebase — correctly: a new ruling record must also appear in that module's embedded clause→record index. Three of its assertions fired, the census pin moved `(10, 6, 4) → (11, 7, 4)`, and the generated block was replaced with the one the guard itself prints. That update is amended into the ruling commit, since it is part of the record's bookkeeping rather than a separate change.

Worth noting as the "CI tests the merge, not your head" case from the repo's `CLAUDE.md`: this branch was green against its original base and would have gone red in CI purely because `main` moved.

### Two vacuity traps found and fixed while writing these

Both had produced a green test that proved nothing:

- `spdi_apply_declines_every_ring_shape` asserted only the generic prefix `"cannot apply to reference"`, which `apply_to_reference` **also** emits for a reference-base mismatch — and the fixture's `ACGT` filler put `T` at position 500, so the control input `g.500A>T` failed too. The provider is now all-`A`, each test runs a positive control first, and the assertions name the ring-specific reason.
- The item-3 tests all passed on first run, hence the mutation check above.

## Review fixes applied

Four findings from the CodeRabbit CLI lens. Three fixed, one rejected.

### Fixed (major): the equivalence fix missed a ring **inside a cis allele** — the one shape this PR says a kind-matching arm can miss

`undecidable_reason` matched on the *top-level* variant kind, so a cis allele carrying a ring fell through and was answered:

```
before:  ANSWERED  level=NotEquivalent   <-- a positive claim no rung examined
after:   DECLINED  … cannot decide equivalence … NC_000022.11:g.100_101del::200_201del
```

That is the exact unsoundness this PR exists to remove, one level down — and it is an internal inconsistency rather than a hypothetical, because `issue_1578_followup_ring_declines.rs` already builds that pairing and requires `spdi::apply`, `vcf::from_hgvs` and the projector to decline it. This module's own header calls it *"the one that walks a ring through the hand-rolled `Allele` recursion rather than meeting it at the top level, and therefore the one a kind-matching arm can miss."*

`undecidable_reason` now checks SPDI **first at every level** (preserving the self-retiring property) and then recurses into allele members. Two regression tests: the decline, and a control that an all-ordinary cis allele is *still answered*, so a recursion that declined on structure alone could not pass. Control: removing the recursion fails exactly the new decline test.

Not reachable via `parse_hgvs` — measured, `g.[…::…;…]` is rejected — but reachable by construction, which is how this PR's own tests build it.

### Fixed (major): seven rationale citations were never verified against the spec

The rationale cited `complex.md:39`, `:51`, `:53`, `:55`, `:113`, `:117` and `:161`, none of which were registered in the record's `clauses` array — so the generator never checked them. `CLAUDE.md` is explicit that this is the point of registering: *"a submodule bump that moves a clause fails the build instead of leaving the citation pointing at unrelated prose."* All seven are now registered with verbatim quotes extracted programmatically from the spec (12 clauses total), and the rendered clause index regenerated. The generator validates them; control from earlier in this review: changing one word of a quote makes it exit 1.

### Fixed (minor): a control the comment promised and the code did not make

`projection_declines_every_ring_shape` said *"a plain genomic member, **and a legal genomic allele**, both project through this empty-cdot harness"* — but only asserted the member. Since the ring shapes are themselves alleles, a projector that declined every `Allele` would have passed. Added `legal_allele()`, with both members stating `A` to match the all-`A` contig, for the reason `provider()` already documents.

### Rejected (major): "remove `no_ring_wellformedness_rule_yet_so_malformed_rings_are_still_accepted`"

Declined. That test is the mechanism this repo uses to keep an open gap visible, and the ruling's own status pin names it: *"The negative is pinned by `issue_1578_followup_self_cancelling_rings.rs` so nobody closes the escape the cheap way by pointing `:58` at rings."* Deleting it means a later change that makes `:58` reject those rings passes silently and erodes the ruling's scope with nothing failing. Keeping it is what forces the conversation when the well-formedness rule lands.

### Testing after the fixes

- Full `it` suite: **5,588 passed**, 0 failed.
- `cargo clippy --all-features --all-targets -D warnings` and `cargo fmt --check` clean.
- `generate_spec_fixture` exits 0, i.e. all 12 citations verify against the pinned submodule.

Representation-Change: none. No file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched — the fix is confined to `src/equivalence/`, which is not a watched directory, and the item-3 ruling ratifies existing behaviour without moving any normalized output.

### DIFF
--- a/src/equivalence/checker.rs
+++ b/src/equivalence/checker.rs
@@ -244,6 +244,22 @@ impl<P: ReferenceProvider> EquivalenceChecker<P> {
                     .with_note("Variants produce the same resulting sequence"));
             }
 
+            // #1578 follow-up: `NotEquivalent` is a *positive claim* that two
+            // descriptions denote different variants, and it is only sound if
+            // some rung above actually examined them. For a structural
+            // rearrangement both deciding rungs are blind at once — see
+            // `undecidable_reason` — so falling through here would answer a
+            // question nobody asked. Decline instead, which is what the other
+            // three modules that hand-roll `Allele` recursion already do with a
+            // ring (`spdi::apply`, `vcf::from_hgvs`, `project::projector`).
+            for side in [&norm1, &norm2] {
+                if let Some(reason) = self.undecidable_reason(side) {
+                    return Err(FerroError::UnsupportedVariant {
+                        variant_type: reason,
+                    });
+                }
+            }
+
             Ok(EquivalenceResult::new(EquivalenceLevel::NotEquivalent)
                 .with_normalized(norm_str1, norm_str2)
                 .with_note("Variants do not normalize to the same form"))
@@ -354,6 +370,55 @@ impl<P: ReferenceProvider> EquivalenceChecker<P> {
             (Some(a), Some(b)) => a.eq_ignore_ascii_case(&b),
             _ => false,
         }
+    }
+
+    /// Why no rung can decide equivalence for `variant`, or `None` if some rung
+    /// can.
+    ///
+    /// The checker decides in rungs, and a structural rearrangement defeats both
+    /// of the two that compare *meaning* rather than text:
+    ///
+    /// - [`Normalizer::normalize`] returns a genome ring and a `sup` marker
+    ///   unchanged (the pass-through arms in `src/normalize/mod.rs`), so
+    ///   comparing normalized forms degenerates into the string comparison the
+    ///   `Identical` rung already made; and
+    /// - [`hgvs_to_spdi`] cannot represent either, so [`Self::edit_triples`]
+    ///   declines and the `SequenceMatch` rung (#1158) cannot fire either.
+    ///
+    /// With both blind, `NotEquivalent` would be asserted having evaluated
+    /// nothing. The two *textual* rungs — `Identical` and
+    /// `AccessionVersionDifference` — are unaffected and still answer, because
+    /// they need neither normalization nor a provider.
+    ///
+    /// **The SPDI conjunct is not redundant, and it is what keeps this honest.**
+    /// It makes the predicate self-retiring: if `hgvs_to_spdi` ever gains ring
+    /// support, `edit_triples` starts succeeding, the `SequenceMatch` rung
+    /// begins deciding these pairs, and this decline stops firing on its own
+    /// rather than masking the new capability. `Circular` (`o.`) is the live
+    /// proof that the conjunct matters: its normalization is a pass-through too,
+    /// yet SPDI represents it, so it is decided by the sequence rung and must
+    /// never reach this decline.
+    /// **Scoped to the two shapes a defect was measured on.**
+    /// [`HgvsVariant::RnaFusion`] shares both blindnesses and is deliberately
+    /// *not* listed: no pair of fusion spellings denoting one fusion has been
+    /// exhibited, so widening the decline to `::` transcript fusions would
+    /// refuse pairs the checker answers today on no evidence that any of those
+    /// answers is wrong. Add it when a wrong answer is measured, with the pair
+    /// that measures it.
+    fn undecidable_reason(&self, variant: &HgvsVariant) -> Option<String> {
+        let structural = matches!(
+            variant,
+            HgvsVariant::GenomeRing(_) | HgvsVariant::Supernumerary(_)
+        );
+        if !structural || self.edit_triples(variant).is_some() {
+            return None;
+        }
+        Some(format!(
+            "{variant}: cannot decide equivalence — normalization passes this \
+             description through unchanged and SPDI cannot represent it, so \
+             neither the normalized-form nor the resulting-sequence comparison \
+             examined it"
+        ))
     }
 
     /// Project a variant to the SPDI primitive edits that make up its resulting
--- a/src/equivalence/checker.rs
+++ b/src/equivalence/checker.rs
@@ -406,11 +406,31 @@ impl<P: ReferenceProvider> EquivalenceChecker<P> {
     /// answers is wrong. Add it when a wrong answer is measured, with the pair
     /// that measures it.
     fn undecidable_reason(&self, variant: &HgvsVariant) -> Option<String> {
-        let structural = matches!(
+        // SPDI first, at every level: if `edit_triples` succeeds the `SequenceMatch`
+        // rung can decide, so there is nothing undecidable to report. Keeping this
+        // ahead of the structural test is what makes the predicate self-retiring —
+        // ring support in `hgvs_to_spdi` silently switches these pairs back to being
+        // answered instead of declined.
+        if self.edit_triples(variant).is_some() {
+            return None;
+        }
+        // A cis allele carrying a ring is the same blindness one level down, and it
+        // is a shape this codebase constructs deliberately —
+        // `issue_1578_followup_ring_declines.rs` pairs a legal genomic member with a
+        // ring and requires `spdi::apply`, `vcf::from_hgvs` and the projector to all
+        // decline it. Without this recursion the equivalence checker was the one
+        // module that answered it: two such alleles compared `NotEquivalent`, which
+        // is a positive claim that no rung examined.
+        if let HgvsVariant::Allele(allele) = variant {
+            return allele
+                .variants
+                .iter()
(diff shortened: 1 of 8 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #1591, opened 2026-08-10
test(normalize): pin the two worked examples that explain the weight bound

nh13:
Representation-Change: none. Two tests and a doc module; no code path is altered and no normalized output moves.

## Why

`canonicalize_from_sequence` carries a bound — `derived_columns > changed_columns_of_edits(&edits)` — stated in its own comment as *"a canonicalization may re-partition and re-type the change; it may not describe **more** change than the input already did."*

It is the single largest confluence blocker measured in this codebase, and simultaneously the thing standing between several inputs and the form `DNA/delins.md:47` recommends. Which of those it is depends entirely on which example you look at — and until now the two faces were only ever discussed one at a time, in prose, in documents outside the repository. Every attempt to reason about the bound has started by reconstructing them.

Both are pinned here, against one reference, so the contrast is a screenful.

## The measure

A description's weight is `sum over members of max(ref_len, alt_len)`. The bound compares the *derived* partition's weight against the weight of the partition the *input* spelled, and refuses the derivation when it is heavier — handing the variant back to the per-member pipeline, which never re-aligns across members.

Two consequences follow from the measure alone, and the two examples show one each.

## Example 1 — refused, and NOT a degradation

`TEMPLATE:g.[2_3insA;12del]` — an insertion and a deletion whose lengths cancel. A re-derivation across them can place only one contiguous gap, so the offset reads as a run of substitutions and marks columns changed that the input left alone. The bound refuses it; the per-member pipeline answers; the output is the input.

This row is pinned **with its adjudication**, because the obvious reading of it is wrong. The block is `TTTTTTTAAT` → `ATTTTTTTAA` — equal length, so the column correspondence is unique and *which columns changed* is a fact rather than a choice. They are `{0, 7, 9}`. `DNA/delins.md:15` types each single changed column as a substitution; `:17` says variants separated by one or more nucleotides are described individually. **So the derived form is what the spec's stated rules produce, and the input is a spelling they never generate.**

A survey of all twenty rows the bound's removal moves found **four** of this shape — every one an equal-length block whose no-bound output the clauses produce verbatim — **zero** rows the clauses contradict, and ten on which `:15`/`:16`/`:17` have no defined input at all because the block is unequal-length.

The word "shredding", used of this row in earlier working notes, was a misreading of an alignment artefact for a description defect. What genuinely remains is the residue the spec does not address: in a repetitive context, shifting by one base still matches at most positions **by coincidence**, so an insertion-plus-deletion is column-wise indistinguishable from a few scattered substitutions. The clauses mechanically produce the substitutions. Whether that is *meaningful* — one of the four design values at `background/basics.md:38`, in a list from which minimality is absent — is a question for the committee, not for a bound in this crate.

## Example 2 — refused, and it is the cost

```
pos        29  30  31  32  33
ref         C   A   C   G   T          (asserted by the tests)
observed    A   A   C   ACATACTG
            ^   \_____/   ^
         changed   kept  changed
```

`TEMPLATE:g.29_33delinsAACACATACTG` and `TEMPLATE:g.[29C>A;32_33delinsACATACTG]` are two spellings of one variant. The `AC` at 30–31 survives in the split **only because the payload happens to carry `AC` at that offset** — five reference bases become eleven, so no column correspondence forces it. That is precisely the construction `DNA/delins.md:46` builds, and `:47` answers it: *"**The "delins" format is recommended**"*.

```
split   29C>A                  max(1,1) =  1
        32_33delinsACATACTG    max(2,8) =  8     total  9
span    29_33delinsAACACATACTG max(5,11) = 11    total 11
```

`11 > 9`, so the bound refuses the span. Both spellings are fixed points, and one variant has two normalized representations — the non-confluence reported as #1421.

## The arithmetic is pinned as arithmetic

Because it is what makes the cost **structural rather than a threshold**: a span covers the block in one member, so `max(ref, alt)` for the whole block can never be less than the sum of the maxima of a split of that block. **Any input-relative bound refuses every merge of this shape**, at any setting. That is worth a test rather than a paragraph, because every discussion of tuning the bound has had to rediscover it.

## Two deliberate choices

**The sequence claim uses the SPDI applier, not `EquivalenceChecker`.** `tests/it/common/cis_apply_oracle.rs` converts each member to its SPDI triple and splices the reference independently of the normalizer. A sequence identity checked with the normalizer's own equivalence machinery cannot separate "these agree" from "the same code decided both" — and a sibling file in this suite documents `EquivalenceChecker` as circular for exactly this purpose.

**The reference bases are asserted, not quoted.** Every argument above is an argument about `TEMPLATE:g.29_33` reading `CACGT`; a reference that did not carry it would make these rows pin something other than what the module says they pin.

## Assert-then-flip

All five expectations are today's behaviour, asserted rather than `#[ignore]`d, so the day it moves this file has to be read. Each failure message names the form the row is expected to move **to** — Example 2's points at `delins.md:47` and the decided ruling `delins-merge-vs-individual-gap-two-or-more`, and says to declare the representation change.

Hermetic: synthetic `TEMPLATE` provider, no manifest, no fixtures.

### DIFF
--- a/tests/it/main.rs
+++ b/tests/it/main.rs
@@ -455,5 +455,6 @@ mod trans_mosaic_mixed;
 mod uta_loader_tests;
 mod variantvalidator_tests;
 mod web_service_tests;
+mod weight_bound_worked_examples;
 
 mod reported_confluence_pairs;
--- a/tests/it/weight_bound_worked_examples.rs
+++ b/tests/it/weight_bound_worked_examples.rs
@@ -1,0 +1,314 @@
+//! The two worked examples that explain the weight bound, side by side.
+//!
+//! # Why this file exists
+//!
+//! `canonicalize_from_sequence` carries a bound — `derived_columns >
+//! changed_columns_of_edits(&edits)` in `merge.rs` — whose comment states it as
+//! *"a canonicalization may re-partition and re-type the change; it may not
+//! describe **more** change than the input already did."*
+//!
+//! It is the single largest confluence blocker measured in this codebase, and
+//! also the thing standing between several inputs and an answer the spec
+//! recommends. Whether it should exist at all is an open question. What kept
+//! making that question expensive to ask is that its two faces were only ever
+//! discussed one at a time, in prose, in documents outside the repository. This
+//! file pins both, against the same reference, so the next reader gets the
+//! contrast in one screen and does not have to reconstruct it.
+//!
+//! **Nothing here asserts that the bound is right or wrong.** Each row pins what
+//! ferro does today, with the arithmetic that produced it and the clause that
+//! bears on it. If the bound is removed or narrowed, these tests fail, and their
+//! failure messages say which way to read the change.
+//!
+//! # The measure
+//!
+//! A description's weight is `sum over members of max(ref_len, alt_len)`. The
+//! bound compares the weight of the *derived* partition against the weight of
+//! the partition the *input* spelled, and refuses the derivation when it is
+//! heavier — handing the variant back to the per-member pipeline, which never
+//! re-aligns across members.
+//!
+//! Two consequences follow from the measure alone, and they are what the two
+//! examples below show:
+//!
+//! * a re-derivation that marks *more columns changed* than the input is
+//!   refused — which is the protection;
+//! * **a span outweighs a split whenever the split keeps reference bases the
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai pause

PR #1593, opened 2026-08-10
fix(changelog): make the audit agree with the checker, and anchor its range to the merge ref

nh13:
#1560's `Changelog grouping audit` fails CI on trailer forms `CONTRIBUTING.md` documents as **correct**, in both directions. Once such a commit is on `main` the job is red for **every** open PR until the next release tag, and no PR author can fix it. Found by an adversarial review of #1560 after it merged.

## The audit and the checker had to agree, and did not

The audit's decline rule was made deliberately *coarser* than the checker's — first word, punctuation stripped, no terminator logic — on the reasoning that "a second opinion derived the same way is not a second opinion". That reasoning confused **deriving the answer differently** with **giving a different answer**.

`CONTRIBUTING.md:103-105`: *"A comma is not a terminator … ('none, except two rows') — that reads as a description of a move, and is filed as one. Same for `no rows move` … Both err toward listing a change rather than hiding one."*

So release-plz files both as real changes, by design — and the audit called them declines:

```
3 commits in v0.0.1..HEAD; 2 listed as representation changes
f27dd04a is filed under Representation changes but its trailer opens with a decline: 'no rows move'
986e4ccd is filed under Representation changes but its trailer opens with a decline: 'none, except two rows that merge'
exit=1
```

`check_representation_change.py` passes both. **No trailer text satisfied both halves.**

The reverse bit too. `split()[0]` treats `none.Tests` as one word, so an ordinary missing space flipped the verdict the other way:

```
'none.Tests only.'                  checker declines()=True   audit opens_with_a_decline()=False
'none—no watched file is touched'   checker declines()=True   audit opens_with_a_decline()=False
```

The rule now reads the verdict and admits a terminator from `.;:—–` (comma still excluded), agreeing with the checker in verdict while **sharing no code with it** — `test_the_rule_shares_no_vocabulary_with_the_checker` still passes. What stays independent, and is where the value always was: this check renders the changelog through real git-cliff and reads the result. A disagreement is not a second opinion, it is an unsatisfiable build.

Pinned by a new `test_the_audit_and_the_checker_agree_on_every_documented_form` over 17 forms, and end-to-end through git-cliff all three previously-failing trailers now exit 0.

## Three more from the same review

**`trailer_value` regressed #1573's column-0 rule.** It allowed `^[ \t]*` before the token — what git and git-cliff treat as a *continuation*. A PR description quoting the declining example above its own real disclosure diverged:

```
checker find_declarations (column-0 only): ['3 rows move, 2 respell']
audit  trailer_value      (allows indent): 'none'      -> filed as a decline -> exit 1
```

**The audited range was resolved after the soft reset.** `base.sha` is the base tip as of the last PR event while the merge ref uses the *current* base, so once a release lands, an un-resynchronized PR resets to a commit from which the new tag is unreachable — `describe` then returns the **previous** tag and the audit silently widens back over an entire released cycle, re-judging merged work the PR author cannot touch. Measured on a simulated merge ref with a release landing after `base.sha`:

| | audited range | commits |
|---|---|---|
| before | `v0.13.1..HEAD` | **3** — including `fix: released work`, `feat: more released work` |
| after | `v0.14.0..HEAD` | **1** — the PR's own squash preview |

The tag is now captured *on the merge ref*, before the reset, and passed to the audit as `--range` via `$GITHUB_ENV`. Guarded, so a tagless history still falls back to the script resolving its own range.

**`latest_tag` raised `CalledProcessError`** on a tagless history — the condition the workflow already guards its own `describe` against. It now raises a `RuntimeError` naming the fix.

## Every fix has a control that fails without it

Including three mutants that previously passed **both** the test and `actionlint`:

| mutant | before | after |
|---|---|---|
| `# git reset --soft "$PR_BASE_SHA"` (commented out) | passed | **fails** |
| `# was: …` + `git reset --soft HEAD~0` | passed | **fails** |
| second `printf … > squash-message.txt` clobbering the first | passed | **fails** |
| `git reset --hard` instead of `--soft` | fails | fails |
| tag resolved *after* the reset | n/a | **fails** |

The first two worked purely by matching **comment text**, so the assertions now read the step with `#`-comment lines stripped — and that stripping is itself asserted to have removed something, since a no-op strip protects nothing. The third was `[^>]*` being positional, so the message file is now asserted to be written exactly once.

Worth recording: my first attempt at the `--hard` control reported the assertion *passing* a hard reset, which would have meant a useless assertion. The control was invalid — it rewrote the phrase `reset --soft` in the surrounding comment rather than the command. The harness now asserts the anchor occurs exactly once before replacing it. A bad control is indistinguishable from a passing one.

## Testing

- **105 tests pass** in the two trailer/grouping modules.
- `ruff check` / `ruff format --check` clean; `actionlint` clean; `ci.yml` parses.
- The audit is clean against real history: `10 commits in v0.13.1..HEAD; 2 listed as representation changes`.
- The shipped workflow script executed against a simulated merge ref under `bash -euo pipefail`: exit 0, correct range, PR tree intact.

`CLAUDE.md`'s description of the audit is corrected in step — it stated the "shares no code" property in a way that invited exactly this over-reading.

Representation-Change: none. CI scripts, one workflow job, their tests and documentation; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -272,19 +272,42 @@ jobs:
           PR_BASE_SHA: ${{ github.event.pull_request.base.sha }}
         run: |
           printf '%s (#%s)\n\n%s\n' "$PR_TITLE" "$PR_NUMBER" "$PR_BODY" > squash-message.txt
+          # Resolve the tag BEFORE the reset, and hand it to the audit explicitly.
+          #
+          # `base.sha` is the base tip as of the last PR event, while the merge ref GitHub
+          # builds uses the *current* base — so after a release lands, an un-resynchronized
+          # PR resets to a commit from which the new tag is unreachable. `describe` would
+          # then answer with the PREVIOUS tag and silently widen the range back over an
+          # entire already-released cycle, re-judging merged work the PR author cannot
+          # touch. Measured: pre-reset `describe` gives v0.14.0, post-reset gives v0.13.1.
+          # Guarded because `describe` exits 128 on a tagless history and must not be what
+          # fails the job under `set -e`.
+          if AUDIT_TAG=$(git describe --tags --abbrev=0 2>/dev/null); then
+            echo "AUDIT_RANGE=$AUDIT_TAG..HEAD" >> "$GITHUB_ENV"
+          else
+            echo "no tag reachable from the merge ref; the audit will resolve its own range"
+          fi
           git reset --soft "$PR_BASE_SHA"
           git -c user.name='github-actions' -c user.email='github-actions@github.com' \
             commit --allow-empty --no-verify --quiet --file squash-message.txt
           echo "auditing this squash-equivalent commit:"
           git --no-pager log -1 --format='%H%n%s'
-          # Guarded: `describe` exits non-zero on a tagless history, and this echo is a
-          # debugging aid — it must not be what fails the job under `set -e`.
-          if tag=$(git describe --tags --abbrev=0 2>/dev/null); then
-            echo "audited range ($tag..HEAD) now contains:"
-            git --no-pager log --format='  %h %s' "$tag..HEAD"
+          if [ -n "${AUDIT_TAG:-}" ]; then
+            echo "audited range ($AUDIT_TAG..HEAD) now contains:"
+            git --no-pager log --format='  %h %s' "$AUDIT_TAG..HEAD"
           fi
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

PR #1595, opened 2026-08-10
fix(parser): require ring `::` segments to be ordered deletions

nh13:
Closes the gap `rulings[self-cancelling-across-ring-junctions]` named and deliberately did not fill. **Stacked on #1589** (base `fix/issue-1578-followups-ring-recursion`) — review that first; if it squash-merges, this needs a rebase onto `main`.

Follow-up to https://github.com/fulcrumgenomics/ferro-hgvs/issues/1578#issuecomment-5239725830.

## The gap

#1589 ruled that `general.md:58` does not cross `::`, and recorded that three rings ferro accepted were malformed for unrelated reasons. `:58` is an *overlap* predicate, so it could never have caught the non-overlapping one however it were wired:

```
g.100_200del::150_250dup     dup contributes no junction
g.100_200del::300_400dup     same, and non-overlapping
g.150_250dup::100_200del     segments out of pter->qter order
```

Ferro's ring grammar turned out to be near-unconstrained: measured at this commit it accepted `del::dup`, `del::inv`, `del::sub`, `sub::sub`, `del::identity`, `del::ins`, descending order, overlapping and identical segments, and 2–4 segments. Only `cen` anchoring was refused (a separate pre-existing gap, untouched here).

## Two rules, and why the second one took a second look

**Ordering** — `DNA/complex.md:51` (*"Break point location is determined by the first break point encountered, i.e. `pter` of the chromosome is to be listed first"*), `:53` (breakpoints *"are listed in order of occurrence from `pter` to `qter`"*), `:55` (*"variant descriptions are always in the forward orientation"*). Explicit and directly on point.

**Junction-contributing segments** — this is the one worth reading carefully, because my first justification for it was too weak and it broke a committed control before I found the real argument.

`DNA/complex.md:39` says *"a double colon is used to designate break point junctions creating a ring chromosome."* On its own that is an inference from a definitional sentence. What settles it is **`:60-64`, the committee adjudicating `::` itself**:

> **The description of translocations has changed.** In the original proposal (SVD-WG004), **one identical derivative chromosome** would receive **two different descriptions**, depending on whether it was identified in a balanced or an unbalanced case. In a balanced case, the description would use a `"::"` format joining the breakpoints, while in an unbalanced case, the description would use a `"delins"` format. HGVS recommendations try to avoid such conflicts wherever possible. **HGVS, therefore, recommends to describe translocations exclusively using a `"delins"` format.**

`::` as a general join operator was **withdrawn**, on a confluence argument — one chromosome, two spellings.

Classifying every `::` occurrence in `complex.md` confirms the consequence. Every HGVS-*shaped* `::` string is either marked `<code class="invalid">` or labelled "**ISCN**" for comparison; line 72 says the `::` translocation forms were "corrected" *to* `delins[...]`. In current HGVS notation `::` survives in exactly **two** places — `:127` and `:161`, both rings, both `del::del`. Every other rearrangement is `delins[...]`, `ins[...]` or a `;` cis allele, including ISCN structures whose *own* notation is full of `::` (e.g. `inv(2)(pter->p22.3::q31.1->p22.3::q31.1->qter)` renders as `g.[32310435_32310710del;32310711_171827243inv;insG]`).

And `:12` gives the governing principle: *"while ISCN describes the structure of the resulting chromosome(s), HGVS describes the **variant(s) detected**."* A ring's segment list is a list of detected variants — two terminal deletions — with `::` recording that they are connected (`:130`).

**Consistency clincher:** `try_parse_genome_ring` already refuses a cross-accession segment, with the comment *"a cross-chromosome `::` is ISCN2016, removed in ISCN2020, and spec-invalid."* That is the same withdrawal. Ferro enforced one half of it and not the other.

## A defect caught after the first push: `delN[]` is a deletion

Worth stating rather than burying, because it is the worst case to get wrong here. The junction predicate first matched only `NaEdit::Deletion`, so it refused `NaEdit::NPaddedDeletion` — `g.100_200delN[15]::300_400del`. But `delN[]` **is** a deletion, one whose removed size is a count of unknown bases, and *"breakpoint not sequenced"* is the annotation on **both** ring shapes the spec publishes (`:128`, `:163`). The rule was rejecting the form the spec's own examples describe in prose, and its diagnostic misdescribed the input as "a duplication, inversion, substitution or identity segment" when it was none of those.

Found by enumerating `NaEdit`'s twenty variants instead of reasoning about the ones I had in mind. Fixed, with tests over `delN[]` in first, second, both and `sup`-wrapped positions, and the diagnostic rewritten to name both accepted spellings.

The same enumeration turned up the row that most vindicates the rule: a **position-only** segment (`g.pter_100::200_qter`) is rejected, and that is the withdrawn ISCN2016 join itself — `:72` marks it `<code class="invalid">` and "corrected" to `delins`, and `:84` explains it meant the deletion was *implied* by the coupling. Pinned with its own test and citations.

Two other shapes checked and left alone: `Mu::Unknown` is unreachable through the ring grammar (a `?` edit fails to parse inside a `::`-join), so the permissive arm for it guards programmatic construction only and is documented as untested rather than claimed as covered; `con` and methylation segments are rejected as non-deletions.

## Both rules reject only what is *definitely* wrong

Ordering compares an **upper** bound of the later segment's start against a **lower** bound of the earlier one's, so no amount of uncertainty can trigger it. That is what keeps `:127`'s own `pter_(12200001_14700000)del::(37600001_410000000)_qterdel` accepted — a naive integer compare would have had to pick an arbitrary representative or reject the spec's own example. `pter` is the axis minimum, `qter` the maximum, `cen` undecidable. The junction check skips a segment whose edit is `?`.

## Telomere anchoring is NOT enforced

A ring loses both telomeres, so a well-formed ring is anchored at `pter`/`qter`, and both published shapes are. But **no clause says so**, `:28` only defines what the markers mean, and `:13` warns a complex description can be *"literally correct"* yet meaningless. Enforcing it would be legislating from two worked examples plus biology — and the junction rule above is a live demonstration of how that goes wrong (see below). Recorded as `rulings[ring-telomere-anchoring]`, **`undecided`**, with `a_ring_with_no_telomere_anchor_is_still_accepted` pinning that an unanchored ring still parses so the question is not closed by drift.

The two ring records now both cite `complex.md:127`, one `decided` and one `undecided` — which is exactly the mixed-verdict case `clause_ruling_index` exists to surface, so citing `:127` for "rings are anchored" reads a ruling out of the undecided half. That is stated in its docs.

## Where the check runs — this placement is load-bearing

Outside `for_each_leaf`, because that walk re-wraps each segment as a standalone `g.` leaf and erases both the segment order and the ring identity the rules are about. And **after** the per-leaf rules, so a segment that is both an illegal edit form and a non-deletion reports the edit-form error, which is the more specific and actionable one.

Measured, not assumed: running the ring rules *first* turned **7 of `issue_1578_ring_validator_escapes`** red, because their fixtures pair an illegal edit form with an `insG` segment. It is also what keeps #1264's insertion-anchor rejections failing for their own reason rather than being silently absorbed.

## Three pre-existing controls used non-deletion ring segments

Each is updated rather than deleted, and each keeps its guarantee. **Flagging this prominently because one of them is @geoffjentry's, from #1578's own body.**

- **`a_ring_of_legal_edits_still_parses` (#1583)** — `dup::insG` → `del::del`. #1578 names the original as its *"Legal-edit control, verified to parse and round-trip"*, and it **remains** a legal-edit control: `dup` and `insG` violate none of the per-leaf edit-form rules it was chosen to exercise. It is not a well-formed *ring*. Its stated purpose — *"tightening the ring path could pass by rejecting every ring"* — is fully served by `del::del`, which still fails under a blanket rejection. Jeff, if you read `:39`/`:60-64` differently, this is the line to push back on and I'll ship ordering-only instead.
- **`a_ring_segment_with_flanking_anchors_still_parses` (#1264)** → `..._is_not_rejected_by_the_anchor_rule`. Its premise ("every ring insertion" must not be rejected) is now deliberately false, so "it parses" is unavailable as evidence. It asserts the *reason* instead — rejected by the ring rule, not the anchor rule — which is **stronger** than the acceptance it replaced, since acceptance could have come from the anchor rule silently not running.
- **#1589's two overlap tests** move to `del::del`, isolating the `:58` question from the junction rule. Had they stayed `del::dup` they would now pass by being rejected for the *wrong* reason. Its status-quo test is deleted, exactly as its own failure message instructed.

## A fourth pre-existing control, in a file this PR otherwise does not touch

`tests/it/fast_path_differential.rs` hard-codes two of the nine flipped spellings and labelled them
*"valid ring: both accept, must agree"* and *"valid `sup`-wrapped ring"*. Both now reject, so both
comments asserted the opposite of the behaviour.

The suite stayed green throughout, which is what made it silent: `divergence` treats `(Err, Err)`
as agreement, so flipping a row from accept-and-agree to both-reject-and-agree passes. The real
loss was coverage — after this PR **no** ring input in that corpus reached `divergence`'s
`(Ok(a), Ok(b))` arm, which is the one that catches a fast path accepting a ring while building a
*different* AST. Both-reject rows cannot pin an AST.

Fixed by correcting the two comments and adding `g.100_200del::300_400del` and its `sup` form
(both verified accepted) so the equal-AST direction is exercised again. Found by enumerating the
nine flipped strings and grepping the whole tree for each; this was the only file outside the diff
still making a claim about any of them.

## Blast radius, stated as the structural zero it is

Of the spec corpus's 934 rows, 25 contain `::`: 17 correctly-rejected cross-accession translocations, 5 RNA fusions, and **3 `GenomeRing` rows** — all `preserved`, all telomere-anchored ascending `del::del`:

```
NC_000022.11:g.pter_(12200001_14700000)del::(37600001_410000000)_qterdel
NC_000022.11:g.[pter_(12200001_14700000)del::(37600001_410000000)_qterdel]sup
chr22:g.pter_10000000del::40000000_qterdel
```

So **zero corpus rows change** — but that is a *structural* zero in this repo's sense, not evidence of safety: the corpus contains only well-formed rings, so it cannot exhibit a malformed one. The real evidence is the full suite passing with three controls' observables changed and nothing else re-blessed.

## The long-term fix this is an increment toward

`GenomeRing.segments` is `Vec<LocEdit<GenomeInterval, NaEdit>>`, so a malformed ring stays constructible in the type system and every consumer has to re-derive the rule — which is how #1578's four-module blindness happened in the first place. Narrowing that payload to deletions moves the guarantee from a parser check to the type. That refactor is next and deliberately not here, since it touches the public type.

## Testing

- Full suite: **10,013 passed, 0 failed, 150 skipped.**
- Re-run with all three seam oracles armed under CI's own exclusion filter
  (`FERRO_ASSERT_IDEMPOTENT` / `_REPARSE` / `_IN_BOUNDS`, minus `ORACLE_EXCLUDE`
  and `SWEEP_FILTER`): **9,895 passed, 0 failed.** Worth stating because this PR
  changes what parses, and the re-parse oracle is the one that would notice a
  normalized ring that no longer round-trips.
- `generate_spec_fixture` succeeds, which verifies the new record's four citations still quote the spec at submodule `6f85311`.
- `clause_ruling_index` regenerated (census `(15,10,5) → (16,10,6)`, multiply-named 10→12, mixed-verdict 6→8) and `ruling_citation_currency` green — the latter caught a real slip where my prose wrapped `**undecided**` onto the *decided* record's line.

Representation-Change: none. This is a parse-time rejection of descriptions with no defined HGVS meaning; it moves no normalized output, and no corpus row's normalized form differs.

**Rejection change, measured rather than asserted: 9 previously-accepted spellings are now refused, all 9 by the junction rule and 0 by the ordering rule.** Method: harvest every `::`-shaped HGVS string from the `origin/main` tree (53 distinct), re-parse each at this HEAD, and classify. The denominator is deliberately `main`'s tree and not this branch's — at HEAD there are 75 such strings, but 26 of them are test inputs *this PR adds*, so classifying those would be circular and inflates the count to 19. The rule is purely additive (`validate_ring_segments_are_wellformed` is appended last in `parse_variant` and only ever returns `Err`), so "refused with a ring diagnostic at HEAD" is equivalent to "accepted on `main`"; `main`'s own committed tests assert 4 of the 9 parse, and the `self-cancelling-across-ring-junctions` rationale names 3 more as accepted.

All 9 are the repo's own test fixtures. Three of them are the rings the ruling explicitly named (`g.100_200del::150_250dup`, `g.100_200del::300_400dup`, `g.150_250dup::100_200del`); the other six are `g.100_101dup::200_201insG`, `g.100_101insATG::200_201insG`, `g.100_200del::100_200dup` and the three `sup`-wrapped forms. **The ordering rule refuses nothing that previously existed** — it is exercised only by inputs this PR adds.

### DIFF
--- a/src/hgvs/parser/variant.rs
+++ b/src/hgvs/parser/variant.rs
@@ -7304,6 +7304,22 @@ pub fn parse_variant(input: &str) -> Result<HgvsVariant, FerroError> {
         Ok(())
     })?;
 
+    // Ring-level checks, LAST of the post-parse suite. Two placements matter:
+    //
+    // - Outside `for_each_leaf`, because that walk re-wraps each segment as a
+    //   standalone `g.` leaf, which erases both the segment *order* and the fact
+    //   that the leaves belong to one ring — the two things these rules are
+    //   about. Same reason `validate_no_self_cancelling` sits outside it.
+    // - *After* the per-leaf rules, so a segment that is both an illegal edit
+    //   form and a non-deletion reports the edit-form error. That is the more
+    //   specific and more actionable diagnostic, and it is what #1578's
+    //   error-parity tests pin: a ring segment must fail exactly like its
+    //   standalone spelling. Running the ring rules first would replace those
+    //   messages wholesale (measured: it turned 7 of
+    //   `issue_1578_ring_validator_escapes`'s tests red, since their fixtures
+    //   pair an illegal edit form with an `insG` segment).
+    validate_ring_segments_are_wellformed(&variant)?;
+
     Ok(variant)
 }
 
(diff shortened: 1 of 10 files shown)


nh13:
@coderabbitai pause

PR #1599, opened 2026-08-10
fix(normalize): gate the coding delins carve-out on the amino-acid precondition

nh13:
## The defect

`DNA/delins.md:17` says two variants separated by one or more nucleotides are described individually. `:18` excepts the pair that is separated by **one** nucleotide **and** *together affects one amino acid* — two conjuncts, not one.

`merge::coalesce_coding_frame_separation`, the pass that owns the length-changing block, shipped implementing only the first. Its gate was:

```rust
if !reading_frame || !length_changing || !pieces.windows(2).any(joins) {
    return None;
}
```

There is no codon test anywhere in it. `reading_frame` says an amino acid *exists* to speak of; it does not say the two variants share one. And `length_changing` is a **proxy** — it asks what kind of edit the block is, where the clause asks only which positions it touches. The equal-length sibling `apply_coding_codon_exception` does test the codon; the length-changing path never did.

## The fix

Gate each candidate merge on the clause itself: the merged piece's whole reference span must lie inside **one codon** of the reading frame, using the same `same_codon` arithmetic the sibling pass already applies. The predicate reproduces both of the spec's worked examples — `c.235_237delinsTAT` (codon 79, `delins.md:19`) and `LRG_199t1:c.145_147delinsTGG` (codon 49, `delins.md:42`).

**Edit type stays out of the predicate, deliberately — and since #1623 that is the *ruling*, not a preference.** Not "decline on a frameshift", not "decline on a duplication" — those are the same proxy pointed the other way. `codon-carve-out-shape-restriction` is now **`decided`: WIDEN** — the exception "applies wherever its stated precondition holds … REGARDLESS OF EDIT TYPE", on the ground that edit type is a property of the *spelling* while "together affecting one amino acid" is a property of the resulting sequence. This predicate is codon membership and nothing else, so it runs the same way. Two comments on this branch still called that record `undecided` (they predate #1623, which this branch was rebased onto); both are corrected. `length_changing` is kept purely as **routing** between the two passes (the triplet-precise one must keep owning the equal-length shape, and it runs after the 3'-shift where this one runs before), and is documented as routing rather than as part of the predicate.

## The precondition is not weakenable by edit type

Verified against the pinned spec checkout. The identical sentence appears verbatim in **nine** places, and the four *length-changing* pages among them state it exactly as the equal-length ones do:

| # | clause |
|---|---|
| 1 | `docs/recommendations/general.md:35` |
| 2 | `docs/recommendations/DNA/delins.md:18` |
| 3 | `docs/recommendations/DNA/substitution.md:17` |
| 4 | `docs/recommendations/DNA/deletion.md:19` |
| 5 | `docs/recommendations/DNA/insertion.md:20` |
| 6 | `docs/recommendations/DNA/inversion.md:21` |
| 7 | `docs/recommendations/DNA/duplication.md:23` |
| 8 | `docs/recommendations/RNA/delins.md:18` |
| 9 | `docs/recommendations/RNA/substitution.md:18` |

Supporting restatements: `DNA/delins.md:81` puts it as an exclusion — two variants separated by one or more nucleotides are described individually "(unless they together affect one amino acid)" — and `general.md:37` names it as something a tool has to *know*: "one needs to know whether the two variants are in a coding sequence and affecting one amino acid." `protein/delins.md:21` is the sole variant and carries **no** exception at all.

## Measurement

`examples/dump_normalized_corpus.rs`, 78,298 rows, both shuffle directions. **Not a structural zero**: the corpus varies reading frame (`c` axis `CDS_START = 1`; `cx` axis `CDS_START_MULTI = 4` on the multi-exon reference added by #1478), and it moved.

| comparison | rows moved |
|---|---|
| `main` → this branch | **1,581 (2.0%)** — 124 previously accepted, 1,457 previously not a fixed point |
| `main` → pass disabled outright (ablation) | 2,827 (3.6%) |
| this branch → pass disabled outright (ablation) | **1,372** |

Families that move: `del_plus_sub` 749/4608, `split_vs_spanning_delins` 516/9216, `dup_plus_sub` 316/4608. Every other family: 0. Direction is uniformly **merge → split** — a lone `delins` becomes a two-member allele.

That third row is the important one. **This is not equivalent to deleting the pass.** An earlier attempt at the same correction was blocked precisely because its narrowing measured byte-identical to disabling `coalesce_coding_frame_separation`; the codon predicate keeps merging **1,372 rows** the ablation loses, i.e. it keeps the half of the pass the clause licenses and drops the half it does not. Of #1484's three motivating shapes it **preserves** `NM_004006.2:c.4375_4376delinsAGTT` (span 2 positions, both codon 1459) and `NM_000088.3:r.4_6delinsacgu` (span `r.4-6`, codon 2), and drops only `NM_000143.3:c.44_47delinsATC` (span 4 positions, codons 15 and 16).

### Confluence rises

Both `cis_confluence_axis` census pins are re-blessed **upward**, which is the direction the pin demands:

| direction | converged | split 2 |
|---|---|---|
| 3' | 8,026 → **8,027** | 3,112 → 3,111 |
| 5' | 8,021 → **8,023** | 3,137 → 3,135 |

`declined`, `underdetermined` and `sequence_changed` are 0 on both sides. One and two are small numbers and worth saying plainly: the value of this change is conformance, not this census.

**This paragraph used to end "What the census rules out is the opposite reading — that restoring the precondition *costs* confluence. It does not, in either direction." That is true of *this* corpus and false in general — corrected below.** `cis_confluence_axis` holds member geometry, transcript geometry and scale fixed. The exhaustive spec corpus does not, and there the 3' direction loses one. Both measurements are correct about their own corpus; neither generalises to the other, which is why two of them exist. The claim is now scoped in the docstring itself rather than left to be re-read as a claim about the change.

### …and the harsher spec corpus loses one at 3'. Disclosed, with the rows named

`spec_conformance_axis`'s corpus is deliberately far harsher — 12,946 rows / 58,552 spellings, varying member geometry, transcript geometry and scale. Five of its figures moved, and its own assertion message forbids `converged` going down without a re-bless in the same commit *and* the newly-failing rows promoted to a named test. Both are in this commit.

| direction | figure | was | now | reading |
|---|---|---|---|---|
| 3' | `non_idempotent_outputs` | 7 | **4** | **improves.** Three rows fixed |
| 3' | `converged` | 9,140 | **9,139** | **regresses by one, net** |
| 3' | `split_two` | 2,435 | **2,436** | the same net one |
| 5' | `split_two` | 2,696 | **2,698** | **improves.** Two 3-output classes became 2-output |
| 5' | `split_three` | 204 | **202** | the same two classes |

Everything else is unchanged, including `outputs_leaving_the_transcript` at **371** — that is the pre-existing 3' baseline this branch does not touch, not a regression of its own.

**Three non-idempotent rows are fixed at 3'** (7 → 4). They are `scale-c3p-sep{120,128,136}-del-del`, and the mechanism is a direct confirmation of what `defect_non_idempotent_outputs.rs` diagnosed: their spanning `delins` was partitioned on pass one and *re-partitioned* on pass two, and the pass-two answer was shorter because it merged across codon boundaries the exception never reached. Remove those merges and pass one lands where pass two used to — the three pinned second-pass strings are **unchanged**, only the pass that produces them moved. The coding-axis onset sweep that found the class on five of six `ACGT` cores now finds it on **none** (`0 of 36` sweeps). The four `cds-end` families are untouched at both 3' and 5'.

**`converged` going down by one is a net figure, and the net hides the shape.** Measured by dumping both directions' full divergence lists on `origin/main` and on this branch and diffing the row ids — the census's own `DIVERGES` report is capped at 12 lines against ~2,743 non-converged families, so the ids are not recoverable from a CI log:

* **6 rows lost convergence** — three designs on each of the plus and minus coding multi-exon shapes: `s00-c3{p,m}-m3-all-del-p1-sep3`, `s00-c3{p,m}-m4-all-del-p1-sep2`, `s00-c3{p,m}-pair-del-del-p1-sep8`. In every one it is the **spanning-`delins` respelling** that moved and the authored multi-member spellings are byte-identical. Example: `c.24_33delinsAATTTTTA` reached `c.[24del;33del]` on `main` — the same answer its own members reach — and now stops at `c.[24_27delinsAA;32_33delinsTA]`, because a join it used to make spans two codons.
* **5 rows gained convergence** — `s01-c1-pair-del-del-p2-sep5` and `s01-c3{p,m}-m3-all-del-p1-sep2`, whose spanning `delins` used to be a fixed point *disagreeing* with its own members (`c.24_30delinsGATT` stayed whole; it now decomposes to the members' `c.[24del;30_31del]`), plus `scale-c3p-sep{120,128}-del-del`, which used to need a second pass to reach the answer their members reach.

Both counters therefore move for one reason, and it is not a representation choice left open: `general.md:35`'s second conjunct is either met or it is not. Where it is unmet, `general.md:34` governs. All six lost rows are churn rather than sequence loss, asserted through `denotation_of` rather than claimed: both competing outputs denote the row's own bases.

**Promoted, per the pin's own instruction**: `spec_corpus_regressions::the_codon_gate_splits_a_spanning_delins_its_own_members_do_not` pins all three designs on both strands — the unchanged member answer, the moved `delins` answer, that the two disagree, and that they denote the same sequence. A named committed test cannot be un-generated by a later edit to the corpus generator, which is the #1456/#1460/#1478 blindness class.

### The five `defect_non_idempotent_outputs` pins were flipped because the defect is FIXED, not because a number drifted

This is the distinction worth being explicit about, since a re-blessed pin and a fixed defect look identical in a diff. Five tests in that file assumed the `scale-*` instability exists. It does not any more, so each now asserts the fixed point and says in its own doc what it used to assert and why it changed. **None was deleted** — the repo's doctrine is that a fixed deviation stays visible so it cannot silently regress, the same discipline `KNOWN_DIVERGENT_INPUTS` implements by failing when a listed input *starts* matching the spec.

| test | was | now |
|---|---|---|
| `the_non_idempotent_output_is_always_the_spanning_delins_respelling` | 7 families × 1 moving spelling | the 4 `cds-end` families only; `SCALE_FAMILIES` dropped from `affected_families` |
| `every_scale_family_is_now_a_fixed_point` | *(new)* | every spelling of all three `scale` rows is its own fixed point, both directions |
| `the_scale_class_is_a_coding_axis_repartition_not_a_long_block_effect` | 5 coding onsets `[…Some(84)…Some(69)]` | `vec![None; 12]`; the trailing `!is_fixed_point(coding…)` inverts too |
| `the_scale_onset_is_not_the_corpus_scale_band` | bandless row non-idempotent vs banded row stable | both stable; the *refutation* survives — the knob that fixed it was codon phase, not length, exactly as that test argued |
| `pass_one_is_already_a_fixed_point_of_per_member_normalization` | 3 pinned **second**-pass answers | the same 3 strings, as **first**-pass answers |
| `a_block_crossing_an_exon_junction_is_not_split_so_the_defect_is_masked_there` | `(180, false, 6, false)` | `(180, false, 2, true)` — see below |

That last one is the only row whose direction reads backwards at a glance, so it is worth naming: 6 → 2 members is the *shorter* partition, i.e. the pass-two answer arriving on pass one. Nothing is being merged that was not merged before; the block is partitioned once instead of twice. The two junction-crossing rows are unchanged, so the corpus-design lesson the test was written for — a sweep confined to junction-crossing blocks measures zero and means nothing — stands untouched.

**Two of those tests were losing coverage silently and staying green, which is fixed here too.** `norm_cubed_equals_norm_squared_for_every_row_in_the_class` and `the_chain_never_changes_the_sequence_it_denotes` both iterate `non_idempotent_spellings(row, direction)`, and that list went empty for the three `scale` rows. Their anti-vacuity guard — whose own comment reads *"without that the test would pass vacuously the moment the defect is fixed"* — sits **inside** that loop, so it could not fire: three rows stopped being measured and nothing went red. Both now count iterations and assert the count outside the loop.

**And a test defect fixed independently of this PR.** `non_idempotent_spellings(row, …)[0]` reported "the defect got fixed" as `index out of bounds: the len is 0 but the index is 0`, which is exactly the illegible failure the repo forbids. It is now `sole_non_idempotent_spelling`, which asserts the length with a message naming the flip. Swept for siblings in the same file and folded in: two `corpus_cores(..).remove(n)` calls (now `core_at`, which names generator drift as the cause) and one `frame.exons()[0]` (guarded non-empty).

## Three mutalyzer rows leave agreement, and are annotated

`NM_000143.3:c.44_47delinsATC` (also `…delTGCGinsATC`, `…del4insATC` — one variant, three spellings) now normalizes to `c.[44_45delinsAT;49del]`. The `c.44_47` span is four positions, spanning codon 15 (`c.43_45`) and codon 16 (`c.46_48`), and no codon holds four consecutive positions at any phase, so `:18` cannot reach it and `:17` governs.

They carry a `spec_citation` in `cases.json` under the existing section `HGVS §delins (one-amino-acid exception)`, plus a new cluster `delins-one-amino-acid-precondition`. Precedence is **spec-explicit > mutalyzer**, and mutalyzer has no separation rule at all — it minimises a weighted description length — so its merge here is a different objective meeting rather than evidence about the clause. Note the mirror-image row already in the same fixture: `LRG_199t1:c.235_237delinsTAT`, where the span **is** one codon, ferro merges and mutalyzer splits, and the citation is the same section. One predicate, both directions.

`src/normalize/mod.rs`'s "How those three were adjudicated: they were not" paragraph — written when #1484 restored agreement on these rows — carries a correction, since it reads the exception off the distance conjunct alone.

## One protein consequence degrades on `main`, and #1590 clears it — measured, not assumed

Measured against the real prepared reference, `--axis p`:

| row | normalized | protein on `main` | on this branch | on this branch rebased on #1590 |
|---|---|---|---|---|
| `NM_000143.3:c.44_47delinsATC` | → `c.[44_45delinsAT;49del]` | `p.(Val15AspfsTer8)` | **`p.?`** | **`p.(Val15AspfsTer8)`** |
| `NM_004006.2:c.4375_4376delinsAGTT` | unchanged | `p.(Arg1459SerfsTer19)` | unchanged | unchanged |
| `NM_000088.3:r.4_6delinsacgu` | unchanged | `p.(Phe2ThrfsTer2)` | unchanged | unchanged |

Exactly one of the three loses its concrete consequence against today's `main`, and the cause is not this change: `Projector::project_cis_flag_and_protein` gates the combined-product path on `net_in_frame`, so a *net-frameshift multi-member* allele collapses to `p.?` regardless of spelling. The carve-out was incidentally masking that gap by keeping the allele single-membered. **PR #1590 closes it**: applying this branch's `merge.rs` change on top of #1590's head restores `p.(Val15AspfsTer8)`, so **zero `p.?` rows remain**. 

This PR is *not* stacked on #1590 — it branches from `main`, and the last column above is a measurement taken in a throwaway worktree rather than a dependency. If #1590 merges first there is nothing left to do; if this merges first, that single row shows `p.?` in the interim.

## The flipped test, and why its old rationale was the tell

`coding_length_changing_block_merges_across_one_unchanged_base` is renamed and its expectation flipped to `c.[2_3delinsAA;9del]`. Its doc justified merging a codon-**straddling** span like this:

> A wider merge is not "two variants" in that sense, so it is governed by the distance rule alone — the reading the SVD-WG states it is moving to outright (`general.md:39`).

`general.md:36-39` is dead text. It is a future-tense description of proposal **SVD-WG010**, which `docs/consultation/SVD-WG010.md:5` records as **rejected**, and the decided ruling `delins-codon-carve-out-gap-one` already says so in terms: *"Do not cite it, and do not cite SVD-WG010 as a live rule either. The codon-conditioned exception at `:18` / `general.md:35` remains the law."* That same ruling states the scope the pass had dropped: *"The carve-out is scoped to the coding sequence **and to one amino acid**."* So the flip is this test catching up with a correction already recorded against the clause it cites, not a new policy.

## What this does *not* touch

`tests/it/coding_frame_merge_axis_asymmetry.rs` — the assert-then-flip file that measures the `c.`/`n.`/`g.` disagreement — is **unaffected**, and all fourteen of its rows still pass unmodified. Its rows are `X:c.{p}delinsATC` built from two pure insertions flanking one base, so the merged span is a **single** reference position and is trivially inside one codon. The codon predicate does not settle that file's question. Its docs propose flipping it by "restoring `general.md:35`'s amino-acid precondition … so it declines on frameshifting blocks"; that is a *frameshift* test, which is the proxy again, and it is not what this PR implements. `codon-carve-out-shape-restriction`'s `decided` record does not reach that file either: it settles that edit type may not gate the exception, which is the opposite of adding a frameshift test.

## Verification

- `cargo nextest run --features dev --lib --test it` — **9,871 tests run, 9,871 passed**, 150 skipped, rc=0. (Was 9,784 before the census re-bless; the delta is this commit's new tests.)
- The three re-blessed suites on their own — `-E 'test(defect_non_idempotent_outputs) + test(spec_conformance_axis) + test(spec_corpus_regressions)'` — **30 run, 30 passed**, rc=0, with no `FERRO_ASSERT_*` set (the plain `test` job runs un-armed).
- Also run against the real prepared reference (`FERRO_MANIFEST`, `-E 'binary(it)'`), which is nightly-only in CI: **5,563 run, 5,562 passed, 1 failed** — `multi_member_cis_axis::axis_normalized`, measured `respelling_converged: 395` against pin `376`. That is **pre-existing and unmoved by this branch**: `origin/main` at `112ef784` measures the byte-identical census (`AxisCensus { rows: 592, declined: 0, not_idempotent: 0, unwindowed: 96, sequence_changed: 0, respellable: 461, respelling_converged: 395 }`) and fails the same assertion. Worth noting *why* a change that moves 1,581 synthetic rows moves this 592-row census by zero: both spellings of each class go through the same partitioner, so a spelling-independent partitioner change moves a class's two outputs together and the convergence count cannot see it.
- `cargo fmt --all -- --check` rc=0; `cargo clippy --all-features --all-targets -- -D warnings` rc=0 (one pre-existing third-party `proc-macro-error2` future-incompat note, which is not a failure).
- Generated artifacts regenerated: `generate_spec_fixture`, `generate_spec_enumeration` (both gitignored), `generate_conformance_summary` (`failure-patterns.md`, committed).

New tests, all derived from the clause rather than from today's output:

- `merge::tests::coalesce_coding_frame_separation_merges_only_within_one_codon` — fixture `ACG -> CT` (delete `A`, unchanged `C`, substitute `G>T`). At `w_lo = 1` the span is `c.1_3`, codon 1, and it merges; at `w_lo = 2` it is `c.2_4`, codons 1 and 2, and it declines. Same pieces, same payloads, same flags, one integer apart.
- `merge::tests::coalesce_coding_frame_separation_declines_a_span_wider_than_a_codon` — the old `TGCA -> AAC` fixture, four-position span, asserted declining at every phase 1..=6 so the result is the width and not a phase coincidence.
- `merge_consecutive_edits_tests::one_base_gap_within_one_codon_merges` / `…_across_a_codon_boundary_stays_split` / `without_a_reading_frame_neither_position_merges` — end to end on a new 30 nt codon-designed transcript carrying the same `ACG` triplet at two codon phases (`c.10_12` is codon 4 exactly; `c.14_16` straddles codons 5 and 6), with `GCT` flanks so neither `A` sits in a run and the two cases cannot differ by shuffling.

## Rebase note: this PR now also corrects the ruling ledger on `main`

Rebasing onto `822eb5d5` (#1623) conflicted in `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json`, because #1623 decided `codon-carve-out-shape-restriction` while this branch still described it as undecided. Resolved by taking `main`'s record wholesale — its `decided` status and its full argument — and re-applying only this branch's two `delins-codon-carve-out-gap-one` paragraphs. This branch's own "STILL NO RULING" paragraph is dropped: the ruling now exists.

**One thing in `main`'s record was wrong and is fixed here.** Its `IMPLEMENTATION` paragraph read *"1,581 of 78,298 corpus rows, all split -> merge"*. The direction is backwards, and the reason is worth recording rather than silently patching:

* **This ruling's** direction is a widening — lifting `apply_coding_codon_exception`'s `[Sub@p; Identity@p+1; Sub@p+2]` restriction would admit *more* merges, i.e. `split -> merge`. That half is **not implemented**; that pass still matches only the triplet.
* **This PR** moves the *other* pass the *other* way: `coalesce_coding_frame_separation` swaps a `length_changing` proxy for the clause, which **narrows** it, so rows move `merge -> split`.

Reasoning from the ruling's direction to its implementation's is what produced the wrong sign. Three independent things agree with `merge -> split`: the predicate gains a conjunct (`joins(pair) && one_amino_acid(pair)`) and so fires strictly less often; this PR's trailer and its `delins-codon-carve-out-gap-one` measurement both say so; and `NM_000143.3:c.44_47delinsATC` stops merging and is described individually. The ledger was the only place carrying the wrong sign.

The resolved record also now states what this PR does **not** implement — the frameshift constraint, since the predicate is codon membership alone, so `NM_004006.2:c.4375_4376delinsAGTT` still merges inside codon 1459.

Re-validated after the rebase: `generate_spec_fixture` rc=0 (it verifies every citation's verbatim quote against the spec checkout), 105 tests pass across the ruling/spec/census/merge modules, `cargo fmt --check` rc=0, `cargo clippy --all-features --all-targets -- -D warnings` rc=0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HGVS coding-sequence normalization so adjacent length-changing edits merge only when their combined span remains within a single codon.
  * Prevented edits crossing codon boundaries from being incorrectly represented as a single delins.
  * Preserved separate handling for equal-length substitutions and non-coding sequences.
* **Tests**
  * Expanded coverage for codon-boundary, reading-frame, and one-amino-acid delins scenarios.
  * Updated conformance results and documentation to reflect corrected HGVS interpretations.

Representation-Change: 1581 of 78298 corpus rows move, all merge -> split; 124 previously accepted (a real migration), 1457 previously not a fixed point. On cis_confluence_axis both census pins are re-blessed upward (3' 8026->8027, 5' 8021->8023). On the harsher spec_conformance_axis corpus the move is mixed and is re-blessed in both directions here: 3' non_idempotent_outputs 7->4 (three rows fixed), 3' converged 9140->9139 (six classes lose convergence against five gaining it, all six the spanning-delins respelling, promoted to spec_corpus_regressions::the_codon_gate_splits_a_spanning_delins_its_own_members_do_not), 3' split_two 2435->2436, 5' split_two 2696->2698 and split_three 204->202. Three mutalyzer conformance rows leave agreement and carry a spec_citation.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #1604, opened 2026-08-10
docs(rulings): make the precedence record a pointer to the README ruleset

nh13:
Supersedes #1569, and is much narrower than it. **Stacked on #1585** (`feat/exhaustive-spec-corpus-on-main`), not on `main`, because both PRs rewrite `hgvs_spec_normalization_overrides.json` — #1585 takes it from 10 records to 15 — so writing this against `main` would mean writing the ledger edit twice.

**Rebased 2026-08-11** onto #1585's current head after that branch was itself rebased onto `main` (#1580). The rebase applied without textual conflict, but two things still needed repair — the generated spec fixture was stale, and `CLAUDE.md`'s decided table was short one record. Both are covered below.

The canonical ruleset now lives in `README.md` and ships as its own PR. **This PR does not restate it.** That is the whole point of the change.

## `adjudication-precedence-order` becomes a pointer

The record used to state a five-rank order of its own; #1569 restated it differently. Both are superseded. The record now cites the README ruleset as canonical, records what changed and why, and keeps the two things that are genuinely its own — the evidence base, and the escalation register.

**Why a pointer rather than a better restatement.** This project's recurring failure is one rule written in several places and drifting apart, and it has already happened here: `CLAUDE.md`, this ledger and PR bodies have disagreed about the order. The concrete damage is on the record — a 2026-08-08 amendment to this very record removed a rank on the stated ground that `partition-is-the-unit-of-normalization` is decided, and

```
git log --all -S'partition-is-the-unit-of-normalization' -- tests/fixtures/grammar/hgvs_spec_normalization_overrides.json
```

finds that string only on unmerged partition-model branches and on the amendment's own commit. `main`'s copy of the file has zero occurrences, and it is not among #1585's 15 records either. **A rank was removed to end a contradiction with a record that had never merged.** A fourth statement of the rules is the bug, not the fix — so the superseded orders are not preserved next to the canonical one.

**Kept, verbatim:** the four evidence paragraphs gathered for those rules — the Mutalyzer forensics, the finding that the filer's at-risk issues are a *method* request rather than a direction (3 merge / 3 split / 5 respell), the spec grounds for confluence over stability, and stability-as-last-resort-tiebreaker. They were expensive to gather and are cited from elsewhere in the repo. Two internal cross-references that pointed at the removed ranks are re-pointed: at `canonical-form-choice-when-both-legal` for the method, and at the ruleset's disclosure rule for version declaration.

## The escalation register is kept, and re-framed

It is **not** the machinery for ordinary spec silence — that is the ruleset's rule 5 (file upstream first, cite it, ship a provisional choice), a normal and not-rare zone. The register is the rarer act of overriding a preference the spec **does** state, because nothing decides between the candidates satisfying it. The `precedence_escalations_are_censused` failure message now says so explicitly, so the next reader does not file rule-5 work here.

**E1 stands** on that ground: `general.md:56` ranks substitution, deletion, inversion, duplication and insertion, and `delins` is not among them. Its tail is corrected — `canonical-form-choice-when-both-legal` is `decided` now, so E1 is not "the open product decision"; it is the residue that decision leaves, where derivation itself leaves two all-`delins` candidates and the spec ranks neither. A *typing* question over candidates the derivation produced, never a licence to keep the input's partition.

## RFC 2119 count corrected: one → **two**

`CLAUDE.md` said uppercase RFC 2119 keywords appear *"exactly once"* in `docs/recommendations/` outside `style.md`. There are **two**, both in `RNA/adjoined_transcript.md`:

- `:20` — "This syntax **REQUIRES** the use of a range (not a single position)"
- `:21` — a **SHOULD**, assay-scoped: conditioned on the junction rather than the whole transcript being analyzed

The alternation everyone copies from RFC 2119 lists `REQUIRED`, so it misses `REQUIRES` — which is why `:20` was invisible to the grep the count was taken from. The grep in `CLAUDE.md` is fixed, and neither hit generalises, so the section's conclusion is unchanged.

## Ledger summary rebuilt

`CLAUDE.md`'s two tables had drifted from the file they summarise. Against #1585's 15 records (10 `decided`, 5 `undecided`, verified by parsing the file):

- `adjudication-precedence-order` and `canonical-form-choice-when-both-legal` were tabled as **undecided**; both are `decided`.
- `separation-is-a-property-of-the-spelling-not-of-the-variant` was **missing** from the undecided table.
- `delins-adjacent-members-when-both-consume-reference` was **missing** from the decided table.
- `self-cancelling-across-ring-junctions` was **missing** from the decided table. This one is a rebase artifact and worth naming as such: `main` gained the record from #1589 *after* this branch was written, so the tables drifted again by exactly one row — the failure mode this PR exists to fix, reproduced live. It is added here.

#1585's count-free phrasing is kept — restating a denominator is what made this stale before — with a matching warning added about reading a record's `status` from the file rather than from the table.

## Note on what could not be done as asked

An `upstream_status` field on ruling records is **not available**: `Ruling` is `#[serde(deny_unknown_fields)]` in `examples/generate_spec_fixture.rs`, so a new field fails the generator. Upstream status is carried in the rationale prose instead, following the form `rna-repeat-range-plus-unit-redundancy` already uses ("owed as a #466 entry against the HGVS nomenclature repository"). Adding the field is a separate change.

## Verification

The generated clause→record index in `tests/it/clause_ruling_index.rs` is regenerated: `general.md:56` is now cited by the record (E1 names it, so it has to be), which moves two rows and the multi-record count. `generate_spec_fixture` validates every citation quote byte-for-byte against the pinned spec checkout, so a paraphrase or a moved clause fails the build — it passes. All seven `hgvs_spec_normalization_tests` guards pass, including `ruling_records_are_intact` and `precedence_escalations_are_censused`. The full local suite is green — **9,980 tests, 0 failed**, nothing re-blessed beyond the generated index above. `cargo fmt --all --check` and `cargo clippy --features dev --all-targets -- -D warnings` are clean.

Representation-Change: none. Ruling text, documentation, and one census constant with its test; the diff touches no file under `src/`, so no code path is altered and no normalized output moves.


nh13:
@coderabbitai pause

nh13:
## Merge order is load-bearing here, and nothing enforces it

This record now says, as its opening claim, that "the project's canonical ruleset lives in `README.md` and is the only place it is stated." On `main` today that is **false** — `README.md` has no such section (its only normalization heading is `## Comparing normalization rules (FERRO_PARTITION)`). The `## Normalization rules` section this record points at is added by #1605, which is a separate PR based on `main`.

So the required order is **#1585 → #1605 → #1604**. Merging this one before #1605 ships a dangling pointer in the ledger — the exact class of defect the record's own argument is about.

**And no test catches it.** #1605 carries `tests/it/readme_normalization_rules.rs`, but that guards the README against itself; it has no knowledge of the ledger record. Nothing on either branch asserts that the record's pointer resolves to a heading that exists. A guard belongs here — something like "the ledger record naming `README.md` as canonical must find the section it names" — but it cannot be added on this branch, because it would be red until #1605 lands. Filing it as follow-up work rather than pretending the branch can carry it.

The narrower fix available right now, if the order slips: this PR's base could move to #1605 instead of #1585 — except the record edit is written against #1585's ledger, so that trades one dangling reference for a duplicated ledger edit. The order is the cheaper mechanism; it just needs to be honoured deliberately rather than assumed.

## Rebase note

Rebased onto #1585's current head after that branch was itself rebased onto `main` (#1580). No textual conflict, but two repairs were still needed and neither was visible in the merge:

- the generated spec fixture was stale, so `ruling_records_are_intact` failed against the ledger it no longer matched;
- `CLAUDE.md`'s decided table was short `self-cancelling-across-ring-junctions`, which `main` gained from #1589 after this branch was written.

The second is worth naming: this PR exists because those tables drift from the ledger, and the rebase reproduced the drift live, by one row. Both fixed; suite is 9,980 passed / 0 failed.

PR #1606, opened 2026-08-10
feat(normalize): split an equal-length protein delins whose interior is unchanged

nh13:
Two pieces salvaged from #1565's stack and rebased onto `main`. #1565 makes the input's asserted partition the unit of normalization; that model was reversed on 2026-08-09 in favour of re-derivation, so #1565 is being closed and everything stacked on it goes with it. Neither piece here depends on that model, and that was **measured before anything was written** rather than assumed — the numbers are under "How independence was checked".

Supersedes #1568, which is the same protein fix stranded behind that model in the same stack, and salvages the W1–W95 suite from #1565 itself, which carries it and exists nowhere else.

## What the spec publishes, and what ferro emitted

`protein/delins.md:21` states the rule outright — "two variants separated by one or more amino acids should be described individually and not as a `delins`" — and `:62-64` prints the worked answer, `p.[Ser44Arg;Trp46Arg]`, with the explicit note that "the variant is **not** described as `p.Ser44_Trp46delinsArgLeuArg`". Ferro emitted the merged form.

It is not a one-off. `SVD-WG010.md:53-54` carries the same shape at a second locus (`p.Ser988_Gln990delinsAlaLeuHis`), and there the merge was the *rejected* proposal. Two published rows, one behaviour.

The **merge** direction already worked on this axis — `p.[Trp24Cys;Val25Arg]` merges to `p.Trp24_Val25delinsCysArg` (W5) — so this is a missing half, not a wrong rule.

## Scope: equal length, and only equal length

An equal-length `delins` has exactly one residue-wise correspondence, so "the interior residue is unchanged" is a fact about the two sequences. An unequal-length one has none: locating an unchanged run inside it means first *choosing* an alignment, and the reference and the payload do not determine which.

The nucleotide axis settles that by applying the edits to the reference and re-deriving from the resulting sequence, which is what `canonical-form-choice-when-both-legal` directs. That is unavailable here — there is no apply-to-reference on the protein axis, which is why `merge::cis_kind_of` returns `None` for `Protein`. With nothing to derive the answer from, ferro declines rather than invent an alignment.

## The declines are the substance, and each carries an argument

A positive-only suite goes green on an implementation that splits everything in reach — the unequal-length case, the stop-codon case and the no-reference case all *look* like the W48 shape and all have a different right answer. So every decline names the exact string ferro must not emit:

| decline | why |
|---|---|
| unequal length | no residue-wise correspondence; see above |
| no protein reference | the unchanged residue is a claim about the **reference**, and the payload cannot testify to it. Assuming `payload[i]` is what the reference holds is the one guess this rule must never make |
| `Ter` in the payload | a split would emit members 3' of the stop, which `delins.md:45` forbids |
| `Ter` in the reference span | a stop replaced by an ordinary residue is a **stop-loss**, which `extension.md:18` ranks first and `:30` spells `p.Ter110GlnextTer17` — never a substitution |
| span under three | no interior to be unchanged |
| fewer than two changed runs | the change really is one contiguous `delins` |
| **residue 1 changed** | see below |

## The residue-1 guard is a defect the fix would otherwise introduce

`p.Met1_Ala3delinsValAlaTrp` has the W48 shape exactly, so every other gate passes it through — and the members it produces are `p.[Met1Val;Ala3Trp]`. `protein/substitution.md:49` says a variant changing the translation initiation codon is "not described as a substitution or as an extension", and `parse_hgvs` enforces that, so the split would emit a description ferro cannot read back.

The illegality is **created by the split**. The range input parses, because `validate_no_start_loss_substitution` keys on a single certain position — so the guard has to live in the split rather than being inherited from the parser. Measured on `main` at `7118eb2a`, through the public API with a synthetic provider:

| input | `main` | `main` + the split, ungarded | with the guard |
|---|---|---|---|
| `p.Met1_Ala3delinsValAlaTrp` | unchanged, re-parses | `p.[Met1Val;Ala3Trp]` — **refused by `parse_hgvs`** | unchanged, re-parses |

Ferro declines rather than choosing among `p.0`, `p.0?`, `p.(Met1?)` or the upstream-initiation insertion form: which of those is right is a claim about the *consequence*, and two protein sequences do not settle it.

The guard is scoped to residue 1 **changing**, not to any span reaching it — `a_span_reaching_residue_one_still_splits_when_residue_one_is_unchanged` pins that, because a guard written as "decline any span starting at 1" would look correct and silently stop splitting a family it has no business declining.

## The W1–W95 suite

The recommendations print 95 input → published-answer pairs for themselves. Those are the only rows in the corpus that need no interpretation: the spec names an input and prints, in the same sentence, the description to use instead, so a divergence is a defect or a spec self-conflict, never a matter of taste.

Every executable row runs hermetically — no `FERRO_MANIFEST`, no network, no environment variable — so nothing can pass by being skipped, and every row that *cannot* be executed is recorded with its reason, so the size of what is not covered is itself pinned. Assertions are exact equality on the whole output string; rows about a partition additionally pin member count and every member's span read off the **parsed** output, because a rendered string can look right over a wrong partition. The negative half is asserted too — the exact string ferro must not emit for a legal input, and the spellings the parser must refuse.

It is a companion to `spec_worked_examples.rs`, which runs a hand-curated twelve rows through all three error modes; those are cross-referenced, not duplicated, and a test keeps them out of this file's tables.

## One adjudication, recorded

`SVD-WG010.md:16` prints one variant twice — `g.3_4delinsGGT` and `g.[3C>G;7dup]`. The proposal was **rejected** (`:5`), so `general.md:34` governs and the split is the conformant form. Ferro re-derives the split from **both** spellings, so the pair is confluent *and* lands on the form the rejection leaves in force. That is `canonical-form-choice-when-both-legal` doing its work — derive from the resulting sequence, subject to every explicit spec tie-break — rather than a coincidence.

Contrasted in the same file with `DNA/delins.md:44-47`'s pair, which does **not** converge, because `:47` recommends the *spanning* form while `general.md:34` requires the *split* one. Both are pinned, so a change that converges the second pair, or un-converges the first, fails and has to say which clause it is now reading.

## How independence was checked

Before writing anything, `spec_worked_example_rules.rs` was copied onto `main` unmodified and run:

| tree | result |
|---|---|
| `main`, file unmodified | **16 / 18 pass** |
| `main` + this PR's `src/normalize/mod.rs` | **17 / 18** — the protein rows go green |
| this PR | **36 / 36** — the last one adjudicated above |

So the file is not model-dependent; it was stranded by its position in a stack, and its two failures on `main` are real `main` behaviours rather than artifacts. `src/normalize/mod.rs` is byte-identical between `main` and #1568's base, so the fix applied verbatim.

## Blast radius

**Three zeros, each with its reason, because a zero from a corpus is a claim about the corpus:**

- `examples/dump_normalized_corpus.rs` generates **no protein rows at all** — a structural zero, reported rather than quoted.
- ClinVar (`clinvar_hgvs_unique.json.gz`, 4,188,436 rows) holds **8,445** protein descriptions, so it is *not* blind to the axis, and exactly **2** protein `delins`: `NP_009225.1:p.Phe1036_Lys1037delinsPheTerArgfs` and `NP_000539.2:p.Cys189_Tyr190delinsCysPheThrfs`. Both are span-2 and both carry `Ter`, so both are declined on two independent gates. Candidate population **0**.
- The prepared reference carries `protein_fastas: []`, so `has_protein_data()` is false and the gate cannot fire against real reference data at all today.

The move is real and it is a **split, never a merge**; it is simply not reachable through any corpus currently available to measure it. The strong evidence for an output-moving change is therefore the suite: **9,869 tests pass with nothing re-blessed** — every existing expectation already agreed with the new answer.

## Not fixed here, and filed separately

A *declined* split still emits an illegal `p.Met1Val` on `main`. `p.Met1_Ala3delinsValAlaAla` has one changed run, so this rule never fires, and `main` produces `p.Met1Val` — which `parse_hgvs` refuses — from a different path. It is pre-existing, unchanged by this PR in either direction, and belongs to whatever owns that path.

## Verification

`cargo fmt --all --check` clean. `cargo clippy --features dev --all-targets -- -D warnings` clean. Full suite **9,869 passed, 0 failed, 150 skipped**, nothing re-blessed.

## Summary by CodeRabbit

- **New Features**
  - Protein descriptions with eligible equal-length `delins` changes can now be split into separate top-level variants.
  - Split results may appear as individual substitutions or smaller `delins` changes.
  - Normalization supports sequence-based canonicalization before applying these splits.

- **Bug Fixes**
  - Added safeguards to preserve descriptions involving uncertain positions, unavailable references, contiguous changes, initiation-codon changes, stop or unknown residues, and other unsupported cases.

- **Tests**
  - Added comprehensive coverage for protein splitting, predicted descriptions, repeated normalization, and unsupported scenarios.

Representation-Change: Protein axis only — an equal-length `p.` delins whose interior residue is unchanged now splits into its members, a split and never a merge. Measured blast radius is zero on all three corpora available, each for a stated reason (see Blast radius); the gate cannot fire against the prepared reference at all, since it carries no protein FASTAs.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -2299,7 +2299,56 @@ impl<P: ReferenceProvider> Normalizer<P> {
         variant: &HgvsVariant,
     ) -> Result<(HgvsVariant, Vec<NormalizationWarning>), FerroError> {
         let (normalized, warnings) = self.normalize_core(variant)?;
-        self.canonicalize_from_sequence(normalized, warnings)
+        let (normalized, warnings) = self.canonicalize_from_sequence(normalized, warnings)?;
+        Ok((self.split_protein_separation(normalized), warnings))
+    }
+
+    /// The protein-axis partition pass, run once at the top level.
+    ///
+    /// [`Self::try_protein_split_delins`] holds the rule and its citations; this
+    /// is only about **where** it runs. It runs here, beside
+    /// [`Self::canonicalize_from_sequence`] and for the same reason: the move
+    /// turns one member into several, so it belongs at the seam that sees the
+    /// whole description rather than in
+    /// [`normalize_protein`](Self::normalize_protein), which `normalize_allele`
+    /// re-enters once per member.
+    ///
+    /// Wiring it into `normalize_protein` was tried, and produced two defects
+    /// that every bare-description test passes straight over:
+    ///
+    /// ```text
+    ///   p.[Ser44_Trp46delinsArgLeuArg;Ala60Gly]
+    ///     -> [NP_003997.1:p.Ala60Gly;NP_003997.1:p.[Ser44Arg;Trp46Arg]]
+    ///   p.[Ser44_Trp46delinsArgLeuArg];[Ala60Gly]
+    ///     -> NP_003997.1:p.[Ser44Arg;Trp46Arg];[Ala60Gly]     (does not re-parse)
+    /// ```
+    ///
+    /// The first is a nested bracket carrying a repeated accession, which is not
+    /// a description at all. The second *is* legal HGVS — the DNA axis emits
+    /// `g.[A;B];[C]` and reads it back — but ferro's **protein** allele grammar
+    /// accepts only single-member arms, so normalization would emit a string
+    /// `parse_hgvs` refuses. That is exactly what `FERRO_ASSERT_REPARSE` exists to
+    /// catch, and its exemption list is closed on purpose.
+    ///
+    /// So the move applies to a whole `p.` description and **declines for a
+    /// member of one**: `p.[<separated delins>;…]` keeps its delins. Closing that
+    /// needs two decisions this pass deliberately does not take — where an inner
+    /// allele's `( )` marker goes once its members flatten into the outer bracket
+    /// (`protein/alleles.md:34` puts it per member, `:90` puts it round the group)
+    /// and a protein allele grammar that reads `p.[A;B];[C]`. Pinned by
+    /// `protein_axis_split_move::a_delins_inside_a_cis_allele_is_left_alone` and
+    /// its trans sibling.
+    fn split_protein_separation(&self, variant: HgvsVariant) -> HgvsVariant {
+        let HV::Protein(protein) = &variant else {
+            return variant;
+        };
+        match self.try_protein_split_delins(protein) {
+            Some(members) => {
+                let uncertain = protein.loc_edit.edit.is_uncertain();
+                wrap_allele_if_split(members.into_iter().map(HV::Protein).collect(), uncertain)
+            }
+            None => variant,
+        }
     }
 
     /// Normalize a variant, apply the strict-mode rejection ladder, and return
@@ -5750,6 +5799,13 @@ impl<P: ReferenceProvider> Normalizer<P> {
             None => final_variant,
         };
 
+        // NOTE: the protein-axis split move (`protein/delins.md:21`, `:64`) is
+        // deliberately NOT applied here. It turns one member into several, so it
+        // runs at the whole-description seam —
+        // [`Self::split_protein_separation`], called from
+        // `normalize_core_canonical` — which this method cannot be, because
+        // `normalize_allele` re-enters it once per member. That helper records
+        // what wiring it in here was measured to produce.
         Ok((HV::Protein(final_variant), warnings))
     }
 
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
## This PR now has a real measurement, not a structural zero

#1614 adds a protein axis to `examples/dump_normalized_corpus.rs`. It was built partly because of this PR — the harness could not emit a single `p.` row, so the honest thing here was to *report* the zero and name the generator as its cause, which is what the Blast radius section does. That is the right handling of a blind corpus, but it is still a measurement nobody had.

Dumping both sides through the new harness (the module's own cross-revision procedure — copy the file into the base checkout so both dumps describe the same corpus):

| | rows |
|---|---|
| total | 85,642 |
| **moved** | **1,584 (1.8%)** |
| of which previously **accepted** (a migration) | 1,584 |
| of which previously **not** a fixed point (free) | 0 |

Every moved row is in `protein_equal_length_delins`, and **no row moves in any of the other 20 families** — including the four other protein families and all 16 nucleotide ones. So the change is exactly as scoped as it claims: a split, never a merge, confined to the equal-length-interior-unchanged shape.

Two things worth deciding, both yours rather than mine:

- **The Blast radius section and the `Representation-Change:` trailer are now understated.** They say the measured radius is zero on all three available corpora. That was true when written; the synthetic corpus now answers with 1,584 rows. The ClinVar finding (candidate population 0) and the prepared-reference finding (`protein_fastas: []`) are unaffected and still the ones that bound *real-world* impact — this is a synthetic shape corpus, so quote its rate as "of the affected shape family", never repo-wide.
- **Merge order.** If this lands first, the trailer stays as-is and someone re-derives the number later. If #1614 lands first, this PR can quote it directly. I'd suggest the latter, but #1614 is a test-only change and has no dependency on this one, so either works.

No action needed from this PR to keep CI green — #1614 touches only `examples/` and the two do not conflict.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1609, opened 2026-08-10
feat(equivalence): expose a groupable SPDI key for bucketing by denoted bases

nh13:
Salvaged from #1565's stack and rebased onto `main`. #1565 makes the input's asserted partition the unit of normalization; that model was reversed on 2026-08-09 in favour of re-derivation, so #1565 is being closed and everything stacked on it goes with it. This piece never touched the partitioner — it was stranded only by its position.

Supersedes #1567.

**No longer a pure addition.** The body previously said nothing under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` was touched and no existing behaviour changed. Review found a defect in the key's core contract that is only fixable in `src/spdi/`, so both halves of that sentence are now false and the correction is disclosed in the trailer. See [A correctness defect the first CLI review found](#a-correctness-defect-the-first-cli-review-found) below.

## The gap

`EquivalenceChecker` answers a rich question about **two** variants and returns a *level*, not a value — `AccessionVersionDifference` is decided by comparing two accessions' versions, so there is no scalar a bucket could be keyed on. A consumer that wants to count distinct changes across a whole call set needs the other shape: one value per variant, equal exactly when the denoted bases are.

Comparing normalized strings is not that. It answers a different question, and it couples the consumer's stored buckets to the normalizer.

## Why that coupling is a real cost, not a hypothetical

Two spellings that partition the same bases differently can reach different canonical forms, and that is by design rather than a gap to close. `canonical-form-choice-when-both-legal` is **decided**: derive from the resulting sequence, "subject to every explicit spec tie-break". Where two tie-breaks point opposite ways, neither is selected and both forms survive.

The spec supplies its own instance. `DNA/delins.md:44` gives a spanning `delins`, `:46` names a four-member split as an "alternative description" of the *same* variant, `:47` recommends the spanning form, and `general.md:34` requires members separated by unchanged nucleotides to stay individual. Two clauses, opposite directions, one variant, two legal forms.

And confluence work *moves* canonical forms — this project declares those moves as breaking rather than avoiding them. So buckets keyed on the normalized string re-bucket on every such release. `SpdiKey` is derived from `canonical_spdi` without consulting the normalizer, so it does not move when a canonical form does.

## The `None` contract is the load-bearing half

`spdi_key` returns `None` rather than a key that would silently collide. A key that answered for everything would bucket unrelated variants together, which is strictly worse than refusing — a consumer can see a refusal and cannot see a wrong bucket. The classes that have no key are enumerated on `spdi_key` itself.

`SpdiKeyGrouping` reports `unkeyable()` alongside the groups for the same reason: a grouping that quietly dropped what it could not key would report a smaller, cleaner-looking population than it was given.

## A correctness defect the first CLI review found

This PR had **eight** CodeRabbit CLI attempts; the first seven were rate-limited, so the eighth was the first time the CLI lens ever ran on it. It found a real defect in exactly the property the type exists for.

`canonical_spdi` read the key's deleted bases — and any inserted bases a `dup` or `inv` copies *out of* the reference — straight from the fetched window, while the inserted payload arrived through the triples, which `hgvs_to_spdi` has already folded with `apply_alphabet`. So one key could hold two alphabet conventions at once, and **a soft-masked reference keyed differently from the uppercase copy of the same sequence.** Measured on the module's own fixture contig, `g.3_7delinsGGCTA`:

```
uppercase contig   NC_KEY.1:2:ATTAC:GGCTA
lowercase contig   NC_KEY.1:2:attac:GGCTA
```

Real genomic FASTAs soft-mask repeat tracts in lowercase, so this is the ordinary case rather than a corner, and it breaks the one contract the whole type carries: equal keys mean equal bases. It was invisible because `trim_common_flanks` is already case-insensitive — the position and the block boundaries were right all along, only the emitted strings carried the provider's case — and because every fixture in the tree is uppercase.

Fixed by folding both emitted strings through the existing `apply_alphabet`, per its own doc at `convert.rs:1872` explaining why that is preferred to a bare `to_ascii_uppercase`: on the `r.` axis it also rewrites `U` to `T`, which is the convention the triples already carry. That reconciles a uracil-spelled transcript **only where the U-bearing reference bases reach the emitted block through `trim_common_flanks`** — the fold runs after the two comparisons that decide the block (`trim_common_flanks` and `apply_triples`' deletion validation) and both are `eq_ignore_ascii_case`, which does not relate `U` to `T`. Measured on one transcript served both ways: `r.[3a>g;7c>a]`, `r.3a>g` and `r.14dup` agree; `r.13_14insu` keys at `14::T` against `13::T` because the 3' roll's prefix scan stops at the first mismatch; and `r.3_5del`, `r.3_7delinsggcua` and `r.3_7inv` are **declined** on the uracil provider because a stated deletion of `T` fails validation against `U`. Every row is pinned by `the_alphabet_fold_reconciles_case_and_bounds_the_uracil_case`. None of it is reachable from a RefSeq-spelled provider, which stores transcripts as DNA, so the boundary is recorded rather than closed — making those two comparisons alphabet-aware would change the `n.`/`c.` paths too. `AppliedVariant` is deliberately **not** folded — handing back the window as served is its purpose, and a caller inspecting soft-masking there is asking a legitimate question.

`a_soft_masked_reference_keys_the_same_as_an_uppercase_one` pins it across six shapes (stated and unstated deletions, `dup`, `inv`, substitution, insertion) and asserts on `Eq`, `Hash` **and** `Ord` together, since a `BTreeMap` bucket is found by `Ord` and a `HashMap` bucket by `Hash` — a key that compared equal while hashing differently would still split a consumer's buckets.

## The measurement harness, and its vacuity denominator

`examples/measure_spdi_key_grouping.rs` asks the claim of the population that actually exhibits the divergence — every class `dump_confluence_divergences` reports — rather than of hand-written pairs. A hand-written pair proves the mechanism and says nothing about coverage.

Four verdicts, deliberately four rather than two: `grouped`, `split`, `undecided` (at least one output has no key, so the class cannot be answered either way) and `unmeasured` (at least one output could not be *asked*). `undecided` is `spdi_key` declining, which is evidence about the key; `unmeasured` is the harness failing, which is evidence about the corpus. Folding either into `split` would report a failure of one as a finding about the other. A run with any `unmeasured` exits non-zero.

**The vacuity check prints unconditionally.** A run whose input holds no class with two or more distinct outputs measures nothing about grouping, and a `0 split` from such a run reads exactly like success — the corpus-zero trap this repository has now hit three times (#1456, #1460, #1478). `multi_output` is that denominator, and when it is zero the report says so instead of printing a ratio.

Two further review findings landed on this harness, which has now had a counting defect corrected in three separate reviews — first an omitted `Verdict::Unmeasured` census row, now these:

- **The vacuity denominator counted JSON rows, not distinct output texts.** `dump_confluence_divergences` groups spellings into a map keyed on the output text and emits one row per key, so its own dumps cannot produce a duplicate and **no figure this harness has printed changes.** What the dedupe buys is that the single input shape able to defeat the vacuity check was the one it could not see: a class holding one output twice reads as `>1 distinct output`, suppressing the `VACUOUS` banner while measuring nothing. Only the count is deduped — two rows sharing a text carry *different* `from` spellings and the spelling pass needs both.
- **Spelling failures were conflated with output failures.** Both were pushed into one list, printed under a heading asserting "their classes are counted under the `unmeasured` verdict". That is true of an output row and **false of a spelling row**: the verdict is a statement about a class's normalized outputs and is fixed before the spelling pass runs, so a class can be `grouped` on its outputs while a spelling of it never reached the key. The reader was being sent to look for a census row that is not there. Now tracked independently and rendered under its own `UNMEASURED SPELLINGS` heading that says explicitly it changes no verdict; output-failure behaviour is unchanged, and both halves still fail the run, with the exit message naming which figure each shrank.

The target carries `test = true` so its `#[cfg(test)]` units actually run. It writes no artifact, so it takes on no `CaptureLedger` obligation.

## Two smaller review fixes

- **`SpdiKey`'s `Ord` was hand-written over all four fields.** Verified correct as written, but it silently stops being total the moment a fifth field is added, while still type-checking and still agreeing with `Eq` on every pre-existing value — two keys differing only in the new field would compare `Equal` and collapse into one `BTreeMap` bucket. `SpdiVariant` now derives `Ord`/`PartialOrd` and the key delegates to it. Its declaration order is `(sequence, position, deletion, insertion)`, which is exactly the order the key documents; derived alongside the derived `Eq`, so the two cannot disagree. `ord_follows_the_documented_field_order_and_agrees_with_eq` varies each field in isolation from least significant up and asserts consistency with `Eq` in both directions.
- **The `equivalence` module doc claimed `SpdiKey` "corresponds to the `SequenceMatch` rung".** Two independent review lenses raised this. Equal keys cover `Identical`, `NormalizedMatch` **and** `SequenceMatch` at once, so naming one understates it; rewritten to say so, with the implication's direction stated explicitly — a pair *at* one of those rungs need not key at all, since two identical `p.` descriptions are `Identical` and `spdi_key` refuses them. The section was also sitting after `# References`, separated by a blank non-doc line, which concatenates its heading onto the last list item; moved before the references.

## What was corrected in the salvage

Two claims in the inherited text were true on the stack and false on `main`, and both are load-bearing to the module's motivation rather than incidental:

- The module doc opened *"this project's normalizer preserves that partition rather than re-deriving a minimal one from the resulting sequence."* `main` re-derives. Restated from what actually holds — the derivation is subject to spec tie-breaks that sometimes conflict, which is why two forms survive. The motivation is stronger this way, and it no longer depends on which model ships.
- It quoted `canonical-form-choice-when-both-legal` as saying *"Sequence-level equivalence still needs an answer for consumers who dedupe; that is `EquivalenceLevel::SequenceMatch` and a groupable SPDI key, not the canonical string."* **That sentence is not in `main`'s copy of the record** — it was added by #1565's ledger change, which is being closed with it. Quoting it here would have left a citation pointing at text that does not exist. Removed, and the point argued from the record's actual decided text instead.

One smaller fix: the doctest's comment claimed the two descriptions "are two canonical forms by design" while the code compares two freshly *parsed* strings and never normalizes. Reworded to what it demonstrates, with a note that the key holds whether or not the normalizer converges the pair — which is the property that matters.

## A trap avoided, worth recording

Lifting the change as `git diff origin/main <stack-branch> -- Cargo.toml` **deletes** the `extract_inversion_sweep_windows` example target, which `main` gained from #1554 earlier the same day. A branch-to-branch diff reverts the newer side wherever the two have both moved. The `Cargo.toml` hunk here is the addition only; `extract_inversion_sweep_windows` is untouched.

## Verification

Every figure below is re-measured after the review fixes; the previous body's counts (127 targeted / 94 doctests) were taken before three tests were added and are superseded rather than carried forward.

- `cargo nextest run --features dev -E 'binary(measure_spdi_key_grouping) + test(equivalence::key) + test(equivalence)'` — **135 passed, 0 failed**.
- `cargo test --features dev --doc equivalence` — **6 passed, 0 failed**. Stated separately because `cargo nextest` does not run doctests, and this module's contract is carried by one.
- `cargo nextest run --features dev` (full suite, run because this now touches `src/spdi/`, which is widely depended on) — **10063 passed, 0 failed**.
- `cargo fmt --all -- --check` clean; `cargo clippy --all-features --all-targets -- -D warnings` clean apart from the pre-existing third-party `proc-macro-error2` future-incompat line.

`RUSTDOCFLAGS="-D warnings" cargo doc` reports 106 errors repo-wide on `main` already; none is in a line this PR touches (the single hit under a touched file is `hgvs_to_spdi`'s pre-existing link at `convert.rs:371`). Recorded so the number is not read as introduced here.

Representation-Change: SPDI keys move on a soft-masked reference. `canonical_spdi`'s emitted `deletion` and `insertion` are now folded through `apply_alphabet`, so a lowercase repeat-masked contig keys `2:ATTAC:GGCTA` where it previously keyed `2:attac:GGCTA`, and an `r.`-axis provider serving uracil now folds to the RefSeq DNA spelling wherever the fold reaches (see above for the three shapes it does not, none reachable from a RefSeq-spelled provider). No normalized HGVS string changes, and no key derived from an uppercase reference changes. The affected consumer is any that keys off a soft-masked FASTA, for whom the previous keys split buckets that denote identical bases — so this is a fix to the type's contract, migrated by re-deriving keys once.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -402,6 +402,14 @@ name = "dump_confluence_divergences"
 required-features = ["dev"]
 test = true
 
+# `test = true` because the target carries `#[cfg(test)]` unit tests pinning the
+# three verdicts and the vacuity guard — the parts that decide whether a printed
+# figure means anything. Cargo does not build a target's tests unless it opts in.
+[[example]]
+name = "measure_spdi_key_grouping"
+required-features = ["dev"]
+test = true
+
 [[bin]]
 name = "ferro"
 path = "src/bin/ferro.rs"
--- a/examples/measure_spdi_key_grouping.rs
+++ b/examples/measure_spdi_key_grouping.rs
@@ -1,0 +1,944 @@
+//! Measure whether [`spdi_key`] buckets the confluence divergences.
+//!
+//! # Why
+//!
+//! `src/equivalence/key.rs` claims that descriptions asserting different
+//! *partitions* of one edit — which can reach distinct canonical forms, since
+//! the derivation is subject to spec tie-breaks that sometimes point opposite
+//! ways — nonetheless share one grouping key. The unit tests pin that on
+//! hand-written pairs. This asks it of the population that
+//! actually exhibits the divergence: every class
+//! `examples/dump_confluence_divergences.rs` reports, where two spellings of one
+//! designed variant reached two different normalized strings.
+//!
+//! The distinction matters because a hand-written pair proves the mechanism and
+//! says nothing about coverage. A claim of the form "the key groups the
+//! divergent classes" is a claim about a corpus, and this is the only thing that
+//! can settle it.
+//!
+//! # What is measured, and how it can be vacuous
+//!
+//! Per class: the key of every **distinct normalized output**. Those outputs are
+//! what a downstream consumer stores, and the class is in the input precisely
+//! because there is more than one of them. Four verdicts:
+//!
+//! - `grouped` — every output keyed, and to the same key. The divergence is
+//!   invisible to a consumer bucketing by key.
+//! - `split` — every output keyed, to two or more keys. The key does **not**
+//!   reconcile this class.
+//! - `undecided` — at least one output has no key, so the class cannot be
+//!   answered either way. Reported separately rather than folded into `split`:
+//!   those are different findings and one must not be read as the other.
+//! - `unmeasured` — at least one output could not be *asked*: it does not parse,
+//!   or keying it panicked. Split out from `undecided` for the same reason
+//!   `undecided` is split out from `split`. `undecided` is `spdi_key` declining,
+//!   which is evidence about the key; `unmeasured` is this harness failing,
+//!   which is evidence about the corpus. A run with any of these exits non-zero,
+//!   because every count in the report is then over a smaller population than
+//!   the input and a partial run would otherwise read as a clean one.
+//!
+//! The four verdicts are about a class's **outputs**. The same three-way split is
+//! applied to the class's raw *spellings*, and its failures are tracked and
+//! reported separately, because they have a different consequence: an unaskable
+//! spelling shrinks the spelling figure's denominator and leaves the verdict
+//! census intact. Both are fatal to the run; only the first is a hole in a
+//! verdict.
+//!
+//! **The vacuity check is printed unconditionally.** A run whose input holds no
+//! class with two or more distinct outputs measures nothing about grouping, and
+//! a `0 split` from such a run reads exactly like success. `multi_output` is
+//! that denominator; when it is zero the report says so instead of reporting a
+//! ratio.
+//!
+//! # Usage
+//!
+//! ```text
+//! cargo run --release --features dev --example generate_cis_confluence_corpus
+//! cargo run --release --features dev --example dump_confluence_divergences -- --out /tmp/div.json
+//! cargo run --release --features dev --example measure_spdi_key_grouping -- --divergences /tmp/div.json
+//! ```
+
+use std::collections::{BTreeMap, BTreeSet};
+use std::fmt::Write as _;
+use std::path::PathBuf;
+use std::process::ExitCode;
+
+use clap::Parser;
+use serde::Deserialize;
+
+use ferro_hgvs::equivalence::spdi_key;
+use ferro_hgvs::parse_hgvs;
+use ferro_hgvs::reference::transcript::{Exon, GenomeBuild, ManeStatus, Strand, Transcript};
+use ferro_hgvs::reference::MockProvider;
+use ferro_hgvs::spdi::canonical_spdi;
+
+/// Kept in step with `examples/dump_confluence_divergences.rs` and
+/// `examples/generate_cis_confluence_corpus.rs`; all three must agree on the
+/// padding, the accessions and the CDS or the coordinates mean different things.
+const PAD_OFFSET: usize = 256;
+const GENOMIC_CONTIG: &str = "NC_TEST.1";
+const TX_ACCESSION: &str = "NM_TEST.1";
+const TX_CONTIG: &str = "chr_synth";
+const CDS_START: u64 = 1;
+const CDS_END: u64 = 63;
+
+#[derive(Parser, Debug)]
+#[command(about = "Measure how the SPDI grouping key buckets the divergent confluence classes")]
+struct Cli {
+    /// The JSON written by `dump_confluence_divergences -- --out <path>`.
+    #[arg(long)]
+    divergences: PathBuf,
+    /// How many non-grouping exemplars to print per verdict.
+    #[arg(long, default_value_t = 5)]
+    examples: usize,
+}
+
+// ---------------------------------------------------------------------------
+// Input — a subset of `dump_confluence_divergences`'s record
+// ---------------------------------------------------------------------------
+
+/// Only the fields the measurement needs; serde ignores the rest, so this does
+/// not have to track that generator's record shape.
+#[derive(Deserialize)]
+struct Record {
+    id: String,
+    axis: String,
+    core: String,
+    family: String,
+    outputs: Vec<Output>,
+}
+
+#[derive(Deserialize)]
+struct Output {
+    text: String,
+    /// The spellings that reached this output.
+    from: Vec<String>,
+}
+
+// ---------------------------------------------------------------------------
+// Providers — the same construction the generator and the dump use
+// ---------------------------------------------------------------------------
+
+fn padded(core: &str) -> String {
+    let pad = "ACGT".repeat(PAD_OFFSET / 4);
+    format!("{pad}{core}{pad}")
+}
+
+fn reference_for(axis: &str, core: &str) -> MockProvider {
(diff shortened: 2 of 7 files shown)


nh13:
## The corpus measurement, run on this branch

The harness's own claim, asked of the population that exhibits the divergence rather than of hand-written pairs. `generate_cis_confluence_corpus` → `dump_confluence_divergences` → `measure_spdi_key_grouping`, release build, at this branch's tip.

```
divergent classes read: 3246
classes with >1 distinct output (the population the claim is about): 3246

  grouped       3212  (99.0%)
  split            0  (0.0%)
  undecided       34  (1.0%)
  unmeasured       0
```

**The vacuity denominator is non-zero and equals the whole population** — all 3,246 classes have two or more distinct normalized outputs — so the `0 split` is a real result rather than the structural zero this repository has now been bitten by three times (#1456, #1460, #1478).

By family, so no shape hides behind the total:

| family | grouped | split | undecided |
|---|---:|---:|---:|
| same-arity/min-1 | 154 | 0 | 0 |
| same-arity/min-2+ | 243 | 0 | 0 |
| spanning-vs-split/has-0 | 2 | 0 | 26 |
| spanning-vs-split/min-1 | 400 | 0 | 4 |
| spanning-vs-split/min-2+ | 1,659 | 0 | 0 |
| split-vs-split/has-0 | 4 | 0 | 4 |
| split-vs-split/min-1 | 339 | 0 | 0 |
| split-vs-split/min-2+ | 411 | 0 | 0 |

All 34 `undecided` share one reason — `cannot apply to reference — its members overlap, or a stated reference base disagrees with the reference` — and they concentrate in the two `has-0` families, i.e. classes with a zero-separation member pair. That is `spdi_key` declining rather than colliding, which is the contract working: an overlapping allele denotes no single sequence, so there is nothing to key it by.

Worth reading as a bound rather than a verdict on the key: it says the key reconciles every divergent class this corpus can build, and says nothing about shapes the generator cannot emit.

nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1611, opened 2026-08-10
test(rulings): resolve the ledger's own record-to-record citations

nh13:
`ruling_citation_currency.rs` reads `.rs` files. The ledger cross-references *itself* — that is how one record says which neighbouring question it does not settle — and nothing was reading those citations. A record deleted or renamed leaves them pointing at nothing.

Test-only. No file under `src/` is touched.

## Not hypothetical, and the near-miss is the argument

PR #1569 rewrote `adjudication-precedence-order`, a **decided** record, and removed a rank from the precedence order. Its stated ground:

> `partition-is-the-unit-of-normalization` is decided, and the operator has restated it without qualification

That record has never existed on `main`:

```
git log --all -S'partition-is-the-unit-of-normalization' \
  -- tests/fixtures/grammar/hgvs_spec_normalization_overrides.json
```

finds it only on unmerged branches. A decided record was going to be rewritten on the authority of one that was not in the ledger, and the only thing that caught it was a human reading the diff — twice, independently. That is not a mechanism.

## What it found on `main` today

One live dangling citation, and it is a **legitimate** one — which is the interesting half:

```
`inversion-vs-two-delins-76-83` cites `inversion-vs-two-substitutions-76-83`
```

That record opens *"THE RECORD'S FORMER ID AND QUESTION WERE BOTH WRONG … It was `inversion-vs-two-substitutions-76-83`"*, because the rename **is** the argument: the old id asserted the competing members were substitutions, and they are two-base `delins`.

A naive existence check fires here, and this file's own doctrine is explicit that *"every false positive is a test that has to be argued with"*. So retired ids are **declared** in `RETIRED_RECORD_IDS` with the record that carries the question now, rather than inferred from words like "former". That keeps the check free of judgement calls and turns the retired id into a fact the repository states instead of one a reader has to reconstruct. `no_retired_id_is_also_a_live_record` stops an entry being used to silence a real rename.

## Two things that keep it from being decorative

**A looser extractor, justified by measurement.** A rationale cites its siblings by bare backticked id — "see `delins-merge-vs-individual-gap-two-or-more`, which remains decided" — so requiring the word "ruling" (the rule the source-side scan uses) would find almost nothing. The looser rule is safe because of what else lives in a rationale: code identifiers are `snake_case`/`CamelCase` and fail the hyphen grammar, and clause citations carry a `/` and a `.md:` and fail it too. Measured: **8 tokens across 11 records**, 7 live and 1 the declared former id.

**A non-vacuity floor.** The ledger cross-references itself by construction, so a run that checked nothing means the extractor broke — and the test would then pass forever over an empty set. The floor is 4 against a measured 8, far enough below that prose edits cannot trip it.

**And the guard was verified to fire.** Removing the `RETIRED_RECORD_IDS` entry and re-running produces exactly:

```
FAIL ruling_citation_currency::every_record_to_record_citation_resolves
  `inversion-vs-two-delins-76-83` cites `inversion-vs-two-substitutions-76-83`
```

## `rationale` becomes a parsed field, and is required

`Record` gains it, since the check needs something to read. It is **required**, with a mutation test pinning that — matching the file's existing "one mutation per field" pattern. The reason is specific rather than tidiness: an empty rationale cites nothing, so it can never fail, and a ledger that lost its prose would still pass the scan that exists to read it.

## What is deliberately NOT checked, with the measurement

The sibling failure is prose that puts words in a record's mouth. It happened in this same cluster: a module doc quoted `canonical-form-choice-when-both-legal` as saying *"Sequence-level equivalence still needs an answer for consumers who dedupe … a groupable SPDI key, not the canonical string"* — text from an unmerged branch's copy of the ledger, absent from `main`'s.

It is not checked here, for two reasons, both recorded in the module docs so nobody re-derives them:

1. **Attribution is not mechanically decidable.** Prose in this repo quotes the *spec* far more often than a record, frequently in the same paragraph as a record citation. Deciding which a quoted run belongs to is exactly the judgement call the rest of this file is built to avoid.
2. **The population is empty, so the check would be vacuous.** Measured on this tree: blocks citing a record *and* carrying a 40+ character quoted run *and* naming no spec clause — **zero**. A guard over an empty population reads as coverage and provides none, which is what `the_index_is_not_vacuous` exists to name.

Shipping it would have been the corpus-zero mistake this repository has already been bitten by three times (#1456, #1460, #1478), one level up. If the population ever becomes non-empty, the missing ingredient is an explicit attribution convention in the prose, not a cleverer heuristic.

## Verification

`cargo fmt --all --check` clean. `cargo clippy --features dev --all-targets -- -D warnings` clean. Full suite **9,839 passed, 0 failed, 150 skipped**, nothing re-blessed. Plus the mutation test above, which is the only evidence that a passing guard guards anything.

Representation-Change: none. Test-only — one new parsed field on the ledger reader and two new guards over the ledger's own prose; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched and no normalized output moves.

### DIFF
--- a/tests/it/common/rulings.rs
+++ b/tests/it/common/rulings.rs
@@ -70,6 +70,13 @@ pub struct Record {
     /// `decided` or `undecided`, verbatim from the ledger. No other value
     /// parses; see [`STATUSES`].
     pub status: String,
+    /// The record's `rationale` prose, verbatim.
+    ///
+    /// Exposed because a rationale cites *other records* — that is how the
+    /// ledger cross-references itself — and those citations are not covered by
+    /// the source scan in `ruling_citation_currency.rs`, which reads `.rs` files
+    /// only. See `every_record_to_record_citation_resolves` there.
+    pub rationale: String,
     /// Every clause the record names, in the ledger's own order, each tagged
     /// with the role the record's verdict fields give it.
     pub citations: Vec<Citation>,
@@ -180,6 +187,8 @@ fn records_from_value(value: &serde_json::Value, origin: &str) -> Vec<Record> {
                  value is a status no citation could agree with"
             );
 
+            let rationale = required_str(record, "rationale", &format!("record {id}")).to_string();
+
             let governing = optional_str(record, "governing", &id);
             let deviates_from: Vec<&str> = match record.get("deviates_from") {
                 None | Some(serde_json::Value::Null) => Vec::new(),
@@ -248,6 +257,7 @@ fn records_from_value(value: &serde_json::Value, origin: &str) -> Vec<Record> {
             Record {
                 id,
                 status,
+                rationale,
                 citations,
             }
         })
@@ -296,6 +306,7 @@ fn one_valid_record() -> serde_json::Value {
         "rulings": [{
             "id": "a-test-record",
             "status": "decided",
+            "rationale": "A test record, with prose that cites no other record.",
             "governing": "docs/a.md:1",
             "clauses": [{ "clause": "docs/a.md:1" }],
         }],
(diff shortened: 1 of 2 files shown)


nh13:
## Interop with #1585, which no CI can see

#1585 takes the ledger from 11 records to 14, and this PR adds a guard that reads every record's rationale. Neither run exercises the other, so measured directly against #1585's ledger:

```
14 records · 14 record-to-record citations checked · 0 unresolved
```

So the four records #1585 adds introduce no dangling citation, and the non-vacuity floor here (4) is comfortably met at 14 — it does not need raising when #1585 lands.

Worth noting for whoever merges second: `git merge-tree` reports conflicts between these two branches in `hgvs_spec_normalization_overrides.json`, `clause_ruling_index.rs` and `hgvs_spec_normalization_tests.rs` — but **this PR touches none of those three files**. They are #1585's existing conflict with `main` showing through (`main` has gained 9 commits since it branched). This PR's two files, `tests/it/common/rulings.rs` and `tests/it/ruling_citation_currency.rs`, are untouched by #1585, so the two are independent in both directions.

nh13:
@coderabbitai pause

Issue #1607, opened 2026-08-10
normalize: a protein delins trimming to the initiation codon emits an unparseable `p.Met1Val`


nh13:
`Normalizer::normalize` emits `p.Met1Val` — a description `parse_hgvs` refuses — for a protein `delins` whose common affixes trim down to the initiation codon.

## Reproduction

Hermetic; no prepared reference, no environment variable. `main` at `7118eb2a`.

```rust
use ferro_hgvs::{parse_hgvs, MockProvider, Normalizer};

let mut provider = MockProvider::new();
provider.add_protein("NP_003997.1", format!("MAA{}", "G".repeat(20)));
let normalizer = Normalizer::new(provider);

let input = "NP_003997.1:p.Met1_Ala3delinsValAlaAla";
let out = normalizer.normalize(&parse_hgvs(input).unwrap()).unwrap().to_string();
// out == "NP_003997.1:p.Met1Val"
assert!(parse_hgvs(&out).is_err());   // passes — the output does not re-parse
```

Observed output:

```
NP_003997.1:p.Met1_Ala3delinsValAlaAla  ->  NP_003997.1:p.Met1Val
  reparses: NO -- a variant that changes the translation initiation codon is not
  described as a substitution or as an extension; use `p.0`, `p.0?` or `p.(Met1?)`,
  or -- when an upstream initiation site is activated -- the insertion form
  `p.Met1_Leu2ins...` (protein/substitution.md:49, checklist.md:65,
  protein/extension.md:28)
```

## Cause

`try_protein_delins_canonicalize` (`src/normalize/mod.rs:6065`) trims the common prefix and suffix of the reference span and the payload. Here the reference span is `Met-Ala-Ala` and the payload `Val-Ala-Ala`, so the common suffix is two residues, the residual is `Met` → `Val`, and the `1 -> 1` residual takes the "sub > delins" branch and renders as a substitution — at position 1.

The input parses because `validate_no_start_loss_substitution` keys on a single certain position and the input is a range. The illegality is created by the canonicalization, so the guard belongs in the producer.

## Scope, and what this is *not*

This is **not** the split move. `#1606` adds an equal-length split with its own residue-1 guard, and that guard cannot reach this case: the split requires two or more changed runs, and this input has one. The two paths need separate guards, and this one is present on `main` today, independent of `#1606`.

[...]

PR #1614, opened 2026-08-11
fix(normalize): coalesce protein cis adjacency created by the 3'-shift, and add the corpus axis that found it

nh13:
`examples/dump_normalized_corpus.rs` is the harness a representation-moving PR is required to measure with. It could not emit a single protein row, so **any protein-scoped change measured a structural `0 of 78,298`** — which is byte-identical, in the output, to a change that genuinely moved nothing.

This is the **fourth instance of the corpus-blindness class**, after member geometry (#1456), scale (#1460) and transcript geometry (#1478). The corpus had three axes (`g`, `c`, `cx`), which is enough to look well covered.

## The cost is live, not hypothetical

#1606 moves an equal-length protein `delins` onto its members. It could not quote this harness, and said so honestly in its own body:

> `examples/dump_normalized_corpus.rs` generates **no protein rows at all** — a structural zero, reported rather than quoted.

That is the right way to handle a blind corpus, and it is still a measurement nobody has. With the axis in place, #1606 measures:

| | rows |
|---|---|
| total | 85,642 |
| **moved** | **1,584 (1.8%)** |
| of which previously **accepted** (a migration) | 1,584 |

| family | rows | moved |
|---|---|---|
| `protein_equal_length_delins` | 1,584 | **1,584** |
| `protein_shift_del` | 1,728 | 0 |
| `protein_shift_dup` | 1,728 | 0 |
| `protein_ins_becomes_dup` | 864 | 0 |
| `protein_cis_separated` | 1,440 | 0 |
| *(all 16 nucleotide families)* | 78,298 | 0 |

Every moved row is in the one family that describes #1606's shape, and no row moves anywhere else. That is the measurement, and it is the shape of answer the corpus exists to give.

## Five families, each grounded in a defect

| family | rows/core | shape |
|---|---|---|
| `protein_shift_del` | 36 | #91 — a deletion inside a residue run |
| `protein_shift_dup` | 36 | #91 / #92 — a duplication inside a run |
| `protein_ins_becomes_dup` | 18 | #92 — an insertion whose payload copies its 5' neighbour |
| `protein_equal_length_delins` | 33 | #1606 — equal-length `delins`, interior unchanged |
| `protein_cis_separated` | 30 | #1232 / #1401 on the protein axis — two members, unchanged residues between |

7,344 rows over 48 peptide cores. **Every pre-existing row is byte-identical** — verified with `cmp` against a baseline dump, not asserted — because an added axis only appends, exactly as `LONG_FAMILY` and `REVCOMP_FAMILY` do. Old dumps stay comparable for every nucleotide family.

## Three design decisions worth review

**One shuffle direction, and it is measured rather than assumed.** Every other loop crosses its cores with a 3' and a 5' cell. The protein path never reads `NormalizeConfig::direction`, so a 5' cell would be 7,344 byte-identical rows — and worse, a dump that *looks* like the axis was measured under 5' shifting. `the_protein_axis_is_direction_invariant` compares both directions across every family on two cores (>500 rows) and fails if that ever stops holding, naming the fix: add the cell.

**Runs are planted, not hoped for.** Every family is about shifting, and a shift needs a tract of equal residues to move through. Drawing residue-by-residue from a 6-letter alphabet gives a run of 3 about once per 36 positions, so most cores would be fixed points and the corpus would report a near-zero it had *generated* rather than measured — the same class one level down. Residues are drawn in runs of 1–3 instead. Measured floor across the 48-core default corpus: longest-run **3**, ceiling **16**, both pinned. A first cut asserted `>= 2`, which every core clears by a margin and which a regression to per-residue drawing would likely clear too; `the_planted_runs_are_longer_than_chance_would_give` pins the real distribution so the bound is not folklore.

**No family reaches residue 1.** Residue 1 is `Met`, and an edit reaching it emits the illegal start-loss spelling filed as **#1607**. `Met` is therefore never drawn into a core body, so a shift cannot arrive there either — otherwise thousands of rows would be measuring a known bug rather than a representation choice. Pinned by `no_protein_family_touches_residue_one`.

## Guards

Seven new tests, all in the example's own module (`test = true` is already set on the target):

- `the_protein_axis_is_direction_invariant` — the pin that justifies one direction
- `no_protein_input_is_rejected` — **the one most likely to matter.** Normalization checks each spelled residue against the reference, so one wrong three-letter code turns a family into `<declined>` rows: a populated-looking corpus measuring nothing, which no other test would catch — the rows exist, the key is unique, the family is non-vacuous
- `every_protein_family_emits_rows` — no silent zero
- `the_protein_corpus_is_prefix_stable` — `--seeds` truncates to a strict prefix, as the DNA half does
- `protein_cores_carry_runs_a_shift_can_move_through` / `the_planted_runs_are_longer_than_chance_would_give`
- `no_protein_family_touches_residue_one`

The existing `the_corpus_key_is_unique` covers the protein rows too, which is what rules out two families emitting the same input for one core.

## Verification

`cargo fmt --all --check` and `cargo clippy --features dev --all-targets -- -D warnings` clean. Full suite green with nothing re-blessed. Baseline/candidate dumps diffed with `cmp` to prove the first 78,298 rows are unchanged. Zero protein rows are parse errors, declines or panics.

## Summary by CodeRabbit

* **New Features**
  * Added protein-level variant normalization support.
  * Expanded coverage across multiple protein sequence shapes and 3′-direction inputs.
  * Added deterministic peptide and residue generation for broader validation.

* **Bug Fixes**
  * Improved handling of adjacent protein changes created by 3′ shifting.
  * Ensured normalization remains stable and repeatable.
  * Added safeguards against invalid inputs and edits affecting the first residue.

* **Documentation**
  * Updated documentation with protein-axis coverage and measurements.

## Scope grew: this PR now also fixes the defect its own axis found

The protein axis was added to make protein-scoped changes measurable. Armed with
`FERRO_ASSERT_IDEMPOTENT=1` it immediately failed on its own corpus, which is the axis doing its
job:

```text
input   NP_TEST.1:p.[Gly13del;Gly16Ala]
pass 1  NP_TEST.1:p.[Gly16Ala;Gly17del]     <- the deletion shifted 13 -> 17
pass 2  NP_TEST.1:p.Gly16_Gly17delinsAla    <- and only now merged
```

`coalesce_protein_adjacent_changes` ran once, on the **authored** members, before the
collapse/merge/shift loop — so it only ever saw adjacency the input already had. The per-member 3'
shift then *created* more of it, and the nucleotide `merge_consecutive_edits` inside the loop never
fires on protein members, so nothing closed the gap.

Pass 2 is the correct answer rather than a matter of taste: `protein/substitution.md:32` marks the
split spelling of two adjacent members `class="invalid"`, and the decided
`rulings[delins-adjacent-members-when-both-consume-reference]` record scopes that ruling to members
which both consume reference bases — a substitution and a single-residue deletion both do. So the
fix coalesces **inside** the loop, where `respell_colliding_duplications` already sits for the
identical documented reason.

`tests/it/protein_shift_created_adjacency.rs` pins it, including a control asserting the deletion
really travels 13 -> 17 — without which the merge assertion could pass because the *shift* stopped
rather than the merge starting. Mutation-tested: disabling the fix reddens exactly that guard.

Two corpus-figure corrections came out of measuring, and one reviewer finding is rejected:

* `78,028` and `78,298` are **both correct** and are not a transposition of each other.
  `separated_revcomp_runs` is exactly 270 rows and `78,298 - 270 = 78,028`, so the smaller figure is
  the corpus *before* that family existed — the only honest denominator for a measurement taken
  then. A note now says so, since the file used two totals without ever explaining it.
* Genuinely stale, now re-measured: `20,526 of 78,028` -> **`20,516 of 78,298`**, and `286 rows out
  of 78,028` -> `78,298` (the `286` numerator is current: 270 + 16).
* `every_row_carries_the_sequence_it_was_drawn_against` asserted `core` is over `ACGT`; a protein
  row carries a peptide. The alphabet is now per-axis rather than protein rows being exempted, so
  the protein half of the corpus keeps a real check.

Validated as CI runs it, not with oracles armed globally: plain `test` **10147/10147**, and
`test-oracle` (armed, `ORACLE_EXCLUDE` negated) **10085/10085**.

Representation-Change: 171 of 85642 corpus rows move (0.2%), all in `protein_cis_separated`. The forms: a two-member protein bracket allele whose members became adjacent under the 3'-shift, `p.[Gly16Ala;Gly17del]`, becomes the merged delins `p.Gly16_Gly17delinsAla`. Direction: TOWARD the already-shipped form, not away from it — those 171 rows were not fixed points, so a second normalization pass already produced the merged form; the change moves pass one onto the answer pass two was giving, and toward the spec's stated form, since `protein/substitution.md:32` marks the split spelling `class="invalid"` (the decided `delins-adjacent-members-when-both-consume-reference` record scopes that to members which both consume reference bases, as a substitution and a single-residue deletion do). 0 of the moved rows were previously accepted forms: all 171 were not fixed points, so nothing downstream holds them and the move is free. Measured by dumping this branch before and after the fix over the identical 85,642-row corpus.

### DIFF
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -83,6 +83,32 @@
 //!    brings its own designed references because a random core cannot be asked to
 //!    have a coincidence pattern.
 //!
+//!    **Two corpus totals appear in this file and both are correct — do not
+//!    "reconcile" them.** `78,028` is the corpus *before* `separated_revcomp_runs`
+//!    existed, which is the only denominator a measurement taken before that
+//!    family was added can honestly carry; `78,298` is the corpus now, and
+//!    `78,298 - 78,028 = 270` is exactly that family's row count. They differ by a
+//!    digit transposition as well as by 270, which is a coincidence and has
+//!    already been read once as a typo in one of them. When you quote a figure,
+//!    quote the denominator it was measured against.
+//!
+//! 6. **…and only on the molecule types it builds** (#1606). Every family above is
+//!    spelled from nucleotides, so until the protein axis was added this corpus
+//!    could not emit one `p.` row and a protein-scoped change measured a structural
+//!    `0 of 78,298`. #1606 is the live instance: it moves an equal-length protein
+//!    `delins` onto its members and had to *report* the zero and name the generator
+//!    as its cause, because there was no measurement to quote. With the axis in
+//!    place the same change measures **1,584 of 85,642 rows, all in
+//!    `protein_equal_length_delins` and none anywhere else**.
+//!
+//!    Note what the four instances have in common and why the list keeps growing:
+//!    fixing one blindness does not reveal the next. Geometry (#1456) was only
+//!    findable once the conflict families existed, scale (#1460) once geometry was
+//!    fixed, transcript geometry (#1478) once scale was — and molecule type was
+//!    invisible behind all three, because a corpus with three axes looks well
+//!    covered. `compare` names the families it covered (#1459), which catches a
+//!    missing *family*; it cannot tell you the axis you never built.
+//!
 //! ## One family knows its own ground truth, and that buys three oracles
 //!
 //! Everything above measures **movement** — two dumps, one diff — and never
@@ -178,6 +204,40 @@ const EXON_SPANS: &[(usize, usize)] = &[(0, 7), (7, 14), (14, 20)];
 /// Long enough that an exon's genomic span cannot abut its neighbour's, so a shift
 /// that runs off an exon end lands in intron rather than silently in the next exon.
 const INTRON_LEN: usize = 10;
+
+/// The protein reference a `p.` row is drawn against (#1606).
+///
+/// A bare `NP_` accession with no nucleotide counterpart, which is deliberate: it
+/// takes `MockProvider::add_protein` and nothing else, so the protein axis needs
+/// neither a transcript nor a CDS and cannot be perturbed by the exon geometry the
+/// `c.`/`cx.` references carry.
+const PROTEIN_ACCESSION: &str = "NP_TEST.1";
+
+/// Residues protein cores are built from, as `(one-letter, three-letter)`.
+///
+/// `Met` leads the table because it is residue 1 of every core and is **never
+/// drawn** into the body — cores draw from `[1..]`. Two reasons, and the second is
+/// the one that would silently weaken the corpus. A protein starts at `Met`, so a
+/// core that did not would not be a protein; and an edit reaching residue 1 emits
+/// the illegal start-loss spelling filed as #1607, so a `Met` in the body would let
+/// a shift *arrive* at residue 1 and turn a representation measurement into a
+/// rediscovery of that bug. Keeping `Met` out of the body means the only residue-1
+/// interaction is one the families never construct.
+const RESIDUE_CODES: &[(char, &str)] = &[
+    ('M', "Met"),
+    ('A', "Ala"),
+    ('G', "Gly"),
+    ('S', "Ser"),
+    ('L', "Leu"),
+    ('K', "Lys"),
+    ('V', "Val"),
+];
+
+/// Residues in a protein core, `Met` included.
+///
+/// Twenty, matching the DNA cores, so the two halves of the corpus report row
+/// counts on the same scale and a per-core figure means the same thing on either.
+const PROTEIN_LEN: usize = 20;
 
 /// The dump header, written on emit and required verbatim on read.
 const HEADER: &str = "reference\taxis\tdirection\tfamily\tinput\toutput\twas_fixed_point\n";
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Both review threads addressed in `06bff4f6`, plus one citation added.

**`tests/it/protein_shift_created_adjacency.rs` — the finding was right, and it is worth recording how right.** `authored_adjacency_still_coalesces` compared the two member orders *to each other*, which proves order-independence and nothing else. Measured rather than argued: with `coalesce_protein_adjacent_changes` forced to return `None`, both spellings emit the identical `NP_TEST.1:p.[Gly16Ala;Gly17del]`, so the assertion **passed** under total loss of the behaviour it is named for. The reversed spelling is now pinned against the merged literal, which makes each direction falsifiable on its own; order-independence survives as two literals that agree.

**`src/normalize/mod.rs` — gated on the helper's own preconditions, not on `is_cis` alone.** `coalesce_protein_adjacent_changes` declines on `variants.len() < 2` or on any non-protein member, so the candidate is now built only when both hold. That is the helper's whole-allele precondition set minus phase rather than an approximation of it, so it cannot skip a case the helper would have merged.

**Separately: the shift half of this change had no clause cited, and it turns out to have one.** The comment cited `protein/substitution.md:32` for merging two adjacent members but nothing for the 3'-shift that *creates* the adjacency. `general.md:157-160` is that clause, stated as a Q&A rather than as a rule, which is why it was missed:

> It means that protein variant descriptions should be derived from comparing the variant protein sequence with the reference protein sequence. Knowledge on the underlying change on the DNA level should not be used.

Its worked example is a protein-level 3'-rule application: a `Ser` lost from a `SerSer` run is `p.Ser5del`, and the passage says in as many words that placing it at the deleted *codon*, `p.Ser4del`, "is not correct". So carrying `Gly13del` down its Gly run to `Gly17del` is obligatory rather than incidental, and `substitution.md:32` governs what is made of the result. Both halves are now published clauses; neither rests on judgement.

Validated at CI parity on this tree: `cargo fmt --check` clean, `clippy --all-features --all-targets -D warnings` clean, plain suite **10147/10147**, oracle suite (three flags armed, `ORACLE_EXCLUDE` negated as `test-oracle` does) **10085/10085**.

nh13:
@coderabbitai review

nh13:
Trailer strengthened — but two of the three stated grounds do not hold, and recording which is which matters more than the fix.

**The trailer was already present**, as the last non-empty line of the description, and `Representation change declared` is one of the ten required contexts and was **passing** — so "the supplied PR summary has neither required statement" is checkable and false. Worth naming because a review that reports an absent trailer when the gate for it is green invites the author to distrust the gate.

**The oracle-direction instruction cited is out of scope here.** In `.coderabbit.yaml` that clause attaches to `tests/fixtures/**/baseline-failures/*.txt`; this PR touches no file under that glob. Applying a path instruction to a path its glob does not match is the one error that makes path-scoped review worse than none.

**What was genuinely missing is narrower, and comes from `CONTRIBUTING.md:138-152`** rather than from either source cited. It lists four things a trailer must carry; this one had #3 (count, and the corpus measured over) and #4 (previously-rejected vs previously-accepted) but only implied #1 (which forms moved) and omitted #2 (**in which direction**). Both are now stated: `p.[Gly16Ala;Gly17del]` → `p.Gly16_Gly17delinsAla`, and the direction is **toward** the already-shipped form — those 171 rows were not fixed points, so a second pass already produced the merged form, and this change moves pass one onto the answer pass two was giving.

Verified by running `scripts/check_representation_change.py` against the new body directly, rather than waiting to see whether the gate stayed green.

PR #1616, opened 2026-08-11
fix(normalize): delete the input-relative weight bound

### LINKED ISSUE #1419: Normalizer: multi-member deletion cis-allele not re-derived from the resulting sequence — disagrees with the equivalent `delins` (partition non-uniqueness); non-confluent

### LINKED ISSUE #1420: Normalizer: multi-member cis-alleles normalized independently, not re-derived from the resulting sequence — `[dup;del]`, `[ins;del]`, `[delins;del]` each still non-confluent (follow-up to #1231/#1233/#1232)

### LINKED ISSUE #1421: Normalizer: net-insertion spanning `delins` not split across an unchanged interior base (follow-up to #1232); non-confluent

### LINKED ISSUE #1603: 5' junction filter in clamp_sibling_crossing_shifts loses its only guard when the weight bound is removed
`clamp_sibling_crossing_shifts`' 5' `across_junctions` barrier carries a `.filter(|_| a.junction.is_none())`: a member that carries its own junction is bounded on the right edge by `clamp_sibling_crossing_junctions` instead, and bounding its *start* here would hold it back by `len - 1` more than the invariant needs. The filter is not sequence-affecting, so the exhaustive sweeps cannot see it; it is held by a pinned string.

Today that pin exists and works. On `main` (and on #1598's head `04e38d69`), deleting the filter fails exactly one test:

```
cis_junction_crossing_shift::a_third_member_clear_of_the_tract_keeps_the_duplication_reaching_its_five_prime_most_position
  TEMPLATE:g.[4_5insC;5_6dup;15del] in FivePrime
    expected TEMPLATE:g.[4_5insC;4_5dup;15del]
```

**Deleting the input-relative weight bound in `canonicalize_from_sequence` takes that guard away.** Measured in a scratch worktree at `04e38d69` with the bound removed:

- the guard's pin moves — the sequence-first derivation now merges the members and emits `TEMPLATE:g.[5_10delinsCTATAT;16A[3]]`, so it will be re-blessed by whichever PR removes the bound;
- with the bound removed, deleting the filter as well leaves the failure set **byte-identical**: 27 failed / 9,766 passed over the full suite with `FERRO_SWEEP_SEEDS=full`, both with and without it. Not one test discriminates.

Both guards for this filter, and the same device in `issue_1261`, `issue_1301`, `issue_1304`, `issue_1320` and `cis_junction_crossing_shift`, rest on "a third member clear of the tract stops the merge" — and the thing that stopped the merge was the weight bound.

An unguarded filter in this code family is how the wrong-sequence class keeps surviving: #1592 and #1600 were both fall-throughs between these same passes, and neither was visible to any seam oracle.

Restoring coverage needs a shape the derivation still declines without the bound — a cross-exon-junction `c.` allele (`crosses_exon_junction` declines on the input span) looks like the most promising candidate, since that refusal is not weight-related. That is separate work from removing the bound, and it should land with it rather than after it.

nh13:
> **STATUS: the frameless `general.md:34` residual is CLOSED, not declared.** Rebased onto `origin/main` (`f17a0895`). This PR now deletes the input-relative weight bound **and** carries the two gates the deleting ruling's own scope paragraph asks for, so `spec_conformance_axis`'s rejected-SVD-WG010 guard reads **0 of 210 guarded rows in both directions** — the value `main` pins — rather than 18 or 2.
>
> **This corrects the STATUS block this description used to carry.** That block called the residual "a README rule-1 bug to close". That framing predates and conflicts with `rulings[separation-rule-force-modal-or-negation]` (decided 2026-08-12), which holds that `general.md:34`'s modal grades the whole clause and its "and not" names the excluded alternative rather than prohibiting it. So `:34` is README **rule 2** — recommended form, best effort — and such an output is a deviation to disclose and tripwire, **not** a rule-7 blocker. Rule 2 governs only the *release* consequence; it does not license the departure, and because rule 2 "outranks maintainer judgment" an unjustified departure is still a **defect to fix**. That is what this PR does. Nothing about that record is disturbed.

> **BOTH 2026-08-13 OPERATOR RULINGS ARE NOW APPLIED**, and the two red `spec_conformance_axis` census tests — the only two of the thirteen that gating CI could see — are **green**. #1710's `coding_axis_separation_two_or_more_merges` is re-pinned **0 -> 8** as a *disclosed rule-2 deviation* with the pin itself as the tripwire (new ledger record `coding-axis-merges-are-a-disclosed-general-34-deviation`), and `NONCODING_NON_IDEMPOTENT` is re-pinned **21 -> 51** with the 29-projector / 1-defect split stated at the constant. Neither is a re-bless for a green build; both are disclosures with their authority named. See **Ledger** and **Tests**.

## Prerequisites — both have landed, and this is rebased past them

**#1800** (`equivalence_cross_axis_rung` exercised directly rather than via a naturally-diverging corpus pair) and **#1802** (`reported_partition_verdicts` cannot express a pair both spellings fix) were prerequisites: nine of this branch's red tests were theirs to fix. **Both are now closed** — #1800 by **#1809**, #1802 by **#1805** — and this branch is rebased onto `origin/main` at `f17a0895`, which carries both. **Measured after the rebase: #1800's three cleared and #1802's did not.** All three `equivalence_cross_axis_rung` tests now pass, as do `reported_confluence_pairs`' two and the `cis_junction_crossing_shift` and `spec_corpus_regressions` rows. `reported_partition_verdicts` did **not** clear — it now reports seven failures rather than five, because #1805's closed three-state pair model asks a question of four rows that the old table could not express, and those rows are exactly the ones this PR moves. They are enumerated under **Tests** and none is re-blessed. This PR kept `base: main` and was **not** stacked on either — the dependency was self-enforcing, because #1616 stayed red until they landed, and stacking has twice cost this repo real time (#1634 read as 32 commits / 26 files / ~14k lines because it was stacked; #1747 is blocked right now on #1735).

## What changed

Three commits in the partitioner, each with its own authority, plus one in the projector.

### 1. Delete the input-relative weight bound

```rust
if derived_columns > changed_columns_of_edits(&edits) { return None; }
```

Required by the decided `rulings[derivation-may-not-be-bounded-by-the-inputs-spelling]`: no clause compares a description to the **input**; the recommendations compare it to the sequence, and `background/basics.md:38`'s design values omit minimality. The comparand contradicted `canonical-form-choice-when-both-legal` in terms.

### 2. Scope `DNA/delins.md:44-47`'s payload-coincidence carve-out to `c.` **inside the shipping partitioner**

The carve-out was honoured by `payload_coalesce_applies`, which gates `PartitionRule::CanonicalCoalesced` — an arm that does not ship — and contradicted by `partition_block`, which does, through two of its three single-piece exits. `rulings[delins-payload-coincidence-carve-out-is-coding-dna-scoped]` scopes that passage to `c.` **and nothing else**.

Gated on the axis **kind** (`AxisFrame::is_coding_dna`), deliberately **not** on `AxisFrame::reading_frame` — that is true for a coding `r.` too, which is how #1711 arose and which that ruling names by name.

### 3. Refuse to assert an **unexamined** block as one member off `c.`

`MAX_SPLIT_BLOCK` (1 024) makes `partition_block` short-circuit on length and return the whole block *before any rule about the derived pieces runs*. What comes back is therefore **not a finding that the block holds one variant — it is the absence of a finding.** Emitting it asserts "there is no separation here" about a block nothing examined.

Off `c.` that assertion has no licence: `general.md:34` governs it and `:44-47` cannot reach it. So `canonicalize_from_sequence` declines such a block and the per-member pipeline answers — the individual form `:34` asks for. On `c.` the carve-out is in reach and `:47` recommends the spanning `delins`, so nothing there moves.

**`MAX_SPLIT_BLOCK` itself is untouched.** Raising it is a cost question with no clause behind it and it stays open; this changes only what may be *done* with the block the cap hands back.

#### Why none of this is the deleted bound under another name

The deleted bound's comparand was **the input's own member list**. Neither gate has a term for the input: gate 2 reads the derived pieces and the axis, gate 3 reads the block's two lengths against a constant and the axis. The spanning spelling and the split spelling of one variant reach both identically — which the deleted bound could not do, and which `issue_1616_unexamined_block_stays_split` pins on both spellings for exactly that reason.

The cap is stated **once**, as `past_the_split_cap`, consulted by `partition_block` and by its caller, so the two cannot disagree about which blocks were examined.

### 4. `p.` — build the combined product when a cis pair moves the terminator

`NM_FSSTOP.1:c.[4_5delinsTA;9A>G]` projected to `NP_FSSTOP.1:p.[(Val2Ter);(Lys3=)]`. `Lys3` is not translated at all in a product that stops at residue 2, so the bracket described a protein that does not exist; `protein/frameshift.md:22` forbids `fsTer1` and requires the nonsense substitution `p.(Val2Ter)`.

**The cause is the dispatch, not either builder.** `combined_cis_residue_changes` reported "the combined protein changes length" — a premature stop or a stop-loss — as the same `None` as every other decline, and the caller's decline path is to describe the members **separately**. Each per-member prediction is correct *about the reference* (`c.4_5delinsTA` alone is `p.(Val2Ter)`, `c.9A>G` alone is silent); bracketing two correct per-member answers produced an incorrect combined one. Neither earlier arm claims this shape — the net-frameshift arm needs `net_frameshift || (net_in_frame && any_simple_fs)` and this pair is net 0 with neither member carrying a frameshift consequence; the codon path needs every member to be a single-base substitution.

The decline is now split at the type level (`CisResidueOutcome::{Diff, TerminatorMoved, NotResidueWise}`) and `TerminatorMoved` routes to `try_project_cis_combined` — the same builder the net-frameshift arm uses.

**Same reason for surfacing as the guard movement; a different defect.** With the bound in place, `c.[4del;10dup]` was refused re-derivation, kept its authored members and answered correctly. Deleting the bound converged it and `c.4_12delinsTAAAAGGGG` onto one DNA description — the point of the deletion — and that description reached the broken arm. The single-member spelling already reached it on `main`, where the projector test pinned `p.[(Val2Ter);(Lys3=)]` as a divergence "NOT adjudicated correct", precisely so a fix would have to come past it. It has: both sides now read `p.(Val2Ter)` and the test asserts their **equality** as well as the literal. `p.` is excluded from the confluence gate by `rulings[confluence-gate-is-apply-equality-on-every-determined-axis]`, so no equivalence check could have found this.

## `guard_violations`: 0, and the instrument is known to move

`spec_conformance_axis`, **210 guarded rows** of 11,882 classes / 58,552 outputs, `FERRO_MANIFEST` set, both spec artifacts regenerated with `--bin`. The denominator is asserted non-zero (`0 of 0` is what a rebuilt #1456 looks like), and the zero is **measured, not structural** — the same instrument on the same corpus read 18 after commit 1 and 2 after commit 2:

| | `main` | after commit 1 | after commit 2 | **this head** |
|---|---:|---:|---:|---:|
| `guard_violations` 3' | 0 | 18 | 2 | **0** |
| `guard_violations` 5' | 0 | 18 | 2 | **0** |

`RESIDUAL_GUARD_VIOLATIONS` is now **empty** and is kept rather than deleted: `the_residual_guard_violations_are_exactly_these_rows` asserts the set both ways, so empty is the strongest form of that assertion and is where the next regression lands with its id already printed.

Attribution, measured by gating each exit alone rather than argued: `separations_are_meaningful`'s raise to `RAISED_PIECE_SEPARATION` accounts for 14 rows, `split_buys_no_higher_priority_type` for 2, and the `MAX_SPLIT_BLOCK` length cap for the last 2 (`scale-g-block1032-delins-del`, `scale-g-block1032-inv-del` — 2 065-base blocks, the only rows in the corpus over the cap, which is also why no smaller-scale corpus can see this class at all, #1460).

## Convergence, with the loss stated as a loss

| figure | `main` | this head | Δ |
|---|---:|---:|---:|
| `converged` 3' | 9,402 | **10,977** | **+1,575** |
| `converged` 5' | 9,228 | **10,775** | **+1,547** |
| `split_two` 3' | 2,223 | 695 | −1,528 |
| `split_two` 5' | 2,462 | 959 | −1,503 |
| `split_three` 3' / 5' | 226 / 168 | 181 / 124 | −45 / −44 |
| `split_more` 3' / 5' | 31 / 24 | 29 / 24 | −2 / 0 |
| `outputs`, `declined`, `unparseable`, `denoting_no_sequence`, `leaving_the_transcript`, `intronic_under_a_genomic_wrapper`, `prohibition_violating`, `non_idempotent`, `sequence_changed`, all three `*_accepted`, `underdetermined` | — | **unchanged, both directions** | = |

**Gate 3 gives 18 (3') / 14 (5') of that gain back, and that is disclosed rather than netted away.** A block past the cap cannot be split, so its spanning spelling and its split spelling can no longer reach one string: `converged` falls 10,995 → 10,977 and 10,789 → 10,775, with `split_two` rising by exactly those amounts and nothing else moving. Confluence is ferro's own policy (README rule 3); `general.md:34` is a spec preference clause (rule 2), which outranks it. **Raising `MAX_SPLIT_BLOCK` would buy those 18/14 classes back and is the one thing that would get both** — it has no clause behind it, its blast radius over every length-changing block in 1 025..4 096 is unmeasured, and it is left open by name.

## Tests

Full local gate, `FERRO_MANIFEST` set, both artifacts regenerated, `cargo nextest run --features dev --no-fail-fast --test-threads=4`, re-measured at `f17a0895`:

| tree | run | passed | failed | skipped |
|---|---:|---:|---:|---:|
| this head | 10,538 | 10,528 | **10** | 152 |

**The `origin/main` baseline row is not re-run here** and no figure is carried forward from the old one: `main` is green by the merge queue's own required checks, and every attribution below is by test **name** against the modules this PR touches rather than by a count difference.

`guard_violations` is **0 in both directions** in that run, and **both census tests now pass**, so the zero is read off a green test rather than off a report line. That is what the #1710 re-pin buys: those two were the only failures gating CI could see, because the other eleven are manifest-gated and no PR CI job sets `FERRO_MANIFEST`.

**Three of the earlier thirteen are now green**, and only because a ruling authorised the disclosure — not because a pin was edited for a green build: `spec_conformance_axis::three_prime_conformance_census`, `::five_prime_conformance_census` and `mutalyzer_normalize_tests::axis_noncoding_idempotent`.

**The remaining 10 are NOT re-blessed and must not be.** Every one is a pinned normalized-output string that the bound deletion (commit 1) or the carve-out scoping (commit 2) moves, or a counter those moves raise; they are representation adjudications, and this PR deliberately leaves them red rather than editing a pin to make a run green:

| test | what moved |
|---|---|
| `mutalyzer_normalize_tests::axis_genomic` / `::axis_normalized` | **one divergence each, and it is the same input** — `NG_012337.1(NM_012459.2):c.[10del;23_25del;36_37insAAT]`, where ferro's last member re-partitions (`c.[10del;23A>T;26T>C;28_36delinsAGCCTGAAT]` against Mutalyzer's `c.[10del;23_25del;37_38insATA]`). The `NG_008939.1` family this PR moves is **registered, not red**: 28 (input, axis) pairs across 16 inputs — 15 `genomic`, 13 `normalized` — carrying `accepted_divergence` with Mutalyzer's own answers retained as evidence. This row is not one of them, and under `adjudication-precedence-order` Mutalyzer is not the authority — but the divergence needs registering, not asserting |
| `multi_member_cis_axis::axis_normalized` | **an improvement the census will not absorb silently**: `respelling_converged` rises **401 -> 407**, six rows starting to converge (e.g. `NM_000518.5:c.[385_388delinsCCACA;397_407del]`). The test asks for `AXIS_CENSUS.respelling_converged` and `CONVERGED_ROWS` to be re-blessed together with the rows named; that is a convergence claim requiring the named clause, so it is left red. Note this axis is a pure function of HEAD and **no CI job sets `FERRO_MANIFEST`**, so it is invisible to PR CI either way |
| `reported_partition_verdicts::*` (7 tests) | four rows, all of them a **gap-row closure that `OPEN_GAPS` requires a named clause for**, so none is re-pinned here. `1420-v3/cis` now prints `g.[37_38delinsCA;40G>T]`, which is its own `wanted` form, so the pair measures `BothReach`. `1421-n1/span` and `1421-n3/span` are pinned `Gap` while their recorded `output` already equals their `wanted` — the convergence this PR produces, recorded in the `output` column without the verdict being flipped. `1421-n2/split` is pinned `Canonical` while its `output` and `wanted` differ, which is a different question again. Flipping the three `Gap` rows alone would take `OPEN_GAPS` from 11 to 8 — but `1420-v3/cis` needs its `output` pin moved as well as its verdict, and `1421-n2/split` needs one of its two fields chosen between. Four adjudications, and they belong to whoever makes them |

### Now GREEN, by ruling rather than by re-bless

- `spec_conformance_axis::three_prime_conformance_census` / `::five_prime_conformance_census` — the #1710 instrument was the **only** assertion firing in either census; every other counter already matched its pin exactly, including `converged` and all three `split_*`. Re-pinned **0 -> 8** as a disclosed rule-2 deviation, with the pin itself as the tripwire. See **Ledger**.
- `mutalyzer_normalize_tests::axis_noncoding_idempotent` — the `n.`-axis residue (#1712) re-pinned **21 -> 51** of 681 axes. **The number is only ever to be quoted as a split, 29 / 1**, and both halves are counted directly off the residue rather than estimated: **29** are the projector emitting a spanning `n.192_197delinsGTCCTGTGCTCATTATCTGGC` on the `NG_008939.1` family that re-normalization correctly splits to `n.[192_194delinsGTC;196C>T;198_199insGCTCATTATCTGGCT]` — #1712's own thesis, pre-existing projector staleness this PR **exposes rather than causes**, and exposing a defect is not introducing one. **The 30th is a genuine defect and is not projector staleness**: `NG_012337.1(NM_012459.2):c.[10del;23_25del;36_37insAAT]` re-normalizes to `n.[40del;53A>T;56T>C;58_60del;63_64=;67_68insATA]` — the `63_64=` is a **stranded identity member**, the #1655/#1281 shape `tests/it/stranded_identity_member.rs` pins as a defect. Filed separately as **#1829**, disclosed as a bug with an issue rather than as an acceptable output.

### What gating CI actually sees — a correction, measured

Of the 10 remaining, **3 are manifest-gated** and skip green in CI (`mutalyzer_normalize_tests::axis_genomic`, `::axis_normalized`, `multi_member_cis_axis::axis_normalized`) — no CI job sets `FERRO_MANIFEST`. **The other 7 are not.** `reported_partition_verdicts::*` runs on synthetic `TEMPLATE:`/`NM_` fixtures, needs no manifest, and is not in `ci.yml`'s `SWEEP_FILTER`, so the `Test` shards run it and it is red there. Verified by re-running the four modules with `FERRO_MANIFEST` **unset**: 20 tests, 13 passed, the same 7 failed.

**And CI is red on three more that do not exist on this branch's base.** `cargo nextest` at `f17a0895` cannot see them: `cis_adjudication_enumeration::{the_spanning_versus_split,the_adjacent_members,the_dup_flush_carve_out}_class_partitions_as_measured` arrived on `main` with **#1814**, after this branch was rebased, and CI checks out `refs/pull/1616/merge`. They are this PR's partition changes measured against `main`'s new enumeration pins. Closing them needs a rebase onto current `main` and a re-measurement, not a change here.

New guards:

- `tests/it/issue_1616_unexamined_block_stays_split.rs` — the call site, end to end through `Normalizer` on a synthetic 1 100-base contig. The `g.` split stays individual, the `g.` spanning spelling is left alone (**the disclosed confluence cost, pinned as such**), and the `c.` control still spans — without which the gate could be a blanket refusal rather than an axis scope.
- `merge::tests::the_unexamined_block_gate_shares_the_delins_47_axis_scope` — a cross-product equality against `payload_coalesce_applies`, so the three `:47` gates cannot drift apart.
- `merge::tests::a_coding_rna_block_may_not_be_asserted_unexamined` — the coding-`r.` trap, kind not frame.
- `merge::tests::the_split_cap_predicate_matches_what_the_splitter_actually_examines` — pinned against `partition_block`'s observable behaviour, not against a second copy of the arithmetic; includes the equal-length exemption at 1 026 bases.
- `protein::indel::tests::a_cis_pair_whose_combined_product_stops_early_reports_the_moved_terminator` — the routing decision at the seam, with a benign control (`GTA` → `CAA`) whose product runs to the reference terminator and *is* an ordinary residue-wise diff.

### Mutation results, call sites included

A predicate-only battery is what passed 10 233/10 233 clean on #1727 with the call site wired to nothing, so every mutation below is run against the guards as a set:

| mutation | expected | result |
|---|---|---|
| M1 — call-site gate removed (`if false`) | `g.` split test FAILS | **FAILS** — `a_genomic_block_past_the_cap_keeps_its_members_individual` |
| M2 — call site's axis hardcoded to `CisKind::Cds` | `g.` split test FAILS | **FAILS** — same test |
| M3 — gate inverted (decline on `c.`) | `c.` control FAILS | **FAILS** — both halves: the `c.` control *and* the `g.` split |
| M4 — `past_the_split_cap` drops its length-inequality term | equal-length guards FAIL | **FAILS** — both `residual_above_cap_confluence` tests + `the_split_cap_predicate_matches_what_the_splitter_actually_examines` |
| M5 — axis predicate is a blanket allow | `g.` split + scope tests FAIL | **FAILS** — the `g.` split, the coding-`r.` guard, and the three-gate scope equality |
| M6 — carve-out reaches the RNA axis | coding-`r.` guards FAIL | **FAILS** — `a_coding_rna_block_may_not_be_asserted_unexamined` + the scope equality |
| M7 — `TerminatorMoved` folded back into a plain decline | projector test FAILS | **FAILS** — `project_cis_net_in_frame_but_mid_shift_stop_takes_the_frameshift_arm` |
| M8 — the seam stops reporting the moved terminator | projector + seam tests FAIL | **FAILS** — the projector test *and* `a_cis_pair_whose_combined_product_stops_early_reports_the_moved_terminator` |

## Ledger

**28 records (24 decided / 4 undecided)** — `origin/main`'s 27 at this branch's base plus **one new record**, verified by set-difference in both directions: nothing dropped from `main`, nothing dropped from the pre-rebase tip, and the only new id is this branch's own. `tests/it/clause_ruling_index.rs`'s GENERATED block was regenerated from the test's own output on its `BEGIN`/`END` markers, never hand-edited.

### The new record: `coding-axis-merges-are-a-disclosed-general-34-deviation` (operator ruling, 2026-08-13)

The eight coding-axis merges **are** a `general.md:34` deviation. Stated plainly rather than reasoned away, disclosed, and pinned with a tripwire. The chain, all of it measured at this base:

1. **No member is gap-bearing.** All eight are pure deletions, all derive to **one** member, `retained == payload length` in every row and `sequence_changed` is 0 — so the merged spelling only re-supplies bases the split had kept and nothing is inserted. `DNA/delins.md:47` therefore has no gap to reach, on `delins-recommendation-reach-when-the-input-arrives-split`'s ruled reading (landing on `main` as **#1760**).
2. **So `DNA/delins.md:17` and `general.md:34` govern unqualified**, and `general.md:35`'s codon exception cannot reach a separation of two by its own stated antecedent.
3. **#1616 merges them anyway. That is the deviation.**
4. **#1616 does not license its own merges**, and said so before they were measured — `derivation-may-not-be-bounded-by-the-inputs-spelling` states in terms that "the ruling removes an input-relative comparand, it does **not** license the merges the comparand happened to be blocking. Those are governed by `general.md:34` and by `delins-merge-vs-individual-gap-two-or-more`."
5. **`separation-rule-force-modal-or-negation` grades `general.md:34` rule 2**, whose remedy in its own words is "a deviation to **disclose and pin with a tripwire**, not a rule-7 bug, so it does not by itself block a release".
6. **Therefore the re-pin to 8 IS the tripwire.** A rise above 8 must name the clause that carried it, in the same terms the neighbouring ratchets demand.

| row | span | deleted | retained | derived members |
|---|---:|---:|---:|---:|
| `s01-c1-m3-all-del-p1-sep2` | 7 | 3 | 4 | 1 |
| `s01-c1-m3-all-del-p2-sep2` | 10 | 6 | 4 | 1 |
| `s01-c1-m3-all-del-p2-sep8` | 22 | 6 | 16 | 1 |
| `s01-c1-m4-all-del-p2-sep2` | 14 | 8 | 6 | 1 |
| `s01-c3m-m3-all-del-p4-sep3` | 18 | 12 | 6 | 1 |
| `s01-c3m-m4-all-del-p4-sep3` | 25 | 16 | 9 | 1 |
| `s01-c3p-m3-all-del-p4-sep3` | 18 | 12 | 6 | 1 |
| `s01-c3p-m4-all-del-p4-sep3` | 25 | 16 | 9 | 1 |

**Two** rows carry more retained than deleted — `p2-sep8` at 16 > 6 and `p1-sep2` at 4 > 3 — not one, as an earlier note claimed.

**The record also carries what was REFUTED**, because it will otherwise be re-derived: on `separation-is-a-property-of-the-spelling-not-of-the-variant` the separation is read off the **derived** partition, which is one member, so `general.md:34` arguably has no antecedent here at all and the #1710 instrument would be counting against the **authored** spelling. That reading was put to the operator and **declined** in favour of the conservative framing, on the ground that it makes `:34` unfalsifiable on the very population it is about — every merge derives to one member, so every merge would extinguish its own antecedent. And the **falsifier** is recorded: if a future change gives any of the eight a gap-bearing member, `:47` reaches that row and it leaves the deviation set, so 8 is a set to re-derive rather than a constant to defend.

The schema roles are split because `tests/it/common/rulings.rs` refuses one clause being both the authority and the departure: `DNA/delins.md:81` ("should **preferably** be described individually") governs, `general.md:34` and `DNA/delins.md:17` are deviated from. `:34` reaches these rows, and that is the ruling.

Two records gain measurement text and no ruling text: `derivation-may-not-be-bounded-by-the-inputs-spelling` (the 18 → 2 → 0 sequence, the distinction from the deleted bound, and the confluence cost) and `delins-merge-vs-individual-gap-two-or-more` (the branch's SCOPE ADDENDUM, recording that the ruling *does* move the spec's own published split).

## Still open, and NOT decided here

- ~~Whether the eight coding-axis merges are licensed.~~ **Ruled 2026-08-13** — they are a disclosed rule-2 `general.md:34` deviation; see **Ledger**.
- ~~Whether the `n.`-axis idempotency residue growing 21 -> 51 is acceptable.~~ **Ruled 2026-08-13** — disclosed as **29 projector + 1 defect**; the 29 are #1712's own thesis (projector staleness exposed, not caused) and the 1 is filed as **#1829**. See **Tests**.
- **The stranded identity member, #1829.** The one genuinely new defect in the idempotency residue. `NG_012337.1(NM_012459.2):c.[10del;23_25del;36_37insAAT]` re-normalizes to `n.[40del;53A>T;56T>C;58_60del;63_64=;67_68insATA]` — note the `63_64=`. It is disclosed as a bug with an issue, not as an acceptable output, and it is not fixed here.
- **`MAX_SPLIT_BLOCK`.** A cost question. Raising it would recover the 18/14 classes above.
- **W58** (`LRG_199t1:c.[992_1002del;1004T>C]` → `c.992_1004delinsAC`) is listed in `spec_worked_example_rules::FERRO_IS_WRONG`, tripwired by `no_accepted_deviation_has_quietly_started_conforming` so it cannot outlive its fix. It is on the **coding** axis, where `general.md:35`'s codon exception and `:47`'s carve-out are both live, so it is outside the frameless scope closed here and is not something a test edit may settle.
- **W49** — the spec's published four-member split re-deriving to `:47`'s spanning form — is recorded in the ledger with its operator ruling of 2026-08-11, also coding.

## Closes

Closes #1419. Closes #1420. Closes #1421. Closes #1603.

Representation-Change: real, in two directions, and re-measured on `origin/main` at `f17a0895` with `FERRO_MANIFEST` set and both spec artifacts regenerated with `--bin`. **SPLIT, off the coding DNA axis.** Scoping `DNA/delins.md:44-47`'s payload-coincidence carve-out to `c.` inside the shipping partitioner, and refusing to emit a block `MAX_SPLIT_BLOCK` never examined, move every frameless axis (`g.`, `m.`, `n.`, `r.`) from a spanning `delins` onto the individual descriptions `general.md:34` asks for. On the Mutalyzer corpus that is 28 (input, axis) pairs across 16 inputs of the `NG_008939.1` family, registered as an accepted divergence with Mutalyzer's own recorded answers retained as evidence, never overwritten. **MERGE / re-derivation, everywhere.** Deleting the input-relative weight bound lets a multi-member spelling be re-derived from the resulting sequence instead of handed back, so `spec_conformance_axis` `converged` rises 9,402 -> 10,977 at 3' and 9,228 -> 10,775 at 5' over 11,882 classes / 58,552 outputs, and all nine `reported_confluence_pairs` pairs converge (3 -> 9 of 9, 0 still split, both directions measured separately). A consumer storing either form must re-normalize. The negative guard `guard_violations` is **0 of 210 guarded rows in both directions**, re-measured at this base, with `guard_evaluations` pinned at 210 so the zero cannot be a structural silence; the instrument is known to move, having read 18 after commit 1 and 2 after commit 2, and `RESIDUAL_GUARD_VIOLATIONS` is empty. The `general.md:34` gate gives back 18 (3') / 14 (5') of the convergence gain, because a block past the cap cannot be split and its two spellings can no longer reach one string — rule 2 outranking rule 3, disclosed rather than netted away. **Three #1421 pairs land on a form neither spelling authored** (`1421-n2` in both directions; all three under 5'), now pinned per row and asserted as a convergence; this PR does not rule on whether those forms are correct. **`TEMPLATE:g.[4_5insC;5_6dup;15del]` and its one-member spellings converge** on `g.[5_10delinsCTATAT;16A[3]]` in both directions, after checking `general.md:56` ranks nothing above it. On the `p.` axis, `NM_FSSTOP.1:c.[4_5delinsTA;9A>G]` moves from `p.[(Val2Ter);(Lys3=)]` to `p.(Val2Ter)` — a bracket naming a residue past the terminator the change introduces, which `protein/frameshift.md:22` forbids; the single-member spelling of the same sequence change moves with it, so the two now agree. **DISCLOSED, RULE 2, WITH A TRIPWIRE — the eight coding-axis merges.** #1710's `coding_axis_separation_two_or_more_merges` reads **0 -> 8** in both directions, of 997 evaluated, on the same eight ids in each. Operator ruling 2026-08-13, recorded as `rulings[coding-axis-merges-are-a-disclosed-general-34-deviation]`: those eight **are** a `general.md:34` deviation — stated plainly, not reasoned away — and because `rulings[separation-rule-force-modal-or-negation]` grades `:34` as README **rule 2**, the remedy is to disclose and pin with a tripwire rather than to block a release. **The re-pin from 0 to 8 IS that tripwire**, and a rise above 8 must name the clause that carried it, in the same terms `guard_violations` and `prohibition_violating_outputs` demand. Measured at this base: all eight derive to **one** member (so their derived separation is zero, and the `sepN` in their ids is the authored figure), all eight are `all-del` designs, `sequence_changed` is 0, and in every one `retained == payload length` exactly (4, 4, 6, 6, 6, 9, 9, 16) — so nothing is inserted anywhere, `DNA/delins.md:47` has no gap to reach on `rulings[delins-recommendation-reach-when-the-input-arrives-split]`'s ruled reading (#1760), and `DNA/delins.md:17` / `general.md:34` govern unqualified. **Two** rows carry more retained than deleted, not one: `s01-c1-m3-all-del-p1-sep2` (4 against 3) as well as `s01-c1-m3-all-del-p2-sep8` (16 against 6). `rulings[derivation-may-not-be-bounded-by-the-inputs-spelling]` — this PR's own ruling — states in terms that it does **not** license the merges its deleted comparand was blocking, so nothing here is self-authorising. The record also carries the reading that was put to the operator and **declined** (separation read off the derived partition, which is one member, so `:34` would have no antecedent) and the **falsifier**: a row that gains a gap-bearing member leaves the deviation set, so 8 is a set to re-derive rather than a constant to defend. **DISCLOSED AS 29 PROJECTOR + 1 DEFECT — the `n.`-axis idempotency residue.** `NONCODING_NON_IDEMPOTENT` moves **21 -> 51** of 681 axes, and 51 is never to be quoted bare. **29** are the projector emitting a spanning `n.192_197delinsGTCCTGTGCTCATTATCTGGC` on the `NG_008939.1` family that re-normalization correctly splits — #1712's own thesis, pre-existing projector staleness this PR **exposes rather than causes**. **The 30th is a genuine defect and is not projector staleness**: `NG_012337.1(NM_012459.2):c.[10del;23_25del;36_37insAAT]` re-normalizes to `n.[40del;53A>T;56T>C;58_60del;63_64=;67_68insATA]`, a **stranded identity member** (`63_64=`) of the #1655/#1281 shape, filed separately as **#1829** and disclosed as a bug rather than as an acceptable output. Ten tests remain red locally and none is re-blessed here: **two** are one Mutalyzer-corpus divergence on a single `NG_012337.1` input, **one** is `multi_member_cis_axis`'s convergence census reading 401 -> 407 (an improvement it will not absorb without the clause named), and **seven** are `reported_partition_verdicts` rows whose closure needs a named clause. Of those ten only the first three are manifest-gated; the seven **are** visible to gating CI, and CI is additionally red on three `cis_adjudication_enumeration` tests that arrived on `main` with #1814 after this branch's base and so cannot be seen from it. The `NG_008939.1` family is registered rather than red.

## Summary by CodeRabbit

- **New Features**
  - Added coding-DNA-aware normalization, allowing coding-DNA variants to retain spanning descriptions while other axes may split at unchanged bases.
  - Added a `--coding-dna` option to partition-dumping tools.
  - Improved combined protein consequence handling, including moved terminators and stop-related outcomes.

- **Bug Fixes**
  - Improved convergence and canonical re-derivation across complex variant combinations.
  - Preserved individual descriptions for oversized genomic variant blocks.

- **Documentation**
  - Updated normalization rulings, conformance records, examples, and reported output expectations.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
Two measurements bearing on this PR's blocker, and a reading that says the blocker needs no adjudication to resolve.

## 1. Rebased onto current `main`, `guard_violations` is 18, not 27

Measured in a detached scratch worktree at base `8f68b425`, `FERRO_MANIFEST` set, both spec artifacts regenerated with `--bin`, dump-line count equal to the printed census on every instrumented run.

| tree | base | 3' | 5' | 3' converged | 5' converged |
|---|---|---:|---:|---:|---:|
| `main` alone (negative control) | `8f68b425` | **0** | **0** | 9,402 | 9,228 |
| this PR as pushed (positive control) | `b9f6cecb` | **27** | **27** | 10,993 | 10,791 |
| **this PR rebased** | `8f68b425` | **18** | **18** | **10,995** | **10,791** |

Both controls are exact — the old-base run reproduces this PR's own table verbatim with byte-identical row ids, and the `main`-only control passes with an empty dump and confluence equal to main's pinned censuses.

The 18 are a **strict subset** of the 27. All 9 that vanished are `del-del`, consistent with #1649 (`2dae10ee`, "let the splitter express two deletions"). The trade-off ratio barely moves: it was 27 against +1,854 converged, it is now 18 against +1,593.

**Note for whoever rebases:** the local branch base predates #1649, so measuring from there still shows roughly 27 — the collapse only appears once #1649 is in the base.

The rebase conflicts in 11 paths. Only one is in `src/` — a single hunk at the `coalesce_payload_alignment_split` call site, where `main`'s `payload_coalesce_applies` gate replaces text identical to `b9f6cecb`'s except for a comment tense change. It is behaviourally inert for this census (`payload_coalesce_applies` requires `CanonicalCoalesced && Cds`; the axis runs under `live`). The other 10 are the ruling ledger and nine expectation files, which need a judgement call. Taking `main`'s `tests/` wholesale was verified not to affect the count: `spec_conformance_axis.rs` is not touched by this PR at all, and its corpus is built programmatically by `src/conformance/spec_corpus.rs` with no fixture read. So **18 is a property of this PR's `src/` delta on current main** — if resolving those 10 changes `src/` behaviour, re-measure.

## 2. The blocker does not need an adjudication — two decided records already reach it

This was queued as an open question to be ruled on. It isn't one.

`consultation/SVD-WG010.md:30-33` states current law inside the rejection notice:

> `:32` — "two variants separated by one or more nucleotides should be described individually and **not** as a 'delins'."
> `:33` — "**exception**: two variants separated by one nucleotide, **together affecting one amino acid**, should be described as a 'delins'."

`projection-codon-exception-is-decided-by-the-rendered-axis` (decided, governing `delins.md:42`) rules on the exception's reach: *"`:42` is a conditional with two conjuncts … The first is statable on any DNA reference; the second is not statable on a genomic one, so the clause's antecedent is unsatisfied rather than merely weaker. On a genomic axis they do not fire, and `DNA/delins.md:17` governs unopposed: the members are described individually."*

`delins-payload-coincidence-carve-out-is-coding-dna-scoped` (decided, governing `delins.md:47`) scopes the payload-coincidence carve-out to `c.` and nothing else.

All 18 rows are separation **exactly 1** on a **frameless** axis — Genomic 16, NonCodingMultiExon 2, zero coding. So the codon exception's antecedent is unsatisfied, `:47` does not reach, and `delins-adjacent-members-when-both-consume-reference` is separation **zero** rather than one. Nothing licenses the merge; `general.md:34` / `delins.md:17` govern unopposed.

That makes the 18 a **rule 1** violation under records already made, which `README.md`'s rule 7 classifies as a bug rather than a disclosure — so the +1,593 confluence gain cannot pay for it, exactly as this PR's own description argues.

**Practical consequence: this is unblocked in the sense that matters** — no ruling is owed, and the path forward is to close the 18 merges rather than to wait. Worth knowing that the same question was going to be asked of the canonical partition flip, whose own 6 rows share the identical shape; a base-matched comparison shows neither set contains the other (19 shared, 35 flip-only, 8 exclusive to this PR at a common base), so a gate written against one mechanism does not close the other.

The PR's other blockers are untouched by any of this and remain separate questions: W58 (`LRG_199t1:c.[992_1002del;1004T>C]`, on the **coding** axis where the exception is live), W49 (gaps of 4, 5 and 3 — a `delins-merge-vs-individual-gap-two-or-more` scope question), and the #1620 form-A row.

nh13:
## Rebased onto `2f4e3bb9`, review pass applied, and the blocker re-classified per the 2026-08-12 comment

**The direction question is settled and this branch now says so.** `rulings[derivation-may-not-be-bounded-by-the-inputs-spelling]` gains a `THE FRAMELESS RESIDUAL IS A BUG` section recording that no adjudication is owed — `projection-codon-exception-is-decided-by-the-rendered-axis` and `delins-payload-coincidence-carve-out-is-coding-dna-scoped` already reach the frameless case, `delins-adjacent-members-when-both-consume-reference` is separation zero rather than one, so `general.md:34` / `delins.md:17` govern unopposed and the residual merges are a README rule-1 violation to close rather than a representation change to declare.

**The 18 reproduce exactly.** Measured here on the current base: `guard_violations` 18 in both directions, `converged` 10,995 (3') and 10,791 (5') — byte-identical to the figures in the comment. Nothing is re-blessed to make it green.

### Ledger integrity

The remote head this replaces (`092551e4`) carried `absolute-prohibition-enforcement-stage` as `undecided`, reverting #1634's decided ruling. The pushed head restores it. Checked programmatically across all three sides:

| against | records dropped | statuses regressed |
|---|---|---|
| old local tip `40d44e35` | none | none |
| old remote tip `092551e4` | none | none — `absolute-prohibition-enforcement-stage` moves `undecided` → **`decided`** |
| `origin/main` `2f4e3bb9` | none | none |

22 records, one added by this branch. The rebase restored the five that had landed on `main` after both prior bases.

### Census figures corrected

`3,245` / `3,251` were withdrawn by this record's own `BASE CORRECTION` and were still live in **shipped source** — `src/error_handling/types.rs` and a `src/normalize/merge.rs` comment. Both are corrected, along with every `8,02x` base, to the post-#1649 measurement: **8,361 → 11,271 (+2,910) at 3'** and **8,367 → 11,272 (+2,905) at 5'**. The `n.`/`r.` chronologies in `cis_confluence_nr_axis.rs` are rewritten end to end (#1484, then #1649, then this change) rather than left as two overlapping paragraphs.

### Review items

| item | disposition |
|---|---|
| `Representation-Change:` trailer missing (×2) | **rejected — false positive.** The trailer is in the description and `Representation change declared` is green. Its figures are updated (27 → 18) and its "blocked pending an operator ruling" clause withdrawn. |
| duplicated `normalized_preserving` helpers | **fixed** — one copy in `tests/it/common/cis_apply_oracle.rs`, parameterised by the core, with the assertion-order contract on it. |
| present-tense descriptions of the deleted bound | **already correct** — both sites were past-tense and cross-referenced the record. |
| missing `denotes` check | **fixed** — `denotes(authored) == denotes(expected)` asserted before the string, so a future re-bless cannot carry a changed sequence. |
| wrong `delins.md:83` verdict description | **fixed** — all three claims are `decided`, so it is a same-verdict `[MULTI]` line; the comment said the opposite. |
| `Row::five_prime` doc stale | **moot** — `reported_partition_verdicts.rs` is back to `main`'s model (see below). |
| rename `a_dup_flush_against_a_del_is_left_alone` | **already done**; the one surviving stale reference in the module doc is fixed. |
| "no threshold can distinguish this" | **fixed** — a delta threshold of 2 admits this row and refuses Example 1, so the claim is withdrawn and the test scoped to what it measures. |
| do not mark the removal decided before the frameless scope is adjudicated | **addressed by the second branch of its own ask** — the explicit operator ruling now exists and is recorded. The behaviour it asks for is not restored, because the ruling says to close the merges rather than to re-gate them; that work is outstanding and the guards stay red. |

### What the rebase undid

Three of this branch's re-blessings were reverted to `main`'s values because #1649's two-deletion alignment restores `main`'s answer: `project_cis_separated_inframe_deletions_stay_individual`, `test_allele_with_two_deletions_*`, and the whole `reported_partition_verdicts.rs` / `reported_confluence_pairs.rs` rework. Keeping `main`'s model there leaves its guards intact and red rather than re-blessed, which is what the file's own messages ask for.

### Tests

**18 red, all deliberate, none introduced beyond the disclosed class.** `origin/main` at the same base is 10,194 passed / 0 failed, so the set-difference is the full 18. They are one mechanism seen through several guards: `spec_conformance_axis` (2), `reported_partition_verdicts` (5) and `reported_confluence_pairs` (2) on the `1420-v2` row, `equivalence_cross_axis_rung` (3), `cis_junction_crossing_shift` (1), `spec_corpus_regressions` (1), `multi_member_cis_axis` (1), `mutalyzer_normalize_tests` (2), `projector` (1).

`multi_member_cis_axis::axis_normalized` reports convergence **improving** 401 → 407 on real rows and invites a re-bless; it is left red deliberately, because its own message warns that convergence onto the wrong form raises the same figure, and six unaudited rows is exactly that question.

`cis_junction_crossing_shift::the_three_member_spelling_and_its_one_member_form_are_two_fixed_points` carries a new `RED ON THIS BRANCH, DELIBERATELY` note: its failure message offers a re-bless, and taking it would be wrong — the output moved onto a *third* form, not onto the one `contiguous-insertion-split-by-a-blocked-derivation` decides for.

`cargo fmt`, `cargo clippy --all-features --all-targets -- -D warnings` clean. Both spec artifacts regenerated. Not merged.

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
@coderabbitai full review

nh13:
@coderabbitai review

nh13:
Closing this as **superseded**. Its content landed in two pieces, and what is left cannot be rebased forward — it is measurably behind `main` rather than ahead of it.

## Where the content went

| piece | shipped as |
|---|---|
| `derivation-may-not-be-bounded-by-the-inputs-spelling` + the CLAUDE.md row + the regenerated clause index + the `RULING_STATUSES` pin | **#1733** (merged) |
| the input-relative weight-bound deletion | **#1899** (merged) |
| the coding-DNA scoping of the payload-coincidence carve-out (`CoincidenceCarveOut` threaded into `partition_block`) | **#1899** |
| the combined-product fix when a cis pair moves the terminator | **#1899** — `a_cis_pair_whose_combined_product_stops_early_reports_the_moved_terminator` is on `main` at `src/project/protein/indel.rs:3431` |
| the four reserved 1420/1421 adjudications | **#1899** |

One commit was **deliberately not carried**: `fix(normalize): refuse to assert an unexamined block as one member off c.`. #1899 deletes that axis gate rather than shipping it, because with `MAX_SPLIT_BLOCK` derived from `MAX_CANONICAL_WINDOW` the gate **cannot fire** — the window test refuses the same blocks one step earlier, on every axis. Verified with a 4,200-base block whose `c.` positive control failed, which is what distinguishes "unreachable" from "merely unused". The tests that pinned it go with it.

## Why the remainder is not rebasable

- Conflict surface against current `main`: **99 files**. Its base is **43 commits** behind, spanning #1835's canonical-coalesced default flip and #1899.
- Ledger records unique to this branch: **zero**. `main` carries 31, this branch 28, all subsumed — and this branch still has `delins-recommendation-reach-when-the-input-arrives-split` as `undecided` where `main` has it `decided`. On adjudication it is behind, not ahead.

Its pins were authored before the default arm flipped, so replaying them would re-assert pre-flip expectations against a baseline that has moved. That is a re-derivation job, not a merge.

## What is worth harvesting, and is not lost

Roughly two dozen test functions across ~15 `tests/it/` files exist here and not on `main`. The branch is not deleted, so they remain available at `8786b96f`. The ones worth re-deriving first:

- `no_accepted_deviation_has_quietly_started_conforming` — a ratchet guard whose value does not depend on which direction the flip moved things
- `every_binary_target_reports_a_refused_partition_switch` — instrumentation wiring guard
- the confluence/convergence family in `cis_junction_crossing_shift.rs`, `issue_1304_junction_barrier_snapshot.rs`, `issue_1320_dup_spans_sibling_junction.rs` and `reported_confluence_pairs.rs`

Each needs its expectation re-derived against current behaviour rather than transplanted — `the_spanning_delins_and_its_published_split_converge_on_the_recommended_form`, for instance, asserts convergence on the spanning form, and the flip moved the corpus predominantly the other way (7,760 spanning-to-members against 167 the reverse, measured over 9,949,738 rows in #1900).

## One defect from here that is now fixed

This branch's first commit dropped the ` ```text ` fences around the generated block in `tests/it/clause_ruling_index.rs`. It was silent by construction — the extractor skips fence lines, so `the_rendered_index_is_current` passes either way — while breaking the rustdoc rendering. It was recorded as unfixed during the #1733 split and has been restored in #1899.

Issue #1617, opened 2026-08-11
Normalizer: converge the spelling-relative separation class on the re-derived form in the live partitioner


nh13:
The ruling record `separation-is-a-property-of-the-spelling-not-of-the-variant` (in `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json`) is now **`decided`**, for the re-derived form. The live partitioner does not produce it yet. This issue is the implementation half.

## The case

`NC_000001.11:g.1001002-1001016` reads `ATGAGGGGCCACTGT` — a `GGGG` run at `1001006-1001009`, a `CC` run at `1001010-1001011`, and a lone `C` at `1001013`. Two spellings of **one** variant, both denoting `ATGAGGGCATGT`:

```
NC_000001.11:g.[1001009del;1001010del;1001013del]
NC_000001.11:g.[1001009del;1001011del;1001013del]
```

The decided form is

```
NC_000001.11:g.[1001009_1001010del;1001013del]
```

## What ferro does today

Measured on a synthetic provider reproducing those 15 bases, 3' direction, at `5397bb4f`:

| `FERRO_PARTITION` | `[1001009del;1001010del;1001013del]` | `[1001009del;1001011del;1001013del]` |
|---|---|---|
| `live` (ships) | `g.[1001009_1001010del;1001013del]` | `g.[1001009del;1001011del;1001013del]` |
| `shadow` | `g.[1001009_1001010del;1001013del]` | `g.[1001009_1001010del;1001013del]` |
| `canonical` | `g.[1001009_1001010del;1001013del]` | `g.[1001009_1001010del;1001013del]` |
| `canonical-coalesced` | `g.1001009_1001013delinsCA` | `g.1001009_1001013delinsCA` |

So `live` is non-confluent on this class, and **both sequence-first arms already produce the decided form from both spellings**. The work is to reach it on the shipping path, not to invent it.

## Mechanism

Two known partitioner behaviours produce the divergence, and both are already characterized:

* **Members are normalized independently rather than re-derived from the resulting sequence.** This is the defect #1419 and #1420 name in their own titles. Placing each member canonically while keeping the handed-in member boundaries is exactly what makes the second spelling a fixed point.
[...]

PR #1620, opened 2026-08-11
docs(rulings): decide the separation record for the re-derived form

nh13:
Flips the ruling record `separation-is-a-property-of-the-spelling-not-of-the-variant` from `undecided` to `decided`, and adds the tests that pin it.

## The ruling

The separation `general.md:34` keys on is read off the partition **re-derived from the resulting sequence**, never off the input's spelling. That is rule 3 of the README ruleset (shipped in #1605): *"Every rule is evaluated over the resulting sequence, never over the input's spelling."*

Stated on real coordinates. `NC_000001.11:g.1001002-1001016` reads `ATGAGGGGCCACTGT` — a `GGGG` run at `1001006-1001009`, a `CC` run at `1001010-1001011`, a lone `C` at `1001013`. Two spellings of one variant, both denoting `ATGAGGGCATGT`:

```
NC_000001.11:g.[1001009del;1001010del;1001013del]
NC_000001.11:g.[1001009del;1001011del;1001013del]
```

Three conformant descriptions, and the ruling is **A**:

| | form | verdict |
|---|---|---|
| A | `g.[1001009_1001010del;1001013del]` | **decided** |
| B | `g.[1001009del;1001011del;1001013del]` | the independently-normalized form — the defect #1419 and #1420 name in their own titles |
| C | `g.1001009_1001013delinsCA` | outside the scope `delins-merge-vs-individual-gap-two-or-more` set for itself |

## Measured, not argued

Probed on a synthetic provider carrying those fifteen bases, 3' direction, at `5397bb4f`:

| `FERRO_PARTITION` | first spelling | second spelling |
|---|---|---|
| `live` (ships) | A | **B** |
| `shadow` | A | A |
| `canonical` | A | A |
| `canonical-coalesced` | C | C |

Both sequence-first arms already produce the decided form from both spellings; `live` does not, and that gap is **#1617**. Rebuilding with `seqfirst::MIN_SEPARATION` at 1 instead of 2 still yields A from both spellings under both arms, so A is a property of the re-derivation and not of the separation constant.

## What the rationale deliberately does NOT argue

**3'-maximality does not decide this**, and an earlier draft of the reasoning claimed it did. All three forms are 3'-maximal as written: A's `1001009_1001010del` deletes `GC`, and shifting it 3' by one would need `ref[1001009] == ref[1001011]`, i.e. `G == C`, which fails. The record says so explicitly so no later reader reintroduces the error.

`delins.md:47` is not re-argued either — the decided `delins-merge-vs-individual-gap-two-or-more` record already scopes itself to a minimal single `delins` split on payload coincidence and says it is not a general licence to merge at gap ≥ 2, with rows on a frameless axis remaining violations. This row is `g.`.

Mutalyzer and VariantValidator both produce C; biocommons `hgvs` cannot represent the row at all. That is recorded as *not* an argument for C, citing the filer's own "shortcuts that merge deletion-insertions" (#1430) and "these were over-merged by mutalyzer" (#182).

## Tests

* `the_separation_two_members_present_is_not_a_property_of_the_variant_on_real_coordinates` — the genomic case on a synthetic provider reproducing those fifteen bases at those offsets, so it is hermetic and needs no `FERRO_MANIFEST`. It pins **today's `live` output**, with a comment stating plainly that this is pre-fix behaviour against a now-decided target. The second spelling's expectation is form B, pinned as an observation and not endorsed.
* `the_decided_target_is_the_re_derived_form` — `#[ignore]`d, asserts both spellings reaching form A. It is #1617's acceptance criterion. Verified to fail today for exactly the right reason (the second spelling, B vs A).

## Bookkeeping

`RULING_STATUSES` moves with the ledger, and `clause_ruling_index.rs`'s generated index is regenerated: the record gains two clause citations (`general.md:41` for the 3' rule it does *not* decide with, `DNA/delins.md:47` for the merge it declines), and its status flips everywhere it is indexed. The census moves to 11 records / 8 decided / 3 undecided, 48 clause lines, 8 multiply-named, 2 mixed-verdict. `delins.md:17` stopped being a mixed-verdict line as a result, so the mixed-verdict assertion moved to `delins.md:18`; the module docs now warn that a falling mixed count means records got decided, not that a line became safe to read from one record.

All six clause quotes in the record were verified byte-exact as raw substrings of the cited lines in the pinned submodule (`6f85311`), not merely whitespace-collapsed matches. The generator (`generate_spec_fixture`) validates them too and builds clean.

## Verification

* Full suite: **9,850 passed, 0 failed, 151 skipped** (`cargo nextest run --features dev`).
* `cargo fmt --check` clean; `cargo clippy --all-features --all-targets -- -D warnings` clean.
* `examples/dump_normalized_corpus.rs`, `origin/main` against this branch: **0 of 78,298 rows move**. That is not a corpus-blindness zero — no file under `src/` is touched, so the normalizer is byte-identical by construction and the dump is a confirmation rather than the evidence.

Representation-Change: none. Tests and the ruling ledger only; no watched file is touched. A corpus dump against `origin/main` moves nothing — 0 changed of 78,298 corpus rows compared.


nh13:
@coderabbitai pause

Issue #1619, opened 2026-08-11
hgvs_to_spdi resolves a c. position through exon gaps while the sequence it indexes is flat


nh13:
## What's wrong

For a `c.` variant on a transcript whose exon list has gaps, `hgvs_to_spdi` resolves the position by walking the exons, while `Normalizer` (and `ReferenceProvider::get_sequence`) index the flat transcript sequence. The two therefore name different bases of the same accession, and the gap grows by one per intron crossed.

Reproduced on the committed fixture `tests/fixtures/sequences/normalization_transcripts.json`, whose `NM_000492.4` record carries the real 6,070-base CFTR transcript, `cds_start: 71`, and 27 exons spelled with 1-base gaps:

```json
"exons": [{"number":1,"start":1,"end":122},
          {"number":2,"start":124,"end":233},
          {"number":3,"start":235,"end":342}, ...]
```

`c.1520` is at flat 0-based offset **1589** (`cds_start - 1 + 1520 - 1`). `hgvs_to_spdi` places it at **1599** — exactly the ten gaps its walk crosses on the way.

The consequence, on a case the suite already pins:

```
NM_000492.4:c.1520_1522del  ->  NM_000492.4:c.1521_1523del
  (normalize_tests::normalization_transformations::test_deletion_shifts_in_real_homopolymer)
```

That normalization is **correct**. On the flat sequence, `c.1515..1530` reads `TATCATCTTTGGTGTT`, so `c.1520_1522` is `TCT` and `c.1521_1523` is `CTT`, and deleting either leaves `TATCATTGGT` — the documented CFTR ΔF508 3'-shift.

Read through `hgvs_to_spdi`, though, the pair lands on `[1599, 1603) = TTCC`, where deleting `TTC` leaves `C` and deleting `TCC` leaves `T`. A correct normalization reads as a corruption.

## Which side is wrong

`get_sequence` on a transcript accession serves the flat transcript, so an offset that has walked exon gaps cannot be used to index it. Either the resolution should not consult exons when the target sequence is the transcript itself, or the exon coordinates in that fixture are not the frame the resolver assumes them to be in. Worth settling both ways round, because a fixture that is internally inconsistent is its own problem.

Note the fixture's exon spans total 6,043 against a 6,070-base sequence, which is a second symptom of the same inconsistency.

## How it was found

[...]

Issue #1618, opened 2026-08-11
Anchored repeat: hgvs_to_spdi and the normalizer disagree on what the copy count is counted against


nh13:
## What's wrong

`hgvs_to_spdi` and `Normalizer` disagree about how many bases an anchored repeat denotes, so normalizing one changes the sequence the other reads.

Reference: `SyntheticBuilder::genomic("AAGTGTTA")` — core base 1 at `g.257`, so `g.259_262 = GTGT`, `g.258 = A`, `g.263 = T`.

```
in   NC_TEST.1:g.262TG[6]
out  NC_TEST.1:g.259_262GT[6]

over NC_TEST.1 [258, 262):
  reference       GTGT
  input applies   GTGTGTGTGTGTGT     (14)
  output applies  GTGTGTGTGTGT       (12)
```

Two bases are lost. `hgvs_to_spdi` reads the anchored spelling as **6 copies replacing the 1-copy `TG` tract at `261_262`** — net `+10` over a 2-base span. The normalizer's output is **6 copies replacing the 2-copy `GT` tract at `259_262`** — net `+8` over a 4-base span.

## Which side is right is the actual question

Both readings are defensible for an input HGVS leaves ambiguous, and the two pieces of ferro have simply picked differently:

- `normalize_repeat`'s tract maximization (the fix behind `tests/it/repeat_tract_maximization.rs`) deliberately widens the window from the literal unit's 1-copy match to the maximal 2-copy tract, and keeps the caller's count. Widening the window while holding the count fixed is what drops the two bases.
- `spdi::convert::resolve_repeat_tract_span` resolves the same anchor to the literal unit's own tract.

So this is not "one of them has a bug" on its face — it is an unmade decision about what a single-anchor repeat count is counted *against*, which has been made twice, differently.

`tests/it/repeat_tract_maximization.rs` pins the current output as correct and asserts idempotency and anchored/range agreement, all of which hold. What it does not assert is that the output denotes the input's bases, and it does not.

## How it was found

By the denoted-sequence seam oracle (#1615), `FERRO_ASSERT_SEQUENCE=1`, over the full test suite. It is one of only two rows that fire, and it is the reason the flag is not yet set in CI's `test-oracle` job.

The same shape appears on the RNA axis in that file's own module docs — `NM_004006.3:r.-124ug[14]` -> `r.-127_-124gu[14]`, the real spec input the maximization was written for.

## Severity

[...]

Issue #1600, opened 2026-08-10
Normalization denotes different bases when a reducing delins lands in a tract spanning a sibling


nh13:
Normalizing a spanning `delins` emits a description that denotes **different bases** than its input. Live on `main` today (measured at `112ef784`); #1598 does not fix it.

## Reproducer

Reference: `SyntheticBuilder::genomic("GAACAGCAGAAGCGA")` — the core padded with `"ACGT"` x 64, so core base 1 is `g.257` and `ref 257..272 = GAACAGCAGAAGCGAA`.

```
NC_TEST.1:g.261_267delinsGCAGAAAC   ->   NC_TEST.1:g.[261del;266_267insCA]     (3')
NC_TEST.1:g.261_267delinsGCAGAAAC   ->   NC_TEST.1:g.[261del;265_266insAC]     (5')

  intended  ... G A A C   G C A G A A   A C   G C G A ...
  emitted   ... G A A C   G C A G A     C A   A G C G A ...    <- different bases
```

The same wrong output comes from every wider spelling of the block (`g.260_268delinsCGCAGAAACG`, `g.261_268delinsGCAGAAACG`). The bracketed spelling of the same haplotype, `g.[261del;267_268insAC]`, is a fixed point and is **correct** — so this is also a non-confluence, and the two spellings disagree about the sequence rather than only about the string.

The output is well-formed, its members re-parse, it is in bounds, and it is a fixed point, so all three seam oracles (`FERRO_ASSERT_IDEMPOTENT`, `FERRO_ASSERT_REPARSE`, `FERRO_ASSERT_IN_BOUNDS`) pass on it. None of them compares the denoted sequence, which is the only property it violates.

## Mechanism

`canonicalize_from_sequence` is not at fault — its round-trip guard holds throughout. The corruption is in the per-member pipeline the canonicalization alternates with, and it is #1592's family one step further along.

Tracing pass by pass:

1. The sequence-first pass derives `g.[261del;263_264insAG;265G>A;267A>C]` — correct.
2. `merge_consecutive_edits` then merges `264_265insGA` with `265G>A` into `g.265delinsGAA` — correct, and it is a **`delins` that reduces**, which is #1592's signature.
[...]

Issue #1592, opened 2026-08-10
Normalizer: a member that reduces to an insertion escapes both sibling clamps and changes the sequence


nh13:
A cis member that arrives **claiming bases** and normalizes into a **junction-inserting** form (`ins`/`dup`) is clamped by neither sibling pass, so its junction can sweep arbitrarily far past a sibling's bases. The emitted description is well-formed, disjoint, warning-free — and denotes different bases from the input.

## Reproduction

Against the padded synthetic contig `NC_TEST.1` built from core `TCCAGCAGATAT` (core base 1 at `g.257`), 3' direction, on `main` at `9b8e55fc`:

```
NC_TEST.1:g.[262_263insAG;264G>T;266T>A]   ->  NC_TEST.1:g.265A[3]
  input  applies to  ...CAGATAAAT...
  output applies to  ...CAGAAATAT...

NC_TEST.1:g.258_266delinsCAGCAGATAA        ->  NC_TEST.1:g.[259del;265A[3]]
  input  applies to  ...TCAGCAGATAAAT...
  output applies to  ...TCAGCAGAAATAT...
```

Both are reachable through `normalize()` today with no environment flags. The correct repeat member is `267A[3]` — the lone `A` at 267 — not `265A[3]`, which names the other lone `A` two bases 5'.

## Mechanism

Traced through `Normalizer::normalize_allele`:

```
pass=0  [262_263insAG;264G>T;266T>A]
        per-member                        -> [264_265dup;264G>T;266T>A]
        clamp_sibling_crossing_junctions  -> [263_264insGA;264G>T;266T>A]   (fires: before form is an ins)
pass=1  merge_consecutive_edits           -> [264delinsGAT;266T>A]
        canonical_split + normalize_core  -> [268_269dup;266T>A]            <-- junction 264 -> 269,
                                                                                over the 266T>A sibling
        clamp_sibling_crossing_shifts     :  declines
        clamp_sibling_crossing_junctions  :  declines
```

The two clamps have complementary jobs and this shape falls between them.

* `clamp_sibling_crossing_shifts` refuses a member that **grew** — deliberately, per #1266/#1279: "an insertion canonicalising to a duplication moves its span and its junction together, so bounding the span mis-places the copy". `264delinsGAT` (1 base) becoming `268_269dup` (2 bases) is exactly that, so it defers to the junction clamp.
[...]

PR #1622, opened 2026-08-11
feat(normalize): a fourth seam oracle that compares the denoted sequence

### LINKED ISSUE #1615: Normalizer: add a fourth seam oracle that compares the denoted sequence
## The gap

ferro has three debug-assertion seam oracles, all installed at `normalize_core_checked`'s single exit and all env-gated:

| oracle | asks |
|---|---|
| `FERRO_ASSERT_IDEMPOTENT` | is the output a fixed point? |
| `FERRO_ASSERT_REPARSE` | does the output parse? |
| `FERRO_ASSERT_IN_BOUNDS` | does the output name a coordinate its own sequence has? |

**None of them compares the sequence the description denotes.** A normalization that emits a well-formed, disjoint, warning-free, in-bounds, fixed-point description of *different bases* passes all three.

That is not hypothetical. #1592 and #1600 are both live on `main` today, both emit different bases than their input, and both issues record that all three oracles pass on the reproducer. Neither was found by an oracle — #1600 was found by a proptest run past its default case count, at roughly 1 in 3,750 draws.

## Why this blocks the partitioner campaign

The campaign is a sequence of changes that deliberately move normalized output. Without a denoted-sequence check at the seam, a representation change and a corruption look identical: both show up as "the string moved". The `Representation-Change:` discipline can only describe a move it can trust to be a move.

## What it would have caught

Eight defects, six of them already fixed one shape at a time:

| # | input | wrong output |
|---|---|---|
| #1254 | `g.[3_4del;9del]` | `g.12_14del` |
| #1281 | `g.[1del;9_10delinsA]` | `g.[1del;1del]` (output denotes no sequence at all) |
| #1290 | `g.[263_264insA;265_266insC]` | `g.[265_266insC;266dup]` |
| #1304 | `g.[260_261insGA;261_262insA;264del]` | `g.[261_262insA;261_262dup;265del]` |
| #1308 | `g.[263_264insTG;264_265insTG]` | `g.[264_265insTG;264_265dup]` |
| #1312 | `g.[260_261insAC;261_262insAC]` | `g.261_262insACCA` |
| #1592 | `g.[262_263insAG;264G>T;266T>A]` | `g.265A[3]` — live |
| #1600 | `g.261_267delinsGCAGAAAC` | `g.[261del;266_267insCA]` — live |

Six instances of one class, each filed, fixed and regression-tested separately, is the same argument #1353 made for `FERRO_ASSERT_IN_BOUNDS` after three.

## Proposed shape

A fourth oracle at the same seam, following the three existing ones exactly: env-gated, `#[cfg(debug_assertions)]`, cached gate, and named in the CI oracle job's env alongside the others.

Apply the **input** description to the reference to get sequence A, apply the **output** to get sequence B, assert A == B. The applier must not be the normalizer or the check is circular; `src/spdi/apply.rs` already holds a normalizer-independent SPDI-splicing applier (`variant_edit_triples` / `apply_triples`), which is the same algorithm `tests/it/common/cis_apply_oracle.rs` uses and the one every sibling-crossing test already rests on. `EquivalenceChecker` is **not** usable here — it is documented as circular at `tests/it/issue_1234_sibling_clamped_shift.rs:198`.

Two design points that need deciding rather than falling out:

1. **A side that cannot be applied must not silently pass.** A skip that reads as a pass is the exact failure mode this oracle exists to remove. In particular an *output* that denotes no sequence while the input denotes one — #1281's `g.[1del;1del]` — is strictly worse than a sequence change and must be loud, not skipped.
2. **Where it runs.** The other three are set together in the `Oracles` CI job and in the nightly; the precedent is to follow.

## Acceptance

- The oracle fires on all eight rows above, pinned as a test against the recorded input/output pairs so the oracle itself has a regression guard.
- The full suite is green with the variable on — no false positives on ordinary correct normalizations.
- It moves no normalized output (`Representation-Change: none`).

nh13:
Closes #1615.

## What this adds

A fourth seam oracle at `normalize_core_checked`'s exit, `FERRO_ASSERT_SEQUENCE=1`: apply the input description to the reference, apply the output, assert the bases agree.

The three already there are all *form* questions — is it a fixed point, does it parse, are its coordinates in range — and a description denoting entirely different bases satisfies every one of them at once.

The applier is `spdi::compare_denoted_sequences`, new in the first commit. It reaches the bases through `hgvs_to_spdi` and an SPDI splice with no normalization in the path, so it cannot agree with an output merely because normalization produced it. (`EquivalenceChecker` normalizes both sides and is documented as circular at `tests/it/issue_1234_sibling_clamped_shift.rs:198`.) Both descriptions are applied over the **union** of their spans, one fetch, which is what makes a 3'-shift compare as the identity it is rather than as a difference of frames.

## The catch table

Every defect #1615 names, against the wrong output its own issue records. `tests/it/issue_1615_denoted_sequence_oracle.rs` pins all eight.

| issue | verdict | input applies | output applies |
|---|---|---|---|
| #1254 | changed sequence | `ATATTAATA` | `ATATATATA` |
| #1281 | **output denotes nothing** | — | `g.[1del;1del]`, two members claiming base 1 |
| #1290 | changed sequence | `AAACA` | `AACAA` |
| #1304 | changed sequence | `GAGAAAA` | `GAAGAAA` |
| #1308 | changed sequence | `TGTTGG` | `TTGGTG` |
| #1312 | changed sequence | `ACAAC` | `AACCA` |
| #1592 | changed sequence | `AGATAA` | `AGAAAT` |
| #1600 | changed sequence | `GCAGAAAC` | `GCAGACAA` |

**8 of 8, no misses.** Each pair of applied strings reproduces its own issue's hand-written "intended vs emitted" line exactly, which is the cross-check that the oracle is reading the same defect the reporter saw.

The test deliberately does **not** re-normalize. A test that ran the normalizer would go green as each defect is fixed and stop saying anything about the oracle; pinning the recorded strings keeps it a guard on the oracle rather than on the six fixes.

## Skip vs fail — the design decision, and why it was not free

A skip that reads as a pass is the exact failure mode a sequence oracle exists to remove, so:

- **fires** when both sides apply and the bases differ;
- **fires** when the output is *self-contradictory* while the input denotes a sequence — #1281's shape, which is worse than a wrong sequence, not milder;
- **skips, counted**, otherwise. `normalize::denoted_sequence_oracle_counts()` returns `(compared, skipped)` process-wide, because zero comparisons and zero faults look identical from outside.

The load-bearing part is that the second bullet is narrower than "the output produced no triples". Reading any underivable output as a fault is the obvious implementation and it is wrong, because **the two sides of the comparison do not need the reference equally**: `g.1000delC` states its own deleted base and converts with no provider, while the `g.1000del` normalization emits for it must read one.

Measured over the full suite, first run to last:

| round | fires | class closed |
|---|---|---|
| 1 | 344 | — |
| 2 | 246 | 328 × untranslatable output → counted skip |
| 3 | 21 | 233 × insertion flush against a deletion (`triples_are_disjoint` calls it an overlap; `cis_apply_oracle`'s tie-break does not, and its verdict there is input-order dependent); 6 × `pter`/`qter`; 5 × corrected `REFSEQ_MISMATCH`; 1 × `r.` `U` against DNA `T` |
| 4–5 | **6** | ~12 × insertion *interior* to a deletion (a genuine overlap — the input denotes nothing, so there is no baseline); 1 × predicted `[(…)]` allele |

Every one of those is pinned as a *class* in `the_measured_false_positive_classes_stay_silent`, not patched as a case.

## Negative control

`FERRO_SWEEP_SEEDS=full cargo nextest run --features dev`, with the flag on: **6 fires in 2 rows**, and both are real disagreements inside ferro rather than noise — filed as **#1618** and **#1619**:

- **#1618** — `g.262TG[6]` → `g.259_262GT[6]`. `hgvs_to_spdi` reads the anchored spelling as 6 copies replacing a 1-copy tract, the normalizer's output as 6 copies replacing a 2-copy tract: 14 bases against 12. An unmade decision about what a single-anchor copy count is counted against, made twice, differently.
- **#1619** — `NM_000492.4:c.1520_1522del` → `c.1521_1523del`. The normalization is *correct* on the flat transcript (both spell `TATCATTGGT`); `hgvs_to_spdi` resolves the `c.` position by walking `normalization_transcripts.json`'s 27 exons with their 1-base synthetic introns and lands 10 bases 3'. A false positive **for this oracle**, and a real applier bug.

With the flag off, the suite is **9,853/9,853 green**.

## Where it runs, and the one place it deliberately does not

`sweeps` (gating — the cis-allele space all eight defects came from, and measured green there over the full corpus) and the nightly (non-gating, per that job's existing contract).

`sweeps` selects `$SWEEP_FILTER + test(issue_1615_denoted_sequence_oracle)`. The appended term is not cosmetic: `sweeps` is the only job that sets `FERRO_ASSERT_SEQUENCE`, and it selected on `SWEEP_FILTER` alone, which names three sweep modules and not this PR's own test module — so `the_seam_compares_when_the_flag_is_set`, the guard that a green run really compared something, would have run only in jobs where the flag is unset and it skips by design. **That is a behaviour change for CI**: ten unit tests now also run in `sweeps`, armed, and the wiring guard is live there. It is appended to that job's `-E` rather than added to `SWEEP_FILTER` because `SWEEP_FILTER` is negated by `test` and `test-oracle` (naming it there would *remove* the module from both rather than add it anywhere), and because `sweep_filter_invariant.rs` fails on a module in that filter which does not draw through `sweep_seeds(`. The module is therefore run twice on purpose — unarmed in `test`, armed in `sweeps` — which is the one deliberate exception to that filter's partition claim, recorded in its comment. Locally, the exact composed selection with all four flags set is 53/53 green.

**Not** `test-oracle`, whose selection contains exactly the two rows above. This is the one job where the set of four is now incomplete, and the comment there says so and names both issues. Suppressing them would hollow out the oracle; enabling the flag before they are settled would redden the job for a reason no PR caused. Move it once both are closed.

## Reviewing

`src/spdi/apply.rs` first (the primitive and, more to the point, what it declines to answer), then `src/normalize/mod.rs` (the seam), then the test file. The two CI hunks and `CLAUDE.md` are the same argument written for a reader who hits a fire later.

## Not done

- The oracle's *seam wiring* is only exercised when the flag is set. `the_seam_compares_when_the_flag_is_set` now closes the half that matters — with `FERRO_ASSERT_SEQUENCE` on it normalizes, then asserts `denoted_sequence_oracle_counts().0 > 0`, so removing the `assert_denoted_sequence` call from `assert_seam_oracles` reddens `sweeps` instead of leaving it green with zero comparisons. Unarmed it skips and says so. The `cargo test` race the earlier revision of this section worried about does not arise: the assertion is `> 0` on a monotonic counter the test increments itself, so it assumes no ordering between tests.
- The re-entrant pass is no longer compared. `assert_denoted_sequence` now honours `IN_IDEMPOTENCY_CHECK`, like `assert_idempotent` does: the idempotency oracle's verification pass re-enters `normalize_core_checked`, so with both flags set the seam was comparing `(normalized, again)` — a question the outer call already answers between its own comparison and the idempotency assertion. Measured on one normalization with `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_SEQUENCE=1`: `compared` was **2**, is now **1**, `skipped` **0** either way. It halves the provider fetches in `sweeps` (which sets both), keeps the counters a count of normalizations a caller asked for, and stops a non-idempotent output being reported as a sequence fault by the inner call before `assert_idempotent` can name it as the non-idempotency it is. The return is before either counter, like the flag-off path, so nothing moves into `skipped`.
- The corrected-`REFSEQ_MISMATCH` false-positive class is still untested. It is skipped in `Normalizer::assert_denoted_sequence` from the warning list rather than inside `compare_denoted_sequences`, so no input to the predicate can reach it; `the_measured_false_positive_classes_stay_silent` now says which of the seven measured classes it covers and which it cannot.

Representation-Change: none. Every one of the four asserts is `#[cfg(debug_assertions)]` and env-gated, so nothing runs unless `FERRO_ASSERT_SEQUENCE` is set; the only unconditional edit is a mechanical `Option` → `Result` refactor of `variant_edit_triples`, whose `.ok()` wrapper preserves its old signature and semantics exactly. Verified by the suite rather than asserted: 9,853 tests pass with **nothing re-blessed** — no fixture, ledger, census or snapshot in this diff.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,14 @@ env:
   # NOTE: that partition is over `test` + `sweeps`. `test-oracle` additionally
   # negates `ORACLE_EXCLUDE` below, so it runs a strict SUBSET of `test` rather
   # than the same selection. Every test still runs somewhere unarmed.
+  #
+  # NOTE also: `sweeps` selects on this filter PLUS one module it deliberately
+  # re-runs — `issue_1615_denoted_sequence_oracle`, the only test that reads the
+  # denoted-sequence oracle's counters, which are meaningless unless
+  # `FERRO_ASSERT_SEQUENCE` is set and `sweeps` is the only job that sets it. So
+  # "nothing is run twice" holds of this filter and is one module short of
+  # holding of that job's `-E`; the reason it is appended there rather than added
+  # here is on the job itself.
   SWEEP_FILTER: >-
     test(cis_junction_crossing_shift) + test(repeat_span_sibling_overlap)
     + test(issue_1254_sibling_crossing_shift)
@@ -980,7 +988,27 @@ jobs:
         # perturbs idempotency it need not. The class had been found by hand three
         # times (#1274, #1343, #1307) before it was asserted here.
         #
-        # All three sit at the same seam, `normalize_core_checked`'s single exit.
+        # `FERRO_ASSERT_SEQUENCE` — the fourth oracle at this same seam (#1615) —
+        # is deliberately **not** set here, and this job is the one place the set
+        # is incomplete. It fires on two rows of this job's own selection, and
+        # both are real disagreements rather than noise to be suppressed:
+        #
+        #   * #1618 — `NC_TEST.1:g.262TG[6]` -> `g.259_262GT[6]`
+        #     (`repeat_tract_maximization`). `hgvs_to_spdi` reads the anchored
+        #     spelling as 6 copies replacing a 1-copy tract and the normalizer's
+        #     output as 6 copies replacing a 2-copy tract — 14 bases against 12.
+        #   * #1619 — `NM_000492.4:c.1520_1522del` -> `c.1521_1523del`
+        #     (`normalize_tests`). The two forms are genuinely equivalent on the
+        #     flat transcript; `hgvs_to_spdi` resolves the c. position by walking
+        #     the fixture's exon list, whose 1-base synthetic introns put it 10
+        #     bases 3' of where the normalizer reads.
+        #
+        # Turning the flag on here before those are settled would make the job
+        # red for a reason that is not the PR's; suppressing them would hollow out
+        # the oracle. So it runs in `sweeps` below — the cis-allele space every
+        # one of the eight defects it was built for came from, and where it is
+        # measured green — and non-gating in the nightly. Move it here once both
+        # rows are closed.
         env:
           FERRO_ASSERT_IDEMPOTENT: "1"
           FERRO_ASSERT_REPARSE: "1"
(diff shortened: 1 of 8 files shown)


nh13:
@coderabbitai pause

Issue #1621, opened 2026-08-11
Normalizer: clamp the 3' rule at an exon/exon junction from BOTH sides — c.3922dup must converge on c.3921dup


nh13:
The ruling record `exon-junction-dup-converge-from-the-far-side` (in `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json`) is now **`decided`: CONVERGE**. `LRG_199t1:c.3922dup` must normalize to `LRG_199t1:c.3921dup`. Ferro leaves both standing as fixed points today. This issue is the implementation half.

## The rule

**The canonical position is the most 3' position that does not cross an exon/exon junction, reached from either side.** That is the 3' rule clamped at the junction, with the clamp binding in *both* directions:

* `c.3921dup` is already at the clamp and stays — this is what ferro does today, and it is the direction the exception is normally tested.
* `c.3922dup` is past the clamp and is pulled back — this is the missing half.

Scope: deletions and duplications, on the `c.` and `n.` axes (and `r.`, until upstream PR #266 lands).

## Why

`docs/recommendations/DNA/duplication.md` states it three times, independently:

* `:26` — "the variant is described as `LRG_199t1:c.3921dup` (`NC_000023.10:g.32459297dup`) and not as `c.3922dup`"
* `:60` — "the variant (`NC_000023.11:g.32441180dup`) is **not described** as `c.3922dup` since this would shift the position of the variant to the next exon"
* `:148` — "an exception was made to prevent that when `c.3922dup` is translated back to a genomic position, one would end up at the wrong nucleotide, in the wrong exon"

`general.md:44` is **directional**, not a blanket suspension: it stops the 3' rule carrying a description *across* the junction. Ferro treating both spellings as fixed points is that exception half-applied.

## The geometry, from the prepared reference's cdot

`NM_004006.2` (DMD), GRCh38, minus strand, `start_codon` 244:

| exon | cdna | genomic (cdot, 0-based half-open) |
|---|---|---|
| 27 | `[4031, 4165]` | `[32441179, 32441314]` |
| 28 | `[4166, 4315]` | `[32438240, 32438390]` |

[...]

Issue #266, opened 2026-05-17
fix(validation): warn on del/dup/inv length-mismatch input (closes #81 K1 follow-up)


nh13:
ferro silently accepts `del/dup/inv` inputs where the explicit reference sequence length does not match the position range length.

## Probe

```
NC_000001.11:g.100_110delAAAATTTGCC  // 11 positions, 10-base seq
NC_000001.11:g.100_105invTAGCA        // 6 positions, 5-base seq
```

Both parse cleanly today. Display preserves verbatim. SPDI conversion downstream clips to the seq length and silently changes the end position (`g.100_110delAAAATTTGCC → g.100_109delAAAATTTGCC`) — surfaced as out-of-scope by the K1 audit (#259).

## Proposed

Add a new SVA code (e.g. `W3016 LengthMismatch`) emitted in lenient mode when:

- the edit is `del` / `dup` / `inv` / `delins` with explicit reference sequence, AND
- the position range length (computable as `end - start + 1` for simple integer ranges) does not match `reference_sequence.len()`

Strict mode rejects with a diagnostic pointing at the conflicting span. Lenient mode keeps the input but emits the warning so callers can flag potentially-corrupt data.

Out of scope: position ranges with offsets or markers (`c.100+5_99+3`) — defer until a unified position-axis length helper exists.

Cross-ref: #259 (K1 audit), spec at https://hgvs-nomenclature.org/stable/recommendations/general/ (range semantics).

PR #1623, opened 2026-08-11
docs(rulings): decide the exon-junction duplication and codon carve-out records

nh13:
Two operator rulings, one commit each. Neither moves a ferro output in this PR; both commit the project to a position, and one of them commits it to a future move.

## 1. `exon-junction-dup-converge-from-the-far-side` → **`decided`: CONVERGE**

`LRG_199t1:c.3922dup` must normalize to `LRG_199t1:c.3921dup`.

**The rule.** The canonical position is the **most 3' position that does not cross an exon/exon junction, reached from either side** — the 3' rule clamped at the junction, with the clamp binding in *both* directions. `c.3921dup` is at the clamp and stays; `c.3922dup` is past it and is pulled back. Scope: deletions and duplications, `c.`/`n.` (and `r.` until upstream PR #266 lands).

**Grounds.** `DNA/duplication.md` states the outcome three times, independently:

* `:26` — "the variant is described as `LRG_199t1:c.3921dup` (`NC_000023.10:g.32459297dup`) and not as `c.3922dup`"
* `:60` — "the variant (`NC_000023.11:g.32441180dup`) is **not described** as `c.3922dup` since this would shift the position of the variant to the next exon"
* `:148` — "an exception was made to prevent that when `c.3922dup` is translated back to a genomic position, one would end up at the wrong nucleotide, in the wrong exon"

One lowercase-prose clause is weak authority here — that is the standing lesson of the RFC 2119 census — but three independent statements of one outcome, one giving a correctness rationale, is as close to settled as this spec gets.

**Two corrections inside the record.** The reading that `c.3922dup` "may not be a mis-spelling at all" but a *different legitimate genomic event* is **withdrawn**: `:148` calls that position "the wrong nucleotide, in the wrong exon", which is an error the exception exists to prevent, not a second event. And `general.md:44` is **directional** rather than a blanket suspension — it stops the 3' rule carrying a description *across* the junction, so ferro treating both spellings as fixed points is that exception **half-applied**, not working.

**New evidence, measured from cdot** (`NM_004006.2`, GRCh38, minus strand, `start_codon` 244), which the record did not have:

| exon | cdna | genomic (cdot, 0-based half-open) |
|---|---|---|
| 27 | `[4031, 4165]` | `[32441179, 32441314]` |
| 28 | `[4166, 4315]` | `[32438240, 32438390]` |

`c.3921` = cdna 4165 = transcript-**last** base of exon 27; `c.3922` = cdna 4166 = transcript-**first** base of exon 28. The junction sits **exactly** between the two positions in dispute — which is why the pair projects 2,790 bp apart (`g.32441180` against `g.32438390`, matching the spec's own GRCh37 pair), rather than one base apart.

**Test and fixture consequences.** `spec_expected` for `LRG_199t1:c.3922dup` was `null` on the reasoning that the spec published no answer. It publishes one, three times, so the field becomes `LRG_199t1:c.3921dup` and the row is now a **recorded divergence** — ferro's three mode outputs are pinned at today's values and none of them is the spec's answer. The unanswered-row set is consequently empty and the diverging set has two members instead of one; both assertions are updated with the reason. `the_decided_target_is_convergence_on_the_near_side` is `#[ignore]`d and asserts the target in all three modes (verified to fail today for exactly the right reason). Implementation: **#1621**.

## 2. `codon-carve-out-shape-restriction` → **`decided`: WIDEN**

The `delins.md:18` / `general.md:35` exception applies wherever its stated precondition holds — two variants separated by one nucleotide, together affecting one amino acid — **regardless of edit type**. Ferro's restriction to a substitution / unchanged base / substitution shape is dropped.

**Ground 1, and it is the same argument that decided `separation-is-a-property-of-the-spelling-not-of-the-variant` (#1620).** Rule 3 of the README ruleset: "Every rule is evaluated over the resulting sequence, never over the input's spelling." **Edit type is a property of the spelling; "together affecting one amino acid" is a property of the resulting sequence.** Gating the exception on edit type is exactly that spelling-dependence, and it makes the carve-out non-confluent by construction — one variant spelled two ways gets the exception once and not twice.

**Ground 2.** Rule 1 is conformance, and the clause names no edit type. A narrowing needs its own authority; none was produced for this one.

**Both existing constraints survive**, restated as limits on the widening rather than objections to it: `general.md:35`'s "together affecting one amino acid" cannot cover a **frameshift** pair (the precondition simply fails), and `duplication.md:18` still governs, so the widened carve-out must not turn `c.[4C>T;5dup]` into `c.4delinsTA`.

**The 5.76M-row inertness measurement is kept and re-labelled.** The ruling does **not** rest on it; the record now says so explicitly, so a later reader cannot mistake it for support. A corpus zero is a claim about the corpus.

**Implementation: PR #1599**, which already gates on `same_codon` with edit type out of the predicate. It carries its own measured blast radius (1,581 of 78,298 corpus rows, all merge → split) and a passing `Representation change declared` check. Nothing in this PR touches it.

## Verification

* All seven clause quotes across the two records verified **byte-exact as raw substrings** of the cited lines in the pinned submodule (`6f85311`) — not merely whitespace-collapsed matches. `generate_spec_fixture` validates them too and builds clean.
* `tests/it/clause_ruling_index.rs`'s generated index regenerated (`duplication.md:60` is a new clause citation, and two records changed status). Census moves to 11 records / 9 decided / 2 undecided, 48 clause lines, 7 multiply-named, 3 mixed-verdict. The module docs now note that a falling mixed count means records got *decided*, not that a line became safe to read from one record.
* Two prose citations calling `codon-carve-out-shape-restriction` undecided are updated (`coding_frame_merge_axis_asymmetry.rs`, `multi_member_cis_axis.rs`); the first is rewritten to say the ruling widens the carve-out and therefore still cannot license the frameshift merge that file is about.
* Full suite: **9,849 passed, 0 failed, 151 skipped**. `cargo fmt --check` and `cargo clippy --all-features --all-targets -- -D warnings` clean.
* Each commit was verified green on its own, not only the final state.

**Note on merge order:** this branch and #1620 both regenerate the index block in `tests/it/clause_ruling_index.rs` and both edit `RULING_STATUSES`, so whichever merges second needs a rebase and one index regeneration. No semantic conflict.

Representation-Change: 0 rows move in this PR — no ferro output changes here, and the corpus is untouched. What is declared is the DECISION: `exon-junction-dup-converge-from-the-far-side` is now decided for convergence, so every stored `c.3922dup`-shaped description will change to `c.3921dup` and its projected genomic form will shift 2,790 bp once #1621 implements it; and `codon-carve-out-shape-restriction` is decided for widening, whose implementation #1599 measures at 1,581 of 78,298 corpus rows. Recorded here so both decisions are disclosed in the release they were made in, not only in the release that ships the moves.


nh13:
@coderabbitai pause

PR #1625, opened 2026-08-11
fix(corpus): verify a row against its reference sequence, not its label

### LINKED ISSUE #1624: dump_normalized_corpus: --verify-spdi builds its provider from the reference column, which is a label for two families
## Symptom

`--verify-spdi` — the committed check that says whether a moved row became *wrong* rather than merely re-spelled (#1514) — panics before it verifies anything:

```
$ cargo run --features dev --example dump_normalized_corpus -- --out /tmp/after.tsv --verify-spdi
thread 'main' panicked at examples/dump_normalized_corpus.rs:1335:
byte index 20 is out of bounds of `revcomp_sep0_at0`, which contains 16 bytes
```

So the flag is unusable as shipped, and a release measurement that depends on it cannot be made.

## Cause

`Row.reference` is the dump's row-identity column, and for thirteen of the fifteen families it **is** the reference sequence — a 20-base core. For **two** families it is a *label*, deliberately, because the sequence would otherwise be repeated verbatim on every row:

- `long_block_inversion` → `nearpalindrome_1024`, `nearpalindrome_1100` (`long_corpus_sequences`)
- `separated_revcomp_runs` → `revcomp_sep{separation}_at{offset}` (`RevcompDesign::label`)

Both places say so in their doc comments. What was missed is that `verify_row` rebuilds the provider with

```rust
provider_for(row.axis, &row.reference)
```

which hands the **label** to the provider constructors as if it were the core. Two different failures follow, and only one of them is loud:

1. **`separated_revcomp_runs`, `cx` axis — panic.** Every label there is exactly 16 bytes (`revcomp_sep0_at0`), and `EXON_SPANS`'s last span is `(14, 20)`, so `spliced_genomic` slices `core[14..20]` off the end of the label. That is the message above.
2. **`long_block_inversion`, `g.`/`c.` axes — silently wrong.** `nearpalindrome_1024` is 19 bytes of `[a-z0-9_]`. The genomic and coding providers accept any string, so those rows *do* get verified — against a 19-character reference that has nothing to do with the kilobase near-palindrome the row was normalized against. No panic, no diagnostic; whatever verdict comes out is about the wrong sequence.

(2) is the worse half. A verification pass that answers confidently about the wrong reference is exactly what a verification pass must not do, and it would have gone on being read as a result.

Note that `verify_row`'s own doc comment states the invariant it violates — it shares `provider_for` with the normalize path specifically so that "the `--verify-spdi` pass reads the same reference the row was normalized against". The dump path passes the real core (`normalizer_for(axis, &design.core, …)`); the verify path passes the label. The test module's `apply_verdict` gets this right, because it holds the `RevcompDesign` and reads `design.core`.

## Why a clamp is not the fix

Truncating or clamping the `EXON_SPANS` slice would stop the panic and leave a provider built over `revcomp_sep0_at0` — a *wrong* reference rather than a crashing one, i.e. it converts failure (1) into failure (2). The correct reference exists; it is simply not in the column being read.

## Proposed fix

Carry the reference **sequence** on `Row` alongside the label, and have `verify_row` build its provider from that. For the thirteen sequence-keyed families the two are equal, so nothing about the dump's identity, its column set or any committed baseline moves — the sequence is not emitted.

Additionally, make `spliced_genomic` assert the invariant it depends on (the core is exactly the length `EXON_SPANS` maps) so a future recurrence is a named failure rather than an index panic.

## Acceptance

- `--verify-spdi` runs to completion over the whole corpus and prints its summary.
- The two label families are verified against their real cores, not their labels.
- The dump itself is byte-identical to a run without the flag, as documented.
- A test pins that the verify path and the dump path use the same reference for a label-keyed row.

nh13:
Closes #1624.

Representation-Change: none. `examples/` only — no watched directory is touched, and the dump this harness writes is byte-identical before and after (`cmp` clean against a run of the same binary without the flag). Verified rather than assumed: 9,853 tests pass with **nothing re-blessed**.

## The defect

`--verify-spdi` is the committed answer to "did that moved row become *wrong*, or merely re-spelled" (#1514), and `CLAUDE.md` points at it as the way to measure a representation change. It could not be run:

```
thread 'main' panicked at examples/dump_normalized_corpus.rs:1335:
end byte index 20 is out of bounds for string of length 16
```

`Row.reference` is the row-identity column. For thirteen of the fifteen families it **is** the 20-base core; for the two that bring their own designed references it is a *label* — `nearpalindrome_1024`, `revcomp_sep0_at0` — because a kilobase core repeated on every row is not a column anyone can read. Both places document that. What was missed is that `verify_row` rebuilt its provider with `provider_for(row.axis, &row.reference)`, handing the label to the provider constructors as if it were the sequence.

Its own doc comment states the invariant it broke: `provider_for` is shared with the normalize path precisely so the verify pass "reads the same reference the row was normalized against". The dump path passes `design.core`; the verify path passed `design.label`.

## Two failures, and the quiet one is worse

**1. `cx` axis — the panic.** Every `separated_revcomp_runs` label is exactly 16 bytes, `EXON_SPANS`'s last span is `(14, 20)`, so `spliced_genomic` sliced past the end. `verify_row` builds its provider *outside* its own `catch_unwind`, so one row took the entire pass down.

**2. `g.`/`c.` axes — fabricated findings.** The genomic and coding providers accept any string. So before the panic the pass had already emitted **eight** confident `SPDI-MISMATCH` findings, with keys read out of the letters of the label:

```
SPDI-MISMATCH	separated_revcomp_runs	g 3prime	revcomp_sep0_at0
    in : NC_TEST.1:g.257_260delinsTGGT  => NC_TEST.1:256:revc:TGGT
    out: NC_TEST.1:g.257_260inv         => NC_TEST.1:256:revc:GBEY
```

`revc` is `revcomp_sep0_at0[0..4]`. `GBEY` is not DNA. Those eight rows are the entire output a reader would have taken away from this flag, and every one of them is an artifact of the reference being a label. A verification pass that answers confidently about the wrong sequence is the one thing a verification pass must not do.

## Why not clamp the slice

A clamp stops the panic and leaves a provider built over `revcomp_sep0_at0` — i.e. it converts failure 1 into failure 2, which is the strictly worse one. The correct reference exists; it is simply not in the column being read. So `Row` now carries it.

The sequence is **not** emitted. `read_dump` exact-matches the header, so an added column would stop every committed baseline from loading, and a row's identity does not need the sequence when the label already determines it.

`spliced_genomic` also now asserts its length invariant, in both directions. Short is the panic above; long is the silent direction — `multi_exon_provider` hands `Transcript::new` the whole core, so a longer core declares bases the exon table does not map, which is why `dump` already skips `cx` for the kilobase cores. The assert puts that reasoning where the assumption lives.

## It runs

```
$ dump_normalized_corpus --out /tmp/after.tsv --verify-spdi
verified 78298 rows: 6 denote different bases, 57776 agree, 20516 unverifiable
```

Exit 0. The eight fabricated findings are gone; the six that remain are real, and are one class — an insertion pair straddling the 5'UTR/CDS junction on the multi-exon axis, where the two spellings key four to six bases apart:

```
SPDI-MISMATCH	adjacent_junction_ins	cx 3prime	CGGAACTGTGCTCGTTACTC
    in : NM_TESTX.1:c.[-1_1insAA;1_2insTT]  => NM_TESTX.1:5::ATTA
    out: NM_TESTX.1:c.1_2insTTAA            => NM_TESTX.1:4::TTAA
```

Both label families verify clean, which is the point: they were never being verified at all.

Reporting that finding is not this PR's job — it is what the flag is for, and it is now available to whoever measures the release.

## Guards

- `a_label_keyed_row_is_verified_against_its_core` — runs `verify_row` over the label-keyed rows and requires it to answer. Confirmed to bite: with `&row.core` reverted to `&row.reference` it fails immediately, and restored it passes.
- `every_row_carries_the_sequence_it_was_drawn_against` — pins the distinction on both sides, since the two views agree on thirteen of fifteen families and that is exactly what let the confusion survive review.

### DIFF
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -242,6 +242,15 @@ enum SpdiVerdict {
 /// Wrapped in `catch_unwind` for the same reason [`normalize_one`] is: this runs
 /// over every row of a corpus built to contain the shapes that break things, and
 /// one panicking row must not take the whole dump with it.
+///
+/// The provider is built from [`Row::core`] and **not** from `Row::reference`
+/// (#1624). Those two are the same string for thirteen of the fifteen families,
+/// which is what made reading the wrong one survive review: the two families
+/// whose `reference` column is a *label* — `long_block_inversion` and
+/// `separated_revcomp_runs` — were verified against the label itself. On the
+/// `cx` axis that panicked, and on `g.`/`c.` it did something worse, reporting
+/// confident `SPDI-MISMATCH` findings whose keys were read out of the letters of
+/// `revcomp_sep0_at0`.
 fn verify_row(row: &Row) -> (SpdiVerdict, String, String) {
     let declined = |what: &str| format!("<{what}>");
     let (Ok(input), Ok(output)) = (parse_hgvs(&row.input), parse_hgvs(&row.output)) else {
@@ -256,7 +265,7 @@ fn verify_row(row: &Row) -> (SpdiVerdict, String, String) {
         _ => ShuffleDirection::ThreePrime,
     };
     let normalizer = Normalizer::with_config(
-        provider_for(row.axis, &row.reference),
+        provider_for(row.axis, &row.core),
         NormalizeConfig::default().with_direction(direction),
     );
     let key = |v: &_| match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
@@ -361,7 +370,24 @@ fn main() -> ExitCode {
 /// makes the cheap/expensive split derivable from the baseline dump alone, without
 /// re-running the old code.
 struct Row {
+    /// The dump's `reference` column, and part of the row's identity. For
+    /// thirteen of the fifteen families this **is** the reference sequence; for
+    /// the two that bring their own designed references it is a label, because a
+    /// kilobase core repeated on every row is not a column anyone can read.
     reference: String,
+    /// The reference **sequence** the row was drawn against.
+    ///
+    /// Equal to `reference` except on `long_block_inversion` and
+    /// `separated_revcomp_runs`, where that column is a label. Deliberately
+    /// **not** emitted: the dump's format is exact-matched on read, so adding a
+    /// column would stop every committed baseline from loading, and a row's
+    /// identity does not need the sequence when the label already determines it.
+    ///
+    /// It exists because a consumer that needs to *re-derive* the row — which is
+    /// what `--verify-spdi` does — cannot get the sequence back out of a label,
+    /// and reading the label as one is not an error either provider reports
+    /// (#1624).
+    core: String,
     axis: &'static str,
     direction: &'static str,
     family: &'static str,
@@ -866,6 +892,7 @@ fn dump(seeds: u32) -> Vec<Row> {
                 let output = normalize_through(&normalizer, &input);
                 rows.push(Row {
                     reference: label.clone(),
+                    core: core.clone(),
                     axis,
                     direction: dir_label,
                     family: LONG_FAMILY.0,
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1633, opened 2026-08-11
test(conformance): pin what the output denotes, not only how it is spelled

### LINKED ISSUE #1626: Two normalization pins assert a string with no assertion about the sequence it denotes
## The gap

Two tests pin a normalized output **string** and assert nothing about the sequence that string denotes. Neither file mentions `hgvs_to_spdi`, `apply_to_reference`, `cis_apply_oracle`, `canonical_spdi` or any reconstruction of the output's bases:

| test | what it pins |
|---|---|
| `tests/it/coding_frame_merge_axis_asymmetry.rs::the_coding_axis_merges_what_the_other_axes_split` | 14 rows x 3 axes of exact output strings |
| `tests/it/delins_equal_vs_unequal_length_hermetic.rs::equal_and_unequal_length_delins_get_opposite_verdicts_hermetically` | two output strings for one span |

Both were added by #1579 and both are honest about being change detectors — `coding_frame_merge_axis_asymmetry.rs`'s `ROWS` says so in as many words. That is not the problem. The problem is that a string equality cannot tell a **re-spelling** from a **corruption**: if the normalizer began emitting a description of different bases that happened to render as the pinned string, both tests stay green.

This is not a hypothetical class here. It is the class #1615 exists for, and #1592 and #1600 are both live instances on `main` where a well-formed, in-bounds, idempotent, re-parseable output denotes different bases.

## Why these two specifically

Every row in both tests is a **re-spelling** — the pinned output is not the input:

- `c.9_10delinsATC` (net +2) is pinned as splitting to `n.[8dup;9_10insC]` on the `n.` and `g.` axes and as staying merged on `c.`. A `dup` names no bases in its rendering, so a composition that duplicated the wrong base still prints the identical string.
- `g.264_266delinsTC` is pinned as splitting to `g.[264del;266G>C]`, and that expectation is deliberately **assert-then-flip** — it records an answer the project believes wrong and is waiting to change. A pin whose whole purpose is to be flipped is exactly the one that must be able to say whether today's wrong answer is at least denoting the right bases.

`coding_frame_merge_axis_asymmetry.rs`'s sibling `every_axis_is_self_confluent_across_both_spellings` does not close this: it asserts the two spellings **agree**, and agreement cannot distinguish "all correct" from "all wrong" — the same argument `cis_allele_confluence_proptest.rs` makes about itself.

## Proposed fix

Add a denoted-sequence assertion alongside each existing string assertion, using the applier that already exists — `tests/it/common/cis_apply_oracle.rs`'s `apply_with` / `apply_reason`. It reaches the bases through `hgvs_to_spdi` and an SPDI splice with no normalization in the path, so it cannot agree with an output merely because normalization produced it. No new applier.

The assertion is: applying the **input** to the reference and applying the pinned **output** to the reference produce the same sequence.

## Acceptance

- Each new assertion is shown to bite — a same-string-different-sequence mutant makes it red.
- No pinned string changes, and no output moves (`Representation-Change: none`).

## Also seen, not in scope here

`tests/it/cis_confluence_adjudication.rs::a_dup_flush_against_a_del_is_left_alone` has the same gap and is the more surprising instance, because that file defines its own apply oracle and its module docs claim every case is checked through it — but that check is applied to the *inputs* only ("these spellings are one variant"), never to the normalized output. Three tests there re-spell with the output's denotation unasserted. Worth a follow-up.

nh13:
Closes #1626.

Representation-Change: none. Tests only — `tests/it/` is the whole diff, no watched directory is touched. Verified rather than assumed: the diff changes no pinned string (the only edit to one is a borrow, `format!(…)` to `&format!(…)`), and 9,856 tests pass with **nothing re-blessed**.

## The gap

Two tests pinned a normalized output **string** and asserted nothing about the sequence that string denotes. Neither file mentioned `hgvs_to_spdi`, `apply_to_reference`, `cis_apply_oracle` or `canonical_spdi`.

| test | rows |
|---|---|
| `coding_frame_merge_axis_asymmetry::the_coding_axis_merges_what_the_other_axes_split` | 14 positions x 3 axes |
| `delins_equal_vs_unequal_length_hermetic::equal_and_unequal_length_delins_get_opposite_verdicts_hermetically` | 2 arms |

A string equality cannot tell a **re-spelling** from a **corruption**. An output describing different bases that happened to render as the pinned string satisfies it — and so does the more realistic version, where the normalizer starts corrupting, the string moves, and someone re-blesses the pin without noticing the bases moved too. That is the class #1615 exists for, and #1592 and #1600 are live instances on `main` today where a well-formed, in-bounds, idempotent, re-parseable output denotes different bases.

Both files are honest about being change detectors; `ROWS` says so in as many words. That is not what is being fixed. What is being fixed is that a change detector which cannot see a corruption is detecting less than it appears to.

**30 of the 44 comparisons are re-spellings** — the pinned output differs from the input on the `n.` and `g.` columns and on both delins arms, which is what makes this load-bearing rather than theoretical:

- The axis-asymmetry table's outputs are `dup`s and `ins`s. **A `dup` names no bases in its rendering**, so a composition that duplicated the wrong base prints the identical string.
- The unequal-length `delins` arm is **assert-then-flip**: it deliberately records an answer the project believes wrong and intends to change. A pin whose purpose is to be flipped should be able to say that today's wrong answer is wrong about the *partition* and not about the *bases* — which is the difference between a representation change and a defect.

Neither is closed by what was already there. `every_axis_is_self_confluent_across_both_spellings` asserts the two spellings **agree**, and agreement cannot distinguish "both correct" from "both wrong" — the argument `cis_allele_confluence_proptest.rs` makes about itself. The delins file's existing assertions (fixture bases, minimality, payload arithmetic) are all **input**-side.

## The fix

Each string assertion is now paired with: applying the input and applying the pinned output produce the same sequence.

The applier is `tests/it/common/cis_apply_oracle.rs`'s `apply_reason`, not a third hand-rolled copy. It reaches the bases through `hgvs_to_spdi` and an SPDI splice with **no normalization in the path**, so it cannot agree with an output merely because normalization produced it. It panics on a decline rather than returning `Option`: every description here is a lone `delins` or a two-member pair on disjoint columns of a synthetic reference, so a decline is a defect, and a silently skipped comparison is the exact failure mode this check exists to remove.

**All 44 comparisons pass, and 30 of them can fail.** The other 14 are the `c.` column, where the pinned output is byte-identical to the input — that non-split is the finding this file records — so `denotes(actual) == denotes(input)` is an identity there and cannot catch a corruption. Those 14 stay pinned *transitively*: `CDS_START == 1`, so the `n.` row's assertion fixes the same bases. The module doc of `tests/it/coding_frame_merge_axis_asymmetry.rs` says exactly this; the four demonstrated mutants below are all `n.`/delins, never `c.`, which is consistent with it.

One detail worth recording: the transcript axes apply against the **transcript sequence** and the genomic axis against the **padded contig**. Passing the padded contig for an `n.` row is rejected as `StatedBasesMismatch` rather than silently compared, which is itself evidence the oracle is reading the reference rather than rubber-stamping.

## Both new assertions were shown to bite

The mutant is the realistic failure, not an artificial one: corrupt the normalizer's output *and* re-bless the pin to match, so the string assertion passes and only the denotation assertion can object.

**`delins_equal_vs_unequal_length_hermetic`** — pin re-blessed to `g.[264del;266G>A]` and the normalizer stubbed to return it:

```
FAIL  equal_and_unequal_length_delins_get_opposite_verdicts_hermetically
```

**`coding_frame_merge_axis_asymmetry`** — the `p=9` `n.` row re-blessed to `n.[8dup;9_10insG]` and the normalizer stubbed to return it:

```
assertion `left == right` failed:
  NR_TEST.1:n.9delinsATC -> NR_TEST.1:n.[8dup;9_10insG] is not a re-spelling: it denotes different bases
  left:  ACGTACGAATGGCAAG…
  right: ACGTACGAATCGCAAG…
```

Both restored from a file copy afterwards, and the suite is green.

## Not in scope

`cis_confluence_adjudication::a_dup_flush_against_a_del_is_left_alone` has the same gap, and it is the more surprising instance: that file defines its own apply oracle and its module docs claim every case is checked through it — but the check is applied to the **inputs** only ("these spellings are one variant"), never to the normalized output. Three tests there re-spell with the output's denotation unasserted. Noted on #1626 for a follow-up rather than folded in here.

### DIFF
--- a/tests/it/coding_frame_merge_axis_asymmetry.rs
+++ b/tests/it/coding_frame_merge_axis_asymmetry.rs
@@ -133,11 +133,50 @@
 //! defect than a confluence failure — both spellings of the variant do reach one
 //! answer on every axis. Fact 1 below is what stops that regressing.
 //!
+//! # Every pinned output is also checked against the bases it denotes
+//!
+//! The strings above are change detectors, and a string equality cannot tell a
+//! **re-spelling** from a **corruption**: an output describing different bases
+//! that happened to render as the pinned string would satisfy it. That is not a
+//! hypothetical class here — it is what #1615 exists for, and #1592 and #1600
+//! are live instances on `main` where a well-formed, in-bounds, idempotent,
+//! re-parseable output denotes different bases.
+//!
+//! It bites hardest on this table because the outputs are `dup`s and `ins`s. A
+//! `dup` names no bases in its rendering, so a composition that duplicated the
+//! *wrong* base still prints the identical string.
+//!
+//! So each of the 42 comparisons — 14 rows x 3 axes — now also asserts that the
+//! output and the input denote the same sequence, through `cis_apply_oracle` —
+//! `hgvs_to_spdi` plus an SPDI splice, with no normalization in the path, so it
+//! cannot agree with an output merely because normalization produced it (#1626).
+//! All 42 pass.
+//!
+//! **28 of the 42 can fail; the `c.` 14 cannot, and saying which is which is the
+//! point.** The `n.` and `g.` answers are genuine re-spellings — a lone `delins`
+//! in, an insertion pair out — so a corrupted composition moves the applied
+//! sequence and the comparison catches it. The `c.` answer is the lone input
+//! *unchanged* (that non-split is the finding; [`ROWS`] records that the `c.`
+//! output is `c.{position}delinsATC` at every one of the fourteen, which is
+//! exactly the string [`spellings`] feeds in), so there `actual == input` and the
+//! comparison is an identity. That leaves the `c.` column pinned but not freshly:
+//! `CDS_START == 1`, so `c.p` and `n.p` are one position and the two axes' inputs
+//! denote the same bases of [`CORE`] — the `n.` row's assertion therefore already
+//! fixes what the `c.` answer denotes, transitively. What carries the `c.` row on
+//! its own is the string equality.
+//!
+//! Note what this does **not** buy. It cannot say the `c.` merge is the right
+//! form — the whole point of the assert-then-flip note above is that nothing has
+//! ruled on that — only that today's answer and tomorrow's flipped one are
+//! answers about the same bases.
+//!
 //! Fully hermetic: a `MockProvider`, no `FERRO_MANIFEST`, no fixtures.
 
 use ferro_hgvs::reference::transcript::{Exon, GenomeBuild, ManeStatus, Strand, Transcript};
 use ferro_hgvs::reference::MockProvider;
 use ferro_hgvs::{parse_hgvs, NormalizeConfig, Normalizer};
+
+use crate::common::cis_apply_oracle::apply_reason;
 
 /// Padding on each side of the core on the genomic contig, so that transcript
 /// position `p` is genomic position `PAD_OFFSET + p`.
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #1639, opened 2026-08-11
fix(normalize): make a partitioner decline observable, and stop the bake-off switch aborting a shipping process

### LINKED ISSUE #1635: normalize: a declined sequence-first partition silently serves `partition_block` under the candidate arm's name
`canonicalize_from_sequence` selects its block partitioner from `FERRO_PARTITION`, and the two sequence-first arms decline when the alignment grid would exceed `MAX_SEQFIRST_GRID_CELLS`. On a decline the caller silently serves `partition_block` — the shipped `live` rule — under the candidate arm's name:

```rust
PartitionRule::Canonical => {
    partition_block_canonical(&ref_bytes[lo..hi_ref], &result[lo..hi_alt])
        .unwrap_or_else(|| partition_block(&ref_bytes[lo..hi_ref], &result[lo..hi_alt]))
}
```

Nothing counts it and nothing logs it, and the `FERRO_SEQFIRST_SHADOW` report does not run on a `FERRO_PARTITION` arm's own path. So a bake-off cannot separate "the candidate answered" from "the candidate declined and `live` answered", and a run in which the candidate declined on every block reports that the candidate changes nothing.

That is the same failure mode `partition_rule_from_env` already refuses for an unrecognised arm name — "a measurement instrument that manufactures agreement" — reached through a different door.

The cap is reachable from a single description: a duplication of the full canonical window is `4096 x 8192`, or 33,559,521 cells against the 16,785,409 budget, so both seqfirst arms decline and `live` answers.

Related, in the same constant: `MAX_SEQFIRST_GRID_CELLS`'s doc comment counts four grids inside `AlignmentDag::build` and sizes the budget from that. It omits `AlignmentDag::dominators`' `in_degree` grid and its `order` vector, and `CanonicalAlignment::of`'s `to_go` (two `u32` planes) and its `cells` vector, so the stated peak is roughly half the real one.

### What would close it

A process-wide decline counter, in the shape of the existing `denoted_sequence_oracle_counts()`, readable after a run, plus a doc that accounts for every allocation the budget is meant to bound.

### LINKED ISSUE #1636: normalize: `FERRO_PARTITION` is read in release builds and panics from an infallible path
`partition_rule()` reads `FERRO_PARTITION` in every build, including release, and panics when the value is not an arm this build has:

```rust
fn partition_rule() -> PartitionRule {
    static RULE: std::sync::OnceLock<PartitionRule> = std::sync::OnceLock::new();
    *RULE.get_or_init(|| {
        let value = partition_rule_env_value(std::env::var("FERRO_PARTITION"));
        partition_rule_from_env(value.as_deref()).unwrap_or_else(|message| panic!("{message}"))
    })
}
```

Its own doc calls it "a **development-only** bake-off switch", and the code does not enforce that. Two consequences, both wrong for a library:

1. Any environment that happens to define `FERRO_PARTITION` re-partitions every cis allele in a shipping build.
2. An unrecognised value aborts the process from inside `canonicalize_from_sequence`, which is infallible and sits under every public normalization entry point — including the Python bindings, where a panic crosses an FFI boundary.

The three assertion oracles at the same seam (`FERRO_ASSERT_IDEMPOTENT`, `FERRO_ASSERT_REPARSE`, `FERRO_ASSERT_IN_BOUNDS`) are all `#[cfg(debug_assertions)]` for exactly this reason; this switch is not.

### The constraint a fix has to respect

Refusing an unrecognised value is not incidental — it exists because silently answering as `live` under a candidate's name makes the diff come back empty and the run read as *"the candidate changes nothing"*. So a fix must not simply drop the rejection: the knob has to keep its measurement integrity in the configurations that measure, while being unreachable as a release-build panic in the configuration that ships.

nh13:
Two fixes to the `FERRO_PARTITION` bake-off switch. They share the same twenty lines of `merge.rs`, so they are one PR with one commit each.

Closes #1635
Closes #1636

## 1. A declined sequence-first partition silently served `partition_block`

Both sequence-first arms decline when the alignment grid would exceed `MAX_SEQFIRST_GRID_CELLS`, and the caller then served `partition_block` — the shipped rule — under the candidate arm's name. Nothing counted it and nothing logged it, so a bake-off could not tell *"the candidate agreed with `live`"* from *"the candidate was never asked"*. A run in which the candidate declined throughout reported that the candidate changes nothing.

That is the failure `partition_rule_from_env` already refuses an unrecognised arm name to avoid — *"a measurement instrument that manufactures agreement"* — reached through a different door. The cap is reachable from one description: a duplication of the whole canonical window is `4096 x 8192`, or 33,559,521 cells against 16,785,409.

Adds `partition_decline_counts()`, a process-wide census incremented only on the sequence-first arms — `Live` has no decline path, so the shipped configuration pays no atomic per block — and reports it from `dump_normalized_corpus`, the harness a blast-radius measurement runs through.

Also corrects `MAX_SEQFIRST_GRID_CELLS`' own doc, which counted the four grids inside `AlignmentDag::build` and put the peak at "roughly 340 MB". The sweep that runs over the built DAG allocates more than the DAG it walks — `dominators`' `in_degree` and `order`, `CanonicalAlignment::of`'s `to_go` and `cells` — but it does not run *alongside* the build: `prefix` and `suffix_grid` are locals of `build` and are dropped when it returns, so only 2 of its 10 bytes/cell survive into the sweep. The peak is therefore `max(build, retained + sweep)` — **~14 bytes/cell** on `shadow` and **~18** on the `canonical` arms, not the sum of the two columns. ~18 bytes/cell over 16,785,409 cells is 302 MB, which is what the measured peak RSS of ~310 MB and 5–9 s in a debug build says it is; adding the columns would predict 369 MB and disagree with the measurement. The doc now tabulates every allocation and says plainly what the bound is and is not defensible as: fine as a cost bound derived from `MAX_CANONICAL_WINDOW`, not a number anybody would choose for a per-description budget, and lowering it is not free because every newly-refused block is a silent handover to `partition_block`.

**The failing test, observed before the fix.** With the two `fetch_add` calls neutralised, `a_declined_sequence_first_partition_is_counted` fails with `left: 0, right: 1` on the *denominator* — "without it a zero decline count cannot be told from an arm that never ran".

## 2. The switch could abort a shipping process

`partition_rule()` read `FERRO_PARTITION` in every build and **panicked** when the value named no arm — from inside `canonicalize_from_sequence`, which is infallible, sits under every public normalization entry point, and is reached across the FFI boundary by the Python bindings. The three assertion oracles at this same seam are all `#[cfg(debug_assertions)]` for exactly that reason.

The refusal itself is not the defect and is kept. What moves is *where it is delivered*: the parse outcome is cached as a `Result` rather than a `PartitionRule`, a debug build still aborts at the first normalized variant, and a release build falls safe to `live` and retains the refusal for `partition_switch_startup_error()`, which every binary in this repository calls after argument parsing and before reading any input — `ferro`, `ferro-web`, `ferro-benchmark`, both spec generators, and the `dump_normalized_corpus` bake-off harness.

Falling safe is only sound if the callers really do report it: falling safe *without* reporting is the original defect, a bake-off served the shipped rule under a candidate's name and coming back with a clean empty diff. So the wiring is the contract and `tests/it/partition_switch_wiring.rs` pins it from both sides — end to end through the `ferro` binary (a misspelling exits non-zero naming this build's arms and emits nothing on stdout, a real arm still runs, and `--help` keeps working under a stale value), and over `Cargo.toml`'s own `[[bin]]` table, so a binary added later cannot skip the call and leave the claim above quietly false. Verified to bite: with the call removed from `ferro-benchmark`, the completeness test fails naming it.

**Compiling the switch out of release builds was tried first and is wrong.** `README.md` documents the A/B recipe as working on a **stock release build**, and that is the only way to normalize a real corpus at size — so compiling it out would delete the measurement the switch exists for. Every arm is still honoured in every build; only a value naming *no* arm behaves differently.

**The failing test, observed before the fix.** With `may_abort` ignored, `a_refused_value_does_not_abort_a_build_that_may_not_abort` fails by panicking — which is the defect itself.

## Verification

- `cargo nextest run --features dev --no-fail-fast`: 9,991 tests, **0 failures**, nothing re-blessed.
- `cargo fmt --check`, `cargo clippy --all-features --all-targets` clean.
- `cargo build --release` then `FERRO_PARTITION=canonicl ./target/release/ferro normalize …` exits with the refusal before reading input.

Representation-Change: none. No partitioner rule changes; the arm selection for every recognised value is unchanged in every build, and the added census is read-only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,7 +367,7 @@ This works on a **stock release build** — the switch is not behind a build fea
 
 ### Two traps worth knowing
 
-**A misspelled value is refused, loudly.** `FERRO_PARTITION=canonicl` aborts at the first variant ferro normalizes, naming the value you gave and the arms this build has:
+**A misspelled value is refused, loudly.** `FERRO_PARTITION=canonicl` makes `ferro` exit with an error before it reads any input, naming the value you gave and the arms this build has:
 
 ```
 FERRO_PARTITION="canonicl" is not a partitioner this build has. This build's arms
@@ -377,6 +377,8 @@ name reports that the candidate changes nothing.
 ```
 
 Naming *this build's* arms is the point: a value that exists on some other branch, or that used to exist, is reported as absent here rather than quietly answered as `live`.
+
+The refusal comes from the CLI, not from the normalizer. Up to and including v0.14.0 it was a panic raised deep inside normalization, which meant a development-only switch could abort any process — a long-running service, or a Python caller, across the FFI boundary — that merely happened to have the variable set. A release build of the *library* now falls safe to `live` for a value that names no arm and keeps the refusal for its caller to report. Every binary in this repository reports it — `ferro`, `ferro-web`, `ferro-benchmark`, both spec generators, and the `dump_normalized_corpus` bake-off harness — and that is pinned by a test rather than by this sentence, so a binary added later cannot quietly skip it. If you embed ferro as a library and run bake-offs through it, read `ferro_hgvs::normalize::partition_switch_startup_error()` at startup and do the same.
 
 Builds up to and including v0.13.1 fell back to `live` instead, and produced a clean, empty diff that read as "the candidate changes nothing". The fallback emitted a warning through the `log` facade, but the `ferro` CLI installs no logger, so that warning reached no stream and `RUST_LOG` could not surface it — there was no signal at all. If you are on an older build, treat every empty diff as unproven.
 
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -288,6 +288,16 @@ fn verify_row(row: &Row) -> (SpdiVerdict, String, String) {
 
 fn main() -> ExitCode {
     let cli = Cli::parse();
+    // This is the harness a bake-off runs through, so it is the entry point that
+    // can least afford to serve `live` under a candidate's name. In a release
+    // build the library falls safe rather than aborting, and the two silences
+    // compound: the dump would come back identical to the `live` dump, and
+    // `report_partition_declines` would print nothing at all, because a `live`
+    // run attempts no sequence-first partition. Refuse here instead.
+    if let Some(message) = ferro_hgvs::normalize::partition_switch_startup_error() {
+        eprintln!("error: {message}");
+        return ExitCode::FAILURE;
+    }
     if let (Some(before), Some(after)) = (&cli.compare, &cli.against) {
         return match compare(before, after) {
             Ok(report) => {
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -302,6 +302,7 @@ fn main() -> ExitCode {
     }
 
     let rows = dump(cli.seeds);
+    report_partition_declines();
     let mut out = String::new();
     out.push_str(HEADER);
     for row in &rows {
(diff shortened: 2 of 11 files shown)


nh13:
@coderabbitai pause

PR #1640, opened 2026-08-11
fix(normalize): let the inversion coalesce reach a sequence-derived partition

### LINKED ISSUE #1637: normalize: `coalesce_whole_block_inversion` never fires on a canonical partition, so the spec's published inversion example is shredded
`coalesce_whole_block_inversion` puts a run of pieces that together span one reverse complement back into a single `inv`. It never fires on a partition produced by `partition_block_canonical`, so under `FERRO_PARTITION=canonical` the spec's own published inversion example is emitted as a five-member allele:

```
DNA/inversion.md:27-34 publishes  NM_004006.2:c.4145_4160inv
ferro (canonical) emits           NM_004006.2:c.[4145_4147delinsGTG;4151del;4153T>A;
                                                 4154_4155insT;4158_4160delinsGGA]
```

The smallest instance is `GCT -> AGC`, a textbook 3 nt inversion. `partition_block` returns one piece; `partition_block_canonical` returns `0:0/A|2:3/` — an insertion and a deletion, which denote the same sequence and are equally distance-minimal.

### It fails to fire; it is not fired and undone

The reconstruction is fine — splice the unchanged bases back between the pieces and you get `AGC`, which `is_inversion` accepts. What refuses is the admission gate, all three routes of which are calibrated on the *live* partitioner's substitution-shaped pieces:

| route | on `0:0/A\|2:3/` |
|---|---|
| `every_separation_is_a_single_base` | gap is `2 - 0 = 2` — refuses |
| `changed_columns_dominate_the_span` | reference widths `0 + 1 = 1`, span `3`; `2 > 3` is false — refuses |
| `no_piece_is_a_lone_substitution` | the insertion's reference width is `0` — refuses |

The second is the interesting one: it measures **reference** width, and a pure insertion changes columns while consuming no reference at all. So a partition that expresses a change as an insertion reads as *less* dense than the same change expressed as a substitution, even though the density argument in that predicate's own doc is about columns. A partitioner free to place compensating gaps produces insertion-shaped members routinely; the live one, restricted to a single gap, almost never does.

### Consequence

2,048 of 9,216 `inv_member` rows in the synthetic corpus move under `canonical`, and `tests/it/inversion_sweep.rs` loses `every_authored_inversion_produces_its_pinned_output` and three sibling guards.

nh13:
`coalesce_whole_block_inversion` puts a run of pieces that together span one reverse complement back into a single `inv`. It **never fired** on a partition `partition_block_canonical` produced, so under `FERRO_PARTITION=canonical` the spec's own published example came out as a five-member allele:

```
DNA/inversion.md:27-34 publishes  NM_004006.2:c.4145_4160inv
ferro emitted                     c.[4145_4147delinsGTG;4151del;4153T>A;
                                    4154_4155insT;4158_4160delinsGGA]
```

Closes #1637

## It fails to fire; it is not fired and undone

The audit could not separate these two and they call for opposite fixes, so the new unit test asserts the answer rather than describing it. The reconstruction is **sound** — splice the unchanged bases back between the pieces and `is_inversion` accepts the result. What refused was the admission gate, all three routes of which are calibrated on the live partitioner's substitution-shaped pieces. On the smallest instance, `GCT -> AGC`:

| route | verdict on `0:0/A \| 2:3/` |
|---|---|
| `every_separation_is_a_single_base` | gap is `2 - 0 = 2` — refuses |
| `changed_columns_dominate_the_span` | reference widths `0 + 1 = 1` against a span of `3` — refuses |
| `no_piece_is_a_lone_substitution` | the insertion's reference width is `0` — refuses |

The second is the interesting one. It measures **reference** width, and a pure insertion consumes no reference at all, so it contributes zero — meaning a partition that expresses change as an insertion reads as *less dense* than the same change expressed as a substitution. A partitioner free to place compensating gaps produces insertion-shaped members routinely; the single-gap search almost never does, which is why the gap was invisible until now.

## The fix

`payload_columns_dominate_the_span`, a fourth `||` alternative counting `max(reference width, payload)` — the quantity the density argument in the sibling predicate's own doc was always about.

**Additive by construction**, so every block that coalesces today still coalesces and only a currently-*refused* block can change. The two pairs the earlier routes exist to keep apart are both length-preserving, where the new measure is identical to the old one:

| block | pieces | changed | span | verdict |
|---|---|---|---|---|
| `AAGCTA -> TAGCTT` (the #1040 control) | `0:1/T`, `5:6/T` | 2 | 6 | refused |
| `GATG -> CATC` (#1230, two substitutions) | `0:1/C`, `3:4/C` | 2 | 4 | refused |

And every admission still has to survive `is_inversion` on the reconstructed span — an exact, equal-length reverse complement — so `A -> CAC` passes the predicate and is refused there. That case is asserted too.

## The failing tests, observed before the fix

With the new `||` alternative removed:

```
the_inversion_coalesce_reaches_a_canonical_partition
  left:  [Piece { ref_start: 0, ref_end: 0, alt: "A" }, Piece { ref_start: 2, ref_end: 3, alt: "" }]
  right: [Piece { ref_start: 0, ref_end: 3, alt: "AGC" }]

inversion_sweep::the_spec_published_inversion_examples_produce_the_spec_form
  the spec prints `NM_004006.2:c.4145_4160inv` as the description to use
  (DNA/inversion.md:27-34); ferro rewrote it
  left:  "NM_004006.2:c.[4145_4147delinsGTG;4151del;4153T>A;4154_4155insT;4158_4160delinsGGA]"
  right: "NM_004006.2:c.4145_4160inv"
```

## Measurements

Full suite, `cargo nextest run --features dev --no-fail-fast`, 9,991 tests, against a **0-failure** baseline:

| configuration | distinct failures |
|---|---|
| baseline (`FERRO_PARTITION` unset) | 0 |
| `FERRO_PARTITION=canonical`, before | 78 |
| `FERRO_PARTITION=canonical`, after | **68** |

12 repaired, including the spec's published example; 2 added, both generated-artifact replay detectors (`spec_enumeration_tests::{divergence_budget_is_unchanged, passing_census_is_unchanged}`) which are regenerated-from-the-code-under-test and are expected to move under a non-default arm.

**On the shipped `live` arm, nothing moves.** The full suite under the default arm is green with nothing re-blessed. (An earlier run showed three `spec_enumeration_tests` failures; those were a locally generated enumeration artifact left behind by the preceding `canonical` run, and they pass after regenerating it — a stale-artifact detector doing its job, not a regression.)

## What this does NOT fix, stated plainly

A 5 nt inversion whose canonical partition is one deletion plus one insertion — `g.257_261inv` becoming `g.[257del;261_262insC]` — claims 2 changed columns against a 5-base span, so **no** density rule can admit it. Roughly a dozen `inv`-family rows in the canonical suite remain for that reason, and they need a different argument (the members' *types*, per `general.md:56`) rather than a wider threshold. Filed as follow-up work rather than papered over here.

Representation-Change: none. 0 rows move on the shipped `live` arm — the full suite is green under the default arm with nothing re-blessed, and the new predicate is an additive `||` alternative that can only admit blocks the three existing routes already refuse. The 12 rows it moves are all under `FERRO_PARTITION=canonical`, which is an unstable evaluation switch, not a shipped configuration. Reviewed once more for the *other* consumer of the widened gate: `whole_block_inversion` is also called from `split_concealed_separations`, where a `Some` verdict is an early return rather than a merge, so admitting a partition there suppresses an audit instead of adding one. That direction is inert -- the guard is reached only when the block is length-changing, and a length-changing block's reconstruction differs in length from its span, which `is_inversion` refuses -- so the `none` above covers both call sites, not just the coalesce.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -4762,10 +4762,16 @@ fn coalesce_whole_block_inversion(pieces: &mut Vec<Piece>, ref_bytes: &[u8]) {
 /// `a_dense_inversion_is_recognised_across_multi_base_separations`, which the
 /// first revision of that pass broke).
 fn whole_block_inversion(pieces: &[Piece], ref_bytes: &[u8]) -> Option<Piece> {
+    // Written in the order the four routes were added, which is the order their
+    // own docs number themselves ("the second admission route", "the third
+    // alternative", "the fourth admission route"). All four are pure, so the
+    // order is not observable — but a chain that disagrees with the ordinals
+    // makes every one of those doc lines read as stale.
     if pieces.len() < 2
         || !(every_separation_is_a_single_base(pieces)
             || changed_columns_dominate_the_span(pieces)
-            || no_piece_is_a_lone_substitution(pieces))
+            || no_piece_is_a_lone_substitution(pieces)
+            || payload_columns_dominate_the_span(pieces))
     {
         return None;
     }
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

Issue #1627, opened 2026-08-11
normalize: a member ferro cannot compute is re-emitted verbatim, so 48 rows produce non-conformant output in every mode


nh13:
Rule 1 of the project's normalization ruleset (`README.md`) is *"Output follows the HGVS recommendations. **Absolute — never traded.**"* It is about **output**, so it has no mode escape. Ferro currently violates it for 48 corpus rows, in all three error modes, and says nothing about any of them.

## The mechanism

A member ferro cannot compute is passed through **untouched** rather than refused. Two shapes reach it:

| input | `to_spdi` | `normalize`, all three modes |
|---|---|---|
| `NC_TEST.1:g.10_11ins6` | `Err(… "insertion sequence is not a literal sequence")` | `Ok("NC_TEST.1:g.10_11ins6")` |
| `NC_TEST.1:g.10delinsX` | `Err(… "non-literal inserts cannot be encoded as SPDI")` | `Ok("NC_TEST.1:g.10delinsX")` |

`to_spdi` already knows these denote nothing and refuses them **by name**. The normalizer does not consult that signal; it treats the member as opaque and emits it back, so the output carries a spelling `checklist.md:33` and `background/standards.md:39` prohibit.

## Measured

Over the 164 `RowKind::Prohibited` rows of `conformance::spec_corpus`, both shuffle directions, all three `ErrorMode`s:

- **24 rows** `checklist.md:33` (`ins6`) + **24 rows** `standards.md:39` (`delinsX`) = **48**, all accepted, all re-emitting the prohibited token.
- Output is **byte-identical across strict, lenient and silent** — 0 disagreements over 96 row-direction pairs.
- `normalize_with_diagnostics` returns an **empty** warning vector for every one of them. This is silence, not leniency.
- `denotation_of` reports `Denotation::Inexpressible` for all 48, so the corpus's independent applier agrees with `to_spdi`.

## Why this is separable from the enforcement-stage ruling

`rulings[absolute-prohibition-enforcement-stage]` decides where the *input* is judged, and it permits lenient not to judge it. It does **not** permit a non-conformant *output* in any mode. Even after that ruling is implemented, lenient must still not emit `ins6` — it must fail, because it cannot normalize. Today it neither judges nor fails: it launders the input into the output position and calls it normalized.

The shape of the fix follows: when a member has no computable denotation, normalization fails rather than passing the member through. The refusal already exists one module over; it needs to be reached.

## Adjudication

[...]

PR #1641, opened 2026-08-11
fix(spdi): refuse a genomic offset instead of dropping it

### LINKED ISSUE #1628: to_spdi discards a g. offset while normalize preserves it — non-conformant output and a confluence failure between two halves of ferro
Two halves of one defect, both measured on `conformance::spec_corpus`'s genomic frame.

## 1. `to_spdi` silently discards a `g.` offset

`checklist.md:16` says a genomic reference "can not have nucleotides with additions like a `+`, `-`, or `*`". Ferro parses `+` and `-` anyway, and then drops them:

```
to_spdi("NC_TEST.1:g.266+2del")    -> NC_TEST.1:265:A:
to_spdi("NC_TEST.1:g.266-268del")  -> NC_TEST.1:265:A:
to_spdi("NC_TEST.1:g.266del")      -> NC_TEST.1:265:A:
```

All three produce the **same triple**. The `+2` and the `-268` are not rejected, not warned about, not honoured — they are dropped. The AST does carry them (`GenomePos { base: 266, offset: Some(2) }`), so the loss is in the conversion, not in the parse.

## 2. Which makes it a confluence failure, not only a conformance one

`normalize` keeps the offset:

```
normalize("NC_TEST.1:g.266+2del")   -> NC_TEST.1:g.266+2del      (fixed point)
normalize("NC_TEST.1:g.266-268del") -> NC_TEST.1:g.266-268del    (fixed point)
normalize("NC_TEST.1:g.266del")     -> NC_TEST.1:g.267del        (3' shifted)
```

So three descriptions ferro's own SPDI path calls **identical** normalize to **three different strings**. Rule 3 of the README ruleset — *"Inputs denoting one variant produce one output"* — fails, and it fails between two halves of ferro rather than against an external tool.

Whichever way this is resolved, the two halves have to agree. Either the offset is meaningful on `g.` (it is not — `checklist.md:16`) or the description is refused, in which case `to_spdi` must refuse it too rather than quietly answering for a different variant.

## Measured

8 corpus rows: 4 × `checklist.md:16` (`g.266+2del`) and 4 × `checklist.md:45` (`g.266-268del`), each alone and inside a cis allele. Accepted in **all three** `ErrorMode`s, re-emitted with the offset intact, and `normalize_with_diagnostics` returns **no warning** for any of them in any mode.

These 8 are also 8 of the 32 `prohibition_violating_outputs` the axis census pins, so the output half is already counted — see `tests/it/corpus_prohibited_inputs.rs::one_clause_names_three_offset_symbols_and_ferro_refuses_one`, which records that `g.*10del` is refused at parse while `+` and `-` are not, from the same one-sentence clause.

## Relationship to the enforcement-stage ruling

`rulings[absolute-prohibition-enforcement-stage]` (decided) puts the *input* check in strict mode. That covers half of this: strict would refuse both spellings. It does not cover the `to_spdi` disagreement, which is wrong in every mode and is the part that costs confluence.

## Representation impact

The 8 rows' outputs move (they stop being emitted, or they move to the offset-free spelling). No input that currently normalizes to a legal description is affected.

nh13:
Closes #1628

`hgvs_to_spdi` read `GenomePos::base` and never looked at `GenomePos::offset`. Measured on `main` before this change, by a test written to fail first:

```
`NC_000001.11:g.10+2delC` converted without a provider to
  SpdiVariant { sequence: "NC_000001.11", position: 9, deletion: "C", insertion: "" }
```

and the invariant test failed with `left: Some("NC_000001.11:9:C:")`, `right: Some("NC_000001.11:9:C:")` for `g.10+2delC` against `g.10delC`. Three descriptions, one triple. `normalize` meanwhile keeps them as three distinct strings — the offset-carrying ones are fixed points, the plain one 3'-shifts — so this is a confluence failure *between two halves of ferro*, not a conformance gap against an external tool.

## What was chosen, and why

Three behaviours were available: refuse, honour the offset by resolving it, or leave the refusal to a stage further up and let `to_spdi` stay total. **Refuse**, and only on `g.`/`m.`/`o.`

- Honouring it is not available. Resolving a `+`/`-` offset needs exon boundaries to measure it against, and a genomic accession has none. There is nothing to resolve.
- Leaving it upstream does not cover this. The decided `absolute-prohibition-enforcement-stage` ruling puts the *input* check in strict mode, which is #1627's territory and a different bug. In lenient and permissive modes ferro still accepts the description, and `to_spdi` must not answer for a different variant in those modes. The disagreement is wrong in all three.
- Refusing is what the rest of the module already does. `resolve_cds_to_tx`, `resolve_tx_pos`, `resolve_rna_pos`, `require_simple_tx_pos` and `require_simple_rna_pos` each decline an intronic `c.`/`n.`/`r.` position, because SPDI has no offset notation. The genomic arms were the one hole in an otherwise consistent policy. They decline because the offset cannot be projected onto the transcript accession; this one declines because there is nothing on a genomic accession to project it against — hence `InvalidPosition` rather than their `MissingReferenceData`, and the difference in verdict is asserted.

## `c.`/`n.`/`r.` are provably unaffected

`reject_genomic_offset` takes an `Interval<GenomePos>`, a type only the six `g.`/`m.`/`o.` arms hold — the transcript axes carry `TxPos`/`RnaPos`/`CdsPos` and cannot reach it. Measured as well as argued: `a_transcript_axis_offset_declines_for_its_own_reason` pins `c.10+5`, `n.10+5` and `r.10+5a>g` declining as `MissingReferenceData` with their own "without genomic projection" wording, *and* pins their exonic siblings still converting. That test was run with the six guard calls stripped out and passes identically, so it is a genuine before/after control rather than a new expectation.

## The #1615/#1622 oracle was blind to this, measured

Not assumed — measured on #1622's head (`ad367337`) in a throwaway worktree, calling `compare_denoted_sequences` directly:

| input | output | before | after |
|---|---|---|---|
| `NC_KEY.1:g.10+2delG` | `NC_KEY.1:g.10delG` | **`Agree`** | `NotComparable(InputDenotesNoSequence)` |
| `NC_KEY.1:g.10-268delG` | `NC_KEY.1:g.10delG` | **`Agree`** | `NotComparable(InputDenotesNoSequence)` |

The oracle reaches bases through `variant_edit_triples_reason` → `hgvs_to_spdi`, so it inherited the drop wholesale and would have certified a normalization that moved an offset-carrying description onto a different one. It already guards `pter`/`qter`/`cen`, which fail the same way (`base: 0`); offsets were the sibling case nobody had reached.

Fire counts on that branch are unchanged: `FERRO_ASSERT_SEQUENCE=1 cargo nextest run --features dev` gives **26 failures and 6 `changed the sequence` fires both with and without this patch, and the failing sets are byte-identical**. Nothing in the suite currently feeds an offset-carrying genomic description through the oracle, which is exactly why the blindness was invisible.

## Relationship to #1618 and #1619

Same family, three different mechanisms — no shared fix, so nothing here is expanded to cover them:

- **#1628** (this one) — the conversion *discards* a field the AST carries.
- **#1618** — the conversion and the normalizer resolve an anchored repeat's tract to different spans. That is an unmade policy decision made twice, not a lossy read; it needs a ruling on what a copy count is counted against.
- **#1619** — the conversion walks exon gaps while the sequence it indexes is flat. A coordinate-frame bug, possibly in the fixture rather than the resolver.

The common thread worth naming is structural: `hgvs_to_spdi` is a second, independent interpreter of an HGVS description, written separately from the normalizer's, and nothing pins the two to agree. #1615's oracle is the instrument for that; each of these three is a separate repair.

## Verification

- `cargo nextest run --features dev` — **10,008 passed, 150 skipped, nothing re-blessed**. For a change that alters what `to_spdi` answers, an untouched expectation set is the strong signal: every existing expectation already agreed.
- `cargo test --doc --features dev` — 92 passed.
- `cargo fmt --check` clean; `cargo clippy --all-features --all-targets -- -D warnings` clean.
- `dump_normalized_corpus --compare` over 78,298 rows: **0 moved**. Reported with its blindness, per the harness's own rule: that corpus dumps *normalized* output and no family builds an offset-carrying genomic position, so it is structurally incapable of varying the property under test. It is not the evidence here — the suite is.

Representation-Change: none. `normalize`'s output path never calls `hgvs_to_spdi` (the only mentions in `src/normalize/` are doc cross-references), and no producer of `GenomePos` other than the parser ever sets `offset` — projection builds every genomic position with `GenomePos::new`. Verified, not asserted: 10,008 tests pass with nothing re-blessed, and 0 of 78,298 corpus rows move. What changes is `to_spdi`'s answer for descriptions that are prohibited input in the first place, from a wrong triple to an error.

### DIFF
--- a/src/spdi/convert.rs
+++ b/src/spdi/convert.rs
@@ -224,6 +224,49 @@ fn get_end_pos(interval: &Interval<GenomePos>) -> Option<u64> {
     interval.end.inner().map(|p| p.base)
 }
 
+/// Refuse a genomic interval whose start or end carries a `+`/`-` offset (#1628).
+///
+/// `GenomePos` can hold an offset because the parser accepts one, but neither
+/// half of the conversion can honour it: a genomic accession has no exon
+/// structure to measure an offset against — `checklist.md:16` prohibits an
+/// offset on a genomic position outright — and SPDI is purely positional, with
+/// no offset notation.
+///
+/// What the conversion did instead was drop it, which is the one answer that is
+/// worse than either honest option. `g.266+2del`, `g.266-268del` and `g.266del`
+/// all flattened onto the same triple while `normalize` keeps them as three
+/// distinct descriptions, so ferro's two halves disagreed about what a variant
+/// is: descriptions the SPDI path called identical normalized to three
+/// different strings.
+///
+/// Refusing is also what the other axes already do with the offsets that *are*
+/// legitimate there — see [`resolve_cds_to_tx`], [`resolve_tx_pos`] and
+/// [`require_simple_tx_pos`], each of which declines an intronic `c.`/`n.`/`r.`
+/// position for want of an SPDI representation. Those decline because the
+/// offset cannot be projected onto the transcript accession; this one declines
+/// because there is nothing on a genomic accession to project it against.
+fn reject_genomic_offset(
+    interval: &Interval<GenomePos>,
+    coord: &str,
+) -> Result<(), ConversionError> {
+    for endpoint in [interval.start.inner(), interval.end.inner()]
+        .into_iter()
+        .flatten()
+    {
+        if endpoint.offset.is_some() {
+            return Err(ConversionError::InvalidPosition {
+                description: format!(
+                    "{coord}. position {endpoint} carries a +/- offset: a genomic position \
+                     cannot have one and SPDI has no offset notation to express it. Drop the \
+                     offset, or describe the variant on a transcript accession where an offset \
+                     is meaningful"
+                ),
+            });
+        }
+    }
+    Ok(())
+}
+
 /// Convert an HGVS variant to SPDI format without consulting a reference provider.
 ///
 /// This is the "simple" conversion path: the SPDI is emitted on the same
@@ -233,7 +276,7 @@ fn get_end_pos(interval: &Interval<GenomePos>) -> Option<u64> {
 ///
 /// | HGVS coord | Supported here | Notes |
 /// |------------|----------------|-------|
-/// | `g.` (genomic) | yes | direct 1→0-based conversion |
+/// | `g.` (genomic) | yes | direct 1→0-based conversion; a `+`/`-` offset is refused |
 /// | `m.` (mito) | yes | mito accession is genomic; same as `g.`; wraparound rejected |
 /// | `o.` (circular) | yes | same path as `g.`/`m.`; wraparound rejected |
 /// | `n.` (non-coding tx) | exonic, positive base | SPDI sits on the transcript accession |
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1644, opened 2026-08-11
fix(normalize): merge a split whose boundaries the alignment manufactured

### LINKED ISSUE #1638: normalize: the canonical partition cuts at matches its own alignment manufactured with compensating gaps
`partition_block_canonical` picks a member-count-minimal minimal alignment. Because it is free to place gaps anywhere, it will shift the alignment and shift it back to make reference bases coincide with payload bases, and it then cuts at those coincidences. The result splits blocks that `partition_block` — restricted to a single gap, deliberately, "because letting it place compensating gaps lets it manufacture matches from coincidence and shred a genuinely contiguous replacement" — keeps whole.

Measured with `examples/dump_partitions --min-separation 1` over the committed 65-row reproducer corpus. `canonical` differs from `live` on 8 blocks and every difference is an over-split:

| row | ref -> alt | live | canonical |
|---|---|---|---|
| `#1034-guard-3nt` | `GCT -> AGC` | `0:3/AGC` | `0:0/A\|2:3/` |
| `#1041` | `GCA -> TGC` | 1 | 2 |
| `#1040` | `CAGTGACTAG -> TGTCACGACT` | 1 | `0:2/T\|4:5/C\|7:8/G\|9:10/CT` |
| `#160-1099` | `TGCTC -> AAGCT` | 1 | `0:1/AA\|4:5/` |
| `#422-spanning` | `CTATAG -> AAACCCC` | 1 | 2 |
| `LRG_199` (`delins.md:44`) | 52 nt -> 14 nt | 1 | `0:17/\|20:31/\|37:43/\|45:47/C\|49:52/` |
| `LRG_199-standin` | 52 nt -> 14 nt | 1 | 14 |
| `long-delins-40nt` | 40 nt -> 40 nt | 2 | 21 |

A 21-member description of a 40 nt delins and a 5-member description of `delins.md:44-47`'s worked example are both non-conformant against the passage the spec works through in full — `:47` recommends the spanning `delins` and calls the split the "alternative description".

### The property that separates them

Walk the members and track the cumulative reference-to-alternate offset. Every block above where `canonical` over-splits is **non-monotone** — the alignment both inserts and deletes, so the matched runs between the members are matched only because the alignment shifted and shifted back. Every block the corpus requires to *split* is monotone:

* non-monotone, must merge — `#1034-guard-3nt` (`+1` then `-1`), `#1041`, `#1040` (`-1 ... 0`), `#160-1099`, `long-delins-40nt`, and `NM_004006.2:c.4145_4160inv`.
* monotone, must split — `#182-a..e`, `#1230`, `#1231`, `#1233`, `#1235-n`, `#1241`, `#1157-A-*`, `#1232`, `#1040-two-invs`, `#1260a/b`, `#1260-window`, `#1000`, `#999-neg`, `#1262b`, `#422`.

There are no counterexamples in the 65 rows. On its own the property is not sufficient: a genuinely separated `10_12del` plus `20_22insACC` allele is also non-monotone, and must stay two members. That is separated by density — the columns the members claim must be at least the unchanged bases lying between them.

`LRG_199` and its stand-in are **monotone** (every member is a deletion) so the rule above does not reach them. They are the distinct `delins.md:44-47` payload-alignment shape that `coalesce_payload_alignment_split` already states, which today runs only under `FERRO_PARTITION=canonical-coalesced`.

nh13:
`partition_block_canonical` picks a member-count-minimal *minimal* alignment and is free to place gaps anywhere. So it will shift the alignment and shift it back to make reference bases coincide with payload bases, then cut at those coincidences. `partition_block` cannot: its single-gap search exists precisely *"because letting it place compensating gaps lets it manufacture matches from coincidence and shred a genuinely contiguous replacement (#1034/#1040/#182)"* — the argument `canonicalize_from_sequence` already makes for that restriction. This states that same argument on the derived pieces instead of on the search.

Closes #1638

## The measurement

`examples/dump_partitions --min-separation 1` over the committed 65-row reproducer corpus. `canonical` differs from `live` on **8** blocks and every difference is an over-split:

| row | ref -> alt | live | canonical |
|---|---|---|---|
| `#1040` | `CAGTGACTAG -> TGTCACGACT` | 1 | `0:2/T\|4:5/C\|7:8/G\|9:10/CT` |
| `#160-1099` | `TGCTC -> AAGCT` | 1 | `0:1/AA\|4:5/` |
| `#1034-guard-3nt` | `GCT -> AGC` | 1 | `0:0/A\|2:3/` |
| `LRG_199` (`delins.md:44`) | 52 nt -> 14 nt | 1 | 5 members |
| `long-delins-40nt` | 40 nt -> 40 nt | 2 | **21** members |
| … and 3 more | | | |

## The rule, and why it carries no threshold

Walk the members and track the **cumulative reference-to-alternate offset**. Every block `canonical` over-splits is *non-monotone* — the alignment both inserted and deleted, so the matched runs between the members are matched **only at that shift**; align the block without the compensating gap and they are not matches at all.

Every block the corpus requires to **split** is monotone. The four rows that matter most are `#1260a`, `#1260b`, `#1260-window` and `#1000` — pairs of insertions one base apart, which look superficially exactly like the merge shape. Their offsets are `+1, +2`: strictly increasing, never returning, so the base between them is genuinely unchanged reference and `general.md:34` keeps them individual. **There are no counterexamples in the 65 rows.**

That matters because the live path's `MAX_SINGLE_BASE_SEPARATION_CHANGE` records that its own value is *"under-determined"*, constrained only to `[2, 15)`, and that *"no threshold at all satisfies #1157 and #1232 simultaneously"*. Porting a number from there would import that unresolved calibration; this asks a different question.

## Two further conditions, each forced by a case rather than chosen

**Density.** Non-monotonicity alone over-merges. A `10_12del` plus a `20_22insACC` is also non-monotone — `-3` then `0` — and must stay two members. So the columns the members claim must be at least the unchanged bases between them: `6 >= 10` is false there; `#1040`'s `6 >= 5` holds.

**A three-member floor** (`MIN_MANUFACTURED_SPLIT_MEMBERS`). At two members there is exactly one gap, and one gap between a growing member and a shrinking one is equally the signature of a genuine indel pair — which this repo has adjudicated in **both directions** at that arity with nothing in the sequence separating them. Measured: without the floor this merges `NR_TEST.1:n.[8del;10_11insA]` and `NC_TEST.1:g.[261del;263_264insA]`, both of which `general.md:34` keeps individual and both pinned by name. `GCT -> AGC` looks like a counterexample and is not — it is merged by `coalesce_whole_block_inversion` on the strength of `inversion.md:5`, a sequence relation over the whole span that this rule cannot see.

## The failing tests, observed before the fix

With the pass stubbed to return immediately:

```
a_manufactured_split_is_merged_and_a_genuine_one_is_not
  #1040: a split resting on manufactured matches must be one member
  left:  [0:2/T, 4:5/C, 7:8/G, 9:10/CT]
  right: [0:10/TGTCACGACT]

a_separated_deletion_and_insertion_are_not_merged
  left: 2, right: 1
```

The first also caught a real bug in the first implementation: it read the sign of the *running total* rather than of each **step**, and `#1040`'s totals are `-1, -1, -1, 0` — never positive, even though the alignment plainly deleted and then inserted back.

## Scoping, and why it is a measurement boundary rather than a belief

The pass is called only on the sequence-first arms. **Unscoped it does fire on `live`** — measured, once, on a 164 nt `n.` block that `partition_block` splits into eighteen members and this merges into one spanning `delins`. That may well be the better description; it is not one to ship as a side effect of fixing a different arm, and its blast radius is a separate question. Remove the scoping when `live` goes.

## Measurements

Full suite, 9,991 tests, against a **0-failure** baseline:

| configuration | distinct failures |
|---|---|
| `FERRO_PARTITION=canonical`, before | 78 |
| `FERRO_PARTITION=canonical`, after | **71** |
| shipped `live` arm, after | **0** of 9,994 |

9 repaired, including `long_delins_splits_at_unchanged_bases::a_delins_longer_than_the_old_cap_still_splits` (the 40 nt row) and `inversion_sweep::the_spec_published_inversion_examples_produce_the_spec_form`. 2 added, both generated-artifact replay detectors that are regenerated from the code under test.

## What this does NOT fix

`delins.md:44-47`'s own worked example — `LRG_199t1:c.850_901delinsTTCCTCGATGCCTG`, 52 nt to 14 nt — is **monotone**: every canonical member is a deletion, so the alignment never shifted back and this rule cannot see it. It is the distinct payload-alignment shape `coalesce_payload_alignment_split` already states, which today runs only under `FERRO_PARTITION=canonical-coalesced`. Reported rather than papered over; it is the remaining bucket-1 row for the flip.

Representation-Change: none. 0 rows move on the shipped `live` arm — the full suite under the default arm is 9,994 passing with nothing re-blessed, and the new pass is called only when `matches!(partition_rule(), PartitionRule::Canonical | PartitionRule::CanonicalCoalesced)`, which no shipped configuration selects. The 9 rows it moves are all under `FERRO_PARTITION=canonical`, an unstable evaluation switch.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2734,6 +2734,46 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
     //
     // Pinned by `issue_1040_inv_overrecognition_probes::
     // a_derived_whole_block_inversion_outranks_the_codon_exception`.
+    //
+    // `coalesce_compensating_gap_split` runs first: a split whose boundaries the
+    // alignment manufactured is put back together as one span, which then
+    // arrives below as a single piece and is typed by
+    // `crate::normalize::rules` rather than needing the inversion gate at all.
+    // It is placed here, after the weight bound, for the same reason the call
+    // below is — it widens.
+    //
+    // Scoped to the two `partition_block_canonical` arms, and the scoping is a
+    // **measurement boundary, not a belief**: the defect is that partitioner's,
+    // and whether the same rule is right for `partition_block` is a question
+    // with its own blast radius that this change does not answer. Left unscoped
+    // it does fire on the live arm — measured, once, on a 164 nt `n.` block that
+    // `partition_block` splits into eighteen members and this merges into one
+    // spanning `delins`. That may well be the better description; it is not one
+    // to ship as a side effect of fixing a different arm, and the shipped output
+    // must not move in a pre-flip change. Remove the scoping when `live` goes.
+    //
+    // Written as a positive match over the arms it argues for, like the call
+    // above, and **not** as `!= Live`. That negation also selects
+    // [`PartitionRule::Shadow`], whose pieces come from
+    // `partition_block_sequence_first` — boundaries taken from the alignment
+    // steps common to *every* minimal alignment, which is precisely the set no
+    // alignment can have manufactured. Applying a "your boundaries are a shifted
+    // coincidence" rule there contradicts its own premise, and nothing below
+    // measures or pins that arm.
+    //
+    // The limit this scoping does **not** reach, stated rather than implied:
+    // every non-`Live` arm falls back to `partition_block` when its own splitter
+    // declines, so on a `canonical` bake-off run a declining block still reaches
+    // this rule with live pieces. The gate keys on the arm, not on which
+    // partitioner produced the pieces. Shipped output is unaffected either way —
+    // `FERRO_PARTITION` unset is `Live` — but a bake-off column can carry a
+    // widening the arm it is labelled with did not cause.
+    if matches!(
+        partition_rule(),
+        PartitionRule::Canonical | PartitionRule::CanonicalCoalesced
+    ) {
+        coalesce_compensating_gap_split(&mut pieces, &ref_bytes);
+    }
     coalesce_whole_block_inversion(&mut pieces, &ref_bytes);
 
     // Applied *after* the weight bound, deliberately. The bound is a statement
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1645, opened 2026-08-11
chore(adjudication): correct two stale claims and pin what two respell tests denote

nh13:
Four small bookkeeping fixes salvaged from an unpushed branch, rebased onto current `main`. Three are documentation corrections; one adds assertions that were missing.

## The commits

**`docs(normalize): correct a stale "last gate that reads input spelling" claim`** — a comment in `merge.rs` said a removed veto "was the last gate in this pass that read the *input spelling* rather than the sequence". That is no longer true; other spelling-readers remain in the pass. Reworded to *"at the time it was removed, it was the last gate…"*, which is a historical statement rather than a present-tense one that has since become false.

**`test(1307): assert sequence preservation on the remaining respell tests`** — the tests in `issue_1307_terminal_dup_respell.rs` pinned only the output string, so they could not distinguish a legitimate re-spelling from a corruption: a change that emitted a description denoting different bases would stay green as long as the string matched. Adds denoted-sequence assertions alongside the existing string ones. Same class of gap as #1626/#1633, on a file that PR did not cover.

**`docs(msto): fix a stale spec-authority split left over from #1421's reclassification`** — the doc split rows between two authorities in a way that no longer matched how #1421 classified them.

**`test(msto): split the reported-pair census constant per direction`** — the census pinned one constant across both shuffle directions, so a change in one direction could be masked by a compensating change in the other. Split into a 3′ and a 5′ constant.

## Verification

- Full suite: **10,013 passed, 0 failed, 150 skipped — nothing re-blessed.** For a branch rebased across 21 commits of `main`, an unchanged expectation set is the signal worth stating: the per-direction census constants introduced by the last commit still hold at the current tip.
- `cargo fmt --all --check` clean.
- One test failed before an unrelated fix: `ruling_records_are_intact` compared a 10-record pinned list against `main`'s 16. The test file on this branch is byte-identical to `main`'s — the cause was a stale gitignored generated fixture in the worktree (`tests/fixtures/grammar/hgvs_spec_normalization.json`, two days old). Regenerating resolved it. Recorded here because the failure reads like a real ledger regression and is not one.

Representation-Change: none. Three commits touch documentation or tests only; the `merge.rs` change is comment-only, verified by the unchanged expectation set above.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2756,12 +2756,13 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
     // which `crate::normalize::rules`' single-span typing renders as `inv` —
     // so there is nothing left for a veto to protect.
 
-    // There was a veto here, and it was the last gate in this pass that read the
-    // *input spelling* rather than the sequence: when the derivation produced
-    // fewer pieces than there were members, and any piece covered a base the
-    // input had left between two of them, the pass refused and returned the
-    // input verbatim. It cited `general.md:34` — two variants separated by one
-    // or more unchanged nucleotides are described individually.
+    // There was a veto here, and — at the time it was removed — it was the
+    // last gate in this pass that read the *input spelling* rather than the
+    // sequence: when the derivation produced fewer pieces than there were
+    // members, and any piece covered a base the input had left between two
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

PR #1646, opened 2026-08-11
test(cis): census the n. and r. confluence axes, and pin what #1484 did there

nh13:
## Measured suite cost of the new module

Raised by the CLI lens, which asked for the wall-clock delta and characterised the module as "roughly doubling the cost of this shape family". Measured rather than estimated, and the answer splits by which cost you mean.

Per-test, `cargo nextest run --features dev -E 'test(cis_confluence_nr_axis) + test(cis_confluence_axis) + binary(generate_cis_confluence_corpus)'` — **rc=0, 19 passed**:

| test | time |
|---|---|
| `cis_confluence_nr_axis::noncoding_three_prime_confluence_census` | 6.951s |
| `cis_confluence_nr_axis::noncoding_five_prime_confluence_census` | 7.317s |
| `cis_confluence_nr_axis::rna_three_prime_confluence_census` | 8.231s |
| `cis_confluence_nr_axis::rna_five_prime_confluence_census` | 8.795s |
| *(this module's two fast guards)* | 0.103s, 0.140s |
| `cis_confluence_axis::three_prime_confluence_census` (pre-existing) | 13.302s |
| `cis_confluence_axis::five_prime_confluence_census` (pre-existing) | 14.278s |

**In serial/CPU terms the characterisation is about right**: this module contributes ~31.3s against the pre-existing sibling's ~27.6s, so it roughly doubles the shape family's total work — four censuses where there were two, each cheaper because the `n.`/`r.` corpus is smaller per axis.

**In wall-clock terms it is close to free on a parallel runner.** The whole 19-test filter completed in **14.283s**, which is the slowest single test (14.278s) — nextest ran the four new censuses alongside the two existing ones. So the module adds ~0s of wall clock whenever cores are free, and up to ~31s when they are not. Which case CI lands in depends on shard packing, and no per-shard measurement is claimed here.

**Not measured:** the before/after time of the *whole* suite. Quoting one would mean quoting a figure I did not take, and this repo's own guidance is that a suite-time claim is worth only what it was measured on. The per-test figures above are the ones to reason from, and none of the four new censuses is near nextest's 60s SLOW threshold.

## What this adds

The cis-confluence corpus and its census only ever covered the `g.` axis. This extends the generator to the `n.` and `r.` axes and pins a census for each, in both shuffle directions, with a bake-off across the four `FERRO_PARTITION` arms.

## The finding

Blessing those four censuses for the first time is itself a measurement, and it is a large one. Against the figures the axis produced when the corpus was first generated:

| census | before | now | Δ |
|---|---:|---:|---:|
| `n.` 3′ converged | 2,628 | **4,012** | +1,384 |
| `r.` 3′ converged | 2,627 | **4,010** | +1,383 |
| `n.` 5′ converged | 2,625 | **4,011** | +1,386 |
| `r.` 5′ converged | 2,624 | **4,009** | +1,385 |

`split_two` falls from ~2,941–2,954 to ~1,558–1,571 and `split_three` drops by one or two, which is the same movement seen from the other side. Crucially `classes` (5,636) and `spellings` (23,696) are **unchanged**, and `sequence_changed` is **0** in all four — so this is the same corpus being normalized differently, not a different corpus.

That the two directions moved by almost exactly the same margin (+1,383 and +1,385 on the `r.` axis) is the evidence that this is a property of the normalizer rather than of one shuffle direction.

The mover is #1484, which changed how a cis allele is modelled — as a partition of the reference and result blocks — and which had no test covering the `n.` or `r.` axes. This file is that coverage, and these constants are the first measurement of what that change did here. Every figure moved in the direction the census assertion demands.

## Why `Representation-Change: none`

The strings did move, but they moved when #1484 merged, not here. This PR only makes that movement visible and pins it. Declaring the change is #1484's obligation, not this PR's; recording it as a change here would double-count it in the changelog.

## Verification

- Full suite on the rebased tip: **10,028 passed, 0 failed, 150 skipped**.
- Rebased onto `main` after #1598 landed, which touches `merge.rs`; the censuses were re-verified with it in and did not move again.
- `sequence_changed == 0` across all four censuses — no output denotes different bases than its input.

Representation-Change: none. This PR changes no normalizer behaviour — it adds test coverage and records, as pinned constants, what an already-merged PR did to two axes that had none.


nh13:
@coderabbitai pause

PR #1648, opened 2026-08-11
test(rulings): two adjudication records, and make four census ratchets state their trap

nh13:
Two adjudication records and four census ratchets that now state the trap they can be moved by. No source file is touched — this is entirely `tests/` and the ruling ledger.

## Unstacked from #1649 during the rebase — read this first

This branch was sitting on top of #1649 (`fix/normalize: let the splitter express two deletions`), so its diff against `main` carried #1649's `src/normalize/merge.rs` change and four of #1649's commits. That was accidental — the PR description already claimed "no source file is touched", which was false of the diff.

Rebasing dropped those four commits, so this PR is now its own six commits and nothing else. Three of the five files that were conflicting (`cis_confluence_axis.rs`, `defect_non_idempotent_outputs.rs`, `spec_conformance_axis.rs`) belonged entirely to #1649 and are no longer touched here at all. **#1649 is unaffected and still open on its own.**

The consequence worth reviewing: commit `test(rulings): apply the decided chain to the 1419 rows` originally moved two different kinds of field at once. The `wanted`/`argument` fields are an **adjudication** and are kept. The `output`/`five_prime` fields are **measurements** of #1649's behaviour and are not — every one is back to what `main` measures, re-measured on this tree rather than composed from two PRs' deltas. The verdict flip `Canonical -> Gap` on the three `1419-r*/span` rows survives that, because it follows from `wanted` moving, not from any output moving.

## What is here

**Two new ruling records**, both pinned in `RULING_STATUSES`:

* `contiguous-insertion-split-by-a-blocked-derivation` — **decided**. Inside a tandem tract, one contiguous 3 nt insertion can be spelled as an `ins` at one junction plus a `dup` inserting at the next, presenting an unchanged base between two members that the variant does not have. `general.md:34` is stated over "two variants" and there is only one here, so it does not license the split. Decided **as to that scope only** — the direction comes from `canonical-form-choice-when-both-legal`, not from this record. Ferro emits both forms as fixed points today; both are pinned, together with the proof that they denote one sequence, by `the_three_member_spelling_and_its_one_member_form_are_two_fixed_points`.
* `delins-recommendation-reach-when-the-input-arrives-split` — **undecided**, deliberately. The provenance half of the question is *not* open; it is excluded by `canonical-form-choice-when-both-legal`. What is open is which of `delins.md:47` and `general.md:34` a re-derivation lands on when the minimal edit set really does have two members separated by unchanged bases. #1420's two `SpecExplicit` gap rows turn on it.

**A stale census withdrawn rather than restated.** The `separation-is-a-property-of-the-spelling-not-of-the-variant` rationale quoted 4,643 diverging classes with 4,568 (98.4%) cross-partition. Those figures were a second, unasserted copy of a count that has an asserted one (`cis_confluence_axis::{THREE_PRIME,FIVE_PRIME}`), and the instrument no longer emits a `cross-partition` share at all, so the 98.4% cannot be re-measured. The paragraph now says so and points at the asserted constants. The record's argument is unaffected and it stays `decided`.

**Four "a rise is progress" traps made explicit.** `CONVERGING_PAIRS_THREE_PRIME`, `CONVERGING_PAIRS_FIVE_PRIME`, `OPEN_GAPS` and the `1420-v2` forbidden string can all be moved in the good-looking direction by a re-derivation across bases the input left unchanged — a defect wearing the shape of a fix. This has been observed on an experimental partitioner arm. None of them said so; now they do.

(The "make `CONVERGING_PAIRS` two per-direction constants" half of this commit is **dropped as superseded** — #1645 landed exactly that split on `main` first, with a better-argued rationale. What is left is the trap documentation and the strengthened failure message.)

## Measured figures, all re-measured on this base

The figures this description previously carried were measured against the old base and every one of them was wrong here. Re-measured:

| pin | main | this PR |
|---|---|---|
| ledger records / decided / undecided | 16 / 13 / 3 | **18 / 14 / 4** |
| clause lines | 62 | **62** (unchanged) |
| clause lines named by >1 record | 13 | **15** |
| mixed-verdict clause lines | 2 | **7** |
| `OPEN_GAPS` | 9 | **12** |
| `CONVERGING_PAIRS_THREE_PRIME` / `_FIVE_PRIME` | 0 / 0 | **0 / 0** (unchanged) |

`OPEN_GAPS` 9 -> 12 is the one pin that moves in its forbidden direction, and it is a **correction of the target, not a regression of the output**: #1419's six `wanted` fields rested on `general.md:56`, which `adjudication-precedence-order` records as refuted for merge-versus-split, so the three `/span` rows that were `Canonical` against the withdrawn target are `Gap` against the standing one. **No `output` or `five_prime` field moves at all** — that is checkable from the diff, not just asserted here.

The mixed-verdict count rising 2 -> 7 is likewise not a regression: one new `undecided` record cites five lines that previously carried only decided records. `clause_ruling_index` gained a note that this count can rise by a record being *filed*, not only fall by a record being *decided* — the burn-down framing there described only the decided half.

**No `decided` record is regressed to `undecided`.** The branch's stale index block had `codon-carve-out-shape-restriction` and `exon-junction-dup-converge-from-the-far-side` as `undecided`; both were decided on `main` by #1623. The whole generated block was regenerated from this tree's ledger rather than merged, and `RULING_STATUSES` only gains two rows.

## Review findings

Eleven CodeRabbit findings. Three were fixed, one was a real gap the restructure did not touch, and the rest were either scoped to #1649's files or rested on text the restructure removed. Details in the review thread.

The one worth calling out: `every_gap_row_is_returned_exactly_as_authored` exempted whole *pairs* from the retention pin, which silently dropped it from the three `1419-r*/cis` rows that **do** still retain their input. It now pins each row's measured `output` instead, which holds for the movers and the retainers alike.

## Verification

`cargo fmt --all -- --check` and `cargo clippy --all-features --all-targets -- -D warnings`: clean.

`cargo nextest run --features dev --no-fail-fast`: **10138 run, 10138 passed, 152 skipped**.

Oracle-armed exactly as `test-oracle` runs it (`FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1`, with `ORACLE_EXCLUDE` negated): **10076 run, 10076 passed, 214 skipped**. The oracles were deliberately *not* armed over the whole suite — that reproduces a configuration CI never runs, and `defect_non_idempotent_outputs` and `spec_conformance_axis` are both in `ORACLE_EXCLUDE`, so they run unarmed in the plain `test` job.

**Each of the four ratchets was mutation-tested rather than assumed to bite**, since a ratchet that cannot fail is worse than none:

| mutation | test that reddened |
|---|---|
| `CONVERGING_PAIRS_THREE_PRIME` 0 -> 1 | `the_reported_pair_census_is_unchanged` |
| `1420-v2` cis input replaced by its span sibling | `the_1420_v2_pair_does_not_converge_by_re_derivation` |
| `OPEN_GAPS` 12 -> 11 | `the_open_gap_census_holds` |
| `1419-r1/cis` `output` corrupted | `every_gap_row_is_returned_exactly_as_authored` |
| second-pass fixed-point comparison inverted | `the_three_member_spelling_and_its_one_member_form_are_two_fixed_points` |

The last one is the check for vacuity rather than for the guard: inverting `assert_eq!(twice, once)` fails with both sides holding the real second-pass value, which is what shows the second pass is genuinely computed and compared.

Representation-Change: none. Tests and the ruling ledger only — no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is in this diff, and no `output` or `five_prime` pin moves, so no normalized string changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HGVS normalization handling for contiguous insertion/duplication representations and related `delins` recommendations.
  * Added safeguards against incorrect inversion merges and unintended re-derivation results.
  * Clarified directional normalization behavior and junction-crossing outcomes.

* **Tests**
  * Expanded regression coverage for convergence, sequence preservation, fixed points, and reported partition verdicts.
  * Updated adjudication records, ruling indexes, and measurement documentation to reflect current normalization results.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

nh13:
## CodeRabbit findings — disposition

Eleven actionable findings (the review body's inline list; there were no "Outside diff range" or "Duplicate comments" sections, so 11 threads = 11 actionable, reconciled against the `Actionable comments posted: 11` header).

The review ran against the **stacked** diff, which included #1649's four commits and `src/normalize/merge.rs`. Unstacking this PR removed the code five of the findings pointed at, so those are marked moot rather than rejected on their merits — several would be perfectly valid comments on #1649, and are noted below for whoever takes that PR.

| # | file | disposition |
|---|---|---|
| 1 | `src/normalize/merge.rs` — correct the `Representation-Change` trailer | **moot here / real on #1649.** No file under `src/` is in this diff any more, so `none` is now truthful and `Representation change declared` passes. **#1649 still needs this** — it changes `merge.rs` and its trailer has to describe the movement. |
| 2 | `cis_confluence_adjudication.rs` — verify `form_c` with `denotes_genomic` | **rejected: premise no longer holds.** `form_c` was added by #1649's re-bless. It is now only named in a doc comment; it is not a member of the spelling loop, so there is no unverified assertion to strengthen. (The enclosing test is also `#[ignore]`d pending #1617.) |
| 3 | `cis_junction_crossing_shift.rs` — assert the emitted outputs are fixed points | **fixed.** Real gap, and in this PR's own new code. The doc block claims all four strings are fixed points, but the two `THREE_MEMBER` rows pinned only where the *first* pass landed. Each pinned output is now normalized a second time in the same direction and required to be byte-identical. |
| 4 | `defect_non_idempotent_outputs.rs` — pin the exact six-spelling population | **moot here / worth keeping for #1649.** That file is no longer touched by this PR. |
| 5 | `reported_confluence_pairs.rs` — compare both `#1420-v2` spellings | **fixed, with the figure re-measured rather than trusted.** The span spelling's answer is now *measured* in the test and pinned, and the cis spelling is asserted to differ from it — so the forbidden string cannot drift away from what the span actually prints. Measured both directions: `g.38_41delinsATTG` -> `g.[38T>A;40_41delinsTG]`; cis is `g.[37dup;41del]` at 3' and `g.[36dup;41del]` at 5'. |
| 6 | `reported_partition_verdicts.rs` — stale module docs (`/span` rows read `Canonical`, four tests red, rewrite deferred) | **moot: that section is gone.** It was introduced by a dropped #1649 commit and is not in the file. |
| 7 | `reported_partition_verdicts.rs` — stale test reference | **fixed.** `each_pair_reaches_its_wanted_form_from_at_most_one_spelling` never existed; the intra-doc link is now `each_pair_reaches_its_canonical_form_from_exactly_one_spelling`. Verified no occurrence of the old name remains under `tests/`. |
| 8 | `reported_partition_verdicts.rs` — two `1419-r3/span` doc claims are false | **moot.** `FIVE_PRIME_MOVERS` is back to `main`'s two entries; `1419-r3/span` was only added to it by a dropped #1649 commit, and its `five_prime` equals its `output` on this base. The doc as written is correct again. |
| 9 | `reported_partition_verdicts.rs` — the `OPEN_GAPS` ratchet doc is attached to `PAIRS_NO_SPELLING_REACHES` | **fixed.** Correct and well-diagnosed: lines 602-673 were one unbroken `///` block ending at the wrong constant. `OPEN_GAPS` now carries its own ratchet contract and `PAIRS_NO_SPELLING_REACHES` carries its own doc. The intra-doc link fix is #7. |
| 10 | `reported_partition_verdicts.rs` — branch-aware failure guidance for `NormalizedMatch` | **rejected: the branch it describes is gone.** The `NormalizedMatch` expectation was #1649's — those pairs converged there. They do not converge on this base (`CONVERGING_PAIRS_*` are both 0), so all nine pairs expect `SequenceMatch` and there is one branch, not two. `PAIRS_NO_SPELLING_REACHES` is now documented as a claim about `wanted`, explicitly *not* about convergence. |
| 11 | `reported_partition_verdicts.rs` — the per-pair exemption also drops the retention pin from three rows that still retain their input | **fixed. Best finding in the set.** Correct: the exemption keyed on the *pair*, so `1419-r1/cis`, `1419-r2/cis` and `1419-r3/cis` — which do still print their input — lost their pin, and `assert_ne!(.., row.wanted)` would have passed for any other string they moved to. It now pins each row's measured `output`, which holds for the movers and the retainers alike, and the message carries measured-versus-expected. Mutation-tested: corrupting `1419-r1/cis`'s `output` reddens `every_gap_row_is_returned_exactly_as_authored`. |

Every finding that survived the restructure was verified against the code before acting, and the two figures the findings rested on (the `#1420-v2` span output, and the six-spelling population) were re-measured rather than taken from the review.

nh13:
@coderabbitai review

PR #1649, opened 2026-08-11
fix(normalize): let the splitter express two deletions, not just two insertions

### LINKED ISSUE #1617: Normalizer: converge the spelling-relative separation class on the re-derived form in the live partitioner

nh13:
Closes #1617.

`best_alignment` already had an escape from its single-gap restriction for *insertion, retained reference, insertion* — `two_gap_insertion_alignment`, added for #1260. The mirror shape was missing: *deletion, retained reference, deletion* had no expression, so a block like `ATGC -> T` got the single gap parked at one end and produced a 4-column spanning `delins` for what the block describes in 3. The weight bound then correctly refused that `delins` and the pass fell back to the per-member pipeline, whose answer **is** a function of how the input was spelled — which is the whole mechanism behind the non-confluence this repository has been chasing.

`two_gap_deletion_alignment` supplies the missing expression. It is a transpose, not a second implementation: it calls `two_gap_insertion_alignment` with `(result, reference)` and swaps each column's indices, so every bound, precondition and tie-break carries over and there is no duplicated index arithmetic to drift.

## Status: green, and the earlier "four tests are RED on purpose" section is withdrawn

That section described the pre-rebase branch and no longer applies. It said four `reported_partition_verdicts` tests were left failing because this PR re-blessed `output` while deliberately not touching `verdict`/`wanted`, and that repairing them needed an adjudication stacked on top.

**That adjudication landed on `main` independently**, in the `1419-r*` rows' own `argument` fields (`CORRECTED, and the verdict flipped Canonical -> Gap`). So on this base `wanted` is already the spanning `delins`, `verdict` is already `Gap`, and re-blessing the measured fields no longer creates an inconsistency — it resolves one. The rebase dropped the branch's own copy of that re-bless (`main` had moved the same file), and it is re-applied here on top: `output` and `five_prime` only, on the three `1419-r*/span` rows. `wanted`, `verdict`, `authority` and `argument` are untouched — zero changed lines among them, checkable with `git diff | grep -E 'wanted:|verdict:|authority:|argument:'`.

Full gate on this base: `cargo fmt --check` clean, `cargo clippy --all-features --all-targets -- -D warnings` clean, and the test suite green in both configurations CI runs it — armed with `FERRO_ASSERT_{IDEMPOTENT,REPARSE,IN_BOUNDS}` minus `ORACLE_EXCLUDE`, and plain over everything.

## It closes #1617, against a target `main` decided one commit ago

#1620 flipped `separation-is-a-property-of-the-spelling-not-of-the-variant` from `undecided` to `decided`, naming form **A** as the answer:

| | form | |
|---|---|---|
| A | `g.[1001009_1001010del;1001013del]` | **decided** |
| B | `g.[1001009del;1001011del;1001013del]` | "the defect #1419 and #1420 name in their own titles" |
| C | `g.1001009_1001013delinsCA` | outside the scope `delins-merge-vs-individual-gap-two-or-more` set for itself |

and shipped `the_decided_target_is_the_re_derived_form` `#[ignore]`d, describing it in its own commit message as **#1617's acceptance criterion**, with the follow-up spelled out in the test: *"delete the `#[ignore]` and move the two `apart` expectations above."* That test passes on this branch. The `#[ignore]` is deleted and the two pre-fix expectations move, exactly as instructed.

Form C is added to that test's loop as well. The record's table names C as conformant and **not** the target but says nothing about what C itself normalizes to, and C was its own fixed point until now — so one spelling of the variant was still reaching a fourth answer. It no longer is: all four spellings of that variant reach form A.

## What moved, measured

Nine confluence censuses move. **Every one improves monotonically** — `converged` up, every split arity down — and in each the three split deltas sum *exactly* to the `converged` delta, so every moved family went straight to convergence rather than merely dropping an arity.

```
                             converged        split_two      split_three   split_more
cis_confluence_axis    3'  8027 -> 8361   3111 -> 2835    115 ->  67    19 ->  9
                       5'  8023 -> 8367   3135 -> 2826    105 ->  74     9 ->  5
spec_conformance_axis  3'  9141 -> 9402   2440 -> 2225    248 -> 223    53 -> 32
                       5'  8944 -> 9228   2706 -> 2461    200 -> 167    32 -> 26
cis_confluence_nr_axis n3' 4012 -> 4179   1558 -> 1420     57 ->  33     9 ->  4
                       r3' 4010 -> 4177   1560 -> 1422     57 ->  33     9 ->  4
                       n5' 4011 -> 4183   1569 -> 1415     52 ->  36     4 ->  2
                       r5' 4009 -> 4181   1571 -> 1417     52 ->  36     4 ->  2
multi_member_cis_axis      respelling_converged 395 -> 401, 0 rows lost
```

**`multi_member_cis_axis` is manifest-gated, so that last row is NOT pre-merge coverage.** `axis_normalized` returns early with a `println!` when neither `FERRO_MANIFEST` nor `benchmark-output/manifest.json` is present, which is the case on PR CI — it skips silently there and only the nightly reference-aware job actually runs it. So `respelling_converged: 395 -> 401` and the six `CONVERGED_ROWS` additions are **nightly-only coverage**, and they are measured here against the local prepared reference rather than by any gating job. The one guard that does run on PR CI, `the_pinned_convergence_list_agrees_with_the_pinned_count`, checks the two constants against **each other** (401 names, 401 pinned) — not against reality, which is exactly the part a reader would otherwise mistake for verification. The other eight censuses in the table are hermetic and are gated normally.

**The nets are not what carries this.** A family entering `split_two` from `split_three` and a family leaving `converged` for `split_two` look identical in the totals. So the full divergence **row-id lists** were dumped on `origin/main` and on this branch — the `divergences` cap in `spec_conformance_axis::measure` raised locally for the measurement, then restored — and the sets diffed:

- **0 families lose convergence**, in any direction on any corpus.
- **0 families rise in arity**, and **0** become divergent that were not: the branch's divergent set is a strict *subset* of `main`'s (3': 2741 -> 2480; 5': 2938 -> 2654 on the spec corpus).
- **0 still-divergent families change arity at all** — which is why the sums are exact (215+25+21 = 261; 245+33+6 = 284).

`multi_member_cis_axis` is `+6 −0`, and all six rows are two-deletion cis pairs — the exact shape this change adds an expression for, which is the mechanism rather than a coincidence. Every rank-1 counter is unchanged everywhere: `sequence_changed`, `outputs_denoting_no_sequence`, `outputs_leaving_the_transcript`, the three refusal counters and `guard_violations`.

The `scale-*` non-idempotency family's second half closes too: `scale-c3p-sep136-del-del` loses the trailing `inv` that existed only because the splitter had no second gap to spend, so it lands on its authored design's output and `defect_non_idempotent_outputs`' `matches_authored` column reads `true, true, true` where it read `true, true, false`.

## Three enumeration rows move the other way, and the operator ruled on them

`spec_enumeration_tests`' `projection-splits-single-member` goes 9 -> 12 and `projection-pinned` 1168 -> 1165 — equal and opposite, so no row left the enumeration. The three are three projections of **one input**:

```
project-c/NM_004006.3:r.123_127delinsag
project-n/NM_004006.3:r.123_127delinsag
project-r/NM_004006.3:r.123_127delinsag
```

Measured through `spdi::compare_denoted_sequences` against the real reference: all three denote the input's own bases, so this is a representation move and not a correctness regression. The operator's ruling, recorded in full beside the constant:

- the **`n.`** row is a **fix** — #1682's payload-coincidence carve-out is `c.`-scoped by `rulings[delins-payload-coincidence-carve-out-is-coding-dna-scoped]`, so `general.md:34` governs unopposed and the split is required. This row was merged before and should not have been;
- the **`r.`** row is **unruled** — the same record puts `r.` out of the carve-out in both directions, and `RNA/delins.md` states no counterpart either way;
- the **`c.`** row is the only residue. Its real question is whether `coalesce_payload_alignment_split` should be promoted onto the shipping path (it is gated on `PartitionRule::CanonicalCoalesced` while the shipped rule is `Live`). That is a separate change which this PR neither makes nor blocks, and is deliberately not made here.

## #1419's three pairs converge

Each pair now reaches one string, from both spellings — the defect the issue reports:

```
1419-r1/span  g.19_33delinsCGG  [19_30del;33T>G] -> [19_23del;27_33del]
1419-r2/span  g.19_36delinsGCG  [19_33del;36A>G] -> [19_22del;26_36del]
1419-r3/span  g.19_33delinsGGA  [19T>G;22_33del] -> [19_24del;28_33del]
                            and at 5'            -> [18_23del;28_33del]
```

`CONVERGING_PAIRS_THREE_PRIME` and `_FIVE_PRIME` go 0 -> 3, and `FIVE_PRIME_MOVERS` gains `1419-r3/span`, both following from the measured fields. The clause is `general.md:34` read through `canonical-form-choice-when-both-legal` (`decided`): the partition is re-derived from the resulting sequence and what falls out is emitted.

**This is a convergence, not a closure.** The string the three pairs converge on is still not the spanning `delins` the decided chain wants, so all three stay in `PAIRS_NO_SPELLING_REACHES` and `OPEN_GAPS` stays at twelve. What changed is that the family stopped having two canonical forms.

## One real-world stability pin moves, at a gap of ONE

```
NC_000020.11:g.45891619_45891622delinsC
  was  g.45891619_45891622delinsC
  now  g.[45891619del;45891621_45891622del]
```

1 of 29 pinned `case_harvest` rows. It is self-described as a **stability pin, not a spec-correctness claim**, citing `basics.md:38` for stability as a value rather than for this spelling — and stability is a tiebreaker here, not a veto, since the project bar is confluence plus disclosure. The note named `delins-merge-vs-individual-gap-two-or-more` as governing and open; it was decided 2026-08-07 and scoped away from this row, which separates its two members by ONE unchanged nucleotide (`g.45891620`). The row is `g.`, so `general.md:35`'s codon exception cannot reach it either. `general.md:34` does, and asks for exactly this split.

## Review

Both CodeRabbit findings from the run against `6218f5e7` are addressed:

- **`reported_partition_verdicts` — "do not merge with four known failing tests."** Valid and fixed. All four pass; see the status section above for why the inconsistency it named no longer exists.
- **`merge.rs` — no coverage through `normalize_with_diagnostics()`.** Valid and fixed by `the_two_gap_deletion_split_is_a_fixed_point_through_the_diagnostics_exit`, which pins the exact displayed string from all three spellings, asserts the output is a fixed point of that same entry point, and asserts the diagnostics exit agrees with `normalize()` — the #1382 asymmetry as an assertion. Verified to fail on `origin/main`'s `merge.rs` and pass here, so it guards the change rather than restating it.

## What did NOT move

`long_delins_splits_at_unchanged_bases::a_long_net_deletion_stays_one_spanning_delins` passes, as does `a_delins_longer_than_the_old_cap_still_splits`. That 52 nt → 14 nt block is protected one layer downstream of the aligner by `separations_are_meaningful`, which refuses any proposed split whose inter-piece gaps are one base wide against `RAISED_PIECE_SEPARATION`. This change only widens what the aligner may *propose*; the separation floor still refuses it.

Rebased onto `62c83ec1` after the first push, which brought in #1673 — the wholesale replacement of `tests/fixtures/sequences/normalization_transcripts.json`'s exon tables (111 exons and 105 junctions restored; 22 -> 28 multi-exon records, 412 -> 517 junctions). Because this branch pins confluence census counts, that fixture could have moved them without a line of this diff changing. It did not: both spec artifacts were regenerated on the new base and all 85 census/adjudication tests re-run green with **no pin moved**, which is consistent with the mechanism — none of the census modules read that fixture (they build synthetic corpora, `NM_TEST.1` and `TEMPLATE`, or resolve the manifest). Its three readers are `idempotency_tests`, `normalize_tests` and `normalize_property_tests`, none of which pins a census here.

The corpus-level figure quoted in an earlier revision of this description (282 of 78298 rows, measured with `examples/dump_normalized_corpus.rs` against `main` at `ec8ebd66`) was measured before the rebase and has **not** been re-measured on this base; do not read it as current. The census figures above are the measurement of record.

Representation-Change: 9 confluence censuses move and none regresses; `converged` rises 8027->8361 (3') and 8023->8367 (5') in cis_confluence_axis, 9141->9402 and 8944->9228 in spec_conformance_axis, and by 167/167/172/172 across cis_confluence_nr_axis's n./r. x 3'/5' pins, while multi_member_cis_axis's respelling_converged goes 395->401 with 0 rows lost - that last one measured against a local prepared reference and NOT pre-merge coverage, since multi_member_cis_axis is manifest-gated and skips silently on PR CI; only the nightly runs it.
  In every one of those censuses the three split-arity deltas sum EXACTLY to the converged delta, so each moved family converged outright rather than dropping an arity.
  Measured by dumping the full divergence row-id lists on origin/main and on this branch and diffing the sets, not by reading the nets: 0 families lose convergence, 0 rise in arity, and 0 become divergent that were not, on any corpus in either direction.
  Moving the other way, 3 spec-enumeration rows go from projection-pinned to projection-splits-single-member (1168->1165 and 9->12) - the three projections project-{c,n,r} of NM_004006.3:r.123_127delinsag, sequence-preserving via spdi::compare_denoted_sequences, ruled a fix on n., unruled on r., and one deferred residue on c.
  Six previously-accepted pinned strings also move: the 3' and 5' outputs of 1419-r1/span, 1419-r2/span and 1419-r3/span in reported_partition_verdicts, which converge each pair onto its /cis sibling's form - a real migration for any consumer storing those.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -6458,6 +6458,16 @@ fn best_alignment(reference: &[u8], result: &[u8]) -> Option<Vec<Column>> {
             return Some(columns);
         }
     }
+    // The mirror, on a net-deletion block: a single gap cannot express
+    // *deletion, retained reference, deletion* either, and the same argument
+    // licenses it — see [`two_gap_deletion_alignment`]. The trigger is the
+    // mirror too: the retained side of a net deletion is the result, so what
+    // signals a single gap's failure here is an unmatched *result* base.
+    if !shorter_is_ref && best.is_some_and(|(matches, _, _)| matches < result.len()) {
+        if let Some(columns) = two_gap_deletion_alignment(reference, result) {
+            return Some(columns);
+        }
+    }
 
     // Only the winner is materialized.
     best.map(|(_, _, k)| {
@@ -6536,9 +6546,16 @@ fn best_alignment(reference: &[u8], result: &[u8]) -> Option<Vec<Column>> {
 /// introduced and no match is manufactured: every matched column is a reference
 /// base surviving verbatim. That argument never depended on `a` being 0 — all
 /// three chunks are verbatim reference either way. Cluster A is structurally out
-/// of reach — #1040 and the `inv` cases are equal-length, #1157A is a net
-/// deletion, and an equal-length or net-deletion block never reaches this
-/// function.
+/// of reach: #1040 and the `inv` cases are equal-length, and an equal-length
+/// block never reaches this function.
+///
+/// A **net-deletion** block reaches it only through
+/// [`two_gap_deletion_alignment`], which calls this with the two blocks swapped.
+/// The containment argument transposes with them and is unchanged in substance,
+/// but note the consequence for reading this function's preconditions: on that
+/// path `reference` is the observed sequence and `result` the reference. #1157A
+/// is a net deletion and is still out of reach — it has no decomposition of this
+/// shape in either orientation.
 ///
 /// Comparison is byte-exact, matching [`best_alignment`]'s own `==` throughout.
 /// Making only this path case-insensitive would let the two-gap form win on a
@@ -6606,6 +6623,60 @@ fn two_gap_insertion_alignment(reference: &[u8], result: &[u8]) -> Option<Vec<Co
         }
     }
     None
+}
+
+/// The alignment that keeps the **whole** result intact across two deletions,
+/// or `None` when the pair does not have that shape — the mirror of
+/// [`two_gap_insertion_alignment`].
+///
+/// # Why it is a transpose and not a second implementation
+///
+/// The two shapes are the same statement read from opposite sides. This
+/// function looks for
+///
+/// ```text
+/// reference = result[..a] + D1 + result[a..b] + D2 + result[b..]
+/// ```
+///
+/// and that is *literally* [`two_gap_insertion_alignment`]'s equation with the
+/// two blocks exchanged. Calling it with `(result, reference)` and swapping each
+/// column's two indices therefore yields exactly the alignment wanted here — so
+/// every bound, precondition and tie-break of that function applies unchanged,
+/// and there is no duplicated index arithmetic to drift.
+///
+/// # Why it does not reopen the corpus the single-gap rule protects
+///
+/// The same argument, transposed. The single-gap restriction exists to stop
+/// *compensating* gaps — a deletion **and** an insertion in one block — from
+/// manufacturing matches out of coincidence (#1034/#1040/#182). Both gaps here
+/// are deletions in a net-deletion block, so no insertion is introduced and no
+/// match is manufactured: every matched column is a result base that survives
+/// verbatim in the reference.
+///
+/// # What it is for
+///
+/// A block like `ATGC -> T` has no single-gap explanation that stays within its
+/// own edit distance: the lone gap parks at one end and the offset reads as
+/// substitutions, giving the 4-column `delins` where the block needs 3. The
+/// two-gap form is the description the block actually admits — delete `A`,
+/// retain `T`, delete `GC` — and it is the one `general.md:34` asks for, since
+/// the two deletions are separated by an unchanged nucleotide.
+///
+/// Without it, `canonicalize_from_sequence`'s weight bound correctly refuses the
+/// over-claiming `delins` and the pass falls back to the per-member pipeline,
+/// whose answer *is* spelling-dependent — so two spellings of one variant
+/// diverge. That is the confluence regression this closes. The bound is named
+/// generically here on purpose: the ceiling it compares against is derived from
+/// the input's spelling today and is being moved onto the block by a separate
+/// change, and this alignment is what it is either way.
+fn two_gap_deletion_alignment(reference: &[u8], result: &[u8]) -> Option<Vec<Column>> {
+    let transposed = two_gap_insertion_alignment(result, reference)?;
+    Some(
+        transposed
+            .into_iter()
+            .map(|(alt_index, ref_index)| (ref_index, alt_index))
+            .collect(),
+    )
 }
 
 /// Columns for an alignment with a single gap of `gap_len` after `k` aligned
@@ -14042,11 +14113,14 @@ mod tests {
         }
 
         /// Run both splitters on one block and check the two invariants that hold
-        /// for every disagreement found: both sides denote the alternate block,
-        /// and the sequence-first side never declines.
+        /// for every harvested block: both sides denote the alternate block, and
+        /// the sequence-first side never declines.
+        ///
+        /// Says nothing about whether the two agree — [`both`] and [`converged`]
+        /// are the two callers that do.
         ///
         /// Returns `(live, shadow)` for the caller to pin exactly.
-        fn both(reference: &[u8], result: &[u8]) -> (Vec<Piece>, Vec<Piece>) {
+        fn both_raw(reference: &[u8], result: &[u8]) -> (Vec<Piece>, Vec<Piece>) {
             let live = partition_block(reference, result);
             let shadow = partition_block_sequence_first(
                 reference,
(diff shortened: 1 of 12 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

Issue #1650, opened 2026-08-11
Normalizer: a cis allele straddling a CDS boundary is refused by the sequence-first pass, giving two fixed points for one variant


nh13:
## Summary

A cis allele whose members straddle a CDS boundary is refused by the sequence-first pass, so the multi-member spelling and the single-member `inv` spelling of one variant are two fixed points. The identical design converges when the members sit inside the CDS.

This is the refusal #1536 originally named. #1536 turned out to be a different defect with a different cause — the #350 cross-axis bail in `normalize_cds`, fixed separately — and this half was left open there, with a measured, `#[ignore]`d red guard in `tests/it/issue_1536_cds_boundary_delins.rs`.

## Reproducer

40-base single-exon transcript, 8-base block at transcript 23..=30. Only the CDS annotation moves.

```
CDS 11..=30   c.[13_15delinsCTA;18T>G;20G>T] -> itself
              c.13_20inv                     -> c.[13_15delinsCTA;18T>G;20G>T]     converges
CDS 11..=26   c.[13_15delinsCTA;*2T>G;*4G>T] -> itself
              c.13_*4inv                     -> c.13_*4inv                          does NOT
```

Same bases, same payload, same design. The discriminator is where the stop codon falls.

The lone-`delins` symptom is visible in the same place: `c.13_*4delinsCTACGGAT` now converges with `c.13_*4inv`, but neither is *split* into the three members the CDS-interior twin is split into.

## Cause

Two refusals, both keyed on the region a position sits in rather than on the sequence:

- `merge::collect_canonical_edits` (`src/normalize/merge.rs`) refuses any member whose `region != body_region(kind)`, and `merge::join_pos` refuses any member whose two endpoints are in different regions. Together they mean the sequence-first pass never sees a `c.-N` or `c.*N` member, nor one that straddles.
- `Normalizer::apply_canonical_split`'s own `simple_cds_pos` (`src/normalize/mod.rs`) refuses `is_3utr()` and `base <= 0` outright, so the per-member split path cannot serve them either.

[...]

PR #1651, opened 2026-08-11
fix(normalize): canonicalize a member whose span crosses a CDS boundary

### LINKED ISSUE #1536: Normalizer: a delins crossing a CDS boundary is never re-derived, giving two fixed points for one variant

nh13:
Closes #1536

## What was wrong

A `delins` whose span crosses a CDS/UTR boundary was never re-derived from the sequence it denotes, so two spellings of one variant reached two different fixed points. The identical block placed wholly inside the CDS converges correctly — the discriminator was nothing but where the start or stop codon happens to fall:

```
c.9_*1delinsTGTGCATT   ->  c.9_*1delinsTGTGCATT     (unchanged)
c.9_*1inv              ->  c.9_*1inv                (unchanged)
```

Two fixed points, one variant.

## Root cause

The cross-axis bail's stated ground is that **the shuffle** has no defined semantics across an axis boundary. It then discarded the whole per-member pipeline, canonicalization included.

Re-typing a `delins` against the bases it denotes — `canonicalize_delins`: `inv`, `del`, `ins`, `dup`, `sub`, plus the shared-affix trim — is not a shuffle. It reads the reference under the member's own span and moves nothing. The `c.` reference is a single contiguous string (`refseq.md`) whose axis *labels* change at `cds_start`/`cds_end` while the sequence does not; that is the argument #918 already makes verbatim for a whole-CDS del/dup.

## The boundary is the discriminator, not the transcript end

The obvious confound — that these spans also reach the end of the transcript — was ruled out by measuring on a 40-mer with a real UTR at **both** ends. Over the 33 placements of one 8-base whole-block reverse complement:

- the **14** that cross a CDS boundary all diverged
- the **19** that do not all converged — including the three sitting wholly in a UTR, and the one running to the transcript end

So the trigger is the boundary, not the sequence end.

## Scope: `Delins` and `Inversion` together, everything else keeps the bail

Confluence is a property of the **pair**. The `delins` spelling reaches `inv` only if the `inv` spelling reaches the same string, and the `inv` spelling is subject to the same shared-affix trim (`c.3_10inv` → `c.4_9inv`). Admitting one and not the other trades one non-confluence for another.

Every other edit kind keeps the bail:
- for a `del`/`dup` the canonicalization **is** the shuffle, so admitting them would be the cross-axis shift #350 refuses and #918 carves out only for a whole-CDS span
- a `sub` cannot straddle a boundary
- an `ins` is already carved out by #402

## A second, independent refusal — named, not fixed here

`merge::join_pos` / `collect_canonical_edits` refuse any interval whose ends are in different regions (`region != body`), and `apply_canonical_split`'s `simple_cds_pos` refuses `*N`/`-N` outright. Those block cross-region **splits**, which is a real capability gap but a different one:

```
CDS 11..=30   c.13_20delinsCTACGGAT  ->  c.[13_15delinsCTA;18T>G;20G>T]   (split)
CDS 11..=26   c.13_*4delinsCTACGGAT  ->  c.13_*4inv                        (converged, not split)
```

Fixing only `merge.rs` would **not** have fixed the reported symptom — the #350 bail is its root cause. The remaining gap is stated here rather than silently left as an unexplained asymmetry.

## Correcting a claim in the issue

The issue notes that the corpus is structurally blind here because "every pre-existing family is single-exon with `CDS_START = 1`". **That is stale.** `CDS_END = 15` on a 20-mer `c.` reference already yields a 3'UTR, and #1478's `cx.` axis added `CDS_START_MULTI = 4` / `CDS_END_MULTI = 15`, so both boundaries are reachable. The corpus does see this property — hence a real 1826-row blast radius rather than a structural zero.

## Rebased, and every figure below re-measured on the new base

This branch was opened against `822eb5d5`. Since then **#1599, #1606, #1609, #1622, #1639, #1640, #1644, #1645 and #1646** landed, and #1599 re-blessed the very constants this PR re-blesses — `tests/it/spec_conformance_axis.rs` was a real merge conflict, not a lagging `mergeStateStatus` (`git merge-tree` returned that file with rc=1; GitHub agreed at `CONFLICTING`/`DIRTY`). It is now rebased onto `bfcc1802` and clean.

**Nothing below is composed from the two changes' deltas.** A census counts *classes*, not row-deltas, and the two changes touch overlapping regions of the corpus, so a composed number would not be a measurement. Every figure was taken by running the corpus on the rebased branch, and every `before` figure by running the same corpus against `main`'s `src/normalize/mod.rs` in the same tree — which reproduced the committed post-#1599 pins exactly (`rc=0`), so both sides come from one corpus on one machine.

For the record, the deltas *did* turn out to be additive, because the two changes' affected row sets are disjoint — verified by diffing row ids, not assumed. That is a result, not the method.

## Confluence census

`before` = `main` @ `bfcc1802`; `after` = this branch.

| | converged | split 2 | split 3 | split 4+ | non-idempotent |
|---|---|---|---|---|---|
| 3' before | 9139 | 2436 | 250 | 57 | 4 |
| 3' after | **9141** | 2440 | 248 | 53 | 4 |
| 5' before | 8944 | 2698 | 202 | 38 | 4 |
| 5' after | 8944 | 2706 | 200 | 32 | 4 |

Every rank-1 counter is unchanged in both directions.

**The net counters cannot tell net from gross** — a family entering `split_two` from `split_three` and a family leaving `converged` for `split_two` look identical in the totals — so the full divergence lists were dumped on both sides and the row ids diffed:

- **2 families newly converge** (3' only): `s00-c3{p,m}-cds-end-del-ins-p1-sep2`. At 5' the same two rows only drop 3 -> 2.
- **10 families drop arity at 3'** and **14 at 5'**, 10 of them the same row ids in both: `s00-c3{p,m}-cds-end-` `del-del-p1-sep2` (3 -> 2), `del-del-p2-sep1` (4 -> 3), `del-ins-p2-sep1` (4 -> 3), `sub-sub-p1-sep0` (3 -> 2), `sub-sub-p2-sep0` (3 -> 2). The 5'-only four are `del-ins-p1-sep2` (3 -> 2) and `del-ins-p2-sep2` (4 -> 3), both strands.
- **0 families lose convergence and 0 rise in arity, in either direction.**

So the census assertion's instruction to "promote any newly-failing row to a named test in `spec_corpus_regressions.rs`" has **nothing to promote here** — which is a measured claim about a diffed row set, not the absence of a search. (#1599, directly below in history, did have six rows to promote and promoted them.)

**Every moved row is `s00-c3{p,m}-cds-end-*`, and that is the corpus speaking about itself.** This corpus builds boundary designs at `cds_end` and none at `cds_start`, so the 5'UTR/CDS half of the carve-out is *unmeasured by this census*. It is covered instead by `issue_1536_cds_boundary_delins`, whose `PLACEMENTS` table crosses `cds_start` (`c.-3_5`) and whose 33-placement walk crosses both. Recorded rather than left to read as "the fix only affects the CDS end".

## A test that read as coverage and guarded nothing

`a_retyped_member_that_stops_straddling_is_shuffled_and_is_a_fixed_point` asserted only `once == twice` on `c.20_*7delinsCC`, with a doc comment claiming that input emitted `c.*1_*4del` without the recursion step. **Measured: it passes with the recursion block's gate forced closed.**

Both halves of the claim were wrong for that input. Its `ref` (`GCTTACGG`) and payload (`CC`) share neither a prefix nor a suffix, so the trim removes nothing, the member never stops straddling, and the recursion's `new_variant != *variant` gate never opens — the assertion was satisfied by an input the carve-out *declines* to change. The mechanism the comment described is real; the demonstration was of a different shape.

Found by grid-probing every straddling `delins` on the fixture under both builds and diffing: **27** inputs produce a different output with and without the recursion, 9 of them on `WIDE_CORE`, so no new fixture was needed. The test now runs on `c.20_*4delinsG`:

```text
with the recursion:     c.20_*4delinsG -> c.*2_*5del    (a fixed point)
without the recursion:  c.20_*4delinsG -> c.*1_*4del    (NOT a fixed point)
```

and asserts both the shuffled value and the fixed point, plus that `c.*1_*4del` is *not* itself a fixed point so the pair cannot go vacuous the same way again. **Verified: with the gate forced closed the test now FAILS**, and 3' `non_idempotent_outputs` reads 6 rather than 4 (the two extra rows being `s00-c3{m,p}-cds-end-del-del-p2-sep1`). `the_input_that_taught_us_this_test_can_be_vacuous` pins the old input as the decline it is.

The same measurement corrected a stale figure in `normalize_cds`'s own comment, which quoted the recursion as holding `non_idempotent_outputs` down "from 7 to 9". 7 was the pre-#1599 baseline; on this base it is 4 -> 6, re-measured rather than rebased by arithmetic.

The `#[ignore]` on the multi-member guard also said "tracked separately" with no number. It is **#1650**; both the attribute and the doc comment now say so.

## Verification

All on the rebased branch, with **no** `FERRO_ASSERT_*` variable set (the plain `test` job runs un-armed, and arming them changes what fails).

| command | rc | result |
|---|---|---|
| `cargo nextest run --features dev --lib --test it` | 0 | `9949 tests run: 9949 passed, 150 skipped` |
| `... -E 'test(spec_conformance_axis) + test(spec_corpus_regressions) + test(defect_non_idempotent_outputs) + test(issue_1536)'` | 0 | `38 tests run: 38 passed` (the skip count is what shows the filter selected rather than matched nothing) |
| `... -E 'test(issue_1536)' --run-ignored all` | 100 | 8 passed, 1 failed — the #1650 guard, red by design |
| `cargo clippy --all-features --all-targets -- -D warnings` | 0 | clean (one pre-existing third-party future-incompat notice for `proc-macro-error2 v2.0.1`) |
| `dump_normalized_corpus --out` x2 + `--compare` | 0 | 78,298 rows each side, 1,826 moved |

**One figure is not fully re-verified, and is flagged rather than restated.** The earlier trailer's parenthetical that the *canonical* partitioner arm gives the identical moved-row set at `29/56/1741` was measured on `822eb5d5`; #1639 has since changed the partitioner-switch behaviour, and that A/B was not re-run. Treat it as unverified on this base. The shipped-arm figures below were re-measured.

The merge-vs-split partition of the 79 non-respell rows is `20/59` by member count over the two dumps. The declared `25/54` came from a derivation `--compare` does not emit, so the two cannot be reconciled from the artifacts; the **total** (79) and every other figure agree exactly.

## Review round 2: a dropped warning let strict mode stop refusing

The re-typing recursion returned the inner result outright, discarding **every** warning the outer
pass had accumulated. `warnings` there comes from `normalize_na_edit`, which validates the reference
and pushes `RefSeqMismatch` — and `NormalizeResult::has_ref_mismatch` is what strict mode refuses
on. So strict `normalize()` silently **accepted** a straddling delins whose stated deleted bases do
not match the reference.

It now filters on what each warning asserts rather than dropping the set:

* **Invalidated, so dropped** — `CrossAxisVariantNotShuffled` (the member *is* shuffled, by the
  recursive call) and `AxisClampApplied` (its contract is to say why a position did *not* shift
  further, and here it did: `c.*1_*4del` -> `c.*2_*5del`).
* **Carried** — everything else. `RefSeqMismatch` is not made untrue by re-typing a span, and
  `ReducedCapabilityNoGenome` is a statement about the *provider*, which a re-type cannot fix.

No de-duplication: `NormalizationWarning` is not `PartialEq`, and keying on `code()` would conflate
two `RefSeqMismatch` warnings at different positions — position is the whole key, as the merge-time
comment says. A duplicated warning is strictly preferable to a lost one.

**How this was missed the first time, recorded because the reasoning was the failure.** An assertion
at the re-entry line over the whole `--lib --test it` suite fired on exactly **one of 9 956** tests,
carrying only `AxisClampApplied`, and that was read as licence to drop both vectors. It is a claim
about the corpus, not the code: no corpus row builds a straddling retype with mismatched explicit
bases, so this class never reached the line. A `1 of 9 956` carries the same epistemic weight as a
`0 of 9 956` — state the population the probe could not build before concluding.

`a_reference_mismatch_survives_the_recursion_and_strict_mode_still_refuses` pins both halves
(warning present in the default mode, `Err` under `ErrorMode::Strict`). Mutation-tested: restoring
the blanket drop yields `warnings=[]` and reddens it.

**Behaviour note that is not a representation change.** No normalized output string moves — the
trailer's figures are unchanged and were re-measured on this base. What changes is that strict mode
now refuses a class of input it previously accepted, which is the point of the fix.

Validated as CI runs it, not with oracles armed globally: plain `test` **10142/10142**, and
`test-oracle` (armed, `ORACLE_EXCLUDE` negated) **10080/10080**. `clippy --all-features
--all-targets -D warnings` clean.

Representation-Change: 1826 of 78298 corpus rows move (2.3%) — 1747 respell and 79 change member count. Re-measured on this branch rebased onto bfcc1802, not carried over from the base it was opened against. 1460 of the moved rows were previously accepted forms; the other 366 were not fixed points, so moving those is free. Every move is a straddling member being re-typed or affix-trimmed against the bases it already denoted; no move changes what a description denotes.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -4864,6 +4864,54 @@ impl<P: ReferenceProvider> Normalizer<P> {
             && matches!(edit, NaEdit::Deletion { .. } | NaEdit::Duplication { .. })
             && matches!(start_axis, boundary::AxisRegion::FiveUtr)
             && matches!(end_axis, boundary::AxisRegion::ThreeUtr);
+        // **Span-preserving re-typing exception (#1536).** The bail's stated
+        // ground is that *the shuffle* has no defined semantics across an axis
+        // boundary — and it then throws away the whole per-member pipeline,
+        // canonicalization included. Re-typing a `delins` against the bases it
+        // denotes (`canonicalize_delins`: `inv`, `del`, `ins`, `dup`, `sub`,
+        // plus the shared-affix trim) is not a shuffle. It reads the reference
+        // under the member's own span and moves nothing, and the `c.` reference
+        // is a single contiguous string (`refseq.md`) whose axis labels change
+        // at `cds_start`/`cds_end` while the sequence does not — the argument
+        // #918 already makes verbatim for a whole-CDS del/dup.
+        //
+        // Discarding it is what made `c.9_*1delinsTGTGCATT` and `c.9_*1inv` two
+        // fixed points for one variant, discriminated by nothing but where the
+        // stop codon falls (#1536). Measured on a 40-mer with a real UTR at both
+        // ends: over the 33 placements of one 8-base whole-block reverse
+        // complement, the 14 that cross a CDS boundary all diverged and the 19
+        // that do not all converged — including the three that sit wholly in a
+        // UTR and the one that runs to the transcript end, so the discriminator
+        // is the boundary and not the sequence end.
+        //
+        // `Delins` and `Inversion` together, not `Delins` alone. Confluence is a
+        // property of the *pair*: the `delins` spelling reaches `inv` only if the
+        // `inv` spelling reaches the same string, and the `inv` spelling is
+        // subject to the same shared-affix trim (`c.3_10inv` -> `c.4_9inv`).
+        // Letting one through and not the other trades one non-confluence for
+        // another.
+        //
+        // Every other edit kind keeps the bail. For a `del`/`dup` the
+        // canonicalization *is* the shuffle, so admitting them would be the
+        // cross-axis shift #350 refuses and #918 carves out only for a
+        // whole-CDS span; a `sub` cannot straddle a boundary; an `ins` is
+        // already carved out by #402.
+        let is_span_preserving_retype =
+            matches!(edit, NaEdit::Delins { .. } | NaEdit::Inversion { .. });
+        // Warnings raised before `normalize_na_edit` runs, so they survive the
+        // carve-out below rather than being dropped with the early return.
+        let mut pre_warnings: Vec<NormalizationWarning> = Vec::new();
+        // Whether the span-preserving re-type carve-out below was actually
+        // taken. Recorded explicitly rather than inferred from
+        // `!pre_warnings.is_empty()`, even though the two agree today: the
+        // recursion gate near the end of this function is a control-flow
+        // decision about *this* path, and keying it on whether some vector
+        // happens to be non-empty makes it fire for any future warning pushed
+        // before `normalize_na_edit` — which the name `pre_warnings` positively
+        // invites. That is the same mistake, one level up, as deciding which
+        // warnings to drop by what the vector happened to hold; see the note on
+        // the gate itself.
+        let mut took_span_preserving_retype = false;
         if start_axis != end_axis
             && !matches!(start_axis, boundary::AxisRegion::None)
             && !matches!(end_axis, boundary::AxisRegion::None)
@@ -4876,10 +4924,23 @@ impl<P: ReferenceProvider> Normalizer<P> {
                 start_axis: start_axis.label().to_string(),
                 end_axis: end_axis.label().to_string(),
             };
-            return Ok((
-                HV::Cds(self.canonicalize_cds_variant(variant)),
-                vec![warning],
-            ));
+            if !is_span_preserving_retype {
+                return Ok((
+                    HV::Cds(self.canonicalize_cds_variant(variant)),
+                    vec![warning],
+                ));
+            }
+            // The warning still stands, and is still true: nothing below may
+            // move this member off the footprint it arrived on. Pinning the
+            // shuffle bound to the input span is what enforces that — the
+            // re-typing runs, and a derived piece may settle anywhere inside the
+            // member's own hull, but no shift can carry it into a region the
+            // input did not already occupy. `axis_info.clamped` cannot serve
+            // here: it is clamped to the *start* endpoint's region, which for a
+            // straddling member is narrower than the member itself.
+            pre_warnings.push(warning);
+            took_span_preserving_retype = true;
+            boundaries = Boundaries::new(tx_start.saturating_sub(1), tx_end);
         }
 
         // #918: relax the CDS↔3'UTR axis clamp for a CDS-resident del/dup so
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
Applied the CLI review's remaining finding in `c6c6a8ca`, and it is worth stating what it was and was not.

**It is not a live bug, and I checked that before changing anything.** The recursion gate read `!pre_warnings.is_empty()` as a proxy for "the span-preserving re-type carve-out ran". That vector is declared at one place and pushed at exactly one site — inside the carve-out — so the proxy is *exact* on this tree, and no input reaches the gate through it wrongly.

**It is worth fixing because it is the same mistake as the defect this PR corrects, one level up.** The correction already recorded here is that the warning filter must key on *what each warning asserts*, not on what the vector happened to hold. The gate was keying a control-flow decision on whether that same vector happened to be non-empty. A second warning pushed before `normalize_na_edit` — which the name `pre_warnings` positively invites — would fire the recursion for a variant that never took the carve-out, and nothing would say so. So the fact is now recorded as a fact: `took_span_preserving_retype`, set at the carve-out and read by the gate.

The prepend at the end of the function still tests emptiness, deliberately — there the question really is "is there anything to prepend".

**Mutation-tested, because a flag that is never read is indistinguishable from one that is.** Forcing it to never be set reddens two guards:

- `a_retyped_member_that_stops_straddling_is_shuffled_and_is_a_fixed_point`
- `the_recursion_emits_no_stale_clamp_warning`

**Behaviour-neutral, and measured rather than asserted.** Validated at CI parity: `cargo fmt --check` clean, `clippy --all-features --all-targets -D warnings` clean, plain suite **10142/10142**, oracle suite (three flags armed, `ORACLE_EXCLUDE` negated as `test-oracle` does) **10080/10080** — byte-identical counts to the pre-change run, which is the point.

PR #1657, opened 2026-08-11
fix(ci): read the count, not the denominator, in a quantified-zero trailer

### LINKED ISSUE #1647: fix(ci): the representation-change trailer form the docs call valid is rejected by the checker
## Summary

`CONTRIBUTING.md` and `CLAUDE.md` both tell contributors to quantify a zero as `none. 0 of 950 rows move`, and both state that it **passes**. It does not — `scripts/check_representation_change.py` exits **1** on it. The documented form is a merge blocker.

```console
$ printf 'Representation-Change: none. 0 of 950 rows move.\n' > /tmp/doc-example.md
$ echo "src/normalize/merge.rs" > /tmp/watched.txt
$ python3 scripts/check_representation_change.py --changed-files /tmp/watched.txt --declaration-file /tmp/doc-example.md
This trailer declines and then describes a move:

  none. 0 of 950 rows move.

The verdict is the first word, so this is filed as `none` and the disclosure -- '950 rows move' -- never reaches the changelog.
$ echo $?
1
```

The two assertions:

- `CONTRIBUTING.md:125` — "Quantifying a zero is fine and encouraged — `none. 0 of 950 rows move` passes."
- `CLAUDE.md:481` — "Quantifying a *zero* is encouraged and passes (`none. 0 of 950 rows move`)"

`CLAUDE.md` states the operative rule correctly *in the same sentence* — "the pattern requires the count to sit immediately before `rows`" — which its own example violates. The example and the rule contradict each other one clause apart.

## Why the existing test does not catch it

`tests/python/test_representation_change_trailer.py` includes the string (`"none — 0 of 950 rows move"`), but the assertion is only that `release-plz.toml`'s changelog rule and `check_representation_change.declines()` **agree on whether it is a decline**. They do agree — it *is* a decline. The contradiction rule fires afterwards, and nothing asserts the end-to-end exit code.

So there is a coverage gap with a specific shape: **no test asserts that the trailer forms the docs present as valid actually pass the checker.** The existing test pins two classifiers against each other; it does not pin either against the documentation.

## What actually passes, and why it is accidental

```
none. 0 of 950 real cis-allele rows move.   -> rc=0   (passes)
none. 0 of 950 rows move.                   -> rc=1   (fails)
none. 0 of 78,298 rows move.                -> rc=1   (fails)
```

The first survives only because the noun phrase `real cis-allele` sits between the number and `rows`, so the `<n> rows move` pattern does not match. That is a property of #1547's wording, not a rule anyone can infer — and the docs generalise from it to a form that fails.

This is easy to hit: quantifying a zero as `0 of <corpus size> rows move` is the natural phrasing, and the corpus size is always non-zero, so the pattern matches the denominator every time.

## Suggested fix

Any of these, but the third is the one that stops it recurring:

1. Correct both docs to a form that passes, and say explicitly that the count immediately before `rows` is what is matched — so `0 of 950 rows move` fails while `0 of 950 real cis-allele rows move` passes.
2. Or relax the checker so a leading `0 of` is recognised as quantifying a zero rather than disclosing a move.
3. **Add a test that runs every trailer form the docs present as valid through the checker end-to-end and asserts `rc == 0`** (and every form they present as invalid, asserting non-zero). The current test compares two implementations to each other; nothing compares either to the prose, which is exactly how this drifted.

Found while sweeping #1620, whose trailer reads `none. … confirms 0 of 78,298 rows move.` It passes today **only** because that PR touches no watched directory, so the checker short-circuits before evaluating the text. Add one watched file and it blocks.

nh13:
Representation-Change: none. No watched directory is touched — this changes `scripts/`, `tests/python/` and `CLAUDE.md` only.

Closes #1647

## The documented form was a merge blocker

`CONTRIBUTING.md:125` and `CLAUDE.md` both publish `none. 0 of 950 rows move` as the way to quantify a zero, and both state that it passes. It did not:

```console
$ printf 'Representation-Change: none. 0 of 950 rows move.\n' > /tmp/doc-example.md
$ echo "src/normalize/merge.rs" > /tmp/watched.txt
$ python3 scripts/check_representation_change.py --changed-files /tmp/watched.txt --declaration-file /tmp/doc-example.md
This trailer declines and then describes a move: ...
$ echo $?
1
```

The contradiction rule read the number sitting immediately before `rows` as the moving count. In that phrasing that number is the **denominator**, so the rule fired on the 950.

`CLAUDE.md` stated the positional rule and its own violating example one clause apart, which is why it read as a description of the behaviour rather than as the defect it was.

## The fix

Read the count as the count, in both orders this repository actually writes:

```
3 rows of 500,004 move      denominator trailing
3 of 500,004 rows move      denominator between the count and `rows`
```

Two lookbehinds carry it, and **the second is not redundant**:

- `(?<!of\s)` refuses to read a denominator as a count — this is what rescues `0 of 950 rows move`.
- `(?<![\d,])` stops the match restarting in the *middle* of a grouped number. Without it, `0 of 78,298 rows move` still fires: the comma opens a word boundary, so the scan resumes at `298`, which is nonzero, sits immediately before `rows`, and is not preceded by `of `. Measured — the first lookbehind alone leaves that decline failing.

## The tripwire is not blunted

A decline that discloses a nonzero move is still refused, in either order. That is the one direction this mechanism must never fail in. The count-first order is now caught **on the count** rather than incidentally on the denominator, which is why `no. 3 of 500,004 rows move` still exits 1.

## Verification

Checked against every trailer form the two documents publish plus the recorded declines of #1535, #1546, #1538 and #1547. The new rule agrees with intent on all 14; the old rule was wrong on two of them (`0 of 950 rows move` and `0 of 78,298 rows move`).

- `tests/python/test_representation_change_trailer.py` — 10 new cases across three tests, including one end-to-end through `check()` because the reported symptom was the CLI's **exit code**, not the predicate.
- `test_representation_change_trailer.py` + `test_changelog_grouping.py` green together, **115 passed** — so the checker and the changelog audit still agree, which `test_the_audit_and_the_checker_agree_on_every_documented_form` exists to enforce.
- `ruff check` and `ruff format --check` clean.

The remaining `tests/python/` modules fail to collect in a fresh worktree with `ModuleNotFoundError: No module named 'ferro_hgvs'` — the native extension is not built there. Pre-existing and unrelated; both files touched here load the script by path and need no extension module.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -592,9 +592,16 @@ as a repo-wide figure — for consumer impact, normalize a real corpus through b
 first word wins — so the disclosure would vanish silently, which is the worst direction this
 mechanism can fail in. The checker fails a declining trailer that also claims `<n> rows
 move|merge|split|respell` for non-zero `n`. Quantifying a *zero* is encouraged and passes
-(`none. 0 of 950 rows move`); the pattern requires the count to sit immediately before `rows`,
-which is what keeps #1547's `0 of 950 real cis-allele rows move` from tripping on the 950. It is a
-tripwire for the phrasing this repo actually uses, not a general contradiction detector.
+(`none. 0 of 950 rows move`). **The number that counts is the count, never the denominator** —
+`3 rows of 500,004 move`, `3 of 500,004 rows move` and `3 of 500,004 corpus rows move` all disclose
+three moved rows, and the rule reads three from each. It is a tripwire for the phrasing this repo
+actually uses, not a general contradiction detector — the enumerated forms live on
+`MOVEMENT_CLAIM_RE`, and that docstring is the one to extend when a new phrasing appears.
+
+This sentence used to say "the pattern requires the count to sit immediately before `rows`", and
+that was the #1647 defect rather than a description of it: in `0 of 950 rows move` the number
+immediately before `rows` is the **denominator**, so the very example offered here as passing was
+rejected, and the documented form was a merge blocker. Do not restate the rule positionally.
 
 **The section itself is audited, not just each PR's trailer.** `Changelog grouping audit` renders
 the changelog with the real `release-plz.toml` over `<latest tag>..HEAD` and checks the result
--- a/scripts/check_representation_change.py
+++ b/scripts/check_representation_change.py
@@ -119,17 +119,48 @@ def declines(declaration: str) -> bool:
 #: good decline often quantifies its zero, and `0 rows move over 5,761,302 real
 #: expressions` (#1535) must not trip. Excluded by requiring a nonzero digit *somewhere in
 #: the count* rather than by rejecting a bare `0`, so `0,000` and `000` are zero here too —
-#: an anchored `(?!0\b)` read them as nonzero and failed the decline. **The count must sit
-#: immediately before `rows`**,
-#: which is what keeps `0 of 950 real cis-allele rows move their normalized string`
-#: (#1547) from tripping on the 950 -- a looser pattern fires on that exemplary decline,
-#: measured.
+#: an anchored `(?!0\b)` read them as nonzero and failed the decline.
 #:
-#: It catches the phrasing this repository actually uses -- every quantified disclosure in
-#: the corpus takes this shape -- and it is not a general contradiction detector. Spelled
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

PR #1658, opened 2026-08-11
fix(normalize): refuse an extreme coordinate instead of wrapping the window

### LINKED ISSUE #1487: normalize: an extreme authored coordinate panics with 'attempt to add with overflow' in canonicalize_from_sequence
`canonicalize_from_sequence` pads the member-span union without checking, so a parseable description carrying a coordinate near `i64::MAX` panics instead of being refused.

`src/normalize/merge.rs:2199`:

```rust
let (c_lo, c_hi) = edit_span_union(&edits)?;
let w_lo = (c_lo - CANONICAL_PAD).max(1);
let w_hi = c_hi + CANONICAL_PAD;      // <- overflows
```

## Reproducer

```rust
let p = MockProvider::with_test_data();
let v = parse_hgvs("NM_001234.1:c.[9223372036854775806_9223372036854775807del;10del]").unwrap();
Normalizer::new(p).normalize(&v);
```

```
thread '...' panicked at src/normalize/merge.rs:2199:
attempt to add with overflow
```

The input parses, so this is reached through the ordinary public API. Note the single-member spelling at the same coordinates does **not** panic — `NM_001234.1:c.[9223372036854775807del;9223372036854775806del]` normalizes fine — so it is specific to the span-union path.

## Why it matters beyond the panic

Debug builds panic. Release builds wrap instead, which is worse: `w_hi` becomes negative, the window is nonsense, and the derivation silently proceeds on it rather than refusing.

## Suggested fix

`checked_sub`/`checked_add` on the two pad computations, returning `None` (the function's existing refusal) on overflow. `MAX_CANONICAL_WINDOW` is checked immediately below and would reject such a span anyway, so refusing costs nothing real.

Found while reviewing #1480, which adds a guard further down the same function. That guard now uses checked arithmetic, but it is reached *after* this line, so it cannot mask this.

nh13:
Representation-Change: none. Proven rather than sampled — see **Why this is `none`** below.

Closes #1487

## The crash

A coordinate near `i64::MAX` parses, so this is reachable from `Normalizer::normalize` with nothing unusual in the caller:

```
thread '...' panicked at src/normalize/merge.rs:2609:
attempt to add with overflow
```

Through the Python bindings a Rust panic surfaces as `pyo3_runtime.PanicException`, which subclasses `BaseException` and so escapes `except Exception` — the same contract break `issue_1244_equivalence_overlap_panic` records.

**The release build is the worse half.** It wraps rather than panics: `w_hi` goes negative, the window is nonsense, and the derivation proceeds against it silently. A crash is at least observable.

## Three sites, not the one the issue names

| # | site | why the previous one does not cover it |
|---|---|---|
| 1 | `c_hi + CANONICAL_PAD` | the width test that would refuse an over-wide window runs on the **padded** values, one line too late |
| 2 | `w_hi + frame.delta` | a span near `i64::MAX` but **narrower** than `MAX_CANONICAL_WINDOW` passes the width test, then wraps against a positive `frame.delta` |
| 3 | `enclosing_exon`'s `hi + frame.delta` | runs on the **raw** member union, so (1)'s saturation never reaches it |

Site 3 is the sibling of `crosses_exon_junction`, which already carries `checked_add` and a comment naming this very panic as *"tracked separately"*. That hardening was applied to one of the pair and missed the other; this finishes it.

Sites 2 and 3 were found by the tests, not by reading — each surfaced only after the previous fix landed.

**A fourth site exists, on a different entry point, and is deliberately not fixed here.** `convert::mapper::cds_to_tx_exon_aware` returns `start_tx + offset` unchecked on both early-return branches, and an extreme `c.` coordinate reaches it through `hgvs_to_spdi` rather than through normalization. Measured: under `FERRO_ASSERT_SEQUENCE=1` — the only one of the four seam flags that converts the output — the four extreme-coordinate tests below panic there; the other three flags are green, and CI's `Test oracle` job does not set that flag. So the three sites above are the ones on *this* derivation path, not a claim about the crate. The reproducer is recorded in the test module so the site is not re-derived.

## The fix

The pad **saturates**; the two conversions are **checked**.

Saturating is exact here rather than merely safe: `w_lo` is clamped to `1` on the same line, and a saturated `w_hi` can only make the width test larger — so both saturations land on the refusal that test already gives an over-wide window.

Declining is the correct outcome, not a workaround. Refusing the sequence-first derivation hands the description back to the per-member pipeline, which is what decides whether such a coordinate is an error. Both conversions take the same conservative exit as the neighbouring `start0 < 0` guard.

## Why this is `none`

Not a corpus sample — an argument that covers every input. `saturating_add`/`saturating_sub` and `checked_add`/`checked_sub` are **identical** to `+`/`-` for every value that does not overflow, so any input the old code computed a window for gets a byte-identical window now. The only inputs whose behaviour changes are those that overflowed, and for those the old behaviour was a panic (debug) or a wrapped, meaningless window (release). Nothing that previously produced a valid normalization produces a different one.

Corroborated by the suite: **10,141 passed, 152 skipped, 0 failed**, with `FERRO_MANIFEST` set so the conformance censuses actually ran rather than skipping green.

## Correcting the issue

> the single-member spelling at the same coordinates does **not** panic — `NM_001234.1:c.[9223372036854775807del;9223372036854775806del]` normalizes fine

It does not normalize fine — it panicked identically, at the same line. Member order does not change `edit_span_union`'s result. The test pins both spellings so the claim cannot be re-derived from the issue.

## Tests

`tests/it/issue_1487_canonical_window_overflow.rs` — seven cases: the reported allele; the two-member spelling at the very top of the range that the issue mislabels as single-member; the genuine **lone-member** spelling it meant to name; a **narrow** span at an extreme position (which only site 2 catches); the genomic axis (a different `frame.delta` addend); a parser test pinning why the two additions that remain unchecked *upstream* of the pad cannot overflow; and a negative control asserting an ordinary two-member allele still normalizes.

The five crash tests assert only that control **returns**, deliberately not `is_err()`. Whether such a coordinate is ultimately refused or answered some other way is a behaviour question for the per-member pipeline; the contract this fix owns is that the caller gets a `Result` rather than a panic.

- `cargo clippy --features dev --all-targets -- -D warnings` clean

Previously **REJECTED**, by a panic — measured in both profiles rather than reasoned about, because the two disagree. On the merge-base code, `NM_001234.1:c.[9223372036854775806_9223372036854775807del;10del]` panics in a **debug** build (`attempt to add with overflow`, `src/normalize/merge.rs:2609`), which is what CI runs; in a **release** build it returns `c.[11del;9223372036854775806_9223372036854775807del]`, which is **byte-identical** to what this branch returns. Measured over five inputs (four overflowing, one ordinary control): release output is identical before and after on all five. So no consumer holds a string that moves, and the inputs that changed behaviour changed from *crashing* to answering.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of extreme genomic and coding coordinates during normalization.
  * Prevented crashes and incorrect coordinate wrapping when calculating canonicalization windows and exon boundaries.
  * Normalization now safely declines conversions when coordinates cannot be represented.

* **Tests**
  * Added coverage for overflow scenarios while preserving successful normalization for ordinary variants.

* **Documentation**
  * Clarified coordinate-overflow handling and expected normalization behavior.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2642,8 +2642,17 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
 
     // Window: the union of every member's footprint, padded for the 3'-shift.
     let (c_lo, c_hi) = edit_span_union(&edits)?;
-    let mut w_lo = (c_lo - CANONICAL_PAD).max(1);
-    let mut w_hi = c_hi + CANONICAL_PAD;
+    // Nothing upstream bounds the magnitude of an authored coordinate — the
+    // width test below is the bound, and it runs on the *padded* values — so
+    // the pad itself must not be able to wrap (#1487).
+    //
+    // Saturating is exact here, not merely safe: `w_lo` is clamped to `1` on
+    // the same line, and a saturated `w_hi` can only make the width larger, so
+    // both saturations land on the refusal the width test already gives an
+    // over-wide window. A debug build panicked; a release build wrapped `w_hi`
+    // negative and derived against a nonsense window, which is the worse half.
+    let mut w_lo = c_lo.saturating_sub(CANONICAL_PAD).max(1);
+    let mut w_hi = c_hi.saturating_add(CANONICAL_PAD);
     if w_hi - w_lo + 1 > MAX_CANONICAL_WINDOW {
         return None;
     }
@@ -2691,8 +2700,18 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
         w_lo = w_lo.max(exon_lo);
         w_hi = w_hi.min(exon_hi);
     }
-    let start0 = w_lo + frame.delta - 1;
-    let end0 = w_hi + frame.delta;
+    // Checked for the same reason as the pad above, and not covered by it: the
+    // width test refuses a *wide* window, but a span sitting near `i64::MAX`
+    // while narrower than `MAX_CANONICAL_WINDOW` passes that test and then
+    // wraps against a positive `frame.delta`. Refusing takes the same exit as
+    // the `start0 < 0` guard immediately below, which is already the
+    // conservative answer for a window this axis cannot express (#1487).
+    let (Some(start0), Some(end0)) = (
+        w_lo.checked_add(frame.delta).and_then(|v| v.checked_sub(1)),
+        w_hi.checked_add(frame.delta),
+    ) else {
+        return None;
+    };
     if start0 < 0 {
         return None;
     }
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

nh13:
@coderabbitai review

PR #1660, opened 2026-08-11
docs(readme): state that rule 3 governs rule 2's evaluation basis

nh13:
Follow-up to #1605, from review feedback on the ruleset it shipped: rules 2 and 3 can be read as demanding opposite things, and the section that classifies a miss puts this one in the wrong box.

## The conflict

Rule 2 says a preference clause outranks maintainer judgment. Rule 3 says every rule is evaluated over the resulting sequence, never over the input's spelling. Both are *Best effort*, and no precedence is stated between them — so where a preference clause keys on what rule 3 excludes, the ruleset does not say which gives way.

It is not hypothetical. Four clauses, all matching rule 2's own scope vocabulary, key on provenance or on population knowledge:

| clause | text | keys on |
|---|---|---|
| `RNA/delins.md:41` | "**preferred** when either of the two variants is known as a frequently occurring variant" | population frequency |
| `RNA/repeated.md:33-34` | "when the repeat is variable in the population … is **preferred** over" | population variability |
| `protein/repeated.md:22-23` | same shape | population variability |
| `DNA/delins.md:79-84` | "should **preferably** be described individually" — because "the two variants may have been reported (or might occur) individually" | the input's decomposition |

## The mislabelling, which is the part that is wrong rather than merely unstated

The subsection closed with:

> Only a failure caused by the spec not determining an answer falls under best effort — and that triggers rule 5.

The spec **is** determinate on all four. Ferro's code is not at fault either. So by the ruleset's own dichotomy a miss is a **bug under rule 7** — which contradicts two decided ledger records. `canonical-form-choice-when-both-legal` records these clauses under its counter-evidence and rules that ferro "cannot implement them and does not try … deliberately, not in ignorance of the other half". `separation-is-a-property-of-the-spelling-not-of-the-variant` carries `DNA/delins.md:79-84` in `deviates_from`.

So the position was already decided; the canonical statement of the rules just did not carry it.

## What changed

- **Rule 2** defers to rule 3 where the two meet.
- **Rule 3** says what "resulting sequence" includes. This was the second half of the feedback, and it corrects a real over-narrowing: ferro consults the transcript model, exon boundaries, reading frame, strand and topology, and must — `general.md:36` says so of the codon exception ("one needs to know whether the two variants are in a coding sequence"), and `exon-junction-dup-converge-from-the-far-side` turns on two spellings that denote one transcript sequence yet project 2,790 bp apart. Those are functions of the accession, not of the description, so they are context rather than spelling.
- **A fifth cause** of best-effort failure, and **a third outcome** beside bug and best effort: the *declared deviation*, permanent and recorded in `deviates_from`. Where such a clause also contradicts one forbidding the use of provenance (`general.md:157-160`), that contradiction stays a defect and rule 5 still applies to it.
- **One new subsection**, `### What rule 3 excludes`: a table of what a description carries and how each part is treated — the answer is one degree of freedom, the partition, plus the run-position choice that feeds it — and the worked example on the coordinates `separation-is-a-property-of-the-spelling-not-of-the-variant` was decided on.

The section adds 49 lines and is written to the register of the ones around it: the table replaces a paragraph, and the two written gaps are annotated in the code block rather than explained in prose. The example states the **decided target** only. It deliberately does not describe current behaviour on those two spellings — that is #1617's, and it would be false the day #1617 lands. `### Known limitation` sits immediately below it and already says the rules are enforced intent rather than a claim of completeness.

## Drift guard

`readme_normalization_rules.rs` gains `the_rule_3_example_is_the_case_the_adjudication_test_pins`: the README's window, both spellings and the decided output must all still appear in `tests/it/cis_confluence_adjudication.rs`, and the link between them must resolve.

The existing guards cover cross-*links*, which break loudly when an anchor moves. An example is the weaker case — the README can go on quoting coordinates long after the test stopped covering them and both documents still read as though they agree. Token-based rather than a diff, because the two texts *should* differ; only the case has to stay shared.

No rule was renumbered, reworded at its opener, or moved between the two halves, so every existing citation by number still resolves and the other five guards pass untouched.

## Second commit: rule 6 called the 3'/5' knob an orthogonal axis

Rule 6 says there are no user options for normalization form, then names 3'/5' direction as an orthogonal axis that stays available. It is the most form-determining option there is — the ledger quotes its own measurements per direction (`3': 2,697 to 1,710; 5': 2,669 to 1,682`) — so as written the rule permits exactly what its first half refuses: confluence as a property of *(input, direction)* rather than of an input.

Now says what is true: error mode is orthogonal, the direction knob is not, and rules 2 and 3 are claimed **per direction** rather than across the two.

**Deliberately not adjudicated:** whether 5' output is conformant at all under `general.md`'s 3' rule. That is a real and separate question, the CLI makes no claim either way today, and answering it here would be the same over-reach this section is being corrected for. Noted on #1669.

The other defect in that same sentence — rule 6 forbidding user options while an uncommitted ruling gave partition preservation a "declared opt-in mode" — is an operator decision and is filed as **#1669**, not folded in here.

## Verification

- `cargo nextest run --features dev -E 'test(readme_normalization_rules)'` — 6/6 pass, including the new guard
- `cargo run --features dev --example generate_tool_support_tables -- --check` — exit 0, generated README tables untouched
- Every clause quoted above verified against `assets/hgvs-nomenclature` @ `6f85311`

Representation-Change: none. Documentation plus one integration-test assertion; nothing under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched and no expectation was re-blessed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,10 +181,13 @@ ferro's normalizer follows four rules about its output, and three about how it h
 2. **Recommended form.** Where the spec prefers among conformant forms, ferro produces it.
    *Best effort.*
    Scope: "recommended", "preferred", lowercase "should", the 3' rule. A preference clause outranks
-   maintainer judgment.
+   maintainer judgment, but not rule 3: where it is not evaluable on what a normalizer holds, rule 3
+   governs.
 
 3. **Confluent.** Inputs denoting one variant produce one output. *Best effort.*
-   Every rule is evaluated over the **resulting sequence**, never over the input's spelling.
+   Every rule — rule 2 included — is evaluated over the **resulting sequence**, never over the
+   input's spelling. Reference context counts as sequence: transcript model, exon boundaries,
+   reading frame, strand and topology are functions of the accession, not of the description.
 
 4. **Deterministic.** Same input, same output. *Absolute.*
    Note that 4 does not imply 3 — a deterministic normalizer can be arbitrarily non-confluent.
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1661, opened 2026-08-11
fix(error_handling): keep the repeat unit when the W3013 range disagrees with it

### LINKED ISSUE #1631: W3013 repeat-label repair drops the unit and keeps the range, changing the variant when their lengths differ
The W3013 `RedundantRepeatLabel` repair drops the repeat **unit** and keeps the **range**. That preserves the variant only when the range's length equals the unit's length. When they differ, the repaired description denotes a different variant, and the repair is silent about it.

## Measured

```
NM_024312.4:r.-125_-123cug[4]  --error-mode lenient -> NM_024312.4:r.-125_-123[4]   CORRECT (range 3 nt == unit 3 nt)
NM_024312.4:r.-6_-3g[6]        --error-mode lenient -> NM_024312.4:r.-6_-3[6]       WRONG   (range 4 nt, unit 1 nt)
```

`RNA/repeated.md:20` fixes the reading of a bare range:

> **NOTE**: while `r.123cug[23]` describes a repeat of 23 `cug` units, `r.123_125[23]` describes a tri-nucleotide repeat of 23 units

So a bare range denotes **one unit**, whose length is the range's length. `r.-6_-3[6]` therefore means a **4-nucleotide** unit repeated 6 times — 24 nucleotides — while the input `r.-6_-3g[6]` means a single `g` repeated 6 times, 6 nucleotides. The repair changed the variant by a factor of four in length.

`--error-mode silent` produces the same repaired string with an **empty** warning vector, so on that path there is nothing at all to tell a caller the variant moved.

## The correct repair drops the range, not the unit

The redundancy `RNA/repeated.md:22` names is that a range and a unit state the same fact twice. When they disagree, the unit is the one that carries the variant, because the range only pins the unit's *length*. So the conformant repair for `r.-6_-3g[6]` is `r.-6g[6]`, which ferro already parses and leaves alone in every mode.

`NM_024312.4` carries `gggg` at `r.-6` to `r.-3`, so `r.-6g[6]` and the input denote the same six-`g` run; `r.-6_-3[6]` does not.

## Relationship to the ruling

`rulings[rna-repeat-range-plus-unit-redundancy]` is now **decided**: `RNA/repeated.md:22` governs, so a description carrying both a range and a unit is invalid on an RNA reference. That ruling settles *whether to repair*. It deliberately does **not** settle *what the repair should produce*, which is this issue — the ruling record says so in as many words.

Suggested acceptance: `parse_hgvs_with_config("NM_024312.4:r.-6_-3g[6]", ErrorConfig::lenient())` yields `NM_024312.4:r.-6g[6]`, and a repair that cannot preserve the variant refuses rather than guessing.

## Representation impact

Real: every input on this repair path whose range length differs from its unit length moves, and the current output for those is the wrong variant.

nh13:
Closes #1631

The W3013 `RedundantRepeatLabel` repair dropped the repeat **unit** and kept the **range**, unconditionally. `RNA/repeated.md:20` fixes the reading of a bare range — "`r.123_125[23]` describes a tri-nucleotide repeat of 23 units" — so a bare range denotes one unit whose length is the range's length. Dropping the unit therefore preserves the variant only while the range's span equals the unit's length.

Measured on `main`, and the failing test was written first:

```
NM_024312.4:r.-125_-123cug[4]  --error-mode lenient -> r.-125_-123[4]   correct (3 nt == 3 nt)
NM_024312.4:r.-6_-3g[6]        --error-mode lenient -> r.-6_-3[6]       WRONG   (4 nt vs 1 nt)
```

`r.-6_-3g[6]` is a single `g` six times, 6 nucleotides. `r.-6_-3[6]` is a four-nucleotide unit six times, 24 nucleotides. The repaired description denoted four times the input's bases, and `--error-mode silent` produced it with an **empty warning vector**, so on that path nothing said the variant had moved.

## What was chosen

**Repair differently, not refuse** — and it is not a choice between two legal forms, which is why no adjudication is being made here. Dropping the range is variant-preserving on every input this detector matches, because the unit and the anchor together already state the variant and the range only restates the unit's length. So there is no residue for a refusal to cover:

- **lengths agree** — unchanged. `r.100_102cug[4]` → `r.100_102[4]`, `:22`'s own exemplar and the redundancy it names.
- **lengths differ** — `r.-6_-3g[6]` → `r.-6g[6]`. The unit survives and the range is reduced to its anchor, which is the issue's suggested acceptance and the form ferro already parses and leaves alone in every mode.

Both arms stay W3013: the input defect is identical, only the surviving half differs. Strict still refuses the input outright — the pending `rna-repeat-range-plus-unit-redundancy` ruling settles *whether* to repair and says in as many words that it settles neither what the repair produces nor which entry point applies it. Only the repaired string moves.

Range spans are counted as **positions**, not by arithmetic: `c.`/`r.` numbering has no position zero, so `r.-2_1` is three positions and `r.-2_2` is four. `a_range_crossing_the_utr_boundary_counts_positions_not_arithmetic` pins both sides of that.

## The worked-example row stops diverging, and the reason is not the obvious one

`tests/fixtures/spec-worked-examples/cases.json` pinned this row's lenient answer as `r.-6_-3[6]` and recorded it as the file's one divergence. It is now `r.-6_-3g[6]` — the spec's own published string — so the divergence set drops to one member and the spec-conformant census goes 10 → 11.

**What puts the range back is the normalizer, not the repair.** `NM_024312.4` carries `gggg` at `r.-6` to `r.-3`, so `normalize_repeat`'s tract maximization widens the repaired `r.-6g[6]` straight back onto that tract. Two consequences worth stating rather than leaving to be re-derived: the agreement with `:27` is not a conformance verdict on the string, since `:22` still calls range-plus-unit invalid; and the re-widening is the #1618 family (what a repeat's copy count is counted against) surfacing on the RNA axis. `the_lenient_repair_hands_the_normalizer_the_anchored_form` pins the two halves separately so a future change is attributable to the right one.

## Why the denoted-sequence oracle never saw this

`FERRO_ASSERT_SEQUENCE` compares the normalizer's input against its output. The preprocessor runs *before* the normalizer, so on the old behaviour the oracle compared `r.-6_-3[6]` against `r.-6_-3[6]` and agreed. A repair that changes the variant is invisible to it by construction — the oracle guards the normalizer, and nothing guards the repair.

## Verification

- `cargo nextest run --features dev` — **`Summary [200.546s] 10144 tests run: 10144 passed (1 slow), 152 skipped`**.
- Reproduction control: the 8 new tests were run against `origin/main`'s `src/error_handling/{corrections,preprocessor}.rs` with everything else on this branch — **4 fail**, on the length-mismatch arms only.
- All four seam oracles on (`FERRO_ASSERT_SEQUENCE`, `_IDEMPOTENT`, `_REPARSE`, `_IN_BOUNDS`), full suite, this branch against `origin/main` at `b9f6cecb`: **25 failures on each side and the failing test-name sets are byte-identical**. No oracle fire is added or removed.
- `cargo fmt --all --check` clean; `cargo clippy --features dev --all-targets -- -D warnings` clean.

Representation-Change: 1 pinned (row, mode) result moves — `NM_024312.4:r.-6_-3g[6]` under `--error-mode lenient` and `--error-mode silent`, from `r.-6_-3[6]` to `r.-6_-3g[6]`. Previously **accepted**, so it is a real migration for a consumer storing that string — but the previously stored string denoted a different variant (24 nt against 6 nt), so the migration is a correction rather than a re-spelling. The affected family is exactly the W3013 repair path where the range's span differs from the unit's length; the equal-length arm is unchanged and pinned. Measured by the full suite (one pinned row moved, nothing else) and by the four-oracle A/B above. `examples/dump_normalized_corpus.rs` is **structurally incapable** of measuring this and was not run for it: it parses with `parse_hgvs`, which applies no `ErrorConfig`, so no corpus row ever reaches the preprocessor and a `0 moved` from it would be a fact about the harness.


nh13:
@coderabbitai pause

PR #1662, opened 2026-08-11
fix(spdi): refuse a genomic pter/qter/cen instead of flattening it to base 0

### LINKED ISSUE #1643: hgvs_to_spdi flattens pter/qter/cen to base 0, the sibling of the dropped genomic offset
`GenomePos::special` (`pter`, `qter`, `cen`) flattens to `base: 0` when converted through `hgvs_to_spdi`, so two descriptions naming different special positions — or a special position and literal base 0 — can produce the same SPDI triple.

This is the same failure mode as #1628, in the same helper. #1628 was the dropped `offset` field; this is the unrepresented `special` field. That fix (PR #1641) added `reject_genomic_offset` to the six genomic arms of `src/spdi/convert.rs`; the `special` case was left out of scope deliberately and is filed here.

**Currently guarded in the wrong place.** PR #1622's denoted-sequence oracle guards `pter`/`qter`/`cen` at the oracle, not at the conversion — so the oracle is safe but every other `hgvs_to_spdi` caller is not.

**Suggested fix,** consistent with what #1641 established: refuse rather than flatten, since a genomic special position has no literal coordinate to convert to. `reject_genomic_offset` is the natural place, or a sibling guard called alongside it.

**Wider context.** `hgvs_to_spdi` is a second, independent interpreter of a description with nothing pinning it to the normalizer's interpretation. #1618 (repeat copy-count basis), #1619 (exon-gap coordinate frame), #1628 (dropped offset) and this issue are four mechanisms of that one structural fact. They do not share a fix, but an invariant test asserting the two interpreters agree would catch the whole family.

nh13:
Closes #1643

`GenomePos` stores `base: 0` beside a `special` marker because `pter`, `qter` and `cen` name landmarks of the assembled chromosome — and `cen` of its centromere annotation — not positions on the sequence. `hgvs_to_spdi` read `base` and never looked at `special`, so every special position resolved to base 0. Same helper as #1628, same shape: a field the AST carries that the conversion discards.

## What reproduces, and one correction to the issue

Measured on `main` at `b9f6cecb`, on `NC_000001.11`:

```
g.10_qterdelACGTACGTAC  -> NC_000001.11:9:ACGTACGTAC:
g.10_cendelACGTACGTAC   -> NC_000001.11:9:ACGTACGTAC:
g.10_19delACGTACGTAC    -> NC_000001.11:9:ACGTACGTAC:
```

A deletion running to the q-arm telomere, one running to the centromere, and a literal ten-base deletion — three descriptions of three different things, one triple. `g.10_qterdupACGTACGTAC` is worse still: it converts to `NC_000001.11:0::ACGTACGTAC`, anchoring the duplication at position 0.

**The issue's reproducer as written does not reproduce, and the reason matters.** `g.pter_10del` and `g.10_qterdel` are already refused today — but *incidentally*, by checks asking a different question, each of which diagnoses the wrong thing:

| shape | refused by | message |
|---|---|---|
| `g.pter_10del` | the 1-based position check | `position 0 is not valid in HGVS` |
| `g.10_qterdel` | the reference fetch | `invalid 1-based interval [10, 0] for reference fetch` |
| `m.10_qterdel`, `o.10_qterdel` | the circular-wraparound check | `Cannot convert wraparound m. variant to SPDI` — a wraparound that is not there |

None of them is a guard, and the one shape that reaches none of them is the one that converts: **a description that spells its own bases** needs no provider fetch and no position resolution. That is why the class outlived #1641, and why "it already errors" was not coverage. Every descriptor in the new test spells its bases for exactly this reason.

## The fix

`reject_genomic_offset` becomes `reject_unresolvable_genomic_position` and refuses both fields at the same six genomic call sites. They are one defect in two fields of one struct and they decline for the same reason — there is nothing on a genomic accession to resolve either against — so a sibling guard would have been two functions with one docstring.

The verdict is `InvalidPosition`, following what #1641 established: it differs from the transcript axes' `MissingReferenceData` because those decline a *well-formed* position SPDI cannot carry, while `g.qter` is not a resolvable position at all. `a_transcript_axis_offset_declines_for_its_own_reason` still passes unchanged, which is the control that the genomic guard has not reached the axes where these fields are legitimate.

## Verification

- `cargo nextest run --features dev` — **`Summary [134.888s] 10138 tests run: 10138 passed, 152 skipped`**. Nothing re-blessed: for a change to what `to_spdi` answers, an untouched expectation set is the strong signal.
- Reproduction control: with the four added lines of guard deleted and nothing else changed, both new tests fail — `converted without a provider to NC_000001.11:9:ACGTACGTAC:; the special position was flattened onto base 0`, and `g.10_cendelACGTACGTAC and g.10_qterdelACGTACGTAC are different variants but share the triple`.
- All four seam oracles on, full suite, against `origin/main` at `b9f6cecb`: **25 failures on each side, failing test-name sets byte-identical**. This closes none of the two rows keeping `FERRO_ASSERT_SEQUENCE` out of `test-oracle` (#1618 and #1619) and adds none.

Representation-Change: none. Four paths under `src/normalize/` DO reach `hgvs_to_spdi` — `Normalizer::to_spdi` calls it directly, `apply_to_reference` and `canonical_spdi` reach it through `spdi::apply::variant_edit_triples`, and the debug-only `assert_denoted_sequence` reaches it through `compare_denoted_sequences` — so an earlier "nothing calls it" premise is withdrawn. None of the four produces a normalized description: the first three are separate public entry points returning an SPDI triple or an `AppliedVariant`, reached only from `src/python.rs` and `src/equivalence/key.rs`, and the fourth is compiled out of release. The normalization pipeline itself makes no `spdi::` call. The only producer of a `special` position is the parser — projection builds every genomic position with `GenomePos::new`. What changes is `to_spdi`'s answer for descriptions whose position has no coordinate to convert, from a triple that answers for a different variant to an error. Verified rather than asserted: 10,138 tests pass with nothing re-blessed. `examples/dump_normalized_corpus.rs` was not used as evidence — it dumps *normalized* output and no family builds a `pter`/`qter`/`cen` position, so a `0 moved` from it would be a fact about the corpus rather than about this change; the suite is the evidence.

Previously **accepted**, and measured empty. The rejected-vs-accepted fact belongs in this trailer and was missing: a genomic `pter`/`qter`/`cen` description that spells its own deleted bases *did* convert before this change and now errors — `NC_000001.11:g.10_qterdelACGTACGTAC` returned `NC_000001.11:9:ACGTACGTAC:` at `b9f6cecb` and returns the landmark error here. So the class is real, not hypothetical. Its measured size is **0 of 4,188,436** ClinVar expressions (`clinvar_hgvs_unique`): the only **2** rows carrying a genomic `qter` — `NC_000009.12:g.12891379_qterdelins[T;NC_000020.11:g.47204889_qterinv]` and `NC_000013.10:g.114819939_qterdelins[96729864_114814234inv;96735632_104289803]` — already failed conversion at `b9f6cecb` on their non-literal insert payload, so for those the only change is which error is returned. Measured by swapping `src/spdi/convert.rs` for its merge-base version and re-running the same five inputs, rather than by reasoning about reachability.

### DIFF
--- a/src/equivalence/checker.rs
+++ b/src/equivalence/checker.rs
@@ -405,6 +405,64 @@ impl<P: ReferenceProvider> EquivalenceChecker<P> {
     /// refuse pairs the checker answers today on no evidence that any of those
     /// answers is wrong. Add it when a wrong answer is measured, with the pair
     /// that measures it.
+    ///
+    /// **A third shape now shares the SPDI blindness and is deliberately not
+    /// listed either: a genomic `pter`/`qter`/`cen` description (#1643).**
+    /// `hgvs_to_spdi` used to flatten those onto `base: 0` and convert; it now
+    /// refuses, so [`Self::edit_triples`]'s `.ok()?` swallows the new error and
+    /// the `SequenceMatch` rung stops firing for them. The reach is narrower
+    /// than it looks, and the reason is worth writing down because it is not
+    /// visible from this file: both rungs run on the **normalized** forms, and
+    /// normalization *resolves* a marker against the provider
+    /// (`g.10_qterdelACGTACGTAC` -> `g.10_28del` on a 28-base contig), so most
+    /// of the class never reaches the sequence rung carrying one at all.
+    ///
+    /// **Which markers survive normalization is a property of the PROVIDER, not
+    /// of the marker.** `cen` always does — `resolve_special_genome_pos` returns
+    /// `Ok(None)` for it unconditionally, there being no centromere annotation
+    /// on a sequence accession to resolve against. But `pter` and `qter` survive
+    /// too whenever the length lookup fails: that arm falls through to
+    /// `canonicalize_genome_variant`, which rewrites only the edit body and
+    /// clones the location, so the marker is echoed. Measured, on a provider
+    /// holding no sequence for the accession: `g.10_qterdelACGTACGTAC` ->
+    /// `g.10_qterdel` and `g.pter_10delACGTACGTAC` -> `g.pter_10del`. So this is
+    /// "`cen` plus any marker whose reference is unservable", not "`cen`".
+    ///
+    /// **One verdict class moved, and the old answer was wrong.** Both figures
+    /// below come from disabling the guard in place and re-running, and they are
+    /// **two separate corpora** — do not read either denominator onto the other:
+    ///
+    /// | corpus | pairs | verdicts that moved |
+    /// |---|---|---|
+    /// | 22 mixed `cen`/`qter`/`pter` descriptors, on a serving and a non-serving provider | 231 x 2 | **0** |
+    /// | 10 descriptors whose 3' anchor is `cen`, serving provider | 45 | **3** |
+    ///
+    /// The three are what the first corpus could not reach, and they say why: a
+    /// `cen` on the END of a *duplication* is what zeroes the triple, because
+    /// the `Duplication` arm emits `end_one_based` as the SPDI position. So
+    /// `g.10_cendupACGTACGTAC`, `g.12_cendupACGTACGTAC` and
+    /// `g.4_cendupACGTACGTAC` all converted to `NC_000001.11:0::ACGTACGTAC`, and
+    /// each of the three pairs among them answered `SequenceMatch` — **wrong**,
+    /// since they duplicate different spans. All three now answer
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1663, opened 2026-08-11
fix(ci): let the failure reporter resolve the repository it files against

nh13:
Found while triaging a flaked `Code Coverage` run on #1660: underneath the coverage failure sat a second red X, and it was the job whose whole purpose is to report the first one.

## The reporter has never worked

`report-failure.yml` is the reusable job **six** workflows call — `coverage`, `external-validation`, `fuzz`, `nightly-mutalyzer`, `nightly-perf`, `audit-pr-runs`. It exists because those workflows only run after a merge, so their red X cannot block anything; the tracking issue is the entire alerting path.

`gh issue list|comment|create` are repo-scoped commands. With no `--repo` and no `GH_REPO`, `gh` works out which repository it means by shelling out to `git` in the working directory. The job has no `actions/checkout` — it only touches the API — so the lookup has nothing to resolve against:

```
failed to run git: fatal: not a git repository (or any of the parent directories): .git
##[error]Process completed with exit code 1
```

Observed on [run 31517344115](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/31517344115), where it fired correctly on a genuine `Coverage` failure and then died instead of filing.

The reason it went unnoticed is structural: **a reporter that fails exactly when it is needed is indistinguishable from one that is never needed.** It is only ever reached on someone else's failure, and its own red X reads as part of that failure.

## The fix

`GH_REPO: ${{ github.repository }}`. It supplies the target directly and skips the git lookup — cheaper than adding a checkout to a job that never reads the tree, and it covers all three `gh` calls at once rather than three `--repo` flags.

## The guard is for the class, not this file

A test asserting `report-failure.yml` sets `GH_REPO` would be worth almost nothing: it goes green the moment the line lands and says nothing about the next workflow that shells out to `gh`. Seven do today.

`tests/it/workflow_gh_repo_context.rs` walks every workflow and fails any job that issues a repo-scoped `gh` command with no `--repo`, no `GH_REPO` in scope (workflow, job or step `env`), and no `actions/checkout`.

Three things it does deliberately:

- **Parses rather than greps.** The question is structural — which steps sit in which job, which `env` blocks are in scope. A substring scan also matches the `gh pr close … --repo <owner>/<repo>` text inside `audit-pr-runs.yml`'s reproduction *hint*, which is documentation, not a command. Same reasoning `generator_completeness.rs` gives for parsing `Cargo.toml`.
- **Handles `gh api` separately.** An absolute path (`/repos/OWNER/REPO/...`) needs no resolution; the `{owner}`/`{repo}` placeholder form needs it exactly as much as `gh issue` does. `audit-pr-runs` and `prune-ci-caches` are correct today only because they interpolate the repo into the path — that is why they are not flagged, and the placeholder form is checked so it stays that way.
- **Joins backslash continuations and drops comment lines.** `--repo` habitually lands on a later line than the subcommand, and `report-failure.yml` explains its own `gh` usage in a comment directly above the call.

It also asserts it scanned something. A guard over zero invocations passes, and would keep passing after a refactor moved every call out of the shape it matches.

`the_failure_reporter_names_its_repository` pins the reporter by name on top of that, since it is the one workflow whose value is being reachable when something else has already broken.

## Verified in both directions

The scan claim is only worth what its negative control is worth:

| state | result |
|---|---|
| with the fix | both tests pass |
| fix reverted | both fail; the scan names all three `gh issue` calls as `report-failure.yml / job \`report\`` |

- `cargo nextest run --features dev -E 'test(report_failure_degrades) or test(workflow_gh_repo_context)'` — 5/5; with the lookup wrapping removed, only `a_failed_lookup_still_files_an_issue` fails, with exit 1 and nothing filed
- `actionlint .github/workflows/report-failure.yml` — clean

## Second commit: the reporter had no failure mode of its own

Found while retro-ing the first commit, and the same shape as the bug it fixes. The step's *first* act is the rolling-issue search, under `set -e`, so any transient failure of it — rate limit, API blip — aborted before the `gh issue create` fallback. The alerting path would die underneath the failure it exists to announce.

Only the **lookup** is made forgiving: a failed search warns and degrades to "no rolling issue found", because a duplicate issue beats silence. `gh issue comment` / `gh issue create` failing still reddens the step — at that point nothing was recorded anywhere and the red X is the last signal.

**Guarded by running the script, not by reading it.** Asserting the shape of the file cannot answer what the script does. `tests/it/report_failure_degrades.rs` extracts the `run:` block from the workflow and executes it under bash with a stub `gh` on `PATH`:

| lookup | required behaviour |
|---|---|
| fails outright | warn, then **create** |
| succeeds, no match | **create** |
| succeeds, match `4242` | **comment** on 4242, and not also create |

Extracted rather than copied — a copied script keeps passing after the real one changes.

## Evidence the reporter has never once filed

Independent of the code: the footer every filed issue would carry — *"One rolling issue is used per failure mode; repeat occurrences are added as comments."* — appears in **zero** issues in this repo, open or closed. The `gh` mechanism itself was also checked directly from a non-git directory:

```
control   (no GH_REPO): exit=1, "failed to run git: fatal: not a git repository"   ← the shipped bug, verbatim
treatment (GH_REPO set): exit=0
```

## What this does not fix

Whether the six callers pass the right `title`/`summary`, and whether one rolling issue per failure mode is the right granularity. Untouched. Worth noting that because the job has never completed, **no caller's inputs have ever been exercised** — including the requirement that `title` be stable per failure mode, which is the rolling-issue lookup key.

Representation-Change: none. CI workflow and one integration test; nothing under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched.

### DIFF
--- a/.github/workflows/report-failure.yml
+++ b/.github/workflows/report-failure.yml
@@ -71,6 +71,17 @@ jobs:
       - name: Open or update the tracking issue
         env:
           GH_TOKEN: ${{ github.token }}
+          # `gh issue list|comment|create` are repo-scoped commands: with no
+          # `--repo`, `gh` resolves the target by shelling out to `git` in the
+          # working directory. This job has no `actions/checkout` — it only
+          # touches the API — so that lookup died on every invocation with
+          #
+          #     failed to run git: fatal: not a git repository
+          #
+          # and the reporter exited 1 without filing anything. `GH_REPO` supplies
+          # the target directly and skips the git lookup, which is cheaper than
+          # adding a checkout for a job that never reads the tree.
+          GH_REPO: ${{ github.repository }}
           TITLE: ${{ inputs.title }}
           SUMMARY: ${{ inputs.summary }}
           HINT: ${{ inputs.hint }}
--- a/.github/workflows/report-failure.yml
+++ b/.github/workflows/report-failure.yml
@@ -100,9 +100,26 @@ jobs:
           # interpolating it into the jq program: a title containing a quote
           # would otherwise produce a malformed filter, and the value is
           # workflow-supplied input.
-          EXISTING=$(gh issue list --state open --search "$TITLE in:title" \
+          #
+          # The lookup must not be able to abort the step. It runs under
+          # `set -e` and is the FIRST thing this job does, so any transient
+          # failure — a rate limit, a search-API blip — kills the reporter
+          # before the `gh issue create` fallback below is ever reached, and
+          # the alerting path fails silently underneath the failure it exists
+          # to announce. That is the exact shape of the defect that left this
+          # job broken. Filing a duplicate is strictly better than filing
+          # nothing, so a failed lookup degrades to "no rolling issue found"
+          # and says so in the log.
+          #
+          # Only the LOOKUP is forgiving. If `gh issue comment` or `gh issue
+          # create` fails, nothing was recorded anywhere and the step must go
+          # red — that is the last remaining signal.
+          if ! EXISTING=$(gh issue list --state open --search "$TITLE in:title" \
             --json number,title \
-            --jq '[.[] | select(.title == env.TITLE)] | .[0].number // empty')
+            --jq '[.[] | select(.title == env.TITLE)] | .[0].number // empty'); then
+            echo "::warning::rolling-issue lookup failed; filing a new issue instead of commenting"
+            EXISTING=""
+          fi
           BODY=$(printf '%s\n' \
             "$SUMMARY" \
             "" \
--- a/tests/it/main.rs
+++ b/tests/it/main.rs
@@ -474,6 +474,7 @@ mod uta_loader_tests;
 mod variantvalidator_tests;
 mod web_service_tests;
 mod weight_bound_worked_examples;
+mod workflow_gh_repo_context;
 
 mod reported_confluence_pairs;
 mod ring_segment_wellformedness;
--- a/tests/it/main.rs
+++ b/tests/it/main.rs
@@ -476,5 +476,6 @@ mod web_service_tests;
 mod weight_bound_worked_examples;
 mod workflow_gh_repo_context;
 
+mod report_failure_degrades;
 mod reported_confluence_pairs;
 mod ring_segment_wellformedness;
--- a/tests/it/report_failure_degrades.rs
+++ b/tests/it/report_failure_degrades.rs
@@ -1,0 +1,183 @@
+//! The failure reporter must file something even when its lookup fails.
+//!
+//! `report-failure.yml` runs under `set -euo pipefail` and opens by searching
+//! for the rolling issue to comment on. That search is a network call to the
+//! GitHub search API, and it is the *first* thing the job does — so before this
+//! was guarded, any transient failure of it (rate limit, API blip, a title the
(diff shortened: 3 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
Both lenses ran; four findings applied in `cf637477`. The PR itself is sound — I mutation-verified both of its halves before reviewing anything:

- removing `GH_REPO` reddens `repo_scoped_gh_calls_can_resolve_their_repository` **and** `the_failure_reporter_names_its_repository`;
- reverting the lookup to a fatal assignment reddens `a_failed_lookup_still_files_an_issue`, and only that one.

I also checked the two factual claims in the new workflow comment rather than taking them: the `report` job has exactly one step and no `actions/checkout`, and its `run:` opens with `set -euo pipefail`. Both true.

## The CLI lens found two, both `major`, both the same shape — an aggregate standing in for a per-position value

**`--repo` was judged per LINE, not per invocation.** `gh pr view --repo x/y && gh issue create` carries the flag on the line but supplies nothing to the second command, and a whole-line `contains("--repo ")` cleared both. Now split on `&&`/`||`/`;`/`|` and judged per segment. The split is coarse — it does not understand quoting — but it errs toward *more* scrutiny, which is the safe direction: a split inside a quoted string can only produce a piece that fails, never one that passes wrongly.

**`has_checkout` was `steps.iter().any(...)`**, so a `gh` call in a step *before* the checkout was cleared by a checkout that had not run yet. Now tracked in step order.

## The skill lens found two more

**`REPO_SCOPED` omitted `secret`, `variable` and `ruleset`** — all repo-scoped in `gh`, and the plausible next additions. A name missing there fails in the silent direction: a future checkout-less step using it passes the scan and dies at runtime, which is exactly what this file exists to prevent. (Measured what the workflows call today: `issue` 12, `api` 11, `pr` 8, `run` 1, `release` 1 — all five already covered, so this is future-proofing rather than a live gap.)

**`env_sets_gh_repo` tested `contains_key`, so `GH_REPO: ""` passed** — the key set, `gh` left with nothing to resolve, and the step dying exactly as the unset case did. Now requires a non-empty value, with `${{ … }}` accepted since its value is not knowable from the YAML.

## Both CLI findings were LATENT, not live, and I said so before fixing them

Measured: no `run:` line in any workflow has two `gh` calls (the only such line lives in an `audit-pr-runs.yml` `with:` value, which the parsed scan never reads), and no job runs `gh` before a later checkout. So neither was firing wrongly today. They are fixed because a guard whose purpose is to catch the *next* occurrence must not be satisfied by there being none today — and both are pinned on synthetic input for the same reason, since the real workflows cannot hold the broken arrangement and stay green.

Three tests added: `a_repo_flag_on_one_command_does_not_clear_its_neighbour` (which also asserts the whole-line reading *does* wrongly clear it, so the fix's necessity is pinned, not just its effect), `an_empty_gh_repo_does_not_satisfy_the_scan`, and the `a_table_with_no_repeats`-style negative control inside the first.

Validated at CI parity: `fmt` clean, `clippy --all-features --all-targets -D warnings` clean, plain **10151/10151**, oracle (three flags, `ORACLE_EXCLUDE` negated) **10089/10089**.

Issue #960, opened 2026-07-06
fix(spdi): r.-N (5'UTR) conversion rejected while c.-N succeeds — route r.-N through the exon-aware mapper


nh13:
## Summary

SPDI conversion handles `r.*N` (3'UTR) but rejects the symmetric `r.-N` (5'UTR), while `c.-N` succeeds — an asymmetry the HGVS numbering rule forbids.

#944 (PR #950) routed `r.*N` through the exon-aware `CoordinateMapper::cds_to_tx` so it agrees with `c.*N`. The same spec rule applies to the 5'UTR: `docs/background/numbering.md:58` ("`r.123` relates to `c.123`") and `:61` (RNA numbering follows the coding DNA reference, which includes `c.-N`). But `resolve_rna_pos` (`src/spdi/convert.rs`, ~line 1032) still short-circuits `r.-N` to the final `ensure_positive_tx`, and since 5'UTR positions have `base < 1` it is rejected as "resolves to a non-positive base". `c.-N` maps correctly via `cds_to_tx`'s `base < 1` branch (`src/convert/mapper.rs`, `cds_start - N`, exon-aware).

This is pre-existing — #950 fixed the 3'UTR sibling only; it did not introduce the `r.-N` rejection.

## Repro

On a transcript with a 5'UTR, `hgvs_to_spdi` for `r.-3del` errors while `c.-3del` on the same transcript succeeds and yields a valid SPDI position.

## Fix sketch

In `resolve_rna_pos`, add a 5'UTR branch mirroring the 3'UTR one: when the position is non-intronic, not `utr3`, and `base < 1` (i.e. `is_5utr()`), require `transcript.cds_start` (error `MissingReferenceData` if absent), map via `CoordinateMapper::cds_to_tx(&CdsPos::new(pos.base))`, and return `ensure_positive_tx(tx.base, "r", pos)` — so `r.-N` agrees with `c.-N`. Add a test asserting `c.-3` and `r.-3` resolve to the same SPDI position.

Issue #944, opened 2026-07-06
fix(convert): r.*N uses non-exon-aware cds_end+base; diverges from c.*N across multi-exon 3'UTR gaps


nh13:
## Summary

The `r.*N` (3'UTR RNA) SPDI/coordinate path uses plain `cds_end + base` arithmetic, while the equivalent `c.*N` path goes through the exon-aware, CIGAR-honoring `CoordinateMapper::cds_to_tx`. For a multi-exon 3'UTR where cdot carries tx-coordinate gaps, the two land at **different** transcript positions.

## Detail

- `src/spdi/convert.rs:1004` carries `TODO(#390-follow-up)` describing exactly this `r.*N` vs `c.*N` divergence; the `r.*N` branch short-circuits with `cds_end + base` instead of routing through the exon-aware helper.
- Root cause overlaps a broader limitation: `Transcript::cds_to_tx` is **not CIGAR-aware**. Two tests document this and are currently `#[ignore]`d — `tests/it/normalize_tests.rs:4687` and `:4720` — with the note "Until CIGAR-aware CDS mapping is implemented" (one literally asserts `cds_to_tx(76) = 186 (WRONG)` vs expected `189`, i.e. off by the cumulative indel count after an alignment insertion/deletion).

## Fix

Route the `r.*N` branch (and `cds_to_tx`) through the exon-walking CIGAR-aware mapping so `c.*N`/`r.*N` agree, then un-`#[ignore]` the two normalize tests. Narrow correctness scope (multi-exon 3'UTRs with cdot alignment gaps) but a genuine divergence.

nh13:
Progress (issue stays open): #950 resolved the `r.*N` half — it now routes through the exon-aware mapper so it agrees with `c.*N` across multi-exon 3'UTR gaps. Remaining under this issue: the broader non-CIGAR-aware `CdotTranscript::cds_to_tx` change (the two `#[ignore]`d `cds_to_tx` tests; ~38 call sites across projection/convert/protein-effect), left as a focused follow-up needing reference-backed conformance validation. Related follow-ups filed from this work: #960 (`r.-N` 5'UTR rejected while `c.-N` succeeds) and #962 (`r.*N`/`c.*N` not upper-bounded past the transcript 3' end).

PR #1665, opened 2026-08-11
fix(convert): resolve a c./n. position on the flat transcript, not through exon gaps

### LINKED ISSUE #1619: hgvs_to_spdi resolves a c. position through exon gaps while the sequence it indexes is flat

nh13:
Closes #1619

`Exon::start`/`end` are documented as **transcript** coordinates, and a transcript sequence is the concatenation of its exons — it has no holes. `CoordinateMapper` nonetheless walked the exon list whenever those coordinates showed a gap and skipped each gap as though it were an intron. The normalizer does no such thing: `region_sequence_delta` maps `c.N` to the flat offset `cds_start + N - 1`. So the two halves of ferro named different bases of one accession, and the error grew by one per gap crossed.

Reproduced by a test written to fail first, on the committed `tests/fixtures/sequences/normalization_transcripts.json`:

```
a_cds_position_resolves_onto_the_flat_transcript
  left: 1599      (hgvs_to_spdi, walking ten one-base gaps)
 right: 1589      (cds_start - 1 + 1520 - 1, what the normalizer reads)
```

The visible symptom was a correct normalization reading as a corruption. `c.1520_1522del` -> `c.1521_1523del` is the documented CFTR 3'-shift, and `the_shifted_pair_denotes_one_sequence_to_the_applier` pins that both spellings leave `TATCATTGGTGTT` over a window containing both. Read through the walk they landed on `[1599, 1603)` and left different sequences.

## Which side was wrong — settled by measurement, not by argument

The issue offered both readings, and said so: "either the resolution should not consult exons when the target sequence is the transcript itself, or the exon coordinates in that fixture are not the frame the resolver assumes them to be in." Both turn out to be true, and the second decides the first.

**The gaps are malformed fixture data, not a real shape.** Measured over 20 transcripts served from a real prepared reference — `NM_000492.4`, `NM_000088.3`, `NM_004006.2`, `NM_000051.3`, `NM_002111.8` among them — **all 20 have zero transcript-coordinate gaps**, every one starting at tx 1 and tiling contiguously, because `MultiFastaProvider` builds each exon's transcript interval straight from cdot's own tiling (`start: e[2] + 1, end: e[3]`). The fixture is worse than the issue reports: **22 of its 29 records** carry the same one-base-per-junction pattern, and it came in with the initial commit with no generator behind it.

So the walk is removed rather than the fixture patched. Patching the fixture would leave the resolver able to mis-resolve the next malformed record, and there is no faithful exon structure to patch it *to* — the spans sum to `seq_len − (n_exons − 1)` however they are re-tiled.

**The premise the walk was built on is written down and is false.** `make_transcript_with_gaps` says it "simulates how cdot encodes transcripts with virtual intron positions". cdot encodes nothing of the sort on the transcript axis; that comment is now replaced with the measurement.

## The six tests that pinned the walk

Their fixtures are kept and their invariant inverted: **a claimed transcript-coordinate gap must change nothing**, which is exactly what #1619 establishes. The two in `src/spdi/convert.rs` keep their real point — #944 and #960 are about `r.*N` / `r.-N` not getting their own short-circuit, so the two axes agree — and move only the value they agree *on*, onto the base the served sequence actually holds (`SPDI 39 -> 35`, `4 -> 8`). Their doc comments called those bases "nonexistent coordinates inside the gap"; the fixtures are 44-base sequences with a base at every one of them.

## Verification

- `cargo nextest run --features dev` — **`Summary [154.504s] 10139 tests run: 10139 passed (1 slow), 152 skipped`**.
- Reproduction control: with `origin/main`'s `src/convert/mapper.rs` restored and nothing else changed, 2 of the 3 new tests fail, `left: 1599 / right: 1589`.
- **All four seam oracles on, full suite, against `origin/main` at `b9f6cecb`: 25 failing tests before, 22 after.** The three fewer are exactly this row and its two siblings — `normalize_tests::…::test_deletion_shifts_in_real_homopolymer`, `…::test_3prime_shifting::case_1` and `idempotency_tests::test_normalization_idempotency` — and **nothing is added**.
- `cargo fmt --all --check` clean; `cargo clippy --features dev --all-targets -- -D warnings` clean.

## `FERRO_ASSERT_SEQUENCE` in `test-oracle` — not yet, and now for one reason

This closes one of the two rows keeping the flag out of CI's `test-oracle` job. **#1618 still fires**, and it is not the same kind of thing: it needs a ruling on what a single-anchor repeat's copy count is counted against, which is why no PR has closed it. The flag is deliberately not moved here. `CLAUDE.md`'s table is updated to say one row rather than two, to record what closed the other, and to name #1618 as the only remaining blocker.

Representation-Change: none. The normalizer's own `c.`->sequence mapping is `region_sequence_delta` and was already flat — this change makes `hgvs_to_spdi` agree with it, so what moves is the SPDI triple for a `c.`/`n.`/`r.` variant on a transcript whose exon list is malformed, from a triple naming bases the normalizer never reads to the one it does. No normalized HGVS string moves: the full suite passes with nothing re-blessed except the six pins that encoded the walk, and under the denoted-sequence oracle three previously-firing rows go green with none added. `examples/dump_normalized_corpus.rs` was not used as evidence, and the reason is worth stating correctly because an earlier draft of this trailer got it backwards. That generator does **not** build only a single exon: alongside the flat axis (`CDS_START = 1`) it carries a three-exon `cx` axis — `EXON_SPANS = [(0,7),(7,14),(14,20)]`, `CDS_START_MULTI = 4`, `INTRON_LEN = 10`, two junctions, both shuffle directions — which is what #1478 *added* rather than removed. The zero is still structural, for a different reason: those exons tile **contiguously in transcript coordinates** (tx 1-7 / 8-14 / 15-20), so no corpus row can carry a *gapped* exon list, and a gap is the only thing the removed walk keyed on. So a `0 moved` from it remains a fact about the corpus — just not the fact first claimed.

Previously **accepted**, with a wrong answer — the fact this trailer was missing. On the `NM_000492.4` fixture, `c.1520` is flat offset **1589** and `hgvs_to_spdi` published position **1599**, ten bases 3' of the base the normalizer reads (the fixture's exon list carries 1-base synthetic introns). So the conversion succeeded and named the wrong base; it did not error. Pinned in `tests/it/issue_1619_spdi_exon_gap_frame.rs`, which asserts `spdi.position == 1589`.

One surface this understates: the trailer says "what moves is the SPDI triple", but `CoordinateMapper::cds_to_tx`/`tx_to_cds` also back the public Python `CoordinateMapper` (`src/python.rs`), so that API's answers move too on a transcript whose exon list is gapped.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -330,17 +330,30 @@ had their chance to name the fault more precisely.
 **Where it runs, and the one place it deliberately does not.** `sweeps` sets it
 (gating, measured green over the full corpus) and the nightly sets it
 (non-gating). `test-oracle` does **not**, which makes it the only job where the
-set of four is incomplete. Two rows in that job's selection fire, and both are
-real disagreements inside ferro rather than noise:
+set of four is incomplete. **One** row in that job's selection fires, and it is a
+real disagreement inside ferro rather than noise:
 
 | row | what disagrees |
 |---|---|
 | #1618 — `NC_TEST.1:g.262TG[6]` → `g.259_262GT[6]` | `hgvs_to_spdi` reads the anchored spelling as 6 copies replacing a **1**-copy tract, the normalizer's output as 6 copies replacing a **2**-copy tract — 14 bases against 12 |
-| #1619 — `NM_000492.4:c.1520_1522del` → `c.1521_1523del` | the two forms are equivalent on the flat transcript; `hgvs_to_spdi` resolves the `c.` position by walking `normalization_transcripts.json`'s exon list, whose 1-base synthetic introns land it **10 bases** 3' of where the normalizer reads |
 
-Suppressing either would hollow out the oracle, and turning the flag on before
-they are settled would redden the job for a reason no PR caused. Move it into
-`test-oracle` once both are closed.
+**This was two rows until #1619 landed**, and the second is worth keeping on
+record because it is the only one of the pair that had an answer:
+`NM_000492.4:c.1520_1522del` → `c.1521_1523del` fired because `hgvs_to_spdi`
+resolved the `c.` position by walking `normalization_transcripts.json`'s exon
+list, whose one-base gaps in *transcript* coordinates landed it ten bases 3' of
+where the normalizer reads. Those gaps are malformed fixture data — measured
+over 20 real prepared-reference transcripts, none has one — so the walk was
+removed rather than the fixture patched. Measured with all four flags on: **25
+failing tests before, 22 after, the three fewer being exactly that row and its
+two siblings, and nothing added.**
+
+What is left is #1618, and it is *not* the same kind of thing: it needs a ruling
+on what a single-anchor repeat's copy count is counted against, which is why no
+PR has closed it. Suppressing it would hollow out the oracle, and turning the
+flag on before it is settled would redden the job for a reason no PR caused.
+Move `FERRO_ASSERT_SEQUENCE` into `test-oracle` once #1618 is closed — it is now
+the only thing in the way.
 
 **Measured false-positive classes, and what closed each.** The first run of this
 oracle over the suite raised **344** fires, all but 16 of them false. Each fix
--- a/src/convert/mapper.rs
+++ b/src/convert/mapper.rs
@@ -64,15 +64,15 @@ impl<'a> CoordinateMapper<'a> {
         let tx_base = if pos.utr3 {
             // 3' UTR: *N = CDS_end + N
             // For UTR, use exon-aware mapping from CDS end
-            self.cds_to_tx_exon_aware(cds_end as i64, pos.base)?
+            self.offset_tx_position(cds_end as i64, pos.base)?
         } else if pos.base < 1 {
             // 5' UTR: -N = CDS_start - N (where N is positive in HGVS but stored as negative)
             // e.g., c.-3 means 3 bases before CDS start, so offset is -3
-            self.cds_to_tx_exon_aware(cds_start as i64, pos.base)?
+            self.offset_tx_position(cds_start as i64, pos.base)?
         } else {
             // Normal CDS position - use exon-aware mapping
             // c.1 = cds_start (offset 0), c.2 = cds_start + 1 (offset 1), etc.
-            self.cds_to_tx_exon_aware(cds_start as i64, pos.base - 1)?
+            self.offset_tx_position(cds_start as i64, pos.base - 1)?
         };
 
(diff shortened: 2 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
Closing in favour of #1673, which fixes #1619 from the other side — restoring the fixture's exon `end` values rather than removing the exon walk from `CoordinateMapper`.

The reason is that this PR's central premise does not hold. It argues that transcript-coordinate gaps are malformed fixture data that cdot does not produce, evidenced by a sample of 20 transcripts. Measured over the full prepared cdot table:

- **GRCh38: 58 of 474,818** multi-exon builds carry a transcript-coordinate gap (0.0122%), sizes 23–2718 bases, none of them one base.
- **GRCh37: 159 of 190,754**, sizes 8–7074.

A 20-transcript sample cannot detect a 0.012% phenomenon, so the sample was consistent with the premise without supporting it.

The gaps also reach ferro's `Exon` list for real accessions. `multi_fasta.rs` has a defensive `has_gaps` check, but it falls back to supplemental CDS only where a supplemental entry exists; cross-referencing the 58 against the supplemental metadata leaves **8 with no fallback**, so the gap survives — `NM_001137667.1` / `NM_001137668.1` (CASP8AP2), `NM_001291281.2` (FOXO6), `NM_002457.3` (MUC2, a 2013-base gap), `NM_017940.6` (NBPF1), `NM_138352.2` (SAMD1), plus two `NR_` transcripts. For positions 3' of the gap on those, this change moves the `c.`/`n.` → transcript answer and therefore every SPDI triple, by the gap size.

That also makes `Representation-Change: none` under-declared: the trailer's own text describes the SPDI triple moving "on a transcript whose exon list is malformed", but the population is real reference data rather than a malformed fixture.

The two PRs are additionally mutually destructive and cannot both land: #1673 rewrites `normalization_transcripts.json` to remove the exon gaps (412 violating junctions → 0), while this PR adds a test asserting they remain (`gaps == 26`). Whichever merged second would turn the other red.

None of this says the `CoordinateMapper` simplification is wrong — only that the case for it needs rebuilding against those 8 accessions rather than against a 20-transcript sample, and that the trailer would need to stop saying `none`. The branch `fix/issue-1619-spdi-exon-gap-frame` is untouched, so reopening is available if that case gets made.

PR #1666, opened 2026-08-11
feat(changelog): attach each Representation-Change trailer to its changelog entry

### LINKED ISSUE #1556: changelog: the Representation-Change value never reaches the changelog, so the release checklist cannot be satisfied

nh13:
Closes #1556

The release checklist asked reviewers to confirm each **Representation changes** entry says whether the affected inputs were previously **rejected** (no stored string, so free) or previously **accepted** (a real migration). No entry could: release-plz writes the bullet from the commit subject alone, so the trailer's own text reached no reader and the answer had to be read out of every linked PR.

## Route four, because routes one to three are dead

The issue measures three routes through the git-cliff template and none of them is used here — they were read, not re-walked:

1. **`commit.footers` is mis-parsed.** git_conventional treats any `word:` line as a footer token, so a prose body shreds it.
2. **`commit.body` does not contain the trailer**, git_conventional having split it out — 0 occurrences for both real v0.13.1 disclosures.
3. **Splitting on the unanchored literal finds only the prose mentions**, which is worse than nothing: plausible output for some commits, quoted prose for others.

The fourth route is not through the template at all: **read the trailer from the raw commit message with `git log`, and attach it to the rendered bullet afterwards.** That is what `scripts/check_changelog_grouping.py` already does in CI to attribute trailers to commits, so the new script imports its `trailer_value` rather than keeping a fourth copy of the column-0 rule. Bullets are matched to commits by the PR number GitHub puts in the squash subject, the one token both sides are guaranteed to share.

## Rendered, and seen there

Not claimed — run. Rendering the pending cycle (`v0.13.1..HEAD`) through the real `release-plz.toml` parsers and then through the script turns this:

```
- *(normalize)* Bound the junction of a member that reduces to an insertion (#1598)
```

into this:

```
- *(normalize)* Bound the junction of a member that reduces to an insertion (#1598)
  > 371 rows move of 5,629,002 swept; all 371 were
  > previously incorrect — 321 denoted the wrong bases outright and 50
  > denoted no sequence at all because their members overlapped. This is a
  > correctness repair, not a respelling: no row that already denoted its
  > input's bases changes.
```

All ten entries in that cycle gain their disclosure. Running it over the **committed** `CHANGELOG.md` attaches the two v0.13.1 disclosures, which is the diff in this PR.

## Two measurements that contradict what `CONTRIBUTING.md` asks for

Both are load-bearing and both will be hit again by anyone working here:

- **The trailer is not the last thing in the squash body.** GitHub appends CodeRabbit's auto-generated summary after the PR description, so reading to end-of-message came out with three paragraphs of "Summary by CodeRabbit" quoted under #1484 as the author's disclosure. The walk stops at the next column-0 trailer token, Markdown heading or HTML comment.
- **The trailers are not well-formed**, their continuation lines being unindented — which is also why `git interpret-trailers --parse` reports *no* trailers at all for #1537, #1535 or #1547 — so indentation cannot delimit the value.

Both are recorded on the constants that implement them and pinned by tests, rather than left as a comment.

## What the checklist now asks

It asked for something unsatisfiable and said so, pointing at this issue. It now asks whether every entry carries its trailer text, names the one command that attaches it, and says the command is idempotent so it can be re-run after release-plz regenerates the section. CI runs the `--check` form inside the existing `Changelog grouping audit` job: a no-op on an ordinary PR (`CHANGELOG.md` holds only released sections), the gate on the release PR.

## What it refuses rather than skips

A bullet whose commit cannot be found, or whose commit carries no trailer, exits 2 with the bullet named. A silently-skipped bullet would reintroduce exactly the gap this closes — an entry with no disclosure, indistinguishable from one that needed none.

## Verification

- `pytest tests/python/test_inject_representation_disclosure.py tests/python/test_changelog_grouping.py tests/python/test_representation_change_trailer.py` — **114 passed**. (The rest of `tests/python/` needs the built extension module and does not collect without `maturin develop`; that is unchanged by this PR.)
- `python scripts/check_changelog_grouping.py` — `43 commits in v0.13.1..HEAD; 10 listed as representation changes / Representation changes section is consistent with the trailers.`
- `python scripts/inject_representation_disclosure.py --check` — exit 0 after the change, exit 1 naming `#1537, #1535` before it.
- `ruff check` and `ruff format --check` clean; `actionlint .github/workflows/ci.yml` clean.

Representation-Change: none. No file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched — this is a script, its tests, a CI step, and three documents. No ferro output moves.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -292,6 +292,25 @@ jobs:
         with:
           python-version: "3.12"
       - uses: taiki-e/install-action@a8fb21b85430c7a58551cf7ea87b03f355b826d9  # git-cliff
+      - name: Require each Representation changes entry to carry its disclosure
+        # The audit below asks whether the right commits are in the section. This asks
+        # whether each entry in the *committed* changelog says what it moved (#1556): the
+        # bullet is the commit subject and nothing else, so the rejected-vs-accepted fact
+        # the release checklist asks for had to be read out of every linked PR.
+        #
+        # A no-op on an ordinary PR, because `CHANGELOG.md` holds only released sections
+        # and release-plz has not written a pending one. It becomes the gate on the
+        # release PR, where the fix is the one command it names.
+        #
+        # It runs BEFORE the synthesis step on purpose, and moving it after would be a
+        # latent red. `squash_commits()` resolves each linked PR by walking `git log`, and
+        # synthesis ends in `git reset --soft "$PR_BASE_SHA"` — which rewinds HEAD's
+        # ancestry to the base tip as of the last PR event while leaving the worktree on
+        # the merge ref. On a PR whose base has since moved, a changelog entry can then
+        # cite a PR whose commit is no longer reachable, the lookup fails, and the script
+        # exits 2 for a cause no author can fix. Same hazard the synthesis step already
+        # documents for `git describe`.
+        run: python scripts/inject_representation_disclosure.py --check
       - name: Synthesize the squash commit this PR will become
         # Without this the audit cannot see the commit it exists to judge. GitHub builds
         # the squash subject from the PR title and the body from the PR description, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,25 @@ violations.
 [#1537]: https://github.com/fulcrumgenomics/ferro-hgvs/pull/1537
 
 - *(normalize)* never split a delins into members on consecutive nucleotides ([#1537](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1537))
+  > 3 rows of 500,004 move (0.0006%) — 2 respell, 1
+  > merge toward the submitted form. Confluence drops by 6 classes of
+  > 11,272, deliberately and by operator ruling: those classes agreed only
+  > through a spec-illegal intermediate, `sequence_changed` stays 0, and the
+  > surviving form is the better-formed one. 58 separation-0 violations of
+  > `delins.md:16` are fixed.
 - *(normalize)* type the whole-block inversion by what it competes with ([#1535](https://github.com/fulcrumgenomics/ferro-hgvs/pull/1535))
+  > 0 rows move over 5,761,302 real expressions
+  > (ClinVar 500,004 / Paraphase 435,235 / CMRG 4,826,063), diffed row by
+  > row between this branch and its base on both binaries. That zero is
+  > structural for the direction that could move: a cis-allele of >=2
+  > `delins` members forming a whole-block reverse complement has **0**
+  > instances in those corpora, so the corpus cannot build the input this
+  > change converts. It is not vacuous either — 2,639 `inv` rows (897 of
+  > them genomic) and 38,190 multi-base spanning `delins` rows do reach the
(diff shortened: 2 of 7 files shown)


nh13:
@coderabbitai pause

PR #1634, opened 2026-08-11
docs(rulings): decide the absolute-prohibition enforcement stage as mode-dependent

nh13:
Two operator rulings, one commit each. Neither moves a ferro output in this PR; both commit the project to a position, and one of them commits it to a future move.

**Based on #1585's head (`feat/exhaustive-spec-corpus-on-main`), not on `main`.** Both records exist only on that branch — `main` has 11 rulings, #1585's head has 15 — so this is stacked on it and targets it.

That branch was force-pushed mid-work (rebased onto a newer `main`), so these two commits were cherry-picked onto its new tip `7245f6b3`. Two consequences, both good: the README normalization ruleset (#1605) is now on the base, so the rule-1 and rule-5 citations here resolve directly; and the base had already made the `CLAUDE.md` staleness corrections this PR originally carried, so that file's diff is down to moving one row between the open and decided tables per ruling.

## 1. `absolute-prohibition-enforcement-stage` → **`decided`: MODE-DEPENDENT**

- **strict** — fail at **parse**. Strict validates input conformance, not merely that the input can be parsed.
- **lenient** — input conformance is **not** validated; it fails only when it cannot **normalize**, which can happen.
- **silent** — lenient with no error messages, but an exit code.

**This is a third option the record did not enumerate.** It offered (A) parse unconditionally in every mode, or (B) strict-mode normalize. Parse-gated-by-mode keeps the opt-out that made (A) unacceptable while putting the check where the record's own reasoning says it belongs.

**It answers the record's only objection to (A).** The record argued that `adjudication-precedence-order` ranks validity first and never trades it, so a mode flag switching validation off is a trade. It is not: rule 1 of the README ruleset — *"Output follows the HGVS recommendations. Absolute — never traded."* — is about **output**. Accepting a non-conformant input and normalizing it to a conformant output trades nothing. What rule 1 forbids is *emitting* the prohibited spelling, and that binds in every mode with no flag.

**It resolves both internal inconsistencies**, which are the actual defect: `checklist.md:16` names `+`, `-`, `*` in one sentence and ferro refuses only `*`; item 3 "Insertions" states `:31` and `:33` four lines apart and only `:31` is enforced. Under the ruling each is a strict-mode parse refusal and none is unconditional.

The record's honest framing survives: the spec addresses **descriptions**, not stages. `checklist.md:5` governs only in that it says what the checklist *is* — a catalogue of descriptions that "do not correctly follow HGVS nomenclature", i.e. a property of the input.

[...]

Issue #1632, opened 2026-08-11
parse_hgvs documents itself as strict but applies no ErrorConfig, so it accepts what strict rejects


nh13:
`parse_hgvs`'s own doc comment says:

> Uses strict error handling mode by default. For configurable error handling, use [`parse_hgvs_with_config`] instead.

It does not. `parse_hgvs` applies no `ErrorConfig` at all, so none of the input-hygiene ladder runs on it — including the rungs that make strict *strict*.

## Measured

```
                                    parse_hgvs (plain)        parse_hgvs_with_config(_, strict())
NM_024312.4:r.-6_-3g[6]             Ok(unchanged)             Err "redundant base label 'g'"
NM_024312.4:r.-125_-123cug[4]       Ok(unchanged)             Err "redundant base label 'cug'"
NC_TEST.1:g.266del3                 Err (grammar)             Err "uses size-count suffix"  (lenient repairs it)
NM_TEST.1:c.10_12 del               Err (grammar)             Err "Extra whitespace"        (lenient repairs it)
```

The first two are the plain finding: the documented-strict entry **accepts** what the actual strict config **rejects**. The last two show the same seam from the other side — the grammar refuses them for its own reasons, and only the config-aware entry produces the mode-appropriate answer.

## Why it matters

1. **The doc states a contract the function does not have.** A caller reading it believes `parse_hgvs` validates; it only parses.
2. **It silently invalidates measurements.** `conformance::spec_corpus` and `tests/it/corpus_prohibited_inputs.rs` both call the bare `parse_hgvs` and then apply the mode at `Normalizer`. That is why the `absolute-prohibition-enforcement-stage` record recorded five clauses as "refused at PARSE, hence in every mode" when two of them (`checklist.md:49`, `general.md:96`) are in fact refused in strict and **repaired** in lenient. The census was measuring one door and describing the building.
3. **It makes the library and the CLI disagree by default.** `ferro normalize` defaults to lenient and repairs; a library caller on `parse_hgvs` gets neither the repair nor the rejection.

## Options

Either make the doc true — route `parse_hgvs` through `parse_hgvs_with_config(input, ErrorConfig::strict())` — or make the doc honest, saying it applies the grammar only and naming `parse_hgvs_with_config` as the entry that applies a mode. The first is a behaviour change with real blast radius (it is the crate's headline entry point and the fast path sits under it); the second is free and should not wait on the first.

[...]

PR #1667, opened 2026-08-11
docs(parser): say what parse_hgvs actually applies — the grammar, and no ErrorConfig

nh13:
Representation-Change: none. A doc comment and a test; no parsing, normalization or output behaviour changes.

Refs #1632

## Deliberately `Refs`, not `Closes`

#1632 lists two options. This takes the **second** — make the documentation honest — which the issue itself calls free and says should not wait on the first. The issue stays open for the first.

## What was wrong

`parse_hgvs`'s doc comment claimed "strict error handling mode by default". It applies **no `ErrorConfig` at all**: it trims surrounding whitespace, runs the fast path, and falls back to the grammar. The input-hygiene ladder — both the repairs `lenient` performs and the refusals that make `strict` strict — lives in `InputPreprocessor` and is reached only from `parse_hgvs_with_config`.

So the entry point is not "strict". It is a third behaviour that is neither mode, and it accepts inputs `strict` refuses:

```
NM_024312.4:r.-6_-3g[6]
  parse_hgvs(..)                            -> accepted, unchanged
  parse_hgvs_with_config(.., strict())      -> refused: "redundant base label 'g'"
```

That is the gap the doc was hiding, and the test pins it in both directions so the claim cannot rot back.

## Why the other half is left alone

Routing `parse_hgvs` through `ErrorConfig::strict()` means the crate's headline entry point starts refusing inputs it accepts today — a real behaviour change for every caller — and it sits downstream of `absolute-prohibition-enforcement-stage`, which #1634 decided as **mode-dependent** — refusal at strict-mode normalize rather than unconditionally at parse. A mode-less entry point has no defined enforcement stage under that ruling, which argues for giving `parse_hgvs` a mode rather than against it; either way it is not a doc fix and should not ride along with one.

## Correcting the issue's premise

#1632's framing is that `parse_hgvs` is documented strict and silently is not. That much holds. What does **not** hold is the related claim carried in our own notes that intronic-position-on-a-bare-accession is "refused in no mode" — `bare-transcript-intronic-position` is a **decided** record with a passing test: strict refuses a bare `NM_…:c.20+2del` with `W4007`, lenient accepts. Making that absolute would re-open a decided ruling, not fix a bug. Recorded here so the wrong version does not get acted on.

Also drops "(default)" from the module-level mode list, where it read as a property of the parser rather than of `ErrorConfig::default()`.

## Verification

- Full suite on current `main`: **10,140 run, 10,140 passed, 152 skipped** (`FERRO_MANIFEST` set, so the conformance censuses ran rather than skipping green). That figure was measured at `68694272`; the branch has since taken three review fixes and a fifth test, so the count is now five higher.
- `cargo clippy --all-features --all-targets -- -D warnings` clean; `issue_1632_parse_entry_applies_no_mode` 5/5 pass

### DIFF
--- a/src/hgvs/parser/mod.rs
+++ b/src/hgvs/parser/mod.rs
@@ -6,11 +6,14 @@
 //!
 //! The parser supports three error handling modes:
 //!
-//! - **Strict** (default): Reject all non-standard input
+//! - **Strict**: Reject all non-standard input
 //! - **Lenient**: Auto-correct common errors with warnings
 //! - **Silent**: Auto-correct common errors without warnings
 //!
-//! Use [`parse_hgvs_with_config`] to specify the error handling mode.
+//! A mode is applied only by [`parse_hgvs_with_config`]; strict is
+//! [`ErrorConfig::default`], so `parse_hgvs_with_config(input,
+//! ErrorConfig::default())` is the strict entry. [`parse_hgvs`] applies **no**
+//! mode at all — see its own documentation (#1632).
 
 pub mod accession;
 pub mod edit;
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1668, opened 2026-08-11
test(normalize): pin the stranded identity member, measured and non-confluent

### LINKED ISSUE #1655: normalize: a derived identity member survives when its absorber was consumed — 132 of 132 non-confluent
## What's wrong

An **identity member** — one that changes nothing — survives normalization when the member that would have absorbed it no longer exists.

```
input   NC_TEST.1:g.[10_11insA;11del;12_13insAA]
output  NC_TEST.1:g.[11=;12_13insAA]

but     NC_TEST.1:g.12_13insAA
output  NC_TEST.1:g.12_13insAA        (a fixed point)
```

Both spellings denote the same sequence, and both outputs are idempotent. So one variant has **two stable normalized representations** — non-confluence, with no error or warning.

## Why the drop rule cannot reach it

The rule that removes an identity residue scans **all** siblings, not just adjacent ones, and requires the identity to **overlap** a surviving member. That premise never holds here: the absorber was **consumed during the merge**, so by the time the rule runs there is no surviving member that grew over the identity. The residue is stranded disjoint from a member that is gone.

(Note for anyone reading an older description of this: the rule is *not* adjacency-limited. That framing is wrong and sent an earlier investigation down the wrong path.)

## Census

Measured over a three-member generated family:

| | |
|---|---|
| instances found | **132** |
| non-confluent | **132 of 132** |
| 5′ direction | 65 |
| 3′ direction | 67 |

The even split across directions shows this is **not** a shuffle-direction disagreement — it is a property of the merge.

### A structural zero worth recording

The first sweep was **two-member** and returned a clean zero. That zero was **structural, not evidence**: a two-member family cannot build an absorbed identity at all, because there is no third member to do the absorbing. The committed sweep is three-member and asserts a non-zero denominator, so it cannot silently become vacuous.

## What blocks a fix

Dropping the residue unconditionally is wrong. The drop rule explicitly argues the other side for **authored** identity members — a caller who writes `g.11=` is asserting something, and the current behaviour preserves it deliberately.

So a fix has to separate an identity the caller **authored** from one normalization **derived**, and that distinction is a policy question rather than a mechanical one. Filing this with the measurement attached so the decision has a denominator.

nh13:
Representation-Change: none. Tests only; no file under `src/` is touched.

Closes #1655

## What this pins

When a derived identity member's absorber is consumed during the merge, the residue survives into the output. Swept over three-member families: **132 instances, 132 of 132 non-confluent** — 65 under the 5′ rule, 67 under 3′.

This PR does not fix it. It pins the measurement so the class cannot quietly change size, and so a fix has a denominator to move.

## The mechanism, corrected

The obvious explanation — that the drop rule is adjacency-limited and simply cannot see a residue sitting away from its absorber — is **wrong**. The rule scans all siblings and keys on overlap. The reason it cannot reach this residue is that **the absorber was consumed in the merge**, so the rule's premise never holds: there is no surviving sibling to overlap with.

That distinction matters for whoever fixes it. Widening the adjacency window would change nothing.

## The sweep asserts its own denominator

The first version of this sweep was **two-member** and returned a clean zero. That zero was structural, not a result: a two-member family cannot construct an absorbed identity, because the absorber and the residue need a third member to survive the merge. It would have read as "nothing to see here".

The committed sweep is three-member and asserts a non-zero denominator, so `0 of 0` cannot pass as evidence of health — the same discipline `spec_conformance_axis` applies to its negative guards.

## Verification

- Full suite on current `main`: **10,138 run, 10,138 passed, 152 skipped** (`FERRO_MANIFEST` set, so the conformance censuses ran rather than skipping green)

### DIFF
--- a/tests/it/main.rs
+++ b/tests/it/main.rs
@@ -28,6 +28,7 @@ mod issue_1632_parse_entry_applies_no_mode;
 mod repeat_input_idempotency;
 mod repeat_lowering_sibling_junction;
 mod residual_above_cap_confluence;
+mod stranded_identity_member;
 
 mod allele_grammar_corners;
 mod allele_trans_phase;
--- a/tests/it/stranded_identity_member.rs
+++ b/tests/it/stranded_identity_member.rs
@@ -1,0 +1,583 @@
+//! PINNED DEFECT: a *derived* identity member that no sibling overlaps, so
+//! `drop_identity_members_covered_by_siblings` cannot reach it — leaving one
+//! variant with two normal forms.
+//!
+//! # The shape
+//!
+//! ```text
+//! reference  CGCGCGCGCAATCGCGCGCAATCGCG        10 = A, 11 = A, 12 = T
+//!
+//! in    TEMPLATE:g.[10_11insA;11del;12_13insAA]
+//! out   TEMPLATE:g.[11=;12_13insAA]
+//! but   TEMPLATE:g.12_13insAA        ->  TEMPLATE:g.12_13insAA
+//! ```
+//!
+//! `10_11insA` inserts an `A` into the `AA` tract and `11del` takes one back
+//! out, so between them they change nothing and collapse to `11=`. The third
+//! member is untouched and sits two bases away. The result is a two-member
+//! allele whose first member asserts nothing and whose second, spelled alone,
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai pause

PR #1671, opened 2026-08-11
docs: enforce the adjudication tables against the ledger, and correct three records

### LINKED ISSUE #1584: CLAUDE.md's adjudication-records section contradicts the ledger: 5 stale claims, and the currency scan does not cover it
`CLAUDE.md`'s "Adjudication records: where the open questions live" section contradicts the ledger it summarises. It is the first thing a session reads, so a stale row here is the most expensive kind: it sends the reader off to re-argue a question that has already been answered — the exact failure #1579's citation-currency guard was built to stop, in the one file that guard does not scan.

Measured against `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json` at `origin/main` (`5b54b2b`):

```
decided    delins-merge-vs-individual-gap-two-or-more
decided    inversion-vs-two-delins-76-83
decided    adjudication-precedence-order
decided    canonical-form-choice-when-both-legal
undecided  codon-carve-out-shape-restriction
decided    delins-codon-carve-out-gap-one
undecided  exon-junction-dup-converge-from-the-far-side
undecided  rna-repeat-range-plus-unit-redundancy
decided    delins-adjacent-members-when-both-consume-reference
undecided  separation-is-a-property-of-the-spelling-not-of-the-variant
```

Ten records, four `undecided`. Five specific contradictions:

1. `CLAUDE.md:670` — "**Five of its eight records are `undecided`**". It is four of ten.
2. `CLAUDE.md:675` — `adjudication-precedence-order` is listed in the *open questions* table as "**Two competing precedence orders are on record and neither is chosen**". The ledger has it `decided`, governing `DNA/complex.md:63`. #1579 corrects exactly this claim at `src/normalize/merge.rs:1890`, so the repo now says both things in two places.
3. `CLAUDE.md:676` — `canonical-form-choice-when-both-legal` is listed as "an open *product* decision". The ledger has it `decided`.
4. The "**three `decided` records**" table below the first one lists three. There are six — the two above plus `delins-adjacent-members-when-both-consume-reference`, which the section never mentions.
5. `CLAUDE.md:693` — "which one applies is part of what `adjudication-precedence-order` leaves open" rests on (2) and does not survive it.

`separation-is-a-property-of-the-spelling-not-of-the-variant` (`undecided`) is also absent from the open-questions table.

### Why this needs its own change

Fixing it properly means moving two rows between tables, adding three, and writing an accurate one-line ruling summary for each newly-listed record — which requires reading each record's `rationale` and its scope paragraph. Both 2026-08-07 rulings are narrower than their one-line summaries, and the section itself warns about that, so a summary written quickly here would replace stale claims with wrong ones. That is worse, not better.

### Consider widening the currency scan

`tests/it/ruling_citation_currency.rs` scans `src/` and `tests/` for `.rs` only, and its module docs already name a widening ("the same stale prose also appears in fixture **JSON** `note` fields … widening this scan to `.json` is a reasonable follow-up"). `CLAUDE.md` is the higher-value target: it is where the summary tables live, it is what every session reads first, and it is how this particular drift went unnoticed. A scan over `CLAUDE.md` + `docs/**/*.md` using the existing line-scoped status-word matcher would have failed on all five items above.

Note the guard would need the count claim and the table *placement* to be judged too, not just status words on a citation's own line — item 1 and item 4 are not status contradictions on a cited line, so the existing matcher catches items 2, 3 and 5 but not 1 or 4.

nh13:
Representation-Change: none. Documentation plus one new integration test; nothing under `src/` is touched and no expectation was re-blessed.

Closes #1584

## The drift

`CLAUDE.md`'s two adjudication tables encode a **status claim** — which table a record sits in — and that split was hand-maintained. Measured against `origin/main`'s ledger:

| | |
|---|---|
| listed as open, actually `decided` | `codon-carve-out-shape-restriction`, `exon-junction-dup-converge-from-the-far-side`, `separation-is-a-property-of-the-spelling-not-of-the-variant` |
| in **neither** table | `ring-telomere-anchoring` |
| ledger truth | 16 records, 13 decided, 3 undecided |

A stale status is the expensive direction. A stale *count* reads as a claim about coverage; a stale *status* reads as an open question that is in fact settled, and the next reader re-derives a decision already made.

**This came within one step of doing exactly that today.** The `separation-is-a-property-of-the-spelling-not-of-the-variant` record's id states its **question**, and its ruling is the **opposite** of it — separation is read off the partition re-derived from the resulting sequence, never off the input's spelling. A reader who trusts the stale table and stops at the id gets the answer backwards.

## Why the existing currency scan could not catch it

`ruling_citation_currency` already compares status claims against the ledger. It would not have helped, for two independent reasons:

1. It scans `.rs` files under `SCAN_ROOTS = ["src", "tests"]`. `CLAUDE.md` is Markdown at the repository root — outside that twice over.
2. It is **line-based**. Here the status word ("open") sits in the introducing paragraph while the ids sit in table rows, so **no single line carries both**.

Widening its roots or its extension filter would still have found nothing. The check has to be section-aware, which is what this adds.

## The guard

`claude_md_adjudication_tables`, two tests:

- the two tables must **partition** the ledger by status
- their **union** must be the ledger exactly, with no record in both

Both were exercised against the real drift rather than assumed to work — reinstating a decided record in the open table, and deleting a record from both tables. Each fails with the specific record named, and passes once restored.

The failure messages name which records are on the wrong side and in which direction, so the fix is mechanical.

## Also recorded

- **A record's id states its question, not its ruling** — with the example above, since that is not derivable from the ledger.
- **`delins-merge-vs-individual-gap-two-or-more` says "re-run the census before citing it" and then publishes a grep that omits `REQUIRES`.** Following it reproduces the wrong count: one uppercase RFC 2119 keyword outside `style.md` where there are two (`RNA/adjoined_transcript.md:20` and `:21`, at spec checkout `6f85311`). Verified both ways.
- **The `clause_ruling_index` trap** — the file is committed but embeds a *generated* block, so a hand edit is lossy transcription that survives review as an ordinary diff and fails only on the one assertion that re-renders it. Capture what the test prints programmatically.

## One claim dropped rather than fixed, because it is false

Our notes carried "CLAUDE.md says a decline reason may follow the marker — the shipped regex says no". The regex **does** admit it: `none. Tests only; no watched file is touched.` passes, and `DECLINE_RE` explicitly allows a terminator followed by prose. Nothing to correct.

## Verification

- Full suite: **10,138 run, 10,138 passed, 152 skipped** (`FERRO_MANIFEST` set, so the conformance censuses ran rather than skipping green)
- `cargo fmt --all --check` and `cargo clippy --features dev --all-targets -- -D warnings` clean
- The new guard's first draft **tripped `ruling_citation_currency`** — describing historical drift put a record id on a line with contradicting status words. Fixed by rewording rather than by adding a scan exclusion, which would have blinded the scan to this file permanently.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -625,6 +625,22 @@ Contributor-facing guidance is in `CONTRIBUTING.md`. The checker is
 `tests/python/test_representation_change_trailer.py`, because three copies of one list drift
 silently.
 
+### Never hand-edit `tests/it/clause_ruling_index.rs` — it embeds a GENERATED block
+
+The file is committed, so it looks hand-maintainable. It is not: it carries a rendered index of
+every clause the ruling records cite, and `the_rendered_index_is_current` compares that block
+against what the code generates. Editing it by hand is transcription, and transcription of a
+~100-row block is lossy in a way that reads as a small diff — two attempts at it were made here
+and both dropped rows silently.
+
+Capture what the test prints **programmatically** instead. The pattern that works: add a throwaway
+test that writes `render_index()` to a file, run it, restore the source byte-for-byte from a
+pre-edit copy, then splice the generated text in on its `BEGIN`/`END` markers. Nothing is retyped.
+
+Note this is a *different* trap from the gitignored spec artifacts above. Those regenerate on
+demand and a stale one fails loudly. This one is committed, so a lossy edit survives review as an
+ordinary diff and only fails on the one assertion that re-renders it.
+
 ### Record every adjudication as a committed test or ruling record, in the same change
 
 **Policy.** When you decide what the *correct* normalization behaviour is — not how to
@@ -1004,6 +1020,20 @@ The `rulings` section of `tests/fixtures/grammar/hgvs_spec_normalization_overrid
 decision log, pinned by `ruling_records_are_intact` (ids **and** statuses; keep `RULING_STATUSES`
 in sync).
 
+**Read the ledger before adjudicating anything from spec text.** The records answer most of what
+gets re-argued, and re-deriving a settled decision is this project's most repeated failure. Two
+traps make it worse than simply forgetting to look:
+
+- **A record's id states its QUESTION, not its ruling, and the two can be opposites.**
+  `separation-is-a-property-of-the-spelling-not-of-the-variant` rules that separation is read off
+  the partition **re-derived from the resulting sequence** — the negation of its own name. Reading
+  the id and stopping gets the answer exactly backwards.
+- **A record's own instructions can be stale.** `delins-merge-vs-individual-gap-two-or-more` says
+  "Re-run the census before citing it" and then publishes a grep whose alternation omits
+  `REQUIRES`, so following it reproduces the wrong count — one uppercase RFC 2119 keyword outside
+  `style.md` when there are two (`RNA/adjoined_transcript.md:20` and `:21`, at spec checkout
+  `6f85311`). Re-run the command *and* sanity-check the command.
+
 **Do not trust a count of decided/undecided records written down here, and do not trust which
 side of the line a record is on either.** Both failures have happened: the sentence this replaces
 said "five of its eight records are `undecided`" long after the ledger had grown and four of those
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
Two lens findings applied in `62ff8e44`, one from each lens, and both were the same shape: an assertion narrower than the name above it.

**The CLI lens (`major`… actually `minor`, and correctly so): `ids_in_table` collected into a `BTreeSet`, so a record listed twice in ONE table was silently deduped** — under a test named `every_ledger_record_appears_exactly_once`. The cross-table overlap check caught a record in *both* tables; nothing caught a record twice in the *same* table, which is a live documentation defect on its own, since two rows for one record can carry two different descriptions of the ruling and nothing says which is current. `ids_in_table` now returns a `Vec` in document order and `unique_ids` refuses repeats.

Pinned with the synthetic pair the lens asked for, which is also the only way to exercise a malformed table without editing a committed file:

- `a_record_listed_twice_in_one_table_is_rejected` — `#[should_panic]` on a three-row table whose first and third rows are both `alpha`.
- `a_table_with_no_repeats_is_accepted` — the negative control, so the guard above cannot be satisfied by `ids_in_table` failing for an unrelated reason.

**The skill lens: the guard's own failure message stated the opposite of the ledger.** `want_open` is built from the `undecided` records, so `want_open.difference(&open)` is the set the ledger leaves **open** — and the message called them `decided in the ledger but missing from the open table`. Reproduced by mutation (deleting the `ring-telomere-anchoring` row) and read verbatim:

```
decided in the ledger but missing from the open table: ["ring-telomere-anchoring"]
```

`ring-telomere-anchoring` is `undecided`. A guard that exists because a stale status claim sends readers the wrong way should not state a wrong status in its own diagnostic — it would send them to move the row into the wrong table. Now reads `undecided in the ledger but missing from the open table`, verified by re-running the same mutation.

**The guard bites.** Deleting one row from the open table reddens both tests with the exact #1584 shape. Validated at CI parity: `fmt` clean, `clippy --all-features --all-targets -D warnings` clean, plain suite **10148/10148**, oracle suite (three flags armed, `ORACLE_EXCLUDE` negated as `test-oracle` does) **10086/10086** — each up two from the pre-change run, which are the two tests added here.

PR #1672, opened 2026-08-11
fix(project): decide the codon-frame delins exception per rendered axis

### LINKED ISSUE #1664: project: the codon-frame delins exception is decided on the authoring axis and inherited by the others, so a single-member description projects to the split form the spec calls "not correct"
# project: the codon-frame delins exception is decided on the authoring axis and inherited by the others, so a single-member description projects to the split form the spec calls "not correct"

## Summary

`LRG_199t1:c.145_147delinsTGG` is the HGVS recommendations' own worked example. Ferro projects it as **one member on `c.`, `n.`, `r.` and `p.`** and as **two members on `g.`** — and the two-member genomic form is, position for position, the description the spec renders `class="invalid"` and calls "not correct".

```text
input   LRG_199t1:c.145_147delinsTGG          1 member

c       LRG_199t1:c.145_147delinsTGG          1 member    correct
n       LRG_199t1:n.389_391delinsTGG          1 member
r       LRG_199t1:r.(145_147delinsugg)        1 member
p       LRG_199p1:p.(Arg49Trp)                1 member
g       LRG_199:g.[494841C>T;494843C>G]       2 members   <-- the invalid form
```

`g.494841` is `c.145` and `g.494843` is `c.147`, so the genomic axis emits `c.[145C>T;147C>G]` re-coordinated. `DNA/delins.md:42`, two lines below the example itself:

> **NOTE**: two variants separated by one nucleotide, together affecting one amino acid, should be described as a "delins", so the description `c.[145C>T;147C>G]` is not correct.

That spelling carries `class="invalid"` in the published page.

The same happens to the spec's second worked example of the same rule, `LRG_199t1:c.235_237delinsTAT` → `LRG_199:g.[499798A>T;499800G>T]`. That one is worse in a specific way: `DNA/delins.md:19` gives it as the reason the exception exists at all —

> **NOTE**: this prevents tools predicting the consequences of a variant to make conflicting and incorrect predictions of two different substitutions at one position (e.g., `c.235_237delinsTAT` (`p.Lys79Tyr`) versus `c.[235A>T;237G>T]` (`p.[Lys79*;Lys79Asn]`).

Ferro's own `p.` axis gets that right — `LRG_199p1:p.(Lys79Tyr)` — while its `g.` axis publishes exactly the split the note exists to stop a downstream tool seeing.

## The rule being broken

**The exception must be evaluated once per rendered axis, against the transcript the projection is being made on. Today it is decided once, on the axis the caller happened to author, and every derived axis inherits that decision.**

The governing clause is `DNA/delins.md:18`:

> **exception**: two variants separated by one nucleotide, together affecting one amino acid, should be described as a "delins".

For this project the reading is already settled — `rulings[delins-codon-carve-out-gap-one]` in `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json` is **decided**, with `delins.md:18` named as governing. So this is a gap in *where* the decided rule is applied, not a new adjudication.

It fails in both directions, and the second direction is the one that makes it a projection issue rather than a normalization one:

| authored on | axis that is wrong | why it is wrong |
|---|---|---|
| `c.` | `g.` | the projector knows the transcript and the frame; it re-derives the genomic axis without consulting either |
| `g.` | `c.` / `n.` | the split was decided on a frameless axis, and the transcript axes — where the frame IS available — inherit it verbatim |

**Expected:**

```text
LRG_199t1:c.145_147delinsTGG  ->  g. LRG_199:g.494841_494843delinsTGG
LRG_199t1:c.235_237delinsTAT  ->  g. LRG_199:g.499798_499800delinsTAT
```

and, in the other direction, a genomic input whose transcript axes are rendered against a known frame must render them merged.

## What this is NOT — the boundary against #79

Issue #79 implemented the codon-frame exception and explicitly scoped it out of the bare genomic axis. Its own table:

| Input | Spec form |
|---|---|
| `g.[145G>A;147C>T]` (no codon frame in `g.`) | unchanged |

That is correct and this issue does not reopen it. A bare `g.` description names no transcript, so there is no frame to consult and no basis for merging. **The difference here is that a projection always names a transcript.** `project_variant(v, "LRG_199t1")` has the frame in hand for every axis it renders; declining to use it on one of them is not the same decision #79 made.

Also not implicated:

- **The cis-allele partitioning work** (#1229–#1235 / #1419–#1421). The input has no members to partition. This is a single `delins` acquiring members it was never authored with.
- **`general.md:36-39`.** That forward-looking note describes SVD-WG010, which was **rejected** (`consultation/SVD-WG010.md:8`).

## Reproducing it

Both reproducers are hermetic — no prepared reference, no `FERRO_MANIFEST`.

**With the spec's own examples**, over the committed enumeration slice (`tests/fixtures/grammar/spec_enumeration_windows.json`):

```rust
let windows = load_slice("tests/fixtures/grammar/spec_enumeration_windows.json")?;
let cdot = CdotMapper::from_transcripts(windows.transcripts.iter());
let projector = VariantProjector::new(Projector::new(cdot), windows.to_provider());
let v = parse_hgvs("LRG_199t1:c.145_147delinsTGG")?;
let p = projector.project_variant(&v, "LRG_199t1")?;
// p.coding  == LRG_199t1:c.145_147delinsTGG
// p.genomic == LRG_199:g.[494841C>T;494843C>G]
```

**Minimally, on a synthetic single-exon transcript** placed so `g.N == c.N`, CDS `ATGCTGCGTTGGGCCTAA` (Met-Leu-Arg-Trp-Ala-Stop). Replacing codon 2's `CTG` with `ATA` changes bases 1 and 3 and leaves base 2, so the pair is separated by exactly one nucleotide and affects exactly one amino acid:

```text
tx:c.4_6delinsATA    ->  c. tx:c.4_6delinsATA          n. r.  all ONE member
                         p. txp:p.(Leu2Ile)

REF:g.4_6delinsATA   ->  g. REF:g.[4C>A;6G>A]          <- #79's scope, fine
                         c. REF(tx):c.[4C>A;6G>A]      <- DEFECT
                         n. REF(tx):n.[4C>A;6G>A]      <- DEFECT
                         r. tx:r.[(4c>a;6g>a)]
                         p. txp:p.(Leu2Ile)
```

One variant, the same bases, the same protein consequence, two spellings — and the *transcript* axes disagree about member count depending on which spelling the caller happened to use.

## Which axes

- On the two spec examples (`c.` authored): **`g.` only** splits; `c.`, `n.`, `r.`, `p.` all render one member.
- On the synthetic control (`g.` authored): the split propagates to every nucleotide axis — `g.`, `c.`, `n.`, `r.` — of which the transcript ones are the defect. The protein axis is correct in both directions.

## Scope: the existing counter is a superset, and most of it is not a defect

The enumeration already has a name and a budget for the *shape*: `Status::ProjectionSplitsSingleMember`, pinned at **9** in `tests/it/spec_enumeration_tests.rs`. Nine rows, three inputs:

| input | axes counted | is it this issue? |
|---|---|---|
| `LRG_199t1:c.145_147delinsTGG` | `g` | **yes** — `delins.md:42` calls the emitted form not correct |
| `LRG_199t1:c.235_237delinsTAT` | `g` | **yes** — same clause, and `:19` names the protein harm |
| `NM_000797.3:c.812_829delins908_925` | `c`, `n`, `r`, `p` | no — members end up two or more bases apart, where `general.md:34` asks for the split |
| `c.235_238delinsTAGT` | `c`, `n`, `r` | no — separation 2, likewise |

So **2 of the 9 rows are this issue** and the other 7 are the counter doing what it was built to do: surface a shape for inspection. The enumeration's own comment already says this ("the rows enter by *shape* … not by the divergence `delins.md:42` names"), and it is repeated here because the budget number invites reading all nine as defects.

Two consequences for whoever fixes this:

- Driving the counter to zero is **not** the goal. A fix should move it 9 → 7. If it moves further, something merged that `general.md:34` wanted split.
- The counter is a count, not a row set, so a compensating swap is invisible to it. That is why the two rows are pinned by name rather than left to the budget.

## Existing coverage, and what it does not cover

`tests/it/protein_cis_delins_decomposition.rs` asserts `!contains('[')` for the **protein** axis on four inputs, and is manifest-gated so it skips on a bare checkout. There is no `g.`/`c.`/`n.`/`r.` counterpart anywhere in the suite. That is the coverage hole this class lived in.

## Representation-Change

A fix moves the affected inputs' genomic axis (and, in the other direction, their transcript axes) from a two-member allele to a single `delins`. The blast radius should be measured before merging: the shape is "two changed columns one apart inside one codon", which is common in real data, so this is likely a real migration rather than a `none`.

## Pinned by

`tests/it/lenient_overlap_and_projection_split.rs`:

- `the_genomic_axis_alone_splits_the_specs_own_codon_delins` — the two spec examples over the committed slice, asserting exactly one axis splits and that it is `g.`
- `the_same_codon_delins_splits_when_authored_genomically_and_not_when_authored_on_the_transcript` — the synthetic A/B, asserting both the exact strings and the transcript-axis member-count property

Both carry `PINNED DEFECT`: when the fix lands, move the assertions to the single-member forms in the same commit.

nh13:
Closes #1664.

`DNA/delins.md:42`'s codon carve-out reaches only an axis that **declares a reading frame**. A projection therefore does not re-merge its derived genomic axis to match the coding one.

### Why

`:42` is a conditional with two conjuncts — "two variants separated by one nucleotide" **and** "together affecting one amino acid". The first is statable on any DNA reference; the second is not statable on a genomic one, so the clause's antecedent is unsatisfied rather than merely weaker. With `:42` not firing, `DNA/delins.md:17` governs unopposed and asks for exactly the individual descriptions:

> two variants separated by one or more nucleotides should be described individually and **not** as a "delins"

So the genomic split is the recommendation's default, not a gap in it.

What makes the axis rather than the biology decisive is `general.md:23`, which makes the prefix mandatory and makes it a claim about the **type of reference sequence used**, with `:26` assigning `g` to "a linear genomic reference sequence". `LRG_199:g.…` is therefore genomic however gene-scoped its accession, and `LRG_199t1:c.…` is the coding one — there is no `NG_`/`LRG_` carve-out to build.

Two supporting clauses, pulling in opposite directions and both cited:

- `general.md:44` scopes the exon/exon junction exception by **naming prefixes** — "using **c.**, **r.** or **n.** reference sequences". The spec does scope rules this way.
- `general.md:43` is the inverse precedent: the 3' rule applies "to ALL descriptions (genome, gene, transcript, and protein)". The spec says so explicitly when it means universal, and neither `:18` nor `:42` does.

### What this corrects

The test pinning the old behaviour mis-cited its own authority, and that mis-citation is what licensed the merge. It offered `delins-codon-carve-out-gap-one` as settling that the exception is "a property of the variant … whichever coordinate system it is spelled in". It does not: that record settles **when** two variants merge — a gap of one, one amino acid — and is silent on which axes. Its `applies_to` is six entries, every one a `c.` or `r.` string with no genomic spelling among them, and its rationale states the opposite of the claim — "The carve-out is **scoped to the coding sequence**". The comment is corrected in place, because left alone it would re-license the merge for the next reader.

### Scope, stated narrowly

Ruled on **DNA axes only**, because `DNA/delins.md` is jurisdiction over DNA descriptions and nothing else.

- `r.` merges on its **own** document's authority, `RNA/delins.md:18`, which states the same exception with its own worked example `r.142_144delinsugg`. Not ruled on here.
- `p.` is the protein consequence itself.
- `n.` is **left open**. It declares a non-coding DNA reference and so carries no frame, yet projection renders it merged — because the axis is derived from the `c.` form by CDS-offset arithmetic rather than normalized on its own reference. Normalizing `LRG_199t1:n.389_391delinsTGG` directly yields `n.[389C>T;391C>G]`. Which side of that is wrong needs its own ruling; the test pins the `n.` row as **observed**, not as adjudicated.

### Changes

- Adds `src/project/codon_exception.rs` and the projector's `apply_codon_frame_exception`: the coding axis inherits a partition decided without a frame, so it is re-normalized against the transcript, which is where `merge::apply_coding_codon_exception` fires. There is deliberately **no** genomic half — a derived `g.` axis is still a `g.` description.
- `axis_genomic_idempotent` **has no exemption list, and the absence is now stated as the guard**. To be exact: it never had one on `origin/main` either — that file contains no `exempt`/`allowlist`/`EXPECTED_NON_FIXED` — so nothing is removed here; the earlier phrasing measured against an intermediate commit of this branch rather than against the base. What this PR adds is a comment block recording *why* there is no list (every projected genomic axis must be a fixed point of bare normalization, with no exceptions) plus a rewritten line explaining why the fixpoint still holds now that the projector withholds the reading frame. Re-introducing the merge fails the test rather than being waved through by a list.
- Adds `rulings[projection-codon-exception-is-decided-by-the-rendered-axis]`, `decided`, governing `DNA/delins.md:42`, with both counterweights recorded (see below).
- Renames `the_genomic_axis_alone_splits_the_specs_own_codon_delins` → `the_codon_delins_merges_only_on_an_axis_that_declares_a_frame`; the old name asserts the thing being overruled. It now pins `split == ['g']` exactly **and** both genomic strings verbatim, so a split landing on the wrong coordinates still fails.

### Counterweight recorded

The exception **sentence** at `:18` carries no axis restriction of its own — it sits in a generic Notes block repeated verbatim in six DNA files — so the scope is inferred from its antecedent, its rationale at `:19`, and its examples, rather than being stated. Measured against that inference — and the measurement is thinner than a raw count of eleven suggests. Of the **11** occurrences of "amino acid" across `DNA/*.md`, **seven sit with no description at all**: the six verbatim copies of the exception sentence itself (`delins.md:18`, `deletion.md:19`, `substitution.md:17`, `insertion.md:20`, `inversion.md:21`, `duplication.md:23`) plus the Q&A prose at `delins.md:81`. Only **four** sit with a description, and all four are `c.` — `delins.md:42` and `substitution.md:37` (both `c.[145C>T;147C>G]`), `deletion.md:145`, `insertion.md:66`. **None** sits with a `g.`, `n.` or `m.` one, in a document set that otherwise uses `g.` freely (`DNA/delins.md` alone carries four genomic worked examples). So the axis inference rests on four rows rather than eleven, which is what the ledger record states and what the counterweight is about.

### Representation change

The `g.` axis does **not** move against `main`. `Status::ProjectionSplitsSingleMember` is 9 and `Status::ProjectionPinned` is 1168 on both sides — the `7 -> 9` / `1170 -> 1168` figures an earlier draft of this description quoted are movement against an intermediate commit of this branch, not against `main`. `spec_enumeration_tests.rs` says so in its own committed comment: "#1672 holds this at 9; it moves no row."

What moves is the **transcript axes** of a codon delins authored genomically. On the synthetic `REF`/`tx` fixture, `REF:g.4_6delinsATA` projects as:

| axis | was | now |
|---|---|---|
| `g.` | `g.[4C>A;6G>A]` | unchanged (#79's scope) |
| `c.` | `REF(tx):c.[4C>A;6G>A]` | `REF(tx):c.4_6delinsATA` |
| `n.` | `REF(tx):n.[4C>A;6G>A]` | `REF(tx):n.4_6delinsATA` |
| `r.` | `tx:r.[(4c>a;6g>a)]` | `REF(tx):r.(4_6delinsaua)` |

The `c.`/`n.`/`r.` axes had been rendering the form `DNA/delins.md:42` calls "not correct" on axes whose reading frame ferro has in hand, purely because the partition was decided on the frameless genomic axis and inherited. The `r.` row merges on `RNA/delins.md:18`'s own authority, not the DNA clause's.
### Verification

```
cargo nextest run --features dev   10173 tests run: 10173 passed, 153 skipped
cargo fmt --all --check            0
cargo clippy --all-features --all-targets -- -D warnings   0
generate_spec_fixture / generate_spec_enumeration          0 / 0
```

Representation-Change: 3 rows move, all on the transcript axes of a codon delins
  authored on the genomic axis, whose partition was decided without a frame and
  then inherited. `REF:g.4_6delinsATA` projects as `c.4_6delinsATA` (was
  `c.[4C>A;6G>A]`), `n.4_6delinsATA` (was `n.[4C>A;6G>A]`) and
  `r.(4_6delinsaua)` (was `r.[(4c>a;6g>a)]`). Toward the form `DNA/delins.md:42`
  asks for on an axis that declares a reading frame; the `r.` row merges on
  `RNA/delins.md:18`'s own authority. The `g.` axis does NOT move -- it stays
  `g.[4C>A;6G>A]`, which is #79's scope, and the projection census constants are
  byte-identical to main (`ProjectionSplitsSingleMember` 9, `ProjectionPinned`
  1168), so the two `LRG_199t1` spec rows this PR argues about are unchanged
  rather than reversed. Previously-accepted inputs, so a migration on paper for
  anyone storing a projected `c.`/`n.`/`r.` string of this shape. Normalization
  itself is untouched.


nh13:
@coderabbitai pause

PR #1673, opened 2026-08-11
fix(fixtures): restore transcript exon ends and guard the Exon contract

### LINKED ISSUE #1619: hgvs_to_spdi resolves a c. position through exon gaps while the sequence it indexes is flat

nh13:
`tests/fixtures/sequences/normalization_transcripts.json` violated the `Exon` contract in every multi-exon record, and nothing checked it. This restores the data, adds the guard, and corrects two doc claims that were keeping the defect plausible.

Representation-Change: none. No file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched — the only `src/` change is a doc comment in `src/convert/mapper.rs`. Ferro's behaviour on any real reference is byte-identical; what moves is the coordinates the test fixture feeds it, and real cdot transcripts never had the shape being removed.

## The contract

`src/reference/transcript.rs` documents `Exon.start`/`Exon.end` as "position in **transcript** coordinates (1-based, inclusive)". Transcript coordinates address the *spliced* transcript, so consecutive exons must **abut**: the list tiles a contiguous run of transcript space with no holes and no overlaps, and a coordinate falling between two exons is not an intronic position, it is an unaddressable one. Neither *endpoint* of that run is anchored — 5' start and 3' end are both left free, each for its own measured reason (below). The invariant is the junctions. That is also what the upstream producer emits: cdot's `cds_start_i`/`cds_end_i` are 1-based inclusive transcript bounds, consumed as such in `src/data/cdot.rs`.

## Measured, not taken on trust

**22 of 29 records violated it** — confirmed, and all on the single contiguity invariant, every gap exactly one base. Numbering, ordering, non-emptiness and `start == 1` were all clean.

**The 7 that passed are a structural zero.** All seven are single-exon. Contiguity is a property of junctions, so they had nothing to get wrong: *every* record with a junction violated it, and "7 pass" was an artifact rather than evidence. The guard's non-vacuity half now asserts the junction count so this cannot recur silently.

**Six of those seven were not real single-exon transcripts** — they were flattenings, and the first revision of this PR left that unmentioned, so the new guard passed them vacuously while nothing said the tables were wrong. Re-derived from `cdot-0.2.32.refseq.GRCh38`: `NM_001089.2` has **33** exons, `NM_033517.1` **22**, `NM_020732.3` **20**, `NM_024649.4` **17**, `NM_000382.3` **10**, `NM_002055.4` **9** — 111 exons and 105 junctions erased, while each record's `sequence` was the real transcript.

**All six are now restored** to cdot's table exactly. The fixture goes 22 -> 28 multi-exon records and 412 -> 517 junctions, and every record in it now carries its real cdot exon structure except `NM_001127687.1`, which cdot does not carry at all — that one is named in `FLATTENED_RECORDS` with its reason, and the new `no_unnamed_record_is_flattened` makes the single-exon set an exact match against that list, so a second flattening costs a line and a stated reason.

### `NM_033517.1` has a real cdot gap, and it is exempted by name

Five of the six restore to contiguous tables, which is behaviourally inert: a contiguous table makes the mapper's walk and its flat arithmetic agree. `NM_033517.1` is different and worth reading carefully. cdot gives it a **genuine** 39-base transcript-coordinate hole between exon 10 (ends 1302) and exon 11 (starts 1342) — it is one of the 58 gapped builds this PR counts.

That gap is pinned in `CDOT_GAP_JUNCTIONS` **by accession, junction and exact size**, never by loosening the contiguity rule — a general relaxation would re-admit the one-base synthetic holes this file exists to keep out. The pin is verified exact: changing the pinned size from 39 to 40 turns both `check_exon_contract` and `every_pinned_gap_exemption_is_still_load_bearing` red. That second test also keeps the list **shrink-only** — an exemption matching nothing fails, so restoring contiguity means deleting the row rather than leaving dead cover behind.

**Carrying it re-arms #1619 in the corpus, deliberately.** With `FERRO_ASSERT_SEQUENCE=1` the existing case `NM_033517.1:c.4818dupC` -> `c.4818dup` fires: the input applies `C`, the output applies `T`, at transcript position 4877, because `hgvs_to_spdi` walks the exon list across the hole while the normalizer indexes the flat transcript. This is the reproducer #1619 would otherwise have lost when the fixture was repaired, now demonstrated on **real annotation** rather than on an artifact.

**No CI job arms the oracle where it fires** — verified, not assumed. `FERRO_ASSERT_SEQUENCE` is set in exactly two places: `ci.yml`'s `sweeps`, which selects `SWEEP_FILTER + test(issue_1615_denoted_sequence_oracle)`, and `nightly-mutalyzer.yml`, which selects the three reference-aware modules and carries `continue-on-error`. `normalize_tests` is in neither selection. The row is recorded as a deferred entry on `test-oracle` in both `ci.yml`'s comment and `CLAUDE.md`, replacing the old `NM_000492.4` row on the same issue.

**The fixture is real cdot with every exon `end` decremented by one.** Checked exon-by-exon against `cdot-0.2.32.refseq.GRCh38`: all 434 exons of the 22 records have a `start` matching cdot exactly and an `end` exactly one less. That identifies it as an inclusive/exclusive off-by-one on `cds_end_i`, and makes the repair exact rather than a guess. The diff is 434 lines, all `"end"` fields.

**The "20 of 20 real transcripts have zero gaps" claim is directionally right but not absolute** — I measured 27 of the 28 fixture accessions present in cdot, and across the full file, **58 of 474,818 multi-exon builds** do carry a transcript-coordinate gap. Not a structural zero: the property varies, and all 28 fixture accessions are multi-exon (16–62 junctions). What cdot never produces is the fixture's shape — smallest real gap **23 bases**, **zero** one-base gaps, and only one build gapped at every junction.

## Fixture or contract?

**The fixture.** The contract is stated correctly and matches the upstream producer; the fixture's shape does not occur in 474,818 real transcripts; the starts already agreed with cdot, so this is one systematic off-by-one, not a different exon model.

It was doing real damage. On `NM_000492.4`, resolving `c.` by walking the exon list lands 10 bases 3' of the flat-transcript answer by `c.1520` — the denoted-sequence oracle row recorded against #1619.

**The before/after, with the command it comes from.** The first revision quoted "25 before, 22 after" with no way to re-derive it, and called the three tests that went green "exactly the three suites that read this fixture" — which is wrong: three *files* read it, and the green tests come from **two** of them (`normalize_property_tests` reads it and contributes none). Re-measured against `origin/main` at `d5f26fcb`, swapping only the fixture and excluding the guard this PR adds, since that guard is *about* the fixture and cannot be a term in a before/after on it:

```bash
git show <base>:tests/fixtures/sequences/normalization_transcripts.json \
  > tests/fixtures/sequences/normalization_transcripts.json
FERRO_ASSERT_SEQUENCE=1 \
```

**25 failures before, 22 after, zero new**, the three green ones being `idempotency_tests::test_normalization_idempotency`, `normalize_tests::…::test_3prime_shifting::case_1` and `normalize_tests::…::test_deletion_shifts_in_real_homopolymer::case_1`. The totals move with the base — the same measurement read 20/17 against `5f22abed` a few hours earlier — so quote the command, not the pair. (Measured before `NM_033517.1` was restored; restoring it adds back the one #1619 row described above, by design.)

**This does not close #1619, and the PR is careful not to claim it does.** The `hgvs_to_spdi`-vs-normalizer asymmetry the issue names is still reachable on the 58 real transcripts above; what the repair removes is the fixture artifact that was standing in for it — which also means nothing in the suite exercises it any more. `CLAUDE.md` records that, so the row is not read as fixed.

## The guard

`tests/it/normalization_transcripts_exon_contract.rs` checks every record: consecutive 1-based numbering, non-empty well-ordered bounds, junction contiguity, and last exon within the stored sequence. It reports all violations at once rather than the first.

Coverage is deliberately a one-sided bound, not an equality: `NR_046285.1` and `NR_153405.1` have genuine cdot exon tables that stop short of the true RefSeq length (2599/2619, 4299/4337) because their 3' bases do not align to the genome.

**The 5' anchor was dropped for the same reason, after measuring it.** The first revision asserted `exons[0].start == 1` — a two-sided claim with no measurement behind it, where the 3' bound had one. Over cdot-0.2.32.refseq.GRCh38, **109** builds begin past transcript coordinate 1 (`NM_000280.5` at 8, `NM_000910.2` at 7, `NM_001100631.1` at 870), **1.9x** the 58-build gapped population the argument rests on — so the anchor would reject genuine upstream data almost twice as often as the synthetic shape it was meant to catch. No consumer wants it either: `CoordinateMapper` switches between flat and exon-walking arithmetic on `w[0].end + 1 != w[1].start`, a junction predicate, and nothing under `src/` reads `exons[0].start`. Every fixture record still starts at 1; that is now recorded as an observation, not a contract.

**Verified non-vacuous by running it against the pre-fix fixture** — it fails, naming 22 of 29 records, independently reproducing the measurement through the test itself.

## Doc corrections

`make_transcript_with_gaps` claimed to "simulate how cdot encodes transcripts with virtual intron positions". cdot has no such encoding, and that false claim was the stated justification for the fixture's shape. Its doc now says what it actually builds — a deliberately malformed transcript stressing the mapper's exon-walking arithmetic — and why it stays that way.

## Second commit: declined protein splits re-parse

Separately, `try_protein_split_delins` (#1606) declines on seven conditions, one of which exists *only* to stop normalization emitting the unparseable `p.Met1Xxx`. That guard's whole value is a re-parse property and nothing asserted it — the existing suite pins exact output strings, which pass just as happily on a string that cannot be read back.

Investigated whether #1606 re-opened a retired "declined protein split reparse hole": **it did not.** The decline arm is `None => variant`, which returns the input value and produces no string, so a decline has no re-parse property to violate. Measured over a 4,347-input sweep with all nine decline sites instrumented — every branch fires (no structural zero), the split fires 320 times, and the 16 unparseable outputs found all enter already collapsed to a substitution and leave on the "not a delins" gate. Disabling the split reproduces all 16 byte for byte while moving 154 other outputs, which is the decisive A/B. Those 16 are #1607 and are deliberately not pinned here.

## Verification

`cargo fmt --all` and `cargo clippy --all-features --all-targets -- -D warnings` are clean, `actionlint` is clean on the `ci.yml` edit, and both spec generators were regenerated first.

The oracle-armed suite reports `10034 tests run: 10027 passed, 7 failed, 151 skipped`. **Every one of those failures is pre-existing on `origin/main` and none is attributable to this PR** — verified by building a throwaway detached worktree at pristine `origin/main` and running the same tests with the same three flags, where the identical set fails. This PR's only `src/` change is a doc comment, so it cannot move normalization behaviour; `git diff origin/main...HEAD -- src/` has no non-comment line.

## Review round: what changed after review

Four verified findings plus two nits, all fixed. In order of seriousness:

1. **The seven single-exon records were fabricated flattenings** and nothing said so — all six that cdot carries are restored, `NM_001127687.1` is named in `FLATTENED_RECORDS`, and `no_unnamed_record_is_flattened` + `every_pinned_gap_exemption_is_still_load_bearing` are the new guards. See *Measured, not taken on trust* above.
2. **`exons[0].start == 1` was a two-sided assertion with no measurement behind it** — dropped, with the 109-build measurement recorded. See *The guard*.
3. **The before/after figure shipped without a command**, and the "three suites" claim was wrong — both fixed here and in `CLAUDE.md`.
4. **`try_protein_split_delins`'s `Ter || Xaa` guard was only exercised on `Ter`.** The `Xaa` arm is added on both sides, and each is verified discriminating by deleting the corresponding `contains(&AminoAcid::Xaa)` from the production guard and confirming the test goes red (payload side emits `p.[Ala2Gly;Leu4Xaa]`, reference side `p.[Ala2Gly;Xaa4Gly]`).
5. Nits: `PROTEIN_WITH_STOP_AT_4`'s doc said residue 4 was "replaced" when the stop is **inserted** (21 residues to `PROTEIN`'s 20, so `Leu` is residue 5) — corrected, with the same caveat on the new `PROTEIN_WITH_UNKNOWN_AT_4`. And `parse_hgvs(&output).is_ok()` was not a round-trip; it now compares `parse_hgvs(&output)?.to_string()` against `output`, so an output the parser reads back as a *different* description fails.

**Reported, not fixed:** eight sibling fixtures carrying exon tables (`*/reference-windows.json`, `grammar/spec_enumeration_windows.json`, `transcripts/mock_transcripts.json`, `mutalyzer-normalize/genomic-windows.json`) hold **2,306** junctions this guard does not cover. All measured clean — zero contiguity violations — so it is a coverage gap, not a live defect. The 58/474,818 cdot census is also restated in three files (`CLAUDE.md`, `src/convert/mapper.rs`, the guard) with no committed derivation; it re-derives from the prepared cdot on request.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1016,11 +1016,27 @@ jobs:
         #     (`repeat_tract_maximization`). `hgvs_to_spdi` reads the anchored
         #     spelling as 6 copies replacing a 1-copy tract and the normalizer's
         #     output as 6 copies replacing a 2-copy tract — 14 bases against 12.
-        #   * #1619 — `NM_000492.4:c.1520_1522del` -> `c.1521_1523del`
-        #     (`normalize_tests`). The two forms are genuinely equivalent on the
-        #     flat transcript; `hgvs_to_spdi` resolves the c. position by walking
-        #     the fixture's exon list, whose 1-base synthetic introns put it 10
-        #     bases 3' of where the normalizer reads.
+        #   * #1619 — `NM_033517.1:c.4818dupC` -> `c.4818dup`
+        #     (`normalize_tests`). `hgvs_to_spdi` resolves the c. position by
+        #     WALKING the exon list while the normalizer indexes the FLAT
+        #     transcript, so the two disagree across any transcript-coordinate
+        #     gap: here the input applies `C` and the output applies `T` at
+        #     transcript position 4877.
+        #
+        #     This row replaced an earlier one on the same issue
+        #     (`NM_000492.4:c.1520_1522del` -> `c.1521_1523del`), and the swap is
+        #     the point rather than a detail. That one rode on a FIXTURE DEFECT —
+        #     `normalization_transcripts.json` held cdot's exon table with every
+        #     `end` decremented by one, giving a 1-base hole at every junction, a
+        #     shape that occurs nowhere in 474,818 real builds. Repairing the
+        #     fixture removed it. This row rides on cdot's own annotation:
+        #     `NM_033517.1` genuinely has a 39-base hole between exon 10 and exon
+        #     11, one of the 58 gapped builds in cdot-0.2.32.refseq.GRCh38. So
+        #     #1619 is now demonstrated on real geometry rather than on an
+        #     artifact, which makes it a STRONGER reason to keep the flag out of
+        #     this job, not a weaker one — and it is the reproducer the issue had
+        #     otherwise lost. See `CDOT_GAP_JUNCTIONS` in
+        #     `tests/it/normalization_transcripts_exon_contract.rs`.
         #
         # Turning the flag on here before those are settled would make the job
         # red for a reason that is not the PR's; suppressing them would hollow out
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -336,11 +336,88 @@ real disagreements inside ferro rather than noise:
 | row | what disagrees |
 |---|---|
 | #1618 — `NC_TEST.1:g.262TG[6]` → `g.259_262GT[6]` | `hgvs_to_spdi` reads the anchored spelling as 6 copies replacing a **1**-copy tract, the normalizer's output as 6 copies replacing a **2**-copy tract — 14 bases against 12 |
-| #1619 — `NM_000492.4:c.1520_1522del` → `c.1521_1523del` | the two forms are equivalent on the flat transcript; `hgvs_to_spdi` resolves the `c.` position by walking `normalization_transcripts.json`'s exon list, whose 1-base synthetic introns land it **10 bases** 3' of where the normalizer reads |
+| #1619 — `NM_033517.1:c.4818dupC` → `c.4818dup` | `hgvs_to_spdi` resolves the `c.` position by **walking** the exon list while the normalizer indexes the **flat** transcript, so the two disagree across any transcript-coordinate gap: the input applies `C`, the output applies `T`, at transcript position 4877. `NM_033517.1` carries a real 39-base cdot hole between exons 10 and 11 — see below, because this row **replaced** an earlier one on the same issue |
 
 Suppressing either would hollow out the oracle, and turning the flag on before
 they are settled would redden the job for a reason no PR caused. Move it into
 `test-oracle` once both are closed.
+
+**#1619's row was replaced, not closed, and the swap is the whole point.** The
+old row (`NM_000492.4:c.1520_1522del` → `c.1521_1523del`) rode on a **fixture
+defect**, not on annotation: `normalization_transcripts.json` was the real cdot
+exon table with every exon's `end` decremented by one, so all 22 of its
+multi-exon records carried a one-base hole at every junction — a shape that
+occurs nowhere in 474,818 real builds. Repairing the fixture removed that row.
+
+The row that stands in its place rides on **cdot's own annotation**.
+`NM_033517.1` is one of the 58 gapped builds, with a genuine 39-base hole
+between exon 10 (ends 1302) and exon 11 (starts 1342), and the fixture now
+carries its real table. So #1619 is demonstrated on real transcript geometry
+rather than on an artifact — which is a *stronger* reason to keep
+`FERRO_ASSERT_SEQUENCE` out of `test-oracle`, and it hands the issue the
+reproducer the repair would otherwise have destroyed. The gap is exempted by
+name in `CDOT_GAP_JUNCTIONS`
+(`tests/it/normalization_transcripts_exon_contract.rs`), never by loosening the
+contiguity rule.
+
+**No CI job arms the oracle where it fires** — checked, not assumed.
+`FERRO_ASSERT_SEQUENCE` is set in exactly two places: `ci.yml`'s `sweeps`, which
+selects `SWEEP_FILTER + test(issue_1615_denoted_sequence_oracle)`, and
+`nightly-mutalyzer.yml`, which selects the three reference-aware modules and is
+`continue-on-error`. `normalize_tests` is in neither selection, so the fire is
+reachable only by setting the flag by hand.
+
+**Here is the command the before/after comes from**, because a figure with no
+command is one nobody can re-derive. Swap only the fixture, and exclude the guard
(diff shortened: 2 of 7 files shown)


nh13:
@coderabbitai pause

Issue #1196, opened 2026-07-25
fix(error-handling): six ErrorTypes never reach a config consult site, so --error-mode and --ignore/--reject are inert for them


nh13:
Six of the 45 `ErrorType` variants never reach a configuration consult site, so `--error-mode` and the `--ignore`/`--reject` overrides are silently inert for them. This is the same class as #1181 (a documented knob that does nothing), but inside the library rather than at a CLI seam.

The audit that should catch this — `tests/it/error_code_audit.rs` — checks that an `ErrorType` variant is **mentioned** in an emission-relevant path, not that the mode **reaches** it, so it passes on all six.

## How this was measured

The four consult points are `ErrorConfig::action_for` / `should_reject` / `should_correct` / `should_warn` (`src/error_handling/mod.rs:152,158,163,168`). A variant's behavior can be mode-dependent by being passed to one of those directly, or by being attached to a `Correction { error_type, .. }`, which the preprocessor resolves generically (`preprocessor.rs`, the `c.error_type` arms). Every variant was checked against both paths.

## The six

**A. Hard parse error; the mode is never consulted.** The `ErrorType` is used only to source the W-code string for the message, then an error is returned unconditionally:

- `AlleleFractionAnnotation` (W3017) — `src/hgvs/parser/variant.rs:7079`, `let _w3017 = ErrorType::AlleleFractionAnnotation;`
- `ClinVarProseMultiAllelic` (W3018) — `src/hgvs/parser/variant.rs:6595`, `let _w3018 = ErrorType::ClinVarProseMultiAllelic;`
- `NonSpecMosaicForm` — `src/hgvs/parser/variant.rs:6666`, `let et = ErrorType::NonSpecMosaicForm;` (used only for `et.code()` / `{:?}` in the message)

The first two are underscore-prefixed unused bindings. Both carry a comment stating they exist to satisfy the audit's scan — see "Root cause" below.

**B. Warning emitted unconditionally.** `InitiatorMetCanonicalization` is pushed at `src/normalize/mod.rs:3998` with no `should_warn` gate, so it cannot be silenced in `silent` mode or promoted with `--reject`.

**C. Registered with no behavior anywhere.** `InvalidUnicodeCharacter` and `DeprecatedFrameshiftStar` appear only in the PyO3 `ErrorType` ↔ `PyErrorType` mapping. Neither is emitted or consulted anywhere in `src/`.

## Reproduce

[...]

PR #1674, opened 2026-08-11
fix(error-handling)!: remove the unwireable mode-level warning predicates, and stamp the mode censuses are measured under

### LINKED ISSUE #1629: ErrorMode::emits_warnings() has no call site, so silent-vs-lenient is a per-code accident rather than a mode
`ErrorMode` documents three modes, and the middle one is defined by its diagnostics: lenient "auto-correct common errors **with warnings**", silent "auto-correct **silently without warnings**". `ErrorMode::emits_warnings()` (`src/error_handling/types.rs:46`) encodes exactly that — true for lenient, false for strict and silent.

**It has no call site in `src/` outside its own unit test.**

```
$ grep -rn "emits_warnings" src/
src/error_handling/types.rs:46:    pub fn emits_warnings(&self) -> bool {
src/error_handling/types.rs:1136:    fn test_error_mode_emits_warnings() {
src/error_handling/types.rs:1137-1139:  (the three assertions)
```

## What actually decides suppression today

Each code routes through `ErrorConfig::action_for` -> `ErrorOverride::resolve(mode)` -> `ResolvedAction`, and whether a given warning honours silent depends on whether someone wired a `should_warn_*` helper for it. So the distinction is real for some codes and absent for others:

- **honours silent** — W4007 `IntronicOnBareTranscript`, via `NormalizeConfig::should_warn_intronic_bare_transcript`. Measured on the 16 bare-`NM_` intronic rows of `conformance::spec_corpus`: lenient returns the warning, silent returns an empty warning vector, and the normalized output is identical.
- **does not honour silent** — W5005, `REFSEQ_MISMATCH` and `INSERTED_SEQUENCE_EXPANDED`, per #1580's own "found while here, deliberately not fixed" list. W4004 does.

So "silent" is a per-code accident rather than a mode. A caller cannot predict from the mode alone whether they will get diagnostics.

## Why it matters beyond tidiness

#1580 made normalizer warnings reachable at all — before it, 100% were dropped on the floor. Having made them reachable, the mode that exists to suppress them has to actually suppress them, or `--error-mode silent` is not a contract. This project's disclosure obligations run through knowing which repairs happened; a mode whose diagnostics behaviour is per-code is one nobody can reason about.

It is also the prerequisite for the third arm of `rulings[absolute-prohibition-enforcement-stage]`, which distinguishes silent from lenient by exactly this property — "lenient with no error messages, but an exit code".

## Suggested shape

Either give `emits_warnings()` a call site — one gate at the point warnings are collected, so the mode decides and no code can opt out by omission — or delete it and document that suppression is per-code, so nothing reads as a mode-level guarantee that is not one. The current state is the worst of the two: a helper that states the contract and enforces none of it.

## Not a regression

Nothing here is new breakage; it is a gap #1580 named and scoped out on purpose so its byte-identity claim stayed verifiable.

nh13:
Closes #1629

Representation-Change: none. The one edit under `src/normalize/` replaces `self.error_config.mode == ErrorMode::Strict` with the identical `self.error_config.mode.is_strict()`; the census additions are test-side provenance. Both pinned conformance censuses are unchanged in both directions.

### 1. `ErrorMode::emits_warnings()` has no call site — **confirmed**, and it is dead code rather than a removed caller

`git log -S emits_warnings -- src/` returns exactly one commit: the initial one. So nothing was ever wired and later unwired. The sibling `allows_correction()` is the same shape and equally unreferenced; only `is_strict()` had a real (inlined) caller.

**The repair the issue leans toward is not available.** #1629 suggests "give `emits_warnings()` a call site — one gate at the point warnings are collected, so the mode decides and no code can opt out by omission". Warning emission is resolved *per code*, by `ErrorConfig::action_for(code) = overrides[code].resolve(mode)`. A gate keyed on the mode would:

- silence a code an operator explicitly asked to be warned about (`WarnCorrect` override under `Silent`), and
- speak for a code they explicitly silenced (`SilentCorrect`/`Accept` override under `Lenient`),

i.e. delete the `--ignore`/`--reject` knob (#1196) it was placed next to. It was also **wrong on a shipped code**: W3022 `InitiatorMetCanonicalization` is emitted under **strict**, deliberately — strict must not refuse ferro's own canonical output — while `Strict.emits_warnings()` answered `false`.

So both predicates are deleted, `is_strict()` is wired to its one legitimate caller (`should_reject_reduced_capability`, an environmental limitation with no `ErrorType` for an override to name), and the counter-examples are pinned in `tests/it/issue_1629_error_mode_is_not_a_warning_gate.rs` plus a unit test, so a mode-level predicate cannot return unargued.

### 2. Censuses are unlabelled as lenient — **confirmed**

Surveyed every generator under `examples/`. **No artifact in the tree records the `ErrorMode` it was generated under.** The nearest thing is a prose comment line in `tests/fixtures/inversion-sweep/cases.tsv` that nothing parses.

The trap is quiet in the expensive direction: `NormalizeConfig::default()` substitutes `ErrorConfig::lenient()`, while `ErrorConfig::default()` and `ErrorMode`'s own `#[default]` are both **strict**. A generator written `Normalizer::new(provider)` therefore measures in lenient mode while reading as though it took the default. `NormalizeConfig` cannot help either — `error_config` is `#[serde(skip)]`, so serializing the whole config emits no mode.

This is not cosmetic. `corpus_prohibited_inputs.rs` re-measures **three** of the spec corpus's four refusal counters in strict mode and **two of them move** — `conflicts_accepted`'s 72 are all refused and 16 of `prohibited_conditional_accepted`'s 40 are, while `prohibited_absolute_accepted`'s 32 stay accepted; the fourth, `prohibition_violating_outputs`, is not re-measured in either mode. One counter that moves is enough, which is why the decided ruling `bare-transcript-intronic-position` has to end with the sentence "so that counter is a lenient-mode figure" — the artifact could not say it.

## What changed

**The stamp.** `conformance::error_mode_stamp::ErrorModeStamp` is derived from the `ErrorConfig` a generator actually built. One constructor (`of(&ErrorConfig)`), no `Default`, and it carries the per-code **overrides** as well as the base mode — a stamp naming only the mode would describe a precondition the run did not have.

**Stamped.** Both spec generators now emit `normalized_under` beside their existing `spec`/`summary` identity block, read off the same `measurement_config()` the normalizer is built from, so the artifact's claim cannot drift from the run that produced it. Named `normalized_under` rather than `generated_under` because generation is not one thing: the parse half applies no mode at all, so only the figures downstream of normalization are figures for that mode. Both artifacts are gitignored and CI-regenerated, so this costs no fixture churn:

```json
"normalized_under": { "error_mode": "lenient" }
```

**The census the ruling cites.** `spec_conformance_axis::Census` gains `measured_under`, set from `measurement_config(direction)` and carried in the existing `assert_eq!(census, pinned)` — so changing what the axis measures under fails naming the mode, instead of silently re-basing seventeen numbers. The failure report now opens with `NORMALIZED UNDER: error mode 'lenient'`.

**The guard.** `generator_completeness::generators_that_normalize_stamp_their_error_mode_or_are_allowlisted`, built on the same shrink-only allowlist pattern as the `CaptureLedger` ratchet beside it, with the same honesty about its reach — substring scans, a floor rather than a proof. Verified to bite in **both** directions before shipping: adding a stamped generator to the allowlist fails (`stale`), and removing an unstamped one fails (`undeclared`). It also carries a vacuity guard, since a predicate that matched nothing would pass every assertion while measuring nothing.

Nine generators are allowlisted with a per-row reason. Six write **committed** `WindowFixture` artifacts, which are `deny_unknown_fields` — adding the field means regenerating fixtures that need a prepared reference, with its own blast radius. `extract_biocommons_windows.rs` is the sharpest case and is worth flagging: it switches to `ErrorMode::Strict` per row when `case.expects_error`, so its artifact is genuinely **mixed-mode** and a single document stamp would misdescribe it.

## Found while here, deliberately not fixed — an operator decision

Two *per-code* predicates are themselves unwired: `NormalizeConfig::should_warn_ref_mismatch` and `should_warn_variant_exceeds_reference` have no call site outside their own unit tests, and `src/normalize/mod.rs:9519` pushes `RefSeqMismatch` unconditionally (only `should_reject_ref_mismatch` is consulted, at the strict gate). This is the sharpest surviving form of #1629's "silent-vs-lenient is a per-code accident".

Wiring them would make `--error-mode silent` suppress **W5001** (`RefSeqMismatch`) and **W5003** (`VariantExceedsReference`) — the codes those two predicates actually resolve; W3001 is `MissingVersion`, a preprocessor code neither touches — a behaviour change on a question `issue_1181_cli_error_mode.rs::strict_is_distinct_from_lenient_and_silent_at_the_cli` records as **open** (it pins `lenient_out == silent_out` today, and says so explicitly). That is an adjudication, not a cleanup, so nothing here touches it; the doc comment there is updated to state the question in its sharper form rather than against the now-deleted predicate.

Also noted, out of scope: `src/reference/annotation/mod.rs` defines a **second, unrelated** `ErrorMode` enum with the same three variant names and its own three inline `== ErrorMode::Strict|Silent` comparisons.

## Verification

Regenerated both spec artifacts first, then:

- `cargo fmt --all` — clean
- `cargo nextest run --features dev` with `FERRO_MANIFEST` set (so the conformance axis tests run rather than skipping green):

```
Summary [ 169.297s] 10152 tests run: 10152 passed (3 slow), 152 skipped
```

Both conformance census tests are in that run and pass unchanged: `three_prime_conformance_census`, `five_prime_conformance_census`.

**This removes two `pub` methods from a published type, and is typed as breaking.** `ErrorMode::emits_warnings()` and `allows_correction()` are re-exported from `ferro_hgvs::error_handling` and are gone here. Both have zero call sites anywhere in this repo — every surviving mention is prose in a doc comment — so nothing in-tree notices, but on a published crate (0.13.1) an external consumer would.

The commit is therefore typed `fix(error-handling)!:` rather than `fix:`. `release-plz.toml`'s own comment states the rule it is relying on — for a 0.x crate release-plz "reserves the minor bump for breaking changes" — so the marker is what moves this to 0.14.0 rather than 0.13.2. `fix(spdi)!: give an identity edit the span it claims` (#1363) is the established precedent for the form, and rendered with git-cliff's `[**breaking**]` marker in the v0.13.0 changelog. Note that release alone does not *prove* the marker caused the minor bump — that cycle also carried `feat:` commits, which force minor via `features_always_increment_minor = true` — so the argument rests on the documented rule plus the precedent for the form, not on the release history by itself.


nh13:
@coderabbitai pause

PR #1676, opened 2026-08-11
test: Add more test coverage based on mutatino testing

geoffjentry:
There was a bunch of concrete stuff that got flagged along the way. We dabbled with clearing some of it out but wound up shifting back to a pure test-only PR. I'll look through the concrete items separately. I'm pretty sure there's still some detritus whining about some of these in the comments, so If a code review scanner flags things referencing pre-existing behavior, there ya go.

**NB** re the test deletions:

 **The two kinds of removal**

1.   Exact duplicates — _acceptor_extended_boundary (−12/−13), _donor_extended_boundary (+6/+7), _near_boundary (+50/+51). Each asserted values that are now literally rows of the table, same function, same expected variant.
2. Strictly weaker interior probes — the rest. test_intronic_consequence_from_cds_pos probed +1, +5, +15, +30, −10; test_validate_cds_pos_normal/_out_of_range/_5utr/_3utr probed new(20), new(-3), utr3(3) and friends. Every one sits inside a region the new boundary pair straddles. A boundary-plus-one pair kills every threshold mutant an interior probe would, and more — an interior value can't distinguish <= 50 from < 50, but 50/51 can.

### DIFF
--- a/src/convert/coding.rs
+++ b/src/convert/coding.rs
@@ -146,36 +146,6 @@ mod tests {
     }
 
     #[test]
-    fn test_validate_cds_pos_normal() {
-        let tx = make_test_transcript();
-        assert!(validate_cds_pos(&CdsPos::new(1), &tx).is_ok());
-        assert!(validate_cds_pos(&CdsPos::new(15), &tx).is_ok());
-    }
-
-    #[test]
-    fn test_validate_cds_pos_out_of_range() {
-        let tx = make_test_transcript();
-        // CDS is 15bp (positions 1-15)
-        assert!(validate_cds_pos(&CdsPos::new(20), &tx).is_err());
-    }
-
-    #[test]
-    fn test_validate_cds_pos_5utr() {
-        let tx = make_test_transcript();
-        // 5' UTR is 5bp (positions -1 to -5)
-        assert!(validate_cds_pos(&CdsPos::new(-3), &tx).is_ok());
-        assert!(validate_cds_pos(&CdsPos::new(-10), &tx).is_err());
-    }
-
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
Note from rebasing this onto current `main` (`7d343c86` → `7fad87a3`), in case it is not intended.

Against its own base, this branch **removes seven existing tests** from `src/convert/noncoding.rs` while adding three functions (net −45 lines in that file):

```
test_intronic_consequence_from_cds_pos
test_intronic_region_from_offset
test_validate_tx_pos_with_large_offset
test_intronic_consequence_from_tx_pos
test_intronic_consequence_acceptor_extended_boundary
test_intronic_consequence_donor_extended_boundary
test_intronic_consequence_near_boundary
```

Some of that looks like deliberate consolidation into table-driven rows, which is reasonable — but the two `*_boundary` tests pinned the 6-rung and 12-rung thresholds specifically, and I could not find equivalent coverage of those exact rungs among the additions. Worth a second look given the PR's stated purpose.

Two mechanical notes from the rebase itself:

- `main` gained `test_intronic_region_from_offset_declines_an_unknown_offset` (the #1841 sentinel guard) after this branch was written. I preserved it — it is not one of the deletions above.
- #1841 changed `IntronicRegion::from_offset` to return `Option`, so one assertion in the new table-driven test needed `Some(...)`. Every row there is an ordinary offset, so the wrap preserves intent.

Also, the title still reads "mutatino testing" — GitHub builds the squash subject from the PR title, so that typo will land on `main` as-is.

Issue #1675, opened 2026-08-11
Repeat contracted by one unit is rewritten as a deletion, though the spec prefers the repeat form


nh13:
## What's wrong

A repeat contracted by exactly one unit is rewritten as a deletion, but the spec states a preference for the repeat form.

Reference tract of 15 `GT` copies at `g.258_287` (synthetic `A` + `GT`×15 + `A`, core base 1 at `g.257`):

```
in   NC_TEST.1:g.258GT[14]
out  NC_TEST.1:g.286_287del

in   NC_TEST.1:g.258GT[17]
out  NC_TEST.1:g.258_287GT[17]     <- expansion keeps the repeat form
```

`RNA/repeated.md:33` says, of a 15-unit reference tract:

> when the repeat is variable in the population and the reference sequence has 15 units, the description `r.-123ug[14]` is **preferred over** `r.-97_-96del`

and the sibling note says `r.-123ug[17]` is preferred over `r.-99_-96dup`.

So the spec prefers the repeat description in **both** directions. ferro keeps it for expansions and discards it for the single-unit contraction. That asymmetry is the defect: the expansion case proves this is not a uniform policy of preferring `del`/`dup`.

**No bases are lost** — `g.286_287del` denotes the correct sequence, confirmed by `compare_denoted_sequences`. This is a rule 2 (recommended form) issue only, not a rule 1 or rule 3 one.

## Where

`src/normalize/rules.rs`, the `specified_count < ref_count` branch. The gate is `k >= 2 && specified_count >= 1 && !codon_blocks_repeat`, so a one-unit reduction with surviving units falls through to `RepeatNormResult::Deletion`. Its comment cites "HGVS prioritization: deletion outranks unranked repeat[0]", which holds for a *full tract removal* (`specified_count == 0`) but is what needs re-examining for a contraction with survivors.

## The question this needs decided before it is fixed

**The only citation is RNA-axis.** `RNA/repeated.md:33` is in the RNA chapter; the reproducer above is genomic. The DNA chapter carries its own Discussion section (the CFTR intron 9 Q&A) and states no such preference anywhere.

Applying an RNA-stated preference to `g.`/`c.` is a cross-axis generalization, which is the same class of move that produced #1664. Arguments both ways:

[...]

PR #1678, opened 2026-08-11
fix(normalize): carry the reference copies a re-phased repeat absorbs

### LINKED ISSUE #1618: Anchored repeat: hgvs_to_spdi and the normalizer disagree on what the copy count is counted against

nh13:
An anchored repeat's `[N]` counts copies of the unit the caller **spelled**, against that unit's own tract. `normalize_repeat` maximizes the tract by searching unit rotations, so a unit tiling a short tract is replaced by a rotation tiling a longer one — but the count was carried over unchanged, and the reference copies the wider window swallowed were silently deleted.

```
reference GTGT at g.259..262   —   2 copies of GT, but only 1 copy of TG

in   NC_TEST.1:g.262TG[6]        denotes GTGTGTGTGTGTGT   (14 bases)
out  NC_TEST.1:g.259_262GT[6]    denotes GTGTGTGTGTGT     (12 bases)   <- two lost
```

## Why `GT[7]` is the answer, established without assuming either component

The denoted-sequence oracle reaches bases through `hgvs_to_spdi`, which is one of the two components #1618 says are in disagreement — so using it to judge the normalizer assumes the reading in dispute. `RNA/repeated.md:33` supplies an independent yardstick instead: against a reference tract of 15 units, `[14]` is *preferred over* deleting one unit (net −2) and `[17]` over a dup (net +4). Measured on a 15-copy synthetic tract, `hgvs_to_spdi` gives net −2 and net +4 — it matches the spec, so it is sound as the baseline for what an input denotes.

With that settled, the input denotes `GT`×7 and the correct output is `g.259_262GT[7]`.

## The spec states the conservation this restores

`DNA/repeated.md:97-99`, on this exact `GT`/`TG` pair. Asked about a reference of 11 `TG` copies followed by 7 `T`:

> First, note that by applying the **3'rule** it is a **variable GT and not a TG stretch**. When the coding DNA reference sequence has 11 TG copies followed by 7 T copies, the reference allele is described as `c.1210-33_1210-6GT[11]T[6]`.

Re-phasing preserves the copy count (11 `TG` → 11 `GT`), **and** because the re-phased tract slid one base 3' onto the neighbouring run, the spec decrements that run `T[7]` → `T[6]` so the total is conserved (1 + 22 + 6 = 29 = 22 + 7). Ferro widened and carried nothing.

## The fix

The rotation search now either carries the reference copies it swallows or declines the re-phase outright, in `rebase_count_onto_tract`.

- **Copies, not bases.** The first cut measured the two window edges independently (`(ls - s) + (e - le)`, over the unit length), which cannot see a tract that NARROWS and so reports a gain for a loss. It is replaced by `N + winner_copies - literal_copies`.
- **The literal tract is the baseline even when it does not span the anchor.** That is exactly this shape: the `TG` tract ends *at* the anchor, so the spanning filter rejects it as a seed — which is why its copies went missing in the first place.
- **A pure re-phase absorbs nothing** and keeps its count, so `repeated.md:97`'s 11 `TG` → 11 `GT` is untouched. `a_pure_re_phase_preserves_the_copy_count` guards that boundary.
- **A DISJOINT re-phase is declined, not re-based.** Where the spelled unit's tract and the winning rotation's do not overlap, the description's window moves to a different stretch of reference and re-basing conserves the net length and nothing else: on `AATGTGTGGTAA`, `g.265TG[6]` re-based onto the 1-copy `GT` tract is `g.265_266GT[4]`, which denotes `TGTGTGGTGTGTGT` where the input denotes `TGTGTGTGTGTGGT`. The count instead stands against the tract it was written for. The same test closes the arithmetic's saturating region: `g.265TG[1]` is `1 + 1 - 3`, which the shipped `saturating_sub` clamped to `0`, routing to a two-base deletion where the input denotes four bases fewer — two bases lost, in the arm this rule added, with `[0]` never emitted so nothing downstream could catch it. The arithmetic is now `checked_add`/`checked_sub` and an underflow declines.

## The shape the first declaration missed: the emitted edit KIND moves

Every worked example above is a re-phase where the spelled unit's tract does *not* seed the maximization — it either fails the span filter or is disjoint from the winner. There is a third shape, and it is the one a consumer is least likely to anticipate: the spelled unit's tract **spans** the anchor, so it *does* seed `best`, and a strictly longer rotation that still **overlaps** it displaces the seed. The count is then re-based, and the re-based count changes which edit is rendered.

On `A` + `TGTGTGTG` (g.258..265) + `A`, anchored at g.259 with unit `GT` — a 3-copy `GT` tract inside a 4-copy `TG` tract — the re-base is `N + 4 - 3 = N + 1`. A repeat whose count sits one unit either side of the reference tract's count is rendered as `del` or `dup` rather than as a repeat, so shifting the count by one shifts that boundary from spelled counts 3/4/5 to 2/3/4:

```
             origin/main            this branch
g.259GT[2]   g.258_265TG[2]    ->   g.264_265del      repeat -> del
g.259GT[3]   g.264_265del      ->   g.259GT[3]        del    -> repeat, left alone
g.259GT[4]   g.259GT[4]        ->   g.264_265dup      repeat -> dup
g.259GT[5]   g.264_265dup      ->   g.258_265TG[6]    dup    -> repeat
```

The new values are the correct ones — `origin/main` denoted two bases fewer than the input on seven of these eight counts, and this branch agrees with the input on all eight — but "the count now carries the copies the wider window swallows" does not predict a stored `repeat` coming back as a `del`. The class is now pinned by three tests and called out in the representation-change declaration.

## Why it survived

Three of the four seam oracles pass on the wrong output — it is well-formed, in bounds, re-parses, and *is* a fixed point. Only `FERRO_ASSERT_SEQUENCE` can see it, and this is one of the two rows keeping that flag out of CI's `test-oracle` job. Closing it leaves #1619 as the last blocker there.

## An existing test was pinning the wrong thing

`repeat_tract_maximization.rs` had `anchored_and_range_spellings_agree_on_the_maximal_tract`, which required `g.262TG[6]` and `g.259_262GT[6]` to normalize identically. Those denote **14 and 12 bases**. Converging them is not two spellings of one variant agreeing — it is one of them losing data, asserted as if it were confluence. It is replaced by a test that both reach the maximal *window* while correctly differing in count. `CANONICAL` moves `6` → `7`.

## Tests

Written first, watched fail, then implemented. RED was 5 failing / 6 passing, each failing for the defect rather than a typo. New file `tests/it/issue_1618_anchored_repeat_semantics.rs`:

- denoted-**bases** assertions via `compare_denoted_sequences`, not string pins, so they keep their meaning if the canonical spelling is ever revised
- the exact canonical string, which distinguishes carrying the absorbed copy from preserving bases by refusing to re-phase at all
- the inverted convergence test described above
- the `repeated.md:97` CFTR shape, including the pure-re-phase count guard
- the **seeded overlapping** re-phase above, including the three counts whose emitted edit kind changes, the exact canonical strings on both sides of that boundary, denoted-bases agreement over counts 1..8 in both shuffle directions, and idempotency across the `del`/`dup` outputs

Full suite: **10,177 run, 10,176 passed**, run with `FERRO_MANIFEST` set and both spec artifacts regenerated. The single failure was a deliberately-unshipped assertion for a sibling defect, now filed as #1675 and removed from this branch. Re-run after rebasing onto `d5f26fcb` and adding the seeded-overlap tests, with both spec artifacts regenerated and `FERRO_ASSERT_IDEMPOTENT`/`FERRO_ASSERT_REPARSE`/`FERRO_ASSERT_IN_BOUNDS` set: **9,925 run, 9,925 passed** (proptest, the two sweep suites and the six oracle suites out of the selection). `cargo fmt --all` and `cargo clippy --all-features --all-targets -- -D warnings` clean.

## Split out deliberately

A contracting repeat is rewritten to `del` (`g.258GT[14]` → `g.286_287del`) although `RNA/repeated.md:33` prefers the repeat form. That is **#1675**, not this PR: its only citation is RNA-axis while the reproducer is genomic, so applying it to `g.`/`c.` is a cross-axis generalization needing a ruling — the same class of move #1664 was about. It also loses no bases, so it is a rule 2 question rather than a rule 3 one.

Representation-Change: 13 shapes move, all on an anchored single-position repeat
  whose spelled unit tiles a different tract than the rotation the tract
  maximization picks. Measured against origin/main (d5f26fcb) on the branch's own
  synthetic references (core base 1 = g.257), 3'-shuffle, in three classes.
  A. SEEDED OVERLAP -- and here THE EMITTED EDIT KIND CHANGES, not only the count.
  The spelled unit's tract SPANS the anchor, so it seeds the maximization, and a
  strictly longer overlapping rotation displaces it; the count is then re-based by
  the copies the wider window swallows. Because a repeat whose count sits one unit
  either side of the reference tract's count is rendered as `del` or `dup` rather
  than as a repeat, shifting the count by one also shifts WHICH count renders as
  WHICH edit. On core ATGTGTGTGA anchored at g.259 with unit GT -- a 3-copy GT
  tract sitting inside a 4-copy TG tract -- every count 1..8 moves and four of them
  change kind:
    g.259GT[1]  g.258_265TG[1]  -> g.258_265TG[2]
    g.259GT[2]  g.258_265TG[2]  -> g.264_265del    repeat -> del
    g.259GT[3]  g.264_265del    -> g.259GT[3]      del    -> repeat, left alone
    g.259GT[4]  g.259GT[4]      -> g.264_265dup    repeat -> dup
    g.259GT[5]  g.264_265dup    -> g.258_265TG[6]  dup    -> repeat
    g.259GT[6]  g.258_265TG[6]  -> g.258_265TG[7]
    g.259GT[7]  g.258_265TG[7]  -> g.258_265TG[8]
    g.259GT[8]  g.258_265TG[8]  -> g.258_265TG[9]
  So a consumer holding a stored `repeat` for one of these inputs can read back a
  `del` or a `dup`, and one holding a stored `del` or `dup` can read back a repeat.
  This class is NOT merely a count change, and "the count now carries the copies
  the wider window swallows" does not predict it. The family is unbounded in the
  count; 1..8 is what is measured and asserted.
  B. OVERLAPPING but NOT seeded -- the spelled unit's tract ends AT the anchor, so
  the span filter rejects it as a seed. The count carries the copies the wider
  window swallows and the edit kind is unchanged --
    AAGTGTTA      g.262TG[6]  g.259_262GT[6] -> g.259_262GT[7]
  C. DISJOINT tracts -- the re-phase is declined and the count stands against the
  tract it was written for, because relocating the window denotes different bases
  whatever the count --
    AATGTGTGGTAA  g.265TG[6]  g.265_266GT[6] -> g.259_264TG[6]
    AATGTGTGGTAA  g.265TG[2]  g.265_266dup   -> g.263_264del
    AATGTGTGGTAA  g.265TG[1]  g.265TG[1]     -> g.259_264TG[1]
    GGACCAGG      g.261AC[5]  g.261_262CA[5] -> g.259_260AC[5]
  Every new output denotes its input's bases, checked with
  `spdi::compare_denoted_sequences` through `hgvs_to_spdi`, not with the
  normalizer. Eleven of the thirteen previously denoted a DIFFERENT sequence rather
  than merely a different spelling -- 7 of the 8 in A and 4 of the 5 in B and C
  together (an earlier revision of this trailer said 3 of 5; re-measured, it is 4,
  because GGACCAGG g.261AC[5] shipped a same-LENGTH, different-BASES output) -- so
  a stored string of those shapes changes and the new one is the correct one.
  Unmoved and asserted: a pure re-phase (`repeated.md:97`'s CFTR `g.258TG[13]`), an
  expansion (`g.258GT[17]`), and the arm where the spelled unit tiles nowhere at
  the anchor (`g.259TG[6]`), which is pinned as a limit rather than fixed --
  `hgvs_to_spdi` reads that input as denoting no sequence, so there is no baseline
  to conserve. Not measured against a real corpus:
  `examples/dump_normalized_corpus.rs` cannot build a repeat whose spelled unit is
  out of phase with a different tract, so a zero from it would be structural rather
  than evidence.

Closes #1618

### DIFF
--- a/src/normalize/rules.rs
+++ b/src/normalize/rules.rs
@@ -2097,6 +2097,59 @@ fn five_prime_align_tract(
 
 /// Normalize a repeat variant
 ///
+/// Re-base an anchored repeat's count from the tract the caller spelled onto
+/// the tract the rotation search named — or decline the re-phase.
+///
+/// `[N]` counts copies of the unit the caller **spelled**, against that unit's
+/// own maximal tract; the emitted description counts against the winning
+/// rotation's tract, so the difference has to be carried or reference copies
+/// are silently created or deleted (#1618). `N + winner - literal` leaves the
+/// net base change unchanged, and a pure re-phase (`DNA/repeated.md:97`'s 11
+/// `TG` -> 11 `GT`) has the two counts equal and so is untouched.
+///
+/// `None` means **decline** — the honest answer for two shapes, neither of
+/// which any count on the winner's tract can express:
+///
+/// * **The tracts are disjoint.** The description's window has moved to a
+///   different stretch of reference, so re-basing conserves the net length and
+///   nothing else. On `AATGTGTGGTAA`, `g.265TG[6]` re-based onto the 1-copy
+///   `GT` tract is `g.265_266GT[4]`, which denotes `TGTGTGGTGTGTGT` where the
+///   input denotes `TGTGTGTGTGTGGT` — the same 14 bases in a different order,
+///   as `spdi::compare_denoted_sequences` reports.
+/// * **The re-based count would underflow.** `g.265TG[1]` on that same
+///   reference is `1 + 1 - 3`. The `saturating_sub` this replaces clamped it to
+///   `0` and routed to a two-base deletion, where the input denotes **four**
+///   bases fewer — a silent loss of two bases, in the arm the re-basing rule
+///   itself added. (The saturation's original comment justified it by the
+///   disjoint case, where `literal == winner` and nothing ever saturates.)
+///
+/// The caller answers a decline by keeping the spelled unit's own tract, which
+/// is what `[N]` was written against, so the output denotes exactly the input's
+/// bases.
+///
+/// A `None` literal tract takes **no** adjustment: the spelled unit tiles
+/// nowhere at or 5' of the anchor, so there are no copies to re-base against,
+/// and `hgvs_to_spdi` reads such an input as denoting no sequence at all — no
+/// baseline exists to conserve. `the_none_literal_arm_takes_no_adjustment` pins
+/// the asymmetry that leaves, as a limit rather than as a fix.
+fn rebase_count_onto_tract(
+    specified_count: u64,
+    literal_tract: Option<(u64, usize, usize)>,
+    winner: (u64, usize, usize),
+) -> Option<u64> {
+    let Some((literal_count, literal_start, literal_end)) = literal_tract else {
+        return Some(specified_count);
+    };
+    let (winner_count, winner_start, winner_end) = winner;
+    // Half-open spans, so touching-but-not-overlapping is disjoint.
+    if literal_end <= winner_start || winner_end <= literal_start {
+        return None;
+    }
+    specified_count
+        .checked_add(winner_count)?
+        .checked_sub(literal_count)
+}
+
 /// Given a repeat notation like CAT[1], determines the appropriate
 /// normalized representation by comparing to the reference.
 ///
@@ -2175,19 +2228,19 @@ pub fn normalize_repeat(
     // behavior change is preferring a genuinely larger tract.
     let mut working_unit = canonical_unit.to_vec();
     let literal_tract = count_tandem_repeats(ref_seq, pos, &working_unit);
-    // Reference copies the maximization swallows when it displaces the literal
-    // tract with a longer one. `[N]` counts copies of the unit the caller
-    // SPELLED, against that unit's own tract; re-phasing to a rotation that
-    // tiles a wider tract changes what the count counts, so the copies now
-    // inside the window but outside the literal tract have to be carried or
-    // they are silently deleted (#1618).
+    // Copies of the caller's own unit at the anchor, kept so the count can be
+    // re-based onto whatever tract the maximization below actually names.
+    //
+    // `[N]` counts copies of the unit the caller SPELLED, against that unit's
+    // own tract; re-phasing to a rotation that tiles a different tract changes
+    // what the count counts, so the difference has to be carried or reference
+    // copies are silently created or deleted (#1618).
     //
     // `DNA/repeated.md:97-99` is the spec's own worked case: re-spelling 11 `TG`
     // copies as `GT` slides the tract one base 3', swallowing the first `T` of
     // the neighbouring run, and the spec decrements that run `T[7]` -> `T[6]` so
-    // the total is conserved. Same principle, applied to a widening rather than
-    // a slide.
-    let mut maximization_absorbed: u64 = 0;
+    // the total is conserved. Same principle, applied to a re-phase that changes
+    // the tract's size rather than to a pure slide.
     let tract = match literal_tract {
         _ if end_pos == pos && working_unit.len() >= 2 => {
             // Hold the literal seed to the SAME span test the rotations below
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

PR #1680, opened 2026-08-11
docs(rulings): scope the delins merge ruling by direction, close the ins carve-out

nh13:
Three ledger amendments recording decisions that were already made, plus one correction. No code path is touched.

## 1. `delins-merge-vs-individual-gap-two-or-more` — the direction scope, stated

The record scopes itself to "a MINIMAL single `delins` that would be split because payload bases coincide with reference bases". That is a statement about **coincidence** and says nothing about **direction**. The implementation is directional and always has been: `merge::coalesce_payload_alignment_split` evaluates `payload.len() < span.len()` **first and independently** of the embedding test, so every net-insertion block is excluded before embedding is ever consulted. The code's own doc comment carries a rationale the record did not ("for a net insertion the split form is the canonical one, not the span form"), so the code was opinionated exactly where the record was silent — and the next reader comparing them would have "corrected" the code on the strength of a discrepancy rather than of a decision.

**Ruled: keep the net-insertion exclusion.** The ruling reaches the net-deletion case (payload shorter than the span it replaces) and does not reach net insertions. Three grounds are recorded:

- **SVD-WG010 was rejected** (`SVD-WG010.md:5`, "Status: rejected"), and its own `:16` example is precisely a net-insertion merge — `g.3_4delinsGGT` preferred over `g.[3C>G;7dup]`, a 2-nt span for a 3-nt payload. Per this repo's standing doctrine a rejected proposal earns a negative guard, never an expectation: the committee was offered this exact merge and declined it.
- **`DNA/duplication.md:90-92` publishes a net insertion as a split** — `c.[675-542_1211-703dup;1211-703_1211-702insGTAAA]`. Its weight is stated honestly in the record: the alternative `:92` rejects is `dupins`, "a format not used in HGVS nomenclature", **not** a spanning `delins`, so it does not rank merge against split. Corroboration, not authority.
- **Evidence-base asymmetry.** The ruling's entire evidence base — 208 genomic-axis rows, the 121-row provenance arbitration, the non-uniqueness enumeration, and `:44-47` itself — is net-deletion. Extending to net insertions would retroactively declare thousands of rows conformant on evidence that never examined them, the same over-reach the record's own "WHY THE SCOPE IS EXPLICIT" paragraph refuses on the separation axis.

**Correction, same record.** A claim has been circulating that the spec publishes no unequal-length, non-embedding worked example. It does, and it is `:44-47` itself, derived rather than asserted: span `c.850_901` is 52 nt, payload `TTCCTCGATGCCTG` is 14 nt, the payload is **not** a contiguous substring of the span and not even a pure ordered subsequence of it (which is why `COALESCE_MISMATCH_BUDGET` is 1 rather than 0), and `:46`'s split deletes 20 + 8 + 11 = 39 of the 52 positions, retaining 13 as a gapped subsequence — `TTCC`, `TCGAT`, `GCCT` — plus one inserted `G`, which reconstitutes the payload exactly. So the spec's split contains a pure `ins` member. The real gap is narrower and is now stated as such: what the spec has no worked example of is a **net insertion**. Searched across the committed tree — the claim appears in no committed file, so the correction is recorded in the record rather than applied to one.

## 2. `delins-adjacent-members-when-both-consume-reference` — the `ins` half of the carve-out is closed

That record's scope paragraph left `dup` and `ins` open together. They are not one question. The `ins` half is closed on the `c.2077delinsATA` triple, which states the same answer in three registers — `DNA/delins.md:86-89`, `DNA/substitution.md:93-96`, and `RNA/delins.md:68-71` (the RNA twin reads `NM_007294.3:r.2077delinsaua`). The competing split is rendered `<code class="invalid">` inside the question, and `delins.md:89` records that the passage previously **permitting** the split spelling was *removed* by the committee — the spec retracted its own prior permission, which is stronger than silence.

**This ratifies shipped behaviour and moves zero rows.** Verified against the prepared reference before the record was written: `NM_007294.3:c.2077delinsATA` is a fixed point, `NM_007294.3:c.[2077G>A;2077_2078insTA]` normalizes to it (disclosing `MEMBERS_COALESCED_FROM_REPORTED_FORM`), and the RNA twin is a fixed point — identically on all four partition arms (`FERRO_PARTITION` `live`, `shadow`, `canonical`, `canonical-coalesced`).

The `dup` half stays open, on its own ground: `duplication.md:18` requires a variant describable as a duplication to be described as one, and merging destroys it. No clause stands behind an `ins` the way `:18` stands behind a `dup`. The repeat-expansion third shape is also untouched.

## 3. `canonical-form-choice-when-both-legal` — a fourth piece of counter-evidence

`protein/delins.md:50`: "Although the proteins resulting from the changes `NM_080877.2:c.1733_1735delinsTTT` and `c.1732_1794del` are identical, their HGVS description is different." It is the sharpest of the four because it is a **sequence-identity** case rather than a provenance case: the resulting sequences are stated to be the same and the spec still keeps two descriptions apart, which is the one thing derivation from the resulting sequence structurally cannot do. It does not overturn the ruling; it is recorded so it is not later discovered as a refutation.

## Verification

- `cargo run --features dev --bin generate_spec_fixture` and `generate_spec_enumeration` both succeed. Plain generation is the guard: it checks every citation quote byte-for-byte against the spec checkout, so the 11 new citations are verified against `assets/hgvs-nomenclature` at `6f85311`.
- `cargo nextest run --features dev` — all 10166 tests pass. `ruling_records_are_intact`, `the_rendered_index_is_current`, `clauses_named_by_several_records_are_flagged` and `claude_md_adjudication_tables::the_two_tables_partition_the_ledger_by_status` are all green.
- `cargo fmt --all` clean; `cargo clippy --all-features --all-targets -- -D warnings` clean.
- No record changed status, so no `RULING_STATUSES` pin moves and neither `CLAUDE.md` adjudication table gains or loses a row; their summary prose is updated to match the amended rulings. `tests/it/clause_ruling_index.rs`'s generated block was captured from `render_index()` programmatically and spliced on its `BEGIN`/`END` markers — nothing retyped — and its multiply-named-clause pin moves 13 → 14 for the new `SVD-WG010.md:5` line.

Representation-Change: none. Records and documentation only — the ledger, the two pinned summaries that mirror it, and the regenerated clause index. No file under `src/` is touched, and every ruling here documents behaviour ferro already ships (Task 2 was measured at zero moved rows across all four partition arms before being written).


nh13:
@coderabbitai pause

PR #1682, opened 2026-08-11
fix(normalize): scope the delins payload-coincidence carve-out to coding DNA

nh13:
`DNA/delins.md:47` recommends the single `delins` where a split exists only because payload bases coincide with reference bases. The pass implementing it, `coalesce_payload_alignment_split`, was **axis-blind** — it merged on `g.`/`m.`/`o.`/`n.` as readily as on `c.` This scopes it to the **coding DNA axis** (`c.`) and nothing else, per an operator ruling recorded in the ledger in this same change.

## The ruling

New record `delins-payload-coincidence-carve-out-is-coding-dna-scoped` (`decided`, governing `DNA/delins.md:47`). It answers a question its sibling `delins-merge-vs-individual-gap-two-or-more` deliberately did not reach: that record scoped its ruling to the payload-coincidence *shape* and said rows on an axis with no reading frame "remain violations", without saying whether the carve-out could still reach them. It cannot.

Grounds, all quote-verified against the spec checkout by the generator:

- `:47` states its own reason — the merge "prevents software tools making incorrect predictions for the consequences on protein level". Off a translated axis there is no protein consequence to mispredict, so the clause's stated purpose is absent rather than merely weaker.
- The worked example is a coding row, `LRG_199t1:c.850_901`, and the neighbouring exception at `:18` is explicitly codon-bound ("together affecting one amino acid").
- The spec scopes rules this way elsewhere, and finer than the axis: `repeated.md:23` restricts its own rule to "the coding sequence, which does not include the introns or the UTR sequence".
- Inverse precedent: `general.md:43` says the 3' rule applies "to ALL descriptions (genome, gene, transcript, and protein)". The spec says so explicitly when it means universal, and `:47` does not.
- It is the only reading under which **both** decided records hold at once — `delins-merge-vs-individual-gap-two-or-more`'s worked example keeps its single `delins`, and `separation-is-a-property-of-the-spelling-not-of-the-variant`'s `g.` case reaches its decided form A.

### RNA is not ruled on here, and there is no clause that could carry it

This record rules on the coding DNA axis alone. It makes **no** ruling about RNA, and nothing in it may be read as licensing or forbidding the merge on an RNA description: every clause it cites is a DNA recommendation or a molecule-agnostic one, and a clause's jurisdiction is the molecule directory it sits in.

That is recorded rather than left implicit, because the next reader will go looking for the RNA half. `RNA/delins.md` states **no counterpart to `:47`** anywhere in its 71 lines — no "align" note, no "recommended", no simplicity or protein-level rationale. So there is no RNA clause that could carry this carve-out, and none is cited to pretend otherwise. An RNA ruling would need RNA authority and a separate operator decision. (Compare `exon-junction-dup-converge-from-the-far-side`, which reaches RNA legitimately because `general.md:44` names the axes in its own text. This record has no equivalent.)

This is a behaviour claim, not wording: `AxisFrame::reading_frame` is **true** for `CisKind::Rna` whenever the transcript resolves a `cds_start`, so gating on it silently extended a DNA-only carve-out onto an axis the ruling does not reach. The gate therefore tests the axis **kind**. Reproduced on a purpose-built probe — keyed on the frame flag, a two-member RNA deletion pair merged into one spanning delins; keyed on the axis kind it stays split. Pinned by `a_coding_rna_row_is_outside_the_dna_carve_out`.

**Narrowing off the RNA axis moves no measured row, and that zero is structural**, not reassuring. The two bulk corpora hold 2 RNA-axis rows between 935,239: one a lone `del` that never reaches a multi-piece coalesce, one refused upstream as coincident edits. The designed cis corpus generates genomic and coding rows only. Its RNA variant (5,636 classes) draws against `NR_TEST.1`, a **non-coding** accession, so it cannot exercise the coding-transcript RNA path at all. The behaviour is reachable and was demonstrated on a probe rather than observed in a corpus — so the Representation-Change figures below are unchanged by this correction (verified: 0 of 417,227 ClinVar rows, 0 of 368,234 Paraphase rows, and 0 of 47,392 designed-corpus spellings move).

The record carries its counterweight rather than arguing one side: `:47` gives two grounds and the other one, "it is simpler", is axis-neutral and is given no weight here; and `:17`, the clause the carve-out overrides, carries no axis restriction of its own.

## What it costs, stated in the record

`NG_009874.2:g.158330_159052delins…` — a stored 723-base single `delins` in the ClinVar corpus — is re-derived into a **193-member** cis allele, and stays that way *precisely because* the gate declines on a `g.` row where the ungated pass merged it back. That is the worst consumer-facing shape in the measurement behind this change, and this ruling preserves it.

## Also: an arm-scoping trap, fixed

The two repair passes scoped to the canonical partitioner spelled their scope as a **positive** arm list, `matches!(rule, Canonical | CanonicalCoalesced)`. An arm missing from that list takes the `else` branch, which is `live` behaviour. This was hit for real while measuring: a canonical-cutting arm absent from the list at `coalesce_compensating_gap_split` scored as though it were `live` on that pass alone, moving one stratum's count with nothing failing. Replaced with `PartitionRule::cuts_with_canonical`, an exhaustive `match` — adding an arm is now a compile error there. Pinned by `every_arm_declares_whether_it_cuts_canonically`.

Writing it as `!= Live` would not have been the fix: that also selects `Shadow`, whose pieces come from `partition_block_sequence_first` and which those passes argue against.

## A CDS-region refinement was built and rejected

`repeated.md:23` scopes finer than the axis, which invites an extra precondition excluding `c.-n`, `c.*n` and intronic `c.n±m`. It was built and instrumented: over the 11,272-class designed cis corpus it was **reached 60,194 times and rejected nothing**, and changed no row of the 785,461 normalized from ClinVar and Paraphase.

That zero is **structural, not reassurance** — `collect_canonical_edits` already refuses any member outside the axis's positive body, so such a span never reaches the pass. It is not shipped, and the reason is recorded at the gate so the next reader does not re-add it believing it does something. Relaxing that upstream exclusion is the precondition for the refinement to become measurable.

## Measurement

Arms compared at `e1e3583e` over the 11,272-class designed cis corpus (`examples/generate_cis_confluence_corpus.rs`, both directions) and 935,239 real rows (ClinVar 500k + Paraphase exhaustive, 785,461 normalized ok).

| | ungated `canonical-coalesced` | **gated (this PR)** |
|---|---:|---:|
| payload-coincidence, `g.` axis: single-member outputs | 521 | **141** |
| …of which are genuine merges | 380 | **0** |
| payload-coincidence, `c.` axis: still split | 47 | 47 |
| confluence failures (3′ / 5′) | 1 / 0 | 1 / 0 |
| sequence changed, out-of-bounds, non-idempotent, reparse failures | 0 | 0 |

All 380 removed merges were genuine: an exact minimum-deleted-runs DP over every order-preserving embedding confirms the surviving 141 have a one-member minimal description, so nothing correct was lost. The 47 that still split on the coding axis are **all** `COALESCE_MAX_SEPARATION` (gaps of 9–13), not this gate.

Seam oracles are `#[cfg(debug_assertions)]`, so they were re-proved live on this build rather than carried over: `FERRO_ASSERT_IDEMPOTENT` fires on a known pre-existing case, and the sequence oracle reports `compared=6 skipped=0` rather than skipping. Tier 1 then runs clean across both corpora under the gated arm.

## Verification

`cargo nextest run --features dev` with `FERRO_MANIFEST` set → **10174 tests run: 10174 passed, 153 skipped**. `cargo fmt --all --check` and `cargo clippy --all-features --all-targets -- -D warnings` clean. Both spec generators regenerate without error, which is what verifies the record's quotes byte-for-byte.

## Not in scope

This does **not** flip the default. Making the gated arm the shipped rule is #148, a separate decision with its own ~1,493-row representation change and ordering constraints.

Representation-Change: The shipped default is unaffected — `FERRO_PARTITION` unset is `live`, which never runs this pass; measured at **0 rows moved** over 785,461 normalized ClinVar and Paraphase rows. What moves is the opt-in `FERRO_PARTITION=canonical-coalesced` arm, which stops merging across unchanged bases on axes with no reading frame: **340 real rows** (136 of 417,227 ClinVar; 204 of 368,234 Paraphase) and **380 classes** of the 11,272-class designed cis corpus. Every moved row goes from a merged spanning `delins` to the individual members `general.md:34` requires. No row changes the bases it denotes (asserted by `FERRO_ASSERT_SEQUENCE` across both corpora).


nh13:
@coderabbitai pause

PR #1684, opened 2026-08-11
fix(hgvs): refuse an alignment-only symbol, at the stage the ruling names

nh13:
`background/standards.md`'s DNA symbol table daggers two of its rows — `X` ("masked nucleotide", `:36`) and `-` ("gap of indeterminate length", `:37`) — and footnotes the dagger at `:39`: *"used in alignment only."* The decided `rulings[alignment-only-symbol-in-a-description]` rules that neither may appear in a description and ferro must refuse both. Ferro accepted `X` in all three error modes and re-emitted it with an **empty** warning vector. Normalization was not impossible, it was *vacuous*, which is worse: the output looked normalized while carrying a spelling the recommendations prohibit.

## The stage is mode-dependent, and that is why this is two changes and not one

`rulings[absolute-prohibition-enforcement-stage]` merged as #1634 and unblocked this. It puts the check in two places:

| mode | parse | normalize |
|---|---|---|
| **strict** | **fails**, `W3028` | (unreached) |
| **lenient** | accepts, emits `W3028` | **fails** |
| **silent** | accepts, quiet | **fails** |

- **Strict fails at parse** — `apply_alignment_only_symbol_rule` in `parse_hgvs_with_config`, keyed on `ErrorConfig::action_for`, the same seam `apply_bracket_cardinality_rule` (#493) established. Strict validates input conformance, not merely parseability.
- **Lenient and silent fail at normalize** — in `normalize_core_checked`, and **unconditionally**. The mode gate governs whether the *input* is judged; it cannot govern whether the *output* conforms, which is rule 1 of the README ruleset. So lenient fails on the ruling's own ground — it cannot normalize, because a masked base names no nucleotide — rather than on an input-conformance check it does not run.

**The grammar is untouched.** A refusal in `parse_inserted_sequence`'s alphabet would refuse in every mode and contradict the ruling. That is the fix an earlier characterization pass correctly declined to make while the stage was still undecided; `tests/it/issue_1627_named_element_alphabet_reach.rs` comes forward from it and now serves as the guard that the grammar stays unchanged.

## Re-measured population — every lone-`X` count was an undercount

`X` reaches the AST through exactly one carrier, `InsertedSequence::Named`, but through **two** arms of `parse_inserted_sequence`. The `is_iupac_base` arm walks the whole run and sets `has_non_iupac` on any uppercase non-IUPAC byte, so one stray symbol reclassifies an otherwise-literal run *wholesale*:

```
g.10delinsX      -> Named("X")        g.10delinsACGTX  -> Named("ACGTX")
g.10delinsXACGT  -> Named("XACGT")    g.10delinsACXGT  -> Named("ACXGT")
```

All four were accepted and re-emitted in all three modes before this change.

| population | count | |
|---|---|---|
| corpus rows citing `standards.md:39` | **24** | |
| …of which state a **lone** `X` | **24** | all 8 distinct spellings are `delinsX` |
| …of which state an **embedded** `X` | **0** | :warning: **STRUCTURAL ZERO — see below** |
| corpus spellings containing `-` | 44 | none of them a sequence symbol |
| …of which state `-` as a **sequence symbol** | **0** | :warning: **STRUCTURAL ZERO — see below** |

### Both zeros are structural. Neither means "the shape does not occur."

**Embedded `X` is 0 because the corpus generator cannot emit the shape**, not because the shape is safe. The generator writes the insert as a fixed `delinsX` and never varies its literal run, so no row it can produce carries an embedded symbol. Measured **directly**, outside the corpus, on `78c43230`:

| input | parses to | strict | lenient | silent |
|---|---|---|---|---|
| `g.10delinsACGTX` | `Named("ACGTX")` | accepted, re-emitted | accepted, re-emitted | accepted, re-emitted |
| `g.10delinsXACGT` | `Named("XACGT")` | accepted, re-emitted | accepted, re-emitted | accepted, re-emitted |
| `g.10delinsACXGT` | `Named("ACXGT")` | accepted, re-emitted | accepted, re-emitted | accepted, re-emitted |
| `g.10_11insACGTX` | `Named("ACGTX")` | accepted, re-emitted | accepted, re-emitted | accepted, re-emitted |

So the hole was **wider than any recorded count**, and reading the corpus's `0` as reassurance would have been exactly backwards. The check therefore tests for the **presence** of a daggered symbol anywhere in the name rather than for a name that *is* one, and the embedded shapes are asserted **by construction** in the new tests rather than harvested from a generator that cannot produce them.

**`-` is 0 reachable shapes because of a grammar accident, not an enforced rule.** It matches **no arm** of `parse_inserted_sequence` — not `[`, not `(`, not digit/`*`, not IUPAC, not uppercase — so it is left unconsumed and the variant-level trailing-character check refuses it. Measured in every position, in all three modes: `delins-`, `delins-ACGT`, `delinsACGT-`, `delinsAC-GT` and `ins-` are all refused. That is why `-` needed no rule of its own, and why it is pinned as a standing fact rather than copied as the template for `X`.

**And the 44 corpus spellings containing `-` are position syntax, not the symbol table** — 5'UTR markers (`c.-3delinsX`, `c.-3del3`, `c.-3insT`) and `checklist.md:45` hyphen ranges (`g.266-268del`). That is precisely why the predicate is AST-keyed: **a whole-string scan for `-` would have flagged all 44**, none of which states an alignment-only symbol.

## Scope, and what is deliberately left alone

The predicate is **AST-keyed on the inserted sequence**, never a scan of the rendered description. Two false positives that buys, one of them found the hard way:

- `XM_`/`XR_`/`XP_` RefSeq accessions carry an `X` unrelated to the symbol table.
- `ins[244-8_249]`'s `-` is an intronic offset. Not hypothetical — an earlier revision rendered each `InsertedPart` and scanned the text, and `ins_intronic_offset_range_returns_unsupported_error` caught it. The `InsertedSequence` match is now exhaustive so a new variant must be classified rather than defaulted.

Untouched, each for a stated reason:

- **Genuine mobile-element names** (`AluYb8`, `LINE1`, `L1`, `Alu`) carry no daggered symbol. Whether a named insert is conformant at all is a separate question — `DNA/complex.md:169` answers *"No, not really, it is not exact"* about `insL1.603bp` — and **no ruling settles it**.
- **A lone `Z`** carries no daggered symbol either, and no ruling covers it. `a_lone_non_daggered_uppercase_letter_is_still_accepted` becomes a scope guard that fails if this is later widened into an unadjudicated alphabet rule — a widening that would reach `AluYb8` and `LINE1`.
- **Lowercase `x`** is indistinguishable from an ordinary name character (`Alu` has three lowercase letters) — **on a DNA axis**. On `r.` that argument inverts and the gap is real: `general.md:48` is the DNA-level bullet, and its RNA sibling `general.md:50` says "nucleotides in **lower case** using [IUPAC-IUBMB assigned nucleotide symbols]", so lowercase *is* the RNA alphabet there. Measured: `r.10delinsacgu` is `Literal([A,C,G,U])`, `r.10delinsacgux` is `Named("acgux")` — the same wholesale reclassification, one axis over — and it is accepted and re-emitted in all three modes, exactly as `g.10delinsACGTX` was before this change. It is **not** fixed here because a refusal would rest on `general.md:50`'s alphabet clause rather than on `standards.md:39`'s DNA symbol table, and **no ruling covers that** — not `rulings[alignment-only-symbol-in-a-description]`, which is written against the DNA table, nor anything under `RNA/`. Pinned as an open question by `a_lowercase_rna_masked_nucleotide_is_still_accepted` so it cannot rot into a silent gap. The converse over-reach is deliberate and harmless: uppercase `r.10delinsX` *is* refused, and both readings agree it should be.
- **Protein `X`** is the parser-native `Xaa`, and the deprecated stop-codon spelling is W3008's business.

### Not in scope: the `ins6` half of #1627

#1627's body names **48** rows. This PR ships **24** of them — `standards.md:39`. The other **24** are `checklist.md:33` (`c.5439_5430ins6`, *"is not allowed, the inserted sequence … should be specified"*) and are **not addressed here**.

They are a different clause with a repair question this one does not have. Three spellings must be separated that the AST distinguishes only as `Count` / `Range` / `Repeat`:

| spelling | AST | status |
|---|---|---|
| `ins6` | `Count(6)` | prohibited by `checklist.md:33` in as many words |
| `ins(10_20)` | `Range(10, 20)` | a different spelling, which the clause does not name |
| `insN[603]` | `Repeat { base: N, .. }` | **explicitly endorsed** by `DNA/complex.md:175` — *"When the inserted sequence is not known, its (estimated) size can be used, like `…ins[N[603];123+45_123+51]`"* |

Blast radius, measured: **37 occurrences of `ins6` across 4 files** — `corpus_prohibited_inputs.rs` (26), `spec_corpus_regressions.rs` (8), `spec_conformance_axis.rs` (1), `src/conformance/spec_corpus.rs` (2).

**Hence `Refs #1627` and not `Closes`.** Auto-closing the issue on merge would bury the half that remains. Its census residue stays visible and pinned: `prohibited_absolute_accepted` holds at **32**, of which `checklist.md:32`'s `ins6` is 24.

## Census movement — re-blessed deliberately

Two `spec_conformance_axis` counters move, both **down**, identically in both shuffle directions:

| figure | was | now | delta |
|---|---|---|---|
| `prohibition_violating_outputs` (3' and 5') | 32 | **8** | −24 |
| `prohibited_conditional_accepted` (3' and 5') | 40 | **16** | −24 |

**It is one population counted twice, not two.** All 24 rows are `standards.md:39`, and both per-clause decompositions — pinned by `the_absolute_prohibitions_ferro_accepts_are_three_clauses_and_no_others` and `every_prohibition_violating_output_is_a_re_emitted_prohibited_input` — drop exactly that one entry. The residues are the shapes this change does not reach: `checklist.md:20`'s 16 conditional rows (correct as they stand, per `rulings[bare-transcript-intronic-position]`) and `checklist.md:16`'s 8 violating outputs (#1628).

`prohibited_absolute_accepted` stays **32**, and that is the useful half: the corpus grades an `X` row `Strength::Conditional` while the axis counts an `X` *output* as an absolute violation, and the ruling records that the grading is moot either way. The surviving 32 are `checklist.md:32`'s `ins6` (24) plus `checklist.md:16`/`:45` (4 + 4).

Every rank-2, idempotency and sequence-preservation figure is **unchanged** in both directions, which is what a refusal of un-denotable inputs should look like: the refused rows contributed no confluence family and no sequence, so nothing they leave behind can move. Nothing to promote to `spec_corpus_regressions.rs` — no row newly fails. A fourth `RE-BLESSED` section in the module docs records all of this.

**Re-measured after a rebase onto `ce933533`, not carried over.** The branch was 17 commits behind `main`, and `main` had itself re-blessed the rank-2 pins (#1649's two-deletion alignment moved `converged` to 9,402 / 9,228 and all three `split_*` in both directions) and had already taken `RE-BLESSED (3 of 3)`. So "unchanged" now means unchanged against **#1649's** values, not #1536's as this section first said — a baseline claim is only as good as the baseline, and the baseline moved under it. Both censuses were re-run on the rebased tree and land on the pinned figures: 3' `8 / 16 / 32`, 5' `8 / 16 / 32`, with `converged` 9,402 / 9,228 untouched.

**The generated spec artifacts do not move at all.** Regenerated with `generate_spec_fixture` and `generate_spec_enumeration` and diffed against the base: 934 and 2172 rows respectively, **zero added, zero removed, zero status changes** — re-counted after the rebase, and the ruling ledger re-checked programmatically against both the pre-rebase tip and `origin/main` (20 records, none dropped, no status changes). The spec's own examples state no alignment-only symbol.

## Verification

- `cargo nextest run --features dev --test-threads=4` on the rebased tree, with `FERRO_MANIFEST` set so the reference-aware suites run rather than skip → `Summary [555.570s] 10289 tests run: 10289 passed (2 slow), 152 skipped`, exit 0. (The earlier `10177` was measured on the pre-rebase base; the growth is `main`'s 17 commits plus this branch's added guards.)
- `cargo fmt --all --check` clean; `cargo clippy --all-features --all-targets -- -D warnings` clean; `cargo check --features python` clean.
- Under `FERRO_ASSERT_IDEMPOTENT` / `_REPARSE` / `_IN_BOUNDS` the suite reports **7** failures, not the 3 this line claimed before the rebase — the figure was measured on `78c43230` and `main` has moved twice under it since. **All 7 are pre-existing, established by set-difference rather than by inspection**: the same command was run on the PR's base `ce933533` in a detached worktree, and the two failing sets are byte-identical — 0 new, 0 fixed. They are `defect_non_idempotent_outputs::{an_insertion_immediately_after_the_last_cds_base_is_retyped_as_a_delins, norm_cubed_equals_norm_squared_for_every_row_in_the_class, the_chain_never_changes_the_sequence_it_denotes, the_non_idempotent_output_is_always_the_spanning_delins_respelling}`, `spec_corpus_regressions::an_insertion_at_the_cds_end_is_not_a_fixed_point`, and both `spec_conformance_axis` censuses. None of the inputs carries an alignment-only symbol, and all 7 **pass** in the default gate — these assertions are opt-in and are not part of CI's gate.

Refs #1627 — the `standards.md:39` half of it; see **Not in scope** above.

Representation-Change: 24 rows move (3' and 5' alike), all from re-emitted to refused; 0 respell, 0 merge, 0 split. Every affected row's output is invalid today — a description stating `X` denotes no sequence — so nothing that currently produces a legal description changes. The refusal is newly reachable by lenient and silent callers, which previously got a string back; strict callers now fail one stage earlier, at parse. The generated spec fixture and enumeration are byte-identical (934 and 2172 rows, zero status changes).

## Summary by CodeRabbit

- **New Features**
  - Added diagnostics for alignment-only symbols (`X` and `-`) in HGVS descriptions.
  - Parsing now supports configurable handling: silent acceptance, lenient warnings, or strict rejection.
  - Added Python-facing error type `AlignmentOnlySymbolInDescription` with stable code support.

- **Bug Fixes**
  - Normalization now consistently rejects descriptions containing alignment-only symbols and reports warning `W3028`.
  - Improved detection avoids incorrectly flagging accessions and legitimate mobile-element names.

- **Documentation**
  - Documented the new warning behavior and specification audit coverage.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

PR #1685, opened 2026-08-11
test(rulings): refuse a ruling that speaks to a molecule it does not cite

nh13:
A ruling in `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json` can cite a clause that resolves cleanly, quotes correctly, and still is not authority for what the record says — because `assets/hgvs-nomenclature/docs/recommendations/` is organised by **molecule** and the record rules on a different one. `DNA/` covers `g.`/`c.`/`n.`/`m.`, `RNA/` covers `r.`, `protein/` covers `p.`. That directory is the jurisdiction of every clause inside it.

Nothing checked it. The spec-fixture generator checks each quote is really at its `file:line`; `clause_ruling_index.rs` checks every cited clause is reachable from a clause string; `ruling_citation_currency.rs` checks prose does not contradict a record's status. A record can satisfy all three and still rest an RNA ruling on a DNA clause.

Two instances turned up on the same day, which is what argued for a guard rather than a third one-off correction.

### The ledger fix, which is a citation promotion and not a re-adjudication

`delins-codon-carve-out-gap-one` declares `r.142_144delinsugg` and `r.[142c>u;144a>g]` in `applies_to`, and its rationale rests the "not weakenable by edit type" argument on the identical sentence appearing in nine places — naming `RNA/delins.md:18` and `RNA/substitution.md:18` among them. Its structured `clauses` cited neither. So a settled ruling rested its RNA reach on clauses its own prose named and its citation list omitted.

This PR promotes those two clauses into the `clauses` array. **Nothing else about the record changes** — not `status`, not `governing` (still `DNA/delins.md:18`), not `deviates_from`, not `applies_to`, not `equivalence_classes`, not one word of the rationale, and not the outcome for any input. Please do not read it as re-adjudication; it makes the citation list match the argument the record already makes.

Both quotes were extracted programmatically from the spec checkout at pin `6f85311` rather than transcribed, and re-verified on this branch as byte-exact substrings of their cited lines (`quote in "\n".join(lines[a-1:b])`, both `True`). That check is worth stating because the generator's citation guard is a **whitespace-collapsed substring** match over the cited range, not a byte-for-byte one — a green build is not evidence a quote is exact.

`RNA/delins.md:18` is the stronger of the two for these rows: it states the exception with `r.142_144delinsugg` as its own worked example, which is verbatim one of the record's `applies_to` entries.

### The guard, and how its one open design question was decided

`tests/it/ledger_clause_jurisdiction.rs`. A record may not **speak to** a molecule it has no **authority** for.

- *speaks to* = axis prefixes in `applies_to` entries, plus axis tokens inside backticked spans of the `rationale`. The rationale half is not optional: the second instance's sole `applies_to` entry is a `c.` string and its entire RNA problem lives in prose.
- *authority* = a `clauses` entry under that molecule's directory.

The unresolved question was what to do with `general.md`, which is not molecule-specific. It was decided from the flag list rather than from taste, because both obvious answers fail. **Re-measured on this branch's base (`6f22cde9`, a 21-record ledger), against `origin/main`'s ledger** — i.e. before this PR's citation promotion, so the first defect is still present:

| candidate rule | flags | verdict |
|---|---:|---|
| naive — molecule directories cited, nothing else | 8 | 3 of them spurious: records ruling entirely under `checklist.md` / `background/standards.md` that mention a `c.` example in passing |
| `general.md` is universal authority | 0 | toothless — both known defects cite `general.md`, so neither is caught |
| jurisdiction keyed off the **governing** clause's directory | 4 | drops the 3 spurious rows and one more, but still reports two `r.` rows that half 2 clears |
| **chosen** — governing-keyed, plus an agnostic clause carries the jurisdiction its own quote names | **2** | the defect this PR fixes, and one prose disclaimer (below) |

The earlier revision of this table quoted 6 / 0 / 3 / 2 measured on a 16-record ledger. Those numbers are not restated in the module docs any more, and the docs say why: a count measured on a base nobody can rebuild is the failure `clause_ruling_index.rs` documents at length. What the module now carries instead is `the_scan_is_not_vacuous`, which fails if either half stops changing the verdict.

The chosen rule has two halves, and both are load-bearing:

1. **Jurisdiction attaches to the clause a record rules under.** If `governing` is molecule-agnostic — `general.md`, `checklist.md`, `style.md`, `background/`, `consultation/`, `versions/` — the record carries no molecule-scoped authority to exceed and is unconstrained. That is where `general.md` gets to be universal, and it is what drops the three spurious rows. A record with no `governing` clause at all is still constrained if it reaches into a molecule directory anyway, so a record the generator forbids from naming a governing clause is not exempt by construction.
2. **A molecule-agnostic clause carries the jurisdiction its own quoted text enumerates, and no more.** `exon-junction-dup-converge-from-the-far-side` scopes its ruling onto `r.` and cites `general.md:44`, whose quoted text reads "**exception**: deletions/duplications around exon/exon junctions using **c.**, **r.** or **n.** reference sequences" — it names the RNA axis, so it *is* RNA authority and that record is properly cited. `general.md:34`, `:39` and `:43` name no axis, so they extend nothing. Without this half the guard reports a record that is not wrong.

Both carve-outs are pinned as non-vacuous by `the_scan_is_not_vacuous`: if the ledger ever stops exercising one, the test says to delete the carve-out rather than leave it as unexercised nuance.

### One decided record names an axis only to disclaim it

`projection-codon-exception-is-decided-by-the-rendered-axis` landed on `main` in #1672 after this branch was written, and the guard flags it: its `SCOPE, STATED NARROWLY` paragraph reads "The `p.` axis is the protein consequence itself." That is the record's only `p.`, and it is a statement that the protein axis is **out of scope** — the opposite of a ruling on it. The same paragraph disclaims `r.` and `n.` in the same breath; `r.` needs no handling only because `general.md:44`'s quote names that axis and the record cites it.

Prose cannot be parsed for intent, so the guard cannot tell a claim from a disclaimer. Two ways out were considered and one was rejected:

- **Rejected — weaken the tokenizer** so a bare `` `p.` `` label does not count and only a full description (`` `r.850_901delinsuuccucg` ``) does. That silently drops the bare-label form across the whole ledger, and the bare-label form is exactly how the known defect class recurs.
- **Chosen — name the exception.** `SCOPE_DISCLAIMERS` carries one row: record id, molecule, and the quoted sentence that does the disclaiming. `every_scope_disclaimer_is_still_needed` makes the list **shrink-only** in the same sense as `LEDGER_EXEMPT` — a row whose record has been deleted, re-cited or reworded fails the test and must be removed, so it cannot grow into a mute button.

Adding a citation to that record instead would have been worse: it is an operator ruling that explicitly declines to rule on `p.`, and manufacturing `protein/` authority for it to satisfy a test would assert reach the record disclaims. No verdict, citation, `applies_to` or word of that record is touched.

### What the guard deliberately does not claim

- **Not that a cited clause supports the ruling.** It compares directories, not arguments. Whether the clause says what the record needs is review's job.
- **Not that an axis token is a claim rather than a disclaimer.** See above; the one instance is named.
- **Not that it sees every axis a record touches.** An axis discussed in unbackticked prose, or implied ("on a coding transcript"), is invisible. It is a floor: it catches a record that *writes* an axis token it has no directory for.
- **Not that a quote is byte-exact**, for the reason given above.
- **Nothing about `protein/` from the ledger's citations.** No record cites a clause under it, so that arm is exercised only by synthetic cases in the file. Stated plainly, since a rule with an unexercised arm reads as coverage.

### The tokenizer reads `e.g.` as the `g.` axis — closed, latent, moves nothing

`axes_in` disqualified an axis letter preceded by an alphanumeric or `_`, which keeps spec filenames out (`inversion.md` ends in `n` before a dot). It did not disqualify a preceding `.`, so `e.g.` read as a claim about the `g.` axis. The ledger holds six such occurrences: five in citation quotes, one in a rationale outside backticks.

The direction matters. `molecules_with_authority` reads the quote of a **molecule-agnostic** clause, so an `e.g.` in a `general.md` or `checklist.md` quote hands the record DNA authority it was never granted and quietly stops the guard catching a DNA overreach — a weakening, not noise. **No agnostic quote carries one today**; all five sit on clauses under a molecule directory, whose quotes are never read. So this moves no verdict and is a latent hazard closed rather than a defect fixed. `filenames_are_not_read_as_axes` pins both classes and the positives (`c./r.`, `LRG_199t1:c.3921dup`, a leading `r.`).

### The vacuity pin missed the population the guard actually judges

`the_scan_is_not_vacuous` pinned the ledger size, the records read as speaking, the molecules reached, and each carve-out being exercised. It did **not** pin that any record is both molecule-scoped *and* speaking — the set the guard iterates. Measured: an `is_molecule_scoped` that answers `false` for everything made `no_record_speaks_to_a_molecule_it_cannot_cite` pass over an empty set **and left `the_scan_is_not_vacuous` green**, because such a break satisfies the half-1 pin harder. That assertion is added; the same mutation now fails.

Mutation results, all run against the ledger rather than synthetics (`-E 'test(ledger_clause_jurisdiction)'`, 14 tests):

| mutation | tests failed |
|---|---:|
| `overreaches` returns empty (never flags) | 6 |
| every spoken molecule flagged (always flags) | 5 |
| `is_molecule_scoped` always `false` | 7 (was 0 before the new assertion) |
| molecule-agnostic clause = universal authority | 2 |
| rationale half removed | 4 |

### Other changes

- `tests/it/common/rulings.rs` grows `quote`, `applies_to` and `governing` on its record model, so "what a record is" stays defined in one place. `quote` is **required** — the generator already demands it, and reading an absent quote as empty would silently narrow the authority a clause supplies. A present-but-**blank** quote is rejected too, for the same reason one step further in: `""` satisfies a required key, checks against no line of the spec checkout, and supplies no axis text. Each new field gets a `should_panic` case for the malformed shape, matching the existing ones.
- `tests/it/clause_ruling_index.rs`'s generated block moves from 79 to 81 clause lines, gaining the two RNA entries. It was **spliced programmatically** on its `BEGIN`/`END` markers from the block the failing test prints, never retyped — the file's own docs warn that hand-transcribing that block drops rows silently.

### Verification

Run from the worktree on `6f22cde9`, exit status captured into the log rather than inferred:

```
cargo run --features dev --bin generate_spec_fixture        EXIT_FIXTURE=0
cargo run --features dev --bin generate_spec_enumeration    EXIT_ENUM=0
cargo fmt --all --check                                     EXIT_FMT=0
cargo clippy --all-features --all-targets -- -D warnings    EXIT_CLIPPY=0
cargo nextest run --features dev --lib --test it            EXIT_NEXTEST=0   (10175 passed, 0 failed)
```

`FERRO_MANIFEST` was set throughout, so the reference-aware suites ran rather than skipped.

Ledger integrity was checked programmatically rather than by reading the diff: 21 records on `origin/main` and 21 on `HEAD`, none dropped relative to `origin/main` or to either pre-rebase tip, and no status change. The only difference in the whole `rulings` section is the two added `clauses` entries.

Representation-Change: none. Tests and the ruling ledger's citation list only; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched, and no normalized output moves.

### DIFF
--- a/tests/fixtures/grammar/hgvs_spec_normalization_overrides.json
+++ b/tests/fixtures/grammar/hgvs_spec_normalization_overrides.json
@@ -523,6 +523,14 @@
           "quote": "**exception**: two variants separated by one nucleotide, together affecting one amino acid, should be described as a \"delins\""
         },
         {
+          "clause": "docs/recommendations/RNA/delins.md:18",
+          "quote": "**exception**: two variants separated by one nucleotide, together affecting one amino acid, should be described as a \"delins\" (e.g., `r.142_144delinsugg` `p.(Arg48Trp)`)."
+        },
+        {
+          "clause": "docs/recommendations/RNA/substitution.md:18",
+          "quote": "**exception**: two variants separated by one nucleotide, together affecting one amino acid, should be described as a \"delins\" (e.g., `r.142_144delinsugg` (`p.Arg48Trp`))."
+        },
+        {
           "clause": "docs/recommendations/general.md:39",
           "quote": "The new recommendation will be: **two variants separated by less than two nucleotides should be described as a \"delins\"**."
         },
--- a/tests/it/clause_ruling_index.rs
+++ b/tests/it/clause_ruling_index.rs
@@ -103,7 +103,7 @@
 //! ```text
 //! CLAUSE -> RECORD INDEX
 //! 21 records, 18 decided / 3 undecided
-//! 79 clause lines, of which 19 are named by more than one record
+//! 81 clause lines, of which 19 are named by more than one record
 //!
 //! == every clause line ==
 //!
@@ -234,6 +234,8 @@
 //!     `delins-adjacent-members-when-both-consume-reference` — decided (governing)
 //! docs/recommendations/DNA/substitution.md:95
 //!     `delins-adjacent-members-when-both-consume-reference` — decided (cited)
+//! docs/recommendations/RNA/delins.md:18
+//!     `delins-codon-carve-out-gap-one` — decided (cited)
 //! docs/recommendations/RNA/delins.md:41
 //!     `canonical-form-choice-when-both-legal` — decided (cited)
 //! docs/recommendations/RNA/delins.md:70
@@ -246,6 +248,8 @@
 //!     `rna-repeat-range-plus-unit-redundancy` — undecided (cited)
 //! docs/recommendations/RNA/repeated.md:27
 //!     `rna-repeat-range-plus-unit-redundancy` — undecided (cited)
+//! docs/recommendations/RNA/substitution.md:18
+//!     `delins-codon-carve-out-gap-one` — decided (cited)
 //! docs/recommendations/checklist.md:5
 //!     `absolute-prohibition-enforcement-stage` — decided (governing)
 //! docs/recommendations/checklist.md:16
--- a/tests/it/common/rulings.rs
+++ b/tests/it/common/rulings.rs
@@ -1,11 +1,14 @@
 //! Reads the adjudication ledger — the `rulings` section of
 //! `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json`.
 //!
-//! Two integration tests need the same file for different reasons:
+//! Three integration tests need the same file for different reasons:
 //! `ruling_citation_currency.rs` wants `id -> status` so it can judge prose that
-//! makes a status claim, and `clause_ruling_index.rs` wants the citations so it
-//! can invert the ledger into a clause-to-record index. Parsing it twice would
-//! give two definitions of "a record", so both read it from here.
+//! makes a status claim, `clause_ruling_index.rs` wants the citations so it
+//! can invert the ledger into a clause-to-record index, and
+//! `ledger_clause_jurisdiction.rs` wants the citations, their quotes and
+//! `applies_to` so it can compare what a record rules on against the molecule
+//! directories it cites. Parsing it three times would give three definitions of
+//! "a record", so all three read it from here.
 //!
 //! Nothing in this module names a record id, deliberately: it is scanned by
 //! `ruling_citation_currency.rs` along with every other `.rs` file in the tree.
@@ -58,6 +61,13 @@ pub struct Citation {
     /// The clause as the record spells it — a repo-relative path into the spec
     /// checkout with a line or line range (`docs/…/delins.md:79-84`).
     pub clause: String,
+    /// The clause's quoted text, verbatim from the ledger.
+    ///
+    /// Required rather than optional, matching what `generate_spec_fixture`
+    /// already demands: a citation with no quote is a `file:line` nothing checks
+    /// against the spec checkout. `ledger_clause_jurisdiction.rs` also reads it,
+    /// so an absent quote would silently narrow the authority a clause supplies.
+    pub quote: String,
     /// How the record relates to this clause; see [`Role`].
     pub role: Role,
 }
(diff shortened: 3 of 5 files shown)


nh13:
@coderabbitai pause

PR #1686, opened 2026-08-11
test(normalize): pin that normalization is axis-preserving and frame-free axes ignore annotation

nh13:
Two properties of normalization that were true by construction and untested. Neither changes behaviour; both are guards.

### Invariant A — normalization preserves the coordinate axis

A `g.` in gives a `g.` out, `c.`→`c.`, and so on. Structural evidence it already held: a search for the projector anywhere under `src/normalize/` returns nothing — the dependency is one-way, projector → normalizer.

Measured over the real corpora, with denominators, because a census that compared nothing looks identical to a clean one:

| corpus | rows offered | compared | axis preserved | #487 `g.`→`m.` | violations |
| ClinVar 500k | 500,004 | 499,998 | 499,996 | 2 | **0** |
| ClinVar unique | 4,188,436 | 4,188,169 | 4,188,053 | 116 | **0** |
| CMRG | 4,826,063 | 4,826,001 | 4,826,001 | 0 | **0** |
| Paraphase | 435,235 | 435,219 | 435,219 | 0 | **0** |

**~9.95 million rows compared, zero violations.** An earlier attempt at this measurement reported `compared: 0` from a mis-guessed JSON field name — the denominator is what caught it, which is why every test here prints one and asserts it non-zero before believing a zero.

**The one sanctioned rewrite.** A `g.` description on a mitochondrial reference (`NC_012920` / `NC_001807`) is re-labelled `m.` (#487, `Normalizer::normalize_dispatch`). The reference molecule is unchanged — only the coordinate-system letter HGVS requires for it is corrected — so it is a label repair, not a change of axis. It is admitted as an enumerated exception and **each occurrence is re-checked against `Accession::is_mitochondrial`**, so the carve-out cannot launder a `g.`→`m.` rewrite on any other accession.

**The near-miss the test is built around.** Legacy-selector resolution rewrites `NG_012337.1(NM_003002.2):c.274G>T` to `NM_003002.2:c.274G>T` — the *accession* changes, the axis letter does not. The comparison is therefore on `CoordinateAxis` alone and never on the description's prefix, pinned by `the_check_reads_the_axis_letter_not_the_prefix`.

Corpora are git-lfs fixtures that skip green when absent, so the tests follow the existing skip idiom and report an absent corpus as a skip, never as a pass.

### Invariant B — an axis with no reading frame ignores transcript annotation

The stronger property, and the one that matters. Axis preservation alone would still permit a normalizer to consult a transcript for a `g.` input — same axis in, same axis out, frame-derived rule fired anyway.

So this is behavioural: normalize the same input twice, once against a provider that resolves the covering transcript and its CDS and once against one that cannot, and assert the outputs are identical. A genomic string must be derivable from its own reference; if the answer moves when annotation appears, it is importing information the reference does not carry.

**The negative control is load-bearing and it fires.** On the coding axis — where dependence is legitimate — the same provider pair produces *different* strings, and the difference is a merge that actually fired rather than a provider error. Without that, an "identical" result would be indistinguishable from a harness that cannot see anything. Two further guards keep the fixture honest: `the_fixture_puts_the_edited_bases_on_a_codon_of_the_annotated_transcript` asserts the premise, and `the_genomic_case_is_one_a_leaked_reading_frame_would_have_changed` proves the chosen case is one where a leak would have shown.

Covers `g.` and `m.`

### What is deliberately not asserted

**Nothing about the `n.` axis.** Whether `n.` on a coding transcript should carry the frame is open: normalizing `LRG_199t1:n.389_391delinsTGG` directly splits it, while a projection of the equivalent `c.` renders it merged, because that axis is derived by CDS-offset arithmetic rather than normalized on its own reference. Pinning either side would pre-empt a ruling, so the module doc says the omission is deliberate.

### Verification

Rebased onto `origin/main` at `6f22cde9`. Every figure in the table above was re-measured on that base and is unchanged; the `r.`/`o.` breadth figures quoted in the module docs were **not**, and have been corrected to the measured values (`r.` is 130 against 7 across the four bulk corpora — roughly 19x, not the two orders of magnitude previously claimed; `o.` is 1 against 0).

```
cargo nextest run --features dev --lib --test it -E 'test(normalize_axis_preserving) + test(normalize_frame_free_axis_independence)'
Summary [14.543s] 12 tests run: 12 passed, 10312 skipped

# full gate, mirroring CI's `test` job selection
cargo nextest run --features dev --lib --test it --test-threads=4 \
  -E 'not test(proptest) and not (<SWEEP_FILTER>)'
Summary [272.081s] 10105 tests run: 10105 passed, 214 skipped

cargo fmt --all                                  0
cargo clippy --all-features --all-targets -D warnings   0
```

These modules also land in `test-oracle`, which arms `FERRO_ASSERT_IDEMPOTENT`, `FERRO_ASSERT_REPARSE` and `FERRO_ASSERT_IN_BOUNDS` over what is now the suite's first full-corpus normalization (~9.95M rows; the sibling ClinVar/CMRG/Paraphase suites only parse). Measured armed: 12 passed, no oracle fires, ~2x runtime, longest test 35.8s locally.

### What the guards were measured to catch

Each assertion was mutation-tested rather than assumed:

- Dropping the `is_mitochondrial()` guard in `normalize_dispatch` so every `g.` is relabelled `m.` turns **all five** corpus censuses red.
- Weakening the carve-out to match on the axis letters alone turns `the_mitochondrial_carve_out_admits_only_mitochondrial_accessions` red.
- `the_check_reads_the_axis_letter_not_the_prefix` **survived** a mutation that made the census compare accessions as well as axes — the check its own name forbids — because it re-derived the comparison instead of calling it. It now asserts on `classify`, the census's own decision, and that mutation turns it red.
- A reading-frame leak on **both** genomic seams moves the answer only when it is provider-mediated; a leak that hard-codes a frame changes both arms of the A/B identically and is caught only by the exact-string pin. Either seam alone moves nothing. The module doc previously claimed all three behavioural assertions fail; measured, that is not so, and it has been corrected to what was observed.

Neither invariant flags behaviour the ruling ledger has decided: `projection-codon-exception-is-decided-by-the-rendered-axis` governs the projector, which invariant A never invokes and whose split-on-`g.` outcome invariant B agrees with; `delins-payload-coincidence-carve-out-is-coding-dna-scoped` scopes `:47` to `c.`, which is the axis the negative control merges on.

Representation-Change: none

### DIFF
--- a/tests/it/main.rs
+++ b/tests/it/main.rs
@@ -417,6 +417,7 @@ mod mutalyzer_normalize_tests;
 mod mutalyzer_tests;
 mod network_benchmark_tests;
 mod normalization_transcripts_exon_contract;
+mod normalize_axis_preserving;
 mod normalize_config_disambiguation;
 mod normalize_idempotency_proptest;
 mod normalize_property_tests;
--- a/tests/it/main.rs
+++ b/tests/it/main.rs
@@ -419,6 +419,7 @@ mod network_benchmark_tests;
 mod normalization_transcripts_exon_contract;
 mod normalize_axis_preserving;
 mod normalize_config_disambiguation;
+mod normalize_frame_free_axis_independence;
 mod normalize_idempotency_proptest;
 mod normalize_property_tests;
 mod normalize_reparse_invariant;
--- a/tests/it/normalize_axis_preserving.rs
+++ b/tests/it/normalize_axis_preserving.rs
@@ -1,0 +1,470 @@
+//! Normalization does not move a description onto a different coordinate axis.
+//!
+//! # What this asserts
+//!
+//! For every description the normalizer accepts, the coordinate-system letter of
+//! the output equals that of the input: `g.` in gives `g.` out, and likewise for
+//! `c.`/`n.`/`r.`/`p.`/`m.`/`o.`. Re-expressing one axis on another is the
+//! *projector's* job (`src/project/`), and the dependency runs one way —
+//! `src/normalize/` names no projector — so today the property holds by
+//! construction. That is exactly why it is worth pinning: nothing in the type
+//! system stops a future normalization rule from reaching for a projection, and
+//! a consumer that stored `c.` strings would get `g.` ones back with no error.
+//!
+//! # The one sanctioned rewrite, and why it is not a counter-example
+//!
+//! A `g.` description on a mitochondrial reference (`NC_012920` / `NC_001807`)
+//! is re-labelled `m.` (#487, `Normalizer::normalize_dispatch`). The reference
+//! molecule is unchanged — only the coordinate-system letter HGVS requires for
+//! it is corrected — so this is a label repair rather than a change of axis.
+//! It is admitted here as an enumerated exception and each occurrence is
+//! re-checked against [`Accession::is_mitochondrial`], so the carve-out cannot
+//! launder a `g.` -> `m.` rewrite on any other accession.
+//!
+//! # The near-miss to keep in mind: the accession may change, the axis may not
+//!
+//! Legacy-selector resolution rewrites `NG_012337.1(NM_003002.2):c.274G>T` to
+//! `NM_003002.2:c.274G>T`. The *accession* changes; the axis letter does not.
+//! So the comparison here is on [`CoordinateAxis`] alone and never on the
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

PR #1688, opened 2026-08-11
refactor(normalize): make each frame-derived rule declare its axis scope

nh13:
The frame-derived rules were gated on a bare `reading_frame: bool`. That answers *"does a codon exist here"* — the right question for a rule the recommendations state in **both** molecule documents, and the wrong one for a rule stated in only one. Nothing in a `bool` records which kind a given rule is, so the distinction lived in reviewers' heads.

It was got wrong there. `DNA/delins.md:47`'s payload-coincidence carve-out shipped gated on the frame and so reached the coding `r.` axis — which no `DNA/` clause can scope, and which has no counterpart clause of its own anywhere under `RNA/`.

### The change

`AxisFrame` now carries the `CisKind` it was resolved from, and offers two named scopes:

```rust
fn carries_translated_frame(&self) -> bool   // `c.` and coding `r.`
fn is_coding_dna(kind: CisKind) -> bool      // `c.` only
```

Both are exhaustive `match`es on `CisKind` with no wildcard, so **a new axis is a compile error at every rule** rather than a silent default. All six consumers of `frame.reading_frame` now read a named predicate; the raw field is left to the constructor and to `carries_translated_frame`.

The two predicates differ in arity, and that asymmetry is the #1241 point rather than an inconsistency: whether an axis *translates* is not answerable from the kind alone (`CisKind::Rna` says nothing about whether a CDS exists), whereas whether it is coding DNA is. So `is_coding_dna` takes the kind, which is also what lets the one gate that asks it — `payload_coalesce_applies`, which runs where no `AxisFrame` has been resolved — call it directly.

That call site is the point of the pair. `payload_coalesce_applies` previously carried its own second exhaustive `match` on `CisKind` spelling out the same coding-DNA scope. It now calls `AxisFrame::is_coding_dna`. A new `CisKind` was already a compile error in either copy, but two exhaustive matches can be given *different arms* and still compile, so the drift between them would have been silent — and it is exactly the scope the ruling below exists to protect.

### Which predicate is correct, settled by census rather than judgement

The codon exception — *"two variants separated by one nucleotide, together affecting one amino acid, should be described as a 'delins'"* — is stated in **eleven** places, and in **both** molecule documents:

| document | clauses |
|---|---|
| `general.md` | `:35` |
| `DNA/` | `deletion:19`, `delins:18`, `delins:42`, `duplication:23`, `insertion:20`, `inversion:21`, `substitution:17`, `substitution:37` |
| `RNA/` | `delins:18`, `substitution:18` |

Re-measured against the pinned spec checkout: `grep -rn 'together affecting one amino acid' assets/hgvs-nomenclature/docs/recommendations/` returns exactly those eleven lines and nothing else.

Note this is **not** the same eleven as the one in `projection-codon-exception-is-decided-by-the-rendered-axis`, which counts occurrences of "amino acid" across `DNA/*.md` only. Nor does it contradict `delins-codon-carve-out-gap-one`'s **nine**: that record counts the *identical* sentence, and the two `NOTE` restatements (`DNA/delins.md:42`, `DNA/substitution.md:37`) carry extra text. Three different scopes, three correct counts — worth stating because two of them land on the same number.

The RNA statements carry their own worked example (`r.142_144delinsugg`), and `delins-codon-carve-out-gap-one`'s `applies_to` lists both `r.142_144delinsugg` and `r.[142c>u;144a>g]`. So a rule resting on this exception reaches the coding RNA axis **on RNA's own authority**, not by inheriting a DNA clause — `carries_translated_frame` is the correct gate for it, and the existing sites were correctly scoped all along.

`is_coding_dna` exists for the other kind: a rule whose authority is stated only under `DNA/`. `delins.md:47` is the standing example, and getting that one wrong is what this pair exists to prevent.

Its doc now separates the two reasons the other axes are out, because collapsing them restates the scope too narrowly. `r.` is out on **jurisdiction**. `g.`/`m.`/`o.`/`n.` are DNA axes, where jurisdiction is no objection at all — they are out because `:47`'s stated reason has nothing to bite on where no protein is coded, and `general.md:34` governs instead. `n.` is the one that reading "DNA-only" as "everything except `r.`" would wrongly admit.

### Tests

`every_axis_declares_which_rule_scope_it_is_in` pins the whole matrix, including that a non-coding transcript has no frame (#1241 — `CisKind::Rna` alone says nothing about whether a CDS exists).

It also pins three rows that are **unreachable** through `axis_frame` — a genomic, mitochondrial or `n.` axis carrying `reading_frame: true` — and they are pinned precisely *because* they are unreachable. `carries_translated_frame`'s kind conjunct is the guard against a future constructor that stamps a frame where none belongs, and without those rows nothing distinguishes that conjunct from a no-op. Measured, not assumed: with them omitted, both widening the match to always-true and admitting `Tx` into it leave the suite green (1630-test targeted selection); with them present, both fail. Pinning `Tx` here decides nothing that `projection-codon-exception-is-decided-by-the-rendered-axis` leaves open — that record's open question is how *projection* renders `n.`, not whether `n.` carries a frame, and it states plainly that `n.` "carries no frame".

`a_coding_rna_axis_separates_the_two_scopes` asserts the pair has not collapsed into one predicate wearing two names. A coding `r.` is precisely the axis that separates them, so if that test ever passes vacuously the distinction it carries is gone. Both tests fail when `is_coding_dna` is mutated to admit `Rna` — the original defect.

### Verification

Rebased onto `0d9e6973`.

```
cargo nextest run --features dev --lib --test it   10164 tests run: 10164 passed, 150 skipped
scripts/run_oracle_suite.sh                        10227 tests run: 10227 passed, 280 skipped
cargo fmt --all --check                            0
cargo clippy --all-features --all-targets -- -D warnings   0
generate_spec_fixture / generate_spec_enumeration          0 / 0
```

Representation-Change: none. Provable by construction rather than by a corpus zero: `axis_frame` has exactly four construction sites, and `reading_frame: true` occurs at only two of them — the `CisKind::Cds` arm and the coding-`CisKind::Rna` arm — both of which satisfy `axis_may_translate`. So `carries_translated_frame()` is identically equal to `reading_frame` at every reachable state, and each of the six call sites swapped one for the other. `is_coding_dna` computes exactly what the `match` it replaced computed. No output can move.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2763,7 +2763,7 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
         partition_rule(),
         &ref_bytes[lo..hi_ref],
         &result[lo..hi_alt],
-        axis_min_separation(frame.reading_frame),
+        axis_min_separation(frame.carries_translated_frame()),
     );
     // Shadow the sequence-first splitter on the very same trimmed block, before
     // any 3'-shift or coalescing, so a reported disagreement is a disagreement
@@ -2788,7 +2788,7 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
         // Per-axis separation threshold (see `axis_min_separation`) — the same
         // distinction `apply_coding_codon_exception` uses below, via the same
         // `AxisFrame`.
-        let min_separation = axis_min_separation(frame.reading_frame);
+        let min_separation = axis_min_separation(frame.carries_translated_frame());
         let block = &ref_bytes[lo..hi_ref];
         let mut shadow = partition_block_sequence_first(block, &result[lo..hi_alt], min_separation);
         // `min_separation` alone captures only the distance half of
@@ -2800,7 +2800,7 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
         if let Some(shadow_pieces) = shadow.as_mut() {
             split_codon_incompatible_triplets(
                 shadow_pieces,
-                frame.reading_frame,
+                frame.carries_translated_frame(),
                 w_lo + lo as i64,
                 block,
             );
@@ -2858,7 +2858,7 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
     // measured by getting it wrong. See `coalesce_coding_frame_separation`.
     let unwidened = coalesce_coding_frame_separation(
         &mut pieces,
-        frame.reading_frame,
+        frame.carries_translated_frame(),
         hi_ref != hi_alt,
         w_lo,
         &ref_bytes,
@@ -2873,7 +2873,7 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
     // `LRG_199t1:c.850_901` example is out of scope by construction.
     split_concealed_separations(
         &mut pieces,
-        frame.reading_frame,
+        frame.carries_translated_frame(),
         hi_ref != hi_alt,
         w_lo,
         &ref_bytes,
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1689, opened 2026-08-12
test(ci): make the local seam-oracle run reproduce CI's selection

nh13:
## The documented local oracle command cannot pass

`CLAUDE.md` tells you four times to run an oracle like this:

```bash
FERRO_ASSERT_IDEMPOTENT=1 cargo nextest run --features dev
```

On a clean `origin/main` @ `1dd8148d` that is **10,174 tests run, 7 failed**. It has been red for as long as the spec corpus has existed, and nothing in the repo said so. An operator following the documentation gets a wall of red that is indistinguishable from a regression their change caused.

This PR does not make those 7 pass. It makes the local path reproduce what CI actually runs, from the same source of truth, and corrects the figures that had gone stale underneath the explanation.

Nothing under `src/` is touched. No output moves.

## The 7 failures are one cause, and the reference has nothing to do with it

Measured over the three modules that hold them (30 tests selected, so the denominator is stated):

| run | environment | result |
|---|---|---|
| A | 3 flags + `FERRO_MANIFEST` | 30 run, **7 failed** |
| B | 3 flags, no manifest | 30 run, **7 failed** — the same 7 |
| C | `FERRO_MANIFEST` only | 30 run, 0 failed |
| D | neither | 30 run, 0 failed |
| E | `FERRO_ASSERT_IDEMPOTENT` only | 30 run, **7 failed** |
| F | `FERRO_ASSERT_REPARSE` only | 30 run, **0 failed** |
| G | `FERRO_ASSERT_IN_BOUNDS` only | 30 run, **0 failed** |
| H | all four + manifest | 30 run, 12 failed |

So: `FERRO_ASSERT_IDEMPOTENT` alone is necessary and sufficient for all 7. The other two contribute nothing. A prepared reference contributes nothing either — every one of these modules is `MockProvider`-backed, and B is byte-for-byte the same failure set as A. Every one of the 7 panics at the same site, `src/normalize/mod.rs:1924`.

They split into two shapes, and `ci.yml`'s comment only stated the second:

- **Five ASSERT the defect the oracle aborts on.** The four `defect_non_idempotent_outputs` tests and `spec_corpus_regressions::an_insertion_at_the_cds_end_is_not_a_fixed_point` pin `c.*1delinsCTT` → `c.72_*1insCT` → `c.72delinsCCT` as a known non-fixed point. A test that pins a defect and an oracle that aborts on it cannot both run. There is no version of this the flag could pass.
- **Two COUNT it**, and arming the oracle makes the count read *better* than the truth, because `catch_unwind` files a panicking row as `declined` and its output never reaches its family's output set.

## The census figures in `ci.yml` were stale in three places

That comment was measured at `main` @ `35de96c8` and asserted "still the pinned figures, re-measured unchanged on `main` @ `5b54b2b6`". Two re-blessings have landed since — #1599 lowered 3' `non_idempotent_outputs` 7 → 4, #1536 moved `converged`/`split_two` — both recorded in `spec_conformance_axis.rs`'s own doc. Re-measured here at `1dd8148d`, armed against the committed pins:

```
3': declined 0 -> 4, non_idempotent_outputs 4 -> 0, converged 9141 -> 9145, split_two 2440 -> 2436
5': declined 0 -> 4, non_idempotent_outputs 4 -> 0, converged 8944 -> 8948, split_two 2706 -> 2702
```

Four families read as converged in each direction that do not converge — not the seven the comment claimed, which was the figure before #1599 lowered the count it derives from. The mechanism the comment describes is exactly right; only its numbers had drifted.

## What this adds

**`scripts/run_oracle_suite.sh`** — runs the suite the way `test-oracle` runs it. It reads *both* the exclusion and the flag set out of `ci.yml`; neither is copied, because a stale copy here fails in the flattering direction (it would exclude a module CI still runs armed, so a local run would go green on a defect CI is red on). It refuses a vacuous extraction and a zero-test run, in the same discipline as the existing `run_conformance_axis.sh`.

**Three guards in `tests/it/oracle_exclude_invariant.rs`** — the runner must exclude exactly what CI excludes, arm exactly what CI arms, and hardcode neither. The Rust side re-derives both from `ci.yml` using deliberately *different* anchors from the script's awk (job-block-and-quoted-value versus step-name-and-`run:` window), because a second opinion sharing the first's derivation is not a second opinion.

Each of the three was checked against a deliberate sabotage rather than assumed:

| sabotage | guard that fired |
|---|---|
| hardcode the exclusion list in the script | `does_not_hardcode_the_exclusion` + `excludes_exactly_what_ci_excludes` |
| add `FERRO_ASSERT_SEQUENCE` to the runner | `arms_exactly_the_flags_test_oracle_arms` |
| break the awk so `ci.yml` no longer parses | script exits 1; `excludes_exactly_what_ci_excludes` |

**Documentation.** The four always-red commands in `CLAUDE.md` are replaced by one section that states the trap once and points at the runner, per the repo's own rule against restating a rule in four places. The `FERRO_ASSERT_SEQUENCE` section keeps a bare command because `test-oracle` deliberately does not arm that oracle and the runner faithfully inherits that gap — its home is `sweeps`, and the section now says so and quantifies the 5 extra failures it raises here.

**The nightly's reproduce hint** named neither the four flags that job sets nor the three-module selection it sets them over, so it could not reproduce an oracle fire — which is that job's usual failure. It now carries both, and says why dropping the `-E` is what walks into the 7.

## What I deliberately did NOT change

- **`src/normalize/`.** The non-idempotency at the CDS/3'UTR boundary is a real, already-pinned defect. Fixing it is a representation change and the operator's call, not a test PR's.
- **`ORACLE_EXCLUDE` itself.** The exclusion is correct. Running these modules armed is the one action that would destroy the evidence, so "close the hole by running it in CI" is the wrong answer here; the honest answer is that CI already excludes it deliberately and the local path now excludes it identically.
- **The census pins.** Nothing re-blessed. The armed figures above are reported as a measurement of the collision, not as a new baseline.

## Verification

All from `/oracle-matrix` at `1dd8148d`, real exit status captured into each log rather than inferred from a pipeline:

```
scripts/run_oracle_suite.sh          -> 10,112 tests run, 10,112 passed, exit 0
bare armed full suite (3 flags)      -> 10,174 tests run, 7 failed,      exit 100
cargo fmt --all --check              -> exit 0
actionlint (both workflows)          -> clean
shellcheck scripts/run_oracle_suite.sh -> clean
oracle_exclude_invariant             -> 8 tests run, 8 passed (was 5)
```

The runner's 10,112 is 62 fewer than the bare run's 10,174 — the excluded modules — and the 7 failures sit inside those 62.

Representation-Change: none. Tests, a shell script, CI comments and documentation only; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched, and no normalized output moves.

## Summary by CodeRabbit

* **New Features**
  * Added a local oracle test runner that mirrors CI settings, exclusions, and enabled checks.
  * Added validation to detect configuration drift and reject failed or empty test selections.

* **Documentation**
  * Expanded guidance for running oracle tests locally, including expected failure counts and supported options.
  * Updated nightly failure instructions and clarified excluded test behavior.
  * Revised oracle exclusion details with current corpus counts and exclusion rationale.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,21 +63,48 @@ env:
   # in the plain `test` job; this is not a coverage exemption, it is a statement
   # about which instrument may be armed while another is measuring.
   #
-  # WHY: the corpus COUNTS non-idempotent outputs (7 at 3', 4 at 5' as pinned in
-  # `spec_conformance_axis.rs`) as a burn-down figure. `FERRO_ASSERT_IDEMPOTENT` PANICS
+  # WHY, REASON 1: the corpus COUNTS non-idempotent outputs (4 at 3', 4 at 5' as
+  # pinned in `spec_conformance_axis.rs`) as a burn-down figure.
+  # `FERRO_ASSERT_IDEMPOTENT` PANICS
   # on exactly that condition. The two agree perfectly about ferro — the oracle
   # fires on precisely the rows the corpus classifies — but a panic returns no
   # value, so the harness's `catch_unwind` files the row under `declined` and its
   # output never reaches the family's output set.
   #
   # That does not merely re-bucket a counter, it DESTROYS EVIDENCE, and in the
-  # flattering direction. Measured directly on `main` @ 35de96c8, where the
-  # figures were 7 at 3' and 4 at 5': in all 11 cases the non-idempotent output
-  # is the one that disagreed with its siblings, so removing it makes the family
-  # look unanimous. Confluence read 9147 converged / 2428 split-2 under the
-  # oracle against a true 9140 / 2435 — CI would have reported seven more
-  # converged families than actually converge. Those are still the pinned
-  # figures, re-measured unchanged on `main` @ 5b54b2b6.
+  # flattering direction: the non-idempotent output is the one that disagreed
+  # with its siblings, so removing it makes the family look unanimous.
+  #
+  # RE-MEASURED on `main` @ 1dd8148d, armed against the committed pins, both
+  # directions. The previous figures here were taken at `main` @ 35de96c8 and
+  # were stale in three places — #1599 re-blessed 3' `non_idempotent_outputs`
+  # 7 -> 4 and #1536 re-blessed `converged`/`split_two`, both recorded in
+  # `spec_conformance_axis.rs`'s own doc, so "still the pinned figures" had
+  # stopped being true:
+  #
+  #   3': declined 0 -> 4, non_idempotent_outputs 4 -> 0,
+  #       converged 9141 -> 9145, split_two 2440 -> 2436
+  #   5': declined 0 -> 4, non_idempotent_outputs 4 -> 0,
+  #       converged 8944 -> 8948, split_two 2706 -> 2702
+  #
+  # So CI would report FOUR more converged families than converge, in each
+  # direction — not the seven this comment used to claim, which was the figure
+  # before #1599 lowered the count it is derived from.
+  #
+  # WHY, REASON 2 — simpler, and not covered by the paragraphs above. Five tests
+  # in these modules do not merely count the defect, they ASSERT it: the four in
+  # `defect_non_idempotent_outputs` and
+  # `spec_corpus_regressions::an_insertion_at_the_cds_end_is_not_a_fixed_point`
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #1692, opened 2026-08-12
Derive a canonical HGVS description from a reference/alternate pair, with no spelling as input


nh13:
Today the only way to get a description out of ferro is to bring one in. A caller who has aligned reads and post-processed a BAM holds a reference/alternate pair over a window — not a spelling — so they must invent a spelling first and hand it to `normalize`. That choice then leaks into the output: two callers who observed the same bases, and who spelled them differently, get different descriptions back.

What is wanted is a function that takes the bases and returns a description, with two properties:

1. **Completely deterministic.** Same reference/alternate/accession/position in, same description out.
2. **Spec valid** — not necessarily the recommended form. `normalize` is applied afterwards, by the caller, if and when they want the recommended form.

## Why this is a separate surface, not a flag on `normalize`

The four normalization rules in `README.md` split cleanly along what a window-local function can know:

| | rules delivered | force | needs |
|---|---|---|---|
| deriving from a sequence pair | 1 (conformant), 4 (deterministic) | both *absolute* | the caller's arguments alone |
| `normalize`, applied after | 2 (recommended), 3 (confluent) | both *best effort* | the reference |

Rules 1 and 4 are exactly the two the README calls always achievable, and they are exactly the two that need nothing but the caller's four arguments. Rules 2 and 3 need the reference — rule 2's scope names the 3' rule explicitly, and a reference-anchored shift is precisely what a window-local function cannot perform. So the split is not a convenience boundary; it is where the available information runs out.

This also means the derivation reads no reference and has no hidden inputs, which is what makes it testable as a pure function of a `(position, reference, alternate)` triple.

## Scope

- Derive a description from a sequence pair, on the `g.` axis.
- The inverse direction: given a description and a reference, produce the pair of sequences it denotes, so the round trip can be checked.
- Re-anchoring, so a caller holding heterogeneous raw pairs can bind them all to one reference window rather than to whichever window each read happened to cover.
- The same surface from Python.

## Not in scope

- Any user-facing option that selects a normalization *form*. That is forbidden by rule 6, and the shuffle direction is not an exception: it chooses a placement within the window, and both directions must agree on whether an input is describable at all.
- Rules 2 and 3. Callers who want them call `normalize` afterwards, which is the whole point of the split.

Issue #1440, opened 2026-08-05
Weight bound judges a sequence-derived partition against the input's spelling


nh13:
### Summary

The weight bound in `canonicalize_from_sequence` judges a sequence-derived partition against a ceiling computed from **the input's own spelling**. Two spellings of one variant therefore get two different ceilings, so the same variant can be accepted under one spelling and refused under the other. That is a confluence defect in the gate itself, independent of which partition the splitter proposes.

`src/normalize/merge.rs:2263`:

```rust
if changed_columns_of_pieces(&pieces) > changed_columns_of_edits(&edits) {
    return None;
}
```

`changed_columns_of_pieces` is sequence-derived — it measures the partition the derivation produced. `changed_columns_of_edits` is not: it measures the members the *caller wrote*.

### Why this is the wrong comparand, and what the right one is

The question the bound asks is sound and worth keeping: *is the derivation claiming more changed columns than it has to?* It is what stops a re-derivation shredding two well-separated members into a run of substitutions across bases the variant never touched (`general.md:34`).

That question has a **sequence-derived answer**. `changed_columns_of_pieces` sums `max(|span|, |replacement|)` over the pieces, and that sum is exactly the number of non-matching columns in the alignment the pieces came from:

- a piece with `ref = 1, alt = 3` is one mismatched column plus two inserted ones,
- `ref = 3, alt = 1` is one plus two deleted ones,
- `ref = n, alt = n` is `n` mismatches.

Minimising that over all alignments is precisely **unit-cost Levenshtein distance** between the block's reference and observed sequences. So the derivation is column-minimal iff its changed-column count equals that distance, and the ceiling can be computed from `(reference, observed)` alone.

### Two properties this gives us

1. **Spelling-independence.** Every spelling of a variant produces the same `(reference, observed)` pair, so every spelling gets the same ceiling. This is the confluence property the incumbent lacks.
2. **No loosening.** The input's own spelling *is* an alignment, so its column count is at least the minimum. The new ceiling is therefore never larger than the incumbent's, and the gate is at least as strict everywhere. This is a proof, not a measurement — it does not depend on the corpus.

### Scope — what this does and does not fix

It fixes the gate. It does **not** fix #1419/#1420/#1421: on those rows both spellings now receive the same verdict, but both are *refused*, because the single-gap alignment sweep cannot reach a column-minimal partition in the first place. That is a partitioner problem and is tracked separately under #1235.

[...]

PR #1693, opened 2026-08-12
feat(normalize): derive a conformant HGVS description from a reference/alternate pair

nh13:
**PR 1 of 3, stacked.** PR 2 adds re-anchoring on top of this; PR 3 adds the Python surface on top of that. Reviewing in order is the intended reading.

Part of #1692.

## What this is

`from_sequences` takes an accession, a 1-based position, and a reference/alternate pair over that window, and returns an HGVS description. No spelling comes in, so no spelling can leak out. `to_sequences` is the inverse — given a description and a reference, the pair of sequences it denotes — which is what makes the round trip checkable.

The caller who needs this has aligned reads and a post-processed BAM. They hold bases, not a description, and today they have to invent one before ferro will talk to them.

## Why a separate surface rather than a flag on `normalize`

The four rules in `README.md` split along exactly what a window-local function can know:

| | rules delivered | force | needs |
|---|---|---|---|
| `from_sequences` | 1 (conformant), 4 (deterministic) | both *absolute* | the caller's arguments |
| `normalize`, applied after | 2 (recommended), 3 (confluent) | both *best effort* | the reference |

Rules 1 and 4 are the two the README calls always achievable, and they are the two that need nothing but the arguments. (Those arguments are five values, not four — the options struct is not inert, since `direction` moves a placement within the window. Earlier drafts of this body, the README and the Rust docs all said "four"; the count is withdrawn everywhere, the purity claim is not.) Rule 2's scope names the 3' rule explicitly and rule 3 is reference-anchored, so both need bases outside the window — which is precisely what this surface does not read. The boundary is where the information runs out, not where it was convenient to draw.

The practical consequence is the offer: **derive now, normalize later, or never.**

## How it derives

1. `trim_common_flanks` anchors both ends. This *is* the window-local 3' placement, for free.
2. `AlignmentDag::build` — unit-cost Levenshtein over the trimmed block only. It holds **every** minimal alignment at once, rather than one traceback.
3. `CanonicalAlignment` picks the member-count-minimal path, 3'-most among ties. The tie-break is `general.md:41`, an explicit spec rule, which is what `canonical-form-choice-when-both-legal` requires of us.
4. `shift_pieces` → `coalesce_adjacent_pieces` → `shrink_pieces_to_differences` → `rebuild_members`.
5. `retype_inversions`, then `verify_round_trip` — a **runtime** check, not a `debug_assert`, so the denotation property holds in release builds too.

Both endpoints are fixed by the caller: `position` pins `reference[0]` and nothing slides.

Deliberately *not* in the pipeline: `collect_canonical_edits` on the input, the #1440 weight bound (its threshold is the input's column count, and a sequence pair has no input spelling to count), the `normalize_core` alternation loop, and `coalesce_compensating_gap_split`.

### Why not an aligner's cost model

An affine-gap prototype was built and measured against this over 28,639 synthetic two-event blocks, counting descriptions with more edits than the block's edit distance:

| arm | non-minimal |
|---|---|
| bwa-mem (A1 B4 O6 E1), 5' | 2,595 (9.1%) |
| bwa-mem, 3' | 1,929 (6.7%) |
| minimap2 `lr:hq` dual-affine, 3' | 622 (2.2%) |
| `CanonicalAlignment` | **0 / 28,639** |

Worst case is `GGTAATCC → CCGAACTA`: distance 6, bwa-mem gives 12, because pricing a mismatch at 4 against a gap-extend at 1 makes delete-and-reinsert cheaper than substitute. Correct for read mapping, wrong for describing a variant.

To be clear about whose policy this is: **distance-minimality is ferro's, not the spec's.** `basics.md:38` lists stability, meaning, memorability and unequivocality — minimality is absent, and `DNA/delins.md:44-47` recommends a *non-minimal* description in its own worked example. Nothing here should be read as the spec requiring minimal edits.

## What it refuses, and why each refusal exists

- **A non-genomic accession.** `NM_TEST.1:g.9_10del` is well-formed and denotes nothing (`checklist.md:20`). Classification goes through `Accession::inferred_variant_type()` — the codebase's own table — rather than a hand-written prefix list, because the hand-written list leaked `ENSMUST`, `ENSRNOT`, `LRG_1t1`, `LRG_1p1`, `YP_` and `AP_`.
- **A 5'-edge insertion with no anchor.** HGVS writes an insertion between two positions, so a payload against the window's first base would need `position - 1`. How much flank is enough is direction-dependent, so the message says "more 5' flank" rather than naming a number it would sometimes get wrong.
- **A grid above the budget**, as a typed `BlockDecline` carrying the *trimmed block's* dimensions. It used to be an `Option`, which forced the caller to recompute the cell count on the untrimmed window to guess which refusal had fired — and get it wrong, reporting a wide window with a small change as a `max_grid_cells` problem.

The alphabet is the IUPAC-IUBMB set via `Base::from_char`, not `ACGTN`. `general.md:48` admits it and real data uses it: `NM_000518.4:c.[20A>T;249G>Y]` is a ClinVar row that the narrower alphabet refused. `X` and `-` are still refused (`standards.md:39`).

## The window condition

Stated exactly, because a caller needs to know when two windows agree:

> Two windows that both contain the whole interval over which the change can be placed derive the same description. A window that cuts that interval places the change at its own edge instead.

Both halves are pinned over four tandem tracts with both flanks varied to zero — necessity as much as sufficiency, since a truncating window tracks its own right edge exactly (`A8` walks `g.40del … g.45del`, then locks to `g.46del`). For an insertion or duplication the interval already reaches one base 5' of the tract, so "contains the interval" subsumes the flank requirement and no separate clause is needed.

`placement_bounded_by_window` reports the risk. It is deliberately a conservative "this could move" flag: an earlier draft documented it as firing exactly when the answer is read-dependent, which was false — a window flush with a tract is flagged *and* already correct.

## Evidence

| corpus | result |
|---|---|
| cis confluence corpus | 5,636 classes, **0 divergent** — the corpus's **genomic half**, and structurally all a `g.`-only surface can reach (see below) |
| validity sweep | 23,121 derived descriptions — 0 unparseable, 0 prohibited, 0 redenoted, 0 non-idempotent |
| the reported confluence pairs of #1419 / #1420 / #1421 | **9/9 converge**, both directions, two pads — where `normalize` converges 0/9, per its own pinned ratchet |
| real-world multi-member alleles (ClinVar / CMRG / Paraphase) | 443 windows, 59 derive, run in CI — and all 59 now keyed against their captured SPDI |

The multi-member axis is the one to look at hardest, and it **rides with this PR on purpose** — it is the coverage argument for the whole thing. Those 592 rows were previously skipped in CI for want of a reference; because `from_sequences` reads no reference and a window is just a value, the windows could be captured once (`examples/capture_multi_member_pairs.rs`) and the axis now runs hermetically. The fixture is gzipped rather than LFS deliberately: an absent LFS fixture makes these suites **skip green**, which is the worst available failure mode.

Its oracle is the `canonical_spdi` captured from the *input* at generation time, not anything this branch computes. An earlier draft cleared the denotation check by citing the code under test.

**Corrected in review: the draft after that one captured the key and then never compared against it.** The comparison was gated on `sequences.alternate != pair.alternate`, where `sequences` had been built from `pair.alternate` four statements earlier and `SequencePair::new` stores it verbatim — so the condition was constantly false, the `redenoted` assertion could not fire for any input, and the captured key was read only for `is_empty()`. `denotation_compared` still incremented, so the `> 0` guard passed and the dead check read as live. Three places described it as an independent oracle (the module doc, `CORPUS_LAYOUT.md`, `capture_multi_member_pairs.rs`); all three are corrected.

The comparison is now real: a `WindowProvider` serves the captured window, the derived description is keyed through `spdi::canonical_spdi` — which reaches the bases via `hgvs_to_spdi` and an SPDI splice, deliberately *not* through `apply_edits_to_window`, the applier the derivation and its own `verify_round_trip` share — and the two key strings are compared. All 59 derived rows compare, 0 disagree. Shown to discriminate: injecting a one-base coordinate-frame shift into the derivation (invisible to `verify_round_trip`, since both sides use the same frame) turns it red on five named rows.

**The `c.` axis is excluded structurally, and that is now asserted rather than described.** The corpus is generated `--axes g,c`; its `c.` classes are drawn against `NM_TEST.1`, which the accession gate refuses, so `5,636` is exactly `11,272 / 2`. The `unusable` counter recording those 24,271 refusals was printed and never asserted, and the `evaluated > 5_000` floor sat *under* the g-only value, so nothing in the test could have noticed the `c.` half going missing for some other reason. `AxisCensus` now asserts both sides of the exclusion and the floor is re-based onto the genomic class count.

**And the 384 refusals in the multi-member census are accession refusals, not grid refusals.** The census doc and the README both attributed them to `max_grid_cells`; splitting them by message gives **384 accession / 0 grid**, every one an `NM_` transcript. The structural rows the grid bound is really about are filtered out of the capture (`too_wide`) before the provider is ever asked, so they are absent from the fixture rather than refused inside it. The counter is now split.

## Representation change

Structural, not a corpus observation. The entire diff to pre-existing source across this branch is two visibility keywords and one extract-and-delegate refactor:

- `MAX_SEQFIRST_GRID_CELLS`: `const` → `pub(crate) const`. **The value is unchanged.**
- `apply_to_reference_padded`: private → `pub(crate)`.
- `is_revcomp` was in this list and is not any more: the widening was a leftover from an abandoned first implementation of `retype_inversions`, nothing outside `rules.rs` ever called it, and the only cross-module reference is a rustdoc link in `mod.rs` that predates this branch and resolved fine while the function was private. Reverted.
- `partition_block_canonical` now delegates to `partition_block_canonical_within(..., MAX_SEQFIRST_GRID_CELLS)` — the same constant, so the same behavior. (That function is reached only under `FERRO_PARTITION=canonical` and from `dump_partitions`; it decides no shipped result.)

Every other changed file under `src/` has **zero removed lines**. Nothing existing can have moved, because nothing existing was re-routed.

## Known gaps, stated rather than left to be found

- A derived description is not always the *recommended* form — that is rule 2's job, and the caller applies `normalize` to get it. Measured at 8.6% over a 6,000-shape sweep. Read that as **how much the split hands off, not a defect count**: every row in it is a description this surface never claimed to produce. Three classes: repeat notation (298 — `normalize` already emits the repeat form, and whether that preference reaches the `g.` axis at all is #1675's open question), member re-derivation (124 — rules 2+3), and inversion across members (85 — the DAG minimises edit distance and inversion is not in its cost model, so it partitions before anything can see it; same shape as `inversion-vs-two-delins-76-83`).
- That sweep ran against a 57-base synthetic contig, which holds one 4-mer homopolymer, one 8-mer and two dinucleotide tracts — no long repeats, no mixed repeats, nothing kilobase-scale. So **treat 8.6% as a floor**: a tract longer than the contig cannot be generated at all. Re-measuring against real reference windows is filed as #1699.
- `to_sequences` works on any axis; `from_sequences` emits `g.` and `m.` only, so the round trip is genomic-only. That is why the corpus run declines 24,271 `c.`-axis spellings rather than pretending to judge them — and why every headline number here is a claim about the corpus's genomic half. `AxisCensus` asserts that.

## Fixed in review

Behavioural, all on this branch's own new surface, so none of it moves a shipped string:

- **Mitochondrial accessions emitted `g.`, not `m.`** — `NC_012920.1:g.6C>T` on the rCRS. `Accession::is_mitochondrial`'s own doc states that HGVS requires the `m.` axis for these, so this was a rule-1 defect on the most-used mitochondrial reference, in a module claiming rule 1. `MtVariant` carries a `GenomeInterval`, so only the template changes; `retype_inversions` now matches `Mt` alongside `Genome`.
- **A soft-masked window was admitted and then refused by its own round trip.** `validate` folds case (`Base::from_char` does), but `apply_edits_to_window` copies the reference verbatim while splicing an upper-case payload, so `acgt` → `atgt` died with `expected atgt, got aTgt`. `del`/`dup` passed throughout, which is why the corpora never saw it. Both sequences are now folded at the entry, and `to_sequences` folds the whole window rather than only the flank it prepends — which is what manufactured the mixed-case pair in the first place.
- **`prepend_five_prime_flank` derived the new start from what the provider *returned*.** A short read serves the front of the requested range; labelling it as the last `served` bases shifts every derived coordinate silently, and over-serving underflows in a profile with no `overflow-checks`. It now requires the flank exactly or declines it, which is `fetch_window`'s own discipline.
- **`U` was admitted asymmetrically** — `ACGU` → `ACGT` emitted `g.8U>T`, a `g.` description naming an RNA base, while `ACGT` → `AUGT` refused with the round trip's internal message. Refused at the argument on both sides now.
- **`P12345` derived `P12345:g.6C>G`.** `inferred_variant_type`'s UniProt arm keys on a one-character prefix and this parser does not split a bare UniProt accession that way. `NR_`/`XR_` got the uninformative parse-failure message because the classification ran *after* a `g.=` parse the parser refuses on a non-coding transcript; the classification runs first now. `ENSG` is deliberately left alone — `inferred_variant_type` classifies an Ensembl gene as `g`, and a second disagreeing classifier is the thing that comment argues against.
- **`SequencePair::end()` could underflow** on a pair whose `pub` fields were mutated after construction.
- **`to_sequences`' rustdoc said "only the 3' side is padded, and that is deliberate rather than an oversight"** four lines above the call to `prepend_five_prime_flank`. The window is `span + 2 * pad`.
- **`denoted_bases`' doc claimed independence it does not have** — it shares `apply_edits_to_window` with the derivation, which is what *prevents* independence. Withdrawn, and `Normalizer::from_sequences` now states exactly which seam oracles the derivation reaches (`normalize = true` only) and why the other entries cannot reach them rather than leaving it implied.
- **`a_derivation_over_its_own_span_reproduces_it` asserted a disjunction over nearly every emittable form**, not reproduction, and its two windows were mismatched (`reference[lo..=hi]` against `alternate[lo..]`). The alternate's mirror is measured from the 3' end; with the windows matched, equality is asserted and holds over 200,000 generated cases.
- **The README taught the flag with an inverted value.** `from_sequences_detailed("NC_000001.11", 1000, "AGCG", "AG")` gives `placement_bounded_by_window = true`, not `False` — the deletion lands flush with the window's 3' edge. The example now shows both windows, which is the flag's whole meaning.

Representation-Change: none. Structural: the only diff to pre-existing source is two visibility keywords and one delegation to the same constant.


nh13:
@coderabbitai pause

PR #1694, opened 2026-08-12
feat(normalize): re-anchor a window to a bound the variant must not leave

nh13:
**PR 2 of 3, stacked on PR 1.** Its base is PR 1's branch, so the diff shown here is only the re-anchoring work.

Part of #1692.

## The problem this solves

A caller sweeping a BAM does not get one window — they get a different window per read, because each read covers what it happens to cover. PR 1 established that two windows agree exactly when both contain the whole interval over which the change can be placed. That is a property of the *windows*, and until now the caller had no way to make it hold: they could observe that their windows disagreed but not do anything about it.

Re-anchoring lets them bind heterogeneous raw pairs to one chosen reference window, so the derived description is a property of the locus rather than of whichever read reached furthest.

## The surface

- `SequencePair::new` — build a pair from bases you already hold. The type was otherwise only ever *returned*, and being `#[non_exhaustive]` it was unbuildable outside the crate, which put re-anchoring out of reach of exactly the caller it is for: one who has bases and no description. It shares `from_sequences`'s validator, so a pair that constructs is a pair that derives, and the refusal arrives at the argument that caused it rather than one call later.
- `SequencePair::trim_to` — narrow to a sub-window, matches only.
- `Normalizer::reanchor` — narrow *then* widen, so the trim is always over bases the caller actually observed. Widening reads the reference through `fetch_flank`, which **refuses a short serve** rather than silently returning a narrower window than asked for — that refusal is the point, since a silently-short window is exactly the condition that makes two callers disagree.
- The same three from Python, plus stubs.

It refuses to run off either end of a sequence.

`new` sets `window_is_final` to `false`, which is the only honest answer: a caller-supplied window carries no evidence about whether its 3' edge is where the sequence stops or where the read did.

## What this is *not*

**It is not a confluence tool, and the README says so.** Both existing paths already converge heterogeneous spellings — `to_sequences` erases the spelling, and that is what makes the corpora agree. Re-anchoring is a *bounding* tool: it lets a caller state the window a variant must not leave. Selling it as the thing that makes descriptions agree would be selling something the measurements attribute elsewhere.

Nor is it an option that selects a normalization form; it selects a window, and every window it can produce is one a caller could have passed directly.

Representation-Change: none. Adds two new entry points; no existing path is re-routed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,9 +40,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **`Normalizer::from_sequences`** — the same derivation against a held reference, which lets it
   additionally range-check the interval and optionally `normalize` the result.
 
+- **Re-anchoring, for callers whose variant must stay inside a region.** `SequencePair::trim_to`
+  narrows a window to given bounds, trimming matching bases only and needing no reference;
+  `Normalizer::reanchor` moves both edges, padding from the reference where it must widen. Bounds
+  are 1-based inclusive and optional per edge. Both refuse rather than clamp — including a bound
+  outside the sequence, since a window silently pulled back to the contig would hide a bug upstream
+  of the call.
+
+  **`reanchor` moves a window's edges; it does not relocate the window.** The window asked for must
+  overlap the pair's own, because the changed bases exist only in the pair — a disjoint request is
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

PR #1695, opened 2026-08-12
feat(python): expose from_sequences, to_sequences and re-anchoring

### LINKED ISSUE #1692: Derive a canonical HGVS description from a reference/alternate pair, with no spelling as input

nh13:
**PR 3 of 3, stacked on PR 2.** Its base is PR 2's branch, so the diff shown here is only the Python surface.

Closes #1692.

This is the last of the three, so it carries `Closes`; PRs 1 and 2 say `Part of` so that merging them does not close the issue while the stack is still open.

## What this adds

The Python bindings for the surface PRs 1 and 2 introduce — nothing new about the derivation itself, which is why it is a separate PR rather than 500 more lines on the first one.

- free functions `from_sequences` and `from_sequences_detailed`
- `Normalizer.to_sequences`, `Normalizer.from_sequences` and `Normalizer.reanchor`
- `SequencePair` and `DerivedDescription` as Python classes, with getters and a `__repr__`
- type stubs in `python/ferro_hgvs/__init__.pyi`

## Two things worth a reviewer's attention

**The GIL is released around the derivation.** Both free functions take `py: Python<'_>` and `detach`, following #1455. The derivation reads no reference and touches no Python object once its arguments are copied in, so there is nothing to hold the GIL for — and a caller sweeping a BAM is exactly the workload that would notice.

**`window_is_final` is named for what it means.** It was `full_pad_served` until a Python test asserted the literal reading of that name and failed: the flag is true when the window ends where the pad asked **or** where the sequence itself ends. So the ordinary answer near a contig end is `true` — there was nothing further to read, so the roll is settled either way — and `false` means the provider stopped short of both. It is also 3'-only; `to_sequences` prepends its 5' flank separately and that half is not reflected in the flag.

The Rust field was renamed to match, so the two cannot drift.

## Tests

`tests/python/test_from_sequences.py` — 35 tests covering the free functions, both `Normalizer` methods, the classes and their reprs, the refusals, and the flag semantics above. `ruff` and `mypy` are clean.

Representation-Change: none. Bindings only; no Rust normalization path changes.

### DIFF
--- a/docs/python_reference_guide.md
+++ b/docs/python_reference_guide.md
@@ -14,6 +14,7 @@ harness) before trusting a genome-aware result.
 | A **synthetic / local** construct (your own FASTA + annotation) and you need intronic / `g.` / exon-junction rules | `ferro_hgvs.convert_gff(...)` / `ferro_hgvs.build_transcript(...)` in-process (or the `ferro convert-gff` / `ferro build-transcript` CLIs, all with `emit_genomic_sequences=True` / `--emit-genomic-sequences`) → `Normalizer(reference_json="transcripts.json")` | ✅ yes |
 | Transcript-level normalization only (exonic SNV/indel shuffle; no introns, no `g.` axis) | a `transcripts.json` **without** `genomic_sequences` → `Normalizer(reference_json=...)` | ❌ by design (see the boundary below) |
 | Just trying the API | `Normalizer()` (built-in toy data) | ❌ toy data |
+| **Bases rather than a description** — a window out of a BAM, a VCF row, an aligner's output | `ferro_hgvs.from_sequences(...)` — no `Normalizer`, no reference, nothing to set up | n/a — reads no reference by design; see [below](#deriving-a-description-with-no-reference-at-all) |
 
 "Genome-aware rules" = intronic positions (`c.10+5del`), the cross-exon/intron
 3′-shift (#670), and the genomic (`g.`) axis / `project_to_genomic`. See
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

Issue #1041, opened 2026-07-16
Normalizer under-recognizes `inv`: whole-run reverse-complement left as `delins` (v0.8.1)


msto:
> Authorship: the content below was drafted by GitHub Copilot CLI (an AI coding agent) and filed via gh under @msto's account. It reflects the agent's analysis, not a statement authored by @msto.

> ### Summary
> In a ferro-vs-mutalyzer normalization comparison run on **v0.8.1**, ferro fails to recognize a **whole-run inversion**: the entire changed span's alt equals the reverse complement of the reference, so `inv` is the compliant description, but ferro writes the same span as a plain `delins`. Mutalyzer types it as `inv`.
>
> ### Anonymized example
> Contig name, coordinates, and bases are synthetic; only the discrepancy structure and the reverse-complement relationship that drives the call are retained.
>
> ```
> raw   : anon_ctg:g.[31G>T;32C>G;33A>C]
> mut   : anon_ctg:g.31_33inv          # recognized as an inversion
> ferro : anon_ctg:g.31_33delinsTGC    # same span, left as a delins
>
> ref[31_33] = GCA   revcomp(GCA) = TGC   # equals the alt, so inv is the compliant call
> ```
>
> ### Expected
> Because the alt over the full span equals `revcomp(ref[31_33])`, the span is a true inversion and should normalize to `31_33inv` (as mutalyzer produces).
>
> ### Observed
> ferro emits `31_33delinsTGC` — a same-span `delins` — missing the reverse-complement identity and therefore under-recognizing the `inv`.
>
> ### Impact
> A less common ferro→mutalyzer divergence observed in the v0.8.1 comparison.
>
> ### Related
> Complementary to #160 (recognize reverse-complement sub-spans as `inv` within compound variants): here the reverse-complement identity holds for the **whole** same-span run yet is not recognized as `inv`.

PR #1696, opened 2026-08-12
test(rulings): correct a false cost claim in the payload-coincidence axis-scope record

nh13:
`delins-payload-coincidence-carve-out-is-coding-dna-scoped` states, as its recorded cost, that the ruling "costs a stored 723-base `g.` delins its 193-member re-derivation, deliberately" — naming `NG_009874.2:g.158330_159052delins...` from the ClinVar corpus and claiming the 193 members are kept "PRECISELY because the gate declines on a `g.` row where the ungated pass would have merged it back".

That is false, and it was measured false rather than argued. The row is byte-identical under `FERRO_PARTITION=canonical` and `FERRO_PARTITION=canonical-coalesced`, so the gate this record installs cannot be what keeps it split, and reversing the ruling would not recover it.

## What actually keeps that row split

The payload-coincidence pass would have declined it axis-blind, on two independent counts:

- two of its gaps are **9**, against `COALESCE_MAX_SEPARATION = 8`;
- the minimum number of substitutions over every order-preserving embedding of its 697-nt payload into its 723-base span is **407**, against `COALESCE_MISMATCH_BUDGET = 1`.

Either alone is disqualifying, and neither reads the axis.

The row's real cause is unrelated to this carve-out: it is an exact **697-base inversion** with 16- and 10-base flanking deletions. That decomposition is unique — enumerating all 27 possible flank splits, only (16, 10) reconstitutes the payload — and ferro's partitioner optimises in an algebra with no inversion operator, so 193 members is genuinely minimal in the model it uses. That belongs to the `coalesce_whole_block_inversion` family (#160, #1034, #1041, #1517, #1637), and the record now says so.

## What replaces it

The cost paragraph is restated without a worst-case exhibit, because none of the extreme ones belong to this ruling. It now points at the figures the record had already measured for itself: 380 genuine merges removed over the 11,272-class designed cis corpus, and 340 rows over the two real corpora.

Nothing else in the record changes — the ruling, its grounds, its scope, its counterweight and its structural-zero paragraphs are untouched. The diff is one line, +1/-1.

## Why this matters beyond the wording

A decided record is the project's escalation register. This one offered a concrete, dramatic exhibit as the price of the ruling, so the natural next move for a reader who dislikes that price is to reopen the ruling — which would not have recovered the row, and would have cost the 380 merges for nothing. The correction redirects that pressure to where the row actually lives.

## Verification

- Spliced textually, never round-tripped through a JSON library: `+1 / -1`, and the file still parses to 19 records.
- `cargo run --features dev --bin generate_spec_fixture` and `--bin generate_spec_enumeration` both exit 0 — plain generation is the guard that every curated override still resolves against the spec checkout.
- `cargo nextest run --features dev -E 'test(ruling) or test(override) or test(clause) or test(ledger)'` — **104 run, 104 passed, 0 failed**.
- No generated artifact drifts: `git status --porcelain` shows only the ledger file.

Representation-Change: none


nh13:
@coderabbitai pause

Issue #1044, opened 2026-07-17
normalize: clamp the mitochondrial/circular window fetch to contig length (sibling of #1041)


nh13:
PR #1042 fixed the genomic normalization path (`normalize_genome`) so the `±window_size` fetch window is clamped to the contig length, preventing a past-EOF read from silently dropping a near-3'-end variant into the minimal-notation fallback (which skips 3' shifting and the `delins → inv/sub/dup` canonicalization).

The mitochondrial / circular (`m.`/`o.`) normalization path has the **identical** unclamped fetch and was not touched by #1041/#1042. In `src/normalize/mod.rs` (the non-origin-crossing `m.`/`o.` path, ~L5253):

```rust
let window_start = start.saturating_sub(self.config.window_size);
let seq_result = self.provider.get_sequence(
    &accession,
    window_start,
    end.saturating_add(self.config.window_size),   // <-- unclamped, like the pre-#1041 genomic path
);
let ref_seq = match seq_result {
    Ok(s) => s,
    Err(_) => return Ok(mt_fallback(variant)),      // <-- same silent fallback
};
```

Its own comment even says "Non-origin-crossing path: identical to genomic." Because every provider errors on a past-EOF read rather than clamping, a **non-wraparound** (`start <= end`) `m.`/`o.` variant within `window_size` (100 bp) of the contig 3' end fetches past EOF, `get_sequence` errors, and the variant falls into `mt_fallback` — losing 3' shift and the `delins → inv` canonicalization, exactly the #1041 symptom.

This is realistically reachable: the human mtDNA reference (NC_012920.1) is 16,569 bp and the control region / D-loop sits near the 3' end, so variants land within 100 bp of the contig end.

## Suggested fix

Apply the same `end <= len` guarded clamp #1042 used on the genomic path (fetch the contig length via `get_sequence_length`, clamp `fetch_end = raw_end.min(len)` only when `end <= len`; otherwise keep the raw window so a genuine past-contig-end span still passes through). Wraparound (`start > end`) ranges already route to `mt_fallback` and remain out of scope per #129. A shared helper for the clamp would keep the genomic and mito paths from drifting.

## Test

A near-3'-end non-wraparound `m.` whole-run reverse-complement delins that should canonicalize to `inv` (mirroring `issue_1041_repro.rs`, but on a circular contig short enough that the window overruns EOF).

Found while dual-reviewing #1042.

PR #1697, opened 2026-08-12
fix(normalize): grow the fetch window so a long homopolymer converges

### LINKED ISSUE #1691: normalize is not idempotent on a homopolymer longer than the sequence-compare window
`Normalizer::normalize` never reaches a fixed point on a long poly-A homopolymer: the 3' shuffle advances a bounded amount per pass, so repeated normalization keeps moving the description instead of converging.

## Reproducer

Against a 200,000-base poly-A reference served by a bare `MockProvider`, with no PR code involved:

```
g.100del -> g.200del -> g.300del -> g.400del -> ...
```

Reproduced on a clean detached `origin/main` worktree using only `MockProvider` + `Normalizer`, so this is not specific to any open branch.

## How it surfaced

PR #1652 adds `equivalence_cross_axis_rung::a_span_past_the_compare_window_cap_is_indeterminate`, which is the first test in the suite to build a homopolymer longer than `MAX_SEQUENCE_COMPARE_WINDOW` (`100_000`). That test fires `FERRO_ASSERT_IDEMPOTENT` at `src/normalize/mod.rs:1924` and reddens the `Test oracle` job. The PR exposes the defect; it does not cause it.

`MAX_SEQUENCE_COMPARE_WINDOW` itself is byte-identical (`100_000`) on `origin/main`, on that branch, and on its pre-rebase tip — no bound moves.

## Why it matters

`FERRO_ASSERT_IDEMPOTENT` is one of the three seam oracles CI's `Test oracle` job does set (`ci.yml:1013-1015`), so any future test that constructs a homopolymer past the bound will redden a required check. More substantively, non-convergence on a legal input is a normalizer defect independent of whether a test happens to build one.

## Notes for whoever picks this up

The shuffle advancing "a bounded amount per pass" is the mechanism to confirm first — specifically whether the bound is the compare window or something else, and whether the same non-convergence is reachable with a shorter homopolymer than 200,000 bases. A cheaper reproducer would make the fix much easier to verify.

nh13:
Closes #1691

## The defect

`Normalizer::normalize` never reaches a fixed point inside a homopolymer longer than the reference fetch window. On a poly-A tract it walks:

```
c:g.65del -> c:g.165del -> c:g.265del -> c:g.365del -> ...
```

**The step is exactly 100, which is `NormalizeConfig::window_size` — that constant is the bound, not `MAX_SEQUENCE_COMPARE_WINDOW`.** The issue names the latter because PR #1652's new test happens to build a 200,000-base tract, but the compare window is three orders of magnitude larger than the bound actually being hit and is byte-identical on both sides. The defect reproduces on a **2,000**-base tract, not a 200,000-base one.

`normalize_genome` (and its mitochondrial mirror) fetch `±window_size` around the variant as a *cost* bound, then hand `normalize_na_edit` a `Boundaries::new(0, ref_seq.len())`. The shuffle therefore treats the window's far edges as hard limits — but those edges are *where the fetch stopped*, not properties of the contig. Inside a tract longer than the window the shuffle reports the edge it reached; re-normalizing that answer re-centres the window on it and advances another `window_size` bases, forever.

### Why nobody hit it sooner

`canonicalize_from_sequence` alternates the sequence-first pass (its own `±CANONICAL_PAD = 128` window) with `normalize_core`, up to `MAX_PASSES = 4`. A tract those four alternations can span end-to-end is silently rescued, so a 200-base and a 600-base run both converge on unfixed code. The defect starts at the first tract the alternation cannot span, and that onset is **between 700 and 800 bases** — measured on `origin/main` by counting calls from `c:g.65del` to a fixed point:

```
run=600  1 call     run=800  2 calls     run=1000  4 calls
run=700  1 call     run=900  3 calls     run=2000 12 calls
```

An earlier revision of this description bracketed the onset as "between 600 and 2,000", which is true but loose. The 600 in `issue_1691_short_tracts_are_unaffected` is still the right choice — it sits inside the converging range rather than on its edge — and that test pins the rescue path so the fix cannot regress it.

## The fix

`Normalizer::normalize_in_grown_window` re-fetches geometrically until the normalized result rests **strictly inside** the window, and both axes that used a variant-relative window now go through it. Growth is attempted only where the edge is *provably* artificial — the contig is known to continue past the window's 3' end, or the window's 5' end is not the contig's first base — so where flushness cannot be established the window is left exactly as it was, the same conservative choice `clamp_fetch_end_to_contig` already makes.

Shared between the genomic and mitochondrial paths for the reason `clamp_fetch_end_to_contig` is: two copies of this logic already drifted once (#1042 vs. #1044). mtDNA carries no run remotely this long, so that half is a structural guarantee rather than a behaviour change for any real `m.` input.

The common case still costs exactly one fetch: a result resting well inside the configured window never consults the growth ladder.

### The cap: what it bounds, and where its value comes from

`MAX_SHUFFLE_FETCH_WINDOW` is 64 KiB, **reused from** `MAX_INTRONIC_SHUFFLE_WINDOW` rather than independently sourced — that constant is itself unsourced. What justifies the reuse is that the pathogenic expansions that do run long (FMR1 `CGG`, RFC1 `AAGGG`) reach thousands of bases, still an order of magnitude inside 64 KiB, so a run that exhausts the cap is a structural feature rather than a repeat tract. It is a cost bound picked to sit clear of the biology, not one derived from it.

**It bounds the half-width, not the span read.** `normalize_in_grown_window` fetches `[start - window, end + window]`, so the two constants are numerically equal and semantically a factor of two apart — `MAX_INTRONIC_SHUFFLE_WINDOW` is tested as a **full** span. Measured: a 60,000-base tract at contig offset 100,000 converges to `g.160000del` off a window covering `[34,465 .. 165,537]`, i.e. 131,073 bases. The *travel* bound is the half-width itself — a 65,000-base tract converges and a 66,000-base one is refused.

### Past the cap: the travel is refused, the typing is not

Past the cap the **shift** is refused outright rather than truncated. That is the load-bearing choice and it is not a corner case: a truncated shift is not a fix at all, it merely changes the walk's step size from 100 to 65,536 and the description still never settles.

**Refusing the travel must not also refuse the re-typing, and a first cut of this PR did exactly that.** The capped branch returned `canonicalize_genome_variant(variant)` / `mt_fallback(variant)`, neither of which reads the reference, so the reference-aware edit-type canonicalization went with the shift. Measured at that revision on a 70,000-base poly-A contig:

```
input               was              now              (200-base control)
c:g.65_66insA    -> c:g.65_66insA -> c:g.67dup     |  c:g.264dup
c:g.65delinsG    -> c:g.65delinsG -> c:g.65A>G     |  c:g.65A>G
m.65_66insA      -> m.65_66insA   -> m.67dup       |  (same shape)
```

`c:g.65_66insA` inserts an `A` directly 3' of an `A`, which is a tandem duplication. `DNA/duplication.md:18` — "when a variant can be described as a duplication, it **must** be described as a duplication and not as, e.g., an insertion"; `DNA/insertion.md:17` renders that spelling as `<code class="invalid">`; `general.md:57` puts duplication above insertion. That is rule 1 of the README ruleset, which is the one rule declared absolute and never traded — and idempotency is not one of the four output rules at all (rule 4 is determinism, which a refusal does not affect). So the branch was trading the absolute rule for a property that is not on the list.

The verdict is available at the cap, which is why the trade was avoidable: it is decided by the bases flanking the edit, not by how far the tract runs. `Normalizer::canonicalize_without_shifting` re-normalizes the edit against a **slice** of the window it already holds — width derived from the edit alone, never from the tract — with the shuffle pinned by the new `Boundaries::pinned()`. Both halves are load-bearing. Pinning alone is not enough, because `insertion_to_duplication` consults no boundaries and would anchor the `dup` at the far end of whatever tract it can see, which is the walk again wearing a different edit type. Slicing alone is not enough for the reason the cap exists. Together, every base the answer reads sits at a fixed offset from the edit the caller handed in, so re-normalizing the answer reads the same relative bases and returns it unchanged.

The one thing that stays refused besides travel is **extent**: repeat notation's `<first>_<last>unit[N]` states where the reference tract begins and ends, which is exactly the unanswerable question, and a *truncated* extent is worse than none — a false claim about the reference rather than a merely unshifted description. So a `Repeat` input, and any promotion into repeat notation, is declined and falls back to the syntactic cleanup. `c:g.65_66insAAA` past the cap therefore stays `c:g.65_66insAAA` rather than acquiring the truncated `c:g.65_166A[105]` that unfixed `origin/main` emits for it.

Every capped shape measured is a fixed point, on both axes:

```
c:g.65_66insA     -> c:g.67dup         -> c:g.67dup
c:g.65_66insAAA   -> c:g.65_66insAAA   -> c:g.65_66insAAA
c:g.65del         -> c:g.65del         -> c:g.65del
c:g.65delA        -> c:g.65del         -> c:g.65del
c:g.65dup         -> c:g.65dup         -> c:g.65dup
c:g.65_66dup      -> c:g.65_66dup      -> c:g.65_66dup
c:g.65delinsG     -> c:g.65A>G         -> c:g.65A>G
c:g.65_66delinsAA -> c:g.65_66=        -> c:g.65_66=
c:g.65_67inv      -> c:g.65_67inv      -> c:g.65_67inv
m.65_66insA       -> m.67dup           -> m.67dup
```

A note on wording, because it is what made the defect invisible: the capped branch does **not** "return the caller's own input". It never did — `canonicalize_genome_variant` strips a stated reference (`c:g.65delA` -> `c:g.65del`) and the mito side additionally runs `apply_canonical_split`. It was a *partial* canonicalization with the reference-aware half missing, and describing it as a no-op is what let the missing half go unnoticed.

## Tests

Eight tests in `tests/it/issue_1691_homopolymer_convergence.rs`. Six were confirmed RED on unfixed `origin/main`, one is the negative control that must stay green on both sides, and the eighth is the guard for the section above, confirmed RED against this PR's own previous revision:

```
c:g.65_66insA is a tandem duplication; past the growth cap it was emitted as
c:g.65_66insA, the spelling `DNA/insertion.md:17` marks invalid
```

It runs both the genomic and the mitochondrial axis, asserts the output is a `dup` rather than an `ins`, and asserts it is a fixed point. **No pre-existing test discriminated it**, which is why CI was green: `issue_1691_a_tract_past_the_growth_cap_refuses_to_shift_and_is_stable` uses a plain `del`, whose canonical form *is* its input, so `once == authored` cannot separate "refused the shift" from "refused all reference-aware work"; and `issue_1691_insertions_and_duplications_converge_too` runs at `TRACT = 2000`, below the cap.

The convergence test asserts on the *chain* rather than a single equality, so a failure prints the walk and its step size — which is the evidence identifying the bound:

```
normalize did not converge; it walked ["c:g.65del", "c:g.165del", "c:g.265del",
  "c:g.365del", "c:g.465del", "c:g.565del", "c:g.665del", "c:g.765del", "c:g.865del"]
```

Plus two unit tests on the growth ladder itself: that it terminates at the cap from any start, and that a configured `window_size` of 0 still advances (doubling alone leaves it at 0 forever, turning a bounded retry into a hang).

## Measurement

`examples/dump_normalized_corpus.rs`, 85,642 rows, both revisions: the two dumps are **byte-identical**.

**That zero is structural and must not be read as safety.** The property this change keys on is *a repeat tract longer than `window_size` (100 bases) that the shuffle can roll through*, and the corpus cannot build one. Measured directly rather than argued: **the largest first-position shift anywhere in the 85,642-row corpus is 14 bases.** The changed code path needs a shuffle to travel at least 100. The corpus misses the trigger by a factor of seven, so it could not have moved a row whatever this change did. The cap-refusal path is further out of reach still — it needs a tract past 64 KiB, and no corpus member's whole sequence is that long.

Real consumer impact is therefore *not* zero, and is not quantified here: any `g.` or `m.` description sitting in a repeat tract longer than 100 bases moves — to the tract's 3' end, which is the position the 3' rule asked for all along, and which the previous output was only an arbitrary way-point toward. Past 64 KiB, a description that was stored as an `ins` or a `delins` may now be stored as the `dup`/`sub`/`inv` the spec requires.

Representation-Change: two classes move. (1) A `g.` or `m.` description whose 3' shuffle runs past `window_size` (100 bases) now converges on the tract's 3' end instead of advancing 100 bases per normalize call. (2) Past the 64 KiB growth cap the shift is refused but the reference-aware edit type is no longer refused with it, so a description that is really a duplication, substitution or inversion is now spelled as one — `g.65_66insA` -> `g.67dup`, `g.65delinsG` -> `g.65A>G` — where before it was echoed in the authored edit type. Repeat notation stays refused past the cap, because its `<first>_<last>unit[N]` extent is unreadable there. Measured 0 of 85,642 corpus rows moved, but that zero is STRUCTURAL and is not evidence of safety — the corpus's largest shift is 14 bases and no member is 64 KiB long, so it cannot reach either changed path.

### DIFF
--- a/src/normalize/boundary.rs
+++ b/src/normalize/boundary.rs
@@ -31,6 +31,28 @@ impl Boundaries {
         Self { left, right }
     }
 
+    /// Bounds that permit no shuffle travel at all, in either direction.
+    ///
+    /// [`crate::normalize::shuffle::shuffle`] walks 3' while
+    /// `new_end < right` and 5' while `new_start > left`, so a pair that is
+    /// empty in *both* senses — `left` above every position, `right` below every
+    /// one — makes each loop fall through on its first test whatever coordinates
+    /// the edit has. That is the point: the alternative is to derive a pinned
+    /// pair from the edit's own indices, which would have to reproduce
+    /// `normalize_na_edit`'s per-edit-type index adjustment (an insertion's
+    /// shuffle start and end are both shifted by one) and could only ever be
+    /// wrong in the same way twice.
+    ///
+    /// Used by `Normalizer::canonicalize_without_shifting` to re-type an edit
+    /// against the reference while refusing to move it — see that method for why
+    /// the two halves are separable.
+    pub fn pinned() -> Self {
+        Self {
+            left: u64::MAX,
+            right: 0,
+        }
+    }
+
     /// Check if a position is within the half-open bounds `[left, right)`.
     /// `right` itself is **not** contained — it's one past the last
     /// reachable position by the half-open convention documented on
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -4124,6 +4124,94 @@ fn resolve_special_genome_pos<P: ReferenceProvider>(
     }
 }
 
+/// Widest **half**-width [`Normalizer::normalize_in_grown_window`] will grow the
+/// fetch to before refusing to shift any further.
+///
+/// **This bounds `window`, not the span read.** `normalize_in_grown_window`
+/// fetches `[start - window, end + window]`, so the widest reference span is
+/// roughly *twice* this plus the variant's own length — measured, a 60,000-base
+/// tract at contig offset 100,000 converges to `g.160000del` off a window
+/// covering `[34,465 .. 165,537]`, 131,073 bases. The *travel* bound is the
+/// half-width itself: a 65,000-base tract converges and a 66,000-base one is
+/// refused. [`MAX_INTRONIC_SHUFFLE_WINDOW`] at its own definition is tested as a
+/// **full** span, so the two constants are numerically equal and semantically a
+/// factor of two apart; do not read one off the other.
+///
+/// The value is **reused from** [`MAX_INTRONIC_SHUFFLE_WINDOW`] rather than
+/// independently sourced — that constant is this module's existing answer to the
+/// same question (how far a shuffle may fetch before the cost stops being worth
+/// the answer) and is itself unsourced. What justifies the reuse is that a run
+/// long enough to exhaust it is a structural feature (an assembly `N` gap, a
+/// synthetic contig) rather than a repeat tract: the pathogenic expansions that
+/// do run long — FMR1 `CGG`, RFC1 `AAGGG` — reach thousands of bases, which is
+/// still an order of magnitude inside 64 KiB. It is a cost bound picked to sit
+/// clear of the biology, not a bound derived from it.
+const MAX_SHUFFLE_FETCH_WINDOW: u64 = MAX_INTRONIC_SHUFFLE_WINDOW;
+
+/// Next window to try, or `None` once [`MAX_SHUFFLE_FETCH_WINDOW`] has been
+/// tried. Doubling keeps the number of re-fetches logarithmic in the distance
+/// the shuffle actually travels, so the overwhelmingly common case — a result
+/// that rests well inside the configured window — costs exactly one fetch and
+/// never consults this function at all.
+///
+/// `max(1)` guards a `window_size` of 0, which would otherwise double to itself
+/// forever.
+fn grow_shuffle_fetch_window(window: u64) -> Option<u64> {
+    if window >= MAX_SHUFFLE_FETCH_WINDOW {
+        return None;
+    }
+    Some(
+        window
+            .max(1)
+            .saturating_mul(2)
+            .min(MAX_SHUFFLE_FETCH_WINDOW),
+    )
+}
+
+/// Bases of reference kept either side of an edit's own span by
+/// [`Normalizer::canonicalize_without_shifting`].
+///
+/// Every type rule that survives the growth cap is decided within one payload
+/// length of the edit: `duplication.md:5` defines a duplication as a copy
+/// inserted "**directly 3'** of the original copy", so seeing the original costs
+/// exactly `|payload|` bases of 5' flank, and the mirror check on the 3' side
+/// costs the same. Delins trimming, the delins type ladder and the inversion
+/// shortening all read inside the span itself.
+///
+/// **The value must depend on the edit and on nothing else.** It is what bounds
+/// how far the capped answer can look, so a flank derived from the reference —
+/// a tract length, a window width — would make the answer depend on the very
+/// distance the cap declined to travel, and the walk would come straight back.
+/// The `max(1)` floor keeps a zero-payload edit (`del`, `inv`) from being handed
+/// a slice with no context at all.
+fn capped_typing_flank(edit: &NaEdit) -> u64 {
+    let payload = match edit {
+        NaEdit::Insertion { sequence } | NaEdit::Delins { sequence, .. } => {
+            sequence.bases().map_or(0, <[Base]>::len)
+        }
+        _ => 0,
+    };
+    (payload as u64).max(1)
+}
+
+/// One settled attempt at normalizing an edit against a fetched reference
+/// window. See [`Normalizer::normalize_in_grown_window`].
+struct WindowedNormalization {
+    /// 0-based offset of the window's first base within the contig. Positions
+    /// convert back with `contig = window_relative + window_start`.
+    window_start: u64,
+    /// The fetched bases.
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

PR #1705, opened 2026-08-12
fix(ci): derive the reporter's provenance sentence from the triggering event

nh13:
`report-failure.yml` ended every tracking issue it files with a fixed sentence: that the workflow does not run on pull requests, and that the failure therefore could not have been caught before merge. That is a claim about the **caller**, hardcoded in the **callee**, so it cannot notice when a caller's triggers change.

It went stale on arrival. #1225 both created this reporter — because `Coverage` stayed red for 15 commits while every PR reported a green board — and gave `coverage.yml` a `pull_request` trigger. From that commit the sentence was false for one of the six callers.

## Measured

Six workflows call the reporter. Exactly one runs on pull requests:

| caller | `pull_request` trigger |
|---|---|
| `coverage.yml` | **yes** |
| `audit-pr-runs.yml`, `external-validation.yml`, `fuzz.yml`, `nightly-perf.yml`, `nightly-mutalyzer.yml` | no |

So the sentence was true of five callers and false of one — and `coverage.yml`'s own header says why it should be: *"Runs on pull requests as well as on `main`. Until #1225 this workflow was push-only, which made it structurally incapable of failing before a merge."*

All five recurrences on the rolling coverage issue carried the false sentence, most recently against a superseded head of an open PR — where the failure was not merely catchable but already visible on the PR.

## The direction of the error is why this is a fix rather than a deletion

It excused a failure that *was* visible on the pull request, and pointed the reader away from the one place the answer was sitting. A wrong claim that discourages checking is worse than no claim.

## The change

The sentence is derived from `github.event_name`, which in a `workflow_call` is the caller's own triggering event. On a pull request it names the PR and says to look there; otherwise it names the actual event. That is a fact about *this run*, so adding a `pull_request` trigger to another caller needs no change here.

Both new reads are `${VAR:-}` rather than bare, and that is load-bearing rather than defensive habit. The step runs under `set -u`, so an unset `EVENT` aborts it — and aborting here files **nothing**, which is the exact failure `report_failure_degrades` exists to prevent. That test caught it, by executing the `run:` block standalone with a stubbed `gh`; it is unreachable from real CI, where `env:` always supplies both, so nothing else would have said so.

## The guard

`tests/it/report_failure_provenance.rs`, in the idiom of `sweep_filter_invariant.rs` and `coderabbit_config_paths.rs`. Four tests, two of which exist to stop the other two passing vacuously:

- the reporter has callers at all — otherwise every assertion below is empty
- at least one caller runs on `pull_request` — because if none did, the old sentence would be **true**, and this guard should be reconsidered deliberately rather than left green for the wrong reason
- the emitted text carries no blanket uncatchable-before-merge claim
- the reporter keys on `github.event_name`, and compares it against `pull_request`, so both branches are reachable

**The scan reads the file with comment lines removed.** The question is what the workflow *emits*, and a comment emits nothing — which also lets the fix quote the offending sentence to explain why it was removed, without failing itself. That is the `SELF`-exclusion problem `sweep_filter_invariant.rs` records, in a different costume.

**Proven to discriminate.** With the blanket claim restored to the emitted body, `the_reporter_does_not_assert_that_failures_were_uncatchable` fails; restored via `cp` and verified byte-identical.

## Verification

`FMT=0`, `CLIPPY=0` (`--all-features --all-targets -D warnings`), `actionlint` clean, YAML parses. 7/7 in the two `report_failure` modules — my 4 new plus the 3 existing, which were red until the `${VAR:-}` fix.

Representation-Change: none. CI and tests only; no file under `src/` is touched.

### DIFF
--- a/.github/workflows/report-failure.yml
+++ b/.github/workflows/report-failure.yml
@@ -1,4 +1,5 @@
-# Reusable failure reporter for workflows that cannot fail on a pull request.
+# Reusable failure reporter for workflows whose red X would otherwise be
+# decoration.
 #
 # A workflow that only ever runs after a merge — on a schedule, or on push to
 # `main` — has no way to stop the break it detects. Its red X is decoration
@@ -7,6 +8,14 @@
 # demonstrated the consequence: it stayed red for 15 commits (`92da6d0` ..
 # `e90dff2`) while every pull request in that window reported a full green
 # board (#1225, #1226).
+#
+# NOT every caller is post-merge-only, and assuming so is what this file got
+# wrong. `coverage.yml` runs on `pull_request` as well as on `main` — added by
+# #1225, the same change that motivated this reporter — so one of the six
+# callers CAN fail before a merge. The issue body therefore derives its
+# provenance sentence from `github.event_name` rather than stating a fixed
+# claim; see the comment on that branch below. Adding a `pull_request` trigger
+# to another caller needs no change here, which is the point.
 #
 # `nightly-soak` already had the right pattern and this is a straight
 # generalisation of it: ONE rolling issue per workflow, updated rather than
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1706, opened 2026-08-12
feat(normalize): recognise inversions post hoc over blocks and runs of pieces

### LINKED ISSUE #1575: normalize: an authored inv is repartitioned into members including a delins, which inversion-vs-two-delins-76-83 decided against
## An authored `inv` is repartitioned into members that include a `delins`

`rulings[inversion-vs-two-delins-76-83]` is `decided`, and it decided this shape:

> `inversion.md:5` governs: a whole-block reverse complement is one `inv` when the members it competes with are `delins`, since `general.md:56` ranks substitution but not `delins`. #1230's substitution case is untouched and still splits

After #1484 widened the sequence-derived axis gate to `c.`/`n.`/`r.`, **25 authored inversions on `NM_004006.2` are repartitioned into member sets that contain a `delins`** — the members the ruling says lose to the whole-block `inv`.

## Measurement

Swept `NM_004006.2:c.{start}_{start+len-1}inv` for every `start` in `101, 108, …, 2999` and every `len` in `{4, 6, 8, 12, 16}` — 2,075 rows, against the prepared reference. 155 repartition, and they fall into three shapes:

| members | rows | against the ruling |
|---|---|---|
| substitutions only | 87 | consistent — the #1230 case it says still splits |
| an `inv` core with flanking substitutions | 43 | not addressed |
| **include a `delins`** | **25** | **the shape `inversion.md:5` is recorded as governing** |

All 25:

```
NM_004006.2:c.157_168inv	NM_004006.2:c.[157_158delinsGC;160C>T;165A>G;167_168delinsAG]
NM_004006.2:c.367_374inv	NM_004006.2:c.[367_369delinsTTC;372G>T;374A>C]
NM_004006.2:c.374_381inv	NM_004006.2:c.[374A>G;376A>T;379_381delinsTTT]
NM_004006.2:c.451_466inv	NM_004006.2:c.[451T>C;454_455delinsTA;458_459inv;462_463delinsGG;466G>A]
NM_004006.2:c.661_676inv	NM_004006.2:c.[661_663delinsTCT;668_669inv;674A>G;676A>T]
NM_004006.2:c.787_802inv	NM_004006.2:c.[787_789delinsACT;793C>A;796T>G;800_802delinsTTC]
NM_004006.2:c.878_889inv	NM_004006.2:c.[878A>T;880_881delinsAA;886_887delinsGG;889A>T]
NM_004006.2:c.913_920inv	NM_004006.2:c.[913_915delinsACA;918T>A;920T>C]
NM_004006.2:c.941_948inv	NM_004006.2:c.[941G>T;943A>G;946_948delinsTCC]
NM_004006.2:c.1067_1078inv	NM_004006.2:c.[1067_1068delinsCC;1071T>A;1074T>A;1077_1078delinsAA]
NM_004006.2:c.1095_1110inv	NM_004006.2:c.[1096_1097delinsTT;1101G>A;1104T>C;1108_1109delinsCC]
NM_004006.2:c.1291_1302inv	NM_004006.2:c.[1291_1292delinsGA;1295A>C;1298G>T;1301_1302delinsCA]
NM_004006.2:c.1655_1670inv	NM_004006.2:c.[1655T>A;1657_1659delinsGTC;1662_1663inv;1666_1668delinsAAG;1670T>A]
NM_004006.2:c.1767_1782inv	NM_004006.2:c.[1767C>A;1771_1772delinsTG;1774_1775inv;1777_1778delinsTT;1782T>G]
NM_004006.2:c.1781_1792inv	NM_004006.2:c.[1783_1785delinsAAC;1788G>T;1790T>C]
NM_004006.2:c.1816_1831inv	NM_004006.2:c.[1816T>C;1819_1822delinsGATC;1825_1826delinsCT;1828C>T;1831G>A]
NM_004006.2:c.1921_1928inv	NM_004006.2:c.[1921_1923delinsCAT;1926A>T;1928G>C]
NM_004006.2:c.1921_1932inv	NM_004006.2:c.[1921_1924delinsCAGC;1929_1930delinsCT;1932G>C]
NM_004006.2:c.1963_1978inv	NM_004006.2:c.[1963_1965delinsTTT;1969_1972inv;1976A>G;1978A>C]
NM_004006.2:c.2390_2405inv	NM_004006.2:c.[2390A>T;2393_2394delinsAT;2396A>C;2399G>T;2401_2402delinsTG;2405A>T]
NM_004006.2:c.2418_2433inv	NM_004006.2:c.[2418A>C;2420_2421delinsGG;2425_2426inv;2430_2431delinsTT;2433G>T]
NM_004006.2:c.2432_2443inv	NM_004006.2:c.[2432_2433delinsAT;2435G>C;2440_2443delinsCACC]
NM_004006.2:c.2516_2531inv	NM_004006.2:c.[2517A>G;2519_2520delinsTC;2527_2528delinsTT;2530C>T]
NM_004006.2:c.2628_2643inv	NM_004006.2:c.[2628A>T;2630_2631delinsAT;2633A>G;2638_2641delinsTTGA;2643A>T]
NM_004006.2:c.2656_2671inv	NM_004006.2:c.[2656_2657delinsTT;2661T>C;2663_2664inv;2666G>A;2670_2671delinsTG]
```

## Not a sequence defect

Every one of these denotes the correct bases: applying the members to the committed transcript sequence reproduces the authored span replaced by its reverse complement, checked per row in `the_members_of_every_repartitioned_row_reproduce_the_inversion` (#1554). So this is a question about which of several correct spellings ships — the `canonical-form-choice-when-both-legal` family — not about correctness.

## What is actually being asked

One of these, and the answer belongs in the ruling record rather than in a test:

1. The ruling's scope is narrower than its wording, and a `delins` member arising from the *sequence-derived partition* is outside it. Then the record should say so, since as written it reads as covering this.
2. The ruling holds and #1484 widened past it for this shape, in which case a whole-block `inv` should win wherever the competing partition contains a `delins`.

Either way it is a representation decision: {len(delins)} of 2,075 rows on one transcript, and the shape generalises to every authored inversion whose block admits a `delins` partition.

Pinned as characterization in `tests/it/inversion_sweep.rs` meanwhile — recorded, not endorsed.

nh13:
Closes #1575

Representation-Change: 62 of 2,075 rows move. An authored inversion that was shredded into an allele comes back as the single `inv`; previously-accepted inputs, so a real migration for a consumer keying on the string. Denotation is unchanged on all 62, verified through the SPDI applier with a firing negative control. Re-derived by set difference against `origin/main` at `cc8407bc`, not carried forward from any earlier measurement: the same 62 rows. Denotation is re-stated from the `1ea75334` run rather than re-derived here; see the table below, which marks which figures were re-run at this base and which were not.

## The defect

`coalesce_whole_block_inversion` asks exactly one question — *do **all** the pieces together span one reverse complement?* — so an inversion reaches the output only when it is the whole block **and** the block's piece geometry happens to line up with it. Two things break that, and both were measured rather than argued.

**The relation was read off the wrong span.** Whether a block is an inversion is a property of the `(reference block, resulting block)` pair and of nothing else, but the test was applied to the **pieces' hull**. By the time this pass runs, `shift_pieces` has moved every pure indel 3', so the hull is routinely a base or two off the block — and `revcomp` does not survive that, because `revcomp(x)[k..]` is `revcomp(x[..len-k])`, not `revcomp(x[k..])`. A hull off by one base asks a different question and answers no.

**The admission gate refused a competitor no clause selects.** `no_piece_is_a_lone_substitution` withdraws the spanning `inv` unless ***no*** competing member is a lone substitution, which also refuses a **mixed** competitor — substitutions beside multi-column members. `NC_000013.10:g.100809575_100810031inv` is a perfect 457-base reverse complement whose canonical partition is 133 pieces, one of them `g.100809575A>T`; that partition leaves 271 of its 457 columns unchanged (re-measured at `cc8407bc`; the `229` measured at `3d2a7a03` does not reproduce), so both density routes read the wrong side of half and the block came out as its own columns. Refusing the `inv` there makes the description turn on whether one interior column happens to be self-complementary — a property of the sequence, not of the event.

**This PR's earlier prose grounded that on `general.md:56`, and that grounding is withdrawn** (see the ruling record below). Read plainly and taken to reach a multi-member allele at all, `:56` argues the *other* way on every row this PR moves: every competitor contains a substitution, which is `:56`'s rank (1), above inversion at (3). `:56` supports neither the widened gate nor the narrow one it replaces, and `conflicting-member-geometry-refusal-scope` is already on record that `:56` does not reach a multi-member allele. `general.md:58` was measured as an alternative ground and does not reach it either — no competitor on any moved row contains a `del`, a `dup` or an `ins`.

## The mechanism

`coalesce_inversion_runs` replaces the single call to `coalesce_whole_block_inversion` at the same point in `canonicalize_from_sequence` — **after** the pieces are settled and shifted, **before** `rebuild_members` renders them. It is a post-hoc pass over the partition, not a change to the alignment algebra, and it runs four routes in order:

| route | question | scope |
|---|---|---|
| **0** | `coalesce_whole_block_inversion`, verbatim | unchanged; if it fires, nothing else runs |
| **A** | does the **block** equal `revcomp` of its own span? | reads `(ref_block, result_block)` directly, so a 3'-shifted terminal indel cannot hide it |
| **B** | is the payload `revcomp` of an **interior** sub-span, flanks falling out as deletions? | single-piece blocks only |
| **C** | does some **run** of consecutive pieces spell an inversion over its own hull? | ≤ `INVERSION_RUN_MAX_PIECES` pieces per window |

Route 0 running first and returning early is what makes the pass additive: a block that coalesces today still coalesces by exactly the path it did.

**Every admission is exact equality against a reverse complement.** No density heuristic, no new threshold. The majority condition on the flanked shape is the same `changed * 2 > span` its two siblings already apply, restated on the payload. `INVERSION_RUN_MAX_PIECES = 256` is a *work* bound on route C alone — the block-level routes read the block and are `O(span)` whatever the piece count, pinned by `the_window_bound_does_not_reach_the_block_level_routes`, which recognises a 700-piece inversion with the bound at 256. Its stated source is one measured point (`NG_009874.2:g.158330_159052` partitions into 193 pieces on the canonical arm) and the next power of two above it; it is not derived from a piece-count distribution and the doc comment says so.

## The adjudication, and where it is recorded

The gate's one new route, `not_every_piece_is_a_lone_substitution`, is `no_piece_is_a_lone_substitution` with `all` weakened to `any`. **This is the adjudication in the PR**, and it is the one place the change takes a position rather than measuring one. It is now filed as a ledger record, `inversion-vs-a-mixed-member-competitor` in `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json`, decided by operator ruling on 2026-08-12.

The record's shape matters and is stated in it rather than paraphrased here. `DNA/inversion.md:5` governs **as a permission**: the `inv` is conformant, the mixed multi-member competitor is conformant too, so the choice between them is a `README.md` rule 6 choice among conformant forms — not a conformance claim, and a build emitting the multi-member form is not violating rule 1. What licenses the `inv` is that notation must not turn on base coincidence, not a ranking of the competitor's member types.

Route attribution was measured, not inferred. Instrumented at the three adoption points at `3d2a7a03`, all 62 moved rows adopt exactly once, at the block route, with only this fifth predicate admitting — routes 1–4 admit none of them. **Re-measured against `origin/main` at `cc8407bc`**, by disabling only this route and re-running the whole sweep: withdrawing it does **not** restore the previous behaviour — movement drops from 62 rows to **10**, and **all ten** land on a third spelling present in neither arm (`c.283_298inv` → `c.[283G>C;286_295inv;298G>C]`, byte-identical to the earlier measurement). **This counterfactual moved with the base** (12 / 3-via-run-scan / 9-third-spelling at `1ea75334` and `c165c939`): `b455a637` (#1720) touches `src/normalize/merge.rs`, and the three rows that used to reach the shipping answer through the run scan now fall back to `main`'s own spelling instead, while `c.2418_2433inv` starts landing on a third spelling. The durable point is strengthened, not weakened — none of the ten is a restoration. Withdrawal would be a new representation change, not a restoration. The per-adoption-point bitmask attribution in the first sentence is the one figure still carrying only its `3d2a7a03` stamp; the withdrawal counterfactual, which is what the ruling turns on, is re-derived at the new base.

#1230's `GATG -> CATC` and #1040's `AAGCTA -> TAGCTT` — where *every* competing member is a one-base substitution — stay split, pinned at every arity by three tests in the new falsification suite.

**Two further `:56` claims survived the four corrections and are corrected here too.** Both are the same shape of failure — a withdrawn claim living on in a second copy — which is the one this repository records as recurring. The inline comment at the `inversion_gate_admits` call site still read *"the competing description is what `general.md:56` ranks against, and that is a statement about members"*, eighty lines above the doc comment that withdraws exactly that. And `not_every_piece_is_a_lone_substitution` still gave its own reason as `:56` stating an argument "more broadly than `:56` says it", and still described the question as a contest it inherits and does not settle — which the ledger record settles, and settles without resting on `:56`. Grepped rather than eyeballed: the `:56` mentions that remain in this diff are the corrections themselves, the `inversion-vs-two-delins-76-83` restatements (that ruling *does* rest on `:56`, and is undisturbed here), and the #1040/#1230 control messages that restate it. The record itself cites `:56` only to record that it argues the other way.

**And `inversion_gate_admits`'s doc now names the predicate that actually executes**, which was the substantive half of that correction. Two of the gate's five disjuncts can never be the deciding one: the third implies the fifth (already recorded), and — not recorded anywhere until now — the second implies the fourth, since both compare `changed * 2 > span` over the same span while the fourth sums `max(reference width, payload length)`, which is never smaller. So the gate is three independent questions, not five: are the pieces close together, does the *payload* dominate the span, is some piece not a lone substitution. Both implications were verified exhaustively over 486 piece sets, with the converses shown to fail so the check is not vacuous. No behaviour changes — the redundant disjuncts are kept so the four carried-over routes stay listed in `whole_block_inversion`'s order.

**Why post-hoc rather than in the alignment algebra.** Inversion is not an edit the aligner can price: it is a relation between a span and its own reverse complement, and admitting it as a DP edge changes the cost model for every block, not just the inverted ones. Recognising it after the partition is settled keeps the change confined to blocks that are *already* an exact reverse complement, which is why the blast radius below is what it is. The cost-invariant alternative — admitting inversion as an edge in the alignment itself — is architecturally stronger and is being proposed separately at `INV_COST = 3`. This PR is not an argument against it.

## What it does not fix

- **`coalesce_whole_block_inversion` remains**, and this PR is another patch around it rather than a replacement for it. Five issues have been filed against inversion under-recognition in this pass — #160, #1034, #1041, #1517, #1637 — and seven PRs have already touched the function itself (`git log -S 'coalesce_whole_block_inversion' -- src/normalize/merge.rs`: #1392, #1470, #1484, #1485, #1535, #1640, #1644). Retiring the pass outright is the edge-admissibility approach's prize, not this PR's.
- **#1541's two-substitution case is untouched**, deliberately: `NM_000500.9:c.[710T>A;713T>A]` and `c.710_713inv` remain two fixed points, because every competing member there is a lone substitution.
- **93 of the 155 repartitioned sweep rows stay split**, and all 93 are that same case — measured, not assumed: every changed column in all 93 blocks is isolated, so there is no multi-column run for the gate to admit on. Six of them nevertheless *render* with a `delins` member (`c.367_374inv -> c.[367_369delinsTTC;372G>T;374A>C]`); reading that as a counter-example is the trap the census doc comment now names. That `delins` is the coding codon carve-out (`general.md:35`, `DNA/delins.md:18`) merging two isolated substitutions one base apart, applied *after* this pass — the partition the gate reads holds no `delins` at all. **Count these by the pieces, never by the rendered members.**
- **The net-insertion mirror of the flanked shape (`[ins; inv; ins]`) is out of scope**, matching the direction `delins-merge-vs-individual-gap-two-or-more` scopes its own licence to. Pinned as a limit by `a_net_insertion_flanked_inversion_is_out_of_scope`.
- **Route B does not cross a separation the partition already found.** Scoped to a single-piece block, because widening it made `NC_TEST.1:g.[265_268inv;270_273del]` come back as `g.265_273delinsGTCAG` and `five_prime_conformance_census` reported "outputs implementing REJECTED SVD-WG010 rose from 0 to 2".
- **Three questions the ruling record explicitly leaves open**: whether `general.md:56` reaches a multi-member allele (decided elsewhere, negatively — the record relies on that holding without extending it), the frameless separation-one family (#1711), and the coding-axis merge questions (#1716 at separation one, and `coalesce_payload_alignment_split` at separation two or more).

## Measurements, on this base

**`tests/it/inversion_sweep.rs` — 2,075 authored `NM_004006.2:c.…inv` rows, prepared reference.** 62 rows move. The outcome census goes from `(1434 unchanged, 460 shifted-still-inv, 26 palindromic-no-op, 155 repartitioned)` to `(1490, 466, 26, 93)`; 56 of the recovered rows come back at the authored span and 6 at a narrowed one. `cases.tsv` and the four census constants are re-pinned in this PR, which is what the three ratchets' own failure messages ask for.

**Denotation, on all 62 moved rows — 62/62 agree, 0 differ, 0 not-comparable.** Checked **non-circularly**, through `spdi::compare_denoted_sequences` on the two *parsed* descriptions rather than by re-normalizing either: the two spellings under test converge on one string, so a check that normalized first would be comparing `x` with `x`. A ±500-base coordinate-shift negative control on the same rows **fires 116/116** (both directions where both stay above `c.1`; 8 of the 124 candidates were skipped because `-500` would go below `c.1`, and every one of the 62 rows still carries at least one firing control).

**`examples/dump_normalized_corpus.rs` — 0 of 85,642 rows move.** That zero is **structural for the population this change moves**, and should not be quoted as a safety result. The corpus does build inversion families, and the reason each is inert is checkable:

- `separated_revcomp_runs` (270 rows) pins `REVCOMP_RUN_WIDTH = 2` — *every* changed run is multi-column — so `no_piece_is_a_lone_substitution` already admits them and `main` already coalesces them to `inv`.
- `long_block_inversion` (16 rows) — both designed cores shred into exactly **four isolated one-base substitutions** (`NC_TEST.1:g.257_1280inv -> g.[267C>G;761G>C;776G>C;1270C>G]`), which is #1040's case and is refused by design.
- No family builds the shape that actually moves: a reverse-complement block whose competing partition is **mixed** — at least one multi-column run *and* at least one isolated column. That is precisely the gap between `REVCOMP_RUN_WIDTH = 2` and `long_block_inversion`'s width-1 runs, and it is what the 62 sweep rows are.

**ClinVar, 500,004 rows through both arms against the prepared reference — 0 move.** The inversion sub-population is **131 inputs**. Of these, **107 normalize** and every one of those does produce a single lone `inv` on both arms — **100 return their own string** and 7 change, 6 dropping a redundant payload (`c.2731_2732invGT` → `c.2731_2732inv`) and one shifting a base (`g.29447680_42554394inv` → `g.29447681_42554393inv`). The remaining **24 decline on both arms**, all with `Intronic variant normalization not supported`; the earlier `124 return their own string` folded those 24 in with the fixed points, which the claim "every one already normalizes" cannot survive. Either way this corpus does not carry the shape. Reported as a real zero with its denominator, not as coverage.

That figure read **125** until the rebase, and the correction is worth naming because of where the wrong number came from: 125 is `EXON_JUNCTION_CROSSING`, the count of sweep rows whose `c.` span crosses an exon junction, which sits a few paragraphs away in the same module. It was never a property of ClinVar at all. Measured now on every criterion available — substring, terminal `inv` edit, `valid` flag, unique inputs — the answer is 131 each time.

**Seam oracles.** `inversion_sweep` runs green under all four (`FERRO_ASSERT_SEQUENCE`, `FERRO_ASSERT_IDEMPOTENT`, `FERRO_ASSERT_IN_BOUNDS`, `FERRO_ASSERT_REPARSE`) apart from the three ratchets this PR re-pins.

**Re-derived again against `origin/main` at `cc8407bc`.** The `1ea75334` stamp this section used to carry was already two bases stale when it was re-read: the branch had been rebased onto `57f71d31` without the stamp moving, and `main` has since reached `cc8407bc`. Across `57f71d31..cc8407bc`, `main` gained thirteen commits, three touching normalization — `0c925605`/#1751, `f187331a`/#1726 and `b455a637`/#1720, the last on `src/normalize/merge.rs` itself. **None moves a row of this sweep on the `main` arm** (its committed `cases.tsv` is byte-identical across the whole range), but #1720 does move the withdrawal counterfactual, which is why that figure is re-derived above rather than carried. The #1708 overlap claim is retired rather than restated: #1708 now sits far below the base and `main` moves no sweep row across the range.

Every figure holds, and each was taken as a **set difference over both arms**, never by comparing totals — a clean textual merge is exactly how a stale count survives:

| figure | at `1ea75334` | re-derived at `cc8407bc` |
|---|---|---|
| sweep rows moved | 62 of 2,075 | **62 of 2,075, the same 62** (set equality) |
| outcome census | (1434, 460, 26, 155) → (1490, 466, 26, 93) | unchanged |
| recovered at the authored / a narrowed span | 56 / 6 | unchanged |
| repartitioned rows remaining, of which rendering a `delins` | 93 / 6 | unchanged |
| withdrawal arm: rows moved / on a third spelling / via the run scan | 12 / 9 / 3 | **10 / 10 / 0 — MOVED, see above** |
| rows the fifth route decides | 59 | **62** |
| competitors containing a substitution / a `delins` / an `inv` member | 59 / 18 / 48 (of 59) | **62 / 19 / 50 (of 62)** |
| run-scan adoption events, route present / withdrawn | 0 / 31 | unchanged (count the first normalize pass only) |
| `dump_normalized_corpus` | 0 of 85,642 | unchanged |
| ClinVar | 0 of 500,004 | unchanged |
| ClinVar inversion sub-population | 131 | unchanged; 100 return their own string, 7 change, **24 decline** |
| #1461: canonical pieces / span | 133 / 457 | unchanged; all four earlier routes still refuse |
| #1461: columns left unchanged / widest piece separation | 229 / 5 | **271 / 6 — the old margins do not reproduce** |
| denotation on the moved rows, and its negative control | 62/62 agree, control 116/116 | **not re-derived at this base** |
| per-adoption-point bitmask attribution | `00001`, all 62 at the block route | **not re-derived at this base** |

The `origin/main` side is a measurement rather than a committed pin read back: `cc8407bc` was checked out into its own worktree and its sweep run there — green on all 11 gate tests — and it reproduces its committed `cases.tsv` for all 2,075 rows. On this branch, `extract_inversion_sweep_windows --check` against the prepared reference reports both committed artifacts up to date, and the instrumented shipping arm rewrites `cases.tsv` byte-identically, so neither probe ran a stale binary. The qualitative findings the ruling rests on are shape properties and do not depend on any of it.

## Tests

A falsification suite (`mod inversion_run_scan`, 14 `#[test]` functions — this said 13) written to break the pass rather than demonstrate it. The refusals are as load-bearing as the recognitions: the #1040 and #1230 controls at every arity, an all-substitution competitor at arity three, an ambiguous flank split (three equally-compliant placements — refused, not picked), a minority payload, the net-insertion mirror, a nested inversion pair, and the window bound not reaching the block-level routes. `a_palindromic_span_is_structurally_unreachable` records a **structural** zero as structural: `trim_common_flanks` removes a self-reverse-complementary block before this pass can see it, so the case proves nothing about palindrome handling.


nh13:
@coderabbitai pause

nh13:
On whether `fc311f30` invalidates the trailer's 62 and the four census constants — it does not, and the confirmation predates the question.

The reasoning behind the concern is right: `tests/fixtures/inversion-sweep/cases.tsv` was last written by `98611e94`, the widening at `fc311f30` is behaviour-bearing at `whole_block_inversion`'s four-route gate, and a strictly-widening gate can only move a count up. So either the widening moves no row inside that corpus, or the fixture and all four constants are stale and 62 is a floor.

It moves no row. The full suite run that produced the sign-off for `fc311f30` was executed **after** the `merge.rs` edits — the only change between it and the commit was a `rustfmt` wrapping nit in `tests/it/issue_1517_inv_priority_over_delins.rs`, which touches no behaviour. Every sweep ratchet ran and passed in it, and none was skipped:

```
every_authored_inversion_produces_its_pinned_output              PASS
the_outcome_census_is_unchanged                                  PASS
every_pinned_output_is_a_fixed_point                             PASS
the_members_of_every_repartitioned_row_reproduce_the_inversion   PASS
no_authored_inversion_leaves_the_inversion_family                PASS
```

`10084 tests run: 10084 passed, 151 skipped`, `0` skipped among `inversion_sweep`. `cargo fmt --check` and `cargo clippy --features dev --all-targets -- -D warnings` both clean.

Why the widening is inert on this corpus is worth stating rather than leaving as luck. The corrected predicate differs from the width test it replaced only on pieces that are **narrow but not substitutions** — a 1-base deletion, a pure insertion, a 1-base `delins`. The sweep's 2,075 rows are authored inversions and their repartitions, whose members are substitutions and multi-base runs; the shapes that discriminate the two predicates are the ones `partition_block_canonical` produces from a *whole-span reverse complement with self-complementary interior columns*, which is #1703's geometry and not what this corpus enumerates. So the zero is structural rather than incidental — the corpus cannot build the shape the change keys on.

That is the same class of claim the repository `CLAUDE.md` warns about ("a corpus zero is a claim about the corpus, not about the change"), so stating it that way rather than as reassurance: the sweep confirms no *authored inversion* moved, and the new guard in `issue_1517_inv_priority_over_delins.rs` is what covers the shape that does.

nh13:
Coordination note, no action implied for this PR's correctness.

A second ledger record touching inversion typing is being drafted in parallel, so whoever lands second will need to **reconcile the records**, not just fix the textual conflict in `tests/it/clause_ruling_index.rs`.

**What this PR carries:** `inversion-vs-a-mixed-member-competitor`, which is not on `main`.

**What is being drafted separately:** a record for the operator ruling on #1703 (nh13, 2026-08-13) —

> Overturn #1230. A whole-span reverse complement types as `inv` **uniformly** — regardless of how much of the interior coincides with the reference, and regardless of whether the competing partition is substitutions, `delins`, or anything else.

It widens `inversion-vs-two-delins-76-83` (currently scoped to `c.76_83inv`) and records that #1230 is overturned. Because that formulation is deliberately general, it may **subsume** `inversion-vs-a-mixed-member-competitor` rather than sit beside it.

Both records also edit the generated clause index and its census pins, so the collision is in two places.

Worth flagging as well: the 457-base `NC_000013.10:g.100809575_100810031inv` this PR reports as a 133-piece partition is the same row #1703 names as its one genuine rule 1 violation — 12 pairs of members on consecutive nucleotides, which `DNA/substitution.md:32` marks `class="invalid"` by name. So this PR's implementation and that ruling are closing the same defect from two directions, which is the main reason the two records should end up as one coherent story rather than two overlapping ones.

PR #1707, opened 2026-08-12
fix(normalize): settle whole-span inversion typing before the block is cut

### LINKED ISSUE #1703: canonical-coalesced re-derives a whole-span inversion as [del;dup]; all spellings converge on the wrong form
## Summary

On the `canonical-coalesced` partition arm, a whole-span reverse complement is re-derived as a `[del;dup]` pair instead of an `inv`. Every spelling converges on the same wrong string, so the defect is invisible to a confluence check — the arm is self-consistent and wrong together.

This blocks the pending default change (#148). The shipped `live` arm is correct and unaffected.

## Reproducer, on real GRCh38 coordinates

`NC_000017.11:g.43044327_43044331` is `GTTAA`; `revcomp(GTTAA)` = `TTAAC`. Columns `43044328` and `43044330` are self-complementary, so the substitution spelling is three *separated* substitutions.

```
g.        43044327  43044328  43044329  43044330  43044331
ref          G         T         T         A         A
             X         |         X         |         X
alt          T         T         A         A         C
```

| input | `FERRO_PARTITION=live` | `FERRO_PARTITION=canonical-coalesced` |
|---|---|---|
| `g.43044327_43044331inv` | `g.43044327_43044331inv` | `g.[43044327del;43044332dup]` |
| `g.[43044327G>T;43044329T>A;43044331A>C]` | `g.43044327_43044331inv` | `g.[43044327del;43044332dup]` |
| `g.43044327_43044331delinsTTAAC` | `g.43044327_43044331inv` | `g.[43044327del;43044332dup]` |

Mutalyzer converges all three on `NC_000017.11:g.43044327_43044331inv`, matching `live`.

The same shape is already pinned synthetically on the `live` behaviour by `issue_1040_inv_overrecognition_probes::every_spelling_of_a_derived_whole_block_inversion_converges_on_inv`, which uses the identical 5-mer core.

## Why the current output is not admissible

`DNA/inversion.md:5` defines an inversion by the content of the whole span — "**more than one nucleotide** replacing the original sequence is the reverse complement of the original sequence" — and the span here is an exact reverse complement. Independently, `DNA/delins.md:5` defines a delins as a replacement "**and which is not** a substitution or inversion", so the spanning `delins` spelling is excluded by definition; ferro accepts it as input but must not emit it.

## Root cause

`shift_pieces` moves the trailing insertion to `43044332` — outside the block hull — before `coalesce_whole_block_inversion` (`src/normalize/merge.rs:5296`, called at `:3075`) runs. That pass then computes a 6-base hull against a 5-base payload and declines.

## Why this cannot be fixed by a partition-time gate

The block's minimal split is genuinely **unique** — 0 ambiguous columns under every cost model tested, at padding 0 through ±3. Three separate candidate gates keyed on forced columns, pinned columns and matched-set uniqueness were measured against it and all four readings (including ungated) produce byte-identical output. Nothing keyed on the ambiguity of the alignment can reach this row, because there is no ambiguity to key on.

## Fix shape

An ordering constraint rather than a new gate: test the trimmed block for `block == revcomp(payload)` and short-circuit to the `inv` **before any cut is considered**, rather than attempting to reassemble one after `shift_pieces` has moved a member outside the hull.

Inversion typing is a predicate over the whole block; cutting is a decision about where to divide it. The typing has to settle first.

## Scope

The `live` arm is correct on all three spellings, so no shipped output moves. This is a prerequisite for #148 rather than a defect in current behaviour.

nh13:
Closes #1703.

On every sequence-first partition arm (`FERRO_PARTITION=shadow`, `canonical`, `canonical-coalesced`) a whole-span reverse complement was re-derived as a `[del;dup]` pair instead of an `inv`. Every spelling reached the same wrong string, so the arm was self-consistent and wrong together and no confluence check could see it. `NC_000017.11:g.43044327_43044331` is `GTTAA`, `revcomp(GTTAA) = TTAAC`, and columns `43044328`/`43044330` are self-complementary.

`DNA/inversion.md:5` defines an inversion by the content of the whole span, and `DNA/delins.md:5` excludes the spanning `delins` spelling by definition. Both are prohibitions read from prose force, so this is a rule 1 bug rather than a disclosure, and it blocks #148.

## The mechanism is not the one the issue diagnosed

The issue attributes the decline to `shift_pieces` moving the trailing insertion outside the block hull, after which `coalesce_whole_block_inversion` "computes a 6-base hull against a 5-base payload and declines". Instrumented before changing anything, the pass never reaches that reconstruction — it declines at the **admission gate**, identically before and after the shift:

```
partition_block_canonical(GTTAA, TTAAC) = [0..1>""] , [5..5>"C"]
  PRE-shift  gate: sep1=false chg=false nolone=false pay=false  -> None
  shift_pieces moves the insertion 5..5 -> 6..6
  POST-shift gate: sep1=false chg=false nolone=false pay=false  -> None
```

So there are **two independent defects**, and neither change reaches the case on its own.

**The gate.** `no_piece_is_a_lone_substitution` measured reference *width* (`>= 2`) while its name and its `general.md:56` reasoning are about substitution-hood. `DNA/delins.md:15` is explicit — a substitution is one nucleotide replaced by one other — so a 1-base deletion (one replaced by none) and a pure insertion (none replaced by one) are not substitutions, and `:56`'s ranking of substitution above inversion reaches neither. The predicate now tests substitution-hood. This is additive: a piece two or more columns wide can never be a one-for-one replacement, so every partition the width test admitted is still admitted, and an admission still has to survive `is_inversion` on the reconstruction.

**The ordering.** `shift_pieces` runs before `coalesce_whole_block_inversion`, and moves a derived insertion off the hull's right edge whenever the next reference base repeats the payload — which is what renders the wrong answer's second member as a `dup`. The verdict is now taken while the pieces still partition the trimmed block, and applied at the existing call site, which keeps the widening **after** the changed-columns weight bound. That placement is load-bearing in both directions: the bound compares the derived weight against the input's, and `GTTAA -> TTAAC` weighs 3 spelled as three substitutions against 5 spelled as one `delins`, so a pre-bound widening is accepted for one spelling of the variant and refused for the other.

## Which function owns the whole-block predicate

`rules::decompose_delins` does not and must not. Its unit is the **maximal contiguous mismatch run**, bounded by identity positions — and a whole-span reverse complement with self-complementary interior columns has identity positions by construction (coincidences come in mirror pairs). On this block it sees three length-1 runs and types three substitutions. The per-member path answers correctly for a different reason: `rules::canonicalize_delins`'s step-4d full-span check runs first and tests the shared-affix-trimmed span as a whole. The cis re-derivation's counterpart is `merge::whole_block_inversion`, and that is where the predicate belongs — one function per path, each testing the whole span.

## Blast radius

**`live` does not move.** Verified byte-for-byte, not sampled: 819,396 descriptions drawn from the committed ClinVar 500k, CMRG and Paraphase fixtures (every `del`/`ins`/`dup`/`delins`/`inv`/multi-member row, 619,396 of them, plus a 200,000-row random sample of the remainder), normalized against the prepared GRCh38 reference through the merge base and this branch. The whole TSV — including the status and detail columns — is byte-identical in **strict** and in **lenient** mode (lenient leaves only 51 rows unnormalized, so the effective denominator is 819,345). The committed `dump_normalized_corpus` harness likewise reports `0 (0.0%)` moved over 85,642 rows across its 21 shape families, with byte-identical dumps.

That zero is structural rather than lucky: `is_inversion` requires an equal-length reverse complement, so the trimmed block is length-preserving, and `partition_block`'s single-gap search cannot express a length-preserving block with any gap at all — so it never produces the compensating deletion/insertion pair this fix admits.

**The arms move by one row.** Over the same 819,396-row corpus, `FERRO_PARTITION=canonical-coalesced` re-types exactly one row: `NM_001410759.1:c.[854del;857dup]` becomes `NM_001410759.1:c.854_856inv`. So the defect is real-corpus reachable, and the fix is narrow.

## Axes and strand

Probed rather than inferred, on the prepared reference, against a minus-strand gene (BRCA1 `NM_007294.4`, CDS `c.571_575` = `GTTAA`) and a plus-strand gene (CFTR `NM_000492.4`, CDS `c.814_818` = `GTTAA`), three spellings each:

| axis | strand | before (`canonical-coalesced`) | after |
|---|---|---|---|
| `g.` | + | `g.[43044327del;43044332dup]` | `g.43044327_43044331inv` |
| `c.` | − | `c.[571del;575_576insC]` | `c.571_575inv` |
| `n.` | − | `n.[684del;688_689insC]` | `n.684_688inv` |
| `r.` | − | `r.[571del;575_576insc]` | `r.571_575inv` |
| `c.` | + | `c.[814del;818_819insC]` | `c.814_818inv` |
| `n.` | + | `n.[884del;888_889insC]` | `n.884_888inv` |
| `r.` | + | `r.[814del;818_819insc]` | `r.814_818inv` |

All 21 rows were wrong before and all 21 converge on the `inv` after, on all three sequence-first arms; `live` gave the `inv` throughout. The defect was axis-blind and strand-blind, which follows from the pipeline — the partition and the gate read only the trimmed block.

Because the `r.` axis moves too, jurisdiction was checked rather than assumed: a `DNA/` clause cannot scope `r.`, and the RNA recommendations state the same three clauses in their own directory — `RNA/inversion.md:5` and `RNA/delins.md:5` word-for-word identical to their DNA twins, and `RNA/delins.md:16` for the substitution definition. `general.md:56` is not molecule-specific. The table is on `no_piece_is_a_lone_substitution`.

## Tests, written first and watched fail

The end-to-end guard has to reach an arm, and `partition_rule()` caches its `FERRO_PARTITION` read in a `OnceLock`. `tests/it/common/cis_apply_oracle.rs` records why a test must not `set_var` to get around that (sound under nextest, false under `cargo test`), so the arm is reached the way `partition_switch_wiring.rs` reaches it — through the `ferro` binary, with the variable in the child's environment and a temporary indexed FASTA as the reference.

* `tests/it/issue_1703_whole_span_inversion_before_the_cut.rs` — all three spellings converging on the `inv` across all four arms, asserted as the cardinality of the output *set* so convergence is observed rather than three constants agreeing; each output cross-checked through `hgvs_to_spdi` so a re-typing that changed the sequence fails loudly; plus two negative controls and a note recording why all four seam oracles passed on the wrong form.
* `merge.rs` units for each defect: the deletion/insertion pair is not two lone substitutions, two genuine substitutions still are, and the whole-block verdict differs either side of the 3'-shift.

Both negative controls are load-bearing, because the fix widens an admission gate. `AAGCTA -> TAGCTT` is also an exact reverse complement but its members are two lone substitutions, which `general.md:56` ranks above inversion, so it must stay split. `GTTAT -> TTATC` has the *identical geometry* to the positive case — a 1-base deletion at the hull start, a 1-base insertion at the hull end, four unchanged bases between, rendering as `g.[81del;86dup]` — and is not a reverse complement, so it must stay split too. Both were green before the fix and stayed green after; only the three intended assertions were red.

## One assertion re-blessed, deliberately

`the_inversion_coalesce_reaches_a_canonical_partition` asserted `!no_piece_is_a_lone_substitution` for `GCT -> AGC`'s insertion-plus-deletion partition. That verdict moved because the **predicate** was corrected, not because the block changed: the pass reached the same `inv` before and after, by route 4 then and by two routes now. The test's subject is untouched, and `the_inversion_coalesce_reaches_the_spec_published_sixteen_nt_example` still pins route 4 as the one that uniquely decides a partition containing a genuine substitution. The re-bless carries that reasoning inline.

Full suite green: 10,267 passed, 0 failed, with both spec artifacts regenerated. `cargo fmt --check` and `cargo clippy --features dev --all-targets -- -D warnings` clean.

Representation-Change: none. 0 of 819,396 corpus rows move on the shipped `live` arm — the strict and lenient TSVs are byte-identical against the merge base, as is the committed `dump_normalized_corpus` dump at 85,642 rows. `FERRO_PARTITION` unset is `live`, and the partition shape this fix admits cannot arise there. The development-only `shadow`, `canonical` and `canonical-coalesced` arms do change: a whole-span reverse complement is now typed `inv` rather than as a `[del;dup]` pair, which over that same corpus re-types exactly one row (`NM_001410759.1:c.[854del;857dup]` becoming `c.854_856inv`).

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2107,6 +2107,18 @@ impl Piece {
         (self.alt.is_empty() && ref_len > 0) || (ref_len == 0 && !self.alt.is_empty())
     }
 
+    /// Whether this piece spells a substitution: exactly one reference
+    /// nucleotide replaced by exactly one other.
+    ///
+    /// `delins.md:15` — "when **one** nucleotide is replaced by **one** other
+    /// nucleotide, the change is a substitution". Both halves are load-bearing
+    /// and a width test alone gets the second one wrong: a 1-base *deletion* is
+    /// also one column wide and is not a substitution. See
+    /// [`no_piece_is_a_lone_substitution`], which is the reason this exists.
+    fn is_substitution(&self) -> bool {
+        self.ref_end.saturating_sub(self.ref_start) == 1 && self.alt.len() == 1
+    }
+
     /// Whether this piece is a pure insertion sitting immediately outside one
     /// end of the fetched window.
     ///
@@ -2878,6 +2890,37 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
         w_lo,
         &ref_bytes,
     );
+    // Whether this block is one `inv`, decided **here** — while the pieces still
+    // partition the trimmed block — and applied far below, after the weight
+    // bound. Two orderings are in tension and this is what reconciles them.
+    //
+    // `inversion.md:5` defines an inversion by the content of the **whole** span,
+    // so the typing is a predicate over the block; where to cut is a decision
+    // about how to divide it. The typing has to settle first. It does not survive
+    // `shift_pieces`: that pass moves a pure indel 3' as far as the reference
+    // allows, and a derived insertion sitting on the hull's right edge moves
+    // *past* it whenever the next reference base repeats the payload. The pieces
+    // then no longer span the block, and the reconstruction
+    // `whole_block_inversion` builds from their hull is a different, longer
+    // string. On `NC_000017.11:g.43044327_43044331` (`GTTAA -> TTAAC`) the hull
+    // becomes 6 bases and the reconstruction `TTAACC` against `GTTAAC`, which is
+    // not a reverse complement — so the answer computed after the shift is `None`
+    // for a block that is an exact reverse complement (#1703).
+    //
+    // The *application* must nevertheless stay after the changed-columns weight
+    // bound, for the reason `coalesce_whole_block_inversion`'s own doc works
+    // through in full: this rule **widens**, and the bound compares the derived
+    // weight against the *input's*, which differs between two spellings of one
+    // variant. `GTTAA -> TTAAC` weighs 3 spelled as three substitutions and 5
+    // spelled as one spanning `delins`, so a pre-bound widening to 5 columns is
+    // accepted for the second and refused for the first. Deciding here and
+    // applying below is what lets both constraints hold at once; collapsing them
+    // either way reintroduces one of the two defects.
+    //
+    // Nothing between here and the application can make this stale. It is a
+    // function of the trimmed block alone — `ref_bytes[lo..hi_ref]` against the
+    // denoted sequence — and neither is touched again.
+    let block_inversion = whole_block_inversion(&pieces, &ref_bytes);
     shift_pieces(&mut pieces, &ref_bytes, direction);
     coalesce_adjacent_pieces(&mut pieces);
     // Partitioning decides where the members are; this decides how wide each
@@ -3065,7 +3108,22 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
     if partition_rule().cuts_with_canonical() {
         coalesce_compensating_gap_split(&mut pieces, &ref_bytes);
     }
-    coalesce_whole_block_inversion(&mut pieces, &ref_bytes);
+    // The typing decided on the trimmed block wins, and the post-shift pieces are
+    // consulted only when it declined. See the `block_inversion` binding above
+    // for why the decision and its application sit either side of the weight
+    // bound.
+    //
+    // Both directions are kept rather than the earlier one replacing the later:
+    // the pre-shift answer is stronger where a member has since left the hull,
+    // and the post-shift answer is the one that reaches a block whose pieces the
+    // passes between the two calls have merged (`coalesce_payload_alignment_split`
+    // and `coalesce_compensating_gap_split` both re-partition). Neither
+    // subsumes the other, and each is a whole-span reverse complement over the
+    // pieces it is asked about, so preferring the earlier one costs nothing.
+    match block_inversion {
+        Some(whole) => pieces = vec![whole],
+        None => coalesce_whole_block_inversion(&mut pieces, &ref_bytes),
+    }
 
     // Applied *after* the weight bound, deliberately. The bound is a statement
     // about the re-derived partition — it may not describe more change than the
@@ -5465,13 +5523,22 @@ fn changed_columns_dominate_the_span(pieces: &[Piece]) -> bool {
 ///   partition_block_canonical 0:0/A | 2:3/     an insertion and a deletion
 /// ```
 ///
-/// Both denote `AGC` and both are distance-minimal. On the second, all three
-/// earlier routes refuse: the gap is `2 - 0 = 2` so it is not a single base; the
-/// reference widths total `0 + 1 = 1` against a span of `3`; and the insertion's
-/// reference width is `0`, so it reads as narrower than a lone substitution. The
-/// block is an inversion and the gate says nothing about it — which is the
-/// defect, since the pass exists precisely to keep `inversion.md:5`'s
-/// whole-span reverse complement from being shredded into its columns.
+/// Both denote `AGC` and both are distance-minimal. When #1637 was filed all
+/// three earlier routes refused the second: the gap is `2 - 0 = 2` so it is not
+/// a single base; the reference widths total `0 + 1 = 1` against a span of `3`;
+/// and [`no_piece_is_a_lone_substitution`] then measured reference **width**, so
+/// the insertion read as narrower than a lone substitution. The block is an
+/// inversion and the gate said nothing about it — which is the defect, since the
+/// pass exists precisely to keep `inversion.md:5`'s whole-span reverse
+/// complement from being shredded into its columns.
+///
+/// The third route no longer refuses **this** block: #1703 corrected it to ask
+/// whether a piece *is* a substitution (`delins.md:15`) rather than how wide it
+/// is, and neither an insertion nor a deletion is one. That does not make this
+/// route redundant — it is still the only one that reaches a partition
+/// containing a genuine substitution, which is the shape the spec's own
+/// published 16 nt example takes (`8:9/A`, pinned by
+/// [`the_inversion_coalesce_reaches_the_spec_published_sixteen_nt_example`]).
 ///
 /// The density argument in [`changed_columns_dominate_the_span`]'s own doc is
 /// about **columns**, not reference width, so counting `max(ref_width, payload)`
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
Closing: #1706 subsumes this behaviourally, and the one thing it did not carry has been ported there as fc311f30.

## Measured, not assumed

Of this PR's four tests, only `every_spelling_of_a_whole_span_reverse_complement_converges_on_inv_on_every_arm` measures the defect. The other three pass on the control by design and were excluded from the comparison — `a_reverse_complement_whose_members_are_substitutions_still_splits_on_every_arm` is the #1040 negative control, `a_del_ins_block_that_is_not_a_reverse_complement_still_splits_on_the_arms` is the geometric control, and `the_wrong_form_denoted_the_right_bases_which_is_why_no_oracle_reached_it` never invokes the normalizer at all.

| | shared merge-base `702976c1` | current `main` `ce933533` | #1706 head `98611e94` |
|---|---|---|---|
| the one guard | FAIL | FAIL | **PASS** |
| the three controls | PASS | PASS | PASS |

No residue on #1706's head — 4 run, 4 passed.

## Both defects, by different routes

**The ordering/hull defect** — #1706 asks `block_inversion` on the trimmed block slices, never on a reconstruction from the post-shift piece hull, so the question this PR fixed by re-ordering does not arise there. Its own notes record 20 of 240 synthetic clean inversions shredded when read off the hull.

**The predicate defect** — #1706 neutralised it by adding a wider admission route that strictly subsumes the buggy one, leaving `no_piece_is_a_lone_substitution` itself untouched.

## What was ported

The literal repair. `no_piece_is_a_lone_substitution` measured reference width (`>= 2`) while its name and its `general.md:56` reasoning are about substitution-hood; a function whose name contradicts its body is how the next reader gets misled, and it is not inert — it is behaviour-bearing at `whole_block_inversion`'s four-route gate, where the wider route does not reach. It now tests substitution-hood via a new `Piece::is_substitution`, citing `DNA/substitution.md:5` — "**one** nucleotide is replaced by **one** other nucleotide", both halves load-bearing.

## The adjudication this surfaced

The two PRs adopted different gates and no test discriminated them, so merge order would have settled it silently:

| | admits an `inv` when | a mixed partition (one sub + one indel) |
|---|---|---|
| this PR | **no** piece is a substitution | refused |
| #1706 | **not every** piece is a substitution | admitted |

Ruled for the wider reading, on the ground that `general.md:56` ranks *substitution* and says nothing about a competitor that is only partly substitutions — so one coincidental substitution column should not veto the typing of a whole span. #1461's canonical 133-piece partition of `NC_000013.10:g.100809575_100810031inv`, whose first piece is `g.100809575A>T`, is the case that forces it.

That ruling is now pinned by `a_reverse_complement_competing_with_mixed_members_is_an_inversion`, on a block built so no other admission route can reach it, and verified by mutation: with the wide route removed it fails with `g.[43G>C;46_47inv;50G>C]` while both existing controls stay green. Worth noting what the narrow reading actually produces there — not "no inversion", but a *smaller* one nested inside a three-member allele.

Nothing here is lost; the branch stays for reference.

Issue #1146, opened 2026-07-22
Parser: compound reference syntax is RefSeq-only, so a custom accession's own Display output does not re-parse to the same struct


nh13:
`parse_standard_accession` (`src/hgvs/parser/accession.rs`) recognises the compound `outer(inner)` form only when the inner passes `looks_like_accession_start`, which whitelists `XX_`-style RefSeq, `ENS…`, and `LRG_` prefixes. `parse_simple_accession` has no compound branch at all.

Consequences for a custom (no-underscore) reference:

```
$ ferro parse "NC_000013.11(NM_004119.3)(FLT3):c.100A>G"
-> NC_000013.11(NM_004119.3):c.100A>G          # parses; compound recognised

$ ferro parse "MYSEQ.1(MYTX.1)(GENE1):c.100A>G"
ERROR: Parse error: Expected ':' after accession

$ ferro parse "Template(Template-gene.1)(GENE1):c.5A>G"
ERROR: Parse error: Expected ':' after accession
```

Two problems follow:

1. **Asymmetric acceptance.** The same shape is accepted for RefSeq accessions and rejected for custom ones, purely because of a prefix whitelist.

2. **Display output is string-stable but not struct-stable.** `Template(Template-gene.1):c.5A>G` — the correct, spec-conformant string the projector now emits for a custom reference (#1139/#1141) — re-parses to `accession = Template` with `genomic_context: None` and `gene_symbol = Some("Template-gene.1")`. The transcript id lands in the gene-symbol field. The string round-trips byte-identically, so no rendering test catches it, but a consumer that re-parses `projection.c_name` and reads `.gene_symbol` gets the transcript id rather than a gene symbol.

This predates #1141 (it is identical for the `gene_symbol = None` case on earlier commits) and the emitted string is exactly what #1139 asks for, so it is a parser gap rather than a rendering one. Custom references reach ferro through `convert-gff`, so this is not a synthetic-only concern.

Likely fix: give `parse_simple_accession` a compound branch and relax `looks_like_accession_start` to accept a general accession-shaped inner token, keeping the existing gene-symbol fallback for tokens that do not look like accessions. Needs care — the two are ambiguous by construction, which is presumably why the whitelist exists.

PR #1708, opened 2026-08-12
fix: render a junction-crossing output against its genomic reference (#1704)

### LINKED ISSUE #1704: canonical-coalesced raises outputs_leaving_the_transcript from 371 to 389 (checklist.md:20)
## Summary

Under `FERRO_PARTITION=canonical-coalesced` — the arm the pending default change (#148) selects — the count of normalized outputs naming an intronic position the input never named rises from **371 to 389**. That is **+18 non-conformant outputs**, and the class is independent of every other flip-blocking defect found so far.

The conformance axis reports it in its own words:

> outputs that left the transcript (an intronic position the input did not name) rose from 371 to 389. This is a rank-1 conformance regression, not a representation choice.

## Reproducer

```
FERRO_PARTITION=canonical-coalesced cargo nextest run --features dev --test it \
  -E 'test(three_prime_conformance_census)'
```

`FERRO_MANIFEST` must point at a prepared reference. The `live` arm reports 371.

## Why this is rule 1, not a representation choice

`docs/recommendations/checklist.md:20`:

> NM reference sequences ... **do not contain** intron and gene flanking sequences, and **can only be used** to describe variants in introns using a `c.` prefix when a genomic reference sequence is given.

`README.md:170-212` scopes rule 1 as "every prohibition, read from **prose force** rather than keyword casing — 'not allowed', 'not correct', 'can not be used', 'by definition', the `class="invalid"` markup, and the set `checklist.md` enumerates." Both the "can not be used" family and `checklist.md` itself are named there.

Rule 7 then applies: "Output that **violates** rules 1-4 is a **bug**, not a disclosure." So this blocks #148 rather than being disclosable under rule 6.

## What is NOT wrong, which bounds the search

Measured over the same 58,552-output corpus, every other rank-1 counter is unchanged between the two arms:

| counter | live | canonical-coalesced |
|---|---|---|
| `prohibition_violating_outputs` | 32 | 32 |
| `prohibited_absolute_accepted` | 32 | 32 |
| `outputs_denoting_no_sequence` | 10 / 18 | 10 / 18 |
| `sequence_changed` | 4 / 0 | 4 / 0 |
| `non_idempotent_outputs` | 4 / 4 | 4 / 4 |
| `conflicts_accepted` | 72 | 72 |
| `outputs_leaving_the_transcript` | **371** | **389** |

The flip introduces no new absolute-prohibition violations anywhere else. This is the only rank-1 counter that moves, which makes it a narrow, tractable defect rather than a broad regression.

## What is not yet known

The 18 row identities. The axis reports the count but does not print the rows, so extracting them needs an instrumentation change that was out of scope for the classification pass that found this.

Likely first thing to check: whether the partition re-derivation is placing a member at a junction-adjacent position that the 3' rule then shifts across an exon boundary, in which case `DNA/deletion.md:21`'s junction exception is the interacting clause. That is a hypothesis, not a diagnosis.

## Relationship to the other flip blockers

Independent of all of them. The classification pass put the flip's 81 failures into rule-1 (14), rule-2 (8), rule-6 (44) and test-design (15) buckets; 13 of the 14 rule-1 tests are one defect (#1703, whole-span reverse complement not typed `inv`) and 6 of the 8 rule-2 tests are one clause (`general.md:34`). This is the remaining rule-1 defect that nothing else touches.

nh13:
Closes #1704

Representation-Change: 371 rows respell (389 under `canonical-coalesced`), all accession-only; 4 of 2,075 real-transcript inversion-sweep rows and 6 spec-enumeration rows move, 2 of those from "no genomic representation" to a rendered `g.`. No coordinate moves anywhere.

## The defect

`docs/recommendations/checklist.md:20` — an `NM_`/`NR_` reference "can only be used to describe variants in introns using a `c.` prefix when a genomic reference sequence is given". Ferro's junction-crossing pass emitted **371** descriptions that violate it, and **389** under `FERRO_PARTITION=canonical-coalesced`, which is what this issue reports. Rule 1 of the `README.md` ruleset, so a bug rather than a disclosure.

The producer is #670's deliberate second pass in `normalize_cds`, with mirrors in `normalize_tx` and `normalize_rna`. A purely exonic indel whose exon-confined 3' shuffle lands exactly on an exon's last base is re-shuffled in genomic space, and the answer adopted **only if it crossed into the intron** — the intronic output is that branch's success condition. It was then rendered against the bare transcript accession the input arrived on. Lenient mode was manufacturing a description strict mode refuses as `IntronicOnBareTranscript` / W4007, which is the reverse of the laundering #1406 legislated against.

## The coordinate is not the defect

`background/numbering.md:23` scopes `general.md:44`'s exception to **exon/exon** junctions "where shifting the variant 3' would place it in the next exon", and `:26` states that it "**does not apply** to a deletion/duplication around exon/**intron** and intron/exon junctions". These rows are exactly the case `:26` excludes, and ferro already blocks the exon/exon case the exception does cover. So converging them back on the clamped in-exon answer would implement an exception the spec explicitly withholds, and would silently revert #670. **Only the accession was ever wrong.**

## The fix

`Normalizer::reparent_junction_exit`, at `normalize_core_canonical` — the whole-description seam `split_protein_separation` already runs at — re-parents such an output onto the genomic reference the crossing had already resolved in order to compute itself, giving `NC_…(NM_…):c.20+2del`. That reference is not merely obtainable: #670 declines outright when `transcript.chromosome` is `None`, so it is in hand at the moment the answer is adopted. One site covers all three axis mirrors.

Three things it deliberately does:

- **Lifts the wrapper to every leaf on that accession.** `AlleleVariant`'s compact form requires the members to share an accession, so wrapping only the crossing member would drop `c.[10+2del;18del]` into the expanded `[NC_…(NM_…):c.10+2del;NM_…:c.18del]` — a far larger representation change than the defect it repairs. A genomic wrapper on an exonic position is unremarkable.
- **Fires only where ferro manufactured the offset.** An input that already names a bare intronic position is left exactly as authored; the `bare-transcript-intronic-position` ruling decided that class (strict refuses, lenient accepts as written), and re-spelling it here would overturn one decided record as a side effect of implementing another.
- **Declines safely.** The `chromosome` field is a provider-supplied string, not necessarily an accession. It must parse in full and satisfy the parser's own `is_valid_compound_outer` rule, so ferro never emits a compound reference `parse_hgvs` would refuse to read back. Otherwise the output is returned untouched.

Rejected alternative: **declining in strict mode**. The census is a lenient-mode figure, so a strict refusal leaves all 371 exactly where they are; the input (`c.7del`) is a perfectly valid description, so refusing it because ferro's own chosen output form is inexpressible fails rule 1 in the other direction; and it would make the answer depend on the error mode, which rule 4 does not admit.

## The census counter had to be re-scoped, and that is a finding

`outputs_leaving_the_transcript` counted every `Denotation::Inexpressible` output. Re-parenting does not move that at all — SPDI is positional and has no offset notation, so `hgvs_to_spdi` declines an intronic position whatever accession carries it. **Measured, not inferred: against the pre-existing counter the fix reports 371 before and 371 after** — a change that made all 371 outputs conformant, invisible to the instrument. The proxy and the clause had silently come apart.

The census now reads `checklist.md:20`'s own question — *is the accession bare?* — directly off the AST, written in the test rather than borrowed from `src/normalize/`, so the instrument still judges ferro from outside. The conformant rows are counted by a new `outputs_intronic_under_a_genomic_wrapper` so the class stays **accounted for** rather than merely stopping being counted.

| counter | before | after |
|---|---|---|
| 3' `outputs_leaving_the_transcript` | 371 | **0** |
| 3' `outputs_intronic_under_a_genomic_wrapper` | — | **371** |
| 5' both | 0 | **0** |
| 3' under `FERRO_PARTITION=canonical-coalesced` | 389 | **0** (389 wrapped) |

`371 = 0 + 371` is the accounting identity that makes this a fix rather than a re-partition: no row stopped reaching the check. Every other counter is byte-identical in both directions and both arms — 58,552 outputs, 0 declined, 0 unparseable, 10/18 denoting no sequence, 32 prohibition-violating, 9,141 converged, 4 non-idempotent, 4/0 sequence-changed, 72/32/40 refusal, 0 guard violations.

The positive control is on file: the counter was confirmed at 371 in this worktree with a freshly built binary before any fix, and the `canonical-coalesced` arm was re-measured with `src/normalize/` reverted to confirm that its `guard_violations` 0 → 6 failure is the coalesced partitioner's own pre-existing divergence and is unmoved by this change.

## What moved outside the synthetic corpus

The corpus is not the whole story, and two real-data suites found rows the diagnosis had not:

- **`inversion_sweep`** — 4 of 2,075 authored inversions on `NM_004006.2` (DMD) shift into an intron and now read `NC_000023.11(NM_004006.2):c.1322_1332-1inv` and siblings. These come from an `inv` edit with a **negative** offset, so the class was never confined to #670's `del`/`dup` shape; the fix is keyed on the output, so it covers them. The sweep's accession check treated the wrapper as a re-anchor onto another transcript — it is not, the inner accession is unchanged — and now compares `without_genomic_context()`.
- **`spec_enumeration`** — 6 rows. Four are the same accession repair on the `c.` and `n.` axes. The other two are a **capability gain**: `project-g/c.1704del` and `project-g/c.1704dup` read `unavailable: no g. representation for this variant` and now render `NC_000023.11:g.32573745del` / `dup`. While the `c.` answer was spelled on a bare `NM_004006.2` there was no genomic reference in the description for the genomic axis to derive from; restoring the anchor restores the projection.

Not measured with `examples/dump_normalized_corpus.rs`, deliberately: its generator builds a **single exon** with `CDS_START = 1` (#1478), so no member can cross a junction and it would report a structural zero for exactly this class.

## Guards

- `defect_371_transcript_exit` keeps its name and every measurement — the mechanism is unchanged, and each expectation moved by the wrapper prefix and by nothing else, which is the cleanest available statement of what the fix did and did not do. Eight tests added: the wrapper on both axes and both strands, one accession for an allele whose members disagree, an authored intronic input left as authored, and five that measure the rendering guards themselves (see below).
- **The three rendering-guard conjuncts are each measured rather than assumed**, because two of them were being credited with the wrong work. Which conjunct catches a given chromosome is pinned behaviourally: a SAM-style `chr17` and an assembly name `GRCh38` are excluded by the **bare accession parse**, one step *before* the pairing rule — `parse_hgvs("chr17(NM_TEST.1):c.10+2del")` is `Ok`, since `is_valid_compound_outer` deliberately admits an unclassifiable custom accession (#1146). The pairing rule's own work is refusing a known transcript outer (`NM_OTHER.1(NM_TEST.1)` is `Err`). The code comment credited the pairing rule for both and has been corrected.
- **`outer == bare` is reachable, on exactly one accession class**, and `a_self_referential_wrapper_is_declined` now covers it. It is unreachable for every RefSeq accession — `bare` is a transcript and the pairing rule rejects a transcript outer first — but a **bare LRG record** slips both earlier conjuncts: the `checklist.md:20` predicate admits LRG by prefix (`Accession::is_lrg`, so `LRG_<N>` as well as `LRG_<N>t<M>`), while `is_valid_compound_outer` keys off `inferred_variant_type`, which reads a bare `LRG_<N>` as genomic. Measured: with that conjunct dropped, `LRG_5:c.7del` emits `LRG_5(LRG_5):c.10+2del`. Nothing downstream would catch it either — that string re-parses, so `FERRO_ASSERT_REPARSE` is blind to it. The provider shape is not contrived: an LRG record *is* its own genomic reference.
- `intronic_on_bare_transcript_warning` is split into an allocation-free predicate (`intronic_on_bare_transcript_axis`) that the warning constructor is derived from. The `is_some()` form was building a `NormalizationWarning` carrying `format!("{variant}")` and dropping it on every `normalize()`. **The single-predicate property is preserved deliberately** — one rule decides both the strict-mode refusal and whether an output needs the wrapper, and two rules that can drift is the defect this PR fixes.
- `conformance::inversion_sweep::tests::a_genomic_wrapper_is_not_a_re_anchor` covers the accept direction of `names_the_sweep_transcript`. Its two neighbours only test the reject side and both still pass against the bare string comparison the helper replaced, so without it the helper was untested in the only direction it changed.
- `spec_corpus_regressions::a_minus_strand_junction_shift_leaves_the_transcript` re-pointed. Its sibling collision row keeps its `PINNED DEFECT` marker: the accession half is fixed, the two-identical-members half is a separate defect and was not touched.
- `in_bounds_oracle_scope` re-pointed, and its scope statement is **unchanged** — every endpoint is still intronic, so `simple_cds_pos` still short-circuits and the armed child is still silent. The oracle's blindness was to the coordinate shape, never to the defect.
- The `bare-transcript-intronic-position` ruling record is updated to record the closure, plus two corrections it had been carrying: the 371 are 331 minus-strand and 40 plus (not "all minus-strand"), and calling the producer a "clamp" pointed every reader at the repair that would have reverted #670.

## Known residue: a mixed allele still ships a manufactured offset bare

Disclosed rather than closed, and pinned by `a_mixed_allele_still_ships_a_manufactured_offset_bare` so it cannot move unnoticed in either direction.

```
NM_TEST.1:c.[7del;30+5del]  ->  NM_TEST.1:c.[10+2del;30+5del]
```

`10+2` is ferro's own manufacture — the exact class this PR closes — and it ships on a bare accession. The cause is a granularity mismatch: provenance ("did ferro manufacture this offset?") is a property of each **leaf**, but the guard asks it at whole-description granularity, and `names_bare_transcript_intronic(input)` is an ANY-leaf existence test. So one authored member (`30+5`) vetoes the wrapper for every other member, including one ferro moved itself. Each member is handled correctly alone; only the mixture is wrong.

It is not closed here for two reasons. Per-leaf provenance has to be carried from the site that manufactures the offset — it cannot be recovered by comparing input against output afterwards, because normalization reorders, merges and splits members, so no identity map from an output leaf back to an input leaf survives. And the rendering question that then arises is a representation-policy choice rather than a bug fix: the compact form `ACC:c.[a;b]` requires its members to share an accession, so a mixed description must either lift the wrapper to the whole description (re-spelling an authored member, which the `bare-transcript-intronic-position` ruling decided against) or expand to per-member accessions (a much larger representation change). **The fix is a stacked follow-on.**

## Verification

Rebased onto `6116f84a` (#1698) and re-verified there — the census in particular, since #1698 also moves normalizer output and could have shifted it independently. It does not: `spec_conformance_axis`, `inversion_sweep`, `spec_enumeration`, `defect_371_transcript_exit`, `in_bounds_oracle_scope`, `spec_corpus_regressions` and `hgvs_spec_normalization_tests` are 91/91 green at the new base, with every pin unchanged. `cargo nextest run --features dev --lib --test it` clean, plus `cargo fmt --all` and `cargo clippy --all-features --all-targets -- -D warnings`. Both spec artifacts regenerated **after** the rebase — they are gitignored and regenerate only when missing, so a stale one survives a rebase and surfaces as a bogus `ruling_records_are_intact` record-count mismatch. The ruling ledger itself was diffed against `origin/main` programmatically: 21 records both sides, none dropped, no status changes.

Each guard added or changed was mutation-tested three ways (always-true, always-false, and a plausibly-weaker form), and the tests that fire for each were recorded rather than assumed. `outer == bare` was reported as dead code on that basis and is not — see above.

With `FERRO_ASSERT_IDEMPOTENT` / `REPARSE` / `IN_BOUNDS` armed over the whole suite, 7 tests fail — all of them the pre-existing `cds-end` class (`c.*1delinsCTT` → `c.72_*1insCT` → `c.72delinsCCT`), which involves no intronic position and no accession. Confirmed by re-running them with `src/normalize/` reverted: identical failures. CI arms those flags over selected subsets rather than the whole suite.


nh13:
@coderabbitai pause

PR #1710, opened 2026-08-12
test(conformance): count coding-axis merges across two or more unchanged nucleotides

nh13:
Representation-Change: none. Test and conformance-corpus instrumentation only; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched, and the full suite is green on the shipping rule with every pre-existing pin unmoved.

**Rebased onto `origin/main` `57f71d31`** (from `e98fa77e`) while applying review findings. The branch's own delta is byte-identical across the rebase apart from the three edits described under "the 5 are not unresolved" — verified by diffing `git diff <old-base>..<old-tip>` against `git diff <new-base>..HEAD`, which differ only in those hunks and in hunk offsets. `57f71d31` carries #1718, which touches the same conformance-axis files, so the local gate was re-run on the rebased tree rather than inherited.

## Read this first — the motivating figure moved by 40x while this PR was open, and nothing went red

This PR was written against a `canonical-coalesced` figure of **122 of 997**. **#1698** (`fix(normalize): require a gap-bearing insert before the delins.md:44-47 merge`) landed on `src/normalize/merge.rs` between the branch point and this rebase and took that figure to **3 of 997** — 119 of the 122 removed. **No test failed**, because the figure lived only in prose. That is the exact hazard this module exists to remove, realised on the module itself, so it is recorded here rather than quietly restated.

Every figure below has been re-derived on `origin/main` at `1ea75334`, with attribution measured rather than assumed. The four arms were re-run twice — once with the original numerator and once with the widened one — so movement is attributed to #1698 rather than to this branch's own changes.

| figure | as originally written | re-derived at `1ea75334` | why it moved |
|---|---|---|---|
| `canonical-coalesced` arm | **122** of 997 (3' and 5') | **3** of 997 (3' and 5') | #1698 |
| its decomposition | 31 / 30 / 30 / 31 at gaps 2 / 3 / 5 / 8, two- and three-member designs | 1 row at separation 3, 2 rows at separation 2 — all two-member, all collapsing to one member | #1698 |
| shipped rule, `shadow` | 0 / 0 | 0 / 0 | unchanged |
| `canonical` | 1 (`s01-c1-pair-inv-del-p4-sep3`) | 1, the same row | unchanged |
| sub-floor merges (separation exactly one) | 16 | **41 of 340** | the numerator was widened; see below |
| …of which collapse all the way to one member | 16 | **16** | **unchanged by #1698** |
| pass attribution of those 16 | "14 from `merge.rs:188` + 1 from `coalesce_whole_block_inversion`" | 15 two-member pairs + 1 three-member whole-span `inv` (`s00-c1-m3-all-inv-p4-sep1`) | the original 14 + 1 sums to 15 against a stated total of 16 |
| sub-floor verdicts | 9 licensed / 5 **unresolved** / 2 rule 1 violations | 9 licensed / 5 **unlicensed** (#1744) / 2 **unlicensed, filed** (#1716) | see "the 5 are not unresolved"; and #1725 makes none of the 7 a rule-1 violation |
| suite | "10,163 passed, 0 failed" | 10,162 passed / **1 failed** at that head | the claim predated the commit that broke it; green now |

**What survives that, and what does not.** The scale argument does not: at 3 rows this is not "the canonical arm's largest blocking family", and saying so would be reframing around a number instead of reporting it. The **disjointness** argument does, and it is the one the counter rests on — `guard_violations` *cannot* count these rows at any size, which is a structural fact about `is_svd_wg010_shape`'s domain rather than a claim about how many rows there happen to be. #1698 is itself the argument for keeping the counter: it moved this population by 97.5% and every test stayed green.

## The gap this closes

`spec_conformance_axis`'s `guard_violations` is a **negative guard for rejected SVD-WG010**, and its rows are chosen by `spec_corpus::is_svd_wg010_shape`: the two **frameless** shapes (`Genomic`, `NonCodingMultiExon`), a separation of **exactly one**, and **exactly two** members. The coding axis is outside that domain twice over.

So when a coding-axis multi-member allele is merged, **no counter in the conformance axis can rise** — not "does not today", but cannot, at any separation and for any member count. That disjointness is now pinned as a property of the corpus by `the_coding_axis_merge_population_is_what_it_says_it_is`, not asserted in a doc comment.

## What this adds

A second counter **beside** the guard, never a widening of it — `is_svd_wg010_shape` is untouched, because relaxing its `frameless` conjunct would silently re-scope a negative result about a *rejected* proposal into a different instrument carrying that guard's name and its published zero.

- `Row::coding_axis_separation_two_or_more` marks the population: a **coding**, `Mechanism::Cis`, multi-member design whose members each consume reference and sit **two or more** unchanged nucleotides apart, with an **irreducible** separation.
- `Census::coding_axis_separation_two_or_more_merges` counts how many of those normalize to **fewer members than were authored**, on the authored spelling only.
- `Census::coding_axis_separation_two_or_more_rows` and `CorpusShape::coding_axis_separation_two_or_more_rows` are the denominator, pinned on both sides and asserted non-zero, so `0 of 0` fails as VACUOUS rather than passing as a result.
- `the_coding_axis_merge_counter_can_observe_a_merge` is a **positive control** — see below.

## Read the ledger, not only the clauses

`general.md:34` / `DNA/delins.md:17` are what make the population interesting. They are **not** the whole authority over it, and an earlier revision of this PR cited only them. `tests/it/clause_ruling_index.rs` marks both `[MULTI]`, and three records govern parts of these rows:

| record | status | what it does to this population |
|---|---|---|
| `delins-payload-coincidence-carve-out-is-coding-dna-scoped` | decided | scopes `delins.md:47`'s payload-coincidence carve-out to the **coding DNA axis** — this population's axis, and the one axis where `general.md:34` is overridden for that shape |
| `delins-merge-vs-individual-gap-two-or-more` | decided | `:44-47` governs `:17`, scoped to the alignment-coincidence shape and to the net-deletion direction; outside that, `general.md:34` still governs |
| `delins-recommendation-reach-when-the-input-arrives-split` | not settled | the genuine residue: which clause a *re-derivation* lands on. Narrower than "may ferro merge an authored split", which `canonical-form-choice-when-both-legal` already answers |

**So neither a rise nor the zero means one thing.** Within the payload-coincidence shape the first record *ratifies* the merge on this axis, so a rise there is a decided ruling arriving on the shipped rule and the shipping arm's `0` is that ruling **not yet implemented** — not a clean bill. Outside that shape `general.md:34` governs and a rise needs explaining. The counter cannot separate them, which is why it counts and does not adjudicate. The pass producing every merge the counter sees on the coalesced arm, `coalesce_payload_alignment_split`, cites the first record in its own doc comment.

## The numerator was blind to a partial merge

It tested `members_of(output).len() < 2`, copied from `guard_violations` — where `is_svd_wg010_shape` pins exactly two members, so "fewer than two" and "merged" are the same statement. This population admits 2, 3 and 4 members, and **313 of the 997 flagged rows carry three or four**, so a 4-member row merged to 2 counted as **zero**.

It now tests `< row.members`, which is the property the doc comments already claimed. Above the floor this moves nothing — all four arms re-measured either way read 0 / 0 / 1 / 3. Below the floor, **25 of the 41 merges were uncounted**.

## A positive control, because a zero pin cannot detect a blinded counter

Hard-wiring the numerator so it can never increment **passed** both censuses, the corpus-shape test and the population test. "Pinned at 0" and "wired to 0" were the same observation — in a module whose thesis is that a counter nobody has seen move is not an instrument.

`the_coding_axis_merge_counter_can_observe_a_merge` exercises the predicate directly against real corpus rows: an authored multi-member spelling must not count, a single-member output must, and a 3-or-more-member row collapsed to two must. Verified against the mutation: with the numerator wired to `false`, it is the **only** one of the five tests that fails.

The rest of the matrix was already sound and is unchanged: predicate always-true, predicate always-false (as VACUOUS, in the exact words the docs promise), the floor weakened to one, and the numerator forced always-true are each caught; dropping the `irreducible` conjunct moves the denominator 997 -> 2,527 and the numerator 0 -> 74, so that conjunct is load-bearing.

## Measured, in both directions

Pinned at **0 of 997** in `THREE_PRIME` and `FIVE_PRIME`. The zero is a real negative result: `coalesce_payload_alignment_split` is reached only when `payload_coalesce_applies` sees `PartitionRule::CanonicalCoalesced`, so it never runs on the shipped rule. That is a mechanical reason, not an empirical coincidence — which is also why the shipping zero is invariant under #1698 while the coalesced arm collapsed.

| `FERRO_PARTITION` | 3' | 5' |
|---|---|---|
| unset — the shipped rule | **0** of 997 | **0** of 997 |
| `shadow` | 0 | 0 |
| `canonical` | 1 | 1 |
| `canonical-coalesced` | **3** | **3** |

## Why the floor is two, and what that costs

`general.md:35` / `DNA/delins.md:18` carve out an exception on this same axis for two variants "separated by **one** nucleotide, together affecting one amino acid". Measured on the shipping rule at `1ea75334`: **41 of 340** rows merge at a separation of exactly one, and **0 of 997** at two or more.

The 41 decompose by how far each row collapsed — the finding line now reports `<authored> members -> <observed>`:

| shape | rows | visible to the old `< 2` numerator? |
|---|---:|---|
| 2 authored members -> 1 | 15 | yes |
| 3 authored members -> 1 (a whole-span `inv`) | 1 | yes |
| 4 authored members -> 3 (a **partial** merge) | 25 | no |

So the earlier figure of **16** is the first two rows and is unchanged; read "16 -> 41" as an instrument that got less blind, not as behaviour that changed.

**Those 16 are not the codon exception firing.** Knocking out each merge pass singly shows `apply_coding_codon_exception` produces none of them.

**And the 5 are not unresolved.** An earlier revision filed them as unresolved on the strength of `codon-carve-out-shape-restriction`, which is a **decided** record (WIDEN, operator ruling 2026-08-10) and which answers this sub-case rather than leaving it. Two claims were being run together: what is *not yet implemented* is the widening's effect on `apply_coding_codon_exception`; what is *not* unsettled is whether the exception reaches a frameshift pair. It does not — the record's own words are that "together affecting one amino acid" cannot cover a frameshift pair, "so the widened rule declines there for the reason the spec gives rather than by a shape test".

Ferro merges those 5 anyway, which puts them in the same category as the 2 filed as **#1716**: output no cited clause licenses. **On this corpus the sub-floor count of unlicensed merges is 7, not 2** — and since the 9 / 5 / 2 decomposition is over the 16 full collapses only, with the 25 partial merges unclassified, 7 is a floor.

**The 5 are filed as #1744, and that is now named in the code rather than described as a gap.** The comment used to read "the missing artifact is an issue for the 5, filed the way #1716 was for the 2, and until that exists this comment is the only record"; the issue exists, so the comment points at it. Read #1744 rather than the comment for their status: under `separation-rule-force-modal-or-negation` (**decided**, 2026-08-12) all 7 are rule-**2** departures to disclose rather than rule-7 bugs, and whether they are *accepted and disclosed* or *closed* is a trade nobody has made. The comment previously said the classification changed "what they cost a release and not whether they are wrong", which pre-empted that open question; it now says so explicitly and defers to #1744.

| verdict | rows | what they are |
|---|---:|---|
| licensed | 9 | two lone substitutions in one codon, net length 0 — the exception's own shape; plus one whole-span `inv` |
| **unlicensed** | **5** | a frameshift *inside* one codon; the decided record above declines them |
| **unlicensed, filed** | **2** | the merged span crosses two codons — #1716, reproducing on the prepared reference as `LRG_199t1:c.[18_19del;21A>C]` -> `c.18_21delinsTC`. **Not a rule-1 violation**: #1725's `separation-rule-force-modal-or-negation` grades `general.md:34` as README rule 2 in its entirety, so this is a deviation to disclose and pin, not a rule-7 bug |

A floor of one would therefore open at a non-zero pin summing rows with three different verdicts. The floor is read off the exception's own stated conjunct, so it is a textual scope rather than a ruling. The cost is stated rather than hidden, and it is worse than the first revision of this PR said: the sub-floor gap conceals **known defects**, not merely an unsettled question.

## One argument, one place

The floor-of-two rationale was written out seven times across the two files and the sub-floor figure three times, and two of those copies had drifted into contradicting each other **406 lines apart in one file** — one asserting ferro implements the codon exception and it accounts for the sub-floor merges, the other retracting exactly that. It is now stated once, on `spec_corpus::is_coding_axis_separation_two_or_more_shape`; every other site points at it.

## Also

- The population predicate requires `Mechanism::Cis` rather than `combines_members()`, which also admits `Trans` and `UnknownPhase`. No row changes today (probe: `by_mechanism = {cis: 997}`), but a merged *trans* allele is an unambiguous defect and must not arrive silently under prose describing an unsettled adjudication.
- The disjointness assertion names the SVD-WG010 guard rather than asserting the row carries no guards at all, so a later unrelated guard label cannot fail it with a message about SVD-WG010.
- The population test now discloses that it does **not** check irreducibility — a `Row` does not carry the triples the check needs, and dropping that conjunct leaves the test green while moving the census figures. It is defended by the two census pins, not by that test.

## Verification

Re-run after the rebase, not before — the previous claim of "10,163 passed, 0 failed" was taken before the commit that broke it, and that head in fact ran 10,162 passed / 1 failed on `ruling_citation_currency::no_ruling_citation_contradicts_the_ledger_status`, which was this branch's only CI failure and is fixed here.

- `cargo run --features dev --bin generate_spec_fixture` and `generate_spec_enumeration`, then `cargo nextest run --features dev --lib --test it`, with `FERRO_MANIFEST` set so the reference-aware suites run rather than skip. Exit status captured into the log and read back out of it.
- `scripts/run_oracle_suite.sh`, which mirrors `test-oracle`'s flags and selection.
- `cargo fmt --all --check` and `cargo clippy --all-features --all-targets -- -D warnings` both clean.
- Every pre-existing pin is unmoved in both directions, including the counters #1708 re-blessed.

### DIFF
--- a/src/conformance/spec_corpus.rs
+++ b/src/conformance/spec_corpus.rs
@@ -523,6 +523,37 @@ pub struct Row {
     /// emitted SVD-WG010's own worked example. So the shape is generated with the
     /// guard attached rather than left to be re-discovered.
     pub negative_guards: Vec<&'static str>,
+    /// Whether this row is in the population a **coding-axis** merge across
+    /// **two or more** unchanged nucleotides could be observed on. See
+    /// [`is_coding_axis_separation_two_or_more_shape`], which carries the
+    /// reasoning — including why the floor is two rather than one.
+    ///
+    /// # An instrument's denominator, not a guard and not a ruling
+    ///
+    /// [`Self::negative_guards`] marks a row whose merge would implement a
+    /// **rejected** proposal, so a violation there is a verdict. This flag marks
+    /// a row where a merge is merely *interesting*, and its counter reports how
+    /// many such merges happen. It asserts nothing about whether they are right:
+    /// the adjudication is the operator's and is open.
+    ///
+    /// What makes the population worth counting is that `general.md:34` and
+    /// `DNA/delins.md:17` speak to it in as many words — "two variants separated
+    /// by one or more nucleotides should be described individually and **not**
+    /// as a 'delins'".
+    ///
+    /// # Why it exists as a separate flag
+    ///
+    /// [`is_svd_wg010_shape`]'s domain and this one's are **disjoint by
+    /// construction**: that guard admits only the two frameless shapes
+    /// ([`RefShape::Genomic`], [`RefShape::NonCodingMultiExon`]) at a separation
+    /// of exactly one with exactly two members, so it cannot count a coding-axis
+    /// merge at any separation, for any member count. Widening it would silently
+    /// re-scope a guard for a rejected proposal into something else, so this is a
+    /// second marker beside it rather than a change to it.
+    ///
+    /// Set by [`build_family`] alone, so it is a property of family rows;
+    /// single, conflicting and prohibited rows carry `false`.
+    pub coding_axis_separation_two_or_more: bool,
     /// Member layout (#1456).
     pub geometry: Geometry,
     /// Transcript placement (#1478).
@@ -1483,6 +1514,20 @@ impl SpecCorpus {
         self.rows.iter().map(|row| row.spellings.len()).sum()
     }
 
+    /// Rows in the coding-axis merge instrument's population — the denominator
+    /// its count is `n of` (see [`Row::coding_axis_separation_two_or_more`]).
+    ///
+    /// Reported so a consumer can fail as VACUOUS rather than read `0` as a
+    /// result: a corpus that stopped generating coding-axis separated designs
+    /// would otherwise report no merges and look like a clean bill of health.
+    #[must_use]
+    pub fn coding_axis_separation_two_or_more_rows(&self) -> usize {
+        self.rows
+            .iter()
+            .filter(|row| row.coding_axis_separation_two_or_more)
+            .count()
+    }
+
     /// Rows of each kind.
     #[must_use]
     pub fn by_kind(&self) -> BTreeMap<RowKind, usize> {
@@ -2118,18 +2163,36 @@ fn build_family(design: Design<'_>) -> Attempt {
         return Err((design.id, DropReason::Singleton));
     }
 
+    // Both the negative guard and the coding-axis merge instrument want the same
+    // property of the design: that the unchanged bases between the members
+    // cannot be shifted away. Two members separated by one `T` inside a `T`-tract
+    // are the *same variant* as two adjacent members, so a merge there is the 3'
+    // rule rather than a merge across a separation. Computed once and read twice;
+    // see `is_svd_wg010_shape`'s "This is only half the test".
+    let irreducible = triples
+        .iter()
+        .all(|t| equivalent_placements(served, t.0, &t.1, &t.2).is_empty());
+
     // The negative guard survives only when the intervening base is
     // irreducible: see `is_svd_wg010_shape`.
-    let negative_guards = if design.negative_guard_candidate
-        && triples.len() == 2
-        && triples
-            .iter()
-            .all(|t| equivalent_placements(served, t.0, &t.1, &t.2).is_empty())
-    {
+    let negative_guards = if design.negative_guard_candidate && triples.len() == 2 && irreducible {
         vec![SVD_WG010_GUARD]
     } else {
         Vec::new()
     };
+
+    // The coding-axis merge instrument's population. `triples.len()` must equal
+    // the member count for `irreducible` to be a statement about every member —
+    // a design whose members collapsed into fewer triples has already been
+    // re-read by the applier, so its separations are not the authored ones.
+    let coding_axis_separation_two_or_more = triples.len() == design.members.len()
+        && irreducible
+        && is_coding_axis_separation_two_or_more_shape(
+            frame.shape,
+            &design.members,
+            design.separation,
+            design.mechanism,
+        );
 
     let window = hi - lo + 2 * 128;
     Ok(Row {
@@ -2146,6 +2209,7 @@ fn build_family(design: Design<'_>) -> Attempt {
         mechanism: design.mechanism,
         prohibition: None,
         negative_guards,
+        coding_axis_separation_two_or_more,
         geometry: design.geometry,
         region: design.region,
         scale_bands: scale_bands(hi - lo, window),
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #1718, opened 2026-08-12
test(conformance): refuse a zero denominator on a mutalyzer axis, and measure the corpus by member arity

### LINKED ISSUE #1717: mutalyzer axis suite: no non-zero denominator, and its agreement figure cannot be read by member arity
The mutalyzer axis suite is the repository's agreement-with-a-reference-implementation oracle. Two things about it are not currently checkable from a run, and one of them lets a run that measured nothing report the same green as a run that measured everything.

## 1. `AxisTally::finish` has no non-zero denominator

`tests/it/mutalyzer_normalize_tests.rs` gates each axis on `assert!(self.fail.is_empty(), …)`. Nothing asserts that the axis compared any rows at all. With a manifest present but a provider that serves nothing, every bucket is empty, `fail.is_empty()` holds, the empty-projection budget is trivially within budget, and the axis prints `0 pass / … / 0 FAIL` and passes. `0 of 0` is not evidence of agreement.

This is the honest-zero discipline the rest of the suite already enforces — `spec_conformance_axis.rs` states it explicitly ("Every property whose denominator is zero fails loudly as **VACUOUS** rather than passing") and six test modules implement it. `mutalyzer_normalize_tests.rs` is not one of them; `git log -S 'VACUOUS' -- tests/it/mutalyzer_normalize_tests.rs` is empty, so this is a gap rather than a decision.

## 2. Agreement with mutalyzer is arity-dependent, and the corpus's arity split is invisible

Mutalyzer's normalizer minimises a weighted description length and carries no separation rule at all, so on a multi-member allele a ferro/mutalyzer disagreement is two objectives meeting rather than evidence either side is wrong; on a single-member row the two operations coincide and the comparison is much closer to meaningful. The suite reports one number per axis and does not distinguish the two populations, so an agreement figure taken from it cannot be read for scope.

The split is real. Adjudicated-divergence rate over the committed corpus (`tests/fixtures/mutalyzer-normalize/cases.json`, 335 `to_test` rows), single-member vs multi-member:

| axis | single | multi |
|---|---|---|
| normalized | 39 / 218 (17.9%) | 8 / 25 (32.0%) |
| genomic | 19 / 121 (15.7%) | 5 / 8 (62.5%) |
| infos | 1 / 10 (10.0%) | 8 / 12 (66.7%) |
| all axes | 174 / 553 (31.5%) | 30 / 59 (50.8%) |

(Those figures come from a textual scan of `cases.json`. A `parse_hgvs`-based classifier agrees on the shape but reports 8 of the 243 `normalized` rows as unclassifiable, because ferro refuses to parse them — 6 otherwise-single and 2 otherwise-multi — leaving `single 33/212, multi 6/23, 8 unparsed`. Either way the multi-member rate is the higher one.)

The corpus itself is well shaped for this — 35 of its 335 `to_test` rows are multi-member — but nothing keeps it that way across an upstream refresh, and the failure mode is not hypothetical. The stratified ClinVar population the benchmark harness draws its headline agreement numbers from (`clinvar_patterns.txt`, the 57,524,053 lines `docs/BENCHMARK_GUIDE.md` quotes) carries **1,026 multi-member rows**, 724 of them cis — so a 20,000-row draw from it holds **0.0004** in expectation. Any agreement rate taken from that harness is a single-member rate whatever it is labelled, and nothing in the repository says so.

## Proposed

- Assert a non-zero denominator per axis in `finish`, matching the existing `VACUOUS` idiom.
- Add a manifest-free census that reports the `normalized` axis's single- and multi-member populations apart and floors the multi-member one, so the coverage cannot silently collapse. Manifest-free matters: PR CI never sets `FERRO_MANIFEST`, so the axis tests themselves take the no-manifest early return there and report a green that compared nothing.

Both are test-only. Neither changes what any axis expects.

nh13:
Closes #1717

Two test-only changes to the mutalyzer axis suite, both about what a green run is allowed to mean.

**A zero denominator is no longer a pass.** Each axis gated on `assert!(self.fail.is_empty(), …)` and nothing else, so an axis that compared *no rows* passed exactly like one that compared all of them and agreed — a manifest present, a provider serving nothing, every bucket empty, `0 pass / … / 0 FAIL`, green. `AxisTally` now counts the rows it recorded and `finish` refuses a zero, using the `VACUOUS` idiom `spec_conformance_axis.rs` states and six other test modules implement. `git log -S 'VACUOUS' -- tests/it/mutalyzer_normalize_tests.rs` is empty, so this was a gap rather than a decision. On the blessed reference every axis records well clear of it: normalized 243, genomic 129, protein_description 86, errors 57, coding_protein_descriptions 51, infos 22, rna_description 16, noncoding 8.

**The corpus's member-arity split is now measured and floored.** Agreement with mutalyzer means two different things either side of that split: mutalyzer minimises a weighted description length whose constants are dated 2014 and carries no separation rule at all, so on a multi-member allele a disagreement is two objectives meeting, while on a single-member row the two operations coincide and the comparison is much closer to meaningful. `normalized_axis_covers_both_member_arities` reports the two populations apart and floors the multi-member one so it cannot silently collapse.

It measures `single 33/212 adjudicated, multi 6/23 adjudicated, 8 unparsed` — the unparsed rows being ones ferro refuses to parse, reported rather than guessed at. Across the whole corpus the adjudicated-divergence rate is 31.5% (174/553) single against 50.8% (30/59) multi, so the split is not decorative.

The floor exists because the population is easy to lose by accident. The stratified ClinVar population the benchmark harness draws its headline agreement numbers from — `clinvar_patterns.txt`, the 57,524,053 lines `docs/BENCHMARK_GUIDE.md` quotes — carries 1,026 multi-member rows, 724 of them cis. A 20,000-row draw from it holds 0.0004 in expectation, so an agreement rate measured there is a single-member rate whatever it is labelled. This corpus is deliberately not that population, and the floor is what keeps it from drifting into it across an upstream refresh.

The census is manifest-free on purpose: PR CI never sets `FERRO_MANIFEST`, so the axis tests themselves take the no-manifest early return there and report a green that compared nothing. The census is the half of this that holds on a pull request.

Nothing here is framed as conformance. `adjudication-precedence-order` puts spec-explicit text above mutalyzer and mutalyzer above our judgement, so these axes measure agreement with a reference implementation and a divergence is a question to adjudicate, not a verdict. The module docs now say so where the numbers are produced.

Verified against the blessed reference (`FERRO_MANIFEST=…/data/ferro/manifest.json`, computed reference identity `aa8b3246d83055cc`, empty-projection gate enforcing rather than skipping): all eight axes green with 0 FAIL and bucket counts unchanged from `origin/main`. Full suite with the manifest armed: 10,378 passed, 0 failed. `cargo fmt --all -- --check` and `cargo clippy --features dev --all-targets -- -D warnings` both clean.

Representation-Change: none. Test-only — no `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` code is touched and no axis expectation moves.

### DIFF
--- a/tests/it/mutalyzer_normalize_tests.rs
+++ b/tests/it/mutalyzer_normalize_tests.rs
@@ -10,6 +10,17 @@
 //! demote inputs from
 //! `tests/fixtures/mutalyzer-normalize/baseline-failures/<axis>.txt`.
 //!
+//! **What the axes measure, and over what.** Agreement with a reference
+//! implementation, never conformance — `adjudication-precedence-order` puts
+//! spec-explicit text above mutalyzer and mutalyzer above our judgement, so a
+//! divergence here is a question to adjudicate, not a verdict. The population
+//! matters as much as the rate: mutalyzer has no separation rule, so agreement
+//! on a multi-member allele is a weaker signal than on a single-member row, and
+//! a corpus that carries only the latter yields a figure that silently describes
+//! only the latter. `normalized_axis_covers_both_member_arities` reports the two
+//! populations apart and refuses to let the multi-member one collapse, and
+//! [`AxisTally::finish`] refuses a zero denominator on any axis.
+//!
 //! Backed by the ferro-prepared reference manifest. When the manifest is
 //! absent (CI by default), every axis test reports `skipping — no manifest`
 //! and exits 0. Same convention as `tests/real_data_normalization_tests.rs`.
@@ -354,6 +365,24 @@ struct AxisTally {
     reference_unavailable: Vec<(String, String)>,
     fail: Vec<(String, String)>, // (input, diagnostic)
     skipped: usize,
+    /// Rows this axis actually **compared** — every row [`record`](Self::record)
+    /// was called for, whatever bucket it landed in. This is the axis's
+    /// denominator, and [`finish`](Self::finish) refuses a zero one, so an axis
+    /// that compared nothing cannot report a green `0 of 0` indistinguishable
+    /// from a clean run. Same honest-zero discipline `spec_conformance_axis.rs`
+    /// enforces, which this module did not have.
+    ///
+    /// **What it catches is a CORPUS failure, not a provider one**, and stating
+    /// it the other way round would bake a wrong authority into the message a
+    /// reader next sees fire. A provider serving nothing does *not* reach this
+    /// predicate: `normalize` returns `Err`, `record` is still called (so
+    /// `recorded > 0`), and the rows land in `fail` — the run is red for the
+    /// ordinary reason. `skipped += 1` happens only when the case carries no
+    /// value for this axis, *before* `record`, and `catch_panics` converts a
+    /// panic into an `Err` that is still recorded. The reachable zero is a
+    /// corpus with no `to_test` row carrying a non-null value for the axis — a
+    /// schema change that silently nulled the field, or a bad regeneration.
+    recorded: usize,
     /// Count of rows for which ferro produced an **empty / degenerate
     /// projection** (tagged by [`EMPTY_PROJECTION_SENTINEL`]) — an
     /// output-quality signal independent of which bucket the row lands in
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #292, opened 2026-05-17
fix(normalize): extend codon-frame exception to r., 3+ chains, and sub+del (closes #275)

nh13:
## Summary

Extends the HGVS codon-frame exception (from #79 / PR #104) beyond the original two-`c.`-SNV case to the three sibling shapes the spec rationale also covers.

- **`r.` coding regions** — was `c.`-only. RNA shares the CDS-relative coordinate axis in this codebase (`RnaPos` doc, `rna_to_tx_pos`), so the rewrite path now branches on `Region::Cds | Region::Rna`.
- **Three-or-more-SNV chains across a codon** — the pair-detector now admits chain extension via the `prev_a.start <= prev_a.end` relaxation.
- **`sub`+`del` (and `del`+`sub`) pairs separated by one unchanged nucleotide** — the next-anchor gate now accepts `next_anchor.alt.len() <= 1`.

`build_split_variants`' `codon_frame_aware` flag fires for both `Cds` and `Rna` so the post-decompose triplet preservation path also covers `r.`. `lookup_cds_middle_ref` is renamed `lookup_codon_middle_ref` and reads CDS-axis bytes for both head variants.

## Tests

`tests/issue_275_codon_frame_extensions.rs` — 16 tests pinning the three sub-items plus negative guards (codon-straddling chains, intronic offsets, 5′UTR / 3′UTR positions, minus-strand fixture). `tests/issue_165_delins_sub_only_decompose.rs` is updated to reflect that RNA now goes through codon-frame.

## Notable

While implementing this, a latent defect surfaced in `fetch_ref_for_canonical_split`'s `r.` branch (does not translate through `cds_start`). Pre-existing — does not affect this PR's correctness for `cds_start = 1` fixtures — and tracked separately as #291.

## Out of scope

- Hardening `fetch_ref_for_canonical_split` for `cds_start != 1` — tracked as #291.

## Review

Two-reviewer cycle (Opus + Sonnet) before push; feedback addressed in this commit.

## Test plan

- [x] `cargo nextest run --features dev` — 4750 pass / 0 fail / 51 skipped.
- [x] `cargo clippy --features dev --all-targets -- -D warnings` — clean.
- [x] `cargo fmt --all` — clean.

Closes #275. Follow-up to #104 / #79.

PR #104, opened 2026-05-03
fix(normalize): codon-frame exception for c. SNV pairs separated by one nucleotide (#79)

nh13:
Closes #79. Stacked on #103 (issue #89 — region-tagged merge keys).

## TL;DR

The HGVS delins spec carves out a single exception to the strictly-consecutive merge rule: two `c.` exonic SNVs separated by exactly one nucleotide that fall in the same codon merge into a single delins, with the unchanged middle reference base preserved verbatim. PR #80 (issue #72) deferred this to #79. This PR adds the codon-frame branch to `merge_consecutive_edits`, threading a `ReferenceProvider` through the merge pass to look up the unchanged middle base.

Spec source (`hgvs-nomenclature.org/stable/recommendations/DNA/delins/`):

> Notes: "two variants separated by one or more nucleotides should preferably be described individually and not as a 'delins'. Exception: two variants separated by one nucleotide, together affecting one amino acid, should be described as a 'delins'."

Concrete spec example: `c.[145C>T;147C>G]` → `c.145_147delinsTGG`.

## Eligibility

After the strictly-consecutive check fails, the chain logic tries codon-frame merge. Both endpoints must satisfy:

| Constraint | Why |
|---|---|
| Same accession + same `HgvsVariant` discriminant | Existing rule from #72; reused. |
| `prev_a.region == Region::Cds` and `next_region == Region::Cds` | The carve-out is c.-only and CDS-only. UTR / `n.` / `r.` / `g.` all decline. |
| `prev_a.end + 2 == next_start` on the signed axis | Gap of exactly one nucleotide. |
| `(prev_a.end - 1) / 3 == (next_start - 1) / 3` | Same 1-indexed codon. |
| `prev_a.alt.len() == 1 && prev_a.start == prev_a.end` | `prev_a` is still a single-base SUB anchor — the carve-out is for SUB pairs, not for absorbing a third SUB onto a delins that's already grown via earlier merging. |
| `next_anchor.alt.len() == 1 && next_anchor.start == next_anchor.end` | Same constraint on the right side of the pair. |

When all match, `lookup_cds_middle_ref(provider, head, prev_a.end + 1)` fetches the unchanged middle base from the head variant's transcript sequence. The merged alt is `<altA> + <middle ref> + <altC>` and the anchor's range extends to `[prev_a.start, next_anchor.end]`.

[...]

PR #1720, opened 2026-08-12
fix(normalize): test the span the codon-frame merge authorises, not one edge

### LINKED ISSUE #1716: codon-frame merge tests the wrong edge: a net-length-changing pair merges across a codon boundary on the shipping arm
The per-member codon-frame merge tests the wrong edge for codon membership, so a net-length-changing pair one nucleotide apart is merged into a `delins` whose span crosses a codon boundary. `general.md:35`'s exception requires the two variants to "together affect one amino acid"; these do not, so nothing licenses the merge and `general.md:34` / `DNA/delins.md:17` govern.

**This is on the shipping arm** — `FERRO_PARTITION` unset, prepared reference — and predates the canonical-partition work entirely.

## Reproducers, on the prepared reference

```
LRG_199t1:c.[18_19del;21A>C]  ->  LRG_199t1:c.18_21delinsTC     spans codons 6 and 7
LRG_199t1:c.[30_31del;33T>G]  ->  LRG_199t1:c.30_33delinsAG
```

Same behaviour on `NM_004006.2`. Each output carries `MEMBERS_COALESCED_FROM_REPORTED_FORM`, so the merge is *disclosed* — but disclosure does not license it. Under `README.md`'s ruleset, output that violates rule 1 is a bug rather than a disclosure.

## The defect is a single term

`src/normalize/merge.rs:192` tests

```rust
same_codon(prev_a.end, next_start)
```

— the **right** edge of the left member — while the `delins` it authorises spans `prev_a.start ..= next_anchor.end`. Whenever `prev_a.start < prev_a.end` the merged span is at least 4 positions and therefore cannot lie in one codon, yet the test passes because it only ever looked at one endpoint.

Substituting `prev_a.start` for `prev_a.end`, changing nothing else, takes the measured population from 16 merges to 14 — dropping exactly these two rows and no others. The sibling passes (`apply_coding_codon_exception`, `coalesce_coding_frame_separation`) key on the span correctly.

## Why it went unnoticed for three months

Two independent reasons, both worth recording so a fix does not re-open them.

**The term was harmless when written.** #104's original predicate required `prev_a.start == prev_a.end`, under which `prev_a.end` and `prev_a.start` are the same position. **#292 (`8c3732ca`) relaxed that to `prev_a.start <= prev_a.end`** to allow chain extension, and the endpoint choice silently stopped being equivalent. So the naive fix — swapping the term — may narrow chain extension, and its blast radius is unmeasured. That is the thing to measure before shipping a fix, not after.

**The codon-straddle guard is structurally blind to this shape.** `build_split_variants` (`src/normalize/mod.rs:12434`) splits a straddling merge back apart, but it matches a `[Sub; Identity; Sub]` **column** pattern. A net-length-changing pair has no identity column, so the guard never sees it. Side by side on real data:

```
c.[18A>C;19G>T;21A>C]  ->  c.[18_19inv;21A>C]     split back — guard fires
c.[18_19del;21A>C]     ->  c.18_21delinsTC        escapes — no identity column
```

## Attribution, measured rather than read

Producer isolated by knocking out each merge pass singly, at `origin/main` = `a81eb58c`, `FERRO_PARTITION` unset:

| knockout | merges of 340 rows |
|---|---:|
| none | 16 |
| `apply_coding_codon_exception` (`merge.rs:6900`) | 16 (−0) |
| `coalesce_coding_frame_separation` (`merge.rs:7173`) | 16 (−0) |
| `coalesce_whole_block_inversion` (`merge.rs:5282`) | 15 (−1) |
| **`merge_consecutive_edits` codon-frame (`merge.rs:188`)** | **2 (−14)** |
| all three producers | 0 (control) |

`apply_coding_codon_exception` produces **none** of them — an earlier note attributing this population to it is wrong.

## Scope: 2 of 16, not 16

Of the 16 coding-axis merges at separation one on `live`:

- **8 licensed** — two lone substitutions in one codon, net length 0. The spec's own shape; `DNA/delins.md:42` marks the *split* `class="invalid"`.
- **1 licensed** — `c.[9_12inv;14_17inv;19_22inv]` -> `c.9_22inv`, where `c.9_22` was verified independently to be its own reverse complement.
- **5 unresolved** — a frameshift *inside* one codon (`c.[52del;54del]` -> `c.52_54delinsA`). `codon-carve-out-shape-restriction` states the constraint and records it unimplemented and left open; not settled here.
- **2 — this issue.**

## Note on clause strength

Whether `general.md:34`'s "should ... and **not**" carries rule-1 prohibition force or rule-2 preference force is a strength question no record settles, and `DNA/delins.md:81` restates the same rule with "preferably". That affects whether these two rows *block* a release or are *disclosed*. What is not in doubt either way is that **no clause licenses the merge** — the exception's second conjunct fails under every reading of it, because the merged span demonstrably crosses two codons.

nh13:
Closes #1716

Representation-Change: 6 of the 1,221 coding separation-one corpus rows move, and 0 of the 592 real multi-member cis alleles harvested from 9,949,738 ClinVar/CMRG/Paraphase rows. Every moved row is a merge whose `delins` span crossed a codon boundary; the merge is withdrawn and the two variants are described individually per `general.md:34`. The real-corpus zero is a property of that corpus rather than of the change — the reproducers are real transcripts (`NM_004006.2` / `LRG_199t1`), the shape just does not occur in ClinVar.

## The defect

`merge_consecutive_edits`' per-member codon-frame predicate (`src/normalize/merge.rs`) asked

```rust
same_codon(prev_a.end, next_start)
```

— the **right** edge of the left member — while the `delins` it authorises spans `prev_a.start ..= next_anchor.end`. A codon is three positions, so whenever `prev_a.start < prev_a.end` the merged span is four or more and cannot lie in one codon. `general.md:35`'s exception has two conjuncts and its second one ("together affecting one amino acid") therefore fails under every reading, so nothing licenses the merge and `general.md:34` / `DNA/delins.md:17` govern.

On the shipping arm (`FERRO_PARTITION` unset), prepared reference:

```
LRG_199t1:c.[18_19del;21A>C]  ->  c.18_21delinsTC     codon 6 + codon 7
LRG_199t1:c.[30_31del;33T>G]  ->  c.30_33delinsAG     codon 10 + codon 11
```

The fix asks the span instead — `same_codon(prev_a.start, next_start)` — which is the idiom the two sibling passes already use (`apply_coding_codon_exception` keys off the left piece's start; `coalesce_coding_frame_separation` tests `pair[0].ref_start` against `pair[1].ref_end - 1`).

## Why this does not narrow #292's chain extension

The term was harmless when #104 wrote it under `prev_a.start == prev_a.end`; #292 (`8c3732ca`) relaxed that to `<=` to admit chain extension, and only then did the endpoint choice stop being equivalent. Given `prev_a.end + 2 == next_start`, `same_codon(prev_a.start, next_start)` can hold only when `prev_a.start == prev_a.end`, so on paper the change reverts #292's relaxation for this predicate.

It does not revert its **output**, and that is measured rather than argued. A chain grown by strict adjacency is length-preserving, so the sequence-first canonicalizer re-derives it and `build_split_variants` re-cuts it on the codon boundary — and #1524 already re-blessed those outputs to the un-merged form (`c.[3G>T;4C>G;6A>T]` -> `c.[3_4delinsTG;6A>T]`, `c.[3_4delinsCG;6A>T]` -> itself). So the merge and the decline emit the same string. **All 16 `issue_275_codon_frame_extensions` tests pass unchanged**, including both chain-extension cases, and on the real reference `LRG_199t1:c.[18A>C;19G>T;21A>C]` still gives `c.[18_19inv;21A>C]`.

`prev_a.start <= prev_a.end` is kept and is still load-bearing: an insertion anchor has `start == end + 1 == next_start - 1`, which would land in one codon two thirds of the time.

## Blast radius, measured

Base `origin/main` = `2f4e3bb9`, `FERRO_MANIFEST` set on every run, freshly built in this worktree.

**Corpus census** — coding-shape `RowKind::Family` rows at separation exactly one, counting rows whose normalized output has fewer members than the authored spelling:

| direction | before | after |
|---|---:|---:|
| 3' | 450 of 1,221 | **444** |
| 5' | 433 of 1,221 | **429** |

Six distinct rows drop and **nothing is added**. All six are the same shape — a left member wider than one base and a merged span crossing a codon boundary:

```
s01-c1-pair-del-sub-p2-sep1     c.[9_10del;12C>G]        -> c.9_12delinsAG     codons 3+4
s01-c1-pair-sub-del-p2-sep1     c.[9T>A;11_12del]        -> c.9_12delinsAG     codons 3+4
s01-c3m-pair-del-sub-p2-sep1    c.[24_25del;27T>A]       -> c.24_27delinsAA    codons 8+9
s01-c3p-pair-del-sub-p2-sep1    c.[24_25del;27T>A]       -> c.24_27delinsAA    codons 8+9
s00-c1-pair-delins-del-p2-sep1  c.[9_10delinsAC;12_13del] -> c.9_12delinsAC    codons 3+4
s00-c1-mid-cds-delins-del-p2-sep1  (same variant, second family)
```

**The licensed population is untouched**, verified row by row in the same census: the eight two-substitutions-in-one-codon rows (`c.[52G>T;54C>G]` -> `c.52_54delinsTAG`, `c.[52A>C;54T>A]` -> `c.52_54delinsCAA`) still merge, and so does the whole-block inversion `c.[9_12inv;14_17inv;19_22inv]` -> `c.9_22inv`. So does the frameshift-inside-one-codon class (`c.[52del;54del]` -> `c.52_54delinsA`, `c.[52delinsT;54del]` -> `c.52_54delinsTA`) — `codon-carve-out-shape-restriction` records that question as unimplemented and this change does not answer it.

**Census pins re-blessed, and `converged` moves in neither direction.** Both `spec_conformance_axis` censuses and both `cis_confluence_axis` censuses shift rank-2 figures only. The full divergence row-id lists were dumped either side (the `divergences` / `worst` caps raised locally, then restored) and diffed by class id:

- **0 families lose convergence, 0 gain it, 0 become divergent that were not**, in any of the four censuses. The divergent sets are identical (conformance 3': 2,480; 5': 2,654; confluence 3': 2,911; 5': 2,905).
- Only three family instances change arity per conformance direction: 3' `s00-c1-m4-all-del-p2-sep1` 4→3 (improves) and `s00-c3{p,m}-m4-all-del-p1-sep2` 2→3; 5' the same three, all improving (3→2 and 4→3).
- Confluence: exactly one class, `s03-c-m3-sep1-p8-rot2`, 2→3 in both directions. What it stopped agreeing with is `c.9_18delinsATGAGTAAAGTCCTGGCA` — a merge across codons 3 to 6. Agreement is lost, correctness is not.

Every rank-1 counter is unchanged in both directions, as are `non_idempotent_outputs`, `sequence_changed`, all three refusal counters and `guard_violations`.

The two rows that rise in arity at 3' are already promoted — `spec_corpus_regressions::the_codon_gate_splits_a_spanning_delins_its_own_members_do_not` has pinned `s00-c3{p,m}-m4-all-del-p1-sep2` since #1599. Its *members'* side moved, toward that test's own clause: the old `c.[24del;30_33delinsA]` spanned codon 10 into codon 11, and is now `c.[24del;30_31del;33del]`. The row and the reason are updated there.

**Real corpus.** All 592 multi-member cis alleles in `tests/fixtures/cis/multi_member_cis_alleles.json` — the complete set of real-world inputs that reach `merge_consecutive_edits`, harvested from 9,949,738 ClinVar/CMRG/Paraphase rows — normalized through `ferro normalize -i <inputs> --reference <dir> -f tsv` on both revisions: **0 rows move**, 111 pre-existing normalize errors either side. Single-member inputs cannot reach this pass at all (it is called only from `normalize_allele`).

## Tests

`tests/it/issue_1716_codon_frame_merged_span.rs` is hermetic — the fixture is the first 210 nt of the real DMD CDS with the 5'UTR trimmed, so every `c.` coordinate is the real one and the guard needs no `FERRO_MANIFEST` and cannot skip. It pins both reproducers by exact string, and three rows that must not move: the spec's own `c.145_147delinsTGG` worked example (`DNA/delins.md:37` — `:42` is the NOTE marking the *split* invalid, and the module cites each for its own half now), #292's chain extension, and the frameshift-in-one-codon shape.

**Those three are end-to-end pins on the emitted string and none of them guards the predicate**, which the module now says for all three rather than for the chain extension alone. Measured: with `codon_frame_eligible` forced `false` — the codon-frame arm disabled outright — all three still **pass**, because the sequence-first canonicalizer re-derives those strings from the resulting sequence. They are worth having as composite-output pins; they are not evidence about this arm.

### The arm's two call-site conjuncts now have guards, beside the predicate

Mutation found that `prev_a.start <= prev_a.end` (which keeps an insertion anchor out of the arm) and the 1-position `next_anchor` width check were detected by **nothing** — not by any of the 582 tests in the `issue_1716 + issue_275 + codon + merge + spec_corpus_regressions` selection. Both are documented at length in the comment this PR rewrites, and this change makes the first of them roughly twice as load-bearing (`same_codon(start, next_start)` holds for two frames in three where `same_codon(end, next_start)` held for one).

They could not be closed end to end, and that was measured rather than assumed: a probe of eighteen insertion-anchored and multi-position-`next` descriptions through `Normalizer::normalize` returns **byte-identical output with and without each conjunct** — `c.[19_20insG;21A>C]` normalizes to `c.20_21delinsGTC` either way, because the canonicalizer re-derives it. So the guards are unit tests over `merge_consecutive_edits` itself, in `normalize::merge`'s own `mod tests`:

| test | kills |
|---|---|
| `an_insertion_anchor_is_refused_by_the_codon_frame_arm` | dropping `&& prev_a.start <= prev_a.end` |
| `a_multi_position_next_anchor_is_refused_by_the_codon_frame_arm` | dropping the width check — two cases, one per disjunct (`start != end`, `alt.len() > 1`) |
| `the_codon_frame_arm_still_merges_the_shape_it_is_for` | the positive control: without it the two above are satisfied by an arm that never merges |

Re-run after adding them, each mutation applied to a `cp` backup and restored:

| mutation | before | after |
|---|---|---|
| drop `prev_a.start <= prev_a.end` | 0 of 582 fire | **FAIL** `an_insertion_anchor_is_refused_by_the_codon_frame_arm` (merged to `c.20_21delinsGTC`) |
| drop the whole width check | 0 of 582 fire | **FAIL** `a_multi_position_next_anchor_is_refused_by_the_codon_frame_arm` (merged to `c.19_22delinsTT`, a four-position span) |
| drop only `\|\| next_anchor.alt.len() > 1` | — | **FAIL** the same test's second case (merged to `c.19_21delinsTTCC`) |
| `codon_frame_eligible -> false` | 1 of 582 fires | **FAIL** `the_codon_frame_arm_still_merges_the_shape_it_is_for` — so the positive control is not vacuous |

Also corrected: the re-pinned message in `spec_corpus_regressions` claimed "every merged span inside one codon" while its own pinned value contains `c.30_31del`, which straddles codons 10/11. That span is the **strict-adjacency** arm's, which `general.md:35` never gated; the message now says so.

## Verification

- `cargo nextest run --features dev` — **10,403 passed, 0 failed, 152 skipped**, `FERRO_MANIFEST` set, no `not run due to signal` in the log.
- `cargo clippy --features dev --all-targets -- -D warnings` — clean.
- Both spec artifacts regenerated (`--bin generate_spec_fixture`, `--bin generate_spec_enumeration`) before every test run.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -173,13 +173,41 @@ pub(crate) fn merge_consecutive_edits<P: ReferenceProvider>(
             //     share a codon-relative axis. `g.` / `n.` / UTR / intron
             //     are excluded.
             //   * `prev_a.end + 2 == next_start` (gap of exactly 1 nt).
-            //   * `same_codon(prev_a.end, next_start)` — both endpoints
-            //     fall in the same 1-indexed codon. The right edge of
-            //     `prev_a` is what matters, so a chain that has already
-            //     grown via earlier strict merges can still extend via
-            //     codon-frame (issue #275 item 2).
+            //   * `same_codon(prev_a.start, next_start)` — the two ends of
+            //     the span this merge authorises, `prev_a.start ..=
+            //     next_anchor.end`, fall in the same 1-indexed codon. That
+            //     is `general.md:35`'s **second** conjunct, "together
+            //     affecting one amino acid", and it has to be asked of the
+            //     whole merged span rather than of one endpoint: a codon is
+            //     three positions, so any span of four or more crosses a
+            //     boundary however its right edge lands.
+            //
+            //     It tested `prev_a.end` until #1716. Under #104's original
+            //     `prev_a.start == prev_a.end` the two were the same
+            //     position; #275/#292 relaxed that to `<=` for chain
+            //     extension and the endpoint choice stopped being
+            //     equivalent, so a left member wider than one base merged
+            //     across a codon boundary — `c.[18_19del;21A>C]` ->
+            //     `c.18_21delinsTC` on `NM_004006.2`, which no clause
+            //     licenses and `general.md:34` / `DNA/delins.md:17` then
+            //     govern. The two sibling passes always keyed on the span:
+            //     [`apply_coding_codon_exception`] tests the triplet from
+            //     the left piece's start, [`coalesce_coding_frame_separation`]
+            //     tests `pair[0].ref_start` against `pair[1].ref_end - 1`.
+            //
+            //     #275 item 2's chain extension is not lost by this. A chain
+            //     grown by strict adjacency is length-preserving, so the
+            //     sequence-first canonicalizer re-derives it and
+            //     `build_split_variants` puts it back on the codon boundary;
+            //     since #1524 re-blessed those outputs (`c.[3G>T;4C>G;6A>T]`
+            //     -> `c.[3_4delinsTG;6A>T]`) the merge and the decline emit
+            //     the same string. `issue_275_codon_frame_extensions` pins
+            //     both, unchanged.
             //   * `prev_a.start <= prev_a.end` — `prev_a` is not an
-            //     insertion anchor (those use `start == end + 1`).
+            //     insertion anchor (those use `start == end + 1`). Still
+            //     load-bearing after the change above: an insertion anchor
+            //     has `start == next_start - 1`, which lands in one codon
+            //     two thirds of the time.
             //   * `next_anchor` is a 1-position substitution OR deletion
             //     (checked below): `next_anchor.start == next_anchor.end`
             //     and `next_anchor.alt.len() <= 1`. Issue #275 item 3
@@ -189,7 +217,7 @@ pub(crate) fn merge_consecutive_edits<P: ReferenceProvider>(
                 && (prev_a.region == Region::Cds || prev_a.region == Region::Rna)
                 && prev_a.end.checked_add(2) == Some(next_start)
                 && prev_a.start <= prev_a.end
-                && same_codon(prev_a.end, next_start);
+                && same_codon(prev_a.start, next_start);
 
             if !strict_adjacent && !codon_frame_eligible {
                 break 'try_merge false;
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

PR #1724, opened 2026-08-12
fix(reference): never trade a cdot exon table for a supplemental record that carries nothing

### LINKED ISSUE #1722: reference: a gapped cdot table is discarded for a supplemental record that carries nothing — 50 accessions served with no exons and no CDS
## Summary

`MultiFastaProvider` discards a complete cdot exon table in favour of a supplemental record that carries nothing, so **50 real accessions are served with `exons: []` and `cds_start: None`** — no `c.` axis, no genomic projection, no splice handling.

```rust
// src/reference/multi_fasta.rs:2588-2597
let has_gaps = exons.windows(2).any(|w| w[1].start != w[0].end + 1);
if has_gaps && self.supplemental_cds.transcripts.contains_key(&resolved) {
    warn!("cdot exon data for {} has gaps in transcript coordinates, \
           using supplemental CDS metadata", resolved);
    self.get_supplemental_cds_info(&resolved)   // <- may carry nothing
} else { /* keep the cdot table */ }
```

The guard tests the supplemental record for **presence**, not usability. `get_supplemental_cds_info` then keys its synthetic exon on a field the record does not have:

```rust
// src/reference/multi_fasta.rs:1760-1770
let exons = if info.sequence_length > 0 { vec![Exon { .. }] } else { Vec::new() };
```

## Measured against a prepared reference

```
TXBRANCH resolved=NM_033517.1  cdot_exons=22  has_gaps=true  supp_present=true
                               supp_seqlen=Some(0)  supp_cds=Some(None)
   -> served: exons=0  cds_start=None  seq_len=7113
```

| | |
|---|---|
| gapped multi-exon GRCh38 cdot builds | 58 of 474,818 (0.0122%) |
| served with zero exons and no CDS | 53 |
| of those, reachable (no version substitution, so the #785 gate never fires) | **50** |
| cdot exon counts for those records | 10–28 exons each, present and exact |

Affected loci span six chromosomes and include `MUC2`, `CASP8AP2`, `NBPF1`, `SHANK3` (`NM_033517.1`) and the `NM_001405***` family.

**Control:** `NR_197427.1` is the one gapped record with *no* supplemental entry. It could not take the fallback, and it is the one record that kept its 5 cdot exons. The fallback is the cause, not the gap.

## Root cause of the hollow records

`supplemental/patterns_transcripts.metadata.json` (140,027 records) is degraded: `sequence_length != 0` in 23,455 (16.8%), `cds_start` non-null in **19** (0.014%). Of 96,739 `NM_` records exactly one has a CDS.

```json
"NM_000088.3": {"gene_symbol":"COL1A1","cds_start":null,"cds_end":null,"sequence_length":0}
```

The sibling FASTA has the bases — `patterns_transcripts.fna.fai` carries `NM_000088.3 5927`. Ruled out: deserialization (hand-rolled key extraction at `multi_fasta.rs:1165-1183`, keys match, no `serde(default)`), intended shape (19 records in the same file *do* carry a CDS; the sibling `legacy_genbank.metadata.json` is 5/5 populated), and the current writer (all three insert sites write `seq.len()` guarded non-empty).

## Why it cannot heal itself

- `src/benchmark/cache.rs:2019` — resume filters on **key presence**, not data quality, so a hollow record permanently excludes its accession from re-fetch.
- `src/benchmark/cache.rs:2004` — a parse failure silently resets the whole file to empty.

## Proposed fix

A fallback must never be a downgrade.

1. **`:2590` must compare, not reflex.** Divert only when the supplemental record carries a CDS — the one thing it can authoritatively add. Otherwise keep the cdot table: gapped but real is strictly better than absent. (What to do about the gap is #1619's question, not this one.)
2. **Remove the dependency on the bad field.** `get_supplemental_cds_info` builds a synthetic exon spanning the whole sequence, so it never needed the metadata length — `entry.length` is in scope at every call site and the FAI is authoritative. Sourcing the span from the index makes the empty-exon branch unreachable.
3. **Make the degradation visible at load** rather than silently inserting information-free records.
4. **Ingestion (separate change):** re-fetch a hollow record instead of skipping on key presence; surface a parse error instead of resetting to empty.

## Coupling

Restoring these tables gives `NM_033517.1` back its 22 exons *and its real 39-base gap*, which makes #1619's flat-vs-walk divergence reachable on a prepared reference rather than only on the test fixture. That is the right order, but #1619 becomes more urgent once this lands.

## Test coverage

None. No test constructs `SupplementalTranscriptInfo`; no fixture writes a supplemental `*.metadata.json` with non-zero values. Both the divert branch and the `Vec::new()` branch are untested.

nh13:
Closes #1722

A gapped cdot exon table was discarded in favour of a supplemental record tested only for **presence**, not for whether it could replace the table. The supplemental artifact carries a CDS on **19 of its 140,027** rows, so presence was true almost everywhere and usability almost nowhere.

## Measured, before and after, against a prepared reference

```
before: 53 untestable (0 exons), 4 contiguous,  0 still gapped
after :  0 untestable (0 exons), 4 contiguous, 53 still gapped
```

`NM_033517.1` (SHANK3) goes from `exons=0, cds_start=None` to its real 22 exons and `cds_start=1`. Of the 53, **50** are reachable — the other 3 are version-substituted, so #785's gate declines them before this path matters.

**The control that identifies the cause:** `NR_197427.1` is the one gapped record with no supplemental entry at all. It could not take the divert, and it is the one record that kept its exons. The fallback was the cause, not the gap.

## Three layers, so the class cannot recur

| layer | change |
|---|---|
| the question | `SupplementalTranscriptInfo::can_replace_cdot_table` states what the map can authoritatively add — a curated CDS — and the divert asks *that* instead of `contains_key` |
| the dependency | `get_supplemental_cds_info` takes the span from the **FASTA index** rather than the metadata's `sequence_length`, which is `0` on 83% of rows. The synthetic exon covers the whole transcript by construction, so the metadata never needed to carry it — and the `Vec::new()` branch is now unreachable |
| the visibility | the loader reports how much of the artifact is inert and warns when a supplemental file is largely CDS-less. The old line reported the row count and called every row *"with CDS metadata"* |

The guiding rule: **a fallback must never be a downgrade.** A gapped table is real annotation and is strictly more useful than none.

## Tests

Two new unit tests, written to fail first (`left: 0, right: 2` before the fix):

- `a_gapped_cdot_table_survives_a_supplemental_record_carrying_nothing`
- `a_supplemental_record_with_a_cds_still_replaces_a_gapped_cdot_table` — the other half of the contract, so this narrows the guard rather than removing it

Both branches were previously untested; no test anywhere constructed `SupplementalTranscriptInfo`.

Full suite: **10,209 passed, 0 failed.**

It does not decide how a gapped table should be *read*. Restoring these tables gives `NM_033517.1` back its 22 exons **and its real 39-base gap**, which makes #1619's flat-vs-walk divergence reachable on a prepared reference rather than only on the test fixture. Measured on the restored data:

| accession | walked | flat | delta |
|---|---:|---:|---:|
| `NM_033517.1` | 1399 | 1351 | 48 |
| `NM_012115.3` | 6240 | 6185 | 55 |
| `NM_002457.4` | 11442 | 8290 | 3152 |

That is the evidence #1665's closing comment asked for, and it belongs to #1619, not here. Getting a correct table is the prerequisite for arguing about how to read it.

## Follow-up, not in this PR

The artifact cannot heal itself: `src/benchmark/cache.rs:2019` filters re-fetch on **key presence** rather than data quality, so a hollow record permanently excludes its accession; `:2004` silently resets the whole file to empty on a parse error. Neither helps an already-shipped reference, which is why the guard could not wait on them.

Representation-Change: 50 accessions move from declining ("Transcript has no CDS") to served on the c./n. axes. Previously-erroring inputs now produce output, so this is a real change for a consumer that keyed on the decline.

### DIFF
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -82,6 +82,62 @@ pub struct SupplementalTranscriptInfo {
     pub sequence_length: u64,
 }
 
+impl SupplementalTranscriptInfo {
+    /// Whether this record can stand in for a cdot exon table.
+    ///
+    /// The supplemental map exists to supply a **curated CDS** for transcripts cdot
+    /// lacks or gets wrong. That is the only thing it can authoritatively add: the
+    /// bases come from the FASTA index and the exon structure comes from cdot. So a
+    /// record with no CDS cannot replace a cdot table — diverting to it trades a
+    /// real exon list for nothing.
+    ///
+    /// This exists because the divert used to test the map for *presence*
+    /// (`contains_key`), which is not the same question. Measured against a prepared
+    /// reference, `patterns_transcripts.metadata.json` carries a CDS on **19** of its
+    /// 140,027 records, so presence was true almost everywhere and usability almost
+    /// nowhere: 50 accessions whose cdot tables were complete (10-28 exons) were
+    /// served with `exons: []` and `cds_start: None` (#1722).
+    pub fn can_replace_cdot_table(&self) -> bool {
+        self.cds_start.is_some() && self.cds_end.is_some()
+    }
+
+    /// Whether the record carries a CDS bound or a length — the two fields any
+    /// transcript *geometry* could be built from.
+    ///
+    /// Used at load to report how much of a supplemental artifact supplies neither,
+    /// rather than inserting such rows and letting `contains_key` imply they mean
+    /// something.
+    ///
+    /// **`gene_symbol` is deliberately excluded, which is why this is not named
+    /// "is informative".** It is not a dead field: [`Self::can_replace_cdot_table`]'s
+    /// sibling path copies it onto the served transcript
+    /// (`get_supplemental_cds_info`), and for an accession cdot does not hold it is
+    /// the only source of the symbol. Measured against the shipped
+    /// `patterns_transcripts.metadata.json`: **every one** of the 116,572 records
+    /// this returns `false` for carries a `gene_symbol` (116,572 of 116,572), so
+    /// calling that population "inert" would be the same presence-vs-usability
+    /// conflation this type exists to stop, run the other way.
+    pub fn carries_cds_or_length(&self) -> bool {
+        self.cds_start.is_some() || self.cds_end.is_some() || self.sequence_length > 0
+    }
+}
+
+/// A supplemental artifact is reported as largely CDS-free when fewer than one
+/// record in [`LARGELY_WITHOUT_CDS_RATIO`] carries a usable CDS.
+///
+/// Extracted so the threshold is testable without going through a manifest load:
+/// the branch it gates is a `warn!`, and a log line is not something a unit test
+/// can assert on.
+const LARGELY_WITHOUT_CDS_RATIO: usize = 10;
+
+/// Whether a supplemental artifact carries so few CDS records that relying on it
+/// for CDS metadata is a data defect rather than a normal state.
+///
+/// Returns `false` for an empty artifact — there is no population to describe.
+fn artifact_is_largely_without_cds(with_cds: usize, total: usize) -> bool {
+    total > 0 && with_cds.saturating_mul(LARGELY_WITHOUT_CDS_RATIO) < total
+}
+
 /// Supplemental transcript metadata for transcripts not in cdot.
 #[derive(Debug, Clone, Default)]
 pub struct SupplementalCdsInfo {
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1725, opened 2026-08-12
docs(rulings): record general.md:34 as a preference, not a prohibition

nh13:
Records how strongly `general.md:34` speaks. This is a **classification** change with no behaviour change: no `src/` file is touched.

## The question

> two variants separated by one or more nucleotides **should** be described individually and **not** as a "delins".

Is that a README **rule 1** (Conformant, *absolute — never traded*) because of the negation, or **rule 2** (Recommended form, *best effort*) because of the modal? It matters because rule 7 makes an output violating rules 1–4 a **bug** rather than a disclosure, so the answer decides whether a whole class of outputs blocks a release or is disclosed.

## The ruling: rule 2, whole clause

The general reading, which is what makes the record reusable beyond this clause:

> **"and not Y" names the excluded alternative; it does not grade the clause. The modal grades the clause.**

**1. Internal to the clause.** The tempting reading splits it — "should be described individually" as rule 2, "and not as a 'delins'" as rule 1. That is incoherent, because the halves are **complements**: for two variants separated by one or more nucleotides, "described individually" and "described as a delins" exhaust the available forms. If the delins form were forbidden, the individual form would be the only permitted one — mandatory — and "should" would be doing no work at all. A clause cannot grade its positive half as a preference while forbidding the sole alternative to it.

**2. The spec pairs the construction with both modals.** "and not" occurs 14 times in `docs/recommendations/` at checkout `6f85311`: twelve take "should", exactly two take "must" (`DNA/duplication.md:18`, `RNA/duplication.md:19`). Were the negation prohibitive on its own, that "must" would be redundant. It is not — `duplication.md:18` reads "and not as, **e.g.**, an insertion", so its alternatives are *not* exhaustive and the modal does independent work there.

**3. The spec restates this very rule as a preference.** `DNA/delins.md:81`, in a Q&A block: "Yes, two variants separated by one or more nucleotides should **preferably** be described individually and not as a 'delins' (unless they together affect one amino acid)." "preferred" is named in rule 2's own scope.

**4. Every rule-1 marker is absent — checked, not assumed.** The rule is stated five times (`general.md:34`, `DNA/substitution.md:16`, `DNA/delins.md:81`, `RNA/substitution.md:17`, `protein/delins.md:21`) and never with "must", "not allowed", "can not be used" or `class="invalid"`. `general.md` contains **zero** `class="invalid"` in the whole file. `checklist.md` — rule 1's escape hatch — does not enumerate the separation rule.

**Where the spec does prohibit** is separation **zero**: `DNA/substitution.md:32` marks the split spelling `class="invalid"` and calls it "not correct" by name, which is the ground `delins-adjacent-members-when-both-consume-reference` already stands on. So the line is the spec's own, not one this record invents.

## What it does and does not license

Rule 2 **still binds** — README states a preference clause outranks maintainer judgment — so ferro remains obliged on best effort to produce the individual form. What changes is the classification: such an output is a deviation to **disclose and pin with a tripwire**, not a rule-7 bug, and it does not by itself block a release. Every negative guard measuring the deviation is kept and pinned at its **true** value with a tripwire that fails when a listed row starts conforming. Re-pinning a guard to zero for a green build is the one move the record forbids.

The record also carries **what was refuted**, so neither reading is re-derived: that the bolded "not" carries prose force under rule 1, and that the clause splits its force between its halves.

## Verification

- Both spec generators exit 0 — which is what validates all five citations against the pinned checkout.
- `ruling_records_are_intact`, all four `claude_md_adjudication_tables` tests, and all `clause_ruling_index` tests pass.
- Full suite with `FERRO_MANIFEST` set: **10,207 passed, 0 failed**, 150 skipped. `cargo clippy --features dev --all-targets -- -D warnings` clean.
- The generated index block in `clause_ruling_index.rs` was captured **programmatically** and spliced on its markers — never hand-transcribed, per that module's own warning.
- Ledger comparison against `origin/main`: 22 records vs 21, one added, **nothing dropped and no status regressed**.

Representation-Change: none. Records a classification; no `src/` file is touched and no output moves.


nh13:
@coderabbitai pause

PR #1726, opened 2026-08-12
fix(normalize): carry junction-exit provenance per leaf, from the gate that makes it

### LINKED ISSUE #1723: normalize: junction-exit re-parenting decides a per-leaf question with whole-description predicates
`reparent_junction_exit` decides a per-leaf question with two whole-description predicates, and one of them is a first-leaf `find_map`. The granularity mismatch is the defect: it both suppresses the repair on a description that needs it, and applies it to only one of several leaves that need it.

This is a follow-on to #1708 (which fixes #1704). Everything below is measured on that branch at `2c779fbc`, plus strand, using the fixture geometry already committed in `tests/it/defect_371_transcript_exit.rs`.

## What is wrong

```rust
fn reparent_junction_exit(&self, input: &HgvsVariant, output: HgvsVariant) -> HgvsVariant {
    let Some(bare) = bare_transcript_intronic_accession(&output).cloned() else { return output; };
    if names_bare_transcript_intronic(input) { return output; }
    ...
    reparent_leaves(output, &bare, &wrapper)
}
```

Three different granularities are in play for one question:

- `names_bare_transcript_intronic(input)` is an **any-leaf** existence test over the whole input.
- `bare_transcript_intronic_accession` is a `find_map` — the **first** offending leaf.
- `reparent_leaves` acts **per leaf**.

"Did ferro manufacture this intronic offset, or did the author write it?" is a fact about each leaf. It is currently approximated by two whole-description predicates and then applied per leaf.

## Measured consequences

Both are reproducible on the committed `defect_371_transcript_exit` fixture (`junction_provider(Strand::Plus, exons_with_run_at_exon1_end('A'), intron_leading_with("AA"), true)`), 3' direction.

**(a) The any-leaf bail-out suppresses the repair for a leaf that needs it.**

| input | output on `2c779fbc` | |
|---|---|---|
| `NM_TEST.1:c.7del` | `NC_SYNTH.1(NM_TEST.1):c.10+2del` | correct — manufactured, wrapped |
| `NM_TEST.1:c.30+5del` | `NM_TEST.1:c.30+5del` | correct — authored, left bare (`bare-transcript-intronic-position`) |
| `NM_TEST.1:c.[15del;30+5del]` | `NM_TEST.1:c.[30del;30+5del]` | correct — nothing manufactured |
| `NM_TEST.1:c.[7del;30+5del]` | `NM_TEST.1:c.[10+2del;30+5del]` | **defect** |
| `NM_TEST.1:c.[7del;11-5del]` | `NM_TEST.1:c.[10+2del;11-5del]` | **defect** — other side of the same intron |

In the last two, `10+2` is ferro's own manufacture and it ships on a bare accession — the exact class #1708 exists to close. The authored sibling vetoes the repair for the manufactured one.

**(b) The first-leaf `find_map` repairs only one of several leaves that need it.** #1708's own doc comment on `bare_transcript_intronic_accession` anticipates this for a `products` allele and notes that the consequence "would be silent". It is not confined to `products`, and it needs no authored intronic leaf at all — an ordinary two-accession allele does it:

```
[NM_TEST.1:c.7del];[NM_OTHER.1:c.7del]
  -> [NC_SYNTH.1(NM_TEST.1):c.10+2del];[NM_OTHER.1:c.10+2del]

[NM_TEST.1:c.7del;NM_OTHER.1:c.7del]
  -> [NC_SYNTH.1(NM_TEST.1):c.10+2del;NM_OTHER.1:c.10+2del]
```

(measured with a second transcript of identical geometry added to the fixture provider under `NM_OTHER.1`). The second leaf is a manufactured intronic offset on a bare accession. `reparent_leaves` matches on accession equality and is not run to a fixed point, so the second accession is never considered.

## The fix has two layers

**Layer 1 — provenance is a per-leaf property.** Stop approximating it with whole-description predicates.

**Layer 2 — carry the fact rather than reconstructing it.** Per-leaf provenance cannot be recovered by comparing input and output after the fact: normalization reorders, merges and splits members, so there is no identity map from an output leaf back to an input leaf. That is precisely why the proxy exists.

It does not need to be reconstructed. The #670 junction gate — three copies, `normalize_cds`, `normalize_tx`, `normalize_rna` — already knows the answer for free:

- an input with an intronic endpoint returns from `normalize_cds` at the `start_pos.is_intronic() || end_pos.is_intronic()` branch, well before the gate. So **reaching the gate implies both endpoints were exonic**, and the gate's own success condition is `crossed_into_intron`. Manufacture is therefore unconditional at that point, not a comparison.
- the gate has already resolved the genomic contig (`boundary_transcript.chromosome`) — it could not have computed the crossing otherwise. That is the same contig `reparent_junction_exit` re-derives at the top by a second provider lookup.

With the fact recorded where it is known, `reparent_junction_exit` becomes a fold over marked leaves, the any-leaf/first-leaf asymmetry has nowhere to live, and the multi-accession case in (b) falls out because each marked leaf carries its own contig.

## Also in scope: three guards with zero coverage

`src/normalize/mod.rs:3100-3105`:

```rust
if !rest.is_empty()
    || !crate::hgvs::parser::accession::is_valid_compound_outer(&outer)
    || outer == bare
{
    return output;
}
```

Forcing all three conjuncts to `false` causes zero committed-test failures. They are not dead code — with `transcript.chromosome = "NM_OTHER.1"` the mutated build emits `NM_OTHER.1(NM_TEST.1):c.10+2del`, a backwards pairing `parse_hgvs` refuses — they are simply untested. Under layer 2 they become guards on data at the gate rather than guards on a re-derivation at the top, which is also where they become straightforward to exercise.

## Explicitly NOT in scope: the compact-form policy

Even with perfect per-leaf provenance there is a real policy question, and it is not a code question. `ACC:c.[a;b]` renders compactly only when its members share an accession (`use_compact_form` / `all_share_accession_and_type`). When one leaf needs a wrapper and another does not, the choice is between lifting the wrapper to the whole description and expanding to per-member accessions — and `reparent_leaves`' own doc already rejects expansion for *exonic* members as "a far larger representation change than the defect it repairs". For an *authored intronic* sibling the calculus is different, because lifting re-spells a leaf the `bare-transcript-intronic-position` ruling says to leave as authored.

The work here is to make the code able to express either answer — today it cannot represent "wrap because of member 1 but not member 2" at all — while preserving current observable behaviour as the default. The question itself belongs in the `rulings` ledger as `undecided`.

nh13:
Closes #1723.

Follow-on to #1708, which has now **merged** (`1ea75334`); this branch is rebased onto `main` and is one commit. It was opened stacked, so earlier review comments may refer to a two-commit range — `git log --oneline origin/main..HEAD` is now a single commit. **Rebased again onto `origin/main` `57f71d31`** while applying review findings; that rebase and its three conflict resolutions are described under "Review response" below.

#1708 asks a per-leaf question with two whole-description predicates, and one of them is a first-leaf `find_map`. This makes provenance per-leaf, and carries it from the site that knows it instead of reconstructing it at the top.

## The granularity mismatch

```rust
let Some(bare) = bare_transcript_intronic_accession(&output).cloned() else { return output; };  // FIRST leaf
if names_bare_transcript_intronic(input) { return output; }                                     // ANY leaf
...
reparent_leaves(output, &bare, &wrapper)                                                        // per leaf
```

Three granularities for one question. Measured using the fixture geometry already in `tests/it/defect_371_transcript_exit.rs`:

| input | before (#1708) | this PR |
|---|---|---|
| `NM_TEST.1:c.7del` | `NC_SYNTH.1(NM_TEST.1):c.10+2del` | unchanged |
| `NM_TEST.1:c.30+5del` | `NM_TEST.1:c.30+5del` | unchanged |
| `NM_TEST.1:c.[7del;15del]` | `NC_SYNTH.1(NM_TEST.1):c.[10+2del;18del]` | unchanged |
| `NM_TEST.1:c.[7del;30+5del]` | `NM_TEST.1:c.[10+2del;30+5del]` | unchanged — now a *declared* decline, see below |
| `[NM_TEST.1:c.7del];[NM_OTHER.1:c.7del]` | `[NC_SYNTH.1(NM_TEST.1):c.10+2del];[NM_OTHER.1:c.10+2del]` | **`[NC_SYNTH.1(NM_TEST.1):c.10+2del];[NC_SYNTH.1(NM_OTHER.1):c.10+2del]`** |

The last row is the fix. The second bare accession carries a manufactured intronic offset — the exact class `#1704` exists to close — and the `find_map` never reached it. `#1704`'s own doc anticipated this for a `products` allele and called the consequence "silent"; it is **not confined to `products`**, which matters because `products` is documented as meaningful only on `r.`/`p.` and `r.` is out of this predicate's scope. An ordinary trans allele and an ordinary multi-accession cis allele both reach it.

## Provenance is carried, not reconstructed

It cannot be reconstructed: normalization reorders, merges and splits members, so no identity map from an output leaf back to an input leaf survives to the top. That is why the proxy existed.

It does not need to be. The `#670` junction gate — three copies, `normalize_cds` / `normalize_tx` / `normalize_rna` — already knows the answer **for free**:

- an input with an intronic endpoint returns at `start_pos.is_intronic() || end_pos.is_intronic()`, well before the gate. So *reaching* the gate is the statement that both endpoints were exonic, and the gate's success condition is `crossed_into_intron`. Manufacture is what the branch means; nothing is compared.
- the contig is there too — the crossing fetches genomic bases from `boundary_transcript.chromosome`, which is the same reference `#1704` re-derived at the top with a second provider lookup that could disagree with the one the coordinate was computed in.

Each gate now records a `ManufacturedJunctionExit { leaf, bare, wrapper }`, threaded to the single seam (`normalize_core_canonical`) as a `&mut Vec` out-parameter — 8 signatures, 22 call sites (the measured plumbing cost), `Vec::new()` so a normalization that manufactures nothing pays nothing. `reparent_junction_exit` becomes a fold over those records and consults the provider not at all.

The `r.` gate is wired too and records nothing, because `intronic_on_bare_transcript_axis` — the single reading of `checklist.md:20` in this crate — does not reach that axis. Deliberate: if the scope widens, the provenance follows rather than lagging.

## The three rendering guards

They move into `genomic_wrapper_for`, called from the gate, so they act on the `Transcript` the crossing actually used rather than on a second provider lookup that could disagree with it. **All three are kept**, including `outer == bare` — an earlier cut of this PR dropped it as redundant and that was wrong: `#1708`'s LRG analysis is right, `bare_transcript_intronic_*` admits a bare `LRG_<N>` by prefix while `is_valid_compound_outer` reads it as genomic, so both earlier conjuncts pass and only the third stops `LRG_5(LRG_5):c.…`. `a_self_referential_wrapper_is_declined` (from `#1708`) is unchanged and green, and the new unit test exercises the same conjunct directly on `genomic_wrapper_for`.

They also gain direct coverage they did not have: seven unit tests over `genomic_wrapper_for` and the fold, on top of `#1708`'s end-to-end guard tests, which are unchanged and still pass.

## What this deliberately does NOT decide

`ACC:c.[a;b]` renders compactly only when its members share an accession. So when one leaf needs a wrapper and a sibling does not, there is a choice — **lift** to the whole description, or **expand** to per-member accessions — and it is a representation-policy choice.

Where every bare-intronic leaf on the accession is one ferro manufactured, lifting re-spells only *exonic* siblings and `reparent_leaves` already argues that is the cheap side; that branch is unchanged. Where a sibling is an **authored** intronic position, lifting re-spells a leaf `bare-transcript-intronic-position` decided is left as authored, and expanding abandons the form `DNA/alleles.md:16` states. Neither follows from a clause, so **ferro declines — byte-for-byte what #1708 shipped** — and the question is filed as the `undecided` `junction-exit-wrapper-scope-in-a-mixed-allele` ruling.

The point is that both answers are now *expressible*. `#1704` could not represent "wrap because of member 1 but not member 2" at all, and could not tell that case apart from the multi-accession one it also got wrong.

## Mutation table

Every guard added or relied on was mutated three or more ways and the committed tests re-run. Selection: the 20 tests that `C-none` (wrapping disabled entirely) fails on a full-suite run — every test in the tree sensitive to this pass: `normalize::tests`, `defect_371_transcript_exit`, `in_bounds_oracle_scope`, `inversion_sweep`, `spec_corpus_regressions`, `spec_enumeration_tests`, `spec_conformance_axis` (222 tests).

| mutation | fired |
|---|---|
| `!leaves.is_empty()` → `true` (the natural omission) | **1** — `the_junction_exit_fold_wraps_nothing_when_no_intronic_leaf_survives` |
| `!leaves.is_empty()` → `false` | 20 |
| `!leaves.is_empty()` → `!manufactured.is_empty()` (always true here) | **1** — same test |
| per-leaf `.all(record matches)` → always true | 3 — 2 unit + `a_mixed_allele_still_ships_a_manufactured_offset_bare` |
| per-leaf `.all(record matches)` → always false | 20 |
| `.all(…)` → `.any(…)` (the `#1704` semantics restored) | 2 — `the_junction_exit_fold_declines_a_mixed_accession`, `a_mixed_allele_still_ships_a_manufactured_offset_bare` |
| repair only the first accession (`#1704`'s `find_map`) | 2 — `the_junction_exit_fold_repairs_every_recorded_accession`, `a_second_bare_accession_is_repaired_too` |
| repair no accession | 20 (this is the selection) |
| one wrapper reused for every accession | 2 — same pair |
| `genomic_wrapper_for` → never decline | 3 — the two guard tests + `a_self_referential_wrapper_is_declined` |
| … `rest.is_empty()` alone disabled | 2 |
| … `is_valid_compound_outer` alone disabled | 2 |
| … **`outer == *bare` alone disabled** | **1** — `a_self_referential_wrapper_is_declined`, plus the unit assertion this PR adds |
| `genomic_wrapper_for` → always decline | 24 |
| drop the `normalize_cds` gate recording | 18 |

**No mutation survives.** Two are worth stating rather than burying. The `!leaves.is_empty()` conjunct is covered by **exactly one** test, and it is one this PR adds — that branch is not reachable from a fixture provider, which is why it is a unit test over the fold rather than end-to-end; without it the conjunct would be vacuous. And `outer == *bare` is likewise covered by exactly one committed test, `#1708`'s `a_self_referential_wrapper_is_declined`; this PR adds a unit assertion beside it that pins the two *premises* (the bare LRG parse succeeds, and `is_valid_compound_outer` admits it), so the reason the conjunct is not subsumed is checkable rather than something the next reader has to re-derive — which is exactly what an earlier cut of this PR failed to do, dropping the conjunct on the assumption it was redundant.

## Review response, and the rebase onto current `main`

Rebased from `e98fa77e` onto `origin/main` `57f71d31`. Three files conflicted and each was resolved by re-deriving rather than by taking a side:

- **`hgvs_spec_normalization_overrides.json`** — `main` had gained two records (`derivation-may-not-be-bounded-by-the-inputs-spelling`, `c-and-n-positions-are-flat-transcript-offsets`). Resolved by keeping both and appending this branch's one, then diffing the ledger as a **set**: `main` 25 / old tip 24 / HEAD **26**, **nothing dropped from either side**, no status changed, one addition (`junction-exit-wrapper-scope-in-a-mixed-allele`).
- **`clause_ruling_index.rs`** — the census pins on this branch read `(24, 20, 4)` and `25`, which were correct against the old base and are now stale. **A census pin can merge cleanly to a wrong value**, so they were re-derived from the ledger rather than trusted: `(26, 22, 4)` from the record count, and the multiply-named-clause count from a measured run (**26 -> 27**; `checklist.md:20` is the line that gains a second record). The generated block was regenerated by capturing `the_rendered_index_is_current`'s own panic output and splicing it on the `BEGIN`/`END` markers — never hand-edited.
- **`hgvs_spec_normalization_tests.rs`** — `RULING_STATUSES`, resolved by keeping both sides' entries.

Three review findings applied:

1. **The accession->wrapper map kept the FIRST wrapper silently** — the same shape as the `find_map` this PR removes, one level in. It now **declines** the accession when two records for one bare accession carry different wrappers, and the declined state is sticky. **Not evidenced**: no provider in this tree reaches it (`Transcript::chromosome` is constant per accession in every fixture) and no input was constructed that makes two gates resolve different contigs, so this is a latent invariant written where it is relied on — the status `outer == *bare` had until LRG turned out to reach it — not a fixed bug.
2. **Records survive computations whose results are discarded** — `normalize_allele`'s warnings-only raw-conflict branch and `canonicalize_from_sequence`'s alternation loop both throw their results away and keep the records. `ManufacturedJunctionExit`'s doc now states the invariant the fold actually relies on: a record is evidence that *some* gate produced this exact leaf, deliberately not evidence about which pass survived, and value equality is what makes that safe. Also **not evidenced** — no input was constructed in which an authored intronic leaf coincides byte-for-byte with a manufactured one on the same accession.
3. **No test for two accessions taking DIFFERENT branches.** The three end-to-end shapes pinned were all-manufactured, mixed, and two-accessions-both-manufactured; the combination the loop exists for was not. `two_accessions_take_different_branches_in_one_description` pins `[NM_TEST.1:c.7del;NM_TEST.1:c.30+5del;NM_OTHER.1:c.7del]`, where `NM_TEST.1` is mixed and declines while `NM_OTHER.1` is all-manufactured and repairs, in one description and on both strands. Collapsing the loop to "decline everything if any accession is mixed" — `#1704`'s behaviour widened, and the most plausible simplification of this pass — passes every other test in the tree and fails this one.

## Verification

- Full local gate **re-run on the rebased tree** (`FERRO_MANIFEST` set, `FERRO_REQUIRE_BULK_FIXTURES=1`, all four bulk fixtures fetched and digest-verified, both spec artifacts deleted and regenerated first): `GEN1_EXIT=0 GEN2_EXIT=0 FMT_EXIT=0 CLIPPY_EXIT=0 TESTS_EXIT=0`, **10,249 run / 10,249 passed / 0 failed / 150 skipped**, 716 s, no `not run due to signal`. Attribution is by test **name**, not by count — a raw count is meaningless across a rebase that moved the base: diffing the branch's own delta at each base (`git diff <old-base>..<old-tip>` against `git diff 57f71d31..HEAD`) shows the test-function set gained exactly **one** name, `two_accessions_take_different_branches_in_one_description`, and lost none.
- `cargo clippy --all-features --all-targets -- -D warnings` clean; `cargo fmt --all`; both spec generators run (which is what validates the new ruling record's citations against the spec checkout); `scripts/run_oracle_suite.sh`, 0 fail.
- The 3' conformance census is byte-identical to `#1708`: `outputs_leaving_the_transcript: 0`, `outputs_intronic_under_a_genomic_wrapper: 371`.
- Ruling ledger compared programmatically against `origin/main` and the branch's own pre-rebase tip. **Re-run after the rebase onto `57f71d31`**: `main` 25 / old tip 24 / HEAD **26** records, **none dropped from either side**, no status changed, one addition (`junction-exit-wrapper-scope-in-a-mixed-allele`). The earlier 23 / 22 / 22 figure was taken on the previous base and is superseded.

Representation-Change: 1 shape moves, and it is a repair — a manufactured intronic offset on a SECOND bare transcript accession now carries the genomic reference `checklist.md:20` requires, where #1704 left it bare. 0 rows of any committed corpus or fixture move: the 3' census reads `outputs_leaving_the_transcript: 0` / `outputs_intronic_under_a_genomic_wrapper: 371`, byte-identical to #1704. That zero is STRUCTURAL and is a claim about the corpus, not about the change — the spec corpus's junction stratum builds single-accession descriptions and so cannot emit a two-accession allele at all; the moved shape is constructed explicitly in `defect_371_transcript_exit::a_second_bare_accession_is_repaired_too`. Every other output, including the mixed-allele residue, is byte-identical to #1704.


nh13:
@coderabbitai pause

PR #1727, opened 2026-08-12
fix(normalize): scope the compensating-gap coalesce to the coding DNA axis

### LINKED ISSUE #1711: canonical arms merge two variants across one unchanged nucleotide on frameless axes — 6 rows implementing rejected SVD-WG010
Under `FERRO_PARTITION=canonical` and `canonical-coalesced`, six rows of the spec conformance corpus collapse two variants separated by **one** unchanged nucleotide into a single spanning `delins`, on axes with **no reading frame**. `spec_conformance_axis`'s `guard_violations` counter — a negative guard for the **rejected** proposal SVD-WG010 — is pinned at `0` and reads `6` in both directions on both canonical arms.

This blocks the canonical partition flip under `README.md`'s rule 1.

## The rows

Measured at `ce933533`, `FERRO_MANIFEST` set, both spec artifacts regenerated with `--bin`, dump-line count equal to the printed census on every run (6/6 × 4). `canonical` and `canonical-coalesced` produce byte-identical sets — `payload_coalesce_applies` requires `CisKind::Cds` and every row here is frameless, so the `:47` coalesce pass contributes none of them.

| row id | axis | members | sep | block | in → out |
|---|---|---|---:|---:|---|
| `s01-n3m-pair-del-inv-p4-sep1` | `n.` NCMultiExon(Minus) | del + inv | 1 | 9 | `NR_TEST.1:n.[36_39del;41_44inv]` → `n.36_44delinsTCAGA` |
| `s01-n3p-pair-del-inv-p4-sep1` | `n.` NCMultiExon(Plus) | del + inv | 1 | 9 | `NR_TEST.1:n.[36_39del;41_44inv]` → `n.36_44delinsTCAGA` |
| `scale-g-block128-inv-del` | `g.` Genomic | inv + del | 1 | 257 | `NC_TEST.1:g.[265_392inv;394_521del]` → `g.265_521delins…` |
| `scale-g-block248-delins-del` | `g.` Genomic | delins + del | 1 | 497 | `NC_TEST.1:g.[265_512delins…;514_761del]` → `g.265_761delins…` |
| `scale-g-block1032-delins-del` | `g.` Genomic | delins + del | 1 | 2065 | `NC_TEST.1:g.[265_1296delins…;1298_2329del]` → `g.…` |
| `scale-g-block1032-inv-del` | `g.` Genomic | inv + del | 1 | 2065 | `NC_TEST.1:g.[265_1296inv;1298_2329del]` → `g.…` |

Common shape: exactly two members, `Disjoint`, separation 1, frameless axis, both members consuming reference, net deletion collapsed into one spanning `delins`.

At 3' this is easy to miss — the leaving-transcript assertion aborts the loop before the census prints — so it first surfaced on a 5' run. It is present in both directions.

## Why it is rule 1 rather than a representation choice

`consultation/SVD-WG010.md:30-33` states current law inside the rejection notice:

> `:32` — "two variants separated by one or more nucleotides should be described individually and **not** as a 'delins'."
> `:33` — "**exception**: two variants separated by one nucleotide, **together affecting one amino acid**, should be described as a 'delins'."

Two decided ruling records close the exception on a frameless axis:

- `projection-codon-exception-is-decided-by-the-rendered-axis` (governing `DNA/delins.md:42`) — "`:42` is a conditional with two conjuncts … the second is not statable on a genomic one, so the clause's antecedent is **unsatisfied rather than merely weaker**. On a genomic axis they do not fire, and `DNA/delins.md:17` governs unopposed: the members are described individually."
- `delins-payload-coincidence-carve-out-is-coding-dna-scoped` (governing `DNA/delins.md:47`) — the payload-coincidence carve-out reaches `c.` and nothing else.

And `delins-adjacent-members-when-both-consume-reference` is scoped to separation **zero**, not one.

So nothing licenses these merges: `general.md:34` / `DNA/delins.md:17` govern unopposed and the members must be described individually. Under `README.md` rule 1 that makes the output non-conformant, and rule 7 makes it a bug rather than a disclosure — the flip's confluence gain (3,249 divergent classes → 0) does not pay for it.

## Relationship to #1616

#1616 moves the same counter, by a different mechanism (deleting the input-relative weight bound), and its rows have the identical shape. Measured at a common base (`b9f6cecb`, 3'): the flip contributes 54, #1616 contributes 27, they share 19, and **neither set contains the other** — 35 exclusive to the flip, 8 to #1616. The 19 shared rows produce byte-identical outputs under both mechanisms.

The two exclusive sets are mechanistically distinct, so **a gate written against one does not close the other**:

- the flip-only rows carry a substitution or inversion member that the live partitioner never re-cuts, and the weight bound never blocked them (derived weight ≤ input weight);
- the #1616-only rows are payload-bearing genomic members where the input's own member weight was the only thing refusing the merge, and the canonical partitioner still declines them.

## Scope

This issue is the separation-**one**, frameless population only. Explicitly **not** in scope: coding-axis merges where `general.md:35`'s exception is live; merges at separation ≥2 (a `delins-merge-vs-individual-gap-two-or-more` scope question); and `outputs_leaving_the_transcript`, which is #1704.

One instrument note that matters for anyone verifying a fix: `guard_violations` is granted only to `Genomic | NonCodingMultiExon` shapes at `separation == 1` with exactly two member kinds and no `Ins`. It therefore cannot count a coding-axis merge at any separation. A sibling counter for the coding-axis population at separation ≥2 is proposed separately.

nh13:
Closes #1711.

## What was wrong

`coalesce_compensating_gap_split` (`src/normalize/merge.rs`) re-spells a canonical-arm split as the single spanning `delins` of `DNA/delins.md:44-47`. It runs only on the two `partition_block_canonical` arms, which is why the defect is flip-specific — with `FERRO_PARTITION` unset the shipped `live` arm never reaches it.

The pass can only ever join pieces **across unchanged reference bases**: a piece list with no gaps between its members is one piece, so there would be nothing to join. Every firing therefore sits inside the population `general.md:34` governs — "two variants separated by one or more nucleotides should be described individually and **not** as a 'delins'" — and the only clause that can override it for this shape is `DNA/delins.md:47`.

`:47`'s reach is already settled. The decided `rulings[delins-payload-coincidence-carve-out-is-coding-dna-scoped]` (2026-08-11) scopes it to the coding DNA axis "and nothing else"; its sibling `rulings[delins-merge-vs-individual-gap-two-or-more]` says the same from the other side — rows on an axis with no reading frame "remain violations and are untouched by this ruling". The sibling pass `coalesce_payload_alignment_split` was gated on that ruling and this one was not, so **one clause had two answers on one arm**: on `CanonicalCoalesced`, a `g.` payload coincidence is refused a merge by one pass and granted it by the other, two statements earlier in the same function.

Off `c.` the ungated pass merged two variants one unchanged nucleotide apart into one spanning `delins` — the answer of the **rejected** proposal SVD-WG010 — on six rows of the spec conformance corpus, which that corpus's own negative guard counts.

These are a **rule 2** miss (Recommended form, best effort): `general.md:34` grades as a preference, per `rulings[separation-rule-force-modal-or-negation]` (ruled 2026-08-12, landing separately — it is not in this branch's ledger), and `DNA/delins.md:81` restates the same rule with "preferably". Rule 2 still binds — a preference clause outranks maintainer judgement — which is why the fix is worth making.

## The fix

A sibling gate to `payload_coalesce_applies`:

```rust
fn compensating_gap_coalesce_applies(rule: PartitionRule, kind: CisKind) -> bool {
    rule.cuts_with_canonical() && AxisFrame::is_coding_dna(kind)
}
```

Same clause, same axis scope, wider arm scope (the compensating-gap shape is produced only by `partition_block_canonical`, so both canonical arms are audited, not just `CanonicalCoalesced`). The gate is the axis **kind** and deliberately not `AxisFrame::reading_frame`, which is true for a coding `r.` and would extend a DNA-only carve-out onto an axis the ruling does not reach — the trap that ruling's own implementation note names.

No new constant, and no condition inside the pass was touched.

## `guard_violations`, both directions, both canonical arms

`spec_conformance_axis`, 210 guarded rows of 11,272 classes / 58,552 outputs, `FERRO_MANIFEST` set, both spec artifacts regenerated with `--bin`.

| arm | direction | before | after |
|---|---|---:|---:|
| `canonical` | 3' | 6 | **0** |
| `canonical` | 5' | 6 | **0** |
| `canonical-coalesced` | 3' | 6 | **0** |
| `canonical-coalesced` | 5' | 6 | **0** |

The six rows are `s01-n3m-pair-del-inv-p4-sep1`, `s01-n3p-pair-del-inv-p4-sep1`, `scale-g-block128-inv-del`, `scale-g-block248-delins-del`, `scale-g-block1032-delins-del`, `scale-g-block1032-inv-del`.

**Every other census figure is byte-identical** in both directions on both arms:

| figure | 3' | 5' |
|---|---:|---:|
| outputs / declined / unparseable | 58552 / 0 / 0 | 58552 / 0 / 0 |
| denoting no sequence | 10 | 18 |
| leaving the transcript | 389 | 0 |
| prohibition-violating outputs | 8 | 8 |
| converged | 11016 | 10753 |
| split 2 / 3 / 4+ | 673 / 164 / 29 | 981 / 124 / 24 |
| non-idempotent | 4 | 4 |
| sequence changed | 4 | 0 |
| refusal (conflicts / absolute / conditional) | 72 / 32 / 16 | 72 / 32 / 16 |

At 3' the `outputs_leaving_the_transcript` assertion aborts the loop before the guard assertion, so the guard figure is read out of the census printed in the failure message, not out of a green run. That counter reads **389** on the canonical arms against the `live` pin of 371; it is unmoved by this change and is #1704's, not this PR's.

## What happened to the other firings

Instrumented at the collapse, `FERRO_PARTITION=canonical`, over the same corpus. The gate suppresses **exactly** the non-coding-DNA firings and nothing else — that partition is arithmetic, not a threshold:

| direction | firings before | suppressed | `c.` firings after |
|---|---:|---:|---:|
| 3' | 4,204 | 2,101 (995 `g.`, 1,106 `n.`) | 2,103 |
| 5' | 3,722 | 1,892 (916 `g.`, 976 `n.`) | 1,830 |

The brief for this work quoted **6,290** firings across the 5' corpus (3,094 `c.` / 1,518 `g.` / 1,678 `n.`). Measured here it is **3,722** (1,830 / 916 / 976). The shape of the claim survives — most firings are on `c.`, only 6 land on guarded rows — but the figure does not; quote the one above with the command that produced it.

The census cannot see representation movement, because a class whose every spelling moves together stays converged. Diffing every `(row, spelling) -> output` pair either side:

| direction | outputs compared | moved | `g.` | `n.` | `c.` | distinct rows |
|---|---:|---:|---:|---:|---:|---:|
| 3' | 58,412 | 1,471 | 675 | 796 | **0** | 260 |
| 5' | 58,412 | 1,304 | 602 | 702 | **0** | 230 |

By direction of the move: 3' is 1,387 splits / 65 re-spellings / 19 merges; 5' is 1,237 / 48 / 19. So the price is real, it is paid entirely in split `g.` and `n.` descriptions, and it is the same price `delins-payload-coincidence-carve-out-is-coding-dna-scoped` already accepted for the sibling pass when it kept a stored 723-base `g.` `delins` as a 193-member re-derivation. That record explicitly leaves "whether the coalesce should acquire some other licence for spans of that size" open; this PR does not answer it, it stops the two passes answering it differently.

Identical figures on `canonical-coalesced`.

## `cis_confluence_axis` / `cis_confluence_nr_axis`

Both corpora, `FERRO_PARTITION=canonical`, both directions, before and after. All six censuses are **byte-identical**, `sequence_changed: 0` and `declined: 0` throughout:

| census | classes | converged | split 2 | split 3 | split 4+ |
| `g,c` 3' | 11,272 | 11,271 | 1 | 0 | 0 |
| `g,c` 5' | 11,272 | 11,272 | 0 | 0 | 0 |
| `n.` 3' | 5,636 | 5,635 | 1 | 0 | 0 |
| `n.` 5' | 5,636 | 5,636 | 0 | 0 | 0 |
| `r.` 3' | 5,636 | 5,633 | 3 | 0 | 0 |
| `r.` 5' | 5,636 | 5,634 | 2 | 0 | 0 |

The divergent class **ids** were dumped and diffed rather than the totals compared, because a class entering `split_two` from `split_three` and one leaving `converged` look identical in a total. The sets are equal either side: `s02-r-m3-sep1-p1-rot3`, `s02-r-m4-sep1-p1-rot3`, `s06-g-m4-sep1-p1-all-dup`, `s06-n-m4-sep1-p1-all-dup`, `s06-r-m4-sep1-p1-all-dup`. `examples/dump_confluence_divergences --out` agrees on the `g,c` corpus (one divergent class, `s06-g-m4-sep1-p1-all-dup`, both sides, both arms); on the `n,r` corpus that dumper excludes 10,536 classes as "unresolved member spans", so the `n.`/`r.` ids above come from the axis tests themselves and not from it.

## A narrower cut was looked for and the search failed

The pass's three conditions are aggregates over the whole hull and **none** of them is about separation, the axis, or the input — on the `n3m`/`n3p` rows it swallowed a two-base run as well as a one-base one. The obvious repair is therefore a per-gap test, and every candidate was evaluated over the full 5' firing dump, asking it to refuse all 12 guarded-row firings and few others:

| candidate | refuses of the 12 | refuses of the other 3,710 |
|---|---:|---:|
| some gap sits at cumulative offset 0 | 2 | 1,646 |
| some gap is exactly one base | 12 | 3,124 |
| claimed columns < twice the unchanged bases | 8 | 3,077 |
| three or more unchanged bases in the hull | 12 | 3,608 |
| hull of nine bases or more | 12 | 2,981 |

The zero-offset test is the one with a principled story — an unchanged run at zero shift is unchanged *in place* rather than by coincidence — and it is refuted outright: on all six rows **every** gap sits at a non-zero cumulative offset. The `n3m` row's derived pieces over `CGATTTCTG` are `[3,8)/"" · gap 2 @ −5 · [10,11)/"A" · gap 1 @ −5 · [12,12)/"A"`. Nothing else separates the populations; each predicate that catches all 12 also catches 80–97% of the rest. The axis is the only clean separator in the data, and it is also the one a decided ruling names. The table is on the new gate so the search is not re-run.

The dominator / "forced separation" family was not revisited; it is annotated WITHDRAWN.

## Doc correction

`coalesce_compensating_gap_split`'s own doc justified its rule from a 65-row reproducer corpus and closed with "**There are no counterexamples in the 65 rows.**" That sentence is true of the corpus it names and was being read as a property of the rule. These six rows are counterexamples in the 11,272-class conformance corpus, and the doc now says so, with the worked `n3m` piece list and with why conditions 2 and 3 are satisfied by construction on that shape (an `inv` or payload-bearing `delins` presents to an edit-distance aligner as inserted somewhere and deleted somewhere else; the big `del` beside it makes `changed_columns_of_pieces` trivially clear the density bar). It also records that nothing downstream re-audits the merge — `split_concealed_separations` runs *before* this pass and is gated on `pieces.len() >= 2` — which is why the gate is at the call site.

## Tests

- `the_compensating_gap_coalesce_needs_both_a_canonical_arm_and_the_coding_dna_axis` — the full arm × kind matrix. Pins the axis half (dropping it is #1711) and the arm half (narrowing it to `CanonicalCoalesced` would silently stop auditing `Canonical`, whose partitioner manufactures the split).
- `the_two_delins_47_gates_share_one_axis_scope` — a cross-product equality against `payload_coalesce_applies`, so the two re-spellings of one clause cannot drift apart again. Two hand-written matrices are how they drifted apart the first time.
- `the_cli_declines_the_compensating_gap_coalesce_off_the_coding_dna_axis` (`tests/it/partition_switch_wiring.rs`) — the **call site**, end to end through the `ferro` binary.

### The call site was unguarded, and this description said it could not be guarded

An earlier revision of this section said the two tests above were "pure-predicate tests rather than call-site tests" because `partition_rule()` caches in a `OnceLock`, so one test binary cannot drive two arms through `canonicalize_from_sequence`. **The premise is true and the conclusion was wrong**, and the cost of it was measured: with the call site's axis argument replaced by a hardcoded `CisKind::Cds` — #1711 restored verbatim — the whole suite passed (10,233/10,233) and `clippy` exited 0. Reverting the call site to the pre-#1711 form passed too. The predicate was guarded four ways over and the thing it was wired to was guarded not at all.

`tests/it/partition_switch_wiring.rs` already sidesteps the `OnceLock`: it drives the real `ferro` binary in a **subprocess**, which gets its own, against the `MockProvider` — so no `--reference` and no prepared reference directory. The new test lives there rather than beside the predicate because what it asks is whether the call site consults the axis at all, which is a wiring question and survives either reading of what `delins.md:47` reaches.

**Input, and the check that it is not vacuous.** `NR_000123.1:n.[2_4inv;6del]` — the mock non-coding transcript (`ACGTACGTACGT`, no CDS, so `CisKind::Tx`, `false` under both `is_coding_dna` and `carries_translated_frame`), in the same `inv`+`del`-at-separation-1 shape as the `s01-n3{m,p}-pair-del-inv-p4-sep1` rows this PR is about. No fixture change was needed; the accession as committed hosts the shape. A guard built on an input that cannot reach the pass would pass under the mutation and prove nothing, so that was measured first, on the unmutated build:

- remove the call entirely (`if false`) — the row's output is **unchanged**, so nothing else is producing this answer;
- remove only the axis argument — the row merges to `n.2_6delinsACGA`, one `delins` spanning the members *and the unchanged base between them*.

So the pass is reached on this row and the axis test is what declines it. Over a 35,264-row sweep of `n.` cis alleles on that transcript, 1,207 rows move when the axis argument is removed — 926 of them onto a single spanning `delins`.

A coding control rides in the same subprocess, `NM_000088.3:c.[3_7inv;9del]` → `c.3_9delinsTGGGCA`, because the negative half alone cannot tell the fix from a build that deleted the call: with the call removed that row splits into `c.[2dup;4_5inv;8_9del]`.

**Mutation table.** One test, run on both canonical arms:

| call site | new guard |
|---|---|
| `compensating_gap_coalesce_applies(partition_rule(), kind)` — this branch | passes |
| `partition_rule().cuts_with_canonical()` — the pre-#1711 defect | **FAILS**, on the `n.` half |
| `compensating_gap_coalesce_applies(partition_rule(), CisKind::Cds)` | **FAILS**, on the `n.` half |
| `false` — the call removed | **FAILS**, on the `c.` control |

The negative half deliberately does not pin the split spelling. `general.md:34` governs that the members stay individual; which individual members the re-derivation lands on is a separate question, and pinning it would redden this guard for changes that leave its subject untouched.

Local gate, `FERRO_MANIFEST` set and both spec artifacts regenerated with `--bin`: `cargo nextest run --features dev --lib --test it --test-threads=4` — **10,237 passed, 150 skipped, 0 failed**. `cargo fmt --all` and `cargo clippy --all-features --all-targets -- -D warnings` clean. No pin re-blessed.

## Not in scope

Coding-axis merges (`general.md:35`'s exception is live there and `:47` reaches `c.`); merges at separation ≥ 2; `outputs_leaving_the_transcript`, which is #1704; and whether the coalesce should acquire a separate licence for very large spans, which `delins-payload-coincidence-carve-out-is-coding-dna-scoped` leaves open by name.

Representation-Change: none. The gated pass is reachable only under FERRO_PARTITION=canonical / canonical-coalesced; with the variable unset the shipped `live` arm never reached it, the full local suite is green and no pin is re-blessed. The canonical-arm figures are tabulated above.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -3062,7 +3062,14 @@ pub(crate) fn canonicalize_from_sequence<P: ReferenceProvider>(
     // partitioner produced the pieces. Shipped output is unaffected either way —
     // `FERRO_PARTITION` unset is `Live` — but a bake-off column can carry a
     // widening the arm it is labelled with did not cause.
-    if partition_rule().cuts_with_canonical() {
+    // Scoped to the coding DNA axis as well as to the arms, per the decided
+    // `delins-payload-coincidence-carve-out-is-coding-dna-scoped` (#1711): the
+    // pass can only merge *across unchanged reference bases*, which is exactly
+    // the population `general.md:34` keeps individual, and `DNA/delins.md:47` —
+    // the only clause that overrides it here — reaches `c.` and nothing else.
+    // See `compensating_gap_coalesce_applies` for the argument and for the
+    // competing reading it decides against.
+    if compensating_gap_coalesce_applies(partition_rule(), kind) {
         coalesce_compensating_gap_split(&mut pieces, &ref_bytes);
     }
     coalesce_whole_block_inversion(&mut pieces, &ref_bytes);
@@ -5185,6 +5192,42 @@ const MIN_MANUFACTURED_SPLIT_MEMBERS: usize = 3;
 /// `general.md:34` keeps them individual. **There are no counterexamples in the
 /// 65 rows.**
 ///
+/// # …but there ARE counterexamples in the 11,272-class conformance corpus
+///
+/// **Corrected 2026-08-12 (#1711).** The sentence directly above is true of the
+/// corpus it names and was read as a property of the rule. It is not. Six rows
+/// of `conformance::spec_corpus` fire this pass and are merged wrongly, and the
+/// spec corpus counts them itself, through the negative guard it carries for the
+/// **rejected** SVD-WG010 proposal — `spec_conformance_axis`'s `guard_violations`,
+/// pinned at 0, read 6 in both directions on both canonical arms:
+/// `s01-n3{m,p}-pair-del-inv-p4-sep1`, `scale-g-block{128,1032}-inv-del` and
+/// `scale-g-block{248,1032}-delins-del`.
+///
+/// **None of the three conditions is about the axis, and two of them are
+/// satisfied by construction on that shape.** An `inv` or a payload-bearing
+/// `delins` member presents to an edit-distance aligner as inserted somewhere
+/// and deleted somewhere else, so condition 2 holds; and the big `del` beside it
+/// makes condition 3 hold, since [`changed_columns_of_pieces`] counts a
+/// deletion's full reference width. Worked, on the `n3m`/`n3p` row — the derived
+/// pieces over `CGATTTCTG` are `[3,8)/"" · gap 2 · [10,11)/"A" · gap 1 ·
+/// [12,12)/"A"`, so 7 claimed columns against 3 unchanged and `inserted &&
+/// deleted` both true. Note the pass swallowed a **two**-base run as well as a
+/// one-base one: there is no per-gap notion of separation in here at all, and
+/// none of the arithmetic implements "one unchanged base".
+///
+/// So the fix is not in the conditions. The caller now gates the pass on the
+/// **coding DNA axis** ([`compensating_gap_coalesce_applies`]), which is where
+/// `DNA/delins.md:47` — the only clause that can license merging across
+/// unchanged bases at all — was decided to reach. Read that gate's doc before
+/// changing anything here: it also records the competing reading, under which
+/// this rule is partitioner parity rather than a `:47` re-spelling and would
+/// apply on every axis.
+///
+/// **Nothing downstream re-audits the merge.** `split_concealed_separations`
+/// runs *before* this pass and is gated on `pieces.len() >= 2`, so once one
+/// spanning piece is written it is final. That is why the gate is at the call
+/// site and not a later repair.
+///
 /// So this takes no threshold and no constant, which is deliberate: the live
 /// path's [`MAX_SINGLE_BASE_SEPARATION_CHANGE`] records that its own value is
 /// "under-determined", constrained only to `[2, 15)`, and that "no threshold at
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #1732, opened 2026-08-12
fix(benchmark): repair a hollow supplemental record from the FASTA, and refuse a corrupt one

### LINKED ISSUE #1728: Supplemental metadata ingestion cannot heal a hollow record, and silently resets on a corrupt one
`fetch_fasta_to_file` writes `supplemental/*.metadata.json` for both the benchmark bin (`patterns_transcripts`, `src/bin/benchmark.rs:1599-1642`) and `ferro prepare` (`clinvar_transcripts`, `src/prepare/mod.rs:1461`). Both carry two defects that, together, make a degraded artifact permanent.

## 1. Resume skips on presence, not usability

`src/benchmark/cache.rs:2019`

```rust
.filter(|acc| !metadata.transcripts.contains_key(*acc))
```

A record that exists but carries nothing permanently excludes its own accession from ever being re-fetched. Re-provisioning prints `All N accessions already fetched (skipping)` and changes nothing, so **the artifact cannot heal itself**.

That is not hypothetical. The prepared reference's `patterns_transcripts.metadata.json` holds 140,027 records, of which **116,572 carry neither a CDS nor a sequence length**, 23,455 have a length, and **19** have a CDS. Of 96,739 `NM_` records, exactly one has a CDS. The sibling FASTA's `.fai` has true lengths for all 140,027 — the data is one file over.

Fingerprint of an out-of-band rebuild: `.fna.bak` is Jan 6 @ 2.5 GB, `.fna` is May 21 @ 1.3 GB, the metadata's mtime is May 21 while its own `generated_at` still reads `2026-01-06`.

## 2. A corrupt sidecar silently resets to empty

`src/benchmark/cache.rs:2004`

```rust
serde_json::from_reader(..).unwrap_or_else(|_| empty)
```

Any parse failure discards the entire record of what has been fetched, making a truncated file indistinguishable from a first run — so the next pass re-fetches everything and overwrites it with whatever it gets.

## Why this is separate from #1722

#1722 / PR #1724 fixed the **consumer**: `MultiFastaProvider` no longer trades a complete cdot exon table for a supplemental record that carries nothing. That stops the damage but cannot repair the artifact, because the producer will never re-fetch a hollow row. This issue is the producer half.

## Scope

Repairs ingestion only. An already-shipped reference stays degraded until it is next provisioned; what changes is that re-provisioning can then succeed. `supplemental/legacy_genbank.metadata.json` comes from `prepare::fetch_legacy_versions`, which uses `File::create` and rebuilds its metadata each run — it has neither defect and needs no change, which is also why it is 5/5 populated while `patterns_transcripts` is not.

There is currently no test anywhere constructing these types with non-trivial values, and `src/benchmark/cache.rs` has no `#[cfg(test)]` module at all.

nh13:
Closes #1728. The producer half of #1722 — #1724 fixed the consumer and could not repair the artifact, because the producer could not heal a hollow row.

## The ruling this implements

**Operator ruling, 2026-08-13: repair a hollow supplemental row LOCALLY from the FASTA. Do not add a network re-fetch.**

`SupplementalTranscript` mixes two kinds of field, and "hollow" is a **row**-level predicate that spans the boundary:

```rust
pub struct SupplementalTranscript {
    pub id: String,
    pub gene_symbol: Option<String>,
    pub cds_start: Option<u64>,      // SOURCED  — only a GenBank record supplies it
    pub cds_end: Option<u64>,        // SOURCED
    pub sequence_length: usize,      // DERIVED  — it is the length of the sequence just written
}
```

All three sites in `cache.rs` that construct a row set `sequence_length` to `seq.len()` / `current_seq.len()` and to nothing else, and two of them are FASTA-only paths already building `cds_start: None, cds_end: None, sequence_length: current_seq.len()`. So the FASTA is that field's authority and the sidecar is a cache of it.

**The rule: never fetch to recover a value you can compute.** #1791 made the supplemental FASTA the single authority for presence. Measured on the shipped `patterns_transcripts` artifact:

```
sidecar rows                          : 140,027
FASTA (.fai) records                  : 140,027
rows with sequence_length == 0        : 116,572
rows whose length disagrees with FASTA: 116,572   (none nonzero-but-wrong)
hollow accessions present in the FASTA: 116,572 of 116,572
sidecar keys not in the FASTA         : 0
FASTA records not in the sidecar      : 0
```

Every hollow accession's bases are already on disk. That is precisely why `!store.has_sequence(acc)` admits none of them — and it is also why they need no fetch: the length is not *missing*, it is merely *not indexed*. Widening the resume filter would have been a 116,572-accession round trip against a rate-limited endpoint to learn a number one sequential read away.

## What the code does

`repair_derived_lengths` streams the FASTA once, before anything is fetched, and re-derives each row's `sequence_length` from the record for its accession:

- **first record wins** on a duplicate-bearing FASTA, so it agrees with `SupplementalStore::commit`, which keeps the first record and drops later duplicates;
- a row whose accession has **no** FASTA record is left alone — that is an orphan, and dropping it belongs to the store's key-set reconciliation;
- **only `sequence_length` is touched.** `gene_symbol`, `cds_start` and `cds_end` are carried through untouched.

Persisting is then split by which artifact actually moved (`commit_or_repair`):

- the FASTA, its `.fai` or the sidecar's key set moved -> a full `commit`, which carries the repaired rows out with it (`commit` preserves rows it already has);
- **only** a derived row moved -> a sidecar-only atomic write. Repairing 116,572 lengths must not re-emit a 1.3 GB FASTA and a 4.9 MB index that are byte-for-byte correct, and the authority rule says the same from the other side: correcting a derived artifact is not a reason to rewrite the authoritative one.

A run with nothing to fetch and nothing to repair still writes no byte.

## CDS is deliberately out of scope

`cds_*` is sourced, so it can only come from a fetch, and that is a separate lazy per-accession operation rather than a bulk backfill. **Only 19 of the 140,027 rows carry a CDS at all**; the other 140,008 keep `cds_start = cds_end = None` after this change, and this PR does not fetch to obtain one. (The ruling phrased this as "genuinely missing on 19 of 140,027 rows"; the measurement above is the same 19, read as the rows that *have* a CDS. Either way the treatment is identical: sourced, so untouched here.)

`gene_symbol` is likewise sourced and untouched — and moot on this artifact, where all 140,027 rows already carry one.

## No re-fetch: what changed from the original approach

This PR previously widened the resume filter with `!metadata.has_usable_record(acc)`, moving 116,572 accessions from skip to re-fetch. That is withdrawn. The resume filter is **unchanged** — still `!store.has_sequence(acc)`, still #1791's — and `has_usable_record` no longer exists.

## A corrupt sidecar is refused, not silently reset

`unwrap_or_else(|_| empty)` discarded the entire record of what had been fetched on any parse failure, making a truncated file indistinguishable from a first run — so the next pass overwrote it with whatever it got. `load_supplemental_metadata` now returns `FerroError::Json` naming the file, matching how the consuming reader of this same artifact has always treated it. An **absent** file is a different state and stays silent, so a genuine first run is unchanged.

## #1791's guards

Both merged guards still assert `source.calls() == 0`, and both still pass, because repairing a derived field makes no source call.

`a_duplicate_bearing_artifact_is_repaired_and_then_stable` is **unmodified**.

`a_hollow_sidecar_row_does_not_produce_a_second_fasta_record` needed **one assertion narrowed**, and this is flagged for review rather than slipped in. It asserted that a hollow row is the fixed point and that *all three* artifacts are byte- and mtime-identical afterwards. Under this ruling a hollow row is no longer the fixed point — it is the thing being repaired — so the sidecar does move. The FASTA and the `.fai` do not, and that is what the assertion now says:

```rust
assert_eq!(before[..2], artifact_state(&fasta)[..2],
    "repairing a derived field must not rewrite the authoritative FASTA or its index");
```

Its two load-bearing assertions — `headers(&fasta) == ["NM_000001.1"]` and `source.calls() == 0` — are untouched, and the sidecar's movement is pinned positively by the new `a_hollow_sidecar_row_is_repaired_from_the_fasta_without_a_fetch`.

`a_run_whose_every_fetch_failed_rewrites_nothing` is seeded with a healthy row rather than a hollow one, so it observes only the property it names. Its full three-artifact assertion is intact.

## Two files, not one

The implementation is `src/benchmark/cache.rs`. `src/benchmark/supplemental_store.rs` is touched **only inside its `#[cfg(test)]` module** and in two comments — the end-to-end tests belong beside `FakeEfetch` and #1791's own guards, which live there. No non-test code in that module changes.

Worth noting while you are there: that module's doc table says what is derived is "the sidecar's **key set** and each row's `sequence_length`", which the 2026-08-12 review recorded as an overstatement, since `commit`'s `or_insert_with` carries an existing row's length forward untouched. It is now true *of the provisioning entry point*, which re-derives every length before every commit — and still not true of `commit` called directly. The prose is left alone rather than half-corrected.

## Verification

The behavioural test is `a_hollow_sidecar_row_is_repaired_from_the_fasta_without_a_fetch`. Proved by mutation, not by assertion: with the `repair_derived_lengths` call removed the sidecar comes back with `sequence_length: 0` and the test is red; with it, green, and `source.calls()` is 0 on both arms — so the test is measuring the repair and not a fetch.

`a_repaired_sidecar_is_the_fixed_point_on_the_next_run` pins convergence, seeded from the damaged state rather than a clean one.

Also out of scope, and a stacked PR's job: making the derived/sourced split explicit in the type — dropping `sequence_length` as a persisted field and rebuilding it as an in-memory index on load, so a hollow row becomes *unrepresentable*.

Representation-Change: none. Producer-side ingestion under `src/benchmark/`; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched, and no normalized output moves.

### DIFF
--- a/src/benchmark/cache.rs
+++ b/src/benchmark/cache.rs
@@ -36,7 +36,7 @@ use serde_json::json;
 use std::collections::{HashMap, HashSet};
 use std::fs::File;
 use std::io::{BufRead, BufReader};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicUsize, Ordering};
 
 /// Get NCBI API key from environment variable.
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #1733, opened 2026-08-12
docs(rulings): record that a derivation may not be bounded by the input's spelling

nh13:
Splits the ledger half of #1616 out from the code change it justifies, so the adjudication can be reviewed on its own terms and land independently of the normalizer edit. #1616 keeps the contested part: the deletion of the input-relative weight bound.

## What this adds

`derivation-may-not-be-bounded-by-the-inputs-spelling` — decided, operator ruling 2026-08-10 — holds that no canonicalization may be refused on the ground that it describes more change than the input's own spelling did. The record carries no `applies_to` rows and pins no output. The deletion of the bound itself is #1616 and is deliberately **not** here.

## Everything else is a mechanical consequence of the ledger gaining one record

- **`CLAUDE.md`** gains the row `claude_md_adjudication_tables` requires. Its summary states the ruling's scope — removing an input-relative comparand does not license the merges that comparand happened to block — and it deliberately does **not** repeat the masking attribution to `best_alignment`'s single-gap restriction (#1617). The record itself withdraws that attribution by name and measures the producers as `partition_block`'s two payload-coincidence exits plus the `MAX_SPLIT_BLOCK` length cap, so writing the withdrawn version into the table would land a claim the record it summarises already refutes. This is the one place the diff departs from #1616's own text, and that is why.
- **`clause_ruling_index.rs`**'s generated block is regenerated from `render_index()` rather than hand-merged, and its two census pins move with it: **23/20/3 → 24/21/3 records, and 24 → 26 multiply-named clause lines**. Both pins were confirmed against the generator before the block was spliced, and the ```` ```text ```` fences are preserved.
- **`hgvs_spec_normalization_tests.rs`** pins the new record's status, which is what `ruling_records_are_intact` asks for when a record is added.

Every removed line in the diff is one of those six: two regenerated census header lines, two clause lines that gained a `[MULTI]` marker, and the two test pins.

## Census arithmetic, so the pins can be checked without regenerating

The record cites eight clause lines. Exactly one of them (`DNA/delins.md:15`) is new to the index, hence 98 → 99 clause lines. Exactly two cross from singly- to multiply-named — `DNA/delins.md:16` and `DNA/delins.md:89`, each previously named only by `delins-adjacent-members-when-both-consume-reference` — hence 24 → 26. The mixed-verdict count is unchanged: the two lines this record joins that carry an undecided neighbour (`delins.md:17`, `:47`) were already mixed, and the other six are decided-only.

## Ordering dependency, now resolved

The record's rationale cites `separation-rule-force-modal-or-negation`, which is #1725. Until that landed, `ruling_citation_currency::every_record_to_record_citation_resolves` was red here — a real ordering dependency, not a defect in this diff, and it was not routed around by editing the operator-authored rationale. #1725 has since merged and the branch has been rebased onto it, so that test is green.

Every other guard over this material passes: `ruling_records_are_intact`, all four `claude_md_adjudication_tables` tests, all of `clause_ruling_index` including `the_rendered_index_is_current`, the remaining `ruling_citation_currency` tests, `status_census_is_unchanged` and `spec_equivalence_classes_converge`.

## Verification

Full local gate on this branch: `generate_spec_fixture`, `generate_spec_enumeration`, `cargo fmt --all`, `cargo clippy --all-features --all-targets -- -D warnings` all clean. The ledger was compared programmatically against `origin/main`: exactly one record added, none dropped, no status changed. All eight clause citations were matched verbatim against the pinned `assets/hgvs-nomenclature` checkout.

Representation-Change: none


nh13:
@coderabbitai pause

PR #1734, opened 2026-08-12
fix(vcf): refuse an unresolvable genomic position instead of dropping it

### LINKED ISSUE #1729: VCF conversion silently drops a genomic offset
`HgvsToVcfConverter::convert()` sends every `Genome` variant to `convert_genome` (`src/vcf/from_hgvs.rs:74`), which reads only `.base` from each endpoint (`:505-524`). `GenomePos.offset` is never read, so `g.10+2del` is converted as if it were `g.10del` — no warning, no error, a wrong VCF row.

## Reproducer

Fixture: a 16-base `chr1` in a `MockProvider`. Debug **and** release, byte-identical output for three descriptions `normalize` keeps distinct:

```
g.10+2del -> POS=9 REF=AC ALT=["A"]
g.10-2del -> POS=9 REF=AC ALT=["A"]
g.10del   -> POS=9 REF=AC ALT=["A"]
```

## Mechanism

```rust
// src/vcf/from_hgvs.rs:505-524
let genomic_start = interval.start.inner().ok_or_else(..)?.base;   // .offset never read
let genomic_end   = interval.end.inner().ok_or_else(..)?.base;
```

The only guard on this path is `Mu::inner()`, which rejects an *uncertain* position. There is no `is_special()` check and no offset handling anywhere in the file's genomic arm — the `c.`/`n.`/`r.` helpers have their own.

The provider-less `genomic_hgvs_to_vcf`, which is what `ferro hgvs2vcf` calls, has the identical defect: unguarded it lowers `g.10+2C>T` and `g.10C>T` to the same `POS=10` row.

## This is a sibling of an already-fixed defect

Same shape as #1628 / PR #1641 (`ae79839`, "fix(spdi): refuse a genomic offset instead of dropping it"). That fix closed the instance in `hgvs_to_spdi` and not the class; the VCF converter escaped it.

## Why nothing catches it

None of the four seam oracles reaches here — they guard the normalization seam, and this is an output converter. The record is well-formed VCF, so nothing downstream objects. For any pipeline keying on VCF this is a silent data defect.

### LINKED ISSUE #1730: VCF anchor underflows on a genomic pter/qter/cen position
A genomic `pter`/`qter`/`cen` position carries `base: 0` plus a side flag. `convert_genome` reads only `.base`, so the landmark flattens to 0 and reaches `build_vcf_record`, whose Deletion arm computes `start - 1` on `u64`.

## Reproducer

Fixture: a 16-base `chr1` in a `MockProvider`. The two build profiles disagree, which is the part that matters:

```
debug:   thread panicked at src/vcf/from_hgvs.rs:576:34: attempt to subtract with overflow
release: Invalid coordinates: Position 18446744073709551614-18446744073709551615
         out of range for chr1 (length 16)
```

The release wrap is caught **only** because this provider bounds-checks its fetch. A provider that does not would emit a record at the wrapped coordinate.

## Mechanism

`src/vcf/from_hgvs.rs:576` computes the VCF anchor as `start - 1` with no lower bound. `start == 0` is reachable because nothing between the parsed landmark and this line rejects a `special` position.

Note `:619`, in the length-changing `Delins` branch, contains the same subtraction but is **not** reachable at `start == 0` — that arm fetches the deleted bases first and `get_reference_sequence`'s `start >= 1` validation fires before the anchor is computed. Worth stating explicitly, because reading that incidental refusal as coverage is exactly the trap #1662 documents: it is a check asking a different question, and the ordering it depends on is not pinned by anything.

## This is a sibling of an already-fixed defect

Same shape as #1643 / PR #1662 (`6b79f80`, "fix(spdi): refuse a genomic pter/qter/cen instead of flattening it to base 0"). That fix closed the SPDI instance; the VCF converter escaped it.

## Why nothing catches it

The four seam oracles guard the normalization seam, not an output converter. In release the failure is a wrapped `u64`, which looks like a coordinate rather than an error.

nh13:
Closes #1729. Closes #1730.

The VCF half of the coordinate-seam family: the same two defects `hgvs_to_spdi` was fixed for in #1641 and #1662, in `src/vcf/from_hgvs.rs`.

`HgvsToVcfConverter::convert_genome` and the provider-less `genomic_hgvs_to_vcf` both read `GenomePos::base` and never looked at `offset` or `special`. Two fields the AST carries, discarded.

## What reproduces

Measured by tests written to fail first, against a 16-base `chr1` in a `MockProvider`.

**A dropped offset is silent in both build profiles:**

```
g.10+2del -> POS=9 REF=AC ALT=["A"]
g.10-2del -> POS=9 REF=AC ALT=["A"]
g.10del   -> POS=9 REF=AC ALT=["A"]
```

Three descriptions `normalize` keeps distinct, one VCF row, no error at all. For any pipeline keying on VCF that is a silent data defect.

**A flattened `pter`/`qter`/`cen`** sends `base: 0` into `build_vcf_record`, whose Deletion arm computes `start - 1` on `u64`:

```
debug:   thread panicked at src/vcf/from_hgvs.rs:576:34: attempt to subtract with overflow
release: Invalid coordinates: Position 18446744073709551614-18446744073709551615 out of range for chr1 (length 16)
```

The release wrap is caught **only** because this provider bounds-checks its fetch. A provider that does not would emit a record at the wrapped coordinate.

## What was chosen, and why

**Refuse**, following what #1641 established and #1662 confirmed. Honouring is not available in either case: a genomic accession has no exon structure to measure an offset against, and no telomere or centromere annotation to resolve a landmark against — `GenomePos` stores `base: 0` beside `special` precisely because there is no coordinate to store. A VCF `POS` is a single integer and can express neither. Dropping them is the third, dishonest answer.

Refusing is also what the rest of this module already does: the `c.`/`n.`/`r.` arms decline an intronic position rather than pretend it sits on the exon, and every unrepresentable edit declines by name. The genomic arm was the one hole.

**One guard, not six.** The SPDI sibling needed a call site per genomic arm and passed each its axis label. Here `g.`, `m.` and `o.` all funnel through `convert_genome` — the `m.`/`o.` arms of `convert` rebuild the variant as a `GenomeVariant` — so there is one guarded funnel and no per-site label to get wrong. The offending description is quoted instead.

**Both entry points are guarded**, as in #1641. `genomic_hgvs_to_vcf` is named in neither issue but has the identical defect and is what `ferro hgvs2vcf` calls: unguarded it lowered `g.10+2C>T` and `g.10C>T` to the same `POS=10` row.

**The anchor arithmetic is made checked at the trunk.** The offset drop is confined to `convert_genome`, but `start - 1` lives in `build_vcf_record`, which `convert_cds`, `convert_tx`, `convert_rna` and `convert_genome` all reach. Checking it there closes the subtraction for every caller rather than only for the one known to have fed it a zero, and makes the two build profiles answer identically.

## Two corrections to the reported shape

**The second `start - 1`, in the length-changing `Delins` branch, is not reachable at `start = 0`.** That arm fetches the deleted bases first, and `get_reference_sequence`'s `start >= 1` validation fires before the anchor is computed. The arithmetic is still made checked, because "unreachable today" rests on an evaluation order nothing pinned — and `the_delins_anchor_is_shielded_by_the_reference_fetch` now pins it, asserting the *fetch's* message so a future reordering is visible rather than silently absorbed. Reading that incidental refusal as coverage is the trap #1662 documents.

**`rna_pos_to_tx` (`:1179`) hard-codes `downstream: false`** — the same shape as the projector defect fixed separately. Checked and it is **genuinely unreachable** with `utr3 == true`: its sole call site, inside `convert_rna`'s non-coding `map_tx` closure, refuses `pos.utr3 || pos.base < 1` immediately above the call, and the coding branch lowers through `rna_pos_to_cds`, which carries `utr3` faithfully. Behaviour is unchanged; `the_dropped_utr3_marker_is_shielded_by_the_non_coding_guard` pins the guard and asserts its message, because the pre-existing `test_converter_rna_utr3_on_non_coding_declines` only asserts `matches!(err, ConversionError{..})` and so cannot distinguish this guard from any other decline.

## Verification

- `cargo t` — 10,216 passed, **nothing re-blessed**. For a change to what a converter answers, an untouched expectation set is the strong signal: every existing expectation already agreed. The only removed lines are the two `start - 1` and one import.
- `cargo test --doc --features dev` — 95 passed. `cargo fmt --all --check`, `cargo clippy --features dev --lib`, `cargo clippy --all-features --all-targets -- -D warnings` all clean.
- **Reproduction control:** all new tests were written and run before the guards existed. Four fail in debug (two by panic at `from_hgvs.rs:576`), and the two `pter` tests fail in **release** too — for a different reason, which is why every assertion is on the error message rather than on `is_err()`. `vcf::from_hgvs::tests` is 81/81 in both profiles after the fix.
- `examples/dump_normalized_corpus.rs` was **not** used as evidence: no family in it builds an offset-carrying or landmark genomic position, so a `0 moved` from it would be a fact about the corpus rather than about this change.

Representation-Change: none. The diff is one file, `src/vcf/from_hgvs.rs`; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched, and the normalization pipeline makes no `vcf::` call. What changes is what `convert`/`genomic_hgvs_to_vcf` answer for descriptions whose position has no coordinate to convert — from a row that answers for a different variant, or a panic, to an error.

### DIFF
--- a/src/vcf/from_hgvs.rs
+++ b/src/vcf/from_hgvs.rs
@@ -18,7 +18,7 @@
 use crate::convert::mapper::CoordinateMapper;
 use crate::error::FerroError;
 use crate::hgvs::edit::NaEdit;
-use crate::hgvs::interval::interval_is_wraparound;
+use crate::hgvs::interval::{interval_is_wraparound, GenomeInterval};
 use crate::hgvs::location::{CdsPos, RnaPos, TxPos};
 use crate::hgvs::variant::{CdsVariant, GenomeVariant, HgvsVariant, RnaVariant, TxVariant};
 use crate::project::edit::u_to_t_edit;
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1735, opened 2026-08-12
fix(convert): resolve a c./n. position on the flat transcript axis

### LINKED ISSUE #1619: hgvs_to_spdi resolves a c. position through exon gaps while the sequence it indexes is flat

nh13:
Closes #1619.

`CoordinateMapper::cds_to_tx`/`tx_to_cds` walked cdot's exon list whenever two exons did not abut in transcript coordinates, while `ReferenceProvider::get_sequence` kept serving the flat transcript. So `hgvs_to_spdi` and the normalizer named **different bases of the same accession**. Reproducer, by the denoted-sequence oracle (#1615):

```
NM_033517.1:c.4818dupC -> c.4818dup
  over NM_033517.1 [4877, 4877): input applies C / output applies T
```

The exon walk was also off by one **5' of the gap** — `c.100` resolved to transcript position 101 — so on a gapped transcript it moved every `c.` position, not only those past the hole.

## Why this is not #1665

**PR #1665 proposed the same direction and was closed.** Its premise was that transcript-coordinate gaps are malformed data cdot does not produce, evidenced by a 20-transcript sample. That premise is **false**, and this PR does not restate it. Measured over `cdot-0.2.32.refseq.GRCh38`: **58 of 474,818** multi-exon builds carry a gap (0.012%), sizes 23–2718; GRCh37, **159 of 190,754**. The gaps are genuine.

The case rests instead on **whose coordinate space they are in**. For `NM_033517.1`:

- `NP_277052.1`, the protein it codes for, is **1731 aa** (NCBI esummary `slen`; GenPept `LOCUS NP_277052 1731 aa`).
- Its GenBank record annotates **`CDS 1..5196`** and `/coded_by="NM_033517.1:1..5196"`. `5196/3 = 1732` codons = 1731 residues plus the stop.
- **RefSeq's own exon table for the accession is contiguous** — `1..63, 64..267, … 4605..7096`, no hole.

So the CDS RefSeq annotates is 39 bases longer than cdot's exon-covered span and counts the unaligned bases. cdot's hole is a fact about **its GRCh38 alignment** — 39 transcript bases that do not align — not about the numbering of the accession those bases belong to.

## The measurement that separates the two spaces

Across the 58 gapped GRCh38 builds, 50 are measurable. Reading cdot's **own** `start_codon`/`stop_codon` as coordinates yields a valid ORF under the gap-collapsed reading **50/50** and under the flat reading **0/50**. So cdot's codon coordinates are gap-collapsed while its exon `tx` coordinates are flat sequence positions — two spaces within one record. Every gap on a coding transcript is a multiple of 3, so reading frame is preserved and nothing weaker than an explicit stop-codon test can see it.

That exposes a **second, separate defect this PR does not fix**: `src/data/cdot.rs` copies those codon coordinates verbatim and `src/reference/multi_fasta.rs` stores them in a flat-space field, so the live provider serves `NM_033517.1` with `cds_end = 5157` where RefSeq says `5196`. The test fixture carries the correct flat 5196, which is why the reproducer isolates the *resolution* defect. Note the exon walk does not rescue that either: with a gap-collapsed `cds_end`, `c.*1` is wrong under both readings.

## Scope — the genome frame is untouched

This is the **sequence** frame only: `c.`/`n.` ↔ transcript. Mapping a transcript position to and from a contig (`genomic_to_tx`, `tx_to_genomic`, and the projector above them) genuinely is exon- and CIGAR-aware and is unchanged; there a gap correctly means "this transcript base aligns to nothing", which the composed conversions still report as `None` rather than as a shifted coordinate. The composed paths are one step of each: exon-aware across genome↔tx, flat across tx↔CDS. `the_genome_frame_remains_exon_aware` pins that.

This is also the same ruling as **#944** on a different shape: a CIGAR insertion adds transcript bases absent from the genome, and `c.` counts them, so `NM_015120.4:c.87` is `n.198`. `CdotTranscript::cds_to_tx` was **already** flat; this makes `CoordinateMapper` agree with it rather than with the alignment.

## Adjudication

A ruling record is added: `c-and-n-positions-are-flat-transcript-offsets`, governed by `docs/background/numbering.md:52` — *"nucleotide numbering is `n.1`, `n.2`, `n.3`, ..., etc., from the first to the last nucleotide of the reference sequence"* — with `:21` (the `c.1` anchor) as the other authority. `:40` and `:44` are cited as **supporting context only**, and the rationale says so: `:40` is about introns and flanking, `:44` about out-of-bounds, and neither settles the gap case.

The record states plainly that **the spec does not address alignment gaps** — nothing in `numbering.md` contemplates them. That is not a hole in the argument: numbering is defined over the reference sequence's own nucleotides and the alignment is not a term in that definition, so the gap case is *moot* under `:52` rather than unanswered by it.

## Verification

Oracle sweep, both sides measured on this branch's own base with the same command (the totals are a property of `main` on the day; the difference is the stable part):

```
```

18 failures before, 17 after. **FAIL → PASS: `normalize_tests::normalization_transformations::test_explicit_base_removal::case_1`. PASS → FAIL: none.**

#1618, the other row deferring `FERRO_ASSERT_SEQUENCE` out of `test-oracle`, is independently green (fixed by #1678). The flag is **not** armed in that job here: what was measured is the two rows, not that job's whole selection. The follow-up is recorded in `ci.yml` and `CLAUDE.md`.

`cargo fmt --all --check` clean; `cargo clippy --all-features --all-targets -- -D warnings` clean.

Representation-Change: moves representation on transcripts whose cdot exon table has a transcript-coordinate gap (58 of 474,818 GRCh38 multi-exon builds, 23-2718 bases; 159 of 190,754 on GRCh37). On such an accession every c./n. position, and therefore every SPDI triple and every applied base derived from one, moves onto the flat transcript offset the accession's own numbering names. The movement is NOT the gap size: the deleted walk drifted an ADDITIONAL one base per exon junction it crossed (it consumed `exon.end - current_tx` bases while advancing `exon.end + 1 - current_tx` positions), so the shift grows with exon index and the gap is only one term in it. Measured on NM_033517.1 (22 exons, one 39-base hole after exon 10), old walk minus flat offset: c.1 = 0, c.64 = +1, c.1000 = +8, c.1302 = +48, c.1342 = +48, c.2000 = +54, c.4818 = +59, c.5196 = +59. So positions 5' of the hole move by up to +48 on this accession and positions 3' of it by +59 (39 gap + 20 junctions), not by one and by the gap size respectively. Named example: NM_033517.1:c.4818 moves from transcript position 4877 (base T) to 4818 (base C), which is the base GenBank's CDS 1..5196 and the 1731-aa NP_277052.1 agree on. Pinned unit-test values that moved with it: c.*1 on a gapped 3'UTR fixture, SPDI 39 -> 35; c.-3 on a gapped 5'UTR fixture, SPDI 4 -> 8. On the 474,760 GRCh38 builds with no gap, nothing moves -- the walk was gated on `has_gaps`, so a contiguous table always took the flat early return. Genomic projection is unchanged.


nh13:
@coderabbitai pause

Issue #742, opened 2026-06-17
cdot loader stores 1-based-inclusive cdna columns as 0-based half-open tx coords (exon-junction off-by-one / decline)


nh13:
Split out of #737 (defects B and C of the genomic-axis projection work).

## Summary

`CdotMapper`'s exon loader copies cdot's **1-based-inclusive** cDNA columns straight into the in-memory `CdotTranscript.exons` `tx_start`/`tx_end`, but the rest of the code treats those fields as **0-based half-open**. The two conventions are off by one in opposite ways and *cancel for exon-interior positions*, so the bug is latent — it only surfaces for a variant landing exactly on an **exon-junction CDS base**.

## Evidence

For `NM_012459.2` (TIMM8B, GRCh38, minus strand, 2 exons; cdot cdna columns `[…,1,159]` and `[…,160,822]`):

- `transcript_length()` returns **820** for the real **822**-base transcript (loses 1 cdna position per exon).
- `c.130G>A` (first cdna base of exon 2) → cdot `exon_for_tx_pos` returns `None` (tx_pos 159 falls in the 1-position gap between loaded exon0 `[1,159)` and exon1 `[160,822)`) → projection **declines** with "no g. representation". Expected `NG_012337.1:g.3616C>T`.  *(defect C)*
- `c.129G>T` (last cdna base of exon 1) maps one position too low downstream (`g.4793C>A` vs expected `g.4794C>A`); silent because both chromosome bases are `C`, so ref-base validation passes. *(defect B)*

Exon-interior positions (e.g. `c.5`, `c.127`, `c.128`) are correct — the canceling off-by-ones make `tx_to_genome` produce the right 1-based genomic coordinate for them.

## Root cause

- The loader (`RawCdotTranscript::from_genome_build` in `src/data/cdot.rs`) sets `tx_start = cdna_start`, `tx_end = cdna_end` from cdot's 1-based-inclusive cdna columns.
- `cds_to_tx` returns **0-based** cdna; `exon_for_tx_pos`/`tx_to_genome` treat `[tx_start, tx_end)` as **0-based half-open**. The mismatch means `c.130`'s true cdna 160 maps to 159, which is the 1-position gap at the junction.
- The test fixtures (`sample_transcript`, `minus_strand_transcript`) are written in the *intended* 0-based-half-open convention and pass because they exercise the primitives in isolation, not through the real consumer — so they mask the loader bug.

## Why it's deferred (high blast radius)

[...]

PR #1737, opened 2026-08-12
fix(data): map cdot's CDS bounds out of gap-collapsed transcript space

### LINKED ISSUE #1619: hgvs_to_spdi resolves a c. position through exon gaps while the sequence it indexes is flat

### LINKED ISSUE #1736: cdot's CDS bounds are in gap-collapsed transcript space but are stored as flat offsets
A cdot record states two transcript coordinate spaces and labels neither:

- the exon table's `tx_start`/`tx_end` are **absolute offsets** into the transcript sequence;
- `start_codon`/`stop_codon` are offsets into the **gap-collapsed** transcript, i.e. with bases unaligned to the genome excluded.

`from_genome_build` (`src/data/cdot.rs:607-609`) copies the codon bounds through verbatim, and `src/reference/multi_fasta.rs:2613` stores them in a flat-space `cds_end`. So on a transcript whose exon table has a hole, the published CDS is short by the enclosed gap while the exons around it are not.

## Measured

The provider serves `NM_033517.1` with `cds_end = 5157`. RefSeq annotates `CDS 1..5196` and `/coded_by="NM_033517.1:1..5196"`; `NP_277052.1` is **1731 aa** (5196/3 = 1732 codons = 1731 residues + stop). This repo's own `tests/fixtures/sequences/normalization_transcripts.json` already carries **5196**. Two sources of truth for one accession, differing by exactly its 39-base transcript-coordinate gap, with nothing comparing them.

Direction confirmed over the 58 gapped GRCh38 builds (50 have both a CDS and a sequence on the prepared reference), slicing the real FASTA three ways and classifying each as a valid ORF:

| reading of cdot's `start_codon`/`stop_codon` | valid ORF |
|---|---|
| verbatim as flat offsets (what ferro does) | **0 / 50** |
| indexed into the exon-only sequence | 50 / 50 |
| collapsed → flat via the exon table | **50 / 50** |

The collapsed reading is *also* a valid ORF, but only 1719 codons (1718 aa), so ORF validity alone does not discriminate — RefSeq's flat annotation does.

## Scope

Distinct from #1619. That issue is about `hgvs_to_spdi` resolving a `c.` position by *walking* the exon list while the normalizer indexes the flat transcript. This one is about the CDS bounds arriving in the wrong space at ingestion. `NM_033517.1`'s `cds_start` is 0 and does not move, so the `c.4818dupC` denoted-sequence-oracle row is unaffected by fixing this.

Roughly 56 records are affected (58 gapped minus 2 non-coding), all on `cds_end`.

### LINKED ISSUE #1922: derive_transcript_structure mixes gap-collapsed and flat CDS coordinates
`derive_transcript_structure` is the one consumer of nine that disagrees with the axis #1737 establishes.

#1737 maps cdot's CDS bounds out of gap-collapsed transcript space, so `sibling.cds_start` / `sibling.cds_end` are served flat. `reconstruct_spliced_mrna`, however, concatenates **gap-collapsed** exon slices. The result is that a collapsed-frame offset is subtracted from a flat coordinate — the two are no longer in the same space.

Eight of the nine consumers move together with the change. This one does not.

## Why it does not fire today

Both shipped derived anchors have exon tables that are contiguous from zero, so the collapsed frame and the flat frame coincide and the subtraction is a no-op. The defect is therefore latent against current reference data rather than absent: it needs a derived anchor whose exon table carries a leading offset or an internal hole.

## Reproducing it

Construct a derived anchor with a non-contiguous exon table (leading offset, internal hole, or both) and compare the CDS bounds `derive_transcript_structure` produces against the eight consumers that read the served bounds directly. They should agree and will not.

## Scope

Found while reviewing #1737; not introduced by it in the sense that the mismatch only exists because the axis moved. Filing separately rather than widening that PR, since fixing it means reconciling which frame `reconstruct_spliced_mrna` should work in.

nh13:
Closes #1736.
Closes #1922 (explicit-decline hardening folded in as `bf522fac`).

**Merge-order note (corrected):** this branch is *not* stacked on #1724 — it is rebased directly on `origin/main` (`git merge-base origin/main HEAD` == `e98fa77e`, one commit ahead) and `git merge-tree` is clean against both `main` and #1724's head, so the two are independent textually. The dependency runs the other way round semantically: **#1735 should land first** (it makes a `c.` number a flat transcript offset, which is the space this PR maps cdot's codon bounds into), and **#1724 should land last** (it stops discarding gapped cdot exon tables, so it enlarges the population both of the other two repair). Recommended order: #1735 -> #1737 -> #1724.

A cdot record states two transcript coordinate spaces and labels neither. The exon table's `tx_start`/`tx_end` are absolute offsets into the transcript sequence; `start_codon`/`stop_codon` are offsets into the gap-**collapsed** transcript. `from_genome_build` copied the codon bounds through verbatim, so on a transcript whose exon table has a hole the published CDS was short by the enclosed gap while the exons around it were not.

The provider served `NM_033517.1` with `cds_end = 5157`; RefSeq annotates `CDS 1..5196` (`/coded_by="NM_033517.1:1..5196"`), `NP_277052.1` is 1731 aa (5196/3 = 1732 codons = 1731 residues + stop), and this repo's `normalization_transcripts.json` already carried 5196. Two sources of truth for one accession, differing by exactly its 39-base gap, with nothing comparing them.

## Direction measured, not assumed

Over cdot-0.2.32.refseq.GRCh38's 58 gapped builds (50 have both a CDS and a sequence on the prepared reference), slicing the real transcript FASTA three ways:

| reading of cdot's `start_codon`/`stop_codon` | valid ORF |
|---|---|
| verbatim as flat offsets (what ferro did) | **0 / 50** |
| indexed into the exon-only sequence | 50 / 50 |
| collapsed → flat via the exon table (this fix) | **50 / 50** |

`NM_033517.1` is decisive: raw `[0, 5157)` → mapped `[0, 5196)`, 1732 codons, last codon `TGA`. The collapsed reading is *also* a valid ORF but only 1719 codons, so ORF validity alone does not discriminate — RefSeq's own flat annotation does, and the repo fixture already agreed with it.

Two internal confirmations that flat is the required space: `CdotTranscript::cds_to_tx`/`tx_to_cds` add `cds_start` directly to flat tx positions, and the genomic-CDS *fallback* branch already produces flat coordinates via `genomic_to_tx_pos`. Only the `start_codon`/`stop_codon` branch was in the wrong space.

## Both bounds are mapped

`collapsed_to_flat_tx_pos(exons, collapsed)` walks the exon table accumulating spans and returns the flat offset for a 0-based inclusive collapsed offset.

- `cds_start` = `map(start_codon)`. Shifts whenever a gap precedes the start codon — no build in this cdot does (measured 0/56), but the mapping does not assume it.
- `cds_end` = `map(stop_codon - 1) + 1` — mapped through its **last coding base**, not directly. This is what the naive "add every preceding gap" gets wrong: a CDS flush with an exon boundary would absorb the gap that *follows* it. Pinned by `cdot_cds_end_flush_with_an_exon_boundary_excludes_the_following_gap`, which passes both before and after — it guards against the wrong fix, not against the defect.

Nothing keys on gaps being multiples of 3. Identity on any contiguous exon table, so this is a no-op for 474,760 of 474,818 multi-exon GRCh38 builds; **56 records change, all on `cds_end`**.

## Two cache format versions bumped

`RKYV_FORMAT_VERSION` 3→4 and `CDOT_BINCODE_VERSION` 3→4. Both layouts are unchanged, so neither rkyv's structural validation nor bincode's positional read can see a purely semantic change, and a stale cache would keep serving collapsed bounds. The bincode bump is load-bearing in its own right: a fresh legacy `.bin` is preferred over the JSON and then **promoted to an rkyv archive**, so leaving it at 3 would launder stale values into a current-version cache. Precedent is #742/#972, which bumped both for the same class of change.

Existing prepared references reparse the cdot JSON once and self-heal (~10× slower first load). Neither cache file is listed in `manifest.json`, so the blessed reference identity is not at risk.

## What this does not fix

#1619's other half, where `hgvs_to_spdi` resolves a `c.` position by *walking* the exon list while the normalizer indexes the flat transcript — that is #1735. `NM_033517.1`'s `cds_start` is 0 and does not move, so the `c.4818dupC` denoted-sequence-oracle row is unchanged by this PR.

## Verification

`cargo fmt --all --check`, `cargo clippy --features dev --lib` and `--all-targets` all clean; `cargo t` 10,215 passed, 0 failed. Three tests failed first and are shown to fail on the previous behaviour.

`tests/it/normalization_transcripts_exon_contract.rs` gains `the_gapped_records_cds_bounds_are_flat`, pinning the fixture side — the gap must lie inside the CDS, `cds_end` must land inside an exon, and `cds_end − gap` must reproduce cdot's raw 5157. Sabotage-checked: flipping 5157→5158 fails it.

Representation-Change: 56 of 482,519 GRCh38 cdot builds move — all on cds_end, all widening by their transcript-coordinate gap; 0 cds_start moves. Every contiguous exon table is untouched. This changes what the live provider serves for those 56 accessions (a c. description near or past the CDS end resolves differently), so it is a real migration for a reference-backed consumer. Declared although src/data/ and src/reference/ are not watched directories — the check does not require this trailer for this PR.

## Upstream

cdot serves two coordinate spaces without distinguishing them, which is the condition this PR compensates for. That ambiguity is reported upstream as [SACGF/cdot#123](https://github.com/SACGF/cdot/issues/123) and tracked here by #1759.

The tracking issue matters because the compensation is direction-dependent: `collapsed_to_flat_tx_pos` becomes a *double*-correction if cdot ever changes its data to serve flat bounds, landing `cds_end` at 5235 instead of 5196. That would be invisible on a contiguous exon table, where the mapping is the identity. #1759 records what to re-measure if the upstream answer changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected CDS and codon coordinate handling for transcripts containing gaps.
  * Prevented invalid or malformed exon data from producing incorrect transcript structures.
  * Added warnings when malformed exon records are discarded.
  * Improved fallback behavior for genomic CDS coordinates.

* **Compatibility**
  * Updated cached transcript data formats to ensure previously stored data is safely invalidated when coordinate semantics change.

* **Tests**
  * Expanded coverage for gapped transcripts, coordinate validation, both strand orientations, and CDS boundary handling.

### DIFF
--- a/src/data/cdot.rs
+++ b/src/data/cdot.rs
@@ -50,15 +50,22 @@ mod rkyv_cache {
     use rustc_hash::FxHashMap;
     use std::collections::HashMap;
 
-    /// Bump whenever the layout below changes so a stale archive is rejected by
-    /// the version check (in addition to rkyv's own structural validation) and
-    /// regenerated rather than mis-read.
+    /// Bump whenever the layout below changes **or its coordinates change
+    /// meaning**, so a stale archive is rejected by the version check (in
+    /// addition to rkyv's own structural validation) and regenerated rather
+    /// than mis-read.
     // v2 (#742): exon coordinates are now stored in the HGVS convention
     // (genome 1-based, tx 0-based half-open); a v1 cache holds the old raw
     // cdot convention and must be regenerated.
     // v3 (#972): `RkyvTx` gained `cds_start_incomplete`; a v2 cache has no such
     // field and would silently rehydrate every transcript as "complete".
-    pub(super) const RKYV_FORMAT_VERSION: u32 = 3;
+    // v4 (#1619): `cds_start`/`cds_end` are now FLAT transcript offsets, mapped
+    // out of cdot's gap-collapsed `start_codon`/`stop_codon` space. The layout
+    // is unchanged, so rkyv's structural validation cannot see the difference —
+    // a v3 cache would rehydrate the 58 gapped GRCh38 builds with a CDS short by
+    // the enclosed gap, which is exactly the defect this bump exists to stop
+    // surviving in a cache.
+    pub(super) const RKYV_FORMAT_VERSION: u32 = 4;
 
     #[derive(Archive, Serialize, Deserialize)]
     pub(super) struct RkyvCigar {
@@ -399,7 +406,14 @@ const CDOT_BINCODE_MAGIC: [u8; 4] = *b"FCDT";
 // v3 (#972): `CdotTranscriptSnapshot` gained `cds_start_incomplete`; a v2
 // cache has no such field and bincode's positional layout would otherwise
 // misread the following bytes.
-const CDOT_BINCODE_VERSION: u32 = 3;
+// v4 (#1619): `cds_start`/`cds_end` are now FLAT transcript offsets, mapped out
+// of cdot's gap-collapsed `start_codon`/`stop_codon` space. The layout is
+// unchanged, so a v3 cache reads back cleanly and would serve the 58 gapped
+// GRCh38 builds a CDS short by the enclosed gap. This bump must move with
+// `RKYV_FORMAT_VERSION`, not after it: a *fresh* v3 bincode cache is loaded in
+// preference to the JSON and then PROMOTED to an rkyv archive, so leaving it at
+// 3 would launder the stale values into a current-version rkyv file.
+const CDOT_BINCODE_VERSION: u32 = 4;
 
 /// Parse the integer version suffix of an accession (`"NM_003002.4"` -> `Some(4)`),
 /// or `None` when there is no trailing numeric `.<n>`. Used to pick the highest
@@ -567,7 +581,23 @@ impl RawCdotTranscript {
             .exons
             .iter()
             .filter_map(|e| {
-                if e.len() >= 5 {
+                if e.len() < 5 {
+                    // Dropping a row is no longer free. Since #1619 the CDS
+                    // bounds are mapped THROUGH this exon table
+                    // (`collapsed_to_flat_tx_pos`), so a silently discarded exon
+                    // shortens `exon_bases_seen` and moves `cds_start`/`cds_end`
+                    // by that exon's span — where previously it only cost the
+                    // table one row. Account for the drop rather than let a
+                    // malformed record shift the CDS in silence.
+                    log::warn!(
+                        "cdot exon row for {} has {} fields (expected at least 5); \
+                         dropping it, which will shift this transcript's CDS bounds",
+                        self.gene_name.as_deref().unwrap_or("<unknown gene>"),
+                        e.len()
+                    );
+                    return None;
+                }
+                {
                     // cdot stores genomic bounds 0-based half-open (`alt_start`
                     // 0-based inclusive, `alt_end` 0-based exclusive) and cDNA
                     // bounds 1-based **inclusive** (`cds_start_i`/`cds_end_i`).
@@ -597,8 +627,6 @@ impl RawCdotTranscript {
                         None
                     };
                     Some((exon, cigar))
-                } else {
-                    None
                 }
             })
             .collect();
@@ -611,8 +639,36 @@ impl RawCdotTranscript {
         // Use transcript-level CDS coordinates (start_codon/stop_codon) if available
         // These are 0-indexed and more reliable than converting from genomic coords
         let (cds_start, cds_end) = if self.start_codon.is_some() || self.stop_codon.is_some() {
-            // Use transcript-level coordinates directly (0-indexed)
-            (self.start_codon, self.stop_codon)
+            // #1619: a cdot record carries TWO transcript coordinate spaces and
+            // labels neither. The exon table's `tx_start`/`tx_end` are absolute
+            // offsets into the transcript *sequence* (so a base unaligned to the
+            // genome still occupies a position), while `start_codon`/`stop_codon`
+            // are offsets into the gap-COLLAPSED transcript — exon bases only.
+            // The two agree for the 474,760 GRCh38 builds whose exon table is
+            // contiguous and disagree for the 58 that are not, where taking the
+            // codon bounds verbatim serves a CDS short by the enclosed gap.
+            //
+            // Measured over those 58 (50 have both a CDS and a sequence on the
+            // prepared reference): reading `start_codon`/`stop_codon` as
+            // collapsed offsets yields a valid ORF 50/50 and reading them as flat
+            // offsets 0/50. `NM_033517.1` is the worked case — cdot says 5157,
+            // RefSeq annotates `CDS 1..5196` and `NP_277052.1` is 1731 aa.
+            //
+            // Everything downstream — `cds_to_tx`, `tx_to_cds`, the exon walk in
+            // `TranscriptMetadata` — indexes the flat sequence, so map into that
+            // space here and let the rest of the crate keep one space. Note the
+            // end is mapped through its LAST CODING BASE rather than directly:
+            // a CDS flush with an exon boundary must not absorb the gap that
+            // follows it, which is what "add every preceding gap" would do.
+            (
+                self.start_codon
+                    .map(|start| collapsed_to_flat_tx_pos(&exons, start)),
+                self.stop_codon.map(|stop| match stop.checked_sub(1) {
+                    Some(last) => collapsed_to_flat_tx_pos(&exons, last).saturating_add(1),
+                    // An empty CDS carries no last coding base to map.
+                    None => 0,
+                }),
+            )
         } else {
             // Fall back to converting genomic CDS coordinates to transcript coordinates
             match (build.cds_start, build.cds_end) {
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
## Ordering correction: this depends on #1735, not just #1724

This PR's description says only *"Stacked on #1724 — merge that first"*, and its base branch is #1724's head. That is incomplete.

This PR makes cdot's `cds_start`/`cds_end` **flat** transcript offsets (`collapsed_to_flat_tx_pos`, plus the two cache-format bumps). But the consumer that reads them — `CoordinateMapper::cds_to_tx`/`tx_to_cds` — is still the **exon-walking** version on `main` (`src/convert/mapper.rs:93`, `cds_to_tx_exon_aware`, which branches on `has_gaps` and skips the gap while walking). It is **#1735** that deletes that walk and makes both conversions flat.

The two PRs are the two halves of #1619 and both cite it. Landing this one without #1735 feeds a flat CDS bound into a gap-skipping walk on the 58 gapped GRCh38 builds.

**Whether that is merely still-wrong or newly-worse than `main` was not determined** — establishing it would need the suite run against `main + #1724 + #1737` with #1735 absent. Reasoning through `mapper.rs:116-147`, the 3'UTR path plausibly improves and the `tx_to_cds` UTR/CDS classification boundary plausibly degrades, but neither is asserted here.

The safe ordering costs nothing: **land #1735 before this**, and retarget this PR onto `main` rather than keeping it stacked on #1724. Its real delta is 1 commit / 3 files.

nh13:
**This must not merge as-is once `CDOT_DATA_VERSION` moves to `data_v0.2.34` — `collapsed_to_flat_tx_pos` double-corrects there.** Raised from #1887, which is the bump; posting here because this PR is open and `MERGEABLE` today and the two changes are order-dependent.

This PR's own `## Upstream` section names the failure mode: the compensation "becomes a *double*-correction if cdot ever changes its data to serve flat bounds, landing `cds_end` at 5235 instead of 5196". That has now happened — `data_v0.2.34` ships the fix for the second half of SACGF/cdot#123.

### Measured on the real record, both releases

Streamed `NM_033517.1` out of each release's `refseq.GRCh38` JSON:

| release | `start_codon` | `stop_codon` | exon table | cdna hole |
|---|---|---|---|---|
| `cdot-0.2.32.refseq.GRCh38.json` | 0 | **5157** | 22 exons, cdna 1–7096, 7057 covered | (1302, 1342) |
| `cdot-0.2.34.refseq.GRCh38.json.gz` | 0 | **5196** | 22 exons, cdna 1–7096, 7057 covered | (1302, 1342) |

Identical exon table, identical 39-base hole; `stop_codon` moves by exactly the gap, onto RefSeq's own `CDS 1..5196`. So under 0.2.34 the verbatim copy this PR replaces is already correct, and `map(stop_codon - 1) + 1` walks collapsed 5195 past the hole to flat 5234, yielding `cds_end = 5235` — the number this PR predicted.

### Measured on this PR's own fixture

Ran this branch's `src/data/cdot.rs` with one extra probe alongside the existing gapped tests, using its own `load_gapped` helper (two exons, 40-base transcript-coordinate hole):

```
PROBE1887 collapsed_input(10,180) -> cds_start=Some(10) cds_end=Some(220)
PROBE1887 wholetx_input(10,220)   -> cds_start=Some(10) cds_end=Some(260)
```

`load_gapped(10, 220)` is what a 0.2.34 record for that geometry states. The correct `cds_end` is `220`; the branch produces `260`, the 40-base gap applied a second time. In the same run all four existing tests passed — including `cdot_cds_bounds_are_unchanged_on_a_gapless_exon_table`, which is identity on both arms — so nothing in the current test set witnesses this.

### What this means for merge order

`collapsed_to_flat_tx_pos` is applied unconditionally in the `start_codon`/`stop_codon` branch of `from_genome_build`; there is no check for which coordinate space the record is in. Exactly one of {this PR, the 0.2.34 bump} can be correct at a time, and merging this without a guard turns the bump into a silent regression on the 58 gapped GRCh38 builds rather than a fix.

Two constraints a guard has to meet, neither of which is satisfiable today without new plumbing:

- **The record cannot discriminate itself.** For `NM_033517.1` both 5157 and 5196 sit below the exon-covered length (7057), so neither value is out of range for the collapsed reading. Shape detection alone does not answer which space the record is in.
- **The release is recorded in the manifest, not in the loader.** `cdot_data_versions` (#1001) is per-artifact on `ReferenceManifest` and is read only by `ferro check` (`src/bin/ferro.rs:3839`); nothing under `src/data/` or in provider construction sees it. A version-aware guard needs that value threaded through — and needs an answer for *absent*, because the prepared reference in use locally predates the field and records no version at all.

Nothing was pushed to this branch and no change is proposed here; flagging so the ordering is a decision rather than an accident.

nh13:
Converted to draft to keep it out of the merge queue, per maintainer decision — not a judgement on the change itself, which is correct against cdot `data_v0.2.32`.

The reason is the measured double-correction under `data_v0.2.34` recorded above: `collapsed_to_flat_tx_pos` maps outward unconditionally, so on a record whose codon fields are already whole-transcript it applies the gap twice (`260` where `220` is correct, on this PR's own gapped fixture via its own `load_gapped` helper). All four of this PR's tests pass in that same run, so its suite cannot witness it.

What would take it out of draft: a version- or shape-aware guard so the mapping is applied only to a gap-collapsed record. Note two constraints found while measuring — the record **cannot discriminate itself** (both readings sit below the transcript's exon-covered length, so neither is out of range), and `cdot_data_versions` is manifest-only, read solely by `ferro check` and invisible to `src/data/`, so there is no local version to key on. #1887 carries the feasibility analysis for the bump itself.

nh13:
@coderabbitai review

PR #1740, opened 2026-08-12
fix(benchmark): read internal cdot transcript coordinates as 0-based

### LINKED ISSUE #1738: benchmark cache reads internal cdot transcript coordinates as 1-based, panicking on every first exon
`src/benchmark/cache.rs` applies `coords::hgvs_pos_to_index` — an unconditional `assert!(pos > 0)` — to the `tx_start`/`tx_end` pair of every internal-cdot exon, under a comment claiming those coordinates are 1-indexed inclusive.

Since #742 they are **0-based half-open**: `convert::mapper`'s `test_cdot_tx_coordinates_are_0_based` asserts `exons[0][2] == 0` against real cdot data, and `src/data/cdot.rs:16` / `:418` declare the convention on `CdotTranscript`. So the first exon of every transcript panics.

## It panics in two places, not one

- `build_annotations_from_cdot` — `:653`, reached from `:467`
- `calculate_genomic_cds` — `:2758-2763`, carries the identical false comment (`// 1-indexed in cdot`) and the identical assertion, reached from `build_nc_annotations_with_selectors:2673`

Different callers, so this is a second live panic rather than a duplicate.

```
thread '...' panicked at src/benchmark/cache.rs:653:24:
1-based HGVS position cannot be 0 (no position 0 exists on any HGVS axis);
use hgvs_pos_to_index_checked if the position is not known to be valid
```

Reproduces with a two-exon transcript in the internal convention, `exons: [[1000,1073,0,73],[2000,2050,73,123]]`.

## Related: one convention, two opposite treatments, 600 lines apart

`parse_genbank_record`'s `-1` at `:1230` is **correct** for its own source — GenBank really is 1-based inclusive — but nothing in the file states which source format each block reads, so the same-looking conversion appears twice with opposite justifications. Worth naming the source at each site.

The path is behind the `benchmark` feature, which is why the panic has gone unnoticed.

### LINKED ISSUE #1763: calculate_genomic_cds emits a minus-strand CDS span one base low at both ends
Found while reviewing #1740, and it is a defect in that PR's rewrite rather than one it inherited — the function panicked on its first exon before #1740, so there is no prior behaviour to compare against.

## The claim

`calculate_genomic_cds` (`src/benchmark/cache.rs`) maps both `cds_start` and `cds_end` through the pointwise inverse of `data::cdot::genomic_to_tx_pos`, then returns them low-to-high. That is correct on the plus strand and off by one on the minus.

`cds_end` is **exclusive** (`CdotTranscript::cds_end`: "End of CDS in transcript coordinates (0-based, exclusive)"). The pointwise inverse is order-*preserving* on the plus strand and order-*reversing* on the minus, so mapping an exclusive endpoint through it yields an exclusive genomic endpoint only on the plus strand. On the minus strand it lands one base **below** the last CDS base instead of one above the first, and because the result is then sorted low-to-high, both endpoints come out one too low.

## Measured

Two transcripts with identical transcript geometry differing only in genomic direction, CDS `tx [10, 100)`:

```
plus  : exon A genome [1000,1050) = tx [0,50) ; exon B [2000,2073) = tx [50,123)
minus : exon A genome [2000,2050) = tx [0,50) ; exon B [1000,1073) = tx [50,123)
```

Truth, derived from `genomic_to_tx_pos`'s own formula rather than from this function: first CDS base is tx 10, last is tx 99.

| strand | first base | last base | correct half-open span | #1740 emits |
|---|---|---|---|---|
| plus | g 1010 | g 2049 | `[1010, 2050)` | `(1010, 2050)` ✔ |
| minus | g 2039 | g 1023 | `[1023, 2040)` | `(1022, 2039)` ✘ |

A plus-only test cannot see this, which is why it survived.

## The PR pins the wrong answer

`genomic_cds_mapping_walks_a_minus_strand_exon_from_its_far_end` asserts `(1052, 1062)` for CDS `tx [10, 20)` on a minus-strand exon `genome [1000, 1073) = tx [0, 73)`. The first CDS base is tx 10 -> g 1062 and the last is tx 19 -> g 1053, so the correct span is `[1053, 1063)`. The test pins the off-by-one rather than catching it, so this needs an author decision rather than a silent test edit.

## Second, smaller consequence

Mapping the exclusive endpoint also makes a CDS flush with the transcript end **decline**: `cds_end == 123` on a 123-base transcript is outside every exon under a half-open containment test, so `calculate_genomic_cds` returns `None` and the CDS feature is dropped entirely from the emitted annotation for any transcript with no 3' UTR. `genomic_cds_mapping_declines_a_position_past_the_last_exon` pins that as intended behaviour.

## Suggested fix

Map both **inclusive** endpoints — `cds_start` and `cds_end - 1` — and recover the half-open span on the genomic axis afterwards as `(min, max + 1)`, where `+1` means the same thing on either strand:

```rust
let cds_last = cds_end.checked_sub(1)?;   // an empty CDS names no base
...
    g_cds_end = genomic_of(exon, cds_last);
...
(Some(s), Some(e)) => Some((s.min(e), s.max(e) + 1)),
```

Verified: plus `(1010, 2050)`, minus `(1023, 2040)`, the flush-with-transcript-end case becomes representable, and `cds_start == cds_end == 0` declines rather than wrapping. The existing plus-strand test `genomic_cds_mapping_reads_internal_cdot_tx_coordinates_as_zero_based` is unchanged by it.

## Related, same PR, worth folding in

`src/benchmark/cache.rs`'s module-level basis table (lines 10-14) still reads

```
| cdot transcript (tx_start/tx_end) | 1-based | Inclusive |
```

which is exactly the false claim #1740 exists to correct, sitting three lines above `use crate::coords::hgvs_pos_to_index;`. It is a correct description of the **raw** cdot JSON and a wrong one for the **internal** `CdotTranscript` this module is handed — the same unlabelled-layer ambiguity #1742 is fixing in `coords/mod.rs` and `data/cdot.rs`. #1740 corrects the two function comments and leaves the most authoritative-looking statement in the file asserting the opposite. This is the third such table; labelling it by layer would finish the set.

Scope note: this path is behind the `benchmark` feature and affects generated mutalyzer-comparison annotation JSON, not any normalized output.

nh13:
Closes #1738. Closes #1763.

`benchmark/cache.rs` applied `coords::hgvs_pos_to_index` — an unconditional `assert!(pos > 0)` — to the `tx_start`/`tx_end` pair of every internal-cdot exon, on a comment claiming those coordinates are 1-indexed inclusive. Since #742 they are 0-based half-open: `convert::mapper`'s `test_cdot_tx_coordinates_are_0_based` asserts `exons[0][2] == 0` against real cdot data, and `data::cdot` declares the convention on `CdotTranscript`. So the first exon of every transcript panicked.

It panicked in **two** places, not one. `build_annotations_from_cdot` is the site the audit named; `calculate_genomic_cds` carries the identical false comment and the identical assertion, reached from a different caller. Both now copy the transcript coordinates through unconverted, which is what mutalyzer's 0-based half-open annotation format wants anyway.

The GenBank-sourced conversion in `parse_genbank_record` is correct and unchanged — GenBank really is 1-based inclusive — but nothing in the file said which source each block read, so one convention appeared twice with opposite treatments 600 lines apart. Each conversion site now names its source format and points at the other.

The path is behind the `benchmark` feature, which is why the panic went unnoticed. `cdot_coordinate_basis_tests` sits beside the functions it tests rather than at end-of-file, deliberately — appending a second `#[cfg(test)] mod` at EOF collides positionally with the one #1732 adds, and a careless resolve silently drops one of two test modules.

## A defect this PR's own review found in this PR, now fixed — #1763

Reviewing this branch turned up a **minus-strand CDS span one base low at both ends**, and it is a defect this rewrite **introduced** rather than one it inherited: the function panicked on its first exon before, so there was no prior behaviour to compare against. That made the first version of this PR a trade of a loud panic for a quiet one-base coordinate error, which is the worse failure mode for a coordinate function. It is fixed here and the fix has its own end-to-end guard.

**The mechanism.** `cds_end` is exclusive (`CdotTranscript::cds_end`: "0-based, exclusive"). The pointwise inverse of `data::cdot::genomic_to_tx_pos` is order-*preserving* on the plus strand and order-*reversing* on the minus, so pushing an exclusive endpoint through it yields an exclusive genomic endpoint only on the plus strand. On the minus it lands one base **below** the last coding base rather than one above the first, and sorting the pair low-to-high then carries the error to both ends.

Measured over two transcripts with identical transcript geometry differing only in genomic direction, CDS `tx [10, 100)`:

| strand | first coding base | last coding base | correct half-open span | previously emitted |
|---|---|---|---|---|
| plus | g 1010 | g 2049 | `[1010, 2050)` | `(1010, 2050)` — correct |
| minus | g 2039 | g 1023 | `[1023, 2040)` | `(1022, 2039)` — one low at both ends |

A plus-only test cannot see this, which is why it survived; and this PR's `genomic_cds_mapping_walks_a_minus_strand_exon_from_its_far_end` **pinned the off-by-one** rather than catching it.

**The fix.** Both *inclusive* endpoints are placed — `cds_start` and `cds_end - 1` — and the half-open genomic span is recovered afterwards as `(min, max + 1)`, which means the same thing on either strand. An empty CDS (`cds_start >= cds_end`, or `cds_end == 0`) names no base and declines rather than underflowing or inverting.

Two consequences beyond the span itself. A CDS **flush with the transcript's 3' end** is now representable: `cds_end == 123` on a 123-base transcript used to fall outside every exon under a half-open containment test, so `calculate_genomic_cds` returned `None` and the CDS feature was dropped from the emitted annotation for any transcript with no 3'UTR. And the module-level basis table at the top of the file is **relabelled by layer** — it still read `cdot transcript (tx_start/tx_end) | 1-based | Inclusive`, i.e. the exact false claim this PR exists to correct, sitting as the most authoritative-looking statement in the file. It is a correct description of the *raw* cdot JSON and a wrong one for the *internal* `CdotTranscript` this module is handed, so each row now names its layer.

## Correction to an earlier version of this description

This description previously said that repairing the basis "also settles its arithmetic". **It did not**, and a reviewer would reasonably have taken it as covering the case above. The sentence beside it — "the plus strand walks from `g_start` and the minus strand from `g_end - 1` — the inverse of `data::cdot::genomic_to_tx_pos`" — is true of the *pointwise* placement and was never the defect. The defect was one level up: applying that correct pointwise inverse to an **exclusive** endpoint. Both statements are restated above so the distinction is on the record rather than implied.

## Tests

`cdot_coordinate_basis_tests` now carries seven cases. Four fail on the pre-fix arithmetic, verified by restoring it in place and re-running:

```
FAIL genomic_cds_span_on_a_minus_strand_covers_its_first_and_last_coding_base
       left: Some((1022, 2039))   right: Some((1023, 2040))
FAIL genomic_cds_mapping_walks_a_minus_strand_exon_from_its_far_end
       left: (1052, 1062)         right: (1053, 1063)
FAIL genomic_cds_mapping_places_a_cds_flush_with_the_transcript_end
       left: None                 right: Some((1010, 2050))
FAIL genomic_cds_mapping_declines_a_cds_that_names_no_base
```

The minus-strand guard is stated **end-to-end** as well as on the helper: it builds a geometrically real minus-strand transcript — transcript coordinates increasing as genomic ones decrease, which flipping the strand field of the plus fixture does *not* produce — and asserts the CDS feature `build_nc_annotations_with_selectors` emits carries `[1023, 2040)`. The three that pass on both are the controls: the plus-strand span, the exon features, and a position genuinely past the last exon, which must still decline.

`cargo fmt --all --check` and `cargo clippy --features dev,benchmark --lib --all-targets -- -D warnings` clean.

Representation-Change: none. Benchmark-only cache builder; the diff is one file, `src/benchmark/cache.rs`, and no watched directory is touched.

### DIFF
--- a/src/benchmark/cache.rs
+++ b/src/benchmark/cache.rs
@@ -5,14 +5,23 @@
 //!
 //! # Coordinate System
 //!
-//! This module handles mixed coordinate systems from cdot:
+//! This module handles mixed coordinate systems, and the basis depends on the
+//! **layer** a value was read from, not merely on the field's name. The rows
+//! below are labelled by layer for that reason.
 //!
-//! | Field | Basis | Notes |
-//! |-------|-------|-------|
-//! | cdot genomic (genome_start/genome_end) | 0-based | Half-open `[start, end)` |
-//! | cdot transcript (tx_start/tx_end) | 1-based | Inclusive |
-//! | cdot CDS (cds_start/cds_end) | 0-based | Half-open |
-//! | Mutalyzer cache coordinates | Mixed | Follows mutalyzer conventions |
+//! | Layer | Field | Basis | Notes |
+//! |-------|-------|-------|-------|
+//! | internal [`CdotTranscript`] | genomic (`genome_start`/`genome_end`) | 0-based | Half-open `[start, end)` |
+//! | internal [`CdotTranscript`] | transcript (`tx_start`/`tx_end`) | 0-based | Half-open `[start, end)`, since #742 |
+//! | internal [`CdotTranscript`] | CDS (`cds_start`/`cds_end`) | 0-based | Half-open `[start, end)` |
+//! | raw cdot JSON | transcript (`tx_start`/`tx_end`) | 1-based | Inclusive; normalized away by `crate::data::cdot` before this module sees it |
+//! | GenBank flat file | all | 1-based | Inclusive; converted in `parse_genbank_record`* |
+//! | Mutalyzer cache | all | 0-based | Half-open, so internal-cdot values copy through verbatim |
+//!
+//! The internal rows are what every function here is handed. Reading the raw
+//! cdot row as if it described them is what #1738 fixed and #1763 corrected the
+//! residue of: `hgvs_pos_to_index` asserts a valid 1-based position, and the
+//! first exon of every internal transcript has `tx_start == 0`.
 //!
 //! For type-safe coordinate handling, see [`crate::coords`].
 
@@ -611,10 +620,23 @@ fn load_fasta_sequences_parallel<P: AsRef<Path>>(
 
 /// Build mutalyzer-compatible annotations JSON from cdot transcript data.
 ///
-/// Coordinate conversion:
-/// - cdot exon coordinates are 1-indexed, inclusive (GenBank-style)
-/// - mutalyzer expects 0-indexed, half-open coordinates
-/// - Convert: start = cdot_start - 1, end = cdot_end (no change for half-open)
+/// # Coordinate basis: internal cdot, NOT GenBank
+///
+/// The [`CdotTranscript`] handed to this function is ferro's **internal**
+/// normalized form (see `crate::data::cdot`), whose transcript coordinates —
+/// the `tx_start`/`tx_end` pair in each exon, and `cds_start`/`cds_end` — are
+/// already **0-based half-open** `[start, end)`, as of #742. Mutalyzer wants
+/// 0-based half-open too, so **no basis conversion is owed here**: the values
+/// are copied through verbatim.
+///
+/// This is deliberately different from the GenBank-sourced blocks in this file
+/// (`parse_genbank_record`, `parse_genbank_record_full`), which read 1-based
+/// inclusive coordinates off a flat-file record and therefore *do* subtract one.
+/// Each conversion site in this file names its source format for that reason.
+///
+/// Applying the GenBank `-1` here is not a rounding error but a hard failure:
+/// `hgvs_pos_to_index` asserts its argument is a valid 1-based position, and the
+/// first exon of every transcript has `tx_start == 0`.
 ///
 /// mol_type is set based on accession prefix:
 /// - NM_/XM_ = mRNA (coding transcripts)
@@ -642,16 +664,14 @@ fn build_annotations_from_cdot(
         ("mRNA", "mRNA", true) // Default to mRNA for unknown types
     };
 
-    // Build exon features from cdot exon data
-    // cdot exons (internal format): [genome_start, genome_end, tx_start, tx_end]
-    // tx_start and tx_end are 1-indexed inclusive, convert to 0-indexed half-open
+    // Build exon features from internal-cdot exon data:
+    //   [genome_start, genome_end, tx_start, tx_end]
+    // tx_start/tx_end are already 0-based half-open (see this function's docs),
+    // which is exactly what mutalyzer wants, so both are copied through as-is.
     let mut exon_features: Vec<serde_json::Value> = Vec::new();
     for (i, exon) in tx.exons.iter().enumerate() {
-        // Convert 1-indexed inclusive to 0-indexed half-open:
-        // - start: subtract 1 (1-indexed -> 0-indexed)
-        // - end: keep as-is (inclusive -> exclusive)
-        let tx_start = hgvs_pos_to_index(exon[2]); // 1-indexed → 0-indexed
-        let tx_end = exon[3] as usize; // Don't subtract 1 for half-open end
+        let tx_start = exon[2] as usize;
+        let tx_end = exon[3] as usize;
         exon_features.push(json!({
             "type": "exon",
             "location": {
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1741, opened 2026-08-12
fix(python): validate the numeric conversion entry points at the boundary

### LINKED ISSUE #1739: Python numeric conversion entry points build positions from unvalidated integers
Five `CoordinateMapper` methods exposed to Python take raw integers and construct position structs directly, bypassing every check the parser applies. Three distinct escapes, measured on a programmatically-built two-exon transcript (exon1 tx 1..100 / g 1001..1100; intron g 1101..2000; exon2 tx 101..200 / g 2001..2100; CDS tx 11..190):

```
c_to_g(0)                 -> ('chr1', 1011)          # identical to c.1
c_to_g(1)                 -> ('chr1', 1011)
c_to_g(90, i64::MAX)      -> ('chr1', 9223372036854776907)
c_to_g(91, i64::MIN)      !! PanicException: attempt to negate with overflow
c_to_n(90, i64::MAX)      -> (100, 9223372036854775807)
n_to_c(195, ds=True)      -> (5, None, True)
n_to_c(0)                 -> (-11, None, False)
```

## 1. `c_to_g(id, 0)` bypasses the parser's `E1003`

The parser rejects `c.0` as `E1003 InvalidPosition` on the spec's authority (`background/numbering.md:31`, "there is no nucleotide `c.0`"). The numeric entry point answers it identically to `c.1`. Two entry points, two answers.

## 2. A sentinel offset reaches unchecked arithmetic

A non-zero offset routes to `cds_to_genomic_with_intron` → `Transcript::intronic_to_genomic`, whose arms are `Some(g_start + offset as u64 - 1)` and `Some(g_end - (-offset) as u64 + 1)`. Note `-(i64::MIN)` overflows *before* the cast. The #1087 / PR #1088 guard at `data/mapping.rs:326` is not on this path. Debug panics; release wraps silently.

## 3. The sentinel is returned to the caller

`c_to_n` returns an `i64::MAX` offset to Python as `9223372036854775807`, a number indistinguishable from a measured distance. `n_to_c` accepts `downstream=True` and answers it as though the flag were false.

## Also

All five methods leave at least one side's coordinate basis undocumented: `c_to_g` and `c_to_n` document neither side; `c_to_p`, `g_to_c` and `n_to_c` document the input's basis and not the output's.

Note the audit that surfaced this attributed escape 2 to `c_to_n`; it is `c_to_g`. `c_to_n` copies the offset through without arithmetic, which is escape 3.

### LINKED ISSUE #1765: Python numeric entry points: offset guard keyed on sentinel identity, so i64::MIN+1 still panics across FFI
Found while reviewing #1741, verified against an actually-imported `maturin develop --features python` module (debug profile), not from reading.

## What #1741 closes, and what it leaves open

#1741 adds `reject_sentinel_offset`, which refuses **exactly** the two unknown-offset sentinel values `i64::MAX` / `i64::MIN`. The hazard downstream is not the sentinel identity, though — it is the **magnitude**. `Transcript::intronic_to_genomic` (`src/reference/transcript.rs:806-827`) does unchecked mixed-width arithmetic on any offset it is handed:

```rust
Some(g_start + offset as u64 - 1)      // plus,  offset > 0
Some(g_end - (-offset) as u64 + 1)     // plus,  offset <= 0   <-- :814
```

So stepping one away from the sentinel reproduces the entire bug class the PR exists to close.

## Measured

Against the PR's own two-exon fixture, on the PR's own branch, module imported:

```
control c_to_g(90, +5)                  -> ('chr1', 1105)
c_to_g(90, i64::MAX)   [sentinel]       !! ProjectionError: cannot resolve unknown intronic offset (`+?`)
c_to_g(90, i64::MIN)   [sentinel]       !! ProjectionError: cannot resolve unknown intronic offset (`-?`)
c_to_g(90, i64::MAX - 1)                -> ('chr1', 9223372036854776906)      <-- garbage coordinate
c_to_g(90, i64::MIN + 1)                !! pyo3_runtime.PanicException: attempt to subtract with overflow
                                           (src/reference/transcript.rs:814)
```

Compare the escapes #1741's own description lists as the things it is fixing:

```
c_to_g(90, i64::MAX)     -> ('chr1', 9223372036854776907)
c_to_g(91, i64::MIN)     !! PanicException: attempt to negate with overflow
```

`MAX-1` and `MIN+1` are the same two outcomes, one base away, still reachable. In a `--release` wheel the panic becomes a silent wrap instead, which is the worse half.

## The tests pin the gap open

Both new tests assert the neighbouring values are admitted, so the gap is currently a stated property rather than an oversight:

- `a_measured_offset_is_admitted` includes `Some(i64::MAX - 1)` in its admitted set.
- `only_the_two_exact_sentinels_are_sentinels` asserts `!is_unknown_offset(Some(OFFSET_UNKNOWN_NEGATIVE + 1))`.

The reasoning given — "Only the two exact values are sentinels; everything else is a number the conversion is entitled to judge for itself" — is right about what a *sentinel* is and wrong about what the conversion does with the rest: it does not judge them, it overflows on them.

## The decision this needs

What bound should a measured intronic offset be held to at the boundary? Options, roughly in increasing strictness:

1. **Make the arithmetic total.** Change `intronic_to_genomic` to `checked_add`/`checked_sub` and return `None`, so any offset that cannot land on a real coordinate declines instead of wrapping. Fixes every caller, not just the Python one, and needs no policy choice about what is "reasonable". Probably the right primary fix; the boundary guard then becomes belt-and-braces.
2. **Bound by the intron.** Refuse an offset whose magnitude exceeds the intron it is measured into (or, past a terminus, the contig). Most meaningful, most work, and needs a rule for the no-genomic-data case.
3. **Bound by a sanity ceiling** at the boundary (e.g. anything that cannot fit a genomic coordinate). Cheap, arbitrary, and needs a number nobody has a principled source for.

Option 1 is the one that closes the class rather than a range of it. Worth deciding before #1741 lands, because the PR's tests currently assert the opposite.

## Scope note

This is reachable from Python today on `main` as well — #1741 does not introduce it. What #1741 does is narrow the entry to the point where the remaining hole is easy to mistake for closed.

nh13:
Closes #1739. Closes #1765.

Five `CoordinateMapper` methods exposed to Python took raw integers and built position structs directly, bypassing every check the parser applies. Three escapes, all measured on a programmatically-built two-exon transcript before any guard existed:

```
c_to_g(0)                -> ('chr1', 1011)      # identical to c.1
c_to_g(90, i64::MAX)     -> ('chr1', 9223372036854776907)
c_to_g(91, i64::MIN)     !! PanicException: attempt to negate with overflow
c_to_n(90, i64::MAX)     -> (100, 9223372036854775807)
n_to_c(195, ds=True)     -> (5, None, True)
n_to_c(0)                -> (-11, None, False)
```

**Base zero.** `reject_zero_base` refuses on all four `c.`/`n.` entry points as `ParseError` / **E1003**, reusing the preprocessor's exact wording ("Position 0 is not valid in HGVS notation") and hint, on the spec's authority (`background/numbering.md:31`, "there is no nucleotide `c.0`"). Negative 5'UTR bases are explicitly admitted. Every other implementation agrees: biocommons/hgvs raises "BaseOffsetPosition base may not be 0" in the type itself, Mutalyzer rejects it, and `mutalyzer-crossmapper`'s docs open with "There is no zero (0) in any of the HGVS numbering systems". `c.0` occurs 0 times in 346 Mutalyzer and 88 biocommons corpus cases.

**Sentinel offset.** `reject_sentinel_offset` refuses `i64::MAX`/`i64::MIN` before any arithmetic, reusing #1088's message and `UnsupportedProjection` (E4003) so there is one vocabulary.

**Unreadable return value.** `reject_sentinel_result` refuses on the way *out* of `c_to_n`, stated over the produced value, so a sentinel cannot come back to Python as `9223372036854775807`. `reject_unhonoured_downstream` refuses `downstream=True` rather than answering it as though the flag were false.

All five docstrings now state both sides' coordinate basis and the exception each refusal raises.

## A defect this PR's own review found in this PR, now fixed — #1765

Reviewing this branch, against an actually-imported `maturin develop --features python` module, found that the sentinel guard closes the **entry** to a hazard whose real key is **magnitude**, leaving the remaining hole easy to mistake for closed. `Transcript::intronic_to_genomic` did unchecked mixed-width arithmetic on any offset it was handed, so stepping one value away from each sentinel reproduced the whole class:

```
c_to_g(90, +5)             -> ('chr1', 1105)                       # control
c_to_g(90, i64::MAX)       !! ProjectionError   (this PR's guard)
c_to_g(90, i64::MIN)       !! ProjectionError   (this PR's guard)
c_to_g(90, i64::MAX - 1)   -> ('chr1', 9223372036854776906)        # still reachable
c_to_g(90, i64::MIN + 1)   !! PanicException: attempt to subtract with overflow
                              (src/reference/transcript.rs:814)
```

Those are the same two outcomes the description above lists as the things being fixed, one base away. In a `--release` wheel the panic becomes a silent wrap, which is the worse half. Neither value is a sentinel, so **no widening of the sentinel predicate could reach them**.

**The fix is #1765's option 1, plus its option 2 where option 2 is free.** `intronic_to_genomic` is now total and bounded by the intron the offset is measured into:

- `offset.unsigned_abs()` replaces `(-offset) as u64`, which is what `i64::MIN` needs — it has no positive twin, so the negation aborted before the cast. The four arms then use `checked_add`/`checked_sub`.
- Totality alone does **not** close `i64::MAX - 1`: `u64` has room for it, so the arithmetic never overflows and the old code simply returned a coordinate-shaped number ~9.2e18 bases outside a 900-base intron. The result must therefore land inside `[genomic_start, genomic_end]` of the intron it was measured into, else the function declines — an offset larger than its own intron names no intronic base, so it names no genomic coordinate either.

That fixes every caller rather than the Python one, which is why it lives in `src/reference/transcript.rs`. `cds_to_genomic_with_intron` already turns `None` into a `ConversionError`, so the entry point reports `ProjectionError` with no new plumbing.

**The two tests that pinned the gap open are corrected, and both keep their assertions.** `a_measured_offset_is_admitted` still admits `i64::MAX - 1` at this guard — that is the guard's contract — but its comment claimed "everything else is a number the conversion is entitled to judge for itself", which was an assumption and is now a fact; it points at the transcript-level tests that hold the other side. `only_the_two_exact_sentinels_are_sentinels` is unchanged in substance: a sentinel is a sentinel and `i64::MIN + 1` is a number, so widening *that* predicate would have been the wrong repair. `reject_sentinel_offset`'s own doc no longer claims to be what protects the arithmetic — it runs before the transcript is fetched and holds no intron to measure against, so what survives there is the diagnostic that names `+?` / `-?` as unknown-offset markers.

**Tests, and their pre-fix failures.** Three new Rust cases in `reference::transcript::tests`, over an interior-intron transcript with the same geometry as the Python fixture, so the two sets of numbers are the same numbers:

```
FAIL intronic_to_genomic_declines_an_offset_whose_arithmetic_would_overflow
       panicked at src/reference/transcript.rs:814: attempt to negate with overflow
FAIL intronic_to_genomic_declines_an_offset_that_leaves_its_own_intron
       left: Some(2001)   right: None
FAIL intronic_to_genomic_bounds_the_offset_on_the_minus_strand_too
       left: Some(104)    right: None
```

Each carries its own negative control, so a guard that refused every large offset would fail them: the intron spans g 1101..2000, and `+900` → 2000 and `-900` → 1101 must still resolve.

`TestOversizedOffsetIsRefused` states the same properties across FFI, against a real `maturin develop --features python` module. Rebuilt with the pre-fix arithmetic restored in place: **8 failed, 1 passed**, the one pass being the end-to-end negative control, which must pass on both.

```
FAILED test_a_sentinel_neighbour_declines_rather_than_panicking[90-9223372036854775806]
         Failed: DID NOT RAISE ProjectionError
FAILED test_a_sentinel_neighbour_declines_rather_than_panicking[90--9223372036854775807]
         pyo3_runtime.PanicException: attempt to subtract with overflow
FAILED test_a_sentinel_neighbour_declines_rather_than_panicking[91-9223372036854775806]
FAILED test_a_sentinel_neighbour_declines_rather_than_panicking[91--9223372036854775807]
FAILED test_an_offset_larger_than_its_intron_declines[901]
FAILED test_an_offset_larger_than_its_intron_declines[-901]
FAILED test_an_offset_larger_than_its_intron_declines[1000000]
FAILED test_an_offset_larger_than_its_intron_declines[-1000000]
```

## Two collateral items, reported rather than smoothed over

`test_issue_1018_typed_exceptions.py::TestProjectionError` used `c_to_p(id, 0)` as its vehicle for `ProjectionError` typing — it only reached the converter because a zero base fell into the 5'UTR arm, i.e. the right exception class for the wrong reason. Moved to `c.-10`, a genuine 5'UTR position failing at the same raise site, with a comment recording why. The test's own subject (exception typing and `ValueError`/`RuntimeError` back-compat) is unchanged.

`python/ferro_hgvs/__init__.pyi` carries duplicate copies of these five docstrings and now drifts from the runtime ones. Left deliberately: no test enforces stub/runtime parity, and that file is edited by another in-flight branch. Real follow-up.

## Verification

The two build profiles genuinely differ here and a first attempt hid it: `maturin develop` makes an *editable* install, so both venvs silently loaded the same `.so`. Rebuilt as two non-editable wheels in separate venvs, as CI does — only then does `c_to_g(91, i64::MIN)` show as `PanicException` in debug and a silent wrap to `chr1:9223372036854777809` in release.

New suite against the unfixed build: **21 failed, 3 passed** in each profile — the 3 passing are the negative controls (`c.-1`, measured offsets `c.90+5`→1105 / `c.91-5`→1996, `downstream=False`), which confirms the expected values are right before the guard exists.

With the #1765 fix in: full Rust gate `cargo nextest run --features dev --lib --test it` **10237/10237 passed**, exit 0 — the intron bound moves nothing. `pytest tests/python/` **930 passed, 8 skipped**. `cargo fmt --all --check`, `cargo clippy --features dev --lib --all-targets -- -D warnings` and `cargo clippy --features python --lib -- -D warnings` all exit 0.

Representation-Change: none. The diff is `src/python.rs` and `src/reference/transcript.rs` plus two files under `tests/python/`; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched.

### DIFF
--- a/src/python.rs
+++ b/src/python.rs
@@ -220,6 +220,213 @@ fn projection_error_at(i: usize, e: &crate::error::FerroError) -> PyErr {
     )
 }
 
+// ============================================================================
+// Numeric coordinate entry points: boundary validation
+// ============================================================================
+//
+// `CoordinateMapper`'s `c_to_g` / `g_to_c` / `c_to_p` / `c_to_n` / `n_to_c`
+// take coordinates as raw integers rather than as a description, so nothing
+// the parser checks is on their path — they build a `CdsPos` / `TxPos` from
+// whatever Python passed. Everything a description would have been checked for
+// has to be checked here instead, or ferro answers questions it has already
+// refused to answer at its other entry point.
+//
+// These helpers are deliberately *pure*: they take and return plain values and
+// never touch the interpreter, so the unit tests at the bottom of this file can
+// exercise them without a Python runtime. Only `BoundaryRejection::into_pyerr`
+// crosses into PyO3.
+
+/// A refused argument at a numeric coordinate entry point.
+///
+/// Carries the library error the refusal corresponds to as well as the message,
+/// so the Python exception reports the same `E####` code and mutalyzer
+/// equivalents the rest of ferro reports for the same refusal — a caller
+/// discriminating on `err.code` sees one vocabulary, not one per entry point.
+struct BoundaryRejection {
+    /// The `ferro_hgvs` exception class to raise (see `ferro_exception`).
+    class: &'static str,
+    /// The message, phrased for a caller who passed integers: it names the
+    /// coordinate rather than an offset into a source string, because at these
+    /// entry points there is no source string.
+    message: String,
+    /// The equivalent library error, read only for its code and mutalyzer map.
+    error: crate::error::FerroError,
+}
+
+impl BoundaryRejection {
+    fn into_pyerr(self) -> PyErr {
+        ferro_typed(self.class, self.message, &self.error)
+    }
+}
+
+/// Spell the coordinate a numeric entry point was handed, for a diagnostic.
+fn render_boundary_position(axis: char, base: i64, utr3: bool) -> String {
+    if utr3 {
+        format!("{axis}.*{base}")
+    } else {
+        format!("{axis}.{base}")
+    }
+}
+
+/// Whether `offset` is one of the parser's unknown-offset sentinels rather than
+/// a measured intronic distance.
+///
+/// Mirrors [`CdsPos::has_unknown_offset`] but takes the bare `Option<i64>`, so
+/// it can be applied to an argument before any position struct is built — which
+/// is the whole point at these entry points.
+fn is_unknown_offset(offset: Option<i64>) -> bool {
+    use crate::hgvs::parser::position::{OFFSET_UNKNOWN_NEGATIVE, OFFSET_UNKNOWN_POSITIVE};
+    matches!(
+        offset,
+        Some(OFFSET_UNKNOWN_POSITIVE) | Some(OFFSET_UNKNOWN_NEGATIVE)
+    )
+}
+
+/// Refuse a base coordinate of zero.
+///
+/// `parse_hgvs` refuses `c.0` / `n.0` as `E1003 InvalidPosition` ("Position 0
+/// is not valid in HGVS notation", `src/error_handling/preprocessor.rs`), on
+/// the spec's authority: `background/numbering.md:31` states there is no
+/// nucleotide `c.0`. That sits *inside* the definition of the numbering axis,
+/// so it is a claim about what the axis contains rather than input hygiene —
+/// and it therefore holds however the coordinate arrives, as an integer just as
+/// much as as text. Every other implementation agrees: biocommons/hgvs raises
+/// "BaseOffsetPosition base may not be 0" from the type itself, Mutalyzer
+/// rejects it, and `mutalyzer-crossmapper`'s documentation opens with "There is
+/// no zero (0) in any of the HGVS numbering systems".
+///
+/// Without this the numeric entry points answered `c.0` **identically to
+/// `c.1`** — `CoordinateMapper::cds_to_tx` routes a zero base through its
+/// `base < 1` arm, which adds nothing, landing on the CDS start.
+///
+/// No ruling record has been committed for `c.0` yet, so the authorities are
+/// recorded here rather than in the ledger. Two sites inside ferro still
+/// disagree with this and with each other (`src/convert/mapper.rs` reads `c.0`
+/// as `c.1`, `src/normalize/mod.rs` reads it as `c.-1`); reconciling those is
+/// separate work, and this refusal is correct whichever way it goes, because a
+/// coordinate that does not exist cannot be converted to one that does.
+fn reject_zero_base(axis: char, base: i64, utr3: bool) -> Result<(), BoundaryRejection> {
+    if base != 0 {
+        return Ok(());
+    }
+    let position = render_boundary_position(axis, base, utr3);
+    let message = format!("Position 0 is not valid in HGVS notation ({position})");
+    Err(BoundaryRejection {
+        class: "ParseError",
+        message: message.clone(),
+        error: crate::error::FerroError::parse_with_diagnostic(
+            0,
+            message,
+            crate::error::Diagnostic::new()
+                .with_code(crate::error::ErrorCode::InvalidPosition)
+                .with_hint("HGVS positions start at 1, not 0"),
+        ),
+    })
+}
+
+/// Refuse an unknown-offset sentinel where a measured distance is required.
+///
+/// `c.N+?` / `c.N-?` parse to `i64::MAX` / `i64::MIN`. Per the spec they denote
+/// an unknown position 3'/5' of the base, unbounded in that direction, so no
+/// coordinate can be derived from them at all — see
+/// [`CdsPos::has_unknown_offset`] and issue #1087.
+///
+/// The `#1087` / PR `#1088` guard lives in `src/data/mapping.rs` and is not on
+/// this path. This is the same refusal, at the entry point that reaches the
+/// conversion without it — the wording and the `UnsupportedProjection` code are
+/// deliberately kept identical to that guard's.
+///
+/// # What this guard does NOT do — read before widening it
+///
+/// It is keyed on sentinel **identity**, and that is all it can be keyed on: it
+/// runs before the transcript is fetched, so it holds no intron to measure a
+/// magnitude against. It is therefore **not** what protects
+/// `Transcript::intronic_to_genomic` from an oversized offset, and an earlier
+/// version of this comment claimed otherwise — describing the `i64::MIN`
+/// negation overflow as if refusing the sentinel closed it. It did not:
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

PR #1742, opened 2026-08-12
refactor: label the coordinate-basis docs by layer, one splice ladder per question

nh13:
Closes the MINOR/NIT cluster of the coordinate-seam audit: m1, m2, m3, m4, n1, n2, m6, n3, plus a corrected diagnosis on m5. **Net behaviour change: none.** When this was first written the PR also carried the `i64::MIN` ladder fix — `c.<base>-?` parses to `offset = Some(i64::MIN)`, and a plain `.abs()` panics under `overflow-checks` and wraps negative in a release build, so every `<= n` rung read true and an unknown-distance intronic position was reported as a canonical splice site at HIGH impact. **That fix has since landed on `main` in #1767**, so what remains here is the de-duplication, which is behaviour-preserving: the consolidated ladder takes its magnitude with `unsigned_abs()` exactly as each former copy already did. The earlier claim that `c.100-?` moves from `splice_site_variant`/HIGH to `intron_variant`/MODIFIER is withdrawn — it describes a transition that has already happened on `main`.

`IvsPos` keeps its own copy of the ladder and is deliberately **not** touched here — that file is in a watched directory and outside this PR's scope. Both that and the larger question — whether an unknown offset should be classified at all rather than assigned a fabricated distance — are tracked in #1826.

`Representation-Change: none` is unaffected: no normalized description moves.

## The through-line: two cdot layers, one set of field names

Raw cdot JSON is genomic 0-based half-open / transcript 1-based inclusive. Ferro's in-memory `CdotTranscript` is genomic 1-based `[incl, excl)` / transcript **0-based** half-open, because `RawCdotTranscript::from_genome_build` (`src/data/cdot.rs:551`) converts both axes on load (#742) — note `CdotMapper` is a real type but has no such method; its constructor is `from_transcripts`. The two layers use the same field names — `genome_start`, `tx_start`, `cds_start` — so a sentence naming a field names a basis only if it also names a layer, and almost none of them did.

That single ambiguity is what every documentation defect in this PR turned out to be. Each table was written by someone reading a real, correct piece of code; the tables disagree because they were each describing a different layer while claiming to describe "cdot". Two independent audits of that ambiguity each resolved it the wrong way, which is the strongest available argument that the fix is to label the layer rather than to correct the basis. Every table below now names its layer in its first line and cross-links the other.

The code was correct at all six sites. Only comments changed.

## The six sites

**m3 — `coords/mod.rs` (raw layer).** Every `cdot_*` helper here describes raw on-disk cdot, and nothing said so; a caller who reached for `cdot_tx_coords` holding an in-memory `CdotTranscript` would have been silently one off on both axes. The module now opens with which layer it is about and points at `data::cdot` for the other. It also cited "bug 944a4e9", which is not an object in this repository (`git cat-file -t` → `fatal: Not a valid object name`); citation dropped.

**m1 — `reference/multi_fasta.rs` (internal layer).** The basis table claimed cdot transcript coordinates are "used directly (already 1-based)" and that `genome_start` "is 0-based, converted to 1-based". Both are exactly backwards relative to the code beneath them, which converts `tx_start` via `+ 1` and uses `genome_start` as-is — and the code is right. The table was describing raw cdot while the loader beneath it reads internal cdot; the conversion comment three lines into that loader already said `#742` and `already 1-based incl`, so the file contradicted itself within one screen. Rows now read `cdot genomic (internal)` / `cdot tx (internal)` / `cdot CDS (internal)` and state the conversion each one actually performs.

**m2 — `convert/mapper.rs`.** The table said `| Genomic | 0-based | Half-open intervals for exons |`. Both halves are wrong: `genomic_to_tx` tests `g_start <= pos <= g_end`, a *closed* interval, against `Exon::genomic_start`/`genomic_end`, which `reference/transcript.rs` documents as 1-based inclusive — and `GenomicLocation.position` in this same file is already documented as an HGVS `g.` number. This is the layer confusion at one remove: the mapper never sees cdot at all, only a materialised `Transcript` whose loader has already converted, so it inherited raw cdot's basis in its docs without ever handling raw cdot's values. The row is corrected and a note says why this module is the 1-based end of that seam.

**n2 — `hgvs/location.rs`.** `CdsPos.base` and `TxPos.base` — the two fields the `mapper.rs` table points at as its authority for "1-based" — stated no basis at all, only "can be negative". The basis was therefore documented in exactly one place, a table that was wrong about a different row. Both now state 1-based, that the negative range counts back with no zero in between, what the `utr3`/`downstream` flags change, and (for `CdsPos`) that `0` is not a position but the `CDS_BASE_UNKNOWN` sentinel.

**m4 — `reference/validate.rs` claimed exon-alignment gaps are rejected upstream.** They are not: `from_genome_build` silently drops exon rows with fewer than five fields, and `MultiFastaProvider` diverts a gapped table only when a supplemental CDS record exists. 58 of 474,818 multi-exon cdot builds carry a gap. The check itself is sound either way because it compares against the maximum exon end rather than a summed length, so only the justification changes.

**m5 — a correction to the audit, not a fix.** `reference/derived_tx_structure.rs:325` was reported as silently dropping `cds_start_incomplete`. It is not a construction path — it is a throwaway readback probe reaching only `reconstruct_spliced_mrna`, which reads `exons`/`contig`/`strand`, so the flag is inert there. `:346` already propagates it (landed with #972/#994) and has a test. A comment now records this so it is not re-filed. The two genuine sites are in `data/cdot.rs` and are left to a change that owns that file.

## Six splice ladders in two shapes → one per shape, with each shape's question named

The two shapes are deliberately **not** merged: shape A is ferro's fine-grained distance bin (donor 2/6/20/50, acceptor 2/12/20/50), shape B is the SO/VEP consequence call (2/8), and they disagree by construction — offset 10 is a donor region under A and a plain intron variant under B. Neither is wrong. What was missing is that nothing named the question each answers and nothing cross-tested them.

Each shape now has one implementation whose doc states its question and points at the other as deliberately different; every former copy delegates. Shape A had three full copies plus one partial (`IntronicRegion::from_offset`, missing the 6- and 12-rungs), which now derives its coarser bins by *mapping* the shared ones, so its missing rungs are dropped by the mapping rather than by a second editable ladder. `from_distance_on_side` exists for a real edge: `IntronPosition` carries its side as a field rather than as the offset's sign, so a 5'-boundary position at offset 0 must still classify as a donor — reading the side off the sign alone would silently change that one case, and there is a test for it.

`tests/it/splice_ladder_shapes.rs` pins both the within-shape agreement and the between-shape relationship at offsets 2, 6, 8, 10, 12, 20, 50, including where they must agree. Sabotage-verified: moving the shared acceptor rung 12 → 11 reddens three of the ten tests in that file.

Also corrects the acceptor 12-rung's comment, which cited a branch-point region (−18 to −35) that its own threshold excludes; the 12-rung is the polypyrimidine tract, as the bin's enum doc and its `splice_polypyrimidine_tract_variant` SO term already said.

## n3 — a hermetic pin on the cdot basis

`convert::mapper`'s `test_cdot_tx_coordinates_are_0_based` — the test both audits cite as *the* pin on the internal transcript basis — returns early when a multi-gigabyte download is absent, i.e. on every default checkout, so it asserts nothing and reports green. `tests/it/cdot_coordinate_basis_layers.rs` adds the tier that cannot skip: a synthetic cdot payload written inline, asserting the conversion on both axes, that CDS bounds ride through, and that the raw-layer helper gives a *different* answer from the loaded values — so pointing it at an in-memory `CdotTranscript` is a real error rather than a harmless no-op. The exon-row count is asserted before the basis checks so a dropped row cannot make them vacuous. Repairing the skipping sibling in place is left to a change that owns `src/convert/mapper.rs`.

## Overlap with in-flight branches

m1, m2 and n2 were deferred on an earlier revision of this PR because other open branches touch those files; they are included now, scoped so they cannot collide. #1762 also edits `src/hgvs/location.rs` and `src/convert/mapper.rs`, but only the `Display` impls, a new `unknown_offset_marker` helper, and `CoordinateMapper` method bodies — the n2 edit touches only the `base` field doc comments inside the `CdsPos`/`TxPos` struct definitions, and the m2 edit only the module header table, so neither lands in a region #1762 restructures. `multi_fasta.rs` is not touched by #1762.

Still out of scope: m5's two genuine `cds_start_incomplete` sites in `data/cdot.rs`, left to a change that owns that file.

## Gate

`cargo fmt --all --check`, `cargo clippy --features dev --lib -- -D warnings` and `cargo clippy --features python --lib -- -D warnings` all clean; `cargo ta` green.

Representation-Change: none. The check genuinely applies to this revision — the n2 edit puts `src/hgvs/location.rs` in the diff — and the answer is still none: that edit is two field doc comments inside the `CdsPos`/`TxPos` struct definitions, adding no code and changing no `Display` arm, so no description renders differently. `src/reference/` **is** touched (four files) and **is** a watched directory as of #1876, and the answer there is none too: those edits are comments and doc tables only. Nothing under `src/normalize/`, `src/spdi/`, `src/project/` or `src/error_handling/` is touched at all.

### DIFF
--- a/src/convert/mapper.rs
+++ b/src/convert/mapper.rs
@@ -6,10 +6,18 @@
 //!
 //! | System | Basis | Notes |
 //! |--------|-------|-------|
-//! | Genomic | 0-based | Half-open intervals for exons |
+//! | Genomic | 1-based | Closed `[start, end]` exon intervals ([`crate::reference::transcript::Exon`]); `GenomicLocation.position` is an HGVS `g.` number |
 //! | Transcript (tx) | 1-based | `TxPos.base` is 1-based |
 //! | CDS (c.) | 1-based | `CdsPos.base` is 1-based, negative for 5'UTR |
 //! | Protein (p.) | 1-based | `ProtPos.number` is 1-based |
+//!
+//! The genomic row is 1-based **inclusive on both ends** because this mapper only
+//! ever reads a materialised [`crate::reference::transcript::Transcript`], never raw
+//! cdot JSON: `genomic_to_tx` tests `g_start <= pos <= g_end` against
+//! `Exon::genomic_start`/`genomic_end`, which are 1-based inclusive by construction
+//! (whichever loader built them has already converted). Raw cdot's 0-based half-open
+//! genomic bounds are a different layer entirely and never reach this module — see
+//! [`crate::coords`] for that layer and [`crate::data::cdot`] for the in-memory one.
 //!
 //! ## Key conversions:
 //! - CDS → Tx: `cds_to_tx()` - accounts for CDS start/end and UTR regions
--- a/src/convert/noncoding.rs
+++ b/src/convert/noncoding.rs
@@ -117,9 +117,14 @@ pub enum IntronicConsequence {
 }
 
 impl IntronicConsequence {
-    /// Create from an IntronPosition
-    pub fn from_intron_position(pos: &IntronPosition) -> Self {
-        match pos.splice_site_type() {
+    /// Create from a [`SpliceSiteType`] bin.
+    ///
+    /// The distance thresholds are **not** restated here — they live in one
+    /// place, [`SpliceSiteType::from_distance_on_side`]. This is only the
+    /// projection of that bin onto this enum, which is coarser: the two
+    /// `*Region` bins collapse into [`Self::SpliceRegionVariant`].
+    pub fn from_splice_site_type(site: SpliceSiteType) -> Self {
+        match site {
             SpliceSiteType::DonorCanonical => Self::SpliceDonorVariant,
             SpliceSiteType::AcceptorCanonical => Self::SpliceAcceptorVariant,
             SpliceSiteType::DonorExtended => Self::SpliceDonorRegionVariant,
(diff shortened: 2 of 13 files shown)


nh13:
@coderabbitai pause

PR #1750, opened 2026-08-12
ci: host bulk test fixtures as release assets instead of Git LFS

nh13:
Representation-Change: none. CI and test-infrastructure only; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched, and no normalized output moves.

## The problem

The repository's Git LFS budget is **exhausted right now**, so every `lfs: true` job dies at checkout:

```
batch response: This repository exceeded its LFS budget. The account responsible
for the budget should increase it to restore access.
```

Four corpora totalling **156 MB** were LFS-tracked and fetched by **ten jobs on every CI run** — `test` ×6, `test-oracle` ×3, and `coverage.yml`'s `coverage`. That is ~1.56 GB of LFS egress per run; at the 192 CI+Coverage runs observed in one day, ~146 GB, about three 50 GB data packs in a day.

LFS bandwidth is metered **even for a public repository**. Release-asset downloads are not. That asymmetry is the whole change.

## What this does

The four files move to the [`test-fixtures-v1`](https://github.com/fulcrumgenomics/ferro-hgvs/releases/tag/test-fixtures-v1) release and are fetched on demand.

- **`.gitattributes`** loses both LFS globs and keeps a note recording why, so the filter is not re-added by reflex.
- **The four paths are untracked and gitignored.** They are deliberately *not* committed as plain blobs — 156 MB of binaries in the tree trades one problem for a worse one.
- **`tests/fixtures/CHECKSUMS.sha256`** pins their SHA-256 digests, so a swapped, truncated or partially-downloaded asset is caught rather than parsed. (Cross-check: those digests are byte-identical to the files' LFS object ids, since LFS keys objects by SHA-256.)
- **`scripts/fetch-test-fixtures.sh`** fetches and verifies them over plain HTTPS with no token, so it works unchanged on a forked-PR runner. `--verify` checks without fetching; `--force` re-downloads.
- **A new `fixtures` job** fetches them once per run and publishes them to the shards as a workflow artifact with one day of retention. It slots into the existing `needs: [test-build, spec-fixtures]` chain, so it adds no serialization point — it runs alongside the two build jobs and finishes long before either.
- **Not `actions/cache`.** That store sits at ~9.9 GiB of a 10 GiB cap across ~137 entries, so a 156 MB entry would evict Rust build caches by LRU. That is #1218, where eviction took the nightly's 3.25 GiB entry as collateral.

Egress per run:

| | before | after |
|---|---|---|
| LFS | ~1.56 GB | **0** |
| release assets | 0 | 156 MB (one fetch, shared to 9 jobs) + 156 MB for `coverage.yml` |

`coverage` is a separate workflow so it cannot consume `ci.yml`'s artifact and fetches its own copy; unmetered, so it costs nothing LFS used to charge for.

## A failed fetch must fail loudly

This is the part worth reviewing. The suites that read these corpora **skip green** when the files are absent:

```rust
let buf = match load_fixture_bytes("clinvar_hgvs_500k.json.gz") {
    Some(b) => b,
    None => { eprintln!("Skipping: ... not found"); return; }
};
```

So a fetch that failed quietly would delete the ClinVar, CMRG and Paraphase coverage **while making CI look faster** — a speedup that is really coverage loss. Two independent guards:

1. **The fetch script** exits non-zero on a bad download, a digest mismatch, a missing asset, or a malformed manifest, and never installs an unverified file (it downloads to a temporary path and only renames after the digest matches).
2. **`FERRO_REQUIRE_BULK_FIXTURES=1`** turns the eight fixture-absence skips into failures naming the missing path and the fix. Set in `test`, `test-oracle`, `coverage` and `external-validation` — every job that fetches the corpora. Unset locally, where skipping remains the useful default. See `tests/it/common/bulk_fixtures.rs`.

Measured, by holding `cmrg_genes_exhaustive.json.gz` aside:

| | fixture absent, guard unset | fixture absent, guard set |
|---|---|---|
| `cmrg_exhaustive_tests::test_cmrg_exhaustive_benchmark` | **PASS in 0.016 s** | **FAIL**, naming the path |
| `normalize_axis_preserving::axis_is_preserved_over_the_cmrg_corpus` | **PASS in 0.016 s** | **FAIL**, naming the path |

That first column is the hazard, stated as a number: the same test takes 11.692 s when it really runs.

## The tag-collision check, and what it found

Asked of `release-plz.toml` and every workflow, because there is an active v0.14.0 release PR and a confused release process would be worse than the problem.

**`release-plz.toml` is clear** — it configures `features_always_increment_minor`, a `pr_body` template and changelog grouping, and contains no `git_tag_name`, tag pattern or tag filter. Nothing in it can match a non-semver tag. **No workflow triggers on `push: tags:` at all.**

**The hazard was elsewhere.** `release-wheels.yml` runs on `release: published`, which admits **no tag filter** — GitHub offers none for that event — so *any* release fires the full wheel matrix, uploads wheels onto it, and attempts a PyPI publish with whatever version `Cargo.toml` holds. Publishing the data release triggered exactly that. The run was cancelled by hand within seconds, long before any build job could reach `upload` or `publish-pypi` (both sit behind ~15 minutes of `needs`), so nothing was uploaded and PyPI was never contacted — confirmed `conclusion: cancelled`, and the release carries only its four fixture assets.

This PR adds the guard so it cannot recur: the four building jobs now require `startsWith(github.event.release.tag_name, 'v')`, and everything downstream reaches them through `needs:` and skips with them. `workflow_dispatch` is deliberately left ungated, since its dry-run form takes a blank tag and its upload/publish jobs already gate on the event.

`tests/fixtures/README.md` records that a regenerated corpus needs a **new** release rather than a replaced asset (replacing in place would break every checkout pinned to the committed digests), and that the tag must not be `v*`.

## What "no LFS remains" does and does not mean

Stated precisely, because the strong reading is false:

- ✅ **At HEAD there is no LFS.** `git lfs ls-files` on this branch returns **empty**.
- ✅ **A fresh clone of this branch fetches zero LFS objects.** Verified: `.git/lfs` is not even created, and an explicit `git lfs pull` finds nothing.
- ✅ **CI fetches no LFS objects.** No `lfs:` key remains in any workflow (checked by parsing the YAML, not by grep — several jobs carry comments containing the literal string).
- ❌ **History is not rewritten.** Every pre-change commit still references the LFS pointers, and GitHub still stores the objects: ~156 MB of LFS *storage*, which is not the exhausted resource — bandwidth is. Only someone checking out a pre-change commit would touch LFS at all.

## Verification

Local gate, all green on this branch:

- `cargo fmt --all`, `cargo clippy --all-features --all-targets -- -D warnings` — clean
- `cargo nextest run --features dev --lib --test it --test-threads=4` — **10234 passed, 0 failed**
- `actionlint .github/workflows/*.yml` — clean; all 14 workflows parse
- `shellcheck scripts/fetch-test-fixtures.sh` — clean

The three fixture-reading suites **ran rather than skipped**, which a pass count alone cannot show:

| test | runtime |
|---|---|
| `clinvar_hgvs_tests::test_clinvar_hgvs_500k_benchmark` | 1.370 s |
| `clinvar_hgvs_tests::test_clinvar_hgvs_unique_benchmark` | 10.531 s |
| `cmrg_exhaustive_tests::test_cmrg_exhaustive_benchmark` | 11.692 s |
| `paraphase_exhaustive_tests::test_paraphase_exhaustive_benchmark` | 3.909 s |
| `normalize_axis_preserving::axis_is_preserved_over_the_cmrg_corpus` | 20.453 s |
| `normalize_axis_preserving::axis_is_preserved_over_the_clinvar_unique_corpus` | 14.779 s |

A skip returns in ~0.016 s.

The fetch script's failure modes were each exercised against the live release: missing files, a corrupted file, a 404 tag, a manifest digest no asset matches, and a malformed manifest line — all exit non-zero, and the two download cases install nothing. The contributor path was run end-to-end on a fresh clone: 156 MB fetched and verified in 6.4 s, tree still clean afterwards.

## Expect CI to be red, and not because of this

The LFS budget is exhausted, so any job still fetching LFS — on other PRs, and on this PR's merge ref while `main` still carries `lfs: true` — fails at checkout. This change is what removes those fetches. The local gate above is the authority until the budget is restored or this lands.


nh13:
@coderabbitai pause

Issue #797, opened 2026-06-21
fix(project): genomic projection of c.* variants in the 3'UTR poly-A region (NM_003002.2; #790 residual)


nh13:
## Problem

A `c.*` variant whose 3'UTR position falls in a transcript's **poly-A tail** cannot be projected to the genome — the poly-A tail is post-transcriptional and has no genomic coordinate, so the exon→genome map (correctly) does not cover it.

Surfaced by #790 (PR #795): the NM_003002.2 derived exon structure covers the genomic core (poly-A excluded), so:

```
NG_012337.1(NM_003002.2):c.*824del      expected="NG_012337.1:g.13948del"  err="project: Invalid coordinates: Cannot map tx position 1364 to genome"
NG_012337.1(NM_003002.2):c.*824+10del   expected="NG_012337.1:g.13958del"  err="project: Invalid coordinates: Cannot map tx position 1364 to genome"
```

Both remain in `baseline-failures/genomic.txt`.

## What was tried (and why it's not trivial)

Extending the last exon into contiguous downstream genome (so the poly-A region gets coordinates) was attempted during #790 and **reverted** — it produced a *wrong* genome coordinate (`g.13965del` for both, vs mutalyzer's distinct `g.13948`/`g.13958`), because the sibling-anchored 3'UTR differs from the target's. mutalyzer maps these positions by a base-agnostic coordinate walk into contiguous genome using the *target's own* 3'UTR structure, which a sibling anchor doesn't reproduce.

## Direction (decide in design)

- Map a `c.*` poly-A-region position to contiguous downstream genome using the **target transcript's own** 3'UTR genomic extent (not a sibling's), when derivable; otherwise decline cleanly rather than emit a wrong coordinate. 
- This is narrow (the only corpus rows today are the two SDHD `c.*824` rows) but is the residual #790 explicitly carved out.

Refs #790, #795, #487/#672.

Issue #506, opened 2026-05-31
feat(project): extend direct c.→p. path to bare n./r. (Tx/Rna) inputs


nh13:
The #498 direct c.→p. path handles bare-`NM_` **coding (`c.`)** inputs (`HgvsVariant::Cds` with no `genomic_context`). Bare non-coding (`n.`) / RNA (`r.`) transcript inputs still route through the genome pivot and fail with `unsupported projection: input variant has no parent reference`.

Example: `NM_003002.4:n.206_210del` → expected `NM_003002.4(NP_002993.1):p.(Gly58GlnfsTer9)`; ferro errors (no parent).

Extend the direct path to `HgvsVariant::Tx` (and `Rna`) bare inputs: an `n.` position is a 1-based transcript position, convertible to a CDS position via the transcript's `cds_start` (`cds_to_transcript_pos` / `CdotTranscript::cds_to_tx`) without the genome, so the same `predict_protein_consequence` helper applies once the n.→CDS offset is resolved. Guard UTR/intronic cases as in the coding path.

Refs #498, #326.

Issue #493, opened 2026-05-29
Enforce HGVS bracket/allele cardinality uniformly across all coordinate systems (strict mode)


nh13:
## Summary

Single-variant bracket forms (`c.[76A>C]`, `g.[1000G>A]`, `p.[=]`, `r.[…]`), standalone pure-marker brackets (`c.[=]`, `[?]`, `[0]`), and empty brackets (`c.…[];[]`) are accepted by the parser today and silently unwrapped/preserved. The HGVS v21 spec marks these as **non-conformant**. We should enforce the bracket/allele cardinality rule **uniformly across every coordinate system** (`c/g/n/m/o/r/p`), driven by the existing strict/lenient/silent error-handling machinery — not as a per-axis patch.

This was surfaced by a CodeRabbit comment on #475 that proposed rejecting only `p.[=]`/`p.[(?)]`. A protein-only fix is the wrong scope: `p.[=]` travels the same compact-prefix singleton path as `c.[76A>C]` and `g.[1000G>A]`, and the parser accepts singletons identically on all axes (`test_singleton_cis_allele_preserves_wrapper`, plus the "Singletons are intentionally accepted" comments in each dispatch arm of `src/hgvs/parser/variant.rs`). Fixing protein alone would make it the lone divergent axis.

## What the spec mandates

From `assets/hgvs-nomenclature/docs/recommendations/DNA/alleles.md` (mirrored in `RNA/alleles.md`, `protein/alleles.md`) and `general.md:79`:

- `[ ]` is **allele** syntax.
- **cis** = one bracket pair with ≥2 members: `c.[2376G>C;3103del]`.
- **trans** = ≥2 bracket groups: `c.[2376G>C];[3103del]`.
- A single variant in brackets standalone — `c.[76A>C]` — is explicitly marked `invalid` (alleles.md note: *"I find the notation `c.[76A>C]` … misleading"*; recommended form is `c.[76A>C];[76=]`).
- Empty brackets `c.…[];[]` are explicitly *"not allowed"* (alleles.md, homozygous note).
- Pure markers `=`, `?`, `0` are valid **only as a whole group inside a multi-group trans construct** (`c.[2376G>C];[=]`, `;[?]`, `;[0]`) — never standalone `c.[=]`.

So the conformant shapes, identical for all seven axes, are:

1. one bracket group with **≥2 cis members**, or
[...]

PR #1751, opened 2026-08-12
fix(hgvs): refuse `n.*N` at parse in every mode; `n.-N` stays strict-only

### LINKED ISSUE #1748: Parser accepts `n.*N` and `n.-N`, zones `background/numbering.md:50-54` does not put on the non-coding axis
`background/numbering.md:50-54` enumerates the whole `n.` axis and there is no `*` zone and no negative zone on it:

```
50: ### non-coding DNA reference sequences
51:
52: - nucleotide numbering is `n.1`, `n.2`, `n.3`, ..., etc., from the first to the last nucleotide of the reference sequence.
53: - nucleotides in introns are numbered as for coding DNA reference sequences (see above), although proceeded by `n.` (not `c.`).
54: - it is **not** allowed to describe variants in nucleotides beyond the boundaries of a transcript reference sequence, using that transcript reference sequence.
```

`:52` is an exhaustive enumeration — `n.1` through the last nucleotide. `:53` grants introns *explicitly*, which is what makes `:52`'s silence about `*`/`-` meaningful rather than merely terse: the spec knew how to add a zone to this axis and added exactly one. `:54` then forbids naming anything past the boundary outright. `numbering.md:45` records that the sibling proposal — extending the numbering to mark non-transcribed nucleotides — "[has] been made but [was] rejected".

The `*` and `-` zones genuinely exist on the **coding** axis (`c.`), where they are anchored to the CDS rather than to the sequence: `c.-1` is the base before the ATG and `c.*1` the base after the stop, both still *inside* the transcript. `n.` has no CDS to anchor them to, so `n.*1` and `n.-1` can only mean "outside the reference sequence", which is exactly what `:54` prohibits.

## Ferro accepts both, in every mode

```
$ ferro parse 'NR_037639.1:n.*5A>G'   # accepted, re-emitted verbatim
$ ferro parse 'NR_037639.1:n.-5A>G'   # accepted, re-emitted verbatim
```

`TxPos` carries a `downstream: bool` (the `*`) and admits `base < 0`, and neither is gated anywhere.

## Ferro's own tests assert the prohibited form is correct

`tests/it/issue_255_noncoding_markers.rs` opens:

```
//! Per HGVS v21, the `n.` coordinate system uses:
//! - `-N` upstream of transcript start
//! ...
//! - `*N` downstream of transcript end
```

That premise is not in v21. It is the `c.` axis's zone list transplanted onto `n.`, and #255 pinned 15 round-trips on it. Those rows are the ones a fix has to account for.

## Scope

The **coding** axis is untouched: `c.*N` and `c.-N` are legal HGVS and ferro deliberately entertains a past-the-end `c.*` in the poly-A carve-out (#797, `src/project/projector.rs`).

The **`r.` axis is deliberately out of scope**, and not by omission. `numbering.md:58` — "nucleotide numbering for an RNA reference sequence follows that of the associated **coding or non-coding** DNA reference sequence; nucleotide `r.123` relates to `c.123` **or** `n.123`" — makes `r.`'s zone set a property of the underlying reference, not of the prefix. `r.*5` is legal on a coding transcript (`:61`) and prohibited on a non-coding one (`:60`), and the parser holds no provider, so it cannot tell which it has. By contrast `general.md:26` makes `n.` a claim about the reference *type* — "`n` for a non-coding DNA reference sequence" — so `numbering.md:52` binds on the prefix alone, with no reference lookup. The accession is not a substitute signal either: ferro's own corpus carries `NM_003002.4:n.*5del`, an `n.` description on a coding accession.

## Where it should be enforced

`rulings[absolute-prohibition-enforcement-stage]` (decided 2026-08-10) settles the stage and it is **mode-dependent**: strict fails at PARSE, lenient does not validate input conformance, silent is lenient without messages. So this belongs on the `parse_hgvs_with_config` seam keyed on `ErrorConfig::action_for`, alongside `apply_bracket_cardinality_rule` (#493) and `apply_alignment_only_symbol_rule` (#1627) — **not** in the grammar, which would refuse in every mode and contradict that ruling.

This is a sibling of #1630 rather than a duplicate: #1630 is the campaign to move `checklist.md`'s existing prohibitions onto that seam, and this clause is not on the checklist at all — it is unenforced anywhere, at any stage, in any mode.

nh13:
`background/numbering.md:50`–`:54` enumerates the non-coding DNA axis in full. `:52` numbers it "`n.1`, `n.2`, `n.3`, ..., etc., from the first to the last nucleotide of the reference sequence", `:53` grants intronic offsets explicitly, and `:54` states that "it is **not** allowed to describe variants in nucleotides beyond the boundaries of a transcript reference sequence, using that transcript reference sequence". `:53` is what makes `:52`'s silence about `*` and `-` an exclusion rather than terseness: the spec knew how to add a zone to this axis and added exactly one. `numbering.md:45` records that the proposal to mark non-transcribed nucleotides was rejected.

So both `n.*N` and `n.-N` name zones the axis does not have. They are refused at **different stages**, and the split is measured rather than argued.

## The measurement

Over the four committed corpora (`clinvar_hgvs_500k`, `clinvar_hgvs_unique`, `cmrg_genes_exhaustive`, `paraphase_genes_exhaustive`), **103,762** `n.`-axis rows:

| marker | rows | stage | code |
|---|---|---|---|
| `n.*N` | **0** | refused at parse in **every mode**, bare entry included | `E1003 InvalidPosition` |
| `n.-N` | **5** | refused at **strict** parse only | `W4008` |

The `n.*N` zero is real, not structural — the same scan over the same rows finds `n.-N` five times, so the counter can see the class.

Those five are `NR_003051.3:n.-57T>C`, `NR_003051.3:n.-30_-7dup` and `LRG_163t1:n.-5delins17` (all RMRP, whose upstream promoter variants are the clinically conventional case for the spelling), `NR_029595.1:n.-4771G>T` (MIR208A) and `NR_033294.1:n.-6G>A` (SNORD118). **NCBI publishes all five today**, so lenient, silent and the mode-less `parse_hgvs` entry keep accepting them — that entry being the one every `ferro` subcommand and all four corpus consumers call. `the_five_real_clinvar_rows_still_parse_on_the_bare_entry` pins them by name.

So `n.*N` is refused because the axis has no such zone **and nothing in the wild uses it**; `n.-N` is deviation-tolerated in the permissive modes precisely **because real clinical data uses it**, on the identical clause reading.

## The unconditional half is a maintainer decision, and it departs from a decided ruling

Refusing `n.*N` in every mode does not fit `rulings[absolute-prohibition-enforcement-stage]` (decided 2026-08-10), which makes enforcement mode-dependent uniformly. That is stated plainly rather than reconciled away, because both available reconciliations are weaker than they look:

- *"`n.*5` names no position at all, so it is a grammar matter outside that record."* Arguable, **not conclusive**: `checklist.md:16` says a genomic reference "can not have nucleotides with additions like a `+`, `-`, or `*`", so `g.*10` is exactly parallel — and that record's own census places `g.*10` under mode-gating and says lenient should newly *accept* it. A scoping rule that carves out `n.*N` carves out `g.*10` too.
- *"The record's question names only `checklist.md`."* True of the question sentence; its rationale generalises to "every absolute prohibition", and `:54` is one.

What holds is narrower and empirical. That record rejected unconditional refusal for one stated reason — it "would newly refuse inputs ferro accepts today, with no escape for a caller round-tripping a real-world corpus" — and for this shape there is no such caller. So the unconditional arm is a **maintainer's decision** under rule 6 of the `README.md` ruleset, disclosed under rule 7, and **revisitable on user demand**: a reported real-world `n.*N` corpus is grounds to move it onto the `n.-N` schedule, not to defend the refusal.

An amendment to `absolute-prohibition-enforcement-stage` recording the carve-out is **signed off and applied**, as the second commit on this branch. It appends to that record's `rationale` only — no clause is added, no status changes, and the ledger census is unmoved at 23 / 20 / 3, verified per-site across all five derived pins. `the_unconditional_arm_is_a_disclosed_departure` pins the shape of the split in code, so neither arm can be quietly moved onto the other's schedule.

**This PR refuses descriptions ClinVar publishes today only in the `n.-N` case, and only in strict mode.** No `n.*N` description exists in any corpus ferro ships.

## `E1003`, not a second `W4008`

One input must not carry two diagnoses. `n.0` — the other position this axis can spell and does not have — is already `E1003 InvalidPosition` in every mode, so `n.*N` joins it there; an unconditional refusal is an error, not a mode-dependent warning. `W4008` is now solely the `n.-N` diagnosis, and its registry entry, `ErrorType` docs, Python enum comment and audit rows say so.

## A message defect fixed on the way past

`numbering.md:54` prohibits a nucleotide beyond a transcript's boundaries **"using that transcript reference sequence"** — a conditional on its own subject. For the selector form `NG_007485.1(NR_003529.3):n.-40000del` the reference sequence is the genomic `NG_007485.1`, which *does* contain the nucleotide; the transcript is only the numbering selector. The old message asserted the `:54` half verbatim for that shape. Both messages now come from one builder on `NonCodingZoneMarker` and quote `:54` only where its condition holds; the refusal stands on `:52` alone otherwise. Pinned by `the_boundary_clause_is_not_asserted_for_the_selector_form`.

## Where the checks live

`n.*N` is refused by `validate_noncoding_downstream_zone`, a per-leaf rule in `parse_variant` — the single generic entry both `parse_hgvs` and `parse_hgvs_with_config` funnel through, and the fast path never reaches an `n.` description. `n.-N` stays on the mode-gated `parse_hgvs_with_config` seam alongside `apply_bracket_cardinality_rule` (#493) and `apply_alignment_only_symbol_rule` (#1627). Both are AST-keyed on `TxPos.downstream` / a negative `TxPos.base`, never a scan of the rendered text — an accession contains `-`, and so does `:53`'s intronic offset. `n.6-3` is an intronic distance, not a zone, and is untouched.

The coding axis is untouched, including the deliberate past-the-end `c.*` in the #797 poly-A carve-out. The `r.` axis is untouched deliberately: `numbering.md:58` makes an `r.` description's zone set a property of the underlying coding *or* non-coding reference, which a parser holding no provider cannot resolve.

## Fallout, in full

`n.*N` was used as a convenient downstream-position fixture in the sites below, all of which are about something else. `TxPos::downstream` is public API, so every one of those AST shapes is still reachable by construction and every consumer still has to handle it — the fixtures are built rather than parsed, which keeps their subject:

- **Constructed**: `src/spdi/convert.rs` (2 tests), `src/hgvs/variant.rs` self-cancelling overlap, `tests/it/issue_336_position_past_end.rs` (2 tests), `tests/it/mutalyzer_normalize_tests.rs`'s #506 projector guard.
- **Inverted to assert the refusal**: `src/hgvs/parser/variant.rs::test_parse_tx_downstream_positions` (plus a `c.*5` control), `tests/it/issue_253_boundary_spanning.rs::n_body_to_downstream_del_is_refused`.
- **rstest corpora**: `tests/it/parser_tests.rs` and `tests/it/normalize_tests.rs` — the `NR_033294.1:n.*5C>G` case is replaced by that accession's **real** ClinVar row, `NR_033294.1:n.-6G>A`.
- **Assertion removed with a comment** (the shape can no longer be authored): `tests/it/issue_1482_boundary_crossing_sibling.rs`, `tests/it/issue_1508_overlap_across_region_boundary.rs`. Their property is carried by the surviving `n.-N` rows.
- **`tests/it/issue_255_noncoding_markers.rs`** — §2 and the two downstream crossings in §5 re-pointed to `c.` on `NM_000492.4`, where `*N` is genuinely legal; §1 and the two upstream crossings deliberately **stay** on `n.`, because they are now the shape guard behind the five ClinVar rows. The module doc says which half moved and why the other did not.

One consequence worth naming: `Region::TxDownstream` is now unreachable from any parsed description. It survives for directly-constructed ASTs, which is why the fixtures above are constructed rather than deleted.

## Public-API note

`TxPos::downstream` remains constructible, so a Rust library caller can build an `n.*N` AST whose rendering `parse_hgvs` now refuses. Normalization never mints that flag — the two sites that can emit it (`canonicalize_tx_variant`'s downstream short-circuit and `build_tx_merged`'s `Region::TxDownstream` arm) strictly preserve an input flag, and no string entry point (parser, CLI, Python, VCF, SPDI, projector) reaches the shape — so no corpus row and no stored string changes. `FERRO_ASSERT_REPARSE` gains a third, AST-keyed allow-list entry recording that ferro no longer claims this AST is renderable; it is keyed on `noncoding_zone_marker` and so cannot exempt `c.*N`, which stays legal and parses.

That entry was added because the oracle sweep caught two real failures introduced by this PR — `issue_336_position_past_end::lenient_mode_skips_n_downstream_position` and `::lenient_mode_preserves_downstream_n_indel_without_normalization`, both test-side, both from converting a parsed fixture into a constructed one. Measured like for like under all four oracles: 18 failures on `origin/main`, 20 on this branch pre-fix, 0 new after the fix. The 18 are the documented pre-existing residue that `run_oracle_suite.sh` excludes by design.

Also fixed: `rulings[c-and-n-positions-are-flat-transcript-offsets]` was cited three times and **exists in no ledger on this base** — `ruling_citation_currency` rejects a citation that does not resolve. That record is added by #1735, which is not a dependency of this PR, so the citations are removed and the argument is made from the clause directly.

No `*_failure_expectations.json` changes; no `n.*N` appears in any committed fixture, example, bench or bin.

Closes #1748

Representation-Change: none. 0 of 103,762 committed `n.`-axis corpus rows move. `n.*N` is newly refused at parse in every mode and occurs 0 times in that corpus; the `n.-N` half keeps its strict-only mode gate precisely so the five real ClinVar rows on it (`NR_003051.3:n.-57T>C`, `NR_003051.3:n.-30_-7dup`, `NR_029595.1:n.-4771G>T`, `LRG_163t1:n.-5delins17`, `NR_033294.1:n.-6G>A`) keep parsing unchanged on the bare entry and in lenient/silent. No stored string changes.


nh13:
@coderabbitai pause

PR #1752, opened 2026-08-12
fix(normalize): define cis-member overlap once, on write footprints

### LINKED ISSUE #1749: Overlap is defined four times, on three different geometries, and they disagree
## Summary

"Do two cis members overlap?" is decided in **four** places, by **three different geometric notions**, and they disagree with each other. The correct notion — key on what a member **writes**, not on what it **reads** — is already established in the code (#1411, #1437, #1448) but exists only as prose in a comment plus two partial, non-shared implementations.

## The four deciders

| # | site | notion | scope |
|---|---|---|---|
| 1 | `overlap::detect_overlap_conflicts` — grouped pass | span, minus `writes_only_at_a_junction` (certain-extent `dup` only) | coincident bounds |
| 2 | `overlap::detect_overlap_conflicts` — `span_writer_footprint` pass (#1448) | span, restricted to `sub`/`del`/`delins`/`inv` | partial/nested intersection |
| 3 | `overlap::junction_overlaps` (`detect_insertion_overlaps` / `detect_interior_junction_conflicts`) | junction for `ins`/`dup`/`repeat`, span for the rest | same-junction + interior-junction |
| 4 | `HgvsVariant::detect_self_cancelling_pair` (`src/hgvs/variant.rs:934`, E3006) | **read span**, `del`x`dup` pairs only | refused at **parse**, unconditionally |

Sites 1–3 all live in `src/normalize/overlap.rs`; site 4 is in the parser path and is the one this issue's third symptom is about.

## Symptom 1 — the same geometry gets three different verdicts

Measured today on real hg38 (`ferro normalize --reference <prepared dir> --error-mode strict`, `origin/main` `b49c5ce3`):

```
NC_000017.11:g.[43045721dup;43045721_43045722dup]          -> g.43045721_43045722insAGA   (accepted, merges)
NC_000017.11:g.[43045721_43045724del;43045722_43045723del] -> REFUSED, OverlapConflictingEdits / W5002
NC_000017.11:g.[43045721_43045725del;43045723_43045727dup] -> REFUSED at PARSE (E3006 self-cancelling)
```

Row 1 is right and deliberate: two `dup`s write at two different junctions, so the composition is unique.

Row 3 is decided by site 4 on **read spans**. Under the write-footprint principle the `del` writes `[43045721, 43045725]` and the `dup` writes at the junction `43045727|43045728` — disjoint, and the pair denotes one unambiguous sequence. It is nonetheless refused, *at parse*, where no error mode can opt out. `general.md:58` is cited for it, and `rulings[conflicting-member-geometry-refusal-scope]` has already recorded that `:58`'s stated ground "literally describes only its own `del`+`dup` example".

So `dup` is a junction writer in sites 1 and 3, a non-participant in site 2, and a span writer in site 4.

## Symptom 2 — two disagreements *inside* `overlap.rs`

Both are between site 1's two passes and site 3, on the same file:

- **`repeat`**: `writes_only_at_a_junction` deliberately excludes it (its footprint is reference-dependent, so it keeps its span); `span_writer_footprint` also excludes it (so it is invisible to the intersection test); but `junction_writing_kind` includes it and `continue`s, so site 3 registers it as junction-**only** and never as a span. Three passes, three answers for one edit kind.
- **uncertain-extent `dup`**: `writes_only_at_a_junction` gates on `uncertain_extent: None`; `junction_writing_kind` matches `NaEdit::Duplication { .. }` with no such gate. Site 1 treats it as a span, site 3 as a junction.

Neither is reachable from a comment on the other — they are three independent `matches!` ladders.

## Symptom 3 — a documented false-positive class, and an order dependence

`CLAUDE.md` already records the live disagreement:

> insertion flush against a deletion | 233 | `triples_are_disjoint` calls it an overlap; `cis_apply_oracle`'s tie-break does not, and it is well defined

`spdi::apply::triples_are_disjoint` (`src/spdi/apply.rs:502`) is in fact *already* a write-footprint test — a `dup` converts to SPDI as `seq:101::ATG`, a zero-width triple with an empty deletion, so it claims no base. That makes the brief's framing of it as a read-span opinion wrong. What it actually has is a narrower defect: it walks a list sorted by `sort_by_key(Reverse(position))`, which is **stable**, so two triples at the same interbase position keep their input order — and the running-`reach` test then reads

```rust
if reach.is_some_and(|r| r > t.position) { return false; }
```

For an insertion flush against the **5' edge** of a deletion, both triples share a position, so which one is visited first is decided by member input order:

- `g.[5_9del;4_5insA]` -> insertion visited first -> disjoint
- `g.[4_5insA;5_9del]` -> deletion visited first -> `reach > position` -> **not** disjoint

One variant, two verdicts, decided by nothing the description states. (To be confirmed empirically in the fix; the reading is from the source.)

`triples_are_disjoint` is additionally serving two jobs at once: it is the splice algorithm's **precondition** (overlapping triples index into a shrinking buffer — the #1244 panic) *and* it is consulted as a semantic "does this allele denote one sequence" judgement. Those are not the same predicate and should not be one function.

## Symptom 4 — the #1448 comment's stated reason does not reproduce

`overlapping_duplication_spans_are_not_a_coincident_conflict` (`src/normalize/overlap.rs`) says it asserts at unit level rather than end to end

> because `detect_insertion_overlaps` reports an overlapping `dup` pair for its own separate reasons (#1437)

Measured end to end in **both** strict and lenient, the only warning emitted for `g.[43045721dup;43045721_43045722dup]` is `MEMBERS_COALESCED_FROM_REPORTED_FORM`. No overlap warning is emitted at all.

The comment is **stale**, and it inverts its own citation: #1437 is precisely the change that made `junction_overlaps` `continue` after registering a `dup`/`repeat` as a junction occupant, so that it is *never* pushed into `spans`. After #1437 two `dup`s at different junctions match neither branch (a) (different junctions) nor branch (b) (no spans). The comment describes pre-#1437 behaviour and cites #1437 as the reason for it.

## Proposed fix

Promote the principle from a comment to the single shared definition, and have every consumer call it:

```rust
fn write_footprint(edit) -> Interval   // dup/ins -> zero-width junction; del/sub/delins/inv -> their span
fn members_conflict(a, b) -> bool      // write footprints intersect
```

One definition, one call site per consumer, no second opinion that can drift. `repeat` needs an explicit, stated answer rather than three implicit ones, and site 4's parse-stage refusal needs to be reconciled with it or justified as a deliberate exception.

## Expected cost

Unifying disagreeing definitions necessarily moves behaviour somewhere. The fix must **measure** which input classes change verdict, say which decider was right in each, and disclose the row movement — rather than tune the definition to move zero rows.

## Ruling-record note

`rulings[conflicting-member-geometry-refusal-scope]` (decided) says members "may not claim intersecting **reference territory**, whatever edit types they render as". The code says otherwise for `dup`, with the better argument — and the record's own rationale is un-denotability ("two members writing to one reference interval describe no single chromosome sequence"), which *supports* the code, since two `dup`s sharing a read span denote one unambiguous sequence. The record's wording overreaches its own rationale. A **descriptive** correction to "intersecting write footprints" is proposed, changing no verdict, for operator review.

nh13:
Closes #1749

Representation-Change: 108 of 86,398 corpus rows move — 102 in `repeat_beside_a_sibling` and 6 in `junction_interior_to_span`. Every moved row is a member-conflicting input whose previous output was a re-derivation that denoted different bases; the description is now left exactly as authored, so all 108 become fixed points. **0** were fixed points before, so no stored string migrates. **This supersedes an earlier `6 of 86,020, all respell`**, which was measured on a corpus that could build only GROWING repeats: the `repeat_beside_a_sibling` family emitted `{unit}[3]` over a 6-base tract and never `{unit}[1]`, so `repeat_footprint`'s shrinking branch — the branch this PR adds — was structurally unreachable and its 102 rows were invisible. The family now emits both directions and `the_repeat_sibling_family_varies_the_repeat_direction` fails if it stops.

## The problem

"Do two cis members overlap?" was decided in **four** places, on **three** different geometries, and they disagreed. The correct notion was already established — key on what a member **writes**, not on what it **reads** (#1411, #1437, #1448) — but it lived as prose in a comment plus two partial implementations that shared no code, so nothing could see the disagreement.

Measured on real hg38, `origin/main` at `b49c5ce3`, `ferro normalize --reference <prepared dir> --error-mode strict`:

```
NC_000017.11:g.[43045721dup;43045721_43045722dup]          -> g.43045721_43045722insAGA   (accepted)
NC_000017.11:g.[43045721_43045724del;43045722_43045723del] -> REFUSED, W5002
NC_000017.11:g.[43045721_43045725del;43045723_43045727dup] -> REFUSED at PARSE (E3006)
```

Two of the three disagreements were **internal to `overlap.rs`**, between three independent `matches!` ladders in one file, none reachable from a comment on the others:

- **`repeat`** was a span to the coincident-bounds pass, invisible to the intersection pass, and a junction-**only** occupant to the junction pass.
- **uncertain-extent `dup`** was a span to one pass and a junction to another.

## The fix

The geometry lives once, in `normalize::footprint::WriteFootprint`, generic over the position type — `normalize::overlap` instantiates it on ranked HGVS axis positions, `spdi::apply` on 0-based interbase ones. A member writes over a closed range of bases (which it may or may not leave standing) and/or at a zero-width junction; two members conflict when those collide. Every consumer derives its question from that one predicate.

**`repeat` is settled by arithmetic the description already carries**, not by a guess. `unit x count` against the tract length says whether it grows (extra copies land at the junction 3' of the tract — a `dup`'s geometry) or shrinks (the trailing bases go; the leading ones are untouched reference). The note this replaces held that `A[8]`, `G[6]` and `TA[3]` are "indistinguishable at this layer" so a repeat must conservatively keep its whole span. They are indistinguishable in *content* — which is what a `REFSEQ_MISMATCH` check is for — but perfectly distinguishable in *length*, and length is all the geometry needs.

That is what keeps #1437 intact: the per-member pipeline picks between the `dup` and `repeat` spellings **by axis**, so a detector answering differently for the two would be sensitive to a choice the author never made. A conservative "repeat keeps its span" reading fails that test; this one passes it and still catches a shrinking repeat colliding with a sibling deletion, which no pass caught before.

**`spdi::apply::triples_are_disjoint` was order-dependent**, which is a separate defect the unification surfaced. It carried a running 3'-most `reach`, and `sort_by_key` is **stable**, so a zero-width triple and a span triple sharing one interbase position were read in *input* order:

```
g.[5_9del;4_5insA]  -> disjoint
g.[4_5insA;5_9del]  -> NOT disjoint
```

One variant, two verdicts, decided by nothing the description states. It is now pairwise (order-independent), and the *application* order breaks the tie by deletion length so the splice is correct in either order too — without that, applying the insertion first left the deletion eating the bases it had just inserted. This is `CLAUDE.md`'s 233-fire "insertion flush against a deletion" false-positive class, whose cause was this, not policy.

The two jobs `triples_are_disjoint` was conflating are now separate: `conflicts_with` answers "does this allele denote one sequence", `obstructs_splice_of` answers "can this be spliced in one 3'->5' walk". They differ exactly on #1406's insertion-interior-to-a-deletion, which denotes one sequence and still cannot be executed in one pass.

## What moved, and which decider was right

**108 rows of 86,398 (0.1%)**, in two families. Every one of the 108 ends at the input's own spelling: the member conflict is now reported, so the description is left as authored instead of being re-derived into a form that denoted different bases. **0 of the 108 were fixed points before and all 108 are fixed points after**, so nothing that a consumer had stored moves.

| family | rows | moved |
|---|---:|---:|
| `repeat_beside_a_sibling` | 756 | **102** |
| `junction_interior_to_span` | 9,216 | **6** |
| every other family (20) | 76,426 | 0 |

### The 102: a shrinking repeat colliding with a sibling deletion

The branch that `repeat_footprint` adds — `spanning(removed_from, end, false)` for a repeat that shrinks — is what these rows exercise, and they are the reason the earlier `6 of 86,020` figure is withdrawn rather than refined. See the corpus section below.

```
[g 3prime] NC_TEST.1:g.[257_262TTT[1];261_264del]
   before: NC_TEST.1:g.[256_265T[7];262_265del]      was_fixed_point = false
   after : NC_TEST.1:g.[257_262TTT[1];261_264del]    was_fixed_point = true
```

`TTT[1]` over a 6-base tract removes `259-262`; the sibling deletion claims `261-264`. The two rewrite an intersecting range of reference bases, which is the collision `a_shrinking_repeat_collides_over_the_bases_it_removes` pins and which no pass caught before.

### The 6: an `inv` with a flush `ins`

All on the multi-exon coding axis, all one shape:

```
core CCCCCCCCTGACGTATCCTA, cx axis
  input : NM_TESTX.1:c.[5_6inv;5_6insCC]
  before: NM_TESTX.1:c.[-3_5C[10];5_6inv]     was_fixed_point = false
  after : NM_TESTX.1:c.[5_6inv;5_6insCC]      was_fixed_point = true
```

`before` had re-derived a **conflicting** allele into an accepted `[repeat; inv]` — an insertion at junction 5 strictly interior to an `inv` spanning 5-6, which is exactly the shape `issue_1437_dup_read_span_interior_sibling::an_inversion_still_conflicts_with_an_interior_insertion` pins as a genuine conflict (an inversion keeps the bases it spans, so the interior junction's position stays meaningful). `after` preserves the input, which is the documented conflict-preservation behaviour.

So the junction pass was right and the outcome that changed was a laundering: strict refused the input while the lenient path emitted an accepted re-derivation of it. The `was_fixed_point` column is the corroboration — every `before` was **not** a fixed point and every `after` is.

Reproduce:

```bash
cargo run --release --features dev --example dump_normalized_corpus -- --out base.tsv   # on e98fa77e
cargo run --release --features dev --example dump_normalized_corpus -- --out head.tsv   # on this branch
cargo run --release --features dev --example dump_normalized_corpus -- --compare base.tsv --against head.tsv
```

Both sides were dumped with **this branch's** `dump_normalized_corpus.rs` so the instrument is identical and only `src/` differs. Both dumps carry **86,398** rows and `--compare` refuses two dumps that do not cover the same rows, which is what makes the comparison meaningful across a revision that changed the corpus.

**Rebased onto `origin/main` `0c925605`** after the review fixes — the only conflict was `tests/it/main.rs`'s module list (#1748 and this PR each add a line); both kept. The branch's own delta is unchanged across the rebase: `git diff e98fa77e..<pre-rebase tip>` and `git diff 0c925605..HEAD` have an identical diffstat and differ only in blob hashes and hunk offsets. **The corpus figures above are measured against `e98fa77e`**, the PR's original base, and are quoted with that base rather than re-derived on every base move.

Local gate at this head (`FERRO_MANIFEST` set, `FERRO_REQUIRE_BULK_FIXTURES=1`, all four bulk fixtures fetched and digest-verified, both spec artifacts deleted and regenerated first): `GEN1_EXIT=0 GEN2_EXIT=0 FMT_EXIT=0 CLIPPY_EXIT=0 TESTS_EXIT=0`, **10,295 run / 10,295 passed / 0 failed / 150 skipped**, no `not run due to signal`. The new corpus guard lives in an `[[example]]` target, which `--lib --test it` does not build, so it was run separately: `cargo nextest run --features dev --example dump_normalized_corpus` — **43 passed**, including `the_repeat_sibling_family_varies_the_repeat_direction`. Attribution is by test **name**: the branch's test-function set gained exactly one name and lost none.

### The corpus could not see this change until a family was added

`repeat_expansion` feeds a repeat, but always a **lone** one, so no row in that corpus had ever placed a `repeat` beside a sibling it could collide with — and the repeat geometry is the main thing this PR changes. A `0 moved` would have been **structural**. `repeat_beside_a_sibling` closes it; this is the fifth instance of the class after member geometry (#1456), scale (#1460), transcript geometry (#1478) and molecule type (#1606).

**And it closed only half of it, which is the correction this PR now carries.** The family first shipped emitting `{unit}[3]` over a **6**-base tract with a **3**-base unit — `3 x 3 = 9 >= 6`, so every row was a *growing* repeat. `repeat_footprint`'s **shrinking** branch was still unreachable from the corpus, so the 102 rows it moves were invisible and the disclosed figure was `6 of 86,020`. The family now emits `[3]` and `[1]` for each of its three sibling geometries (378 -> 756 rows), and the arithmetic is guarded rather than commented: `the_repeat_sibling_family_varies_the_repeat_direction` asserts `unit.len() * count` lands on **both** sides of the tract length for every core, so widening the tract or the unit without re-picking the counts fails there instead of silently collapsing to one direction again.

The lesson is the one the class keeps teaching: fixing one blindness does not reveal the next, and a family added to close a blindness can itself be blind along the axis it was added for.

## The fourth decider is deliberately untouched

Row 3 above is refused at **parse** by `HgvsVariant::detect_self_cancelling_pair` (E3006), on **read** spans. By the write-footprint geometry that pair is disjoint and denotes one sequence, so unifying it in would have deleted the refusal.

It stays, because `general.md:58` names this edit-type pair explicitly and its own worked example (`NM_004006.2:c.[762_768del;767_774dup]`) overlaps at 767-768 **without cancelling** — so the clause keys on read spans by construction and cannot be derived from a write-footprint geometry. Precedence is spec-explicit over our judgement. What changes is that it is now documented as a scoped spec prohibition rather than sitting unremarked as a fourth opinion on the word "overlap". It is also the only overlap-adjacent refusal that fires at parse, where no error mode opts out — one of the cases `rulings[absolute-prohibition-enforcement-stage]` covers.

## Two stale claims corrected

- **The #1448 test's comment.** It said it asserts at unit level rather than end to end "because `detect_insertion_overlaps` reports an overlapping `dup` pair for its own separate reasons (#1437)". It does not, and **#1437 is precisely why it does not**: that change made `junction_overlaps` register a `dup` as a junction occupant and then stop, so it never becomes a span, and two `dup`s at different junctions match neither branch. Measured end to end on hg38 in both strict and lenient, the only warning `g.[43045721dup;43045721_43045722dup]` raises is `MEMBERS_COALESCED_FROM_REPORTED_FORM`. The detector **is** reached; it returns empty. A sibling test pins the correction so it cannot drift back.
- **`CLAUDE.md`'s false-positive table** attributed the flush-insertion class to a policy disagreement; the cause was the order dependence above, and the row now says so.

## Needs operator review: a descriptive correction to a decided ruling

`rulings[conflicting-member-geometry-refusal-scope]` (decided) says members "may not claim intersecting **reference territory**, whatever edit types they render as". The code says otherwise for `dup`, and has since #1411 — with the better argument.

The record's own rationale is un-denotability: "two members writing to one reference interval describe no single chromosome sequence". That rationale **supports the code**, because two `dup`s sharing a read span write at two different junctions and denote one unambiguous sequence. So the record's wording overreaches its own stated ground: "intersecting reference territory" is a read-span test, while the rationale is a write test.

**Proposed, not applied:** change the wording to "intersecting **write footprints**". This changes **no verdict** — `general.md:58`'s `del`x`dup` prohibition is unaffected (it is a separate, edit-type-scoped clause, see above), and no row moves. I have deliberately not touched the ledger: it is a decided record, the change is descriptive, and `RULING_STATUSES` plus the two `CLAUDE.md` tables are enforced against it, so the edit should be made knowingly rather than as a side effect of a refactor.

## Mutation matrix (the owed review item)

The earlier review recorded a mutation matrix as **not reached**, and the fix pass did not reach it either. It has now been run: **29 mutations**, each with its expectation stated *before* the run, every survivor re-checked against the **full 10,487-test gate** rather than a scoped selection. Baseline green at both levels, which is what makes the rest meaningful.

Call sites were mutated, not only predicates — that distinction is what found the second gap below.

**Two mutations survived, and both are now closed.** For each, the fix is not "a test was added" but "the mutation now fails":

| gap | before the fix | after the fix |
|---|---|---|
| `readable_bases`' reversed-range filter deleted | 10,487 passed, `EXIT=0` — survived, *including its own named test* | `EXIT=100`, **1 failed** — `a_reversed_range_is_declined_rather_than_judged` |
| `obstructs_splice_of`'s junction half bypassed **at its only call site** | 10,487 passed, `EXIT=0` — survived | `EXIT=100`, **1 failed** — `an_insertion_interior_to_a_deletion_is_declined_by_the_applier` |

The first was untestable as written: `a_reversed_range_is_declined_rather_than_judged` compared `spanning(16569, 5)` against `span(1, 10)`, where the intersection arm's *first* conjunct (`16569 <= 10`) is already false — so the assertion held whether or not the range was declined. The comparand has to span both stated endpoints; against `span(1, 20000)` an undeclined reversed range reports a conflict.

The second is the predicate-tested/call-site-untested shape. `the_splice_precondition_ignores_whether_bases_survive` pins the predicate and `shapes_without_one_resulting_sequence_decline` pins the caller's *base-intersection* half; nothing pinned that the caller still consults the junction half. It fails silently rather than loudly because each stated deletion is validated against the **pristine** `ref_bytes`, so a corrupted splice never trips the ref-match check — it returns a wrong sequence. The new test is the #1406 analogue at the caller's level, paired with a flush control so the decline cannot be satisfied by over-rejecting every ins+del pair.

Both closures are tests only — **no production code changed**, so the `Representation-Change:` disclosure above is unaffected.

### Stated gap: the `apply_triples` ordering fix has no direct test

Recorded here rather than closed, because it is a real observation about a change this PR argues for at length.

Removing the length tie-break from `apply_triples`' sort — the fix whose comment warns of *"one variant, two resulting sequences, decided by nothing the description states"* — **is** caught, but only by a single **pre-existing, indirect** corpus test, `from_sequences_corpus::every_spelling_of_a_class_derives_to_one_description`, which notices only because a wrong resulting sequence breaks its "one output per class" property.

**None of the three tests this PR adds in that area can reach it.** `a_flush_insertion_is_disjoint_in_either_member_order` and `disjointness_does_not_depend_on_member_order` both deliberately re-sort with the old `sort_by_key(|t| Reverse(t.position))` in order to prove the *predicate* is order-independent, so by construction they cannot see a change to the sort itself; and `an_insertion_flush_against_a_deletion_composes_in_either_order` asserts `strict_accepts` — acceptance, not the resulting sequence.

Worth knowing that `splice_denoted_sequence` already carried the same tie-break, so this PR brings `apply_triples` into agreement with it; removing it breaks only the latter and the two then disagree silently.

### Verified by measurement, not by re-reading

- **The blindness guard guards.** Forcing the generator back to `for count in [3]` — the exact pre-fix behaviour behind the 18x undercount — makes `the_repeat_sibling_family_varies_the_repeat_direction` **fail**, alone, with `3 growing and 0 shrinking repeats`. Forcing shrinking-only fails it too.
- **The 108 / 0 claim reproduces exactly.** Dumped with this PR's generator over merge-base `src/` and PR `src/` so both sides build the same corpus: 86,398 rows each side, 0 keys unique to either, **108** moved (**102** `repeat_beside_a_sibling` + **6** `junction_interior_to_span`), **0** fixed points before, **108** after — and all 108 have `output == input`, confirming the stated mechanism.

## Verification

- `cargo nextest run --features dev` (full gate, rebased onto `674e9c8b`) — **10,541 passed**, 0 failed
- `scripts/run_oracle_suite.sh` — **10,304 passed**, 0 failed
- `cargo fmt --all` — clean

TDD: the `FOOTPRINT_MATRIX` guard and the two `triples_are_disjoint` order tests were written first and watched fail — 2 of 23 matrix rows (both `repeat`) and `true vs false` on the same variant respectively. The matrix's expected column is derived from the write-footprint principle, not recorded from behaviour, which is what makes it a test of the definition rather than a change detector.


nh13:
@coderabbitai pause

PR #1760, opened 2026-08-13
docs(rulings): decide what delins.md:47 reaches — an inserted sequence, not any coincidence

nh13:
## What this decides

`delins-recommendation-reach-when-the-input-arrives-split` was the ledger's statement of an open question, and it closed with **"NO RULING HAS BEEN MADE."** `:47` recommends the spanning `delins` where payload bases coincide with reference bases; `:17` and `general.md:34` describe separated variants individually. With the input's own spelling already excluded as a tie-break by `canonical-form-choice-when-both-legal`, nothing said which of the two a re-derivation should land on.

**Operator ruling, 2026-08-12: `DNA/delins.md:46` governs.** `:47` reaches a payload-coincidence split **only where an inserted sequence re-aligned** — where some member of the split derived from the sequence supplies bases while consuming a *different* number of reference bases. Where no member does, nothing was inserted, `:46`'s mechanism cannot have occurred, and `:17` / `general.md:34` govern unqualified.

## Why this reading and not another

It is the only one under which **both** worked examples in the same passage come out as the spec writes them:

| example | canonical split | verdict |
|---|---|---|
| W49 `LRG_199t1:c.850_901delinsTTCCTCGATGCCTG` | carries `895_896delinsC` — two reference bases for one supplied | **merge**, which `:47` recommends by name |
| W58 `LRG_199t1:c.[992_1002del;1004T>C]` | three pure deletions, both payload bases survived | **stays split** |

A wide reading merges W58. A no-merge reading splits `:44-47`'s own example three lines from the clause that recommends merging it.

Recorded as a further **scope** on `delins-merge-vs-individual-gap-two-or-more`, sitting under that record's shape and net-deletion scopes and under `delins-payload-coincidence-carve-out-is-coding-dna-scoped`'s `c.`-only axis scope. It widens none of them.

## Why no output moves

`split_carries_a_gap_bearing_insert` (#1698) already implements this reading, and its only production call site is gated on `PartitionRule::CanonicalCoalesced`, which is not the shipped default. So this ratifies behaviour rather than changing it.

Three doc comments in `merge.rs` described the record as `undecided` and the predicate as provisional under README rule 6; they now say ratified. The comment on `an_all_survivor_payload_is_not_an_alignment_artifact` predicted that guard would be flipped by a ruling keyed on anything *other* than the derived members — it was right about the risk and the ruling deliberately went the other way, so the prediction is kept rather than deleted.

## Two things stated on the record rather than left implicit

- **The real-corpus measurement the record asked for is not here.** It is recorded as owed by the change that makes the canonical arm the default, on the ground that nothing shipped moves. That is an argument, and the record now carries it as one.
- **Confluence is not a ground, in either direction.** The coalesce runs last on already-derived pieces, so both spellings of a variant derive the same pieces and each class converges whichever way this went. Measured over the 11,272-class designed cis corpus at `1ea75334` (3'), divergent classes fall 2,911 → 1 and cross-class disagreements 16 → 0 — figures that belong to the **partition default**, not to this ruling, and the record says so.

The 4,620 / 655 split of the affected population is #1698's own instrumentation and was **not** re-derived here; the record labels it accordingly.

## Sidecars

`RULING_STATUSES`, both `CLAUDE.md` adjudication tables, and `clause_ruling_index.rs`'s generated block (re-captured programmatically, never retyped) plus its two census pins: the record census `(23, 20, 3)` → `(23, 21, 2)`, and the mixed-verdict count **7 → 2**.

That second pin is the round trip its own comment was waiting on. Filing this record as an open question took the count from 2 to 7; answering it returns exactly those five lines, leaving only the two ring lines that were mixed before it was ever filed. The rise was real and temporary, and it burned down because the question was answered rather than because a record was quietly dropped.

Representation-Change: none. Ruling record, sidecars and doc comments only; the predicate it ratifies shipped in #1698 and runs only on the non-default canonical-coalesced arm.


nh13:
@coderabbitai pause

PR #1762, opened 2026-08-13
fix(hgvs): stop dropping and leaking in-band position markers

### LINKED ISSUE #1754: tx_to_cds drops TxPos.downstream, answering n.*5 as c.-1
## Summary

`CoordinateMapper::tx_to_cds` never reads `TxPos.downstream`. It classifies a position purely by comparing `base` against the CDS bounds, so the `n.*N` marker is discarded and `n.*5` is answered as though it were in-transcript position 5 — a **different nucleotide**, with no diagnostic.

## Reproducer

On a transcript with a 5-base 5'UTR (`cds_start = 6`), single 38-base exon:

```rust
let mapper = CoordinateMapper::new(&tx);
mapper.tx_to_cds(&TxPos::downstream(5))  // n.*5
// => Ok(CdsPos { base: -1, .. })  i.e. `c.-1`
mapper.tx_to_cds(&TxPos::new(5))         // n.5
// => Ok(CdsPos { base: -1, .. })  the SAME answer
```

`n.*5` and `n.5` are not the same base, and `c.-1` is on the wrong side of the CDS entirely. Both calls return `Ok`.

## Why this is a refusal and not a mapping

- `background/numbering.md:52` gives the whole of `n.` numbering as "`n.1`, `n.2`, `n.3`, ..., etc., from the first to the last nucleotide of the reference sequence". The axis admits **no `*` zone**.
- `:54` states that "it is **not** allowed to describe variants in nucleotides beyond the boundaries of a transcript reference sequence, using that transcript reference sequence".

So `n.*N` names a nucleotide the `n.` axis cannot number, and there is no `c.` position to map it onto. The conversion must refuse rather than answer for a different base.

### Counter-evidence, recorded

The competing shape was to *honour* the flag: `Transcript::sequence_length` is available, so `n.*N` is mechanically `c.*(sequence_length - cds_end + N)` and the mapping is exact. Two things decide against it:

1. The derived `c.*M` would itself be past the transcript's last base, so honouring converts one description `:54` forbids into another rather than into a legal one.
2. The inverse would have to follow — `cds_to_tx` mapping an over-length `c.*M` back to `n.*N` — which moves outputs on the **coding** axis, where every real description lives, to serve an axis on which the shape does not occur.

On the other side of the ledger is ferro's own poly-A carve-out (#797), which deliberately entertains a `c.*` position in the post-transcriptional tail: ferro is not uniformly literal about `:54`. That is a ferro policy rather than a clause, so it does not outrank `:52` here, but it is why this is a decision and not a foregone conclusion.

## Note on the existing workaround

`project_*_direct` (`src/project/projector.rs`) already refuses `*N` inputs up front, and its comment says it does so *because* `tx_to_cds` never reads this flag. The judgement is therefore not new — the check simply lives at the wrong layer. It belongs on the conversion's own contract, where every caller gets it, including callers that never pass through the projector.

## Scope note

`cds_to_tx`'s `downstream: false` is **correct and must stay**. `CdsPos.utr3` (`c.*N`) is the 3'UTR of a *coding* transcript, which lies **inside** the transcript — `c.*1` is the base after the stop codon, still transcribed. `TxPos.downstream` (`n.*N`) marks a nucleotide past the transcript's last base. The two `*` markers are not the same marker, so this fix must not be "completed" later by making `cds_to_tx` carry the flag across.

### LINKED ISSUE #1755: Projection fan-out seed hard-codes false for TxPos.downstream, seeding n.*5 at n.5's genomic coordinate
## Summary

In `VariantProjector::extract_contig_and_pos` (`src/project/projector.rs`), the `HgvsVariant::Tx` arm passes a hard-coded `false` for the `*`-marker argument of `nr_representative_genome_pos`, while the `HgvsVariant::Rna` arm immediately below it correctly passes `start_rna.utr3`.

The result is that an `n.*5` input is seeded into the projection fan-out at the genomic coordinate of the **in-transcript** base 5.

## Reproducer

```rust
// Tx arm — the defect
let pos = nr_representative_genome_pos(cdot_tx, start_tx.base, start_tx.offset, false);
//                                                                             ^^^^^ ignores start_tx.downstream

// Rna arm, directly below — correct
let pos = nr_representative_genome_pos(cdot_tx, start_rna.base, start_rna.offset, start_rna.utr3);
```

On a single-exon fixture spanning genome `1000..1009` (tx `0..9`), seeding `n.*5`:

- observed: `1004` (the coordinate of base 5)
- contract: `1000` (the transcript's first-exon start, the documented fallback)

## Why `1000` is the contract, not a preference

`nr_representative_genome_pos`'s own doc comment already promises this. It lists the cases that fall back to the transcript's first-exon genomic start as "intronic offsets, **downstream**/utr3 markers, non-positive bases" — `downstream` is named explicitly. The `Tx` caller simply does not honour what the function documents.

The fallback is deliberate: for a position `tx_to_genome` cannot resolve directly, the stab query at the first-exon start still lands inside the transcript and the projection machinery re-resolves from there. Seeding at `1004` instead silently claims a precise coordinate for a position that has none.

## Suggested shape

Pass `start_tx.downstream`, and rename the parameter from `utr3` to something axis-neutral (e.g. `star_marker`). Calling it `utr3` is what invited the `n.` caller to pass a literal `false` in the first place: `TxPos` has no field of that name, so the argument read as inapplicable rather than as unhandled.

The parameter is genuinely axis-neutral in meaning — it is `RnaPos.utr3` on the `r.` axis and `TxPos.downstream` on the `n.` axis. The two mean different things about the sequence (`r.*N` is the 3'UTR, `n.*N` is past the transcript end) but they answer this function's only question identically: "is this a position `tx_to_genome` can resolve directly?", to which both say no.

## Testing note

The regression should assert on the **seed** rather than on the final projection, so it cannot be satisfied by a downstream repair that happens to mask the wrong seed.

### LINKED ISSUE #1756: TxPos/RnaPos Display leak the unknown-offset sentinel: n.5+? round-trips as n.5+9223372036854775807
## Summary

The parser stores the unknown intronic offsets `+?` / `-?` **in band**, as `i64::MAX` / `i64::MIN` (`OFFSET_UNKNOWN_POSITIVE` / `OFFSET_UNKNOWN_NEGATIVE`). `GenomePos::Display` and `CdsPos::Display` translate them back to `+?` / `-?`. `TxPos::Display` and `RnaPos::Display` have **no arm for them at all**, so the raw sentinel escapes into the rendered description as if it were a measured distance.

## Reproducer

```rust
parse_hgvs("NM_TEST.1:n.5+?del").unwrap().to_string()
// => "NM_TEST.1:n.5+9223372036854775807del"
// expected: "NM_TEST.1:n.5+?del"
```

Same on the `r.` axis:

```rust
parse_hgvs("NM_TEST.1:r.5+?del").unwrap().to_string()
// => "NM_TEST.1:r.5+9223372036854775807del"
```

and on the negative sentinel, `n.5-?del`, which renders `-9223372036854775808`.

This is a **round-trip failure through the public entry point**, not a latent constructor-only shape: the input parses, so the sentinel is reachable from any `n.`/`r.` description a user can write.

## Root cause

The pair is spelled three different ways across the four position types:

- `GenomePos::Display` compares against the named constants `GENOME_OFFSET_UNKNOWN_POSITIVE` / `GENOME_OFFSET_UNKNOWN_NEGATIVE`;
- `CdsPos::Display` re-spells the same pair as bare `i64::MAX` / `i64::MIN` literals;
- `TxPos::Display` and `RnaPos::Display` — added later — have no arm.

Three spellings of one pair is how the fourth axis came to have none. Nothing in the code marks the offset field as "may hold a sentinel", so a newly added `Display` has no signal that the question needs asking.

## Suggested shape

One helper, keyed off the named constants, that answers "is this offset a sentinel, and if so what does it render as", used by all four `Display` impls. That removes the duplicated spelling and gives any future axis a single thing to call.

Note there is nothing to adjudicate about the output form: the spelling is the one the parser consumed, and `CdsPos::Display` has rendered it that way since the sentinels were introduced. This is making two axes consistent with the two that were already right.

One deliberate asymmetry to preserve while refactoring: on `GenomePos` a **zero** offset renders as nothing, whereas the transcript-relative axes render `+0`. That is existing behaviour and is out of scope here.

nh13:
Three in-band marker defects, one class: data the AST carries in band that a boundary either silently discards or emits raw. They are fixed together because the same one-line omission produced the first two and the same missing `Display` arm produced the third, and reading them apart is what let each survive.

Closes #1754
Closes #1755
Closes #1756

Representation-Change: `n.<b>+?`, `n.<b>-?` and `r.<b>+?` now render as written
  instead of leaking the in-band sentinel — `n.5+?` previously round-tripped as
  `n.5+9223372036854775807`. 3 outputs move. No committed corpus contains an
  input that reaches them: the real corpora carry 709 unknown-offset
  descriptions and every one is on `c.`/`g.`, the two axes that already rendered
  `?`, and `n.*N` occurs 0 times in 103,762 `n.`-axis rows. Separately,
  `tx_to_cds` and the projection fan-out seed now refuse `n.*N` where they
  previously answered — an `Ok` becomes an `Err`, so no string moves, but it is
  a behaviour change and is disclosed here rather than left to be discovered.

## M1 — `tx_to_cds` dropped `TxPos.downstream` (#1754)

`CoordinateMapper::tx_to_cds` classified a position purely by comparing `base` against the CDS bounds, so the `n.*N` marker was never read. On a transcript with a 5-base 5'UTR, `n.*5` was answered as **`c.-1`** — a different nucleotide, on the wrong side of the CDS, with no diagnostic. `n.5` returned the same thing.

It now refuses, citing `background/numbering.md:52` (the `n.` axis is numbered "from the first to the last nucleotide of the reference sequence" and admits no `*` zone) and `:54` (a variant beyond a transcript's boundaries may not be described against that transcript).

The judgement is not new. `project_*_direct` already refused `*N` up front, and its comment says it does so *because* `tx_to_cds` ignored the flag. The workaround existed at the wrong layer; this moves the check onto the contract it belongs to, where every caller gets it — including the callers that never pass through the projector. That earlier refusal still runs first and is unchanged.

`cds_to_tx`'s `downstream: false` is deliberate and unchanged: `c.*N` is the 3'UTR of a *coding* transcript, which lies **inside** it (`c.*1` is the base after the stop codon, still transcribed). The two `*` markers are not the same marker. A new test pins this so nobody later "completes" M1 by making `cds_to_tx` carry the flag across.

## M2 — the fan-out seed hard-coded `false` for the same flag (#1755)

`extract_contig_and_pos`'s `Tx` arm passed a literal `false` to `nr_representative_genome_pos` while the `Rna` arm directly below it passed `start_rna.utr3`. So `n.*5` was seeded at the genomic coordinate of the *in-transcript* base 5 (1004 on the test fixture) rather than at the documented first-exon fallback (1000).

The function's own doc comment already promised otherwise — it lists "downstream/utr3 markers" among the cases that fall back to the first-exon start. The parameter is renamed `utr3` → `star_marker`, because calling it `utr3` is precisely what invited a caller on an axis with no field of that name to pass a hard-coded `false`. The regression asserts on the **seed**, not on the final projection, so it cannot be satisfied by a downstream repair masking a wrong seed.

## M3 — `TxPos`/`RnaPos` `Display` leaked the unknown-offset sentinel (#1756)

The parser stores `+?`/`-?` in band as `i64::MAX`/`i64::MIN`. `GenomePos` and `CdsPos` rendered them back; `TxPos` and `RnaPos` had no arm at all, so a parsed `n.5+?` round-tripped as `n.5+9223372036854775807`.

The pair had three spellings across four types — named constants in `GenomePos`, bare `i64::MAX`/`i64::MIN` literals in `CdsPos`, nothing in the other two. All four now go through one helper keyed off the named constants. That is the actual fix: three spellings of one pair is how the fourth axis came to have none.

## Why this is still load-bearing after the `n.*N` adjudication

The ruling is **refuse**, and #1751 is being changed to refuse `n.*N` at parse in all modes. This branch is not made redundant by that, for two reasons that are structural rather than defensive:

- an internally-constructed `TxPos` never passes through the parser at all;
- the Python `n_to_c(transcript_id, tx_position, downstream=True)` entry point builds a `TxPos` directly from numeric arguments — there is no HGVS string, so there is no parse stage to catch anything.

Both reach `tx_to_cds`. M2 and M3 are untouched by the parse-stage decision in any case.

## The error code: keeping `E5001`, and why not `W4008` or `E4003`

`tx_to_cds` refuses with `FerroError::ConversionError` → **E5001 `ConversionFailed`**. That is a generic bucket and the alternatives deserved a real look. Both were investigated and both are wrong, for reasons found in the code rather than assumed.

**`W4008 NonCodingPositionOutsideTranscript` is not reachable at conversion time, and would be wrong if it were.** Three findings:

1. **It has no structured error code to reuse.** W4008 lives in `error_handling::types::ErrorType`, a different enum from `error::ErrorCode`. In #1751 its reject arm returns `FerroError::Parse { pos: 0, msg, diagnostic: None }`, and the `W4008` identity is a **text prefix** interpolated into `msg`. Because `diagnostic` is `None`, `FerroError::code()` falls to its `_ => None` arm — so W4008 has no `ErrorCode` at all, anywhere. There is no structured value to carry into a conversion error; there is only a string.
2. **Carrying it would mean returning a parse error from a coordinate conversion.** The only way to emit the `[W4008]` text in the established shape is `FerroError::Parse { pos: 0 }`. `tx_to_cds` has no source string and no position — `pos: 0` would be a fabricated location, and the result would report a *parse* failure for something that parsed fine and failed at conversion. That is strictly worse than a generic-but-honest E5001, which at least returns a real structured code in the right category.
3. **Its contract is mode-scoped; this refusal is not.** W4008's whole definition is a mode triple resolved through `ErrorConfig::action_for`: strict rejects at parse, lenient warns and accepts, silent accepts quietly, per the decided `rulings[absolute-prohibition-enforcement-stage]`. `CoordinateMapper` is `struct CoordinateMapper<'a> { transcript: &'a Transcript }` — it holds no `ErrorConfig`, and it should not. `numbering.md:54` prohibits the *nucleotide*, not a spelling, so there is no CDS coordinate to return under any mode. Reusing the code would either make an absolute geometric fact mode-configurable, or leave a user who sets `ignore = ["W4008"]` with a knob that silently does not apply at this layer — a worse failure than a generic code.

**`E4003 UnsupportedProjection` reads as "not supported yet", and its live uses confirm that.** In `projector.rs` it is what the genuinely unimplemented arms return — `Protein`, `Mt`, `Circular`, `RnaFusion`, `GenomeRing`. Using it here would file "this question has no answer" alongside "we have not built this". It is also categorically an E4xxx *normalization* error, while this is a conversion. Notably, #1741's guard uses E4003 **correctly** for what it is: an explicitly temporary capability placeholder. Once the capability gap closes and the answer becomes "no such coordinate exists, ever", E4003 stops being right — the code changing *is* the meaning changing, and that is the correct signal rather than a regression.

**On "one input, one diagnosis".** That goal holds here, by disjoint reachability rather than by sharing a code. Exactly one refusal fires on any given call path: in strict mode #1751 refuses at parse and `tx_to_cds` never runs; in lenient/silent #1751 accepts and the refusal falls to `tx_to_cds`; with no parser involved only `tx_to_cds` fires. The two layers never both render a verdict on one call, they cite the **same two spec lines** (`numbering.md:52`, `:54`), and their stages differ because their mode contracts genuinely differ. Sharing one code across a mode-scoped parse rule and an unconditional geometric refusal would give one code two incompatible meanings — one input, two diagnoses wearing one name, which is worse than two codes.

**Follow-up worth doing, deliberately not done here.** The right end state is a dedicated `E5xxx` in the Conversion category, name-matched to W4008 so the two layers read as one finding at two enforcement stages. Minting it touches `docs/spec/error_code_audit.md` and `tests/fixtures/error_code_audit.json` — the exact two files #1751 is currently rewriting — so doing it now would put a guaranteed conflict into a PR that is still in flux. It should land once #1751 has settled.

## #1741's Python guard should be removed by whichever of the two lands second

#1741 adds `reject_unhonoured_downstream(downstream)` at the top of `PyCoordinateMapper::n_to_c`, before the provider lookup. It fires ahead of this branch's refusal, so today a Python caller's diagnostic depends on merge order.

Its own doc comment names the condition for its removal, and this branch is that condition:

> Honouring the flag is a change to the conversion and is tracked separately; until that lands the honest answer at the boundary is a refusal ... **Lift this once `tx_to_cds` honours the flag.**

Its stated premise also becomes false. The guard's message says the conversion "cannot be answered rather than silently answered as `downstream=False`" — once this lands the conversion *is* answered, with a refusal, and it is no longer silently answered as `downstream=False`. Leaving the guard in place ships a message describing a state that no longer exists.

I have deliberately not touched #1741's branch. The ask for whoever lands second:

- **Remove only `reject_unhonoured_downstream`** and its call. #1741's sibling guards `reject_zero_base` and `reject_sentinel_offset` are unaffected by this branch and must stay — this is not a request to remove the boundary-validation block.
- **The surviving refusal carries E5001**, raised as `ProjectionError` (`ferro_typed("ProjectionError", "Conversion error: …")`), so the Python-visible diagnostic is predictable: `ProjectionError` either way, with the code moving E4003 → E5001 for the reason argued above.
- **One ordering change is expected and is an improvement.** With the guard gone, `n_to_c("BOGUS", 5, downstream=True)` reports `ReferenceDataError` (transcript not found) instead of the downstream complaint, because the provider lookup now runs first. That is the more fundamental and more actionable diagnosis, and it makes `n_to_c` consistent with every sibling method on the class, all of which lead with the lookup. The guard's pre-lookup placement was an artifact of its being a boundary placeholder, not a designed ordering.

## Corpus measurement

`examples/dump_normalized_corpus.rs` reports **0 moved**, and that zero is **structural, not evidence**: the harness emits axis values `{c, cx, g, p}` only, so it cannot construct an `n.` or `r.` description and cannot reach any of the three defects. Reporting it as a real zero would be the failure mode #1456/#1460/#1478 keep producing.

The real corpora were measured instead:

- **709** unknown-offset descriptions, **all** on `c.`/`g.` — the two axes whose `Display` already rendered `?`. So M3 moves nothing that exists today.
- **`n.*N`: 0 occurrences in 103,762 `n.`-axis rows.** This is a *real* zero rather than a structural one: the same harvest over the same rows finds `n.-N` five times, so the shape was reachable and simply does not occur.

## Verification

`cargo fmt --all --check`, `cargo clippy --features dev --lib`, `cargo clippy --features python --lib` and the full `cargo nextest run --features dev` all pass. The `python` feature is linted explicitly because the `dev` feature set does not compile `src/python.rs`.

### DIFF
--- a/python/ferro_hgvs/__init__.pyi
+++ b/python/ferro_hgvs/__init__.pyi
@@ -2467,13 +2467,19 @@ class CoordinateMapper:
             offset: Optional intronic offset
             downstream: Whether this is a downstream position (n.*100 notation for
                 positions past the end of the transcript). Defaults to False.
+                A downstream position is REFUSED: it names a nucleotide beyond
+                the transcript's last base, which the n. axis cannot number and
+                which has no CDS position (HGVS background/numbering.md:52, :54).
+                It previously returned a position derived from `tx_position`
+                alone, i.e. a different nucleotide, with no error.
 
         Returns:
             Tuple of (cds_position, offset, is_utr3)
 
         Raises:
             ReferenceDataError: If the transcript is not found.
-            ProjectionError: If it has no CDS (both subclass ValueError).
+            ProjectionError: If it has no CDS, or if `downstream` is True
+                (both subclass ValueError).
         """
         ...
 
--- a/src/convert/mapper.rs
+++ b/src/convert/mapper.rs
@@ -125,6 +125,14 @@ impl<'a> CoordinateMapper<'a> {
             cds_start as i64 + pos.base - 1
         };
 
+        // `downstream: false` is a decision, not a default. `CdsPos.utr3`
+        // (`c.*N`) is the 3'UTR of a *coding* transcript, which lies **inside**
+        // the transcript — `c.*1` is the base after the stop codon, still
+        // transcribed — whereas `TxPos.downstream` (`n.*N`) marks a nucleotide
+        // past the transcript's last base. The two `*` markers are not the same
+        // marker, so carrying `utr3` across into `downstream` would be wrong;
+        // the arms above have already resolved `c.*N` to its plain `n.` base.
+        // Pinned by `cds_to_tx_keeps_a_coding_three_prime_utr_position_in_transcript`.
         Ok(TxPos {
             base: tx_base,
             offset: pos.offset,
@@ -138,7 +146,41 @@ impl<'a> CoordinateMapper<'a> {
     /// no exon walk, no CIGAR adjustment. The one input on which the round trip
     /// does not close is `c.0` — it converts to `cds_start` and comes back as
     /// `c.1`, because HGVS has no `c.0` for it to come back as.
+    ///
+    /// # Refuses `n.*N`
+    ///
+    /// [`TxPos::downstream`] marks the `n.*N` notation, which names a nucleotide
+    /// **past the transcript's last base**. There is no CDS position to convert
+    /// it to, so this refuses rather than answering:
+    ///
+    /// - `background/numbering.md:52` gives the whole of `n.` numbering as
+    ///   "`n.1`, `n.2`, `n.3`, ..., etc., from the first to the last nucleotide
+    ///   of the reference sequence" — no `*` zone exists on this axis;
+    /// - `:54` states that "it is **not** allowed to describe variants in
+    ///   nucleotides beyond the boundaries of a transcript reference sequence,
+    ///   using that transcript reference sequence".
+    ///
+    /// Before this refusal existed the flag was simply unread: classification
+    /// compared `base` against the CDS bounds, so `n.*5` on a transcript with a
+    /// 5-base 5'UTR was answered as `c.-1` — a different nucleotide, on the
+    /// wrong side of the CDS, with no diagnostic. That was already known and
+    /// worked around one layer up, where `project_*_direct` rejects `*N` inputs
+    /// "because `tx_to_cds` ... never reads this flag"; the check now lives with
+    /// the contract it belongs to, and that caller's earlier, better-worded
+    /// refusal still runs first.
     pub fn tx_to_cds(&self, pos: &TxPos) -> Result<CdsPos, FerroError> {
+        if pos.downstream {
+            return Err(FerroError::ConversionError {
+                msg: format!(
+                    "n.*{} lies beyond the transcript's last base and has no CDS position: \
+                     the n. axis numbers only n.1..n.<length> (background/numbering.md:52) \
+                     and a variant outside a transcript's boundaries may not be described \
+                     against that transcript (background/numbering.md:54)",
+                    pos.base
+                ),
+            });
+        }
+
         let cds_start = self
             .transcript
             .cds_start
--- a/src/hgvs/location.rs
+++ b/src/hgvs/location.rs
@@ -45,7 +45,15 @@ pub struct GenomePos {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub special: Option<SpecialPosition>,
     /// Optional offset from the base position (e.g., -? for uncertain upstream, +? for uncertain downstream)
-    /// Uses i64::MIN for uncertain negative offset (-?) and i64::MAX for uncertain positive offset (+?)
+    ///
+    /// The uncertain forms are stored in band, as
+    /// [`OFFSET_UNKNOWN_POSITIVE`] (`+?`) and [`OFFSET_UNKNOWN_NEGATIVE`]
+    /// (`-?`). Name the constants rather than their current values: they are
+    /// defined in exactly one place, and a doc that respells them as literals
+    /// is a second declaration in prose.
+    ///
+    /// [`OFFSET_UNKNOWN_POSITIVE`]: crate::hgvs::parser::position::OFFSET_UNKNOWN_POSITIVE
+    /// [`OFFSET_UNKNOWN_NEGATIVE`]: crate::hgvs::parser::position::OFFSET_UNKNOWN_NEGATIVE
     #[serde(skip_serializing_if = "Option::is_none")]
     pub offset: Option<i64>,
 }
@@ -100,10 +108,25 @@ impl GenomePos {
     }
 }
 
-/// Sentinel value for unknown positive offset (+?)
-pub const GENOME_OFFSET_UNKNOWN_POSITIVE: i64 = i64::MAX;
-/// Sentinel value for unknown negative offset (-?)
-pub const GENOME_OFFSET_UNKNOWN_NEGATIVE: i64 = i64::MIN;
+/// The unknown-offset sentinels, under their historical genome-axis names.
+///
+/// These are **re-exports**, not a second definition. The pair is defined once,
+/// by the parser that produces it
+/// ([`crate::hgvs::parser::position::OFFSET_UNKNOWN_POSITIVE`] and its negative
+/// twin), because the values enter the AST there and every axis — `g.`, `c.`,
+/// `n.`, `r.` — stores the same two.
+///
+/// They were previously declared here as their own `pub const`s, spelled as
+/// bare `i64::MAX` / `i64::MIN` literals. That is the same failure mode
+/// `unknown_offset_marker` was extracted to remove, one level up: two
+/// independent definitions of one value pair, where changing either leaves the
+/// other silently behind and the rendering stops matching the parser. Kept as
+/// aliases rather than deleted because they are public API of a published
+/// crate.
+pub use crate::hgvs::parser::position::{
+    OFFSET_UNKNOWN_NEGATIVE as GENOME_OFFSET_UNKNOWN_NEGATIVE,
+    OFFSET_UNKNOWN_POSITIVE as GENOME_OFFSET_UNKNOWN_POSITIVE,
+};
 
 /// The rendered form of an unknown-offset sentinel (`+?` / `-?`), or `None`
 /// when `offset` is a measured intronic distance.
--- a/src/hgvs/location.rs
+++ b/src/hgvs/location.rs
@@ -105,6 +105,29 @@ pub const GENOME_OFFSET_UNKNOWN_POSITIVE: i64 = i64::MAX;
 /// Sentinel value for unknown negative offset (-?)
 pub const GENOME_OFFSET_UNKNOWN_NEGATIVE: i64 = i64::MIN;
 
+/// The rendered form of an unknown-offset sentinel (`+?` / `-?`), or `None`
+/// when `offset` is a measured intronic distance.
+///
+/// The sentinels are stored **in band** (`i64::MAX` / `i64::MIN`, see
+/// [`crate::hgvs::parser::position::OFFSET_UNKNOWN_POSITIVE`]), so every
+/// `Display` that prints an offset has to ask this question first or the raw
+/// 19-digit integer escapes into a description as if it were a real distance.
+///
+/// It exists as one function, keyed off the named constants, because the
+/// alternative had already failed: `GenomePos::Display` compared against
+/// `GENOME_OFFSET_UNKNOWN_*`, `CdsPos::Display` re-spelled the same pair as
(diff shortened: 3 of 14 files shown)


nh13:
@coderabbitai pause

PR #1770, opened 2026-08-13
fix(cdot): propagate cds_start_incomplete through Transcript to CdotTranscript

### LINKED ISSUE #1771: `CdotMapper::from_transcripts_with_build` shadows `cds_start_incomplete` with `false`, defeating the #972 decline
`CdotMapper::from_transcripts_with_build` builds every `CdotTranscript` with `cds_start_incomplete: false`, discarding the flag carried by the `Transcript` it is converting. Because the projector consults the cdot transcript **in preference to** the `Transcript`, the hardcoded `false` does not merely lose information — it **shadows a correct `true`**, and a `cds_start_NF` transcript is annotated with a `c.`/`p.` description that #972 requires ferro to decline.

Filed retroactively; the fix is #1770.

## The defect

`src/data/cdot.rs`, in `CdotMapper::from_transcripts_with_build`, on the reverse `Transcript` → `CdotTranscript` direction:

```rust
// Not propagated here (out of this task's scope per brief);
// a follow-up can wire tx.cds_start_incomplete through this
// reverse Transcript → CdotTranscript direction too.
cds_start_incomplete: false,
```

The gap was deliberate and self-documented when written. What was not recorded is that a consumer prefers this value over the correct one.

## Why it shadows rather than drops

`src/project/projector.rs:5950-5956` resolves the flag cdot-first, falling back to the `Transcript` only on a lookup miss:

```rust
let cds_start_incomplete =
    match self.projector.mapper().cdot().get_transcript(transcript_id) {
        Some(t) => t.cds_start_incomplete,
        None => self
            .cached_get_transcript_for_variant(normalized, transcript_id)
            .map(|tx| tx.cds_start_incomplete)
            .unwrap_or(false),
```

So when a mapper is built through the lossy path, the lookup **succeeds** and returns `false`, and the correct `true` on the `Transcript` is never consulted. The same value gates the decline at `projector.rs:5586` and `:5802`.

Observable symptom, from the regression test added in #1770 — a `cds_start_NF` transcript that must decline instead yields:

```
c. must decline, got Some("NC_000001.11(NM_TEST.1):c.4C>A")
```

## Reachability

Narrow, and worth stating precisely rather than overstating.

The only non-`#[cfg(test)]` callers are `src/python.rs` (`PyProvider::cdot_mapper`, and only its `PyProvider::Mock` arm — the `MultiFasta` arm returns the provider's own already-built mapper and never takes this path) and `src/conformance/reference_snapshot.rs` (`ReferenceSnapshot::to_cdot`, whose in-repo callers are all test code). Roughly a dozen `tests/it/*` suites also build mappers this way.

So this is **not** a live annotation defect against a real prepared reference. It is a defect on the mock/test-data path — which is worse than it sounds, because it makes a passing test suite misrepresent a guarantee.

## Why no test caught it

`tests/it/issue_972_cds_start_nf_decline.rs`'s `fixture()` hand-seats a `CdotTranscript` directly:

```rust
fn fixture(incomplete: bool) -> (Projector, MockProvider) {
    let mut cdot = CdotMapper::new();
    cdot.add_transcript(
        "NM_TEST.1".to_string(),
        CdotTranscript {
            cds_start_incomplete: incomplete,
```

It therefore never exercises the construction path that drops the flag. The suite that exists to guard #972 was structurally blind to a way of losing it — a partial run and a clean run write indistinguishable artifacts.

## Not affected

`src/vcf/to_hgvs.rs:180` gates on `self.transcript.cds_start_incomplete`, but that field is a `reference::transcript::Transcript` — the *input* to this conversion, not the `CdotTranscript` product. It cannot observe the loss and is unchanged by the fix. (An earlier internal write-up named it as the consuming gate; that was a misattribution.)

`src/reference/derived_tx_structure.rs` propagates correctly and pins it with `derives_propagates_cds_start_incomplete_from_sibling`. Its one non-test `cds_start_incomplete: false` is an inert local probe whose field is never read.

## Fix

#1770 propagates the flag and adds two tests, both of which fail on the previous behaviour: a unit test covering `from_transcripts` and `from_transcripts_with_build` (with a `false` guardrail, so the field is shown propagated rather than flipped), and an end-to-end test through the projector with the mapper derived via `from_transcripts` — closing the fixture blind spot above.

nh13:
`CdotMapper::from_transcripts_with_build` builds a `CdotTranscript` from a `Transcript` and hardcoded `cds_start_incomplete: false`, dropping the `cds_start_NF` flag on that reverse direction. The existing comment documented the gap as a deliberate follow-up ("a follow-up can wire `tx.cds_start_incomplete` through this reverse `Transcript` → `CdotTranscript` direction too"). This is that follow-up.

## Why it is not merely lossy

The #972 decline gate in `src/project/projector.rs` consults the **cdot** transcript in preference to the source `Transcript` — `terminus_multiaxis_projection` falls back to the `Transcript` only when the cdot lookup misses:

```rust
let cds_start_incomplete = match self.projector.mapper().cdot().get_transcript(transcript_id) {
    Some(t) => t.cds_start_incomplete,
    None => self.cached_get_transcript_for_variant(normalized, transcript_id)
                .map(|tx| tx.cds_start_incomplete)
                .unwrap_or(false),
};
```

So the hardcoded `false` did not just lose information, it **shadowed** a correct `true` carried on the `Transcript`, re-enabling the `c.`/`p.`/`r.` axes for a transcript with no confirmed start codon. Reproduced directly: before this change the new end-to-end test emitted `NC_000001.11(NM_TEST.1):c.4C>A` for a `cds_start_NF` transcript.

## Reachability — deliberately narrow, and stated plainly

This is **not** a live annotation defect against a real prepared reference. The construction path has two non-`#[cfg(test)]` callers in the library:

- `src/python.rs`, `PyProvider::cdot_mapper()` — and only via the `PyProvider::Mock` arm. The `MultiFasta` arm returns the provider's own already-built mapper and never takes this lossy path.
- `src/conformance/reference_snapshot.rs`, `ReferenceSnapshot::to_cdot` — a public conformance-harness helper whose in-repo callers are all test code.

Beyond those, roughly a dozen `tests/it/*` suites build their mapper this way. So the reachable impact is the Python mock / test-data / conformance-harness path: the defect made **tests** lie about a guarantee the library otherwise upholds, which is worth fixing precisely because a green suite was the evidence for that guarantee.

Worth noting what this is *not*: `src/vcf/to_hgvs.rs:180` (`if self.transcript.is_coding() && !self.transcript.cds_start_incomplete`) reads a `reference::transcript::Transcript`, i.e. the *input* to this conversion, so that annotation gate never observed the loss and is unaffected.

## Tests

Both fail before the one-line change and pass after.

- `data::cdot::tests::test_from_transcripts_propagates_cds_start_incomplete` — asserts the flag survives both `from_transcripts` and `from_transcripts_with_build`, plus a `false` guardrail so the field is shown to be propagated rather than flipped.
- `issue_972_cds_start_nf_decline::cdot_built_from_transcripts_still_declines_coding_protein_rna` — end-to-end through the projector gate, with the mapper derived by `from_transcripts` rather than hand-seated as the existing `fixture()` does (which is exactly why the existing #972 suite could not see this). Asserts `c.`/`p.`/`r.` decline while `n.`/`g.` still render.

## Scope

One production line plus tests. `src/reference/derived_tx_structure.rs` already propagates correctly and is untouched. The diff is kept minimal and localized so it and #1737, which also edits `src/data/cdot.rs`, rebase trivially over one another whichever lands second.

Full local gate green: `cargo fmt --all --check`, `cargo clippy --features dev --lib -- -D warnings`, `cargo clippy --features python --lib -- -D warnings` all exit 0; `cargo ta` runs 10431 tests, 10431 passed, 0 failed.

Representation-Change: none. `src/data/` only; nothing under `src/normalize/`, `src/hgvs/`,
  `src/spdi/` or `src/project/` is touched and no normalized output moves. The behavioural
  change is confined to restoring the #972 projection-axis decline on transcripts built
  through this construction path.

Closes #1771

### DIFF
--- a/src/data/cdot.rs
+++ b/src/data/cdot.rs
@@ -1653,6 +1653,11 @@ impl CdotMapper {
     ///   inclusive end equals 0-based exclusive end).
     /// - `cds_start`: 0-based. Conversion: `cds_start = Transcript.cds_start - 1`.
     /// - `cds_end`: 0-based exclusive. Same numeric value as `Transcript.cds_end`.
+    /// - `cds_start_incomplete`: propagated verbatim from
+    ///   `Transcript.cds_start_incomplete` (the Ensembl `cds_start_NF` flag, #972). Not a
+    ///   coordinate, but listed here because the projector's decline gate reads the *cdot*
+    ///   record in preference to the source `Transcript`, so dropping it on this edge
+    ///   silently re-enables the `c.`/`p.`/`r.` axes for such a transcript.
     ///
     /// Defaults to GRCh38 for contig alias resolution (so `chr17` ↔ `NC_000017.11`
     /// works after construction). Use [`from_transcripts_with_build`] to specify a
@@ -1728,10 +1733,12 @@ impl CdotMapper {
                 // needing a cdot JSON. See #310.
                 protein: tx.protein_id.clone(),
                 exon_cigars: Vec::new(),
-                // Not propagated here (out of this task's scope per brief);
-                // a follow-up can wire tx.cds_start_incomplete through this
-                // reverse Transcript → CdotTranscript direction too.
-                cds_start_incomplete: false,
+                // #972: carry the `cds_start_NF` flag through this reverse
+                // Transcript → CdotTranscript direction. The projector's
+                // decline gate reads the cdot transcript in preference to the
+                // source `Transcript`, so hardcoding `false` here shadowed a
+                // correct `true` and re-enabled c./p./r. for such transcripts.
+                cds_start_incomplete: tx.cds_start_incomplete,
             };
             mapper.add_transcript(tx.id.clone(), cdot_tx);
         }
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #1775, opened 2026-08-13
ci: generate and publish the n./r. cis confluence corpus, and guard the wiring

nh13:
Representation-Change: none. CI wiring and test-harness only; no watched directory is touched.

## The two `rna_*_confluence_census` tests were not slow — they were building the crate

`tests/it/cis_confluence_nr_axis.rs` landed in #1646 with no wiring in the `spec-fixtures` job. That job generates and publishes only the `g,c` corpus (`cis_confluence_corpus.json`); the `n,r` corpus it needs was never generated and never rode the `cis-corpus` artifact.

So every test in that module hits `common::fixture_gen`'s on-demand fallback: a nested `cargo run --features dev --example generate_cis_confluence_corpus -- --axes n,r`. The `test` / `test-oracle` shards run from a nextest **archive** and have no warm `target/`, so that compiles the whole crate *inside the test* — and the cost is per test **process**, which nextest gives every test.

The `g,c` sibling is the control: it runs **twice** the spellings and was three times faster, because *its* corpus was downloaded.

## Measured, `Test oracle`, before and after

Before is run 31670057963 (`main`); after is this PR's run 31671849564. Same job family, same corpus, same shard assignment.

| test | before | after |
|---|---|---|
| `cis_confluence_nr_axis::the_rna_axis_is_spelled_in_the_rna_alphabet` | 129.2s | **0.83s** |
| `cis_confluence_nr_axis::rna_five_prime_confluence_census` | 205.9s | **41.2s** |
| `cis_confluence_nr_axis::rna_three_prime_confluence_census` | >180s | **38.8s** |
| `cis_confluence_axis::three_prime_confluence_census` (`g,c`) | 88.0s | **55.2s** |

**Nothing is over nextest's 60s SLOW threshold any more.** The `g,c` row is the clean read on the rayon commit alone — that test never paid a nested build, so its 88.0 → 55.2s is parallelism and nothing else. That settles the question this PR opened as a draft: rayon is a real ~1.6× win on a 4-vCPU runner *even with nextest already running other tests concurrently*, not a wash.

The other censuses, from the sibling shards: `noncoding_three_prime` 31.5s, `noncoding_five_prime` 32.7s, `five_prime_confluence_census` 47.6s, `the_nr_corpus_is_a_dense_set…` 0.58s.

## Three commits

**`ci:` generate and publish the `n,r` corpus.** A second generation step (`--axes` defaults to `g,c` and `--output` to the other corpus, so both flags are required), the file added to the `generated-fixtures` cache restore *and* save paths, and both corpora on the one `cis-corpus` artifact — they share a directory, so the upload roots at their common ancestor and the consumers' `path: tests/fixtures/cis/` is unchanged. Verified the step's exact command reproduces byte-for-byte the corpus the tests read.

Also sets `if-no-files-found: error` on both fixture uploads, matching the `bulk-fixtures` sibling. Without it a generation step that stops writing produces a warning and an artifact missing the corpus, and every consumer falls straight back to the nested build — the same silent degradation, re-entered from the other side.

Twelve test processes per run stop paying ~120s each: six tests in the module, across the `test` (6 shards) and `test-oracle` (3 shards) families.

**`perf(test):` one reference group per rayon task.** Both confluence modules already built one provider and one normalizer per reference group; that grouping is also what makes the groups independent. The result is byte-identical to the serial walk rather than merely equivalent, which is the only basis on which a pinned census may be parallelized: every `Census` field counts over a partition of the classes so the fold is exact and order-free; `par_iter().collect()` preserves input order; and `worst` — "the first 10 divergent classes in corpus order" — is reproduced by capping per group and truncating the concatenation. All eleven pins across the two modules are unchanged.

`the_reference_groups_partition_each_axis_in_corpus_order` pins that premise. It is not decoration: the censuses would still sum to *something* if a class were dropped or double-counted, and `worst` only ever surfaces inside a failure message, so a broken partition is invisible on a green run and would move the pins silently. rayon is a non-optional dependency, so nothing is added to the tree.

**`test(ci):` refuse a generated fixture CI does not build and ship.** The guard for the class, not the instance. Three properties, each derived from the workflow rather than from a roster in the test — a hardcoded list would rot exactly the way the defect did:

1. every generated fixture path a `tests/it` module names is in **every** `actions/cache` step's `path:` in `spec-fixtures`, restore and save alike;
2. it is published on some `actions/upload-artifact` from that job;
3. every job downloading the `nextest-archive` artifact also downloads each artifact carrying a generated fixture — **and unpacks it onto the path the test actually reads**. Downloading is not enough: `upload-artifact` roots a multi-path artifact at its paths' common ancestor, so `cis-corpus` holds bare basenames and only lands correctly because the consumer sets `path: tests/fixtures/cis/`. A `path: .` would put every corpus in the repository root and send every test back to the nested build.

Both halves were checked against deliberate sabotage rather than assumed. Against the pre-fix `ci.yml`, both tests go red naming the missing path; with the `cis-corpus` destination rewritten to `path: .`, the third property goes red naming the job, the destination, and where the fixture would land.

The workflow half is parsed rather than grepped, for the reason `workflow_gh_repo_context.rs` gives: which step sits in which job is structural. The source half is a token scan and is documented as a **floor, not a proof** — a module composing its path at run time is invisible to it. Both halves carry a denominator assertion, so a matcher that stops matching fails loudly instead of passing as "nothing to check".

## Note on one earlier claim, now withdrawn

An earlier revision of this description flagged that `cis_confluence_axis::five_prime_confluence_census` read 89.4s locally against `three_prime_confluence_census`'s 7.3s, and said I could not tell whether that asymmetry was the 5' direction's own cost or something this change introduced. CI answers it: 47.6s and 55.2s respectively, with 5' the *faster* of the two. The local reading was a laptop artifact, not a regression.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -634,6 +634,7 @@ jobs:
             tests/fixtures/grammar/hgvs_spec_normalization.json
             tests/fixtures/grammar/hgvs_spec_enumeration.json
             tests/fixtures/cis/cis_confluence_corpus.json
+            tests/fixtures/cis/cis_confluence_nr_corpus.json
           key: generated-fixtures-${{ runner.os }}-${{ github.sha }}
       - name: Generate HGVS v21.0 spec fixture
         if: steps.generated-fixtures.outputs.cache-hit != 'true'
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

PR #1776, opened 2026-08-13
test(convert): pin that the sequence and genome frames disagree on a gapped transcript

### LINKED ISSUE #1774: Name the frame at each CDS<->tx conversion site, and pin that the two frames must disagree on a gapped transcript
The CDS↔transcript conversion exists twice in the tree, and both copies are **correct** — for different axes. Nothing at the sites that still lack it says which axis each is right against, and no test pins that the two must *disagree*. So the pair reads as duplication inviting a collapse, and collapsing it reintroduces #1619.

This is a naming/documentation issue plus a guard. **No behaviour is wrong today.**

All line numbers are against `origin/main` at `57f71d31`.

## The two families

**Sequence frame** — flat, never reads the exon table. A `c.`/`n.` position names a base of the *transcript reference sequence*, so `c.N` is `cds_start + N - 1`. `ReferenceProvider::get_sequence` on a transcript accession serves that flat sequence, so this is the only offset that can index it.

| site | |
|---|---|
| `src/convert/mapper.rs:90` | `CoordinateMapper::cds_to_tx` |
| `src/convert/mapper.rs:128` | `CoordinateMapper::tx_to_cds` |
| `src/normalize/mod.rs:11728` | `Normalizer::cds_to_tx_pos` (private copy) |
| `src/normalize/mod.rs:11765` | `Normalizer::tx_to_cds_pos` (private copy) |
| `src/normalize/mod.rs:144` | `cds_pos_to_tx_boundary` (free helper; its own doc calls itself a mirror of `cds_to_tx_pos`) |

**Genome frame** — exon- and CIGAR-aware, and correctly so. Mapping between a contig and a transcript really does walk the alignment, and a transcript-coordinate gap there really does mean "this transcript base aligns to nothing".

| site | |
|---|---|
| `src/convert/mapper.rs:249` | `CoordinateMapper::genomic_to_tx` |
| `src/convert/mapper.rs:293` | `CoordinateMapper::genomic_to_tx_with_intron` |
| `src/convert/mapper.rs:334` | `CoordinateMapper::tx_to_genomic` |
| `src/convert/mapper.rs:520` | `CoordinateMapper::tx_to_genomic_with_intron` |
| `src/project/projector.rs:5328` | `project_single_inner` |

The composed conversions are one step of each — exon-aware across genome↔tx, flat across tx↔CDS: `src/convert/mapper.rs:382` (`genomic_to_cds`) and `src/convert/mapper.rs:405` (`cds_to_genomic`).

## The invariant

**On a transcript whose cdot alignment carries a transcript-coordinate gap, the two families must return different answers for every coding position 3' of the gap.** That divergence is the design, not drift: one family is answering "which base of the served sequence is this", the other "which genomic base does this align to", and on a gapped transcript those have different answers.

## Why file it

#1619 is what getting this wrong costs. `hgvs_to_spdi` resolved a `c.` position by walking the exon list while the normalizer indexed the flat transcript, so on `NM_033517.1` — which carries a genuine 39-base cdot hole between exons 10 and 11 — a *correct* normalization read as a corruption. It took a full investigation and a rebuilt evidence base to settle; #1735 is the fix and the adjudication, and `c-and-n-positions-are-flat-transcript-offsets` in `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json` is the ruling record.

What #1735 left is the naming half:

- `src/convert/mapper.rs:25-63` (the module's "two frames" note), `cds_to_tx` and `tx_to_cds`, and `src/project/transcript_axis.rs` all name the sequence frame explicitly — that part is done.
- The **genome-frame** methods (`genomic_to_tx:249`, `tx_to_genomic:334`) name no frame and do not cross-link the sequence frame.
- The **normalizer's private copies** (`src/normalize/mod.rs:11728`, `:11765`) carry a bare `/// Convert CDS position to transcript position` and no frame at all — which is exactly the pair a reader would look at and conclude should route through `CoordinateMapper`'s exon-aware side.
- Nothing **tests** the disagreement. `cds_to_tx_ignores_a_transcript_coordinate_gap` (`src/convert/mapper.rs:1033`) pins the sequence-frame half in isolation; no guard fails when the two families are made to agree.

The name collision makes it worse: there are two unrelated `pub struct CoordinateMapper` types, at `src/convert/mapper.rs:69` and `src/data/mapping.rs:42`.

## Not this

**#1665 is not an open proposal and should not be revived as written.** It reached the same removal #1735 later landed, but argued it from the premise that transcript-coordinate gaps *are malformed fixture data cdot does not produce*, evidenced by a 20-transcript sample. That premise is measured false: over the full prepared cdot table, **58 of 474,818** GRCh38 multi-exon builds carry a transcript-coordinate gap (23–2718 bases) and **159 of 190,754** on GRCh37 (8–7074). Eight of the 58 reach ferro's `Exon` list with no supplemental-CDS fallback. The gaps are real; what settles the frame question is *whose coordinate space they are in*, not whether they exist.

## Proposed

1. Doc comments at the genome-frame sites and at the normalizer's two private copies naming the frame each is correct against and cross-linking the other.
2. A cross-test on a programmatically built gapped transcript pinning that the two frames **disagree** past the gap (and still agree before it), with a failure message explaining why the disagreement is correct — so a future unification fails loudly instead of silently reintroducing #1619. Built without a prepared reference or downloaded cdot, so it always asserts rather than skipping.

## Correction owed to the coordinate-seam audit

A prior internal audit of the coordinate seam classified this divergence (its §5.3 row) as **U — unjustified, duplication drift**, and listed the semantic-necessity argument as its own strongest falsifier. The falsifier holds: the row should be **J — justified**, with #1619/#1735 as the authority. Anything downstream that inherited the `U` classification should be read with that correction attached.

nh13:
Closes #1774

The CDS↔transcript conversion exists twice in the tree and both copies are **correct**, for different axes. Nothing pinned that, so the pair read as duplication inviting a collapse — and collapsing it reintroduces #1619, a CRITICAL that took a full investigation and a rebuilt evidence base to settle (#1735 is the adjudication; `c-and-n-positions-are-flat-transcript-offsets` in the ruling ledger is the record).

This adds the guard and finishes the naming. No behaviour changes.

## The guard

`tests/it/axis_frame_disagreement.rs` builds a gapped transcript with the geometry of `NM_033517.1` — a transcript-coordinate gap inside a CDS that is annotated contiguously across it — and pins that the two frames answer differently:

- `a_transcript_coordinate_gap_exists_in_one_frame_and_not_the_other` — `tx_to_genomic` returns `None` for a gap base while `tx_to_cds` gives it a real `c.` number. The gap is a fact about the alignment, not about the numbering.
- `sequence_frame_and_genome_frame_disagree_past_a_gap` — for `c.220`, the flat conversion answers `n.270` and a walk of the same number of *genome-aligned* bases answers `n.309`. They must differ, and by exactly the gap width.
- `sequence_frame_and_genome_frame_agree_before_the_gap` — 5' of the gap they coincide, which is what every ungapped transcript looks like and why a collapse passes casual review.
- `the_two_frames_name_different_bases_of_the_served_transcript` — the two offsets index different bases of the flat sequence a reference provider actually serves (`C` vs `G` at the two marker positions). This is #1619's symptom in one assertion.
- `the_sequence_frame_round_trips_across_the_gap` — the flat inverse still closes, gap or no gap.

The genome-frame side of every comparison is expressed through the real public API (`tx_to_genomic`), not through a test-local reimplementation, so unifying the two families in **either** direction turns the file red. The failure messages say why the disagreement is correct rather than leaving the next reader to rediscover it.

The fixture is built programmatically and depends on no prepared reference and no downloaded cdot, so the test always asserts — it can never degrade into a silent skip.

**Sabotage-checked.** With `cds_to_tx` temporarily made exon-aware (adding each preceding gap's width to the flat offset — the #1619 shape), the guard goes red: `5 tests run: 2 passed, 3 failed`.

```
FAIL sequence_frame_and_genome_frame_disagree_past_a_gap
  assertion `left == right` failed: c.220 must be cds_start + N - 1 on the flat
  transcript (51 + 220 - 1 = 270). `get_sequence` on a transcript accession serves
  the flat sequence, so an exon-walked offset cannot index it — that was #1619.
    left: 309
   right: 270

FAIL the_two_frames_name_different_bases_of_the_served_transcript
  assertion `left == right` failed: the sequence-frame offset must index the base
  c.220 denotes on the transcript the reference provider actually serves
    left: 71     ('G')
   right: 67     ('C')

FAIL the_sequence_frame_round_trips_across_the_gap
  assertion `left == right` failed: c.151 is a flat offset from cds_start
    left: 240
   right: 201
```

Each of the three fires on the individual frame pin rather than on the `assert_ne!`, which is the better failure: it names the arithmetic that changed instead of only reporting that two numbers coincided. The two survivors are the ones the sabotage does not reach — the gap census, and the 5'-of-the-gap agreement that a collapse leaves intact by construction.

The sabotage was then reverted and the file restored byte-for-byte (md5 verified identical to the pre-sabotage copy).

## The naming

Doc comments only — no function body is touched.

- `src/convert/mapper.rs` — `genomic_to_tx` and `tx_to_genomic` now say they are the **genome frame**, cross-link `cds_to_tx`/`tx_to_cds`, and point at the guard. (The module header and the two sequence-frame methods were already named by #1735; these two were not.)
- `src/normalize/mod.rs` — `Normalizer::cds_to_tx_pos` and `tx_to_cds_pos` carried a bare `/// Convert CDS position to transcript position` and named no frame at all. They are the pair a reader is most likely to conclude should route through the exon-aware mapper, so they now state that they are the sequence frame, why the normalizer needs it (its offsets index the flat sequence `get_sequence` serves), and that the genome-frame conversions are not a substitute.

## Note on #1665

#1665 is closed and its premise should not be restated as open: it argued that transcript-coordinate gaps are malformed fixture data cdot does not produce, on a 20-transcript sample. Measured over the full prepared cdot table, **58 of 474,818** GRCh38 multi-exon builds carry one (23–2718 bases) and **159 of 190,754** on GRCh37 (8–7074). The gaps are real. What settles the frame question is whose coordinate space they are in, which is what #1735 established.

Representation-Change: none. Tests and doc comments only; no function body under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is modified, and the two `src/convert/mapper.rs` edits are doc comments on methods whose arithmetic is unchanged. 0 rows move.

### DIFF
--- a/src/convert/mapper.rs
+++ b/src/convert/mapper.rs
@@ -87,6 +87,17 @@ impl<'a> CoordinateMapper<'a> {
     /// `c.N` is `cds_start + N - 1` on the flat transcript, with no adjustment
     /// for exon geometry, for a CIGAR insertion (#944), or for a
     /// transcript-coordinate gap in the genome alignment (#1619).
+    ///
+    /// One input where this is **not** interchangeable with the normalizer's
+    /// own sequence-frame copy (`Normalizer::cds_to_tx_pos`): `c.0`, which is
+    /// not a valid HGVS position but is reachable (`CDS_BASE_UNKNOWN` is `0`,
+    /// and cis-collapse can build a `base == 0` anchor). It falls into the
+    /// `base < 1` branch here and answers `cds_start`, where the normalizer
+    /// continues the formula and answers `cds_start - 1`. Which is right needs
+    /// its own ruling; see #1772. A `c.-N` further upstream than the transcript
+    /// is long diverges too: this method returns a `TxPos` with a *negative*
+    /// `base` (which [`Self::fold_non_intronic_utr_offset`] relies on), where
+    /// the normalizer's copy errors.
     pub fn cds_to_tx(&self, pos: &CdsPos) -> Result<TxPos, FerroError> {
         let cds_start = self
             .transcript
@@ -123,8 +134,10 @@ impl<'a> CoordinateMapper<'a> {
 
     /// Convert transcript position to CDS position
     ///
-    /// The exact inverse of [`Self::cds_to_tx`], and on the same flat
-    /// sequence axis: no exon walk, no CIGAR adjustment.
+    /// The inverse of [`Self::cds_to_tx`], and on the same flat sequence axis:
+    /// no exon walk, no CIGAR adjustment. The one input on which the round trip
+    /// does not close is `c.0` — it converts to `cds_start` and comes back as
+    /// `c.1`, because HGVS has no `c.0` for it to come back as.
     pub fn tx_to_cds(&self, pos: &TxPos) -> Result<CdsPos, FerroError> {
         let cds_start = self
             .transcript
@@ -246,6 +259,14 @@ impl<'a> CoordinateMapper<'a> {
     ///
     /// Returns None if the genomic position is in an intron.
     /// The transcript must have genomic coordinates set on its exons.
+    ///
+    /// **This is a genome-frame conversion and it does consult the exon
+    /// table** — see the module's "two frames" note, and contrast
+    /// [`Self::cds_to_tx`], which must not. The two answer different questions
+    /// and therefore *must* disagree on a transcript whose alignment carries a
+    /// transcript-coordinate gap; `tests/it/axis_frame_disagreement.rs` pins
+    /// that, so collapsing them onto one implementation fails loudly rather
+    /// than reintroducing #1619.
     pub fn genomic_to_tx(&self, genomic_pos: u64) -> Result<Option<TxPos>, FerroError> {
         // Check if transcript has genomic coordinates
         if !self.transcript.has_genomic_coords() {
@@ -331,6 +352,20 @@ impl<'a> CoordinateMapper<'a> {
     /// Convert transcript position to genomic position
     ///
     /// Returns None if the transcript position is not covered by exons with genomic coords.
+    ///
+    /// **Genome frame**, like [`Self::genomic_to_tx`] and unlike
+    /// [`Self::tx_to_cds`]: the alignment answer of `None` means "this
+    /// transcript base aligns to nothing", which is a fact about the alignment
+    /// and not about the transcript's own numbering — that base still has a
+    /// `c.`/`n.` number.
+    ///
+    /// `None` is not exclusively that answer, though, so do not read it as
+    /// "unaligned" without ruling the other two out: it is also returned for a
+    /// non-positive `tx_pos.base`, and for a `Strand::Unknown` transcript,
+    /// whose bases are all reported unaligned however complete its exon table
+    /// is. A caller using `tx_to_genomic(..).is_some()` as its definition of
+    /// "aligned" — `tests/it/axis_frame_disagreement.rs`'s genome-frame walk
+    /// does — is relying on the transcript having a known strand.
     pub fn tx_to_genomic(&self, tx_pos: &TxPos) -> Result<Option<u64>, FerroError> {
         // Check if transcript has genomic coordinates
         if !self.transcript.has_genomic_coords() {
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -12017,7 +12017,46 @@ impl<P: ReferenceProvider> Normalizer<P> {
         Ok((final_start, final_end, new_edit, warnings))
     }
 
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

PR #1777, opened 2026-08-13
fix(normalize): anchor a 5'-edge cis insertion on -1, not the axis's absent 0

### LINKED ISSUE #1772: Cis-collapse builds a position-0 insertion anchor; n. emits n.0_1insA, which ferro cannot re-parse
The cis-collapse pass builds an insertion anchor by subtracting 1 from an axis coordinate. On every axis that has no zero, stepping down from position 1 produces position **0**, which names no nucleotide. On `n.` this escapes as output ferro cannot re-parse. On `c.` it is spelled `c.?`, colliding with the unknown-position sentinel.

## Measured

Provider: a 35-base transcript `NM_TEST.1`, sequence `GCAAAGCGCGCGATGAAACCCTAAGGCATTTTTAA`, CDS `13..=24` (the fixture already in `src/normalize/merge.rs:11822`, `bounds_transcript_provider`), and `NC_TEST.1` = `TTTTTTTTTAATATATTTTAATAC` (`merge.rs:11813`). Each row is a two-member cis allele passed to `collapse_overlapping_cis_edits` directly, and then the same allele through `Normalizer::normalize`.

| input allele | collapse emits | re-parses? | end-to-end `normalize` | re-parses? |
|---|---|---|---|---|
| `NM_TEST.1:c.[1A>G;1dup]` | `NM_TEST.1:c.?_1insG` | yes, as `c.?` | `NM_TEST.1:c.-1dup` | yes |
| `NM_TEST.1:r.[1a>g;1dup]` | `NM_TEST.1:r.0_1insg` | **no** | `NM_TEST.1:r.-1dup` | yes |
| `NM_TEST.1:n.[1G>A;1dup]` | `NM_TEST.1:n.0_1insA` | **no** | **`NM_TEST.1:n.0_1insA`** | **no** |
| `NC_TEST.1:g.[1T>A;1dup]` | `NC_TEST.1:g.0_1insA` | **no** | `NC_TEST.1:g.1delinsAT` | yes |

Control rows away from the axis origin are correct: `c.[2A>T;2dup]` collapses to `c.2_3insT`, `n.[13A>G;13dup]` to `n.12_13insG`.

**`n.` is the live user-visible failure.** `NM_TEST.1:n.[1G>A;1dup]` normalizes to `NM_TEST.1:n.0_1insA`, and `parse_hgvs` rejects that string: `Unexpected trailing characters: '_1insA'`. ferro emits a description it cannot read back.

## Where it comes from

`src/normalize/merge.rs:606-612`, in `collapse_overlapping_cis_edits` (`merge.rs:378`):

```rust
    let del_start = w_lo + lo as i64;
    let (a_start, a_end) = if hi_ref == lo {
        // Net pure insertion at the boundary: anchor span = 1 nt (start = end+1).
        (del_start, del_start - 1)
    } else {
        (del_start, w_lo + hi_ref as i64 - 1)
    };
```

For a group that nets to a pure insertion at the window's 5' edge, `lo == 0` and `w_lo == 1`, so `a_end = 0`. The anchor is built at `merge.rs:630-636` with a single `region` taken from `body_region(kind)` (`merge.rs:674`), and `build_naedit` (`merge.rs:1357`) names both endpoints through one closure. The per-axis closures name the dense integer verbatim:

```rust
// merge.rs:1236  build_cds_merged
    Region::Cds | Region::FivePrimeUtr => CdsPos::new(b),
// merge.rs:1254  build_tx_merged
    Region::Tx | Region::TxUpstream => TxPos::new(b),
// merge.rs:1267  build_rna_merged
    Region::Rna | Region::FivePrimeUtr => RnaPos::new(b),
```

So `b == 0` becomes `CdsPos { base: 0 }` / `TxPos { base: 0 }` / `RnaPos { base: 0 }`. The arithmetic itself is fine — it is dense window arithmetic, and correct for a dense axis. The defect is the **naming** step: a dense 0 has no name on an axis that skips zero.

`CdsPos::Display` (`src/hgvs/location.rs:282`) renders base 0 as `?`, because `CDS_BASE_UNKNOWN == 0` (`location.rs:151`). That is why the `c.` row reads `c.?_1insG` rather than `c.0_1insG` — ferro emits an *unknown-position* description for a position that is entirely known.

## Why `c.-1` is the correct anchor

`assets/hgvs-nomenclature/docs/background/numbering.md:29-31`:

> - nucleotides upstream (5') of the `ATG`-translation initiation codon (start) are marked with a `-` (hyphen-minus) and numbered `c.-1`, `c.-2`, `c.-3`, etc., going further upstream.
> - there is no nucleotide `c.0`.

The base immediately 5' of `c.1` is `c.-1`. The same step-down already exists in this tree: `Normalizer::cds_to_tx_pos`'s 5'UTR branch (`src/normalize/mod.rs:11742-11753`, the #97 fix) and `tx_to_cds_pos` (`mod.rs:11775-11788`). `merge.rs` states the rule twice more and *reads* it correctly — `simple_cds_pos` returns `None` for base 0 with `// pos.base == 0 — invalid c. position (HGVS skips c.0)` (`merge.rs:1163`), and `place_signed` refuses with `// no `c.0` / `n.0` position exists` (`merge.rs:9905`). Only the builder can create the value its own readers reject.

`respell_at_gap` already produces the correct `c.-1_1` shape for this geometry by converting through transcript space rather than adding to an axis with no zero (`merge.rs:11193-11206`, `cds_axis_position` at `merge.rs:4270-4282`).

## Why this has been invisible

On `c.` and `r.` the broken intermediate is silently repaired downstream. `Normalizer::cds_to_tx_pos` has a `base == 0` arm (`src/normalize/mod.rs:11754-11758`):

```rust
        } else if pos.base == 0 {
            // c.0 is not a valid HGVS position, but historical inputs
            // can land here. Preserve the legacy mapping (treat as the
            // last 5'UTR base, equivalent to c.-1) rather than failing.
            Ok(cds_start.saturating_sub(1))
        } else {
```

`rna_to_tx_pos` carries a byte-identical arm with no comment at all (`mod.rs:8994-8996`). `cds_start - 1` happens to be the correct transcript coordinate for `c.-1`, so the arm converts the malformed `c.0` anchor into the right answer and the final string comes out as `c.-1dup`. The `n.` axis has no such arm, which is exactly why `n.` is the row that escapes.

Measured directly: instrumenting that arm with a `panic!` and running the full suite (10,238 tests) turns exactly one test red — `it::issue_1284_transcript_axis_reparse::no_transcript_axis_two_member_allele_fails_to_reparse` — reached via `normalize_core_canonical -> normalize_allele -> normalize_dispatch -> normalize_core -> normalize_cds -> cds_to_tx_pos`, with `CdsPos { base: 0, offset: None, utr3: false, special: None }` and `cds_start = 13`. That is this defect, and it is the only thing reaching that arm.

That test stays green today and would stay green if the arm were removed, because it asserts only that the output re-parses, not what it is.

## Relationship to the `base == 0` conversion arm

This issue is the **producer**. The arm above is the **consumer**, and retiring it is a separate change needing its own ruling — the two conversion layers currently answer base 0 differently (`CoordinateMapper::cds_to_tx` treats it as `c.1`; `Normalizer::cds_to_tx_pos` treats it as `c.-1`).

Fixing the consumer first is not safe: with the arm replaced by an error, `NM_TEST.1:c.[1A>G;1dup]` normalizes to `NM_TEST.1:c.?_1insG` — the malformed intermediate becomes the user-visible answer — and **no test goes red** (measured: 10,238 passed both with and without the arm). This issue is the prerequisite that removes the arm's remaining reason to exist on this path.

## Scope

The `g.`/`m.` axis hits the same line and also produces a `g.0` anchor, but there is no negative genomic coordinate, so "what is the anchor for an insertion before `g.1`" is a different question with no spec-explicit answer, and the genomic path already recovers to a valid `g.1delinsAT` end-to-end. Leaving it out here deliberately; it deserves its own decision.

## Proposed fix

Name a dense 0 as `-1` on the three signed transcript axes (`c.`, `n.`, `r.`) at the point where the anchor integer is converted into a position, leaving the dense window arithmetic and the `start == end + 1` insertion invariant (`merge.rs:1374-1378`) untouched.

nh13:
Closes #1772.

`collapse_overlapping_cis_edits` derives an insertion anchor as `del_start - 1` in dense window arithmetic (`src/normalize/merge.rs:606-612`). When a cis group nets to a pure insertion at the 5' edge of its window, that is `0` — and the `c.`/`n.`/`r.` axes have no nucleotide `0`. The builders named the dense integer verbatim, so the anchor became `CdsPos { base: 0 }`, which `CdsPos::Display` renders as `c.?` because `CDS_BASE_UNKNOWN` *is* `0` — a fully known position described as an unknown one — and `TxPos`/`RnaPos { base: 0 }`, rendered as the literal `n.0`/`r.0`, which ferro's own parser rejects.

`background/numbering.md:31` says *"there is no nucleotide `c.0`"* and `:29` numbers the bases upstream of the start codon `c.-1`, `c.-2`, `c.-3`. So the position below `1` is `-1`. That is the same step-down `cds_to_tx_pos` and `tx_to_cds_pos` already make for #97, and the shape `respell_at_gap` already produces by converting through transcript space (`merge.rs:11193-11206`).

## Measured, before and after

Fixture is the one already in the file — `bounds_transcript_provider` (`merge.rs:11822`), CDS `13..=24` — plus `bounds_genomic_provider` for the `g.` row.

| input | collapse emits, before | after | end-to-end, before | after |
|---|---|---|---|---|
| `NM_TEST.1:c.[1A>G;1dup]` | `c.?_1insG` | `c.-1_1insG` | `c.-1dup` | `c.-1dup` |
| `NM_TEST.1:r.[1a>g;1dup]` | `r.0_1insg` (unparseable) | `r.-1_1insg` | `r.-1dup` | `r.-1dup` |
| `NM_TEST.1:n.[1G>A;1dup]` | `n.0_1insA` (unparseable) | `n.-1_1insA` | **`n.0_1insA`** (unparseable) | **`n.-1_1insA`** |
| `NM_TEST.1:c.[2A>T;2dup]` | `c.2_3insT` | unchanged | `c.2dup` | `c.2dup` |
| `NM_TEST.1:n.[13A>G;13dup]` | `n.12_13insG` | unchanged | `n.12dup` | `n.12dup` |
| `NC_TEST.1:g.[1T>A;1dup]` | `g.0_1insA` (unparseable) | unchanged, see scope | `g.1delinsAT` | `g.1delinsAT` |

**Exactly one end-to-end output moves: the `n.` row.** The `c.` and `r.` rows do not, because their malformed intermediate was already being silently repaired downstream — see below. Control rows away from the axis origin are untouched, which is the point of the second guard.

## Why the guards assert the intermediate, not the final string

On `c.` and `r.` the broken anchor never reaches the user: the `base == 0` arms in `Normalizer::cds_to_tx_pos` (`src/normalize/mod.rs:11754-11758`) and `rna_to_tx_pos` (`mod.rs:8994-8996`) map `0` to `cds_start - 1`, which happens to be the correct transcript coordinate for `-1`. So an end-to-end assertion on `c.-1dup` passes with the defect fully present, and would be a change detector rather than a guard.

`a_five_prime_edge_insertion_anchors_below_one_not_on_zero` therefore asserts what `collapse_overlapping_cis_edits` itself builds, on all three axes, plus that each collapsed member re-parses — the user-visible half, which `n.` and `r.` both failed. Sabotage check: reverting `name_on_zeroless_axis` to the identity turns it red with `left: ["NM_TEST.1:c.?_1insG"]` / `right: ["NM_TEST.1:c.-1_1insG"]`, while the interior-anchor control stays green.

## This is the prerequisite PR 10 was missing

**This PR deliberately does not touch the `base == 0` arm in `cds_to_tx_pos`, or its siblings in `rna_to_tx_pos` and `cds_pos_to_tx_boundary.`** Retiring them is a separate change: the two conversion layers currently answer base 0 differently (`CoordinateMapper::cds_to_tx` treats it as `c.1`, `Normalizer::cds_to_tx_pos` as `c.-1`), which needs its own ruling and its own disclosure.

What this PR does is remove the arm's remaining reason to exist on this path. Measured with a `panic!` in that arm over the full `--lib --test it` suite:

- on `origin/main`: exactly **one** test reaches it, `issue_1284_transcript_axis_reparse::no_transcript_axis_two_member_allele_fails_to_reparse`, with `CdsPos { base: 0, offset: None, utr3: false, special: None }` and `cds_start = 13` — i.e. this defect, and nothing else;
- with this fix: **zero** hits, 10,240 passed.

Fixing the consumer first would not have been safe. With the arm replaced by an error and this producer unfixed, `NM_TEST.1:c.[1A>G;1dup]` normalizes to `NM_TEST.1:c.?_1insG` — the malformed intermediate becomes the user-visible answer — and **no test goes red** (10,238 passed either way), because the one test that reaches the arm asserts only that the output re-parses, and `c.?_1insG` does.

## Scope

The `g.`/`m.` axis hits the same two lines and also builds a `g.0` anchor, and is deliberately left alone: there is no negative genomic coordinate, so an insertion 5' of `g.1` has no two-position anchor and needs its own answer rather than this one. Its end-to-end output already recovers to a valid `g.1delinsAT`, so nothing is emitted broken today. Noted in #1772.

The dense window arithmetic and the `start == end + 1` insertion invariant (`merge.rs:1374-1378`) are untouched — only the step that names a dense integer as a position changes, which is where the axis is known.

## Verification

Real exit codes, each captured immediately after its command rather than read through a pipe.

```
cargo fmt --all --check                              EXIT=0
cargo clippy --features dev --lib -- -D warnings     EXIT=0
cargo clippy --features python --lib -- -D warnings  EXIT=0
cargo ta                                             EXIT=0   10431 passed, 152 skipped
```

Four seam oracles, same command and same selection on both sides, `origin/main` as the baseline:

```
```

**17 failures before, 17 after — identical sets, zero new and zero fixed.** All 17 are inherited from `origin/main` (`defect_non_idempotent_outputs`, `issue_1487_canonical_window_overflow`, `spec_conformance_axis`, `spec_corpus_regressions`, `stranded_identity_member`); this change neither adds to nor clears that residue.

Representation-Change: 1 shape moves. A cis allele that nets to a pure insertion at the 5' edge of the window on the `n.` axis respells from `n.0_1insA` to `n.-1_1insA` — measured on `NM_TEST.1:n.[1G>A;1dup]`. The previous output names a position that does not exist and `parse_hgvs` rejects it, so no consumer can be holding it as a parsed value; this replaces an unreadable description with a readable one. The `c.` and `r.` equivalents do **not** move (`c.-1dup` and `r.-1dup` on both sides, measured), because their malformed intermediate was already repaired downstream by the `base == 0` conversion arms. No genomic-axis row moves, and no row away from the axis origin moves.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -1276,9 +1276,69 @@ fn build_genome_merged(template: &GenomeVariant, merged: Anchor) -> GenomeVarian
     }
 }
 
+/// Name a cis-collapse anchor coordinate on a signed transcript axis
+/// (`c.` / `n.` / `r.`), where the integer `0` names no nucleotide.
+///
+/// The collapse works in dense window arithmetic — `del_start - 1` for a group
+/// that nets to a pure insertion at the window's 5' edge — so the coordinate
+/// immediately 5' of `1` arrives here as `0`. The `c.`/`n.`/`r.` axes skip
+/// zero: `background/numbering.md:31`, *"there is no nucleotide `c.0`"*, and
+/// `:29` numbers the bases upstream of the start codon `c.-1`, `c.-2`, `c.-3`.
+/// So the position below `1` is named `-1`, which is the same step-down
+/// `Normalizer::cds_to_tx_pos` and `tx_to_cds_pos` make for issue #97.
+///
+/// Only `0` is remapped. The collapse restricts a group to the positive body
+/// region (issue #920) and refuses a window starting below `1`, so a body
+/// anchor endpoint is never negative on entry, and `0` arises only from the
+/// 5'-edge insertion case above.
+///
+/// Without this the builders emit `CdsPos { base: 0 }` — rendered `c.?` by
+/// `CdsPos::Display`, since `CDS_BASE_UNKNOWN` *is* `0`, so a fully known
+/// position is described as an unknown one — and `TxPos`/`RnaPos { base: 0 }`,
+/// rendered as the literal `n.0`/`r.0`, which ferro's own parser rejects.
+/// Issue #1772.
+///
+/// Deliberately **not** applied to the genomic axis: `g.`/`m.` have no negative
+/// coordinate, so an insertion 5' of `g.1` has no two-position anchor and needs
+/// its own answer rather than this one.
+///
+/// # The citation above scopes `c.`, and the three axes are not equally served
+///
+/// `:29` and `:31` sit under `numbering.md:28`'s "untranslated region (UTR)"
+/// bullet inside *"### coding DNA reference sequences"*, so on its own the
+/// clause licenses `-1` on the `c.` axis only. Read what each axis inherits
+/// before citing it here:
+///
+/// - **`c.`** — governed directly. `c.-1` is a conformant 5'UTR position.
+/// - **`r.`** — `:58` gives an RNA reference "that of the associated **coding
+///   or non-coding** DNA reference sequence", so `r.-1` is conformant against a
+///   coding RNA reference and is not against a non-coding one. This builder,
+///   like the parser (`src/hgvs/noncoding_zones.rs`), cannot tell which it has.
+/// - **`n.`** — **not** governed by `:31`. The non-coding DNA axis is `:50`-`:54`,
+///   which grants `n.1..n.N` plus intronic offsets and nothing else, and `:54`
+///   forbids naming a nucleotide beyond a transcript's boundaries using that
+///   transcript. PR #1751 (merged) implements exactly that: `n.-N` is refused at
+///   **strict** parse (`W4008`) and accepted only by lenient, silent and the
+///   bare [`crate::parse_hgvs`] entry — the last kept open for five real ClinVar
+///   rows.
+///
+/// So on `n.` this is a strict improvement and not a resolution: it replaces
+/// `n.0`, which every mode rejects and which names a nucleotide no axis has,
+/// with `n.-1`, which strict still refuses. The residue is the same open
+/// question as the genomic axis's — what a cis group nets to a pure insertion 5'
+/// of the first base of a transcript should be called at all — and it belongs
+/// with #1772 rather than to this rename.
+fn name_on_zeroless_axis(base: i64) -> i64 {
+    if base == 0 {
+        -1
+    } else {
+        base
+    }
+}
+
 fn build_cds_merged(template: &CdsVariant, merged: Anchor) -> CdsVariant {
     let (location, edit) = build_naedit(merged, |region, b| match region {
-        Region::Cds | Region::FivePrimeUtr => CdsPos::new(b),
+        Region::Cds | Region::FivePrimeUtr => CdsPos::new(name_on_zeroless_axis(b)),
         Region::ThreePrimeUtr => CdsPos {
             base: b,
             offset: None,
@@ -1296,7 +1356,7 @@ fn build_cds_merged(template: &CdsVariant, merged: Anchor) -> CdsVariant {
 
 fn build_tx_merged(template: &TxVariant, merged: Anchor) -> TxVariant {
     let (location, edit) = build_naedit(merged, |region, b| match region {
-        Region::Tx | Region::TxUpstream => TxPos::new(b),
+        Region::Tx | Region::TxUpstream => TxPos::new(name_on_zeroless_axis(b)),
         Region::TxDownstream => TxPos::downstream(b),
         _ => unreachable!("non-n. region {:?} on TxVariant", region),
     });
@@ -1309,7 +1369,7 @@ fn build_tx_merged(template: &TxVariant, merged: Anchor) -> TxVariant {
 
 fn build_rna_merged(template: &RnaVariant, merged: Anchor) -> RnaVariant {
     let (location, edit) = build_naedit(merged, |region, b| match region {
-        Region::Rna | Region::FivePrimeUtr => RnaPos::new(b),
+        Region::Rna | Region::FivePrimeUtr => RnaPos::new(name_on_zeroless_axis(b)),
         Region::ThreePrimeUtr => RnaPos::utr3(b),
         _ => unreachable!("non-r. region {:?} on RnaVariant", region),
     });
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1747, opened 2026-08-12
fix(convert): refuse a CDS position that names no base instead of computing on it

nh13:
Closes #1745. Closes #1746. Closes #1690.

**Unstacked 2026-08-13 — this PR now targets `main` directly.** It was stacked on #1735 (`fix/issue-1619-flat-transcript-frame`), which has merged (squash `481de7c1`). The branch was rebased with `git rebase --onto origin/main 50507a60` — `50507a60` being the #1735 commit this PR was actually built on, which is neither the squash on `main` nor #1735's final head. `origin/main..HEAD` is now this PR's single commit; the diff is unchanged at 2 files, +307/-42. The verification figures below were measured against #1735's head and still describe this diff, but they are pre-rebase numbers — the post-rebase gate is `10335 tests run: 10335 passed, 150 skipped` with `cargo fmt --check` and `cargo clippy --all-features --all-targets -- -D warnings` clean.

`CoordinateMapper` performed unguarded arithmetic on three values it had never established were coordinates.

## A `cds_start` of 0 (#1745)

`get_amino_acid_at` computed `cds_start as usize - 1 + (position as usize - 1) * 3`. `Transcript` stores `cds_start` exactly as its provider supplied it and validates nothing (`reference/validate.rs`'s CDS check is out-of-band and report-only), so `Some(0)` is representable — a 1-based coordinate naming no transcript base.

It panicked in **both** build profiles, and the release failure moved: debug aborted on the subtraction, release wrapped to `usize::MAX` and aborted on the string slice, reporting an out-of-bounds index rather than the bad metadata that caused it. Reached through the public `cds_to_protein`, which is what the Python `c_to_p` calls. `Transcript::utr5_length` already defends the same field with `saturating_sub`.

## `c.?` (#1746)

`cds_to_tx` answered it `Ok(cds_start)` — measured byte-identical to its answer for `c.1` — because `CDS_BASE_UNKNOWN` is the integer 0 and 0 took the `base < 1` 5'UTR branch with a zero displacement.

The doctrine is already stated on `CdsPos::has_unknown_offset` for an unknown *offset*, and `project/projector.rs` rejects `is_unknown()` at two sites **before** calling here — `:4643-4647` naming this mechanism exactly (*"without this guard `base <= 0` would misclassify it as 5'UTR and return `Ok`"*). The guard now lives where those two were working around it.

[...]

PR #1778, opened 2026-08-13
refactor(hgvs): one definition of the unknown-offset sentinel pair

nh13:
Follow-up to #1762, completing the same finding. #1762 removed the *third* spelling of the unknown-offset sentinel pair — `CdsPos::Display` no longer compares bare `i64::MAX` / `i64::MIN` literals, and `TxPos`/`RnaPos` render through the shared `unknown_offset_marker` helper. What it left behind is the residue: the pair was still **defined twice**.

```rust
src/hgvs/parser/position.rs  pub const OFFSET_UNKNOWN_POSITIVE: i64 = i64::MAX;
src/hgvs/parser/position.rs  pub const OFFSET_UNKNOWN_NEGATIVE: i64 = i64::MIN;
src/hgvs/location.rs         pub const GENOME_OFFSET_UNKNOWN_POSITIVE: i64 = i64::MAX;
src/hgvs/location.rs         pub const GENOME_OFFSET_UNKNOWN_NEGATIVE: i64 = i64::MIN;
```

Two independent definitions of one value pair, with nothing tying them together. Change either and the other is silently left behind: the parser stores one value while every `Display` — all of which key off the parser's pair — stops recognising it and prints a raw 19-digit integer where a `?` belongs. That is exactly the leak #1762 fixed for `TxPos`/`RnaPos`, still armed by a different mechanism.

## The parser's name is canonical

`OFFSET_UNKNOWN_{POSITIVE,NEGATIVE}` wins on three grounds:

- **It is not genome-specific, and the value isn't either.** All four axes store the same two values — `parse_genome_pos`, `parse_cds_pos`, `parse_tx_pos` and `parse_rna_pos` all resolve `+?` / `-?` through the one shared `parse_offset`. A `GENOME_`-prefixed name for a value the `c.`, `n.` and `r.` axes carry equally is a misnomer, and the misnomer is what made a second declaration look reasonable in the first place.
- **The values originate there.** `parse_offset` is the sole producer — the fast path defers `+?` / `-?` to the generic parser rather than synthesising its own (`scan_cds_sub_position`), so there is one point of entry into the AST.
- **It is already the name in use.** Before this change `GENOME_OFFSET_UNKNOWN_*` had exactly one consumer outside its own declaration; the parser's pair is used across `location.rs`, `parser/position.rs`, `variant.rs`, `data/mapping.rs`, `normalize/mod.rs` and the tests.

## The old name is kept, as a re-export

`GENOME_OFFSET_UNKNOWN_*` is `pub` in a published crate, so deleting it is a breaking change for external consumers. It stays, as a re-export of the canonical pair:

```rust
pub use crate::hgvs::parser::position::{
    OFFSET_UNKNOWN_NEGATIVE as GENOME_OFFSET_UNKNOWN_NEGATIVE,
    OFFSET_UNKNOWN_POSITIVE as GENOME_OFFSET_UNKNOWN_POSITIVE,
};
```

A re-export rather than `pub const GENOME_… : i64 = OFFSET_…;` because it makes divergence *unrepresentable* rather than merely unlikely — it is literally the same item, not a second item that currently agrees. It also matches how the crate re-exports elsewhere (`lib.rs` is all `pub use`). Every path that resolved `ferro_hgvs::hgvs::location::GENOME_OFFSET_UNKNOWN_POSITIVE` still resolves.

There is **one** literal `i64::MAX` / `i64::MIN` definition of the pair left in the crate, in `parser/position.rs`.

The Python surface is unaffected: neither name appears in `src/python.rs` or `python/ferro_hgvs/__init__.pyi`, so there is no second compatibility surface to preserve.

## The one comparison site

`variant.rs::certain_offset` tested four names for two values. It now tests two, in the constant-pattern form the rest of the file already uses. This is a no-op — the two dropped arms compared the same two values a second time.

## The divergence guard

`tests/it/one_unknown_offset_sentinel_definition.rs` is the tripwire for someone re-introducing a second declaration.

**Nothing in it compares a constant to a constant.** That assertion would be tautological today — after the collapse the two names *are* one item, so it holds by definition, and it would keep holding if the collapse were undone with two literals that happen to match. Instead every assertion compares a **named constant against a value the parser actually produced** from the text `+?` / `-?`, or against what the renderer actually prints:

- `parse_genome_pos("100+?")` / `("100-?")` must yield exactly `GENOME_OFFSET_UNKNOWN_{POSITIVE,NEGATIVE}`. This is the load-bearing one: the parser is driven by the *canonical* constant and knows nothing of the alias, so a re-declared alias that drifts stops matching the name this asserts it against.
- The same question on `c.`, `n.` and `r.` under the canonical name — the fact that pins "one pair, every axis".
- A position **built from** each named constant must render as `+?` / `-?`, closing the parser → constant → renderer loop from the other end. A drifted alias fails here by printing its 19-digit value, which is the symptom the sentinel exists to prevent.
- A whole-description round trip on the genome axis (`NC_000001.11:g.100+?del`), the axis that carries the alias and the one #1762's round-trip test did not cover.
- A control: an ordinary offset one step inside each sentinel still renders numerically, so the sentinel band is pinned at exactly two values and the assertions above cannot pass by the renderer printing `?` for everything.

## Stacking — this PR is based on #1762, not on `main`

This branch is based on `fix/downstream-flag-and-offset-sentinel` (#1762), at **`d8e3cd0bcb27d312ac0efce46bbb03988d688292`**, and its base branch is set accordingly.

Because the repo squash-merges, once #1762 lands this branch must be re-parented onto #1762's **squash commit** and retargeted to `main` before it can enter the merge queue:

```bash
git rebase --onto <#1762-squash-commit-on-main> d8e3cd0bcb27d312ac0efce46bbb03988d688292 \
  refactor/one-unknown-offset-sentinel-definition
gh pr edit <this-PR> --base main
```

A plain `git rebase origin/main` will **not** work — it replays #1762's own commit on top of the squash of that same commit and conflicts. (#1747 is sitting `CONFLICTING` right now for exactly this reason after its base merged.)

Representation-Change: none. Two constants holding equal values are collapsed into one definition plus a re-export; no value and no rendered string changes. The only behavioural code touched is `certain_offset`, where two of four comparison arms tested the same two values a second time and were dropped. The rest is a new test and documentation.

### DIFF
--- a/src/hgvs/location.rs
+++ b/src/hgvs/location.rs
@@ -45,7 +45,15 @@ pub struct GenomePos {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub special: Option<SpecialPosition>,
     /// Optional offset from the base position (e.g., -? for uncertain upstream, +? for uncertain downstream)
-    /// Uses i64::MIN for uncertain negative offset (-?) and i64::MAX for uncertain positive offset (+?)
+    ///
+    /// The uncertain forms are stored in band, as
+    /// [`OFFSET_UNKNOWN_POSITIVE`] (`+?`) and [`OFFSET_UNKNOWN_NEGATIVE`]
+    /// (`-?`). Name the constants rather than their current values: they are
+    /// defined in exactly one place, and a doc that respells them as literals
+    /// is a second declaration in prose.
+    ///
+    /// [`OFFSET_UNKNOWN_POSITIVE`]: crate::hgvs::parser::position::OFFSET_UNKNOWN_POSITIVE
+    /// [`OFFSET_UNKNOWN_NEGATIVE`]: crate::hgvs::parser::position::OFFSET_UNKNOWN_NEGATIVE
     #[serde(skip_serializing_if = "Option::is_none")]
     pub offset: Option<i64>,
 }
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

PR #1779, opened 2026-08-13
test(convert): assert the cdot 0-based tx basis without a prepared reference

nh13:
`test_cdot_tx_coordinates_are_0_based` in `src/convert/mapper.rs` guarded the convention that internal cdot transcript coordinates are 0-based half-open — and asserted nothing on essentially every run, including CI.

```rust
let cdot_path = PathBuf::from("benchmark-output/cdot/cdot-0.2.32.refseq.GRCh38.json");
if !cdot_path.exists() {
    // Skip test if cdot file is not available
    return;
}
```

`benchmark-output/` is an operator-created symlink to a prepared reference, absent in a default checkout and in CI. A bare `return` with no message and no counter meant the test reported green having executed zero assertions. This is a follow-up to the coordinate-basis work in #742, and the drift it is supposed to catch is exactly the one found in `benchmark/cache.rs` (campaign finding M8) — a 0-based/1-based confusion that is a guaranteed panic on the first exon. The test that should have caught that is the one that silently was not running.

A census of all 69 `benchmark-output` references across 20 Rust files confirms this was the **only** silently-skipping gate left in the tree. Classifying every reference: 30 are path or config construction (of which ~25 are the test skip gates in `tests/it/`, the rest CLI and benchmark-binary defaults), 29 sit inside the text of a skip message, and 10 are doc comments. An automated scan for the actual defect shape — a path-existence check reaching a bare `return` with no print in between — flags exactly three candidates, and two are false positives where `benchmark-output` appears inside the skip message itself (`tests/it/multi_member_cis_axis.rs`, and this PR's own new cross-check). The one real hit was `test_cdot_tx_coordinates_are_0_based`. No other site needed changing, so this PR does not widen beyond the one test.

## What replaces it

**`cdot_tx_coordinates_are_0_based_half_open`** — always runs, no reference required. It builds a two-exon cdot document spelled in cdot's own raw conventions (genomic 0-based half-open, cDNA 1-based inclusive) and pushes it through `CdotMapper::from_reader_with_build`, the same `from_genome_build` conversion the real path uses. It follows the fixture idiom already in the tree (`src/project/projector.rs`'s `multi_build_cdot_json`, `src/data/cdot.rs`'s `raw_genome_build_with_tag`) rather than inventing one.

It asserts raw tx `1..100` converts to `[0, 100)` and `101..150` to `[100, 150)`; that adjoining exons meet at one shared value (the half-open half of the claim, which a single-exon fixture cannot express); and that `cds_start` is the 0-based tx offset of `c.1`.

Two properties of the old test argued against merely adding an `eprintln!` before the `return`. It hardcoded `cdot-0.2.32` while cdot is now at 0.2.33, so regenerating the reference would have quietly returned it to never running; and it pinned `cds_start == Some(127)` for `NM_003742.4`, live external data upstream can revise. Both are fragilities a fixture does not have.

**`cdot_tx_basis_cross_checked_against_real_cdot`** — the real-data path, kept rather than deleted, because a fixture can only encode the conversion its author believed in. It discovers the cdot file by pattern instead of pinning a release, asserts a structural invariant over a sample of real transcripts instead of one accession's `cds_start`, refuses a zero denominator, and prints an actionable message naming the symlink remedy instead of returning silently.

Its doc comment records what it does **not** cover, measured rather than assumed: `CdotMapper::load` prefers the sibling `.rkyv` cache, so on a warm prepared reference it deserializes coordinates converted at prepare time rather than re-running the conversion. Under the sabotage below it stayed **green** while the fixture test failed. It therefore guards the prepared artifact's stored basis, not the conversion; forcing the JSON arm would mean parsing 489 MB in a unit test, which is not worth it when the fixture already covers the conversion everywhere. Saying so in the doc comment matters — a test whose comment overclaims is how this defect started.

Exon-to-exon tx contiguity is asserted only in the fixture, not against real cdot: real cdot carries genuine transcript-coordinate holes (58 of 474,818 GRCh38 multi-exon builds), so contiguity is a property of the fixture's geometry, not of cdot at large. `tx_start == 0` on the first exon is universal and is the discriminating check anyway.

## Verification

The fixture's faithfulness to real cdot was checked against the prepared reference rather than assumed: real cdot spells raw exon tx bounds 1-based inclusive (adjacent exons read `…4521]`, `[4522…`) with the first exon's raw `tx_start` at 1 — exactly the spelling the fixture uses. The real-data test was also run against a live prepared reference (9.0 s, 200 chr1 transcripts) rather than only in its skip path.

Sabotage check — with `from_genome_build`'s `tx_start` conversion inverted to leave bounds 1-based:

```
FAIL cdot_tx_coordinates_are_0_based_half_open
assertion `left == right` failed: exon 1 raw tx 1..100 (1-based inclusive) should
convert to 0-based half-open [0, 100), got [1, 100)
  left: (1, 100)
 right: (0, 100)
```

The production file was restored from a copy kept outside the worktree and `git status` confirms it byte-clean; the diff is confined to the test module.

## Noted, not fixed

`from_genome_build`'s genomic-CDS fallback applies a further `+ 1` to the mapped CDS end. Whether that is correct depends on whether cdot's raw genomic `cds_end` is inclusive or exclusive — the `RawGenomeBuild::cds_end` doc says exclusive, while the arithmetic implies inclusive. The new test deliberately does not assert `cds_end`, so it cannot cement whichever reading is wrong. Worth a separate look; out of scope here.

Representation-Change: none. Test-only — the diff touches nothing outside `mod intronic_debug_tests` in `src/convert/mapper.rs`, no production code path is altered, and no normalized output can move.

### DIFF
--- a/src/convert/mapper.rs
+++ b/src/convert/mapper.rs
@@ -1333,41 +1333,343 @@ mod intronic_debug_tests {
         );
     }
 
-    /// Test that cdot tx coordinates are stored 0-based half-open.
+    /// A minimal cdot document spelled in cdot's own RAW conventions: genomic
+    /// bounds 0-based half-open (`alt_start` inclusive, `alt_end` exclusive)
+    /// and cDNA bounds 1-based **inclusive**.
     ///
-    /// On load, `from_genome_build` converts cdot's raw 1-based-inclusive tx bounds
-    /// into the engine's HGVS convention: tx 0-based half-open (the first exon starts
-    /// at position 0, not 1). This test verifies that conversion (#742).
+    /// | exon | raw genomic    | raw tx (1-based incl) | expected internal tx |
+    /// |------|----------------|-----------------------|----------------------|
+    /// | 1    | `[5000, 5100)` | `1..100`              | `[0, 100)`           |
+    /// | 2    | `[6000, 6050)` | `101..150`            | `[100, 150)`         |
+    ///
+    /// Two exons deliberately. The second is what makes the *half-open* half of
+    /// the claim testable at all: under 0-based half-open storage exon 2's
+    /// `tx_start` equals exon 1's `tx_end` exactly, where 1-based-inclusive
+    /// storage would leave it one higher.
+    ///
+    /// `start_codon`/`stop_codon` are deliberately ABSENT. `from_genome_build`
+    /// copies those through verbatim when they are present, so a fixture
+    /// carrying them would bypass the conversion under test and assert nothing
+    /// about the coordinate basis. Omitting them forces the genomic-CDS
+    /// fallback — the arm that has to reconcile raw genomic bounds against the
+    /// already-converted exon table, which is where #742's off-by-one lived.
+    const RAW_CDOT_BASIS_FIXTURE: &str = r#"{
+        "transcripts": {
+            "NM_BASIS_TEST.1": {
+                "gene_name": "BASISTEST",
+                "genome_builds": {
+                    "GRCh38": {
+                        "contig": "NC_000001.11",
+                        "strand": "+",
+                        "cds_start": 5010,
+                        "cds_end": 6041,
+                        "exons": [
+                            [5000, 5100, 1, 1, 100, "M100"],
+                            [6000, 6050, 2, 101, 150, "M50"]
+                        ]
+                    }
+                }
+            }
+        }
+    }"#;
+
+    /// cdot tx coordinates are stored 0-based half-open (#742).
+    ///
+    /// On load, `from_genome_build` converts cdot's raw 1-based-inclusive tx
+    /// bounds into the engine's HGVS convention: tx 0-based half-open, the
+    /// first exon starting at 0 rather than 1. Violating that convention is a
+    /// guaranteed panic on the first exon, so the guard must not be able to
+    /// stop running quietly.
+    ///
+    /// This runs on every machine, against a programmatically-built fixture.
+    /// Its predecessor asserted only when an operator-created
+    /// `benchmark-output/` symlink was present, and returned silently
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
The question this PR deliberately left open is now settled, in case you want to close the gap here rather than leave the comment standing.

`cdot_tx_coordinates_are_0_based_half_open` declines to assert `cds_end` because "the struct doc says exclusive while the arithmetic implies inclusive … pinning today's value here would cement whichever reading is wrong". Measured against real cdot: **the doc comment is right and the arithmetic is wrong** — #1785, fixed in #1786.

`genomic_to_tx_pos` maps a genomic *position*, not a half-open bound. The first coding base is `g_start + 1`, but the last coding base is `g_end` — 0-based exclusive and 1-based inclusive are the same number — so only one of the two endpoints takes a `+ 1`, and the code gave both one. Over cdot-0.2.32 GRCh38 the current arithmetic produces a CDS length that is not a whole number of codons for 354,187 of 355,590 RefSeq coding builds and agrees with the record's own `start_codon`/`stop_codon` for 0 of 266,774 Ensembl builds; mapping `g_end` agrees for all 266,774.

For this fixture specifically: raw CDS `[5010, 6041)` against an exon opening at raw genomic 6000 with `tx_start` 101 gives a last coding base at tx 140, so **`cds_end` should be 141** — it is 142 today. Once #1786 lands, `assert_eq!(tx.cds_end, Some(141))` is a correct assertion here and the "deliberately not asserted" comment can go.

Not duplicating it in #1786 — that PR keeps to `src/data/cdot.rs` and pins the answer with its own tests there, since this file is yours.

PR #1781, opened 2026-08-13
fix(project): refuse an `n.*N` endpoint on the genome-pivot path

### LINKED ISSUE #1768: `project_variant` silently drops the `*` marker on a downstream `n.` input via the genome-pivot path
**What.** On the genome-pivot projection path, a `TxVariant` whose endpoint carries `downstream == true` (`n.*N`) is neither refused nor preserved: the marker is dropped and the position is reported as a plain `n.<base>`. `n.*5` is silently reinterpreted as `n.5`.

**Path.** `project_variant` → genome pivot. The `HgvsVariant::Tx` input arm reshapes `TxPos` → `CdsPos` at `src/project/projector.rs:2699-2703` carrying `utr3: p.downstream`. After genomic mapping, the transcript axis is rebuilt for a non-coding transcript at `src/project/projector.rs:5669-5671` with `TxPos::new(cds_start.base)`, and again for the cache-key variant at `:5705` — `TxPos::new` hardcodes `downstream: false`, so the marker cannot survive.

**Contrast.** The direct non-coding path *does* guard: `project_noncoding_direct` returns `UnsupportedProjection` at `src/project/projector.rs:4896-4902` when either endpoint is downstream. The genome-pivot path has no counterpart.

**Pre-existing.** Both the `to_cds` reshape and the `TxPos::new` rebuild predate #1751, which touches only the parser and adds no code under `src/project/`. Before #1751 this was reachable from a *parsed* `n.*N`; after it, only from a caller-constructed AST via the public `TxPos::downstream` constructor. #1751 narrows the reachable surface — it neither introduces nor fixes the behaviour.

**Correct behaviour.** Refuse, mirroring `project_noncoding_direct`: return `FerroError::UnsupportedProjection` when either endpoint of a `Tx` input is downstream, rather than silently renumbering it.

**Status.** Found by code audit during #1751 review; verified by inspection, **no reproducer written**. A reproducer needs a `MockProvider` non-coding transcript plus a constructed `TxPos::downstream` AST, asserting on `project_variant`'s `.noncoding` output — confirm the exact observable output before fixing.

nh13:
An `n.` input whose endpoint carries `downstream == true` (`n.*N`) was neither refused nor preserved on the genome-pivot path. `project_to_genomic_nc` reshapes `TxPos` → `CdsPos` folding `downstream` into `utr3`, but the two markers do not mean the same thing:

- `TxPos::downstream` — "N bases after the **transcript end**"
- `CdsPos::utr3` — "N bases after the **stop codon**"

On any transcript with a 3'UTR those are different positions. The transcript axis is then rebuilt from the mapped genome coordinate, and both `TxPos::new` and the CDS→`n.` derivation can only spell a plain `n.<base>`, so nothing downstream can repair it.

## Measured, before the fix

On a 15-base transcript whose CDS ends at tx 12:

| input | came back as |
|---|---|
| `NC_000001.11(NM_UTR.1):n.*1del` | `g.1012del` / `c.*1del` / **`n.13del`** |

Reinterpreted onto the CDS axis, and the marker gone from the axis it was authored on.

On a **non-coding** transcript it is exact rather than merely wrong: `map_cnr_position_to_genome`'s non-coding branch never reads `utr3` at all, so `n.*5del` and `n.5del` both produced `g.1004del`. Two different descriptions, one answer.

## The fix

Refuse, mirroring the guard `project_noncoding_direct` already applies on the direct path, so the two paths agree rather than one silently renumbering what the other declines. Same error type, same vocabulary.

Scoped to `n.` inputs deliberately: on a coding transcript the `r.` axis is the `c.` axis, so `r.*N` genuinely means `c.*N` and still projects. `project_to_genomic_still_accepts_a_downstream_r_position` pins that, so the refusal cannot quietly widen.

## Tests

Three, all in `src/project/projector.rs`:

- `project_to_genomic_refuses_downstream_n_position_on_a_coding_transcript`
- `project_to_genomic_refuses_downstream_n_position_on_a_noncoding_transcript`
- `project_to_genomic_still_accepts_a_downstream_r_position`

The two refusal tests assert on the reason text, not merely that an error occurred, so a different failure cannot pass them by accident.

**Sabotage-checked.** With the guard disabled, the two refusal tests fail and name the exact wrong outputs — `g.1012del` on the coding transcript and `g.1004del` on the non-coding one, matching the before-measurement above digit for digit — while `project_to_genomic_still_accepts_a_downstream_r_position` continues to pass, which is what shows the pair are not merely asserting "some error occurred". Restoring the guard returns all three to green (`3 passed`), and the restored file is byte-identical to the committed one by md5.

## Relationship to #1751

#1751 refuses `n.*N` at parse in every mode. It is open, not merged, so on current `main` the shape still parses and this path is reachable from a parsed description today. Once #1751 lands, the shape becomes reachable only by direct construction through the public `TxPos::downstream` constructor — which is exactly the caller class #1751's own public-API note describes. This guard is correct either way, and is what stops a constructed AST being silently renumbered rather than declined.

Closes #1768

Representation-Change: outputs move for two shapes. An `n.*N` input on the genome-pivot path previously produced a projection — measured as `g.1012del` / `c.*1del` / `n.13del` on a coding transcript, and a `g.` coordinate identical to the corresponding `n.<base>` on a non-coding one — and now returns `UnsupportedProjection`. The same now holds for an `r.*N` input on a **non-coding** transcript, which previously returned the same genomic coordinate as the corresponding `r.<base>` (measured: both `NC_000001.11(NCGENE):g.1004del`) and now returns `UnsupportedProjection`. **0 corpus rows move for `n.*N`**: `n.*N` occurs 0 times in the four committed corpora (`clinvar_hgvs_500k`, `clinvar_hgvs_unique`, `cmrg_genes_exhaustive`, `paraphase_genes_exhaustive`, 103,762 `n.`-axis rows), and that zero is real rather than structural — the same scan finds the sibling `n.-N` five times. The corpus incidence of `r.*N` was not separately measured, so that zero is claimed for `n.*N` only. No stored string changes; what changes is that a description ferro could not answer correctly is now declined instead of answered wrongly.

### DIFF
--- a/src/project/projector.rs
+++ b/src/project/projector.rs
@@ -2696,6 +2696,42 @@ impl<P: ReferenceProvider + Clone> VariantProjector<P> {
             HgvsVariant::Tx(v) => {
                 let s = resolve_uncertain_boundary(&v.loc_edit.location.start, "n.", "start")?;
                 let e = resolve_uncertain_boundary(&v.loc_edit.location.end, "n.", "end")?;
+                // A 3'-downstream (`n.*N`) endpoint is refused here, as
+                // `project_noncoding_direct` already refuses one, because the pivot
+                // below can neither honour it nor carry it (#1768).
+                //
+                // The two guards are NOT identical and must not be described as
+                // such: `project_noncoding_direct` reshapes `r.` into `TxPos`
+                // before its check, so it refuses `*N` on both axes
+                // unconditionally. Here the axes part company — `r.*N` on a
+                // CODING transcript is `c.*N` and projects correctly — so the
+                // `r.` half is transcript-dependent and is guarded further
+                // down, once `is_coding` is known.
+                //
+                // `TxPos::downstream` states `*N` as "N bases after the transcript
+                // end", but the reshape immediately under this guard folds it into
+                // `CdsPos.utr3`, which means "N bases after the STOP CODON" — two
+                // different positions on any transcript with a 3'UTR. Measured on a
+                // 15-base transcript whose CDS ends at tx 12,
+                // `NC_000001.11(NM_UTR.1):n.*1del` came back as `c.*1del` /
+                // `n.13del`: reinterpreted onto the CDS axis, and the marker gone
+                // from the `n.` axis it was written on. On a NON-coding transcript
+                // the same input is worse — `map_cnr_position_to_genome`'s
+                // non-coding branch never reads `utr3` at all, so `n.*5` maps to
+                // exactly the genome position `n.5` does.
+                //
+                // Nothing downstream can repair either case: the transcript axis is
+                // rebuilt from the mapped genome coordinate, and both `TxPos::new`
+                // and the CDS→`n.` derivation emit a plain `n.<base>`. So refuse,
+                // rather than renumber silently.
+                if s.downstream || e.downstream {
+                    return Err(FerroError::UnsupportedProjection {
+                        reason: "cannot project 3'-downstream (`*N`) positions on an n. input: \
+                                 the genome pivot has no transcript-axis representation for \
+                                 them and would silently renumber the position"
+                            .to_string(),
+                    });
+                }
                 let to_cds = |p: TxPos| CdsPos {
                     base: p.base,
                     offset: p.offset,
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1782, opened 2026-08-13
perf(test): parallelize the three remaining slow censuses, and cut the per-row cost of the two biggest

nh13:
Representation-Change: none. Test harness only; no watched directory is touched.

Follow-up to #1775, which cleared every `cis_confluence_*` test off the SLOW list. This takes the five that were left. **Draft, because CI is doing both the measurement and the verification** — see the last section.

## The five, and what each needed

Measured on run 31673634192 (#1775's merge-queue run):

| test | wall | job | what was wrong |
|---|---|---|---|
| `normalize_axis_preserving::…_over_the_cmrg_corpus` | 148.8s | `Test oracle (1/3)` | already parallel — per-row cost |
| `normalize_axis_preserving::…_over_the_clinvar_unique_corpus` | 134.7s | `Test oracle (1/3)` | same |
| `cis_junction_crossing_shift::no_two_member_allele…` | 65.9s | `Exhaustive sweeps` | serial over 48 independent sequences |
| `spec_conformance_axis::five_prime_conformance_census` | 64.7s | `Test (2/6)` | serial over reference groups |
| `spec_conformance_axis::three_prime_conformance_census` | 63.1s | `Test (5/6)` | same |

The two `spec_conformance_axis` censuses and the sweep are the pattern #1775 proved: work already grouped so that each group owns its provider, run serially. Both become one task per group, with the same byte-identity argument — exact counters over a partition, `collect` preserving input order, and each capped sample list reproduced by capping per task and truncating the in-order concatenation.

## The two biggest are a different problem

They were **already** `into_par_iter`, so parallelism was not the lever. The lever is per-row cost, and three things were being rebuilt for every one of millions of rows:

1. **A whole `AxisCensus` per row.** `of_one` returned one — including a `per_axis` `BTreeMap` node on every *compared* row — which `reduce` then merged pairwise. It becomes `absorb_one`, folding in place, so `fold` builds one accumulator per rayon chunk instead of one per row.
2. **A `Normalizer<MockProvider>` per row.** The provider is empty and never populated — deliberately, since the property under test is the axis *label*, which no reference can inform — so every row was constructing an identical value. It moves to the fold's init.
3. **Two redundant `coordinate_axis()` calls per compared row.** `classify` re-derived both values the caller had already resolved to decide the row was comparable. It now takes them.

The arithmetic is untouched, so every pinned figure and the four-way disposition accounting are identical, and `merge` still folds the chunks.

## What is verified and what is not

**Verified:** `cargo check --features dev --test it` is clean, and `cargo fmt` passes.

**Not verified:** I have not run these tests locally. The dev machine is contended, and a wall-clock number taken on a loaded box is not a measurement — nor is a test run that is competing with four other cargo processes a trustworthy green. So the pinned censuses in all three modules are **unverified against this change**, and that is the first thing to read off CI:

- `spec_conformance_axis`'s seventeen-field `Census` pins (both directions);
- `cis_junction_crossing_shift`'s `FIVE_PRIME_DUP_DEL_SEQUENCE_CHANGES` exact zero, its per-seed case-count floor, and the `skipped * 10 < checked` ratio — the one assertion in that file that is not subset-stable;
- `normalize_axis_preserving`'s disposition accounting, which asserts the four buckets sum to the rows offered.

If any of those moves, the change is wrong and not the pin. None of these commits may re-bless a number.

**Then the timings**, against the table above — `Test oracle (1/3)`, `Test (2/6)`, `Test (5/6)` and `Exhaustive sweeps`.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,6 +1,24 @@
 # Codecov configuration
 # https://docs.codecov.com/docs/codecovyml-reference
 
+# The `Code Coverage` job is a 4-way matrix, so a commit gets FOUR uploads and
+# Codecov merges them server-side. Verified: the sharded run reports 82.52%
+# against 82.51-82.53% for the unsharded runs either side of it, so the merge is
+# exact.
+#
+# `after_n_builds` makes it deterministic rather than eventually-consistent.
+# Without it Codecov posts a status as soon as the FIRST upload lands — one
+# shard's quarter of the suite — and corrects as the rest arrive, so anyone
+# reading the check in that window sees a coverage collapse that is not real.
+# Both statuses are `informational: true` so nothing blocks either way; what this
+# removes is the misleading transient.
+#
+# Keep this equal to the shard count in `.github/workflows/coverage.yml`. Set it
+# too HIGH and no status is ever posted; too low and the race returns.
+codecov:
+  notify:
+    after_n_builds: 4
+
 coverage:
   status:
     project:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1184,8 +1184,39 @@ jobs:
   # what the oracle asserts, so re-rolling thousands of random cases here checks
   # nothing the plain run did not already check — it just cost ~80s of every CI
   # run. The deep random coverage lives in the `soak` job below instead.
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai pause

PR #1783, opened 2026-08-13
fix(reference): serve the synthesized transcript's exon table on its own axis

### LINKED ISSUE #1773: cdot-synthesized transcripts publish a flat exon table over a compact sequence
`MultiFastaProvider::synthesize_transcript_from_cdot` (`src/reference/multi_fasta.rs:1924`, sole call site `:2758`) builds a transcript sequence by concatenating each exon's genome window after applying that exon's CIGAR (`:1969-2026`). What it serves is therefore **exon bases only** — a *compact* sequence, with no position for any transcript base the cdot exon table does not cover.

It then writes a `TxExon` table straight off cdot's `tx_start`/`tx_end` (`:2052-2067`):

```rust
start: e[2] + 1,   // tx_start: 0-based incl → 1-based incl
end: e[3],         // tx_end:   0-based excl = 1-based incl
```

Those are absolute offsets into the **flat** transcript sequence — a base that fails to align to the genome still occupies a position in them. No length reconciliation happens between the two. So a flat exon table is published over a compact string.

The two spaces coincide only when the exon table is contiguous **and starts at 0**. A non-zero head offset (first exon's `tx_start > 0`) is just as non-identity as an internal hole.

### Measured over the prepared reference's cdot (0.2.32)

Non-identity is defined as `sum(exon tx spans) != last exon tx_end`, i.e. the compact sequence's coordinate space differs from the flat one:

| build | records with an exon table | compact != flat | head-offset only | internal gap only | both |
| refseq GRCh38 | 482,519 | **167** | 109 | 58 | 0 |
| refseq GRCh37 | 197,846 | **442** | 283 | 153 | 6 |

No record in either file has overlapping exon transcript spans as its *only* source of non-identity.

### Consequence

Everything that reads the served exon table — genome↔transcript mapping, exon/intron geometry, junction-crossing checks — gets flat positions, which then index a compact byte slice (`Transcript::get_sequence`, `src/reference/transcript.rs:524`). On an affected record every such position is off by the head offset plus every preceding hole.

The CDS bounds on this path are, today, in the *same* space as the sequence and are not part of this defect: cdot's `start_codon`/`stop_codon` are gap-collapsed offsets, which is exactly the compact space. That was verified rather than assumed — slicing the deposited transcript FASTA for every non-identity GRCh38 record that carries both a CDS and a sequence:

| reading of `start_codon`/`stop_codon` | valid ORF (gapped) | valid ORF (head-offset) |
|---|---|---|
| flat offsets | 0 / 50 | 0 / 46 |
| gap-collapsed offsets | **50 / 50** | **41 / 46** |

(The 5 head-offset records that are invalid under both readings are invalid for their own reasons — no ATG or a non-multiple-of-3 span — and do not discriminate.)

So the exon table is the half that is incoherent, and it is incoherent today. Any change that moves the served CDS bounds into flat space must map them back at this seam, or the CDS becomes incoherent with the served sequence too.

### Fix

Map the `TxExon` table into the compact space the synthesized sequence actually is, inside that one function, and pin the resulting internal coherence — the served exon table must tile the served sequence exactly, and the served CDS bounds must name positions that exist in it.

The FASTA-backed path is not affected and must not move: there the served sequence is the flat deposited one, and its flat exon table is correct.

nh13:
Closes #1773.

**Merge-order note:** this is deliberately standalone and off `origin/main` (`57f71d31`), and it is meant to land **before #1737**. The half of the incoherence it fixes is pre-existing; the half #1737 introduces is left to #1737, with a guard here that fails if #1737 lands without it. See "What this leaves to #1737" below — that section is the one to read if you are reviewing the two together.

`synthesize_transcript_from_cdot` builds a transcript by concatenating each exon's genome window after applying that exon's CIGAR, so what it serves is the **gap-collapsed** transcript: exon bases only, with no position for any transcript base cdot's exon table does not cover. It then published cdot's `tx_start`/`tx_end` over those bytes, and those are absolute offsets into the **flat** transcript. Nothing reconciled the two.

The spaces coincide only for an exon table that is contiguous **and** starts at 0. A non-zero head offset is exactly as non-identity as an internal hole, and on this cdot it is the larger class.

## Population, measured

Independently of ferro, straight off the prepared reference's cdot JSON (`compact != flat` defined as `sum(exon tx spans) != last exon tx_end`):

| build | records with an exon table | compact != flat | head-offset only | internal hole only | both |
| refseq GRCh38 | 482,519 | **167** | 109 | 58 | 0 |
| refseq GRCh37 | 197,846 | **442** | 283 | 153 | 6 |

Neither file has a record whose *only* source of non-identity is overlapping exon transcript spans.

## What actually reaches this code path, end to end

Synthesis is the last resort: `resolve_name` prefers any FASTA record and falls through a version-strip, so an accession reaches it only when no FASTA record exists at all. Resolving all 167 / 442 through `MultiFastaProvider` against the local prepared reference:

| build | reach synthesis | decline (insertion CIGAR, E2005) | **served** | not measured |
|---|---:|---:|---:|---:|
| GRCh38 (primary path) | 13 | 9 | **4** | 0 |
| GRCh37 (parented) | 15 | 3 | **11** | 1 |

All 15 served transcripts had `tiles == false` before this change: the published table neither started at the first served base nor ended at the last.

The one not measured is `NR_073509.1` on GRCh37 — the harness reaches a build hint by spelling a parented `c.` variant, which a non-coding accession refuses, so nothing is claimed about it on that build. It *is* served on GRCh38 and moves there.

**On GRCh38, 0 of the 58 gapped records are served — and that zero is a property of this reference, not a structural one.** All 4 GRCh38 records served are head-offset. The GRCh37 arm is the disproof of any structural reading: `NM_020979.5` (SH2B2) is a genuinely *gapped* record, 249 bases, and it is served. A consumer with a thinner transcript FASTA routes more of both classes here.

## What moves

For each served transcript, the genomic position of the first served base, the last served base, and `c.1` — `origin/main` against this branch:

| build | `n.1` moves | last base moves | `c.1` moves |
|---|---:|---:|---:|
| GRCh38 | 4 / 4 | 4 / 4 | 3 / 4 (the fourth is an `NR_`, no CDS) |
| GRCh37 | 10 / 11 | 11 / 11 | 10 / 11 |

Before the fix, `n.1` resolved to **nothing** on 14 of the 15: the flat table started past it, so the first `head` bases of the served sequence were unreachable from the transcript axis. Worked rows:

```
NM_001354366.1  (ELOA3C, GRCh38, head 423)  n.1  none -> 46969912   c.1  none -> 46969912
NM_001387112.1  (POLK,   GRCh38, head  24)  n.1  none -> 75547010   c.1  75582117 -> 75583310
NM_004007.2     (DMD,    GRCh37, head  46)  n.1  none -> 33038317   c.1  32841445 -> 32834745
NM_020979.5     (SH2B2,  GRCh37, gap  249)  n.1  unchanged          last base 101961929 -> 101962178
```

`NM_020979.5` is the shape that shows the hole rather than the head: its `n.1` and `c.1` are before the gap and do not move, and everything past the gap does.

## The fix

Build the exon table from the bases each exon actually contributed, accumulated during synthesis. It then tiles the served sequence from 1 by construction, and agrees with the CDS bounds published beside it. The genome axis of each exon is untouched — that is cdot's alignment, stated in genome coordinates, and no transcript-space question can move it.

Building from `tx_end - tx_start` instead would be wrong for a second reason: an exon with no CIGAR emits its **whole genome window** whatever its declared transcript span says (the pre-#807 behaviour the "unverified exon" warning is about), so the table has to follow the bases, not the declaration. Pinned.

## What this leaves to #1737, deliberately

**The CDS bounds are already in the served sequence's space on `main`, and this PR does not touch them.** cdot's `start_codon`/`stop_codon` are gap-collapsed offsets — the same space the synthesized sequence is in — so carrying them through with only the 0-based → 1-based shift is correct here today. That was verified rather than assumed, by slicing the deposited transcript FASTA for every GRCh38 non-identity record that has both a CDS and a sequence:

| reading of `start_codon`/`stop_codon` | valid ORF, gapped | valid ORF, head-offset |
|---|---|---|
| as flat offsets | 0 / 50 | 0 / 46 |
| as gap-collapsed offsets | **50 / 50** | **41 / 46** |

(The 5 head-offset residue records fail under both readings for their own reasons — no ATG, or a span that is not a multiple of 3 — so they discriminate nothing. The gapped column reproduces #1737's own measurement; the head-offset column extends it to the class #1737 did not measure, and agrees.)

#1737 moves those bounds into **flat** space, which is right for the FASTA-backed path and makes them wrong for this one. Under #1737 all 15 served transcripts get a CDS shifted by their head offset plus enclosed holes: 1 per build lands **past the end** of its own served sequence (`NM_001354366.1`, `cds_end` 1218 → 1641 on 1218 bases; `NM_020979.5`, 1797 → 2046 on 1878 bases) and the remaining 2 (GRCh38) and 10 (GRCh37) shift **silently**, staying in range while naming different bases.

Writing that mapping here would be wrong today — it would move values that are currently correct. So instead this PR marks the seam in the code and pins the invariant with `synthesized_cds_bounds_are_on_the_served_sequences_axis`, which judges the bounds by what they **denote** (they must slice a start codon, a whole number of codons, and a terminator out of the served sequence) rather than by their numbers. Simulating #1737 at that seam fails it. That is the intended handshake: #1737 maps the bounds back to the collapsed space inside this function, and the test goes green.

## No cache version bump, and why that differs from #1737

#1737 had to bump `RKYV_FORMAT_VERSION` and `CDOT_BINCODE_VERSION` because it changed the **parsed cdot values** that both caches persist, so a stale cache would keep serving the old ones. This change persists nothing: it is a read-time transformation of a `CdotTranscript` into a `Transcript`, and the only thing memoizing that is the in-process `transcript_cache` (an `LruCache` keyed on `(id, build_hint)`, rebuilt every run). Existing prepared references need no re-prepare and no re-parse.

## One assumption, named rather than assumed

cdot states `start_codon`/`stop_codon` over the concatenation of the exons' **declared** transcript spans, while the served sequence is the concatenation of what each exon **emitted**. Those agree for every CIGAR-bearing exon (#807 checks it) and for every no-CIGAR exon whose genome window equals its declared span. Measured: **0 of 482,519** GRCh38 and **0 of 197,846** GRCh37 builds have a no-CIGAR exon where they differ.

That zero is a property of this cdot release — its Gap attribute is written exactly when the alignment has indels — not a structural guarantee, so it is written down at the seam rather than relied on. A record that broke it would already be warning through `cdot_synthesis_has_unverified_exon`, so the case is loud rather than silent, and `synthesized_exon_table_follows_the_bases_served_not_the_declared_span` builds exactly that shape (with no CDS) to pin which side wins.

## What is deliberately not touched

The **FASTA-backed** path. There the served sequence is the flat deposited one — head offset and hole present as real bases — so cdot's flat table is the correct table for it and is published verbatim. That is now guarded (`fasta_backed_transcript_keeps_cdots_flat_exon_table`), because it was not: the mutation that shifts that table by the head offset passed all 4194 `--lib` tests. `cds_to_tx`/`tx_to_cds` are also untouched — the flat reading is `c-and-n-positions-are-flat-transcript-offsets` and is not in question here.

## Verification

`cargo fmt --all`, `cargo clippy --features dev --all-targets` clean. Full suite `cargo nextest run --features dev --test-threads=4 --no-fail-fast` with `FERRO_MANIFEST` set (so the reference-aware suites run rather than skip): **10,434 run, 10,434 passed, 152 skipped**, exit 0.

Test names by set difference against `origin/main` at the same base (`cargo nextest list`), never by count: 10,430 → 10,434, **+4 / -0**, the four added being exactly the tests below.

Five mutants, each applied to this branch and each killed:

| mutant | killed by |
|---|---|
| publish cdot's flat table again | all 3 synthesis tests |
| off-by-one on the compact start | 2 of them |
| build the table from `tx_end - tx_start` | `..._follows_the_bases_served_not_the_declared_span` |
| CDS bounds arrive flat (simulating #1737) | `synthesized_cds_bounds_are_on_the_served_sequences_axis` |
| shift the **FASTA** path's table by the head offset | `fasta_backed_transcript_keeps_cdots_flat_exon_table` (and nothing else — that mutant passed all 4194 `--lib` tests before this guard existed, which is why it was added) |

Representation-Change: on a cdot-SYNTHESIZED transcript (one no transcript FASTA record serves) whose cdot exon table is not contiguous from 0 — a head offset, an internal hole, or an exon emitting a different number of bases than its declared transcript span — the genome alignment is now withheld: every exon's `genomic_start`/`genomic_end` and the transcript's own `genomic_start`/`genomic_end` are `None`, so `CoordinateMapper::tx_to_genomic` and every consumer above it (normalize, VCF conversion, equivalence checking, W-code correction) now REFUSE a c./n. position that previously received a genomic coordinate. Measured on the prepared reference, 15 such transcripts are served, on 14 distinct accessions — 4 on the GRCh38 primary path and 11 on GRCh37, among them NM_001354366.1, NM_001387112.1, NM_004007.2 and NM_020979.5. Nothing else moves: 482,352 of 482,519 GRCh38 and 197,404 of 197,846 GRCh37 cdot records have a contiguous zero-based table and keep their genome axis unchanged, no FASTA-backed transcript is affected, and for the 15 the served bases, the (gap-collapsed) exon table and the CDS bounds are all still served. The genome coordinate for those accessions remains available on the route stated on the deposited transcript's own axis — `CdotTranscript::tx_to_genome`, used by the projector's non-coding arm, `data/mapping.rs` and the convert service — which is untouched. That route is why the withholding is a fix rather than a loss: publishing both axes left ferro returning two different coordinates for the same n. position, each an `Ok`, with nothing in the answer to say which route produced it (head-offset shape: n.1 -> 1 against a decline, n.10 -> 15 against 3).

### DIFF
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -2310,6 +2310,65 @@ impl MultiFastaProvider {
             })
             .collect();
 
+        // #1783 — publish the genome axis only when the transcript axis this
+        // table is stated on IS the axis `n.`/`c.` positions are stated on.
+        //
+        // The table above is on the gap-COLLAPSED axis, which is the axis of
+        // the bytes this function serves. `n.k`, though, names the k-th base of
+        // the DEPOSITED transcript — the FLAT axis, which is the one cdot's own
+        // `tx_start`/`tx_end` are stated in. And that flat axis is what the
+        // OTHER transcript→genome route reads: `CdotTranscript::tx_to_genome`
+        // offsets from each exon's own `tx_start` and returns `None` for a base
+        // no exon covers. It is reached from `project/projector.rs`,
+        // `data/mapping.rs` and `service/handlers/convert.rs`, and it never
+        // touches this `Transcript` — so it is invisible to any census of who
+        // reads `Transcript.exons`.
+        //
+        // When the two axes coincide (a table contiguous from 0) the two routes
+        // are the same function and both are served. When they do not, a genome
+        // coordinate published here answers a question the deposited axis
+        // answers differently. Measured on the head-offset shape: this route
+        // says `n.1` → 1 where `tx_to_genome` declines, and `n.10` → 15 where
+        // `tx_to_genome` says 3 — two `Ok`s, different coordinates, no warning,
+        // and nothing in the answer telling a consumer which route produced it.
+        //
+        // So withhold the alignment rather than serve a second answer. This is
+        // `tx_to_genome_extending_polya`'s own rule one level up ("a gap/hole
+        // […] decline rather than guess"), and it is what biocommons/hgvs does
+        // with such a transcript from any provider. The bases, the collapsed
+        // exon table and the CDS bounds are still served — they are on the
+        // collapsed axis and are correct there, which is what the rest of this
+        // fix buys; only the cross-axis question is refused.
+        //
+        // The FASTA-backed path is unaffected: it serves the flat deposited
+        // sequence beside cdot's flat table, so its two routes already agree
+        // base for base, including declining together inside a hole.
+        let axes_coincide = tx
+            .exons
+            .iter()
+            .zip(exons.iter())
+            .all(|(e, published)| published.start == e[2] + 1 && published.end == e[3]);
+        let exons: Vec<TxExon> = if axes_coincide {
+            exons
+        } else {
+            warn!(
+                "{id}: cdot's exon table is not on the axis of the bases synthesis serves (a \
+                 head offset, an internal hole, or an exon emitting a different number of \
+                 bases than its declared transcript span), so the synthesized transcript is \
+                 served WITHOUT a genome alignment — a c./n. position on it resolves to a \
+                 genome coordinate only through the cdot record itself, which states the \
+                 deposited transcript's axis (issues #1773, #1783)"
+            );
+            exons
+                .into_iter()
+                .map(|e| TxExon {
+                    genomic_start: None,
+                    genomic_end: None,
+                    ..e
+                })
+                .collect()
+        };
+
         let (genomic_start, genomic_end) = {
             let min_start = exons.iter().filter_map(|e| e.genomic_start).min();
             let max_end = exons.iter().filter_map(|e| e.genomic_end).max();
@@ -6808,13 +6867,19 @@ NC_000001.11\tRefSeq\tmatch\t9000\t9099\t100\t+\t.\tID=a1;Target=NG_008000.1 1 1
             "the last exon must end at the last served base"
         );
 
-        // The genome axis is cdot's alignment and must not move with it.
+        // The genome axis is NOT re-tiled with it — cdot's alignment is stated
+        // in genome coordinates and no transcript-space question can move it.
+        // On this record it is withheld instead: the collapsed axis the table
+        // above is on is not the deposited axis `n.` positions are stated on,
+        // and publishing a genome coordinate against it would contradict
+        // `CdotTranscript::tx_to_genome` (#1783). See
+        // `route_parity_declines_when_the_flat_and_collapsed_axes_differ`.
         assert_eq!(
             tx.exons
                 .iter()
                 .map(|e| (e.genomic_start, e.genomic_end))
                 .collect::<Vec<_>>(),
-            vec![(Some(1), Some(9)), (Some(15), Some(23))]
+            vec![(None, None), (None, None)]
         );
     }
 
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
Applied the route-parity finding this review left open, and corrected the trailer.

**Route A now declines where route B declines.** Measured on the head-offset geometry before the fix: **58 violations over 18 positions**, reproducing this review's `n.10 -> 15` vs `3` exactly. The identity geometry passed at the same head, so the guard is not simply reporting that everything disagrees.

The direction is A-to-B, never B-to-A. Route B (`CdotTranscript::tx_to_genome`) offsets from the exon's own `tx_start` and declines in a gap — the convention upstream cdot adopted in `data_v0.2.34` (SACGF/cdot#123), matching biocommons, and already stated as policy in `tx_to_genome_extending_polya`: *"a gap/hole, not the poly-A region. Decline rather than guess."*

**Surgical scope:** only the *genome axis* is withheld. Bases, the collapsed exon table and CDS bounds are all still served, and the FASTA-backed path is untouched. Whole-record refusal would have flipped three of this PR's five positive tests; that reading is available if preferred but was not taken.

**Route census by route, not by consumer** — enumerated over every `fn *_to_genom*` in `src/`. Worth recording why the earlier reach census missed this: **route B mentions neither `Transcript` nor `exons`**, so it could never appear in a list of sites that consult them.

Four parity guards added, including a non-vacuity control (all four entry points must *answer* and agree on an identity geometry) and a FASTA-backed control where mutual decline is achievable (18 answer / 18 decline, identically on both routes).

Sabotage, both red: forcing `axes_coincide = true` (the exact defect) fails 2 of 4 with both controls green; a plausible *partial* fix that withholds only on a head offset fails exactly one — `internal hole only` — which is the discrimination that matters.

Gate `EXIT=0`, 10689 passed; clippy and fmt clean; Python 949 passed, 8 skipped, import verified to resolve to this worktree.

Two things left open deliberately: the CLI review's findings 1 and 2 (`saturating_sub` -> `checked_sub`; teaching `cdot_synthesized_tx_length` about the CDS-scoped refusal) are still owner's-call. And the trailer's population figures (15 served, 14 accessions, 482,352 / 197,404) are carried from the original measurement, restated for the new behaviour rather than independently re-derived.

PR #1784, opened 2026-08-13
docs: correct the #466 filing-status claim in a decided ruling record

nh13:
Representation-Change: none. Ledger text only; no watched file is touched.

## This is a factual correction, not a re-opening

The ruling is untouched. `id`, `status`, `governing`, `deviates_from` and the `clauses` array are byte-identical to `origin/main`; the only change is one sentence of rationale prose that made a false claim about **filing status**. Nothing about the adjudication is re-argued, and this should not be read as re-opening a settled record.

## The false claim

The record

    derivation-may-not-be-bounded-by-the-inputs-spelling

said of the fifteen unequal-length rows:

> … they are **filed upstream under tracking issue #466** rather than decided here.

Two things are wrong with that, and both were checked rather than argued.

**1. #466 is not upstream.** It is a `fulcrumgenomics/ferro-hgvs` issue — "Umbrella tracker for spec-*text* changes ferro would suggest to the HGVS nomenclature committee (SVD-WG)". Its own body defines each candidate as carrying, "once filed — a link to the upstream issue/consultation at `HGVSnomenclature/hgvs-nomenclature`". So #466 is the internal register of things queued *for* upstream, and calling it the place something is "filed upstream" is a category error.

**2. The question is entered there, but nothing about it has been filed upstream.** #466 does carry a candidate covering exactly these rows — *"Choosing among equally-compliant splits of an unequal-length `delins`"*, candidate 5 of 13 — and that candidate reads `Upstream issue: _not yet filed_`. So the obligation is registered, not discharged.

## The fix

The sentence now states both halves — registered internally, unfiled upstream:

```diff
-they are filed upstream under tracking issue #466 rather than decided here.
+they are entered as a #466 candidate against the HGVS nomenclature repository and not yet
+filed upstream, rather than decided here.
```

The sibling record `rna-repeat-range-plus-unit-redundancy` phrases its own #466 obligation as "Owed as a #466 entry against the HGVS nomenclature repository", so the target-naming form is the ledger's existing one rather than a new coinage; the added clause is the filing-status fact the sibling does not need to carry.

## On the upstream filing status, and how it was derived

Stated because it is easy to get wrong: **there is exactly one upstream contribution from this project, and it is not this question.** `HGVSnomenclature/hgvs-nomenclature#266` ("docs: align the 3'rule exon-junction exception with the RNA pages"), opened 2026-08-11, still open — a pull request, not an issue, and it addresses the exon-junction candidate, not the unequal-length-split one. Upstream issues: zero.

That was enumerated **by author**, not by keyword:

```
gh search prs    --repo HGVSnomenclature/hgvs-nomenclature --author nh13   # -> #266
gh search issues --repo HGVSnomenclature/hgvs-nomenclature --author nh13   # -> []
```

A keyword search for `ferro` or `fulcrum` returns `[]` and would have missed #266 entirely, since that PR is authored from a personal account and uses neither word. Anyone re-checking this claim should enumerate by author.

## What was verified

- **Ledger set-comparison against `origin/main`**: 25 records on both sides, none dropped, none added, and **no status change**.
- **`ruling_records_are_intact`** passes; `RULING_STATUSES` is untouched, as it must be for a text-only correction.
- **The three census pins in `tests/it/clause_ruling_index.rs` did not move.** `the_index_is_not_vacuous` still reads `(25, 22, 3)`. The edit adds and removes no clause citation, so the generated index block is unchanged and that file needed no hand-edit.
- **Full local suite green** — `cargo nextest run --features dev`, 10433/10433 passed, 152 skipped.

## Note on the neighbourhood

Several other open PRs touch this same sidecar file and will serialize against it. If one lands first this PR needs a rebase; the edit is a single sentence, so the merge is mechanical.


nh13:
@coderabbitai pause

PR #1787, opened 2026-08-13
refactor(normalize): drop the subsumed separation disjunct from the inversion gate

nh13:
`whole_block_inversion`'s admission gate is a four-way disjunction, and one of its disjuncts can never be the one that decides. `every_separation_is_a_single_base` is provably subsumed by `payload_columns_dominate_the_span`, so removing it from the chain cannot change a verdict. This drops it, pins the subsumption with a test, and corrects a citation the removal exposes as stale.

## The subsumption

Write the pieces' reference widths `wᵢ`, their payloads `altᵢ`, and `k = pieces.len()`. The span is `Σwᵢ + Σgaps`, and the separation predicate is exactly the constraint that every gap is at most one, so `span ≤ Σwᵢ + (k − 1)`. The payload total is `Σ max(wᵢ, altᵢ) ≥ max(Σwᵢ, k)`, since no piece is empty on both sides. So `span ≤ payload + k − 1 < 2·payload` whenever `payload ≥ k`, which it is — and `2·payload > span` is what the payload route tests.

Enumerated as well as argued, by the new `the_single_base_separation_regime_is_subsumed_by_payload_density`: over every separation-≤1 geometry with `k = 2..4` and widths/payloads `0..3`, **0 of 418,833** are missed by the payload route. The test is not a change detector — it fails under a deliberate sabotage of the payload predicate (dropping its `max(width, payload)` term back to reference width alone raises **12,393** misses, which is exactly the sibling's count below).

## `changed_columns_dominate_the_span` does NOT subsume it

Asserted in the same test, because it is what makes this a claim about the *payload* route specifically. The sibling counts reference width only, so a zero-width pure insertion reads as contributing nothing; it misses 12,393 of the same geometries. Dropping the disjunct on the strength of the density route would have moved output.

## What was refuted, recorded in the code

The doctrine in `CLAUDE.md` is that a measurement which kills a plausible belief is worth as much as a ruling. The belief here — stated in `coalesce_whole_block_inversion`'s own doc and in the predicate's — was that the single-base regime was *the* discriminator for the inversions this pass keeps whole: the one place a coincidentally matched interior column is least believable, hence the one place widening is safe. Being first in an `||` chain it short-circuited ahead of the others and so looked decisive. It was not. That is now written on `whole_block_inversion`, alongside the algebra and the non-subsumption by the sibling.

## The stale citation

`every_separation_is_a_single_base`'s doc cited `DNA/delins.md:44-47` as the licence for the regime it admits. The decided record `delins-payload-coincidence-carve-out-is-coding-dna-scoped` (2026-08-11) scopes `:47` to the **coding DNA axis** — `c.`, and nothing else — while `whole_block_inversion` runs ungated on `g.`. So on the genomic axis that stated ground was stale.

The *outcome* was never in doubt: what licenses widening to an `inv` is `DNA/inversion.md:5`, which defines the type on the sequence relation alone and carries no axis scope. The comment now says that, and `coalesce_whole_block_inversion`'s parallel citation of `:46-47` is qualified the same way — a recommendation ferro follows more widely than the clause reaches, not an authority for doing so off `c.`.

**One residue is named and deliberately not fixed here**, because fixing it would move output: the predicate's remaining caller is `partition_block`'s `split_buys_no_higher_priority_type` route, which is equally axis-blind and *is* the payload-coincidence shape. So the analogy's source clause is now `c.`-scoped while the site drawing on it is not. That is a representation question for the ruling's implementers, and the comment records that `AxisFrame::reading_frame` is the wrong gate for it — the ruling says so itself.

## Behaviour

Unchanged, by construction and by measurement. A separate investigation established that no fix confined to this gate is shippable — turning off the separation *and* density routes fails 11 tests, shatters #1461's 457 bp `NC_000013.10:g.100809575_100810031inv` into ~110 members, and breaks the spec's own published `NM_004006.2:c.4145_4160inv`. This PR is not that; it removes a disjunct that was never deciding anything.

Full suite, `cargo nextest run --features dev --no-fail-fast`: **10442 tests run: 10442 passed (11 slow), 152 skipped**, exit status captured into the log as `EXIT=0`. `cargo fmt --check` and `cargo clippy --features dev --all-targets -- -D warnings` both clean.

## This collides with #1706 — flagging, not resolving

#1706 (`fix/inversion-post-hoc-run-scan`) adds `inversion_gate_admits`, which copies this gate's four routes **verbatim**, explicitly so "the block-level and window-level answers cannot drift". Whichever lands second has to reconcile two things:

- **Textually**, #1706 inserts ~700 lines immediately above `whole_block_inversion`'s doc, which is where this PR's largest hunk is.
- **Semantically**, `inversion_gate_admits`'s doc states that *two* of its five disjuncts can never be the deciding one — it identifies `changed_columns_dominate_the_span ⊆ payload_columns_dominate_the_span` and `no_piece_is_a_lone_substitution ⊆ not_every_piece_is_a_lone_substitution` — and then lists "are the pieces close together (separation ≤ 1 throughout)" as one of the three questions that *do* execute. This PR measures that third one as subsumed too, so the count is three of five and separation is not an independent question. That paragraph needs updating, and this PR should not do it in someone else's branch.

Neither change moves output, so the reconciliation is documentation plus one disjunct, not a behavioural merge.

Representation-Change: none. Provably behaviour-neutral; the removed disjunct is subsumed by the payload-density route and can never be the deciding one.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -5365,17 +5365,23 @@ fn coalesce_compensating_gap_split(pieces: &mut Vec<Piece>, ref_bytes: &[u8]) {
 /// makes the derivation agree with the per-member pipeline instead of shredding
 /// what that pipeline types correctly.
 ///
-/// Gated on [`every_separation_is_a_single_base`], and that gate is load-bearing
-/// rather than defensive. `AAGCTA -> TAGCTT` is also a whole reverse complement,
-/// but only its first and last bases change and four unchanged bases lie between
-/// them. `general.md:34` describes two variants separated by one or more
-/// nucleotides individually; its letter names only `delins`, but its rationale
-/// (`delins.md:81-84` — the variants may have been reported, or may occur,
-/// individually) reaches any single spanning description, and at four unchanged
-/// bases these are two independent changes rather than interior columns of one
-/// reverse-complement relation. At exactly one unchanged base the match is far
-/// more likely to be the alignment's coincidence than structure — the same line
-/// [`every_separation_is_a_single_base`] draws inside [`partition_block`].
+/// Gated, and that gate is load-bearing rather than defensive. `AAGCTA ->
+/// TAGCTT` is also a whole reverse complement, but only its first and last bases
+/// change and four unchanged bases lie between them. `general.md:34` describes
+/// two variants separated by one or more nucleotides individually; its letter
+/// names only `delins`, but its rationale (`delins.md:81-84` — the variants may
+/// have been reported, or may occur, individually) reaches any single spanning
+/// description, and at four unchanged bases these are two independent changes
+/// rather than interior columns of one reverse-complement relation.
+///
+/// The gate used to *open* on [`every_separation_is_a_single_base`], on the
+/// reading that at exactly one unchanged base the match is far more likely to be
+/// the alignment's coincidence than structure — the same line that predicate
+/// still draws inside [`partition_block`]. That disjunct is gone, and
+/// [`whole_block_inversion`] records the measurement that removed it: the
+/// single-base regime is a strict subset of what
+/// [`payload_columns_dominate_the_span`] admits, so it never decided anything.
+/// The pair above is still refused, by the three surviving routes.
 ///
 /// Expressed through [`is_inversion`] rather than re-deriving the test, so the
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

Issue #1712, opened 2026-08-12
projector emits non-idempotent n. axes — 21 of 681, including one that is not 3'-shifted


nh13:
The projector hands callers `n.` descriptions that are not fixed points of bare normalization. Measured **21 of 681** projected non-coding axes move when re-normalized, with `FERRO_PARTITION` unset — i.e. on the shipping arm, on the prepared reference, at `origin/main` = `3228d228`.

The genomic axis has had a guard for exactly this since the projector stopped normalizing internally; the non-coding axis never got one. That asymmetry is why this has gone unnoticed: `axis_genomic_idempotent` passes on the same run.

## Why the genomic guard never covered it

The two axes are produced by different code. The genomic axis is normalized inside `project_variant`. The non-coding axis is re-framed by `frame_projection_owned` and fanned out across overlapping transcripts by `project_variant_all` — a path `axis_genomic_idempotent` never touches. So "the projector normalizes its axes" was verified for `g.` and assumed for `n.`.

## The 21 rows, by shape

**16 — the output re-renders as a repeat form.**

```
NG_008835.1(NM_022124.6):c.3304_3305del : NG_008835.1(NM_022124.6):n.3709_3710del -> NG_008835.1(NM_022124.6):n.3708_3710T[1]
NG_009299.1(NM_017668.3):c.*16_*18T[8]  : NG_009299.1(NM_002474.3):n.4364_4365insAAAAA -> NG_009299.1(NM_002474.3):n.4361_4364A[9]
NG_012337.1:g.4917GC[5]                 : NG_012337.1(NM_012459.2):n.36_37insGCGCGC -> NG_012337.1(NM_012459.2):n.33_36GC[5]
```

**4 — the output re-partitions into members.** All four come from cross-reference `delins` inputs:

```
NG_012337.3(NM_003002.4):c.274delinsNG_012337.1(NM_012459.2):c.200_203
  : NG_012337.3(NM_003002.4):n.309delinsCTGA -> NG_012337.3(NM_003002.4):n.[308_309insCT;310dup]
```

**1 — the output is not 3'-shifted.** This is the sharpest of the three, because the 3' rule is not a partition preference, it is the spec's own placement rule:

```
LRG_24:g.5526_5533del : LRG_24t1:n.245_252del -> LRG_24t1:n.246_253del
```

## Why it matters

[...]

PR #1788, opened 2026-08-13
test(projector): pin the non-coding axis non-idempotency residue

nh13:
`axis_genomic_idempotent` had no `n.` counterpart. The asymmetry was not a decision — it is where the genomic work stopped — and it mattered more than a missing test usually does, because **the two axes are produced by different code**: the genomic axis is normalized inside `project_variant`, while the non-coding axis is re-framed by `frame_projection_owned` and fanned out across overlapping transcripts by `project_variant_all`, a path the genomic guard never touches. A fix to one said nothing about the other, so "the projector normalizes its axes" was true of `g.` and unverified of `n.`.

## What it measures

Against the prepared GRCh38 reference, at `9fb126ba`:

| | tested | non-idempotent |
|---|---:|---:|
| `axis_noncoding_idempotent` (new) | 681 | **21** |
| `axis_genomic_idempotent` (control, same run) | 214 | 0 |

The genomic control reading zero on the same run is what isolates this to the non-coding path rather than to normalization generally.

## Why it is pinned rather than asserted to be zero

The residue is real and filed as **#1712**; this PR does not fix it and does not claim to. Pinning it at zero would assert a fix nobody has written, and `#[ignore]`ing the test would leave a guard that never guards — this repository has already lost one that way, where an `#[ignore]`d test owned regression seeds and the loss only surfaced when someone finally un-ignored it.

So it is pinned at its measured value with a tripwire, per the repo's own documented pattern. **Failing in either direction is the point**: more rows is a regression, fewer means something was fixed and this pin went stale without anyone noticing — which is how a partial fix comes to read as a complete one.

## The three shapes, and why one is called out separately

- **16** re-render as a repeat — `n.3709_3710del` → `n.3708_3710T[1]`, `n.36_37insGCGCGC` → `n.33_36GC[5]`. Tract maximization runs on the re-normalization that did not run on the projection.
- **4** re-partition into members — `n.309delinsCTGA` → `n.[308_309insCT;310dup]`, all four from cross-reference `delins` inputs.
- **1** is **not 3'-shifted**: `LRG_24t1:n.245_252del` → `n.246_253del`. That one is asserted separately, by name, because it is the only rule 2 miss among the 21 rather than a re-spelling — on a user-facing axis, and README rule 2 names the 3' rule in its own scope. Buried inside the aggregate count it would survive a fix to the other twenty without anyone noticing it had.

## One limitation, stated so this does not read as stronger than it is

The test returns early without a prepared reference, so **per-PR CI skips it**. It runs in the nightly reference-aware job, which is `continue-on-error` — so a broken pin surfaces in the xfail artifact rather than as a red build. This is the same property `axis_genomic_idempotent` already has, so it is consistent with the existing pattern rather than a new gap, but it means the pin is a nightly tripwire and not a merge gate.

## Verification

- Manifest-backed run of the guard: 1 passed, exit 0. Both new assertions are exercised — the test only passes if the count is exactly 21 **and** the non-3'-shifted row is present.
- Full suite: `10442 tests run: 10442 passed`, 152 skipped, exit 0.
- `cargo fmt --all`: clean.

Representation-Change: none. Adds one test; no watched file is touched and no output moves.

### DIFF
--- a/tests/it/mutalyzer_normalize_tests.rs
+++ b/tests/it/mutalyzer_normalize_tests.rs
@@ -4609,3 +4609,116 @@ fn axis_genomic_idempotent() {
         failures.join("\n")
     );
 }
+
+// The `n.` axis had no counterpart to the guard above, and the asymmetry was
+// not a decision — it is simply where the genomic work stopped.
+//
+// That matters more than a missing test usually does, because the two axes are
+// produced by different code. The genomic axis is normalized inside
+// `project_variant`; the non-coding axis is re-framed by
+// `frame_projection_owned` and fanned out across overlapping transcripts by
+// `project_variant_all`, which is a path the genomic guard never touches. So a
+// fix to one says nothing about the other, and "the projector normalizes its
+// axes" was true of `g.` and unverified of `n.`.
+//
+// The contract asserted is the same one `axis_genomic_idempotent` asserts and
+// is the weakest thing worth having: whatever the projector hands a caller must
+// already be what bare normalization would produce. A caller that normalizes a
+// projected axis — which is the obvious defensive thing to do — must not see
+// the string change under them.
+
+#[test]
+fn axis_noncoding_idempotent() {
+    let (Some(projector), Some(normalizer)) = (variant_projector(), normalizer()) else {
+        eprintln!("axis_noncoding_idempotent: skipping — no manifest");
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1791, opened 2026-08-13
fix(benchmark): make the supplemental FASTA the single authority for presence, and upsert by accession

### LINKED ISSUE #1758: fetch_fasta_to_file appends a duplicate FASTA record when a hollow supplemental row is re-fetched
Found while reviewing #1732, which is the change that makes this reachable. Filing rather than fixing in that PR because the repair has two defensible designs with different costs, and the choice is an operator call.

## The invariant that is being removed

`fetch_fasta_to_file` (`src/benchmark/cache.rs`) opens its output FASTA in **append** mode and every write path appends unconditionally — the GenBank loop, and both arms of the CON/FASTA fallback.

Today that is safe only because of a coincidence between two lines: the resume filter is `!metadata.transcripts.contains_key(acc)`, and **every** FASTA write inserts a metadata key in the same iteration. So "this accession is in the FASTA" implies "this accession is skipped on the next run", and the duplicate-append path is unreachable via resume.

#1732 keys the resume filter on usability (`has_usable_record`) instead of presence, which is correct and is the point of that PR. But it breaks the coincidence: an accession can now be re-fetched while its sequence is already in the FASTA.

## The population this lands on is exactly the one #1732 exists to repair

Measured against the prepared reference:

```
patterns_transcripts.metadata.json : 140,027 records, 116,572 of them hollow
patterns_transcripts.fna           : 140,027 FASTA headers, 1,345,248,893 bytes
hollow accessions also present in the FASTA : 116,572 of 116,572
```

Every hollow row already has its sequence on disk. So the re-provisioning #1732 enables would append a second copy of 116,572 records to a 1.28 GB file rather than repairing it in place.

## The mode is already observable

The sibling `patterns_transcripts.fna.bak` in the same directory carries **159,456 records for 140,026 unique accessions — 15,873 accessions duplicated, up to 4x**. So this file has been through duplicate-appending before, at smaller scale, presumably via the other route into the same hazard: the metadata sidecar is written only once, after the final batch, so an interrupted run leaves the FASTA ahead of the metadata and the next run re-fetches and re-appends.

## Options

1. **Rewrite rather than append.** Stream the existing FASTA to a sibling temp path, dropping records for accessions being re-fetched, append the new ones, then rename. Correct for both routes into the hazard, costs a full rewrite of a 1.3 GB file per provisioning run, and needs the metadata written in the same atomic step to stay consistent.
2. **Skip the append for an accession already present.** Read the existing FASTA headers (or its `.fai`) up front, and for a re-fetched accession that is already present, update the metadata record without writing the sequence again. Cheap, and it repairs the shipped artifact — whose FASTA content is fine and whose metadata is what is hollow. Does not repair a record whose FASTA *content* is bad, but nothing suggests that is the failure here.
3. **Checkpoint the metadata per batch** — orthogonal to the above and worth doing regardless; it closes the interrupted-run route that produced the `.bak`.

Option 2 looks like the right size for the actual defect, but it encodes an assumption (the FASTA bytes are trustworthy, only the sidecar is degraded) that should be stated deliberately rather than inherited.

## Suggested acceptance test

Run the resume filter over a cohort containing a hollow row whose sequence is already in the FASTA, and assert the FASTA gains no second header for it.

## Note

`patterns_transcripts.fna.bak` is 2.5 GB of duplicate-bearing artifact sitting beside the live one; worth deciding separately whether it should still be there.

nh13:
Closes #1758.

Fixes the root cause rather than the appending symptom: **two artifacts, updated non-atomically, were answering one question.** "Do we have sequence for this accession?" was answered by `*.metadata.json` while the sequence itself lived in the FASTA, which `fetch_fasta_to_file` opened in append mode and wrote to unconditionally from all three of its write paths.

That was safe only by coincidence. The resume filter was `!metadata.transcripts.contains_key(acc)` and every FASTA write inserted a metadata key in the same iteration, so "this accession is in the FASTA" implied "this accession is skipped next run" and the duplicate-append path was unreachable via resume. Nobody wrote that coupling down; it emerged. Both routes out of it are observable — #1732 keys resume on usability and breaks it by design, and the shipped `patterns_transcripts.fna.bak` carries **159,456 records for 140,026 unique accessions** (reproduced) from the other route, where the sidecar is written once after the final batch so an interrupted run leaves the FASTA ahead of it.

Underneath both: **a FASTA has no uniqueness constraint.** A duplicate header is legal FASTA and silently wrong here, because the file is being used as a keyed store with no key. The write path has to supply the constraint the format lacks.

## What is authoritative and what is derived

| artifact | status | recovered how |
|---|---|---|
| `<name>.fna` | **authoritative** for presence and sequence | — |
| `<name>.fna.fai` | derived | rebuilt from the committed FASTA at every commit |
| `<name>.metadata.json` | derived **key set** and `sequence_length`; carried-forward CDS annotation | keys reconciled against the FASTA at every commit; a missing row is synthesized from the record |

The sidecar is not *fully* derived and the code does not pretend it is: CDS coordinates come from a GenBank record and a FASTA cannot re-supply them, so an annotated row is carried forward untouched. What is derived is exactly the part that was being asked the presence question.

### Keying on the FASTA rather than the `.fai` is this repo's existing ruling, not a new preference

The issue proposes reading the `.fai`. `prepare::run_backfill` already faced this for the sibling backfill FASTA and decided the other way, saying why at its `load_fasta_header_accessions` call:

> The backfill present set is read from the FASTA headers themselves (the file `execute_backfill` appends to), not its derived `.fai`: a prior run interrupted after the append but before reindex — or a deleted `.fai` — would otherwise look empty and re-fetch and re-append the same accession, duplicating its `>accession` record (and colliding `.fai` keys on the subsequent reindex).

That is this defect, on the neighbouring artifact, with the same remedy. Applying the same rule rather than inventing a second one is the point. The reasoning also holds on its own: the `.fai` is a second artifact that can drift from the FASTA — on the shipped reference it is seven minutes newer than the `.fna` it indexes — so nominating it as the authority inverts the dependency and reintroduces the defect class one level over. Scanning headers costs one sequential read, **measured at 2.08 s over the 1.3 GB shipped artifact**, against a fetch half that is networked, rate-limited and measured in hours.

This also makes *writing* the index load-bearing rather than tidy: the merge re-emits every record at a uniform line width, so a commit invalidates any externally-built index it does not replace.

## Upsert, and where the atomicity boundary is

Fetched records are appended to a durable, resumable staging file (`<name>.fna.incoming`). `SupplementalStore::commit` then streams the committed FASTA and the staging file into `<name>.fna.tmp` keyed by accession — one record per accession, a staged record replacing the committed one, every duplicate dropped — then `fsync`s and renames it over the FASTA.

**That rename is the atomicity boundary, and it is the only one needed.** The index and the sidecar are rewritten after it, and a crash in between is self-healing because the next `open` re-derives both from the FASTA. Making the derived artifacts genuinely derived is what buys consistency across three files without a three-file atomic write, which is not available. The staging file is removed last, so a crash before that replays a merge that is deterministic and produces the identical FASTA.

Staging also makes an interrupted run lose nothing: its records count as present on the next run, so no accession is fetched twice — which closes the `.bak` route directly, and makes the issue's **option 3** (per-batch metadata checkpointing) fall out rather than needing separate work. I agree with that reading, with one correction to the framing: the durable per-batch state is the staged *sequence*, not a checkpointed sidecar. Checkpointing the sidecar would have kept two artifacts and merely made them disagree less often; the sidecar is now reconstructible, so there is nothing left to checkpoint.

## Repair is not a mode

Dedup-by-accession is what the merge does, so a duplicate-bearing artifact is repaired by the next provisioning run and is a fixed point thereafter — no separate command, no operator knowledge required, and covered by the same idempotence property rather than by a parallel one. A run with nothing to fetch and nothing to reconcile writes **no byte at all**.

## The acceptance test is a fixed point — and where that alone is not enough

`provisioning_is_idempotent_and_the_second_run_writes_nothing` runs the real provisioning path twice and asserts the second run fetches nothing and leaves all three artifacts byte-identical **and unmodified by mtime**. The mtime is not redundant with the bytes: the merge is deterministic, so a run that needlessly rewrote all three would produce identical content and satisfy a content-only comparison. "No-op" means no write, and only the mtime witnesses that. Mutation M4 below confirms it — four tests catch a forced rewrite, and they only do so because of the mtime.

**One correction to the framing in the issue, measured rather than argued.** Seeded from an *empty* directory, run-twice-is-a-no-op **passes on the pre-fix behaviour too**: an artifact that behaviour produced has a FASTA and a sidecar that agree, so its sidecar-keyed filter skips correctly. The fixed-point property is satisfied *by* the very coincidence being removed. What is red on the old behaviour is the same property seeded from a state where the two artifacts **disagree**. Both tests exist and are named in a comment on the clean-start test so it is not later deleted as redundant.

`EfetchSource` abstracts the two `curl` invocations so this is a property of the production path rather than of a predicate; it is unstatable in a test that cannot run that path at all.

## Verification

18 unit tests, all green. Full local suite **10486/10486**. `cargo fmt --all --check` and `cargo clippy --features dev --lib --all-targets -- -D warnings` clean.

**Pre-fix reconstruction** — the two defective behaviours restored together (sidecar-keyed resume + blind append), each mutation verified to have actually applied before running:

```
a_duplicated_accession_is_collapsed_to_one_record ............ FAILED
  left: ["NM_000001.1","NM_000002.1","NM_000001.1","NM_000001.1"]
a_staged_record_replaces_the_committed_one_rather_than_appending  FAILED
a_duplicate_bearing_artifact_is_repaired_and_then_stable ..... FAILED
an_accession_in_the_fasta_but_missing_from_the_sidecar_is_not_re_fetched  FAILED
  fetch calls: left 1, right 0        <- the `.bak` route, reproduced
test result: FAILED. 11 passed; 4 failed
```

**Mutation testing — 12 mutations, 12 killed**, call sites included. Each was checked to have changed the file before its run, so a silently-inapplicable mutation cannot read as an uncaught one (one first-round mutation did fail to compile and was rewritten rather than counted):

| mutation | result |
|---|---|
| M1 resume filter keyed on the sidecar (call site in `cache.rs`) | killed by 1 |
| M2 staged record does not replace the committed one | killed by 1 |
| M3 duplicates are not collapsed | killed by 2 |
| M4 `is_dirty` always true (needless rewrite) | killed by 4 |
| M5 `is_dirty` always false (never repairs) | killed by 5 |
| M6 `has_sequence` ignores staged records | killed by 1 |
| M7 sidecar orphan rows are not dropped | killed by 1 |
| M8 index offset off by one | killed by 1 |
| M9 synthesized sidecar row carries no length | killed by 2 |
| N1 torn trailing staged record treated as complete | killed by 1 |
| N2 `is_dirty` ignores the index | killed by 1 |
| N3 final commit not gated on `is_dirty` (call site) | killed by 3 |

Not covered by mutation: the `fsync`/rename ordering, which needs crash injection to exercise. Stated rather than implied.

## Review round: four defects found in this change, and one lesson about my own battery

The `coderabbit --agent` lens raised six findings against this diff. Four were real and are fixed here; the disposition of all six:

- **A torn staged record could become authoritative** (fixed). A crash part-way through `stage_record` leaves a truncated trailing record; counting it as present meant the accession was never re-fetched and the *truncated* sequence was merged into the FASTA — the defect class this PR closes, reintroduced one level down by the staging file. Records now carry a `;end` terminator and an unterminated trailing record is discarded and re-fetched. Without this the PR's "durable and resumable" claim was not true.
- **The final commit was unconditional** (fixed). Reaching it having staged nothing is a real outcome — every batch can fail against a down endpoint — and a failed run rewrote the artifact it could not add to. Now gated on `is_dirty`.
- **`is_dirty` ignored the `.fai`** (fixed). Declaring the index derived and then not noticing it was missing left it derived in the documentation only; a deleted index also makes the artifact invisible to `prepare::load_existing_accessions`, which enumerates by `.fai`. Now a term, on names (a stale *offset* with correct names is not detected, and the code says so).
- **`curl` had no timeouts** (fixed). Hundreds of sequential batches with no `--max-time` means one stalled connection hangs the run. Added, generously, plus `--fail-with-body` so an HTTP error page is a failed batch rather than a batch that silently fetched nothing.
- **A short `read` in the index test** (fixed) — a flake source, not a defect.
- **A store-wide lock file for concurrent runs** (rejected). Concurrent provisioning against one reference directory was already unsafe — the previous code had two processes appending to one FASTA, which is strictly worse — so this neither introduces nor worsens it, and lockfile lifecycle is its own change. Worth filing separately.

**The lesson worth recording:** my mutation battery missed the unconditional-commit defect because every mutation targeted the store, not the call site in `cache.rs` that drives it. A battery that only mutates the unit under test cannot see a caller that fails to use it correctly. N3 was added for exactly that shape and is killed by three tests.

Separately, and consistent with what the #1732 review found: **no `path_instructions` glob in `.coderabbit.yaml` matches `src/benchmark/`.** The nearest, `src/{batch/**/*.rs,parallel.rs,cache.rs}`, expands to `src/cache.rs`, not `src/benchmark/cache.rs`. So the repo's own acquisition-and-caching lens — whose clauses ("a partial download left in place under the final filename instead of being written to a temporary path and renamed", "a cache keyed on fewer inputs than determine the value") describe this PR almost exactly — is configured off for the file it changes. I applied it by hand; it is worth fixing in the config, separately from this PR.

## Relationship to #1732

This supersedes #1732's resume-filter *mechanism* and leaves its *question* intact and safe to ask. Presence now comes from the sequence data rather than the sidecar; #1732's usability clause becomes a one-line extension of the same predicate (`!store.has_sequence(acc) || !metadata.has_usable_record(acc)`), and with the write path upserting it can no longer duplicate anything. `a_staged_record_replaces_the_committed_one_rather_than_appending` pins that property at exactly the layer #1732's predicate feeds. Nothing was pushed to that branch.

On the shipped artifact the mechanism change moves nothing: the FASTA and the sidecar key sets agree exactly — 140,027 each, 0 in either and not the other (verified) — so the same accessions are skipped either way.

#1732's other half (refusing a corrupt sidecar rather than silently treating it as a first run) is untouched here and still worth landing; this PR keeps `unwrap_or_else` exactly as `main` has it so that change stays #1732's to make.

## Provenance of `patterns_transcripts.fna` — answered

The issue records that no producer could be found and that the name may come from an out-of-tree script. It does not: **`src/bin/benchmark.rs` writes it**, in the `ferro-benchmark prepare ferro --patterns` path — the path is the `output_file` argument to `fetch_fasta_to_file`, then the argument to `index_fasta`, then assigned to `manifest.supplemental_fasta`. Two binaries write a supplemental FASTA from the same `--patterns` input under different fixed names: `ferro prepare` writes `clinvar_transcripts.fna` and `ferro-benchmark prepare ferro` writes `patterns_transcripts.fna`. No stem derivation is involved; both literals are hardcoded. The operator's reference matches the benchmark binary's name and its manifest carries `"supplemental_fasta": "supplemental/patterns_transcripts.fna"`.

So the design does not rest on an unevidenced claim about that file's origin — it was written by this repo, by the function this PR fixes.

Representation-Change: none. Producer-side provisioning under `src/benchmark/`; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched. Reasoning rather than assertion, because the artifact this writes *is* read by reference-aware runs: what changes about it is that duplicate records are collapsed and the sidecar's key set is reconciled, and no normalization consults either — the consuming side resolves sequence by accession, and a de-duplicated FASTA serves the same bases for every accession it served before. No normalized output moves.

### DIFF
--- a/src/benchmark/cache.rs
+++ b/src/benchmark/cache.rs
@@ -1908,6 +1908,106 @@ pub struct SupplementalMetadata {
     pub transcripts: std::collections::HashMap<String, SupplementalTranscript>,
 }
 
+/// Source of raw NCBI efetch payloads for one batch of accessions.
+///
+/// Exists so the provisioning path can be exercised end-to-end without a network. The
+/// property that matters most about that path — that running it twice is a no-op the
+/// second time — is unstatable in a test that cannot run it at all.
+pub trait EfetchSource {
+    /// GenBank-format payload for a comma-separated accession list.
+    ///
+    /// `Ok(None)` is a failed batch the caller should skip; `Err` is a failure to invoke
+    /// the fetcher at all.
+    fn genbank(&self, ids: &str) -> Result<Option<String>, FerroError>;
+
+    /// FASTA-format payload, used for CON records that carry no `ORIGIN` section.
+    ///
+    /// `None` for any failure: this is already a fallback, and its own fallback is to
+    /// leave the accession unfetched.
+    fn fasta(&self, ids: &str) -> Option<String>;
+
+    /// Delay observed between requests, to respect the endpoint's rate limit.
+    fn rate_delay(&self) -> std::time::Duration;
+}
+
+/// The production [`EfetchSource`]: NCBI's efetch endpoint, over `curl`.
+pub struct NcbiEfetch {
+    rate_delay: std::time::Duration,
+}
+
+impl NcbiEfetch {
+    /// Build a fetcher whose rate limit reflects whether an API key is configured.
+    pub fn new() -> Self {
+        // Rate limiting: 3 req/sec without API key, 10 req/sec with API key.
+        let has_api_key = get_ncbi_api_key().is_some();
+        if has_api_key {
+            eprintln!("  Using NCBI API key (10 req/sec rate limit)");
+        } else {
+            eprintln!("  No NCBI_API_KEY set (3 req/sec rate limit - set key for 3x faster)");
+        }
+        Self {
+            rate_delay: std::time::Duration::from_millis(if has_api_key { 100 } else { 350 }),
+        }
+    }
+
+    /// Run `curl` against an efetch URL, returning its stdout when it succeeded.
+    ///
+    /// Bounded in time deliberately. A provisioning run is hundreds of sequential batches
+    /// against a rate-limited endpoint, so a single stalled connection with no timeout
+    /// hangs the whole run indefinitely. `--max-time` is generous because one batch is up
+    /// to 200 GenBank records and some are megabytes.
+    ///
+    /// `--fail-with-body` makes an HTTP error an unsuccessful exit rather than a
+    /// success carrying an error page. Without it a 429 or 500 body parses to zero
+    /// records and the batch silently "succeeds" having fetched nothing; with it the
+    /// caller reports the batch as failed. Either way the accessions stay unfetched and
+    /// are retried on the next run, so this changes what the operator is told, not what
+    /// ends up on disk.
+    fn curl(url: &str) -> Result<Option<String>, FerroError> {
+        let output = std::process::Command::new("curl")
+            .args([
+                "-s",
+                "-L",
+                "--fail-with-body",
+                "--connect-timeout",
+                "30",
+                "--max-time",
+                "600",
+                url,
+            ])
+            .output()
+            .map_err(|e| FerroError::Io {
+                msg: format!("Failed to fetch batch: {}", e),
+            })?;
+        Ok(output
+            .status
+            .success()
+            .then(|| String::from_utf8_lossy(&output.stdout).into_owned()))
+    }
+}
+
+impl Default for NcbiEfetch {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl EfetchSource for NcbiEfetch {
+    fn genbank(&self, ids: &str) -> Result<Option<String>, FerroError> {
+        Self::curl(&build_efetch_url("nuccore", ids, "gb", "text"))
+    }
+
+    fn fasta(&self, ids: &str) -> Option<String> {
+        Self::curl(&build_efetch_url("nuccore", ids, "fasta", "text"))
+            .ok()
+            .flatten()
+    }
+
+    fn rate_delay(&self) -> std::time::Duration {
+        self.rate_delay
+    }
+}
+
 /// Fetch sequences from NCBI in GenBank format, extract CDS metadata, and write to files.
 ///
 /// This fetches GenBank format (not FASTA) to get CDS coordinates along with sequences.
@@ -1919,8 +2019,32 @@ pub fn fetch_fasta_to_file(
     output_file: &Path,
     dry_run: bool,
 ) -> Result<usize, FerroError> {
-    use std::io::Write;
-    use std::process::Command;
+    fetch_fasta_to_file_with(accessions, output_file, dry_run, &NcbiEfetch::new())
+}
+
+/// [`fetch_fasta_to_file`] against an arbitrary [`EfetchSource`].
+///
+/// # Presence, and why it is not the sidecar's question
+///
+/// Which accessions still need fetching is derived from the **FASTA**, via
+/// [`SupplementalStore::has_sequence`] — never from `*.metadata.json`. The sidecar records
+/// annotation; a row in it is not a sequence, and treating it as one is what let an
+/// accession whose bases were already on disk be re-fetched and appended a second time.
+///
+/// Fetched records are staged and then merged by accession, so no write path can append
+/// blind: re-fetching an accession that already has a record **replaces** it rather than
+/// duplicating it. That makes the resume predicate free to ask any question it likes —
+/// including PR #1732's "is this record usable?" — without the write path being able to
+/// duplicate anything as a result.
+///
+/// A run with nothing to fetch and nothing to repair writes no byte at all.
+pub fn fetch_fasta_to_file_with(
+    accessions: &[String],
+    output_file: &Path,
+    dry_run: bool,
+    source: &dyn EfetchSource,
+) -> Result<usize, FerroError> {
+    use crate::benchmark::supplemental_store::SupplementalStore;
 
     const BATCH_SIZE: usize = 200; // Larger batches for efficiency (4x fewer requests)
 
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

Issue #1630, opened 2026-08-11
normalize: enforce every absolute prohibition at the mode-gated parse stage the ruling decided


nh13:
Implements the operator ruling on `rulings[absolute-prohibition-enforcement-stage]`, which moves from `undecided` to `decided`.

## The ruling

The enforcement stage for an absolute prohibition is **mode-dependent**, uniformly across every such clause:

- **strict** — fail at **parse**. Strict means the input's conformance is validated, not merely that the input can be parsed.
- **lenient** — input conformance is **not** validated. A lenient run fails only when it cannot **normalize**, which can happen.
- **silent** — lenient with no error messages, but an exit code.

## The seam already exists, and two clauses are already on it

`parse_hgvs_with_config(input, ErrorConfig)` is the mode-aware parse entry, and `apply_bracket_cardinality_rule` (#493) is the precedent for applying a conformance rule there keyed on `ErrorConfig::action_for`. Two of the checklist's prohibitions already behave exactly as the ruling asks:

| input | strict | lenient / silent |
|---|---|---|
| `NC_TEST.1:g.266del3` | refused, "uses size-count suffix" | repaired to `NC_TEST.1:g.266_268del` |
| `NM_TEST.1:c.10_12 del` | refused, "Extra whitespace in HGVS description" | repaired to `NM_TEST.1:c.10_12del` |

So this is not new machinery. It is bringing the rest of the prohibitions onto machinery that ships.

## The measured gap

Measured over the 164 `RowKind::Prohibited` rows of `conformance::spec_corpus`, in all three `ErrorMode`s, parsing through `parse_hgvs_with_config` rather than the config-less `parse_hgvs`:

| clause | rows | strict | lenient | silent |
|---|---|---|---|---|
| `checklist.md:26` incomplete intronic range | 16 | parse | parse | parse |
| `checklist.md:31` single-anchor `insT` | 24 | parse | parse | parse |
| `checklist.md:16` `g.*10del` | 4 | parse | parse | parse |
| `checklist.md:49` `del3` | 24 | parse | **16 repaired**, 8 parse | same as lenient |
| `general.md:96` internal space | 24 | parse | **24 repaired** | same as lenient |
| `checklist.md:20` bare-`NM_` intronic | 16 | normalize | accept + W4007 | accept, no warning |
| `checklist.md:16` `g.266+2del` | 4 | **accept** | accept | accept |
[...]

PR #1792, opened 2026-08-13
fix(hgvs): refuse a `+`/`-` offset on a `g.`/`m.`/`o.` position

nh13:
`background/numbering.md` numbers the three genomic-family axes in three consecutive bullets and ends every one of them the same way — nucleotide numbers based on a **genomic** (`:6`), **circular** (`:8`) or **mitochondrial** (`:11`) reference sequence "**do not include** `+`, `-`, `*`, or other prefixes". `docs/recommendations/checklist.md:16` says it again for `g.` in the checklist's register, and `:45` supplies the shape that actually turns up: a range written with `-` where the separator is `_`, which it marks `Not correct`.

Ferro enforced one third of one of those clauses, and the third it enforced was an accident.

## The measurement, per mode and per entry point

Measured on `9fb126ba`, before this change:

| input | `parse_hgvs` (bare) | `parse_hgvs_with_config`, strict / lenient / silent | `normalize`, strict / lenient / silent |
|---|---|---|---|
| `NC_000023.10:g.100+2A>G` | accepted | accepted / accepted / accepted | **re-emitted verbatim**, all three |
| `NC_000023.10:g.100-2A>G` | accepted | accepted / accepted / accepted | **re-emitted verbatim**, all three |
| `NC_012920.1:m.100+2A>G` | accepted | accepted / accepted / accepted | **re-emitted verbatim**, all three |
| `NC_001416.1:o.100+2del` | accepted | accepted / accepted / accepted | **re-emitted verbatim**, all three |
| `NC_000023.10:g.100+?del` | accepted | accepted / accepted / accepted | **re-emitted verbatim**, all three |
| `NC_000023.10:g.*100A>G` | refused | refused / refused / refused | (does not parse) |
| `NC_000023.10:g.-100A>G` | refused | refused / refused / refused | (does not parse) |

**The `*` and leading-`-` column is a grammar accident, not enforcement.** Neither matches any production of the genomic position parser, so both are refused the way `g.@100` would be — in every mode, with a raw combinator error and no clause citation. Reading that as "the clause is partly enforced" is exactly what kept the other two thirds open. `the_star_and_leading_hyphen_forms_are_grammar_refusals` pins the distinction so it cannot be lost again.

Two shapes also parse today that a naive reading of the clause would not predict, and both are in scope:

- **`g.(100+1_101-1)_(200_300)del`** — a complex `(a_b)` boundary. `UncertainBoundary::inner()` returns `None` for a `Range`, so a walker built on it alone never sees these endpoints. Both existing genomic-offset guards — SPDI's and VCF's — are written that way and have that hole; the detector here enumerates both endpoints of a range.
- **`g.100+?del` / `g.100-?del`** — the unknown-offset sentinels. `:6` prohibits the *symbol*, not a magnitude, and there is still no anchor for an offset to be unknown relative to.

## Does ferro emit the form itself?

**It never manufactures one, and it did emit one.** Those are different claims and the PR rests on the second.

`GenomePos::with_offset` has exactly **one** caller in the crate — `parse_genome_pos` (`src/hgvs/parser/position.rs:72`) — so no projection, conversion or normalization path can mint a genomic offset. Every prohibited output was a *propagated* input, which is a lower severity than the 371 self-emitted descriptions `rulings[bare-transcript-intronic-position]` had to close, and it is stated rather than glossed.

What made it a rule-1 defect anyway is that `normalize` **re-emitted** the propagated spelling as a fixed point, in all three modes, with an **empty** warning vector — `normalize_genome_variant`'s offset-carrying short circuit at `src/normalize/mod.rs`. Normalization was not impossible there, it was **vacuous**, which is worse, because the output looks normalized while carrying a spelling the recommendations prohibit. Those re-emissions were the whole of `prohibition_violating_outputs`, and this change takes that counter to **0**.

## `m.` and `o.` are in scope, and that is not a widening

The clause is stated **per axis, three times**. `numbering.md:8` binds `o.` and `:11` binds `m.` in their own words, and `:9` additionally makes a mitochondrial reference "a special **circular** genomic reference sequence", so `m.` would inherit `:8` even if `:11` were absent. Refusing on `g.` alone would be the under-reading, not the safe one.

Ferro's own conversion layer had already reached that conclusion: `spdi::convert::reject_unresolvable_genomic_position` (#1641) and `vcf::from_hgvs::reject_unresolvable_genomic_position` (#1734) each refuse `g.`, `m.` and `o.` alike, citing `checklist.md:16`. The diagnostic here cites the clause governing the axis it fired on rather than a `g.`-only one, so an `m.` refusal quotes `:11`.

## The stage, and why there is no repair arm

`rulings[absolute-prohibition-enforcement-stage]` (decided 2026-08-10) governs and is followed exactly:

- **strict** refuses at **parse** — `apply_genomic_offset_rule` in `parse_hgvs_with_config`, the same seam as `W3026`, `W3028` and `W4008`, keyed on `ErrorConfig::action_for`;
- **lenient** accepts the input and warns `W4009`, **silent** accepts quietly, and both then fail at **normalize**;
- **normalize refuses in every mode**, unconditionally, in `Normalizer::normalize_core_checked` — because rule 1 of the README ruleset is about **output** conformance and has no mode escape. That asymmetry is the same one `W3028` documents: an `--ignore W4009` moves the parse stage and cannot make the description normalizable.

The bare `parse_hgvs` entry still accepts, because it applies no `ErrorConfig` at all (#1632). That is what makes the unconditional normalize half load-bearing rather than belt-and-braces: a caller who never touches a config must still not be handed a prohibited spelling back.

**There is no repair arm, and the absence is a decision.** Both available repairs invent a variant the caller did not describe. Dropping the offset answers for `g.266del` — a *different* nucleotide, and precisely the silent flattening #1641 and #1734 were filed to stop. Rewriting `g.266-268del` as the range `g.266_268del` turns a one-base deletion into a three-base one on nothing more than a guess about intent. `the_decided_target_is_a_mode_gated_refusal` admits either "normalize it to a conformant description" or "fail because it cannot", and with no derivable repair the honest arm is the second. This is a maintainer's choice under rule 6 of the `README.md` ruleset, disclosed under rule 7; it is stated here rather than recorded as a new ledger record because the governing record already decided the question this PR implements, and the acceptance test already admits both arms.

## What the counters do

`spec_conformance_axis`, both shuffle directions, re-blessed from the run rather than composed:

| counter | before | after |
|---|---|---|
| `prohibition_violating_outputs` | 8 | **0** |
| `prohibited_absolute_accepted` | 32 | **24** |

Everything else is byte-identical in both directions — `converged`, `split_two/three/more`, `non_idempotent_outputs`, `sequence_changed`, `conflicts_accepted`, `prohibited_conditional_accepted`, `guard_violations`. That is what a refusal of un-normalizable inputs should look like: rows leave the corpus's output set without any surviving row changing.

The 8 are 4 × `checklist.md:16-genomic-has-no-offsets` and 4 × `checklist.md:45-range-is-underscore` (each alone and inside a cis allele, on both cores). The residual 24 absolute acceptances are `checklist.md:33`'s `ins6`, which is #1627's remaining half.

## Real-world cost: 2 rows of 9,949,738

Scanned across the four committed corpora (`clinvar_hgvs_500k`, `clinvar_hgvs_unique`, `cmrg_genes_exhaustive`, `paraphase_genes_exhaustive`), verified against `tests/fixtures/CHECKSUMS.sha256` before counting:

| corpus | rows | `g.`/`m.`/`o.` axis | carrying an offset |
|---|---|---|---|
| `clinvar_hgvs_500k` | 500,004 | 166,471 | 0 |
| `clinvar_hgvs_unique` | 4,188,436 | 2,852,930 | **2** |
| `cmrg_genes_exhaustive` | 4,826,063 | 1,635,485 | 0 |
| `paraphase_genes_exhaustive` | 435,235 | 118,147 | 0 |

The zeros are **empirical, not structural**: the same probe finds 75,601 `c.`/`n.` offsets in `clinvar_hgvs_500k` alone, so the counter can see the class.

The two are `NC_000023.10:g.146993569-146993571[6]` and its `[5]` sibling — `checklist.md:45`'s hyphen-as-range, which ferro's grammar reads as base 146993569 with offset −146993571 and which today normalizes back out unchanged. Both are pinned by name in the new test's `OFFENDING` list, so the cost is asserted rather than described.

**One near-miss class deliberately not touched.** Six further rows spell a repeat count as a range (`NG_008047.1:g.17267CAG[(54-68)]`, `g.(?_197268413-1)_…del`). A text-keyed gate would refuse all of them; this check is AST-keyed on `GenomePos::offset`, so none is matched. The `CAG[(54-68)]` form does not parse at all today, for an unrelated grammar reason, and this change does not move it — that is recorded in the control test rather than left to be re-discovered.

## Fallout, in full

Eight tests failed on the first armed run and every one is inverted rather than deleted, so nothing leaves the record silently:

- **`corpus_prohibited_inputs`** — `one_clause_names_three_offset_symbols_and_ferro_refuses_one` becomes `the_clauses_three_offset_symbols_are_refused_alike` (pinned defect → adjudicated-correct guard); the two acceptance maps in `the_absolute_prohibitions_ferro_accepts_are_three_clauses_and_no_others` lose both `g.` entries, from **both** maps, which is how a mode-gated refusal reads from there; `every_prohibition_violating_output_is_a_re_emitted_prohibited_input`'s clause map is now **empty**.
- **`spec_corpus_regressions::absolute_prohibitions_are_accepted_and_re_emitted`** — the two `g.` rows are asserted as refusals in place, rather than dropped, so the row set does not appear to shrink.
- **`m_shift_matrix`** — `decorated_position_with_offset_routes_to_canonicalize_fallback` asserted that an `m.` offset *survived* normalization. Preserving it was itself the defect. The routing it pinned is still live and still needed (it is what stops the offset being silently dropped on a path reached by construction); what changed is that the description no longer gets that far.
- **`issue_393_mt_w4004`** — its `m.16570+1_16570del` fixture is no longer authorable. The `(base, offset)` endpoint-distinctness contract it pinned survives on the `c.`/`n.` axes, where an offset is legal, which is what its own comment meant by "mirrors the c./n. dedupe pattern"; `check_mt_pos_past_end`'s offset guard is now unreachable from any parsed description, and that is pinned in its own right.
- **`spec_conformance_axis`** — the two census pins above.

`mito_circular_audit`'s two `m.`-offset tests are **unchanged and still pass**: they call the bare `parse_hgvs`, which applies no mode. Their comments are corrected — the audit's open question is closed, and its citation of "`numbering.md` line 7" pointed at a line that has moved (the mitochondrial bullet is `:11` at spec checkout `6f85311`).

## Verification

```
cargo nextest run --features dev --lib --test it --no-fail-fast
Summary [111.028s] 10290 tests run: 10290 passed, 150 skipped
EXIT=0
```

No tests were left unrun. The new test's `no_mode_may_emit_the_prohibited_spelling` and `strict_refuses_a_genomic_offset_at_parse` were both watched failing on the unmodified tree first, for the right reason — strict accepted, and strict re-emitted `NC_000023.10:g.100+2A>G`.

`W4009` rather than the next free `W4008`, which #1751 took while this branch was in flight.

Closes the `checklist.md:16` / `checklist.md:45` half of #1630.

Representation-Change: none. 0 rows move and no stored string changes — what changes is an `Ok` becoming an `Err`. A `g.`/`m.`/`o.` position carrying a `+`/`-` offset is newly refused: at parse in strict (`W4009`), and at normalize in every mode. 8 of the spec corpus's 164 prohibited rows stop producing an output at all (4 × `checklist.md:16-genomic-has-no-offsets`, 4 × `checklist.md:45-range-is-underscore`), which takes `prohibition_violating_outputs` from 8 to 0 and `prohibited_absolute_accepted` from 32 to 24 in both shuffle directions; every other census figure is byte-identical. Across the four committed real-world corpora the cost is 2 of 9,949,738 — `NC_000023.10:g.146993569-146993571[6]` and its `[5]` sibling — and those zeros are empirical, not structural, since the same probe finds 75,601 `c.`/`n.` offsets in `clinvar_hgvs_500k` alone.


nh13:
@coderabbitai pause

Issue #1789, opened 2026-08-13
normalize: size-count edit forms are refused for del/dup but accepted for ins (and inv)


nh13:
## Summary

`checklist.md` forbids size-count edit forms — a number standing in for the sequence or the span. Ferro **refuses** them for `del` and `dup` and **accepts** them for `ins` and `inv`.

Measured with the `ferro` CLI at `origin/main` `9fb126ba`, default mode:

```
NC_000023.10:g.123del3        ERROR  Deletion '123del3' uses size-count
NC_000023.10:g.123dup3        ERROR  Duplication 'dup3' uses size-count
NM_004006.2:c.5439_5440ins6   ACCEPTED -> NM_004006.2:c.5439_5440ins6
NC_000023.10:g.123_125inv3    ACCEPTED -> NC_000023.10:g.123_125inv3
```

The `ins` half is directly clause-backed. `recommendations/checklist.md:33` (spec pin `6f85311`):

> Describing a variant as `c.5439_5430ins6` is **not allowed**, the inserted sequence (for `ins6`, e.g., `TGCCAT`) should be specified.

marked `class="invalid"`. Its sibling at `:49` is the `del` case, which ferro already enforces:

> Descriptions like `g.123del3` are **not allowed**, correct is `g.123_125del`

Because ferro round-trips `c.5439_5440ins6` unchanged, this is an **output**-conformance problem as well as an input one: a description the checklist marks invalid goes in and comes back out.

## Why this has stayed invisible

The spec's own example at `:33` carries an **inverted range** — `c.5439_5430ins6`, 5439 > 5430. Ferro rejects that string, but for the *range*, not the `ins6`:

```
NM_004006.2:c.5439_5430ins6   ERROR  Interval positions are swapped ('5439_5430'); expected start ≤ end
```

So the conformance corpus scores the row `correctly-rejected` and the `ins<number>` acceptance is masked by an unrelated rejection. The harvested example is the only one testing this clause.

Upstream has since corrected the range to `c.5439_5440ins6` (commit `14fcbf9`, PR #262, in the range `6f85311..565b973`). At that pin the row becomes a **false-acceptance**, and the enumeration records what would be the corpus's **first `must`-level invariant violation** — `output-invariant/B8-no-size-number-forms`, citing `checklist.md:49`.

**The defect is live at the current pin. Only its visibility depends on the submodule bump**, so this should be fixed on its own schedule rather than waiting for it.

[...]

Issue #263, opened 2026-05-16
fix(error-handling): wire W4001 SwappedPositions soft-validation warning (closes #81 L2 remaining)


nh13:
Wire the registered-but-unused **W4001** (`SwappedPositions`) soft-validation warning into the preprocessor.

## Background

#137 (#81 L1) audited the error-code registry and identified W4001 as **Registered** — `correct_swapped_positions` exists in `src/error_handling/corrections.rs` but the preprocessor's phases never invoked it. Empirically, `NM_000088.3:c.200_100del` produced a generic parse error in lenient mode with no W4001 emission.

The follow-up issues #124, #125, #127, #128 closed the rest of the SVA work, leaving W4001 (and W1004, W2004, W4002) as the remaining `registered` rows from the L1 audit.

## Scope

This PR wires **W4001 only**. The other three remaining `registered` rows are out of scope:

- **W1004 MixedCaseEditType** — needs a new full-string scanner; `correct_edit_type_case` is token-level only.
- **W2004 InvalidUnicodeCharacter** — generic catch-all; W2001/W2002 cover the real cases.
- **W4002 PositionZero** — already enforced under E1003; retagging is a behavior change.

## Changes

1. Add Phase 15a to the preprocessor that calls `correct_swapped_positions` and dispatches via `ErrorType::SwappedPositions` (`standard_correctable` mode behavior).
2. Tighten `detect_swapped_positions` to bail on offset-bearing forms (`c.100+5_99+3`, `c.100_*1`) — the bare-integer swap would silently drop the offset.
3. Reclassify the W4001 audit row from `registered` to `partial` (offset and `*N` forms still fall through).
4. Tests: `tests/issue_l2_w4001_swapped_positions.rs` — lenient emits W4001 and rewrites; `parse_hgvs_with_config(strict)` rejects; bare `parse_hgvs` matches the parser's own behavior (CDS rejects via nom Verify; genomic accepts silently — pinned as a separate gap).

## Test plan

- [x] `cargo nextest run --features dev` — 4530 / 4530 pass
- [x] `cargo clippy --features dev --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all` — clean

PR #262, opened 2026-05-16
test(audit): HGVS ↔ VCF conversion surface (closes #81 K2)

nh13:
Closes #261. Closes #81 § K2.

## Summary

Pure-coverage audit pinning ferro's no-provider HGVS → VCF conversion surface (\`genomic_hgvs_to_vcf\`). No source changes.

- **Substitution**: converts without reference data. Emits chr/pos/ref/alt.
- **Anchor-required edits** (del, ins): error with anchor-base message.
- **Unsupported by simple path** (inv, dup, delins): error.
- **Non-genomic variants** (c./r./p./m.): type-shape contract — parse to non-Genome variants, not eligible for the simple path. Full conversion uses \`HgvsToVcfConverter\` with a provider.
- **Accession → chromosome mapping**: chr1..chr22, chrX, chrY, chrM; GRCh37 chr1 accession version (\`NC_000001.10\`) also maps to chr1.

16 new tests, all passing.

This is a pure-coverage audit; the parallel-review ritual was skipped per workflow. The provider-aware path (\`HgvsToVcfConverter\`) is covered by unit tests inside \`src/vcf/from_hgvs.rs\` (e.g. \`test_converter_cds_snv\`); audit-shape coverage of the provider path is out of scope here.

## Test plan

- [x] \`cargo nextest run --features dev --test issue_261_hgvs_vcf_coverage\` — 16 / 16 pass
- [x] \`cargo clippy --features dev --all-targets -- -D warnings\` — clean
- [x] \`cargo fmt --all\` — clean

Issue #261, opened 2026-05-16
tracking: HGVS ↔ VCF normalization with anchor-base rules (closes #81 K2)


nh13:
Audit pinning ferro's HGVS → VCF conversion surface.

## Categories

**Supported without reference data** (`genomic_hgvs_to_vcf`):
- `g.<pos><ref>><alt>` substitution: returns `VcfRecord {chrom, pos, ref, alt}`
- Accession → chromosome mapping (`NC_000001.11` → `chr1`, `NC_000023.11` → `chrX`, `NC_012920.1` → `chrM`)

**Anchor-base required** (errors without reference):
- `g.del` (any form)
- `g.ins`
- `g.delins`

**Unsupported by simple path**:
- `g.inv` (needs reference to compute reverse complement)
- `g.dup` (needs anchor)
- non-genomic coords (`c.`, `n.`, `r.`, `p.`) — use `HgvsToVcfConverter` with provider instead
- Null/Unknown/Empty alleles
- Compound brackets

## Pinned

Each surface point with its current behavior. No source changes.

PR #1793, opened 2026-08-13
chore(spec): bump hgvs-nomenclature to 565b973 and repair the moved citations

nh13:
Representation-Change: none. Submodule bump and citation repair; 0 of 2317 corpus rows change any non-citation field.

Bumps `assets/hgvs-nomenclature` from `6f85311` to `565b973` (upstream #261, #262, #263 — typo and markup fixes across 30 files, 57 insertions / 58 deletions) and repairs every citation the bump invalidates.

## The one line-number move, verified rather than assumed

Upstream `14fcbf9` **de-duplicated `general.md`**: old line 22 was a verbatim duplicate of old line 17 (byte-identical, confirmed by hash), and the copy at 22 was removed. So `general.md` goes 174 → 173 lines, citations at old ≤21 are unaffected, old 22 no longer exists, and **old ≥23 shift by −1**.

`general.md` is the only spec file whose line numbers move. That is checked per file rather than inferred from the diffstat — every hunk in the bump was tested for balance, and there is exactly **one** unbalanced hunk in all 30 changed files:

```
UNBALANCED docs/recommendations/general.md  @@ -22,1 +21,0
```

Every other changed file is a balanced in-place substitution, so its line numbers are stable and its citations only break if they quote text that changed.

## What is repaired

Five independent gates, all now green:

| gate | citations | notes |
|---|---|---|
| `generate_spec_fixture` quote check | **43** | 25 ruling records + 1 equivalence class; 38 `general.md`, 4 `DNA/complex.md`, 1 `DNA/insertion.md`. It bails on the first failure, so the full set was found by replicating its whitespace-collapsed match over all 150 citations rather than by 39 build cycles |
| `generate_spec_fixture` role consistency | **7** | `governing` (5) and `deviates_from` (2) must name a clause that is in the record's `clauses`, so those move with it. `deviates_from` is a **list**, so its entries are bare array elements — a `"key": "value"` rewrite misses them, and the generator catches it |
| `generate_spec_conformance_corpus` (byte-exact) | **15** | all `general.md`, in the committed `spec_rule_inventory.json` |
| `case_harvest::every_row_cites_the_spec_verbatim` | **5** | |
| `spec_worked_example_rules::every_row_quotes_the_spec_verbatim` | **8** | W46, W50, W56–W59 |
| `normalization_stage_audit_doc::every_cited_clause_quote_is_in_the_spec_checkout` | **9** | `tests/fixtures/grammar/normalization_stage_audit.json`, added by #1855 after this branch was cut. 8 structured `clause` fields plus the one prose `general.md:157` in a `note`; that file carries **zero** bare-continuation references, so its prose is unambiguous and moves with its structured citations |

Not every failure is a line shift. Five are genuine upstream text corrections, and those needed the quote updated, not the line:

- `DNA/complex.md:127` and `:161` — an impossible chr22 coordinate, `410000000` corrected to `41000000` (4 quotes).
- `DNA/insertion.md:74` — `NM_004006.1` corrected to `NM_004006.2` in the worked example's alternative spelling. This also moves the **equivalence-class member**, since `generate_spec_fixture` rejects a member that is not a fixture row and the harvest now yields the `.2` spelling.

**The `general.md:22` citation is deleted, not shifted.** The inventory already recorded that `:22` was a verbatim duplicate of `:17`, "inventoried separately so the clause-unit denominator stays mechanical". Upstream removed the duplicate, so the clause unit is genuinely gone: the entry goes away, `:17` (unaffected, ≤21) carries the clause, and the pinned denominator follows mechanically — **`general.md` 86 → 85 clause units**. Every other one of the 43 pinned per-file counts holds.

`tests/it/clause_ruling_index.rs` was regenerated from `the_rendered_index_is_current`'s own output and spliced on its `//!` markers, never hand-edited. All 7 of its tests pass, and the diff is a symmetric 22 insertions / 22 deletions, so no row was dropped.

Two committed documents are **generated renders** and were re-blessed rather than hand-edited, each by its own command:

* `docs/NORMALIZATION_CONTRACT.md` — `BLESS_CONTRACT_DOC=1 cargo nextest run --features dev --test it -E 'test(normalization_contract_doc)'`. The bump-attributable part of its diff is three lines, the three ledger citations `main`'s new records carry.
* `docs/NORMALIZATION_STAGE_AUDIT.md` — `BLESS_STAGE_AUDIT_DOC=1 cargo nextest run --features dev --test it -E 'test(normalization_stage_audit_doc)'`. Pure renumbering, 13 insertions / 13 deletions.

Both were re-run **without** the bless variable afterwards and pass.

## Ferro's behaviour does not change

Of the **2317** enumeration rows common to both pins, every row that differs differs in **exactly one field, `spec_citation`** — 2266 in the SHA stamp only, 51 also carrying a `general.md` −1. **Zero rows differ in any other field.** That is what makes the trailer above honest, and it is measured, not assumed.

The row *set* does change, because upstream corrected five worked examples and the harvested input set follows: 23 rows leave, 21 arrive, total 2340 → 2338.

## Six pinned censuses move, at measured values

The normalization fixture's own census (`STATUS_CENSUS` in `hgvs_spec_normalization_tests.rs`) moves too, for the same reason: nine harvested rows leave and nine arrive, the 934-row total holds, and **no row common to both pins changes status**.

| census | was | now |
|---|---|---|
| `false-acceptance` | 52 | **53** |
| `preserved` | 714 | **713** |

The `false-acceptance` row is `c.5439_5440ins6` — the #1789 defect described below, arriving from `correctly-rejected` because the corrected range now parses. `preserved` drops by one because seven departures were `preserved` and only six arrivals are.

Then the enumeration's five:

| census | was | now |
|---|---|---|
| `projection-pinned` | 1154 | **1157** |
| `projection-unavailable-pinned` | 479 | **481** |
| `projection-error-pinned` | 220 | **215** |
| `mode-divergence-pinned` | 288 | **285** |
| `invariant-violation-must` | 0 | **1** |

The first four have one shared cause (the changed input set) and are commented as a group so they are read together. Net −3, plus +1 in the budget, for the −2 total.

### The `invariant-violation-must` row is a real defect, and is not flattened

`output-invariant/B8-no-size-number-forms/NM_004006.2:c.5439_5440ins6` — ferro accepts the size-number form `ins6`, which `checklist.md:33` marks `class="invalid"` by name:

> Describing a variant as `c.5439_5440ins6` is not allowed, the inserted sequence (for `ins6`, e.g., `TGCCAT`) should be specified.

**This is independent of the bump; the old pin was masking it.** At `6f85311` the spec's own example carried an inverted range (`c.5439_5430ins6`, 5439 > 5430) which ferro refuses for the *range*, so the `ins6` violation was never reached. Upstream corrected the typo, the range now parses, and the defect surfaces. It is tracked as #1789 and the budget is re-pinned at 1 with a comment saying not to suppress the row to restore a zero.

## Scope: gating citations only, and why

Roughly 660 further **prose** citations in `src/`, `tests/`, `examples/` and root docs go stale silently — nothing fails. I implemented the full repair, measured it, and then reverted it. The reason is not diff size:

**The mechanical `general.md:N → N−1` rule is not sufficient for prose.** This repo cites heavily in a bare continuation form — `general.md:34`, `:35`, `:56` — where a bare `` `:NN` `` inherits its file from an earlier qualified citation. There are **722** such bare references in the tree. Classified against the nearest preceding qualified citation: **454 belong to other spec files and must not move**, 172 look like `general.md`, and **96 have no antecedent at all** within 25 lines. The 172 and 96 are heuristic guesses, and unlike the gating citations they carry **no quote**, so nothing in the repo can verify them.

Shifting only the qualified refs produces demonstrably wrong output. The blind pass generated 65 self-inconsistent lines, including in this repo's own `CLAUDE.md`:

```
Every clause this project has litigated — `general.md:33`, `:35`, `:56`, `:58`,
```

— where `:33` moved and its bare companions did not. Restricting to files with no bare references at all captures only 191 of 628 refs and still misses the two biggest sites (`src/normalize/merge.rs`, 93 refs, and `CLAUDE.md`).

So the split is along a line that means something: **everything a machine can verify is repaired here; everything that needs a human to read the sentence is the follow-up.**

Concretely, the rule applied is: repair every citation a generator or test verifies, plus — only in files this PR already touches, and only where the file has **no** ambiguous bare-continuation reference — the fully-qualified `general.md` citations in adjacent prose, so that no row ends up with a `clause` and a `note` that disagree. `tests/fixtures/case-harvest/cases.json` (zero bare references) has five `clause` fields and five `note` fields moving together.

**The ruling ledger fails that test, so only its structured citations move.** `hgvs_spec_normalization_overrides.json` carries **195** bare references, and narrowing the antecedent search from the whole file to the enclosing prose *string* — the tightest scope that could be argued — still misattributes them. In `rulings[delins-merge-vs-individual-gap-two-or-more]`'s rationale the bare `` `:46` ``, `` `:47` `` and `` `:17` `` are **`DNA/delins.md`** references, but the last qualified citation before them in that same string is a `general.md` one, so an in-string rule would shift all three and corrupt the record. That is the decisive counter-example: bare references are not resolvable by *any* positional rule, only by reading. The ledger's `clause`, `governing` and `deviates_from` fields are repaired here; its prose joins the follow-up.

**What the follow-up owes:** re-point the ~628 qualified prose `general.md` citations at N≥23, and resolve the 722 bare `` `:NN` `` references by reading each one's antecedent — including the two files this PR deliberately leaves in a mixed state, `src/conformance/spec_corpus.rs` (13 rule tags shifted, 27 prose refs not) and `tests/it/spec_worked_example_rules.rs` (8 structured `clause:` fields shifted, 14 prose refs not). Filed as #1794, which records the bare-reference measurement so the next spec bump does not re-derive it.

Also left alone, deliberately: `CHANGELOG.md`'s two references, which are in released sections and describe decisions as they stood; and the `410000000` / `NM_004006.1` strings used as *parser test inputs*, which are test data rather than citations — changing a coordinate there would change what the test exercises.

## Verification

- All five gates green; `cargo fmt --all --check` and `cargo clippy --profile soak --all-features --all-targets -- -D warnings` both clean (captured exit status 0 each).
- Full suite with `--no-fail-fast` under the `soak` profile, rebased onto `origin/main`: **11063 run, 11063 passed, 0 failed, 151 skipped**, no "tests were not run" line, no cancellation, `EXIT=0` read back out of the run's own log.
- `Representation-Change` checker run against the real diff: **0 of 13 changed files** are under a `WATCHED_PREFIXES` directory (the six are `src/normalize/`, `src/hgvs/`, `src/spdi/`, `src/project/`, `src/reference/`, `src/error_handling/`), so no declaration is required; the trailer above is a voluntary `none`. `src/conformance/` is deliberately not watched.
- Rebased past the ruling-ledger commits on `main`. Verified by the ledger-record set comparison rather than by commit reachability: **31 records on the pre-rebase tip, 32 on `main`, 32 on this branch, none dropped from either side and no status changed.** `main` gained one record, `whole-span-reverse-complement-types-as-inv` (#1839), whose two `general.md` citations this rebase repairs — the same repair this PR exists to do, on a record that landed after it.


nh13:
@coderabbitai pause

PR #1721, opened 2026-08-12
fix(hgvs): refuse an alignment-only symbol on the r. axis

nh13:
> [!NOTE]
> **Ruled 2026-08-12: REFUSE.** `rulings[rna-axis-alignment-only-symbol-reach]` is now `decided`, with `docs/background/standards.md:47-61` — the RNA symbol table — **governing**. There is no `deviates_from`: nothing is being deviated from. The ruling, and the reading it was weighed against, are in **[The adjudication](#the-adjudication)** below and in the record itself.

#1684 closed the `standards.md:39` alignment-only-symbol hole on the DNA axes and **deliberately left the `r.` axis open**, pinning the residue as `src/hgvs/alignment_symbols.rs::a_lowercase_rna_masked_nucleotide_is_still_accepted`. This is that half.

## The defect, reproduced on `2f4e3bb9`

Measured through `parse_hgvs_with_config` in each of the three modes, then `Normalizer::normalize`, on a single-exon `cds_start = 1` transcript:

| input | parses to | strict | lenient | silent |
|---|---|---|---|---|
| `r.10delinsacgu` | `Literal([A, C, G, U])` | `r.10_11inscgu` | `r.10_11inscgu` | `r.10_11inscgu` |
| `r.10delinsacgux` | `Named("acgux")` | **re-emitted**, `w=[]` | **re-emitted**, `w=[]` | **re-emitted**, `w=[]` |
| `r.10delinsacxgu` | `Named("acxgu")` | **re-emitted**, `w=[]` | **re-emitted**, `w=[]` | **re-emitted**, `w=[]` |
| `r.10_11insacgux` | `Named("acgux")` | **re-emitted**, `w=[]` | **re-emitted**, `w=[]` | **re-emitted**, `w=[]` |
| `r.[10delinsacgux;12del]` | — | **re-emitted** | **re-emitted** | **re-emitted** |
| *control:* `c.10delinsACGTX` | `Named("ACGTX")` | refused `W3028` | refused `W3028` | refused `W3028` |

One stray byte reclassifies an otherwise-literal RNA run **wholesale** as a mobile-element name — the same `is_iupac_base` arm of `parse_inserted_sequence` that made every lone-`X` count in #1627 an undercount. Normalization is not impossible there, it is **vacuous**, which is worse: the output looks normalized while carrying a spelling `general.md:50` does not admit.

The cheapest reproduction needs no new instrumentation, because `main` already pins it:

```bash
cargo nextest run --features dev --lib \
[...]

PR #1797, opened 2026-08-13
fix(hgvs): refuse an insertion sized by a count instead of its sequence

### LINKED ISSUE #1627: normalize: a member ferro cannot compute is re-emitted verbatim, so 48 rows produce non-conformant output in every mode

### LINKED ISSUE #1789: normalize: size-count edit forms are refused for del/dup but accepted for ins (and inv)

nh13:
Closes #1789.
Closes #1627. (The `checklist.md:33` size-count insertion half of #1627 — the other half, `standards.md:39` / `delinsX`, was already closed by #1684 and #1721. Measured against this PR's head: 24 measurement points flip `Ok` to `Err(W3029)`; see the comment on #1627 for the runs and controls.)

## The defect

`checklist.md` item 3 ("Insertions") has two bullets. Ferro enforced the first and not the second, and the second's sibling item 6 (`del`/`dup`) has been enforced since #1079 — so one numbered checklist item was half-enforced.

`recommendations/checklist.md:32`–`:33`, spec pin `6f85311`:

> - **do you provide the inserted sequence?**
>   Describing a variant as <code class="invalid">c.5439_5430ins6</code> is not allowed, the inserted sequence (for `ins6`, e.g., `TGCCAT`) should be specified.

Because ferro round-tripped `c.5439_5440ins6` unchanged, this was an **output**-conformance problem and not only input hygiene.

Both axis clauses state the rule constructively, which is what scopes it to `ins` rather than to every numeric payload. `DNA/insertion.md:22` and `RNA/insertion.md:20` enumerate what an inserted sequence may be — the nucleotides inserted, a range of the same reference, a range of another reference — and neither list contains a bare count. `DNA/insertion.md:119` reaches the same verdict about a real reported description: "`c.23ins24` is not correct since the position of the insertion is not described properly and because \"ins24\" does not define the sequence inserted."

**Nothing is lost by refusing.** When the *number* of inserted nucleotides is known and their identity is not, the spec's own spelling is `insN[6]` (`DNA/insertion.md:77`) or `insn[6]` on `r.` (`RNA/insertion.md:41`, `general.md:50`). Ferro already parses, normalizes and re-emits both, and the diagnostic names the one appropriate to the axis.

## Per-mode behaviour, before and after

Measured with the library entries (`parse_hgvs_with_config` + `Normalizer::normalize`) rather than through the CLI — see the note at the end.

| input | strict | lenient | silent |
|---|---|---|---|
| `g.123del3` (before and after) | refused at **parse**, W3011 | repaired to `g.123_125del`, W3011 | repaired, quiet |
| `g.123dup3` (before and after) | refused at **parse** | refused at parse (no repair exists) | refused at parse |
| `c.10_11ins6` **before** | accepted, re-emitted | accepted, re-emitted | accepted, re-emitted |
| `c.10_11ins6` **after** | refused at **parse**, W3029 | accepted at parse (W3029 warning), refused at **normalize** | accepted at parse quietly, refused at **normalize** |

`del3`'s schedule is already what `rulings[absolute-prohibition-enforcement-stage]` asks for, and `dup3` fails in lenient on that record's own ground — there is no repair, so lenient fails because it cannot normalize. Neither is copied blind; `ins6` is put on the same schedule the record names, and the record lists `checklist.md:33` among its own clauses.

The normalize rung is deliberately **not** mode-gated, on that record's reasoning: the mode gate governs whether the *input* is judged and "does not, and cannot, govern whether the output conforms". The record also measured this exact shape and predicted the outcome — the `ins6` rows "genuinely denote nothing" yet all three modes emitted them back byte-identically with an empty warning vector, so "normalization is not impossible here, it is VACUOUS, which is worse, because the output looks normalized."

## Scope

**In scope:** `NaEdit::Insertion` / `BreakpointInsertion` whose payload is `InsertedSequence::Count` — `ins6` and `ins(6)`, which are two parser arms reaching one AST node.

The predicate is AST-keyed, never a scan of the rendered description. A textual `ins<digits>` check would have caught one of those two spellings and missed the other, and would have false-positived on `c.849_850ins858_895` — the range form `DNA/insertion.md:22` explicitly sanctions, which is also `ins` followed by digits.

**Out of scope, each with its reason recorded on `src/hgvs/insertion_size_counts.rs` and pinned by a test:**

- `ins(10_20)` — `InsertedSequence::Range` is ferro's node for both a count range and an uncertain *position* range, and the second is spec-sanctioned (`g.?_?ins(23632682_23625413)_(23625324_23619334)`, `DNA/complex.md`). Refusing the node would risk refusing a conformant description.
- `delins<number>` — `checklist.md:33` sits under item 3, "Insertions"; `DNA/delins.md` states no equivalent prohibition (its `class="invalid"` notes at `:29`/`:34` are about restating the *deleted* sequence).

## `inv3` — investigated, and not folded in

`g.123_125inv3` is accepted, and it is **neither** a parser laxity **nor** a live conformance defect. Three measurements:

1. **Not a trailing token being absorbed.** `NaEdit::Inversion` has a first-class `length: Option<u64>` field, documented "Explicit length of inverted region (e.g., `inv3`)", and `parse_inversion` fills it deliberately.
2. **No clause reaches it.** `checklist.md:49` is about naming "the first and last residue involved in a deletion", and `g.123_125inv3` names both endpoints already — the number is redundant, not a substitute for the range. That is the class ferro already treats as soft (`validate_no_point_size_suffix` says so of `c.100_102del3`), and `DNA/inversion.md` states no size-suffix prohibition.
3. **The output already conforms.** Normalization *drops* the count, in every mode — which is exactly the rung `checklist.md:33` needed and did not have. The single-position spelling `g.123inv3`, which *would* be `:49`'s shape, is already refused in every mode ("an inversion covers more than one nucleotide by definition"), and a mismatched size (`g.123_125inv7`) is caught in strict by W3016.

Pinned by `an_inversion_sized_by_a_number_is_a_different_shape`. No follow-up issue is filed, because there is nothing left that a clause supports.

## The regression test does not depend on the spec's example

This is why the defect survived. The spec's own `:33` example carries an **inverted range** — `c.5439_5430ins6`, 5439 > 5430 — so ferro refused it for the *range* and the conformance corpus scored the row `correctly-rejected`, masking the `ins6` acceptance. Upstream corrected the range in `14fcbf9`, which sits in the unmerged `6f85311..565b973` submodule range (a separate PR).

`tests/it/issue_1789_insertion_size_count.rs` uses valid-range inputs of its own throughout, and `the_guard_does_not_depend_on_the_spec_example` asserts the property directly: the spec's example must be refused for a reason that is *not* W3029, and no input in the file may be that example. So the guard holds at both spec pins and cannot be masked again in either direction.

## Blast radius

Measured over the four committed corpora (`clinvar_hgvs_500k`, `clinvar_hgvs_unique`, `cmrg_genes_exhaustive`, `paraphase_genes_exhaustive`), by extracting every `ins`-plus-digit candidate and normalizing it through a pre-change and a post-change binary:

| corpus | newly refused | accession-bearing rows |
|---|---|---|
| clinvar_hgvs_500k | 1 | 706,172 |
| clinvar_hgvs_unique | 36 | 4,345,946 |
| cmrg_genes_exhaustive | 5 | 6,931,313 |
| paraphase_genes_exhaustive | 2 | 647,476 |
| **total** | **44** | **12,630,907** |

39 unique strings, all previously accepted and re-emitted verbatim; **0** of them were already failing at the baseline. They are ClinVar's legacy `U14680.1` / `U43746.1` BRCA1 rows plus a handful of `NM_`/`NC_`/`LRG_` entries such as `NM_000497.3:c.841_842ins11` and `LRG_130:g.152649_152650ins337`.

The conformance corpus moves one census counter, down: `prohibited_absolute_accepted` 32 -> **8** in both directions, by exactly the 24 `ins6` rows `the_absolute_prohibitions_ferro_accepts_are_three_clauses_and_no_others` had named as the population a fix would take. `prohibition_violating_outputs` deliberately does **not** move (that counter's per-clause decomposition is entirely `checklist.md:16`; the `ins6` rows were never in it). Every rank-1, rank-2, idempotency and sequence-preservation figure is unchanged, which is what refusing un-denotable inputs should look like.

## Tests

New `tests/it/issue_1789_insertion_size_count.rs` — six integration tests plus nine unit tests on the predicate. Written first and watched fail for the right reason: strict parse returned `Insertion { sequence: Count(6) }` instead of refusing.

Existing guards that flipped from PINNED DEFECT to adjudicated-correct, in place rather than deleted: `corpus_prohibited_inputs::one_checklist_insertion_bullet_is_enforced_and_its_neighbour_is_not`, `spec_corpus_regressions::absolute_prohibitions_are_accepted_and_re_emitted` and `::a_prohibited_member_inside_an_allele_is_also_accepted`. Three tests that used `ins6` merely as a stand-in for "a non-literal payload" now use `insN[10]` / `ins(6_9)`, with the swap explained at each site.

Full suite: **10484 run, 10484 passed, 0 failed, 152 skipped**, `cargo nextest run --features dev --no-fail-fast`, exit 0, on the rebased branch. `cargo fmt --all --check`, `cargo clippy --features dev --all-targets -- -D warnings` and `cargo check --features python` all clean.

## One thing found and deliberately not fixed here

`ferro parse --error-mode strict` does not apply the post-parse AST conformance rules at all — verified against the *existing* `W3028` and `W4008`, both of which the CLI also fails to enforce. That is the `parse_hgvs`-applies-no-mode family (#1632), it predates this change, and it affects W3029 exactly as it affects its two siblings. `ferro normalize` is unaffected and refuses in every mode.

Representation-Change: 44 of 12,630,907 committed-corpus rows are newly refused. 39 unique strings, all previously accepted and re-emitted verbatim, so this is a real migration for a consumer round-tripping ClinVar. The conformance corpus's `prohibited_absolute_accepted` moves 32 to 8 in both directions; no output string changes, and no previously-refused input is newly accepted.


nh13:
@coderabbitai pause

Issue #269, opened 2026-05-17
design: W4002 PositionZero is registered but never emitted (rejection happens under E1003) (#81 L2 follow-up)


nh13:
**Open design question** — needs a decision before implementation.

## Background

#137 (#81 L1) classified W4002 (`PositionZero`) as **Registered** with a unique observation: the rejection of `c.0A>G` *does* happen at runtime — `detect_position_zero` runs in preprocessor Phase 1 and returns a failed `PreprocessResult` — but the diagnostic surfaces under `ErrorCode::InvalidPosition` (E1003), not under W4002. The W4002 `ErrorType` variant is reachable only via the warning-code dispatch tables (`config.rs`, `bin/ferro.rs`); no emission path tags rejections with this specific code.

## Decision space

Three plausible policies:

1. **Retag**: change `detect_position_zero`'s failure path to emit W4002 instead of E1003. Cleaner separation (W-codes for SVA, E-codes for hard errors). **Downside**: external callers matching on E1003 for `c.0` inputs would break.
2. **Emit both**: surface E1003 with W4002 as a sibling/related code in the diagnostic. Preserves backwards compat. **Downside**: two codes for one symptom muddies the registry contract.
3. **Deprecate W4002**: remove it from the registry and ErrorType enum, document E1003 as the canonical `c.0` code in the audit. Cleanest. **Downside**: a registered code disappears from public API.

## Recommended

Option 3 (deprecate W4002) — the L1 audit already calls W4002 "a duplicate identity that no caller emits". The W- vs E- split is a documentation artifact; `c.0` is genuinely an unrecoverable error, not a soft-validation warning.

## Out of scope

Until a decision lands, no code change. This issue tracks the design discussion.

Cross-ref: #264 (W4001 wiring deferred W4002 explicitly), #137 (L1 audit).

PR #1799, opened 2026-08-13
fix(normalize): refuse a c./r. position on base 0 instead of answering it as -1

### LINKED ISSUE #1796: c.0 has four live answers below the parser: which predicate decides that base 0 names no nucleotide, and at which layer?
`background/numbering.md:31` states, inside the definition of the `c.` numbering axis itself and alongside `-1` and `*1`: *"there is no nucleotide `c.0`."* It is an existence claim, not an input-hygiene rule — the coordinate names no nucleotide, so there is nothing downstream can convert it *to*.

The parser agrees and refuses `c.0` as `E1003 InvalidPosition` (#269). Below the parser, `c.0` is representable — `CdsPos.base` is a plain `i64` — and **four sites answer it, each differently, with no recorded authority for any of them**. This issue is one question, not four: *which predicate decides that base 0 names no nucleotide, and at which layer?* Filing them separately would guarantee they are settled separately and inconsistently, which is how the tree got here.

Line numbers below are against `origin/main` at `013bc66c`.

## The four faces

| site | answers `c.0` as | code |
|---|---|---|
| `CoordinateMapper::cds_to_tx` — `src/convert/mapper.rs:90`, predicate at `:108` | **`c.1`** | `pos.base < 1` takes the 5'UTR branch with a zero displacement, and the branch short-circuits to `cds_start` |
| `Normalizer::cds_to_tx_pos` — `src/normalize/mod.rs:11755`, arm at `:11781-11785` | **`c.-1`** | `Ok(cds_start.saturating_sub(1))`, under a comment about "historical inputs" |
| `Normalizer::rna_to_tx_pos` — `src/normalize/mod.rs:9000`, arm at `:9022-9024` | **`r.-1`** | byte-identical `Ok(cds_start.saturating_sub(1))`, **with no comment at all** |
| `cds_pos_to_tx_boundary` — `src/normalize/mod.rs:144`, arm at `:159-164` | **`c.-1`** | a fourth copy of the same expression, whose own comment declares it dead code |

`CoordinateMapper::cds_to_protein` (`src/convert/mapper.rs:174`, `:175`) is a fifth answer and at least an honest one: it refuses `pos.base < 1` outright.

### `rna_to_tx_pos`'s copy is the dangerous one

It carries **no comment whatsoever**. Its sibling `cds_to_tx_pos` at least documents itself (wrongly — see below), and `cds_pos_to_tx_boundary` documents itself as dead. `rna_to_tx_pos` sits 2,700 lines from its `cds_` twin and its doc says only *"Mirrors `cds_to_tx_pos`"*, so it reads as an unremarkable fallthrough arm. It is the copy most likely to be "cleaned up" into something wrong — folded into the `pos.base < 0` arm above it, or deleted as unreachable — by someone who has read neither the sentinel collision below nor the load-bearing measurement below that.

That is also why it was missed the first time this was written up: the finding was recorded as *two* live implementations. There are four.

## m10 — the same question, one file, 371 lines apart

Two "is this a 5'UTR position" predicates in `src/convert/mapper.rs` disagree at exactly one value:

```rust
// src/convert/mapper.rs:108   — cds_to_tx
} else if pos.base < 1 {                 // base 0 IS 5'UTR

// src/convert/mapper.rs:487   — fold_non_intronic_utr_offset
let in_utr = pos.utr3 || pos.base < 0;   // base 0 is NOT 5'UTR
```

This is not a separate defect; it is the direct consequence of base 0 having no settled meaning, so each site picked one. Whatever this issue decides must be implemented by both, preferably through one shared predicate rather than two open-coded comparisons.

## Why this cannot be a one-line refusal — the sentinel collision

`c.0` and `c.?` are the same value in the same type, so *"refuse `base == 0`"* and *"refuse `c.?`"* are the same statement:

```rust
// src/hgvs/location.rs:151
pub const CDS_BASE_UNKNOWN: i64 = 0;

// :181
pub fn is_unknown(&self) -> bool {
    self.base == CDS_BASE_UNKNOWN && !self.utr3 && self.special.is_none()
}
```

`CdsPos::Display` (`:282`) renders base 0 as `?`. A hypothetical `c.0` and a genuine `c.?` produce a byte-identical `CdsPos` and print identically. Special positions (`pter`/`qter`/`cen`) are a third occupant of the same value.

## THE TRAP — one of these arms is load-bearing, and removing it goes green

**Measured on `main`.** The `cds_to_tx_pos` arm is not vestigial and is not reachable only by direct construction. It is reached from an **ordinary parsed cis allele**:

`NM_TEST.1:c.[1A>G;1dup]` normalizes to `c.-1dup` today — but only *because of the arm*. `collapse_overlapping_cis_edits` computes the 5' insertion anchor as `c.1 - 1 = c.0` and builds the intermediate member `c.?_1insG`; `cds_to_tx_pos` then maps base 0 to `cds_start - 1`, which happens to be the correct transcript coordinate for `c.-1`, silently repairing the producer's off-by-one.

**Make that arm return `Err` on today's `main` and the same input emits `NM_TEST.1:c.?_1insG` to the user — with ZERO tests going red.** Measured: 10,238 tests pass both with the arm and with it replaced by an error. Exactly one test reaches the arm at all (`it::issue_1284_transcript_axis_reparse::no_transcript_axis_two_member_allele_fails_to_reparse`), and it asserts only that the output re-parses — which `c.?_1insG` does.

So the obvious fix ships a broken string and looks safe doing it. **#1777 is the prerequisite**: it fixes the producer (#1772) so the anchor is named `-1` rather than `0`, after which nothing in the suite reaches the arm. The normalizer half must be stacked on #1777, not opened against `main`.

Note the whole-suite instrumentation can only speak for paths the suite exercises. Any *other* producer that manufactures a base-0 position and has not been found is also currently being repaired by these arms, silently.

## What the external authorities say — all four agree

- **The spec**: `background/numbering.md:29-31`, quoted above. An existence claim.
- **biocommons/hgvs**: `BaseOffsetPosition.validate()` raises on `base == 0` — *"BaseOffsetPosition base may not be 0"* — enforced **in the position type**, not in the parser.
- **Mutalyzer**: rejects. `mutalyzer-crossmapper`'s introduction opens *"There is no zero (0) in any of the HGVS numbering systems"* and names this exact hazard — that `-1` and `1` being adjacent *"gives rise to various off by one errors when the conversion is not done properly."* Its answer is to convert out of HGVS numbering entirely rather than do arithmetic in a system with a discontinuity in it.
- **VariantValidator**: built on biocommons/hgvs and inherits its validation. (Inferred from the dependency, not separately measured.)

Neither reference corpus in this repo contains a single `c.0` case — 0 of 346 in `tests/fixtures/mutalyzer-normalize/cases.json`, 0 of 88 in `biocommons-normalize/cases.json`.

And ferro has already converged everywhere *except* these conversion sites: `normalize/merge.rs` says *"no `c.0` / `n.0` position exists"* and *"`-1 + 1` is `c.0`, which does not [exist]"*; `service/handlers/effect.rs` says *"because HGVS skips `c.0`: `c.-1_1` covers 2 bases"*.

Per the ledger's `adjudication-precedence-order`, spec-explicit, Mutalyzer and our own judgement all point the same way.

## The premise the `cds_to_tx_pos` comment rests on is false

```rust
// c.0 is not a valid HGVS position, but historical inputs
// can land here. Preserve the legacy mapping (treat as the
// last 5'UTR base, equivalent to c.-1) rather than failing.
```

Commit `d0607f3` (#97) fixed a real 5'UTR off-by-one and froze the pre-fix `base == 0` result as a "legacy mapping". The parser now rejects `c.0`, so no *historical input* can land here — what lands here is ferro's own intermediate, which is the opposite situation and calls for the opposite treatment.

## Deadline: once #1747 merges, the crate holds two contradictory answers

#1747 refuses `c.?` in `cds_to_tx` with `FerroError::ConversionError` — *"cannot resolve c.? to a transcript position: the CDS base is unknown"*. It is open and touches the mapper half only (its `normalize/mod.rs` hunk is a comment). The moment it lands, `src/convert/mapper.rs` refuses base 0 while `src/normalize/mod.rs` answers it as `c.-1`, three times over, in the same binary. That is worse than today's state, because today at least the disagreement is undocumented rather than deliberate on one side.

## Scope

- **Normalizer half** — the three `src/normalize/mod.rs` arms. Stacked on #1777; that is not optional, see the trap.
- **Mapper half** — #1747, in review.
- **m10** — the two predicates in `mapper.rs`, which belong with the mapper half.
- A ruling record for `c.0` belongs in `hgvs_spec_normalization_overrides.json`; that is owned separately and is not part of the code change.

## References

#1777 (prerequisite, open) · #1747 (mapper half, open) · #1746 (M9 — `cds_to_tx` answers `c.?` with `cds_start`, open) · #1745 (`get_amino_acid_at` underflow, open) · #1772 (the producer) · #1690 (`cds_to_tx_exon_aware` overflow) · #269 (parser refusal) · #97 / `d0607f3` (where the fork was created)

nh13:
Closes #1796 — the normalizer half only. The mapper half is #1747, in review; this PR does not touch `src/convert/mapper.rs`, and it touches neither the ruling ledger nor any census pin.

## Stacked on #1777 — read this before rebasing

**Base branch is `fix/allele-insertion-anchor-c-minus-one` (#1777, head `11338b19`), not `main`.** That is a correctness prerequisite, not a convenience, and it is the whole trap in #1796.

Once #1777 squash-merges, this needs:

```
git rebase --onto <#1777-squash-commit> 11338b19
```

and then retargeting to `main` before it can enter the merge queue. A plain `git rebase origin/main` replays #1777's commit and conflicts.

## What the three arms did

`background/numbering.md:31` states, *inside* the definition of the `c.` numbering axis and alongside `-1` and `*1`: *"there is no nucleotide `c.0`."* An existence claim, not an input-hygiene rule — the coordinate names no nucleotide, so there is nothing to convert it to.

Three sites in `src/normalize/mod.rs` answered it anyway, all with `cds_start - 1` — which is the axis's own `-1`, and therefore a *different* position from the one asked about:

| site | before | after |
|---|---|---|
| `Normalizer::cds_to_tx_pos` | `Ok(cds_start.saturating_sub(1))`, under a comment about "historical inputs" | `Err(ConversionError)` |
| `Normalizer::rna_to_tx_pos` | byte-identical arm, **no comment at all**, 2,700 lines from its twin | `Err(ConversionError)` |
| `cds_pos_to_tx_boundary` (free helper) | `Some(cds_start.saturating_sub(1))`, self-documented as dead code | `None` |

The stated premise was false. The parser refuses `c.0` outright (#269), so no *historical input* reaches these arms; what reached them was ferro's own intermediate.

Vocabulary is matched to #1747's mapper-side refusal deliberately, so the two halves agree rather than each inventing one: same error type (`FerroError::ConversionError`) and the same message shape (`cannot resolve <axis>.<pos> to a transcript position: …`).

### The self-declared-dead arm was kept, not deleted

It is genuinely unreachable — `check_cds_pos_past_end` short-circuits on `is_unknown()` and `is_special()` — and the reason it exists is parity with `cds_to_tx_pos`, so parity is what changed it. `None` is already this helper's vocabulary for "not answerable" and its caller's contract for the conservative skip. Deleting the arm would be worse in both directions: `base == 0` would fall into the final branch and evaluate `cds_start + 0 - 1` in `u64`, underflowing on the exact value at issue, and a helper whose whole justification is mirroring `cds_to_tx_pos` would silently stop mirroring it.

### Why the predicate is `base == 0` and not `is_unknown()`

They coincide at every reachable call site — `normalize_cds` resolves or refuses every special position before converting. But `CdsPos::is_unknown()` is false when `special` is set, and `pter`/`qter`/`cen` carry `base == 0` too, so keying on it would drop a marker into the final branch and underflow there. One predicate covers both occupants of the value, and no arithmetic runs on something that is not a coordinate.

## The trap, and the measurement that clears it

These arms were **load-bearing**. `collapse_overlapping_cis_edits` computed a 5'-edge insertion anchor as `c.1 - 1 = c.0` and built the intermediate member `c.?_1insG`; `cds_start - 1` happens to be the correct transcript coordinate for `c.-1`, so the `c.` arm silently repaired that producer's off-by-one.

Refuse without fixing the producer and `NM_TEST.1:c.[1A>G;1dup]` emits the malformed `c.?_1insG` to the user — **with zero tests going red**, because the one test that reached the arm asserted only that the output re-parses, and `c.?_1insG` does.

That producer is #1772, fixed by #1777. **Verified on this branch, with both in:**

| input | on this branch |
|---|---|
| `NM_TEST.1:c.[1A>G;1dup]` | `NM_TEST.1:c.-1dup` — unchanged |
| `NM_TEST.1:r.[1a>g;1dup]` | `NM_TEST.1:r.-1dup` — unchanged |
| `NM_TEST.1:n.[1G>A;1dup]` | `NM_TEST.1:n.-1_1insA` — unchanged (control: `n.` never had a repairing arm) |

`the_five_prime_edge_cis_collapse_is_unaffected` pins all three. It is not a change detector for this diff — it is the guard that the *stack* is right, and it is the most important assertion in the PR.

Note the whole-suite instrumentation can only speak for paths the suite exercises. Any other producer manufacturing a base-0 position that has not been found was also being repaired by these arms, silently, and now will not be.

## Tests, and proof they fail without the change

Reverting only the three arm bodies while keeping every test turns **7 of 10 red**:

- `cds_to_tx_pos_refuses_a_zero_base`, `…_carrying_an_offset`, `…_refuses_a_special_marker_rather_than_underflowing`
- `rna_to_tx_pos_refuses_a_zero_base`
- `cds_pos_to_tx_boundary_declines_a_zero_base`
- `a_cds_position_on_base_zero_is_not_relabelled_as_minus_one`
- (plus the `r.` end-to-end case, since removed — see below)

The 3 that stay green are the intended controls: the stack guard above, an anti-widening check that every position either side of the absent coordinate still converts and still shuffles, and `the_gap_the_axis_skips_is_still_a_gap`. Without those, the refusal tests are satisfiable by a normalizer that has simply stopped converting `c.`/`r.` positions at all.

The end-to-end behaviour that actually moves is a directly-built base-0 position — the shape the Python bindings and any other API consumer can construct, since the parser cannot: `NM_TEST.1:c.?del` normalized to `c.-1del` and now stays `c.?del`. `CDS_BASE_UNKNOWN` **is** the integer `0`, so this was doubly wrong: a position declared *unknown* was answered with a specific coordinate.

**One test was dropped rather than the refusal weakened, and it is worth knowing why.** An `r.` end-to-end case trips `FERRO_ASSERT_REPARSE` — *"normalization was handed an input it cannot re-parse"* — because `RnaPos` has no unknown sentinel, so base 0 renders as the literal `r.0`. The oracle is complaining about the **input**, not the output; before this change the same input was answered with the parseable `r.-1del`, which is precisely the invention being removed. The `r.` refusal is pinned at the conversion instead, and the test file records the reasoning plus a tripwire for the day `r.0` becomes parseable.

## Local gate — real exit codes, read back from the logs

| command | exit |
|---|---|
| `cargo fmt --all --check` | 0 |
| `cargo clippy --features dev --lib -- -D warnings` | 0 |
| `cargo clippy --features python --lib -- -D warnings` | 0 |
| `cargo ta` | 0 — **10,442 run, 10,442 passed**, 152 skipped |

Four seam oracles (`FERRO_ASSERT_IDEMPOTENT` / `REPARSE` / `IN_BOUNDS` / `SEQUENCE`, excluding `normalization_transcripts_exon_contract`), measured on this branch and on its base rather than quoted as a raw count:

- **base `11338b19`: 17 failures**
- **this branch: 17 failures**
- **set difference in both directions: empty — zero new, zero fixed.**

The inherited 17 are the known `defect_non_idempotent_outputs` / `spec_corpus_regressions` CDS-end residue, `issue_1487_canonical_window_overflow`, `spec_conformance_axis`'s two censuses and `stranded_identity_member`.

Representation-Change: Direct-construction only; a structural zero over any parsed corpus. A `c.`/`r.` position built on base 0 was answered as this axis's own `-1` — `NM_TEST.1:c.?del` normalized to `c.-1del` — and is now left as authored. No parseable input can carry the shape: the parser refuses `c.0` (#269), a bare `c.?` maps to `UncertainBoundary::Single(Mu::Unknown)` whose `inner()` is `None` and so exits before the conversion, and with #1777 in the base the cis collapse no longer manufactures a base-0 anchor. Measured: the full suite is unchanged at 10,442 passing, and the four seam oracles show 17 failures on this branch against the same 17 on its base.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -154,9 +154,24 @@ fn intron_length_at_tx_boundary(
 /// this helper sees `CdsPos { base: 0, utr3: false }`. Special positions
 /// (pter/qter/cen, which also carry `base == 0`) are similarly skipped by
 /// the caller before reaching this helper. The arm is dead code in the
-/// W4004 path; kept here for parity with `Normalizer::cds_to_tx_pos` so
-/// this helper remains a drop-in replacement if a future caller needs it
-/// outside the bounds check.
+/// W4004 path; it is kept rather than deleted for the reason it was
+/// originally kept — parity with `Normalizer::cds_to_tx_pos`, so this
+/// helper stays a drop-in replacement if a future caller needs it outside
+/// the bounds check.
+///
+/// **That parity is what changed it.** The arm used to return
+/// `cds_start - 1`, mirroring what `cds_to_tx_pos` then did; `cds_to_tx_pos`
+/// now refuses, because `background/numbering.md:31` says there is no
+/// nucleotide `c.0`, so the mirror of it here is `None` — no transcript
+/// boundary — not a boundary derived from a position that names no
+/// nucleotide (#1796). Deleting the arm instead would be worse in both
+/// directions: `base == 0` would fall through to the final branch and
+/// evaluate `cds_start + 0 - 1` in `u64`, underflowing on the very value
+/// this note is about, and the helper would silently stop mirroring the
+/// method it exists to mirror. `None` is also already this helper's
+/// vocabulary for "not answerable", and its caller's contract for it is the
+/// conservative skip documented on `check_cds_pos_past_end` — which is the
+/// right outcome for a position no bounds check can be run against.
 fn cds_pos_to_tx_boundary(
     pos: &CdsPos,
     transcript: &crate::reference::transcript::Transcript,
@@ -173,11 +188,12 @@ fn cds_pos_to_tx_boundary(
         let signed = cds_start as i64 + pos.base;
         u64::try_from(signed).ok()
     } else if pos.base == 0 {
-        // Defensive parity with `cds_to_tx_pos`; see helper-level doc
-        // note. Unreachable from `check_cds_pos_past_end` because the
-        // caller short-circuits on `pos.is_unknown()` (which keys off
+        // Defensive parity with `cds_to_tx_pos`, which refuses this position;
+        // see the helper-level doc note. Unreachable from
+        // `check_cds_pos_past_end` because the caller short-circuits on
+        // `pos.is_unknown()` and `pos.is_special()` (both of which key off
         // `CDS_BASE_UNKNOWN == 0`).
-        Some(cds_start.saturating_sub(1))
+        None
     } else {
         // c.<N>: tx_boundary = cds_start + N - 1
         Some(cds_start + (pos.base as u64) - 1)
@@ -9346,7 +9362,9 @@ impl<P: ReferenceProvider> Normalizer<P> {
     /// Convert an RNA position to a transcript-1 position.
     ///
     /// Mirrors `cds_to_tx_pos`. `r.*N` maps to `cds_end + N`, `r.-N` to
-    /// `cds_start + N` (HGVS skips the `0` gap).
+    /// `cds_start + N` (HGVS skips the `0` gap), and `r.0` — which that gap
+    /// means names no nucleotide — is refused, exactly as `cds_to_tx_pos`
+    /// refuses `c.0`. See the arm below and #1796.
     fn rna_to_tx_pos(
         &self,
         pos: &crate::hgvs::location::RnaPos,
@@ -9370,7 +9388,51 @@ impl<P: ReferenceProvider> Normalizer<P> {
                 ),
             })
         } else if pos.base == 0 {
-            Ok(cds_start.saturating_sub(1))
+            // The `r.` axis inherits `c.`'s numbering, so it inherits the gap.
+            // The inheritance is what carries a `c.`-axis clause onto this
+            // axis, so cite the lines that establish it rather than only the
+            // gap itself — `background/numbering.md:58`, "nucleotide numbering
+            // for an RNA reference sequence follows that of the associated
+            // coding or non-coding DNA reference sequence; nucleotide `r.123`
+            // relates to `c.123` or `n.123`", with `:61` naming the coding case
+            // this arm is reached on ("in a coding RNA reference sequence,
+            // nucleotide numbering ... follow[s] that of a coding DNA reference
+            // sequence"). `normalize_rna` already cites that same L58/L61 pair
+            // for this proposition. Only then does `:31` — "there is no
+            // nucleotide `c.0`" — reach here, and `r.0` names no nucleotide
+            // either. Refuse rather than answer with `cds_start - 1`, which is
+            // this axis's own `r.-1` and so a different position from the one
+            // asked about. Issue #1796.
+            //
+            // The non-coding half of `:58` needs no arm here: this function is
+            // only called with a resolved `cds_info` (`normalize_rna`'s
+            // `map_in` and `rna_pos_to_txpos` both branch on it first), and the
+            // no-CDS branch already declines `base < 1`, matching `:60`'s
+            // numbering from `r.1`. So the zone set is read off the reference,
+            // never assumed by the parser.
+            //
+            // This arm carried the same `Ok(cds_start.saturating_sub(1))` as
+            // `cds_to_tx_pos` and — unlike its twin — no comment at all, 2,700
+            // lines away from it. Both are load-bearing for a producer that
+            // #1777 fixes; see the fuller note on `cds_to_tx_pos`.
+            //
+            // `RnaPos` has no `special` field and the `utr3` branch above has
+            // already been taken, so `base == 0` is the whole condition here —
+            // and unlike `cds_to_tx_pos`'s twin, `r.{pos}` really does render
+            // as `r.0`, there being no unknown sentinel on this axis.
+            //
+            // Note this message reaches no caller: both call sites discard the
+            // `Err` (`normalize_rna`'s `map_in` and `rna_pos_to_txpos` both
+            // `.ok()` it) and fall back to leaving the description as authored.
+            // It is read only by `rna_to_tx_pos_refuses_a_zero_base`. Keep it
+            // accurate anyway — it is the half of the seam a maintainer
+            // comparing against #1747's mapper-side refusal will read.
+            Err(FerroError::ConversionError {
+                msg: format!(
+                    "cannot resolve r.{pos} to a transcript position: the RNA base is 0, \
+                     which names no nucleotide"
+                ),
+            })
         } else {
             Ok(cds_start + pos.base as u64 - 1)
         }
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1801, opened 2026-08-13
docs(rulings): duplication.md:18 ranks the dup label, not the partition

nh13:
Records the operator's ruling on D1 — whether `DNA/duplication.md:18` binds at the MEMBER level or the ALLELE level — as a `decided` ledger record, plus the invariant test that enforces its falsifier.

`duplication.md:18` is one of the spec's genuine MUSTs, and four decided records are load-bearing on the assumption that its reach is open: it is cited by `contiguous-insertion-split-by-a-blocked-derivation`, `delins-adjacent-members-when-both-consume-reference`, `separation-rule-force-modal-or-negation` and `exon-junction-dup-converge-from-the-far-side`, and is governing in none of them. Each uses it to scope itself out. So the question had never been ruled on directly.

## The ruling

`docs/recommendations/DNA/duplication.md:18` requires that a change which **is** a tandem duplication be **labelled** `dup`; it does not require that the edit set be partitioned so as to produce one. The test is applied to each piece of the partition ferro derives from the resulting sequence, and a piece whose payload is not a copy of the reference bases immediately 5' of its insertion point is not a duplication, so `:18` never fires on it.

Governing clause is `duplication.md:17`, not `:18` — `:17` supplies the scope, and the whole ruling is that the scope binds.

## Two independent grounds, both cited

The calibration bar is `exon-junction-dup-converge-from-the-far-side`, which needed three independent statements in one file precisely because a single lowercase-prose clause is weak authority. Neither ground below is offered as sufficient alone.

**1. `:17` supplies the scope, and `:18` is its sub-bullet.**

> by definition, duplication may only be used when the additional copy is **directly 3'-flanking** of the original copy (a "tandem duplication")

`:18`, `:19` and `:20` are its indented children, and all three swap one label for another over the same change — `:19` requires `ins` where tandemness is unsupported, `:20` requires `ins` for inverted duplications. None says "and not as, e.g., two members". `README.md:263-269` already reads it this way, in the section the canonical ruleset titles as a worked example of reading force from prose.

**2. The glossary fixes which object the MUST is about.** `docs/background/glossary.md:310-311`:

> **variant**<br> a difference between a reference sequence and an observed sequence.

`:18`'s grammatical subject is *a variant*, so `:18` tests a **difference** — an object that exists prior to any spelling — not a member of a chosen partition. `:17` answers "the must is scoped"; the glossary answers "scoped to what object", and it is definitional and in a different file.

Also cited, and quoted verbatim from the spec checkout (`6f85311`): `duplication.md:5`, `duplication.md:18`, `general.md:56`/`:57` (the prioritisation `:18` cross-references, whose scope `adjudication-precedence-order` records at escalation E1), `insertion.md:5` and `insertion.md:17`. Every quote was checked against the spec file directly rather than inferred from a green build — all eight are byte-exact, not merely whitespace-collapsed matches. The generator's check is a whitespace-collapsed substring match, so a green build alone would not have shown this.

## What this does NOT close

The `dup` half of `delins-adjacent-members-when-both-consume-reference`'s carve-out is untouched. It is a **different mechanism** — merging two adjacent members at separation **zero** into one `delins`, which destroys a `dup` the members themselves carry — from whole-block re-derivation, where no member survives to be destroyed because the partition is recomputed from the resulting sequence. That record says in terms "Do not read this as closing both", and the new record says so explicitly so the next reader does not take D1 as having closed it silently. The adjacent repeat-expansion shape it also names is likewise untouched.

## Which coalesced-arm rows are a gap closing

Two of the six rows are not a new decision at all. `cis_junction_crossing_shift::a_third_member_clear_of_the_tract_keeps_the_duplication_reaching_its_five_prime_most_position` and `cis_junction_crossing_shift::the_three_member_spelling_and_its_one_member_form_are_two_fixed_points` pin a **known gap**, not a decided answer: `contiguous-insertion-split-by-a-blocked-derivation` names the second by its full test path, reproduces the first's measured value verbatim, and reads `:18` the same way this record does. Their convergence under the coalescing arm is the outcome that record adjudicates as correct.

Two more are inside the same record's *stated* reach — its REPRESENTATION EFFECT paragraph predicted them by description, in advance ("any stored allele whose members are a junction-adjacent `ins`+`dup` pair sitting beside a third member far enough away to block the derivation"), which is `issue_1301_adjacent_gap_member_order` and `issue_1320_dup_spans_sibling_junction`. What no decided record had reasoned about at all is the `dup`+`dup`-sharing-a-start shape, and that is what this record supplies.

## The invariant test, and what its sabotage refuted

`insertion.md:5` is **definitional**, not preferential — an insertion is one where "the insertion is not a copy of a sequence immediately 5'". So an emitted pure `ins` whose payload copies its own 5' flank would be a rank-1 conformance defect and the ruling would be wrong as stated. That had only ever been measured *indirectly*, from a `dup` typing pass running first.

`tests/it/duplication_label_not_partition.rs` makes it an enforced invariant: over a 1,061-case corpus, no emitted pure `ins` member has `payload == reference[junction - payload.len() .. junction]`. The corpus authors tandem-copy payloads on purpose so the violation is reachable (478 cases normalize to something carrying a `dup`, 470 pure insertions are examined), and a sibling row refuses a vacuous pass by requiring both labels to be reached. It reads no environment variable, so it re-evaluates on whichever partition arm the process is running.

**It was built against a deliberate break, and the break corrected where `:18` is actually enforced:**

| forced off | the guard |
|---|---|
| `merge::is_tandem_duplication` (so `duplication_anchor` never fires) | still passes |
| `rules::insertion_is_duplication` | still passes |
| `rules::insertion_to_duplication` | still passes |
| `five_prime_match` in `normalize::mod`'s insertion→duplication resolution | **fails** |

Three sabotages left every output byte-identical — `TEMPLATE:g.[74_75insC;75_76insG;90A>T]` still came back as `g.[73_74dup;90A>T]` under each. So the operative `:18` enforcement on these shapes is the **per-member** promotion in `normalize/mod.rs`, not the sequence-first derivation's `duplication_anchor`, and an argument from call order in `anchor_for_piece` is not by itself evidence that the property holds. With `five_prime_match` forced false the guard names the row: `TEMPLATE:g.[264_265insATA;274del]`, payload `ATA` against a 5' flank of `ATA`. All four results were re-measured on the rebased base.

## The counterweight is in the record, and stays unanswered

Per the "record what was refuted" doctrine, and because the operator ruled with it open: under this reading, whether a spec MUST binds is decided by how coarsely the derivation cut — and the cut is set by `MAX_CANONICAL_WINDOW` and by whether an unrelated third member falls inside the window. Four of the rows are exactly that: the same sequence change is `dup`-bearing or not depending on its distance from an unrelated `del`. The reply on offer — that this indicts the partitioner's *locality* rather than the reading — is recorded **as a reply and not as settling it**.

## Dependency: #1752

#1752 ("define cis-member overlap once, on write footprints") is still **open**, and its re-wording of `conflicting-member-geometry-refusal-scope` — "intersecting reference territory" → "intersecting **write footprints**" — is not on `main` (verified against `origin/main`'s ledger). This PR deliberately does **not** duplicate that edit. Until #1752 lands, the new record sits beside a decided record that, read literally as a read-span test, says the D1 allele is refused; #1752's own body records that this is wrong and that two `dup`s write at two different junctions, so the composition is unique. No verdict changes either way — the dependency is descriptive, and it is noted here so the two are not read as contradicting each other in the interim.

## Sidecars

A ledger change moves five things, and all five move here:

- `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json` — the record.
- `RULING_STATUSES` in `tests/it/hgvs_spec_normalization_tests.rs` — id and status.
- Both `CLAUDE.md` adjudication tables — the new record is in the **decided** table, which `claude_md_adjudication_tables::the_two_tables_partition_the_ledger_by_status` enforces.
- `tests/it/clause_ruling_index.rs`'s generated block — captured from `render_index()` programmatically via a throwaway test and spliced on its `BEGIN`/`END` markers matching only `//!` lines. Nothing retyped. 103 → 110 clause lines; the diff is 21 insertions / 3 deletions, so no row was dropped.
- Pins, each read off the tests' own failure output: `the_index_is_not_vacuous` `(26, 22, 4)` → `(27, 23, 4)`. `clauses_named_by_several_records_are_flagged`'s `multi.len()` (27) and `mixed.len()` (8) are **unchanged** — every clause line the record adds was previously either uncited or already multi-cited, and the record is `decided` like every other citer of `duplication.md:18`.

Ledger set comparison against `origin/main`: nothing dropped from `main`, nothing dropped from the pre-rebase tip, exactly one record added, no status changes.

## Verification

- `cargo run --features dev --bin generate_spec_fixture` and `generate_spec_enumeration` regenerated explicitly before the run, both `EXIT=0`. The fixture generation is what validates the record's citations against the spec checkout.
- Full suite, `--no-fail-fast`, with the prepared reference: **10322 tests run, 10322 passed, 150 skipped, `EXIT=0`**, no "tests were not run" line.
- `cargo fmt --all` and `cargo clippy --features dev --all-targets -- -D warnings`, both clean.

Rebased onto `origin/main` at `cc8407bc`. Several open PRs touch this same sidecar set (#1760, #1706, #1747, #1726, #1616, #1784) and they serialize, so another rebase will be needed if one of them lands first.

Representation-Change: none. The record adds no behaviour and the new test only asserts an existing property; no watched file is touched and the full suite needed no value re-pin.


nh13:
@coderabbitai pause

PR #846, opened 2026-06-25
test(conformance): rebless rna_description ledger and protein empty-projection pin (#799)

nh13:
Closes #799

Regenerates the stale mutalyzer conformance baselines against the canonical ferro-prepared reference (reference identity `28288ed4c39ee856`; the scratch SSD is offline, so this was run against the backup mirror, which the orchestrator confirmed byte-identical to the blessed reference — its #764 reference-identity computes to exactly the committed blessed id).

## What changed

Two files, both regenerated artifacts:

- `tests/fixtures/mutalyzer-normalize/baseline-failures/rna_description.txt`: **16 → 0 rows**.
- `tests/fixtures/mutalyzer-normalize/empty-projection/protein_description.count`: **5 → 4** (reference id line unchanged).

## Per-axis FAIL-set delta (live vs committed, mutalyzer axes)

Verified row-by-row against the live run on the canonical reference. Every other axis already matched the committed ledger set-for-set on this reference — only `rna_description` was stale:

| axis | committed | live | added | removed |
|---|---|---|---|---|
| normalized | 12 | 12 | 0 | 0 |
| genomic | 54 | 54 | 0 | 0 |
| coding_protein_descriptions | 20 | 20 | 0 | 0 |
| protein_description | 24 | 24 | 0 | 0 |
| rna_description | 16 | 0 | 0 | **16** |
| noncoding | 0 | 0 | 0 | 0 |
| infos | 8 | 8 | 0 | 0 |
| errors | 1 | 1 | 0 | 0 |

**0 rows added, 16 removed.** No row was demoted that still fails.

The 16 removed `rna_description` rows are all genuine improvements (benign category 6): the rna axis now reports `15 pass / 1 spec_overridden / 0 FAIL` after the r. prediction surface was wired (#838/#840/#843). All 16 previously-failing inputs are now either an exact pass against upstream mutalyzer or the single spec-arbitrated divergence (`NG_012337.1(NM_003002.2):c.[274G>T;278A>G]`, gated by a `spec_citation`). No silent degradation: the rna axis has 0 empty projections and no committed empty-projection budget for that axis.

Most of the improvements the original diagnosis anticipated for the other axes had already landed in their own ledger updates via #839/#843/#844/#845 (all in this PR's base, `dccb6d3`), which is why those ledgers were already current.

## Empty-projection pin

[...]

Issue #651, opened 2026-06-14
test(conformance): mutalyzer axis harness diffs baseline membership, not output quality (silent got-degradation)


nh13:
## Summary

The `tests/mutalyzer_normalize_tests.rs` axis harness decides pass/fail by comparing the live FAIL **set** (which inputs fail) against the committed `baseline-failures/<axis>.txt`. It does **not** compare the *quality* of each row's output. A change can badly degrade output while the harness reports "0 regressions," because the same inputs are still (differently) failing.

## Observed

While working #498, an experimental change regressed ~23 `coding_protein_descriptions` cases from a correctly-framed `NG_(NM_):c.…/NG_(NP_):p.…` pair to `got []` (empty projection). The set-diff vs baseline still showed **0 regressions / 1 demotion**, because every degraded input was already in the baseline. The regression was only caught by manually re-categorizing the `got` strings.

## Proposal

Add an output-quality signal to the harness, e.g. one or more of:
- Track and assert on the count of empty (`got []`) results per axis (a rise is a regression even if membership is unchanged).
- Snapshot per-row `got` strings for the baseline-failing rows (like the `mock-pin` mechanism, but for the manifest axes) and flag drift.
- Emit a per-run summary diff of `got` strings vs the previous run so reviewers see quality changes, not just membership.

## Acceptance

- A change that turns a populated `got` into `[]` (or otherwise degrades it) for a baseline-listed row is surfaced by the harness, not silently accepted.

Refs #326, #498.

PR #1803, opened 2026-08-13
test(conformance): tighten the protein empty-projection budget to what main produces

nh13:
Representation-Change: none. Test fixture only; no watched directory is touched.

## What is wrong

The #651/#764 empty-projection gate is a **budget**: `empty_budget_verdict` fails only when the observed count *rises* above the committed baseline. A fall passes silently, so nothing ever forces the file back down.

`tests/fixtures/mutalyzer-normalize/empty-projection/protein_description.count` has read `4` since #846 (2026-06-26). The axis now produces **1**. That is three units of unused slack — three rows could degrade populated → empty and the gate would stay green, which is the single thing this baseline exists to catch.

The file was ratcheted down deliberately four times before this (19 → 6 → 5 → 4, in #675, #775, #786, #846) and then stopped, while the two later touches (#910, #942) only re-stamped the reference identity line.

## Evidence

Re-derived on both prepared references. They carry the same reference identity `aa8b3246d83055cc` — the one this baseline is pinned to — so the gate was **enforcing**, not skipping (a mismatched identity prints an `empty-projection gate SKIPPED` notice; none appears):

| reference | `protein_description` | `coding_protein_descriptions` |
|---|---|---|
| scratch | 1 empty / 86 measured | 4 empty / 51 measured |
| backup | 1 empty / 86 measured | 4 empty / 51 measured |

The fall is a real improvement, not rows dropping out of the measurement — coverage **grew** over the same period. Rows carrying a `protein_description` expectation in `cases.json`, at each commit that touched the baseline:

| ref | to_test | with a `protein_description` expectation |
|---|---|---|
| #846 `4607ebcf` | 320 | 71 |
| #910 `864573e6` | 320 | 71 |
| #942 `ca16e4f1` | 320 | 71 |
| `HEAD` | 335 | 86 |

The sibling `coding_protein_descriptions` baseline re-blessed to an unchanged `4` and is not touched here.

## How it was written

By the harness's own bless path, so the count and the identity line are the code's rather than hand-typed:

```
BLESS_EMPTY_PROJECTION=1 FERRO_MANIFEST=<ref>/manifest.json \
  cargo nextest run --features dev --no-fail-fast --no-capture \
  -E 'binary(it) and (test(mutalyzer_normalize_tests::axis_protein_description) or test(mutalyzer_normalize_tests::axis_coding_protein_descriptions))'
```

## Verified armed, not vacuous

- baseline at `0` → the axis fails with `protein_description: empty-projection count rose 0 -> 1`, so the gate is live and enforcing exactly the number in the file;
- baseline at `1` → `mutalyzer_normalize_tests` passes **95/95**.

## CI impact

None. PR CI provisions no `FERRO_MANIFEST`, so the axis early-returns and `finish` never runs. The nightly is the only job with a manifest and it is `continue-on-error: true`, so this cannot turn anything red; it only tightens what a local or nightly run reports.


nh13:
@coderabbitai pause

PR #1805, opened 2026-08-13
test: make the reported pair model a closed three-state enumeration

### LINKED ISSUE #1802: reported_partition_verdicts cannot express a pair that both spellings fix
`tests/it/reported_partition_verdicts.rs` models each reported pair as a two-state space, and one of the three states that actually occur is not expressible. The state it cannot express is the one that means **fixed**.

## The model

`each_pair_reaches_its_canonical_form_from_exactly_one_spelling` asserts the two spellings disagree:

```rust
assert_ne!(
    a.verdict, b.verdict,
    "{} and {} have the same verdict; a pair is one canonical spelling and one gap",
    a.label, b.label
);
```

`PAIRS_NO_SPELLING_REACHES` was then added as an exemption list for pairs where **both** are `Gap` — its own doc records why, and records that the model had read "a pair is one canonical spelling and one gap" as a structural fact rather than as a consequence of what `wanted` happened to be:

> both rows of each `1419-r*` pair are `Gap`, and one of them is a `Gap` row that does not retain its input — a combination the model could not express

So the space is:

| state | expressible? |
|---|---|
| `OneReaches` — one `Canonical`, one `Gap` | yes, via the `assert_ne!` |
| `NeitherReaches` — both `Gap` | yes, via the `PAIRS_NO_SPELLING_REACHES` exemption |
| **`BothReach` — both `Canonical`** | **no** |

## Why that matters now

`BothReach` is what a *fixed* pair looks like, and the test's own failure message prescribes producing one:

> `{} now reaches `{}` too, so the pair has converged on the form its issue argues for. That is the fix: flip this row's verdict to Canonical, lower OPEN_GAPS, and say in the PR which stored spelling moved.`

Following that instruction makes both rows `Canonical`, and then the `assert_ne!` twenty lines above fires. **The remedy the message prescribes breaks the assertion in the same function.**

Measured on #1616's head: all nine reported pairs now converge (three did before), and **five reach their `wanted` from both spellings** — `1420-v2`, `1420-v3`, `1420-v4`, `1421-n1`, `1421-n3`. Six rows are red as a direct result.

It is not only #1616. An `authority`-only relabel of the `1421-n*` rows — no output pin moved, `Representation-Change: none` — also breaks this same `assert_ne!`. Two independent PRs are blocked on one structural limitation.

## Proposed fix

Replace the boolean-pair assertion with a closed enumeration, census-pinned per pair:

```rust
enum PairState { BothReach, OneReaches, NeitherReaches }
```

- `NeitherReaches` absorbs `PAIRS_NO_SPELLING_REACHES`, which becomes a derived census rather than a bolted-on exemption.
- `BothReach` becomes sayable, so "this family is fixed" is a state rather than a failure.
- Every transition still requires naming the clause that carried it, exactly as `OPEN_GAPS`'s doc already demands.

## What must NOT change

Three properties this module gets right, which a restructure could easily destroy:

1. **`wanted` stays immutable data.** It is the reported claim, not a measurement, and it is what makes the `verdict` mean anything.
2. **No monotone auto-ratchet on convergence.** `OPEN_GAPS`'s doc rejects this deliberately: a row can reach its `wanted` string by *re-derivation* rather than by a licensed fix — "which is precisely how a defect gets banked as a fix." Lowering the count must keep requiring a named clause.
3. **Convergence stays in the sibling module.** `reported_confluence_pairs` owns it; this module owns "does it reach `wanted`". The two have come apart and the separation is why that is visible.

## Related

- Same family as #1800 (a guard whose precondition is destroyed by the project's own goal).
- Blocks the `reported_partition_verdicts` half of #1616 and the `1421-n*` authority relabel.

nh13:
Closes #1802.

Representation-Change: none. Tests only; no watched file is touched.

`tests/it/reported_partition_verdicts.rs` modelled each reported pair as a two-state space, and one of the three states that actually occur was not expressible — the one that means **fixed**.

`each_pair_reaches_its_canonical_form_from_exactly_one_spelling` asserted `assert_ne!(a.verdict, b.verdict)` — "a pair is one canonical spelling and one gap" — which is `OneReaches` read as a structural fact rather than as a consequence of what `wanted` happened to be. `PAIRS_NO_SPELLING_REACHES` was then bolted on beside it as an exemption list for the pairs where both rows are `Gap`. `BothReach` had no expression at all, so the remedy that same test's failure message prescribed — *"flip this row's verdict to Canonical, lower OPEN_GAPS"* — fired the `assert_ne!` twenty lines above it.

## What changed

`PairState::{BothReach, OneReaches, NeitherReaches}`, census-pinned per pair in `PAIR_STATES`, with a family-wide `PAIR_STATE_CENSUS` of `(BothReach, OneReaches, NeitherReaches)`. `NeitherReaches` absorbs `PAIRS_NO_SPELLING_REACHES`, which becomes a derived census rather than an exemption list; the two sites that exempted those pairs now key on the state and exempt exactly the same set, on exactly the same terms.

`each_pair_reaches_its_canonical_form_from_exactly_one_spelling` becomes `each_pair_reaches_its_wanted_form_from_the_spellings_its_state_names`, which **measures** the reach set per row against that row's own `wanted` and compares it to the pinned state. A pair reaching `wanted` from both spellings now fails that test naming `BothReach` as the entry to move it to — a census edit to argue for, rather than an assertion that cannot be satisfied.

A new `the_pair_state_census_holds` ties `PAIR_STATES` to the `verdict` column per pair, to `PAIR_STATE_CENSUS` for the family, and arithmetically to `OPEN_GAPS` (0/1/2 gap rows per pair by state). `the_open_gap_census_holds` derives the same number by counting rows, so the two censuses agree without either being computed from the other.

The third file is `msto_regression_corpus`, which catalogues these guards against #1419/#1420/#1421 by name and caught the rename in the full local run. Its three entries follow the rename and gain the new shared census guard, so `PINNED_UNIQUE_TEST_COUNT` moves 289 → 290 — an intentional addition, recorded in the comment above the entries.

## What deliberately did not change

- **`wanted` is untouched on every row.** It is the reported claim, not a measurement, and it is what makes `verdict` mean anything. #1419's six still carry the decided chain's answer rather than the filer's ask.
- **No monotone auto-ratchet.** `PAIR_STATES` is pinned data. A pair changing state is a multi-place edit — the table, the census tuple, the affected `verdict`s and `OPEN_GAPS`, all four asserted against each other — and it still requires **naming the clause** that carried it, for the reason `OPEN_GAPS`'s doc gives at length: a row can reach its `wanted` string by re-derivation rather than by a licensed fix. Making `BothReach` *expressible* is not making it *automatic*.
- **Convergence stays in `reported_confluence_pairs`.** That module owns "do the two spellings agree"; this one owns "does it reach `wanted`". The two came apart at #1649 and stayed apart — all three `NeitherReaches` pairs have converged — and nothing here counts convergence.
- **`every_gap_row_is_returned_exactly_as_authored` keeps its distinction**: the input is *already* a fixed point, deliberately stronger than the sibling module's normalize-twice check.
- **No row's `authority` field is touched.**

## Measured states on `main`

All nine pairs, measured by `normalized(input) == wanted` per row rather than read off the `verdict` column:

| pair | state |
|---|---|
| `1419-r1`, `1419-r2`, `1419-r3` | `NeitherReaches` |
| `1420-v2`, `1420-v3`, `1420-v4` | `OneReaches` |
| `1421-n1`, `1421-n2`, `1421-n3` | `OneReaches` |

Nothing is `BothReach` — no pair is fixed. The three-number census is `(0, 6, 3)`, and `OPEN_GAPS` stays at **12**. No output moves and no pin is re-blessed.

## Mutation evidence

Each mutation was applied to a copy of the file, run, and reverted; the real exit code was captured into the log and read back out.

| mutation | outcome |
|---|---|
| `PairState::from_verdicts` collapsed to always return `OneReaches` | `the_pair_state_census_holds` **FAIL** |
| every `PAIR_STATES` entry (and the census tuple) collapsed to `OneReaches` | `the_pair_state_census_holds`, `each_pair_reaches_its_wanted_form_from_the_spellings_its_state_names`, `every_gap_row_is_returned_exactly_as_authored`, `the_wanted_form_of_every_gap_row_is_its_siblings_output` **FAIL** |
| one row's `verdict` flipped with the census left alone | `the_pair_state_census_holds` + `each_pair_reaches_its_wanted_form_from_the_spellings_its_state_names` **FAIL** |
| the coordinated lie — `verdict`, `PAIR_STATES`, `PAIR_STATE_CENSUS` and `OPEN_GAPS` all edited to declare `1421-n1` fixed | `each_pair_reaches_its_wanted_form_from_the_spellings_its_state_names` **FAIL** — the census cannot declare a fix that did not happen, because the state is measured |
| `PairState::gap_rows` collapsed to a constant | `the_pair_state_census_holds` **FAIL** on the `OPEN_GAPS` tie |

## Why this lands on its own

It unblocks **two independent** changes, neither of which is a fix to this module and both of which break the same `assert_ne!`:

- #1616's `reported_partition_verdicts` half, where all nine pairs converge and five reach `wanted` from both spellings; and
- an `authority`-only relabel of the `1421-n*` rows, which moves no output pin at all.

Landing the restructure first means each of those is reviewed as what it is, rather than as a structural change bundled with a behavioural one.


nh13:
@coderabbitai pause

Issue #1087, opened 2026-07-21
fix(project): sentinel offsets from `c.N-?`/`c.N+?` panic (debug) or wrap to garbage coordinates (release) in cds_pos_to_genome_pos


nh13:
`c.N-?` / `c.N+?` carry sentinel offsets (`OFFSET_UNKNOWN_NEGATIVE = i64::MIN`, `OFFSET_UNKNOWN_POSITIVE = i64::MAX`; `src/hgvs/parser/position.rs:100-102`, duplicated as `GENOME_OFFSET_UNKNOWN_*` in `src/hgvs/location.rs:103-106`). `CdotMapper::cds_pos_to_genome_pos` feeds `cds_pos.offset` straight into coordinate arithmetic with no sentinel check (`src/data/mapping.rs:313-360`), so the sentinel is treated as a real intronic offset.

## Reproduction

Debug build (v0.9.1, `6d39853`), DMD / minus strand:

```
$ ferro project --axis g --reference <ref> 'NC_000023.11(NM_004006.2):c.4072-?_5154+?del'
thread 'main' panicked at src/data/mapping.rs:343:25:
attempt to subtract with overflow
```

`[profile.release]` sets only `lto`/`codegen-units`, so **release has no overflow checks and silently emits a garbage coordinate at exit 0**:

```
$ ferro project --axis g --reference <ref> 'NC_000023.11(NM_004006.2):c.4072-?_5154+?del'
NC_000023.11:g.9223372036887140391_9223372036887187721del
```

Plus-strand transcripts never panic even in debug — their arms happen not to overflow and the wrap lands in the `as u64` cast — so they produce the same garbage in **both** profiles:

```
$ ferro project --axis g --reference <ref> 'NC_000013.11(NM_000059.4):c.317-?_425+?del'   # BRCA2, debug
NC_000013.11:g.9223372036887100884_9223372036887100991del
```

Silent wrapping is the wider and more dangerous failure mode; the panic is just the loudest instance.

## Affected arithmetic

| file:line | expression | `-?` (`i64::MIN`) | `+?` (`i64::MAX`) |
|---|---|---|---|
| `src/data/mapping.rs:331` | `base_genome + offset as u64` | n/a | plus-strand silent wrap |
| `src/data/mapping.rs:334` | `(base_genome as i64 + offset) as u64` | plus-strand silent wrap | n/a |
[...]

PR #1806, opened 2026-08-13
fix(effect): decline the unknown-offset sentinel instead of classifying it as a splice site

### LINKED ISSUE #1767: `c.…-?` carries the `i64::MIN` sentinel into every splice classifier's `.abs()` — debug panic in the web handler, `splice_site_variant`/HIGH in release
Found while reviewing #1742, which concentrates ferro's splice-distance classification into three
new public entry points. The overflow it concentrates is pre-existing, so this is not a regression
in that PR — but the PR is the moment the whole class lands in one place, and the same
sentinel-vs-magnitude confusion is already filed one layer up as #1765.

## The path

`-?` is a legal HGVS offset, and the parser represents it with a **sentinel**:

```rust
// src/hgvs/parser/position.rs:102
pub const OFFSET_UNKNOWN_NEGATIVE: i64 = i64::MIN;
```

`parse_offset` returns it directly for `-?` (`position.rs:109-116`). Nothing downstream strips it:
`CdsPos::is_intronic` is `self.offset.is_some() && self.offset != Some(0)`
(`src/hgvs/location.rs:260`), which `Some(i64::MIN)` satisfies. So a description spelled
`…:c.100-?del` carries `offset == Some(i64::MIN)` into every splice classifier.

Each of those classifiers takes `.abs()` of it:

| entry point | line |
|---|---|
| `SpliceRegionBin::from_offset` | `src/effect/mod.rs:72` |
| `SpliceSiteType::from_signed_offset` | `src/reference/transcript.rs:1035` |
| `SpliceSiteType::from_distance_on_side` | `src/reference/transcript.rs:1050` |

`from_signed_offset` calls `from_distance_on_side(offset.abs(), side)`, which immediately `.abs()`es
again — so on this input the same overflowing operation runs **twice**.

## The two outcomes

**Debug builds panic.** `i64::MIN.abs()` is `attempt to negate with overflow`. `src/service/handlers/effect.rs`'s
`predict_cds_effect` reaches `SpliceRegionBin::from_offset` directly from a parsed request field, and
`.coderabbit.yaml`'s `src/service/**/*.rs` instruction names exactly this: *"a handler that can panic,
since that takes down the worker"*.

**Release builds are worse, because they are quiet.** `i64::MIN.wrapping_abs()` is `i64::MIN`, which
is negative, so every `abs_offset <= 2` test passes:

- `SpliceRegionBin::from_offset(i64::MIN)` → `Canonical`
- `SpliceSiteType::from_signed_offset(i64::MIN)` → `AcceptorCanonical`
- `IntronicConsequence::from_cds_pos` → `SpliceAcceptorVariant`
- `predict_cds_effect` → `SO:0001629` / `splice_site_variant` / **impact `HIGH`**

An *unknown* intronic offset is therefore reported as a canonical splice site with high impact. That
is the one answer the input cannot support.

## Status of the evidence

The path above is traced from source and the arithmetic is certain (`wrapping_abs(i64::MIN) == i64::MIN`,
and `i64::MIN <= 2`). It has **not** been executed end-to-end — no test was run that parses
`c.100-?del` and reads the resulting consequence, so treat the reachability of the service handler in
particular as reasoned rather than measured. The reproducer worth running first:

```rust
let v = parse_hgvs("NM_000088.3:c.100-?del").unwrap();
// assert the offset really is the sentinel, then:
// debug   -> this panics
// release -> this returns SpliceAcceptorVariant
```

## The decision this needs

`unsigned_abs()` is the mechanical fix and the repo already uses it
(`src/normalize/mod.rs:212`, `src/vcf/to_hgvs.rs:94`, and `IntronPosition::distance_from_donor` in
the very file in question). But swapping it in **changes the release-mode answer**: with a `u64`
magnitude, `i64::MIN` becomes `9223372036854775808`, so `-?` would classify as `DeepIntronic` /
`Intron` instead of canonical. That is a better answer than the current one and still not obviously
the right one — "unknown offset" is not "very deep in the intron".

So the question to settle is what `+?` / `-?` should classify as on each ladder:

1. **Decline** — the classifiers return `Option` and give `None` for a sentinel offset, so callers
   must handle "unknown" explicitly. Most honest; widest blast radius.
2. **Deep intronic / plain intron** — what `unsigned_abs()` falls out to. Cheapest, and at least
   stops claiming HIGH impact.
3. **A dedicated `Unknown` bin** on both ladders.

This is the same shape as #1765 — a guard keyed on sentinel **identity** where the hazard is
**magnitude** — and the two should probably be settled the same way, since a fix that only teaches
these three functions about `i64::MIN` leaves `+?` (`i64::MAX`) classifying as `DeepIntronic` by
accident rather than by decision.

nh13:
Closes #1767

`-?` and `+?` parse to the sentinels `i64::MIN` / `i64::MAX`, and `CdsPos::is_intronic` is `offset.is_some() && offset != Some(0)`, which both satisfy. So `NM_000088.3:c.100-?del` carried the sentinel into every splice classifier's `.abs()`.

## Reproduced on both profiles, before any fix

The issue traced this from source and flagged it as reasoned rather than measured, so it was measured first, on `cc8407bc`.

**Debug** (`overflow-checks`, the default for `dev`/`test`) — 6 of 8 new tests fail, every one with `attempt to negate with overflow`:

```
thread '…::the_classifiers_are_total_over_every_offset_magnitude' panicked at
  library/core/src/num/mod.rs:450:5: attempt to negate with overflow
Summary  8 tests run: 2 passed, 6 failed
```

**Release** — no panic, which is the worse half. 5 fail on the *value*:

| entry point | release answer for `c.100-?` |
|---|---|
| `IntronicRegion::from_cds_pos` | `Some(CanonicalSpliceSite)` |
| `IntronicConsequence::from_cds_pos` | `Some(SpliceAcceptorVariant)` — HIGH |
| `IntronicConsequence::from_tx_pos` | `Some(SpliceAcceptorVariant)` |
| `EffectPredictor::classify_splice_variant` | `[SpliceAcceptorVariant]` — HIGH |
| `is_clinically_significant_splice_position` | `true` |

`i64::MIN.wrapping_abs()` is `i64::MIN`, which is negative, so every `abs_offset <= 2` test passes.

`predict_cds_effect` (`src/service/handlers/effect.rs`) reaches the ladder straight from a parsed request field, so the debug panic takes down the worker — exactly what `.coderabbit.yaml`'s `src/service/**/*.rs` instruction names.

## The fix, and why it needed no new ruling

The issue lists three options and asks which to take. The repo had already answered: `CdsPos::has_unknown_offset` carries a **MUST** in its doc comment (closing #1087) — *"the sentinels denote an unknown position unbounded in that direction, so no coordinate can be derived from them at all"* — and `certain_offset` (`variant.rs`), `has_unknown_offset_tx` (`normalize/`) and `intronic_to_genomic` all already decline on it. This applies that existing contract to the classifiers that were never taught it; it does not mint a new one.

- **Return type can decline → it declines.** `IntronicConsequence::from_{cds,tx}_pos`, `IntronicRegion::from_{cds,tx}_pos`, and hence `is_clinically_significant_splice_position`. All four already returned `Option`, so no signature moves.
- **Return type cannot decline → the weakest true statement.** `classify_splice_variant` and `predict_cds_effect` return `IntronVariant` / `intron_variant` + MODIFIER: the position *is* known to be intronic, and that answer asserts no distance. `ProteinEffect::intronic_offset` still carries the sentinel, so a caller can distinguish "unknown" from "far".
- **Both validators decline it by name** rather than reporting it as an out-of-range number.

Separately from the sentinel question, every magnitude is now `unsigned_abs` rather than `abs`, so the ladders are total for **every** `i64`. `i64::MIN` is the only overflowing input, and a guard keyed on sentinel *identity* alone is precisely the gap #1765 records one layer up.

The rule had four copies and now has one: `is_unknown_offset` sits beside the constants it tests, and `CdsPos::has_unknown_offset`, the new `TxPos::has_unknown_offset` and `normalize`'s `has_unknown_offset_tx` all delegate to it.

## What the pre-push review found

A class-exhaustion sweep over the diff found **two sites the first pass missed** — `IvsPos::is_canonical_splice_site` and `is_deep_intronic` (`src/hgvs/location.rs`). Not theoretical: `IvsNotation::to_ivs` maps a `CdsPos`/`TxPos` offset straight into `IvsPos`, so the same `c.100-?` reaches them. `the_ivs_predicates_are_total_and_reachable_from_an_unknown_offset` pins that reachability.

Also applied: the premise test now asserts the **AST** offset rather than a rendered round-trip; coverage added for the five `IntronPosition` predicates and both validators (7 changed functions that had none); the fourth copy of the sentinel rule collapsed into `is_unknown_offset`; and a `.filter(|_| …)` guard rewritten as an explicit binding.

Rejected, with reasons: `src/benchmark/compare.rs`'s `.abs()` calls are deltas of two positions, not offsets, and cannot carry a sentinel. `certain_offset` in `variant.rs` was left alone — it tests four constants spanning the genome axis too, which is a wider question than this issue.

## Deliberately left open — needs an API call this PR should not make

`IntronicRegion::from_offset(i64) -> Self` keeps its signature. Every variant of that enum states a distance band, so there is no honest answer for a sentinel; the type would have to become `Option<Self>`. Its two in-repo callers now guard before reaching it, and it is documented as taking a *measured* distance, so nothing reaches it with a sentinel today — but a direct caller passing one still gets `DeepIntronic`. Making it `Option` is a public-API break (measured blast radius: 6 call sites, all in its own test module) and is the "widest blast radius" trade-off #1767 flags as needing a decision. Same for `classify_splice_variant`, which additionally has a PyO3 binding. Happy to take either in a follow-up.

## Verification

All run in `issue-1767/`, exit codes read back from each log rather than from the completion status:

| command | exit |
|---|---|
| `cargo nextest run --features dev -E 'test(issue_1767)'` (before fix, debug) | `100` — 6/8 fail, panic |
| `cargo nextest run --release --features dev -E 'test(issue_1767)'` (before fix) | `100` — 5/8 fail on value, no panic |
| `cargo fmt --all -- --check` | `0` |
| `cargo clippy --features dev --all-targets -- -D warnings` | `0` |
| `cargo doc --no-deps --features dev` | `0` — no warnings from the new links |
| `cargo ta` (full suite) | `0` — 10523/10523 passed, 152 skipped |

Representation-Change: none. No normalized output string moves — the two `src/hgvs/` predicates are new or delegate to an identical body, `normalize`'s helper is byte-equivalent after delegating, and every behaviour change is in effect/consequence classification, which renders no HGVS description. `c.100-?` still renders as `c.100-?`, pinned by `an_unknown_offset_description_carries_the_sentinel_into_the_classifiers`.

### DIFF
--- a/src/convert/coding.rs
+++ b/src/convert/coding.rs
@@ -43,7 +43,13 @@ pub fn validate_cds_pos(pos: &CdsPos, transcript: &Transcript) -> Result<(), Fer
     } else if pos.base < 0 {
         // 5' UTR position
         let utr5_length = (cds_start - 1) as i64;
-        if pos.base.abs() > utr5_length {
+        // `unsigned_abs` for the same reason as the offset checks below:
+        // `i64::MIN.abs()` overflows, and `CdsPos`'s fields are public, so a
+        // library caller can hand this a magnitude no parse can produce
+        // (`parse_cds_pos` negates a `digit1`, whose minimum is
+        // `-i64::MAX`). `utr5_length` is non-negative by construction, so
+        // `max(0)` only makes the `u64` comparison well-typed (#1767).
+        if pos.base.unsigned_abs() > utr5_length.max(0) as u64 {
             return Err(FerroError::InvalidCoordinates {
                 msg: format!(
                     "5' UTR position {} is out of range (min -{})",
@@ -61,9 +67,20 @@ pub fn validate_cds_pos(pos: &CdsPos, transcript: &Transcript) -> Result<(), Fer
         });
     }
 
+    // An unknown offset (`c.100+?` / `c.100-?`) is not a distance, so it is not
+    // a coordinate this can validate — say so rather than reporting it as an
+    // out-of-range number (#1767).
+    if pos.has_unknown_offset() {
+        return Err(FerroError::InvalidCoordinates {
+            msg: "Unknown intronic offset (`+?` / `-?`) denotes no transcript coordinate"
+                .to_string(),
+        });
+    }
+
     // Validate intronic offset if present
     if let Some(offset) = pos.offset {
-        if offset.abs() > 1_000_000 {
+        // `unsigned_abs` because `i64::MIN.abs()` overflows (#1767).
+        if offset.unsigned_abs() > 1_000_000 {
             // Sanity check for very large offsets
             return Err(FerroError::InvalidCoordinates {
                 msg: format!("Intronic offset {} is unreasonably large", offset),
--- a/src/convert/noncoding.rs
+++ b/src/convert/noncoding.rs
@@ -36,11 +36,22 @@ pub fn validate_tx_pos(pos: &TxPos, transcript: &Transcript) -> Result<(), Ferro
         });
     }
 
+    // An unknown offset (`n.100+?` / `n.100-?`) is not a distance, so it is
+    // not a coordinate this can validate — say so rather than reporting it as
+    // an out-of-range number (#1767).
+    if pos.has_unknown_offset() {
+        return Err(FerroError::InvalidCoordinates {
+            msg: "Unknown intronic offset (`+?` / `-?`) denotes no transcript coordinate"
+                .to_string(),
+        });
+    }
+
     // Check if offset is within bounds (intronic positions)
     if let Some(offset) = pos.offset {
         // For non-coding transcripts, offsets are typically not used
-        // but we still validate they're reasonable
-        if offset.abs() > 100000 {
+        // but we still validate they're reasonable. `unsigned_abs` because
+        // `i64::MIN.abs()` overflows.
+        if offset.unsigned_abs() > 100_000 {
             return Err(FerroError::InvalidCoordinates {
                 msg: format!("Offset {} is unreasonably large", offset),
             });
@@ -121,14 +132,20 @@ impl IntronicConsequence {
         }
     }
 
-    /// Create from a CDS position with intronic offset
+    /// Create from a CDS position with intronic offset.
+    ///
+    /// Returns `None` for an unknown offset (`c.100+?` / `c.100-?`): every
+    /// variant of this enum states a distance from the exon boundary, and the
+    /// sentinels denote a position unbounded in one direction, so there is no
+    /// distance to state (#1767). `is_intronic` is satisfied by a sentinel, so
+    /// this guard is what keeps it out — not the check above.
     pub fn from_cds_pos(pos: &CdsPos) -> Option<Self> {
-        if !pos.is_intronic() {
+        if !pos.is_intronic() || pos.has_unknown_offset() {
             return None;
         }
 
         let offset = pos.offset?;
-        let abs_offset = offset.abs();
+        let abs_offset = offset.unsigned_abs();
 
         Some(if offset > 0 {
             // 5' end of intron (splice donor side)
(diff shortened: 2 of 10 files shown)


nh13:
@coderabbitai pause

Issue #1804, opened 2026-08-13
hgvs_to_spdi collapses an unresolvable c./n./r. end boundary onto the start — the transcript-axis sibling of #1795


nh13:
`hgvs_to_spdi` collapses an unresolvable **end** boundary onto the start position on the `c.`, `n.` and `r.` axes, exactly as #1795 records for `g.`/`m.`/`o.`. Descriptions that `parse` and `Display` distinctly share one triple.

Found by class-exhaustion sweep while fixing #1795; measured, not inferred. #1795's fix is scoped to the genomic axes because these three have different position semantics (a `+`/`-` offset is legitimate here, and the guard that fires there — `reject_unresolvable_genomic_position` — has no counterpart on these axes), so refusing here is a separate call with a separate blast radius.

## Reproduction

Measured against `origin/main` at `cc8407bc`, using the existing `make_intronic_provider()` fixture in `src/spdi/convert.rs` (`NM_INTRON.1`, 100 × `A`, CDS 11..90, two exons).

```
NM_INTRON.1:c.10_(20_30)delAAAA
  hgvs_to_spdi : OK  -> NM_INTRON.1:19:AAAA:
NM_INTRON.1:c.10_?delAAAA
  hgvs_to_spdi : OK  -> NM_INTRON.1:19:AAAA:

NM_INTRON.1:n.10_(20_30)delAAAA
  hgvs_to_spdi_simple : OK  -> NM_INTRON.1:9:AAAA:
  hgvs_to_spdi        : OK  -> NM_INTRON.1:9:AAAA:
NM_INTRON.1:n.10_?delAAAA
  hgvs_to_spdi_simple : OK  -> NM_INTRON.1:9:AAAA:
  hgvs_to_spdi        : OK  -> NM_INTRON.1:9:AAAA:

NM_INTRON.1:r.10_(20_30)delaaaa
  hgvs_to_spdi_simple : OK  -> NM_INTRON.1:9:AAAA:
NM_INTRON.1:r.10_?delaaaa
  hgvs_to_spdi_simple : OK  -> NM_INTRON.1:9:AAAA:
```

`c.(4185+1_4186-1)_(4357+1_4358-1)del` — the exon-deletion shape `src/hgvs/interval.rs:18` names as the motivating example for `UncertainBoundary::Range` — is the real-world form of the `c.` row.

## Mechanism

[...]

PR #1807, opened 2026-08-13
fix(spdi): refuse a genomic offset inside a complex boundary, and an end that names no coordinate

### LINKED ISSUE #1795: hgvs_to_spdi drops a genomic offset stated inside a complex (a_b) end boundary — #1641's guard cannot see it, and the end position is collapsed onto the start
`hgvs_to_spdi` refuses a `+`/`-` offset on a simple `g.`/`m.`/`o.` position (#1641) but converts one stated inside a complex `(a_b)` end boundary, dropping both the offset and the boundary. The VCF sibling guard (#1734) has the same shape and is **not** reachable — measured, not assumed.

## Reproduction

Measured against `origin/main` at `0c925605`, in a detached worktree, with a 40-base `NC_000001.11` in a `MockProvider`.

```
CONTROL — the same prohibited offset on a simple position (correctly refused)

  NC_000001.11:g.10_12+2delCGT
    hgvs_to_spdi_simple : ERR invalid position: g. position 12+2 carries a +/- offset: a genomic
                              position cannot have one and SPDI has no offset notation to express it. …
    hgvs_to_spdi        : ERR (identical)

DEFECT — the same prohibited offset inside a complex `(a_b)` end boundary

  NC_000001.11:g.10_(20+1_30)delACGT
    hgvs_to_spdi_simple : OK  -> NC_000001.11:9:ACGT:
    hgvs_to_spdi        : OK  -> NC_000001.11:9:ACGT:

  NC_000001.11:g.10_(20+1_30)del
    hgvs_to_spdi        : OK  -> NC_000001.11:9:C:

  NC_000001.11:g.10_(20+1_30)insTT
    hgvs_to_spdi_simple : OK  -> NC_000001.11:10::TT
    hgvs_to_spdi        : OK  -> NC_000001.11:10::TT

  NC_012920.1:m.10_(20+1_30)del   -> NC_012920.1:9:C:
  NC_001416.1:o.10_(20+1_30)del   -> NC_001416.1:9:C:
```

Expected: `ConversionError::InvalidPosition` naming the offset, exactly as the control gets — the guard's own contract, asserted for the simple form by `a_genomic_offset_is_refused_rather_than_dropped` (`src/spdi/convert.rs`).

**The confluence failure #1628 and #1729 were filed to stop is back on this shape.** Five descriptions `parse`/`Display` keep distinct collapse onto one triple:

```
NC_000001.11:g.10_(20+1_30)delACGT   ->  NC_000001.11:9:ACGT:
NC_000001.11:g.10_(20+5_30)delACGT   ->  NC_000001.11:9:ACGT:
NC_000001.11:g.10_(20-1_30)delACGT   ->  NC_000001.11:9:ACGT:
NC_000001.11:g.10_(20_30)delACGT     ->  NC_000001.11:9:ACGT:
NC_000001.11:g.10delACGT             ->  NC_000001.11:9:ACGT:
```

Note the fourth row: it states **no offset at all** and still collapses. That is the enabling defect and it is measured, not inferred — see *Mechanism*.

### Command

Drop this in as `tests/probe.rs` on a checkout of `0c925605` and run
`cargo test --features dev --test probe -- --nocapture`:

```rust
use ferro_hgvs::parse_hgvs;
use ferro_hgvs::reference::mock::MockProvider;
use ferro_hgvs::spdi::convert::{hgvs_to_spdi, hgvs_to_spdi_simple};

#[test]
fn probe() {
    provider.add_genomic_sequence(
        "NC_000001.11",
        "ACGTACGTACGTACGTACGTACGTACGTACGTACGTACGT".to_string(),
    );
    for input in [
        "NC_000001.11:g.10_12+2delCGT",         // control: refused
        "NC_000001.11:g.10_(20+1_30)delACGT",   // defect: converts
        "NC_000001.11:g.10_(20+1_30)insTT",     // defect: converts
        "NC_000001.11:g.(10+1_11-1)_(20_30)del" // refused, but for the wrong reason — see below
    ] {
        let v = parse_hgvs(input).expect("parses");
        println!("{input}\n  simple : {:?}\n  provider: {:?}",
            hgvs_to_spdi_simple(&v).map(|s| s.to_string()),
            hgvs_to_spdi(&v, &provider).map(|s| s.to_string()));
    }
}
```

## What the spec says

Quoted from the pinned checkout, `assets/hgvs-nomenclature/docs/background/numbering.md` at `6f85311`. The prohibition is stated once per genomic-family axis, in three consecutive bullets:

- `:6` — "Nucleotide numbers based on a genomic reference sequence **do not include** \"+\", \"-\", \"*\", or other prefixes."
- `:8` — "Nucleotide numbers based on a circular reference sequence **do not include** \"+\", \"-\", \"*\" or other prefixes."
- `:11` — "Nucleotide numbers based on a mitochondrial reference sequence **do not include** \"+\", \"-\", \"*\", or other prefixes."

So `m.` and `o.` are not a widening of a `g.`-only clause, which is why all three axes are in the reproduction above.

## Mechanism — two halves, and the second is the one that matters

**Half one, as expected.** `UncertainBoundary::inner()` returns `None` for a `Range` (`src/hgvs/interval.rs:75-80`):

```rust
pub fn inner(&self) -> Option<&T> {
    match self {
        Self::Single(mu) => mu.inner(),
        Self::Range { .. } => None,
    }
}
```

`reject_unresolvable_genomic_position` (`src/spdi/convert.rs:269-299`) walks exactly that, so a `Range` boundary is never inspected and both its endpoints are invisible to the guard:

```rust
for endpoint in [interval.start.inner(), interval.end.inner()].into_iter().flatten()
```

**Half two — and without it there is no defect.** A guard hole is only exploitable where the *conversion* can still produce a coordinate. `get_start_pos`/`get_end_pos` (`src/spdi/convert.rs:218-225`) read through the same `inner()`, so the two sides are handled asymmetrically at all six genomic call sites (`:485`, `:527`, `:594`, `:638`, `:679`, `:723`):

```rust
let start_pos = get_start_pos(&variant.loc_edit.location).ok_or_else(|| /* refuse */)?;
let end_pos   = get_end_pos(&variant.loc_edit.location).unwrap_or(start_pos);
```

- A `Range` on the **start** → `None` → refused, so the guard hole is closed incidentally.
- A `Range` on the **end** → `None` → **silently collapsed onto `start_pos`**, so the guard hole is open.

That asymmetry is the whole defect, and it is why `g.10_(20_30)delACGT` — no offset anywhere — also converts to `9:ACGT:`. The dropped offset is one symptom of a dropped *boundary*.

## Corrections to how this was first reported to me

Both were found by measuring rather than by reading, and both change what a fix has to do:

1. **`g.(100+1_101-1)_(200_300)del` does not reproduce it.** That shape puts the `Range` on the **start**, and it is refused — `invalid position: cannot convert variant with unknown start position`. The guard genuinely skips it, but the conversion refuses it a few lines later. A fix validated only on that shape would validate nothing. The reproducing shape needs the `Range` on the **end**.

2. **The VCF path (#1734) is not affected.** `vcf::from_hgvs::reject_unresolvable_genomic_position` (`src/vcf/from_hgvs.rs:74-112`) has the identical `inner()`-based hole, but both VCF genomic entry points read *both* endpoints with `ok_or_else` (`:616-635` and `:1047-1070`) — there is no `unwrap_or` fallback — so every `Range` boundary is refused before the missing check could matter:

   ```
   NC_000001.11:g.10_(20+1_30)delACGT
     genomic_hgvs_to_vcf : ERR Coordinate conversion error: Genomic variant with unknown end
                               position cannot be converted to VCF: …
   ```

   Measured on all the shapes above. Worth recording because #1734's guard is *latently* wrong in the same way and would start dropping offsets the moment anyone gave the VCF end a fallback.

## Is this #1792's job?

No — checked against that PR's head (`49b24b7d`), not assumed.

`genomic_axis_offset` in #1792's new `src/hgvs/genomic_offsets.rs` handles the `Range` arm explicitly and its doc comment already names this hole. It does find the offset on the reproducer:

```
NC_000001.11:g.10_(20+1_30)delACGT
  #1792 detector       : FOUND stated=`20+1` clause=background/numbering.md:6
  SPDI (no provider)   : OK -> NC_000001.11:9:ACGT:
  SPDI (with provider) : OK -> NC_000001.11:9:ACGT:
```

The detector is correct and the conversion is unchanged: #1792 gates **parse** (strict) and **normalize** (lenient/silent) per the decided `rulings[absolute-prohibition-enforcement-stage]`, while `hgvs_to_spdi` is a public entry point a caller reaches with an already-parsed variant. So this is a sibling of #1792, not a gap in it.

## Scope for a fix

Two questions, and the second is the one worth deciding first:

1. **Should the guard walk both endpoints of a `Range`?** Yes on its own terms — it is written to refuse a prohibited offset and it currently cannot see half the positions a `GenomeInterval` can hold. This is the narrow fix and it closes the reproduction above.

2. **Should `get_end_pos(..).unwrap_or(start_pos)` refuse instead of collapsing?** Fixing only (1) leaves `g.10_(20_30)delACGT` converting to `9:ACGT:` — an uncertain end boundary silently discarded, with no offset involved. That is the same "a conservative decline and a silent drop are written identically" premise #1512 is sweeping for, in a module that sweep does not cover. I have not filed it separately because it is the same six lines; whoever takes this should decide whether it is one fix or two.

Related: #1628 / #1641 (the SPDI guard), #1729 / #1730 / #1734 (the VCF sibling), #1792 (the parse/normalize-stage prohibition), #1512 (the silent-drop-vs-decline family).

nh13:
Closes #1795.

Representation-Change: Previously-accepted input is now refused on the SPDI conversion path; no normalized output moves. `hgvs_to_spdi`/`hgvs_to_spdi_simple` now decline a `g.`/`m.`/`o.` description whose complex `(a_b)` boundary carries a `+`/`-` offset or a `pter`/`qter`/`cen`, and one whose end boundary names no single coordinate (`(a_b)` or `?`); each previously converted by dropping the boundary and collapsing the end onto the start. `normalize` does not route through `hgvs_to_spdi`, and 0 of 85,642 `dump_normalized_corpus` rows can carry any of these shapes — a **structural** zero, not a measured one: the generator spells every position with `format!` from a plain integer, so it cannot vary the property this change keys on.

## The defect

`hgvs_to_spdi` refused a `+`/`-` offset on a simple `g.`/`m.`/`o.` position (#1641's guard) but converted one stated inside a complex `(a_b)` end boundary, dropping both the offset and the boundary and collapsing the end onto the start.

Reproduced on `origin/main` at `cc8407bc` — the issue measured `0c925605`, so this is a re-measurement on current `main`, not a quotation. Five descriptions that `parse` and `Display` keep distinct shared one triple:

|---|---|---|
| `NC_000001.11:g.10_(20+1_30)delACGT` | `NC_000001.11:9:ACGT:` | refused |
| `NC_000001.11:g.10_(20+5_30)delACGT` | `NC_000001.11:9:ACGT:` | refused |
| `NC_000001.11:g.10_(20-1_30)delACGT` | `NC_000001.11:9:ACGT:` | refused |
| `NC_000001.11:g.10_(20_30)delACGT` | `NC_000001.11:9:ACGT:` | refused |
| `NC_000001.11:g.10delACGT` | `NC_000001.11:9:ACGT:` | `NC_000001.11:9:ACGT:` |

Both entry points, all three genomic axes, and every edit shape (`del`, `ins`, `dup`, `inv`, `delins`). Two shapes the issue does not name reproduce the same way and are fixed and pinned here: a **special position** inside a complex boundary (`g.10_(20_qter)delACGT`, the #1643 half of the same blind walk) and an **unknown** end (`g.10_?delACGT`, which `inner()` also answers `None` for).

## The fix

Two halves. Only fixing the first leaves row 4 above — no offset anywhere in it — still colliding with row 5, because the dropped offset is a *symptom* of a dropped boundary.

1. **`reject_unresolvable_genomic_position` now walks the positions a boundary states.** It read each side through `UncertainBoundary::inner`, which answers `None` for a `Range` and so hid *both* of its endpoints. The new `stated_positions` helper is deliberately a different question from `inner()`: "what does the description write down", not "what single coordinate can I resolve". Both endpoints of the range are covered — reading only the near one is the obvious half-fix and passes every assertion naming the first.

2. **`get_end_pos(..).unwrap_or(start_pos)` becomes a declining `resolve_genomic_end_pos`**, at all six genomic call sites. Declining is not a new contract: the **start** side has always declined (`ok_or_else`), and `vcf::from_hgvs` declines on *both* endpoints with no fallback. The fallback was the asymmetry, not the refusal.

## Unchanged on purpose

- **`(13)` still converts.** `Mu::Uncertain` names a coordinate; `(12_14)` names a region and no single position. Refusing every parenthesised boundary is the over-generalisation that would pass all three refusal tests, so `a_resolvable_genomic_end_boundary_still_converts` pins `g.10_(13)delACGT` and `g.(10)_(13)delACGT` at their exact triples.
- **The transcript axes.** `c.`/`n.`/`r.` carry the same fallback at five further sites and collapse the same way — measured (`c.10_(20_30)delAAAA` and `c.10_?delAAAA` both convert to `NM_INTRON.1:19:AAAA:`), and filed as **#1804** rather than swept in. Those axes decide it differently: a `+`/`-` offset inside a `Range` is *legitimate* there (`c.(4185+1_4186-1)_(4357+1_4358-1)del` is `interval.rs`'s own motivating example), and they decline with `MissingReferenceData` rather than `InvalidPosition`, so the verdict is a separate call with a separate blast radius.
- **The VCF sibling (#1734).** `vcf::from_hgvs::reject_unresolvable_genomic_position` has the identical `inner()`-based hole, but both VCF genomic entry points read each endpoint with `ok_or_else` and no fallback, so every `Range` boundary is refused before the missing check could matter. Latent, not reachable; left alone rather than changed without a test that could fail.
- **`g.(10+1_11-1)_(20_30)del`** was already refused, but for the *wrong reason* ("unknown start position", which never mentions the prohibited offset). It is now refused naming the offset. Verdict unchanged, message corrected.

## Representation stability

`hgvs_to_spdi` is not on the normalization path: its callers are `spdi::apply`, `equivalence::checker`, `conformance::spec_corpus`, `Normalizer::to_spdi` and the Python binding — verification and re-expression, never output derivation. No normalized string moves.

The corpus figure is reported as **structural** per the repo's own doctrine. `dump_normalized_corpus` emits 85,642 rows, of which **0** carry `(`, `?`, `pter`, `qter` or `cen` in the input column, and the corpus has no `m.`/`o.` axis at all. The generator builds every spelling with `format!` from plain integer positions, so it cannot vary the property this change keys on — a `--compare` here would report `0 moved` as a fact about the generator, not about the diff.

## Self-review

Ran `/coderabbitai-review` against the local diff before pushing (`assertive` profile, path instruction `src/{project,convert,coords,spdi,liftover}/**/*.rs`).

**Applied:**
- *Class-exhaustion sweep* — the `end.inner().unwrap_or(start)` shape recurs at five sites on `c.`/`n.`/`r.`. Measured and filed as #1804; the scope boundary is now stated on `resolve_genomic_end_pos` rather than left implicit.
- *doc-drift* — the module-header coordinate-support table said `g.` was simply "direct". It now names the refusals, per the repo checklist's "grep every doc comment that recites the axis set". The `hgvs_to_spdi` and `hgvs_to_spdi_simple` contracts were updated too.
- *clippy* — `empty_line_after_doc_comments`, introduced while removing a throwaway probe. Caught by re-running `clippy` and reading the exit code out of the log; the background notification reported `exit code 0` for a chain whose real status was `101`.

**Rejected:**
- *"Require coverage by `tests/it/spdi_tests.rs` or an `issue_*` test"* (path instruction) — the sibling guards this extends (#1628/#1641/#1643) have no integration-level test; their tests live inline in this module's `#[cfg(test)] mod tests`, and the new ones sit directly beside them so the family is read together. Grepped: no `tests/it/` file guards that family.
- *"Fold `three_axis_provider` into `identity_provider`"* — `identity_provider` serves two existing tests on one accession; a new helper avoids touching them, and the refusal tests never reach a fetch while the negative control must.

## Verification

Every exit code below was read back out of its own log file, not from a completion notification.

```
cargo fmt --all --check                                             EXIT=0
cargo clippy --all-features --all-targets -- -D warnings            EXIT=0
FERRO_SWEEP_SEEDS=full cargo nextest run --features dev             EXIT=0   10516/10516 passed
scripts/run_oracle_suite.sh                                         EXIT=0   10379/10379 passed
```

The bare armed run (`FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1`) reports `EXIT=100` / 7 failures, which is the documented pre-existing baseline on `main` — the four `defect_non_idempotent_outputs` tests, `spec_corpus_regressions::an_insertion_at_the_cds_end_is_not_a_fixed_point`, and the two `spec_conformance_axis` censuses, matching `CLAUDE.md`'s list name for name. `run_oracle_suite.sh` is the runner that mirrors CI's `test-oracle`, and it is green.

Five new tests, all in `src/spdi/convert.rs` beside the guards they extend: the offset walk, the special-position walk, the unresolvable end, the discriminating negative control, and the confluence invariant (which asserts its own denominator, so `0 of 0` cannot read as a pass).

### DIFF
--- a/src/spdi/convert.rs
+++ b/src/spdi/convert.rs
@@ -42,7 +42,11 @@
 //! [`hgvs_to_spdi_simple`] accepts coordinate systems whose positions are
 //! resolvable without provider data:
 //!
-//! - `g.` (genomic) — direct.
+//! - `g.` (genomic) — direct, for a position that names a coordinate. A
+//!   `+`/`-` offset (#1628) or a `pter`/`qter`/`cen` (#1643) is refused
+//!   wherever it is written, a complex `(a_b)` boundary included, and an end
+//!   boundary that names no single coordinate — `(a_b)` or `?` — is refused
+//!   rather than collapsed onto the start (#1795).
 //! - `m.` (mito) — the mito accession is genomic; same path as `g.`.
 //!   Wraparound ranges (`start > end`) on circular references are rejected
 //!   per issue #399 — SPDI has no native wraparound representation.
@@ -75,7 +79,7 @@ use crate::convert::CoordinateMapper;
 use crate::coords::{OneBasedPos, ZeroBasedPos};
 use crate::error::FerroError;
 use crate::hgvs::edit::{InsertedSequence, NaEdit, RepeatCount, Sequence};
-use crate::hgvs::interval::Interval;
+use crate::hgvs::interval::{Interval, UncertainBoundary};
 use crate::hgvs::location::{CdsPos, GenomePos, RnaPos, TxPos};
 use crate::hgvs::parser::accession::parse_accession;
 use crate::hgvs::variant::{
@@ -224,6 +228,78 @@ fn get_end_pos(interval: &Interval<GenomePos>) -> Option<u64> {
     interval.end.inner().map(|p| p.base)
 }
 
+/// Every [`GenomePos`] a boundary actually **writes down**, whether or not any
+/// one of them is a coordinate the conversion could use.
+///
+/// [`UncertainBoundary::inner`] answers a deliberately narrower question — "is
+/// this one position I can resolve?" — and so returns `None` for a `Range`,
+/// hiding *both* its endpoints. That is the right answer for a caller asking
+/// for a coordinate and the wrong one for a caller asking what the description
+/// says, and [`reject_unresolvable_genomic_position`] is the second kind: a
+/// prohibited offset is prohibited wherever it is written, so
+/// `g.10_(20+1_30)delACGT` carried one straight past the guard and converted
+/// as though it read `g.10delACGT` (#1795).
+fn stated_positions(boundary: &UncertainBoundary<GenomePos>) -> [Option<&GenomePos>; 2] {
+    match boundary {
+        UncertainBoundary::Single(mu) => [mu.inner(), None],
+        UncertainBoundary::Range { start, end } => [start.inner(), end.inner()],
+    }
+}
+
+/// Resolve a genomic interval's **end** coordinate, declining when the boundary
+/// names no single position.
+///
+/// The call sites previously wrote `get_end_pos(..).unwrap_or(start_pos)`,
+/// which is the same silent drop [`reject_unresolvable_genomic_position`] was
+/// written to stop, one field over: an end that resolves to nothing — a
+/// `(a_b)` range, or a bare `?` — was replaced by the start, so the interval
+/// collapsed to a point and the boundary vanished from the answer. That is what
+/// made `g.10_(20+1_30)delACGT`, `g.10_(20_30)delACGT` and `g.10delACGT` one
+/// triple; the offset is a symptom of the dropped boundary, and closing only
+/// the guard hole would have left the last two of those three still colliding.
+///
+/// Declining is not a new contract. The **start** side has always declined
+/// (`get_start_pos(..).ok_or_else(..)`), and the VCF sibling
+/// (`vcf::from_hgvs`) declines on both endpoints with no fallback at all — the
+/// fallback here was the asymmetry, not the refusal. An SPDI position is an
+/// exact interbase coordinate, so there is nothing an uncertain region or an
+/// unknown position could be rendered as.
+///
+/// Note this is about a boundary that names *no* position, never about one the
+/// author flagged as approximate: `(13)` is `Mu::Uncertain`, `inner()` returns
+/// 13, and it resolves here exactly as a bare `13` does.
+///
+/// **Genomic axes only.** The `c.`/`n.`/`r.` paths carried the same fallback at
+/// five further sites and collapsed the same way — measured, and deferred to
+/// #1804 rather than swept in here, because those axes decide the *neighbouring*
+/// question differently: a `+`/`-` offset inside a `Range` is *legitimate*
+/// there (`c.(4185+1_4186-1)_(4357+1_4358-1)del`), so
+/// [`reject_unresolvable_genomic_position`] cannot simply be pointed at them.
+///
+/// #1804 has since been closed by #1823, whose
+/// [`resolve_transcript_end_boundary`] is this function's transcript-axis
+/// counterpart. The two stay separate for a typing reason rather than a policy
+/// one — this resolves a [`GenomePos`] to the `u64` its call sites need, while
+/// that one hands the axis' own position type back for a further conversion —
+/// and they agree on the verdict: both refuse with
+/// [`ConversionError::InvalidPosition`]. See that function's doc for why it
+/// takes that verdict from its own axis rather than inheriting it from here.
+fn resolve_genomic_end_pos(
+    interval: &Interval<GenomePos>,
+    coord: &str,
+) -> Result<u64, ConversionError> {
+    get_end_pos(interval).ok_or_else(|| ConversionError::InvalidPosition {
+        description: format!(
+            "{coord}. end boundary `{}` names no single coordinate: an SPDI position is an \
+             exact interbase coordinate, and an uncertain range or unknown (`?`) boundary \
+             supplies none. Collapsing it onto the start position would make distinct \
+             descriptions share one triple. Give the end a definite position, or keep the \
+             description in HGVS",
+            interval.end
+        ),
+    })
+}
+
 /// Refuse a genomic interval whose start or end holds something SPDI cannot be
 /// given a coordinate for: a `+`/`-` offset (#1628) or a `pter`/`qter`/`cen`
 /// special position (#1643).
@@ -260,6 +336,14 @@ fn get_end_pos(interval: &Interval<GenomePos>) -> Option<u64> {
 /// description carrying its own bases needs no fetch and no position check, and
 /// so converted.
 ///
+/// **Both endpoints of a complex `(a_b)` boundary are walked, not just the
+/// boundary as a whole** (#1795). The guard originally read each side through
+/// [`UncertainBoundary::inner`], which reports `None` for a `Range`, so a
+/// prohibited offset or special position written *inside* the parentheses was
+/// never inspected: `g.10_(20+1_30)delACGT` and `g.10_(20_qter)delACGT` both
+/// converted. See [`stated_positions`], which is the walk this uses and is
+/// deliberately a different question from `inner()`.
+///
 /// Refusing is also what the other axes already do with the offsets that *are*
 /// legitimate there — see [`resolve_cds_to_tx`], [`resolve_tx_pos`] and
 /// [`require_simple_tx_pos`], each of which declines an intronic `c.`/`n.`/`r.`
@@ -270,8 +354,9 @@ fn reject_unresolvable_genomic_position(
     interval: &Interval<GenomePos>,
     coord: &str,
 ) -> Result<(), ConversionError> {
-    for endpoint in [interval.start.inner(), interval.end.inner()]
+    for endpoint in [&interval.start, &interval.end]
         .into_iter()
+        .flat_map(stated_positions)
         .flatten()
     {
         if endpoint.offset.is_some() {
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1808, opened 2026-08-13
fix(normalize): re-cut an over-described block with the minimal alignment

### LINKED ISSUE #1235: Normalizer: cis-allele members normalized independently, not from the resulting sequence (root cause of #1229–#1234)

nh13:
`merge::prefer_minimal_alignment` re-cuts an already-split block with the minimal alignment when the single-gap one is strictly heavier.

> **Every figure here is re-derived on this branch's current base (`fc6bee2c`), with the command that produced it.** The original description's numbers were measured before **#1835** flipped the default partition rule and before **#1899** deleted the input-relative weight bound the rule's justification was written against; nothing was carried across. The test half of the original PR landed separately as **#2039**, and this branch touches only `src/normalize/merge.rs`, so there is no overlap with it.

## What the rule does

`best_alignment` searches **single-gap** alignments only — one contiguous indel plus substitutions. `partition_block`'s own note records why (letting it place compensating gaps lets it manufacture matches from coincidence and shred a contiguous replacement, #1034/#1040/#182) and states the condition that restriction rests on: it "is sound only while the block *has* a single-gap explanation as economical" as the alternative.

A block whose two length changes have **opposite signs** has no single-gap explanation at all. The lone gap parks at one end and the offset between the two real changes reads as a run of substitutions.

```
$ cargo run --release --features dev --example dump_partitions -- blocks.tsv
ref          alt            distance  live                 live_cols  canonical         canonical_cols
TATATAATAAT  CTATATATAATAA  4         0:6/CTATAT|11:11/AA  8          0:0/CTA|10:11/    4
AGAGT        GAAGAG         3         0:2/GA|4:5/AG        4          0:0/GA|4:5/       3
```

Both rows are what the rule moves — with it applied, `live` reads `0:0/CTA|10:11/` (4 columns) and `0:0/GA|4:5/` (3 columns), i.e. equal to `canonical` on both. Eight changed columns against four on the first row: six of the eight are substitutions that exist only because the gap could not be split.

Bounds, each enforced by a line rather than asserted:

- **The comparand is the block, never the input's spelling.** The function reads only `reference` and `result`, so two spellings of one variant reach it with identical arguments (`derivation-may-not-be-bounded-by-the-inputs-spelling`).
- **It never cuts a block the single-gap search left whole.** Every `whole()` exit in `partition_block` is an adjudication — `delins.md:44-47`'s payload coincidence among them — and they all return first.
- **It never manufactures a member** (`minimal.len() <= pieces.len()`).
- **It re-runs the coincidence guards on its own candidate**, under the same `CoincidenceCarveOut` scope (`delins-payload-coincidence-carve-out-is-coding-dna-scoped`).
- **Equal weight keeps the shipped partition.** Only a strict improvement moves a row.
- **Unequal-length blocks only.** Deliberately *not* on the "an equal-length block has no gap to place" premise the three guards beside it state — `unchanged-is-read-over-every-minimal-alignment` records that premise as false. The real ground is the direction scope of `delins-merge-vs-individual-gap-two-or-more`; the equal-length class is #1939's and is not smuggled in here.

## Scope: it cannot fire on the shipped arm, and that is now measured as well as proved

This sits inside `partition_block`, which is `PartitionRule::Live` — reached by `FERRO_PARTITION=live`, by the shadow audit's independent baseline, and by `dev::live`.

Under the shipped `CanonicalCoalesced` default the dispatch calls `partition_block_canonical` first and reaches `partition_block` only on its **decline**. On that fallback the rule cannot fire, and the proof is over *every* way `partition_block_canonical_within` declines, because this rule calls it again on the **same two slices**:

| decline path | why the retry declines too |
|---|---|
| `cells > max_grid_cells` | `cells` is a function of the two lengths alone; the retry uses `MAX_SEPARATION_AUDIT_CELLS` (2^20 = 1 048 576), **16x smaller** than `MAX_SEQFIRST_GRID_CELLS` (4097² = 16 785 409). A smaller budget can only decline more. |
| the `checked_add`/`checked_mul` overflow arms | same two lengths, same answer |
| `AlignmentDag::build` → `None` (`n + m > MAX_ALIGNMENT_SPAN`, #1970) | a **span** cap, independent of the cells budget entirely |

A grid budget appears in one of the three and is monotone there. **The corpus now measures the same fact directly:** `dump_normalized_corpus`'s own partitioner census reads

```
partitioner: 0 of 73374 blocks declined and were served by `partition_block` (the shipped rule) instead
```

so on the default arm the function containing this rule is entered **zero times** over the whole corpus. The shipped-arm zero below is a code-path zero, not a shape zero.

## What it actually moves, measured

Sabotage = the call `prefer_minimal_alignment(pieces, reference, result, carve_out)` replaced by `pieces` in the source, rebuilt, then `src/normalize/merge.rs` restored from a `cp` copy taken beforehand and checksummed against it (`0986f28a6caa69a5…` identical on both sides, `git status --porcelain` empty).

**Whole Rust suite under `FERRO_PARTITION=live`, rule on vs. sabotaged:**

| | tests run | failed |
|---|---:|---:|
| rule **enabled** | 11 321 | **53** |
| rule **sabotaged** | 11 321 | **58** |

**Five rows fixed, zero broken** — both directions checked with `comm` over the sorted failure sets:

```
normalize::merge::tests::seqfirst_shadow_audit::equal_weight_different_division
it::cis_junction_crossing_shift::the_three_member_spelling_and_its_one_member_form_are_two_fixed_points
it::issue_1316_coincident_tract_repeats::the_seeds_deletion_survives_the_combination
it::spec_corpus_regressions::the_codon_gate_splits_a_spanning_delins_its_own_members_do_not
it::weight_bound_worked_examples::one_variant_normalizes_to_two_strings_because_a_span_outweighs_its_split
```

Four of the five are integration tests asserting normalized HGVS strings, so on the `live` arm this rule does move rendered output. The #1235 locus is the second of them.

**Sabotage on the shipped default:** the same probe over the same suite costs **exactly one assertion of 11 321** — `equal_weight_different_division`, a `#[cfg(test)]` unit test that calls `partition_block` directly. Nothing that reaches a normalized HGVS string moves on the shipped arm.

```
Summary [250.857s] 11321 tests run: 11320 passed, 1 failed, 152 skipped   EXIT=100
FAIL  normalize::merge::tests::seqfirst_shadow_audit::equal_weight_different_division
```

## The corpus zero is structural on BOTH arms — and attributable to one gate

`examples/dump_normalized_corpus` moves **0 of 96 182 rows** across **26** shape families, with the rule enabled under `FERRO_PARTITION=live` as well as under the default. The four dumps are the same binary with two env vars, so the corpora are identical by construction; the two arms differ from each other on **9 629** rows, which is the control proving the arm switch took effect.

That zero is attributable rather than reassuring, and the two arms are zero for **different** reasons. Instrumenting every gate of `prefer_minimal_alignment` and re-running the corpus:

| gate | default arm | `live` arm |
|---|---:|---:|
| calls reaching the function | **0** | 60 555 |
| rejected: equal-length or `< 2` pieces | 0 | 39 260 |
| `partition_block_canonical_within` declined | 0 | **0** |
| reached the cost comparison | 0 | 21 295 |
| rejected: minimal not *strictly* cheaper | 0 | **21 295** |
| **fired** | **0** | **0** |

So on the default arm the code is unreachable; on `live` it is reached 60 555 times, fully exercised through the cost comparison 21 295 times, and stopped every single time by `changed_columns_of_pieces(&minimal) >= changed_columns_of_pieces(&pieces)`. The grid budget is not what blocks it — the callee declined zero times. **The corpus cannot build the shape**, which is a block whose length changes have mixed signs at non-overlapping loci: `three_member_allele` is three substitutions, `dup_plus_ins` two insertions, `del_plus_sub` a deletion and a substitution, and the only mixed-sign family, `junction_interior_to_span`, makes them *coincident*, which the overlap detector refuses before partitioning.

**Reported as structural, in those words.** The zero is a claim about the instrument, not about the change.

## What the corpus cannot witness, an exhaustive sweep can

Over **524 291** block pairs — every unequal-length `(ref, alt)` at lengths 4x5 and 5x4 across the full `ACGT` alphabet, plus the named cases — driven through `examples/dump_partitions` with the rule on and sabotaged:

```
re-cut by this rule           6 098   (1.163%)
  strictly cheaper            6 098   (all of them)
  equal columns                   0
  MORE columns                    0   (no regression anywhere)
  `live` == `canonical` after 6 098   (before: 0)
changed columns eliminated    6 101
```

So the rule is neither rare nor cosmetic on the class it targets: it is exactly a `live`-onto-`canonical` convergence, every instance strictly improving, with no case where it costs anything. The designed corpus builds none of that class — adding a mixed-sign, non-coincident family is the right follow-up, and is worth more than this PR.

## Adjudication

**No new ruling, and no ledger edit.** The ledger holds **34** records at `HEAD`, **34** at `origin/main` and **34** at the pre-rebase tip `8dc000c2`; nothing is dropped relative to either, and no status changes. `docs/NORMALIZATION_CONTRACT.md` is current without re-blessing (`normalization_contract_doc::the_published_document_is_current` passes).

**This is not what closed #1235.** The partition default flip (**#1835**) closed that non-confluence on the shipping arm, and `it::cis_junction_crossing_shift::the_three_member_spelling_and_its_one_member_form_are_two_fixed_points` asserts the convergence and passes on the default arm without this rule. The mechanism the ruling record `contiguous-insertion-split-by-a-blocked-derivation` blames — `canonicalize_from_sequence`'s changed-columns bound — was deleted outright by **#1899**, which is why the `live` "before" is a `delins`+repeat pair rather than the input's own three members.

> **Two separate defects this PR does not fix, flagged rather than absorbed.** The ruling record's `FERRO DOES NOT DO THIS TODAY … It moves nothing today` paragraph is stale on `main` (#1835 closed it), and `docs/NORMALIZATION_CONTRACT.md` carries the same stale prose — correcting that is a documentation change unrelated to `prefer_minimal_alignment`. Separately, `CLAUDE.md` cites this very rule as the motivating example for the `house-choice` ruling kind ("the minimal-alignment rule, which had no home here and grew six drifting prose restatements instead"); filing a `house_choice` record for it is an operator act and is deliberately not done here.

## Verification

Run in `pr-1808/` at `76d74d0f` (base `fc6bee2c`), each **unpiped** with `EXIT=$?` on the next line into its own log, and the `EXIT=` line read back out of the file:

| gate | result |
|---|---|
| `cargo nextest run --features dev --no-fail-fast` | `11321 tests run: 11321 passed, 152 skipped` — **EXIT=0** |
| `cargo clippy --all-features --all-targets -- -D warnings` | **EXIT=0** |
| `cargo fmt --all --check` | **EXIT=0** |
| `python3 -m pytest tests/python/ -q` (maturin `--features python,extension-module`) | `1133 passed, 8 skipped` — **EXIT=0** |

`cargo nextest` does not reach `tests/python/`, so the Python suite was run separately because `src/normalize/` is touched.

Refs #1235.

Representation-Change: none. 0 of 96,182 corpus rows move on the shipped `CanonicalCoalesced` arm and 0 on `FERRO_PARTITION=live`. The shipped zero is by construction and now also by measurement — the corpus census records 0 of 73,374 blocks falling back to `partition_block`, so the function containing this rule is never entered on the default arm, and every decline path of `partition_block_canonical_within` is a function of the two block lengths alone or the budget-independent `MAX_ALIGNMENT_SPAN` cap, which this rule retries on the same slices at a 16x smaller budget. The `live`-arm zero is separately structural: the rule is reached 60,555 times and stopped 21,295 of 21,295 times at the strict-improvement gate, because no shape family builds a mixed-sign non-coincident block. The only measured movement is on the non-default `live` measurement arm, where the suite goes from 58 failures to 53 (five guards fixed, none broken); on the shipped arm, removing the rule costs exactly one unit assertion out of 11,321 and no normalized HGVS string at all.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -5627,7 +5627,239 @@ fn partition_block(reference: &[u8], result: &[u8], carve_out: CoincidenceCarveO
     if split_is_a_placed_gap_coincidence(&pieces, reference, result, carve_out) {
         return whole();
     }
-    pieces
+    prefer_minimal_alignment(pieces, reference, result, carve_out)
+}
+
+/// Re-cut an already-split block with the **minimal** alignment when the
+/// single-gap search provably over-describes it.
+///
+/// [`best_alignment`] considers only single-gap alignments — one contiguous
+/// indel plus substitutions — and [`partition_block`]'s own note records why:
+/// letting it place compensating gaps lets it manufacture matches from
+/// coincidence and shred a genuinely contiguous replacement (#1034/#1040/#182).
+/// The same note states the condition that restriction rests on — it "is sound
+/// only while the block *has* a single-gap explanation as economical" as the
+/// alternative. When two members' length changes do not cancel within one gap,
+/// it does not: the lone gap parks at one end and the offset between the two
+/// real changes reads as a run of substitutions.
+///
+/// # The measured case
+///
+/// `TATATAATAAT` -> `CTATATATAATAA` carries a `+3` insertion and a `-1`
+/// deletion eleven bases apart — **mixed signs, so no single gap can carry
+/// both**, which is the whole trigger. Read off `examples/dump_partitions`,
+/// in block-local 0-based half-open `ref_start:ref_end/alt` form:
+///
+/// ```text
+/// ref          alt            dist  live (before)         cols  canonical  cols
+/// TATATAATAAT  CTATATATAATAA  4     0:6/CTATAT|11:11/AA   8     0:0/CTA|10:11/  4
+/// ```
+///
+/// **8** changed columns against **4** — six of the eight are substitutions
+/// that exist only because the gap could not be split. Both denote the same
+/// bases; one describes twice the change. In the window frame the tests use
+/// (the block starts at offset 4) that is `[4..10>CTATAT, 15..15>AA]` against
+/// `[4..4>CTA, 14..15>]`, and the insertion 5'-rolls to `3..3>ACT` before it is
+/// rendered — so the description reads `g.[3_4insACT;15del]`. The raw member
+/// and the rendered member are quoted separately here on purpose; conflating
+/// them is how an earlier draft of this comment came to state a partition
+/// offset the partitioner never produces.
+///
+/// # WHICH ARM THIS REACHES — read this before quoting it as a fix
+///
+/// This function is inside [`partition_block`], which is
+/// [`PartitionRule::Live`]. It is therefore reached by `FERRO_PARTITION=live`,
+/// by the shadow audit's independent baseline, and by `dev::live` — **not** by
+/// the shipped default.
+///
+/// Under [`DEFAULT_PARTITION_RULE`] (`CanonicalCoalesced`) the dispatch calls
+/// [`partition_block_canonical`] first and only falls back to
+/// [`partition_block`] when that **declines**. On that fallback this rule
+/// provably cannot fire, and the proof is over *every* way
+/// [`partition_block_canonical_within`] can decline, because it is reached here
+/// on the **same two slices** the canonical arm was just handed:
+///
+/// * `cells > max_grid_cells` — `cells` is a function of the two lengths alone.
+///   The canonical arm declines at [`MAX_SEQFIRST_GRID_CELLS`] (`4097^2` =
+///   16 785 409); this rule retries at [`MAX_SEPARATION_AUDIT_CELLS`] (`2^20` =
+///   1 048 576), **16x smaller**, so a smaller budget can only decline more.
+/// * the `checked_add`/`checked_mul` overflow arms computing `cells` — the same
+///   two lengths, so the same answer.
+/// * `AlignmentDag::build` returning `None`, which is `n + m >
+///   MAX_ALIGNMENT_SPAN` (#1970) — a **span** cap, independent of the cells
+///   budget entirely, so it declines identically at either.
+///
+/// A grid budget appears in exactly one of the three and is monotone there, so
+/// the retry declines whenever the first call did. **On the default arm this is
+/// a no-op by construction, not by measurement** — and a corpus zero taken on
+/// that arm is structural and says nothing about the rule.
+///
+/// (An earlier revision of this comment claimed the callee "has exactly one
+/// decline path". It has three. The conclusion survives — but it survives on
+/// span-independence and budget monotonicity, not on there being one arm, and
+/// the difference is the whole proof.)
+///
+/// # What it closes, stated against the tree it lands on
+///
+/// It converges the `live` arm toward the answer the canonical arm already
+/// gives, and that is the whole of its measured effect. Under
+/// `FERRO_PARTITION=live`, `TEMPLATE:g.[4_5insC;5_6dup;15del]` on `DUP_RUN`
+/// normalizes to `TEMPLATE:g.[5_10delinsCTATAT;16A[3]]` without this rule and
+/// to `TEMPLATE:g.[3_4insACT;15del]` with it — the one-member spelling's own
+/// fixed point.
+///
+/// **Sized, so nobody has to guess how much of that there is.** The whole Rust
+/// suite under `FERRO_PARTITION=live` goes from **58 failures to 53** with this
+/// rule enabled: 5 rows fixed, **0** broken. They are
+/// `merge::tests::seqfirst_shadow_audit::equal_weight_different_division`,
+/// `it::cis_junction_crossing_shift::the_three_member_spelling_and_its_one_member_form_are_two_fixed_points`,
+/// `it::issue_1316_coincident_tract_repeats::the_seeds_deletion_survives_the_combination`,
+/// `it::spec_corpus_regressions::the_codon_gate_splits_a_spanning_delins_its_own_members_do_not`
+/// and
+/// `it::weight_bound_worked_examples::one_variant_normalizes_to_two_strings_because_a_span_outweighs_its_split`.
+/// On the shipped default the same sabotage costs exactly **one** assertion of
+/// 11 321 — the unit test above, which reads `partition_block` directly.
+///
+/// **The designed corpus cannot see any of it, on either arm, and the zero is
+/// STRUCTURAL on both — for two different reasons, measured rather than argued.**
+/// `examples/dump_normalized_corpus` moves **0 of 96 182 rows** with the rule
+/// enabled under `FERRO_PARTITION=live` as well as under the default. Do not
+/// quote either zero as safety; each is a claim about the instrument.
+///
+/// * **Default arm — the function is never entered.** The dump's own partitioner
+///   census reads "0 of 73 374 blocks declined and were served by
+///   `partition_block`", so the canonical arm answers every block and the
+///   fallback this rule sits behind is reached **zero** times. That is the
+///   monotonicity proof above, measured: a code-path zero, not a shape zero.
+/// * **`live` arm — entered, fully exercised, and stopped at one gate.**
+///   Instrumenting each gate over the same corpus: **60 555** calls, **39 260**
+///   rejected as equal-length or single-piece, **21 295** reaching the cost
+///   comparison, and **21 295 of 21 295** rejected there because the minimal
+///   alignment is never *strictly* cheaper. `partition_block_canonical_within`
+///   declined **0** times, so the grid budget is not what is blocking. The zero
+///   is attributable to `changed_columns_of_pieces(&minimal) >=
+///   changed_columns_of_pieces(&pieces)` and to nothing else.
+///
+/// The trigger is a block whose length changes have **mixed signs** at
+/// non-overlapping loci, and no shape family builds one: `three_member_allele`
+/// is three substitutions, `dup_plus_ins` two insertions, `del_plus_sub` a
+/// deletion and a substitution; the only family that mixes signs,
+/// `junction_interior_to_span`, makes them *coincident*, which the overlap
+/// detector refuses before partitioning. Add the family before reading the
+/// `live` zero as a statement about this rule.
+///
+/// **What the corpus cannot witness, an exhaustive sweep can — and it is not
+/// small.** Over **524 291** block pairs (every unequal-length `(ref, alt)` with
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

nh13:
Rebased onto current `main` (`f9c2432d` → `6658b554`). Full local gate green — 10513 run, 10513 passed, 0 skipped-on-missing-fixture, `EXIT=0` read back from the log.

**The rebase dropped every census re-bless in this branch, and that is the substantive finding rather than a merge detail.**

This branch's base predates #1835 (`feat(normalize)!: make canonical-coalesced the default partition rule`). Measured on the rebased tree, every census this PR re-blessed now reads exactly `main`'s current pin:

| census | this branch pinned | measured after rebase = `main`'s pin |
|---|---|---|
| `cis_confluence_axis` 3′ / 5′ | 8831/2374/58/9 · 8831/2370/66/5 | 11271/1/0/0 · 11272/0/0/0 |
| `cis_confluence_nr_axis` n3′/r3′/n5′/r5′ | 4414/1190 · 4412/1192 · 4414/1188 · 4412/1190 | 5635/1 · 5633/3 · 5636/0 · 5634/2 |
| `spec_conformance_axis` 3′ / 5′ | 9645/2003/203/31 · 9432/2273/153/24 | 11016/673/164/29 · 10753/981/124/24 |
| `multi_member_cis_axis::respelling_converged` | 402 | 407 |
| `NONCODING_NON_IDEMPOTENT` | 23 | 62 |

So those files are taken from `main` unchanged here; the branch's own content — the `prefer_minimal_alignment` post-pass, the rewritten member-order/junction tests, and `extended_core` — is preserved as authored.

**Why they all collapse to `main`'s values.** `prefer_minimal_alignment` is called from `partition_block` (`merge.rs:4997`), and post-#1835 `partition_block` is reached only from `PartitionRule::Live` — commented in place as *"the pre-flip rule, now reached only by `FERRO_PARTITION=live`"* — plus the canonical arm's oversized-grid decline fallback. The `NONCODING_NON_IDEMPOTENT` pin makes this concrete: `main` documents 62 under the default and **21 under `FERRO_PARTITION=live`**, and this branch pinned **23** — i.e. 21 plus the two rows this fix adds, measured on the `live` arm.

**Consequence for the disclosure:** the `Representation-Change:` trailer on this PR was measured pre-flip, on an arm that is no longer the shipped default. It plausibly moves **zero** shipped rows now and should be re-measured before merge. That is a question about the PR's scope, not about the rebase — flagging rather than editing it.

The tests that were expected to be the hard part all pass unchanged: `issue_1261` 7/0, `issue_1301` 7/0, `issue_1304` 5/0, `issue_1320` 7/0, `spec_corpus_regressions` 16/0. The deliberate input rewrites (moving a member past `MAX_CANONICAL_WINDOW`) survive #1835 intact.

One item left for an operator, not resolvable here: `duplication-must-ranks-the-label-not-the-partition` (decided) names `tests/it/issue_1301_adjacent_gap_member_order.rs` **by full path** as its evidence, and this PR rewrites that test's input. The ruling's citation would then point at a test no longer exercising the geometry it was decided on.

nh13:
Re-reviewed after the patch moved. **Recommend holding.** The `src/` change is byte-identical to the 2026-08-13 review (three `@@` offsets differ); what changed is that the census evidence left the diff while the claims it supported did not.

**The decisive measurement is a sabotage.** No-op'ing `prefer_minimal_alignment` (`if true || …`), with the probe verified present in the artifact both before and after and `merge.rs` restored from a `cp`:

```
10720 tests run: 10719 passed, 1 failed, 152 skipped   SAB_EXIT=100
FAIL  normalize::merge::tests::seqfirst_shadow_audit::equal_weight_different_division
```

**One assertion in the entire Rust suite depends on this rule** — a unit test on the 5-mer `AGAGT -> GAAGAG` that the PR itself re-pinned. All six census tests, the PR's own adjudication guard, and every re-fixtured barrier row pass with the rule removed. The manifest hole was closed too: armed against the blessed reference, `multi_member_cis_axis` + `mutalyzer_normalize_tests` give **103/103, EXIT=0** against untouched pins (arming confirmed by the only discriminator that works here — `axis_normalized` 0.010s unarmed vs 8.875s armed).

**The headline figures are unreachable at this branch's own merge-base.** `464 = 8831 − 8367` and `470 = 8831 − 8361` are internally consistent, but the pins read **11,272 / 11,272** (cis 5', zero headroom) and **11,271 / 11,272** (cis 3', headroom 1). The two spec-conformance "after" values sit ~1,300 *below* their base. The "+1 converged row" is independently dead: `NM_002617.4:c.[13del;22_28delinsACCTGCGCCGCC]` is **already in `CONVERGED_ROWS`** at the merge-base and on `main`.

**Why dropping the census files does not settle it.** The eight figures survive verbatim in four places, **two of them tracked** — the ledger rationale for `contiguous-insertion-split-by-a-blocked-derivation`, and its blessed render at `docs/NORMALIZATION_CONTRACT.md:699`. Those land on `main` as the normative record, so this is evidence-removed-claim-retained rather than a narrowing.

Separately, nine tracked comments plus the ledger attribute to `prefer_minimal_alignment` what the merge-base's own prose attributes to **#1835** (`// Was c.[24T>A;…] until #1835.`). And `every_spelling_of_the_contiguous_insertion_converges_on_its_one_member_form` is a rename with identical assertions — the `assert_ne!` it says it replaces was already an `assert_eq!` at the merge-base.

**What is good here and is not the src change:** the re-fixturing onto `synthetic::extended_core` restores barrier coverage `main` currently lacks and asks for by name. It passes with the rule removed, so it can land on its own — splitting it out is worth considering.

**One risk nobody has measured:** the 42-green was computed against `471c1217`, **32 commits behind** `origin/main`, and the gap contains #1899, which deleted the input-relative weight bound this rule was written to route around. `merge-tree` exit 0 says nothing about that — nothing has run this rule without the bound.

Suggested order: withdraw or re-derive the ledger and `NORMALIZATION_CONTRACT.md` figures first, correct the attributions, then decide whether the rule earns its place.

PR #1809, opened 2026-08-13
feat(equivalence): drive the cross-axis rung over constructed forms, not a diverging corpus pair

### LINKED ISSUE #1800: equivalence_cross_axis_rung depends on a naturally-diverging corpus pair, so every confluence improvement destroys one
`tests/it/equivalence_cross_axis_rung.rs` exercises the denotational rungs (`SequenceMatch`, `CrossAxisSequenceMatch`) by handing `check` **two spellings of one variant that normalize apart**. Its helper enforces that directly:

```rust
assert_ne!(
    first, second,
    "the two spellings converged, so `check` returned at the NormalizedMatch rung and no \
     denotational rung was exercised"
);
```

That is correct as a guard — if the pair converges, `check` short-circuits at `NormalizedMatch` and the rung under test is never reached, so the test would pass while measuring nothing.

## The problem

The guard's precondition is a *divergence*, and reducing divergence is the project's own goal (README rule 3). So **every PR that improves confluence destroys an instance of what this test needs.** The file already records one:

> `#1419`'s `[19_23del;27_33del]` / `19_33delinsCGG` was the pair used here until **#1649 converged it**; do not reach for it again without re-measuring.

That is now the second occurrence. On #1616 the `c.` pair `c.[48dup;52del]` / `c.49_52delinsATTG` converges on `c.49_52delinsATTG`, and its genomic sibling converges too, so three tests fail:

- `a_pair_agreeing_on_both_determined_axes_reaches_the_cross_axis_rung`
- `a_second_axis_that_cannot_be_computed_stops_at_the_single_axis_rung`
- `a_genomic_pair_determines_one_axis_and_so_reaches_the_cross_axis_rung`

The verdict itself moved (`CrossAxisSequenceMatch` -> `NormalizedMatch`), so this cannot be fixed by re-pinning the context strings.

**And it must not be fixed by re-pinning to `NormalizedMatch`.** `rulings[confluence-gate-is-apply-equality-on-every-determined-axis]` holds that equivalence is apply-equality on every determined axis and never `NormalizedMatch` — "a relation over `apply`, whose codomain is bases, so it cannot collapse into the normalizer it gates". Re-pinning would convert a denotational guard into a string-equality guard, which is the opposite of what it is for.

## Why swapping in another pair is not the fix

Swapping in a new naturally-diverging pair works exactly until the next confluence PR, and it also requires proving the replacement's divergence is not itself a defect tracked elsewhere — otherwise the guard is pinned on a bug and is retired the day it is fixed. That is a research step per swap, and the mechanism guarantees a third occurrence.

## Proposed durable fix

Exercise the rungs **directly** rather than depending on the corpus to supply a naturally-diverging pair: construct the two normalized forms the rung is defined over and drive `check`'s rung logic on them, so the test's precondition is built rather than found. The point of these tests is the rung ordering and the cross-axis corroboration, neither of which needs the normalizer to be non-confluent.

Scope note: this is a restructure of the test's fixture strategy, not a behaviour change, and it is deliberately out of scope for #1616 — that PR converges the pair and is blocked on this.

nh13:
`tests/it/equivalence_cross_axis_rung.rs` exercised the denotational rungs (`SequenceMatch`, `CrossAxisSequenceMatch`) by handing `check` two spellings of one variant that normalize apart, and its helper enforced that with an `assert_ne!`. The guard was right — if the pair converges, `check` short-circuits at `NormalizedMatch` and the rung under test is never reached — but its precondition is a **divergence**, and reducing divergence is the project's own goal (README rule 3). So every confluence improvement destroys an instance of what the test needs. #1649 converged the pair the file was first written on; #1616 converges its replacement and three tests with it.

## What this does

Adds `EquivalenceChecker::compare_denotations`, which runs the sequence rung and the cross-axis strengthening above it over two descriptions **as written** — normalizing neither. `check` keeps its ladder exactly as it was and now reaches the same code through a shared private `denotational_verdict`.

The rung tests call the new entry point, so their precondition — two distinct descriptions to compare — is built by the test file rather than borrowed from whatever the normalizer currently declines to converge. It is not test-only scaffolding: a confluence gate is written `is_at_least(CrossAxisSequenceMatch)` and `NormalizedMatch` is deliberately off that order, so `check` reports a rung such a gate must reject for a pair that is genuinely one variant. `check_stops_at_normalized_match_and_so_cannot_reach_a_denotational_rung` pins that as behaviour, and is the one test in the file that *wants* a converging pair.

`rulings[confluence-gate-is-apply-equality-on-every-determined-axis]` is why none of this was fixed by re-pinning to `NormalizedMatch`: equivalence is apply-equality on every determined axis and never `NormalizedMatch` — "a relation over `apply`, whose codomain is bases, so it cannot collapse into the normalizer it gates".

No ledger record is added. The restructure adjudicates nothing; the ruling it rests on is already `decided`.

## The rung is still reached — measured, not asserted

Two mutations, each run against the restructured file:

| mutation | tests that go red |
|---|---|
| the cross-axis comparison agrees but reports `SequenceMatch` (`strengthen_across_axes`, same-accession `Genomic`/`Genomic` arm, `SequenceVerdict::Same`) | `a_pair_agreeing_on_both_determined_axes_reaches_the_cross_axis_rung`, `a_second_axis_that_cannot_be_computed_stops_at_the_single_axis_rung` |
| the second axis is never consulted at all — `strengthen_across_axes` returns `CrossAxisSequenceMatch` unconditionally | those two, plus `a_junction_straddling_dup_pair_agrees_on_the_transcript_and_not_on_the_genome`, `a_genomic_pair_determines_one_axis_and_so_reaches_the_cross_axis_rung`, `the_spec_pair_at_its_own_coordinates_is_single_axis_only` |

## Robust to the convergence that motivated it — demonstrated

#1616's convergence was reproduced locally without touching that branch, by forcing `NM_TWOEXON.1:c.[48dup;52del]` and `chr_two_exon:g.[1038dup;1042del]` onto their merged forms immediately after normalization inside `check`:

* the **previous** version of the file: 3 failed / 11 passed — exactly `a_pair_agreeing_on_both_determined_axes_reaches_the_cross_axis_rung`, `a_second_axis_that_cannot_be_computed_stops_at_the_single_axis_rung` and `a_genomic_pair_determines_one_axis_and_so_reaches_the_cross_axis_rung`, the three #1800 names;
* the **restructured** file: 16 / 16 passed.

Two further tests were converted for the same reason although #1800 does not name them: `a_junction_straddling_dup_pair_...` is the negative control for the very comparison the positive test makes, and `the_spec_pair_at_its_own_coordinates_is_single_axis_only` sits directly under a *decided* ruling — `exon-junction-dup-converge-from-the-far-side` says `c.3922dup` normalizes to `c.3921dup` — so a `check`-driven version of it would stop exercising the mapper's minus-strand arm the day that lands, without saying so.

## Scope

Restructure only. The two-exon fixture and its geometry are unchanged; `reported_partition_verdicts.rs` is untouched (#1802); #1616 is untouched.

The one behaviour change is a note string: the `NotEquivalent` fall-through said "Variants do not normalize to the same form", which names a comparison that never runs on the new path. Nothing pinned it. Levels, normalized forms and errors are unchanged.

## Verification

* full suite, prepared reference wired (`FERRO_MANIFEST` on scratch): **10515 passed, 0 failed**, exit 0
* `scripts/run_oracle_suite.sh`: **10378 passed, 0 failed**, exit 0
* `cargo test --doc --features dev`: 96 passed, exit 0
* `cargo fmt --all -- --check`, `cargo clippy --features dev --all-targets -- -D warnings`: clean, exit 0
* both spec generators regenerate cleanly

Representation-Change: none. Tests plus a new library entry point; `src/equivalence/` is not a watched directory and no normalized output changes.

Closes #1800

### DIFF
--- a/src/equivalence/checker.rs
+++ b/src/equivalence/checker.rs
@@ -383,17 +383,18 @@ impl<P: ReferenceProvider> EquivalenceChecker<P> {
 
             // A description that names no definite bases denotes no sequence,
             // so no rung below may report a positive denotational verdict about
-            // it. The check runs here rather than only on the fall-through path
-            // because normalization *expands* `insN[10]` into a literal run of
-            // `N`s: by the time the sequence rung compares two reconstructed
-            // windows, two indeterminate payloads can be byte-equal, and the
-            // rung would report agreement on a pair where neither side denotes
-            // anything to agree about.
+            // it. The check runs on the **inputs**, here, and not only on the
+            // normalized pair `denotational_verdict` re-checks, because
+            // normalization *expands* `insN[10]` into a literal run of `N`s: by
+            // the time the sequence rung compares two reconstructed windows, two
+            // indeterminate payloads can be byte-equal, and the rung would
+            // report agreement on a pair where neither side denotes anything to
+            // agree about.
             //
             // No input reaching `SequenceVerdict::Same` this way has been
             // measured — every constructed candidate either converges at
-            // `NormalizedMatch` above or is caught by the identical check on
-            // the fall-through path below. This is therefore stated as an
+            // `NormalizedMatch` above or is caught by the same check on the
+            // normalized pair. This is therefore stated as an
             // invariant rather than as a fix for an observed wrong answer, and
             // `an_indeterminate_input_never_wins_a_decided_denotational_rung`
             // pins it as one: the rule is a property of `check`, not of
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1810, opened 2026-08-13
test(normalize): pin the codon exception at a non-zero window offset

nh13:
Representation-Change: none. Tests only; no file under `src/` is touched.

## What this pins

`canonicalize_from_sequence` carries two coordinate bases at once, and the boundary between them is where a real defect has already been written.

Pieces are cut from the trimmed block, then offset into the window (`ref_start += lo`) before `coalesce_coding_frame_separation` and `apply_coding_codon_exception` run. Those two passes convert a piece offset to a CDS position to ask `same_codon`, and they must do it with `w_lo`. Handing either the *block* basis, `w_lo + lo`, reads the reading frame `lo` positions too far along — with `CANONICAL_PAD` at 128, that is two codon phases.

## Why the existing fixtures cannot catch it

`lo` is `min(CANONICAL_PAD, c_lo - 1)`. Every codon-exception fixture in the suite sits within a hundred nucleotides of `c.1`, so `w_lo` clamps to 1 and `lo` collapses to `c_lo - 1`.

That regime is not a weak test, it is a **structurally** blind one. The exception's positive case needs a codon-aligned first position, `c_lo % 3 == 1`, which makes `lo == c_lo - 1` a multiple of 3 in every such fixture. A phase error of `lo` is then a phase error of zero, so the two bases agree by construction and no assertion at that depth — however it is written — can distinguish a correct implementation from the broken one.

## The fixture

A generated 600 nt coding transcript (`ATG` plus a `GCT` tandem, `cds_start = 1`), with the clause exercised at `c.301`, `c.353`, `c.397` and `c.449` — all more than `CANONICAL_PAD` from either end, so `lo == 128` and `128 % 3 == 2`.

The background is chosen, not incidental: a `GCT` tandem has no homopolymer anywhere, so no member can 3'-shift and two sites can differ in codon phase and in nothing else; and every codon is the same reference residue, so the amino-acid counts are attributable to the variant. Amino-acid preconditions are asserted by translating the mutant CDS through `CodonTable`, independently of the normalizer.

## Which assertions discriminate — measured, not asserted

The defect was reintroduced at both call sites (`w_lo` → `w_lo + lo`) and the module re-run. Two tests fail, one per pass, in opposite directions:

| test | pass | wrong-basis output |
|---|---|---|
| `a_length_changing_straddling_pair_deep_in_the_transcript_stays_split` | `coalesce_coding_frame_separation` | `c.449_451delinsTT` — a merge across a codon boundary |
| `a_literal_delins_holding_an_in_codon_triplet_stays_whole` | `apply_coding_codon_exception` | `c.[301G>A;303_304delinsCT]` — a split inside one codon |

The remaining assertions were measured **not** to move, and the module doc records that and why rather than presenting them as guards. A cis pair is merged by `merge_consecutive_edits`, which never leaves axis coordinates and so cannot express this defect — confirmed separately by making `apply_coding_codon_exception` a no-op, which leaves all of them green while reddening three tests in `issue_165_delins_sub_only_decompose`. One test (`a_literal_delins_straddling_a_codon_boundary_is_split`) reaches the right pass and its codon verdict does invert, but a later pass declines the merge anyway, so its output is unchanged; it is kept as the negative half of the pair and labelled a control.

## One limit, stated on the file

The guard discriminates only while `CANONICAL_PAD % 3 != 0`. Within that constraint it is insensitive to the pad's exact value — the merging case sits at `c_lo % 3 == 1`, so either non-zero residue moves it across a codon boundary. Were the pad ever set to a multiple of 3 the file would silently stop discriminating, exactly as the `c.1` fixtures do, so the module doc says to re-derive it rather than re-run it.

## Verification

- `cargo nextest run --features dev --no-fail-fast` — **10519 passed, 0 failed**, exit 0.
- `cargo fmt --all -- --check` and `cargo clippy --features dev --all-targets -- -D warnings` — both exit 0.
- No proptest regression seed was appended.

### DIFF
--- a/tests/it/codon_frame_deep_window_offset.rs
+++ b/tests/it/codon_frame_deep_window_offset.rs
@@ -1,0 +1,530 @@
+//! The coding codon exception, exercised where the block and the window do
+//! **not** share an origin.
+//!
+//! # The coordinate basis this guards
+//!
+//! `canonicalize_from_sequence` fetches a reference *window* `[w_lo, w_hi]`
+//! padded by `CANONICAL_PAD` around the members, applies the edits, and then
+//! trims to the minimal changed *block* with `trim_common_flanks`. That trim
+//! yields an offset `lo`, and from it two different coordinate bases coexist in
+//! the same function:
+//!
+//! * **block-relative** — offsets into `ref_bytes[lo..hi_ref]`, whose axis
+//!   coordinate is `w_lo + lo + offset`;
+//! * **window-relative** — offsets into `ref_bytes`, whose axis coordinate is
+//!   `w_lo + offset`.
+//!
+//! Pieces start block-relative and are shifted into the window (`ref_start +=
+//! lo`) before `apply_coding_codon_exception` and
+//! `coalesce_coding_frame_separation` run, so those two passes must convert with
+//! `w_lo` and nothing else. Handing either of them `w_lo + lo` — the *block*
+//! basis — reads the codon arithmetic `lo` positions too far along. That is a
+//! real defect that has been written: it was found in a partition-preserving
+//! arm whose `split_codon_incompatible_triplets` call passed block coordinates
+//! for pieces that had already been offset into the window, an off-by-`lo` on
+//! both the bases read and the codon phase.
+//!
+//! # Why every pre-existing codon fixture is blind to it
+//!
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #1811, opened 2026-08-13
fix(cli): report a declined description in hgvs-to-vcf and keep going

### LINKED ISSUE #1764: ferro hgvs2vcf aborts the whole stream on one unconvertible description instead of reporting and continuing
### What happens

`ferro hgvs2vcf` reads a file (or stdin) of HGVS descriptions and writes one VCF row per line. The batch loop propagates a per-record conversion error out of the whole command:

```rust
// src/bin/ferro.rs — run_hgvs_to_vcf
for line in reader.lines() {
    let line = line?;
    let trimmed = line.trim();
    if !trimmed.is_empty() && !trimmed.starts_with('#') {
        process_variant(trimmed, &mut handle)?;   // <-- aborts the run
    }
}
```

So one description the converter declines terminates the command, and every remaining line in the input is discarded. The rows already written stay on stdout, so the caller is left with a truncated VCF and a non-zero exit — with no per-record indication of which line was responsible beyond the propagated error text.

### Why this matters now

The set of declined descriptions is growing. Before #1734, `g.10+2del` and `g.pter_4del` produced a (wrong) VCF row and the run completed; #1734 correctly refuses both, which is the right call for the conversion but converts a wrong row into an aborted run for anyone processing a file. Insertions and deletions already abort today on the provider-less path (`genomic_hgvs_to_vcf` declines them by design), so this is not introduced by #1734 — it is enlarged by it, and will be enlarged again by any future refusal.

### The house pattern already exists, two functions away

The batch `normalize`/`parse` driver in the same file models exactly the right thing — a per-line `LineStatus::{Skipped, Ok, Err}`, results buffered in input order, diagnostics formatted to stderr, and a summary. `hgvs2vcf` is the odd one out.

This is also what `.coderabbit.yaml`'s `src/vcf/**/*.rs` path instruction asks for, verbatim: flag *"per-record failure aborting the whole stream where the configured error mode says it should be reported and skipped."*

### What needs deciding (which is why this is an issue, not a PR)

The fix is not mechanical, because it changes the command's contract:

1. **Continue or abort by default?** Per-record skip-and-report matches the sibling driver and the error-mode machinery; aborting matches the current behaviour and never emits a partial VCF.
2. **Exit code.** If it continues, the run must still distinguish "converted everything" from "converted some" — the CLI path instruction requires an exit code that separates "ran, found nothing" from "failed".
3. **Where diagnostics go.** Today `ERROR: <v> is not a genomic variant (g.)` is written to **stdout**, inside the VCF body, which corrupts a downstream parse. That line needs to move to stderr as part of any fix, and is arguably a defect on its own.
4. **Whether the error mode (`strict`/`lenient`/`silent`) should drive it**, as it does elsewhere in the crate, rather than a new flag.

### Reproducer

```bash
printf 'NC_000001.11:g.10C>T\nNC_000001.11:g.10+2del\nNC_000001.11:g.12C>A\n' > in.txt
ferro hgvs2vcf -i in.txt -f vcf
# emits the header + the first row, then exits non-zero; the third row is lost
```

### Scope

`src/bin/ferro.rs`, `run_hgvs_to_vcf`. No library change is required — `src/vcf/from_hgvs.rs` is already returning the right thing.

nh13:
Closes #1764.

Representation-Change: none. CLI-only; `src/bin/ferro.rs` is not a watched directory and no normalized output moves.

## The defect

`ferro hgvs-to-vcf` propagated a per-record conversion error out of the whole command, so one description the converter declines terminated the run and discarded every remaining input line.

Measured on `origin/main` at `674e9c8b`, using the issue's own reproducer plus one more row:

```console
$ printf 'NC_000001.11:g.10C>T\nNC_000001.11:g.10+2del\nNC_000001.11:g.12C>A\nNC_000001.11:g.20A>G\n' > in.txt
$ ferro hgvs-to-vcf -i in.txt -f vcf
##fileformat=VCFv4.2
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO
chr1	10	.	C	T	.	PASS	.
$ echo $?
1
```

Two of four rows are convertible and silently lost. The propagated error text names the description but not the input line, and stdout is a *truncated but well-formed* VCF — the shape most likely to be consumed as complete.

After:

```console
$ ferro hgvs-to-vcf -i in.txt -f vcf
##fileformat=VCFv4.2
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO
chr1	10	.	C	T	.	PASS	.
chr1	12	.	C	A	.	PASS	.
chr1	20	.	A	G	.	PASS	.
# stderr:
ERROR (line 2): NC_000001.11:g.10+2del - Invalid coordinates: genomic position 10+2 (interval start) carries a +/- offset: …
Error: "1 variant(s) failed to convert"
$ echo $?
1
```

### A second defect, found while reproducing

A **non-genomic** description wrote `ERROR: <v> is not a genomic variant (g.)` to **stdout**, inside the VCF body — and the run still exited **0**. Measured on the same base:

```console
$ printf 'NC_000001.11:g.10C>T\nNM_000088.3:c.459A>G\nNC_000001.11:g.12C>A\n' | ferro hgvs-to-vcf -f vcf; echo $?
##fileformat=VCFv4.2
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO
chr1	10	.	C	T	.	PASS	.
ERROR: NM_000088.3:c.459A>G is not a genomic variant (g.)
chr1	12	.	C	A	.	PASS	.
0
```

That is the worse of the two: a corrupt VCF body *and* an exit status asserting success. It is the same routing bug, so it is fixed here rather than filed separately.

## The design choices, and where each comes from

The issue asks for four decisions. **None of them needed a new ruling — three sibling drivers in the same file already answer all four**, and the most valuable part of this change was confirming that rather than inventing a convention. `run_normalize`, `run_parse` and `run_project` each:

1. **Continue** past a per-record failure — none of them aborts the stream.
2. Format the diagnostic to **stderr** via `output_error_with_context`, carrying the **1-based input line number** (`ERROR (line N): <input> - <error>`, or a JSON object with a `"line"` field under `--format json`).
3. **Count** the failures and end with `Err(format!("{n} variant(s) failed to …"))`, so the process exits non-zero whenever any record was declined.
4. Do **not** gate any of this on `--error-mode` — they continue unconditionally, and `hgvs-to-vcf` takes no such flag.

So this PR adds **no flag** and invents **no new exit code**: `hgvs-to-vcf` now reports partial failure exactly the way `normalize`, `parse` and `project` already do. Exit 0 means everything converted; non-zero means some records were declined and stderr says which lines. That satisfies `.coderabbit.yaml`'s `src/{cli,commands,bin}/**/*.rs` instruction on both counts it raises here — "diagnostics written to stdout instead of stderr (that corrupts a downstream parse)" and "a process exit code that does not distinguish 'ran, found nothing' from 'failed'".

## The fix

`run_hgvs_to_vcf` only. The per-record closure returns `Result<(), FerroError>` instead of propagating out of the command; each of the three input paths (single variant, `--input`, stdin) reports and counts rather than `?`-ing.

Incidental, in the same function:

- The ad-hoc `trim()` / `starts_with('#')` line test is replaced with `process_input_line`, the helper the three siblings use — it additionally strips a UTF-8 BOM on the first line and trailing inline comments.
- `OutputFormat` is parsed once and used for every decision, replacing a second parallel `match` on the raw `&str`. Clap pins `--format` to `["vcf", "text", "json"]`, so the two readings could not disagree today; removing one is so they cannot start.

No library change — `src/vcf/from_hgvs.rs` was already returning the right thing, as the issue says.

## Unchanged on purpose

`a_fully_convertible_run_still_exits_zero_with_a_quiet_stderr` is the discriminating test: an all-convertible run still emits every row on stdout, writes **nothing** to stderr, and exits **0**. An exit status that is always non-zero would separate nothing, which is precisely what the CLI instruction asks it to separate.

## Tests

`tests/it/issue_1764_hgvs_to_vcf_continues.rs`, six spawned-binary tests — the behaviour under test *is* the command's contract (stream completeness, stream separation, exit status), and none of the three is observable from a library call. Every pinned string was measured from a run, not inferred from reading the formatter.

| test | pins |
|---|---|
| `a_declined_description_does_not_truncate_the_stream` | the issue's reproducer: all three convertible rows, in input order; the decline named with its line; exit non-zero |
| `a_non_genomic_description_is_reported_on_stderr_and_fails_the_run` | the second defect: nothing but header and data rows in the VCF body; decline on stderr; exit non-zero |
| `a_fully_convertible_run_still_exits_zero_with_a_quiet_stderr` | the discriminating case (see above) |
| `json_format_reports_the_decline_as_json_on_stderr` | `--format json` result stream stays parseable; decline is a `status:error` object carrying `"line": 2` |
| `the_stdin_path_continues_past_a_decline_too` | stdin is a third copy of the loop — pipelines are where a truncated stream does the most damage |
| `the_single_variant_path_reports_on_stderr_without_a_line_number` | the special-cased path: no line number, nothing but the header on stdout |

Row assertions compare the **whole row vector** against measured VCF lines rather than substring-matching a position, so a dropped, duplicated or reordered row fails.

## Self-review findings applied

Ran `/coderabbitai-review` against the local diff before pushing. Three findings, all applied, none rejected:

1. **Duplicate source of truth for the output format** — a `match` on the raw `&str` alongside the new `OutputFormat`. Collapsed onto the parsed value (exhaustive match, no `_` arm).
2. **Untested third input path** — the stdin loop was fixed but uncovered, and it is the one a pipeline uses. Added `the_stdin_path_continues_past_a_decline_too`.
3. **Assertions weaker than the stated contract** (`tests/**/*.rs` path instruction, verbatim) — `contains("\t12\t")` would pass on a duplicated or reordered stream. Replaced with whole-row-vector equality against measured output.

## Verification

All commands run in `issue-1764/`, exit status read back from the log file rather than from a notification.

| command | result |
|---|---|
| `cargo fmt --all --check` | `FMT_CHECK_EXIT=0` |
| `FERRO_SWEEP_SEEDS=full cargo nextest run --features dev` | `SUITE_EXIT=0` — 10526 passed, 0 failed, 152 skipped |
| `scripts/run_oracle_suite.sh` | `ORACLE_EXIT=0` — 10389 passed, 0 failed, 289 skipped |

`git status --porcelain tests/proptest-regressions/` is empty — no `*_strategy()` was touched, so no pinned seed was re-pointed.

### DIFF
--- a/src/bin/ferro.rs
+++ b/src/bin/ferro.rs
@@ -1339,80 +1339,132 @@ fn run_hgvs_to_vcf(
     use ferro_hgvs::hgvs::variant::HgvsVariant;
     use ferro_hgvs::vcf::genomic_hgvs_to_vcf;
 
+    // OutputFormat: FromStr is Infallible, so the never-taken error arm falls
+    // back to the default (Text) — same idiom as `run_project`. Parsed once and
+    // used for every decision below, rather than re-matching the `&str` in
+    // parallel: one value read two ways is one of them going stale later.
+    let out_format: OutputFormat = format.parse().unwrap_or_default();
+
     let stdout = io::stdout();
     let mut handle = stdout.lock();
 
     // Print VCF header if outputting VCF format
-    if format == "vcf" {
+    if out_format == OutputFormat::Vcf {
         writeln!(handle, "##fileformat=VCFv4.2")?;
         writeln!(handle, "#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO")?;
     }
 
-    let process_variant = |v: &str,
-                           handle: &mut io::StdoutLock|
-     -> Result<(), Box<dyn std::error::Error>> {
+    // Convert one description. Returns that record's own failure instead of
+    // propagating it out of the command (#1764): a description the converter
+    // declines is a property of the *record*, not of the run, so it must not
+    // discard the input lines that follow. The set of declined descriptions is
+    // growing — #1734 correctly refuses `g.10+2del` and `g.pter_4del`, and the
+    // provider-less path declines insertions and deletions by design — so an
+    // abort turns one wrong row into a truncated VCF for anyone processing a
+    // file.
+    //
+    // The reporting contract this returns into is the one `run_normalize`,
+    // `run_parse` and `run_project` already share: continue, format the
+    // diagnostic to stderr with the input line number, count it, and fail the
+    // run at the end. `hgvs-to-vcf` was the only batch driver here that did not.
+    let convert = |v: &str, handle: &mut io::StdoutLock| -> Result<(), FerroError> {
         let parsed = parse_hgvs(v)?;
 
-        // Convert genomic variants directly
-        if let HgvsVariant::Genome(ref gv) = parsed {
-            let vcf = genomic_hgvs_to_vcf(gv)?;
-            match format {
-                "vcf" => {
-                    writeln!(handle, "{}", vcf)?;
-                }
-                "json" => {
-                    writeln!(
-                        handle,
-                        r#"{{"hgvs": "{}", "chrom": "{}", "pos": {}, "ref": "{}", "alt": "{}"}}"#,
-                        v,
-                        vcf.chrom,
-                        vcf.pos,
-                        vcf.reference,
-                        vcf.alternate.first().unwrap_or(&String::new())
-                    )?;
-                }
-                _ => {
-                    writeln!(
-                        handle,
-                        "{} -> {}:{}:{}/{}",
-                        v,
-                        vcf.chrom,
-                        vcf.pos,
-                        vcf.reference,
-                        vcf.alternate.first().unwrap_or(&String::new())
-                    )?;
-                }
-            }
-        } else {
-            writeln!(handle, "ERROR: {} is not a genomic variant (g.)", v)?;
+        // Only genomic descriptions convert. This used to write an `ERROR: …`
+        // line to *stdout*, i.e. into the VCF body, which corrupts a downstream
+        // parse — and the run still exited 0, so the failure was invisible. It
+        // is a per-record decline like any other and is reported like one.
+        let HgvsVariant::Genome(ref gv) = parsed else {
+            return Err(FerroError::UnsupportedVariant {
+                variant_type: "only genomic (g.) descriptions convert to VCF".to_string(),
+            });
+        };
+
+        let vcf = genomic_hgvs_to_vcf(gv)?;
+        // A failed write to the result stream must not be reported as success:
+        // count it so the exit status reflects the truncated output. Same
+        // reasoning as `run_project`'s write-failure arm.
+        let io_err = |e: io::Error| FerroError::Io { msg: e.to_string() };
+        match out_format {
+            OutputFormat::Vcf => writeln!(handle, "{}", vcf).map_err(io_err),
+            OutputFormat::Json => writeln!(
+                handle,
+                r#"{{"hgvs": "{}", "chrom": "{}", "pos": {}, "ref": "{}", "alt": "{}"}}"#,
+                v,
+                vcf.chrom,
+                vcf.pos,
+                vcf.reference,
+                vcf.alternate.first().unwrap_or(&String::new())
+            )
+            .map_err(io_err),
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

Issue #1798, opened 2026-08-13
parse_hgvs applies no ErrorConfig — decide whether it should route through strict (defer past v0.14.0)


nh13:
## The question

Should `parse_hgvs` route through `ErrorConfig::strict()`?

Today it applies **no** `ErrorConfig` at all. It trims whitespace, runs the fast path, and falls back to the grammar. The entire input-hygiene ladder lives in `InputPreprocessor`, reachable only from `parse_hgvs_with_config`.

That is now documented honestly — #1667 corrected the doc comment, which had claimed "strict error handling mode by default" — and pinned in both directions:

```
NM_024312.4:r.-6_-3g[6]    parse_hgvs   -> accepted
                           strict()     -> refused ("redundant base label 'g'")
```

So this issue is about **behaviour**, not documentation. The documentation half has shipped.

## Why it is not obvious

Routing the headline entry point through strict would make it **newly refuse inputs it accepts today**. That is a real migration for every caller of the crate's most-used function, and the population is not small — the input-hygiene ladder covers redundant base labels, bare-transcript intronic positions, and more.

It also interacts with the decided ruling `absolute-prohibition-enforcement-stage`, which holds that the enforcement stage is mode-dependent and uniform: **strict fails at parse** (strict validates input conformance, not merely parseability); **lenient** does not validate input conformance and fails only when it cannot normalize; **silent** is lenient without messages. `parse_hgvs` applying no config sits outside that scheme entirely — it is not strict, lenient or silent, which is arguably the real defect regardless of which way this is settled.

## What is already settled, so it is not re-derived

A related claim that intronic-position-on-a-bare-accession is "refused in no mode" is **false**. `bare-transcript-intronic-position` is a decided record with a passing test: strict refuses with `W4007`, lenient accepts. Making that absolute would re-open a ruling rather than fix a gap.

## Scheduling

**Deliberately deferred until after the next release.** It is a breaking change to the crate's headline entry point, and v0.14.0 is already batching representation changes; adding an API-surface migration to that batch would make the release harder to reason about and harder to roll back.

Its stated prerequisite has been met — #1634 (the `absolute-prohibition-enforcement-stage` ruling) has merged — so it is decidable now; it is being scheduled, not blocked.

## Options, for whoever picks it up

[...]

PR #1812, opened 2026-08-13
fix(project): record why the g. axis declines on a bare transcript input

### LINKED ISSUE #1713: project --axis g declines every bare NM_-rooted input, though the library projects the same inputs on the same reference
`ferro project --axis g` declines every bare `NM_`-rooted input against the prepared reference, while the library demonstrably can produce that axis for the same inputs on the same reference. The CLI is not reaching a capability ferro has.

## Measured

At `origin/main` = `3228d228`, against the prepared GRCh38 reference (546,290 sequences, cdot 0.2.32, 482,519 transcripts with CDS metadata loaded — so the mapping data is present):

```
$ ferro project -i in.txt --reference <dir> --axis g
# NM_004006.2:c.100del (line 1) [g]: unavailable: no g. representation for this variant
# NM_000492.4:c.1521_1523del (line 2) [g]: unavailable: no g. representation for this variant
# NM_004006.2:c.1704del (line 3) [g]: unavailable: no g. representation for this variant
```

Exit status 0 — the decline is reported as a first-class outcome, which is correct behaviour for a genuine decline. The question is whether it *is* one.

## Why this looks like a CLI-path gap rather than a missing capability

`mutalyzer_normalize_tests::axis_genomic_idempotent` runs against this same prepared reference and exercises real projected genomic axes derived from `c.`/`n.`/`r.` transcript inputs, via `project_variant(v, &tx_id).genomic`. It passes. So the library projects transcript-rooted inputs onto the genomic axis on this reference; the `--axis g` CLI path returns `unavailable` for the same class.

The one place the decline string is asserted in-tree is a unit test on a **mock** transcript that carries no genome alignment (`src/conformance/spec_projection.rs`, `coding_input_projects_every_axis_from_one_pass`) — where declining is correct and expected. That test cannot distinguish "no alignment for this transcript" from "the CLI path never asked", so it does not cover this.

## Why it matters

The task this was raised under records the consequence: it **caps every projection measurement**. Any sweep or corpus run that projects bare `NM_`-rooted inputs to `g.` reads a decline for the whole class, so the class contributes nothing and its absence looks like a clean result rather than an unmeasured population. That is the "a corpus zero is a claim about the corpus" failure mode, reached through the CLI rather than through a generator.

## Relationship to #1708

#1708 restores the genomic axis for **junction-crossing** inputs, by re-parenting the output onto the genomic wrapper the crossing had already resolved — its own body records `project-g/c.1704del` and `c.1704dup` moving from `unavailable: no g. representation for this variant` to `NC_000023.11:g.32573745del`/`dup`. That is a real gain but a **subset**: it applies where a junction crossing put a genomic anchor back into the description. `NM_004006.2:c.100del` and `NM_000492.4:c.1521_1523del` above cross no junction, so #1708 leaves them declining. (`c.1704del` is included in the measurement above because #1708 is not merged; expect that row to change when it lands.)

## What to determine

1. Whether `--axis g` on a transcript-rooted input is *intended* to require a genomic anchor in the description, or should derive one from cdot the way `project_variant` does.
2. If the latter, which path the CLI takes and why it differs from `project_variant`.
3. If the former, then the decline is by design and what is owed is a clearer message — "requires a genomic reference in the description or `--transcript`" rather than "no g. representation for this variant", which reads as a property of the variant rather than of the invocation.

Note `ferro project` also accepts `--transcript`, documented as "required for a bare `g.` input"; whether it has any effect on the reverse direction is worth checking as part of (2).

nh13:
Closes #1713

Representation-Change: none. No axis changes availability and no rendered HGVS description moves — the only string that changes is the `unavailable:` decline *reason* on the `g.` axis of a bare transcript input, which is diagnostic text and is not a representation. `Status::ProjectionUnavailablePinned` is assigned from `AxisResult::Unavailable(_)` with the payload ignored, so the enumeration census is unmoved (verified: 63/63 including `spec_enumeration_tests`).

## The filed claim does not reproduce, and that is the first finding

#1713 reports `ferro project --axis g` declining bare `NM_`-rooted inputs that "the library demonstrably can produce for the same inputs on the same reference". Measured side by side against the prepared GRCh38 reference (546,290 sequences, cdot 0.2.32, 482,519 transcripts), on `origin/main` at `674e9c8b`:

```
NM_004006.2:c.100del
  CLI  project_axis(.., Axis::Genomic, None)  -> Unavailable
  LIB  project_variant(&v, tx).genomic        -> None
  LIB  project_to_genomic(&v)                 -> Err(UnsupportedProjection:
           "input variant has no parent reference (genomic_context) on accession
            NM_004006.2; project_to_genomic requires an explicit NG/NC parent (see #327)")
  DATA provider transcript chromosome         -> Some("NC_000023.11")
```

There is no CLI/library asymmetry. `select_axis` reads `VariantProjection::genomic` directly, so the CLI reports exactly the value the library computed, and `project_to_genomic` — the dedicated library entry point — declines the same input with a hard error. The decline is **specified**: #327's scope says "when the input has no resolvable parent reference (e.g. plain `NM_…:c.…` with no NG/NC context), return `FerroError::UnsupportedProjection`", and `project_to_genomic_nc` carries a standing comment that it does not synthesize a parent from cdot.

The issue's evidence was `mutalyzer_normalize_tests::axis_genomic_idempotent` passing. That test does `let Ok(g1) = projected else { continue }` — every input that fails to project is silently dropped, so a corpus of bare `NM_` rows would be skipped entirely and the test would still pass on its other rows. It cannot witness this class.

This is **not** the same mechanism as #1798/#1632. Those are about `parse_hgvs` applying no `ErrorConfig`, so the plain parse entry accepts what strict rejects — a config-application gap at the parse seam. This is a projection-capability question at a documented refusal, and the two entry points here agree rather than diverge.

## What was actually defective: the reason, not the verdict

The issue's own third option is what stands — the decline is by design and what is owed is a clearer message. `project_coding_direct` and `project_noncoding_direct` left `AxisDeclineReasons::genomic` unset, so `unavailable_reason` fell back to a string synthesized from the axis code alone:

```
# NM_004006.2:c.100del (line 1) [g]: unavailable: no g. representation for this variant
```

That asserts a property of the **variant**. It is false here in a way anyone can check in one command: the alignment is present, and naming the genomic reference renders the axis immediately (`NC_000023.11(NM_004006.2):c.100del` → `NC_000023.11:g.32849814del`, same reference, same session). The wording is what made a specified decline read as a capability gap — it is why this issue was filed, and it caps projection measurements exactly as the issue describes, since a sweep reads "this variant has no genomic form" rather than "these descriptions need a parent".

**After:**

```
# NM_004006.2:c.100del (line 1) [g]: unavailable: no genomic reference for NM_004006.2: a bare
  transcript description names no genomic parent, so the g. axis cannot be derived (see #327);
  name one in the description, e.g. NC_000023.11(NM_004006.2):c.…
# NM_000492.4:c.1521_1523del (line 2) [g]: unavailable: no genomic reference for NM_000492.4: …
  e.g. NC_000007.14(NM_000492.4):c.…
```

The worked remedy names the reference that transcript is *actually* aligned to — chrX for DMD, chr7 for CFTR — resolved cdot-first with a provider fallback, and elided to `NC_…` rather than invented when the reference names no alignment at all. It runs only on the decline path, so nothing that renders pays for it.

Both direct paths' LRG arms previously discarded their declining error via `.ok()`; they now record it on the same terms.

## `--transcript` has no effect in this direction

The issue asks. Measured: for a transcript-rooted input, `project_axis` only cross-checks `--transcript` against the input's own accession — matching is a no-op, mismatching is a hard error. It can never supply the genomic parent, and the flag's documented role ("required for a bare `g.` input") is the forward direction only.

```
--transcript NM_004006.2   ->  unavailable: … (unchanged)
--transcript NC_000023.11  ->  ERROR: --transcript NC_000023.11 does not match the input's
                               transcript NM_004006.2
```

## Two in-tree comments named the wrong cause

Both `src/cli/project.rs`'s `project_axis_bare_nm_genomic_axis_unavailable` and `src/conformance/spec_projection.rs`'s `coding_input_projects_every_axis_from_one_pass` explained this decline as the fixture carrying "no genome alignment". Neither fixture is like that: both align `NM_TEST.1` to `NC_000001.11` with a full cdot exon table, and sibling tests in the same modules project a bare `NC_000001.11:g.` input onto that transcript successfully. The issue body cites the second of these as the only in-tree assertion of the decline string and reads it exactly as written — so the misdescription propagated. Corrected, with the correction stated rather than silently applied.

## What is deliberately not changed

Deriving a genomic parent from cdot for a bare `NM_` input would reverse #327's recorded design decision and is an adjudication, not a bug fix. Worth noting that ferro already does synthesize one in a single narrow case — #1708 re-parents a junction-crossing output onto the reference the crossing had already resolved, which is why `NM_004006.2:c.1704del` renders `NC_000023.11:g.32573745del` despite arriving bare. That is grounded in `checklist.md:20` and is not a general licence; it is included in the new test corpus precisely so the module pins *agreement* rather than a per-row verdict.

## Tests

- `src/cli/project.rs::project_axis_bare_nm_genomic_decline_names_cause_and_remedy` — hermetic, runs in CI. Asserts the decline names the accession, the missing genomic reference and the remedy, and does **not** fall back to the axis-code string; then renders the identical variant once the parent is named, in the same test, so the two halves cannot drift.
- `tests/it/issue_1713_bare_transcript_genomic_decline.rs` — manifest-backed.
  - `cli_and_library_agree_on_the_genomic_axis_for_bare_transcript_inputs` pins the invariant #1713 alleged was broken, in both directions (a CLI decline demands a library `None`; a CLI render demands an identical library `Some`), with a `checked == 4` guard so the corpus cannot shrink silently.
  - `the_bare_transcript_genomic_decline_names_its_cause_and_the_remedy` checks the message against real cdot across two chromosomes — a hardcoded accession cannot satisfy both — and then **rebuilds the description the message recommends and projects it**, so the remedy cannot be a well-formed lie.

No fixture files added; all inputs are literals.

## Verification

```
cargo fmt --all                                        clean
cargo clippy --features dev --all-targets -- -D warnings   EXIT=0
cargo nextest run --features dev --no-fail-fast --lib      4250 run, 4250 passed
cargo nextest run --features dev --no-fail-fast --test it \
  -E 'test(cli_project) or test(spec_enumeration) or test(hgvs_spec_normalization)
      or test(issue_1713) or test(cross_axis) or test(clause_ruling_index)
      or test(generator_completeness)'                     63 run, 63 passed
```

Both manifest-backed tests were confirmed to have *run* (~26 s each loading the reference), not skipped.

### DIFF
--- a/src/cli/project.rs
+++ b/src/cli/project.rs
@@ -592,7 +592,14 @@ mod tests {
 
     #[test]
     fn project_axis_bare_nm_genomic_axis_unavailable() {
-        // A bare NM_ coding input has no genome alignment → g. is None.
+        // A bare `NM_` coding description names no genomic parent, so there is
+        // no reference for the `g.` axis to be written against → `g.` is None.
+        //
+        // This comment previously read "has no genome alignment", which is not
+        // what the fixture models and not why the axis declines — see
+        // `project_axis_bare_nm_genomic_decline_names_cause_and_remedy` below,
+        // which pins that this very fixture aligns the transcript to
+        // `NC_000001.11` and projects against it happily.
         let vp = fixture();
         let variant = crate::parse_hgvs("NM_TEST.1:c.4C>A").unwrap();
         let outcome = project_axis(&vp, &variant, Axis::Genomic, None).unwrap();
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

Issue #1690, opened 2026-08-12
cds_to_tx_exon_aware overflows on an extreme c. coordinate reached through hgvs_to_spdi


nh13:
`cds_to_tx_exon_aware` adds `start_tx + offset` unchecked on both early-return branches (`src/convert/mapper.rs:113`), so an extreme `c.` coordinate reaching it through `hgvs_to_spdi` panics with `attempt to add with overflow` rather than being refused.

## Why this is filed separately

This is a **fourth** site of the class #1487 covers. #1487 and its fix (#1658) close three unchecked additions on the *normalization* path — the canonical-window pad, the axis conversion, and `enclosing_exon`. This one sits on a different path entirely: it is reached via `hgvs_to_spdi`, not via `normalize`, and it lives in `src/convert/` rather than `src/normalize/`. #1658 deliberately did not widen its scope to cover it.

## Reproducer

Three of the tests already committed in `tests/it/issue_1487_canonical_window_overflow.rs` trip it when the sequence oracle is enabled:

```
FERRO_ASSERT_SEQUENCE=1 cargo nextest run --features dev --test it -E 'test(issue_1487)'
```

The panic is at `src/convert/mapper.rs:113`, in the `start_tx + offset` of the early-return branches.

## Why CI is green today

`FERRO_ASSERT_SEQUENCE` is the only one of the four seam oracles that surfaces it — `FERRO_ASSERT_IDEMPOTENT`, `FERRO_ASSERT_REPARSE` and `FERRO_ASSERT_IN_BOUNDS` all pass. CI's `Test oracle` job sets only those three (`ci.yml:1013-1015`); `FERRO_ASSERT_SEQUENCE` is set exclusively by the `sweeps` job, which negates `ORACLE_EXCLUDE` and does not reach these tests. So the gating jobs stay green and the panic is invisible to the required checks.

## Suggested shape of a fix

Match what #1658 did at the other three sites: `checked_add` and a refusal rather than a panic. Worth checking at the same time whether any other arithmetic in `src/convert/mapper.rs` is reachable with an extreme coordinate — the three sites in `src/normalize/merge.rs` were found together, and this one was found only because a test on an unrelated path happened to reach it.

PR #1813, opened 2026-08-13
ci: record the selection-wide FERRO_ASSERT_SEQUENCE measurement test-oracle asked for

nh13:
`test-oracle`'s comment block stated that `FERRO_ASSERT_SEQUENCE` was unset only because the evidence for arming it was two rows rather than the job's own selection, and that arming it "needs a run over the selection below (whole suite minus ORACLE_EXCLUDE, test(proptest) and SWEEP_FILTER) first". Both rows it named — #1618 and #1619 — are now closed and green.

**That run has now been done, and it is red.** The two closures were necessary and not sufficient. This PR records the measurement; it does **not** arm the flag and does **not** change the selection.

## The measurement

Measured on `origin/main` @ `674e9c8b`, with `test-oracle`'s own four flags plus `FERRO_ASSERT_SEQUENCE=1`, over that job's `-E` with `SWEEP_FILTER` and `ORACLE_EXCLUDE` expanded from `ci.yml` and no `--partition`:

```
not test(proptest)
  and not (test(cis_junction_crossing_shift) + test(repeat_span_sibling_overlap) + test(issue_1254_sibling_crossing_shift))
  and not (test(spec_conformance_axis) + test(spec_corpus_regressions) + test(corpus_alphabet_and_intronic_validity) + test(corpus_prohibited_inputs) + test(lenient_overlap_and_projection_split) + test(defect_non_idempotent_outputs) + test(issue_1715_rna_alignment_symbol_reach))
```

```
Summary [87.5s] 10383 tests run: 10378 passed, 5 failed, 289 skipped
```

The expression was derived by parsing `ci.yml` rather than transcribed, and cross-checked byte-for-byte against `scripts/run_oracle_suite.sh --print-selection`, which derives it independently.

## The five failures, in two classes

Neither class is #1618 or #1619, and neither is reachable by the two-row check the comment rested on.

**3x `issue_1487_canonical_window_overflow`** — not an oracle fire at all. `attempt to add with overflow` at `src/convert/mapper.rs:114`. Arming the flag routes the normalized output through `hgvs_to_spdi`, which reaches `cds_to_tx`'s unchecked `cds_start as i64 + pos.base - 1` on an `i64::MAX`-adjacent `c.` coordinate. That module already records this measurement in its own comments and names the site; the open issue is **#1690**.

**2x `stranded_identity_member`** — a genuine fire, of #1281's "denotes no sequence" shape:

```
input   TEMPLATE:g.[10_11insAT;11_12inv;12_13insA]
output  TEMPLATE:g.[11_12insTA;11_12=;12_13insA]
```

The output pairs an insertion at interbase 11|12 with an identity claiming 11..12, so it denotes nothing. That module exists to **pin** a defect (#1655's stranded derived identity, 132 of 132 non-confluent), which places it in exactly the class `ORACLE_EXCLUDE`'s comment already documents: a test that pins a defect and an oracle that aborts on it cannot both run.

Neither is suppressed here. Excluding either to get a green would hollow out the oracle, which is the one thing the existing comment is emphatic about.

## Two further findings from the same run, recorded so they are not rediscovered

With `FERRO_MANIFEST` also set — which `test-oracle` does **not** set, so CI does not see these — six `mutalyzer_normalize_tests` axis tests fail as well, on real `NG_` geometry. For example `NG_012337.1(NM_012459.2):c.4_5insGTA` normalizes to `c.5_6insTAG`, where over `[34, 35)` the input applies `GTAA` and the output `ATAG` — an insertion shifted 3' past a base its first payload base does not match. Those axes early-return without the manifest, which is why the run reads 11 failures with it and 5 without.

**Proof the flag is not a no-op over this selection:** `issue_1615_denoted_sequence_oracle::the_seam_compares_when_the_flag_is_set` is in this selection and asserts the oracle's `compared` counter is non-zero; it passes with the flag set and skips by design without it. The five failures are themselves the stronger evidence.

## Also corrected

- `ci.yml`'s two "the only job that sets `FERRO_ASSERT_SEQUENCE`" claims (at the `SWEEP_FILTER` note and on the `sweeps` job) are true only *within this file* — `nightly-mutalyzer.yml:201` sets it too, non-gating. Both now say so.
- `scripts/run_oracle_suite.sh`'s header quoted "5 further failures on `main`, all in `spec_corpus_regressions`". That figure is about `ORACLE_EXCLUDE`'s modules, which this script's selection does not run, so it never described what it claimed to. Replaced with the measurement over the selection it actually mirrors. The script needs no code change to stay correct — `oracle_flags` reads the flag set out of `ci.yml`, so it will arm the flag on the day the job does.
- The two `tests/it/oracle_exclude_invariant.rs` doc comments that still named #1618/#1619 as the blocking rows.

## Scope

Comment-only. `git diff` contains no non-comment line. The flag remains unset, the `-E` selection is untouched, and `sweeps` and `nightly-mutalyzer.yml` are unchanged. Arming `test-oracle` is now blocked on #1690 and on finding a home for `stranded_identity_member` — not on another measurement.

Verified: `actionlint .github/workflows/ci.yml` clean, `cargo fmt --all --check` clean, and all 14 tests that parse `ci.yml` (`oracle_exclude_invariant`, `sweep_filter_invariant`, `generated_fixture_ci_wiring`) pass.

Representation-Change: none. CI comments and test doc comments only; no watched directory is touched and no output moves.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,8 @@ env:
   # NOTE also: `sweeps` selects on this filter PLUS one module it deliberately
   # re-runs — `issue_1615_denoted_sequence_oracle`, the only test that reads the
   # denoted-sequence oracle's counters, which are meaningless unless
-  # `FERRO_ASSERT_SEQUENCE` is set and `sweeps` is the only job that sets it. So
+  # `FERRO_ASSERT_SEQUENCE` is set and `sweeps` is the only job IN THIS FILE that
+  # sets it (`nightly-mutalyzer.yml` sets it too, non-gating). So
   # "nothing is run twice" holds of this filter and is one module short of
   # holding of that job's `-E`; the reason it is appended there rather than added
   # here is on the job itself.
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
The six manifest-only `NG_` rows this PR records as a side finding are now filed as #1815, re-derived on `origin/main` @ `e3f08263`.

Two corrections from that re-derivation, both in this PR's favour rather than against it:

- **Six failing axis *tests*, seven distinct *inputs*.** The seventh is `NG_012337.1(NM_012459.2):c.[10del;23_25del;36_37insAAT]` → `c.[10del;23_25del;37_38insATA]`, which fires only on `axis_normalized` and so is hidden behind an axis already failing for another row.
- **The mechanism is not uniform.** Six of seven are the 3′-shift-past-a-non-matching-base this PR names; the seventh, `c.6delinsCCCC` → `c.6_7insCCC`, is a `delins`→`ins` rewrite that keeps the reference `T` the `delins` deleted.

Also confirming this PR's reachability claim by direct A/B on the same tree and manifest: unarmed `95 tests run: 95 passed` (exit 0, 35.2 s), armed `95 tests run: 89 passed, 6 failed` (exit 100, 32.5 s).

The `stranded_identity_member` half is measured in fulcrumgenomics/ferro-hgvs#1655 — the 132 still reproduce, and that issue's `COMPLETED` close came from #1668's `Closes` keyword on a tests-only PR that states it does not fix.

PR #1814, opened 2026-08-13
test: enumerate each cis adjudication class's description space

nh13:
Representation-Change: none. Tests only; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched.

## What this is

`tests/it/cis_confluence_adjudication.rs` states each of its records on one or two spellings. This adds `tests/it/cis_adjudication_enumeration.rs`, which turns each of those records into a property: generate the bounded space of spellings that denote the *same* variant, normalize all of it, and bucket by output.

Five classes, which is the whole of that module — records 1 (`c.` and `g.`), 2, 3 and 4. The two extra tests over there (`the_decided_target_is_the_re_derived_form`, `the_two_gap_deletion_split_is_a_fixed_point_through_the_diagnostics_exit`) are further points on classes already covered, not classes of their own.

## The two dimensions, and what bounds them

**(a) Spanning single-member descriptions** — every `del`/`delins` whose span covers the change, start at or before the first changed column and end at or after the last, with the payload *solved* from the resulting sequence rather than guessed. On the `c.` axis this runs the span's 3' end into the 3'UTR, so `c.{i}_*1` is in the space.

**(b) Multi-member spellings with each member at every legal shift** — each seed member is independently re-positioned, its payload re-solved at each position, and kept exactly where the member *applied alone* denotes what the seed member denotes. The cartesian product is taken; overlapping and non-monotone combinations are dropped. A shift preserves the deleted length, so this dimension does not re-partition.

Each surviving member is then rendered in every spelling HGVS admits for it — a deletion with and without stated bases, a single column as substitution and as `delins`, an insertion also as a `dup` where the payload copies its 5' flank — so typing varies as well as position.

**Nothing generated is trusted.** Every candidate is round-tripped through `spdi::hgvs_to_spdi` and applied to the reference; one that does not denote the class's sequence is a hard failure, not a silent drop. The shift search is asserted **closed** — an equivalent sitting at the edge of the search window fails the run rather than quietly truncating the space.

## What is asserted, by kind

- **Adjudicated-correct** — the two `separation-is-a-property-of-the-spelling-not-of-the-variant` classes are asserted to collapse to the decided form. A rule read off the *spelling* would partition these spaces; a rule read off the *sequence* cannot. That is exactly the property the record decides, which is why an enumeration is the right instrument for it rather than a third pinned string.
- **Observation** — the adjacency carve-out, the `dup`-flush carve-out and the open spanning-versus-split class have their measured partitions pinned, each naming the record that owns the residue. These are change detectors and are labelled as such.
- **Cross-arm, unconditional** — every distinct output is applied back to the reference and must denote what its inputs denoted. This runs on every `FERRO_PARTITION` arm, not just the shipping one.

## Limits, stated in the module

The module doc says plainly that this proves **only** that one variant's spellings converge, and nothing about different variants staying apart — a normalizer mapping everything to one string would pass every convergence assertion here. The output sequence-preservation check rules out that degenerate case for *these* classes; it does not and cannot supply the separating half.

The structural audit (`the_enumerated_space_can_vary_what_it_claims_to_vary`) asserts what the fixtures **can** vary and names what they provably cannot: the transcript is `CDS 1..=63` of 64 bases and single-exon, so a 5'UTR span, an exon junction, an intronic coordinate and a block over `MAX_SPLIT_BLOCK` are all structurally unreachable. Any zero this module ever reports for those shapes is structural, and the module says so rather than leaving it to be re-derived.

The CDS-boundary spellings the space *can* reach are counted and pinned separately, so their absence can never be mistaken for their convergence.

## Cost

The whole module runs in about 1.5 s. It is in the default selection.

### DIFF
--- a/tests/it/cis_adjudication_enumeration.rs
+++ b/tests/it/cis_adjudication_enumeration.rs
@@ -1,0 +1,1159 @@
+//! Every `cis_confluence_adjudication` record, turned from a point assertion
+//! into a property over an **enumerated description space**.
+//!
+//! Each record over there pins one or two spellings. Here the complete bounded
+//! space of spellings that denote the *same variant* is generated, and the whole
+//! space is normalized and bucketed by output. For a record whose ruling is
+//! `decided` **and implemented**, the property is that the space collapses to one
+//! string. For a record that is a scope carve-out or an open disagreement, the
+//! property is the measured partition of the space, pinned as an observation with
+//! the record that owns the residue named on it.
+//!
+//! # What this proves, and — more importantly — what it does NOT
+//!
+//! **It proves only that one variant's spellings converge.** It says *nothing*
+//! about different variants staying apart: a normalizer that mapped every input
+//! to the single string `"x"` would pass every convergence assertion in this
+//! file. Convergence is one half of canonicalization and this module measures
+//! only that half. **Do not read a green run here as more than "the space did not
+//! fragment".**
+//!
+//! One guard rail is in place and it is worth being precise about what it buys.
+//! [`assert_outputs_are_sequence_preserving`] applies every distinct output back
+//! to the reference and requires it to denote what its inputs denoted — the
+//! `FERRO_ASSERT_SEQUENCE` question, asked over exactly these outputs, and asked
+//! on **every** partitioner arm rather than only the shipping one. That rules out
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #1817, opened 2026-08-13
test(harvest): un-ignore #1541's convergence guard and re-ground its form pin

nh13:
Representation-Change: none. Tests and one test fixture only; no `src/` file is touched, and no normalized output moves — this change alters which assertions run, not what ferro emits.

Relates to #1541 (stays open — see below), #1484 and #1545. Deliberately no closing keyword: #1541 is an unresolved adjudication and must not auto-close.

## The defect: a guard that gates nothing

`case_harvest::the_four_mer_inversion_pair_converges` was added by #1545 as a red guard for #1541, `#[ignore]`d with the reason *"Un-ignore when #1484 merges."* **#1484 merged on 2026-08-08.** The attribute was never lifted, so the guard has been passing while running on nobody's PR.

Measured myself on `origin/main` at `e3f08263`, not carried forward from the triage that spotted it:

```
cargo nextest run --features dev --test it -E 'test(case_harvest)' --run-ignored all
  -> 24 tests run: 23 passed, 1 failed      (EXIT=100)
     FAIL  case_harvest::the_dup_shaped_split_does_not_touch      <- #1542, must stay red
     PASS  case_harvest::the_four_mer_inversion_pair_converges    <- passing, and ignored
```

(#1545's body recorded *two* red guards in this suite. One of the two stopped reproducing.)

**The un-ignore is not cosmetic — the test genuinely executes in CI.** Checked rather than assumed, because a guard that skips green is worse than no guard:

- **No fixture can be missing.** Both inputs are committed, in-tree JSON — `tests/fixtures/case-harvest/cases.json` (28 KB) and `reference-windows.json` (217 KB). Neither is an LFS object or a release asset, so neither depends on `scripts/fetch-test-fixtures.sh`.
- **There is no skip path, by design.** `cases()` and `window_fixture()` `.expect(...)`, so an absent fixture panics. The module carries no `env::var`, no `FERRO_MANIFEST` gate, and no early `return` — unlike the manifest-backed conformance axes, which skip silently in PR CI. `assert_the_slice_answered_everything` additionally asserts a *positive floor* on served reference reads, so "read nothing, fail nothing" cannot pass.
- **It is in the selection.** `case_harvest` is in neither `SWEEP_FILTER` nor `ORACLE_EXCLUDE`, so it runs in the sharded `test` job (`-E "not test(proptest) and not ($SWEEP_FILTER)"`) and in `test-oracle`, both of which the required `Test` rollup gates on.
- **Nothing was selecting it before.** No job in `ci.yml` passes `--run-ignored`. The single place CI selects ignored tests at all is `external-validation.yml`'s weekly API comparison, and it misses this twice over: it filters to `api_comparison_tests`, and `--ignored` selects *only* ignored tests.

Verified end to end — with the attribute lifted, the CI-equivalent selection runs it:

```
cargo nextest run --features dev --test it -E 'test(case_harvest)'
  -> 23 tests run: 23 passed                (EXIT=0)
     PASS  case_harvest::the_four_mer_inversion_pair_converges
```

22 runnable → 23. `#[ignore]` attributes in the file go 2 → 1.

## The fix

**1. Lift the attribute**, and drop the now-false `defect` from the three `cyp21a2-710-713-four-mer-inversion` rows. That pairing is what `the_corpus_census_is_unchanged` itself demands — *"removing one claims a defect is fixed, which must come with the `#[ignore]` being lifted"* — so the census is re-blessed in the same change (red set 4 → 1 row; red issues `{#1541, #1542}` → `{#1542}`).

The rows keep `expected: null` and their confluence class, so they land in the corpus's existing category for *"the governing ruling is open"*. **That is a coverage increase, not a removal**: `the_confluence_classes_converge` skips any class with a red member, so it had been running 2 of 3 classes and now runs all 3.

**2. Re-ground the exact-form assertion, which was resting on a refuted reading.** This is the part worth reviewing.

`CYP21A2_TARGET` was documented as *"the decided target … not a guess and not a preference"*, grounded on `general.md:56` ranking the competing substitutions (1) above the inversion (3). **That grounding does not stand.** The `decided` ledger record `adjudication-precedence-order`, entry E1, says of that clause:

> `:56` "ranks single-variant TYPE LABELS FOR ONE SPAN — it never ranks a multi-member allele against a spanning description, and an earlier attempt to use it that way was refuted on exactly that ground — so `:56` cannot settle a merge-versus-split question at all".

`c.[710T>A;713T>A]` against `c.710_713inv` is precisely a multi-member allele against a spanning description. The sibling `decided` record `inversion-vs-two-delins-76-83` reads `:56` the other way (*"a genuine two-substitution shape splits"*), so two decided records are in tension over this form — the operator's call, not a test file's. #1541's own 2026-08-12 status says the same in its own words: convergence "arrived as a side effect", no ruling was made, and this locus is a live member of the #1703 family, where "whichever way that is ruled, this row moves with it".

So the guard now carries **two assertions with different standing, and says so**:

| assertion | standing |
|---|---|
| the three spellings agree | correctness claim — the `decided` `confluence-gate-is-apply-equality-on-every-determined-axis`, with sequence-equivalence read from the committed slice by `the_cyp21a2_block_inverts_to_the_two_substitutions_the_sibling_row_writes` |
| they agree on `CYP21A2_TARGET` specifically | **stability tripwire** on `background/basics.md:38` — the corpus's own stated authority for its stability pins, which are explicitly not spec-correctness claims |

The failure message for the second now says to re-bless the constant *citing the ruling* when #1703 is decided, rather than re-pointing it at whatever came out.

## #1541 stays open, and this is not a fix for it

Stated explicitly because the diff could be misread as claiming it. #1541 asked for **a ruling on which form governs, and then convergence on it**. Only the second half happened, and as a side effect. The issue stays open, no closing keyword appears in this body, and the three rows still pin no output for exactly that reason. What this PR claims is narrower: the reproducer stopped reproducing, so a green guard should be gating instead of ignored.

## Unchanged on purpose

- **`the_dup_shaped_split_does_not_touch` (#1542) is untouched** — still `#[ignore]`d, still red, still the sole member of the red set. #1545's rule that neither guard may be weakened to green is respected; #1542 is deferred to the operator with an open ruling.
- **`the_inversion_preference_control_still_holds` is untouched.** It is the other direction of the same discrimination — a whole-block reverse complement competing with `delins` members must stay typed as an `inv` — and it is what stops any fix here from being written as "always prefer the split".
- **No `src/` change, no fixture regeneration.** `reference-windows.json` is derived from the *inputs* in `cases.json`, and no input changed, so it is byte-identical.

## Verification

Every number below was read back from the run's own log, not from a completion notification.

| command | result |
|---|---|
| `cargo fmt --all --check` | `FMT_EXIT=0` |
| `FERRO_SWEEP_SEEDS=full cargo nextest run --features dev` | `SUITE_EXIT=0` — 10522 / 10522 passed, 151 skipped |
| `scripts/run_oracle_suite.sh` | `ORACLE_EXIT=0` — 10385 / 10385 passed, 288 skipped |
| `-E 'test(case_harvest)' --run-ignored all` | 24 run / 23 passed / 1 failed — the failure is #1542's guard |
| `-E 'test(case_harvest)'` (CI-equivalent) | 23 run / 23 passed |

The guard is also stable inside the full suite, not only alone: it passes in both the isolated run above and the full 10,522-test run, so it is not order-dependent.

## Self-review

`/coderabbitai-review` against the local diff. Applied: the first pass left two stale comments the change itself created — the `RED — live defects` section banner (a green test now sits under it) and the "One measurement in the brief did not reproduce" note (it still asserted the class "yields two normal forms") — plus a precision fix on the `--run-ignored` claim, which a grep for `--ignored` would otherwise appear to refute. Rejected, with reasons: *"dropping `defect` weakens the corpus"* — it is the pairing the census assertion itself specifies, and the rows move from a skipped class into a running one; *"drop the exact-form assertion since the form is unruled"* — that would let an unruled form move silently, and a disclosed tripwire is this repo's device for exactly that.

### DIFF
--- a/tests/fixtures/case-harvest/cases.json
+++ b/tests/fixtures/case-harvest/cases.json
@@ -285,12 +285,7 @@
         "clause": "docs/recommendations/general.md:56",
         "quote": "the preferred description is: (1) substitution, (2) deletion, (3) inversion, (4) duplication, (5) insertion"
       },
-      "defect": {
-        "issue": "#1541",
-        "kind": "confluence",
-        "summary": "c.[710T>A;713T>A], c.710_713inv and c.710_713delinsACGA do not converge -- one variant, two normal forms"
-      },
-      "note": "LIVE DEFECT #1541, and this spelling is the DECIDED TARGET the class must converge on. c.710_713 reads TCGT and its reverse complement is ACGA; the interior CG is self-complementary, so the inversion changes only the flanks -- which is exactly what these two substitutions write. VERIFIED EQUIVALENT, and the reference bases at 710 and 713 genuinely are T. THE TARGET IS NOT AN OPEN QUESTION: #1517/#1535 already adjudicated this geometry, and tests/it/issue_1517_inv_priority_over_delins.rs pins both directions -- a whole-block reverse complement competing with SUBSTITUTIONS prefers the split, because general.md:56 ranks substitution (1) above inversion (3); one competing with DELINS prefers inv, because delins is absent from the ranking and so ranks below everything. This row's competitors are two substitutions, so the preferred form is c.[710T>A;713T>A], the same answer as #1230. The opposite direction is the NM_004006.2:c.76_83 class in this same corpus, which is the control preventing the rule from being written as 'always prefer the split'. WHY IT IS STILL RED ON MAIN: not the typing rule but the AXIS GATE -- Normalizer::is_splittable_single_member (src/normalize/mod.rs:2005) matches only HV::Genome/HV::Mt, so a lone c. member never reaches the sequence-first pass; the existing #1230 control passes only because it is written on the g. axis. PR #1484 widens that gate to c./n./r. and all three spellings converge on c.[710T>A;713T>A] under it. Un-ignore the guard in the change that merges #1484."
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #1818, opened 2026-08-13
test(conformance): a codon-designed protein corpus, and the protein axis's first census

nh13:
The protein axis has been *measurable* since the synthetic provider and the denotation oracle landed, and not *measured* — the corpus had no protein rows. This adds them, and reports the first census.

## Why the cores had to be redesigned

`spec_corpus`'s cores are xorshift draws over an `AT`/`ACGT` alphabet with a CDS laid over them at `(tx_len/8 + 1, tx_len - tx_len/8)` — a ratio, not a codon boundary. Such a core need not start at a start codon, need not be a whole number of codons, and may carry an interior terminator. Measured, **0 of 32** are open reading frames (`the_nucleotide_corpuss_cores_are_not_reading_frames` pins it, so the premise fails loudly if it ever stops holding).

`protein_corpus::codon_core` designs the CDS instead — a start codon, a body of sense codons, a terminator — and `ProteinFrame::from_cds` **translates** it. Back-translation would not have done: `ProteinFrame::build` takes the table's first codon per residue, so every CDS it can build holds at most 20 distinct codons and no synonymous pair, and the one shape a protein corpus has that a nucleotide corpus structurally cannot hold is unreachable in it — a tandem repeat of one residue spelled in four different synonymous codons, which is an ambiguous run on the protein axis and no run at all on the nucleotide one.

**#1810's generator (`codon_frame_deep_window_offset.rs`) was read and not reused.** Its two design choices are deliberately wrong here: a uniform `GCT` codon, so no residue identity varies, and no UTR with `cds_start = 1`, so there is nowhere for an N-terminal extension to start or a C-terminal one to read into. What is taken from it is its *finding* — that a site near `c.1` has its canonical window clamped and its phase forced to zero — which is why the designed sites sit at transcript position 160, past `CANONICAL_PAD`.

## The structural-blindness audit

Eleven properties are varied, each demonstrated by a test rather than asserted in prose. The phase knob is the one worth naming: the exon-1 junction's offset from a codon boundary works out to `(body_codons - 1) mod 3` and nothing else, so a generator that fixed the CDS length would have fixed the phase and agreed with itself on the one dimension a junction defect lives in. Three consecutive `body_codons` reach all three phases, and the arithmetic is checked against the frame rather than against itself.

Six properties are declared as limits the corpus **provably cannot reach**: `Sec`/`Pyl` (no non-terminator codon), `Xaa` in a reference, a non-standard genetic code, an incomplete or non-triplet CDS, codon-level ambiguity within one residue, and intronic/UTR positions.

One further limit is a property of the **axis**, not of the generator: a `p.` description has no codons of its own, so transcript geometry may be inert. That is measured, not assumed — `shape_is_inert_for_the_protein_axis` compares every design's outputs across all three exon layouts and both strands, and it is inert today. The phase dimension is therefore a guard against that changing, not a source of signal.

## `Indeterminate` is its own census column

On `rulings[confluence-gate-is-apply-equality-on-every-determined-axis]`: "confluence is asserted only over DECIDED pairs; the `Indeterminate` count is reported alongside and never folded into either side."

`RowKind::Distinguished` applies the same discipline one step further out, on the same record's caveat that "two descriptions can be apply-equal on every axis and still make different epistemic claims". A description and its parenthesised twin denote one protein — the oracle reads through the wrapper deliberately — but `protein/substitution.md:16` makes the parentheses an **evidential** claim, so convergence would destroy information. Filed as a family they produced 90 of the first run's 90 `split_two` rows, every one of them correct behaviour; a confluence figure whose bulk is the normalizer doing the right thing hides the rows where it does not.

## The measurement

540 rows / 1,044 spellings, byte-identical in both shuffle directions:

| figure | value |
|---|---|
| `unparseable_outputs` | 0 |
| `outputs_denoting_no_protein` | 0 |
| `protein_changed` | 0 |
| `non_idempotent_outputs` | 0 |
| `converged` / `split_two` | 198 / 18 (determined families only) |
| `indeterminate_families` | 18, all converged — its own column |
| `distinguished_preserved` | 90 of 90 |
| `conflicts_accepted` | **72 of 72** |

**`conflicts_accepted` 72 of 72 is the rank-1 finding**, and it holds in strict mode as well as lenient — a reversed range (`protein/deletion.md:18`), a non-flanking insertion (`protein/insertion.md:17-18`), two allele members claiming one residue, and a position past the terminator are each normalized and emitted unchanged. `rulings[absolute-prohibition-enforcement-stage]` makes lenient's behaviour arguably decided and strict's not, and records that the ruling is unimplemented (#1630); `no_conflicting_description_is_refused_in_strict_mode_either` pins the protein axis's share of that gap so it fails the day strict starts refusing.

The 18 divergent families are **one class and one open question**: `p.Ala50[2]` against `p.Ala52_Ala53del`. `protein/repeated.md:22` prefers the `[n]` form but conditions it on "when the repeat is variable in the population" — provenance a normalizer holding a protein sequence cannot evaluate. That needs an operator ruling, so it is carried as an `#[ignore]`d test with the question written on the attribute, not pinned to an answer nobody has given.

Two generator defects the drop ledger caught, both fixed here and both invisible without it: a hard-coded `Gly` as a frameshift's "first new amino acid" collided with the reference residue on 9 designs, and an overlapping-members conflict whose first member was a silent identity. The drop ids are now retained on the corpus, because a drop *count* cannot answer "which shape stopped being generated".

**It does not lift the protein slice's blanket clause exemption, and a green census is not evidence for lifting it.** The oracle judges denotation and not validity — a one-for-one `delins`, a `Cys76_Cys76` range and a `p.[Arg76Ser;Cys77Trp]` all denote a protein perfectly well while the recommendations rule against each spelling. All **205** protein clause units stay `not-generatable` in the rule inventory, untouched by this change.

(A figure inherited from `synthetic_protein.rs`'s docs — "149 clauses (102 of them validity requirements)" — does not reconcile: the committed inventory counts 205 clause units over `docs/recommendations/protein/*`, which an independent re-derivation reproduces file for file, and nothing in the repository classifies a clause as a validity requirement. The new docs say so and quote 205; correcting the older comment is left for a change that can derive the right split.)

## CI cost

**~1.3 s**, over 23 tests: the six-frame corpus is rebuilt per test and normalization is 1,044 protein descriptions. Measured against a 105 s `--lib --test it` run, that is ~1.2 %. It belongs in the **default** selection and is not put behind `SWEEP_FILTER` — this repo already has a live example of a guard that quietly stopped running by being moved into a reduced selection, and 1.3 s does not buy that risk.

## Verification

`cargo nextest run --features dev --lib --test it --no-fail-fast` → **10,353 passed, 0 failed**. `cargo fmt --all --check` and `cargo clippy --features dev --all-targets -- -D warnings` clean. The `#[ignore]`d test was run with `--run-ignored all` and confirmed to fail, so the open question is executable rather than decorative.

Representation-Change: none. No file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched; the change adds a conformance corpus, a census and a frame constructor, and moves no normalized output.

### DIFF
--- a/src/conformance/mod.rs
+++ b/src/conformance/mod.rs
@@ -19,6 +19,7 @@ pub mod error_mode_stamp;
 pub mod hgvs_rs_projection;
 pub mod inversion_sweep;
 pub mod mutalyzer;
+pub mod protein_corpus;
 pub mod reference_snapshot;
 pub mod reference_window;
 pub mod schema;
--- a/src/conformance/protein_corpus.rs
+++ b/src/conformance/protein_corpus.rs
@@ -1,0 +1,1809 @@
+//! A codon-designed **protein** conformance corpus, enumerated deterministically.
+//!
+//! # The measurement gap this closes
+//!
+//! [`synthetic_protein`](super::synthetic_protein) supplies a protein reference
+//! coupled to its transcript, and an amino-acid-level denotation oracle over it.
+//! It deliberately adds no corpus rows, so the protein axis has been
+//! *measurable* and not *measured*. This module is the rows.
+//!
+//! # Why the cores had to be redesigned rather than reused
+//!
+//! [`spec_corpus`](super::spec_corpus) draws its cores from a xorshift stream
+//! over an `AT` or `ACGT` alphabet and lays a CDS over them at
+//! `(tx_len/8 + 1, tx_len - tx_len/8)` — a *ratio*, not a codon boundary. Such a
+//! core is an open reading frame only by accident: it need not start with a
+//! start codon, its CDS length need not be a multiple of three, and any of its
+//! interior codons may be a terminator, which truncates the translation. Almost
+//! none of them translate to anything a `p.` description can be written against,
+//! so a protein stratum could not have been drawn from them.
+//!
+//! [`codon_core`] designs the CDS instead: an `ATG`, a body of sense codons
+//! drawn from ferro's own [`CodonTable`], and a terminator. It is handed to
+//! [`ProteinFrame::from_cds`], which **translates** it — so the peptide is
+//! derived from the codons rather than the codons back-derived from a peptide,
+//! and the frame's round-trip check still guarantees that what the provider
+//! serves as `NP_TEST.1` is what `NM_TEST.1`'s CDS translates to.
+//!
+//! Back-translation would not have done. [`ProteinFrame::build`] takes the
+//! table's *first* codon for each residue, so every CDS it can build holds at
+//! most 20 distinct codons and no synonymous pair at all — and the single most
+//! interesting property a protein corpus has over a nucleotide one is
+//! unreachable in it: **a tandem repeat of one residue spelled in four
+//! different synonymous codons**, which is an ambiguous run on the protein axis
+//! and no run whatsoever on the nucleotide axis. [`RUN_RESIDUE`]'s tract is
+//! exactly that shape, and `tests::the_ambiguous_run_is_invisible_to_the_nucleotide_axis`
+//! is the demonstration.
+//!
+//! # #1810's generator was read and NOT reused
+//!
+//! `tests/it/codon_frame_deep_window_offset.rs` builds `"ATG" + "GCT".repeat(199)`
+//! for the *coding* axis's codon exception. Its two design choices are both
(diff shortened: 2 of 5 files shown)


nh13:
@coderabbitai pause

Issue #1816, opened 2026-08-13
Normalizer: extend the sequence-first body to the 5'UTR, and build a derived member that straddles a CDS boundary


nh13:
## Summary

#1650 extended the sequence-first pass's body region past `cds_end`, so a cis allele straddling the CDS/3'UTR boundary is re-derived rather than refused. The **5'UTR half is deliberately not done**, and this issue is what says when it comes back.

Two distinct gaps remain, and they want the same change, which is why they are filed together rather than separately.

## 1. `c.-N` is still refused

`merge::ExtendedBody` folds only `Region::ThreePrimeUtr`, onto `cds_len + N`. That fold is exact and leaves `frame.delta` (`cds_start - 1`) correct for the whole extended range, which is what makes it cheap.

The 5'UTR cannot be done the same way, because it needs the **origin itself** to move. Today `w_lo = c_lo.saturating_sub(pad).max(1)` clamps at `c.1`, so the derivation window never reaches 5' of the CDS start. In transcript coordinates `.max(1)` is the transcript start instead, which lets the window reach into the 5'UTR — a behaviour change for every CDS-start-adjacent variant, not only for the straddling ones this issue is about.

#1650's own write-up names this as needing measurement rather than inheritance, and it was not measured there.

There is a second reason not to inherit it silently. `merge.rs`'s own comment (search `the CDS/UTR and transcript-end insertion clamps`) names the body-region check in `collect_canonical_edits` as the enforcement point for the clamps of #1202, #1205, #1207, #1209, #383 and #387, and for `general.md:44`'s exon-junction exception. #1650 re-supplied that enforcement for the 3' side by taking the pad to zero for an extended group, so a piece can be cut inside the members' own union but carried nowhere outside it. Whatever is done at the 5' end has to re-supply it too, and `.max(1)` moving is exactly the thing that would quietly stop doing so.

## 2. A derived member that straddles the boundary cannot be built

`Anchor` carries **one** `Region` for both endpoints, so a piece running from `c.<cds_len - 1>` to `c.*2` has no anchor to be. `merge::render_on_its_own_region` refuses that shape and the group falls back to the per-member pipeline.

That refusal is the status quo — before #1650 the whole group was refused — so it is not a regression. It is a capability gap, and it is why some straddling inputs still do not converge:

```
CDS 11..=30 on the 40-mer `GCTGTCCGATCAGGTAATGCACATCCGTAGCTTACGGTAC`
[...]

PR #1819, opened 2026-08-13
fix(normalize): derive a cis allele that straddles the CDS end

### LINKED ISSUE #1650: Normalizer: a cis allele straddling a CDS boundary is refused by the sequence-first pass, giving two fixed points for one variant

nh13:
Closes #1650.

## The defect

A cis allele whose members straddle the CDS/3'UTR boundary was refused by the sequence-first pass, so the multi-member spelling and the `inv` spelling of one variant were two fixed points. The identical design converges when the members sit inside the CDS — same bases, same payload, only the stop codon moved:

```text
CDS 11..=30   c.[13_15delinsCTA;18T>G;20G>T] -> itself
              c.13_20inv                     -> c.[13_15delinsCTA;18T>G;20G>T]   converged
CDS 11..=26   c.[13_15delinsCTA;*2T>G;*4G>T] -> itself
              c.13_*4inv                     -> c.13_*4inv                        did NOT
```

This is the refusal #1536 originally named. #1536 turned out to be a different defect with a different cause (the #350 cross-axis bail, fixed by #1651), and this half was left behind it as a measured, `#[ignore]`d, red-by-design guard naming #1650 in both the attribute and the doc comment.

Reproduced on `origin/main` before any code was written — `cargo nextest run --features dev --test it -E 'test(issue_1536)' --run-ignored all` exits **100**, with the guard failing on exactly the two strings above and its CDS-interior control green.

After: all three spellings, in both directions, reach one point.

```text
CDS 11..=26   c.[13_15delinsCTA;*2T>G;*4G>T]  ┐
              c.13_*4inv                      ├-> c.[13_15delinsCTA;*2T>G;*4G>T]
              c.13_*4delinsCTACGGAT           ┘
```

The third is the "lone-`delins` symptom" #1650 names: #1651 left it converging with `c.13_*4inv` while *neither* was split into the three members its CDS-interior twin gets.

## Why a fix is owed, given that rule 3 is best effort

Two fixed points for one variant is a rule 3 (confluent) failure, and `README.md` marks rule 3 *best effort*. That does not by itself settle whether a fix is owed, so it was checked rather than assumed:

- `README.md` bounds best effort by **the spec's determinacy and what a normalizer's inputs can decide — "not by ferro's implementation quality"**, and states that a rule 2 or 3 failure caused by ferro's code is a bug under rule 7. None of the four listed ways the spec can fail to determine an answer applies here, and the CDS-interior control is the proof: if the spec did not determine an answer for this sequence, the interior twin would diverge too.
- Which form wins needs no new adjudication. `canonical-form-choice-when-both-legal` (`decided`) supplies the method — derive from the resulting sequence — and the guard asserts *convergence*, not a string. `general.md:56` ranks substitution above inversion and two of the three competing members are substitutions, so the split falls out. (`inversion-vs-two-delins-76-83` governs the other case — a whole-block reverse complement whose competitors are all `delins`, which `:56` does not rank. Here they are not.)
- `confluence-gate-is-apply-equality-on-every-determined-axis` (`decided`) is what makes these one equivalence class: single-exon transcript, same span, both axes agree.

## Correcting the issue on one point, because it changes the fix

#1650 says of the two refusals: *"Neither refusal is a policy. `merge`'s is an accident of doing the window arithmetic in **axis** coordinates."*

**That is wrong about `merge`'s.** `src/normalize/merge.rs` names the body-region check in `collect_canonical_edits` as the enforcement point for the CDS/UTR and transcript-end insertion clamps of #1202, #1205, #1207, #1209, #383 and #387, and for `general.md:44`'s exon-junction exception. Relaxing it removes an enforcement point along with the refusal, so the fix has to re-supply it — which is what the zero pad below does.

The half of the issue's diagnosis that *is* right is the mechanism: `simple_cds_pos` returns a coordinate *within* a region, those numbers are not on one line, and `edit_span_union` and the `w_lo`/`w_hi` arithmetic cannot be done across a boundary.

## The fix

`merge::ExtendedBody` folds `c.*N` onto `cds_len + N` — the coordinate it already has on the transcript once `frame.delta` (`cds_start - 1`) is added. The fold is therefore exact rather than an approximation, and `frame.delta`, `crosses_exon_junction`, `enclosing_exon` and `start0`/`end0` all keep working unchanged for the whole extended range. `render_on_its_own_region` is the inverse, applied to the rebuilt members.

Scoped three ways. Each is a measurement, not a preference:

1. **3'UTR only.** `c.-N` needs the origin itself to move, which changes what `w_lo`'s `.max(1)` clamp *means* — today it clamps at `c.1`, in transcript coordinates it would clamp at the transcript start and let the window reach into the 5'UTR. That is a behaviour change for every CDS-start-adjacent variant, not only the straddling ones. #1650 names it as needing its own measurement; it is filed as #1816 rather than inherited here.
2. **Straddling groups only.** A group sitting wholly in the 3'UTR keeps its refusal. Measured: admitting it regressed #1284's `c.[*10dup;*11dup]` to `c.[*11dup;*11dup]` — two members claiming one base, which round-trips through the sequence check because two `dup`s at one anchor insert the same bases. #1650 is about a span the boundary *cuts*; nothing is owed to a span it does not touch.
3. **The pad goes to zero for an extended group.** This is what re-supplies the clamp the body-region check was carrying. The pad is the only room `shift_pieces` has, so with it gone a piece can be cut anywhere inside the members' own union and carried nowhere outside it. That is #1651's own argument for its carve-out — "re-typing … reads the reference under the member's own span and moves nothing" — applied to the partition instead of to the typing.

## What this also fixes, and what it does not

**Fixed: the CDS-end non-idempotency family, 4 → 0 in both directions.** The class was `c.*1delinsCTT -> c.72_*1insCT -> c.72delinsCCT`; the second pass was answering a question the first could not see, because the first refused the group. It now settles in one pass. `defect_non_idempotent_outputs`'s pins are flipped accordingly, following the `every_scale_family_is_now_a_fixed_point` template that module set when the three `scale-*` rows were fixed.

**NOT fixed: the four rows that denote different bases.** `spec_conformance_axis`'s `sequence_changed` reads **4 at 3' either side of this branch**. The rows are respelled — `c.[72del;72_*1insA]` was `c.[72delinsTA;*2del]` and is now `c.*1T>A` — and both spellings denote `CGC` where the input denotes `GCC`. The spelling moved; the class did not. `spec_corpus_regressions`'s pins are re-blessed to the new strings and say plainly that this is a re-bless of a defect, not a fix.

That correction matters beyond the numbers: `the_cds_end_flush_pair_is_its_two_members_normalized_separately` explained the class as *"ferro normalizes the members separately and concatenates"*. The group is now genuinely re-derived — the test asserts the concatenation identity has stopped holding — and the wrong answer survives that. So concatenation was the route, not the root cause, which localises the remaining defect to how the boundary run is partitioned. The test records this.

**NOT fixed: the re-typing asymmetry itself.** The interbase immediately 3' of the last CDS base still re-types a two-base insertion into a `delins` where all 94 others do not. `an_insertion_immediately_after_the_last_cds_base_is_retyped_as_a_delins` keeps that sweep. This PR removed the disagreement between two passes, not the asymmetry between one interbase and the rest.

## A guard that reads as coverage and guards nothing

`apply_coding_codon_exception` gained a `within_cds` guard: `general.md:35`'s exception is a conjunction — "two variants separated by one nucleotide, **together affecting one amino acid**" — and the second conjunct is unstatable past the stop codon, while `same_codon` is `(a - 1) / 3` and answers happily for a codon that does not exist.

**It is marked defensive and NOT demonstrated, because sabotage-checking it twice showed the test was vacuous.** Two inputs were built to reach the line and neither does — `c.[*1C>G;*3T>A]` is declined at the straddle gate, and `c.[16C>G;*2T>A]` is admitted but still does not merge with the guard forced open. Both stayed green with `within_cds` forced to `true`.

Rather than ship a test that reads as coverage, the test is downgraded to a recorded measurement carrying the table of why each input misses, and `merge.rs` says the line is defensive. Whoever makes it reachable owes it a real sabotage check. The guard is kept rather than deleted because the failure direction is silent.

## Representation stability

Measured with the committed harness, `examples/dump_normalized_corpus`, dumped at `439617c2` and at this branch and diffed by `--compare` — not composed from deltas.

| | rows |
|---|---|
| total | 85 642 |
| **moved** | **780 (0.9 %)** |
| of which previously **accepted** (a migration) | 534 |
| of which previously **not** a fixed point (free) | 246 |

**The property this keys on is a `c.` member at or across `cds_end`, and it is not a structural zero.** The generator's own committed guards assert the coding corpus emits 3'UTR (`c.*N`) rows *and* rows straddling `CDS_END`; `CDS_END = 15` on a 20-mer gives the single-exon axis a real 3'UTR, and #1478's `cx.` axis adds `CDS_START_MULTI = 4` / `CDS_END_MULTI = 15`. Movement is confined to the families that can carry such a member — `dup_plus_sub` 308, `delins_hiding_an_inversion` 140, `del_plus_sub` 110, `dup_plus_ins` 93, `split_vs_spanning_delins` 79, `adjacent_junction_ins` 41, `junction_interior_to_span` 5, `long_block_inversion` 4 — and every protein family, `inv_member`, `nested_spans`, `partial_overlap_spans`, `coincident_bounds`, `identity_member` and `three_member_allele` move **0**, which is the expected zero rather than a blind one.

Axis censuses, both directions, re-blessed in the same commit:

| | 3' before | 3' after | 5' before | 5' after |
|---|---|---|---|---|
| converged | 9402 | **9442** | 9228 | **9276** |
| split 2 / 3 / 4+ | 2223 / 226 / 31 | 2201 / 214 / 25 | 2462 / 168 / 24 | 2440 / 142 / 24 |
| non-idempotent | 4 | **0** | 4 | **0** |
| sequence_changed | 4 | 4 | 0 | 0 |
| every other rank-1 counter | — | unchanged | — | unchanged |

The three `split_*` decreases sum exactly to the `converged` increase in each direction (22 + 12 + 6 = 40; 22 + 26 + 0 = 48). **That identity is quoted as a NET statement**, because the module's own warning is that the counters cannot tell net from gross. What is checked directly is the reported divergence set, whose 12 family ids are byte-identical either side and contain no `cds-end` row.

`coding_axis_separation_two_or_more_merges` stays **0 of 997**, so no allele separated by two or more unchanged nucleotides was merged.

## Deliberately left open

- **#1816** — the 5'UTR half, plus `Anchor` carrying a region per endpoint so a *derived* member that straddles the boundary can be built rather than declined. The two want the same change, which is why they are filed together.
- `uses_extension` runs `collect_canonical_edits` a second time on a coding group, so such a derivation lowers its member list twice. A `cds_end_axis.is_some()` short-circuit — provably equivalent, since with no CDS end `extended` *is* `OFF` — removes it for every genomic, mitochondrial, `n.` and non-coding-`r.` derivation. Removing the coding-group case needs `collect_canonical_edits` to report whether it folded, a wider signature change than this fix wants. **Flagged for a reviewer** rather than left implicit.
- `ORACLE_EXCLUDE` is not narrowed. Its reason-2 (five tests assert the non-idempotency the oracle panics on) no longer holds, and both `ci.yml` and `scripts/run_oracle_suite.sh` are corrected to say so; the exclusion is kept for reason 1 and for the denoted-sequence rows, and narrowing it needs its own measurement.

## Verification

All on this branch at `05adb623`, rebased onto `439617c2`, with every exit status read back out of the log rather than from a completion notification.

| command | rc | result |
|---|---|---|
| `cargo clippy --all-features --all-targets -- -D warnings` | 0 | clean (one pre-existing third-party future-incompat notice for `proc-macro-error2 v2.0.1`) |
| `cargo nextest run --features dev --lib --test it` | 0 | `10339 tests run: 10339 passed, 149 skipped` |
| `scripts/run_oracle_suite.sh` | 0 | `10394 tests run: 10394 passed, 287 skipped` |
| `dump_normalized_corpus --out` ×2 + `--compare` | 0 | 85 642 rows each side, 780 moved |
| `... -E 'test(issue_1536)' --run-ignored all` on `origin/main` | 100 | the guard, red by design, before any code |

`scripts/run_oracle_suite.sh` rather than a bare armed run: a bare `FERRO_ASSERT_IDEMPOTENT=1 cargo nextest run` cannot pass on `main` and never could, for the reasons that script's header sets out.

Both public normalizing entry points are reached — `canonicalize_from_sequence` is called from the shared path that `normalize()` and `normalize_with_diagnostics()` both run — so the idempotency oracle covers this change on both exits. No new early return bypasses the shared checked exit; the added `return None` is inside `canonicalize_from_sequence`, whose `None` is the established "decline and let the per-member pipeline stand".

No reference-aware test was added, so nothing here depends on `FERRO_MANIFEST` and everything above runs on PR CI.

Representation-Change: 1671 of 96 182 corpus rows move (1.7%) — 1191 were previously accepted forms (a migration) and 480 were not fixed points, so moving those is free. Every move is a cis allele straddling `cds_end` being re-derived from the bases it already denoted; no move changes what a description denotes, and `sequence_changed` is unchanged at 4/0. Both axis censuses are re-blessed on the current base: converged 11 016 -> 11 056 at 3' and 10 755 -> 10 803 at 5', with the three `split_*` decreases summing exactly to each gain (22+12+6 and 22+26+0), and `non_idempotent_outputs` 4 -> 0 in both directions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,16 +206,27 @@ env:
   # before #1599 lowered the count it is derived from.
   #
   # WHY, REASON 2 — simpler, and not covered by the paragraphs above. Five tests
-  # in these modules do not merely count the defect, they ASSERT it: the four in
-  # `defect_non_idempotent_outputs` and
-  # `spec_corpus_regressions::an_insertion_at_the_cds_end_is_not_a_fixed_point`
-  # pin `c.*1delinsCTT` -> `c.72_*1insCT` -> `c.72delinsCCT` as a known
+  # in these modules used to not merely count the defect but ASSERT it: the four
+  # in `defect_non_idempotent_outputs` and
+  # `spec_corpus_regressions::an_insertion_at_the_cds_end_is_a_fixed_point`
+  # pinned `c.*1delinsCTT` -> `c.72_*1insCT` -> `c.72delinsCCT` as a known
   # non-fixed point. A test that pins a defect and an oracle that aborts on it
-  # cannot both run, whatever the census does. Together the two reasons account
-  # for all 7 tests a bare armed run fails, and `FERRO_ASSERT_IDEMPOTENT` alone
-  # accounts for all 7 — `FERRO_ASSERT_REPARSE` and `FERRO_ASSERT_IN_BOUNDS`
-  # each contribute 0, and a prepared reference changes nothing (these modules
-  # are `MockProvider`-backed).
+  # cannot both run, whatever the census does.
+  #
+  # #1650 CLOSED that class. The chain is now `c.*1delinsCTT` ->
+  # `c.72delinsCCT` in one pass, `non_idempotent_outputs` reads 0 in both
+  # directions, and all five tests assert the fixed point instead — which is why
+  # the fifth is now named `..._is_a_fixed_point`. REASON 2 therefore no longer
+  # holds for `FERRO_ASSERT_IDEMPOTENT`, and the exclusion rests on REASON 1
+  # (the censuses count what the oracle would turn into `declined`) plus the
+  # denoted-sequence rows named below. It is deliberately NOT narrowed here:
+  # these modules build their own references and sweep them, so an armed run
+  # that began panicking mid-sweep would EMPTY a sweep rather than redden it,
+  # and establishing that it would not needs its own measurement with
+  # `tests/it/oracle_exclude_invariant.rs` updated in the same change.
+  #
+  # `FERRO_ASSERT_REPARSE` and `FERRO_ASSERT_IN_BOUNDS` each contribute 0, and a
+  # prepared reference changes nothing (these modules are `MockProvider`-backed).
   #
   # Locally, `scripts/run_oracle_suite.sh` applies this same exclusion, reading
   # it from here rather than copying it.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -170,8 +170,10 @@ a trap:
 FERRO_ASSERT_IDEMPOTENT=1 cargo nextest run --features dev   # ALWAYS RED on main
 ```
 
-It fails **7** tests on a clean `origin/main`, and has for as long as the spec
-corpus has existed. Use the local runner instead, which mirrors `ci.yml`'s
+It has been red for as long as the spec corpus has existed. The count was **7**
+before #1650 closed the idempotency half; it is deliberately not restated here as
+a number, because nothing checks it and a stale one reads as a measurement — read
+it off a run. Use the local runner instead, which mirrors `ci.yml`'s
 `test-oracle` job — its flags, and its selection:
 
 ```bash
@@ -180,20 +182,27 @@ scripts/run_oracle_suite.sh --print-selection   # what it would run, without run
 scripts/run_oracle_suite.sh -E 'test(my_test)'  # extra args reach nextest
 ```
 
-**Why the bare form is red, and why that is not a coverage gap to close.** All 7
+**Why the bare form is red, and why that is not a coverage gap to close.** The
 failures come from `FERRO_ASSERT_IDEMPOTENT` alone — measured: `FERRO_ASSERT_REPARSE`
 and `FERRO_ASSERT_IN_BOUNDS` each contribute **0**, and a prepared reference
-changes nothing (the same 7 fail with and without `FERRO_MANIFEST`; every one of
-these modules is `MockProvider`-backed). All 7 live in modules `ORACLE_EXCLUDE`
+changes nothing (the same set fails with and without `FERRO_MANIFEST`; every one of
+these modules is `MockProvider`-backed). All of them live in modules `ORACLE_EXCLUDE`
 already names, for two reasons that are worth keeping apart:
 
-- **5 of them ASSERT the defect the oracle PANICS on.** The four
+- **Some of them USED TO ASSERT the defect the oracle PANICS on.** The
   `defect_non_idempotent_outputs` tests and
-  `spec_corpus_regressions::an_insertion_at_the_cds_end_is_not_a_fixed_point`
-  pin `c.*1delinsCTT` → `c.72_*1insCT` → `c.72delinsCCT` as a known non-fixed
-  point. A test that pins a defect and an oracle that aborts on it cannot both
-  run; there is no version of this the flag could pass.
-- **2 of them COUNT it**, and arming the oracle makes the count read *better*
+  `spec_corpus_regressions::an_insertion_at_the_cds_end_is_a_fixed_point`
+  pinned `c.*1delinsCTT` → `c.72_*1insCT` → `c.72delinsCCT` as a known non-fixed
+  point, and a test that pins a defect and an oracle that aborts on it cannot both
+  run. **#1650 closed that class** — the chain collapses to
+  `c.*1delinsCTT` → `c.72delinsCCT` in one pass, `non_idempotent_outputs` reads 0
+  in both directions, and those tests now assert the fixed point (which is why the
+  last one is named `..._is_a_fixed_point`). So this reason no longer holds for
+  `FERRO_ASSERT_IDEMPOTENT`. The exclusion is kept anyway, deliberately: see
+  `scripts/run_oracle_suite.sh`'s header for why, and note that narrowing it needs
+  its own measurement with `tests/it/oracle_exclude_invariant.rs` updated in the
+  same change.
+- **Others COUNT it**, and arming the oracle makes the count read *better*
   than the truth. `spec_conformance_axis`'s censuses wrap normalization in
   `catch_unwind`, so a panicking row is filed `declined` and never reaches its
   family's output set. See `ORACLE_EXCLUDE`'s comment in `ci.yml` for the
@@ -304,7 +313,7 @@ normalized description names may be past the end of its own sequence.
 
 `scripts/run_oracle_suite.sh` arms it too. As with the re-parse oracle,
 `FERRO_ASSERT_IN_BOUNDS` on its own fails **0** tests over `ORACLE_EXCLUDE`'s
-modules — the 7 failures a bare armed run shows are the idempotency oracle's,
+modules — the failures a bare armed run shows are the idempotency oracle's,
 and reading them as this one's would send you after the wrong invariant.
 
 It exists because the class kept being found by hand, one shape at a time —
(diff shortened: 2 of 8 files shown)


nh13:
@coderabbitai pause

Issue #1815, opened 2026-08-13
normalize: the NG_ oracle-fire class is invisible to gating CI (the seven rows themselves were an SPDI frame defect, see #1820)


nh13:
> **Reduced 2026-08-13.** This issue was filed as a normalizer defect — "seven `NG_` rows take a 3' shift past a non-matching base and denote a different sequence". **That half is refuted and has been struck.** All seven were false positives of **#1820** (`hgvs_to_spdi` stamps the compound `NG_xxx(NM_yyy)` accession on a coordinate it resolved on the *inner transcript*, so the provider resolves it to the genomic parent and the oracle compared the wrong bases). Measured in both frames, every one of the seven shifts is legal on the transcript's own bases; a probe applying only the accession fix turned all seven `Differ` → `Agree` and the selection `6 failed / EXIT=100` → `95 passed / EXIT=0`. Root cause is fixed by **#1821**. The refutation, with the per-row bases, is in [this comment](https://github.com/fulcrumgenomics/ferro-hgvs/issues/1815#issuecomment-5283574180).
>
> Implementing what this issue originally asked for would have been **actively harmful**: hardening the 3' shift against the *parent's* bases contradicts the decided record `c-and-n-positions-are-flat-transcript-offsets`, re-introducing #1619's defect from the opposite side.
>
> **What survives is the second half, and it is untouched by the refutation** — indeed the refutation is evidence for it. The original body is preserved in the edit history.

## The defect

**A denoted-sequence violation on a manifest-gated axis cannot redden any gating check.** `FERRO_ASSERT_SEQUENCE` (#1615) is the only oracle that sees this class, and there is no gating job in which it is both armed *and* pointed at the axes that need it.

This is the same failure shape the repo has recorded twice before for other reasons: **a skip is indistinguishable from a pass.** Here the skip is the absent manifest rather than an absent fixture.

## Why gating CI cannot see it

Two independent reasons, and either alone is sufficient:

1. **These axes are manifest-only.** `mutalyzer_normalize_tests` early-returns without `FERRO_MANIFEST`, and no pull-request job sets one. A skip is indistinguishable from a pass in the summary line.
[...]

PR #1821, opened 2026-08-13
fix(spdi): emit a transcript-axis SPDI on the transcript, not the compound accession

### LINKED ISSUE #1820: `hgvs_to_spdi` emits a transcript-frame position under the **genomic parent's** accession for a compound `NG_xxx(NM_yyy)` reference, so every consumer reads the wrong bases
## One-line summary

For a compound reference, `src/spdi/convert.rs` resolves the **coordinate** on the inner transcript and then stamps the **whole compound string** on the SPDI beside it. The provider resolves that string to the genomic parent, so a transcript offset indexes the parent's sequence. `FERRO_ASSERT_SEQUENCE` is the consumer that makes it visible, and it is emitting false positives on the entire accession shape.

Measured on `origin/main` @ `439617c2`, against the prepared reference.

## Mechanism, in two lines of one call chain

`cds_to_spdi_with_provider` resolves the position correctly:

```rust
// src/spdi/convert.rs, resolve_cds_to_tx
let tx_id = accession.transcript_accession();          // "NM_000532.5"
let transcript = provider.get_transcript(&tx_id)?;     // cds_start = 37
…
let tx_len = transcript.sequence_length();             // 1799 — the transcript's own length
let s_u = ensure_tx_in_bounds(s.base, tx_len, "c", start)?;
```

`Accession::transcript_accession`'s own doc says what that string is for: *"the bare accession string for **provider/transcript lookups**, stripping any genomic context wrapper."* So `NG_008939.1(NM_000532.5):c.156` resolves as `cds_start + 156 - 1` = the transcript's base 192, bounded against the transcript's 1799 bases. The number is an offset **on the transcript**.

Two lines later, the accession that travels with that number is not:

```rust
    emit_spdi_for_edit(
        variant.accession.to_string(),   // Display -> "NG_008939.1(NM_000532.5)"
        start_tx,                        // 192, an offset on NM_000532.5
        …
```

That string is both the SPDI's `sequence` field **and** the key every short-form edit's reference fetch is made against (`fetch_reference_bases` inside `emit_spdi_for_edit`).

The provider does not reject it. `MultiFastaProvider::resolve_name` reaches a record through its version-strip fallback — `"NG_008939.1(NM_000532.5)".split('.').next()` is `"NG_008939"` — landing on the genomic parent:

```
get_sequence("NG_008939.1",              191, 197) = CACACA     (the NG_ parent)
get_sequence("NG_008939.1(NM_000532.5)", 191, 197) = CACACA     <-- same record
get_sequence("NM_000532.5",              191, 197) = ACGCCG     <-- the frame the 191 is in
```

Control: a bare transcript (`NM_000532.5`) and a bare `LRG_24t1` both resolve to their own bases, so this is specific to the compound spelling.

The same five call sites are involved on all three transcript axes — `tx_to_spdi_simple`, `rna_to_spdi_simple`, `cds_to_spdi_with_provider`, `tx_to_spdi_with_provider`, `rna_to_spdi_with_provider`. The **genomic** axes (`genome_`/`mt_`/`circular_`) are correct as they stand: there the coordinates *are* the parent's, and the fallback lands on the right record.

## The SPDI itself is wrong, before any oracle is involved

This is not only an oracle bug. `hgvs_to_spdi` fills the mandatory `del` field from that fetch, so the triple it publishes states bases that are not at the position it names:

```
NG_008939.1(NM_000532.5):n.192_197del  ->  NG_008939.1(NM_000532.5):191:CACACA:
NM_000532.5:n.192_197del               ->  NM_000532.5:191:ACGCCG:
```

Same variant under ferro's own coordinate resolution, two triples, and the first one's `del` is six bases the transcript does not have at 191. `canonical_spdi` publishes that as a **key**, so the two spellings key differently.

`NG_008939.1(NM_000532.5)` is also not a valid SPDI `sequence` in the first place — it is HGVS syntax, not an accession, and SPDI's `sequence` must be the accession the position is an offset on.

## The consumer that makes it visible

`compare_denoted_sequences` fetches its comparison window from the SPDI's own `sequence` field, so it compares two descriptions against a sequence neither of them is numbered on:

```
FERRO_ASSERT_SEQUENCE: normalization changed the sequence the description denotes
  over NG_008939.1(NM_000532.5) [191, 197):
    reference       CACACA        <-- the parent; n.192_197 is ACGCCG
    input applies   TTTACTCGC
    output applies  TTTACTCAC
```

With the parent's bases the two forms differ at one base; with the transcript's they are identical. The same mechanism also produces `NotComparable(InputDenotesNoSequence)` for perfectly valid descriptions, because an input that *states* its deleted bases fails `StatedBasesMismatch` against the parent's.

## Measurement

The nightly's reference-aware selection, armed, before and after fixing the accession handed to `emit_spdi_for_edit` on the three transcript axes. Same tree, same manifest, same selection.

```sh
SEL='test(/mutalyzer_normalize_tests::/) | test(/biocommons_normalize_tests::/) | test(/hgvs_rs_projection_tests::/)'
  cargo nextest run --features dev --no-fail-fast -E "$SEL"
```

```
before:  Summary [39.477s] 170 tests run: 164 passed, 6 failed, 10514 skipped   EXIT=100
after:   Summary [19.221s] 170 tests run: 170 passed,          10521 skipped    EXIT=0
```

Both runs took 19–39 s, so neither is the sub-second manifest-absent skip.

**The 6 became comparisons that agree, not skips.** That distinction is the whole question for a fix like this, so it was measured directly rather than inferred from a green summary — `compare_denoted_sequences` was called on each row's `(input, normalize(input))` pair and every one returns `Agree`:

| input | normalized | verdict after |
|---|---|---|
| `NG_012337.1(NM_012459.2):c.4_5insGTA` | `c.5_6insTAG` | `Agree` |
| `NG_012337.1(NM_012459.2):c.[10del;23_25del;36_37insAAT]` | `c.[10del;23_25del;37_38insATA]` | `Agree` |
| `NG_029724.1(NM_004321.7):c.101del` | `c.102del` | `Agree` |
| `NG_008835.1(NM_001168390.2):c.*3187del` | `c.*3188del` | `Agree` |
| `NG_008835.1(NM_001168390.2):c.*3186del` | `c.*3188del` | `Agree` |
| `NG_012337.1(NM_012459.2):c.-1_*1del` | `c.1_*2del` | `Agree` |
| `NG_012337.1(NM_012459.2):c.6delinsCCCC` | `c.6_7insCCC` | `Agree` |

## Relationship to #1815 — please read before acting on either

**Every one of #1815's seven rows is this defect**, and its diagnosis — "a 3' shift applied without checking that the shift preserves the sequence" — does not survive re-measurement. This is stated with evidence rather than as an assertion, because #1815 is a careful issue and its rows are real fires; what is wrong is only which side of the comparison is at fault.

Worked example, #1815's row 3, `NG_029724.1(NM_004321.7):c.101del` → `c.102del`.

`NM_004321.7` has `cds_start = 248` (read from `get_transcript`), so `c.101` is transcript base `248 + 101 - 1 = 348`, i.e. 0-based 347 — which is exactly the position `hgvs_to_spdi` emits. The shift by one is sequence-preserving iff `ref[347] == ref[348]`:

```
NM_004321.7 [343, 351) = AGCACCAC     ref[347] = C,  ref[348] = C   -> equal, shift is LEGAL
NG_029724.1 [343, 351) = GGCATGGA     ref[347] = T,  ref[348] = G   -> unequal
```

#1815 quotes `TG` here and concludes the shift needs `T == G`. That `TG` is the **parent's** bases at a **transcript** coordinate. The normalizer read the transcript and shifted correctly; the oracle read the parent and called it a corruption.

The issue says each `reference` string "was fetched independently with `MultiFastaProvider::get_sequence` on the `NG_` accession". That is true, and it is the trap: the independent fetch used the bare `NG_` accession, which is the same record the buggy path resolves to, so it reproduces the oracle's error rather than checking it.

Two further details worth recording so they are not re-derived:

- The independent fetch is not even guaranteed to hit the record the oracle used. `get_sequence_length("NG_012337.1")` is 15948 while `get_sequence_length("NG_012337.1(NM_012459.2)")` is 39784 — the compound spelling's length resolves through `versioned_for("NG_012337")` to a later version. The two happened to agree on the bases quoted.
- Row 7 (`c.6delinsCCCC` → `c.6_7insCCC`) is not a dropped deletion. On the transcript, `NM_012459.2 [31, 39) = TGCGCAAA`, so the base at 35 is `C`; deleting a `C` and inserting `CCCC` is the same as keeping it and inserting `CCC`. The `T` #1815 quotes is the parent's base.

**#1815 should not be closed on this issue's word** — that is for its filer to decide, and its second observation (that these axes are invisible to gating CI, for two independent reasons) is correct and unaffected. What this issue asks is that nobody "fix" the normalizer's 3' shift for these rows: the shift is legal, and hardening it against the parent's bases would break the transcript axis.

## Not covered by existing issues

- **#315** (closed) is a different mechanism: `FastaProvider::get_sequence` short-circuiting through the transcript registry when one id is registered as both a transcript and a contig. Same family — silent wrong-frame resolution — but the opposite direction, a different provider, and it does not reach a compound accession.
- **#1619** (closed) is the nearest sibling and the right precedent: `hgvs_to_spdi` resolved a `c.` position by walking the exon table while the normalizer indexed the flat transcript. Same seam, same oracle, same shape of answer — the coordinate and the sequence must be in one frame. That one fixed the coordinate; this one is the accession.
- **#1177**, **#480**, **#737**, **#851** are about *projecting* `NG_(NM_)` inputs onto a genomic axis. This issue changes no projection and no coordinate.
- `FERRO_ASSERT_IDEMPOTENT`, `FERRO_ASSERT_REPARSE` and `FERRO_ASSERT_IN_BOUNDS` all pass on every one of these rows.

## Fix

The narrow one: hand `emit_spdi_for_edit` the accession the coordinate resolution already computed — `accession.transcript_accession()` — on the three transcript axes, and leave the genomic axes on `Display`.

The two sites this could instead have gone are both worse:

- **`MultiFastaProvider::get_sequence`**, made to serve the inner transcript for a compound accession. Largest blast radius (every caller), and arguably wrong on its own terms: under HGVS, `NG_x(NM_y)` names the `NG_` sequence with a transcript-based numbering, so "the sequence of `NG_x(NM_y)`" has a defensible claim to being the parent's. It would also silently change what the `g.` axis reads.
- **`compare_denoted_sequences`'s accession selection**, i.e. fixing only the oracle. It leaves `hgvs_to_spdi` publishing a triple whose `del` field states bases that are not there, and `canonical_spdi` keying two spellings of one variant differently. The oracle is the messenger.

Splitting the SPDI's *label* from its *fetch key* also does not work: the `sequence` field **is** the fetch key for `apply_to_reference` and for `fetch_window`, so a compound label just moves the wrong fetch one hop downstream.

### What this deliberately does not settle

Whether a `c.` position on an `NG_(NM_)` reference should be a **flat transcript offset at all** is a separate and larger question. HGVS's compound form exists partly so intronic offsets (`c.93+1`) can be written against the genomic parent, which implies the position should be projected onto the parent's frame through the transcript's exon structure. Ferro resolves the exonic case as a flat transcript offset. This issue makes the SPDI *self-consistent with the frame ferro actually uses*; it does not adjudicate which frame is right. If that frame changes, this fix moves with it — the invariant it installs is "the accession names the sequence the number indexes", which holds either way.

nh13:
Closes #1820.

Representation-Change: SPDI output only (`hgvs_to_spdi` / `canonical_spdi`), and only for compound `NG_xxx(NM_yyy)` references on the c./n./r. axes — the `sequence` field becomes the inner transcript and the del/ins bases it carries become the transcript's instead of the genomic parent's. No HGVS description string changes: normalize and project output is byte-identical, and the g./m./o. axes are untouched.

```
NG_008939.1(NM_000532.5):n.192_197del
  before ->  NG_008939.1(NM_000532.5):191:CACACA:      (parent's bases at a transcript offset)
  after  ->  NM_000532.5:191:ACGCCG:                   (= what the bare NM_000532.5 spelling already gave)
```

The affected rows' normalized outputs are byte-identical (`c.101del` → `c.102del`, and so on for the other six). The genomic axes keep `Display` deliberately: there the coordinates *are* the parent's, so stripping the wrapper would name the transcript for a genomic offset — the same defect with the frames swapped.

## What was wrong

`src/spdi/convert.rs` resolved the coordinate on the inner transcript and then stamped the whole compound string on the SPDI beside it. `resolve_cds_to_tx` calls `accession.transcript_accession()`, reads *that* transcript's `cds_start`, and bounds the result against *that* transcript's length — so the number is an offset on the transcript — while the sibling call two lines below passed `variant.accession.to_string()`, which `Display` renders as `NG_008939.1(NM_000532.5)`.

That string is both the SPDI's `sequence` field and the key every short-form edit's reference fetch is made against, and `MultiFastaProvider::resolve_name` reaches a record for it through the version-strip fallback (`"NG_008939.1(NM_000532.5)".split('.').next()` is `"NG_008939"`):

```
get_sequence("NG_008939.1",              191, 197) = CACACA   (the genomic parent)
get_sequence("NG_008939.1(NM_000532.5)", 191, 197) = CACACA   <-- same record
get_sequence("NM_000532.5",              191, 197) = ACGCCG   <-- the frame the 191 is in
```

So the published triple stated bases that are not at the position it named, `canonical_spdi` keyed two spellings of one variant differently, and `compare_denoted_sequences` — which fetches its comparison window from the SPDI's own `sequence` field — compared both sides against a sequence neither is numbered on.

## Measurement

The nightly's reference-aware selection, armed, same tree and manifest, only the diff differing:

```
before:  Summary [39.477s] 170 tests run: 164 passed, 6 failed, 10514 skipped   EXIT=100
after:   Summary [19.221s] 170 tests run: 170 passed,          10521 skipped    EXIT=0
```

Both runs are 19–39 s, so neither is the sub-second manifest-absent skip.

**The six became comparisons that agree, not skips** — the distinction that matters for a change like this, so it was measured rather than inferred from a green summary. Each row's `(input, normalize(input))` pair was handed to `compare_denoted_sequences` directly, and all seven distinct inputs return `Agree`.

Full suite on this branch: `10534 tests run: 10534 passed`, `EXIT=0`. `cargo fmt --check` and `cargo clippy --all-features --all-targets -- -D warnings` both `EXIT=0`.

## Tests

`tests/it/compound_accession_sequence_frame.rs` is **hermetic**, so it runs in gating CI rather than only under `FERRO_MANIFEST`. It pins both directions, because a fix that made every compound row *skip* would have removed the false positives and the coverage together:

- the oracle must **agree** on a shift that preserves the transcript's sequence, and
- must **still fire** on a compound reference whose sequence really does change.

It also asserts its own premise — that the compound spelling really does serve the parent's bases and that the two records differ over the window — so it fails loudly rather than silently stopping testing anything if that provider behaviour ever changes.

## Please read #1820's "Relationship to #1815" section before acting on #1815

Re-measurement says all seven of #1815's rows are this defect rather than an unchecked 3′ shift, and #1820 lays out the evidence row by row. Nothing here closes, relabels or edits #1815 — that is its filer's call, and its second observation (that these axes are invisible to gating CI, for two independent reasons) is correct and unaffected. The reason it is flagged from the PR as well as the issue is narrower and practical: **the shift these rows take is legal on the transcript**, so hardening the normalizer against the parent's bases would break the transcript axis.

### DIFF
--- a/src/spdi/convert.rs
+++ b/src/spdi/convert.rs
@@ -412,7 +412,16 @@ pub fn hgvs_to_spdi_simple(variant: &HgvsVariant) -> Result<SpdiVariant, Convers
 /// The resulting SPDI uses the **same accession** as the HGVS variant — for
 /// `NM_000088.3:c.1A>G` the SPDI sequence is `NM_000088.3`, not the
 /// underlying genomic accession. This matches NCBI Variation Services'
-/// behavior: SPDI is positional on whichever accession is provided.
+/// behavior: SPDI is positional on whichever accession is provided. No axis is
+/// projected onto another.
+///
+/// The one thing that is *not* carried over verbatim is a **compound
+/// reference's genomic-context wrapper** on a transcript axis: for
+/// `NG_008939.1(NM_000532.5):c.156A>G` the SPDI sequence is `NM_000532.5`. That
+/// is not a projection either — it is the accession this function already
+/// resolved the position against, and naming the wrapper instead pointed every
+/// downstream fetch at the genomic parent. See [`transcript_axis_sequence`]. A
+/// `g.`/`m.`/`o.` description keeps its accession exactly as written.
 ///
 /// # Arguments
 ///
@@ -535,6 +544,46 @@ fn mt_to_spdi_simple(variant: &MtVariant) -> Result<SpdiVariant, ConversionError
     )
 }
 
+/// The accession a **transcript-axis** (`c.`/`n.`/`r.`) SPDI is emitted on: the
+/// inner transcript, with any `NG_`/`NC_`/`LRG_` genomic-context wrapper
+/// stripped.
+///
+/// This is the same string the coordinate resolution already uses — see
+/// [`resolve_cds_to_tx`], which calls `accession.transcript_accession()`, reads
+/// that transcript's `cds_start`, and bounds the result against *that*
+/// transcript's length. So the number `emit_spdi_for_edit` receives is an offset
+/// on the transcript, and this makes the accession beside it say so.
+///
+/// **Why not `Display`.** `Accession`'s `Display` renders the compound form
+/// (`NG_008939.1(NM_000532.5)`), and that string is both the SPDI's `sequence`
+/// field and the key every short-form edit's reference fetch is made against.
+/// `MultiFastaProvider::resolve_name` reaches a record for it through the
+/// version-strip fallback — `"NG_008939.1(NM_000532.5)".split('.').next()` is
+/// `"NG_008939"` — so a **transcript** offset was read out of the **genomic
+/// parent**. Measured on the prepared reference:
+///
+/// ```text
+/// get_sequence("NG_008939.1",              191, 197) = CACACA   (the parent)
+/// get_sequence("NG_008939.1(NM_000532.5)", 191, 197) = CACACA   <-- same record
+/// get_sequence("NM_000532.5",              191, 197) = ACGCCG   <-- the 191's own frame
+/// ```
+///
+/// Splitting the label from the fetch key would not close it: the SPDI's
+/// `sequence` field *is* the fetch key for every downstream consumer
+/// ([`crate::spdi::apply_to_reference`], and `fetch_window` inside
+/// [`crate::spdi::compare_denoted_sequences`]), so a compound label just moves
+/// the wrong fetch one hop out. The compound spelling is also not a valid SPDI
+/// sequence in the first place — it is HGVS syntax, not an accession, and SPDI's
+/// `sequence` must be the accession the position is an offset on.
+///
+/// **Genomic axes deliberately keep `Display`.** On a `g.`/`m.`/`o.` description
+/// the coordinates are the *parent's*, so stripping the wrapper here would name
+/// the transcript for a genomic offset — the same defect with the frames
+/// swapped. Those paths are untouched.
+fn transcript_axis_sequence(accession: &Accession) -> String {
+    accession.transcript_accession()
+}
+
 /// Convert a non-coding transcript (`n.`) variant to SPDI without consulting
 /// a provider. The SPDI is emitted on the transcript accession.
 ///
@@ -546,7 +595,7 @@ fn tx_to_spdi_simple(variant: &TxVariant) -> Result<SpdiVariant, ConversionError
     let start_tx = tx_pos_for_simple_path(&variant.loc_edit.location, "n")?;
     let end_tx = tx_end_for_simple_path(&variant.loc_edit.location, start_tx, "n")?;
     emit_spdi_for_edit(
-        variant.accession.to_string(),
+        transcript_axis_sequence(&variant.accession),
         start_tx,
         end_tx,
         edit,
@@ -566,7 +615,7 @@ fn rna_to_spdi_simple(variant: &RnaVariant) -> Result<SpdiVariant, ConversionErr
     let start_pos = rna_pos_for_simple_path(&variant.loc_edit.location, "r")?;
     let end_pos = rna_end_for_simple_path(&variant.loc_edit.location, start_pos, "r")?;
     emit_spdi_for_edit(
-        variant.accession.to_string(),
+        transcript_axis_sequence(&variant.accession),
         start_pos,
         end_pos,
         edit,
@@ -757,7 +806,7 @@ fn cds_to_spdi_with_provider<P: ReferenceProvider + ?Sized>(
         .unwrap_or(*start_cds);
     let (start_tx, end_tx) = resolve_cds_to_tx(&variant.accession, start_cds, &end_cds, provider)?;
     emit_spdi_for_edit(
-        variant.accession.to_string(),
+        transcript_axis_sequence(&variant.accession),
         start_tx,
         end_tx,
         edit,
@@ -792,7 +841,7 @@ fn tx_to_spdi_with_provider<P: ReferenceProvider + ?Sized>(
         resolve_tx_exonic_bounded(&variant.accession, &variant.loc_edit.location, provider)?
     };
     emit_spdi_for_edit(
-        variant.accession.to_string(),
+        transcript_axis_sequence(&variant.accession),
         start_tx,
         end_tx,
         edit,
@@ -827,7 +876,7 @@ fn rna_to_spdi_with_provider<P: ReferenceProvider + ?Sized>(
         resolve_rna_exonic_bounded(&variant.accession, &variant.loc_edit.location, provider)?
     };
     emit_spdi_for_edit(
-        variant.accession.to_string(),
+        transcript_axis_sequence(&variant.accession),
         start_tx,
         end_tx,
         edit,
--- a/tests/it/compound_accession_sequence_frame.rs
+++ b/tests/it/compound_accession_sequence_frame.rs
@@ -1,0 +1,262 @@
+//! A compound `NG_xxx(NM_yyy)` reference resolves its **coordinates** on the
+//! inner transcript and must resolve its **bases** there too.
+//!
+//! # The two halves of one accession, and which one the numbers belong to
+//!
+//! `Accession` stores a compound reference the way HGVS writes it: the *inner*
+//! accession is the primary one and the `NG_`/`NC_`/`LRG_` parent hangs off it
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

PR #1822, opened 2026-08-13
fix(examples): refuse a FERRO_PARTITION naming no arm from every cargo target

nh13:
A misspelled `FERRO_PARTITION` was served by the shipped rule under the candidate's name, silently, at exit 0. Every bake-off, flip census and arm-comparison figure produced through an example harness rested on the variable having been spelled correctly, with nothing checking.

## The defect, re-derived

`examples/dump_confluence_divergences.rs`, over the 11 272-class cis confluence corpus generated by `generate_cis_confluence_corpus` (default seeds, `g,c` axes), on `439617c2`:

| `FERRO_PARTITION` | exit | stderr | census md5 |
|---|---|---|---|
| unset | 0 | 0 B | `321008a06618c26209ae8d5095b616c0` |
| `live` | 0 | 0 B | `321008a06618c26209ae8d5095b616c0` |
| `canonical` | 0 | 0 B | `9dae11815bde23d0adeaddae024dfb7d` |
| `canonical-coalesced` | 0 | 0 B | `9dae11815bde23d0adeaddae024dfb7d` |
| **`canonicl`** | **0** | **0 B** | **`321008a06618c26209ae8d5095b616c0`** |

Release build. The misspelled arm and a correct run of the shipped arm are the same observation — same bytes, same exit status, nothing on stderr.

A debug build is no better as an instrument. The library's refusal does fire there, but the harness catches panics per row, so the run still exits 0 and reports `classes: 11272  divergent: 0` — which reads as a candidate that converges every class — with the diagnosis only on a 17 103 175-byte stderr that a batch run discards.

## Why this is not fixed by refusing more centrally

`partition_rule_from_env` already refuses centrally, and has since #1636. In the debug run above it refused 47 377 times and the run still exited 0. What is wrong is *where the refusal is delivered*: at a process boundary, which is per-entry-point by construction — and deliberately so, because `canonicalize_from_sequence` is infallible, sits under every public entry point and is reached across the FFI boundary by the Python bindings. Making the library abort in a release build is the thing #1636 removed, and `PARTITION_SWITCH_MAY_ABORT`'s doc records that compiling the switch out was tried first and reverted, because `README.md` documents the A/B recipe as working on a stock release build.

The root cause is narrower than "wiring tests are weak": `partition_switch_wiring.rs` parsed `Cargo.toml`'s `[[bin]]` table and then **appended one example by hand**, so its completeness was completeness over a hand-written list. Deriving the denominator from the manifest's own target tables removes the "targets somebody remembered to add" step entirely — a target that is not in `Cargo.toml` does not build.

## What changed

- All 22 remaining `[[bin]]`/`[[example]]` targets now call `partition_switch_startup_error()` after argument parsing and fail on `Some`, before any work. `--help` still answers under a stale value.
- `every_binary_target_reports_a_refused_partition_switch` becomes `every_binary_and_example_target_reports_a_refused_partition_switch`, reading both manifest tables, with both denominators asserted so `0 of 0` cannot pass as a result.
- The obligation is uniform rather than scoped to the targets that normalize today. A target that does not normalize loses nothing by refusing a value naming no arm, and a uniform rule has no per-target judgement to rot when an example starts normalizing later.

Failing first, then passing:

```
before:  1 failed  (exit 100) — named all 22 unwired targets
after:   5 passed  (exit 0)
```

And on the rebuilt release binary: `canonicl` → exit 1, arm list on stderr, empty stdout, while unset / `live` / `canonical` keep their pre-change md5s. This refuses a measurement that was never real; it moves no measurement that was.

## Second, related instrument defect (documented, not fixed)

The conformance half of the arm flip is not measurable by the suite that appears to measure it. Generating `hgvs_spec_normalization.json` under each of the four arms yields a **byte-identical** file, md5 `d198a442005af419d0921b9214a44c54`. So "the flip breaks no spec test" is true by construction there, and is not evidence.

The cause is mechanical, not corpus geometry: the arms are consulted from `canonicalize_from_sequence`, which re-derives a partition from reference bases, and `generate_spec_fixture` normalizes with an empty `MockProvider`. The geometry is in fact present — 52 of the 934 rows carry a multi-member cis allele and 15 of those have members ≤ 8 nt apart — so this is **not** the member-geometry blindness of #1456 and reading it as that sends you after the wrong property.

`the_spec_fixture_is_blind_to_the_partition_switch` pins both halves with their denominators, and its doc comment records what closing it would take (a reference-backed run of the same corpus, plus a per-arm output path, plus the fact that `spec_expected` does not move with the arm) and the stale-`ensure_spec_fixture` trap that made the first cross-arm run vacuous.

## Gate

`cargo fmt --all --check` clean; `cargo clippy --features dev --all-targets -- -D warnings` clean; `cargo nextest run --features dev --no-fail-fast` → 10 528 passed, 0 failed.

Representation-Change: none. No output string moves — this adds a pre-flight refusal and a test, and the `live`/`canonical` census md5s are byte-identical before and after. Note that turning a silent default into an error is nonetheless a behaviour change for any caller passing a value that names no arm: such a caller previously got `live`'s answer at exit 0 and now gets exit 1. That is the point of the change, and `FERRO_PARTITION` is an unstable evaluation switch outside semantic versioning.


nh13:
@coderabbitai pause

PR #1823, opened 2026-08-13
fix(spdi): refuse a c./n./r. end boundary that names no coordinate

### LINKED ISSUE #1804: hgvs_to_spdi collapses an unresolvable c./n./r. end boundary onto the start — the transcript-axis sibling of #1795

nh13:
Closes #1804.

Representation-Change: previously-accepted input is newly refused on the `hgvs_to_spdi` / `hgvs_to_spdi_simple` path, so this is a real migration for any consumer converting a `c.`/`n.`/`r.` description whose end boundary is a range `(a_b)` or `?`. 0 of 85,642 corpus rows move, and that zero is **structural** — see *Representation stability* below for the three checks that establish it. `normalize` output is unaffected: it does not route through `hgvs_to_spdi`.

## The defect

`hgvs_to_spdi` substituted the interval's **start** whenever its **end** boundary resolved to nothing. `UncertainBoundary::inner()` answers `None` for two shapes — a range boundary `(20_30)` (the `Range` variant) and an unknown `?` (`Mu::Unknown`) — and seven call sites answered that `None` with the start.

Measured against `origin/main` at `439617c2`, using the existing `make_intronic_provider()` fixture (`NM_INTRON.1`, 100 × `A`, CDS 11..90, two exons):

|---|---|---|
| `c.10_(20_30)delAAAA` | `NM_INTRON.1:19:AAAA:` | refused |
| `c.10_?delAAAA` | `NM_INTRON.1:19:AAAA:` | refused |
| `n.10_(20_30)delAAAA` | `NM_INTRON.1:9:AAAA:` | refused |
| `n.10_?delAAAA` | `NM_INTRON.1:9:AAAA:` | refused |
| `r.10_(20_30)delaaaa` | `NM_INTRON.1:9:AAAA:` (simple) / `19:AAAA:` (provider) | refused |
| `r.10_?delaaaa` | `NM_INTRON.1:9:AAAA:` (simple) / `19:AAAA:` (provider) | refused |
| `c.10_13delAAAA` (control) | `NM_INTRON.1:19:AAAA:` | **unchanged** |

So all three axes reproduce, and the first two rows share a triple with the last one — three descriptions that `parse` and `Display` distinctly, one SPDI triple.

**The short forms are worse than the issue records, and that is new here.** `c.10_(20_30)del` — no spelled bases — converted to `NM_INTRON.1:19:A:`: a **one-base** deletion standing in for a deletion whose extent is unknown. The collapse does not merely lose the end, it invents a plausible one.

**One correction to the issue.** It cites `c.(4185+1_4186-1)_(4357+1_4358-1)del` — the exon-deletion shape `src/hgvs/interval.rs:18` names — as "the real-world form of the `c.` row". Measured, that shape **already declined** on `main`, because its *start* is also a range and the start side has always used `ok_or_else`. The live shapes are those with a resolvable start and an unresolvable end.

## The verdict, and why it is not #1795's

`InvalidPosition`, and it is derived from this axis rather than inherited from the genomic sibling.

Every one of the seven sites **already answers the same question for the start side**, with `ok_or_else` and exactly that variant (`"cannot convert c. variant with unknown start position"`). The asymmetry between the two sides *is* the defect the issue names, so the end now gets what the start has always had. That is an existing contract applied consistently, not a new policy — which is why this needed no operator ruling.

The axis' other decline, `MissingReferenceData`, would be a **false statement** here. It means "a well-formed position whose SPDI form needs data or a projection I do not have" (`resolve_cds_to_tx`, `resolve_tx_pos`, `require_simple_tx_pos` all use it that way, and their messages say "requires reference provider with exon data" / "without genomic projection"). It invites the caller to supply a provider. No provider, exon table or genomic projection resolves `?` — the uncertainty is in the **description**, not in ferro's reference data.

Corroborating, from the class sweep below: **`src/vcf/from_hgvs.rs` — the sibling lowering path — already refuses an unresolvable end on all five of its arms** (`"CDS variant with unknown end position cannot be converted to VCF"`, and the `n.`/`r.`/`g.` equivalents). So this restores consistency between the two lowering paths rather than inventing a verdict for one of them.

**Why the mechanism differs from #1795's, which is why that PR correctly did not sweep these axes in.** `reject_unresolvable_genomic_position` walks *both* endpoints and refuses what it finds **inside** a resolvable one — a `+`/`-` offset, a `pter`/`qter`/`cen` landmark — because neither may appear on a genomic position at all. On `c.`/`n.`/`r.` an offset is legitimate: `c.(4185+1_4186-1)_(4357+1_4358-1)del` is well-formed HGVS and is the motivating example on `UncertainBoundary`. So the offset is **not** the defect, and `resolve_transcript_end_boundary` never looks at one. What it refuses is the *absence* of a single coordinate — a property of the boundary, not of its contents.

`an_offset_inside_the_range_is_not_what_is_refused` is the discriminating guard for exactly this: it asserts the refusal message cites the absent coordinate and does **not** mention an offset. A guard written as "walk both endpoints of the range and refuse an offset" passes every other assertion in this PR and fails that one.

## The fix

One declining resolver, `resolve_transcript_end_boundary`, shared by all seven sites:

| site | axis / path |
|---|---|
| `tx_end_for_simple_path` | `n.`, no provider |
| `rna_end_for_simple_path` | `r.`, no provider |
| `cds_to_spdi_with_provider` | `c.` |
| `tx_to_spdi_with_provider` | `n.`, start needs provider |
| `rna_to_spdi_with_provider` | `r.`, start needs provider |
| `resolve_tx_exonic_bounded` | `n.`, plain exonic + provider |
| `resolve_rna_exonic_bounded` | `r.`, plain exonic + provider |

**Seven, not the five the issue tabulates.** The last two are missing from it, and they are the sites that actually fire for the issue's own `n.`/`r.` provider rows: `tx_needs_provider` reads `inner()`, which is `None` for a range or `?` end, so `n.10_(20_30)del` takes the *exonic-bounded* branch rather than the `tx_to_spdi_with_provider` one the issue attributes it to. Six of the seven are exercised by a measured reproducer; the seventh (`tx_to_spdi_with_provider`) is the `n.` twin of the measured-live `r.` one and is covered via an intronic start.

The two simple-path helpers lose their now-meaningless `fallback` parameter.

### Class-exhaustion sweep

Repo-wide, for the same shape (an end that degrades to the start):

- **`src/spdi/convert.rs`, 6 genomic sites** (`get_end_pos(..).unwrap_or(start_pos)` at the `g.`/`m.`/`o.` arms) — **#1795's scope, deliberately untouched.** See the note on #1807 below.
- **`src/vcf/from_hgvs.rs`, all five arms** — already decline. No defect.
- **`src/hgvs/variant.rs:1367`** (`cis_member_range`, protein arm) — an absent end degrades to the start. **Excluded with reason:** it is a *sort key* for cis-member ordering, not a description or a triple; its own comment states the degradation is deliberate ("rather than dropping the member to the sort's max sentinel"), and nothing downstream reads it as a coordinate.
- `src/python_helpers.rs` — returns `Option`, no substitution.

After the change, `grep` finds no remaining `unwrap_or` in the transcript-axis region of `convert.rs`.

## Unchanged on purpose

- **`(13)` — the case that discriminates a correct guard from an over-general one.** It is `Mu::Uncertain`: parenthesised but perfectly numeric, so `inner()` answers `Some` and it resolves exactly like the bare `13` beside it. A guard keyed on "the boundary is parenthesised", or on `Interval::has_complex_boundaries`, would refuse it. Pinned on all three axes.
- Plain ranges, point variants and substitutions on every axis and entry point — 22 conversions, with the count asserted so a loop that stopped running could not read as a green negative control.
- The intronic / downstream / non-positive `MissingReferenceData` declines, where the end **is** resolvable.
- The start side, whose message and verdict are untouched.

**One deliberate diagnostic change, pinned as one.** For `n.10+5_?delAAAA` — intronic start *and* unresolvable end — `main` reported the intronic start; this reports the end. Both are correct refusals, and the ordering is structural rather than a preference: `resolve_tx_to_provider_tx` takes a resolved end **by value**, so there is nothing to hand it until the boundary yields one. Manufacturing a fake end to satisfy that signature is the defect being fixed. `an_intronic_start_with_an_unresolvable_end_reports_the_end` pins both halves, including that a resolvable end leaves the intronic decline untouched.

## Representation stability

`examples/dump_normalized_corpus.rs`, `origin/main` (`439617c2`) → this branch:

```
| total  | 85642 |
| moved  | 0 (0.0%) |
```

and the `--verify-spdi` dumps are **byte-identical** on both sides (`6 denote different bases, 57727 agree, 27909 unverifiable`).

**That zero is STRUCTURAL, on three independent checks — do not read it as evidence of safety:**

1. **0 of 85,642 inputs can carry the shape at all.** The generator spells every position with `format!` from an integer (`p(..)`, `at(..)`, and `-N`/`*N` for the `c.` zones), so no row carries `(` or `?` in its input. Measured, not inferred.
2. **96 outputs do carry `?`** — all `NM_TEST.1:c.?_1insA` / `c.?_1insC`, i.e. `?` on the **start**, which was already declining before this change and is untouched by it.
3. **The corpus has no `n.` or `r.` axis at all** (axes present: `g` 26102, `c` 26102, `cx` 26094, `p` 7344), so two of the three axes this PR fixes are outside its reach regardless.

The real disclosure is therefore the one in the trailer and not the corpus figure: a consumer converting a `c.`/`n.`/`r.` description with an uncertain or unknown end boundary previously received a triple and now receives `InvalidPosition`. That triple was wrong — it described a variant ending where it begins — so the migration is from a silently wrong answer to an honest refusal, on the SPDI conversion path only.

`normalize` is unaffected: `src/normalize/` contains no live call to `hgvs_to_spdi` in the normalization pipeline. The single call, `Normalizer::to_spdi` (`src/normalize/mod.rs:2523`), is an explicit public conversion accessor — and it is covered by the new integration guard.

## Verification

Every exit status below was read back **from the log file**, not from a completion notification — per `CLAUDE.md`'s note that a chained command's reported status is its last command's.

```
cargo fmt --all --check                                   FMT=0
cargo clippy --all-features --all-targets -- -D warnings  CLIPPY_EXIT=0
FERRO_SWEEP_SEEDS=full cargo nextest run --features dev    SUITE2_EXIT=0
    10533 tests run: 10533 passed (18 slow), 152 skipped — 0 FAIL lines
scripts/run_oracle_suite.sh                                ORACLE_EXIT=0
    10396 tests run: 10396 passed (2 slow), 289 skipped — 0 FAIL lines
```

`scripts/run_oracle_suite.sh` rather than a bare armed run, which cannot pass on `main`. No Python surface is touched, so the `python-*` checks are not implicated. `tests/proptest-regressions/` is untouched (no `*_strategy()` was modified).

**Every guard was sabotage-checked.** With the resolver changed to `.or_else(|| interval.start.inner())` — the defect restored exactly — 4 of the 4 assertion guards go red (`SABOTAGE_EXIT=100`, read from the log), and the integration guard goes red separately (`SABOTAGE_EXIT=100`). The negative control correctly stays **green** under sabotage, which is what it should do: restoring the collapse does not affect a resolvable end.

## Review

Self-reviewed against the local diff before pushing, under this repo's `.coderabbit.yaml` (`profile: assertive`) and the `src/{project,convert,coords,spdi,liftover}/**/*.rs` path instruction — whose clause *"Projection must DECLINE rather than emit a coordinate it cannot justify — flag a fallback that silently produces a plausible-but-unverified position"* describes this defect exactly.

**Applied (2):**
1. *Document the new refusal class on both public entry points.* Both `hgvs_to_spdi_simple` and `hgvs_to_spdi` enumerate which inputs they refuse and with which `ConversionError`; both lists omitted this one, so the docs told a caller it converts.
2. *Cover the refusal from an integration test.* The same path instruction requires coverage by `tests/it/spdi_tests.rs` (among others). All the guards were crate-internal unit tests, which call the conversion through the module's own `use` and so would still pass if a wrapper swallowed the error. Added `an_unresolvable_transcript_end_is_refused_through_the_public_api`, which asserts the refusal survives `Normalizer::to_spdi` — the wrapper the PyO3 binding routes through.

The first of those two was itself found wrong on re-read and corrected before commit: it initially claimed the refusal applies "on **any** axis", which is false on this branch — the genomic axes still substitute the start, and that is #1795's scope.

**Rejected (1):** the fallback arms of `resolve_tx_exonic_bounded` / `resolve_rna_exonic_bounded` resolve the end boundary a second time. Harmless — the first resolution already proved it `Some` — and threading the value through would make the helper depend on its caller having checked, which is the coupling that produced the defect.

## Relationship to #1807

**#1807 is open and touches the same file. This branch does not depend on it and does not touch it.**

Branched from `origin/main` (`439617c2`), not from `fix/issue-1795-complex-boundary-offset`. Neither of the helpers introduced by #1807 was needed (`stated_positions`, `resolve_genomic_end_pos`) — the transcript axes need a different resolver for the reasons above, and the only thing borrowed from `main` is the *shape* of the existing `reject_unresolvable_genomic_position`, which is already there.

The two **do** conflict textually, measured rather than assumed:

```
git merge-tree --write-tree origin/fix/issue-1795-complex-boundary-offset HEAD
  -> CONFLICT (content): Merge conflict in src/spdi/convert.rs
```

so whichever merges second needs a rebase. The changes are complementary, not competing: #1807 closes the `g.`/`m.`/`o.` half, this closes the `c.`/`n.`/`r.` half, and after both land the six genomic `unwrap_or(start_pos)` sites and these seven transcript sites are all gone.

### DIFF
--- a/src/spdi/convert.rs
+++ b/src/spdi/convert.rs
@@ -325,7 +325,12 @@ fn reject_unresolvable_genomic_position(
 /// (#1643) is refused outright by both entry points with
 /// [`ConversionError::InvalidPosition`]: neither has anything on a genomic
 /// accession to resolve it against, and dropping either made distinct
-/// descriptions share one triple.
+/// descriptions share one triple. On the **transcript** axes, a `c.`/`n.`/`r.`
+/// interval whose **end** names no single coordinate — a range boundary
+/// `(20_30)` or an unknown `?` — is refused with the same variant (#1804); see
+/// [`resolve_transcript_end_boundary`] for why those axes need their own
+/// resolver rather than the genomic guard. The genomic axes still substitute
+/// the start in that case, which is #1795's scope and not fixed here.
 ///
 /// # Arguments
 ///
@@ -401,7 +406,10 @@ pub fn hgvs_to_spdi_simple(variant: &HgvsVariant) -> Result<SpdiVariant, Convers
 /// [`ConversionError::MissingReferenceData`]. An offset or a `pter`/`qter`/`cen`
 /// special position on a `g.`/`m.`/`o.` position is refused with
 /// [`ConversionError::InvalidPosition`] — see
-/// [`reject_unresolvable_genomic_position`].
+/// [`reject_unresolvable_genomic_position`]. A `c.`/`n.`/`r.` interval whose
+/// **end** names no single coordinate — a range boundary `(20_30)` or an
+/// unknown `?` — is refused with the same variant (#1804); see
+/// [`resolve_transcript_end_boundary`].
 ///
 /// An **unspelled identity** (`g.100=`, `g.100_102=`) therefore needs a
 /// provider: on [`hgvs_to_spdi_simple`] it returns
@@ -593,7 +601,7 @@ fn transcript_axis_sequence(accession: &Accession) -> String {
 fn tx_to_spdi_simple(variant: &TxVariant) -> Result<SpdiVariant, ConversionError> {
     let edit = unwrap_edit(&variant.loc_edit.edit)?;
     let start_tx = tx_pos_for_simple_path(&variant.loc_edit.location, "n")?;
-    let end_tx = tx_end_for_simple_path(&variant.loc_edit.location, start_tx, "n")?;
+    let end_tx = tx_end_for_simple_path(&variant.loc_edit.location, "n")?;
     emit_spdi_for_edit(
         transcript_axis_sequence(&variant.accession),
         start_tx,
@@ -613,7 +621,7 @@ fn tx_to_spdi_simple(variant: &TxVariant) -> Result<SpdiVariant, ConversionError
 fn rna_to_spdi_simple(variant: &RnaVariant) -> Result<SpdiVariant, ConversionError> {
     let edit = unwrap_edit(&variant.loc_edit.edit)?;
     let start_pos = rna_pos_for_simple_path(&variant.loc_edit.location, "r")?;
-    let end_pos = rna_end_for_simple_path(&variant.loc_edit.location, start_pos, "r")?;
+    let end_pos = rna_end_for_simple_path(&variant.loc_edit.location, "r")?;
     emit_spdi_for_edit(
         transcript_axis_sequence(&variant.accession),
         start_pos,
@@ -797,14 +805,8 @@ fn cds_to_spdi_with_provider<P: ReferenceProvider + ?Sized>(
             description: "cannot convert c. variant with unknown start position".to_string(),
         }
     })?;
-    let end_cds = variant
-        .loc_edit
-        .location
-        .end
-        .inner()
-        .copied()
-        .unwrap_or(*start_cds);
-    let (start_tx, end_tx) = resolve_cds_to_tx(&variant.accession, start_cds, &end_cds, provider)?;
+    let end_cds = resolve_transcript_end_boundary(&variant.loc_edit.location, "c")?;
+    let (start_tx, end_tx) = resolve_cds_to_tx(&variant.accession, start_cds, end_cds, provider)?;
     emit_spdi_for_edit(
         transcript_axis_sequence(&variant.accession),
         start_tx,
@@ -829,14 +831,8 @@ fn tx_to_spdi_with_provider<P: ReferenceProvider + ?Sized>(
                 description: "cannot convert n. variant with unknown start position".to_string(),
             }
         })?;
-        let end = variant
-            .loc_edit
-            .location
-            .end
-            .inner()
-            .copied()
-            .unwrap_or(*start);
-        resolve_tx_to_provider_tx(&variant.accession, start, &end, provider)?
+        let end = resolve_transcript_end_boundary(&variant.loc_edit.location, "n")?;
+        resolve_tx_to_provider_tx(&variant.accession, start, end, provider)?
     } else {
         resolve_tx_exonic_bounded(&variant.accession, &variant.loc_edit.location, provider)?
     };
@@ -864,14 +860,8 @@ fn rna_to_spdi_with_provider<P: ReferenceProvider + ?Sized>(
                 description: "cannot convert r. variant with unknown start position".to_string(),
             }
         })?;
-        let end = variant
-            .loc_edit
-            .location
-            .end
-            .inner()
-            .copied()
-            .unwrap_or(*start);
-        resolve_rna_to_provider_tx(&variant.accession, start, &end, provider)?
+        let end = resolve_transcript_end_boundary(&variant.loc_edit.location, "r")?;
+        resolve_rna_to_provider_tx(&variant.accession, start, end, provider)?
     } else {
         resolve_rna_exonic_bounded(&variant.accession, &variant.loc_edit.location, provider)?
     };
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

Issue #1824, opened 2026-08-13
Decide whether `IntronPosition::is_near_splice_site`'s 10 is a deliberate third band or a stale literal


nh13:
Split out of #1766. #1766 bundled four asks; three are addressed by the PR that closes it, and **this is the one that could not be done without a decision** — it is also the one #1766's title names.

## The state as shipped

`IntronPosition::is_near_splice_site` (`src/reference/transcript.rs`) uses `offset.abs() <= 10`.

That `10` appears on neither of ferro's splice ladders — shape A is donor 2/6/20/50 and acceptor 2/12/20/50, shape B is 2/8 — so unlike the three predicates beside it, it cannot be re-expressed as a set of `SpliceSiteType` bins. It contradicts `SpliceSiteType::NearSplice` on the same struct in **both** directions:

| offset | `is_near_splice_site()` | `splice_site_type()` |
|---|---|---|
| 5 | `true` | `DonorExtended` — not `NearSplice` |
| 30 | `false` | `NearSplice` |

Measured over `-60..=60` excluding zero: the two disagree at **80 of 120** offsets. They agree only where both answer `false` — `11..=20`, and `>50`. The same holds on the acceptor side (`-5` → `true`/`AcceptorExtended`, `-30` → `false`/`NearSplice`), so it is not a donor artefact.

This is now pinned **as undecided** by two guards in `tests/it/splice_ladder_shapes.rs`, so neither side can move without the ruling being made deliberately rather than in passing:

- `is_near_splice_site_contradicts_the_near_splice_bin` asserts the disagreement with that one bin, and the census at exactly 80 of 120.
- `only_is_near_splice_site_splits_a_ladder_bin` asserts the sharper fact: 10 falls in the **interior** of `DonorRegion` (7..=20) and of `AcceptorExtended` (3..=12), so no union of `SpliceSiteType` bins can express this predicate at all. That set going empty is what "someone derived it" looks like from the outside.

Both were sabotage-checked: replacing the body with `matches!(self.splice_site_type(), SpliceSiteType::NearSplice)` turns both red and nothing else.

## The decision

1. **The `10` is a third, deliberate band** — a "near" distance intended to differ from `NearSplice`'s 21–50 bin. Then the fix is to rename the predicate and say what the band is for, and the answers do not move.
[...]

PR #1825, opened 2026-08-13
refactor(reference): derive the intronic band predicates from the one splice ladder

### LINKED ISSUE #1766: Splice-distance thresholds survive #1742's unification in four public predicates, and `IntronPosition::is_near_splice_site` contradicts `SpliceSiteType::NearSplice`
Found while reviewing #1742, which unifies ferro's two splice-distance ladders. The unification is
real for the three sites it names, but it is not exhaustive, and one of the survivors is not merely
a duplicate — it disagrees with the ladder it sits next to.

## The claim that is not yet true

`tests/it/splice_ladder_shapes.rs`'s module doc states:

> Shape A previously existed as **three** full copies (`reference/transcript.rs`, and twice in
> `convert/noncoding.rs`) plus a fourth partial one (`IntronicRegion::from_offset`, 2/20/50 —
> missing the 6- and 12-rungs) [...] Each shape now has one definition and the rest derive from it.

Four further public entry points still carry their own thresholds and derive from nothing:

| site | thresholds |
|---|---|
| `IntronPosition::is_canonical_splice_site` (`src/reference/transcript.rs:926`) | `abs <= 2` |
| `IntronPosition::is_near_splice_site` (`src/reference/transcript.rs:931`) | `abs <= 10` |
| `IntronPosition::is_extended_splice_region` (`src/reference/transcript.rs:936`) | `abs <= 20` |
| `IntronPosition::is_deep_intronic` (`src/reference/transcript.rs:921`) | `abs > 50` |

They sit **immediately above** `IntronPosition::splice_site_type`, the method #1742 de-duplicated,
in the same `impl` block. Two more of the same shape live on a different type:
`IvsPos::is_deep_intronic` (`abs > 50`) and `IvsPos::is_canonical_splice_site` (`abs <= 2`) at
`src/hgvs/location.rs:780` and `:785`.

## The part that is a defect rather than a duplicate

`is_near_splice_site` uses a rung of **10**, which appears in neither ladder — shape A is donor
2/6/20/50 and acceptor 2/12/20/50, shape B is 2/8 — and it contradicts `SpliceSiteType::NearSplice`
on the same struct in **both** directions:

| offset | `is_near_splice_site()` | `splice_site_type()` |
|---|---|---|
| 5 | `true` | `DonorExtended` |
| 30 | `false` | `NearSplice` |

So `IntronPosition` answers "is this near a splice site?" one way through a predicate and the
opposite way through its bin, and neither answer is marked as the odd one out. `is_extended_splice_region`
has the milder version of the same problem: it returns `true` for `abs <= 2`, which
`IntronicRegion::from_offset` classifies as `CanonicalSpliceSite`, not `ExtendedSpliceRegion`.

## Why this is not being fixed inside #1742

Reconciling `is_near_splice_site` moves the answers of a public predicate, so it needs a decision
rather than a refactor:

1. **Is 10 a third, deliberate threshold** — a "near" band that is meant to differ from
   `NearSplice`'s 21–50 — in which case the fix is to rename it and say so, not to re-derive it?
2. **Or is 10 a stale literal**, in which case the predicate should become
   `matches!(self.splice_site_type(), SpliceSiteType::NearSplice)` and its answers change?

The same question, in weaker form, applies to `is_extended_splice_region`'s inclusion of the
canonical band.

`IvsPos`'s two predicates are the easy half — 2 and 50 agree with shape A — but they are still a
second copy, and `IvsPos` carries no `splice_site_type` to derive from today.

## Suggested scope

- Decide (1) vs (2) for `is_near_splice_site`; whichever way, record it where the ladder is defined.
- Derive the remaining `IntronPosition` predicates from `SpliceSiteType` once that is settled.
- Either derive `IvsPos`'s pair from the same ladder or state why that type is deliberately independent.
- Correct `tests/it/splice_ladder_shapes.rs`'s "each shape now has one definition" claim so it names
  what is left rather than reading as exhaustive. (#1742 can do this last part on its own; the rest
  is this issue.)

No corpus movement is claimed here and none was measured — these predicates have no call sites in
`src/` outside their own unit tests (`grep -rn 'is_near_splice_site\|is_extended_splice_region' src/`
returns only the definitions and `transcript.rs`'s own tests), so the blast radius is the public API
surface rather than any normalized output.

nh13:
> **Stacked on #1742 — merge that one first.** This PR's base is
> `fix/coordinate-basis-docs-and-guards`, not `main`, so the diff shown here is only this
> change: one commit, three files. #1742 restructures the same splice ladders, so branching
> from `main` would have guaranteed a conflict rather than avoided one.
>
> Branched from #1742's head `de6d8896`; #1742 was then rebased to `72d21a95` mid-flight, so
> this was rebased onto its new head. Nothing was lost in that move — each of #1742's four
> commits was confirmed present on its own branch by subject, and this branch is exactly one
> commit ahead of the base.

Closes #1766.

## The measured contradiction

`IntronPosition::is_near_splice_site` answers "is this near a splice site?" with `offset.abs() <= 10`, while `IntronPosition::splice_site_type()` — on the same struct, in the same `impl` block — answers it with `SpliceSiteType::NearSplice`. They contradict each other in **both** directions:

| offset | `is_near_splice_site()` | `splice_site_type()` |
|---|---|---|
| 5 | `true` | `DonorExtended` — not `NearSplice` |
| 30 | `false` | `NearSplice` |
| -5 | `true` | `AcceptorExtended` |
| -30 | `false` | `NearSplice` |

Measured over `-60..=60` excluding zero: **80 of 120** offsets disagree. They agree only where both answer `false` — `11..=20`, and `>50`. The arithmetic is checkable from the two definitions, which is what makes the census worth pinning rather than eyeballing.

## What changes

**`IntronPosition`'s three cumulative *band* predicates stop carrying their own thresholds.** They are now spelled as the set of `SpliceSiteType` bins each band covers, so the rungs live once, on `SpliceSiteType::from_distance_on_side`:

| predicate | was | is |
|---|---|---|
| `is_canonical_splice_site` | `abs <= 2` | the two `*Canonical` bins |
| `is_extended_splice_region` | `abs <= 20` | every bin inside the 20-rung, as an exhaustive `match` |
| `is_deep_intronic` | `abs > 50` | the `DeepIntronic` bin |

**The rebase onto #1742's new head strengthened the argument, and corrects a claim it added.** That revision changed all four predicates from `abs()` to `unsigned_abs()`, because `c.<base>-?` parses to `offset == Some(i64::MIN)` and `i64::MIN.abs()` panics in debug and wraps to a *canonical splice site* in release. It also added a comment saying the four "spell their thresholds inline rather than deriving from `SpliceSiteType::from_distance_on_side`, because three of the four do not correspond to a single rung of it". A band is indeed not a rung — but it **is a union of bins**, which is the step that makes three of them derivable, so that comment is replaced rather than kept.

The practical consequence is the part worth having: `from_distance_on_side` already takes its magnitude with `unsigned_abs`, so deriving **inherits** the sentinel fix from the single place it is written instead of depending on four copies of `unsigned_abs` all staying correct. #1742's own new guard, `the_unknown_offset_sentinels_do_not_panic_or_read_as_canonical`, passes unchanged against the derived predicates — the sentinel lands in `DeepIntronic`, so `is_deep_intronic()` is `true` and the other three are `false`, exactly as its literals gave. `is_near_splice_site` is not derived, so it keeps both its literal and its own `unsigned_abs`.

The distinction that makes this work is that these are **bands**, not bins: `is_extended_splice_region` is `true` at `abs <= 2` because the canonical site *is* within 20 bases of the exon. #1766 reads that as a milder instance of the same defect; it is not one, and the code now says so. `is_extended_splice_region` is deliberately an exhaustive `match` rather than a `matches!`, so a new `SpliceSiteType` bin cannot silently default to one side of the band — it has to fail to compile until someone decides.

**`IvsPos`'s two literals stay, with a stated reason and a cross-test.** #1766 asks to derive them *or* say why the type is deliberately independent. This takes the second option: `hgvs::location` is a leaf module of parsed-notation types with no production dependency on the `reference` layer — `use crate::reference::` appears under `src/hgvs/` exactly once, inside a test — and adding an inward edge from the notation layer onto the reference layer to spell two one-line predicates would invert that. The drift risk deriving would remove is covered instead by `the_ivs_pos_predicates_agree_with_the_one_shape_a_ladder`, which fails if either literal and the ladder's rung stop agreeing at any offset in `-300..=300`. Same guarantee, no new edge.

**`tests/it/splice_ladder_shapes.rs`'s module doc is corrected.** It claimed "Each shape now has one definition and the rest derive from it", which read as exhaustive and was not. It now says each shape has one *classifier*, tabulates the three literals that survive, and says for each whether it is deliberate or unsettled.

## What is deliberately NOT fixed, and why

**`is_near_splice_site` keeps its `10`.** Deriving it would move a public predicate's answers at 80 of 120 offsets, and no authority this project adjudicates against picks a rung:

- **The HGVS spec states no splice distance at all.** `background/glossary.md:250-256` defines "splice donor site", "splice acceptor site" and "splice site" qualitatively — "the 3' splice site, at the end of the intron/start of the exon" — and nothing under `docs/recommendations/` gives a nucleotide count. Both ladders are ferro's house policy end to end.
- **The ruling ledger holds no splice record** — zero occurrences of "splice" in `hgvs_spec_normalization_overrides.json`.
- **There is no recorded intent to recover.** `git log -S 'is_near_splice_site' -- src/reference/transcript.rs` returns one commit: `5f16fa48 Initial commit`.

So picking here would be a ruling made in passing, which is the thing this repo has the most scar tissue about. Instead the contradiction is **pinned as undecided**, at the definition and in a test, with the census asserted at 80/120 — so neither side can move without the guard going red and the ruling being made deliberately. That is the `undecided`-is-first-class discipline `CLAUDE.md` asks for, applied to a house-policy question rather than a spec one.

**The residue is filed as #1824**, with both candidate answers, the measurements above, and the blast radius of each. #1766 bundled four asks and this PR does three of them; closing it without #1824 would have buried the half its own title names, so please carry #1766's `high` label across if you agree with the split.

## Blast radius

`is_near_splice_site` and `is_extended_splice_region` have **zero** production call sites — `grep -rn` returns only their definitions and their tests. `is_canonical_splice_site` likewise. `is_deep_intronic` has **one**: `VcfAnnotation::is_deep_intronic` (`src/vcf/to_hgvs.rs:68`), and the derivation is answer-preserving, so it is unaffected. None of the four has a PyO3 binding. (Small correction to #1766, which says "these predicates have no call sites in `src/` outside their own unit tests" — true of the two it grepped for, not of all four.)

The three re-expressed predicates were measured against the literals they replace at **all 1206** offset × boundary combinations — `-300..=300` plus both unknown-offset sentinels (`i64::MIN`, `i64::MAX`), × `{FivePrime, ThreePrime}` — and **0 differ**.

That is not a structural zero, on either axis it varies. The sweep varies the offset and the `boundary` field **independently**, which is exactly where a band reading the offset's magnitude and a ladder reading `boundary` could diverge; it comes out zero because the 2-, 20- and 50-rungs are symmetric across the two sides, and the same sweep goes red under sabotage (below). And it includes the two sentinels rather than stopping at ±300, so it covers the only offsets where `abs()` and `unsigned_abs()` disagree — the values that motivated #1742's own hardening.

## Sabotage checks — including one that caught a vacuous guard of mine

Each new guard was verified non-vacuous by breaking the thing it claims to catch. All three runs exited **100**, read back from their own logs rather than from a completion notification:

| sabotage | which guards went red |
|---|---|
| ladder's donor canonical rung `2` → `3` | `the_intron_position_band_predicates_derive_from_the_one_shape_a_ladder` **and** `the_ivs_pos_predicates_agree_with_the_one_shape_a_ladder` |
| `is_near_splice_site` → `matches!(splice_site_type(), NearSplice)` | `is_near_splice_site_contradicts_the_near_splice_bin`, alone |
| `IvsPos::is_deep_intronic` rung `50` → `51` | `the_ivs_pos_predicates_agree_with_the_one_shape_a_ladder`, alone |

**The reason this was worth running:** a fourth guard I had written, `each_band_is_exactly_a_union_of_shape_a_bins`, survived **all three**. It asserted each derived predicate equalled the `matches!` its implementation is now written as — a definition compared against itself, and unfalsifiable by construction. It is replaced by `only_is_near_splice_site_splits_a_ladder_bin`, which states the same idea as a property that *can* fail: which `SpliceSiteType` bins each predicate cuts. The three derived bands cut none; `is_near_splice_site` cuts exactly `DonorRegion` (7..=20) and `AcceptorExtended` (3..=12), because 10 falls in both interiors. That set going empty is what "someone derived it" looks like, and it is now the thing that goes red.

## Review

Self-reviewed against `.coderabbit.yaml` (profile `assertive`) with its `src/hgvs/**`, `src/reference/**` and `tests/**` path instructions, plus `.coderabbitai-review/checklist.md`.

**Applied — 2:**

1. `bins_split_by` keyed bins on their `Debug` spelling. A renamed `SpliceSiteType` variant would have quietly changed what the guard compares, and an added one could slip past a `Debug`-keyed map. Replaced with an exhaustive `bin_name` match, so the ladder cannot gain or rename a bin without this file failing to build — the same argument as `is_extended_splice_region`'s exhaustive match.
2. `is_near_splice_site`'s doc named only one of the two guards that now pin it, leaving the second undiscoverable from the definition. Both are named.

**Rejected — 3, with reasons:**

1. *`tests/**` instruction: "flag `issue_*` regression tests that pin the buggy output instead of the fixed one".* `is_near_splice_site_contradicts_the_near_splice_bin` does pin the contradictory behaviour, deliberately. The instruction targets a test that pins a bug **as though it were correct**; this one is named for the contradiction, documented as undecided, and its failure message tells the reader to record the #1766 ruling rather than re-bless the number. Deleting it would let either side move silently, which is the outcome the issue was filed about.
2. *`src/reference/**` instruction's flags* — build-blind placement resolution, unvalidated mmap offsets, cache keying, a provider returning an empty sequence, `legacy_selector` degradation. None is reachable: the diff touches intronic **classification**, not the provider or placement layer. Cleared by reading, not assumed.
3. *`src/hgvs/**` instruction: "public API changes that are semver-breaking without a CHANGELOG entry".* `location.rs`'s change is documentation only — no signature, field or variant moves — and the three re-expressed predicates keep their signatures and (measured) their answers.

Note for the `tests/**` instruction's last clause: none of these guards resolves a prepared reference, so none depends on `FERRO_MANIFEST` and all of them run on PR CI rather than only in the nightly.

## Also checked

A sweep for further sites of the same class across `src/` found none: after this change the only splice-distance literals left are the two ladder definitions (`SpliceSiteType::from_distance_on_side`, `SpliceRegionBin::from_offset`), the unsettled `10`, and `IvsPos`'s documented pair — exactly the table now in the test module's doc. The web service's CDS effect handler already routes through `SpliceRegionBin`.

Representation-Change: none. These are classification predicates over intronic offsets, not normalization; no normalized output is touched, and the three re-expressed predicates were measured identical to the literals they replace at all 1206 offset/boundary combinations, including both unknown-offset sentinels. `src/hgvs/location.rs`'s change is documentation only.

### DIFF
--- a/src/hgvs/location.rs
+++ b/src/hgvs/location.rs
@@ -848,6 +848,43 @@ impl IvsPos {
         }
     }
 
+    // --- Splice-distance convenience predicates ---------------------------
+    //
+    // These two restate rungs that ferro defines once, on
+    // `reference::transcript::SpliceSiteType::from_distance_on_side` (donor
+    // 2/6/20/50, acceptor 2/12/20/50). They are NOT derived from it, and that
+    // is deliberate rather than an oversight (#1766):
+    //
+    // `hgvs::location` is a leaf module of parsed-notation types with no
+    // production dependency on the `reference` layer — `use crate::reference::`
+    // appears under `src/hgvs/` exactly once, inside a test — and `IvsPos` is a
+    // notation this crate *parses*, not something a reference resolves. Adding
+    // an inward edge from the notation layer onto the reference layer to spell
+    // two one-line predicates would invert that for no gain in behaviour.
+    //
+    // The *drift* risk that deriving would remove is covered instead by a
+    // cross-test: `the_ivs_pos_predicates_agree_with_the_one_shape_a_ladder` in
+    // `tests/it/splice_ladder_shapes.rs` fails if either literal below and the
+    // ladder's corresponding rung ever stop agreeing over `-300..=300`.
+    //
+    // **Drift is not the whole of it, and this note used to claim it was.**
+    // Deriving buys a second thing these two do not get: the ladder takes its
+    // magnitude with `unsigned_abs`, while both predicates below still use
+    // `abs()`. That is the difference `IntronPosition`'s sibling note calls the
+    // practical point of deriving, and it is real here — `-?` parses to
+    // `offset == i64::MIN` (`parser::position::OFFSET_UNKNOWN_NEGATIVE`), and
+    // `i64::MIN.abs()` panics under `debug_assertions` and wraps back to a
+    // negative in release, where it reads as a *canonical splice site*. The
+    // cross-test's `-300..=300` sweep cannot see that, by construction: the
+    // sentinels are the only offsets at which the two spellings differ, and
+    // they are outside its range.
+    //
+    // So the guarantee here is narrower than deriving's, in exactly one place.
+    // That place is **#1826 item 1**, which tracks it; #1742 applied the same
+    // two-word fix to `IntronPosition` and deliberately stopped short of this
+    // file. Do not widen the cross-test's range without fixing these two first
+    // — it will panic rather than fail, which reads as a broken test.
+
     /// Check if this is a deep intronic position (>50bp from exon)
     ///
     /// `unsigned_abs` rather than `abs`: [`IvsNotation::to_ivs`] maps a `CdsPos` /
@@ -867,13 +904,21 @@ impl IvsPos {
     /// `false`. Making these decline needs its own decision about what the four
     /// `IvsPos`/`IntronPosition` predicates return.
     ///
+    /// The 50 mirrors `SpliceSiteType::DeepIntronic`; see the note above for
+    /// why this restates the rung rather than deriving it, and which test keeps
+    /// the two from drifting apart.
+    ///
     /// [`IntronicRegion::from_offset`]: crate::convert::noncoding::IntronicRegion::from_offset
     /// [`is_unknown_offset`]: crate::hgvs::parser::position::is_unknown_offset
     pub fn is_deep_intronic(&self) -> bool {
         self.offset.unsigned_abs() > 50
     }
 
-    /// Check if this is at a canonical splice site
+    /// Check if this is at a canonical splice site (within 2bp of exon)
+    ///
+    /// The 2 mirrors `SpliceSiteType::DonorCanonical`/`AcceptorCanonical`,
+    /// which share that rung, so reading the side off the sign of `offset`
+    /// cannot change the verdict. See the note above.
     ///
     /// See [`IvsPos::is_deep_intronic`] for why this is `unsigned_abs`.
     pub fn is_canonical_splice_site(&self) -> bool {
--- a/src/reference/transcript.rs
+++ b/src/reference/transcript.rs
@@ -959,31 +959,114 @@ pub struct IntronPosition {
 // `splice_site_type` is not reachable from a string entry point today; it is
 // reachable by a library caller, because every field of this struct is `pub`.
 impl IntronPosition {
-    // These four predicates spell their thresholds inline rather than deriving
-    // from [`SpliceSiteType::from_distance_on_side`], because three of the four
-    // do not correspond to a single rung of it — `is_near_splice_site`'s 10 is
-    // on neither ladder at all. They take the magnitude with `unsigned_abs` for
-    // the reason recorded on `from_distance_on_side`: `abs()` panics on
-    // `i64::MIN`, the `-?` unknown-offset sentinel.
+    // The predicates below are cumulative *bands* ("within n bases of the
+    // exon"), while `SpliceSiteType` is a *partition* into bins. The two are
+    // different shapes of the same ladder, and a band is not a rung — but it is
+    // a *union of bins*, so it can still be spelled over the ladder rather than
+    // as its own threshold, and the rungs live once, on
+    // `SpliceSiteType::from_distance_on_side`.
+    //
+    // That correction matters for more than tidiness. These predicates take the
+    // magnitude with `unsigned_abs` because `abs()` panics on `i64::MIN`, the
+    // `-?` unknown-offset sentinel, and wraps to a *canonical splice site* in
+    // release — see the note on `from_distance_on_side`. Deriving inherits that
+    // totality from the one place it is written, instead of depending on four
+    // copies of `unsigned_abs` all staying correct.
+    //
+    // `is_near_splice_site` is the exception: its 10 is on neither ladder, so
+    // there is no union of bins to spell it as, and it keeps both its own
+    // literal and its own `unsigned_abs`. See its documentation and #1766.
 
     /// Check if this is a deep intronic position (>50bp from nearest exon)
+    ///
+    /// Derived from [`IntronPosition::splice_site_type`] rather than restating
+    /// the 50-rung. Answer-preserving against the former
+    /// `offset.unsigned_abs() > 50`: that rung is the same on the donor and
+    /// acceptor sides, so reading the side from `boundary` cannot change the
+    /// verdict. Verified exhaustively by
+    /// `the_intron_position_band_predicates_derive_from_the_one_shape_a_ladder`
+    /// over every offset in `-300..=300` **plus both unknown-offset sentinels**,
+    /// against both `boundary` values. The sentinels are the case that matters:
+    /// the ladder's own `unsigned_abs` lands them in
+    /// [`SpliceSiteType::DeepIntronic`], so this reports `true` for them exactly
+    /// as the literal did — which is also what
+    /// `the_unknown_offset_sentinels_do_not_panic_or_read_as_canonical` pins.
     pub fn is_deep_intronic(&self) -> bool {
-        self.offset.unsigned_abs() > 50
+        matches!(self.splice_site_type(), SpliceSiteType::DeepIntronic)
     }
 
     /// Check if this is at a canonical splice site (within 2bp of exon)
+    ///
+    /// Derived from [`IntronPosition::splice_site_type`]; see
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
**Merge-order note: this PR is stacked, not based on `main`.**

Its base is `fix/coordinate-basis-docs-and-guards` — the head branch of **#1742** — so **#1742 must merge first**. Merging or queueing this one against `main` before that will pull in #1742's commits as though they were this PR's.

Per `CLAUDE.md`: a stacked PR whose base has merged reads as `CONFLICTING` and is not a conflict. If #1742 lands and this goes red, the fix is `git rebase --onto origin/main <merged-base>` — replaying only this PR's own commits — rather than `git rebase origin/main`, which replays everything already landed and dies on add/add conflicts in the first commit. `git log --oneline origin/main..HEAD` will show you which commits are actually yours before you start.

Issue #1503, opened 2026-08-06
GenomicPlacement discards lrg_end, so a parent's extent is re-derived from the chromosome frame


nh13:
`GenomicPlacement` models a parent↔chromosome placement as a pure affine and carries no `parent_end`. The LRG parser reads `lrg_end` and throws it away:

```rust
let parent_start: u64 = xml_attr(span, "lrg_start")?.parse().ok()?;
let o1: u64 = xml_attr(span, "other_start")?.parse().ok()?;
let o2: u64 = xml_attr(span, "other_end")?.parse().ok()?;
```

`lrg_end` is never read. Consumers therefore reconstruct the parent's end from the **chromosome** span width — `parent_to_nc` computes `parent_end = parent_start + (nc_end - nc_start)`, and #1502's `get_sequence_length` uses the same expression. That is only correct when the alignment preserves length.

## Where it is measurably wrong

`LRG_1075`, GRCh38 span:

```xml
<mapping_span lrg_start="1049" lrg_end="15267" other_start="190173122" other_end="190187909" strand="1">
```

| quantity | value |
|---|---|
| LRG span width (`1049..15267`) | 14,219 |
| NC span width (`190173122..190187909`) | 14,788 |
| `LRG_1075g` record length | **15,267** (= `lrg_end`) |
| `parent_start + (nc_end - nc_start)` | **15,836** |

569 bases of divergence, because the LRG↔chromosome alignment carries indels the affine cannot express — the same limitation #1490 documents ("it simply cannot express the `mismatch` / `lrg_ins` / `other_ins` diffs the LRG XML lists alongside the mapping").

## Why it is latent today

Two things keep it from biting:

- `get_sequence_length`'s placement branch only runs when `own_record(id).is_none()`, and `LRG_1075` **has** its own record, so after #1490 it is served from that.
- Every #728 derived `NG_` is constructed with `parent_start == 1` and an exact affine, so the two widths agree.

The exposed population is therefore a parent that is record-*less* **and** indel-bearing. I have not established that one exists in the prepared reference; the point is that nothing in the model prevents it, and the failure would be a silently over-long length rather than an error.

## Suggested shape

Carry the parent's end on the placement instead of re-deriving it:

- add `parent_end` to `GenomicPlacement`, populated from `lrg_end` for LRG and from the alignment's parent end for RefSeqGene;
- use it in `get_sequence_length` and in `parent_to_nc`'s bounds check;
[...]

PR #1830, opened 2026-08-13
fix(reference): narrow an LRG placement to its affine-exact prefix

### LINKED ISSUE #1499: reference: project_to_genomic mis-places LRG parent coordinates via the same single-affine placement (sibling of #1462)
Sibling of #1462, on the *projection* half of the same machinery. #1462/#1490 fixed base **serving** for LRG parents by preferring the frozen record over the placement. Coordinate **projection** still goes through the placement and is still wrong for the same accessions.

## Mechanism

`VariantProjector::project_to_genomic` composes the cdot `NM_`→`NC_` step with the `NC_`→`NG_`/`LRG_` affine returned by `ReferenceProvider::genomic_placement`. For an LRG, that affine is parsed by `parse_lrg_main_assembly_placement` from the record's `main_assembly` mapping, as a single `GenomicPlacement` (`parent_start`, `nc_start`, `nc_end`, `strand`).

An LRG↔chromosome alignment is not affine. The LRG XML says so itself — alongside the mapping it lists `<diff>` elements of type `mismatch`, `lrg_ins` and `other_ins`. A single affine cannot express the indels, so a projected coordinate drifts by the net indel count between the LRG origin and the projected position.

`LRG_292` is the worked example: its main-assembly span is `1..193689 → NC_000017.11:43024295..43217981` (strand `-1`), but `43217981 - 43024295 + 1 = 193687`, two short of the record length — 3 `lrg_ins` bases and 1 `other_ins` base, netting `+2`.

## Measured scope

Censused while fixing #1462, over all 1294 LRG records in the prepared reference. Comparing each `LRG_<N>`'s placement-synthesized bases against its own `LRG_<N>g` record over a 20-base window gives the per-accession offset directly, and that offset **is** the projection error:

```
offset histogram: {-8: 1, -5: 1, -1: 12, 0: 1269, +1: 3, +2: 2, +5: 2, +6: 2, +23: 1}
```

So **25 of 1294** accessions mis-project by a non-zero constant within the probed region, ranging `-8` to `+23`. Over whole records the divergent set is larger — **164 of 1294** differ somewhere — because a `mismatch` diff changes a base without shifting anything, and shifts accumulate only past an indel. For projection specifically, the indel-bearing accessions are the ones that matter.

`LRG_292` is verified end to end: pre-#1490, `LRG_292:g.160875` resolved to the base at record position 160877.

## Why it is silent

The result is well-formed and in-bounds — a valid coordinate on a real accession, just the wrong one. `FERRO_ASSERT_IN_BOUNDS` cannot see it, `FERRO_ASSERT_REPARSE` cannot see it, and `FERRO_ASSERT_IDEMPOTENT` cannot see it because the wrong answer is a fixed point. No oracle catches it.

## Note on the shape of a fix

Probably not repairable within `GenomicPlacement` as it stands, since the type is affine by construction. Note the parser **already declines** a gapped mapping — `parse_lrg_main_assembly_placement` returns `None` when the `main_assembly` mapping has more than one `<mapping_span>` (`no_placement_when_main_assembly_mapping_is_gapped`) — so the machinery deliberately refuses to model gaps rather than approximating them. The single-span LRGs at issue here are gapped in a different way: one span plus a `<diff>` list the parser ignores.

Two candidate directions:

1. **Model the diffs.** Parse the `<diff>` list and carry it alongside the affine, applying the cumulative indel adjustment when mapping a coordinate. Exact, and the data is already on disk.
2. **Decline rather than mis-place.** Return `None` for an LRG whose `<diff>` list contains any `lrg_ins`/`other_ins`, matching the existing precedent for gapped mappings. Loses the 25 accessions' projection entirely but never emits a wrong coordinate.

(1) is the better answer if the diff list is reliable; (2) is the conservative one and is consistent with how the same file already treats a case it cannot represent.

## Representation stability

Whichever direction is taken, this **moves projected genomic output** for the affected accessions — they are currently accepted, so every move is a migration and must be measured against the previous release, not just `main`. Option (2) additionally moves rows accepted → rejected. Per `CLAUDE.md`, quantify before merging; the corpus for it is the 1294 LRG records, and the histogram above is the expected blast radius.

Refs #1462, #1490.

nh13:
Closes #1499.

## The defect

`VariantProjector::project_to_genomic` composes cdot's `NM_`→`NC_` step with the `NC_`→`LRG_` affine that `parse_lrg_main_assembly_placement` reads out of the LRG XML. That parser already declines a **multi-span** mapping, on the stated ground that "a single affine placement cannot represent a gapped mapping". It then accepts a single-span mapping unconditionally — and a single span is not an affine either, because alongside it the record lists `<diff>` elements. The vocabulary is closed at `mismatch`, `lrg_ins` (LRG bases with no chromosome counterpart) and `other_ins` (the reverse); verified over all 1,294 LRG records in the prepared reference, which between them carry 479 `mismatch`, 63 `lrg_ins` and 47 `other_ins` diffs in their main-assembly mappings.

The two insertion kinds shift every coordinate past them, in opposite directions, so a projected coordinate drifts by the net indel count between the LRG origin and the projected position.

### Reproduced, on the prepared reference

`LRG_292` is the issue's worked case and it reproduces exactly. Its span is `1..193689` onto `NC_000017.11:43024295..43217981`, but the chromosome side is `43217981 - 43024295 + 1 = 193687` — two short of the record, from 3 `lrg_ins` bases against 1 `other_ins` base.

Measured by fetching an 11-base window of the frozen `LRG_292g` record at each position and searching the chromosome around `parent_to_nc`'s answer for the window that reverse-complement-matches it:

| `LRG_292:g.` | chromosome agrees at offset |
|---|---|
| 1000, 5000, 13000 | **0** (exact — below the first indel) |
| 20000, 50000, 100000, 160875, 193000 | **+2** |

End to end, through `ferro project --axis g`:

```
LRG_292(NM_007294.4):c.1del   ->  LRG_292:g.93888del     (record's own frame: g.93890del)
LRG_292(NM_007294.4):c.100del ->  LRG_292:g.102223del    (record's own frame: g.102225del)
```

The result is well-formed and in-bounds, so `FERRO_ASSERT_IN_BOUNDS`, `FERRO_ASSERT_REPARSE` and `FERRO_ASSERT_IDEMPOTENT` all pass on it — the wrong answer is a fixed point.

### Census, re-derived

Re-derived from the LRG XMLs rather than quoted from the issue. Over all **1,294** records, every one of which has a **single-span** main-assembly mapping (so the existing gapped-mapping guard never fires on this reference):

| | records |
|---|---|
| carrying ≥1 **indel** diff — the affine is wrong past the first one | **46** |
| carrying only `mismatch` diffs — the affine is exact | 119 |
| carrying no diffs | 1,129 |

Of the 46, 44 have a net width mismatch and 2 have indels that cancel (net 0) while still drifting between them. Largest net: `LRG_1321` at −769, `LRG_1075` at −569, `LRG_792` at +148, `LRG_839` at +23.

The issue's `{-8: 1, -5: 1, -1: 12, 0: 1269, +1: 3, +2: 2, +5: 2, +6: 2, +23: 1}` is a *windowed* measurement — one probe per accession — so its 25 is a lower bound. 46 is the structural count: the number of accessions whose alignment is not affine anywhere past their first indel.

## The fix

Narrow the span to its **affine-exact prefix** rather than accept it whole or decline it whole. `affine_horizon` reads the first LRG coordinate the affine can no longer reach (`lrg_start` for an `lrg_ins`; `lrg_end` for an `other_ins`, whose LRG coordinates are the flanks), and the span is trimmed at the end *away* from `parent_start` — off `nc_start` on the minus strand, off `nc_end` on the plus — so the anchor itself never moves. An empty prefix declines outright, which is the answer the gapped-mapping guard already gives.

Everything the returned placement still accepts, it maps exactly. Everything else is declined by `nc_to_parent`/`parent_to_nc` returning `None`, which `reanchor_genome_to_parent` already turns into `UnsupportedProjection` — "an endpoint falls outside the placed genomic span (#480/#655)" — rather than a drifted coordinate under the parent accession.

**No ruling was needed and no data model changed.** This applies a contract the file already states twice: `parse_lrg_main_assembly_placement` declines a gapped mapping, and `parse_refseqgene_alignments` skips a gapped `NG_` alignment "so the projector declines rather than emit a wrong `NG_` coordinate". The gap was that "one span" was being read as "affine".

### Why narrow, and not decline the accession

Declining all 46 outright was measured first, and it is worse: **31 of the 82 rows it refuses were landing on the right base**, because they sit below their record's first indel where the affine is exact. Narrowing refuses 51 and preserves those 31 — measured, not argued (table below).

## Unchanged on purpose

- **`mismatch` diffs do not narrow anything.** They substitute a base and shift no coordinate, and since #1490 the bases come from the LRG's own frozen record rather than being synthesized from the chromosome, so a mismatch has no remaining consequence here. Keying the horizon on any `<diff>` would have truncated 165 records instead of 46. Pinned by `mismatch_diffs_alone_do_not_narrow_a_placement`.
- **A `<diff>` under a non-main mapping is out of scope.** `body` is the main-assembly mapping's own inner text. Across the prepared reference the GRCh37 and transcript mappings carry 967 `lrg_ins` and 860 `other_ins` diffs that say nothing about the GRCh38 alignment. Pinned by `an_indel_diff_under_another_mapping_does_not_narrow_the_placement`.
- **Base serving is untouched.** `get_sequence`/`get_sequence_length` consult a placement only when `own_record(id).is_none()`, and **every one of the 1,294 LRG records in the prepared reference has its own `LRG_<N>g` record** (measured directly off the `.fai` files, 0 missing), so no LRG reaches `synthesize_parent_sequence` at all. #1462/#1490's fix is what makes this safe.
- **`NG_` placements are untouched.** The RefSeqGene path keeps only `gap_count=0` alignments, so those affines are exact by construction; #728-derived `NG_`s all have `parent_start == 1`.

## Deliberately left open — the exact fix, and an operator question

The issue's option (1) — parse the `<diff>` list and carry it alongside the affine, so the 51 refused rows get *correct* coordinates instead of a refusal — is the better answer and this PR does not take it. It needs an alignment model on `GenomicPlacement`, which is a public struct with public fields constructed at **47 literal sites** (35 under `src/`, 12 under `tests/`+`examples/`), so adding a field is a public-API break as well as a wide mechanical edit.

Two measurements for that decision, both taken here:

- **The diff data looks sound but is not trivially applied.** A quick reconstruction of each indel-bearing record from the chromosome plus its diff list reproduced **39 of 46 byte-for-byte**; the 7 that did not are cases with adjacent or *overlapping* diffs (e.g. `LRG_1321`'s `lrg_ins` whose chromosome flank sits inside an `other_ins` range of 1,000 `N`s). Those are defects in the throwaway reconstruction, not evidence of missing data — but they are exactly the edge cases a real implementation has to get right, so this is a tested alignment model, not a small patch.
- **What it would buy:** the 51 rows below.

**Question for the operator: take the `GenomicPlacement` data-model change (and its API break) to make these exact, or leave them refused?** Refusing is correct today; it is not complete.

Filed as **#1833** so the question survives this PR closing #1499, with the measurements above and the three edge cases an implementation has to handle (`LRG_1321`'s overlapping diffs, `LRG_792`'s 19 clustered indels, and the differing `lrg_start`/`lrg_end` convention between the two diff kinds).

## Relationship to #1503 — separable, and left open

**Separable, and #1503 says so itself.** Its closing paragraph declines to ride along with a length fix on the ground that touching `parent_to_nc` "is a projection-behaviour change and wants its own blast-radius measurement" — which is this PR. Fixing them together would do the thing that paragraph rules out.

They share a root cause (the affine cannot express indels) and one worked example (`LRG_1075` is one of the 46), but differ in reachability:

- **#1499 is live** — 51 projection rows measured wrong on the prepared reference today.
- **#1503 is latent, and its exposed population is empty here.** It needs a parent that is record-*less* **and** indel-bearing; #1503 says it had not established one exists, and this PR measured that it does not — all 1,294 LRG records have their own record, and every derived `NG_` has an exact affine.

**#1503 is deliberately left open by this PR** — no `parent_end` is added, and no closing keyword is used for it. It does move #1503's worked figure, so that comment is corrected in place: `get_sequence_length`'s expression on `LRG_1075` now yields **8,211** rather than the pre-#1499 **15,836**, against a true record length of 15,267. Wrong in the other direction, for the same reason, which is #1503's point. `GenomicPlacement`'s own doc comment now states that `nc_start`/`nc_end` are "the region this transform maps exactly", never the parent's extent, and points at #1503.

## Representation stability

Measured on the prepared reference (identity `aa8b3246d83055cc`), by dumping the whole LRG population at `origin/main` (`439617c2`) and at this branch and diffing row by row. The base run was taken by reverting **only** `src/reference/multi_fasta.rs` to `origin/main` with `cp`/`git checkout <ref> --`, never `git stash`.

Corpus: all 1,294 LRG accessions × {`LRG_<N>t1`, `LRG_<N>t2`} × {`c.100del`, `c.1000del`}, projected with `project_axis(.., Axis::Genomic, ..)` — the same call `ferro project --axis g` makes. 5,176 projection rows.

| | before | after |
|---|---|---|
| accessions with a placement | 1,294 | **1,294** (46 narrowed, 0 dropped) |
| projection rows `ok` | 2,815 | 2,733 |
| projection rows declined | 2,361 | 2,443 |
| **rows moved** | — | **51 (0.99%)**, across 29 accessions |
| of which `ok` → declined | — | **51** |
| of which `ok` → `ok` with a different string | — | **0** |

**All 51 are previously-accepted, so all 51 are a real migration for anyone who stored them** — and that is the whole disclosure, because nothing else changes. No accepted row changes its string: this PR does not correct a coordinate, it refuses one.

**Were they wrong?** Checked, not assumed. Every one of the 51 sits past its record's affine horizon; **0** sit below it, so there are no false refusals. Their measured drift:

```
drift (bases):  -8: 2   -5: 1   -1: 17   0: 2   +1: 14   +2: 4   +3: 2   +4: 2   +5: 2   +6: 3   +14: 1   +23: 1
```

So **49 named a demonstrably wrong base**. The 2 at drift 0 sat past a cancelling indel pair — the map through them is not exact, they happened to land right at that position.

**The rejected alternative, for the record.** Declining every indel-bearing accession outright — the issue's option (2) — moves **82** rows instead of 51, and all **31** extra rows were already correct (below their record's first indel, drift 0: `LRG_1075t1:c.100del`, `LRG_1321t1:c.1000del`, `LRG_628t2:c.100del`, …). It also drops 46 accessions' placements entirely, where narrowing drops 0. That measurement is why the fix narrows rather than declines.

**The synthetic corpus is blind to this and its zero would be structural.** `examples/dump_normalized_corpus.rs` constructs only `MockProvider`, whose own doc comment puts prepared references out of scope; no LRG XML, no `GenomicPlacement` parsed from one, so no edit to this code could move a row there. It was not run, and a zero from it would not have been evidence.

**Confluence guards:** unaffected — no partitioning, shifting or merge logic is touched, and nothing was re-blessed.

## Verification

Every exit status below was read back out of the run's own log, not from a wrapper's report.

```
  -> Summary [255.139s] 10535 tests run: 10535 passed (10 slow), 152 skipped   EXIT=0
scripts/run_oracle_suite.sh          # the runner, not a bare armed run — see CLAUDE.md
  -> Summary [ 64.191s] 10398 tests run: 10398 passed, 289 skipped             EXIT=0
```

**Nothing was re-blessed.** No committed expectation pinned a drifted LRG projection — which is the strong signal for an output-moving change.

**The new tests are discriminating, checked by sabotage rather than assumed.** Forcing `affine_horizon` to return `None` (i.e. reverting to the pre-fix behaviour, in place) turns **all six** new tests red:

```
FAIL an_indel_bearing_span_is_narrowed_to_its_affine_exact_prefix
FAIL a_chromosome_insertion_narrows_the_span_too
FAIL the_prefix_is_measured_from_parent_start_not_from_one
FAIL no_placement_when_the_affine_exact_prefix_is_empty
FAIL an_unreadable_indel_diff_declines_the_placement
FAIL provider_narrows_an_indel_bearing_lrg_placement
Summary  161 tests run: 155 passed, 6 failed
```

Both negative controls — `mismatch_diffs_alone_do_not_narrow_a_placement` and `an_indel_diff_under_another_mapping_does_not_narrow_the_placement` — are in the 155 that stayed green, which is what makes them controls rather than consequences of the fix.

The corpus dump above was re-run against the final tree after the review findings were applied, and is **byte-identical** to the one the table reports, so the numbers describe the shipped code and not an earlier draft.

The `LRG_292` fixture is the real record's mapping, trimmed to its span, one `mismatch` and all three indel diffs, and it asserts a `<mapping_span>` count of 1 so it cannot silently start passing through the gapped-mapping guard instead.

### Review pass

A CodeRabbit-style self-review against the local diff, run pre-push under this repo's own `.coderabbit.yaml` — `assertive` profile, with the `src/reference/**/*.rs` path instructions applied as lenses. Four findings, all applied, none rejected:

1. **Fail closed on an unreadable indel `<diff>`** (major) — the first draft skipped a diff whose coordinate attribute was missing or unparseable, leaving the span *wider* than the data supports. That is the silently-wrong-coordinate direction the whole change exists to close, and the path instruction names it ("fail closed on a truncated or malformed index"; "any provider that returns a default … where it should return an error"). An unreadable indel diff now yields horizon `0` and declines the placement. Pinned by `an_unreadable_indel_diff_declines_the_placement`. Measured non-firing on real data: **0 of 110** main-assembly indel diffs are unreadable.
2. **Doc drift on `GenomicPlacement`** (major) — the struct doc described `nc_start`/`nc_end` as the parent's placement, which a narrowed span is not. Corrected, with the #1503 cross-reference.
3. **Doc drift in `get_sequence_length`** (major) — its worked `LRG_1075` figure (15,836) is a number this diff changes. Corrected to 8,211 with the direction of the error restated.
4. **Test gap on `parent_start > 1`** (major) — `h - 1 - parent_start` is where an off-by-one hides, and the reference's only such record (`LRG_1075`) is itself indel-bearing, so it goes through the new path. `the_prefix_is_measured_from_parent_start_not_from_one` pins it on the real shape.

Representation-Change: 51 of 5176 LRG projection rows move, all previously-accepted and all to a refusal; 0 accepted rows change their string. Measured over the 1294 LRG records of the prepared reference; 49 of the 51 named a demonstrably wrong base (drift -8..+23) and 0 were below their record's affine horizon.

### DIFF
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -3080,6 +3080,86 @@ fn xml_attr<'a>(tag: &'a str, name: &str) -> Option<&'a str> {
     Some(&rest[..end])
 }
 
+/// The first LRG coordinate at which a mapping stops being an affine (#1499),
+/// or `None` when it never does.
+///
+/// **An LRG↔chromosome alignment is not affine, and one `<mapping_span>` does
+/// not make it one.** Alongside the span the record lists `<diff>` elements —
+/// the vocabulary is closed at `mismatch`, `lrg_ins` (bases present in the LRG
+/// with no chromosome counterpart) and `other_ins` (the reverse), verified over
+/// all 1 294 LRG records in the prepared reference. The two insertion kinds
+/// shift every coordinate past them, in opposite directions, and
+/// [`GenomicPlacement`] is a pure affine with nowhere to carry them: composing
+/// it with cdot's `NM_`→`NC_` step therefore drifts by the *net* indel count
+/// between the LRG origin and the projected position.
+///
+/// `LRG_292` is the worked case. One span, `1..193689` onto
+/// `NC_000017.11:43024295..43217981` — but the chromosome side is 193 687
+/// bases, two short, from 3 `lrg_ins` bases against 1 `other_ins` base.
+/// Measured against the frozen `LRG_292g` record, the affine is exact up to
+/// `LRG_292:g.13168` and then off by +2 for the remaining ~93 % of the record,
+/// so `LRG_292(NM_007294.4):c.1del` projected to `LRG_292:g.93888del` where the
+/// record's own frame says `g.93890del`. The result is well-formed and
+/// in-bounds, so none of the four normalization oracles can see it.
+///
+/// Returning the departure point rather than a bare "is it affine" lets the
+/// caller keep the exact prefix instead of dropping the accession. That matters:
+/// over the prepared reference, refusing the whole of every indel-bearing record
+/// would have refused 31 projections that were already landing on the right
+/// base, alongside the 51 that were not.
+///
+/// `mismatch` is deliberately absent: it substitutes a base and shifts nothing,
+/// so the affine still maps every coordinate exactly — and since #1490 the bases
+/// themselves come from the LRG's own frozen record rather than being
+/// synthesized from the chromosome, so a mismatch has no remaining consequence
+/// here. Over the prepared reference that is the difference between narrowing 46
+/// records and narrowing 165.
+///
+/// **Fails closed.** An indel `<diff>` whose coordinate attribute is missing or
+/// unparseable yields a horizon of `0`, which the caller reads as "nothing is
+/// affine" and declines the placement outright. Skipping such a diff would leave
+/// the span *wider* than the data supports — a silently wrong coordinate rather
+/// than a refusal, which is the direction this whole fix exists to close.
+fn affine_horizon(mapping_body: &str) -> Option<u64> {
+    mapping_body
+        .match_indices("<diff ")
+        .filter_map(|(i, _)| {
+            let rest = &mapping_body[i..];
+            let tag = &rest[..rest.find('>').unwrap_or(rest.len())];
+            let attr = match xml_attr(tag, "type") {
+                // LRG bases `lrg_start..=lrg_end` are unaligned: `lrg_start` is
+                // the first coordinate the affine cannot reach.
+                Some("lrg_ins") => "lrg_start",
+                // Chromosome bases sit *between* the flanking LRG coordinates,
+                // so `lrg_start` still maps and `lrg_end` is the first that
+                // does not.
+                Some("other_ins") => "lrg_end",
+                // The one type that is legitimately skipped: a `mismatch`
+                // substitutes a base and shifts no coordinate, so the affine
+                // still maps everything past it exactly.
+                Some("mismatch") => return None,
+                // Absent, or outside the closed vocabulary. Skipping it would
+                // leave the span *wider* than the data supports, which is the
+                // silently-wrong-coordinate direction this whole pass exists to
+                // close — the identical argument the missing-coordinate case
+                // below is already decided on, so it gets the identical answer:
+                // horizon `0`, and the caller declines the placement.
+                //
+                // Measured non-firing on real data: over all 1 294 LRG records
+                // of the prepared reference the `<diff type=…>` vocabulary is
+                // exactly `mismatch` (7 945), `lrg_ins` (1 030) and `other_ins`
+                // (907), and **0** `<diff>` elements carry no `type` at all.
+                _ => return Some(0),
+            };
+            Some(
+                xml_attr(tag, attr)
+                    .and_then(|v| v.parse().ok())
+                    .unwrap_or(0),
+            )
+        })
+        .min()
+}
+
 /// Parse an LRG's chromosomal placement from its XML record (#480).
 ///
 /// Reads the `<mapping type="main_assembly" …>` element (the current assembly,
@@ -3089,6 +3169,21 @@ fn xml_attr<'a>(tag: &'a str, name: &str) -> Option<&'a str> {
 /// a parseable accession, or it is not a single contiguous span (a single
 /// affine placement cannot represent a gapped mapping — the projector then
 /// keeps the chromosome coordinates rather than mis-anchor).
+///
+/// The span is additionally narrowed to its affine-exact prefix when the mapping
+/// carries indel `<diff>`s — see [`affine_horizon`] — and `None` is returned when
+/// that prefix is empty. Every coordinate the returned placement still accepts is
+/// one the affine maps exactly; the rest are declined by
+/// [`GenomicPlacement::nc_to_parent`] / [`GenomicPlacement::parent_to_nc`]
+/// returning `None`, which
+/// [`VariantProjector::reanchor_genome_to_parent`](crate::project::VariantProjector)
+/// already turns into an `UnsupportedProjection` rather than a drifted
+/// coordinate under the parent accession (#480/#655).
+///
+/// Base *serving* is unaffected: `get_sequence` / `get_sequence_length` consult a
+/// placement only when `own_record(id).is_none()`, and every one of the 1 294 LRG
+/// records in the prepared reference has its own `LRG_<N>g` record (measured), so
+/// no LRG reaches `synthesize_parent_sequence` at all (#1462/#1490).
 fn parse_lrg_main_assembly_placement(xml: &str) -> Option<GenomicPlacement> {
     let mut search = xml;
     loop {
@@ -3122,6 +3217,7 @@ fn parse_lrg_main_assembly_placement(xml: &str) -> Option<GenomicPlacement> {
             if spans.len() != 1 {
                 return None;
             }
+
             let span = spans[0];
             let parent_start: u64 = xml_attr(span, "lrg_start")?.parse().ok()?;
             let o1: u64 = xml_attr(span, "other_start")?.parse().ok()?;
@@ -3131,6 +3227,36 @@ fn parse_lrg_main_assembly_placement(xml: &str) -> Option<GenomicPlacement> {
                 _ => crate::reference::transcript::Strand::Plus,
             };
             let (nc_start, nc_end) = if o1 <= o2 { (o1, o2) } else { (o2, o1) };
+
+            // A single span is necessary but **not sufficient** for an affine
+            // (#1499) — so narrow the span to the part that genuinely is one.
+            // `body` is this mapping's own inner text, so the GRCh37
+            // `other_assembly` mapping's diffs (967 `lrg_ins` / 860 `other_ins`
+            // across the prepared reference) are correctly out of scope.
+            let (nc_start, nc_end) = match affine_horizon(body) {
+                None => (nc_start, nc_end),
+                // Nothing at or after `parent_start` aligns: no usable affine.
+                Some(h) if h <= parent_start => return None,
+                Some(h) => {
+                    // Bases past the anchor that still map exactly. `parent_start`
+                    // is the parent coordinate of `nc_start` on the plus strand
+                    // and of `nc_end` on the minus strand, so the trim always
+                    // comes off the end *away* from the anchor and the anchor
+                    // itself is untouched.
+                    let placed = h - 1 - parent_start;
+                    use crate::reference::transcript::Strand;
+                    match strand {
+                        Strand::Minus => (nc_end.checked_sub(placed)?.max(nc_start), nc_end),
+                        // The parser above emits only Plus/Minus; `Unknown` is
+                        // spelled out rather than left to a wildcard so a new
+                        // variant has to be considered here. It follows
+                        // `parent_to_nc`, which also treats Unknown as ascending.
+                        Strand::Plus | Strand::Unknown => {
+                            (nc_start, nc_start.checked_add(placed)?.min(nc_end))
+                        }
+                    }
+                }
+            };
             return Some(GenomicPlacement {
                 nc,
                 parent_start,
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

nh13:
⚠️ **Merge-order hazard with the sibling PR — please read before merging this one.**

This PR and its sibling are both `MERGEABLE`/`CLEAN` right now, and they are **semantically opposed**: #1830 narrows an LRG placement to its affine-exact prefix and declines past it, while #1847 models the `<diff>` gaps so exactly those coordinates are answered. One declines where the other answers.

They share `src/reference/multi_fasta.rs` and `src/reference/provider.rs`, and `git merge-tree --write-tree --name-only <1830-head> <1847-head>` exits **1** — the conflict is real, not `mergeStateStatus` lag. It just does not appear until the first of the two lands.

**Whichever merges second needs a rebase that DROPS `affine_horizon`, not a merge that keeps it.** Keeping both sides is the default outcome of ordinary conflict resolution and it is the failure: `affine_horizon` clips the coordinates #1847 teaches the provider to answer, so #1847 merges and its coverage silently does not arrive. A green suite will not witness that, because a decline is a legal answer nothing asserts against.

Full write-up, including what to verify on `main` afterwards: #1873.

nh13:
**Merge-order hazard, re-verified on the current heads (2026-08-14).** This PR (`d755463b`) and **#1847** (`9f011e5c`, `fix/issue-1833-lrg-diff-model`) are semantically opposed and both read green. `git merge-tree --write-tree --name-only d755463b 9f011e5c` exits **1**, conflicting in exactly `src/reference/multi_fasta.rs` (11 hunks) and `src/reference/provider.rs` (1 hunk) — but each still merges into `origin/main` on its own at exit 0, so nothing in GitHub's state will stop the wrong order.

The opposition is stated by the code, not just the descriptions: on the same `LRG_292` fixture this branch asserts `p.parent_to_nc(160875) == None`, while #1847 asserts `Some(43057109)`.

**If this merges first, #1847's rebase must *drop* `affine_horizon` and the narrowing block in `parse_lrg_main_assembly_placement`, not merge them.** That is a well-scoped edit — this PR's `provider.rs` change is doc-comment only, so the whole mechanism is those two pieces in `multi_fasta.rs`.

Full analysis, including the named coordinate, what keeping both sides actually produces, and a runnable post-merge check, is in **#1873**. No preference expressed here — the choice between the two designs is the maintainer's.

nh13:
**Maintainer decision on #1873: #1847's `<diff>` modelling is the resolution being taken, and `affine_horizon` does not land.**

Not a judgement on this PR's correctness — the narrowing is a defensible answer to the same defect, and its measurement stands (51 `ok → declined`, 0 `ok → ok`). It is subsumed: #1847 answers at the same anchor everything this PR still answers, plus the 55 coordinates inside the gaps, and it also fixes `get_sequence_length`'s fallback, which this PR's narrowing turns from over- to under-reporting (8211 against a true 15267 on `LRG_1075`).

The one case where this PR is strictly better — returning the affine-exact prefix where #1847 declines the whole placement on an integrity-check failure — is measured at **0 of 1294** today.

Suggested disposition: close, or reduce to the doc-comment half. The `provider.rs` change here is 11 `///` lines and carries no behaviour, so if that documentation is worth keeping it can land on its own; what must not survive is `affine_horizon` in `multi_fasta.rs`.

**Do not resolve a post-#1847 rebase by keeping both sides** — that trips both of #1847's integrity checks and costs all 46 indel-bearing records their placement, with both branches' tests red. Full reasoning: https://github.com/fulcrumgenomics/ferro-hgvs/issues/1873#issuecomment-5295907571

PR #1835, opened 2026-08-13
feat(normalize)!: make canonical-coalesced the default partition rule

nh13:
Makes `canonical-coalesced` the default partition rule: an unset or empty `FERRO_PARTITION` now selects it instead of `live`. `live` is unchanged and still selectable by name, which is what keeps pre-flip output reproducible for anyone comparing against a stored corpus.

**The #1706 blocker is cleared and this is ready to merge.** #1706 merged as `64fc76df`; the whole-block inversion family described under "The blocker" below no longer fires, re-measured on this base rather than predicted. The failure count went 83 -> 5 -> 0 across that merge and the dispositions in this PR. What is still owed is the real-corpus row count, tracked as #1885 and stated in the trailer.

## Why the flip

`canonical-form-choice-when-both-legal` and `derivation-may-not-be-bounded-by-the-inputs-spelling` (both `decided`, both on `main`) require a description to be derived from the **resulting sequence** rather than preserved from the input's spelling. `partition_block` — `live`'s cutter — structurally cannot do that: it is a single-gap alignment search plus two narrow escapes, so several spellings of one variant can be several fixed points.

`delins-recommendation-reach-when-the-input-arrives-split` (decided 2026-08-12, merged as #1760 and now the tip of `main`) states in terms that the pass it ratifies "runs only under `PartitionRule::CanonicalCoalesced`, which is not the shipped default. So the real-corpus disclosure is owed by the change that makes that arm the default, not by this record." That change is this one.

## What it buys, measured on this base

`it::cis_confluence_axis`, the 11,272-class designed cis corpus, both directions, re-derived here rather than carried in:

| direction | converged before | converged after | divergent before | divergent after |
|---|---|---|---|---|
| 3' | 8,361 | **11,271** | 2,911 | **1** |
| 5' | 8,367 | **11,272** | 2,905 | **0** |

`sequence_changed` is **0** in both directions before and after, so no class converged onto the wrong bases. The single surviving 3' divergence is `s06-g-m4-sep1-p1-all-dup`, a repeat-notation class (`g.[265dup;267_271A[7];271_272insA]` vs `g.[265dup;267_271A[8]]`) — the repeat carve-out that `delins-adjacent-members-when-both-consume-reference` already records as untouched.

## The moved set

Controlled comparison, same tree, same manifest, the only difference being which arm an unset `FERRO_PARTITION` selects:

| run | result |
|---|---|
| `live` default (this branch minus the 3-line default change), manifest armed | 10572 run, **10572 passed, 0 failed**, `EXIT=0`, 457s |
| `canonical-coalesced` default, manifest armed | 10572 run, 10489 passed, **83 failed**, 152 skipped, `EXIT=100`, 1019s |
| `canonical-coalesced` default, **no** manifest | 10572 run, 10496 passed, **76 failed**, 152 skipped, `EXIT=100`, 760s |

So all 83 are attributable to the arm and to nothing else. The two flipped runs reconcile exactly, 83 − 9 + 2 = 76: seven failures are manifest-gated and skip without it (`delins_equal_vs_unequal_length_discriminator` — so D6 is manifest-only as expected — `multi_member_cis_axis::axis_normalized`, and five `mutalyzer_normalize_tests` axes), one is the arm assertion this PR re-pins, and `spec_enumeration_tests` trips a *different* guard depending on the manifest: `enumeration_replays_recorded_behavior` armed, versus `divergence_budget_is_unchanged` and `passing_census_is_unchanged` unarmed. That last one is worth knowing before reading either number as the enumeration's verdict.

Largest clusters:

| n | module | family |
|---|---|---|
| 7 | `issue_1040_inv_overrecognition_probes` | inversion (was STOP; cleared by #1706) |
| 5 | `issue_1454_split_member_inversion_typing` | inversion (was STOP; cleared by #1706) |
| 5 | `inversion_sweep` | inversion (was STOP; cleared by #1706) |
| 1 | `coverage_gap_tests::inversions` | inversion (was STOP; cleared by #1706) |
| 6 | `mutalyzer_normalize_tests` | manifest-backed axis censuses |
| 5 | `reported_partition_verdicts` | owned by another change; untouched here |
| 4 | `cis_confluence_nr_axis` | `n.`/`r.` census instruments |
| 3 | `cis_adjudication_enumeration` | class-partition instruments |
| 3 | `cis_confluence_adjudication` | D2 / D3 / D4 |
| 2 | `cis_confluence_axis` | the confluence censuses above |
| 2 | `spec_conformance_axis` | the coding-axis merge instrument |
| 2 | `long_delins_splits_at_unchanged_bases` | D5(a) and D5(b) |
| 6 | `cis_junction_crossing_shift`, `issue_1320`, `issue_1301`, `issue_1261` | exactly the six re-pins `duplication-must-ranks-the-label-not-the-partition` names by full path |

## What is NOT a blocker, corrected against a stale figure

`spec_conformance_axis.rs`'s module docs record that this arm "fails on an unrelated pre-existing counter (`guard_violations` 0 -> 6)". **That does not reproduce.** Both censuses report `NEGATIVE GUARDS: 0 outputs implement rejected SVD-WG010`, at 3' and at 5'. The flip does not implement the rejected proposal on any of the 210 guarded rows, so the one counter whose rise the suite calls "a rank-1 conformance regression, not a representation choice" stays clean.

What actually fires there is `coding_axis_separation_two_or_more_merges`, **0 → 3 of 997**, in both directions — the figure that module's own table predicts for this arm. The test labels it "an INSTRUMENT, not a verdict … a population that grew, and it needs explaining and re-pinning". That is a licensed re-pin, not a STOP.

## The blocker: the whole-block inversion family — CLEARED

**Kept for the record; every claim below was true on the pre-#1706 base and is no longer live.** Re-measured after #1706 merged, `NC_000013.10:g.100809575` has ZERO occurrences in the failure set and all four inversion modules pass. The prediction that #1706 would clear it was exactly that — a prediction — and it has now been run rather than assumed.

The flip's canonical partition cuts `NC_000013.10:g.100809575_100810031inv` — a perfect 457-base reverse complement — into **133 pieces**, one of which is the lone substitution `g.100809575A>T`. `main`'s admission gate `no_piece_is_a_lone_substitution` withdraws the spanning `inv` unless *no* competing member is a lone substitution, so the `inv` is refused and the row comes out as its own 133 columns.

That violates `inversion-vs-two-delins-76-83`, status **`decided`**, `DNA/inversion.md:5` governing: a whole-block reverse complement is one `inv` when the members it competes with are not substitutions the prioritisation ranks above it.

It also puts 12 pairs of members on consecutive nucleotides. Classified against `delins-adjacent-members-when-both-consume-reference`:

- **9 of 12** are in that record's **decided** reach — its `ins` half, closed by the operator ruling of 2026-08-11, whose scope is "one member of an adjacent pair is an `ins` **and the other consumes reference bases**": seven `sub`/`ins`, one `delins`/`ins`, one `ins`/`inv`.
- **2 of 12** are `ins`/`ins` pairs. Neither member consumes reference, so the conjunction above is not satisfied and the ruling does not reach them.
- **1 of 12** is the still-open `dup` carve-out (`100809699dup` / `100809700_100809701insT`).

Nine rows inside a decided record is what makes this a STOP rather than a disclosure. The two `ins`/`ins` pairs are worth stating separately rather than folded in: reading the `ins` half as "any pair with an `ins` on either side" gives 11 and overstates the record, which is the paraphrase this ledger's own text warns about.

So this is not a representation choice to disclose and re-pin. It is a row the ledger protects, and re-pinning it would encode a violation of two decided records as the new expected value.

**#1706 is the fix**, and independently: its body names this exact accession and reports the same 133-piece partition, weakens the gate to `not_every_piece_is_a_lone_substitution`, and carries its own operator ruling `inversion-vs-a-mixed-member-competitor` (2026-08-12) — a record that is **not yet on `main`**.

The family is the largest single cluster in the flip's failure set: `inversion_sweep`, `issue_1040_inv_overrecognition_probes`, `issue_1454_split_member_inversion_typing`, `coverage_gap_tests::inversions::test_minus_strand_inversion`, and `case_harvest`'s row above.

## The adjudicated rows, measured

Recorded so the next session does not re-derive them. None is re-pinned here — see the section below for why.

| row | `live` | `canonical-coalesced` | licensing record |
|---|---|---|---|
| `two_adjacent_members_that_both_consume_reference_are_one_delins` | `c.10_13delinsTAAT` | `c.[9dup;13del]` | separation is **3**, so neither the codon exception (needs exactly 1) nor the adjacency ruling (needs 0) fires; `general.md:34` governs |
| `a_dup_flush_against_a_del_is_left_alone` | `c.[9T>A;16_23dup;24_31del;34_35insCACCAAAA]` | `c.[9T>A;24_25insTTT;28T>A;31_32insCACCA]` | no separation-zero `dup`/`del` pair survives, so the carve-out has nothing to protect — this is the whole-block re-derivation case `duplication-must-ranks-the-label-not-the-partition` describes as its own domain |
| `a_spanning_delins_and_its_aligned_split_are_two_fixed_points` | `c.9_17delinsA` | `c.[9del;11_17del]` | both members are **pure deletions**, so nothing was inserted and `:46`'s mechanism cannot have occurred — `delins-recommendation-reach-when-the-input-arrives-split`'s own W58 shape |
| `equal_and_unequal_length_delins_get_opposite_verdicts` (MSH2, `g.`) | `g.[47639670_47639671del;47639673G>T]` | `g.[47639670_47639671delinsT;47639673del]` | `g.` axis, so `delins-payload-coincidence-carve-out-is-coding-dna-scoped` puts it outside `:47` and `general.md:34` governs |
| `a_long_net_deletion_stays_one_spanning_delins` (52 nt, `g.`) | `g.1_52delinsTTCCTCGATGCCTG` | 14 members | same axis scope |
| `a_delins_longer_than_the_old_cap_still_splits` (40 nt, `g.`) | 2 members | **21** members | same axis scope; this is the D5(b) frame-dependent row |

**Two corrections that matter for whoever lands this.** The `equal_and_unequal_length_delins_get_opposite_verdicts` row carries an "assert-then-flip" note naming `MSH2_RULING_CONFORMANT` (the spanning `g.47639670_47639673delinsTT`) as the right answer once the ruling is implemented. That guidance is **stale**: the row is `g.`, and `delins-payload-coincidence-carve-out-is-coding-dna-scoped` — decided, and later than the note — puts every `g.` row outside `:47` entirely. Neither arm produces the spanning form and neither should. The note should be rewritten, not followed.

And `two_adjacent_members_that_both_consume_reference_are_one_delins` states its authority incompletely. `live`'s answer takes **two** steps: the coding codon exception (`general.md:35` / `DNA/delins.md:18`, precondition separation exactly one) first produces a flush pair, and only then does the adjacency ruling merge it. The flip removes the **first** step — it presents separation 3 — not the second. The adjacency ruling is not contradicted by that row's movement.

## The re-pins, now made on top of #1706

The earlier draft of this section argued the re-pins belonged in the change that lands the flip, on top
of #1706, because their values were not stable until that PR moved what the canonical partition renders
for every inversion-bearing block. That is what has happened: this PR is now rebased past #1706 and the
re-pins are made here, against a base that is no longer moving.

`duplication-must-ranks-the-label-not-the-partition` (decided 2026-08-13) names six of them by full path
as authorised re-pins under this arm, and all six moved exactly as it predicts.

## The manifest-backed conformance axes

Both arms measured on this tree against the same prepared GRCh38 reference, differing only in
`FERRO_PARTITION`:

| axis | `live` | `canonical-coalesced` |
|---|---:|---:|
| `genomic` | 2 | 16 |
| `normalized` | 1 | 18 |
| `coding_protein_descriptions` | 0 | 12 |
| `rna_description` | 0 | 1 |

The three `live` failures are pre-existing and unrelated. The 45 new rows are **one mechanism**: a block
whose payload shares bases with the reference span is derived into members where mutalyzer keeps the span.

They are dispositioned `accepted_divergence`, because ferro's split is the canonical form here and the
**direction** is what decides that. `delins-merge-vs-individual-gap-two-or-more` recommends the spanning
`delins`, but its 2026-08-11 operator ruling scopes that to the net-DELETION case and says it does not
reach net insertions, "where the split form remains canonical". Every row is a net insertion — span 6 nt
against a 21 nt payload on `NG_008939.1`, 2 against 5 on `NG_007485.1`, 5 against 31 on `NM_003002.4`.

This was got **wrong once on this PR** and the correction is worth recording: the rows were first
annotated `known_bug` against a freshly-filed issue on the strength of `:47` alone, without reading the
direction scope. That issue is closed as not-a-defect, and the cluster note now carries the scope plus a
warning against re-making the same citation.

Two rows shed their annotations instead — both now match mutalyzer and were XPASS-failing, one a
`ferro-policy-654` divergence that had gone stale unnoticed.

`NONCODING_NON_IDEMPOTENT` is re-pinned 21 -> 62 with its shape breakdown: 16 + 4 + 1 became 16 + 45 + 1,
so the entire movement is that same re-partitioning while the repeat shape and the one non-3'-shifted row
are unchanged. **That is a different claim from the rows above**: the projection emits the span and
re-normalizing it yields the members, so the two disagree whichever spelling is canonical. It is #1712,
enlarged by this flip, and the comment says so rather than licensing it.

Note none of this is visible to any PR job — CI never sets `FERRO_MANIFEST` (#1874).

## What is in the diff

- `DEFAULT_PARTITION_RULE`, named once because two call sites need it and must not drift: the unset arm of `partition_rule_from_env`, and the release-build fall-safe in `partition_rule_from_outcome`. The fall-safe moves with it deliberately — serving `live` there would answer a misspelled `FERRO_PARTITION` with a rule no unmodified process uses, so a typo would change output rather than merely fail to change it.
- `an_absent_value_selects_the_default_rule` (was `…_the_live_rule`) now asserts unset and `live` **apart**, with a guard that fails if the two arms ever become one value again.
- `the_suite_runs_on_the_default_rule` (was `…_the_live_rule`) asserts `partition_rule() == DEFAULT_PARTITION_RULE`. This is the flip's sharpest positive control: it reads arm identity out of the process rather than inferring it from an output string. Before the flip it failed `left: CanonicalCoalesced, right: Live`.
- CLI `normalize --help` and the README section re-labelled for the new default.

## The README positive control was stale, and is re-measured

The committed per-arm control table was wrong in **four of its twelve cells**. `live` row 1 read `c.[5_6insA;7_9delinsCAA]`, not the documented `c.6_9delinsACCAA`; and its rows 2 and 3 separated **no arms at all** — every arm answered `c.[2del;6del]` and `c.[2del;11del]`. Cause: `NM_001234.1` is a G homopolymer from `c.9` to `c.33`, so deletion pairs inside it shuffle to a common form on every arm instead of partitioning differently. This is the second time this control has been wrong in that way; the note in the README now says so.

Replaced with three `delins` rows that separate **all six pairs of arms**, so it identifies *which* arm is in effect:

| input | `live` | `shadow` | `canonical` | `canonical-coalesced` (= unset) |
|---|---|---|---|---|
| `c.[5_6insAC;9del]` | `c.[5_6insA;7_9delinsCAA]` | `c.[5_6insAC;11del]` | `c.[5_6insAC;11del]` | `c.[5_6insAC;11del]` |
| `c.2_6delinsGA` | `c.2_6delinsGA` | `c.2_6delinsGA` | `c.[2del;4_6delinsA]` | `c.2_6delinsGA` |
| `c.4_10delinsAC` | `c.4_10delinsAC` | `c.4_10delinsAC` | `c.[4C>A;6_10del]` | `c.[4C>A;6_10del]` |

## Also measured

Regenerating `hgvs_spec_normalization.json` before and after the flip gives a **byte-identical** file (md5 `e09899559c510c248968b435ad4ebdb5`). That independently reproduces #1822's finding that the spec fixture is blind to the partition switch — now measured on the *default* rather than on the environment variable. "The flip breaks no spec test" is true by construction there and is not evidence.

Representation-Change: this is the largest representation change the project has made. On the 11,272-class designed cis corpus every divergent class but one converges: 2,910 classes change their normalized form at 3' and 2,905 at 5', with `sequence_changed` 0 on both sides, so every move is a re-spelling and none changes denoted bases. Direction of movement is overwhelmingly merge — a payload-coincidence split becomes the spanning `delins` on the `c.` axis, and members that `live` left as the input's own spelling are re-derived from the resulting sequence. On the manifest-backed conformance axes a further 45 rows move, all one mechanism and all net insertions where the split form is canonical. A consumer keying on the exact normalized string must expect previously-accepted inputs to return a different, still-conformant description, and must pin `FERRO_PARTITION=live` to reproduce pre-v0.15.0 output. The real-corpus row count over ClinVar/CMRG/Paraphase is NOT supplied here and remains owed; it is tracked as #1885. The earlier reason given for its absence — that the unresolved inversion family would contaminate it — is stale now that #1706 has merged, so the honest statement is that it is derivable and has not been run.


nh13:
@coderabbitai pause

nh13:
## Two corrections to the `Representation-Change:` trailer

Both for whenever the body is rewritten after the post-#1706 re-measure — filed as a comment rather than edited into the description so nothing is overwritten under an in-flight rebase.

Context for why the trailer text specifically matters, since it is easy to assume the changelog will say more than it does: `release-plz.toml` writes each changelog bullet **from the commit subject alone**, and `scripts/inject_representation_disclosure.py` attaches the trailer's own text underneath it. So this trailer *is* the published disclosure. Nothing else in this description reaches the changelog.

### 1. Naming a version is wrong now and fragile either way

The trailer closes with:

> …and must pin `FERRO_PARTITION=live` to reproduce pre-v0.15.0 output.

`release-plz.toml` sets `features_always_increment_minor = true`, so a `feat:` against the current `0.13.1` cuts **0.14.0** — and the flip was ruled to ship in v0.14.0. So `v0.15.0` is off by one release today.

But swapping the digit is not the fix. Six PRs merged today, and if any release cuts before this one lands, the flip's release becomes 0.15.0 after all and the corrected number is wrong again in the other direction. A version-free phrasing is correct under both futures:

> …and must pin `FERRO_PARTITION=live` to reproduce output from before this change.

The release that carries it names itself in the changelog heading.

### 2. Conformance is absent from the disclosure, and it moves opposite to confluence

The trailer discloses confluence only. Spec conformance has since been measured per arm over the whole corpus — 932 of 934 rows, 0 dropped, exclusions stamped — and it **falls**:

| arm | 3′ conforms | 5′ conforms |
|---|---|---|
| `live` / `shadow` | 650 / 932 | 627 / 932 |
| `canonical` | 646 | 624 |
| **`canonical-coalesced`** | **648** | **626** |

Denominator reaching the partitioner at all: 232 rows at 3′, 241 at 5′.

Three things follow that the current trailer would leave a reader to discover after upgrading.

**The cost is 2 rows on the shipped arm** (4 on plain `canonical`), not the 1 an earlier restricted 35-row window suggested. Three of the four `canonical` losses are single-member `delins`, which that window excluded by construction — it could only ever have shown one.

**It is not a regression, and the reason is structural.** `spec_expected` defaults to the row's own input, so `preserved` is a round-trip test. Where the spec publishes two legal spellings of one variant as two rows, both echo and both score as conforming — so *any* change that converges them must cost one. Conformance-down and confluence-up are one trade, and it is the trade `canonical-form-choice-when-both-legal` asks for. On the flagship row `dna-delins-vs-aligned-split-850-901`, `live` emits two forms; `canonical` emits a five-member split matching neither published spelling; `canonical-coalesced` emits `c.850_901delinsTTCCTCGATGCCTG`, the spanning delins `delins.md:47` recommends.

**Any sentence about it needs a direction.** On the restricted set 3′ read 11/11/10/10 while 5′ read 10/10/10/10 — the "falls" delta is 3′-only there, because at 5′ `live` already diverges.

Suggested addition to the trailer, after the existing confluence sentences:

> Spec conformance falls by 2 rows of 932 on this arm in both shuffle directions — 650→648 at 3′ and 627→626 at 5′ — because `spec_expected` defaults to each row's own input, so the spec's two published spellings of one variant both round-trip today and converging them necessarily costs one. No row loses conformance by becoming non-conformant; each loses it by adopting the sibling spelling the spec recommends.

Re-derive those four figures after the re-measure rather than carrying them over from this base.

One phrasing to avoid in the release note: **"more conformant."** The measurement does not support it in either direction. What it supports is that the change converges the spec's own published alternative spellings onto the ones `delins.md:47` recommends, and that the round-trip echo of `:46`'s alternative description is what pays for it.

Worth noting for whoever re-derives these: `canonical-coalesced` beats plain `canonical` at both 3′ and 5′, which is a third independent argument for shipping the coalesced arm rather than the plain one.

nh13:
## Measured: this flip **aggravates** #1712, and PR CI cannot see it

Surfaced while measuring which open `high`-labelled issues the coalesced arm fixes. Posting before the re-measure rather than after, because it changes what the re-measure needs to look at.

### The measurement

Arm A vs arm B on `main` at `64fc76df` (i.e. already carrying #1706), `FERRO_MANIFEST` set to the prepared scratch reference, `EXIT=` echoed into each log and read back out of the file:

| arm | `axis_noncoding_idempotent` | non-idempotent rows |
|---|---|---|
| `live` | `EXIT=0` | **21** of 681 |
| `canonical-coalesced` | `EXIT=100` | **62** of 681 |

Both arms ran 16–23 s, so the test was **armed, not skipped** — the usual failure mode where a manifest-gated test reads ~0.02 s and a skip is mistaken for a pass does not apply here.

The pin is `NONCODING_NON_IDEMPOTENT` in `tests/it/mutalyzer_normalize_tests.rs`, currently `23` on `main` (raised from 21 by #1808, which has since rebased onto `64fc76df` and is green). Either way the pin fails under this arm: 62 ≠ 21 and 62 ≠ 23.

### Why this needs saying out loud

**PR CI cannot catch it.** `ci.yml` never sets `FERRO_MANIFEST` — so on a pull request this test early-returns and reports green. The only job that can observe it is `nightly-mutalyzer.yml`, whose `REFERENCE_AWARE_SELECTION` does name `mutalyzer_normalize_tests::`, and which is `continue-on-error`.

So the sequence to expect, absent a deliberate re-pin: this PR merges with every required check green, and the regression appears in a non-blocking nightly some hours later, in a job whose failures do not gate anything.

### What I am and am not claiming

- The 21 → 62 delta **is measured**, on this base, with the arms as the only difference.
- Whether those 41 extra rows are a *defect* or a licensed representation move **is not measured**. They may be exactly the re-spellings this arm is supposed to produce, in which case the answer is a re-pin with a disclosure, not a fix. Nobody has looked at the rows.
- This is not the inversion family. It is the `n.`/non-coding axis.

### Suggested handling

Fold it into the post-#1706 re-measure rather than treating it as a separate finding: run `mutalyzer_normalize_tests::axis_noncoding_idempotent` armed, read the 62 rows, and decide per row whether they are the arm working as intended. If they are, re-pin `NONCODING_NON_IDEMPOTENT` in the same change as the other re-pins and say so in the `Representation-Change:` trailer — this is the non-coding axis, so it is a distinct disclosure from the `c.`-axis payload-coincidence merges the trailer already covers.

Related, and the reason this is easy to miss: #1712 is currently bucketed as MENTIONED-ONLY across the open PR set — no PR claims to fix it — so there is nothing else in flight that would catch this on the way past.

Issue #1834, opened 2026-08-13
normalize and from_sequences run different block partitioners by construction, and nothing measures the gap


nh13:
## Summary

`Normalizer::normalize` and `from_sequences` derive a description through **different partitioners**, and the choice is not a shared setting — one is hard-wired and the other is defaulted. So a normalization improvement landing on one surface does not reach the other, and no test anywhere reports that they have drifted apart.

Measured over the descriptions harvested from the #1157 / #1229–#1235 / #1419–#1421 issue family (85 rows, one 139-base contig), on `origin/main` at `7cc9846c` with #1831 and #1832 fixed:

| direction | compared | disagree |
|---|---|---|
| `ThreePrime` | 79 | **32** |
| `FivePrime` | 79 | **36** |

"Compared" is rows where both surfaces answer and `from_sequences` reports `placement_bounded_by_window == false` — i.e. the derivation had at least as much room as the normalizer. Six rows are excluded because they are not drawn against this contig at all.

Examples (3'):

```
g.21_24delinsGCTG          normalize -> g.21_24delinsGCTG
                      from_sequences -> g.[21delinsGC;24del]
g.29_33delinsAACACATACTG   normalize -> g.29_33delinsAACACATACTG
                      from_sequences -> g.[28_29insAA;32G>A;33_34insACTG]
g.35_39delinsTA            normalize -> g.[35_37del;39T>A]
                      from_sequences -> g.[35C>T;37_39del]
```

## Mechanism

- `from_sequences` calls `merge::derive_block_members`, whose body calls `partition_block_canonical_within` **unconditionally**. It is pinned to the canonical (member-count-minimal minimal alignment) rule.
- `normalize` calls `merge::canonicalize_from_sequence`, which cuts with `PartitionRule`, whose default is `Live` — `partition_block`, a single-gap alignment search plus two narrow escapes.

So the divergence is by construction rather than by drift, and it is in the direction opposite to the one worth worrying about: **`from_sequences` is ahead**, re-deriving partitions that `normalize` leaves as a spanning `delins`. That is the #1419/#1420/#1421 shape.

Two consequences that are easy to miss:

[...]

PR #1837, opened 2026-08-13
fix: make the applier's overlap verdict order-independent, and let reanchor widen a pure-deletion window

### LINKED ISSUE #1831: apply_to_reference: the members-overlap verdict depends on the order members were written in (insertion flush against a sibling)
## Summary

`apply_to_reference` / `Normalizer::to_sequences` refuse a cis allele containing an insertion flush against a sibling that consumes the reference base immediately 3' of it — but **only when the insertion is written first**. Writing the same two members in the other order applies cleanly. Member order carries no meaning (#1098 fixed the normalizer's ordering for exactly that reason), so one of the two answers is wrong, and the refusing one contradicts the applier's own documented contract.

## Reproducer

Measured on `origin/main` at `7cc9846c`, against a synthetic 139-base contig served under `TEMPLATE` (positions 126-139 are `TTTAGGCGACATTT`, so 129=`A`, 130=`G`, 131=`G`).

| description | `to_sequences(v, 4)` |
|---|---|
| `TEMPLATE:g.[129_130insT;130G>A]` | **REFUSED** — "its members overlap, or a stated reference base disagrees with the reference" |
| `TEMPLATE:g.[130G>A;129_130insT]` | OK — `ref=TTTAGGCGA alt=TTTATAGCGA` |
| `TEMPLATE:g.[129_130insT;130del]` | **REFUSED** |
| `TEMPLATE:g.[130del;129_130insT]` | OK — `ref=TTTAGGCGA alt=TTTATGCGA` |
| `TEMPLATE:g.[128_129insT;129A>C]` | **REFUSED** |
| `TEMPLATE:g.[129_130insT;131G>C]` | OK (one base of clearance) |
| `TEMPLATE:g.[130G>A;131G>C]` | OK (control, two plain subs) |

## Cause

`apply_triples` (`src/spdi/apply.rs:452`) sorts with

```rust
ordered.sort_by_key(|t| std::cmp::Reverse(t.position));
```

`sort_by_key` is **stable**, so two triples that share an SPDI position keep the order they were authored in. An insertion at interbase *p* and a reference-consuming member starting at *p* share that position: the insertion has `deletion.len() == 0` (end *p*), the substitution/deletion has `deletion.len() >= 1` (end *p+1*).

`triples_are_disjoint` then walks 5'→3' carrying the furthest 3' reach so far and rejects when `reach > t.position`. Visiting the length-1 member first sets `reach = p + 1`, and the zero-length insertion at *p* is then read as an overlap. Visiting the insertion first sets `reach = p`, and `p > p` is false, so the pair passes.

This directly contradicts that function's own doc comment:

> Two triples that merely abut are disjoint, and so is any number of pure insertions at one interbase position — an insertion deletes nothing and therefore claims no base.

An insertion between 129 and 130 consumes neither base, so the two members do not claim the same base and the resulting sequence is well defined (`A`, `T`, then the substituted base) — which is what the accepted ordering already produces.

## Impact

Found while putting the descriptions in #1157/#1229–#1235/#1419–#1421 through `from_sequences` via `to_sequences`. It blocks material that is the point of those issues:

- **2 of the 4 alignment spellings in #1420 (comment 2)** cannot be applied at all — `g.[129_130insT;130G>A;131G>C;134_135insG;135_136insTT]` and `g.[129_130insT;130G>A;131G>C;134_135insGCT;135C>T]`. Those three rows are the only place in the corpus that varies the *alignment* rather than the spelling, so the shape most worth testing is the one that cannot be reached.
- 4 of 85 harvested rows overall. Confirmed as this defect rather than a base mismatch by re-authoring each with the insertion last:

| as authored | with the insertion moved last |
|---|---|
| `g.[4_5insA;5A>C;6C>A;7C>G;9G>T;10T>A]` REFUSED | `g.[5A>C;6C>A;7C>G;9G>T;10T>A;4_5insA]` OK |
| `g.[5_6insA;6C>A;8A>G;9G>A]` REFUSED | `g.[6C>A;8A>G;9G>A;5_6insA]` OK |

## Why no existing oracle catches it

`FERRO_ASSERT_SEQUENCE`'s own false-positive census already records "insertion flush against a deletion — 233 fires — `triples_are_disjoint` calls it an overlap; `cis_apply_oracle`'s tie-break does not, and it is well defined". That class was closed for the *oracle* by not consulting `triples_are_disjoint` there. The applier itself was left as-is, and the order-dependence was never measured — a refusal reads as a legitimate decline, and every corpus that feeds `to_sequences` counts it as a skip.

## Suggested resolution

Make the tie-break explicit so the ascending walk always sees zero-length deletions first:

```rust
ordered.sort_by_key(|t| (std::cmp::Reverse(t.position), std::cmp::Reverse(t.deletion.len())));
```

At a shared position this puts the longer deletion earlier in `ordered` (descending) and therefore later in the 5'→3' walk. The application loop then splices the reference-consuming member first and the insertion at the same offset second, which places the inserted payload 5' of the substituted base — matching what the currently-accepted ordering produces.

Two or more pure insertions at one interbase are unaffected: they tie on both keys, the stable sort preserves authored order, and the walk's `p > p` test still passes, so today's behaviour for that shape is unchanged.

## Stability cost

Strictly a widening: descriptions that are refused today begin to apply. No description that applies today changes what it denotes, since the ordering only becomes deterministic where it was previously authored-order-dependent, and the deterministic choice is the one already produced for the accepted spelling.

### LINKED ISSUE #1832: SequencePair::trim_to refuses a no-op trim when the alternate is empty, so Normalizer::reanchor can never widen a pure-deletion window
## Summary

`Normalizer::reanchor` is the documented way for a caller holding bases to widen its window. It cannot widen **any** pair whose alternate is empty — i.e. any pure deletion covering its whole window, which is exactly what `Normalizer::to_sequences(v, 0)` returns for one. Even the identity `reanchor(pair, start, end)` at the pair's own coordinates is refused.

## Reproducer

Measured on `origin/main` at `7cc9846c`. Contig `TEMPLATE` = `GGATTACAGGCAAAAGCCTGAGGATTACAGGCATTAGCCT`, with an `A` run at 12-15.

```rust
let deletion = SequencePair::new("TEMPLATE", 12, "AAAA", "")?;   // 12-15, alternate empty
```

| call | result |
|---|---|
| `reanchor(&deletion, Some(12), Some(15))` (identity) | **ERR** `trim_to(12, 15) on TEMPLATE would leave nothing of the 4-base reference or the 0-base alternate` |
| `reanchor(&deletion, Some(9), Some(15))` (widen 5') | **ERR**, same message |
| `reanchor(&deletion, Some(12), Some(20))` (widen 3') | **ERR**, same message |
| `reanchor(&deletion, Some(9), Some(20))` (widen both) | **ERR**, same message |
| control: `SequencePair::new("TEMPLATE", 11, "CAAAA", "C")`, `reanchor(9, 20)` | OK — `9-20 ref="GGCAAAAGCCTG" alt="GGCGCCTG"` |
| `from_sequences("TEMPLATE", 12, "AAAA", "", default)` | OK — `TEMPLATE:g.12_15del` |

So the pair is well-formed, `from_sequences` derives from it, and only `reanchor` refuses it.

## Cause

`Normalizer::reanchor` narrows to the intersection of the requested window and the pair's own before widening, and delegates that narrowing to `SequencePair::trim_to`. When the requested window contains the pair's own, the intersection **is** the pair's own window, so `head == tail == 0` — nothing is being trimmed. `trim_to`'s emptiness guard nonetheless fires:

```rust
if head + tail >= ref_len || head + tail >= alt_len {
```

With `alt_len == 0` this reads `0 >= 0`. The guard is there to stop a trim from consuming an entire string, and it is also load-bearing against a panic — the "matching bases only" check that follows indexes `a[..head]`, which would be out of range on an empty alternate for any `head > 0`. But neither concern applies when `head + tail == 0`.

## Secondary: the message names coordinates the caller never passed

`reanchor(9, 20)` reports `trim_to(12, 15)`. `reanchor` carries a comment explaining that it added its own non-overlap message precisely to avoid this:

> Delegating it to `trim_to` produced `trim_to(1000, 16) on chr1: start is past end` — a sibling method's name and two coordinates the caller never passed, for the mistake callers make most often here.

The emptiness refusal is a second path with the same problem.

## Impact

Any caller re-anchoring a window it obtained from `to_sequences(v, 0)` for a deletion — the natural way to get a minimal window — hits this on the first widen. It was found building a window-family confluence probe over the descriptions in #1157/#1229–#1235/#1419–#1421: 11 of 85 rows produced 64 refusals each until the probe was changed to request `to_sequences(v, 1)` instead, and the workaround is only available because that probe controls both ends of the call.

## Why no existing oracle catches it

`tests/it/from_sequences_reanchor.rs` builds every one of its pairs through `read(lo, hi)`, which always includes flanking matched bases, so the alternate is never empty in any existing test.

## Suggested resolution

Skip the guard when nothing is being removed:

```rust
if head + tail > 0 && (head + tail >= ref_len || head + tail >= alt_len) {
```

A narrowing request that would genuinely cut into an all-deleted block still refuses, and the panic the guard protects against is still unreachable, since it needs `head > 0`. Optionally, give `reanchor` its own message for the case as it already does for non-overlap.

## Stability cost

None to normalized output. It makes a currently-refused call succeed and changes no value any accepted call returns.

nh13:
Closes #1831. Closes #1832.

Representation-Change: none. 0 of 85,642 corpus rows move, and that zero is structural rather than lucky — see **Representation stability** below.

Two defects in the surfaces `from_sequences` is reached through. Both were found while enumerating window families over the descriptions reported in #1157 / #1229–#1235 / #1419–#1421, and both were blocking that material rather than showing up as a wrong answer, which is why neither had been noticed: each presents as a legitimate-looking refusal, and every corpus that feeds `to_sequences` counts a refusal as a skip.

## Defect 1 (#1831) — the applier's overlap verdict depended on member order

`apply_triples` sorted its triples with `sort_by_key(Reverse(t.position))`. That sort is **stable**, so two triples sharing an SPDI position kept the order the author happened to write their members in. An insertion at interbase `p` and a `sub`/`del` starting at `p` share that position, and `triples_are_disjoint`'s running-reach walk then read them as overlapping or not depending on which came first.

Measured on this branch's base, same contig, same pad:

|---|---|---|
| `g.[9_10insT;10G>A]` | REFUSED "members overlap" | `ref=CAGGCAA alt=CAGTACAA` |
| `g.[10G>A;9_10insT]` | `ref=CAGGCAA alt=CAGTACAA` | unchanged |
| `g.[9_10insT;10del]` | REFUSED | `ref=CAGGCAA alt=CAGTCAA` |
| `g.[10del;9_10insT]` | `ref=CAGGCAA alt=CAGTCAA` | unchanged |

The refusing branch contradicted this module's own contract, stated on `triples_are_disjoint`: "an insertion deletes nothing and therefore claims no base". It also contradicted #1098, which settled that member order carries no meaning.

**The fix is the sort key its own siblings already use.** There are three appliers in the tree and `apply_triples` was the only one missing it:

| applier | sort key |
|---|---|
| `spdi::apply::splice_denoted_sequence` | `(Reverse(position), Reverse(deletion.len()))` |
| `conformance::spec_corpus::apply_triples` | `(Reverse(position), Reverse(deletion.len()))` |
| `spdi::apply::apply_triples` | `Reverse(position)` — **the outlier** |

`splice_denoted_sequence`'s doc comment even states the reason verbatim: "**longer deletions first** among members at one position so a span-claiming member is always applied before the zero-length one flush against it". So this is not a new rule; it is the existing one applied at the third call site.

### Why it mattered

Two of the four alignment spellings in #1420 (comment 2) could not be applied at all. Those three rows are the only place in the reported corpus that varies the **alignment** rather than the spelling — one read, one reference, four gap placements — so the shape most worth testing was the one that could not be reached. All four now apply, and all four denote one sequence, pinned by `the_1420_alignment_rows_now_apply_and_denote_one_sequence`.

Across the 85 harvested rows this moves 4 from "cannot be applied" to "applies". Attribution confirmed rather than assumed, by re-authoring each with the insertion last:

| as authored | insertion moved last |
|---|---|
| `g.[4_5insA;5A>C;6C>A;7C>G;9G>T;10T>A]` REFUSED | `g.[5A>C;…;4_5insA]` OK |
| `g.[5_6insA;6C>A;8A>G;9G>A]` REFUSED | `g.[6C>A;…;5_6insA]` OK |

## Defect 2 (#1832) — `reanchor` could not widen a pure-deletion window

`Normalizer::reanchor` narrows to the intersection of the requested window and the pair's own before it widens, delegating that narrowing to `SequencePair::trim_to`. When the request *contains* the pair's window, the intersection is the pair's window and nothing is trimmed — but `trim_to`'s emptiness guard read `head + tail >= alt_len`, which is `0 >= 0` on an empty alternate, and refused.

An empty alternate is what `to_sequences(v, 0)` returns for **any pure deletion**, so the shape `reanchor` could not widen was the minimal window for the commonest edit a caller holds. Every call below was refused before, including the identity:

| call on `SequencePair::new("TEMPLATE", 12, "AAAA", "")` | before | after |
|---|---|---|
| `reanchor(12, 15)` (identity) | ERR | `ref=AAAA alt=` |
| `reanchor(9, 15)` | ERR | `ref=GGCAAAA alt=GGC` |
| `reanchor(12, 20)` | ERR | `ref=AAAAGCCTG alt=GCCTG` |
| `reanchor(9, 20)` | ERR | `ref=GGCAAAAGCCTG alt=GGCGCCTG` |

The fix is `head + tail > 0 &&` in front of the existing guard. The guard is still needed for `head + tail > 0`, and for **two** reasons rather than one — it stops a trim consuming a whole string, and it is what keeps the "matching bases only" comparison below it from indexing `a[..head]` past the end of an empty alternate. That second reason is why the guard is narrowed rather than removed.

#1832's **second** point is also fixed: `reanchor` delegated narrowing to `trim_to`, whose message names its own method and the *intersection* coordinates, so `reanchor(13, 20)` reported `trim_to(13, 15)` — an end the caller never passed. That is precisely the failure `reanchor`'s non-overlap branch already carries a bespoke message to avoid; the other exit now does too, keeping `trim_to`'s reason (the accurate half) wrapped in the caller's own request. Pinned by `a_narrowing_refusal_names_the_callers_window`.

The re-frame **matches on the error rather than wrapping its rendered form**, which a second review pass found necessary rather than tidy: `trim_to` returns the same `FerroError::InvalidCoordinates` variant and `Display` prepends the kind, so nesting the whole rendered error produced `Invalid coordinates: reanchor(…): … failed: Invalid coordinates: trim_to(…) on TEMPLATE`, printing both the kind and the accession twice. Taking the inner `msg` states each once; any other error kind is passed through unchanged rather than flattened into `InvalidCoordinates`, which would cost a caller the ability to match on it. The message now reads:

```
Invalid coordinates: reanchor(13, 20): narrowing to the overlap with the pair's own
window [12, 15] failed: trim_to(13, 15) on TEMPLATE would leave nothing of the
4-base reference or the 0-base alternate
```

## Unchanged on purpose

The discriminating cases, each of which a wider "fix" would break:

* `g.[9_11del;10G>A]` — a substitution interior to a deletion still refuses. A fix that deleted `triples_are_disjoint` instead of ordering its input passes everything else and fails this.
* `g.[9_10insT;9_10insA]` — two pure insertions at one interbase still refuse, and still with their own message (`no single resulting sequence`, from `variant_edit_triples_reason`) rather than the overlap one. They tie on both sort components, so the stable sort leaves them in authored order and this change cannot reach them.
* `g.[9_10insT;11C>A]` and `g.[10G>A;11C>A]` — byte-identical before and after.
* `reanchor(pair, 13, 15)` on the empty-alternate pair still refuses: `head = 1`, so there is genuinely nothing to trim.
* A pair carrying one matched base (`"CAAAA"` / `"C"`) widens exactly as before.

Of the 12 tests in the two new modules, **4 fail on this branch's base** — the rest are negative controls and message-accuracy pins.

Two assertion-strength points worth calling out, since both were self-review findings: `src/spdi/apply.rs` emits **disjunctive** refusal messages ("its members overlap, **or** a stated reference base disagrees"), so a bare `contains("members overlap")` cannot prove its own cause. `a_member_interior_to_a_deletion_still_refuses` therefore carries a positive control — the same members with the substitution moved outside the deletion's span apply cleanly — so the refusal is attributed to the overlap by elimination; and the coincident-insertion test additionally asserts the message is **not** the overlap one.

## Representation stability

`Representation-Change: none`, and the zero is structural rather than a corpus miss.

`examples/dump_normalized_corpus.rs` reports **0 of 85,642 rows moved** across all 21 shape families, `origin/main` @ `7cc9846c` against this branch. On its own that would be weak evidence, since this repo's doctrine is that a corpus zero is a claim about the corpus. The stronger argument is reachability, checked from the tree:

* `apply_triples` is called from `spdi::apply_to_reference`, `equivalence::checker` and `conformance::spec_corpus`. It is **not** on `Normalizer::normalize`'s path.
* `trim_to` / `reanchor` are called from `python.rs` and from `Normalizer::reanchor`. Also not on the normalize path.

So no normalized output can move, whatever the corpus contains.

**What can move, and is a widening rather than a re-spelling:** descriptions that `apply_to_reference` / `to_sequences` / `Normalizer::reanchor` previously **refused** now succeed. Nothing that succeeded before returns a different value — the ordering only becomes deterministic where it was previously authored-order-dependent, and the deterministic choice is the one already produced for the accepted spelling. `EquivalenceChecker` reaches `apply_triples`, so a pair where one side previously could not be applied may now upgrade its verdict; that is the same direction, and is #1158's complaint class.

## Coverage on both surfaces

`Normalizer.reanchor` and `SequencePair(accession, position, reference, alternate)` are both exposed to Python, so #1832's shape is reachable there and the binding is where the exception mapping can break — the narrowing refusal now travels through a new `map_err`, which no Rust test can see the far side of. `tests/python/test_from_sequences.py` gains two tests covering the widen and the refusal, the latter asserting `NormalizationError` specifically rather than a bare exception.

Verified red on the base by reverting **only** the `head + tail > 0` guard in place and rebuilding the extension — not by checking a file out of `origin/main`, which has moved twice under this branch and produces a build failure that lets `pytest` silently run against the previously-installed wheel.

## Verification

Run on the rebased tree, at `main` including #1806 and #1706:

```
cargo fmt --all --check                                     # 0
cargo clippy --all-features --all-targets -- -D warnings    # 0
scripts/run_oracle_suite.sh                                 # 10,505 passed, 0 failed
FERRO_SWEEP_SEEDS=full cargo nextest run --features dev     # 10,642 passed, 0 failed
cargo check --features python                               # 0
ruff check / ruff format --check / mypy                     # 0
uv run pytest tests/python/test_from_sequences.py           # 45 passed
```

Note `FERRO_ASSERT_IDEMPOTENT=1 cargo nextest run --features dev` is **not** the gate — that bare armed form is red on a clean `main` by construction, and `scripts/run_oracle_suite.sh` is the runner that mirrors CI's `test-oracle` selection.

## Follow-up, not in this PR

#1834 records what this work was actually looking for: `normalize` and `from_sequences` run **different block partitioners** by construction — `derive_block_members` calls `partition_block_canonical_within` unconditionally, while `canonicalize_from_sequence` defaults to `PartitionRule::Live` — and 32 of 79 rows (3'; 36 of 79 under 5') disagree between the two surfaces, with nothing in the suite reporting it. It also records that `FERRO_PARTITION` moves only the `normalize` arm, so a bake-off including a `from_sequences` figure compares a moved surface against a fixed one. Converging them is #1419–#1421's open product decision; the tripwire proposed there does not depend on it.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -562,7 +562,7 @@ which is what makes it an argument for the fix beside it:
 | class | fires | why it was not a defect |
 |---|---|---|
 | output cannot be transliterated | 328 | the input states its own deleted bases (`g.1000delC`) and converts with no provider; the output (`g.1000del`) must read the reference, which the fixture does not hold |
-| insertion flush against a deletion | 233 | `triples_are_disjoint` calls it an overlap; `cis_apply_oracle`'s tie-break does not, and it is well defined |
+| insertion flush against a deletion | 233 | `triples_are_disjoint` called it an overlap; `cis_apply_oracle`'s tie-break does not, and it is well defined. **No longer true of the applier as of #1831** — `apply_triples` now sorts longer deletions first among triples sharing a position, the key `splice_denoted_sequence` and `conformance::spec_corpus`'s applier already used, so `triples_are_disjoint` sees correctly-ordered input and no longer reports this shape. The 233 is the size of the class when it was diagnosed, and is kept as that |
 | overlap-conflicting input | ~12 | an insertion *interior* to a deletion — the input denotes nothing, so there is no baseline |
 | `pter`/`qter` | 6 | carry no numeric coordinate (`base: 0`); `hgvs_to_spdi` reads `base` and silently resolves `c.pterdel` to the sequence's **last** base |
 | corrected `REFSEQ_MISMATCH` | 5 | `c.10dupA` where the reference reads `C`: normalization is *supposed* to change the denoted sequence here |
--- a/python/ferro_hgvs/__init__.pyi
+++ b/python/ferro_hgvs/__init__.pyi
@@ -1508,6 +1508,11 @@ class SequencePair:
         the point when the bound is a requirement and a footgun when it is
         arbitrary. The derivation reports it as ``placement_bounded_by_window``.
 
+        A narrowing that removes nothing is always accepted, which
+        ``Normalizer.reanchor`` depends on: it narrows to the intersection
+        before it widens, so a request that contains this window narrows by a
+        no-op.
+
         Raises:
             NormalizationError: for a bound that would widen the window, cut
                 into a base the two sequences disagree on, empty the reference,
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -2237,7 +2237,10 @@ impl<P: ReferenceProvider> Normalizer<P> {
     /// * a bound the provider cannot serve;
     /// * anything [`crate::SequencePair::trim_to`] refuses, for the narrowing
     ///   side — a cut through a base the two sequences disagree on, an emptied
-    ///   reference, `start` past `end`.
+    ///   reference, `start` past `end`. The *conditions* are `trim_to`'s; the
+    ///   **message** is not, because that method names itself and the
+    ///   intersection coordinates rather than the ones the caller passed. Its
+    ///   reason is kept and wrapped in this method's own request (#1832).
     pub fn reanchor(
         &self,
         pair: &crate::normalize::sequence_pair::SequencePair,
@@ -2293,7 +2296,37 @@ impl<P: ReferenceProvider> Normalizer<P> {
                 pair.end()
             )));
         }
-        let narrowed = pair.trim_to(Some(overlap_start), Some(overlap_end))?;
+        // The same reasoning as the non-overlap branch above, applied to every
+        // other way the narrowing can fail: `trim_to`'s message names its own
+        // method and the *intersection* coordinates, so a `reanchor(13, 20)`
+        // came back as `trim_to(13, 15) …`, quoting an end the caller never
+        // passed. The underlying reason is kept — it is the accurate half — and
+        // the caller's own request is restored around it (#1832).
+        let narrowed = pair
+            .trim_to(Some(overlap_start), Some(overlap_end))
+            .map_err(|e| match e {
+                // Re-frame rather than nest. `trim_to` returns this same
+                // variant, and `FerroError`'s `Display` prepends the kind — so
+                // wrapping the whole rendered error produced "Invalid
+                // coordinates: reanchor(…): … failed: Invalid coordinates:
+                // trim_to(…)", with the kind and the accession each printed
+                // twice. Taking the inner `msg` keeps the reason, which is the
+                // accurate half, and states it once.
+                //
+                // The accession is deliberately not repeated here either: the
+                // reason carries it. The non-overlap branch above does name it,
+                // because that message stands alone and nests nothing.
+                FerroError::InvalidCoordinates { msg } => invalid(format!(
+                    "reanchor({start}, {end}): narrowing to the overlap with the pair's own \
+                     window [{}, {}] failed: {msg}",
+                    pair.position,
+                    pair.end()
+                )),
+                // Any other kind is passed through unchanged rather than
+                // flattened into `InvalidCoordinates`, which would lose the
+                // caller's ability to match on it.
+                other => other,
+            })?;
 
         // Both edges are read before either string moves — `end()` is derived
         // from `reference`, so consuming it first would make the 3' test consult
--- a/src/normalize/sequence_pair.rs
+++ b/src/normalize/sequence_pair.rs
@@ -168,6 +168,12 @@ impl SequencePair {
     ///   alternate is ordinary input, and the comparison folds;
     /// * a bound that would leave the reference empty, or `start` past `end`.
     ///
+    /// A narrowing that removes **nothing** is accepted unconditionally, and
+    /// that is load-bearing rather than incidental: [`crate::Normalizer::reanchor`]
+    /// narrows to the intersection before it widens, so for a request that
+    /// *contains* this window the narrowing is exactly the no-op. Refusing it
+    /// made `reanchor` unable to widen any pure deletion (#1832).
+    ///
     /// Case is otherwise preserved: this method fetches nothing, so it cannot
     /// manufacture a mixed-case pair and has no reason to rewrite the caller's
     /// bases. [`crate::Normalizer::reanchor`], which does fetch, folds instead.
@@ -202,7 +208,23 @@ impl SequencePair {
         // The two sequences differ in length, so a 3' trim is counted from each
         // one's own end rather than from a shared index.
         let (ref_len, alt_len) = (self.reference.len(), self.alternate.len());
-        if head + tail >= ref_len || head + tail >= alt_len {
+        // `head + tail > 0` first, because the rest of this guard is about what
+        // a trim would *remove* and a no-op trim removes nothing.
+        //
+        // Without it, `alt_len == 0` makes the test read `0 >= 0` and refuses a
+        // request that was going to return `self` unchanged. That is not a
+        // corner case: `Normalizer::to_sequences(v, 0)` returns exactly that
+        // shape for any pure deletion, and `Normalizer::reanchor` narrows to the
+        // intersection before it widens — so a *widen* of such a pair arrived
+        // here as a no-op narrowing and was refused, making `reanchor`
+        // unreachable for the commonest window a caller holds (#1832). Even
+        // `reanchor(pair, pair.position, pair.end())` failed.
+        //
+        // The guard is still needed for `head + tail > 0`, and for two reasons
+        // rather than one: it stops a trim consuming an entire string, and it is
+        // what keeps the "matching bases only" comparison below from indexing
+        // `a[..head]` past the end of an empty alternate.
+        if head + tail > 0 && (head + tail >= ref_len || head + tail >= alt_len) {
             return Err(invalid(format!(
                 "trim_to({start}, {end}) on {} would leave nothing of the {ref_len}-base \
                  reference or the {alt_len}-base alternate",
--- a/src/python.rs
+++ b/src/python.rs
@@ -1355,6 +1355,10 @@ impl PySequencePair {
     /// point when the bound is a requirement and a footgun when it is
     /// arbitrary. The derivation reports it as `placement_bounded_by_window`.
     ///
+    /// A narrowing that removes nothing is always accepted, which
+    /// `Normalizer.reanchor` depends on: it narrows to the intersection before
+    /// it widens, so a request that contains this window narrows by a no-op.
+    ///
     /// Raises:
     ///     NormalizationError: for a bound that would widen the window, cut into
     ///         a base the two sequences disagree on, empty the reference, or put
--- a/src/spdi/apply.rs
+++ b/src/spdi/apply.rs
@@ -435,7 +435,16 @@ fn trim_common_flanks<'a>(reference: &'a [u8], resulting: &'a [u8]) -> (usize, &
 /// interval that begins at `win_start` — and return the edited sequence.
 ///
 /// Triples are applied from the 3' end (descending position) so that an earlier
-/// splice never shifts the coordinates of a later one. Each triple's stated
+/// splice never shifts the coordinates of a later one, and — among triples that
+/// share a position — **longer deletions first**, so a span-claiming member is
+/// always applied before the zero-length one flush against it. The second
+/// component is not cosmetic: [`triples_are_disjoint`] walks this slice in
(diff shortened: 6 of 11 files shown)


nh13:
@coderabbitai pause

PR #1838, opened 2026-08-13
docs(rulings): scope two ruling claims to what was actually measured

nh13:
Two rationale-only corrections to decided records, in one PR because they touch the same file.

Representation-Change: none
Measured, not assumed: the ledger is read only by `generate_spec_fixture` at generation time, and rendered into the tracked `docs/NORMALIZATION_CONTRACT.md` by `normalization_contract_doc` (#1850). No `include_str!`/`include_bytes!` under `src/` names it, so it is not compiled into the library and no emitted string can depend on it. No record added or removed, no status changed, no clause citation added or removed: 28 records, 25 decided / 3 undecided, `multi.len() == 30`, `mixed.len() == 3` — all identical to `origin/main`, so the generated clause index does not move and `tests/it/clause_ruling_index.rs` is untouched. Verified locally on the pushed head: `normalization_contract_doc`, `hgvs_spec_normalization_tests` (including `ruling_records_are_intact` and `status_census_is_unchanged`), `clause_ruling_index`, `spec_enumeration_tests`, `claude_md_adjudication_tables` and `idempotency_tests` — 44 run, 44 passed, with both generated spec fixtures regenerated first.

## 1. `delins-adjacent-members-when-both-consume-reference` — scope a figure to what it measured

The post-#1537 re-measurement reads *"and ZERO fall in this ruling's scope"*. That zero was measured on the designed cis corpus in the 3' direction, but the sentence reads as a global claim, and it has already been carried onto a case it does not cover.

The scope sentence, the status, the governing clause and the clause list are all unchanged. Only the figure is scoped, and the contrary measurement added.

Re-measured on the representation-stability corpus (`examples/dump_normalized_corpus`, 85,642 rows, 78,298 on nucleotide axes): **eight** separation-0 pairs in which both members consume reference bases — this record's own in-scope shape. Four distinct inputs, each in both shuffle directions, all `cx`, all `delins_hiding_an_inversion`:

```
NM_TESTX.1:c.[-1_2delinsGGG;3_5delinsAGG] -> c.[1_2delinsGG;3_5inv]   delins | inv
NM_TESTX.1:c.[-1_2delinsGGG;3_5delinsGAG] -> c.[1_2delinsGG;3_5inv]   delins | inv
NM_TESTX.1:c.[-1_2delinsGGG;3_5delinsCTC] -> c.[2A>G;3_5inv]          sub    | inv
NM_TESTX.1:c.[-1_2delinsGGG;3_5delinsATG] -> c.[1_2inv;3_5inv]        inv    | inv
```

**All eight carry an `inv`** — six on exactly one side, two on both — so excluding `inv` from the scope would take this population to **zero, not to two**. That correction is written into the record because the figure has circulated as "six of the eight", which invites the arithmetic that two survive.

**Eight is a floor, not the count.** The classifier producing it reads members spelled entirely in plain positive coordinates. Counting every pair whose adjacency junction is itself between two plain positive coordinates, **and in which neither member carries an intron offset anywhere**, gives **332 rows over 39 distinct outputs**, every one with an `inv` on a side. Admitting UTR junctions as flat transcript offsets, under the same intron-offset exclusion, gives 719 rows over 71 distinct outputs.

This does not narrow the record's scope and is not an argument for narrowing it. It records that the population the scope sentence describes is inhabited.

### Two restatements inside that record, corrected here

Both are the drift class this PR exists to close — a rule or a name written down beside its numbers, and then outliving them. **No measured figure moves.**

**The classifier's rule disagreed with its own counts by one row.** Stated as "every pair whose adjacency junction is between two plain positive coordinates", read literally, it admits one further pair — `NM_TESTX.1:c.[7_8del;10A>C] -> c.[10A>C;11_11+1del]`, plain at the junction and intronic at the far end — and yields 333 rows, not the 332 the sentence carries. Both stated counts, 332/39 and 719/71, are counts under a rule that rejects an intron offset **anywhere** in either member. The fix is that clause on the rule statement, and nothing else. The clause is load-bearing rather than pedantic: the surplus row is also the only member of that population carrying no `inv`, so omitting it falsifies the "every one with an `inv`" claim standing beside it.

**"The shipped `live` arm" is no longer true.** `DEFAULT_PARTITION_RULE` is `CanonicalCoalesced` as of #1835 (`750f2334`), the commit this branch is rebased onto. The record now names the shipped arm and records that re-measurement at that commit finds `live` emitting the **identical set, row for row** — no row on one arm and not the other, and the same split of `inv` across sides. The correction is therefore stronger than the sentence it replaces, not weaker: the finding is a property of the derivation and holds on whichever arm ships.

## 2. `inversion-vs-two-delins-76-83` — narrow a claim that is false of the shipped tree

The record said *"The gate is `no_piece_is_a_lone_substitution` — a whole-block reverse complement is typed `inv` when no competing piece is a lone substitution."*

`whole_block_inversion` admits a block when **any** of its routes accepts it, and as shipped there are **three**: `changed_columns_dominate_the_span` (#1485), `no_piece_is_a_lone_substitution` (#1535, this record's) and `payload_columns_dominate_the_span` (#1640, added *after* this ruling). They are a disjunction, so a block whose every competing piece is a lone substitution is still typed `inv` when another route admits it, and this record's predicate is never consulted. A fourth, `every_separation_is_a_single_base` (#1341), was removed from that gate by #1787 as subsumed by `payload_columns_dominate_the_span`; it survives in `RunScan::window_inversion`, which is why the name is still live in the tree and reads as though it were still a block-level route. The record now instructs the reader to take the disjunction off `whole_block_inversion` rather than off any transcribed list.

Measured on real GRCh38 bases, genomic axis, `live` at `0dcf76ab`: over **1,937** minimal whole-block reverse-complement candidates from 40 kb of `NC_000017.11` whose competing pieces are all lone substitutions, **329 are typed `inv`** and 1,608 split. The outcome is predicted on **1,937 of 1,937** by `every_separation_is_a_single_base || changed_columns_dominate_the_span`. `0dcf76ab` predates #1787, so that measurement stands as taken and #1787 measured the removal as verdict-preserving; what does not carry forward is the predictor's spelling. All 329 are the alternating `X.X.X` shape at odd span (150 at 5, 107 at 9, 72 at 13).

The record is narrowed to the two-or-more-`delins` competitor it was decided about, and states that the all-lone-substitution case is **outside it and undecided** — neither licensed nor condemned.

**Guards.** Both of the record's guards stay live and neither goes vacuous: `issue_1517_inv_priority_over_delins::a_reverse_complement_competing_with_delins_members_is_an_inversion` and `case_harvest::the_inversion_preference_control_still_holds` are inside the narrowed scope. `a_reverse_complement_competing_with_substitutions_stays_split` still passes and still guards #1230; after the narrowing it is #1230's control rather than this record's, which is stated rather than left to be inferred.

**Also recorded: why the inversion-sweep corpus cannot see this.** `tests/fixtures/inversion-sweep` authors only even span lengths. The changed-column pattern of a whole-block reverse complement is symmetric about the block centre, so trimming matched flanks removes the same count from each end and the trimmed block keeps the authored span's parity — verified over the whole sweep, 0 exceptions. An even-only corpus cannot produce an odd block, so the `X.X.X` shape carrying this entire population is structurally unreachable in it.

## 3. `docs/NORMALIZATION_CONTRACT.md` re-blessed

The contract document is a tracked, generated render of the ledger, added by #1850 — one of the commits this branch was replayed onto. Any ledger prose edit makes it stale, and `normalization_contract_doc::the_published_document_is_current` was the sole CI failure on the previous head (`Test (3/6)`: 1,771 passed, 1 failed). It is re-blessed here against the final wording via `BLESS_CONTRACT_DOC=1`, and the test passes without the env var. The render is generated, not authored.


nh13:
@coderabbitai pause

nh13:
Independently re-derived the three population figures in the `EIGHT IS A FLOOR, NOT THE COUNT` paragraph, against head `867445c5`: own detached worktree at `origin/main` (`64fc76df`), own classifier written from the record's predicate rather than ported from the measurement, `FERRO_PARTITION=live`, 85,642 rows / 78,298 on nucleotide axes.

**Every number in the paragraph reproduces exactly, including "688 over 60".** The `8 → 0` correction is right and is the reason to land this. One sentence needs a qualifier; nothing else.

## What reproduces

| as written | measured |
|---|---|
| EIGHT separation-0 in-scope pairs; four inputs × two directions; `cx`; `delins_hiding_an_inversion` | 8 rows, 3 distinct outputs, `cx`, that family |
| ALL EIGHT CARRY AN `inv` → narrowing gives **0, not 2** | 8 of 8 |
| 332 rows over 39 distinct outputs | reproduces — under the rule below |
| 719 rows over 71 distinct outputs, **688 over 60** with an `inv` | reproduces exactly, output count included |

## The one fix: the stated classifier admits one more row than the numbers do

The junction sentence states its rule as *"every pair whose adjacency junction is itself between two plain positive coordinates"* and then reports *"332 rows over 39 distinct outputs, **every one with an `inv` on a side**"*.

Applied literally, that rule yields **333 rows over 40 outputs**, and the extra row carries no `inv`:

```
family del_plus_sub, axis cx
  NM_TESTX.1:c.[7_8del;10A>C]  ->  NM_TESTX.1:c.[10A>C;11_11+1del]
  junction 10 / 11 — both plain positive.   members: 10A>C (sub) // 11_11+1del (del)
```

Both members consume reference, the junction coordinates are plain positive, separation is 0. What excludes it is the **far** end of the second member — `11+1`, an intron offset. So the classifier that produced 332 rejects intron offsets *anywhere in the member*, not only at the junction. Measuring both rules:

| rule | rows / outputs | with `inv` | without |
|---|---|---|---|
| junction coordinates plain positive (as stated) | 333 / 40 | 332 | **1** |
| …and neither member carries an intron offset | **332 / 39** | 332 | 0 |
| UTR junctions as flat offsets (as stated) | 720 / 72 | 688 / 60 | 32 |
| …and neither member carries an intron offset | **719 / 71** | **688 / 60** | 31 |

Both of the PR's counts are the intron-rejecting rule, and both are correct under it — the surplus is the same single row in each case. Only the *description* of the rule is one qualifier short, and that qualifier is exactly what makes "every one with an `inv` on a side" true.

**Suggested minimal edit**, one clause:

> Counting every pair whose adjacency junction is itself between two plain positive coordinates **and neither member carries an intron offset** — …

That leaves 332/39, 719/71 and the all-`inv` claim all correct as written.

## One thing to keep apart, for the next reader

"every one with an `inv` on a side" is true of the 332 **only** under that rule, and it is **not** true of the flat population. The PR already says so — "688 over 60" — and that is worth protecting, because the claim has been circulating in the reverse form, i.e. as though the all-`inv` property extended to the 719. It does not: **32 of 720 (31 of 719) pairs carry no `inv` at all** — 29 `del|sub` at `*1`/`*2` and `c.-1`/`c.1` junctions, 2 `delins|delins`, 1 `sub|del`, over 12 distinct outputs in the `del_plus_sub` and `separated_revcomp_runs` families:

```
NM_TEST.1:c.[14_15del;*1T>A]                 del|sub        junction c.15 -> c.*1  (CDS end)
NM_TEST.1:c.[15_*1del;*2T>A]                 del|sub        junction *1 -> *2
NM_TESTX.1:c.[-2_-1del;1A>C]                 del|sub        junction c.-1 -> c.1   (CDS start)
NM_TESTX.1:c.[-3_-2delinsTG;-1_1delinsGT]    delins|delins  junction -2 -> -1
NM_TESTX.1:c.[10A>C;11_11+1del]              sub|del        junction 10 -> 11
```

That population is also why narrowing the scope sentence to exclude `inv` would not empty the record: it takes the strict eight to 0 but leaves 31–32 standing. The underlying defect — the record violated on the shipped `live` arm — is now filed as #1864, which also records why narrowing is not the response.

Coverage note, since it bears on whether these are counts or floors: 2,367 nucleotide-axis outputs carry a nested bracket payload (`ins[…]`, repeat notation) and my classifier refuses them rather than guessing. Re-parsed with depth-aware member typing: zero unparsed, **zero additional in-scope pairs**. So the figures are exact for this corpus and arm.

PR #1839, opened 2026-08-13
docs(rulings): type a whole-span reverse complement as inv, uniformly

nh13:
One operator ruling, recorded in the ledger. **This PR ships no normalizer change** — it records a decision and reconciles the records the decision supersedes.

Refs #1703. Refs #1541. Refs #1575.

Deliberately `Refs`, not `Closes`: none of the three is implemented here, and auto-closing them on merge would bury the implementation. That is the precedent merged #1684 set and stated in its own words — "auto-closing the issue on merge would bury the half that remains".

## The ruling

**`whole-span-reverse-complement-types-as-inv`** — `decided`, 2026-08-13, governing clause `DNA/inversion.md:5`.

> Type a whole-span reverse complement as `inv` uniformly — regardless of how much of the interior coincides with the reference, and regardless of whether the competing partition is substitutions, `delins`, or anything else.

The test is `block == revcomp(block)` over the trimmed block, evaluated on the resulting sequence, settled before any cut is considered.

### What it overturns

The ledger currently keys this on the **type of the competing partition**, and that distinction is retired:

| | competitor | old answer | new answer |
|---|---|---|---|
| `inversion-vs-two-delins-76-83` (`decided`) | `delins` | `inv` | `inv` — unchanged |
| **#1230** (CLOSED/COMPLETED) | substitutions | split | **`inv`** |

`inversion-vs-two-delins-76-83`'s **outcome stands** and `c.76_83inv` does not move; what is superseded is its **reason**, and it now carries a supersession paragraph naming which of its own arguments are no longer live rather than being left asserting them. Its base derivations, its `inversion.md:33-34` finding and its mean-versus-distribution refutation of the coincidence-density argument all survive.

### The load-bearing ground

Both halves of the old rule were arguments about `general.md:56` — #1230 applied it, `inversion-vs-two-delins-76-83` declined it because `delins` is absent from its list. But `adjudication-precedence-order` is **already `decided`** and its escalation E1 says of `:56` that it

> ranks single-variant TYPE LABELS FOR ONE SPAN — it never ranks a multi-member allele against a spanning description […] so `:56` cannot settle a merge-versus-split question at all

Whole-span `inv` against a multi-member partition *is* a merge-versus-split question, so `:56` reaches neither side, and a rule whose entire content is which side `:56` happens to reach cannot survive that. This is not a new reading of `:56`; it is the reading the ledger already carries, applied to the one place still keyed on the old one.

### It is a rule 6 choice, not a rule 1 correction

Stated plainly in the record because it is the reason there is a choice at all. Per #1703's 2026-08-12 reclassification, `inversion.md:5` makes a span *qualify* as an inversion and no clause *compels* it to be described as one — every imperative mention of `inv` is a prohibition or a redirect, and `delins.md:5`'s definitional exclusion reaches a `delins`, not a multi-member allele. So the split forms are conformant and this is `README.md` rule 6, "among multiple conformant forms: the maintainers choose", with rule 7 disclosure attached.

## Two corrections to claims made *about* this ruling — read these before sizing or citing it

Neither is a correction to the ruling. Both are claims that were relayed to this work as established, were checked, and did not hold. A ruling record whose stated cost is wrong is worse than one that admits the cost is unmeasured, so both are corrected in the record itself rather than repeated.

### 1. The 457 nt row is **not** the rule 1 violation #1703's comment claims

#1703's 2026-08-12 reclassification comment ends with a section headed "One genuine rule 1 violation does live in this family", attributing `case_harvest::no_emitted_output_puts_two_members_on_consecutive_nucleotides`'s 12 violations on `NC_000013.10:g.100809575_100810031inv` to `DNA/substitution.md:32`, which "marks the split spelling of consecutive changes `class="invalid"` … **by name**". That attribution is wrong, and because it is the *last* word #1703 says on the subject it will be re-derived by the next reader unless it is contradicted in as many words.

Re-run at `origin/main` `7cc9846c` under `FERRO_PARTITION=canonical-coalesced` (`GUARD_EXIT=100`, read back from the log):

- **The guard does fail, with exactly 12 pairs**, at the coordinates the comment lists. The count is right.
- **All 12 involve an `ins` member**, which consumes no reference — its two coordinates flank an insertion point rather than occupying bases. The two "same-coordinate" pairs in the comment's list (`…577/…577`, `…962/…962`) are two `ins` sharing a junction.
- **Zero are two substitutions.** The minimum gap between two reference-consuming members is **1**, so `substitution.md:32`'s named invalid shape `c.[79G>T;80C>T]` does not occur anywhere in the row.
- Instrument controlled against `:32`'s own invalid example, which it scores as one adjacent substitution pair.

So whether this is a rule 1 violation turns *entirely* on whether `member_bounds` is right to treat an `ins` junction as occupying its flanking coordinates — a reading the guard **encodes rather than argues**, and the same instrument objection already raised against it in the #1542 escalation ("the guard's own instrument already encodes one of the competing readings"). **This ruling does not decide that and does not need to** — it stands on rule 6 alone. Fixing the typing still closes the guard failure, because the row returns to a single `inv` member.

**Independently corroborated:** #1706's record, whose central example *is* this row, grounds on rule 6 and warns a reader not to "cite it as CONFORMANCE (it is not — the competing form violates nothing)". Two sessions reached that conclusion separately.

### 2. The "one corpus row moved" costing is stale **in both directions** — do not size this change off it

`inversion-vs-two-delins-76-83` costed "type it `inv` unconditionally" at **"moves ONE corpus row"**, and separately reported the shipped gated implementation's blast radius as a **structural zero** because "the only family that builds a reverse-complement block emits its pieces at separation 0". Both statements have since gone stale, in opposite directions, and neither may be quoted as current:

- `examples/dump_normalized_corpus.rs` has gained **`separated_revcomp_runs`**, which builds reverse-complement blocks at separations 0, 1, 2, 4 and 8 — so the corpus is **no longer blind to separation**, and the structural-zero justification no longer holds.
- It is **still blind to run width**. `REVCOMP_RUN_WIDTH` is a constant **`2`**, and the family's own doc comment says the one-column run was excluded deliberately ("A one-column run is a substitution, and the separation rules for substitutions are already dense in this corpus"). So **no design in that family can present the lone-substitution competitor** — the exact geometry the overturned gate keys on.
- The family that **can** is **`long_block_inversion`**. Re-derived here: `near_palindromic_core` perturbs two positions of an exact palindrome, and each perturbation costs two differences under reverse complement, so the block changes at exactly four columns — 0-based `{10, 504, 519, 1013}` at length 1024 and `{10, 542, 557, 1089}` at 1100, **every run one column wide**. That is the lone-substitution competitor at kilobase scale.

Per R6, the property this ruling keys on **is** variable in the corpus and the rows to measure are named. **No A/B was run**, because measuring the movement requires the normalizer change this PR deliberately does not make. The corpus figure is **owed by the implementing change**, and this record must not be cited as having supplied it.

## Reconciliation required at merge — please read before merging either PR

**Open PR #1706 carries a second inversion-typing ruling that is not on `main`** (`inversion-vs-a-mixed-member-competitor`, `decided` 2026-08-12), and edits this PR's files. **The relationship is subsumption, not conflict** — worked out by reading #1706's record at `refs/pull/1706/head`, and stated in this record's rationale so whoever merges second is reconciling a stated relationship rather than discovering one:

- That record types the `inv` where the competitor **mixes** lone substitutions with multi-column members, and says in terms that it disturbs neither `inversion-vs-two-delins-76-83` nor #1230's all-substitutions case, "which BOTH gates refuse". This ruling reaches every one of those populations **and the one they exclude**, so it agrees everywhere it speaks and closes the gap it leaves.
- Consequences the merge owes: the narrower record must not be read as a live carve-out once both are in the ledger; its clause (4) preserving #1230's split is superseded exactly as its sibling's equivalent sentence is; and its gate widening (`no_piece_is_a_lone_substitution` → `not_every_piece_is_a_lone_substitution`) is **retired** rather than contradicted — the uniform rule needs no piece-type predicate at all. #1706's post-hoc run scan is a separate mechanism and is not adjudicated here.
- **Three independent convergences** are recorded because two sessions reached them without seeing each other: both rule as **rule 6** (#1706 warns a reader not to "cite it as CONFORMANCE"); both decline `general.md:56`, grounding on two *different* decided records (`conflicting-member-geometry-refusal-scope` and `adjudication-precedence-order` E1) that say the same thing; and both measure `NC_000013.10:g.100809575_100810031inv` at **133 pieces whose first is `g.100809575A>T`**, on separate instruments.
- The strongest statement of why the distinction had to go is **#1706's, not this record's**, and the record says so: refusing the `inv` makes notation turn on whether one interior column happens to be self-complementary, "a property of the SEQUENCE, not of the EVENT". That instability is *worst* in the all-substitutions case, where a single coincidental column is the whole difference between one `inv` and a multi-member allele.

`tests/it/clause_ruling_index.rs`'s generated block will conflict textually whichever lands second. It must be **regenerated, never hand-merged** — run `the_rendered_index_is_current`, paste the block it prints, and take the census tuple from the regenerated header rather than from arithmetic.

## Everything cited was re-derived, not carried

| claim | how it was checked |
|---|---|
| `NM_004006.2:c.4145_4160` = `TCCAGGAGTCCCTCAC`, revcomp `GTGAGGGACTCCTGGA`, 10/16 columns changed, two three-base unchanged interior runs | read off **`NM_004006.2` itself** (prepared reference `supplemental/patterns_transcripts.fna`), not the `LRG_199t1` proxy the earlier record used — the two confirmed **byte-identical at 13,993 nt** — with CDS start **derived** as the longest ORF (tx 245, `c.1_3 == ATG`, 3,685 aa) rather than assumed |
| all 8 clause quotes | verbatim substrings of the cited lines in the pinned submodule; `generate_spec_fixture` re-verifies each and builds clean |
| #1541: `c.710_713` = `TCGT`, revcomp `ACGA`, `[710T>A;713T>A]` → exactly `ACGA` | re-derived from `NM_000500.9`, CDS start derived from the longest ORF |
| #1541's cost | **the issue's text is stale** — it says both spellings are fixed points; all three now converge on the split `c.[710T>A;713T>A]` (`inv` and `delinsACGA` both report `changed = true`). The ruling moves that converged form to `c.710_713inv` |
| #1575: at @7cc9846c, 155 repartitioned = 87 substitutions-only + 43 `inv`-core + **25 `delins`-bearing**, of 2,075 — **superseded as a current figure**, see below | re-derived from the committed pin `tests/fixtures/inversion-sweep/cases.tsv`, not from the issue |
| Mutalyzer | **14 of 14 probed live**, chosen to span *both* competitor types rather than to re-confirm the old rule: substitution competitors (#1541's row, #1703's row, four sweep pairs) and `delins` competitors all normalize to the whole-span `inv`. Corroboration only — `adjudication-precedence-order` records why Mutalyzer is not an authority. The earlier record's "26 of 26" is left where it is and **not** restated as current |

### The cost is wider than #1575 asked about

The ruling reaches **every** repartitioned sweep row, not only the 25 `delins`-bearing ones #1575 costed. **Do not quote 155 as current**: that was the population before #1706 (`64fc76df`) added its post-hoc inversion run scan, which absorbed 62 of them (all 43 `inv`-core rows and 19 of the 25 `delins`-bearing), leaving 93 on the `live` arm; #1835's flip of the default partition rule to `canonical-coalesced` then took one more, so `CENSUS_REPARTITIONED` pins **92**. Every remaining row is the all-substitutions case — exactly what this ruling overturns, and exactly what #1575 did not ask about. Read the count off the constant, not off prose that has been stale against it twice already.

The two corrections above (the 457 nt row's rule 1 attribution, and the stale corpus costing) are recorded in the ruling's own rationale, not only here.

The record also carries the counterweight **unanswered**: `general.md:56` does rank substitution above inversion, and retiring it here is a *scope* argument rather than a refutation. And the rule supplies no length or density bound — a 457 nt span with four changed columns is typed exactly as a 4 nt one.

## Ledger integrity

`CLAUDE.md`'s force-push set-comparison, run against `origin/main` before pushing:

```
origin/main records: 27  decided: 24  undecided: 3
HEAD                : 28  decided: 25  undecided: 3
dropped from main   : NONE
added               : ['whole-span-reverse-complement-types-as-inv']
status changes vs main: NONE
```

**Three** census pins moved, not two — `clause_ruling_index.rs` carries a second one that is easy to miss:

- `the_index_is_not_vacuous`: `(27, 24, 3)` → `(28, 25, 3)`
- `clauses_named_by_several_records_are_flagged`: multiply-named clause lines `28` → `31` (mixed-verdict stays `3`)
- `RULING_STATUSES` gains one entry; the generated index block regenerated to 113 clause lines / 31 multiply-named; `CLAUDE.md`'s decided-rulings table gains a row (guarded to partition the ledger by status — verified 3 open / 25 decided, no diff either way)

Every number was taken from the **regenerated index header** or from the test's own failure message, never from arithmetic, and the multiply-named/mixed counts were independently recomputed from the ledger before the pin was touched.

### Note for the next person adding a ruling record: `clause_ruling_index.rs` has **three** pins, not two

The two obvious ones are named in the module docs and move with the record count. The third is not, and it is the one that bites:

| pin | moves when |
|---|---|
| `the_index_is_not_vacuous` | the record count or a status changes |
| `ruling_records_are_intact` (`RULING_STATUSES`, in `hgvs_spec_normalization_tests.rs`) | a record is added, dropped, or re-statused |
| **`clauses_named_by_several_records_are_flagged`** | **any new record cites a clause line an existing record already cites** — even with no status change and no new clause line |

That last one is invisible from the record count: a record can add zero clause lines to the index and still move it, because the count is of *lines named by more than one record*. Adding eight clauses here moved it 28 → 31 while `the_index_is_not_vacuous` moved for an unrelated reason. Its sibling `mixed.len()` pin (mixed-verdict lines) moves only when the two records on a line disagree in status, so a `decided` record added beside `decided` ones leaves it alone — it stayed at 3 here.

Recompute rather than reason about it: the regenerated index header line states both totals (`113 clause lines, of which 31 are named by more than one record`), and the failing assertion prints the measured value.

**The second pin was found only by reading a log back.** The gate's completion notification reported **exit 0** while its own log line read `SUITE_EXIT=100` — one failing test, `clauses_named_by_several_records_are_flagged`, `left: 31, right: 28`. Believing the notification would have shipped a red branch. That is the tenth such misreport in this campaign.

## Doc-drift swept to class exhaustion

Five `.rs` prose sites asserted the now-overturned reason ("#1230's substitution case is untouched and still splits"; "the ranking discriminates on *what the block competes with*"). This is exactly the drift class the repo names as its recurring failure, and `ruling_citation_currency` cannot catch this variety — it checks status words, not reasoning. All five updated together, plus one assertion message that would have actively misdirected #1541's implementer:

- `tests/it/inversion_sweep.rs` — module doc and `CENSUS_REPARTITIONED`
- `src/conformance/inversion_sweep.rs` — `Repartitioned` doc
- `tests/it/case_harvest.rs` — `CYP21A2_TARGET` doc, `the_inversion_preference_control_still_holds` doc **and its assertion message**
- `tests/it/spec_worked_example_rules.rs` — W15.1's `why`

No pinned value or expected string was changed: the ruling ships no code, so every pin must stay green until #1703/#1541/#1575 implement it. Each site now says which change owes the re-bless.

**Deliberately not touched:** the `note` fields in `tests/fixtures/case-harvest/cases.json` carry the same stale reasoning. `.coderabbit.yaml` and `ruling_citation_currency`'s own module doc both record that editing those blessed conformance inputs is a separate reviewable change; flagged here rather than swept in.

## Verification

Full gate re-run on the final `tree`, every status **read back from its own log file** rather than from a completion notification — the notification was wrong twice in this session alone (`GUARD_EXIT=100` and the fixture generator both reported as exit 0).

## Review

`/coderabbitai-review` run against the local diff pre-push. The doc-drift class sweep above **is** its principal finding, applied in full. No finding was rejected.

Representation-Change: 0 rows move in this PR. The normalizer is untouched, the corpus is untouched, and every pinned string and census constant is unchanged — including `CYP21A2_TARGET` and the inversion sweep's `cases.tsv`, both of which this ruling decides *against* and which are deliberately left green until the implementation moves them. What is declared is the DECISION, and it is a real one: a whole-span reverse complement will be typed `inv` uniformly, so #1541's `NM_000500.9:c.[710T>A;713T>A]` becomes `c.710_713inv`, the rows the `NM_004006.2` inversion sweep still repartitions return to their authored `inv` — that population is `CENSUS_REPARTITIONED` = **92** on the default arm at `origin/main` (93 on `live`), **not** the 155 this record was first costed on and not only the 25 `delins`-bearing ones #1575 costed — and #1230's checked-in guards are flipped. The real-corpus measurement is **owed by the implementing change and is not supplied here** — no A/B was run, because running one requires the normalizer change this PR deliberately does not make. Disclosed in the release the decision was made in, not only in the release that ships the moves.


nh13:
@coderabbitai pause

nh13:
**This reads `CONFLICTING` but does not need conflict resolution — it needs a rebase.**

The conflict is a stale base, not competing edits: **#1706 has merged** (it is `main`'s tip commit `64fc76df`, *"recognise inversions post hoc over blocks and runs of pieces"*), and this branch predates it. Both touch the ruling ledger and `tests/it/clause_ruling_index.rs`, so the overlap is real but it is already-landed work rather than a rival change.

`git fetch origin main && git rebase origin/main` should be the whole of it.

Two cautions specific to this branch, both from `CLAUDE.md`:

- **Do not hand-merge `tests/it/clause_ruling_index.rs`.** It carries three census pins and a ~392-line generated block; regenerate instead, and `the_rendered_index_is_current` prints the exact block to paste.
- **This is a ledger-bearing branch, so check the records rather than the commits** after rebasing — the set-comparison script under *"For a ledger-bearing branch, check the records, not the commits"*. Nothing in the old tip or in `main` may be missing from `HEAD`. A ruling was destroyed here once by a push that passed its lease, and that is the check which would have caught it.

nh13:
This is `CONFLICTING` again after #1838 merged, and it is not a stale base — `git merge-tree` against current `main` reports exactly one conflicted file:

```
tests/fixtures/grammar/hgvs_spec_normalization_overrides.json
```

Both sides rewrite the **same `rationale` string** of the `inversion-vs-two-delins-76-83` record: #1838 scopes its claims to what was actually measured, this PR adds the whole-span-reverse-complement supersession paragraph.

Flagging rather than resolving, because composing two competing rewrites of one ruling's rationale is an adjudication rather than a merge — taking either side wholesale silently discards the other's claim.

(Context: this branch was rebased onto `750f2334` earlier today as `822ac9ba` → `55629b27`, which was clean at the time; #1838 landed afterwards.)

nh13:
Rebased onto current `main` (`55629b27` → `1d9aa3d9`) and the ledger conflict resolved. Full local gate green — 10561 run, 10561 passed, 0 skipped-on-missing-fixture, `EXIT=0` read back from the log.

**The conflict was not a contradiction — the two rewrites are sequential**, which is only visible once both are in hand.

- **#1838** measured the all-lone-substitution population — 329 typed `inv` against 1,608 split, over 1,937 candidates — narrowed this record away from it, and recorded it as UNDECIDED ("this record neither licenses those 329 nor condemns them").
- **This PR** adds `whole-span-reverse-complement-types-as-inv` (`decided`), which decides exactly that population, uniformly `inv`, and overturns #1230.

So #1838 supplies the evidence and this PR rules on it. The new record already anticipated the merge — it says the older record "has been amended to say so rather than left asserting the distinction," which is precisely this PR's edit.

**Resolution:** `inversion-vs-two-delins-76-83` now carries #1838's rationale in full — its correction of the `no_piece_is_a_lone_substitution` note (that it is a disjunction, not "the gate"), its scope narrowing, the 1,937/329/1,608 measurement, and the inversion-sweep parity finding — followed by this PR's `SUPERSEDED IN PART` block verbatim. Nothing from either side was dropped or paraphrased.

**Ledger integrity, checked against both parents rather than against the commits:**

```
HEAD=32  main=31  pr=29
dropped from main: NONE
dropped from PR  : NONE
status changes vs main: none
added vs main: whole-span-reverse-complement-types-as-inv (decided)
```

That check was worth running twice. An earlier attempt at this composition was built against a `main` snapshot taken minutes before #1899 merged, and would have silently dropped **two** records it had just added — `coding-axis-merges-are-a-disclosed-general-34-deviation` and `unchanged-is-read-over-every-minimal-alignment`. The record comparison caught it; the commit-level checks would not have.

Derived artifacts regenerated rather than hand-edited: the `clause_ruling_index` block re-rendered from the guard's own output (482 lines), `docs/NORMALIZATION_CONTRACT.md` re-blessed via `BLESS_CONTRACT_DOC=1`. Two pins moved with the ledger and are re-blessed here: the census `(31, 28, 3)` → `(32, 29, 3)`, and multiply-named clause lines `34` → `36`.

One thing deliberately **not** written into the record, flagged here instead: under the new ruling, #1838's 1,608 split rows are non-conformant and would move to `inv`. That figure is absent from the new record's own cost section, which was measured on a tree that did not carry #1838's measurement — the two were written on sibling branches the same day. It is an implied cost for the implementer to re-derive on a current tree, not a disclosure either record has made.

PR #1840, opened 2026-08-13
fix(normalize): make a cis partition independent of the shuffle direction

### LINKED ISSUE #1542: Normalizer: a dup-shaped split still emits members at separation 0 (delins.md:16), surviving #1537

nh13:
Closes #1542.

Representation-Change: 54 of 85,642 corpus rows move, 44 merge / 10 respell. All 54 fall in the two inversion families (`inv_member` 38, `delins_hiding_an_inversion` 16); the other 19 families move 0. 19 of the 54 are on the shipped 3' direction, 35 on the internal 5' oracle. NOT a migration: all 54 were previously not fixed points, so no stored string was stable for any of them, and 38 become fixed points. Measured with examples/dump_normalized_corpus.rs against origin/main@2d8b490b with only src/normalize/merge.rs swapped. The previously declared 150 was measured on the pre-#1835 base and is withdrawn.

## The defect

`NC_000017.11:g.80110044_80110047delinsGTTGG` — reference span `CTGA`, payload `GTTGG` — normalized to two different strings depending on nothing but the shuffle direction:

```
3'  ->  g.[80110044C>G;80110045dup;80110047A>G]     three members
5'  ->  g.[80110044delinsGT;80110047A>G]            two members
```

Measured on the branch base across all four `FERRO_PARTITION` arms × both directions: **7 of the 8 configurations already emitted the merged two-member form; only `live`/3' produced the split.** Both answers are fixed points, and on `live`/3' the *merged* form normalizes **into** the split — so it is a fixed point reached from both sides, and `FERRO_ASSERT_IDEMPOTENT`, `FERRO_ASSERT_REPARSE`, `FERRO_ASSERT_IN_BOUNDS`, `FERRO_ASSERT_SEQUENCE` and every confluence class are satisfied by the wrong answer.

**Mechanism.** `partition_block` cuts the block at edit-op granularity — `[C>G][ins T][A>G]` — regardless of direction, and `coalesce_adjacent_pieces` (the `delins.md:16` enforcement point) runs *after* `shift_pieces`. The inserted `T` copies the very base that separates it from the substitution, so both junctions denote one sequence: at 5' it rolls onto `44|45`, abuts the substitution and is merged; at 3' it rolls onto `45|46`, escapes the merge, and is then spelled `dup`. The **member count** was therefore a function of `ShuffleDirection`.

## The issue's stated predicate was wrong, and so was the instrument

#1542 reports this as a `delins.md:16` violation at separation 0. It is not, and the corpus-wide guard's one exemption for this row was buying nothing.

`tests/it/case_harvest.rs::member_bounds` was a textual leading-number parse, so it read `80110045dup` as occupying `[80110045, 80110045]` — the base it duplicates. A duplication **inserts** and consumes no reference position. Re-instrumented on SPDI write footprints (`hgvs_to_spdi`, the same geometry `examples/dump_confluence_divergences.rs` documents), the three members sit at separation **[1, 1]**. The old instrument was wrong on a second shape too: `a_b insT` read as the two-base span `[a, b]`, where `insertion.md:15` calls those the **flanking** nucleotides.

## The fix

`shift_and_coalesce_direction_symmetrically` (`src/normalize/merge.rs`) replaces the bare `shift_pieces` + `coalesce_adjacent_pieces` pair at both of its call sites. It asks the mirror direction as well, and where the mirror closes a gap this direction did not, adopts that partition and re-shifts it the way the caller asked.

**The rule: the shuffle direction may move a member's placement; it may not change how many members there are.** Adopting the *merged* side is not a preference for merging — a separation that one legal placement erases was never a separation, while a gap both placements leave open is real.

The mirror is only computed where its answer can differ, and the precondition is exact rather than heuristic: `shift_pieces` moves nothing but pure indels, so a block with none of those places identically both ways.

### Authority

The operator's 2026-08-13 ruling on #1542, resting on two `decided` ledger records:

- **`separation-is-a-property-of-the-spelling-not-of-the-variant`** — the separation `general.md:34` keys on is read off the partition **re-derived from the resulting sequence**, never off a spelling. A partition that varies with the shuffle direction is read off a placement.
- **`duplication-must-ranks-the-label-not-the-partition`** — `duplication.md:18` ranks the *label* applied to a derived change and never requires a partition that exposes one. This is whole-block re-derivation, and the re-derived piece `CT → GTT` replaces a base as well as adding one, so `duplication.md:17`'s scope fails and `:18` never fires.

**This is not the reserved `dup` half of `delins-adjacent-members-when-both-consume-reference`.** That carve-out is about merging two adjacent **members** at separation zero, one carrying a `dup` the merge would destroy. Here the input is a single spanning `delins` carrying no members at all, the emitted members measure separation `[1, 1]` rather than zero, and the `dup` is *manufactured* by `live`/3' rather than carried. Structurally the pass also cannot reach that shape: `pieces_from_columns` flushes a run on an unchanged column, so consecutive pieces are always at separation one or more.

## Unchanged on purpose

`partition_block` and `best_alignment` are **not touched**. A first attempt repaired the partition itself, by merging pieces whose separating column is not a dominator of every minimal alignment. It reddened 10 tests including `tie_break_by_separation` (#1262/#1232), `two_gap_insertion_alignment` (#1260), `splitter_reproducer_corpus` and `seqfirst_shadow_audit` — settled behaviour with its own citations, which this issue has no authority to move. The measurement that redirected the fix: of `AAA→CA` (#1262), `TTTTT→TA` (#1232), `A→CAC` (#1260) and #1542's own `CTGA→GTTGG`, **only #1542's shape is direction-dependent**; the other three already normalize identically in both directions. So the defect is the direction dependence, and the fix is scoped to it.

## The guard, restated

`the_dup_shaped_split_does_not_touch` was `#[ignore]`d and red on a predicate the row never violated. It is replaced by `the_dup_shaped_split_is_the_partition_of_its_own_resulting_sequence`, which runs (not ignored) and asserts two things that fail independently:

1. the row's answer does not depend on the shuffle direction;
2. its member footprints equal those of the partition **re-derived from the resulting sequence** by `from_sequences_detailed` — which holds no provider, cuts with a different partitioner, and never sees this output, so it is not the normalizer checking itself. The `(reference, resulting)` pair is built from the input's own stated payload, not from anything normalization produced.

**Sabotage-checked, both halves separately**, because a guard that stays green under sabotage is worse than none:

| sabotage | result |
|---|---|
| mirror disabled (`false &&`) | RED on half 1 — `Ok("…C>G;…dup;…A>G")` vs `Ok("…delinsGT;…A>G")` |
| mirror disabled **and** half 1 neutered | RED on half 2 — footprints `[(…44,…44),(…46,…45),(…47,…47)]` vs `[(…44,…44),(…47,…47)]` |

## `allows_adjacent_members` was removable, and is removed

`Case::allows_adjacent_members` was a permanent single-row hole in the corpus-wide `no_emitted_output_puts_two_members_on_consecutive_nucleotides`. With the instrument corrected the row it exempted is not a violation, so the predicate is deleted and **no row is exempt**. `DefectKind::AdjacentMembers` is **kept**: classifying which guard a future red row is expected to fail is a different job from waving that guard off, and it has no other enumerating pass to fall out of.

Re-instrumenting also *widens* the guard — a member the textual reader declined (a UTR position, an intronic offset) now resolves through the provider instead of being skipped.

## Verification

Read back from the logs, not from completion notifications.

```
cargo fmt --all --check                                  # FMT_EXIT=0
cargo clippy --all-features --all-targets -- -D warnings # CLIPPY_EXIT=0
FERRO_SWEEP_SEEDS=full cargo nextest run --features dev   # EXIT=0
scripts/run_oracle_suite.sh                               # EXIT=0
```

**Corpus A/B** — the committed harness, `examples/dump_normalized_corpus.rs`, dumped at the branch base and at the candidate with only `src/normalize/merge.rs` swapped (`md5` confirmed against `git show HEAD:src/normalize/merge.rs` each way, after an initial baseline taken against a *moved* `origin/main` was discarded):

| | rows |
|---|---|
| total | 85 642 |
| **moved** | **150 (0.18%)** |
| direction | 150 `3prime`, **0** `5prime` |
| shape family | 150 `split_vs_spanning_delins`, 0 in the other 20 |
| member count | 150 of `3 → 2` |
| denote different bases | **0** — `--verify-spdi` residual is 6 before and 6 after, the same 6 rows in `adjacent_junction_ins` |

**R6 — the property this keys on** is *a multi-piece block carrying at least one pure-indel piece whose legal placements straddle the base separating it from a sibling*. This is **not** a structural zero: `split_vs_spanning_delins` builds exactly that shape and 150 of its 9 216 rows moved, so the corpus can vary the property and did. The corpus is deliberately enriched for churn-prone shapes, so 0.18% is a rate within these families, never a library-wide figure.

**Confluence** (`cis_confluence_axis`, `cis_confluence_nr_axis` — the 11 272-class and 5 636-class designed cis corpora). Three pins re-blessed, each in the direction the pins' own instruction demands (`converged` only up, no divergence figure up), and `sequence_changed` stays 0 throughout:

| census | converged | `split 2` |
|---|---|---|
| `g.`/`c.` 3' | 8 361 → **8 363** | 2 834 → 2 832 |
| `n.` 3' | 4 179 → **4 181** | 1 420 → 1 418 |
| `r.` 3' | 4 177 → **4 179** | 1 422 → 1 420 |
| all three **5'** | **unchanged** | **unchanged** |

The 5' halves being untouched is the shape of the finding rather than a happy accident: the fix moves the 3' answer onto the 5' one.

**Not measured against the operator's brief figure.** The ruling anticipated ~84 classes carrying this signature. That is a different predicate over a different corpus; the figures above are re-derived here rather than inherited, per R6.

## Review

`/coderabbitai-review` against the local diff before pushing (profile `assertive`, repo `path_instructions` walked clause by clause). Two findings applied: a doc-drift claim that `DefectKind::AdjacentMembers` had been deleted when it was kept, and a hardcoded span length beside a named window constant (now derived, with the `CTGA` premise asserted inline). One rejected: the re-derivation reads through a plain rather than audited provider — `genomic_bases` panics naming the span it could not serve, so it fails loud rather than skipping, which is the property the audit discipline exists to protect.

## Coverage note

`tests/it/case_harvest.rs` is hermetic — it reads a committed reference slice, sets no `FERRO_MANIFEST`, and has no skip path — so this is pre-merge coverage on PR CI, not nightly-only.


nh13:
@coderabbitai pause

PR #1842, opened 2026-08-13
test(spec): measure spec conformance per FERRO_PARTITION arm over a real reference

nh13:
The `FERRO_PARTITION` flip is framed as making ferro more conformant and more confluent. Confluence is measured. **Conformance is not**, and the suite that appears to measure it is vacuous by construction — so both directions of that claim are currently unsupported. This adds the instrument that measures it, and reports the first numbers off it.

## The defect, re-derived rather than inherited

`generate_spec_fixture` normalizes with an empty `MockProvider`, so no row can fetch a block and `merge::canonicalize_from_sequence` — the switch's only consumer — is never reached. Generating the fixture under each of the four arms at `7cc9846c` gives a **byte-identical file**, md5 `e09899559c510c248968b435ad4ebdb5` on `live`, `shadow`, `canonical` and `canonical-coalesced` alike, with `generated_utc` pinned to `fixture-byte-stable` so the comparison is real. (An earlier measurement recorded `d198a442005af419d0921b9214a44c54`; that was taken before #1760 added a ruling record. The identity *across arms* is what matters and it reproduces.)

That blindness is pinned by `the_spec_fixture_is_blind_to_the_partition_switch` on #1822. This PR does not duplicate it and changes neither the fixture nor that suite.

## Where `spec_expected` comes from — it is not ferro-derived

Established from `build_rows`. Of 934 rows, **905** take the default: the harvested spec string itself, with a mechanical default accession prepended. **27** are `null` from a `<code class="invalid">` marker. **2** are curated strings, and exactly **one** of those is ferro-derived — `NP_060250.2:p.Gln746_Lys747ins*63`, whose override pins ferro's own `insTer63` rendering and says so in its note. It is reported separately as `ferro_derived_expectations` and excluded from a second numerator, so the measurement can be read with or without it. Comparing ferro to this column is therefore not circular.

**But note what the column means.** `spec_expected` defaults to the row's own input, so `preserved` is a *round-trip* test: "does ferro echo the spec's string back". Where the spec publishes two legal spellings of one variant as two rows, both echo, both score as conforming, and a change that converges them must cost one of them. That is exactly what the measurement below finds, and it is why "more conformant" cannot be read off this corpus without care.

## What is added

- `normalize::partition_blocks_cut()` — blocks cut on **every** arm. `partition_decline_counts` counts only the sequence-first arms, so it is structurally zero on `live` and cannot answer "did this corpus reach the switch at all". Debug-only, like the four normalization oracles and for the same reason: the shipped `Live` path must not grow a per-block atomic to serve a measurement.
- `examples/measure_spec_conformance_per_arm.rs` — re-normalizes the corpus over a prepared manifest, one process per arm, classifying with the same `classify` taxonomy the fixture uses. It measures ferro against the spec once per arm; `spec_expected` does not move with the arm, so this is **never** an arm-versus-arm comparison. Rows are counted, never skipped: a refusal is a taxonomy bucket, a panic is its own status and a ledger drop.

## The measurement

Prepared reference, 3' direction, `lenient` (the config the fixture records). Restricted to the 35 multi-member cis rows that are the partitioner-eligible population and are not chromosome-scale; `restricted_to` and `skipped_by_restriction` are stamped in the artifact so this cannot be read as a full census.

| arm | conforms | diverges | reached partitioner | blocks cut |
|---|---|---|---|---|
| `live` | 11 / 35 | 13 | 14 | 14 |
| `shadow` | 11 / 35 | 13 | 14 | 14 |
| `canonical` | 10 / 35 | 14 | 14 | 15 |
| `canonical-coalesced` | 10 / 35 | 14 | 14 | 15 |

**The effective denominator is 14** — measured, not inferred from a re-implementation of the gate. It is non-zero, so the numbers are evidence rather than structure.

**The delta is one row**, and it is `DNA/delins.md:44-47`'s own worked example:

```
input   c.[850_869del;874_881del;887_897del;901_902insG]
live    c.[850_869del;874_881del;887_897del;901_902insG]      preserved
canon   c.[850_866del;870_880del;887_892del;895_896delinsC;899_901del]   diverges
c-coal  c.850_901delinsTTCCTCGATGCCTG                          diverges
```

Its sibling spelling `LRG_199t1:c.850_901delinsTTCCTCGATGCCTG` is a corpus row too, and the pair is the curated equivalence class `dna-delins-vs-aligned-split-850-901`. Measured per arm on the coordinate form:

| arm | distinct coordinate forms | verdict |
|---|---|---|
| `live` | 2 | non-confluent |
| `canonical` | 1 — `c.[850_866del;…;899_901del]` | confluent |
| `canonical-coalesced` | 1 — `c.850_901delinsTTCCTCGATGCCTG` | confluent |

So the flip **converges the class**, and `canonical-coalesced` converges it onto the spanning delins `delins.md:47` actually recommends, while `canonical` converges it onto a five-member split matching neither published spelling.

**Read together: this corpus does not show the flip improving conformance. It shows conformance falling by one row and confluence rising by one class, on the same row** — the round-trip echo of `:46`'s "alternative description" is what is traded for convergence. That is precisely what `canonical-form-choice-when-both-legal` rules should happen. A release note claiming "more conformant" cannot cite this corpus; one claiming "more confluent, at the cost of one echoed alternative spelling" can.

## What the corpus can and cannot vary here

It **can** vary member count (52 rows carry a multi-member cis allele), coordinate system (`c`/`g`/`n`/`r`/`m`/`o`/`p`), member separation, and payload coincidence. It **cannot** vary transcript geometry — the accessions are fixed and mostly `NM_004006.x` — and it carries no designed sweep, so absence of movement in a stratum is weak evidence.

Two limits are stated rather than hidden. `protein_fastas` is empty in the reference used, so the 218 `p.` rows cannot be reference-backed; `sequence_first_pass` refuses protein anyway, so they cannot reach the switch either way. And a **full 934-row per-arm census was not obtained**: several rows are effectively unbounded in cost under a real reference — `NC_000002.12:g.pter_8247756delins[NC_000011.10:g.pter_15825272]` ran over 6 minutes and `c.(516+1_517-1)_(1970_6774)dup` likewise — so the whole-corpus number is still open. `--progress-every` exists so that a future run diagnoses this in seconds rather than presenting as a hang.

Representation-Change: none. The only `src/` change is a debug-only census counter incremented at `partition_block_for_rule`; it is read by nothing on the normalization path and is compiled out of release builds.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -333,6 +333,11 @@ required-features = ["dev"]
 test = true
 
 [[example]]
+name = "measure_spec_conformance_per_arm"
+required-features = ["dev"]
+test = true
+
+[[example]]
 name = "extract_spec_worked_example_windows"
 required-features = ["dev"]
 # `#[path]`-includes `examples/common/recording.rs`, which carries `#[cfg(test)]`
--- a/examples/measure_spec_conformance_per_arm.rs
+++ b/examples/measure_spec_conformance_per_arm.rs
@@ -7,10 +7,20 @@
 //! can fetch a block and `merge::canonicalize_from_sequence` — the only consumer
 //! of the `FERRO_PARTITION` switch — is never reached. The fixture is therefore
 //! **byte-identical on all four arms**, and every assertion the spec suite makes
-//! is constant across them by construction. That blindness is pinned by
-//! `it::hgvs_spec_normalization_tests::the_spec_fixture_is_blind_to_the_partition_switch`;
-//! this generator is the measurement that closes it, and it deliberately does
-//! not change the fixture or that suite.
+//! is constant across them by construction.
+//!
+//! **That blindness is not yet pinned by anything in this tree.** The guard is
+//! `it::hgvs_spec_normalization_tests::the_spec_fixture_is_blind_to_the_partition_switch`,
+//! which lands with #1822 — open, not merged — so until it does, the byte-identity
+//! above rests on the measurement recorded in this PR's description rather than on
+//! a test. Do not cite it as coverage before checking it exists:
+//!
+//! ```text
+//! grep -rn the_spec_fixture_is_blind_to_the_partition_switch tests/
+//! ```
+//!
+//! This generator is the measurement that closes the blindness, and it
+//! deliberately does not change the fixture or that suite.
 //!
 //! # What it measures, stated so nobody reads it as something else
 //!
@@ -100,6 +110,31 @@ struct Cli {
     /// Shuffle direction to normalize under.
     #[arg(long, default_value = "three-prime")]
     direction: Direction,
+
+    /// Measure only the rows named in this file, one `input` per line.
+    ///
+    /// The cross-arm shortcut, and it is exact rather than a sample. Every bail
+    /// on the way to `partition_block_for_rule` is arm-independent — the switch
+    /// is first consulted *at* that call — so a row that cuts no block on one
+    /// arm cuts none on any, and its output is identical across all four. Run
+    /// the full corpus once, take the rows whose `blocks_cut` is non-zero, and
+    /// the remaining arms need only those. Roughly 900 of 934 rows are then
+    /// skipped, which is the difference between a half-hour arm and a brisk one.
+    ///
+    /// Reported in the artifact as `restricted_to`, so a census taken this way
+    /// can never be mistaken for a full one.
+    #[arg(long)]
+    only_inputs: Option<PathBuf>,
+
+    /// Print a progress line every N rows. Zero disables.
+    ///
+    /// Not a nicety: the first full run of this generator spent 29 minutes of
+    /// CPU without finishing, and with no per-row output there was no way to
+    /// tell a slow corpus from a stalled one. The line carries the elapsed time
+    /// and the row being measured, so a pathological row (a whole-chromosome
+    /// inversion, a `pter`-anchored delins) names itself.
+    #[arg(long, default_value_t = 25)]
+    progress_every: usize,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, clap::ValueEnum)]
@@ -172,9 +207,25 @@ struct Measurement {
     /// built from a partial pass, and nothing else in the file would say so.
     normalized_under: ErrorModeStamp,
     /// One row in, one row out. A refusal is *recorded* rather than dropped, so
-    /// `dropped` is expected to be zero — the stamp is what lets a reader see
-    /// that rather than assume it.
+    /// the only drop this pass can produce is a panic.
+    ///
+    /// **`dropped` is always zero here, and that is a property of the control
+    /// flow rather than a measurement**: `ledger.finish()?` refuses before the
+    /// artifact is written, so a run with a drop produces no file at all. Read
+    /// this field as "attempted == succeeded == the row count below", which is
+    /// the claim it can actually support — not as evidence that drops were
+    /// looked for and found to be none. A dropped row is reported on stderr by
+    /// the refusal, never here.
     completeness: CaptureCounts,
+    /// `Some` when `--only-inputs` restricted the run. A census taken this way
+    /// covers only the named rows and is not comparable to a full one.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    restricted_to: Option<String>,
+    /// Rows the restriction skipped. Zero on a full run.
+    skipped_by_restriction: usize,
+    /// Wall seconds this measurement took, for the record. Not a benchmark —
+    /// these runs share a heavily loaded machine.
+    elapsed_seconds: u64,
     total: usize,
     by_status: BTreeMap<String, usize>,
     /// `preserved` — ferro renders the spec's stated form. The conformance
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
Extending this measurement rather than redoing it: I reproduced the 3′ table exactly, then closed the two caveats it states — the missing 5′ direction and the unobtainable whole-corpus census. Everything below is measured on `64fc76df` with this PR's three commits cherry-picked, against the same prepared reference, one process per arm.

**Positive control first.** The 35-row 3′ table reproduces byte for byte on a different base: `live` 11, `shadow` 11, `canonical` 10, `canonical-coalesced` 10, `reached_partitioner` 14, `blocks_cut` 14/14/15/15.

## The whole-corpus census is reachable — only two rows are unbounded, not the two named here

The description names `g.pter_8247756delins[NC_000011.10:g.pter_15825272]` and `c.(516+1_517-1)_(1970_6774)dup`. **Both complete at 0 s**, and I checked that on *this PR's own built binary* at `7cc9846c` as well as on my branch, so it is not a base difference. The pter row is refused upstream with `cross-reference shape not yet supported`, so its payload is never resolved at all; the uncertain-range dup is `preserved`.

The two rows that actually do not terminate are:

- `NC_000004.12:g.134850793_134850794ins[NC_000023.11:g.89555676_100352080]`
- `NC_000004.12:g.134850793_134850794ins[NC_000023.11:g.89555676_100352080inv]`

`ins[ACC:g.A_B]` **is** a supported cross-reference shape, so the 10,796,405 bp payload gets expanded and handed to `normalize::rules::insertion_to_duplication`, which rotates it once per base and rebuilds it each time — Θ(payload²), measured at ~4× per doubling, 1.4 GB RSS, extrapolating to ~52 min for that row. Filed with the stack profile and the cost law as #1846. `MAX_SEQFIRST_GRID_CELLS` is not the missing guard: it is real and works, but it sits in `merge::canonicalize_from_sequence`, downstream of where this cost is paid.

Payload size is not the predictor, which is why I measured row by row rather than capping: 18 corpus rows declare a payload ≥ 100 kb and 17 ≥ 1 Mb (one is 224 Mb), and only these two stall.

**With those two excluded the corpus completes in 6 s** — 932 of 934 rows, `dropped: 0`, on every arm and both directions.

## Whole-corpus census, 932/934 rows

`skipped_by_restriction: 0`; exclusions stamped in the artifact as `excluded_by` / `excluded_reason` / `excluded_inputs` (see the patch below).

**3′** — effective denominator (`reached_partitioner`) **232**

| arm | conforms | blocks cut |
|---|---|---|
| `live` | 650 / 932 | 233 |
| `shadow` | 650 / 932 | 235 |
| `canonical` | 646 / 932 | 243 |
| `canonical-coalesced` | 648 / 932 | 241 |

**5′** — effective denominator **241**

| arm | conforms | blocks cut |
|---|---|---|
| `live` | 627 / 932 | 242 |
| `shadow` | 627 / 932 | 244 |
| `canonical` | 624 / 932 | 252 |
| `canonical-coalesced` | 626 / 932 | 250 |

`shadow` is identical to `live` on conformance in both directions — 0 rows lost, 0 gained — while cutting more blocks, which is what a comparison-only arm should look like.

## The cost is four rows, not one — and the restriction is why it read as one

Rows that are `preserved` on `live` and stop being so, whole corpus, 3′:

| arm | rows lost |
|---|---|
| `canonical` | `c.[850_869del;874_881del;887_897del;901_902insG]`, `LRG_199t1:c.850_901delinsTTCCTCGATGCCTG`, `LRG_199t1:c.9002_9009delinsTTT`, `LRG_199t1:c.992_1004delinsAC` |
| `canonical-coalesced` | `c.[850_869del;...;901_902insG]`, `LRG_199t1:c.992_1004delinsAC` |

No arm gains a row anywhere. Three of the four are **single-member** `delins` rows, so the restriction to multi-member cis alleles excluded them by construction — the 35-row window could only ever have shown the one it showed. This does not contradict the PR's reading, it widens it: the trade is 4 rows for `canonical` and 2 for `canonical-coalesced`, and `canonical-coalesced` is consistently the better of the two (648 vs 646 at 3′, 626 vs 624 at 5′).

## 5′ diverges, and the headline delta is 3′-only

On the restricted 35-row set the two directions do not agree:

| direction | `live` | `shadow` | `canonical` | `canonical-coalesced` | reached |
|---|---|---|---|---|---|
| 3′ | 11 | 11 | 10 | 10 | 14 |
| 5′ | 10 | 10 | 10 | 10 | 15 |

**At 5′ all four arms tie at 10/35.** The "conformance falls 11 → 10" delta does not exist at 5′, because `live` already diverges on `c.[850_869del;874_881del;887_897del;901_902insG]` there — it is one of exactly three gate rows whose `live` verdict differs between directions. The same row is absent from the 5′ whole-corpus loss list for that reason.

So the release-note sentence wants a direction on it. "More confluent, at the cost of one echoed alternative spelling" is a **3′** statement about the restricted set. On the whole corpus the cost is 4 rows at 3′ and 3 at 5′ for `canonical`, 2 and 1 for `canonical-coalesced`.

## On the accession caveat

I checked whether the missing re-anchoring reaches these numbers: it does not. `classify` only ever compares a row against its **own** `spec_expected`, never against another row, and of the 780 rows carrying a `spec_expected`, **0** have a target/expected accession mismatch — `build_rows` prefixes both sides identically. Re-anchoring matters for the equivalence-class table, which is cross-row; the conformance table is not. I left the comparison unmodified so 3′ and 5′ come from one instrument.

## Handover

The only change I made is measurement-only: `--exclude-inputs` plus a mandatory `--exclusion-reason`, stamped into the artifact so a census cannot silently report a smaller denominator. One signed commit, `cargo fmt` and `clippy -D warnings` clean:

`campaigns/patches/1842-exclude-inputs-stamping.patch`

**It belongs on this branch, not a separate PR** — a standalone PR would carry a duplicate of this PR's whole diff. Take it, rewrite it, or drop it. The exclusion list is two lines and is reproducible from #1846.

`Representation-Change: none` — no `src/` change beyond the counter already in this PR.

Issue #1826, opened 2026-08-13
Unknown intronic offset (`c.<base>-?`) is classified as a canonical splice site with HIGH impact


nh13:
`c.<base>-?` parses to `offset = Some(i64::MIN)` (`OFFSET_UNKNOWN_NEGATIVE`), and `is_intronic()` is true for it, so an unknown-distance intronic position reaches the splice-distance ladder like any other. The ladder then takes `.abs()` of that offset.

`i64::MIN.abs()` has no representable result. In a debug build it panics (`attempt to negate with overflow`); in a release build it wraps back to `i64::MIN`, which is negative, so every `<= n` rung comparison reads true and the position is classified at the **first** rung.

The consequence is a silent misclassification in the direction that matters: a position whose distance from the splice site is *unknown* is reported as a **canonical splice site with HIGH impact**. It is not a parse rejection or an obviously wrong number — it is a confident answer about a position the description explicitly declines to locate.

This is reachable from the web-service handler (`src/handlers/effect.rs:136`), not only from library callers.

## Where the ladder is

The shared rung bodies, plus `IntronPosition`'s four inline-threshold predicates. `IvsPos` in `src/hgvs/location.rs` has the same shape and is **not** covered by the fix in #1742 — that file is in a watched directory and outside that PR's diff, so it was deliberately left alone.

## Status

The two shared bodies and `IntronPosition`'s four predicates are fixed on #1742's branch (`unsigned_abs()` plus a guard), because that PR's de-duplication happens to reduce the fix to two words. The fix is verified by sabotage: reverting one line reddens the suite with `attempt to negate with overflow`.

What remains after #1742 lands:

1. `IvsPos`'s copy of the ladder in `src/hgvs/location.rs`.
2. The larger question, which is deliberately **not** answered by the `unsigned_abs()` fix: whether an unknown offset should be *classified at all*. Treating `-?` as "distance 0" or as "very large" are both fabrications. Declining — returning `None` and letting the caller see that no consequence could be determined — is the semantically right answer, but it is a public-API and response-shape change and wants its own decision.

The `unsigned_abs()` fix removes the panic and the wrap; it does not make the resulting classification meaningful. Item 2 is the real fix.

## Reproducing

Any `c.<base>-?` description through the effect classifier. Under `overflow-checks` it panics; without them it returns `splice_site_variant` / HIGH.

nh13:
## Verified fixed and superseded — closing

[...]

PR #1843, opened 2026-08-13
refactor(effect)!: decline an unknown offset from the two splice classifiers that could not

### LINKED ISSUE #1841: Make `IntronicRegion::from_offset` and `classify_splice_variant` return `Option` — neither can answer for an unknown-offset sentinel
Merged PR #1806 (`5bf0b4ff`) settled #1767 by teaching every reachable splice classifier about the
unknown-offset sentinels (`-?` / `+?`, carried as `i64::MIN` / `i64::MAX`). It deliberately stopped
short of two functions, and wrote that down under "Deliberately left open — needs an API call this PR
should not make". #1767 is now closed, so this is the follow-up that carries that half.

## The two signatures

Both are on `origin/main` at `5bf0b4ff` exactly as #1806 left them.

**1. `IntronicRegion::from_offset(offset: i64) -> Self`** — `src/convert/noncoding.rs:280`

```rust
pub fn from_offset(offset: i64) -> Self {
    let abs_offset = offset.unsigned_abs();
    if abs_offset <= 2 { Self::CanonicalSpliceSite }
    else if abs_offset <= 20 { Self::ExtendedSpliceRegion }
    else if abs_offset <= 50 { Self::NearSpliceSite }
    else { Self::DeepIntronic }
}
```

Every variant of `IntronicRegion` names a **distance band**. A sentinel states that the distance is
unknown, so none of the four is a true answer for it — and the function's own doc comment says so:

> a sentinel passed here anyway lands in `DeepIntronic`, which is a distance claim its input does not
> support.

A doc comment is the weakest way to express a precondition. `Option<Self>` is the type system saying
the same thing.

**2. `EffectPredictor::classify_splice_variant(&self, offset: i64) -> ProteinEffect`** — `src/effect/mod.rs:404`

#1806 gave this one the weakest *true* statement rather than a decline — `Consequence::IntronVariant`
/ `Impact::Modifier` — precisely because the return type cannot decline:

> This entry point returns a bare `ProteinEffect` and so cannot decline the way the `Option`-returning
> classifiers do, and `IntronVariant` is the weakest true statement available.

That is a correct answer under the constraint, and the constraint is the signature. `IntronVariant`
still reads as a positive classification to a caller that is switching on the consequence; `None`
reads as "this input has no answer", which is the actual state of affairs.

## This is API honesty, not a live defect

Stated plainly so nobody re-triages it as a crash:

- `IntronicRegion::from_offset`'s two in-repo callers — `from_cds_pos` and `from_tx_pos` — screen with
  `has_unknown_offset()` before reaching it (#1806), and both already return `Option`. Its other call
  sites are its own unit tests. So no sentinel reaches it from inside this repo. A **direct external
  caller** passing one still gets `DeepIntronic`.
- `classify_splice_variant` *is* reachable with a sentinel — `src/bin/ferro.rs:2715` passes
  `start_pos.offset` through unscreened — but it handles the sentinel internally and answers
  `IntronVariant`, which is true. It is not wrong; it is just unable to say "no answer".
- Neither is on the web-service path. `predict_cds_effect`
  (`src/service/handlers/effect.rs:578`) has its own `has_unknown_offset()` screen and does not call
  either function. The panic and the `splice_site_variant`/HIGH release answer that #1767 was filed
  for are both fixed and are not what this issue is about.

So the arithmetic is safe on every `i64` (`unsigned_abs`, total), the answers are true, and the
remaining gap is that two public signatures promise an answer they cannot always give.

## The break

`classify_splice_variant` has a **PyO3 binding** — `src/python.rs:4042`, declared in
`python/ferro_hgvs/__init__.pyi:1380` as `def classify_splice_variant(self, offset: int) -> ProteinEffect`.
Making the Rust function return `Option<ProteinEffect>` makes the Python method return
`ProteinEffect | None`, which is a break for any Python caller that indexes into the result
unconditionally. That is the point of the change — the Python surface should express "no answer" as
`None` rather than hand back a `ProteinEffect` that merely declines to be specific — but it is a
break, and it must ship as one.

Measured blast radius on `origin/main` at `5bf0b4ff`:

| symbol | in-repo call sites |
|---|---|
| `IntronicRegion::from_offset` | 2 internal (`from_cds_pos`, `from_tx_pos`, both already `Option`-returning), 6 in its own unit-test module, 1 in `tests/it/issue_1767_unknown_offset_splice_classifiers.rs` |
| `classify_splice_variant` | 1 real (`src/bin/ferro.rs:2715`), 1 PyO3 binding, 4 in `src/effect/mod.rs`'s unit tests, 2 in `tests/it/issue_1767_unknown_offset_splice_classifiers.rs` |

There are **no** Python tests for `classify_splice_variant` today (`git grep` over `tests/python/`
returns nothing), so the binding's behaviour on a sentinel is currently unpinned in either direction.

The `ferro effect` CLI is the one caller that has to grow a new arm: today an unknown offset prints
`intron_variant`/MODIFIER, and with a declining classifier it has to say it cannot predict an effect.
`run_effect` already ends with `Err(format!("Variant type not supported for effect prediction: {v}"))`
for the shapes it cannot handle, so there is an existing convention to match rather than a new exit
status to invent.

## Approved

The operator has **approved taking the break** for a major bump. This issue exists so the decision is
recorded against something trackable — #1767 is closed and would otherwise have buried it.

Refs #1767, and PR #1806 which measured the trade-off and left it open.

Related: #1826, whose item 2 asks the same question in wider terms ("whether an unknown offset should
be *classified at all* … a public-API and response-shape change [that] wants its own decision"). This issue is the narrow, decided form of
that item 2, scoped to the two signatures and their PyO3 surface; #1826 should be reviewed separately
for whatever it still holds.

Its item 1 — `IvsPos`'s copy of the ladder — was already addressed in #1806.

nh13:
Closes #1841.

Merged PR #1806 settled #1767 by teaching every reachable splice classifier about the unknown-offset sentinels, and deliberately left two functions with their signatures under "Deliberately left open — needs an API call this PR should not make". The operator approved taking that break. This is it.

```rust
-pub fn from_offset(offset: i64) -> Self                                  // IntronicRegion
+pub fn from_offset(offset: i64) -> Option<Self>

-pub fn classify_splice_variant(&self, offset: i64) -> ProteinEffect      // EffectPredictor
+pub fn classify_splice_variant(&self, offset: i64) -> Option<ProteinEffect>
```

## Semver — deliberately breaking, and typed as such

Both functions are `pub` on a published crate (0.13.1), and `classify_splice_variant` additionally has a PyO3 binding, so the Python method's return type moves from `ProteinEffect` to `ProteinEffect | None`. A Python caller reading `.consequences` off the result unconditionally now raises `AttributeError` on `+?` / `-?`.

Typed `refactor(effect)!:`, with the `!` **in the PR title as well as the commit subject** — squash-merge builds the commit from the title, and that is what release-plz reads. `release-plz.toml`'s own comment states the rule this relies on: for a 0.x crate release-plz reserves the minor bump for breaking changes, so the marker is what moves this to 0.14.0 rather than 0.13.2, and git-cliff renders it with the `[**breaking**]` marker. `fix(error-handling)!` (#1674), `fix(spdi)!` (#1363) and `refactor!` (#1168) are the precedents for the form; like all six landed `!` commits, this one hand-edits no changelog — `CONTRIBUTING.md` is explicit that `CHANGELOG.md` is generated by release-plz and must not be edited in a feature PR.

## Why `Option` is the correct answer

`-?` and `+?` are legal HGVS. They say the position is intronic and decline to say how far from the boundary it sits, and the parser carries that as `i64::MIN` / `i64::MAX`.

- **`IntronicRegion`** has four variants and every one of them names a **distance band**. A sentinel supports none of them. Before this change a sentinel landed in `DeepIntronic`, and the function's own doc comment said so in as many words — "a distance claim its input does not support".
- **`classify_splice_variant`** answered `IntronVariant` / `Modifier`, which #1806 chose as "the weakest true statement available" *given a signature that could not decline*. It is true, and it still reads as a positive classification to a caller matching on the consequence.

The rule itself is not new: `CdsPos::has_unknown_offset`'s doc comment carries a MUST from #1087 — the sentinels denote a position unbounded in one direction, so no coordinate can be derived from them at all — and `IntronicConsequence::from_{cds,tx}_pos`, `IntronicRegion::from_{cds,tx}_pos` and both validators already decline on it. This applies that existing contract to the two entry points whose return type was in the way.

## Not a live defect — this is API honesty

Stated plainly so it is not re-triaged as a crash:

- `IntronicRegion::from_offset`'s two in-repo callers screened before reaching it, so no sentinel reached it from inside this repo. A **direct external caller** got `DeepIntronic`.
- `classify_splice_variant` *is* reachable with a sentinel from `ferro effect`, and its answer was true.
- Neither is on the web-service path — `predict_cds_effect` has its own screen and calls neither. The panic and the `splice_site_variant`/HIGH release answer #1767 was filed for are both fixed and are not what this changes.

## Call sites

| symbol | sites | what changed |
|---|---|---|
| `IntronicRegion::from_offset` | `from_cds_pos`, `from_tx_pos` | `.map` → `.and_then`; their own `has_unknown_offset()` screens are **collapsed into `from_offset`**, so the rule has one site rather than three |
| | 6 in its own unit-test module | wrapped in `Some(...)` |
| `classify_splice_variant` | `src/bin/ferro.rs` (the only real caller) | new `None` arm, below |
| | `src/python.rs` PyO3 binding | `-> Option<PyProteinEffect>`, plus `python/ferro_hgvs/__init__.pyi` |
| | 4 in `src/effect/mod.rs`'s unit tests, 6 in `tests/it/issue_1767_*` | `.expect` / rewritten |

## The one behaviour change, measured on both sides

`ferro effect` is the only shipped surface whose output moves. Measured in this worktree, `origin/main` `5bf0b4ff` against the branch, same binary path, same inputs:

| input | before | after |
|---|---|---|
| `NM_000088.3:c.100-?del` | `intron_variant (MODIFIER)`, exit **0** | `Error: "Cannot predict an effect for an unknown intronic offset …"`, exit **1** |
| `NM_000088.3:c.100-2del` | `splice_acceptor_variant (HIGH)`, exit 0 | **unchanged** |
| `NM_000088.3:c.100del` | `coding_sequence_variant (MODIFIER)`, exit 0 | **unchanged** |

The refusal matches `run_effect`'s existing unsupported-type arm rather than inventing an exit status. Under `-f json` the single-variant path now emits no JSON object at all for that input, which is stated in the code comment because a pipeline consumer sees it.

**Batch mode does not regress into #1764.** `--input` and stdin already catch a per-line `Err`, report it to **stderr** as `ERROR: <line> - <e>` and continue, so one declining line does not truncate the stream. Verified by reading both loops, not assumed.

## Tests

New: `tests/it/issue_1841_option_returning_splice_classifiers.rs` (5 tests, registered in `tests/it/main.rs`), `tests/python/test_issue_1841_splice_classifier_decline.py` (8 tests), plus a unit test beside each changed function.

**The discriminating case in every one of them is `i64::MIN + 1` / `i64::MAX - 1`.** Those are not sentinels, so they must still classify as `DeepIntronic` / `IntronVariant`. A decline keyed on "a very large magnitude" rather than on the sentinel passes every assertion that only looks at `i64::MIN` and `i64::MAX` — and that is exactly the gap #1765 records one layer up.

**Sabotage-checked rather than assumed.** With the two `is_unknown_offset` early-returns neutered (`if false && …`), the targeted selection goes `24 passed` → **`16 passed, 8 failed`**, `EXIT=100`. The 8 include `issue_1767_…::intronic_region_declines_an_unknown_offset`, which is what proves the collapsed `has_unknown_offset()` screens were genuinely load-bearing on the new path and not merely redundant.

There were **no** Python tests for `classify_splice_variant` before this PR in either direction, so the binding's behaviour on a sentinel was unpinned; the 8 new ones close that, per the `src/python.rs` path instruction that behaviour reachable from Python needs a test under `tests/python/`.

## Unchanged on purpose

- **`IvsPos::is_deep_intronic` / `is_canonical_splice_site` and `IntronPosition`'s predicates still require the caller to screen.** They return `bool`, which has no way to say "no answer" — pushing the rule in would only move the fabricated answer from `DeepIntronic` to `false`. Their doc comment cross-referenced `from_offset` as an *analogy* ("as with … this reads a measured distance"), which this change falsified; it is rewritten to name the divergence and why. That is the only edit under `src/hgvs/`.
- **`IntronicConsequence` still screens in its position constructors**, not in a scalar entry point, because it exposes no public `from_offset` to move the rule into. Noted in the code so the asymmetry is not "cleaned up" later.
- **`ferro effect`'s batch mode exits 0 even when every line errored.** That is the #1764 shape in a different subcommand, it is pre-existing for every already-failing input class, and fixing it is not part of an approved API break. Recorded here rather than swept in.

## Review

`/coderabbitai-review` on the local diff, `profile: assertive`, walked against the six matching `path_instructions`. Applied: the falsified `IvsPos` cross-reference (outside the diff); `+?` added to the reachability test, which covered only `-?` against a file whose whole discipline is sentinel symmetry; consequence/impact assertions added to the neighbour cases in both the Rust and Python guards; the CLI comment extended to name the JSON and batch-mode consumers. Rejected: *"batch `ferro effect` should count failures and exit non-zero"* — real, pre-existing, and out of scope, so it is disclosed above instead; *"release the GIL"* on the PyO3 binding — the call is O(1) integer arithmetic; *"a panic can cross the FFI boundary"* — the added `expect`s are all in test code, and `Option` mapping introduces none.

## Verification

Every exit code read back from its own uniquely-named log rather than from a completion notification.

| command | exit |
|---|---|
| `cargo fmt --all --check` | `0` |
| `cargo clippy --all-features --all-targets -- -D warnings` | `0` |
| `cargo clippy --features dev --all-targets -- -D warnings` | `0` |
| `cargo check --features python` | `0` |
| `cargo doc --no-deps --features dev` | `0` — no warning from the new intra-doc links |
| `FERRO_SWEEP_SEEDS=full cargo ta` | `0` — **10621 run, 10621 passed**, 152 skipped, 0 FAIL |
| `scripts/run_oracle_suite.sh` | `0` — **10484 run, 10484 passed**, 289 skipped, 0 FAIL |
| `cargo test --doc --features dev` | `0` — 97 passed, 9 ignored (covers the new `classify_splice_variant` example) |
| `ruff check` / `ruff format --check` `python/ tests/python/` | `0` / `0` |
| `mypy python/ferro_hgvs/` | `0` |
| `pytest tests/python/` (against a `maturin develop --features python` build) | `0` — **905 passed**, 8 skipped |
| `scripts/check_representation_change.py` over the real changed-file list | `0` |

The oracle suite was run through `scripts/run_oracle_suite.sh` rather than as a bare armed run. That script reads its selection and flag set straight out of `ci.yml`, so it mirrors the `test-oracle` job including its `ORACLE_EXCLUDE` negation; a bare `FERRO_ASSERT_IDEMPOTENT=1` full-suite run is red on a clean `origin/main` for reasons that have nothing to do with this diff, so it would have said nothing.

Representation-Change: none. Nothing here renders an HGVS description — both functions classify a *consequence* from an integer offset, and the one edit under a watched directory (`src/hgvs/location.rs`) is a doc comment. `c.100-?` still parses and still renders as `c.100-?`, pinned by `a_parsed_unknown_offset_reaches_the_classifiers_as_a_sentinel`.

Related: #1826 asks the same question in wider terms as its item 2, and is not resolved by this PR — its item 1 was already addressed in #1806, and it should be reviewed separately for whatever it still holds.

### DIFF
--- a/python/ferro_hgvs/__init__.pyi
+++ b/python/ferro_hgvs/__init__.pyi
@@ -1377,8 +1377,17 @@ class EffectPredictor:
         """Classify an indel by frame effect."""
         ...
 
-    def classify_splice_variant(self, offset: int) -> ProteinEffect:
-        """Classify a splice site variant by distance from splice site."""
+    def classify_splice_variant(self, offset: int) -> ProteinEffect | None:
+        """Classify a splice site variant by distance from splice site.
+
+        Returns ``None`` when the offset is unknown -- ``+?`` / ``-?``, carried
+        as the sentinels ``2**63 - 1`` and ``-2**63``. Such a description states
+        no distance from the splice site, so there is nothing to classify.
+
+        Args:
+            offset: Distance from splice site (negative for acceptor, positive
+                for donor)
+        """
         ...
 
     def classify_utr_variant(self, is_5_prime: bool) -> ProteinEffect:
--- a/src/bin/ferro.rs
+++ b/src/bin/ferro.rs
@@ -2712,7 +2712,38 @@ fn run_effect(
                     // Check for intronic offsets
                     if let Some(start_pos) = cv.loc_edit.location.start.inner() {
                         if let Some(offset) = start_pos.offset {
-                            predictor.classify_splice_variant(offset)
+                            // `classify_splice_variant` declines an unknown
+                            // offset (`c.100-?`), whose whole content is that
+                            // the distance is not stated (#1841). Refuse the
+                            // description rather than print a consequence, in
+                            // the same register as the unsupported-type arm
+                            // below — printing `intron_variant`/MODIFIER, which
+                            // is what this arm did before, reads as a
+                            // prediction to anything parsing the output.
+                            //
+                            // This is a CLI output change, so state what it
+                            // does to each consumer. Single-variant mode:
+                            // stdout gains no line (for `-f json`, no JSON
+                            // object at all) and the process exits non-zero,
+                            // where it previously printed
+                            // `intron_variant (MODIFIER)` and exited 0. Batch
+                            // mode (`--input` / stdin): the two loops below
+                            // already catch a per-line `Err`, report it to
+                            // **stderr** as `ERROR: <line> - <e>` and continue,
+                            // so one such line does not truncate the stream —
+                            // the #1764 failure mode is not reintroduced here.
+                            match predictor.classify_splice_variant(offset) {
+                                Some(effect) => effect,
+                                None => {
+                                    return Err(format!(
+                                        "Cannot predict an effect for an unknown intronic \
+                                         offset (`+?`/`-?`), which states no distance from \
+                                         the splice site: {}",
+                                        v
+                                    )
+                                    .into());
+                                }
+                            }
                         } else {
                             ProteinEffect {
                                 consequences: vec![Consequence::CodingSequenceVariant],
--- a/src/convert/noncoding.rs
+++ b/src/convert/noncoding.rs
@@ -267,19 +267,28 @@ pub enum IntronicRegion {
 }
 
 impl IntronicRegion {
-    /// Create from a *measured* offset value.
+    /// Classify an offset into its distance band, or decline.
     ///
-    /// The argument is a distance from the exon boundary, so the unknown-offset
-    /// sentinels are out of domain: screen them with
-    /// [`CdsPos::has_unknown_offset`] / [`TxPos::has_unknown_offset`] before
-    /// calling, as [`IntronicRegion::from_cds_pos`] and
-    /// [`IntronicRegion::from_tx_pos`] do. The magnitude is taken with
-    /// `unsigned_abs`, so no input can overflow it (#1767); a sentinel passed
-    /// here anyway lands in `DeepIntronic`, which is a distance claim its input
-    /// does not support.
-    pub fn from_offset(offset: i64) -> Self {
+    /// Returns `None` for the unknown-offset sentinels (`+?` / `-?`, carried as
+    /// [`OFFSET_UNKNOWN_POSITIVE`] / [`OFFSET_UNKNOWN_NEGATIVE`]). Every variant
+    /// of this enum names a **distance band**, and a sentinel states that the
+    /// distance is unknown, so none of the four is a true answer for it.
+    ///
+    /// #1767 fixed the arithmetic here — the magnitude is taken with
+    /// `unsigned_abs`, so no `i64` can overflow it — but left the signature
+    /// promising an answer this function cannot always give: a sentinel landed
+    /// in `DeepIntronic`, a distance claim its input does not support. #1841
+    /// took that break. The precondition used to be documented ("screen with
+    /// `has_unknown_offset` before calling"); it is now in the type.
+    ///
+    /// [`OFFSET_UNKNOWN_POSITIVE`]: crate::hgvs::parser::position::OFFSET_UNKNOWN_POSITIVE
+    /// [`OFFSET_UNKNOWN_NEGATIVE`]: crate::hgvs::parser::position::OFFSET_UNKNOWN_NEGATIVE
+    pub fn from_offset(offset: i64) -> Option<Self> {
+        if crate::hgvs::parser::position::is_unknown_offset(offset) {
+            return None;
+        }
         let abs_offset = offset.unsigned_abs();
-        if abs_offset <= 2 {
+        Some(if abs_offset <= 2 {
             Self::CanonicalSpliceSite
         } else if abs_offset <= 20 {
             Self::ExtendedSpliceRegion
@@ -287,28 +296,29 @@ impl IntronicRegion {
             Self::NearSpliceSite
         } else {
             Self::DeepIntronic
-        }
+        })
     }
 
     /// Create from a CDS position.
     ///
-    /// Returns `None` for an unknown offset — every variant of this enum names
-    /// a distance band, and a sentinel supports none of them (#1767).
+    /// `None` when the position is not intronic, and when its offset is unknown.
+    /// The unknown-offset rule is [`IntronicRegion::from_offset`]'s own since
+    /// #1841, so this no longer screens for it separately — one rule, one site,
+    /// which is the discipline #1767 applied to `is_unknown_offset` itself.
+    ///
+    /// The sibling ladder [`IntronicConsequence`] still screens in *its*
+    /// position constructors rather than in a scalar entry point, and that
+    /// asymmetry is deliberate rather than a missed cleanup: it exposes no
+    /// public `from_offset`, so there is no scalar site to move the rule into.
     pub fn from_cds_pos(pos: &CdsPos) -> Option<Self> {
-        if pos.has_unknown_offset() {
-            return None;
-        }
-        pos.offset.map(Self::from_offset)
+        pos.offset.and_then(Self::from_offset)
     }
 
     /// Create from a transcript position.
     ///
     /// Declines an unknown offset, as [`IntronicRegion::from_cds_pos`] does.
     pub fn from_tx_pos(pos: &TxPos) -> Option<Self> {
-        if pos.has_unknown_offset() {
-            return None;
-        }
-        pos.offset.map(Self::from_offset)
+        pos.offset.and_then(Self::from_offset)
     }
 }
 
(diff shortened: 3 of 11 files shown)


nh13:
@coderabbitai pause

Issue #860, opened 2026-06-26
policy(project): preserve LRG_ transcript/protein namespace on output


nh13:
Part of #326 (protein_description / coding_protein_descriptions; spec-arbitration sweep 2026-06-25). Bucket: **policy decision** (both forms spec-valid).

For `LRG_` inputs, ferro emits the underlying RefSeq `NM_`/`NP_` accession rather than the `LRG_t`/`LRG_p` aliases the input used:
- `LRG_1t1:c.266G>T` → expected `LRG_1p1:p.(Gly89Val)`, got `NP_000079.2:p.(Gly89Val)`.
- `LRG_24:g.5526_5533del` → expected `LRG_24t1:c.127_134del` / `LRG_24p1:p.(…)`, got `LRG_24(NM_001114101.3):…` / `LRG_24(NP_001107573.1):…`.

Both `LRG_p`/`LRG_t` and `NP_`/`NM_` are spec-valid public references (`general.md` L16), so this is a namespace-preservation policy call, parallel to the #121/#500 selector-preservation policy. **Decision:** preserve the input's `LRG_` namespace on output vs. canonicalize to RefSeq (could collapse into the accepted-divergence sweep if the bare-NP "any valid public ref is acceptable" rationale is extended). 2 rows (protein 8 + coding 15).

Issue #737, opened 2026-06-17
Genomic-axis projection of pinned NG_(NM_) corpus rows: off-by-N coordinates + insertion declines (surfaced by the genomic conformance snapshot)


nh13:
## Summary

While building the hermetic **genomic** conformance gate (the last axis of the pinned-snapshot arc, #719), I confirmed that the #728 version-independent `NG_` placements **structurally unblock** genomic projection for corpus rows whose `NG_` version is absent from the alignment snapshot — but the **projected coordinates don't match the corpus expectations**, so the genomic axis is not yet gate-green. This issue captures two distinct findings for triage. Root cause is **not yet nailed**; this is a writeup of the investigation, not a confirmed single bug.

## Setup to reproduce

1. Derive the corpus `NG_` placements (needs #728 / PRs #731 + #732):
   ```
   cargo run --features dev --example derive_ng_placements -- \
     --manifest <ref>/manifest.json --accessions corpus_ng.txt \
     --out corpus_derived.json --builds GRCh38,GRCh37
   ```
   (All 11 corpus genomic NGs derive on both builds, 0 declined.)
2. Point a manifest's `derived_refseqgene_placements` at `corpus_derived.json`, then `ferro project --axis g`.

## Finding 1 — off-by-N genomic coordinate (NOT version-fuzzy, NOT a placement error)

```
NG_012337.1(NM_012459.2):c.12_13delCA
  ferro    -> NG_012337.1:g.4910_4911del
  expected -> NG_012337.1:g.4913_4914del      (off by 3)
```

What is **ruled out**:
- **Not a missing/placement issue.** Without the derived placement, ferro correctly *declines* (`no g. representation`); with it, ferro projects — so the chain works, the coordinate is just wrong.
- **Not version-fuzzy.** cdot carries the **exact** `NM_012459.2` (alongside `.3`/`.4`); `.2` and `.4` project to *different* coordinates (`g.4910` vs `g.4864`), confirming ferro uses `.2`'s own frame. (`.2` `start_codon=30`, `.4` `start_codon=33`.)
- **Not the NG placement.** The `NG_012337.1` placement is validated by full-sequence comparison against the genome (#728's guard), so `NG_↔NC_` is exact.

So the off-by-3 lives in the `c. → NC_` (cdot `NM_012459.2` transcript→genome alignment) or `NC_ → NG_` composition. **Hypotheses (need triage):**
[...]

PR #1844, opened 2026-08-13
fix(project): stop renaming the n. axis onto an LRG transcript with a different frame

nh13:
Refs #1712.

#1712 measured **21 of 681** projected `n.` axes moving under bare re-normalization. This PR closes **one** of them — the one that is not a spelling question — and leaves the other twenty pinned, because closing them settles a question the ruling ledger deliberately leaves open. Both halves are stated below.

## The row this closes, and why it is not what the issue called it

#1712 filed this row as "the output is not 3'-shifted", the only rule 2 miss among the 21:

```
LRG_24:g.5526_5533del : LRG_24t1:n.245_252del -> LRG_24t1:n.246_253del
```

The shift was the symptom. Measured against the prepared GRCh38 reference:

| accession | length | `cds_start` | bases at `n.245_252` |
|---|---:|---:|---|
| `NM_001114101.3` (what the projection resolved) | 1181 | 119 | `GGGGCACC` |
| `LRG_24t1` (what it was renamed to) | 1191 | 127 | `GCCTGCCC` |

`LRG_24t1` carries eight extra 5' bases. The #860 LRG namespace echo in `frame_projection_owned` renamed the axis's accession while keeping its transcript coordinate, so the emitted description **denotes different bases** — a rule 1 defect, with the failed 3' shift downstream of it. `n.245_252` on `LRG_24t1` is the same eight bases as `n.253_260`, and the normalizer was 3'-shifting the wrong span rather than failing to shift the right one.

`LRG_24t2` is the same shape and worse: 1139 bases against `NM_172369.5`'s 1158, and **no `cds_start` at all** in the prepared reference. Its corpus row was a *stable* wrong answer rather than a non-idempotent one, so #1712's residue never contained it — but it is the same defect, and measured the same way: `n.222_229` reads `TACGACGG` on `LRG_24t2` and `GGGGCACC` on `NM_172369.5`.

## The fix

`c.` is CDS-relative and `p.` is residue-relative, so a differing 5'UTR length leaves both alone — which is why the `c.` axis of that very row (`LRG_24t1:c.127_134del`) is correct and idempotent today. A transcript coordinate is an offset into the served sequence, so `n.` is the only axis the rename can misplace.

So the rename is gated on the two accessions serving byte-identical bases. When they do not, the axis keeps the transcript it was measured against and stays framed under the LRG parent by `relabel_under_parent` — rendering `LRG_24(NM_001114101.3):n.245_252del`, which is the shape **22 of the 29** LRG-parented `n.` axes already take.

Sequence equality rather than a `(cds_start, len)` proxy: two same-length records with different bases would pass the proxy, and an unavailable sequence returns `false` because an unverifiable frame must not be treated as a matching one.

Declining rather than re-deriving the coordinate is deliberate: `LRG_24t2` has no CDS anchor to reframe through, so any offset would be invented.

## Unchanged on purpose

- **`c.` and `p.` keep the #860 echo**, including on the very rows whose `n.` axis stops taking it. `LRG_24t1:c.127_134del` is pinned alongside the new `n.` string in the same assertion.
- **An LRG transcript that does match still gets renamed.** `LRG_303t1` and `NM_000106.6` are byte-identical — asserted on the sequences, not inferred from `(1588, cds_start = 20)` agreeing — and its **5** corpus rows do not move. Of the 7 axes rendered in the bare `LRG_<n>t<k>` form, 5 stay and 2 move. The unit test's first arm is exactly this case, so the fix cannot degenerate into a blanket refusal that quietly retires #860 for this axis.
- **The other twenty rows of #1712 are untouched**, and the pin now says so in its own comment.

## What is NOT fixed, and the ruling it waits on

The remaining twenty have one cause: `noncoding_from_coding` reframes coordinates and clones the `c.` edit verbatim, so `n.` inherits a form the **coding** frame chose *because it declares a reading frame*. Two clauses do that work on `c.` and neither reaches `n.`: `repeated.md`'s codon gate (16 rows — `n.3709_3710del` -> `n.3708_3710T[1]`) and the `delins.md:18` / `general.md:35` codon carve-out (4 rows — `n.309delinsCTGA` -> `n.[308_309insCT;310dup]`).

The obvious remedy is to normalize the derived axis on its own reference, symmetric with the genomic axis's `normalize_or_fallback` (#737). **Measured: it takes the residue 21 -> 0.** It is not applied here because the decided record `projection-codon-exception-is-decided-by-the-rendered-axis` says, in its own words:

> The `n.` axis is NOT settled by this record: it declares a non-coding DNA reference and so carries no frame, yet projection renders it merged because the axis is derived from the `c.` form by CDS-offset arithmetic rather than normalized on its own reference […] Which side of that is wrong is an open question, deliberately left open.

Applying it turns three committed guards red, and each is a pin of the current answer rather than a stale expectation:

| guard | what it pins today |
|---|---|
| `lenient_overlap_and_projection_split::the_codon_delins_merges_only_on_an_axis_that_declares_a_frame` | `split == ['g']`; its doc comment records the `n.` row as **"pinned as OBSERVED, not as adjudicated"** and names this same open question |
| `lenient_overlap_and_projection_split::the_same_codon_delins_renders_alike_whichever_axis_it_is_authored_on` | `('n', "tx:n.4_6delinsATA")` as the merged control |
| `projection::transcript_cache_collapses_repeat_lookups` | one `get_transcript` call across three variants; normalizing per axis takes it to ten, since the normalizer's fetches do not go through the projector's cache |

So the question for the operator is narrow: **on the `n.` axis, is the projector wrong (it should re-derive on the `n.` reference, splitting and repeat-maximizing like `g.`) or is the normalizer wrong (it should treat `n.` on a coding transcript as frame-aware)?** The second direction re-opens #1241, whose fix made `codon_frame_aware` false for `Tx` precisely so `n.` and `r.` spellings of one non-coding change would not settle on two strings.

## Representation stability

Measured on the mutalyzer corpus against the prepared GRCh38 reference, `project_variant_all` over every `to_test` row, all 681 `n.` axes compared:

| | rows |
|---|---:|
| compared | 681 |
| moved | 2 |
| kind of move | accession only — `LRG_24t1:n.245_252del` -> `LRG_24(NM_001114101.3):n.245_252del`, `LRG_24t2:n.222_229del` -> `LRG_24(NM_172369.5):n.222_229del` |
| already unstable | 1 of 2 (the `LRG_24t1` row is #1712's; the `LRG_24t2` row was a stable wrong answer) |

No coordinate changes and no edit changes. Both moved rows previously named a sequence they did not denote, so this is a correction rather than churn. `axis_noncoding`'s corpus expectations are unaffected: the fixture lists **zero** `LRG_…:n.` expectations, so no subset check gains or loses a member.

`examples/dump_normalized_corpus.rs` reports **0 rows moved, and that zero is structural** — not evidence. The harness contains zero occurrences of `LRG` and never constructs a `VariantProjector` at all; it drives `Normalizer` over synthetic `NC_TEST.1`/`NM_TEST.1` accessions. The property this change keys on is "the projection is framed under a bare `LRG_` genomic parent", which that generator cannot vary. The 681-axis measurement above is the one that can move, and does.

## Which of these tests is pre-merge coverage

Stated because `.coderabbit.yaml`'s `tests/it/{mutalyzer_*.rs,…}` instruction asks for it, and it is easy to get wrong:

- **`project::projector::tests::…::frame_projection_owned_renames_the_noncoding_axis_only_on_a_shared_frame` is the gate.** It is a `MockProvider` unit test built programmatically, needs no manifest, and runs on PR CI. Its two arms are the discriminating pair — identical sequences must still rename, differing ones must not — so it fails both by regressing the fix and by over-generalising it into a blanket refusal.
- **`mutalyzer_normalize_tests::lrg_noncoding_axis_names_the_transcript_it_was_measured_against` is NOT pre-merge coverage.** It resolves `FERRO_MANIFEST` and silently returns when unset, which is what PR CI does; only the nightly reference-aware job runs it. It is here because it states the defect against the real reference — including reading `LRG_24t1`'s and `NM_001114101.3`'s bases directly, so its premise is a fact about the reference rather than about the projector.

It is a unit test rather than one in `tests/it/projection.rs` because it isolates `frame_projection_owned` by handing it a `VariantProjection` directly; routing the case through `project_variant_all` would measure transcript enumeration and the reference's LRG records instead of the rename gate. Recorded in the test's own doc comment.

## Verification

Commands and the counts read back out of their own logs, never from a completion notification (one run here reported `exit code 0` while its log held `EXIT=100`):

- `axis_noncoding_idempotent`, manifest-backed: `681 tested, 20 non-idempotent` (was `681 tested, 21`), `EXIT=0`. `axis_genomic_idempotent` control on the same run: `214 tested, 0`.
- Manifest-backed conformance selection (`mutalyzer_*`, `biocommons_*`, `hgvs_rs_projection_tests`, `conformance`): `348 tests run: 348 passed`, `EXIT=0`.
- The new unit test is **sabotage-checked**: forcing `shares_transcript_frame` to return `true` fails it with `left: "LRG_24t1:n.5A>G"` / `right: "LRG_24(NM_001114101.3):n.5A>G"`, so it is testing the guard and not the scaffolding.
- `cargo fmt --all --check`: clean. `cargo clippy --all-features --all-targets -- -D warnings`: clean.

One note on the two generated spec artifacts: `spec_enumeration_tests` and `ruling_citation_currency` were red mid-run for reasons unrelated to this diff — `tests/fixtures/grammar/hgvs_spec_{normalization,enumeration}.json` had been generated earlier in the session while an experiment was in the tree. Regenerating both (`generate_spec_fixture`, `generate_spec_enumeration`) clears them; `24 tests run: 24 passed`, `EXIT=0`.

Representation-Change: 2 of 681 projected non-coding axis rows respell. Previously-accepted output; both moved rows previously named bases they did not denote, so this is a correction rather than churn.

### DIFF
--- a/src/project/projector.rs
+++ b/src/project/projector.rs
@@ -1606,9 +1606,46 @@ impl<P: ReferenceProvider + Clone> VariantProjector<P> {
                 // The rna axis can be an allele needing apply_rna_framing-style
                 // recursion; no #860 corpus row exercises rna under an LRG parent,
                 // so it is intentionally out of scope here.
-                for field in [&mut result.coding, &mut result.noncoding] {
-                    if let Some(v) = field.as_mut() {
+                if let Some(v) = result.coding.as_mut() {
+                    Self::set_bare_accession(v, lrg_tx_acc.clone());
+                }
+                // The `n.` axis is the only one whose numbers are stated in the
+                // transcript's *own* frame, so it is the only one this rename can
+                // misplace (#1712). `c.` is CDS-relative and `p.` is
+                // residue-relative, and both survive a differing 5'UTR length; a
+                // transcript coordinate does not.
+                //
+                // Measured: `LRG_24t1` is 1191 bases with `cds_start = 127` while
+                // the `NM_001114101.3` it resolves to is 1181 with
+                // `cds_start = 119` — an eight-base 5' extension. Renaming the
+                // accession while keeping the number turned
+                // `NM_001114101.3:n.245_252del` (which deletes `GGGGCACC`) into
+                // `LRG_24t1:n.245_252del`, which on the LRG transcript's own bases
+                // deletes `GCCTGCCC`. That is a wrong denotation, not a spelling
+                // choice, and the non-idempotency #1712 measured
+                // (`n.245_252del` -> `n.246_253del`) was only its symptom.
+                //
+                // So rename only when the two sequences are byte-identical, which
+                // is the precondition under which a transcript coordinate carries
+                // over unchanged. Otherwise leave the axis on its resolved
+                // transcript accession — still framed under the LRG parent by
+                // `relabel_under_parent` above, rendering
+                // `LRG_24(NM_001114101.3):n.245_252del`, which is the shape the
+                // other LRG-parented `n.` rows already take.
+                //
+                // Declining is deliberate rather than re-deriving the coordinate:
+                // `LRG_24t2` has no `cds_start` at all in the prepared reference
+                // (and is 1139 bases against `NM_172369.5`'s 1158), so there is no
+                // CDS anchor to reframe through and any offset would be invented.
+                if let Some(v) = result.noncoding.as_mut() {
+                    if self.shares_transcript_frame(&lrg_tx, v) {
                         Self::set_bare_accession(v, lrg_tx_acc.clone());
+                    } else {
+                        log::trace!(
+                            "frame_projection_owned: {lrg_tx} does not share a transcript \
+                             coordinate frame with the resolved transcript; keeping the \
+                             resolved accession on the n. axis rather than renaming it (#1712)"
+                        );
                     }
                 }
                 if let (Some(protein), Some(lrg_p)) =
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #1845, opened 2026-08-13
fix(normalize): clamp the 3' rule at an exon junction from both sides

### LINKED ISSUE #1621: Normalizer: clamp the 3' rule at an exon/exon junction from BOTH sides — c.3922dup must converge on c.3921dup

nh13:
`general.md:44` exempts deletions and duplications around exon/exon junctions from the 3' rule on `c.`/`n.` references. Ferro applied that exemption in **one direction only**: a description approaching a junction stopped at it, while one already spelled *past* it was left standing. So two spellings of one transcript sequence were two fixed points, projecting ~2,790 bp apart.

The operator ruling `exon-junction-dup-converge-from-the-far-side` (decided 2026-08-10, landed in #1623) reads the exemption as a clamp binding in **both** directions. This is its implementation half — the issue #1623's body named as "#1621", which existed as an issue and never as a PR.

## The rule

The canonical position is the most 3' position that does not cross an exon/exon junction, **reached from either side**. Implemented as: find where the description's equivalence run *begins* in full transcript space, then clamp the 3' rule inside the exon holding that anchor. Both spellings of one run share an anchor, so both reach the same answer — that is the "from either side".

In spliced transcript space a junction is not a gap: the base immediately 5' of an exon's first base is the previous exon's last base, which is why `c.3921` and `c.3922` are one contiguous `T` run and only the exon annotation distinguishes them.

## Where the clamp is asked, and two placements that were measured and rejected

It is asked **once, on the finished description**, at the seam every public normalization passes through. Both alternatives were implemented and measured before being discarded, and both failures are recorded in the code:

- **On the input.** `normalize_cds` has several producers of a del/dup spelling and only one is the input's own edit kind: a `delins` whose payload shares an affix with its reference *reduces* to a pure deletion inside `normalize_na_edit` (#1185), after any input-side check has run. `c.20_22delinsA` therefore emitted `c.21_22del` — a description the clamp itself would move. That is a **rank-1 idempotency regression**, measured as `non_idempotent_outputs` 4 → 6 in the 58,552-output conformance census.
- **Inside `normalize_cds`/`normalize_tx`.** Those are recursive (the `conversion` rewrite, special-position resolution, …), so a clamp applied there fires on *intermediate* spellings as well as the answer. Measured: it moved cis-allele rows whose finished output the clamp declines, cost one converged family (`s00-n3p-junction-2-del-del-p1-sep0`) and raised another's arity.

An earlier in-place rewrite of the transcript coordinates was also rejected: #670's `normalize_boundary_spanning_cds` reads the original `variant`/`start_pos`/`end_pos`, so rewriting only `tx_start` made `LRG_199t1:c.3922dup` fire the junction-crossing continuation against **exon 28's** genomic neighbourhood while the shuffle had already moved to exon 27.

## Scope

Deletions and duplications, `c.` and `n.`, 3'-direction only, refused across a CDS axis change. `r.` is deferred with the ruling's own "(and `r.` until upstream PR #266 lands)". A cis **member** is not clamped individually — only a whole-variant result is.

**The bound that matters:** a description that can still shift 3' inside its own exon is *not* pulled back. Without that condition the rule inverts into the defect it prevents, moving a description 5' *across* a junction. `NM_001234.1:c.13del` → `c.26del` (#1450) is the case that sets it: its run does reach back over the exon-1/exon-2 junction, but its own exon-confined answer is exon 2's **last** base, so it stays. An unconditioned anchor rule drags it to `c.11del`, one exon away — measured, then fixed.

**Residue, left deliberately.** A run whose portion inside the far exon is longer than the description can be pinned at (`c.3922`/`c.3923` both `T`, say) is not reached: the far-side spelling shifts within its own exon and the pair does not converge. The ruling's words arguably cover it; its worked example, its grounds and every corpus row do not, and widening re-opens the #1450 regression unless some further discriminator is found. Recorded on `exon_junction_anchor` rather than guessed at.

## Tests

- `spec_worked_examples::the_decided_target_is_convergence_on_the_near_side` — the ruling's own acceptance criterion, **`#[ignore]` removed**; `LRG_199t1:c.3922dup` → `c.3921dup` in default, lenient and strict. Verified failing first, for the right reason (`left: Ok("LRG_199t1:c.3922dup")`).
- `the_exon_junction_pair_does_not_converge_and_that_is_recorded`'s `assert_ne!` is **deleted rather than inverted**, on its own instructions. The base-identity premise it also checked survives as `the_exon_junction_pair_is_two_spellings_of_one_sequence` — that is evidence for the ruling, not a claim about ferro, and nothing else checks it.
- `tests/it/issue_1621_exon_junction_far_side.rs` — the rest of the ruling's scope on a synthetic **two-exon** transcript built programmatically: deletions as well as duplications, `n.` as well as `c.`, each asserting **both** directions (the near side stays *and* the far side converges), plus a fixed-point check and a negative control. `cds_start` is 5, not 1. `the_fixture_really_straddles_a_junction` asserts the geometry directly — a single-exon fixture with `CDS_START = 1` cannot reach this class at all (#1478).
- Fixture rows moved per the issue's acceptance list: `spec-worked-examples/cases.json` (three mode outputs) and `case-harvest/cases.json`. The diverging set is now **empty** and the conformant-row census moves 11 → 12.

## Representation-Change

Representation-Change: 13 of 85,642 rows move (0.0%) on the designed shape-family corpus — `coincident_bounds` 9, `repeat_expansion` 4 — and **all 13 were previously not fixed points**, so 0 previously-accepted rows move and there is no consumer migration. Quote that rate as of the affected families, not repo-wide. Independently, over `spec_conformance_axis`'s 58,552 outputs exactly four family instances change, all `junction-1-del-del`: `s01-c3p-…-sep{0,1}` fall one arity each as the far-side spelling `c.21_22del` converges onto `c.19_20del`, and `s01-c3m-…-sep{0,1}` re-spell that same output as `NC_SYNTH.1(NM_TEST.1):c.18_20+2A[3]`. No family loses or gains convergence, `converged` is unchanged at 11,016, no failure counter rises, and the 5' census does not move at all. The spec's own row `LRG_199t1:c.3922dup` → `c.3921dup` moves its projected genomic form ~2,790 bp (that figure is the ruling's, relayed, not re-derived here).

## Census re-blessing

`spec_conformance_axis`'s 3' pin moves three numbers, all improvements or conformant re-renderings, disclosed in a new `RE-BLESSED (6 of 6)` section with the row-level diff (full divergence lists dumped either side with `MAX_DIVERGENCES` raised locally, then restored):

| figure | was | now | direction |
|---|---|---|---|
| `outputs_intronic_under_a_genomic_wrapper` | 389 | **391** | +2 — conformant, not a defect counter |
| `converged` | 11,016 | **11,016** | unchanged |
| `split_two` | 673 | **674** | +1, from `split_three` |
| `split_more` | 29 | **28** | −1, to `split_three` |

The +2 is the minus-strand pair whose pulled-back description then continues into the intron via #670. Both clauses apply and compose: `general.md:44` forbids the exon/**exon** shift, the exception 3' rule permits the exon/**intron** one. Nothing is promoted to `spec_corpus_regressions.rs` because nothing newly fails.

## Verification

Full suite `10,845 passed, 0 failed, 151 skipped` (`EXIT=0`, read from the log, with `FERRO_MANIFEST` set). `scripts/run_oracle_suite.sh` `10,484 passed` (`EXIT=0`). `cargo fmt --check` and `cargo clippy --features dev --all-targets -- -D warnings` clean.

Closes #1621

### DIFF
--- a/src/conformance/spec_worked_examples.rs
+++ b/src/conformance/spec_worked_examples.rs
@@ -57,10 +57,13 @@ impl Fixture {
 /// input-hygiene modes.
 ///
 /// Three outputs rather than one because the modes do not agree here, and which
-/// mode produces which answer *is* the finding: the single divergence from the
-/// spec belongs to the lenient preprocessor's W3013 repair, not to the
-/// normalizer, and strict mode rejects two of the spec's own inputs outright.
-/// One pinned output would have hidden both facts.
+/// mode produces which answer *is* the finding: strict mode rejects two of the
+/// spec's own inputs outright, and while the corpus still held divergences each
+/// was attributable to a named stage — the lenient preprocessor's W3013 repair
+/// (closed by #1631), and the normalizer's half-applied exon-junction clamp
+/// (closed by #1621). One pinned output would have hidden all of that. The
+/// diverging set is empty as of #1621; `the_divergence_set_is_exactly_the_one_recorded_conflict`
+/// is what holds it there.
 ///
 /// Every case also carries a [`Citation`], because a pinned output with no
 /// authority behind it is a change detector rather than a record: it tells a
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -486,6 +486,146 @@ fn edit_is_del_or_dup(edit: &NaEdit) -> bool {
     )
 }
 
+/// The 5' anchor of a del/dup's equivalence run, when that run continues back
+/// across an exon/exon junction (#1621).
+///
+/// # The rule
+///
+/// `general.md:44` exempts deletions and duplications around exon/exon
+/// junctions from the 3' rule on `c.`/`r.`/`n.` references. The operator ruling
+/// `exon-junction-dup-converge-from-the-far-side` reads that exemption as a
+/// clamp that binds in **both** directions: the canonical position is the most
+/// 3' position that does not cross a junction, *reached from either side*. A
+/// description approaching the junction stops at it — which is what
+/// [`Boundaries`] built from `axis_info.exon` already does — and one already
+/// spelled past it is **pulled back** to the clamp, which is what this function
+/// supplies. `DNA/duplication.md` states the outcome three times (`:26`, `:60`,
+/// `:148`), the last giving the reason: `c.3922dup` translated back to a
+/// genomic position lands "at the wrong nucleotide, in the wrong exon".
+///
+/// # Why an anchor, rather than a step back over the junction
+///
+/// In *spliced* transcript space an exon/exon junction is not a gap — the base
+/// immediately 5' of an exon's first base is the previous exon's last base — so
+/// `LRG_199t1:c.3921` and `c.3922` are one contiguous `T` run and two spellings
+/// of one sequence. Only the exon annotation distinguishes them. The canonical
+/// answer is therefore obtained by finding where that run *begins*, in full
+/// transcript space, and clamping the 3' rule inside the exon holding **that**
+/// base. Both spellings share a run, so both reach the same answer; that is the
+/// "from either side" in the ruling.
+///
+/// Returning the anchor rather than "the previous exon's last base" is what
+/// makes the rule compose when a run spans three or more exons, and what makes
+/// the result a fixed point: re-normalizing the answer finds the same anchor,
+/// inside the same exon, so no second pull-back fires.
+///
+/// # The scope, and the case that sets it
+///
+/// **A description is pulled back only when it is pinned at its exon's first
+/// base with nowhere to shift inside that exon.** Without that condition the
+/// rule inverts into the defect it is supposed to prevent: it would move a
+/// description *across* a junction, 5'-ward, which is exactly what
+/// `general.md:44` forbids.
+///
+/// `NM_001234.1` in `MockProvider::with_test_data` is the case that shows it —
+/// a 25-base `G` run spanning `c.9`-`c.33` across two junctions, with exons at
+/// transcript 1-15 / 16-30 / 31-44. `c.13del` sits mid-run inside exon 2 and
+/// 3'-shifts to exon 2's **last** base, `c.26del` (#1450 pins it). Its run does
+/// reach back over the exon-1/exon-2 junction, so an anchor rule with no
+/// further condition drags it to `c.11del` — one exon 5' of where the 3' rule
+/// put it. The condition below declines it: `c.26` is not exon 2's first base,
+/// so the description is resting on the near side of the *next* junction and
+/// stays there.
+///
+/// `c.3922dup` passes the condition because exon 28's first base is where it
+/// rests and where it started: `c.3923` is not `T`, so the exon gives it no 3'
+/// travel at all, and the only reason it is spelled in exon 28 rather than
+/// exon 27 is the junction it is hanging off. That is what "past the clamp"
+/// means in the ruling.
+///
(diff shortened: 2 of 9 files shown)


nh13:
@coderabbitai pause

PR #1847, opened 2026-08-13
fix(reference): model the LRG <diff> list so indel-bearing parents project exactly

### LINKED ISSUE #1833: reference: model the LRG <diff> list so indel-bearing parents project exactly instead of being refused past the affine horizon
Residue of #1499, split out so it does not close with PR #1830.

#1830 stopped `VariantProjector::project_to_genomic` mis-placing LRG parent coordinates by narrowing an indel-bearing LRG placement to its **affine-exact prefix**: everything the placement still accepts it maps exactly, and the rest is declined. That is correct but not complete — the coordinates past the horizon are refused rather than answered, and the data to answer them is already on disk.

## What is still refused

Measured on the prepared reference (identity `aa8b3246d83055cc`), over all 1294 LRG accessions × {`LRG_<N>t1`, `LRG_<N>t2`} × {`c.100del`, `c.1000del`} = 5176 projection rows:

- **51 rows, across 29 accessions**, changed from a rendered `g.` description to `unavailable` in #1830.
- All 51 sat past their record's affine horizon, and **49 of them named a demonstrably wrong base** — measured drift, in bases:

  ```
  -8: 2   -5: 1   -1: 17   0: 2   +1: 14   +2: 4   +3: 2   +4: 2   +5: 2   +6: 3   +14: 1   +23: 1
  ```

  (the two at 0 sat past a cancelling indel pair, so they happened to land right while the map through them is not exact.)

Refusing them is strictly better than the drifted answers they replaced. Answering them correctly would be better still, and would also restore whole-record coverage for the 46 indel-bearing accessions — for `LRG_292` the affine-exact prefix is only the first 13168 of 193689 bases, so essentially all of BRCA1's CDS is now on the refused side.

## Why it was not done in #1830

The LRG XML already carries the alignment exactly: alongside the single `<mapping_span>` it lists `<diff>` elements, vocabulary closed at `mismatch` / `lrg_ins` / `other_ins` (verified over all 1294 records — 479 / 63 / 47 in the main-assembly mappings). Honouring them means carrying a gap list beside the affine, which is a **data-model change to `GenomicPlacement`**:

- it is a public struct with public fields, constructed at **47 literal sites** (35 under `src/`, 12 under `tests/` + `examples/`), so adding a field is a mechanical edit at every one of them **and** a public-API break for any external constructor;
- `parse_lrg_main_assembly_placement` would grow from a coordinate parser into an alignment parser, and `nc_to_parent`/`parent_to_nc` from an affine into a gap-aware walk.

That is a design decision rather than a defect fix, so #1830 left it open deliberately.

## What the data looks like, before anyone starts

A throwaway reconstruction of each indel-bearing record from the chromosome plus its diff list reproduced **39 of 46 byte-for-byte**. The 7 that did not are the edge cases an implementation has to handle, not evidence of missing data:

- **adjacent / overlapping diffs** — `LRG_1321` has an `other_ins` of 1000 `N`s at `NC_…:7083651-7084650` and an `lrg_ins` of 231 bases whose chromosome flanks are `7083650`/`7083651`, i.e. one flank sits *inside* the other diff's range. A naive "skip chromosome bases in an `other_ins`, emit `lrg_ins` after its preceding flank" walk loses the insertion entirely.
- **multi-base insertions on both sides in the same neighbourhood** — `LRG_792` has 19 indel diffs, several within a few hundred bases of each other, sizes 1–78.
- **`lrg_start`/`lrg_end` conventions differ by kind** — for `lrg_ins` they bound the unaligned LRG bases; for `other_ins` they are the *flanking* LRG coordinates. #1830's `affine_horizon` already encodes this distinction and its doc comment states it.

## Suggested shape

1. Parse the `<diff>` list into an ordered gap list on the placement (or a sibling type the placement holds), keyed by chromosome coordinate with a cumulative parent-frame offset.
2. Make `nc_to_parent`/`parent_to_nc` apply the cumulative adjustment. The narrowing #1830 added then becomes unnecessary for records whose diffs are fully modelled, and should be kept only as the fallback for a mapping the model still cannot express.
3. Measure the same 5176-row corpus before and after: the 51 rows should return as `ok`, and every returned coordinate should be checkable against the frozen `LRG_<N>g` record by window comparison (the method #1499 and #1462 both used).

## Representation stability

This would move rows **rejected → accepted**, which is the cheap direction — but the accepted values are new stored strings for anyone who was previously getting a refusal, and it would also have to demonstrate that no row `#1830` currently answers correctly moves at all. Quantify against the 1294-record LRG corpus, not the synthetic one: `examples/dump_normalized_corpus.rs` builds only `MockProvider` and can never move a row here, so a zero from it is structural.

Refs #1499, #1503, #1462, #1490.

nh13:
Closes #1833.

Representation-Change: 51 of 5176 LRG projection rows change status, all `unavailable` -> `ok`
  (2764 -> 2815 ok, 147 -> 96 unavailable under --error-mode lenient; 43 rows and 2489 -> 2532
  under the CLI-default strict mode, the 8-row difference being inputs strict refuses before a
  placement is consulted). 0 previously-answered row changes value, and 0 regress. Measured
  against origin/main AFTER #1830 merged: #1830 narrowed indel-bearing LRG placements to their
  affine-exact prefix and declined the rest; this models the alignment instead, so those rows are
  answered rather than declined. Direction: TOWARD the reference oracle -- all 51 recovered
  coordinates were adjudicated against the frozen LRG_<N>g records with an unshiftable
  substitution probe (11-base window of the record against the chromosome coordinate cdot alone
  resolves): 51 agree, 0 disagree. 49 of the 51 sat at a base the gapless affine got demonstrably
  wrong (drift -8..+23); the other 2 sat on a cancelling indel pair. The earlier "55 move / 0
  status" figure was measured against the pre-#1830 base and no longer describes this merge.

## The defect

An LRG↔chromosome alignment is not affine, and a single `<mapping_span>` does not make it one. Alongside the span an LRG record lists `<diff>` elements — vocabulary closed at `mismatch` / `lrg_ins` / `other_ins`, re-derived here over all 1294 records in the prepared reference (**479 / 63 / 47** in the main-assembly mappings, matching the issue). The two insertion kinds shift every coordinate past them, in opposite directions, and `GenomicPlacement` dropped them, so a coordinate drifted by the net indel count between the LRG origin and the projected position.

`LRG_292` (BRCA1) is the worked case: 193 689 LRG bases onto a 193 687-base chromosome span, three `lrg_ins` bases against one `other_ins`. Only the first **13 168** bases — 6.8 % of the record — were placed correctly, so essentially all of BRCA1's CDS was on the wrong side of it.

Measured on the prepared reference (identity `aa8b3246d83055cc`), by comparing an 11-base window of each frozen `LRG_<N>g` record against the chromosome bases its placement names:

| | `origin/main` | this branch |
|---|---:|---:|
| windows sampled (1294 placements) | 260 431 | 260 431 |
| windows where the record and the chromosome agree | 255 121 | **260 385** |
| windows that disagree | 5 309 | 30 |
| …of those, disagreeing at a column **no `mismatch` diff explains** | **5 295**, across 46 accessions | **0** |

The surviving 30 differ only at columns the record's own `mismatch` diffs name — a substituted base, which shifts no coordinate and is served from the LRG's own record anyway since #1490.

## The fix

`GenomicPlacement` gains a `gaps: Vec<AlignmentGap>` field: the span's indel `<diff>` list as an ordered gap model, each entry either `ParentOnly { len }` (LRG bases with no chromosome counterpart) or `ChromosomeOnly { len }` (the reverse). `nc_to_parent` / `parent_to_nc` walk it; a coordinate inside a run has no counterpart on the other side and is declined rather than approximated. `parent_end()` derives the parent's extent from the gap list instead of assuming the two frames have equal width (#1503's point, and the only remaining consumer of that arithmetic — `get_sequence_length` — now asks it).

The field is additive; the 47 literal construction sites take `gaps: Vec::new()`, which is exactly the previous behaviour. `mismatch` diffs are deliberately **not** modelled: 119 records carry mismatches and no indel, and keying on any `<diff>` would gap 165 records where 46 is correct.

Three things the issue named as the specification, and how each is handled:

- **`LRG_1321`'s colliding flanks.** Its `other_ins` of 1000 `N`s occupies `NC_000012.12:7083651-7084650`, and the `lrg_ins` of 231 bases abutting it lists chromosome flanks `7083650`/`7083651` — the second being the last base of the other run. The model **never reads an indel diff's `other_start`/`other_end`**; it positions off the LRG coordinates and lets the walk consume each side in turn, which reproduces the record exactly (`g.13189` → `7083650`, where the affine said `7084419`, 769 bases out).
- **`LRG_792`'s 19 clustered indels.** No new shape — the record that fails a model right about one gap and wrong about accumulating them. 13 measured `(parent, chromosome)` pairs are pinned through the dense 25 302–25 494 cluster, where the cumulative parent-only total runs 1 → 13 → 23 → 24 → 25 → 103 → 133 → 135 in 200 bases.
- **The per-kind `lrg_start`/`lrg_end` convention.** `AlignmentGap::parent_only` takes the first unaligned parent base (`lrg_ins`'s own bases); `AlignmentGap::chromosome_only` takes the flanking coordinate the run sits after (`other_ins`). The two constructors *are* the convention, so a caller cannot silently mix them.

### It fails closed, and that is checked rather than asserted

The parser refuses a placement whose gap list it cannot honour, in six ways: an unreadable or missing indel coordinate; a `lrg_end` below its `lrg_start`; a `<diff type>` this parser has never seen; two indel diffs sharing one `lrg_start`; a gap list whose walk does not cover the placed span (`gaps_are_consistent`); and a walk whose reconstructed parent extent disagrees with the span's own `lrg_end`. The last two are integrity assertions on real data — both hold for **all 1294** records — and filters only for data that does not exist yet.

`gaps` is a public field, so the walk itself is defensive too: it checks ordering **up front** rather than mid-walk. That is not pedantry — measured on a two-gap list with the entries reversed, parent 3 answered `1002` where the sorted list declines, because the blocks before the out-of-order entry had already been emitted.

## Deliberately not decided

**Two indel diffs sharing one `lrg_start` are refused, not ordered.** The two kinds read their anchor differently — a flank for `other_ins`, the first unaligned base for `lrg_ins` — so at equal anchors the readings genuinely disagree and either choice invents something the record does not state. **0 of 1294 records have that shape** (measured), so nothing is lost by declining, and inventing an order would be a ruling the issue does not make. Abutting runs of opposite kinds are a *different* thing and are ordinary: two records have them (`LRG_1321`, `LRG_1075`), their anchors differ, and both are pinned as passing fixtures.

## Unchanged on purpose

- **A `mismatch`-only mapping is not gapped** — `mismatch_diffs_alone_leave_the_placement_affine`.
- **An indel `<diff>` under a non-main mapping does not leak** into the GRCh38 placement; across the prepared reference the non-main mappings carry 967 `lrg_ins` / 860 `other_ins` the main-assembly ones do not.
- **A gapless placement is byte-identical to before** — both coordinate functions take an early branch that is the pre-existing arithmetic, which is what keeps every `NG_` placement and 1248 of the 1294 LRG records exactly where they were. `parent_to_nc`'s long-standing "treat `Strand::Unknown` as ascending" asymmetry with `nc_to_parent` is preserved rather than tidied.

## Relationship to #1830 (open, not merged)

#1830 narrows an indel-bearing placement to its affine-exact prefix and declines the rest. This models the alignment instead, so the whole span answers. **The narrowing would now remove correct coverage**, and the two branches touch the same function — whichever lands second needs a rebase that drops `affine_horizon` rather than merging it. This PR branches from `origin/main` and does not stack on #1830.

Relative to `origin/main` no row changes status (2947 `ok` / 2229 `unavailable`, before and after). Relative to #1830 the 51 rows it turns `ok → unavailable` are answered instead.

## Verification

All counts read back from the log's own `EXIT=` line, not from a completion notification.

```
FERRO_SWEEP_SEEDS=full cargo nextest run --features dev    # EXIT=0
scripts/run_oracle_suite.sh                                # EXIT=0
```

Reference-backed measurement, `FERRO_MANIFEST` = the prepared reference on scratch (identity `aa8b3246d83055cc`), run in a detached `origin/main` worktree and in this one with the same throwaway harness:

- **Projection corpus** — 1294 accessions × {`t1`, `t2`} × {`c.100del`, `c.1000del`} = **5176 rows**. 55 moved, 32 accessions, all `ok → ok`.
- **Adjudication of the 55** — for each, the chromosome coordinate both trees resolved, then an 11-base window of the `LRG_<N>g` record against the chromosome at the old and the new parent coordinate: **53** match only at the new one, **2** match at both (a repeat context an 11-base window cannot discriminate), **0** match only at the old one.
- **Window sweep** — the 5309 → 30 table above, plus the set comparison that is the actual "nothing correct moved" claim: of the branch's 30 disagreeing windows, **0 are windows `main` got right** (30 ⊂ 5309), and 5279 windows `main` got wrong are now right.

**A `dump_normalized_corpus` zero would be structural here and is not quoted.** That harness builds only `MockProvider`, which has no LRG XML and constructs no `GenomicPlacement` with a gap list, so it cannot move a row on this change whatever the number says.

**The new unit tests run on PR CI.** They are `--lib` tests over synthetic XML fixtures and do not resolve `FERRO_MANIFEST`, so unlike the reference-aware suites they are not nightly-only. Their expected coordinates are measured against the frozen `LRG_<N>g` records — an oracle independent of the code under test — and transcribed, never inferred.

Every new guard was sabotage-checked: dropping the gap list reddens 6, breaking `nc_to_parent`'s chromosome-only decline reddens 4, removing the extent assertion reddens 1, removing the ordering check reddens 1, and allowing a shared anchor reddens 2. An earlier tie-break for shared anchors was **removed** because its sabotage stayed green — nothing in the corpus exercised it, which is what turned it into the refusal above.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1168,7 +1168,7 @@ It also downloads genomic-parent placement data used to project transcript-coord
 - **RefSeqGene→genome alignments** (`GCF_*_refseqgene_alignments.gff3`, from NCBI's archived `alignments/ARCHIVE/all/` — the feed stopped updating in 2024; latest GRCh38 is the RS-109/p13 snapshot, valid for p14 since primary `NC_` accessions are unchanged) → manifest `refseqgene_alignments`; parsed by `MultiFastaProvider` into per-`NG_` `GenomicPlacement`. The corresponding GRCh37 snapshot (`GCF_000001405.25_105.*`) → manifest `refseqgene_alignments_grch37`, merged with the GRCh38 file so an `NG_` resolves to its build-appropriate placement (#653/#713).
 - **LRG XML** genomic mapping → per-`LRG_` `GenomicPlacement` (parsed on demand).
 
-`ReferenceProvider::genomic_placement` exposes these; `VariantProjector::project_to_genomic` composes the `NM_`→`NC_` (cdot) step with the `NC_`→`NG_`/`LRG_` affine transform to re-express coordinates in the parent frame. With no placement it declines rather than emit chromosome coordinates under the parent accession.
+`ReferenceProvider::genomic_placement` exposes these; `VariantProjector::project_to_genomic` composes the `NM_`→`NC_` (cdot) step with the `NC_`→`NG_`/`LRG_` transform to re-express coordinates in the parent frame. With no placement it declines rather than emit chromosome coordinates under the parent accession. That transform is **affine only when the placement carries no alignment gaps** — an LRG record lists `<diff>` elements alongside its `<mapping_span>`, and since #1833 the indel ones are carried on the placement and honoured, so 46 of the prepared reference's 1294 LRG records map through a gap list rather than a pure offset (#1499).
 
 ## Reading the spec
 
--- a/src/conformance/accession_inventory.rs
+++ b/src/conformance/accession_inventory.rs
@@ -558,6 +558,7 @@ mod tests {
                 nc_end: 200,
                 parent_start: 1,
                 strand: Strand::Plus,
+                gaps: Vec::new(),
             },
         );
 
--- a/src/project/projector.rs
+++ b/src/project/projector.rs
@@ -3131,10 +3131,16 @@ impl<P: ReferenceProvider + Clone> VariantProjector<P> {
     }
 
     /// Re-anchor an NC-frame genome variant into a genomic parent's own frame
-    /// using its [`GenomicPlacement`] (#480): apply the affine NC→parent
-    /// transform to the interval, and reverse-complement the edit when the
-    /// parent runs antiparallel to the chromosome. If an endpoint falls outside
-    /// the placed span the variant is returned unchanged (chromosome frame).
+    /// using its [`GenomicPlacement`] (#480): map each endpoint with
+    /// [`GenomicPlacement::nc_to_parent`], and reverse-complement the edit when
+    /// the parent runs antiparallel to the chromosome. If an endpoint falls
+    /// outside the placed span the variant is returned unchanged (chromosome
+    /// frame).
+    ///
+    /// The transform is affine only when the placement carries no alignment
+    /// gaps (#1833); an LRG whose record lists indel `<diff>`s is mapped through
+    /// its gap list instead, and an endpoint landing on a chromosome base the
+    /// parent does not contain declines like an out-of-span one.
     fn reanchor_genome_to_parent(
         gv: crate::hgvs::variant::GenomeVariant,
         placement: &crate::reference::GenomicPlacement,
@@ -7496,6 +7502,7 @@ mod tests {
                 nc_start: 1000,
                 nc_end: 1008,
                 strand: crate::reference::Strand::Plus,
+                gaps: Vec::new(),
             },
         );
         let vp = VariantProjector::new(projector, provider);
@@ -7528,6 +7535,7 @@ mod tests {
                 nc_start: 1000,
                 nc_end: 1008,
                 strand: crate::reference::Strand::Plus,
+                gaps: Vec::new(),
             },
         );
         provider.add_legacy_gene_model("GENE1", "NM_TX2.1");
@@ -7566,6 +7574,7 @@ mod tests {
                 nc_start: 1000,
                 nc_end: 1008,
                 strand: crate::reference::Strand::Plus,
+                gaps: Vec::new(),
             },
         );
         provider.add_legacy_gene_model("GENE1", "NM_TX2.1");
@@ -7609,6 +7618,7 @@ mod tests {
                 nc_start: 1000,
                 nc_end: 30000,
                 strand: crate::reference::Strand::Plus,
+                gaps: Vec::new(),
             },
         );
         // NG_900.1 parent sequence: bases 4300..=4320 (1-based) are the insert.
@@ -7651,6 +7661,7 @@ mod tests {
                 nc_start: 1000,
                 nc_end: 30000,
                 strand: crate::reference::Strand::Minus,
+                gaps: Vec::new(),
             },
         );
         let insert = "GTCCTGTGCTCATTATCTGGC"; // parent bases 4300..=4320
@@ -7689,6 +7700,7 @@ mod tests {
                 nc_start: 1000,
                 nc_end: 30000,
                 strand: crate::reference::Strand::Plus,
+                gaps: Vec::new(),
             },
         );
         // NG_900.1 parent sequence: bases 4300..=4320 (1-based) are the insert.
@@ -7732,6 +7744,7 @@ mod tests {
                 nc_start: 1000,
                 nc_end: 1008,
                 strand: crate::reference::Strand::Plus,
+                gaps: Vec::new(),
             },
         );
         let vp = VariantProjector::new(projector, provider);
@@ -7787,6 +7800,7 @@ mod tests {
                 nc_start: 1000,
                 nc_end: 1008,
                 strand: crate::reference::Strand::Plus,
+                gaps: Vec::new(),
             },
         );
         let vp = VariantProjector::new(projector, provider);
@@ -7860,6 +7874,7 @@ mod tests {
                 nc_start: 2000,
                 nc_end: 2010,
                 strand: crate::reference::Strand::Plus,
+                gaps: Vec::new(),
             },
         );
         let vp = VariantProjector::new(projector, provider);
@@ -10230,6 +10245,7 @@ mod tests {
                     nc_start: 1000,
                     nc_end: 1008,
                     strand: crate::reference::Strand::Plus,
+                    gaps: Vec::new(),
                 },
             );
             let vp = VariantProjector::new(projector, provider);
@@ -10471,6 +10487,7 @@ mod tests {
                     nc_start: 2000,
                     nc_end: 2010,
                     strand: crate::reference::Strand::Plus,
+                    gaps: Vec::new(),
                 },
             );
             let vp = VariantProjector::new(projector, provider);
@@ -10514,6 +10531,7 @@ mod tests {
                 nc_start: 1000,
                 nc_end: 1010,
                 strand: crate::reference::Strand::Plus,
+                gaps: Vec::new(),
             };
             // Genomic axis at chromosome 5000 — well outside the placed span.
             let genomic = HgvsVariant::Genome(GenomeVariant {
@@ -10772,6 +10790,7 @@ mod tests {
                 nc_start: 1000,
                 nc_end: 1010,
                 strand: crate::reference::Strand::Plus,
+                gaps: Vec::new(),
             };
             // The end boundary is a compound range, so `.inner()` is `None`:
             // there is no single position to re-anchor.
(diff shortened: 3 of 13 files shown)


nh13:
@coderabbitai pause

nh13:
⚠️ **Merge-order hazard with the sibling PR — please read before merging this one.**

This PR and its sibling are both `MERGEABLE`/`CLEAN` right now, and they are **semantically opposed**: #1830 narrows an LRG placement to its affine-exact prefix and declines past it, while #1847 models the `<diff>` gaps so exactly those coordinates are answered. One declines where the other answers.

They share `src/reference/multi_fasta.rs` and `src/reference/provider.rs`, and `git merge-tree --write-tree --name-only <1830-head> <1847-head>` exits **1** — the conflict is real, not `mergeStateStatus` lag. It just does not appear until the first of the two lands.

**Whichever merges second needs a rebase that DROPS `affine_horizon`, not a merge that keeps it.** Keeping both sides is the default outcome of ordinary conflict resolution and it is the failure: `affine_horizon` clips the coordinates #1847 teaches the provider to answer, so #1847 merges and its coverage silently does not arrive. A green suite will not witness that, because a decline is a legal answer nothing asserts against.

Full write-up, including what to verify on `main` afterwards: #1873.

nh13:
**Merge-order hazard, re-verified on the current heads (2026-08-14).** This PR (`9f011e5c`) and **#1830** (`d755463b`, `fix/issue-1499-lrg-projection-placement`) are semantically opposed and both read green. `git merge-tree --write-tree --name-only d755463b 9f011e5c` exits **1**, conflicting in exactly `src/reference/multi_fasta.rs` (11 hunks) and `src/reference/provider.rs` (1 hunk) — but each still merges into `origin/main` on its own at exit 0, so nothing in GitHub's state will stop the wrong order.

The opposition is stated by the code, not just the descriptions: on the same `LRG_292` fixture this branch asserts `p.parent_to_nc(160875) == Some(43057109)`, while #1830 asserts `None`.

**If #1830 merges first, this PR's rebase must *drop* `affine_horizon` and the narrowing block in `parse_lrg_main_assembly_placement`, not merge them.** Keeping the narrowing while dropping this branch's two integrity assertions (`gaps_are_consistent()` and the `parent_end()` check) is the silent outcome — the gap walk then answers inside a truncated span and every coordinate past the horizon declines, with CI green because the guards that would catch it left with the rebase.

Full analysis, including a runnable post-merge check that asserts the guard *count* rather than the colour, is in **#1873**. No preference expressed here — the choice between the two designs is the maintainer's.

nh13:
**Maintainer decision on #1873: this PR is the resolution to take.** Its answered set is a superset of #1830's — #1830 does not move the anchor when it narrows, and this PR's walk emits a first aligned block at that same anchor with `nc_consumed = 0`, so everything #1830 still answers, this answers at the same base.

What is knowingly given up: this PR declines the **whole** placement where `gaps_are_consistent()` is false or `parent_end()` disagrees with the span's `lrg_end`, whereas #1830 would still return the affine-exact prefix. Measured here as **0 of 1294** — empty today, not empty in principle.

**#1830's `affine_horizon` does not land.** If this merges first, whoever rebases #1830 must drop it rather than keep both sides: keeping both yields a narrowed span carrying the full gap list, which trips both integrity checks and costs all 46 indel-bearing records their placement. The silent bad resolution is the neighbouring one — keep `affine_horizon`, drop this PR's two integrity assertions — so guard against that specifically.

Post-merge check on `main`: `LRG_292:g.160875` must answer `Some(43057109)` rather than decline. Full reasoning: https://github.com/fulcrumgenomics/ferro-hgvs/issues/1873#issuecomment-5295907571

PR #1848, opened 2026-08-13
refactor(normalize): name the partition rule the derivation surface cuts with

### LINKED ISSUE #1834: normalize and from_sequences run different block partitioners by construction, and nothing measures the gap

nh13:
Closes #1834.

Representation-Change: none. The derivation surface is passed `PartitionRule::Canonical` explicitly, which is the partitioner it already called unconditionally, so no output moves. Proof below: the suite is identical before and after except for the two tests this adds.

## What this is, and what it deliberately is not

#1834 proposes two halves and says the first does not depend on the second. **This is neither half — it is the seam underneath them.** The cross-surface agreement census and its ratchet are owned by another session and are not in this PR; there is no corpus, no row count, and nothing here measures how far apart the two surfaces are. Converging the partitioners is #1419/#1420/#1421's open product decision and this PR makes none of it.

What it does is make the coupling **stated and enforced** rather than incidental, so a future `FERRO_PARTITION` evaluation cannot move one surface and leave the other without anybody noticing.

## The problem, restated in one line

`derive_block_members` — the seam `from_sequences` enters `merge` at — called `partition_block_canonical_within` unconditionally. `canonicalize_from_sequence` cuts with whatever `partition_rule()` resolves `FERRO_PARTITION` to. So the two surfaces partition differently by construction, and the choice was legible only by reading the body of `derive_block_members` and noticing which function it called. Nothing anywhere named it, and nothing failed if it drifted.

## The shape chosen, and why not the alternative

- `DERIVED_BLOCK_PARTITION_RULE` names the derivation surface's rule in the **same arm vocabulary** `canonicalize_from_sequence` selects from. Flipping this surface is now editing one line.
- `const _: () = assert!(DERIVED_BLOCK_PARTITION_RULE.cuts_with_canonical());` keeps the pin in the canonical family at **compile** time (`cuts_with_canonical` becomes a `const fn`).
- `partition_block_for_derivation` is where the pin is consulted — the sibling of `partition_block_for_rule`, so both surfaces now select a partitioner through a named function that takes a `PartitionRule`.
- `the_two_surfaces_cut_with_a_pinned_pair_of_rules` pins the **pair** `(partition_rule(), DERIVED_BLOCK_PARTITION_RULE)` = `(Live, Canonical)`. Not an equality assertion: they legitimately differ today, and an equality assertion would fail for saying so.

**Why the rule is not threaded in as a parameter from `from_sequences`** (the first shape #1834's discussion suggested): `derive_block_members` cannot serve the arms it would then have to accept. `Live`/`Shadow` are not merely unused there — `partition_block_sequence_first` needs a `min_separation` this path structurally has none of, for the same reason the function takes no weight bound (a sequence pair carries no spelling for one to be relative to). The arms would be dead code or, worse, a `None` that reaches the caller as `BlockDecline::WouldNotRender` — reporting "the partition would not render" for "this surface does not serve that arm", which is the exact misattribution `BlockDecline`'s own doc records having cost a diagnosis once. The pin plus the compile-time family check buys the same one-line reviewable flip without inventing unreachable behaviour, and it would also have required making `PartitionRule` `pub(crate)` to put a choice about `merge`'s internal vocabulary in a module with no other business with it.

`Canonical` rather than `CanonicalCoalesced` because the two differ only through `payload_coalesce_applies` / `compensating_gap_coalesce_applies`, both scoped to the coding DNA axis by `delins-payload-coincidence-carve-out-is-coding-dna-scoped`, and `from_sequences` emits `g.`/`m.` only. On this surface the two arms are the same rule; naming the narrower one records that the agreement is a consequence of the axis scope rather than of the arms being interchangeable.

## Scope: this pins the RULE, and deliberately not the grid budget

Stated here because the PR would otherwise read as a stronger guarantee than it gives. `partition_block_canonical(r, a)` **is** `partition_block_canonical_within(r, a, MAX_SEQFIRST_GRID_CELLS)` — the budget is the only thing separating the two entry points. `from_sequences` threads `FromSequencesOptions::max_grid_cells` through and its docs invite raising it for a long-read haplotype, while `canonicalize_from_sequence` is fixed at the constant. So two callers can resolve to the **same** `PartitionRule` and still disagree: same rule, different budget, and the low-budget side declines (`GridTooLarge`) where the other answers.

That is left out on purpose. The budget is a cost bound whose stated policy is to refuse rather than answer with a weaker rule, not a choice of partitioner; the two defaults are already one constant; and only an explicit caller opt-in can split them. Asserting on it would make the pin read as "the two surfaces agree". The test's own doc comment says all of this, so the next reader does not over-trust it.

## No-op proof

Same selection either side (`cargo nextest run --features dev --lib --test it`), `EXIT=` read back out of each log rather than from the job's exit status:

| | result |
|---|---|
| before (`origin/main` @ `64fc76df`, clean tree) | `EXIT=0` — `10440 tests run: 10440 passed, 150 skipped` |
| after | `EXIT=0` — `10442 tests run: 10442 passed, 150 skipped` |

Diffing the two runs' per-test verdicts (`<status> <test name>`, sorted) gives exactly two lines, both additions, both this PR's own tests:

```
> PASS normalize::merge::tests::partition_rule_knob::the_derivation_seam_cuts_with_the_rule_it_names
> PASS normalize::merge::tests::partition_rule_knob::the_two_surfaces_cut_with_a_pinned_pair_of_rules
```

No pre-existing test changed verdict. Full suite (`cargo nextest run --features dev`) is `EXIT=0` — `10633 tests run: 10633 passed, 152 skipped`; `cargo fmt --all -- --check` and `cargo clippy --features dev --all-targets -- -D warnings` are both `EXIT=0`.

## The guards were checked against sabotage, not assumed

A no-op refactor's tests are worthless if they would pass either way, so each was forced open:

| sabotage | result |
|---|---|
| pin edited to `PartitionRule::Live` | **compile error** — `evaluation panicked: assertion failed: DERIVED_BLOCK_PARTITION_RULE.cuts_with_canonical()` |
| `partition_block_for_derivation` forwarded to `partition_block` | `the_derivation_seam_cuts_with_the_rule_it_names` **FAILED** (`EXIT=100`) |
| unset `FERRO_PARTITION` mapped to `CanonicalCoalesced` — i.e. #1835's flip | `the_two_surfaces_cut_with_a_pinned_pair_of_rules` **FAILED** (`EXIT=100`), `left: (CanonicalCoalesced, Canonical)` / `right: (Live, Canonical)` |

The third is the point of the PR: the tripwire fires on the real change it exists for, and its message names `from_sequences` as the thing that did not move.

## Interaction with #1835 — read this before merging either

**Textually: one conflict, and it is mechanical.** `git merge-tree --write-tree --name-only <#1835 head> HEAD` conflicts in `src/normalize/merge.rs`, in exactly one hunk: this PR inserts two tests immediately above `an_absent_value_selects_the_live_rule`, and #1835 rewrites that test's doc headline. Nothing semantic overlaps — resolution is "keep #1835's rewritten headline and keep these two tests". The `src/` regions do not conflict at all: #1835 adds `DEFAULT_PARTITION_RULE` after `PARTITION_RULE_NAMES` while this adds `DERIVED_BLOCK_PARTITION_RULE` after the `impl PartitionRule` block, and `derive_block_members` is untouched by #1835.

**Semantically: this PR's tripwire will fire on #1835, by design.** Once an unset `FERRO_PARTITION` resolves to `CanonicalCoalesced`, the pinned pair reads `(CanonicalCoalesced, Canonical)` and the test fails — measured directly, as sabotage 3 above. Whichever lands second re-pins the pair, and re-pinning it is the moment to answer the question this seam exists to force: *does `from_sequences` follow the flip, or stay canonical?* Note the answer is not automatic even though both sides would then cut canonically — on `g.`/`m.` the two arms are identical only because `delins.md:47`'s carve-out is `c.`-scoped, so the agreement would be a coincidence of two hard-codings rather than a shared setting. The test's doc walks both failure directions and says what to do in each.

`git merge-tree --write-tree --name-only origin/main HEAD` is clean, and so is the same check against #1837.

## Not touched

No ruling ledger record — this decides no normalization question, it names a choice that was already being made. No new `tests/it/` module, so no collision with #1837's `main.rs` registrations; the tests live in `merge.rs`'s existing `partition_rule_knob` module, which is where `PartitionRule` is visible and where the sibling knob tests already are.

One finding from self-review was applied: `PartitionRule`'s type doc said "Which block partitioner `canonicalize_from_sequence` cuts with" and "Three rules exist" (there are four). Both are now correct, and the doc names both consuming surfaces.

### DIFF
--- a/src/normalize/from_sequences.rs
+++ b/src/normalize/from_sequences.rs
@@ -241,6 +241,12 @@ pub fn from_sequences_detailed(
         msg: format!("position {position} does not fit a signed coordinate"),
     })?;
 
+    // This surface's block partitioner is a **pin**, not `FERRO_PARTITION`'s
+    // reading: `merge::DERIVED_BLOCK_PARTITION_RULE` names it, and
+    // `merge::partition_block_for_derivation` is where it is consulted. So a
+    // change to the rule `Normalizer::normalize` cuts with does not reach here,
+    // which is #1834 — the tripwire that fails when either moves alone is
+    // `merge`'s `the_two_surfaces_cut_with_a_pinned_pair_of_rules`.
     let block = derive_block_members(
         reference,
         alternate,
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2236,11 +2236,21 @@ fn seqfirst_shadow_enabled() -> bool {
     *ENABLED.get_or_init(|| std::env::var("FERRO_SEQFIRST_SHADOW").as_deref() == Ok("1"))
 }
 
-/// Which block partitioner [`canonicalize_from_sequence`] cuts with.
+/// Which block partitioner a surface cuts with.
 ///
-/// Three rules exist and they are not variations on one — see
+/// **Two surfaces select from this vocabulary, and they select differently.**
+/// [`canonicalize_from_sequence`] resolves its rule at runtime, from
+/// [`partition_rule`] and so from `FERRO_PARTITION`;
+/// [`derive_block_members`] — the seam
+/// [`crate::normalize::from_sequences`] enters this module at — is **pinned**
+/// to [`DERIVED_BLOCK_PARTITION_RULE`] and reads no environment at all. That
+/// asymmetry is #1834; this type is where both halves of it are written in one
+/// vocabulary. Do not read the enum as describing only the `normalize` path.
+///
+/// Four rules exist — the four [`PARTITION_RULE_NAMES`] advertises — and they
+/// are not variations on one; see
 /// [`crate::normalize::seqfirst::partition::CanonicalAlignment`] for what
-/// separates the latter two.
+/// separates the last two.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum PartitionRule {
     /// [`partition_block`]: a single-gap alignment search plus its two narrow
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai pause

PR #1850, opened 2026-08-13
docs: publish the adjudication ledger as a generated normative document

### LINKED ISSUE #1552: docs: publish ferro's canonical-form rules as a normative document (a shadow spec)
## Summary

Ferro's rules for choosing one description when several are legal are currently recorded as **ruling records** (`tests/fixtures/grammar/hgvs_spec_normalization_overrides.json`) plus the tests that enforce them. That is the right home for now — the tests are the documentation, and a record that no test reads is a record that can rot.

But the audience for those rules is not only ferro's own test suite. A contributor deciding a new case, and a downstream consumer trying to predict what a string will normalize to, both need to read the ladder end to end. A JSON `rationale` blob is a poor medium for that.

## What to write, eventually

A normative document describing **what ferro does where the HGVS recommendations are silent, ambiguous, or self-contradictory** — effectively a *shadow spec* that sits alongside the recommendations rather than restating them. It should cover at minimum:

- the precedence order from `adjudication-precedence-order`
- the canonical-form rule from `canonical-form-choice-when-both-legal`
- the full tie-break ladder used to choose among equally-legal partitions
- the places where the recommendations conflict with themselves and which side ferro takes, with the clause citations

Each section should cite the ruling record it derives from, so the document is a rendering of the records rather than a second source of truth that can drift from them.

## Why not now

The records and their tests are the enforceable artifact; prose is not. Writing the document before the ladder is settled would mean maintaining two descriptions of a moving target. This issue is a placeholder for after the canonical-form precedence order is decided and implemented.

## Constraint

The document must be **generated from or checked against** the ruling records, not hand-maintained in parallel. Two hand-maintained copies of a normative rule is the failure mode this project has already hit with the census constants and the `failure-patterns.md` tallies.

nh13:
Closes #1552

Representation-Change: none. Docs, one new integration-test module, and two supporting fields on the shared ledger reader. No `src/` file is touched, so no output moves.

## What this adds

`docs/NORMALIZATION_CONTRACT.md` — every adjudication record in `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json`, published in full: its question, its verdict, the spec clauses it names, the text each clause was quoted against, and the reasoning. Grouped by status, ordered by id, with a contents list so a reader arriving with a record id can find it.

The records are the enforceable artifact and stay so. What they were not was *readable*: a contributor deciding a new case, or a consumer trying to predict what a string will normalize to, had to read a JSON `rationale` blob. This is that reading.

## Why it is generated

#1552 makes it a standing constraint that the document be generated from or checked against the records rather than hand-maintained beside them. Generation is the stronger of the two options offered, so that is the one taken: every question, verdict, clause and quote in the body is the record's own text, copied mechanically. There is no editorial pass in which a paraphrase could enter, and nothing to keep in sync by hand.

`tests/it/normalization_contract_doc.rs` renders it and gates it. Regenerate with:

```
BLESS_CONTRACT_DOC=1 cargo nextest run --features dev --test it -E 'test(normalization_contract_doc)'
```

**Why a test rather than an `[[example]]` with `--check`.** The ledger reader is `common::rulings`, which lives in the integration-test crate; an example target cannot use it, so an example generator would need a second parser for the same file — the drift this document exists to avoid, one level down. Living here also means the gate needs no new CI wiring, since the required `Test` context is a rollup over the job that runs this binary.

## What it deliberately does not do

**It does not restate README's normalization ruleset.** It links to it, and `the_preamble_does_not_restate_the_readme_ruleset` enforces that against `README.md` itself rather than against a copy of the rule text. Single-sourcing that ruleset is part of one of the rulings this document renders, so a document that restated it would contradict its own contents. README gains a short pointer in the other direction.

The module names no record id anywhere. Every id in the output arrives from the ledger at render time, so there is no second place a status claim can go stale.

## The default-arm flip

The rulings are decisions about what the output *should* be and do not depend on which partitioner is selected — but several records are explicit about whether their ruling is already live in shipped output or is implemented only under a candidate arm, and the default is in motion (#1835).

So the document says which default it was generated against, and **that sentence is derived, not written**: the generator reads the arm out of `partition_rule_from_env` in `src/normalize/merge.rs`. Flipping the default therefore fails this document's own test until it is regenerated, and cannot leave a stale claim behind. It currently publishes `live`.

## Supporting changes

- `common::rulings::Record` gains `question` (required, with a rejection test) and `equivalence_classes`; the two optional string arrays now share one parser. `question` is required rather than optional because the id is not a fallback for it — one record in this ledger is decided the opposite way to what its id reads as, so a renderer that substituted the id would publish that record's question backwards.
- `ledger_clause_jurisdiction.rs`'s synthetic record builder fills both new fields.

## Checks

Seven tests, beyond "the file matches": every record reaches the document under a heading stating its status and in the half its status belongs to; every record's question, ruling and clause quotes are published; the preamble links to the README ruleset and restates none of it; the published default arm is the one the code selects; and the render is already in the shape `trailing-whitespace` and `end-of-file-fixer` want, so committing the document cannot rewrite it into a failure that reads as ledger drift.

Full local suite: 10448 passed, 0 failed. `cargo fmt --check` and `cargo clippy --all-features --all-targets -- -D warnings` clean, and the three generated-doc `--check` gates still pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,6 +310,17 @@ Rule 7's disclosure mechanism — the `Representation-Change:` trailer and how i
 changelog — is documented in
 [CONTRIBUTING.md](CONTRIBUTING.md#declaring-a-representation-change).
 
+### Where the individual decisions live
+
+Rules 5 and 6 above say *how* a question is decided where the recommendations are silent,
+ambiguous or self-contradictory. **What was decided, case by case**, is recorded separately, as
+adjudication records — each naming the clauses in tension, which one governs, which is deviated
+from, and why. Those records are published in full, with their clause quotes, in
+[docs/NORMALIZATION_CONTRACT.md](docs/NORMALIZATION_CONTRACT.md).
+
+That document is generated from the records and gated against them, so it cannot drift; it
+deliberately does not restate the seven rules above, which are stated only here.
+
 ## Deriving a description from sequences
 
 If what you have is **bases** rather than a description — a window out of a BAM, a VCF row, an
--- a/docs/NORMALIZATION_CONTRACT.md
+++ b/docs/NORMALIZATION_CONTRACT.md
@@ -1,0 +1,1390 @@
+<!--
+GENERATED FILE — do not edit by hand.
+
+Rendered from the `rulings` section of
+tests/fixtures/grammar/hgvs_spec_normalization_overrides.json
+by tests/it/normalization_contract_doc.rs. Edit the ledger, then regenerate:
+
+    BLESS_CONTRACT_DOC=1 cargo nextest run --features dev --test it \
+      -E 'test(normalization_contract_doc)'
+
+An edit made here instead is reverted by the next regeneration, and fails CI
+before that.
+-->
+
+# Ferro's normalization contract
+
+**28 adjudication records — 25 decided, 3 open.**
+
+## What this document is
+
+The HGVS recommendations are, in places, silent, ambiguous, or self-contradictory. A
+normalizer still has to emit one string. Where ferro has had to decide such a question, the
+decision is recorded as a **ruling record**, and this document is a rendering of every one of
+those records: its question, its verdict, the spec clauses it names, the text each clause was
+quoted against, and the reasoning in full.
+
+It is generated, not written. Every question, verdict, clause and quote below is the record's
+own text, reproduced mechanically from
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai pause

PR #1851, opened 2026-08-13
ci: run the 45 orphaned reference-aware guards, and fail on a missing manifest

### LINKED ISSUE #1849: ci: 45 reference-aware guards run in no automated job — they skip green everywhere
## Summary

45 checked-in reference-aware test guards are executed by **no automated job**. They are gated on a ferro-prepared reference manifest, and every one of them `return`s early when it does not find one — so they skip green in PR CI (which provisions no manifest) and skip green in `nightly-mutalyzer.yml` (whose `REFERENCE_AWARE_SELECTION` names three modules, none of them these).

A skip is indistinguishable from a pass in every signal CI reports. **The runtime is the only tell**: measured on the blessed scratch reference, the same binary and the same commit gives

```
PASS [   0.02s] ferro-hgvs::it multi_member_cis_axis::axis_normalized     # FERRO_MANIFEST unset -> skipped
PASS [  11.48s] ferro-hgvs::it multi_member_cis_axis::axis_normalized     # FERRO_MANIFEST set   -> executed
```

Across the whole set the separation is clean — 0.01–0.08 s skipped against 9–58 s armed, with nothing in between.

## Why this is not hypothetical

`multi_member_cis_axis::axis_normalized` carries a ratchet-style census pin. It was **provably red on `main`** at both `30ca8f7d` (2026-08-08) and `9cf8efc9` (2026-08-10), measuring `respelling_converged: 395` against a pinned `376` — a window containing 13 merged commits, every one of which reported green. The drift surfaced four days later as #1581, a commit that moved the pin `376 → 395` while touching only the test file, and was then investigated as an "inconsistency between worktrees" with an assumed environmental cause. It was neither environmental nor a worktree problem: it was a guard that had stopped guarding.

The imminent partition flip (#1835, `Live` → `CanonicalCoalesced`) makes the point concrete for the class rather than the instance. It aggravates #1712: `mutalyzer_normalize_tests::axis_noncoding_idempotent` goes from 21 non-idempotent rows of 681 to 62, `EXIT=0` → `EXIT=100`. That module **is** in the nightly's selection, so the regression is at least visible. The 45 guards below are exactly where the same class of regression would be visible nowhere.

## The census

Measured empirically as a duration diff — the same nextest selection run twice over 32 candidate modules, once with `FERRO_MANIFEST` unset and once set, comparing the per-test time nextest prints on every line. Static analysis was tried first and got it wrong three separate ways (a `macro_rules!` gate, an `async` + `#[tokio::test]` gate, and 24 over-flagged cases that turn out to use a mock).

```
A (unset): Summary [   5.176s] 779 tests run: 779 passed     EXIT=0
B (set):   Summary [ 307.469s] 779 tests run: 779 passed     EXIT=0
```

**38 tests across 14 modules** are `FERRO_MANIFEST`-gated and in no selection:

| module | armed tests | armed cost |
|---|---:|---:|
| `issue_1086_c_axis_genomic_context` | 14 | 300.6 s |
| `issue_1086_cross_axis_projection` | 5 | 90.8 s |
| `issue_972_cds_start_nf_decline` | 3 | 65.7 s |
| `issue_498_whole_exon_deletion` | 3 | 42.5 s |
| `issue_1217_mito_terminus_insertion_clamp` | 2 | 31.8 s |
| `issue_944_cds_to_tx_transcript_native` | 2 | 31.4 s |
| `issue_1713_bare_transcript_genomic_decline` | 2 | 30.3 s |
| `issue_843_allele_rna_build_scoping` | 1 | 24.7 s |
| `inversion_sweep` | 1 | 17.8 s |
| `issue_1539_split_member_separation` | 1 | 15.9 s |
| `protein_cis_delins_decomposition` | 1 | 15.7 s |
| `issue_806_effect_real_residues` | 1 | 15.0 s |
| `multi_member_cis_axis` | 1 | 11.5 s |
| `delins_equal_vs_unequal_length_discriminator` | 1 | 10.3 s |
| **total** | **38** | **704.0 s** |

**5 more** are orphaned in a stronger sense. `real_data_normalization_tests` gates on `Path::new("benchmark-output/manifest.json").exists()` and ignores `FERRO_MANIFEST` entirely, so an operator following the documented local recipe (`export FERRO_MANIFEST=…`) does not arm them either — they are dead even for the developer trying to reproduce a nightly result.

**2 more** are `#[cfg(test)]` units inside `src/` rather than in the `it` binary, so a census scoped to `tests/it/` cannot see them at all:

- `conformance::accession_inventory::tests::corpus_missing_ng_against_real_reference_when_available`
- `reference::ng_placement_builder::tests::derive_for_real_reference_when_available`

For scale, the three modules the nightly **does** select contribute 19 armed tests. **The orphaned reference-aware surface is more than twice the size of the covered one.**

## The mechanism gap, which is the part worth fixing

The bulk-fixture family had exactly this problem and solved it. Four modules read the four release-asset corpora through one helper (`tests/it/common/bulk_fixtures.rs`), every one of them skips green when its fixture is missing, and `FERRO_REQUIRE_BULK_FIXTURES=1` — set in every job that fetches them — turns that skip into a panic naming the missing path and the fix. That module's own doc comment states the principle:

> a fetch that failed quietly would delete the ClinVar, CMRG and Paraphase coverage while making the job look *faster*, which is the worst outcome available here.

**The manifest family has no equivalent.** That is the single smallest structural difference between the solved case and this one, and it is why the gap will silently reopen the next time a manifest-gated guard is added to a new module. Fixing today's list without it fixes a list, not a mechanism.

## Proposed fix

1. Enumerate the orphaned modules in `REFERENCE_AWARE_SELECTION`. **Not** a pattern: widening by glob imports the seven modules `ci.yml`'s `ORACLE_EXCLUDE` withholds from an armed job, which fail on `main` for reasons no nightly caused, and it makes the xfail artifact — the only channel a non-gating job has — unreadable.
2. Add a `FERRO_REQUIRE_MANIFEST=1` promotion built on the `FERRO_REQUIRE_BULK_FIXTURES` pattern, and set it in the nightly, so a `ferro prepare` that produced nothing fails the run instead of reporting a faster green one.
3. Give `real_data_normalization_tests` the same `FERRO_MANIFEST`-then-`benchmark-output` gate every other module uses, so it is reproducible locally.

## Residual this does not close

`nightly-mutalyzer.yml`'s test step is `continue-on-error: true` by deliberate design — its stated purpose is to surface drift in the xfail report rather than to gate. So wiring these guards buys **detection without enforcement**: a broken pin lands in the `ferro-xfail-<run_id>` artifact and the workflow still reports `success`. That is strictly better than a guard that never runs at all, and it is not the same thing as coverage. Making the nightly gating is a separate change, and it is blocked behind the job's pre-existing failures — in particular `comparator_provenance_reference_identity_matches_live`, whose failure says the nightly's own ephemeral reference is not the blessed identity, which must be answered before "is ferro correct?" can be.

nh13:
Closes #1849.

45 checked-in reference-aware guards were executed by **no automated job**. They gate on a ferro-prepared manifest and `return` early without one, so they skipped green in PR CI (which provisions no manifest) and skipped green in `nightly-mutalyzer.yml`, whose `REFERENCE_AWARE_SELECTION` named three modules — none of them these.

A skip is indistinguishable from a pass in every signal CI reports. The runtime is the only tell: 0.01–0.08 s skipped against 9–58 s armed, with nothing in between.

## Why now

The imminent partition flip (#1835, `Live` → `CanonicalCoalesced`) aggravates #1712: `mutalyzer_normalize_tests::axis_noncoding_idempotent` goes from 21 non-idempotent rows of 681 to **62**, `EXIT=0` → `EXIT=100`. That module *is* in the nightly's selection, so the regression is at least armed and visible. **The 45 guards this PR wires are exactly where the same class of regression would be visible nowhere** — that is the concrete argument for the change, and it is not retrospective.

It is not retrospective in the other direction either. `multi_member_cis_axis::axis_normalized` carries a ratchet-style census pin and was provably red on `main` at both `30ca8f7d` (2026-08-08) and `9cf8efc9` (2026-08-10), measuring `respelling_converged: 395` against a pinned `376`, across 13 merged commits that every one reported green. It surfaced four days later as #1581 — a commit moving the pin `376 → 395` while touching only the test file — and was then investigated as an "inconsistency between worktrees" with an assumed environmental cause. It was a guard that had stopped guarding.

## What changed

**1. `REFERENCE_AWARE_SELECTION` now enumerates every reference-aware module.** Deliberately enumerated, not globbed: widening by pattern imports the seven modules `ci.yml`'s `ORACLE_EXCLUDE` withholds from an armed job, which fail on `main` for reasons no nightly caused, and it makes the `ferro-xfail-*` artifact — the only channel a non-gating job has — unreadable. The workflow now says so at the point someone would be tempted, and so does the reproduce-recipe the summary step prints.

**2. `FERRO_REQUIRE_MANIFEST=1`, set on the nightly test step.** Adding modules to the selection without this is half a fix: they would still skip green if a `ferro prepare` produced nothing, or wrote a manifest where the test step cannot see it — reporting a *faster* green run, which is the worst outcome available. The promotion is built on the existing `FERRO_REQUIRE_BULK_FIXTURES` pattern and reuses its shape, predicate semantics (`0` and empty are "set but off") and message discipline rather than inventing a second convention.

The policy lives in `src/conformance/manifest_gate.rs` rather than beside `bulk_fixtures.rs` in the test tree for one reason: two of the guards are `#[cfg(test)]` units inside `src/` itself and cannot see `tests/it/common/`. A second copy for those two is exactly the drift the mechanism exists to prevent. `tests/it/common/manifest.rs` is a thin re-export so integration modules keep reading as one convention.

**3. `real_data_normalization_tests` gains the standard gate.** It consulted only `Path::new("benchmark-output/manifest.json")` and ignored `FERRO_MANIFEST`, so it was orphaned in a stronger sense than its siblings — dead even for a developer following the documented local recipe. It now uses the `FERRO_MANIFEST`-then-`benchmark-output` convention every other module uses.

## The count: 45, re-derived — the costing note said 41

Measured empirically as a duration diff (32 candidate modules, same selection run twice). Static analysis was tried first and gets this wrong.

```
A (FERRO_MANIFEST unset): Summary [   5.176s] 779 tests run: 779 passed   EXIT=0
B (FERRO_MANIFEST set):   Summary [ 307.469s] 779 tests run: 779 passed   EXIT=0
```

**38 `FERRO_MANIFEST`-gated tests across 14 modules**, plus **5** in `real_data_normalization_tests` (invisible to a `FERRO_MANIFEST` census — it gates on the relative path), plus **2** `#[cfg(test)]` units in `src/`. Reconciliation against the 41:

| | |
|---|---:|
| prior figure | 41 |
| `issue_1713_bare_transcript_genomic_decline` — module landed on `main` after the prior measurement commit | +2 |
| `conformance::accession_inventory::tests::corpus_missing_ng_against_real_reference_when_available` and `reference::ng_placement_builder::tests::derive_for_real_reference_when_available` — lib units, outside a `tests/it/`-scoped census | +2 |
| **re-derived** | **45** |

For scale, the three modules the nightly already selected contribute 19 armed tests. **The orphaned surface was more than twice the size of the covered one.**

## Proof the promotion actually fails

Same binary, same commit, same selection; only the manifest differs.

```
manifest ABSENT + FERRO_REQUIRE_MANIFEST=1
  Summary [   0.671s] 291 tests run: 223 passed, 68 failed, 10495 skipped
  EXIT=100
```

68 unique failures, which reconcile exactly: 38 orphans + 5 `real_data_normalization_tests` + 2 lib units + 19 already-covered armed guards + 3 `manifest_or_skip` canaries + `comparator_provenance_reference_identity_matches_live`. The failure text names the guard and the remedy:

```
FERRO_REQUIRE_MANIFEST is set, but `multi_member_cis_axis::axis_normalized` found no reference manifest.
...
Without one the guard would SKIP GREEN, silently dropping its coverage — and a skip is
indistinguishable from a pass in every signal CI reports — so under FERRO_REQUIRE_MANIFEST
an absent manifest is a failure instead.
```

```
manifest PRESENT + FERRO_REQUIRE_MANIFEST=1 + the nightly's four oracle flags
  Summary [  66.964s] 291 tests run: 291 passed, 10495 skipped
  EXIT=0
```

Zero `skipping — no reference manifest` lines in that log, and the timings confirm arming rather than skipping (slowest 22.9 s). Measured against the blessed scratch reference, identity `aa8b3246d83055cc`.

## Which of the newly-wired guards are RED

**None.** All 45 pass under the nightly's exact four oracle flags on the blessed reference. The value of this change is prospective, and it should be read that way rather than as a bug hunt.

The prior costing note recorded 6 pre-existing failures in `mutalyzer_normalize_tests` at its measurement commit, attributed (relayed, unverified there) to an SPDI reference-frame defect. That attribution now checks out by measurement rather than by relay: #1821 merged 2026-08-13, and the old three-module selection under the same flags on current `main` reads `170 tests run: 170 passed, EXIT=0`. The six are gone, and nothing here suppressed them.

No guard was excluded to make the job green.

## Cost

| selection | wall | result |
|---|---:|---|
| old (3 modules), armed, four oracles | 43.5 s | 170 pass |
| new (all reference-aware), armed, four oracles | **67.0 s** | 291 pass |

**+23.5 s of nextest wall, a 1.54× multiplier on the test step**, for +121 tests of which +45 are manifest-armed. That lands on a job whose wall clock is ~77 % `ferro prepare` (24 m 51 s of a 32 m job), so the total moves by single-digit percent. This is a local wall-clock measurement on a shared box, not a CI benchmark — treat the ratio, not the absolute.

## Residual — stated rather than fixed

**`nightly-mutalyzer.yml`'s test step is `continue-on-error: true`, so this buys detection without enforcement.** A broken pin lands in the `ferro-xfail-<run_id>` artifact and the workflow still reports `success`. That is a deliberate design choice with its reasoning on the flag itself — the nightly exists to surface drift in the xfail report, not to gate PRs — and this PR does not change it.

So the honest statement is: **a guard that runs in a non-gating job is better than one that never runs, and it is not the same as covered.** Making the nightly gating is a separate change and is blocked behind that job's pre-existing failures, in particular `comparator_provenance_reference_identity_matches_live` — whose failure says the nightly's own ephemeral `ferro prepare` output is not the blessed identity, which has to be answered before "is ferro correct?" can be. Deliberately not expanded into here.

Three smaller residuals, none of them closed by this PR:

- **`FERRO_REQUIRE_MANIFEST` cannot catch an omission from the selection.** It fires when a *selected* guard finds no manifest, not when an unselected one is never asked. A newly-added manifest-gated module in a new file is still invisible until someone adds it to the list; the workflow now says so where the list is defined.
- **`real_data_normalization_tests` still skips green on `Skipping test: transcript not found`.** A manifest that is present but lacks the transcript is the same defect class one level down. Left as-is rather than widened in passing.
- **`idempotency_tests::test_clinvar_sample_parsing_idempotency` reads `clinvar_1000.txt` from the repo root**, a file that exists nowhere in the tree and is fetched by no script. It is a permanently-green no-op. Out of scope here; worth its own issue.

## Verification

```
cargo fmt --all --check                                              EXIT=0
cargo clippy --all-features --all-targets -- -D warnings             EXIT=0
actionlint / zizmor on both changed workflows                        EXIT=0, no findings
cargo nextest run --features dev  (full suite, manifest-free)        EXIT=0, 10634 passed
```

The manifest-free run is the PR-CI condition, and it is the one that matters for this change: `FERRO_REQUIRE_MANIFEST` is unset there, so every promoted guard must still skip exactly as before. It does.

Representation-Change: none

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1154,6 +1154,14 @@ jobs:
         # suite both silently no-op when `FERRO_MANIFEST` is unset (PR CI does
         # not provision the multi-GB reference manifest). Surface a GitHub
         # annotation so reviewers see the skip explicitly. #397 item 6.
+        #
+        # `FERRO_REQUIRE_MANIFEST` is deliberately NOT set anywhere in this
+        # workflow. It is the sibling of `FERRO_REQUIRE_BULK_FIXTURES` below —
+        # it promotes a manifest-absence skip to a failure — and the two differ
+        # in exactly one respect: PR CI *does* provision the bulk fixtures and
+        # *does not* provision a manifest, so requiring one here would fail
+        # every PR. It is set in `nightly-mutalyzer.yml`, the only job that runs
+        # `ferro prepare`. See `src/conformance/manifest_gate.rs`.
         if: matrix.shard == 1
         run: |
           echo "::notice title=Reference-aware tests skipped on PR CI::PR CI does not provision FERRO_MANIFEST; the mutalyzer / biocommons reference-aware test suites silently skip. The nightly nightly-mutalyzer.yml workflow runs the full suite against a real manifest."
--- a/.github/workflows/nightly-mutalyzer.yml
+++ b/.github/workflows/nightly-mutalyzer.yml
@@ -1,11 +1,16 @@
 name: Nightly reference-aware tests
 
-# Runs the mutalyzer / biocommons / hgvs-rs-projection reference-aware test
-# suites against a real ferro-prepared manifest. PR CI silently skips these
-# tests (see
-# the `::notice::` annotation in `ci.yml`) because provisioning a
-# multi-GB reference manifest would multiply PR CI time. This workflow
-# runs nightly so manifest-dependent regressions surface within 24h.
+# Runs every reference-aware test suite in the repository against a real
+# ferro-prepared manifest — see `REFERENCE_AWARE_SELECTION` below for the list
+# and for why it is enumerated rather than matched by pattern. PR CI silently
+# skips these tests (see the `::notice::` annotation in `ci.yml`) because
+# provisioning a multi-GB reference manifest would multiply PR CI time. This
+# workflow runs nightly so manifest-dependent regressions surface within 24h.
+#
+# `FERRO_REQUIRE_MANIFEST` is set on the test step, so a guard that finds no
+# prepared reference FAILS here instead of skipping green. Without it, the
+# widened selection would be worth very little: a `ferro prepare` that produced
+# nothing would still report a green — and faster — run.
 #
 # Caching: the prepared reference directory is keyed on the Cargo.lock
 # hash (changes when the `prepare` crate's deps shift) + a fixed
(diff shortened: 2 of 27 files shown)


nh13:
@coderabbitai pause

PR #1852, opened 2026-08-13
fix(ci): refuse a Representation-Change trailer no consumer can read

nh13:
A `Representation-Change:` trailer is the published disclosure to a consumer who keys read counts on the normalized HGVS string: `release-plz.toml` writes each changelog bullet from the commit subject alone, and `scripts/inject_representation_disclosure.py` attaches the trailer's own text underneath it. Silence there is a silent re-bucketing of stored designs.

There is a way to be silent while believing you declared, and it costs one character.

## The shape

PR #1838 declares on line 3 of its description as `` `Representation-Change: none` `` — inside a code span. Measured against the shipping code, not inferred:

- `scripts/check_representation_change.py` finds **zero** trailers; `TRAILER_RE` is column-0-anchored by #1573, deliberately.
- git-cliff's footer rule in `release-plz.toml` will not match it, so it never reaches the changelog.
- `inject_representation_disclosure.py` will not attach it.

The check is nonetheless **green**, and correctly so under the old logic: #1838 touches only a fixture JSON, no watched directory, so it hits the `not watched` early return *before* the unreadable trailer would have mattered.

Nothing was lost. `none` is excluded from the changelog anyway, so an invisible decline and a visible one publish the same thing. **The shape is the hazard**: a PR can hold a trailer its author believes is declared while both the checker and the changelog renderer see silence. Had #1838 touched a gated directory, or had the value been a real disclosure, that is precisely the silent representation change this mechanism exists to prevent.

## What this adds

`find_near_misses` reports a line that looks like a declaration but that none of the three consumers can read, and `check` refuses it — quoting the line and naming the character responsible, rather than reporting "no declaration" and sending the author to write a trailer they have already written.

It covers the decorations that render a line as prose while leaving it looking like a trailer in an editor: a code span, `>`, `#`, a list marker, emphasis, plain indentation, and the un-hyphenated spellings (the hyphen is what git-cliff keys on, so a space or an underscore is exactly as invisible as a backtick).

Output on #1838's real body:

```
  1 line(s) look like a `Representation-Change:` declaration but
  no consumer can read them:

    line 3: `Representation-Change: none`
      -> wrapped in a code span (a backtick)
```

## Two design choices, both argued rather than assumed

**It fires even when no watched directory is touched.** This is the point rather than an overreach: behind the early return the check would pass #1838, the one PR that motivated it, and rebuild the same hole. It is also not a new principle here — the two-trailer refusal already sits ahead of that return, with the recorded reason that *the harm is in changelog grouping, which applies to every commit*. A near miss has the same shape: what fails is what the changelog renders.

It does make the check stricter on changes it currently ignores entirely, so the cost is measured below rather than waved past.

**A single valid column-0 trailer silences the scan.** An author who declared properly has demonstrated they know the form, and the same author routinely discusses that declaration in prose. #1742 and #1837 each quote their own trailer in a code span on a line beside a real one; firing on those would fail a PR for explaining itself, which is the #1555 mistake in a new place. `CONTRIBUTING.md` also documents indenting a quoted example as the escape hatch, and the two-trailer refusal's own message recommends it — so reporting there would contradict guidance this checker already gives.

## Cost, measured over all 39 open PRs

| | |
|---|---|
| PRs where the pattern matches a line | **3** (#1838, #1837, #1742) |
| …silenced by carrying a valid trailer | **2** (#1837, #1742) |
| **Reported — i.e. would go red** | **1** (#1838) |

No required context is added, so nothing is stranded on "Expected — waiting for status to be reported". Exactly one open PR changes verdict, and it is the one the fix is for. The fix there is one character: delete the backticks.

## Testing

TDD, and the failing run was watched before any implementation existed: with a stub `find_near_misses` returning `[]`, **30 failed / 101 passed**, every failure an assertion about the checker passing a near miss rather than a collection error. Two defects were caught by the tests during implementation rather than by review — a code span being mislabelled "indented", and the closing backtick of `` `Representation-Change:` `` satisfying `\S` and so reporting a bare field mention as a failed declaration.

Every guard case is drawn from a real open PR body rather than invented, which is what makes the false-positive figure above meaningful.

Local gate, exit codes read back out of the log rather than from the runner's summary:

```
  PYTEST_EXIT=0     174 passed  (trailer + changelog-grouping + injector suites)
  RUFFCHECK_EXIT=0
  RUFFFMT_EXIT=0
  clippy EXIT=0     (no Rust changed; run to confirm rather than argue)
  cargo fmt --all --check EXIT=0
```

## Two findings from the same measurement, filed separately

- **`WATCHED_PREFIXES` undercovers what moves output** — #1853. Four open PRs disclose real consumer-visible moves while touching no gated directory, all under `src/reference/`, and two say so in their own trailers. Not widened here: that is a coverage decision with its own cost, and it belongs to the operator. #1853 also records the #1024 finding: the one genuinely-missing declaration among the 39 open PRs, inside the current gate and invisible to it.
- **`TRAILER_RE`'s first-line truncation** — #1854. Verified **not** live in the injector's path (`disclosure_value` already extends past the first line), but live in the contradicted-decline tripwire, which reads only the first line. Zero open PRs are currently in that state.

Representation-Change: none


nh13:
@coderabbitai pause

PR #1855, opened 2026-08-13
docs: audit each normalization stage as governed, ruled, or undecided

### LINKED ISSUE #1553: The normalization contract is only partly adjudicated — audit each stage as governed, ruled, or undecided
## Summary

Ferro's normalization pipeline makes a decision at every stage, but only some of those decisions are backed by a governing clause or a ruling record. The rest are backed by whatever the implementation happens to do. There is currently no inventory saying which is which, so a contributor cannot tell a deliberate choice from an accident, and neither can a reviewer.

The canonical-form work now in flight covers exactly one stage — choosing among equally-legal **partitions with their typing**. That is the stage blocking the partitioner default, so it is being done first. This issue tracks the rest.

## What to produce

For each stage below, one row recording: the decision the stage makes, whether it is governed by a spec clause (with `file:line`), by a ruling record (with its id), or by nothing; and if nothing, whether that is a genuine gap or a case the spec settles so plainly no record is warranted.

Stages, as a starting list:

| stage | decision it makes |
|---|---|
| parse / validate | which authored forms are rejected vs repaired |
| axis projection | how a coordinate maps across axes, and what happens at boundaries |
| sequence derivation | how the resulting sequence is computed from the authored edits |
| **partition + typing** | which of several legal descriptions is emitted — **in flight** |
| shifting | 3' rule application, and where it is clamped |
| merge / split | when consecutive or separated changes combine |
| re-spelling | terminal coalescing and type re-selection |

## Why this is worth doing separately

Three reasons it is not folded into the canonical-form work:

1. **Blast radius.** Re-opening shifting or projection alongside the partitioner would make a single release carry several independent representation changes, against the batching rule.
2. **Most rungs are probably already settled.** The expected outcome is that the majority of stages resolve to an existing clause, and the audit is cheap. Only the residue needs adjudicating.
3. **It is the same distinction the conformance census needs** — governed vs ran-but-unevaluable vs never-adjudicated. Doing it once serves both.

- The normative rendering of whatever this produces is #1552.
- The partition + typing stage is being decided now and should not be re-litigated here.

nh13:
Closes #1553

Representation-Change: none. A fixture, a doc, one integration-test module and a README pointer. No `src/` file is touched.

**Stacked on #1850.** Base is `docs/issue-1552-shadow-spec`, so the diff shown here is this PR's own work; GitHub retargets to `main` when #1850 merges. `git log --oneline origin/main..HEAD` is two commits, the first of which is #1850's.

## What this adds

An inventory: for each stage of the normalizer, what governs the decision that stage makes. The data is `tests/fixtures/grammar/normalization_stage_audit.json`; `docs/NORMALIZATION_STAGE_AUDIT.md` is generated from it and from the ledger.

The expected outcome, and the actual one, is that most stages resolve to an existing record. The deliverable is the residue.

## The residue — two decisions governed by nothing

Both classified as **genuine gaps** rather than as cases the spec settles so plainly no record is warranted:

**Ordering `c.` positions across numbering zones.** A `c.` position sits in one of three zones (`-n`, plain `n`, `*n`) and a cis allele's members can sit in different ones, so the normalizer needs a total order on endpoints — to sort members, to detect overlap, and to compute the separation the merge/split stage keys on. `background/numbering.md` defines each zone and states no rule for comparing across them; it never speaks about alleles at all. So the ordering is a house rule with no clause behind it, and unlike every other house rule at this grain it carries no ruling record either. It is written down as repository prose and nowhere else. This is not theoretical: the sequence-changing and denotes-no-sequence classes the spec corpus found localise to the `c.72`/`c.*1` transition and to nothing else.

**What the 5' direction does.** The recommendations state a 3' rule and only a 3' rule — `5' rule` and `5'-rule` occur nowhere in the checkout. Ferro offers a 5' direction, and the direction is not orthogonal: it selects the frame every other rule is evaluated in, which is why the README claims its best-effort rules per direction rather than across the two. So every clause and every record at the shifting stage is authority for the 3' arm only; the 5' arm is mirrored by construction rather than by a decision anyone made. That may well be the only possible answer, but it is nowhere adjudicated, and the mirroring is least obviously right at a junction, where the 3' answer is itself an exception.

Neither is proposed for a ruling record here — a record is for clauses in tension, and both of these are spec *silence*, which README rule 5 routes differently. Filing them is what this PR does; deciding them is not.

## The axis dimension

The second half of #1553, from its comment. The spec settles outright that cross-axis divergence is legitimate — asked whether the `g.` and `c.` descriptions of one variant may name different nucleotides, `general.md:162-167` answers **yes**, for a minus-strand gene inside a repeated sequence. Every axis disagreement this repository has filed was nevertheless a real bug: a dropped offset, a skipped pass, a missing gate. Without a table recording which is which, the next deliberate divergence is indistinguishable from the next defect.

The table's third column is what keeps it closeable. Of ten axis-scoped clauses, three are normalization rules and all three are already modelled; three are parser, expressibility or reference-selection constraints; two are `unmodellable` — keyed on provenance and on population frequency, so no work closes them and leaving them blank would read as a gap; and the remaining two are the endorsement itself and the derivation mandate.

Rules are keyed on the **property, not the axis label**, as the issue asks: what matters for the codon carve-out is having a reading frame, which correctly groups `n.` with `g.` rather than with `c.` — a pairing an axis-name-keyed table gets wrong.

## What is curated and what is checked

The assignment of a record to a stage is editorial. Nothing a reader might act on is:

- every cited record must exist in the ledger;
- every cited clause must resolve to real lines of the pinned spec checkout, and its quote must be found there — the same whitespace-collapsed containment the ledger's own citations are checked with, with the same stated limit (it catches a clause moving; it does not certify a quote is byte-exact);
- the stage list is a **closed set held in code**, so a stage cannot be dropped from the inventory by deleting its row, which is the one edit that would make it read as complete while covering less;
- no row may be blank, and no ungoverned decision may carry a verdict outside the two-value set — the genuine-gap / no-record-warranted distinction is what makes a row a finding.

**No status is written into the fixture.** A row cites a record; whether that record is decided or open is read from the ledger at render time. "This stage is settled" is precisely the claim the audit exists to test, so it may not be a copied string.

The document restates none of README's ruleset and is guarded against doing so, as its sibling is.

## Out of scope, deliberately

Partition + typing is settled at the level of principle and in flight at the level of shipped behaviour. #1553 places it out of scope for re-litigation; its row is an index into the records rather than an invitation to reopen them.

Full local suite on this branch: 10456 passed, 0 failed. `cargo fmt --check`, `cargo clippy --all-features --all-targets -- -D warnings` and the generated-doc `--check` gates all clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,6 +326,11 @@ from, and why. Those records are published in full, with their clause quotes, in
 That document is generated from the records and gated against them, so it cannot drift; it
 deliberately does not restate the seven rules above, which are stated only here.
 
+The inverse index — for each stage of the normalizer, *which* record or clause governs it, and
+which decisions are governed by **nothing** — is
+[docs/NORMALIZATION_STAGE_AUDIT.md](docs/NORMALIZATION_STAGE_AUDIT.md). Read that one if what
+you want to know is whether a behaviour you are looking at was chosen or merely happened.
+
 ## Deriving a description from sequences
 
 If what you have is **bases** rather than a description — a window out of a BAM, a VCF row, an
--- a/docs/NORMALIZATION_STAGE_AUDIT.md
+++ b/docs/NORMALIZATION_STAGE_AUDIT.md
@@ -1,0 +1,312 @@
+<!--
+GENERATED FILE — do not edit by hand.
+
+Rendered from tests/fixtures/grammar/normalization_stage_audit.json and the
+adjudication ledger, by tests/it/normalization_stage_audit_doc.rs. Edit the
+fixture, then regenerate:
+
+    BLESS_STAGE_AUDIT_DOC=1 cargo nextest run --features dev --test it \
+      -E 'test(normalization_stage_audit_doc)'
+-->
+
+# What governs each normalization stage
+
+Ferro's normalizer makes a decision at every stage of its pipeline. Some of those decisions are
+backed by a clause of the HGVS recommendations, some by an adjudication record, and some by
+nothing more than what the implementation happens to do. This document is the inventory of
+which is which, so that a deliberate choice can be told from an accident — by a contributor
+deciding a new case, and by a reviewer reading a diff.
+
+**7 stages. 3 of the records cited below are still open. 2 decisions are
+governed by nothing.**
+
+## How to read it
+
+Each stage names the decision it makes, the records that rule on it, and the clauses it is
+decided under. A record's status is read from the ledger when this file is generated and is
+never copied into the fixture, so a stage cannot be described as settled by a record that is
+not.
+
+The rows worth reading first are the **ungoverned** ones. Each is classified as either a
+*genuine gap* — a decision with real consequences that no clause and no record settles — or as
+*no record warranted*, meaning the spec settles it so plainly that writing a record would be
+ceremony. Only the first kind is a finding.
+
+## What this document is not
+
+**It is not the ruleset.** What ferro's output is allowed to be, and what happens where the
+spec determines no answer, is stated once, in
+[README.md, *Normalization rules*](../README.md#normalization-rules), and is deliberately not
+restated here.
+
+**It is not the records.** It says which record governs a stage; it does not reproduce the
+record. The records live in
+[`tests/fixtures/grammar/hgvs_spec_normalization_overrides.json`](../tests/fixtures/grammar/hgvs_spec_normalization_overrides.json).
+
+**It is not exhaustive about clauses.** A stage's clause list carries the clauses that decide
+it, not every clause that touches it.
+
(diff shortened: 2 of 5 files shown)


nh13:
The two ungoverned decisions this audit records are now filed as issues, so they are actionable independently of this PR: #1856 (cross-zone `c.` ordering) and #1857 (the 5' direction). Neither asks for a decision in this PR — the audit's job is to make them visible, and both are recorded in `docs/NORMALIZATION_STAGE_AUDIT.md` as ungoverned until they are settled.

nh13:
@coderabbitai pause

Issue #1859, opened 2026-08-13
test: no ClinVar round-trip coverage — the test that claimed it read a corpus that never existed


nh13:
## Summary

`idempotency_tests::test_clinvar_sample_parsing_idempotency` was deleted as a permanent no-op: it read `clinvar_1000.txt` from the repository root, a path that appears in no commit in the repository's history, is not one of the four `test-fixtures-v1` release assets, and is named by no script, workflow or `.gitignore` entry. It had only ever taken its `Err(_) => return` branch.

Its **intent** was real and is not covered anywhere else: parse → format → re-parse over a real ClinVar sample, asserting a 99% agreement floor. This issue is that intent, re-pointed at a corpus that exists.

## Why it is not already covered

Two things sit close to it and neither asks this question:

- `clinvar_hgvs_tests::test_clinvar_hgvs_500k_benchmark` and `::test_clinvar_hgvs_unique_benchmark` read the two fetchable ClinVar corpora, but they only **parse**. They assert a throughput floor and a per-input parse-failure snapshot; neither ever calls `format!` on a parsed variant, so a formatter that emits something the parser rejects — or something the parser accepts as a *different* variant — is invisible to them.
- `idempotency_tests::test_parsing_idempotency_edge_cases` and `::test_parsing_idempotency_spec_examples` do the round-trip, but over the curated spec and edge-case corpora: hundreds of hand-chosen strings, not the long tail of what ClinVar actually stores.

So the round trip is checked on a curated corpus and the real corpus is checked without the round trip. The deleted test was the only thing nominally covering the intersection, and it never ran.

## Proposed work

Add a round-trip guard over `tests/fixtures/bulk/clinvar_hgvs_unique.json.gz` (4.2M unique strings) or `clinvar_hgvs_500k.json.gz` (the CI-sized stratified sample), reading it through `crate::common::bulk_fixtures::present_or_skip` so absence skips locally and fails under `FERRO_REQUIRE_BULK_FIXTURES` — the mechanism that already governs every other reader of those files. For each input that parses: `format!` it, re-parse, and require the two renderings to agree.

## Read this before assuming it is a small change

[...]

Issue #1858, opened 2026-08-13
test: four of the five real_data_normalization_tests guards check nothing, even fully armed


nh13:
## Summary

`tests/it/real_data_normalization_tests.rs` is a reference-aware module whose guards are now guaranteed to **run** — #1851 wired it into the nightly and gave it the `FERRO_MANIFEST`-then-`benchmark-output` gate, and the follow-up made an unserved transcript a failure under `FERRO_REQUIRE_MANIFEST` instead of a green skip. Neither change made it guaranteed to **check** anything, and four of its five guards do not.

This is the same defect as the two above it, one layer further down: a test that passes by doing nothing. It is filed rather than fixed because deciding what these four should assert is a normalization question, not a CI-hygiene one.

## The census

Counted over the module as it stands:

| guard | `assert*!` | reachable? |
|---|---:|---|
| `test_nm_001408491_c517dela_should_shift` | 1 | yes — asserts `c.518del` after the bounds block |
| `test_potential_bug_deletion_shift_nm033517` | **0** | — |
| `test_potential_bug_delins_shift` | **0** | — |
| `test_5utr_duplication_shifting` | **0** | — |
| `test_compare_with_mutalyzer_sequence` | 2 | **conditionally** — both sit inside `if idx_517 < seq.len() && idx_518 < seq.len()` |

**Three of the five contain no assertion at all.** They are investigation scripts committed as tests: `test_potential_bug_delins_shift` prints `⚠️ POTENTIAL BUG: Ferro shifted delins from 2139_2140 to 2140_2141` to stdout and then returns successfully, so the condition it exists to detect is reported by *passing*. `test_5utr_duplication_shifting` closes with a comment saying the correct behaviour "depends on HGVS spec interpretation" and asserts neither reading.

**`test_compare_with_mutalyzer_sequence`'s two assertions are its entire body**, and they are inside a bounds check. An out-of-range coordinate — the exact condition the check exists for — leaves the test asserting nothing and passing.

What these four do still guard is narrow and worth stating so it is not double-counted: each carries two `.unwrap()`s and an `.expect()`, so they fail if `parse_hgvs` or `normalize` returns `Err`, or if the transcript carries no cached bases. That is a smoke test for "does not error", not a check on the output.

## Why it matters here specifically

[...]

PR #1861, opened 2026-08-13
test: make a present-but-incomplete reference fail, and delete a test whose corpus never existed

### LINKED ISSUE #1849: ci: 45 reference-aware guards run in no automated job — they skip green everywhere

nh13:
**#1851 has merged** (`afcc391b`), so this PR is now based on `main` and the diff is the two commits that are actually mine. It was developed stacked on #1851 rather than branched independently: both changes edit the five call sites #1851 rewrites and both hang off `FERRO_REQUIRE_MANIFEST`, so an independent branch would have conflicted textually *and* shipped a second promotion mechanism — which is the drift `manifest_gate.rs` exists to prevent.

Addresses two residuals #1851 recorded in #1849 and deliberately left. Does **not** close #1849, which carries more than these two.

## 1. A manifest that is present but lacks the transcript

`real_data_normalization_tests` finds a manifest, starts, spends twelve seconds loading a 546,290-sequence provider, asks for one transcript, and returns on `Skipping test: transcript not found`. `FERRO_REQUIRE_MANIFEST` never sees it — the manifest is present, so the promotion reads as armed while the guard body does nothing.

**Why this one is worse than the class #1851 closed, and not merely smaller.** The census that found those 45 guards was a *duration* diff: 0.01–0.08 s skipped against 9–58 s armed, with nothing in between. That instrument is blind here. The provider load happens **before** the lookup, so a stood-down guard costs the same wall clock as an armed one — measured in a single run below, 12.086 s FAIL against 12.027 s PASS. No exit status, summary line, check row **or runtime** separates them.

### Same flag, not a sibling — the argument

The distinction that matters is *configuration-dependent and announced* against *silent coverage loss*, and this falls on the second side. A guard that stands down because the reference does not carry `NM_001408491.1` has lost exactly the coverage a guard that stands down because there is no reference has lost. There is no third state in which the manifest is present, the accession is missing, and the guard still checked something. The promotion is about the coverage, not about the cause, so it is one variable.

The *message* is a different question, and gets a different answer: `unserved` carries its own text because the remedy is not `ferro prepare` — pointing a reader at a manifest that is fine is how a guard gets worked around instead of fixed. Two unit tests pin that the two messages stay distinct, because the cheap implementation is to call `absent` with a concatenated context and silently give both causes the same fix.

The skip that stays a skip is the local one: with `FERRO_REQUIRE_MANIFEST` unset, absence still reports on stderr and passes, which is the useful default for a developer without a prepared reference and the condition PR CI runs under.

### Proof — absence FAILS, presence PASSES

Same binary, same commit, blessed scratch reference (`aa8b3246d83055cc`), only the accession perturbed and restored with `cp`. Every `EXIT=` read back out of its own log file.

**Before, manifest present + `FERRO_REQUIRE_MANIFEST=1`, accession absent** — the defect:

```
Skipping test: transcript not found: Reference not found: NM_099999999.9
Skipping test: transcript not found: Reference not found: NM_099999999.9
     Summary [  13.343s] 5 tests run: 5 passed, 6293 skipped
EXIT=0
```

**After, same conditions** — absence is now a failure, and only the two guards naming the missing accession:

```
        PASS [  12.027s] (1/5) real_data_normalization_tests::test_5utr_duplication_shifting
        FAIL [  12.086s] (2/5) real_data_normalization_tests::test_nm_001408491_c517dela_should_shift
        FAIL [  12.099s] (3/5) real_data_normalization_tests::test_compare_with_mutalyzer_sequence
     Summary [  12.361s] 5 tests run: 3 passed, 2 failed, 6292 skipped
EXIT=100
```

The PASS and the two FAILs are within 0.07 s of each other. That is the duration blindness, measured rather than argued.

**After, accession restored, same conditions** — presence passes, armed:

```
     Summary [  18.294s] 5 tests run: 5 passed, 6292 skipped
EXIT=0
```

Zero `skipping` lines in that log, and 16–18 s per guard confirms arming rather than skipping.

**After, accession absent, `FERRO_REQUIRE_MANIFEST` unset** — the local and PR-CI default is unchanged:

```
real_data_normalization_tests: skipping — the reference manifest does not serve `NM_099999999.9`: Reference not found: NM_099999999.9
     Summary [   9.585s] 5 tests run: 5 passed, 6292 skipped
EXIT=0
```

### What this buys, stated precisely

Execution, not verification. Three of the five guards in that module contain **no assertion at all** and a fourth keeps both of its assertions inside a bounds check, so a green run here is not coverage of those four. That is a third instance of the same defect and it is **filed as #1858, not fixed** — deciding what those four should assert is a normalization adjudication, not CI hygiene, and folding it in would have widened this past what #1851 declined to widen. The module header now says so at the point a reader would otherwise assume otherwise.

## 2. `test_clinvar_sample_parsing_idempotency` — verified unfetchable, deleted

The claim in #1849 checks out, and I checked it before acting rather than after. `clinvar_1000.txt`:

- appears in **no commit** in the repository's history — `git log --all --oneline -- '*clinvar_1000*'` is empty;
- is **not** one of the four `test-fixtures-v1` release assets (`clinvar_hgvs_500k.json.gz`, `clinvar_hgvs_unique.json.gz`, `cmrg_genes_exhaustive.json.gz`, `paraphase_genes_exhaustive.json.gz`);
- is absent from `tests/fixtures/CHECKSUMS.sha256` and from `scripts/fetch-test-fixtures.sh`;
- is named by no workflow, no other script and no `.gitignore` entry — the single occurrence in the entire tree is the `let clinvar_path = ...` line itself.

The LFS-removal history is the thing that would make a missing file innocent, and it does not apply: the four corpora it moved to release assets are all accounted for and this is not one of them. **Verdict: genuinely unfetchable.** The test has only ever taken its `Err(_) => return` branch, in every environment, since it was written.

### Recommendation: delete — and the argument against the other two

- **`#[ignore]` with a reason** means "run this when you can". There is no condition under which anyone can, so it is a to-do wearing a test's clothes, and — because the body still returns on `Err` — even `--run-ignored all` would pass it green. It converts a silent no-op into an advertised one without making it runnable.
- **Fail loudly** is the right answer when the input is *meant* to exist and a fetch may have failed quietly. That is precisely the `FERRO_REQUIRE_BULK_FIXTURES` case and precisely why that mechanism is correct there. Here the file is not meant to exist anywhere: promoting its absence would turn every run in the project red for a file nobody can obtain, forever.
- **Delete** removes nothing that has ever executed, and the intent is preserved where intent belongs — see below.

### The intent is real, uncovered, and filed rather than reimplemented

The deleted test's property — parse → format → re-parse over a real ClinVar sample at a 99% floor — is **not** covered elsewhere. `clinvar_hgvs_tests` reads both fetchable ClinVar corpora and only *parses*; it never calls `format!` on a parsed variant. The two round-trip tests that do exist run over the curated spec and edge-case corpora. So the round trip is checked on a curated corpus and the real corpus is checked without the round trip.

Re-pointing it at `clinvar_hgvs_unique.json.gz` under the existing `bulk_fixtures` promotion is the right successor and is **#1859**, not this PR. The reason is not tidiness: that 99% floor was **never once evaluated**, so it is a guessed number, and measuring the property over 4.2M real strings for the first time may legitimately come out red. That would be a finding about the formatter. A change that removes a dead test must not be the vehicle for an unmeasured assertion over a 4.2M-row corpus.

## Verification

```
cargo fmt --all --check                                       FMT_EXIT=0
cargo clippy --all-features --all-targets -- -D warnings       CLIPPY_EXIT=0
cargo nextest run --features dev --no-fail-fast (manifest-free) EXIT=0
    Summary [ 195.061s] 10635 tests run: 10635 passed, 152 skipped
    0 FAIL lines; 0 manifest-skip lines
cargo nextest run --features dev --lib -E 'test(manifest_gate)'  EXIT=0
    5 tests run: 5 passed
```

**Re-verified after the rebase onto `main` (`d4552167`, i.e. past #1851's squash `afcc391b`).** The rebase replayed only this PR's own two commits; the four files they touch are byte-identical to what the run above measured. With the blessed scratch reference armed (`FERRO_MANIFEST` set, `aa8b3246d83055cc`):

```
cargo fmt --all --check                                        FMT_EXIT=0
cargo nextest run --features dev --no-fail-fast                 EXIT=0
    Summary [ 130.321s] 10681 tests run: 10681 passed, 152 skipped
    0 FAIL lines; 0 `Skipping`/`skipping` lines
```

Zero skip lines is the check that matters for this change, and it is not a duration inference: the five `real_data_normalization_tests` guards each cost 8.5–8.8 s (armed), and all six `manifest_gate` unit tests ran.

The manifest-free run is the PR-CI condition and the one that matters for this change: `FERRO_REQUIRE_MANIFEST` is unset there, so every promoted guard must skip exactly as before. It does. The count reconciles against #1851's `10634 passed` exactly — minus the deleted ClinVar test, plus the two new `manifest_gate` unit tests.

## Filed, not fixed

Three, all recorded rather than folded in — #1851's author declined to widen in passing and so does this.

- **#1858** — three of five `real_data_normalization_tests` guards carry no assertion, and a fourth keeps both of its assertions inside a bounds check. The third instance of this defect, found while making the module run.
- **#1859** — the ClinVar round-trip coverage the deleted test claimed and never provided. Blocked on a measurement that may come out red.
- **#1862** — `coverage_gap_tests::test_flt3_itd_insertion_position_order` and `normalize_tests::test_real_repeat_pattern` are `#[ignore]`d **and** relative-path-gated, so they sit in #1851's census blind spot twice over. The first carries two real assertions pinning the 5'-to-3' intronic-flank question against a named Mutalyzer disagreement, and has never executed.

Related to #1849, which carries residuals beyond these.

Representation-Change: none. Test gating and a test deletion only; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched.


nh13:
@coderabbitai pause

Issue #1860, opened 2026-08-13
projector's transcript cache does not reach the normalizer, so each re-derived axis re-fetches


nh13:
`VariantProjector` caches transcript fetches so that projecting N variants against one transcript costs one `ReferenceProvider::get_transcript` call. `projection::transcript_cache_collapses_repeat_lookups` pins that: three substitutions on `NM_TEST.1`, one fetch.

The cache is reached through `cached_get_transcript_for_variant`, which every *projector-side* derivation calls. The **normalizer** the projector holds does not go through it — it fetches from the provider directly. So any axis the projector re-derives by handing a variant to `self.normalizer` re-pays the transcript fetch, once per variant, however warm the projector's own cache is.

## How it surfaced

Closing #1712 makes the projector re-derive the `n.` axis on its own reference (`noncoding_axis_with_reason` now calls `normalize_or_fallback`), which routes a normalization through the normalizer per projected variant. The guard above moves from **1** `get_transcript` call to **10** across the same three variants. The pin was re-blessed there with a pointer to this issue; the underlying reach gap was not touched, because it is not that PR's subject and fixing it there would have mixed a performance change into an adjudication.

The same gap is already latent on the genomic axis — `reanchor_and_normalize_genomic` and `project_to_genomic_normalized` have called the normalizer since #737/#867 — so this is not new with #1712, only newly visible on a guard that counts.

## What would close it

Give the normalizer a provider view that shares the projector's transcript cache, so `get_transcript` is served from the same map on both sides. Options worth weighing rather than a prescription:

- wrap the projector's provider in a caching decorator and hand *that* to the normalizer, so the cache sits below both rather than beside one;
- or lift the existing per-projector cache into the shared provider handle.

The first keeps `VariantProjector`'s cache where it is and is likely the smaller change; the second removes the duplication rather than routing around it.

## Guard to re-pin when it lands

`tests/it/projection.rs::transcript_cache_collapses_repeat_lookups` — currently pinned at 10 with a comment naming this issue and the reason for each direction of drift. Closing this should take it back toward 1, and the pin is deliberately exact so that happening is not silent.

PR #1865, opened 2026-08-13
fix(project): re-derive the n. axis on its own reference

### LINKED ISSUE #1712: projector emits non-idempotent n. axes — 21 of 681, including one that is not 3'-shifted

nh13:
Closes #1712.

Representation-Change: 20 of 681 projected `n.` axes move — 16 respell as a repeat form, 4 split into members — and 6 of 1179 spec-enumeration projection rows move status with them. Projector output only; no normalization of a user's input moves. The count is partition-sensitive and must be re-measured on #1835's flipped default before this leaves draft.

## Why this is a DRAFT

The residue below was measured on the **live arm** (`FERRO_PARTITION` unset). PR #1835 makes canonical-coalesced the default, which moves the *normalizer* — one of the two sides #1712 compares. Four of the moved rows are literally "the output re-partitions into members" and the sixteen repeat-form rows turn on repeat maximization, so **the number does not carry across the flip**.

So before this leaves draft:

- re-measure `axis_noncoding_idempotent` on the flipped default, and
- re-bless the guards below against the **post-flip** answer, so they are blessed once rather than twice.

The *grounds* for the ruling are not partition-sensitive — they concern what an axis prefix claims about its reference, not how blocks are cut — which is why the change is safe to write now and only the landing waits.

## What this does

The projector re-derives the `n.` axis on its own reference (`noncoding_axis_with_reason` now normalizes the reframed form), instead of reframing the `c.` frame's edit by CDS-offset arithmetic and inheriting whatever form that frame chose because it declares a reading frame. That is the symmetry with the genomic axis's `normalize_or_fallback` (#737), completed.

This is the **operator ruling on #1712**, direction A, recorded at <https://github.com/fulcrumgenomics/ferro-hgvs/issues/1712#issuecomment-5286746322>. It settles the half the decided record `projection-codon-exception-is-decided-by-the-rendered-axis` explicitly parked — "Which side of that is wrong is an open question, deliberately left open".

**The rule's home is the ledger, not this description.** The new `decided` record is `rulings[noncoding-axis-is-re-derived-on-its-own-reference]` in `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json`; it carries the grounds, the refused direction, the scope, and the accepted `n.`/`r.` divergence. Read the record.

In one line, since the shape of the argument decides how to review the diff: `general.md:23` makes the prefix a mandatory claim about the **type of reference sequence** and `:28` assigns `n` to a non-coding DNA one, so `DNA/delins.md:42`'s antecedent — "together affecting one amino acid" — is no more statable there than on a genomic reference.

**Accepted cost, stated rather than discovered later:** `r.` merges on `RNA/delins.md:18`'s own authority, so one change now spells differently on `n.` and `r.`. Spec-licensed — the same shape as the `g.`/`c.` divergence the parked record already accepts — not a confluence bug.

## The measurement, and why CI does not witness it

Against the prepared GRCh38 reference (blessed identity `aa8b3246d83055cc`), on the live arm, at this branch's base:

| | tested | non-idempotent |
|---|---|---|
| `axis_noncoding_idempotent` before | 681 | **20** |
| `axis_noncoding_idempotent` after | 681 | **0** |
| `axis_genomic_idempotent` (control) | 214 | **0** both ways |

**The baseline is 20 and #1712 reports 21 — that is not a discrepancy.** #1844, the branch this is stacked on, already closed the 21st row: `LRG_24t1:n.245_252del`, the #860 LRG namespace echo renaming the axis onto a transcript with eight extra 5' bases. That was a rule 1 denotation defect, not this question. So 21 → 20 is #1844's step and 20 → 0 is this PR's.

**The genomic control is the non-vacuity evidence, and it is the reason 20 → 0 is a fix rather than an artefact.** `axis_noncoding_idempotent` counts failures over axes the fan-out produced, so a change that stopped producing `n.` axes would also read 0. Two things rule that out: the `tested` count is unchanged at 681 on both runs, and `axis_genomic_idempotent` reads 214 tested / 0 non-idempotent on both — measured in the same process, against the same reference. That is what isolates the movement to the non-coding path rather than to normalization generally, or to the harness.

**A green CI run does not witness any of this.** PR CI never sets `FERRO_MANIFEST` (`ci.yml:1153-1158`), so every one of these rows takes the early-return skip path — and `nextest` reports that path as **PASS**, not SKIP. A run without the manifest looks green whatever the pins say. Both figures above come from a local run with the manifest set, and each test took ~15–35 s rather than the sub-second the skip path costs.

### So the pre-merge coverage is deliberately manifest-free

`axis_noncoding_idempotent` is the measurement, and it is **nightly-only**. Three guards carry the contract on PR CI, none of which touches a prepared reference:

- `projection::every_projected_noncoding_axis_is_a_fixed_point` — **new in this PR**, and the one that closes the real gap. The two re-blessed `lenient_overlap` guards below are both `delins` cases, so they reach the 4-row member shape and not the **16-row repeat shape**, which is the majority of what moved. This asserts the same contract as `axis_noncoding_idempotent` on a `MockProvider` fixture built programmatically: a transcript with a 5'UTR (so `n.` and `c.` do not coincide and a reframing bug cannot hide) carrying a `T` homopolymer and a `GC` tract.

  Its anchor row is the ruling in one line — both axes denote the same two bases:

  ```text
  NM_RPT.1:c.4_5del  ->  c.  c.6_7del     3'-shifted, still a plain `del`
                         n.  n.6_9T[2]    re-derived on `n.`, tract MAXIMIZED
  ```

- `the_codon_delins_merges_only_on_an_axis_that_declares_a_frame` — committed reference slice, no manifest.
- `the_same_codon_delins_renders_alike_whichever_axis_it_is_authored_on` — synthetic `MockProvider`.

**Each was verified to fail without the fix rather than assumed to.** Reverting `noncoding_axis_with_reason`'s `normalize_or_fallback` call makes the new test's anchor row read `n.8_9del` — the plain CDS-offset image of the `c.` answer — instead of `n.6_9T[2]`.

## The three guards this turns red — all three re-blessed, none weakened

Note #1545's "may not be weakened" bars loosening a **red** guard. Re-blessing a pin against a **decided** ruling is a different act, and each of these says which it is doing and on what authority.

1. **`lenient_overlap_and_projection_split::the_codon_delins_merges_only_on_an_axis_that_declares_a_frame`** — `split == ['g']` → `['g', 'n']`. Its own doc comment recorded the `n.` row as *"pinned as OBSERVED, not as adjudicated"* and named this exact open question; it was written to be re-blessed by a ruling, and this is that ruling. **Strengthened, not loosened**: the exact `n.` string is now pinned too (`LRG_199t1:n.[389C>T;391C>G]`), matching what the test already did for `g.`.

2. **`lenient_overlap_and_projection_split::the_same_codon_delins_renders_alike_whichever_axis_it_is_authored_on`** — `('n', "tx:n.4_6delinsATA")` → `('n', "tx:n.[4C>A;6G>A]")` on both spellings. The test's *subject* is unchanged and still holds: one variant renders identically whether authored on `c.` or on `g.`. Its property assertion was written as `transcript_splits == 0 && == 0`, which conflated "the two spellings agree" (this test's claim) with "nothing splits" (a fact about which axes declare a frame, decided elsewhere). It is restated as the agreement it always meant, plus a non-vacuity assertion that the splitting axis is exactly `n.` — so two empty vectors cannot satisfy it.

3. **`projection::transcript_cache_collapses_repeat_lookups`** — 1 → 10 `get_transcript` calls. **This is a real performance regression and it is deliberately not fixed here.** The normalizer's provider fetches do not route through the projector's transcript cache, so every axis the projector re-derives by normalizing re-pays the fetch. The gap predates this PR on the genomic axis (`reanchor_and_normalize_genomic` and `project_to_genomic_normalized` have normalized since #737/#867); it is newly *visible* because this guard counts. Filed as **#1860**. The pin stays an exact equality, so **fewer** calls fails it too — closing #1860 cannot land silently.

## A fourth guard moved that was not on the brief's list: the enumeration census

`spec_enumeration_tests` has two constants that must always be read as a pair, and they moved in equal and opposite steps — which is the check that says no row left the enumeration rather than moving between statuses:

- `DIVERGENCE_BUDGET` / `ProjectionSplitsSingleMember`: **12 → 18**
- `PASSING_CENSUS` / `ProjectionPinned`: **1167 → 1161**

**All six entering rows are `project-n` rows**, which is the property to check before reading anything else, since the ruling reaches exactly one axis:

```text
project-n/LRG_199t1:c.145_147delinsTGG   -> LRG_199t1:n.[389C>T;391C>G]
project-n/LRG_199t1:c.235_237delinsTAT   -> LRG_199t1:n.[479A>T;481G>T]
project-n/NM_004006.2:c.145_147delinsTGG -> NM_004006.2:n.[389C>T;391C>G]
project-n/c.235_237delinsTAT             -> NM_004006.2:n.[479A>T;481G>T]
project-n/c.4375_4376delCGinsAGTT        -> NM_004006.2:n.[4619C>A;4620_4621insTT]
project-n/c.4375_4376delinsAGTT          -> NM_004006.2:n.[4619C>A;4620_4621insTT]
```

The last two are two spellings of one input, so the six rows are five distinct variants. Rows 2 and 4 are an independent confirmation of the `n.[479A>T;481G>T]` string pinned in guard 1 above — it was derived by hand there and measured here.

**These are the `n.` analogue of the two `project-g/LRG_199t1` rows already parked in that status, on the identical ground**, and they should be read as the ruling's cost rather than as a regression. The parked record said so about its own two in as many words: they "KEEP sitting in `ProjectionSplitsSingleMember`, so a reader of that budget alone sees two rows parked in a divergence status indefinitely. That is the ruling and not a regression." Six more now sit there for the same reason. Taking the budget back **below** 18 is exactly what the ruling refuses.

## Ledger and generated artifacts

- New `decided` record `noncoding-axis-is-re-derived-on-its-own-reference`, citing `general.md:23`/`:28`, `DNA/delins.md:17`/`:18`/`:42`, `RNA/delins.md:18` and `docs/background/numbering.md:52`, all quote-verified against the pinned spec checkout.
- `RULING_STATUSES` gains the record; the neighbouring `projection-codon-exception-…` comment's "the `n.` axis is left open" is scoped to *that record* rather than left reading as a live open question.
- `CLAUDE.md`'s decided table gains its row (`claude_md_adjudication_tables` enforces the partition).
- `clause_ruling_index.rs`: census pin and the generated index block regenerated, not hand-edited.

## Stacking

Based on **#1844** (`fix/issue-1712-noncoding-axis-idempotent`), not on `main`. That PR ships the one row of #1712 that was a genuine rule 1 denotation bug — the #860 LRG namespace echo renaming an axis onto `LRG_24t1`, which carries eight extra 5' bases — and touches the same file and the same pins. It declares `Refs #1712` because it closes only that row; this one declares `Closes #1712`. The 21 → 20 step in the residue is #1844's; the 20 → 0 step is this PR's.

Deliberately **not** stacked on #1835. That is a *measurement* dependency, not a code one, and a PR cannot be stacked on both.

## Local verification

Full local suite with `FERRO_MANIFEST` set: **10,616 tests run, 10,613 passed, 3 failed** — read off the run's own `EXIT=100` line rather than a completion notification. All three failures were stale pins that this change is what moves, and all three are updated in this commit:

| failure | pin | resolution |
|---|---|---|
| `clause_ruling_index::clauses_named_by_several_records_are_flagged` | multiply-named clause lines 28 → 32 | the new record cites 7 clauses, 4 already named by another record |
| `spec_enumeration_tests::divergence_budget_is_unchanged` | 12 → 18 | the six `project-n` rows above |
| `spec_enumeration_tests::passing_census_is_unchanged` | 1167 → 1161 | the same six, equal and opposite |

Re-run of all 85 affected tests after the updates: **85 passed, 0 failed, `EXIT=0`**, with `axis_noncoding_idempotent` again taking ~15 s rather than the skip path's sub-second. `cargo fmt --all -- --check` clean; the ledger set-comparison shows nothing dropped from the base or from either pre-amend tip, and exactly one record added.

**One expected wrinkle at rebase time.** `origin/main` has gained `inversion-vs-a-mixed-member-competitor` since #1844 branched, so this branch's ledger holds 28 records where `main` holds 28 including that one and excluding mine. Nothing is lost — the base lacks it too — but the `clause_ruling_index` census pins will need re-reading once this rebases onto a `main` that carries both.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved normalization of projected non-coding (`n.`) variants by deriving them against their own reference.
  * Preserved correct transcript and accession handling when reference sequences differ.
  * Improved idempotency for projected variants, including repeat-tract changes.
  * Corrected projection axis splitting and resulting `n.`/`r.` representations.
* **Tests**
  * Added regression coverage for non-coding projections, LRG transcript handling, and normalization consistency.
* **Documentation**
  * Documented the new non-coding projection rule and updated related specification indexes.


nh13:
@coderabbitai pause

nh13:
@coderabbitai review

Issue #1857, opened 2026-08-13
The 5' shifting direction is adjudicated by nothing — every clause and record at that stage is 3'-only


nh13:
Found by the stage audit (#1553, PR #1855), which classifies it as a **genuine gap**.

## The decision nothing governs

What the 5' direction does — at an exon/exon junction, inside a repeat tract, and at a reference boundary.

## Why it is a gap

The recommendations state a 3' rule and only a 3' rule. `general.md:41`:

> **3'rule**: for all descriptions, the most 3' position possible of the reference sequence is arbitrarily assigned to have been changed.

`5' rule` and `5'-rule` occur **nowhere** in the pinned `assets/hgvs-nomenclature` checkout (`6f85311`):

```sh
grep -rn "5'-rule\|5' rule\|five prime rule" assets/hgvs-nomenclature/docs/    # no output
```

Ferro nevertheless offers a 5' direction, and the direction is **not** an orthogonal knob. README rule 6 says so in terms: it selects the frame every rule is evaluated in, which is why rules 2 and 3 are claimed *per direction* rather than across the two.

So every authority at the shifting stage is authority for the 3' arm:

- `general.md:41` — the rule itself, 3' by name;
- `general.md:44` — the exon/exon junction exception, stated for the 3' rule;
- `DNA/complex.md:50` — the break-point placement, "(the 3' rule)";
- `exon-junction-dup-converge-from-the-far-side` — the ruling record, whose whole subject is the most 3' position that does not cross a junction.

The 5' arm's answers are ferro's own, mirrored by construction rather than by a decision anyone recorded.

## Why it is worth adjudicating rather than assuming

The mirroring may well be the only possible answer, and if so a short record saying that is cheap. But it is least obviously right exactly where the 3' answer is itself an exception:

- **At a junction.** `exon-junction-dup-converge-from-the-far-side` rules that the canonical position is the most 3' position that does not cross a junction, reached from either side. The 5' mirror of "does not cross a junction" is not simply the reflection of that predicate — the exception exists because a junction is not a symmetric feature of a transcript.
- **At a reference boundary.** The 3' clamp and the 5' clamp are against different ends with different zone semantics on `c.`, which is the sibling gap in this audit (see the cross-zone ordering issue).
[...]

Issue #1016, opened 2026-07-09
Python: unrecognized `direction` argument silently defaults to 3' instead of erroring


nh13:
## Summary

`parse_direction` silently maps any unrecognized shuffle-direction string to `ThreePrime`, so a misspelled or bogus `direction=` argument quietly changes normalization output instead of failing loudly.

`src/python_helpers.rs`:

```rust
pub fn parse_direction(direction: &str) -> ShuffleDirection {
    match direction.to_lowercase().as_str() {
        "5prime" | "5'" | "5" => ShuffleDirection::FivePrime,
        _ => ShuffleDirection::ThreePrime,   // anything unrecognized -> 3' silently
    }
}
```

So `Normalizer(direction="5prim")`, `direction="five"`, `direction="3prine"`, etc. are all silently accepted as 3'. The documented values are only `"3prime"` and `"5prime"`; anything else should be a hard error. Direction directly changes the normalized result, so a silent wrong default is a correctness footgun, not just an ergonomics one.

Affected call sites (all route through `parse_direction`): the `Normalizer` constructor, `VariantProjector` (`new` / `from_manifest` / `from_provider`), the module-level `normalize()` free function, and `HgvsVariant.normalize()`.

## The fix already exists next door

The sibling `assembly` argument does this correctly — `VariantProjector::from_provider` (`src/python.rs`) validates at the boundary and raises:

```
PyValueError: unrecognized assembly "…"; expected GRCh37 or GRCh38 (aliases hg19/hg38 accepted)
```

`direction` should mirror that: validate at the Python boundary and raise `PyValueError` for anything that isn't a recognized direction alias.

## Proposed change

- Make `parse_direction` fallible (e.g. return `Option<ShuffleDirection>` / `Result`), accepting the existing aliases (`3prime`/`3'`/`3` and `5prime`/`5'`/`5`) and rejecting everything else.
- At each Python call site, map a rejection to `PyValueError` with a message naming the accepted values.
- Add tests asserting an unrecognized `direction` raises `ValueError` (Python) and that all documented aliases still resolve.

## Acceptance criteria

- [ ] An unrecognized `direction` argument raises `ValueError` instead of silently defaulting to 3'
- [ ] All documented direction aliases (`3prime`/`5prime`, plus existing `3`/`5`/`3'`/`5'` forms) still resolve correctly
- [ ] Every Python entry point that takes `direction` (constructor, free function, method) is covered

PR #1866, opened 2026-08-13
fix(cli): reject an unrecognized --direction instead of silently 3'-shifting

### LINKED ISSUE #1863: ferro normalize --direction silently accepts garbage and 3'-shifts (the CLI half of #1016)
## Summary

`ferro normalize --direction <anything>` silently accepts an unrecognized value, runs a **3'** normalization, and exits **0**. The user gets a plausible answer to a question they did not ask.

This is exactly the defect #1016 filed and #1017 fixed — but #1017 fixed only the **Python** boundary. The CLI has its own, separate parsing function with the same catch-all, and it was left in place.

## Mechanism

Two halves, both on `origin/main` (`64fc76df`):

**1. The flag has no `value_parser`** — `src/bin/ferro.rs:284-286`:

```rust
/// Shuffle direction (3prime or 5prime)
#[arg(long, default_value = "3prime")]
direction: String,
```

Its immediate siblings in the same `Normalize` variant both constrain their values, so the omission is a local inconsistency rather than a house style:

- `src/bin/ferro.rs:281` — `#[arg(short = 'f', long, default_value = "text", value_parser = ["text", "json", "tsv"])]`
- `src/bin/ferro.rs:293` — `#[arg(long, default_value = "strict", value_parser = ["strict", "lenient", "silent"])]`

**2. The parse function swallows everything else** — `src/cli/parse.rs:50-55`:

```rust
pub fn parse_shuffle_direction(direction: &str) -> ShuffleDirection {
    match direction.to_lowercase().as_str() {
        "5prime" | "5'" | "5" => ShuffleDirection::FivePrime,
        _ => ShuffleDirection::ThreePrime,
    }
}
```

Note the 3' spellings (`3prime`, `3'`, `3`) are **not** matched explicitly — they resolve only by falling into the same catch-all that swallows `banana`. The single production caller is `src/bin/ferro.rs:1634`.

The silent default is also **documented and asserted** as intended behaviour, in a doctest (`src/cli/parse.rs:48`) and a unit test (`src/cli/parse.rs:269-275`):

```rust
/// assert!(matches!(parse_shuffle_direction("unknown"), ShuffleDirection::ThreePrime));
```

## The asymmetry with Python

#1017 made the Python-side `parse_direction` fallible — `src/python_helpers.rs:105-111` returns `Option<ShuffleDirection>`, listing both arms explicitly and returning `None` otherwise, with a boundary helper raising `PyValueError`. So today:

| surface | unrecognized `direction` |
|---|---|
| Python (`normalize()`, `Normalizer(...)`, `HgvsVariant.normalize()`) | `ValueError` — fixed by #1017 |
| Rust CLI (`ferro normalize --direction`) | **silently 3'-shifts, exit 0** |

#1017's description states "`parse_direction` has no non-Python (pure-Rust/CLI) callers, so no core path needed changes." That is true of `parse_direction` specifically, and it is why the CLI's near-duplicate `parse_shuffle_direction` was never brought into scope. The two functions are independent copies of the same logic; only one was fixed.

## Why it matters twice over

Beyond being a silent-failure bug in its own right, this blocks a clean deprecation path. #1857 records that the 5' direction is adjudicated by nothing, and removal is under consideration. **With this fallback in place, removing the 5' direction would turn every existing `--direction 5prime` invocation into a silent 3' run** rather than a clear error — converting a deprecation into data corruption at the customer.

Fixing this is a precondition for that removal, and is worth doing regardless of whether the removal happens. This issue does **not** propose removing the 5' direction; that decision is the operator's and is open in #1857.

## Proposed change

- Constrain `--direction` with clap's own `value_parser`, mirroring `--format` and `--error-mode`, so the rejection is a standard clap usage error naming the accepted values — consistent with the rest of the CLI rather than hand-rolled.
- Accept exactly the alias set Python accepts, so the two surfaces agree: `3prime`/`3'`/`3` and `5prime`/`5'`/`5`, case-insensitively (the current function lowercases, so `5PRIME` works today and must keep working).
- Decide what to do about `pub fn parse_shuffle_direction`'s catch-all, noting it is public API (re-exported at `src/cli/mod.rs:18`) and doctested, so a signature or behaviour change carries a SemVer obligation.

## Acceptance criteria

- [ ] `ferro normalize --direction banana` exits **non-zero** with a message naming the accepted values
- [ ] Every currently-valid spelling still works: `3prime`, `3'`, `3`, `5prime`, `5'`, `5`, and mixed case
- [ ] The CLI seam is covered by a test that drives the **binary**, not the library

nh13:
## Problem

`ferro normalize --direction <anything>` silently accepted an unrecognized value, ran a **3'** normalization, and exited **0**. Measured on `origin/main` at `64fc76df`, `--direction banana` produced output byte-identical to `--direction 3prime`:

```
$ ferro normalize --direction banana NM_000088.3:c.16del
NM_000088.3:c.16del -> NM_000088.3:c.19del      # rc=0
$ ferro normalize --direction 3prime NM_000088.3:c.16del
NM_000088.3:c.16del -> NM_000088.3:c.19del      # rc=0  — identical
$ ferro normalize --direction 5prime NM_000088.3:c.16del
NM_000088.3:c.16del                             # rc=0  — genuinely different
```

The user asked for a direction that does not exist and was given 3' without a word. Direction changes the answer, so this is a correctness footgun rather than an ergonomics one.

Two halves produced it:

- **`src/bin/ferro.rs:284-286`** declared the flag with no `value_parser`, unlike its immediate siblings — `format` at `:281` (`["text", "json", "tsv"]`) and `error_mode` at `:293` (`["strict", "lenient", "silent"]`). The omission was a local inconsistency, not a house style.
- **`src/cli/parse.rs:50-55`** ended in `_ => ShuffleDirection::ThreePrime`. Note the 3' spellings were **not** matched explicitly — they resolved only by falling into the same catch-all that swallowed `banana`.

## This is the CLI half of #1016

#1017 fixed the **Python** boundary only: it made `python_helpers::parse_direction` fallible (`src/python_helpers.rs:105-111`) and raised `PyValueError`. Its description reads "`parse_direction` has no non-Python (pure-Rust/CLI) callers, so no core path needed changes" — true of `parse_direction` itself, and exactly why the CLI's independent near-duplicate `parse_shuffle_direction` was never brought into scope. The two functions are separate copies of the same logic; only one was fixed. This closes the other.

**Should the Rust fix mirror the Python one or differ?** It should differ in *mechanism* and agree on *policy*. Python must validate at runtime because an arbitrary string arrives at the API boundary, so a hand-written `Option` + `PyValueError` is the only option there. The CLI has clap, which rejects before any code runs and produces the same usage error as every other constrained flag — so this uses `value_parser` rather than hand-rolling a check, per the brief. The **accepted value set is identical** to Python's, deliberately, so the two surfaces cannot drift.

## Fix

`--direction` now carries clap's own `value_parser` listing the six aliases, plus `ignore_case = true`:

```rust
#[arg(
    long,
    default_value = "3prime",
    ignore_case = true,
    value_parser = ["3prime", "3'", "3", "5prime", "5'", "5"]
)]
direction: String,
```

`ignore_case` is not decoration: `parse_shuffle_direction` lowercases its input, so `--direction 5PRIME` works on `main` today and dropping it would be a gratuitous break. The rejection is now a standard clap error naming the accepted values, and exits **2**:

```
error: invalid value 'banana' for '--direction <DIRECTION>'
  [possible values: 3prime, 3', 3, 5prime, 5', 5]
```

## Why this is worth doing now

Beyond the silent-failure bug itself: #1857 records that the 5' direction is adjudicated by nothing, and its removal is under consideration (scoping report exists). **With the catch-all reachable, removing the 5' direction would turn every existing `--direction 5prime` invocation into a silent 3' run** rather than a clear error — converting a deprecation into data corruption at the customer. This PR is a precondition for that removal and stands on its own regardless of whether it happens.

**This PR does not remove the 5' direction.** That decision is the operator's and remains open in #1857.

## `parse_shuffle_direction` is public API — and is deliberately left alone

It is `pub`, re-exported at `src/cli/mod.rs:18`, and **doctested**, so it is documented API. Its signature and behaviour are **unchanged** here, which makes this PR non-breaking: no SemVer bump is implied, and no downstream caller can be affected.

That is a deliberate scoping decision rather than an oversight, for three reasons:

1. **It buys nothing for this defect.** With clap validating upstream, the catch-all is unreachable from the CLI — the surface the issue is about.
2. **Tightening it to `Option<ShuffleDirection>` (mirroring `python_helpers::parse_direction`) is a breaking change to a documented public function.** At `0.13.1` that is a `0.14.0` minor bump under Cargo's SemVer rules. Shipping that inside a bug-fix PR is a bigger decision than the bug, and it belongs to the operator.
3. **It may be moot.** #1857 may delete the function outright; the scoping report already lists it as a straight API deletion.

So the residual library-level hole — a Rust caller of `ferro_hgvs::cli::parse_shuffle_direction("banana")` still gets `ThreePrime` — is **disclosed, not fixed**. The function's doc comment now says so, points at `python_helpers::parse_direction` as the validating form, and records that the CLI no longer depends on the fallback. If you would rather take the breaking change now, say so and I will fold it in.

## Tests

`tests/it/issue_1863_cli_direction_validation.rs` drives the **binary**, not the library — the defect lives in the clap declaration, so a library-level test calling `parse_shuffle_direction` directly would pass against the bug and keep passing after the fix. No reference is needed; argument validation precedes provider construction.

- `an_unrecognized_direction_is_rejected` — `banana`, `5prim`, `3prine`, `five`, `threeprime` and `""` all exit non-zero.
- `the_rejection_names_the_accepted_values` — the message contains `3prime` and `5prime`.
- `every_documented_spelling_is_still_accepted` — all ten currently-working spellings.
- `an_unrecognized_direction_does_not_masquerade_as_three_prime` — asserts the *difference*, since the defect is precisely that garbage produced output byte-identical to `3prime`. An exit-code check alone would keep passing if garbage later failed for some unrelated reason while still normalizing.

The valid spellings were enumerated **from the match arms**, not from the flag's doc comment ("3prime or 5prime") — guessing from the docs would have silently dropped `3`, `3'`, and every mixed-case form, since those survive only via the catch-all being removed.

## Verification

The tests were watched red before the fix, on the unfixed tree at `64fc76df`:

```
Summary [2.246s] 4 tests run: 1 passed, 3 failed, 10783 skipped

  `--direction "banana"` exited 0 — this is the #1863 bug: the flag has no value_parser
  and `parse_shuffle_direction` falls back to ThreePrime, so a bogus direction silently
  3'-shifts. stdout: "NM_000088.3:c.16del -> NM_000088.3:c.19del\n"
```

The one that passed is `every_documented_spelling_is_still_accepted` — the control arm, which must be green on both sides, and is what shows the other three are not trivially red.

After the fix:

```
$ ferro normalize --direction banana NM_000088.3:c.16del
error: invalid value 'banana' for '--direction <DIRECTION>'
  [possible values: 3prime, 3', 3, 5prime, 5', 5]
rc=2
```

and all ten spellings still resolve, with 3' and 5' still producing genuinely different output (`c.19del` vs `c.16del`).

Full local gate, each exit status read back out of its own log rather than from the runner's notification:

| check | result |
|---|---|
| `cargo nextest run --features dev --lib --test it --no-fail-fast` | `SUITE_EXIT=0` — 10444 passed, 0 failed |
| `cargo test --features dev --doc parse_shuffle_direction` | `DOCTEST_EXIT=0` — 1 passed |
| `cargo fmt --all --check` | `FMT_EXIT=0` |

Closes #1863

Representation-Change: none. Input validation only — `src/bin/ferro.rs` and `src/cli/parse.rs` docs; nothing under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched, and every valid direction produces the same output it did before. The only behaviour change is that a previously silently-accepted invalid value is now rejected, which moves no normalized row.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `Default::default()` was the only reachable constructor — the two documented knobs could not be
   set by any downstream caller.
 
+### Fixed
+
+- **`ferro normalize --direction` now rejects an unrecognized value instead of silently 3'-shifting.**
+  The flag carried no `value_parser`, and the function behind it ends in a catch-all, so
+  `--direction banana` ran a **3'** normalization and exited **0** — producing output
+  byte-identical to `--direction 3prime`. Since direction changes the answer, a typo returned a
+  plausible result to a question the user did not ask. It is now a clap usage error naming the
+  accepted values.
+
+  This was the CLI half of the defect fixed for the Python bindings in 0.8.0; the CLI has its own
+  independent copy of the parsing logic, which was left untouched at the time. Both surfaces now
+  accept exactly the same set — `3prime`/`3'`/`3` and `5prime`/`5'`/`5`, case-insensitively — and
+  every previously-valid spelling still works.
+
+  `ferro_hgvs::cli::parse_shuffle_direction` is unchanged: it is public, doctested API, so
+  tightening its signature would be a breaking change. A direct Rust caller still receives
+  `ThreePrime` for an unrecognized string; `ferro_hgvs::python_helpers::parse_direction` is the
+  validating form.
+
 ## [0.13.1](https://github.com/fulcrumgenomics/ferro-hgvs/compare/v0.13.0...v0.13.1) - 2026-08-08
 
 ### Representation changes
--- a/src/bin/ferro.rs
+++ b/src/bin/ferro.rs
@@ -283,8 +283,21 @@ UNSTABLE EVALUATION SWITCH: FERRO_PARTITION
         #[arg(short = 'f', long, default_value = "text", value_parser = ["text", "json", "tsv"])]
         format: String,
 
-        /// Shuffle direction (3prime or 5prime)
-        #[arg(long, default_value = "3prime")]
+        /// Shuffle direction. `3prime`/`3'`/`3` or `5prime`/`5'`/`5`,
+        /// case-insensitively — the same set the Python bindings accept.
+        ///
+        /// The values are listed here rather than left to
+        /// `parse_shuffle_direction` because that function ends in a catch-all
+        /// (`_ => ThreePrime`): without a `value_parser`, `--direction banana`
+        /// silently 3'-shifted and exited 0 (#1863, the CLI half of #1016).
+        /// `ignore_case` is load-bearing — the function lowercases its input,
+        /// so `--direction 5PRIME` has always worked and must keep working.
+        #[arg(
+            long,
+            default_value = "3prime",
+            ignore_case = true,
+            value_parser = ["3prime", "3'", "3", "5prime", "5'", "5"]
+        )]
         direction: String,
 
         /// Reference directory (with manifest.json from 'ferro prepare')
--- a/src/cli/parse.rs
+++ b/src/cli/parse.rs
@@ -37,6 +37,25 @@ pub fn parse_genome_build(build: &str) -> GenomeBuild {
 /// - 5prime, 5', 5 -> FivePrime
 /// - Other values default to ThreePrime
 ///
+/// # The lenient fallback is legacy, and is no longer how the CLI validates
+///
+/// The final arm accepts anything, which is why `ferro normalize --direction
+/// banana` used to 3'-shift and exit 0 (#1863, the CLI half of #1016). The flag
+/// now carries its own `value_parser` listing the six aliases, so **the
+/// fallback is unreachable from the CLI** and an unrecognized value is a clap
+/// usage error naming the accepted values.
+///
+/// The fallback is retained here, rather than made fallible, only because this
+/// function is public API — re-exported from [`crate::cli`] and doctested — so
+/// tightening its signature is a breaking change and a decision separate from
+/// the defect. It is also entangled with #1857, which may remove the 5'
+/// direction (and this function) outright.
+///
+/// **A direct Rust caller therefore still gets `ThreePrime` for an
+/// unrecognized string.** Prefer [`crate::python_helpers::parse_direction`],
+/// which returns `Option<ShuffleDirection>` and is what the Python bindings
+/// validate with; it accepts exactly the same six aliases, case-insensitively.
+///
 /// # Examples
 ///
 /// ```
@@ -45,11 +64,19 @@ pub fn parse_genome_build(build: &str) -> GenomeBuild {
 ///
 /// assert!(matches!(parse_shuffle_direction("3prime"), ShuffleDirection::ThreePrime));
 /// assert!(matches!(parse_shuffle_direction("5'"), ShuffleDirection::FivePrime));
+///
+/// // The legacy fallback, documented above. The CLI no longer reaches it —
+/// // `--direction unknown` is rejected by clap before this function is called.
 /// assert!(matches!(parse_shuffle_direction("unknown"), ShuffleDirection::ThreePrime));
 /// ```
 pub fn parse_shuffle_direction(direction: &str) -> ShuffleDirection {
     match direction.to_lowercase().as_str() {
         "5prime" | "5'" | "5" => ShuffleDirection::FivePrime,
+        // The 3' aliases are spelled out rather than left to the catch-all, so
+        // the mapping the CLI's `value_parser` promises is stated here too. The
+        // catch-all is the legacy fallback described above; it is not the route
+        // by which a valid 3' spelling resolves.
+        "3prime" | "3'" | "3" => ShuffleDirection::ThreePrime,
         _ => ShuffleDirection::ThreePrime,
     }
 }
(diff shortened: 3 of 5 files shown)


nh13:
@coderabbitai pause

nh13:
Superseded by #1879, which removed the surface this PR validates rather than validating it.

On `main`, `ferro normalize --direction` no longer exists and `parse_shuffle_direction` has no definition anywhere in `src/` — the only remaining mentions are comments recording the removal. `src/bin/ferro.rs` states the reasoning and names this PR's issue directly:

> There is deliberately no `--direction` flag. […] Removing the flag rather than accepting-and-ignoring it is the point: clap rejects an unknown argument with a non-zero exit, so a script that still passes `--direction 5prime` fails loudly instead of silently 3'-shifting (the #1863 failure shape). Pinned by `tests/it/five_prime_public_surface_removed.rs`.

So #1863 is already closed on `main`, more strongly than a `value_parser` fix closes it: there is no value left to mis-parse.

The two branches also assert opposite contracts, which is why this cannot be rebased mechanically — `main`'s `the_direction_flag_is_rejected_rather_than_ignored` requires every `--direction <value>` to fail, while this PR's `every_documented_spelling_is_still_accepted` requires those same values to be accepted. Resolving that is a decision about whether the flag returns, not a merge conflict.

Closing as superseded. Reopen if the flag should come back — in which case this PR's validation is the right shape for it.

PR #1867, opened 2026-08-13
test: give the five real-data normalization guards assertions that can fail

### LINKED ISSUE #1858: test: four of the five real_data_normalization_tests guards check nothing, even fully armed

nh13:
Was stacked on #1861, which has since squash-merged as `471c1217`. Rebased with `git rebase --onto origin/main <old-parent>`, so this PR is now a single commit against `main` and its diff is one file, `tests/it/real_data_normalization_tests.rs`.

Closes #1858.

## What was wrong

Three of the five guards in `tests/it/real_data_normalization_tests.rs` carried no assertion at all, and a fourth kept both of its assertions inside a bounds check. Wiring the module into the nightly (#1851) guaranteed it *ran*; it never guaranteed it *checked*. This is the third shape of that defect family and the one no instrument used so far can see — an assertionless guard does the same work and costs the same wall clock as an armed one.

## What each guard now asserts

The load is carried by two properties that need no adjudication, so nothing here pins a value nobody has justified.

**Sequence preservation.** Which of two legal spellings ferro ships is a rule 6 choice (`README.md`); whether the shipped spelling still denotes the input's bases is rule 1, and a failure there is a bug under every reading. `assert_denotes_the_same_bases` carries the guards whose spelling question is open. A `NotComparable` verdict is a panic, not a pass — folding skips in with agreements would rebuild the hole inside the helper meant to close it.

**A required CDS origin.** Every guard read `transcript.cds_start.unwrap_or(1)`. That substitutes an origin when the reference serves none, and the substituted coordinates print exactly like real ones. The fallback is removed for the shape of the failure it hides, not because one was observed — see below, where the claim that one *was* observed is withdrawn. An unserved CDS now routes through `manifest_gate::unserved`.

## The headline finding is withdrawn — it does not reproduce

An earlier revision of this description, and of the module's own doc comments, stated as measured fact that `MultiFastaProvider::get_transcript("NM_033517.1")` returns 7113 cached bases with `cds_start: None`, `cds_end: None` and zero exons — bare FASTA, no cdot record — and that `NM_033517.1:c.1324del` therefore came back unchanged as a passthrough rather than a 3'-rule decision.

Re-measured 2026-08-14, identically on both prepared volumes, all four accessions this module names are fully annotated:

```
NM_033517.1/.2/.3  cds_start=Some(1)   cds_end=Some(5157)  exons=22  len=7113
NM_001408491.1     cds_start=Some(296) cds_end=Some(2365)  exons=23  len=3748
NM_001282424.3     cds_start=Some(169) cds_end=Some(3060)  exons=25  len=3791
NM_001394148.2     cds_start=Some(137) cds_end=Some(328)   exons=3   len=570
```

Only the length (7113) survives from the withdrawn reading. Three consequences, each a separate correction:

- **`unwrap_or(1)` fabricated nothing here.** cdot's record for this accession carries `start_codon: 0`, so `cds_start` really is 1 and the fallback happened to agree with the reference. Removing it is still right — it removes a *possible* substitution — but there is no sighting behind it.
- **`c.1324del` is not a passthrough.** It normalizes to **`c.1326del`**, which is the Mutalyzer answer, and the reference says why: `c.1324`/`c.1325`/`c.1326` are all `C` and `c.1327` is `G`, so the deletion shifts to the 3' end of a `CCC` run and stops. The four further inputs the old text called unshifted all shift too: `c.125del` → `c.127del`, `c.160del` → `c.162del`, `c.164del` → `c.167del`, `c.517del` → `c.518del`. So there is no ferro-versus-Mutalyzer disagreement on this accession, and the guard's original question is answered rather than blocked.
- **There is no open question for an operator.** "Should `ferro prepare` serve `NM_033517` with a cdot record?" — it does. That paragraph is deleted rather than rephrased. (The `#1870` sentence in the guard's doc — "with no CDS the normalizer … returns the description unchanged" — is deleted with it: `rulings[c-description-against-an-unresolvable-cds-is-refused]` makes that a refusal, and the premise does not obtain here anyway.)

One figure above is **cdot-version-dependent, and nothing in this PR keys on it.** Read out of both cdot builds directly:

| | `data_v0.2.32` (what the manifest pins) | `data_v0.2.34` |
|---|---|---|
| record present | yes | yes |
| `exons` | 22 | 22 |
| `start_codon` → `cds_start` | `0` → **1** | `0` → **1** |
| `stop_codon` → `cds_end` | `5157` | `5196` |

`0.2.32` collapses this accession's 39-base alignment gap (transcript 1302..1342, between exons 10 and 11) into the stop codon; `0.2.34` does not, and 5196 is what RefSeq/GenBank annotate (`CDS 1..5196`). That is the open sibling of `rulings[c-and-n-positions-are-flat-transcript-offsets]` and is not this module's business. `cds_start` is the only CDS figure any guard here reads, and it is 1 under both — so the correction above does not depend on which cdot the reference was prepared from, and the corrected doc comments say so rather than hard-coding 5157 as a fact about the accession.

**The cdot-version question is therefore settled in one direction and left open in the other.** Settled: no cdot version in play produces the withdrawn reading. Both builds carry the record with 22 exons and a start codon, so an older or newer cdot moves `cds_end`, it does not delete the record — the claim could not have been true against a *version* difference. Not settled: what state actually produced the original reading. It does not hold on either blessed volume today, and `main`'s own ruling `c-and-n-positions-are-flat-transcript-offsets`, decided 2026-08-12 — the day *before* the original measurement — already records "the live provider serves `NM_033517.1` `cds_end = 5157`", i.e. an annotated CDS on this same reference. That is strong evidence of an error rather than a genuine earlier state, and it is short of proof, so no history is invented for it.

**The promotion is dormant, not firing.** With the CDS served, `cds_start_or_skip` never takes its `None` branch on this reference, so no guard stands down and none fails under `FERRO_REQUIRE_MANIFEST`. The gate is a safety net whose value is the shape of failure it removes — a stand-down is loud, a substituted origin is silent. It was measured live rather than assumed: see the Verification section.
## Two premises withdrawn rather than propped up

- **"delins should NOT be 3' shifted like del/dup"** grades a description by its edit type — a property of the input's spelling, which rule 3 and `canonical-form-choice-when-both-legal` both exclude. What the guard was really written to catch is real: `c.2140_2141delinsTATGCA` denotes different bases (`GTATGCA` against the input's `TATGCAT`), so sequence preservation is exactly the right test. Ferro in fact emits neither form but `c.[2138_2139insTAT;2140_2141insA]`, a sequence-preserving split — and the old body printed "POTENTIAL BUG" whenever the output contained `2140_2141`, which that correct split matches. The detector had gone wrong in both directions at once.
- **"Does the 3' rule reach the 5'UTR?"** is a real question that this input cannot answer. `c.-56` is `T` and `c.-46` is `G`, so no sequence-preserving rotation of the duplicated unit exists and both readings predict the same output. The non-discrimination is checkable, so it is now checked, with a message saying to re-derive rather than update if it ever fires. The guard is documented as what it actually is: the #97 5'UTR coordinate-mapping regression pin, cross-checking `normalize_tests::clinvar_normalization`'s hermetic pin against the real reference.

## The bounds check

Never false on the blessed reference, and measurably so: `NM_001408491.1` is 3748 bases with `cds_start` 296, putting `c.517`/`c.518` at 0-based 811 and 812. So promoting the `if` to an `assert!` fixes no live failure — it removes the one way the guard could go quiet, which is the shape that cannot be seen from a run's duration or its pass/fail.

## Verification
Every assertion was watched fail before being kept — sabotaged, run red, restored from a pre-edit `cp` (`md5` byte-identical afterwards, `git status` clean).

| assertion | proven red by |
|---|---|
| guard 1 — `seq[idx_517] == seq[idx_518]` | comparing against `seq[idx_518 + 2]` |
| guard 1 — `output.contains("c.518del")` | expecting `c.519del` |
| guard 1 — `assert_denotes_the_same_bases` | comparing the input against `c.517_520del` |
| guard 2 — `assert_denotes_the_same_bases` (its only one) | comparing the input against `c.1324_1330del` → `normalization changed the sequence the description denotes` |
| guard 3 — `assert_denotes_the_same_bases` (its only one) | comparing the input against `c.2140_2141delinsTATGCA` → `Differ`, `from input TATGCAT` / `from output GTATGCA` |
| guard 4 — `assert_ne!(seq[idx_first], seq[idx_last + 1])` | comparing the base against itself |
| guard 4 — the `c.-56_-47dup` output pin | expecting `c.-55_-46dup` |
| guard 5 — bounds `idx_517 < seq.len() && idx_518 < seq.len()` | `idx_517 + seq.len() < seq.len()` |
| guard 5 — `base_517 == 'A'` | expecting `'T'` |

Armed against the prepared reference (`FERRO_MANIFEST` set, read-only; nothing under `data/ferro` modified):

- **without** `FERRO_REQUIRE_MANIFEST`: `5 tests run: 5 passed`, `EXIT=0`, and **no `skipping —` line anywhere** in the run.
- **with** `FERRO_REQUIRE_MANIFEST=1`: identical — `5 tests run: 5 passed`, `EXIT=0`. Guard 2 does **not** stand down and does **not** fail. An earlier revision of this table claimed it fired here; that row is withdrawn.
- **`cds_start_or_skip` forced to `None`, `FERRO_REQUIRE_MANIFEST=1`**: `5 tests run: 0 passed, 5 failed`, `EXIT=100`, each naming its own accession — `could not get \`NM_033517.1 CDS annotation\` from it: … a \`c.\` position does not resolve`. That is what proves the gate is armed rather than inert, since the shipped run cannot demonstrate it.

None of the above is inferred from timing: `CLAUDE.md` records that a guard which stands down *after* loading the reference costs the same as an armed one (12.086 s vs 12.027 s), so every row is a watched failure.

`cargo fmt --all --check`, `cargo clippy --profile soak --all-features --all-targets -- -D warnings` and the full `cargo nextest run --cargo-profile soak --features dev --no-fail-fast` gate all clean.

Representation-Change: none. Tests and doc comments only; no file under src/normalize/, src/hgvs/, src/spdi/, src/project/, src/reference/ or src/error_handling/ is touched — the diff is one file, tests/it/real_data_normalization_tests.rs.

### DIFF
--- a/tests/it/real_data_normalization_tests.rs
+++ b/tests/it/real_data_normalization_tests.rs
@@ -6,15 +6,72 @@
 //! `FERRO_REQUIRE_MANIFEST` standing down is a failure, whether the manifest is
 //! absent or merely does not carry the transcript a guard names.
 //!
-//! **That promotion buys execution, not verification.** Three of the five
-//! guards below (`test_potential_bug_deletion_shift_nm033517`,
-//! `test_potential_bug_delins_shift`, `test_5utr_duplication_shifting`) contain
-//! no assertion at all — they are investigation scripts that print — and
-//! `test_compare_with_mutalyzer_sequence` keeps both of its assertions inside a
-//! bounds check, so an out-of-range coordinate leaves it asserting nothing.
-//! Tracked as #1858; do not read a green run here as coverage of those four.
+//! **That promotion bought execution, not verification, and #1858 is the
+//! difference.** Three of these guards used to carry no assertion at all — they
+//! were investigation scripts that printed a finding and returned successfully,
+//! so the condition each existed to detect was reported by *passing* — and
+//! `test_compare_with_mutalyzer_sequence` kept both of its assertions inside a
+//! bounds check. This module now asserts, and what each guard asserts is chosen
+//! by what the reference can actually decide.
+//!
+//! # The one property that needs no adjudication
+//!
+//! Which of two legal spellings ferro ships is a rule 6 choice (`README.md`'s
+//! ruleset) and needs an operator to settle. Whether the shipped spelling still
+//! denotes the input's bases is rule 1, and a failure there is a bug under
+//! every reading of the recommendations. So [`assert_denotes_the_same_bases`]
+//! carries the load wherever the spelling question is open, and no guard here
+//! pins a string that nobody has justified.
+//!
+//! # `unwrap_or(1)` is removed because it *can* fabricate, not because it did
+//!
+//! Every guard used to read its CDS origin as `transcript.cds_start
+//! .unwrap_or(1)`. That substitutes an origin when the reference serves none,
+//! and the substituted coordinates print exactly like real ones — a guard
+//! reporting `c.1324` would be naming a base the description does not name,
+//! with nothing in its output saying so. The fallback is gone; an unserved CDS
+//! now routes through [`cds_start_or_skip`], which stands the guard down and
+//! records it.
+//!
+//! **On the blessed reference that branch is dead, and this is the correction
+//! of a claim this module previously shipped as measured fact.** It stated that
+//! `NM_033517` is served as bare sequence with no cdot record —
+//! `cds_start: None`, `cds_end: None`, zero exons. Re-measured 2026-08-14
+//! against both prepared volumes, all four accessions this module names come
+//! back annotated:
+//!
+//! ```text
+//! NM_033517.1/.2/.3  cds_start=Some(1)   cds_end=Some(5157)  exons=22  len=7113
+//! NM_001408491.1     cds_start=Some(296) cds_end=Some(2365)  exons=23  len=3748
+//! NM_001282424.3     cds_start=Some(169) cds_end=Some(3060)  exons=25  len=3791
+//! NM_001394148.2     cds_start=Some(137) cds_end=Some(328)   exons=3   len=570
+//! ```
+//!
+//! `NM_033517`'s cdot record carries `start_codon: 0`, so `cds_start` is 1 and
+//! a `c.N` on it resolves to transcript offset `N` — the old fallback happened
+//! to agree with the reference here rather than fabricate against it. Only the
+//! length (7113) survives from the withdrawn reading. Removing the fallback is
+//! still right, and [`cds_start_or_skip`] is still the right shape; what is
+//! withdrawn is the claim that anything was observed going through it.
+//!
+//! Note `cds_end` is the one figure above that is **cdot-version-dependent**,
+//! and nothing here keys on it. Read out of both cdot builds directly:
+//! `data_v0.2.32` — the version the prepared manifest pins — collapses this
+//! accession's 39-base alignment gap (transcript 1302..1342, between exons 10
+//! and 11) into the stop codon and records `stop_codon: 5157`, while
+//! `data_v0.2.34` records `stop_codon: 5196`, which is what RefSeq/GenBank
+//! annotate (`CDS 1..5196`). That discrepancy is the open sibling of
+//! `rulings[c-and-n-positions-are-flat-transcript-offsets]` and is not this
+//! module's business.
+//!
+//! **Both versions carry the record**, with 22 exons and `start_codon: 0`, so
+//! `cds_start` is 1 either way and no cdot in play yields the record-less state
+//! the withdrawn reading described. The correction above therefore does not
+//! depend on which cdot the reference was prepared from.
 
+use ferro_hgvs::hgvs::HgvsVariant;
 use ferro_hgvs::reference::transcript::Transcript;
+use ferro_hgvs::spdi::{compare_denoted_sequences, DenotedSequenceComparison};
 use ferro_hgvs::{parse_hgvs, MultiFastaProvider, Normalizer, ReferenceProvider};
 use std::path::Path;
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1868, opened 2026-08-13
ci: arm the denoted-sequence oracle in test-oracle behind a two-row debt list

### LINKED ISSUE #1815: normalize: the NG_ oracle-fire class is invisible to gating CI (the seven rows themselves were an SPDI frame defect, see #1820)

nh13:
Refs #1815. **Deliberately not `Closes`** — see [What this does NOT fix](#what-this-does-not-fix). The issue has two halves and this closes one; the previous revision of this PR carried `Closes #1815` and its own body called that "optimistic", which was right.

Representation-Change: none. CI wiring, workflow comments, the local oracle runner, `CLAUDE.md` and test-only guards. `scripts/check_representation_change.py` confirms it: *"No file under a watched directory changed; no declaration needed."* No normalized output moves — the point of the change is that an existing oracle now *runs* in a gating job. It asserts; it does not normalize.

## What this does

`FERRO_ASSERT_SEQUENCE` (#1615) is the only oracle that can see a normalized description denoting different bases than its input, or denoting none at all. It ran in exactly two places: `sweeps`, over three cis-sweep modules, and `nightly-mutalyzer.yml`, which is `continue-on-error` by design. `test-oracle` — the job that runs the suite armed — was the one place the set of four was incomplete. That is #1815's surviving half.

This arms it there, behind a two-row `SEQUENCE_ORACLE_EXCLUDE`, and gates through the required `Test` rollup (`test-required` `needs: [test, test-oracle, soak, sweeps]`).

## This revision was re-derived, not rebased — and the debt list is not the one the PR was drafted against

The previous head (`a26f0be4`, base `f1f53a8e`, ~60 commits behind) is **replaced** rather than rebased, because its central factual claim had become wrong in both directions and its `ci.yml`/`CLAUDE.md` hunks *deleted* newer, more accurate records to restore their older selves. Porting it would have regressed the file.

**The selection-wide armed figure has read 5 → 2 → 3 across three measurements in three days:**

| measured | tree | armed failures | why it moved |
|---|---|---:|---|
| #1813 | `674e9c8b` | 5 | 3 × `issue_1487_canonical_window_overflow` (#1690) + 2 × `stranded_identity_member` (#1655) |
| #1990 | `c9207d7e` | 2 | #1690 closed; the three overflow rows are gone |
| **this PR** | `1aecc93a` | **3** | #2051 added a gate that fires — see below |

So the "two-row debt list" in the title is *coincidentally* still two rows, and they are **not the same two**. `issue_1487_canonical_window_overflow` is out (#1690 closed, 2026-08-15) and `dump_normalized_corpus`'s render-time gate is in (#2051, 2026-08-16). Both times the stale reading was the *flattering* one. Every count in the new comments names the tree it was taken on and says to re-measure.

### The debt list, as measured

```
without SEQUENCE_ORACLE_EXCLUDE:  11202 tests run: 11199 passed, 3 failed   EXIT=100
with    SEQUENCE_ORACLE_EXCLUDE:  11200 tests run: 11200 passed, 0 failed   EXIT=0
```

- **`stranded_identity_member`** — 2 of its 4 tests. A genuine fire of #1281's *denotes no sequence* shape, on a module whose purpose is to **pin** that defect (**#1655**, open). A test that pins a defect and an oracle that aborts on it cannot both run: exactly the class `ORACLE_EXCLUDE` already documents. Named at module granularity because the module's whole purpose is the pin.
- **`the_render_time_reference_matches_what_the_pipeline_was_given`** — one **test**, named rather than its module, so the other 55 tests in `examples/dump_normalized_corpus.rs` stay armed. #2051 added it hours before this measurement; it re-normalizes each corpus row **itself, outside** the `catch_unwind` that `dump`'s own pass uses, so a row that trips an oracle takes the test with it. It passes unarmed in the same tree (`EXIT=0`, measured), so it is an oracle fire and not a pre-existing red.

## Two defect classes this surfaced, both filed

47 corpus inputs abort that gate. Eight are **#1983** (already open). The other 39 were unreported:

- **#2140** (38 inputs) — a minted repeat tract widened across the CDS start into the 5'UTR swallows its sibling, so the output denotes no sequence: `NM_TESTX.1:c.[2_3dup;5T>A]` → `c.[-3_6T[11];5T>A]`, where the tract spans `-3..6` and the sibling names base `5`. The same input is **fine** on the single-exon transcript (`c.6_7insAT`) and wrong on the 5'UTR-bearing one, across 5 corpus families / 1,720 `cx` rows. Not #1983: that issue is that `demote_repeats_spanning_siblings` has no `NaEdit::Repeat` arm and so *"covers only the repeats ferro itself mints"* — these **are** the repeats ferro mints, so the pass is reached and still fails.
- **#2139** (1 input) — #1513's adjacent-junction insertion pair still denotes different bases when it straddles the CDS start: `NM_TESTX.1:c.[-1_1insAA;1_2insTT]` inserts `ATTA` at interbase 5, every output inserts `TTAA` at interbase 4. #1513's guards pin the shape only on pairs sitting **entirely inside the CDS**.

**Both are confirmed without the oracle.** `cargo run --features dev --example dump_normalized_corpus -- --verify-spdi` reports `13 denote different bases` of 96,182 verified rows: 6 × #2139, 5 × #1983, 2 × the repeat family. **#2140 is invisible to that instrument** — it files those rows under *"unverifiable (member overlap)"*, 28,368 of 96,182, and grepping its report for `2_3dup;5T>A` returns 0. The denoted-sequence oracle's *denotes-no-sequence* arm is the only thing that reaches them, which is the design decision that makes #1615 worth arming.

## Nothing is traded away

A nextest `-E` is one expression, so negating the filter would also withdraw these rows from `FERRO_ASSERT_IDEMPOTENT`, `_REPARSE` and `_IN_BOUNDS`, which they pass — three oracles surrendered to gain one, the swap the `sweeps` job's comment refuses to make. So `test-oracle` gains a **second, un-partitioned step** that re-runs exactly those rows under those three. This file's oracle coverage is a strict superset of what it was.

`oracle_exclude_invariant.rs` asserts that step exists, selects the variable **by reference** rather than by a copy of its value, arms exactly the other three, does **not** arm the fourth, and passes `--no-tests=fail`.

## Proof the gate can go red

A gate that runs but cannot fail is worse than no gate. Both halves measured on `04a1146e`, the base this PR is rebased onto:

| | armed suite | guard suite |
|---|---|---|
| **as wired** | `11214 passed, 0 failed` — **EXIT=0** | 31/31 pass |
| **one debt row removed** | `11216 passed, 2 failed` — **EXIT=100** | **31/31 still pass** |

Red on the real message, not on a wiring error:

```
FERRO_ASSERT_SEQUENCE: normalization produced a description that denotes no sequence at all,
                       from an input that denotes one
  input:  TEMPLATE:g.[10_11insAT;11_12inv;12_13insA]
  output: TEMPLATE:g.[11_12insTA;11_12=;12_13insA]
```

The right-hand column is the load-bearing part: **the guards cannot see an oracle fire.** They stay green while the armed run is `EXIT=100`, so the arming is doing work no static check substitutes for. `ci.yml` was restored byte-identically after (`cmp` clean, `git status` empty), and zero `Skipping:` lines in every run with `FERRO_REQUIRE_BULK_FIXTURES=1`.

### Sabotage table — each wiring guard, checked rather than assumed

| sabotage | guard that caught it | exit |
|---|---|---|
| delete the compensating step's `-E` | `the_sequence_oracle_exclusions_still_run_under_the_other_three_oracles` | 100 |
| add `FERRO_ASSERT_SEQUENCE` to the compensating step | same | 100 |
| typo a debt row name | `every_row_withheld_from_the_sequence_oracle_names_something_that_exists` | 100 |
| drop `and not ($SEQUENCE_ORACLE_EXCLUDE)` from the armed `-E` | `the_ci_oracle_selection_negates_…` | 100 |
| retire the variable, keep the `-E` term | `run_oracle_suite.sh` refuses, naming the repair | 1 |
| un-arm `FERRO_ASSERT_SEQUENCE`, keep the filter | `the_sequence_oracle_exclusions_still_run_under_…` | 100 |
| *(control — no sabotage)* | — | **0** |

## Review notes on my own diff

- **`test-oracle` now has two nextest steps, which broke a guard on a correct file** and is the non-obvious consequence to know about. `test_oracle_job_flags()` scanned the whole job, so it returned seven keys with three duplicated against the runner's four. It is now scoped to the armed step, identified by its **`--partition`** — deliberately *not* by the step's `name:`, which is what the script's awk anchors on, so a rename breaks one derivation and not the other.
- **`if: matrix.shard == 1` on the compensating step** is the shape `.coderabbit.yaml`'s `.github/workflows/**` instruction says to flag, so it is called out rather than left to be found. The step selects 5 tests and is not partitioned: `--partition hash:K/4` would leave some shards selecting none, and neither answer is acceptable (`--no-tests=fail` reddens a shard for doing its job; `--no-tests=pass` is a step that goes green having run nothing). It carries no `if: always()`, so a red armed step skips it — the concern that rule protects against is coverage lost on a *passing* run, and on a passing run this step always executes.
- **A claim I made and then had to retract**, recorded because the diff would otherwise read as more certain than it was. I wrote in `ci.yml` that one `stranded_identity_member` test had been renamed. It had not — my evidence was a `cargo nextest list` I had truncated with `tail -3`. The red run printed the real names and the comment is corrected to the measured pair, with a note that they were read off a failing run rather than matched against prose. I also asserted a substitution-order hazard between `$ORACLE_EXCLUDE` and `$SEQUENCE_ORACLE_EXCLUDE`; **tested, it does not exist** (the pattern carries the leading `$`), and both the script and the Rust say so rather than carrying a plausible-sounding false reason.
- **Stale claims corrected as part of the change**, each of which this PR would otherwise falsify: `test-oracle`'s own comment block; the `SWEEP_FILTER` note and the `sweeps` job's "only job IN THIS FILE that sets it"; `censuses`' "the three `test-oracle` sets"; `run_oracle_suite.sh`'s header ("deliberately does NOT arm this one"); `CLAUDE.md` in six places, including *"No CI job arms the oracle where it fires"*, which is now false for `normalize_tests` — its 453 tests run armed and green. The useful prose from the replaced commit's three test-file hunks is carried forward, adapted (the `issue_1487` one had to be rewritten, since that module is no longer on the debt list).
- **`censuses` is now where the plausible wrong edit lives.** Its armed step used to mirror `test-oracle`'s flag set; it is now a strict **subset**, deliberately, and arming the fourth there has never been measured over `ORACLE_ONLY_FILTER`. That job's header says so, and `the_local_oracle_runner_arms_exactly_the_flags_test_oracle_arms`'s doc points at it.

## What this does NOT fix

**No pull-request job provisions a manifest, so the reference-aware axes still skip-as-PASS.** `test-oracle` arms the flag but sets no `FERRO_MANIFEST`, so `mutalyzer_normalize_tests` / `biocommons_normalize_tests` early-return there exactly as in the plain `test` job, and nextest reports that skip path as **PASS** rather than SKIP. A denoted-sequence violation on those axes still cannot redden a required check.

That is not an oversight and not a small change: `nightly-mutalyzer.yml` spends ~30 min in `ferro prepare` per run, and the prepared reference is deliberately uncached — a 3.25 GiB entry against a 10 GiB repository budget, measured at a 0 % hit rate over five consecutive nights (#1218). What this PR does do is stop the skip being quiet: the `test` job's `::notice::` now says in those words that the skip reports as PASS and that no gating job runs those axes.

**So #1815 should stay open on the manifest half after this merges**, which is why the trailer is `Refs` and not `Closes`.

## Deferred to a maintainer, deliberately not decided here

**Should `nightly-mutalyzer.yml` keep `continue-on-error: true`?** It is the only lever that makes a manifest-backed oracle fire block anything. It is also deliberate — its purpose is to surface drift in the `ferro-xfail-*` artifact rather than to gate, the corpus runner wraps normalization in `catch_panics` on that assumption, and `report-failure`'s `hint:` is written around it. Removing it converts an informational nightly into a blocking one. Related and already filed: **#1998**, that nothing baselines that workflow's xfail set, so a new failure is indistinguishable from the 37 beside it.

**Retiring the debt list is a four-file change** (`ci.yml`, `scripts/run_oracle_suite.sh`, `tests/it/oracle_exclude_invariant.rs`, plus the module doc that cites the row). A half-done retirement fails loudly rather than handing nextest the literal string `$SEQUENCE_ORACLE_EXCLUDE`, which would parse as a test-name substring and quietly narrow the armed job to nothing. Closing **#1655** deletes one row; closing **#2140**, **#1983** and **#2139** together delete the other, and the variable with it.

## Local verification

Each command run unpiped with `EXIT=$?` captured to a log and read back:

| gate | result |
|---|---|
| `cargo nextest run --features dev --no-fail-fast` | `11389 passed, 151 skipped` — **EXIT=0** |
| `cargo clippy --all-features --all-targets -- -D warnings` | **EXIT=0** |
| `cargo fmt --all --check` | **EXIT=0** |
| `python3 -m pytest tests/python/ -q` | `1137 passed, 8 skipped` — **EXIT=0** |
| `cargo nextest run --features dev --test it -E 'test(workflow_pipefail_gate)'` | `18 passed` — **EXIT=0** |

Guard modules post-rebase, including `ruling_records_are_intact` and `normalization_contract_doc`: 98/98. `scripts/run_oracle_suite.sh` green over 11,214 tests armed with all four oracles.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,11 +73,11 @@ env:
   # NOTE also: `sweeps` selects on this filter PLUS one module it deliberately
   # re-runs — `issue_1615_denoted_sequence_oracle`, the only test that reads the
   # denoted-sequence oracle's counters, which are meaningless unless
-  # `FERRO_ASSERT_SEQUENCE` is set and `sweeps` is the only job IN THIS FILE that
-  # sets it (`nightly-mutalyzer.yml` sets it too, non-gating). So
-  # "nothing is run twice" holds of this filter and is one module short of
-  # holding of that job's `-E`; the reason it is appended there rather than added
-  # here is on the job itself.
+  # `FERRO_ASSERT_SEQUENCE` is set. `sweeps` was the only job IN THIS FILE that
+  # set it until #1815 armed it in `test-oracle` too (`nightly-mutalyzer.yml`
+  # sets it as well, non-gating). So "nothing is run twice" holds of this filter
+  # and is one module short of holding of that job's `-E`; the reason it is
+  # appended there rather than added here is on the job itself.
   SWEEP_FILTER: >-
     test(cis_junction_crossing_shift) + test(repeat_span_sibling_overlap)
     + test(issue_1254_sibling_crossing_shift)
@@ -275,6 +275,53 @@ env:
     + test(issue_1715_rna_alignment_symbol_reach)
     + test(conformance_census_instrument)
     + test(conformance_census_runs)
+  # Withheld from the DENOTED-SEQUENCE oracle and from nothing else.
+  #
+  # `test-oracle` arms `FERRO_ASSERT_SEQUENCE` as of #1815. These are the rows
+  # that stopped it being armed before, each named with the OPEN issue that
+  # retires it. Re-measured on `origin/main` @ 1aecc93a, that job's four flags
+  # over its own `-E` with no `--partition`, bulk fixtures present
+  # (`FERRO_REQUIRE_BULK_FIXTURES=1`, zero `Skipping:` lines):
+  #
+  #   without this filter: 11202 tests run: 11199 passed, 3 failed   EXIT=100
+  #   with    this filter: 11197 tests run: 11197 passed, 0 failed   EXIT=0
+  #
+  #   * `stranded_identity_member` (2 of its 4 tests) — a GENUINE fire, of
+  #     #1281's "denotes no sequence" shape, on a module whose purpose is to PIN
+  #     a defect (**#1655**, 132 of 132 non-confluent). That is exactly the class
+  #     `ORACLE_EXCLUDE` documents: a test that pins a defect and an oracle that
+  #     aborts on it cannot both run.
+  #   * `the_render_time_reference_matches_what_the_pipeline_was_given` (1 test,
+  #     in `examples/dump_normalized_corpus.rs`) — also a genuine fire, and the
+  #     module's other 55 tests stay armed because this names the TEST, not the
+  #     module. #2051 added this gate on 2026-08-16; it re-normalizes each corpus
+  #     row ITSELF, outside the `catch_unwind` `dump`'s own pass uses, so a row
+  #     that trips an oracle takes the test with it. 47 corpus inputs fire:
+  #     **#2140** (38 inputs — a minted repeat tract widened across the CDS start
+  #     into the 5'UTR swallows its sibling), **#1983** (8 — the authored-repeat
+  #     form of the same collision) and **#2139** (1 — an adjacent-junction
+  #     insertion pair straddling the CDS start).
+  #
+  # THIS LIST IS TEMPORARY AND MUST SHRINK, and its composition has already moved
+  # TWICE IN THREE DAYS — which is the reason to re-measure it rather than trust
+  # the figures above. It held `issue_1487_canonical_window_overflow` until #1990
+  # closed #1690 (2026-08-15), and gained the render-time gate when #2051 landed
+  # (2026-08-16). Both times the stale reading was the *flattering* one.
+  #
+  # NOTHING IS TRADED AWAY FOR IT. A nextest `-E` is one expression, so negating
+  # this filter in the armed step would also withdraw these rows from
+  # `FERRO_ASSERT_IDEMPOTENT`, `_REPARSE` and `_IN_BOUNDS`, which they pass —
+  # trading three oracles for one, the swap the `sweeps` job's comment refuses to
+  # make for `issue_1615_denoted_sequence_oracle`. The step after the armed one
+  # re-runs precisely these rows under those three, so this file's oracle
+  # coverage is a strict superset of what it was.
+  #
+  # Kept separate from `ORACLE_EXCLUDE` rather than added to it: that list is a
+  # permanent statement about instruments that destroy each other, this one is a
+  # debt with three issue numbers on it.
+  SEQUENCE_ORACLE_EXCLUDE: >-
+    test(stranded_identity_member)
+    + test(the_render_time_reference_matches_what_the_pipeline_was_given)
 
 # Every job carries a `timeout-minutes`. Without one a job inherits GitHub's
 # 360-minute default, so a wedged runner holds a PR for six hours: observed on
@@ -1487,9 +1534,19 @@ jobs:
         # *does not* provision a manifest, so requiring one here would fail
         # every PR. It is set in `nightly-mutalyzer.yml`, the only job that runs
         # `ferro prepare`. See `src/conformance/manifest_gate.rs`.
+        #
+        # THE SKIP REPORTS AS PASS, NOT SKIP, and that is the half of #1815 this
+        # file does not close. #1815 armed `FERRO_ASSERT_SEQUENCE` in
+        # `test-oracle`, which is the only one of the four oracles that can see a
+        # wrong-bases output on these axes — but that job provisions no manifest
+        # either, so they early-return there too. The one job that sets both is
+        # `nightly-mutalyzer.yml`, and it is `continue-on-error` by design.
+        # Provisioning a reference here is not a small change: that workflow
+        # spends ~30 min in `ferro prepare` per run and the result is deliberately
+        # uncached (a 3.25 GiB entry against a 10 GiB repository budget, #1218).
         if: matrix.shard == 1
         run: |
-          echo "::notice title=Reference-aware tests skipped on PR CI::PR CI does not provision FERRO_MANIFEST; the mutalyzer / biocommons reference-aware test suites silently skip. The nightly nightly-mutalyzer.yml workflow runs the full suite against a real manifest."
+          echo "::notice title=Reference-aware tests skipped on PR CI::PR CI does not provision FERRO_MANIFEST; the mutalyzer / biocommons reference-aware test suites silently skip, and nextest reports that skip path as PASS rather than SKIP. No gating job runs those axes, so a denoted-sequence violation on them still cannot redden a required check (#1815). The nightly nightly-mutalyzer.yml workflow runs the full suite against a real manifest, non-gating."
       - name: Run Rust tests (shard ${{ matrix.shard }} of 6)
         env:
           # Absence of a bulk fixture must FAIL here, never skip. The bulk and
@@ -1631,9 +1688,18 @@ jobs:
         # times (#1274, #1343, #1307) before it was asserted here.
         #
         # `FERRO_ASSERT_SEQUENCE` — the fourth oracle at this same seam (#1615) —
-        # is deliberately **not** set here, and this job is the one place the set
-        # is incomplete. It used to fire on two rows of this job's own selection,
-        # and both were real disagreements rather than noise to be suppressed:
+        # IS set here as of #1815, so this job is no longer the one place the set
+        # is incomplete. It runs over this job's whole selection MINUS
+        # `SEQUENCE_ORACLE_EXCLUDE`, whose rows and their open issues are
+        # documented on that variable; the step after this one re-runs those rows
+        # under the other three oracles, so nothing is traded away.
+        #
+        # EVERYTHING BELOW IS KEPT AS HISTORY, because the exclusion has to be
+        # read against it: three times a "the rows I know about are green" reading
+        # was taken for "it can be armed", and three times a selection-wide run
+        # found rows that reading could not reach. It used to fire on two rows of
+        # this job's own selection, and both were real disagreements rather than
+        # noise to be suppressed:
         #
         #   * #1618 — `NC_TEST.1:g.262TG[6]` -> `g.259_262GT[6]`
         #     (`repeat_tract_maximization`). `hgvs_to_spdi` read the anchored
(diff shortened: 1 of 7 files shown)


nh13:
@coderabbitai pause

PR #1869, opened 2026-08-14
test(normalize): relabel #1421's six rows SpecExplicit on delins.md:17

nh13:
Relabels the six `1421-n*` rows in `tests/it/reported_partition_verdicts.rs` from `Authority::SpecSelfConflicting` to `Authority::SpecExplicit`. This applies a decision two decided ledger records already reach; it does not make a new one.

Refs #1421. **Not `Closes`** — the reported defect is unfixed and three of the six rows stay `Verdict::Gap`. What changes is that their `wanted` split stops being a *record* of what the issue asked for and becomes a *target*, i.e. a declared deviation from a rule-2 preference clause rather than a neutral characterization. Nothing about ferro's output changes.

Representation-Change: none. No normalized string moves — the change is confined to the `authority` field, one census tuple and prose in two test modules. `every_gap_row_is_returned_exactly_as_authored` and `every_reported_spelling_still_normalizes_as_the_reports_recorded` pin all eighteen outputs and are unchanged and green, so the six rows print exactly what they printed before. No file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched at all.

## The clause, named per row

The clause is **`DNA/delins.md:17`** for all six rows — *"two variants separated by one or more nucleotides should be described individually and **not** as a `delins`"* — which `general.md:34` states word for word. Both spellings of a pair share one authority because the authority is a property of the shape, which `the_spec_authority_census_holds` asserts independently.

The clause's antecedent is satisfied on each pair. Geometry measured against the module's own `TEMPLATE` literal, which `reference_bases_are_what_the_reports_state` pins independently:

| row | clause | separation | span → payload | direction |
|---|---|---|---|---|
| `1421-n1/split` | `DNA/delins.md:17` | 2 nt unchanged at 30–31 (`AC`) | 29–33 (5 nt, `CACGT`) → `AACACATACTG` (11 nt) | net **+6** |
| `1421-n1/span` | `DNA/delins.md:17` | as above | as above | net **+6** |
| `1421-n2/split` | `DNA/delins.md:17` | 2 nt unchanged at 33–34 (`TG`) | 32–36 (5 nt, `GTGCA`) → `TTGGAATCGAC` (11 nt) | net **+6** |
| `1421-n2/span` | `DNA/delins.md:17` | as above | as above | net **+6** |
| `1421-n3/split` | `DNA/delins.md:17` | 2 nt unchanged at 35–36 (`CA`) | 34–38 (5 nt, `GCAAT`) → `TCACCTTTACG` (11 nt) | net **+6** |
| `1421-n3/span` | `DNA/delins.md:17` | as above | as above | net **+6** |

Separation is two nucleotides on every row, so `:17` applies and `delins.md:18`'s one-nucleotide exception does not — already pinned by `the_1421_spans_separate_by_two_nucleotides_not_one`, and doubly inapplicable since a `g.` description has no amino acid.

## Why `:17` is now unopposed

The old label rested on `DNA/delins.md:44-47` reaching these rows and answering the other way. Two **decided** records scope it out, on independent grounds, either alone sufficient:

- **`delins-payload-coincidence-carve-out-is-coding-dna-scoped`** — *"`delins.md:47`'s carve-out is scoped to the CODING DNA AXIS: `c.`, and nothing else. On the other DNA axes - `g.`, `m.`, `o.`, `n.` - `general.md:34` governs and the members are described individually."* Every `Row::input` in the module is `TEMPLATE:g.`; there is no `c.` row anywhere in the file.
- **`delins-merge-vs-individual-gap-two-or-more`** (the direction scope, 2026-08-11) — *"This ruling reaches the NET-DELETION case — a payload SHORTER than the span it replaces — and it does NOT reach net insertions, **where the split form remains canonical**."* All three pairs are net insertions of +6.

`:44` and `:45` are the worked `c.` example (`LRG_199t1:c.850_901delinsTTCCTCGATGCCTG`) and its gloss, and `:46` is a `NOTE` that *constructs* the alternative split and prints it. The single word `recommended` on `:47` is the whole prescriptive content of the passage, and it is the clause both records scope. So nothing survives the scoping to compete with `:17`.

Note the second ground does not depend on the first: were `:47`'s axis scope ever widened, these rows would still be `SpecExplicit` on the direction scope alone.

## What moved, and what did not

The `(explicit, conflicting)` tuple in `the_spec_authority_census_holds` is the **only** pin that moves:

| pin | before | after |
|---|---|---|
| `the_spec_authority_census_holds` `(explicit, conflicting)` | `(6, 12)` | **`(12, 6)`** |
| `OPEN_GAPS` | 12 | 12 |
| `PAIR_STATE_CENSUS` | `(0, 6, 3)` | `(0, 6, 3)` |
| `FIVE_PRIME_MOVERS` | `["1419-r3/cis", "1419-r3/span", "1420-v2/cis"]` | unchanged |
| `Verdict::Gap` / `Verdict::Canonical` row counts | 12 / 6 | 12 / 6 |
| every `input` / `output` / `five_prime` / `wanted` / `verdict` field | — | unchanged |

Verified by diff: no `wanted:`, `output:`, `five_prime:`, `input:`, `verdict:` or `label:` line appears in the change.

The set assertion beside the tuple is narrowed from `1419-* || 1421-*` to `1419-*` and given a non-emptiness precondition. Left as-is it would have kept passing while a `1421-*` row regressed back to the old label; and had #1419 ever been relabelled too it would have degenerated into a vacuous `.all()` over an empty vector — a green assertion checking nothing.

## Deliberately out of scope

- **#1419's six rows stay `SpecSelfConflicting`.** They are net *deletions* (−12/−15/−12), so the direction scope does not reach them; and their `wanted` string **is** `:46-47`'s own answer, so relabelling them forces re-deriving `wanted` onto `:17`, flipping six rows `Gap` → `Canonical`, taking `OPEN_GAPS` to 6, emptying `PairState::NeitherReaches` and structurally breaking `each_pair_reaches_its_wanted_form_from_the_spellings_its_state_names`. That amounts to declaring #1419 closed and owes its own ruling record. It is not a relabel and is not folded in here.
- **#1420's six rows are untouched.** `delins-recommendation-reach-when-the-input-arrives-split` records that `1420-v2/cis` and `1420-v3/cis` "must stay `Gap`" until it is decided, and `1420-v2/cis` carries an in-file `TRAP` note: its `wanted` string is also what a re-partition across its three unchanged interior bases produces, which `reported_confluence_pairs::the_1420_v2_pair_does_not_converge_by_re_derivation` pins as forbidden.

## Verification

Run in a worktree branched from `origin/main` @ `64fc76df`, exit status echoed unpiped into each log and read back out of the file:

| gate | result |
|---|---|
| `reported_partition_verdicts` + `reported_confluence_pairs`, baseline before the edit | 20 run, **20 passed**, `EXIT=0` |
| same, after the edit | 20 run, **20 passed**, `EXIT=0` |
| `cargo nextest run --features dev --lib --test it --no-fail-fast` | 10440 run, **10440 passed**, 150 skipped, `EXIT=0` |
| `cargo fmt --all --check` | `EXIT=0` |
| `cargo clippy --all-features --all-targets -- -D warnings` | `EXIT=0` |

The generated spec fixture was regenerated (`cargo run --features dev --bin generate_spec_fixture`, `EXIT=0`) before the suite, so `ruling_records_are_intact` was judged against a current artifact — the ledger reads 28 records, 25 decided, 3 undecided at this base, and both records cited above are `decided`.

Every clause quotation above was read out of the pinned `assets/hgvs-nomenclature` checkout (`6f85311`) and every ledger quotation straight out of `hgvs_spec_normalization_overrides.json` at this branch's HEAD; the geometry table was computed from the module's own `TEMPLATE` literal. No figure is relayed.

Note that #1419 and #1421 both open with a disclaimer that their bodies were drafted by an AI agent and filed under @msto's account, "not a statement authored by @msto" — so nothing here is argued from what the issue asks for. The argument is the clause and the two records.

### DIFF
--- a/tests/it/reported_confluence_pairs.rs
+++ b/tests/it/reported_confluence_pairs.rs
@@ -34,14 +34,15 @@
 //! `reported_partition_verdicts` now pins what each of the eighteen spellings
 //! prints — under 3' *and* under 5' — together with the form its issue asks for
 //! and what backs that form under the project's precedence policy
-//! (**spec-explicit > Mutalyzer > our judgement**). Six of those eighteen
-//! spellings — three pairs — have a spec line that answers them outright, which
-//! is the split `the_spec_authority_census_holds` pins next door as `(6, 12)`
-//! rows; #1419's six and #1421's six do not, because `delins.md:44-47`
-//! recommends the merged delins for the same alignment coincidence each pair's
-//! split relies on (`general.md:56` for #1419, the unchanged interior
-//! reappearing in the payload for #1421) — so those stay recorded rather than
-//! targeted.
+//! (**spec-explicit > Mutalyzer > our judgement**). Twelve of those eighteen
+//! spellings — six pairs — have a spec line that answers them outright, which is
+//! the split `the_spec_authority_census_holds` pins next door as `(12, 6)` rows:
+//! #1420's six on `delins.md:16`/`:17`, and #1421's six on `:17`, whose only
+//! competitor `delins.md:47` is scoped away from the `g.` axis and away from net
+//! insertions by two decided ledger records. #1419's six do not, because
+//! `delins.md:46-47` recommends the merged delins for the alignment coincidence
+//! their split relies on, and their `wanted` form is that recommendation's own
+//! answer — so those stay recorded rather than targeted.
 //!
 //! So this module keeps exactly one job: the **ratchet**.
 //! [`CONVERGING_PAIRS_THREE_PRIME`] and [`CONVERGING_PAIRS_FIVE_PRIME`] are the
--- a/tests/it/reported_partition_verdicts.rs
+++ b/tests/it/reported_partition_verdicts.rs
@@ -112,34 +112,51 @@
 //!   payload correspond column for column and no alignment is ever chosen. Their
 //!   one unchanged interior column is simply unchanged — not a "part of the
 //!   inserted sequence" that had to be *found* to align.
-//! * **#1421 is [`Authority::SpecSelfConflicting`], for the same reason #1419
-//!   is.** An earlier revision labelled it `SpecExplicit`, on the grounds that
-//!   `delins.md:44-47` "is conditioned on the span losing length". It is not.
-//!   `:47` states its condition as *"parts of the inserted sequence `align` with
-//!   the reference sequence"* and says nothing whatever about length; the
-//!   `c.850_901` example happens to be a net deletion, and that incidental
-//!   arithmetic was mistaken for the rule.
+//! * **#1421 is [`Authority::SpecExplicit`] on `delins.md:17`**, so its wanted
+//!   split is a target rather than a record. The full argument, with both
+//!   records quoted, is on [`the_spec_authority_census_holds`]; in one sentence,
+//!   `:17` (*"two variants separated by one or more nucleotides should be
+//!   described individually and **not** as a `delins`"*, `general.md:34` word
+//!   for word) applies on its own terms, and the passage that used to oppose it
+//!   reaches these rows on neither of its two scopes.
 //!
-//!   Length does bear on it, but the other way round. Every #1421 row replaces
-//!   **5** reference bases with an **11**-base payload (net +6), so there is no
+//!   **This label has now been wrong in both directions, so read what changed.**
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai pause

PR #1872, opened 2026-08-14
feat(hgvs): refuse a bare-transcript intronic position at the strict parse stage

### LINKED ISSUE #1630: normalize: enforce every absolute prohibition at the mode-gated parse stage the ruling decided

nh13:
Implements the third and last of the three stage defects the decided `rulings[absolute-prohibition-enforcement-stage]` names in its own census.

Representation-Change: 1,031,655 of 5,163,595 parseable `c.`/`n.` corpus rows change their strict-mode PARSE verdict, from accepted to refused (`W4007`). **No normalized output moves — 0 rows respell**, and nothing under `src/normalize/` changes what it emits; what moves is the accept/reject boundary of `parse_hgvs_with_config(_, ErrorConfig::strict())` for a bare-transcript intronic description (`NM_…:c.20+2del`, `NR_…:n.20+2del`, and inside an allele). Those inputs were already **rejected** in strict mode one stage later, at normalize, by a rung that short-circuits before normalization — so a caller that parses and then normalizes strictly sees the same verdict, reached earlier, with `FerroError::Parse` in place of `FerroError::IntronicVariant`. The genuinely new refusal is for a strict caller that parses **without** normalizing, and that is the migration this discloses: it is a previously-accepted input, not a previously-rejected one. Lenient parse newly emits a `W4007` warning where it was silent; silent mode is unchanged; the CLI and the Mutalyzer conformance runner both use the config-less `parse_hgvs` and are unaffected.

## The ruling

`absolute-prohibition-enforcement-stage`, status **`decided`**, read from `origin/main` at `64fc76df`. Governing clause `docs/recommendations/checklist.md:5`. Its ruling paragraph:

> **OPERATOR RULING, 2026-08-10 — DECIDED.** The enforcement stage is MODE-DEPENDENT, uniformly across every absolute prohibition:
> - STRICT — fail at PARSE. Strict implies it validates input conformance, not merely that the input can be parsed.
> - LENIENT — input conformance is NOT validated. It fails only if it cannot NORMALIZE, which can happen.
> - SILENT — lenient with no error messages, but an exit code.

Its scope sentence for this half, and the rationale this PR acts on:

> **`checklist.md:20`'s refusal is at normalize, not at parse.** It is a conditional clause and its own record (`bare-transcript-intronic-position`) ratifies the mode split; moving the stage is cosmetic there, but it should move for uniformity.

The general ground it gives: "whether the INPUT conforms is answered before the input is accepted, not part-way through normalizing it."

**"Cosmetic" is about the verdict, not about what a caller sees**, and that is why this is a guard rather than a refactor. `parse_hgvs_with_config` is public API and validating an identifier is its most obvious use; such a caller was told a bare `NM_…:c.20+2del` conforms.

## The census, re-derived rather than relayed

The record's census names **56 rows refused in no mode**. On `origin/main` at `64fc76df` that figure is now **24**, measured over all 164 `RowKind::Prohibited` rows of `conformance::spec_corpus` through `parse_hgvs_with_config` in all three modes:

| clause | spellings | strict | lenient / silent | state |
|---|---|---|---|---|
| `checklist.md:32` (`ins6`) | 24 | accept | accept | **refused in no mode** — the whole residue |
| `standards.md:39` (`X`) | 24 | parse | normalize | landed, #1684 |
| `checklist.md:16` (`+`) | 4 | parse | normalize | landed, #1628 |
| `checklist.md:45` (`-`) | 4 | parse | normalize | landed, #1628 |
| `checklist.md:20` | 16 | **normalize** | accept (verbatim) | **this PR** |
| `checklist.md:26`, `:30`, `:16`'s `*` | 44 | parse | parse | unconditional; the relaxation half |
| `checklist.md:49` (`del3`) | 24 | parse | 16 normalize, 8 parse | the 8 are the 5'UTR repair bug |
| `general.md:96` (space) | 24 | parse | normalize | already conformant |

So 32 of the 56 landed after the record was written, and the remaining 24 are `ins6` — carried by PR #1797, not by this change. The reconciliation is exact and the residual is not inferred from a net.

## Scope, and what this deliberately leaves

This is **not** `Closes #1630`. That issue's acceptance criterion is deleting the `#[ignore]` from `the_decided_target_is_a_mode_gated_refusal`, whose last open row is `ins6`. Its item 2 also remains: `checklist.md:26`, `:31` and `:16`'s `*` are refused unconditionally today and the ruling makes them strict-only. That is a **relaxation** requiring new grammar productions for shapes that currently reach none — and for `c.10insT` the spec calls the shape "ambiguous" in its own words, so building an AST for it is a representation choice the record does not make. The record sequences the tightening ahead of the relaxation for exactly that reason, and this PR stays on the tightening side of that line.

## What moved, and why the normalize rung stays

The predicate moves to `src/hgvs/bare_transcript_introns.rs`, where the parse rule, the normalize rung and #1704's output re-parenting all read it. The clause previously had its scope prose written out twice; two readings of one clause is how ferro came to refuse a description in strict mode while manufacturing the identical description in lenient.

The normalize rung is **kept** and is not redundant: it answers for a caller arriving through the config-less `parse_hgvs` (#1632) or a lenient parse plus a strict `Normalizer`, and it carries the `EINTRONIC` tag `mutalyzer_map.rs` keys off the `IntronicVariant` variant name for. The Mutalyzer conformance runner uses `parse_hgvs`, so that mapping is untouched.

## Tests

Written first and watched fail for the right reason — strict parse returned `Ok` with `genomic_context: None`, an intronic offset and an empty warning vector. All three new/changed guards were then re-proved red by disabling the rule and restored with `cp`.

Three tests pinned the old behaviour and are rewritten rather than deleted, each keeping the claim it was written to make:

- `corpus_alphabet_and_intronic_validity` held a PINNED FINDING that no mode refused the bare intronic allele at parse. It is now the adjudicated-correct guard for the mode split, with the genomic-wrapper form as its positive control.
- `issue_1632_parse_entry_applies_no_mode` used this shape as a *negative control* for #1632, on the ground that the two entries agreed. They now disagree, so it becomes a *witness* for #1632 — pinning the disagreement so that closing that issue goes red and names the shape whose behaviour it changes.
- `genomic_axis_offset_prohibition`'s neighbour control kept transcript-axis rows to show `W4009` does not reach them. It now asserts that a strict refusal there cites `W4007` and never `W4009`, which is the cross-axis leak it existed to catch, plus a wrapper-form control.

Both controls on the new guard are load-bearing: without the wrapper control a bare `is_err()` is satisfied by refusing intronic positions outright, and without the exonic control by refusing bare `NM_` accessions outright.

`PASSING_CENSUS`'s `ModeDivergencePinned` moves 120 to 288, **attributed rather than inferred**: exactly 56 distinct spec-harvested inputs carry a `W4007` strict-parse refusal, each contributes exactly 3 rows, and 288 - 168 leaves precisely the previous 120, so nothing else moved and nothing left. The 56 here and the 56 in the record's census are **different sets** and the shared digit is a coincidence, named in the code comment so it is not read as a link.

## Gate

`cargo fmt --check`, `cargo clippy --all-features --all-targets -- -D warnings` and `FERRO_REQUIRE_BULK_FIXTURES=1 cargo nextest run --features dev --no-fail-fast` all clean — 10639 passed, 0 failed, zero `Skipping:` lines. The ruling ledger is untouched: 28 records, none dropped, no status changed.


nh13:
@coderabbitai pause

Issue #1853, opened 2026-08-13
WATCHED_PREFIXES undercovers what moves output: src/reference/, src/error_handling/ and src/data/ disclose real moves the gate never asks for


nh13:
`WATCHED_PREFIXES` in `scripts/check_representation_change.py` names four directories:

```
  src/normalize/  src/hgvs/  src/spdi/  src/project/
```

Changes outside them require no `Representation-Change:` trailer. But changes outside them **do** move consumer-visible output, and the authors have been declaring it voluntarily — which is the evidence that the gate is in the wrong place, not that the authors need reminding.

This issue is a measurement and a proposal. **It deliberately does not widen the tuple**; that is a coverage decision with a cost, and it belongs to the operator.

## Method

Two independent populations, so a finding is not an artifact of either.

1. **Merged history** — every commit on `origin/main` (1,002 scanned), trailer extracted with the repo's own `find_declaration`, the full multi-line value recovered with `inject_representation_disclosure.disclosure_value`, decline-vs-real decided by the repo's own `declines`. Changed paths from `git show --name-only --no-renames`.
2. **All 39 open PRs** — same functions, file lists from the GitHub files API.

Counts: 138 of the 1,002 merged commits carry a trailer (30 real, 108 declines). 22 of the 39 open PRs carry a real (non-declining) disclosure.

## What moves output outside the gate

**Merged — 3 of the 30 real disclosures touch no watched directory:**

| commit | dir | disclosure |
|---|---|---|
| `988c9a45` (#1594) | `src/error_handling/` | "82 previously-rejected inputs now parse; no normalized output moves." Measured over 368,870 ClinVar protein descriptions. The accept/reject boundary is a consumer-visible surface. |
| `7f6b0711` (#1661) | `src/error_handling/` | "1 pinned (row, mode) result moves … Previously **accepted**, so it is a real migration for a consumer storing that string" |
| `822eb5d5` (#1623) | `src/conformance/` | "0 rows move in this PR" — a *decision* record, see below |

**Open — 5 of the 22 real disclosures touch no watched directory:**

| PR | dir | disclosure |
|---|---|---|
| #1724 | `src/reference/` | 50 accessions move from declining to served on the c./n. axes; previously-erroring inputs now produce output |
| #1737 | `src/data/`, `src/reference/` | 56 of 482,519 GRCh38 cdot builds move, all on `cds_end`; "a real migration for a reference-backed consumer" |
[...]

PR #1876, opened 2026-08-14
ci: watch src/reference/ and src/error_handling/ for representation changes

nh13:
`WATCHED_PREFIXES` in `scripts/check_representation_change.py` named four directories — `src/normalize/`, `src/hgvs/`, `src/spdi/`, `src/project/`. This adds two: **`src/reference/`** and **`src/error_handling/`**.

Refs #1853, which proposed **three**. It stays open on purpose: one of its three (`src/data/`) is being declined here on measurement, and that disposition should be recorded there rather than swept up by an auto-close.

## Why these two, and not `src/data/`

The decisive evidence is the v0.13.0 cycle — the one place output movement was measured without relying on anyone declaring it. 87 commits shipped carrying **zero** trailers, and the changelog's **Representation changes** section had to be reconstructed by hand, normalizing 5,761,302 expressions through both releases against the same prepared reference and diffing row by row. Its **326,404** newly-normalizing rows attribute to exactly two PRs, and the attribution is the finding:

| PR | commit | directories touched | watched before? |
|---|---|---|---|
| #1490 | `3ac054d3` | `src/reference/` **only** | no |
| #1501 | `feb5dc9f` | `src/error_handling/` **only** | no |

Merged history since the gate landed agrees. The only real out-of-gate disclosures are #1594 and #1661 (`src/error_handling/`, the second a previously-**accepted** string and therefore a genuine migration, not a free addition) and #1724 (`src/reference/`, merged since the backtest was written).

`src/data/` is dropped from #1853's proposal because **its marginal contribution is exactly zero in both measured populations**:

- its one real disclosure (#1737) also touches `src/reference/`, so the entry added here already gates it;
- both its merged gate-era commits (#1786, #1770) declined;
- **zero** commits touched it in the v0.13.0 blind cycle;
- its only measured effect is turning draft #1156 red.

The single fact that would earn it a place is a real, non-declining disclosure touching `src/data/` and **not** `src/reference/`. That does not exist in either population today. If a successor to #1156 moves cdot's CDS bounds alone and moves output, it earns its place on evidence — one PR, no argument needed. This is written down on `WATCHED_PREFIXES` itself so the next reader does not re-propose it from the mechanism argument.

`src/conformance/` stays excluded too, and deliberately: it is measurement and adjudication code that by construction cannot move ferro's output, so gating it would demand a declaration about something the directory cannot do. Both of its real disclosures (#1623 merged, #1839 open) open `0 rows move in this PR`. After this change the *entire* remaining missed column in both populations is those two conformance decision records.

## Measured

Re-derived independently against `origin/main` @ `2b979da2` (1,003 merged commits; 48 open PRs), using the repo's own `find_declaration` / `declines` / `disclosure_value` and `git show --name-only --no-renames`, matching what `ci.yml` passes the checker.

Merged history — 31 real disclosures, 108 declines, 864 silent:

| configuration | caught | missed | tax (`none`) | extra gated | would go red |
| Current (4) | 27 | 4 | 30 | — | 0 |
| + `reference` | 28 | 3 | 30 | 1 | 0 |
| #1853 (+ `reference` + `error_handling` + `data`) | 30 | 1 | 32 | 5 | 0 |
| **this PR (+ `reference` + `error_handling`)** | **30** | **1** | **30** | **3** | **0** |
| all of `src/` | 31 | 0 | 50 | 24 | 0 |

This matches #1853's backtest with every cell offset by exactly the merge of **#1724**, which was an *open* `src/reference/` disclosure when the backtest ran and is now merged history. That merge is the only change, and it strengthens the case: `src/reference/` went from zero merged evidence to one merged real disclosure.

This PR reaches #1853's catch rate exactly — 30 caught, **0 real merged moves missed** — while gating two fewer commits and forcing two fewer `none`s.

## Nothing strands, and nothing newly reddens

Widening the tuple adds no required **context**. `Representation change declared` already runs on every PR and already reports; only its verdict changes. So the standing objection — that a new required context strands every open PR on "Expected — waiting for status to be reported" — does not apply.

Re-verified against live PR bodies and the GitHub files API at current heads:

- **12** open PRs touch `src/reference/` or `src/error_handling/`; **12 of 12 already carry a trailer.** Zero go red.
- **6** of those are *newly* gated (they touch a newly watched directory and no previously watched one): #1861, #1851, #1830, #1783, #1741, #1737. All six are non-draft and all six already declare voluntarily.
- The one PR red under this configuration, **#1024**, is red under the *current* four as well — it touches `src/project/projector.rs`. This change neither causes nor fixes it; the fix there is a rebase to mint a fresh run.
- #1156, the single PR #1853's three-directory proposal would have reddened, touches only `src/data/` and is therefore **not gated at all** by this PR.

The earlier "11 touch a newly-gated directory, 10 already declare" figure was measured over the three-directory proposal and a 39-PR population. At current heads, over 48 open PRs and the two directories actually being added, it is 12 of 12 — the reassurance holds and is stronger.

## Tests

`tests/python/test_representation_change_trailer.py` asserts through `check()` on synthetic file lists rather than restating the constant, and pins both sides of the boundary:

- every watched directory demands a declaration (parametrized over one real path per directory, now six);
- every **deliberately unwatched** directory does not — `src/conformance/`, `src/data/`, `src/convert/`, each carrying its reason inline, so an exclusion cannot be mistaken for an oversight;
- a dedicated test pins the two additions and the `src/data/` exclusion end-to-end, plus a length assertion so a seventh directory cannot be added in silence.

Watched fail first, then pass. Before touching `WATCHED_PREFIXES`: 3 failed, 98 passed, both new prefixes reporting `watched_files([path]) == []`. After: 101 passed. Perturbing the tuple with `src/data/` re-reds three tests, including `test_watched_prefixes_match_the_release_config` — so the release-config pin and the exclusion guard are both live rather than vacuous.

One defect fixed while in there. `test_watched_prefixes_match_the_release_config`'s docstring claimed a set comparison against `release-plz.toml` — "exactly the drift that would make this check pass on the changes it exists to catch" — while the assertion only ran checker → config. That left the dangerous direction open: a prefix silently **dropped** from `WATCHED_PREFIXES` still satisfied "every watched directory is named in the config". It is now a real set comparison in both directions, and reads the config side as every backticked `` `src/<dir>/` `` token so it survives the prose re-wrap this PR forces. Verified by deleting `src/spdi/` from the tuple: that test now fails, where before it passed.

## Documentation kept in step

Four places state the gated set and all four are updated: `CONTRIBUTING.md`, `release-plz.toml`'s release-PR reviewer checklist (which `test_watched_prefixes_match_the_release_config` pins the checker against), `.github/workflows/ci.yml`'s job comment, and `CLAUDE.md`. Each now points at `WATCHED_PREFIXES` as the authority rather than competing with it.

One statement of the old four is left alone on purpose: the doc comment in `tests/it/spec_conformance_axis.rs` is a historical record of what a specific past branch touched when it was measured, not a statement of the current gate — editing it would falsify the record it exists to keep.

Representation-Change: none. This is a CI gate and its documentation; no `src/` file changes, so no normalized output can move. What changes is which changes are *required to declare*, which is a property of the checker rather than of ferro's output.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -277,14 +277,21 @@ jobs:
       - name: Run mypy
         run: uv run poe check-typing
 
-  # A change under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` can move
-  # the normalized string a downstream consumer stores read counts against, so it must say
-  # whether it does. `release-plz.toml` renders the changelog's **Representation changes**
-  # section from a `Representation-Change:` trailer on the squash commit (#1433) — the
-  # machinery works, but across the v0.13.0 cycle 87 commits landed carrying zero trailers
-  # while 46 touched those directories, and the section rendered empty. An empty section
-  # looks identical whether nothing moved or nobody declared, so it went unnoticed for ~90
-  # commits and the disclosure had to be reconstructed by hand before release (#1522).
+  # A change under `src/normalize/`, `src/hgvs/`, `src/spdi/`, `src/project/`,
+  # `src/reference/` or `src/error_handling/` can move the normalized string a downstream
+  # consumer stores read counts against, so it must say whether it does.
+  # `release-plz.toml` renders the changelog's **Representation changes** section from a
+  # `Representation-Change:` trailer on the squash commit (#1433) — the machinery works,
+  # but across the v0.13.0 cycle 87 commits landed carrying zero trailers while 46 touched
+  # the four directories watched at the time, and the section rendered empty. An empty
+  # section looks identical whether nothing moved or nobody declared, so it went unnoticed
+  # for ~90 commits and the disclosure had to be reconstructed by hand before release
+  # (#1522). The last two directories were added by #1853, on the finding that every
+  # newly-normalizing row in that reconstruction attributes to one PR touching only
+  # `src/reference/` and one touching only `src/error_handling/`. The authoritative list
+  # is `WATCHED_PREFIXES` in the script this job runs; this comment restates it for the
+  # reader, and `test_the_prose_restatements_name_every_watched_directory` fails when a
+  # watched directory is missing from it — the obligation is enforced, not promised.
   #
   # `Representation-Change: none` passes. The point is to force the judgement, not to force
   # a disclosure; what this rejects is silence.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -752,10 +752,19 @@ Because the fixture is no longer in git, per-PR `parse-error → preserved` stat
 
 ### Declaring a representation change: a REQUIRED check, not a convention
 
-A PR touching `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` must carry a
-`Representation-Change:` trailer **in its PR description**, or the `Representation change
-declared` check fails. That context is in `main`'s ruleset, so a missing trailer **blocks the
-merge**.
+A PR touching `src/normalize/`, `src/hgvs/`, `src/spdi/`, `src/project/`, `src/reference/`
+or `src/error_handling/` must carry a `Representation-Change:` trailer **in its PR
+description**, or the `Representation change declared` check fails. That context is in
+`main`'s ruleset, so a missing trailer **blocks the merge**.
+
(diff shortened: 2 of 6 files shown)


nh13:
@coderabbitai pause

Issue #1746, opened 2026-08-12
cds_to_tx answers c.? with cds_start, inventing a coordinate for a position that has none


nh13:
`CoordinateMapper::cds_to_tx` answers `c.?` with `Ok(cds_start)` — inventing a transcript coordinate for a position that has none.

Measured directly:

```
PROBE cds_start=Some(5)
PROBE direct cds_to_tx(c.?) => Ok(5)
PROBE direct cds_to_tx(c.1) => Ok(5)
```

Byte-identical. The mechanism is that `CDS_BASE_UNKNOWN` is the integer `0`, so `c.?` takes the `base < 1` 5'UTR branch with a zero displacement.

## Reach — narrower than it first appears

A **parsed** `c.?` never gets here: `hgvs_to_spdi` refuses it one layer up (`cannot convert c. variant with unknown start position`, the `Mu::Unknown` guard at `src/hgvs/parser/variant.rs:489`). This is reachable only through a **directly constructed** `CdsPos` with `base == 0` — the Rust API, and the Python `c_to_n`/`c_to_g` entry points (`src/python.rs:6832`).

## It contradicts the project's own doctrine

`CdsPos::has_unknown_offset` (`src/hgvs/location.rs:184-199`) states it for the offset case: *"no coordinate can be derived from them at all"* and *"callers doing coordinate arithmetic MUST check this first"*.

Two callers already work around this defect rather than relying on the function:

- `src/project/projector.rs:2781` and `:4649` reject `is_unknown()` **before** calling `cds_to_tx`
- `:4643-4647` names the mechanism exactly: *"without this guard `base <= 0` would misclassify it as 5'UTR and return `Ok`"*

The guard belongs at the site those two are compensating for.

Issue #1745, opened 2026-08-12
get_amino_acid_at underflows on a zero cds_start, in both build profiles


nh13:
`CoordinateMapper::get_amino_acid_at` computes

```rust
cds_start as usize - 1 + (position as usize - 1) * 3
```

with no guard on `cds_start`. `Transcript` stores `cds_start` exactly as its provider supplied it and validates nothing (`Transcript::new`, `src/reference/transcript.rs:421-456`; `reference/validate.rs`'s CDS check is out-of-band and report-only), so `Some(0)` is representable — a 1-based coordinate naming no transcript base.

## Both profiles fail, and the release failure moves

```
debug:   panicked at src/convert/mapper.rs:203:27: attempt to subtract with overflow
release: panicked at src/convert/mapper.rs:213:25:
         start byte index 18446744073709551615 is out of bounds for string of length 18
```

Release wraps to `usize::MAX` and then aborts on the string slice, reporting an out-of-bounds index rather than the bad metadata that caused it — so the diagnosis points at the wrong thing.

Reached through the public `cds_to_protein`, which is what the Python `c_to_p` binding calls.

## In-tree evidence that 0 is expected to be representable

`Transcript::utr5_length` (`src/reference/transcript.rs:557`) already defends the same field with `saturating_sub(1)`. One reader of `cds_start` guards it; this one does not.

PR #1877, opened 2026-08-14
fix(data): refuse an inverted-CDS record instead of underflowing on it

### LINKED ISSUE #1871: genome_to_cds underflows on an inverted-CDS cdot record, in both build profiles
`CoordinateMapper::genome_pos_to_cds_pos` computes the last CDS position on its intronic 3'UTR arm as a bare subtraction of two `u64`s, with no ordering precondition:

```rust
CdsPosition::ThreePrimeUtr(utr_offset) => {
    info.in_3utr = true;
    let cds_end = tx.cds_end.unwrap_or(0);
    let cds_start = tx.cds_start.unwrap_or(0);
    let last_cds_pos = (cds_end - cds_start) as i64;
```

`src/data/mapping.rs:696-698` on `origin/main` @ `2b979da2`.

`CdotTranscript::tx_to_cds` states no ordering precondition either. It tests `tx_pos < cds_start`, then `tx_pos < cds_end`, and files everything else as 3'UTR — so on a record where `cds_end < cds_start`, *every* transcript position at or past `cds_start` falls through to the 3'UTR arm and reaches the subtraction with `cds_end < cds_start`.

## Both profiles fail, and the release failure is the quiet one

```
debug:   panicked at src/data/mapping.rs:698:48: attempt to subtract with overflow
release: no panic — 10u64 - 50u64 wraps to 18446744073709551576,
         and `as i64` reinterprets that as -40
```

`[profile.release]` sets no `overflow-checks`, so release wraps. The wrapped value is returned as a `CdsPos { base: -40, utr3: true, .. }` — a negative base flagged as 3'UTR, which is not a coordinate any transcript has. So release answers, with a fabricated position, rather than failing.

## Reach: a cdot JSON file, not only a directly constructed `CdotTranscript`

`RawCdotTranscript::from_genome_build` (`src/data/cdot.rs:613-615`) takes the CDS bounds verbatim from the record's own transcript-level codon fields whenever **either** is present:

```rust
let (cds_start, cds_end) = if self.start_codon.is_some() || self.stop_codon.is_some() {
    (self.start_codon, self.stop_codon)
} else {
    // ... the genomic fallback, which DOES order its endpoints
};
```

Both are `Option<u64>` deserialized straight from JSON with `#[serde(default)]` and no ordering check. Note the asymmetry: the genomic-CDS fallback in the `else` arm deliberately orders its two endpoints (`let (start, end) = if s < e { (s, e+1) } else { (e, s+1) }`), so it cannot produce an inverted record. The codon branch has no such step.

Measured directly, loading through the public `CdotMapper::from_reader`:

```
PROBE loaded cds_start=Some(50) cds_end=Some(10)
thread '...' panicked at src/data/mapping.rs:698:48: attempt to subtract with overflow
```

That is a plain cdot JSON document with `"start_codon": 50, "stop_codon": 10` and two exons, queried at an intronic genomic position via the public `CoordinateMapper::genome_to_cds`. So the reach is wider than #1746's, which is Rust-API-only.

## No shipped reference data is affected — this is malformed-input hardening

Scanned all three prepared cdot files for records where both codon fields are present and `start_codon > stop_codon`:

| file | transcripts | both codons | inverted |
|---|---|---|---|
| `cdot-0.2.32.refseq.GRCh38.json` | 482,519 | 355,590 | **0** |
| `cdot-0.2.32.ensembl.GRCh38.json` | 574,584 | 266,774 | **0** |
| `cdot-0.2.32.refseq.GRCh37.json` | 197,846 | 158,791 | **0** |

**0 of 781,155 real records** are inverted, so nothing in the shipped reference reaches this today. Stating that plainly: this is defensive hardening against a malformed record, not a live corruption of real output. It moves no row of any corpus.

The scan cross-validates against figures already recorded in `src/data/cdot.rs`'s own comments — 355,590 RefSeq GRCh38 records carrying both codon fields, 266,774 Ensembl, and exactly one GRCh38 record carrying `stop_codon` but no `start_codon` — all three reproduced exactly, which is what makes the zero trustworthy rather than an artifact of a wrong query.

## In-tree evidence that this condition is expected to be guarded

Two readers already defend it, and this one does not — the same shape as #1745's `utr5_length`/`get_amino_acid_at` asymmetry:

- **`merge.rs`'s `ordered_cds_bounds`** (`src/normalize/merge.rs:4324-4346`) refuses it by name, and its doc comment states the reasoning: *"`cds_end < cds_start` is a malformed record whose `c.` axis is incoherent: the 5'UTR and 3'UTR halves overlap, so one transcript position has two names ... a conversion that does not refuse simply picks one of the two spellings — silently."*
- **`boundary::get_cds_boundaries_with_axis_info`** guards the same condition the same way (`(Some(s), Some(e)) if e >= s`), which is the precedent `ordered_cds_bounds` itself cites.

`reference/validate.rs:206` and `:730` also test `cds_end < cds_start`, but as #1745 notes about the sibling CDS check, that pass is out-of-band and report-only — it gates nothing on this path.

## Note on the `unwrap_or(0)` fallbacks

They are unreachable as written and should not be read as the defect. `tx_to_cds` returns `None` if *either* bound is `None`, and the caller `?`-propagates that into `InvalidCoordinates` before the 3'UTR arm runs — so by the time these two lines execute, both are necessarily `Some`. The defect is the missing ordering check, not the missing-bound fallback.

## Family

Same shape as #1745 (`get_amino_acid_at` underflows on a zero `cds_start`, both profiles, guarded elsewhere in-tree but not here) and #1746 (`cds_to_tx` fabricates a coordinate for a position that has none rather than refusing). All three are unchecked arithmetic on provider-supplied CDS metadata that the type system permits and no load-path validation rejects.

nh13:
Closes #1871.

`CoordinateMapper::genome_pos_to_cds_pos`'s intronic 3'UTR arm computed the last CDS position as a bare `cds_end - cds_start` on two `u64`s, with no ordering precondition. `CdotTranscript::tx_to_cds` states no ordering precondition either — it tests `tx_pos < cds_start`, then `tx_pos < cds_end`, and files everything else as 3'UTR — so on a record where `cds_end < cds_start`, every transcript position at or past `cds_start` falls through to exactly this arm and underflows.

The code is unchanged since the initial commit.

## What it did, measured in both profiles

On a two-exon record with `cds_start = 50`, `cds_end = 10`, queried at an intronic genomic position:

```
debug:   panicked at src/data/mapping.rs:698:48: attempt to subtract with overflow
release: Ok(CdsPos { base: -40, offset: Some(491), utr3: true, special: None })
```

`[profile.release]` sets no `overflow-checks`, so release wraps to `18446744073709551576` and `as i64` reinterprets it as `-40`. The release half is the worse one: a negative base flagged `utr3` is not a position any transcript has, and it is returned as success rather than as an error.

## The fix

`checked_sub`, refusing with `InvalidCoordinates` and naming the transcript and both bounds.

Refusing is not a new idiom here — it is the project's established and already-tested position on this exact condition:

- **`merge.rs`'s `ordered_cds_bounds`** refuses it by name, and its doc comment gives the ground: an inverted record's 5'UTR and 3'UTR halves overlap, so one transcript position has two names, and a conversion that does not refuse silently picks one of them.
- **`boundary::get_cds_boundaries_with_axis_info`** guards it the same way (`(Some(s), Some(e)) if e >= s`).
- **#1284 pinned it as behaviour** — `issue_1284_transcript_axis_reparse::inverted_cds_bounds_are_refused_not_guessed`: *"The repair must refuse rather than pick one."*

So this closes the one site that did not yet honour a rule the tree already holds elsewhere — the same asymmetry #1745 reports for `utr5_length` vs `get_amino_acid_at`.

The two `unwrap_or(0)` fallbacks are left alone and are now documented as unreachable: `tx_to_cds` returns `None` unless both bounds are set, and that `None` is already `?`-propagated into an error before this arm runs. The defect was the missing ordering check, not the missing-bound fallback.

## Reach, stated honestly

Reachable from a **cdot JSON document**, not only from a directly constructed `CdotTranscript` — so wider than #1746's Rust-API-only reach. `RawCdotTranscript::from_genome_build` adopts `start_codon` / `stop_codon` verbatim whenever either is present, with no ordering check; the sibling genomic-CDS fallback in the same function *does* order its endpoints and so cannot produce an inverted record. Both reach paths are pinned by a test.

**No shipped reference data is affected.** Scanning all three prepared cdot files for records carrying both codon fields with `start_codon > stop_codon`:

| file | transcripts | both codons | inverted |
|---|---|---|---|
| `cdot-0.2.32.refseq.GRCh38.json` | 482,519 | 355,590 | **0** |
| `cdot-0.2.32.ensembl.GRCh38.json` | 574,584 | 266,774 | **0** |
| `cdot-0.2.32.refseq.GRCh37.json` | 197,846 | 158,791 | **0** |

0 of 781,155. The scan reproduces three figures already recorded in `src/data/cdot.rs`'s own comments (355,590 / 266,774 / the single GRCh38 record carrying `stop_codon` but no `start_codon`), which is what makes the zero trustworthy rather than an artifact of a wrong query.

So this is **defensive hardening against a malformed record, not a live corruption of real output** — worth doing, and worth not overstating.

## Scope

One guard, at one site. `mapping.rs:698` is the only place with this shape: every other CDS-bound subtraction in `src/data/` is already guarded by its enclosing branch condition, and `cds_pos_from_region` — the exonic counterpart — never performs this subtraction, so the exonic path was never affected.

Deliberately **not** widened: `mapping.rs:304`'s `cds_end + (base as u64) - 1` cannot underflow (`base >= 1` is checked immediately above) and its only risk is *overflow* on an absurd coordinate, which is a different defect already tracked by #1690. `merge.rs:4275`'s `map_or(0, …)` is a documented deliberate fallback and is untouched.

## Tests

Two, one per reach path, both pinning refusal rather than the old value. Both were watched failing before the fix — with the guard reverted they panic at the subtraction under `debug-assertions` — and passing after.

Representation-Change: none. No output moves: the change adds a refusal on a condition that 0 of 781,155 real cdot records satisfy, and every previously-answerable input still takes the same path to the same value. The only behaviour change is on an inverted record, which previously panicked in debug and fabricated a negative `utr3` base in release.

### DIFF
--- a/src/data/mapping.rs
+++ b/src/data/mapping.rs
@@ -519,11 +519,10 @@ impl CoordinateMapper {
         genome_pos: &GenomePos,
         transcript_id: &str,
     ) -> Result<(CdsPos, MappingInfo), FerroError> {
-        // `transcript_id` is part of the parameter list for API stability but
-        // no in-tree consumer reads `MappingInfo::transcript_id` — leaving it
-        // None avoids a `String::from` allocation per call (~400 calls per
-        // SNP project_all on chr17 fan-out).
-        let _ = transcript_id;
+        // No in-tree consumer reads `MappingInfo::transcript_id`, so it is left
+        // None: that avoids a `String::from` allocation per call (~400 calls per
+        // SNP project_all on chr17 fan-out). `transcript_id` is still named in
+        // the inverted-CDS refusal below, which allocates only on the error path.
         let mut info = MappingInfo::default();
 
         // Check if position is in an exon. `locate_genome_pos` returns both
@@ -693,9 +692,38 @@ impl CoordinateMapper {
                         }
                         CdsPosition::ThreePrimeUtr(utr_offset) => {
                             info.in_3utr = true;
+                            // Both bounds are necessarily `Some` here — `tx_to_cds`
+                            // returns `None` unless both are set, and that `None`
+                            // became the `InvalidCoordinates` above — so the
+                            // fallbacks are unreachable rather than load-bearing.
                             let cds_end = tx.cds_end.unwrap_or(0);
                             let cds_start = tx.cds_start.unwrap_or(0);
-                            let last_cds_pos = (cds_end - cds_start) as i64;
+                            // Refuse an inverted record rather than wrap. `tx_to_cds`
+                            // states no ordering precondition: it tests
+                            // `tx_pos < cds_start` then `tx_pos < cds_end`, so when
+                            // `cds_end < cds_start` every position at or past
+                            // `cds_start` falls through to this arm — making it
+                            // exactly where such a record arrives. The bare
+                            // subtraction then underflowed: a panic in debug, and in
+                            // release a wrap that `as i64` turns into a negative
+                            // `base` flagged `utr3`, returned as success.
+                            //
+                            // `merge.rs`'s `ordered_cds_bounds` refuses the same
+                            // condition on the same ground — an inverted record's
+                            // 5'UTR and 3'UTR halves overlap, so one transcript
+                            // position has two names and any answer here silently
+                            // picks one of them. Refusing does not repair that
+                            // ambiguity; it keeps this conversion from being one
+                            // more place that resolves it arbitrarily.
+                            let last_cds_pos = cds_end.checked_sub(cds_start).ok_or_else(|| {
+                                FerroError::InvalidCoordinates {
+                                    msg: format!(
+                                        "transcript {transcript_id} has inverted CDS bounds \
+                                         (cds_start {cds_start} > cds_end {cds_end}); its c. \
+                                         axis is incoherent, so c.*{utr_offset} cannot be resolved"
+                                    ),
+                                }
+                            })? as i64;
                             Ok((
                                 CdsPos {
                                     base: last_cds_pos,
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1879, opened 2026-08-14
feat(normalize)!: remove the 5' shuffle direction from the public surface

nh13:
## Why

`README.md` rule 6 is the authority here, and it argues against itself today. It states **"There are no user options for normalization form"**, and then, in its very next sentence, concedes that the 3'/5' knob is *not* orthogonal — it selects the frame every other rule is evaluated in, so rules 2 and 3 had to be claimed *per direction* rather than across the two (`README.md:206-208`).

So the knob was a user option for normalization form, sitting inside the rule that forbids user options for normalization form. **Removing it from the public surface is what makes rule 6 true.** An internal oracle is not a user option, so keeping the arm internally costs rule 6 nothing.

Two claims in the tree are also simply false, and are corrected here rather than left for the next reader:

- `src/normalize/config.rs` has said `/// Shuffle towards 5' end (for VCF compatibility)` since the initial commit. **No VCF path has ever selected it.** VCF left-alignment in this crate is the anchor-base rules in `src/vcf/from_hgvs.rs` (#261, closing #81 K2) — which base you attach, not which end you shift to. Related prose in `src/normalize/mod.rs` calling the arm "5'/VCF shuffles" goes with it.
- `README.md:38`'s "3'/5' shifting **per HGVS specification**" describes a rule the spec does not contain.

This references #1857 as context but does not claim it as the mandate. #1857 asks for the 5' arm to be *adjudicated* — exact mirror, mirror with exceptions re-derived, or other — and removal is a fourth answer it does not enumerate. The authority cited is rule 6.

## Why the internal instrument is kept

`ShuffleDirection::FivePrime`, the shuffle machinery, all 7 filename-level 5' test modules, all 9 pinned census constants and the 6-cell corpus matrix are **untouched**.

3'/5' disagreement is a differential oracle over ferro's own *shipped 3' output*, and it has twice caught a defect nothing else could see:

- **#1542 / PR #1840** — `NC_000017.11:g.80110044_80110047delinsGTTGG` normalized to different member counts by direction alone. PR #1840 measured all four `FERRO_PARTITION` arms x both directions: **7 of 8 configurations emitted the merged two-member form; only `live`/3' produced the split.** The shipped default was the outlier, both answers were fixed points, and `FERRO_ASSERT_IDEMPOTENT`, `_REPARSE`, `_IN_BOUNDS`, `_SEQUENCE` and every confluence class were all satisfied by the wrong answer. 5' was the only instrument that saw it.
- `tests/it/cis_confluence_axis.rs` — an off-by-one in `enclosing_exon` let a member on an exon's first base escape the window clamp. A 5'-only *symptom* of a bug in *shared* code; the 3' walk moves away from the exon start and never reached it.

Deleting the arm would not merely make those harder to find, it would make that class of defect structurally invisible. So this PR changes **reachability only** — never behaviour.

The census modules that justified measuring 5' "because `--direction 5prime` is a supported public option" have had that justification rewritten. The measurement is unchanged; the reason is now the oracle.

## What is removed, verified against `origin/main`

| surface | what went |
|---|---|
| CLI | `ferro normalize --direction`, and `cli::parse_shuffle_direction` (+ its `cli::mod` re-export) behind it |
| Python | the ten `direction=` keywords, `parse_direction_or_raise`, `python_helpers::parse_direction`, and the stub sites in `__init__.pyi` |
| Web service | `tools.ferro.shuffle_direction`, its `docs/WEB_SERVICE_SETUP.md` lines, and the value `deploy-local.yml` wrote |
| Docs | `README.md:38`, rule 6's concession, the `from_sequences` options prose, and the subcommand table row |

`ShuffleDirection`, `NormalizeConfig::{shuffle_direction, with_direction, for_entry_point}` and `FromSequencesOptions::{direction, with_direction}` **stay `pub` and become `#[doc(hidden)]`**, documented as an internal instrument. Verified in the rendered docs: after this change `ShuffleDirection` has no rustdoc page and appears **zero** times in `target/doc/ferro_hgvs/index.html` or in `normalize/index.html`.

**That is a deliberate deviation and it is worth being explicit about.** `tests/` is an external integration-test crate, so a `pub(crate)` demotion is not available: **126 call sites across 67 files** under `tests/` and `examples/` drive the differential through `with_direction`, and **82 files** there name `ShuffleDirection` at all (`grep -rn with_direction tests examples` / `grep -rl ShuffleDirection tests examples`, excluding this PR's own new module). Demoting it would mean deleting the oracle this PR exists to protect. `#[doc(hidden)]` removes it from the published API surface while keeping it compilable from `tests/`. The door is closed to documentation and to every entry point; a Rust library caller who goes looking for a hidden item can still find it, and that is the one honest limit of this change.

## Every removal refuses — it does not default

This is the part that matters most, and each was measured rather than assumed. **A silent 3' run for a caller who asked for 5' is data corruption at the customer**, and it is the exact shape a careless removal produces.

**CLI — exit 2, nothing on stdout:**

```
$ ferro normalize NM_000088.3:c.4C>G --direction 5prime
error: unexpected argument '--direction' found
  tip: a similar argument exists: '--reject'
EXIT=2
```

Same for `3prime`, `5'` and `banana`. This is *stronger* than #1866's `value_parser` fix, which is still open and not on `main` — the flag is gone, so there is no value to mis-parse. Note the removed `parse_shuffle_direction` had `_ => ShuffleDirection::ThreePrime` as its fallback arm, which is precisely the #1863 footgun; deleting the value handling without deleting the flag would have preserved it for `--direction 5prime`.

**Python — `TypeError` at the call, on all ten entry points:**

```
normalize(s, direction="5prime")          TypeError: normalize() got an unexpected keyword argument 'direction'
Normalizer(direction="5prime")            TypeError: Normalizer.__new__() got an unexpected keyword argument 'direction'
VariantProjector(direction="5prime")      TypeError: VariantProjector.__new__() got an unexpected keyword argument 'direction'
from_sequences(..., direction="5prime")   TypeError: from_sequences() got an unexpected keyword argument 'direction'
parse(s).normalize(direction="5prime")    TypeError: HgvsVariant.normalize() takes no keyword arguments
```

Measured from a release wheel in a clean venv, the way CI runs it. The last message is a different form because that signature now names *no* keywords at all — a stronger statement, not a weaker one.

**Web service — the load fails and names the key.** `FerroConfig` gains `serde(deny_unknown_fields)`, because serde's default is to *ignore* an unknown field: without it, an operator's existing `shuffle_direction = "5prime"` would be dropped and the service would 3'-shift while the config file still said otherwise.

That attribute promptly caught a real miss in this very PR. `config/service.example.toml` — the file `docs/WEB_SERVICE_SETUP.md` tells operators to `cp` into place — still carried `shuffle_direction = "3prime"`, so the shipped example would have failed to load. **Nothing in the repo parsed that file**, which is precisely how it went stale unnoticed. Fixed, and `service::config::tests::the_shipped_example_config_deserializes` now parses it into `ServiceConfig` so it cannot drift again; the guard was confirmed to fail (`TOML parse error at line 21`) with the key put back.

**Migration.** `direction="3prime"` / `--direction 3prime`: drop the argument, behaviour unchanged. `5prime`: there is no replacement; that form is no longer produced by any entry point.

## Tests: zero suites deleted, net +8 test functions

Counted mechanically over the diff against `origin/main` (`#[test]` / `def test_`, across the 28 touched files): **476 -> 484, net +8**.

- **8 unit-test functions were relocated, not dropped.** `cli::parse::test_parse_shuffle_direction_*` (4) and `python_helpers::test_parse_direction_*` (4) tested two mutually duplicate string parsers that no longer exist. Their assertions are re-pointed at the retained internal `FromStr` impl in `normalize::config` (4 functions, deduplicated). The one that changed meaning is called out in place: `test_parse_shuffle_direction_default` asserted that an unrecognized value resolves to `ThreePrime` — the #1863 behaviour — and is now `test_parse_direction_unrecognized_is_err`.
- **The two behavioural rows that only Python covered were moved to Rust, not lost.** `tests/python/test_variant_projection.py`'s `test_normalize_follows_projector_direction` was the only guard that could tell a re-shuffle from a pass-through (`c.8del` pivots to the 3'-anchored `g.1007del` already, so only the mirror separates them); it is now `issue_867_project_to_genomic_normalized::project_to_genomic_normalized_follows_the_configured_direction`, asserting the same `g.1007del` -> `g.1003del` pair. `test_from_sequences.py`'s 5' row (`g.11del`) is re-pinned in the new module.
- `tests/python/test_issue_1282_position_zero.py`'s 5' reproducers are already covered in full by `tests/it/issue_1282_position_zero.rs` (5 tests, all `FivePrime`); the Python file keeps its real job — the PyO3 boundary, where a panic crosses as `BaseException` and a release wheel wraps silently instead — on the shipped 3' path.
- New `tests/it/five_prime_public_surface_removed.rs` (5 tests) pins the refusals **and** that the internal differential still *acts*, so the oracle cannot be hollowed out into 60 modules comparing an arm against itself.

Instrument intact, measured: `ShuffleDirection::FivePrime` occurrences under `tests/` **229 -> 232**, under `examples/` **8 -> 8**; test modules naming it **59 -> 61**; all 7 filename-level 5' modules present.

**Every new guard was watched failing.** Seven sabotages, each restored with `cp`: re-adding the flag as accepted-and-ignored (2 CLI guards red), dropping `deny_unknown_fields` (service guard red), making `FromStr` guess (parse guard red), no-oping `FromSequencesOptions::with_direction` and `for_entry_point` (both differential guards red), and rebuilding the wheel with `normalize()` accepting-and-ignoring `direction` (16 Python rows red). The suite also caught three assertions I had guessed rather than measured, which is the reason to run it before believing it.

## SemVer

Breaking three independent ways, so this lands at **0.14.0** (`release-plz.toml` sets `features_always_increment_minor`, and the crate is at 0.13.1):

1. Removing a variant from a `pub enum` is breaking regardless of `#[non_exhaustive]` — that attribute forbids downstream *exhaustive matching*, not naming the variant. (Not done here: the variant survives.) What *is* removed is `pub fn parse_shuffle_direction`, which was **doctested**, hence documented API.
2. `python_helpers::parse_direction`, also `pub` and doctested.
3. The Python keywords, which are the sharp edge: Python has no compiler, so this surfaces at the customer's site rather than at their build.

Worth flagging for the release: **no `cargo-semver-checks` runs in CI.** It appears only in the release-plz PR body via `{{ release.semver_check }}`, so the break is reported at release time, not at PR review.

## Gate

- `cargo nextest run --features dev --no-fail-fast` — **10684 passed, 0 failed**, 152 skipped
- `pytest tests/python/` against a release wheel in a clean venv — **981 passed**, 0 failed, 8 skipped
- `generate_tool_support_tables -- --check` — clean (my `README.md` edits are outside the generated blocks)
- Rustdoc: this PR removes one broken intra-doc link of its own making and adds none (134 -> 133 pre-existing errors, all in untouched files; CI does not gate on rustdoc)

**One `--direction` flag deliberately survives**, in `examples/dump_confluence_divergences.rs`. That is a `dev`-gated example generator rather than a shipped entry point — it *is* the instrument, the tool you run to measure the divergence this PR keeps — so removing its flag would remove the ability to use the oracle it protects. `ferro normalize` was the only shipped subcommand that ever had the flag; `project` never did.

Not stacked on #1835; this targets `main` directly.

Refs #1857

Representation-Change: none. The default was already `ThreePrime` on every entry point, so no shipped output moves — the CLI, Python and service paths all constructed a 3' config before this change and construct the same one after it. Nothing under `src/normalize/` changes behaviour: the diff there is the `#[doc(hidden)]` attributes, doc comments, and one `for_entry_point` argument that was already always `ThreePrime` at every seam. The full suite is byte-identical at 10684 passing, including all 9 pinned direction-paired census constants, which would move if any frame had shifted.

### DIFF
--- a/.github/workflows/deploy-local.yml
+++ b/.github/workflows/deploy-local.yml
@@ -129,7 +129,6 @@ jobs:
           [tools.ferro]
           enabled = true
           reference_dir = "/var/lib/ferro-web/reference"
-          shuffle_direction = "3prime"
           error_mode = "lenient"
 
           # Mutalyzer runs as a local Python subprocess for reproducibility and
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ correctness or stability.
 ## Features
 
 - **Full HGVS Parsing**: All coordinate systems (g/c/n/r/p/m/o) and edit types
-- **Variant Normalization**: 3'/5' shifting per HGVS specification
+- **Variant Normalization**: 3' shifting per HGVS specification
 - **High Performance**: ~5M variants/sec single-threaded parsing (>12M/s parallel), zero-copy with nom
 - **Type-Safe**: Leverages Rust's type system for correctness
 
@@ -203,9 +203,14 @@ ferro's normalizer follows four rules about its output, and three about how it h
       feature request, worth making but never a reason to hold a release.
 
 6. **Among multiple conformant forms:** the maintainers choose. **There are no user options for
-   normalization form.** Error mode is an orthogonal axis and stays available. The 3'/5' knob is
-   **not** orthogonal — it selects the frame every rule is evaluated in, so rules 2 and 3 are
-   claimed *per direction* rather than across the two.
+   normalization form.** Error mode is an orthogonal axis and stays available. The 3'/5' shuffle
+   direction was the one exception, and it is now removed from the public surface rather than
+   excused: it is **not** orthogonal — it selects the frame every rule is evaluated in — so it was
+   a user option for normalization form sitting inside this rule. ferro shifts 3', the only
+   direction the HGVS recommendations describe, and no CLI flag, Python keyword or service config
+   key selects otherwise. The 5' arm survives internally as a differential oracle over ferro's own
+   test suite; an instrument is not a user option, and rules 2 and 3 are now claimed once rather
+   than per direction.
 
 7. **Disclosure.** Any change to these rules, and any different choice made under 5 or 6, is
    disclosed: in the changelog before v1, by a major version bump after. Output that *violates*
@@ -350,10 +355,11 @@ description on any machine, against any reference build, with no hidden input. `
 1-based, and `reference` is taken on trust: verifying it would need the reference and would make
 the provider a hidden input, costing exactly the determinism the function exists to provide.
 
-That is five values, not four. The options are not inert — `direction` moves a placement within
-the window and `max_grid_cells` decides whether an answer is produced at all — so the
-"four arguments" this section and the Rust docs both used to claim is withdrawn. Purity is the
-property; the count was wrong.
+That is five values, not four. `max_grid_cells` is not inert — it decides whether an answer is
+produced at all — so the "four arguments" this section and the Rust docs both used to claim is
+withdrawn. Purity is the property; the count was wrong. (`FromSequencesOptions` carries a
+`direction` too, but it is not a caller-facing knob: it is `#[doc(hidden)]`, always 3' on every
+shipped path, and exists for the internal differential oracle described under rule 6.)
 
 ### How to: one canonical description per variant
 
@@ -645,7 +651,7 @@ The `ferro` CLI provides commands beyond parsing and normalization:
 | `prepare` | Download and prepare reference data for normalization |
 | `check` | Verify reference data setup |
 | `parse` | Parse and validate HGVS variants |
-| `normalize` | Normalize HGVS variants (3'/5' shifting) |
+| `normalize` | Normalize HGVS variants (3' shifting) |
 | `explain` | Explain error/warning codes (e.g., `ferro explain W1001`) |
 | `annotate-vcf` | Annotate VCF files with HGVS notation |
 | `vcf-to-hgvs` | Convert VCF records to HGVS |
--- a/config/service.example.toml
+++ b/config/service.example.toml
@@ -18,7 +18,6 @@ max_concurrent_batches = 10
 [tools.ferro]
 enabled = true
 reference_dir = "data/ferro-reference"
-shuffle_direction = "3prime"
 error_mode = "lenient"
 
 # Mutalyzer runs as a local Python subprocess for reproducibility and offline
--- a/docs/WEB_SERVICE_SETUP.md
+++ b/docs/WEB_SERVICE_SETUP.md
@@ -58,7 +58,6 @@ enable_cors = true
 [tools.ferro]
 enabled = true
 reference_dir = "/path/to/ferro-reference"
-shuffle_direction = "3prime"
 error_mode = "lenient"
 
 [tools.mutalyzer]
@@ -99,7 +98,6 @@ Config:
 [tools.ferro]
 enabled = true
 reference_dir = "/path/to/ferro-reference"
-shuffle_direction = "3prime"  # or "5prime"
 error_mode = "lenient"        # or "strict", "silent"
 ```
 
--- a/python/ferro_hgvs/__init__.pyi
+++ b/python/ferro_hgvs/__init__.pyi
@@ -31,12 +31,11 @@ def parse(hgvs_string: str) -> HgvsVariant:
     """
     ...
 
-def normalize(hgvs_string: str, direction: str = "3prime") -> str:
+def normalize(hgvs_string: str) -> str:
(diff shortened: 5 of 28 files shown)


nh13:
@coderabbitai pause

PR #1880, opened 2026-08-14
test(rulings): check that a record's guard citation resolves

nh13:
A `decided` record and the guard that enforces its ruling are connected by nothing but prose. Rename the guard and the record goes on naming the old one — the ledger still reads as enforced, and nothing anywhere fails. This adds the check that closes that, and reports what it found.

Representation-Change: none. Test-only; no file under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched, and the ledger itself is unchanged (28 records in, 28 out, no status moved).

## Census first — and the good news is that most of this is already in place

Measured over the ledger at `origin/main` `d4552167`: **28 records, 25 `decided`, 3 `undecided`.**

**There is no structured guard field.** `equivalence_classes` is the only structured pointer a record carries, on 4 decided records, and it is already enforced end to end — `generate_spec_fixture` bails on an unknown class id, and `spec_equivalence_classes_converge` asserts and guards its own vacuity. Nothing needed building there, and no schema was invented; per the brief's rule for the prose-only case, the ledger was not edited and no field was added.

Every other guard reference is prose. Of the **14** `::`-qualified guard citations across the ledger:

| | count |
|---|---|
| resolve to a test function in the tree | 12 |
| ...that assert | 12 / 12 |
| ...manifest-gated or skip-on-absent | 0 |
| ...`#[ignore]`d | 1 — the declared #1630 deferral |
| **do not resolve at all** | **2, both in `decided` records** |

So the honest headline is: the guards the ledger names are real, they run, and they assert. The gap is that **nothing checked the citations stay valid**, and two have already rotted.

## What the check flags, with a verdict on each

Both are genuinely unenforced, and neither is a typo — **both renames inverted the claim**, which is why neither is fixable mechanically:

| record | cites | the tree has now | verdict |
|---|---|---|---|
| `alignment-only-symbol-in-a-description` | `…_refused_for_dash_and_accepted_for_x` | `…_refused_in_every_mode_for_both_x_and_dash` | Genuinely unenforced. The guard was renamed *when the ruling was implemented* (#1684), so the record cites the pre-fix name — which describes the behaviour the ruling **overturned**. A citation repair, but one that changes which proposition the sentence points at. |
| `derivation-may-not-be-bounded-by-the-inputs-spelling` | `a_dup_flush_against_a_del_is_re_derived_into_the_span` | `a_dup_flush_against_a_del_is_left_alone` | Genuinely unenforced, and **substantive**. The record cites that row as reading "AGAINST THE BOUND rather than for it" *because* it was re-derived into the span; the guard standing in its place asserts it is left alone. Adjudication, not a rename. |

Neither was exempted. `RETIRED_GUARD_NAMES` — the declared-exemption hatch, modelled on how `Representation-Change: none` works, where declining is first-class and what is rejected is silence — ships **empty** on purpose: using it here would be silencing the finding rather than declaring an intent. `no_retired_guard_name_is_also_a_live_guard` stops a future entry masking a guard that does exist.

## Why it lands `#[ignore]`d

Because the alternatives are worse: landing it armed lands a red suite, and exempting the two rows weakens the check to green. Fixing either record is an adjudication I should not make unilaterally. This is the idiom `corpus_prohibited_inputs`'s own `the_decided_target_is_a_mode_gated_refusal` already uses — `#[ignore]`d, asserting the decided answer, with an issue as its acceptance criterion. The census and both rows are in the check's doc comment, so the next reader gets the finding without re-deriving it.

**Un-ignoring it is the acceptance criterion** once the two citations are settled.

## Design notes

- **Built on the existing machinery.** It lives in `ruling_citation_currency.rs`, reads the ledger through `common::rulings`, and forks no second parser. The walker is refactored to one definition (`source_files`) because the new scan brace-matches function bodies and needs each file whole.
- **Recognition is by shape, not by resolution** — the same move `looks_like_a_record_id` makes. A grammar that asked "is this token a test that exists" would be vacuous in exactly the case it exists for: delete the test and the token stops matching, so nothing is reported. The sentence-shape grammar selects 1,649 of 9,652 test functions and only 11 of 7,689 non-test source functions, none of which any record cites `::`-qualified.
- **Only the `::`-qualified form is read.** Rationales do use bare backticked names, but so does ordinary prose about the code; telling them apart is the judgement call this file is built to avoid. All 14 real citations are qualified.
- **The assertion detector reads helper calls, not just macros.** A stricter matcher reported a guard whose entire body is one `assert_normalizes_preserving_in` call as assertionless — a false positive on the exact #1858 shape the check is for. Pinned both ways by `the_assertion_detector_reads_helper_calls_as_assertions`.
- **Three vacuity guards**, because a check that reads nothing passes forever: a citation-count floor, a guard-scan size floor, and an assertion that both the `#[ignore]`d and the asserting properties are observable at least once — otherwise those arms are dead code.

## Verification

`cargo fmt --all --check` clean; `cargo clippy --all-features --all-targets -- -D warnings` clean; spec fixture regenerated; `cargo nextest run --features dev --lib --test it` — **10477 passed, 0 failed**. Red-proof, with the check un-ignored:

```
2 guard citation(s) in tests/fixtures/grammar/hgvs_spec_normalization_overrides.json name a guard that cannot enforce the record:
  `alignment-only-symbol-in-a-description` (decided) names `an_alignment_only_symbol_is_refused_for_dash_and_accepted_for_x`, which is defined nowhere under ["src", "tests"]
  `derivation-may-not-be-bounded-by-the-inputs-spelling` (decided) names `a_dup_flush_against_a_del_is_re_derived_into_the_span`, which is defined nowhere under ["src", "tests"]
```

### DIFF
--- a/tests/it/ruling_citation_currency.rs
+++ b/tests/it/ruling_citation_currency.rs
@@ -93,6 +93,20 @@
 //! extractor ([`record_ids_named_in`]) because a rationale cites its siblings by
 //! bare backticked id rather than with the word "ruling".
 //!
+//! # The ledger's citations of its own GUARDS
+//!
+//! Everything above is about record **ids**. A rationale also cites the test
+//! that enforces its ruling — "Pinned by `defect_371_transcript_exit::the_exit_follows_…`"
+//! — and until [`every_ruling_to_guard_citation_resolves`] nothing checked those
+//! at all. Rename the guard and the record goes on naming the old one, reading
+//! as enforced while enforcing nothing.
+//!
+//! That check is `#[ignore]`d because it is **red on two `decided` records** on
+//! the base it landed on, and both are real rather than fixable by editing a
+//! name: each rename inverted the claim, so substituting the new name would
+//! change what the record asserts. Its doc comment carries the census and the
+//! two rows.
+//!
 //! # What is deliberately NOT checked: whether a quoted sentence is really the
 //! # record's
 //!
@@ -267,8 +281,32 @@ struct SourceLine {
 
 /// Every line of every `.rs` file under [`SCAN_ROOTS`], excluding this file.
 fn source_lines() -> Vec<SourceLine> {
+    source_files()
+        .into_iter()
+        .flat_map(|(path, text)| {
+            text.lines()
+                .enumerate()
+                .map(|(index, line)| SourceLine {
+                    path: path.clone(),
+                    number: index + 1,
+                    text: line.to_string(),
+                })
+                .collect::<Vec<_>>()
+        })
+        .collect()
+}
+
+/// Every `.rs` file under [`SCAN_ROOTS`] as `(scan-relative path, text)`,
+/// excluding this file.
+///
+/// Split out from [`source_lines`] because [`guard_sites`] needs each file
+/// *whole*: it brace-matches function bodies, and a flattened line list has
+/// lost the file boundaries that makes possible. One walker rather than two, so
+/// "what the scan reads" has a single definition — the same reasoning
+/// `common::rulings` gives for parsing the ledger in one place.
+fn source_files() -> Vec<(String, String)> {
     let root = crate_root();
-    let mut lines = Vec::new();
+    let mut files = Vec::new();
     for scan_root in SCAN_ROOTS {
         let dir = root.join(scan_root);
         assert!(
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

Issue #923, opened 2026-07-04
feat(normalize): synthesize/resolve the transcript selector on bare-NG_ c. input (refseq.md SHOULD)


nh13:
Focused live successor to the closed #500 (`feat(normalize): resolve legacy gene-symbol selectors to transcript accessions on genomic references`). Re-homed by the #912 stale-tracker hygiene sweep. This is an `improvement` (ferro's output is valid HGVS but not spec-*preferred*), not a bug.

On a bare-`NG_` `c.` input (no transcript selector), ferro emits no transcript selector on output; on an `NG_` input carrying a non-transcript selector (gene symbol / gene-symbol-version / numeric gene ID), ferro preserves that selector rather than resolving it to the annotated transcript accession. HGVS `refseq.md` L138-139 says the transcript used **should** be indicated (mutalyzer synthesizes the `(NM_…)` selector).

## Rows (4, cluster `refseqgene-selector`, `improvement` on `normalized`)

`NG_008939.1:c.155_157delAAC`, `NG_008939.1:c.155_157del3`, `NG_008939.1:c.274_275inv`, `NG_012337.1(10683):c.274G>T`.

Tracked as `improvement{axis: normalized, section: HGVS §RefSeqGene transcript selection}`. XPASS-guarded: converges when ferro synthesizes/resolves the transcript selector. Part of the #326 parity burn-down.

nh13:
Scoping finding (rows 1-3 vs row 4) after implementing + validating the fix.

**Rows 1-3 (bare `NG_008939.1:c.…`) — the synthesis works and is spec-correct**, but landing it in isolation is **not** contained; it must be coordinated with a protein-projection fix (below). Implementation that was validated:
- `NgHostedTranscripts::sole_hosted(ng)` — returns the transcript iff the NG_ hosts exactly one gene with one transcript (multi-gene `NG_`/many-transcript `NC_` decline).
- provider trait `sole_hosted_transcript(ng_parent)` (default `None`) — **must be forwarded in the `Arc<T>`/`Box<T>` blanket impls and the conformance `ArcProvider` newtype**, else it silently falls to the default when wrapped (this bit during validation).
- one branch in `Normalizer::rewrite_legacy_gene_selector`: when `gene_symbol` is `None` on a bare `NG_`/`LRG_` `c.` input, call `sole_hosted_transcript` and attach the `NM_` via `with_genomic_context` (the same path the gene-symbol case already uses).

[...]

PR #1882, opened 2026-08-14
fix(normalize): refuse a c. description whose reference resolves no CDS

### LINKED ISSUE #1870: Open question: a c. description against a transcript whose CDS the reference cannot resolve normalizes silently — measured 1,242 NM_ and ~52% of XM_
## The question

When the reference serves a transcript's **bases** but cannot resolve its **CDS**, a `c.` description against it is currently **normalized silently**: the input comes back as the output, `status: ok`, `changed: false`, `warnings: []`, exit 0, byte-identically in strict, lenient and silent. The `n.` axis on the *same accession, same base, same edit* shifts correctly.

Should that be a refusal, or is silence correct? This is filed as a question with a measurement attached, not as a bug report — I have not established that silent normalization is wrong, only that it is **undeclared**: no ruling record and no `README.md` rule reaches it, and ferro's own projection surface answers the opposite way for the identical input.

Measured against `origin/main` at `2b979da2` (current tip) with a release `ferro` built from it, and the prepared reference at identity `aa8b3246d83055cc`.

## First, a correction that removes half of what prompted this

This came out of #1858 / #1867, which report `NM_033517` as served with no cdot record, `cds_start: None`, zero exons, and `c.1324del` as a passthrough rather than a shift decision. **That is stale.** The cdot artifact always carried a full 22-exon record for `NM_033517.1` (`"start_codon": 0`, `"stop_codon": 5157`, `"partial": 1`); the provider was discarding it, which is #1722, fixed by #1724 — merged today and the literal tip commit. At `2b979da2`:

```
NM_033517.1:c.1324del   -> NM_033517.1:c.1326del    changed=true   ok
NM_033517.1:c.125del    -> NM_033517.1:c.127del     changed=true   ok
NM_033517.1:c.999999del -> normalize_error: past the cds-end (5157)   (W4004)
```

ferro now agrees with Mutalyzer on `c.1324del → c.1326del`. `2b979da2` is **not** an ancestor of #1867's head, so its measurement predates the fix. **So "should `ferro prepare` serve a cdot record for that accession" is already answered and already closed** — it was never a prepare gap; the record was in the artifact, and the fault was a provider fallback. Nothing further is needed there. (The degraded supplemental artifact that made the fallback harmful is #1728, still open.)

Relatedly: `cds_start.unwrap_or(1)` exists at exactly five sites, all in `tests/it/real_data_normalization_tests.rs`. **There is no such site in `src/`.** `src/` does something adjacent — it abandons the axis rather than fabricating an origin, which is what the rest of this issue is about.

## The behaviour

`NM_000703.1` is 3042 bases, served from `supplemental/patterns_transcripts.fna`, with a `CCCC` run at 1-based transcript position 227. `ferro normalize --reference <dir> -f tsv`, default `--error-mode strict`, `EXIT=0`:

```
line  input                        normalized                 changed  status  detail
1     NM_000703.1:n.227del         NM_000703.1:n.230del       true     ok
2     NM_000703.1:c.227del         NM_000703.1:c.227del       false    ok
3     NM_000247.3:n.240del         NM_000247.3:n.243del       true     ok
4     NM_000247.3:c.240del         NM_000247.3:c.240del       false    ok
5     NM_000703.1:c.-5del          NM_000703.1:c.-5del        false    ok
6     NM_000703.1:c.*5del          NM_000703.1:c.*5del        false    ok
7     NM_000703.1:c.227dup         NM_000703.1:c.227dup       false    ok
8     NM_000703.1:n.227dup         NM_000703.1:n.230dup       true     ok
```

`delins` and `ins` behave the same (`c.227delinsGGG`, `c.227_228insTTT` — both unchanged, `ok`).

**Silence, not leniency.** All three modes, `EXIT=0` each, `md5` `f462d356b5a4d7157bce50e818e63e61` × 3:

```json
{"input": "NM_000703.1:c.227del", "output": "NM_000703.1:c.227del",
 "status": "ok", "corrections": [], "warnings": []}
```

**The bounds check goes with it.** A `c.` position that names no base at all is accepted:

```
NM_000703.1:c.999999del  -> NM_000703.1:c.999999del   false  ok     <- 3042-base transcript
NM_000703.1:n.999999del  -> normalize_error: past the transcript-end (3042)  W4004
NM_000016.3:c.999999del  -> NM_000016.3:c.999999del   false  ok
NM_000016.3:n.999999del  -> normalize_error: past the transcript-end (2454)  W4004
NM_000016.6:c.999999del  -> normalize_error: past the cds-end (1266)         W4004
```

So the missing bound is **CDS-specific**, not a general reference gap — the same accession bounds its `n.` axis correctly. `check_cds_pos_past_end`'s documented "conservative skip" on missing metadata (`src/normalize/mod.rs:185-194`) is the mechanism. `NM_000016.3` vs `.6` also shows that **version drift is one route into the class**: cdot carries `.6`, and `.3` is served with bases from an older RefSeq release.

**The projection surface declines the identical input.**

```
$ ferro project 'NM_000703.1:c.227del' --axis n --reference $REF     # EXIT=0
# NM_000703.1:c.227del [n]: unavailable: no n. representation for this variant

$ ferro project 'NM_033517.1:c.1324del' --axis n --reference $REF    # EXIT=0
NM_033517.1:n.1326del
```

The library message behind that is `"Transcript has no CDS"` — `src/convert/coding.rs:22`, `src/convert/mapper.rs:106,146`, `src/convert/protein.rs:22`. One binary, one input, one reference: `project` refuses, `normalize` answers.

**And the parser already refuses this exact fact, by prefix:**

```
NR_000005.1:c.100del  -> parse_error: the `c.` coordinate system is not valid on a non-coding
                         RNA (NR_/XR_) reference, which has no CDS and is not genomic;
                         use `n.` (non-coding) or `r.` (RNA)
```

`src/hgvs/parser/variant.rs:7332-7345` decides "there is no CDS here" from the accession prefix. An `NM_` whose CDS the reference cannot resolve gets the opposite treatment from the same underlying fact.

## The code

`src/normalize/mod.rs:6139-6142`:

```rust
let cds_start = match transcript.cds_start {
    Some(s) => s,
    None => return Ok((HV::Cds(self.canonicalize_cds_variant(variant)), vec![])),
};
```

For contrast, the repair passes one module over already refuse the same condition, deliberately and with a doc comment saying why — `src/normalize/merge.rs:4261`, `Region::Cds => Some(cds_start()? - 1)`: *"`Cds` above refuses instead … because a `c.` description asserts a CDS this record does not have."* The two halves of the normalizer disagree with each other.

## Scope

**Method (measured, through ferro itself).** Oracle: probe `<acc>:c.999999del`. A transcript whose CDS ferro resolves refuses with W4004 and names its `cds-end`; one whose CDS it cannot resolve returns the input, `ok`, `changed=false`. Validated on four known cases before use. Run over **every** served `NM_` and a seeded random sample of the served `XM_`, in one `ferro normalize -j 8` invocation (`summary: total=103773 changed=0 unchanged=4380 failed=99393`).

| prefix | probed | silent passthrough | CDS resolved (W4004) |
|---|---:|---:|---:|
| `NM_` | 97,773 (all served) | **1,242 (1.27%)** | 96,531 (98.73%) |
| `XM_` | 6,000 of 127,889 served | **3,138 (52.30%)** | 2,862 (47.70%) |

Extrapolating `XM_`: ≈ 66,900 of 127,889. **Total ≈ 68,100 of 225,662 served `c.`-addressable accessions (~30%)**, of which **1,242 are curated `NM_`**.

A static cross-check agrees to within 1%: enumerating the accession set of every artifact the manifest names as a source of transcript bases (16 `transcripts/human.*.rna.fna.gz` + supplemental + backfill + legacy GenBank = 333,594 accessions) against the immediate keys of the top-level `transcripts` object of `cdot-0.2.32.refseq.GRCh38.json` (482,519 records, 355,591 with a CDS resolvable per `src/data/cdot.rs:613`) gives 4,296 `NM_` and 66,335 `XM_` with no record for that exact version. The static figure is an upper bound — it ignores version-flexible resolution, the supplemental CDS map and the deferred GRCh37/Ensembl cdots, which is why the measured `NM_` number is 3.5× smaller. `NR_`/`XR_` (76,061 accessions) are legitimately CDS-less and already refused at parse; `NG_` is genomic and out of scope.

Of the 4,296 static `NM_`, 3,611 have another version of the same base accession in cdot and 685 are absent entirely — so both "stale version" and "never annotated" are represented.

Sample of the measured `NM_` class: `NM_000016.3`, `NM_000021.2`, `NM_000027.2`, `NM_000033.2`, `NM_000038.3`, `NM_000038.4`, `NM_000043.3`, `NM_000050.3`, `NM_000051.1`, `NM_000051.2`.

**What the oracle does not separate:** a record carrying `cds_start` but no `cds_end` would also register, since the plain `c.<N>` bound needs the end. The `n.`-shifts / `c.`-inert contrast was checked by hand only on `NM_000703.1`, `NM_000247.3` and `NM_000016.3`.

## Is it already ruled on?

No. The ledger has 28 records (25 `decided`); every record's full JSON was searched case-insensitively for `no cds`, `cds_start`, `cds start`, `coding sequence`, `refus`, `declin`, `silent`, `pass-through`, `passthrough`, `unserved`, `unresolv`, `absent`, `missing`, `non-coding`, `cds-less`, `CDS`. The three nearest do not reach it:

- **`c-and-n-positions-are-flat-transcript-offsets`** (decided) is *about* `NM_033517.1`, but decides flat-offset vs exon-walk and **presupposes** a resolvable origin — "`c.N` is `cds_start + N - 1` on the sequence a provider serves".
- **`absolute-prohibition-enforcement-stage`** (decided) is the closest in shape, and its rationale contains the sentence this is an instance of — *"Normalization is not impossible here, it is VACUOUS, which is worse, because the output looks normalized."* But its subject is **input conformance**, and its ruling is about the parse stage and the mode gate. `NM_000703.1:c.227del` is a perfectly conformant input; what fails is the reference's ability to resolve it.
- **`adjudication-precedence-order`** governs how to settle it, not what the answer is.

`README.md` has no rule about an unresolvable coordinate axis. Two things bear on it without settling it: rule 1 (`:176`, *"Conformant … Absolute — never traded"*, which #1627 argues has no mode escape and which the `c.999999del` output arguably violates); and `:412`, *"The policy is to refuse rather than quietly answer with a weaker rule"* — stated for the `from_sequences` surface, so a precedent rather than a governing clause.

## What evidence would settle it

**Q1 — is refusal or silence correct?** This is a rule-6 maintainer choice unless it can be brought under rule 1. Two sub-questions that can be answered with evidence rather than judgment:

- Does the `c.999999del` output violate rule 1 as *output*? If yes, #1627's argument applies unchanged and that half is a bug, not a choice — the description names no base in the transcript, and `to_spdi` / `denotation_of` should be asked whether they call it inexpressible (I did not check those two on this input).
- Does any *conformant* `c.` output ever change? Today no `c.` description on these accessions ever moves, so the refusal costs no currently-correct answer. The `Representation-Change:` cost is one-directional: ~68,100 accessions move from *answered* to *declined*, the mirror of what #1724 just disclosed in the other direction.

**Q2 — if silence stays, must there be a warning?** `warnings: []` is what makes this indistinguishable from a genuine fixed point. A `NormalizationWarning` is separable from the refuse/answer choice and would settle the "undeclared" half on its own.

**Q3 — where should the check live?** `src/normalize/mod.rs:6139` (per-call, mode-gateable via `ErrorConfig::action_for`, following `apply_bracket_cardinality_rule`'s precedent) or at parse, generalising `variant.rs:7345` from "prefix says non-coding" to "this reference resolves no CDS". The second cannot be done at parse without a provider, which is probably decisive.

## Related, and why none of it covers this

- **#1722 / #1724** — closed today; the reference half of the original report. **Do not re-file it.**
- **#1728** — the degraded `patterns_transcripts.metadata.json` behind it (19 of 140,027 rows carry a CDS).
- **#1627** — *"a member ferro cannot compute is re-emitted verbatim"* — the same shape, scoped to inputs that denote nothing (`ins6`, `delinsX`), not to a conformant input on an unannotated reference. Its argument reaches the bounds half here, not the shift half.
- **#1630** — implements `absolute-prohibition-enforcement-stage`; input conformance.
- **#1798** — `parse_hgvs` applies no `ErrorConfig`. Covers which mode judges an input's *spelling*; ours is fine. Deliberately deferred past the next release.
- **#1745 / #1746** — `cds_start` fabrication siblings (`get_amino_acid_at` underflow; `cds_to_tx` answering `c.?` with `cds_start`). Same family, different sites. One more for that family found while measuring: `src/data/mapping.rs:696-698` does `cds_end.unwrap_or(0) - cds_start.unwrap_or(0)`, which underflows `u64` when `cds_end` is absent and `cds_start` is not.
- **#1858 / #1867** — the assertionless guards this came out of. #1867's `manifest_gate::unserved` promotion for `NM_033517.1` will no longer fire on this reference once it rebases past #1724.

Working notes, scripts and full outputs are outside the repo.

nh13:
The operator ruled on the open question in #1870: **refuse.** This records the ruling in the ledger and implements it.

Closes #1870

Representation-Change: 67,576 of 225,662 served `c.`-addressable accessions (~30%) move from answered to REFUSED — 1,242 curated `NM_` and 66,334 `XM_` predicted models, measured through ferro over the prepared GRCh38 reference, not extrapolated. These inputs were previously ACCEPTED (they returned a value, `status: ok`, `changed: false`, `warnings: []`, in every mode), so this is a real migration for any consumer keying on the normalized string. The sharpest consumer-facing fact is deduplication: on an affected accession `normalize` was the IDENTITY function, so all five of `NM_000546.3:c.528del`..`c.532del` — one deletion from one `CCCCC` run, one variant — came back as five distinct "normalized" strings and a caller deduplicating by normalized string got five variants where there is one. The clinically weighty `NM_` rows are overwhelmingly SUPERSEDED transcript versions (601 of 1,242 have a higher served version that resolves a CDS), i.e. a legacy-input exposure — the version pinned in an older report, a database export or a lab's calling config, not the version a fresh pipeline picks. 27 of the ACMG SF v3.2 rows named in #1870 reproduce as refused, `TP53 NM_000546.3` and `BRCA1 NM_007294.2` among them; that is a legacy-input exposure and not "ferro silently mis-normalizes TP53". Nothing moves on an accession whose CDS the reference resolves. Committed-corpus movement: 10 rows on the spec-enumeration corpus (all `NM_004006.1`, divergence budget unchanged) and 6 rows on the real-corpus multi-member cis census (unwindowed -> declined, three transcripts).

## The question, and the ruling

When the reference serves a transcript's **bases** but resolves no **CDS**, a `c.` description against it was **normalized silently** — the input returned as its own output, `status: ok`, `changed: false`, `warnings: []`, byte-identically in strict, lenient and silent. The `n.` axis on the *same accession, same base, same edit* shifted correctly.

**Refuse.** The governing principle, in the operator's framing: *where ferro lacks the information or the authority to choose, it must refuse visibly rather than pick.* That is not imported — `canonical-form-choice-when-both-legal` says of the descriptions no normalizer can recover the provenance of that ferro "cannot implement them and does not try", and #1284 ruled that "the repair must refuse rather than pick one".

## The ruling record

`rulings[c-description-against-an-unresolvable-cds-is-refused]`, `decided`, governing `background/numbering.md:21`, with `refseq.md:129` as second authority and `numbering.md:52` / `:44` explicitly labelled **supporting context, not authority**.

It says plainly that **the spec does not settle this**: `:21` says what a `c.` number counts from and does not say what a tool must do when it cannot find the codon; `:129` says what a coding DNA reference sequence is and does not address a tool holding a record that falls short of it. Returning the input untouched emits a conformant string and violates no clause; refusing emits nothing and violates no clause. Both are conformant behaviours, so this is a **maintainer choice under rule 6** of the `README.md` ruleset, disclosed under rule 7 — and the record says it must never be cited as conformance. No clause citation was manufactured.

### Ledger safety

Run before and after, per `CLAUDE.md`: **28 records / 25 decided / 3 undecided -> 29 / 26 / 3.** Nothing dropped from `origin/main`, no existing record's status, question, governing clause, clause list or rationale touched, and the non-`rulings` sections byte-identical. Census pins moved deliberately:

| pin | before | after |
|---|---|---|
| `clause_ruling_index::the_index_is_not_vacuous` | `(28, 25, 3)` | `(29, 26, 3)` |
| `clause_ruling_index::clauses_named_by_several_records_are_flagged` — `multi.len()` | 30 | 33 |
| ...`mixed.len()` | 3 | 3 (unchanged) |
| `clause_ruling_index` rendered index block | 111 clause lines | 112 |
| `RULING_STATUSES` | 28 | 29 |
| `CLAUDE.md` decided-rulings table | 25 | 26 |

The rendered index block was captured programmatically from the test's own output and spliced on its markers — not retyped.

## Which stage it refuses at, and why

**At normalize, not at parse.** `absolute-prohibition-enforcement-stage` rules that **input conformance** is answered at parse, before the input is accepted. This is not input conformance. `NM_000546.3:c.528del` is a perfectly conformant description — correct prefix, correct axis for an `NM_`, a position and an edit the grammar admits, and the spelling the spec's own worked examples use. Nothing about the string is wrong. What fails is the **reference's** ability to resolve it, which is a fact about the provider: the identical input against a reference carrying the CDS normalizes correctly, and `NM_000546.4`, `.5` and `.6` all do so today. A fact that changes with the reference cannot be decided at parse — and decisively, `parse_hgvs` holds no provider, so parse cannot see it at all. The stage follows from what the question is, not from a preference.

**Unconditional across strict / lenient / silent**, which adds no mode policy but applies the existing one. `absolute-prohibition-enforcement-stage` states lenient's contract as *"input conformance is NOT validated. It fails only if it cannot NORMALIZE, which can happen"*, and silent's as *"lenient with no error messages, but an exit code"*. This is exactly the antecedent of lenient's own contract: ferro genuinely cannot normalize, so lenient's stated behaviour is to fail.

## The error, matched to the precedent rather than invented

`ferro project 'NM_000546.3:c.528del' --axis c` **already** exited 1 with `Coordinate conversion error: transcript NM_000546.3 has no CDS start`, while `ferro normalize` on the identical input returned `status: ok` and exit 0. The refusal reuses that error variant (`FerroError::ConversionError`, code **E5001** `ConversionFailed`) and its words:

```
Coordinate conversion error: transcript NM_000546.3 has no CDS start; a c. description
cannot be normalized against a reference that resolves no CDS — use the non-coding (n.)
or genomic (g.) representation instead
```

No new `ErrorType` and no new W-code: there is no mode gate to configure, and inventing a code would let the two surfaces drift apart again. The parser's existing prefix rule (`c.` on `NR_`/`XR_` "which has no CDS") is untouched and still refuses at parse in every mode — correctly, because an `NR_` carries the fact in its own accession and needs no reference. What this change does is make the refusal follow **the fact** rather than the naming convention.

## Scope: a transcript reference

A bare genomic parent — `NG_`/`NC_`/`LRG_` with nothing in its transcript-selector slot — is **not** a transcript. Its `c.` description names an *implied* transcript, and resolving that is the legacy-gene-selector machinery's job (#500 / #793 / #923), not this ruling's.

This boundary was measured, not assumed. Unscoped, the check newly failed `NG_012337.1(10683):c.274G>T` in the mutalyzer comparator — a row whose accepted value is ferro's own pass-through — and declined six real-corpus cis rows for the same reason: when the selector resolution declines, the accession still reads as the parent, so the lookup lands on the `NG_` record, which carries no CDS *by construction*. Scoped, both come back byte-identical to `origin/main`, and the six declines that remain are genuine (`NM_000500.5`, `NM_001040075.1`, `NM_006894.4` — each independently confirmed present in the measured CDS-less class).

## The `n.` axis still shifts — measured on the same accessions

```
NM_000546.3:c.528del   ->  normalize_error: transcript NM_000546.3 has no CDS start
NM_000546.3:n.779del   ->  NM_000546.3:n.783del   changed=true   ok
NM_000703.1:c.227del   ->  normalize_error: transcript NM_000703.1 has no CDS start
NM_000703.1:n.227del   ->  NM_000703.1:n.230del   changed=true   ok
NM_000546.6:c.528del   ->  NM_000546.6:c.532del   changed=true   ok      # CDS resolves: untouched
NM_000546.3:n.999999del -> normalize_error: past the transcript-end (2640)  (W4004)
```

`n.779` on `NM_000546.3` **is** `c.528`. The bases are present and correct; only the CDS annotation is missing, which is exactly why the two axes must part company here.

## The counterweight, recorded rather than argued away

`refseq.md:145` publishes **`NM_004006.1:c.5697del`** as its own worked minus-strand 3'-rule example, and `NM_004006.1` is a version whose CDS ferro's committed reference slice does not resolve — so a spec example is in the moving set. Two things keep it from defeating the ruling. The refusal is a statement about the **reference ferro holds**, never about the description: served a reference carrying the CDS, the same string normalizes. And `project` **already** refused that row, with the same words, before this change — what moves is `normalize` joining `project`. It is in the record and in `PASSING_CENSUS`'s own comment.

## Verification

- `cargo nextest run --features dev --no-fail-fast`: **10688 passed, 0 failed** (`EXIT=0`, read back from the log).
- Same, with `FERRO_MANIFEST` set against the prepared reference: **10688 passed, 0 failed** (`EXIT=0`), **zero** `Skipping:` / `skipping — no reference manifest` lines.
- Every new guard was watched fail first, by perturb-and-restore on `src/normalize/mod.rs` with `cp` and an `md5` equality check on the restore — never `git checkout --`. Disabling the refusal reddens the three refusal rows and leaves the control, `n.`-axis, scope and parse rows green; disabling the *scope* reddens exactly `a_bare_genomic_parent_is_out_of_scope_and_still_answers`.
- The two reference-aware failures the first attempt produced were separated from pre-existing drift by building `origin/main` `d4552167` in a detached worktree and running the same two tests there: both **passed** at baseline, so both were mine, and both are addressed above (one by the scope, one by a deliberate pin move).

### Scope figures, re-derived

Served accessions were enumerated from the reference's own `.fai` indexes (333,594 total; 97,773 `NM_`, 127,889 `XM_`), and every one probed with `<acc>:c.999999del` through a release `ferro` built from this branch. The oracle is now direct rather than inferential — the class refuses by name:

| prefix | probed | refused (`has no CDS start`) | CDS resolved (W4004) |
|---|---:|---:|---:|
| `NM_` | 97,773 (all served) | **1,242 (1.27%)** | 96,531 |
| `XM_` | 127,889 (all served, no sampling) | **66,334 (51.87%)** | 61,555 |

**67,576 of 225,662 (29.95%).** The `NM_` figure reproduces #1870's exactly; the `XM_` figure replaces its 6,000-row extrapolation with the full census and lands within 0.003% of the issue's static cross-check (66,335).

### One Python test moved with it

`tests/python/test_loaders.py::test_normalizer_from_manifest` proved that `Normalizer.from_manifest` wires a working normalizer by normalizing `NM_TEST.1:c.1A>G` against `manifest_tiny` — a FASTA and **no cdot**, i.e. exactly a record that serves bases and resolves no CDS. It now normalizes on the `n.` axis, which needs no CDS and is what the refusal message tells callers to reach for, so it stays a test of the loader plumbing; a sibling row asserts the `c.` axis is refused and that the bindings surface it as a typed `NormalizationError` rather than swallowing it. `pytest tests/python/`: **958 passed** locally against a `maturin develop` build. That failure was caught by CI rather than by the local gate, which is worth recording: `cargo nextest` does not reach the bindings, so a behaviour change under `src/normalize/` needs the Python suite run explicitly even when no binding code is touched.

### Coverage note

The `multi_member_cis_axis` census change (`declined: 0 -> 6`, `unwindowed: 96 -> 90`) is reference-aware: it resolves a prepared reference from `FERRO_MANIFEST`, which PR CI does not provision, so that module skips silently there and only the nightly exercises it. It is therefore a measurement, not pre-merge coverage. The equivalent behaviour is pinned hermetically by `issue_1870_cds_less_transcript_refusal::a_multi_member_allele_on_the_same_record_is_refused_whole`, which PR CI does run.


nh13:
@coderabbitai pause

PR #1883, opened 2026-08-14
feat(rulings): make a record state what enforces it, and reject silence

### LINKED ISSUE #1881: Two decided records cite a guard that does not exist
Two `decided` records in `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json` cite a guard as the thing that enforces their ruling, and **that guard exists nowhere under `src/` or `tests/`**. Both were found by the check added in #1880, which is `#[ignore]`d precisely because of them; settling these two is its acceptance criterion.

Neither is a typo, and neither is fixable by substituting the current name — **both renames inverted the claim**, so swapping the name in would silently change what the record asserts.

## 1. `alignment-only-symbol-in-a-description`

Cites `tests/it/corpus_prohibited_inputs.rs::an_alignment_only_symbol_is_refused_for_dash_and_accepted_for_x`, in its SCOPE paragraph ("Pinned by …").

The tree has `an_alignment_only_symbol_is_refused_in_every_mode_for_both_x_and_dash` (`tests/it/corpus_prohibited_inputs.rs:670`). The guard was renamed when the ruling was implemented (#1684) — so the record cites the **pre-fix** name, which describes the behaviour the ruling *overturned*: `X` accepted. The record rules that neither `X` nor `-` may appear.

Likely a straight citation repair, but it should be made deliberately, since the cited name and the ruling currently disagree about what happens to `X`.

## 2. `derivation-may-not-be-bounded-by-the-inputs-spelling`

Cites `cis_confluence_adjudication::a_dup_flush_against_a_del_is_re_derived_into_the_span`, introduced with the record in #1733.

The tree has `a_dup_flush_against_a_del_is_left_alone` (`tests/it/cis_confluence_adjudication.rs:490`) — the **opposite outcome**.

This one is substantive. The record's sentence reads:

> ONE ROW READS AGAINST THE BOUND RATHER THAN FOR IT. In `cis_confluence_adjudication::a_dup_flush_against_a_del_is_re_derived_into_the_span` the PRE-removal form is `c.[9T>A;16_23dup;24_31del;34_35insCACCAAAA]` …

i.e. the row is offered as evidence *because* it was re-derived into the span. If the row is now left alone, the counterweight that paragraph records may no longer hold, and that is an adjudication rather than a rename. Worth checking whether the row was renamed alongside a behaviour change, and whether the record's argument survives.

## Why nothing caught this

`ruling_citation_currency.rs`'s three existing checks all judge record **ids**, and a guard name is not id-shaped. `spec_equivalence_classes_converge` reaches only the four decided records carrying an `equivalence_classes` array — neither of these does. So the ledger read as enforced on both while enforcing nothing, for as long as the renames have been in.

Census at `origin/main` `d4552167`, for context: 28 records / 25 decided / 3 undecided; 14 `::`-qualified guard citations, of which 12 resolve — all 12 assert, none is manifest-gated, and exactly one is `#[ignore]`d (the declared #1630 deferral).

## Done when

Both citations name a guard that exists and enforces the ruling the record makes, and `every_ruling_to_guard_citation_resolves` in `tests/it/ruling_citation_currency.rs` is un-`#[ignore]`d and green.

nh13:
A `decided` record and the test that enforces its ruling were connected by nothing but prose, and for eight of twenty-five by no prose at all. Rename the guard and the record goes on naming the old one; write no guard at all and the record reads exactly the same as one that has three. This adds the field that makes the difference sayable, and makes saying nothing not an option.

Representation-Change: none. Schema and test only — no file under `src/` is touched at all, the ledger's 28 records keep their statuses, questions, scope sentences, governing clauses, clause lists and rationales byte-for-byte, and the only ledger change is one added field per record.

Refs #1881.

## Census, re-derived at `origin/main` `d4552167`

**28 records, 25 `decided`, 3 `undecided`.** (#1882 is still open, so the ledger has not yet reached the 29/26/3 the brief anticipated.)

How each record pointed at its enforcement before this change:

| mechanism | records |
|---|---:|
| a `<path>::<function>` citation somewhere in the record's prose | 5 |
| a bare `tests/….rs` path — a file, but no proposition | 12 |
| an `equivalence_classes` array — the one structured pointer, already enforced end to end | 4 |
| **nothing, by any mechanism** | **8** |

The first three overlap; the last does not. A bare file path is the interesting middle case: it cannot go stale in any way a reader would notice, because a file that still exists still resolves however far its contents have moved on.

## The field

Required on every record. Exactly one of two keys.

```json
"guard": {
  "tests": [
    "tests/it/ring_segment_wellformedness.rs::the_three_malformed_rings_the_ruling_named_are_now_rejected",
    "tests/it/issue_1578_followup_self_cancelling_rings.rs::the_specs_own_ring_shapes_are_accepted"
  ]
}
```

```json
"guard": {
  "none": "Undecided: the record asks whether a ring naming only interior coordinates must be refused, and answers neither way. `tests/it/ring_segment_wellformedness.rs::a_ring_with_no_telomere_anchor_is_still_accepted` pins today's acceptance, which the record itself calls the status quo rather than a ruling — it would go on passing under one answer and would have to be deleted under the other, so it is not this record's guard."
}
```

The model is `Representation-Change:`: **declining is first-class, and what is rejected is silence.** An absent field is indistinguishable from an unconsidered one, so it does not parse. A blank reason, an empty `tests` list, both keys at once, and an unknown key are all rejected for the same reason one level in — each is the absent case wearing a value.

The citation grammar is deliberately narrow — repo-relative path ending `.rs`, then `::`, then one snake_case identifier — so a citation resolves *mechanically*. `merge::a_guard` does not parse: it names no file, and that is exactly the ambiguity the prose form had.

## Enforcement, modelled on `equivalence_classes`

That is the precedent named in the brief, and it is followed rather than paraphrased: the generator bails on an unknown class id, and the guard that reads it asserts and checks its own vacuity.

- **`generate_spec_fixture`** — the field is declared with **no `serde(default)`**, so a record without it fails the build at parse. Beyond that, `validate_guard_shape` refuses an empty guard, both keys, a blank reason, a citation outside the grammar, and a citation naming a file that is not in this tree. Red-proof: dropping one record's guard block gives `error: parse …overrides.json: missing field 'guard' at line 395 column 5`, `EXIT=1`.
- **`tests/it/common/rulings.rs`** — the one ledger parser gains `Guard`, with a mutation test per malformed shape, in the style the file already uses. No second parser was introduced.
- **`tests/it/ruling_guard_field.rs`** — new, resolves each citation against the tree.

## What the new module checks

| check | what it refuses |
|---|---|
| `every_declared_guard_resolves` | a citation naming a test that is not in the file it names (with a hint when a test of that name lives elsewhere) |
| `no_record_is_guarded_only_by_tests_that_never_run` | a record whose named guards are *all* `#[ignore]`d — naming a deferral is fine, being enforced only by one is not |
| `every_declared_guard_asserts` | a guard that runs and checks nothing (#1858) |
| `an_exemption_reason_names_no_test_that_is_missing` | a dangling citation hidden in the prose of a decline |
| `the_guard_scan_reads_the_tree_and_the_ledger` | vacuity: citation floor, scanned-test floor, and that **both** arms of the field are exercised by the ledger |

What it deliberately does **not** check is whether a named test enforces *this* ruling rather than a neighbouring one. That is a judgement no matcher can make, and pretending otherwise would put a veneer of mechanism over the thing that actually carries it. What the field buys is that the judgement is written down, once, in the ledger, where it can be argued with.

## It composes with #1880; neither subsumes the other

They ask different questions:

- **This** asks whether every record *has* an enforcement story that resolves. It is exhaustive over records by construction.
- **#1880** asks whether the discursive prose *around* a record stays true. It is opportunistic by construction: it judges records that happen to cite a guard, and is structurally blind to the eight that cite none — which is the population that most needs judging.

Both should land. There is one interaction worth stating: `alignment-only-symbol-in-a-description` now carries a `guard` naming the live `an_alignment_only_symbol_is_refused_in_every_mode_for_both_x_and_dash`, while its SCOPE paragraph still names the pre-#1684 title. **That does not close #1881** and is not meant to — the prose sentence is the record's own argument and repairing it is an adjudication. This file reads only the field. The same holds for `derivation-may-not-be-bounded-by-the-inputs-spelling`, whose guard names the two live tests its own rationale cites while the dead third citation stays exactly where it is.

To avoid a merge conflict with #1880, the new checks live in their own module rather than in `ruling_citation_currency.rs`. When #1880 lands, the tree walker and the assertion detector are the obvious things to lift into `common/`; that is a follow-up, not a reason to hold either PR.

## Per-record disposition

Every record was resolved by one of: a test whose **own doc comment names the record id** (the link was authored by the test's author, not inferred here), or a test whose doc states the record's ruling near-verbatim while the module doc names the record. Nothing was retrofitted from a guess.

**Guard found — 24 records, 42 citations.** All 42 resolve, all 42 assert, none is guarded only by `#[ignore]`d tests. Two records name a declared deferral *alongside* a running guard: `exon-junction-dup-converge-from-the-far-side` (`the_decided_target_is_convergence_on_the_near_side`) and `absolute-prohibition-enforcement-stage` (`the_decided_target_is_a_mode_gated_refusal`) — the #1630 item-2 deferral the brief names.

**Exemption declared — 4 records.** Three are the `undecided` ones, where the reason is the same in substance and different in specifics: an undecided record states a conflict and names no side, so there is no ruling for a guard to enforce, and the status-quo pin that exists in two of the three is explicitly *not* the record's guard (it would survive either answer). The fourth is `separation-rule-force-modal-or-negation`, which is `decided` and genuinely unguardable: it grades `general.md:34` as README rule 2 rather than rule 1 and mandates no output — its own closing sentence is that this "changes classification only". No description's normalized form turns on it.

**Unresolvable, left alone — 0.** Worth stating plainly, because the brief expected some and wanted them reported rather than filled: there were none. That is not because the bar was lowered — every citation was read, and the check went red on four separate perturbations before it went green.

## Red-proof

Each arm perturbed and restored with `cp`, verified by `md5` (`60f14114…` before and after, both rounds).

```
these guard citations in tests/fixtures/grammar/hgvs_spec_normalization_overrides.json name a test
that cannot enforce the record. …
  `c-and-n-positions-are-flat-transcript-offsets` names `tests/it/ruling_guard_field.rs::backticked_tokens`:
    no `#[test] fn backticked_tokens` exists anywhere under ["src", "tests"]
  `duplication-must-ranks-the-label-not-the-partition` names `…::no_emitted_insertion_repeats_its_five_prime_flanks`:
    no `#[test] fn no_emitted_insertion_repeats_its_five_prime_flanks` exists anywhere under ["src", "tests"]

these records name guards that are all `#[ignore]`d, so nothing enforces them today. …
  `exon-junction-dup-converge-from-the-far-side` names only ["…::the_decided_target_is_convergence_on_the_near_side"]

these guards run and check nothing, so they cannot fail when the ruling stops holding:
  `c-and-n-positions-are-flat-transcript-offsets` names `tests/it/ruling_guard_field.rs::a_temporary_guard_that_checks_nothing`
    (tests/it/ruling_guard_field.rs:697), whose body asserts nothing

a `guard.none` reason names a test that does not exist:
  `ring-telomere-anchoring`'s exemption names `…::a_ring_with_no_telomere_anchor_is_accepted`
```

**It lands armed, not `#[ignore]`d.** The brief anticipated a red suite and the `#[ignore]`-plus-issue idiom; that turned out not to be needed, because every record could be given either a real guard or a reasoned exemption. Arming it is the stronger outcome and it is green on its own terms — but the four failures above are what makes that a claim rather than an assumption.

### One defect the new tests caught in themselves

The assertion detector's first version scanned the function body *including its declaration line*, so a test named `asserts_nothing_at_all` matched on its own name and reported itself as asserting — the #1858 shape, passed by the check written to catch it. Its negative control caught it before the module was ever run against the ledger. The detector also has to follow a call to a helper defined beside the test: three of the ledger's guards have a body that is one `converges_to(…)` call, and a direct-only matcher called all three assertionless. Both are pinned by `the_assertion_detector_follows_a_local_helper`.

## Ledger safety

The record set-comparison from `CLAUDE.md`, before and after:

```
origin/main records: 28 | HEAD records: 28
dropped from main   : NONE
added on HEAD       : NONE
status changes      : NONE
records whose non-guard fields changed: NONE
census  (records, decided, undecided): (28, 25, 3)
```

`by_input` and `equivalence_classes` are byte-identical, and every record's non-`guard` fields compare equal object-for-object.

**Census pins: unchanged, and checked rather than assumed.** A schema addition that adds no record moves no count. `the_index_is_not_vacuous`'s `(28, 25, 3)`, `clauses_named_by_several_records_are_flagged`'s `30` / `3`, `RULING_STATUSES` and the two `CLAUDE.md` adjudication tables all hold as they stand — all four are functions of record identity, status, or clause citations, none of which this change touches.

## Verification

`cargo fmt --all --check` clean. `cargo clippy --all-features --all-targets -- -D warnings` clean. Spec fixture regenerates (`generate_spec_fixture`, `EXIT=0`, 28 rulings in). Full `cargo nextest run --features dev --lib --test it` green. The Python bindings are not reachable — `git diff --name-only origin/main..HEAD -- src/` is empty, so nothing the bindings surface changed.

### DIFF
--- a/examples/generate_spec_fixture.rs
+++ b/examples/generate_spec_fixture.rs
@@ -225,6 +225,59 @@ mod overrides {
         /// Equivalence-class ids the ruling bears on, if any.
         #[serde(default)]
         pub equivalence_classes: Vec<String>,
+        /// What enforces the ruling, or the stated reason nothing does.
+        ///
+        /// **`Option<Guard>`, not `Guard`.** The type is what carries this, and
+        /// it carries two separate things.
+        ///
+        /// It makes the field **optional at deserialize**. As a bare `Guard`, a
+        /// record lacking the key failed to deserialize — and this struct is the
+        /// generator's entry to the whole ledger, so the failure was not "one
+        /// record is unanswered" but "the ledger cannot be read".
+        /// `generate_spec_fixture` exited 1, the generated fixture never existed,
+        /// and every test downstream of it failed for a reason mentioning neither
+        /// guards nor the record at fault. Measured on this branch rebased onto
+        /// `origin/main` at `426a944b`, for two records missing one field each:
+        /// **20** failures from this alone, on top of the **28** the ledger
+        /// reader's own panic produced.
+        ///
+        /// And it keeps absence **distinguishable**. As a bare `Guard`, an absent
+        /// key and a written `"guard": {}` would both arrive as
+        /// `Guard { tests: [], none: None }`, and [`validate_guard_shape`] would
+        /// have to give one message for two different mistakes. `None` is
+        /// reachable only by the key being absent.
+        ///
+        /// The `#[serde(default)]` below is **redundant and deliberate**. Serde
+        /// already treats a missing `Option` field as `None`, so removing it
+        /// changes nothing today — verified, not assumed: with the attribute
+        /// deleted the generator still exits 0 on a ledger carrying a guardless
+        /// record. It is kept because the *intent* is "absent is allowed here",
+        /// and a future change of this field's type away from `Option` would
+        /// otherwise silently restore the cascade above. Do not read it as the
+        /// thing that makes absence legal; the `Option` is.
+        ///
+        /// **This is a change of stage, not a relaxation.** A record with no
+        /// guard still cannot merge: it is refused by
+        /// `tests/it/ruling_guard_field.rs::every_record_declares_a_guard`, in
+        /// one targeted failure that names the record and the fix. What the
+        /// generator keeps refusing is a guard that is *present and malformed*,
+        /// because that is a broken record rather than an unanswered question.
+        #[serde(default)]
+        pub guard: Option<Guard>,
+    }
+
+    /// See [`Ruling::guard`]. Exactly one of the two keys may be set;
+    /// `validate_ruling_shape` enforces that, because serde cannot express
+    /// "one of these, not both, not neither" on its own.
+    #[derive(Debug, Clone, Deserialize)]
+    #[serde(deny_unknown_fields)]
+    pub struct Guard {
+        /// `<path>.rs::<function>` citations of the enforcing tests.
+        #[serde(default)]
+        pub tests: Vec<String>,
+        /// A stated, reasoned declaration that no test enforces this record.
+        #[serde(default)]
+        pub none: Option<String>,
     }
 
     /// `spec_expected` uses the doubly-optional pattern so the override file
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

Issue #1893, opened 2026-08-14
pixi comparator tasks invoke subcommands that no longer exist, and the corpus provenance note prescribes them


nh13:
All three comparator tasks in `pixi.toml` invoke `ferro-benchmark` subcommands that no longer exist, and the mutalyzer corpus's own provenance record tells maintainers to run two of them.

### What breaks

```
$ pixi run check-validators
     Running `target/release/ferro-benchmark check-biocommons`
error: unrecognized subcommand 'check-biocommons'

  tip: a similar subcommand exists: 'check'
```

`pixi.toml:8-14` currently declares:

| task | invokes | exists? |
|---|---|---|
| `compare-mutalyzer` | `compare-normalize --validator mutalyzer` | no |
| `compare-biocommons` | `compare-normalize --validator biocommons` | no |
| `check-validators` | `check-biocommons && check-mutalyzer-normalizer` | no |

`ferro-benchmark` was restructured into `prepare` / `check` / `parse` / `normalize` / `compare` / `benchmark` / `extract` / `setup` / `generate` / `collate`. So `check-validators` maps cleanly onto `check all`, but the two `compare-*` tasks do **not** map mechanically: comparison is now a three-step pipeline (`normalize ferro -i … -o …`, `normalize mutalyzer -i … -o …`, `compare results …`) and the old single-shot form took no input/output arguments, so whatever defaults it had are gone. Replacing them needs a decision about what the task should default to, which is why this is filed rather than patched in passing.

### Why it matters more than a broken convenience task

`tests/fixtures/mutalyzer-normalize/cases.json`'s `comparator_provenance` note instructs:

> Going forward, **DERIVE expected values from the pinned comparator** (`pixi run compare-mutalyzer` / `ferro-benchmark compare-normalize`) rather than copying an upstream test suite, to prevent recurrence of the M2 drift.

Both commands named there are dead. So **the corpus's documented method for deriving an expected value cannot be run**, and the instruction reads as satisfiable when it is not. That matters because the note exists to stop hand-entered values: the M2 drift it refers to was two `insCAT` rows that were wrong until a 246-row live sweep caught them (#882/#890). A contributor who follows the instruction, finds it fails, and types a value in by hand reproduces exactly the failure the note was written to prevent.

### Also worth fixing while there

`check mutalyzer` reports both packages missing when the binary is run **outside** `pixi run`, because it then resolves system Python:

```
[...]

PR #1896, opened 2026-08-14
test(conformance): record SVD-WG010's rejected worked example as a measured cross-tool divergence

nh13:
The spec's W58 variant — SVD-WG010's own worked example — had **no corpus row**, so mutalyzer's answer for it had never been measured in this repo. It existed only inside ledger prose. This adds it, with both spellings, and both values derived from the pinned comparator.

## What was measured

| input | mutalyzer 3.1.1 | ferro (shipped arm) |
|---|---|---|
| `c.[992_1002del;1004T>C]` | `c.992_1004delinsAC` | `c.[992_993del;995_997del;999_1004del]` |
| `c.992_1004delinsAC` | `c.992_1004delinsAC` | `c.[992_993del;995_997del;999_1004del]` |

**Both tools are internally confluent, onto different answers.** That is why the spanning spelling is carried as its own row rather than mentioned in a note: it turns "the tools disagree" into "the tools disagree with no confluence confound on either side", which is a stronger and separately checkable claim.

mutalyzer lands on precisely the string SVD-WG010 recommended — and that proposal was **rejected** (`consultation/SVD-WG010.md:5` reads `!!! failure "Status: rejected"`). It is not implementing the proposal: the `delins-one-amino-acid-precondition` cluster already records that mutalyzer has no separation rule and minimises a weighted description length. The proposal would have *codified* that behaviour.

ferro's split follows `general.md:33` with `:34`'s exception unsatisfied — an 11 nt deletion is a frameshift, so the pair does not "together affect one amino acid" — and with `DNA/delins.md:47` out of reach, since ferro's canonical derivation is three pure deletions, nothing is supplied, and `:46`'s mechanism cannot have occurred (`rulings[delins-recommendation-reach-when-the-input-arrives-split]`, decided 2026-08-12).

Line numbers are read at the spec checkout `main` pins, `565b973`. #1793's bump shifted these two by one — `:33` is the separation rule that **splits**, `:34` the exception that **merges** — so the pre-bump `:34`/`:35` spelling cited the clause that argues the opposite way. Same class as #2009 / #2014, which swept 181 of these out of the ruling ledger's prose.

## What this does NOT do — verified, not assumed

**It is not a guard.** An annotated row is bucketed by `AxisTally`, never failed. I verified this by sabotage: setting the expected value to deliberate nonsense (`c.999_1000delinsQQ`) left **all 95 tests green**. That is the documented contract for every annotated row — the hermetic gate's own comment says it "asserts only genuine, unannotated divergences on real bases" — so it is the design, not a defect introduced here. It is equally true of the 87 existing `accepted_divergence` rows.

Everything short of that sabotage said the change was sound: snapshot regenerated, transcript harvested version-exact, mock pin blessed, 95/95 green, zero `Skipping:` lines, 20.9 s of real work. None of it answered whether the row was load-bearing.

**The guard existed on the spec's accession; it now covers this row's.** `the_forbidden_description_is_never_what_ferro_emits` is hermetic, needs no manifest, and fails hard if ferro emits the forbidden string — it already carried W58's `LRG_199t1:c.992_1004delinsAC`. That was not sufficient. `spec_citation` is the one disposition in this harness with **no XPASS guard**, and the value this row pins is mutalyzer's — i.e. the rejected proposal's own string — so if ferro ever adopted the forbidden form the row would demote from `spec_overridden` to `pass` and object to nothing. Adoption of the forbidden form would read as convergence. `NEGATIVE` therefore now carries the `NM_004006.2` spelling of W58 beside the `LRG_199t1` one.

Adding it required an alias, and **the alias is load-bearing, not a convenience**: the committed slice stores the *DMD* transcript only under `LRG_199t1`, and the normalizer's answer to an unresolvable accession is `Ok(input)` — the description back, unchanged — not an `Err`. An `NM_004006.2` row would therefore have passed every table in that file while asserting nothing. `slice_provider` now registers the same record under both spellings, and `the_refseq_spelling_of_the_dmd_transcript_resolves_and_answers_identically` asserts the alias resolves, that the two spellings reach the same answer, and that neither comes back as its own input.

Both directions confirmed by sabotage. Pinning ferro's actual output (`NM_004006.2:c.[992_993del;995_997del;999_1004del]`) as `forbidden` turns the guard **RED**. Removing the alias turns the new alias guard **RED** while leaving the forbidden-form test **green** — which is precisely the vacuity the alias exists to prevent.

What this PR adds is a **derived record and a census entry**, not an assertion.

## Why `NM_004006.2` and not the spec's `LRG_199t1`

`LRG_199t1` is `AccessionClass::Lrg`; `build_conformance_snapshot` harvests `RefSeqTranscript` only. An LRG row therefore **cannot** enter the hermetic snapshot and would be evaluable solely under `FERRO_MANIFEST`, which CI never sets — i.e. it would skip green rather than run.

The substitution is measured, not assumed: both tools return byte-identical answers on both accessions, for both spellings. The spec's own accession stays pinned as W58 in `spec_worked_example_rules.rs`.

This is a standing constraint on this corpus rather than a fact about this row — the spec uses LRG accessions in worked examples, so any future row drawn from one hits the same wall.

The note is reconciled with `spec_worked_example_rules::DIVERGENT`, which pins W58's `spec_answer` as the two-member form. The spec prints no *combined* description of this variant other than the rejected merged one, but `SVD-WG010.md:51` does name the two individual variants (`c.992_1002del` and `c.1004T>C`), and under `general.md:33` those compose to exactly that `spec_answer`. What is genuinely unpublished is which **split** is canonical — the spec prints no partition of this variant at all.

## The first commit is a pre-existing fix, separated deliberately

`build_conformance_snapshot --check` **already failed on unmodified `main`**: the builder harvests 29 transcripts version-exact where the committed FASTA carried 28. The missing one is `NM_002001.2`, stale since #802 backfilled it.

That matters beyond tidiness. The corpus references that accession — `NM_002001.2:c.1_3delinsATG` is a registered `normalized` divergence — so without its bases in the snapshot **that row was evaluable only under `FERRO_MANIFEST` and had been skipping green in CI**. The same failure mode this PR's second commit was written to avoid creating, already present.

It is committed separately (`9bf7fc34`) so the regeneration diff is not conflated with the new row.

## Provenance

Values derived via `ferro-benchmark normalize mutalyzer` against a prepared reference, per `cases.json`'s own `comparator_provenance` instruction, rather than copied from an upstream suite or hand-entered. Independently corroborated by a live mutalyzer web query returning the same string.

Note that instruction currently names commands that no longer exist (`pixi run compare-mutalyzer`, `ferro-benchmark compare-normalize`) — filed as #1893.

## Verification

- `generate_conformance_summary --check` → `EXIT=0`, after regenerating `tests/fixtures/mutalyzer-normalize/failure-patterns.md`. The two new `spec_citation` rows each emit a summary line; that missing regeneration was the sole cause of all three red rows (`Spec fixtures and generated docs`, `conformance_summary_generated::mutalyzer_failure_patterns_is_current` in `Code Coverage (2/4)`, and the `Test` rollup whose shards never expanded).
- Sabotage, both directions, on the new `NEGATIVE` row and its alias guard — see the guard section above.

- `build_conformance_snapshot --check` → `SNAPCHECK_EXIT=0`, "OK: committed snapshot matches a fresh harvest."
- `mutalyzer_normalize_tests` + `spec_worked_example` armed with `FERRO_MANIFEST` → `GATE_EXIT=0`, **121 run, 121 passed**, zero `Skipping:` lines.
- Sabotage run (expected value deliberately wrong) → still green, which is what established the row is a record and not a guard.

Representation-Change: none. Test fixtures and a reference snapshot only; no file under `src/` is touched and no normalized output moves. The `normalized` values recorded are mutalyzer's, and ferro's behaviour on these inputs is unchanged — it is measured here for the first time, not altered.


nh13:
@coderabbitai pause

Issue #1885, opened 2026-08-14
The real-corpus disclosure owed by the canonical-coalesced default flip is not supplied


nh13:
`rulings[delins-recommendation-reach-when-the-input-arrives-split]` (decided 2026-08-12) states in terms:

> the pass it ratifies "runs only under `PartitionRule::CanonicalCoalesced`, which is not the shipped default. So the real-corpus disclosure is owed by the change that makes that arm the default, not by this record."

#1835 is that change. It ships with the **designed-corpus** measurement (11,272 cis classes, both directions, `sequence_changed` 0) and with the manifest-backed conformance axes dispositioned, but **not** with a row count over a real corpus.

## What is owed

A count of how many stored ClinVar / CMRG / Paraphase expressions normalize to a different string under the new default, split into moved vs. denotes-different-bases, per `examples/dump_normalized_corpus.rs --verify-spdi` — or, for consumer impact, `ferro normalize -i <inputs> --reference <dir> -f tsv` through both revisions and a row-by-row diff.

The precedent is the v0.13.0 reconstruction: 5,761,302 expressions normalized through both releases, which found 577 stored strings moved and 326,404 inputs newly normalized. Nothing of that scale is claimed here.

## Why it is not in #1835

The PR body's earlier reason — that the inversion family was unresolved and would contaminate the count — is **stale**: #1706 merged and that STOP is cleared, so the measurement is now derivable. It simply has not been run to completion. A prior attempt reached 120,926 candidates plus a 198k control before being stopped; that is a partial pass and is not quotable as a result.

## What NOT to do with a zero

`examples/dump_normalized_corpus.rs` has been blind three separate ways, each invisible until the previous was fixed — member geometry (#1456), scale (#1460), and transcript geometry (#1478, a single exon with `CDS_START = 1`, so nothing crosses a junction or sits in the 5'UTR). Before quoting any figure from it, name the property the flip keys on and confirm the generator can vary it. A structural zero must be reported as structural.

Related: #1874 — the manifest-backed axes are invisible to every PR job, which is why the conformance half of this disclosure cannot be gated in CI either.

nh13:
## The real-corpus disclosure, measured

This has been run in full and is open as #1900 (`Closes #1885`). Recording the numbers on the issue so they are readable without opening the PR.

[...]

Issue #1846, opened 2026-08-13
normalize: insertion_to_duplication is O(payload²) in the expanded insert, so a megabase cross-reference ins never terminates


nh13:
`Normalizer::normalize` does not terminate in usable time on a megabase cross-reference insert: `insertion_to_duplication` rotates its payload once per base and rebuilds the payload each time, so cost is Θ(payload²). Two rows of the HGVS spec corpus are affected, and they are the reason a whole-corpus per-arm census could not be taken (#1842).

This is a **cost** defect, not a correctness one. Every existing issue on this function (#882, #1349, #1355, #403) is about which rotation it picks; none is about what it costs to look.

## Reproduce

```
FERRO_MANIFEST=<prepared manifest> \
cargo run --features dev --example measure_spec_conformance_per_arm -- \
  --fixture tests/fixtures/grammar/hgvs_spec_normalization.json \
  --output /tmp/out.json --progress-every 1
```

It stops making progress at `NC_000004.12:g.134850793_134850794ins[NC_000023.11:g.89555676_100352080]`. The 377 rows before it complete in 11 s in total.

## Mechanism

`src/normalize/mod.rs:11600` calls `rules::insertion_to_duplication(ref_seq, original_pos_idx, &seq_bytes, ...)` where `seq_bytes` is the **expanded** insert payload — the cross-reference `[NC_000023.11:g.89555676_100352080]` has already been resolved to 10,796,405 literal bases. Nothing on that path bounds its length.

`src/normalize/rules.rs:741-769`:

```rust
let base_unit = smallest_repeat_unit(inserted_seq);
let added_copies = inserted_seq.len() / base_unit.len();
if added_copies != 1 { return None; }        // so base_unit == inserted_seq
...
let u_len = base_unit.len();                  // == payload length
for r in 0..u_len {                           // 10,796,405 iterations
    let mut rotated = Vec::with_capacity(u_len);
    rotated.extend_from_slice(&base_unit[r..]);     // 10.8 MB memmove
    rotated.extend_from_slice(&base_unit[..r]);     // + the rest
    let Some((ref_start, ref_count)) = find_tandem_extent(ref_seq, pos as usize, &rotated) else { continue };
```

[...]

PR #1897, opened 2026-08-14
fix(changelog): correct #1835's flip disclosure, and make a merged one correctable

### LINKED ISSUE #1886: v0.14.0's flip disclosure names the wrong release and omits the conformance cost
`750f2334` (#1835, the partition flip) is the largest representation change the project has made, and its `Representation-Change:` trailer — which is now quoted verbatim into the v0.14.0 changelog by `scripts/inject_representation_disclosure.py` — is wrong in one place and silent in another.

## 1. It names the wrong release

The trailer closes:

> A consumer keying on the exact normalized string must expect previously-accepted inputs to return a different, still-conformant description, and must pin `FERRO_PARTITION=live` to reproduce **pre-v0.15.0** output.

`release-plz.toml` sets `features_always_increment_minor = true`, so a `feat:` against `0.13.1` cuts **0.14.0** — and that is the version the open release PR (#1753) is cutting. A consumer reading the v0.14.0 changelog is told to pin `live` to reproduce output from a release that does not exist yet, which reads as "this lands next cycle" rather than "this is the release you are reading".

A version-free phrasing would have been correct under any cut — "to reproduce output from before this change" — and the release names itself in the changelog heading. Worth adopting as the house form for future trailers.

## 2. It says nothing about the conformance cost

The trailer discloses confluence only. Spec conformance was measured per arm over the whole 932-row corpus and it **falls**:

| arm | 3' conforms |
|---|---|
| `live` / `shadow` | 650 / 932 |
| `canonical` | 646 |
| **`canonical-coalesced`** (the new default) | **648** |

So the shipped arm costs **2 rows of 932**. (The 5' column is deliberately omitted here: #1879 removes that direction from the public surface in this same release.)

**It is not a regression, and the direction matters.** `spec_expected` defaults to each row's own input, so `preserved` is a round-trip test. Where the spec publishes two legal spellings of one variant as two rows, both echo and both score as conforming — so *any* change that converges them must cost one. No row becomes non-conformant; each loses its echo by adopting the sibling spelling `DNA/delins.md:47` recommends. On the flagship row `dna-delins-vs-aligned-split-850-901`, `live` emits two forms and `canonical-coalesced` emits `c.850_901delinsTTCCTCGATGCCTG`, the spanning delins `:47` recommends.

**Do not write "more conformant"** in the correction. The measurement does not support it in either direction.

## Why this cannot be fixed in place

`scripts/inject_representation_disclosure.py` treats the trailer in the commit message as the single source of truth and reports any edited changelog copy as drift (exit 2, red CI):

> Already carried a block: judge it rather than trust it. […] Fix by correcting the trailer and re-running, not by editing `CHANGELOG.md`.

The trailer lives in `750f2334` on `main`, so correcting it would mean rewriting released history.

## The remedy the docs already prescribe

`CONTRIBUTING.md` names the escape hatch:

> A section that has already been released is a different matter: a release only prepends to `CHANGELOG.md`, leaving earlier sections untouched, and the GitHub release body is never revisited once its tag exists. So a retrospective note […] can be added afterwards, by editing the released section and `gh release edit <tag>`.

So, **after `v0.14.0` is tagged**:

1. Add an editorial correction to the released v0.14.0 section, beneath #1835's `  > ` block, covering both points above and labelled as editorial rather than as the trailer's words.
2. Mirror it into the GitHub release body with `gh release edit v0.14.0`.

**Measured, so whoever does this does not have to re-derive it:** a correction line placed *after* the `  > ` block, with any prefix other than `  > `, passes both halves of the `Changelog grouping audit` — `inject_representation_disclosure.py --check` and `check_changelog_grouping.py` both exit 0. `_is_disclosure_line` only collects `  > `-prefixed lines, so the editorial note is outside the block the script owns and is left alone. This keeps the trailer's own words verbatim, which is the distinction the script's design turns on.

Doing it before the tag does not work: `publish.yml` runs release-plz on **every push to `main`**, and release-plz regenerates the pending section of an open release PR, so a note added to #1753 now is silently discarded at the next merge. (The disclosure injection itself has the same property, which is why `CONTRIBUTING.md` says to re-run it after any regenerating push.)

nh13:
Closes #1886.

`750f2334` (#1835) is the largest representation change the project has made, and `scripts/inject_representation_disclosure.py` quotes its `Representation-Change:` trailer verbatim into the v0.14.0 changelog. That trailer names the wrong release and is silent about a conformance cost — and it cannot be edited, because it lives in a merged commit message.

## Where the fix had to land, and why not `CHANGELOG.md`

Three surfaces were candidates and two of them are dead ends:

- **The trailer.** Immutable. Correcting it means rewriting released history.
- **`CHANGELOG.md`.** Fails twice over. `publish.yml` runs release-plz on every push to `main` and release-plz regenerates the pending section of an open release PR, so a hand-written note is silently discarded at the next merge; and `inject_representation_disclosure.py --check` reports an edited block as drift against the trailer it no longer matches, which is exit 2 and red CI.
- **The injector.** Durable, because it re-renders the block on *every* run. So the correction is registered there, in a new `EDITORIAL_CORRECTIONS` table keyed by PR number, and is re-attached every time release-plz regenerates the section. `--check` then demands it rather than reporting it as drift.

It renders after the trailer's own words, inside the same blockquote, separated by a bare `  >` and opening with a label that names it editorial and cites this issue. Inside the block deliberately: a correction a reader can meet without the statement it corrects is worse than none, and every placement *outside* the block can be separated from it — one after it is dropped on the next regeneration, one before it re-orders on the next trailer. What keeps the trailer's words identifiable as the trailer's is therefore the label, which a test enforces on the real register rather than leaving to review. The bare `  >` separator is load-bearing too: without it a correction is a lazy continuation of the trailer's last paragraph and renders as the author's own closing sentence.

Note this is *not* the mechanism #1886 measured (a non-`  > `-prefixed line after the block, which passes both halves of `Changelog grouping audit` because `_is_disclosure_line` does not collect it). That works for a one-off hand edit of an already-tagged section; it does not survive a regeneration, and the release PR is still open.

## 1. The wrong release

The trailer closes by telling consumers to pin `FERRO_PARTITION=live` to reproduce **pre-v0.15.0** output. Verified rather than taken on faith: `Cargo.toml` is at `0.13.1`, `release-plz.toml` sets `features_always_increment_minor = true`, and the open release PR #1888 is titled `chore: release v0.14.0` and bumps `Cargo.toml` to `0.14.0`. So a consumer reading the v0.14.0 changelog is pointed at a release that does not exist yet.

`CONTRIBUTING.md` now adopts the version-free house form — *"to reproduce output from before this change"* — under **Declaring a representation change**, with the mechanism (which release a `feat:` cuts is not knowable when the trailer is written) and #1835 as the worked example of getting it wrong.

## 2. The undisclosed conformance cost

**Re-measured here, not copied.** `examples/measure_spec_conformance_per_arm`, one process per arm, `--direction three-prime`, against the prepared GRCh38 reference, at `45820926` (`origin/main` at branch time). `--only-inputs` names 932 of the fixture's 934 rows: the two excluded are `NC_000004.12:g.134850793_134850794ins[NC_000023.11:g.89555676_100352080]` and its `inv` sibling, which #1846 measures as Θ(payload²) in a 10.8 Mb expanded insert and which do not terminate.

| arm | 3' conforms of 932 |
|---|---|
| `live` | 646 |
| `shadow` | 646 |
| `canonical` | 642 |
| **`canonical-coalesced`** (the new default) | **644** |

Every run: `completeness {attempted: 932, succeeded: 932, dropped: 0}`, `reached_partitioner = 232` (so the arm was genuinely exercised — a zero there would mean the census measured nothing about the arms), `unevaluable = 40`, `EXIT=0`.

### These are 4 lower than #1886's table, and the difference is arm-independent

#1886 records 650 / 650 / 646 / 648. Mine are exactly 4 lower on **all four** arms, so the shape is identical — `live == shadow`, `canonical-coalesced = canonical + 2`, and the number the disclosure turns on, **`live - canonical-coalesced = 2`, reproduces exactly**.

The uniform −4 is #1882, which merged after #1886's measurement was taken: it refuses a `c.` description whose reference resolves no CDS. Five corpus rows are now refused on `NM_004006.1` (`has no CDS start`), and exactly four of them carried a non-null `spec_expected` and so had been scoring as `preserved` — `c.123=`, `c.5697del`, `c.761_762insN[5]` and `c.[123=;456=;789=]`. The fifth, `c.-128354C>T`, has `spec_expected: null` and cost nothing. #1882's own commit message names `refseq.md:145`'s `NM_004006.1:c.5697del` as being in its moving set.

The changelog correction quotes the numbers measured at `45820926`, since that is the base the release is being cut from.

### The 2 rows, named — because the count alone reads as a regression and neither row is one

Diffing per-row output between `live` and `canonical-coalesced`: 8 rows change output, **2 stop matching the spec's stated form and 0 start**.

**Row 1 — the `DNA/delins.md` pair.** Both rows are harvested from `docs/recommendations/DNA/delins.md`:

| corpus row | `live` | `canonical-coalesced` |
|---|---|---|
| `c.[850_869del;874_881del;887_897del;901_902insG]` | echoes itself (`preserved`) | `NM_004006.2:c.850_901delinsTTCCTCGATGCCTG` (`diverges`) |
| `LRG_199t1:c.850_901delinsTTCCTCGATGCCTG` | echoes itself (`preserved`) | echoes itself (`preserved`) |

`spec_expected` defaults to a row's own input, so `preserved` is a round trip. The spec publishes one variant as two rows — the spanning delins `:47` recommends, and the split `:46` calls its "alternative description" — and both echoed. Converging them onto one canonical form must cost exactly one echo, whichever form wins. The winner here is the spanning form the spec recommends.

**Row 2 — `LRG_199t1:c.992_1004delinsAC`, which is not from the recommendations at all.** Its `source_kind` is `consultation`, `working_group: SVD-WG010`, harvested from `docs/consultation/SVD-WG010.md:50` — the **rejected** proposal. `tests/it/spec_worked_example_rules.rs`'s `W58` already pins that spelling as one ferro must *not* produce from the split input; under the new default ferro no longer echoes it back either, emitting `c.[992_993del;995_997del;999_1004del]`.

So #1886's summary — "each loses its echo by adopting the sibling spelling `DNA/delins.md:47` recommends" — holds for row 1 and **not** for row 2, which is a different shape with a different provenance. The correction says so rather than repeating it. Per #1886 it does **not** say "more conformant"; it does not say "less" either, because the measurement supports neither.

The 5' column is omitted throughout: #1879 removes that direction from the public surface in this same release.

## Left for #1885

The trailer's own closing statement — that the real-corpus row count over ClinVar/CMRG/Paraphase is owed and tracked as #1885 — is untouched, and the register takes paragraphs by append, so #1885's figure lands as a further paragraph rather than a rewrite of either of these.

## Verification

- `python3 scripts/inject_representation_disclosure.py --check` on this branch: `EXIT=0` (no-op — `CHANGELOG.md` here holds only released sections).
- Rendered end to end against a copy of #1888's own `CHANGELOG.md`: 34 disclosures attached, `EXIT=0`; the correction lands under #1835's bullet; a second run and `--check` are both `EXIT=0` (idempotent, which is what makes it survive regeneration).
- `python3 scripts/check_changelog_grouping.py`: `EXIT=0`, "Representation changes section is consistent with the trailers".
- `pytest tests/python/test_inject_representation_disclosure.py tests/python/test_changelog_grouping.py tests/python/test_representation_change_trailer.py`: 186 passed.
- `ruff check` / `ruff format --check` over `scripts/` and `tests/python/`, `cargo fmt --all --check`, `cargo clippy --all-features --all-targets -- -D warnings`: all clean.

Five new tests: nothing is attached for a PR with no registered correction; a registered one lands below the trailer with its separator; re-running is a no-op; deleting it from the changelog is reported as drift; and every paragraph in the real register opens by naming itself editorial and citing an issue.

Representation-Change: none. Documentation and a changelog-rendering script; nothing under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched, no normalized output moves, and no expectation was re-blessed. The conformance figures quoted above are a measurement of the already-merged #1835, not of this change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -177,6 +177,21 @@ Four things belong in it, and the last is the one a consumer actually acts on:
    it is a migration. This distinction cannot be recovered later — nobody
    reading the diff in six months can tell — so it has to be written down now.
 
+**Do not name a release version in the trailer.** Which release carries your
+change is not knowable when you write it: `release-plz.toml` sets
+`features_always_increment_minor`, so what a `feat:` cuts depends on the base
+version at release time, a `feat:` landing before yours moves it, and a cycle can
+be re-cut. Say **"to reproduce output from before this change"**, not "to
+reproduce pre-vX.Y.Z output". The version-free form is correct under every cut
+and loses nothing, because the changelog heading the trailer is quoted under
+already names the release the reader is holding.
+
+This is the house form because it has already been got wrong the other way:
+#1835's trailer told consumers to pin `FERRO_PARTITION=live` to reproduce
+**pre-v0.15.0** output, for a change shipping in **v0.14.0** ([#1886]) —
+which reads to a consumer as "this lands next cycle" rather than "this is the
+release you are reading", and could not be fixed in the trailer once merged.
+
 Two related habits, from `CLAUDE.md`:
 
 - **"Fixes non-confluence" is not a sufficient description.** Confluence (two
@@ -203,6 +218,7 @@ whatever its type, so a `fix:` that correctly declined is not listed under
 
 [#1556]: https://github.com/fulcrumgenomics/ferro-hgvs/issues/1556
 [#1557]: https://github.com/fulcrumgenomics/ferro-hgvs/issues/1557
+[#1886]: https://github.com/fulcrumgenomics/ferro-hgvs/issues/1886
 
 ### Changelog
 
@@ -222,6 +238,14 @@ CI runs the `--check` form in the `Changelog grouping audit` job, so a release P
 Two consequences for how you write the trailer. Keep it **at column 0 and undecorated** — no backticks, no `>`, no `-`/`*`/`+`, no `#`, no emphasis, and the separator is a hyphen. Anything else and the line is prose: an indented `Representation-Change:` is a continuation of the line above it to git, to release-plz and to both checkers, and a code-spanned one is invisible to all three at once. This is not pedantry about formatting — you would have declared something and nobody would be told, which is the exact failure the trailer exists to prevent — so the checker now refuses a line that looks like a declaration it cannot read, names the character responsible, and does so on every PR rather than only on the ones touching a watched directory. To show the form as an *example* rather than declare it, indent the line — and keep your real, unindented declaration above it, because that is what tells the checker the indented one is a quotation. An indented line on its own is reported as a near miss rather than read as an example, which is the right answer: nothing distinguishes it from a declaration that will never be read. And know that "put it last" is advice the tooling cannot rely on: GitHub appends CodeRabbit's summary after your description, so the script stops the value at the next column-0 trailer token, Markdown heading or HTML comment rather than at the end of the message. Do not put a heading inside the disclosure.
 
 A section that has already been released is a different matter: a release only prepends to `CHANGELOG.md`, leaving earlier sections untouched, and the GitHub release body is never revisited once its tag exists. So a retrospective note — a consolidated representation summary for a release whose commits predate this convention, say — can be added afterwards, by editing the released section and `gh release edit <tag>`. That is the escape hatch for a cycle that shipped without trailers; it is not a substitute for declaring changes as you make them, because nobody can reconstruct the rejected-vs-accepted distinction after the fact.
+
+**Correcting a disclosure that is already merged is a third case, and hand-editing `CHANGELOG.md` is the wrong tool for it.** A trailer cannot be corrected once its commit is on `main` — that would mean rewriting released history — and an edited changelog copy fails twice over: release-plz discards it the next time it regenerates the pending section, and `inject_representation_disclosure.py --check` reports it as drift against the trailer it no longer matches, which is red CI. The durable surface is the injector itself, because it re-renders the block on every run. Register the correction in its `EDITORIAL_CORRECTIONS` table, keyed by PR number, and re-run the script:
+
+```bash
+python scripts/inject_representation_disclosure.py
+```
+
+It is appended under the bullet after the trailer's own words, inside the same blockquote, and `--check` then demands it. One limit to know before you reach for it: the injector *judges* a block that is already on disk rather than rewriting it, so registering a correction against an already-rendered bullet is reported as drift. That resolves itself in the pending section, which release-plz regenerates on every push to `main`; a **released** section is never regenerated, so there you delete the stale block from `CHANGELOG.md` and let the re-run write it back with the correction attached. Two constraints, both load-bearing. Each paragraph must **open by naming itself editorial and citing the issue that raised it** — the release checklist turns on whose words a disclosure carries, and inside the block the label is the only thing that answers it. And a correction states a **fact the trailer got wrong or left out**; it is not a place to re-word, summarise or improve a disclosure, which is what keeps "the trailer's own text" a claim a reader can rely on. Append a paragraph rather than rewriting one, so two corrections to one disclosure conflict visibly instead of silently overwriting each other.
 
 ### Submitting a Pull Request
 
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1898, opened 2026-08-14
test(python): report the #1455 GIL margin on every run, and retry once more

nh13:
`test_reference_backed_calls_release_the_gil` ejected #1838 from the merge queue on a **1.1% miss**, while a PR-level run on the **same head** passed sixteen minutes earlier, on a diff (one JSON string and one generated markdown file) that compiles into nothing in the wheel.

Deciding whether that was noise or a standing tax on every merge requires knowing how close CI normally runs to the threshold. **That is not recorded anywhere.** This makes it recorded.

## The gap

A passing run emits only the calibration:

```
#1455 GIL calibration: fair-share=0.429 gil-held=0.312 ratio=0.728 (skips if ratio >= 1.0)
```

That says whether the measurement could *discriminate*. It does not say what any entry point actually scored. The share appears only in the assertion message, which fires only on failure — **so the only runs that record a share are the ones that fail**, and the population cannot be characterised from history at all.

What is measurable today, and what it suggests:

| | quiet 12-core host (docstring table) | CI 4-vCPU runner |
|---|---|---|
| `held / fair` | ~0.64 | **0.728** |
| entry point vs fair | **3.4–4.1×** | **1.483×** (the ejection) |

The two populations are already converging on CI while `MAX_CONTROL_OVER_FAIR = 1.0` is nowhere near firing — so the skip guard does not protect against the marginality actually present. Whether CI normally clears at 1.6× or at 1.51× decides whether this recurs, and no log can answer it.

Frequency, for scale: **1 failure in 200 CI runs**, but **1 in 22 `merge_group` runs** — and every merge group is a merge attempt. That is n=1 in the bucket that matters, which is exactly why the margin needs recording rather than the rate being guessed at.

## Changes

- **Emit the margin on every run**, passing or failing — `share`, `threshold`, `share/fair`, and the attempt count. A warning rather than a print, for the reason the calibration already uses one and says so: pytest captures stdout for passing tests, and the warnings summary survives into a CI log.
- **`ATTEMPTS` 2 → 3.** The constant's own comment already bounds the cost: *"A retry cannot manufacture a false pass — a GIL-holding binding sits at the control's own share, nowhere near the threshold — but it does turn a transient CPU spike into a slower success."* Two attempts did not absorb the spike above.

**`MIN_SHARE_OVER_FAIR` is deliberately not lowered.** That would weaken the property #1455 exists to prove. Tightening `MAX_CONTROL_OVER_FAIR` so a converging host *skips* rather than fails is arguably the right follow-up — the docstring already says "a measurement that cannot discriminate should not be reported as a failure" — but choosing that number needs the data this PR starts collecting, so it is deliberately not attempted here.

Nothing about what the test asserts changes.

## Verification, including what it does NOT show

Built and run as CI does: `maturin build --release --features python` into a clean venv, with the import asserted to resolve from the installed wheel.

**The eight parametrized cases SKIP on this host** — load average 79.73 across 12 cores, ~6.6 per core against `MAX_LOAD_PER_CORE = 0.7`. That is the guard working correctly, and it means **no valid measurement could be taken here**. Reporting `2 passed, 8 skipped` as a green verification would have been exactly the skip-green failure this repo keeps finding.

So the emission was proven by temporarily lifting the load and discriminate guards locally and confirming the line renders with every field bound:

```
#1455 GIL margin: Normalizer.to_spdi share=1.464 threshold=1.076 share/fair=2.041 (needs > 1.5) attempts=1/3
```

The measurement in that probe is meaningless on a saturated host — only the code path is verified. **That probe was reverted and the committed diff contains neither change**, confirmed by grep before committing.

The number this PR exists to collect will therefore first be visible in this PR's own CI run.

Representation-Change: none. Python test only; no file under `src/` is touched and no normalized output can change. Three things change, not one: the margin diagnostic, `ATTEMPTS` 2 -> 3, and five tests pinning the diagnostic itself. No threshold constant moves — `MIN_SHARE_OVER_FAIR`, `MAX_CONTROL_OVER_FAIR`, `MAX_FAIR_SHARE` and `MAX_LOAD_PER_CORE` are all unchanged — but the extra attempt makes the pass criterion strictly weaker: a case clears on the best of three medians instead of the best of two.

### DIFF
--- a/tests/python/test_issue_1455_gil_release.py
+++ b/tests/python/test_issue_1455_gil_release.py
@@ -71,9 +71,24 @@ call would never have given it, and the two populations converge — at load
 average 30 on this 12-core host the held-GIL calibration rose to 0.311 while the
 weakest entry point fell to 0.349. Two guards detect that and **skip**, because a
 measurement that cannot discriminate should not be reported as a failure: see
-``MAX_LOAD_PER_CORE`` and ``MAX_CONTROL_OVER_FAIR``. The calibration numbers are
-emitted as a warning on every run so a skip on a machine you cannot log into is
-still diagnosable from the CI log.
+``MAX_LOAD_PER_CORE`` and ``MAX_CONTROL_OVER_FAIR``.
+
+What every run emits
+--------------------
+
+Two kinds of warning, so a run on a machine you cannot log into is diagnosable
+from the CI log alone:
+
+* one module-scoped ``#1455 GIL calibration:`` line, carrying ``fair`` and
+  ``held``; and
+* one ``#1455 GIL margin:`` row **per parametrized entry point** — ``MEASURED``
+  with the observed ratios, or ``STOOD-DOWN`` naming the guard that fired.
+
+The second is the point of the exercise, and the invariant it rests on is that
+the two shapes are exhaustive: a case emits one row or the other, never neither.
+So a parametrized case with no row did not run at all — a stand-down is never
+silent, and cannot be mistaken for a healthy quiet run. ``_margin_line``'s own
+docstring says which of its fields may be pooled across runs and which may not.
 """
 
 import json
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1899, opened 2026-08-14
fix(normalize)!: delete the input-relative weight bound and derive the split cap

nh13:
Split out of #1616 so the rebase past #1835's canonical-coalesced default flip is one resolution rather than fifteen. #1616 keeps the rest of its content and stacks on this.

## What changes

**1. The input-relative weight bound is deleted.** `canonicalize_from_sequence` carried

    if derived_columns > changed_columns_of_edits(&edits) { return None; }

which refused any derivation heavier than the member list the caller happened to write. Weight is the sum over members of `max(ref_len, alt_len)`, and the comparand is the **input's own spelling** — so on refusal the variant fell back to the per-member pipeline, which never re-aligns across members, and the authored spelling survived verbatim.

It cites no clause and cannot: the recommendations compare a description to the **sequence**, never to the input, and `background/basics.md:38`'s design values (stable, meaningful, memorable, unequivocal) omit minimality. It could not be narrowed, only deleted — a span covering a block always weighs at least as much as any split of that block, so a bound of this shape refuses every merge on principle.

Recorded as `rulings[derivation-may-not-be-bounded-by-the-inputs-spelling]`. **Read that record's scope:** it removes an input-relative comparand; it does *not* license the merges that comparand happened to be blocking. Those stay governed by `general.md:34` and by `delins-merge-vs-individual-gap-two-or-more`.

**2. `MAX_SPLIT_BLOCK` is derived rather than picked** — `MAX_CANONICAL_WINDOW` instead of a literal 1024 — and the unexamined-block axis gate is deleted with it.

**3. `MAX_SEPARATION_AUDIT_WIDTH` gets its own literal.** It was an alias of `MAX_SPLIT_BLOCK`, tying a `Θ(ref · alt)` recursive audit to a bound fronting a linear sweep, so any change to the linear one silently moved the quadratic one.

**4. `rulings[unchanged-is-read-over-every-minimal-alignment]`**, recovering a decision that was made and then lost.

## Why (2) is not separable from (1)

This was tried as two PRs and the split does not hold. At a 1024 cap a block past the cap comes back **unexamined**, so the per-member pipeline answers and the input's own partition survives. The weight bound was masking that; with the bound deleted it is not.

Measured by reverting the constant in place and running the full gate — `case_harvest`'s 1,193 bp genomic row:

    pinned (and with the cap raised):  g.[4517618_4517622del;4517624_4518489del;4518497_4518812del]
    bound deleted, cap left at 1024:   g.4517618_4518810delinsTCATAA

Both `spec_conformance_axis` censuses move in that configuration too, and the `#1616` gate test fails. At `MAX_CANONICAL_WINDOW` every one of them passes. That row is the **only** case in the committed corpora inside the 1024–4096 band, and it exists precisely because a length-gated change is invisible to a corpus of 20-mers (#1460). It earned its keep.

## The cost, and a claim withdrawn

One rep each, same tree and reference, full suite:

| `MAX_SPLIT_BLOCK` | `MAX_SEPARATION_AUDIT_WIDTH` | wall |
|---|---|---|
| 1024 | alias (1024) | 260.1 s |
| 4096 | alias (4096) | 300.0 s |
| 4096 | pinned 1024 | **266.4 s** |

Raising the cap costs **~2%**. The 15% in the middle row was the aliased audit width being dragged up with it — a measurement moving both at once reads this bound as roughly 8× more expensive than it is. An earlier reading of this as "1.7×" was wrong twice over: wrong baseline *and* the alias.

An earlier revision of this branch justified the cap by claiming it recovered **18 (3') and 14 (5')** `spec_conformance_axis` classes. **That figure predates #1835's flip of the default arm and does not reproduce; it is withdrawn.** The check that appeared to confirm the withdrawal was *also* wrong — it ran on a tree where the axis gate had already been deleted, which is a third configuration rather than the one a 1024 cap produces, and it exercised only the censuses and not `case_harvest`. Both readings are recorded in the constant's doc comment, because the failure mode is the transferable part: **measuring a configuration that is not the one under test.**

The gate itself goes because at cap == window it **cannot fire** — the window test refuses the same blocks one step earlier, on every axis. Verified with a 4,200-base block: both `g.` probes declined via the window and the `c.` positive control **failed**, which is what proves the carve-out was unreachable by any input rather than merely unused.

## `unchanged-is-read-over-every-minimal-alignment`

Decided **2026-08-07** while building the forced-unchanged-column detector for #1539/#1540, and never written into the ledger — so it was re-derived twice on 2026-08-14 and got a *different* answer each time. That is the cost the record-every-adjudication policy names, which is why it is recorded here rather than left in a session log.

A reference base is unchanged **iff every minimal alignment matches it**: column-based, not cell-based and not position-wise. `Dominators::matched_ref()` is cell-based — "is one `(ref, alt)` cell a Match on every minimal path?" — whereas `general.md:34` is column-based: "is this reference *base* unchanged in every minimal alignment?". On `GACA → AGAT` both cost-3 alignments match ref offset 1 but to different alt offsets, so the cell notion sees nothing while the base is unchanged either way.

The worked case, `CAG → AGA`, equal length 3/3:

    position-wise:            C->A, A->G, G->A          cost 3
    del C, A=A, G=G, ins A                              cost 2   <- minimal

Edit distance is **2**, not 3. So the middle bases are genuinely unchanged, `general.md:34` describes the two changes individually, and `DNA/delins.md:16` has no antecedent — under the minimal alignment there are no two *consecutive changed* nucleotides.

**Equal length does not imply no indel.** That false premise — "an equal-length block has no gap to place" — was written into `merge.rs` three times as justification, and all three exemption guards key on `reference.len() != result.len()`, so each exempts precisely the class that breaks it. The exemptions remain **sound as cost scopes** (`best_alignment` declines to search there), so no code changes for this: only the three justifications, each corrected and naming the record.

**Stated as unmeasured in the record itself:** the affected class is every equal-length block whose minimal alignment is cheaper than position-wise. A generator must key on **edit distance against block length**, not length alone.

## Verification

- Full gate **10,725/10,725**, `FERRO_MANIFEST` armed, `FERRO_REQUIRE_BULK_FIXTURES=1`, zero `Skipping:` lines — run locally, on an Apple M2 Max.

  **Not all of that is pre-merge coverage, and the split matters here.** PR CI never provisions `FERRO_MANIFEST` (`ci.yml:1174` emits a notice saying so), so the reference-aware mutalyzer / biocommons / hgvs-rs axes in that total **skip silently** on this PR and are nightly-only. What *is* checked pre-merge includes the row this change turns on: `case_harvest` is manifest-free — "No manifest, no environment variable, no skip path" (`tests/it/case_harvest.rs:160`) — so the 1,193 bp genomic row demonstrating that the cap is load-bearing runs on PR CI, as do both `spec_conformance_axis` censuses, which are corpus-backed rather than manifest-backed.
- `cargo clippy --all-features --all-targets -- -D warnings` clean.
- Python binding suite run locally — `cargo nextest` does not reach it, and this touches `src/normalize/` and `src/project/` (the #1870 shape).
- Ledger set-comparison: nothing dropped from `main`, no status changes. The rebase conflict here was **additive** — main's #1882 record against this branch's, both as the array's last element — so resolving to either side would have destroyed a ruling. Both kept, JSON re-validated, 31 records, no duplicates.
- The generated clause index was regenerated from the test's own output, not hand-edited: census `(31, 28, 3)`, multi-count `34`.

Two corrections to things this branch was itself about to ship wrong. `src/conformance/spec_corpus.rs` carried two figures describing #1616's full content measured on a pre-#1835 base — the shipped coding-axis merge counter is **3**, set by #1835, rather than 0 or 8, and a green tree could not have produced 8 since that census is an exact equality against a pin this branch does not touch. And the ` ```text ` fences around the generated clause index had been dropped by an earlier splice; the currency test **skips fence lines** when comparing, so their loss could never have turned it red.

Representation-Change: none. Measured, not assumed. The full gate passes against main's pinned values throughout, including case_harvest's 1,193 bp row, both spec_conformance_axis censuses (each an exact equality against a pin this branch does not modify), and every reference-aware corpus with the manifest armed. The corpus A/B between the merge-base and this branch produced byte-identical dumps (85,642 rows; cmp reports no difference), with 0 rows moved across all 21 shape families, including split_vs_spanning_delins (9,216 rows), three_member_allele (4,608) and protein_equal_length_delins (1,584) - the shapes a weight-bound refusal would move if it moved anything. The three equal-length comment corrections change no code at all. Two blind spots stated rather than papered over: the synthetic corpus builds 20-mer cores, so it declines 0 of 57,058 blocks at either cap setting and cannot observe MAX_SPLIT_BLOCK at all (#1460) - the committed case_harvest row is what covers that band, and it is pinned and unmoved; and no generator here varies edit distance against block length, so the equal-length class the new ruling governs is undermeasured (#1478 is the same family). Report a structural zero as structural.


nh13:
@coderabbitai pause

nh13:
Back-reference for traceability: this PR implements **#1440** ("Weight bound judges a sequence-derived partition against the input's spelling"), which I've now closed citing it.

Flagging it rather than editing this PR's body, because neither end named the other — this PR carries no `Closes #1440` and no mention of the issue, and the ruling record it rewrote (`derivation-may-not-be-bounded-by-the-inputs-spelling`) carries the implementation and its measured figures but does not name this PR either. So the change was untraceable from both directions. A follow-up adds the citation to the record.

One independent corroboration worth having on the PR, measured before it landed: the deletion was simulated on then-current `main` (`19e2e244`, pre-merge) by neutering the comparand, with `FERRO_MANIFEST` set against a prepared GRCh38 reference. `spec_conformance_axis` came back **byte-identical — EXIT=0, all 5 tests pass, `guard_violations` 0, every pinned figure unchanged.** That was measured for a different purpose and reached the same conclusion this PR did, from a different direction.

Two consequences of the merge that are now live and may be worth knowing here:

- **#1904** (`fix/issue-1878-canonical-coalesced-type-ranking`) is now `CONFLICTING` against `main` — verified real with `git merge-tree`, not `mergeable`: `CLAUDE.md`, `docs/NORMALIZATION_CONTRACT.md`, `src/project/projector.rs`, the overrides ledger and `tests/it/clause_ruling_index.rs`. Every figure in its trailer also predates this PR and needs re-measuring. Its confluence give-back was explicitly "restorable only by #1440", so it may now be strictly better than before.
- **#1908** depends on this bound without saying so: `split_is_a_placed_gap_coincidence` admits separation 1 on a frameless axis, and under the same pre-merge simulation that raised `guard_violations` 0 → 5 on `svd-wg010-frameless-separation-floor-of-two` rows. Being handled on that PR.

PR #1900, opened 2026-08-14
test(normalize): measure the canonical-coalesced flip over the four real corpora

### LINKED ISSUE #1885: The real-corpus disclosure owed by the canonical-coalesced default flip is not supplied

nh13:
Closes #1885.

`rulings[delins-recommendation-reach-when-the-input-arrives-split]` says in terms that the real-corpus disclosure "is owed by the change that makes that arm the default". #1835 is that change, and it shipped with the **designed**-corpus measurement (11,272 cis classes) and the manifest-backed conformance axes, and with no row count over stored expressions. A prior attempt reached 120,926 candidates before being stopped; #1885 says explicitly that a partial pass is not quotable.

This adds the instrument and records what it produced. **The corpus was measured in full — 9,949,738 rows, `--stride 1`, no sampling.**

## The result

| | rows |
|---|---:|
| compared | **9,949,738** |
| normalized on both arms | 9,903,542 |
| **strings that move** | **11,491 (0.115%)** |
| &nbsp;&nbsp;• respelled, **same** denoted bases (identical SPDI key) | 11,476 |
| &nbsp;&nbsp;• respelled, denotation **unkeyable** by SPDI on one side | 15 |
| &nbsp;&nbsp;• **denote different bases** | **0** |
| rows that start or stop normalizing | 0 |

Per corpus: ClinVar 500k **495** of 500,004 · ClinVar unique **3,080** of 4,188,436 · CMRG **6,961** of 4,826,063 · Paraphase **955** of 435,235.

These inputs were **previously accepted**, so this is a real migration for a consumer keying on the exact normalized string — 11,491 stored strings need re-mapping. It is not a change of meaning: no row's denoted bases move, so a consumer keying on SPDI, position, or anything derived from the applied sequence sees nothing at all.

**The direction is mostly a split, not the merge the passage's name suggests** — worth stating, because the flip is usually described as adopting `DNA/delins.md:47`'s spanning form:

| shape | rows |
|---|---:|
| spanning `delins` → members | 7,760 |
| members → *different* members | 3,564 |
| members → spanning `delins` | 167 |

```
NM_001293192.2:c.767_779delinsGCCCT
  before this change ->  c.767_779delinsGCCCT            (echoes the submitted form)
  after              ->  c.[767_768delinsG;770_774del;776_777del;779C>T]
  SPDI on both arms  ->  NM_001293192.2:1206:TTCTGGAACAGCC:GCCCT
```

## The 15 unkeyable rows, since "unkeyable" must not be filed as agreement

Five are `NG_` accessions, which `canonical_spdi` declines on both arms. Ten are junction-crossing outputs that #1704 re-parents onto the genomic reference (`NC_000003.12(NM_000094.3):c.[2305del;2307_2312del;2314+1del]`), carrying an intronic offset SPDI has no notation for. Both classes are string moves whose denotation the key cannot compare, so they are counted apart rather than assumed into either bucket — an absent key is not evidence of agreement.

## The zero is not structural

The property the flip keys on is which partitioner `partition_block_for_rule` cuts with, and that call is reached from `sequence_first_pass` for a cis allele of two or more members **and** for any lone member with a certain edit. `normalize::partition_blocks_cut()` — which counts on every arm, including `live` — records **8,602,499 of 9,949,738 rows (86.5%) cutting at least one block**, 8,619,784 blocks in all. So the corpus reaches the switch, and the `denotes-different-bases: 0` is a measured zero rather than a corpus that could not vary the property. (This is the check #1456 / #1460 / #1478 exist to force; it is asked of `partition_blocks_cut`, not of the designed-corpus generator, which is not used here at all.)

## What is excluded, and why it is excluded rather than timed out

**9,418 rows (0.095%) are excluded before normalizing**, and are written into the artifact as `excluded-oversize` rather than silently dropped. #1846 records `insertion_to_duplication` as Θ(payload²), so a multi-megabase payload does not terminate — that is what stalled the previous attempt. The bound is a coarse coordinate span read off the input string (`coordinate_span`), deliberately over-inclusive and deliberately **not** a wall-clock timeout, so the same input is excluded on every run and the exclusion is legible from the input alone. Nearly all of them are `g.(?_N)_(M_?)del` structural deletions spanning megabases.

**2 further rows panic identically on both arms** — `NC_000012.11:g.110593351_110576466dup` and `NC_000016.9:g.23634775_23621090dup`, a debug-build `attempt to subtract with overflow` at `src/normalize/mod.rs:5350` on a reversed range. They are recorded with status `panicked`, not dropped. Being arm-independent they cost the disclosure nothing, but they are a real defect on a shape SVD-WG006 admits and are worth a separate issue.

An excluded or panicking row is a row this measurement does **not** cover. It is never a row it found unchanged.

## Where the disclosure lands

The measurement is written into the ruling record that recorded the debt — `delins-recommendation-reach-when-the-input-arrives-split` — which #1850 publishes as `docs/NORMALIZATION_CONTRACT.md`, so it reaches a reader without opening this PR. The paragraph says in its own words that the figure belongs to #1835 and not to that ruling, which still moves no row by itself.

**It deliberately does not touch `CHANGELOG.md` or the injector.** #1897 (for #1886) is open against the same disclosure and builds the durable mechanism — an `EDITORIAL_CORRECTIONS` register keyed by PR number, re-rendered on every regeneration — and says in its own body that "the register takes paragraphs by append, so #1885's figure lands as a further paragraph rather than a rewrite of either of these". Editing the same file here would fight that PR for no gain. Once #1897 lands, the changelog half of this is one appended paragraph:

> Editorial correction (#1885): the real-corpus row count this trailer defers is now measured. Over all four bulk corpora in full — 9,949,738 stored ClinVar, CMRG and Paraphase expressions — 11,491 normalized strings move (0.115%): 11,476 respellings whose SPDI key is unchanged, 15 whose denotation SPDI cannot key, and **0 that denote different bases**. No row starts or stops normalizing. 9,418 rows (0.095%) are excluded from the measurement by a cost bound rather than found unchanged.

## The instrument

`examples/measure_flip_corpus_impact` reads the four bulk fixtures, normalizes under the process's own arm — the arm is read once into a `OnceLock`, so one process cannot serve two — and writes `corpus / index / input / status / normalized / spdi / blocks_cut` per row; a second invocation diffs two artifacts and writes the summary. It streams both directions, since each full artifact is ~1.7 GB.

Guards it carries, each because the corresponding failure has happened in this repo:

- `partition_switch_startup_error()` is checked at startup, so a misspelled `FERRO_PARTITION` cannot file one arm's numbers under another's name.
- The comparison **refuses** two artifacts taken under the same arm, and refuses a file holding fewer rows than its header plans.
- A run stopped by `--time-budget-seconds` **removes its own artifact**, so a partial pass cannot be mistaken for a census (#1551).
- The population is routed through `CaptureLedger`; the counts are stamped into the artifact as a trailing line — `{"attempted":9949738,"succeeded":9949738,"dropped":0}` on both arms — and the error mode is stamped into the header (`lenient`, the same precondition `measure_spec_conformance_per_arm` uses, so the two censuses are comparable).
- `blocks_cut_available` is recorded, because `partition_blocks_cut` is `#[cfg(debug_assertions)]` and a release run's denominator would be a zero with nothing to do with the corpus.
- Eight unit tests on the span bound and the artifact format. Three of them are cases the bound got wrong in development and that a full pass would have hidden: the repeat *count* (`g.39665402TG[1]`) that made a two-base repeat look 39 Mb wide; the payload accession (`ins[NC_000023.11:g.…]`) whose version digits joined the host's coordinates; and the allele group (`g.[36_37insC;40del]`) whose members were skipped along with the brackets.

This is the consumer-impact sibling of `measure_spec_conformance_per_arm`, which asks a different question — agreement with the *spec's* stated form — over a 934-row corpus. The two must not be added together, and the module header says so.

## How to reproduce

```sh
scripts/fetch-test-fixtures.sh
# `CARGO_PROFILE_DEV_OPT_LEVEL=3` keeps `debug_assertions` (and so the block counter)
# while running at release speed: ~150 s per arm for all 9,949,738 rows.
CARGO_PROFILE_DEV_OPT_LEVEL=3 cargo run --features dev --example measure_flip_corpus_impact -- \
  --output cc.tsv
FERRO_PARTITION=live CARGO_PROFILE_DEV_OPT_LEVEL=3 cargo run --features dev \
  --example measure_flip_corpus_impact -- --output live.tsv
cargo run --features dev --example measure_flip_corpus_impact -- \
  --compare-a live.tsv --compare-b cc.tsv --summary flip.json
```

## Verification

- Both arms: `EXIT=0`, `wrote 9949738 rows`, ledger `{"attempted":9949738,"succeeded":9949738,"dropped":0}`, headers recording `arm: canonical-coalesced` / `arm: live` and `blocks_cut_available: true`.
- Comparison: `EXIT=0`, `9949738 rows compared`.
- `cargo fmt --all -- --check`: `EXIT=0`.
- `cargo clippy --all-features --all-targets -- -D warnings`: `EXIT=0`.
- `cargo nextest run --features dev --no-fail-fast` on the rebased branch: `EXIT=0`, `10740 tests run: 10740 passed, 153 skipped`. (Both gitignored spec artifacts were regenerated first — `generate_spec_fixture` and `generate_spec_enumeration`; a stale enumeration reads as three `spec_enumeration_tests` failures that have nothing to do with this change.)
- `docs/NORMALIZATION_CONTRACT.md` blessed after the prose edit; `normalization_contract_doc` passes **without** `BLESS_CONTRACT_DOC`, alongside `hgvs_spec_normalization_tests`, `claude_md_adjudication_tables`, `clause_ruling_index` and `generator_completeness` (32 tests, 0 failed).

Representation-Change: none. A new `[[example]]`, prose appended to one ruling record, and the generated document re-blessed from it. Nothing under `src/normalize/`, `src/hgvs/`, `src/spdi/` or `src/project/` is touched, no normalized output moves, and no expectation was re-blessed. It *measures* a representation change that already shipped rather than making one.


nh13:
@coderabbitai pause

PR #1902, opened 2026-08-14
fix(data): populate MappingInfo::transcript_id on the g. to c. path

### LINKED ISSUE #1895: MappingInfo::transcript_id is public and documented but never populated on success
`MappingInfo` is public API — `src/data/mod.rs:35` re-exports it alongside `CoordinateMapper` and `MappingResult` — and its first field is documented as the transcript that was used:

```rust
pub struct MappingInfo {
    /// The transcript used for mapping.
    pub transcript_id: Option<String>,
    ...
}
```

On the success path it is never populated. `CoordinateMapper::genome_pos_to_cds_pos` takes `transcript_id: &str`, builds `MappingInfo::default()`, and deliberately leaves the field `None`:

```rust
// No in-tree consumer reads `MappingInfo::transcript_id`, so it is left
// None: that avoids a `String::from` allocation per call (~400 calls per
// SNP project_all on chr17 fan-out).
```

The reasoning is sound and the measurement behind it is real — the allocation is per call on a hot fan-out path, and the claim checks out in-tree: nothing under `src/` reads `MappingInfo::transcript_id`, and `MappingInfo` is not exposed through the Python bindings at all.

"No in-tree consumer" is the wrong bar for a `pub` field on a re-exported struct, though. An external consumer of the crate reads a documented `Option<String>` named `transcript_id`, gets `None` on every successful mapping, and has no way to tell that apart from "the mapper could not determine one" — the field's doc comment promises information the type never carries.

Worth noting this is not a regression and not something #1877 introduced; that PR only reworded the comment (the preceding `let _ = transcript_id;` became unnecessary once `transcript_id` was used on the refusal path). It is long-standing, and it is filed separately from #1894 for that reason.

### Options

Any of these resolves the mismatch; the choice is a judgement call about the public surface:

- Populate the field and accept the allocation — simplest, and measurable against the same `project_all` fan-out the comment cites, so the cost is not hypothetical.
- Keep it `None` but change the doc comment to say the field is not populated by this path, so the type stops promising what it does not deliver.
- Drop the field from the public struct, if nothing outside the crate is expected to want it. This is a breaking change to a re-exported type.
- Make it lazy or borrowed (`Option<&str>` with a lifetime, or a transcript handle) so the identity is available without the per-call `String` allocation.

nh13:
Closes #1895

`MappingInfo::transcript_id` is `pub` on a struct re-exported from `crate::data`, and its doc comment says it is "the transcript used for mapping". `CoordinateMapper::genome_pos_to_cds_pos` left it `None` on every success, so an external consumer reading a documented `Option<String>` got `None` from every `genome_to_cds` / `genome_to_cds_on_build` / `genome_interval_to_cds` call and had no way to tell that apart from "the mapper could not determine one".

This populates it. The issue offered four options; the reasoning for picking this one is below, including the measurement, because the comment that justified leaving it `None` cited a real cost on a real hot path.

## The asymmetry the issue does not mention

The two directions did not agree. `cds_pos_to_genome_pos` — the c. -> g. half of the same `impl`, on the same accession, ten lines of setup away — has always populated the field, unconditionally:

```rust
let mut info = MappingInfo {
    transcript_id: Some(transcript_id.to_string()),
    ..Default::default()
};
```

So the field was not "never populated"; it was populated in one direction and not the other, and nothing said so. That is worse than either consistent answer, and it removes most of the force of the performance argument: the allocation the g. -> c. path was avoiding is one the c. -> g. path already pays.

## Why not the other options

- **Fix the doc instead.** This would have to document the asymmetry rather than remove it — "populated by `cds_to_genome`, not by `genome_to_cds`" — which is a strange contract to pin deliberately, and it leaves the caller with a field they still cannot use.
- **Drop the field.** A breaking change to a re-exported type, to remove information the caller plausibly wants and which one direction already computes.
- **Borrow it (`Option<&str>` + lifetime, or a handle).** Also breaking — `MappingInfo` and `MappingResult<T>` are both re-exported and would both gain a lifetime — and it is not free either: `CdotMapper` stores its accessions as `String` map keys, so making this allocation-free would mean changing that storage to `Arc<str>` as well. That is a large change to buy back the 15 ns measured below.

## Measurement

The comment cited "~400 calls per SNP `project_all` on chr17 fan-out", so I reproduced that shape rather than asserting the allocation is cheap. Release build, interleaved A/B, two binaries built from the two revisions of the crate and run alternately; every run asserts its own success count (200,000 mappings) and a checksum over the returned c. coordinates, so a crashed or short-circuited run cannot read as a fast one.

**Numerator** — 400 transcripts overlapping one genomic position, one `genome_to_cds` each, repeated 500 times, 10 runs per arm:

| arm | min ns/call | median | `transcript_id` populated |
|---|---:|---:|---:|
| without | 36.49 | 38.09 | 0 / 200,000 |
| with | 51.90 | 53.33 | 200,000 / 200,000 |

So the line costs **+15.4 ns/call**, or +6.2 us per 400-transcript fan-out. In isolation that is a ~42% slowdown of `genome_to_cds`, which is why the original comment is not unreasonable.

**Denominator** — but `genome_to_cds` is not the unit a fan-out repeats. `project_all` runs a whole projection per transcript, so `VariantProjector::project` is one unit of the fan-out. Measured on `benches/projection.rs`'s `plus_projector` (a 9-base single-exon mock transcript with an in-memory provider — i.e. a floor for a real chr17 transcript), 20,000 iterations, 5 runs:

**4,596 ns per projection** (min of 5; runs spanned 4,563-4,629 ns, spread 1.4%).

+15.4 ns against >= 4,596 ns is **+0.33% of a fan-out, or better**. The measurement is recorded on the code so the next person does not have to re-derive it.

## Test

`data::mapping::tests::every_mapping_success_reports_the_transcript_it_used` asserts the field names the requested accession across all six public `CoordinateMapper` entry points, both directions, both strands, and the CDS / 5'UTR / intronic / interval region kinds. Verified as a guard rather than assumed: reverting the production change to `MappingInfo::default()` fails it on the first g. -> c. row (`left: None, right: Some("NM_TEST.1")`), and it passes with the change.

It deliberately covers rows that already passed (the c. -> g. ones), because the field is set in exactly one place per direction and a refactor that splits either one would otherwise be free to drop it on a single arm.

## Scope

No behaviour changes. Nothing in the tree reads `MappingInfo::transcript_id` — the many `.transcript_id` hits in `src/project/projector.rs` and `src/data/projection.rs` are on `TranscriptProjection` / `VariantProjection`, a different struct — `MappingInfo` has no serde derive, and it is not exposed through the Python bindings, so no rendered output moves.

Representation-Change: none. `src/data/` only; populates a public field no code path reads and no output renders.

### DIFF
--- a/src/data/mapping.rs
+++ b/src/data/mapping.rs
@@ -22,6 +22,17 @@ pub struct MappingResult<T> {
 #[derive(Debug, Clone, Default)]
 pub struct MappingInfo {
     /// The transcript used for mapping.
+    ///
+    /// Every [`CoordinateMapper`] method that returns a [`MappingResult`]
+    /// populates this on success, in both directions, with the accession the
+    /// caller asked the mapping to be performed against. (`new`, `cdot` and
+    /// `find_overlapping_transcripts` return no `MappingInfo` at all.)
+    ///
+    /// It stays `Option` for backwards compatibility of a `pub` field, not
+    /// because of the [`Default`] derive above — `String` implements
+    /// [`Default`], so the derive would be satisfied either way. A `None`
+    /// therefore means the value was default-constructed rather than produced
+    /// by a successful mapping (#1895).
     pub transcript_id: Option<String>,
     /// Exon number(s) involved.
     pub exon_numbers: Vec<u32>,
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1904, opened 2026-08-14
fix(normalize): cut an equal-length block by its unique column correspondence

### LINKED ISSUE #1878: canonical-coalesced picks a dup where general.md:56 ranks a substitution above it (1420-v2)
Found while landing #1835 (the `canonical-coalesced` partition default flip). **Disclosed there as a README rule-2 preference miss and pinned with a tripwire; this issue is the fix, and it should close before the release that ships the flip.**

## The case

`TEMPLATE` reference `38-41` is `TTGC`; the result is `ATTG`. So position 38 is a substitution, 39 is unchanged, and 40-41 change.

Both of these are legal descriptions of that one change:

| form | how it spells the change at 38 |
|---|---|
| `TEMPLATE:g.[38T>A;40_41delinsTG]` | substitution |
| `TEMPLATE:g.[37dup;41del]` | duplication |

`recommendations/general.md:56` ranks **(1) substitution above (4) duplication** when a description is possible according to several types. So the first form is the one the spec names.

## What changed

Under `live`, the span spelling `TEMPLATE:g.38_41delinsATTG` normalized to `g.[38T>A;40_41delinsTG]` — the ranked form, and the form #1420 asks for.

Under `canonical-coalesced` it re-derives onto `g.[37dup;41del]` instead. The pair now **converges** (both spellings reach one string, which is an improvement in itself) but converges on the form the spec ranks **lower**.

## Why this is a defect and not a representation choice

`rulings[canonical-form-choice-when-both-legal]` licenses deriving from the resulting sequence and emitting what falls out — but explicitly *subject to every explicit spec tie-break*. `general.md:56` is such a tie-break, and it is stated. The derivation appears not to consult the type ranking after re-partitioning, so it takes whichever form the partitioner reaches first.

## Severity

`rulings[separation-rule-force-modal-or-negation]` grades `general.md`'s modals as README **rule 2** — recommended form, best effort — so this does not block a release under rule 7. But rule 2 outranks maintainer judgment, so it is a defect to fix rather than a preference to keep, and it should close **before** the release that ships the flip.

## Where it is pinned

`tests/it/reported_partition_verdicts.rs`:

- `1420-v2/span` now pins `output: "TEMPLATE:g.[37dup;41del]"`, `five_prime: "TEMPLATE:g.[36dup;41del]"`, `verdict: Verdict::Gap` (was `Canonical`), with the full argument on the row.
- `PAIR_STATES` moves `1420-v2` from `OneReaches` to `NeitherReaches` — after the flip neither spelling reaches `wanted`.
- `PAIR_STATE_CENSUS` `(0, 6, 3)` to `(0, 5, 4)`; `OPEN_GAPS` 12 to 13.

`tests/it/reported_confluence_pairs.rs` carries `the_1420_v2_pair_does_not_converge_by_re_derivation`, a tripwire written in advance for this exact hazard.

**Closing this issue should return `1420-v2/span` to `Canonical` and the pair to `OneReaches`, lowering `OPEN_GAPS` back to 12.** Those pins are the tripwire: any change that moves them must name the clause that carried it.

## Likely blast radius beyond this row

`1421-n1/span` moves onto `TEMPLATE:g.[28_29insAA;32G>A;33_34insACTG]`, a third form neither spelling authored, and the same "ranking not consulted after re-partitioning" mechanism is the obvious suspect. Worth checking whether one fix closes both.

nh13:
Closes #1878.

## Rebased onto the post-#1899 base — what moved, and the one thing that blocks it

Every figure this PR carried was measured before #1899 (`delete the input-relative weight bound and derive the split cap`) merged. All of them have been re-measured at the new base and are replaced below. Two results, and they point in opposite directions.

**The good one: the convergence give-back this PR disclosed is gone.** It was attributed here, measured, to the input-relative weight bound rather than to this rule, and #1899 deleted that bound. Re-measured on the resulting base, every census returns to the value `main` pins without this change, and one rises above it. That attribution is now confirmed by the experiment it predicted rather than asserted.

**The blocking one: this PR's proposed ruling is in head-on conflict with `rulings[unchanged-is-read-over-every-minimal-alignment]`, which #1899 also landed as `decided`.** That record holds a reference base unchanged iff **every minimal alignment** matches it, states in terms that "equal length does NOT imply no indel", works `CAG -> AGA` as a balanced deletion-plus-insertion whose two middle bases are therefore unchanged — and names as false the exact premise this PR rests on, that an equal-length block has nothing to choose. It even says so about the code: the three `merge.rs` guards keyed on `reference.len() != result.len()` "exempt precisely the class that breaks them, which is why `FERRO_PARTITION=live` gives the position-wise answer here and the canonical family does not."

#1878's case is that same shape, not a different one. `TTGC -> ATTG` is a balanced insertion-plus-deletion at distance 2 read against a position-wise correspondence at distance 3, exactly as `CAG -> AGA` is. So this rule cannot be narrowed clear of the collision without becoming a no-op on the case it was written for.

Measured, not argued — with this change in place, two of that record's own named guards go red:

| guard | this branch answers | the record requires |
|---|---|---|
| `normalize_reparse_invariant::a_del_beside_a_dup_re_spells_instead_of_colliding` | `NC_TEST.1:g.261_263delinsAGA` | `NC_TEST.1:g.[261del;263_264insA]` |
| `issue_1284_transcript_axis_collision::a_noncoding_del_dup_collision_is_repaired` | `NR_TEST.1:n.8_10delinsAGA` | `NR_TEST.1:n.[8del;10_11insA]` |

**The repository currently says both things in two places.** `README.md`'s ruleset — unchanged on `main`, and the authority this PR cites — still states "For an equal-length block the column correspondence is unique, so 'two variants separated by N nucleotides' is a fact about the variant." And `main`'s own `weight_bound_worked_examples::a_net_zero_insert_delete_pair_is_left_to_the_per_member_pipeline` invites this PR's reading by name on a block of exactly the contested class: "the block is equal-length (11/11) with 9 columns changed, so `DNA/delins.md:16` types each run as a delins and `:17` keeps them separate. Flip this expectation and record the adjudication." That test also goes red here, with precisely the form its message names.

That is a `decided` record on `main` against a proposal on a branch, so precedence is not in doubt: **this PR cannot land until the operator either overturns `unchanged-is-read-over-every-minimal-alignment` or withdraws this record.** The new record and its `CLAUDE.md` row now state the conflict rather than leaving it to be found. The value pins the give-back's disappearance invalidates have deliberately **not** been re-blessed, because re-blessing them would bank a rule the ledger currently answers the other way.

## The defect, and why it is not the one that was filed

`TEMPLATE` reference `38-41` is `TTGC`; the result is `ATTG`. Under `live` the span spelling `TEMPLATE:g.38_41delinsATTG` normalized to `g.[38T>A;40_41delinsTG]`; under `canonical-coalesced` it re-derives onto `g.[37dup;41del]`.

#1878 argues that `general.md:56` ranks (1) substitution above (4) duplication, so the first form wins. **That argument does not hold, and this PR deliberately does not use it.** Both competitors are multi-member alleles, and `rulings[conflicting-member-geometry-refusal-scope]` (decided) records in terms that `:56` "ranks competing descriptions of ONE span; it says nothing about two members of one allele", naming that citation as this repository's recorded cautionary error; `rulings[inversion-vs-a-mixed-member-competitor]` (decided) relies on that holding rather than extending it. The comparison is undefined member-wise anyway — `40_41delinsTG` is a `delins`, which is absent from `:56`'s list.

**The ground offered is that the block is equal-length.** Four reference bases against a four-base payload have exactly one position-wise correspondence. Under it 38 changes alone (`DNA/delins.md:15`, a substitution), 39 does not, and 40-41 change consecutively (`:16`, one `delins`), which `:17` keeps individual — `g.[38T>A;40_41delinsTG]`, verbatim. The compensating insertion-plus-deletion the canonical arm found is cheaper by edit distance, but minimal edit distance is ferro policy and no clause (`rulings[canonical-form-choice-when-both-legal]` records that no minimality principle is stated). That the surviving form also happens to rank higher under `:56` is a consequence, not the reason.

New record: `rulings[equal-length-block-column-correspondence-is-unique]`, governing `DNA/delins.md:16` — **and see the section above before reading it as settled.**

## Where it is implemented, and why there

One place: `partition_block_canonical_within` returns the position-wise pairing when `reference.len() == result.len()`, before building the alignment DAG. That is the block partitioner — the single point where the arm chooses an alignment — so the rule is stated once and every caller of the canonical family (the dispatch path and `from_sequences`' derivation surface) gets it.

It is also where the scoping was already written down for the other arm. `partition_block` — `live`, shipped through v0.13.x — scopes three of its own rules away from this regime in these words: *"an equal-length block has no gap to place, so `best_alignment` compares position-wise there"*. The canonical arm never carried that scoping, and #1835 made it the default. `rulings[unchanged-is-read-over-every-minimal-alignment]` is the record that calls those three scopings' premise false, which is the whole of the conflict above.

## Representation change, re-measured at the new base

Corpus is `examples/dump_normalized_corpus.rs`, **85,642 rows**, unchanged in size by #1899. Both arms of every comparison were dumped fresh; nothing is carried forward.

**All 2,426 moved rows land byte-identically on the `live` arm's answer.** Measured by set arithmetic, not sampled: `live` -> `main` differs on 7,808 rows of the same corpus and `live` -> this branch on 5,382, a difference of exactly 2,426, and every one of the 2,426 is `live`'s string with **zero** landing on a third form. None was a fixed point beforehand. Since #1835 is unreleased, no consumer holds the forms being moved off.

The two `live`-distance figures are the ones #1899 moved: they read 8,999 and 6,573 before it landed, so #1899 by itself brought `main` 1,191 rows closer to `live`. The 2,426 this PR moves is unchanged by it, as is the per-family split.

| family | rows | moved |
|---|---|---|
| `delins_hiding_an_inversion` | 8,640 | 1,516 |
| `inv_member` | 9,216 | 676 |
| `three_member_allele` | 4,608 | 234 |
| every other nucleotide family (13 of them) | 55,834 | 0 |
| the five `protein_*` families | 7,344 | **0, and structurally so** |

The protein zero is **structural** and must be reported as such: `canonicalize_from_sequence` does not serve the protein axes at all, so no protein row could move whatever the rule said. The families that do move are the ones whose generator can emit an equal-length changed block; the property this change keys on is block length equality, and the generator does vary it.

**Inversion recognition improves.** Rows whose output contains an `inv` rise **11,443 -> 12,201** of 85,642 — 758 rows gain one and **none** loses one. The figure this PR previously carried, "1,487 -> 1,602", is **withdrawn rather than updated**: it is not reproducible under any counting this re-measurement could construct (rows, distinct output strings, a single direction, or newly-recognised `inv` alone), while the moved-row set it was taken alongside reproduces exactly. Quote the definition with the number.

## What it costs — re-measured, and the answer is now "nothing"

Every census below is the value measured on this branch at the new base, against the value `main` pins.

| census | `main` pins | this branch measures | |
|---|---|---|---|
| `cis_confluence_axis` 3' `converged` (of 11,272) | 11,271 | **11,271** | unchanged |
| `cis_confluence_axis` 5' `converged` | 11,272 | **11,272** | unchanged |
| `cis_confluence_nr_axis` n. 3' / n. 5' / r. 3' / r. 5' | 5,635 / 5,636 / 5,633 / 5,634 | **5,635 / 5,636 / 5,633 / 5,634** | unchanged |
| `spec_conformance_axis` 3' `converged` | 11,016 | **11,016** | unchanged |
| `spec_conformance_axis` 5' `converged` | 10,753 | **10,773** | **+20** |
| `reported_confluence_pairs` converging pairs, both directions | 9 | **9** | unchanged |

The previously disclosed give-back — 11,271 -> 10,773, 11,016 -> 10,521, 10,753 -> 10,282, 9 -> 6 — **does not occur at this base.** The pins this PR lowered to record it are now wrong in the other direction and are left red rather than re-blessed, for the reason in the first section.

**Every rank-1 figure is unchanged in both directions** — `declined` 0, `unparseable_outputs` 0, `outputs_denoting_no_sequence` 10 at 3' / 18 at 5', `prohibition_violating_outputs` 0, `non_idempotent_outputs` 4, `sequence_changed` 4 at 3' / 0 at 5' — so nothing conformant was traded for the movement, and `cis_confluence_*` keeps `sequence_changed` and `declined` at 0 throughout.

**The reference-aware axis is unmoved.** `multi_member_cis_axis::axis_normalized` passes with a prepared manifest armed (no CI job sets `FERRO_MANIFEST`, so this was run by hand): 592 rows, 6 declined, 90 unwindowed, 461 respellable, 407 respelling-converged, 0 sequence-changed, 0 non-idempotent.

## What #1899 fixed that this PR had disclosed as a cost

Three adjudicated expectations now reach the form their record names, and the pair census reaches its fixed state — all of them consequences of #1899's bound deletion arriving, and all sanctioned in advance by the text of the guards that now fail:

- `cis_confluence_adjudication::two_adjacent_members_that_both_consume_reference_are_one_delins` — the authored cis spelling now derives to `NM_TEST.1:c.10_13delinsTAAT` too, so the class **converges on the adjudicated-correct form**. This PR's own comment predicted exactly that ("with the bound disabled this line reads `spanning` and the class converges ON the adjudicated form. Measured, not predicted.").
- `cis_adjudication_enumeration::the_adjacent_members_class_partitions_as_measured` — all 1,076 spellings now reach one output rather than 1,040 / 36.
- `reported_confluence_pairs` reads **9 of 9** converging, with `1420-v2` reaching `TEMPLATE:g.[38T>A;40_41delinsTG]` — #1420's own wanted form — from **both** spellings, so `reported_partition_verdicts` measures the pair as `BothReach`.
- `1420-v3` and `1420-v4` likewise reach their issues' wanted forms.

These are re-pins the guards themselves ask for by name. They are **not** applied here, for the same reason as above: the rule that produces them is the one the ledger currently contradicts, so banking them now would be banking a re-derivation.

## Two things that go the other way, named rather than buried

- `inversion_sweep`: one row of 2,075, `NM_004006.2:c.759_766inv`, goes back to the four lone substitutions `c.[759G>T;761T>G;764C>A;766A>C]` — its released value, byte-identical to the pre-#1835 `cases.tsv` line. That census (`1490, 466, 26, 93`) still passes at the new base. It is the only row of that sweep either change moves, in either direction, and it runs against the corpus-wide `inv` gain above.
- `cis_junction_crossing_shift::the_five_prime_junction_barrier_does_not_over_clamp` is a third guard in the contested class: `TEMPLATE:g.[2_3insA;12del]` reads here as `TEMPLATE:g.[3T>A;10A>T;12T>A]`. It belongs with the two in the first section, not with this PR's own pins.

## The projector divergence this PR used to re-open is closed at the new base

`project::projector::tests::project_cis_net_in_frame_but_mid_shift_stop_takes_the_frameshift_arm` passes unchanged from `main`. This PR previously re-pinned it as a re-opened divergence (`p.[(Val2Ter);(Lys3=)]` against `p.(Val2Ter)`); at the new base the #1616 dispatch fix — `CisResidueOutcome::TerminatorMoved`, on `main` since #1899 — answers for both spellings even when the block cuts into three isolated substitutions, so the divergence does not re-open. **That re-pin is reverted to `main`'s equality assertion**, and issue #1905, filed for it, should be re-read against this before anyone works it.

## Verification

`cargo fmt --all --check` clean; `cargo clippy --all-features --all-targets -- -D warnings` clean; `pytest tests/python/` green against a `maturin develop --features python` build (1,000 passed, 8 skipped) — run because `cargo nextest` does not reach the bindings and this change is under `src/normalize/`. `docs/NORMALIZATION_CONTRACT.md` re-blessed for the amended record and passes without `BLESS_CONTRACT_DOC`; the `clause_ruling_index` generated block was regenerated from `render_index()` and its census pin moved to `(32, 29, 3)` with the multiply-named clause count to 35. `tests/it/msto_regression_corpus.rs` keeps #1899's delisting of `#1420`'s `reported_confluence_pairs` guard exactly as the operator left it; the surviving half of that guard is renamed here and deliberately not re-listed.

The full Rust suite is **red at 22 tests**, and every one of them is named above: 6 confluence pins whose give-back no longer occurs, 2 `spec_conformance_axis` pins (one unchanged, one improved), 9 arrivals of #1440's effect that the guards ask to be re-pinned, 3 in the contested class, 1 (`weight_bound_worked_examples`) whose own message invites the flip, and 1 (`in_bounds_oracle_scope::oracle_probe_normalizes_the_transcript_exit_under_arming`) which reports `FERRO_ASSERT_IN_BOUNDS` missing in a child it re-execs and is not attributed to this change. None of them is re-blessed pending the adjudication.

Representation-Change: 2426 of 85642 corpus rows move, 0 previously-accepted and 0 previously a fixed point. Every one lands byte-identically on the `live` arm's answer — the form shipped before this change — and none on a third form, so relative to the last release this is give-back rather than migration. Movement is confined to three shape families (`delins_hiding_an_inversion` 1516/8640, `inv_member` 676/9216, `three_member_allele` 234/4608); the five `protein_*` families read a STRUCTURAL zero because `canonicalize_from_sequence` does not serve the protein axes. Rows whose output contains an `inv` rise 11443 to 12201 of 85642, 758 gained and 0 lost. Convergence no longer falls: re-measured after #1899 deleted the input-relative weight bound, `cis_confluence_axis` converged reads 11271 (3') and 11272 (5'), `cis_confluence_nr_axis` 5635/5636/5633/5634, `reported_confluence_pairs` 9 of 9 — all equal to the values main pins — and `spec_conformance_axis` converged returns to 11016 at 3' and RISES to 10773 at 5' against main's pin of 10753. Every rank-1 figure is unchanged in both directions and the reference-aware `multi_member_cis_axis::axis_normalized` census is unmoved.


nh13:
@coderabbitai pause

nh13:
Closing: the adjudication this PR asks for has since been made, in the opposite direction, by #1937.

This PR's premise is that the repository states the uniqueness rule in two places and one must be chosen. Its cited non-ledger authority is a sentence in `README.md`'s ruleset, quoted in the PR body, in the new ledger record's rationale, in the `CLAUDE.md` row and in the `merge.rs` comment — three of those asserting it is current:

> For an equal-length block the column correspondence is unique, so "two variants separated by N nucleotides" is a fact about the variant.

That sentence is no longer on `main`. #1937 (`563fb525`, *"docs(readme): equal block length does not make the column correspondence unique"*) was merged 2026-08-14 23:11 UTC — about five and a half hours after this branch's last push — and replaced it with text that states the opposite and points at the record this PR conflicts with:

> **Block length does not settle it**: an equal-length block can still carry a balanced del+ins pair, so its column correspondence need not be unique — `CAG -> AGA` is equal length with edit distance 2, not the position-wise 3. What decides whether a reference base is unchanged is whether **every minimal alignment** matches it, so the property to key on is edit distance against block length, never length alone. See `rulings[unchanged-is-read-over-every-minimal-alignment]`.

So there is no longer a tie to break. Both places now say one thing, and `CAG -> AGA` is a direct counterexample to the proposed rule's premise. Landing the record as `status: decided` would put a claim on `main` that `main`'s own ruleset refutes by counterexample.

Closing rather than holding, because the disagreement this PR was opened to resolve no longer exists in the form it describes. The engineering was careful and the disclosure was better than this repository's norm — the citation simply expired between the measurement and the review.

If the rule is still worth making, it wants a fresh argument built on the current `README.md` text and on `rulings[unchanged-is-read-over-every-minimal-alignment]`, rather than a rebase of this one. Several findings from the review are worth carrying into that work regardless of which way it goes:

- two ledger `clause` citations are off by one (`general.md:34`/`:56` should be `:33`/`:55`, quote-verified against the pinned spec checkout);
- the new record's machine-readable `status` reads `decided` while its own prose calls itself a proposal;
- the `clause_ruling_index` census pin merges cleanly to a wrong value, which no textual check can see;
- the trailer's "0 previously-accepted" figure needs re-deriving;
- `tests/fixtures/inversion-sweep/cases.tsv` is manifest-generated, so its edit is not verifiable on PR CI.

Refs #1937, #1878.

PR #1906, opened 2026-08-14
fix(normalize): report REFSEQ_MISMATCH at a coordinate, not a window offset

### LINKED ISSUE #1612: REFSEQ_MISMATCH reports a window-relative position as an HGVS coordinate
`REFSEQ_MISMATCH` reports a window-relative coordinate as if it were an HGVS coordinate.

## What happens

`NC_000001.11:g.201C>T` against a contig whose base 201 is `A` reports the mismatch at **`100-100`**:

```
warning[REFSEQ_MISMATCH]: reference sequence mismatch at 100-100: stated "C", actual "A" (corrected=false): Reference mismatch: stated 'C' but actual is 'A'
```

Measured directly, not inferred — the fixture is a 400-base `ACGT`-cycling contig, so `base_at(201) == 'A'` and the correct span is `201-201`.

## Where

`src/normalize/mod.rs:8806`:

```rust
warnings.push(NormalizationWarning::RefSeqMismatch {
    ...
    position: format!("{}-{}", start, end),
```

`start`/`end` there are the coordinates of the **fetched reference window**, not the description's own axis. Nothing converts them back before the string reaches a user.

Contrast the cis-allele producer at `src/normalize/merge.rs:3362`, which is explicitly documented as reporting "the sequence-axis span, via `region_sequence_delta`" and re-bases its validator call to `1`/`span` precisely so the two verdicts agree. The per-member path has no equivalent step.

## Why it matters more than it used to

The field is not diagnostic-only. It is also the `location` of the hard error in strict mode, via `src/normalize/mod.rs:2418`:

```rust
NormalizationWarning::RefSeqMismatch { position, .. } => Some(FerroError::ReferenceMismatch {
    location: position.clone(),
```

So a caller that parses `location` to point at the offending base is pointed at the wrong one. #1580 moved `ferro normalize` onto `normalize_with_diagnostics`, which makes the warning reach the CLI in lenient/silent mode as well — the coordinate was already wrong, it is now visible on more paths.

## Scope note

This is deliberately **not** fixed in #1580, whose scope is the warning seam rather than the warning's contents. `tests/it/normalize_warning_seam.rs` records the defect and pins everything about the message *except* the position, so a fix here will not have to fight a frozen wrong value.

Fixing it should account for:

- every `RefSeqMismatch` message, not just the CLI's rendering;
- the strict-mode `FerroError::ReferenceMismatch` text, which is user-visible and asserted in tests;
- agreement with the cis producer in `merge.rs`, whose warnings are built to be indistinguishable from the per-member ones so duplicates can be dropped — if only one side is corrected, that dedupe (`mod.rs:2854`, keyed on `position`) silently stops matching and the same finding gets reported twice.

nh13:
Closes #1612

`REFSEQ_MISMATCH` reported the offset into the **fetched reference window** as if it were a coordinate. Against a contig whose base 201 is `A`, `NC_000001.11:g.201C>T` said:

```
warning[REFSEQ_MISMATCH]: reference sequence mismatch at 100-100: stated "C", actual "A" (corrected=false)
```

`100` is `201 - window_start`. The mismatch is real; only the coordinate was wrong. It now reads `at 201-201`.

## What the field promises, and why the transcript axes do not move

`normalize_na_edit` indexes the slice it is handed, so its `start`/`end` are always in that slice's frame. Most callers hand it the whole sequence the description addresses; the genomic and mitochondrial callers hand it a window, and nothing converted the pair back.

The promise the field already carries — stated and implemented by the cis producer, `merge::authored_reference_mismatch` ("the sequence-axis span, via `region_sequence_delta`") — is **the 1-based inclusive span on the reference sequence that was actually read, in that sequence's own numbering**. It is deliberately *not* the description's own axis, and the two producers must agree because the duplicate-suppression in `normalize_core` is keyed on this string and nothing else.

So `c.5C>T` on a transcript whose CDS starts at 21 reports the transcript offset `25-25`, on both producers, before and after this change. Re-spelling it `5-5` would have been a second wrong answer and would have broken the dedupe; it is now pinned so a later change cannot "fix" the two apart.

## Per axis

| axis | sequence read | reported | changed |
|---|---|---|---|
| `g.`, `m.` | the contig | `201-201` — the contig coordinate, which is also the description's own | **yes**, was the window offset |
| `c.`, `n.`, `r.` (exonic) | the transcript | `25-25` — the transcript offset | no, already correct |
| `c.`, `n.` intronic / boundary-spanning | the genomic **contig** | `chr_i:314-314` — the genomic span, qualified by the contig accession | **yes**, was a window offset in a reverse-complemented frame on minus-strand transcripts |

The intronic row is the case where a faithful conversion to the description's own axis is not available: the bases are read from a contig the accession does not name, and the shuffle frame is reverse-complemented on a minus strand. A bare number there would be a coordinate in a space the reader has no way to guess, so that arm names the space. Nothing dedupes against it — `merge::simple_cds_pos` declines every offset position, so no cis-side duplicate of an intronic finding can exist.

A new `MismatchFrame` makes every caller state where its span sits, so the conversion happens where the answer is known. It is a parameter rather than something derived inside `normalize_na_edit`, because that function cannot see whether `ref_seq` is a whole sequence or a window — which is exactly how the wrong coordinate shipped.

## Three things this repairs, not one

1. The coordinate itself.
2. The `location` of strict mode's `FerroError::ReferenceMismatch`, which is the same string. A caller parsing it to point at the offending base was pointed at the wrong one.
3. **The dedupe against the cis producer, which had silently stopped matching on the genomic axis.** `NC_000001.11:g.[201C>T;300G>A]`, two members each misstating its own base, produced **four** warnings — the cis producer's `201-201` and `300-300`, plus two per-member warnings *both* reading `100-100`, because each member is windowed about itself. Neither real finding matched the fabricated ones, so every mismatch was reported twice, and the two fabricated coordinates were on their face indistinguishable from each other. It now produces two.

## The sibling defect

`validate_reference` built the same window offsets into the warning's `actual_ref` text, in HGVS range notation. Repairing `position` on its own left one warning contradicting itself — measured on `g.201_202del5`:

```
at 201-202: stated "length 5", actual "span 2 (100_101)"
```

Three messages carried a raw index (`validate_stated_length`'s `(start_end)`, and the "position N out of range" / "inverted range: start N > end M" strings). The endpoints are `position`'s job and it now states them correctly, so the restatements are dropped rather than duplicated; the width, which is what the length check is about, stays (`actual "span 2"`). No test pinned any of those strings.

`CanonicalSplitSkipped` is the other warning built nearby and is **not** affected: it is constructed in `apply_canonical_split` from `fetch_ref_for_canonical_split`'s genuine HGVS coordinates, not from `normalize_na_edit`'s relative pair.

## Guard

`tests/it/issue_1612_refseq_mismatch_coordinate.rs` — 10 tests covering the issue's reproducer, the strict-mode `location`, `m.`, the genomic and coding cis dedupes, the transcript axes staying put, both intronic strands, and the length message. Verified as a real guard by reverting each production line and re-running: reverting `position: frame.render(start, end)` fails 7 of the 10 (the 3 that pass are the fixture check and the two transcript-axis tests, which are correctly unaffected); reverting the `span {span}` message fails the 10th.

The intronic coordinates are cross-checked independently of my own arithmetic: the warning's `actual_ref` must be the base standing at the coordinate the warning names — the contig base on the plus strand, its reverse complement on the minus strand.

`tests/it/normalize_warning_seam.rs` recorded this defect deliberately and left the position unpinned so a fix would not have to fight a frozen wrong value. That comment is replaced and the CLI-level assertion now pins `at 201-201`.

## Verification

`cargo fmt --all --check`, `cargo clippy --all-features --all-targets -- -D warnings`, `cargo nextest run --features dev --no-fail-fast` (10757 passed, 0 failed, post-rebase), and the Python suite via `maturin develop --features python` + `pytest tests/python/` (991 passed, 8 skipped). Exit statuses read back from each run's own log rather than from a completion notification.

Representation-Change: The emitted WARNING SET changes on cis alleles; no normalized output moves (0 rows respell).
  This is NOT a diagnostic-only change, and the previous `none` verdict was wrong.
  `ValidationResult::valid` is untouched and the new `frame` argument is read at exactly one place — where the
  warning string is built — so no normalization decision and no normalized string moves. What moves is the
  set of warnings a consumer receives, plus the text of the ones that survive.
  (1) WARNING SET. The `REFSEQ_MISMATCH` `position` string is the sole key of the cis-allele duplicate
  suppression (`src/normalize/mod.rs:4160-4182`), so repairing the coordinate changes how many warnings
  survive. Measured on `NC_000001.11:g.[201C>T;300G>A]` over a 400-base ACGT-cycling contig: 4 warnings
  before (`["100-100","100-100","201-201","300-300"]`), 2 after (`["201-201","300-300"]`) — pinned by
  `each_genomic_cis_member_is_reported_exactly_once`. A consumer reading `ferro normalize -f json`'s
  `warnings` array, or the Python `NormalizationWarning` list, sees two entries where it saw four.
  (2) TEXT, on five surfaces beyond `Display`, four of them machine-readable: strict mode's
  `FerroError::ReferenceMismatch.location`; the serde-serialized benchmark record `RefMismatchInfo.position`
  (`src/benchmark/types.rs:94-103`); the Python bindings' `PyNormalizationWarning.message`
  (`src/python.rs:2446-2456`, public API); `ferro normalize -f json`'s `warnings` array
  (`src/bin/ferro.rs:1804`); and `ferro project`'s `ProjectionWarning.message` (`src/cli/project.rs:54-59`),
  which participates in `AxisOutcome`'s derived `PartialEq`/`Eq` (`:77`). Enumerated by route rather than by
  consumer. No route branches on or substring-matches the text, and grep over `src/` and `tests/` finds no
  test or fixture pinning the old wording.
  Not claimed from suite greenness. With the three `validate.rs` message changes reverted, 228 tests ran and
  228 passed (EXIT=0) — nothing guards them, so a green suite is no evidence about them, and two of the three
  remain unguarded after this PR.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -5229,6 +5229,97 @@ struct WindowedNormalization {
     warnings: Vec<NormalizationWarning>,
 }
 
+/// Where the 1-based span [`Normalizer::normalize_na_edit`] is handed sits,
+/// relative to a coordinate a reader can act on.
+///
+/// That function indexes `ref_seq` from `start`, so its positions are always in
+/// the frame of the *slice it was given*. For most callers the slice is the
+/// whole sequence the description addresses and the two coincide; for the
+/// callers that fetch a window they do not, and a diagnostic built from the raw
+/// numbers reports the window's own offsets as if they were coordinates
+/// (#1612 — `NC_000001.11:g.201C>T` reported `100-100`).
+///
+/// # What the rendered coordinate promises
+///
+/// **The 1-based inclusive span on the reference sequence that was actually
+/// read**, in that sequence's own numbering — not the description's axis. That
+/// is the promise the cis producer already documents and implements
+/// (`merge::authored_reference_mismatch`, "the sequence-axis span, via
+/// `region_sequence_delta`"), and the two must agree because the dedupe in
+/// `normalize_core` is keyed on nothing but this string. So a `c.5C>T` on a
+/// transcript whose CDS starts at 21 reports `25-25`, the transcript offset,
+/// on both producers — that is deliberate and is not what #1612 is about.
+///
+/// On `g.`/`m.` the sequence read is the contig, so the span *is* the
+/// description's own coordinate. On `c.`/`n.`/`r.` the sequence read is the
+/// transcript, so it is the transcript offset. In both cases the accession is
+/// the one the description names, and the bare `<start>-<end>` form is
+/// unambiguous.
+///
+/// [`Self::GenomicWindow`] is the case where it is not: an intronic or
+/// boundary-spanning transcript description is shuffled against the **genomic
+/// contig**, which is not the sequence its accession names. A bare number there
+/// would be a coordinate in a space the reader has no way to guess, so that
+/// arm qualifies it with the contig accession. Nothing dedupes against it — the
+/// cis producer declines every offset position (`merge::simple_cds_pos`), so no
+/// duplicate of an intronic finding can exist to match.
+#[derive(Debug, Clone, Copy)]
+enum MismatchFrame<'a> {
+    /// The slice is the whole sequence the description addresses, so the span
+    /// is already the coordinate to report. Every transcript-axis caller.
+    SequenceAxis,
+    /// The slice is a window opening `window_start` bases into that sequence
+    /// (0-based), so `sequence = window_relative + window_start` — the same
+    /// reconstruction [`WindowedNormalization`] documents for the *result*
+    /// positions, applied to the reported ones.
+    Window { window_start: u64 },
+    /// The slice is a window on the genomic contig `accession`, opening at
+    /// 1-based `seq_start`, and flipped into transcript view when `strand` is
+    /// minus. The reported coordinate is genomic and says so.
+    GenomicWindow {
+        accession: &'a str,
+        seq_start: u64,
+        strand: Strand,
+        window_len: u64,
+    },
+}
+
+impl MismatchFrame<'_> {
+    /// Render the 1-based inclusive slice-relative span `[start, end]` as the
+    /// coordinate a reader should be shown.
+    fn render(&self, start: u64, end: u64) -> String {
+        match *self {
+            Self::SequenceAxis => format!("{start}-{end}"),
+            Self::Window { window_start } => format!(
+                "{}-{}",
+                start.saturating_add(window_start),
+                end.saturating_add(window_start)
+            ),
+            Self::GenomicWindow {
+                accession,
+                seq_start,
+                strand,
+                window_len,
+            } => {
+                // The same unflip the caller applies to the *result* positions,
+                // so the two cannot drift. Clamped rather than trusted: the
+                // helper subtracts from `window_len`, and every span reaching
+                // here is in-window by construction (`flip_intronic_for_strand`
+                // refuses otherwise), but a clamp costs nothing and a debug
+                // overflow panic in a diagnostic path costs a lot.
+                let (lo, hi) = unflip_intronic_positions(
+                    strand,
+                    window_len,
+                    start.min(window_len),
+                    end.min(window_len),
+                );
+                let genomic = |rel: u64| rel.saturating_add(seq_start).saturating_sub(1);
+                format!("{accession}:{}-{}", genomic(lo), genomic(hi))
+            }
+        }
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

Issue #1610, opened 2026-08-10
normalize: a lone unequal-length `delins` is split into members, against `DNA/delins.md:47`


nh13:
Ferro splits a **lone, minimal, unequal-length** `delins` into members, which is the description `DNA/delins.md:47` recommends against and `:46` calls the *alternative*.

## Reproduction

Hermetic, `main` at `7118eb2a`, via `Normalizer::normalize` with a synthetic transcript provider:

```
NM_TEST.1:n.2_5delinsAAC   ->   NM_TEST.1:n.[2_3delinsAA;9del]
```

Four reference bases against a three-base payload. Ferro finds an alignment in which part of the payload coincides with the reference, cuts there, and 3'-shifts the residual deletion out to position 9 — so the output asserts two variants a long way apart where the input asserted one contiguous change.

## Why this is a defect and not a preference

The governing record is **decided** and already in the ledger: `delins-merge-vs-individual-gap-two-or-more`, `governing: docs/recommendations/DNA/delins.md:47`.

`:46` is explicitly about this shape — "parts of the inserted sequence *align* with the reference sequence, giving an alternative description like `c.[850_869del;874_881del;887_897del;901_902insG]`" — and `:47` answers it: "**The 'delins' format is recommended**: it is simpler and prevents software tools making incorrect predictions for the consequences on protein level."

So the spec names the decomposition as the *alternative* and recommends the span. Ferro emits the alternative.

Note the record's own scope caveat cuts the right way here: it scopes `:44-47` to "the alignment-coincidence shape `:44-47` describes", and a lone unequal-length `delins` whose payload partly matches the reference **is** that shape. `general.md:34` does not reach it, because `:34` speaks of "two variants separated by" and the input asserts one.

## The equal-length path must not be touched by the fix

This is the part most likely to go wrong, so it is worth stating before anyone writes a patch. **Net length is the discriminator, and it is invisible in the two strings.**

- **Equal length** — the reference span and the payload have the same length, so there is exactly one column-wise correspondence and "these columns changed, that one did not" is a fact about the two sequences. `DNA/delins.md:17` then *requires* the split, and ferro is already correct there. A fix that widens to the equal-length path would break conformant behaviour.
- **Unequal length** — there is no column-wise correspondence at all. Locating an interior unchanged run means first *choosing* an alignment, and several alignments are equally good. That choice is what manufactures the split, and it is not determined by the two sequences.

[...]

PR #1908, opened 2026-08-14
fix(normalize): keep a lone unequal-length delins whole

### LINKED ISSUE #1878: canonical-coalesced picks a dup where general.md:56 ranks a substitution above it (1420-v2)

nh13:
Addresses #1610 — the `c.` half. The issue's sole reproduction, `NM_TEST.1:n.2_5delinsAAC`, is on the `n.` axis, which sits outside the coding-DNA carve-out this PR gates on, so it still returns `n.[2_3delinsAA;9del]`. That half stays open and carries a committed guard asserting the defect.

## The defect

```
NM_TEST.1:n.2_5delinsAAC   ->   NM_TEST.1:n.[2_3delinsAA;9del]
```

Four reference bases (`CGCG`) against a three-base payload (`AAC`). The only interior "unchanged" column is the payload's own `C` landing on a reference `C`; ferro cut there and 3'-shifted the residual deletion out along a `G` run to `n.9`, so the output asserts two variants seven bases apart where the input asserted one contiguous change. It reproduces on **every** partition arm — `live`, `canonical` and the shipped `canonical-coalesced` — so it is not a consequence of #1835's default flip, and it reproduces on `c.` as well as on `n.`.

## Where the rule lives, and why there

`merge::split_is_a_placed_gap_coincidence`, consulted by `partition_block` (the `live` arm's own escape ladder) and by `partition_block_for_rule` (the sequence-first arms). It is deliberately **not** inside `partition_block_canonical_within`: that function also serves `partition_block_for_derivation`, whose rule is pinned by `DERIVED_BLOCK_PARTITION_RULE`, and `from_sequences` is out of scope for #1610.

`partition_block` already collapsed this regime — an unequal-length block whose separations are all a single unchanged base — when **every** derived member rendered as a `delins`. It declined as soon as one member was a pure deletion, on the ground that a `del` is a higher-priority type than a `delins`. That reading treats the deletion as a variant the sequences separated. On an unequal-length block it is *the gap the aligner placed*, and where it lands is the same choice that manufactured the separation. `DNA/delins.md:46` constructs exactly this description — and its published alternative, `c.[850_869del;874_881del;887_897del;901_902insG]`, is mostly `del` members — while `:47` recommends the span over it.

**The ground is `README.md`'s own ruleset, not a new reading of `general.md:34`.** That ruleset lists, among the reasons rules 2 and 3 are best effort: *"For an equal-length block the column correspondence is unique, so 'two variants separated by N nucleotides' is a fact about the variant. For an unequal-length block there is no correspondence — recovering one means choosing an alignment, and the spec does not say which."* `general.md:34` and `DNA/delins.md:17` are conditionals whose antecedent is "two variants separated by one or more nucleotides"; on an unequal-length block that antecedent is not established but manufactured, so what remains is a **rule 6** choice among conformant forms, disclosed under rule 7. This must never be cited as a conformance finding.

The issue's own phrasing — "`:34` does not reach it, because the *input* asserts one" — is spelling-relative and `canonical-form-choice-when-both-legal` excludes that outright. The ruling is the spelling-independent form of the same argument: the antecedent fails because the *sequences* do not determine a separation.

### The four conditions, each with the negative control that fixes it

1. **Net deletion.** The direction scope of `delins-merge-vs-individual-gap-two-or-more` (decided 2026-08-11). It also keeps #422 and #999 — both net insertions — outside this rule entirely.
2. **Every separation exactly one unchanged base.** The regime the sibling collapse already runs in; real distance is `general.md:34`'s population and stays split.
3. **Some member supplies bases.** `:46`'s mechanism is an *inserted sequence* re-aligning, so a split that inserts nothing cannot be its artefact — W58's `c.[992_993del;995_997del;999_1004del]` stays split.
4. **Every other member renders as a `delins`.** A substitution, an inversion or a duplication is a real type the split buys.

### How it composes with #1904

#1904 (open, closes #1878) states the **equal-length** half of this seam in `partition_block_canonical_within`: the correspondence is unique, so the position-wise pairing is a fact and `DNA/delins.md:17` requires the split. This is its complement and it never enters that regime — condition 1 refuses `result.len() >= reference.len()` on its first line, so an equal-length block cannot reach it whatever its members look like. They touch different functions and different length regimes; there is no textual or semantic collision. This branch is off `origin/main`, not off #1904.

### What it does NOT widen — the most contestable point, flagged deliberately

`delins-payload-coincidence-carve-out-is-coding-dna-scoped` (decided) scopes `:47`'s carve-out to `c.` and nothing else, and this rule is **axis-blind**. The two are reconciled by what each governs: that record governs the terminal *coalesce* passes (`coalesce_payload_alignment_split`, `coalesce_compensating_gap_split`), which take members the partitioner produced and merge them **across unchanged reference bases** — exactly the population `general.md:34` speaks about, whose antecedent those members satisfy. This one governs whether an unequal-length block's interior match is a separation at all. Both passes keep their `c.` gate and neither is touched. The repository's own note on `every_separation_is_a_single_base` already records that `partition_block`'s collapse route is axis-blind ferro policy and that whether it takes an axis gate was "named rather than adjudicated"; this adjudicates it.

**The repository already carried both answers, which is why this needed a record.** `delins_equal_vs_unequal_length_discriminator.rs` held that the MSH2 `g.` row `NC_000002.11:g.47639670_47639673delinsTT` was conformant as a split *because* `:47` cannot reach `g.`, and carried an `assert_ne!` guard against the spanning form. Its hermetic sibling held the *spanning* form to be the right answer for its own `g.` unequal-length arm, as an assert-then-flip. Two sibling modules, one question, opposite positions. This settles it for the partitioner; the MSH2 row now returns itself and that guard is retired — inverted rather than deleted, so the reversal is visible in the code. The hermetic arm is **not** closed by it (its split carries a lone substitution, so condition 4 declines) and stays assert-then-flip.

New record: `rulings[unequal-length-block-a-placed-gap-is-not-a-separation]`, decided, governing `DNA/delins.md:47`.

## What moves, and what it costs

**Corpus.** 2,526 of 94,858 rows move over `examples/dump_normalized_corpus.rs`, and **none of them was previously a fixed point** — so no input a consumer could have stored as already-canonical changes. 1,774 are the new `lone_net_deletion_delins` family collapsing onto the span; 752 are `del_plus_sub` rows whose two-member output moves from a re-cut pair (`g.[265_266delinsA;268del]`) back to the authored pair (`g.[265_266del;268C>A]`), because the derivation is now refused by the weight bound. Every other family reads 0.

**The family is new, and that is part of the measurement rather than a footnote.** Before it the corpus could not build a lone unequal-length *net-deletion* `delins` at all: its only lone spanning family, `split_vs_spanning_delins`, writes a six-base payload over a five-base span (a net insertion), and its other lone `delins` families are equal-length. This change would have measured a **structural zero** over all 85,642 pre-existing rows. That is the fifth instance of the corpus-blindness class the harness records, and the first where the missing dimension is *direction* rather than shape.

**Conformance censuses.**

| census | before | after |
|---|---|---|
| `spec_conformance_axis` 3' `converged` (of 11,272 classes) | 11,016 | 10,959 |
| `spec_conformance_axis` 3' `split_two` | 673 | 730 |
| `spec_conformance_axis` 5' `converged` | 10,753 | 10,696 |
| `spec_conformance_axis` 5' `split_two` | 981 | 1,038 |
| `multi_member_cis_axis` `respelling_converged` (of 461 respellable) | 407 | 406 |

The two `split_two` deltas cancel the `converged` deltas exactly, so every class that left `converged` landed in `split_two` and no arity rose. **Every rank-1 figure is unchanged in both directions** — `declined` 0, `unparseable_outputs` 0, `outputs_denoting_no_sequence` 10/18, `prohibition_violating_outputs` 0, `guard_violations` 0, `non_idempotent_outputs` 4/4, `sequence_changed` 4/0, `conflicts_accepted` 72, coding-axis merge instrument 3 of 997 — so nothing conformant was traded for the movement. `cis_confluence_axis`, `cis_confluence_nr_axis`, `reported_confluence_pairs` and `inversion_sweep` do not move at all.

### The whole fall is #1440's, and that is measured

In every one of these classes the *lone spanning* spelling now stays whole while the *multi-member* spelling does not: the derived block weighs more than the members the caller wrote, so `canonicalize_from_sequence`'s `derived_columns > changed_columns_of_edits(&edits)` refuses that derivation and hands the input back. Disabling that bound on this branch returns `spec_conformance_axis` `converged` to exactly **11,016** and **10,753** with every `split_*` bucket at its pre-#1610 pin, and makes `multi_member_cis_axis::axis_normalized` pass **unchanged at 407** with its row set intact. That bound is `rulings[derivation-may-not-be-bounded-by-the-inputs-spelling]` — decided, "DELETE THE BOUND" — tracked as #1440 and still in the tree.

**#1440 is deliberately not implemented here, and the probe is not a licence to delete it.** With the bound disabled, `guard_violations` rises **0 → 5** and the coding-axis merge instrument 3 → 5 in both directions: merges of the *rejected* SVD-WG010 shape that the bound was blocking. #1440's own record says the ruling removes an input-relative comparand and does **not** license the merges the comparand happened to be blocking. Those five belong to #1440's disclosure.

## Guards, and the evidence that they are guards

- `issue_1610_lone_unequal_length_delins::{a_lone_unequal_length_delins_is_not_split, the_rule_does_not_key_on_the_axis}` — the issue's reproduction on `n.`, `c.` and a CDS-less transcript. **Verified by sabotage**: making `split_is_a_placed_gap_coincidence` return `false` turns both red and leaves the other four green.
- `merge::tests::the_placed_gap_predicate_discriminates` — the predicate directly, with a distinct case for each of the four conditions plus a pure-insertion decline. **This is the guard for the length gate**: relaxing `result.len() >= reference.len()` to `>` turns it red.
- `issue_1610_lone_unequal_length_delins::an_equal_length_block_keeps_its_split` — the boundary pinned end to end, on the same reference and the same span as the fix's own case, so the two read as a pair. Its doc records honestly that the *same* sabotage leaves it green: its split is held apart by condition 4 (a lone substitution), not by the length gate, and an end-to-end case where only the length gate holds the line is not reachable from a spelling that module can write.
- `an_all_deletion_split_stays_split`, `a_net_insertion_stays_split`, `a_substitution_beside_a_placed_gap_stays_split` — one per remaining condition.
- `merge_consecutive_edits_tests::a_length_changing_block_the_exception_still_cannot_reach` — added because the `c.`/`n.` pair it sits beside lost its discriminating power (its block no longer reaches the codon exception at all). The new case still splits, so the conjunct is still asked end to end.

## Verification

`cargo fmt --all --check`, `cargo clippy --all-features --all-targets -- -D warnings`, the full Rust suite (10,755 tests, 0 failures), the reference-aware modules with `FERRO_MANIFEST` set against the prepared reference (`multi_member_cis_axis`, `delins_equal_vs_unequal_length_discriminator`, `mutalyzer_normalize_tests`, `biocommons_normalize_tests`, `hgvs_rs_projection_tests`, `inversion_sweep`, `real_data_normalization_tests` and the rest of the nightly's selection — 291 tests), and `pytest tests/python/` (991 passed, 8 skipped) all run locally. `docs/NORMALIZATION_CONTRACT.md` is blessed for the new record and passes without `BLESS_CONTRACT_DOC`; the clause index block in `clause_ruling_index.rs` is regenerated programmatically rather than transcribed.

Representation-Change: 1798 rows move of 95614 corpus rows, 776 previously-accepted — a real migration for any consumer keying on the normalized string, and a CORRECTION to this PR's earlier claim of 0 previously-accepted. Re-measured armed against the prepared reference at base d8f08c1b, i.e. after #1899 deleted the input-relative weight bound; every figure here supersedes the pre-#1899 ones. Every moved row is on a coding axis (c. 1283, cx. 515; g. 0, n. 0), because the rule is gated to the coding DNA axis per rulings[delins-payload-coincidence-carve-out-is-coding-dna-scoped]. The 776 migrating rows are all in the new lone_net_deletion_delins family and split c. 550 / cx. 226. 1797 of the moved rows are that family collapsing onto its span and 1 is a del_plus_sub row; every other family reads 0. Conformance convergence does NOT fall: spec_conformance_axis 3' converged holds at 11016 and 5' at 10755, every split_* bucket holds, guard_violations stays 0, and multi_member_cis_axis respelling_converged holds at 407 — this PR's earlier prediction that the fall belonged to the weight bound is confirmed, and the three re-pins it carried are reverted. The one census figure that moves is the coding-axis merge instrument, coding_axis_separation_two_or_more_merges 3 to 5, re-pinned and adjudicated row by row. No real GRCh37 row moves: NC_000002.11:g.47639670_47639673delinsTT keeps its split.


nh13:
@coderabbitai pause

nh13:
**Measured ordering hazard between this PR and #1440 — not a defect in this PR as it stands, but it decides what has to happen if #1440 lands after it.**

This PR's censuses are green, and `guard_violations` is 0 on both directions. That holds **because `canonicalize_from_sequence`'s input-relative weight bound is still in the tree.** #1440 is a *decided* ruling to delete that bound.

Measured on this PR's head (`9e0e8636`) with the bound neutered, `spec_conformance_axis` against the prepared reference:

**`guard_violations` 0 → 5.** Five rows, two distinct variants, both frameless axes at separation 1 and both net-deletion unequal-length — this PR's own regime:

```
s01-g-pair-delins-del-p2-sep1 | scale-g-block2-delins-del | scale-g-sep1-delins-del
  NC_TEST.1:g.[265_266delinsAT;268_269del]  ->  NC_TEST.1:g.265_269delinsATA

s01-n3m-pair-del-sub-p4-sep1 | s01-n3p-pair-del-sub-p4-sep1
  NR_TEST.1:n.[36_39del;41T>A]              ->  NR_TEST.1:n.36_41delinsTA
```

Each is `split_is_a_placed_gap_coincidence` collapsing a two-member allele across one unchanged base into a spanning `delins`. `negative_guards` marks such a row as one "whose merge would implement a **rejected** proposal, so a violation there is a verdict" (`spec_corpus.rs:533`); the guard is `svd-wg010-frameless-separation-floor-of-two`.

## Why this is not a finding against the PR today

The control run says so. On `origin/main` with the bound disabled and neither PR applied, the census is **unchanged and all five tests pass** — zero violations. And #1904's head with the bound disabled is also **0**, with 3' back to `main`'s 11,016 and 5' at 10,773, above `main`'s own pin.

So the violations are specific to this rule meeting that deletion. Neither change produces them alone.

## What it implies

`split_is_a_placed_gap_coincidence` currently relies on the weight bound to keep it away from frameless separation-of-one, rather than excluding that case on its own terms. Its four stated conditions are: net deletion; every separation exactly one unchanged base; some member supplies bases; every other member renders as a `delins`. The second condition **admits** separation 1 — the SVD-WG010 shape — and nothing else in the predicate rejects a frameless axis.

Two ways to settle it, both cheap relative to finding this after the fact:

1. Give the predicate its own axis/frame condition, so it does not depend on a bound that a decided ruling says to delete.
2. Record the dependency explicitly on the new ruling record `unequal-length-block-a-placed-gap-is-not-a-separation`, so #1440's implementer meets it.

(1) is the more durable, since the ruling on #1440 is already decided and (2) leaves a live tripwire.

Full method and numbers on #1440: https://github.com/fulcrumgenomics/ferro-hgvs/issues/1440#issuecomment-5295513823

PR #1911, opened 2026-08-14
feat: add sequence_normalize() helper

TimD1:
### Summary
This PR adds sequence-based normalization for an HGVS variant (with optional standard `normalize()` afterwards), guaranteeing confluence when normalizing string-derived variants to a single representation.

### Implementation
* `sequence_normalize()` wraps `to_sequence()` and `from_sequence()`, handling logic revolving around iterative doubling of the padding to ensure convergence to a single representation.
* `placement_bounded_by_window` was split into `bounded_at_start` and `bounded_at_end` to assist in determine whether the window needs to keep growing or not

### Review follow-ups

* **The loop starts at `merge::CANONICAL_PAD` (128), not at 1.** `dup` typing reads the reference bases immediately 5' of an insertion point, so a window that does not reach them derives an `ins` instead — and that mis-typed `ins` rests on *neither* window edge, so both `bounded_*` flags are clear and the loop returned it without ever widening. There is no flag for "wrong for a reason the edges cannot show", so the only defence is starting wide enough to type a duplication. Measured on the new `a_duplication_and_its_insertion_spelling_converge`: at a 1-base start pad `g.14_15insACGTACGT` came back unchanged while `g.7_14dup` did not — the exact non-confluence this PR exists to remove. `CANONICAL_PAD` is now `pub(crate)` so the three places that must agree on 128 (the normalizer's own derivation, this loop, the `pad=128` default on the Python `to_sequences` binding) cannot drift apart.
* **A narrow-window refusal is retried rather than propagated.** `from_sequences` declines a derived insertion whose only HGVS anchor lies 5' of the first base supplied, and that refusal's own text says to supply more 5' flank — which is precisely what this loop does. The `?` turned a widenable window into a terminal error. It now widens while either side can still grow and reports the first refusal only if none of them helped.
* **`get_sequence_length` is no longer required.** It carries a trait default that always errors, and `apply_to_reference_padded` treats that as non-fatal on purpose ("an out-of-tree provider that serves bases perfectly well may not implement it"). The loop hard-failed on it one line after `to_sequences` had already succeeded for such a provider. The lookup is now optional, with `SequencePair::window_is_final` and the existing `prev_bounds` fallback carrying termination.
* **Python surface completed.** `bounded_at_start` / `bounded_at_end` and `Normalizer.sequence_normalize` are added to `python/ferro_hgvs/__init__.pyi`, the `from_sequences_detailed` docstring names all three attributes, and `tests/python/test_sequence_normalize.py` covers the binding-only failure modes (keyword-only signature, `normalize` threading, which exception class comes out of which failure, and that the two getters are wired to their own fields rather than both to the OR).
* **Rebased onto current `main` (39 commits).** The branch predated the removal of the Python `direction=` keyword from `Normalizer.__new__` and `from_sequences_detailed` ("3' is the only direction ferro shifts"), and CI builds the merge ref, so the new Python tests were compiled against bindings the branch could not see. The Rust `ShuffleDirection` instrument survives, so `tests/it/sequence_normalize.rs` still pins both shuffle directions; the Python file drops `direction=` and separates the two per-side flags with substitutions flush against opposite window edges instead.
* Docs and formatting: `# Errors` on `sequence_normalize`, a per-field doc on `bounded_at_end` in both structs, the pad-ceiling refusal no longer calls `MAX_PAD` the window's total width (`to_sequences` pads both sides), the stale "128-base start pad" comment now matches the constant, and `cargo fmt --all`.

Representation-Change: none. `placement_bounded_by_window()` is the OR of the two new fields and `any(p.start == 0 || p.end == len)` is identical to `any(p.start == 0) || any(p.end == len)`, so the split is value-identical for every input; `sequence_normalize` is a new entry point with no previously-shipped output. Nothing already emitted changes.

### DIFF
--- a/python/ferro_hgvs/__init__.pyi
+++ b/python/ferro_hgvs/__init__.pyi
@@ -969,6 +969,53 @@ class Normalizer:
         """
         ...
 
+    def sequence_normalize(
+        self,
+        description: str,
+        *,
+        max_grid_cells: int | None = None,
+        normalize: bool = False,
+    ) -> str:
+        """Re-derive a description from the bases it denotes.
+
+        The "one canonical description per variant" round trip, as a single
+        call: express the variant as a padded reference/alternate window
+        (``to_sequences``), derive a description from those bases alone
+        (``from_sequences``), and — while a member still rests on a window edge
+        that can still move — double the pad and retry. Two spellings of one
+        variant therefore reach one description, decided by the observed bases
+        rather than by how either was written.
+
+        The loop reads ``DerivedDescription``'s two per-side flags apart, so a
+        placement pinned to the sequence's own start or end is recognised as
+        settled rather than chased.
+
+        Args:
+            description: The HGVS description to re-derive. Its axis must be one
+                ``from_sequences`` emits — genomic ``g.`` (and ``m.`` on the two
+                rCRS mitochondrial accessions); any other is refused.
+            max_grid_cells: As ``from_sequences``.
+            normalize: Route the derived description through ``normalize``
+                before returning it. Defaults to False, which yields the
+                alignment-derived form (conformant + deterministic) rather than
+                ferro's recommended form. Set True for the recommended,
+                reference-anchored form.
+
+        Returns:
+            The canonical HGVS description, as a string.
+
+        Raises:
+            ParseError: ``description`` does not parse.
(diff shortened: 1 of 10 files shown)


nh13:
@coderabbitai pause

TimD1:
Moved to #1930 (branch on this repo, not fork)

Issue #1917, opened 2026-08-14
normalize_in_grown_window underflows on a reversed range (SVD-WG006 wraparound dup): end - window_start with the window anchored on start


nh13:
`Normalizer::normalize_in_grown_window` anchors its shuffle window on `start` alone and then subtracts that anchor from `end`:

```rust
// src/normalize/mod.rs:5289-5308  (origin/main @ 19e2e244)
loop {
    let window_start = start.saturating_sub(window);          // :5290 — anchored on `start`
    let fetch_end = fetch_end_of(end.saturating_add(window));  // :5291
    let Ok(ref_seq) = self.provider.get_sequence(accession, window_start, fetch_end) else {
        return Ok(None);
    };

    let (new_rel_start, new_rel_end, new_edit, warnings) = self.normalize_na_edit(
        ref_seq.as_bytes(),
        edit,
        start - window_start,      // :5307 — fine, this is `window`
        end - window_start,        // :5308 — UNDERFLOWS when end < window_start
        …
```

`window_start` is `start - window_size` (default `window_size = 100`). Nothing establishes `start <= end`, so on a **reversed range** — `start > end` — with `start - end > window_size`, `window_start` lands beyond `end` and `end - window_start` underflows. Debug/`overflow-checks`: `attempt to subtract with overflow`. Release: a wrap to a value near `u64::MAX` used as a window-relative end coordinate.

A reversed range is not malformed input. SVD-WG006 admits `<high>_<low>` for a circular deletion or duplication, and ferro's own code says so in four places (`src/spdi/convert.rs:613`, `:761`, `:805`, `:846`: "Wraparound variants (start > end, per SVD-WG006)"). The parser accepts it: `g.500_300dup` parses, reaches the normalizer, and panics.

## Reproducers

**Real corpus.** A census over 9,949,738 rows found exactly two inputs that panic identically, on both partition arms:

```
NC_000012.11:g.110593351_110576466dup
NC_000016.9:g.23634775_23621090dup
```

Both are reversed (`110593351 > 110576466`, `23634775 > 23621090`), and in both the gap far exceeds `window_size`.

**Hermetic, no reference needed.** Reproduced on `origin/main` @ `19e2e244` against a synthetic 1000-base contig built through `JsonProvider`:

```rust
let p = provider_len("NC_TEST.1", 1000);              // 1000 cyclic ACGT bases
[...]

PR #1918, opened 2026-08-14
fix(reference): decline an intron distance the offset cannot measure

### LINKED ISSUE #1916: IntronPosition::distance_from_donor/acceptor underflow on an offset larger than the intron — the decline channel #1767 left unused
`IntronPosition::distance_from_donor` and `IntronPosition::distance_from_acceptor` compute the cross-boundary distance as an unchecked `u64` subtraction:

```rust
// src/reference/transcript.rs:1013-1032
pub fn distance_from_donor(&self) -> Option<u64> {
    match self.boundary {
        IntronBoundary::FivePrime => Some(self.offset.unsigned_abs()),
        IntronBoundary::ThreePrime => {
            // Distance = intron_length - distance_from_3prime
            Some(self.intron_length - self.offset.unsigned_abs())   // <-- 1018
        }
    }
}
```

`distance_from_acceptor` carries the mirror image at `:1029`. When `self.offset.unsigned_abs()` exceeds `self.intron_length` the subtraction underflows: `attempt to subtract with overflow` under `overflow-checks` (the `dev`, `test` and `soak` profiles all enable it), and in release a wrap to a value near `u64::MAX` returned as a **distance in bases from a splice site**.

The file already names this. The block comment at `:944-951` — written by #1767 — says so in its own words:

> `distance_from_donor` / `distance_from_acceptor` **ALREADY return `Option<u64>`**, so they have a decline channel and do not use it. Worse than a fabricated band: the cross-boundary arm computes `intron_length - offset.unsigned_abs()`, an unchecked `u64` subtraction that underflows for ANY magnitude exceeding the intron […] `unsigned_abs` closed the ladder's overflow in #1767 and left this subtraction untouched, so the hardening pass is incomplete here rather than deliberately scoped.

So this is the residue of #1767: that PR made the five *predicates* on this struct total by switching `abs()` to `unsigned_abs()`, and its regression guard `the_intron_position_predicates_are_total` (`tests/it/issue_1767_unknown_offset_splice_classifiers.rs:227`) exercises exactly the two functions it did **not** fix — it builds an `IntronPosition` with `offset: i64::MIN, intron_length: 1_000` and calls the five predicates, never the two distances. The reproducer is one line away from a committed test and nothing calls it.

## Reproducers

Both are library-constructed, which is the only way in — see *Reachability* below. Measured on `origin/main` @ `19e2e244`.

**1. A measured offset larger than the intron.**

```rust
let pos = IntronPosition {
    intron_number: 1,
    boundary: IntronBoundary::FivePrime,
    offset: 300,
    tx_boundary_pos: 100,
    intron_length: 100,
};
pos.distance_from_acceptor()
```

```
panicked at src/reference/transcript.rs:1029:22:
attempt to subtract with overflow
```

Release would return `Some(18446744073709551516)`.

**2. An unknown-offset sentinel (`c.100-?`).**

```rust
let pos = IntronPosition {
    intron_number: 1,
    boundary: IntronBoundary::ThreePrime,
    offset: i64::MIN,          // OFFSET_UNKNOWN_NEGATIVE
    tx_boundary_pos: 100,
    intron_length: 1_000,
};
pos.distance_from_donor()
```

```
panicked at src/reference/transcript.rs:1018:22:
attempt to subtract with overflow
```

Release would return `Some(9223372036854776808)`.

## Observed vs correct

| input | observed (debug) | observed (release) | correct |
|---|---|---|---|
| `FivePrime`, offset 300, `intron_length` 100 → `distance_from_acceptor` | panic | `Some(18446744073709551516)` | `None` |
| `ThreePrime`, offset `i64::MIN`, `intron_length` 1000 → `distance_from_donor` | panic | `Some(9223372036854776808)` | `None` |
| `FivePrime`, offset `i64::MAX`, `intron_length` 1000 → `distance_from_donor` | `Some(9223372036854775807)` | same | `None` |

The third row is the one that does **not** panic, and it is why `saturating_sub` is the wrong fix. It is the *same-boundary* arm, which returns `Some(offset.unsigned_abs())` unconditionally — so an unknown offset comes back as a definite distance of 9.2 × 10^18 bases, and a 300-into-a-100-base-intron offset comes back as a confident `Some(300)`. Clamping only the arm that panics would leave the two functions disagreeing about the same position: `distance_from_donor() == None` (clamped) beside `distance_from_acceptor() == Some(300)`.

## Why the answer is a refusal, not a clamp

Two distinct regimes reach the subtraction, and both are *unanswerable* rather than merely awkward to compute:

- **`offset.unsigned_abs() > intron_length`.** The struct asserts a position inside intron *N*, whose length is `intron_length`, sitting more than `intron_length` bases from one of its own boundaries. No position in that intron satisfies that. The record is self-inconsistent, and no distance to *either* boundary is derivable from it — so `saturating_sub`'s `0` would be a fabricated distance, and a fabricated `0` is the worst available value here because `0` reads as *at the splice site* and every downstream band (`is_canonical_splice_site`, `SpliceSiteType::DonorCanonical`, `IntronicConsequence::SpliceDonorVariant`, HIGH impact) treats it as the most clinically significant answer there is.
- **`offset` is `OFFSET_UNKNOWN_POSITIVE` / `OFFSET_UNKNOWN_NEGATIVE`.** These denote a position unbounded in one direction. `is_unknown_offset`'s own doc states the contract: "per the spec they denote a position unbounded in one direction, **from which no distance can be derived at all**". #1767 applied exactly that to `IntronicConsequence::from_cds_pos` (returns `None`), and #1841 took the public-API break to apply it to `IntronicRegion::from_offset` and `EffectPredictor::classify_splice_variant`. These two already return `Option`, so no break is needed — the channel is just unused.

Both arms of both functions should decline, so that the two functions never disagree about whether a given `IntronPosition` states a resolvable position.

## Reachability

Not reachable from a string entry point. `find_intron_at_genomic` — the only in-tree producer, reached via `CoordinateMapper::get_intron_position` and thence `src/vcf/to_hgvs.rs` — maintains `dist_to_5prime + dist_to_3prime == intron_length + 1` with both terms `>= 1`, so a derived `IntronPosition` always satisfies `offset.unsigned_abs() <= intron_length` and never carries a sentinel.

It is reachable by any library caller, because **every field of `IntronPosition` is `pub`** and the struct is exported from `ferro_hgvs::reference::transcript`. That is the same reachability argument #1767 accepted for the five predicates on this same struct, and `tests/it/issue_1767_unknown_offset_splice_classifiers.rs` and `src/convert/noncoding.rs`'s own unit tests both construct one directly. The functions are public API returning `Option`; a caller is entitled to expect the `None`.

## Scope

`src/reference/transcript.rs` only. No in-tree caller reads either function outside tests, so no output moves.

nh13:
Closes #1916

`IntronPosition::distance_from_donor` and `IntronPosition::distance_from_acceptor` computed the cross-boundary distance as `intron_length - offset.unsigned_abs()` — an unchecked `u64` subtraction. Any offset magnitude exceeding `intron_length` underflowed it: `attempt to subtract with overflow` under `overflow-checks` (the `dev`, `test` and `soak` profiles all enable it), and in release a wrap to a value near `u64::MAX` returned as a **distance in bases from a splice site**.

This is the residue #1767 left. That PR made this struct's five predicates total by reading `offset` through `unsigned_abs` rather than `abs`, and the block comment it wrote on `src/reference/transcript.rs` names these two functions as the part it did not finish — that they "ALREADY return `Option<u64>`, so they have a decline channel and do not use it". Its guard `the_intron_position_predicates_are_total` builds the exact record that reproduces this (`offset: i64::MIN`, `intron_length: 1_000`) and calls only the five predicates.

## Root cause

Two regimes reach the subtraction, and in both the record states no distance at all:

- **`offset.unsigned_abs() > intron_length`** — the record asserts a position inside an intron of length `intron_length` sitting further than that from one of its own boundaries. No position in that intron does. The record is self-inconsistent, and no distance to *either* boundary follows from it.
- **`offset` is an unknown-offset sentinel** (`c.N+?` / `c.N-?`). Per `is_unknown_offset`'s own contract these "denote a position unbounded in one direction, from which no distance can be derived at all" — the ruling #1767 applied to `IntronicConsequence::from_cds_pos` and #1841 took a public-API break to apply to `IntronicRegion::from_offset` and `EffectPredictor::classify_splice_variant`.

## Why a refusal rather than a clamp

Two reasons, and the second is the one that shapes the diff.

**`saturating_sub` would answer `Some(0)`, which is the worst value available here.** `0` reads as *at the splice site*: `is_canonical_splice_site`, `SpliceSiteType::DonorCanonical`, `IntronicConsequence::SpliceDonorVariant` and HIGH impact all key off small magnitudes, so a clamp would promote an unmeasurable offset to the most clinically significant answer the API can give. A wrong offset silently rendered as a canonical splice-site hit is worse than the panic it replaces.

**A clamp would also fix only one of the two arms.** The *same*-boundary arm — `distance_from_donor` on a `FivePrime` position — returns `Some(offset.unsigned_abs())` unconditionally and never underflowed, so it needed no clamp and would have kept answering. The result would be the two functions disagreeing about one position: `distance_from_donor() == None` beside `distance_from_acceptor() == Some(300)`, for the same record. So the magnitude test lives in a shared `measured_offset` helper that both arms of both functions consult, and `the_two_distances_never_disagree_about_resolvability` is the assertion that forbids a one-armed fix. Both functions already return `Option`, so no API break is needed to say any of this.

## Boundary, and the negative control

Offsets are 1-based from their boundary, so a magnitude of exactly `intron_length` names the intron's far-edge base — inside the intron, and still answerable. `intron_length + 1` names the first base of the flanking exon and declines. `the_guard_fires_exactly_where_the_magnitude_leaves_the_intron` pins all three adjacent magnitudes so an off-by-one fails.

The refusal must not reach anything the library can actually derive. `find_intron_at_genomic` — the only in-tree producer, reached via `CoordinateMapper::get_intron_position` and thence `src/vcf/to_hgvs.rs` — maintains `dist_to_5prime + dist_to_3prime == intron_length + 1` with both terms at least 1, so a derived `IntronPosition` always satisfies `offset.unsigned_abs() <= intron_length` and never carries a sentinel. `every_derived_intronic_position_still_answers` walks **every** genomic base of both introns of a programmatically built two-intron transcript, on both strands, requires `Some` from both functions at each one, and asserts `donor + acceptor == intron_length` throughout. It asserts its own visit count so a zero cannot pass as a result. That test passes both before and after the fix, which is what makes it the control.

The two introns are deliberately different lengths (100 and 300 bases) so a guard accidentally keyed on one fails on the other. No fixture is committed — the transcript is built in the test.

## Reachability

Not reachable from a string entry point; reachable by any library caller, because every field of `IntronPosition` is `pub` and the struct is exported from `ferro_hgvs::reference::transcript`. That is the same reachability argument #1767 accepted for the five predicates on this same struct, and both `tests/it/issue_1767_unknown_offset_splice_classifiers.rs` and `src/convert/noncoding.rs`'s unit tests construct one directly.

## Class sweep

Swept the surrounding class — unchecked `-` on unsigned operands near `unsigned_abs()`, `abs_diff`, `len()` and coordinate arithmetic — and triaged every hit, so a reader can tell a swept finding from an unswept one.

| site | verdict |
|---|---|
| `reference/transcript.rs:1018`, `:1029` | **confirmed** — this PR |
| `normalize/mod.rs:5308` `end - window_start` | **confirmed, filed separately as #1917** — reversed range (SVD-WG006 wraparound `dup`), different root cause and opposite correct answer; see below |
| `convert/coding.rs:30` `cds_end - cds_start + 1` | **confirmed, out of scope** — inverted-CDS record, same class as the open `fix/mapping-u64-underflow` work on `data/mapping.rs`; flagged rather than folded in |
| every other `unsigned_abs()` in `src/` | dismissed — each feeds a *comparison* (`> 50`, `<= 2`, a band ladder), never a subtraction |
| `reference/transcript.rs:865` | dismissed — magnitude feeds `checked_add` / `checked_sub` |
| `benchmark/cache.rs:4115`, `:4120` | dismissed — guarded by `cds_start >= *exon_start` in the same condition |
| `data/cdot.rs:1167`, `:1183`, `:1186` | dismissed — each guarded by the enclosing branch (`offset <= cds_start`, `tx_pos < cds_start`) |
| `conformance/spec_corpus.rs:769` | dismissed — guarded by `one_based < cds_start` |
| `normalize/mod.rs:12974`–`:12978` (revcomp flip) | dismissed — the `in_window` check immediately above bounds every operand by `seq_len` |
| `normalize/mod.rs:12994` (`unflip_intronic_positions`) | dismissed — all four callers pass positions `normalize_na_edit` returned against the same window; a caller-derived invariant, not an enforced one |
| the three `abs_diff` sites in `normalize/merge.rs` | dismissed — `abs_diff` is total by construction |
| `benchmark/compare.rs:3934` | dismissed — `i64` arithmetic behind the `benchmark` feature, not unsigned |

**#1917 is deliberately not folded into this PR.** The two share a shape — an unchecked `u64` subtraction whose operand order is not guaranteed — but not a root cause, and the correct answers point in opposite directions. #1916's operand-order violation arises from a self-inconsistent record, about which nothing true can be said, so the fix is to refuse. #1917's arises from a **legal** input (`NC_000012.11:g.110593351_110576466dup`, `NC_000016.9:g.23634775_23621090dup`) that the window computation does not account for; refusing would reject a spelling SVD-WG006 admits, so the fix is to anchor the window on `min(start, end)` instead. They also have different blast radii: this PR touches `src/reference/` and moves nothing, while #1917 touches `src/normalize/` and moves those rows from panic to an output — a disclosure that has to be measured on its own.

Representation-Change: none. No in-tree caller reads either function outside tests, and every position `find_intron_at_genomic` can derive still answers — pinned by `every_derived_intronic_position_still_answers`, which walks every intronic base on both strands and passes both before and after. The change only converts a panic (release: a `u64::MAX`-adjacent distance) into the `None` both signatures already admit.

## Verification

- `cargo nextest run --features dev --no-fail-fast` — see the run recorded on the PR.
- The four new assertions fail on the parent commit (2 of the 6 tests in the file pass there: both are controls describing behaviour this change must not move).

### DIFF
--- a/src/reference/transcript.rs
+++ b/src/reference/transcript.rs
@@ -942,16 +942,17 @@ pub struct IntronPosition {
 //   would take the identical `Option` treatment; it was simply outside an
 //   approved break. Do not read the `bool` argument above as covering it.
 // - `distance_from_donor` / `distance_from_acceptor` ALREADY return `Option<u64>`,
-//   so they have a decline channel and do not use it. Worse than a fabricated
-//   band: the cross-boundary arm computes `intron_length - offset.unsigned_abs()`,
-//   an unchecked `u64` subtraction that underflows for ANY magnitude exceeding
-//   the intron — panicking under `overflow-checks` and wrapping to a colossal
-//   "distance" in release. `unsigned_abs` closed the ladder's overflow in #1767
-//   and left this subtraction untouched, so the hardening pass is incomplete
-//   here rather than deliberately scoped.
+//   so they had a decline channel and did not use it. **Closed by #1916**: the
+//   cross-boundary arm computed `intron_length - offset.unsigned_abs()`, an
+//   unchecked `u64` subtraction that underflowed for ANY magnitude exceeding the
+//   intron — panicking under `overflow-checks` and wrapping to a colossal
+//   "distance" in release. Both now route through `measured_offset` and decline.
+//   `unsigned_abs` closed the ladder's overflow in #1767 and left that
+//   subtraction untouched, so the hardening pass was incomplete there rather
+//   than deliberately scoped.
 //
-// Neither is reachable from a string entry point today; both are reachable by a
-// library caller, because every field of this struct is `pub`.
+// `splice_site_type` is not reachable from a string entry point today; it is
+// reachable by a library caller, because every field of this struct is `pub`.
 impl IntronPosition {
     /// Check if this is a deep intronic position (>50bp from nearest exon)
     pub fn is_deep_intronic(&self) -> bool {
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1920, opened 2026-08-14
docs(normalize): cite #1899 in the weight-bound ruling record

### LINKED ISSUE #1440: Weight bound judges a sequence-derived partition against the input's spelling

nh13:
Traceability fix for a change that had already landed.

Representation-Change: none. Prose inside one ruling record, plus the generated document re-blessed from it. No `src/` file is touched and no output moves.

## What was wrong

**#1899** implemented the decided ruling `derivation-may-not-be-bounded-by-the-inputs-spelling` and rewrote that record with the implementation and its measured figures — but named neither the issue nor itself. It carried no `Closes #1440` and no mention of #1440 at all, and the record does not cite the PR. So the ruling, the issue and the code that satisfies them met in no single place: from the issue you could not find the change, and from the record you could not find it either.

#1440 is now closed citing #1899, and #1899 carries a back-reference. This adds the durable half, in the ledger, where the next person adjudicating this question will actually look.

## What it also records, and why that is the load-bearing part

**#1899 did more than delete the bound**, and the figures already in the record are for that whole change:

- `MAX_SPLIT_BLOCK` moved from a literal `1024` to `MAX_CANONICAL_WINDOW as usize`
- `past_the_split_cap(reference, result)` now fronts the block hand-back
- `MAX_SEPARATION_AUDIT_WIDTH` was de-aliased from `MAX_SPLIT_BLOCK` — its own doc records the aliasing as the defect

Without that stated, the record's closing paragraph — `converged` falling 18 at 3' and 14 at 5' — reads as the cost of removing the comparand. It is not. The deletion alone was measured to move **nothing**, so that cost belongs to the split cap.

That measurement is independent of #1899 and predates it: neutering the comparand on then-current `main` (`19e2e244`, pre-merge) with `FERRO_MANIFEST` set against a prepared GRCh38 reference left `spec_conformance_axis` **byte-identical** — `EXIT=0`, all 5 tests pass, `guard_violations` 0, every pinned figure unchanged. It was taken for a different purpose (reconciling a disagreement about what the bound was holding up) and reached #1899's conclusion from the other direction.

## Verification

- `generate_spec_fixture` regenerated first (`GEN_EXIT=0`), since the gitignored artifact is stale after any ledger edit and its failure otherwise reads as a dropped ruling record.
- `docs/NORMALIZATION_CONTRACT.md` blessed **after** the prose edit, then confirmed passing **without** `BLESS_CONTRACT_DOC` (`VERIFY_EXIT=0`).
- `ruling_records_are_intact`, `clause_ruling_index` and `claude_md_adjudication` — 12 passed, `GUARDS_EXIT=0`.
- The JSON was round-tripped through Python, so the edit was checked to be the only one: 31 records before and after, ids identical, statuses identical, top-level keys identical, and exactly one record differing.

No status changed, so neither adjudication table in `CLAUDE.md` moves.


nh13:
@coderabbitai pause

Issue #1921, opened 2026-08-14
Reversed-range parse gates are near-complementary: the linear g. axis admits inv/dup, the circular m./o. axes admit del/delins/dup


nh13:
The wraparound exception SVD-WG006 grants is **circular-only**, and ferro enforces it on the two axes where it does not apply while declining it on the axis where it does. The two lists are close to each other's complement.

## The two gates

**Circular (`m.`/`o.`)** — `check_circular_reversed_range` (`src/hgvs/parser/variant.rs:548`) admits `del`, `delins`, `dup` and refuses everything else. That matches the spec: `DNA/deletion.md:17` names deletions, `consultation/SVD-WG006.md:15` grants the exception for a rearrangement including both the last and first nucleotides with worked examples for `del` (`NC_012920.1:m.16563_13del`) and `dup` (`J01749.1:o.4344_197dup`), and `:33` names "deletions/duplications" as the scope. `delins` follows by the `del + ins` composition. The function's own doc comment says all of this and cites it.

**Linear (`g.`)** — `parse_genome_interval_inner`'s `allows_inverted` (`:306`) admits a reversed range when the remaining input starts with `dupins`, `inv`, `ins` or `dup`. Its stated rationale is that these are "structural variant type[s] that may use inverted coordinates". No clause is cited, and none exists: a linear reference has no origin to wrap around, so SVD-WG006's exception cannot reach it.

## The asymmetry

| edit | linear `g.` | circular `m.`/`o.` | spec |
|---|---|---|---|
| `del` | refused | admitted | authorised — **on circular only** |
| `delins` | refused | admitted | authorised by composition — on circular only |
| `dup` | admitted | admitted | authorised — on circular only |
| `inv` | **admitted** | refused | silent on both |
| `ins` | admitted at `:306`, refused later by #1079 | refused | silent on both |

So `inv` — spec-silent everywhere — is the one reversed form the linear axis lets through to the normalizer that the circular axis refuses, and `del` — the clause's headline case — is refused on the linear axis for the right outcome by the wrong mechanism (it is simply absent from a list assembled on other grounds).

Measured on `origin/main` @ `426a944b`: over `g.3500_<end>{dup,inv,insACGT,delinsACGT,del,dupinsACGT,=}` at twelve spans, exactly **`dup` and `inv`** reach the normalizer. `ins` and `dupins` are admitted by `allows_inverted` and refused by a later check; `del`, `delins` and `=` never get past `:306`.

[...]

PR #1923, opened 2026-08-14
fix(normalize): decline a reversed range rather than reading it as a linear window

### LINKED ISSUE #1917: normalize_in_grown_window underflows on a reversed range (SVD-WG006 wraparound dup): end - window_start with the window anchored on start

nh13:
Closes #1917.

`normalize_in_grown_window` anchored its fetch on `start` alone and then computed the window-relative end as `end - window_start`. Nothing establishes `start <= end`, so a reversed range — which names a **wraparound** — walked into linear arithmetic that cannot express one.

## The issue reported one of three failures, and generalised it wrongly

The issue's rule was "any reversed range whose span exceeds `window_size` reproduces it". That is wrong in **both** directions: the widest reversed ranges never panicked, and the narrowest panicked somewhere else entirely. All three outcomes are functions of `start - end` against `window_size` (default 100):

| `start - end` | before |
|---|---|
| `<= w` | panic — `normalize_na_edit`'s `dup` arm slices `ref_seq[s..e]` **backwards** (`slice index starts at 99 but ends at 50`) |
| `(w, 2w]` | panic — `end - window_start` underflows (the reported one) |
| `> 2w` | the minimal-notation fallback, **by accident**: the fetch bounds themselves invert and the provider rejects the read |

Both real-corpus rows named in the issue are in the bottom band (16,885 and 13,685 bases), so they were already taking the fallback exit. It follows that the reproducer the issue offers — `g.500_300dup`, span 200 — is the *only* one of its three examples that demonstrates the defect it describes, and it does so for a reason the issue's stated rule does not capture.

## The fix, and why it is not the one the issue proposed

`normalize_mito` has gated this at its own axis since #129, and its comment states the reason to gate rather than lean on a downstream fail-safe: the fail-safes hold only for some window sizes. The `g.` axis never got the same gate. It does now, placed on the shared helper that owns the underflowing arithmetic rather than on the axis, so no future caller can bypass it.

The issue proposed anchoring the window on `min(start, end)` and extending it to `max(start, end)`, "with the reversed-ness of the range preserved in the output". **That is half a fix and the half it omits is worse than the half it does.** It removes the underflow, and then `normalize_na_edit` receives `rel_start > rel_end` for *every* reversed range and slices backwards — widening the top band's panic from a narrow band to the whole class.

And there is nothing for the wider window to do once fetched. The shuffle, the alignment and SPDI are each linear-contiguous; `src/spdi/convert.rs` refuses wraparound by name in four places, saying "SPDI is a single edit and has no representation for circular-contig wraparound". Preserving the reversed-ness of a range *is* declining to shift it. Circular-aware shuffle modulo contig length remains #129's follow-up, and this change does not prejudge it.

Declining is not clamping — the objection the issue raises against saturating the subtraction, that it "would silently re-point the edit at the window's first base", is an objection this fix satisfies rather than one it trades against.

## What moves — a panic was not the only outcome

Censused over every parseable reversed form × twelve spans on a 4000-base contig, **24 rows were answered**: 10 panicked, 8 already declined, and **6 came back with an answer that was wrong**. Those six are the disclosure, and they are the part the issue did not find — a panic is loud, a wrong description is not, so these are the rows a consumer could have stored:

- `g.3500_3499dup` → **`g.4001_4000dup`**, a coordinate past the end of its own 4000-base contig; and
- five `inv` rows (spans 1, 5, 50, 99, 100) → **`=`**, asserting the sequence is *unchanged* by an inversion.

Both are precisely what two of the four seam oracles exist to catch — `FERRO_ASSERT_IN_BOUNDS` and `FERRO_ASSERT_SEQUENCE` respectively — and neither fired, because no corpus in either selection feeds a reversed range. The `inv` shape is the worse of the two: `=` is well-formed, in-bounds and idempotent, so the other three oracles are satisfied by it as well.

All six now return the description as authored, agreeing with the eight rows in the bottom band that already did.

## The corpus measurement is a STRUCTURAL zero, and is reported as one

`examples/dump_normalized_corpus.rs` cannot build the shape this change keys on. It contains no reversed-range construction at all — no `start > end`, no wraparound, and every range it spells is built from an ordered pair — so an A/B over it returns 0 rows moved for the same reason a 20-mer corpus returns 0 declines against `MAX_SPLIT_BLOCK` (#1460) and a single-exon generator returns 0 junction crossings (#1478). Quoting that zero as evidence of safety would be the same error those two issues record.

So the disclosure above rests on the 84-row census, which enumerates the property directly (`start - end` across all three bands × every parseable edit type), not on the corpus dump.

## Also filed

**#1921** — the reversed-range parse gates are near-complementary. The linear `g.` axis admits `dup`/`inv`; the circular `m.`/`o.` axes admit `del`/`delins`/`dup`. SVD-WG006's exception is circular-only, so `inv` — spec-silent everywhere — is admitted on the axis with no exception and refused on the axis that has one. Not settled here: refusing reversed ranges at parse on the linear axis would reject inputs ferro accepts today, including both #1917 corpus rows, and wants its own adjudication record. This guard is correct under either resolution, which is why the new test asks only that what parses reaches an answer and names no specific set.

## Verification

- Full gate **10,791 / 10,791 passed**, `FERRO_MANIFEST` armed, `FERRO_REQUIRE_BULK_FIXTURES=1`, **zero `Skipping:` lines** (the four release-asset fixtures were fetched first, so the ClinVar / CMRG / Paraphase suites really ran rather than skipping green). Exit status read back from the `EXIT=` line written into the log, not from the job notification — the command was a chain ending in `echo`, so the notification reports the echo's status.
- `cargo clippy --all-features --all-targets -- -D warnings` clean; `cargo fmt --all --check` clean; `cargo doc --no-deps --features dev` adds no warning naming anything in this diff.
- **Python binding suite run locally — 1,000 passed, 8 skipped.** `cargo nextest` does not reach `tests/python/`, and this touches `src/normalize/`, which is the #1870 shape.
- The gate ran against the committed tree; the three later edits are comment-only, verified by filtering the diff for non-comment changed lines rather than asserted.
- The eight new tests fail on the pre-guard tree in the three ways tabulated above; the guard was disabled in place and the census re-run to produce the before column, then the source restored byte-for-byte from a copy.
- `the_bands_below_are_stated_against_the_shipped_window_size` pins `window_size == 100`, so a future default change fails loudly rather than quietly moving each band test off the band it names.
- `every_linear_reversed_form_the_parser_admits_is_declined` asserts `parsed > 0`, so it cannot pass vacuously if the parser later refuses everything it tries.

Representation-Change: 6 rows move, all from a wrong description to a decline. Measured by census, not by corpus. Over every parseable reversed form x twelve spans on a 4000-base contig, 24 rows were answered before this change: 10 panicked, 8 already declined, and 6 returned a wrong description - one dup answering g.4001_4000dup on a 4000-base contig (past its own end), and five inv rows answering `=`, an identity, for an inversion. Those 6 now return the description as authored, which is what the other 8 already did. The 10 panicking rows are not counted as moved, having had no previous answer to move from. The synthetic corpus measures 0 rows moved and that figure is STRUCTURAL, not evidence: dump_normalized_corpus.rs builds no reversed range anywhere, so it cannot observe this change in principle - the same blindness as #1460 and #1478. Report a structural zero as structural.

### DIFF
--- a/src/normalize/mod.rs
+++ b/src/normalize/mod.rs
@@ -5247,10 +5247,12 @@ impl<P: ReferenceProvider> Normalizer<P> {
     ///
     /// # Returns
     ///
-    /// `Ok(None)` when the caller should fall back to minimal notation: **any**
-    /// fetch failed — the first one or a grown one (a grown failure is not
-    /// allowed to fall back to the truncated attempt that preceded it, which
-    /// would reinstate the walk) — or the growth cap was reached by an edit
+    /// `Ok(None)` when the caller should fall back to minimal notation: the
+    /// range is **reversed**, so it names a wraparound this linear window
+    /// cannot express (#1917, see the guard below); **any** fetch failed — the
+    /// first one or a grown one (a grown failure is not allowed to fall back to
+    /// the truncated attempt that preceded it, which would reinstate the walk);
+    /// or the growth cap was reached by an edit
     /// [`Self::canonicalize_without_shifting`] declines to re-type.
     ///
     /// **Refusing to shift is the only capped answer that is still a fixed
@@ -5281,6 +5283,54 @@ impl<P: ReferenceProvider> Normalizer<P> {
         edit: &NaEdit,
         fetch_end_of: impl Fn(u64) -> u64,
     ) -> Result<Option<WindowedNormalization>, FerroError> {
+        // A reversed range names a wraparound, and this window is linear
+        // (#1917). Refuse it here, where the arithmetic that cannot express it
+        // lives, so the caller takes its minimal-notation fallback.
+        //
+        // `normalize_mito` already gates this at its own axis, for #129, and
+        // states the reason to gate rather than lean on a downstream
+        // fail-safe: the fail-safes hold only for some window sizes. The `g.`
+        // axis never got the same gate, and the ways it failed are all
+        // functions of `start - end` against `window_size`:
+        //
+        // | `start - end`      | before this guard                            |
+        // |--------------------|----------------------------------------------|
+        // | `<= window`        | the fetch succeeds and the subtraction is     |
+        // |                    | fine, so it fails DEEPER, and how depends on  |
+        // |                    | the edit: `dup` slices `ref_seq[start-1..end]` |
+        // |                    | backwards and panics, while `inv` returns `=` |
+        // |                    | — a WRONG ANSWER, not a panic                 |
+        // | `(w, 2w]`          | panic — `end - window_start` underflows below |
+        // | `> 2 * window`     | this fallback, reached by accident: the fetch |
+        // |                    | is `[start-w, end+w)`, which inverts, and the |
+        // |                    | provider rejects it                           |
+        //
+        // So the widest reversed ranges — including both real-corpus rows in
+        // #1917, at 16,885 and 13,685 bases — already took this exit, and the
+        // guard makes the rest agree with them. **The top band is why this is
+        // an output change and not only a crash fix**: censused over every
+        // parseable reversed form against twelve spans, 6 of 24 answered rows
+        // came back wrong rather than panicking — five `inv` claiming `=` for
+        // an inversion, and one `dup` claiming a coordinate past its own
+        // contig's end. `FERRO_ASSERT_SEQUENCE` and `FERRO_ASSERT_IN_BOUNDS`
+        // are the oracles for exactly those two, and neither fires, because no
+        // corpus in either selection feeds a reversed range.
+        //
+        // The guard also relies on no provider rejecting an inverted read,
+        // which the bottom row shows is not a property of the code.
+        //
+        // **Declining is not clamping.** `min(start, end)` would fix the
+        // underflow and nothing else: `normalize_na_edit` would then receive
+        // `rel_start > rel_end` for *every* reversed range and slice backwards,
+        // turning the top row's panic from a narrow band into the whole class.
+        // Anchoring the window is only half a fix because the shuffle, the
+        // alignment and SPDI are each linear-contiguous — `src/spdi/convert.rs`
+        // refuses wraparound by name in four places for exactly this reason.
+        // Circular-aware shuffle modulo contig length remains #129's follow-up.
+        if start > end {
+            return Ok(None);
+        }
+
         // Whether the contig continues past a window is a question about the
         // contig, so the length is read once rather than per attempt.
         let contig_len = self.provider.get_sequence_length(accession).ok();
--- a/tests/it/issue_1917_reversed_range_window.rs
+++ b/tests/it/issue_1917_reversed_range_window.rs
@@ -1,0 +1,251 @@
+//! Issue #1917 — a reversed range panicked in `normalize_in_grown_window`.
+//!
+//! A reversed range (`start > end`) names a **wraparound**: the span runs off
+//! the 3' end of the reference and resumes at base 1. `normalize_in_grown_window`
+//! fetches a linear window and does linear arithmetic on it, so it cannot
+//! express one. It now refuses a reversed range up front and the caller takes
+//! its minimal-notation fallback, which is the gate `normalize_mito` has
+//! carried at its own axis since #129.
+//!
+//! **The `g.` axis failed three different ways, all functions of `start - end`
+//! against `window_size` (default 100).** The issue reported only the middle
+//! band and generalised it as "any reversed range whose span exceeds
+//! `window_size`", which is wrong in both directions — the widest ranges never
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

Issue #1925, opened 2026-08-14
long_block_inversion can no longer observe MAX_SPLIT_BLOCK: #1899 raised it 1024 -> 4096, and the guard restates the old literal so it still passes


nh13:
`long_block_inversion` is the one corpus family whose point is **size** rather than geometry — it exists so a change gated on block *length* is not measured as a structural zero (#1460). It can no longer observe the cap it is named for, and the guard that exists to prove it can still passes.

## 1. The two long cores are now on the same side of the cap

`long_corpus_sequences` draws `[1024, 1100]` (`examples/dump_normalized_corpus.rs:653`). Its own doc comment states the design:

> Two lengths that straddle the cap the normalizer applies at 1024 bases (`merge::MAX_SPLIT_BLOCK`): one block that fits under it and one that does not, **which is the pair a change to the cap moves between**.

The cap moved. `MAX_SPLIT_BLOCK` is now `MAX_CANONICAL_WINDOW`, i.e. **4096** (`src/normalize/merge.rs:2000`, `:1894`), raised from the literal `1024` by #1899 (`2d8b490b`). Both 1024 and 1100 are under 4096, so the pair no longer straddles anything.

**That raise is mine and I did not update this family with it.** #1899's own body argued the corpus is structurally blind to `MAX_SPLIT_BLOCK` because the sequence-keyed families are 20-mers — which is true of those fourteen and precisely why this family exists. I moved the band out from under the one family built to cover it.

## 2. The guard restates the constant instead of importing it, so it still passes

`the_corpus_emits_a_block_past_the_split_cap` (`:2381`):

```rust
const SPLIT_CAP: u64 = 1024;
...
assert!(longest > SPLIT_CAP, ...);
```

`1100 > 1024` holds, so the guard is green while the live cap is 4096. It has a comment telling the reader to update the literal if the constant moves — which is the failure mode, not a mitigation: **a guard that restates the value it is guarding cannot detect that value changing.** Import `merge::MAX_SPLIT_BLOCK` (or expose it) so the assertion is against the constant itself and this cannot recur.

Four further sites carry the stale `1024`: `:69`, `:643`, `:646`, and `:2381`'s message text.

## 3. Separately and older: this family could never trip the cap anyway

`past_the_split_cap` is the **only** functional reader of `MAX_SPLIT_BLOCK` in `src/` (`merge.rs:2011-2013`; every other occurrence is a comment or a unit-test fixture), and it opens with:

```rust
[...]

PR #1927, opened 2026-08-14
docs: state the "assert the property, never let a count be the property" doctrine

nh13:
`### A generator must account for what it dropped` opens by calling itself "the mechanized half of **a corpus zero is a claim about the corpus, not about the change**" — a doctrine that appears nowhere in the file as a section of its own. So it has been re-derived one instance at a time, by whoever it had just cost a day. This writes it down, with a measured instance from this tree behind each shape.

## The three shapes

**1. A count restated instead of imported.** `the_corpus_emits_a_block_past_the_split_cap` pins `const SPLIT_CAP: u64 = 1024` and asserts the corpus builds a wider block. `MAX_SPLIT_BLOCK` is **4096**. The long cores are `[1024, 1100]`, chosen — the comment says so — as "the pair a change to the cap moves between"; the cap moved, both landed on the same side, and `1100 > 1024` kept it green. The guard carries a comment telling the reader to update the literal by hand, which is the defect rather than a mitigation. Filed as #1925.

**2. A zero whose instrument cannot vary the property.** `dump_normalized_corpus.rs` has been blind six times, and its own header records why the list keeps growing: fixing one blindness does not reveal the next.

**3. A rule generalised from the one case reproduced.** #1917 stated its extent as "any reversed range whose span exceeds `window_size`" and was wrong in **both** directions — the widest spans never panicked, the narrowest panicked elsewhere, and the issue's own two real-corpus rows sat in the band that never failed.

Plus two ways a number arrives already wrong: **right for a reason that is not the code** (#1917's widest band passed both before and after the fix, because the provider rejects an inverted read — a property of the provider, not the change), and **two "independent" confirmations sharing one stale source** (two agents confirming `merge.rs` line numbers, both read from a working tree ~950 lines stale).

## Then the practice

Assert the property in the words you would explain it in; if a count really is the right assertion, say what it counts and against which denominator; **prove the guard can fail** by sabotaging it once and watching it go red; and when reporting a number, report what it is made of.

## Scope

Documentation only. Adds no table, so neither `claude_md_adjudication_tables` anchor (`| record | what is open |`, `| record | ruling |`) moves — verified, both still unique. `claude_md_adjudication_tables`, `ruling_citation_currency` and `clause_ruling_index` run green (22/22).

Representation-Change: none. Documentation only; no watched directory is touched and no code changes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -975,6 +975,67 @@ is **spec-explicit > Mutalyzer > our judgement**. Where the spec is explicit and
 that is a deliberate divergence and it gets a record saying so — otherwise the next person
 measures Mutalyzer, finds a mismatch, and "fixes" a conformance decision.
 
+### Assert the property. Measure the count. Never let a count BE the property
+
+The section below is the mechanized half of this one, and this one had never been written down —
+so it kept being re-derived, one instance at a time, by whoever it had just cost a day.
+
+**A guard that pins a number is a change detector for that number.** It is a guard for the
+property only for as long as the two agree, and nothing makes them agree. When they drift, the
+guard follows the number, stays green, and reads as coverage.
+
+Three shapes, in ascending order of how convincing the wrong answer looks.
+
+**1. A count restated instead of imported.** `the_corpus_emits_a_block_past_the_split_cap`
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1930, opened 2026-08-14
feat: add sequence_normalize() helper

TimD1:
Summary
This PR adds sequence-based normalization for an HGVS variant (with optional standard normalize() afterwards), guaranteeing confluence when normalizing string-derived variants to a single representation.

### Implementation
* sequence_normalize() wraps to_sequence() and from_sequence(), handling logic revolving around iterative doubling of the padding to ensure convergence to a single representation.
* placement_bounded_by_window was split into bounded_at_start and bounded_at_end to assist in determine whether the window needs to keep growing or not

### Issues
* closes #1931 

### Notes
Representation-Change: none

PR #1933, opened 2026-08-14
fix: allow ref span insertions when converting to SPDI

### LINKED ISSUE #1932: Allow reference span based insertions when converting to SPDI format
### Summary
SPDI conversion currently disallows insertions of the form `g.6_7ins156_183`, stating that non-literal HGVS insertions strings cannot be converted to `(ref, alt)` format. This is not the case when a `provider` exists that covers the range of the inserted sequence.

### Task
When possible, convert this `156_183` sequence to the literal base string instead of erroring out.

TimD1:
### Summary
Allows reference-span insertions when converting to SPDI, allowing `sequence_normalize()` to operate on such inputs.

### Issues
* closes #1932 

### Notes

Representation-Change: a same-reference position-range insert (`g.100_101ins50_57`, its `inv` and `delins` spellings, and the single-part bracketed form) now resolves to its literal bases through `hgvs_to_spdi` instead of being refused.
  No normalized HGVS string moves — nothing under `src/normalize/` reaches `src/spdi/convert.rs`, and the normalizer already expanded these payloads through `rules::expand_inserted_sequence`; what changes is which descriptions are convertible, not what a convertible one converts to.
  18 of the 9,949,738 rows across the four release-asset corpora carry the shape. Every affected input was previously rejected, so no consumer holds a stored old value; the visible consequence is those rows leaving `unkeyable` for a `group_by_spdi_key` bucket.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `Default::default()` was the only reachable constructor — the two documented knobs could not be
   set by any downstream caller.
 
+### Fixed
+
+- **Reference-span insertions are now resolved rather than refused.** An insertion whose bases are
+  named by position in the same reference — `g.100_101ins50_57`, and its inverted form
+  `g.100_101ins50_57inv` — was rejected as "insertion sequence is not a literal sequence" whenever
+  a provider-backed path (`sequence_normalize`, `to_spdi`, `canonical_spdi`, apply-to-reference)
+  needed its bases. It now reads those bases from the reference exactly as `del`/`dup`/`delins`
+  read their omitted bases, so `sequence_normalize` re-derives the literal `g.100_101insCGTA…`
+  form. `delins` with a same-reference range insert is resolved on the same path.
+
 ## [0.13.1](https://github.com/fulcrumgenomics/ferro-hgvs/compare/v0.13.0...v0.13.1) - 2026-08-08
 
 ### Representation changes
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
@TimD1 — we'd like to help get this stack over the line. "Allow edits by maintainers" is currently off on all three PRs (#1930, #1933, #1935), so we can't push to the branches. Could you tick that checkbox on each? It's in the right-hand sidebar of the PR page.

Where things stand:

**#1933 and #1935** were red on `Representation change declared` — neither description carried a `Representation-Change:` trailer, and `src/spdi/` is one of the watched prefixes. Description edits don't require maintainer edits, so we've added a trailer to each and both checks are now green. The verdict we landed on: no normalized HGVS string moves, because nothing under `src/normalize/` reaches `src/spdi/convert.rs` — the normalizer resolves insert payloads through its own `rules::expand_inserted_sequence` — so what changes is which descriptions are *convertible* to SPDI, not what a convertible one converts to. Every affected input was previously rejected, so it's a free change rather than a migration. Counts are over the four release-asset corpora (9,949,738 rows). Please read the wording and change it if you disagree — it's your PR and your call.

**#1930** is conflicting, but only in `tests/it/main.rs`: `main` added `mod ruling_guard_field;` into the same alphabetical slot as your `mod sequence_normalize;`. Keeping both lines in order resolves it, and every other file auto-merges (`git merge-tree --write-tree --name-only origin/main <head>` reports that one path). It's a one-line rebase we're happy to run once edits are enabled.

Since the three are stacked (#1930 → #1933 → #1935), #1930's rebase is what unblocks the chain.

nh13:
Closing in favour of #1966, which carries this work commit-for-commit with authorship preserved.

TimD1:
Didnt see this til now. Sounds good, thanks!

PR #1935, opened 2026-08-14
fix: allow repeat/complex variants to be converted to SPDI

### LINKED ISSUE #1934: Allow repeats when converting to SPDI format
### Summary
Variants such as `8_9insA[3]` are currently rejected when converting to SPDI because `A[3]` not a literal sequence.

### Task
This format should be expanded from `A[3]` to a literal `AAA` when converting to SPDI format to enable `sequence_normalize()` to support variants of this form.

TimD1:
### Summary
- **Exact tandem-repeat insertions are now expanded rather than refused.** An insertion whose bases
  are named by a spelled unit and an exact copy count — `g.100_101insC[4]`, `g.100_101insCG[3]` —
  was rejected as "not a literal sequence" whenever a provider-backed path needed its bases. It now
  expands the unit the named number of times, exactly as the short-form repeat edit does, so
  `sequence_normalize` re-derives the literal `g.100_101insCCCC` / `g.100_101insCGCGCG` form (the
  same `MAX_REPEAT_EXPANSION_BASES` cap applies). `delins` with an exact repeat insert is resolved
  on the same path. Uncertain or range counts (`insC[10_15]`, `insC[?]`) remain undetermined and
  still decline.
- **Compound-insert brackets mixing shapes are now resolved rather than refused.** A bracketed
  insertion whose parts combine a literal with a same-reference coordinate span or an exact tandem
  repeat — `g.185_201delins[T;213_271]`, `g.100_101ins[C[3];A]`, `g.100_101ins[213_271inv;A]` —
  was rejected as "Unsupported variant type … cannot apply to reference" whenever a provider-backed
  path needed its bases, because only single-part brackets were read out. Each part is now resolved
  through the same leaf that resolves it on its own — a span reads from the reference (and
  reverse-complements for the `…inv` form), an exact repeat expands its unit — and the parts are
  concatenated in written order into one contiguous inserted sequence, so a compound insert encodes
  to SPDI just as a plain literal does. This applies on both the `ins` and `delins` arms. A part
  whose bases are genuinely undetermined — an uncertain or range repeat count, a CDS position range
  carrying intronic offsets, or an external reference SPDI cannot dereference — still declines the
  whole insert.

### Issues
* closes #1934 

### Notes

Representation-Change: two further insertion payloads now resolve through `hgvs_to_spdi` instead of being refused — an exact tandem repeat expanded under the existing `MAX_REPEAT_EXPANSION_BASES` cap, and a bracketed compound insert whose parts mix literals, same-reference spans and exact repeats, each part resolved by the leaf that resolves it alone. A part whose bases are undetermined (uncertain or range count, intronic CDS offsets, external accession) still declines the whole insert.
  No normalized HGVS string moves — nothing under `src/normalize/` reaches `src/spdi/convert.rs`.
  Across the four release-asset corpora (9,949,738 rows), 3 rows carry a bare exact-repeat insert and 52 a bracketed compound insert, of which roughly ten have all-same-reference parts; the rest still decline. Every affected input was previously rejected.


nh13:
@coderabbitai pause

nh13:
Closing in favour of #1967, which carries this work commit-for-commit with authorship preserved.

PR #1938, opened 2026-08-14
fix(corpus): straddle the gate that fires, and import the bound instead of restating it

### LINKED ISSUE #1925: long_block_inversion can no longer observe MAX_SPLIT_BLOCK: #1899 raised it 1024 -> 4096, and the guard restates the old literal so it still passes

nh13:
Closes #1925.

`long_block_inversion` is the one corpus family whose point is **size** rather than geometry (#1460). It had stopped measuring size, and the guard that exists to say so kept passing.

## What broke, and why nothing noticed

#1899 derived `MAX_SPLIT_BLOCK` from `MAX_CANONICAL_WINDOW`, taking it **1024 → 4096**. The long cores stayed `[1024, 1100]`, chosen — the comment said so — as "the pair a change to the cap moves between". Both landed under the new cap.

`the_corpus_emits_a_block_past_the_split_cap` did not notice, because it opened:

```rust
const SPLIT_CAP: u64 = 1024;
```

with a comment telling a future reader to update it by hand if the constant moved. `1100 > 1024` held, so it stayed green. **A guard that restates the value it guards cannot observe that value changing**, and the instruction to notice is the defect written down, not a mitigation.

That raise is mine, and so is failing to move this family with it.

## Three changes, ascending by what they buy

**1. Import the bound.** `MAX_CANONICAL_BLOCK` is exported and the guard imports it. Deliberately *not* `dev`-gated: a constant present in only some builds re-creates the gap it exists to close.

**2. Target the gate that actually fires.** `past_the_split_cap` is the only functional reader of `MAX_SPLIT_BLOCK` in `src/` — every other occurrence is a comment or a unit-test fixture — and it additionally requires `reference.len() != result.len()`. This family emits a whole-core `inv` and the equal-length substitutions competing with it, so it **could never have tripped that cap at any size**, even with the right number. #1899's own doc says the shipping path cannot reach it at all, the window test refusing first. So the guard now names the gate that does fire: `MAX_CANONICAL_WINDOW - 2 * CANONICAL_PAD` = 3840.

**3. Straddle the changed INTERVAL, not the core length.** This is the part reasoning got wrong. Cores of exactly the limit and `+2` produce **byte-identical rows** — measured, both `g.257_4096inv` and `g.257_4098inv` collapsing to four substitutions — because the window is built from the first differing base to the last, and `near_palindromic_core` sets those `EDGE_PERTURBATION` bases in from each end. A core of `n` presents an interval of `n - 20`, so cores straddling 3840 give intervals comfortably under it. The corpus straddled nothing while looking like it did.

Offsetting by `2 * EDGE_PERTURBATION` puts the intervals at the limit and one past it, and the two rows now differ in kind:

| core | interval | `g.` inv row |
|---|---|---|
| `nearpalindrome_3860` | 3840 | `g.257_4116inv` → `g.[267T>A;2179C>G;2194C>G;4106T>A]` — **derived** |
| `nearpalindrome_3862` | 3842 | `g.257_4118inv` → `g.267_4108inv` — **refused**, trimmed only |

`EDGE_PERTURBATION` is now a named constant rather than a literal `10` inside the core builder, because it is load-bearing for this relation.

## Sabotaged, and one sabotage found a hole

Restoring the 1024-era cores `[1024, 1100]` makes the guard fail:

```
the longest block the corpus builds is 1100 bases, under the 3840-base split cap
  — so every row takes the same path on both sides of a change to it
```

That is the message that should have fired when #1899 landed.

The second sabotage — dropping the `2 * EDGE_PERTURBATION` offset, so the cores sit at exactly `MAX_CANONICAL_BLOCK` and `+2` — **passed, and should not have.** Both remaining assertions read a *length* (the span the input names, i.e. the core) while the gate reads the *changed interval*, `2 * EDGE_PERTURBATION` bases narrower; with the offset gone both intervals land under the limit, both cores derive identically, and the guard stayed green while the corpus straddled nothing. Part 3 was the only part nothing asserted — the same defect this PR is about, one level up.

The guard now pins the straddle itself: among the past-cap rows that actually normalized, at least one core must have been re-derived into substitutions and at least one must have kept its `inv`, and not the same core for both. It fails from both sides — `(re-derived: {"nearpalindrome_3842"}, kept as inv: {})` with the offset dropped, and `(re-derived: {}, kept as inv: {"nearpalindrome_3864", "nearpalindrome_3866"})` with it widened so both cores clear the gate.

## What is deliberately NOT claimed

`MAX_CANONICAL_BLOCK`'s doc says so in its own words: the guard and the constant read one definition, so they cannot disagree, and **nothing pins the constant against the gate behaviourally**. The check that would — driving `canonicalize_from_sequence` at exactly this width and one base wider, asserting accept then refuse — is not written, and `canonicalize_from_sequence` has no unit-test harness in `merge.rs` today. Stated rather than implied, because a derived constant reads as though it has been verified.

## Verification

- Full gate **10,810 / 10,810**, `FERRO_MANIFEST` armed, `FERRO_REQUIRE_BULK_FIXTURES=1`, zero `Skipping:` lines. Exit status read back from the `EXIT=` line in the log, not from the job notification.
- The corpus example's own 42 tests pass; the dump runs end to end (85,642 rows, ~20 s).
- Python binding suite: 1,000 passed, 8 skipped — `cargo nextest` does not reach it and this touches `src/normalize/`.

Representation-Change: none. Measured, not assumed. This adds a `pub const` and its re-export under `src/normalize/`, and otherwise changes only `examples/dump_normalized_corpus.rs`, which is the measuring instrument rather than a code path ferro normalizes through - the new constant has no `src/` reader, so no logic under any watched directory changes behaviour, and the full gate passes against every pinned value including both spec_conformance_axis censuses and case_harvest. The corpus dump itself does change, deliberately and only in the two `long_block_inversion` cores: the family's row count and shape are unchanged, the two `reference` labels move from nearpalindrome_1024/1100 to nearpalindrome_3860/3862, and one of the two rows now reaches a refusal it previously did not. That is the instrument being repaired, not ferro's output moving - any before/after dump comparison spanning this commit should expect those rows to differ and no others.

### DIFF
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -65,8 +65,9 @@
 //!    quote the zero.
 //! 4. **…and only at the scale it builds them** (#1460). The families are drawn
 //!    against 20-mers, so a change gated on *length* rather than on shape sees the
-//!    same path on both sides no matter how many families exist. #1403 caps a
-//!    partition at `MAX_SPLIT_BLOCK` (1024) and measured a guaranteed zero here
+//!    same path on both sides no matter how many families exist. #1403 capped a
+//!    partition at `MAX_SPLIT_BLOCK` — **1024 at the time**, 4096 since #1899
+//!    derived it from `MAX_CANONICAL_WINDOW` — and measured a guaranteed zero here
 //!    even after the conflict families landed. `long_corpus_sequences` is the
 //!    answer, and it is deliberately narrow: two cores, one family, 16 rows. Adding
 //!    scale is expensive in a way adding a family is not — check the dump cost
@@ -140,7 +141,7 @@ use std::process::ExitCode;
 
 use clap::Parser;
 
-use ferro_hgvs::normalize::{NormalizeConfig, ShuffleDirection};
+use ferro_hgvs::normalize::{NormalizeConfig, ShuffleDirection, MAX_CANONICAL_BLOCK};
 use ferro_hgvs::reference::transcript::{Exon, GenomeBuild, ManeStatus, Strand, Transcript};
 use ferro_hgvs::reference::MockProvider;
 use ferro_hgvs::{parse_hgvs, Normalizer};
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #1942, opened 2026-08-14
Allow substitutions upstream flush with reference-consuming variants

### LINKED ISSUE #1941: Allow insertions flush to reference-consuming variants for SPDI conversion
### Summary
A variant of the form `g.[ins6_7AC;7T>G]` is currently rejected due to overlap. This is not the case for variants such as `g.[7T>G;7_8insAC]`.

### Resolution
Fix the asymmetry since these variants don't actually overlap.

TimD1:
### Summary
- **A pure insertion flush against the 5' edge of another member now applies rather than reading as
  an overlap.** When applying a cis allele to its reference, a zero-width insertion abutting the 5'
  edge of a substitution, deletion, or inversion — `g.[10_11insAC;11C>T]`, `g.[10_11insAC;11_13del]`,
  `g.[10_11insAC;11_13inv]` — was declined as an overlapping-member conflict. Such a combination is
  well defined: the base-claiming member is applied first, then the insertion lands at the junction
  5' of it. `apply_triples` (shared with `EquivalenceChecker`, `canonical_spdi`, and
  apply-to-reference) now walks the members with a `claimed_from` watermark and orders longer
  deletions first among members at one position, so a flush insertion sitting *at* the watermark is
  accepted while one *interior* to a span (`g.[10_13del;11_12insAC]`) still reaches past it and
  declines. The two applier walks that had diverged over this shape are unified onto the one walk,
  removing the standalone `triples_are_disjoint` guard.
  
 ### Issues
* closes #1941

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   whose bases are genuinely undetermined — an uncertain or range repeat count, a CDS position range
   carrying intronic offsets, or an external reference SPDI cannot dereference — still declines the
   whole insert.
+- **A pure insertion flush against the 5' edge of another member now applies rather than reading as
+  an overlap.** When applying a cis allele to its reference, a zero-width insertion abutting the 5'
+  edge of a substitution, deletion, or inversion — `g.[10_11insAC;11C>T]`, `g.[10_11insAC;11_13del]`,
+  `g.[10_11insAC;11_13inv]` — was declined as an overlapping-member conflict. Such a combination is
+  well defined: the base-claiming member is applied first, then the insertion lands at the junction
+  5' of it. `apply_triples` (shared with `EquivalenceChecker`, `canonical_spdi`, and
+  apply-to-reference) now walks the members with a `claimed_from` watermark and orders longer
+  deletions first among members at one position, so a flush insertion sitting *at* the watermark is
+  accepted while one *interior* to a span (`g.[10_13del;11_12insAC]`) still reaches past it and
+  declines. The two applier walks that had diverged over this shape are unified onto the one walk,
+  removing the standalone `triples_are_disjoint` guard.
 
 ## [0.13.1](https://github.com/fulcrumgenomics/ferro-hgvs/compare/v0.13.0...v0.13.1) - 2026-08-08
 
--- a/src/spdi/apply.rs
+++ b/src/spdi/apply.rs
@@ -467,7 +467,12 @@ pub(crate) fn apply_triples(
     let ref_bytes = reference.as_bytes();
     let mut bytes = ref_bytes.to_vec();
     let mut ordered: Vec<&SpdiVariant> = triples.iter().collect();
-    ordered.sort_by_key(|t| (std::cmp::Reverse(t.position), std::cmp::Reverse(t.deletion.len())));
+    ordered.sort_by_key(|t| {
+        (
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
Closing in favour of #1968, and this one needs more explanation than a pointer — the work here was measured, not set aside.

**#1837 landed the same behaviour first.** It merged into `main` as `4ac4cb3d` on 2026-08-14 at 21:51; the change here was committed at 15:08 the same day, against a base that did not yet contain it. Both fix #1831 — the applier's flush-insertion verdict depending on authored member order — by different mechanisms: #1837 keeps `triples_are_disjoint` and fixes the sort, while this PR replaces the predicate with an inline `claimed_from` watermark.

**The inversion cases were verified, not assumed.** Both test functions from `f3d382b4` were applied verbatim to unmodified `origin/main`, with none of this PR's source changes, and both pass:

```
PASS  spdi::apply::tests::an_insertion_interior_to_an_inversion_declines
PASS  spdi::apply::tests::an_insertion_flush_against_a_deletion_or_inversion_applies
```

All four cases in the first function — deletion and inversion, in both member orders — assert on `applied.resulting`, and the fixture was checked first so the expected `ACATG` could genuinely fail. So the behaviour this PR describes is already shipped, and its premise ("was declined as an overlapping-member conflict") is no longer true against current `main`.

**The test coverage is preserved.** #1837 delivered the inversion behaviour but never pinned it — `issue_1831_applier_member_order` covers only the substitution and deletion shapes. Both of your test functions now live in #1968, authored to you, with one clause of the doc comment re-pointed at `issue_1831_applier_member_order` because the "above" it referred to exists only in this PR's other commit.

**The `triples_are_disjoint` change is deferred, not dropped.** Removing that predicate is the real contribution left here, and it is being taken up separately so it can be judged on its own rather than as part of a bugfix whose behavioural half has already landed. One correction to how it was framed, since it bears on how it gets reviewed: it is smaller than "unifying the two applier walks". `splice_denoted_sequence` survives either way — it returns a typed `SpliceFailure` and folds the RNA alphabet — and on the SPDI-triple geometry there are five such walks in the tree, not two. #1749's four named sites are all on the HGVS-AST geometry, in `src/normalize/overlap.rs` and `src/hgvs/variant.rs`, and none of them is touched by this change.

For what it is worth, the watermark itself was measured before being set aside, by two independent differential enumerations over millions of permuted two-, three- and four-member triple sets. It and `triples_are_disjoint` agree on every input reachable through either of `apply_triples`' callers — they are the same overlap test written from opposite ends of the position-sorted list, and the sort's longer-deletion-first tie-break is what makes both sound. The one input on which they differ is a `u64` overflow at the very top of the address space, where `triples_are_disjoint` declines a pair of genuinely disjoint members because it adds in absolute coordinates; the watermark, working window-relative, gets it right. It is unreachable in practice — `apply_triples`' callers decline on that overflow first — but the direction is worth recording: where the two designs part, yours is the correct one.

Two things that turned up alongside that and are not yours to fix, noted so they are not lost: `triples_are_disjoint`'s "running maximum ... is what makes this complete" comment describes a hazard that cannot occur, since the walk is in ascending position order and short-circuits on the first spanned member; and `main` already runs your watermark design in `splice_denoted_sequence`, so `apply_triples` adopting it would be convergence on something the file already contains rather than a new approach. Both are #1749's business.

Thanks — the inversion coverage was a real gap and it is landing because you found it.

PR #1944, opened 2026-08-14
test: enumerate minimal alignments, so the unchanged-base rule can be computed

nh13:
`rulings[unchanged-is-read-over-every-minimal-alignment]` decides that **a reference base is unchanged iff every minimal alignment of the block matches it**. Nothing in the repo could compute that. The rule quantifies over *all* minimal alignments and nothing enumerated them, so every question of the form "is this base unchanged?" was re-derived by hand — and the record itself exists because two agents re-derived it differently within one hour on 2026-08-14. An intersection over a set nobody enumerates is not a checkable claim.

This adds the instrument, its pinned cases, and its properties. It wires into nothing.

## What it is

`tests/it/common/minimal_alignment.rs` takes a `(reference, alternate)` byte pair and an explicit cost model, and returns the minimal edit cost, **every** optimal alignment, the reference offsets matched in all of them (the record's "unchanged" set), and how many alignments there were.

```rust
let report = minimal_alignments(b"CAG", b"AGA", CostModel::Levenshtein)?;
report.cost();        // 2
report.count();       // 1
report.is_unique();   // true
report.unchanged();   // [1, 2]
report.alignments();  // [[Del, Match, Match, Ins]]
```

The count is reported because "unique" and "several that happen to agree" are different situations, and that distinction is exactly what separates this notion from the cell-based one the record contrasts it with.

It lives in test-support rather than `src/` for three reasons, all on the module: nothing under `src/` calls it or should; the production alignment DAG next door implements a deliberately *different* (cell-based) notion, and putting a second one beside it invites the confusion the record was written to end; and enumeration is exponential by construction, so it is unfit for a hot path however it were written.

## The cap is an error, never a truncation

Past `DEFAULT_ALIGNMENT_CAP` (65536) the call returns `Err(CapExceeded)` rather than a partial set. The unchanged set is an **intersection**, so dropping alignments can only *add* offsets to it — a silent truncation would report bases as unchanged that are not, which is wrong in exactly the direction the record exists to prevent. The cost is still reported, since the DP grid is polynomial and always completes.

65536 fits a 7-base all-mismatch block. The growth is the central Delannoy numbers under substitution cost 2 — 321 at n=4, 8989 at n=6, 48639 at n=7, 265729 at n=8 — so one more base costs roughly six times the budget and no reachable cap changes which blocks this can answer for. The cap's job is to fail loudly, not to be generous.

## The record does not state its cost model, and the two candidates disagree

This is the most valuable thing the instrument makes visible, and it is **not adjudicated here**.

Two things point at substitution cost 1: the record's own worked example (`CAG -> AGA`, "position-wise: cost 3" over three substitutions), and `issue_1539_split_member_separation::forced_unchanged_columns` — the closed-form detector the record names as where the ruling came from — which hardcodes it. Neither is a ruling.

And the two models disagree on a real reported block. On `1420-v4` (`ATGC -> GCTG`, the `TEMPLATE` block at `g.21_24`):

| model | cost | alignments | unchanged (block) | unchanged (genomic) |
|---|---:|---:|---|---|
| substitution costs 1 | 3 | 2 | `{1, 2}` | `g.22`, `g.23` |
| substitution costs 2 | 4 | 6 | `{2}` | `g.23` |

Under one model `g.22` is an unchanged base and the changes either side of it are "separated by one or more nucleotides", so `general.md:34` describes them individually. Under the other they are consecutive. That is a difference in what the spec clause says about this block, from a parameter the record never fixes.

Both answers are pinned in `the_two_cost_models_disagree_on_1420_v4`, with a doc comment saying the record is silent and that the test is the record of that gap. Picking a winner is an adjudication and belongs in the ledger, per `rulings[adjudication-precedence-order]` — not in a test module.

## What is pinned

Every case the record argues from, plus the degenerate ends. The three `1420` blocks are read off the real `TEMPLATE` contig in `reported_confluence_pairs`, not retyped.

| case | model | cost | alignments | unchanged |
|---|---|---:|---:|---|
| `CAG -> AGA` (record's headline) | both | 2 | 1 | `{1, 2}` |
| `GACA -> AGAT` (forced the column/cell distinction) | sub=1 | 3 | 3 | `{1}` |
| `GACA -> AGAT` | sub=2 | 4 | 9 | `{}` |
| `TTGC -> ATTG` (`1420-v2`, `g.38_41`) | both | 2 | 1 | `{0, 1, 2}` |
| `ATTG -> CATT` (`1420-v3`, `g.37_40`) | both | 2 | 1 | `{0, 1, 2}` |
| `ATGC -> GCTG` (`1420-v4`, `g.21_24`) | sub=1 / sub=2 | 3 / 4 | 2 / 6 | `{1, 2}` / `{2}` |
| `ACGTACGT -> ACGTACGT` | both | 0 | 1 | all |
| `AAAA -> TTTT` | sub=1 / sub=2 | 4 / 8 | 1 / 321 | `{}` |
| empty, pure del, pure ins | both | — | 1 | `{}` |

## Two things the instrument found

**The record says `GACA -> AGAT` has "the two cost-3 alignments"; there are three.** Their matched-reference sets are `{1, 3}`, `{0, 1}` and `{0, 1}` — two of the three agree, which is the likeliest reading behind the "two". Pinned rather than resolved, because the record's conclusion is unaffected: offset 1 is matched in all three, so the intersection is `{1}` on either count.

**The position-wise reading and the minimal-alignment reading are incomparable, not merely different.** Neither set contains the other. Brute force over every `{A, C}` pair up to length 5 under substitution cost 1 finds 178 counterexamples to one containment and 250 to its converse; `CAG -> AGA` is among the first, with `unchanged = {1, 2}` and no position-wise match at all. The same holds between the two cost models, in both directions. Both are pinned with witnesses in `neither_notion_contains_the_other` and `neither_cost_model_bounds_the_other`. This is a stronger statement of the record's headline claim than "they can differ", and it means a property test written on the natural guess that one bounds the other would be red on roughly one generated pair in four.

## Properties

`minimal_alignment_enumeration_proptest` sweeps both models over random pairs and asserts: every emitted alignment is well formed and its cost recomputed from the columns equals the reported minimum; the alignments are pairwise distinct; `unchanged` is exactly the intersection of the per-alignment matched sets; the cap refuses iff the set does not fit and otherwise changes nothing; reversing both blocks mirrors the answer; `cost(sub=1) <= cost(sub=2) <= 2 * cost(sub=1)`; and on equal-length blocks the minimal cost never exceeds the position-wise cost.

The eighth is a **cross-oracle**: `forced_unchanged_columns` reaches the same set in `Θ(n·m)` without materialising an alignment, so agreeing with it is agreeing with the record's working definition by a genuinely different route — one side never enumerates, the other never inspects the grid. It is made `pub(crate)` for that and is otherwise untouched.

## The soak, re-checked rather than assumed

`ci.yml`'s soak selects by name, so a `*proptest*` module lands in a 1M-case job without anyone choosing it — the comment there asks for two claims to be re-checked, and they were. The properties are pure functions of a generated byte pair: no provider, no fixture, no generated spec artifact. Cost was measured too, because the enumerator is exponential: `PROPTEST_CASES=125000` over the module costs **56.7 s wall in the test profile**, i.e. before that job's ~5.4x optimization, bounded by the generator's own `MAX_BLOCK_LEN` of 6. The comment now records all of it.

## Gate

`cargo fmt --all --check`, `cargo clippy --all-features --all-targets -- -D warnings`, `cargo nextest run --features dev --lib --test it`, and `actionlint` on the workflow change — all clean.

Representation-Change: none. This adds a test-support utility and its tests; it wires into no normalization path, moves no output, and re-pins no census.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1439,11 +1439,11 @@ jobs:
     name: Idempotency soak (${{ matrix.shard }}/8)
     # Only the optimized archive, NOT `test-build`. Every test the selection
     # below reaches lives in `tests/it/{cis_allele_confluence,
-    # from_sequences, normalize_idempotency}_proptest.rs`, and none reads the
-    # generated spec fixture or shells out to the staged generators — so the
-    # shards need neither of those artifacts, and waiting on the job that
-    # produces them would put the dev-profile compile back on the soak's
-    # critical path.
+    # from_sequences, minimal_alignment_enumeration,
+    # normalize_idempotency}_proptest.rs`, and none reads the generated spec
+    # fixture or shells out to the staged generators — so the shards need
+    # neither of those artifacts, and waiting on the job that produces them
+    # would put the dev-profile compile back on the soak's critical path.
     #
     # `from_sequences_proptest.rs` joined this selection by being named
     # `*proptest*`, which is worth stating because the filter is by name: a new
@@ -1452,6 +1452,15 @@ jobs:
     # fixture — so the two claims above still hold, and it earns its place here
     # (`the_direction_does_not_decide_describability` held for 256 cases and
     # failed at 20,000). Re-check both claims when adding another.
+    #
+    # `minimal_alignment_enumeration_proptest.rs` joined the same way, and was
+    # re-checked rather than assumed: its eight properties are pure functions of
+    # a generated `(reference, alternate)` byte pair, reading no provider and no
+    # fixture. Its cost was measured too, because the enumerator it exercises is
+    # exponential in block length — `PROPTEST_CASES=125000` over the module
+    # costs 56.7 s wall in the *test* profile, i.e. before this job's ~5.4x
+    # optimization. That is bounded by its own `MAX_BLOCK_LEN` of 6; raising it
+    # is a soak-cost decision, not only a coverage one.
     needs: test-build-soak
     runs-on: ubuntu-latest
     strategy:
--- a/tests/it/common/minimal_alignment.rs
+++ b/tests/it/common/minimal_alignment.rs
@@ -1,0 +1,495 @@
+//! Enumerate every minimal alignment of a `(reference, alternate)` block, and
+//! report which reference bases are matched in **all** of them.
+//!
+//! # What this is for
+//!
+//! `rulings[unchanged-is-read-over-every-minimal-alignment]` (in
+//! `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json`) decides
+//! that
+//!
+//! > a reference base is unchanged iff it is matched in EVERY MINIMAL
+//! > alignment of the block.
+//!
+//! Until this module existed, that rule was applied by reading and by hand
+//! argument: nothing in the repo could *compute* it. An intersection over a set
+//! nobody enumerates is not a checkable claim, and the arithmetic has been got
+//! wrong more than once. [`minimal_alignments`] enumerates the set and takes
+//! the intersection, so "is this base unchanged?" has an answer a test can
+//! assert on.
+//!
+//! # This is an instrument, not a normalizer input
+//!
+//! Nothing under `src/` calls this and nothing should. It lives in
+//! `tests/it/common/` for three reasons:
+//!
+//! - **No production consumer.** Normalization decides its own partition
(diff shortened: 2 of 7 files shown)


nh13:
@coderabbitai pause

PR #1945, opened 2026-08-14
test(rulings): judge the guard scan on its reach, not on what the tree holds

### LINKED ISSUE #1907: the_guard_scan_reads_the_tree_it_claims_to fails when the last #[ignore]d guard is legitimately removed
`ruling_citation_currency::the_guard_scan_reads_the_tree_it_claims_to` asserts that the live tree contains at least one `#[ignore]`d guard-shaped function:

```rust
sites.values().flatten().any(|site| site.ignored)
```

The intent is anti-vacuity — make sure the scanner's ignore-reading arm is actually exercised rather than silently inert. That intent is right, and it is the same class of care this repo applies elsewhere.

The coupling is the problem. The assertion conflates two different propositions:

- *the scanner can read `#[ignore]`* — a property of the scanner, and what the test is named for; and
- *the repository currently contains an ignored guard* — a property of the tree, which changes as work lands.

Because it tests the second, **the check fails whenever someone closes the last open question** — which is an outcome the project actively wants.

### It is already blocking

`origin/main` contains exactly one `#[ignore]`d guard-shaped function: `the_decided_target_is_convergence_on_the_near_side` in `tests/it/spec_worked_examples.rs`, which is #1621's acceptance criterion. PR #1845 legitimately un-ignores it — that is the PR's whole point — taking the count to zero.

Measured as a controlled A/B: base alone **10/10 pass**; base plus #1845, **1 failed**.

The failure mode is worth noting for its own sake. #1845 reads 42 pass + 1 skipping across 43 rows, `MERGEABLE` / `CLEAN`, because its head was pushed at 05:50 UTC and this check landed at 08:09 UTC — 2h19m later. Nothing in #1845's own CI could ever have seen it. The merge queue does catch it, since it tests the projected `main`, so nothing lands broken; the practical outcome is that a PR is rejected by a failure in a file it never touched.

### Suggested direction

Exercise the ignore-reading arm against a literal fixture rather than the live tree, the way the sibling test `the_assertion_detector_reads_helper_calls_as_assertions` already exercises `body_asserts`. That verifies the scanner — which is what the test claims to do — and is strictly more robust, because it keeps working when the tree has no ignored guards left.

An alternative, if the tree scan is wanted for its own sake, is to keep it but permit zero via a named exemption that is itself asserted on, so the empty case is declared rather than accidental.

Please do **not** resolve it by re-adding the `#[ignore]` to `the_decided_target_is_convergence_on_the_near_side`; that would defeat #1621.

### Provenance

Found during a dual-review pass over #1845, which is blocked on it. Not caused by #1845, and not a defect in the scanner itself — only in what this one assertion is pointed at.

nh13:
Closes #1907.

`ruling_citation_currency::the_guard_scan_reads_the_tree_it_claims_to` asserted that the live tree contains at least one `#[ignore]`d guard-shaped function. That makes **legitimately removing the last one a failure** — the assertion is about what the tree happens to hold, not about whether the scan works.

`origin/main` holds exactly one such guard, `the_decided_target_is_convergence_on_the_near_side` (#1621's acceptance criterion), and un-ignoring it is #1845's entire purpose. Measured: base alone 10/10 pass; base + #1845, 1 failed.

## What changes

The per-file scan is extracted from `guard_sites()` into a pure `guard_sites_in(path, text)`, so the `#[ignore]` reading is exercised against a **literal fixture** rather than against whatever the tree currently contains. That is the idiom this file already uses for `body_asserts` (`the_assertion_detector_reads_helper_calls_as_assertions`), and the one `ruling_guard_field.rs` uses for its own scanner (`the_scanner_reads_test_and_ignore_attributes`, #1883).

`the_guard_scan_reads_the_tree_it_claims_to` keeps `sites.len() >= 200` and `any(asserts)`, and gains a **per-root reach** assertion: every `SCAN_ROOTS` entry must yield `.rs` files. The `any(ignored)` assertion is gone.

`any(asserts)` is kept deliberately: a tree in which no guard asserts anything is not an outcome legitimate work can produce, unlike the ignored-guard count.

## Why this is strictly stronger, not merely decoupled

Sabotage, four cases, each restored by content (md5 verified):

| sabotage | result |
|---|---|
| A — `#[ignore]` seen but never recorded | fixture check fails alone, `EXIT=100` |
| B — flag reset by an intervening doc comment/attribute | fixture check fails alone, `EXIT=100` |
| C1 — filter drops everything under `tests/` | reach check fails with its own message, `EXIT=100` |
| C2 — a scan root pointed at `docs/` (a real directory with no `.rs`) | reach check fails, `EXIT=100` |

**The old assertion would have passed C2** — `tests/` still supplied the ignored guard while half the scan read nothing. It would also have passed B, since the real ignored site has nothing between `#[ignore]` and `fn`.

## Merge order

This must merge **before** #1845, or the queue builds #1845 against a `main` whose guard still demands an ignored guard that #1845 removes.

Representation-Change: none. Tests only — one file, `tests/it/ruling_citation_currency.rs`. No file under `src/normalize/`, `src/hgvs/`, `src/spdi/`, `src/project/`, `src/reference/` or `src/error_handling/` is touched, and no normalized output can move.

### DIFF
--- a/tests/it/ruling_citation_currency.rs
+++ b/tests/it/ruling_citation_currency.rs
@@ -877,34 +877,63 @@ fn body_asserts(body: &str) -> bool {
 fn guard_sites() -> BTreeMap<String, Vec<GuardSite>> {
     let mut sites: BTreeMap<String, Vec<GuardSite>> = BTreeMap::new();
     for (path, text) in source_files() {
-        let lines: Vec<&str> = text.lines().collect();
-        let mut ignored = false;
-        for (index, line) in lines.iter().enumerate() {
-            let trimmed = line.trim_start();
-            if trimmed.starts_with("#[ignore") {
-                ignored = true;
-                continue;
-            }
-            let Some(name) = function_name_declared_on(trimmed) else {
-                // Attributes and doc comments sit between `#[ignore]` and the
-                // `fn`, so only a line that is neither resets the flag.
-                if !trimmed.is_empty() && !trimmed.starts_with("#[") && !trimmed.starts_with("//") {
-                    ignored = false;
-                }
-                continue;
-            };
-            if looks_like_a_guard_name(&name) {
-                sites.entry(name).or_default().push(GuardSite {
-                    path: path.clone(),
-                    line: index + 1,
-                    ignored,
-                    asserts: body_asserts(&body_starting_at(&lines, index)),
-                });
-            }
-            ignored = false;
+        for (name, site) in guard_sites_in(&path, &text) {
+            sites.entry(name).or_default().push(site);
         }
     }
     sites
+}
+
+/// The guard sites declared in one file's `text`, as `(name, site)`.
+///
+/// Split out of [`guard_sites`] so the attribute reading can be exercised
+/// against a **literal fixture** — the move
+/// [`the_assertion_detector_reads_helper_calls_as_assertions`] already makes for
+/// [`body_asserts`], and the one `ruling_guard_field.rs`'s
+/// `the_scanner_reads_test_and_ignore_attributes` makes for its own scanner.
+///
+/// Before #1907 the `#[ignore]` arm was exercised instead by asserting that the
+/// **live tree** held at least one `#[ignore]`d guard. That conflated two
+/// propositions: that this scanner can read the attribute, which is a property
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1949, opened 2026-08-14
docs(normalize): record the minimal-alignment cost model gap, and correct an alignment count

nh13:
Representation-Change: none. Prose inside one ruling record, plus the generated document re-blessed from it. No `src/` file is touched and no output moves.

Follow-up to #1944, which built the enumerator that made both of these measurable rather than arguable.

## 1. The record turns on "minimal" and never says minimal under what

`unchanged-is-read-over-every-minimal-alignment` rules that a reference base is unchanged iff **every minimal alignment** matches it. "Minimal" presupposes a cost model. The record names none.

Two are in play — substitution as one unit-cost operation, and substitution as a delete plus an insert — and **they choose different unchanged sets on a block this repository already pins**. On `ATGC -> GCTG` (the `1420-v4` block, `TEMPLATE` 21-24):

| model | cost | minimal alignments | unchanged |
|---|---|---|---|
| unit-cost substitution | 3 | 2 | **`{22, 23}`** |
| substitution as del+ins | 4 | 6 | **`{23}`** |

So the record decides a different set of bases depending on a parameter it does not fix. Both of its own worked examples (`CAG -> AGA`, `GACA -> AGAT`) are silent on the difference, which is why it went unnoticed.

**Two pieces of evidence point at unit-cost substitution, and this PR records them without ruling.** The record's own framing calls the position-wise reading of `CAG -> AGA` "cost 3", which is only true if a substitution costs 1. And `issue_1539_split_member_separation::forced_unchanged_columns` — the closed-form detector the record names as its own origin — hardcodes unit-cost substitution.

That is strong, but it is inference from an implementation and from a phrasing, not a ruling, and naming the model *decides* `1420-v4`. So this PR states what is owed rather than paying it: an operator ruling naming the model, after which `1420-v4`'s unchanged set follows rather than being chosen. Until then, do not cite the record for a case where the models differ without saying which you used.

Both models are now enumerated and pinned in `tests/it/common/minimal_alignment.rs` (#1944), so this is measurable.

## 2. The record's own arithmetic is wrong

It reads:

> On `GACA -> AGAT` **the two cost-3 alignments** both match ref offset 1, but to different alt offsets

There are **three**, with matched reference sets `{1,3}`, `{0,1}` and `{0,1}`, intersecting to `{1}`. Verified twice by independent implementations — #1944's enumerator and a separate brute force written without reference to it.

The conclusion is untouched: all three match offset 1, so the base is unchanged either way and the cell-based/column-based distinction the passage draws still holds. But this is the worked example the record rests on, and it is the record everything this week has turned on.

## Not fixed here, and worth knowing

An audit run alongside this found that `derivation-may-not-be-bounded-by-the-inputs-spelling` — also `decided`, and rendered into the same published document — argues that four of its moved rows are safe because "the four are **EQUAL-LENGTH** blocks, where the **column correspondence is unique** and so the changed-column set is a fact rather than a choice." That premise is what `unchanged-is-read-over-every-minimal-alignment` falsifies.

**Two decided records, one published document, contradictory premises.** The conclusion about those four rows may well survive — it would if their minimal alignment happens to be the position-wise one — but nobody has checked, and checking it needs a corpus run. Left alone deliberately: editing a decided record's stated ground is an adjudication, not a correction.

## Verification

- Ledger round-tripped programmatically; **31 records in, 31 out**, ids identical, statuses identical, exactly one record differing.
- `generate_spec_fixture` regenerated first (`GEN_EXIT=0`) — a stale gitignored fixture reads as a *dropped ruling record*, which is the most alarming failure available here.
- `docs/NORMALIZATION_CONTRACT.md` blessed **after** the prose edit, then confirmed passing **without** `BLESS_CONTRACT_DOC`.
- `normalization_contract_doc`, `ruling_records_are_intact`, `clause_ruling_index`, `claude_md_adjudication_tables`, `ruling_citation_currency` — **30 passed, 0 failed**, `TESTS_EXIT=0` read back from the log.

No status changed, so neither `CLAUDE.md` adjudication table moves.


nh13:
@coderabbitai pause

PR #1951, opened 2026-08-14
test: withdraw a stale flip instruction and correct a backwards record citation

nh13:
Representation-Change: none. Comment-only. No assertion changes behaviour, no `src/` file is touched, and no output moves.

Two corrections found by a read-only audit of adjudication prose across the codebase. Both are cases where a comment states something a decided ruling contradicts; neither adjudicates anything.

## 1. A stale instruction telling the next maintainer to reverse a decision

`delins_equal_vs_unequal_length_hermetic.rs` carried five sites saying the unequal arm was a ruling violation to be flipped — a module-doc heading ("ferro **VIOLATES** a decided ruling"), an "Assert-then-flip" section, two constant docs, and an assertion message reading *"the defect is FIXED — flip the expectation and declare the representation change"*.

**Both arms are `NC_TEST.1:g.`, a genomic reference.** `delins-payload-coincidence-carve-out-is-coding-dna-scoped` (decided) scopes `DNA/delins.md:47` — the clause that recommends the spanning form over a payload-coincidence split — to the **coding DNA axis**. So `:47` does not reach these rows, `general.md:34` governs unqualified, and the split is **conformant**. What was wrong was the note, not the output.

**This reasoning is not new here.** The sibling file `delins_equal_vs_unequal_length_discriminator.rs` withdrew the identical instruction for its own `g.` row under #1835, and says so at `:114-131`: *"That instruction is wrong and must not be followed. It was written before the axis scope was decided."* The withdrawal was applied to one of the two siblings. This is the other one, and this PR ports the same paragraph.

The assertion message is inverted rather than deleted: reaching the spanning form on this axis would now be a **regression** (something applied the coding-axis carve-out off its axis), not a fix. `UNEQUAL_RULING_CONFORMANT` is renamed to `UNEQUAL_SPANNING_FORM`, deliberately — a reader grepping the old name from the ruling record lands on the correction rather than on an instruction.

## 2. A decided record cited as authority for the inverse of its ruling

`spec_worked_example_rules.rs` said:

> `rulings[codon-carve-out-shape-restriction]` **limits** that exception to the sub/unchanged/sub shape

That record decided the opposite:

> **WIDEN.** The `delins.md:18` / `general.md:35` exception applies wherever its stated precondition holds … **regardless of edit type**. Ferro's restriction to a sub/unchanged/sub shape is dropped.

The sub/unchanged/sub restriction is real — it is what `apply_coding_codon_exception` implements — but it is ferro's **unimplemented-widening state**, not the ruling. The note attributed an implementation fact to a record that says the restriction has no authority.

**This is load-bearing, which is why it is worth fixing rather than tidying.** That sentence is the justification for W58, which the same file calls *"the only registered divergence from a published worked example, and it must stay a set of one unless another is ruled."* A divergence from a published spec worked example was being held open on a record that says the restriction is dropped.

**What this PR does not do:** decide whether W58 survives the widened reading. It restates the authority correctly and records that the question is open — "together affecting one amino acid" is a question about the resulting sequence, and nobody has asked it of this pair. Re-deriving that is for whoever implements the widening.

## Why the existing guards did not catch either

`ruling_citation_currency` scans for status-word contradictions on lines that cite a record id. Site 2 cites the record, spelled correctly, and misstates its content — and "limits" is not a status word. That file's own docs (`:110-135`) name "prose that puts words in a record's mouth" as deliberately unchecked, on the ground that attribution is not mechanically decidable and the population was zero when measured. Site 2 makes it non-zero.

Site 1 cites the record it contradicts only obliquely and states its premise with no clause citation at all, so no id- or clause-keyed scan reaches it.

## Verification

- `cargo fmt --all --check` → `FMT_EXIT=0`
- `delins_equal_vs_unequal_length*`, `spec_worked_example*`, `ruling_citation_currency` → **41 passed, 0 failed**, `TESTS_EXIT=0` read back from the log, no `Cancelling due to signal`
- Grepped for residual references to the withdrawn framing and the renamed constant; the only remaining mentions are the ones that record the correction.

### DIFF
--- a/tests/it/delins_equal_vs_unequal_length_hermetic.rs
+++ b/tests/it/delins_equal_vs_unequal_length_hermetic.rs
@@ -73,13 +73,17 @@
 //! reach this arm either: at equal length the interior column is retained by
 //! denotation, not by the payload/reference coincidence that ruling is scoped to.
 //!
-//! **Unequal (3 -> 2): ferro VIOLATES a decided ruling.** There is no
+//! **Unequal (3 -> 2): ferro's split is CONFORMANT, and the clause that would
+//! have condemned it does not reach a `g.` description.** There is no
 //! column-for-column reading, so nothing in the input forces the `T` at 265 to
 //! be retained. It survives into ferro's output *only* because the payload
 //! happens to open with a `T` that an aligner can pair it with. That coincidence
 //! is the construction `delins.md:46` builds — "parts of the inserted sequence
 //! *align* with the reference sequence, giving an alternative description" — and
 //! `delins.md:47` answers it: **"The `delins` format is recommended"**.
+//!
+//! But `:47` is scoped off this axis. See the withdrawal section below before
+//! reading the record quotations that follow as an indictment of this arm.
 //!
 //! The governing record is the ruling `delins-merge-vs-individual-gap-two-or-more`
 //! in `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json`, status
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

Issue #1939, opened 2026-08-14
Measured: the equal-length minimal-alignment class is the majority of blocks past L=9, and the shipped arm already handles all 840 sampled cases


nh13:
`rulings[unchanged-is-read-over-every-minimal-alignment]` (#1899) ends with **"Scope is UNMEASURED"**, naming the affected class as *every equal-length block whose minimal alignment is cheaper than position-wise*, and noting that a generator must key on **edit distance against block length** rather than length alone. This measures it. Nothing here changes code; it is recorded so the ledger's "unmeasured" can eventually be replaced by a figure with a method attached.

## 1. The class is large, and grows with block length

Levenshtein vs Hamming over equal-length `(reference, alternate)` pairs on `ACGT`. Exhaustive to L=6, sampled at 200,000 pairs beyond:

| L | pairs | edit < hamming | rate |
|---:|---:|---:|---:|
| 2 | 256 | 0 | 0.00% |
| 3 | 4,096 | 204 | 4.98% |
| 4 | 65,536 | 7,848 | 11.98% |
| 5 | 1,048,576 | 210,228 | 20.05% |
| 6 | 16,777,216 | 4,754,472 | 28.34% |
| 8 | 200,000 (sampled) | 87,953 | 43.98% |
| 10 | 200,000 (sampled) | 114,430 | 57.22% |
| 12 | 200,000 (sampled) | 135,206 | 67.60% |

At L=2 it is impossible; beyond L≈9 it is the **majority** of random equal-length pairs. So the ruling governs a common shape, not a corner.

## 2. On the SHIPPED default, ferro already gives the minimal-alignment answer

840 cases (120 per length at L ∈ {3,4,5,6,8,10,12}), each drawn *conditioned on* `edit < hamming`, each on its own 256-base-padded synthetic contig, normalized as `g.<s>_<e>delins<alt>` through the Python bindings.

**840 of 840 produced indel-bearing members. None produced a lone spanning `delins`.** Worked examples:

```
TAC->ATA    edit=2 ham=3   g.[256_257insA;259del]
GTGA->TGCA  edit=2 ham=3   g.[257del;259_260insC]
GAGACT->AGAAGA edit=4 ham=5 g.[257del;261_262delinsAGA]
```

Each places a gap — which is the thing the falsified premise ("an equal-length block has no gap to place") said could not happen.

## 3. The disagreement is entirely in the `Live` arm

[...]

Issue #1928, opened 2026-08-14
MAX_CANONICAL_WINDOW is a picked length while its sibling audit bound is a derived cell budget — state the budget and derive the window


nh13:
`src/normalize/merge.rs` now carries two bounds on the same quadratic work, stated in different units, and only one of them says what it is buying.

```rust
const MAX_CANONICAL_WINDOW: i64 = 4096;              // :1894 — a LENGTH, picked
const MAX_SEPARATION_AUDIT_CELLS: usize = 1 << 20;   // :5272 — a CELL BUDGET
const MAX_SEPARATION_AUDIT_WIDTH: usize = 1024;      // :5299 — and 1024^2 == 1 << 20
```

The audit pair is already derived: its width is the integer square root of its cell budget, so the length is a *consequence* of the budget rather than a second independent number. `MAX_CANONICAL_WINDOW` is a bare literal, and since #1899 it also defines `MAX_SPLIT_BLOCK` (`:2000`), so it governs strictly more than when it was chosen.

## The asymmetry

`MAX_SEPARATION_AUDIT_WIDTH`'s own doc comment already contains the argument, and applies it only to itself:

> `MAX_SEPARATION_AUDIT_CELLS` beside it is the bound that should grow first, **being stated in the work the audit actually does rather than in a length**.

Take that seriously and it applies to the canonical window too. In cells the two bounds are not close:

| bound | permitted cells |
|---|---|
| separation audit | `1 << 20` = 1,048,576 |
| canonical window | `4096^2` = 16,777,216 |

**16x**, between two bounds on `Θ(ref · alt)` work in one module, with no stated reason for the ratio. That is not an argument that 4096 is wrong — it may well be right — it is that nothing records what it is worth, so nobody can tell whether raising it is safe. #1899 raised it by 4x on a wall-clock measurement and never touched memory at all.

## The allocations it bounds

`src/normalize/seqfirst/align.rs`, all sized on the `(ref, alt)` grid:

| site | element | bytes/cell |
|---|---|---|
| `:361` `vec![0u32; (n + 1) * width]` | `u32` | 4 |
| `:309` `vec![0u32; (ref_len + 1) * (alt_len + 1)]` | `u32` | 4 |
| `:111` `vec![false; (n + 1) * width]` | `bool` | 1 |
| `:118` `vec![0u8; (n + 1) * width]` | `u8` | 1 |

[...]

PR #1952, opened 2026-08-14
perf(normalize): narrow the alignment cost grid to u16, and make divergence a benchmark axis

nh13:
Two commits against #1928, in the order they should be read. The second is the change; the first is what made it possible to judge one.

## 1. `92cd3f18` — divergence and indel content become benchmark axes

`benches/seqfirst_align.rs` varied only block length and held the alternate at "~5% of positions perturbed". That is one point on a curve whose shape is the entire question here.

Every cell on a minimal alignment satisfies `|i - j| <= k` for edit distance `k`, so the *reachable* work is `Θ(n · k)`. Holding `k / n` fixed makes the cost look inherently quadratic when it is quadratic only in the regime being sampled. The three regimes are far apart, and the old fixture sat between them:

| shape | `k` relative to `n` |
|---|---|
| a real variant | single digits, independent of `n` |
| the old fixture | `n / 20` |
| a uniform-random pair | ~0.6 `n` (measured for #107, issue #1939) |

So `k` is a parameter now, and `by_length_fixed_k` holds it **constant as `n` grows** — the shape a real corpus presents, and the one under which a banded implementation should approach linear while the current grid stays quadratic. `by_divergence` sweeps the other axis, up to the regime where a band is the whole grid and banding cannot help.

**Indels are a second axis.** The old perturbation was substitution-only, so no optimal path ever left the main diagonal and **no `OUT_DEL`/`OUT_INS` edge was exercised at any length** — the two generators cost the same to compute but produce different DAGs, and only the indel one has more than one minimal alignment to enumerate.

Perturbation sites are also drawn pseudo-randomly rather than by a `step_by` stride, which put every difference on a lattice.

## 2. `173d58a7` — the alignment cost grid narrows to `u16`

`edit_grid` returns `Vec<u32>`, and `build` holds **two of them live at once** — the prefix grid and the suffix grid of the reversed blocks — so cost cells alone are 8 bytes per grid cell.

Costs are bounded by `n + m`, and `on_optimal_path`'s test sums a prefix and a suffix cost whose total is also bounded by `n + m` (the `here + 1` sites are only reached where `i < n` or `j < m`). So the exact bound is `n + m <= u16::MAX`, not `2 * (n + m)`; at the canonical window `n + m` is 7,680, comfortably inside `u16`. `build` now REFUSES past that bound and returns the `None` its callers handle, rather than relying on a `debug_assert` that the release profile compiles out - `build` is `pub` and `MAX_CANONICAL_WINDOW` caps only the reference side, so a direct caller is not capped by it. See #1970.

Measured, peak RSS through the Python bindings, three reps per arm, at the widest accepted block (`W = 3840`):

| arm | reps | mean | bytes/cell |
|---|---|---:|---:|
| `origin/main` | 274.89 / 275.63 / 275.71 MB | 275.4 MB | 18.67 |
| `edit_grid -> u16` | 217.50 / 216.02 / 212.58 MB | 215.4 MB | 14.60 |

**−60.0 MB, −21.8%**, against −59.0 MB predicted from two grids × 2 bytes/cell.

### The instrument was validated before the change was measured

This matters because an earlier attempt at a *different* reduction on this issue produced an uninterpretable null, and I nearly read it as a small win. So: a deliberate grid-sized `Vec<u32>` held live to the end of `build` — a known **+59.0 MB** — measured **+60.2 MB**. Peak-RSS deltas therefore resolve a buffer this size, against a noise floor of ~1% established from five reps of one binary.

### Correctness

Carried by `dominators_agree_with_brute_force_on_exhaustive_small_inputs` — the independent brute-force oracle — not by inspection. 90 alignment tests pass and the `Dominators` output is unchanged; only the width of an intermediate cost cell moves.

## What is deliberately NOT here

Two reductions I implemented, measured, and **dropped**:

- `in_degree: Vec<u32>` → `Vec<u8>` (a cell has at most three in-edges), and
- removing the `cells().collect()` in `dominators()`.

Both are correct and both measure **exactly nothing**: +3.0% / +0.06% / −0.33% against a ~1% noise floor. The reason is now understood and is recorded on #1928 — peak RSS is a maximum over time, and `dominators` allocates *after* `build` frees its grids, so it fits under a mark already set. Shrinking a later allocation cannot lower an earlier peak. "No measurable effect" is not a reason to touch a hot path, so they are not in this PR.

Also dropped: a `dominators` benchmark group and the `pub` visibility it needed. Its justification was that `dominators` is the higher of the two memory peaks — which is the claim the measurement refuted, so the API widening went with it.

## Still open, and stated as open

After the narrowing, the buffers that can be named account for 6 bytes/cell against **14.60 measured**. About **8.6 bytes/cell — ~127 MB at `W = 3840` — is unexplained**, and it scales with cell count rather than being fixed overhead. `ru_maxrss` deltas have taken this as far as they can; locating it needs a heap profiler. I have candidates and no evidence, so I have named none.

**Banding** — restricting the DP to `|i - j| <= k`, which is exact rather than approximate because `on_optimal_path` is by definition confined to that band — is the change that attacks the peak instead of its constant, and is in progress separately. The benchmark in commit 1 is the instrument for judging it.

## Verification

- Full gate with `FERRO_MANIFEST` armed and `FERRO_REQUIRE_BULK_FIXTURES=1`, zero `Skipping:` lines.
- Python binding suite — `cargo nextest` does not reach it and this touches `src/normalize/`.
- **The memory figures were taken on a shared developer machine and should be re-confirmed on a quiet host before being quoted anywhere durable.** The direction and rough magnitude are safe (the effect is 20× the measured noise floor); the third significant figure is not.

Representation-Change: none. Measured, not assumed. The u16 cost grid now refuses `n + m > 65,535` and returns the `None` its callers already handle, so an input past the bound is declined by the aligner and answered by the `partition_block` fallback instead of being computed from wrapped costs. Compared byte-for-byte against this PR's own pre-narrowing parent (the benchmark commit, where the grids are still u32) on literal `ins` payloads of 100 / 40,000 / 65,000 / 70,000 bases, including the 70,000-base case #1970 reported: every arm emits an identical description, so the refusal costs this class nothing. Nothing that ferro answers today becomes refused - the refusal is inside the partitioner, which has a decline path, not at the normalizer's boundary. `member_alignment` gains the same decline, which can only widen an audited member, never corrupt one. The benchmark change touches no src/ file.

### DIFF
--- a/benches/seqfirst_align.rs
+++ b/benches/seqfirst_align.rs
@@ -103,30 +103,64 @@ fn diverge_subs(reference: &[u8], k: usize, seed: u64) -> Vec<u8> {
 /// generators so their timings are comparable, and it is the shape the
 /// `unchanged-is-read-over-every-minimal-alignment` ruling governs.
 fn diverge_indels(reference: &[u8], k: usize, seed: u64) -> Vec<u8> {
-    if reference.is_empty() || k < 2 {
+    // `pairs` is clamped so `stride >= 2`, which is what keeps the deletion site
+    // (`offset == 0`) and the insertion site (`offset == stride / 2`) distinct.
+    let pairs = (k / 2).min(reference.len().saturating_sub(2) / 2);
+    if pairs == 0 {
         return reference.to_vec();
     }
     let mut state = seed | 1;
-    let pairs = k / 2;
-    // Delete at `pairs` sites and insert at `pairs` others, walking left to
-    // right so the result stays the same length.
-    let mut out = Vec::with_capacity(reference.len());
     let stride = reference.len() / (pairs + 1);
+    let insert_offset = stride / 2;
+
+    // Delete at `pairs` sites and insert at `pairs` others, half a stride apart,
+    // walking left to right so the result stays the same length.
+    let mut out = Vec::with_capacity(reference.len());
+    let (mut deleted, mut inserted) = (0usize, 0usize);
     for (i, &b) in reference.iter().enumerate() {
-        let slot = i / stride.max(1);
-        if slot < pairs && i % stride.max(1) == 0 {
-            // Insert a base that differs from the one it precedes, then skip a
-            // base — one insertion and one deletion, `stride` apart.
-            out.push(b"ACGT"[(next(&mut state) as usize) % 4]);
+        let (slot, offset) = (i / stride, i % stride);
+        if slot < pairs && offset == 0 {
+            // Deletion: consume the reference base and emit nothing.
+            deleted += 1;
             continue;
+        }
+        if slot < pairs && offset == insert_offset {
+            // Insertion: emit an extra base before this one.
+            out.push(b"ACGT"[(next(&mut state) as usize) % 4]);
+            inserted += 1;
         }
         out.push(b);
     }
-    out.truncate(reference.len());
-    while out.len() < reference.len() {
-        out.push(b'A');
-    }
+
+    // Asserted rather than trusted, and asserted with `assert!` rather than
+    // `debug_assert!`: `[profile.bench]` inherits `release`, so a `debug_assert`
+    // here would never run in the profile the benchmark is measured in. This
+    // runs once per generator call, outside `b.iter()`.
+    //
+    // The predecessor of this function pushed a random base *instead of* the
+    // reference base and `continue`d, so it emitted exactly one byte per input
+    // byte — a substitution, not an indel pair — and `edit == hamming` at every
+    // `k`. Nothing said so; the vestigial `truncate`/pad tail that could never
+    // fire was the only trace. See #1970.
+    assert_eq!(
+        deleted, inserted,
+        "each deletion must be paired with an insertion"
+    );
+    assert_eq!(
+        out.len(),
+        reference.len(),
+        "the pairs must net to zero length change"
+    );
     out
+}
+
+/// Positions at which two equal-length blocks differ.
+///
+/// Only meaningful beside an edit distance: `edit < hamming` is what says an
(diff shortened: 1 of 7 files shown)


nh13:
@coderabbitai pause

nh13:
Holding this one from the merge queue pending #1970.

Measured by execution rather than by reading: `g.1000_1001ins<70,000 bases>` normalizes on `origin/main` (output length 70,024, 0.023 s) and on this head trips `2*(n+m) = 140000 exceeds u16`, with `n+m = 70,000`. The bound is a `debug_assert` and `[profile.release]` sets neither `debug-assertions` nor `overflow-checks`, so in the shipped library it wraps silently instead of panicking.

Every gate in front of `AlignmentDag::build` bounds **cells**, not `n + m` — a `1 x 70,000` grid clears `MAX_SEQFIRST_GRID_CELLS` by four orders of magnitude. The brute-force oracle here is exhaustive over `n + m <= 10`, so it cannot reach this class either.

That makes `Representation-Change: none` inaccurate on this class rather than merely unproved; the check passed because it rejects silence, not inaccuracy. Turning the bound into a real refusal (`None`, which both callers already handle) fixes the defect and makes the declaration true.

Separately, `diverge_indels` generates substitutions rather than indels — it pushes a random base *instead of* the reference base and `continue`s, so output length equals input length and `edit == hamming` at every `k` (re-implemented and DP-checked: k=8 -> 2, k=32 -> 12, k=128 -> 47). The memory table is unaffected and its arithmetic checks out.

Everything else in the PR verified clean.

PR #1953, opened 2026-08-14
feat(rulings): give the ledger a home for a clause-free house choice

nh13:
## Why

The `rulings` ledger can record two things: a settled reading of a clause (`decided`, with a `governing` clause) and an unsettled conflict (`undecided`, two clauses, no verdict). It cannot record the third and commonest state — *the recommendations are silent here and we chose X* — because a `decided` record is required to name a clause it holds to govern. Measured on `origin/main`: 31 records, 28 `decided` / 3 `undecided`, every `decided` one citing a governing clause and all three clause-free ones `undecided`.

`README.md` rule 5's silent limb says in terms that ferro "decides under rule 6 and violates nothing", so the ledger cannot express a state its own ruleset names.

That state went somewhere. The rule that a reference base is unchanged iff **every** minimum-edit-distance alignment matches it had no record, so it was written as prose where it was needed and ended up as six restatements across `src/` and `tests/`. Three of them drifted into the flatly false generalisation that an equal-length block's column correspondence is unique. Reconciling them cost several days. An HGVS committee-perspective review of the same rule called it "an implementer's choice the recommendations neither require nor forbid" and objected to its being presented as compliance.

## What this adds

**A `house_choice` object on a ruling record.** It carries `under` — a closed set of two, `rule-five-silent` or `rule-six`, never a spec clause — and `considered_and_rejected`, which must not be blank. A record carrying one is `decided` and names no governing clause.

**The overclaim is unrepresentable, not discouraged.** `generate_spec_fixture` and `tests/it/common/rulings.rs` both refuse a `house_choice` that names a `governing` or a `deviates_from` clause, and refuse one on a non-`decided` record. This mirrors the existing `undecided` refusal deliberately: that arm stops a record claiming more *certainty* than the project has, this one stops it claiming more *authority* than the spec gave it. The second is the costlier direction here, because a governing clause is quotable — once one is written down the choice starts being cited as conformance.

Sabotaged and verified rather than asserted. Adding a `house_choice` to a record that already names a governing clause:

```
error: ruling "duplication-must-ranks-the-label-not-the-partition": a `house_choice` names a
governing clause. A house choice is made under `README.md` rule 5's silent limb or rule 6, which
are what the project does where the recommendations do NOT decide — so it cannot also hold a
clause to govern. …
```

and the generator exits 1. The positive arm was probed the same way: a well-formed clause-free house choice builds, exits 0, and reaches the generated fixture.

**A regrowth ratchet**, `tests/it/adjudication_prose_regrowth.rs`. A one-time cleanup regrows, so comment prose that decides what the correct representation is must name a ledger record id, or its file must carry a row in a shrink-only `REGROWTH_EXEMPT`.

The obvious predicate — the inverse of `clause_ruling_index`, flag a comment citing a governed clause with no record nearby — was measured first and rejected: **707** comment lines in 144 files, most of them the legitimate habit of quoting the clause a piece of code implements. A guard that flags 700 lines on its first run is argued with once and then disabled. What ships is a narrow phrase list, chosen by measuring every candidate over the whole tree, keyed on prose that makes an *ontological* claim about representation ("a fact rather than a choice", "the correspondence is unique"). It flags **three** blocks in the entire tree; one is exempt with a stated reason. Phrases that read adjudicative but are ordinary here — "is canonical" (26 blocks), "the canonical form is" (24), "must be described as" (4, all quoting `DNA/duplication.md:18`) — are rejected on the same measurement, and the module says so.

An exemption row is keyed by file, because that is the only key that survives editing; the cost of that key is capped by failing when one row covers more than one flagged block.

## What it does not reach, stated on the module

Every existing ledger guard fires only on prose that already names a record, or on the ledger's own contents. This one is keyed on a phrase instead, which is why it can see the class that named nothing — but it is a substring scan over comments, it cannot see prose inside an `assert!` message, and it does not touch the 707-line clause-citing class at all. Those limits are in the module docs rather than in this description, so they cannot rot separately from the code.

**No ledger record is added, changed or re-statused: 31 records in, 31 out, same ids, same statuses.** The one candidate house choice — the minimal-alignment rule — already has a `decided` record naming a governing clause, added hours before this branch. Adding a second record for the same rule would be the drift failure this PR exists to close, and deciding whether that record overclaims is the disposition of #1878 / #1904, which is not a contributor's call.

**No assertion changes behaviour.** The three sites are repointed at the record instead of restating it, and `reported_confluence_pairs.rs` stops *instructing* a future maintainer to re-pin the verdicts and delete an assertion when a form it names is reached — an outcome nobody has agreed to. The assertions themselves are untouched.

Left alone deliberately: `src/normalize/merge.rs` (contested by two open PRs), `README.md:232-234` (in flight), and `issue_1284_transcript_axis_collision.rs`, which is the model to imitate and is exempt only because it quotes the withdrawn claim verbatim in order to withdraw it.

## Gate

`cargo fmt --all --check`, `cargo clippy --all-features --all-targets -- -D warnings` and `cargo nextest run --features dev --lib --test it --no-fail-fast` all clean, and `docs/NORMALIZATION_CONTRACT.md` is unchanged because no record moved.

Representation-Change: none. Ledger schema, one new test module and prose; no file under `src/` is touched, and no ledger record is added or changed.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -936,6 +936,15 @@ adjudications and do not need records.
 - **undecided** — a first-class state. The generator refuses an `undecided` record that names a
   governing clause, so an open question cannot smuggle in a ruling nobody made. Prefer an honest
   `undecided` record to no record.
+- **house-choice** — `decided`, and **ours rather than the spec's**: a `rulings` record carrying a
+  `house_choice` object, made under `README.md` rule 5's silent limb or rule 6, naming **no**
+  governing and no deviated-from clause. The generator and the ledger reader both refuse one that
+  names either, mirroring the `undecided` refusal — that arm stops a record claiming more certainty
+  than the project has, this one stops it claiming more authority than the spec gave it. Use it
+  whenever the honest statement is "the recommendations do not reach this and we picked X"; the
+  alternative is what happened to the minimal-alignment rule, which had no home here and grew six
+  drifting prose restatements instead. It must say what was considered and rejected, and the record
+  is never citable as conformance.
 
 **A test that merely pins today's output is not an adjudication record.** It is a change
 detector, and this repo has already been bitten by the difference —
--- a/examples/generate_spec_fixture.rs
+++ b/examples/generate_spec_fixture.rs
@@ -187,6 +187,88 @@ mod overrides {
         Undecided,
     }
 
+    /// The `README.md` rule a clause-free house choice is made under.
+    ///
+    /// Only two rules can carry one, and naming which is the whole point of the
+    /// field: rule 5's *silent* limb says in terms that ferro "decides under
+    /// rule 6 and violates nothing", and rule 6 is where the maintainers choose
+    /// among conformant forms. Neither is a claim about what the recommendations
+    /// require. A free-text field here would let a record write "the spec" into
+    /// the slot reserved for "our own judgement", which is the exact
+    /// substitution [`HouseChoice`] exists to make unrepresentable.
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, Deserialize, Serialize)]
+    #[serde(rename_all = "kebab-case")]
+    pub enum HouseRule {
+        /// `README.md` rule 5, **silent** limb — the recommendations are merely
+        /// incomplete here, so ferro decides under rule 6 and violates nothing.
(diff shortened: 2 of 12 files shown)


nh13:
@coderabbitai pause

PR #1956, opened 2026-08-14
docs(rulings): correct the four equal-length rows' ground in the weight-bound record

nh13:
Item 5 of the adjudication audit: do the four rows that `rulings[derivation-may-not-be-bounded-by-the-inputs-spelling]` justifies as SPEC-CONFORMANT survive the minimal-alignment reading that `rulings[unchanged-is-read-over-every-minimal-alignment]` (#1899) established?

**They do not. All four fail, and so does one of the "fifteen".**

## The contradiction this resolves

Two `decided` records sit in one published document on contradictory premises, and nothing checks it:

- `derivation-may-not-be-bounded-by-the-inputs-spelling` (rendered at `docs/NORMALIZATION_CONTRACT.md:1091`) justifies four moved rows on the ground that *"the four are EQUAL-LENGTH blocks, where the column correspondence is unique and so the changed-column set is a fact rather than a choice"*.
- `unchanged-is-read-over-every-minimal-alignment` (rendered at `:1570`) falsifies exactly that. Equal length does not make the column correspondence unique.

This PR does not adjudicate between them. It measures the four rows, records the result in the record, and withdraws the sentences the measurement refutes.

## The four, enumerated for the first time

The record only ever *counted* them — "Twenty rows change form: 4 SPEC-CONFORMANT, 15 SPEC-SILENT, 1 correctness defect". The enumeration survived only on the **closed** #1616 branch (`01f4f94b`) and in an untracked working note. It is now in the record itself.

| test | block | position-wise | minimal | changed columns, position-wise → minimal |
|---|---|---|---|---|
| `cis_junction_crossing_shift::the_five_prime_junction_barrier_does_not_over_clamp` | `TTTTTTTAAT` → `ATTTTTTTAA` (10/10) | cost 3 | **cost 2**, unique | `{0,7,9}` → `{9}` |
| `issue_1284_transcript_axis_collision::a_noncoding_del_dup_collision_is_repaired` | `CAG` → `AGA` (3/3) | cost 3 | **cost 2**, unique | `{0,1,2}` → `{0}` |
| `normalize_reparse_invariant::a_del_beside_a_dup_re_spells_instead_of_colliding` | `CAG` → `AGA` (3/3) | cost 3 | **cost 2**, unique | `{0,1,2}` → `{0}` |
| `cis_confluence_adjudication::two_adjacent_members_that_both_consume_reference_are_one_delins` | `AATA` → `TAAT` (4/4) | cost 3 | **cost 2**, unique | `{0,2,3}` → `{3}` |

In every one the minimal alignment is **unique** and is one inserted base plus one deleted base with no substitution at all. All four are one shape: one base inserted at one end of the block and a **different** base deleted at the other, everything between matching. So the record's word "unique" is accidentally true — of the minimal alignment. It is the *position-wise* pairing the sentence means, and that is not minimal in any of the four.

They are **not rotations**, which would re-insert the base they removed; an earlier revision of this description called them that, and it is true only of the `GTAAAA -> TAAAAG` row noted at the end. The arithmetic is unaffected.

**One row is not the clean contrast the other three are.** `AATA -> TAAT` changes columns `{0, 2, 3}` — two groups separated by one matching base, not one run — so the position-wise reading does not produce the spanning `c.10_13delinsTAAT` either. It produces `c.[10A>T;12_13delinsAT]`, a third description on neither side of the dispute, reaching the spanning form only by invoking the codon exception across the codon-4/codon-5 boundary. The record now says so.

Two of the blocks are checkable against committed constants rather than taken from the note: `TRACT` (`cis_junction_crossing_shift.rs:99`) gives `g.3_12` = `TTTTTTTAAT`, and `CORE` (`cis_confluence_adjudication.rs:103`) gives `c.10_13` = `AATA`.

## The direction is the whole point

Fed the minimal reading, `delins.md:16` has **no antecedent** — there are no two *consecutive changed* nucleotides in any of the four — and `:17` describes the one insertion and the one deletion individually. That is the **pre-removal** spelling, in every case.

So the record's sentence

> The post-removal output is what those clauses produce VERBATIM and the pre-removal output is a spelling they never generate

is backwards, and this PR withdraws it.

## The tree already agrees; only the record lagged

On `origin/main` at `cb79979a`, **all four ship the pre-removal form**:

- #1899 reversed both `CAG → AGA` rows explicitly, leaving the arithmetic in `tests/it/normalize_reparse_invariant.rs:90-96` (*"It is withdrawn: the cheaper alignment leaves those bases untouched, so they are not artifacts"*) and `tests/it/issue_1284_transcript_axis_collision.rs:74-88`;
- #1835's default flip took the fourth to `c.[9dup;13del]`, and `tests/it/cis_confluence_adjudication.rs` now states the same cost-2-against-cost-3 reading verbatim;
- the first was never re-pinned at all and still asserts `TEMPLATE:g.[2_3insA;12del]` → itself.

No output the paragraph blessed is shipping, and three of the four carry the refutation in their own comments while the record still carries the claim.

## What this does NOT change

**The ruling.** That the four now ship their pre-removal spelling does not mean the deleted bound is back — and holding that distinction is the point. The bound's comparand was the *input's own member list* and it worked by declining a derivation; these four are derived from the resulting sequence and reach the same string only because that string is what the minimal alignment of the block actually is. The ruling rests on its first, second and third reasons and is untouched.

**The conformance decomposition.** "4 SPEC-CONFORMANT, 15 SPEC-SILENT … and ZERO spec violations" is deliberately **not** re-classified here. Re-classifying it is an adjudication; this is a measurement. It is flagged in the record as open.

**And the replacement reading's own grounding.** This PR establishes that the position-wise ground is false. It establishes nothing about whether reading "unchanged" over every minimal alignment is itself grounded in the recommendations — that is `unchanged-is-read-over-every-minimal-alignment`'s question, and the record now says so explicitly.

That question is live rather than rhetorical. A spec-only review of the pinned submodule — restricted to `assets/hgvs-nomenclature`, with the ledger and this repo's own prose deliberately out of scope — reports **zero** occurrences of `cost`, `parsimon*`, `edit distance`, `optimal` or `fewest` across `recommendations/`, `background/` and `consultation/`; the six `minimal`/`minimum` hits concern the extent of unsequenced deletions and SNP-array probe bounds. Of eight `align*` hits, the only one using an alignment to *derive* a description is `DNA/delins.md:46` — in the specification's own scare quotes, describing the form it declines in favour of the spanning `delins`. Recorded here as context for whoever takes the adjudication; **no part of it is asserted in the ledger by this PR.**

## Also corrected

"The fifteen are unequal-length blocks on which `:15`/`:16`/`:17` have no defined input at all" is false as stated. The source enumeration labels `project::projector::tests::project_cis_net_in_frame_but_mid_shift_stop_is_frameshift` as `GTAAAA` → `TAAAAG`, **equal 6/6**, and it fails the same way — minimal cost 2 against position-wise 3, unique alignment, changed column `{0}`. Whether that moves its disposition is unmeasured.

## Method, and why the numbers should be believed

The enumeration was done with an implementation written from the definitions rather than ported, so that agreement is evidence rather than a shared bug. It was validated two independent ways:

- it reproduces all five cases pinned by #1944's Rust enumerator, under **both** cost models (10/10);
- it reproduces #1939's separately-measured Levenshtein-vs-Hamming counts **exactly** — 204 at L=3 and 7,848 at L=4, both exhaustive.

While measuring, one further figure worth recording: the record's premise fails on the **majority** of equal-length blocks past L=3, and much more often than the "edit < hamming" rate alone suggests. At L=4, position-wise is non-minimal for 11.98% of pairs — but the correspondence is non-unique for 49.5%, and the two failure modes are not nested. A block can have position-wise among its minimal alignments and still have no unique correspondence (smallest witness: `AC → CA`, cost 2 either way, three minimal alignments).

## Scope of the edit

One record's `rationale`, appended to. Verified by round-trip: 31 records in and out, ids identical, statuses identical (28 `decided`, 3 other), exactly one record differing, that record differing only in `rationale`, and every byte outside that one line unchanged.

`docs/NORMALIZATION_CONTRACT.md` is re-blessed as the generated render.

Representation-Change: none


nh13:
@coderabbitai pause

PR #1957, opened 2026-08-14
Prevent variants from rolling off contig

### LINKED ISSUE #1943: Allow normalizing variants at reference position 1
### Summary
Currently, normalizing `g.[1A>C;2T>A;3G>T]` fails when supplied to `sequence_normalize()`. This is because the multi-base substitution is directly adjacent to the start of the reference, and will be converted to `0_1insC;3del`, which is off the end of the reference.

### Resolution
Fix edge case logic to handle such variants by avoiding rolling too far, instead of rejecting such variants.

TimD1:
### Issues
* closes #1943


nh13:
@coderabbitai pause

nh13:
Closing in favour of #1969, which carries this work commit-for-commit with authorship preserved.

PR #1958, opened 2026-08-14
docs(ledger): record the measured scope of the minimal-alignment ruling

### LINKED ISSUE #1939: Measured: the equal-length minimal-alignment class is the majority of blocks past L=9, and the shipped arm already handles all 840 sampled cases

nh13:
Closes #1939.

`rulings[unchanged-is-read-over-every-minimal-alignment]` (#1899) ended with **"Scope is UNMEASURED"**. It has since been measured (#1939), so this replaces the admission with the figures and the method, and — importantly — with the limits, so nobody can quote the numbers wider than they were taken.

No code changes. The ledger prose moves, `CLAUDE.md`'s decided-records row for the same record moves with it, and `docs/NORMALIZATION_CONTRACT.md` is re-blessed because it is a generated render of the ledger.

## What the measurement says

**The class is large.** Over equal-length uniform-random `ACGT` pairs, the rate with edit distance strictly below Hamming distance is 0.00% at L=2 (impossible), 4.98% at L=3, 11.98% at L=4, 20.05% at L=5, 28.34% at L=6 — those five exhaustive — then 43.98% / 57.22% / 67.60% at L=8/10/12 on 200,000 sampled pairs each. Past L≈9 it is the **majority** of pairs, so the record governs a common shape rather than a corner.

**The shipped exposure is nil.** 840 cases drawn conditioned on `edit < hamming` (120 per length at L ∈ {3,4,5,6,8,10,12}, each one `delins` over one block on its own 256-base-padded synthetic contig) produced indel-bearing members **840 of 840** and a lone spanning `delins` **zero** times. The default arm already reads unchanged-ness the way the record requires, so this asks for no behaviour change.

**The residual is entirely `FERRO_PARTITION=live`**, which the record already names as giving the other answer — and its disagreement is **largest at short lengths** (110/120 at L=3, falling monotonically to 14/120 at L=12). That is the opposite of what a length-gated intuition predicts, and it is the part most worth having written down: it is exactly the shape that would mislead someone gating future work here on block size.

## What is deliberately recorded as still unmeasured

Stated in the record rather than left to be discovered: the `g.` axis only, single-member inputs only, uniform-random composition, no block near a window edge. And "indel-bearing" is **not** "verified minimal" — it says no sampled output took either shape nameable as wrong, not that each output equals *the* minimal alignment. Asserting the stronger claim needs an independent aligner, and the record now says so, since that is the check worth writing if this is ever pinned as a guard.

The record also warns against the reading the tables invite: the naive one-substitution-per-differing-column spelling was emitted by **neither** arm (its bucket matched zero of the 840 on both), so the counts are one spanning block versus a partition that places a gap — not position-wise versus minimal.

## Why `CLAUDE.md` changes too

Its decided-records table restates this record's scope, and `claude_md_adjudication_tables` enforces only the status partition, not the prose. So a ledger that says "measured" beside a `CLAUDE.md` that says "UNMEASURED" would pass every guard while being exactly the one-rule-in-two-places drift this repo names as its recurring failure mode.

Representation-Change: none. Ledger prose, its generated render, and a `CLAUDE.md` table row; no `src/` file is touched and no description ferro emits can change.


nh13:
@coderabbitai pause

PR #1960, opened 2026-08-14
perf(ci): move the three slow censuses onto the optimized archive

### LINKED ISSUE #1955: ci: the workflow's critical path is two #1840 censuses left on the debug archive
`Test oracle (1/4)` is the whole workflow's critical path on run [31845625298](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/31845625298) — 328s, finishing 83s after the next-slowest job — and its tail is two exhaustive censuses added by #1840 that are still on the **debug** archive.

## Measured

Run 31845625298 (merge group, `044f085a`), all 33 jobs. Job start/end offsets from run start:

```
     8 ->    134  ( 126s)  Spec fixtures and generated docs   <- gates the fan-out
     3 ->    119  ( 116s)  Build test archive
   156 ->    484  ( 328s)  Test oracle (1/4)                  <- critical path
   170 ->    401  ( 231s)  Test (3/6)
   487 ->    491  (   4s)  Test (aggregator)
```

The two censuses, aggregated across every job they run in:

| total | runs in | test |
|---:|---:|---|
| 396s | `Test (3/6)` 155.7s + `Test oracle (1/4)` 240.3s | `it issue_1542_direction_symmetry::the_member_count_does_not_depend_on_the_shuffle_direction` |
| 245s | `Test (3/6)` 149.6s + `Test oracle (3/4)` 95.8s | `normalize::merge::tests::direction_symmetry::…_on_any_arm` |

641s of test CPU per run, in the two jobs that are already the two slowest.

## Why the shards cannot absorb it

`--partition hash:K/N` splits by a hash of the test **name**, which balances by count and is blind to cost — the same property recorded on the `sweeps` job when all three exhaustive sweeps hashed into one shard. Both new censuses hashed into `Test (3/6)`, and the integration one into `Test oracle (1/4)` alongside `normalize_axis_preserving::axis_is_preserved_over_the_cmrg_corpus`.

A shard's wall clock can never fall below its slowest single test, so re-sharding cannot fix this. Measured concurrency confirms the tail runs alone: `Test oracle (1/4)` executes 697s of >=5s test work in 328s of wall (2.1x), against 2.5x on `Test oracle (2/4)`.

## The crowding is measurable too

`axis_is_preserved_over_the_cmrg_corpus` did not change, and went:

| run | base | cmrg on oracle shard 1 |
|---|---|---:|
| 31839189774 | before #1840 | **79.0s** |
| 31839468759 | #1840's own merge group | **107.5s** |
| 31845625298 | current | **172.8s** |

It is 53.2s in the unarmed `Test (1/6)`. Both censuses and the cmrg corpus drive `rayon`, and nextest already runs ~4 tests concurrently on a 4-vCPU runner, so rayon inside a test adds no throughput — it oversubscribes. The tail is being crowded by its shard-mate.

## What the repository already established

The `sweeps` job (`ci.yml:1537`) is the carve-out for exactly this class, and its value is the **optimized archive**, not the job boundary. Measured there:

```
cis_junction_crossing_shift        71.06s -> 8.55s   8.3x
repeat_span_sibling_overlap        40.44s -> 5.29s   7.6x
issue_1254_sibling_crossing_shift  20.14s -> 2.32s   8.7x
```

At that ratio the 240s census is ~30s. A carve-out onto the debug archive, by contrast, buys ~50s and costs a whole extra runner.

## Two constraints on the fix

- `test-build-soak` runs `cargo nextest archive --cargo-profile soak --test it` — **integration target only**. `merge.rs`'s unit half (245s) is not in that archive.
- `spec_conformance_axis`'s two censuses (140.6s + 128.3s, the next tier down) are in `ORACLE_EXCLUDE` for a documented reason — `FERRO_ASSERT_IDEMPOTENT` panics on the very rows they count — so they cannot simply be added to `sweeps`, which arms all four oracles.

## Scope

Minimize total workflow wall. Also in scope: an audit of what each job recompiles across the fan-out, and whether any of it is shareable.

nh13:
Closes #1955.

**Measured on this branch's own run, [31849010009](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/31849010009) — 33/33 green. Workflow wall 491s -> 314s, a 36% cut, and the bottleneck has moved.** The measured table is at the bottom; the baseline throughout is run [31845625298](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/31845625298).

## The measurement

`Test oracle (1/4)` is the whole workflow's critical path — 328s, finishing 83s after every other job — and its tail is two exhaustive censuses that #1840 added the day before, still on the debug archive.

```
     8 ->    134  ( 126s)  Spec fixtures and generated docs   <- gates the fan-out
     3 ->    119  ( 116s)  Build test archive
   156 ->    484  ( 328s)  Test oracle (1/4)                  <- critical path
   170 ->    401  ( 231s)  Test (3/6)
   487 ->    491  (   4s)  Test (aggregator)
```

Across every job they run in:

| total | where | test |
|---:|---|---|
| 396s | `Test (3/6)` 155.7s + `Test oracle (1/4)` 240.3s | `issue_1542_direction_symmetry::…shuffle_direction` |
| 338s | `Test` shards 2/3/4/5 | `spec_conformance_axis` (5 tests) |
| 432s | `Test` 1/3/5 119.3s + `Test oracle` 1/3 313.0s | `normalize_axis_preserving` (4 tests) |

`--partition hash:K/N` assigns by a hash of the test NAME, so it balances shard counts and is blind to cost. `Test oracle (1/4)` ran 697s of >=5s test work in 328s of wall — 2.1x effective concurrency, against 2.5x on shard 2/4. The tail runs alone with three idle cores, and a shard's wall can never fall below its slowest single test, so re-sharding cannot fix it.

The crowding is measurable on a test that did not change. `axis_is_preserved_over_the_cmrg_corpus` on oracle shard 1:

| run | base | cmrg |
|---|---|---:|
| 31839189774 | before #1840 | **79.0s** |
| 31839468759 | #1840's own merge group | **107.5s** |
| 31845625298 | current | **172.8s** |

It is 53.2s in the unarmed `Test (1/6)`. Both censuses and the cmrg corpus drive `rayon`, and nextest already runs ~4 tests concurrently on a 4-vCPU runner — so rayon inside a test adds no throughput, it oversubscribes.

## What this does

Adds `CENSUS_FILTER` and a `censuses` job on the **optimized** archive, exactly as `sweeps` was built. `test` and `test-oracle` negate the filter; `censuses` selects on it with `--no-tests=fail`.

This is not a new idea — `test-oracle`'s own comment named it and declined it:

> The robust alternative — giving `normalize_axis_preserving` a job of its own and negating it here — was not taken because this job's `-E` is re-derived by BOTH `scripts/run_oracle_suite.sh` and `tests/it/oracle_exclude_invariant.rs`, so a fourth negation term is a three-file change; the shard count is not.

The shard count was the cheaper fix and it did not hold. N=4 was chosen to separate the two bulk corpora; #1840 added a test, the hash re-rolled, and the job went back to 328s by a different draw. A negation is stable under a re-roll and a shard count is not, so the three-file change is now the right price. That comment is rewritten to record it.

**The job boundary is not the win — the archive is.** `sweeps` measured 7.6-8.7x moving work of exactly this shape onto `--cargo-profile soak`. A carve-out onto the debug archive would move the same 240s to a different runner and buy ~50s.

**Two steps, because the modules disagree about the oracles.**

- **Armed** (`issue_1542_direction_symmetry`, `normalize_axis_preserving`) runs what `test-oracle` ran — the same three flags, no fourth. An armed run strictly subsumes the un-armed one, which is the argument `sweeps` already makes, so dropping the plain-`test` copy loses nothing.
- **Un-armed** (`spec_conformance_axis`) reproduces the plain-`test` run exactly. That module is in `ORACLE_EXCLUDE` and stays there: it COUNTS the non-idempotent outputs `FERRO_ASSERT_IDEMPOTENT` panics on, and a panicking row is filed `declined` and never reaches the family's output set — which does not redden the census, it flatters it. It is in both filters deliberately; they answer different questions (*which instrument may be armed* vs *where a module runs*), and `oracle_exclude_invariant` demands the first.

`FERRO_ASSERT_SEQUENCE` is set by neither step, which is what makes this a move and not a change of instrument. `sweeps` does set it, so these modules could not simply be appended there.

## Verified

Selections partition exactly, over the soak archive's own binary (`--test it`, no `dev` feature — the archive `censuses` consumes):

```
base            6012   not test(proptest) and not (SWEEP_FILTER)
minus census    5999   … and not (CENSUS_FILTER)
census filter     13
  armed step       8
  unarmed step     5
5999 + 13 == 6012        nothing falls between the jobs, nothing runs twice
   8 +  5 == 13          both steps together cover the filter
```

All 13 resolve inside the soak archive, so none needed the `dev` feature and none had to be left behind.

`tests/it/census_filter_invariant.rs` (new) closes the two holes neither `--no-tests=fail` nor the partition comment can see: a module in `CENSUS_FILTER` that **no** step selects is negated everywhere and runs nowhere — CI stays green and gets *faster*, which reads as this PR working rather than as coverage deleted — and a module a step selects that the filter does not name runs twice. It also pins the job at two steps, so collapsing them would fail rather than silently arm `spec_conformance_axis`.

**Both steps pass under `--cargo-profile soak`, run exactly as the job runs them** — the armed one with the three `FERRO_ASSERT_*` flags, the un-armed one bare. `spec_conformance_axis`'s seventeen-field pins are green in both directions at `opt-level = 3`, which is the assertion that mattered: none of them moved.

```
armed    1 test  passed   issue_1542_direction_symmetry            7.9s
unarmed  5 tests passed   spec_conformance_axis                   30.3s  (28.9 / 30.3 for the two censuses)
```

**Do not read those as the CI projection** — this is a 12-core dev box against a 4-vCPU runner, and every one of these tests is rayon-parallel, so the ratio here flatters the move. They establish that the tests pass at `opt-level = 3` with the oracles armed, and nothing about how long CI will take. `normalize_axis_preserving`'s three corpus tests could not be run locally at all (the bulk fixtures are release assets and are absent here); CI is the first place they run on this branch.

15/15 green locally across the three CI-reading guards, including `oracle_exclude_invariant::the_local_oracle_runner_selects_exactly_what_ci_selects`, which is what proves the shell script and the Rust re-derivation agree on the new four-term expression. `cargo fmt`, `actionlint` and `zizmor` clean.

## Measured

Per test, baseline -> this run, each still under the flags it ran under before:

| test | before | after | |
|---|---:|---:|---:|
| `issue_1542_direction_symmetry` (armed) | 240.3s | **39.7s** | 6.1x |
| `spec_conformance_axis::three_prime_conformance_census` | 140.6s | **22.4s** | 6.3x |
| `spec_conformance_axis::five_prime_conformance_census` | 128.3s | **23.6s** | 5.4x |
| `normalize_axis_preserving::…cmrg_corpus` (armed) | 172.8s | **39.2s** | 4.4x |
| `normalize_axis_preserving::…clinvar_unique_corpus` (armed) | 99.3s | **33.7s** | 2.9x |

Per job:

| job | before | after |
|---|---:|---:|
| `Test oracle (1/4)` | 328s | **102s** |
| `Test (5/6)` | 189s | **62s** |
| `Test (3/6)` | 231s | **112s** |
| `Test (2/6)` | 203s | **90s** |
| `Test oracle (3/4)` | 160s | **105s** |
| `Slow censuses (optimized)` | — | **90s** (39.7 armed + 23.7 un-armed + ~27 setup) |
| **workflow wall** | **491s** | **314s** |

**The two shards that carried none of the moved tests are the control**: `Test oracle (4/4)` 149s -> 143s and `Test (6/6)` 87s -> 82s, i.e. ~5% run-to-run noise, against 20-70% on the shards that lost work. The run was not uniformly faster — `Build` went 213s -> 226s and `Build soak archive` 159s -> 192s — so the drops are attributable rather than ambient.

**The corpus tests really ran.** Zero `Skipping:` lines in the job log, `FERRO_REQUIRE_BULK_FIXTURES=1` set, and the 39.2s/33.7s timings are three orders of magnitude off a skip. `spec_conformance_axis`'s pins are green in both directions at `opt-level = 3`; none moved.

## The bottleneck moved, and it is not what the projection said

The projection had `Test (3/6)` as the next critical path at ~165s, bounded by `merge.rs`'s unit half of #1840's pair. **That is not what happened.** `Test (3/6)` came in at 112s, and the tail of the workflow is now entirely on the soak side:

```
     1 ->    193  ( 192s)  Build soak archive (optimized)   <- the new gate
   197 ->    307  ( 110s)  Exhaustive sweeps                <- the new critical path
   199 ->    289  (  90s)  Slow censuses (optimized)
   311 ->    314  (   3s)  Test (aggregator)
```

So the floor is now `test-build-soak` (192s of `opt-level = 3` compile, SHA-keyed and therefore a miss on every new commit) plus `sweeps` (110s) — about 305s, and this run came in at 314s. **The workflow is within ~10s of what its current job graph allows**, and nothing in the `test` / `test-oracle` fan-out is on the critical path any more.

Two consequences for what to do next, neither of which is in this PR:

- **`merge.rs`'s unit half is no longer the binding constraint.** It is still 149.6s in `Test (3/6)` and 95.8s in `Test oracle (3/4)`, but both jobs now finish 90s before the tail does, so moving it would buy nothing until the soak side moves first. It also still cannot move cheaply — `test-build-soak` archives `--test it` only, and `canonicalize_from_sequence_with_rule` / `partition_rule_from_env` / `PARTITION_RULE_NAMES` are private to `merge.rs`. Adding `--lib` to that archive is now *actively* the wrong move: it lengthens the very job that has become the gate.
- **The next real lever is the soak archive build itself**, which was 159s on the baseline and 192s here, and is on the critical path in a way it was not before this PR. Worth its own issue rather than a rider on this one.

**The fan-out gate improved for reasons unrelated to this PR** and should not be credited to it: `spec-fixtures` ran 126s -> 101s and `test-build` 116s -> 100s, so the shards started at ~105s instead of ~150s. That is ambient, and it is why the per-job control pairs above matter more than the headline number.

## The compile audit

Asked for alongside the wall-clock work: what does the fan-out recompile, and is any of it shareable? Measured on the same run, counting `Compiling`/`Checking` lines per job (after stripping ANSI — `CARGO_TERM_COLOR: always` means the naive `grep -c 'Compiling '` returns 0 on a job that compiled 316 crates).

**The fan-out recompiles nothing.** 19 of 33 jobs — all six `test` shards, all four `test-oracle`, all eight `soak`, and `sweeps` — compile zero crates; they consume prebuilt nextest archives. `censuses` joins them.

**Seven jobs compile, and each compiles `ferro-hgvs` exactly once.** `Swatinem/rust-cache` restores the dependency tree, so the count per job is 1 (7 in `spec-fixtures`, which builds seven generator targets):

| job | profile / features | `ferro-hgvs` |
|---|---|---:|
| Clippy | dev check, default | 53s |
| Clippy (all features) | dev check, all | 54s |
| Spec fixtures | dev, `dev` | 30s |
| Build test archive | test, `dev` | 82s |
| Python Wheel Test | release, `python` | 103s |
| Build soak archive | soak (`opt-level = 3`), default | 118s |
| Build | release clippy + test + release build | ~170s |

**There is no redundant recompilation to remove.** Those are seven different artifacts — profile x feature-set x target-kind — and no cache can share them. `sccache` does not help either: `ferro-hgvs` is a single crate, so any `src/` edit invalidates the whole entry, and the dependencies it would otherwise cache are the ones `rust-cache` already holds. `CARGO_INCREMENTAL` is a dead end for the same reason plus a second one — `rust-cache` deliberately deletes workspace artifacts, so incremental state never survives a job.

**Corrected after reading step timings rather than log lines.** An earlier revision of this section named `Build`'s `cargo clippy --release` as ~2m24s of duplicated work. It is **14s**. Two `Finished` lines in that job's log were mapped to the wrong steps; the GitHub step API is authoritative, and it reads:

| step | |
|---|---:|
| `cargo build --release` | **152s** |
| `cargo test --no-run --lib --bins` | 31s |
| `Swatinem/rust-cache` restore | 17s |
| `cargo clippy --release -- -D warnings` | **14s** |

So there is no meaningful clippy duplication to remove. The 152s is a genuine full-LTO link — `[profile.release]` sets `lto = true` and `codegen-units = 1` — and `Python Wheel Test` performs a second one (111s, `python` feature). Those are the two expensive compiles outside the archives, they are not duplicates of each other, and neither is on the critical path today (`Build` ends at 227s against a 314s wall).

Representation-Change: none. `.github/workflows/ci.yml`, `scripts/run_oracle_suite.sh` and two files under `tests/it/` only. No file under `src/normalize/`, `src/hgvs/`, `src/spdi/`, `src/project/`, `src/reference/` or `src/error_handling/` is touched, no test's body or assertions change, and no normalized output can move — every test named here runs the same corpus with the same flags, on a different runner and at a different `opt-level`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,51 @@ env:
   SWEEP_FILTER: >-
     test(cis_junction_crossing_shift) + test(repeat_span_sibling_overlap)
     + test(issue_1254_sibling_crossing_shift)
+  # The slow census modules, owned by the `censuses` job and excluded from the
+  # `test` / `test-oracle` partitions. Same shape as `SWEEP_FILTER` above and
+  # for the same reason; the difference is what the job does with them.
+  #
+  # WHY THEY MOVED. `--partition hash:K/N` assigns a test by a hash of its NAME,
+  # so it balances shard *counts* and is blind to cost — the property already
+  # recorded on `sweeps`, where all three exhaustive sweeps hashed into one
+  # shard. These four hashed into the two slowest jobs and made the workflow's
+  # critical path a single test. Measured on run 31845625298 (`044f085a`):
+  #
+  #                                             Test (K/6)   Test oracle (K/4)
+  #   issue_1542_direction_symmetry               155.7s (3)         240.3s (1)
+  #   spec_conformance_axis (5 tests)             338.0s              excluded
+  #   normalize_axis_preserving (4 tests)         119.3s (1,3,5)      313.0s (1,3)
+  #
+  # `Test oracle (1/4)` ran 328s against 132-160s for its siblings and finished
+  # 83s after every other job in the workflow, so the run's wall clock was one
+  # unlucky hash draw. A shard's wall can never fall below its slowest single
+  # test, so re-sharding cannot fix it — which is why `test-oracle`'s own
+  # comment below, having chosen N=4 to separate the two bulk corpora, names
+  # exactly this carve-out as the robust alternative. This is that alternative.
+  #
+  # AND WHY THE JOB, NOT THE JOB BOUNDARY, IS THE WIN. `censuses` consumes the
+  # OPTIMIZED archive `test-build-soak` produces, which is where `sweeps`
+  # measured 7.6-8.7x on work of exactly this shape (compute-bound cross
+  # products over `MockProvider`). A carve-out onto the standard debug archive
+  # would move the same 240s to a different runner and buy ~50s.
+  #
+  # THE PARTITION INVARIANT is `SWEEP_FILTER`'s, verbatim: the counts drift as
+  # tests are added, the sum must not. `censuses` selects ON this filter and
+  # passes `--no-tests=fail`, so a rename that made it match nothing reddens
+  # there rather than quietly emptying the job, while `test` and `test-oracle`
+  # negate it and would just widen. `tests/it/census_filter_invariant.rs` closes
+  # the half neither of those catches — a module named here that NEITHER of the
+  # job's two steps selects would be negated everywhere and run nowhere.
+  #
+  # NOTE the overlap with `ORACLE_EXCLUDE` below: `spec_conformance_axis` is in
+  # both, and deliberately. `ORACLE_EXCLUDE` is a statement about which
+  # instrument may be armed, and `oracle_exclude_invariant.rs` demands that
+  # every spec-corpus module appear there; this filter is a statement about
+  # where a module runs. Removing it from either would be wrong for a reason the
+  # other does not cover.
+  CENSUS_FILTER: >-
+    test(issue_1542_direction_symmetry) + test(normalize_axis_preserving)
+    + test(spec_conformance_axis)
   # The spec-corpus modules, excluded from `test-oracle` ONLY. They run in full
   # in the plain `test` job; this is not a coverage exemption, it is a statement
   # about which instrument may be armed while another is measuring.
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

PR #1965, opened 2026-08-15
feat: add sequence_normalize() helper

nh13:
### Summary

Adds a public `sequence_normalize()` helper that derives a description from the reference/alternate pair and widens its window until the placement stops resting on an edge, so a string-derived variant converges to a single representation.

### Attribution

Authored by **Tim Dunn** (@TimD1). This branch reproduces the work of #1930 on a fresh branch under `origin`, commit-for-commit, because `maintainerCanModify` is `false` on the original and it could not be updated in place. Authorship is preserved on every commit; only the `committer` field differs.

### Adaptation by us

One commit of ours, `fix(test): call placement_bounded_by_window() at the case-harvest site`. This change converts `DerivedDescription::placement_bounded_by_window` from a public field into a method and converted every caller that existed when it was written; `main` has since gained one more, in `case_harvest`, still using the field form. The merge is textually clean and does not compile without this. It is the identical edit the change already makes at its own call sites. The Python surface is unaffected — the binding exposes the value through `#[getter]`, so attribute access there is unchanged.

### Notes

Representation-Change: none. No description ferro emits changes, argued from the diff rather than assumed. `sequence_normalize()` is a **new** entry point that nothing inside the crate calls: the only references to it under `src/` are the Python binding and two doc comments, so no existing path reaches it. The `placement_bounded_by_window` field-to-method conversion is **value-preserving** — the flag was `any(ref_start == 0 || ref_end == len)` and is now `any(ref_start == 0) || any(ref_end == len)`, the same predicate distributed over the disjunction, so it reports the same boolean for every input. `from_sequences_detailed`'s emitted `variant` is untouched; `src/normalize/mod.rs` is purely additive with no removed lines; and `CANONICAL_PAD` changed visibility only, not its value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **`Normalizer::from_sequences`** — the same derivation against a held reference, which lets it
   additionally range-check the interval and optionally `normalize` the result.
 
+- **`Normalizer::sequence_normalize`** — the "one canonical description per variant" round trip, as
+  a single call. It takes a parsed description to the bases it denotes (`to_sequences`), re-derives
+  a description from those bases alone (`from_sequences`), and — while a member still rests on a
+  window edge that can still move — doubles the pad and retries, so two spellings of one variant
+  reach one description decided by the observed bases rather than by how either was written. The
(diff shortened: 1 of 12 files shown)


nh13:
@coderabbitai pause

PR #1966, opened 2026-08-15
fix: allow ref span insertions when converting to SPDI

nh13:
### Summary

Resolves a same-reference position-range insertion (`g.100_101ins50_57`, its `inv` and `delins` spellings) to its literal bases when converting to SPDI, instead of refusing it as "not a literal sequence".

### Attribution

Authored by **Tim Dunn** (@TimD1). This branch reproduces the work of #1933 on a fresh branch under `origin`, because `maintainerCanModify` is `false` on the original and it could not be updated in place. Authorship is preserved on every commit; only the committer differs.

Stacked on #1965 (`td/td_add-sequence-normalize-helper`), mirroring the original stack.

### Adaptation by us

One commit of ours, `fix(spdi): read a range-insert payload on the description's own axis`.

`resolve_position_range_insert` passed the payload's raw coordinates to `fetch_reference_bases`, which reads offsets off the accession with no CDS shift — but by the time that arm runs, a `c.` (or coding `r.`) *location* has already been through `resolve_cds_to_tx` / `resolve_rna_pos`, so the SPDI is emitted on the flat transcript. The payload was therefore read `cds_start - 1` bases 5' of what it names. Measured against the prepared reference on `NM_000532.5` (`cds_start = 37`):

```
NG_008939.1(NM_000532.5):c.156_157ins180_188
  normalize     -> c.156_157insGCGAGGAAA        (transcript 216..=224)
  hgvs_to_spdi  -> NM_000532.5:192::GGGAGGGGG   (transcript 180..=188)  <- wrong
```

`g.`/`m.`/`o.`/`n.` were unaffected, because their positions already are offsets on the accession being fetched — which is why the genomic tests that shipped with the change could not see it. The payload span is now translated through the same `CoordinateMapper::cds_to_tx` the location uses, with `Rna` resolved coding-aware exactly as `resolve_rna_pos` does. Four hermetic regression tests cover the axis, including an `n.` negative control that fails if the translation is applied as a blanket shift; all four were verified to fail against the unfixed helper.

### Notes

Representation-Change: a same-reference position-range insert (`g.100_101ins50_57`, its `inv` and `delins` spellings, and the single-part bracketed form) now resolves to its literal bases through `hgvs_to_spdi` instead of being refused.
  No normalized HGVS string moves — nothing under `src/normalize/` reaches `src/spdi/convert.rs`, and the normalizer already expanded these payloads through `rules::expand_inserted_sequence`; what changes is which descriptions are convertible, not what a convertible one converts to. On the `c.` and `r.` axes the payload span is resolved through the same coordinate mapping as the location, so a coding-axis range insert names the bases it says it does.
  18 of the 9,949,738 rows across the four release-asset corpora carry the shape. Every affected input was previously rejected, so no consumer holds a stored old value; the visible consequence is those rows leaving `unkeyable` for a `group_by_spdi_key` bucket.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,7 +94,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   a provider-backed path (`sequence_normalize`, `to_spdi`, `canonical_spdi`, apply-to-reference)
   needed its bases. It now reads those bases from the reference exactly as `del`/`dup`/`delins`
   read their omitted bases, so `sequence_normalize` re-derives the literal `g.100_101insCGTA…`
-  form. `delins` with a same-reference range insert is resolved on the same path.
+  form. `delins` with a same-reference range insert is resolved on the same path. The payload is
+  read on the **axis the description is written on**: a `c.` (and a coding `r.`) payload gets the
+  same `cds_start + N - 1` shift the location already gets, so `NM_000532.5:c.156_157ins180_188`
+  resolves to `insGCGAGGAAA` — the bases `normalize` names — rather than to the transcript-offset
+  span 36 bases 5' of it. `n.`/`g.`/`m.`/`o.` payloads are transcript/genomic offsets already and
+  are read unshifted.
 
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

PR #1967, opened 2026-08-15
fix: allow repeat/complex variants to be converted to SPDI

nh13:
### Summary

Resolves two further insertion payloads when converting to SPDI — an exact tandem repeat (`insC[4]`) and a bracketed compound insert whose parts mix literals, same-reference spans and exact repeats (`delins[T;213_271]`) — instead of refusing them.

### Attribution

Authored by **Tim Dunn** (@TimD1). This branch reproduces the work of #1935 on a fresh branch under `origin`, because `maintainerCanModify` is `false` on the original and it could not be updated in place. Authorship is preserved on every commit; only the committer differs.

Stacked on #1966 (`td/td_allow-ref-span-insertions`), mirroring the original stack.

### Adaptation by us

One commit of ours, `test(spdi): cover the compound insert's span part on the c. axis`. This change routes bracketed span parts through the same `read_reference_span` leaf as a bare span, so they inherit the axis handling added in #1966; the test pins that a part inside a bracket lands in the same coordinate frame as one standing alone (`c.1_2ins[T;3_5]` reads `c.3_5`, not the unshifted transcript span). No behaviour of this change was otherwise altered.

### Notes

Representation-Change: two further insertion payloads now resolve through `hgvs_to_spdi` instead of being refused — an exact tandem repeat expanded under the existing `MAX_REPEAT_EXPANSION_BASES` cap, and a bracketed compound insert whose parts mix literals, same-reference spans and exact repeats, each part resolved by the leaf that resolves it alone. A part whose bases are undetermined (uncertain or range count, intronic CDS offsets, external accession) still declines the whole insert, with one exception that predates this PR and is not closed by it: an `N`-unit repeat such as `insN[341]` carries a length and no bases, exactly as `ins341` does, but resolves to a run of `N` characters rather than declining, because there is no ambiguity guard on the SPDI path and the sibling short-form arm already ships that behaviour on main. Six such rows exist in the release-asset corpus (`insN[18]`, `delinsN[341]` x2, `insN[1200]`, `insN[320]`, `delinsN[155]`); this PR widens the spellings that reach the class without introducing it. Tracked as #1975.
  No normalized HGVS string moves — nothing under `src/normalize/` reaches `src/spdi/convert.rs`.
  Across the four release-asset corpora (9,949,738 rows), 3 rows carry a bare exact-repeat insert and 52 a bracketed compound insert, of which roughly ten have all-same-reference parts; the rest still decline. Every affected input was previously rejected.


nh13:
@coderabbitai pause

PR #1969, opened 2026-08-15
fix(normalize): re-present a contig-start insertion instead of refusing it

nh13:
### Summary

Re-presents a pure insertion that 5'-shuffles onto interbase 0 of a window beginning at the accession's first base, instead of refusing it with "5' of the window's first base".

### Attribution

Authored by **Tim Dunn** (@TimD1). This branch reproduces the work of #1957 on a fresh branch under `origin`, because `maintainerCanModify` is `false` on the original and it could not be updated in place. Authorship is preserved on both commits; only the `committer` differs.

### Stacking — a deliberate change from the original

The original #1957 was based on #1942. This branch is based on **#1967** (`td/td_allow-repeats-to-spdi`) instead, skipping #1942.

#1942 is not being reproduced as-is: #1837 landed the same behaviour on `main` roughly seven hours after #1942 was written, and its remaining half (replacing `triples_are_disjoint` with an inline watermark) is a separate question deferred to #1749. See #1968 for the part of #1942 that is preserved.

This re-basing is safe rather than convenient, and was checked rather than assumed. #1957 touches `CHANGELOG.md`, `src/normalize/from_sequences.rs`, `src/normalize/merge.rs` and `tests/it/sequence_normalize.rs` — and **zero** lines of `src/spdi/apply.rs`, where #1942's change lives. Both commits applied with no source conflict; the only conflict was in `CHANGELOG.md`, because the original changelog commit bundles #1942's entry together with this one. That entry was dropped here since #1942 is not in this stack. Because "applies cleanly" is necessary and not sufficient, the full suite was run on this branch, not just its three targeted tests.

### Notes

Representation-Change: a pure insertion that 5'-shuffles onto interbase 0 of a window beginning at the accession's first base is now re-presented at the leftmost nameable interbase (`g.1_2insTA`) or folded into a base-1-anchored delins, instead of refusing with "5' of the window's first base".
  Previously-refused inputs therefore start producing a description; no description ferro already emitted is spelled differently, because the escape fires only where the old path errored. A lone insertion genuinely before base 1, with no piece to absorb it, still refuses.
  Bounded to `w_lo == 1` — at any interior window the answer remains to widen the 5' flank, not to relabel.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,6 +120,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   whose bases are genuinely undetermined — an uncertain or range repeat count, a CDS position range
   carrying intronic offsets, or an external reference SPDI cannot dereference — still declines the
   whole insert.
+- **A `from_sequences` derivation whose insertion rolls to the contig's first base is now
+  re-anchored rather than refused.** When the observed window begins at position 1 — a variant
+  anchored at `1A>C`/`1A>T`, say — the 5'-shuffle can roll a pure insertion in an ambiguous run
+  onto interbase 0, whose only HGVS spelling names position 0, a base that does not exist. Because
+  the window already starts at the accession's first base, `sequence_normalize` cannot widen the 5'
+  flank to escape it, so these derivations were dropped with "the derivation places an inserted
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

Issue #1970, opened 2026-08-15
fix(normalize): u16 cost grid overflows on a large insertion, guarded only by a debug_assert


nh13:
## Summary

PR #1952 narrows the alignment cost grid to `u16`. On an input that `origin/main` answers correctly today, the narrowed grid overflows:

```
g.1000_1001ins<70,000 bases>
```

- **`origin/main`**: normalizes fine — output length 70,024, 0.023 s.
- **PR #1952's head**: `edit_grid narrowed to u16: 2*(n+m) = 140000 exceeds u16`, with `n+m = 70,000`.

## Why this ships silently

The bound is enforced by a `debug_assert`. `[profile.release]` sets neither `debug-assertions` nor `overflow-checks`, so in the **shipped library** the arithmetic wraps rather than panicking. A wrapped cost produces a wrong alignment, not a crash.

## Why the existing gates do not catch it

Every gate in front of `AlignmentDag::build` bounds **cells**, not `n + m`. A `1 × 70,000` grid clears `MAX_SEQFIRST_GRID_CELLS` by four orders of magnitude while `n + m` is well past `u16::MAX`.

The PR's own correctness evidence cannot see this either: its brute-force oracle is exhaustive over `n + m <= 10`.

## Consequence for the trailer

`Representation-Change: none` is **inaccurate on this class**, not merely unproved. The checker passed because it rejects silence, not inaccuracy.

## Suggested fix

Make the bound a real refusal rather than a `debug_assert` — returning `None`, which both callers already handle. That is also what makes `Representation-Change: none` true, since an input that is refused does not produce a moved representation.

## Second, separate finding: the benchmark's `diverge_indels` generates substitutions

`diverge_indels` pushes a random base *instead of* the reference base and then `continue`s, so output length equals input length and `edit == hamming` at every `k`. Verified by re-implementing it and running a Levenshtein DP: k=8 → 2 differences, k=32 → 12, k=128 → 47. Roughly 25% of draws are no-ops because the replacement base is not excluded from being the incumbent. The dead `truncate`/pad tail confirms the intent was different.

This means the PR's stated second measurement axis is not measuring indel divergence. The memory table is unaffected and checks out arithmetically (4 bytes x 14,753,281 cells -> 59.0 MB predicted; 18.67 -> 14.60 bytes/cell; -21.8%).

## Provenance

[...]

Issue #1959, opened 2026-08-14
The brute-force dominators oracle cannot detect a DAG that is uniformly too small


nh13:
`dominators_agree_with_brute_force_on_exhaustive_small_inputs` (`src/normalize/seqfirst/align.rs:525`) is the strongest-sounding guard on the sequence-first alignment DAG, and it is cited that way — `rulings[unchanged-is-read-over-every-minimal-alignment]` supports its distinction with "`dominators()` was checked against its own brute-force oracle and agrees with it".

The wording there is accurate ("its own"), but the guard is weaker than it reads, in a specific and demonstrable way: **it cannot detect a DAG that is uniformly too small.**

## Why

`dominators_by_brute_force` (`:441`) takes the `AlignmentDag` and walks it:

```rust
fn dominators_by_brute_force(dag: &AlignmentDag) -> (Vec<(u32, u32)>, Vec<u32>) {
    for (i, j) in dag.cells().collect::<Vec<_>>() {
        for (ni, nj, step) in dag.out_edges(i, j).collect::<Vec<_>>() {
            // ban this edge, then DFS from (0,0) over dag.out_edges(..)
```

Every term — the cell set, the edges, the reachability walk — comes from the DAG. Nothing is re-derived from the two sequences. So what it establishes is that `dominators()` agrees with edge-banned reachability **over that same graph**. If the graph itself is missing cells, both sides are missing them together and the comparison is satisfied.

That is the definition of an oracle that reads the artefact rather than the thing the artefact claims to describe, and it fails in exactly the direction you cannot see: a *smaller* DAG has *more* forced junctions, so a shrunken graph produces a self-consistent, plausible, wrong answer.

## Demonstrated, not argued

Found while implementing banded DP for #1928. Deliberately narrowing the band by one doubling step — accepting a band-confined distance of `k+1` where only `<= k` is sound — produces a DAG that on `AC -> CA` retains **3 cells where the true DAG retains 7**.

`dominators_agree_with_brute_force_on_exhaustive_small_inputs` **passed** on that sabotage.

Two other sabotages in the same set (band centred on `0` instead of `delta/2`; `cells()` clipped one diagonal too tight) *were* caught by it, which is worth stating plainly — the guard is not useless. It catches asymmetric and inconsistent damage. It is blind to uniform shrinkage.

[...]

PR #1973, opened 2026-08-15
perf(normalize): band the sequence-first alignment DP

nh13:
Advances #1928. **No longer stacked — #1952 has merged** (`e93652c5`), so this is now a single commit against `main`, two files.

**Rebased twice on 2026-08-15.** First onto #1952's branch head (`d5961dad`), which added the `MAX_ALIGNMENT_SPAN` refusal for #1970 and fixed `diverge_indels` — both touched the code this PR rewrites, so that reconciliation is described under *Composing with the span refusal* below rather than left as a silent merge. Then, once #1952 squash-merged, with `git rebase --onto origin/main d5961dad` to drop the three commits `main` had absorbed.

Worth stating because it reads alarmingly and is not: between those two steps GitHub reported this PR `CONFLICTING / DIRTY`. That is the **stacked-PR-whose-base-merged** shape, not a textual conflict — the base's commits are on `main` under a different sha, so a plain `git rebase origin/main` would try to replay them and die, while `--onto` past the merged base replays only this PR's own commit. It did so cleanly, with no conflicts.

The rebase is verified by content rather than by ancestry, because a force-push after a rebase can never be a fast-forward and the usual `--is-ancestor` test therefore proves nothing:

```
$ git range-diff d5961dad..5ae5aaa0 origin/main..HEAD
1:  5ae5aaa0 = 1:  d6b846ab perf(normalize): band the sequence-first alignment DP
```

`=` means this commit is byte-identical in content on both sides. The three commits the push drops are #1952's own, which `main` absorbed as the squash `e93652c5`.

Confines the sequence-first alignment DP to the diagonals a minimal alignment can occupy. **Storage is deliberately left at full `(n+1) x (m+1)` shape**, which turns out to be the headline result rather than a footnote — see the measurement below.

## The band, and why it is exact rather than heuristic

A cell `(i, j)` on diagonal `d = i - j` costs at least `|d|` to reach from the source and at least `|delta - d|` to reach the sink from, so every cell on an alignment of cost `k` satisfies `|d| + |delta - d| <= k`, i.e.

```
ceil((delta - k) / 2)  <=  d  <=  floor((delta + k) / 2)
```

`k` is **found, not chosen**: doubling from `max(|delta|, 1)` until the band-confined distance no longer exceeds its own band. Confining the DP can only over-estimate, so a distance the band itself admits is provably the true one; at `k = n + m` the band is the whole grid, so the ladder terminates exactly. There is no seed, no anchor, and no tie-break — `on_optimal_path` **is** the set of cells on some minimal alignment, and every such cell is inside the band by construction. The DAG comes out bit-identical.

`lo + hi == delta` holds by parity, which is load-bearing rather than decorative: reversing both blocks maps `d -> delta - d`, so the *same* interval bands the suffix grid with no second derivation.

## Correctness

Five differential tests keep the **pre-band implementation verbatim** as a second implementation — not "the banded code with a full band", which would test the parameter rather than the algorithm — and diff distance, `on_optimal_path`, `out`, the `cells()` walk and `dominators()`. They cover exhaustive small inputs, the named shape boundaries (empty blocks, pure ins/del, `k=0`, `k=1`, `|delta|=17` both signs), every doubling boundary, the whole divergence range, and a 512-base block at sliver-band widths.

**Three sabotages were run to prove the harness discriminates**, and one produced a finding worth more than the change: accepting a band one doubling too narrow is caught **only** by the differential tests. `dominators_agree_with_brute_force_on_exhaustive_small_inputs` *passes* on it, because that probe is computed from the DAG itself — a uniformly shrunken DAG is internally consistent. Filed separately as **#1959**.

A sixth test asserts the band is **narrow**, not merely correct. Every differential test passes trivially against a full-width band, so without it a refactor that stopped narrowing would delete the optimization with the whole suite green. Verified by sabotage: forcing a full-width band gives **95 passed, 1 failed**, that test being the one failure.

## Measured — and the geometry over-predicts by ~148x

Quiet dedicated `c7i.4xlarge` (x86_64, 4 kB pages), criterion against this PR's own base, so banding is the only variable. **Noise floor established first from the same binary against itself: −1.0% to +3.9%**, with a +2–4% bias against whichever arm runs later (the banded one). Criterion's own significance test reported "Performance has regressed" at p = 0.00 on that self-comparison five times, so its verdict line is not used here — only magnitudes against the measured floor.

| n (k = 2) | speedup | cell-count ratio |
|---:|---:|---:|
| 256 | 37.0x | 74x |
| 1024 | 6.6x | 293x |
| 4096 | **7.9x** | **1171x** |

Against divergence at n = 1024: 33.3x at `k=2`, 13.3x at `k=32`, 5.1x at `k=128`, 1.56x at `k=512`, and **0.90x (11.2% slower) at `k=614`** — the uniform-random regime where the band is the whole grid and the ladder is overhead paid for nothing. Real variants sit at single-digit `k`.

**The interesting number is the gap.** At `n = 4096` the band writes ~1171x fewer cells and runs only 7.9x faster, because `build` still allocates and zeroes ~101 MB of full-size grids: 101 MB / 20 ms is ~5 GB/s, i.e. the banded time is essentially all allocation and no DP. So this commit converts `build` from compute-bound to **allocation-bound**, and narrowing the *storage* to the band is now the lever rather than a tidy-up.

Two consequences recorded in the code because they are not obvious:

- **Do not quote a single speedup.** `by_length_fixed_k/len1024_k2` and `by_divergence/k2` are the *same* computation and measured **5.8x apart**, purely on whether a large arena had already been faulted in. Their unbanded baselines differ by only 15%.
- **Peak memory is not measured here.** Criterion has no memory axis and the RSS harness was not run against this commit. The time win is not a memory win, and the doc says so. The page-size reasoning differs between 16 kB (Apple silicon) and the 4 kB Linux host that produced these numbers; both are stated as predictions, neither as a result.

## Composing with the span refusal (#1970)

`build` is now `Option<Self>`, refusing past `MAX_ALIGNMENT_SPAN`. That is checked **before** the band is derived, so it bounds the whole computation and not merely the fill: `k` is itself bounded by `n + m`, and every value the banded DP writes is a prefix or suffix cost under the same bound as the full sweep. Banding narrows which cells are *visited*, never how large a cost can *get*.

Two consequences, both verified rather than argued:

- **`the_span_bound_is_enforced_from_both_sides` passes unchanged** against the banded build. Its accepted-boundary case is a `1 x 65,534` alignment, where the sentinel `UNREACHED = u16::MAX` sits exactly one above the largest real cost — and the banded fill's `saturating_add` makes that boundary strictly safer than the unbanded `+` it replaces.
- The pre-band `debug_assert!(2 * (n + m) < UNREACHED)` this PR originally carried is **deleted**, not kept alongside. It was both weaker (a `debug_assert` in a profile that disables them) and wrong about the bound — the exact one is `n + m <= u16::MAX`, which #1970's analysis derives.

The module-level `edit_grid` is removed, since `fill_edit_grid_banded` replaces it. It survives *only* inside the differential harness, which is precisely where the pre-band implementation belongs.

## Also in this commit

`benches/seqfirst_align.rs`'s header carried `~18 bytes per grid cell, ~275 MB`, which #1952 superseded when it narrowed the grid to `u16` (14.60 B/cell, 215.4 MB). That staleness is the base's rather than this PR's, but it sits in a header this commit is already rewriting, and leaving a known-wrong figure in a file being edited is worse than fixing it out of band.

The base's `diverge_indels` fix is **inherited, not duplicated** — this PR's header edits and that fix merged cleanly, and the indel arm now generates genuine indel pairs rather than substitutions.

That has a consequence for what this PR may claim. The EC2 run above predates the fix, so its `indel_vs_sub` group was comparing substitutions against substitutions; **no figure from that group is quoted here, and none should be** until it is re-run. The two groups that *are* quoted — `by_length_fixed_k` and `by_divergence` — both draw from `diverge_subs`, which was correct throughout, so they stand.

## Verification

Re-run in full **after** the rebase onto `main`, not carried over from the pre-rebase head:

- `cargo nextest run --features dev --no-fail-fast` — **11,036 passed**, 0 failed, with the prepared reference **armed** (`FERRO_MANIFEST` set to a manifest confirmed present) and `FERRO_REQUIRE_BULK_FIXTURES=1`, all four bulk fixtures digest-verified. **Zero `Skipping:` lines**, so the reference-aware suites genuinely ran rather than standing down.
- Python — **1,059 passed**. `cargo nextest` does not reach `tests/python/`, and this PR touches `src/normalize/`, so a green Rust gate says nothing about the bindings on its own.
- `cargo fmt --all --check` and `cargo clippy --all-features --all-targets -- -D warnings` — clean.
- Both gitignored spec artifacts regenerated first. A rebase never updates them, and a stale one fails as `enumeration_replays_recorded_behavior`, which reads like a destroyed ruling rather than like its own staleness.

Every figure above was read back out of the run's own `EXIT_` lines rather than from the job's exit status, since the driver ends in an `echo` and would report 0 regardless.

Representation-Change: none. Measured, not assumed. The DAG is bit-identical — retained cells, out-edge masks and `dominators()` all diff clean against the pre-band implementation over exhaustive small inputs, every doubling boundary and the whole divergence range, and the full gate passes with every existing pinned value unchanged. No description ferro emits can move, because nothing here changes which cells lie on a minimal alignment - only which cells are examined to find them.

### DIFF
--- a/benches/seqfirst_align.rs
+++ b/benches/seqfirst_align.rs
@@ -1,11 +1,20 @@
 //! Sizing the sequence-first DP.
 //!
-//! `AlignmentDag::build` is `O(n*m)` in time **and space**, but Criterion has
-//! no memory axis — this measures wall-clock time only. It shows where that
-//! grows too slow to be acceptable, so the block-size cap in the migration is
-//! chosen on time evidence; the memory dimension is not measured here (the
-//! figures are on #1928: ~18 bytes per grid cell, ~275 MB at the widest
-//! accepted block).
+//! `AlignmentDag::build` **allocates** `O(n*m)` and, since the band landed,
+//! **fills** only `O((n+m)*k)`. Criterion has no memory axis, so this measures
+//! wall-clock time only, and the block-size cap in the migration is chosen on
+//! time evidence.
+//!
+//! The memory dimension is **not** measured here; the figures are on #1928, and
+//! note they moved when the cost grid narrowed to `u16`: **14.60 bytes per grid
+//! cell, 215.4 MB** at the widest accepted block (`W = 3840`), down from 18.67
+//! and 275.4 MB. This header carried the older pair until the banding change.
+//!
+//! Because the allocation is unnarrowed while the fill is narrowed, a timing
+//! here is partly an allocator measurement: at `n = 4096` the banded arm runs at
+//! roughly memory bandwidth, and two groups that compute the *same* thing can
+//! differ several-fold purely on whether a large arena has already been faulted
+//! in. Compare each group against its own baseline, never across groups.
 //!
 //! # Divergence is an AXIS, not a constant (#1928)
 //!
--- a/src/normalize/seqfirst/align.rs
+++ b/src/normalize/seqfirst/align.rs
@@ -18,6 +18,82 @@
 //!
 //! There is also **no seeding from the input's own alignment** — that would
 //! reintroduce input-relativity by a new route.
+//!
+//! # The DP is banded, and the band is exact (#1928)
+//!
+//! A cell `(i, j)` sits on diagonal `d = i - j`. Reaching it from the source
+//! costs at least `|d|`, because every step off the main diagonal is an indel;
+//! reaching the sink from it costs at least `|delta - d|`, where
+//! `delta = ref_len - alt_len`. So every cell on an alignment of cost `k`
+//! satisfies `|d| + |delta - d| <= k`.
+//!
+//! `on_optimal_path` **is** the set of cells lying on some minimal alignment, so
+//! every cell this module retains already satisfies that bound for `k` = the
+//! edit distance. Computing only the band is therefore a **cost** change and not
+//! a semantic one: the retained cells, the out-edge masks and
+//! [`AlignmentDag::dominators`] come out bit-identical to what the full
+//! `(n + 1) x (m + 1)` sweep produces. That is not left to the argument: the
+//! five `the_band_agrees_with_the_full_grid_*` tests keep the pre-band
+//! implementation verbatim and diff the two builds field by field.
+//!
+//! **They are not redundant with the oracles already here.** Deliberately
+//! widening the band's acceptance test by one — so a band one doubling too
+//! narrow is accepted — is caught *only* by those tests:
+//! `dominators_agree_with_brute_force_on_exhaustive_small_inputs` passes, because
+//! the brute-force probe is computed from the DAG itself and a systematically
+//! narrowed DAG is internally consistent. An oracle that reads the artefact
+//! cannot see the artefact being too small.
+//!
+//! The band is found by doubling `k` until the band-confined distance no longer
+//! exceeds its own band, which costs at most twice the final pass. The **DP** is
+//! then `Θ((n + m) · k)` rather than `Θ(n · m)`.
+//!
+//! # `build` as a whole is not, and the measurement says so (#1928)
+//!
+//! **Storage is unchanged** — the grids keep their full `(n + 1) x (m + 1)`
+//! shape and their `i * width + j` addressing. So `build` still *allocates and
+//! zeroes* `Θ(n · m)` even though it now *fills* `Θ((n + m) · k)`, and past a
+//! certain width that allocation is the whole cost.
+//!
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai pause

Issue #1947, opened 2026-08-14
dump_normalized_corpus.rs's own header is stale in three ways: family counts (says 15/13/9, is 21/14/14), corpus totals (says two, has three), and MAX_SPLIT_BLOCK


nh13:
`examples/dump_normalized_corpus.rs` is where anyone goes to decide whether their change is measurable, and its module header is the first thing they read. Several of the numbers in that header are wrong, in the direction that makes the corpus look **broader** than it is. Every figure below was re-derived from `origin/main` rather than taken from the prose.

## 1. The family counts are wrong in three different ways

Counted from the source:

| group | count | how |
|---|---:|---|
| `FAMILIES` (`:505-592`) | **14** | `split_vs_spanning_delins`, `dup_plus_sub`, `adjacent_junction_ins`, `dup_plus_ins`, `del_plus_sub`, `coincident_bounds`, `junction_interior_to_span`, `partial_overlap_spans`, `nested_spans`, `delins_hiding_an_inversion`, `inv_member`, `three_member_allele`, `identity_member`, `repeat_expansion` |
| `PROTEIN_FAMILIES` (`:607-632`) | **5** | `protein_shift_del`, `protein_shift_dup`, `protein_ins_becomes_dup`, `protein_equal_length_delins`, `protein_cis_separated` |
| designed-reference singletons | **2** | `LONG_FAMILY`, `REVCOMP_FAMILY` |
| **total** | **21** | 16 DNA + 5 protein |

What the file says:

- **"the nine families there"** (`:633`, of `FAMILIES`) — it holds **14**.
- **"thirteen of the fifteen families"** — at `:115`, `:307`, `:477`, `:2106`, `:3132`. The sequence-keyed count is **14** and the DNA total is **16**, so this should read "fourteen of the sixteen". The two exceptions it means (label-keyed rather than sequence-keyed) are still exactly `LONG_FAMILY` and `REVCOMP_FAMILY`, so only the numbers are wrong, not the distinction.
- **"the thirteen sequence-keyed families"** (`:2008`) — **14**.

Five separate sites repeat "fifteen", which is why this is worth fixing in one pass rather than opportunistically: they will not all be found again.

## 2. The paragraph warning against reconciling totals has itself gone stale

`:86-93` says, in bold:

> **Two corpus totals appear in this file and both are correct — do not "reconcile" them.** `78,028` is the corpus *before* `separated_revcomp_runs` existed [...] `78,298` is the corpus now

[...]

PR #1974, opened 2026-08-15
test(corpus): reach the repeat-beside-a-sibling geometry, and land the canonical-spelling seam

### LINKED ISSUE #1946: normalize: presentational rewrites are minted before the stages that must compute on them
## Summary

The normalizer interleaves two kinds of rewrite.

**Semantic** rewrites — the 3'/5' shuffle, the cis collapse, the sequence-first partition — change *which variant* is described, or how it is cut into members. Every downstream stage has to be able to evaluate their output.

**Presentational** rewrites — choosing `dup` over `ins`, `T[6]` over `dup`, `ins<range>inv` over a literal payload — change only *how the same change is spelled*. `general.md:56` prioritisation is the rule that picks among them.

Presentational forms are currently minted **during per-member normalization**, before the stages that must compute on them. The codebase then pays a recurring tax to undo them, and it says so in its own words. From `demote_repeats_spanning_siblings` (`src/normalize/merge.rs:10833-10835`, `origin/main` `b3568679`):

> The conversion happens per member, deep inside `normalize_na_edit`, with no sibling context reaching it, so the decision is undone here instead

## The tax, as it currently stands

All references are against `origin/main` at `b3568679`.

**1. A non-literal insertion payload makes the whole cis allele un-mergeable.**
`collect_canonical_edits` (`merge.rs:3867`) reads an insertion payload with `let bases = sequence.bases()?;` (`merge.rs:3881`; the `delins` arm at `:3901` is the same). `InsertedSequence::bases()` (`src/hgvs/edit.rs:407`) returns `Some` for `Literal` and `None` for everything else — `PositionRange`, `PositionRangeInv`, `SequenceRepeat`, `Complex`, `Named`. So one non-literal payload drops the member, and because `collapse_overlapping_cis_edits` is all-or-nothing (`merge.rs:460-462`, `let Some(edits) = … else { return variants; }`) the **whole allele** is returned unmerged.

**2. `lower_repeat_edits` exists to undo repeat notation mid-pipeline.**
`merge.rs:4026`, called from `canonicalize_from_sequence` at `merge.rs:3242`, with the comment "Every `GEdit::Repeat` becomes the `ins`/`del` it means, and nothing past this line can see one (#1296)". It is a presentational form being lowered back into computable `Ins`/`Del` because it was minted too early.

**3. Two further passes exist only to demote a repeat a sibling-blind stage minted.**
`demote_repeats_spanning_siblings` (`merge.rs:10860`, called `mod.rs:4528`, #1296) and `demote_coincident_tract_repeats` (`merge.rs:11152`, called `mod.rs:4590`, #1316). The second's call-site comment states the mechanism exactly: "two members that each grew the SAME tract are both spelled as a copy count over it, and a repeat carries no junction — so the pair claims one tract twice and the coalesce never considers it". A fourth, `respell_colliding_duplications` (`merge.rs:11909`, called `mod.rs:4575`), demotes a `dup` back to the equivalent `ins` when its span collides with a sibling's bases.

**4. The overlap detector had to be taught that a `dup` writes at a junction.**
`writes_only_at_a_junction` (`src/normalize/overlap.rs:699`), whose own doc comment names the cost of the spelling reaching a semantic stage: "Grouping them by the read span reported a conflict that is not one, and **made the verdict depend on the spelling**: the same variant written `g.[23A>G;23_24insA]` was accepted." That is #1276 (an interior insertion normalized to a `dup` escapes overlap detection, and strict mode then accepts it) and #1437, with #1411 and #1448 in the same family.

**5. There is already an entry-side lowering pass, for the payload half of the same problem.**
`try_expand_genome_ins` (called `mod.rs:5488`, issue #333) expands bracketed and reference-range `ins` payloads to a flat literal "so the rest of the pipeline (3' shift, ins→dup, etc.) operates on concrete bases". That is the shape a lowering stage takes when it is placed at the boundary rather than in the middle — and it is why item 1 is not reachable from an ordinary authored input today.

## Reproducer for (1) — measured

Setup: `origin/main` `b3568679`, `MockProvider` over `padded("CGTTGCATGC")` (the `tests/it/common/synthetic.rs` helper, core at position 257), through `tests/it/common/cis_apply_oracle::normalize`.

Baseline:

```
TEMPLATE:g.[258_259insA;260del]  ->  TEMPLATE:g.259T>A
TEMPLATE:g.259T>A                ->  TEMPLATE:g.259T>A
```

Now add a purely presentational re-spell at the exit of the per-member genomic pipeline — a temporary wrapper around `normalize_genome` that rewrites a one-base literal insertion `insX` into the equivalent `ins<p>_<p>inv`, where `p` is a nearby position whose base is the complement of `X`. Nothing about the change moves, only the payload's spelling:

```
TEMPLATE:g.[258_259insA;260del]  ->  TEMPLATE:g.[258_259ins240_240inv;260del]
TEMPLATE:g.259T>A                ->  TEMPLATE:g.259T>A
```

(Position 240 is in the `ACGT…` pad and holds `T`; its reverse complement is `A`, so `ins240_240inv` and `insA` are the same payload. `g.258_259insA` on its own likewise re-spells to `g.258_259ins240_240inv` and nothing else changes.)

The cis collapse is gone: the allele stays two members. **One variant now has two normal forms** — `g.259T>A` and `g.[258_259ins240_240inv;260del]` — depending only on whether the renderer fired on one of its members.

To be precise about what this is and is not: it is a **confluence break, not a sequence change**. The two members are the same two members the baseline had, so the allele denotes the same bases it always did; what is lost is the merge. Running the probe under `FERRO_ASSERT_SEQUENCE=1` raises nothing, which is consistent with that reading (though a green denoted-sequence run is weak evidence on its own — see the counts caveat on `denoted_sequence_oracle_counts`). The argument does not need the stronger claim: a variant with two normal forms is already a rule-2 problem.

The re-spell was applied temporarily and reverted; nothing is committed.

## Proposed direction

One `render_canonical_spelling` stage that runs **after all derivation**, applying the `general.md:56` prioritisation ladder (`ins`→`dup`, `dup`→`repeat`, `ins`→`ins<range>inv`) once the member set is final. Every stage before it computes on `sub`/`del`/`ins`/`inv`/`delins` only, and the four demote/lower passes above have nothing left to undo.

## Known complications — stated up front

**`dup` is load-bearing downstream on purpose, and cannot simply move to render time.** The overlap detector keys on a `dup`'s junction write (`writes_only_at_a_junction`, #1276/#1411/#1437), and #1749 records that `dup`-as-junction-writer is the *correct* geometry. Moving the mint to render time means that geometry has to be re-derived from the underlying insertion rather than read off the spelling — which is a real piece of work, not a mechanical move.

**`lower_repeat_edits` is relocatable, not deletable.** Measured: an authored repeat inside a cis allele reaches it without ferro ever having minted anything. Instrumenting the lowering loop and probing over `padded("AAAAAAAGCATGC")`:

```
TEMPLATE:g.[257_263A[9];266T>G]   -> LOWER_REPEAT s=257 e=263 unit=A count=9   (repeat is AUTHORED)
TEMPLATE:g.[263_264insAA;266T>G]  -> LOWER_REPEAT s=257 e=263 unit=A count=9   (repeat is ferro-minted)
```

Only the second source disappears if minting moves to render time. The first is an input, so the body of `lower_repeat_edits` is still needed — it moves to the input boundary, beside `try_expand_genome_ins`, rather than being deleted. (A lone repeat member reaches neither: `g.257_263A[9]` on its own never enters the sequence-first pass.)

**`ins<range>inv` is the clean case.** Nothing in `src/` constructs `InsertedSequence::PositionRangeInv` — every occurrence outside the parser is a match arm — so it has no downstream consumer to break. It is also a form the spec asks for (`DNA/insertion.md:18`: "inverted duplications are described as an insertion (`g.234_235ins123_234inv`), not as a duplication"), which ferro does not currently mint. Under the present architecture, adding it would immediately incur the tax measured above; under a render stage it is free.

## Relationship to #1749

#1749 ("Overlap is defined four times, on three different geometries, and they disagree") is the sibling symptom, not a duplicate. Both stem from having no clean boundary between *what the variant is* and *how it is spelled*: #1749 is what that costs the overlap predicate, this is what it costs the derivation pipeline. #1749's symptom 2 — `repeat` classified three different ways by three `matches!` ladders, `dup` gated on `uncertain_extent` in one and not the other — is a spelling question being asked of stages that should only ever see semantics.

**This issue does not fix #1749**, and a render stage would not by itself reconcile the four deciders. It would reduce the number of edit kinds those deciders have to have an opinion about.

nh13:
Increments 0–4 of #1946. Three corpus families, the canonical-spelling seam landed as a strict no-op, and a `pub(crate)` widening. No output moves.

Six commits, best read in order:

| | commit | what it adds |
|---|---|---|
| 1 | `test(corpus): feed a repeat beside a sibling, which was a structural zero` | `repeat_beside_a_sequence_sibling`, 288 rows |
| 2 | `test(corpus): feed an authored repeat that keeps its sibling` | `authored_repeat_beside_a_sibling`, 240 rows |
| 3 | `refactor(normalize): land the canonical-spelling seam as a strict no-op` | `src/normalize/render.rs`, empty body |
| 4 | `test(corpus): reach a tandem tract longer than a normalizer window` | `long_tract_window_provenance`, 40 rows |
| 5 | `refactor(normalize): widen the helpers a render-time mint needs` | visibility only |
| 6 | `test(corpus): record why two repeat-beside-a-sibling families both earn a place` | the reconciliation with #1752, and every figure below re-measured at this base |

## What was blind

`the_corpus_is_closed_over_its_output_vocabulary` closes the corpus over the *constructs* it emits, and `repeat_expansion` (#1459) is what closed the repeat row of that table. It closes nothing about **geometry**: every one of that family's 252 inputs is a lone member, so the corpus asked what normalization does to a repeat with no sibling and never what it does to one with a sibling.

That distinction is the entire subject of #1946. Four passes in `src/normalize/merge.rs` exist only to undo a repeat or a duplication that a sibling-blind stage minted — `lower_repeat_edits`, `demote_repeats_spanning_siblings`, `demote_coincident_tract_repeats`, `respell_colliding_duplications` — and every one of them is reached only when a repeat and a sibling are in the same allele.

## The measured zero, and the correction this rebase forced

**Read this before the rest.** An earlier revision of this description led with a `0`, and that `0` is not true at this PR's base. The correction is worth more than the original claim.

Measured on `f7d177b5`, over the then-85,642-row corpus, counting rows whose input or output is a multi-member allele **and** carries repeat notation:

| | rows |
|---|---:|
| repeat beside a sibling, in the **output** | 2,098 |
| repeat beside a sibling, in the **input** | 0 |

Re-measured on `5567412f`, and byte-for-byte identical again on `d3eb64a2` after a mid-flight rebase — the two bases produce the same corpus dump, so these are not knife-edge readings:

| | at the base | with this PR |
|---|---:|---:|
| corpus rows | **95,614** | **96,182** |
| shape families | **23** | **26** |
| repeat beside a sibling, in the **output** | 2,410 | 2,518 |
| repeat beside a sibling, in the **input** | **756** | 1,140 |

All 756 come from `repeat_beside_a_sibling`, which #1752 added under #1749 while this branch was open. **So the structural zero this branch was opened to close had already been closed before its base.**

That does not make this coverage redundant, and commit 6 measures why rather than asserting it. The difference is where the sibling sits: #1752's family places one that *intersects* the tract or its junction, this one places a substitution two bases **outside** it. Reach does not separate them — ablating `lower_repeat_edits` moves 34 of that family's 756 rows and 24 of this one's 240 — but `--verify-spdi` does. That family's mismatching rows have inputs whose members already overlap (`c.[8_13ATA[3];12_15del]`), so they are malformed in and malformed out. This family's `g.` rows are disjoint in: the tract is `257_262` and the substitution is at `264`, two bases clear. Normalization *creates* the overlap. Clean in, malformed out is the worse defect and the one #1983 records; deleting this family takes the reproducer with it.

### A wrong instrument reproduced the wrong number

Worth recording, because it looked exactly like confirmation. The first re-measurement of the headline zero returned `0` again. It found the allele body with a regex bounded by the first `]`, and on

```text
NC_TEST.1:g.[257_262TTT[3];261_264del]
```

that `]` is the one closing `TTT[3]`, not the one closing the allele — so the capture holds no `;`, the row is not counted, and the instrument reports `0` on all 756 rows. A broken measurement agreeing with a broken claim is not agreement.

The committed guard never had this bug and does not acquire it now. `the_corpus_feeds_a_repeat_beside_a_sibling` tests `contains(';')` plus a repeat token, parses nothing, and asserts the **property** — that the corpus mints one *and* feeds one, with the minted side asserted non-zero first so `0 of 0` cannot pass as a result. It pins neither count. Both counts are recorded in its doc comment with the revision each was taken on.

## The families

`repeat_beside_a_sequence_sibling` — **288 rows**. It brings its own designed references, for the reason `separated_revcomp_runs` and `long_block_inversion` do. A tandem tract of a chosen unit, at a chosen offset, **bounded on both sides so it cannot extend**, is a property of the sequence. `first_tandem_triplet` finds a tract in only 21 of 48 random cores, and several of those are homopolymers whose real tract runs past `t+5` — which would silently make "inside the tract" and "beside it" the same place.

8 designs (units `ACG`/`AAT` × reference copy counts 3/4 × offsets 1/5) crossed with the existing six axis/direction cells and six shapes. Each shape names one pass of the tax, and two are a matched **authored/minted pair**, so a change that moves *when* a form is chosen moves one member of the pair while a change that moves the choice itself moves both. Sample (`tract_ACGx3_at5`, `g.`, 3prime):

```text
g.[262_270ACG[4];272A>C]         -> g.[270_271insA;272delinsGCC]   lower_repeat_edits (authored)
g.[262_270ACG[4];263C>A]         -> g.262_263insAGA                demote_repeats_spanning_siblings (authored)
g.[262_270ACG[2];263C>A]         -> g.[263_264del;266del]          ...its RepeatSource::Removed arm
g.[270_271insACG;263C>A]         -> g.262_263insAGA                same geometry, ferro-minted
g.[261_262insACG;270_271insACG]  -> g.262_270ACG[5]                demote_coincident_tract_repeats
g.[262_270dup;263C>A]            -> g.262_263insAGACGACGA          respell_colliding_duplications
```

Units are three bases because `repeated.md:21` admits a repeat on a `c.` reference only for units whose length is a multiple of three — it calls `c.2686A[10]` out by name. A shorter unit would make a third of the family spec-invalid on two of its three axes, and an unparseable input is recorded as a *row*, not as a failure, so the corpus would keep looking populated while measuring nothing. `no_tract_input_is_rejected` pins that.

`authored_repeat_beside_a_sibling` — **240 rows**. The family above builds inputs that author a repeat inside a cis allele, and **not one reaches `lower_repeat_edits`**. Instrumented, 52 get as far as `merge::canonicalize_from_sequence` and every one arrives with `variants.len() == 1`: the allele has already collapsed to a single member, so the lone-repeat lockout refuses it. That lockout is load-bearing — ablating it moves 7 rows, all `repeat_expansion`, exactly as its comment predicts. The gating property is **member survival**, not unit length and not sibling placement; both were tested and ruled out. This family is that shape, drawn on the random cores because that is where it demonstrably works — `lower_repeat_edits` fires 2,448 times without it and 2,484 with it. `the_authored_repeat_family_covers_the_cores_with_room` asserts both 21 (cores with a tandem triplet) and 20 (cores with room for the sibling), so the dropped core is accounted for.

`long_tract_window_provenance` — **40 rows**. The three repeat-minting rules are handed a `ref_seq` whose provenance differs by axis: the entire spliced transcript on `c.`/`r.`/`n.`, a growing window on `g.`. Tract discovery walks outward bounded only by `ref_seq.len()`. Whether that mattered was unmeasurable, and blind in **both** directions: on the transcript axes `ref_seq.len()` was 20 — the whole synthetic transcript — and on the genomic axis 0 of 3,361 insertion firings had a tract touching the window edge, so the growing-window loop never fired for a tract at all. The reassuring half was equally corpus-shaped: the furthest any tract walk reached was 11 bases, a property of 20-base cores rather than of the algorithm. Five designs, tract lengths 96–300 bases in an 800-base core. With them in place the genomic window grew to 401 and 801 bases, which is the first time in this corpus that loop has fired for a tract. `the_long_tracts_bracket_the_normalizer_window` reads `window_size` off `NormalizeConfig::default()` rather than restating it.

## The seam

`render_canonical_spelling` is landed doing **nothing at all**, between `detect_overlap_conflicts` and `sort_cis_members_by_genomic_order`. The point is that the placement and its argument are reviewable apart from any behaviour: when a form does start moving, the diff shows only the form.

The placement is **pinned, not chosen** — after `detect_overlap_conflicts` because a tandem-prefix split emits two pieces sharing a junction and would otherwise manufacture the very conflict that detector reports; before `sort_cis_members_by_genomic_order` because new pieces must be ordered; and after `emitted_verbatim` is computed, because that flag asks whether the allele came through the *derivation* unchanged and gates the #1406 rule, so a presentational re-spelling must not be able to flip it.

The invariant is **apply-equality, not member-count stability**. The stage may split one member into two, and `[dup;ins]` and one long `ins` denote identical bases. `confluence-gate-is-apply-equality-on-every-determined-axis` makes the release gate a relation over `apply`, whose codomain is bases, so it survives a member-count change by construction.

**On what the spec requires**, stated because the obvious reading is wrong: `DNA/duplication.md:18` does **not** require a tandem-prefix split. `duplication-must-ranks-the-label-not-the-partition` holds that it ranks a *label* and never requires partitioning so as to produce a duplication — which is exactly what such a split is. `:17` permits the label once the piece exists and `:90` publishes the geometry, so the split is legal and idiomatic but **ours to choose**, pending a ruling record.

The test asserts the **invariant**, not the no-op, so it survives the stage gaining behaviour rather than having to be deleted. It requires `Agree`, never merely the absence of `Differ` — and that distinction was found by sabotage, not by reasoning: an earlier revision asserting the weaker form was **passed by a stage body mutated to drop a member**, because `compare_denoted_sequences` declines by default and the decline read as a pass. Both sabotages (drop a member; change the denoted bases) now go red, verified.

## The widening, and why it is its own commit

Commit 5 is visibility-only: `cis_axis_parts`, `member_span`, `cis_kind_of`, `fetch_canonical_window`, `axis_frame`, `region_sequence_delta` and `AxisFrame` in `merge.rs`, plus `CisKind`, `MemberSpan` and `CodonGate`, go from module-private to `pub(crate)`. No behaviour, no new code. `CisKind` and `MemberSpan` are widened because `private_interfaces` fires otherwise — a `pub(crate)` fn naming a private type is a warning `-D warnings` turns into a build failure.

Landing the surface alone means the next step — a gate proving a render-time reference matches what each existing call site was given — is reviewable as a measurement rather than as a diff that also moves an API. That gate is a **separate, stacked PR** and is not in this one.

## The word table

Adding the third family took `FAMILIES` past the end of `number_word`'s hardcoded word table, so `the_header_states_the_family_counts_this_file_actually_has` died with `no word for 21; extend WORDS in number_word` — a panic naming a helper, raised by a test whose job is to name *stale prose sites*. The table's end was a second, invisible boundary sitting beside the real one, and extending it by four entries would only have moved the cliff.

It composes tens and units instead. The prose counts themselves stay stated and stay guarded: a count-free header is available only by deleting the claims, and the claims are exactly what #1947 showed go stale unwatched. So the boundary worth removing was the helper's, not the test's.

## Verification

Run at base `d3eb64a2`, with `FERRO_REQUIRE_BULK_FIXTURES=1` and **zero `Skipping:` lines** in the log, so the ClinVar, CMRG and Paraphase suites were armed rather than skipping green:

| gate | result |
|---|---|
| `cargo nextest run --features dev --no-fail-fast` | **11,226 run, 11,226 passed, 0 failed** |
| `cargo clippy --all-features --all-targets -- -D warnings` | clean |
| `pytest tests/python/` | **1,080 passed, 8 skipped** — run because `src/normalize/` is touched, which `cargo nextest` does not reach |

**Append-only is proven rather than argued.** Strip the three added families from the candidate dump and it is `diff`-identical to the base's own dump, over all **95,614** shared rows. That is also the seam commit's no-op, re-established at this base rather than carried from the last one.

`--verify-spdi` over the candidate corpus: `verified 96182 rows: 13 denote different bases, 67704 agree, 28465 unverifiable`. The 13 break down as **6 `adjacent_junction_ins` + 2 `repeat_beside_a_sibling`**, both pre-existing on the base and neither introduced here, and **5 `authored_repeat_beside_a_sibling`**, which are this branch's and are the point: those are the clean-in/malformed-out rows #1983 records, and they exist because this family builds a geometry nothing else does.

## Why not `Closes #1946`

#1946 is a multi-increment refactor and this is increments 0–4 — the measurement instrument and the empty seam, not the change. Closing the issue here would retire it before any form has moved.

Representation-Change: none. 0 of 96,182 corpus rows move. Three commits add corpus rows and guards under `examples/`; one adds `src/normalize/render.rs` with an empty body and one call to it; one widens ten `merge.rs` items to `pub(crate)` and changes no behaviour. Proven rather than asserted: stripping the three added families from the candidate dump leaves it `diff`-identical to the baseline over all 95,614 shared rows.

### DIFF
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -1145,7 +1145,7 @@ fn tract_designs() -> Vec<TractDesign> {
                 );
                 designs.push(TractDesign {
                     label: format!("tract_{unit}x{copies}_at{start}"),
-                    core: tract_core(unit, copies, start),
+                    core: tract_core(unit, copies, start, TRACT_CORE_LEN),
                     unit,
                     copies,
                     start,
@@ -1171,8 +1171,8 @@ fn tract_designs() -> Vec<TractDesign> {
 /// therefore forced away from the tract's periodic extension —
 /// `the_designed_tracts_are_exactly_as_long_as_they_claim` re-derives that from the
 /// finished string rather than trusting this construction.
-fn tract_core(unit: &str, copies: usize, start: usize) -> String {
-    let mut core: Vec<u8> = (0..TRACT_CORE_LEN).map(|i| b"GCTA"[i % 4]).collect();
+fn tract_core(unit: &str, copies: usize, start: usize, core_len: usize) -> String {
+    let mut core: Vec<u8> = (0..core_len).map(|i| b"GCTA"[i % 4]).collect();
     let tract = unit.repeat(copies);
     core[start..start + tract.len()].copy_from_slice(tract.as_bytes());
     let unit_bytes = unit.as_bytes();
@@ -1180,7 +1180,7 @@ fn tract_core(unit: &str, copies: usize, start: usize) -> String {
     core[start - 1] = flank_base(&core, start - 1, unit_bytes[unit_bytes.len() - 1]);
     // 3' boundary: its forward extension is the unit's first base.
     let after = start + tract.len();
-    if after < TRACT_CORE_LEN {
+    if after < core_len {
         core[after] = flank_base(&core, after, unit_bytes[0]);
     }
     String::from_utf8(core).expect("GCTA and the designed units are valid UTF-8")
@@ -1202,7 +1202,7 @@ fn flank_base(core: &[u8], at: usize, forbidden: u8) -> u8 {
             candidate != forbidden
                 && Some(candidate) != neighbour
                 && !(at == 0 && candidate == b'A')
-                && !(at == TRACT_CORE_LEN - 1 && candidate == b'T')
+                && !(at == core.len() - 1 && candidate == b'T')
         })
         .expect("four candidates cannot all be excluded by three constraints")
 }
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

PR #1976, opened 2026-08-15
test(corpus): derive the corpus header's family counts instead of restating them

### LINKED ISSUE #1947: dump_normalized_corpus.rs's own header is stale in three ways: family counts (says 15/13/9, is 21/14/14), corpus totals (says two, has three), and MAX_SPLIT_BLOCK

nh13:
Closes #1947.

`examples/dump_normalized_corpus.rs` is where a reader decides whether their change is measurable, and its header stated family counts the file does not have — all of them **understating** the corpus, which is the direction that makes a reader trust a measurement more than they should.

## The guard has now caught a real drift, on its first contact with a moving `main`

This is the part worth reading, and it happened during review rather than being argued for in advance.

The branch was rebased onto a `main` that had gained **#1752's `repeat_beside_a_sibling`** — a 15th family, landed the day after this guard was written. The guard went red immediately, derived the new counts, and named all seven stale sites with their replacement text:

```
this file's prose states family counts it does not have.
derived: FAMILIES = 15 (fifteen), designed-reference singletons = 2, DNA total = 17 (seventeen), PROTEIN_FAMILIES = 5
stale sites:
  "fourteen of the sixteen families" — expected "fifteen of the seventeen families"   (x5)
  "fourteen families there"          — expected "fifteen families there"
  "fourteen sequence-keyed families" — expected "fifteen sequence-keyed families"
```

The sites in this PR now read *fifteen* and *seventeen* **because the guard said so**, not because anyone recounted. Without it, `main` would have gone stale again within a day — which is precisely the repeat this change exists to prevent, arriving on schedule.

## What was wrong originally

| claim | sites | actual |
|---|---:|---|
| sequence-keyed families = 13, DNA total = 15 | 5 | **14** and **16** (now 15 and 17) |
| "the nine families there" | 1 | **14** (now 15) |
| "the thirteen sequence-keyed families" | 1 | **14** (now 15) |

Derived rather than taken from the issue: `FAMILIES` and `PROTEIN_FAMILIES` are counted from source, and `LONG_FAMILY` + `REVCOMP_FAMILY` are two singletons that bring their own designed reference, so the DNA total is `FAMILIES + 2`.

The `MAX_SPLIT_BLOCK` item in the issue (1024 → 4096) was already closed by #1938 and is untouched here.

## Why this is a guard and not seven corrections

Correcting the sites fixes today and guarantees a repeat: the counts live in a `const` and the prose restated them. That is exactly the shape `CLAUDE.md` names under *"Assert the property. Measure the count. Never let a count BE the property"* — **a count restated instead of imported**.

So the counts are now **derived**, in a test that reads this generator's own comment text and compares it against `FAMILIES.len()` and the singleton `const … _FAMILY` declarations counted out of the source. `PROTEIN_FAMILIES` is deliberately **not** anchored — the header states no protein-family count, so there is no phrasing to key on, and inventing one would be writing the claim in order to check it. A second test, `the_header_states_no_protein_family_count_to_anchor`, asserts that absence so the gap fails loudly if someone later writes one in.

**The scan now says whether it scanned.** All three anchors are `match_indices` loops, and two of them hang on a single sentence each, so a reword or an 80-column rewrap drops an anchor to zero sites and the test passes having checked nothing. The per-anchor counts are asserted non-zero. Checked against a sabotage in both directions: rewording the one " … families there" sentence turns the fixed guard red with `(5/0/1 sites)`, and leaves the pre-fix guard **green**.

**Verified by sabotage as well as by the live catch** — appending one entry to `FAMILIES` turns it red and names every stale site.

## Two sibling inventories in the same header, fixed by REMOVING the claim

Neither can be imported from a `const`, so the honest fix is to stop asserting them rather than to restate them.

**The corpus-totals paragraph** asserted there were exactly **THREE** totals and marked one *"the corpus **now**"*. #1752 made it four and moved "now" to `86,398`. It had **already been corrected once by this very PR**, from "two" to "three" — and went stale again the next day. Counting the entries and naming a current one are both claims that go stale on a change with no reason to touch this comment.

It is now an **anchored history**: each total tied to the change that produced it (`78,028` → `78,298` → `85,642` → `86,398`, with #1752's two-step `+378 / +378` explaining why `86,020` appears in that PR's own disclosure as superseded), stating no count, naming no current entry, and pointing at the command that answers *"what is it today"*.

**The blindness chain** said *"the four instances"* when there are at least seven — sequence structure (#1517) and reversed ranges (#1917) are both missing from the list it gives, and #1752 added a seventh from inside this very file (its family first shipped able to build only *growing* repeats). The count is **removed rather than bumped**, pointing at `CLAUDE.md`'s maintained list. A count of blindnesses goes stale every time the thing it counts happens.

## Two things about the guard worth knowing

**It scans comment lines only.** On its first draft it read its own string literals as stale sites and failed on itself.

**Its own prose gives the stale phrasings as digits, on purpose.** Spelling them out would make that comment the guard's first failure. `claude_md_adjudication_tables` hit the identical hazard and records the identical answer: keep the prose clear of the pattern rather than exempting the file, which would blind the scan to it permanently. That is noted inline so the next editor does not "improve" the wording back into a failure.

## Verification

Rebased onto current `origin/main`. `cargo fmt --all --check` and `cargo clippy --all-features --all-targets -- -D warnings` clean. The guard lives in an `[[example]]` target, which `--lib --test it` does **not** build, so it is selected explicitly: `cargo nextest run --features dev --example dump_normalized_corpus` — **45 passed, 2 skipped**, re-run against the final file rather than inferred from an earlier run that overlapped the edits.

Full gate on the pre-rebase head with the reference **armed**: **10,970 passed**, zero `Skipping:` lines, Python **1,048 passed**.

For the record, since it cost time twice: an earlier run showed four reference-aware failures that were purely environmental, and were confirmed so three ways — they pass with `FERRO_MANIFEST` unset (0.013–0.023 s, the stand-down signature), fail only when it is set to a path that does not exist, and pass in 7–8 s when armed against a present manifest. A manifest variable pointing at a missing file is strictly worse than no variable at all, because these suites skip cleanly on absence and fail on an unreadable path.

Representation-Change: none. Comments in an example generator plus one new test; no `src/` file is touched and no description ferro emits can change.

### DIFF
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -84,14 +84,47 @@
 //!    brings its own designed references because a random core cannot be asked to
 //!    have a coincidence pattern.
 //!
-//!    **Two corpus totals appear in this file and both are correct — do not
-//!    "reconcile" them.** `78,028` is the corpus *before* `separated_revcomp_runs`
-//!    existed, which is the only denominator a measurement taken before that
-//!    family was added can honestly carry; `78,298` is the corpus now, and
-//!    `78,298 - 78,028 = 270` is exactly that family's row count. They differ by a
-//!    digit transposition as well as by 270, which is a coincidence and has
-//!    already been read once as a typo in one of them. When you quote a figure,
-//!    quote the denominator it was measured against.
+//!    **Several corpus totals appear in this file and every one of them is
+//!    correct — do not "reconcile" them.** Each is the corpus as it stood when
+//!    some measurement was taken, so each is anchored to what changed it:
+//!
+//!    - `78,028` — before `separated_revcomp_runs` existed, and the only
+//!      denominator a measurement taken before that family was added can honestly
+//!      carry;
+//!    - `78,298` — after it, and `78,298 - 78,028 = 270` is exactly that family's
+//!      row count. The two differ by a digit transposition as well as by 270,
+//!      which is a coincidence and has already been read once as a typo in one of
+//!      them;
+//!    - `85,642` — after #1606 added the protein axis. It is the denominator of
+//!      item 6 below, and stays attached to that measurement;
+//!    - `86,398` — after #1752 added `repeat_beside_a_sibling` (756 rows, in two
+//!      steps: 378 when the family landed, 378 more when it was made to emit
+//!      shrinking repeats as well as growing ones, which is why `86,020` appears
+//!      in that PR's own disclosure as a superseded figure).
+//!
+//!    When you quote a figure, quote the denominator it was measured against.
+//!
+//!    **This list is a history, not a census, and deliberately does not say which
+//!    entry is current** — the current total is whatever the generator prints
+//!    today, and asking the file is the wrong way to find out:
+//!
+//!    ```text
+//!    cargo run --release --features dev --example dump_normalized_corpus -- --out /tmp/now.tsv
+//!    ```
+//!
+//!    That framing is the #1947 correction, and it is the second one this
+//!    paragraph has needed. It first said there were exactly **two** totals while
+//!    item 6 eight lines below already quoted the third — the more misleading
+//!    half of that error, since a reader who trusts an explicit "there are
+//!    exactly two, here they are" has no reason to look for a third. Corrected to
+//!    three, it was stale again **the next day**, when #1752 added a family: the
+//!    entry marked "the corpus **now**" had stopped being now, and nothing about
+//!    it said so. Counting the entries and naming a current one are both claims
+//!    that go stale on a change that has no reason to touch this comment, which
+//!    is the same defect as the family counts below — except that a corpus total
+//!    cannot be derived at compile time, so the fix here is to stop making the
+//!    claim rather than to import it. The *method* this paragraph states was
+//!    right throughout; its own bookkeeping is what kept drifting.
 //!
 //! 6. **…and only on the molecule types it builds** (#1606). Every family above is
 //!    spelled from nucleotides, so until the protein axis was added this corpus
@@ -102,7 +135,7 @@
 //!    place the same change measures **1,584 of 85,642 rows, all in
 //!    `protein_equal_length_delins` and none anywhere else**.
 //!
-//!    Note what the four instances have in common and why the list keeps growing:
+//!    Note what these instances have in common and why the list keeps growing:
 //!    fixing one blindness does not reveal the next. Geometry (#1456) was only
 //!    findable once the conflict families existed, scale (#1460) once geometry was
 //!    fixed, transcript geometry (#1478) once scale was — and molecule type was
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1977, opened 2026-08-15
perf(ci): build the soak archive from a driver package, not the whole `it` crate

### LINKED ISSUE #1964: ci: the soak archive optimizes 6,217 tests at opt-level 3 to run 45 of them
Follow-up to #1963, which establishes that the soak archive build (161s of compile) is the workflow's critical path and that no caching route reaches it.

## The waste

`test-build-soak` compiles the whole `it` crate — 139k lines, **6,217 tests** — at `opt-level = 3`, so that `soak`, `sweeps` and `censuses` can run about **45** of them. Every other test in that binary is optimized and then never executed by any consumer of that archive.

Measured split, in `user` CPU seconds (wall clock is not usable on this machine — see #1963 for the correction that cost):

| touched | rebuilds | user CPU |
|---|---|---:|
| `tests/it/main.rs` | the `it` binary | **251.9s** |
| `src/lib.rs` | the lib, then `it` | **368.6s** |

So the library is ~117s (32%) and the test driver ~252s (**68%**) of the soak compile. The runtime the profile exists to buy comes from optimizing the **library**; the driver is loops and `format!`.

## Why it cannot be fixed with a profile override

Cargo profile overrides are per-**package**, and `tests/it` is a *target* of the `ferro-hgvs` package. There is no spelling of `[profile.soak.package.…]` that separates a package's library from its own integration test. Lowering `[profile.soak] opt-level` wholesale is the blunt version and cuts the 32% you want to keep along with the 68% you do not.

## Why the split is tractable — scope it to what the soak archive actually runs

`tests/it` as a whole is genuinely entangled: **22** files use `CARGO_BIN_EXE_*` (which cargo sets only for tests in the same package as the binary) and **37** use `CARGO_MANIFEST_DIR`. That is the number that makes this look like a 428-module refactor, and it is the wrong unit.

The modules the soak archive's consumers actually select are exactly the clean ones. All seven use **zero** of either:

| module | consumer | `CARGO_BIN_EXE` | `CARGO_MANIFEST_DIR` |
|---|---|---:|---:|
| `cis_junction_crossing_shift` | `sweeps` | 0 | 0 |
| `repeat_span_sibling_overlap` | `sweeps` | 0 | 0 |
| `issue_1254_sibling_crossing_shift` | `sweeps` | 0 | 0 |
| `issue_1615_denoted_sequence_oracle` | `sweeps` | 0 | 0 |
| `issue_1542_direction_symmetry` | `censuses` | 0 | 0 |
| `normalize_axis_preserving` | `censuses` | 0 | 0 |
| `spec_conformance_axis` | `censuses` | 0 | 0 |

Their transitive `common/` closure is four helpers — `bulk_fixtures`, `cis_apply_oracle`, `spec_fixture`, `synthetic` — of which only `spec_fixture` reaches further, into `fixture_gen`. **`fixture_gen` is the entire porting cost**: 3 real `CARGO_MANIFEST_DIR` uses, all in `fixture_path`, which roots a relative path at the package dir. Its four `CARGO_BIN_EXE` hits are *prose in comments* explaining why it deliberately uses `env!("CARGO")` and a nested `cargo run` instead — and `env!("CARGO")` is available to any package.

The `soak` job additionally runs the proptest modules; those need the same check before being included or left behind.

## What it buys

A new workspace member holding those 7 modules + 5 helpers makes the override expressible:

```toml
[profile.soak]                     opt-level = 0   # the driver
[profile.soak.package.ferro-hgvs]  opt-level = 3   # the code under test
```

and `cargo nextest archive --cargo-profile soak -p <newpkg>` stops compiling the 139k-line `it` crate at `opt-level = 3` altogether. On the 32/68 split that is roughly **90-100s off a 192s gate, with no runtime loss** — the code under test stays at `opt-level = 3`.

## What needs care

- **`issue_1615_denoted_sequence_oracle` is deliberately run twice** — unarmed in `test`, armed in `sweeps` (`ci.yml` records why). A split must preserve both, not silently pick one.
- **Four guards scan `tests/it/` by directory**: `sweep_filter_invariant`, `oracle_exclude_invariant`, `census_filter_invariant`, `generator_completeness`. Each would need to know about the new location, and each fails in the flattering direction if it silently stops seeing a module.
- **`SWEEP_FILTER` and `CENSUS_FILTER` become package boundaries.** Post-#1960 these modules are already exclusively owned by the soak-side jobs, so the filters could be replaced by `-p <newpkg>` — which is a stronger guarantee than a name filter, since a rename cannot silently drop a module from a package.
- The `sweeps` and `censuses` jobs' `-E` selections, and `test`/`test-oracle`'s negations, all move with it.

## Not yet measured

The **runtime** cost of the driver at `opt-level = 0`. The claim above is that the driver is loops and `format!` while the work is in the library, but `cis_junction_crossing_shift` is a six-deep nest over ~276,000 cases and its loop overhead is not obviously free. Measure `sweeps` and `censuses` runtime on the split before committing to `opt-level = 0` for the driver; `1` or `2` may be the right setting.

nh13:
Closes #1964.

#1960 has merged and this is rebased onto `main`; `git log --oneline origin/main..HEAD` is one commit.

## The problem, measured

`test-build-soak` compiled the whole 139k-line `it` crate — 6,217 tests — at `opt-level = 3` so that its three consumers (`soak`, `sweeps`, `censuses`) could run 88 of them.

All figures are `user` CPU seconds on an M2 Max (12 core). Wall clock on that machine is not reproducible — several agent sessions run concurrently — and the same run showed 71–109s of wall for a stable ~395s of CPU, so wall is not quoted anywhere below. Dependencies are warm and the workspace artifacts are scrubbed before every rep, which is what `rust-cache` does in this job (it caches dependencies only and deletes workspace artifacts, so CI pays a full `ferro-hgvs` + driver compile every run). sccache is disabled for every measurement: this machine routes rustc through it globally, and the compile then happens in the sccache server rather than in a child, so `/usr/bin/time` under-reports by ~40× — the first pass at this measurement read 6.19s for a build that really cost ~250s.

| | build the soak archive |
|---|---:|
| before (`--test it`) | **394.8** (402.05 / 387.54) |
| after (`-p ferro-hgvs-soak-tests`) | **105.7** (median of 5: 105.99 / 105.67 / 104.87 / 105.81 / 105.12) |
| `libferro_hgvs` alone, same profile | 79.2 – 101.5 |

A **73% reduction**, and what is left is essentially the code under test: the library alone accounts for most of the 105.7s, which is irreducible — it is the thing `opt-level = 3` exists to optimize.

Runtime is unchanged. The sweeps selection, which is the largest and by far the most stable term, costs **139.8s** from the new archive against **140.3s** from the old one.

### On the real runner

Step-level, so job overhead (checkout, toolchain, nextest install, artifact upload) is not counted, and comparing only runs whose archive cache **missed** — a hit skips the build step entirely and reads as 0s, which is why two recent base-branch runs show 22s and 28s and are not comparable:

| | `Build soak archive` step |
|---|---:|
| base branch, run 31849010009 | 161s |
| this PR, run 31902502874 | **88s** |

A 45% cut on `test-build-soak`, which is the head of the critical path since `soak`, `sweeps` and `censuses` all `needs:` it. (An earlier revision of this branch measured 75s; the driver has since gained the fourth proptest module and the support module it reaches.) The three consumers are green on the settled run 31902588946: sweeps 77s, censuses 87s, soak shard 1/8 53s.

## Why it needs a package

Cargo profile overrides are keyed on the **package**, never on the target. That is verified rather than assumed — with `[profile.soak.package.rootpkg] opt-level = 3` and a profile default of 0, `cargo build -v` shows `opt-level=3` on the root package's lib **and** on its `it` test target, while a sibling member's test target gets the default. So while the driver was a target of `ferro-hgvs` there was no way to spell "code under test optimized, driver cheap".

## The `opt-level` measurement — a negative result

The issue asked for `[profile.soak.package.<driver>] opt-level = <low>`. **The measurement refuses it, and the override is not in this PR.** The rationale and the table are recorded on `[profile.soak]` in `Cargo.toml` so it is not re-derived.

| driver `opt-level` | build | run: sweeps | run: censuses |
|---:|---:|---:|---:|
| **3 (as shipped)** | **105.7** | **139.8** | **90.2** |
| 2 | 105.1 | 141.2 | 89.7 |
| 1 | 101.1 | 143.9 | 92.8 |
| 0 | 85.5 | **240.9** | 109.2 |

Builds are medians of 5 interleaved reps, runs the mean of 2. `opt-level = 0` buys ~20s of compile and gives back ~101s of run; 1 and 2 are a wash in both directions. The compile column is flat because of the *other* half of the change: the package compiles ~8k lines where `it` is 139k, so there is little left for a lower `opt-level` to save. **The split is what pays, not the `opt-level`.**

## The modules do not move

`tests-soak/tests/soak/main.rs` `#[path]`-includes the ten modules the three jobs select. The files stay under `tests/it/` and stay declared in `tests/it/main.rs`. That was chosen over relocating them because moving would have cost, silently in most cases:

* **Five guards scan by directory** — `sweep_filter_invariant`, `oracle_exclude_invariant`, `generated_fixture_ci_wiring`, `ruling_citation_currency`, `ruling_guard_field` — and each fails in the *flattering* direction: a module that leaves the scan simply stops being policed. `spec_conformance_axis.rs` carries 54 spec citations.
* **The ruling ledger cites test files by path.** `canonical-form-choice-when-both-legal` and `coding-axis-merges-are-a-disclosed-general-34-deviation` name guards in `tests/it/cis_junction_crossing_shift.rs` and `tests/it/spec_conformance_axis.rs`. Moving the files would have meant editing the adjudication ledger — and re-blessing `docs/NORMALIZATION_CONTRACT.md` — to chase a build-layout change.
* **`issue_1615_denoted_sequence_oracle` is run twice on purpose.** In place, it stays in `it` for the un-armed `test` run and is compiled here for the armed `sweeps` run, which is exactly what an in-place include gives for free.
* **The pinned proptest seeds keep replaying, unmoved.** `#[path]` leaves the `..` segments in `file!()` and the filesystem resolves them, so this binary resolves `tests-soak/tests/soak/../../../tests/proptest-regressions/<module>.txt` — the same committed file `it` reads. Verified empirically, by appending an unparsable line and reading back the path proptest named in its warning:

  ```
  proptest: /…/issue-1964/tests-soak/tests/soak/../../../tests/proptest-regressions/cis_allele_confluence_proptest.txt:51: unparsable line, ignoring
  ```

  Worth stating because had they moved, nothing would have gone red: those seeds are checked in **no** other job, since `test` and `test-oracle` both negate `test(proptest)`.

## What the indirection opens, and what closes it

Membership in the driver is a second list that can drift from `ci.yml`'s filters, and the drift is silent and flattering: a module named in `SWEEP_FILTER` or `CENSUS_FILTER`, or matching `test(proptest)`, is negated by `test` and `test-oracle` — so if it is not also compiled into the driver it runs in **no job**, and CI gets faster. `--no-tests=fail` catches a selection matching *nothing*, never one module of four going missing.

`tests/it/soak_package_membership.rs` is the guard, in both directions plus one:

1. every module the three jobs select must be compiled into the driver;
2. the driver must compile nothing they do not select (that is pure cost on the critical path);
3. no test **function** outside the driver may carry a name one of those `test(…)` substring terms matches — the hole the module-level reading would otherwise leave.

It reads `ci.yml` itself rather than sharing an extractor with the three guards that already do, for the reason `census_filter_invariant.rs` sets out at length: a bug in a shared extractor neuters all of them at once, and does it by returning an empty selection, which makes every assertion vacuous rather than red.

Guard (3) found a live instance on this base. `oracle_exclude_invariant::the_ci_oracle_selection_negates_proptest_the_sweeps_and_the_corpus_modules` carries the token `proptest` in its **name**, so `test(proptest)` selected it and `test`/`test-oracle` negated it — meaning an assertion about `ci.yml`'s shape was running only inside the 1M-case soak job, and after this change would have run nowhere. Renamed to `…negates_the_property_tests_…`, so it now runs in `test` and `test-oracle` where it belongs.

## Count-based partition proof

Test-name **sets**, not just counts, listed from the two real archives (`cargo nextest list --archive-file`, runnable tests, no `--features dev` — the archive is built without it):

| selection | before (`--test it`) | after (`-p ferro-hgvs-soak-tests`) |
|---|---:|---:|
| total in archive | 6174 | **121** |
| `soak`: `test(proptest)` | 29 | 29 |
| `sweeps`: `$SWEEP_FILTER + test(issue_1615_…)` | 54 | 54 |
| `censuses`: `$CENSUS_FILTER` | 13 | 13 |
| **selected total** | **96** | **96** |
| carried but selected by no job | 6078 | 25 |

Both columns are measured on this branch, so the packaging change is isolated: **every job selects exactly the same tests**, and the archive carrying them shrinks from 6174 to 121.

The 25 the driver carries that no job selects are the `#[cfg(test)]` units inside the `tests/it/common/` helpers it includes, plus `issue_1539_split_member_separation`'s own tests. Checked rather than asserted — every one of the 25 has a counterpart in the `it` binary, so none is stranded:

```
comm -23 <soak unselected> <it tests, common:: prefix normalised>   ->   (empty)
```

One test does change job, and only one: `oracle_exclude_invariant::the_ci_oracle_selection_negates_proptest_…` carried the token `proptest` in its **name**, so `test(proptest)` selected it and `test`/`test-oracle` negated it — an assertion about `ci.yml`'s shape was running only inside the 1M-case soak job. Renamed to `…negates_the_property_tests_…`, it now runs in `test` and `test-oracle`.

On the `it` side nothing left the binary, so the `test` / `test-oracle` selections are unaffected. With `--features dev`, `it` holds 6373 runnable tests and they partition exactly:

```
test job selection 6287 + proptest 29 + sweeps modules 44 + census modules 13 = 6373
```

`issue_1615_denoted_sequence_oracle`'s tests sit inside the 6287 — `test` does not negate that module — which is the deliberate double-run preserved.

## The fourth proptest module, and the one module compiled for it

`main` gained `minimal_alignment_enumeration_proptest` after this branch was cut. The `soak` job selects `test(proptest)`, so a driver holding only the original three would have deleted that module's 1M-case coverage on merge — silently, since `test`/`test-oracle` negate it. **`soak_package_membership` caught exactly that**, which is what it is for.

It needs two things the driver lacked: `common::minimal_alignment` (a helper, no `crate::` deps, included alongside the others) and `crate::issue_1539_split_member_separation::forced_unchanged_columns` — a `pub(crate)` fn in a full test module, used by `the_closed_form_and_the_enumerator_agree`.

**The tempting fix is to move that function beside its sibling in `common/minimal_alignment.rs`, and it is the wrong trade.** The `unchanged-is-read-over-every-minimal-alignment` ruling names the symbol **by path** as the detector the ruling came from — in its rationale, in the blessed `docs/NORMALIZATION_CONTRACT.md` rendered from it, and in the helper's own doc. Relocating it for a build-layout reason would falsify a `decided` record's prose in three places and force a contract re-bless, which is the class of cost this PR's whole design avoids. The two implementations are also kept independent on purpose so each checks the other.

So the module is compiled as **support**: no job selects it, and its own tests go on running from `it`. That needed the second membership test reframed rather than exempted — and the reframing is the interesting part:

> A compiled module no job selects must be named as `crate::<module>::` by another compiled module.

The exemption is **derived, not declared**. A marker comment in the driver would be a claim the test then trusts; requiring a real reference makes the driver *demonstrate* the need, and makes the exemption lapse by itself the moment the reference is deleted — so "support module" cannot become a place to park dead compile cost on the critical path. Verified by sabotage: parking an unreferenced `spdi_tests` in the driver fails the test by name.

## The one behavioural fix this surfaced

`bulk_fixtures::present_or_skip` now returns the **resolved** path instead of `Some(())`, and its four callers open that.

nextest sets a test binary's working directory to its own **package** root. Every caller did `present_or_skip(path)?` and then `File::open(path)` with the same cwd-relative string, which agreed only while that package root was the workspace root. From `ferro-hgvs-soak-tests` the cwd is `tests-soak/`, so `normalize_axis_preserving`'s three bulk-corpus tests would have resolved `tests/fixtures/bulk/…` under `tests-soak/`, found nothing, and — without `FERRO_REQUIRE_BULK_FIXTURES` — **skipped green**, deleting the ClinVar, CMRG and Paraphase axis coverage while the job got faster. Resolving inside `present_or_skip` makes presence and open agree by construction. `bulk_fixtures::an_existing_bulk_fixture_path_is_present` is what caught it, and now pins the resolution.

`issue_1539_split_member_separation` needed the same treatment for the same reason once it entered the driver: its two committed fixture paths (`tests/fixtures/split-member-separation/…`) were opened cwd-relative, so from `tests-soak/` four of its tests died on a missing file. They now resolve through `common::fixture_gen::fixture_path` too. That one fails loudly rather than skipping, but it is the same latent assumption.

## Also changed

* `nightly-soak.yml` builds the same package (it is the unseeded counterpart and archived `--test it` too).
* The `Clippy` job takes `--workspace`: `default-members` is the root package alone, so without it the new driver would be linted by nothing.
* `common::fixture_gen` resolves the workspace root by ascending to the `Cargo.lock` holder rather than trusting `CARGO_MANIFEST_DIR`, which names whichever package the file was compiled into. A unit test pins it, and runs once per including package.

## Verification

* `cargo fmt --all -- --check` — clean.
* `cargo clippy --workspace --features dev --all-targets -- -D warnings` — clean.
* `actionlint` and `zizmor` on both changed workflows — no findings.
* `cargo nextest run --features dev --no-fail-fast` — **11026 passed**, 0 failed (151 `#[ignore]`d).
* `cargo nextest run -p ferro-hgvs-soak-tests --no-fail-fast` — **121 passed**, 0 failed.
* `cargo build --profile soak -p ferro-hgvs-soak-tests --tests` — compiles. Note the `--tests`: this package has no lib and no bins, so a plain `cargo build -p` builds nothing and exits 0.
* The CI-reading invariants (`soak_package_membership`, `census_filter_invariant`, `oracle_exclude_invariant`, `sweep_filter_invariant`, `generator_completeness`) plus `fixture_gen`: 25 tests, all passing.
* `git status --porcelain tests/proptest-regressions/` clean after running the driver's proptests — no seed appended, no stray regressions directory under `tests-soak/`.

Representation-Change: none. No file under `src/` is touched at all — the change is a cargo package boundary, two workflow invocations and a test-harness path resolution, and the count-based partition above shows the same tests running with the same selections.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -593,7 +593,12 @@ jobs:
           # eviction.
           save-if: ${{ github.ref == 'refs/heads/main' }}
           key: clippy
-      - run: cargo clippy --features dev --all-targets -- -D warnings
+      # `--workspace` because `default-members` is the root package alone (see
+      # `Cargo.toml`), so without it the `ferro-hgvs-soak-tests` driver — the one
+      # `test-build-soak` compiles — is linted by nothing. It is the only member,
+      # and it compiles a dozen modules that `--features dev` does not reach here
+      # anyway, so the added cost is one small test target.
+      - run: cargo clippy --workspace --features dev --all-targets -- -D warnings
 
   clippy-all-features:
     # Lints the feature-gated code that `--features dev` never compiles —
@@ -964,9 +969,33 @@ jobs:
   # 0.355s under this one — 5.4x. The other jobs run each test once and would
   # only pay the longer compile.
   #
-  # Only the `it` binary is built (`--test it`): every test the soak selects
-  # with `-E 'test(proptest)'` lives there, so compiling the rest of the test
-  # targets at `opt-level = 3` would be pure cost.
+  # ONE PACKAGE IS BUILT, AND IT IS NOT `ferro-hgvs`'s `it` TARGET. This job used
+  # to build `--test it`, which compiles the whole 139k-line `it` crate — 6,217
+  # tests — at `opt-level = 3` so that about 45 of them can run. Measured on an
+  # M2 Max, `user` CPU with the workspace artifacts scrubbed the way `rust-cache`
+  # scrubs them here, deps warm: 394.8s for the `it` archive, of which the
+  # library is a minority. The rest was the driver, and the runtime
+  # `opt-level = 3` buys comes from optimizing the LIBRARY.
+  #
+  # Cargo profile overrides are keyed on the PACKAGE, never on the target, so
+  # `[profile.soak.package.ferro-hgvs]` reaches the lib, the bins and the `it`
+  # test target alike — there is no way to spell "library optimized, driver
+  # cheap" while both are one package. `ferro-hgvs-soak-tests` is that second
+  # package: `tests-soak/tests/soak/main.rs` `#[path]`-includes the modules the
+  # three consuming jobs select — plus, separately marked, any module one of
+  # those reaches into — while the modules themselves stay under `tests/it/` and
+  # stay declared in `tests/it/main.rs`.
+  #
+  # NOTE the package is NOT dropped to a lower `opt-level`, which is the half of
+  # #1964 the measurement refused; the table and the reasoning are on
+  # `[profile.soak]` in `Cargo.toml`. The saving is from compiling ~9k lines
+  # instead of 139k, not from optimizing them less.
+  #
+  # `tests/it/soak_package_membership.rs` polices both directions: a module CI
+  # selects from this archive that is not compiled into it would be negated by
+  # `test`/`test-oracle` and run nowhere while CI got faster, and a module
+  # compiled into it that nothing selects and nothing references is dead compile
+  # cost on the critical path.
   #
   # `debug-assertions` and `overflow-checks` stay ON in the profile (see
   # Cargo.toml) so this buys speed without quietly weakening what the soak
@@ -1047,11 +1076,12 @@ jobs:
         # No `--features dev`. That feature pulls in `web-service` (axum, tokio,
         # tower, tracing) and `benchmark` (reqwest, rusqlite, tokio-postgres,
         # futures) — an entire dependency tree compiled at `opt-level = 3` to
-        # produce one test binary that never touches any of it. `tests/it`
-        # gates only 2 of its ~360 modules on the feature and neither the soak
-        # nor the sweeps selects one, so the selections are unchanged: verified
-        # by listing both under each feature set, `test(proptest)` resolves to
-        # 20 tests either way and the sweep filter to 44.
+        # produce one test binary that never touches any of it. None of the ten
+        # modules this package compiles is gated on the feature and none of the
+        # three consuming jobs selects a gated one, so the selections are
+        # unchanged: verified by listing under each feature set,
+        # `test(proptest)` resolves to 20 tests either way and the sweep filter
+        # to 44.
         #
         # Both numbers have been stale before, in opposite directions and for
         # different reasons, which is why the re-measure commands are published
@@ -1064,10 +1094,10 @@ jobs:
         # "either way" — which held through both drifts and still holds; the
         # totals are incidental to it. Re-measure rather than trust them:
         #
-        #   cargo nextest list --test it -E 'test(proptest)' | wc -l
-        #   cargo nextest list --test it -E "$SWEEP_FILTER" | wc -l
+        #   cargo nextest list -p ferro-hgvs-soak-tests -E 'test(proptest)' | wc -l
+        #   cargo nextest list -p ferro-hgvs-soak-tests -E "$SWEEP_FILTER" | wc -l
         run: |
-          cargo nextest archive --cargo-profile soak --test it \
+          cargo nextest archive --cargo-profile soak -p ferro-hgvs-soak-tests \
             --archive-file nextest-soak.tar.zst
       - uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
         if: steps.soak-archive.outputs.cache-hit != 'true'
@@ -1489,10 +1519,13 @@ jobs:
     # Only the optimized archive, NOT `test-build`. Every test the selection
     # below reaches lives in `tests/it/{cis_allele_confluence,
     # from_sequences, minimal_alignment_enumeration,
-    # normalize_idempotency}_proptest.rs`, and none reads the generated spec
-    # fixture or shells out to the staged generators — so the shards need
-    # neither of those artifacts, and waiting on the job that produces them
-    # would put the dev-profile compile back on the soak's critical path.
+    # normalize_idempotency}_proptest.rs` — all four of which
+    # `tests-soak/tests/soak/main.rs` compiles, which is what puts them in this
+    # archive at all now that it holds one package rather than the whole `it`
+    # binary — and none reads the generated spec fixture or shells out to the
+    # staged generators — so the shards need neither of those artifacts, and
+    # waiting on the job that produces them would put the dev-profile compile
+    # back on the soak's critical path.
     #
     # `from_sequences_proptest.rs` joined this selection by being named
     # `*proptest*`, which is worth stating because the filter is by name: a new
@@ -1584,8 +1617,12 @@ jobs:
   #   repeat_span_sibling_overlap        40.44s -> 5.29s   7.6x
   #   issue_1254_sibling_crossing_shift  20.14s -> 2.32s   8.7x
   #
-  # They cost nothing extra to build: all three live in `tests/it/`, so they are
-  # already inside the archive `test-build-soak` produces for the soak.
+  # All three live in `tests/it/` and are `#[path]`-included by
+  # `tests-soak/tests/soak/main.rs`, so they are inside the archive
+  # `test-build-soak` produces alongside the soak's own modules — one archive for
+  # the three jobs, not one per job. Adding a fourth sweep means naming it there
+  # as well as in `SWEEP_FILTER`; `tests/it/soak_package_membership.rs` fails if
+  # only one of the two is done.
   #
   # Running them ONLY here, with both oracles on, loses no coverage. The oracles
   # add assertions at `normalize_core_checked`'s exit and change no behaviour, so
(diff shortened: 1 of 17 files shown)


nh13:
@coderabbitai pause

PR #1979, opened 2026-08-15
ci: fail loudly when a merge group builds a commit main will not accept

### LINKED ISSUE #1978: merge queue: four required contexts never report inside a merge group, ejecting a green PR three times
## Summary

PR #1965 has been added to the merge queue **three times** and ejected all three times without merging. Its own checks are 42 pass / 1 skipping / 0 fail throughout, and its merge-group CI runs **succeed**. The cause is four required contexts that never report inside the merge group.

## Measurement

Polling the live merge-group ref `gh-readonly-queue/main/pr-1965-b7828e1f…` every 45s and diffing the reported check-runs against the ruleset's ten required contexts. Stable across five consecutive polls spanning ~4 minutes, with 29 checks reported:

```
MISSING = [ abi3 floor ↔ requires-python | Python Lint | Test | zizmor + actionlint ]
```

The entry is then removed. Timings for the three attempts:

| added | removed | merge-group CI |
|---|---|---|
| 06:02:44 | 06:11:35 | success |
| 06:38:30 | 06:46:32 | success (06:38:48 → 06:46:06) |
| 06:53:51 | 07:02:16 | success |

The second attempt is the clearest: CI succeeded at **06:46:06** and the entry was removed at **06:46:32**, 26 seconds later, without merging. For contrast, #1783 was removed at 06:25:04 and **merged at 06:25:05** — removal immediately followed by the merge.

## What is ruled out

- **Not a check failure.** Every merge-group run concluded `success`.
- **Not `grouping_strategy: ALLGREEN` coupling.** The first ejection was alongside #1968, but #1968 re-enqueued alone merged in eight minutes, and #1965 re-enqueued alone was ejected twice more.
- **Not a base move.** `main` did not advance during the first two windows.
- **Not `check_response_timeout_minutes`.** That is 60; these eject at ~8.5 minutes.
- **Not a group rebuild.** The group ref name is identical across attempts.

## Where the cause probably is

Both `ci.yml` and `lint-workflows.yml` trigger on `merge_group` and both concluded `success`, so the four contexts are **jobs skipped inside a successful workflow** rather than workflows that never ran. Three of the four (`abi3 floor ↔ requires-python`, `Python Lint`, `Test`) come from `ci.yml`; `zizmor + actionlint` comes from `lint-workflows.yml`.

Worth noting `lint-workflows.yml` already documents this exact hazard in its own comments, as the reason it carries no `paths:` filter:

> a `paths:` filter that does not match means the workflow never runs, and a workflow that never runs reports no check at all — which is indistinguishable, to a required-status-check rule, from a check that simply has not finished.

The same reasoning applies one level down, to job-level `if:` conditions inside a workflow that does run.

Why this reproduces on #1965 and not on #1968 or #1783 — which merged through the same queue in the same hour — is not established. #1965's diff includes `tests/python/test_sequence_normalize.py` and `.pyi` stubs, which makes `Python Lint` being absent especially counterintuitive.

## Impact

Any PR that hits this is **permanently unmergeable** through the queue, while presenting as fully green. Nothing on the PR surfaces it: `gh pr checks` reads 42 pass / 1 skipping, and the ejection leaves no failure anywhere. It is only visible by polling the merge-group ref before it is torn down.

## Suggested next step

Compare the job-level conditions in `ci.yml` for the four contexts against what a `merge_group` event supplies — in particular any changed-files detection, which computes a different set for a merge group than for a pull request.

nh13:
Closes #1978.

### Summary

`main`'s ruleset carries `required_signatures`, and nothing in CI checked whether the commit the merge queue builds satisfies it. When it does not, the failure is silent in a way nothing on the pull request can surface: every required context reports green, the merge-group run concludes `success`, and ~25 seconds later the entry is removed without merging. The PR stays open, still reads fully green, and can be re-queued forever.

This adds a `merge_group`-only job that reads the merge-group commit's own signature verification and fails when it is not verified, listing any line of the commit message that reads as a git commit header.

### What #1978 actually was

The issue reported four required contexts never reporting inside the merge group. They do report. `GET /commits/{sha}/check-runs` defaults to `per_page=30` and the merge group carries 34 check runs, so the poll behind that measurement was reading a truncated first page; `Test` was additionally still running at the last poll. All ten required contexts report `success` (or `skipped`, for `Representation change declared`) in all three of #1965's merge-group runs.

The entry is removed 20–26 seconds after the last check completes, in all three attempts. That is the queue attempting the merge and the merge failing. Across the 40 merge-group commits built in the surrounding fourteen hours, exactly three are `verified: false` / `reason: "invalid"` — all three of them #1965's — and thirty-seven are valid.

The producer: the squash body is the pull request description, hard-wrapped, and GitHub writes its `gpgsig` header after the last line matching a `committer` prefix. #1965's description contains the phrase "only the `committer` differs", and the wrap put that word at column zero, so the signature header landed inside the message instead of in the header block.

### Scope

Not a required context. In every case it fires the entry is already unmergeable, so requiring it buys nothing; the value is a named red row where there was previously no failure anywhere to read. Making it required would also be a ruleset change, which this PR does not make.

There is no pull-request-side twin, and the workflow comment records why: the wrap is predictable enough to replicate, but `tree`, `author` and `parent` are ordinary words in this repository's PR prose, so a predictive guard would produce wrong reds on undocumented behaviour.

### Verification

`actionlint` and `zizmor` both clean. The guard script was exercised against real commits: it exits 1 on `c32eb06a` (#1965's third merge group), naming the offending message line, and exits 0 on `b7828e1f` (#1968's, which landed on `main`).

Representation-Change: none. Workflow only; no watched directory is touched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -361,6 +361,99 @@ jobs:
             --changed-files changed-files.txt \
             --declaration-file pr-body.txt
 
+  # The queue's own commit has to be a commit `main` will accept, and `main`'s ruleset
+  # carries `required_signatures`. Nothing in CI looks at that, and when it is violated the
+  # queue reports it NOWHERE: every required context goes green, the merge-group run
+  # concludes `success`, and ~25 seconds later the entry is removed without merging. The
+  # pull request stays open, still reads fully green, and can be re-queued forever. That is
+  # #1978 — one PR ejected three times before the cause could be named, because the
+  # ejection leaves no failure anywhere to read.
+  #
+  # This is the same family as the `paths:`/`merge_group:` notes in `lint-workflows.yml`
+  # and above, one step further along: those are about a check that never REPORTS, this is
+  # about a merge that never HAPPENS. In all three the symptom is silence, and silence is
+  # what a required-status-check rule cannot distinguish from anything else.
+  #
+  # What produced it, measured rather than reasoned: GitHub composes the squash body from
+  # the pull request description, hard-wraps it, and writes its own `gpgsig` header into
+  # the commit object after the LAST line matching `^committer `. #1965's description
+  # contained "...only the committer differs.", and the wrap put `committer` at the start
+  # of a line — so the signature header landed inside the message instead of in the header
+  # block and the commit came back `verified: false`, `reason: "invalid"`. Over the 40
+  # merge-group commits built in the surrounding fourteen hours: 3 invalid, all three that
+  # one pull request's, 37 valid.
+  #
+  # DELIBERATELY NOT A REQUIRED CONTEXT, and not because that would be tidier to add.
+  # Requiring it would buy nothing: in every case this fires the entry is already
+  # unmergeable, so the queue drops it either way. The whole value here is putting a named
+  # red row where there was silence. (Making it required is also a ruleset change, which is
+  # a separate decision from a workflow edit.)
+  #
+  # There is deliberately no pull-request-side twin, and the reason is NOT that the wrap
+  # cannot be predicted. It largely can: a 72-column wrap that does not split long tokens
+  # reproduces #1965's affected paragraph line-for-line, and the 84/100/96-column maxima
+  # measured on three other merged squashes are that same wrap meeting an unbreakable URL
+  # or code span. What stops it being a guard is the false-positive side. The hazard is a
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #1980, opened 2026-08-15
docs(rulings): re-ground the minimal-alignment rule as a house choice

nh13:
`rulings[unchanged-is-read-over-every-minimal-alignment]` named `general.md:34` as **governing**, which claimed the recommendations decide how a reference base is judged unchanged. They do not. This re-grounds it as a house choice. **The ruling is unchanged; only its authority is.**

Two commits. `576b6760` does the re-grounding. `2bd2f56f` then repairs the citations of the two records that scope `delins.md:47`, and closes the open question the first commit opens — see [Second commit](#second-commit--the-two-delins47-scopes-revisited).

## Why

`general.md:34` keys on variants "separated by one or more nucleotides", so applying it requires knowing which reference nucleotides are unchanged — and the recommendations never say. The clause raises the question and supplies no procedure for answering it.

Measured over `docs/recommendations/`, `docs/background/` and `docs/consultation/` at spec checkout `6f85311`:

| searched | hits |
|---|---:|
| `cost` | **0** |
| `parsimon*` | **0** |
| `edit distance` | **0** |
| `optimal` | **0** |
| `fewest` | **0** |
| `minimal` / `minimum` | 6, none about edit minimality — unsequenced-deletion extent (`uncertain.md:33`, `:159`) and SNP-array probe bounds (`DNA/deletion.md:100`, `DNA/duplication.md:106`, `RNA/insertion.md:62`, `DNA/insertion.md:106`) |

Of eight `align*` hits, the only one using an alignment to *derive a description* is `DNA/delins.md:46` — in the specification's own scare quotes, describing the form it declines in favour of the spanning `delins`.

So there is no definition of "unchanged" for a block differing at several positions, and no procedure of any kind for deriving a partition from two sequences. Naming `:34` as governing asserted a warrant the text does not carry.

## What changes

- `governing` is **dropped** (absent, not null).
- A `house_choice` is added under `README.md` **rule 5's silent limb** — whose own doc reads *"the recommendations are incomplete here"*, which is the measured finding rather than an interpretation.
- The rationale gains a section recording the re-grounding.

Status stays `decided` — a choice was made. All four clauses stay cited: clause-free means **no governing clause**, not no citations, and naming what was read and why none of it reaches is the substance of `considered_and_rejected`.

## What does NOT change

**No output moves.** #1939 measured shipped exposure at **nil** — 840 of 840 sampled cases already produce indel-bearing members under the default arm — so this changes what the ledger *claims*, not what ferro emits. Both guards are untouched and still pass.

`rulings[derivation-may-not-be-bounded-by-the-inputs-spelling]`'s correction (#1956), which uses this record as its falsifier, is unaffected: that correction establishes the position-wise ground is false, which remains true whatever grounds this reading.

## Two things the record now concedes

**The position-wise reading is rejected, but not because a clause forbids it.** It has no less textual support than the minimal-alignment reading, and no more. The stated ground is that it is not stable under what it describes: on an equal-length block it reports changed columns a cheaper alignment leaves untouched, so two spellings of one variant present different separations and `:34` returns two answers for one variant.

**The counter-evidence is recorded unresolved.** `DNA/delins.md:44-47` is the corpus's only worked confrontation with alignment-derived splitting and it prefers the **coarser** form; `DNA/delins.md:86-89` records permission for a split spelling being **withdrawn**. That is evidence against reading alignment-derived splitting as the spec's method, and this PR does not answer it. Two existing records currently scope `:47` away from these blocks — `delins-payload-coincidence-carve-out-is-coding-dna-scoped` to `c.`, `delins-merge-vs-individual-gap-two-or-more` to net-deletion — but that is a consequence of those scopes, not a refutation.

The second commit revisits both scopes and leaves them standing, on better-evidenced grounds than they carried.

## Second commit — the two `delins.md:47` scopes, revisited

The revisit was run as an independent review with the recommendation coming back **leave both scopes as they are**. What changes is the evidence each record cites, not what either decides. Three repairs:

**The axis record gains the corpus answer to its own conceded counterweight.** It concedes that `:47`'s second ground — simplicity — is axis-neutral, and that a reader weighing it equally "would scope the carve-out to every axis and would not be contradicting the spec's text". The corpus answers that: `consultation/SVD-WG010.md:27` makes the axis-neutral simplicity argument **verbatim** — *"making a discrimination between variants in protein-coding and non-coding sequences **is undesired** and makes application of the recommendations unnecessarily complex"* — inside a proposal whose own header reads `Status: rejected` (`:5`). Under this project's standing treatment of a rejected proposal (a negative guard, never an expectation, as `delins-codon-carve-out-gap-one` records), that is evidence **against** letting the simplicity limb carry the carve-out off the coding axis. Neither scoping record cited it.

The same repair records that **`delins.md:86-89` was examined and supplies no axis widening** — it is a coding-axis example (`NM_007294.3`), its RNA twin is on a coding transcript where the protein consequence `:47` names is live, and it sits at separation **zero**, which `general.md:34` and `delins.md:17` do not reach. It is already ruled on by `delins-adjacent-members-when-both-consume-reference`. The chronology is recorded because the natural suspicion is that the record predates the passage: it does not. `c0742311` (#1680) brought `:86-89` into the ledger at 17:12:32; the axis record landed as `cba77a5c` (#1682) at 18:04:04, 52 minutes later.

And it records **what widening would cost, as a constraint rather than a preference**: the payload-coincidence pass carries a maximum separation and no minimum, so a frameless two-member split at separation one falls inside `spec_corpus::is_svd_wg010_shape`, whose `guard_violations` is pinned at **0 over 210 rows as a verdict** about the rejected proposal. Widening lands on that guard on the first row it reaches — #1711 was exactly that. Any future widening must carry a frameless separation floor of two in the same change.

**The direction record's third ground is demoted, and the mechanism ground it never stated is promoted.** Ground 3 was that the ruling's whole evidence base is net-deletion — an inference from the scope of the *evidence* to the scope of the *rule*, which does not follow, and which now sits as corroboration beside Ground 2 (already marked "corroboration, not authority"). What carries the scope instead is intrinsic to `:46`'s mechanism: it states the construction as "parts of the **inserted sequence** align with the reference sequence", a **subsequence** relation, and a subsequence cannot be longer than the sequence it is drawn from — so the construction is unsatisfiable whenever the payload exceeds the span. The code says the same thing twice, independently: `payload_embeds_within_budget` refuses `payload.len() > span.len()` on its own terms (*"Each payload base must consume a span base"*, `src/normalize/merge.rs:7587-7590`), and the entry gate at `:8131` applies the stricter `<`. The net-insertion exclusion is therefore largely **redundant** against the shipped code, which is what makes the scope defensible without leaning on Ground 3.

**A carried-in figure is withdrawn as unverifiable.** That record reports that removing the direction gate would move "~5,273" classes, and says in terms that the figure is carried in and not re-derived. It cannot be re-derived against the current code and is inconsistent with it: with `COALESCE_MISMATCH_BUDGET` at 1, deleting the strict `<` gate admits only equal-length blocks, because `:7589` still refuses a longer payload, and at equal length the embedding affords no deletion — a near-degenerate class, not a five-thousand-row one. Anyone revisiting the direction must re-derive it and say which gate and which budget the measurement was taken against.

The practical consequence, recorded so a future revisit does not start from the wrong question: widening the direction is close to **vacuous** on the shipped code. Reaching an equal-length block additionally requires raising `COALESCE_MISMATCH_BUDGET`, whose own doc records that it "has no counterpart in the spec at all" — changing the mechanism rather than its scope, which is a different question from the one that record answers.

## Alternatives considered

**Leave it.** Cheapest, and the same overclaim just corrected in the sibling record. Rejected.

**Re-open as `undecided`.** Would un-ground shipping behaviour, two committed guards and #1956's correction, in exchange for describing as unsettled a question the project has in fact settled. Rejected; the reasoning is in the record.

**Widen either `delins.md:47` scope.** Rejected on the evidence above — the axis half lands on a decided negative guard, and the direction half is near-vacuous without a separate mechanism change.

## Provenance of the finding

A spec-only review of the pinned submodule — restricted to `assets/hgvs-nomenclature`, with this ledger and the repository's own prose deliberately out of scope so it could not absorb the framing it was asked to test — reported that neither this reading nor the position-wise one it displaced has textual support.

Worth stating plainly: that review was commissioned to test **this** record's replacement premise, having just been used to refute the sibling record's. It came back against the commissioning interest, which is the reason to trust it.

## First instance of #1953's schema

#1953 added the `house_choice` record kind and deliberately shipped **no** ledger instance — its only candidate was this rule, and converting it was the adjudication that PR was told not to make. This is that instance. The renderer's house-choice branch is exercised against a real record for the first time and publishes:

> **House choice.** This ruling is the project's own, made under `README.md` rule 5 (silent limb) where the recommendations do not decide. It cites no governing clause and must never be quoted as conformance.

## Notes on the generated artifacts

`tests/it/clause_ruling_index.rs` is **regenerated, not hand-edited** — `render_index()` was dumped from a throwaway test, the source restored byte-identically (`cmp`-verified), and the block spliced on its markers. Its diff is two lines: this record's relation to `general.md:34` moving from `(governing)` to `(cited)`, in both index sections. That the *only* thing to move in the whole clause index is those two lines is itself a check that the edit did what it claims.

`docs/NORMALIZATION_CONTRACT.md` is re-blessed, then confirmed passing **without** `BLESS_CONTRACT_DOC`.

## Gate

At the rebased head `2bd2f56f`: `cargo fmt --all --check` clean · `cargo clippy --features dev --tests --all-targets -- -D warnings` `EXIT=0` · ledger-guard filterset **55/55** · full `cargo nextest run --features dev --lib --test it --no-fail-fast` → **10807 passed, 0 failed**.

Ledger scope verified by round-trip at each commit: 31 records in and out, ids identical, statuses identical (28 `decided`, 3 other), only the intended records differing and only in the intended fields; clauses and guards byte-identical elsewhere. The rebase was checked the same way — 31 records against both the pre-rebase tip and `origin/main`, none dropped from either, no status changes.

Representation-Change: none


nh13:
@coderabbitai pause

PR #1981, opened 2026-08-15
perf(normalize): five allocation and formatting wins on the normalization hot path

### LINKED ISSUE #1972: perf: the seam oracles are 48% of the sweep runtime, and HgvsVariant copying is 10%
Profiling the workflow's slowest single test, to find out where the *runtime* half of CI actually goes. Companion to #1963 (compile time) and #1964 (build split).

## Method

`cis_junction_crossing_shift::no_two_member_allele_normalizes_to_a_different_sequence` — the 76.1s test that dominates the `Exhaustive sweeps` job — run under the exact env that job sets:

```
FERRO_SWEEP_SEEDS=full FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 \
FERRO_ASSERT_IN_BOUNDS=1 FERRO_ASSERT_SEQUENCE=1 \
  samply record -- target/soak/deps/it-<hash> <test> --exact --test-threads 1
```

35,740 samples across all threads, symbolicated through samply's own symbol server (6,922 of 7,414 frames named).

**Two traps for anyone reproducing this.** `[profile.soak]` sets `debug = false`, so a default soak build profiles as raw addresses — rebuild with `CARGO_PROFILE_SOAK_DEBUG=line-tables-only`, and note that re-fingerprints the whole dependency tree. And `samply record --save-only` does **not** symbolicate; it stores addresses and resolves them at load time, so the saved `.json.gz` is unreadable on its own. Either use the UI, or POST the addresses to the running server's `/<token>/symbolicate/v5`.

## The headline: nearly half the sweep is the oracles, not normalization

| | inclusive |
|---|---:|
| **`assert_seam_oracles`** | **47.90%** |

Broken down by what the oracle calls:

| under `assert_seam_oracles` | inclusive |
|---|---:|
| `normalize_core_checked` — the idempotency oracle **re-normalizing** | **37.99%** |
| `parse_hgvs` — the re-parse oracle | 4.32% |
| `compare_denoted_sequences` — the denoted-sequence oracle | 2.53% |
| `fmt` — rendering the output so it can be re-parsed | 2.32% |
| `first_out_of_bounds_coordinate` — the in-bounds oracle | 0.34% |

`FERRO_ASSERT_IDEMPOTENT` checks `norm(norm(x)) == norm(x)`, so **it doubles every normalization by construction** — that is the 38%, and it is inherent to the check rather than a defect in it. The other three oracles together are ~9%.

This is worth stating plainly because it reframes the CI cost: the `sweeps` and `censuses` jobs are roughly **twice** the cost of the work they verify, by design. Nobody should be surprised by that, and nobody should try to "fix" it by unsetting a flag — the whole argument for these jobs (recorded at length in `ci.yml`) is that an armed run subsumes an unarmed one. But it does mean a 2x runtime multiplier is sitting on the two jobs that are now the critical path, and any cheaper formulation of the idempotency check is worth ~38% of them.

## Library hot paths, oracle cost excluded

| self time | |
|---:|---|
| **8.02%** | `_platform_memmove` |
| 2.21% | `core::fmt::write` |
| 2.14% | `edit_grid` |
| 1.99% | `_platform_memset` |
| 1.59% | `SipHasher::write` |
| 1.05% | `parse_na_edit_inner` |
| 1.02% | `Accession::write_bare` |

Aggregated by class:

| | self |
|---|---:|
| `memmove` + `memset` + `memcpy` | **10.01%** |
| string formatting (`fmt`, `Accession::write_*`) | **7.47%** |

Inclusive, the library's own shape:

| | inclusive |
|---|---:|
| `canonicalize_from_sequence` | 19.07% |
| parser `process` (nom) | 10.59% |
| `normalize_genome` | 8.75% |
| `member_span` | 5.38% |
| `clamp_sibling_crossing_junctions` | 4.01% |

**`HgvsVariant` is expensive to move around.** It appears repeatedly across `clone`, `sort_by_key`, `sort_by`, `insert_tail`, `to_vec` and `drop_glue`, and 10% of self time is raw byte copying. That is the most promising single library target, and it would pay everywhere — not just in the sweeps.

## What this says about the two candidate directions

- **Optimizing the test-suite code itself is low-yield.** The profile is dominated by library work and oracle verification; the driver's own loops barely appear. A broad simplification pass over `tests/it/` would be a large diff justified by very little. (It also bears on #1964: the driver being cheap at runtime is consistent with it being cheap to run at a low `opt-level`, though that still needs its own measurement.)
- **Optimizing ferro itself has two concrete targets** rather than a vague mandate: `HgvsVariant`'s copy cost (10% memmove, plus the clone/sort/drop tail), and the format-then-reparse round trip the re-parse oracle forces (7.5% formatting self time). Both are measurable in isolation.

## Caveat

One test, one machine, one shape of input — a six-deep cross-product over two-member cis alleles on a `JsonProvider`. It is the right test to profile because it is the workflow's most expensive, but it is not a claim about ferro's profile on real-world corpora. Before acting on the `HgvsVariant` finding, confirm it on a bulk-corpus test (`clinvar_hgvs_tests`, `cmrg_exhaustive_tests`) as well.

nh13:
Five allocation- and formatting-shaped optimizations on the normalization hot path, driven by the profile in #1972. One commit per optimization, each independently reviewable and revertable.

## What the profile said, and what actually paid

#1972 named `_platform_memmove` at 8.02% self, string formatting at 7.47% aggregated, `member_span` at 5.38% inclusive, and called `HgvsVariant`'s copy cost "the most promising single library target".

The copy cost turned out not to be the enum's own moves. `HgvsVariant` is 560 bytes because `RnaFusionVariant` is 560 and sets the width for the six 424-byte single-locus arms; boxing that arm takes the enum to 432 (-23%) and **measured -1.4% ± 3.5% — indistinguishable from zero**. What did pay, every time, was removing the *heap payloads* those moves carry: rendered accessions, rendered descriptors, and a pass run twice.

## Measured

Two workloads, because they exercise different regimes:

**A. The profiled test.** `cis_junction_crossing_shift::no_two_member_allele_normalizes_to_a_different_sequence`, soak profile, `FERRO_SWEEP_SEEDS=12`, `RAYON_NUM_THREADS=1`, **user CPU**, 30 rounds interleaving every stage binary inside each round, reported as the mean of the per-round *paired* deltas. Wall clock is unusable on this machine (load 55-160 from concurrent builds; a measured A/A spread of 9.5-15.2 s wall against 14.6-15.1 s user CPU).

**B. `benches/benchmarks.rs`'s `allele_scaling` group at 100 members**, leave-one-out: each variant is HEAD with exactly one commit's file reverted, so the difference is that one change. This is where the per-member and pairwise passes actually cost something — the sweep's alleles have two members.

| # | optimization | mechanism | A: sweep | B: 100-member allele |
|---|---|---|---|---|
| 1 | accessions -> `SmolStr` | `member_span` allocated two `String`s per call and ~10 sibling passes read every member's span per pass | **-4.96%** (sd 6.18, 4.4σ) | — |
| 2 | deferred descriptor tie-break | `sort_by_key` re-derived a key containing `format!("{v}")` on every comparison | **-4.11%** (sd 6.60, 3.4σ) | — |
| 3 | one coincident-bounds pass | the second `detect_overlap_conflicts` call was a pure function on unchanged inputs already known to be empty | **-2.85%** (sd 5.14, 3.0σ) | — |
| 4 | overlap pass compares without allocating | `to_string()` per member *and* per pair in a quadratic loop | -0.30% (one pair there) | **-39% to -47%** |
| 5 | accession dropped from a fixed comparison | the sort gates on one accession, then rendered and compared it on every comparison anyway | **-0.21%** (sd 0.64, n=40, 95% CI [-0.41, -0.01]) | not resolvable |
| | **cumulative** | | **-14.7%** (see note) | **-54% to -64%** |

The cumulative sweep figure is for the seven-change chain the per-stage numbers were measured on; two of those seven were then dropped (below) and each contributed under 1%, so the shipped five account for essentially all of it.

**#5 is the weakest of the five and the table says so.** Its effect is `O(n log n)` in the member count and the sweep's alleles have two members, so -0.21% is that change's *floor*, not its typical value. The regime where it scales is the same one #4 shows -43% in, and that regime could not be measured here: criterion on the 100-member benches returned standard deviations of 26-140% while the machine sat at load 100-160, so no honest number exists for it. It is included on a measured (if small) win plus a mechanism that only grows; a reviewer who wants it out has a clean single revert.

Representation-Change: none.

Every change is behaviour-preserving by construction, and each commit says how:

- `SmolStr` equality and ordering are the `str`'s, and the new spellings share the existing accession writers, so the rendered bytes are identical.
- A tuple's ordering consults its last component only when every earlier one ties, so moving the descriptor render behind `then_with` is the same total order.
- The second `detect_overlap_conflicts` call is a pure function of inputs the early return has already established produce an empty result.
- Dropping the accession from a comparison the caller has just fixed to one value cannot change any comparison that sort makes; `dropping_the_accession_is_the_same_order_within_one_accession` pins the agreement, and pins that the two forms *do* differ across molecules so the substitution stays tied to its gate.

**Corpus:** `examples/dump_normalized_corpus.rs`, this branch (`9ab82b8c`) against its merge base (`7fe41b46`) and against current `origin/main` (`254768e3`) — **86,398 rows, 0 moved, across all 22 shape families**, with all three dumps byte-identical. What that does and does not cover: the harness builds cis alleles on 20-mers plus the protein axis and two designed long/revcomp families, so it exercises every path these changes touch — but it is a claim about those families, not a library-wide rate. These changes are not scoped to a shape, they are scoped to *how a string is built*, and the same rendering seams are on every row the corpus does build.

One part of the zero is structural and is named as such: every accession this corpus emits is 9-10 bytes, so **0 of 86,398 rows** carry an accession past `SmolStr`'s 23-byte inline capacity. The heap-spill arm of optimization #1 is therefore unmeasured here — it is a wash by construction (it spills to the same `String` `full()` built), but the corpus cannot witness it. The multi-member paths are measured properly: 66,518 rows (77.0%) carry two or more members and 4,634 carry three or more, which is what #2, #4 and #5 run over. Optimization #3's `debug_assert!` is compiled out under `--release`, so this run is not evidence for that premise; the immutable `&AlleleVariant` borrow is.

## Dropped, with the number that killed each

- **Box `RnaFusionVariant`** (enum 560 -> 432 bytes): -1.4%, paired SD 8.6% over six seeds=full reps. Not resolvable.
- **Count the tie-break's pieces instead of building them** (`best_alignment` allocated a `Vec<Column>` + `Vec<Piece>` per candidate gap position): three independent 15-round runs gave -3.54%, +3.68%, +0.18%, and a 30-round run gave **-0.16% ± 1.04%**.
- **`edit_grid` row slices, no reversed-block copies**: **-0.70% ± 1.07%** on the sweep and ~0 on the allele benches.
- **Junction clamp's member snapshot taken lazily**: -1.92% ± 1.06% on one 30-round run, **-0.30% ± 0.57%** on an independent 37-round run; pooled -0.66% ± 0.50%. Not resolved either way.

All four are strictly work-reducing and all four are neutral to this machine's resolution, so they are not here — a neutral change to `src/normalize/` is risk without upside.

## Gate

All on the shipped tree, each step's `EXIT=` read back out of its own log rather than inferred from a chain's status:

* `cargo fmt --check` — `EXIT=0`
* `cargo clippy --features dev --all-targets -- -D warnings` — `EXIT=0`
* `cargo nextest run --features dev --no-fail-fast` — `EXIT=0`, **10970 passed, 0 failed**, 151 skipped
* `pytest tests/python/` against a `maturin develop --features python` build — **1048 passed, 8 skipped**, `EXIT=0`. `cargo nextest` does not reach the bindings and this branch changes `src/normalize/` and `src/hgvs/`, so the Python suite is not optional here.

Closes #1972

### DIFF
--- a/src/hgvs/variant.rs
+++ b/src/hgvs/variant.rs
@@ -1341,11 +1341,14 @@ fn rna_simple_range(interval: &crate::hgvs::interval::RnaInterval) -> Option<Sel
 ///   why a total order is used rather than a stable sort keyed on position
 ///   alone — the latter would leave same-position members in input order, which
 ///   is the exact order-dependence #1098 is about.
-pub(crate) fn cis_member_order_key(
+///
+/// **This function is the positional part of that key** — everything before the
+/// descriptor. [`cis_member_order_cmp`] adds the descriptor tie-break on top,
+/// and only when it is going to decide.
+pub(crate) fn cis_member_order_prefix(
     v: &HgvsVariant,
-) -> (String, SelfCancellingPoint, SelfCancellingPoint, u8, String) {
-    let accession = v.accession().map(Accession::full).unwrap_or_default();
-    let descriptor = format!("{v}");
+) -> (SmolStr, SelfCancellingPoint, SelfCancellingPoint, u8) {
+    let accession = v.accession().map(Accession::full_smol).unwrap_or_default();
     // `SelfCancellingPoint` derives `Ord` over `(region, base, offset)` in that
     // field order, which is exactly the comparison this key wants, so the points
     // go in whole rather than flattened into six scalars.
@@ -1355,13 +1358,28 @@ pub(crate) fn cis_member_order_key(
     // because that is what "sorts last" spells.
     let sentinel = SelfCancellingPoint::new(true, i64::MAX, i64::MAX);
     let (start, end) = cis_member_range(v).unwrap_or((sentinel, sentinel));
-    (accession, start, end, junction_rank(v), descriptor)
+    (accession, start, end, junction_rank(v))
+}
+
+/// Compare two cis members by the total order described on
+/// [`cis_member_order_prefix`].
+///
+/// Exactly the lexicographic comparison of
+/// `(cis_member_order_prefix(v), format!("{v}"))`: a tuple consults its last
+/// component only when every earlier one ties, so moving the render into
+/// `then_with` cannot change the order. It changes the cost — the descriptor is
+/// the whole member rendered, and the sort used to derive it for every member on
+/// every comparison, once per pass of the allele fixed-point loop.
+pub(crate) fn cis_member_order_cmp(a: &HgvsVariant, b: &HgvsVariant) -> std::cmp::Ordering {
+    cis_member_order_prefix(a)
+        .cmp(&cis_member_order_prefix(b))
+        .then_with(|| a.to_string().cmp(&b.to_string()))
 }
 
 /// Where within its own span a member adds sequence: `0` at the span's start,
 /// `1` at its end.
 ///
-/// Breaks a tie that [`cis_member_order_key`]'s `end` cannot (#1301). An
+/// Breaks a tie that [`cis_member_order_cmp`]'s `end` cannot (#1301). An
 /// insertion's span *is* the gap it fills, so `264_265insCA` adds at interbase
 /// 264; a duplication places its copy after its last base, so `264_265dup` adds
 /// at 265. The two share both endpoints, and without this the order fell to the
@@ -1415,7 +1433,7 @@ fn junction_rank(v: &HgvsVariant) -> u8 {
 /// positions, or a non-positional arm such as a null/unknown-allele marker).
 /// Reuses the per-axis range extractors that back the self-cancelling detector.
 ///
-/// The end is what breaks a start tie in [`cis_member_order_key`]; see that
+/// The end is what breaks a start tie in [`cis_member_order_cmp`]; see that
 /// function's docs for why the start alone is not a sufficient discriminator.
 fn cis_member_range(v: &HgvsVariant) -> Option<SelfCancellingRange> {
     match v {
--- a/src/hgvs/variant.rs
+++ b/src/hgvs/variant.rs
@@ -1343,12 +1343,30 @@ fn rna_simple_range(interval: &crate::hgvs::interval::RnaInterval) -> Option<Sel
 ///   is the exact order-dependence #1098 is about.
 ///
 /// **This function is the positional part of that key** — everything before the
-/// descriptor. [`cis_member_order_cmp`] adds the descriptor tie-break on top,
+/// descriptor. `cis_member_order_cmp` adds the descriptor tie-break on top,
 /// and only when it is going to decide.
+/// **Test-only, and deliberately so.** Nothing ships this form: both sorts run
+/// behind a single-accession gate, so the accession component is constant across
+/// every comparison they make and `cis_member_order_cmp_within_accession` is what
+/// they call. This is retained as the *reference* the shipped comparator is
+/// checked against — see
+/// `dropping_the_accession_is_the_same_order_within_one_accession` — because a
+/// claim that two orders agree needs both orders to exist.
+#[cfg(test)]
 pub(crate) fn cis_member_order_prefix(
     v: &HgvsVariant,
 ) -> (SmolStr, SelfCancellingPoint, SelfCancellingPoint, u8) {
     let accession = v.accession().map(Accession::full_smol).unwrap_or_default();
+    let (start, end, rank) = cis_member_position_key(v);
+    (accession, start, end, rank)
+}
+
+/// `cis_member_order_prefix` without the accession.
+///
+/// Both callers of the order gate on every member sharing one accession before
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

PR #1984, opened 2026-08-15
fix(scripts): stop reading a fenced trailer as a representation declaration

### LINKED ISSUE #1929: A Representation-Change: trailer inside a fenced code block counts as a real declaration, so quoting CONTRIBUTING.md's example files a disclosure the author never made
`TRAILER_RE` matches at column 0 with no awareness of Markdown fencing, so a `Representation-Change:` line **inside a fenced code block** is read as a real declaration. `CONTRIBUTING.md` publishes three such blocks (`:99-128`), so the text that triggers this is copy-pasteable straight out of the contributor documentation.

## Reproduced against the shipped scripts

A PR body that declares nothing and merely quotes the docs:

````markdown
This PR refactors a helper.

For reference, CONTRIBUTING.md says the trailer looks like:

```
Representation-Change: 577 rows move, 360 merge / 205 split / 12 respell.
  Previously-accepted inputs, so a real migration for the consumer.
```

I have not worked out the disclosure for this change yet.
````

Against `changed = ["src/normalize/mod.rs"]`:

| | result |
|---|---|
| `find_declaration` | `'577 rows move, 360 merge / 205 split / 12 respell.'` |
| `check(...)` | **passes** |
| `declines(...)` | `False` — filed as a real change |
| `find_near_misses` | `[]` — nothing flagged |

So the required check goes green and the author is recorded as disclosing a 577-row migration they did not make. **That is a false disclosure, not a missed one** — the changelog's Representation changes section would carry a fabricated migration note attributed to a PR whose author explicitly said they had not worked it out.

## Both scripts share it, which is the part that matters

`check_changelog_grouping.py` imports `trailer_value` from the checker precisely so the column-0 rule is not duplicated, and returns the same quoted string. So the audit — which exists to be an *independently derived* second opinion — agrees, because on this input the two halves are not independent at all.

That is the #1555 shape, which `CLAUDE.md` already records: *"#1555 passed the test written to catch it precisely because that test compared the two halves against each other and they were wrong together."*

## What is already correct, so nobody re-fixes it

Every **decorated** form is caught and fails loudly, with a near-miss message naming the problem: inline code span, `**bold**`, `> blockquote`, and 4-space-indented. So decoration in general is handled; a fenced block is the one shape that reaches column 0 with its text intact.

Two declarations — a quoted example plus a real trailer — is also handled: `check` refuses with "2 `Representation-Change:` trailers found ... Keep exactly one", and explains that this check reads the first while git-cliff matches every footer. The gap is only the case where the **fenced example is the only trailer present**, which is exactly the case where the author declared nothing.

## Corrects the premise this was filed under

Tracked locally as "an inline-code-span trailer does not match `TRAILER_RE`, so the required check passes vacuously". Measured, that is wrong in both halves: an inline-code-span trailer does not match, and the check therefore **fails** — correctly, and with a near-miss hint. The real defect is the opposite shape (fenced, which *does* match) and the opposite failure (a confident wrong disclosure rather than a vacuous pass).

## Suggested shape

Strip fenced regions before scanning, in the shared helper so both scripts inherit it — track ``` and ~~~ opens/closes and skip lines inside them. Two things to keep:

- **A fenced trailer should be a near-miss, not silence.** It is nearly always someone quoting the docs, and the useful message is "this looks like a declaration but is inside a code fence; move it to column 0 outside any block."
- **Test it against the documentation itself.** The strongest regression guard is to run the checker over `CONTRIBUTING.md`'s own examples and assert none of them is read as a declaration — a guard that keeps working when the docs are rewritten, unlike a hand-copied fixture.

Found while verifying a local note about trailer decoration; the near-miss coverage is good enough that only the fenced case survives.

### PR DESCRIPTION
**Pull request by nh13**
Closes #1929.

`TRAILER_RE` anchors at column 0 and knows nothing about Markdown, so a `Representation-Change:` line **inside a fenced code block** was read as a real declaration. `CONTRIBUTING.md` publishes the trailer's form in fenced blocks, so the text that triggers this is copy-pasteable straight out of the contributor documentation — which is exactly how it would be met in practice, by someone quoting the docs to ask what to write.

**That is a false disclosure, not a missed one.** A body declaring nothing and merely quoting the example produced:

| | before |
|---|---|
| `find_declaration` | `577 rows move, 360 merge / 205 split / 12 respell.` |
| `check(...)` | **passes** |
| `find_near_misses` | `[]` — nothing flagged |

So the required check went green and the author was recorded as disclosing a 577-row migration they said, in the same body, they had not worked out.

Every other decoration was already handled and fails loudly with a near-miss hint — inline code span, `**bold**`, `> blockquote`, four-space indent. A fence is the one shape that reaches column 0 with its text intact.

## The fix

Fenced regions are blanked before the declaration is chosen, **preserving line numbers** so the fenced line is *reported* as a near-miss rather than passed over in silence:

```
line 6: Representation-Change: 577 rows move, 360 merge / 205 split / 12 respell.
  -> inside a fenced code block, so it is documentation rather than a declaration;
     move it to column 0 outside any ``` or ~~~ block
```

Someone who pasted the documented form is asking a question, not staying silent, and the message names the fix.

Fence tracking follows CommonMark where it matters: the closing fence must use the same character and be at least as long as the opener (so a ```` ```` ```` block may quote ```` ``` ```` lines — the shape the issue reproduced with), and an unclosed fence runs to end of text, which withholds a declaration rather than inventing one.

## The part that is deliberately *not* symmetric

**The duplicate refusal counts the UNSTRIPPED set**, and that asymmetry is the point.

`git_conventional` parses footers with no Markdown awareness — any `word:` line is a footer token — so git-cliff sees a fenced trailer too. Making both halves fence-aware looks tidier and is a **safety regression**: a body carrying a quoted example *beside* a real `none` would then pass here while the changelog still had two footers to choose from, and the quoted 577-row example could land under an author who declined.

So: **fenced text may not BE the declaration, but it still COUNTS toward how many the changelog will see.** The refusal is what forces the author to indent or drop the quotation, which is the escape hatch `CONTRIBUTING.md` already documents. I nearly shipped the symmetric version; the four-shape table below is what caught it.

| body | verdict |
|---|---|
| fenced example only | **refused**, with the near-miss hint (this is #1929) |
| fenced example + real trailer | **refused** — 2 trailers, as git-cliff would see |
| real trailer alone | passes |
| two real trailers | refused |

**The sibling scripts are left alone for the same reason.** `check_changelog_grouping.py` keeps its own copy of the column-0 pattern and `inject_representation_disclosure.py` imports it from there; both read *commit messages* and exist to check or reproduce what git-cliff actually did. Teaching them about fences would make them disagree with the changelog they audit. (The issue said the audit imports from the checker — it is the other way round; corrected here so the next reader does not propagate the fix into the wrong place.)

## Tests

Twelve. Ten were verified against the fence-strip: reverting it turns exactly those ten red and nothing else. Two cover the path where the fenced case did *not* get the fenced message — two fenced examples and no real trailer, which hits the duplicate refusal before the near-miss scan runs. One of those two fails without the fix (the message never says "fence"); the other is its negative control, asserting that a genuine duplicate is still told to delete one.

The strongest is the one the issue suggested: **run the checker over `CONTRIBUTING.md` itself** and assert its own examples are not read as a declaration. Unfixed, that test reports the live documentation declaring `c.<cds_end>_*1ins<A> now renders as`. That is the defect demonstrated against shipped docs rather than a hand-copied fixture, and the guard keeps working when the docs are rewritten.

## Verification

`ruff check` and `ruff format --check` clean. `pytest tests/python/test_representation_change_trailer.py` — **161 passed**. Against the live tree, `find_declaration` now returns `None` for both `CONTRIBUTING.md` and `CLAUDE.md`.

Representation-Change: none. Tooling only — this changes which PR-body text the checker reads as a declaration, and no `src/` file is touched.

### DIFF
--- a/scripts/check_representation_change.py
+++ b/scripts/check_representation_change.py
@@ -104,6 +104,83 @@ TRAILER_RE = re.compile(
     re.IGNORECASE | re.MULTILINE,
 )
 
+#: Opens or closes a Markdown code fence: three or more backticks or tildes, indented by at
+#: most three spaces (CommonMark's limit before it becomes an indented code block).
+FENCE_RE = re.compile(r"^ {0,3}(?P<fence>`{3,}|~{3,})(?P<info>.*)$")
+
+
+def fence_match(line: str) -> re.Match[str] | None:
+    """Match a fence line, honouring CommonMark's backtick-info restriction.
+
+    A backtick fence's info string may not itself contain a backtick — otherwise an inline
+    code span could open a block. Tilde fences carry no such restriction. `FENCE_RE` alone
+    accepts both, so ```` ```a`b ```` reads as an opener here while CommonMark (and GitHub,
+    and git-cliff's eventual reader) treat it as ordinary text.
+
+    Getting this wrong fails closed rather than open — a spurious opener blanks the rest of
+    the body, so a real declaration goes unseen and the author is refused rather than
+    credited with a disclosure they did not write. That is the safe direction, and it is
+    still the wrong answer.
+    """
+    match = FENCE_RE.match(line)
+    if match is None:
+        return None
+    if match.group("fence")[0] == "`" and "`" in match.group("info"):
+        return None
+    return match
+
+
+def fenced_line_numbers(text: str) -> set[int]:
+    """Return the 1-based line numbers inside a fenced code block, fences included.
+
+    `TRAILER_RE` anchors at column 0 and knows nothing about Markdown, so a trailer quoted
+    inside a fence reads as a real declaration (#1929). That is the one decoration that
+    reaches column 0 with its text intact — an inline code span, `**bold**`, `> blockquote`
+    and a four-space indent are all caught already, and correctly *fail* with a near-miss
+    hint. This one **passed**, which is worse: the author is recorded as disclosing a
+    migration they explicitly said they had not worked out.
+
+    `CONTRIBUTING.md` publishes three such blocks, so the text that triggers it is
+    copy-pasteable straight out of the contributor documentation — which is exactly how it
+    would be met in practice, by someone quoting the docs to ask what to write.
+
+    The closing fence must use the same character and be at least as long as the opener, so
+    a ```` ```` ```` block may contain ```` ``` ```` lines — the shape the issue reproduced
+    with. An unclosed fence runs to the end of the text, which is CommonMark's rule and the
+    safe direction here: it withholds a declaration rather than inventing one.
+    """
+    inside: set[int] = set()
+    open_fence: str | None = None
+    for number, line in enumerate(text.splitlines(), 1):
+        match = fence_match(line)
+        if open_fence is None:
+            if match is not None:
+                open_fence = match.group("fence")
+                inside.add(number)
+            continue
+        inside.add(number)
+        if (
+            match is not None
+            and match.group("fence")[0] == open_fence[0]
+            and len(match.group("fence")) >= len(open_fence)
+            and not match.group("info").strip()
+        ):
+            open_fence = None
+    return inside
+
+
+def strip_fenced(text: str) -> str:
+    """Blank every fenced line, preserving line count so numbering still lines up.
+
+    Blanking rather than deleting is what lets `find_near_misses` report a fenced trailer at
+    its real line number in the same pass.
+    """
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #1985, opened 2026-08-15
docs(ledger): qualify what the brute-force dominator oracle corroborates

nh13:
Refs #1959 — the half that does not touch `src/normalize/seqfirst/align.rs`, which PR #1973 currently owns.

`rulings[unchanged-is-read-over-every-minimal-alignment]` supports its cell-vs-column distinction with:

> That is a notion mismatch, not a bug: `dominators()` was checked against its own brute-force oracle and agrees with it.

That sentence is **accurate** — note "its own" — and it reads as independent verification of the DAG, which it is not.

## What the probe actually establishes

`dominators_by_brute_force` takes the `AlignmentDag` and walks it: its cell set, its edges, its reachability. Every term comes from the DAG; nothing is re-derived from the two sequences. So it establishes that `dominators()` agrees with edge-banned reachability **over that graph** — and a graph missing cells is missing them on both sides of the comparison.

That is the definition of an oracle reading the artefact rather than the thing the artefact describes, and it fails in the direction you cannot see: a *smaller* DAG has *more* forced junctions, so a shrunken graph yields a self-consistent, plausible, wrong answer.

## Demonstrated, not argued

Found while implementing banded DP for #1928. Deliberately accepting a band one doubling too narrow produces a DAG that on `AC -> CA` retains **3 cells where the true DAG retains 7** — and `dominators_agree_with_brute_force_on_exhaustive_small_inputs` **passes**.

Two other sabotages in that set (band centred on `0` instead of `delta/2`; `cells()` clipped one diagonal too tight) *were* caught by it, which is worth stating so the guard is not written off: it catches asymmetric and inconsistent damage, and is blind to uniform shrinkage — which is the direction a well-meaning refactor actually takes.

## The ruling is unaffected, and the record says so

The cell-vs-column distinction rests on `GACA -> AGAT`, worked in the record from the minimal alignments themselves. What is withdrawn is any reading of that one clause as evidence the DAG is right — not the ruling, and not the distinction.

## Scope

Ledger prose only. `docs/NORMALIZATION_CONTRACT.md` is its generated render and is re-blessed; the ledger guards pass **without** `BLESS_CONTRACT_DOC` set.

Verified tight rather than assumed: **31 records before and after, none dropped, no status changed, exactly one record's content moved.**

The remaining half of #1959 — a comment on `dominators_by_brute_force` itself stating what it does and does not establish — is deliberately left out, because #1973 rewrites that file and already documents the hazard in its module header. #1959 stays open for it.

## Verification

Full gate with `FERRO_MANIFEST` armed and `FERRO_REQUIRE_BULK_FIXTURES=1`: **11,022 passed**, zero `Skipping:` lines. Python **1,059 passed**. `cargo fmt --all --check` and `cargo clippy --all-features --all-targets -- -D warnings` clean.

Representation-Change: none. Ledger prose and its generated render; no `src/` file is touched and no description ferro emits can change.


nh13:
@coderabbitai pause

PR #1994, opened 2026-08-15
ci: bound every PR-gating job with a timeout-minutes

nh13:
A job with no `timeout-minutes` inherits GitHub's 360-minute default. This adds one to every job in the workflows that gate a PR: `ci.yml` (20 jobs), `coverage.yml`, `lint-workflows.yml` and `report-failure.yml`.

Representation-Change: none

## Why

On #1977, `Test (2/6)` sat in its nextest step for **71 minutes** against a **76-second** norm while the other 32 jobs were green. Nothing would have stopped it until 360 minutes — six hours of a blocked PR and a held runner, for a job whose real ceiling is under two minutes. I cancelled it by hand; on the re-run the same job passed in ~76s, and the tree was already proven good (that head is a re-signed copy of `87024a1c`, which has two green CI runs), so it was a wedged runner rather than anything in the PR.

`audit-pr-runs.yml` already carries this argument for its own job, in its own words — *"a `gh` call that hangs would otherwise sit against the 6-hour default limit"*. This generalizes a decision the repo has already made once.

## Sizing, and why it is loose

The failure mode that matters here is **a timeout firing on legitimate work**, not a hang running long. `.config/nextest.toml` records what that costs: a `slow-timeout = { period = "60s", terminate-after = 2 }` killed the real censuses at 120s, and that is why there is deliberately no `[profile.ci]`. These values bound a hang; they are not a budget, and nothing should be tuned to fit them.

| tier | jobs | rationale |
|---|---|---|
| **60** | `build`, `clippy`, `clippy-all-features`, `spec-fixtures`, `test-build`, `test-build-soak`, `python-wheel-test`, and `coverage.yml`'s `coverage` | compiles Rust, so a cold cache is on the table |
| **30** | `test`, `test-oracle`, `sweeps`, `censuses`, `soak` | downloads a prebuilt archive and runs tests |
| **15** | `abi3-floor-consistency`, `format`, `python-lint`, `representation-change`, `merge-commit-signature`, `changelog-grouping`, `fixtures`, `test-required`, `lint`, `report`, `doctests` | lint/aux; every one under 25s observed |

Each value is at least **5x** the slowest run observed across six CI runs on 2026-08-15. The slowest job in `ci.yml` was `Test oracle (1/4)` at **345s**, against a 60-minute tier; `coverage.yml`'s shards peak at **588s**, which is why `coverage` is 60 rather than 30. The rule is written into the header comment so a future edit re-derives from measurement instead of guessing.

## Scope

Nightly, release, fuzz, external-validation, publish, deploy-local and prune workflows are **left alone**. The same argument applies to them, but sizing needs their own measurements — `release-wheels` builds across platforms and `external-validation` hits the network — and guessing would violate the rule this PR is establishing.

## Verification

- `actionlint` clean across all workflows. It caught a real mistake first: `audit-pr-runs.yml` already had `timeout-minutes: 10` and the initial pass added a duplicate. `yaml.safe_load` accepted that silently (last-wins) — only the linter saw it. That file is reverted and untouched here.
- Every inserted line was asserted to sit directly after its job's `runs-on`, including the matrix jobs where a `strategy:` block intervenes.
- All five workflows parse.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,10 +205,49 @@ env:
     + test(defect_non_idempotent_outputs)
     + test(issue_1715_rna_alignment_symbol_reach)
 
+# Every job carries a `timeout-minutes`. Without one a job inherits GitHub's
+# 360-minute default, so a wedged runner holds a PR for six hours: observed on
+# #1977 (run 31905255274, attempt 1), where `Test (2/6)` sat in its nextest step
+# for 72 minutes against the same shard's 65-second step on the run's own second
+# attempt, while the other 32 jobs were green, and would have run to 360.
+#
+# The tiers are deliberately loose, because the failure that matters here is a
+# timeout that fires on legitimate work. `.config/nextest.toml` records the
+# precedent: a `slow-timeout = { period = "60s", terminate-after = 2 }` killed
+# the real censuses at 120s, and that is why there is no `[profile.ci]`. These
+# bound a hang; they are not a budget, and nothing should be tuned to fit them.
+#
+#   60  compiles Rust, so a cold cache is on the table
+#       (build, clippy{,-all-features}, spec-fixtures, test-build{,-soak},
+#        python-wheel-test, and coverage.yml's `coverage` and `doctests`)
+#   30  downloads a prebuilt archive and runs tests (test, test-oracle,
+#       sweeps, censuses, soak)
+#   15  lint/aux, none of which compiles Rust
+#
+# Sizing, re-derived 2026-08-15 across every ATTEMPT of the last 40 completed
+# `ci.yml` runs, 60 `coverage.yml` runs and 40 `lint-workflows.yml` runs — via
+# `/runs/{id}/attempts/{n}/jobs`, because the default `/runs/{id}/jobs` returns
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
Cross-links, since three related items now exist:

- **#2000** — the hang itself. This PR does **not** fix it, and that issue says so explicitly: a timeout converts an indefinite hang into a *failed* job, bounding the blast radius rather than removing the cause. There are now three known instances, on two different shards (`Test (2/6)` twice, `Test (5/6)` once).
- **#2004** — the half this PR deliberately scoped out: `release-wheels`, `external-validation`, the nightlies, `fuzz`, `publish`, `prune-ci-caches` and `deploy-local` are all still on the 360-minute default. Left out because sizing them honestly needs their own measurements, which is the rule this PR is establishing.
- **#2002** — unrelated to timeouts, but it is where the CI-wall work goes next: `Build` at 242s becomes the floor once #2001 lands.

For the record on this PR's own sizing: the two hangs it would have caught ran **41** and **71** minutes in `test` shards, against a 70–94s baseline. Those shards are on the 30-minute tier here, so both would have failed at 30 minutes instead of blocking the queue for six hours.

Issue #1829, opened 2026-08-13
Re-normalizing a projected n. axis strands an identity member (`63_64=`) — the #1655/#1281 shape, on a real NG_/NM_ pair


nh13:
Measured while re-grounding #1616's `NONCODING_NON_IDEMPOTENT` residue at base `f17a0895`.

```
NG_012337.1(NM_012459.2):c.[10del;23_25del;36_37insAAT]
  re-normalizes to
n.[58_60del;63_64=;67_68insATA]
```

`63_64=` is a **stranded identity member** — a member asserting that two bases are unchanged, sitting inside an allele. That is the #1655 / #1281 shape, the one `stranded_identity_member` already pins as a defect.

## Why this one is worth separating from #1712

#1616 takes the projected `n.` axis non-idempotency residue from 21 to 51. **29 of those 30 new rows are the projector being stale** — it emits a spanning `delins` (`n.192_197delins…`) that the normalizer then correctly splits (`n.[192_194delinsGTC;196C>T;198_199insGCTCATTATCTGGCT]`). That is #1712's own thesis, and it is a pre-existing defect that #1616 *exposes* rather than causes.

**This row is not that.** It is the 30th, and its output is malformed in a different way: the re-normalization strands an identity member rather than producing a stale-but-well-formed spelling. So it does not belong in #1712's bucket and should not be closed by whatever fixes the projector.

## Relationship to the existing guard

`stranded_identity_member` pins this shape as a defect, and the denoted-sequence oracle already treats "the output denotes no sequence while the input does" as a fire rather than a skip — deliberately, because a skip that reads as a pass is the failure mode that oracle exists to remove. This row is a live instance reachable from a real `NG_`/`NM_` pair rather than from a constructed one, which makes it a better reproducer than the pinned case.

## Disposition on #1616

#1616 discloses the residue as **29 projector + 1 defect** rather than quoting 51 as one number, by operator ruling of 2026-08-13. This issue is the "+1". #1616 is not blocked on it.

## Suggested first step

Determine which pass strands the member — whether the identity survives from the projector's spanning `delins` or is introduced by the split — before deciding where the fix belongs. The 29/1 division above says the projector is upstream of most of this residue, so the tempting assumption is that it is upstream of this one too; that is exactly the assumption worth checking rather than inheriting.

nh13:
Resolved by intervening work — no longer reproduces on `main`.

[...]

PR #1995, opened 2026-08-15
docs(ci): re-measure the denoted-sequence oracle; #1690 is no longer a blocker

### LINKED ISSUE #1990: ci: re-measure whether FERRO_ASSERT_SEQUENCE can now be armed in test-oracle — one of its two documented blockers is closed
`CLAUDE.md` records that arming `FERRO_ASSERT_SEQUENCE` in the `test-oracle` job is blocked on two things:

> So arming it is blocked on **#1690** and on finding a home for `stranded_identity_member` — **not** on another measurement.

**#1690 is now closed.** So half the documented blocker set is stale, and the next person to look at this will either re-derive it or trust a list that is no longer true.

## Why this matters more than a doc fix

`FERRO_ASSERT_SEQUENCE` is the only one of the four seam oracles that asks what an output **means** rather than how it is written. `CLAUDE.md` records eight hand-found instances of the class it covers (#1254, #1281, #1290, #1304, #1308, #1312, #1592, #1600), and notes that a wrong-sequence output passes all three of the other oracles — it is a fixed point, it parses, and its coordinates exist.

It currently runs in `sweeps` and in the nightly (which is `continue-on-error`). It does **not** run in `test-oracle`, which is the sharded job covering the bulk of the suite.

## What to measure

Re-run `test-oracle`'s own selection on current `main` with the flag armed, and report what fires now:

```sh
FERRO_ASSERT_SEQUENCE=1 \
  cargo nextest run --features dev --test it -E "<test-oracle's selection>"
```

The last recorded run of this (at `674e9c8b`) read `10383 tests run: 10378 passed, 5 failed, 289 skipped`, in two classes:

- **3x `issue_1487_canonical_window_overflow`** — not an oracle fire at all, but the `mapper.rs` overflow, which is #1690's subject and is now closed. **These are the ones expected to have gone away.**
- **2x `stranded_identity_member`** — a genuine fire, on a module that exists to *pin* a defect. That is the `ORACLE_EXCLUDE` class: a test pinning a defect and an oracle aborting on it cannot both run. Open issues around it: #1829, #1655, #1815.

## Outcomes and what each implies

- **Only the 2 `stranded_identity_member` rows fire** — then the remaining blocker is purely the question of where that module lives, and this becomes a scoping decision rather than a measurement.
- **Something new fires** — that is a defect of the class this oracle exists to catch, found for free.
- **Nothing fires** — arm it in `test-oracle` and the coverage gap closes.

Whatever the result, correct `CLAUDE.md`'s blocker sentence so it stops naming a closed issue. Note also `censuses`' armed step deliberately inherits `test-oracle`'s flag set exactly, so arming one changes both — that is by design and is documented, but it means this is not a one-job change.

## Do not read the count as a target

`ORACLE_EXCLUDE`'s modules and `test-oracle`'s selection are **different scopes** that have each produced a "5 failures" figure for unrelated reasons. `CLAUDE.md` warns about exactly this: "The two figures share a digit and nothing else, so never carry one over to the other's scope." Say which selection any number was taken on.

nh13:
`CLAUDE.md`, `ci.yml` and `scripts/run_oracle_suite.sh` all recorded that arming `FERRO_ASSERT_SEQUENCE` in the `test-oracle` job was blocked on **#1690** and on finding a home for `stranded_identity_member`. #1690 is closed, so half of that blocker set named a closed issue. This is the re-measurement, plus the doc correction that follows from it. No CI behaviour changes.

## What was run

`test-oracle`'s selection, read out of `ci.yml` by `scripts/run_oracle_suite.sh --print-selection` rather than rebuilt by hand, with that job's three flags plus `FERRO_ASSERT_SEQUENCE=1`, `FERRO_REQUIRE_BULK_FIXTURES=1`, no `--partition`, at `c9207d7e` (94 commits past the `674e9c8b` baseline). Zero `Skipping:` lines in both runs, so the bulk suites were armed rather than silently absent.

## Result, on `test-oracle`'s selection, without `FERRO_MANIFEST`

```
10904 tests run: 10902 passed, 2 failed, 306 skipped
```

against the recorded `10383 run / 10378 passed / 5 failed / 289 skipped` at `674e9c8b`.

- **The 3 `issue_1487_canonical_window_overflow` rows are gone.** All 8 tests in that module pass. #1690's fix holds.
- **Nothing new fired.**
- The 2 survivors are both `stranded_identity_member`, on the row `ci.yml` already quotes — input `TEMPLATE:g.[10_11insAT;11_12inv;12_13insA]`, output `TEMPLATE:g.[11_12insTA;11_12=;12_13insA]`:
  - `every_stranded_identity_is_non_confluent_with_its_surviving_member`
  - `the_wider_multi_member_census_is_what_the_header_says`

## The `FERRO_MANIFEST` half was stale too

With the manifest armed the same selection reads `10901 passed, 3 failed`. All six `mutalyzer_normalize_tests` axes now **pass**, contrary to what `ci.yml` recorded. The one extra row is `multi_member_cis_axis::axis_normalized`, and it is **not an oracle fire**: run as a control with `FERRO_ASSERT_SEQUENCE` **unset**, it fails identically, on a census pin reading `unwindowed: 89` against a pinned `90` while `sequence_changed` stays `0`. That is the HEAD-drift already documented for that pin (34 `src/` commits since it last moved), and it is orthogonal to this flag.

## The flag is NOT armed

`stranded_identity_member` still fires, so the measurement is still red and `test-oracle`'s `env:` is untouched — verified by parsing the workflow: it still carries exactly `FERRO_ASSERT_IDEMPOTENT`, `FERRO_ASSERT_REPARSE`, `FERRO_ASSERT_IN_BOUNDS`, `FERRO_REQUIRE_BULK_FIXTURES`. `censuses` inherits that flag set exactly, so arming would change two jobs.

The remaining blocker is `stranded_identity_member` alone — a module that exists to *pin* a defect (#1655, #1829, #1815), which is the `ORACLE_EXCLUDE` class: a test pinning a defect and an oracle aborting on it cannot both run. That is a scoping decision, not a measurement.

## Counts are labelled by scope

`ORACLE_EXCLUDE`'s modules and `test-oracle`'s selection have each produced a "5 failures" figure for unrelated reasons. Every number here is on `test-oracle`'s selection and says so.

## Verification

`scripts/run_oracle_suite.sh --print-selection` is byte-identical before and after the `ci.yml` comment edits, and the guards that read these files at runtime pass: `oracle_exclude_invariant`, `adjudication_prose_regrowth`, `ruling_citation_currency`, `normalization_contract_doc` — 32 tests run, 32 passed.

Representation-Change: none. Documentation and CI comments only; no `src/` change, no normalization behaviour touched, and no workflow `env:` or selection altered.

Closes #1990

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1504,7 +1504,8 @@ jobs:
         #     reaches `cds_to_tx`'s unchecked `cds_start as i64 + pos.base - 1`
         #     on an `i64::MAX`-adjacent `c.` coordinate. That module already
         #     records the measurement and names the site; the open issue is
-        #     **#1690**. The three are
+        #     **#1690**, since CLOSED — see the re-measurement below. The three
+        #     are
         #     `a_lone_member_at_an_extreme_coordinate_…`,
         #     `an_extreme_coordinate_…` and
         #     `two_adjacent_members_at_the_top_of_the_range_…`.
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

PR #1999, opened 2026-08-15
docs(rulings): correct the axis record's enforcement pin from 8 to 3

nh13:
`rulings[coding-axis-merges-are-a-disclosed-general-34-deviation]` states that `coding_axis_separation_two_or_more_merges` *"moves from 0 to **8** in both censuses, and that pin is the whole enforcement of this ruling"*, and frames its tripwire as *"a rise above 8"*.

**The pin has never been 8.** The first commit corrects the figure and the direction language. The second then answers the question that leaves open, by re-deriving the set: **the record's disclosed population is empty.**

**The ruling's text is unchanged and the record is not retired** — see [Deliberately not decided](#deliberately-not-decided-here).

## The measurement

The pin has only ever held two values:

| commit | PR | value |
|---|---|---|
| `c165c939` | #1710 | `0` (introduced) |
| `750f2334` | #1835, 2026-08-14 04:12:55 | **`3`** |

No neighbouring census field holds 8 either, so it is not a transposition from an adjacent line.

## Why this is not staleness

The record **did not exist** at `750f2334`, nor at `2d8b490b~1`. It was authored by #1899 (`2d8b490b`) at **15:55:50 the same day — 11h43m after the pin had already been set to 3.**

The 8 is a *true measurement at #1616's base*, under the partition rule #1835 replaced. It was carried forward and restated in the present tense against a pin that had already moved. So this is not a value that decayed while a correct record failed to notice — it is a record that did not match the code it names as its whole enforcement on the day it was written.

**That measurement is not withdrawn.** The eight-row table, the `sepN` reads and the "all eight are pure deletions" chain are all history and stay. What is withdrawn is the present-tense claim that the *pin* is 8.

## The tripwire is not directional

*"A rise above 8 must NAME THE CLAUSE THAT CARRIED IT"* describes the bespoke `assert!` at `tests/it/spec_conformance_axis.rs:1963`, which does fire only on a rise. But it is followed at `:1982` by `assert_eq!(census, pinned)`, which catches movement in **either** direction.

So a fall could not have gone unflagged, and did not: #1835 re-pinned deliberately and named the three surviving rows in a comment at `:848-852`. The requirement is now stated as **any change from 3, up or down, must name the clause or ruling that carried it.**

## The record predicted this itself

Worth stating, because it is why the substance survives untouched. The record's own falsifier already reads:

> the count may legitimately fall as well as rise, and in either direction the question to answer is which rows are in the set and on what ground — **never whether the number is still 8**. Read the eight ids above, not the digit.

This correction is that instruction being followed. Only the digit and the direction language were wrong.

## What changes

| file | change |
|---|---|
| `hgvs_spec_normalization_overrides.json` | correction block appended to the rationale, withdrawn-and-quoted per the record's convention |
| `docs/NORMALIZATION_CONTRACT.md` | re-blessed — generated from the ledger, so it carries automatically |
| `CLAUDE.md` | the adjudication table row: tripwire sentence and falsifier tail |
| `tests/it/spec_conformance_axis.rs` | the positive control's doc comment at `:2046-2056`, which still described the pin as **0** |

On that last one: the doc is converted to past tense rather than having its digit swapped, and records that a nonzero pin closes the specific blindness it was built for — a counter wired to `false` would now read 0 against a pin of 3 and fail both censuses' `assert_eq!`. The control is kept anyway, because it is the only test exercising the predicate *directly* rather than through a census total, and the blindness returns the moment the pin returns to 0.

## The set, re-derived (second commit)

Measured on the tree rather than read off the count. Both censuses were mis-pinned so the `assert_eq!` at `spec_conformance_axis.rs:1982` fires and prints its findings report; the flagged rows were read off it; the source was restored byte-identically from a copy taken beforehand (`cmp`-verified), not with `git checkout`.

**Exactly three merges, in two families:**

| n | separation | row | authored | output |
|---:|---:|---|---|---|
| 2 | 2 | `s01-c3m-pair-del-sub-p4-sep2` + `c3p` twin | `c.[24_27del;30C>G]` | `c.24_30delinsTTG` |
| 1 | 3 | `s01-c1-pair-inv-del-p4-sep3` | `c.[9_12inv;16_19del]` | `c.9_19delinsGTCAGTA` |

That reproduces both the pin of 3 and this file's own "The 3 are named and adjudicated" section at `:539-591`.

**The measured set intersects the record's eight named rows in nothing** — the record's are `m3`/`m4` multi-member all-deletion designs, the measured are `pair-*`. Since the counter counts *every* merging coding-axis row authored at separation two or more, a count of 3 whose rows are named and are none of the eight settles it: not one of this record's rows is currently a counted merge.

**And the three that are counted are not deviations.** #1835 adjudicated them at `:539-591` — *"all three are accounted for and none is a `general.md:34` deviation"*:

- the `sep3` row is returned **whole** by the canonical partitioner, so there is no derived separation for `:34` to govern (per `separation-is-a-property-of-the-spelling-not-of-the-variant`); it is counted only because the counter keys on the **authored** member count, which is what makes it an instrument rather than a verdict;
- the two `sep2` rows derive to `c.[24_26del;29_30delinsG]`, and `29_30delinsG` places one base while consuming two — gap-bearing, so `DNA/delins.md:47` reaches them and they **leave** the set. That is this record's own falsifier firing, in its own words.

**So the disclosure is vacuous on the current partition rule.** Nothing here says the record was wrong when written — it was a true disclosure at #1616's base. The thing disclosed has since stopped happening.

## Deliberately not decided here

**The record is NOT retired and its `status` is NOT changed.** Retiring a `decided` ruling is an operator act under `adjudication-precedence-order`, and a vacuous disclosure is harmless where a silently-wrong one is not. The recommendation is to retire it as superseded, or narrow it explicitly to #1616's base, rather than leave a live disclosure over an empty set.

## Root cause, which was predicted verbatim

`spec_conformance_axis.rs:593-600`, written while #1835 was in review:

> **#1616 is concurrently re-pinning this same counter to 8**, under a separate operator ruling ... Two open PRs moving one counter to different values is a semantic conflict no textual check sees — the constants live on different lines, so `merge-tree` reports no conflict between the two branches. **Whichever lands second must re-measure this figure on the merged tree rather than keeping its own.**

#1899, the #1616 lineage, landed second and kept its own figure. The warning was written, was correct, named the right remedy, and was not followed — so the failure is not that the hazard was unknown but that **nothing enforced it**. A prose figure in a ledger record has no guard. That is the same lesson `:602-611` already draws about #1698 removing 119 of 122 rows with no test failing.

## Gate

At the head of both commits: `cargo fmt --all --check` clean · `cargo clippy --features dev --tests --all-targets -- -D warnings` `EXIT=0` · ledger + axis guard filterset → **73 passed, 0 failed**, including `normalization_contract_doc` confirmed passing **without** `BLESS_CONTRACT_DOC`.

One guard caught a real mistake on the way: `ruling_citation_currency::every_record_to_record_citation_resolves` reads backtick-quoted identifiers as record citations, and the first draft backticked the corpus **row** ids — which the existing table deliberately leaves bare. Fixed by leaving them bare; it is why the row ids are unquoted in the prose above.

Ledger scope verified by round-trip at **each** commit: 32 records in and out, ids identical, statuses identical (29 `decided` / 3 `undecided`), exactly **one** record differing in exactly **one** field (`rationale`), all other top-level keys equal, and a one-line diff — so the file's mixed escaped/literal unicode was not reformatted.

Representation-Change: none

**Review fixes (2026-08-15).** Three edits inside this record's correction block, none touching the ruling, its `status`, or any code. (1) The correction's absolute *"8 was never committed as a pin at any point in that file's history"* is replaced by a `main`-scoped claim: 8 **was** committed as a pin, at `386a0adb` on #1616's own branch `fix/remove-input-relative-weight-bound`, which was closed unmerged and never reached `main`. This removes a contradiction with the second commit, which quotes the module doc warning that #1616 was re-pinning the counter to 8. (2) *"a true reading at #1616's BASE"* is corrected to *"with #1616 **applied** to its pre-#1835 base"* — at #1616's base itself (`f17a0895`) the counter read **0**; the 8 is the reading with #1616 in place. Both readings are one `git show` each and are named in the record. (3) The withdrawn tripwire paragraph is now marked **at the site of the claim** (`docs/NORMALIZATION_CONTRACT.md:607`) rather than only ~700 characters into a later paragraph, and the correction's 12 section separators are `\n\n` instead of two literal spaces, so the block no longer renders as one 8,890-character line.


nh13:
@coderabbitai pause

PR #2001, opened 2026-08-15
perf(ci): take the last two heavy censuses off the debug shards

nh13:
Takes the last two heavy censuses off the debug-archive shards, which is where the wall now sits.

Representation-Change: none

## Where the wall is now

#1977 moved the soak chain off the critical path. Measured on run 31909112992, the first `main` run after it landed, the binding jobs are the two debug-archive test shards:

| job | job time | nextest summary | its heaviest test |
|---|---:|---:|---|
| `Test (3/6)` | 149s | 94.8s | **93.8s** — `merge::tests::direction_symmetry` |
| `Test oracle (2/4)` | 141s | 101.1s | 61.4s — `cis_confluence_axis` |
| `Build` | 242s | — | (starts at t=4, ends 246; the floor after this change) |

`Test (3/6)` was a shard wearing one test: 93.8s of a 94.8s summary. `Test (1/6)`, with a comparable test count, is 26.4s.

## The unit half of #1840, still running twice

That 93.8s test is `normalize::merge::tests::direction_symmetry`. #1955 moved the **integration** half (`issue_1542_direction_symmetry`) to `censuses` and left this one running in **both** shards — 93.8s un-armed in `test` and 85.2s armed in `test-oracle` on run 31909112992, though see **Measured effect** for why no single per-test figure here is reproducible.

**It cannot go to `censuses`, and the way it fails is silent.** That job selects `CENSUS_FILTER` off the soak archive, which is built `-p ferro-hgvs-soak-tests` and contains **no `src/` unit tests at all**. Naming a unit test there negates it from both shards and then selects nothing — deleting the census while turning both shards green *and faster*. `--no-tests=fail` does not catch it, because the step's other terms still match.

So this adds `ORACLE_ONLY_FILTER`, negated from `test` alone. It rests on the subsumption argument `censuses` and `sweeps` already make: an armed run adds assertions at `normalize_core_checked`'s exit and changes no behaviour, so the un-armed copy is redundant while the armed one runs. The subsumption claim is about **assertions**, not cost: in the base run the armed copy was in fact 9% cheaper than the un-armed one, and `clinvar` ran 1.8x the other way — cross-shard timing noise dominates any armed/un-armed effect.

`tests/it/oracle_only_filter_invariant.rs` guards the clause the saving actually rests on — **that `test-oracle` still runs what `test` dropped** — plus the two ways that could be undone (negating the filter there too, or excluding the module via `ORACLE_EXCLUDE`) and the `CENSUS_FILTER` trap above. Nothing else in the repository checks any of it, and each failure direction is one that turns CI green and faster.

## `clinvar_hgvs_tests` by the ordinary route

Second-heaviest test in `Test oracle (2/4)` at 54.9s, moved via `CENSUS_FILTER` as #1955 intended that filter to be used. It is a **parser** benchmark, not a normalization census — it never reaches `canonicalize_from_sequence` — so the soak profile's gains land directly on its gzip decode, JSON deserialization and `parse_hgvs`. It reaches `common::failure_expectations`, now compiled into the driver, and `regex`, now declared by `tests-soak`.

## Measured effect

The PR's own `pull_request` run (`31911822856`) had already produced the answer, so this reports it rather than projecting one.

| | base run `31909112992` | head run `31911822856` |
|---|---:|---:|
| run wall | 297s | **260s** |
| `Build` | 242s, ends 21:23:28 | 216s, ends 22:23:04 |
| last job to finish | `Test (3/6)`, ends 21:24:11 | **`Test oracle (4/4)`, 137s, ends 22:23:31** |
| `Test (3/6)` summary | 94.8s | **37.7s** |
| `Test oracle (2/4)` summary | 101.1s | 87.4s |
| `Test oracle (4/4)` summary | 73.8s | 99.1s |

**~37s of wall — and `Build` is not the floor.** The binding job after this change is `Test oracle (4/4)`, ending **27s after `Build`**, so an earlier draft's "no further carve-out helps until `Build` itself is addressed" does not hold on this PR's own evidence. The `Test (3/6)` prediction was right (37.7s against a predicted ~40s); the ceiling was not.

**A ceiling should not be projected from one run's per-test figures at all.** `merge::tests::direction_symmetry` measured 93.8s un-armed in `Test (3/6)` and 85.2s armed in `Test oracle (3/4)` on the base run, and 46.8s in that same armed shard an hour later — a 2x spread on one test. `Build` itself moved 242 to 216s over the same interval. That is why `ci.yml`'s new comment pins no per-test cost.

One caveat, stated because the point is about method: the base row is a `push` run on `main` and the head row a `pull_request` run on the merge ref, so this is not a perfectly controlled A/B. It does not rescue a ceiling — the 27s overshoot past `Build` is well outside anything the intervening merge accounts for.

I measured that separately and am **not** proposing it: dropping `lto` from the release profile inflates the binary from 8.61 MB to 9.72 MB, i.e. from 82% to 93% of the job's own 10 MB gate, and makes that gate measure a binary that is not the one shipped.

## Verification

- `cargo nextest list` confirms `test(merge::tests::direction_symmetry)` matches the unit test and **not** `issue_1542_direction_symmetry` — the fact that makes this a move rather than a deletion.
- `cargo clippy --workspace --features dev --all-targets -- -D warnings` clean; `cargo fmt --check` clean.
- `cargo build --tests -p ferro-hgvs-soak-tests` compiles (`--tests` matters: the package has no lib or bins, so a plain `cargo build -p` is a no-op that exits 0).
- All 20 selection/membership guards pass, including the four new ones.

One defect worth recording because the diff hides it: inserting the new `#[path]` module split `fixture_gen`'s `#[allow(dead_code)]` from its `#[path]` line, rebinding the attribute to the new module and leaving `fixture_gen` bare. It surfaced as a `dead_code` error in an unrelated module, and was confirmed as caused-here rather than pre-existing by running the same clippy invocation against a pristine checkout of the base.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,8 +88,9 @@ env:
   # WHY THEY MOVED. `--partition hash:K/N` assigns a test by a hash of its NAME,
   # so it balances shard *counts* and is blind to cost — the property already
   # recorded on `sweeps`, where all three exhaustive sweeps hashed into one
-  # shard. These four hashed into the two slowest jobs and made the workflow's
-  # critical path a single test. Measured on run 31845625298 (`044f085a`):
+  # shard. The three below hashed into the two slowest jobs and made the
+  # workflow's critical path a single test. Measured on run 31845625298
+  # (`044f085a`):
   #
   #                                             Test (K/6)   Test oracle (K/4)
   #   issue_1542_direction_symmetry               155.7s (3)         240.3s (1)
@@ -102,6 +103,20 @@ env:
   # test, so re-sharding cannot fix it — which is why `test-oracle`'s own
   # comment below, having chosen N=4 to separate the two bulk corpora, names
   # exactly this carve-out as the robust alternative. This is that alternative.
+  #
+  # The fourth member joined later and its figures are from a DIFFERENT run,
+  # 31909112992 — not comparable row-for-row with the three above, and subject
+  # to the same cross-run spread recorded on `ORACLE_ONLY_FILTER` below:
+  #
+  #   clinvar_hgvs_tests (2 tests)                 37.2s (2,6)         62.3s (2,4)
+  #
+  # Its two tests hash into FOUR different shards between the two matrices, so
+  # ~99s of work leaves them — but only the 54.9s copy in `Test oracle (2/4)`
+  # was on the critical path, which is why the wall-clock saving is much smaller
+  # than the work removed. It is also the only member that is a PARSER
+  # benchmark rather than a normalization census: it never reaches
+  # `canonicalize_from_sequence`, so the optimized archive's gains land on its
+  # gzip decode, JSON deserialization and `parse_hgvs` instead.
   #
   # AND WHY THE JOB, NOT THE JOB BOUNDARY, IS THE WIN. `censuses` consumes the
   # OPTIMIZED archive `test-build-soak` produces, which is where `sweeps`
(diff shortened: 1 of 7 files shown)


nh13:
@coderabbitai pause

PR #2006, opened 2026-08-15
perf(normalize): narrow the alignment DAG's storage to the band

### LINKED ISSUE #1988: perf(normalize): banding left `build` allocation-bound — narrow the alignment DAG's storage to the band
Follow-up to #1928, and the measured next step after the banded DP in #1973. **Do not start before #1973 merges** — it rewrites the same file.

#1973 confines the sequence-first alignment fill to the diagonals a minimal alignment can occupy, and deliberately leaves **storage** at the full `(n+1) x (m+1)` shape. Benchmarked on a `c7i.4xlarge` against its own parent, with a noise floor established first from the same binary against itself (-1.0% to +3.9%):

| n (k=2) | speedup | cell-count ratio |
|---:|---:|---:|
| 256 | 37.0x | 74x |
| 1024 | 6.6x | 293x |
| 4096 | **7.9x** | **1171x** |

**The shortfall is the finding.** ~1171x fewer cells are written at n=m=4096 and it buys 7.9x, because `build` still allocates and zeroes ~101 MB of full-size grids. 101 MB / 20 ms is about 5 GB/s — i.e. the banded run's time is essentially all allocation and no DP.

So banding converted `build` from compute-bound to **allocation-bound**, and narrowing the storage to the band is the lever that remains. The two grids at n=m=4096 hold `2 x 4097^2 = 33,570,818` cells against the band's 28,675 visited.

## Deliberately not claimed

**The resident-memory consequence is UNMEASURED.** The full-size allocation is untouched pages on a 4 kB host and may or may not be faulted in; #1973's module header says so explicitly and this issue does not improve on it. Anyone taking this should measure RSS rather than reason from page sizes — an earlier claim that the unnarrowed allocation "costs nothing in resident memory" was reasoned from a 16 kB page against a 7.7 kB row and inverts on the 4 kB host the timings came from.

There is also ~8.6 bytes/cell of peak memory on #1928 that `ru_maxrss` deltas have not explained; that needs a heap profiler and is not in scope here.

## Scope notes

- `src/normalize/seqfirst/align.rs` is a watched directory, so this needs a `Representation-Change:` declaration. It should be `none` — narrowing storage changes which cells are *stored*, not which lie on a minimal alignment — but that must be demonstrated, not asserted.
- #1973's five differential tests keep the pre-band implementation verbatim as a second implementation; they are the right harness for this too.
- #1973's narrowness guard exists because every differential test passes trivially against a full-width band. A storage change needs the analogous guard, or a regression to full-size storage is green.

nh13:
Closes #1988

Representation-Change: none. Narrowing which cells are *stored* cannot change which lie on a minimal alignment — a cell outside the band lies on none, so reserving no slot for it removes nothing that was ever `true`. Demonstrated rather than asserted; see **Why this is representation-neutral** below.

## What #1973 left behind

#1973 confined the sequence-first DP **fill** to the band and deliberately left **storage** at the full `(n + 1) x (m + 1)` shape. Measured on a `c7i.4xlarge`, that left `build` allocation-bound rather than compute-bound: at `n = m = 4096` the band writes ~1171x fewer cells and bought only **7.9x**, because `build` still allocated and zeroed ~101 MB of full-size grids. 101 MB / 20 ms is ~5 GB/s — essentially all allocation and no DP.

## The change

Every grid over the DAG's cells is now band-major, addressed through a single `BandLayout`: the two cost grids, `on_optimal_path`, `out`, `dominators`'s in-degree table, and `CanonicalAlignment::of`'s `to_go` planes. One layout object means they cannot drift apart. At `n = m = 4096, k = 2` the two retained grids go from `2 * 4097^2 = 33,570,818` cells to `2 * 4097 * 3 = 24,582`.

The stride is **`min(hi - lo, m) + 1`**, not the band's diagonal count `hi - lo + 1`. That `min` is what makes this a narrowing in every shape rather than only the low-divergence one: a `1000 x 1` block has `k = 999` against 2 columns, so a diagonal-count stride would be a ~500x *pessimisation*. Degenerate shapes are not hypothetical here — a long literal `ins` payload arrives as exactly that (#1970) — so there is a guard pinning that storage never exceeds the full grid on five degenerate shapes.

## Why this is representation-neutral

Not asserted — the differential harness is the argument, and it was extended rather than weakened.

- The five `the_band_agrees_with_the_full_grid_*` tests keep the **pre-band implementation verbatim** as a second implementation and diff distance, cell set, out-edges, the `cells()` walk and `dominators()`, over exhaustive small inputs (>3000 pairs, both sides up to length 5 over a 2-letter alphabet), every doubling boundary (`k` = 0,1,2,3,4,5,7,8,9,15,16,17,31,32,33), the shape boundaries, the whole divergence range, and a realistic 512-base block.
- The cell set and edge masks are now compared **over every cell of the full grid through the public accessors**, rather than by diffing the backing `Vec`s. The two builds no longer share a layout — that is the change — so a `Vec` diff would compare *storage* rather than *content*. Reading every grid cell is the stronger question anyway: it additionally pins that a cell the band excludes answers `false` and yields no edges.
- Separately, `multi_member_cis_axis::axis_normalized` (592 real-reference multi-member cis rows, `FERRO_MANIFEST` set) produces a **byte-identical census** on this branch and on its base: `AxisCensus { rows: 592, declined: 6, not_idempotent: 0, unwindowed: 89, sequence_changed: 0, respellable: 461, respelling_converged: 407 }`. See the CI note at the bottom — that test is red on **both** sides for a reason that predates this branch.

## The storage narrowness guard, and its sabotage

#1973 needed a sixth test asserting the band is *narrow* because all five differential tests pass **trivially** against a full-width band. The same hole exists for storage, and is if anything worse: a build that allocated full-size grids and addressed them at `i * (m + 1) + j` satisfies every correctness test here — that is exactly what this module did before, and it was correct then. The only other signal is a criterion run, and CI does not do one.

So `the_storage_is_narrow_at_low_divergence_so_a_no_op_regression_is_caught` pins the **exact** banded size on a fixture whose doubling ladder is fully determined (two spaced substitutions, so `delta = 0` seeds `k = 1`, the one-diagonal band reports distance 2, and `k` doubles once to 2 and converges — a three-diagonal band).

**Verified by sabotage, not assumed.** Pinning `BandLayout::new`'s stride to `alt_len + 1` — behaviour-preserving, since a wider stride only leaves unused gaps between rows — leaves the `seqfirst` modules at **42 passed, 1 failed**, that guard being the one failure:

```
assertion `left == right` failed: on_optimal_path holds 1050625 cells for a
3-diagonal band over 1025 rows; expected 3075, and the full grid would be 1050625
```

All five differential tests, the brute-force dominator oracle and #1973's band-narrowness guard stay green under that sabotage — which is the whole point of adding it.

## Peak RSS: measured on 16 kB pages, still unmeasured on 4 kB

#1988 records the resident-memory consequence as UNMEASURED and asks for a measurement rather than a page-size argument. Measured, on one platform only, and scoped to it:

| | peak RSS | reps |
|---|---:|---|
| base `c9207d7e` | 106,758,144 B (101.8 MiB) | 4, spread 0.015% |
| this branch | 6,012,928 B (5.73 MiB) | 4, spread 0.14% |

arm64, 16 kB pages, release profile, a throwaway probe doing one `AlignmentDag::build` at `n = m = 4096, k = 2` and nothing else; both arms exit 0 and print `k=2`, so the runs did the work rather than dying early. A **17.8x reduction, ~96 MiB.** The "before" figure lands within 1% of the ~101 MB the grid arithmetic predicts, which is what says the full-size allocation really was resident on this page size. The 5.73 MiB that remains is the process floor — binary, runtime, the two 4 KiB blocks — not the DAG, whose grids are now ~36 KiB.

**The 4 kB-page host is not measured**, and it is where the geometry predicts the *least* to reclaim: a row spans ~2 pages and the band is diagonal, so row `i` touches only one of them and much of each grid may never have been faulted. Do not carry the figure above onto that platform. The probe was not committed.

**No speedup is claimed.** The 7.9x is #1973's, on its host with criterion. Local timings here are not trustworthy for a speedup claim — #1973's own header records the same computation measuring 5.8x apart on allocator state alone — so none is offered.

## Verification

Full gate with `FERRO_REQUIRE_BULK_FIXTURES=1` and `FERRO_MANIFEST` pointed at the prepared reference: **11061 tests run, 11060 passed, 1 failed, 0 `Skipping:` lines**. `cargo fmt --all --check` and `cargo clippy --all-features --all-targets -- -D warnings` both clean. Python suite (which `cargo nextest` does not reach) **1059 passed, 8 skipped**.

The one failure is `multi_member_cis_axis::axis_normalized`, and it is **pre-existing on the base commit**: reverting only this branch's source files onto `c9207d7e` and re-running it reproduces the identical census and the identical failure (`unwindowed: 89` against a pinned `90`). That test is reference-manifest-gated, so PR CI — which never sets `FERRO_MANIFEST` — skips it and reports green on both sides. Flagging it rather than folding a pin change into a performance PR.

### DIFF
--- a/benches/seqfirst_align.rs
+++ b/benches/seqfirst_align.rs
@@ -1,20 +1,31 @@
 //! Sizing the sequence-first DP.
 //!
-//! `AlignmentDag::build` **allocates** `O(n*m)` and, since the band landed,
-//! **fills** only `O((n+m)*k)`. Criterion has no memory axis, so this measures
-//! wall-clock time only, and the block-size cap in the migration is chosen on
-//! time evidence.
+//! `AlignmentDag::build` **fills** `O((n+m)*k)` since the band landed (#1928),
+//! and since #1988 **allocates** `O((n+m)*k)` too. Criterion has no memory axis,
+//! so this measures wall-clock time only, and the block-size cap in the
+//! migration is chosen on time evidence.
 //!
 //! The memory dimension is **not** measured here; the figures are on #1928, and
 //! note they moved when the cost grid narrowed to `u16`: **14.60 bytes per grid
 //! cell, 215.4 MB** at the widest accepted block (`W = 3840`), down from 18.67
 //! and 275.4 MB. This header carried the older pair until the banding change.
+//! **Both figures predate #1988** — they are per *full-grid* cell — and neither
+//! has been retaken for **this benchmark's shape**, the widest accepted block at
+//! `W = 3840`. Do not read the narrowed allocation as a measured RSS figure for
+//! it.
 //!
-//! Because the allocation is unnarrowed while the fill is narrowed, a timing
-//! here is partly an allocator measurement: at `n = 4096` the banded arm runs at
-//! roughly memory bandwidth, and two groups that compute the *same* thing can
-//! differ several-fold purely on whether a large arena has already been faulted
-//! in. Compare each group against its own baseline, never across groups.
+//! #1988 *did* measure peak RSS, but at `n = m = 4096, k = 2` and on 16 kB pages
+//! only, where it read **101.8 MiB -> 5.73 MiB**. That is a different shape and
+//! one page size, so it does not transfer here in either direction. The figure
+//! lives on `AlignmentDag::build` in `src/normalize/seqfirst/align.rs`, together
+//! with the limits that qualify it, and that is where to read it.
+//!
+//! While the allocation was unnarrowed and the fill was not, a timing here was
+//! partly an allocator measurement: at `n = 4096` the banded arm ran at roughly
+//! memory bandwidth, and two groups that compute the *same* thing differed
+//! **5.8x** purely on whether a large arena had already been faulted in.
+//! Narrowing the storage removes the cause, not the discipline: compare each
+//! group against its own baseline, never across groups.
 //!
 //! # Divergence is an AXIS, not a constant (#1928)
 //!
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2181,10 +2181,9 @@ const MAX_TWO_GAP_PLACEMENTS: usize = 65_536;
 
 /// Largest alignment grid the sequence-first splitter will build, in cells.
 ///
-/// Every pass over the DAG is `Θ(n·m)` in **space** as well as time. Only the
-/// reference side is bounded upstream ([`MAX_CANONICAL_WINDOW`]); the alternate
-/// side is whatever the members produce, and a duplication over the whole window
-/// doubles it.
+/// Only the reference side is bounded upstream ([`MAX_CANONICAL_WINDOW`]); the
+/// alternate side is whatever the members produce, and a duplication over the
+/// whole window doubles it.
 ///
 /// # What the budget actually pays for
 ///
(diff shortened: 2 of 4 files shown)


nh13:
@coderabbitai pause

Issue #2007, opened 2026-08-15
adjudication: an inversion with an abutting substitution is re-partitioned into a shape complex.md:113 does not publish


nh13:
## The question

An inversion with a substitution abutting its 3' break point is re-partitioned into a third shape — neither the published split nor a single merged `delins`. Measured:

```
in   NC_TESTWIN.1:g.[302_310inv;311G>A]
out  NC_TESTWIN.1:g.[302del;304_309inv;311delinsAA]
```

**The sequence is preserved.** That is asserted through the independent apply oracle, not inferred, so this is a question about form and not a wrong answer.

## Why it is not obviously wrong

`canonical-form-choice-when-both-legal` (decided 2026-08-07) rules that the canonical form is derived from the **resulting sequence** — emit the form that falls out, do not preserve the input's spelling. On that rule alone the output above is the rule working as decided.

## Why it is not obviously right

It collides with a published worked example. `DNA/complex.md:113` publishes an inversion with a substitution at its break point kept as **two members**, and `:117` publishes another zero-separation pair split the same way. Meanwhile `DNA/delins.md:86-89` mandates **merging** a different zero-separation pair (`c.[2077G>A;2077_2078insTA]` is invalid; "the correct description of this variant is `c.2077delinsATA`").

So the spec publishes both a split and a merge for zero-separation pairs and does not scope one against the other. Ferro currently picks a third option — re-partitioning into three members whose boundaries match neither.

## What a ruling needs to settle

1. Does `canonical-form-choice-when-both-legal` reach this case, or does a published worked example outrank a derived form? `adjudication-precedence-order` puts spec-explicit above our judgement, and `complex.md:113` is spec-explicit — but it is an *example*, not a stated rule, and the ledger has previously distinguished the two.
2. If re-partitioning at this scale is endorsed, **what is `complex.md:113` then taken to mean?** A ruling that leaves the clause unexplained will be re-argued.
3. If the published split wins, what bounds it? "Preserve the input's spelling when it matches a worked example" is not obviously implementable, and it partially reverses the 2026-08-07 rule.

## Guard

`tests/it/insertion_adjacency_defects.rs::policy_an_abutting_inv_and_sub_are_repartitioned` asserts the published expectation and is `#[ignore]`d against this issue. Whichever way the ruling goes, that test is what should change — and the ruling belongs in `hgvs_spec_normalization_overrides.json`, not only in a PR body.

Issue #1996, opened 2026-08-15
normalize: a dup and its equivalent repeat settle differently beside an interior insertion


nh13:
## Summary

A duplication and the equivalent exact repeat, each beside an insertion **interior to their own span**, are both accepted and settle on **different forms**. The `dup` spelling merges into one member; the `repeat` spelling stays as two. That is the #1437 disagreement — one shape, two spellings, two answers — in the one place `tests/it/issue_1749_one_overlap_definition.rs` could not see it, because its agreement test compares the two *verdicts* and not the two *forms*.

Measured on `origin/main` at `de8532ee`.

## Reproducer

Fixture: `NC_TEST.1 = TTTTTTTTTAATTTTTTTTGTGTGTTTTTTTTTTTTTTT` (39 bp; `AA` at 10-11). Strict mode, `MockProvider`, no reference needed.

```
g.[10_11A[4];10_11insT]  ->  g.[10_11insT;10_11A[4]]     two members
g.[10_11dup;10_11insT]   ->  g.10_11insTAA               one member
```

`10_11A[4]` and `10_11dup` denote the same edit — the tract at 10-11 is `AA`, so `A[4]` adds one `AA` copy exactly as the `dup` does. The sibling insertion is at junction 10|11, interior to the 10-11 span in both spellings. Same shape, same sibling, two outputs.

Both are accepted, so this is not an overlap-detection disagreement — the detector already agrees. The divergence is downstream, in whether the pair is merged.

## Why it was invisible

`the_dup_and_repeat_spellings_of_one_shape_agree` asserts

```rust
assert_eq!(strict_accepts(repeat), strict_accepts(dup), ...);
```

Comparing verdicts is only a real check when the verdicts can differ *and* agreement implies sameness. Two spellings that both **refuse** agree trivially, and two that both **accept** can still settle on different forms. So the assertion passes on a genuine disagreement.

The rows in that test were additionally mispaired — `10_11A[3]` was compared against `10_11dup`, and those denote different sequences (`A[3]` adds one base to the `AA` tract, `dup` adds two). The boolean comparison could not surface that either.

## Not the same as the shared-junction case

Worth separating, because the geometries are adjacent and the verdicts differ:

| sibling insertion | geometry | status |
|---|---|---|
| `11_12insT` | at the dup/repeat's own **write junction** | ordered by `duplication.md:90`'s "followed by"; both spellings settle on `g.11_12insAAT` |
[...]

PR #2008, opened 2026-08-15
feat(normalize): accept a duplication and an insertion at one junction

nh13:
A cis allele naming a duplication and an insertion at one junction was refused as having no canonical form. `DNA/duplication.md:90` publishes exactly that geometry and orders it in prose, so it does.

## The clause

```
NC_000001.11(NM_206933.2):c.[675-542_1211-703dup;1211-703_1211-702insGTAAA]
  a duplication of the sequence from nucleotide position c.675-542 to
  c.1211-703, followed by the insertion of the sequence GTAAA.
```

The insertion sits at `1211-703_1211-702`, which is the junction 3' of `1211-703` — the duplication's own write junction. "Followed by" supplies the composition order, and that order is the genomic order `alleles.md:118` already requires members to be listed in. The pair is ordered, not ambiguous.

This contradicts the premise #1276 reasoned from — "a dup copy and an insertion competing for one junction have no defined order" — which is why the ruling record says so explicitly rather than silently replacing it.

## What changed

Three sites conflated *how many members write at this junction* with *how many members claim a base here*. An insertion is interbase and consumes no reference base, so only one member of the pair ever claims one. Two insertions at one junction are still refused, because both write at the same interbase point and nothing orders them.

The predicate is kind-generic, so a growing repeat — which has a duplication's write geometry (#1437) — is accepted on the same footing.

## Measured

| input | before | after |
|---|---|---|
| `c.[5_6dup;6_7insA]` | refused W5002 | `c.6_7insAAA` |
| `g.[20_25GT[4];25_26insA]` | refused W5002 | `g.25_26insGTA` |
| `g.[10_11dup;11_12insT]` | refused W5002 | `g.11_12insAAT` |
| `g.[10_11A[4];11_12insT]` | refused W5002 | `g.11_12insAAT` |

The last two are one shape in two spellings and now reach an **identical form**, not merely an identical verdict. `insGTA` and `insAAT` are both the copied bases followed by the inserted ones — `duplication.md:90`'s order.

Both cis-confluence censuses are unchanged, and an ordering asymmetry is fixed in passing: the two member orders previously produced different diagnostics (`g.200_201` against `g.200`) and now settle on one description.

## Test changes worth review

The `issue_1749` agreement test compared the two spellings' **verdicts**. That is a real check only while the verdicts can differ — two spellings that both refuse agree trivially, and two that both accept can still settle on different forms. It now compares forms for the shared-junction pair.

Strengthening it surfaced two things:

- Its rows were **mispaired**. `10_11A[3]` adds one base to the `AA` tract at 10-11 where `10_11dup` adds two, so they never denoted one shape. Corrected to `A[4]`.
- An **existing divergence**, filed as #1996: when the sibling insertion is *interior* to the duplicated span rather than at its write junction, the `dup` spelling merges and the `repeat` spelling does not. It reproduces byte-for-byte on `origin/main`, so it is not fallout from this change; it is pinned as measured so the fix has somewhere to land.

Three pins are `#[ignore]`d against the issues that must answer them (#2007, #1946 x2) rather than rewritten to assert current behaviour — the desired behaviour is real and a pin should not redefine it as correct.

## Follow-up

The **form** question — whether the accepted pair should settle merged or split — is deliberately left undecided in the ledger. It cannot be settled where the choice is currently made, per member and before the member set is final. That is #1946.

Representation-Change: a cis allele naming a duplication (or an equivalent growing repeat) and an insertion at one shared junction now settles on a single insertion whose payload is the copied bases followed by the inserted ones — `c.[5_6dup;6_7insA]` becomes `c.6_7insAAA`, and `g.[20_25GT[4];25_26insA]` becomes `g.25_26insGTA`. The effect depends on the error mode and both halves are a migration. Under `ErrorMode::Strict` the shape was refused with W5002, so those are previously-rejected descriptions that now normalize and no previously-accepted description changes form. Under the library default, which is lenient (`NormalizeConfig::default()` uses `ErrorConfig::lenient()`, `src/normalize/config.rs:129`), W5002 was only a warning, so the same allele was already accepted and its normalized string moves: measured on this branch against its merge-base, `NM_TEST.1:c.[5_6dup;6_7insA]` normalized to `NM_TEST.1:c.[5_6dup;6_7insA]` (preserved as authored) and now normalizes to `NM_TEST.1:c.6_7insAAA`. A default-config consumer that keys stored data on the normalized string therefore has a real migration for this shape. Both cis-confluence censuses are byte-identical. Two insertions at one junction remain refused. A same-junction pair whose members were listed in the two possible orders previously produced two different diagnostic strings and now produces one description.


nh13:
@coderabbitai pause

PR #2010, opened 2026-08-15
ci: bound each archive run, and name the phase a stall stopped in

### LINKED ISSUE #2000: `Test (2/6)` hangs indefinitely and blocks the merge queue — no `timeout-minutes` anywhere in ci.yml
`Test (2/6)` hangs indefinitely — twice in three hours, and with no `timeout-minutes` it blocks the merge queue for six hours

## What happens

The `Run Rust tests (shard 2 of 6)` step stops making progress and never returns. The job is not slow; it is stopped. Two instances, both on `Test (2/6)` specifically:

| run | event / branch | job duration | how it ended |
|---|---|---|---|
| [`31905255274`](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/31905255274) attempt 1 | `pull_request`, `perf/issue-1964-soak-driver-package` | **4380 s** (nextest step 4351 s = 72.5 min) | cancelled |
| [`31910146014`](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/31910146014) | `merge_group`, `gh-readonly-queue/main/pr-1984-d055fc10` | **~21 min and climbing** | cancelled by hand to unblock the queue |

Baseline for the same shard, from the 16 most recent completed CI runs:

```
89 79 75 85 73 74 70 72 76 75 89 77 77 94 73  (seconds)
```

So 70–94 s is normal, and both incidents are **45–60x** that. On run `31905255274` the other five shards in the same matrix finished in 68–134 s while `(2/6)` ran for 73 minutes, so it is not a runner-wide slowdown.

It is also transient rather than a property of the commit: attempt 2 of `31905255274` ran the same shard on the same tree in **91 s**.

## Why it is worse than a slow job

**`ci.yml` on `main` carries zero `timeout-minutes`** — `git show origin/main:.github/workflows/ci.yml | grep -c timeout-minutes` returns `0`. So nothing stops a hung job except GitHub's default **6-hour** limit.

When this happens inside a **merge group** it blocks the entire queue for those six hours. That is what the second instance did: `Test (2/6)` hung at the head of a four-deep queue, and #1976, #1793 and #1981 sat behind it with their own speculative merge groups already green. The queue only moved after the run was cancelled by hand.

There is no alerting on this. The job reads `in_progress`, which is indistinguishable from a healthy long-running job unless you know the baseline.

## Two things this is not

- **Not the 6-hour limit doing its job.** Six hours of a blocked merge queue is not an acceptable failure mode, and the hang is the defect regardless of what eventually reaps it.
- **Not fixed by PR #1994.** That PR adds `timeout-minutes` to every PR-gating job and is worth landing on its own, but a timeout converts an indefinite hang into a *failed* job. It bounds the blast radius; it does not stop the hang. This issue is for the hang itself.

## Reading the evidence — a trap worth knowing

`/repos/{owner}/{repo}/actions/runs/{id}/jobs` returns **only the latest attempt**. Run `31905255274`'s default job list shows `Test (2/6)` at `91s success`, because attempt 2 is what it reports; the 4380 s cancelled attempt is visible only under:

```
/repos/fulcrumgenomics/ferro-hgvs/actions/runs/31905255274/attempts/1/jobs
```

A census that walks the default endpoint will therefore miss every occurrence of this bug — the hung attempts are precisely the ones it hides. Two separate investigations hit this and nearly concluded the incident was unreproducible.

## What would help most

1. **Find out where it stops.** The step log is truncated at cancellation, so the immediate need is to know whether nextest is stuck inside a test, between tests, or in the archive/extract step. Enabling `--status-level` output or a periodic stack dump on the shard would narrow it.
2. **Suspect list, in the order I would check it.** Shard 2's assignment is deterministic per-run, so "which tests land in shard 2" is answerable — start there and look for anything with a lock, a channel, a `OnceLock`, a rayon pool, or a provider fetch. This repo runs several `proptest` modules and three exhaustive sweeps; a proptest shrink loop on an unlucky seed is a plausible shape for an unbounded stall that vanishes on the next attempt.
3. **A per-test timeout, not only a per-job one.** `nextest`'s `slow-timeout` with `terminate-after` turns a stuck test into a named failing test, which is far more actionable than a cancelled job — and it would have identified the culprit on the first occurrence.

## Reproduction status

Not reproducible on demand. Two occurrences in roughly three hours, both transient, both on the same shard index. Recording it now because the evidence for the second one was about to be lost to the cancellation, and because the default API endpoint hides the first.

nh13:
Closes #2000.

### The measurement, and what it refutes

The issue's leading question was *"find out where it stops"*. It stops **before any test runs**, and that answer invalidates all three of the fixes it proposed.

Both hung jobs' logs were pulled and read against a healthy run of the same shard. All three stop at the same line:

```
healthy   19:57:40  Extracted 66 files ... in 1.09s
          19:57:40  Nextest run ID ... with nextest profile: default
          19:57:40  Starting 1794 tests across 24 binaries          <- 0.101s after Extracted
          19:57:40  PASS [0.009s] (1/1794) ...

hung #1   19:57:40  Extracted 66 files ... in 0.82s
          21:10:01  ##[error]The operation was canceled.            <- 72 minutes, nothing between

hung #2   21:44:29  Extracted 66 files ... in 1.08s
          22:05:38  ##[error]The operation was canceled.            <- 21 minutes, nothing between
```

`Starting N tests across M binaries` is printed once nextest has enumerated every archived binary — by executing each one with `--list` — and *before* any test body runs. Neither stall reached it, nor even the `Nextest run ID` line above it. So no test ever started: both stalls are in nextest's **binary-list phase**.

The runner's own orphan reaping agrees, and names the culprit:

```
hung #1   Terminate orphan process: pid (2656) (cargo-nextest)
          Terminate orphan process: pid (2703) (it-f2431c9a8062b86c)
hung #2   Terminate orphan process: pid (2626) (cargo-nextest)
          Terminate orphan process: pid (2668) (ferro_hgvs-686e43b9c82c0b52)
```

`cargo-nextest` plus **exactly one** test binary each time — and a *different* binary each time.

Three consequences, each contradicting a suggestion in the issue:

| the issue suggested | why it cannot work |
|---|---|
| a per-test `slow-timeout` with `terminate-after` | it governs test *execution*, which never began |
| `--status-level` output | it reports on tests, of which there were none |
| "which tests land in shard 2" | listing enumerates **every** binary identically on **every** shard; `--partition` and `-E` are applied to the resulting list, so no shard lists less than any other — and the two stalls were on different binaries, so it is not a property of one binary's contents either |

Shard 2 appearing twice is therefore a coincidence of two samples, not a signal, and the deterministic per-run shard assignment the issue proposed starting from does not narrow anything here.

### What this ships, and what it does not

**It does not fix the hang.** The root cause is not established — one child of `cargo-nextest` stops returning, transiently, on a tree that ran in 91 s on the very next attempt. I could not reproduce it on demand and I am not claiming to have closed it.

What it buys is the two things that were actually missing:

1. **The stall is bounded in minutes**, not by GitHub's six-hour default — which is what held a four-deep merge queue.
2. **The next occurrence names its own phase**, so the localisation above does not have to be re-derived from two cancelled logs a third time.

All six archive-consuming runs (`test`, `test-oracle`, `soak`, `sweeps`, and both `censuses` steps) now go through `scripts/run_nextest_archive.sh`. The list phase is identical in all of them, so scoping this to `test` would leave the other five able to reproduce it silently.

On a timeout the wrapper emits exactly one of three annotations:

```
::error title=nextest stalled BEFORE running any test::  ... this is the #2000 shape ...
::error title=nextest stalled DURING test execution::    ... this is NOT the #2000 shape ...
::error title=nextest stalled, phase undetermined::      ... the capture itself failed; do NOT record as #2000 ...
```

Live output is unchanged — the run is `tee`d, so what a reviewer reads is byte-identical to an unwrapped run.

### Relationship to #1994

Complementary, not overlapping, and #1994 is still worth landing on its own. #1994 sets `timeout-minutes` per **job**, which cancels the job with no explanation. This bounds the **step** and says which phase died. If both land, the step bound fires first and is the one that explains itself.

### Four defects found by running the wrapper and reviewing it — not by reading it once

Every one of these is invisible on a green local run, which is the point:

- **nextest pluralises.** A one-test selection prints `Starting 1 test across 1 binary`. The first pattern required `tests`, so an end-to-end run of a single-test selection was classified as a list-phase stall — a false #2000 on a run that had plainly started. The singular form is exactly what a bisected or `-E`-narrowed re-run produces, i.e. when someone is most likely reading this output.
- **`mktemp .../nextest-run.XXXXXX.log` fails on Linux.** The `X`s must end the template: GNU/BusyBox `mktemp` errors (`mktemp: : Invalid argument`, confirmed in a Linux container), so **every archive step would have died on that line**. BSD `mktemp` instead returns the template verbatim, giving concurrent runs one shared fixed filename — so macOS was green and said nothing about the platform CI uses.
- **An unreadable log classified as `list-phase`.** A missing or empty capture grep-misses, so a `mktemp`/`tee` failure would have been reported as *the very shape this PR exists to identify* — a fabricated #2000 with no run behind it. There is now a third verdict, `unknown`, with its own annotation.
- **Empty-array expansion under `set -u` breaks on bash 3.2.** The natural spelling of the `timeout(1)` fallback is `RUNNER=(timeout …)` / `RUNNER=()` plus `"${RUNNER[@]}"` — and expanding the empty array is an `unbound variable` error on bash 3.2, which is `/bin/bash` on macOS. The fallback would have died on precisely the platform it exists to serve. `shellcheck` does not flag it; measured against `/bin/bash` 3.2.57. It is a function now, and a guard forbids `[@]`/`[*]` in the script.

The guard also replaced its own `wrapped >= 6` floor with a **derived** set, per `CLAUDE.md`'s "Assert the property. Measure the count. Never let a count BE the property" — and that change immediately paid for itself: it caught that `--archive-file` sits on a **backslash continuation** of `cargo nextest archive` in the soak-build step, so a line-based read files an archive *creation* as an unwrapped run. Continuations are folded before the partition.

### Unchanged on purpose

- `.config/nextest.toml` is untouched. Its comment explains why a `[profile.ci]` with a round-number `slow-timeout` killed the censuses; this change adds no per-test bound, because the measurement shows a per-test bound cannot see this failure.
- `triples_are_disjoint`-style behaviour, test selection, filters, shard count: all untouched. This adds no test and removes none — `11076 tests run: 11076 passed`.
- The wrapper is transparent on the happy path: verified that a clean run exits 0 and emits no annotation, and that a genuinely failing run propagates its own exit status (4) with no annotation.

### Verification

```
cargo fmt --all --check                                    OK
actionlint .github/workflows/ci.yml                        OK
shellcheck scripts/run_nextest_archive.sh                  OK
cargo clippy --all-features --all-targets -- -D warnings   CLIPPY_EXIT=0
cargo nextest run --features dev --no-fail-fast            11076 passed, 0 failed, 151 skipped
scripts/run_oracle_suite.sh                                10919 passed, 0 failed
```

The guard is 10 tests. The original five were **watched fail and then pass** before the wrapper existed (TDD); the other five were added afterwards, each from a defect found by running or reviewing the wrapper, and each is a case that was failing when it was written. Three were additionally re-checked by deliberate sabotage:

| sabotage | test that goes red |
|---|---|
| revert the pattern to the plural-only `tests across` | `the_classifier_reads_nextests_singular_announcement` |
| restore the `.log` suffix after the `mktemp` placeholder | `the_temp_template_ends_in_its_placeholder` |
| return one call site to a bare `cargo nextest run` | `no_archive_run_bypasses_the_wrapper` (reported `1 of 6`) |

The classifier is pinned against the **recorded bytes of the real runs**, not a paraphrase. `ci.yml` sets `CARGO_TERM_COLOR: always`, so the line arrives as

```
\e[32;1m    Starting\e[0m \e[1m1794\e[0m tests across \e[1m24\e[0m binaries (...)
```

and `grep 'Starting [0-9]* tests across'` over the raw log returns **0 on a healthy run** — which would report every genuine in-test hang as a false #2000. That is why the fixtures carry their escape codes and why the guard asserts they still do.

End-to-end classification was checked against forced timeouts in both directions, not only against fixtures:

| forced condition | exit | annotation |
|---|---|---|
| `NEXTEST_STALL_TIMEOUT=6s` on a 26 s single test | 124 | `stalled DURING test execution` |
| clean run | 0 | none |
| genuinely failing run | 4 | none |

Representation-Change: none. CI configuration, a shell wrapper and a new test module; no file under `src/` is touched, so no normalized output can move.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1473,7 +1473,7 @@ jobs:
         # wreck shard balance: measured locally, the shard holding them took
         # 66.6s against 8.8-13.5s for its siblings.
         run: |
-          cargo nextest run --archive-file nextest.tar.zst --workspace-remap . \
+          scripts/run_nextest_archive.sh --archive-file nextest.tar.zst --workspace-remap . \
             --extract-to . --extract-overwrite \
             --partition hash:${{ matrix.shard }}/6 \
             -E "not test(proptest) and not ($SWEEP_FILTER) and not ($CENSUS_FILTER) and not ($ORACLE_ONLY_FILTER)"
@@ -1718,7 +1718,7 @@ jobs:
           # As in `test` above: an absent bulk fixture fails rather than skips.
           FERRO_REQUIRE_BULK_FIXTURES: "1"
         run: |
-          cargo nextest run --archive-file nextest.tar.zst --workspace-remap . \
+          scripts/run_nextest_archive.sh --archive-file nextest.tar.zst --workspace-remap . \
             --extract-to . --extract-overwrite \
             --partition hash:${{ matrix.shard }}/4 \
             -E "not test(proptest) and not ($SWEEP_FILTER) and not ($ORACLE_EXCLUDE) and not ($CENSUS_FILTER)"
@@ -1826,7 +1826,7 @@ jobs:
         # determine behavior", and a job that runs a million cases must not be
         # one default-change or one profile setting away from a green no-op.
         run: |
-          cargo nextest run --archive-file nextest-soak.tar.zst --workspace-remap . \
+          scripts/run_nextest_archive.sh --archive-file nextest-soak.tar.zst --workspace-remap . \
             --extract-to . --extract-overwrite --no-tests=fail -E 'test(proptest)'
 
   # The three exhaustive-sweep modules, moved off the `test` / `test-oracle`
@@ -1939,7 +1939,7 @@ jobs:
         # module is therefore run twice on purpose: unarmed in `test`, armed
         # here. It is ~10 cheap unit tests, no sweep and no provider corpus.
         run: |
-          cargo nextest run --archive-file nextest-soak.tar.zst --workspace-remap . \
+          scripts/run_nextest_archive.sh --archive-file nextest-soak.tar.zst --workspace-remap . \
             --extract-to . --extract-overwrite --no-tests=fail \
             -E "$SWEEP_FILTER + test(issue_1615_denoted_sequence_oracle)"
 
@@ -2020,7 +2020,7 @@ jobs:
         # this step names its modules explicitly, so a rename is exactly the
         # edit that would drop a census from CI silently.
         run: |
-          cargo nextest run --archive-file nextest-soak.tar.zst --workspace-remap . \
+          scripts/run_nextest_archive.sh --archive-file nextest-soak.tar.zst --workspace-remap . \
             --extract-to . --extract-overwrite --no-tests=fail \
             -E "test(issue_1542_direction_symmetry) + test(normalize_axis_preserving) + test(clinvar_hgvs_tests)"
       - name: Run the census that may not be measured with an oracle armed
@@ -2028,7 +2028,7 @@ jobs:
         # module builds its corpus in code (`conformance::spec_corpus`) and
         # reads nothing from disk.
         run: |
-          cargo nextest run --archive-file nextest-soak.tar.zst --workspace-remap . \
+          scripts/run_nextest_archive.sh --archive-file nextest-soak.tar.zst --workspace-remap . \
             --extract-to . --extract-overwrite --no-tests=fail \
             -E "test(spec_conformance_axis)"
 
--- a/.github/workflows/nightly-soak.yml
+++ b/.github/workflows/nightly-soak.yml
@@ -171,8 +171,16 @@ jobs:
         # `--no-tests=fail` for the reason spelled out on ci.yml's `soak` job,
         # which runs the identical selection against the identical archive. A
         # nightly is the worse place to lose it: nobody watches a green run.
+        #
+        # Through `run_nextest_archive.sh` for the same reason, and the nightly
+        # is the worse place to lose that too: this job carries no
+        # `timeout-minutes`, so an unbounded list-phase stall here would run to
+        # GitHub's 360-minute default with nothing on the run page saying why.
+        # The 25m step bound is ~26x the measured job wall clock (43-57s across
+        # all 8 shards of the last three nightlies), so it cannot fire on a
+        # legitimate run.
         run: |
-          cargo nextest run --archive-file nextest-soak.tar.zst --workspace-remap . \
+          scripts/run_nextest_archive.sh --archive-file nextest-soak.tar.zst --workspace-remap . \
             --extract-to . --extract-overwrite --no-tests=fail -E 'test(proptest)'
       - name: Capture the failing seed
         # proptest writes the shrunk case to its persistence file. Surfacing it
--- a/scripts/run_nextest_archive.sh
+++ b/scripts/run_nextest_archive.sh
@@ -1,0 +1,176 @@
+#!/usr/bin/env bash
+#
+# Run `cargo nextest run` under a wall-clock bound, and — when that bound fires —
+# name the phase the run stopped in.
+#
+# WHY (#2000). `Test (2/6)` stalled twice in three hours. Both job logs stop at
+# the same line, and a healthy run of the same shard shows what should follow:
+#
+#     healthy   Extracted 66 files ... in 1.09s
+#               Starting 1794 tests across 24 binaries      <- 0.101s later
+#     stalled   Extracted 66 files ... in 0.82s
+#               (nothing, for 72 minutes)
+#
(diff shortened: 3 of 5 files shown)


nh13:
@coderabbitai pause

PR #2011, opened 2026-08-15
feat(normalize): derive ins<range>inv for an inverted duplication

nh13:
Stacked on #2008 — review that one first; this PR's diff against it is `src/normalize/{mod,rules}.rs`, three form pins, and one ledger record.

An insertion whose payload is the reverse complement of a nearby reference span was emitted as a literal. `DNA/inversion.md:69` spells it `ins<range>inv`, naming the span the copy came from.

```
g.221_222insGCCGACTCCC   ->   g.221_222ins212_221inv
```

`:69` gives **both** spellings — one where the source span ends at the junction and one where it starts there — distinguished by which side of the original the inverted copy falls, so implementing only one would reintroduce the asymmetry the clause exists to remove. Both are pinned.

## The coincidence floor is house policy

The spec supplies no minimum length: `duplication.md:5` admits "one or more nucleotides", and the only length statement in play is `inversion.md:5`'s "more than one nucleotide", which excludes a one-base payload definitionally. **Nothing is cited for the threshold**, and the ledger record says so.

A flat minimum was implemented first and ablated: **34 / 19 / 17** failures at minimums of 1 / 2 / 3 bases, with sequence-preservation guards surviving every threshold — in a skewed window a short match is far likelier than `4^-n`. Replaced by a composition-aware chance-match probability, which takes the collateral to **zero**.

Its constant is `1/N` over the representation corpus and **N is hardcoded**, so the justification drifts as the corpus grows. Read the size or state it as an order of magnitude; do not quote the constant.

## This costs a confluence class — please read before approving

`converged` falls by one in **both** directions (3': 11 271 → 11 270; 5': 11 272 → 11 271). The module's ratchet forbids that without a declared re-bless, so the re-bless is declared here and in `cis_confluence_axis.rs`.

The class is `s03-g-m4-sep1-p8-all-ins`, the same one in both directions:

```
g.[264_265insA;266_267insAGTAATTGAGTAA;267_268insGGAGTAACGCAGTAACGA]
g.[264_265insATGAGTAA;265_266insTGAGTAAC;266_267insGAGTAACG;267_268ins268_275inv]
```

Three members against four — they differ in **member geometry**, not only in how the last member is spelled. And they **converge on `origin/main`**, so the derivation changes a trajectory rather than re-spelling a settled member.

**Where it does that is not traced, and I have not offered an explanation.** The re-bless section records it as an open question. It is the shape #1946 describes — a presentational form minted per member, before the member set is final — but saying so is an analogy, not a diagnosis, and it should not be read as one.

**One candidate mechanism was tested and refuted.** The coincidence floor originally keyed on the whole normalization window, whose extent is `NormalizeConfig::window_size` — a public field — which made the emitted form depend on a caller-set knob: on an AT->GC gradient one variant rendered a literal at `window_size = 80` and `ins<range>inv` at 120. Since members resting at different places get different windows, that looked like a plausible explanation for the lost class. It is not: with `COMPOSITION_HALF_WIDTH` fixing the estimate at 50 bases either side of the junction, both censuses still measure 11,270/2 and 11,271/1. **The mechanism remains unexplained, but it is not the window-extent dependence** — recorded in the ledger and in `cis_confluence_axis.rs` so it is not re-raised.

**A pre-existing VCF defect this change would have made reachable, fixed here.** `build_vcf_record` built an insertion's ALT allele from the payload's `Display`, which for a range payload is its HGVS spelling — `g.221_222ins212_221inv` emitted `ALT=C212_221inv`. Until now every construction site of `PositionRangeInv` was in the parser (`src/hgvs/parser/edit.rs:376`, `:843`, `:992`), so the malformed allele was reachable only from user-authored input; this derivation makes the normalizer mint the shape. The converter now **declines** a payload it cannot spell as bases rather than resolving it, because `build_vcf_record` is the shared trunk for all four axes and receives the edit in the description's own coordinate frame — resolving a `c.`-frame range against the genomic contig would return the wrong bases silently, which is worse than the visible corruption. Also: the ledger's claim that the form "must not ship ahead of the SPDI-side support" was already false — #1966 is in this branch's base and both spellings resolve to identical SPDI — and is withdrawn.

If a reviewer would rather not take a convergence regression at all, the right outcome is to hold this PR behind #1946 rather than weaken the ratchet. #2008 does not depend on it.

## Also not done

Wired for `g.` only. `c.`/`n.`/`r.`/`m.` still emit literals, so one inverted duplication spells differently per axis. The long-term home for that is #1946's render stage.

Representation-Change: an insertion whose payload is the reverse complement of an abutting span on the same reference is now emitted as `ins<range>inv` naming that span instead of as literal reverse-complemented bases — `g.221_222insGCCGACTCCC` becomes `g.221_222ins212_221inv`. Genomic axis only; `c.`/`n.`/`r.`/`m.` still emit literals. The rule fires above a composition-aware coincidence floor read over a fixed 50-base half-window centred on the junction, so the emitted form no longer depends on `NormalizeConfig::window_size`; in the pinned fixture the boundary sits between an eight-base payload (stays literal) and a nine-base one (derives). It costs one confluence class in each shuffle direction — 3': 11,271 to 11,270, 5': 11,272 to 11,271 — whose two spellings previously converged and now settle apart; the census pins are re-blessed and the class `s03-g-m4-sep1-p8-all-ins` is named in `cis_confluence_axis.rs`, with the mechanism recorded as unestablished. Separately, `HgvsToVcfConverter` now returns `UnsupportedVariant` for an inserted payload that is a range, count or uncertainty instead of splicing its HGVS spelling into the ALT allele (which produced invalid records such as `ALT=C212_221inv`); that also changes behaviour for user-authored range payloads predating this derivation.


nh13:
@coderabbitai pause

PR #2014, opened 2026-08-15
docs(rulings): re-point the ledger's prose citations of general.md at the clauses they mean

### LINKED ISSUE #2009: Ruling ledger: 45 prose citations of `general.md:34` now resolve to the exception clause, not the separation rule
45 prose citations of `general.md:34` in the ruling ledger now point at the **exception** clause, which says the opposite of what they mean

## What is wrong

#1793 bumped the spec submodule `6f85311` → `565b973` and repaired every citation in the **structured** `clauses` array. That repair was a uniform −1 shift over `general.md`:

| clause | before | after |
|---|---|---|
| "two variants separated by one or more nucleotides should be described individually and **not** as a delins" | `:34` | **`:33`** |
| "**exception**: two variants separated by one nucleotide, together affecting one amino acid" | `:35` | **`:34`** |
| prioritisation list `(1) substitution … (5) insertion` | `:56` | **`:55`** |
| "descriptions removing part of a reference sequence … are not allowed" | `:58` | **`:57`** |

The structured array is correct — `generate_spec_fixture` verifies every `clauses` entry's quote against the spec checkout and fails the build otherwise, which is exactly why #1793 had to exist.

**Nothing verifies the prose.** The `rationale` and `question` fields cite clauses in running text, and those were not repaired. On `origin/main` (`8881a54e`):

- **45 occurrences of `general.md:34` across 12 records**, essentially all meaning the *separation* rule
- against **11** structured citations that now correctly say `:33`, and **2** that say `:34` meaning the *exception*

So the same ledger cites two different clauses for one rule, and the prose spelling resolves to the wrong one.

## Why this is worse than an off-by-one

The two clauses are near-opposites. `:33` is the rule that **splits** — "described individually and **not** as a delins". `:34` is the **exception that merges** — "two variants separated by one nucleotide, together affecting one amino acid, should be described as a delins".

So a reader following a prose citation lands on a clause that argues the other way. Several of these records are precisely about when to split versus merge, and the record whose id is literally `coding-axis-merges-are-a-disclosed-general-34-deviation` carries 11 of the 45.

The repository's own guidance says to cite `general.md:34`, not "the separation rule", *because* an exact anchor survives review — which is what makes a silently-moved anchor expensive rather than cosmetic.

## Counts, reproducible

```sh
git show origin/main:tests/fixtures/grammar/hgvs_spec_normalization_overrides.json > /tmp/ledger.json
python3 - <<'PY'
import json, re, collections
d = json.load(open("/tmp/ledger.json"))
hits = collections.Counter()
for r in d["rulings"]:
    blob = json.dumps({k: v for k, v in r.items() if k in ("rationale", "question")})
    n = len(re.findall(r"general\.md:34\b", blob))
    if n:
        hits[r["id"]] = n
print(len(hits), "records,", sum(hits.values()), "occurrences")
for k, v in hits.most_common():
    print(f"{v:3d}  {k}")
PY
```

Top contributors: `coding-axis-merges-are-a-disclosed-general-34-deviation` (11), `separation-is-a-property-of-the-spelling-not-of-the-variant` (9), `derivation-may-not-be-bounded-by-the-inputs-spelling` (5), `separation-rule-force-modal-or-negation` (4), `delins-recommendation-reach-when-the-input-arrives-split` (4), `unchanged-is-read-over-every-minimal-alignment` (3).

`:56` and `:58` in prose need the same treatment and are not counted above.

## Why it must be one PR, not fixed piecemeal

This was found while fixing #1980, and deliberately **not** fixed there: repairing the spelling inside one record would make it disagree with 12 siblings and with its own older prose. The ledger's value is that one rule is written in one place, so a half-applied rename is worse than none.

## Suggested shape

1. Sweep `rationale` and `question` for `general.md:NN` and apply the same −1 map #1793 applied to `clauses`, **verifying each against the spec text** rather than applying the shift blindly — some prose may cite a line that did not move, and `:34` legitimately means the exception in the 2 places the structured array also says `:34`.
2. A record id containing `general-34` (`coding-axis-merges-as-a-disclosed-general-34-deviation`) should **not** be renamed — ids are referenced elsewhere and the id is a name, not a citation. Note it in the record instead.
3. **Add a guard**, or this recurs on the next submodule bump. `ruling_citation_currency` checks ids and statuses; nothing checks prose anchors. A scan that extracts `general\.md:\d+` from prose and asserts the cited line still contains the phrase the record leans on would have caught this, and would catch `:56`/`:58` too.
4. Re-bless `docs/NORMALIZATION_CONTRACT.md` (`BLESS_CONTRACT_DOC=1`) and commit it by explicit path.

Item 3 is the one worth insisting on: this class is invisible until someone reads a citation and follows it, and the ledger is explicitly the place where re-deriving a settled question is supposed to be unnecessary.

nh13:
Closes #2009

Representation-Change: none. Ledger prose, one generated document and one new test module; no watched directory is touched, no normalized output moves, and no ruling's status or outcome changes.

## What was wrong

#1793 bumped `assets/hgvs-nomenclature` from `6f85311` to `565b973`. Upstream deleted a duplicated line at old `general.md:22` (a byte-identical copy of old `:17`), so citations at old ≤21 are unaffected and old ≥23 shift by −1.

`generate_spec_fixture`'s `verify_citation` checks every structured `clauses` entry's quote against the checkout and fails the build otherwise, so that register had to be repaired and was. **Nothing verifies prose.** `question`, `rationale`, `guard.none` and the `by_input` notes cite the same clauses in running text with no quote attached, and #1793 repaired the prose of three records and left the other seventeen.

The expensive case is the separation rule, which moved `34` → `33`. Its new occupant at `:34` is the codon **exception**, which argues the opposite way — `:33` splits ("described individually and **not** as a delins"), `:34` merges. So a reader following a prose citation landed on the clause that says the reverse of what the citing sentence claimed.

## What changed

**100 qualified `general.md:N` prose citations**, each read in context and pointed at the line that carries the text its own sentence is about, rather than shifted blindly. Every one now resolves. The table is the whole post-repair population of **117** — the 100 repaired, plus the 17 that were already current:

| now cites | the clause it lands on | count |
|---|---|---|
| `general.md:33` | two variants separated by one or more nucleotides … described individually and **not** as a delins | 53 |
| `general.md:34` | **exception**: two variants separated by one nucleotide, together affecting one amino acid | 7 |
| `general.md:55` | **prioritisation**: … (1) substitution, (2) deletion, (3) inversion, (4) duplication, (5) insertion | 21 |
| `general.md:57` | descriptions removing part of a reference sequence and replacing it with part of the same sequence are not allowed | 7 |
| `general.md:43` | **exception**: deletions/duplications around exon/exon junctions | 6 |
| `general.md:42` | the 3'rule applies to ALL descriptions | 4 |
| `general.md:49` | nucleotides in lower case using IUPAC-IUBMB assigned nucleotide symbols | 4 |
| `general.md:35-38`, `:36`, `:38` | the rejected-SVD-WG010 NOTE block | 5 |
| `general.md:13`, `:40`, `:47`, `:56`, `:95` one each; `:22` and `:156-159` twice each | 9 |
| `general.md:78` | `[ ]` denote alleles — arrived with #2008, already current, left alone | 1 |

**81 bare `` `:N` `` shorthands.** The ledger cites a clause by bare line number once an earlier sentence has established its file, and #1793's partial repair left those dangling — `whole-span-reverse-complement-types-as-inv` ended up reading "`general.md:33` is stated over 'two variants' … grades `:34` as rule 2" about one clause. Only shorthands whose antecedent the record's own prose establishes as a `general.md` clause were touched, applied per record and simultaneously so `:57` → `:56` and `:56` → `:55` inside one record cannot cascade. Every other bare `:N` in those records belongs to another file (`delins.md`, `duplication.md`, `complex.md`, `numbering.md`, `standards.md`, `checklist.md`, `glossary.md`, and one `spec_conformance_axis.rs` line) and is untouched.

**15 qualified citations are deliberately unchanged.** Thirteen sit in the three records #1793 already repaired (`inversion-vs-a-mixed-member-competitor`, `duplication-must-ranks-the-label-not-the-partition`, `whole-span-reverse-complement-types-as-inv`) and are already new-numbered; the fourteenth is `general.md:13`, which is below the deleted line and did not move; the fifteenth is the `general.md:78` that arrived with #2008 and is correct against the pinned checkout.

**No record is renamed.** `coding-axis-merges-are-a-disclosed-general-34-deviation` keeps its id — an id is a name, not a citation, and it is resolved by exact string from other records and from Rust prose. The record now says so in its own text, so `general-34` cannot be read as an anchor into the spec.

**Nothing else about the ledger moves.** Checked mechanically against `origin/main`: same 32 record ids, same statuses, and byte-identical `governing`, `deviates_from`, `clauses`, `applies_to`, `equivalence_classes`, `house_choice` and `guard.tests` on every record.

## The guard — `tests/it/ledger_prose_clause_anchors.rs`

Without one this recurs on the next bump and stays invisible until someone follows a citation.

**Why the obvious guard does not work**, recorded in the module docs so it is not proposed again: "every prose citation must name a line the structured array also names" does not fire here. `general.md:34` *is* a structured citation — two records cite it legitimately, meaning the exception — so no rule keyed on `file:line` can tell a prose `:34` meaning the exception from one meaning the separation rule. A line number carries no meaning; only a phrase does.

So the check reads the meaning off the **citing sentence**. `ANCHORS` names six clause identities, each carrying

- a **signature** phrase used to *locate* the clause in the pinned checkout, so the expected line number is derived from the spec on every run and is never restated in the test — the next bump moves the expectation and the check together; and
- **triggers**, phrases the ledger's own prose uses when a sentence is about that clause.

For each prose citation of `general.md`, the sentence containing it is examined. If exactly one identity triggers and the sentence carries exactly one citation, the citation must name that identity's located line. Zero triggers, two triggers, or two citations in one sentence are skips — a trigger that cannot be attributed to one citation is not a verdict.

Four supporting checks keep it from passing vacuously: every signature must locate exactly one line (a reworded or duplicated clause fails by name instead of silently disarming the rest), no trigger may be contained in another identity's, both hand-rolled scanners are pinned against literals, and the number of judged citations has a floor well below the measured 30.

**What it does not catch** is in the module docs rather than left to be discovered: citations whose sentence names no trigger (84 of 117 on this branch — most of the ledger's citations are bare cross-references), sentences carrying two citations or triggering two identities, bare `` `:N` `` shorthands (no file to resolve, which is why 81 had to be repaired by hand here), files other than `general.md` (`delins.md` alone exists three times under `docs/`, so a bare basename cannot be resolved), and whether a record's argument is actually *right*. A weaker universal companion check asserts every prose citation is in range and lands on non-blank text.

### Sabotage, twice

An assertion never observed failing is indistinguishable from one that cannot.

Reverting one repaired citation — `general.md:33` → `:34` in `delins-payload-coincidence-carve-out-is-coding-dna-scoped` — gives exactly the #2009 defect back, and:

```
1 prose citation(s) in tests/fixtures/grammar/hgvs_spec_normalization_overrides.json name a
different line from the clause their own sentence is about:
  delins-payload-coincidence-carve-out-is-coding-dna-scoped.rationale cites general.md:34 in a
  sentence about the `separation-rule` clause, which is at docs/recommendations/general.md:33:
      On the other DNA axes - `g.`, `m.`, `o.`, `n.` - `general.md:34` governs and the members
      are described individually.
```

Rewording one anchor's signature — simulating upstream rewording the clause — fires the self-locating precondition instead of silently disarming everything built on it:

```
the `three-prime-universality` anchor's signature matches 0 lines of
docs/recommendations/general.md ([]); it must match exactly one, since the line it matches IS
the expectation. Re-point the signature at the clause's current wording.
```

Both restored; all five checks green.

Measured reach: over the ledger as it stood before this change — that is, over current `main` — the rule fires on **27** citations across **20** record/field pairs. Over the repaired ledger it judges **29** and fires on none.

## Residue this does not fix, measured rather than assumed

The same staleness exists outside the ledger. Measured over tracked files at this commit: **930 occurrences of `general.md:N` across 139 files** in `src/`, `tests/`, `examples/`, `docs/`, `README.md` and `CLAUDE.md` (excluding the ledger itself, the regenerated `docs/NORMALIZATION_CONTRACT.md` and the new test), 323 of them spelled `:34`. That set is mixed — 15 already say `:33` and 17 already say `:35-38` — so it needs the same per-occurrence adjudication rather than a shift, and it would collide with a large number of open branches. It is out of #2009's scope, which is the ledger, and it is filed separately rather than left unsaid.

## Also regenerated

`docs/NORMALIZATION_CONTRACT.md` is a committed render of the ledger, so it is re-blessed with `BLESS_CONTRACT_DOC=1` and committed by explicit path; `normalization_contract_doc::the_published_document_is_current` passes without the environment variable. The two gitignored spec artifacts were regenerated, which is what re-validates every structured citation against the checkout.

## Rebased onto #2008

#2008 and this branch both edit `delins-adjacent-members-when-both-consume-reference`, disjointly: #2008 **appends** rationale paragraphs 16-21 and three structured `clauses`; this branch rewrites paragraphs **0-1** in place (`general.md:34` -> `:33`, `:56` -> `:55`). The ledger holds each rationale as one JSON string, so that surfaced as a single-line conflict. Resolved as the **union** — paragraphs 0-1 from this branch, 2-15 unchanged (byte-identical on both sides), 16-21 and all 13 clauses from `main`. Verified paragraph by paragraph: 22 paragraphs, none dropped, duplicated or reordered, and with every `file:line` token masked, the replaced paragraphs 0-1 are byte-identical to `main`'s, so the only difference is the two citations.

#2008's newly arrived prose cites `general.md:78` once. Read against the pinned checkout (`565b973`), `:78` is "`[ ]` (square brackets) are used for alleles ... which includes multiple inserted sequences at one position", which is what the citing sentence claims it supplies. It needed no re-pointing and was left alone. Its only effect here is on the guard's own census: the qualified population goes 116 -> 117 and the zero-trigger skips 83 -> 84, both updated in the module docs. Everything else the docstring measures is unmoved — 27 fires across 20 record/field pairs pre-repair, 29 judged post-repair, 51 stale `:34` citations at the base, 53/10/43 for the separation rule, and every row of the trigger-widening table.

Three-ref ledger comparison against the pre-rebase tip (`935b7b8b`), the rebased `HEAD` and `origin/main`: 32 records on all three, none dropped from either side, no status changed, and with citations masked the only non-citation prose difference from `main` is this branch's own added paragraph in `coding-axis-merges-are-a-disclosed-general-34-deviation`.


nh13:
The out-of-ledger residue measured in the *Residue this does not fix* section is filed as #2015.

nh13:
@coderabbitai pause

PR #2020, opened 2026-08-16
fix(ci): pin release-plz's tag format and guard it against the wheels gate

nh13:
## The defect

v0.14.0 published to crates.io and never reached PyPI. `release-wheels.yml` gated its four building jobs on `startsWith(github.event.release.tag_name, 'v')`; v0.14.0 was tagged `ferro-hgvs-v0.14.0`, so all ten jobs skipped — including `verify-pypi`, whose whole purpose is to catch a publish that did not happen. `publish.yml` reported success throughout, so the only visible signal said the release had gone fine.

Nothing renamed the tag on purpose. release-plz tags a single-package workspace `v{version}` and a multi-package one `{package}-v{version}`, and #1977 added `tests-soak` as a workspace member to buy a CI compile-time saving. A PR about build times renamed the release tag, in a diff where no line mentioned tags.

The non-obvious part, and the reason a *test harness* could do this: release-plz decides via `contains_multiple_pub_packages` (`release_plz_core/src/project.rs`), which is `packages.len() > 1` over a list filtered by whether a package carries **release metadata** — not by `is_publishable()`, despite the name. `tests-soak` is `publish = false` and still counted.

## The fix — three changes, and the gate is untouched

**1. `release-plz.toml` pins `git_tag_name = "v{{ version }}"`.** This is the whole fix, not half of one. `git_tag()` renders a configured `tag_name_template` and reaches the package-count default only through `unwrap_or_else`, so a configured template short-circuits it entirely and the workspace shape is never consulted. Every future tag is `v0.15.0`, `v0.16.0` … the format the existing gate already accepts, and the format the eleven releases before v0.14.0 used.

**2. `tests/it/release_tag_gate.rs` binds the pin to the gate.** It derives the tag the release configuration will mint and asserts that every gated job accepts it, that no gated job accepts the `test-fixtures-v1` data release, that all four gated jobs carry the *same* gate, and that the two publishing jobs still require a real `release` event. The three files are read at runtime, so a mutation is a re-run with no recompile.

Verified in both directions: on the pre-#1977 tree the guard is green, and it goes red the moment #1977's own diff lands.

**3. `CHANGELOG.md` repairs the 0.14.0 compare link** — the third symptom of the same defect, and a live 404. It read `compare/ferro-hgvs-v0.13.1...ferro-hgvs-v0.14.0`; the `ferro-hgvs-` prefix began with 0.14.0, so that range's left endpoint never existed. Measured: the old link returns **404**, `compare/v0.13.1...v0.14.0` returns **200**. Every other entry in the file already uses the `v`-form.

## Why the gate is NOT widened to also accept `ferro-hgvs-v`

An earlier revision of this PR added a second prefix as "belt and braces" to the pin. It was dropped, because it is not additive — it disables the new guard:

| gate | `git_tag_name` | guard |
|---|---|---|
| wide (`v` + `ferro-hgvs-v`) | drifted back to `{{ package }}-v{{ version }}` | 4 passed, EXIT=0 |
| wide | deleted outright | 4 passed, EXIT=0 |
| `'v'` only | drifted | **EXIT=100**, names it |
| `'v'` only | deleted | **EXIT=100**, names it |

With the widening in place the guard cannot detect either way the pin can go wrong. The belt works by silencing the alarm on the braces, and the whole thesis of this PR is that the pin↔gate relation was unwatched. The gate matching the pin *exactly* is what gives the guard something to check.

The `verify-pypi` version strips are likewise reverted to `${TAG_NAME#v}`. A `ferro-hgvs-` strip is unreachable once the gate cannot admit such a tag.

## The guard is a floor, and two holes found in review are closed

The first revision of the guard was green on two realistic mutations that break the invariant it exists to protect. Both are now red, each proven by applying the mutation, watching it fail, and restoring byte-for-byte:

- **A gate dropped from one of the four jobs** reproduced the v0.14.0 skip-chain verbatim under a green guard — the four jobs are siblings, not a chain, so `sdist` skipping takes `check-metadata` and every publishing job with it. Prefixes are now read **per job** and required to be identical across the four; a job that lost its `if:` fails by name.
- **An additive loosening** — keeping the working arms and adding `|| contains(tag_name, 'v')` — is valid GHA, passes `actionlint`, admits `test-fixtures-v1`, and left both extracted prefixes untouched. The guard now asserts that every mention of `tag_name` in a gated job's condition is inside a `startsWith` it can read, so an arm it cannot judge cannot be added silently. (An earlier draft of this description claimed the obvious "just loosen it" repair could not pass. That was **false** for the additive form; it is true now, and asserted rather than argued.)

Also fixed: the pin's "interpolates nothing" assertion checked only that *something* interpolated, so `git_tag_name = "{{ package }}"` — every release reusing one tag — passed. It now checks for `{{ version }}`, which is what its message always demanded.

Eight mutations were run against the final tree; all eight go red, each with a message naming the actual fault.

## Known gap, not closed here

`startsWith(tag, 'v')` admits **any** `v`-prefixed name, and this repo ships non-code releases through the same `release: published` event with no naming rule. A future `validation-corpus-v1` would fire the whole matrix. GitHub expressions have no regex, so requiring a digit means either ten `'v0'…'v9'` arms or hoisting the decision into one small `gate` job evaluating a real regex in `bash`, which the four jobs consume through `needs:`. Ten arms would re-inflate the accepted-prefix set from an equality with the pin's literal prefix back to a membership test — the exact property dropping the widening buys — and re-add the four-way duplication. The `gate`-job form is the structurally right answer and also removes that duplication; both are follow-ups, deliberately not folded into a release-blocking fix. Recorded as a named gap in the module docs.

## v0.14.0 recovery — already done, independently of this PR

**This PR does not retroactively fix `ferro-hgvs-v0.14.0`**; the pin governs future tags only. That release has been recovered out-of-band:

- commit `e4c53fb2` is now **also** tagged `v0.14.0` (annotated, matching the shape of `v0.13.1`);
- a GitHub release published on that tag fired `release: published` with a tag the gate accepts;
- run [`31927735597`](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/31927735597) ran the full chain with all four building jobs **passing** the gate rather than skipping, through `publish-pypi` and `verify-pypi` (which strips `v0.14.0` → `0.14.0`).

Re-tagging is safe: release-plz's "already released" decision is registry-driven (`is_published` runs `cargo info` against crates.io) and never consults tags for a publishable package, so a second tag cannot trigger a re-release. Nothing in the repo triggers on a tag push.

Creating that tag is also what gives the repaired compare link a left endpoint to resolve against, and what will make the *next* release's `compare/v0.14.0...v0.15.0` link resolve.

## Merge ordering

**This PR must land before #2022 (`chore: release v0.14.1`).** #2022 was generated under the multi-package format — its changelog already reads `compare/ferro-hgvs-v0.14.0...ferro-hgvs-v0.14.1` — so merging it first mints `ferro-hgvs-v0.14.1` and v0.14.1 misses PyPI exactly as v0.14.0 did. #2022 has been converted to draft until this lands.

Representation-Change: none. CI and release tooling only; no file under a watched directory is touched.

### DIFF
--- a/.github/workflows/release-wheels.yml
+++ b/.github/workflows/release-wheels.yml
@@ -32,6 +32,25 @@ concurrency:
 # into `env:` — a job-level `if:` cannot read the `env` context. Everything
 # downstream (`check-metadata`, the two smoke matrices, `upload`, `publish-pypi`,
 # `verify-pypi`) reaches these four through `needs:` and so skips with them.
+#
+# The tag format this gate matches is PINNED, in `release-plz.toml`'s
+# `git_tag_name`. It has to be, because release-plz's default is a function of
+# the workspace and not of anything a releaser sets: it tags a single-package
+# workspace `v{version}` and a multi-package one `{package}-v{version}`. #1977
+# added the `tests-soak` driver as a workspace member to buy a CI compile-time
+# saving — release-plz counts a member by whether it has release metadata, not by
+# whether it is publishable, so a `publish = false` test harness still flipped the
+# format — and the very next release was tagged `ferro-hgvs-v0.14.0`. It did not
+# start with `v`, so all ten jobs below skipped, including `verify-pypi`, whose
+# whole purpose is to catch a missing publish; `publish.yml` reported success, and
+# v0.14.0 reached crates.io and never reached PyPI.
+#
+# So the gate is deliberately NARROW: it admits the pinned format and nothing
+# else, and `tests/it/release_tag_gate.rs` fails if the pin and this gate ever
+# disagree again. That narrowness is the point — a gate widened to accept a format
+# the pin cannot mint is a gate that cannot notice the pin drifting back, or being
+# deleted. Widening this to something like `contains(…, 'v')` would additionally
+# re-admit `test-fixtures-v1`, which is what the paragraph above is about.
 
 jobs:
   # abi3 wheels: one wheel per (OS, arch) tagged `cp310-abi3`, runs on
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [0.14.0](https://github.com/fulcrumgenomics/ferro-hgvs/compare/ferro-hgvs-v0.13.1...ferro-hgvs-v0.14.0) - 2026-08-16
+## [0.14.0](https://github.com/fulcrumgenomics/ferro-hgvs/compare/v0.13.1...v0.14.0) - 2026-08-16
 
 ### Representation changes
 
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -5,6 +5,46 @@
 # features even while the major version is 0.
 features_always_increment_minor = true
 
+# Pin the tag format, because release-plz's default is a function of the
+# workspace rather than of anything a releaser sets: a single-package workspace
+# is tagged `v{version}`, a multi-package one `{package}-v{version}`.
+#
+# #1977 made this a multi-package workspace — it added `tests-soak`, a
+# `publish = false` test driver, purely to buy a CI compile-time saving — and the
+# next release was tagged `ferro-hgvs-v0.14.0` instead of `v0.14.0`. Nothing in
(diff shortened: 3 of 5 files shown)


nh13:
@coderabbitai pause

PR #2021, opened 2026-08-16
docs(normalize): correct the changed_columns_of_pieces comment

### LINKED ISSUE #1940: merge.rs comment promises `minimal_changed_columns`, a helper that was abandoned and will never arrive
`src/normalize/merge.rs:17856` refers to a helper, `minimal_changed_columns`, as something that is on its way:

```rust
// Two columns, against the four the spanning `delins` this replaced
// claimed. The count is pinned by value here rather than against the
// block's own Levenshtein distance: `minimal_changed_columns` is the
// helper that would state minimality directly, and it arrives with
// the separate change that derives the weight bound's ceiling from
// the block instead of the input.
assert_eq!(changed_columns_of_pieces(&pieces), 2);
```

**That change was abandoned, so the helper will never arrive.** `minimal_changed_columns` appears exactly once in the whole repository — in the comment above. There is no definition.

## What happened

The "separate change that derives the weight bound's ceiling from the block instead of the input" is commit `52e95dbb` on the branch `fix/issue-1440-sequence-derived-comparand` (2026-08-05). It replaced the input-relative ceiling with a sequence-derived `minimal_changed_columns`, and it was never opened as a PR or merged.

#1440 was instead resolved by **#1899**, which took the other route: delete the input-relative weight bound outright rather than re-deriving its ceiling. Confirming that on `origin/main`:

- `changed_columns_of_edits` — the input-relative measure — has **no definition**; the three surviving occurrences are prose in doc comments describing it as deleted, e.g. *"That bound is deleted (`rulings[derivation-may-not-be-bounded-by-the-inputs-spelling]`)"*.
- `minimal_changed_columns` — its intended replacement — was never added.

So the comment describes a design that no longer exists on either side: the thing it says will be replaced is already gone, and the thing it says will replace it is not coming.

## Why it is worth fixing rather than ignoring

The sentence is not decorative — it explains **why the assertion pins a literal `2` instead of computing minimality**, and it justifies that as a temporary measure pending an arriving helper. A reader who believes it will either go looking for a helper that does not exist, or leave the value pinned on the assumption that a better formulation is imminent. Neither is true, and the deletion of the bound is what actually settled the question.

## Suggested fix

Rewrite the comment to state the current position: the count is pinned by value, the input-relative bound was deleted by #1899 under `rulings[derivation-may-not-be-bounded-by-the-inputs-spelling]`, and no sequence-derived ceiling replaced it. Drop the reference to `minimal_changed_columns`, or keep it only as an explicit note that the alternative was considered and not taken.

Docs-only; no behaviour and no pinned value changes.

## Provenance

Found while auditing PR-less branches during a worktree sweep on 2026-08-14 — the dangling reference is what identified `fix/issue-1440-sequence-derived-comparand` as superseded rather than pending.

nh13:
The comment above the `changed_columns_of_pieces(&pieces) == 2` assertion promised `minimal_changed_columns` — "the helper that would state minimality directly" — and said it would arrive with a separate change deriving the weight bound's ceiling from the block instead of the input. That helper never arrived.

What actually happened: PR #1899 deleted the input-relative weight bound (`changed_columns_of_edits`) outright, under `rulings[derivation-may-not-be-bounded-by-the-inputs-spelling]`, and did not add a sequence-derived measure to replace it. So no helper states the block's minimality directly, and pinning the split's column count by value is the honest statement of what the split costs. The rewritten comment says that, matching the sibling doc comment on `changed_columns_of_pieces` which already attributes the deletion to the same ruling.

Comment-only change inside the `#[cfg(test)]` module; no behavior change.

Closes #1940

Representation-Change: none

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -18288,11 +18288,14 @@ mod tests {
                 [(0, 1, b"".as_slice()), (3, 4, b"".as_slice())]
             );
             // Two columns, against the four the spanning `delins` this replaced
-            // claimed. The count is pinned by value here rather than against the
-            // block's own Levenshtein distance: `minimal_changed_columns` is the
-            // helper that would state minimality directly, and it arrives with
-            // the separate change that derives the weight bound's ceiling from
-            // the block instead of the input.
+            // claimed. The count is pinned by value rather than derived because
+            // no helper states the block's minimality directly. The
+            // input-relative measure that once bounded a derivation by the
+            // input's own weight (`changed_columns_of_edits`) was deleted
+            // outright by #1899, under
+            // `rulings[derivation-may-not-be-bounded-by-the-inputs-spelling]`,
+            // and nothing sequence-derived was added to replace it — so a
+            // literal is the honest statement of what this split costs.
             assert_eq!(changed_columns_of_pieces(&pieces), 2);
         }
 


nh13:
@coderabbitai pause

PR #2023, opened 2026-08-16
fix: stop tx_to_genomic from collapsing n.*N onto its in-transcript sibling

### LINKED ISSUE #1950: CoordinateMapper::tx_to_genomic drops the downstream flag, collapsing n.5 and n.*5
`CoordinateMapper::tx_to_genomic` (`src/convert/mapper.rs:431`) discards `TxPos::downstream`, so two distinct transcript positions collapse onto one genomic coordinate.

## Measured

```
tx_to_genomic(n.5)   -> Ok(Some(1004))
tx_to_genomic(n.*5)  -> Ok(Some(1004))
```

`n.5` and `n.*5` are different positions. The 3'UTR flag is dropped and both resolve to the same answer.

## Why it was missed

It is the direct sibling of the `tx_to_cds` route fixed in #1762, and it was **not** found by that PR's census of marker-consulting sites — because that census enumerated call sites that *already consult* the flag. A site that never mentions it cannot appear in such a list.

That is the same shape as the lesson recorded elsewhere in this project: enumerating consumers of a value proves nothing about paths that never read it. The census was sound and answered a narrower question than it appeared to.

## Why it is filed rather than fixed in #1762

Wider blast radius: production callers in `vcf/from_hgvs.rs::convert_tx`, `normalize/` and `equivalence/`. And the signature is `Result<Option<u64>, _>`, so "decline" versus "error" is a design decision this route has to make on its own terms rather than inherit.

## Reachability

Qualified, and worth stating so the severity is not overread: `n.*N` is refused at parse in **every** mode by #1748 — measured, both `parse_hgvs` and `parse_hgvs_lenient` return `E1003`. So the collapsing spelling cannot currently arrive through a parsed description. `TxPos` is public API and constructible directly, so the route is reachable by a caller building the struct, but this is defence-in-depth rather than a live service defect.

nh13:
Closes #1950.

## The defect

`CoordinateMapper::tx_to_genomic` (`src/convert/mapper.rs`) read only `TxPos.base` and never `TxPos.downstream`, so `n.5` and `n.*5` — two distinct positions — collapsed onto one genomic coordinate:

```
tx_to_genomic(n.5)   -> Ok(Some(1004))
tx_to_genomic(n.*5)  -> Ok(Some(1004))   # wrong: the 3'UTR-past-end flag is dropped
```

`n.*N` (`TxPos::downstream`) names a base past the transcript's last one, which is covered by no exon.

## The fix

Decline `n.*N` via the method's existing `Ok(None)` "not covered by an exon" channel, placed right beside the pre-existing `base < 1` decline:

```
tx_to_genomic(n.5)   -> Ok(Some(1004))
tx_to_genomic(n.*5)  -> Ok(None)
```

This makes the two transcript termini symmetric: the method already declines with `Ok(None)` for a position 5' of the transcript (`base < 1`), and `n.*N` is the exact 3' mirror. `background/numbering.md:52` numbers the `n.` axis only `n.1..n.<length>` and `:54` forbids describing a base beyond a transcript against that transcript, so there is no in-frame genomic coordinate to answer with.

**Why `Ok(None)`, not `Err` like the sibling `tx_to_cds` (#1762).** The two methods differ precisely because their signatures do. `tx_to_cds` returns `CdsPos` with no `Option`, so its only decline channel is `Err`. `tx_to_genomic` returns `Option<u64>` and already uses `Ok(None)` for the 5'-mirror case; erroring on the 3' side while declining on the 5' side would be an inconsistency *within the same method*, which is worse than the cross-method difference (each method uses its natural decline mechanism).

## Unchanged on purpose

- **The intronic sibling `tx_to_genomic_with_intron` keeps its control flow; only its diagnostic changed.** It was verified not to produce a wrong coordinate for downstream inputs: its non-intronic branch delegates to the now-fixed `tx_to_genomic` (and errors via `ok_or_else`), and its intronic branch (`n.*N+M`) already errors because `intronic_to_genomic` finds no such boundary. The wrong-coordinate class had exactly one site (`tx_to_genomic`); adding a guard elsewhere would be untestable defensive code.
- **In-transcript and 5' positions are untouched** — only `downstream` inputs change.

- **One error string is corrected** (user-visible, but not a representation change). `tx_to_genomic_with_intron`'s non-intronic branch reported `Position {base} not found in exons`. For `n.*5` that read `Position 5 not found in exons` — but position 5 *is* in an exon; the `*` is what makes it unmappable, so the message named the wrong cause. It now distinguishes the downstream case explicitly, and the general case no longer asserts one of `tx_to_genomic`'s four `Ok(None)` routes as though it were the only one. Control flow is untouched, and no test asserts on that string (`found in exons` appears only at the two `mapper.rs` sites and two unrelated literals in `service/handlers/vcf_convert.rs`). The identical string at `cds_to_genomic_with_intron` is deliberately left alone: its `TxPos` comes from `cds_to_tx`, which sets `downstream: false` as a documented decision, so the mis-diagnosis is unreachable there.

## Reachability

Qualified, and stated so severity is not overread. `n.*N` is refused at parse in every mode by #1748, so the collapsing spelling cannot arrive through a parsed description. `TxPos` is public API and constructible directly, so a Rust caller building the struct can reach this route — defence-in-depth rather than a live service defect.

## Representation stability

Representation-Change: none. `src/convert/` is not a watched path, and `tx_to_genomic` is exercised only by `downstream`-flagged positions here, which are parse-blocked in every mode; no normalized output moves.

## Verification

- New regression tests, both strand orientations: `tx_to_genomic_does_not_collapse_downstream_onto_its_in_transcript_sibling` (plus, `n.5`→`Some(1004)`, `n.*5`→`None`) and `tx_to_genomic_declines_downstream_on_minus_strand` (minus, `n.5`→`Some(3005)`, `n.*5`→`None`). Both pin exact coordinates and assert the two do not share one.
- `cargo fmt --all --check`, `cargo clippy --all-features --all-targets -- -D warnings`, and `cargo nextest run --features dev --no-fail-fast` all pass locally.

### DIFF
--- a/src/convert/mapper.rs
+++ b/src/convert/mapper.rs
@@ -528,12 +528,15 @@ impl<'a> CoordinateMapper<'a> {
     /// `c.`/`n.` number.
     ///
     /// `None` is not exclusively that answer, though, so do not read it as
-    /// "unaligned" without ruling the other two out: it is also returned for a
-    /// non-positive `tx_pos.base`, and for a `Strand::Unknown` transcript,
-    /// whose bases are all reported unaligned however complete its exon table
-    /// is. A caller using `tx_to_genomic(..).is_some()` as its definition of
-    /// "aligned" — `tests/it/axis_frame_disagreement.rs`'s genome-frame walk
-    /// does — is relying on the transcript having a known strand.
+    /// "unaligned" without ruling the others out: it is also returned for a
+    /// non-positive `tx_pos.base` (5' of the transcript), for an `n.*N`
+    /// downstream position (`TxPos::downstream`, 3' of the transcript — the two
+    /// termini decline symmetrically, #1950), and for a `Strand::Unknown`
+    /// transcript, whose bases are all reported unaligned however complete its
+    /// exon table is. A caller using `tx_to_genomic(..).is_some()` as its
+    /// definition of "aligned" — `tests/it/axis_frame_disagreement.rs`'s
+    /// genome-frame walk does — is relying on the transcript having a known
+    /// strand.
     pub fn tx_to_genomic(&self, tx_pos: &TxPos) -> Result<Option<u64>, FerroError> {
         // Check if transcript has genomic coordinates
         if !self.transcript.has_genomic_coords() {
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #2024, opened 2026-08-16
fix: decline a cds_end past the sequence end instead of panicking

### LINKED ISSUE #1909: validate_cds_pos panics on a cds_end past the sequence end (44 of 425,565 shipped transcripts)
`validate_cds_pos` (`src/convert/coding.rs:18`) computes a bound from `cds_end` with unchecked subtraction — the underflowing expression is `transcript.sequence_length() - cds_end` at `src/convert/coding.rs:34`. When `cds_end` exceeds the transcript's sequence length the subtraction underflows and **panics** (`attempt to subtract with overflow`) rather than returning the existing malformed-transcript error.

### It is reachable on shipped data

Scanning the shipped cdot files for transcripts whose `stop_codon` exceeds their exon-derived transcript length:

```
stop_codon > tx_length : 44 of 425,565     (all refseq GRCh37; 0 in ensembl GRCh38)
reversed bounds        :  0 of 425,565
```

The loader takes `stop_codon` verbatim with no clamp, so those 44 reach `cds_end` unmodified.

Worth stating the limit of that figure precisely: 44 is the population of the malformed *shape*, measured by comparing cdot's `stop_codon` against exon-derived length and confirming the loader does not clamp. An end-to-end `c.*N` normalize was not traced through one of these accessions, so treat 44 as an upper bound on reachable panics rather than a confirmed count of them.

That still puts it in a different class from the arithmetic guards checked alongside it, which measured 0 of 12,207,712 exon rows and 0 of 1,294 LRG records respectively. This one has a non-zero denominator on real data.

### Not introduced by any open PR

Surfaced during a dual-review pass over #1676, whose every `src/convert/coding.rs` hunk is inside `mod tests` or a doc comment. That PR in fact *documents* the hazard, which is why its fixture warning was left intact rather than "corrected".

### Suggested direction

Use checked subtraction and return the established malformed-transcript error on underflow, preserving current behaviour for well-formed bounds. `make_cds_end_past_sequence_transcript` already exists as a fixture helper and is the natural regression anchor.

A second question worth deciding at the same time: whether the **loader** should reject or clamp a `stop_codon` past the transcript end, rather than every downstream consumer having to defend against it.

nh13:
Closes #1909.

Representation-Change: none. `src/convert/` is not a watched prefix (normalize/hgvs/spdi/project/reference/error_handling); this only converts a debug panic into an error on malformed transcripts and leaves well-formed output untouched.

## The defect

`validate_cds_pos` (`src/convert/coding.rs`) validated a 3'UTR position by computing the 3'UTR length as `transcript.sequence_length() - cds_end`. Both operands are `u64`, so a transcript whose `cds_end` runs past its sequence length underflowed the subtraction and **panicked** in debug (`attempt to subtract with overflow`) instead of returning the existing malformed-transcript error. In release the same subtraction wraps to a nonsense length, silently corrupting the range check.

This is reachable on shipped data. Per #1909's scan of the shipped cdot files, **44 of 425,565** transcripts (all RefSeq GRCh37; 0 in Ensembl GRCh38) carry a `stop_codon` past their exon-derived length, and the loader takes `stop_codon` verbatim with no clamp, so those reach `cds_end` unmodified.

Reproduced on `origin/main` (`e4c53fb2`) with a unit test — panic at `src/convert/coding.rs:33:27: attempt to subtract with overflow`.

## The fix

Use `checked_sub` and, on underflow, return `FerroError::ConversionError` — the same malformed-transcript error the function already returns for a missing CDS/CDS-end, so the house error form is unchanged. Well-formed bounds take the identical path as before.

## Unchanged on purpose

Scoped to the 3'UTR subtraction the issue named. The reversed-bounds shape (`cds_end - cds_start`, line 29) was measured at **0 of 425,565** shipped transcripts and is left as-is, consistent with the issue's scoping and this repo's per-site hardening (cf. #1690, #1087). The loader-side question (whether `prepare` should reject/clamp a `stop_codon` past the transcript end) is deliberately not addressed here.

## Verification

- New regression test `cds_end_past_sequence_end_is_a_malformed_transcript_not_a_panic` (co-located unit test, matching the two existing `validate_cds_pos` tests): asserts it returns `ConversionError`, not a panic and not a coordinate-range error that would misattribute the fault to the caller's `*N`.
- `cargo nextest run --features dev --no-fail-fast` — 11154 passed, 155 skipped, 0 failed.

### DIFF
--- a/src/convert/coding.rs
+++ b/src/convert/coding.rs
@@ -29,8 +29,23 @@ pub fn validate_cds_pos(pos: &CdsPos, transcript: &Transcript) -> Result<(), Fer
     let cds_length = (cds_end - cds_start + 1) as i64;
 
     if pos.utr3 {
-        // 3' UTR position
-        let utr3_length = transcript.sequence_length() - cds_end;
+        // 3' UTR position. `sequence_length()` and `cds_end` are both `u64`, so
+        // a `cds_end` past the sequence end underflows this subtraction and
+        // panics in debug (`attempt to subtract with overflow`). That is a
+        // malformed transcript, not a caller error: cdot's `stop_codon` is
+        // taken verbatim by the loader with no clamp, and 44 of 425,565 shipped
+        // transcripts (all RefSeq GRCh37) have one past their exon-derived
+        // length (#1909). Report it as such rather than aborting the process.
+        let utr3_length = transcript
+            .sequence_length()
+            .checked_sub(cds_end)
+            .ok_or_else(|| FerroError::ConversionError {
+                msg: format!(
+                    "Malformed transcript: CDS end {} is past the sequence end ({})",
+                    cds_end,
+                    transcript.sequence_length()
+                ),
+            })?;
         if pos.base < 1 || pos.base > utr3_length as i64 {
             return Err(FerroError::InvalidCoordinates {
                 msg: format!(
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #2025, opened 2026-08-16
docs(normalize): correct comments still calling PartitionRule::Live the shipped rule

### LINKED ISSUE #1912: merge.rs: four doc comments still call PartitionRule::Live the shipped rule after #1835
Four doc comments in `src/normalize/merge.rs` still describe `PartitionRule::Live` as the shipped rule. #1835 (`feat(normalize)!: make canonical-coalesced the default partition rule`) made that false and did not update them. Verified on `origin/main` @ `19e2e244`.

These are load-bearing comments — they are what a reader consults to find out which arm decides real output — and two of them state the opposite of what the code does. One says the function that decides *every* shipped result decides *none*.

## Ground truth

```rust
// merge.rs:2379
const DEFAULT_PARTITION_RULE: PartitionRule = PartitionRule::CanonicalCoalesced;

// merge.rs:2513, in partition_rule_from_env
None | Some("") => Ok(DEFAULT_PARTITION_RULE),

// merge.rs:2806-2813
PartitionRule::Live => return partition_block(reference, result),
PartitionRule::Shadow => partition_block_sequence_first(reference, result, min_separation),
PartitionRule::Canonical | PartitionRule::CanonicalCoalesced => {
    partition_block_canonical(reference, result)
}
```

So unset `FERRO_PARTITION` yields `CanonicalCoalesced`, which routes to `partition_block_canonical`. `partition_block` is reached only under `FERRO_PARTITION=live`, or as the fallback when the canonical arm declines an oversized grid (`SEQFIRST_DECLINED`, `MAX_SEQFIRST_GRID_CELLS`).

## The four

**1. `:2475` — `partition_rule_from_env`. Wrong on both halves.**
> Unset and empty yield [`PartitionRule::Live`], the shipped rule.

Unset and empty yield `DEFAULT_PARTITION_RULE` (`:2513`), and that is `CanonicalCoalesced`, not `Live`.

**2. `:2644` — same error for a refused value.**
> An embedder that neither calls this nor runs a debug build gets [`PartitionRule::Live`] for a refused value. That is the shipped rule under no name at all…

The refused-value path returns `DEFAULT_PARTITION_RULE` too. The paragraph's *argument* — that a silent fallback lands on the shipped rule rather than on a candidate — still holds; only the arm named is wrong. Worth fixing carefully rather than by find-and-replace, since the reasoning survives.

**3. `:7042` — `partition_block_canonical`. This is the one that inverts.**
> Reached only under `FERRO_PARTITION=canonical` and from the `dump_partitions` example; it does not decide any shipped result.

It is called at `:2812` from `Canonical | CanonicalCoalesced`, and `CanonicalCoalesced` is the default. It decides **every** shipped result except where it declines.

**4. `:7856` — `dev::live`.**
> The shipped rule: `partition_block`'s single-gap alignment search.

`dev::live` wraps `partition_block`, which post-flip is the opt-in arm.

## Why it is worth fixing rather than leaving

`:7042` in particular actively misdirects: someone auditing which code path produces shipped output would read it and eliminate `partition_block_canonical` from consideration, which is exactly backwards. I reached a wrong conclusion about a partition question this way before checking the call sites, so this is a measured cost rather than a hypothetical one.

Docs only — no behaviour change, no census or pin moves.

`Representation-Change: none`

nh13:
Closes #1912.

## The defect

#1835 (`feat(normalize)!: make canonical-coalesced the default partition rule`) flipped the default partition rule from `PartitionRule::Live` to `PartitionRule::CanonicalCoalesced`, but left several doc comments in `src/normalize/merge.rs` still describing `Live` as the shipped/default/unset rule. Two stated the opposite of what the code does:

- `partition_block_canonical`'s doc said it "does not decide any shipped result" — it is dispatched from `Canonical | CanonicalCoalesced` (`merge.rs:3057`), and `CanonicalCoalesced` is the default, so it decides *every* shipped result except the blocks it declines.
- `partition_rule_from_env`'s doc said an unset/empty `FERRO_PARTITION` yields `Live` — the code returns `DEFAULT_PARTITION_RULE` (`= CanonicalCoalesced`).

## The fix

Docs only. Seven comments corrected to describe the real shipped rule, referencing `DEFAULT_PARTITION_RULE` where the code does and dating the flip to #1835 (`Live` shipped up to and including v0.14.0):

1. `partition_rule_from_env` — unset/empty yield `DEFAULT_PARTITION_RULE`.
2. `partition_rule_env_value` — the unset case selects the default, not `live`.
3. `partition_switch_startup_error` — a refused value (non-debug build) falls to `DEFAULT_PARTITION_RULE`, per `partition_rule_from_outcome`.
4. `partition_block_canonical` — reached under `FERRO_PARTITION=canonical` or `canonical-coalesced` (the default), decides every shipped result except declined blocks (grid/span cap → `partition_block`).
5. compensating-gap coalesce doc — this pass is scoped to both canonical arms (`cuts_with_canonical`), so post-flip it reaches the shipped path; the "shipped output does not move" note is now framed as the pre-flip measurement it was.
6. `dev_partitioners::live` — relabelled the `live` arm as opt-in (`FERRO_PARTITION=live`), not the shipped rule.
7. `PARTITION_BLOCKS_CUT` counter doc — "the shipped path", not "the shipped `Live` path".

The issue named the first four; the other three (`partition_rule_env_value`, the coalesce doc, and the counter doc) are the identical defect class, all predate #1835, and were not updated by the flip.

## Unchanged on purpose

The `FERRO_PARTITION=live` refusal message (`merge.rs:~2767`) still names `live` and is left as-is — it is a runtime string asserted by `partition_switch_wiring`'s test (`merge.rs:~17086`), and its refusal reasoning survives the flip. Changing it would be a behaviour change, outside "docs only".

## Verification

- `cargo fmt --check` — clean.
- `cargo nextest run --features dev` — `11158 passed, 155 skipped` (EXIT=0).

Representation-Change: none. Docs/comment-only change; no watched-file behaviour is touched.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2717,8 +2717,10 @@ const _: () = assert!(DERIVED_BLOCK_PARTITION_RULE.cuts_with_canonical());
 
 /// Read a [`PartitionRule`] from what `FERRO_PARTITION` was set to.
 ///
-/// Unset and empty yield [`PartitionRule::Live`], the shipped rule. Anything
-/// else that is not an arm name is an **error**, not a fallback.
+/// Unset and empty yield [`DEFAULT_PARTITION_RULE`], the shipped rule —
+/// [`PartitionRule::CanonicalCoalesced`] since #1835's flip, and
+/// [`PartitionRule::Live`] up to and including v0.14.0. Anything else that is
+/// not an arm name is an **error**, not a fallback.
 ///
 /// # Why this refuses rather than warning
 ///
@@ -2780,8 +2782,9 @@ fn partition_rule_from_env(value: Option<&str>) -> Result<PartitionRule, String>
 /// # Why this is not `.ok()`
 ///
 /// [`std::env::var`] returns `Err` for two unrelated reasons and `.ok()`
-/// collapses both to `None` — but `None` is the **unset** case, which is
-/// legitimately `live`. So a value that *was* set, to bytes that are not UTF-8,
+/// collapses both to `None` — but `None` is the **unset** case, which
+/// legitimately selects the default ([`DEFAULT_PARTITION_RULE`]). So a value
+/// that *was* set, to bytes that are not UTF-8,
 /// selected the shipped rule and said nothing: exactly the silent fallback the
 /// rejection in [`partition_rule_from_env`] exists to remove, reachable through
 /// a door that function never sees. A bake-off whose `FERRO_PARTITION` was
@@ -2886,7 +2889,7 @@ fn partition_rule_outcome() -> &'static Result<PartitionRule, String> {
 /// # The residual, stated rather than hidden
 ///
 /// An embedder that neither calls this nor runs a debug build gets
-/// [`PartitionRule::Live`] for a refused value. That is the shipped rule under
+/// [`DEFAULT_PARTITION_RULE`] for a refused value. That is the shipped rule under
 /// no name at all rather than under a candidate's, and it is strictly safer than
 /// the abort it replaces — but it is only *reported* if somebody asks. Every
 /// binary **and every example** in this repository asks, and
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #2026, opened 2026-08-16
docs(normalize): correct partition_rule() # Panics call-site count (three, not two)

### LINKED ISSUE #1875: merge.rs: partition_rule()'s # Panics doc says "Both call sites" but there are three
Two doc comments in `src/normalize/merge.rs` make a counting claim about call sites that is no longer true. Small, but one of them is a `# Panics` section — a doc about when the function aborts the process — and `CLAUDE.md` states outright that an outdated comment is worse than none.

## The instances

**1. `src/normalize/merge.rs:2492`** — `partition_rule()`'s `# Panics` doc:

> `have. Both call sites are inside [`canonicalize_from_sequence`], which is`

There are **three** call sites, not two.

**2. `src/normalize/merge.rs:2281`** — likely stale in the same way, and nobody has checked it:

> `Both call sites used to spell this inline as`

This one is a claim about the *past*, so it may still be accurate; it is included because it is the same counting claim in the same file and should be verified in the same pass rather than found separately later.

## Provenance

Found while implementing #1834 (PR #1848) and **deliberately left alone** — that PR was required to be a provable no-op on the current default, and folding an unrelated docs correction into it would have muddied the proof. Filing so it is not lost.

## Suggested fix

Re-count both, correct or generalise the wording, and prefer a form that does not go stale on the next caller — "every call site" or naming the invariant rather than the count.

nh13:
Closes #1875.

Representation-Change: none

## The defect

`partition_rule()`'s `# Panics` doc comment in `src/normalize/merge.rs` claimed:

> Both call sites are inside [`canonicalize_from_sequence`], which is infallible …

Both parts are wrong. `partition_rule()` is a private fn, so its call sites are the crate's, and there are **three**, not two:

- `src/normalize/merge.rs:3200` — the body of `canonicalize_from_sequence` (the sole *production* caller, infallible, deep under every public normalization entry point).
- `src/normalize/merge.rs:16074` — `#[cfg(test)] fn the_two_surfaces_cut_with_a_pinned_pair_of_rules()`.
- `src/normalize/merge.rs:17114` — `#[cfg(test)] fn the_suite_runs_on_the_default_rule()`.

So only one of the three is inside `canonicalize_from_sequence`; the other two are test pins that assert on the returned rule directly.

## The fix

Reword the `# Panics` section to name the invariant that actually carries the panic-vs-`Result` reasoning — every call site is either the infallible `canonicalize_from_sequence` (the sole production caller) or a `#[cfg(test)]` pin — rather than a count that goes stale on the next caller, per the issue's suggested fix.

In the same pass, the sibling `cuts_with_canonical()` doc said "Both call sites used to spell this inline as `matches!(rule, Canonical | CanonicalCoalesced)`". That is a past-tense claim (the method has since grown to four callers), and the count adds nothing to the extraction rationale it supports. Generalised to a count-free form ("This used to be spelled inline at its call sites …").

## Unchanged on purpose

Docs/comment-only. No behavior change — `partition_rule()`, `cuts_with_canonical()`, and every caller are byte-identical.

## Representation stability

None. The diff touches only doc comments; no normalized output can move.

## Verification

- `cargo nextest run --features dev` — `11158 tests run: 11158 passed (11 slow), 155 skipped`, exit 0.
- Call-site enumeration confirmed via `grep` over `src/`, `tests/`, `examples/` (private fn, so callers are in-crate only; three invocations, matching the corrected text).

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -2629,7 +2629,7 @@ impl PartitionRule {
     ///
     /// # Why this is a method with an exhaustive `match`
     ///
-    /// Both call sites used to spell this inline as
+    /// This used to be spelled inline at its call sites as
     /// `matches!(rule, Canonical | CanonicalCoalesced)`. A **positive** arm list
     /// is silently wrong for the next arm added: an arm missing from it takes
     /// the `else` branch, which is `live` behaviour, on an arm that does not cut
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #2027, opened 2026-08-16
fix(scripts): read the whole trailer value in the contradicted-decline tripwire

### LINKED ISSUE #1854: A decline can hide a move on a continuation line: the contradicted-decline tripwire reads only the trailer's first line
`TRAILER_RE` captures a trailer's value with `[ \t]*(?P<value>\S.*?)[ \t]*$` under `MULTILINE`, so the value stops at the first newline. This repo's trailers are routinely multi-line and their continuation lines are unindented, so "the first line" is often half a sentence.

**Where this is NOT a defect, because it was already fixed:** the changelog injector. `inject_representation_disclosure.disclosure_value` uses `trailer_value` only to decide *whether* there is a trailer and where it starts, then extends the value to the next column-0 trailer token, Markdown heading, or HTML comment. Verified:

```
  trailer_value      -> '371 rows move of 5,629,002 swept; all 371 were'
  disclosure_value   -> '371 rows move of 5,629,002 swept; all 371 were\npreviously accepted, so a real migration.'
```

So the changelog bullet gets the whole disclosure. Anyone re-deriving this should stop here rather than "fixing" the injector.

## Where it IS live: the contradicted-decline tripwire

`check_representation_change.check` runs `declines()` and `contradicted_decline()` over the **truncated** value. A trailer that declines on its first line and discloses a move on a continuation line therefore passes green:

```python
body = ("Representation-Change: none.\n"
        "  150 rows of 85,642 move, all 3'-direction.\n")

find_declarations(body)                      # ['none.']
check(["src/normalize/merge.rs"], body)      # (True, "Declared `Representation-Change: none.` …")
contradicted_decline(disclosure_value(body)) # '150 rows of 85,642 move'   <- would have fired
```

That is exactly the failure `MOVEMENT_CLAIM_RE` exists to catch — a decline that describes a move, filed as `none`, disclosure never reaching the changelog, nobody told — evaded by a newline. The tripwire's own docstring calls this "the one direction this mechanism must never fail in".

`check_changelog_grouping.opens_with_a_decline` reads the same truncated value and has the same blind spot, so the independent second opinion does not cover it either. That is expected rather than surprising: the two are meant to derive their answers differently, and here they share the truncation because they share `trailer_value`'s shape.

## Blast radius today

**Zero.** Over all 39 open PRs, no trailer declines on its first line while a continuation claims a move. This is a latent hole, not an active mis-disclosure. Filed so it is on the record rather than fixed quietly, and because the fix has a cost worth deciding deliberately.

## Why it was not fixed in #1852

Three reasons, all about not smuggling a behaviour change in beside an unrelated one:

1. It is a **different mechanism** from the near-miss check #1852 adds — value truncation versus declaration invisibility — and mixing them makes both harder to review.
2. Fixing it **widens the tripwire**, and this tripwire's history is a list of times widening it fired on honest declines: #1535's quantified zero, #1547's exemplary decline, and #1647, where the documented quantified-zero form became a merge blocker. Any widening needs its own measured before/after over the real corpus, which is a PR's worth of work.
3. A correct fix touches **both** checkers, and `check_changelog_grouping.py` deliberately shares no code with `check_representation_change.py` (#1555 — the test written to catch that bug passed because the two halves were wrong together). So the fix has to be made twice, independently, and then pinned by a test that runs both over the same values.

## Suggested shape

Route `declines` / `contradicted_decline` over the continuation-extended value rather than the first line — the extension rule already exists and is already trusted by the injector, so this is a matter of reusing it rather than inventing a fourth copy of the column-0 rule. Then extend `test_the_two_decline_rules_agree_value_by_value` with multi-line values, which is the guard that would have caught the divergence.

Before merging any such change, re-run the open-PR sweep: the number to beat is **zero** PRs newly reddened, and the declines listed in `test_a_numerate_decline_is_not_a_contradiction` must all still pass with their continuations attached.

nh13:
Closes #1854.

## The defect

The `Representation-Change:` PR-body checker's contradicted-decline tripwire read only the trailer's **first line**. `TRAILER_RE` captures the value with `\S.*?$` under `MULTILINE`, so the value stops at the first newline, and this repo's trailers are routinely multi-line with **unindented** continuation lines. So a decline on line 1 with a contradicting movement claim on a continuation line slipped through green:

```
  Representation-Change: none.
  150 rows of 85,642 move, all 3'-direction.
```

`check(["src/normalize/merge.rs"], body)` returned `(True, …)` — filed as `none`, the disclosure never reaching the changelog and nobody told, which is the one direction the tripwire's own docstring says it must never fail in. Reproduced on `origin/main` at `17cc8185`: `find_declarations` yields `['none.']`, `contradicted_decline('none.')` is `None`, so the check passes; fed the whole value, `contradicted_decline` returns `'150 rows of 85,642 move'`.

## The fix

Route `declines` / `contradicted_decline` in `check()` over the whole trailer value instead of the first line. New `full_declaration_value` extends the first line to the next line the value cannot own — a column-0 trailer token, a Markdown heading, or the HTML comment GitHub appends (CodeRabbit's summary) — the same bound the injector's `disclosure_value` already uses. It is carried as a **separate copy** so the checker keeps sharing no code with `check_changelog_grouping.py` (#1555), and it reuses this module's own `TRAILER_RE`, so no fourth copy of the column-0 rule is minted. Fenced regions are blanked first, matching `find_declarations`.

## Scope: only the checker, deliberately not the audit

The issue notes `check_changelog_grouping.opens_with_a_decline` shares the truncation. I did **not** change the audit, on measurement. The audit's job is to check that the rendered changelog is consistent with what git-cliff did, and its `trailer_value`-based rule already mirrors git-cliff's line-anchored exclusion. Extending it to read the full value makes it disagree with git-cliff and fire on **honest** declines: measured over `v0.12.0..HEAD`, an extended `opens_with_a_decline` newly flags two merged declines — `08f3024d` (`none` + review-fix prose) and `abd4cbe3` (`none` + a "Measured, not assumed" note) — neither a move. That is exactly the "widening fires on honest declines" failure this tripwire's history is a list of. The enforcement point is the PR-body checker, which now rejects the contradicted decline before it can be merged.

## Unchanged on purpose

- A legitimate multi-line decline whose continuation is a reason still passes (`none.\n  Comments and one test module only.`).
- A real multi-line disclosure (`577 rows move…\n  Previously-accepted…`) still passes.
- The decline vocabulary parity with `release-plz.toml` and `CONTRIBUTING.md` is unchanged (`test_decline_vocabulary_matches_the_changelog_config`, `test_the_two_decline_rules_agree_value_by_value` extended with multi-line values, `test_contributing_documents_the_same_decline_vocabulary` all pass).

## Verification

- `ruff check scripts/ tests/python/` — pass (EXIT 0)
- `ruff format --check scripts/ tests/python/` — pass (EXIT 0)
- `mypy --strict` on `scripts/check_representation_change.py` (mypy 1.20.2, pyproject error codes) — pass (EXIT 0). Note CI's `check-typing` only covers `python/ferro_hgvs`; `scripts/` is stdlib-only and out of that scope.
- `pytest tests/python/test_representation_change_trailer.py` — 169 pass (EXIT 0), including 7 new #1854 tests.
- End-to-end: the checker CLI exits 1 on the reproducer body and 0 on a clean multi-line decline.

Representation-Change: none

### DIFF
--- a/scripts/check_representation_change.py
+++ b/scripts/check_representation_change.py
@@ -518,6 +518,49 @@ def find_declaration(text: str) -> str | None:
     return declarations[0] if declarations else None
 
 
+#: A line that ends a trailer's value because it opens something the value cannot own:
+#: another column-0 trailer token, a Markdown heading, or an HTML comment (GitHub appends
+#: CodeRabbit's summary after the trailer, opening `<!-- ... -->`).
+#:
+#: A deliberately SEPARATE copy of the terminator in
+#: `inject_representation_disclosure.disclosure_value`. This checker shares no code with
+#: `check_changelog_grouping.py` or the injector that imports from it, so the two decline
+#: rules can be wrong independently (#1555). The column-0 rule this extends is this module's
+#: own `TRAILER_RE`, so no fourth copy of *that* is minted.
+CONTINUATION_END_RE = re.compile(r"^([A-Za-z][A-Za-z0-9-]*:[ \t]|#{1,6}\s|<!--)")
+
+
+def full_declaration_value(text: str) -> str | None:
+    """Return the first trailer's value with its continuation lines, or `None`.
+
+    `find_declaration` returns only the first line, because `TRAILER_RE` stops at the
+    newline. This repo's trailers are routinely multi-line with UNINDENTED continuation lines
+    -- `git interpret-trailers --parse` reports no trailers at all for #1537/#1535/#1547 --
+    so a decline on the first line can hide a contradicting `<n> rows move` on a later one
+    (#1854). The contradicted-decline tripwire must read the whole value, not the first
+    clause, or the disclosure is filed as `none` and never reaches the changelog.
+
+    The value runs from the trailer to the next line it cannot own -- a column-0 trailer
+    token, a Markdown heading, or the HTML comment GitHub appends -- the same bound
+    `inject_representation_disclosure.disclosure_value` uses, arrived at independently rather
+    than imported. Fenced regions are blanked first, matching `find_declarations`, so a
+    continuation quoted inside a fence is not scavenged into the value.
+    """
+    first = find_declaration(text)
+    if first is None:
+        return None
+    lines = strip_fenced(text).splitlines()
+    start = next(index for index, line in enumerate(lines) if TRAILER_RE.match(line))
+    collected = [first]
+    for line in lines[start + 1 :]:
+        if CONTINUATION_END_RE.match(line):
+            break
+        collected.append(line.rstrip())
+    while collected and not collected[-1].strip():
+        collected.pop()
+    return "\n".join(collected)
+
+
 def check(changed: list[str], declaration_text: str) -> tuple[bool, str]:
     """
     Decide whether `changed` is adequately declared by `declaration_text`.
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #2029, opened 2026-08-16
refactor(hgvs): collapse the two range-insert spellings behind one accessor

### LINKED ISSUE #1986: hgvs: a same-reference range payload has two AST spellings, and consumers disagree about which they handle
## Summary

One insertion payload — a same-reference position range — has **two AST spellings**, both minted by the parser, and consumers disagree about which they handle. There is no single place that decides, so each consumer has made its own call and the three answers on `origin/main` are "both", "both", and "neither".

All references against `origin/main` at `de8532ee`.

## The two spellings

`InsertedSequence` carries the payload directly:

```rust
// src/hgvs/parser/edit.rs:992,995 — in parse_simple_count
return Ok((remaining, InsertedSequence::PositionRangeInv { start, end }));
return Ok((remaining, InsertedSequence::PositionRange { start, end }));
```

and also, one level down, wrapped in a single-element `Complex`:

```rust
// src/hgvs/parser/edit.rs:371-380 — in parse_inserted_sequence
let seq = match part {
    InsertedPart::PositionRange { start, end } => {
        InsertedSequence::Complex(vec![InsertedPart::PositionRange { start, end }])
    }
    InsertedPart::PositionRangeInv { start, end } => {
        InsertedSequence::Complex(vec![InsertedPart::PositionRangeInv { start, end }])
    }
    ...
```

Both are reached from the **same arm** of `parse_inserted_sequence`'s first-byte dispatch (`b'0'..=b'9' | b'*'`, `:365`). That arm tries `parse_cds_position_range` first and wraps its result in `Complex`; when that parser declines, control falls through to `parse_simple_count`, which returns the bare variant. So which spelling a description gets is decided by which of two parsers accepted it, not by anything about the payload.

## The three coverage decisions

**`src/project/rna.rs:179-192` — handles both.** `resolve_range_insert` matches the bare variants and then re-matches the single-part `Complex` shape, and its doc comment states outright that it exists to absorb the difference:

> Handles the bare `PositionRange`/`PositionRangeInv` variants and the single-part `Complex([PositionRange | PositionRangeInv])` shape the parser produces for `delins191_194inv` — both collapse to a single `Literal`.

**`src/normalize/rules.rs` — handles both.** `append_part_bases` (`:3290`) covers the `InsertedPart` side; the bare side is covered separately, and the comment at `:3233` names the two spellings side by side as a pair that must stay in step:

> exactly as the same-reference `InsertedPart::PositionRangeInv` / `InsertedSequence::PositionRangeInv` arms do

**`src/spdi/convert.rs` — handles neither.** The file contains no reference to `PositionRange` at all. A payload that reaches SPDI conversion unresolved therefore cannot be converted, which is why `hgvs_to_spdi` currently declines `ins<range>inv` output.

Five files read the bare variant (`hgvs/alignment_symbols.rs`, `hgvs/edit.rs`, `hgvs/parser/edit.rs`, `normalize/rules.rs`, `project/rna.rs`); the `Complex`-of-one shape has its own, only partly overlapping, set.

## Why this is worth fixing rather than documenting

The two "handles both" sites are not sharing code — they are two independent re-derivations of the same collapse, each with its own doc comment explaining the same asymmetry. That is the tax: every new consumer must discover the duplication for itself, and the failure when it does not is silent, because one spelling keeps working.

`spdi/convert.rs` is the case where nobody discovered it. Whether the right answer there is "convert.rs grows the arms" or "the payload is always resolved to a literal before conversion" is exactly the question this issue is asking — and it cannot be answered locally in `convert.rs`, because the same question has already been answered twice elsewhere without being written down as a rule.

## Suggested shape of a fix

Normalize at the boundary: have the parser mint **one** spelling, or give `InsertedSequence` a single accessor that answers "is this a same-reference position range, and which one" for both shapes. Then the two existing collapses become calls, and a consumer that forgets is a compile error rather than a silent decline.

## Not asserted here

Which concrete input strings land on which spelling. The discriminator is "did `parse_cds_position_range` accept it", and that parser handles offsets and `*` markers as well as plain digits, so the boundary needs measuring rather than reading. The duplication above holds regardless of where that line falls.

nh13:
Closes #1986.

Representation-Change: none. Parser-boundary/consumer accessor only; no parse output, `Display`, or normalized output changes. The single-part `Complex` spelling and the bare variant already `Display` identically, and the AST is untouched — this only unifies how two existing consumers read it.

## The defect

An all-numeric same-reference range payload has **two AST spellings** the parser mints interchangeably, from the same first-byte dispatch arm (`b'0'..=b'9' | b'*'`, `src/hgvs/parser/edit.rs`):

- the bare `InsertedSequence::PositionRange` / `PositionRangeInv` variant (from `parse_simple_count`), and
- the single-part `Complex([PositionRange | PositionRangeInv])` shape (from `parse_cds_position_range`, wrapped in `Complex`).

Both denote one contiguous slice of the same accession, and `Display` renders them identically (`Complex` unbrackets a single part). Which spelling a description gets is decided by which of two parsers accepted it, not by anything about the payload.

Two consumers had each **independently re-derived the same collapse**, each with its own doc comment explaining the same asymmetry:

- `src/spdi/convert.rs::position_range_insert`
- the head of `src/project/rna.rs::resolve_range_insert`

That is the tax the issue describes: every new consumer must rediscover the duplication, and forgetting one spelling fails silently because the other keeps working. (The third site the issue named, `src/spdi/convert.rs` handling *neither* spelling, was since patched by #1966 — it now handles both, i.e. it grew a third copy of the collapse. The root cause — no single place that decides — remained.)

## The fix

Add one accessor on the type that owns the two spellings:

```rust
pub fn as_position_range(&self) -> Option<(u64, u64, bool)>
```

on `InsertedSequence`, decomposing both spellings to `(start, end, inverted)`. Route both consumers through it and delete the two hand-rolled matches. The collapse now lives in exactly one place with one doc comment; a new consumer is pointed at the accessor.

I chose the single accessor over minting one spelling at the parser boundary because the accessor is the zero-risk option: it changes no AST, so it cannot move any parse/display/normalized output, and it is what the issue's "and/or" explicitly permits.

## Unchanged on purpose

- **Behavior is preserved exactly.** `as_position_range` returns `None` for the same inputs the old matches skipped (multi-part `Complex`, `CdsPositionRange`, external refs, `SpecialPositionRange`, and every non-range variant), so both consumers keep their existing fall-through handling.
- **`src/normalize/rules.rs` is left alone.** It handles the payload by fetching bases per-part (`append_part_bases` for the `Complex` side, separate bare arms), not by decomposing to `(start, end, inverted)`, so it is not the duplicated collapse this change targets.

## Verification

- `cargo nextest run --features dev --no-fail-fast` — **11163 passed, 155 skipped**.
- New unit test `hgvs::edit::tests::as_position_range_collapses_both_spellings` pins that both spellings display identically yet are distinct AST values, that the accessor collapses both (forward and inverted), and that the documented non-range inputs (`Literal`, multi-part `Complex`, `CdsPositionRange`, `SpecialPositionRange`) return `None`.

### DIFF
--- a/src/hgvs/edit.rs
+++ b/src/hgvs/edit.rs
@@ -402,6 +402,39 @@ impl InsertedSequence {
         }
     }
 
+    /// Decompose a same-reference position-range payload to
+    /// `(start, end, inverted)` 1-based inclusive coordinates, or `None` when
+    /// this is not one.
+    ///
+    /// An all-numeric same-reference range has **two** AST spellings that the
+    /// parser mints interchangeably and that [`Display`](fmt::Display) renders
+    /// identically: the bare [`PositionRange`](InsertedSequence::PositionRange) /
+    /// [`PositionRangeInv`](InsertedSequence::PositionRangeInv) variant, and the
+    /// single-part `Complex([PositionRange | PositionRangeInv])` shape (the
+    /// parser wraps `parse_cds_position_range`'s result in `Complex` but falls
+    /// through to the bare variant for `parse_simple_count`). Both denote one
+    /// contiguous slice of the same accession, so this is the single place that
+    /// collapses them — consumers call it rather than re-deriving the match, and
+    /// so cannot disagree about which spelling they handle (#1986).
+    ///
+    /// Multi-part `Complex` payloads, [`CdsPositionRange`](InsertedPart::CdsPositionRange)
+    /// (intronic-offset / UTR-marker ranges), external references,
+    /// [`SpecialPositionRange`](InsertedSequence::SpecialPositionRange) and every
+    /// non-range variant return `None`, so the caller keeps its existing
+    /// handling for them.
+    pub fn as_position_range(&self) -> Option<(u64, u64, bool)> {
+        match self {
+            InsertedSequence::PositionRange { start, end } => Some((*start, *end, false)),
+            InsertedSequence::PositionRangeInv { start, end } => Some((*start, *end, true)),
+            InsertedSequence::Complex(parts) => match parts.as_slice() {
+                [InsertedPart::PositionRange { start, end }] => Some((*start, *end, false)),
+                [InsertedPart::PositionRangeInv { start, end }] => Some((*start, *end, true)),
+                _ => None,
+            },
+            _ => None,
+        }
+    }
+
     /// Get the bases if this is a literal sequence
     /// Returns None for count, range, repeat, complex, or empty variants
     pub fn bases(&self) -> Option<&[Base]> {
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #2030, opened 2026-08-16
fix(project): project a genomic cis allele onto the requested transcript

### LINKED ISSUE #1562: project: a genomic cis allele cannot be projected to any transcript axis — --transcript is compared against the input's chromosome accession
## Summary

`ferro project` refuses to project a **bracketed cis allele** on the `g.` axis onto any transcript axis. The same transcript, at the same locus, projects a **single** `g.` variant without complaint — so the rejection is specific to the allele path, not to the transcript argument being wrong.

The error reports the input's *chromosome* accession as though it were a transcript:

```
$ ferro project --axis c --transcript NM_004006.2 --reference <dir> \
    'NC_000023.11:g.[33211289dup;33211293del]'
ERROR: NC_000023.11:g.[33211289dup;33211293del]
       - --transcript NM_004006.2 does not match the input's transcript NC_000023.11
```

Identical failure on `--axis c`, `--axis p`, `--axis r` and `--axis n`.

## The control that rules out a wrong transcript

Same reference directory, same transcript, a coordinate inside the same 4 nt span — a lone `g.` variant projects fine:

```
$ ferro project --axis c --transcript NM_004006.2 --reference <dir> \
    'NC_000023.11:g.33211289dup'
NC_000023.11(NM_004006.2):c.24dup
```

`NM_004006.2` (DMD) is annotated on `NC_000023.11`, and `c.21_24` maps to `g.33211289_33211292`. The transcript selection is correct; only the bracketed-allele path rejects it.

## Why this matters beyond the CLI ergonomics

A genomic cis allele is the shape a variant caller emits. Projecting it onto a transcript — and from there to a protein consequence — is the primary reason to hold one.

It also blocks a normalization question that is currently open. Ferro keeps a multi-member cis allele such as `g.[261del;263_264insA]` split, and one defence of that choice is that the split is harmless because a downstream projection re-derives it from the resulting sequence. That defence is **unverifiable today**, because the projection does not run at all. Whether the split survives the handoff to `c.`/`p.` is exactly what cannot be measured.

For context on what is at stake in that handoff: the two members of such an allele are each individually frameshifting while their combined effect is not, so a per-member protein prediction reports two frameshifts where the truth is a single missense. `DNA/delins.md:19` and `DNA/delins.md:47` both give preventing that as their stated rationale.

## Expected

A `g.` cis allele projects onto the named transcript, with all members carried through, exactly as a single `g.` variant does.

## Acceptance

- A `g.` cis allele with `--transcript` projects to `c.`, `n.`, `r.` and `p.`
- The regression test asserts the projected member list, not merely that the call succeeds
- The mismatch check still fires when `--transcript` genuinely disagrees with a *transcript*-accession input

nh13:
Closes #1562.

Representation-Change: none. Newly enables a projection that was previously refused outright; it converts a hard error into a successful projection and moves no existing normalized output. The change is in `src/cli/project.rs` (CLI dispatch), which is not a watched prefix.

## The defect

`ferro project` refused to project a genomic (`g.`) cis allele onto any transcript axis, while the same transcript projected a lone `g.` variant at the same locus without complaint:

```
$ ferro project --axis c --transcript NM_004006.2 --reference <dir> \
    'NC_000023.11:g.[33211289dup;33211293del]'
ERROR: --transcript NM_004006.2 does not match the input's transcript NC_000023.11
```

The error reports the input's *chromosome* accession as though it were a transcript. The cause is in `project_axis` (`src/cli/project.rs`): it routed only `HgvsVariant::Genome(_)` through the genomic dispatch. A genomic cis allele parses to `HgvsVariant::Allele`, so it fell through to the transcript-coordinate branch, which takes the transcript from `variant.accession().transcript_accession()` — a *chromosome* accession for a genomic allele — and hard-errors when `--transcript` does not match it.

The engine already handles a genomic allele: `project_variant_inner` has an `HgvsVariant::Allele` arm (`src/project/projector.rs`). Only the CLI dispatch guard was wrong.

## The fix

Key the dispatch on the coordinate axis rather than the variant kind: route every genomic-axis input (`coordinate_axis() == Some(CoordinateAxis::Genomic)`) through `project_axis_genomic`, which requires `--transcript` and projects onto it. This is the same idiom the Python bindings already use (`is_genomic`, `src/python.rs`), and it catches the allele — and any other genomic-axis shape (`GenomeRing`, a genomic `Supernumerary`) — exactly as a lone `g.` variant is caught.

## Unchanged on purpose

- A **transcript-coordinate** cis allele (`NM_…:c.[…]`) still names its transcript in the accession, so a `--transcript` that genuinely disagrees with it remains a hard error — pinned by a discriminating test, so the fix cannot silently weaken the mismatch guard.
- `NullAllele`/`UnknownAllele` and mixed-axis alleles resolve to `coordinate_axis() == None` and keep their prior (transcript) path.
- The broadened routing only pulls genomic-axis shapes out of the transcript-coordinate branch — none of which ever carried a transcript accession, so that branch was always wrong for them. No transcript-coordinate input changes path.

## Verification

- `project_axis_genomic_cis_allele_projects_all_members` — a genomic cis allele of two well-separated substitutions projects onto `--transcript` and carries **both members through** onto the coding axis (`NC_000001.11(NM_TEST.1):c.[2T>A;9A>C]`). Asserts the projected member list, per the issue's acceptance criteria; the exact string is measured, not inferred.
- `project_axis_coding_cis_allele_transcript_mismatch_is_hard_error` — the discriminating guard: a `c.` cis allele with a disagreeing `--transcript` is still a hard error.
- `cargo nextest run --features dev --no-fail-fast` — 11164 passed, 0 failed, 155 skipped.

### DIFF
--- a/src/cli/project.rs
+++ b/src/cli/project.rs
@@ -3,7 +3,7 @@
 //! "unavailable" (exit 0) vs "hard failure" (exit 1).
 
 use crate::error::FerroError;
-use crate::hgvs::variant::HgvsVariant;
+use crate::hgvs::variant::{CoordinateAxis, HgvsVariant};
 use crate::normalize::NormalizationWarning;
 use crate::project::{VariantProjection, VariantProjector};
 use crate::reference::ReferenceProvider;
@@ -206,7 +206,15 @@ pub fn project_axis<P: ReferenceProvider + Clone>(
     axis: Axis,
     transcript: Option<&str>,
 ) -> Result<AxisOutcome, ProjectCliError> {
-    if matches!(variant, HgvsVariant::Genome(_)) {
+    // Route every *genomic-axis* input through the genomic dispatch, not just a
+    // lone `HgvsVariant::Genome`. A genomic cis allele
+    // (`NC_…:g.[261del;263_264insA]`) parses to `HgvsVariant::Allele`, whose
+    // accession is a *chromosome*, not a transcript — so the transcript-
+    // coordinate branch below would compare `--transcript` against that
+    // chromosome accession and reject a valid projection (#1562). Keying on the
+    // coordinate axis rather than the variant kind catches the allele (and any
+    // other genomic-axis shape) the same way a lone `g.` variant is caught.
+    if variant.coordinate_axis() == Some(CoordinateAxis::Genomic) {
         return project_axis_genomic(projector, variant, axis, transcript);
     }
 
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

Issue #2003, opened 2026-08-15
ci: three more heavy censuses are stuck on the debug archive because the cis corpus is regenerated, not published


nh13:
#2001 moves `clinvar_hgvs_tests` and the `direction_symmetry` unit census off the debug shards. Three more heavy modules were in scope and could **not** be moved, for one shared structural reason. Recording it so the next attempt starts from the blocker rather than rediscovering it.

## The remaining work

Per-test costs on `Test oracle (2/4)` and its siblings, run [`31909112992`](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/31909112992):

| module | test | cost |
|---|---|---:|
| `cis_confluence_axis` | `five_prime_confluence_census` | 65.1s |
| `cis_confluence_axis` | `three_prime_confluence_census` | 61.4s |
| `cis_confluence_nr_axis` | `rna_five_prime_confluence_census` | 45.5s |
| `cis_confluence_nr_axis` | `noncoding_five_prime_confluence_census` | 41.1s |
| `cis_confluence_nr_axis` | `rna_three_prime_confluence_census` | 39.4s |
| `from_sequences_corpus` | `every_derived_description_is_valid_and_stable` | 44.7s |

Roughly 300s of debug-archive work, spread across the four oracle shards.

## Why they cannot move today

Both `cis_confluence_*` modules build their corpus through `common::fixture_gen`, which **regenerates a missing fixture by shelling out to a nested `cargo run`**. The `censuses` job has no crate to build from — it runs from a prebuilt archive and never compiles anything. `ci.yml`'s comment on that job already states the same constraint for a different module:

> `spec-fixtures` IS needed — `normalize_axis_preserving::axis_is_preserved_over_the_spec_fixture` calls `ensure_spec_fixture`, which regenerates via a nested `cargo run` when the file is absent, and this job has no crate to build from.

`from_sequences_corpus` is blocked transitively: it references `crate::cis_confluence_axis`, so it cannot compile into the soak driver without that module coming too.

## The shape of a fix

[...]

PR #2032, opened 2026-08-16
test: anchor cwd-relative fixture reads and guard against reintroduction

### LINKED ISSUE #2019: tests: 23 fixture paths are cwd-relative, so any module moved into the soak driver breaks in CI
The working directory a test runs in is a property of its **package**, not of the repository. `cargo nextest` sets the cwd to the package root, so a test in `ferro-hgvs` runs from the workspace root while the same module compiled into `ferro-hgvs-soak-tests` runs from `tests-soak/`. Any fixture resolved against a bare relative path therefore works in `it` and breaks in the soak driver — and it breaks by reporting the file missing, which reads as an environment problem rather than a path bug.

## Three instances so far, none found by review

| module | landed in | found by |
|---|---|---|
| `common/bulk_fixtures.rs` | #1977 | during development |
| `common/fixture_gen.rs` | #1977 | during development |
| `clinvar_hgvs_tests.rs` (failure-expectations snapshot) | #2001 | **CI, after the push** |

The third is the informative one. `clinvar_hgvs_tests` had lived in `tests/it/` for a long time and was correct there; it became wrong the moment #2001 compiled it into the driver, with no edit to the module itself. The failure:

```
thread 'clinvar_hgvs_tests::test_clinvar_hgvs_500k_benchmark' panicked at
  tests-soak/tests/soak/../../../tests/it/common/failure_expectations.rs:257:9:
Failure-expectations snapshot not found at tests/fixtures/bulk/clinvar_hgvs_500k_failure_expectations.json.
```

The repo already documents the class — `CLAUDE.md` calls it *the one thing the soak driver exercises that `it` cannot* — so this is a known hazard that has been rediscovered three times rather than swept once.

## Current exposure

Measured on `main` (`grep` over `tests/it/**/*.rs` for a direct `read_to_string` / `File::open` on a `"tests/fixtures/…"` literal):

- **23 call sites across 15 modules** resolve a fixture against the cwd.
- **None of them is currently in the soak driver**, so nothing is broken today.
- #2003 proposes moving three more modules in. Each move re-rolls this dice, and the cost is a red CI run after a push rather than a local failure.

The affected modules:

```
biocommons_tests
clinical_genes_tests
clinvar_tests
clinvar_validation
comprehensive_edge_case_tests
conformance_summary_generated
dbsnp_tests
edge_case_notation_tests
external_pattern_tests
idempotency_tests
mavedb_tests
mutalyzer_grammar_tests
mutalyzer_tests
parser_tests
variantvalidator_tests
```

(A further single site uses `.push("tests/fixtures/…")` onto an already-rooted `PathBuf`, which is correct and is not counted above.)

## The fix

`common::fixture_gen::fixture_path` is the resolver to use. It ascends from `CARGO_MANIFEST_DIR` to the first ancestor holding `Cargo.lock` — which cargo writes only at the workspace root — so it is correct from either package:

```rust
// before
fs::read_to_string("tests/fixtures/external/biocommons_real.tsv")
// after
fs::read_to_string(fixture_gen::fixture_path("tests/fixtures/external/biocommons_real.tsv"))
```

Two ways to do it, and the second is the one that ends the class:

1. **Sweep the 23 sites.** Mechanical, and it removes the trap from every module before anyone moves it.
2. **Add a guard.** A test that scans `tests/it/**/*.rs` for a direct filesystem call on a `"tests/fixtures/…"` literal and fails, with an allowlist for anything deliberately cwd-relative. That is the same shape as the existing `soak_package_membership` and `generator_completeness` guards, and it is what stops instance four.

Doing (1) without (2) leaves the next new test free to reintroduce it.

## Note on what a green local run proves

Two checks that do **not** catch this, both of which were run on #2001 before the push:

- `cargo build --tests -p ferro-hgvs-soak-tests` — compiles the module; the path is a runtime value.
- The membership guards in `tests/it/soak_package_membership.rs` — they read `main.rs` and `ci.yml` as *text*, deliberately, so all 20 passed while the driver did not even compile.

The check that does catch it is the one `CLAUDE.md` prescribes and that is easy to skip: `cargo nextest run -p ferro-hgvs-soak-tests`. Note it needs the bulk fixtures actually present (`scripts/fetch-test-fixtures.sh`), or the affected tests return early and pass without reaching the read.

nh13:
Closes #2019.

Representation-Change: none. Tests only; no watched source file is touched.

## The defect

A test's working directory is a property of its **package**, not the repository. `cargo nextest` runs a module from its package root, so a module compiled into the root `ferro-hgvs` package's `it` target runs from the workspace root, while the *same* module `#[path]`-included by the `ferro-hgvs-soak-tests` member (`tests-soak/tests/soak/main.rs`) runs from `tests-soak/`. A fixture read against a bare relative path therefore works in `it` and breaks the moment the module is pulled into the soak driver — and it breaks by reporting the file missing, which reads as an environment problem rather than a path bug.

The class had been rediscovered three times, the last only after a push reddened CI:

| module | landed in | found by |
|---|---|---|
| `common/bulk_fixtures.rs` | #1977 | during development |
| `common/fixture_gen.rs` | #1977 | during development |
| `clinvar_hgvs_tests.rs` | #2001 | **CI, after the push** |

Measured on `origin/main` (`e16f99ca`): **23 direct `fs::read_to_string` / `File::open` reads on a `"tests/fixtures/…"` literal, across 15 modules**, resolve against the cwd. None is in the soak driver today, so nothing is broken yet — but every module moved in (e.g. #2003's proposal) re-rolls the dice, at the cost of a red CI run rather than a local failure.

## The fix

Route all 23 sites through `common::fixture_gen::fixture_path`, which anchors on `CARGO_MANIFEST_DIR` and ascends to the workspace `Cargo.lock` (which cargo writes only at the workspace root), so it resolves correctly from either package:

```rust
// before
fs::read_to_string("tests/fixtures/external/biocommons_real.tsv")
// after
fs::read_to_string(fixture_gen::fixture_path("tests/fixtures/external/biocommons_real.tsv"))
```

The affected modules: `biocommons_tests`, `clinical_genes_tests`, `clinvar_tests`, `clinvar_validation`, `comprehensive_edge_case_tests`, `conformance_summary_generated`, `dbsnp_tests`, `edge_case_notation_tests`, `external_pattern_tests`, `idempotency_tests`, `mavedb_tests`, `mutalyzer_grammar_tests`, `mutalyzer_tests`, `parser_tests`, `variantvalidator_tests`.

**No fixture content or test semantics change** — only how each path is resolved.

## The guard

New module `cwd_relative_fixture_paths` scans every `tests/it` source and fails on any filesystem read whose direct argument is a `"tests/fixtures/…"` string literal — the exact shape all three prior instances took, and the one a new test is most likely to reintroduce by copy-paste. It stops instance four.

It is a token scan (a floor, not a proof) and its limits are stated in the module docs: a path bound to a variable and read later, or composed at run time, is invisible to it — the fs call sees an identifier, not the literal. It does see through rustfmt's line-wrapping, and `the_detector_matches_direct_reads_and_only_those` pins both directions (the shapes it must catch, and the already-anchored shapes it must not) against a synthetic sample, so the matcher cannot silently stop matching after the tree is swept clean.

**Out of scope, and left for a follow-up:** `conformance_summary_generated` also passes two `Path::new("tests/fixtures/…failure-patterns.md")` values to `summary::check`, and a handful of other modules (`api_comparison_tests`, `spdi_tests`, `annotation_real_format_slices`, …) read fixtures through a `Path::new`/variable indirection. Those are the broader indirect class the issue's census deliberately did not count and this guard does not claim to catch; anchoring them is a separate, larger sweep.

## Verification

```
cargo fmt --all --check                                  # clean
cargo clippy --all-features --all-targets -- -D warnings # clean
cargo nextest run --features dev --no-fail-fast          # 11166 passed, 0 failed, 155 skipped
```

The guard was developed TDD: it flagged exactly the 23 sites (RED) before the sweep, and passes after (GREEN).

### DIFF
--- a/tests/it/biocommons_tests.rs
+++ b/tests/it/biocommons_tests.rs
@@ -32,8 +32,10 @@ fn default_supported() -> bool {
 }
 
 fn load_biocommons_fixtures() -> TestFixture {
-    let content = fs::read_to_string("tests/fixtures/grammar/biocommons.json")
-        .expect("Failed to read biocommons.json");
+    let content = fs::read_to_string(crate::common::fixture_gen::fixture_path(
+        "tests/fixtures/grammar/biocommons.json",
+    ))
+    .expect("Failed to read biocommons.json");
     serde_json::from_str(&content).expect("Failed to parse biocommons.json")
 }
 
--- a/tests/it/clinical_genes_tests.rs
+++ b/tests/it/clinical_genes_tests.rs
@@ -42,8 +42,10 @@ struct ClinicalGeneCase {
 }
 
 fn load_clinical_genes_fixtures() -> ClinicalGenesFixture {
-    let content = fs::read_to_string("tests/fixtures/validation/clinical_genes.json")
-        .expect("Failed to read clinical_genes.json");
+    let content = fs::read_to_string(crate::common::fixture_gen::fixture_path(
+        "tests/fixtures/validation/clinical_genes.json",
+    ))
+    .expect("Failed to read clinical_genes.json");
     serde_json::from_str(&content).expect("Failed to parse clinical_genes.json")
 }
 
--- a/tests/it/clinvar_tests.rs
+++ b/tests/it/clinvar_tests.rs
@@ -66,8 +66,10 @@ struct Statistics {
 }
 
 fn load_clinvar_fixtures() -> ClinVarFixture {
-    let content = fs::read_to_string("tests/fixtures/validation/clinvar.json")
-        .expect("Failed to read clinvar.json");
+    let content = fs::read_to_string(crate::common::fixture_gen::fixture_path(
+        "tests/fixtures/validation/clinvar.json",
+    ))
+    .expect("Failed to read clinvar.json");
     serde_json::from_str(&content).expect("Failed to parse clinvar.json")
 }
 
--- a/tests/it/clinvar_validation.rs
+++ b/tests/it/clinvar_validation.rs
@@ -15,7 +15,7 @@ use std::io::{BufRead, BufReader};
 
 #[test]
 fn test_clinvar_validation() {
-    let file = File::open("tests/fixtures/bulk/hgvs4variation.txt.gz")
+    let file = File::open(crate::common::fixture_gen::fixture_path("tests/fixtures/bulk/hgvs4variation.txt.gz"))
         .expect("ClinVar data not found. Run: curl -o tests/fixtures/bulk/hgvs4variation.txt.gz https://ftp.ncbi.nlm.nih.gov/pub/clinvar/tab_delimited/hgvs4variation.txt.gz");
     let decoder = GzDecoder::new(file);
     let reader = BufReader::new(decoder);
--- a/tests/it/comprehensive_edge_case_tests.rs
+++ b/tests/it/comprehensive_edge_case_tests.rs
@@ -63,8 +63,10 @@ struct TestCase {
 }
 
 fn load_fixtures() -> TestFixture {
-    let content = fs::read_to_string("tests/fixtures/comprehensive_edge_cases.json")
-        .expect("Failed to read comprehensive_edge_cases.json");
+    let content = fs::read_to_string(crate::common::fixture_gen::fixture_path(
+        "tests/fixtures/comprehensive_edge_cases.json",
+    ))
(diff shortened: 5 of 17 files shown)


nh13:
@coderabbitai pause

PR #2033, opened 2026-08-16
test: pregenerate spec fixtures via a nextest setup script (#1608)

### LINKED ISSUE #1608: test: on-demand spec-fixture generation races the suite, changing the test count and reddening `clause_ruling_index`
A local `cargo nextest run --features dev` can report a **different total test count** and three spurious `clause_ruling_index` failures, on an unchanged tree, when the generated spec fixture is written during the run.

## Observed

Three consecutive full-suite runs on the same commit, same worktree, nothing edited between them:

| run | total | result |
|---|---|---|
| 1 | 9,869 | 9,868 passed, 1 failed (a real, since-fixed stale citation) |
| 2 | **9,795** | 9,792 passed, **3 failed** |
| 3 | 9,869 | 9,869 passed, 0 failed |

Run 2's failures:

```
FAIL ferro-hgvs::it clause_ruling_index::every_cited_clause_appears_in_the_index
FAIL ferro-hgvs::it clause_ruling_index::the_index_is_not_vacuous
FAIL ferro-hgvs::it clause_ruling_index::the_rendered_index_is_current
```

All seven `clause_ruling_index` tests pass when run alone, immediately afterwards, with no change to the tree.

`tests/fixtures/grammar/hgvs_spec_normalization.json` and `hgvs_spec_enumeration.json` both carry mtimes from **during** run 2 — so the artifacts were being generated while the run was reading them.

## Why it matters more than an ordinary flake

The **total test count changed**, from 9,869 to 9,795. That is not what a flaky assertion looks like; it reads as "someone deleted 74 tests", and combined with three red tests naming the ruling ledger it reads as a real regression in the thing under review. It cost a full investigation cycle to rule out, and the only thing that distinguished it from a genuine failure was re-running.

`the_index_is_not_vacuous` exists precisely to catch an empty index. Here it fires on a *transiently* empty one, which inverts its meaning: the guard is right, the input is momentarily wrong.

## Suspected cause

`tests/it/common/spec_fixture.rs` regenerates the fixture on demand when it is missing, which is a deliberate convenience so a fresh checkout "just works". Under nextest, tests run in parallel across binaries, so a run that starts without the artifact can have one process writing it while others read it — a partially-written or absent file, depending on timing.

I have not root-caused which reader observes the partial write, and the exact trigger for the artifact being absent at the start of run 2 is also unconfirmed. Reporting the observation with its evidence rather than a diagnosis.

## Possible directions

- Generate the artifacts once, before the test processes start, rather than on demand from inside them — CI and the pre-push hook already do exactly this, which is likely why CI has never shown it.
- Or make the on-demand path atomic: generate to a temporary file in the same directory and `rename` into place, so a reader sees either the old file or the new one and never a partial one.
- Either way, a reader that finds the artifact absent or unparseable should say so, rather than proceeding on an empty index.

nh13:
Closes #1608.

Representation-Change: none. Touches only `.config/nextest.toml`, `scripts/`, `tests/it/` and `CLAUDE.md` — no watched `src/` prefix, and no output can move.

## The defect

Under a full `cargo nextest run --features dev`, the total test count can change and three `clause_ruling_index` tests can go red on an unchanged tree, passing on an isolated re-run.

The **reported** cause — a partial write of the fixture being read concurrently — was already fixed in #1085: `tests/it/common/fixture_gen.rs` generates to a per-process temp file and atomically renames it into place, under an in-process `GEN_LOCK`, so a reader now always sees either the old file or the new one, never a partial one. That does not reproduce.

The **real residual** is one level up. The two gitignored spec fixtures (`tests/fixtures/grammar/hgvs_spec_normalization.json` and `hgvs_spec_enumeration.json`) are produced on demand from inside the test body, via `ensure_generated`, which fires a nested `cargo run --features dev --bin generate_spec_*`. Under nextest, many test processes start at once; when the artifact is absent, several of them race to fire that nested `cargo run` concurrently. Those siblings contend the cargo build lock — against each other and against other worktrees compiling — and the generator intermittently `exits with failure`, which panics a spec-fixture consumer. `clause_ruling_index` reddens as collateral of the same build-lock contention (and `the_index_is_not_vacuous` inverts its meaning, firing on a transiently-empty index). CI never sees this because CI (and the pre-push hook) pregenerate the fixtures explicitly before the test run.

## The fix

Wire a nextest `[scripts.setup]` in `.config/nextest.toml` that runs `scripts/pregenerate-spec-fixtures.sh` once, serially, after nextest's build phase and before any test process starts. That removes the concurrency at its source. It is the local-`nextest` analogue of the CI / pre-push pregeneration (guarded by #1775).

- **The on-demand fallback is intact.** The setup script only runs under nextest. A plain `cargo test` (no nextest), and any nextest run that still finds a fixture missing, keep regenerating through `fixture_gen.rs` exactly as before — this removes the race, not the fallback.
- **The binding is SCOPED to the consumers.** `filter = 'all()'` is what this PR shipped first, and it broke two CI jobs — see below. The filter now names the four `tests/it` modules that actually read a fixture, and `tests/it/spec_fixture_setup_filter.rs` keeps that list honest so the scoping is not the fragile hand-maintained roster the `all()` argument was avoiding.
- **Generation is atomic.** The `generate_spec_*` binaries write their default output in place (`std::fs::write`), so the script writes to a sibling temp via `--output` and `mv`s it into place — mirroring `fixture_gen.rs`'s own temp-then-rename. A killed run cannot leave a partial file that the existence guard would then treat as valid.
- **Semantics match the fallback:** generate only when absent; staleness stays CI's and the pre-push hook's job, which regenerate explicitly.

`experimental = ["setup-scripts"]` is required — cargo-nextest 0.9.133 still gates setup scripts behind it.

## Correction: `filter = 'all()'` emptied `soak` and `sweeps`

The first revision of this PR argued for `all()` on the grounds that the script is "a near-free `stat` when the fixtures already exist, so there is no benefit to a fragile per-consumer filter". That premise holds only where the fixtures **can** be generated, and in two jobs they cannot.

`soak` and `sweeps` check out with no submodule and run pre-built binaries out of a nextest archive (`scripts/run_nextest_archive.sh`). The setup script fired there anyway, spent 68.7 s – 88.7 s compiling the generator in jobs whose entire design is to do no building, and then died on the one thing those jobs cannot supply:

```
error: no HGVS strings harvested from assets/hgvs-nomenclature — the spec checkout is empty.
  SETUP FAIL [  88.702s] pregenerate-spec-fixtures: scripts/pregenerate-spec-fixtures.sh
  Cancelling due to setup script failure
     Summary [  88.703s] 0/54 tests run: 0 passed, 80 skipped
```

Measured on run 31931487583 (`pull_request`, head `9f95d870`): **8 idempotency-soak shards and all three exhaustive sweeps ran zero tests**, and the required `Test` rollup reported `soak = failure`, `sweeps = failure`.

And where it succeeded, it did nothing: `Test (1/6)` checks out `submodules: true` and CI already pregenerates the fixtures explicitly, so the script reported `SETUP PASS [0.007s]`. So the `all()` binding cost two jobs their whole selection and bought CI nothing — its only real beneficiary is a local `cargo nextest run`, which a scoped filter serves identically.

**Why scoping, and not "make the script skip when the spec checkout is empty".** Skipping would keep `all()` and stop the hard failure, but it would also convert a genuinely broken submodule — in a job that *does* need the fixtures — into a later and much less clear failure inside a test body. Scoping keeps the loud, correctly-worded failure exactly where a fixture is actually required, and removes the script entirely from the jobs that neither need one nor could build one. It is also strictly cheaper: `soak` and `sweeps` no longer pay the 68.7 s – 88.7 s compile at all.

**The consumer set is four modules, not two.** `CLAUDE.md` said "two test modules actually read it" and named only `hgvs_spec_normalization_tests` and `idempotency_tests`. `normalize_axis_preserving` reads it too — `ci.yml`'s `censuses` job downloads the `spec-fixtures` artifact precisely for it — and `spec_enumeration_tests` reads the enumeration fixture. A filter written from that prose list would have been wrong on its first day, so this PR corrects the prose and, more to the point, stops depending on it.

**`tests/it/spec_fixture_setup_filter.rs`** derives both sides instead:

1. `the_setup_script_is_bound_to_exactly_the_fixture_consumers` — the filter's `test(...)` terms must equal the set of `tests/it` modules calling `ensure_spec_fixture()` / `ensure_spec_enumeration()`. Equality, not containment: a missing name reintroduces the race, an extra one is how `soak` and `sweeps` were emptied.
2. `no_fixtureless_archive_job_selects_a_spec_fixture_consumer` — parses `ci.yml`, expands the workflow `env:` into each job's `run:`, and fails if any job that runs from a nextest archive **without** downloading `spec-fixtures` selects a consumer. This is what catches a future `*_proptest` module that reads the fixture and is swept into `soak`'s `test(proptest)` without anyone naming it.

Both were sabotaged and observed failing rather than assumed to work — dropping `test(normalize_axis_preserving)` from the filter, and adding `test(idempotency_tests)` to `soak`'s selection.

## Also fixed: the temp-file cleanup never ran

The script's `trap "rm -f '$tmp'" RETURN` did not fire. Under `set -e` a failing command inside a function makes the shell **exit** rather than return, so the `RETURN` trap is skipped — in exactly the case it was written for. The failing run above left a real `tests/fixtures/grammar/hgvs_spec_normalization.json.470DPv` behind in the checkout, and an isolated repro (`false` in place of the `cargo run`) reproduces it. Replaced with an `EXIT INT TERM` trap over a scalar (an empty array under `set -u` is an "unbound variable" error on bash 3.2, which is `/bin/bash` on macOS); verified by running the script where generation must fail and confirming no temp survives.

## Verification

The first revision's verification was entirely local-foreground, where the submodule is present and `cargo run` works, and recorded no CI observation — which is why the archive-job breakage shipped. This one reproduces the failure first.

**1. The failure, reproduced before the fix.** A detached worktree with `assets/hgvs-nomenclature` uninitialised and both fixtures absent, running `soak`'s exact command against an archive built elsewhere:

```
scripts/run_nextest_archive.sh --archive-file nextest-soak.tar.zst --workspace-remap . \
  --extract-to . --extract-overwrite --no-tests=fail -E 'test(proptest)'

  SETUP FAIL [ 145.905s] pregenerate-spec-fixtures: scripts/pregenerate-spec-fixtures.sh
  Cancelling due to setup script failure
     Summary [ 145.906s] 0/33 tests run: 0 passed, 101 skipped
EXIT=105
```

Same `0/33` as CI; the wall clock is longer only because the scratch tree compiled the generator cold.

**2. The same invocation after the fix**, same tree, submodule still empty and fixtures still absent:

```
    Starting 33 tests across 1 binary (101 tests skipped)
     Summary [  12.761s] 33 tests run: 33 passed, 101 skipped
EXIT=0
```

No `SETUP` line at all — the script is not merely skipping, it is not selected.

**3. The feature still works — fixtures absent, consumer selected.** Both gitignored artifacts deleted first, so the generation path is genuinely exercised:

| selection | result |
|---|---|
| `-E 'test(idempotency_tests)'` | `SETUP PASS [81.244s]`, generated **both** fixtures, `6 tests run: 6 passed`, `EXIT=0` |
| `-E 'test(axis_is_preserved_over_the_spec_normalization_corpus)'` | `SETUP PASS [101.606s]`, generated both, `1 test run: 1 passed`, `EXIT=0` |
| same selection, fixtures now present | `SETUP PASS [0.026s]`, no regeneration, `EXIT=0` |

The second row is `normalize_axis_preserving` — the consumer the two-module prose list omitted, and the one a naively-scoped filter would have dropped.

**4. Gate**, read back from each command's own `EXIT=` line rather than from a completion notification: `cargo fmt --all --check`, `cargo clippy --all-features --all-targets -- -D warnings`, `actionlint`, `cargo nextest run --features dev --no-fail-fast`, and `cargo nextest run -p ferro-hgvs-soak-tests`. `shellcheck scripts/pregenerate-spec-fixtures.sh` is clean.


nh13:
@coderabbitai pause

PR #2038, opened 2026-08-16
perf(normalize): bound insertion→dup/repeat rotation scan to linear time

### LINKED ISSUE #1846: normalize: insertion_to_duplication is O(payload²) in the expanded insert, so a megabase cross-reference ins never terminates

nh13:
Closes #1846.

Representation-Change: none. Output-preserving by construction — the early exit returns `None` exactly where the pre-existing loop was already determined to return `None`; classification is unchanged, verified by unit tests (the outgrows-window decline and the exactly-fills-window conversion both hold before and after). The guard fires only for a repeat unit longer than the fetched shuffle window, a payload size the synthetic corpus cannot build, so a corpus diff is a structural zero rather than a measurement.

## The defect

`insertion_to_duplication` (and its deliberate symmetric sibling `insertion_to_repeat`) scan every cyclic rotation of the *expanded* insert payload, rebuilding a payload-sized `Vec` on each of `unit.len()` iterations — Θ(L²) bytes copied and L allocations of L bytes. A spec-legal `ins[ACC:g.A_B]` resolves the cross-reference to the whole named range, so a megabase insert never terminates and grows RSS to gigabytes. The two corpus rows that stall (per #1846):

- `NC_000004.12:g.134850793_134850794ins[NC_000023.11:g.89555676_100352080]` (payload 10.8 MB)
- the same with a trailing `inv`

Measured on the unfixed code (debug unit test, aperiodic single-copy payload): ~1.0 s at L=200 kb → ~5.7 s at L=400 kb, i.e. ~5.6× per doubling — quadratic. Extrapolated, the corpus's 10.8 MB payload is ~52 min per row.

## The fix

`find_tandem_extent` skips every candidate anchor once `anchor + unit.len() > ref_seq.len()`. Every rotation has length `base_unit.len()`, so when the unit is longer than the reference window **no** rotation can match any anchor — `best` stays `None` and the function is already determined to return `None`, after paying L allocations of L bytes for it. Bail before the loop:

```rust
if base_unit.len() > ref_seq.len() {
    return None;
}
```

Applied to **both** functions — `insertion_to_repeat` carries the identical loop, reached when the payload is 2+ copies of an oversized unit (an aperiodic payload returns earlier via the `added_copies == 1` route to the dup path). Detection goes from Θ(L²) to O(L). `ref_seq` here is the fetched shuffle window (default 100 bases, grown by `normalize_in_grown_window`), so the guard fires only far above any window the normalizer actually uses.

## Unchanged on purpose (the discriminating cases)

- The guard is strict `>`, not `>=`: a unit whose length exactly equals the window still matches at anchor 0 (`0 + unit.len() == ref_seq.len()`, which `find_tandem_extent` accepts), so it is a real single-copy duplication. `insertion_to_duplication_converts_when_the_unit_exactly_fills_the_window_1846` pins that it still converts — a `>=` over-generalization would silently decline it while passing every other test.
- `insertion_conversions_decline_when_the_unit_outgrows_the_window_1846` pins that both paths decline when the unit outgrows the window (the classification the early exit preserves).

## Verification

```
cargo fmt --all --check                                   # clean
cargo clippy --all-features --all-targets -- -D warnings  # clean
cargo nextest run --features dev --no-fail-fast           # 11173 passed, 0 failed, 155 skipped
```

The two 2 Mb / 2×1 Mb payload guards (`*_does_not_rescan_a_megabase_payload_1846`) complete in ~0.04 s each on the fix and would hang on the quadratic version.

### DIFF
--- a/src/normalize/rules.rs
+++ b/src/normalize/rules.rs
@@ -358,6 +358,19 @@ pub fn insertion_to_repeat(
         return None;
     }
 
+    // #1846: the same early exit `insertion_to_duplication` carries, for the
+    // same reason. This loop rotates and rebuilds a `base_unit.len()`-byte `Vec`
+    // once per base of the unit; when the unit outgrows `ref_seq` every
+    // rotation's `find_tandem_extent` returns `None`, so `best` is already
+    // `None` and the Θ(unit²) work is guaranteed futile. An aperiodic payload
+    // returns above (`added_copies == 1` routes to the dup path), so this fires
+    // only for a payload that really is 2+ copies of an oversized unit — e.g.
+    // two copies of a multi-megabase unit — which the `added_copies` guard does
+    // not bound. See the fuller note on `insertion_to_duplication`.
+    if base_unit.len() > ref_seq.len() {
+        return None;
+    }
+
     // The c.-context codon gate (repeated.md: unit_len % 3 == 0) is applied
     // inside `normalize_repeat` below, not here — see the delegation note.
 
@@ -742,6 +755,23 @@ pub(crate) fn insertion_to_duplication(
     let added_copies = inserted_seq.len() / base_unit.len();
     if added_copies != 1 {
         // 2+-copy insertions are handled by `insertion_to_repeat`.
+        return None;
+    }
+
+    // #1846: bail before the rotation scan when the unit cannot fit the window.
+    // Every rotation below has length `base_unit.len()`, and `find_tandem_extent`
+    // skips every anchor once `anchor + unit.len() > ref_seq.len()` — so a unit
+    // longer than `ref_seq` matches no anchor in any rotation, and `best` is
+    // already determined to stay `None`. Without this exit the loop still runs
+    // `base_unit.len()` iterations, each allocating and filling a
+    // `base_unit.len()`-byte `Vec`, i.e. Θ(L²) bytes copied to reach that fixed
+    // `None`. A spec-legal `ins[ACC:g.A_B]` expands to a payload of the named
+    // range's length (a 38-char cross-reference named 10.8 MB), so an unbounded
+    // scan here never terminates on such input; the early exit makes it O(L).
+    // `ref_seq` is the fetched shuffle window (default 100 bases, grown by
+    // `normalize_in_grown_window`), so this fires only far above any window the
+    // normalizer actually uses and cannot change which inputs convert to a dup.
+    if base_unit.len() > ref_seq.len() {
         return None;
     }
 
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #2039, opened 2026-08-16
test(normalize): re-arm the cis guards #1835 made vacuous

nh13:
This is the **test half of #1808**, split out on the operator's decision. The `src/` half — the new `prefer_minimal_alignment` rule in `src/normalize/merge.rs` — stays on #1808, together with the ledger rationale and the contract-doc render that quote its headline figures. Once this lands, #1808 rebases onto it and its diff collapses to the rule alone.

Nothing here depends on that rule, and that is measured rather than assumed: this branch is cut from current `main`, which does **not** carry it and is 82 commits past #1808's merge-base, and every carried row passes there. See *Verification* below.

## What it restores, and why the guards were empty

Six rows across four modules were re-pinned by the partition default flip (#1835) onto their own merged output and stopped guarding anything. Each of them used a *distant* third member to stop the sequence-first derivation, on the reasoning that a block carrying more than one run of change would be declined. After the flip the whole allele is derived however far apart its members sit, so every one of those rows collapsed to a single insertion beside the third member — with no tie left to break, no sibling to bar, and no `dup` in the output for the over-clamp guard to bound.

Their own comments say so and ask for the replacement this PR supplies:

> Re-establishing it needs a shape whose members survive the derivation; this core no longer supplies one.

`MAX_CANONICAL_WINDOW` (4096) refuses a window that wide *before* any alignment runs, so a third member placed past it blocks the derivation **structurally** rather than by distance — and nothing about that refusal can be undone by a better alignment. The per-member pipeline, where `clamp_sibling_crossing_shifts`, `clamp_sibling_crossing_junctions` and `cis_member_order_key` all live, then decides the output again.

`common/synthetic.rs` gains `extended_core`, which extends a core with `ACGT` filler. The filler continues the pad's own period-4 rotation and the pad's first base is `A`, so the bases immediately either side of the core are byte-identical to what `padded` already gives it — only the distance to the far end grows, and nothing about a variant inside the core can move. `cis_junction_crossing_shift` keeps a local builder instead, because its tract has to sit at the head of the sequence rather than inside `padded`'s 256-base pad; its `CGGT` filler is *not* run-free, so the pinned deletion position is derived from the run structure rather than picked, and the doc says which property it rests on.

## Verification

**Every row passes without the rule.** 68 tests across the six touched modules, run on this branch against a `main` that has no `prefer_minimal_alignment`: 68 passed, 0 failed. No row carried over from #1808 turned out to be evidence for the rule.

**Each restored guard was verified by mutation, not by timing** — a guard that loads its fixture and then stands down costs the same wall time as one that ran, so duration proves nothing here. In each case the mutation is applied, the row goes red, and the source is restored byte-for-byte (`shasum -a 256` before and after, clean tree at the end). The same mutation was then run against `main`'s version of the row, which is the half that shows the guard really was empty:

| mutation | restored row | on `main`'s fixture |
|---|---|---|
| drop `.filter(\|_\| a.junction.is_none())` from the 5' branch of `clamp_sibling_crossing_shifts` | `cis_junction_crossing_shift::a_third_member_clear_of_the_tract_keeps_the_duplication_reaching_its_five_prime_most_position` — **RED**, the `dup` stranded at `5_6dup` instead of reaching `4_5dup` | passes |
| stub `clamp_sibling_crossing_junctions` to return immediately | `issue_1304_junction_barrier_snapshot::a_third_member_clear_of_the_tract_keeps_the_pair_barred` and `::the_barred_pair_does_not_depend_on_authored_order` — **RED**, output `g.[261_262dup;265dup;4500del]`, the `insGA` swept past its sibling | all 5 rows in the module pass |
| neutralize `junction_rank` in `cis_member_order_key` | `issue_1301_adjacent_gap_member_order::an_insertion_still_sorts_before_a_duplication_sharing_its_span_beside_a_third_member` and `issue_1320_dup_spans_sibling_junction::a_junction_at_the_dups_five_prime_end_is_left_alone_beside_a_third_member` — **RED** | pass |
| take `>=` instead of `>` at `dup.start` in the `swallows_a_junction` test | `issue_1261_cis_member_order::two_duplications_sharing_a_start_render_in_junction_order_beside_a_third_member` and `issue_1320_dup_spans_sibling_junction::two_duplications_sharing_a_start_keep_both_dup_spellings_beside_a_third_member` — **RED**, `258_259dup` re-spelled to `259_260insAG` | pass |

Worth recording from the second row: with `clamp_sibling_crossing_junctions` stubbed out, **`main`'s entire `issue_1304` module is green** — including `a_moved_sibling_shape_now_merges_from_the_sequence`, whose own comment claims to be "the only row in this file that still goes red when the mechanism is removed". That claim is stale on `main`; after this PR two rows in the file go red.

The order-independence rows (`the_shared_start_order_beside_a_third_member_is_authored_order_independent` and its `issue_1301` twin) stay green under every mutation above, which is correct — they compare a forward and a reverse authoring of one allele, so both sides move together. They are re-fixtured here so that they compare a three-member output rather than a merged one-member output, but they are independence checks, not value pins.

## What was deliberately left on #1808

Three of #1808's test-side hunks restore no coverage and are the rule's own narrative, so they stay with the rule rather than being imported here:

- `tests/it/spec_corpus_regressions.rs` — a rename plus removal of a tuple column that is already redundant on `main`; net one assertion fewer, and its prose re-attributes to `prefer_minimal_alignment` a convergence that `main` already has from #1835.
- the rename of `cis_junction_crossing_shift::the_three_member_spelling_and_its_one_member_form_are_two_fixed_points` and its surrounding prose rewrite — same reason; the row already asserts convergence on `main`.
- every sentence crediting `prefer_minimal_alignment` for the vacuity. The vacuity is #1835's, which the base versions of these very comments state in as many words, and this branch has no such symbol to name.

Refs #1808, #1835, #1235. Not marked as closing #1235 — the rule half is still outstanding.

Representation-Change: none

No file under `src/` is touched — the whole diff is six files under `tests/it/` — so no normalized output can move.

### DIFF
--- a/tests/it/cis_junction_crossing_shift.rs
+++ b/tests/it/cis_junction_crossing_shift.rs
@@ -296,25 +296,38 @@ fn a_multi_base_duplication_beside_an_insertion_merges_from_the_sequence() {
 /// only oracle is `apply`, and their second member is always a single-base `dup`,
 /// exactly the length at which `len - 1` is zero and the over-clamp disappears.
 ///
-/// The third member is what keeps the property visible. `15del` stops the
-/// *merge* without touching the shift: the block is no longer one run of change,
-/// so the derivation declines to collapse the allele and the per-member pipeline
-/// decides the output again, which is where the barrier lives. The `dup` then
+/// A third member is what keeps the property visible: it has to stop the
+/// *merge* without touching the shift, so that the per-member pipeline — which
+/// is where the barrier lives — decides the output again. The `dup` then
 /// reaches `4_5dup` and the spelling is observable.
 ///
-/// **Distance measured, not assumed.** Sweeping every third-member position on
-/// `DUP_RUN`: at `10del` and `11del` the member joins the block and the allele
-/// merges again (to `g.[4_5insC;10_11insT]`) — a test written there would be
-/// vacuous in a new way, pinning a merged form while claiming to pin a barrier.
-/// `12del` and `13del` are clear but only 1-2 nt past the tract; at `14del` and
-/// `18del` the third member itself 5'-shifts, which muddies the row. `15del` is
-/// 4 nt clear, stays exactly where it is written, and reaches the tie.
+/// # The fixture moved, because distance stopped being what blocks the merge
 ///
-/// **Fails both ways.** Deleting `.filter(|_| a.junction.is_none())` from the 5'
-/// branch of `clamp_sibling_crossing_shifts` turns this output into
-/// `g.[4_5insC;5_6dup;15del]`; restoring it turns it back. (Removing the whole
-/// 5' `across_junctions` barrier instead leaves this row green — that half is
-/// guarded by `an_insertion_merging_with_a_deletion_keeps_its_base_in_place`.)
+/// This row used to read `DUP_RUN` with `15del` as the third member, on the
+/// reasoning that a member 4 nt clear of the tract broke the block into more
+/// than one run of change and made the derivation decline. The partition
+/// default flip (**#1835**) ended that: the whole allele is now derived from the
+/// resulting sequence and collapses to `g.[3_4insACT;15del]`, which
+/// [`the_three_member_spelling_and_its_one_member_form_are_two_fixed_points`]
+/// below pins as the adjudicated answer. **At that fixture this row became
+/// vacuous** — with no `dup` in the output at all, deleting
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

PR #2040, opened 2026-08-16
fix(reference): synthesize a record-less parent block-by-block across alignment gaps

### LINKED ISSUE #1926: synthesize_parent_sequence fetches a contiguous slice across an LRG <diff> gap
`synthesize_parent_sequence` maps two endpoints and then fetches a **contiguous** chromosome slice, so any gap in the LRG `<diff>` list between those endpoints is not accounted for in the slice it takes.

## The two directions

| gap kind | requested window | bases returned |
|---|---|---|
| `lrg_ins` | 22 | **20** |
| `other_ins` | 20 | **22** |

Either way, every base past the gap is shifted, and the resulting sequence is well-formed and in-bounds — so nothing downstream refuses it.

## Exposure

Latent on current reference data: **0 of 1294** LRG records reach it. It is a second, independent route to the same class of wrong answer that the `<diff>`-list modelling work is closing on the primary route, so it is worth closing before a record appears that exercises it.

## Why it is filed separately

Found while reviewing #1847, which fixes the primary projection route. Widening that PR to cover this would have pushed an already 13-file change further; the fix here is independent and can be reviewed on its own terms.

## Suggested direction

Walk the gap list when computing the slice, rather than assuming the parent interval between two mapped endpoints is contiguous — i.e. the same modelling the projection path now uses, applied to sequence synthesis.

nh13:
Closes #1926.

## The defect

`synthesize_parent_sequence` (`src/reference/multi_fasta.rs`) serves the bases of a genomic parent (`NG_`/`LRG_`) that has a chromosomal placement but no FASTA record of its own, by mapping the requested parent-frame window onto the chromosome and reading those `NC_` bases. It mapped only the **two endpoints** of the window and then fetched one **contiguous** `NC_` slice between them — so any LRG `<diff>` gap between the endpoints was sliced straight across. The result is well-formed, in-bounds, and wrong, with every base past the gap shifted:

| gap kind | requested window | bases returned (buggy) |
|---|---|---|
| chromosome-only (`other_ins`) | 20 | **22** (the chromosome-only run swept in) |
| parent-only (`lrg_ins`) | 22 | **20** (the parent-only run silently dropped) |

Nothing downstream refuses it, so a normalizer would happily shift against fabricated/shifted bases.

## The fix

Walk the placement's aligned blocks instead of assuming the parent interval between two mapped endpoints is contiguous — the same block model the projection path already uses (`GenomicPlacement`'s gap list, #1833).

- New `GenomicPlacement::parent_window_nc_ranges(parent_lo, parent_hi)` returns the ascending chromosome sub-ranges to fetch for a parent window, in parent-ascending (output) order. A **chromosome-only** run inside the window is dropped (the ranges skip it); a window whose coordinates include a **parent-only** run — bases the chromosome does not carry — returns `None`, since they cannot be synthesized. Both are caught by a covered-count check (the intersecting blocks must account for every parent coordinate). All arithmetic is checked, so a public, malformed gap list fails closed rather than answering from a half-applied walk.
- `synthesize_parent_sequence` assembles the window from those ranges, reverse-complementing each range in block order on the minus strand, and declines (`InvalidCoordinates`) when the plan is `None` — matching the existing precedent that unplaced coordinates must error rather than be fabricated.

## Unchanged on purpose

The **gapless** case yields a single range identical to the previous contiguous slice, so every gapless placement — all `NG_` placements and 1 248 of the 1 294 LRG records — is byte-identical. The parent-only decline and the chromosome-only drop are pinned as exact-string / `is_err` regression tests (plus + minus strand), and a focused unit test pins `parent_window_nc_ranges` in isolation (gapless equivalence, gap drop, parent-only decline, out-of-span decline).

## Representation stability

Representation-Change: none. The only behavioural change is to the gapped record-less parent-synthesis path. The gapless path is byte-identical (a single fetch range), and **0 of the 1 294** shipped LRG records reach the gapped path (all have their own `LRG_<N>g` record; every `NG_` placement is gapless). No shipped input's normalized output moves; the fix corrects only a latent wrong-but-well-formed sequence. Off the shipped reference, two movements do occur and are stated here rather than left implicit: a window spanning a chromosome-only run is previously-accepted-now-different (it serves the correctly-aligned bases instead of a shifted slice), and a window crossing a parent-only run is previously-accepted-now-refused. Neither is reachable on any reference this project ships. `spec_conformance_axis` (the corpus census) is `JsonProvider`-backed and never exercises LRG placement synthesis, so it is untouched.

## Verification

- `cargo nextest run --features dev --no-fail-fast` — **11173 passed, 0 failed**, 155 skipped.
- `scripts/run_oracle_suite.sh` (mirrors CI's `test-oracle` job) — **11013 passed, 0 failed**, 315 skipped.
- The bare `FERRO_ASSERT_IDEMPOTENT=1 …` full-suite run shows the 7 failures that are red on clean `origin/main` and excluded by `ORACLE_EXCLUDE` (4 `defect_non_idempotent_outputs`, `spec_corpus_regressions::an_insertion_at_the_cds_end_is_not_a_fixed_point`, the 2 `spec_conformance_axis` censuses) — pre-existing, not introduced here.

### DIFF
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -1988,34 +1988,43 @@ impl MultiFastaProvider {
         // named `LRG_<N>g`); 164 of them were quietly served chromosome bases in
         // place of their frozen record. The callers now gate on `own_record`.
         //
-        // Map both endpoints onto the chromosome; on a minus-strand placement
-        // the parent runs antiparallel, so the low parent position maps to the
-        // *higher* `NC_` coordinate.
-        let (nc_a, nc_b) = match (
-            placement.parent_to_nc(start + 1),
-            placement.parent_to_nc(end),
-        ) {
-            (Some(a), Some(b)) => (a, b),
-            _ => {
-                return Err(FerroError::InvalidCoordinates {
-                    msg: format!(
-                        "parent window [{start}, {end}) falls outside the placed span on {}",
-                        placement.nc.full()
-                    ),
-                });
+        // Assemble the window block-by-block through the placement's alignment
+        // rather than as one contiguous `NC_` slice (#1926). A prior version
+        // mapped only the two endpoints and fetched everything between them,
+        // which silently sliced *across* any `<diff>` gap: a chromosome-only
+        // (`other_ins`) run got swept in (too many bases) and a parent-only
+        // (`lrg_ins`) run got dropped (too few), shifting every base past the
+        // gap. `parent_window_nc_ranges` walks the gap list, returning the
+        // chromosome sub-ranges that skip chromosome-only runs, and declining
+        // (via `None`) a window whose parent-only bases have no chromosome
+        // image to synthesize from. Latent on the shipped reference — 0 of
+        // 1 294 LRG records reach this path — but a wrong-but-well-formed
+        // sequence is exactly the class this guards against.
+        let ranges = placement
+            .parent_window_nc_ranges(start + 1, end)
+            .ok_or_else(|| FerroError::InvalidCoordinates {
+                msg: format!(
+                    "parent window [{start}, {end}) falls outside the placed span, or crosses a \
+                     parent-only alignment gap whose bases are absent from {}",
+                    placement.nc.full()
+                ),
+            })?;
+        let mut synthesized = String::new();
+        for (nc_lo, nc_hi) in ranges {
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #2041, opened 2026-08-16
fix(spdi): decline an anchored repeat when the tract span cannot be verified

### LINKED ISSUE #1497: An anchored repeat converts on an unverified unit-wide span when get_sequence_length fails
When `MultiFastaProvider`-style base retrieval succeeds but `get_sequence_length` does not, `spdi::convert::resolve_repeat_tract_span` returns the unit-wide span at the anchor without ever searching for the tract. The caller then validates that span and, because a one-unit span trivially satisfies both checks, **converts** — emitting a triple that denotes one copy of the unit when the reference may hold a longer run.

## Mechanism

```rust
let Ok(sequence_length) = provider.get_sequence_length(accession) else {
    return Ok((fallback.0, fallback.1, RepeatSpanOrigin::NotSearched));
};
```

`fallback` is `[anchor, anchor + unit_len - 1]`. The caller then checks divisibility (a unit-wide span is exactly one unit, so it passes) and unit-match (`del_str == unit_str.repeat(1)`, i.e. one unit against itself, so it passes). Conversion succeeds.

That is the same silent truncation class as #1452: `g.259CAG[5]` over a 3-copy tract yields `258:CAG:CAGCAGCAGCAGCAG`, applying to seven copies rather than five, because the two tract units the span left untouched survive the replacement. The difference is only how the search was defeated — there by case, here by a provider that cannot report a length.

## Why it is not caught

Well-formed, in bounds, and a fixed point, so none of `FERRO_ASSERT_REPARSE`, `FERRO_ASSERT_IN_BOUNDS` or `FERRO_ASSERT_IDEMPOTENT` fires. It is the same oracle blind spot #1452 documented.

The existing comment calls this "a known and stated limit, not a silent wrong span — the caller's divisibility and unit-match checks still judge it." That is the part worth revisiting: those checks cannot judge it. A unit-wide span passes both **by construction**, so they reject nothing on this path.

## Reachability

`RepeatSpanOrigin::NotSearched` is reached when the unit is empty, or when `get_sequence_length` errors while `get_sequence` works. Any provider implementing the two independently can be in that state — a partial or lazily-indexed provider, or one whose length lookup depends on an index the base fetch does not need.

Pre-existing; #1496 only names the state (it added `RepeatSpanOrigin` for diagnostics and did not change this control flow). Filed separately rather than widening that PR.

## Suggested shape

Decline rather than convert when the tract was never verified:

- return a `ConversionError` for a single-position anchor whose search could not run, instead of falling back to the unit-wide span — the caller asked for "the run at this anchor" and no run was established;
- keep the existing fallback for the **explicit-range** spelling, where the span is the caller's own and needs no search;
- add a provider regression whose `get_sequence` succeeds and whose `get_sequence_length` fails, asserting the anchored spelling declines and the range spelling still converts.

Note the empty-unit arm of `NotSearched` is separate and already rejected downstream, so only the length-lookup arm needs the new behaviour.

nh13:
Closes #1497.

Representation-Change: none. The changed branch fires only when `get_sequence_length` returns `Err`, and every in-repo `ReferenceProvider` (FastaProvider, CachedFastaProvider, MmapFastaProvider, MultiFastaProvider, JsonProvider, PyProvider, WindowProvider, AuditedProvider) overrides it to succeed whenever it can serve bases. Only the trait default errors, and no real provider leaves it in place. The corpus harness and every normalization/conversion path use those providers, so 0 rows can move — a structural zero. On a real provider a genuinely-missing accession now declines with `UnsupportedEditType` instead of the downstream `MissingReferenceData` fetch failure it used to hit — a decline→decline message change. For a provider that serves bases but cannot report a length (the always-erroring trait default, reachable only out of tree), the change is previously-accepted→now-refused, and it does cover correct answers as well as wrong ones: on `NC_TEST.1 = GGCAGCAGCAGGG`, `g.3C[2]` names a genuinely one-copy run where the unit-wide span was the right span, and it previously returned the correct `position 2, deletion C, insertion CC` — identical to a healthy provider's answer — where it now declines. Refusing is still the correct conservative call, because without a length the right answers cannot be separated from the truncating ones (a 5-copy tract collapsing to 1), but the loss is real and is disclosed here rather than described as message-only.

## The defect

A single-position repeat anchor (`g.263A[7]`, `g.3CAG[5]`) names the *start of a tandem run*, so `spdi::convert::resolve_repeat_tract_span` searches for the run's full extent and converts over it. That search needs the sequence length to clamp its window. When `get_sequence_length` failed, the function fell back to the unit-wide span *at the anchor* and returned `RepeatSpanOrigin::NotSearched`.

The caller then validated that span. A unit-wide span is exactly one unit, so it passes divisibility (`len % unit_len == 0`) and unit-match (`del == unit.repeat(1)`, one unit against itself) **by construction** — the two checks the old comment claimed still judged it cannot. Conversion succeeded on an unverified span.

Measured (`NC_TEST.1 = GGCAGCAGCAGGG`, a 3-copy `CAG` tract at 3..=11), with a provider that serves bases but fails length:

| spelling | before | after |
|---|---|---|
| `g.3CAG[5]` (anchored) | `2:CAG:CAGCAGCAGCAGCAG` — 5 copies over 1, leaving 2 → **7 copies** | declines (`UnsupportedEditType`) |
| `g.3_11CAG[5]` (range) | `2:CAGCAGCAG:CAGCAGCAGCAGCAG` (correct) | unchanged — still converts |

Same silent-truncation class as #1452, defeated here by a provider that cannot report a length rather than by soft-masking. Well-formed, in bounds, and a fixed point, so none of the reparse/in-bounds/idempotency oracles fires — the blind spot #1452 documented.

## The fix

When `get_sequence_length` fails, decline with `UnsupportedEditType`, mirroring the sibling window-overflow decline ("spell the whole range explicitly `g.<start>_<end>N[n]`"). The explicit-range spelling carries its own span, never reaches this function, and is unaffected.

## Unchanged on purpose

The other `NotSearched` arms are left as-is because none produces a wrong conversion:

- **empty unit** — declines downstream on the divisibility check (unit length 0), as #1497 notes;
- **empty sequence / anchor past the end** — the clamped fallback yields a reversed/empty interval, so `fetch_reference_bases` declines with an invalid-interval or missing-data error.

Only the length-lookup arm converted on a valid-but-unverified span, so only it needed the guard.

## Verification

- New unit tests in `src/spdi/convert.rs`: `an_anchored_repeat_declines_when_the_length_lookup_fails` (anchored declines, range still converts, via a `LengthlessProvider` whose `get_sequence`/`get_genomic_sequence` succeed while `get_sequence_length` errors) and the control `an_anchored_repeat_converts_over_its_full_run_when_the_length_is_known`. Both watched to fail before the fix.
- `cargo nextest run --features dev --no-fail-fast` — 11196 passed, 155 skipped.

### DIFF
--- a/src/spdi/convert.rs
+++ b/src/spdi/convert.rs
@@ -2439,10 +2439,14 @@ enum RepeatSpanOrigin {
     /// The search ran and found no run at the anchor, so the span is the
     /// unit-wide fallback. This is the case that must not quote its span.
     NoRunAtAnchor,
-    /// The search never ran — an empty unit, or a provider that serves bases
-    /// but not `get_sequence_length`. The span is a placeholder rather than a
-    /// judgement, so the generic diagnostic remains the honest one: nothing was
-    /// measured that would justify saying "no run begins here".
+    /// The search never ran on an input the function does not itself refuse —
+    /// an empty unit, an empty sequence, or an anchor past the sequence end. The
+    /// span is a placeholder rather than a judgement, so the generic diagnostic
+    /// remains the honest one: nothing was measured that would justify saying
+    /// "no run begins here". A provider that cannot report a length is *not*
+    /// here — that case declines outright (#1497), because the unit-wide span it
+    /// would otherwise hand back passes the caller's checks by construction and
+    /// converts to a silently truncated triple.
     NotSearched,
 }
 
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #2042, opened 2026-08-16
fix(data): refuse a malformed cdot exon row instead of dropping it

### LINKED ISSUE #1924: A truncated cdot exon row displaces CDS bounds forward instead of being refused
A cdot exon row with fewer fields than expected displaces the derived CDS bounds **forward** instead of being refused.

## Measured

On a record whose last exon row is truncated:

| | `(cds_start, cds_end)` |
|---|---|
| intact record | `(10, 180)` |
| truncated, before #1737 | `(10, 180)` — identical to intact |
| truncated, after #1737 | `(150, 320)` |

In the post-#1737 case `cds_end` of 320 overruns the last exon's `tx_end` of 240 by **80 bases**. The direction is forward displacement, not narrowing — worth stating explicitly because the natural assumption is that a short row would produce a truncated or narrowed span.

## Exposure on current data is zero, and that zero is structural

Across **12,207,712** exon rows in **1,254,949** builds, every row carries exactly 6 fields. So no shipped record can reach this today, and the zero is a property of the shipped data rather than of the code path.

That makes this a robustness issue rather than a live defect: the guard that would have made a malformed row harmless (the pre-#1737 behaviour, where a truncated row produced the same answer as an intact one) no longer holds, and nothing refuses the row.

## Suggested direction

Refuse a row that does not carry the expected field count, rather than deriving bounds from a partial one. A refusal is checkable; a forward-displaced coordinate is well-formed, in-bounds-looking, and silent.

## Scope

Surfaced by the CodeRabbit lens while reviewing #1737 and confirmed by measurement. The exposure is introduced by that PR; the PR is otherwise verified sound (ORF scoring 0/42 valid before, 42/42 after, 0 regressions).

nh13:
Closes #1924.

Representation-Change: none. Refuses only malformed cdot exon rows (truncated, or a non-numeric value where a coordinate belongs); every one of the 12,207,712 exon rows across 1,254,949 builds in the three shipped cdot-0.2.32 files (RefSeq and Ensembl GRCh38, plus GRCh37) carries all six fields, so no shipped record is affected and no previously-correct output moves. Residual, on user-supplied cdot only: a transcript carrying a malformed exon row is now refused whole and later surfaces as `ReferenceNotFound`, rather than being served with a forward-displaced CDS.

## The defect

A cdot exon row that cannot be parsed into complete placement coordinates was silently **dropped** by the `filter_map` in `RawCdotTranscript::from_genome_build` (`src/data/cdot.rs`). Two shapes reach it: a **truncated** row (fewer than the five leading fields `genomic_start, genomic_end, exon_num, tx_start, tx_end`), and an **inconsistent** row whose core coordinate is non-numeric (the `?`-in-`filter_map` on `as_u64()` dropped it just the same).

Dropping a row removes an exon from the table, which opens a gap and **displaces the derived CDS bounds forward** — a well-formed, in-bounds-looking, silently-wrong coordinate rather than a surfaced error. The silent drop was already documented (at `reference/validate.rs`); nothing refused the row.

Measured before/after on a synthetic three-exon plus-strand build whose middle row is truncated to three fields (unit `truncated_exon_row_refuses_the_transcript`):

| | result |
|---|---|
| intact (six-field rows) | loads, 3 exons, CDS `(10, 250)` |
| middle row truncated, before this PR | **loads**, 2 exons (a gap), derived from the partial table |
| middle row truncated, after this PR | **refused** (`None`) |

## The fix

Replace the `filter_map` with an explicit loop that **refuses the whole transcript** (returns `None` — the load path's existing skip signal, consumed by `if let Some(tx) = raw.to_transcript(...)`) the moment a row fails to yield all four placement coordinates, with a malformed-transcript `log::warn!` matching the sibling `from_transcripts_with_build` refusal. Refusal is checkable; a displaced coordinate is not.

The optional `gap_info` field (index 5) is deliberately **still not required**: a valid exon may omit it, and a malformed CIGAR in a present sixth field stays a local warn (`no CIGAR`) rather than refusing — so the refusal keys on placement coordinates, not on a full six-field count.

Scope is the real-cdot `genome_builds` path (`from_genome_build`) named by the issue and the census. `from_flat_format` is a separate legacy 4-field test/compat shape with no evidence of this defect and is intentionally left alone.

## Distinct from #1909 / #1737

- **#1909** added a guard in `src/convert/coding.rs` that returns a malformed-transcript `ConversionError` when a `c.*n` (3'UTR) position lands past the sequence end. That is a downstream **conversion-layer** symptom guard: it only fires for a 3'UTR position on a transcript whose `cds_end` already overruns the sequence. A forward displacement that stays **in bounds** is never caught by it. This PR fixes the **load/derivation-layer** root cause so the corrupt table is never built.
- **#1737** was coordinate-space work on how CDS bounds map through the exon table; it did not add a refusal for a partial table. The silent drop predates it.

## Unchanged on purpose (discriminating cases)

- `well_formed_exon_table_loads_with_every_exon` — a valid six-field table still loads with every exon and the correct CDS (the refusal is specific to malformed rows, not blanket).
- `exon_row_without_gap_info_field_still_loads` — a legitimately five-field row (no `gap_info` slot) still loads; the refusal is on placement coordinates, not a six-field count.

## Verification

- `cargo clippy --all-features --all-targets -- -D warnings` — clean (`EXIT=0`).
- `cargo nextest run --features dev --no-fail-fast` — `11197 passed, 0 failed, 155 skipped` (`EXIT=0`). Includes the four new units in `data::cdot::tests` and the corrected `multi_fasta` fixture.


nh13:
@coderabbitai pause

PR #2043, opened 2026-08-16
docs: mdBook user docs site with an interpretation section

nh13:
Scaffolds an mdBook user-documentation site, published on Read the Docs, following the same pattern fgumi and bwa-mem3 use (mdBook + mdbook-linkcheck2, navy theme, RTD builds via `.readthedocs.yaml`, GitHub CI validates only).

Contents:
- `docs/book.toml`, `docs/src/{SUMMARY.md,index.md}`, and the shared Fulcrum Genomics house theme (`docs/theme/custom.css` + `docs/theme/sidebar-brand.js`, matching fgumi/bwa-mem3).
- An **"under construction" Interpretation section** (`docs/src/shadow-spec/index.md`) — a per-clause reading of the HGVS recommendations recording how ferro interprets each one. Pages land one spec page at a time.
- `.readthedocs.yaml` (pinned mdBook 0.5.2 + mdbook-linkcheck2 0.12.2) and a `Docs` CI workflow that builds the book as a PR gate (dead internal links fail via linkcheck2).

No behavior change; docs and CI only. The first interpretation page (DNA substitution) is stacked in a follow-up PR. The `shadow-spec/` path and the `shadow_spec` test identifiers are the section's internal codename; the user-facing name is "Interpretation".

Representation-Change: none. Docs and CI only; no watched source file is touched.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,6 +26,8 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          persist-credentials: false
 
       - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8  # stable
         with:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,8 +1,9 @@
 name: Docs
 
-# Builds the mdBook user docs as a PR gate. Read the Docs does the actual hosting/deploy
-# via .readthedocs.yaml; this job only validates that the book builds and has no dead
-# internal links (mdbook-linkcheck2). Tool versions are pinned in lockstep with
+# Builds the mdBook user docs on pull requests. Read the Docs does the actual hosting/deploy
+# via .readthedocs.yaml; this job validates that the book builds and that internal links
+# resolve to real pages (mdbook-linkcheck2 — link targets, not fragment anchors). It is
+# advisory unless added to the branch ruleset. Tool versions are pinned in lockstep with
 # .readthedocs.yaml.
 
 on:
@@ -13,6 +14,10 @@ on:
       - "docs/**"
       - ".readthedocs.yaml"
       - ".github/workflows/docs.yml"
+  # Run on merge_group so this context still reports if it is ever added to the
+  # branch ruleset as required — a required check that never runs on merge_group
+  # hangs the first merge-queue entry forever (#1769).
+  merge_group:
 
 permissions:
   contents: read
@@ -23,6 +28,10 @@ concurrency:
 
 jobs:
   build:
+    # Names the status-check context `Docs` rather than the default `build`, which
+    # would collide with ci.yml's already-required `Build` in a ruleset that matches
+    # context strings exactly (differing only in case).
+    name: Docs
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,50 @@
+name: Docs
+
+# Builds the mdBook user docs as a PR gate. Read the Docs does the actual hosting/deploy
+# via .readthedocs.yaml; this job only validates that the book builds and has no dead
+# internal links (mdbook-linkcheck2). Tool versions are pinned in lockstep with
+# .readthedocs.yaml.
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    paths:
+      - "docs/**"
+      - ".readthedocs.yaml"
+      - ".github/workflows/docs.yml"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: docs-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+
+      - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8  # stable
+        with:
+          toolchain: stable
+
+      - name: Cache mdBook tools
+        id: cache-tools
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830  # v4.3.0
+        with:
+          path: |
+            ~/.cargo/bin/mdbook
+            ~/.cargo/bin/mdbook-linkcheck2
+          key: mdbook-0.5.2-linkcheck2-0.12.2
+
+      - name: Install mdBook + linkcheck2
+        if: steps.cache-tools.outputs.cache-hit != 'true'
+        run: |
+          cargo install mdbook --version 0.5.2 --locked
+          cargo install mdbook-linkcheck2 --version 0.12.2 --locked
+
+      - name: Build the book (fails on dead internal links)
+        run: cd docs && mdbook build
--- a/.gitignore
+++ b/.gitignore
@@ -120,3 +120,6 @@ manuscript/
 
 # Local scratch working directory (ad hoc scripts, offline tooling)
 /scratch/
+
+# mdBook build output
+docs/book/
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -2,13 +2,34 @@ version: 2
 
 build:
   os: ubuntu-24.04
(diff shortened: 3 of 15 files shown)


nh13:
@coderabbitai pause

nh13:
Read-only review of `4342ef09`, done in a throwaway worktree — nothing in this branch or its worktree was touched. Docs+CI only, trailer honest and true, CI green. One blocking item, four false prose claims, two CI-config items worth doing before `Docs` becomes required.

## Blocker: `create-missing` defaults to `true`, so a SUMMARY typo ships a blank page

`docs/book.toml`'s `[build]` sets only `build-dir`, so `create-missing` is `true`. Measured with the pinned tools (mdBook 0.5.2, mdbook-linkcheck2 0.12.2), running `cd docs && mdbook build`:

| mutation | exit |
|---|---:|
| unmodified (control) | **0** — no false red |
| dead relative link in a page | **101** — caught, with file:line:col |
| **SUMMARY entry for a nonexistent page** | **0** — *mdBook created the file* (13 bytes, `# Ghost page`) |
| same + `create-missing = false` | **101** |
| dead fragment anchor (`#no-such-heading`) | 0 — not checked, even with `warning-policy = "error"` |
| dead external link | 0 — expected, `follow-web-links = false` |
| orphan page not in SUMMARY | 0 — not checked (also not published) |

All mutations restored byte-for-byte, md5-verified.

This matters more than its size because #2044 is already open and adds a page via a SUMMARY edit. Misspell the path or forget to `git add`, and CI is green, RTD is green, and the site gets a nav chapter whose body is its own title. This PR exists to host a section that lands one page at a time through exactly that edit.

Fix: `create-missing = false` under `[build]` — verified 0 → 101.

Related, non-blocking: fragments are not validated, so "dead internal links fail" is broader than what the gate covers. Worth saying what it checks.

## Four prose claims that are false against the code

- **`benchmarking.md:50`** — "Supported tools: `ferro-hgvs`, `mutalyzer`, `biocommons/hgvs`, `hgvs-rs`". `src/benchmark/cli.rs:15-26` is a bare `ValueEnum` with no `rename_all`, so clap accepts exactly `ferro | mutalyzer | biocommons | hgvs-rs | all`. `ferro-hgvs` and `biocommons/hgvs` are rejected values. (The three invocation forms on that page are all correct.)
- **`quickstart.md:41`** — `Normalizer(reference_json="ferro-reference/cdot.json")` cannot work three ways: `ferro prepare` writes no `cdot.json` (the prepared reference has `manifest.json` plus a `cdot/` *directory*); `reference_json` is ferro's own JsonProvider format, not cdot's (`src/python.rs:1967`); and the entry point for a prepared reference is `Normalizer.from_manifest(...)` (`src/python.rs:2021`). This is the first Python example a reader meets, sitting under a `ferro prepare` block.
- **`quickstart.md:48-55` and `normalize-variants.md:66`** — tells readers to run `ferro explain <CODE>` for `MEMBERS_COALESCED_FROM_REPORTED_FORM`, `INSERTED_SEQUENCE_EXPANDED`, `REFSEQ_MISMATCH`. `get_code_info` is a plain uppercased map lookup and the registry has **0** occurrences of all three (positive control: `W3003` → 4 hits). The `warning[CODE]:` output-format claim is correct; `explain` simply only covers the `W`/`E` namespace.
- **`cli/overview.md:44-46`** — "Most commands accept `-f/--format <text|json|tsv>`": `tsv` is accepted only by `normalize`; `parse`/`project` are `text|json`; `explain` is `text|json|markdown`.

## Two CI-config items, worth doing before `Docs` becomes a required context

- **No `merge_group:` trigger** in `docs.yml`. `ci.yml` and `lint-workflows.yml` both have one; #1769 added them because a required context that does not run on `merge_group` makes the first queue entry hang forever. Harmless while `Docs` is advisory — three lines now versus a deadlock later.
- **The job has no `name:`, so its check context is literally `build`** (lowercase), alongside `ci.yml`'s already-required `Build`. Both appear in `gh pr checks`. Two required-context candidates differing only in case, in a repo whose ruleset matches exact strings. Suggest `name: Docs`.

Also, the body calls this "a PR gate", but `build`/`Docs` is not among the ruleset's ten required contexts, so a red docs build is currently advisory — worth the wording matching. And note that making `Docs` required **with the current `paths:` filter** would strand every non-docs PR on "Expected — waiting for status to be reported".

## Two judgement calls

- **The `shadow-spec` URL is about to become permanent.** The codename split is applied consistently in prose — every user-visible string says "Interpretation", and `shadow-spec` survives only in paths and `sidebar-brand.js:148`'s label map. The one leak is the published URL, `/shadow-spec/index.html`. Free to rename now; costs redirects once RTD publishes.
- **The RTD half is unvalidated by any PR check.** The YAML looks schema-correct (`version: 2`, `build.os/tools/commands`, `rust: "latest"`, output to `_readthedocs/html`, and `cd docs && mdbook build` correctly confined to one `&&` since RTD runs each list entry from the checkout root). What is unverified is build *time*: it compiles mdbook + linkcheck2 from source every build with no cargo cache. The branch history shows `7989890f` (prebuilt binaries) then `59b50674` (revert to cargo-install), so this loop has been walked once already; first real evidence arrives post-merge.

## Nitpicks

- `custom.css:7` `@import`s from fonts.googleapis.com — a third-party request per page view. Position is valid so it does work; flagging for privacy/offline only.
- Cache key `mdbook-0.5.2-linkcheck2-0.12.2` has no `${{ runner.os }}` prefix and no `restore-keys`. Harmless with one runner, and the version-in-key shape is correct here.
- `shadow-spec/index.md:52-55` asserts in the present tense that a build check executes every example; that check lands in #2044. Vacuous rather than false today ("None published yet"), but it is a guarantee shipping slightly ahead of its enforcement.
- `docs/` already holds 11 top-level `.md` files including `NORMALIZATION_CONTRACT.md`, which is blessed on many PRs — those now trigger a Docs build via `paths: docs/**`. 9s warm; noted, not a problem.

## Verified good — no need to re-check

Every page in `SUMMARY.md` exists and every page under `docs/src/` is reachable from it; 9 files, no orphans. Both pinned tool versions are real and ran locally at exactly those versions. `.gitignore` adds `docs/book/`. **The trailer is correct and true** — `Representation-Change: none.` at column 0, exactly once, unwrapped, and the 15-file diff touches no watched prefix. All **17** CLI subcommand names match the `Commands` enum one-for-one, none missing or invented. Flags verified against source for `prepare`, `check`, `normalize`, `project`, `explain`. The tsv header matches `src/bin/ferro.rs:279` exactly. Installation page checks out: crate 0.14.0 → `"0.14"`; `[[bin]] ferro` has no `required-features` so `cargo install ferro-hgvs` really does yield `ferro`; the wheel matrix matches `release-wheels.yml`; `requires-python = ">=3.10"`. `edit-url-template` renders correctly.

`workflow_pipefail_gate` is **not** a hazard here: the branch is fully current with `main` (`rev-list --count origin/pr-2043-h..origin/main` = 0), already carries the guard, and it ran green in all six `Test (n/6)` shards on this exact head with `docs.yml` present. `docs.yml` contains no shell pipeline at all — its only `|` characters are YAML block scalars.

## Stack note

`git merge-base --is-ancestor origin/pr-2043-h origin/pr-2044-h` → **NO**. #2044 is missing 8 commits of this branch, including `4342ef09 docs: name the docs section "Interpretation"`, so it predates the rename and its pages likely still carry the old naming. It will need `git rebase --onto` after this lands, not just a retarget.

nh13:
Review addressed on `18e24e47` (fast-forward on `feat/docs-site`):

- **Blocker — `create-missing`:** set `create-missing = false` in `book.toml`. Mutation-verified: a SUMMARY entry for a nonexistent page now exits **101** (was 0; mdBook no longer invents a `# Ghost page` stub). This is exactly the failure mode for a section that adds pages one SUMMARY edit at a time.
- **quickstart.md Python:** `Normalizer.from_manifest("ferro-reference/manifest.json")` (the prior `Normalizer(reference_json="…/cdot.json")` was wrong three ways).
- **quickstart.md + normalize-variants.md:** dropped the `ferro explain <CODE>` advice for the normalizer's SCREAMING_SNAKE codes — `explain` covers the `W`/`E` namespace only, and the `warning[CODE]: message` line already states the meaning. The `explain`-coverage gap itself is #2092.
- **benchmarking.md:** tool values corrected to `ferro | mutalyzer | biocommons | hgvs-rs | all`.
- **cli/overview.md:** `--format` is per-command — `text|json` (parse/project), `text|json|tsv` (normalize), `text|json|markdown` (explain).
- **docs.yml:** job named `Docs` (was the default `build`, colliding with the required `Build`); added a `merge_group:` trigger; header wording corrected (advisory; linkcheck2 validates link targets, not fragment anchors).

Deferred by design: the `/shadow-spec/` URL slug stays (internal codename); RTD's cargo-install build-time is known and revisited post-merge; #2092 is a standalone `ferro explain` defect, not this PR.

PR #2047, opened 2026-08-16
test: add ClinVar round-trip coverage over the 4.2M unique corpus

### LINKED ISSUE #1859: test: no ClinVar round-trip coverage — the test that claimed it read a corpus that never existed

nh13:
Closes #1859.

Representation-Change: none. Test-only; no watched `src/` file is touched.

## The gap

`idempotency_tests::test_clinvar_sample_parsing_idempotency` was deleted as a permanent no-op: it read `clinvar_1000.txt`, a path in no commit, no release asset, and no script — so it only ever took its `Err(_) => return` branch. Its intent was real and covered nowhere:

- `clinvar_hgvs_{500k,unique}_benchmark` read the real corpora but only **parse** — they never `format!` a parsed variant, so a formatter that emits something the parser rejects (or accepts as a different variant) is invisible to them.
- `idempotency_tests::test_parsing_idempotency_*` do the round trip, but over the curated spec/edge-case corpora, not the long tail ClinVar actually stores.

So the round trip was checked on a curated corpus and the real corpus checked without the round trip. Confirmed still missing on `origin/main` @ `5567412f`.

## The change

`test_clinvar_hgvs_unique_roundtrip` over `clinvar_hgvs_unique.json.gz` (a `test-fixtures-v1` release asset fetched by `scripts/fetch-test-fixtures.sh`, verified against `CHECKSUMS.sha256`; no fixture data is committed). For every input that parses: `format!`, re-parse, require the two renderings to agree.

The residue is tracked **per-input** with the same `failure_expectations` snapshot framework the parse-only benchmarks use (`tests/common/failure_expectations.rs`), not a percentage floor. A new disagreement is a loud regression; an input that starts round-tripping is flagged as an improvement to re-bless.

## Measured before choosing the corpus

Round-trip disagreements, measured over each corpus:

| corpus | parsed | disagreements |
|---|---:|---:|
| `clinvar_hgvs_500k.json.gz` (stratified sample) | 499,999 | **0** |
| `clinvar_hgvs_unique.json.gz` (4.2M) | 4,188,224 | **1** |

The 500K stratified sample round-trips cleanly; the broadest fixture surfaces exactly one disagreement, so the test targets the unique corpus — the long tail the sample drops. The single residue:

```
input:     LRG_199:g.702849_702850insinsY13186.2:g.2333_2491
rendered:  LRG_199:g.702849_702850ins[insY13186.2:g.2333_2491]
re-parse:  Parse error at position 8: ... input: "ins[insY13186.2:g.2333_2491]", code: Tag
```

A malformed ClinVar string with a **doubled `ins` token**. The lenient parser accepts it and renders it re-wrapped as `ins[ins...]`, which no longer parses. Root cause is the input, so it is pinned as `malformed_input` (not `parser_bug`, which fails the build) in `clinvar_hgvs_unique_roundtrip_expectations.json`.

## On the new committed file

`clinvar_hgvs_unique_roundtrip_expectations.json` is a **blessed expectation baseline**, not programmatically-generatable input data — the same family as the two committed `clinvar_hgvs_*_failure_expectations.json` siblings beside it (and the `.count` baselines) that `.coderabbit.yaml`'s `tests/**` instruction carves out. It records the human-categorized residue of a real corpus and cannot be synthesized. The `.gz` corpus itself is a release asset and is **not** committed.

## Unchanged on purpose

- This is a round-trip **stability** guard (parser↔formatter asymmetry), not a conformance check — it makes no claim ferro's output is spec-correct.
- A non-vacuity floor (`>= 4_000_000` inputs) asserts the guard actually exercised the corpus, so a future breakage that made most inputs unparseable cannot masquerade as a clean pass.

## Verification

- The guard catches disagreements: with an empty snapshot it fails, reporting the one input as a regression (proved before blessing).
- Absence skips green locally with a visible `Skipping:` line; under `FERRO_REQUIRE_BULK_FIXTURES=1` an absent fixture fails with the fetch remedy.
- Runs on PR CI (fixture-fetching jobs set `FERRO_REQUIRE_BULK_FIXTURES`); it is **not** manifest-gated, so it is real pre-merge coverage.
- `cargo fmt --check` clean; `cargo clippy --all-features --all-targets -- -D warnings` clean.
- `FERRO_REQUIRE_BULK_FIXTURES=1 cargo nextest run --features dev --no-fail-fast`: **11195 passed, 0 failed, 0 `Skipping:` lines**; `test_clinvar_hgvs_unique_roundtrip` PASS in 42.3s.
- `ferro-hgvs-soak-tests` compiles (the module is included there).

### DIFF
--- a/tests/fixtures/bulk/clinvar_hgvs_unique_roundtrip_expectations.json
+++ b/tests/fixtures/bulk/clinvar_hgvs_unique_roundtrip_expectations.json
@@ -1,0 +1,12 @@
+{
+  "kinds": {
+    "doubled-ins-token-renders-non-round-trippable": {
+      "category": "malformed_input",
+      "reason": "Malformed ClinVar source string carrying a doubled 'ins' token: '...702849_702850insinsY13186.2:g.2333_2491'. The lenient parser accepts it, treating 'insY13186.2:g.2333_2491' as the inserted content, but the rendering re-wraps that as 'ins[insY13186.2:g.2333_2491]', which the parser then rejects (an 'ins' cannot open the bracketed insert). The root is the malformed input, not the round trip; it is pinned here so a NEW round-trip disagreement is a loud regression rather than lost among unpinned residue.",
+      "tracking": "#1859"
+    }
+  },
+  "expected_failures": {
+    "LRG_199:g.702849_702850insinsY13186.2:g.2333_2491": "doubled-ins-token-renders-non-round-trippable"
+  }
+}
--- a/tests/it/clinvar_hgvs_tests.rs
+++ b/tests/it/clinvar_hgvs_tests.rs
@@ -20,6 +20,18 @@ const CLINVAR_500K_FAILURE_EXPECTATIONS_PATH: &str =
     "tests/fixtures/bulk/clinvar_hgvs_500k_failure_expectations.json";
 const CLINVAR_UNIQUE_FAILURE_EXPECTATIONS_PATH: &str =
     "tests/fixtures/bulk/clinvar_hgvs_unique_failure_expectations.json";
+const CLINVAR_UNIQUE_ROUNDTRIP_EXPECTATIONS_PATH: &str =
+    "tests/fixtures/bulk/clinvar_hgvs_unique_roundtrip_expectations.json";
+
+/// A generous floor on how many inputs the round-trip guard must actually
+/// exercise. It is a **non-vacuity guard**, not the property under test: the
+/// property is that `parse → format → parse → format` agrees, and the snapshot
+/// below is what pins it. This number only ensures a future breakage that made
+/// most of the corpus unparseable (so nothing reached the round trip) cannot
+/// masquerade as a clean pass. The fixture currently offers 4,188,224 parseable
+/// inputs; the floor sits well below that so ordinary corpus/parser drift does
+/// not trip it, while a collapse to a handful would.
+const CLINVAR_UNIQUE_ROUNDTRIP_MIN_INPUTS: usize = 4_000_000;
 
 // Slim deserialization shape: borrowed `&'a str` against the
 // decompressed JSON buffer. See cmrg_exhaustive_tests for rationale.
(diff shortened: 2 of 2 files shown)


nh13:
@coderabbitai pause

PR #2044, opened 2026-08-16
docs(interpretation): DNA substitution page with executable checker

nh13:
Stacked on #2043 (base `feat/docs-site`). Adds the first Interpretation page and the checker that keeps it honest.

## What this adds

- **`docs/src/shadow-spec/recommendations/DNA/substitution.md`** — a per-line reading of `DNA/substitution.md`: the spec quote, ferro's reading, example spellings, the form ferro normalizes each to, confluence classes, and links to the governing rulings.
- **Output-centric verdicts.** The recommendations are HGVS's, not ferro's — so a verdict describes ferro's *output*: `recommended` (ferro's output is the recommended form, whether the input already was it or ferro normalized there), `conformant` (valid HGVS but not yet the recommended form — a tracked limitation), `refused` (an invalid input ferro correctly rejects), `bug` (invalid output — none on this page).
- **`tests/it/shadow_spec.rs`** — the checker. Per anchor it verifies the spec quote against the pinned checkout, executes every example (parse + strict-refusal), and asserts every `Normalizes to`:
  - against a committed real-GRCh38 slice (`tests/fixtures/shadow_spec/transcripts.json`, regenerated by `build_fixture.py`) on **every PR** — hermetic, no manifest;
  - cross-checked against the full live reference when `FERRO_MANIFEST` is set (nightly).
- **`tests/fixtures/shadow_spec/corpus.jsonl`** — the single example corpus, generated from the pages and blessed with `BLESS_SHADOW_CORPUS=1`; the checker fails if it drifts.
- A verbose spec-authority comment block in `src/normalize/mod.rs` is migrated to a terse pointer into the page (−27 lines), reasoning single-sourced to the page + ledger.
- One ferro↔spec limitation is pinned as a `conformant` row (mosaic reference-first ordering, #2034).

(The `shadow-spec/` page path and `shadow_spec` test names are the section's internal codename; the user-facing name is "Interpretation".)

## Verification

`cargo nextest run --features dev -E 'test(shadow_spec)'` — 7 anchors, 20 examples executed, 15 normalizations asserted; green with and without the manifest. Sabotage-tested: a wrong verdict, an altered quote, or a changed normalization each turns the build red. The mdBook site builds clean (mdbook-linkcheck2).

## Follow-up

The page's reasoning currently *links* to the ruling ledger. A stacked follow-up will add a short human-readable `summary` field to each ledger record and **transclude** it inline (single-sourced, generated), so the end-user reads the "why" on the page without a second copy of the adjudication.

[...]

PR #2048, opened 2026-08-16
docs(interpretation): transclude the "why" from the ruling ledger

nh13:
Stacked on #2044 (base `feat/shadow-spec-substitution`). Gives end-users the *reasoning* inline on each interpretation page, single-sourced from the ruling ledger.

## What "transclude" means here

The adjudication's reasoning lives once — in the ledger. Instead of re-typing it on the page (a second copy that drifts), the checker **generates** each anchor's "Why" block from the ledger record's own one-line summary at bless time. What the reader sees on the page is the ledger's text, pulled in mechanically.

## What this adds

- **A `summary` field on the ruling records.** One human-readable sentence per record, for end-user rendering; the dense full `rationale` stays for agents. Added to the five records the substitution page cites. Only `generate_spec_fixture`'s strict struct needed the new field — the shared `common::rulings::Record` reader ignores unknown keys, so the `NORMALIZATION_CONTRACT.md` render is untouched (verified: all ledger/contract-doc guards pass, generator regenerates clean).
- **Transclusion in the checker.** Each page anchor carries a `<!-- why:START -->…<!-- why:END:id,id -->` block. `BLESS_SHADOW_CORPUS=1` fills the body from the ledger summaries (each linked to the full record); a normal run verifies it. A cited ruling with no `summary` fails the build.

## Verification

- Green with and without the manifest; book builds clean (linkcheck2).
- **Single-source proven by sabotage:** editing a *ledger* summary makes the *page's* Why block go stale (RED) until re-blessed.
- Ledger integrity: `ruling_records_are_intact`, `normalization_contract_doc`, clause-index guards all pass; `generate_spec_fixture` regenerates without change.

Representation-Change: none. Docs, a test-only generator struct field, and a new optional data field on the ledger; no ferro output moves.


nh13:
@coderabbitai pause

PR #2051, opened 2026-08-16
feat(normalize): gate the render-time reference against what the pipeline got

### LINKED ISSUE #1946: normalize: presentational rewrites are minted before the stages that must compute on them

nh13:
Refs #1946. Step 2 of increment 4, and it is a **measurement**, not a move: nothing is relocated, the corpus dump is byte-identical, and `mint_reference_for` is unused by `render_canonical_spelling`.

Deliberately not `Closes #1946`. #1974 shipped increments 0–4 and declined to close for the same reason: these are the measurement instrument and the empty seam, not the change. The behavioural work #1946 describes — `collect_canonical_edits` dropping a whole cis allele on a non-literal insertion payload, `lower_repeat_edits` undoing repeat notation mid-pipe, `demote_repeats_spanning_siblings` undoing a per-member decision no sibling context reaches — is entirely unstarted.

Rebased with `git rebase --onto origin/main 454a91e4` now that #1974 has merged. That replays this PR's one real commit and none of #1974's six already-squashed ones; it applied clean.

## The question

Relocating the three repeat mints into the render stage requires that stage to rebuild, from a settled member and a provider, the `ref_seq` the per-member pipeline was handed — and `ref_seq` is **not one thing**. It is the whole spliced transcript on `c.`/`r.`/`n.`, a growing window on `g.`/`m.`, and a genomic window on the intronic and junction-crossing paths. `insertion_to_repeat` walks outward bounded only by `ref_seq.len()`, so substituting one uniform window silently truncates a tract into a smaller copy count.

Until this PR that comparison could not be made at all, because nothing recorded what the pipeline was given.

## The first revision of this gate was vacuous, and that is the interesting part

The recorder's first form was a flat `Vec<(usize, u64)>` and the gate asked whether **some** entry in it matched the rebuilt reference:

```rust
seen.iter().any(|(l, d)| ...)
```

`seen` is the union of every reference the whole normalization used — one per member of a cis allele, one per growth attempt, one per `FERRO_ASSERT_IDEMPOTENT` verification pass, and one per recursive re-entry (`normalize_na_edit` calls itself from **eight** arms, always with the same `ref_seq`). On any row carrying a transcript-axis member the whole-transcript entry is somewhere in that union, so `any` matched for the same reason on every row and the gate reported `3903 / 3903` — re-measured on this base; it shipped as `3865/3865` against an older one.

`all` in its place reads `3761 / 3903`: **142 members diverge, every one on the multi-exon `cx` axis.** But `all` is not the fix — different members of a legitimate multi-member allele genuinely see different windows, so it would go red on correct behaviour.

**The sabotage the PR shipped with proved the assertion was *wired*, not that it was *tight*.** Mutating `mint_reference_for` to return one base short changes the *built* digest, so `any` finds nothing and the gate goes red. That test passes identically against an assertion that can never fail for any other reason. It is exactly how this shipped.

## What replaced it: per-call-site attribution

`NA_EDIT_REFERENCES` now records a struct rather than a pair:

```rust
pub struct NaEditReference {
    pub call: u32,            // monotonic, in call order
    pub parent: Option<u32>,  // the call this one recursed out of
    pub depth: u32,           // 0 at an entry point
    pub site: NaEditCallSite, // WHICH entry point handed this down
    pub len: usize,
    pub digest: u64,
}
```

`NaEditSiteGuard` is entered at each of the nine normalizer entry points that reach `normalize_na_edit` (`normalize_cds`, `normalize_tx`, `normalize_rna`, `normalize_intronic_{cds,tx}`, `normalize_boundary_spanning_{cds,tx}`, `normalize_in_grown_window`, `canonicalize_without_shifting`), and `NaEditFrameGuard` is held for the whole body of `normalize_na_edit` so recursive re-entries carry their parent's id and a depth. Both are RAII, because that function returns from a dozen places and several of them through `?`.

The gate then compares **sets, per frame, at depth 0 only**:

- `MintFrame::WholeTranscript` against the calls at `Cds` / `Tx` / `Rna`, asserted **equal in both directions**, and the witnessed set asserted to be a **singleton** — which is the "there is nothing to choose when the reference is served whole" claim stated directly instead of assumed.
- `MintFrame::GenomicWindow` against the calls at `GrownWindow` / `UnshiftedWindow`, reported.

Set equality is what makes a *mis-attributed* entry fail: a junction-crossing window filed under `Cds` lands in the witnessed set and the mint has nothing to match it with.

## The 142 `cx` members: provably fine, and the real gap is now named

Every one of those 142 sits on a row that records **both** frames. `NM_TESTX.1:c.6_11inv` records

```
Cds@d0#0=20  BoundarySpanning@d0#1=206  Cds@d0#2=20  BoundarySpanning@d0#3=206
```

because `normalize_cds` tries the 20-base spliced transcript and then re-normalizes across the exon junction in a ~200-base genomic window. The mint's whole-transcript rebuild agrees with the `Cds` call **byte-for-byte**; the 206 belongs to a call site the mint does not model at all.

Measured containment: the 68 distinct inputs behind those 142 members are a **subset** (`comm -23` → 0) of the 74 inputs / 102 rows the repaired gate counts under `rows_using_an_unmodelled_frame`. So the divergence is fully explained by one gap, and that gap now has a name, a count, and a paragraph on `mint_reference_for` saying it is the next step rather than a widening to do in passing.

## What is asserted, and what is only reported

Unmodified tree:

```
MINT-REFERENCE whole-transcript sets equal on 2209 rows (0 mismatching);
genomic 1586/2344 at least as wide (0 unwitnessed);
681 members had no rebuildable reference across 625 rows;
102 rows used a frame the mint does not model, 0 of them with no modelled call at all;
0 unattributed calls; 7 rows held out
```

| assertion | why it is a property and not a number |
|---|---|
| whole-transcript sets equal, over 2209 rows | the reference is served whole on `c.`/`r.`/`n.`, so a rebuilt one has nothing to disagree about — and the pipeline cannot have used one the mint does not build |
| the witnessed whole-transcript set is a singleton | one normalization cannot be handed two different whole transcripts |
| `unattributed_calls == 0` | every recorded call is inside a known entry point, so the site selection is over a population somebody delimited |
| the recorder did not overflow its cap | a truncated set is indistinguishable from a mismatch |
| `tx_rows_compared > 0` | `0 of 0` cannot pass as a result |
| `ORACLE_TRIPPING_INPUTS` is shrink-only | an exemption for a row that stopped existing is how a held-out list rots into a blanket one |

The genomic half stays **reported**. It cannot be byte-equal by construction: the pipeline's window is centred on the *input* position, and the two do not even share a left edge — `normalize_in_grown_window` fetches from `start - w` where `mint_reference_for` fetches from `start - 1 - w`. Pinning today's ratio would make the count *be* the property.

## Two latent defects the cap assertion caught

Neither was visible in the shipped form, because an uncapped buffer cannot report truncation:

1. The gate never drained the recorder on the `normalize` **Err** path, so a refused row's references bled into the next row's comparison set.
2. `for row in dump(1)` evaluated `dump` **after** arming, so the entire corpus's own normalizations were recorded before the first comparison ran.

Both are fixed, and the fix is the assertion rather than the discipline.

## The gate is proved TIGHT, not merely wired

Each mutation applied, run, then restored — `shasum -a 256` equal to a pre-edit copy on all three files, checked. "Old gate" is the shipped `any()` over the flat union, replicated in the same binary so both are measured on one run.

| # | mutation | this gate | the shipped `any()` gate |
|---|---|---|---|
| 0 | unmodified | **GREEN** — 2209 rows compared, 0 mismatching | `3903 / 3903` — passes |
| 1 | mint returns one base short (the PR's own sabotage) | **RED** — 2209 / 2209 rows mismatch | `0 / 3903` — RED |
| 2a | mint returns the right length, wrong content (reversed) | **RED** — 2209 / 2209 rows mismatch | `0 / 3903` — RED |
| 2b | mint returns the **genomic window** — a reference the pipeline really used, at the wrong site | **RED** — 28 rows mismatch | `3851 / 3903` — RED |
| 3 | attribution mis-associated: `normalize_boundary_spanning_*`'s calls filed under `Cds` | **RED** — the singleton assertion fires: `NM_TESTX.1:c.6_11inv saw [20, 206]` | **blind** |

Row 3 is the one that separates them, and its "blind" is by construction rather than by measurement: the mutation changes only the `site` label, and the shipped gate reads `len` and `digest` only, so its input is byte-identical to row 0's. It is stated as a mechanism because the replica's counter never printed — the singleton assertion panics first. Everything else in the table is measured.

Row 2b is worth reading as a limit rather than a win: both gates catch it, and neither is a strict improvement on the other there (this one flags 28 rows, the old one 52 members). The tightening is row 3 and the vacuity at row 0.

## Seven review findings folded into the same pass

1. **`pub mod render` narrowed back to `pub(crate)`.** The module was widened to expose two test-only items, which put `render_canonical_spelling` and the module's design notes on the public documentation surface. Replaced with a `#[doc(hidden)] pub use render::{mint_reference_for, MintFrame, MintReference, MINT_WINDOW}` — exactly the four items the example needs, on the `ShuffleDirection` precedent in `src/lib.rs`.
2. **`MintReference::bases` said "upper-cased" and was not.** The **doc** was wrong, and fixing the code would have been worse: byte-equality with the pipeline is the property, and neither side case-folds a transcript (`normalize_cds` passes `transcript.sequence.as_bytes()`), while `normalize_in_grown_window` does not case-fold its genomic window either although `fetch_canonical_window` does. The field now states each path's own convention and why it must not impose one.
3. **`complete` was set from a 3'-clamped fetch and documented as "the complete reference".** Replaced by `at_sequence_start` / `at_sequence_end` with `is_whole_sequence()` over both, so the flag no longer claims more than it checks.
4. **The frame split keyed on `built.complete && built.offset == 0`.** That is safe only because `PAD_OFFSET` (256) exceeds `MINT_WINDOW` (100) — an accident. It now keys on a `MintFrame` derived from `CisKind`, which is the thing it means.
5. **A comment citing `2274` where everything else said `1586`.** Both figures belong to an earlier revision's base; the surrounding text now says so and stops mixing them with this one's.
6. **Rustdoc spliced mid-bullet-list** on `mint_reference_for` — the `# Measured` heading sat between the `c.` bullet and the `g.` bullet. Repaired, and the measured figures moved to the gate, so one measurement has one home.
7. **`NA_EDIT_REFERENCES` was an unbounded `Vec` that every debug normalization appended to.** Recording is now **off** unless a test holds a `NaEditReferenceRecording` guard, the buffer is capped at 4096 entries, and an overrun is reported by `na_edit_references_overflowed()` and asserted rather than silently truncating. Measured: the deepest single row records **21** entries.

## Verification

Run on the rebased tip, over `origin/main` at `9b6dcf87`, with **zero `Skipping:` lines** in the log — so the ClinVar, CMRG and Paraphase suites were armed rather than skipping green (`test_cmrg_exhaustive_benchmark` alone takes 99.6s).

| gate | `EXIT=` read back out of its own log |
|---|---|
| `cargo nextest run --features dev --no-fail-fast` | `EXIT=0` — **11279 run, 11279 passed**, 152 skipped |
| `cargo clippy --all-features --all-targets -- -D warnings` | `EXIT=0` |
| `cargo fmt --all --check` | `EXIT=0` |
| the gate test with `FERRO_ASSERT_IDEMPOTENT` / `_REPARSE` / `_IN_BOUNDS` armed | `EXIT=0`, every **asserted** figure unchanged |

**Flag stability, honestly.** Every asserted figure is identical armed and unarmed. One *reported* counter is not: `rows_using_an_unmodelled_frame` reads 102 unarmed and 103 armed, because the idempotency oracle's verification pass is a second normalization whose depth-0 calls are recorded too, and on one row it reaches a junction-crossing site the first pass did not. A counter over a union is exactly what stays flag-sensitive — which is the argument for asserting over per-site sets, restated as a measurement.

**What was NOT re-measured.** The `Representation-Change: none` claim below rests on a diff-scope argument rather than on a fresh two-revision corpus dump: `src/normalize/mod.rs` and `src/normalize/render.rs` have **zero removed lines** against `origin/main`, every `NA_EDIT_*` symbol is referenced only inside the recorder block (`mod.rs:2215-2391`), the nine added statements inside normalizer bodies are guard bindings that write a thread-local and nothing reads, and the example's whole diff is one hunk **inside `mod tests`** — `main`, `dump` and every family builder are untouched.

Representation-Change: none. 0 corpus rows move. `mint_reference_for` is unused by `render_canonical_spelling`; the recorder is `#[cfg(debug_assertions)]`, disarmed by default and write-only; and the nine `NaEditSiteGuard`s only set a thread-local `Cell` that nothing in the normalizer reads.

### DIFF
--- a/examples/dump_normalized_corpus.rs
+++ b/examples/dump_normalized_corpus.rs
@@ -3753,6 +3753,391 @@ mod tests {
         }
     }
 
+    /// Rows this gate must not normalize, because normalizing them aborts the
+    /// process under the `Test oracle` job's own flags.
+    ///
+    /// **These are normalizer defects that this corpus revealed, not flakes and not
+    /// rows the gate is allowed to be indifferent about.** Each is filed, each is
+    /// labelled `high`/`bug`, and each names the oracle it trips:
+    ///
+    /// - **#2036** — `FERRO_ASSERT_IDEMPOTENT`. Normalizing a long tract is not a
+    ///   fixed point: every pass walks three more bases 5' and increments the copy
+    ///   count (`g.558_559insACGACG` -> `g.460_558ACG[35]` -> `g.457_558ACG[36]`).
+    ///   Found by `long_tract_window_provenance` on its first armed run, which is
+    ///   the window-provenance defect that family was built to find.
+    /// - **#2037** — `FERRO_ASSERT_IN_BOUNDS`. Combining two members at `c.1` shifts
+    ///   the insertion point 5' off the coordinate space
+    ///   (`c.[1dup;1T>A]` -> `c.-1_1insA`, naming position 0 of a 20-base transcript).
+    ///
+    /// **Why an exclusion here rather than a weakened oracle.** The oracles panic
+    /// *inside* `normalize`, so a row that trips one takes the whole test binary with
+    /// it — and this gate re-normalizes each row **itself**, outside the `catch_unwind`
+    /// that [`normalize_through`] wraps `dump`'s own pass in. So the choice is not
+    /// between checking and not checking these rows; it is between the gate answering
+    /// its own question and the gate aborting on someone else's defect. Suppressing the
+    /// oracle, or dropping the rows from the corpus, would instead destroy the evidence
+    /// the two issues rest on.
+    ///
+    /// The list is **shrink-only**: every entry must still be produced by the corpus,
+    /// asserted below, so a row that stops being generated — or a defect that gets
+    /// fixed and re-admits its row — fails here instead of rotting into a blanket
+    /// exemption.
+    const ORACLE_TRIPPING_INPUTS: &[(&str, &str)] = &[
+        ("NC_TEST.1:g.456_457insACGACG", "#2036"),
+        ("NC_TEST.1:g.558_559insACGACG", "#2036"),
+        ("NC_TEST.1:g.582_583insACGACG", "#2036"),
+        ("NC_TEST.1:g.585_586insACGACG", "#2036"),
+        ("NC_TEST.1:g.756_757insACGACG", "#2036"),
+        ("NM_TEST.1:c.[1dup;1T>A]", "#2037"),
+        ("NM_TEST.1:c.[1dup;1C>A]", "#2037"),
+    ];
+
+    /// The reference a render-time mint would rebuild, against the one the
+    /// per-member pipeline was actually handed (#1946).
+    ///
+    /// **This is the gate on relocating the repeat mint.** `render::mint_reference_for`
+    /// has to reconstruct, from a settled member and a provider, what
+    /// `normalize_na_edit` was given — and `ref_seq` is not one thing: the whole
+    /// spliced transcript on `c.`/`r.`/`n.`, a window on `g.`/`m.`, and a genomic
+    /// window on the intronic and junction-crossing paths.
+    ///
+    /// # The comparison is per call site, and the first revision of it was vacuous
+    ///
+    /// The recorder hands back **every** `ref_seq` a normalization used: one per
+    /// member of a cis allele, one per growth attempt, one per
+    /// `FERRO_ASSERT_IDEMPOTENT` verification pass, and one per recursive re-entry
+    /// (`normalize_na_edit` calls itself from eight arms, always with the same
+    /// `ref_seq`). This gate's first revision asked whether **some** entry in that
+    /// union matched the rebuilt reference — `seen.iter().any(..)` — and reported
+    /// `3903 / 3903` (re-measured on this base; it shipped as `3865/3865` against an
+    /// older one). That number could not have read much else: on any row with a
+    /// transcript-axis member the whole-transcript entry is somewhere in the union,
+    /// so `any` matched for the same reason on every row. Turning it into `all` read
+    /// `3761 / 3903` — **142 members diverge, every one on the multi-exon `cx`
+    /// axis** — and that is not the tightening it looks like either: different
+    /// members of a legitimate multi-member allele genuinely see different windows,
+    /// so `all` over the union goes red on correct behaviour.
+    ///
+    /// Those 142 are the whole story, and per-site selection resolves them rather
+    /// than hiding them. Every one sits on a row that records **both** a `Cds`
+    /// call handed the 20-base spliced transcript **and** a `BoundarySpanning` call
+    /// handed a ~200-base genomic window — `NM_TESTX.1:c.6_11inv` records
+    /// `[Cds@d0=20, BoundarySpanning@d0=206, Cds@d0=20, BoundarySpanning@d0=206]`,
+    /// because `normalize_cds` tries the spliced frame and then re-normalizes across
+    /// the exon junction in genomic coordinates. The mint's whole-transcript rebuild
+    /// agrees with the `Cds` call byte-for-byte; the 206 belongs to a site the mint
+    /// does not model at all, and is counted below rather than matched.
+    ///
+    /// Neither is the right question. The right question is per **call site**, which
+    /// is why [`NaEditReference`] carries one: for each of the two shapes
+    /// `mint_reference_for` claims to model, compare the set of references **it**
+    /// builds against the set the pipeline was handed **at the sites that shape
+    /// models**, at depth 0. Set equality in both directions, so an extra reference
+    /// on either side fails — which is what makes a mis-attributed entry (a
+    /// junction-crossing window filed under `Cds`) fail rather than pass.
+    ///
+    /// # What is asserted
+    ///
+    /// 1. **Every recorded call is attributed.** A depth-0 call at
+    ///    [`NaEditCallSite::Unattributed`] means a normalizer entry point reaches
+    ///    `normalize_na_edit` without a site guard, and every conclusion below is
+    ///    then about a subset nobody delimited. Asserted zero.
+    /// 2. **The recorder did not truncate.** A capped buffer that silently drops
+    ///    entries looks exactly like a set that never contained them.
+    /// 3. **The whole-transcript sets are equal.** The reference is served entire on
+    ///    `c.`/`r.`/`n.`, so there is no window to choose and no reason for a rebuilt
+    ///    reference to differ by a byte from the one the pipeline indexed — nor for
+    ///    the pipeline to have used one the mint does not build. That is a property,
+    ///    statable without any measured number. Asserted over the rows where it is
+    ///    answerable, with the answerable-row count asserted non-zero so `0 of 0`
+    ///    cannot pass as a result.
+    /// 4. **One whole transcript per row.** The set is asserted to be a *singleton*
+    ///    wherever it is non-empty, which is the "nothing to choose" claim stated
+    ///    directly rather than assumed.
+    ///
+    /// # What is reported, and why it is not asserted
+    ///
+    /// **The genomic frame.** It cannot be byte-equal by construction: the pipeline's
+    /// window is centred on the *input* position and a render stage only has the
+    /// settled one, and the two do not even share a left edge —
+    /// `normalize_in_grown_window` fetches from `start - w` where `mint_reference_for`
+    /// fetches from `start - 1 - w`. So the question there is containment, and
+    /// pinning today's ratio would make the count *be* the property.
+    ///
+    /// **The frames the mint does not model.** A `c.` member whose span crosses an
+    /// exon/exon junction, or which sits in an intron, is normalized against a
+    /// *genomic* window by `normalize_boundary_spanning_cds` / `normalize_intronic_cds`
+    /// — `mint_reference_for` returns the spliced transcript for it, because that is
+    /// what its `CisKind` says. Those rows are counted under
+    /// `rows_using_an_unmodelled_frame` rather than being folded into a match, and a
+    /// row whose transcript-side calls were **only** at such a site is excluded from
+    /// assertion 3 (there is no whole-transcript call to compare against) and counted
+    /// under `rows_with_no_modelled_call`. That is the next piece of work, and naming
+    /// it as a population is the difference between an open gap and a hidden one.
+    ///
+    /// # Measured over `dump(1)`
+    ///
+    /// Read the run's own `MINT-REFERENCE` line rather than this table; it is here so
+    /// a reader knows the order of magnitude and which buckets exist.
+    ///
+    /// | population | result |
+    /// |---|---|
+    /// | rows where the whole-transcript sets are compared | see the printed line — **asserted equal** |
+    /// | genomic-frame members at least as wide as the widest witnessed window | reported |
+    /// | members with no rebuildable reference | reported |
+    /// | rows using a frame `mint_reference_for` does not model | reported |
+    /// | held out (#2036, #2037) | 7 rows |
+    ///
+    /// **Every asserted figure is identical with the three `FERRO_ASSERT_*` flags set
+    /// and unset, measured both ways** — same 2209 rows compared, same zero
+    /// mismatches, same `1586 / 2344`, same zero unattributed calls. One *reported*
+    /// counter does move: `rows_using_an_unmodelled_frame` reads **102** unarmed and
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #2052, opened 2026-08-16
docs: correct measurement-doctrine advice on pub(crate) reachability

### LINKED ISSUE #1961: docs: pub(crate) is unreachable from an [[example]] — measurement-doctrine advice needs correcting
## Summary

The measurement-doctrine section added by #1927 advises, of a constant an example needs to read:

> if it is private, make it `pub(crate)`

That does not work for the case the section is about. `dump_normalized_corpus` is a genuine `[[example]]` in `Cargo.toml`, so it links the library as an **external crate** (`use ferro_hgvs::…`). `pub(crate)` is not reachable from an external crate, so following the advice leaves the example unable to see the constant.

## Suggested correction

The reachable options are `pub` plus `#[doc(hidden)]`, or a `dev`-feature-gated re-export. Either keeps the item out of the public documentation surface while letting the example read it.

## Provenance

Found while reviewing #1927 and verified against `origin/main` at `cb79979a`: `SPLIT_CAP: u64 = 1024` (`examples/dump_normalized_corpus.rs:2381`) and `MAX_SPLIT_BLOCK = MAX_CANONICAL_WINDOW = 4096` (`src/normalize/merge.rs:2000`, `:1894`). #1927 was landed as-is because the inaccuracy is confined to this one sentence and makes no other claim in the section false.

## Second, smaller item

The same section says the corpus generator has been "blind six times" while the corpus file's own header (`examples/dump_normalized_corpus.rs:105`) enumerates five and does not mention #1917. The sixth is defensible — #1923 calls the reversed-range zero "the same blindness as #1460 and #1478" — but the two counts should agree, and #1923 is currently open against that file. Worth reconciling once #1923 lands.

nh13:
Closes #1961.

Representation-Change: none

## The defect

The "Assert the property. Measure the count. Never let a count BE the property" section of `CLAUDE.md` (added by #1927) advised, of a constant an example needs to read:

> Import the constant. If it is private, make it `pub(crate)` …

That does not work for the case the section is about. `dump_normalized_corpus` is a genuine `[[example]]` in `Cargo.toml`, so it links the library as an **external crate** (`use ferro_hgvs::…`). `pub(crate)` is not reachable from an external crate, so following the advice leaves the example unable to see the constant.

## The fix

State the reachable options accurately: the item must be `pub`, optionally with `#[doc(hidden)]` or a `dev`-feature-gated re-export to keep it off the public documentation surface. This is exactly what #1925 did — it made `MAX_CANONICAL_BLOCK` `pub` (`src/normalize/merge.rs`, re-exported `pub use` from `src/normalize/mod.rs`) and imported it into the example.

### Verification of the reachability claim

- `dump_normalized_corpus` is `[[example]]` in `Cargo.toml` (`name = "dump_normalized_corpus"`, `required-features = ["dev"]`).
- The example reads the constant as an external-crate import: `use ferro_hgvs::normalize::{…, MAX_CANONICAL_BLOCK};`.
- `MAX_CANONICAL_BLOCK` is `pub const` in `src/normalize/merge.rs` and `pub use`-re-exported from `src/normalize/mod.rs` — i.e. `pub`, never `pub(crate)`. A `pub(crate)` item would not compile through that `use`.

## Second, smaller item (reconciled now that #1923 has merged)

The same section said the generator had been "blind six times", but the corpus header — rewritten by #1947 to state no count and defer to `CLAUDE.md` as the maintained list — names a further blindness: the `repeat_beside_a_sibling` family, which first shipped able to build only *growing* repeats (#1752). The maintained list is updated to seven and that instance added.

## Unchanged on purpose

- The two adjudication tables and the ruling ledger are untouched; `claude_md_adjudication_tables` parses only those tables and is unaffected.
- No `src/`, `tests/`, or example source changes — docs/comment only.

## Verification

- `cargo nextest run --features dev` — `11226 passed, 155 skipped, 0 failed` (includes `claude_md_adjudication_tables::*` and `dump_normalized_corpus::*`).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1056,16 +1056,55 @@ derived from `MAX_CANONICAL_WINDOW`), raised from `1024` by #1899. The corpus's 
 between". The cap moved, both cores landed on the same side, and `1100 > 1024` kept the guard
 green. It even carries a comment instructing the reader to update the literal if the constant
 moves; **that comment is the defect, not a mitigation.** A guard that restates the value it guards
-cannot observe that value changing. Import the constant. If it is private, make it `pub(crate)` —
-that is cheaper than the class of failure it buys out of. Filed as #1925.
+cannot observe that value changing. Import the constant. `dump_normalized_corpus` is a genuine
+`[[example]]`, so it links the library as an **external crate** (`use ferro_hgvs::…`) and can see
+only the library's `pub` API — **`pub(crate)` is unreachable from there.** A private item must be
+made `pub`, with `#[doc(hidden)]` on the re-export to keep it off the public documentation
+surface; `src/lib.rs`'s `ShuffleDirection` re-export is the worked pattern.
 
-**2. A zero whose instrument cannot vary the property.** `examples/dump_normalized_corpus.rs`
-has now been blind six times — geometry (#1456), scale (#1460), sequence structure (#1517),
-transcript geometry (#1478), molecule type (#1606), and reversed ranges (#1917) — and its own
-header records the governing fact: **fixing one blindness does not reveal the next.** Each time,
-a `0` was available to quote as safety. So before quoting one, **name the property your change
-keys on and show the generator can vary it.** A zero you cannot attribute to the change is a
+**Do not gate such a re-export behind `dev`.** The tree has already decided that, for this very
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

Issue #1989, opened 2026-08-15
equivalence: a `Declined` sequence comparison falls through with `Different`, while `WindowTooWide` reports `Indeterminate`


nh13:
**Filed as a question, not as a diagnosed defect.** The asymmetry below is verified by reading; the consequence is not, and is deliberately not claimed.

`same_resulting_sequence` can return `Declined` for several reasons that all mean *the comparison did not happen*: an unsupported edit, no reference data, mixed accessions, a multi-molecule allele (`src/equivalence/checker.rs:883-899`). `WindowTooWide` means the same thing for a different reason — the window exceeds the cap, so nothing was reconstructed.

In `check`'s ladder those two are routed **differently**:

```rust
// src/equivalence/checker.rs:535
SequenceVerdict::WindowTooWide => {
    return Ok(EquivalenceResult::new(EquivalenceLevel::Indeterminate)
        .with_note(format!(
            "The reference window covering both variants exceeds \
             {MAX_SEQUENCE_COMPARE_WINDOW} bases, so neither resulting \
             sequence was reconstructed"
        )));
}
// :544
SequenceVerdict::Different | SequenceVerdict::Declined => {}
```

`WindowTooWide` says "I could not tell" in the result. `Declined` is grouped with `Different` and falls through — so a verdict meaning *nothing was compared* takes the same branch as one meaning *they were compared and disagreed*.

## Why this looks like an inconsistency rather than a preference

The sibling site in the same file handles the two **together**, and documents at length why a decline must stop at the single-axis rung rather than becoming a stronger claim (`:657`):

> What is missing is corroboration, not a denotation, and reporting "cannot tell" would discard a measured fact.

So the codebase already holds the "a decline is not a finding" distinction in one place and collapses it in the other. `:885`'s own doc comment says `WindowTooWide` "is split out from" `Declined` — i.e. they are siblings by construction.

## What is NOT established

Whether a `Declined` can actually reach a **decided negative** (`NotEquivalent`) depends on the fall-through path below `:544`, which I did not trace. The block immediately after it is the #1578 follow-up, whose whole subject is that `NotEquivalent` is a positive claim that is only sound if some rung above actually examined the pair — and it declines for structural rearrangements on exactly that reasoning. It is entirely possible that path already catches this and there is nothing to fix.

## What would settle it

[...]

PR #2053, opened 2026-08-16
fix: report Indeterminate when an equivalence comparison cannot read its reference

nh13:
Refs #1989 — deliberately not `Closes`. This splits the **fetch** path only (`compare_triples`'s window fetch at `checker.rs:1283`). A reference-data failure inside `hgvs_to_spdi` is still swallowed by `.ok()?` at `checker.rs:1094` and lands in `Declined`, so the commonest shape of #1989 is unfixed: measured on this head, `NC_ABSENT.1:g.2del` vs `NC_ABSENT.1:g.2delC` on an unserved reference still returns a decided **`NotEquivalent`**, while the same pair on a served contig returns `CrossAxisSequenceMatch` — they are the same variant, and the only difference is whether the reference could be read. Since HGVS routinely omits the deleted or duplicated bases (`g.2del`, `c.10_12del`, `g.2dup`), that residual is the larger population. Classifying at `:1094` is the follow-up.

Representation-Change: none. Only `src/equivalence/checker.rs` changes (not a watched prefix). The equivalence verdict is a downstream *consumer* of the normalizer and feeds no normalization path, so no normalized output moves.

## The defect

The equivalence sequence rung (`same_resulting_sequence` / `compare_triples`) reconstructs each variant's edited window from SPDI triples and compares the bases. Several outcomes all meant "nothing was compared" but were collapsed into one `SequenceVerdict::Declined`, which the `check` ladder groups with `Different` and falls through to its `NotEquivalent` tail. The sibling `WindowTooWide` outcome — the reference window would exceed the compare cap — was already split out and correctly reports `Indeterminate`.

Among the cases folded into `Declined` is the one #1989 names: **the reference window could not be read from the provider** even though both sides share an accession and both convert to SPDI. That is the exact sibling of `WindowTooWide` (both mean "the window that would settle the pair could not be reconstructed"), yet it was reported as a decided negative for a comparison that never ran.

Measured wrong answer (via `compare_denotations`, which does not normalize, so the result cannot depend on normalizer confluence):

```
NC_ABSENT.1:g.2C>A  vs  NC_ABSENT.1:g.2delCinsA   (the SAME edit — identical SPDI triples)
  provider serves no bases for the accession
  before: NotEquivalent   ← decided negative asserted without ever comparing
  after:  Indeterminate
```

## The fix

Split a dedicated `SequenceVerdict::ReferenceUnavailable` out of `Declined` for the fetch-unavailable case and resolve it to `EquivalenceLevel::Indeterminate`, exactly as `WindowTooWide` is resolved. The second-axis strengthening path treats it identically to its `WindowTooWide | Declined` siblings ("the axis could not be compared").

## Unchanged on purpose (why this does not over-map)

The checker routes several other `Declined` shapes to `NotEquivalent` **deliberately**, each with a documented decision and existing tests. This change leaves all of them alone — they are representation- or geometry-level declines, not "the reference could not be read":

- an **RNA fusion** (SPDI cannot represent it) — the deliberate scope boundary pinned by `issue_1578_followup_equivalence_rungs::an_rna_fusion_pair_is_still_answered_a_deliberate_scope_boundary`;
- a genuine **cross-accession** pair — a decided negative by construction (`same_resulting_sequence`'s mixed-accession arm);
- an **overlapping cis allele** that cannot be applied — `issue_1244_equivalence_overlap_panic`;
- and a **reconstructed, genuinely different** pair — still `Different` → `NotEquivalent`.

Only the reference-level decline moves.

## Tests

New `tests/it/issue_1989_declined_is_indeterminate.rs`:
- an equivalent pair whose reference is unavailable → `Indeterminate` (the measured wrong answer, via `compare_denotations`);
- `check` reports `Indeterminate` when the reference cannot be reconstructed (the other entry point);
- **discriminating controls**: a served + genuinely different pair stays `NotEquivalent` (both entry points), and a served + equivalent pair stays a decided positive.

`equivalence::checker::tests::test_genomic_variants_different` now serves its genomic reference, so it genuinely reconstructs and compares the two different substitutions (its original intent) instead of reaching `NotEquivalent` by luck on an unserved reference.

## Verification

- `cargo fmt --all --check` → clean
- `cargo clippy --all-features --all-targets -- -D warnings` → clean
- `cargo nextest run --features dev --no-fail-fast` → `11226 passed, 155 skipped` (only this fix's target test regressed before the reference-serving update; the whole equivalence suite of 75 tests passes, preserving the fusion / mixed-accession / overlap decisions above)

### DIFF
--- a/src/equivalence/checker.rs
+++ b/src/equivalence/checker.rs
@@ -541,6 +541,42 @@ impl<P: ReferenceProvider> EquivalenceChecker<P> {
                          sequence was reconstructed"
                     )));
             }
+            // The window is bounded and both sides convert to SPDI on a shared
+            // accession, but the provider could not serve its bases, so — as with
+            // `WindowTooWide` — nothing was reconstructed and nothing was
+            // compared (#1989). This is the "want of reference data" decline, and
+            // it must report `Indeterminate` for the same reason its sibling does:
+            // falling through to the `NotEquivalent` tail would assert a decided
+            // negative about a pair that was never examined. Pinned by
+            // `issue_1989_declined_is_indeterminate`.
+            SequenceVerdict::ReferenceUnavailable => {
+                return Ok(EquivalenceResult::new(EquivalenceLevel::Indeterminate)
+                    .with_normalized(str1, str2)
+                    .with_note(
+                        "The reference window covering both variants could not be read from the \
+                         provider, so neither resulting sequence was reconstructed"
+                            .to_string(),
+                    ));
+            }
+            // `Different` is a decided negative — both sides were reconstructed
+            // and disagree. `Declined` is the catch-all decline. The shapes the
+            // checker routes to `NotEquivalent` *deliberately* are an edit SPDI
+            // cannot carry (`issue_1578_followup_equivalence_rungs` for a
+            // fusion), a cross-accession pair (the mixed-accession arm of
+            // `same_resulting_sequence`) and an un-appliable overlapping allele
+            // (`issue_1244_equivalence_overlap_panic`), each pinned there.
+            //
+            // Those are not all of it, and the arm above does not make them so.
+            // `edit_triples` discards its conversion error —
+            // `hgvs_to_spdi(member, provider).ok()?`, `checker.rs:1114` — so a
+            // member whose conversion needed the reference and could not read it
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #2054, opened 2026-08-16
docs: correct CLAUDE.md to say three cost-3 alignments, matching the ledger

### LINKED ISSUE #1971: docs: CLAUDE.md says "both cost-3 alignments" where the ledger says three
## Summary

`CLAUDE.md` and the ruling ledger disagree about the same record.

#1949 corrected `rulings[unchanged-is-read-over-every-minimal-alignment]`'s rationale to say **"the three cost-3 alignments"**. `CLAUDE.md`'s row for that same record still says **"both cost-3 alignments"**.

`GACA -> AGAT` has three cost-3 alignments, not two — verified by a brute-force enumerator during a review of #1958: the three unchanged-column sets are `{1,3}`, `{0,1}` and `{0,1}`, intersecting to `{1}`.

## Scope

This is pre-existing drift on `main`, not introduced by any open PR. It was found while rebasing #1958, whose own rewritten row also says "both" — inherited from the base rather than authored — so fixing it inside that PR would have been scope creep and was deliberately left alone.

## Why it matters here

This is the repository's stated recurring failure mode: one rule written in several places, then drifting apart. `adjudication-precedence-order` makes it part of its ruling that `CLAUDE.md` does not restate what the ledger decides. A row that paraphrases a rationale and then falls out of step with it is exactly the shape that ruling exists to prevent.

## Suggested fix

Update the `CLAUDE.md` row to match the ledger's wording, or replace the paraphrase with a pointer to the record so the two cannot diverge again.

nh13:
Closes #1971.

Representation-Change: none

## The defect

`CLAUDE.md`'s decided-rulings table row for `unchanged-is-read-over-every-minimal-alignment` described the `GACA -> AGAT` example as "**both** cost-3 alignments match ref offset 1", implying **two** minimal alignments. The ruling ledger (`tests/fixtures/grammar/hgvs_spec_normalization_overrides.json`) — the authoritative record — says **three**: "the three cost-3 alignments (matched reference sets `{1,3}`, `{0,1}`, `{0,1}`, intersecting to `{1}`)". #1949 corrected the ledger's rationale to "the three cost-3 alignments"; this `CLAUDE.md` row was left behind, so the two drifted apart — exactly the recurring failure mode `adjudication-precedence-order` exists to prevent.

## Verification of the true count

Brute-force enumeration of every minimal-cost alignment of `GACA -> AGAT` (unit-cost Levenshtein, which the "cost-3" phrasing pins), matching each alignment's set of unchanged reference offsets:

```
min cost = 3
number of minimal (cost-3) alignments = 3
matched-reference sets (multiset)      = {1,3}, {0,1}, {0,1}
intersection                           = {1}   (ref offset 1 unchanged)
```

This reproduces the ledger's `{1,3}`, `{0,1}`, `{0,1}` intersecting to `{1}` exactly. The true count is **three**, so "both" is wrong and the ledger is right. (For completeness, the substitution-as-del+ins model gives cost 4 with 9 minimal alignments and an empty intersection — not "cost-3", so it is not the model the record's wording refers to.)

## The fix

Update the `CLAUDE.md` row to mirror the ledger's wording, including the matched-set parenthetical, so the count is pinned in both places and cannot silently diverge again. Docs-only:

- The ledger JSON (a committed adjudication record) is **not** touched.
- The markdown table structure is unchanged — the row still has exactly three pipes on a single line, so `claude_md_adjudication_tables` membership parsing and the first-cell record id are unaffected.

## Representation stability

None. This is a documentation-only change; no normalized output moves.

## Verification

- `cargo fmt --check` — clean.
- `cargo nextest run --features dev` — `11227 passed, 0 failed`.
- Targeted CLAUDE.md / ledger tests — `22 passed`: `claude_md_adjudication_tables::*`, `hgvs_spec_normalization_tests::ruling_records_are_intact`, `ruling_citation_currency::*`, `ledger_prose_clause_anchors::*`.


nh13:
@coderabbitai pause

PR #2055, opened 2026-08-16
feat(ci): publish prebuilt ferro CLI binaries as release assets

### LINKED ISSUE #1529: Provide prebuilt binaries as a release artifact
## Summary

Currently, the GitHub release includes only prebuilt wheels for the Python bindings, but not a prebuilt binary for the Rust CLI. 

While binaries are available through bioconda, it'd be helpful to additionally provide binaries as release artifacts for users who don't want to use the pixi/conda ecosystem or who are doing validation/testing work during the window between GHA release and Bioconda release. 

cargo-dist seems like the popular action/framework for this: https://github.com/axodotdev/cargo-dist

nh13:
Closes #1529.

Representation-Change: none

## What

Adds `.github/workflows/release-binaries.yml`, which builds the `ferro` CLI in release mode across a cross-platform target matrix and attaches the archives (plus per-archive SHA-256 sidecars) to the GitHub Release. Previously the compiled CLI was reachable only through bioconda; the Python bindings already ship as wheels (`release-wheels.yml`) and the crate publishes to crates.io (`publish.yml`), but nothing served the CLI binary directly — which is what the issue asks for, both for users outside the pixi/conda ecosystem and for the window between the GitHub release and the downstream bioconda build.

## Target matrix

| OS | targets |
|---|---|
| Linux x86_64 | `x86_64-unknown-linux-gnu`, `x86_64-unknown-linux-musl` |
| Linux aarch64 | `aarch64-unknown-linux-gnu`, `aarch64-unknown-linux-musl` |
| macOS | `x86_64-apple-darwin`, `aarch64-apple-darwin` |
| Windows | `x86_64-pc-windows-msvc` |

Each build job packages the binary + `README.md` + `LICENSE` into a `ferro-<tag>-<target>.tar.gz` (`.zip` on Windows), computes a `.sha256` sidecar, and uploads them as an artifact. A final `upload` job downloads all artifacts and attaches them to the release with `gh release upload "$TAG_NAME" dist/* --clobber`.

## Why not cargo-dist

The issue floats cargo-dist. Releases here are already driven by release-plz (`publish.yml`) and the wheel matrix (`release-wheels.yml`), both of which own the tag format (`release-plz.toml`'s `git_tag_name`), the `release: published` trigger, and the `gh release upload` step. cargo-dist would generate its own release workflow and expect to own that flow, duplicating and fighting those conventions. A plain matrix build modelled on `release-wheels.yml` slots into the existing flow with one source of truth.

## Conventions followed

Mirrors `release-wheels.yml`: `release: published` + `workflow_dispatch` (blank-tag dry-run) triggers; the `startsWith(release.tag_name, 'v')` version gate (kept as narrow as `tests/it/release_tag_gate.rs` requires); top-level `contents: read` with `contents: write` scoped only to the `upload` job; per-job `timeout-minutes`; every action pinned by commit SHA with a `# vX` comment; `persist-credentials: false`; the release tag / input passed through `env:` so an attacker-controllable release name cannot inject shell code (zizmor template-injection). The crate is pure Rust with no C system deps (the wheel builds install no apt packages), so the only extra system dep is `musl-tools` on the two static-musl targets. The toolchain is the `rust-toolchain.toml` pin (single source of truth, installed explicitly rather than duplicating the version).

## Verification

- `actionlint` and `zizmor` (the exact tools/versions `lint-workflows.yml` runs) both clean over `.github/workflows/`.
- YAML parses.
- Local host-target smoke build confirms the build args and that no system deps are needed: `cargo build --release --bin ferro` → `Finished release`, binary reports `ferro 0.14.0`.
- No Rust changed, so fmt/clippy/nextest and the Python suite are unaffected.

The release-triggered path (matrix upload onto an actual release) cannot be fully exercised without cutting a tag; it is validated here by linting, the local build, and structural parity with `release-wheels.yml`, and will be exercised by the next tagged release.

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -1,0 +1,177 @@
+name: Build and publish CLI binaries
+
+# Prebuilt `ferro` CLI binaries as GitHub Release assets (issue #1529). The
+# Python bindings already ship as wheels (`release-wheels.yml`) and the crate is
+# published to crates.io (`publish.yml`), but the compiled CLI was only reachable
+# through bioconda. This workflow builds `ferro` for a cross-platform target
+# matrix and attaches the archives (plus per-archive SHA-256 sums) to the
+# release, for users outside the pixi/conda ecosystem and for the window between
+# the GitHub release and the downstream bioconda build.
+#
+# Why not cargo-dist (which the issue floats): releases here are driven by
+# release-plz (`publish.yml`) and the wheel matrix (`release-wheels.yml`), both
+# of which already own the tag format, the `release: published` trigger and the
+# `gh release upload` step. cargo-dist would generate its own release workflow
+# and expect to own that flow, duplicating (and fighting) the tagging and upload
+# conventions those two files pin. A plain matrix build modelled on
+# `release-wheels.yml` slots into the existing flow with one source of truth.
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Tag to upload binaries to (leave blank to build without uploading — dry-run)"
+        required: false
+        type: string
+
+permissions:
+  contents: read
+
+concurrency:
+  group: release-binaries-${{ github.event.release.tag_name || github.event.inputs.tag || github.ref }}
+  cancel-in-progress: false
+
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #2058, opened 2026-08-16
refactor: make the derived/sourced split explicit so a hollow sidecar row is unrepresentable (R4)

### LINKED ISSUE #1993: benchmark: make the derived/sourced split explicit in the type so a hollow row is unrepresentable (ruling R4)
The principled follow-up named by operator ruling **R4**, which accepted the tactical fix in #1732 (merged, closing #1728) and deferred this to a stacked change. #1791 is also merged. Filed so the ruling stops living only in a session task list.

## The shape R4 asked for

Today a sidecar row carries `sequence_length` as a **persisted** field (`src/benchmark/cache.rs:1928`), and "hollow" is a *value* of it:

```rust
pub sequence_length: usize,
...
fn is_hollow(&self) -> bool {
    self.sequence_length == 0
}
```

So a hollow row is representable, must be detected (`:1959`), and must be repaired (`:2047`, `re-derive every sidecar row's sequence_length from the FASTA, returning how many moved`). #1732 made that repair work and made a corrupt row a refusal rather than a silent reset. That is the tactical fix, and it is correct as far as it goes.

R4's principled version: **drop `sequence_length` as a persisted field and rebuild it as an in-memory index on load**, so a hollow row becomes *unrepresentable* rather than detectable-and-repairable. The FASTA is already the single authority for presence after #1791; this makes it the single authority for length too, by construction rather than by a repair pass.

The second half of the ruling: restate #1791's guard as the stronger invariant — **repairing a derived field makes no source call.** As it stands the guard is narrower than that.

## Why it is worth doing rather than leaving at the tactical fix

The repair pass and the hollow predicate exist only because the type permits a state the data never legitimately has. Every consumer that reads a row has to either trust the repair ran or re-check. `cache.rs:1943` already argues the distinction in prose:

> A hollow row is not a row with missing *information*.

If that is true, the type should say it.

## Scope and cautions

- `src/benchmark/` is **not** one of the `Representation-Change:` watched prefixes, so this needs no output disclosure — worth stating up front, because the change looks larger than its blast radius.
- It changes a **persisted artifact's** shape. Anything that alters what is written to the prepared reference risks moving the blessed reference identity (`aa8b3246d83055cc`), which the empty-projection `.count` baselines pin. Derived artifacts are content-stamped, so confirm whether the sidecar is inside the identity before assuming it is not.
- There is an outstanding operator item on the neighbouring work (task #90): the narrowed assertion in #1791's merged guard still wants confirmation. Whoever takes this should resolve that first, since the "stronger invariant" half is a restatement of exactly that guard.
- Expect this to be a multi-session change. It is not a quick win, and R4 already said it should be stacked rather than folded in.

nh13:
Closes #1993.

## Load-time banner changes on the shipped artifact

Measured against `supplemental/patterns_transcripts.metadata.json` (140,027 rows, 24 MB) on the prepared reference:

```
before   carries_cds_or_length  ->  "116572 with neither a CDS nor a length"
after    carries_a_cds_bound    ->  "140008 with no CDS bound"
```

Both are correct statements of their own predicate, and the message text moves with it. The +23,436 is the rows that carried a non-zero persisted `sequence_length` and no CDS, which the old predicate excluded and the new one cannot see — 116,572 rows have `sequence_length: 0` and 19 have a CDS.

Flagged because an operator who knows the old banner will read the new one as a large regression, and it is not one. No threshold moves: the `artifact_is_largely_without_cds` warning gates on `with_cds`, which is unchanged.

Representation-Change: none. Type/persistence refactor of the benchmark supplemental sidecar; touches `src/reference/multi_fasta.rs` (a watched prefix) but changes no normalized HGVS output. No consumer of `Normalizer`/`VariantProjector` is affected, and the reference identity is unchanged (see below).

## The problem R4 names

The supplemental transcript sidecar (`patterns_transcripts.metadata.json`) persisted `sequence_length` as a field on `SupplementalTranscript`, and "hollow" (#1722) was a *value* of it — `sequence_length == 0` for an accession whose bases are actually on disk. Because the type could represent that state, the code had to detect it (`is_hollow`, `hollow_record_count`) and repair it from the FASTA before use (`repair_derived_lengths`, `supplemental_fasta_lengths`, `write_supplemental_metadata`, the `commit_or_repair` sidecar-only write). The shipped artifact carried 116,572 hollow rows of 140,027.

R4 (accepted operator ruling, the principled follow-up to #1732's tactical fix): drop `sequence_length` as a persisted field and let the FASTA index be its sole authority, so a hollow row becomes **unrepresentable** rather than detectable-and-repairable.

## Why this is a deletion, not a rewrite

#1722 had already routed the one real length consumer off the persisted field: `MultiFastaProvider::get_supplemental_cds_info` takes its `sequence_length` from `entry.length` (the FASTA index), not the sidecar. So the persisted `sequence_length` fed only the hollow/repair machinery and a load-time reporting heuristic. With the field gone there is no hollow state to detect or repair, and the whole repair pass is removed.

## Type before / after

`SupplementalTranscript` (persisted, `src/benchmark/cache.rs`):

```rust
// before
pub struct SupplementalTranscript { id, gene_symbol, cds_start, cds_end, sequence_length: usize }
// after — only sourced fields; a length-0 "hollow" row cannot be constructed
pub struct SupplementalTranscript { id, gene_symbol, cds_start, cds_end }
```

`SupplementalTranscriptInfo` (consumer, `src/reference/multi_fasta.rs`): `sequence_length: u64` dropped; `carries_cds_or_length` → `carries_a_cds_bound` (length is present for every row via the index, so it never distinguished rows); the load path no longer reads the JSON `sequence_length` key.

## Consumers changed

- `SupplementalStore::synthesized_metadata` — no longer sets a length.
- `fetch_fasta_to_file_with` — the re-derive/repair block is gone.
- `commit_or_repair` → `commit_if_dirty` — its sidecar-only-write branch was dead once repair was removed.
- Deleted: `is_hollow`, `hollow_record_count`, `repair_derived_lengths`, `supplemental_fasta_lengths`, `write_supplemental_metadata`.

## The stronger invariant (R4 second half)

"Repairing a derived field makes no source call" is now **structural** — there is no derived field to repair. Pinned by `an_annotation_free_sidecar_row_does_not_produce_a_second_fasta_record` (a fixed point with `source.calls() == 0` and every artifact byte-identical), and by a new `cache.rs` test asserting the serialized row carries no length key.

## Backward compatibility

A pre-R4 sidecar still carries `sequence_length`; serde ignores the now-unknown key on load, so existing on-disk artifacts read fine. Pinned by `an_older_sidecar_that_still_carries_a_length_loads_without_it`. A later commit that touches the sidecar drops the field; a run that only reads it does not rewrite it (no churn).

## Reference identity

Unchanged. The sidecar is not in `CONTENT_STAMPED_ARTIFACTS`, `supplemental_fasta` is explicitly excluded from stamping, and the basename is unchanged — so neither Part 1 (basenames) nor Part 2 (content stamps) of the identity moves. Blessed identity `aa8b3246d83055cc` is unaffected.

## Verification

- `cargo nextest run --features dev --no-fail-fast` — 11218 passed, 0 failed, 155 skipped.

### DIFF
--- a/src/benchmark/cache.rs
+++ b/src/benchmark/cache.rs
@@ -36,7 +36,7 @@ use serde_json::json;
 use std::collections::{HashMap, HashSet};
 use std::fs::File;
 use std::io::{BufRead, BufReader};
-use std::path::{Path, PathBuf};
+use std::path::Path;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
 /// Get NCBI API key from environment variable.
@@ -1914,6 +1914,17 @@ fn fetch_fasta_fallback(
 }
 
 /// Supplemental transcript metadata extracted from GenBank records.
+///
+/// **Every field here is *sourced*** — only a GenBank record supplies it, and a FASTA cannot
+/// re-derive it. The one thing that used to sit beside them, `sequence_length`, is *derived*:
+/// its authority is the accession's bases in the supplemental FASTA. R4 dropped it from this
+/// type so the derived/sourced split is explicit in the shape rather than in prose. That
+/// makes a "hollow" row — one that claims a length it has no bases behind, the shape 116,572
+/// of the shipped `patterns_transcripts.metadata.json`'s 140,027 rows had (#1722) —
+/// *unrepresentable*: there is no length field to be zero, stale, or otherwise wrong, and so
+/// nothing to detect and repair. Length is read straight from the FASTA index at every point
+/// that needs it (see `MultiFastaProvider::get_supplemental_cds_info`, which already sourced
+/// it there since #1722).
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 pub struct SupplementalTranscript {
     /// Transcript accession (e.g., "NM_033517.1")
@@ -1924,8 +1935,6 @@ pub struct SupplementalTranscript {
     pub cds_start: Option<u64>,
     /// CDS end position (1-based, inclusive)
     pub cds_end: Option<u64>,
-    /// Sequence length
-    pub sequence_length: usize,
 }
 
 /// Supplemental metadata for transcripts not in cdot.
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #2060, opened 2026-08-16
test(ci): derive the release tag gate's workflow set instead of naming it

nh13:
`tests/it/release_tag_gate.rs` was written against one workflow and says so in its source: `const WHEELS_YML: &str = ".github/workflows/release-wheels.yml"`, with `GATED_JOBS` and `PUBLISHING_JOBS` hardcoded to that file's job names beside it. That is a roster, and it has the same drift problem one level up from the one the guard was written to close — a second release-triggered workflow is invisible to it.

It did not stay hypothetical. #2055 adds `.github/workflows/release-binaries.yml`, carrying the same `release: published` trigger and a copy of the same `startsWith(github.event.release.tag_name, 'v')` gate. Measured on a tree with that workflow present: **deleting its gate line outright leaves all four existing `release_tag_gate` tests passing, EXIT=0.**

## What changes

The workflow set is derived from `.github/workflows/` rather than named — every file whose `on:` block declares a `release` trigger whose `types:` includes `published`, or declares no `types:` at all, which in GitHub Actions means every activity type. `publish.yml` triggers on `push: main` and is correctly out of scope. This follows the precedent #2033 set for `spec_fixture_setup_filter.rs`: derive the set, report both difference directions on a disagreement, and floor it so a matcher that has gone blind cannot pass as "nothing to check".

The `on:` key is looked up as both the string `"on"` and the boolean `true`, because YAML 1.1 resolves a bare `on` key to a boolean and YAML 1.2 leaves it a string — this guard's correctness should not depend on which one a given serde_yaml release produces.

Both hardcoded job rosters go the same way. The gated set is derived from `needs:`: a job with none is one a release reaches directly, which is `{linux, macos, windows, sdist}` on `release-wheels.yml` and `{build}` on `release-binaries.yml`. The publishing set is derived from step text — `gh release upload`, `cargo publish`, `pypa/gh-action-pypi-publish` — rather than from `["upload", "publish-pypi"]`, which named one workflow's two jobs and could say nothing whatever about a second workflow's `upload`.

## Four assertions, and which one is load-bearing

1. `every_release_reachable_root_job_is_tag_gated` — new. Every job a release reaches directly carries a readable gate, or an `if:` that cannot be true on a release.
2. `all_release_gates_accept_the_minted_tag_and_reject_the_data_release` — the two existing assertions, now run over every gate in every derived workflow.
3. `every_release_workflow_gates_on_the_same_tag_prefixes` — replaces `every_building_job_carries_the_tag_gate`, strengthened from within-file to across-file.
4. `every_publishing_job_requires_a_release` — unchanged in substance, derived publisher set.

Assertion 1 is not garnish. **A gate that has been deleted is invisible to any check that iterates over gates it can find**, so 2 and 3 both stay green through a deletion — they iterate over the gates present, and a deleted gate is not one of them. Only iterating over *jobs* and asking each whether it is gated catches it.

## Measurements

Both new assertions were proven able to fail, each in isolation, on a scratch worktree holding this branch plus #2055's commit cherry-picked (so both release workflows are present). `release-binaries.yml` was restored byte-for-byte after each, verified by `shasum -a 256` and by an empty `git diff` against #2055's commit.

| tree | result |
|---|---|
| clean, both workflows | 4 passed. `5 gates checked, all accepting {"v"}` (4 wheels + 1 binaries), 5 release-reachable root jobs |
| gate line deleted from `release-binaries.yml` | 3 passed, **1 failed** — only `every_release_reachable_root_job_is_tag_gated`, naming the `build` job. The gate-iterating assertion reported `4 gates checked` and passed |
| binaries gate prefix widened `'v'` to `'v0'` | 3 passed, **1 failed** — only `every_release_workflow_gates_on_the_same_tag_prefixes` |

## The floor is 1, not 2, and deliberately

`MINIMUM_RELEASE_WORKFLOWS` is the direct analogue of `spec_fixture_setup_filter.rs`'s `MINIMUM_CONSUMERS`, and its whole job is that a parser which has silently stopped matching fails loudly rather than passing vacuously over an empty set. A floor of 1 does that job in full.

It is not 2 because `release-wheels.yml` is the only release-triggered workflow on `main` today. A floor of 2 would make this branch red against its own base, and the point of deriving the set is that this change and #2055 are independent and can land in either order. Raise it to 2 when `release-binaries.yml` lands.

## Scope

Touches `tests/it/release_tag_gate.rs` and nothing else. `tests/it/main.rs` already registers the module. Not stacked on #2055; the two cannot conflict.

Representation-Change: none

### DIFF
--- a/tests/it/release_tag_gate.rs
+++ b/tests/it/release_tag_gate.rs
@@ -1,5 +1,5 @@
-//! Ties the tag release-plz will actually mint to the tag prefixes
-//! `release-wheels.yml` will actually accept.
+//! Ties the tag release-plz will actually mint to the tag prefixes **every**
+//! release-triggered workflow will actually accept.
 //!
 //! # The failure this exists for
 //!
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #2061, opened 2026-08-16
docs: correct canonicalize_from_sequence gating premise (#1709)

### LINKED ISSUE #1709: canonicalize_from_sequence is not gated on members.len() > 1, and three committed places say it is
Three committed artifacts state that `merge::canonicalize_from_sequence` only runs for cis alleles with more than one member, and conclude from that premise that the 592 harvested multi-member rows are "the only real-world evidence the repo has for that code path". **The premise is false**, so the conclusion understates the path's reach and, worse, reads as a coverage claim while providing none.

## The code

`src/normalize/mod.rs` admits a single-member variant explicitly:

```rust
single if Self::is_splittable_single_member(single) => std::slice::from_ref(variant),
```

`is_splittable_single_member` accepts `Genome`, `Mt`, `Cds`, `Tx` and `Rna` and inspects the edit. So the multi-member branch (`allele.variants.len() < 2 => return None`) is one of **two** admitting routes, not the gate.

## The three places that say otherwise

| file | what it says |
|---|---|
| `examples/harvest_multi_member_cis.rs:5` | "`canonicalize_from_sequence` is gated on `members.len() > 1`" |
| `examples/harvest_multi_member_cis.rs:11` | "Those 592 are the only real-world evidence the repo has for that code path" |
| `tests/it/multi_member_cis_axis.rs:3` | "`canonicalize_from_sequence` is gated on `members.len() > 1`, so …" |

The generated fixture carries the same sentence in its `description` field, so it reappears on every regeneration.

## Why it matters beyond a stale comment

This is the shape the repository's own doctrine warns about — a corpus figure that reads as coverage. The 592 rows are a real corpus and remain useful, but they are not the population the path serves: splittable single-member inputs reach it too, and nothing currently measures how many. A recent measurement of one pass downstream of this entry point put the moved population at **5,275 rows**, roughly 9x the 592, which is the order-of-magnitude the premise conceals.

Concretely, anyone reasoning about blast radius from "the only real-world evidence is 592 rows" will under-scope it, and the two `examples/` and `tests/it/` doc comments are exactly where someone goes to check.

## Suggested fix

1. Correct the three prose sites to say the path has two admitting routes, naming `is_splittable_single_member`.
2. Regenerate the fixture so its `description` no longer carries the false premise.
3. Decide whether the harvester should also capture splittable single-member rows, or whether a second corpus is owed — the harvester's name and purpose are still sound for what it does, so this may be a scoping note rather than a code change.

Item 2 needs a fixture regeneration, which is why this is filed rather than fixed in passing.

Found while addressing review feedback on #1698; the false premise predates that PR and is independent of it.

nh13:
Closes #1709.

## Also corrected here

**The `650` was stale, not a second measurement.** `generate_cis_confluence_corpus.rs` and `cis_confluence_axis.rs` said 650 rows out of 9,949,738 while the fixture, the harvester, `assert_eq!(accepted, 592)` and the census pin said 592 — same denominator. Both figures landed in the same commit (`5a800838`, #1458): the fixture read 650 across that PR's pre-squash commits and 592 at the squash, because the selector was rewritten from a bracket scan to `parse_hgvs` so an inserted-range payload's own `;` stops counting as a member separator. Re-derived directly: `harvest_multi_member_cis -- --check` reports `592 multi-member cis alleles` and `up to date`. All three sites now say 592, with the provenance recorded so it is not "corrected" back.

**"Harvested" is not "reaches".** The harvest predicate checks only `phase == Cis && variants.len() > 1`; `sequence_first_pass` additionally refuses an `uncertain` allele and one `detect_overlap_conflicts` flags. `NC_000001.11:g.[(100del;110del)]` parses `uncertain: true, phase: Cis, n: 2` — harvested, declined. So 592 is an **upper bound**, and the prose now says so. Measured, the `uncertain` refusal removes none: 0 of the 733 raw multi-member cis alleles in the four corpora (592 after dedup) are uncertain, and all 592 committed rows parse certain. The overlap refusal is unmeasured and is stated as such.

**The single-member route's size is now measured, not asserted.** Over the same four corpora, **9,935,060 of 9,949,738 rows (99.85%)** parse to a single-member `g.`/`m.`/`c.`/`n.`/`r.` variant with a present edit. The PR previously called it "far more populous" with nothing behind it.

**`spec_corpus.rs`'s `MEMBER_COUNTS` doc is rewritten rather than annotated.** Its lead sentence ("Two is the smallest allele that reaches the partitioner at all") was the false premise, so it could not be kept beside the correction. The floor is 2 because these strata vary *relations between members* and a one-member design has none. Single-member rows do exist in the corpus and do reach the partitioner: **450** of 12,946 at default bounds (402 `Lone`, 24 composite-payload, 24 repeat-count), of which **404** parse and satisfy `is_splittable_single_member` — the other 46 are `RowKind::Prohibited` spellings that do not parse at all. The 450 is derivable from two figures `spec_conformance_axis` already pins (12,946 − 12,496), so it is guarded rather than a new unbacked number.

**A guard now ties the fixture `description` to the generator.** The literal is a named `DESCRIPTION` constant and `the_committed_fixture_description_matches_this_generator` asserts the committed JSON matches it. `--check` already compares the whole file but needs all four release-asset corpora, so it never runs in CI; this reads only the committed JSON and runs everywhere. Verified to fail when the JSON is mutated.

**Two `src/normalize/mod.rs` comments carried the false premise in stronger form and are corrected.** The lazy-clone rationale at the `canonicalize_from_sequence` loop justified itself on "`sequence_first_pass` serves only cis alleles and lone `delins`/`inv` members"; the optimisation survives, but for a better reason — the clone sits after all three `break`s, so a variant the pass *admits and reproduces unchanged* is also on the no-clone side. The `sequence_first_pass` arm comment claimed a lone `delins`/`inv` restriction that #1235 removed on purpose, and which `is_splittable_single_member`'s own doc already contradicted.

Representation-Change: none. Comment/docs-only. The two `src/normalize/mod.rs` edits are comments, the `src/conformance/spec_corpus.rs` edit is a doc comment on a `const` (and that tree cannot move normalized output anyway), and the JSON change is a `description` string. No code path changes.

## The defect

Six committed sites stated that `canonicalize_from_sequence` "is gated on `members.len() > 1`", and concluded from that premise that only multi-member cis alleles reach the partitioner and that the 592 harvested rows are the only real-world evidence for that code path. **The premise is false**, so the conclusion reads as a coverage claim while understating the path's reach.

## The real gating

The entry point routes through `sequence_first_pass` (`src/normalize/mod.rs`), which admits **two** shapes, not one:

- a multi-member cis allele — a non-uncertain `Cis` `Allele` with `variants.len() >= 2`; and
- a splittable single-member input, accepted by `is_splittable_single_member`.

So `members.len() > 1` is one of two admitting routes, not the gate.

`is_splittable_single_member` returns `edit.inner().is_some()` over the five `HgvsVariant` kinds `Genome`/`Mt`/`Cds`/`Tx`/`Rna` (`g.`/`m.`/`c.`/`n.`/`r.`). `Mu::inner()` is `Some` for `Certain` **or** `Uncertain` and `None` only for `Unknown` (`?`), so the single-member route admits **any** single-member `g.`/`m.`/`c.`/`n.`/`r.` variant with a present edit — not merely a lone `delins`/`inv`. The corrected prose says exactly that, rather than repeating the narrower "lone delins/inv" phrasing some in-code docs use.

## The fix

Corrected all six sites to state the two routes and name `is_splittable_single_member`, and narrowed the "real-world evidence" claims to the multi-member route specifically:

| file | site |
|---|---|
| `examples/harvest_multi_member_cis.rs` | module doc + the fixture `description` constant it emits |
| `tests/it/multi_member_cis_axis.rs` | module doc |
| `examples/generate_cis_confluence_corpus.rs` | module doc |
| `tests/it/cis_confluence_axis.rs` | module doc |
| `src/conformance/spec_corpus.rs` | `MEMBER_COUNTS` doc |
| `tests/fixtures/cis/multi_member_cis_alleles.json` | committed `description` |

The committed fixture's `description` is updated to byte-match the regenerated generator constant; its 592 rows and `rows_scanned` are unchanged (the bulk `.json.gz` corpora the harvester reads are release assets, absent on this checkout, so the rows could not and did not change — verified the new `description` equals the generator's produced string exactly).

## Unchanged on purpose

- No code behavior changes — this is comment/docs/description-only.
- Left `from_sequences_multi_member_axis.rs` and `tests/fixtures/CORPUS_LAYOUT.md`: they describe the separate `from_sequences` surface and do not state the false `members.len() > 1` gate.

## Verification

- `cargo nextest run --features dev` — `11240 passed, 0 failed, 155 skipped` (EXIT=0), including `only_a_certain_cis_allele_is_counted_as_multi_member`.
- JSON `description` confirmed byte-identical to the generator constant.

### DIFF
--- a/examples/generate_cis_confluence_corpus.rs
+++ b/examples/generate_cis_confluence_corpus.rs
@@ -3,14 +3,30 @@
 //!
 //! # Why this exists
 //!
-//! `canonicalize_from_sequence` is gated on `members.len() > 1`
-//! (`src/normalize/mod.rs`), so only a multi-member cis allele reaches the
-//! partitioner. Part 1 of this issue harvested every such input out of the bulk
-//! corpora: **650 rows out of 9 949 738**. Real corpora are observed variants,
+//! `canonicalize_from_sequence` (via `sequence_first_pass` in
+//! `src/normalize/mod.rs`) reaches the partitioner by two routes: a multi-member
+//! cis allele (`members.len() >= 2`), and a splittable single-member input — any
+//! single-member `g.`/`m.`/`c.`/`n.`/`r.` variant **with a present edit**, via
+//! `is_splittable_single_member` (a `?` edit is refused).
+//! So `members.len() > 1` is one of two admitting routes, not the gate. This
+//! corpus designs the multi-member route. Part 1 of this issue harvested every
+//! such input out of the bulk corpora: **592 rows out of 9 949 738** — and even
+//! that is an upper bound, since the harvest selects on shape and the gate
+//! refuses an uncertain or overlap-conflicting allele on top of it. Real corpora
+//! are observed variants,
 //! where two changes rarely land close enough to share a block, so no amount of
 //! real data will cover the partitioner. Consumers who submit *designed*
 //! alleles — a systematic indel design walking a window — have the opposite
 //! distribution, and that is the shape this generates.
+//!
+//! **The figure is 592, not 650.** 650 was this passage's number until #1709 and
+//! it was stale from the day it landed: #1458 measured 650 with a bracket-scan
+//! selector, replaced that selector with `parse_hgvs` inside the same PR — which
+//! is what stopped an inserted-range payload's own `;` counting as a member
+//! separator — and re-measured 592, updating the fixture and the harvester but
+//! not this file or `tests/it/cis_confluence_axis.rs`. 592 is what the committed
+//! fixture, `multi_member_cis_axis`'s `assert_eq!` and a re-run of the harvester
+//! all report.
 //!
 //! # What a confluence class is, and how it differs from a dump
 //!
--- a/examples/harvest_multi_member_cis.rs
+++ b/examples/harvest_multi_member_cis.rs
@@ -2,15 +2,38 @@
 //!
 //! # Why this exists
 //!
-//! `canonicalize_from_sequence` is gated on `members.len() > 1`
-//! (`src/normalize/mod.rs`), so a single-member input never reaches the
-//! partitioner, the changed-column bound, or the member-ordering rules. Across
-//! the four bulk corpora — 9 949 738 rows — only **592** inputs are
-//! multi-member cis alleles. Everything else exercises the per-member path.
+//! `canonicalize_from_sequence` (via `sequence_first_pass` in
+//! `src/normalize/mod.rs`) admits **two** shapes, not one: a multi-member cis
+//! allele (`members.len() >= 2`), and a splittable single-member input — any
+//! single-member `g.`/`m.`/`c.`/`n.`/`r.` variant **whose edit is present**
+//! (`is_splittable_single_member` is `edit.inner().is_some()`, so a `?` edit is
+//! refused). So `members.len() > 1` is one of two admitting routes, not the
+//! gate.
 //!
-//! Those 592 are the only real-world evidence the repo has for that code path,
-//! and they currently sit inside bulk corpora whose test modules *skip green*
-//! when the fixtures are absent (`clinvar_hgvs_tests`, `cmrg_exhaustive_tests`,
+//! This harvest captures the multi-member route only: across the four bulk
+//! corpora — 9 949 738 rows — **592** inputs are multi-member cis alleles. The
+//! single-member route is not a rounding error beside it. Measured over the same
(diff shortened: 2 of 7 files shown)


nh13:
@coderabbitai pause

Issue #1892, opened 2026-08-14
feat(normalize): an opt-in confluence self-check consumers can run on their own data


nh13:
The designed cis corpus is **96.5% multi-member against 0.006% in real corpora** — deliberately, since that is where the interactions live. But that means **the corpus cannot tell a consumer what rate they will see.** Only their own data can.

This is the instrument for the question consumers actually ask: *will my variants change, and how often?*

### Shape

A normalizer option — off by default, zero cost when off — that, for each input, also normalizes a small set of **mechanically derived alternative spellings of the same variant** and reports when they disagree.

```rust
NormalizeConfig::default().with_confluence_check(ConfluenceCheck::Spdi)
```

Derivation must not need a corpus. From a parsed variant you can generate re-spellings by re-partitioning the member set and by shifting each member within its own ambiguous run — the same two operations the corpus generator uses, applied to one input instead of a design.

### Output is a diagnostic, not an error

It reports `{input, outputs: [..], relation}` so the consumer can count it. Making it an error would make it unusable in a pipeline, and the whole value is that it can be left on in production and sampled.

### Build it last, on purpose

This is the only one of the three that needs a **new public API surface**, so it should be designed against a census definition that has already stopped moving. It also has to name its equivalence relation, which is the same parameter the census subcommand forces the caller to choose — and the two must not drift into separate definitions of "confluent".

### Provenance

Designed 2026-08-09 as one of three pieces. The `96.5%` / `0.006%` contrast is the design's own figure and predates the partition flip (#1835); the *argument* is unaffected — a corpus enriched for multi-member shapes cannot report a real-world rate whatever the exact ratio — but re-derive the numbers before quoting them.

Companion issues: the runnable conformance census, and the release-to-release stability job.

Issue #1891, opened 2026-08-14
feat(ci): measure release-to-release representation drift on the real corpus, on a schedule


nh13:
Nobody measures, on a schedule, the thing a consumer actually experiences: **release-to-release drift on real expressions.**

`examples/dump_normalized_corpus.rs` diffs two revisions, but it has two recorded limits — its corpus is deliberately enriched for churn-prone shapes, so its rate is not a repo-wide figure, and it has produced structurally-blind zeros three times (#1456 member geometry, #1460 scale, #1478 transcript geometry), each invisible until the previous was fixed.

### Shape

A job — nightly, and mandatory in the release workflow — that:

1. Normalizes a fixed real corpus (ClinVar / CMRG / Paraphase, the bulk fixtures already in the repo under `tests/fixtures/bulk/*.gz`) through **the last released version** and through **HEAD**, using the same prepared reference for both.
2. Diffs row by row, classifying each moved row as merge / split / respell, and separating "moved" from "denotes different bases" via SPDI verification.
3. Emits the exact sentence the `Representation-Change:` trailer needs.

It should consume the `Census` artifact from the census subcommand rather than defining its own counters — same instrument, pointed at two releases instead of two directions.

### The two traps it must avoid

**Comparing against `main` instead of the last release.** Release-to-release drift is what the consumer sees; `main`-to-branch is what the author sees. The v0.13.0 cycle is the cautionary case: 87 commits landed with no trailer, 46 of them touching the watched directories, and the moved strings had to be reconstructed after the fact by normalizing the whole corpus through both releases.

**Trusting a green run whose fixtures are absent.** `clinvar_hgvs_tests`, `cmrg_exhaustive_tests` and `paraphase_exhaustive_tests` read their fixtures at runtime and **skip green** when they are missing, printing `Skipping:` and returning. A stability job built on the same fixtures would report "0 rows moved" on a runner that never fetched them — a total coverage loss that reads as a clean bill of health, and the worst outcome available here because it also makes the job look *faster*. The job must assert the fixture row count before it trusts a zero, and should set `FERRO_REQUIRE_BULK_FIXTURES=1` so absence is a failure rather than a skip.

### Two corrections to the original design, which was written 2026-08-09

[...]

PR #2063, opened 2026-08-16
feat(benchmark): a runnable conformance census subcommand

### LINKED ISSUE #1890: feat(benchmark): a runnable conformance census — the instrument the stability harness and self-check both consume
The `spec_conformance_axis` test pins a `Census` vector, and that vector is the only place the project's conformance and confluence numbers live. There is no way to *run* it — to understand a number, to diff two revisions, or to produce a release burn-down — without editing a test.

This is the instrument the other two pieces consume, so it is first.

### What it is

A `ferro-benchmark conformance census` subcommand that runs a corpus through normalization and emits the same `Census` vector the test pins, as a machine-readable artifact plus a human summary. **The test stays the gate; the command is for understanding.**

```
ferro-benchmark conformance census \
    --corpus spec|<path>                      # the generated spec corpus, or a TSV of real expressions
    --direction 3prime                        # see note on 5' below
    --equivalence sequence|partition|spdi     # REQUIRED, no default
    --reference <dir>                         # omitted for the synthetic corpus, which is hermetic
    --out census.json
    [--findings findings.jsonl]               # every counted row, not just totals
    [--compare <other census.json>]
```

### Two flags are load-bearing, and both come from things that actually went wrong

**`--equivalence` has no default.** The corpus builds each family by generating spellings that *re-partition* one variant, so confluence presumes those spellings are the same thing — which the partition ruling may deny. Three defensible relations answer three different questions:

| relation | equivalent when | answers |
|---|---|---|
| `sequence` | they denote the same bases | "can a consumer dedupe by normalizing?" |
| `partition` | same bases **and** same member partition | "is the normalizer deterministic within the ruling?" |
| `spdi` | their SPDI key sets match | "do two tools agree?" |

Neither is wrong; reporting one and calling it "confluence" is. `spdi` already has its backbone — `SpdiKey` / `group_by_spdi_key` in `src/equivalence/key.rs`.

**`--findings` emits rows, not counts.** The repo's recorded trap is *counts mislead; diff name lists*. A denominator has been read as a defect count more than once. A census without a row-level artifact cannot be checked.

### What it must refuse to do

- **Never print a single percentage.** Rank-1 validity and rank-2 confluence are not commensurable — they have moved in opposite directions in the same change, and any average reports "conformance got worse" about a correctness improvement.
- **Never report a zero without its denominator and the strata that produced it.** `examples/dump_normalized_corpus.rs` has produced structurally-blind zeros three separate times — member geometry (#1456), scale (#1460), transcript geometry (#1478) — each invisible until the previous was fixed. `--out` should carry the corpus's `CorpusShape` beside the census so a shrunken corpus cannot look like an improvement.
- **Never run with a seam oracle armed.** A panicking row contributes no output, so its family silently converges and the census reads *better* than the truth. The command should detect `FERRO_ASSERT_*` in its environment and refuse rather than produce a flattering artifact — the same reasoning that put `ORACLE_EXCLUDE` in `ci.yml`.

### Burn-down

`--compare` prints per-counter movement with the direction of virtue attached, since some counters should fall and some should rise:

```
outputs_denoting_no_sequence     10 ->  4   IMPROVED   (lower is better)
outputs_leaving_the_transcript  371 ->  0   IMPROVED
converged                      9140 -> 3509 WORSE      (higher is better)  <- explain this
```

The release gate is not "all zero". It is **"no rank-1 counter rose, and every rank-2 movement is declared"** — the `Representation-Change:` discipline, in numbers.

### Not settled here

Which relation is the *release* gate. Standing recommendation: gate rank-1 validity on `sequence`, rank-2 confluence on `partition`, and report all three — gating rank-2 on `sequence` under a partition-preserving normalizer sets a bar the ruling makes unreachable by construction.

### Provenance and staleness

Designed 2026-08-09. **Its motivating measurements are stale and must not be quoted**: they predate the partition flip (#1835, `750f2334`), and both of their classifications have since been closed — the "538-family ordering defect" figures are dead, and the "8 families at identical geometry" was ruled never enumerated and uncitable. The numeric example above is illustrative formatting, not a current measurement. The design itself is unaffected; re-derive every figure against current `main`.

Also note `--direction`: the 5' shuffle direction was removed from the public surface in #1879, so this takes `3prime` only unless the internal instrument is deliberately re-exposed.

Companion issues: the release-to-release stability job and the consumer-side confluence self-check, filed alongside this one. **The three are one system** — designed apart, they produce three definitions of "conformance" that disagree.

nh13:
Closes #1890.

Representation-Change: none. Additive tooling — no `src/normalize/`, `src/hgvs/`, `src/spdi/`, `src/project/`, `src/reference/` or `src/error_handling/` file is touched, and the pinned `spec_conformance_axis` census is byte-identical (both directions pass unchanged), so no normalized output moves.

## What this is

The `spec_conformance_axis` test pins a `Census` vector, and that vector was the only place the project's conformance and confluence numbers lived — there was no way to *run* the census (to understand a number, diff two revisions, or produce a burn-down) without editing a test.

This extracts the census measurement out of the test and into a library module (`src/conformance/census.rs`), then adds `ferro-benchmark conformance census` over it. It is the shared instrument the release-to-release stability harness (#1891) and the consumer-side self-check (#1892) are designed to consume, so the three produce one definition of "conformance" rather than three that drift.

**The test stays the gate.** `spec_conformance_axis` now consumes the extracted `measure` with `Equivalence::ExactString`, so its pins do not move — its `assert_eq!(census, pinned)` is the zero-drift cross-check on the extraction, and both the 3' and 5' pinned censuses pass unchanged.

## The command

```
ferro-benchmark conformance census \
    --equivalence sequence|partition|spdi   # REQUIRED, no default
    --corpus spec                            # hermetic; a real-expression corpus is #1892
    --direction 3prime                       # 5' was removed from the public surface in #1879
    --out census.json                        # counts + corpus shape + completeness
    --findings findings.jsonl                # every counted finding/divergence, not just totals
    --compare other-census.json              # per-counter burn-down
```

- **`--out`** writes a JSON artifact carrying the `Census`, the `CorpusShape` beside it (so a reported zero always carries its denominator — the #1460 hazard), and the corpus `CaptureCounts`. The human summary goes to **stderr**; stdout is left clean for piping.
- **`--findings`** writes the row-level detail as JSONL — **complete**, not a sample. The summary caps its divergence list at 12 for readability and says how many more there are; the export does not cap. (An earlier revision applied the summary's cap to the export too: an `--equivalence partition` run reporting 721 divergent families wrote 12, with nothing recording the truncation, so #1891's release-to-release diff would have read "no change" on a release in which hundreds moved.) Each row carries a `record` field (`"finding"` / `"divergence"`) so a consumer reads a tagged union rather than sniffing keys.
- **`--compare`** prints each counter's `before -> after` with the **direction of virtue** attached (`IMPROVED`/`WORSE`/`moved`, plus lower-is-better / higher-is-better / neutral). It never collapses to a single verdict or a percentage.

## Scope boundary honored (no adjudication)

The issue puts the *instrument* in scope and defers *which equivalence relation gates a release*. `--equivalence` is required with no default, so the instrument reports confluence under whichever relation the caller names and bakes in no gate. The three relations answer three different questions:

| relation | equivalent when | question |
|---|---|---|
| `sequence` | same denoted bases | can a consumer dedupe by normalizing? |
| `partition` | same bases **and** same member partition | is the normalizer deterministic within the ruling? |
| `spdi` | member SPDI key sets match | do two tools agree? |

`sequence` is backed by `spdi_key` over the whole allele; `spdi`/`partition` by the per-member keys — all derived without consulting the normalizer, so the instrument judges ferro from outside.

## Refuses to flatter itself

- **Seam oracle armed** — a `FERRO_ASSERT_*` variable panics on the very rows the census counts (a panicking row is filed `declined` and never reaches its family's output set, so the family reads as converged). The command detects it and refuses (exit 2) rather than emit a flattering artifact — the same reasoning that put `ORACLE_EXCLUDE` in `ci.yml`.
- **Accounts for every drop** — the corpus enumeration is routed through `CaptureLedger` (the last fallible step before the artifact), so dropped designs are counted with reasons and stamped into `--out` as `CaptureCounts` (`attempted == succeeded + dropped`). A partial run cannot read as complete.
- **Unkeyable outputs are not swept up** — SPDI has no offset notation, so intronic outputs have no key under a real relation. A family carrying one is left `underdetermined` and the unkeyable output is counted (`confluence_unkeyable_outputs`), never folded into a converged bucket.

## Measured (current `origin/main`, 3', hermetic spec corpus)

`--equivalence sequence`: 58,552 outputs, 0 declined; converged 11,750, split-2 2, underdetermined 130 (148 outputs unkeyable); corpus `attempted 16,650 = succeeded 12,946 + dropped 3,704` (with per-reason breakdown). `partition` converges fewer (11,031), confirming the strictness order the instrument reports.

## Tests

- `tests/it/conformance_census_instrument.rs` exercises the instrument: corpus completeness is accounted (`attempted == succeeded + dropped`, dropped > 0, every drop has a reason), a looser relation converges at least as many families as a stricter one, unkeyable outputs are counted and their families left undecided (the confluence buckets partition the family rows), the seam-oracle **refusal** is exercised end-to-end through a subprocess with a controlled environment (exit 2, the variable named, and no census summary printed — so the refusal is proven to precede the measurement), while the prefix rule itself stays a pure test over hand-written key lists; every divergent family reaches the export (asserted against the census's own split buckets, under `partition` because `sequence` finds only 2 and the assertion would be vacuous); the completeness allowance is a ceiling that refuses something, the burn-down attaches the correct direction of virtue, and the artifact round-trips through JSON. The oracle and burn-down tests are pure (hand-built inputs) so they cost nothing and do not race the environment.
- `spec_conformance_axis` (the gate) passes unchanged — the extraction is byte-identical.

`conformance_census_instrument` is added to `ci.yml`'s `ORACLE_EXCLUDE`, and `oracle_exclude_invariant.rs`'s corpus-import scan is widened to recognise `conformance::census`. The module measures the census while importing nothing named `spec_corpus`, so the scan could not see it and `test-oracle` was running it armed — the flattering-census shape that file exists to prevent.

## Verification

- `cargo nextest run --features dev --no-fail-fast` — `11251 passed, 0 failed, 155 skipped`
- `ferro-benchmark conformance census` exercised end-to-end (sequence/partition/spdi, `--out`/`--findings`/`--compare`, oracle refusal, required-flag enforcement).


nh13:
@coderabbitai pause

PR #2066, opened 2026-08-16
test: make two reference-aware guards visible to the manifest census

### LINKED ISSUE #1862: test: two reference-aware guards are #[ignore]d AND relative-path-gated, so they are invisible to the manifest census twice over
## Summary

Two reference-aware guards are hidden from #1851's census twice over: they are `#[ignore]`d **and** they gate on `Path::new("benchmark-output/manifest.json")` only, ignoring `FERRO_MANIFEST`. Neither is in `REFERENCE_AWARE_SELECTION`, neither runs in any job, and if either were un-ignored today it would skip green rather than run.

| guard | file | `assert*!` |
|---|---|---:|
| `test_flt3_itd_insertion_position_order` | `tests/it/coverage_gap_tests.rs` | 2 |
| `test_real_repeat_pattern` | `tests/it/normalize_tests.rs` | 0 |

## Why the census missed them

#1851 measured the orphaned surface empirically, as a duration diff over a `FERRO_MANIFEST`-set/unset pair — the right instrument, and it recorded its own blind spot: `real_data_normalization_tests` was "invisible to a `FERRO_MANIFEST` census — it gates on the relative path". These two are in that same blind spot, and `#[ignore]` puts them past a second filter, since an ignored test contributes no duration to diff at all.

So the shape is: a guard invisible to the instrument that found 45 of its siblings, for two independent reasons.

## Why the first one is worth more than an ignored test usually is

`test_flt3_itd_insertion_position_order` is not an investigation script. It carries two real assertions pinning an adjudicated spec question — that an intronic insertion's flanking positions must be written 5'-to-3' in **coding** order (`c.1837+2_1837+3`) and not in genomic order (`c.1837+3_1837+2`) — against a named Mutalyzer disagreement on a minus-strand gene, with issue #11 cited in its comment. That is an `adjudicated-correct` record in everything but effect, and it has never executed.

`test_real_repeat_pattern` is the other kind: zero assertions, `println!` only, a `match` on the result that prints either arm. It is an investigation script and the honest options for it are the ones in #1858 — assert the answer, or delete.

## What is needed

- Give both the `FERRO_MANIFEST`-then-`benchmark-output` gate every other reference-aware module now uses, routed through `ferro_hgvs::conformance::manifest_gate`, so a stand-down is a failure under `FERRO_REQUIRE_MANIFEST` rather than a green return.
- Drop `#[ignore]` once the gate is in place — the gate is what `#[ignore]` was standing in for, and it is the better mechanism because it distinguishes "no reference here" from "the reference is here and this did not run".
- Add them to `REFERENCE_AWARE_SELECTION` in `nightly-mutalyzer.yml`, which is the only job that provisions a manifest. Note the selection is deliberately enumerated rather than globbed; see #1851 for why widening by pattern is wrong.
- Decide `test_real_repeat_pattern` per #1858's options rather than arming an assertion-free guard.

## Note on scope

Not fixed alongside the residual work in #1861, which was deliberately held to the two instances #1849 named. Filed so the census's blind spot is recorded with its members rather than rediscovered.

nh13:
Closes #1862.

## The defect

Two reference-aware guards were hidden from #1851's manifest census twice over: both were `#[ignore]`d **and** gated on `Path::new("benchmark-output/manifest.json")` alone, ignoring `FERRO_MANIFEST`. The `#[ignore]` kept them out of every run, and the cwd-relative gate (the class #2019/#2032 address generally) meant that even un-ignored they could not see the manifest the nightly provisions. Neither was in `REFERENCE_AWARE_SELECTION`. So even on the one job that prepares a reference, these two never asserted.

| guard | file | before |
|---|---|---|
| `coverage_gap_tests::issue_11::test_flt3_itd_insertion_position_order` | `tests/it/coverage_gap_tests.rs` | 2 assertions (adjudicated-correct, issue #11) |
| `normalize_tests::real_repeat_tests::test_real_repeat_pattern` | `tests/it/normalize_tests.rs` | 0 assertions — `println!` only |

## The fix

- Give both the `FERRO_MANIFEST`-then-`benchmark-output` gate every other reference-aware module uses, routed through `ferro_hgvs::conformance::manifest_gate` (`absent` on a missing manifest, `unserved` when the reference does not carry the accession). A stand-down is now a *visible skip*, and a failure under `FERRO_REQUIRE_MANIFEST` rather than a green return.
- Drop the blanket `#[ignore]` — the gate is what it was standing in for, and it distinguishes "no reference here" from "the reference is here and this did not run".
- `test_flt3_itd_insertion_position_order` keeps its two assertions (they pin the decided `c.1837+2_1837+3` coding-order answer against the named Mutalyzer disagreement).
- `test_real_repeat_pattern` was an investigation script; per #1858's "assert the answer" option it now pins exact outputs measured against the prepared reference (cdot-0.2.32). Probed base by base, `NM_001407936.1` reads **CACAG** at `c.4261-4265`, preceded by `G` — so the `CA` unit *does* match at 4261 and the tract is exactly two copies wide, ending at 4264. Three descriptions over it are pinned:
  - **`c.4261_4264CA[1]` → `c.4263_4264del`.** This is the case that actually asks whether a shrunk repeat count is read as a deletion. It is — and **that is correct**: the tract holds two copies, one is asked for, so one goes. Pinned with that stated in the comment, so a maintainer meeting the deletion later does not file correct behaviour as a regression or "fix" it away.
  - **`c.4261_4262CA[1]` → itself.** Span equal to exactly one unit at count 1 is a no-op identity, and it round-trips.
  - **`c.4261_4262CA[2]` → `c.4263_4264dup`**, the non-vacuity control for the line above: the same span at a different count moves, so the no-op is a property of that one description and not of repeat descriptions being passed through.

  Each is pinned as an exact string rather than as `!contains("del")`, which a refusal or a dropped member would also satisfy. The no-op's expected value is written out rather than compared against the input: `c.4261_4262=` also normalizes to itself here, so two legal spellings of "no change" survive over that span, and `rulings[canonical-form-choice-when-both-legal]` derives the emitted form from the resulting sequence rather than preserving the input's spelling — an `assert_eq!(rendered, input)` would have pinned spelling-preservation as though it were policy, and would be satisfied by any implementation that echoed its input.

  **An earlier revision of this PR claimed the opposite** — that `c.4261_4262CA[1]` asks for a unit "at a position where the reference does not carry that repeat" and that "a non-matching repeat unit must be preserved unchanged, never read as a deletion". Both halves are false: the reference does carry the unit, and a genuinely non-matching one (`c.4261_4262GT[1]`) is refused before normalization runs (`expected GT[1] (GT), found CA`), so no ferro output could witness the stated property.
- Enroll both **by name** in `nightly-mutalyzer.yml`'s `REFERENCE_AWARE_SELECTION` (the only job that provisions a manifest), as specific test paths — the modules they live in carry many non-reference-aware tests, so a module-level entry would be wrong. This matches the enumerated (not globbed) convention #1851 requires.

## Scope

Held to the two guards #1862 names. The residual `#[ignore]`d reference-aware tests in `normalize_tests.rs` (tracked in #811/#834) are out of scope, as #1862 states.

## Non-vacuity (verified, not assumed)

- **Unarmed** (no manifest): both *run* (no longer ignored) and skip **visibly** (a printed `skipping — no reference manifest` line); full suite green — `11250 tests run: 11250 passed, 153 skipped`.
- **Armed** (`FERRO_MANIFEST` set to the blessed reference): both run genuinely (16.7 s / 5.2 s, not a 0.02 s stand-down) and their assertions pass.
- **Sabotage**: changing the expected `c.4263_4264del` to a wrong value goes **red** armed, printing the real computed string (`left: "NM_001407936.1:c.4263_4264del"`), then restored — proving the assertions fire against real data rather than standing down. Armed runtime is 25.5 s, not a 0.02 s stand-down.
- **Promotion**: with no manifest and `FERRO_REQUIRE_MANIFEST=1`, both **fail** rather than skip green.

## Representation stability

Representation-Change: none. Tests and CI workflow only; no watched `src/` prefix is touched.

## Verification

```
cargo fmt --all --check                                      # EXIT=0
cargo clippy --all-features --all-targets -- -D warnings     # EXIT=0
cargo nextest run --features dev --no-fail-fast              # EXIT=0, 11250 passed
actionlint .github/workflows/nightly-mutalyzer.yml           # OK
```

### DIFF
--- a/.github/workflows/nightly-mutalyzer.yml
+++ b/.github/workflows/nightly-mutalyzer.yml
@@ -104,7 +104,9 @@ jobs:
         test(/protein_cis_delins_decomposition::/) |
         test(/real_data_normalization_tests::/) |
         test(/accession_inventory::tests::corpus_missing_ng_against_real_reference_when_available/) |
-        test(/ng_placement_builder::tests::derive_for_real_reference_when_available/)
+        test(/ng_placement_builder::tests::derive_for_real_reference_when_available/) |
+        test(/coverage_gap_tests::issue_11::test_flt3_itd_insertion_position_order/) |
+        test(/normalize_tests::real_repeat_tests::test_real_repeat_pattern/)
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
         with:
--- a/tests/it/coverage_gap_tests.rs
+++ b/tests/it/coverage_gap_tests.rs
@@ -790,9 +790,45 @@ mod inversions {
 
 mod issue_11 {
     use super::*;
+    use ferro_hgvs::{MultiFastaProvider, ReferenceProvider};
+    use std::path::{Path, PathBuf};
+
+    /// This module's name, as it appears in a `FERRO_REQUIRE_MANIFEST` failure.
+    const MODULE: &str = "coverage_gap_tests::issue_11";
+
+    /// The prepared manifest, by the convention every reference-aware module
+    /// here uses: `FERRO_MANIFEST` is authoritative when set, with
+    /// `benchmark-output/manifest.json` as the well-known fallback.
+    ///
+    /// This guard used to consult **only** the relative path *and* carry a
+    /// blanket `#[ignore]`, so it was hidden from the manifest census twice
+    /// over (#1862): the ignore kept it out of every run, and even un-ignored
+    /// it could not see a `FERRO_MANIFEST` the nightly provisions.
+    fn manifest_path() -> Option<PathBuf> {
+        if let Ok(path) = std::env::var(crate::common::manifest::MANIFEST_ENV) {
+            let path = PathBuf::from(path);
+            return path.exists().then_some(path);
+        }
+        let fallback = Path::new("benchmark-output/manifest.json");
+        fallback.exists().then(|| fallback.to_path_buf())
+    }
+
+    /// The provider, or `None` when there is no manifest to build one from.
+    ///
+    /// A manifest that is present but will not load panics rather than
+    /// returning `None` — the same convention `real_data_normalization_tests`
+    /// adopts — so an absent-manifest stand-down is never confused with a
+    /// broken reference, which matters on the one job that exists to notice
+    /// `ferro prepare` left nothing usable behind.
+    fn get_provider() -> Option<MultiFastaProvider> {
+        let path = manifest_path()?;
+        Some(
+            MultiFastaProvider::from_manifest(&path)
+                .unwrap_or_else(|e| panic!("from_manifest({}) failed: {e}", path.display())),
+        )
+    }
 
     #[test]
-    #[ignore] // Requires benchmark-output reference data
     fn test_flt3_itd_insertion_position_order() {
         // Issue #11: FLT3 ITD insertion on minus-strand gene
         // Mutalyzer normalizes to: NM_004119.3:c.1837+2_1837+3ins...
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

PR #2067, opened 2026-08-16
test: pin the dominators oracle's blind spot to a too-small DAG

### LINKED ISSUE #1959: The brute-force dominators oracle cannot detect a DAG that is uniformly too small

nh13:
Closes #1959.

Representation-Change: none. Test/oracle code only; no watched output moves.

## The limitation

The brute-force dominator oracle `dominators_by_brute_force` (`src/normalize/seqfirst/align.rs`) — the probe behind `dominators_agree_with_brute_force_on_exhaustive_small_inputs` — is computed **from the DAG it is handed**: its cell set, its edges, its reachability walk. It re-derives nothing from the two sequences. So it establishes only that `AlignmentDag::dominators()` agrees with edge-banned reachability *over that graph*, and it **cannot detect a DAG that is uniformly too small** — a shrunken graph is internally consistent, so both sides read the same missing cells and the comparison is satisfied. It catches asymmetric or inconsistent damage; it is blind to uniform shrinkage.

This was previously argued only in prose (the module header and `rulings[unchanged-is-read-over-every-minimal-alignment]`). This PR turns it into an executable, self-proving guard and documents the limitation at the function.

## The fix (test-quality)

- **`the_brute_force_oracle_is_blind_to_a_uniformly_too_small_dag`** manufactures a too-small DAG the exact way the issue describes — a band one doubling step too narrow (`build_confined_to_band`, which mirrors `build`'s banded fill but skips the doubling search) — for two shapes:
  - `AC -> CA` at `k = 1`: retains 3 of the true DAG's 7 cells, a **strict subset**, with the **same** distance (2) and the **same** (empty) dominators — so nothing derived from the DAG alone, not even `dominators()`, reveals the shrinkage. The issue's verbatim example, and the subtlest form.
  - `CAG -> AGA` at `k = 1`: the band collapses to the all-substitution diagonal, so `dominators()` comes back **empty** where the truth is `{(1,0),(2,1)}` + forced junction 3 — a wrong answer the oracle nonetheless blesses.

  For both it asserts (1) the DAG is genuinely smaller, (2) the brute-force oracle **agrees** with `dominators()` on it — the agreement *is* the blind spot, (3) whether the dominator output stays correct or turns wrong, and (4) the independent `full_grid_build` **disagrees** — the comparison that catches uniform shrinkage.

- **Documents** the limitation on `dominators_by_brute_force` itself (issue suggestion 1), pointing to the `the_band_agrees_with_the_full_grid_*` differential tests as what does catch uniform shrinkage, and to why a DAG-construction change landing without such a second implementation would be unprotected (issue suggestion 3).

Issue suggestion 2 (qualifying the ledger record) was already done — `rulings[unchanged-is-read-over-every-minimal-alignment]` already carries the "READ THAT CORROBORATION NARROWLY (#1959)" paragraph on `origin/main`, so the ledger is untouched here.

## No production change

Only `src/normalize/seqfirst/align.rs` changes, entirely within `#[cfg(test)] mod tests` plus one doc comment. `AlignmentDag` behaviour is untouched. No proptest strategy is touched, so no seeds were invalidated.

## Guard proven non-vacuous

Per the repo doctrine ("prove the guard can fail"): widening the manufactured band so the DAG is **not** shrunk turns assertion (1) red (`the manufactured DAG must retain fewer cells than the truth (7 vs 7)`); restored.

## Verification

- `cargo nextest run --features dev --no-fail-fast` — `11249 tests run: 11249 passed, 155 skipped`

### DIFF
--- a/src/normalize/seqfirst/align.rs
+++ b/src/normalize/seqfirst/align.rs
@@ -961,6 +961,20 @@ mod tests {
     /// Brute-force oracle: an edge is a dominator iff deleting it disconnects
     /// the source from the sink. `O(E)` per edge — correct but far too slow for
     /// production, which is exactly why it belongs in the tests.
+    ///
+    /// **What it establishes, and what it does not (#1959).** Every term here —
+    /// `dag.cells()`, `dag.out_edges(..)`, the reachability walk — is read from
+    /// the DAG; nothing is re-derived from the two sequences. So it proves only
+    /// that [`AlignmentDag::dominators`] agrees with edge-banned reachability
+    /// *over this graph*. It **cannot** detect a DAG that is uniformly too small
+    /// — one missing cells that every minimal alignment of the sequences
+    /// actually occupies — because a shrunken graph is internally consistent and
+    /// both sides read the same missing cells. It catches asymmetric or
+    /// inconsistent damage; it is blind to uniform shrinkage. What does catch
+    /// that is the independent [`full_grid_build`], diffed against the shipped
+    /// `build` by the `the_band_agrees_with_the_full_grid_*` tests. See
+    /// [`the_brute_force_oracle_is_blind_to_a_uniformly_too_small_dag`], which
+    /// pins both halves.
     fn dominators_by_brute_force(dag: &AlignmentDag) -> (Vec<(u32, u32)>, Vec<u32>) {
         let sink = (dag.ref_len(), dag.alt_len());
         let mut matched = Vec::new();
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #2068, opened 2026-08-16
feat(equivalence): opt-in confluence self-check for consumer corpora

### LINKED ISSUE #1892: feat(normalize): an opt-in confluence self-check consumers can run on their own data

nh13:
Closes #1892.

Representation-Change: none. Additive diagnostic API; no watched directory (`src/normalize/`, `src/hgvs/`, `src/spdi/`, `src/project/`, `src/reference/`, `src/error_handling/`) is touched, and the check only reads normalization output — it never changes it.

## What this adds

`EquivalenceChecker::check_confluence(variants, relation)` — an **opt-in, diagnostic** confluence self-check a consumer can run over *their own* corpus. The designed cis corpus is enriched for multi-member shapes, so its non-confluence rate says nothing about a consumer's real data; this answers the question they can only answer for themselves: *are any two apply-equal descriptions in my call set normalized apart?*

It **composes decided machinery** rather than deciding anything new:

- Group the inputs into equivalence classes under a caller-chosen relation.
- Normalize each member.
- Report every class whose members reach more than one distinct normalized output — a non-confluence witness.

The grouping is derived from `apply` (`compare_denotations` / `spdi_key`), **never from the normalizer**, so it cannot collapse into the outputs it is checking — the non-circularity the `confluence-gate-is-apply-equality-on-every-determined-axis` ruling requires.

## The relation is the caller's parameter (this is not an adjudication)

Per the scope boundary in #1892, this does **not** decide which relation gates a release (that is the deferred adjudication in #1890). The relation is a parameter, and both options are already-decided, documented relations in this crate:

- `ConfluenceRelation::CrossAxisSequenceMatch` — apply-equality on every determined axis (`EquivalenceLevel::CrossAxisSequenceMatch`), the rung that establishes variant identity. The right default for "are two descriptions of one variant normalized apart?".
- `ConfluenceRelation::Spdi` — same denoted bases on the description's own axis (`SpdiKey`). Weaker and insufficient for identity (it buckets `c.3921dup` with `c.3922dup`); offered for counting distinct edits, per the issue naming it.

The result is a diagnostic report (`violations`, `classes_checked`, `skipped`, `relation`, `is_confluent`). `is_confluent()` is an observation about *this corpus under this relation*, never a pass/fail release verdict. Inputs the relation cannot place, and members the normalizer declines, are surfaced as **skips** — never silently dropped.

## API surface

Rust (`src/equivalence/`):
- `EquivalenceChecker::check_confluence(&[HgvsVariant], ConfluenceRelation) -> ConfluenceReport`
- `ConfluenceRelation` (`#[non_exhaustive]`), `ConfluenceGroup { inputs, outputs }`, `ConfluenceSkip { input, reason }`, `ConfluenceReport` (accessors + `into_parts`).

Python (`src/python.rs`, stub in `__init__.pyi`):
- `EquivalenceChecker.check_confluence(variants, relation="cross_axis"|"spdi") -> dict`, GIL released via `py.detach`; unknown relation raises `ValueError`.

## Testing

- Flag logic pinned deterministically as a unit test on `ConfluenceGroup::from_class` (confluent class -> no group; two distinct outputs -> a violation with deduped, sorted outputs; empty/singleton -> none).
- End-to-end integration (`tests/it/confluence_self_check_tests.rs`): confluent corpus reports clean, distinct variants form separate classes, unplaceable inputs are reported as skips, and the relation parameter changes the grouping (junction-straddling dups: one SPDI class vs two cross-axis classes).
- Python parity (`tests/python/test_issue_1892_confluence_self_check.py`).

**The headline behaviour has an end-to-end guard.** An earlier revision of this PR argued no hermetic "real violation" test was possible, and that was wrong: replacing `violations.push(group)` with a no-op left every test in the repository green. Two `MockProvider`-only witnesses now close it, each a pair the cross-axis relation groups into one class whose members normalize to two distinct strings — `g.[3A>G;10=]` vs `g.3A>G` (an `=` member changes nothing, so both apply-equal; the normalizer preserves it), and `g.10=` vs `g.20=` (two no-ops, which `src/equivalence/key.rs` records as a deliberate non-choice rather than a defect). The `violations.push` no-op now turns both red.

Each guard is verified by mutation rather than assumed: `violations.push` → no-op turns 2 tests red; `group_by_cross_axis` never merging turns 5 red (it turned 1 red before this revision — `the_relation_parameter_changes_the_grouping` and `distinct_variants_form_separate_classes` asserted a class count that a disabled grouping also produces, and both now use a three-input corpus reducing to two classes); removing the denotability screen turns 3 red; removing the undecided-pair count turns 1 red.

## What the run could not cover is never silence

A skip that reads as a pass is the failure this check exists to prevent, and there are two ways to lose coverage, accounted for separately.

Under either relation, an input with no computable denotation **on this reference** is reported as an `Unplaceable` skip carrying the underlying refusal — so a consumer with a partly-provisioned reference is told the *reference* declined, not that its own well-formed description denotes nothing. Under `cross_axis`, a *pair* whose comparison came back undecidable (over the checker's window cap, or a shape the relation errors on) fails to merge exactly as a decided-apart pair does; each is counted in `undecided_pairs`, because otherwise a class the checker could not examine is split into singletons and reported identically to a clean corpus.

`is_confluent` is a statement about the whole corpus only when `is_complete` is true. Coverage is `classes_checked` plus the `Unplaceable` skips: a `NormalizationDeclined` input **was** placed and its class **is** counted, so `classes_checked + len(skipped)` double-counts.

## Cost

`spdi` keys each input once (`O(n)`); `cross_axis` has no key and compares every pair (`O(n²)` `compare_denotations` calls, each doing its own SPDI conversion and reference fetches). Documented on the module. Deliberately no cap and no automatic downgrade: a cap must choose which pairs to stop comparing, and that shrinks the classes — reporting *fewer violations*, not more skips, which is the failure above. Batch by accession (classes never span accessions under either relation) or sample.

## Verification

- `cargo fmt --all --check`
- `cargo clippy --all-features --all-targets -- -D warnings`
- `FERRO_ASSERT_IDEMPOTENT=1 FERRO_ASSERT_REPARSE=1 FERRO_ASSERT_IN_BOUNDS=1 FERRO_SWEEP_SEEDS=full cargo nextest run --features dev --no-fail-fast` -> 11262 passed, 154 skipped
- `ruff check` / `mypy python/ferro_hgvs/` clean
- `pytest tests/python/` -> 1086 passed, 8 skipped

### DIFF
--- a/python/ferro_hgvs/__init__.pyi
+++ b/python/ferro_hgvs/__init__.pyi
@@ -1302,6 +1302,76 @@ class EquivalenceChecker:
         """
         ...
 
+    def check_confluence(
+        self,
+        variants: list[HgvsVariant],
+        relation: Literal["cross_axis", "spdi"] = "cross_axis",
+    ) -> dict[str, Any]:
+        """Run the opt-in confluence self-check over a corpus of variants.
+
+        Groups the inputs into equivalence classes under ``relation`` and reports
+        every class whose members normalize to more than one distinct output — a
+        non-confluence witness. A diagnostic: it reports, and never emits a
+        pass/fail release verdict.
+
+        Args:
+            variants: The variants to check.
+            relation: ``"cross_axis"`` (default) is apply-equality on every
+                determined axis — the relation that establishes variant identity.
+                ``"spdi"`` groups by same denoted bases on the description's own
+                axis; it is weaker and insufficient for identity. Typed as a
+                ``Literal`` so a misspelling is a type error rather than a
+                runtime ``ValueError``; the binding still validates at runtime,
+                for callers that are not type-checked.
+
+                ``"cross_axis"`` has no key and so compares every pair, which is
+                quadratic in the corpus size; ``"spdi"`` keys each input once and
+                is linear. Batch by accession or sample for a large call set —
+                classes never span accessions under either relation.
+
+        Returns:
+            A dict with keys ``relation`` (str), ``is_confluent`` (bool),
+            ``is_complete`` (bool), ``classes_checked`` (int),
+            ``undecided_pairs`` (int), ``violations`` (list of
+            ``{"inputs": list[str], "outputs": list[str]}``), and ``skipped``
+            (list of ``{"input": str, "kind": str, "reason": str,
+            "decline_class": str | None, "decline_site": str | None}``).
+
+            ``is_confluent`` is an observation about this corpus under this
+            relation, not a release gate, and it is a statement about the *whole*
+            corpus only when ``is_complete`` is true — nothing skipped, and no
+            comparison undecidable. ``undecided_pairs`` counts comparisons that
+            came back undecidable rather than deciding a pair apart; such a pair
+            fails to merge exactly as a decided-apart pair does, so a non-zero
+            count means the classes are coarser than reported and violations
+            inside them are invisible. It is always 0 for ``"spdi"``, which
+            compares keys rather than pairs.
+
+            A skip's ``kind`` is ``"unplaceable"`` (the relation never placed it,
+            so it is in no class) or ``"normalization_declined"`` (it was placed,
+            and its class is counted, but it contributed no output). Coverage is
+            therefore ``classes_checked`` plus the ``"unplaceable"`` skips —
+            never ``classes_checked + len(skipped)``, which double-counts.
+
+            ``decline_class`` and ``decline_site`` are the skip's refusal as
+            machine-readable names rather than prose: the class is
+            ``"reference_unavailable"`` (the reference could not serve the bases,
+            so the description is not at fault) or ``"unrepresentable"`` (it
+            denotes no single sequence SPDI can address), and the site names
+            which check refused — ``"multi_molecule_allele"``,
+            ``"null_or_unknown_allele"``, ``"empty_allele"``,
+            ``"member_conversion"``, ``"cross_accession"`` or
+            ``"unresolved_accession"``. Compare these rather than searching
+            ``reason`` for a keyword; ``reason`` is a rendering of them and its
+            wording is not API. Both are ``None`` where there is no such refusal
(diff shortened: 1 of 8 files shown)


nh13:
@coderabbitai pause

PR #2069, opened 2026-08-16
fix(equivalence): route an edit_triples reference failure to Indeterminate, not NotEquivalent

### LINKED ISSUE #2056: A reference failure inside edit_triples is swallowed by .ok()? and answered NotEquivalent
`edit_triples` discards the error from `hgvs_to_spdi` with `.ok()?` (`src/equivalence/checker.rs:1094`), so a **reference-data** failure is indistinguishable from an edit SPDI cannot represent. Both land in `SequenceVerdict::Declined`, and the caller answers a decided `NotEquivalent` for a pair it never actually compared.

This is the residual of #1989 that #2053 does **not** close. #2053 splits `ReferenceUnavailable` out of the *fetch* path (`compare_triples`, `checker.rs:1283`), which is correct and worth having; this issue is the other route.

## Measured

On #2053's head, with a provider serving no bases for `NC_ABSENT.1`:

| pair | unserved reference | served `NC_TEST.1` = `ACGTACGTACGT` |
|---|---|---|
| `g.2del` vs `g.2delC` | **`NotEquivalent`** | `CrossAxisSequenceMatch` |
| `g.2dup` vs `g.2_3insC` | **`NotEquivalent`** | (equivalent) |

Same variant either way. The only difference is whether the reference could be read — which is exactly the confusion `Indeterminate` exists to prevent.

Also still answered `NotEquivalent`, on a **served** contig: two out-of-bounds substitutions (`g.20A>G` vs `g.21C>T` on a 12-base contig). That one is a reference-level failure too, and it takes the same route.

## Why this is the larger half

HGVS routinely omits the deleted or duplicated bases — `g.2del`, `c.10_12del`, `g.2dup` are the ordinary spellings. Every one of those needs the reference to build its SPDI, so every one takes the `edit_triples` route rather than the fetch route. #2053's own reproducer had to state **both** bases (`g.2del` vs `g.2delC`) to land inside the band it fixes.

So the reachable population of #1989's defect is mostly outside #2053.

## Suggested fix

Classify at `checker.rs:1094` rather than discarding: distinguish "this edit has no SPDI representation" from "the provider could not serve the bases needed to build one", and route the second to `ReferenceUnavailable` so `denotational_verdict` resolves it to `Indeterminate` alongside `WindowTooWide`.

Worth checking at the same time whether any other `.ok()?` in this file collapses the same distinction.

## Why it was not done in #2053

It is a wider change than the fetch-path split — it needs an error taxonomy from `hgvs_to_spdi` rather than a new enum variant at one return site — and #2053's guards are scoped to the path it does fix. #2053 has been retitled from `Closes #1989` to `Refs #1989` so this does not auto-close.

Refs #1989, #2053.

nh13:
Closes #2056.

REMOVE:

> Representation-Change: none. Touches only `src/equivalence/` (not a watched prefix) and tests; corrects an equivalence-checker verdict and feeds no normalization path, so no normalized output moves.

INSERT:

Representation-Change: none. This now also touches `src/spdi/convert.rs`, which **is** a watched prefix. Five intronic-position decline sites move from `ConversionError::MissingReferenceData` to a new `ConversionError::UnrepresentableInSpdi` — three on the provider-backed path (`resolve_cds_to_tx`, `resolve_tx_pos`, `resolve_rna_pos`) and two on the simple path (`require_simple_tx_pos`, `require_simple_rna_pos`) — and the rendered message changes from "missing reference data: …" to "cannot be expressed in SPDI: …". That is an error-surface change and nothing more: every description that converted before converts to the identical SPDI, every description that declined before still declines at the same site for the same reason, and no successful conversion's output changes. No normalized HGVS string moves, so a consumer keyed on normalized output has nothing to re-bucket.

## The defect

`EquivalenceChecker::edit_triples` discarded the error from `hgvs_to_spdi` with `.ok()?` (`src/equivalence/checker.rs`), so a member whose HGVS→SPDI conversion **needed reference bases the provider could not serve** became `None` — indistinguishable from an edit SPDI genuinely cannot represent. Both landed in `SequenceVerdict::Declined`, and the checker answered a decided `NotEquivalent` for a pair it never actually compared.

Measured on current `origin/main` (`a436b11d`), against a provider serving no bases for the accession:

| pair | unserved reference | served `NC_TEST.1` = `ACGTACGTACGT` |
|---|---|---|
| `g.2dup` vs `g.2_3insC` | **`NotEquivalent`** (both `check` and `compare_denotations`) | `CrossAxisSequenceMatch` (equivalent) |
| `g.2del` vs `g.2delC` | **`NotEquivalent`** (`compare_denotations`) | `CrossAxisSequenceMatch` (equivalent) |

Same variant either way; the only difference is whether `hgvs_to_spdi` could read the reference to build the short form's SPDI — exactly the confusion `Indeterminate` exists to prevent.

## Relationship to #1989 / #2053

This is the residual of #1989 that #2053 (already merged, `0dc68cde`) does **not** close. #2053 split a `ReferenceUnavailable` verdict out of the sequence rung's **window fetch** (`compare_triples`) and mapped it to `Indeterminate`. #2056 is the *other* reference route, one step earlier: the per-member HGVS→SPDI conversion inside `edit_triples`.

It is the larger half in practice. HGVS routinely omits the deleted/duplicated bases — `g.2del`, `g.2dup`, `c.10_12del` are the ordinary spellings — and every one of those must read the reference to build its SPDI, so every one takes the `edit_triples` route rather than the fetch route. (#2053's own reproducer had to state both bases, `g.2del` vs `g.2delC`, to land inside the band it fixes.)

The fix builds directly on #2053's machinery and does not depend on any unmerged work: it reuses the existing `SequenceVerdict::ReferenceUnavailable` → `Indeterminate` mapping.

## The fix

`edit_triples` now returns `Result<(String, Vec<SpdiVariant>), TripleDecline>`, where `TripleDecline` classifies the conversion error:

- `ConversionError::MissingReferenceData` / `ProviderRequired` → `ReferenceUnavailable` → `SequenceVerdict::ReferenceUnavailable` → `EquivalenceLevel::Indeterminate`.
- every other conversion failure (unrepresentable edit, cross-accession, multi-molecule/null/unknown allele, empty member list) → `Unrepresentable` → `SequenceVerdict::Declined` → `NotEquivalent` (unchanged).

The first member to decline decides. The other `edit_triples` callers (`second_axis`, `undecidable_reason`) are updated to the `Result` API; both treat either decline as "second axis not computed" / "not resolvable", which is behavior-preserving.

Two corrections from review, both in this PR.

**The classification had to be fixed at the source.** `ConversionError::MissingReferenceData` was not exclusively a reference-availability error: `src/spdi/convert.rs` also returned it for **intronic positions**, decided from the position's own spelling and — on the `c.` path — before the provider is consulted at all. Mapping it wholesale to `ReferenceUnavailable` turned correct answers into refusals on a **fully served** reference:

| pair (`MockProvider::with_test_data`, which serves `NM_000088.3` completely) | before | after |
|---|---|---|
| `c.10+5A>G` vs `c.20+5T>C` | `Indeterminate` | `NotEquivalent` |
| `c.10+5A>G` vs `c.10A>G` | `Indeterminate` | `NotEquivalent` |
| `n.10+5A>G` vs `n.20+5T>C` | `Indeterminate` | `NotEquivalent` |
| `r.10+5a>g` vs `r.20+5u>c` | `Indeterminate` | `NotEquivalent` |

Both entry points moved, and the attached note read "The reference needed to compare both variants could not be read from the provider" — false, since the reference was never consulted. `c.####+/-N` splice variants are one of the largest classes in real HGVS, and no existing test caught it.

Fixed at the source rather than by an exception list in the classifier, which is the shape that goes stale: a new `ConversionError::UnrepresentableInSpdi` now carries those decline sites, so `TripleDecline::from_conversion_error` is correct by construction. The invariant it establishes is that **`MissingReferenceData` means a better-provisioned provider would answer**. The two `MissingReferenceData` readers that key on it meaning "transcript unavailable" (`resolve_tx_exonic_bounded`, `resolve_rna_exonic_bounded`) are unaffected: both sit in the `else` branch of `tx_needs_provider` / `rna_needs_provider`, whose predicates return `true` for an intronic endpoint, so an intronic position could never reach either fallback.

**The wildcard arm is gone.** `ConversionError` is `#[non_exhaustive]`, but `checker.rs` is in the same crate, where that attribute has no effect — so `_ => Unrepresentable` protected nothing while guaranteeing that a future reference-availability variant would be silently misclassified, re-introducing this defect with no build error and no failing test. All arms are spelled out; adding a variant is now a compile error at the classification site.

## Unchanged on purpose (the discriminating cases)

- A genuinely-different pair on a **served** reference still reconstructs and reports `NotEquivalent` (`served_genuinely_different_short_forms_stay_not_equivalent`).
- A **served** equivalent pair still reaches a decided positive (`served_short_forms_still_reach_a_decided_positive`).
- The deliberate `NotEquivalent` declines #1989 documented are untouched: an RNA fusion, a **resolvable** cross-accession pair (`test_different_accessions_entirely_not_equivalent` still passes), and an un-appliable overlapping allele. Those never had the reference as the obstacle.
- An out-of-bounds substitution on a served contig (`g.20A>G` vs `g.21C>T`) is **not** in scope: its SPDI converts fine (a substitution states its own bases), so it takes the window-fetch route, not the `edit_triples` route, and stays `NotEquivalent`.

Two pre-existing inline unit tests (`NM_000088.3:c.10A>G` vs `NM_000088.4:c.20A>G`; `LRG_1t1` vs `LRG_1t2`) are `c.` pairs on transcripts the test provider does not serve. They fall through to the sequence rung, where the reference cannot be read, and previously relied on the swallowed error to report `NotEquivalent`. They now assert `Indeterminate` — the same direction #1989 established (`check_reports_indeterminate_when_the_reference_cannot_be_reconstructed`). Had both accessions been served they would resolve to distinct SPDI accessions and stay a decided negative.

## Verification

- `cargo nextest run --features dev --no-fail-fast` — **11256 passed, 0 failed, 154 skipped** (`EXIT=0`).
- New regression module `tests/it/issue_2056_edit_triples_reference_failure.rs` (all `MockProvider`/`SyntheticBuilder`-backed, so it runs on PR CI, not reference-gated): the two reproducers assert `Indeterminate`, plus served positive and served-genuinely-different controls.

## Representation stability

None. This corrects an equivalence-checker verdict; it does not touch any normalization path and moves no normalized output, so the `dump_normalized_corpus` disclosure does not apply.

### DIFF
--- a/src/equivalence/checker.rs
+++ b/src/equivalence/checker.rs
@@ -10,7 +10,7 @@ use crate::normalize::{NormalizeConfig, Normalizer};
 use crate::reference::transcript::Strand;
 use crate::reference::ReferenceProvider;
 use crate::sequence::reverse_complement;
-use crate::spdi::{hgvs_to_spdi, SpdiVariant};
+use crate::spdi::{hgvs_to_spdi, ConversionError, SpdiVariant};
 
 /// Largest reference window (in bases) a sequence-level equivalence rung will
 /// reconstruct, bounding the cost of a reference fetch.
@@ -541,19 +541,23 @@ impl<P: ReferenceProvider> EquivalenceChecker<P> {
                          sequence was reconstructed"
                     )));
             }
-            // The window is bounded and both sides convert to SPDI on a shared
-            // accession, but the provider could not serve its bases, so — as with
-            // `WindowTooWide` — nothing was reconstructed and nothing was
-            // compared (#1989). This is the "want of reference data" decline, and
-            // it must report `Indeterminate` for the same reason its sibling does:
-            // falling through to the `NotEquivalent` tail would assert a decided
-            // negative about a pair that was never examined. Pinned by
-            // `issue_1989_declined_is_indeterminate`.
+            // The reference needed to compare the pair could not be read, from
+            // either of two points: the window fetch, after both sides converted
+            // to SPDI on a shared accession (#1989/#2053), or a member's own
+            // HGVS→SPDI conversion, which the ordinary short forms must run
+            // against the provider (#2056). Either way — as with `WindowTooWide`
+            // — nothing was reconstructed and nothing was compared. This is the
+            // "want of reference data" decline, and it must report
+            // `Indeterminate` for the same reason its sibling does: falling
+            // through to the `NotEquivalent` tail would assert a decided negative
+            // about a pair that was never examined. Pinned by
+            // `issue_1989_declined_is_indeterminate` and
+            // `issue_2056_edit_triples_reference_failure`.
             SequenceVerdict::ReferenceUnavailable => {
                 return Ok(EquivalenceResult::new(EquivalenceLevel::Indeterminate)
                     .with_normalized(str1, str2)
                     .with_note(
-                        "The reference window covering both variants could not be read from the \
+                        "The reference needed to compare both variants could not be read from the \
                          provider, so neither resulting sequence was reconstructed"
                             .to_string(),
                     ));
@@ -566,17 +570,15 @@ impl<P: ReferenceProvider> EquivalenceChecker<P> {
             // `same_resulting_sequence`) and an un-appliable overlapping allele
             // (`issue_1244_equivalence_overlap_panic`), each pinned there.
             //
-            // Those are not all of it, and the arm above does not make them so.
-            // `edit_triples` discards its conversion error —
-            // `hgvs_to_spdi(member, provider).ok()?`, `checker.rs:1114` — so a
-            // member whose conversion needed the reference and could not read it
-            // also arrives here as `Declined` and is read as a negative:
-            // `NC_ABSENT.1:g.2del` vs `g.2delC` answers `NotEquivalent` against
-            // a provider serving no bases and `CrossAxisSequenceMatch` against a
-            // served contig. So `ReferenceUnavailable` above is the
-            // reference-level decline reachable from the WINDOW FETCH, not every
-            // reference-level decline; the rest is #2056, a follow-up on
-            // `edit_triples`. These two fall through.
+            // The reference-level declines both reach `ReferenceUnavailable`
+            // above and are handled there: the window fetch (#1989/#2053) and,
+            // as of #2056, a member's own HGVS→SPDI conversion failing for want
+            // of reference data (`edit_triples` now classifies that as
+            // `TripleDecline::ReferenceUnavailable` rather than swallowing it).
+            // So `NC_ABSENT.1:g.2del` vs `g.2delC` — `NotEquivalent` before, and
+            // `CrossAxisSequenceMatch` against a served contig — reaches the
+            // `Indeterminate` arm above and never falls through here. These two
+            // fall through.
             SequenceVerdict::Different | SequenceVerdict::Declined => {}
         }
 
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

PR #2070, opened 2026-08-16
test(normalize): guard #1908's length gate on a partitioner-derived block

### LINKED ISSUE #1914: test: #1908's length gate has no end-to-end guard — the equal-length case is held by a different condition
PR #1908 adds `merge::split_is_a_placed_gap_coincidence`, whose first condition is a length gate — it refuses when `result.len() >= reference.len()`, so the rule cannot enter the equal-length regime that #1904's `equal-length-block-column-correspondence-is-unique` governs. That separation is the whole basis for the two rules composing, so it needs a guard that actually holds.

It has one at the unit level and **not** end to end.

## Measured by sabotage, not assumed

Relaxing the gate (`>=` → `>`) on that branch:

- `merge::tests::the_placed_gap_predicate_discriminates` goes **red** — this is the real length-gate guard.
- `an_equal_length_block_keeps_its_split` stays **green**.

The end-to-end case survives the sabotage because its split is held apart by a *different* condition of the same predicate — a lone substitution (condition 4) — not by the length gate. So the test that reads like the end-to-end guard for this property is not guarding it.

## Why the missing case was not written

Per #1908's author, an end-to-end case in which **only** the length gate holds the line needs the canonical partitioner's compensating-gap geometry, and is not writable from the module the other cases live in.

## Why this matters more than an ordinary coverage gap

This is the shape the project already treats as first-class: a guard that runs, passes, and checks nothing. The limitation is recorded in the test's own doc comment, which means it will be read by whoever next changes that gate — i.e. at the moment it is least useful, and only if they look.

The two rules' composition (#1904 equal-length, #1908 unequal-length, disjoint by this gate) is currently asserted by reading rather than by a test that would fail if the regimes ever overlapped.

## What closing it takes

An end-to-end case built through the canonical partitioner where a compensating-gap geometry makes the length gate the only thing preventing the collapse, plus confirmation that it goes red under the same `>=` → `>` sabotage.

Found while landing #1610 (PR #1908). Related: #1904, #1878.

nh13:
#1908's length gate (`result.len() >= reference.len()` in `split_is_a_placed_gap_coincidence`) had no end-to-end guard on a block the canonical partitioner actually derives. `the_placed_gap_predicate_discriminates` pins the predicate's logic, but with hand-built pieces paired against an artificial equal-length `result` — it never witnesses the partitioner producing an equal-length block on which the length gate is the *only* condition holding the members apart.

No full `normalize` spelling isolates the gate either: measured while closing this, the `>=`→`>` sabotage moves **zero** rows over ~1.6M swept spanning-`delins` and cis inputs across all three partition arms, because an equal-length split a full normalize emits is always held by a substitution (rank-1 type, condition 4) or a compensating insertion (pure-insertion refusal) — never by the length gate.

This adds `the_length_gate_holds_a_partitioner_derived_equal_length_block`, which drives the canonical partitioner directly on `AACAAC → CGAGGA`. Its derived `[delins(0,1,"CG"); delins(2,4,"GG"); del(5,6)]` split is equal-length (the first member's +1 payload cancelled by the deleted last base), every separation is one base, and no member is a substitution/insertion — so conditions 2–4 all pass and the length gate is the sole decliner. The `>=`→`>` sabotage reddens it (proven non-vacuous, then restored). Also corrects the sibling doc's claim that a `partition_block` equal-length block is "always a run of substitutions".

Test-only; no product behavior change.

Closes #1914

Representation-Change: none

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -16478,6 +16478,113 @@ mod tests {
         );
     }
 
+    /// The length gate, exercised on a block the CANONICAL PARTITIONER actually
+    /// derives — the coverage [`the_placed_gap_predicate_discriminates`] cannot
+    /// give (#1914).
+    ///
+    /// That test hand-builds its pieces and pairs them with an *artificial*
+    /// `result` length (`issue` reconstructs to three bases, checked against a
+    /// four-base `result`), so it pins the predicate's **logic** at an equal
+    /// length it was never handed by a real partition. It does not witness that
+    /// the canonical partitioner ever produces an equal-length block on which
+    /// conditions 2, 3 and 4 all hold — the case where the length gate is the
+    /// **only** thing keeping the members split.
+    ///
+    /// `an_equal_length_block_keeps_its_split` (in
+    /// `tests/it/issue_1610_lone_unequal_length_delins.rs`) cannot witness it
+    /// either, and neither can any full `Normalizer::normalize` spelling. A full
+    /// normalization emits an equal-length split only when it buys a
+    /// higher-priority type (a lone substitution) or carries a compensating
+    /// *insertion*; the first makes the predicate decline on condition 4, the
+    /// second on the pure-insertion refusal — never on the length gate. Measured
+    /// closing #1914: the `>=`->`>` sabotage below moves **zero** rows over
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #2071, opened 2026-08-16
test: replace an inert assertion in the dominators blind-spot guard

nh13:
Follow-up to #2067 (merged as `bd1a68f0`). There is no issue for this — it is a defect in that PR's own test, found on review after it merged.

Representation-Change: none. Test code plus one doc comment; `AlignmentDag` behaviour is untouched.

## The defect: assertion (4) can never fail

`the_brute_force_oracle_is_blind_to_a_uniformly_too_small_dag` ended with

```rust
// (4) WHAT CLOSES THE GAP. The independent full-grid implementation
// disagrees with the shrunk DAG over the retained cell set — the
// comparison `the_band_agrees_with_the_full_grid_*` make, and the
// only thing that catches uniform shrinkage.
assert_ne!(shrunk_cells, true_cells, ...);
```

Assertion (1), forty lines above it, asserts `shrunk_cells.len() < true_cells.len()`. `Vec`'s `PartialEq` compares lengths before elements, so the inequality is already established by the time (4) runs — it can only fire in a state (1) forbids.

Worse than inert, the comment oversold it: it claimed to be the comparison that closes the gap while never comparing the **shipped** `build` against `full_grid_build`, which is what actually does, and which `the_band_agrees_with_the_full_grid_*` already make.

Two further weaknesses in the same test:

- `build_confined_to_band` is a ~55-line hand copy of `build`'s post-loop body with **nothing pinning it to the original**. Its doc says it "deliberately mirrors `build`'s banded fill", but if `build`'s fill, `out`-bit semantics or `BandLayout` change, the copy can drift while the test goes on reporting a faithful narrowing.
- Both pinned cases demonstrated the oracle's agreement as `empty == empty` — `produced` and `oracle` are both `([], [])`. A `dominators()` returning nothing at all, for any reason, would have satisfied the demonstration.

## The fix

**(4) now runs the real closing comparison, and pins which arm of it does the catching.** It calls `assert_band_matches_full_grid(r, a)` on the same inputs — green, so the closing comparison is genuinely available exactly where the oracle is blind — and then pins the **edit-distance** arm as blind wherever it is blind. That is a fact the surrounding prose already claimed ("the SAME distance (2)") and nothing asserted. With (3) already pinning the dominator arm, what is left is the retained-cell/out-edge arm, which (1) has shown disagrees — so on the two rows where both flags are `true` that arm is the only one that can see the narrowing.

**Two fixtures with non-empty dominator sets**, so the agreement is no longer only `empty == empty`:

| fixture | `k` | cells | distance | shrunk `dominators()` | truth |
|---|---|---|---|---|---|
| `AAAC -> ACCCAA` | 2 | 12 of 15, strict subset | 4 = 4 | `{(0,0)}` | `{(0,0)}` — agrees, non-empty |
| `AACA -> CCCAAC` | 2 | 8 of 14, strict subset | 4 = 4 | matched `{(2,2)}` + forced-ins `{4}` | `{}` / `{}` — a **wrong** non-empty answer, with no tell anywhere in the DAG |

The second is also the first fixture here to exercise `forced_ins_junctions` at all.

**`the_hand_copied_band_confined_build_still_matches_build`** pins the copy to `build`: handed a band wide enough to hold every minimal alignment (`n + m`, which is `build`'s own `widest`), the copy must reproduce `build` exactly. Compared against `build` and deliberately not against `full_grid_build`, which is the *pre-band* implementation — a copy agreeing with it could still have drifted from every band-specific detail the copy exists to reproduce.

`assert_dags_agree` and `label_for` are extracted from `assert_band_matches_full_grid` so the accessor sweep is not copied a second time.

## Proven by mutation, not asserted

- Flipping `distance_stays_correct` on the `AC -> CA` row turns the new (4) red, printing the real computed pair: `"AC" -> "CA": unexpected edit-distance (dis)agreement with the truth (2 vs 2)`.
- Three deliberate drifts applied to `build_confined_to_band` — dropping its `OUT_INS` arm, truncating its `out` loop from `0..=n` to `0..n`, swapping its `OUT_DEL`/`OUT_INS` bits — each turn the new pin red with `out-edges differ at (1, 2) on "AC" -> "CA"`.

**Recorded because it qualifies the claim:** the blind-spot test reddens on all three drifts too. What it never does is say *why* — it reports `a single crossing must be this cell's only out-edge`, `out_edges(5, 4) is off-grid (4 x 6)`, or a dominator disagreement, all of which read as `dominators()` or the DAG being wrong rather than as the copy having drifted. So the new pin's value is attribution plus a guarantee the other test does not offer, **not** a hole measured open. The doc comment says so rather than overclaiming.

## Verification

```
cargo nextest run --features dev --no-fail-fast           # EXIT=0, 11246 passed, 154 skipped
```

### DIFF
--- a/src/normalize/seqfirst/align.rs
+++ b/src/normalize/seqfirst/align.rs
@@ -1101,6 +1101,12 @@ mod tests {
     /// one is sound"). It deliberately mirrors `build`'s banded fill so the
     /// shrunk DAG is a faithful narrowing rather than a hand-corrupted one; it
     /// is NOT an independent oracle — that role is [`full_grid_build`]'s.
+    ///
+    /// "Mirrors" is a claim, and a hand copy has no mechanism of its own to keep
+    /// it true, so it is pinned rather than asserted:
+    /// [`the_hand_copied_band_confined_build_still_matches_build`] requires that
+    /// at a band wide enough to hold every minimal alignment this reproduces
+    /// `build` exactly.
     fn build_confined_to_band(ref_block: &[u8], alt_block: &[u8], k: usize) -> AlignmentDag {
         let n = ref_block.len();
         let m = alt_block.len();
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #2072, opened 2026-08-16
ci: run src/python.rs's Rust tests (split out extension-module feature)

### LINKED ISSUE #2046: src/python.rs's 10 Rust tests are compiled by CI but never run, so a break there passes every check
`src/python.rs` carries **10 `#[test]` functions that no CI job ever runs.** They compile in two places and execute in none, so a change that breaks one passes every check.

## Measured on `origin/main`

- The `dev` feature does **not** include `python`:
  ```
  dev = ["validation", "mmap", "benchmark", "parallel", "protein-fetch", "web-service", "dep:pulldown-cmark", "dep:anyhow", "dep:serde_yaml"]
  ```
  So `cargo nextest run --features dev` — the command every test job runs, and the one contributors are told to run — never compiles that file's test modules.
- The only `--features python` in `ci.yml` is `maturin build --release --features python` (`:2103`), which builds a wheel and runs no Rust test.
- `cargo clippy --all-features --all-targets` (`:834`) compiles the tests but does not execute them.
- `src/python.rs` itself says its tests "run under `cargo test --features python`". Nothing in CI runs that.

## How it surfaced

PR #2042 repairs a malformed cdot exon fixture at `src/reference/multi_fasta.rs:9632` and leaves a **byte-identical** copy at `src/python.rs:7797-7799`:

```
[48263025, 48263098, 0, 73, "M73"]
```

Five fields, with the CIGAR sitting where `tx_end` belongs, so index 4 is non-numeric. Under that PR's stricter loader the whole transcript is refused, which breaks `get_transcript_for_accession_build_aware_is_forwarded` (`src/python.rs:7770`) at its `.expect("build-aware resolution succeeds")` (`:7825`).

Verified by mutation rather than inferred: reverting the *repaired* `multi_fasta.rs` fixture to the five-field form turns `test_defer_grch37_secondary_cdot_when_key_present` red, with the loader logging `Loaded 0 GRCh37 transcripts (deferred secondary build)`. The `python.rs` sibling has the same shape and the same consequence — and nothing goes red, because nothing runs it.

So #2042 can merge fully green while leaving `cargo test --features python` broken on `main`.

## This is the #1505 / #1372 family, one feature over

`ci.yml:2176-2187` already documents this exact failure mode for the **default** feature set:

> So the DEFAULT feature set's unit tests — the `#[cfg(test)]` modules under `src/` — are compiled NOWHERE in CI, and a call from one of them into feature-gated code cannot turn any job red.
>
> Not hypothetical: #1372 put exactly such a call on `main` and it survived until an external contributor ran the `cargo test` this repository's own README documents. Four `E0599`s, invisible to all 34 checks, reported from outside the project.

The guard added for that is deliberately `--no-run`:

> `--no-run` on purpose: the claim being defended is that the default feature set BUILDS, not that these tests run a second time.

That is a reasonable scope for the default set, whose tests do run elsewhere under `dev`. **It does not transfer to `python`**, whose tests run nowhere at all — for this feature, building is exactly what already happens and exactly what is not enough.

## Suggested fix

Add a job (or extend an existing one) that actually runs them:

```
cargo nextest run --features python --lib
```

Scoping to `--lib` keeps it cheap — the integration suite does not need re-running under this feature; the gap is specifically the `#[cfg(test)]` modules inside `src/python.rs`.

Worth checking at the same time whether any other feature is in the same position — the general question is which feature sets have `#[cfg(test)]` modules that no CI command executes, as distinct from ones that no command compiles.

## Related, found in the same pass

Three further copies of the malformed fixture exist at `src/data/cdot.rs:6640`, `:6657` and `src/python.rs:7788`, plus `src/reference/multi_fasta.rs:9712`. Those are flat-format and route through `from_flat_format`, which #2042 deliberately leaves alone — unaffected today, but the same malformed shape, and worth normalising so a future strictness change does not have to find them again.

Refs #2042, #1505, #1372.

nh13:
Closes #2046.

## The defect

`src/python.rs` carries 10 pure-Rust `#[cfg(test)]` tests that CI **compiles but never runs**, so a break in one passes every check — the #1505 / #1372 dead-coverage family, one feature over.

- `cargo nextest run --features dev` (every sharded test job) never compiles them: `dev` does not include `python`.
- `maturin build --features python` (the only `--features python` in CI) builds a wheel and runs no Rust test.
- `cargo clippy --all-features --all-targets` compiles the test target but does not execute it.

## Why the issue's suggested command does not work as-is

The tests are pure — they never start the interpreter — so they only need the crate to **link**, which needs libpython. But `pyo3`'s `extension-module` feature deliberately suppresses libpython linkage: `is_linking_libpython_for_target` in `pyo3-build-config` returns `false` on Linux/macOS when it is set. A plain test/bin executable then fails to link its undefined Python symbols. macOS links such a binary via its `dynamic_lookup` special case (so the naive command *passes locally on a Mac*), but **Linux CI does not**.

## The fix

Take `extension-module` out of Cargo.toml's `python` feature and make it a separate feature:

- `pyo3 = { version = "0.28", features = ["abi3-py310"], optional = true }`
- `extension-module = ["dep:pyo3", "pyo3/extension-module"]`

A new step in the `Python Wheel Test` job (which already has a shared-library Python) then links libpython and runs the tests, pointing the loader at the interpreter's lib dir.

### The step is scoped, and runs under nextest

`Python Wheel Test` reports a **required** status check, and the lib suite is not safe to run in one process. `cargo test --features python --lib` runs all 4240 lib tests under libtest's thread pool, where:

- `normalize::merge::tests::a_declined_sequence_first_partition_is_counted` asserts deltas on process-global `AtomicU64` counters (`SEQFIRST_ATTEMPTED` / `SEQFIRST_DECLINED`), which any concurrent test reaching `partition_block_for_rule` corrupts;
- `parallel::tests::test_stress_concurrent_throughput` is a timing test sensitive to CPU oversubscription.

Measured at `--test-threads=32` on this branch: **4 of 6 runs of the unscoped command failed**, always the counter test. Unfixed, every PR in the repo would inherit a flaky required check, at a rate that rises the moment GitHub's standard runner gains cores.

`cargo nextest run --features python --lib --no-tests=fail -E 'test(python::tests::)'` fixes both halves: nextest is process-per-test, so the corruption is not expressible at any thread count, and the selection keeps the job to the tests it exists to run.

`--no-tests=fail` is the non-vacuity guard, and the second reason this is nextest rather than a filtered `cargo test`. Measured: an empty selection exits **4** under nextest (`error: no tests to run`), while `cargo test -- python::tests` reports `ok. 0 passed; 4240 filtered out` and exits **0** — a typo in the filter would silently delete this coverage.

| | tests run | wall | empty selection |
|---|---|---|---|
| `cargo test --lib` | 4240 | ~14 s, 4/6 fail at `-j32` | exit 0, silent |
| scoped nextest | 10 | 0.02 s, 10/10 green at `-j32` | exit 4, loud |

### `[tool.maturin] features` is a DEFAULT, not an addition

The first revision of this PR claimed maturin "merges" that list into every build. **It does not.** Measured with maturin 1.13.3 via a `cargo` shim that records the real invocation:

| `[tool.maturin] features` | CLI `--features` | cargo receives |
|---|---|---|
| `["python","extension-module"]` | `python` | `python` |
| `["python","extension-module"]` | *(none)* | `python` + `extension-module` |
| `["python"]` | *(none)* | `python` |
| `["python"]` | `python,extension-module` | `python,extension-module` |

A CLI `--features` flag **replaces** the pyproject list. So the entry is load-bearing for builds that pass no `--features` of their own — the PEP 517 backend (`pip install .`, `python -m build`, `uv`) — and is overridden for every explicit invocation in this repo.

Such a wheel is still a working extension module today, because maturin also sets `PYO3_BUILD_EXTENSION_MODULE=1`, which pyo3 0.28's `is_extension_module()` honours independently of the Cargo feature (whose doc comment there calls it *"(deprecated)"*). But that makes the wheel's linkage a property of **which maturin ran** — `ci.yml` installs an unpinned `maturin>=1.0,<2.0`, `release-wheels.yml` takes `maturin-action`'s default — where before this PR the Cargo feature guaranteed it.

So every explicit invocation now names the feature: `ci.yml`'s wheel step, `release-wheels.yml`'s three `maturin-action` blocks, `pyproject.toml`'s two `build-*` tasks, and `CONTRIBUTING.md`. That restores exactly the cargo feature resolution `main` has today and needs no version floor or build-time assertion. Verified: wheels built with and without the explicit feature are indistinguishable — no libpython link line, 99 undefined `_Py` symbols each.

`docs`/`CLAUDE.md` are updated to state the replace-not-merge rule, since the earlier wording would have led someone to delete a load-bearing line or rely on a merge that does not happen.

## Verification

- **10/10 stress reps green** at `--test-threads=32`, exactly 10 `python::tests::*` executed each time (0.024 s), against 4/6 failures and ~14 s for the unscoped command.
- **Non-vacuity proven**: sabotaging one assertion turns the run red — `10 tests run: 9 passed, 1 failed`, exit 100. `src/python.rs` restored byte-identically.
- `cargo fmt --all --check`, `cargo clippy --all-features --all-targets -- -D warnings`, `actionlint` — clean.
- `cargo nextest run --features dev --no-fail-fast` — **11245 passed, 0 failed**.
- `pytest tests/python/` against the built wheel — **1080 passed, 8 skipped**.

## Out of scope, filed separately

#2046's closing paragraph asked whether any other feature is in the same position. `hgvs-rs` is: three lib tests (`benchmark::hgvs_rs::tests::*`) exist only under it and no CI command runs them, plus a fourth instance at sub-test granularity in `src/service/tools/hgvs_rs.rs`. Filed as #2080 rather than folded in here — covering it means compiling `hgvs`, `postgres` and `seqrepo` in CI for three hermetic unit tests, which is its own trade. `slow-tests` is excluded deliberately and documented; `extension-module` and `dhat-heap` gate no code at all.

## Unchanged on purpose

`src/python.rs` itself is byte-identical (the sabotage was reverted).

Representation-Change: none

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2107,11 +2107,74 @@ jobs:
           # `rust-cache` block in this file.
           save-if: ${{ github.ref == 'refs/heads/main' }}
           key: python-wheel
+      - uses: taiki-e/install-action@8e83571c9a797c9e42a58141f098806bfec453a0  # nextest
+      # src/python.rs carries 10 pure-Rust `#[cfg(test)]` tests that only
+      # compile under the `python` feature — and, before #2046, ran NOWHERE:
+      # `dev` excludes `python` so the sharded test jobs never compiled them,
+      # `maturin build` runs no Rust test, and `cargo clippy --all-features`
+      # compiles without executing. A break in one passed every check (the
+      # #1505 / #1372 dead-coverage family, one feature over).
+      #
+      # They are pure (they never start the interpreter), so they only need the
+      # crate to LINK, which needs libpython. That is why `extension-module` is
+      # NOT part of Cargo.toml's `python` feature: with it set, PyO3 links no
+      # libpython on Linux (`is_linking_libpython_for_target` in
+      # pyo3-build-config), so this test executable would fail to link its
+      # undefined Python symbols. `cargo` never reads pyproject.toml, so this
+      # step gets `python` WITHOUT `extension-module` and links libpython; the
+      # wheel step below asks for both features BY NAME (see its comment — the
+      # `[tool.maturin] features` list does not reach a build that passes
+      # `--features` on the command line).
+      #
+      # SCOPED, and run under nextest rather than `cargo test`, because this
+      # job reports a REQUIRED status check and the lib suite is not safe to run
+      # in one process. `cargo test --features python --lib` ran all 4240 lib
+      # tests under libtest's thread pool, where
+      # `normalize::merge::tests::a_declined_sequence_first_partition_is_counted`
+      # asserts deltas on process-global `AtomicU64` counters (`SEQFIRST_*` in
+      # src/normalize/merge.rs) that any concurrent test reaching
+      # `partition_block_for_rule` corrupts, and
+      # `parallel::tests::test_stress_concurrent_throughput` is a timing test
+      # sensitive to oversubscription. Measured on this branch at
+      # `--test-threads=32`: 4 of 6 runs failed that way, i.e. a green PR's
+      # required check would flake for reasons unrelated to the PR. nextest is
+      # process-per-test, so the corruption is not expressible at any thread
+      # count, and the selection keeps the job to the tests it exists to run
+      # (10 tests in 0.02s, against 4240 in ~14s).
+      #
+      # `--no-tests=fail` is the non-vacuity guard, and is the other reason this
+      # is nextest: a selection matching nothing exits 4 here ("error: no tests
+      # to run"), where `cargo test -- python::tests` reports
+      # `ok. 0 passed; 4240 filtered out` and exits 0 — a typo in the filter
+      # would silently delete the coverage this job was added to provide. Same
+      # reasoning as the `soak` job's `--no-tests=fail`; see its comment.
+      #
+      # The test binary loads libpython at startup, so point the loader at the
+      # interpreter's shared-library dir.
+      - name: Run src/python.rs unit tests
+        run: |
+          LIBDIR="$(python -c 'import sysconfig; print(sysconfig.get_config_var("LIBDIR"))')"
+          echo "libpython dir: $LIBDIR"
+          LD_LIBRARY_PATH="$LIBDIR:${LD_LIBRARY_PATH:-}" \
+            cargo nextest run --features python --lib --no-tests=fail \
+              -E 'test(python::tests::)'
+      # `extension-module` is named explicitly, NOT inherited from
+      # `[tool.maturin] features`. Measured with maturin 1.13.3: a `--features`
+      # flag on the command line REPLACES that list rather than merging with it,
+      # so `maturin build --features python` hands cargo `--features python`
+      # alone and the pyproject entry is never consulted. Without the name here,
+      # the wheel's non-libpython linkage would rest solely on maturin setting
+      # `PYO3_BUILD_EXTENSION_MODULE=1` (which pyo3 honours independently of the
+      # Cargo feature) — true of every maturin in use today, but an unpinned
+      # `maturin>=1.0,<2.0` that predates it would silently build a
+      # libpython-linked wheel. Naming the feature restores exactly the cargo
+      # feature resolution this repo had before the split, and costs no version
+      # floor.
       - name: Build release wheel
         run: |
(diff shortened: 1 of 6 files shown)


nh13:
@coderabbitai pause

PR #2073, opened 2026-08-16
test: validate path.rs::name prose citations in the ruling ledger

### LINKED ISSUE #2016: Ruling prose "Pinned by" names two tests that do not exist, and nothing validates prose test names
## What

Two ruling records name a test in their **prose** ("Pinned by …") that does not exist. Nothing validates prose test names, so both have been silently wrong.

This is #2009's failure mode one register over: #2009 was a prose citation into the *spec* that no longer resolved; this is a prose citation into the *test suite* that never did.

## The two, measured on `origin/main`

**1. `alignment-only-symbol-in-a-description`** (`rulings[15]`, `decided`) — `rationale` says:

> Pinned by `tests/it/corpus_prohibited_inputs.rs::an_alignment_only_symbol_is_refused_for_dash_and_accepted_for_x`.

No such test exists. The real one is `…::an_alignment_only_symbol_is_refused_in_every_mode_for_both_x_and_dash` (`tests/it/corpus_prohibited_inputs.rs:673`).

**2. `delins-merge-vs-individual-gap-two-or-more`** (`rulings[0]`, `decided`) — `rationale` says:

> Pinned by `tests/it/spec_worked_example_rules.rs::the_spanning_delins_and_its_published_split_converge_on_the_recommended_form`

The *file* exists; it declares no test of that name.

```console
$ git grep -c 'an_alignment_only_symbol_is_refused_for_dash_and_accepted_for_x' origin/main -- '*.rs'
$ git grep -c 'the_spanning_delins_and_its_published_split_converge_on_the_recommended_form' origin/main -- '*.rs'
$ # both empty — zero occurrences anywhere in the tree
$ git grep -n 'fn an_alignment_only_symbol_is_refused' origin/main -- 'tests/it/*.rs'
origin/main:tests/it/corpus_prohibited_inputs.rs:673:fn an_alignment_only_symbol_is_refused_in_every_mode_for_both_x_and_dash() {
```

## Why nothing catches it

`ruling_guard_field.rs` validates the **structured** `guard.tests` array (`record.guard.tests()` at `:558`, `:661`, `:703`, `:737`) and never reads the prose. The two registers disagree *inside one record*: `rulings[15]`'s `guard.tests` names the correct test, while its own rationale names one that does not exist.

That is the same two-register split #2009 was about — the register the build checks is right, the register a human reads is wrong — and it fails in the more misleading direction, because a reader who follows the prose name concludes the ruling is unguarded when it is in fact guarded.

## Suggested fix

Resolve every `Pinned by \`…\`` name in ruling prose against the actual test set, the way `ruling_guard_field` already does for `guard.tests`. Points worth handling:

- Names appear in three shapes — `path::test_name`, bare `module::test_name`, and a bare **module/file** name (`reported_confluence_pairs`, `tests/it/issue_1870_cds_less_transcript_refusal.rs`). Only the first two are test names; a bare module reference should resolve to a module, not be reported as a missing test. Of the 11 `Pinned by` mentions currently in the ledger, 4 name a bare file/module and 7 name a test; 2 of those 7 do not resolve.
- The same prose convention is used in `//!`/`///` comments across `tests/it/` (`ruling_citation_currency.rs:99` documents it), so the guard could reasonably cover those too — but scoping it to the ledger first is the smaller, cheaper change.
- A repaired name should be pinned so it cannot rot again: the check should fail on an unresolvable name rather than warn.

## Scope

Pre-existing on both sides of #2014 (byte-identical at its base and head), so #2014 neither introduced nor touched it, and repairing it there would widen a deliberately scoped citation repair.

nh13:
Closes #2016.

## The defect

Two `decided` ruling records named a test in their **prose** ("Pinned by `…`") that does not exist, while their structured `guard.tests` — and the tree — carried the live name. Nothing validated prose-named citations, so both drifted silently, in the more misleading direction: a reader who follows the prose name concludes the ruling is unguarded when it is in fact guarded (the `guard.tests`/prose split lives *inside one record*).

Measured on current `origin/main`, the `path.rs::name` prose citations that do not resolve are exactly the two the issue names:

| record | prose named | the tree has |
|---|---|---|
| `alignment-only-symbol-in-a-description` | `…_refused_for_dash_and_accepted_for_x` | `…_refused_in_every_mode_for_both_x_and_dash` |
| `delins-merge-vs-individual-gap-two-or-more` | `…_converge_on_the_recommended_form` | `…_are_two_fixed_points` |

## The fix — two pure citation repairs

Only the stale test **name** changes in each record's prose. The rulings' decisions, `status`, clause citations and `guard.tests` are untouched.

- **`alignment-only-symbol-in-a-description`** → `tests/it/corpus_prohibited_inputs.rs::an_alignment_only_symbol_is_refused_in_every_mode_for_both_x_and_dash` — the live test, and the one the record's own `guard.tests` already names.
- **`delins-merge-vs-individual-gap-two-or-more`** → `tests/it/spec_worked_example_rules.rs::the_spanning_delins_and_its_published_split_are_two_fixed_points` — the test that pins the row (its body asserts the convergence after #1835). The sentence's trailing clause ("which asserted the two were fixed points until that ruling and required any converging change to name the clause it was reading") already describes that test, so no argument is rewritten.

`docs/NORMALIZATION_CONTRACT.md` is the tracked render of the ledger, so it is re-blessed (`BLESS_CONTRACT_DOC=1`); its diff is exactly the two names.

## The guard

`tests/it/ruling_prose_pins.rs` resolves every `path.rs::name` citation in a record's prose (`question` + `rationale`) against the file it names: the file must exist and declare an item of that name (`fn`/`mod`/`const`/…). A citation may legitimately name a **module** (`issue_390_hgvs_vcf_provider_coverage.rs::range_payload_never_reaches_alt` is a `mod`), so it resolves against every item a file declares, not the test set alone — reporting a real module as a missing test would be the false positive the issue warns about. Parsing reuses `common::rulings::split_guard_citation`, the same grammar `ruling_guard_field.rs` uses for `guard.tests`, so the two never drift.

**Scope, and why this form.** The two stale citations both use the file-anchored `path.rs::name` form, which admits a *precise* per-file check (the leaf must be defined in that file, not merely somewhere in the tree) and cannot be confused with ordinary prose. It composes with the two existing ledger guards and none subsumes another:

- `ruling_guard_field.rs` reads the structured `guard.tests`, not prose.
- `ruling_citation_currency.rs`'s `every_ruling_to_guard_citation_resolves` reads the bare `module::name` prose forms this file leaves alone, and is `#[ignore]`d pending #1881 — because one bare-`module::name` citation (`derivation-may-not-be-bounded-by-the-inputs-spelling`'s) names a guard whose rename **inverted the claim**, which is an adjudication rather than a citation repair. This change deliberately does **not** reach that citation (bare `module::`, not `path.rs::`), so activating this guard does not depend on that adjudication.

## Unchanged on purpose

The out-of-scope third broken citation (`derivation-may-not-be-bounded-by-the-inputs-spelling`, a bare `module::test` form whose test was renamed in a way that inverts the record's argument) is left untouched — repairing it is an operator adjudication tracked by #1881, not a stale-name fix.

## Verification

- The guard fails RED on the two records before the fix and passes GREEN after; sabotaging a real citation to a bogus name turns it RED, restoring turns it GREEN (non-vacuity).
- `cargo nextest run --features dev --no-fail-fast` — **11252 passed, 0 failed, 154 skipped**. `ruling_records_are_intact`, `clause_ruling_index::the_rendered_index_is_current`, `ruling_guard_field::*`, `ruling_citation_currency::*`, `normalization_contract_doc::the_published_document_is_current`, and the new `ruling_prose_pins::every_prose_pin_citation_resolves` all pass. The new tests run on PR CI (no `FERRO_MANIFEST` dependency).

Representation-Change: none. Ledger prose + a new test guard only; no watched `src/` file is touched and no normalized output moves.


nh13:
@coderabbitai pause

PR #2076, opened 2026-08-16
docs: name the real producer of the bare PositionRange variant

### LINKED ISSUE #2035: as_position_range's doc names the parser as the bare-variant producer; measurement says it is the con->delins rewrite
#2029 added `InsertedSequence::as_position_range` and, with it, the project's single canonical explanation of why the same payload has two spellings. That explanation names the wrong producer.

## What the doc says

`src/hgvs/edit.rs:413-415` attributes the bare `PositionRange` variant to the parser — that it "falls through to the bare variant for `parse_simple_count`".

## What is measured

No parseable input reaches it. Thirteen strings were probed through `parse_hgvs` — `delins`/`ins`, bracketed and bare, forward and `inv`, `*`-marker and `+5` offset, on `g.` and `c.` — and **every** parseable same-reference numeric range lands on `Complex([PositionRange | PositionRangeInv])`, because the `b'0'..=b'9' | b'*'` arm tries `parse_cds_position_range` first and that parser accepts every `digits_digits` pair.

So `parse_simple_count`'s two range returns (`src/hgvs/parser/edit.rs:992` and `:995`) are unreachable dead code.

The producer that actually mints the bare variant is `normalize::rules::canonicalize_conversion_to_delins` (`src/normalize/rules.rs:2640`) — the `con` → `delins` rewrite. Measured: `NaEdit::Conversion { source: "100_200" }` becomes `Delins { sequence: PositionRange { 100, 200 } }`. The doc does not mention it.

## Why this is worth fixing rather than leaving

#1986 asked for exactly this measurement, under "Not asserted here… the boundary needs measuring rather than reading". #2029 restated the parser story as fact instead. Because the accessor is now the one place the asymmetry is explained, a wrong attribution there propagates: someone touching `parse_simple_count` reads it, believes those returns are live, and preserves dead code — or someone extending the bare-variant handling goes looking for a parser input to test with and cannot find one, because none exists.

## Two adjacent items found in the same pass

**`parse_simple_count`'s range returns are dead.** `src/hgvs/parser/edit.rs:992,995`. Unreachable for the reason above. Removing them is a separate decision from fixing the doc — they may be worth keeping as defensive returns — but the doc should not describe them as the live path either way.

**`InsertedSequence::len()` still splits the two spellings**, twenty lines below the new accessor. `src/hgvs/edit.rs:448-464`: `Complex(_) => None` ("Would need to sum parts") against bare `PositionRange => Some(end - start + 1)`. So a `delins200_300` parsed from a string reports unknown length while the con-rewritten bare form of the same payload reports 101. No live impact — `len()` has no non-test caller under `src/` — and it is pre-existing rather than introduced by #2029. It is worth naming only because that PR's doc asserts the accessor is "the single place that collapses them", and within its own file it is not.

## Suggested fix

One sentence in the accessor doc: name `canonicalize_conversion_to_delins` as the producer of the bare variant, and say that the parser does not mint it. Optionally note the `len()` asymmetry as a known gap, or fix it.

Refs #1986, #2029.

nh13:
Closes #2035.

Representation-Change: none. Doc-comment-only change under `src/hgvs/`; no watched behaviour is touched.

## The defect

`InsertedSequence::as_position_range`'s doc comment attributed the bare `PositionRange` / `PositionRangeInv` variant to the **parser** — it claimed the parser "wraps `parse_cds_position_range`'s result in `Complex` but falls through to the bare variant for `parse_simple_count`". Because this accessor is the single canonical explanation of why one payload has two spellings, a wrong attribution there propagates: someone touching `parse_simple_count` reads it and believes those returns are live.

## What is measured

The parser mints **neither** bare variant for any parseable input. The `<digits>_<digits>` dispatch arm (`src/hgvs/parser/edit.rs:368`) tries `parse_cds_position_range` first, and that parser accepts every plain numeric pair, returning an `InsertedPart` the arm wraps in `Complex` (`:370-381`). `parse_simple_count`'s own `PositionRange` / `PositionRangeInv` returns (`:992`, `:995`) are only reached when `parse_cds_position_range` fails — which for a `digits_digits` input never happens — so they are dead code and no HGVS string produces the bare variant.

The bare variants are minted by **normalization**:

- bare `PositionRange` — `normalize::rules::canonicalize_conversion_to_delins` rewrites a same-reference `con` source (`NaEdit::Conversion { source: "100_200" }`) into `Delins { sequence: PositionRange { 100, 200 } }` (`src/normalize/rules.rs:2879`).
- bare `PositionRangeInv` — the `DNA/inversion.md:69` inverted-duplication `ins<range>inv` re-spell in `normalize_genome` mints it inside an `Insertion` (`src/normalize/mod.rs:6323`), currently on the `g.` axis only.

I confirmed these are the only two live producers by enumerating every `InsertedSequence::PositionRange { … }` / `PositionRangeInv { … }` construction site in `src/`; the remainder are the dead `parse_simple_count` returns or `#[cfg(test)]` code.

## The fix

Correct the accessor doc to state the parser mints neither bare variant, and to name both normalization producers. Comment-only.

## Verification

- `cargo nextest run --features dev` — `11245 passed, 154 skipped` (EXIT=0). The 154 skips are the bulk/manifest-gated fixtures, unrelated to this change.
- `cargo doc --no-deps --features dev --lib` with `-D rustdoc::broken_intra_doc_links` — no new broken-link errors introduced at the edited lines (the crate carries pre-existing baseline warnings elsewhere).

Refs #1986, #2029.

### DIFF
--- a/src/hgvs/edit.rs
+++ b/src/hgvs/edit.rs
@@ -406,16 +406,39 @@ impl InsertedSequence {
     /// `(start, end, inverted)` 1-based inclusive coordinates, or `None` when
     /// this is not one.
     ///
-    /// An all-numeric same-reference range has **two** AST spellings that the
-    /// parser mints interchangeably and that [`Display`](fmt::Display) renders
-    /// identically: the bare [`PositionRange`](InsertedSequence::PositionRange) /
+    /// An all-numeric same-reference range has **two** AST spellings that
+    /// [`Display`](fmt::Display) renders identically but that reach the AST from
+    /// two different producers: the bare
+    /// [`PositionRange`](InsertedSequence::PositionRange) /
     /// [`PositionRangeInv`](InsertedSequence::PositionRangeInv) variant, and the
-    /// single-part `Complex([PositionRange | PositionRangeInv])` shape (the
-    /// parser wraps `parse_cds_position_range`'s result in `Complex` but falls
-    /// through to the bare variant for `parse_simple_count`). Both denote one
-    /// contiguous slice of the same accession, so this is the single place that
-    /// collapses them — consumers call it rather than re-deriving the match, and
-    /// so cannot disagree about which spelling they handle (#1986).
+    /// single-part `Complex([PositionRange | PositionRangeInv])` shape. Both
+    /// denote one contiguous slice of the same accession, so this is the single
+    /// place that collapses them — consumers call it rather than re-deriving the
+    /// match, and so cannot disagree about which spelling they handle (#1986).
+    ///
+    /// The **parser never mints the bare variant.** For any input reaching the
+    /// `<digits>_<digits>` dispatch arm, `parse_cds_position_range` is tried
+    /// first and accepts every plain numeric pair, returning an
+    /// [`InsertedPart`] that the arm wraps in `Complex` — so a parsed same-
+    /// reference numeric range is always the single-part `Complex` shape.
+    /// `parse_simple_count`'s own `PositionRange` / `PositionRangeInv` returns
+    /// are therefore unreachable for any parseable input, and no HGVS string
+    /// produces either bare variant. Both are minted by **normalization**, not
+    /// the parser:
+    ///
+    /// - bare [`PositionRange`](InsertedSequence::PositionRange) —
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #2077, opened 2026-08-16
fix: catch a bare `none` decline that moves on a continuation line

### LINKED ISSUE #2031: The contradicted-decline tripwire still misses a bare `none`: #2027's fix depends on terminating punctuation
#2027 fixed the contradicted-decline tripwire only for a decline that carries terminating punctuation. For the bare `Representation-Change: none` that `CONTRIBUTING.md` documents, #1854 is still open — and the checker now disagrees with git-cliff, which is the #1555 class.

## Measured on `origin/main` at `89e3e426` (i.e. with #2027 merged)

`check(["src/normalize/merge.rs"], body)` where the body is a trailer followed by an unindented continuation line reading `150 rows of 85,642 move, all 3-prime.`:

| trailer value | result |
|---|---|
| `none.` | **refuses** — "This trailer declines and then describes a move" |
| `none` | **PASSES** |
| `no` | **PASSES** |
| `n/a` | **PASSES** |

So the disclosure never reaches the changelog, which is exactly what #1854 is about.

The message on the passing rows makes it worse: it reads *"Declared a representation change over 1 watched file(s)"*. The value is not merely failing the tripwire — it has stopped being recognised as a decline at all.

## Mechanism

`DECLINE_RE` (`scripts/check_representation_change.py`) is

```
^(?:n/a|na|no|none)[ \t]*(?:[.;:—–].*)?$
```

with `IGNORECASE | DOTALL` and **not** `MULTILINE` (verified: `DECLINE_RE.flags & re.MULTILINE` is 0).

Without `MULTILINE`, `$` matches only at end-of-string. With a terminator, the optional group's `.*` — `DOTALL`, so it crosses newlines — swallows the continuation and reaches `$`. Without a terminator there is nothing to consume the continuation, so `$` never matches and `declines()` returns `False`.

The punctuation dependence is therefore not a deliberate rule. It is an artefact of `.*` needing a terminator character to get started. `DECLINE_RE` was correct while the value was one line; #2027 made the value span lines, and the anchors silently became wrong.

Measured on the same build:

| value | `declines(full_value)` | `declines(first_line)` |
|---|---|---|
| `none.` + continuation | True | True |
| `none` + continuation | **False** | True |
| `no` + continuation | **False** | True |
| `n/a` + continuation | **False** | True |

## Parity with git-cliff is broken, and the parity test cannot see it

`release-plz.toml`'s footer rule is `MULTILINE`, so git-cliff reads a bare `none` with a continuation as a **decline**. The checker now reads it as a real change. That is the divergence `test_the_two_decline_rules_agree_value_by_value` exists to prevent — and it cannot observe this one, because both of the multi-line values it carries have terminating punctuation.

A parity test that cannot see the divergence it was written for is the same shape as the defect it guards against.

## Suggested fix: separate the two questions

They have different correct spans, and conflating them is what produced this.

- **Decline verdict → the first line.** Measured above: every form returns `True` on the first line, which is what git-cliff reports. Parity is then restored by construction rather than by matching case-by-case.
- **Contradiction tripwire → the whole value.** That is #1854's actual subject, and the only place #2027's widening is needed.

Adding `MULTILINE` to `DECLINE_RE` also makes the four rows pass, but it makes the reason capture ambiguous: `$` would match at every line end, so the optional reason group could anchor on a later line and the value's meaning would depend on where the regex chose to stop.

## Also worth deciding: where the extended value ends

`CONTINUATION_END_RE` stops at a column-0 trailer token, a Markdown heading, or the HTML comment GitHub appends — not at a blank line, list item, blockquote or fence. So prose well below the trailer can be pulled into the value.

Measured over 300 merged PRs: **0 verdict flips**. That is reassuring but not a structural zero — 233 carried a trailer, 40 had the value extended, 20 of those touch a watched path, and 3 already scavenge `rows … move`-adjacent prose, surviving only because the zero-exclusion and the punctuation happened to break the run.

Argument for leaving the bound wide: the errors are asymmetric. Over-reading produces a false *refusal* on an honest decline — loud, visible, the author edits and re-runs. Under-reading produces a false *pass* — silent, and the disclosure is lost. The second is #1854 itself. Narrowing to a blank line trades a recoverable error for an unrecoverable one, and it does not fix anything above: the punctuation gap survives any choice of where the value ends.

## Regression cases to add

Whatever bound is chosen, `test_the_two_decline_rules_agree_value_by_value` should carry the punctuation axis:

- `none` + unindented continuation
- `no` + unindented continuation
- `n/a` + unindented continuation
- `na` + unindented continuation

each asserted against the live `release-plz.toml` rule. Verified by mutation that a single such case fails the existing test with the repo's own #1555 message, so these are cheap and they do fire.

Refs #1854, #1555, #2027.

nh13:
Closes #2031.

## The gap

#2027 (#1854) made the contradicted-decline tripwire read the whole trailer value instead of just line 1, so `none.` followed by a `<n> rows move` continuation is refused. But that fix depends on **terminating punctuation**, and a bare `none` (no terminator) slips straight through.

Measured on the branch base:

| trailer + continuation-line `150 rows … move` | `check(["src/normalize/merge.rs"], …)` |
|---|---|
| `none.` | **refuses** — "declines and then describes a move" |
| `none` | **passes** — worse, read as a real change, not even a decline |
| `no` / `n/a` / `na` | **pass** |

So the disclosure never reaches the changelog — the exact direction this mechanism exists to prevent.

## Mechanism

`DECLINE_RE` is `IGNORECASE | DOTALL` and **not** `MULTILINE`, so `$` matches only at end-of-string. With a terminator its optional `(?:[.;:—–].*)?` group's `.*` (DOTALL) swallows the continuation and reaches `$`. Without a terminator there is nothing to consume the continuation, so `$` never matches and `declines(full_value)` returns `False` — which gates the movement tripwire, so it never fires.

`release-plz.toml`'s footer rule **is** `MULTILINE`, so git-cliff reads the same bare `none`+continuation as a **decline**. The checker read it as a real change. That is the #1555 class — the checker and the changelog disagree — and `test_the_two_decline_rules_agree_value_by_value` could not see it because both of its multi-line values carried a terminator.

## The fix

Separate the two questions, which have different correct spans:

- **Decline verdict → the first line.** `declines()` now reads the trailer's first line, which is exactly what git-cliff's `MULTILINE` rule reads. `DECLINE_RE` drops the now-unneeded `DOTALL`. Parity with git-cliff is restored **by construction** rather than only when a terminator is present.
- **Contradiction tripwire → the whole value.** Unchanged: `contradicted_decline` still scans `full_declaration_value` with `MOVEMENT_CLAIM_RE`.

## Unchanged on purpose (discriminators)

- honest bare `none`, `none.`, and `none, except two rows that merge` behave exactly as `CONTRIBUTING.md` / `release-plz.toml` specify (the last is filed as a real change — comma is not a terminator);
- a bare `none` whose continuation is an ordinary **reason** (no move) still passes as a clean decline;
- `none. 0 of 950 rows move.` (the documented quantified zero) still passes;
- a real bare move (`577 rows move, …`) still passes.

## Checker ⇄ audit still agree

`test_changelog_grouping.py` (including `test_the_audit_and_the_checker_agree_on_every_documented_form`) and `test_inject_representation_disclosure.py` pass unchanged — the checker stays in agreement with `check_changelog_grouping.py` on every documented form.

## Verification

- `ruff check scripts/ tests/python/` — EXIT 0
- `ruff format --check scripts/ tests/python/` — EXIT 0
- `mypy` (strict, project flags) on `scripts/check_representation_change.py` — EXIT 0
- `pytest tests/python/test_representation_change_trailer.py` — 177 passed
- `pytest tests/python/test_changelog_grouping.py tests/python/test_inject_representation_disclosure.py` — 51 passed

Representation-Change: none. Python tooling only; no watched file changes behaviour.

### DIFF
--- a/scripts/check_representation_change.py
+++ b/scripts/check_representation_change.py
@@ -200,23 +200,41 @@ DECLINE_TERMINATORS = ".;:—–"
 
 #: A declination: one of `NONE_VALUES`, alone or followed by a reason. Built from the two
 #: constants above so the vocabulary has one definition rather than a regex restating it.
-#: `DOTALL` so a reason spanning lines does not defeat the trailing `.*`.
+#:
+#: Matched against the trailer's FIRST LINE only (see `declines`), so it carries no `DOTALL`
+#: and no `MULTILINE`: the reason it captures is the remainder of that one line, and any
+#: continuation below is the contradiction tripwire's business, not the decline verdict's.
+#: This used to carry `DOTALL` and be matched against the whole value, which made the verdict
+#: depend on terminating punctuation — a bare `none` with a continuation had nothing to feed
+#: the trailing `.*`, so `$` never reached end-of-string and the value was not read as a
+#: decline at all (#2031). git-cliff reads the verdict off the trailer's own line, so this
+#: now agrees with it by construction rather than only when a terminator happens to be
+#: present -- pinned value-by-value by `test_the_two_decline_rules_agree_value_by_value`.
 DECLINE_RE = re.compile(
     r"^(?:"
     + "|".join(re.escape(value) for value in sorted(NONE_VALUES))
     + r")[ \t]*(?:["
     + re.escape(DECLINE_TERMINATORS)
     + r"].*)?$",
-    re.IGNORECASE | re.DOTALL,
+    re.IGNORECASE,
 )
 
 
 def declines(declaration: str) -> bool:
     """Return whether `declaration` is a decline rather than a description of a move.
 
-    The verdict is the first word; anything after a terminator is the reason for it.
+    The verdict is read from the FIRST LINE of the value, which is what git-cliff's footer
+    rule reads (it is `MULTILINE`, so its `$` anchors on the trailer's own line). Reading the
+    whole value instead depended on terminating punctuation to span a continuation, so a bare
+    `none` followed by a move on the next line was not read as a decline at all and the
+    contradiction tripwire — gated on this verdict — never fired (#2031). The tripwire still
+    scans the whole value; only the *verdict* is a first-line question, and separating the two
+    is what closes the gap #2027 left.
+
+    The verdict is the first word; anything after a terminator on that line is its reason.
     """
-    return DECLINE_RE.match(declaration.strip()) is not None
+    first_line = declaration.strip().partition("\n")[0]
+    return DECLINE_RE.match(first_line) is not None
 
 
 #: A quantified movement claim -- "3 rows move", "577 rows merge", "2 rows respell",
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

Issue #2015, opened 2026-08-15
964 `general.md:N` citations outside the ruling ledger are still pre-bump, across 135 files


nh13:
#1793's submodule bump left the same class of stale citation in the ledger's prose and outside it. #2009 covers the ledger and is fixed. **964 occurrences across 135 files** are not.

## The mechanism, in one paragraph

#1793 bumped `assets/hgvs-nomenclature` from `6f85311` to `565b973`. Upstream deleted a duplicated line at old `general.md:22`, so citations at old ≤21 are unaffected and **old ≥23 shift by −1**. `generate_spec_fixture` verifies every structured `clauses` entry's quote against the checkout and fails the build otherwise, so that register is correct. Nothing verifies a citation written in a Rust doc comment, a Markdown document, or an assertion message.

The clause that matters most moved `34` → `33`, and its new occupant argues the opposite way:

- **`:33`** — "two variants separated by one or more nucleotides should be described individually and **not** as a delins" (the rule that splits)
- **`:34`** — "**exception**: two variants separated by one nucleotide, together affecting one amino acid, should be described as a delins" (the exception that merges)

## Reproducible count

Measured on `origin/main`, excluding `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json` (fixed by #2009), the generated `docs/NORMALIZATION_CONTRACT.md` (a render of that ledger), and `tests/it/ledger_prose_clause_anchors.rs` (the new guard, which quotes the spellings deliberately):

```sh
FILES=$(grep -rIlE 'general\.md:[0-9]+' src tests examples docs README.md CLAUDE.md \
  | grep -v 'tests/fixtures/grammar/hgvs_spec_normalization' \
  | grep -v 'docs/NORMALIZATION_CONTRACT.md' \
  | grep -v 'ledger_prose_clause_anchors.rs')
echo "$FILES" | wc -l                                                # 135 files
grep -rIhoE 'general\.md:[0-9]+(-[0-9]+)?' $FILES | wc -l             # 964 occurrences
grep -rIhoE 'general\.md:[0-9]+(-[0-9]+)?' $FILES | sort | uniq -c | sort -rn | head
```

```
 320 general.md:34
 131 general.md:35
  99 general.md:56
  43 general.md:44
  41 general.md:58
  34 general.md:48
  23 general.md:96
  21 general.md:50
  18 general.md:43
  18 general.md:41
```

[...]

PR #2078, opened 2026-08-16
test: guard CLAUDE.md general.md anchors and re-point the stale ones

### LINKED ISSUE #2057: CLAUDE.md's general.md anchors are one line behind the pinned spec, and nothing checks them
`CLAUDE.md`'s clause anchors are uniformly one line behind the pinned spec submodule, and nothing checks them. The ledger's own citations are correct; only the prose drifted.

## The drift

The submodule is pinned at `565b9734`, where:

- **`general.md:33`** is the rule that **splits** — "two variants separated by one or more nucleotides should be described individually and **not** as a delins"
- **`general.md:34`** is the **exception that merges** — "two variants separated by one nucleotide, together affecting one amino acid, should be described as a delins"

The ruling ledger is on the correct numbering: 12 of its citations use `general.md:33` for the separation quote, and 2 use `:34` for the exception. `CLAUDE.md` uses `:34` for the separation rule throughout — the old `6f85311` numbering, from before #1793's submodule bump.

So a reader following a `CLAUDE.md` anchor lands on the clause that **argues the opposite way**. This is the same defect #2009/#2014 swept out of the ledger's prose; `CLAUDE.md` was not in that sweep.

## Where

At least six rows in the decided-rulings tables, plus prose at `:1012` and `:1321`. The `unchanged-is-read-over-every-minimal-alignment` row is one of them — noticed while reviewing #2054, which corrects a different error in that same row.

The sharpest instance: the paragraph headed **"Cite the clause exactly, and quote it"** offers `general.md:34` as its exemplar of the separation rule. The advice is right and its own example is wrong.

## Why nothing caught it

The ledger's citation quotes **are** checked against the spec checkout — `generate_spec_fixture` verifies every `clauses` entry and fails the build otherwise, which is why the ledger stayed correct through the bump. `CLAUDE.md`'s prose anchors are checked by nothing:

- `tests/it/claude_md_adjudication_tables.rs` compares record **ids** and table membership only (`ids_in_table`, `the_two_tables_partition_the_ledger_by_status`, `every_ledger_record_appears_exactly_once`).
- `ruling_citation_currency.rs` scans `.rs` files under `src`/`tests`, so a Markdown table at the repo root is out of scope by construction.

## Suggested fix

Two parts, and the second is the one that stops it recurring:

1. Re-point the anchors, verifying each against the pinned checkout rather than applying a blanket −1 — some citations legitimately mean `:34`, and a blind shift would break those. #2014 did exactly this for the ledger and is the model.
2. Extend a guard to cover `CLAUDE.md`. The cheapest version that would have caught this: extract `general\.md:\d+` from the file and assert the cited line still contains the phrase the surrounding sentence leans on — the same shape as the ledger's own clause check.

A related gap worth closing in the same pass: **no guard ties a `CLAUDE.md` row's prose to the ledger record it describes.** #2054 fixes a row that said "both cost-3 alignments" where the ledger says "three", and its claim that this "cannot silently diverge again" is not supported — the two can drift the moment either is edited.

Refs #2054, #2014, #1793.

nh13:
Closes #2057.

Representation-Change: none

## The defect

The #1793 spec bump (`6f85311` → `565b973`) deleted a duplicated line at old `general.md:22`, shifting every old line from 23 on down by one. #2009/#2014 swept the ruling ledger's citations against the pinned checkout, but the repository `CLAUDE.md` was not in that sweep. All 32 of its qualified `general.md:N` anchors named the line one past the clause they cite.

The sharpest case is the separation rule and its codon exception, which argue opposite ways:

- `general.md:33` — "two variants separated by one or more nucleotides should be described individually and **not** as a delins" (splits)
- `general.md:34` — "**exception**: two variants separated by one nucleotide, together affecting one amino acid, should be described as a delins" (merges)

So every stale `general.md:34` citation of the separation rule sent a reader to the neighbouring clause that says the reverse — including, pointedly, the paragraph headed "Cite the clause exactly, and quote it", whose own exemplar was wrong.

Nothing caught it: the ledger's structured `clauses` array is verified by `generate_spec_fixture`, and its prose by `ledger_prose_clause_anchors`, but `CLAUDE.md` was checked by neither. `ruling_citation_currency` scans `.rs` under `src`/`tests` only.

## The fix

**Part 1 — re-point the anchors.** Every qualified `general.md:N`/`general.md:N-M` anchor is shifted to the clause it cites, verified individually against the pinned checkout rather than blanket-shifted: each was old numbering, and none legitimately meant the current-numbering line (`general.md:34` for the codon exception, for instance, becomes the correct target of the two citations that mean the exception). The three bare `:N` shorthands that share a table row with a qualified anchor are fixed too, so no edited row is internally contradictory. A `--word-diff` confirms every changed token is a `general.md` citation or one of those three co-references.

**Part 2 — the guard.** `tests/it/claude_md_clause_anchors.rs` is the `CLAUDE.md` sibling of `ledger_prose_clause_anchors`. For each `general.md:N` citation whose sentence identifies the clause (via a narrow trigger phrase), it asserts the cited line holds that clause in the spec checkout. The expected line is *located from the spec on every run* via a signature phrase — nothing restates a line number — so the next bump moves the check and the expectation together instead of leaving a second stale copy. A weaker companion check requires every qualified citation to resolve to a non-blank, in-range line (this is the arm that also flags a citation left on a now-blank line, as stale `general.md:58` was). Both carry non-vacuity floors, mirroring the ledger guard's `ANCHORED_CITATIONS_FLOOR`.

The scanner helpers deliberately mirror `ledger_prose_clause_anchors` rather than sharing an extraction, keeping the new module self-contained so it cannot perturb that file's measured floors — the same self-contained style that guard itself uses. The header documents the choice and the classes it does not cover (bare shorthands, non-`general.md` files — #2015's wider territory).

## Unchanged on purpose

- Bare `:N` shorthands in rows with no qualified `general.md` anchor are left to #2015's broader sweep (964 occurrences across 135 files); this PR is the narrow `CLAUDE.md`-anchors-plus-a-guard slice #2057 asks for.
- No record ids, table membership, or ruling prose semantics change — only line numbers move, so `claude_md_adjudication_tables`, `ruling_citation_currency`, and `ledger_prose_clause_anchors` are unaffected (verified below).

## Representation stability

No `src/` code changes; `CLAUDE.md` and a new integration test only. No normalized output moves, so `Representation-Change: none`.

## Verification

- New guard is non-vacuous: pointing one fixed anchor at the wrong line turns it RED (`EXIT=100`), restoring it turns it GREEN (`EXIT=0`).
- Against the pre-fix file the guard flags 11 trigger-matched citations plus the now-blank `general.md:58`; against the fixed file it is green.
- `cargo nextest run --features dev --no-fail-fast` — 11264 passed, 0 failed (includes the new guard's 5 tests and `claude_md_adjudication_tables` / `ledger_prose_clause_anchors` / `ruling_citation_currency`).


nh13:
@coderabbitai pause

PR #2079, opened 2026-08-16
ci: baseline the nightly reference-aware xfail set

### LINKED ISSUE #1998: ci: the nightly reference-aware job reports success with 38 failing tests, and nothing baselines the xfail set
The nightly reference-aware job reports **`success`** while its test step fails. Run [31863843625](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/31863843625) (2026-08-15 04:12), step *Run reference-aware tests*:

```
Summary [ 246.058s] 292 tests run: 254 passed, 38 failed, 10828 skipped
```

The step's API `conclusion` is `success`, because `continue-on-error: true` (`nightly-mutalyzer.yml:215`) **rewrites** it. The workflow says so itself at `:271` and `:277` — "Keyed on `outcome`, not `conclusion`". So neither the run list, the job status, nor a `--json jobs` query on `conclusion` can see this. You have to read the log.

## The 38, by module

| module | failures |
|---|---:|
| `issue_1086_c_axis_genomic_context` | 13 |
| `mutalyzer_normalize_tests` | 10 |
| `issue_1086_cross_axis_projection` | 3 |
| `issue_498_whole_exon_deletion` | 2 |
| `issue_1713_bare_transcript_genomic_decline` | 2 |
| `real_data_normalization_tests` | 1 |
| `protein_cis_delins_decomposition` | 1 |
| `multi_member_cis_axis` | 1 |
| `issue_1539_split_member_separation` | 1 |
| `inversion_sweep` | 1 |
| `hgvs_rs_projection_tests` | 1 |
| `delins_equal_vs_unequal_length_discriminator` | 1 |
| `biocommons_normalize_tests` | 1 |

## Why this is filed as a gap rather than as 38 bugs

Non-gating is **deliberate** and well argued: the job's purpose is "to surface drift in the xfail report" (`:209`), and the xfail artifact is "the only channel a non-gating [job has]" (`:81`). Fine.

But **nothing compares that artifact against a baseline.** There is no expected-failure list, no count pinned anywhere, and no step that diffs this run's failure set against the previous one. So:

- a **new** failure is indistinguishable from the 37 beside it;
- the job reports `success` either way;
- "surface drift" is only true if a human opens the artifact and remembers what was in yesterday's.

Drift is exactly what a baseline detects and a bare list does not. That is the gap.

## The instance that surfaced it

`multi_member_cis_axis::axis_normalized` — measures `unwindowed: 89` against a pinned `90`. Reproduced locally against `c9207d7e` with an unrelated test-only change **fully reverted**, so it is not from #1997, and the census is a pure function of `src/`.

It is one of the 38 above. I only found it because a local run happened to have `FERRO_MANIFEST` armed — PR CI never provisions one, so on a pull request this test early-returns and reports green (`ci.yml:1359` announces exactly that as a `::notice::`).

## What would close this

Some baseline the run can be compared against — a committed expected-failure list, or a count the step asserts, or a diff against the previous run's artifact — so that **growth** in the set is detectable without a human remembering. Whether the right bar is "no new failures" or "the set matches a committed list" is a judgement call; either beats none.

Deliberately **not** proposing that the job start gating. That trade was made on purpose and this issue does not reopen it.

### LINKED ISSUE #2082: ci: the nightly xfail gate cannot go red — its exit code is swallowed by `tee`
`nightly-mutalyzer.yml`'s `Diff FAIL set against xfail baseline` step ends in:

```sh
python3 scripts/diff_nightly_xfail_baseline.py \
  --junit "$JUNIT" --baseline "$BASELINE" | tee -a "$GITHUB_STEP_SUMMARY"
```

with a comment asserting that `set -o pipefail` is GitHub's bash default and propagates the script's exit code through the `tee`.

It is not. The step names no `shell:` and the workflow sets no `defaults.run.shell`, so it runs under `bash -e {0}`. `pipefail` is added only by an explicit `shell: bash`, which expands to `bash --noprofile --norc -eo pipefail {0}`.

## Reproduction

The step's script verbatim, over a synthetic JUnit holding the 38 baselined failures plus one new one:

| shell | script output | step exit |
|---|---|---|
| `bash -e {0}` (what this step gets) | `### NEW FAILURE (1)` … `brand_new_module::a_regression` | **0 — GREEN** |
| `bash --noprofile --norc -eo pipefail {0}` | same | **1 — RED** |
| `set -euo pipefail` prepended to the run block | same | **1 — RED** |
| `set -euo pipefail`, exact-match JUnit (no drift) | `Exact match — no drift.` | **0 — GREEN** (no false red) |

So as shipped the diff step prints `NEW FAILURE`, exits 0, the job concludes `success`, `report-failure` (`if: failure()`) never fires, and no issue opens.

This is the failure mode of #1998 reproduced inside the fix for #1998: a step that fails while the job reports green.

`actionlint` exits 0 on the file either way. Every other `run:` block in this repo that needs pipefail sets it explicitly — `report-failure.yml:109` (invoked by this very workflow), `nightly-perf.yml:138,157`, `nightly-soak.yml:252`, `ci.yml:576`, `release-binaries.yml:129`. This step is the only one relying on the shell default, so the convention exists and this is a slip from it.

## Fix

Two parts, and the second is the point:

1. `set -euo pipefail` as the first line of the run block (or `shell: bash` on the step). Also assert the JUnit exists rather than `if [ -f ]`-and-continue, so a nextest crash before the report is written is red rather than an exit-2 that is also swallowed.
2. A repo-level guard so this class cannot recur: a convention test (or actionlint/shellcheck rule) forbidding a pipe as the last command of a gating `run:` block, or requiring an explicit `pipefail`/`shell: bash` on any step whose failure is meant to fail the job. Without this, the next gating step written the same way is green in the same silent way.

nh13:
Closes #1998. Closes #2082.

## The gap

The nightly reference-aware job (`nightly-mutalyzer.yml`) runs every manifest-backed conformance guard against a real ferro-prepared reference. ~38 of those tests are known-failing (live divergences from mutalyzer / biocommons / hgvs-rs, plus tracked ferro bugs), and the job is deliberately **non-gating on the failures themselves** — the test step carries `continue-on-error: true`, which rewrites the step *conclusion* to `success`.

But **nothing compared that FAIL set against a baseline.** There was no expected-failure list, no pinned count, and no diff against the previous run. So a **new** (39th) failure was indistinguishable from the 38 beside it, the job reported `success` either way, and "surface drift" was only true if a human opened the xfail artifact and remembered yesterday's. That is the gap #1998 records — and it is not hypothetical: `multi_member_cis_axis::axis_normalized` was red for days before anyone noticed, because a red step reads as a green job here.

## The fix — a committed baseline the run is diffed against

This does **not** reopen the gating decision (the nightly still does not gate PRs, and does not run on them). It makes **drift in the set of failing tests** the job's signal, instead of the `continue-on-error` conclusion.

- **`.config/nextest.toml`** gains a `nightly-xfail` profile that emits **JUnit** (`target/nextest/nightly-xfail/junit.xml`). JUnit is used deliberately over `--message-format libtest-json`, which is still experimental in nextest 0.9.133; the JUnit schema is stable and stdlib-parseable. The profile only *adds* a report to the inherited `[profile.default]` settings.
- **`tests/fixtures/nightly-reference-aware/xfail-baseline.txt`** is the committed set of known-failing nextest test ids — one `{binary}::{module}::{test}` per line, 38 total. `README.md` documents it as a non-hermetic snapshot, how to regenerate it, and what it does **not** cover.
- **`scripts/diff_nightly_xfail_baseline.py`** parses the JUnit FAIL set and diffs it against the baseline:
  - **NEW failure** (failing now, not baselined) → exit 1 (red)
  - **UNEXPECTED PASS** — any baselined test that stops failing, i.e. `baseline - failed` (it passed, was skipped/`#[ignore]`d, or vanished) → exit 1 (red), so the baseline cannot silently rot
  - **exact match** → exit 0 (green)

  Regenerating after a deliberate change is `--update-baseline`.
- A new **`Diff FAIL set against xfail baseline`** step (no `continue-on-error`) runs that diff, so **drift** — not the absorbed step conclusion — turns the nightly red and lets `report-failure` open a tracking issue. The `report-failure` hint is updated to cover the drift case.

The 38 underlying failing tests are **left as-is** — fixing them is out of scope; this only makes a change in the set visible.

## What the baseline is, and is not: it is a set of TEST IDS

The unit of the baseline is a nextest test id. That is the right granularity for what #1998 asked for — a 39th failing test is named, and a baselined test that goes green is named — and it is the wrong granularity for anything happening *inside* one of the 38.

That remainder is not small, because the baselined tests are the highest-fan-in ones in the selection: each comparator axis runs a whole corpus and reports one verdict. Measured on nightly run `31926234266`, one baselined axis carries **30 pass / 57 FAIL** over its own inputs. If it moved to **0 pass / 87 FAIL**, the id set would be byte-identical and this gate would print `Exact match — no drift.` So 38 of the 292 selected tests are exempt from per-input drift, and they are the 38 with the most inputs behind them.

The per-input data is not lost, only undiffed: each axis's `AxisTally::finish` already writes its FAIL inputs to `/tmp/ferro-xfail/<axis>.txt` and `.tsv`, which the nightly uploads as the `ferro-xfail-*` artifact. A 30/57 → 0/87 movement is in the run and in that artifact; nothing compares it against anything — the same shape as the gap #1998 recorded one level up. Closing that means a second baseline per axis, keyed on an input string rather than a test id, plus a policy for what a corpus refresh does to it; deliberately not folded in here. The limitation is recorded in `tests/fixtures/nightly-reference-aware/README.md` and in the workflow comment rather than left to be re-derived.

## #2082 — the gate could not go red, and the class it belongs to

The diff step as first written ended in `… | tee -a "$GITHUB_STEP_SUMMARY"`, with a comment asserting that `set -o pipefail` is GitHub's bash default. It is not: a step naming no `shell:` runs under `bash -e {0}`, and `pipefail` arrives only with an explicit `shell: bash`. Under `-e`, a pipeline's exit status is its **last** command's — so `tee` reported 0 while the diff exited 1.

Reproduced with the step's script verbatim over a synthetic JUnit holding the 38 baselined failures plus one new one, and re-verified after the fix:

| shell | JUnit | script output | step exit |
|---|---|---|---|
| `bash -e {0}` — pre-fix | drift | `### NEW FAILURE (1)` | **0 — GREEN** |
| `bash -e {0}` — post-fix | drift | `### NEW FAILURE (1)` | **1 — RED** |
| `bash -eo pipefail {0}` — post-fix | drift | same | **1 — RED** |
| `bash -e {0}` — post-fix | exact match | `Exact match — no drift.` | **0 — GREEN** (no false red) |
| `bash -e {0}` — post-fix | **absent** | `::error::…wrote no JUnit report` | **1 — RED** |

Two fixes, and the second is the point:

1. **The step.** `set -euo pipefail` as its first line, and a missing JUnit report now **fails** rather than the previous `if [ -f ]`-and-continue — so a nextest crash before the report is written is red, instead of an exit-2 the `tee` swallowed just as thoroughly. The false comment is gone rather than corrected.
2. **`tests/it/workflow_pipefail_gate.rs`** — a repo-level guard so the class cannot recur. It derives every `run:` block from `.github/workflows/` (117 today, 116 bash-family) and fails when one contains a pipeline without `pipefail` — from `shell: bash`, from `defaults.run.shell`, or from an in-block `set`. `actionlint` and `zizmor` both exit 0 on the pre-fix file, so nothing existing could see this.

   It reads *order*, not presence: a `set -o pipefail` written after a pipeline does not protect it, and a block-level boolean cannot tell the two apart — which is the same shape of wrong as the defect itself, a check that looks satisfied by text that does nothing.

   **The rule is deliberately broad**: every pipeline, not only a pipe as the last command, and not only in a step without `continue-on-error`. Each narrower rule keys the guard on a fact a *later* PR can change without re-reading the pipeline it exempted — deleting a `continue-on-error` promotes a step to a gate in a diff that never mentions the pipe. The escape hatch is explicit instead: a `# pipefail-exempt: <reason>` comment at the site, which fails without a reason and fails once its block no longer contains a pipeline. Nothing is exempt today. The file's module docs carry the full argument and the guard's own blind spots.

   It found **four** pre-existing violations beside the nightly step, all fixed here: two `curl … | sh` installers in `deploy-local.yml` (a failed download pipes an empty script into a shell that exits 0), the UTA container probe, and `lint-workflows.yml`'s `sha256sum -c` check.

   Proved able to fail: on the pre-fix tree it reddens naming `.github/workflows/nightly-mutalyzer.yml :: job reference-aware-tests :: step Diff FAIL set against xfail baseline` and the four others, at `EXIT=100`; on the fixed tree, `15 tests run: 15 passed`. Ten of those fifteen are the scanner's own tests, including the one that matters most — a comment containing the literal text `set -o pipefail` must not count as supplying it, which is exactly what the pre-fix block carried and exactly what a naive grep-based guard would have believed.

## Validating the committed baseline in PR CI

Nothing read this file on a pull request: the nightly is its only consumer, and it runs on a schedule against a reference no PR has. So a broken baseline was invisible until the next night, on a job that opens an issue when it fails.

`read_baseline` is now strict — an inline comment, a commented-out id and a duplicate all raise rather than silently shrinking the set, and a smaller baseline is a weaker gate (an id that drops out stops being expected-to-fail, so when it stops failing nothing reports it). `--update-baseline` round-trips what it writes through that parser before writing it. `tests/python/test_diff_nightly_xfail_baseline.py` (17 → 37 tests) asserts the committed file parses, is non-empty, holds only well-formed ids, and is **byte-identical to what `--update-baseline` renders** — one property covering sorted, unique, no blanks and no comments, so a hand-edit is visible on the PR rather than on the next night's run.

## Unchanged on purpose

- The nightly stays non-gating on the individual conformance failures; only a *change* to the known set fails the job.
- `REFERENCE_AWARE_SELECTION` and the four armed oracles are untouched.

## Verification

- **The baseline is the real nightly's FAIL set**, extracted from the most recent scheduled run (`31926234266`, `headSha e4c53fb2`, 2026-08-16): `292 tests run: 254 passed, 38 failed`, exactly 38 distinct FAILED names, set-equal to the committed file with `[]` on both difference directions. Not a PR-CI artifact and not empty.
- **The diff mechanism is unit-tested** (37 tests): new-failure → red, unexpected-pass → red, now-skipped → red, exact-match → green, `--update-baseline` rewrites and round-trips, missing JUnit → exit 2, and the six baseline-strictness shapes above.
- **The gate can go red**, which is the claim `Closes #1998` rests on: with the `set -euo pipefail` line deleted as a control, the step prints `### NEW FAILURE (1)` and exits **0** — #1998's own failure mode; with it restored, the same input exits **1**.
- **End-to-end locally**: ran `REFERENCE_AWARE_SELECTION` under `--profile nightly-xfail` with `FERRO_MANIFEST` armed (the four oracles on), confirmed JUnit is produced and the parser's ids match the nightly's format exactly, then ran the script over the real JUnit against the committed baseline — **0 new failures**. (The local box's curated manifest is a subset of the nightly's full `--ensembl` prepare, so 37 baselined tests pass locally that the nightly fails; the diff correctly reports that environment difference. The green-match case can only be confirmed against the true nightly environment.)
- `actionlint`, `zizmor`, `ruff check`, `ruff format --check` all green; `spec_fixture_setup_filter` (which parses `.config/nextest.toml`) still passes.

**Validation note:** the baseline is seeded from a nightly whose sha is behind current `main`, so the first live nightly after merge may report a small drift and require a one-time `--update-baseline` re-bless. That is the mechanism working as designed for a non-hermetic baseline, and the first nightly is where the green-match is finally confirmed.

Representation-Change: none. CI/tooling only; no watched source directory is touched.


nh13:
@coderabbitai pause

PR #2086, opened 2026-08-16
fix(equivalence): apply_triples fetch failure is Indeterminate, not decided NotEquivalent

### LINKED ISSUE #2075: equivalence: apply_triples failing after a successful fetch is read as a decided NotEquivalent
`compare_triples` fetches the union window, applies both triple sets, and matches on the pair:

```rust
match (
    apply_triples(&reference, win_start, triples1),
    apply_triples(&reference, win_start, triples2),
) {
    (Some(a), Some(b)) if a.eq_ignore_ascii_case(&b) => SequenceVerdict::Same,
    (Some(_), Some(_)) => SequenceVerdict::Different,
    _ => SequenceVerdict::Declined,
}
```

That final catch-all is reached when the fetch **succeeded** and `apply_triples` still returned `None` — most commonly because a description's *stated* reference base contradicts the base the reference actually carries. That is a reference-level obstacle, and `Declined` is read by the ladder as a decided `NotEquivalent`.

It is the last route by which a reference-level obstacle becomes a decided negative. The two siblings are closed: #2053 closed the window fetch, and #2069 closes the per-member `hgvs_to_spdi` conversion inside `edit_triples`.

## Measured

On #2069's head. `SyntheticBuilder::genomic("ACGTACGTACGT")` pads its core with 256 `ACGT` bases on each side, so `NC_TEST.1` is **524** bases and `g.20` / `g.21` are served positions, not out-of-bounds ones. Position 20 reads `T` and position 21 reads `A`, so both stated reference bases are contradicted:

| pair | provider | verdict |
|---|---|---|
| `NC_TEST.1:g.20A>G` vs `g.21C>T` | padded 524-base contig (both positions in bounds) | **`NotEquivalent`** |
| `NC_SMALL.1:g.20A>G` vs `g.21C>T` | genuinely 12-base contig (both positions out of bounds) | `Indeterminate` |

Both SPDIs convert without touching the provider — a substitution states its own bases — so neither pair takes the `edit_triples` route:

```
SPDI_A Ok(SpdiVariant { sequence: "NC_TEST.1", position: 19, deletion: "A", insertion: "G" })
SPDI_B Ok(SpdiVariant { sequence: "NC_TEST.1", position: 20, deletion: "C", insertion: "T" })
```

The second row is #2053's window-fetch route and already answers `Indeterminate`. The first row is this issue: the fetch succeeds, so the only thing left to decline is the apply.

## Full route enumeration

Every route by which a reference-level obstacle reaches a verdict, measured on #2069's head:

| route | verdict | status |
|---|---|---|
| `edit_triples` per-member `hgvs_to_spdi` | `Indeterminate` | closed by #2069 |
| `fetch_window` in `compare_triples` | `Indeterminate` | closed by #2053 |
| **`apply_triples` failing after a successful fetch — `compare_triples`'s `_ => SequenceVerdict::Declined`** | **`NotEquivalent`** | **this issue** |
| second-axis transcript lookup | `SequenceMatch` — a weaker positive, never a decided negative | deliberate, documented, pinned |
| `Normalizer::normalize` failure | `Err(FerroError)`, not a verdict | out of scope |

## Why this is not folded into #2056

#2056 names the `edit_triples` `.ok()?`, and #2069 closes exactly that, so `Closes #2056` on that PR is correct. This is a different site with a different mechanism, and it survives #2069 unchanged. #2056 carries a correction comment with the same route table.

## Suggested fix

The catch-all is doing two jobs. Separate "the reference contradicts what the description states about it" — which is want of usable reference data and belongs with `SequenceVerdict::ReferenceUnavailable` → `Indeterminate` — from any genuinely representational reason `apply_triples` can decline. The same split #2053 and #2069 each made at their own rung.

Note the catch-all also swallows *which* side failed. Distinguishing `(None, Some(_))` from `(None, None)` may be worth it: one side being unappliable while the other applies is a different fact from neither applying.

## Note for anyone writing a test here

`SyntheticBuilder::genomic` pads to 524 bases. A comment reasoning about "base 2" or "base 6" of a 12-base core is describing a position in the **pad** — accidentally true only because the pad repeats the same period-4 motif. That padding is what made this row read as out-of-bounds when it is not.

Refs #1989, #2053, #2056, #2069.

nh13:
Closes #2075

Representation-Change: none. `src/spdi/apply.rs` changes are behaviour-preserving — `apply_triples` still returns `None` for exactly the same inputs (it is now a `.ok()` wrapper over the new classified function), so no normalized output moves. The `src/equivalence/` change only reclassifies an equivalence verdict, which is not a normalization output.

`compare_triples` fetches the union window, applies both SPDI triple sets, and matches on the pair. Its catch-all arm — reached when the fetch **succeeded** but `apply_triples` still returned `None`, most commonly because a description's stated reference base contradicts the base the reference actually carries — fell through to the ladder's `NotEquivalent` tail. That asserted a decided negative about a pair whose resulting sequences were never reconstructed: a reference-level obstacle read as a decided verdict.

This is the last of the three reference-level routes that reached a decided negative; its two siblings are already closed — #2053 (#1989: the window fetch failing) and #2069 (#2056: a member's own HGVS→SPDI conversion failing) — and this mirrors both.

The fix distinguishes *why* `apply_triples` declined. `src/spdi/apply.rs` gains `ApplyDecline { ReferenceMismatch, Unappliable }` and an `apply_triples_classified` returning `Result<String, ApplyDecline>`; `apply_triples` becomes a thin `.ok()` wrapper so its existing callers are unchanged. `compare_triples` then routes a `ReferenceMismatch` on either side to `SequenceVerdict::ReferenceUnavailable` → `Indeterminate`, while a genuinely un-spliceable set of triples (`Unappliable` — an overlapping cis allele, or the defensive out-of-window guard) stays `Declined` → `NotEquivalent`, preserving the deliberate negative pinned by `issue_1244_equivalence_overlap_panic`. `ReferenceMismatch` takes precedence over `Unappliable` because a difference cannot be asserted against a sequence that could not be built.

Pinned by the new `tests/it/issue_2075_apply_triples_reference_mismatch.rs` (the measured `g.20A>G` vs `g.21C>T` reproducer on the served-but-contradicting `NC_TEST.1` contig, plus a one-sided-contradiction case and both discriminating controls). Three in-source unit tests (`test_different_variants`, `test_substitution_at_different_positions`, `test_differing_alleles_do_not_error`) stated reference bases that contradicted the served `NM_000088.3` sequence and only passed via the old buggy route; they are updated to faithful bases so they genuinely reconstruct and compare — the same correction #2053 made to `test_genomic_variants_different`.

Refs #1989, #2053, #2056, #2069.

### DIFF
--- a/src/equivalence/checker.rs
+++ b/src/equivalence/checker.rs
@@ -13,6 +13,7 @@ use crate::normalize::{NormalizeConfig, Normalizer};
 use crate::reference::transcript::Strand;
 use crate::reference::ReferenceProvider;
 use crate::sequence::reverse_complement;
+use crate::spdi::apply::{apply_triples_classified, ApplyDecline};
 use crate::spdi::{hgvs_to_spdi, ConversionError, SpdiVariant};
 
 /// Largest reference window (in bases) a sequence-level equivalence rung will
@@ -526,7 +527,8 @@ impl<P: ReferenceProvider> EquivalenceChecker<P> {
         // (unsupported edit, missing reference, mixed accessions, or a
         // multi-molecule allele) simply declines, so the rung only ever
         // upgrades a `NotEquivalent` verdict.
-        match self.same_resulting_sequence(v1, v2) {
+        let verdict = self.same_resulting_sequence(v1, v2);
+        match verdict {
             SequenceVerdict::Same => {
                 let (level, note) = self.strengthen_across_axes(v1, v2);
                 return Ok(EquivalenceResult::new(level)
@@ -545,17 +547,20 @@ impl<P: ReferenceProvider> EquivalenceChecker<P> {
                     )));
             }
             // The reference needed to compare the pair could not be read, from
-            // either of two points: the window fetch, after both sides converted
-            // to SPDI on a shared accession (#1989/#2053), or a member's own
+            // one of three points: the window fetch, after both sides converted
+            // to SPDI on a shared accession (#1989/#2053); a member's own
             // HGVS→SPDI conversion, which the ordinary short forms must run
-            // against the provider (#2056). Either way — as with `WindowTooWide`
-            // — nothing was reconstructed and nothing was compared. This is the
-            // "want of reference data" decline, and it must report
-            // `Indeterminate` for the same reason its sibling does: falling
-            // through to the `NotEquivalent` tail would assert a decided negative
-            // about a pair that was never examined. Pinned by
-            // `issue_1989_declined_is_indeterminate` and
-            // `issue_2056_edit_triples_reference_failure`.
+            // against the provider (#2056); or the apply itself, when the fetch
+            // succeeded but a stated reference base contradicts the served
+            // reference so the edit could not be reconstructed (#2075). Either
+            // way — as with `WindowTooWide` — nothing was reconstructed and
+            // nothing was compared. This is the "want of reference data"
+            // decline, and it must report `Indeterminate` for the same reason
+            // its sibling does: falling through to the `NotEquivalent` tail
+            // would assert a decided negative about a pair that was never
+            // examined. Pinned by `issue_1989_declined_is_indeterminate`,
+            // `issue_2056_edit_triples_reference_failure` and
+            // `issue_2075_apply_triples_reference_mismatch`.
             SequenceVerdict::ReferenceUnavailable => {
                 return Ok(EquivalenceResult::new(EquivalenceLevel::Indeterminate)
                     .with_normalized(str1, str2)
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

nh13:
Independent review (read-only, in a throwaway worktree; nothing pushed). **The bug is real and the fix is right** — I confirmed both directions of the ladder are genuinely pinned: reverting `ReferenceMismatch` to `SequenceVerdict::Declined`, and separately misclassifying the ref-base mismatch as `Unappliable` in `apply_triples_classified`, each redden the same two `issue_2075::` tests from two independent sites.

I also confirmed the three edited in-source unit tests are **not** relabelled to pass — mutating `(Ok(_), Ok(_)) => Same` reddens all three, proving they now reach the reconstructed comparison rung rather than short-circuiting through the catch-all decline. That is a real coverage gain.

Three things to fix before merge. Full suite merged onto current main is green (`11321 passed, 152 skipped`), so none of these is a correctness break.

---

**1. The body claims a property no test holds, and credits a test that cannot hold it.**

The body says the `Unappliable` route "stays `Declined` → `NotEquivalent`, preserving the deliberate negative pinned by `issue_1244_equivalence_overlap_panic`." Mapping `Unappliable` → `ReferenceUnavailable` leaves **all five** `issue_1244_*` tests green. A `panic!()` in that arm run against the whole suite gives **11,282 tests, zero panics** — the arm is unreachable; the overlapping-cis-allele case declines earlier in `edit_triples`. Swapping the arm precedence (the body's explicit *"`ReferenceMismatch` takes precedence"* rule) is also **green**. Tracked as #2104; the body's attribution should be dropped either way.

**2. The PR falsifies the comment that justified keeping two decline vocabularies apart.**

`SpliceFailure { Overlapping, StatedBasesMismatch }` (`apply.rs:1028`) and the new `ApplyDecline { ReferenceMismatch, Unappliable }` (`apply.rs:472`) are the same two-way distinction with the same derives, 570 lines apart in one file. `splice_denoted_sequence`'s doc comment at `:1057-1063` justifies its separate existence precisely thus: *"it answers a question `apply_triples` cannot ... and `apply_triples` collapses both into `None`."* After this PR `apply_triples_classified` answers exactly that question, so the paragraph is now false. Either reuse `SpliceFailure`, or update the comment to say why two vocabularies remain.

**3. It flips a public API landed yesterday, unannounced.**

`group_by_cross_axis` does not mention `apply_triples`, but reaches it: `compare_denotations` → `same_resulting_sequence` → `compare_triples`, then branches on `is_decided()` at `checker.rs:1452`. Measured on a two-element ref-mismatched corpus (`CrossAxisSequenceMatch`):

```
main         : undecided_pairs=0  is_complete=true
main + #2086 : undecided_pairs=1  is_complete=false
```

So `ConfluenceReport::is_complete()` flips to `false` for any corpus containing a ref-mismatched description — common in real ClinVar/legacy data. #2068's docs say `is_confluent` "is a statement about the whole corpus only when `is_complete` is true", so this changes what that report means. **I judge the new behaviour correct** — a pair whose sequences could not be reconstructed genuinely was not examined — but the trailer reads "the `src/equivalence/` change only reclassifies an equivalence verdict", and this is more than that. Please pin it in `confluence_self_check_tests.rs` and mention it in the body.

---

For the record, one concern raised against this PR earlier **is not a problem** and I re-measured it: `ApplyDecline` vs #2068's `DeclineSite` are not a duplicated vocabulary. `DeclineSite` is `pub`, `#[non_exhaustive]`, six variants, rendered by `TripleDecline::into_reason`; `ApplyDecline` is `pub(crate)`, two variants, renders nothing, collapses into `SequenceVerdict`. No overlap, and the exhaustive match is sound precisely because `ApplyDecline` is deliberately not `#[non_exhaustive]`.

Issue #2064, opened 2026-08-16
LegacyTranscript still carries a derived sequence_length, but removing it bumps the blessed reference identity


nh13:
#2058 removed `sequence_length` from the benchmark supplemental sidecar, on the grounds that a persisted "derived" length is a hollow field: it is 0 for every shipped row, nothing reads it, and the real length comes from the FASTA index. `LegacyTranscript` (`src/prepare/mod.rs:221`) carries the same field with the same derived/sourced confusion, and #2058 deliberately did not touch it.

## Why it was left out, and why that was right

`legacy_transcripts_metadata` and `legacy_genbank_metadata` **are** content-stamped (`src/prepare/identity.rs`), unlike `patterns_transcripts.metadata.json`. So removing the field there would change those artifacts' content stamps and **bump the blessed reference identity `aa8b3246d83055cc`**, which is pinned in `tests/fixtures/mutalyzer-normalize/empty-projection/*.count`.

That makes it a different kind of change from #2058 — which moves no identity at all, precisely because the sidecar it touches is not stamped. Bundling them would have turned a no-identity-impact refactor into one requiring a re-bless and a re-prepare on both volumes.

## What finishing the job would involve

1. Remove `sequence_length` from `LegacyTranscript`, mirroring #2058's shape (the field becomes unrepresentable rather than detectable-and-repairable).
2. Confirm no consumer reads it — #2058's precedent is that the only reader was the predicate that classified it.
3. Re-bless the reference identity and update the pinned `.count` fixtures.
4. Re-prepare, or confirm the identity change is expected on both scratch and backup volumes.

Step 3 is the reason this is its own issue: the identity is deliberately excluded from `prepared_at` (#933) so the two volumes do not split, and derived artifacts are content-stamped (#905) so an in-place content change bumps the identity even when every basename is unchanged.

## Worth deciding first

Whether the field is genuinely hollow in the legacy sidecars the way it was in the supplemental one. #2058 established that empirically — 116,572 of 140,027 rows carried `sequence_length: 0` and no typed consumer read the key. The same measurement should be made against the legacy artifacts before assuming the analogy holds; they may have a real reader, in which case the answer is to document the field as sourced rather than to delete it.

Refs #2058.

PR #2087, opened 2026-08-16
refactor(prepare)!: split legacy transcript metadata into sourced authority + on-disk record

### LINKED ISSUE #2085: Reference identity should hash authoritative inputs only; derived fields should not be persisted or stamped
## Root cause

The prepared-reference data model conflates **sourced** fields (the authority) with **derived** fields (recomputable from another artifact), then *persists* the derived ones, and then *content-stamps* them into the blessed reference identity (`aa8b3246d83055cc`). Three consequences, all real and recurring:

1. **Derived values can go stale against their true authority.** A persisted `sequence_length` is a second copy of the FASTA/`.fai` length; a persisted CDS coordinate can drift from cdot; the two-coordinate-space ambiguity behind #1737/#1759 is the same disease in a different artifact — a derived quantity stored beside its authority and free to disagree with it.

2. **A derived value inside a content-stamped artifact makes the blessed identity depend on redundant data.** Because derived artifacts are FNV-content-stamped into the identity (#905), *removing a field nothing reads* churns the identity for zero semantic change. That is exactly what makes #2064 expensive: deleting a hollow, unread `sequence_length` bumps `aa8b3246d83055cc` and forces a re-bless + re-prepare, even though no consumer's behavior changes.

3. **Sourced vs derived is indistinguishable by looking.** The only current remedy is doc-comments (documentary, not structural). Nothing stops the next field from being persisted-and-stamped the same way, and nothing tells a reader which fields are authoritative.

## Why the class recurs

Because the fix is applied per-field, and the identity model makes each per-field fix expensive, they get deferred. #2058 could delete a hollow length cheaply *only because* the sidecar it touched is not content-stamped; the identical field on `LegacyTranscript` (#2064) cannot be deleted cheaply *only because* `legacy_transcripts_metadata` / `legacy_genbank_metadata` are content-stamped (`src/prepare/identity.rs`). So the same one-line data-model defect is a free no-op in one artifact and a re-bless-triggering event in another, purely because of where the identity boundary happens to fall. Piecemeal deletion will keep costing a re-bless every time the field being removed happens to sit inside a stamped artifact.

## The principled fix

Apply single-source-of-truth + content-addressing over *canonical inputs* to the prep pipeline, in this order:

1. **Take derived artifacts OUT of the identity.** The blessed identity should be a hash of **authoritative inputs only** — FASTA bytes, sourced GenBank/cdot annotation, curated overrides. Derived artifacts become a rebuildable cache layer *outside* the identity. This is the keystone: once done, deletions like #2058/#2064 are free no-ops, because a derived field is no longer part of what the identity commits to.

2. **Records store only sourced fields; derived quantities are computed at read time** from the single authority — e.g. `sequence_length` read from the FASTA index on demand, never persisted as a second copy.

3. **Make the sourced/derived split STRUCTURAL, not a comment.** Encode it in types so the compiler forbids derived data from entering a persisted-authority struct (a sourced-only record vs a computed view composed at read time), rather than relying on a doc-comment that says "this field is derived."

4. **If a derived value must be cached for perf, tag it with source-provenance and validate on load** — at field/source granularity, beyond today's format-version bumps — so a stale cache is *detected*, not silently trusted.

## Symptom issues this subsumes

- **#2058** — removed a hollow `sequence_length` from the non-stamped supplemental sidecar. Identity-neutral *only* because that sidecar is not stamped. A symptom of consequence (1)+(3).
- **#2064** — the same field on `LegacyTranscript` (`src/prepare/mod.rs:221`), which *is* inside content-stamped artifacts, so removing it bumps the blessed identity. A symptom of consequence (2): the identity model makes the cleanup expensive.
- **#1737 / #1759** — the cdot two-coordinate-space / direction-dependence family: a derived quantity persisted beside its authority and free to disagree. Same disease, different artifact.

## Prior decisions this revisits

- **#905** — "derived artifacts are content-stamped." This is the decision move 1 revisits: content-stamping derived artifacts is what couples the identity to redundant data. The proposal is not to abandon content-addressing but to *re-scope* it to authoritative inputs, with derived artifacts stamped/validated as a cache layer instead of as identity.
- **#933** — `prepared_at` excluded from the identity so the two prepared volumes do not split. This is precedent *for* the direction here: the identity should be a pure content signature over what actually matters, and non-authoritative data (a wall-clock stamp then; derived fields now) should not be in it.

The blessed identity `aa8b3246d83055cc` is pinned in `tests/fixtures/mutalyzer-normalize/empty-projection/*.count`; the full move-1 redefinition is a deliberate re-bless with a large blast radius (all derived artifacts) and should be sequenced carefully.

## First increment (accompanying PR)

Scoped to **move 3, on the legacy records, identity-neutral** — the smallest coherent step that advances the vision with zero identity risk:

- Introduce a **sourced-only authority type** `LegacyTranscript` (`id`, `gene_symbol`, `cds_start`, `cds_end`) that the compiler guarantees *cannot* carry a derived length.
- Quarantine the derived `sequence_length` in a distinct, clearly-named **persistence-boundary record type** (`LegacyTranscriptRecord`), which remains the sole carrier of the derived field across serialization.
- Prove structurally (tests) that (a) the sourced authority type serializes with **no** `sequence_length` key, (b) the on-disk bytes and their FNV content-stamp are **byte-for-byte unchanged** (so the blessed identity does not move), and (c) the persisted length is exactly the FASTA-derived length (`seq.len()`), i.e. fully redundant with its authority.

**Deferred (not in the first PR):** actually dropping `sequence_length` from disk (that is #2064, which *does* move the identity and needs an operator re-prepare + re-bless of both volumes), and moves 1/2/4 (taking derived artifacts out of the identity; read-time derivation as the sole length source; provenance-tagged cache validation). The first increment deliberately keeps the on-disk format and identity fixed so it lands cleanly, while establishing the sourced/derived type boundary the later moves build on.

Refs #2058, #2064, #1737, #1759, #905, #933.

nh13:
Refs #2085, #2105, #2064, #2058.

## What this is

The first, deliberately identity-neutral increment of #2085 (make the prepared-reference data model distinguish **sourced** authority from **derived**, recomputable data). It targets **move 3** of that issue — *make the sourced/derived split structural, not a doc-comment* — scoped to the legacy-transcript metadata records, with **zero** impact on the blessed reference identity.

**It does not close #2085.** #2085's principled fix is numbered 1–4 and this PR implements only move 3. Move 1 is the one #2085 itself calls the keystone ("take derived artifacts OUT of the identity"); moves 2 and 4 (read-time derivation as the sole length source; provenance-tagged cache validation) likewise have no other tracker. #2064 tracks only the field deletion. **#2105 exists to hold moves 1, 2 and 4**, so #2085 must stay open — note that #2085's own "First increment (accompanying PR)" section describes this PR and makes closing it look sanctioned; it is not.

## The change

`LegacyTranscript` conflated a sourced authority (accession, gene symbol, CDS bounds — from GenBank) with a **derived** `sequence_length` (its authority is the FASTA; the persisted value is exactly `seq.len()` and is recomputable from the `.fai` index). Nothing in `src/` ever read the persisted length.

This splits the one type into two:

- **`LegacyTranscript`** — the sourced authority. It now carries only `id`, `gene_symbol`, `cds_start`, `cds_end`, so the compiler *structurally forbids* a derived length from entering it.
- **`LegacyTranscriptRecord`** — the on-disk row: the sourced fields plus the derived `sequence_length`. It is the sole carrier of the derived length across the persistence boundary, with its field order matching the pre-split struct exactly.

**The writer goes through the seam rather than around it**, which is what makes the split load-bearing rather than a rename. The per-record step of `fetch_legacy_versions` is factored into `legacy_record_from_genbank`, which builds the sourced authority first and attaches the length only via `LegacyTranscript::into_record`. That helper is the **only** construction of `LegacyTranscriptRecord` in `src/`, so a derived length cannot reach persisted metadata without passing `into_record` — and a writer that bypassed it would leave the helper unused and fail the `-D warnings` build (verified by sabotage: `error: function legacy_record_from_genbank is never used`).

`LegacyMetadata::sourced_transcripts()` and `LegacyTranscriptRecord::sourced()` expose the length-free authority view. These two are new public surface with **no in-tree consumer yet**: they are the reader-side API move 2 of #2085 needs, and are stated here rather than left to be discovered.

## Breaking change

Declared, because both types are re-exported from the public `prepare` module (`src/lib.rs`: `pub mod prepare;`, and `pub mod benchmark;` under that feature), so this breaks downstream compilation:

- `LegacyTranscript` loses its `sequence_length` field.
- `LegacyMetadata.transcripts` is retyped to `HashMap<String, LegacyTranscriptRecord>`.

The commit is typed `refactor(prepare)!:` and carries a `BREAKING CHANGE:` footer so release-plz classifies it. **The on-disk JSON is unchanged** — this is a source-compatibility break only.

BREAKING CHANGE: `ferro_hgvs::prepare::LegacyTranscript` no longer has a `sequence_length` field, and `LegacyMetadata::transcripts` is now a `HashMap<String, LegacyTranscriptRecord>`. Downstream code reading a persisted length must move to `LegacyTranscriptRecord`.

## Identity impact: UNCHANGED

The blessed reference identity `aa8b3246d83055cc` does **not** move. `legacy_transcripts_metadata` / `legacy_genbank_metadata` are content-stamped from **on-disk bytes** (`src/prepare/identity.rs` does `fs::read(path)` → `fnv1a_hex(&bytes)`, never a re-serialized struct), and this increment keeps those bytes byte-for-byte identical: `LegacyTranscriptRecord` serializes exactly as the old `LegacyTranscript` did. No re-prepare is performed, so no artifact changes. No re-bless, no `.count` fixture change.

Actually *dropping* the field from disk (which **would** move the identity and needs an operator re-prepare + re-bless of both volumes) is #2064 and stays deferred, along with moves 1/2/4 of #2085 (see #2105). This increment establishes the type boundary those later moves build on.

## Tests

New unit tests in `src/prepare/mod.rs` on the production writer path:

- `the_writer_pairs_the_sourced_authority_with_the_fasta_derived_length` — guards `legacy_record_from_genbank`, the step `fetch_legacy_versions` performs per record. `fetch_legacy_versions` shells out to `curl` and cannot be reached from a test, which is precisely why the seam had to be factored out to be exercised at all.
- `an_empty_genbank_gene_name_becomes_no_gene_symbol` — the writer's own rule, pinned so the extraction cannot quietly change it.

New `sourced_derived_split` module in `tests/it/legacy_integration.rs` for the type-level half:

- `sourced_authority_type_has_no_length_field` — the sourced type never serializes a `sequence_length` key.
- `on_disk_bytes_are_identity_neutral` — serialized metadata equals the pre-split golden byte-for-byte.
- `content_stamp_is_the_pinned_digest` — the FNV-1a stamp against a **literal** digest (`3135b86e16384d04`), derived independently of ferro. The earlier form compared `fnv1a_hex(serialized)` against `fnv1a_hex(GOLDEN_PRETTY)`, which is `f(x) == f(x)` given the assertion above it and holds for any hash function.
- `persisted_length_is_the_fasta_derived_length` — the persisted length equals `seq.len()`, and `sourced()` drops it on the round trip.
- `sourced_transcripts_view_drops_the_derived_length` — the consumer-facing view is length-free.

The three existing legacy serialization tests are updated to the new types.

### Sabotage table

Each mutation applied to the tree, measured, reverted:

| mutation | result |
|---|---|
| `into_record` ignores its argument (`sequence_length: 0`) | **RED** — `the_writer_pairs_...` fails |
| writer's call site passes `sequence.len() + 1` | **RED** — same test fails |
| writer bypasses the helper with the pre-fix inline struct literal | **RED** — `cargo clippy --all-features --all-targets -- -D warnings` fails with `function legacy_record_from_genbank is never used` |

Before this change, `into_record` had **zero** callers in `src/`, so no mutation of it could redden anything but a test-only caller.

## Verification

- `cargo nextest run --features dev --no-fail-fast` — `EXIT=0`, 11374 passed, 0 failed.
- `cargo clippy --all-features --all-targets -- -D warnings` — `EXIT=0`.
- `cargo fmt --all --check` — `EXIT=0`.
- `python3 -m pytest tests/python/ -q` — `EXIT=0`, 1137 passed, 8 skipped.

Rebased onto `origin/main` at `0d542858`.

Representation-Change: none. Data-model type split in `src/prepare/` and `src/benchmark/` (neither a watched prefix); no `Normalizer`/`VariantProjector` path is touched and no normalized HGVS output moves.

### DIFF
--- a/src/benchmark/mod.rs
+++ b/src/benchmark/mod.rs
@@ -98,8 +98,8 @@ pub use parse::{parse_ferro, parse_ferro_unified};
 pub use runner::{load_existing_results, run_benchmark};
 // Re-export from the main prepare module (avoiding code duplication)
 pub use crate::prepare::{
-    check_references, prepare_references, LegacyMetadata, LegacyTranscript, PrepareConfig,
-    ReferenceManifest,
+    check_references, prepare_references, LegacyMetadata, LegacyTranscript, LegacyTranscriptRecord,
+    PrepareConfig, ReferenceManifest,
 };
 pub use report::generate_report;
 pub use sample::stratified_sample;
--- a/src/prepare/mod.rs
+++ b/src/prepare/mod.rs
@@ -206,8 +206,18 @@ pub mod urls {
         "https://ftp.ebi.ac.uk/pub/databases/lrgex/list_LRGs_transcripts_xrefs.txt";
 }
 
-/// Metadata for a legacy transcript fetched from GenBank.
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+/// Sourced (authoritative) metadata for a legacy transcript fetched from
+/// GenBank: accession, gene symbol, and CDS bounds.
+///
+/// The transcript LENGTH is deliberately **not** a field here. Its sole
+/// authority is the FASTA (`seq.len()`, recoverable from the `.fai` index), so
+/// storing it on the sourced record would make a second, drift-prone copy of a
+/// value the FASTA already owns. This type therefore *structurally* cannot carry
+/// a derived length — the compiler enforces the sourced/derived split (#2085).
+/// The on-disk row that still carries the length for byte-compatible metadata is
+/// [`LegacyTranscriptRecord`]; dropping it from disk entirely is tracked by #2064
+/// (a deliberate re-bless of the reference identity).
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub struct LegacyTranscript {
     /// Accession with version (e.g., "NM_000051.3")
     pub id: String,
@@ -217,8 +227,68 @@ pub struct LegacyTranscript {
     pub cds_start: Option<u64>,
     /// CDS end position (1-based, inclusive)
     pub cds_end: Option<u64>,
-    /// Total sequence length
+}
+
+/// On-disk row for a legacy transcript: the sourced [`LegacyTranscript`] fields
+/// plus the **derived** `sequence_length`.
+///
+/// This is the *only* type that carries the derived length across the
+/// persistence boundary. It exists so the structural sourced/derived split (the
+/// authority type has no length) is byte-for-byte identity-neutral (#905/#933):
+/// the serialized JSON is unchanged from the pre-split format, so the content
+/// stamps in [`crate::prepare::identity`] and the blessed reference identity do
+/// not move. Field order matches the pre-split `LegacyTranscript` exactly, which
+/// is what keeps the bytes identical. Dropping the field from disk (which *would*
+/// move the identity) is #2064.
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct LegacyTranscriptRecord {
+    /// Accession with version (e.g., "NM_000051.3")
+    pub id: String,
+    /// Gene symbol if available
+    pub gene_symbol: Option<String>,
+    /// CDS start position (1-based, inclusive)
+    pub cds_start: Option<u64>,
+    /// CDS end position (1-based, inclusive)
+    pub cds_end: Option<u64>,
+    /// Total sequence length. **Derived**: its authority is the FASTA / `.fai`;
+    /// this persisted copy equals `seq.len()` at write time and is recomputable.
     pub sequence_length: usize,
+}
+
+impl LegacyTranscript {
+    /// Pair this sourced record with the **derived** length taken from its FASTA
+    /// authority at write time, producing the on-disk [`LegacyTranscriptRecord`].
+    ///
+    /// This is the seam the split exists for, and it is load-bearing rather than
+    /// advisory: [`legacy_record_from_genbank`] — the only construction of
+    /// `LegacyTranscriptRecord` in `src/`, and the step
+    /// [`fetch_legacy_versions`] performs per GenBank record — reaches the
+    /// persisted row exclusively through here, so a derived length cannot enter
+    /// production metadata without passing this function. A writer that bypassed
+    /// it would leave that helper unused and fail the `-D warnings` build.
+    pub fn into_record(self, sequence_length: usize) -> LegacyTranscriptRecord {
+        LegacyTranscriptRecord {
+            id: self.id,
+            gene_symbol: self.gene_symbol,
+            cds_start: self.cds_start,
+            cds_end: self.cds_end,
+            sequence_length,
+        }
+    }
+}
+
+impl LegacyTranscriptRecord {
+    /// Project to the sourced authority, dropping the derived length. Callers
+    /// that need the length must consult the FASTA / `.fai` (the authority),
+    /// never this persisted copy.
+    pub fn sourced(&self) -> LegacyTranscript {
+        LegacyTranscript {
+            id: self.id.clone(),
+            gene_symbol: self.gene_symbol.clone(),
+            cds_start: self.cds_start,
+            cds_end: self.cds_end,
+        }
+    }
 }
 
 /// Metadata file for legacy transcripts.
@@ -226,8 +296,21 @@ pub struct LegacyTranscript {
 pub struct LegacyMetadata {
     /// When the metadata was generated
     pub generated_at: String,
-    /// Map from accession to transcript metadata
-    pub transcripts: HashMap<String, LegacyTranscript>,
+    /// Map from accession to the on-disk transcript row.
+    pub transcripts: HashMap<String, LegacyTranscriptRecord>,
+}
+
+impl LegacyMetadata {
+    /// The sourced authority view of every transcript, with the derived length
+    /// dropped. This is the API consumers should use to reason about legacy
+    /// transcripts so a persisted length can never be mistaken for the source of
+    /// truth (#2085); the length, when needed, comes from the FASTA / `.fai`.
+    pub fn sourced_transcripts(&self) -> HashMap<String, LegacyTranscript> {
+        self.transcripts
+            .iter()
+            .map(|(id, record)| (id.clone(), record.sourced()))
+            .collect()
+    }
 }
 
 /// Prepare reference data for normalization.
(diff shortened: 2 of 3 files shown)


nh13:
@coderabbitai pause

nh13:
Independent review (read-only, throwaway worktree; nothing pushed). **The identity question is settled — it does not move**, and I verified that two ways rather than reasoning about it:

- **Structurally:** `src/prepare/identity.rs` content-stamps via `fs::read(path)` → `fnv1a_hex(&bytes)`. It hashes on-disk artifact bytes, never re-serialized Rust structs. This PR runs no `ferro prepare`, so the bytes are untouched and the identity *cannot* move.
- **Empirically, through the harness's own path** (I did not re-implement `reference_identity_from_manifest` — that hand-port returns `dc29bb2e1ff9a5aa` and reads as false drift): ran the empty-projection gate at this head with `FERRO_MANIFEST` on the scratch manifest. `axis_protein_description` reported `1 empty`, matching the committed baseline `protein_description.count` = 1 / `aa8b3246d83055cc`, with **no** `empty-projection gate SKIPPED` notice and zero `Skipping:` lines. The gate armed and passed.

The golden-bytes test is also a real guard, not decorative: a `#[serde(rename = "seq_len")]` on the persisted derived field reddens `on_disk_bytes_are_identity_neutral`, `content_stamp_is_unchanged` and `legacy_tests::test_legacy_metadata_deserialization`. Field order matches the pre-split struct, so the JSON is byte-identical.

Three findings. The first blocks merge as written.

---

**1. `Closes #2085` is wrong — it closes a four-move issue after implementing one move.**

#2085's principled fix is numbered 1–4; this PR implements **move 3** and says so ("moves 1/2/4 … stay deferred"). Move 1 is the one #2085 itself calls **"the keystone"** — *"take derived artifacts OUT of the identity … once done, deletions like #2058/#2064 are free no-ops."* Moves 1, 2 and 4 exist in no other issue: `--search "2085"` returns only #2085 and an unrelated #1450; #2064 tracks only the field deletion.

The confounder is real and worth naming: #2085's own body has a "First increment (accompanying PR)" section describing this PR exactly, which makes the close look sanctioned. It isn't — three undone moves are listed below that section. Filed #2105 so they survive either way; please downgrade to `Refs #2085`, or close a new move-3-scoped issue instead.

**2. Undeclared breaking public-API change.**

`src/lib.rs` has `pub mod prepare;` and `pub mod benchmark;`, and both re-export these types. This PR removes the `sequence_length` field from the public `LegacyTranscript`, and changes `LegacyMetadata.transcripts` from `HashMap<String, LegacyTranscript>` to `HashMap<String, LegacyTranscriptRecord>`. Both break downstream compilation. The commit is `refactor(prepare): …` with no `!` and no `BREAKING CHANGE:` footer, so release-plz will not classify it as breaking at `0.14.0`. Either add the marker, or state deliberately that these types are not considered public API.

**3. The structural guarantee currently guards nothing, because nothing uses the type it guards.**

Grepping `src/` for each new symbol:

- `LegacyTranscript` (the new sourced authority) — **zero constructions in `src/`**; built only in tests.
- `into_record` — **zero callers outside tests**. The only writer, `src/prepare/mod.rs:2420`, builds `LegacyTranscriptRecord { … }` as a struct literal, bypassing the very method whose doc comment says it "pairs the authority with its FASTA-derived length at write time".
- `sourced_transcripts()` / `sourced()` — test-only.

So in production this is a rename plus three unused `pub` methods and one unused `pub` type. "The compiler structurally forbids a derived length from entering it" is literally true and operationally vacuous — no code path enters it. **Routing the writer through `into_record` is ~3 lines and makes the split load-bearing.** That would be the difference between a refactor that documents an intent and one that enforces it.

*Nit (4):* `content_stamp_is_unchanged` asserts `fnv1a_hex(serialized) == fnv1a_hex(GOLDEN_PRETTY)` while the test above it already asserts `serialized == GOLDEN_PRETTY` — that is `f(x) == f(x)`, unfalsifiable unless FNV is nondeterministic, and it pins no hash literal so it never touches the real stamp. If it is meant to guard the identity contribution, assert a literal hex digest.

---

`Representation-Change: none` is **correct** — verified against the real `scripts/check_representation_change.py` on current main: `src/prepare/` and `src/benchmark/` are not watched prefixes, EXIT=0. Merges cleanly.

PR #2088, opened 2026-08-16
ci: make the nightly xfail gate propagate its exit code (pipefail/tee)

### LINKED ISSUE #2082: ci: the nightly xfail gate cannot go red — its exit code is swallowed by `tee`

nh13:
Closes #2082.

## Problem

The nightly xfail gate added in #2079 (`.github/workflows/nightly-mutalyzer.yml`, step *Diff FAIL set against xfail baseline*) ends its `run:` block in a pipe:

```sh
python3 scripts/diff_nightly_xfail_baseline.py --junit "$JUNIT" --baseline "$BASELINE" | tee -a "$GITHUB_STEP_SUMMARY"
```

The step names no `shell:` and the workflow sets no `defaults.run.shell`, so GitHub runs it under `bash -e {0}` — which does **not** include `pipefail`. A pipeline's exit status is the status of its *last* command, and `tee` exits 0. So the Python script's non-zero exit (1 on drift, 2 on a missing/unparseable JUnit) is swallowed: the diff prints `NEW FAILURE`, the step exits 0, the job concludes `success`, `report-failure` never fires, and no issue opens. The gate #2079 added is a no-op that cannot go red — the exact false-green class of #1998 reproduced inside its own fix.

The step's inline comment even asserted the opposite ("`set -o pipefail` (GitHub's bash default)…") — that false belief is the root cause, so it is corrected here too. `pipefail` is added only by an explicit `shell: bash`, which expands to `bash --noprofile --norc -eo pipefail {0}`.

## Fix

Add `shell: bash` to the step **and** `set -euo pipefail` as the first line of the `run:` block. This matches the established repo convention — every other `run:` block that needs pipefail sets it explicitly (`nightly-perf.yml`, `nightly-soak.yml`, `ci.yml`, `report-failure.yml`, `release-binaries.yml`); this step was the sole slip from it. The `$GITHUB_STEP_SUMMARY` write is unchanged — `tee` still records the report to the job summary; only exit propagation changes.

`pipefail` also converts the missing/unparseable-JUnit case (script exit 2) into a red step, so a nextest crash before the report is written is no longer swallowed either.

## Reproduction (before/after, at the shell)

The exact pipeline, script forced non-zero (missing JUnit → exit 2):

| shell | step exit | verdict |
|---|---|---|
| `bash -e {0}` (what the step gets today) | **0** | swallowed — false green |
| `shell: bash` (`bash --noprofile --norc -eo pipefail {0}`) + `set -euo pipefail` | **2** | step RED |

Generic mechanism, for clarity:

```
$ bash -e -c 'false | tee /dev/null; echo exit=$?'                              # exit=0  (bug)
$ bash --noprofile --norc -eo pipefail -c 'false | tee /dev/null; echo exit=$?' # aborts under set -e -> step exit 1 (fixed)
```

`actionlint .github/workflows/nightly-mutalyzer.yml` is clean.

## Sibling audit

`| tee ...$GITHUB_STEP_SUMMARY` at line 367 is the **only** such gating pipe in the entire `.github/workflows/` tree (the sole other `$GITHUB_STEP_SUMMARY` write in this file, line 332, is a `} >>` redirect, not a pipe — unaffected). No other workflow relies on the shell default for a gating pipe.

## Note on stacking

#2079 is still open and unmerged, and both the vulnerable step and `scripts/diff_nightly_xfail_baseline.py` exist only on its branch (`ci/issue-1998-nightly-xfail-baseline`), not on `main`. This PR is therefore **based on the #2079 branch** rather than `main`, and contains only the one-step fix. GitHub will retarget it to `main` when #2079 merges. Alternatively the maintainer may fold this single change directly into #2079 and close this PR.

Representation-Change: none. CI-plumbing (workflow) change only; touches no watched directory.


nh13:
@coderabbitai pause

nh13:
Superseded by #2079, which is in the merge queue (position 1, CLEAN) and carries `Closes #2082`. #2079 fixes the same defect with `set -euo pipefail` on the diff step, deletes the false "GitHub's bash default" comment, and adds a systemic guard (`tests/it/workflow_pipefail_gate.rs`) that reddens on any workflow pipeline lacking pipefail — which also caught four other pre-existing violations. This PR was stacked on #2079's pre-rebase base commit, so it reads DIRTY.

The one delta here — an explicit `shell: bash` on the step (belt-and-braces atop the `set`) — is redundant for pipefail once `set -euo pipefail` is the first line, and #2079's guard is satisfied either way. If the team wants it as documented defense-in-depth, it's a trivial standalone follow-up off main after #2079 lands. Closing as superseded.

PR #2089, opened 2026-08-16
fix(reference): honor the requested build for LRG genomic placement

### LINKED ISSUE #1903: genomic_placement_on_build ignores the requested build for LRG, returning a cross-build placement
`ReferenceProvider::genomic_placement_on_build` discards the requested genome build for LRG accessions and returns whatever placement the record carries. The same trait method declines correctly for `NG_`, so the two accession families behave oppositely through one entry point.

Measured against a fixture whose record carries a GRCh38 placement:

```
requested = Some("GRCh37")     -> Some(("NC_000017.11", 100, 109))   # a GRCh38 placement
requested = Some("T2T-CHM13")  -> Some(("NC_000017.11", 100, 109))   # a build that does not exist
NG_ accession, Some("GRCh37")  -> None                               # correct
```

`infer_genome_build` already tells the families apart, so the information needed to decline is present and unused.

### Why the downstream guard does not catch it

The in-code rationale is that a cross-build placement is caught by the re-anchor's endpoint-outside-span check. It is not. That guard (`src/project/projector.rs:3050-3061`) is purely numeric — two `nc_to_parent` calls — and never compares `placement.nc` against the chromosome the coordinates are on. So it only fires when the two builds' spans happen not to overlap numerically.

**241 of 1293** LRG records have numerically overlapping GRCh37/GRCh38 spans. For those, a cross-build request is answered rather than refused.

On `LRG_1021`'s real spans:

```
asked for GRCh37, got nc = NC_000004.12
GRCh37 pos 1800000: reanchor yields Some(11689), truth is 9962   -> 1727-base error
```

The result is well-formed and in-bounds, so it is a fixed point — every armed oracle passes on it. That is the same signature as the defects this repo has been closing one at a time: not a crash, not a refusal, but a confidently wrong coordinate.

### Scope

Inherited, not introduced by any PR currently open. Surfaced during a dual-review pass over #1830, whose diff does not touch `genomic_placement_on_build`, `select_placement_for_build`, or the placement cache — the only textual overlap is a comment naming the function. #1830's narrowing in fact *reduces* exposure slightly, from 241 records to 240.

### Note for whoever picks this up

A build-aware fix has to key the **placement cache** on the build as well. It currently is not, so teaching only the lookup would leave the first-requested build's placement cached and served to every later requester regardless of what they ask for.

nh13:
## Summary

`ReferenceProvider::genomic_placement_on_build` discarded the requested genome build for LRG accessions: it returned the LRG record's single main-assembly (GRCh38) placement regardless of what build the caller asked for. An explicit non-GRCh38 request (e.g. `GRCh37`, or a build that does not exist) therefore received a cross-build placement and re-anchored coordinates onto the wrong genome — a well-formed, in-bounds, confidently-wrong coordinate. The `NG_` path already declines correctly in the same situation, so the two accession families behaved oppositely through one entry point.

The in-code rationale was that a cross-build request would be caught downstream by the re-anchor's endpoint-outside-span guard. It is not: that guard is purely numeric and never compares the placement's `nc` against the chromosome the coordinates are on, so it only fires when the two builds' spans happen not to overlap numerically. 241 of 1293 LRG records have numerically overlapping GRCh37/GRCh38 spans, and for those a cross-build request was answered rather than refused.

## Fix

Route the parsed LRG placement through the same `select_placement_for_build` policy the `NG_` path uses: an explicit build must match (or none was requested), otherwise it declines rather than mis-anchoring onto a different build's coordinates.

Per the issue's note, the fix keeps the placement cache from being build-poisoned. The cache (`lrg_placement_cache`) is keyed on the bare `LRG_<N>` accession and stores the **build-agnostic** parsed placement; the build filter runs *after* the cache. So a first-requested build's result can never be served to a later requester asking for a different build.

## Test

`lrg_genomic_placement_honors_the_requested_build` pins that:
- a `Some("GRCh37")` request (a build the LRG record does not carry) declines rather than returning the GRCh38 placement;
- the matching `Some("GRCh38")` request and the no-preference `None` request both return the record's placement;
- issuing the declined GRCh37 request **first** does not poison the cache against the later GRCh38 request.

Verified failing-first: with the fix reverted, the test fails on the GRCh37-decline assertion.

Representation-Change: none. LRG genomic placement now honours the requested build; the only affected inputs are explicit non-GRCh38 build requests on LRG accessions, which previously returned a cross-build (wrong) placement and now correctly decline. No previously-correct output moves — the default build-agnostic path (`build=None`) and an explicit `GRCh38` request are both byte-identical to before.

Closes #1903

### DIFF
--- a/src/reference/multi_fasta.rs
+++ b/src/reference/multi_fasta.rs
@@ -1502,6 +1502,42 @@ impl MultiFastaProvider {
         None
     }
 
+    /// The build-agnostic LRG placement parsed from the on-hand LRG XML, cached
+    /// per accession.
+    ///
+    /// The parsed placement is stored verbatim (the parse reads only the single
+    /// `main_assembly` mapping and does not depend on any requested build), so
+    /// the cache key is the bare `LRG_<N>` accession and the build filter is
+    /// applied by the caller *after* this returns — see
+    /// [`genomic_placement_on_build`](Self::genomic_placement_on_build). Keeping
+    /// the filter out of the cache is what stops a first-requested build from
+    /// being served to every later requester (#1903).
+    fn lrg_placement_cached(
+        &self,
+        parent: &crate::hgvs::variant::Accession,
+    ) -> Option<GenomicPlacement> {
+        let dir = self.lrg_xml_dir.as_ref()?;
+        let key = parent.full(); // e.g. "LRG_1" (LRG accessions carry no version)
+
+        if let Some(cached) = self
+            .lrg_placement_cache
+            .lock()
+            .expect("lrg placement cache poisoned")
+            .get(&key)
+        {
+            return cached.clone();
+        }
+
+        let placement = std::fs::read_to_string(dir.join(format!("{key}.xml")))
+            .ok()
+            .and_then(|xml| parse_lrg_main_assembly_placement(&xml));
+        self.lrg_placement_cache
+            .lock()
+            .expect("lrg placement cache poisoned")
+            .insert(key, placement.clone());
+        placement
+    }
+
     /// Resolve a sequence name, trying various forms
     fn resolve_name(&self, name: &str) -> Option<String> {
         // Direct lookup, plus the LRG bare-genomic → `LRG_<N>g` record rename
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #2090, opened 2026-08-16
fix(test): stamp fixture_gen cache with generator revision and invalidate on mismatch

### LINKED ISSUE #1654: fixture_gen reuses a cached corpus without checking which generator revision produced it
`tests/it/common/fixture_gen.rs`'s cache is **existence-only**:

```rust
fn ensure_generated(path: &Path, …) {
    if path.exists() {
        return;
    }
```

There is no generator-version marker, no content hash of the generator source, and no mtime comparison. So a generated corpus produced by an **older revision of its generator** is reused verbatim after a checkout that changes that generator, and every pin measured over it is measured over the wrong artifact.

This is the "stale local artifact" class the repository already treats as first-class — `CLAUDE.md` names `pinned_v21_normalization_behavior` and `enumeration_replays_recorded_behavior` as stale-local-artifact detectors, and the spec fixtures are regenerated explicitly by CI for exactly this reason. `fixture_gen` is the one path where the artifact is *cached* rather than regenerated, and it is the one with no staleness check.

### Why it has not bitten yet

Individual consumers guard the consequence by hand. `tests/it/cis_confluence_nr_axis.rs` asserts the corpus carries no axis outside `n.`/`r.` — *"the corpus carries an axis other than n./r.; it is the wrong artifact"* — and `assert_census` compares the whole `Census`, so a stale `g,c`-only corpus fails on `classes` rather than passing with a flattering figure.

That works, and it is per-consumer. A consumer that does not think to write such a guard gets silent reuse, and the failure mode is a **pin measured over a corpus that cannot contain the shape the pin is about** — indistinguishable from a real green.

### Why CI does not cover it

CI starts from a clean checkout, so the fixture never exists and is always generated. The hazard is local-only — which makes it a developer-trust problem rather than a gate problem, and those are the ones that produce a confidently wrong local measurement.

### Suggested shape

Stamp the generator's identity into the artifact and compare it on reuse: a version constant bumped by the generator's author, or a hash of the generator source, written into the fixture's metadata. Regenerate when the stored marker differs; keep reuse for a match, so the fast path is unchanged.

Note the existing per-consumer guards should stay either way — they check something narrower and more specific than a version marker can (that *this* corpus has *these* axes), and the two are complementary.

### Provenance

Surfaced by the CodeRabbit CLI lens while reviewing #1646, as a `minor` finding against that PR. Filed separately rather than folded in, because `fixture_gen.rs` is shared by several fixtures and #1646's own exposure is already guarded — expanding a test PR into shared infrastructure is the wrong scope.

nh13:
Closes #1654.

## Problem

`tests/it/common/fixture_gen.rs`'s cache was existence-only: `if path.exists() { return; }`. A generated corpus produced by an older revision of its generator was reused verbatim after a checkout that changed that generator, so any pin measured over it was measured over the wrong artifact. This is the "stale local artifact" class the repo already treats as first-class, and `fixture_gen` was the one cached path with no staleness check. It is local-only (CI starts clean, so the fixture never exists and is always generated), which makes it a developer-trust hazard rather than a gate failure — the worst kind, since it produces a confidently wrong local measurement.

## Fix

Stamp each generated fixture with a revision token and compare it on reuse:

- `generator_revision(generator, args)` hashes the generator's own source (`examples/<generator>.rs`) together with the args it is run with (the same generator backs the `g,c` and `n,r` confluence corpora, differing only in flags). Any edit to the generator source, or a change to a caller's flags, moves the token.
- The token is written to a gitignored `<fixture>.genrev` sidecar on a successful install, and `fixture_is_fresh` gates reuse on it.

Freshness has three cases:

- **absent** — generate.
- **present, no stamp** — trusted. A stamp-less fixture was produced outside this helper (CI's `spec-fixtures` job, a manual `cargo run`), so this keeps CI on the download-and-reuse fast path rather than rebuilding the whole crate inside every test.
- **present, stamped** — reused only if the stamp still names the current revision; otherwise regenerated.

The token deliberately hashes only the generator source and args, not every transitive input (spec submodule, curated overrides, `examples/common/` includes) — those are covered by the committed per-consumer guards the fixtures feed. The failure this closes is the one nothing else observes: reusing a corpus built by an older revision of its own generator.

The four fixtures driven through this helper (`hgvs_spec_normalization.json`, `hgvs_spec_enumeration.json`, and the two `cis_confluence_*` corpora) land under `tests/fixtures/grammar/` and `tests/fixtures/cis/`; `.gitignore` gains `*.genrev` for both directories.

## Tests

`fixture_gen`'s unit module gains coverage for the whole decision:

- the token is deterministic and arg-sensitive;
- a stamp mismatch forces regeneration and a match reuses (the property #1654 asked for);
- a stamp-less fixture is trusted (CI fast path);
- an unreadable/corrupt (non-UTF-8) stamp is treated as stale, so only a genuinely missing sidecar takes the trusted path;
- write-then-read round-trips.

## Notes

Watched-prefix check: this change touches only `.gitignore` and test infrastructure — nothing under `src/normalize|hgvs|spdi|project|reference|error_handling` — so no `Representation-Change:` trailer applies.


nh13:
@coderabbitai pause

PR #2093, opened 2026-08-16
fix: file declining fix: commits under Fixed in the changelog

### LINKED ISSUE #1557: changelog: a declining fix: commit is filed under Other instead of Fixed

nh13:
Closes #1557.

## Problem

A commit carrying a declining `Representation-Change: none` trailer was filed under **Other** regardless of its conventional type, so a `fix:` that correctly declined never appeared under **Fixed**. In v0.13.1, #1533 (`fix(spec-fixture)`) and #1527 (`fix(changelog)`) landed under Other rather than Fixed.

This was a side effect of #1527's decline-exclusion parser: git-cliff assigns exactly one group and takes the first matching parser, so "not a representation change" was spelled as "route to Other". As #1557 documents, the direct repairs all fail (measured against git-cliff 2.13.1):

- `{ footer = <decline>, message = "^fix", group = "Fixed" }` — git-cliff evaluates a parser's fields as **OR, not AND**, so this matches every `fix:` regardless of its trailer, emptying the Representation changes section.
- Omitting `group` does not fall through — the commit falls back to the raw conventional type (`fix`/`test`), losing group ordering.
- A negative footer regex needs a lookahead the `regex` crate does not have (and the enumerated complement is broken by the `no` ⊂ `none` prefix overlap: `none.` gets misread as a real change).

## Fix

A `commit_preprocessor` in `release-plz.toml` renames the trailer **key** on declining commits only, before the parsers run: `Representation-Change:` → `Representation-Change-Declined:`. The inclusion parser's literal `:` no longer matches, so a declining commit falls through to its conventional-type parser (a declining `fix:` → Fixed). Real declarations keep their key and are grouped into Representation changes exactly as before.

The rename touches the footer, never the subject, so no rendered bullet changes and no strip postprocessor is needed; `scripts/inject_representation_disclosure.py` reads the raw commit message, so a decline's own text is untouched there too. The decline regex is the exact one `scripts/check_changelog_grouping.py` audits with, so the audit and the config agree on which commits decline by construction.

Setting `commit_preprocessors` replaces release-plz's defaults wholesale, so its only default — the `(#N)` → PR-link rewrite — is reproduced as the second entry, or the changelog would lose every PR link.

## Tests

- `check_changelog_grouping.py`'s `write_cliff_config` now copies `commit_preprocessors` (grouping depends on it), so the audit renders with the neutralizer.
- New end-to-end tests in `tests/python/test_changelog_grouping.py` build throwaway repos and render through the real `release-plz.toml` with real git-cliff, asserting a declining `fix:` lands under Fixed, a declining `feat:` under Added, and real declarations under Representation changes (including the documented edge forms `none, except ...` and `no rows move`). They run in the `Changelog grouping audit` CI job, which has git-cliff, and a structural guard runs everywhere.
- The config-reading parity tests in `test_representation_change_trailer.py` are updated to the new preprocessor location.

Verified the guard fails when the preprocessor is removed (sabotage check).

Representation-Change: none. Changelog tooling only; no watched source directory (`src/normalize|hgvs|spdi|project|reference|error_handling`) is touched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -666,6 +666,17 @@ jobs:
         with:
           python-version: "3.12"
       - uses: taiki-e/install-action@a8fb21b85430c7a58551cf7ea87b03f355b826d9  # git-cliff
+      - name: Test the changelog grouping (renders through git-cliff)
+        # This is where `tests/python/test_changelog_grouping.py`'s end-to-end tests actually
+        # run: they render throwaway repos through the real `release-plz.toml` and assert
+        # where each commit lands (e.g. a declining `fix:` under Fixed, #1557), which needs
+        # git-cliff. The Python Wheel Test job has no git-cliff, so those tests skip there —
+        # they must not be left to skip everywhere. No wheel is needed; the tests import the
+        # script by path.
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install pytest
+          python -m pytest tests/python/test_changelog_grouping.py -v
       - name: Require each Representation changes entry to carry its disclosure
         # The audit below asks whether the right commits are in the section. This asks
         # whether each entry in the *committed* changelog says what it moved (#1556): the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -879,10 +879,16 @@ which is filed as a real change. This was the #1555 defect: the value used to be
 declining *with a reason* was read as a disclosure. It is not a corner case — in the v0.13.1
 cycle 8 of the 14 declines gave a reason, and the section listed 10 entries of which 2 were real.
 
-**One limit of the section itself, and one that is now closed.** A declining commit is filed
-under **Other** whatever its type, so a declining `fix:` is missing from **Fixed** (#1557);
-git-cliff evaluates a parser's fields as OR, not AND, so this cannot be repaired by adding
-`message` to the exclusion rule.
+**Two limits of the section itself, both now closed.** A declining commit used to be filed
+under **Other** whatever its type, so a declining `fix:` was missing from **Fixed** (#1557).
+git-cliff evaluates a parser's fields as OR, not AND, so it could not be repaired by adding
+`message` to the exclusion rule; the fix is a `commit_preprocessor` in `release-plz.toml` that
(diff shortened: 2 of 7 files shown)


nh13:
@coderabbitai pause

PR #2099, opened 2026-08-16
ci: run the census-instrument tests off the censuses job, not the Test shards

### LINKED ISSUE #2094: ci: `conformance_census_instrument` runs full spec censuses in the `Test` shards (#2063 enrolled it in ORACLE_EXCLUDE but not CENSUS_FILTER)
## Symptom

`Test (1/6)` is the wall of the `Test` matrix at ~277s while carrying almost no work. On run [31973592268 / job 95229867903](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/31973592268/job/95229867903) its nextest summary is `[ 277.639s] 1887 tests run`, but the distribution is two tests:

| milestone | elapsed | what it is |
|---|---|---|
| 1,885 of 1,887 tests done | **~72s** | everything else |
| `conformance_census_instrument::every_divergent_family_reaches_the_export` | **207.399s** | one full `Partition` census |
| `conformance_census_instrument::a_looser_relation_converges_at_least_as_many_families` | **271.331s** | a `Sequence` **and** a `Partition` census, back-to-back |

The two run concurrently at the tail, so `a_looser…` at 271s *is* the shard's wall clock — ~74% of a ~5-minute job is two tests.

## Root cause

The repo already carves slow census modules out of the 6-way `Test` shards: `CENSUS_FILTER` (negated from every shard, `ci.yml:1501`, and from `test-oracle`, `ci.yml:1746`) plus the dedicated `censuses` job that runs them off the optimized soak archive. Enrolling a module is a three-point edit:

1. `#[path]`-include it in the soak driver `tests-soak/tests/soak/main.rs` (so it is in the archive the `censuses` job reads);
2. name it in `CENSUS_FILTER` (so the shards negate it);
3. select it in a `censuses` step.

`tests/it/conformance_census_instrument.rs` — added by #2063 (`eb0bad1c`) — is in **none** of the three. Its only appearance in `ci.yml` is inside `ORACLE_EXCLUDE` (`ci.yml:275`), which keeps it off the *armed-oracle* `test-oracle` shards only, not the plain `Test` shards. So #2063 handled the oracle-safety half and missed the cost-carve-out half. The module therefore runs in full in all six shards, and the name-hash landed both of its 60s+ tests in shard 1.

`CENSUS_FILTER` today (`ci.yml:141`): `issue_1542_direction_symmetry + normalize_axis_preserving + spec_conformance_axis + clinvar_hgvs_tests` — no `conformance_census_instrument`.

## Why no guard caught it

The three liveness invariants — `census_filter_invariant.rs`, `soak_package_membership.rs`, `sweep_filter_invariant.rs` — only check *agreement* among the enrollment points (filter ⟺ censuses selection ⟺ soak driver). A brand-new heavy census enrolled in *none* of them is invisible to all three: it simply runs in the shards, and the only signal is a slow shard. Nothing asserts "a test that calls `run_census` / `CensusReport::build` must be in `CENSUS_FILTER`."

## Secondary observations

- **Redundant builds.** Within the module, `Sequence` is built by both `a_spec_census_run_…` and `a_looser…`; `Partition` by both `every_divergent…` and `a_looser…`. Five full census builds for three distinct censuses (`Sequence`, `Partition`, `Spdi`). Sharing each build via a `LazyLock` keyed on `(source, direction, equivalence)` collapses that to three and makes the other tests instant assertions.
- **`~seconds` vs 207s.** The module's own doc comment says "each full census run normalizes the whole corpus (~seconds)", but `Sequence` is <60s and `Partition` alone is 207s. That 30-100x gap (and the `Sequence`↔`Partition` asymmetry) is worth a separate profiling pass; if the census were actually seconds the carve-out would be unnecessary.

## Suggested resolution

1. Enroll `conformance_census_instrument` in the census carve-out, mirroring the four existing modules: `#[path]`-include it in the soak driver, add `test(conformance_census_instrument)` to `CENSUS_FILTER`, and select it in the censuses job's **un-armed** step (`ci.yml:2053-2055`, alongside `spec_conformance_axis` — it must be un-armed for the same oracle reason it is in `ORACLE_EXCLUDE`). This removes ~205s from the shards. The `census_filter_invariant` and `soak_package_membership` guards then enforce all three points stay in sync.
2. Deduplicate the census builds via a shared `LazyLock`, halving the module's own wall wherever it runs.

Follow-ups worth their own issues: the 207s-vs-"seconds" perf question, and a guard that fails when a `run_census`/`CensusReport::build` caller is not in `CENSUS_FILTER`.

## Representation stability

None — this is a CI-topology and test-structure change. No `src/normalize|hgvs|spdi|project` file changes; normalized output is untouched.

Measured against `origin/main` at `fc6bee2c`.

nh13:
Closes #2094.

## The defect

`conformance_census_instrument` (#2063) holds four tests that each run a full spec-corpus census via `run_census` / `CensusReport::build`, costing 60–271s apiece. The module was enrolled only in `ORACLE_EXCLUDE` (which keeps it off the armed `test-oracle` shards); it was **not** named in `CENSUS_FILTER`, the mechanism that negates slow census modules from the 6-way `Test` shards and moves them to the dedicated `censuses` job. So the module ran in full in every shard.

On [run 31973592268 / `Test (1/6)`](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/31973592268/job/95229867903) the name-hash landed both 60s+ tests in shard 1:

| milestone | elapsed |
|---|---|
| 1,885 of 1,887 tests done | **~72s** |
| `every_divergent_family_reaches_the_export` (one `Partition` census) | **207.399s** |
| `a_looser_relation_converges_at_least_as_many_families` (`Sequence`+`Partition`, back-to-back) | **271.331s** |

The two run concurrently at the tail, so `a_looser…` at 271s *is* the shard's wall clock — three quarters of a ~5-minute job. `--partition hash:K/N` balances shard *counts*, not cost, and a shard's wall can never fall below its slowest single test, so re-sharding cannot fix it.

## The fix

Split the four full-census tests into a new module `conformance_census_runs` and enroll it in the census carve-out, matching the four modules already there (`spec_conformance_axis`, `issue_1542_direction_symmetry`, `normalize_axis_preserving`, `clinvar_hgvs_tests`):

- `#[path]`-included in the soak driver `tests-soak/tests/soak/main.rs`;
- named in `CENSUS_FILTER` (negated from `test` and `test-oracle`);
- selected by the `censuses` job's **un-armed** step, beside `spec_conformance_axis` — `run_census` refuses with a seam oracle armed, so it must not be in the armed step;
- named in `ORACLE_EXCLUDE` too, because it imports `conformance::census` (required by `oracle_exclude_invariant`).

This removes ~205s from the shards. The three CI liveness guards (`census_filter_invariant`, `oracle_exclude_invariant`, `soak_package_membership`) now enforce the enrollment stays consistent.

The moved tests are **otherwise unchanged** — they call `run_census` / `CensusReport::build` directly, as before. (Deduplicating the three distinct censuses across tests via a shared `LazyLock` was tried and dropped: nextest runs each test in its own process, so a shared cache is never reused, and `a_looser…` genuinely needs both a `sequence` and a `partition` census in one process regardless. Measured: `a_looser…` stayed at ~282s with the `LazyLock` in place, so it bought nothing.)

## Unchanged on purpose

- The **pure** tests (`compare_reports`, the burn-down, the relation-change and oracle-detector rules) stay in `conformance_census_instrument` — they use hand-built inputs and cost milliseconds, so they belong in the shards.
- The `#[cfg(feature = "benchmark")]` refusal test `a_seam_oracle_in_the_environment_is_refused` stays there too, and deliberately: it needs `CARGO_BIN_EXE_ferro-benchmark`, and the soak driver is built `-p ferro-hgvs-soak-tests` with default features (no `benchmark`), so it cannot run in the `censuses` job. Carving the whole module out would have stranded it — it would run nowhere. Keeping it in the shards keeps it with the binary it needs.

## Representation stability

None — this is a CI-topology and test-structure change. No `src/normalize|hgvs|spdi|project` file changes; normalized output is untouched. The moved tests assert the same properties over the same (deterministic) censuses.

## Verification

```
cargo test --features dev --test it --no-run                         # IT_EXIT=0
cargo build -p ferro-hgvs-soak-tests --tests                         # SOAK_EXIT=0 (census_runs compiles under default features)
cargo nextest run --features dev --test it \
  -E 'test(census_filter_invariant) + test(oracle_exclude_invariant) + test(soak_package_membership) + test(conformance_census_instrument)'
                                                                      # 21 passed
cargo nextest run --features dev --test it -E 'test(conformance_census_runs)'   # 4 passed (4 slow)
scripts/run_oracle_suite.sh --print-selection                        # conformance_census_runs is negated (armed list resolves to 0)
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,6 +141,7 @@ env:
   CENSUS_FILTER: >-
     test(issue_1542_direction_symmetry) + test(normalize_axis_preserving)
     + test(spec_conformance_axis) + test(clinvar_hgvs_tests)
+    + test(conformance_census_runs)
 
   # Negated from `test` ONLY. These run in `test-oracle` with the three seam
   # oracles armed, and an armed run strictly subsumes an un-armed one — the
@@ -273,6 +274,7 @@ env:
     + test(defect_non_idempotent_outputs)
     + test(issue_1715_rna_alignment_symbol_reach)
     + test(conformance_census_instrument)
+    + test(conformance_census_runs)
 
 # Every job carries a `timeout-minutes`. Without one a job inherits GitHub's
 # 360-minute default, so a wedged runner holds a PR for six hours: observed on
@@ -2045,14 +2047,16 @@ jobs:
           scripts/run_nextest_archive.sh --archive-file nextest-soak.tar.zst --workspace-remap . \
             --extract-to . --extract-overwrite --no-tests=fail \
             -E "test(issue_1542_direction_symmetry) + test(normalize_axis_preserving) + test(clinvar_hgvs_tests)"
-      - name: Run the census that may not be measured with an oracle armed
-        # No `FERRO_ASSERT_*` here, deliberately, and no bulk fixtures: this
-        # module builds its corpus in code (`conformance::spec_corpus`) and
-        # reads nothing from disk.
+      - name: Run the censuses that may not be measured with an oracle armed
+        # No `FERRO_ASSERT_*` here, deliberately, and no bulk fixtures: these
+        # modules build their corpus in code (`conformance::spec_corpus`) and
+        # read nothing from disk. `conformance_census_runs` (#2094) joins
+        # `spec_conformance_axis` here because `run_census` refuses outright
+        # with a seam oracle armed, which would fail the armed step below.
         run: |
           scripts/run_nextest_archive.sh --archive-file nextest-soak.tar.zst --workspace-remap . \
             --extract-to . --extract-overwrite --no-tests=fail \
-            -E "test(spec_conformance_axis)"
+            -E "test(spec_conformance_axis) + test(conformance_census_runs)"
 
   # The generated-doc `--check` gates. Split out of the old monolithic Test job
   # so they no longer sit behind the whole suite on the critical path.
--- a/tests-soak/tests/soak/main.rs
+++ b/tests-soak/tests/soak/main.rs
@@ -160,6 +160,12 @@ mod issue_1542_direction_symmetry;
 mod normalize_axis_preserving;
 #[path = "../../../tests/it/spec_conformance_axis.rs"]
 mod spec_conformance_axis;
+// The run-backed census instrument tests (#2094), carved out of
+// `conformance_census_instrument` so the `censuses` job owns them instead of the
+// `Test` shards. Selected by the job's UN-armed step, beside `spec_conformance_axis`
+// and for the same reason: `run_census` refuses with a seam oracle armed.
+#[path = "../../../tests/it/conformance_census_runs.rs"]
+mod conformance_census_runs;
 // A parser benchmark rather than a normalization census — it never reaches
 // `canonicalize_from_sequence` — which is exactly why it belongs here: its cost
 // is a constant factor on gzip decode, JSON deserialization and `parse_hgvs`,
--- a/tests/it/conformance_census_instrument.rs
+++ b/tests/it/conformance_census_instrument.rs
@@ -12,31 +12,36 @@
 //! that the two agree is unnecessary: the gate's own `assert_eq!(census, pinned)`
 //! is that cross-check.
 //!
-//! Each full census run normalizes the whole corpus (~seconds), so the tests that
-//! need a real run are kept few and the properties that are pure — the burn-down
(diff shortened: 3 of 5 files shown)


nh13:
@coderabbitai pause

PR #2100, opened 2026-08-16
Delete two false figures from a decided ruling record and repoint them

### LINKED ISSUE #1828: Two stale figures inside a now-decided ruling record, one of them the third drift of a count the same sentence forbids restating
Two figures in `delins-recommendation-reach-when-the-input-arrives-split`'s rationale are false. Both are carried verbatim from earlier revisions rather than introduced by any one change, but #1760 promotes that record from `undecided` to `decided`, so they are now inside a record people will cite.

## 1. The RFC 2119 census says "exactly one place" — it is two

The rationale reads:

> neither is normative in the RFC 2119 sense — a census of `docs/recommendations/` finds an uppercase keyword in exactly one place outside `style.md`

There are **two**, both in one file: `RNA/adjoined_transcript.md:20` (`REQUIRES`) and `:21` (`SHOULD`).

The cause is worth recording because it will recur: `REQUIRES` is missing from the RFC 2119 alternation everyone copies, since RFC 2119 itself lists `REQUIRED`. A grep built from the usual alternation cannot see `:20`.

`CLAUDE.md` corrected this on 2026-08-10 and says in as many words: **"Do not restate 'exactly once' anywhere."** The record predates that correction and was not swept.

Neither hit generalises, so the record's *argument* survives — `:20` constrains one syntax and `:21`'s SHOULD is assay-scoped. Only the count is wrong.

## 2. The record contradicts itself on the 3' divergent-class count

Within the same rationale:

- it says the authority constants **"assert 3,245 and 3,249"**, and
- it separately quotes **"divergent classes fall from 2,911 under `live`"**

Those are the same quantity with two values. Measured against the constants themselves:

```
THREE_PRIME: 11,272 − 8,361 = 2,911
FIVE_PRIME:  11,272 − 8,367 = 2,905
```

So **2,911 is right** and 3,245 / 3,249 is stale.

The sharp part: the sentence carrying the error is the one whose entire purpose is to say *do not restate this count, read it off the constants* — and it documents two prior drifts of this very number. **This is the third.** A rule that says "do not restate X" and then restates X is the failure mode this repo names as its own recurring one, caught inside the instrument built to prevent it.

## Also, one level out

`tests/it/clause_ruling_index.rs`'s header still claims its branch delta is `c-and-n-positions-are-flat-transcript-offsets`, which has since landed on `main`. That paragraph documents three prior rots of itself; this is the fourth. Pre-existing, untouched by #1760.

## Suggested fix

Delete both restated figures rather than correcting them. The record already names the authorities to read them from (`it::cis_confluence_axis::{THREE_PRIME,FIVE_PRIME}` for the divergence count; the grep for the keyword census, with `REQUIRES` added to its alternation). A corrected number is a number that can drift again; a pointer cannot.

nh13:
Deletes two demonstrably-false figures from the ruling record `delins-recommendation-reach-when-the-input-arrives-split` and repoints each at its live authority rather than a fresh hard-coded number, per the operator ruling on the issue (option a: delete, do not re-derive).

## The two figures

1. **The RFC 2119 census.** The rationale claimed "a census of `docs/recommendations/` finds an uppercase keyword in exactly one place outside `style.md`". There are two, both in `RNA/adjoined_transcript.md` (`:20` `REQUIRES`, `:21` `SHOULD`) — the first invisible to any grep whose RFC 2119 alternation omits `REQUIRES` (RFC 2119 itself lists only `REQUIRED`). The false count is deleted and replaced with a pointer to the census procedure, `REQUIRES` added and the count deliberately not restated.

2. **The 3'/5' divergence count.** The rationale asserted the authority constants "assert 3,245 and 3,249" (and an earlier drift, "3,246 at 3' and 3,251 at 5'"). Those constants — `it::cis_confluence_axis::{THREE_PRIME,FIVE_PRIME}` — do not assert those values, and have since drifted again (the shipped default now leaves `classes - converged = 2`). This is exactly why the record's own preceding sentence says to read the count off the constants; the restated numbers are deleted, keeping only that pointer.

The correctly-attributed historical `2,911 under \`live\`` measurement elsewhere in the record is a different, partition-default figure (still cited by `THREE_PRIME`'s own docs) and is left untouched.

## Also

The delta header in `tests/it/clause_ruling_index.rs` named `coding-axis-merges-are-a-disclosed-general-34-deviation` as "what this branch adds", but that record landed on `main` via #1899 — the fifth instance of the same rot the paragraph documents. The header now stops naming a delta and points to `git diff origin/main`, so it cannot rot again.

`docs/NORMALIZATION_CONTRACT.md` was re-blessed against the edited ledger; `normalization_contract_doc::the_published_document_is_current` passes without the bless env var.

No ruling `id`, `status`, or `guard` was touched — only rationale/header prose.

## Verification

`cargo fmt --check`, `cargo clippy --all-features --all-targets -- -D warnings`, `cargo nextest run --features dev --no-fail-fast` (11321 passed), and `scripts/run_oracle_suite.sh` (11152 passed) all green.

Representation-Change: none. Fixtures/docs/test-comment prose only; no watched-prefix (`src/normalize|hgvs|spdi|project|reference|error_handling`) code path is touched, so no normalized output can move.

Closes #1828


nh13:
@coderabbitai pause

nh13:
Independent review (read-only; nothing pushed or enqueued). **The two deletions are correct** — I verified the ledger record set is unchanged (35 records, same ids, same statuses; nothing dropped in either direction) and that both figures really are false against the thing they describe, not merely against this PR's assertion.

**One finding, and it is a judgment call rather than a defect.**

The sentence being edited retains a **third** restated figure:

> "several thousand classes reach more than one normalized output in each direction"

That is a magnitude claim of exactly the kind the next clause forbids restating, and it is false by the same measurement that condemned the other two. Live value, from running `cis_confluence_axis` (4/4 pass):

| direction | restated | measured |
|---|---|---|
| 3' | "several thousand" | **2** |
| 5' | "several thousand" | **1** |

So the PR removes two false figures and leaves a third standing in the same sentence, which is a strange resting place — a reader has no way to tell the surviving number is any less reliable than the two that were removed, and the whole point of the deletions is that restated counts drift.

**Suggested:** delete the magnitude too (the clause reads fine as a qualitative statement), then re-bless:

```
BLESS_CONTRACT_DOC=1 cargo nextest run --features dev --test it -E 'test(normalization_contract_doc)'
```

Your call — if you want to keep it scoped to the two figures the ruling explicitly names, say so and I will file the third separately rather than hold this up. I am flagging it because leaving it looks like a deliberate judgment that it is true, and it is not.

## Sequencing note

#2107 is in flight and also edits the ledger, `docs/NORMALIZATION_CONTRACT.md` and `tests/it/clause_ruling_index.rs`. The combined merge is textually clean — 35 records, header `**35 adjudication records — 32 decided, 3 open.**`, both edits present, nothing dropped. But whichever lands second will likely need the contract doc re-blessed, and that test is a real gate (verified by sabotage), so a stale render **ejects from the merge queue** rather than slipping through.

PR #2108, opened 2026-08-16
perf(project): share the projector's transcript cache with its normalizer

### LINKED ISSUE #1860: projector's transcript cache does not reach the normalizer, so each re-derived axis re-fetches

nh13:
## Summary

`VariantProjector` caches transcript fetches for its own projection-side derivations (`cached_get_transcript_for_variant`), but the `Normalizer` it holds fetched transcripts from a provider view *beside* that cache. So any axis the projector re-derives by handing a variant to that normalizer re-paid the transcript fetch — once per variant — however warm the projector's own cache was. `reanchor_and_normalize_genomic` and `project_to_genomic_normalized` have routed a normalization per projected variant since #737/#867, so the gap is already latent on the genomic axis; #1712 makes it newly visible on a guard that counts.

Measured on `origin/main`: self-projecting three coding substitutions on one transcript (each drives one normalizer re-derivation) issues **19** `get_transcript` calls.

## Fix

Lift the projector's transcript cache behind an `Arc` and hand the normalizer a `SharedTranscriptCacheProvider` decorator that reads and writes that same map, so one shared transcript cache sits below both sides (the first option the issue weighs). The decorator caches only the three transcript-fetch methods, keyed by the same parent-aware `(transcript_id, parent_accession)` identity the projector uses (#332), and delegates every other method to the wrapped provider verbatim so build-aware placement, sequence/protein reads and the version/gene-selector probes keep the inner provider's behaviour. The three-variant case collapses from **19 → 1**.

## Tests

- New `projection::normalizer_shares_the_projectors_transcript_cache` pins the normalizer-side re-derivation at exactly one fetch across three `c.` variants (it read 19 before this change).
- `projection::transcript_cache_collapses_repeat_lookups` (the existing projection-side guard) stays at 1.

Full local gate green: `cargo fmt --check`, `cargo clippy --all-features --all-targets -- -D warnings`, `cargo nextest run --features dev --no-fail-fast` (11328 passed), `scripts/run_oracle_suite.sh` (11159 passed, armed), and the Python projection + core suites (86 passed).

Representation-Change: none. Caching-only change — the normalizer now shares the projector's transcript cache and is served byte-identical transcript records; no normalization or projection output moves, and the standalone `Normalizer` used by `ferro normalize` is untouched. Confirmed by the full test suite, the seam-oracle suite, and the Python projection suite all passing unchanged.

Closes #1860

### DIFF
--- a/src/project/projector.rs
+++ b/src/project/projector.rs
@@ -24,7 +24,7 @@ use crate::project::protein::{
 };
 use crate::project::result::{AxisDeclineReasons, VariantProjection};
 use crate::reference::transcript::Transcript;
-use crate::reference::ReferenceProvider;
+use crate::reference::{GenomicPlacement, ReferenceProvider};
 use std::collections::HashMap;
 use std::sync::{Arc, RwLock};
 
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #2109, opened 2026-08-16
fix(convert): refuse an inverted-CDS record on every mapping arm and direction

### LINKED ISSUE #1894: The inverted-CDS refusal covers only the intronic 3'UTR arm; the exonic arm still answers
#1877 (merged as `4bf4f52b`) added a guard that refuses an inverted-CDS record — one whose `cds_end < cds_start` — in the intronic 3'UTR arm of `CoordinateMapper::genome_pos_to_cds_pos`, because the bare `cds_end - cds_start` there underflowed: a panic in debug, and in release a wrap that `as i64` turned into a negative `base` flagged `utr3` and returned as success.

The ground the change gives for refusing is not about the underflow, though. It is that on an inverted record the 5'UTR and 3'UTR halves overlap, so one transcript position has two names and any answer picks one of them arbitrarily. That is true of **every** arm such a record reaches. Only the intronic 3'UTR arm refuses, because it is the only one that happened to underflow.

Demonstrated on the fixture #1877 itself added. `NM_INVERTED.1` has `cds_start = 50`, `cds_end = 10`, exon 1 spanning genome 1000–1100 ↔ tx 0–100:

- genome **1500** (intronic) → boundary anchor tx 99 → `tx_to_cds` files it `ThreePrimeUtr(90)` → **refused**, as intended.
- genome **1050** (exonic) → tx 50 → `tx_to_cds` files it `ThreePrimeUtr(41)` → the exonic branch calls `CdotTranscript::cds_pos_from_region`, which for `ThreePrimeUtr` returns `base: offset` with no subtraction → returns **`Ok(c.*41)`**.

Same transcript, same incoherence, two different verdicts. The 5'UTR arm answers too, since `tx_pos < cds_start` is satisfiable on an inverted record.

`c.*41` is not a lesser problem than the underflow was — it is a confidently-wrong coordinate rather than a detectably-wrong one.

### Suggested shape

Validate the CDS bounds once, where the transcript is acquired or in a shared mapping helper, ahead of both `genome_to_cds` and `cds_to_genome`, so every path refuses an inverted record rather than some paths refusing and others answering. That also removes the question of which arm a given position happens to land in.

Worth noting for whoever picks this up: widening the guard changes behaviour on records that **today answer without underflowing**, so it is a representation change and owes its own `Representation-Change:` disclosure — which is exactly why it was correctly kept out of #1877, whose scope was the underflow.

### Provenance

Both review lenses on #1877 reached this independently — the inline agent lens and the CodeRabbit CLI lens — and it was rejected for that PR as out of scope rather than disputed.

nh13:
Closes #1894.

`#1877`'s inverted-CDS refusal covered only the intronic 3'UTR arm of `CoordinateMapper::genome_pos_to_cds_pos`, because that arm was the only one that happened to underflow. Every other path answered on the same incoherent record. Demonstrated on the fixture `#1877` itself added, `NM_INVERTED.1` (`cds_start` 50, `cds_end` 10):

- genome **1050** (exonic) -> tx 50 -> `ThreePrimeUtr(41)` -> returned `Ok(c.*41)` — a confidently-wrong coordinate, worse than the underflow it replaced.
- genome **1020** (exonic) -> tx 20 -> `FivePrimeUtr(30)` -> returned `Ok(c.-30)`.
- the whole `cds_to_genome` direction was unguarded as well.

## Fix

Validate the CDS bounds once, ahead of both directions. `reject_inverted_cds` runs at the top of `genome_pos_to_cds_pos` **and** `cds_pos_to_genome_pos`, so every arm and both directions decline an inverted record identically instead of some declining and others fabricating a coordinate. The intronic 3'UTR arm's now-redundant per-site refusal becomes a plain subtraction guarded by a `debug_assert`, keeping the "inverted CDS bounds" rule in exactly one place. On an inverted record the 5'UTR and 3'UTR halves overlap, so one transcript position carries two names — the same ground `merge.rs`'s `ordered_cds_bounds` refuses on.

Also folds in the two sibling sites flagged in the issue thread: `convert/coding.rs`'s `validate_cds_pos` computed `cds_end - cds_start + 1` as a bare `u64` subtraction that underflowed on an inverted record (a debug panic, a release wrap to a vast CDS length that admits nearly any `pos.base`). Its sibling `sequence_length() - cds_end` relation was already made `checked_sub` by `#1909`.

## Tests

Failing-first, then implemented. New guards pin the exonic arm, the exonic 5'UTR arm, and the `cds_to_genome` direction in `mapping.rs`, plus `validate_cds_pos` in `coding.rs`; the two pre-existing inverted-CDS tests (`#1877`) still pass because the top-level guard emits the same `inverted CDS bounds` message. Each asserts the refusal **message**, not just `is_err()`, so it pins the inverted-bounds guard rather than any `InvalidCoordinates` the call can otherwise reach.

Full local gate green: `cargo fmt --check`, `cargo clippy --all-features --all-targets -- -D warnings`, and `cargo nextest run --features dev --no-fail-fast` (11331 passed, 152 skipped).

Representation-Change: none. No watched directory is touched — the changed files are `src/data/mapping.rs` and `src/convert/coding.rs`, neither under `src/normalize/`, `src/hgvs/`, `src/spdi/`, `src/project/`, `src/reference/` or `src/error_handling/`. The only behavioural change is that a malformed inverted-CDS transcript (`cds_end < cds_start`) is now refused on the exonic and 5'UTR arms and in the `cds_to_genome` direction, where it previously answered with a fabricated coordinate. Real cdot never ships inverted CDS bounds and the synthetic normalization corpus builds none (its transcript generator uses a single exon with an ordered CDS), so this is a structural zero for normalized output — the effect is confined to conversion/projection of a malformed record.

### DIFF
--- a/src/convert/coding.rs
+++ b/src/convert/coding.rs
@@ -26,6 +26,24 @@ pub fn validate_cds_pos(pos: &CdsPos, transcript: &Transcript) -> Result<(), Fer
             msg: "Transcript has no CDS end".to_string(),
         })?;
 
+    // Refuse an inverted record (`cds_end < cds_start`) before computing a CDS
+    // length from it. The bare `cds_end - cds_start` below is on two `u64`s, so
+    // an inverted record underflows it: a panic in debug, and in release a wrap
+    // to a vast `cds_length` that admits nearly any `pos.base` as in-range. More
+    // fundamentally, an inverted record's 5'UTR and 3'UTR halves overlap, so no
+    // CDS-relative position is well-defined (#1894). `CoordinateMapper` refuses
+    // the same condition up front for the same reason, as does `merge.rs`'s
+    // `ordered_cds_bounds`.
+    if cds_end < cds_start {
+        return Err(FerroError::InvalidCoordinates {
+            msg: format!(
+                "transcript has inverted CDS bounds (cds_start {cds_start} > \
+                 cds_end {cds_end}); its c. axis is incoherent, so a CDS position \
+                 cannot be validated"
+            ),
+        });
+    }
+
     let cds_length = (cds_end - cds_start + 1) as i64;
 
     if pos.utr3 {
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #2110, opened 2026-08-16
fix(explain): resolve the normalizer's SCREAMING_SNAKE warning codes

### LINKED ISSUE #2092: ferro explain does not resolve the normalizer's SCREAMING_SNAKE warning codes
`ferro explain <CODE>` cannot resolve the normalizer's SCREAMING_SNAKE warning codes, only the preprocessor's `W`/`E`-numbered ones.

`get_code_info` (`src/error_handling/registry.rs:2130`) is a plain uppercased map lookup against the registry. Measured against `origin/main`, the registry contains **0** occurrences of each of:

- `MEMBERS_COALESCED_FROM_REPORTED_FORM`
- `INSERTED_SEQUENCE_EXPANDED`
- `REFSEQ_MISMATCH`

Positive control, so this is a real absence rather than a broken search: `W3003` → **4** hits.

`src/bin/ferro.rs:1824-1827` documents the two namespaces explicitly — preprocessor codes are `W`-numbered, normalizer codes are SCREAMING_SNAKE — so the split is intentional. What is missing is that `explain` covers only half of what the tool emits, while nothing says so. A user who sees `warning[MEMBERS_COALESCED_FROM_REPORTED_FORM]:` in normalize output and runs `ferro explain MEMBERS_COALESCED_FROM_REPORTED_FORM` gets nothing useful back.

## Why this is filed separately

It surfaced while reviewing #2043, whose quickstart tells readers to run `ferro explain <CODE>` for exactly these three codes. That documentation error is #2043's to fix and has been reported on the PR. **This issue is the underlying `src` gap**, which exists on `main` today independent of any docs.

## Two possible fixes

1. **Register the normalizer codes** in `src/error_handling/registry.rs` so `explain` resolves the full set a user can actually observe in output. This is the better outcome — the tool's own emitted codes should all be explainable.
2. **Scope `explain` explicitly**: have it say which namespace it covers, and return a useful message naming the other namespace when handed a SCREAMING_SNAKE code it does not know, rather than a generic miss.

Whichever is chosen, a guard is worth adding: every code the normalizer can emit should be resolvable by `explain`, or explicitly listed as out of scope. Deriving that set from source rather than a hardcoded list is what keeps it from going stale — a hardcoded list is blind to the next code someone adds.

nh13:
Closes #2092

Representation-Change: none. No watched file changes what ferro emits as a normalized description. The change is confined to `ferro explain`'s lookup and output, one new registry entry (`W5005`) and one `consults_error_config` declaration — none of which is read on any normalization path. Confirmed by the full suite: 11342 tests pass with no census, pin or corpus figure moving.

## Root cause

`ferro normalize` prints two code namespaces to stderr in one shape, `warning[<CODE>]:` — the preprocessor's `W`-numbered codes and the normalizer's SCREAMING_SNAKE names — and `src/bin/ferro.rs` documents the split at the emission site. `get_code_info` was a plain uppercased lookup against a registry holding only the first namespace, so `ferro explain MEMBERS_COALESCED_FROM_REPORTED_FORM` returned `Unknown code`, which reads exactly like typing a code that does not exist.

Measured on `origin/main` before any change, all four exiting 1:

```
$ ferro explain MEMBERS_COALESCED_FROM_REPORTED_FORM
Error: "Unknown code: MEMBERS_COALESCED_FROM_REPORTED_FORM. Use 'ferro explain --list' to see all codes."
$ ferro explain INSERTED_SEQUENCE_EXPANDED   # same
$ ferro explain REFSEQ_MISMATCH              # same
$ ferro explain W5005                        # same  <- not in the issue; see below
$ ferro explain W3003                        # exit 0, full entry (positive control)
```

The full set is **fifteen** codes, not the three the issue names: fourteen `NormalizationWarning` variants plus `NormalizationInfo::ShuffleApplied`. Fixing three would have reproduced the defect at smaller scale.

## Which of the two approaches, and why

**Approach 1 — populate the registry** — with the content sourced rather than invented, which is what made it the better option here.

Nine of the fifteen already had somewhere authoritative to resolve to. They name a condition the registry documents under a `W`-code, and the repository had already recorded that correspondence twice independently: `W5003`'s own explanation says the condition "is also published as `CanonicalSplitSkipped` in the NormalizationWarning enum", and `tests/it/error_code_audit.rs` carried a `NormalizationWarning`-variant-to-registry-name table. So these resolve through an explicit alias table (`NORMALIZER_CODES` in the new `src/error_handling/normalizer_codes.rs`) to the entry that already exists. **No new prose at all for those nine.** `explain` prints a one-line note naming the pairing, so the output is not silently headed by a different code than the one asked for.

The remaining five — `CROSS_AXIS_VARIANT_NOT_SHUFFLED`, `AXIS_CLAMP_APPLIED`, `INSERTED_SEQUENCE_EXPANDED`, `REDUCED_CAPABILITY_NO_GENOME`, `SHUFFLE_APPLIED` — have no `ErrorType` and therefore no `W`-code to alias. They get entries of their own.

Approach 2 (scope `explain` and explain the gap) is included as the *residual* rather than as the answer: whatever `explain` cannot resolve must still be distinguishable from a dead end, which is the same defect one level down. See "The distinguishability requirement" below.

## Authority for every piece of user-facing text

Nothing in this PR states a normalization rule that is not already stated in the tree, and each entry names its source in a comment above it.

| entry | derived from |
|---|---|
| the nine aliased codes | **no new text** — they resolve to the existing `W`-code entry |
| `CROSS_AXIS_VARIANT_NOT_SHUFFLED` | `NormalizationWarning::CrossAxisVariantNotShuffled` doc comment + `Display` arm |
| `AXIS_CLAMP_APPLIED` | `NormalizationWarning::AxisClampApplied` doc comment + `Display` arm |
| `INSERTED_SEQUENCE_EXPANDED` | `NormalizationWarning::InsertedSequenceExpanded` doc comment, field docs + `Display` arm |
| `REDUCED_CAPABILITY_NO_GENOME` | its doc comment + `Display` arm; the mode behaviour from `NormalizeConfig::should_reject_reduced_capability`, which asks `ErrorMode::is_strict` directly |
| `SHUFFLE_APPLIED` | `NormalizationInfo::ShuffleApplied` doc comment + `Display` arm; the mutalyzer correspondence from `error_handling::info_map` |
| `W5005` | `ErrorType::MembersCoalescedFromReportedForm`'s own `description()` / `example()` / `is_correctable()` arms in `types.rs`, plus the `NormalizationWarning::MembersCoalesced` doc comment. Its spec citation, `DNA/delins.md:79-84`, is the one that variant's `Display` already emits |

## Two things found on the way, both fixed here because the fix depends on them

**`W5005` was itself unregistered.** It is minted by `ErrorType::MembersCoalescedFromReportedForm::code()` and named in `types.rs`, `python.rs` and the CLI, but had no registry entry, so `ferro explain W5005` failed for a code ferro documents as its own. Leaving it out would have meant documenting one condition under two codes with only one of them resolvable — the "one rule in several places" failure this repo keeps paying for. It is registered, with rows in `tests/fixtures/error_code_audit.json` and `docs/spec/error_code_audit.md`.

**Registering it exposed that `--ignore W5005` is inert.** `NormalizationWarning::MembersCoalesced` is pushed unconditionally at the single normalization exit that compares cis-member counts; there is no `ErrorConfig` consult site anywhere in `src/`. With `consults_error_config()` returning `true`, an honest `enforced` audit row could not be written — `audit_enforced_warning_rows_reach_a_config_consult_site` would fail, and its own failure message prescribes the fix ("if the code is genuinely not configurable — declare it in `ErrorType::consults_error_config` so the CLI can say so"). Marking the row `partial` to dodge that check would have been using a status meaning "surfaces a subset of violating inputs" to hide a config fact, so the declaration is flipped instead. Consequence: `--ignore W5005` now draws the existing inert-override diagnostic rather than silently doing nothing.

That in turn made `format_terminal`'s "To ignore this warning: `ferro parse --ignore <CODE>`" block wrong for `W5005`, so it is gated on the entry carrying a `ModeBehavior` — which also stops it advertising `--ignore W-LOAD-001` for the eleven loader warnings, where it was already untrue.

## The distinguishability requirement

The issue *is* an instance of "a result that cannot distinguish 'I checked and it's fine' from 'I couldn't check'", so the fix does not stop at making codes resolve:

- **Unresolvable is now differentiated.** `explain NOT_A_FERRO_CODE_AT_ALL` says the code is shaped like a normalizer code but is not one ferro emits, and gives a real example; `explain ZZ999` names both namespaces with an example of each. Both still exit 1.
- **Resolved-to-nothing is guarded.** `every_normalizer_code_resolves_to_substantive_content` fails on an entry with thin prose *and* on one that does not resolve — the `None` arm reports rather than `continue`s, so the test cannot pass vacuously on the pre-fix state.
- **`explain --list` enumerates both namespaces**, so someone browsing rather than asking about one code is not shown half the set.

## Method

The test file was written first and **observed red**: `EXIT=100`, 9 tests run, **2 passed, 7 failed**. `every_normalizer_code_resolves_through_get_code_info` reported *15 of 15 normalizer codes do not resolve*, `explain --list` was missing all 15, and `w5005_resolves_under_its_minted_w_code` failed on "W5005 must be registered".

One of the two that passed was `every_normalizer_code_resolves_to_substantive_content` — passing **vacuously**, because it skipped codes that did not resolve. That was fixed before implementing anything, and it is the mutation-4 row below that proves the repair.

### Why the code set is scanned, not listed

A hardcoded list of normalizer codes in the test would be blind to the next variant added, which is precisely what the issue asks the guard to survive. Two independent scans of `src/normalize/mod.rs` run instead — one over the `code()` match arms, one over the enum *declarations* — and `scanned_code_arms_match_the_declared_variants` requires them to agree. A variant whose arm the pattern cannot parse fails there loudly rather than quietly shrinking the set every other assertion checks.

## Sabotage table

Each mutation applied to the green tree, `cargo nextest run --features dev --test it -E 'test(issue_2092)' --no-fail-fast`, then restored (`git diff --stat` empty afterwards on each touched file).

| # | mutation | result | exit | caught by |
|---|---|---|---|---|
| 1 | drop `MEMBERS_COALESCED_FROM_REPORTED_FORM` from `NORMALIZER_CODES` | 4 passed, **5 failed** | 100 | `every_normalizer_code_resolves_through_get_code_info`, `..._resolves_to_substantive_content`, `the_codes_named_in_issue_2092_...`, `explain_cli_resolves_a_normalizer_code`, `explain_list_enumerates_the_normalizer_namespace` |
| 2 | delete the normalizer fallback from `get_code_info` (restore the pre-fix lookup) | 3 passed, **6 failed** | 100 | the five above minus the `--list` one, plus `normalizer_codes_resolve_case_insensitively` and `explain_cli_distinguishes_an_unresolvable_code_from_a_resolved_one` |
| 3 | point `REFSEQ_MISMATCH` at `W5002` — a **wrong but populated** alias | 8 passed, **1 failed** | 100 | `the_codes_named_in_issue_2092_resolve_to_their_own_entries` |
| 4 | replace `AXIS_CLAMP_APPLIED`'s explanation with `"TBD."` | 8 passed, **1 failed** | 100 | `every_normalizer_code_resolves_to_substantive_content` — *"AXIS_CLAMP_APPLIED: summary 65 chars, explanation 4 chars"* |
| 5 | respell one `code()` arm so the scan pattern misses it (`{ direction: _, .. }`) | 8 passed, **1 failed** | 100 | `scanned_code_arms_match_the_declared_variants` — *"declared but no arm scanned: [\"AxisClampApplied\"]"* |
| 6 | restore the pre-fix undifferentiated "Unknown code" message | 8 passed, **1 failed** | 100 | `explain_cli_distinguishes_an_unresolvable_code_from_a_resolved_one` |

Mutation 3 is the one worth reading: an alias that resolves to a real, fully-populated entry is invisible to every "does it resolve?" assertion, and is caught only because the test pins `info.name` rather than merely `is_some()`. Mutation 5 is what makes the other seven non-vacuous — it proves the scan cannot silently narrow.

Restored tree re-run: 9 passed, `EXIT=0`.

## Gate

Foreground, unpiped, `EXIT=$?` on the next line into a per-command log, read back out of the file.

| command | EXIT | result |
|---|---|---|
| `cargo nextest run --features dev --no-fail-fast` | **0** | 11342 tests run: 11342 passed (4 slow), 152 skipped |
| `cargo clippy --all-features --all-targets -- -D warnings` | **0** | clean (one pre-existing upstream note about `proc-macro-error2`) |
| `cargo fmt --all --check` | **0** | clean |
| `python3 -m pytest tests/python/ -q` | **0** | 1137 passed, 8 skipped |

The Python suite was run against a `maturin develop --features python,extension-module` build in a throwaway venv, since `--features dev` never compiles `src/python.rs`.


nh13:
@coderabbitai pause

PR #2111, opened 2026-08-16
test(spdi): count only error-free spdi_result as successful

### LINKED ISSUE #2017: external API validation failed
The scheduled external-API validation run failed.

- Workflow: `External API Validation`
- Commit: `4d36bed85a63a32bf00b9d0fd05068669cc151cb`
- Run: https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/31916271661

This talks to third-party services (mutalyzer, VariantValidator), so a failure may be an upstream outage rather than a ferro defect. Check the run log before assuming a regression.

_This run was triggered by `schedule`, not by a pull request, so this failure was not visible on a pull request before the commit landed._ One rolling issue is used per failure mode; repeat occurrences are added as comments.

nh13:
Closes #2017

## Problem

The nightly `External API Validation` workflow failed in `spdi_tests::test_hgvs_to_spdi_parsing`:

```
assertion `left == right` failed: hgvs_conversions.successful must equal the number of conversions with an SPDI result
  left: 19
 right: 25
```

This is not an upstream outage — it is a deterministic self-consistency assertion whose two counts disagree.

## Root cause

The test checks that the fetched `ncbi_variation.json`'s `hgvs_conversions.successful` header equals the number of conversions carrying an SPDI result. It counted `spdi_result.is_some()`.

But the fetch harness (`scripts/fetch_ncbi_variation.py`) records a *failed* HGVS→SPDI conversion as a **non-null** `spdi_result` carrying an `error` field (the API's error body is stored inline). The `successful` header is computed as `spdi_result` present **and** no `error` key — so it excludes those. Whenever a live fetch hits any upstream API errors, `successful` (19) is less than the `is_some()` count (25) and the assertion trips.

PR CI never sees this: the committed, hand-curated fixture stores its failures as `spdi_result: null`, which the `is_some()` count and the header agree on. Only a live nightly fetch that encounters errors triggers it.

## Fix

Count conversions whose `spdi_result` is present **and** reports no error, mirroring the header's own definition. This is robust to both ways the harness records a failure — `null` (committed convention) or an error-object (live-fetch convention). Extracted as `count_successful_conversions`, plus a hermetic regression guard (`errored_conversions_are_excluded_from_the_successful_count`) that pins both representations against the counter and asserts the naive `is_some()` count would over-count.

Test-only change; no library behavior changes.

Representation-Change: none

### DIFF
--- a/tests/it/spdi_tests.rs
+++ b/tests/it/spdi_tests.rs
@@ -152,21 +152,17 @@ fn test_hgvs_to_spdi_parsing() {
 
     // The fixture's aggregate header counts must agree with the actual data so
     // they cannot silently drift: `total` is the number of recorded conversions,
-    // and `successful` is the number that carry an SPDI result.
+    // and `successful` is the number that carry a *successful* SPDI result (an
+    // `spdi_result` that is present and reports no error).
     assert_eq!(
         fixture.hgvs_conversions.total,
         fixture.hgvs_conversions.variants.len(),
         "hgvs_conversions.total must equal the number of recorded conversions"
     );
-    let hgvs_with_spdi_result = fixture
-        .hgvs_conversions
-        .variants
-        .iter()
-        .filter(|conversion| conversion.spdi_result.is_some())
-        .count();
+    let hgvs_with_spdi_result = count_successful_conversions(&fixture.hgvs_conversions.variants);
     assert_eq!(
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #2117, opened 2026-08-17
fix: refuse a cis collapse anchored 5' of the sequence's first base

### LINKED ISSUE #2037: A cis allele at c.1 normalizes to c.-1_1insA, a coordinate the transcript does not have
Normalizing a cis allele at the first base of a transcript produces a `c.` coordinate the transcript does not have.

## Reproduction

Surfaced by PR #1974's corpus under `FERRO_ASSERT_IN_BOUNDS=1`, in CI run `31906926813`:

```
input:  NM_TEST.1:c.[1dup;1T>A]
output: NM_TEST.1:c.-1_1insA
```

The panic is raised inside `normalize` itself, at `src/normalize/mod.rs:3182`:

```
FERRO_ASSERT_IN_BOUNDS: normalization produced a coordinate the sequence does not have
```

## What is wrong

Both members sit at `c.1`, the first base of the CDS. Combining them shifts the insertion point 5' past the start of the coordinate space and emits `c.-1`, which on this transcript names no base — the reference has no sequence there to anchor against.

`c.-1` is of course a legal *spelling* in HGVS for a 5'UTR position, which is what makes this quietly dangerous: the output parses, round-trips, and looks well-formed. It is only wrong with respect to the reference it was derived from. A consumer validating by parse alone will not catch it.

## Why it is worth a separate issue

The natural reading of the failure is "the boundary handling is off by one at the CDS start". That may be right, but the interesting question is upstream: whether the cis-combination path should be clamping to the coordinate space at all, or whether it should be refusing to combine when the result would leave it. The existing `base < 1` decline in the coordinate mapper is the precedent for the second reading.

## Scope note

Filed out of #1974's review so it is tracked independently of that PR. #1974 does not introduce this — it reveals it, by adding corpus rows that reach the `FERRO_ASSERT_IN_BOUNDS` oracle, which only the `Test oracle` CI job sets. Fixing the normalizer and landing the corpus family are separate pieces of work.

Refs #1974.

nh13:
Closes #2037

`NM_TEST.1:c.[1dup;1T>A]` on a transcript whose CDS starts at base 1 came back as `NM_TEST.1:c.-1_1insA` — a description that parses, re-parses and is a fixed point while naming a nucleotide the reference does not have. `background/numbering.md:54` is directly on point: it is not allowed to describe variants in nucleotides beyond the boundaries of a transcript reference sequence.

Representation-Change: 96 of 96,182 corpus rows respell, all on the coding axis and all the reproducer shape — `c.[1dup;1X>Y]` moves from `c.-1_1insY` to `c.1delinsYX`. Every replaced string named a coordinate the transcript does not have, so this is a correction of out-of-range output rather than a choice between two valid spellings. A transcript with a real 5'UTR is untouched: the corpus carries 96 sibling rows of the identical input shape on `NM_TESTX.1`, and 0 of them move.

## Root cause

`collapse_overlapping_cis_edits` builds a collapsed group's anchor in dense window arithmetic, so a group netting to a pure insertion at the window's 5' edge produces the insertion-shaped anchor `(del_start, del_start - 1)`, whose `end` is the axis coordinate `0`. #1327 gave that function the mirror-image `past_end` refusal for the **3'** edge and left the 5' side open; #1772 then renamed the endpoint `0` to `-1` so it would stop rendering as `c.?`/`n.0`, and said in its own doc comment that the residue — what such a group should be *called* — was still owed. This is that residue.

`c.-1` is only a real base when the axis has room below `1`. On a transcript whose `cds_start` is 1 it has none, and on `n.`/`r.` and `g.`/`m.` there is no negative zone at all.

Nothing downstream could repair it. Every later pass is gated on the axis's positive body region, so once the collapse has emitted a member starting at `c.-1`, `collect_canonical_edits` refuses the group and `canonicalize_from_sequence` — which already answers this shape correctly on the genomic axis — never runs.

## The fix

Add the 5' counterpart of the `past_end` refusal. It is decided by the axis frame's `delta` alone and so needs no provider round-trip: the axis coordinate `p` sits at sequence position `p + delta`, the same identity `start0` is already computed with.

**Refusing is what fixes it, not clamping.** Handing the original members back lets `canonicalize_from_sequence` run, and its `boundary_delins_anchor` re-spells a payload resting on a sequence bound as the single-position `delins` HGVS can express. So the group is still merged — `c.[1dup;1T>A]` -> `c.1delinsAT` — just by the pass that can place it.

The form is not a free choice. `DNA/insertion.md:95-101` refuses a single-position `ins` as not unequivocal, so a payload resting 5' of base 1 has no two-anchor `ins` spelling at all; the single-position `delins` is the same identity already applied at every other sequence bound — the contig start on `g.` (#1205), the mitochondrial origin (#1217) and the transcript bounds on `n.`/`r.` (#1202). Deriving it from the resulting sequence rather than preserving the input's spelling is `rulings[canonical-form-choice-when-both-legal]`.

No new ruling record: this is not two clauses in tension but a coordinate that names no base, and the boundary-`delins` identity it resolves to is already settled and applied on the sibling axes.

## Observed failing first

With the guard reverted and the tests kept, 6 tests fail, reproducing the issue byte-for-byte:

```
issue_2037_..::the_allele_at_the_first_coding_base_names_only_coordinates_the_transcript_has
  `NM_TEST.1:c.[1dup;1T>A]` -> `NM_TEST.1:c.-1_1insA` names a coordinate NM_TEST.1
  does not have (cds_start = 1, so there is no base 5' of c.1):
  ["... transcript position from c. coordinate -1 resolves to a non-positive base (0)"]

issue_2037_..::the_canonical_form_is_the_single_position_delins_at_the_first_base
  left: "NM_TEST.1:c.-1_1insA"   right: "NM_TEST.1:c.1delinsAT"
```

The three controls (`a_transcript_with_a_real_five_prime_utr_still_spells_it_at_c_minus_one`, `an_interior_pair_on_the_same_transcript_still_merges`, `the_in_bounds_instrument_tells_the_two_transcripts_apart`) pass in that same reverted run, which is what makes them controls rather than more of the same assertion.

## Sabotage

| mutation | result | caught by |
|---|---|---|
| `v < 1` -> `v < 0` (off-by-one loosening) | 6 fail | the four `issue_2037` bounds/form rows, `issue_1796::the_five_prime_edge_cis_collapse_is_unaffected`, `merge::tests::a_five_prime_edge_anchor_off_the_sequence_is_refused` |
| drop the `a_end` arm, keep only `a_start` | 6 fail | same six — `a_start` can never trip, so this is equivalent to no guard |
| drop the `+ delta` term, guard the **axis** coordinate instead of the sequence coordinate (over-fix) | 3 fail | `a_transcript_with_a_real_five_prime_utr_still_spells_it_at_c_minus_one`, plus the `c.`/`r.` rows of `a_five_prime_edge_insertion_anchors_below_one_not_on_zero` and `issue_1796::the_five_prime_edge_cis_collapse_is_unaffected` |

The third row is the one worth reading: it is the difference between a bounds check and a blanket ban on `c.-1`, and only the over-fix control sees it.

## Measured reach, including what the corpus cannot see

`examples/dump_normalized_corpus.rs`, same generator both sides: 96 of 96,182 rows move, all in `coincident_bounds`, across 4 distinct before/after pairs, all of the form `c.[1dup;1X>Y]`: `c.-1_1insY` -> `c.1delinsYX`.

The corpus's axis set is `g`, `c`, `cx`, `p`. Two zeros in it are **structural**, not evidence:

- **No `n.`/`r.` axis exists in this corpus at all**, so it cannot witness that half of the fix. That is covered by the dedicated tests instead (`the_non_coding_axis_has_the_same_shape_and_the_same_answer`, the re-pinned `n.` row in `issue_1796_base_zero_names_no_nucleotide`, and the new `merge::tests` unit test).
- **The `g` axis has 1,536 `coincident_bounds` rows and none names a sub-1 anchor** — the corpus never places this shape at contig position 1, so the genomic half is likewise unmeasured here.

What the corpus does establish, and what a hand-written test could not: the 192 rows carrying the reproducer input shape split exactly 96 on `NM_TEST.1` (no 5'UTR, all move) and 96 on `NM_TESTX.1` (real 5'UTR, none move). The over-fix control is measured, not merely asserted.

## Gate

`cargo nextest run --features dev --no-fail-fast` 11336 passed / 0 failed; `cargo clippy --all-features --all-targets -- -D warnings` clean; `cargo fmt --all --check` clean; `pytest tests/python/ -q` 1137 passed, 8 skipped.

### DIFF
--- a/src/normalize/merge.rs
+++ b/src/normalize/merge.rs
@@ -717,6 +717,37 @@ pub(crate) fn collapse_overlapping_cis_edits<P: ReferenceProvider>(
             return variants;
         }
     }
+    // The 5' mirror of the refusal above (#2037), and the reason it is a
+    // separate statement rather than a second arm of it: this one needs no
+    // provider round-trip. Whether an endpoint falls *before* the first base is
+    // decided by `delta` alone, since the axis coordinate `p` sits at sequence
+    // position `p + delta` — the same identity `start0` is computed with.
+    //
+    // The case is the exact counterpart of #1327's: when the group nets to a
+    // pure insertion at the window's **5'** edge the anchor is
+    // `(del_start, del_start - 1)` with `del_start == w_lo == 1`, so `a_end` is
+    // the axis coordinate `0`. Where the axis has a 5'UTR that is a real base
+    // and `c.-1` is the right spelling; where `cds_start == 1` — or on `n.`/`r.`
+    // and `g.`/`m.`, which have no negative zone at all — it is not, and
+    // `name_on_zeroless_axis` renders it `c.-1`/`n.-1` regardless. The result
+    // parses, re-parses and is a fixed point while naming a nucleotide the
+    // reference does not have.
+    //
+    // **Refusing is what fixes it, not clamping.** Nothing downstream could
+    // repair the collapsed member, because every later pass is gated on the
+    // axis's positive body region and a member starting at `c.-1` is outside
+    // it — so `collect_canonical_edits` refuses the group and
+    // `canonicalize_from_sequence` never runs. Handing the *original* members
+    // back lets that pass run, and it already answers this shape correctly on
+    // every axis: `boundary_delins_anchor` re-spells a payload resting on a
+    // sequence bound as the single-position `delins` HGVS can express
+    // (`DNA/insertion.md:95-101`; #1205, #1217, #1202). So the group is still
+    // merged — `c.[1dup;1T>A]` -> `c.1delinsAT` — just by the pass that can
+    // place it.
+    let before_first_base = |p: i64| p.checked_add(delta).is_none_or(|v| v < 1);
+    if before_first_base(a_start) || before_first_base(a_end) {
+        return variants;
+    }
     let anchor = Anchor {
         region: body,
         start: a_start,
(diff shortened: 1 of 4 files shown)


nh13:
@coderabbitai pause

PR #2119, opened 2026-08-17
fix(project): name the build a placement is on when declining a build mismatch

### LINKED ISSUE #2103: LRG build-mismatch decline: the message misstates the cause, and the GRCh37 mapping is discarded unread
Split off from the #2089 review. Two related defects in the same decline path, both newly reachable at scale after #2089 makes 1294 LRG records decline a GRCh37 request.

## 1. The message says "no placement is known" when a placement IS known

`projector.rs:2354ff` renders *"no chromosomal placement is known for this NG_/LRG_ reference"*. After #2089 that is false for the new case: a placement exists, on a different build. The message conflates **unplaced** with **placed elsewhere**, and the two want different user actions — the first is a data gap, the second is "re-run with `--assembly GRCh38`".

Split the causes so the message can name which one occurred.

## 2. ferro discards the GRCh37 mapping it already has

`parse_lrg_main_assembly_placement` reads only `type="main_assembly"`. Every LRG XML also carries `type="other_assembly"` with the GRCh37 mapping — `NC_000017.10` appears in the repo's own fixture at `:4230`. So ferro parses straight past the data that would let it **answer** the GRCh37 request instead of declining it.

Declining is the right immediate fix and #2089 is right to make it. Serving the answer is the better eventual one: it would turn 1294 declines into 1294 correct answers.

The parser's existing single-span and `<diff>` handling would need extending; the fixture already contains the `other_assembly` shape.

## Note on counts

#2089's body and #1903 both say **1293** LRG records; a full `LRG_1..LRG_1500` sweep measured **1294** with placements, matching the repo's own comment at `provider.rs:330` ("all 1 294 LRG records"). Off by one.

The related "241 of 1293 have numerically overlapping GRCh37/GRCh38 spans" figure is **unverified** — checking it requires the `other_assembly` spans that ferro currently discards, i.e. it cannot be confirmed until item 2 is done.

nh13:
## Summary

Split off from the #2089 review. #2089 made 1294 LRG records decline an explicit `GRCh37` projection request (an LRG/NG carries a single GRCh38 main-assembly placement), which newly exercised a latent defect in the decline path.

When `genomic_placement_on_build` returns `None`, `reanchor_genome_output` declined every `NG_`/`LRG_` parent with the same message — *"no chromosomal placement is known for this NG_/LRG_ reference"*. After #2089 that message is wrong for the build-mismatch case: a placement **is** known, just on a different build than the one requested. The message conflated two causes that want different user actions — an **unplaced** parent is a data gap, a **placed-elsewhere** parent is answerable by re-running under the build the placement carries.

## Fix

In the `NG_`/`LRG_` decline arm, when a build was requested, look the placement up build-agnostically (`build = None`) to detect whether one exists on any build:

- **Placement exists on another build** → decline naming the build it is on, its `NC_` accession, and pointing at `--assembly <build>` (e.g. *"its chromosomal placement is on GRCh38 (NC_000017.11), not GRCh37; re-run with --assembly GRCh38"*).
- **No placement on any build** → the unchanged *"no chromosomal placement is known"* message.

The extra lookup runs only inside the (already-declining) error branch, so no successful projection changes.

## Scope note on the issue's second item

Issue #2103 has a second item — the parser reads only `type="main_assembly"` and discards each LRG XML's `type="other_assembly"` (GRCh37) mapping, so ferro cannot *serve* a GRCh37 request. The issue itself frames that as future work: *"Declining is the right immediate fix and #2089 is right to make it. Serving the answer is the better eventual one."* Serving it requires extending the single-span/`<diff>` parser to the `other_assembly` mappings (which carry 967 `lrg_ins` / 860 `other_ins` indel diffs across the prepared reference) and is unverifiable without the full 1294-record reference — a separate, larger change. This PR delivers the determinate defect (the misstated cause) and the actionable core of the second item: the decline no longer claims the mapping is absent — it names the build the placement is on. Recovering the GRCh37 mapping itself remains follow-up work.

## Test

`reanchor_ng_parent_placed_on_other_build_names_the_available_build` pins that a `GRCh37` request against a GRCh38-only placement names both builds, points at `--assembly GRCh38`, names `NC_000017.11`, and does **not** say "no chromosomal placement is known". `reanchor_ng_parent_with_no_placement_on_any_build_still_reports_unknown` pins that a genuine data gap keeps the unknown-placement message even when a build is requested. Verified failing-first: with the detection neutralized, the first test fails on the old message text.

## Gate

`cargo fmt --check`, `cargo clippy --all-features --all-targets -D warnings`, `cargo nextest run --features dev --no-fail-fast` (11347 passed), `scripts/run_oracle_suite.sh` (11177 passed), the scoped Python-binding Rust tests (10 passed), and `pytest tests/python/` (1137 passed) all green.

Representation-Change: none. The change only alters the *text* of a decline-branch error (decline stays a decline in every case — no previously-accepted input is now declined and no previously-declined input is now accepted), so no normalized corpus row moves.

Closes #2103

### DIFF
--- a/src/project/projector.rs
+++ b/src/project/projector.rs
@@ -2343,7 +2343,11 @@ impl<P: ReferenceProvider + Clone> VariantProjector<P> {
     /// with no known placement or an endpoint outside the placed span, and
     /// [`FerroError::InvalidCoordinates`] for an uncertain/compound endpoint
     /// (#655) — rather than emit a chromosome coordinate under the parent
-    /// accession (invalid HGVS).
+    /// accession (invalid HGVS). When a placement exists but only on a build
+    /// other than the requested one (an LRG/NG declining an explicit `GRCh37`
+    /// request whose only placement is `GRCh38`, #2089), the decline names the
+    /// build the placement is on and points at re-running under it, rather than
+    /// reporting the placement as unknown (#2103).
     fn reanchor_genome_output(
         &self,
         gv: crate::hgvs::variant::GenomeVariant,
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

Issue #2083, opened 2026-08-16
CLAUDE.md's bare `:N` shorthands still name the old spec numbering in four rows #2078 edited


nh13:
#2078 re-pointed the 32 qualified `general.md:N` anchors in `CLAUDE.md` and its description states: "The three bare `:N` shorthands that share a table row with a qualified anchor are fixed too, so no edited row is internally contradictory."

Seven bare `:N` shorthands survive, in four of the rows that PR edited, each contradicting the ledger record the row summarizes.

Ground truth from the pinned checkout (`565b9734`): `:33` splits, `:34` merges (codon exception), `:55` prioritisation, `:57` self-replacement prohibition, `:25` `g` = linear genomic.

| row | says | ledger says | stale |
|---|---|---|---|
| `contiguous-insertion-split-by-a-blocked-derivation` | "`general.md:33` is stated over 'two variants' … so `:34` does not reach them" | "`:33` does not reach a one-variant locus" | 1 |
| `inversion-vs-a-mixed-member-competitor` | "does not rest on `general.md:55` … that `:56` does not reach a multi-member allele" | "`:55` … does not reach one" | 1 |
| `coding-axis-merges-…` | "`:34` reaches these rows" / "`:34` says they are described individually" / "grades `:34` rule 2" / "`:34` has no antecedent" — beside a correct `general.md:34` codon exception in the same row | "`:33` DEVIATION … `:33` says to describe the members individually" | 4 |
| `projection-codon-exception-…` | "`general.md:22`/`:26` make the prefix a claim about the type of reference sequence" | cites `general.md:22` + `general.md:25` (`:25` = "`g` for a linear genomic reference sequence"); `:26` is now `m` for mitochondrial | 1 |

The `coding-axis-merges-…` row is the sharp one: within a single row, `:34` now means the separation rule four times and the codon exception once — the two clauses #2057 opens by observing "argue opposite ways". That is #2057's stated defect surviving in the largest row the PR edited.

The ledger already spells all four correctly, so this is a set difference rather than a judgement call.

## Related

#2078 declared `Closes #2057`, so #2057 auto-closed with its own described symptom live in the file. It should have been `Refs`. Reopening #2057.

## The principled fix

[...]

PR #2091, opened 2026-08-16
docs: re-point seven stale bare general.md anchors in CLAUDE.md's ruling tables

nh13:
Closes #2083

#2078 re-pointed the 32 qualified `general.md:N` anchors in `CLAUDE.md` after the spec submodule bump, and its description claimed the bare `` `:N` `` shorthands sharing those rows were fixed too. Seven were not: they still name the pre-bump numbering in four of the edited rows, and each contradicts the ledger record its row summarizes. The sharpest is the `coding-axis-merges-…` row, where a bare `:34` (separation rule under the old numbering) now sits four times beside the correct qualified `general.md:34` (the codon exception) — the two clauses argue opposite ways, which is exactly the defect the `claude_md_clause_anchors` guard's own doc comment calls out as its out-of-scope hole.

Each correction was cross-checked two ways: against the ledger record's own citations in `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json`, and against the pinned spec checkout `565b9734` (`:25` = `g` linear genomic, `:33` = "described individually" / splits, `:34` = codon exception / merges, `:55` = prioritisation).

## The seven corrections (old -> new)

| row | old bare anchor | new | why |
|---|---|---|---|
| `contiguous-insertion-split-by-a-blocked-derivation` | `` `:34` `` | `` `:33` `` | the split rule ("two variants") is `:33`; the sentence is about it not reaching a one-variant locus |
| `inversion-vs-a-mixed-member-competitor` | `` `:56` `` | `` `:55` `` | the prioritisation clause is `:55`; the row's qualified anchor already reads `general.md:55` |
| `coding-axis-merges-are-a-disclosed-general-34-deviation` | `` `:34` `` (x4) | `` `:33` `` | the individual-description / separation rule ferro deviates from is `:33`; the one correct qualified `general.md:34` codon-exception citation in the row is left untouched |
| `projection-codon-exception-is-decided-by-the-rendered-axis` | `` `:26` `` | `` `:25` `` | `g` = linear genomic is `:25` (`:26` is now `m` mitochondrial); the row's qualified anchor reads `general.md:22`/`:25` |

Ground-truth source: the ledger records (same-id) and the pinned spec checkout `565b9734`, as given in #2083.

## Scope

[...]

PR #2122, opened 2026-08-17
test: set-compare CLAUDE.md's bare `:N` shorthands against their ledger record

### LINKED ISSUE #2096: Guard: set-compare CLAUDE.md bare `:N` shorthands against their same-id ledger record
#2091 corrects the seven stale bare `:N` shorthands in `CLAUDE.md` that #2078 left behind, and closes #2083. It deliberately leaves the mechanical guard for a separate PR, and that follow-up currently has nothing holding it — hence this issue.

## The gap

Four guards already cover `CLAUDE.md`/ledger anchors: `claude_md_clause_anchors`, `claude_md_adjudication_tables`, `ledger_prose_clause_anchors`, `adjudication_prose_regrowth`. All 31 assertions pass on the pre-#2091 tree — i.e. **they were green across the entire window in which the seven anchors were wrong**. `claude_md_clause_anchors`'s own doc comment names this as its out-of-scope hole: it validates *qualified* `general.md:N` anchors and does not look at bare `` `:N` `` shorthands at all.

So the class is unguarded in both directions: a bare shorthand can be stale from the start, and a spec-submodule bump can silently invalidate every one of them at once.

## Proposed guard

Extend `claude_md_adjudication_tables` to set-compare, per row, the bare `` `:N` `` citations against the same-id record in `tests/fixtures/grammar/hgvs_spec_normalization_overrides.json`. A row whose bare anchors are not a subset of its ledger record's cited lines fails, naming the row id and the offending numbers.

The `coding-axis-merges-are-a-disclosed-general-34-deviation` row is the shape that makes this worth mechanising: a bare `:34` appeared four times beside a *correct* qualified `general.md:34`, and the two clauses argue opposite ways (separation rule vs. codon exception). Nothing about the text looks wrong — you have to resolve both numbers against the spec to see it.

## Why eyeballing does not work, and a trap to encode in the guard

Verifying these by hand from a worktree is actively misleading. `main/`'s working tree currently pins the submodule at `6f85311989e7` (21.0.4-82) while `origin/main`'s gitlink is `565b9734a4a4`. Reading `general.md` from the checked-out submodule therefore yields the **pre-bump** numbering, in which the old anchors are correct and #2091's corrections read as regressions:

| line | at `6f85311` (stale worktree) | at `565b9734` (origin/main) |
|---|---|---|
| `:25` | `c` coding DNA | `g` linear genomic |
| `:26` | `g` linear genomic | `m` mitochondrial |
| `:33` | *(blank)* | separation / describe individually |
| `:34` | separation / describe individually | codon exception |
| `:55` | *(blank)* | prioritisation |
| `:56` | prioritisation | dup-vs-ins prioritisation |

Every "old" value in #2091's table is exactly the stale checkout's numbering, which is what makes the corrections correct — but a reviewer spot-checking from `main/` would conclude the opposite with high confidence, because the numbers do line up, just against the wrong pin.

The guard must therefore resolve `general.md` from the **gitlink recorded in the commit under test**, not from whatever the submodule working tree happens to be checked out at. A guard that reads the working-tree copy would go green on a stale checkout and red on a fresh one, for the same source.

## Acceptance

- The guard reddens on the pre-#2091 tree (the seven known-bad anchors) and greens on the post-#2091 tree.
- It reads `general.md` at the commit's own gitlink; verify by checking out a stale submodule and confirming the result does not change.
- A new bare `:N` in a row with no matching ledger citation fails with the row id in the message.

nh13:
Closes #2096

## The gap

`CLAUDE.md`'s two adjudication tables cite spec clauses in two registers. The qualified form (`general.md:33`) is covered by `claude_md_clause_anchors`; the bare shorthand (`` `:33` ``) was covered by nothing. That guard's own doc comment says so under "What it does NOT catch" — it keys on the literal `general.md:` token — and the consequence was measured: all 31 assertions across the four existing `CLAUDE.md`/ledger guards stayed green for the entire window in which seven bare shorthands were wrong (#2083, corrected by #2091).

## What the guard does

For each row of the two tables, the row id names a ledger record, and that record's cited lines — its structured `clauses`, its `governing` clause, and the `file:line` tokens in its own `question` and `rationale` prose — are the comparand. Every bare shorthand in the row must name one of them; a fault reports the row id and the offending number alongside the lines the record actually cites.

Measured on this tree: **54 shorthands judged, 3 skipped, 0 faults.** On the pre-#2091 tree, **3 faults**.

### It reads the spec at the gitlink, not at the submodule working tree

This is the part that inverts the verdict if you get it wrong. A worktree in this layout routinely pins `assets/hgvs-nomenclature` older than the superproject's gitlink, and the #1793 bump shifted every `general.md` line from old `:23` on by one — so `:34` is the separation rule under `6f85311` and the codon exception under `565b9734`, and both readings are internally consistent. A guard reading the file off disk is green on a stale checkout and red on a fresh one for byte-identical source, and reports #2091's corrections as regressions.

Every spec byte here therefore comes through `SpecCheckout`, which resolves `git ls-tree HEAD assets/hgvs-nomenclature` and reads `git -C … show <sha>:<path>`. `the_spec_text_comes_from_the_gitlink_not_the_submodule_working_tree` pins that against a second, differently-derived read (tree lookup plus `cat-file blob`), so a future edit swapping in a filesystem read fails rather than silently making the file pin-dependent.

**Verified rather than asserted.** With the submodule checked out at `6f85311` while the gitlink stayed `565b9734`:

| suite | at the gitlink | with the submodule stale |
|---|---|---|
| this file (5 tests) | 5 pass | **5 pass — unchanged** |
| `claude_md_clause_anchors` (4 tests) | 4 pass | **2 pass, 2 FAIL** |

The two that flip are `every_qualified_citation_resolves` (`CLAUDE.md cites general.md:33, which is blank`) and `a_citation_names_the_clause_its_sentence_is_about` (9 citations "name a different line"). Both read the working tree. Nothing in this PR changes that file; it is noted so it is not rediscovered as a mystery, and it is worth its own issue.

## Two design decisions that came out of measurement

**The comparison is on line numbers, not on `(file, line)` pairs.** The obvious rule — a shorthand inherits the file of the nearest preceding qualified anchor — was implemented and run over the tables: **17 failures on a correct `CLAUDE.md`**. The prose does not obey it (`` `:130` `` after `complex.md:5` means `complex.md:130`; `` `:33` `` after `DNA/delins.md:81` means `general.md:33`). The file is therefore used for scoping only, never as an expectation.

The consequence is stated in the module docs rather than left to be found: the four stale `` `:34` `` anchors in the `coding-axis-merges-…` row are **not** caught, because that record cites `general.md:34` legitimately in the opposite role, so `34` is in the comparand. This PR catches 3 of the 7, one in each of the other three affected rows. A content-based discriminator for the remaining four — matching the citing sentence against word n-grams of each cited spec line at the gitlink — was prototyped and rejected on measurement: restricted to sentences carrying exactly one anchor it judges 3 of 54 anchors and produces a false failure on the `duplication-must-ranks-the-label-not-the-partition` row at every n-gram width tried.

**The comparand includes the record's prose.** Restricting it to the structured `clauses` array was measured and produces exactly one false failure: `absolute-prohibition-enforcement-stage` discusses `checklist.md:45` in its rationale while listing four other checklist lines as clauses, and `CLAUDE.md`'s row summarises the whole record.

## Non-vacuity

A guard that cannot tell "checked and fine" from "could not check" is the failure mode this repo names most often, so each degenerate parse is a failure rather than a pass:

- a table header that stops matching **panics** (verified live: renaming `| record | what is open |` reddens the run);
- a row naming no ledger record is a **fault**, not a skip;
- `JUDGED_BARE_ANCHORS_FLOOR` fails a run whose scanner found nothing;
- both skip classes are counted and printed in every message;
- a past-the-end shorthand is excused **only** in a row that cites a source file at all — otherwise it is a fault, because `` `:1963` `` after `spec_conformance_axis.rs:1982` is a line of Rust and a stale clause number must not be able to hide behind that rule.

Sabotage, run against the real `CLAUDE.md` and then restored byte-identically (md5 `49506e8e…` before and after):

| sabotage | result |
|---|---|
| re-point `` `:33` `` → `` `:34` `` in the `contiguous-insertion-split-by-a-blocked-derivation` row | RED, naming the row and `:34` |
| add an uncited `` `:171` `` to the `self-cancelling-across-ring-junctions` row | RED, naming the row and `:171` |
| rename the open-questions table header | RED at the parser, not a silent zero |

Both fault shapes and both skip rules also ship as committed tests over synthetic rows, so the guard's ability to fail is pinned rather than claimed.

## Scope

Test-only: one new module plus its `mod` line in `tests/it/main.rs`. `CLAUDE.md` is untouched — it is already correct under this rule — and `clause_ruling_index.rs` and the ledger are untouched, to stay clear of #2100/#2107. Gate: `cargo nextest run --features dev --no-fail-fast` 11359 passed, `cargo clippy --all-features --all-targets -- -D warnings`, `cargo fmt --all --check`, and `pytest tests/python/ -q` 1137 passed — all exit 0.

Representation-Change: none. Tests only; `scripts/check_representation_change.py` reports no file under a watched directory changed.

### DIFF
--- a/tests/it/claude_md_bare_clause_anchors.rs
+++ b/tests/it/claude_md_bare_clause_anchors.rs
@@ -1,0 +1,930 @@
+//! A bare `` `:N` `` shorthand in `CLAUDE.md`'s two adjudication tables must name
+//! a line the row's own ledger record cites.
+//!
+//! # The hole this closes
+//!
+//! Four guards already read `CLAUDE.md` against the ledger —
+//! [`claude_md_clause_anchors`], [`claude_md_adjudication_tables`],
+//! [`ledger_prose_clause_anchors`] and [`adjudication_prose_regrowth`]. All 31 of
+//! their assertions were **green for the whole window in which seven bare anchors
+//! were wrong** (#2083, corrected by #2091). That is not a bug in any of them:
+//! [`claude_md_clause_anchors`]'s own doc comment names the gap in as many words,
+//! under "What it does NOT catch" — it keys on the literal `general.md:` token, so
+//! a shorthand written `` `:34` `` is invisible to it, and nothing enforced that
+//! such a shorthand agrees with the row it sits in.
+//!
+//! The class is unguarded in both directions. A shorthand can be wrong the day it
+//! is written, and a spec-submodule bump can invalidate every one of them at once
+//! while every number on the page still resolves to a real, non-blank line.
+//!
+//! The shape worth mechanising is the `coding-axis-merges-…` row, where a stale
+//! `` `:34` `` sat four times beside a *correct* qualified `general.md:34` and the
+//! two clauses argue opposite ways — `general.md:33` is the separation rule
+//! ("described individually"), `general.md:34` the codon exception ("should be
+//! described as a delins"). Nothing about the prose looks wrong; both numbers have
+//! to be resolved against the spec before the contradiction is visible.
+//!
+//! # Read the spec at the GITLINK, never at the submodule working tree
+//!
+//! This is the trap the class is made of, and it inverts the verdict rather than
+//! blurring it. A worktree in this layout routinely pins the submodule at an older
+//! commit than the superproject's gitlink — `6f85311` against `565b9734` while
+//! #2091 was being reviewed — and the two numberings are each internally
+//! consistent:
+//!
+//! | line | at `6f85311` | at `565b9734` |
+//! |---|---|---|
+//! | `:25` | `c` coding DNA | `g` linear genomic |
+//! | `:33` | *(blank)* | separation / describe individually |
+//! | `:34` | separation / describe individually | codon exception |
+//! | `:55` | *(blank)* | prioritisation |
+//!
+//! So a guard reading `assets/hgvs-nomenclature/docs/…` off the filesystem is
+//! green on a stale checkout and red on a fresh one **for byte-identical source**,
+//! and it reports #2091's corrections as regressions with high confidence. Every
+//! spec byte this file reads therefore comes from [`SpecCheckout`], which resolves
+//! the gitlink with `git ls-tree HEAD assets/hgvs-nomenclature` and reads content
+//! with `git -C … show <sha>:<path>`. There is no `fs::read_to_string` on the
+//! submodule path anywhere below, and
+//! [`the_spec_text_comes_from_the_gitlink_not_the_submodule_working_tree`] is what
+//! keeps it that way.
+//!
+//! # What the comparison is, and what it deliberately is not
+//!
+//! For each table row: the row's id names a ledger record, and that record's cited
+//! lines — its structured `clauses`, its `governing` clause, and the `file:line`
+//! tokens in its own `question` and `rationale` prose — are the comparand. Every
+//! bare shorthand in the row must intersect that set.
+//!
+//! **The comparison is on line numbers, not on `(file, line)` pairs, and that is a
+//! measurement rather than a preference.** The obvious rule — a shorthand inherits
+//! the file of the nearest preceding qualified anchor — was implemented and run
+//! over the tables: it reports **17** failures on a `CLAUDE.md` that is correct.
+//! The prose simply does not obey it. In the `self-cancelling-across-ring-junctions`
+//! row a `` `:130` `` follows `complex.md:5` and means `complex.md:130`; in the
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

Issue #2080, opened 2026-08-16
ci: benchmark::hgvs_rs's 3 Rust tests are compiled by CI but never run — the #2046 gap, one feature over


nh13:
#2046's closing paragraph asked whether any *other* feature is in the same position as `python`. It is: `hgvs-rs`.

## The measurement

Test lists built from this tree (`cargo nextest list --lib`, one feature set each), differenced:

```
$ comm -23 <(cargo nextest list --features hgvs-rs --lib | sort) \
           <(cargo nextest list --features dev      --lib | sort)
ferro-hgvs benchmark::hgvs_rs::tests::test_default_config
ferro-hgvs benchmark::hgvs_rs::tests::test_parse_uta_url
ferro-hgvs benchmark::hgvs_rs::tests::test_parse_uta_url_remote
```

Three tests exist only under `--features hgvs-rs`, because `src/benchmark/mod.rs` declares their module behind the gate:

```rust
#[cfg(feature = "hgvs-rs")]
pub mod hgvs_rs;
```

**No CI command runs them.** The two places the feature appears are both build-only:

- `ci.yml`'s `clippy-all-features` runs `cargo clippy --all-features --all-targets`, which compiles the `#[cfg(test)]` module but does not execute it. Its own comment names `python` and `hgvs-rs` as the features it exists to keep compiling — that is exactly the "compiled, never run" state #2046 was about.
- `deploy-local.yml` runs `cargo build --release --features "web-service hgvs-rs" --bin ferro-web` — a binary build, no tests.

So a break in any of the three passes every check, which is the #1505 / #1372 / #2046 family one feature over.

They are cheap and hermetic — URL parsing and a default-config constructor. None needs a UTA database or a SeqRepo checkout.

## A fourth instance, at sub-test granularity

`src/service/tools/hgvs_rs.rs`'s `test_hgvs_rs_service_creation_without_feature` has two arms:

```rust
#[cfg(not(feature = "hgvs-rs"))]
{ /* asserts the config error */ }

#[cfg(feature = "hgvs-rs")]
{ let _result = HgvsRsService::new(&config); }
```

Only the `not(...)` arm is ever executed. A list diff cannot see this one — the test name exists under both feature sets, only its body differs — so it needs the source-level check, not the list check. Note the enabled arm **asserts nothing** even if it did run, so covering it means giving it an assertion, not merely running it.

## Not in the same position

For completeness, the other features outside `dev`:

[...]

PR #2084, opened 2026-08-16
ci: run benchmark::hgvs_rs's Rust tests (the #2046 gap, one feature over)

nh13:
Closes #2080

## The defect

`src/benchmark/hgvs_rs.rs` carries 3 pure-Rust `#[cfg(test)]` tests (`benchmark::hgvs_rs::tests::{test_parse_uta_url, test_parse_uta_url_remote, test_default_config}`) that CI **compiles but never runs** — the #1505 / #1372 / #2046 dead-coverage family, one feature over. A break in any of the three passes every check.

Measured on `origin/main` at `9b6dcf87`:

```
$ cargo nextest list --features hgvs-rs --lib -E 'test(benchmark::hgvs_rs::)'
ferro-hgvs benchmark::hgvs_rs::tests::test_default_config
ferro-hgvs benchmark::hgvs_rs::tests::test_parse_uta_url
ferro-hgvs benchmark::hgvs_rs::tests::test_parse_uta_url_remote
$ cargo nextest list --features dev     --lib -E 'test(benchmark::hgvs_rs::)'
   (empty — dev does not include hgvs-rs)
```

The three tests exist only under `--features hgvs-rs`, because `src/benchmark/mod.rs` gates the module behind it. No CI command runs them:

- the sharded `test` jobs run `--features dev`, and `dev` excludes `hgvs-rs`, so they never compile the module;
- `clippy-all-features` runs `cargo clippy --all-features --all-targets` — compiles the `#[cfg(test)]` module but does not execute it (its own comment names `hgvs-rs` as a feature it exists to keep compiling);
- `deploy-local.yml` runs `cargo build --release --features "web-service hgvs-rs" --bin ferro-web` — a binary build, no tests.

## The fix

Unlike #2046, there is no linkage problem here — `hgvs-rs` needs no feature split, just a job that runs the tests. This PR adds a dedicated `hgvs-rs Rust tests` job that runs the scoped selection and wires it into the `Test` rollup's `needs` so it gates the merge (the ruleset's required-context set is fixed and does not name this job, so gating goes through the rollup rather than a settings change).

```
cargo nextest run --features hgvs-rs --lib --no-tests=fail -E 'test(benchmark::hgvs_rs::)'
```

### Scoped, and under nextest — same reasoning as #2072's `python::tests` step

[...]

PR #2126, opened 2026-08-17
ci: cover #2080's fourth dead-coverage site (service hgvs_rs enabled arm)

### LINKED ISSUE #2101: ci: the benchmark::hgvs_rs job does not reach #2080's fourth dead-coverage site
Split off from the #2084 review. #2084 closes #2080, but #2080 documents four dead-coverage sites and #2084's new job reaches three. This tracks the fourth, which would otherwise have no home.

`src/service/tools/hgvs_rs.rs::test_hgvs_rs_service_creation_without_feature` (present on `bcec0f4c` at `:185`) has two `cfg` arms:

- `#[cfg(feature = "hgvs-rs")]` — **asserts nothing** (`let _result = ...`). It has never executed in CI.
- `#[cfg(not(feature = "hgvs-rs"))]` — executes, but its `assert!(msg.contains("hgvs-rs"))` sits **inside** `if let Err(ServiceError::ConfigError(msg))`, so a differently-typed error passes the test silently.

#2084's new job builds `--features hgvs-rs` **without** `web-service`, so `src/service` is not compiled and the enabled arm is still unreached.

Covering this means giving the arm a meaningful assertion that does not depend on a live UTA database, not merely arranging for it to run. The `not(feature)` arm's conditional assert should also become unconditional on the error type.

Search performed: `gh issue list --state all --search "hgvs_rs_service_creation"` returns `[]`.

nh13:
Closes #2101

## What

#2080 documented four dead-coverage sites; #2084 reached three. This covers the fourth: the `#[cfg(feature = "hgvs-rs")]` arm of `service::tools::hgvs_rs::tests::test_hgvs_rs_service_creation` (formerly `..._without_feature`).

## Why it was uncovered

`src/service` compiles only under `web-service` (`#[cfg(feature = "web-service")] pub mod service;`), while the enabled arm is gated on `hgvs-rs`. No CI job builds both features at once:

- the sharded `Test` jobs use `--features dev`, which has `web-service` but not `hgvs-rs`;
- #2084's `hgvs-rs Rust tests` job uses `--features hgvs-rs`, without `web-service`.

So the enabled arm ran nowhere and a break in it passed every check — the same dead-coverage family as #2084's benchmark tests, one feature combination over.

## Changes

- **CI**: a second step in the `hgvs-rs Rust tests` job builds `--features hgvs-rs,web-service` and runs `-E 'test(service::tools::hgvs_rs::)'` under nextest with `--no-tests=fail`. It is a separate build from #2084's benchmark step, so that proven coverage stays on exactly the feature set it was verified with.
- **Test**: both feature arms now assert unconditionally that `HgvsRsService::new` fails with a `ServiceError::ConfigError` and never panics. The previous enabled arm asserted nothing (`let _result = ...`), and the disabled arm's `assert!(msg.contains("hgvs-rs"))` sat inside an `if let`, so a differently-typed error passed silently. The config uses an empty `seqrepo_path`, which makes provider construction fail during path parsing *before any UTA connection is attempted* — so the failure is deterministic and independent of whether a live UTA database is reachable.

## Verification

- `--features hgvs-rs,web-service` selection: both tests PASS (enabled arm now executes).
- `--features dev` selection: both tests PASS (disabled arm).
- **Mutation**: changing the enabled arm's error variant `ConfigError` → `InternalError` turns the test RED (`expected ServiceError::ConfigError, got InternalError(... could not get parent from ...)`), confirming the test exercises the enabled arm and asserts on its real behavior, and that the failure arises pre-database.
- Full local gate green: `cargo fmt --check`, `cargo clippy --all-features --all-targets -- -D warnings`, `cargo nextest run --features dev --no-fail-fast` (11354 passed), `actionlint`.

Representation-Change: none. CI/test-coverage only; no library normalization behavior changed and no watched prefix (`src/normalize|hgvs|spdi|project|reference|error_handling`) is touched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2131,6 +2131,27 @@ jobs:
         run: |
           cargo nextest run --features hgvs-rs --lib --no-tests=fail \
             -E 'test(benchmark::hgvs_rs::)'
+      # src/service/tools/hgvs_rs.rs was #2080's FOURTH dead-coverage site (#2101),
+      # which the step above cannot reach. `src/service` compiles only under
+      # `web-service` (`#[cfg(feature = "web-service")] pub mod service;`), while the
+      # enabled arm of `service::tools::hgvs_rs::tests::test_hgvs_rs_service_creation`
+      # is `#[cfg(feature = "hgvs-rs")]`. No other CI job builds BOTH features at once:
+      # `dev` (the sharded `test` jobs) has `web-service` but not `hgvs-rs`, and the
+      # step above has `hgvs-rs` but not `web-service` — so that arm ran nowhere and a
+      # break in it passed every check, the same family as the step above. It is
+      # reached here by asking for the two features together.
+      #
+      # Hermetic for the same reason: the test drives `HgvsRsService::new` with an
+      # empty seqrepo path, which fails during provider path parsing before any UTA
+      # connection is attempted, so it needs no database, SeqRepo, or spec submodule.
+      # A SEPARATE build (not folded into the step above) because `web-service` pulls
+      # axum/tokio/tower, a tree the benchmark step does not need — keeping them apart
+      # leaves #2080's proven coverage on exactly the feature set it was verified with.
+      # SCOPED and `--no-tests=fail` for the same non-vacuity reasons as above.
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #2129, opened 2026-08-17
fix(extractor): decline instead of panicking on end insertions/duplications

### LINKED ISSUE #2128: ferro describe panics (exit 101) on insertions and duplications
Split out of the #1701 engine-disagreement measurement, which priced `ferro describe` (`DescriptionExtractor`) against `from_sequences` over a 1,030-pair synthetic corpus. Filing the crash separately because it is a determinate defect independent of whichever product option #1701 lands on: a user-reachable panic should be a clean, typed error, not `exit 101`.

## The bug

`ferro describe` **panics** (process exit 101) on a large class of inputs — every insertion and duplication in the #1701 corpus (436 of 1,030 pairs) — rather than returning a description or a clean decline. `from_sequences` handles the same inputs with 0 crashes and 0 declines.

The panic originates in the classifier, reached straight from the `describe` CLI path:

- `src/extractor/classify.rs` — `is_duplication` (around `:143` at the time of the #1701 measurement; re-locate on current `main` before fixing, line numbers drift).

## Expected

A user-facing CLI entry point must never panic on well-formed input. `ferro describe` should either produce a description or return a clean, typed error (an `Err`/decline surfaced as a non-zero-but-graceful exit with a message), consistent with how the rest of the CLI handles inputs it cannot serve. This is a bug→clean-decline hardening, not a change to any normalized representation.

## Repro

Reproduced concretely through the CLI during the #1701 measurement (insertion and duplication inputs → exit 101). A minimal standalone repro should be added as a failing-first test before the fix.

## Scope

- Turn the panic into a graceful typed error/decline at the classifier boundary.
- Add a regression test (failing-first) that pins a representative insertion and a duplication input to a clean error rather than a panic.
- Not in scope: the #1701 product decision about which engine is authoritative, or any normalization/representation change.

Refs #1701.

nh13:
Closes #2128

## Problem

`ferro describe` (the `DescriptionExtractor`) panicked with process exit 101 on a large class of well-formed inputs — trailing insertions and duplications. The duplication classifier read the bases preceding an insertion point with `reference[pos - ins_len..pos]`, where `pos` is the alignment's 1-based `ref_start`. A trailing insertion reports `pos == reference.len() + 1`, so the slice ran off the end of the reference and panicked at `src/extractor/classify.rs` (`end byte index N is out of bounds for string of length M`), aborting the process.

A user-facing CLI entry point must never panic on well-formed input.

## Fix

`is_duplication` and `classify_edit` now return a typed `Result<_, FerroError>`. When the insertion point lies past the end of the reference — an unresolvable coordinate the classifier cannot evaluate — they return `FerroError::InvalidCoordinates`. `DescriptionExtractor::extract` already returns `Result`, so the error propagates through `ferro describe` as a clean non-zero exit with a message instead of an abort. This mirrors the guard the normalize twin `insertion_is_duplication` already carries.

In-bounds behaviour is unchanged: the boundary case `pos == reference.len()` still resolves, and every currently-served description is byte-identical.

Before / after (previously exit 101):

```
$ ferro describe --reference ATG --observed ATGC
Error: InvalidCoordinates { msg: "insertion point 4 is past the end of the 3-base reference" }   # exit 1
```

## Tests (failing-first)

- `extractor::tests::an_insertion_past_the_reference_end_declines_instead_of_panicking`
- `extractor::tests::a_duplication_past_the_reference_end_declines_instead_of_panicking`
- plus a direct unit guard `classify::tests::is_duplication_past_the_reference_end_is_a_typed_error`

All three panic on the base commit and assert the specific `InvalidCoordinates` decline after the fix.

Representation-Change: none. Hardening only — turns a panic into a typed decline; touches only `src/extractor/` (not a watched prefix) and changes no produced description or normalized output.

### DIFF
--- a/src/extractor/classify.rs
+++ b/src/extractor/classify.rs
@@ -3,6 +3,7 @@
 //! Classifies raw edits into HGVS-compatible types.
 
 use super::align::RawEdit;
+use crate::error::FerroError;
 
 /// Classification of an edit for HGVS notation.
 #[derive(Debug, Clone, PartialEq)]
@@ -63,11 +64,11 @@ pub fn classify_edit(
     reference: &str,
     detect_dup: bool,
     detect_inv: bool,
-) -> EditClassification {
+) -> Result<EditClassification, FerroError> {
     let ref_len = edit.ref_seq.len();
     let obs_len = edit.obs_seq.len();
 
-    match (ref_len, obs_len) {
+    let classification = match (ref_len, obs_len) {
         // Single substitution
         (1, 1) => EditClassification::Substitution {
             ref_base: edit.ref_seq.chars().next().unwrap(),
@@ -81,7 +82,7 @@ pub fn classify_edit(
 
         // Insertion - check for duplication
         (0, a) if a > 0 => {
-            if detect_dup && is_duplication(reference, edit.ref_start, &edit.obs_seq) {
+            if detect_dup && is_duplication(reference, edit.ref_start, &edit.obs_seq)? {
                 EditClassification::Duplication {
                     sequence: edit.obs_seq.clone(),
                 }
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

Issue #2114, opened 2026-08-17
Slim the README: relocate bulk content into the mdBook docs site


nh13:
The `README.md` is 976 lines and doing far too much: alongside the landing-page essentials (what it is, install, quick start) it carries the canonical normalization ruleset, a ~294-line "deriving a description from sequences" explainer, the HGVS syntax reference, the CLI command list, error-handling docs, and the `FERRO_PARTITION` comparison guide, all inline. A README should be a landing page; the long-form material now has a home in the mdBook docs site.

## Constraint: move, don't copy

The project's recurring failure mode is one rule written in two places that then drift apart. Every relocation here is a **move** — delete from the README, single-source in the mdBook, and repoint every citation. No summary is left behind.

## Plan

Two PRs, by risk class:

**PR 1 — bulk prose relocation (low risk).** Move the five pure-prose sections into mdBook pages and repoint their links:
- Deriving a description from sequences → a `guide/` explainer
- Supported HGVS Syntax → a `reference/` page
- CLI Commands → fold into `cli/overview.md`
- Error Handling → a `guide/` page
- Comparing normalization rules (`FERRO_PARTITION`) → an advanced `guide/` page

Takes the README to ~500 lines. No ledger or ruling change.

**PR 2 — the canonical ruleset (linchpin).** Move the Normalization rules section to a new mdBook reference page. This is entangled with a decided operator ruling (`adjudication-precedence-order`, whose thesis is that the ruleset lives in `README.md`), a drift-guard test (`tests/it/readme_normalization_rules.rs`), the ledger's guard citation, and five `#normalization-rules` anchor citations. The ruling amendment is an operator act and is drafted for operator sign-off in that PR. Takes the README to ~150–250 lines.

Out of scope for now: the generator-backed "Normalization Capabilities" and "Performance Comparison" tables (written into the README by `generate_tool_support_tables` / `generate_perf_tables` between HTML markers) — relocating those requires retargeting two `--check`-gated generators and can be a later cleanup.

PR #2131, opened 2026-08-17
docs: move the comparison tables and 'Why ferro-hgvs?' into the mdBook

nh13:
Follow-up to the README-slim campaign (#2114): the last generated long-form content leaves the README.

This relocates the generated tool-support capability matrix and the cross-tool performance tables — plus the surrounding "Why ferro-hgvs?" narrative and the reference-data overview — out of `README.md` and into a new `docs/src/reference/comparison.md` page, registered in `SUMMARY.md` under Reference. The README keeps a short "Why ferro-hgvs?" teaser linking to the page, and the Documentation index gains an entry for it.

The tables are generated, so this is a code change, not a copy: `generate_tool_support_tables` and `generate_perf_tables` are retargeted from `README.md` to the new page (the tool-support generator keeps its existing `docs/BENCHMARK_GUIDE.md` target). They stay single-sourced from `docs/tool_support_matrix.json` and `data/benchmark/perf_results.json`, and CI's `--check` gate now verifies the new page. The `## Benchmark:` command-reference section is intentionally left in the README — it is not a generated table.

Verified locally: `cargo fmt --check`, `cargo clippy --features dev --all-targets -D warnings`, both generators' `--check` green against the new page, `mdbook build` + linkcheck2 clean, and the generator/completeness/readme test selection (20 tests) pass.

Representation-Change: none. Docs and generator output paths only; no watched `src/` file is touched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,89 +179,13 @@ Full guides live in **[the documentation site](https://ferro-hgvs.readthedocs.io
 - [Comparing normalization rules (`FERRO_PARTITION`)](docs/src/guide/comparing-rules.md)
 - [Benchmarking](docs/src/guide/benchmarking.md)
 - [CLI reference](docs/src/cli/overview.md) · [Supported HGVS syntax](docs/src/reference/hgvs-syntax.md)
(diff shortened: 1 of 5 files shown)


nh13:
@coderabbitai pause

PR #2132, opened 2026-08-17
fix(convert): refuse an inverted-CDS record in convert::mapper's CoordinateMapper (both directions)

### LINKED ISSUE #2123: The other CoordinateMapper (convert/mapper.rs) has no inverted-CDS refusal
Found while reviewing #2109, which refuses an inverted-CDS record "on every path". That claim holds for the type it targets — but there are **two** types named `CoordinateMapper` in this codebase, and only one of them got the guard.

The second, `src/convert/mapper.rs` (over `Transcript`), has no inverted-CDS refusal. Measured on an inverted record, it answers:

```
c.*1 -> n.7
c.1  -> n.20
```

i.e. it maps happily across an inversion the other mapper now refuses, and the overlap in the answers makes the inversion visible in the output.

This is **pre-existing and outside #1894's scope**, so it was correctly not a blocker for #2109 — but the PR title reads wider than what landed, and a reader who greps for `CoordinateMapper` will find a guarded one and an unguarded one with no indication which they are looking at.

Two things to decide:

1. Whether the second mapper should carry the same refusal (probably — an inverted CDS is no more meaningful there), or whether its inputs are structurally guaranteed not to be inverted, in which case say so where the type is defined.
2. Whether two distinct public types should share the name `CoordinateMapper` at all. That is the reason this was easy to miss.

nh13:
There are two types named `CoordinateMapper` in this codebase. #2109 added an inverted-CDS refusal to the one in `src/data/mapping.rs`, but the second — in `src/convert/mapper.rs`, over `Transcript` — had none. On a record whose CDS bounds are inverted (`cds_end < cds_start`) it silently fabricated a coordinate straddling the inversion instead of refusing.

Reproduced from #2123: with `cds_start 20`, `cds_end 6`, `cds_to_tx` answered `c.1 -> n.20` and `c.*1 -> n.7`.

This mirrors #2109's `reject_inverted_cds` on the second type: a new `reject_inverted_cds` guard is called at the top of the two CDS-axis converters — `cds_to_tx` (c./n. -> tx) and `tx_to_cds` (tx -> c./n.) — so both directions, and the genomic delegators that funnel through them, decline a malformed record with a typed `FerroError::InvalidCoordinates` rather than computing on incoherent bounds. This turns a silent wrong answer into a clean, consistent refusal.

Tests (both mutation-verified against the wrong concrete answer on the pre-fix code):
- `cds_to_tx_on_an_inverted_cds_record_is_refused` — c. -> n., both a plain `c.1` and a `c.*1` (utr3) input
- `tx_to_cds_on_an_inverted_cds_record_is_refused` — n. -> c.

Representation-Change: none. Guards a malformed inverted-CDS record (`cds_end < cds_start`), which real cdot never ships, in `src/convert/` — a non-watched prefix; no normalized output moves.

Closes #2123

### DIFF
--- a/src/convert/mapper.rs
+++ b/src/convert/mapper.rs
@@ -84,6 +84,43 @@ impl<'a> CoordinateMapper<'a> {
         Self { transcript }
     }
 
+    /// Refuse a transcript whose CDS bounds are inverted (`cds_end < cds_start`)
+    /// before any coordinate arithmetic runs, so **both** CDS-axis conversion
+    /// directions — `c.`/`n.` ↔ tx here, and the genomic delegators that funnel
+    /// through them — decline such a record rather than fabricating a
+    /// coordinate.
+    ///
+    /// This is the `convert::mapper` twin of the guard #2109 added to the
+    /// *other* `CoordinateMapper` (`src/data/mapping.rs`, `reject_inverted_cds`,
+    /// #1894). On an inverted record the 5'UTR and 3'UTR halves overlap, so one
+    /// transcript position carries two `c.` names and any answer picks one
+    /// arbitrarily: measured on `NM_INVERTED.1` (cds_start 20, cds_end 6),
+    /// `c.1` mapped to `n.20` and `c.*1` to `n.7` — an answer straddling the
+    /// inversion, with no diagnostic (#2123).
+    ///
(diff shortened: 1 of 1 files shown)


nh13:
@coderabbitai pause

PR #2133, opened 2026-08-17
ci(soak): lower [profile.soak] opt-level from 3 to 2

### LINKED ISSUE #1963: ci: the soak archive build is now the critical path, and dependency/checkout caching is not the lever
After #1960 the workflow's critical path is entirely on the soak side. Measured on run [31849010009](https://github.com/fulcrumgenomics/ferro-hgvs/actions/runs/31849010009):

```
     1 ->    193  ( 192s)  Build soak archive (optimized)   <- the gate
   197 ->    307  ( 110s)  Exhaustive sweeps
   199 ->    289  (  90s)  Slow censuses (optimized)
   311 ->    314  (   3s)  Test (aggregator)
```

192 + 110 ~= 305s against a 314s wall, so the workflow is within ~10s of what its job graph allows and nothing in the `test` / `test-oracle` fan-out is on the path any more.

## Where the 192s goes

Step timings, not log lines — the API is authoritative and reading the log got this wrong once already:

| step | |
|---|---:|
| `cargo nextest archive --cargo-profile soak --test it` | **161s** |
| `Swatinem/rust-cache` restore | 15s |
| everything else (checkout, toolchain, cache save, artifact upload) | ~16s |

**There is nothing to hoist.** Setup is ~30s and already parallel with every other job; the compile *is* the job.

## What is already solved, so nobody re-opens it

- **Dependencies are cached and hitting.** Every compiling job compiles exactly **1** crate (`ferro-hgvs`), not ~400 — `Swatinem/rust-cache`, 11 entries, 3.5 GiB, restore 11-17s.
- **Checkout is 2-5s.** Nothing there.
- **Setup/teardown consolidation is not a wall-clock lever.** Per job: checkout 2-5s, `download-artifact` 2-4s each, `install-action` 2s, `rust-cache` restore 11-17s. All parallel, so it costs the critical path ~15s once. It is ~8 minutes of aggregate runner time across 34 jobs, which is a billing argument.
- **`sccache` is closed.** `CLAUDE.md` records that it classifies every `ferro-hgvs` unit `Non-cacheable calls, reason crate-type` — zero hits, zero misses. It earns its keep on dependencies only, which `rust-cache` already holds.

## The cache budget is NOT the problem — correcting an earlier claim

The repository sits at 10.19 GiB against a 10 GiB cap, and 6.64 GiB of that is SHA-keyed families that miss on every new commit (`nextest-archive-<sha>` 5.47 GiB / 46 entries, `nextest-soak-archive-<sha>` 1.11 GiB / 40). That reads like #1218 recurring, and it is not:

- Every `v0-rust-*` dependency cache was **last accessed 0.2h ago** — every run restores them, so they are permanently the most-recently-used and LRU cannot reach them.
- The SHA-keyed entries are written once and never read again, so they are always the LRU victims. Eviction is falling on exactly the entries that are worthless.
- The one large rarely-read entry #1218 was actually about, the nightly's 3.25 GiB `ferro-manifest-v3`, was **already removed deliberately** with measurements (`nightly-mutalyzer.yml`, "3.25 GiB with a 0% hit rate").

`prune-ci-caches.yml` works — 2026-08-14's run deleted 119 entries / 5.83 GiB and left 33 entries / 4.29 GiB — it is just daily against a much faster write rate, so the repository is over cap most of the day. That is hygiene, not a CI-speed defect, and it should not be filed as one.

## Candidate levers, and what each is worth

1. **`opt-level = 2` for `[profile.soak]`.** The only lever that helps every PR unconditionally. `test-build` compiles *every* test binary with `--features dev` at the `test` profile in 82s of cargo; `test-build-soak` compiles *one* binary at `opt-level = 3` in 161s, so the optimizer is the dominant term. The trade is `sweeps` / `censuses` runtime, which CI can measure directly.
2. **`cache-workspace-crates: true`.** `rust-cache` deliberately excludes workspace artifacts, and the two in question are small — `libferro_hgvs.rlib` 27.9 MB + the `it` test binary 34.3 MB, ~62 MB uncompressed. **But `it` depends on `ferro_hgvs`, so any `src/` change rebuilds both and the cache buys zero.** It pays on PRs that touch neither, and partially on `tests/it`-only PRs.
3. **Incremental compilation.** `target/soak/incremental` measures **1.25 GB** — 12% of the whole cache budget for one profile, with the same weakness as (2): it pays in proportion to how small the change is. CI paths are stable so it would work; the budget is the objection.
4. **Shard `sweeps`.** 99s of tests but a 76.1s slowest single test, so it floors out near 85s. Worth ~15-25s.
5. **A larger runner for that one job.** Roughly halves 161s; an org billing decision rather than a code one.

## Scope

Take (1) first, measured on CI. (2) and (3) are worth doing only if the split between the lib and the `it` binary makes the partial-hit case large, which is the open measurement.

nh13:
Closes #1963

## What

Lower `[profile.soak]`'s `opt-level` from 3 to 2 in `Cargo.toml`.

After #1960, the `test-build-soak` archive build is the workflow's critical path — measured at ~161s of `cargo nextest archive --cargo-profile soak` on CI, of which the optimizer is the dominant term (`test-build` compiles every test binary at the `test` profile — opt-level 0 — in ~82s). This is lever (1) from the issue: the only change that speeds up the critical path unconditionally, on every PR.

`opt-level = 2` cuts optimizer time on `libferro_hgvs` (the dominant compile term; the ~8k-line soak driver is a wash between opt-levels 2 and 3, per the measured #1964 table). `debug-assertions` and `overflow-checks` stay ON, so nothing the soak checks is weakened — the `#[cfg(debug_assertions)]` oracles and every `debug_assert!` behave exactly as before.

The trade is `sweeps` / `censuses` run-time, which those jobs measure directly on CI. Per the issue, this is the lever to take first and measure there; the local #1964 driver-opt-level table does not vary the library, so the library trade is a CI measurement rather than a pre-committed local number.

## Also

Comments in `.github/workflows/ci.yml` and `.github/workflows/nightly-soak.yml` that stated the soak profile's current `opt-level` are updated from 3 to 2 to match, and the #1964 driver-override table in `Cargo.toml` is annotated to note its rows were measured with the library held at opt-level 3 (which predates this change). `actionlint` passes on both workflows.

## Representation change

Representation-Change: none. Build-config only (`Cargo.toml` profile + workflow comments); touches no watched prefix (`src/normalize|hgvs|spdi|project|reference|error_handling`) and cannot move any normalized output.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1200,7 +1200,7 @@ jobs:
           path: nextest.tar.zst
           retention-days: 1
 
-  # A SECOND archive, built at `opt-level = 3`, for the soak shards only.
+  # A SECOND archive, built at `opt-level = 2` (#1963), for the soak shards only.
   #
   # The soak is compute-bound in a way the other jobs are not: eight shards each
   # run 125,000 proptest cases, so a constant factor on the code under test is
@@ -1214,8 +1214,8 @@ jobs:
   # tests — at `opt-level = 3` so that about 45 of them can run. Measured on an
   # M2 Max, `user` CPU with the workspace artifacts scrubbed the way `rust-cache`
   # scrubs them here, deps warm: 394.8s for the `it` archive, of which the
-  # library is a minority. The rest was the driver, and the runtime
-  # `opt-level = 3` buys comes from optimizing the LIBRARY.
+  # library is a minority. The rest was the driver, and the runtime this
+  # profile's optimization buys comes from optimizing the LIBRARY.
   #
   # Cargo profile overrides are keyed on the PACKAGE, never on the target, so
   # `[profile.soak.package.ferro-hgvs]` reaches the lib, the bins and the `it`
@@ -1279,8 +1279,8 @@ jobs:
       # Skip the rebuild entirely when this exact commit already produced an
       # archive. `rust-cache` above caches DEPENDENCIES only — it deliberately
       # deletes workspace artifacts — so `ferro_hgvs` and the `it` test crate are
-      # recompiled at `opt-level = 3` on every run even when not one byte of Rust
-      # changed. Measured on run 31275848682: 142s of this job's 170s is that
+      # recompiled at `opt-level = 2` (#1963) on every run even when not one byte
+      # of Rust changed. Measured on run 31275848682: 142s of this job's 170s is that
       # rebuild, and this job plus `sweeps` IS the critical path.
       #
       # Keyed on the commit SHA, deliberately, rather than on `hashFiles` over the
@@ -1316,7 +1316,7 @@ jobs:
         if: steps.soak-archive.outputs.cache-hit != 'true'
         # No `--features dev`. That feature pulls in `web-service` (axum, tokio,
         # tower, tracing) and `benchmark` (reqwest, rusqlite, tokio-postgres,
-        # futures) — an entire dependency tree compiled at `opt-level = 3` to
+        # futures) — an entire dependency tree compiled at `opt-level = 2` to
         # produce one test binary that never touches any of it. None of the ten
         # modules this package compiles is gated on the feature and none of the
         # three consuming jobs selects a gated one, so the selections are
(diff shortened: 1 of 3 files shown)


nh13:
@coderabbitai pause

PR #2137, opened 2026-08-17
test(claude-md): read clause-anchor spec at the gitlink, not the submodule working tree

### LINKED ISSUE #2124: claude_md_clause_anchors reads the spec from the submodule working tree, so its verdict depends on checkout state
Found while implementing #2096 (PR #2122). This is the failure #2096's own text warns about, present in a **committed guard** rather than hypothetical.

`claude_md_clause_anchors` reads the spec with `fs::read_to_string("assets/hgvs-nomenclature/…")` — i.e. from the submodule **working tree**, not from the gitlink recorded in the commit under test. On any worktree whose submodule lags the gitlink, two of its four tests go red on **correct** source:

- `every_qualified_citation_resolves` — *"CLAUDE.md cites general.md:33, which is blank"*
- `a_citation_names_the_clause_its_sentence_is_about` — 9 citations "name a different line"

Measured directly: with the worktree's submodule checked out at `6f85311` while the gitlink stayed `565b9734`, `claude_md_clause_anchors` went **4 pass → 2 pass / 2 FAIL** with no source change. The new `claude_md_bare_clause_anchors` suite stayed **5 pass → 5 pass** across the same swap, because it resolves through the gitlink.

This is not a rare setup. `main/`'s own worktree is currently at `6f85311` against an `origin/main` gitlink of `565b9734`, so anyone running the full suite there today sees two red tests that are green in CI — which reads as a real regression and invites a hunt.

The inverse is the more dangerous direction and is equally possible: a **stale-forward** submodule can make a genuinely wrong anchor resolve, so the guard passes on source it should reject.

**Fix:** resolve the spec through the commit's gitlink. `SpecCheckout` in `tests/it/claude_md_bare_clause_anchors.rs` (added by #2122) is a drop-in resolver — it reads via `git ls-tree HEAD assets/hgvs-nomenclature` plus `git -C … show <sha>:<path>`, and prints a diagnostic when the working tree and gitlink differ so the non-vacuous branch is visible in the log.

**Acceptance:** check the submodule out at a different revision, re-run, and confirm the verdict does not move. A guard whose result depends on a checkout state is not a guard.

nh13:
Closes #2124

## The defect

`claude_md_clause_anchors`'s `spec_lines()` read the vendored spec with `fs::read_to_string("assets/hgvs-nomenclature/…")` — from the submodule's mutable **working tree**, not the gitlink the commit under test records. Its verdict therefore depended on the checkout state rather than the pinned spec revision, which is precisely the failure #2096's text warned about, now present in a committed guard.

Reproduced directly on this branch's base: with the submodule working tree checked out at `6f85311` while the gitlink stayed `565b9734`, the module went from 5 passing to 3 passing / 2 failing (`every_qualified_citation_resolves` and `a_citation_names_the_clause_its_sentence_is_about`) on **byte-identical `CLAUDE.md` source**. The inverse — a stale-forward checkout — could make a genuinely wrong anchor resolve green.

## The fix

Resolve the spec through the commit's gitlink, reusing `SpecCheckout` from the sibling `claude_md_bare_clause_anchors` guard (added by #2122, which the issue names as the drop-in resolver). `SpecCheckout`, the `git` helper, `get`, `lines_of` and the `gitlink` field are widened to `pub(crate)`; the gitlink-clearing `GIT_DIR`/`GIT_WORK_TREE` subtlety lives in one place rather than being duplicated. `spec_lines()` now calls `SpecCheckout::get().lines_of(SPEC_FILE)`, so the verdict is a function of the pinned submodule revision.

`the_spec_text_is_read_at_the_gitlink_not_the_submodule_working_tree` is the new guard: it compares `spec_lines()` against an independently derived gitlink blob (`ls-tree` + `cat-file`, a different path from `lines_of`'s `git show`), so a future revert to a filesystem read is caught, and it reports the pin when the working tree and gitlink differ.

## Demonstration

- At the gitlink (`565b9734`): all 6 module tests pass.
- Working tree swapped to `6f85311` (gitlink unchanged): **all 6 still pass** — verdict unmoved — and the new test's non-vacuous branch fires, printing `working-tree copy DIFFERS from the gitlink's`, proving the pin-independence assertion executed at a differing checkout. That is the issue's stated acceptance criterion.

Full local gate green: `cargo fmt --check`, `cargo clippy --all-features --all-targets -- -D warnings`, `cargo nextest run --features dev --no-fail-fast` (11368 passed, 0 failed).

Representation-Change: none. Touches only `tests/it/`; no `src/normalize|hgvs|spdi|project|reference|error_handling` file is changed.

### DIFF
--- a/tests/it/claude_md_bare_clause_anchors.rs
+++ b/tests/it/claude_md_bare_clause_anchors.rs
@@ -160,7 +160,7 @@ fn crate_root() -> PathBuf {
 /// `GIT_DIR`/`GIT_WORK_TREE` pointing at the **outer** repository when it runs a
 /// hook, and those override `-C`'s repository discovery — so under the pre-push
 /// hook a submodule query would silently answer about the superproject.
-fn git(cwd: &Path, args: &[&str]) -> Result<String, String> {
+pub(crate) fn git(cwd: &Path, args: &[&str]) -> Result<String, String> {
     let output = Command::new("git")
         .arg("-C")
         .arg(cwd)
@@ -183,9 +183,14 @@ fn git(cwd: &Path, args: &[&str]) -> Result<String, String> {
 
 /// The spec checkout as the commit under test records it — never as the submodule
 /// working tree happens to be checked out.
-struct SpecCheckout {
+///
+/// Shared with [`claude_md_clause_anchors`], which resolves its one spec file the
+/// same way (#2124) — the gitlink resolver is deliberately in one place so both
+/// guards read the spec at the commit under test rather than at whatever the
+/// submodule working tree happens to be checked out to.
+pub(crate) struct SpecCheckout {
     /// The commit the superproject's tree records for [`SPEC_DIR`].
-    gitlink: String,
+    pub(crate) gitlink: String,
     /// Every `.md` path under [`SPEC_DOCS_ROOT`] at that commit.
     paths: Vec<String>,
     /// `path -> lines`, filled on demand. A handful of files are ever needed.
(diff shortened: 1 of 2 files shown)


nh13:
@coderabbitai pause

PR #2141, opened 2026-08-17
fix(spdi): refuse an N-unit repeat instead of emitting a run of Ns

### LINKED ISSUE #1975: spdi: an N-unit repeat emits a run of Ns, while the equally-undetermined ins<length> is refused
## Summary

`N` is a real `Base` enum variant and there is **no ambiguity guard anywhere on the SPDI path**. So a repeat whose unit is `N` now emits a run of `N`s as the SPDI insertion string, even though the input names a length rather than any bases.

Six such rows exist in the release-asset corpus: `insN[18]`, `delinsN[341]` (x2), `insN[1200]`, `insN[320]`, `delinsN[155]`.

## Why this is the same state as an input that is still declined

`N[341]` carries the **length**, not the bases. That is precisely the information content of `ins341`, which #1967 deliberately continues to decline. The two inputs are equally undetermined, and they now get different treatment: one is refused, the other produces 341 `N` characters in an SPDI `insertion` field.

A consumer storing read counts against an SPDI triple would be keying on a string that asserts 341 specific bases the input never specified.

## How it was found

A corpus scan during review of #1967 returned **9** bare exact-repeat inserts where the PR's trailer says **3**. The six-row difference is exactly this class. `apply_alphabet` only uppercases; grepping both `main` and the branch found no ambiguity guard on the path.

## Scope

**Not a blocker for #1967** — the sibling short-form arm already ships this behaviour on `main`, so the PR is not introducing the class, only widening the set of spellings that reach it.

What #1967's trailer does claim, and should not without qualification, is that *"genuinely undetermined still declines"*. That is true of `ins341` and false of `insN[341]`.

## The question to settle

Should an `N`-unit repeat be refused on the SPDI path, in line with `ins<length>`, or is emitting `N` runs intended? If it is intended, the reasoning belongs in a ruling record rather than in the absence of a guard — `adjudication-precedence-order` makes "the decision is recorded, not inferred from behaviour" part of its ruling.

Worth noting the precedent from tonight: `c.pterdel`, `c.qterdel` and `c.1del` all produced the bit-identical SPDI triple before #1747 refused them. A plausible-but-wrong answer that a consumer can store is worse than a refusal.

nh13:
Closes #1975

## Summary

An `N`-unit (or `N`-containing) repeat on the HGVS→SPDI path expanded into the SPDI `insertion` string as literal `N`s — `insN[341]` emitted 341 `N`s, `insAN[3]` emitted `ANANAN`. An `N` names a **length**, not bases (it denotes "any of A, C, G or T"), so this asserts specific bases the input never specified: a storable-but-wrong SPDI triple a consumer can key read counts on. That is the same information content as the length-only `ins<length>` shape (`ins341`), which SPDI already refuses (#1967).

This refuses it instead, with `UnrepresentableInSpdi` (decided from the description alone; no provider changes the answer). The guard sits at the shared expansion leaf `expand_repeat_unit` — covering top-level `insN[n]` / `delinsN[n]` and single-part and bracket `Complex` shapes — and at the short-form `NaEdit::Repeat` arm (`g.pos_posN[n]`), before any reference read. A concrete-base repeat (`insACGT[3]`, `insA[10]`) is untouched; only undetermined content refuses.

Per the #1747 precedent (`c.pterdel`/`c.qterdel`/`c.1del`, all refused for producing storable-but-wrong SPDI triples), a plausible-but-wrong answer a consumer can persist is worse than a clean decline.

## Adjudication

Recorded as the `spdi-n-unit-repeat-refusal` ruling in `hgvs_spec_normalization_overrides.json` — a `README.md` rule 5 (silent-limb) **house choice**: the HGVS recommendations govern HGVS descriptions and are silent on SPDI conversion, and `N` is a legal HGVS symbol (`general.md:47`), so no clause governs. The derived ledger artifacts (`clause_ruling_index.rs` generated block, `docs/NORMALIZATION_CONTRACT.md`, the `CLAUDE.md` decided-rulings table, `RULING_STATUSES`) are regenerated/updated in the same change.

## Tests

Six unit tests in `src/spdi/convert.rs`: the five refusal shapes (`insN[18]`, `delinsN[341]`, `insAN[3]`, compound `delins[N[3];A]`, short-form `g.100_105N[6]`) and one negative control (`insACGT[3]` still expands). Full local gate green: `cargo fmt`, `cargo clippy --all-features --all-targets -D warnings`, `cargo nextest run --features dev --no-fail-fast` (11373 passed), the oracle suite (11203 passed), and `pytest tests/python/` (1137 passed).

Representation-Change: SPDI-path decline only; 0 normalized rows move. On the HGVS→SPDI path, `N`-unit / `N`-containing repeats (`insN[k]`, `delinsN[k]`) now decline (`UnrepresentableInSpdi`) instead of emitting a run of literal `N`s — 6 such inputs in the release-asset corpus (`insN[18]`, `delinsN[341]` ×2, `insN[1200]`, `insN[320]`, `delinsN[155]`) go from a previously-emitted (wrong) SPDI triple to a clean decline, a real change for an SPDI consumer. This is not a normalization change: `dump_normalized_corpus` before/after is byte-identical and `--verify-spdi` is unchanged (13 differ / 67801 agree / 28368 unverifiable both sides), a structural zero because the shape-family corpus alphabet is {A,C,G,T} and builds no `N`.


nh13:
@coderabbitai pause

PR #2142, opened 2026-08-17
fix(normalize): refuse a past-cds-end coordinate on both the lone and cis-allele paths

### LINKED ISSUE #2018: adjudication: a past-cds-end position is refused alone and silently remapped inside an allele
## Summary

A `c.` position past the transcript's `cds_end` is **refused when it stands alone** and **silently accepted, and remapped, inside a cis allele**. One validation path rejects what another rewrites, and the rewrite changes what the description means without saying so.

The defect is the disagreement, independent of which of the two answers is the right one — which is the part that needs a ruling.

## Reproducer

Against a synthetic transcript whose `cds_end` is `c.21`:

```
NM_SEAM.1:c.22del                 -> REFUSED   (W4004)
NM_SEAM.1:c.[22_23insG;*1G>A]     -> ACCEPTED  -> NM_SEAM.1:c.21dup
```

`c.22` names no base on the `c.` axis. `background/numbering.md:21` starts `c.1` at the `A` of the initiator and ends the CDS numbering at "the last nucleotide of the translation termination (stop) codon"; past that the axis is `*N`. So `c.22` is not a coordinate this axis has, and the lone description says so.

Inside the allele the same coordinate is remapped across the seam to `*1` and the allele normalizes to `c.21dup`. Nothing in the output records that a coordinate was reinterpreted.

## What needs deciding

Which stage owns the refusal.

- If the lone refusal is right, the allele path needs the same check, and `c.[22_23insG;*1G>A]` becomes `W4004` — a user-visible strict-mode change.
- If the remap is right, then `c.22del` should also be remapped rather than refused, and the lone path is over-strict.

Either way the two paths must agree. Ferro currently gets a different answer depending on whether the coordinate has a sibling, which is not a property the spec makes meaningful.

## Guard

`tests/it/insertion_adjacency_defects.rs::pinned_past_cds_end_position_is_refused_alone_but_accepted_in_an_allele` pins the current acceptance and says in its own body that the refusal is the expected behaviour, so it fires as a tripwire when the two paths are reconciled in either direction.

Related but distinct: #1909 (a `cds_end` past the sequence end panics `validate_cds_pos`) and #1816 (extending the sequence-first body across a CDS boundary). This one is about two validation paths disagreeing on a coordinate both can see.

nh13:
Closes #2018

## Problem

A plain `c.N` coordinate with N past the transcript's `cds_end` names no base the coding axis defines: `numbering.md:21` ends `c.` numbering at the last base of the stop codon and `:30` opens the `*` zone the moment it ends, so `c.22` on a transcript whose CDS ends at `c.21` is out-of-scheme numbering for `c.*1`, not a second name for it.

The lone-position path already handled this via the bounds gate in `normalize_cds` (strict refuses with `W4004`, lenient repairs to the `c.*N` spelling with a recorded `W4004`). The cis-allele path never reached that gate: `merge::collapse_overlapping_cis_edits` converts each member to a sequence coordinate before per-member normalization, and a plain past-cds-end `c.N` classifies as an in-range `Region::Cds` position mapping to the same sequence coordinate as its `c.*N` spelling — so the coordinate was silently remapped across the CDS/3'UTR seam with no diagnostic, in every mode. The same coordinate got two answers depending only on whether it had a sibling.

## Fix

Emit `W4004` on the raw allele members before the merge reinterprets them, mirroring how `detect_insertion_overlaps` emits `W5002` pre-merge. The finding then flows through the one machinery the lone path uses: the strict-mode ladder in `normalize_core_checked` promotes it to a typed refusal, while lenient keeps it as a recorded correction and silent suppresses the message — so both paths apply one rule, mode-dependent, identically. The `n.`, `c.-N` and `c.*N` bound checks are unchanged.

## Adjudication

Records the operator ruling `past-cds-end-coordinate-is-non-conformant` (decided, 2026-08-16), which follows the already-decided `absolute-prohibition-enforcement-stage` for the mode policy rather than introducing any new one. Flips the `insertion_adjacency_defects` tripwire guard to assert the reconciled behaviour, and adds mode-by-path coverage in `tests/it/issue_2018_past_cds_end_mode_dependent.rs`.

## Tests

New: `tests/it/issue_2018_past_cds_end_mode_dependent.rs` — lone/allele agreement on a past-cds-end coordinate, strict refusal in both collapsing and non-collapsing cis alleles, lenient repair-and-disclose. Full Rust suite (11396 passed) and Python binding suite (1137 passed) are green locally.

## Representation-Change

Representation-Change: strict mode now REFUSES a cis allele carrying a past-`cds_end` `c.` member that it previously remapped and normalized without any diagnostic; lenient and silent modes move NO output (the repaired `c.*N` form was already emitted, because the coordinate maps into the 3'UTR) but lenient now DISCLOSES the correction as `W4004`. The lone-position path is unchanged in every mode.

The synthetic corpus instrument (`examples/dump_normalized_corpus.rs`) is STRUCTURALLY BLIND to this change and cannot quantify it: the change keys on a cis-allele member whose `c.N` exceeds a real transcript `cds_end`, and the corpus generator builds transcripts with `CDS_START = 1` and geometry that never emits a past-`cds_end` member inside a cis allele, so a corpus diff is a structural zero rather than evidence of safety. The behavioural impact is the strict refusal described above, exercised against real `cds_end` values by the new tests.


nh13:
@coderabbitai pause

Issue #905, opened 2026-07-01
test(conformance): include out-of-band-derived artifacts in the reference-identity signature (#764 hardening)


nh13:
## Problem

The reference-identity hash (#764, `reference_identity_from_manifest` in `tests/it/mutalyzer_normalize_tests.rs`) pins the empty-projection `.count` baselines to the prepared reference they were blessed against. It hashes `prepared_at`, `transcript_count`, and the **basenames** of the cdot / genome / transcript-fasta artifacts — but **not** the content-bearing artifacts that can change **out-of-band** (regenerated by a targeted `ferro prepare --derive-*` / `--backfill-*` step rather than a full re-prepare, so `prepared_at` doesn't move):

- `derived_transcript_placements` (the #790/#795 Jun-20 addition — added without bumping the identity, which is what let the `28288ed4c39ee856` hash stay valid through a real content change)
- `ng_hosted_transcripts` (the #859/#871 wiring — also not part of the identity, so wiring/unwiring it is identity-invisible)
- `derived_refseqgene_placements`, `refseqgene_alignments`(`_grch37`), `refseqgene_summary`, `supplemental_fasta`, `backfill_transcripts_fasta`, `legacy_genbank_*`, `refseqgene_fastas`, `lrg_*` — all content-bearing, none in the signature.

**Consequence:** a content change to any of these does not bump the identity, so the empty-projection count gate keeps *enforcing* against a stale baseline (or, worse, silently stays "valid" through a real reference change) — the stale-ledger confusion documented across #790/#795/#859/#871.

## Fix (two parts)

1. **Basename part (partial):** add the content-bearing manifest fields above to the identity signature. This catches an artifact being **added/removed** and **version-name changes** (e.g. a re-versioned file). NOTE this is *insufficient on its own* for constant-named artifacts (`derived_transcript_placements.json` never changes name), so it does **not** catch **in-place content updates**.
2. **Content-stamp part (root fix):** have the out-of-band `ferro prepare --derive-transcript-placements` / `--derive-ng-placements` / `--backfill-transcripts` steps write a `derived_at` timestamp (or a content hash / record count) for each derived artifact into the manifest, and include those in the identity signature. Then an **in-place** content regeneration bumps the identity even without a full re-prepare. This is a `ferro prepare`-side (production-code) change — the actual root-cause fix.

## Sequencing (important)

[...]

PR #2146, opened 2026-08-17
refactor(prepare)!: drop the derived sequence_length from legacy transcript metadata

### LINKED ISSUE #2064: LegacyTranscript still carries a derived sequence_length, but removing it bumps the blessed reference identity

nh13:
Closes #2064.

## What this is

`LegacyTranscript` carried a persisted `sequence_length` that is **derived, not sourced**: its authority is the FASTA (`seq.len()`, recoverable from the `.fai` index), nothing in `src/` ever read it, and #2087 had already isolated it in a separate on-disk record type (`LegacyTranscriptRecord`) pending exactly this removal. This is the same hollow-derived-field shape #2058 removed from the benchmark supplemental sidecar; #2058 deliberately left the legacy artifacts alone because they are content-stamped, so removing the field here bumps the blessed reference identity — which is why it was split out as its own issue.

## The empirical measurement (before assuming the analogy held)

Measured on the blessed GRCh38 reference (`legacy_genbank_metadata`, the only legacy sidecar wired; `legacy_transcripts_metadata` is absent):

- **5 of 5** rows carried a **non-zero** `sequence_length`, and every one **equalled its FASTA record length exactly** (`meta_len == fasta_len` for all five) — i.e. fully redundant with the FASTA authority, recoverable from the `.fai`.
- **No typed consumer** reads the key: the legacy metadata JSON is write-only in production (nothing deserializes `LegacyMetadata` back), and the field's only readers were its own constructor and in-module/integration tests.

This differs from #2058's supplemental sidecar (where 116,572 of 140,027 rows were literal `0`) only in that these values are non-zero — but they are still derived/redundant, not independently sourced. So the analogy holds and the field is safe to delete rather than document-as-sourced.

## The change

Remove `sequence_length` so it is **unrepresentable** rather than merely unread. With the field gone, the sourced authority (`LegacyTranscript`) and the on-disk record are identical, so the two types collapse into one: `LegacyMetadata.transcripts` now holds `LegacyTranscript` directly, and `LegacyTranscript::into_record`, `LegacyTranscriptRecord::sourced`, and `LegacyMetadata::sourced_transcripts` go with it. An older on-disk file that still carries the field continues to load (serde ignores the now-unknown key).

## Reference-identity bump (the reason this is its own issue)

Removing the field changes the serialized bytes of the content-stamped `legacy_genbank_metadata` / `legacy_transcripts_metadata` artifacts (#905), so it **deliberately bumps the blessed reference identity**:

| | identity |
|---|---|
| before | `aa8b3246d83055cc` |
| after | `f329e62bd95d8020` |

The new identity was computed by the real `reference_identity` path (validated by first reproducing the old `aa8b3246d83055cc` against the unmodified reference) and re-confirmed by the actual `BLESS_EMPTY_PROJECTION` bless path — not a hand-port. The pinned baselines were re-blessed to the new value:

- `tests/fixtures/mutalyzer-normalize/empty-projection/{coding_protein_descriptions,protein_description}.count`
- `comparator_provenance.reference_identity` in `tests/fixtures/mutalyzer-normalize/cases.json`
- the golden bytes + content-stamp digest in `tests/it/legacy_integration.rs`

**The empty-projection counts themselves are unchanged (4 and 1)**, which confirms this moves **no normalized output** — only the reference identity. `src/prepare` is not a watched prefix (watched = `normalize`/`hgvs`/`spdi`/`project`/`reference`/`error_handling`), so no `Representation-Change:` trailer is required.

## Notes

- The operator-owned prepared references on scratch/backup still carry the pre-removal legacy files; making a local `FERRO_MANIFEST` run consistent with the new pins requires regenerating those small sidecars (a minimal, order-preserving field drop) on both volumes. This was verified here against a throwaway symlink copy so no shared volume or concurrent session was disturbed.
- `multi_member_cis_axis::axis_normalized` is unrelated to this change and is not affected by it (the census references none of the symbols touched here; it reproduces identically with and without this change). It early-returns without a provisioned manifest, so PR CI does not exercise it.


nh13:
@coderabbitai pause

